PR #5789, opened 2015-08-14
Stringify into window.env only what is used on the client

paglias:


paglias:
Also it removes everything that can be calculated only once (like requires, variables valid for every user, ...) from the locals middleware so that it's not calculated at every request

website/src/controllers/dataexport.js
```diff
@@ -99,10 +99,12 @@ expressres.jsonstring = function(obj, headers, status) {
 
 
 dataexport.avatarPage = function(req, res) {
+  var env = _.defaults({user:user}, res.locals.habitrpg);
```
crookedneighbor: user is defined on line 104

website/src/middlewares/locals.js
```diff
@@ -29,23 +53,18 @@ module.exports = function(req, res, next) {
...
 
   // Put query-string party (& guild but use partyInvite for backward compatibility)
   // invitations into session to be handled later
-  try{
-    req.session.partyInvite = JSON.parse(utils.decrypt(req.query.partyInvite));
-  } catch(e){}
+  if(req.query.partyInvite){
+    try{
+      req.session.partyInvite = JSON.parse(utils.decrypt(req.query.partyInvite));
+    } catch(e){}
```
gisikw, line 67: I know I'm commenting on things unrelated, but...empty catches make me nervous. Are we _sure_ we shouldn't do any error-handling here? And if so, let's add a trailing comment to the catch to that effect.

website/src/controllers/dataexport.js
paglias: right, reverted to original version

website/src/middlewares/locals.js
paglias, line 67: I was unsure on wheter we could remove the try/catch at all since I added that `if`. I think it's there because `utils.decrypt` and `JSON.parse` throws an error and would break everything if for any reason a user passed an invalid query string

gisikw, line 67: Oh yeah, I'm sure it's still valuable to have the catch in there, but if it does break, we should be responding in some way (probably throwing an error code)

paglias, line 67: Hmm, I'm not sure for the user would be useful to see an error code, maybe for us but until we don't have proper logging in place I'm not sure. I can add a TODO comment

paglias:
Dataexport page is broken, don't merge

paglias:
There are a few other tweaks that could be done like not loading `mods` on front page and everything except `user` on avatar export but the most of it has been done and any further tweak will become more difficult to maintain.

paglias:
After endlesses fixes, it's ready, merging

PR #5792, opened 2015-08-15
remove unused default tasks

Alys:
Removes commented-out code that used to give users default tasks and removes matching unused locale strings. Fixes https://github.com/HabitRPG/habitrpg/issues/5631#issuecomment-130619901 (comment by GitHubSphinx which identified a problem with one of those strings).

I'm assuming we don't need these unused default tasks for the foreseeable future, and if we ever do, we can get them from version history. However if there's any disagreements, I'll cancel this PR.

I'll merge in a couple of days if no complaints.


crookedneighbor:
No need to wait for this!

PR #5794, opened 2015-08-15
Upgrade to Express 4

paglias:
Absolutely do not merge, needs extra testing
- ~~Request limit set to 100kb~~ delayed
- ~~res.json() -> res.status().json()~~ delayed until we have batchUpdate

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,19 +4,24 @@
   "version": "0.0.0-152",
   "main": "./website/src/server.js",
   "dependencies": {
+    "accepts": "^1.2.12",
     "amazon-payments": "0.0.4",
     "amplitude": "^1.0.6",
     "async": "~0.9.0",
     "aws-sdk": "^2.0.25",
     "babel": "^5.5.4",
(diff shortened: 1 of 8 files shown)


paglias:
Useful links:
- Migrating from 3.x to 4.x https://github.com/strongloop/express/wiki/Migrating-from-3.x-to-4.x
- Moving to Express 4 http://expressjs.com/guide/migrating-4.html

paglias:
This is ready except than for a few deprecations that we should fix, they work for now but will stop in the next major version:

```
deprecate res.json(status, obj) -- use res.status(status).json(obj) instead
deprecate res.jsonp(status, obj) -- use res.status(status).jsonp(obj) instead
deprecate res.send(status, body) -- use res.status(status).send(body) instead
express deprecated res.send(status): Use res.sendStatus(status) instead
```

The fact is there are around ~200 occurences of them...

paglias:
Will reopen later in a bigger PR

PR #5796, opened 2015-08-15
Change 2 locale strings to exclude html character codes not shown on the front page

### LINKED ISSUE #5571: Bug on frontpage: Dailies&mldr
capture

ghost:
Strings dreimQuote and teamSample3 on the front page contain symbol
codes `&mldr;` and `&rsquo;` which are not rendered and shown correctly.

Screenshot

Instead of replacing symbol codes with plain text `...` and `'` respectively (as suggested previously) I've decided to just enable html rendering for the following reasons:
- there must have been some reasons symbol codes were used initially =)
- this fix won't require translators revisit translations
- there are already a few comments for these strings on Transifex and I'll add some more clarification if/when merged - so I believe there's not much confusion among translators caused by these symbol codes
- previous alternative suggestion in the ticket discussion haven't received any comments/support
- I'm making a commit to bring some attention to this minor issue as it's the front page we have it on =) Even if the quote is soon to be replaced with the new one, the other string will still be there, I guess
- I'll be happy to change it if there are other suggestions

Fixed: Screenshot

Closes #5571


crookedneighbor:
In general, we should limit what is allowed to use unescaped html. I know it's not super convenient for the translators, but a better solution will be to change the original strings.

paglias:
In this case I don't think there's anything wrong with having unescaped
html, it's not user input and we already have it for a lot of strings
Il 15/ago/2015 10:43 PM, "Blade Barringer" notifications@github.com ha
scritto:

ghost:
@paglias @crookedneighbor thank you for your comments! 
I believe translators are quite accustomed to revisiting strings as changes are made to English source, so it's OK. In this case it's just the matter of choosing whether to keep character codes or switch to plain text. And I believe there's no strict reasoning preventing from using either of the alternatives.

I could find no other occurrences of these symbols on the site. I've traced the changes back to front page being untranslatable - and it seems the codes have just been used following the common practices, found on other sites.

Visual changes are very difficult to notice - a screenshot from the previous discussion: screenshot.

So I'm switching the PR back to my initial suggestion of replacing mldr and rsquo with plain text `...` and `'` respectively because
- it's just the way all other occurrences of `...` and `'` are handled on the site
- it won't support further possible inconsistencies 
- it's easier to understand for translators
- it doesn't overuse unescaped html
- (once again) the change in visual appearance is rather difficult to notice and I don't think there might be any functional change

**EDIT**
Screenshot of changes (combined) 
scr

crookedneighbor:
:+1: 

Notated.

PR #5798, opened 2015-08-16
prevent Orb of Rebirth from removing free gear (equipment that has zero gold cost) - partial fix for https://github.com/HabitRPG/habitrpg/issues/5750

Alys:
Partial fix for https://github.com/HabitRPG/habitrpg/issues/5750 "Tasks page loading/running extremely slowly after rebirth -- too many equipment/gear pieces in Rewards column"

This PR:
- Causes the Orb of Rebirth to not remove any free equipment.
- Adds a test for not losing free gear on rebirth.
- Adjusts the text on the Orb's modal to explain that free gear is not lost (underlined in screenshot below).
- Adjusts wording in the Seasonal Store to remove mention of unlocking the Item Shop: _"If you've used the Orb of Rebirth, you can repurchase this equipment in the Rewards Column ~~after you unlock the Item Shop~~."_

(NB: The Orb also does not remove the Training Sword, but that behaviour already existed and was not added by this PR.)

@lemoness, do you want this released on any particular date?

screen shot 2015-08-16 at 1 16 38 pm


SabreCat:
We might consider removing the "as if you had created a new account" clause--that's rather an overstatement of what it does.

crookedneighbor:
I agree

lemoness:
Agree as well

Alys:
I've removed "as if you had created a new account".

@lemoness, do you want this released on any particular date? Or are there more changes you'd like?

lemoness:
Nope, it can go out at any time! Just notify me when it is live.

PR #5806, opened 2015-08-17
Update information about max levels for buffs.

### LINKED ISSUE #4084: Stats page doesn't explain that effects of level are capped at level 100 (e.g., Perfect Day bonus maxes out)
On the stats and achievements page, it says, 

"X Perfect Days
Completed all active Dailies on 134 days. With this achievement you get a +level/2 buff to all attributes for the next day."

My character is level 118. If this were correct, my buff to all attributes should be 59, but in fact it is only 50. I think the buff from the perfect day has been set to max out at level 100, contrary to the text.

UUID: 66440307-9628-4eca-b448-ba008e82a3b3
Browser: Chrome
OS: Windows 7

gbear605:
Fix #4084.


crookedneighbor:
Seems reasonable to me. @Alys This is accurate, correct?

Alys:
Correct.

@Gbear605 What's your User ID (from Settings -> API)?

gbear605:
@Alys 2893ce69-f5df-41a5-8d4f-59fe286660cf

Alys:
Thank you @Gbear605! I've noted that as putting you half way to tier 2.

PR #5848, opened 2015-08-26
Added ability to close menu when a child element has closeMenu directive

### LINKED ISSUE #5845: Guild chat notification window closing when x is clicked
In the past, you could dismiss one or more of the notifications in the little guild/party chat notification window and then click on the guild you wanted to read. Now the box closes after each click, which is not as user-friendly as it was before. Someone has mentioned this problem in the Tavern, so I know it's not just something odd that's happened to my browser. 

It happens in both Chrome and Firefox on my pc running Windows 10. When I asked about this in the Aspiring Socialites guild, @Alys said that it was an unintended consequence of a fix to make menus no longer stay open inappropriately when you change to another page. However, in this particular case, the new behaviour is inconvenient so she suggested that things should be adjusted to get the old behaviour back for this particular situation.

TheHollidayInn:
To fix #5845 , I have included the ability for a child element to have the closeMenu directive and still change the scope. I change the structure of the _expandMenu variable to an object to deal with $parent scopes. Here is a stackoverflow link about the problem: http://stackoverflow.com/questions/16928341/update-parent-scope-variable .

### DIFF
--- a/website/public/js/directives/expand-menu.directive.js
+++ b/website/public/js/directives/expand-menu.directive.js
@@ -11,9 +11,7 @@
       restrict: 'A',
       link: function($scope, element, attrs) {
         element.on('click', function(event) {
-          if (!$scope._expandedMenu) {
-            $scope._expandedMenu = {};
-          }
+          $scope._expandedMenu = $scope._expandedMenu || {};
           $scope._expandedMenu.menu = ($scope._expandedMenu.menu === attrs.menu) ? null : attrs.menu;
           $scope.$apply()
         });
--- a/website/public/js/directives/expand-menu.directive.js
+++ b/website/public/js/directives/expand-menu.directive.js
@@ -11,7 +11,10 @@
       restrict: 'A',
       link: function($scope, element, attrs) {
         element.on('click', function(event) {
-          $scope._expandedMenu = ($scope._expandedMenu === attrs.menu) ? null : attrs.menu;
+          if (!$scope._expandedMenu) {
+            $scope._expandedMenu = {};
+          }
+          $scope._expandedMenu.menu = ($scope._expandedMenu.menu === attrs.menu) ? null : attrs.menu;
           $scope.$apply()
         });
       }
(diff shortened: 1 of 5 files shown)


website/public/js/directives/expand-menu.directive.js
```diff
@@ -11,7 +11,10 @@
       restrict: 'A',
       link: function($scope, element, attrs) {
         element.on('click', function(event) {
-          $scope._expandedMenu = ($scope._expandedMenu === attrs.menu) ? null : attrs.menu;
+          if (!$scope._expandedMenu) {
```
crookedneighbor: This can be simplified to:

``` js
$scope._expandedMenu = $scope._expandedMenu || {};
```

TheHollidayInn:
Updated!

website/public/js/directives/expand-menu.directive.js
```diff
@@ -11,7 +11,8 @@
       restrict: 'A',
       link: function($scope, element, attrs) {
         element.on('click', function(event) {
-          $scope._expandedMenu = ($scope._expandedMenu === attrs.menu) ? null : attrs.menu;
+          $scope._expandedMenu = $scope._expandedMenu || {};
+          $scope._expandedMenu.menu = ($scope._expandedMenu.menu === attrs.menu) ? null : attrs.menu;
```
crookedneighbor, line 15: Just curious, is there a technical reason it needs to be an object?

TheHollidayInn, line 15: Yea, I explained it above in my initial PR post.

crookedneighbor, line 15: Sorry, just had a baby and my brain is :dizzy:

crookedneighbor, line 15: Makes sense :+1:

crookedneighbor:
Looks reasonable to me. Thanks!

Notated.

TheHollidayInn:
No worries. And, congratulations!!!

Dewines:
Thanks for sorting this out. The chat notification seems to be working just like it did before and I can dismiss the ones that I don't want to read right now without the window closing.

PR #5850, opened 2015-08-27
Set image-rendering to pixelated

### LINKED ISSUE #5846: Pixel art images should use image-rendering:pixelated
https://developer.mozilla.org/en/docs/Web/CSS/image-rendering

Basically, whenever you zoom in, the browser tries to "correct" the image without knowing it's pixel art, meaning that the image looks kinda jpegy or something. Apparently image-rendering is a new thing that allows you to specify "hey, this is pixel art, don't do any of that magic" and preserves the pixel-artiness of it.

First image is "normal", second is with image-rendering set to pixelated.

nonpixelated
pixelated

RossBrunton:
Closes #5846 , see there for more details.

### DIFF
--- a/website/public/css/index.styl
+++ b/website/public/css/index.styl
@@ -194,3 +194,11 @@ a.label
 
 .markdown-preview markdown code
   white-space inherit
+
+// By default everything should render as pixelart, except images from img tags (which tend to be things like gravatars, 
+// screenshots and so on)
+*
+  image-rendering: pixelated
+
+img
+  image-rendering: auto


crookedneighbor:
Would it make more sense to add this style to the spritesmith css template?

RossBrunton:
Maybe? I'm sorry but I don't quite know my way around your codebase. >_<

I only put it in index.styl because I couldn't find anything more appropriate in website/public/css.

crookedneighbor:
Why don't you just try adding it here and see what happens: https://github.com/HabitRPG/habitrpg/blob/develop/common/css/css.template.mustache

RossBrunton:
Okay, I added them to the "item blocks" or whatever, and it seems to work. However, this does seem to result in a lot of "image-rendering" properties rules in the file, which could be a problem. But anyway, I guess it's your call where best to put it.

crookedneighbor:
I think you're right, it makes more sense to use the `*` version.

There are a bunch of pixel art images in the community guidelines, so instead of applying `image-rendering: auto` to all `img` tags, I created a class to override it where needed and applied it to those images that are not pixel art. 

An alternative solution would be to apply it all img tags, and then have a `img.pixelated` rule that overrides that.

RossBrunton:
5c84b8fa-8744-4604-9b26-cc90ffcddfbf , thanks.

Alys:
Tier awarded. Thank you @RossBrunton! You can read more about [contributor tiers and rewards here](http://habitrpg.wikia.com/wiki/Contributor_Rewards).

PR #5863, opened 2015-08-31
Directives test task focus

TheHollidayInn:
This tests the task-focus directive.


TheHollidayInn:
I changed the element from an anchor element to an input element, and that fixed the issue.

crookedneighbor:
Good work. Notated. 

I ended up removing the focusMe directive entirely and renamed this directive focusElement to make it more generic. Works great. Thanks for your work on writing the test.

TheHollidayInn:
Thanks! And, cool, I think that is much better.

PR #5864, opened 2015-08-31
Fixed the wrong redirect

gisikw:


website/src/middlewares/redirects.js
```diff
@@ -21,7 +21,7 @@ function isProxied(req) {
 
 module.exports.forceSSL = function(req, res, next){
   if(isHTTP(req) && !isProxied(req)) {
-    return res.redirect(301, BASE_URL + req.url);
```
paglias, line 24: should't be a 301? we want to force SSL for all request

PR #5867, opened 2015-08-31
WIP: Allow content to be inactive

crookedneighbor:
# :warning: DO NOT MERGE :warning:
## Problem

Content needs to be available on the site in advance of it being released.

_Use Case:_ User sees new quest has been released and immediately invites party to quest. Meanwhile, a party member, who had the site open before the deploy, and thus does not 
## Solution

Provide inactive statuses for content
## Still needs to be done
- [ ] Allow armoire items to be inactive
- [ ] Allow quests to be inactive
- [ ] Allow mystery sets to be inactive
- [ ] Prevent purchasing inactive items via api
- [ ] Prevent inactive items from being passed back in content call
## Concerns

Is this the best way to go?

Maybe it's enough to look up quest details via a content api route?

### DIFF
--- a/common/script/content.coffee
+++ b/common/script/content.coffee
@@ -2225,6 +2225,7 @@ api.backgrounds =
       text: t('backgroundTwinklyPartyLightsText')
       notes: t('backgroundTwinklyPartyLightsNotes')
   backgrounds092015:
+    inactive: true
     market:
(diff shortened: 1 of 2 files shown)


crookedneighbor:
I don't think this is quite the right way to do this yet. Going to close this and ruminate over it some more.

PR #5872, opened 2015-09-01
Firebase changes

paglias:


test/api/groups.coffee
```diff
@@ -151,7 +151,7 @@ describe "Guilds", ->
           done()
 
   context "removing users groups", ->
-    it "allows guild leaders to remove a member", (done) ->
+    it "allows guild leaders to remove a member (but not themselves)", (done) ->
```
paglias, line 154: @crookedneighbor I've modified a test to be sure a change is working, it's ok to modify it or you prefer a new test to be created?

crookedneighbor, line 154: Technically, each test should be testing one thing, ie: `it "allows guild leaders to remove a member"`, `it "does not allow guild leader to remove themselves"`, `it "does not allow non-guild leaders to remove a member"`

Though, that said, we're deprecating these tests soon in favor of more app-agnostic api tests. I'm hoping to set up the new testing suite soon (in ES2015!), so in the meantime, do whatever you want with these

paglias:
@crookedneighbor I plan to do some other changes, I'll add a "WIP: do not merge" at the top

paglias:
in the next PR

crookedneighbor:
Makes sense. This one seemed more like a minor bug fix than anything related specifically to firebase.

paglias:
Yes, I just started looking at the code, found this small bug and decided to fix it but it's not specifically related

PR #5875, opened 2015-09-02
Conditional buy gem modal

AnitaYCheng:
- Customizes buy-gems modal that originates from the header menu, depending on the subscription status of the user.
- For non-subscriber, modal provides 3 methods of buying gems: by purchasing, subscribing to buy gems with gold, or suggestions about contributing or winning challenges.
- Produces a shortcut to buy gems with gold for subscribers under their conversion cap.
- If subscriber has exceeded their gold-to-gem conversion cap, modal provides shortcut to purchase gems.
- Cleans up wording for subscription mixin.
- Uses branded payment buttons if available.


website/public/css/shared.styl
```diff
@@ -10,11 +10,44 @@
     th.feature-name
       text-align center
       font-weight: bold
+
 .dashed-underline
   border-bottom:1px dashed #000
   text-decoration:none
   cursor pointer
 
+.form-inline .form-control
```
crookedneighbor: We should be cautious about overriding bootstrap defaults.

website/views/shared/modals/buy-gems.jade
```diff
@@ -1,30 +1,76 @@
...
+          
+            div(ng-include="'partials/options.settings.subscription.html'")
+
+          .container-fluid
+            h3=env.t('freeGemsTitle')
+            p!=env.t('subFreeGemsHow')
 
   .modal-footer
     button.btn.btn-default(ng-click='$close()')=env.t('cancel')
+
```
crookedneighbor: Why the extra new line here?

AnitaYCheng: Oof not sure! Probably an error in copy-pasting!

website/public/css/shared.styl
AnitaYCheng: I agree. Should I add an extra class to the markup?

AnitaYCheng: Oh, I also found a missing colon with line 17 that I'll change.

AnitaYCheng:
I've added specificity to the forms class involved, and removed the extra line.

AnitaYCheng:
Thanks for the suggestions @crookedneighbor, I've added specificity to the forms class involved, and removed the extra line.

Alys:
Some feedback:

For a non-subscriber, the "Obtain Gems for Free" section is at the bottom of the modal, and that means it's beneath the fold for a small screen,

Alys:
Some feedback:

For a non-subscriber, the "Obtain Gems for Free" section is at the bottom of the modal, and that means it's beneath the fold for a small screen (my laptop, tablets probably). This is what I see if I don't scroll down:
screen shot 2015-10-04 at 6 55 45 am

Even an impatient user with a large screen might not notice the "Obtain Gems for Free" section. If we moved it up to between "Purchase Gems" and "Become a Subscriber", then it would help to reduce the endless stream of questions like "how do I get gems without money" and it would give a larger separate between the two purchase options, further reducing the chance of people clicking on the wrong button.

It is common to see new subscribers ask why they didn't get gems because they think that getting gems immediately is part of the subscription benefits. Occasionally they even lodge a PayPal dispute against their subscription payment because of this. Can we add words to explain that you do not get gems immediately, without it sounding like a negative thing? @lemoness ?

When a subscriber clicks on the gem in the header, they see only the option to buy gems with gold. I would really it if they also saw their subscription summary and cancel button, as they can now on the production site. If they've used the gem button to create their subscription, they might not know about the Settings -> Subscription page. We sometimes get PayPal disputes because people can't work out how to cancel a subscription, so keeping it as obvious as possible will prevent that problem getting worse. I think it also wouldn't hurt to show the "Obtain Gems for Free" section because if they've seen the links there before, that might be where they go when they want to find them again.

[EDIT: Ignore this bit. The subscription had already been cancelled in the account I was looking at.]  _When a subscriber goes to Settings -> Subscription, they no longer see the cancel button. It must be there._

Alys:
Sorry, ignore this part:
"When a subscriber goes to Settings -> Subscription, they no longer see the cancel button. It must be there."
I was using an account that had a cancelled subscription.

patriq:
Wait can you get gems for free?

Alys:
Sorry, ignore this part:
"When a subscriber goes to Settings -> Subscription, they no longer see the cancel button. It must be there."
I was using an account that had a cancelled subscription.

I guess that means that for a stupid user, the screen doesn't indicate clearly enough that the subscription has already been cancelled. ;)  Can we change "Subscription will become inactive on 01/10/2016" to "Your subscription has been cancelled and will become inactive on 01/10/2016"?

Or maybe leave those words as they are now, but have the extra words "already cancelled" at the start on a red background so that they look something like like the cancellation button. That would help anyone who's looking for the cancellation button if they'd forgotten that they'd already cancelled. When we advise users about how to cancel, we tell them to look for the red button, and users do forget about actions they've already taken. If they're in a period of severe financial stress, they might be too stressed to read words and will just be looking for visual clues.

Can the "Resubscribe" heading be moved to above the subscription options? See the orange arrow:
screen shot 2015-10-04 at 7 14 51 am

Alys:
@PatriqDesigns Only from challenges and by contributing to Habitica. For more information, please go to Help -> Ask a Question at the habitica.com website rather than asking here. This conversation is specifically about changing the appearance of the subscription screens.

Alys:
This bit is good. ;)

screen shot 2015-10-04 at 7 31 14 am

paglias:
@AnitaYCheng one thing I noticed is that when you're subscribed it looks like there isn't anymore an option to buy gems directly (20 gems for 5$)

lemoness:
Re: the "no gems immediately" explanation - what if instead, after you
subscribed, it triggered a popupthat contained a thank-you message and some
information, like how to buy gems and how to get mystic hourglasses? I
think it might help subscription churn in addition to reducing questions.

Alys:
@lemoness  I like that idea a lot! Can it have a big, obvious "How to Get your Gems" heading with a gem image next to it, to really draw the user's attention?

Should it mentioned using Fix Character Values to give yourself gold, to help reduce anger in those who were expecting gems immediately?

lemoness:
I think both of those are great ideas!

Possible tex (feedback welcome)t:

 Thank You For Subscribing!

[img: three subscriber boxes]

 You're the best! As a subscriber, you're helping Habitica stay strong. To
show our gratitude, you will receive:
- Unique subscriber equipment for each month you subscribe
- The ability to buy gems with gold
- Double your daily drop cap
  
  Plus, for every 3 consecutive months that you subscribe, you will earn a
  Mystic Hourglass, which unlocks exclusive items from the [Time Travelers](link).
  
  How to Get Your Gems [img: three gems]
- Go to the [Market](Inventory > Market)
- Scroll to the bottom
- Click the gem to buy gems with 20 gold each!

The longer you stay a subscriber, the more gems you can buy each month.
Want all your gems at once? Just go to Settings > Site > Fix Character
Values and give yourself as much gold as you need to buy all your gems.

Thanks again for subscribing!

crookedneighbor:
@AnitaYCheng Closing this now, but feel free to re-open it once the merge conflicts are sorted out and you're ready for it to be reviewed.

Issue #5871, opened 2015-09-01
Handle groups memberships and invitations issues


paglias:
There are currently several cases that causes groups memberships and invitations data not to be up to date anymore.
### When the last member of a group leaves (except public guilds)
- [x] the group should be deleted (possible PR is #5362, issue #5136 )
- [x] if guild, invitations to guild should be removed from invited users' profiles
- [x]  if party, invitations to party should be removed from invited users' profiles
- [x] challenges should be deleted
### When an user cancel his account
- [x] user should be removed from all the groups it belongs to 
- [ ] user should be removed from all the groups he's invited to
### When leader leaves
- [ ] Most senior member should be made leader #5737 
### When an invitation is rejected
- [ ] Invitations are not removed from group's invitations. See https://github.com/HabitRPG/habitrpg/blob/develop/website/public/js/controllers/groupsCtrl.js#L453 for guilds and https://github.com/HabitRPG/habitrpg/blob/develop/website/public/js/controllers/groupsCtrl.js#L538 for parties  -> create api route for this
### When user joins empty guild
- [x] should become the new leader
##

Alys:
"Most senior member should be made leader, it isn't?" -- Yes. I've updated the post to remove the question. 

Although we might want to make it the most senior member who has logged in within the past x days, to avoid inactive leaders.

Alys:
To really clean up all abandoned groups, we should remove all deleted user accounts from groups. That will likely produce a new set of groups that are now empty.

paglias:
Yes, it's a lot of work... but it really needs to be done for the new chat feature. @gisikw any idea?

TheHollidayInn:
Is this not fixed by my current PR 
"it should be removed from all the groups it belongs to (both as member or when invited)"
https://github.com/HabitRPG/habitrpg/pull/5264?

crookedneighbor:
I believe it is, but I haven't finished reviewing it yet.

TheHollidayInn:
Sweet. Just making sure. I'm going to try to work on this one, then: When user joins empty guild.

TheHollidayInn:
I think this has been fixed "When an invitation is rejected" when @crookedneighbor moved the group.js files into multiple files. The reject function seems to update the invitations. Is that correct?

paglias:
@TheHollidayInn they are removed from the user's invites but not from the groups'

TheHollidayInn:
[...]

PR #5877, opened 2015-09-02
Groups now make joining user the leader when they are empty

TheHollidayInn:
This is to address one of the requirements on #5871 . When a group is empty, we add the first user to join as the leader.

### DIFF
--- a/website/src/controllers/groups.js
+++ b/website/src/controllers/groups.js
@@ -462,6 +462,9 @@ api.join = function(req, res, next) {
   if(!isUserInvited) return res.json(401, {err: "Can't join a group you're not invited to."});
 
   if (!_.contains(group.members, user._id)){
+    if (group.members.length === 0) {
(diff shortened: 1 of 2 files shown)


website/src/controllers/groups.js
```diff
@@ -462,6 +462,9 @@ api.join = function(req, res, next) {
   if(!isUserInvited) return res.json(401, {err: "Can't join a group you're not invited to."});
 
   if (!_.contains(group.members, user._id)){
+    if ( group.members.length == 0 ) {
```
crookedneighbor: No surrounding spaces around `group.members.length == 0` and please use `===`.

TheHollidayInn:
Eh, sorry. I was coding some PHP earlier, so I got mixed up. Should be good now.

paglias:
@crookedneighbor does it looks good to you?

Issue #5836, opened 2015-08-24
"Old News" page is not very pretty


SabreCat:
We have a static page for viewing old news entries. However, it has several notable problems:
- None of the headings are bold, the way they are on Bailey's modal, making everything run together.
- The current news is included, which is not such a bad thing, except that...
- ...the "older news" link still appears, to no purpose.

zedutch:
@SabreCat Could you take a quick look at PR #5884? I might have handled this issue so I think it can be closed. Unless you have some remarks/comments about my solution or more suggestions for the design of this page?

PR #5884, opened 2015-09-03
Update the look of the static old news page (Fix for #5836)

zedutch:
This is a fix for #5836.
I've redesigned the static/old-news page to more or less follow the same design as Bailey's modal.
I've also removed the "Old news" link when you're already on the old news page. I did this using CSS, but there might be a better way of doing this using expressJS?

This is my first pull request for Habitica, so please let me know if I've changed things I shouldn't have or if I didn't follow some coding standard. I think I've read about everything there is on contributing on the wiki and in these READMEs but I might have missed something...

Anyway, I would love some feedback from more experienced Habitica-contributors.

### DIFF
--- a/website/views/shared/new-stuff.jade
+++ b/website/views/shared/new-stuff.jade
@@ -1,14 +1,14 @@
-h5 NEW ITEMS IN THE ENCHANTED ARMOIRE AND SEPTEMBER BACKGROUNDS!
+h5.big_news_header NEW ITEMS IN THE ENCHANTED ARMOIRE AND SEPTEMBER BACKGROUNDS!
   hr
   tr
     td
       .background_tavern.pull-right
-      h5 September Backgrounds Revealed
+      h5.small_news_header September Backgrounds Revealed
       p There are three new avatar backgrounds in the <a href='/#/options/profile/backgrounds'>Background Shop</a>! Now your avatar can visit the Habitica Tavern, shop in the Habitica Market, or ride mounts in the Habitica Stable! 
   tr
     td
       .promo_enchanted_armoire_201509.pull-right
-      h5 New Items in the Enchanted Armoire!
+      h5.small_news_header New Items in the Enchanted Armoire!
       p There is new equipment in the Enchanted Armoire, a 100 GP Reward in the Rewards Column which unlocks after you've attained Ultimate Gear! 
       br
       p Click on the Enchanted Armoire for a random chance at special Equipment, including the Yellow Hairbow, Red Floppy Hat, Gold Winged Staff, and the Plague Doctor Item Set! It may also give you random XP or food items. We'll be adding new equipment to it during the first week of each month, but even when you've exhausted the current supply, you can keep clicking for a chance at food and XP.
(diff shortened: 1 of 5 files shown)


crookedneighbor:
I don't think adding special classes to the headings is the right way to go.

I think it makes more sense to actually make them hierarchical, instead if using `h5` For everything. IE, the heading you've called `big_news_header` should actually be an `h1` tag. 

This may require some adjustments to the new stuff modal styles. 

After that, I think a new class, something like `.static-old-news`, should be added to the container div for the old news page. Then you can target `.static-old-news h1` and `.static-old-news h5` with special styles (if needed).

Further, I see what you did by hiding the `hr` tag and using border-bottom to simulate it, since the other ones are missing the `hr` tag, but I think it makes more sense to add those `hr` tags back in. There's weird formatting happening with the images that are pulled right, removing the border bottom rule and adding the `hr` tag back in to each news section _should_ fix it.

Alys:
Strictly speaking, there should be only one h1 heading on any web page, so the .big_news_header headings should be changed to h2. However Habitica's code might not (yet) follow that standard. This might be an appropriate time to start though.

zedutch:
I've used H4 instead of the big_news_header since that's more or less the same size. It felt kind of weird to use a bigger header (like H1 or H2) and then downsize it...
I've changed everything @crookedneighbor suggested.
I've also made sure Bailey's news modal isn't broken and didn't change too much. The title is 4px bigger though, but I don't think that's a problem? It doesn't look to weird to me anyway...

crookedneighbor:
You are right Alys (:baby: brain)

website/views/shared/new-stuff.jade
```diff
@@ -17,11 +17,12 @@ h5 NEW ITEMS IN THE ENCHANTED ARMOIRE AND SEPTEMBER BACKGROUNDS!
       p.small.muted by Lemoness and SabreCat
       p.small.muted Art by Starsystemic and Kiwibot
 
-hr
-a(href='/static/old-news', target='_blank') Read older news
+hr.remove-on-static
```
crookedneighbor: I'd try adding a surrounding div instead of applying this class twice:

``` jade
.remove-on-static
  hr 
  a(href...
```

Alys:
@crookedneighbor :baby::two_hearts::smiley:

@zedutchgandalf This is more an FYI in case it's useful to you later, rather than something that I think has to be changed now (because I'm not sure the rest of the site abides by this):

With HTML, it's always appropriate to use the correct h[1-6] tag for the text in question and then downsize its appearance. The number is never meant to be an indication of text size, but instead indicates the "level" of the heading. h1 is always the primary heading on the page and there should be only one of it (I suppose that would be "Habitica" in this case; if you want an analogy with books, then h1 would be the title of a book). h2 is always the next heading in the page's hierarchy, and there can be as many h2's as there are major sections on the page (chapter titles in a book). If a major section has subsections, then the title for each of those would be h3; sub-subsections would be h4, etc. 

Thus the h[1-6] tags are a way of identifying the "meaning" (semantics) of a piece of text - saying that "this text is a heading of level n". When you're merely looking at a web page, this might not seem to be important, but following these guidelines can have great benefits. For example, if a visually impaired person is having a webpage read to them by software, the software can use the h\* headings to determine what the major sections are and can give the user options to jump ahead to each section. Also, if h\* tags are used across a website in a consistently meaningful fashion (which I'm not sure we do yet), then you can apply site-wide styles to each level of h\* tag and be confident that the relatives sizes, boldness, etc will always be correct on every page.

If this page ends up needing further changes, it might be worth changing the h4 tags to h2 and adjusting the styles to match, but you might not want to do that if the page is basically finished now. :)

zedutch:
@Alys Yes, I definitely see what you mean. Especially from a semantics standpoint it would be better to use h2. But don't the h5 tags have to change to h3 then too?
But I'll definitely change it later today (and crookedneighbor's remove-on-static suggestion as well)

website/views/shared/new-stuff.jade
zedutch: @crookedneighbor Is this actually the way I should hide those elements? Or is there some way to make sure they're not even on the page instead of just hidden?

crookedneighbor:
@zedutchgandalf I think you're right about letting express do the work. I ended up doing this: https://github.com/HabitRPG/habitrpg/blob/develop/website/views/shared/new-stuff.jade#L20-L22

zedutch:
@crookedneighbor Thanks! Express is the only library here I haven't really used before and I'm still finding my way with it :)
My user ID: b8c11a9c-2716-4e12-bab5-0be536896496

zedutch:
@crookedneighbor  Also, what's the rule for when the dev branch gets merged into master? I've noticed the last merge was around march this year, which is quite long for an update... Or does the public site get updated from the dev branch/some other way?

crookedneighbor:
For this repo, `develop` _is_ the main branch.

One thing I'd recommend is to branch off, rather than committing directly to your fork of the `develop` branch. That way you can work on multiple things at once, and if a PR doesn't get merged in, you don't have to mess with resetting your develop branch to stay in sync.

---

PR #5888, opened 2015-09-04
Groups delete when last member leaves

TheHollidayInn:
This should fix the first two points on #5871 . I started this yesterday before we talked about rebasing on the new branch, so I will rebase this. I just wanted to get the code up, so you can review it.

@paglias and @crookedneighbor


paglias:
The first point has already been fixed, but tests are missing so we can use yours, the other one instead has yet to be fixed

crookedneighbor:
There's one API test that is failing.

TheHollidayInn:
Okay, I rebased and fixed the test on my local machine. Travis CI has a timeout issue. I'm not sure what that means.

crookedneighbor:
Travis was having trouble a little while ago. I'll give it a kick and see if it's resolved now.

paglias:
it'll have to be  up to date with branch https://github.com/HabitRPG/habitrpg/tree/paglias/firebase-requirements but it looks good. The part regarding groups getting deleted when last member leaves, though, has already been addressed so you'll have to remove it

TheHollidayInn:
Working on it now. I like the function abstraction from the controller.

paglias:
@TheHollidayInn Ok, we've merged that branch in develop so you can work once again with it

TheHollidayInn:
Ah, okay. I was going to create a PR for that branch. This one is up to date with the changes.

website/src/controllers/groups.js
```diff
@@ -517,6 +517,7 @@ api.leave = function(req, res, next) {
     user = group = keep = null;
     return res.send(204);
   });
+
```
crookedneighbor: Extra line break here.

TheHollidayInn:
Removed!

website/src/models/group.js
```diff
@@ -404,11 +404,29 @@ GroupSchema.methods.leave = function(user, keep, mainCb){
           group.type === 'party' ||
           (group.type === 'guild' && group.privacy === 'private')
       )){
-        // TODO remove invitations to this group
-        groupWasRemoved = true;
-        Group.remove({
-          _id: group._id
-        }, cb);
+        async.waterfall([
```
crookedneighbor: This should probably be handled in a pre or post hook for remove: http://mongoosejs.com/docs/middleware.html

paglias: I think in this case we can do it here ad it's the only place ( so far ) where a group is deleted. But for future changes maybe it's better a mideware

TheHollidayInn:
Okay, I made the change. I had to rewrite the remove call. Apparently, you are supposed to call findOne and then remove: https://github.com/Automattic/mongoose/issues/439 .

website/src/models/group.js
```diff
@@ -84,6 +84,27 @@ GroupSchema.pre('save', function(next){
   next();
 })
 
+GroupSchema.pre('remove', function(next) {
+  var group = this;
+  async.waterfall([
+    function(cb2) {
```
crookedneighbor: the variable `cb` is no longer overloaded, so we can use that instead of `cb2`

website/src/models/group.js
```diff
@@ -404,11 +425,12 @@ GroupSchema.methods.leave = function(user, keep, mainCb){
           group.type === 'party' ||
           (group.type === 'guild' && group.privacy === 'private')
       )){
-        // TODO remove invitations to this group
         groupWasRemoved = true;
-        Group.remove({
+        Group.findOne({
```
paglias: I don't think we need to `find` the document again, we can just to `group.remove()` as it's already an instance of Group

TheHollidayInn:
Updated!

website/src/models/group.js
```diff
@@ -404,11 +425,8 @@ GroupSchema.methods.leave = function(user, keep, mainCb){
           group.type === 'party' ||
           (group.type === 'guild' && group.privacy === 'private')
       )){
-        // TODO remove invitations to this group
         groupWasRemoved = true;
```
crookedneighbor: I think this all looks much nicer. 

If you want extra special bonus points, you could move the firebase removal code into a post hook. Then we wouldn't have to use this `groupWasRemoved` variable.

TheHollidayInn:
I'm down with bonus points. I just made the update :)

crookedneighbor:
@paglias this looks good to me. What do you think?

paglias:
It looks good! I'm merging and adding an index for `invitations.guilds.id` on users' collection

website/src/models/group.js
```diff
@@ -84,6 +84,31 @@ GroupSchema.pre('save', function(next){
   next();
 })
 
+GroupSchema.pre('remove', function(next) {
```
paglias, line 87: the only thing @TheHollidayInn is that this maybe should be in a post middleware rather than a pre... I'll fix it in develop

paglias, line 87: I think I'll leave this as it's, because we have no clear way of handling errors in a post middleware

crookedneighbor, line 87: I was going to mention the same thing.

crookedneighbor:
Notated toward your next tier. Thanks for working on this!

PR #5640, opened 2015-07-18
Adjust Cron when Custom Day Start is Changed Fixes #4488

carolstone:
The problem was that cron could be tricked into running over and over again by pushing the offset of dayStart forward an hour at a time. This solution will only allow cron to be run twice in quick succession if the user changes dayStart backward (the extreme case is allowing cron to run with a dayStart of 23 and then changing the dayStart offset to 0). But this behavior is much more logical than having cron run again if a player allows cron to run and then sets the offset to one hour in the future.

Modified common/script/index.coffee to display dayStart, but send back a variable of dayStart for processing, instead of having the object directly updated at the time of edit.

Modified website/views/options/settings.jade to allow a timestamp to be
cast to an iso timestamp because lastCron is saved as an iso timestamp,
and as a straight moment()-based timestamp in order to compare timestamps.

Modified website/public/js/controllers/settingsCtrl.js to do the
following:
   Accept the new proposed dayStart
   Calculate the timestamp of the old dayStart and the new dayStart

   These two steps were necessary because dayStart is an integer between
   and including 0 and 23. I tested and confirmed the current production
   dayStart does not allow 24 to be entered but does allow 0. I was
   careful to NOT change how that worked.

   Cast the old dayStart, the new dayStart and lastCron to the moment()
   version of time so they could be compared.

   Cast the new dayStart to the iso timestamp for storage in lastCron

   The important change is the following:

   if oldDayStart < lastCron AND lastCron < newDayStart then lastCron
   should be set to newDayStart. I modified this to include if oldDayStart
   = lastCron, although I think that's pretty unlikely to be possible.
   When I tested this, my lastCron was a mere 17 seconds after my
   oldDayStart, so it seemed to me that the equal case should be included
   with the less than case.

PR #5890, opened 2015-09-05
Change custom day start to ask for confirmation and prevent immediate cron

crookedneighbor:
Includes commits from #5640
## Problem

Adjusting the custom day start could cause cron to run immediately.
## Solution

Ask user to confirm selection and display the next time cron will run. 

Under the hood, this sets the user's last cron to that moment, so it shouldn't trigger cron to run immediately.

This also solves the bug where a user crons (say at midnight) and then is able to set their cds to 1am and cron again at 1.


crookedneighbor:
The text in the confirmation modal will need to be adjusted and the warning text (and wiki articles) will need to be changed to show the new behavior.

Also, maybe it would be more appropriate to include the warning text in the confirmation modal?

@alys ^

crookedneighbor:
screen shot 2015-09-05 at 3 45 22 pm

crookedneighbor:
Also, why is the selection not a drop down instead of an input box :frowning: ?

website/views/shared/modals/settings/change-day-start.jade
```diff
@@ -0,0 +1,12 @@
+.text-center
+  .modal-header
+    h3=env.t('changeCustomDayStart')
+
+  .modal-body
+    .alert.alert-danger
+      p=env.t('nextCron', {time: '{{nextCron | date: "M/d/yy \'@\' h:mm a"}}'})
```
Alys: This should use the user's preferred date format.

crookedneighbor: Good call.

common/locales/en/settings.json
```diff
@@ -39,6 +39,9 @@
     "xml": "(XML)",
     "json": "(JSON)",
     "customDayStart": "Custom Day Start",
+    "changeCustomDayStart": "Change Custom Day Start?",
+    "sureChangeCustomDayStart": "Are you sure you want to change your custom day start?",
+    "nextCron": "WARNING: Your dailies will reset next on <%= time %>.",
```
Alys: Cron doesn't run exactly at the CDS (unless you're using Habitica at that time), so the wording should reflect this. Maybe something like "Your Dailies will next reset the first time you use Habitica after <%= time %>."

lemoness:
I think the text of the confirmation modal is fine.

However, the warning text definitely needs to be changed. Are there any parts of the following paragraphs that still apply, or can we simply remove them?

"Complete all your Dailies before changing the Custom Day Start or Rest in the Inn that day. Changing your Custom Day Start may cause Cron to run immediately, but after the first day it works as expected.

Allow a window of two hours for the change to take effect. For example, if it is currently set to 0 (midnight), change it before 10pm; if you want to set it to 9pm, change it before 7pm."

Alys:
I'm not certain the red warning formatting is needed. With the display of the next time that cron will run and the ability to back out, this is now a safe operation. It's a big improvement over the current situation, but the warning formatting makes it look more scary. :)

Regarding the current warning text, which is:

```
Complete all your Dailies before changing the Custom Day Start 
or Rest in the Inn that day. Changing your Custom Day Start may
cause Cron to run immediately, but after the first day it works as expected.

Allow a window of two hours for the change to take effect. For example,
if it is currently set to 0 (midnight), change it before 10pm; if you want to
set it to 9pm, change it before 7pm.
```

I think all of that could be deleted now - is that right Blade?

It's probably worth including in the confirmation screen text like, "Make sure you have completed your Dailies before this time!"

Alys:
Oh, sorry Lemoness, I hadn't reloaded before I typed that.

crookedneighbor:
How's this?

screen shot 2015-09-06 at 10 15 31 am

PR #5900, opened 2015-09-09
Adjustments to quest routes

crookedneighbor:
Prevents quest leader's from leaving a party if they have started a quest.

Allows non quest leaders to leave a party if the quest is not yet active.

I also have some code to allow group leaders to cancel, abort or force begin a quest, but want some feedback from @alys and @lemoness before committing it.


crookedneighbor:
This will be one as part of v3 of the API.

Issue #5442, opened 2015-06-21
"Please choose the Mystery Item Set you would like." Where?


Cassolotl:
I'm in the Time Traveller section, and it says "please choose the Mystery Item Set you would like" but there is no obvious place to choose and no explanation.

![screen shot 2015-06-21 at 09 40 38](https://cloud.githubusercontent.com/assets/6607512/8271043/1760bd7c-17fa-11e5-810f-614f91d6093a.png)

If it's because I've got them all (and I don't think I have), there should be some "you've got them all" message, right? And it says "You can see a list of the past item sets [here](http://habitrpg.wikia.com/wiki/Mystery_Item)!" which links to the wiki page showing all the past items, but that's not a way to select an item you would like.

Alys:
You're right that the wiki won't let you choose the items but it will let you see them all, and you can compare them with the "Mystery" section in you Equipment page. However it is most likely that you do have them. I've just tested the Time Travellers page and it's working as expected for me.

Alys:
@lemoness Would you like a message to be added to that page for the case when the user already owns all Mystery Items?

Cassolotl:
@Alys @lemoness 

> Would you like a message to be added to that page for the case when the user already owns all Mystery Items?

That would be brilliant, and would have resolved my confusion immediately. :)

lemoness:
Yes, good idea!

"Congratulations! You already own all of the Mystery Items. Thanks for
supporting the site!"

Alys:
I've advertised this in Aspiring Coders.

melarish:
Working on this right now. So far I've managed to add a Debug menu function to add an Hourglass to my test user, so I can actually get to the Mystics screen. Will post again if I have more updates/issues.

melarish:
Does this look okay? 

![image](https://cloud.githubusercontent.com/assets/5783115/9741274/70ddc8b4-5651-11e5-85d3-0d5013a65767.png)

Alys:
@melarish I think it would look nicer if there was a bit more whitespace above the new message - perhaps about 1em or thereabouts? But that's basically what we're after, so feel free to do a PR!

crookedneighbor:
Would it make more sense to be in an alert box?

Alys:
Alert boxes need to be dismissed. I'm not sure the extra step is necessary here.

crookedneighbor:
Sorry, I meant the bootstrap alert class.

melarish:
[...]

PR #5904, opened 2015-09-09
"Please choose the Mystery Item Set you would like." Where? #5442

melarish:
Added a message that shows when user has all mystery items, instead of just showing an empty shop.


package.json
```diff
@@ -9,15 +9,23 @@
     "async": "~0.9.0",
     "aws-sdk": "^2.0.25",
     "babel": "^5.5.4",
+    "bl": "~1.0.0",
```
crookedneighbor, line 12: What are all these changes for?

paglias:
@melarish your repo is quite out of date, 258 commits behind our version, you should update it and then push your changes again

package.json
melarish, line 12: Modules that were missing when I was setting up my local copy. I installed them one by one but thought others may find it helpful if the package file gets updated, so they get installed automatically.
Nothing to do with this particular issue though, so up to you whether to ignore them :)

melarish:
@paglias My bad, I was doing git fetch instead of git pull. Merged it now :)

package.json
crookedneighbor, line 12: Many of those are likely dependencies of our dependencies. IE, a module we require depends on another module. And since the module we require failed to install, it's dependencies also failed.

I will clean this up when I review your PR.

crookedneighbor:
@melarish The reason your add hourglass changes didn't stick was because you were using `User.set`, but hourglasses, like gems, can't be set that way. I created a new route (only available in a dev environment) to add an hourglass. See https://github.com/HabitRPG/habitrpg/commit/e554309822f9e5f9dff2a0ea12583b7d12deaeb1 and https://github.com/HabitRPG/habitrpg/commit/b6f02d388f353e82aeb548f63cd33fcb8bf9748f

melarish:
Oic; maybe that's also why the items I bought weren't sticking. Thanks for fixing that.

User ID: d97516e4-81d0-4f60-bf03-95f7330925ab

Alys:
@melarish I've awarded you a contributor tier for this. I'm very sorry about the delay!

crookedneighbor:
Whoops, sorry, I dropped the ball on that :grimacing:

PR #5912, opened 2015-09-11
Removed user's party invites when last user leaves party

TheHollidayInn:
This PR removes party invites when the last member of a party leaves. This fixes part of #5871

### DIFF
--- a/website/src/models/group.js
+++ b/website/src/models/group.js
@@ -87,46 +87,30 @@ GroupSchema.pre('save', function(next){
 
 GroupSchema.pre('remove', function(next) {
   var group = this;
-  if ( group.type == "guild" ) {
-    async.waterfall([
-      function(cb) {
-        User.find({
-          'invitations.guilds.id': group._id
-        }, cb);
-      },
-      function(users, cb) {
-        if (users) {
-          users.forEach(function (user, index, array) {
+  async.waterfall([
+    function(cb) {
+      var invitationQuery = {};
+      var groupType = group.type;
+      //Add an 's' to group type guild because the model has the plural version
+      if (group.type == "guild") groupType += "s";
+      invitationQuery['invitations.' + groupType + '.id'] = group._id;
+      User.find(invitationQuery, cb);
+    },
+    function(users, cb) {
+      if (users) {
+        users.forEach(function (user, index, array) {
+          if ( group.type == "party" ) {
+            user.invitations.party = {};
+          } else {
             var i = _.findIndex(user.invitations.guilds, {id: group._id});
             user.invitations.guilds.splice(i, 1);
-            user.save();
-          });
-        }
-        cb();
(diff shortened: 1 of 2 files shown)


website/src/models/group.js
```diff
@@ -87,23 +87,46 @@ GroupSchema.pre('save', function(next){
 
 GroupSchema.pre('remove', function(next) {
   var group = this;
-  async.waterfall([
-    function(cb) {
-      User.find({
-        'invitations.guilds.id': group._id
```
crookedneighbor, line 93: Could you do something like:

``` javascript
var invitationQuery = {};
invitationQuery['invitations.' + group.type + '.id'] = group._id;

// do stuff

User.find(invitationQuery, cb);

// do stuff
users.forEach(function(user, index, array) {

  if (group.type === 'party') { 
      user.invitations.party = {};
  } else {
    var i = _.findIndex(user.invitations.guilds, {id: group._id});
    user.invitations.guilds.splice(i, 1);
  }

  user.save();
});

// do stuff
```

website/src/models/group.js
```diff
@@ -87,23 +87,46 @@ GroupSchema.pre('save', function(next){
...
+    async.waterfall([
+      function(cb) {
+        User.find({
+          'invitations.party.id': group._id
+        }, cb);
+      },
+      function(users, cb) {
+        if (users) {
+          users.forEach(function (user, index, array) {
+            if (user.invitations.party.id === group._id) {
```
crookedneighbor: This shouldn't be necessary, right? Since the query is specifically for the party invitations?

website/src/models/group.js
TheHollidayInn, line 93: I can try, but I believe the invitations are structured differently. The party invitations weren't accepting the same type of array queries as the guilds were.

website/src/models/group.js
TheHollidayInn: Exactly. I was just mixed up by the structure, I think. Pushing updates soon.

TheHollidayInn:
Updated! The structure is only different when deleting, so that worked. (That's what I mixed up before). Also, I had to append an 's' because of the the structure for guilds is `guilds.invitations` rather than `guild.invitations` where guild is `group.type`.

crookedneighbor:
:+1:

Notated.

gisikw:
Sorry, but I had to revert this commit due to performance issues with a heavily-populated Users collection. Please add a migration to add an index for `users.invitations.party.id` at your convenience, and we can pull this back in :)

Cheers!

gisikw:
Ignore me for the moment - we don't currently have a system for managing indexes, so this will be put back in, and we'll set up the index manually.

TheHollidayInn:
Sounds good. Let me know if there is anything I can do to help!

PR #5916, opened 2015-09-11
Added test to check that group invites are removed when user cancels account

TheHollidayInn:
Part of #5871 is "user should be removed from all the groups he's invited to" . I believe this is already happening unless I am misunderstanding the goal. I added tests to show that when a user leaves, their invites are removed form the groups invite list.


crookedneighbor:
Notated.

Issue #5930, opened 2015-09-15
Quest leader Inactive, Can't start Quest


P33bs:
The quest starting member was inactive for a very long time so they were booted from the party. No new quest leader was assigned and now we cannot start a new quest or cancel the current one. Group ID is: b58e0cb1-c6e1-4936-8519-aba1dbdb0fc1

crookedneighbor:
I've reset the stalled quest.

katieshmatie:
I have a similar problem. Our quest leader has gone inactive, and all party members except one have accepted. The remaining member is inactive as well, so we're stuck waiting for one of them to log back in (I don't want to boot them from the party). I'm trying to find the correct forum to request an admin to cancel the quest invite so an active member can start up a new quest.

Group ID is 05e7137d-b3fd-4c69-8106-54983903f2c1

Alys:
@katieshmatie This is the correct place to post, as we advised you in the Newbies guild. I've deleted the quest from your party.

PR #5934, opened 2015-09-15
Adjustments to group quest behavior

crookedneighbor:
This is a modified version of: #5900

Prevents quest leaders from leaving party when there is a quest (whether active or inactive).

Allows other users to leave party when quest is going on, whether active or inactive.

Includes the client side changes from #5900. API changes will need to wait until the iOS app is updated to handle the new behavior.


crookedneighbor:
Added some logic to remove quest data from group if user is booted from party after starting a quest.

Should eliminate issues like these: #5930

crookedneighbor:
It also returns quest scroll to booted user if the quest is active.

PR #5935, opened 2015-09-15
Ensure NPM scripts always use local binary

### LINKED ISSUE #5933: Use $(npm bin) in npm scripts
Hello,

At the moment, `package.json`refers to "best-used-globally" module binaries in its npm scripts.
Their presence in the user's PATH is assumed, though `gulp` and `bower` are explicitely stated as dependencies in the package.json.
This causes the app's build cycle to break (especially `npm install` because of the postinstall script)
on all environments that do not match this requirement.
Most of the time, even though it's kinda annoying anyway, the user can just run `npm install -g bower` and fix the issue.
Unfortunately it's not the case 100% of the time, so I guess it'd be better to simply refer to "preferGlobally" binaries using `$(npm bin)` inside npm scripts.
## 

Tim

ziir:
This probably requires an update of the documentation.
Globally installing modules such as gulp or bower is no longer necessary for npm scripts to work.

Fixes #5933

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,11 +79,11 @@
     "node": "0.10.x"
   },
   "scripts": {
-    "test": "gulp test",
-    "start": "gulp run:dev",
-    "sprites": "gulp sprites:compile",
-    "postinstall": "bower --config.interactive=false install -f; gulp build;",
-    "coverage": "COVERAGE=true mocha --require register-handlers.js --reporter html-cov > coverage.html; open coverage.html"
+    "test": "$(npm bin)/gulp test",
+    "start": "$(npm bin)/gulp run:dev",
+    "sprites": "$(npm bin)/gulp sprites:compile",
+    "postinstall": "$(npm bin)/bower --config.interactive=false install -f; $(npm bin)/gulp build;",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -31,7 +31,6 @@
     "grunt-contrib-watch": "~0.6.1",
     "grunt-hashres": "~0.4.1",
     "grunt-karma": "~0.6.2",
-    "gulp": "^3.9.0",
```
crookedneighbor, line 34: Doesn't gulp still need to be installed locally?

ziir, line 34: Yes it does, sorry about that

crookedneighbor:
Beautiful. I recently made the change from using `./node_modules/.bin/<whatever>` to just `<whatever>` forgetting that not all systems default to using the npm bin path. Using `$(npm bin)` is much nicer.

crookedneighbor:

ziir:
Thank you for such a swift response!
Yup, I just signed up :) My user id is c2fcde39-8331-4ac1-becd-f3ebe80befc2

crookedneighbor:

PR #5938, opened 2015-09-16
Time Travel Stable

SabreCat:
Allows purchase of pets and mounts previously awarded only at the conclusion of World Boss events, via use of Mystic Hourglasses given to long-time subscribers.


test/spec/controllers/inventoryCtrlSpec.js
```diff
@@ -224,14 +224,56 @@ describe('Inventory Controller', function() {
...
   describe('#hasAllTimeTravelerItems', function() {
-    it('returns false if there are items left in the time traveler store', function() {
+    it('returns false if items remain for purchase with Mystic Hourglasses', function() {
       expect(scope.hasAllTimeTravelerItems()).to.eql(false);
     });
 
     it('returns true if there are no items left to purchase', inject(function(Content) {
       sandbox.stub(Content, 'timeTravelerStore').returns({});
-
+      for (var pet in Content.timeTravelStable.pets) {
```
crookedneighbor, line 237: You can just use the native js `foreach`, right? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

test/spec/controllers/inventoryCtrlSpec.js
```diff
@@ -224,14 +224,56 @@ describe('Inventory Controller', function() {
   });
 
   describe('#hasAllTimeTravelerItems', function() {
-    it('returns false if there are items left in the time traveler store', function() {
+    it('returns false if items remain for purchase with Mystic Hourglasses', function() {
       expect(scope.hasAllTimeTravelerItems()).to.eql(false);
     });
 
     it('returns true if there are no items left to purchase', inject(function(Content) {
       sandbox.stub(Content, 'timeTravelerStore').returns({});
```
crookedneighbor: Probably better to do what you're doing with the mounts and pets.

test/spec/controllers/inventoryCtrlSpec.js
SabreCat, line 237: Who knew that was a thing? Sure. xD

test/spec/controllers/inventoryCtrlSpec.js
SabreCat: Ah, iterate over all the mystery item sets and give them to the user? OK, I'll see if I can figure that out.

test/spec/controllers/inventoryCtrlSpec.js
SabreCat, line 237: Since I ended up turning Content.timetravelStable[whatever] into objects, `forEach` doesn't work.

test/spec/controllers/inventoryCtrlSpec.js
SabreCat: Fixed. Though my means of figuring out which items to give the `user` are a little kludgey--suggestions welcome!

PR #5943, opened 2015-09-17
Revert "Removed user's party invites when last user leaves party"

gisikw:
Reverts HabitRPG/habitrpg#5912

Unfortunately, this has the potential to cause very severe performance problems when a party is removed. While we have an index specified in the database for `users.invites.guilds.id`, there's no such index for `users.invites.party.id`. I believe this was causing significant performance issues.

Reverted pending the addition of a migration to the original PR which adds the necessary index.

### DIFF
--- a/website/src/models/group.js
+++ b/website/src/models/group.js
@@ -89,22 +89,15 @@ GroupSchema.pre('remove', function(next) {
   var group = this;
   async.waterfall([
     function(cb) {
-      var invitationQuery = {};
-      var groupType = group.type;
-      //Add an 's' to group type guild because the model has the plural version
-      if (group.type == "guild") groupType += "s";
-      invitationQuery['invitations.' + groupType + '.id'] = group._id;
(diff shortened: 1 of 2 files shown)


website/src/models/group.js
```diff
@@ -89,22 +89,15 @@ GroupSchema.pre('remove', function(next) {
   var group = this;
   async.waterfall([
     function(cb) {
-      var invitationQuery = {};
-      var groupType = group.type;
-      //Add an 's' to group type guild because the model has the plural version
-      if (group.type == "guild") groupType += "s";
-      invitationQuery['invitations.' + groupType + '.id'] = group._id;
```
crookedneighbor, line 96: So, do you think that the `invitations.party.id` query was causing the problem, or something else?

gisikw, line 96: I ran a test query against invitations.party.id, and it took >60s to run. However, we still experienced timeouts after reverting this PR, so there's still more to be found.

PR #5954, opened 2015-09-18
Fall Festival 2015

SabreCat:
Event opener for autumn 2015! NPC sprites, limited edition class gear sets, Seasonal Shop, and candy food.


test/common/algos.mocha.coffee
```diff
@@ -518,12 +518,12 @@ describe 'User', ->
       expect(user.items.hatchingPotions).to.eql {}
       expect(user.items.food).to.eql {}
 
-    it 'gets honey', ->
+    it 'gets red candy', ->
```
crookedneighbor: The fact that we have to change this test for an event is a code smell.

SabreCat: Suggestions? It's another of those goofy things where we're testing a random output.

crookedneighbor: Well, what is the purpose of this test? It's not necessarily that the drop system gives honey, or that it gives red candy, right? Seems to me that the purpose of this test is that the drop system, from some range or predictable random to another range gives food. So it might be better written as (pseudocode here):

```
var MIN_RANGE_FOR_FOOD = .7; // not sure what the exact value is
var MAX_RANGE_FOR_FOOD = .9; // not sure what the exact value is

for( MIN_RANGE_FOR_FOOD to MAX_RANGE_FOR_FOOD) {
  it ('drops food', function() {
     sinon.stub(user.fns, 'predictableRandom', cycle [0, 0, SOME_VALUE]);
     user.ops.score {params: { id: @task_id, direction: 'up'}};

    expect(user.items.eggs).to.be.empty;
    expect(user.items.hatchingPotions).to.be.empty;
    expect(user.items.food).to.not.be.empty;
  });
}
```

I would need to verify that predictable random works that way. Haven't really examined the user ops stuff

SabreCat: No, it sounds like you've got the concept right! I should be able to do something like the above.

Issue #5945, opened 2015-09-17
link to Newbies Guild / Ask a Question in the Tavern's Resources box


Alys:
As discussed in the Aspiring Socialites guild, the Resources box in the Tavern should have a link to the Newbies Guild. We probably want the link to include the text "Ask a Question" so it's obvious what it's for. 

I suggest the link's text should be "Ask a Question (Newbies Guild)" but look below to see if there's a comment from @lemoness suggesting something better. :)

I'm advertising this in the Aspiring Coders guild as an entry-level project.

LuneFox:
I did some research, and noticed that  `habitrpg/website/views/options/social/tavern.jade` contains the table itself, and  `habitrpg/test/spec/translations.js` contains default names of the strings in English (why "test", though?). If these two files are the only ones that are related to solving this task, I can try adding the field.

crookedneighbor:
@LuneFox `test/spec/translations.js` is a fixture that is compiled when a test is run so that the test suite has access to the translation strings.

IE: if my test is `expect(foo).to.eql('some string we change alot');`, the test will break everytime we change the string. However, if we use the translation string variable:

`bar: 'some string we change alot'`

`expect(foo).to.eql(window.env.t('bar'))` - no matter what we change it to, the test assertion remains valid.

In fact, you cannot commit changes to the translations fixture because it is ignored by git.

lemoness:
As a title, "Ask a Question (Newbies Guild)" works well!

LuneFox:
Ah, it makes sense. I also used to keep all "full" strings in a separate file and only referred to variables in main program when I was coding for myself. So the only thing that has to be done is just adding the extra field to tavern.jade using some new short name for the string? I'll try.

LuneFox:
Okay... here we go. https://github.com/HabitRPG/habitrpg/pull/5955

SabreCat:
@crookedneighbor: Ooh, I didn't know that about testing string outputs. That would have made things easier for me on some of my recent Hourglass stuff. `>.<`

PR #5955, opened 2015-09-18
Added link to Newbies Guild in the Tavern

LuneFox:
Sigh. I hope I did this right.

### DIFF
--- a/common/locales/en/groups.json
+++ b/common/locales/en/groups.json
@@ -12,6 +12,7 @@
     "community": "Community Forum",
     "dataTool": "Data Display Tool",
     "resources": "Resources",
+    "askQuestion": "Ask a Question (Newbies Guild)",
     "tavernTalk": "Tavern Talk",
     "tavernAlert1": "Note: if you're reporting a bug, the developers won't see it here. Please",
     "tavernAlert2": "use GitHub instead",


Alys:
Hi @LuneFox 

Any text on the site that appears in `=env.t('...')` is configured in the files under `common/locales/en/`. Those files are copied to Transifex for translators to create translated versions. See the README.md file in `common/locales/en/` for more information. Could you please edit one of those files to add the new string? Use the same file that contains strings for other items in the Resources box. If you commit the change to the same branch as your other changes, and then push to `origin`, it will appear in this pull request.

The `test/spec/translations.js` file should not be edited manually because it's compiled automatically when a test is run. It shouldn't have even been possible for you to commit it, but we need to adjust the `.gitignore` file to handle subdirectories under `test`.

Also, when you're creating a pull request, if you put "fixes #5945" (i.e., the issue number) in the title or description, github creates an automatic link to the issue and closes the issue if the pull request is merged.

LuneFox:
@Alys Ah, I see. I got it wrong at first. Okay, let me see what I can do. (Both GitHub and this project are new to me, so mistakes were expected.)

LuneFox:
I guess editing the file that is created during the test won't hurt it, for it will be created anew every time. But is there any way to discard a specific change from the pull, by the way?

crookedneighbor:
@LuneFox Don't worry about it. I will cherry pick all but the commit that changed the fixture.

LuneFox:
@crookedneighbor Oh, okay. To be honest I just wanted to check if the files in the test folder really can't be commited like you said. I expected some kind of an error, but suddenly it passed.

crookedneighbor:
Oh, actually, the file I was thinking of was: `/test/spec/mocks/translations.js`

The one in `/test/spec/translations.js` is a mistake and should be deleted. Will do that when I merge in your PR.

crookedneighbor:

LuneFox:
@crookedneighbor Thanks :) My ID is 1f27893f-3808-4724-9725-f46dab93faca

crookedneighbor:

LuneFox:
@crookedneighbor Thanks again :) I'll keep that in mind.

PR #5956, opened 2015-09-19
Added lastCron condition to senior member and tests

TheHollidayInn:
This is for #5871 - Most senior member should be made leader . The code to make a member the leader when the current leader leaves was already there. I modified it to add lastCron check to make sure the user is active. However, we might need to add other parameters to ensure the member is considered "Senior".

Also, I am having an issued when inviting and joining the two users. I'm posting this so you can start reviewing, and if you know what the issue is. Otherwise, I will debug this tomorrow.

### DIFF
--- a/website/src/controllers/groups.js
+++ b/website/src/controllers/groups.js
@@ -518,12 +518,10 @@ api.leave = function(req, res, next) {
   // When removing the user from challenges, should we keep the tasks?
   var keep = (/^remove-all/i).test(req.query.keep) ? 'remove-all' : 'keep-all';
 
-  Group.findOne({"_id": group._id}).populate('members').exec(function (err, group) {
-    group.leave(user, keep, function(err){
-      if (err) return next(err);
-      user = group = keep = null;
-      return res.send(204);
-    });
+  group.leave(user, keep, function(err){
+    if (err) return next(err);
+    user = group = keep = null;
+    return res.send(204);
   });
 };
 
--- a/website/src/controllers/groups.js
+++ b/website/src/controllers/groups.js
@@ -473,12 +473,12 @@ api.join = function(req, res, next) {
   if(!isUserInvited) return res.json(401, {err: "Can't join a group you're not invited to."});
 
   if (!_.contains(group.members, user._id)){
+
     if (group.members.length === 0) {
       group.leader = user._id;
     }
 
     group.members.push(user._id);
-
     if (group.invites.length > 0) {
      group.invites.splice(_.indexOf(group.invites, user._id), 1);
     }
@@ -518,10 +518,12 @@ api.leave = function(req, res, next) {
   // When removing the user from challenges, should we keep the tasks?
   var keep = (/^remove-all/i).test(req.query.keep) ? 'remove-all' : 'keep-all';
 
-  group.leave(user, keep, function(err){
-    if (err) return next(err);
-    user = group = keep = null;
-    return res.send(204);
+  Group.findOne({"_id": group._id}).populate('members').exec(function (err, group) {
+    group.leave(user, keep, function(err){
+      if (err) return next(err);
+      user = group = keep = null;
+      return res.send(204);
+    });
   });
 };
 
--- a/website/src/models/group.js
+++ b/website/src/models/group.js
@@ -444,13 +444,21 @@ GroupSchema.methods.leave = function(user, keep, mainCb){
         var leader = group.leader;
 
         if(leader == user._id || !~group.members.indexOf(leader)){
-          var seniorMember = _.find(group.members, function (m) {return m != user._id});
 
+          var seniorMember = undefined;
+          _.forEach(group.members, function (m) {
+            if (!seniorMember) {
+              seniorMember = m;
+            } else if (m._id != user._id && m.lastCron > seniorMember.lastCron) {
+              seniorMember = m;
+            }
+            return
+          });
           // could not exist in case of public guild with 1 member who is leaving
           if(seniorMember){
             if (leader == user._id || !~group.members.indexOf(leader)) {
               update['$set'] = update['$set'] || {};
-              update['$set'].leader = seniorMember;
+              update['$set'].leader = seniorMember._id;
             }
           }
         }
--- a/website/src/models/group.js
+++ b/website/src/models/group.js
@@ -449,7 +449,7 @@ GroupSchema.methods.leave = function(user, keep, mainCb){
           _.forEach(group.members, function (m) {
             if (!seniorMember) {
               seniorMember = m;
-            } else if (m._id != user._id && m.lastCron > seniorMember.lastCron) {
+            } else if (m._id != user._id && m.lastCron > seniorMember.lastCron && flags.chatRevoked === false && flags.blocked === false) {
               seniorMember = m;
             }
             return
--- a/website/src/models/group.js
+++ b/website/src/models/group.js
@@ -446,25 +446,43 @@ GroupSchema.methods.leave = function(user, keep, mainCb){
         if(leader == user._id || !~group.members.indexOf(leader)){
 
           var seniorMember = undefined;
-          _.forEach(group.members, function (m) {
-            if (!seniorMember) {
-              seniorMember = m;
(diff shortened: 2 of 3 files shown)


website/src/controllers/groups.js
```diff
@@ -473,14 +473,15 @@ api.join = function(req, res, next) {
...
+
     if (group.members.length === 0) {
       group.leader = user._id;
     }
 
     group.members.push(user._id);
-
     if (group.invites.length > 0) {
      group.invites.splice(_.indexOf(group.invites, user._id), 1);
+     console.log(group.invites);
```
TheHollidayInn: This is the section giving me 'error: VersionError: No matching document found.' .

TheHollidayInn:
Alright, turns out that the async.map function was allowing enough time for the invites to update when using /join, so I used mapSeries: http://promise-nuggets.github.io/articles/15-map-in-series.html .

website/src/controllers/groups.js
```diff
@@ -518,10 +518,12 @@ api.leave = function(req, res, next) {
   // When removing the user from challenges, should we keep the tasks?
   var keep = (/^remove-all/i).test(req.query.keep) ? 'remove-all' : 'keep-all';
 
-  group.leave(user, keep, function(err){
-    if (err) return next(err);
-    user = group = keep = null;
-    return res.send(204);
+  Group.findOne({"_id": group._id}).populate('members').exec(function (err, group) {
```
paglias: @TheHollidayInn are we sure this has to be fetched once again? We already have a group object from the db, and ids don't need to be populated

TheHollidayInn: @paglias I'm not completely sure, but I am pretty sure.

If there is a way to populate the users into the live group object, then we don't need to.

The reason I think we need the members populates is shown in my additions to selecting the senior member. Before, we simply selected the next id that was not equal to the leader's id. @Alys stated she wanted to make sure the user had logged in recently, so I need to populate the users to access the lastCron property. Also, I believe there should be more criteria in selecting the senior member, so we will probably need more properties from the members. 

We can use projects to lighten the query though, as we don't need every property.

Alys:
Other criteria would be `flags.chatRevoked` is false and `auth.blocked` is false. I.e., this would prevent a user being automatically made a guild leader if their account had been blocked or their chat privileges had been revoked. @lemoness FYI.

Note that this is only for automatic leadership changes. We'd want to discuss this further (probably elsewhere) before applying the same criterion to manual leadership changes.

TheHollidayInn:
Sweet. I added the new criteria as well.

Alys:
It's `auth.blocked` not `flags.blocked`.

Also, I'm sorry, I should have mentioned this: On an account that hasn't been blocked/revoked, those attributes don't exist at all (no `true` nor `false` value).  Does `=== false` include the case where the attribute doesn't exist?

paglias:
@TheHollidayInn now that I think about it, there's a problem with populating groups' members. We don't do it on the site because it's soo heavy for big groups with loads of members. It would crash the site if it happens on a big group. i''m sorry but for now this is out of question

TheHollidayInn:
@Alys , I believe it does not. I will need to change that to (!auth.blocked).

@paglias Make sense to me. I have another idea. I'll make changes tomorrow.

Alys:
What if we picked the first person in the list of members, fetched the data for just that person, and if it happened that they weren't suitable, repeated with the person(s) after that?

TheHollidayInn:
That's almost exactly what I was thinking.  :)

crookedneighbor:
This might be a bad idea, but what about:

``` js
var today = new Date();
var lastWeek = new Date(today.getFullYear(), today.getMonth(), today.getDate() - 7);

User.findOne({
  '_id': { $in : group.members }, 
  'auth.blocked' : { $exists: false }, 
  'flags.chatRevoked': { $exists: false },
  'lastCron': { $gt: lastWeek }
}, function(err, user) {
  if (err) return next(err);

  if (!user) { 
    // no applicable member is found, assign first member of group.members
  } else {
    // Assign matching user id to leader
  }

  // Save group
});
```

May also want to splice the group.members array so we're only looking at a subset of the group. 

Also, we'd probably need to put indexes on all those fields.

crookedneighbor:
@paglias Your thoughts on ^

TheHollidayInn:
@crookedneighbor That is similar to what I was thinking. Although, `'_id': { $in : group.members },` this will need to be an AND query with another check to make sure the _id is not the current leader. I'm testing the code now.

paglias:
@crookedneighbor that seems good, yeah we just need to remember to create a compound index on those fields

TheHollidayInn:
I made the updates and included some changes to the query to maintain consistency. Do I need to created the compound index?

paglias:
I'll create it when I get a chance unless someone else does it
Il 22/set/2015 01:38 PM, "Keith Holliday" notifications@github.com ha
scritto:

TheHollidayInn:
`db.groups.members.createIndex( { _id: 1, auth.blocked: 1, flags.chatRevoked: 1, lastCron: -1 } )`

Is this what you need? I haven't messed around with compound indexes yet.

paglias:
Yes but you can't create it, I have to do it on MongoLab and currently I'm
travelling so it's going to be 20 hours or so before I get home
Il 23/set/2015 10:25 AM, "Keith Holliday" notifications@github.com ha
scritto:

TheHollidayInn:
Gotcha. I figured it was something I didn't have access to. No rush and thanks!

paglias:
@TheHollidayInn the index has been created

@crookedneighbor if it look good to you, we can merge

crookedneighbor:
Did you intend to include the angular changes around the challenge filters in this PR?

TheHollidayInn:
Nope. My bad. I removed the commit.

website/src/models/group.js
```diff
@@ -444,19 +444,45 @@ GroupSchema.methods.leave = function(user, keep, mainCb){
...
+          var today = new Date();
+          var lastWeek = new Date(today.getFullYear(), today.getMonth(), today.getDate() - 7);
+
+          User.findOne({
+            '_id': { $in : group.members, $ne : leader },
+            'auth.blocked' : { $exists: false },
+            'flags.chatRevoked': { $exists: false },
+            'lastCron': { $gt: lastWeek }
+          })
+          .sort({'lastCron': -1})
```
crookedneighbor, line 458: Since we're using findOne, I'm not certain that this sort function is actually doing anything. Can you dig up some documentation figuring out if it is one way or the other?

paglias, line 458: According to this http://stackoverflow.com/questions/25058007/mongodb-difference-between-orderby-and-sort, `sort` should be avalaible only to `find()` and not `findOne()` but they also mention that it should produce an error, which in this case doesn't seem the case.

Anyway the documentation about `findOne()` says that the document is returned according to natural order http://docs.mongodb.org/manual/reference/method/db.collection.findOne/

The solutions seems to be:
1. Use `find()` with `limit(1)` and `sort()`
2. Use `findOne()` with `$orderBy`

paglias, line 458: The main reason is that `sort()` is applied to a cursor and while `find()` is indeed one, `findOne()` isn't

TheHollidayInn, line 458: Thanks for looking this up, @paglias . This is what I read as well. The code seem to be more stable after adding the sort, but the documentation says otherwise. I will add the $orderBy and push in a few minutes.

TheHollidayInn, line 458: Actually, we are using mongoose, not Mongodb directly, so when we use findOne it returns a query: http://mongoosejs.com/docs/api.html#query_Query-findOne . Then, we can call sort on that query: http://mongoosejs.com/docs/api.html#query_Query-sort . 

The answer in this post shows a few sorting methods on findOne: http://stackoverflow.com/questions/13443069/mongoose-request-order-by .

paglias, line 458: good catch!

crookedneighbor:
Blocked by the api test pr.

crookedneighbor:

crookedneighbor:
We're accelerating our work on V3 of the API, as such, we're not going to accept anymore V2 PRs. Note that this is a desired feature, so we'll want it to be a part of v3.

PR #5962, opened 2015-09-21
Magic Potions (**MERGE AFTER 9/30 DEPLOY - SEE COMMENTS**)

SabreCat:
Implements "magic" hatching potions, which are only available for purchase with Gems, and which can hatch brand new recolors of the core pet list.
## To do:
- [x] Can't buy limited-time magic potions outside of event
- [x] -- Generalize behavior of `canBuy` flag across app
- [x] -- Clearly mark magic potions as having limited time availability
- [x] Can't hatch magic potion versions of quest pets
- [x] -- Note in popover text that these potions cannot be used on quest eggs
- [x] -- Don't highlight quest pet eggs when magic potion is selected
- [x] Don't show placeholders for magic potion versions of pets/mounts if the potions are unavailable for purchase and user does not have any of those potions
- [x] Add sprites for new color potion, pets, and mounts
- [x] Pet/mount pages do not have placeholders for magic potion versions of quest pets/mounts
- [x] Premium pet/mount colors don't count for or against Beast/Mount Master achievements
- [x] Premium pets like all food types
- [x] Premium pets/mounts are not released when user activates Key to the Kennels


crookedneighbor:
Regarding the only purchase during event thing: what happens if you set canBuy to false instead of allowing it to default to true?

crookedneighbor:
Would putting the potion in seasonal shop make it clear that it was a limited time option?

SabreCat:
@crookedneighbor: Re `canBuy`: Nothing, yet, other than that it doesn't show up in the Market. Still working on that piece.

Seasonal Shop: that's not a bad idea! I'll check in with @lemoness, maybe that's the way to go.

common/script/index.coffee
```diff
@@ -822,7 +822,8 @@ api.wrap = (user, main=true) ->
           item = content[type][key]
           price = item.value / 4
         return cb?({code:404,message:":key not found for Content.#{type}"},req) unless item
-        return cb?({code:401, message: i18n.t('notEnoughGems', req.language)}) if (user.balance < price) or !user.balance
+        return cb?({code:403, message: i18n.t('messageNotAvailable', req.language)}) if type is 'hatchingPotions' and not item.canBuy
```
SabreCat: It might be better to generalize this so we can flag any buyable item with `canBuy:false` to turn it off, but at the moment we're not consistent in the meaning of `canBuy` across the app. I'll put some work into this next.

common/locales/en/messages.json
```diff
@@ -14,6 +15,7 @@
   "messageEquipped": " <%= itemText %> equipped.",
   "messageUnEquipped": "<%= itemText %> un-equipped.",
   "messageMissingEggPotion": "You're missing either that egg or that potion",
+  "messageInvalidEggPotionCombo": "Invalid egg+potion combination. Try hatching a different egg!",
```
crookedneighbor: Is this a little mechanical sounding?

SabreCat: Probably so. Maybe "You can't hatch that egg with this potion. Try a different egg!"?

crookedneighbor: I like that better. @lemoness ?

lemoness: Does that only show up with quest eggs, or also of you already have the
pet? If it's just quest eggs, we can make it more descriptive:

"You can't hatch Quest Pet Eggs with Magic Hatching Potions! Try a
different egg."  That way the user understands why it isn't working.

SabreCat: It'd only be quest eggs. The existing error message will still come up for trying to duplicate a pet combination. Thanks, I'll put that phrasing in!

SabreCat:
As of commit feff6fa, this is ready to roll out (albeit with the remaining `gulp compile:sprites` chore still to do). OK to merge only if no other production deploy is scheduled between now and the Magic Potions release date as listed on Trello.

SabreCat:
After a closer look at Trello, we do have a Bailey deploy planned for Wednesday. This should merge after that.

PR #5963, opened 2015-09-22
Tasks Collection [ON HOLD]

paglias:
**Work in progress.**
### API v1
- [x] deprecated
### Mongoose 4 Mandatory
- [x] upgrade version in package.json
- [ ] driver v2.x 1. If you are connecting to a replica set, you must specify the replicaSet option in the connection string. **add ?replicaSet=name** to production connection uri, can be found in admin panel under "To connect..."
- [x] driver v2.x 2. update returns a result object from the MongoDB server, rather than just the number affected. The second parameter to the callback will now look like { ok: 1, n: 3 } rather than simply the number affected.
- [x] autoinc not compatible
### Tasks Collection
- [x] startDate default should be a function? https://github.com/HabitRPG/habitrpg/commit/3e701e144d7ab4c08064b36ccde23c68b4920a9a#diff-a4823628c26fce6f17432a7c8d667e2bR54
- [x] ~~use model on browser or check shared's tasksDefaults~~ Not for now
- [x] ~~User's `history` array~~ going to keep it
- [x] ~~Tags in their own collection?~~ Not for now
- [x] ~~Inbox in its own collection?~~ Not for now
- [x] In migration remove the necessity of post('save') checks, especially for users data
- [x] shared.uuid use node-uuid
- [ ] except for user's ops, use express 4 syntax for responses
- [ ] use discriminators instead of having a single model fro habits/rewards/todos/dailys https://github.com/Automattic/mongoose/pull/1647
- [ ] use cron to fix possible errors in user.tasksOrder
- [ ] add indexes in foreground
- [ ] prevent challenge task to have userId and make task.challenge not editable
- [x] migrate challenges.csv method 
- [x] migrate challenges.getMember
### Node v4
- [x] Upgrade to Node 4
- [x] Enable New Relic
- [ ] newrelic add on for heroku
- [ ] newrelic subscription
- [ ] Upgrade deps
### Migration
- [ ] lower case only emails (code)
- [ ] lower case only emails (migration)
- [ ] fix task difficulty values (code)
- [ ] fix task difficulty value (migration)
- [ ] indexes
### NPM 3
- [ ] karma-chai-plugins@0.1.3 UNMET PEER DEPENDENCY chai@1.7.2 (avalaible in last version of karma-chai-plugins)
- [ ] karma-requirejs UNMET PEER DEPENDENCY requirejs@~2.1
- [ ] chai-as-promised@3.3.1 requires a peer of chai@>= 1.0.2 < 2 but none was installed.
### MongoDB 3
- [ ] http://docs.mongodb.org/manual/release-notes/3.0-upgrade/
- wait for wired tiger support on mongolab? DELAYED
### Misc
- [ ] internal card and preparations for api v3


test/common/dailies.coffee
```diff
@@ -49,9 +50,15 @@ newUser = (addTasks=true)->
...
 
+getTasksForUser = (user) ->
+  return user.habits.concat(user.dailys)
+          .concat(user.todos).concat(user.rewards)
+
 cron = (usr, missedDays=1) ->
   usr.lastCron = moment().subtract(missedDays,'days')
-  usr.fns.cron()
+  daysMissed = usr.fns.shouldCronRun()
+  if daysMissed isnt 0
```
crookedneighbor, line 60: This is the problem. You've changed daysMissed to return true or false.

crookedneighbor, line 60: I'll investigate some more.

crookedneighbor, line 60: Well, the good news is that it is working as intended, so it must be something wrong with how the test is written. I'm guessing that since the object is no longer a part of the user, the mock daily object isn't actually being updated.

paglias, line 60: @crookedneighbor shouldCronRun initially returned true or false but then I moved it to return numbers so it shouldn't be the problem. The problem is either with js inheritance or in cron itself but since I didn't change anything in cron the second option seems a bit strange

website/src/routes/apiv1.js
```diff
@@ -1,173 +1,13 @@
 var express = require('express');
```
crookedneighbor, line 1: Do we even need this anymore? Is api v1 supported at all?

paglias, line 1: It was deprecated but working, we can keep the message for 1-2 weeks and then remove it

Issue #4602, opened 2015-01-31
Clicking on a challenge from a guild opens the challenges page with all my guilds' challenges displayed at once


Salstar24:
Hi, never reported bugs before so I hope I'm doing this right!

When I click on a challenge from any guild page, it opens the Social -> Challenges page with all my guilds ticked and all their challenges displaying. I can work round this by manually clicking Filter -> Groups to 'None' and then re-clicking just the guild I was interested in to view their challenges. Still, I gather this isn't what's supposed to happen?

I use the Google Chrome browser, version 40.0.2214.93 m.
##

gkjt:
I do not think this is browser specific, I am getting it on Firefox 35.0.1 too.

crookedneighbor:
The intended behavior is for the screen to scroll down to the open challenge on the challenges page (it doesn't, obviously), but I love the idea of setting the filter to only show challenges for that particular group.

Any devs opposed to that? If so, why?

lemoness:
Not opposed - I think that would be a nice touch. Still, would like the
intended scroll to occur as well.

Salstar24:
Just wanted to report that the scroll thing did work once for me, the very first time I had tried it that day, but then it stopped working. Maybe it's particular guilds that don't trigger the scrolling? :/

(PS Just wanted to say that I'm super-impressed by the quick response and in general the commitment to responding across the website to newbies like me. :+1: )

TheHollidayInn:
Is this ticket a duplicate of https://github.com/HabitRPG/habitrpg/issues/1797 ?

deilann:
No. #1797 is about making sure when you click on a link to a challenge you
don't have to scroll to see it. This ticket is about when you click on a
link to a group challenge from a group, the filters should show only
challenges from that group.

TheHollidayInn:
Ah, well it seemed similar because of what @crookedneighbor said the intended behavior is.

Alys:
Fixed by @TheHollidayInn in https://github.com/HabitRPG/habitrpg/pull/4861 which should appear on the HabitRPG website in a couple of days.

Thanks for suggesting this @Salstar24! I've tested the fix and it really is quite lovely to see the filtered view of challenges.

Alys:
[...]

PR #5968, opened 2015-09-23
Added param to filter challenges by guild when a challenge is first clicked from guild

TheHollidayInn:
This is a fix for #4602 and a reboot for https://github.com/HabitRPG/habitrpg/pull/4861 . I tried to simply update, but I deleted the old branch. 

This adds a parameter to the routes that will be set if a user clicks to view a challenge from the guild, then hits back to challenges from the challenge detail page. 

I tried adding tests, but I was unable to call the private method. I then tried to set the variable and call digest, but that causes a need for mocking a few $http requests. I wasn't sure if I was allowed to move the function to a $scope function. So, any help there would be nice.

### DIFF
--- a/test/spec/controllers/challengesCtrlSpec.js
+++ b/test/spec/controllers/challengesCtrlSpec.js
@@ -148,6 +148,19 @@ describe('Challenges Controller', function() {
         expect(scope.filterChallenges(notOwnMem)).to.eql(false);
         expect(scope.filterChallenges(notOwnNotMem)).to.eql(true);
       });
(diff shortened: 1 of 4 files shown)


crookedneighbor:
Blocked by the api test pr.

crookedneighbor:

crookedneighbor:
Never mind my last comment, this doesn't do anything server side.

SabreCat:
Will be live soon! Noted toward tier 7.

PR #5972, opened 2015-09-24
World Boss dry run

SabreCat:
Adds a placeholder version of the upcoming World Boss to the site. With this in place, we can initialize the quest in the database and watch how fast the community progresses on it, without actually displaying the boss in the Tavern. That will help us tune the final stats for the monster before rolling it out for real.


SabreCat:
Per conversation with @vIiRuS, this approach would likely cause a bogus world boss to appear in mobile apps. Closing this PR; gonna take a different tack for our dry run.

PR #5973, opened 2015-09-24
Track daily quest progress in server-side analytics

SabreCat:
Adds progressUp and progressDown event properties to the Cron analytics event. With these, we can total up the site population's quest progress, both positive and negative, for the purposes of estimating likely pace of battle with a World Boss. Note that these values are tracked regardless of whether or not the user is currently questing.


crookedneighbor:
Seems reasonable to me. How easy is it to aggregate these stats in Anplitude?

PR #5976, opened 2015-09-25
Refactor Content Files

crookedneighbor:
## Problem

The Content file is too large and unwieldy and we're moving away from coffeescript.
## Solution
- Splits content files into separate modules instead of one gigantic file
- Uses ES2015. Files are located in `/common/script/src/content` and get compiled to `/common/dist/script/content/`. The idea will be to move them back to `common/script/content` once all of the common scripts are converted form coffeescript to es6
## Todos
- [ ] Add comment blocks at the top of the file for each section explaining the data type for each kind of content (explain the magic that the helpers are doing).
- [x] Decide if the warrior gear set is the correct way to organize the data and if so, do the same for the rest of it (can be done after merging)
- [x]  Add some testing around each content file to make sure their attributes that resolve into text don't turn into "Error processing string" or "String not found"
- [x] Extract more looping logic into helpers
## Misc Notes:
- In order to support better localization I added mountText for each mount. Since the word for the pet versus mount could be different in other languages.


common/script/src/content/mystery-sets.js
```diff
@@ -0,0 +1,128 @@
+import {each, where} from 'lodash';
+import {flat as flattenedGear} from './gear/index';
+
+let mysterySets = {
+  201402: {
+    start: '2014-02-22',
+    end: '2014-02-28',
+    text: 'Winged Messenger Set'
```
SabreCat, line 8: I don't know why we don't translate these at present. Make them into `translator` calls?

common/script/src/content/pets-mounts/special-mounts.js
```diff
@@ -0,0 +1,13 @@
+// special mounts are {key:i18n}
```
SabreCat: Maybe not in this PR, but I would love to see this structure go away.

common/script/src/content/quests/gold-purchasable.js
```diff
@@ -0,0 +1,119 @@
+import {each, defaults} from 'lodash';
+import {
+  translator as t,
+  merge,
+  formatForTranslator
+} from '../helpers';
+
+let dilatoryDistressSeries = {
```
SabreCat, line 7: If we make quest series into arrays, we could handle the sequence via index rather than the weird `previous` attribute.

common/script/src/content/time-traveler-store.js
```diff
@@ -0,0 +1,15 @@
+import {keys, reduce} from 'lodash';
+import mystery from './mystery-sets';
+
+function timeTravelerStore(owned) {
+  var ownedKeys = keys((typeof owned.toObject === "function" ? owned.toObject() : void 0) || owned);
```
SabreCat, line 5: Lines like this are one of the handful of things I'll miss CoffeeScript for. :\

common/script/src/content/quests/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+let allQuests = merge([
+  worldQuests,
+  holidayQuests,
+  petQuests,
+  unlockableQuests,
+  goldPurchasableQuests
+]);
+
+each(allQuests, function(quest, key) {
```
crookedneighbor: @TODO: Move to helpers

common/script/src/content/quests/pet.js
```diff
@@ -0,0 +1,365 @@
...
+      str: 1.5
+    },
+    drop: {
+      gp: 37,
+      exp: 275,
+    }
+  }
+};
+
+each(petQuests, (quest, name) => {
```
crookedneighbor: @TODO: Move to helpers

common/script/src/content/time-traveler-store.js
crookedneighbor, line 5: I plan to refactor this after this gets merged in to make it more readable.

package.json
```diff
@@ -45,6 +45,8 @@
     "jade": "~1.11.0",
     "js2xmlparser": "~0.1.2",
     "lodash": "~2.4.1",
+    "lodash.camelcase": "^3.0.1",
```
crookedneighbor, line 49: These two modules should be removed after we upgrade lodash to v3

common/script/src/content/pets-mounts/special-mounts.js
crookedneighbor: Yes, definitely.

paglias:
@crookedneighbor just a quick question, why are we transforming styles instead of just using babel's require hook? https://babeljs.io/docs/setup/#babel_register

crookedneighbor:
@paglias I think I was having issues because we're using coffeescript, which also takes over the require hook. Eventually we'll switch out Browserfy for Babelify. 

Now that I've fully switched over the content files, I'll see if I can wrangle the require hook to work.

Alys:
bug report: When you equip a Spooky pet or mount, the change appears on the client but is not saved on the server - a sync will revert you to your previously equipped pet or mount. I have not tried **every** other kind of pet and mount, but have tried several and it's only happening for Spooky ones.

Alys:
bug report: Feeding a Spooky pet works on the client but is not saved to the server. A sync will return the food to the food tray and return the pet to unfed. This doesn't happen with the few Generation 1 pets that I've tried it with. I have not tried it with quest pets.

Alys:
bug report: Hatching a Spooky pet causes the egg and potion to disappear permanently (not regained on sync), but the pet does not appear on the pet page. I have not noticed this with hatching golden Generation 1 pets (they work as expected). I have not tried with other colours or types of pets.

Alys:
FYI not a bug - these things all work:
- Key to the Kennels for releasing pets, and mounts, and both.
- Fortify Potion
- Orb of Rebirth (only brief testing done after using it)
- all four avatar buffs (Spooky Sparkles, etc)
- sending gift cards
- Resting in the Inn
- inviting a player to a party

I'll update this comment with other things that work - not everything, just the things that would be likely to make people scream if they did have bugs. :)

crookedneighbor:
Whoops. I created the logic for premium pets, but forgot to actually export them.

Issue #5928, opened 2015-09-15
Missing/Wrong link in faq.json


ghost:
In the new `faq.json` file, the string `iosFaqAnswer10` contains the following markdown text: [...]`Subscribe on the [website](link) and unlock the ability`[...], while it should certainly be [...]`Subscribe on the [website](https://habitica.com/#/options/settings/subscription) and unlock the ability`[...]
##

librarianmage:
Has this been taken care of?

Alys:
Yes, thanks!

PR #6000, opened 2015-10-02
Issue #5928 - Replaced 'link' with actual URL in faq.json

Gondulf:


crookedneighbor:
I've awarded you two contributor tiers for both your pull requests. You can read more about [contributor tiers and rewards here](http://habitrpg.wikia.com/wiki/Contributor_Rewards).

Issue #5811, opened 2015-08-18
Feature request. sell all option. 


SHMAUS-Carter:
I always have an abundance of eggs and would love to have the ability to sell them all vs having to click the sell button 40 or more times. 
##

crookedneighbor:
@SHMAUS-Carter Thanks for suggesting it. I think that's a fine idea.

In the future, feature requests [should go through our Trello board](https://trello.com/c/odmhIqyW/440-read-first-how-to-request-a-feature-table-of-contents) so our community can vet them first. Going to leave this open though, because I talked with the staff already, and they think this is a great idea. 

Will advertise it in the [aspiring coders guild](https://habitica.com/#/options/groups/guilds/a141a5b9-0d69-46cd-93e6-2b76f18bcc54).

crookedneighbor:
Whoops, linked to the aspiring pixel artists guild by mistake. Here's the link for aspiring coders: https://habitica.com/#/options/groups/guilds/68d4a01e-db97-4786-8ee3-05d612c5af6f

SHMAUS-Carter:
I do apologize for posting in the wrong section I did not even know you had a Trello board that you use for feature request. I'm so used to feature request and issues being reported in the same place.

crookedneighbor:
No worries. Like I said, this is a great idea. 

We do feature requests on Trello so a broader community (more than just devs) can discuss it.

SHMAUS-Carter:
That makes sense. I keep forgetting that your average user (broader community) dose not know of github much less take an active role in filling out issues etc.

melarish:
I'll try building this. Is there any way to assign yourself to an issue?

Alys:
No there isn't. Just commenting as you did is enough. :)

Ask here if you run into problems or want advice!

melarish:
Here's what I've got (the button is there for eggs, pots and food but no point uploading screens for all). Does the whitespace need any improvements? Should there be a proper plural of 'Strawberry' or leave it as is?
<img width="1332" alt="screenshot 2015-09-27 23 48 03" src="https://cloud.githubusercontent.com/assets/5783115/10125620/279a2162-6573-11e5-99e5-2026988ac913.png">
<img width="1318" alt="screenshot 2015-09-27 23 48 29" src="https://cloud.githubusercontent.com/assets/5783115/10125622/327c1b26-6573-11e5-8a13-3e511544b2d4.png">

crookedneighbor:
[...]

PR #6005, opened 2015-10-03
 Feature request. sell all option. #5811

melarish:
My local Habit doesn't work after the most recent git pull but it was working earlier so should still be OK (as long as Spooky Potions didn't change code for regular items).


website/public/js/controllers/inventoryCtrl.js
```diff
@@ -103,6 +103,18 @@ habitrpg.controller("InventoryCtrl",
       }
     }
 
+    $scope.sellAllInventory = function() {
+      var selected = $scope.selectedEgg ? 'selectedEgg' : $scope.selectedPotion ? 'selectedPotion' : $scope.selectedFood ? 'selectedFood' : undefined;
+      if (selected) {
+        var type = $scope.selectedEgg ? 'eggs' : $scope.selectedPotion ? 'hatchingPotions' : $scope.selectedFood ? 'food' : undefined;
+        var amount = user.items[type][$scope[selected].key]; // user.items changes with every iteration of sell, so need to store in a variable first
+        for (i = 0; i < amount; i++) {
+          user.ops.sell({params:{type:type, key: $scope[selected].key}});
```
crookedneighbor, line 112: Rather than calling a for loop, I think it would be better to adjust the user ops function to accept a quantity and have it default to 1.

crookedneighbor, line 112: Doing that would also allow us to just use the `sellInventory` function and pass in a quantity parameter if more than one item is being sold.

crookedneighbor:
@melarish Any update on this?

melarish:
I was fiddling with it the other day but didn't quite get it to work. Haven't had time to look at it again but definitely still on it!

crookedneighbor:
This will require changes to the v2 of the API. We are actively starting work on v3 of the api, so we definitely want this change, but it's going to have to wait until v3.

Going to close this now, but know that the change _is_ still desired.

PR #6008, opened 2015-10-04
Typo in comment string

locks:
Replaces `euqipment` with `equipment`


crookedneighbor:
I will count this toward a contributor tier.

What's your User ID? You can find it in [Settings > API](https://habitrpg.com/#/options/settings/api). Be sure to only copy the User ID and _not_ your API Token.

locks:
Appreciated :D
My User ID is `167b0c32-713a-412d-b589-d745d158a9b4`.

crookedneighbor:

PR #6009, opened 2015-10-04
Typo in test description string

locks:
Replaces `euqipment` with `equipment`.


crookedneighbor:
Thanks for this. I'm including this as part of your first tier award, since both changes were to typos (the same word even!)

locks:
Makes sense! :)

PR #6013, opened 2015-10-05
Added group member api endpoint with offset and limit

TheHollidayInn:
I would like use to add paging, infinite scroll, and member searching to groups (and eventually challenges). I believe it starts with a member endpoint that contains offset and limits. Is this the correct way to begin? If so, then I can begin adding the front-end as well as some other parameters for navigating members.

### DIFF
--- a/website/src/controllers/groups.js
+++ b/website/src/controllers/groups.js
@@ -452,28 +452,25 @@ api.members = function(req, res, next) {
   var limit = req.body.limit;
   var offset = req.body.offset;
 
-  var q = (gid == 'party')
-    ? Group.findOne({type: 'party', members: {'$in': [user._id]}})
-    : Group.findOne({$or:[
-        {_id:gid, privacy:'public'},
-        {_id:gid, privacy:'private', members: {$in:[user._id]}} // if the group is private, only return if they have access
-      ]});
+  if (limit > 15) {
+    return res.send(401, {err: "Limit too high"});
+  }
+
+  var q = Group.findOne({
+    $or:[
+      {_id:gid, privacy:'public'},
+      {_id:gid, privacy:'private', members: {$in:[user._id]}} // if the group is private, only return if they have access
+    ]
+  }, {members: 1});
 
   q.populate({
     path: 'members',
     select: 'name _id',
     options: { limit: limit, offset: offset },
   });
-  
+
   q.exec(function(err, group){
     if (err) return next(err);
-    if(!group){
-      if(gid !== 'party') return res.json(404,{err: "Group not found or you don't have access."});
-
-      // Don't send a 404 when querying for a party even if it doesn't exist
-      // so that users with no party don't get a 404 on every access to the site
-      return res.json(group);
-    }
     return res.json(group);
     gid = null;
   });
--- a/website/src/controllers/groups.js
+++ b/website/src/controllers/groups.js
@@ -443,6 +443,42 @@ api.likeChatMessage = function(req, res, next) {
   })
 }
 
+/**
+ * Get group members
+ */
+api.members = function(req, res, next) {
+  var user = res.locals.user;
+  var gid = req.params.gid;
+  var limit = req.body.limit;
+  var offset = req.body.offset;
+
+  var q = (gid == 'party')
+    ? Group.findOne({type: 'party', members: {'$in': [user._id]}})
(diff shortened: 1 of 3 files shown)


paglias:
@TheHollidayInn this is great, two small notes though:
- Anything that could result in degraded performances (a lot of queries, big populate queries is off limits until we start working on an API v3 (will happen!) so a member endpoint ideally should be limited to a maximum limit of users, I would say 15 so that it's the same as the members list
- Regarding API v3 it's possible that once we start working on it, we'll stop accepting pull requests for API v2 and eventually all methods we decide to keep will have to be migrated, but won't happen in at least a month so feel free to work on it

website/src/controllers/groups.js
```diff
@@ -443,6 +443,42 @@ api.likeChatMessage = function(req, res, next) {
...
+  var user = res.locals.user;
+  var gid = req.params.gid;
+  var limit = req.body.limit;
+  var offset = req.body.offset;
+
+  var q = (gid == 'party')
+    ? Group.findOne({type: 'party', members: {'$in': [user._id]}})
+    : Group.findOne({$or:[
+        {_id:gid, privacy:'public'},
+        {_id:gid, privacy:'private', members: {$in:[user._id]}} // if the group is private, only return if they have access
```
crookedneighbor: parties always have a privacy type of `'private'`, couldn't we just adjust the query based on that?

crookedneighbor:
Haven't had a chance to look at the code super closely, but this route should not return the whole group object. The object it returns should probably look something like:

``` js
{
  _id: group_id
  members: [ 
    // members 
  ],
  limit: number,
  offset: number,
  total_number_of_offsets: number
}
```

@paglias Thoughts on that?

website/src/controllers/groups.js
```diff
@@ -443,6 +443,42 @@ api.likeChatMessage = function(req, res, next) {
...
+  q.populate({
+    path: 'members',
+    select: 'name _id',
+    options: { limit: limit, offset: offset },
+  });
+  
+  q.exec(function(err, group){
+    if (err) return next(err);
+    if(!group){
+      if(gid !== 'party') return res.json(404,{err: "Group not found or you don't have access."});
```
crookedneighbor: This shouldn't be necessary for the members route.

website/src/controllers/groups.js
```diff
@@ -443,6 +443,42 @@ api.likeChatMessage = function(req, res, next) {
   })
 }
 
+/**
+ * Get group members
+ */
+api.members = function(req, res, next) {
+  var user = res.locals.user;
+  var gid = req.params.gid;
+  var limit = req.body.limit;
```
crookedneighbor, line 452: This should probably be something like:

```
var limit = req.body.limit;

if (limit > 15) {
  return res.send(401, 'Limit too high');
}
```

paglias:
@crookedneighbor agreed. Also it's not clear to me where this route should be used? If the aim is to get a list of groups and make it searchable, the members route is not necessary

crookedneighbor:
@paglias I was under the impression that this would be used to populate the members section of the group page and you could paginate through them to see more members.

website/src/controllers/groups.js
TheHollidayInn: Yep, I just took this restriction query from the Get api endpoint.

TheHollidayInn:
Thanks, guys. 

@paglias sounds good about v3. I will help with that when needed. Also, this api route is for paging the members, just like @crookedneighbor suggests.

@crookedneighbor if I return the total count of the members, I will need to do a second query (I believe). Is that alright?

crookedneighbor:
This is currently it's blocked by a [new API test infrastructure we're building](https://github.com/HabitRPG/habitrpg/pull/6062). As soon that is merged in, we can add tests for this behavior and merge this in. 

If you'd like to contribute to the tests so it can be merged in sooner, just comment on the PR with ones you are doing and when you're done, PR into the branch:

```
https://github.com/crookedneighbor/habitrpg/compare/new_api_test_infrastructure...<your_github_username>:<your_branch_to_pr>
```

crookedneighbor:

crookedneighbor:
We're accelerating our work on V3 of the API, as such, we're not going to accept anymore V2 PRs. Note that this is a desired feature, so we'll want it to be a part of v3.

PR #6014, opened 2015-10-05
Toggle chevron direction

ahstro:
When editing a task, there are some expandable options available with a chevron next to them (e.g. 'Tags' and 'Advanced Options'). This commit makes the chevrons' directions toggle based on whether the related option is expanded or collapsed, up and down respectively.

Before:
habitica-chevron-before
After:
habitica-chevron-after

If I did everything correctly, this should only apply to elements with the `option-title` and `mega` classes; I've tested all the tasks I could find, and it seems to work fine, but I'm not overconfident with the Habitica codebase, so I recommend QAint properly before merging (if it's a desired change, that is).

### DIFF
--- a/website/public/css/tasks.styl
+++ b/website/public/css/tasks.styl
@@ -441,6 +441,8 @@ form
         font-size: 0.75em
         content: "\E114"
         padding-left: 0.75em
+      &.active:after
+        content: "\E113"
     //&:not(.mega):after
       //content: ':'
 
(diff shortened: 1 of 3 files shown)


crookedneighbor:
Finally got around to reviewing this. It's great!

ahstro:
Oh, cool. My user ID is `9e763ea4-fc38-432f-bf5e-161014fa5283`

crookedneighbor:

PR #6020, opened 2015-10-07
Fixes issue #5550

### LINKED ISSUE #5550: Wrong max level messages after rebirth
I used the Orb of Rebirth after reaching level 104. All related messages and texts state that my max level was 100. I had my only other rebirth at 100.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23540721-wrong-max-level-messages-after-rebirth?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Thrashmandicoot:
Resolves https://github.com/HabitRPG/habitrpg/issues/5550

@crookedneighbor not sure if stepping on toes, saw you had started on issue, but not sure if you had resolved.


Alys:
What we're wanting for this fix is really just a text change, not a behaviour change. Habitica is optimised for up to level 100 and no further. Players can proceed further for as long as they wish, but the game treats all levels above 100 as if they were 100. So for Rebirth, when you Rebirth at level 100 or higher, the messages you see are the same, no matter the level:
- "Reborn, max level 100+"
- "The highest level you've attained is 100 or higher!"
- "...begin your next new Adventure when you are at least level 100."

If you use Rebirth below level 100, then it lists the exact level.  That's what lemoness was getting at in https://github.com/HabitRPG/habitrpg/issues/5550#issuecomment-119012181

So, in the code, we still need to use api.capByLevel. The change is that when `lvl` = 100 (after api.capByLevel has been applied), then the messages should appear as above, instead of saying merely "100" as they do now. Let me know if I haven't explained this clearly!

Thrashmandicoot:
hi @Alys, I think to fully resolve this problem, we need more than a text change. Perhaps it's worth a discussion?

I'm not actually changing any of the level cap functionality other than removing it from the rebirth function, specifically on line 545 (perhaps I was unclear on this in the commit message). 

The rebirth function, from my understanding, takes the users existing level and stores it in a variable before starting the character over to lvl 1. The lvl is then applied to the achievement for rebirth, and to their max lvl for rebirth, which then displays on the users character page.

The text fix suggestion would only partially fix the problem as it's static and doesn't update variables that may matter to players.

For instance, as a player, what if I wish to know what the highest lvl I attained was after rebirth?
"Reborn, max level 100+" gives an answer that it was over 100, but doesn't let me know if I say, hit 131 etc.

I've included a screencapture that shows how this simple code change fixes the issue fully, based on my change:
lvlfix

Unless I'm not seeing something important, this minor code change will only impact the variables used for rebirth achievement, and the rebirth icon that displays the players max lvl as per my screenshot of a character who's been rebirthed after my code fix.

Alys:
I do understand what you're wanting to do in your code change, however that change isn't how we'd like #5550 to be fixed. Because Habitica is optimised for levels up to 100 only, we make no distinction for any levels above 100. You can reach higher levels, but your attribute points don't increase after 100, and the Rebirth message deliberately doesn't state the level after 100. 

The problem with the current message is that it makes it sound like you rebirthed at exactly level 100. We want it changed so that it indicates that you might have rebirthed at level 100 or above. However we don't want the exact level to be mentioned if it is above level 100. 

In brief, the text is meant to be static when the level is 100 or above. For details, please see lemoness's comments in https://github.com/HabitRPG/habitrpg/issues/5550.

Thrashmandicoot:
I'll close this pull request then.

Alys:
I'm sorry your time was wasted on this! Thank you for being willing to contribute.

If you're looking for other issues to work on, the [entry level coding](https://github.com/HabitRPG/habitrpg/issues?utf8=%E2%9C%93&q=label%3A%22entry+level+coding%22+) issues are good if you're new to Habitica's code, and there's many other issues that are more complex if that's what you're after.

Also, there's this new issue that I thought of while we were discussing this: https://github.com/HabitRPG/habitrpg/issues/6021
It's not ready to be worked on yet, but it might interest you.

Thrashmandicoot:
Hi @Alys, definitely not a time waster. It's a good thought exercise, and I appreciate your feedback on my fixes, it only helps me get better when working on new projects.

If you don't mind me going further, and this also may be a reason this issue hasn't been fixed yet (it was opened in July) is because the simple changes being asked for impact a lot more points in the code than not, due to embeds.

If we want to change the outputs of each piece individually, here's the points of code we need to touch:

lines 124 -> 125 of achievements.jade:

``` javascript
=env.t('rebirthOrb')
    |&nbsp;{{::profile.achievements.rebirthLevel}}.
```

the above impacts line 21 of rebirth.json here:

``` javascript
"rebirthOrb": "Used an Orb of Rebirth to start over after attaining Level"
```

we also impact line 18 of rebirth.json here:

``` javascript
"rebirthAchievement": "You've begun a new adventure! This is Rebirth <%= number %> for you, and the highest Level you've attained is <%= level %>. To stack this Achievement, begin your next new adventure when you've reached an even higher Level!"
```

to add text here we need to either put a conditional in place before the embed is calculated that attaches 100+ as a string (from a number) if the player is 100 or over, or we need to add the calculation within the embed itself (the uglier option).

we also need to change index.jade at lines 29 and 30:

``` javascript
span.glyphicon.glyphicon-plus-sign(ng-show='profile.achievements.rebirths',
        tooltip=env.t('reborn', {reLevel: "{{profile.achievements.rebirthLevel}}"}), tooltip-append-to-body="true")
```

the above will impact the environment variable used in line 24 of rebirth.json:

``` javascript
"reborn": "Reborn, max level <%= reLevel %>"
```

The above is why I'm hesitant, because we go from a simple 1 line fix (my original contrib), to hitting 5 different areas of code across 3 files.

Perhaps I'm bias because the solution I suggest is my own, but I feel if the simple solution offers more functionality (in terms of accurately giving the player their previous lvl) and impacts less in terms of code, why wouldn't we choose the lazy way ;).

Alys:
Yes, when I said we wanted just a text change, that was somewhat misleading. I meant that the desired end result in the front end is just a text change (e.g., from "100" to "100+"), rather than a behaviour change (from "100" to "123"). The code changes will be much as you describe. 

In fact they'll probably be a little more complex than you describe because we need to consider translations for languages that have a different word order than English. For example, for the sentence "The highest level you've attained is [n]/[100 or higher]!", we might need to implement it as two different strings: 
- "The highest level you've attained is [n]!" (for n < 100, where n is a variable)
- "The highest level you've attained is 100 or higher!" (static text with no variables)

That will probably be necessary because we can't simply insert "100 or higher" at the end of the sentence and expect it to make sense in all languages.

A change to insert the exact level for levels higher than 100 would of course be much simpler, but it does not match our intention of making no distinction for any levels above 100.

Thrashmandicoot:
Maybe we just change both to "The highest level you've attained is 100+!"?

It would make sense to players at both level 100 and level 101 etc.

PR #6033, opened 2015-10-10
Add iOS smart app banner

phillipthelen:
Adds a smart app banner to the front page that informs users of our iOS app when opened on iOS


paglias:
@vIiRuS screenshot? Just to see how much space it takes

phillipthelen:
img_0001

PR #6034, opened 2015-10-11
prevent user being removed from challenge if deleting challenge tasks fails - fixes #5506

### LINKED ISSUE #5506: if deleting challenge tasks fails, user is removed from challenge anyway
When leaving a challenge and deleting challenge tasks, a rare bug can cause the challenge ID to become "lost" at the point where the tasks are deleted. This used to result in all non-challenge tasks being deleted (https://github.com/HabitRPG/habitrpg/issues/4466) but now it merely results in the tasks remaining while the user is removed from the challenge (https://github.com/HabitRPG/habitrpg/pull/5504). This means that the user has to delete the tasks manually.

It would be better if the user was not removed from the challenge if deleting the tasks fails. I.e., the code that removes the user should check for success/failure of deleting the tasks and not take other actions if the deletion is not successful.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23201751-if-deleting-challenge-tasks-fails-user-is-removed-from-challenge-anyway?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremyabbott:
This is a pre-PR PR for review/guidance. I still need to write tests for the modified behavior.

### DIFF
--- a/website/src/controllers/challenges.js
+++ b/website/src/controllers/challenges.js
@@ -403,6 +403,14 @@ api.leave = function(req, res, next){
 
   async.waterfall([
     function(cb){
+      var i = user.challenges.indexOf(cid)
+      if (~i) user.challenges.splice(i,1);
+      user.unlink({cid:cid, keep:keep}, function(err){
+        if (err) return cb(err);
+        cb(null);
+      });
+    },
+    function(cb){
       Challenge.findByIdAndUpdate(cid, {$pull:{members:user._id}}, cb);
     },
     function(chal, cb){
@@ -411,13 +419,8 @@ api.leave = function(req, res, next){
       // _.size(challenge.members). We can't do it in pre(save) because we're calling findByIdAndUpdate above.
       if (chal)
         Challenge.update({_id:cid}, {$set:{memberCount:_.size(chal.members)}}).exec();
-
-      var i = user.challenges.indexOf(cid)
-      if (~i) user.challenges.splice(i,1);
-      user.unlink({cid:cid, keep:keep}, function(err){
-        if (err) return cb(err);
-        cb(null, chal);
-      })
+        
+      cb(null, chal);
     }
   ], function(err, chal){
     if(err) return next(err);
--- a/website/src/controllers/user.js
+++ b/website/src/controllers/user.js
@@ -67,7 +67,7 @@ var findTask = function(req, res) {
 */
 
 /**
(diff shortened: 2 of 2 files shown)


crookedneighbor:
This is currently blocked by a [new API test infrastructure we're building](https://github.com/HabitRPG/habitrpg/pull/6062). As soon that is merged in, we can add tests for this behavior and merge this in. 

From your description, it looks like you need some guidance? What are your questions?

jeremyabbott:
@crookedneighbor thanks for getting back to me. I'm actually not stuck... I just haven't had time to go back and write tests.

jeremyabbott:
@crookedneighbor do you need help with the API Test migration?

crookedneighbor:
Sure do. :)

If you'd like to contribute to that PR, you can do so by creating a pull request into that branch, like this:

```
https://github.com/crookedneighbor/habitrpg/compare/new_api_test_infrastructure...<your_github_username>:<your_branch_to_pr>
```

Just replace `<your_github_username>` and `<your_branch_to_pr>` with the applicable names.

jeremyabbott:
Awesome. I'm looking at the new pattern for tests now. I'm somewhat of a node noob. Which gulp task actually runs the tests? I've tried `gulp test:api` and it starts but never finishes.

crookedneighbor:
`gulp test:api` is the correct command. It requires Mongo to be running. I'll have to add a check for that and throw an error if it is not running.

jeremyabbott:
Got it working! Thanks for the feedback!

crookedneighbor:
The PR is merged in. Let me know if you need help writing tests.

jeremyabbott:
I started working on a test for the PUT /user end point based on the pattern you started. I'll try to finish it up tomorrow. I know there shouldn't be much to it. Should I create an issue for it? Thanks for the heads up!

crookedneighbor:
Don't create an issue, but definitely open up a separate pull request.

crookedneighbor:
As we begin working on API v3, we're stopping work on API v2, so I'm going to close this now.

PR #6042, opened 2015-10-13
More and better modals; Social Media sharing

SabreCat:
Moved checklists to Trello: https://trello.com/c/ydzUjKH6/340-social-media-sharing

Feedback requested:
- Current layouts require a lot of `style` attributes to get things to line up correctly, because they use `herobox` and `character-sprites` elements designed to arrange the box used in the site's upper-left corner and in party display. (At least I've figured out how not to do hacky things like negative margins!) Are there better approaches we could use here?


website/public/js/controllers/notificationCtrl.js
```diff
@@ -53,6 +60,19 @@ habitrpg.controller('NotificationCtrl',
        Notification.mp(mana);
     });
 
+    $rootScope.$watch('user.stats.lvl', function(after, before) {
+      if (after === before) return;
```
crookedneighbor: Shouldn't this be `if (before >= after) return;`?

SabreCat: Ah, sure. That'd mean we can remove an extra `if` at line 69, as well. I thought I recalled watches acting weird if you didn't have a strict-equality check at the beginning, but maybe we're safe when we're working with a numeric attribute no user will ever lack.

website/public/js/services/socialServices.js
```diff
@@ -0,0 +1,41 @@
...
+          js = d.createElement(s); js.id = id;
+          js.src = "//connect.facebook.net/en_US/sdk.js#xfbml=1&version=v2.5";
+          fjs.parentNode.insertBefore(js, fjs);
+        }(document, 'script', 'facebook-jssdk'));
+      } else {
+        FB.XFBML.parse(); // http://stackoverflow.com/questions/29133563/
+      }
+
+      // Tumblr
+      $.getScript('https://assets.tumblr.com/share-button.js');
```
SabreCat, line 27: It would be lovely if Tumblr has a way, like Twitter and Facebook, to rebuild widgets with the script already loaded. But I couldn't find one. So this results in the Tumblr script potentially getting loaded into memory multiple times over the course of the session. Maybe we can reverse-engineer `share-button.js` a bit? :P

website/views/social/level-up.jade
```diff
@@ -0,0 +1,27 @@
+doctype html
+html(ng-app='habitrpg')
+  head
+    meta(charset='utf-8')
+    link(href='//fonts.googleapis.com/css?family=Lato:400,700', rel='stylesheet', type='text/css')
+    title Habitica | Your Life the Role Playing Game
+    script(type='text/javascript').
+      window.env = !{JSON.stringify(env._.pick(env, env.clientVars))};
+    != env.getManifestFiles("tmp_static_front")
```
SabreCat: Adding these three lines is when `social/level-up` started redirecting to `static/front`.

website/public/js/controllers/notificationCtrl.js
```diff
@@ -53,6 +60,19 @@ habitrpg.controller('NotificationCtrl',
        Notification.mp(mana);
     });
 
+    $rootScope.$watch('user.stats.lvl', function(after, before) {
```
SabreCat, line 63: I tried to write a test for this, but the `$watch` function didn't seem to run upon a `scope.$digest()` during tests.

website/views/social/level-up.jade
SabreCat: Maybe I need a new set of manifest data instead of trying to reuse `tmp_static_front`?

website/public/js/services/socialServices.js
crookedneighbor, line 27: Can't you set a variable `tumblrIsLoaded` and check for that before loading?

SabreCat, line 27: That'd prevent fetching the script more than once, but it wouldn't activate the piece that goes and sets up the button.

common/locales/en/front.json
```diff
@@ -74,7 +74,7 @@
     "infhQuote": "[Habitica] has really helped me impart structure to my life in graduate school.",
     "invalidEmail": "A valid email address is required in order to perform a password reset.",
     "irishfeet123Quote": "I've had horrible habits with clearing my place completely after meals and leaving cups all over the place. [Habitica] has cured that!",
-    "joinOthers": "Join 850,000 people making it fun to achieve goals!",
+    "joinOthers": "Join 900,000 people making it fun to achieve goals!",
```
crookedneighbor, line 77: This should be string with an interpolation:

```
Join <%= number %> people making it fun to achieve goals!
```

common/locales/en/front.json
```diff
@@ -182,5 +182,11 @@
     "tweet": "Tweet",
     "apps": "Apps",
     "notifyAndroidApp": "Want us to notify you when the Android app is ready? Sign up for this mailing list!",
-    "checkOutIOSApp": "Check out our new iOS App!"
+    "checkOutIOSApp": "Check out our new iOS App!",
+    "imagine1": "Imagine if improving your life were as fun as playing a game.",
+    "landingCopy1": "Advance in the game by completing your real-life tasks.",
+    "landingCopy2": "Battle monsters with friends to stay accountable to your goals.",
+    "landingCopy3": "Join over 900,000 people having fun as they improve their lives.",
```
crookedneighbor, line 189: Same here. 

```
Join over <%= number %> people having fun as they improve their lives.
```

test/spec/controllers/inventoryCtrlSpec.js
```diff
@@ -53,14 +56,36 @@ describe('Inventory Controller', function() {
...
+      expect(rootScope.openModal).to.have.been.calledOnce;
+      expect(rootScope.openModal).to.have.been.calledWith('hatchPet');
+    });
+    
+    it('does not show pet hatching modal if user has opted out', function(){
+      user.preferences.suppressModals.hatchPet = true;
+      scope.chooseEgg('Cactus');
+      scope.choosePotion('Base');
+
+      expect(rootScope.openModal).to.not.be.called;
```
crookedneighbor, line 87: Excellent tests!

website/public/js/controllers/inventoryCtrl.js
```diff
@@ -112,6 +115,10 @@ habitrpg.controller("InventoryCtrl",
       var potName = Content.hatchingPotions[potion.key].text();
       if (!$window.confirm(window.env.t('hatchAPot', {potion: potName, egg: eggName}))) return;
       user.ops.hatch({params:{egg:egg.key, hatchingPotion:potion.key}});
+      if (!user.preferences.suppressModals.hatchPet) {
+        $rootScope.hatchedPet = {egg: eggName, potion: potName, eggKey: egg.key, pet: 'Pet-' + egg.key + '-' + potion.key};
```
crookedneighbor, line 119: Why are we putting this on rootscope? Couldn't you just insert it into the scope of the modal?

website/public/js/controllers/inventoryCtrl.js
```diff
@@ -152,12 +160,16 @@ habitrpg.controller("InventoryCtrl",
         $scope.selectedFood = null;
 
         _updateDropAnimalCount(user.items);
+        if ($scope.mountCount > startingMounts && !user.preferences.suppressModals.raisePet) {
+          $rootScope.raisedPet = {displayName: petDisplayName, spriteName: pet, egg: egg, potion: potion}
```
crookedneighbor, line 164: Same question here, I think we can just insert this into the scope of the modal.

website/public/js/controllers/notificationCtrl.js
```diff
@@ -53,6 +60,23 @@ habitrpg.controller('NotificationCtrl',
...
+      if (after <= before) return; 
+      Notification.lvl();
+      $rootScope.playSound('Level_Up');
+      if (User.user._tmp && User.user._tmp.drop && (User.user._tmp.drop.type === 'Quest')) return;
+      if (after === 3) return; // Drop system unlock. FIXME can we do this without hardcoding?
```
crookedneighbor, line 68: You need a map. IE:

```
var skippableLevels = {
  '3': 'drop system',
  '10':  'class system',
  '50': 'orb of rebirth'
}

if (skippableLevels['' + after]) return;
```

crookedneighbor, line 68: the `'' + after` bit is to cast the number as a string.

website/public/js/services/notificationServices.js
```diff
@@ -108,7 +108,6 @@ angular.module("habitrpg").factory("Notification",
   var stack_topright = {"dir1": "down", "dir2": "left", "spacing1": 15, "spacing2": 15, "firstpos1": 60};
 
   function _notify(html, type, icon) {
-      console.log(type)
```
crookedneighbor, line 111: ABOUT TIME :)

website/views/shared/modals/achievements.jade
```diff
@@ -1,99 +1,134 @@
...
+          a.twitter-share-button(href='https://twitter.com/intent/tweet?text=#{tweet}&via=habitica&url=#{env.BASE_URL}/social/achievement&count=none')=env.t('tweet')
+        .col-xs-4
+          .fb-share-button(data-href='#{env.BASE_URL}/social/achievement', data-layout='button')
+        .col-xs-4
+          a.tumblr-share-button(data-href='#{env.BASE_URL}/social/achievement', data-notes='none')
+
+mixin achievementAvatar(badge,margin)
+  .container-fluid
+    .row
+      .col-xs-4(style='padding:0')
```
crookedneighbor, line 17: I'm not a fan of these inline styles. :/

website/views/social/achievement.jade
```diff
@@ -0,0 +1,9 @@
+extends ./layout
+
+block extraHead
+  - var socialPage = 'achievement'
+  - var socialTitle = 'Earned an Achievement!'
+  - var socialDescription = 'I earned a new Achievement on Habitica for working to improve my life!'
```
crookedneighbor, line 6: Did we want to make these translatable?

SabreCat, line 6: Ah, can we? I'm fuzzy on how that works. This is something that'd get crawled and displayed when somebody examines a Tweet (or FB share, etc.) on a timeline. Would the meta tags get translated then?

website/views/shared/modals/achievements.jade
SabreCat, line 17: Neither am I; I commented on that to some extent in the PR description. But so many things needed fiddly little tweaks that it seemed like no matter how I implemented it, it'd add a bunch of ugly clutter somewhere. Suggestions?

website/public/js/controllers/notificationCtrl.js
SabreCat, line 68: OK. Right in the same controller, then? I guess that would be easier to update, though it'd still be nice to have the milestone list in a more central place.

website/public/js/controllers/inventoryCtrl.js
SabreCat, line 119: How does that go? Use `$scope` instead of `$rootScope`? Though wouldn't that be on `inventoryCtrl`'s scope? (ugh, scopes)

test/spec/controllers/inventoryCtrlSpec.js
SabreCat, line 87: Needs more of them, though. See my comment above about trouble with `$watch` on some of it.

website/views/social/achievement.jade
crookedneighbor, line 6: I'm also fuzzy on that.

website/public/js/controllers/notificationCtrl.js
crookedneighbor, line 68: For now it should be in the controller. If there's a need to have it multiple places, we can put it in a milestones service.

website/public/js/controllers/inventoryCtrl.js
crookedneighbor, line 119: Let me see if I can find the thing I'm thinking about.

crookedneighbor, line 119: Yeah, you [inject it](http://stackoverflow.com/a/26747669/2601552) using [the resolve option](https://github.com/HabitRPG/habitrpg/blob/develop/website/public/js/controllers/rootCtrl.js#L135)

crookedneighbor, line 119: We really abuse `$rootScope` in the app :cry: 

If I remember correctly, objects on `$rootScope` don't get garbage collected as easily as those on `$scope`

SabreCat, line 119: Another bad pattern I've absorbed by this being my first major dev project. Add it to the pile. :disappointed:

crookedneighbor, line 119: It's not your fault. The app has been in this state for a while now. Bad code begets bad code...

website/public/js/controllers/inventoryCtrl.js
```diff
@@ -152,12 +160,16 @@ habitrpg.controller("InventoryCtrl",
         $scope.selectedFood = null;
 
         _updateDropAnimalCount(user.items);
+        if ($scope.mountCount > startingMounts && !user.preferences.suppressModals.raisePet) {
```
Alys, line 163: Using `mountCount` prevents the modal being shown when quest pets are raised to mounts (mountCount is for tracking Mount Master progress). Elsewhere, `statCalc.totalCount(profile.items.mounts)` is used for counting all mounts.

PR #6046, opened 2015-10-14
Handle new possible iOS in-app-purchases

phillipthelen:
This adds handling for the new iOS IAP tiers. also changes the buyGems function to handle different amounts of gems.

### DIFF
--- a/website/src/controllers/payments/index.js
+++ b/website/src/controllers/payments/index.js
@@ -137,7 +137,7 @@ exports.cancelSubscription = function(data, cb) {
 }
 
 exports.buyGems = function(data, cb) {
-  var amt = data.amount ? data.amount : 5;
+  var amt = data.amount || 5;
   amt = data.gift ? data.gift.gems.amount/4 : amt;
   (data.gift ? data.gift.member : data.user).balance += amt;
   data.user.purchased.txnCount++;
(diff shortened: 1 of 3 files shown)


crookedneighbor:
@vIiRuS Will this be compatible with the current ios app? And the old mobile app?

crookedneighbor:
Actually, the old mobile app didn't do gem purchases, did it?

negue:
Only the old android app has gem purchases

phillipthelen:
As it also has the old identifier for 20 gems (but awards 21) it is also compatible with the old/current ios version

website/src/controllers/payments/index.js
```diff
@@ -137,7 +137,8 @@ exports.cancelSubscription = function(data, cb) {
 }
 
 exports.buyGems = function(data, cb) {
-  var amt = data.gift ? data.gift.gems.amount/4 : 5;
+  var amt = data.amount ? data.amount : 5;
```
crookedneighbor: This should probably be: `var amt = data.amount || 5;`

crookedneighbor:
One style comment, otherwise, :+1:

phillipthelen:
Wasn't aware that you could do that. Changed it :)

PR #6049, opened 2015-10-15
Updated vagrant provisioning script to update npm to the latest version

nivl4:
Updated vagrant provisioning script to update npm to the latest version as the older version (1.4.28) was causing the following error during npm install.

npm ERR! EEXIST, open '/home/vagrant/.npm/9fb4c098-adable-stream-1-0-33-package-tgz.lock'
File exists: /home/vagrant/.npm/9fb4c098-adable-stream-1-0-33-package-tgz.lock
Move it away, and try again.

npm ERR! System Linux 3.2.0-23-generic
npm ERR! command "/usr/bin/node" "/usr/bin/npm" "install" "--no-bin-links"
npm ERR! cwd /vagrant
npm ERR! node -v v0.10.37
npm ERR! npm -v 1.4.28
npm ERR! path /home/vagrant/.npm/9fb4c098-adable-stream-1-0-33-package-tgz.lock
npm ERR! code EEXIST
npm ERR! errno 47
npm ERR! not ok code 0

### DIFF
--- a/vagrant.sh
+++ b/vagrant.sh
@@ -68,6 +68,8 @@ echo Updating repositories...
 apt-get update -qq
 echo Installing node.js
 apt-get install -qq nodejs
+echo Updating npm...
+npm install -g npm
 
 cd /vagrant
 


Alys:
At a guess, this is unlikely to be a permanent fix for the problem of a lock file existing, unless the latest version of npm handles lock files differently. I suspect that the problem you had with that particular lock file was fixed because the upgrade process probably deletes old lock files. If you do run into "npm ERR! EEXIST" for lock files in future, try deleting everything in /home/vagrant/.npm/

However [it's recommended](https://docs.npmjs.com/getting-started/installing-node) that npm be upgraded after the nodejs package has been installed, so this change is definitely something we want. Thank you @nivl4. I've awarded you a contributor tier for this.

nivl4:
@Alys thanks for the contributor title! :) I'm now a blacksmith. yay! About the fix, I think I did run "rm -rf ~/.npm" and "rm -rf ~/node_modules" but it was still throwing the same error when I tried. Then I googled a bit, it seems it might be a npm bug according to this: https://github.com/Shippable/support/issues/713 Not sure if it's a solution, but upgrading npm seem to magically fix this problem.

Alys:
Oh that's very interesting. It sounds like the latest version does handle lock files differently, which is excellent news! Thank you!

PR #6057, opened 2015-10-16
Refactor API unit tests

hartalex:
Move GET /user tests from status.coffee to users.coffee.
Remove GET /status test that uses user id and token.

18be44b2-acf0-47a8-a7af-1dcde2ff09a2


crookedneighbor:

Issue #3621, opened 2014-06-17
/user/tags API is not RESTful


kusold:
POST /user/tags should return the newly created Tag object, not the list of all the tags.

There should be two more routes:
GET /user/tags/:id to only get the tag for the id
GET /user/tags to get a list of all the tags. Currently the POST /user/tags is taking this responsibility

The GETs could be sneaked into the API without any issues, but the change for POST would be breaking and better suited for v3.
##

lefnire:
TMK there's a few more bits that aren't restful. A v3 would be very useful. I'm going to be migrating user objects from tasks stored as arrays to stored as objects, and that will be on v3. I'll probably not document v3 until we get more stuff in there such as a restful API

kusold:
I'll try and add the ones that I find. I only discovered the tag ones because I'm building out a [npm package](https://www.npmjs.org/package/habitrpg-api) for the API while also trying to tackle ToDoist syncing.

me-and:
You can also do a GET `/user` to pick up a list of tags, which at least doesn't require POSTing if you want  the tag list without changing state anywhere.

librarianmage:
Is this finished?

Alys:
Closing this now that API v3 has been released.

PR #6066, opened 2015-10-19
Add GET routes for /user/tags and /user/tags/{id}

czan:
An operation that is useful for tools to be able to do is to list a user's tags without having to pull down the entire user object (which is likely significantly larger than the tags). This commit adds a route to GET /user/tags, to get a list of the user's tags, as well as a GET /user/tags/{id} to get the details of a particular tag.

This is relevant to issue #3621. No existing routes were changed, though, so the API is not completely restful.

### DIFF
--- a/common/script/index.coffee
+++ b/common/script/index.coffee
@@ -673,6 +673,14 @@ api.wrap = (user, main=true) ->
         user.tags.splice to, 0, user.tags.splice(from, 1)[0]
         cb? null, user.tags
 
+      getTags: (req, cb) ->
+        cb? null, user.tags
+
+      getTag: (req, cb) ->
+        tid = req.params.id
+        i = _.findIndex user.tags, {id: tid}
+        return cb?({code:404,message:i18n.t('messageTagNotFound', req.language)}) if !~i
+        cb? null, user.tags[i]
 
       updateTag: (req, cb) ->
         tid = req.params.id
--- a/test/api/user/GET-user_tags.test.js
+++ b/test/api/user/GET-user_tags.test.js
@@ -1,0 +1,20 @@
+import {
+  generateUser,
+  requester,
+} from '../../helpers/api.helper';
+
+describe('GET /user/tags', () => {
+  let api, user;
+
+  beforeEach(() => {
+    return generateUser().then((usr) => {
+      user = usr;
+      api = requester(usr);
+    });
+  });
+
+  it('gets the user\'s tags', () => {
+    return expect(api.get('/user/tags'))
+      .to.eventually.eql(user.tags);
+  });
+});
--- a/test/api/user/GET-user_tags_id.test.js
+++ b/test/api/user/GET-user_tags_id.test.js
@@ -1,0 +1,25 @@
+import {
+  generateUser,
+  requester,
+} from '../../helpers/api.helper';
+
+describe('GET /user/tags/id', () => {
+  let api, user;
+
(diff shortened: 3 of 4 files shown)


crookedneighbor:
This reasonable to me, but currently it's blocked by a [new API test infrastructure we're building](https://github.com/HabitRPG/habitrpg/pull/6062). As soon that is merged in, we can add tests for this behavior and merge this in.

crookedneighbor:
The PR is merged in. Let me know if you need help writing tests.

czan:
I've rebased my changes and added tests for the GET routes (I was considering adding tests for the other /user/tags routes too, but decided I'd be better off keeping this pull request focused on the GET ones).

test/api/user/GET_user_tags_id.test.js
```diff
@@ -0,0 +1,41 @@
...
+    }).catch(done);
+  });
+
+  it('gets a user\'s tag by id', (done) => {
+    api.get('/user/tags/' + user.tags[0].id)
+      .then((fetchedTag) => {
+        expect(fetchedTag).to.eql(user.tags[0]);
+        done();
+      })
+      .catch((err) => {
```
crookedneighbor: This can be simplified to:

``` js
.catch(done);
```

test/api/user/GET_user_tags_id.test.js
```diff
@@ -0,0 +1,41 @@
...
+  it('fails for non-existent tags', (done) => {
+    api.get('/user/tags/not-an-id')
+      .then((fetched) => {
+        done(new Error('Request should have failed'));
+      })
+      .catch((err) => {
+        expect(err.code).to.eql(404);
+        done();
+      })
+      .catch((err) => {
```
crookedneighbor: This second catch block shouldn't be necessary.

crookedneighbor:
Overall, looks really good. Just a few things to clean up and it should be ready to merge.

test/api/user/GET_user_tags_id.test.js
czan: I think we need both of these, actually. The first one checks that the error code is correct (because the `api.get(...)` call should return a promise which errors), then the second one catches the error from the `expect(...)` line (if there is one) and passes it to `done`.

czan:
Okay, I've replaced those unnecessary functions wrapping `done` and combined all my commits into the one concise commit for merging. Let me know if there are any more changes you want me to make!

test/api/user/GET_user_tags_id.test.js
czan: I've thought more about this now, and I think what I have now is actually not the right way to do this. The `.then` step actually needs both a success and error handler at that point, not a separate `.then` followed by a `.catch`. I'll fix that up now.

czan:
Assuming the tests pass, this most recent commit should be ready to merge now. I've converted the tests to use promises instead of the `done` callback.

crookedneighbor:
The tests look great. Could you change their names to `GET-user_tags...` instead of `GET_user_tags...`, just to be consistent with the other tests?

czan:
Done! Sorry about that. I hadn't noticed the difference, but it's fixed up now.

crookedneighbor:

czan:
Cheers!

User id: 3e3ef99b-3ee3-4114-a779-78332cb6e714

crookedneighbor:

PR #6075, opened 2015-10-21
Update EXTENDEDCHANGELOG.md

claushellsing:
small typos fix

### DIFF
--- a/EXTENDEDCHANGELOG.md
+++ b/EXTENDEDCHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


Alys:
This file contains the comments from commits and so it should keep those comments exactly as specified in the commits, even when there's typos, for consistency with the repository. For the commit comment that you edited, you can see the original at  https://github.com/HabitRPG/habitrpg/commit/d3f4a561fdf137e5d8f406bae03be4fef1caff22

Also I think it's likely that this file was once updated by an automatic process, and so probably shouldn't have manual edits done to it. We don't seem to be using it any more, regardless.

PR #6082, opened 2015-10-22
fix a minor syntax error when opening reroll modal.

anilanar:
This is a very minor syntax error that Angular rightfully complained about when clicked on fortify potion.


crookedneighbor:
Thanks for catching this!

anilanar:
@crookedneighbor Thanks!
b55a4bfe-b048-4e4c-b086-aa58145b56e1

crookedneighbor:

Issue #6043, opened 2015-10-14
Subscriber Mystery Items can be given to users after their subscription has terminated


Alys:
It's possible for Mystery Items to keep on being given to a user after their subscription has ended, if the user has stopped using Habitica.

Details:

When a player terminates a subscription, the `purchased.plan.dateTerminated` field is set to the appropriate future date. Every time cron runs, it checks that field and if the termination date has arrived, it deletes the user's subscription details (but does not delete any Mystery Items that the user has not yet claimed). 

This means that subscription termination depends on cron running, so if a user stops using Habitica when they unsubscribe, their subscription remains active until/if they do log in again.

The script that assigns Mystery Items to subscribers (`migrations/mystery_items.js`), checks only for the existence of a subscription plan. It doesn't check the termination date. Therefore, users who unsubscribed and stopped logging in keep on being given the Mystery Items. If they do eventually log in, their subscription will be terminated immediately but all the accumulated, unclaimed Mystery Items will still be available for them to claim.

I don't think there's any other exploits related to the subscription termination mechanism since almost all of the subscriber benefits are related to users actually using Habitica (e.g., drop cap, gem buying). It's possible that history entries will avoid being purged for a longer period than the user has paid for, but I consider that a very minor exploit, if it happens - so minor, I can't be bothered investigating properly. ;) 

So I think the best way to fix this is to adjust `migrations/mystery_items.js` so that it checks termination date in addition to checking for the existence of a plan (and a new database index might then be needed). I don't think it's appropriate to change the subscription termination mechanism because it seems correct in every other case. I also don't think that cron should delete accumulated Mystery Items because that would sometimes delete items that the user has paid for. Opposing opinions are welcome. :)
##

TheHollidayInn:
I believe we can close this now since the solution was merged.

Pichu0102:
Hit by this bug myself. Username is Pichu0102.

ghost:
^Pichu0102's UUID is 0d91f7d6-588d-4d54-822f-8bf8c4b2141c . My fault, I wasn't clear when I said to comment.

Alys:
@Pichu0102 I've removed the excess Mystery Items from your account. Thank you for letting us know about this!

@Cantras Thanks!

PR #6106, opened 2015-10-25
Changed Mongo query to check DateTerminated to prevent accumulation o…

kylestratis:
…f mystery items when subscription is ended.

This fixes issue #6043

### DIFF
--- a/migrations/mystery_items.js
+++ b/migrations/mystery_items.js
@@ -27,6 +27,6 @@ if (_id) {
   db.users.update({_id:_id}, update);
 } else {
   // multiple (once @ start of event)
-  db.users.update({'purchased.plan.customerId':{$ne:null}}, update, {multi:true});
+  db.users.update({$and: [{'purchased.plan.customerId':{$ne:null}},{$or: [{'purchased.plan.dateTerminated':{$gte:new Date()}},{'purchased.plan.dateTerminated':{$eq:null}}]}]}, update, {multi:true});
 }
 
--- a/migrations/mystery_items.js
+++ b/migrations/mystery_items.js
@@ -27,6 +27,9 @@ if (_id) {
   db.users.update({_id:_id}, update);
 } else {
   // multiple (once @ start of event)
-  db.users.update({$and: [{'purchased.plan.customerId':{$ne:null}},{$or: [{'purchased.plan.dateTerminated':{$gte:new Date()}},{'purchased.plan.dateTerminated':{$eq:null}}]}]}, update, {multi:true});
+  db.users.update([{$and: [{'purchased.plan.customerId':{$ne:null}},
+                           {$or: [{'purchased.plan.dateTerminated':{$gte:new Date()}},
+                                  {'purchased.plan.dateTerminated':{$exists:false}},
+                                  {'purchased.plan.dateTerminated':{$eq:null}}]}]}], update, {multi:true});
+  });
 }
-
--- a/migrations/mystery_items.js
+++ b/migrations/mystery_items.js
@@ -27,9 +27,14 @@ if (_id) {
   db.users.update({_id:_id}, update);
 } else {
   // multiple (once @ start of event)
-  db.users.update([{$and: [{'purchased.plan.customerId':{$ne:null}},
-                           {$or: [{'purchased.plan.dateTerminated':{$gte:new Date()}},
-                                  {'purchased.plan.dateTerminated':{$exists:false}},
-                                  {'purchased.plan.dateTerminated':{$eq:null}}]}]}], update, {multi:true});
-  });
+  db.users.update([
+    {$and: [
+      { 'purchased.plan.customerId': { $ne:null } },
+      {$or: [
+        { 'purchased.plan.dateTerminated': { $gte: new Date() } },
+        { 'purchased.plan.dateTerminated': { $exists:false } },
+        { 'purchased.plan.dateTerminated': { $eq:null } }
+      ]}
+    ]}
+  ], update, { multi:true });
 }
--- a/migrations/mystery_items.js
+++ b/migrations/mystery_items.js
@@ -1,8 +1,8 @@
 var _id = '';
 var update = {
   $addToSet: {
-    'purchased.plan.mysteryItems':{
-      $each:['head_mystery_201509','armor_mystery_201509']
+    'purchased.plan.mysteryItems': {
+      $each: ['head_mystery_201509', 'armor_mystery_201509']
     }
   }
 };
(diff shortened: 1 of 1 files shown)


crookedneighbor:
Pinging @alys

Alys:
Yes, I saw it. :)

Thank you @kylestratis! I'll check this out tonight if I can, but more likely over the next few days.

For this month's Mystery Items, we'll keep using the current script.

kylestratis:
No problem @Alys, always happy to help where I can! I look forward to working on more issues soon!

Alys:
@kylestratis Sorry about the delay.

Could you please also add an $exists check for the case where the user's document does not contain the `purchased.plan.dateTerminated` field?

kylestratis:
@Alys Done. I looked for some examples in the code and tested my solution before pushing and it appears to work, but I'm new to Mongo so let me know if anything needs to be fixed up!

migrations/mystery_items.js
```diff
@@ -27,6 +27,8 @@ if (_id) {
   db.users.update({_id:_id}, update);
 } else {
   // multiple (once @ start of event)
-  db.users.update({'purchased.plan.customerId':{$ne:null}}, update, {multi:true});
+  db.users.find({'purchased.plan.dateTerminated':{$exists:true}}).forEach(function(user) {
```
crookedneighbor: Am I remembering correctly that there is a find and update method?

crookedneighbor: Could this be helpful? https://docs.mongodb.org/manual/reference/method/db.collection.findAndModify/

kylestratis: @crookedneighbor I did not see that at first (my google-fu must be weak today!), thanks! Upon further investigation, though, it seems as if findAndModify will only modify a single document, regardless of how many the find operation returns. Considering the use of the multi option in the old update statement, would it be safe to guess that we need the ability to modify multiple docs at once?

Alys: Yes, multiple docs at once is still needed.

However, I'm sorry, I wasn't clear before when I spoke about the $exists check for the case where the user's document does not contain the purchased.plan.dateTerminated field.

We need to find any user that has a value for 'purchased.plan.customerId' (i.e., they've bought a subscription at some time in the past) and whose subscription has not terminated. To determine if it has not terminated, you need to find any user where any one of the following is true:
- the purchased.plan.dateTerminated field does not exist
- the purchased.plan.dateTerminated field has a null value
- the purchased.plan.dateTerminated field has a date in the future

So the three dateTerminated checks need to be in the same $or clause. Let me know if that doesn't make sense! :)

crookedneighbor: @kylestratis You're right regarding `findAndModify`. Any progress on @Alys's instructions?

kylestratis:
@crookedneighbor @Alys 
Not yet, unfortunately, I've been pulling lots of late nights at work. I hope to take a deeper look at it tonight after work, hopefully I'll have the code updated, tested, and committed tonight or tomorrow.

crookedneighbor:
No worries, just wanted to make sure you weren't waiting on us :smile:

kylestratis:
@crookedneighbor @Alys 
I have made the requested changes. I removed the $and clause and instead found all users where customerID was not null (this check was done in the $and clause, which I realized wasn't necessary), then added the $exists:null check to the or clause in the update. If this was done incorrectly, please let me know!

migrations/mystery_items.js
```diff
@@ -27,6 +27,7 @@ if (_id) {
   db.users.update({_id:_id}, update);
 } else {
   // multiple (once @ start of event)
-  db.users.update({'purchased.plan.customerId':{$ne:null}}, update, {multi:true});
-}
-
+  db.users.find({'purchased.plan.customerId':{$ne:null}}).forEach(function(user) {
```
crookedneighbor: Is the find and forEach necessary? Can the update call not be done all at once? Is there a performative difference?

kylestratis: Looking at the docs again, I don't think it is necessary, and I can't recall why I added it in the first place. I'll make the change shortly - I think I will just move the customerId check back into the update query in an $and clause.

kylestratis:
Moved $and clause back into update query, tried to make the query slightly more readable, since it is fairly long. Thanks for the patience @Alys and @crookedneighbor!

migrations/mystery_items.js
```diff
@@ -27,6 +27,9 @@ if (_id) {
   db.users.update({_id:_id}, update);
 } else {
   // multiple (once @ start of event)
-  db.users.update({'purchased.plan.customerId':{$ne:null}}, update, {multi:true});
-}
-
+  db.users.update([{$and: [{'purchased.plan.customerId':{$ne:null}},
```
crookedneighbor: If @alys and @paglias can sign off on the query, the only thing I'd change is formatting:

``` js
db.users.update([
  { $and: [
    { 'purchased.plan.customerId': { $ne: null } },
    { $or: [
      { 'purchased.plan.dateTerminated': { $gte: new Date() }},
      { 'purchased.plan.dateTerminated': { $exists: false } },
      { 'purchased.plan.dateTerminated': { $eq: null } }
    ]}
  ]}
], update, { multi: true });
```

kylestratis: @crookedneighbor would you prefer the longer single line style?

crookedneighbor: No. The code sample I posted should be identical in content to your code, it just handles spacing and line breaks differently to make it more readable.

kylestratis: I see, I misread the example as something else. I've pushed my changes.

migrations/mystery_items.js
```diff
@@ -27,6 +27,14 @@ if (_id) {
   db.users.update({_id:_id}, update);
 } else {
   // multiple (once @ start of event)
-  db.users.update({'purchased.plan.customerId':{$ne:null}}, update, {multi:true});
-}
-
+  db.users.update([
+    {$and: [
+      { 'purchased.plan.customerId': { $ne:null } },
```
crookedneighbor: Nitpick here, but could you put a space between the `:` and the value for each of these?

kylestratis: Not a problem, I thought I did it for them all, but I definitely missed them. I did the same elsewhere in the file as well for consistency's sake.

SabreCat:
Note for reviewers: make sure to check out competing PR #6292.

SabreCat:
Hey @kylestratis, can you check on the merge conflict here? (It's probably just that another month of subscriber items has gone by in the meantime - sorry about that!) Let us know if you need a hand with it.

TheHollidayInn:
Eh, didn't me to duplicate efforts here. I did a PR without seeing this. I believe $and is implied, so I don't think you need the wrapper.

as for the $or, what happens when a user resubs? Won't the dateTerminated field exist? If so, then these two conditions will prevent users who have been terminated, but have resubscribed from getting the items.
 { 'purchased.plan.dateTerminated': { $exists: false } },
  { 'purchased.plan.dateTerminated': { $eq: null } }

Alys:
@TheHollidayInn the dateTerminated field should not exist, or should have a falsy value, when the user has any subscription that doesn't have a termination date. 

@kylestratis I'm sorry about the delay on this. I meant to review it last weekend but completely forgot. I'll definitely do it this weekend if it doesn't get done before then! From a quick look at it, it does what we want, but I want to run some tests to make sure.

Alys:
I'll fix the merge conflicts for this when I run tests for it this weekend.

Alys:
I've tested this and it works as desired. It excludes about 1500 users, all of whom have subscription termination dates in the past, so that's a win! Thank you @kylestratis! Tier 2 awarded.

Merged in https://github.com/HabitRPG/habitrpg/commit/42d1fd93f3e90ab8648b47d789bff7f32b41eacc (with conflicts fixed and $and removed since it's the default option for multiple queries).

PR #6107, opened 2015-10-26
Create initial PUT /user test

jeremyabbott:
@crookedneighbor how is this for an initial test (test/api/user/PUT-user.test.js)? Thank you again for the help!

### DIFF
--- a/test/api/user/PUT-user.test.js
+++ b/test/api/user/PUT-user.test.js
@@ -1,0 +1,27 @@
+import {
+  generateUser,
+  requester,
+} from '../../helpers/api.helper';
+
+describe('PUT /user', () => {
+  let api, user;
+  
+  beforeEach((done) => {
+    return generateUser().then((usr) => {
+      user = usr;
+      api = requester(usr);
+      done();
+    });
+  });
+  
+  it('updates the user', () => {
+    let api = requester(user);
+    return api.get('/user')
+    .then((fetchedUser) => {  
+      return api.put('/user', {'profile.name' : 'Frodo'});
+    })
+    .then((updatedUser) => {
+      expect(updatedUser.profile.name).to.eql('Frodo');
+    });
+  });
+});
--- a/test/api/user/DELETE-user.test.js
+++ b/test/api/user/DELETE-user.test.js
@@ -9,7 +9,7 @@ describe('DELETE /user', () => {
   beforeEach(() => {
     return generateUser().then((user) => {
       api = requester(user);
-    })
+    });
   });
 
   it('deletes the user', () => {


test/api/user/PUT-user.test.js
```diff
@@ -0,0 +1,27 @@
+import {
+  generateUser,
+  requester,
+} from '../../helpers/api.helper';
+
+describe('PUT /user', () => {
+  let api, user;
+  
+  beforeEach((done) => {
+    generateUser().then((usr) => {
```
crookedneighbor: the beforeEach block can return a promise as well:

``` js
let user;

beforeEach(() => {
  return generateUser().then((usr) => {
    user = usr;
  });
});
```

test/api/user/PUT-user.test.js
```diff
@@ -0,0 +1,27 @@
...
+      user = usr;
+      api = requester(usr);
+      done();
+    }).catch(done);
+  });
+  
+  it("updates the user", () => {
```
crookedneighbor: We try to use single quotes in our javascript whenever possible.

jeremyabbott:
@crookedneighbor updated with feedback. Thank you for your patience too! I really appreciate it.

crookedneighbor:
I made some adjustments and added some more tests. Thanks for getting this started. https://github.com/HabitRPG/habitrpg/blob/develop/test/api/user/PUT-user.test.js

jeremyabbott:
Wow! Thanks for the feedback. It's helping me grok the idea of protected paths concept. Is it cool if I keep working on tests?

My user ID is: f2fdbe8a-25d9-4c38-933f-7a6ff5885062

crookedneighbor:

Feel free to keep adding tests if you'd like. The dream is to move over all the tests from the legacy api tests so we can remove that directory entirely and provide coverage for every route. 

The only ones that you shouldn't work on are the ones related to tasks, as I'll be doing that as part of a big tasks collection PR.

jeremyabbott:
Thank you!

I started looking at the legacy tests this morning. I'll start replacing
those. Thank you again.

On Mon, Oct 26, 2015 at 9:40 PM, Blade Barringer notifications@github.com
wrote:

> Tier awarded! You can read more about contributor tiers and rewards here
> http://habitrpg.wikia.com/wiki/Contributor_Rewards.
> 
> Note that as your tier increases, it takes progressively more work to
> achieve the next one, so for the higher tiers, larger PRs or a few PRs are
> required. There's some information about that at Contributor Rewards, but
> feel free to ask here if you have questions!
> 
> Feel free to keep adding tests if you'd like. The dream is to move over
> all the tests from the legacy api tests so we can remove that directory
> entirely and provide coverage for every route.
> 
> The only ones that you shouldn't work on are the ones related to tasks, as
> I'll be doing that as part of a big tasks collection PR.
> 
> —
> https://github.com/HabitRPG/habitrpg/pull/6107#issuecomment-151351141.

## 

Thanks,
Jeremy

---

Q: Why is this email three sentences or less?
A: http://three.sentenc.es

PR #6117, opened 2015-10-27
Fix a bug that would allow users to add Mystic Hourglass through API.

nivl4:
Fix a bug that would allow users to add Mystic Hourglass through API.

See the similar pull request: https://github.com/HabitRPG/habitrpg/pull/6115

Users would be able to add gems through the following API request:
POST: /user/batch-update
DATA: [{"op": "addHourglass"}]

*This is a quick fix, a better thought out design for batch-update might needed.
As I don't understand the API fully now, there might other dev backdoors.


Alys:
Bug fix has been applied in https://github.com/HabitRPG/habitrpg/commit/a29c4127811aea1ba5a488cd134d95fe37f6e23f since this PR contains some irrelevant code.

This will count towards a contributor tier. Please tell us your UUID.

nivl4:
UUID: 51102111-ed4a-4f13-972b-705a3c1e5808

Alys:
Tier awarded for #6115 and #6117. You can read more about [contributor tiers and rewards here](http://habitrpg.wikia.com/wiki/Contributor_Rewards).

PR #6122, opened 2015-10-27
Batch update adjustment

crookedneighbor:
Gives sane error if op doesn't exist.

### DIFF
--- a/website/src/controllers/user.js
+++ b/website/src/controllers/user.js
@@ -606,6 +606,7 @@ api.batchUpdate = function(req, res, next) {
           return cb(code+": "+ (data.message ? data.message : data.err ? data.err : JSON.stringify(data)));
         return cb();
       };
+      if(!api[_req.op]) { return cb(_req.op + ' operation not found'); }
       api[_req.op](_req, res, cb);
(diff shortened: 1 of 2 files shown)


website/src/controllers/user.js
```diff
@@ -606,6 +600,7 @@ api.batchUpdate = function(req, res, next) {
           return cb(code+": "+ (data.message ? data.message : data.err ? data.err : JSON.stringify(data)));
         return cb();
       };
+      if(!api[_req.op]) { return cb(_req.op + ' operation not found'); }
```
crookedneighbor, line 603: @paglias Does this look right to you?

paglias, line 603: @crookedneighbor it's fine but due to this line https://github.com/crookedneighbor/habitrpg/blob/batch_update_adjustment/website/src/controllers/user.js#L617 any error is treated as an internal server error. But I think that passing a different error code to the response woul break the site so don't worry, we'll fix it later

crookedneighbor, line 603: Yeah, I noticed that. Made me :cry:

PR #6126, opened 2015-10-28
Docs project

batorni:


Alys:
Thank you for this, but can you please edit the description to provide a summary of the changes this PR makes?

crookedneighbor:
Looks like all these commits are from a different user from September 2013. I assume this PR is a mistake, but feel free to re-open it if that is not the case.

batorni:
Shoot! My bad. Must've accidentally submitted that PR while I was looking through the project.

PR #6150, opened 2015-10-31
Implements user in solo group tests

jeremyabbott:
Implements tests for user in solo group.

### DIFF
--- a/test/api/user/DELETE-user.test.js
+++ b/test/api/user/DELETE-user.test.js
@@ -1,15 +1,18 @@
 import {
+  checkExistence,
+  generateGroup,
   generateUser,
   requester,
   translate as t,
 } from '../../helpers/api.helper';
 
 describe('DELETE /user', () => {
-  let api;
+  let api, user;
 
   beforeEach(() => {
-    return generateUser().then((user) => {
-      api = requester(user);
+    return generateUser().then((usr) => {
+      api = requester(usr);
+      user = usr;
     });
   });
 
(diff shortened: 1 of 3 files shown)


test/api/user/DELETE-user.test.js
```diff
@@ -1,15 +1,21 @@
 import {
+  checkExistence,
   generateUser,
   requester,
   translate as t,
 } from '../../helpers/api.helper';
 
+import {
+  generateGroup
```
crookedneighbor: This can be added in the object above.

test/api/user/DELETE-user.test.js
```diff
@@ -27,10 +33,30 @@ describe('DELETE /user', () => {
   });
 
   context('user in solo group', () => {
-
-    it('deletes party when user is the only member');
-
-    it('deletes private guild when user is the only member');
+    let group;
+    it('deletes party when user is the only member', () => {
+      return expect(generateGroup(user, {
```
crookedneighbor: I think it makes sense to pull this out into a before block, since it's less about testing the route and more about setting up to test the route. 

In that case, it probably makes sense to create two contexts, one for the guild and one for the party.

jeremyabbott:
@crookedneighbor thanks again. I made the changes. I need to squash again, but I wanted to see something first. Basically "beforeEach" fails locally in nested contexts, but "before" does not. I'm not sure why. I wanted to see if it fails in Travis. I've seen it fail in other tests that I haven't updated.

jeremyabbott:
@crookedneighbor so the new tests fail with before but not beforeEach. Very strange though.

crookedneighbor:
Tier awarded!

Made a few alterations for style: https://github.com/HabitRPG/habitrpg/commit/da06e0736a94b0e83550161e38d2c21b1d09df7e?w=1

PR #6151, opened 2015-10-31
added some comments

kyl3r92:
I added some comments, Might try improving code in future, but since I'm new to Github (came here for Hacktoberfest) I thought I'd start with a small pull request.

please review

### DIFF
--- a/migrations/20130307_remove_duff_histories.js
+++ b/migrations/20130307_remove_duff_histories.js
@@ -11,7 +11,11 @@ db.users.find().forEach(function(user){
             task.history = _.filter(task.history, function(h){return !!h.value})
         }
     });
-
+    
+    
+    /**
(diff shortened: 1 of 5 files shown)


crookedneighbor:
Thanks for your interest in the project, but I don't think these changes are necessary. 

First, many of the files you changed are old migration files which will not be used again. 

Second, many of the comments look unnecessary because the code is written in such a way that, with minimal context, it should be obvious what it is doing.

kyl3r92:
Thank you for the reply, I agree, still I am happy to have created a pull request. I now understand how working together on open-source projects works. I used Github with Eclise / Subversion, it was very simple. 
So thanks for the reply, I might inspect the code further and try to contribute helpful pull requests

crookedneighbor:
Please do :smile:

PR #6152, opened 2015-10-31
Prevent cron running when a user's timezone changes

Alys:
**DO NOT MERGE - MORE TESTING IS BEING DONE**

@blade @paglias It would be wise to have extra eyes on this.

@lemoness I'm seeking approval from you for new behaviour. I believe we should deploy this to production before 1 November USA time.

_tl;dr  This will (I think) prevent second crons due to timezone or daylight savings changes. But it could also sometimes prevent a legitimate first cron on a day when the zone changes._

After a user's timezone has changed (either because of travel or daylight savings changes), cron can run even if it has already run once that day (e.g., if the timezone change moves the user back to before the custom day start time). The second cron can cause great damage because most Dailies are not completed.

This PR adds a check at the start of cron to compare the user's previous timezone (which is stored in `user.preferences.timezoneOffset`) to the current timezone (which is captured by Habitica when the user logs in and which is then passed through HTTP headers to cron). If the two timezones are different, cron does not run, and it updates `user.preferences.timezoneOffset`

I believe this PR will prevent players and their party mates taking unfair damage from all timezone changes.

The disadvantage is that it will prevent legitimate crons from running in situations where cron has not already run on the day when the timezone change is noticed. However that will be less distressing for many users than having unfair damage. The worst part will be that their Dailies will not be unticked for reuse. We'll need to tell the Socialites to expect questions about that (the solution is to open up Fix Character Values in one window to capture all your stats, then untick all your Dailies, then save the FCV form to reapply your correct stats).

One catch with timezone changes is that Habitica notices the change only when the user reloads or logs in using the login form (not when they merely use Habitica or sync). This means that a user's PC might have its timezone changed today, but if the user doesn't log out of Habitica or reload until tomorrow, then the missed cron won't happen until tomorrow.

Improvements for the future:
- Find some way to identify if cron has already run that day and only prevent a new cron if it has. Maybe it's enough to just compare the previous cron time with the current time and if they are close, then assume the second cron is not legitimate. However I'm too brain-warped by timezone maths to be certain of that yet. The requirement to reload or log in might prevent this being reliable.
- If that can't be done, then when the timezone changes, allow some parts of cron to run but not all (e.g., untick Dailies but do not assign damage from them). We'd need to think carefully about exactly what behaviour we wanted.

### DIFF
--- a/common/script/index.coffee
+++ b/common/script/index.coffee
@@ -1697,6 +1697,18 @@ api.wrap = (user, main=true) ->
     cron: (options={}) ->
       now = +options.now || +new Date
 
+      # If the user's timezone has changed (due to travel or daylight savings),
+      # cron can be triggered twice in one day. So if a zone change is detected,
+      # we ASSUME cron has already run today and prevent a second one.
+      # FIXME: This can probably also prevent the day's first cron under some
+      # circumstances (but that's less serious than damage from an extra cron).
+      userTimezoneOffset = +user.preferences.timezoneOffset || 0
+      browserTimezoneOffset = +options.timezoneOffset || userTimezoneOffset
+      if userTimezoneOffset != browserTimezoneOffset
+        user.lastCron = now
+        user.preferences.timezoneOffset = browserTimezoneOffset
+        return
+
       daysMissed = api.daysSince user.lastCron, _.defaults({now}, user.preferences)
       return unless daysMissed > 0
 
--- a/common/script/index.coffee
+++ b/common/script/index.coffee
@@ -47,11 +47,7 @@ api.planGemLimits =
 sanitizeOptions = (o) ->
   dayStart = if (!_.isNaN(+o.dayStart) and 0 <= +o.dayStart <= 24) then +o.dayStart else 0
   timezoneOffset = if o.timezoneOffset then +(o.timezoneOffset) else +moment().zone()
-  # console.log("     - sanitizeOptions -")
-  # console.log {timezoneOffset}
-  # console.log {oNow:moment(o.now).format('YYYY-MM-DD HH:mm:ss')}
   now = if o.now then moment(o.now).zone(timezoneOffset) else moment(+new Date).zone(timezoneOffset)
-  console.log {nowWithOffset:moment(now).format('YYYY-MM-DD HH:mm:ss')}
   # return a new object, we don't want to add "now" to user object
   {dayStart, timezoneOffset, now}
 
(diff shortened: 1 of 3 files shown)


crookedneighbor:
I'm concerned about API only users (mobile app only). How does this work with that?

paglias:
@Alys in case of a 1 hour timezone change wouldn't it be better to check for a 25 hours difference between now and lastCron to decide whether run cron or not? In case of negative changes of 1 hour, the check would wait for 23 hours and so on..

Alys:
"I'm concerned about API only users (mobile app only). How does this work with that?"

In the worst case, they'll have the second, damaging cron, just as if this PR didn't exist. I've put in defaults for the two offset variables in the cron code, so that should cover any cases where either offset can't be read. Let me know if I"m wrong. :)

Alys:
"in case of a 1 hour timezone change wouldn't it be better to check for a 25 hours difference between now and lastCron to decide whether run cron or not? In case of negative changes of 1 hour, the check would wait for 23 hours and so on.."

I think it depends on when the user cronned. If they were up late at night and cronned shortly before their CDS and then shortly after that the timezone changed, it's a different situation than if they cronned many hours before the change. I don't think that the difference from lastCron is reliable enough.

crookedneighbor:
This looks sane to me, but I won't be able to give it as thorough review a review as I'd like.

Alys:
Actually I am going to make a change to allow some parts of cron to run. It will be easy for some of them.

Alys:
DO NOT MERGE! I've pushed my latest change in case anyone wants to review it but I'm still testing, and @lemoness might want to comment on this.

The first few sections of cron will now run when a timezone change is detected (e.g., subscription processing will occur, which is important if this was actually the first cron of the day).

This PR now makes a timezone-change cron behave exactly as if the user was resting in the inn, but with the exception that buffs are not cleared because users would find that annoying if it was the second cron of the day. (And of course the user doesn't have to check out of the inn afterwards.)

Dailies and their checklists WILL become unticked when the timezone-change cron runs, just in case it was the first cron of the day. I'm in two minds about whether that's a good thing though, because it will be annoying if it was the second cron. I can change that part back again if anyone says so.

Alys:
New thought..... I think this can be even more accurate.

Previously, when the user changed their timezone (or DST), the cron code didn't have access to the old timezone. It had to calculate whether cron should run based on the new timezone. However now it does have access to the old timezone.

So I'm thinking that when the cron code detects that the timezone has changed, it should use the OLD timezone to work out if cron should run. It then sets `user.preferences.timezoneOffset` to the new timezone value. If cron should run, it runs as normal. If cron should not run, it sets `lastCron` to the current moment to prevent an accidental cron from the user's new time.

I'm going to go off and have a think about this over a cup of tea. If anyone has any opinions, please comment!

crookedneighbor:
This looks really good to me @Alys. I say we try it.

Alys:
Thank you, but I'm finding problems with it - or at least with my most recent comment. I'm still working on that idea, but it's likely that I'll go with the option where cron behaves as if you're resting in the inn. Regardless, it won't be worse than it would be without this PR. :)

Alys:
Needs more work.

common/script/index.coffee
```diff
@@ -1695,10 +1695,93 @@ api.wrap = (user, main=true) ->
...
+        # No timezone change; normal cron processing.
+        daysMissed = daysMissedOldZone
+        # console.log("\nsame zones:")
+      else
+        # console.log("\ndifferent zones:")
+        # console.log {userTimezoneOffset}
+
+        endOfDay = api.startOfDay(_.defaults {now}, user.preferences)
+        endOfDay = moment(endOfDay).add({days:1})
+        endOfDay = endOfDay.subtract({minutes:1})
```
Alys, line 1732: This endOfDay value needs to be assigned to lastCron under some circumstances to prevent any chance of a cron running before the user's next CDS. But I can't get that darned value right. I'm sure it's a simple fix that will make me want to scream when I see it.

Alys:
Closing until I've done some more work on this.

PR #6171, opened 2015-11-03
Flag messages in api

crookedneighbor:
Assist for https://github.com/HabitRPG/habitrpg/pull/6065

I resisted the urge to refactor the whole groups controller.
# Problem

We hide flagged messages on the client, but not through the api. This requires the mods to delete the messages even after they are flagged. @srrvnn did a good job getting this started, but it needed tests.
# Solution

Write tests and extract code into a private function to reduce duplication.

@paglias @Alys


website/src/models/group.js
```diff
@@ -475,11 +475,6 @@ GroupSchema.methods.leave = function(user, keep, mainCb){
 
 GroupSchema.methods.toJSON = function() {
   var doc = this.toObject();
-  if(doc.chat){
```
crookedneighbor, line 478: I elected to do this at the api level instead of here. The reason being, if someone used the `GET /groups/id/chat` route, the flags object was still coming through.

paglias, line 478: strange that this didn't work...

paglias:
Seems good to me. Just one thing: hidden messages should be shown only to admins or also to mods?

Alys:
Hidden messages should be shown to anyone of contributor tier 8 and above (mods and staff).

Alys:
Hidden messages should be shown to anyone of contributor tier 8 and above (mods and staff) and we mark those users with `user.contributor.admin`

website/src/models/group.js
crookedneighbor, line 478: It worked for the regular `GET /groups/:id` route, but not for the `GET /groups/:id/chat`

PR #6176, opened 2015-11-04
Api v3 add babel

crookedneighbor:
WIP


.eslintrc
```diff
@@ -76,10 +76,11 @@
         "camelcase": 2,
         "comma-spacing": 2,
         "comma-style": [2, "last"],
+        "comma-dangle": [2, "always-multiline"],
```
crookedneighbor, line 79: I like having dangling commas, because it makes it easier to extend an object later, but I understand if I'm the only one who likes it and this can be removed.

.eslintrc
```diff
@@ -76,10 +76,11 @@
...
         "computed-property-spacing": [2, "never"],
         "consistent-this": [2, "self"],
         "func-names": 2,
-        "func-style": [2, "expression"],
+        "func-style": [2, "declaration", { "allowArrowFunctions": true }],
```
crookedneighbor, line 83: This allows us to use normal functions when exporting a module

```
export default function nameOfFunction () {
```

And arrow functions when declaring a variable:

```
let someFunction = () => {
```

In my opinion, that's how it should be.

.eslintrc
```diff
@@ -108,8 +109,7 @@
         "space-unary-ops": 2,
         "spaced-comment": [2, "always", { exceptions: ["-"]}],
         "padded-blocks": [2, "never"],
-        "no-multiple-empty-lines": [2, {max: 2}],
-        "lines-around-comment": [2, { "beforeBlockComment": true, "beforeLineComment": true, "allowBlockStart": true }]
```
crookedneighbor, line 112: This rule bugged me, so I took it out. If you feel strongly about it, let me know.

.eslintrc
paglias, line 79: :+1:

.eslintrc
paglias, line 83: :+1:

website/src/middlewares/api-v3/index.js
```diff
@@ -1,10 +1,8 @@
-'use strict';
-
 // This module is only used to attach middlewares to the express app
 
-let errorHandler = require('./errorHandler');
+import errorHandler from './errorHandler';
 
-module.exports = function attachMiddlewares (app) {
+export default function middleware (app) {
```
paglias: I would keep the attachMiddlewares name, this isn't a proper middleware but a function that takes the app and attaches several middlewares to it

crookedneighbor: Ahh, didn't realize it had a name before. When I first worked on it, I believe it was an anonymous function. Will change.

PR #6177, opened 2015-11-04
Add Client identifier and act on it during registration

phillipthelen:
Add a x-client header and use it to only create all sample tasks when registering from the website. Also set the tutorials to done (for the platform the user did not register on).

### DIFF
--- a/website/public/js/controllers/rootCtrl.js
+++ b/website/public/js/controllers/rootCtrl.js
@@ -36,6 +36,8 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
     $rootScope.toJson = angular.toJson;
     $rootScope.Payments = Payments;
 
+    $http.defaults.headers.common['x-client'] = "habitica-web";
+
     // Angular UI Router
     $rootScope.$state = $state;
     $rootScope.$stateParams = $stateParams;
--- a/website/src/models/user.js
+++ b/website/src/models/user.js
@@ -515,6 +515,34 @@ UserSchema.pre('save', function(next) {
         return newTask;
       });
     });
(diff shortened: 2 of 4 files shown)


website/src/models/user.js
```diff
@@ -509,6 +515,34 @@ UserSchema.pre('save', function(next) {
...
+      this.flags.tutorial.common.party = true;
+      this.flags.tutorial.common.pets = true;
+      this.flags.tutorial.common.gems = true;
+      this.flags.tutorial.common.skills = true;
+      this.flags.tutorial.common.classes = true;
+      this.flags.tutorial.common.tavern = true;
+      this.flags.tutorial.common.equipment = true;
+      this.flags.tutorial.common.items = true;
+    } else {
+      self.flags.showTour = false;
```
crookedneighbor, line 532: self or this?

website/public/js/controllers/rootCtrl.js
```diff
@@ -36,6 +36,8 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
     $rootScope.toJson = angular.toJson;
     $rootScope.Payments = Payments;
 
+    $http.defaults.headers.common['x-client'] = "habitica-web";
```
paglias, line 39: $http.defaults should be configured only once and then it'll be applied to all requests, it's not necessary to add it twice

phillipthelen, line 39: it's set twice, because rootCtrl isn't loaded yet before registration/login, but the header was needed during registration.

paglias, line 39: you're right, if it's not a big problem would you mind putting it in public/js/app.js like the other headers?

PR #6196, opened 2015-11-07
Add unit test helpers

crookedneighbor:
Some helpers for making unit tests easier to write.


test/helpers/api-unit.helper.js
```diff
@@ -0,0 +1,46 @@
...
+import i18n from '../../common/script/src/i18n';
+require('coffee-script');
+i18n.translations = require('../../website/src/libs/i18n.js').translations;
+
+afterEach(() => {
+  sandbox.restore();
+});
+
+export function generateUser(options={}) {
+  return new User(options)._doc;
```
crookedneighbor: Not sure how I feel about reaching into the private property `_doc` to generate a fake user. Seems perilous. I definitely wouldn't do this in production code.

For tests, if something about the implementation changed in the future, this would break all the tests, but this should be the only place it'd need to be adjusted to fix it.

paglias: what about using `toObject()`?

crookedneighbor: I thought I tried it and it didn't work like I expected. But I just tried it again and it does, so we'll go with that. 

```
¯\_(ツ)_/¯
```

PR #6198, opened 2015-11-07
V3 constants

crookedneighbor:
This lib is intended for any constants that require logic. 

IE, `IS_PROD` requires an assertion that `NODE_ENV` is equal to `'production'`. 

All other environmental variables should just use `nconf.get('name')`.


crookedneighbor:
@paglias Does this seem reasonable to you? Is there a better approach to take?

paglias:
Very reasonable. I planned to do something similar. The only thing is whether to add these constants as nconf variables so we don't have to import separately config and constants

crookedneighbor:
@paglias Are you thinking something like:

``` js
nconf.set('IS_PROD', process.env.NODE_ENV === 'production');
nconf.set('IS_DEV', process.env.NODE_ENV === 'development');
```

paglias:
Yeah, more or less, but I would use nconf's values not process.env's ones:

```
nconf.set('IS_PROD', nconf.get('NODE_ENV') === 'production');
nconf.set('IS_DEV', nconf.get('NODE_ENV') === 'development');
```

crookedneighbor:
Closing this in favor of https://github.com/HabitRPG/habitrpg/pull/6202

PR #6203, opened 2015-11-08
V3 - make linter happy

crookedneighbor:


website/src/models/user.js
```diff
@@ -643,27 +648,22 @@ schema.methods.unlink = function (options, cb) {
 
   let self = this;
 
-  switch (keep) {
```
crookedneighbor, line 646: The linter saw these as indentation errors.

In general, since none of these cases required us to fall through to another option, it's safer to use if/else if than a switch statement. Less likelihood of someone forgetting to put in a break statement.

website/src/models/user.js
```diff
@@ -581,12 +591,7 @@ schema.pre('save', function postSaveUser (next) {
   }
 
   if (!this.profile.name) {
-    let fb = this.auth.facebook;
-
-    this.profile.name =
-      (this.auth.local && this.auth.local.username) ||
-      (fb && (fb.displayName || fb.name || fb.username || `${fb.first_name && fb.first_name} ${fb.last_name}`)) ||
-      'Anonymous';
+    this.profile.name = _setProfileName(this);
```
crookedneighbor, line 594: Extracted this into a private method since the logic was so messy.

website/src/models/user.js
```diff
@@ -27,7 +27,7 @@ export let schema = new Schema({
     facebook: Schema.Types.Mixed, // TODO validate
     local: {
       email: String,
-      hashed_password: String,
+      hashed_password: String, // eslint-disable-line camelcase
```
crookedneighbor, line 30: For https://github.com/HabitRPG/habitrpg/issues/6193, we should convert the password key to hashedPassword to conform to the camelcase rule.

website/src/models/user.js
```diff
@@ -688,7 +688,6 @@ mongoose.model('User')
   .then((foundMods) => {
     // Using push to maintain the reference to mods
     mods.push(...foundMods);
-  })
-  .catch((err) => {
+  }, (err) => {
```
paglias: @crookedneighbor would you mind adding a TODO comment here about replacing this with a catch in future? The build is stuck but otherwise works well!

PR #6208, opened 2015-11-10
Flag adjustments

crookedneighbor:


website/src/controllers/api-v2/groups.js
```diff
@@ -259,11 +259,14 @@ api.attachGroup = function(req, res, next) {
...
 
+    var unchangedChat = _.cloneDeep(group.chat);
+
     if (!user.contributor.admin) {
-      _purgeFlagInfoFromChat(group);
+      _purgeFlagInfoFromChat(group, user);
     }
 
     res.locals.group = group;
+    res.locals.unchangedGroupChat = unchangedChat;
```
crookedneighbor: Not sure if this is the best way to do this, but it essentially keeps a record of the unchanged chats that can be applied to the flag and like calls 

@paglias @Alys

website/src/controllers/api-v2/groups.js
```diff
@@ -436,7 +442,8 @@ api.seenMessage = function(req,res,next){
 api.likeChatMessage = function(req, res, next) {
   var user = res.locals.user;
   var group = res.locals.group;
-  var message = _.find(group.chat, {id: req.params.mid});
+  var chat = res.locals.unchangedGroupChat;
+  var message = _.find(chat, {id: req.params.mid});
```
crookedneighbor: Is there a way to reach into an array on a mongo doc and just update something in it? I feel like that's how this should be handled instead of how I've implemented it.

 I know how to do it if it's an object, not sure how to do it with an array.

website/src/controllers/api-v2/groups.js
```diff
@@ -445,12 +452,14 @@ api.likeChatMessage = function(req, res, next) {
...
     message.likes[user._id] = true;
   }
+  group.chat = chat;
   group.markModified('chat');
   group.save(function(err,_saved){
     if (err) return next(err);
     // @TODO: We're sending back the entire array of chats back
     // Should we just send back the object of the single chat message?
     // If not, should we update the group chat when a chat is liked?
+    _purgeFlagInfoFromChat(_saved, user);
```
crookedneighbor: Since we send back the whole chat array, we have to re-purge it of flagged messages :(

crookedneighbor:
Closed with https://github.com/HabitRPG/habitrpg/commit/117a254cc8938f79dc68739abbfcc18b234044a5

PR #6211, opened 2015-11-11
Add analytics service to v3

crookedneighbor:
Ports v2 analytics service as middleware for v3.

### DIFF
--- a/website/src/middlewares/api-v3/analytics.js
+++ b/website/src/middlewares/api-v3/analytics.js
@@ -1,0 +1,27 @@
+import nconf from 'nconf';
+import {
+ track,
+ trackPurchase,
+ mockAnalyticsService,
+} from '../../libs/api-v3/analyticsService';
+
+function _generateService () {
+  let service;
+
+  if (nconf.get('IS_PROD')) {
+    service = {
+      track,
+      trackPurchase,
+    };
+  } else {
+    service = mockAnalyticsService;
+  }
+
+  return service;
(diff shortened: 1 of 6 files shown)


paglias:
There are a few small things I'm not convinced about, I'll comment once I get to a computer later but wait before merging

crookedneighbor:
Sure thing.

website/src/middlewares/api-v3/analytics.js
```diff
@@ -0,0 +1,27 @@
...
+    };
+  } else {
+    service = mockAnalyticsService;
+  }
+
+  return service;
+}
+
+export default function attachAnalytics (req, res, next) {
+  res.analytics = _generateService();
```
paglias: The _generateService() function should only run once, not on every request since it's always the same

crookedneighbor: I was thinking that too. This was mainly for being able to test prod and non-prod. We could memo-ize it.

crookedneighbor: Or we could just not care about testing prod vs non-prod for this.

paglias: I would say in this case better to prioritize speed

crookedneighbor:
Merged with https://github.com/HabitRPG/habitrpg/commit/cdb05e1b424f74be10c67a448e151a4944b6882d

PR #6214, opened 2015-11-11
Add tests for password utilities

paglias:
@crookedneighbor


website/src/libs/api-v3/password.js
```diff
@@ -1,5 +1,7 @@
+// Utilities for working with passwords
```
crookedneighbor, line 1: I'm conflicted about comments like this. Isn't it obvious that this is a lib about passwords?

paglias, line 1: Hmm probably yes, it's quite obvious

PR #6218, opened 2015-11-12
Increase code coverage for clearflags route

crookedneighbor:
@Alys @SabreCat @paglias

### DIFF
--- a/website/src/controllers/api-v2/groups.js
+++ b/website/src/controllers/api-v2/groups.js
@@ -412,8 +412,9 @@ api.clearFlagCount = function(req, res, next){
   if(user.contributor.admin){
     message.flagCount = 0;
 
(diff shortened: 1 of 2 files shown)


website/src/controllers/api-v2/groups.js
```diff
@@ -412,8 +412,9 @@ api.clearFlagCount = function(req, res, next){
   if(user.contributor.admin){
     message.flagCount = 0;
 
-    group.markModified('chat');
-    group.save(function(err,_saved){
+    Group.update({_id: group._id, 'chat.id': message.id}, {'$set': {
```
crookedneighbor, line 415: Not strictly necessary now. Not sure if there are performance implications for using `Group.update` over `group.save` but this should protect against changes in the future of a similar nature.

PR #6219, opened 2015-11-12
Js ify common

crookedneighbor:
# :coffee: Problem :coffee:

We're moving away from Coffeescript in favor of small modules written in ES2015. Many of our shared files are in Coffeescript.
# Solution

As the first step of a multi-step process, we should just ditch coffeescript altogether and use JS. The actual content of these files has not changed, so everything will work exactly as it did before.

Going forward, we should methodically extract each of the pieces of this file into separate modules.

I've set up browserify with babelify already, so each module can be written in ES2015.


crookedneighbor:
Bonus: no more stupid crazy nodemon freak out when the app starts up in a dev environment.

crookedneighbor:
I'm planning on merging this tomorrow if no objects.

PR #6222, opened 2015-11-12
Move old logging lib v2 directory

crookedneighbor:


tasks/gulp-console.js
```diff
@@ -2,7 +2,7 @@ import 'coffee-script';
 
 import mongoose from 'mongoose';
 import autoinc  from 'mongoose-id-autoinc';
-import logging  from '../website/src/libs/logging';
+import logging  from '../website/src/libs/api-v2/logging';
```
paglias, line 5: here we should use the new logging library

website/src/server.js
```diff
@@ -1,7 +1,7 @@
 // TODO cleanup all comments when finished API v3
 
 import nconf from 'nconf';
-import logging from './libs/logging';
+import logging from './libs/api-v2/logging';
```
paglias, line 4: and here

website/src/index.js
```diff
@@ -2,12 +2,12 @@
...
 
 // TODO remove this once we've fully converted over
-require('coffee-script'); 
+require('coffee-script');
 
 // Only do the minimal amount of work before forking just in case of a dyno restart
 var cluster = require('cluster');
 var nconf = require('nconf');
-var logging = require('./libs/logging');
+var logging = require('./libs/api-v2/logging');
```
paglias, line 10: new logging library here too

PR #6224, opened 2015-11-12
API v3 - i18n lib and getUserLanguage middleware

paglias:
@crookedneighbor I'll leave comments where I'm having problems with tests


test/api/v3/unit/middlewares/getUserLanguage.test.js
```diff
@@ -0,0 +1,89 @@
...
+            language: 'it',
+          },
+        },
+      };
+
+      getUserLanguage(req, res, next);
+      expect(req.language).to.equal('it');
+    });
+
+    xit('falls back to english if the user preferred language is not avalaible', () => {
```
paglias: this one would need `accepts` to be stubbed (as the other tests that are missing)

test/api/v3/unit/middlewares/getUserLanguage.test.js
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  describe('request with session', () => {
+    it('uses the user preferred language if avalaible', () => {
+      req.session = {
+        userId: 123
+      };
+
+      getUserLanguage(req, res, next);
+      expect(req.language).to.equal('it');
```
paglias, line 105: here it says that req.language is `undefined`, but https://github.com/HabitRPG/habitrpg/pull/6224/files#diff-5c9ea3bfb113e62ed2afc85bbc6db7a3R78 logs `it`

website/src/libs/api-v3/i18n.js
```diff
@@ -0,0 +1,98 @@
...
+fs.readdirSync(localePath).forEach((file) => {
+  if (file === 'en' || fs.statSync(path.join(localePath, file)).isDirectory() === false) return;
+  _loadTranslations(file);
+
+  // Merge missing strings from english
+  _.defaults(translations[file], translations.en);
+});
+
+// Add translations to shared
+shared.i18n.translations = translations;
```
crookedneighbor, line 48: This seems backwards to me. Why can't shared import the translations?

website/src/libs/api-v3/i18n.js
```diff
@@ -0,0 +1,98 @@
...
+
+// A map of languages that have different versions and the relative versions
+export let multipleVersionsLanguages = {
+  es: ['es-419', 'es-mx', 'es-gt', 'es-cr', 'es-pa', 'es-do', 'es-ve', 'es-co', 'es-pe',
+        'es-ar', 'es-ec', 'es-cl', 'es-uy', 'es-py', 'es-bo', 'es-sv', 'es-hn',
+        'es-ni', 'es-pr'],
+  zh: ['zh-tw'],
+};
+
+// Export en strings only, temporary solution for mobile
```
crookedneighbor, line 108: I think these comments about the mobile app can be removed. Ideally, habitrpg-mobile will never use api-v3

paglias, line 108: yep, but since this is going to be the only i18n lib when api v3 it goes out, I would like to wait for the android app to be discontinued before removing them

website/src/libs/api-v3/i18n.js
paglias, line 48: It was this way in the old code, probably there's a reason for it, the first one that comes to my mind is that the compiling step for the client would load a lot more data and also doesn't support the fs module

website/src/libs/api-v3/i18n.js
crookedneighbor, line 108: If that's the case, we'll need to adjust the android app to use this new path, I believe.

website/src/libs/api-v3/i18n.js
crookedneighbor, line 48: Ahh, that makes a lot of sense.

website/src/libs/api-v3/i18n.js
paglias, line 108: Ok, I thought this code was used in some other way. Then maybe this should be removed, I think we can review it later. If you know where it's used because I couldn't find it

website/src/middlewares/api-v3/getUserLanguage.js
```diff
@@ -0,0 +1,85 @@
...
+      return 'en';
+    }
+  } else if (matches.length > 0) {
+    return matches[0].toLowerCase();
+  } else {
+    return 'en';
+  }
+}
+
+function _getFromUser (user, req) {
```
crookedneighbor, line 51: I think it would make more sense like this:

```
let preferredLang;

if (user && user.preferences.language) {
  preferredLang = user.preferences.language;
} else {
  preferredLang = _getFromBrowser(req);
}

let lang = translations[preferredLang] ? preferredLang : 'en';

return lang;
```

The user's preferences _should_ be in a good state, but this will catch it if it isn't.

crookedneighbor, line 51: ^ Updated

crookedneighbor, line 51: Should the if check `user && user.preferences && user.preferences.language`?

paglias, line 51: yeah your code seems cleaner. About the if, i don't know, we can add the extra check but it isn't really necessary unless someone really messes with their data in a way that we don't allow. But definitively better than crashing everything. So yeah let's add it.

Actually for the code, something like this is even better:

```
let preferredLang = user && user.preferences && user.preferences.language;
let lang = translations[preferredLang] ? preferredLang : _getFromBrowser(req);

return lang;

```

Issue #6204, opened 2015-11-08
Timezone used by Habitica should be visible to the user


Alys:
When a player logs in to the website (i.e., literally logs in using the login form) or when they reload the website, their browser reads the timezone that their PC is set to and uses that for determining when cron should run and which Dailies should be visible. The timezone is stored in         `preferences.timezoneOffset` as a negative minutes offset (e.g., UTC+10:00 is -600).

It would be useful for this timezone to be visible to the user, to reduce questions about which timezone Habitica uses and to make it clear when the user's PC is set to the wrong timezone. (It would also be a first step in implementing a manual timezone setting https://github.com/HabitRPG/habitrpg/issues/6079.)

The timezone could be displayed on the [Settings -> Site](https://habitica.com/#/options/settings/settings) screen under the Custom Day Start form. It should be displayed as a numerical zone, relative to UTC, e.g., **UTC+10:00**

It should include a message that states these things (not necessarily with these words or in this order):
- Habitica uses the timezone set on your PC / mobile device.
- If you use two or more PCs / devices, the timezone should be the same on them all otherwise cron will be unpredictable.
- The timezone displayed here is the one that Habitica currently believes is yours.
- If that timezone is wrong, first reload the website to ensure that Habitica is reading the most recent timezone information from your PC. If it is still wrong, adjust the timezone on your PC.

We might also want to include more advanced instructions for users who do use more than one device. For example, after getting the timezone correct on one PC, repeat that process on all other PCs. If you use mobile devices, log in to Habitica there, and the come back to the Settings -> Site screen in the website and sync (NOT reload) to see which timezone the mobile device has set Habitica to; if it's incorrect, adjust your mobile device's timezone.

A developer working on this can use placeholder text until we determine the exact wording we want to use.

AminArria:
I'll work on this.

This timezone would be displayed inside the Custom Day Start form right? Not in another form below it

Alys:
Thanks @AminArria!

It should be in its own form under the CDS form. They're two different concepts so they need to be kept separate, and we don't want the users thinking that CDS is a good way to compensate for a wrong timezone.

PR #6228, opened 2015-11-13
Timezone information

AminArria:
Early PR for #6204 

Sorry about the extra commits, they somehow slipped


.gitignore
```diff
@@ -29,3 +29,5 @@ coverage.html
 common/dist/scripts/*
 
 test/spec/mocks/translations.js
+
+*.sublime*
```
crookedneighbor: You should add this to your global gitignore instead of to the project's gitignore

crookedneighbor:
Here's a screenshot of how it looks:

screen shot 2015-11-13 at 10 14 16 pm

@Alys @lemoness Thoughts?

website/public/js/filters/timezoneOffsetToUtc.js
```diff
@@ -0,0 +1,18 @@
+angular.module('habitrpg')
+  .filter('timezoneOffsetToUtc', function () {
+    return function (offset) {
+      if (offset > 0) {
```
crookedneighbor: You could make this:

``` js
var sign = offset > 0 ? '-' : '+';
```

common/locales/en/settings.json
```diff
@@ -140,5 +140,8 @@
     "gemCapExtra": "Gem Cap Extra:",
     "mysticHourglasses": "Mystic Hourglasses:",
     "paypal": "PayPal",
-    "amazonPayments": "Amazon Payments"
+    "amazonPayments": "Amazon Payments",
+    "timezone": "Timezone",
+    "timezoneUTC": "Your timezone is: <b><%= utc %></b>",
```
crookedneighbor: I feel like this should be either `"Your timezone: <b><%= utc %></b>"` or `"Your timezone is <b><%= utc %></b>"`

@lemoness Wording thoughts?

website/views/options/settings.jade
```diff
@@ -105,6 +105,18 @@ script(type='text/ng-template', id='partials/options.settings.settings.html')
                     ng-disabled='dayStart == user.preferences.dayStart')
                     =env.t('saveCustomDayStart')
 
+            hr
+
+            h5=env.t('timezone')
+            .form-horizontal
+              .form-group
+                .col-sm-12
+                  p
```
crookedneighbor: I'd prefer this to be on a single line:

```
p!=env.t('timezoneUTC', {utc: "{{ user.preferences.timezoneOffset | timezoneOffsetToUtc }}"})
```

Same with the `p` below

crookedneighbor:
Would it be more useful for the site to show what time it currently thinks it is, rather than the technical details of what utc offset it thinks it is?

common/locales/en/settings.json
Alys: I'd like it to say something more like "Habitica has detected from your PC / mobile device that your timezone is ..." or "Habitica has been informed by your PC / mobile device that your timezone is ...".

That makes it clearer where the problem lies when the value is wrong, and prevent Habitica appearing to be incorrect.

Alys:
I think showing what time it thinks it is would be more confusing than showing the timezone, and also make it less obvious what the fix is ("Do I need to change my PC clock??")

Also, it really is the timeZONE that's critical, especially during travel or daylight savings changes.

Alys:
Also, it really is the timeZONE that's critical, especially during travel or daylight savings changes.

Alys:
I'd like this:
"If that timezone is wrong, first reload the website to ensure that Habitica is reading the most recent timezone information from your PC. If it is still wrong, adjust the timezone on your PC." 
to be immediately below the timezone. That's the critical piece of information that people need to get their zone correct.

We could remove "Habitica uses the timezone set on your PC / mobile device." and "The timezone displayed here is the one that Habitica currently believes is yours." if we use the wording "Habitica has been informed by your PC / mobile device that your timezone is ..." at the point where the timezone is listed.

Sorry @AminArria, I realise I'm suggesting changes to the sample text that I provided myself. Seeing the screenshot has clarified my thoughts.

Alys:
What do others think about this suggestion from the issue:

We might also want to include more advanced instructions for users who do use more than one device. For example, after getting the timezone correct on one PC, repeat that process on all other PCs. If you use mobile devices, log in to Habitica there, and then come back to the Settings -> Site screen in the website and sync (NOT reload) to see which timezone the mobile device has set Habitica to; if it's incorrect, adjust your mobile device's timezone.

It would make the section more complex, but those instructions are important for people whose devices are on different zones. Obviously we could use better words than I've used there. :)

Alys:
On second thoughts... one of the reasons I want this feature is because we get questions like "What timezone does Habitica use?", so it **will** help to have a clear statement that "Habitica uses the timezone set on your PC / mobile device."  I'm going to pull @AminArria's code down to my local install and play around with the words until they are the way I have in mind, and then I'll upload a new screenshot for further comments.

website/public/js/filters/timezoneOffsetToUtc.js
```diff
@@ -0,0 +1,18 @@
...
+        var sign = '-';
+      } else {
+        var sign = '+';
+      }
+
+      offset = Math.abs(offset) / 60;
+
+      var hour = Math.floor(offset);
+
+      var minutes = (offset - hour) * 60;
```
Alys: Could you please add something here so that if the number of minutes is less than 10, it prints a leading zero? For example, 10:00 instead of 10:0.

AminArria:
@crookedneighbor  nice suggestion, specially the ternary operator, not used to using it jajaja,

@Alys Hadn't notice the missing 0 when values where less than 10 thanks.

Tomorrow I'll update with the changes.

Alys:
Here's what I have in mind for the text (I used "time zone" for consistency with the text in the CDS section):

Time Zone

Habitica uses the time zone set on your PC, which is: UTC+10:00

If that time zone is wrong, first reload this page using your browser's reload or refresh button to ensure that Habitica has the most recent information. If it is still wrong, adjust the time zone on your PC and then reload this page again.

If you use Habitica on other PCs or mobile devices, the time zone must be the same on them all. If your Dailies have been reseting at the wrong time, repeat this check on all other PCs and on a browser on your mobile devices.

screen shot 2015-11-15 at 4 08 11 pm

@lemoness what do you think? I know it's wordy but it contains everything a player needs to troubleshoot problems caused by incorrect zones.

lemoness:
I would bold the sentence "If you use Habitica on other..." because that's
pretty crucial information.

The use of UTC might confuse some players, but I'm not sure if there's an
easy way to address that.

Alys:
We could add a link to https://en.wikipedia.org/wiki/List_of_time_zones_by_UTC_offset for those that don't understand the UTC format.

Alys:
@AminArria Sorry, I've only just noticed your commits from five days ago (GitHub notifies us for comments but not for added commits unfortunately). They look good, thank you.

My proposed text in [this comment](https://github.com/HabitRPG/habitrpg/pull/6228#issuecomment-156785009) hasn't had any further changes suggested, except for Lemoness's request for bolding "If you use Habitica on other PCs or mobile devices, the time zone must be the same on them all", so you can go ahead and add that text when you're ready. Please use `<strong>` for bold, rather than `<b>`, for consistency with other strings.

AminArria:
@Alys updated the text

Alys:
@AminArria Thank you! I like how this looks! It will certainly help us with user support.

I'm merging this now as it is, but if we see indications that players are confused by UTC, we'll add a link to https://en.wikipedia.org/wiki/List_of_time_zones_by_UTC_offset or similar.

This will count towards a contributor tier. Please tell us your User ID from Settings -> API.

Final screenshot, for the record:

screen shot 2015-11-24 at 9 10 48 pm

AminArria:
@Alys d1c621be-418e-479d-8892-95a962533ca9

Alys:
Tier awarded.

PR #6237, opened 2015-11-14
port firebase lib to api v3

paglias:


website/src/libs/api-v3/firebase.js
```diff
@@ -0,0 +1,68 @@
+import Firebase from 'firebase';
+import nconf from 'nconf';
+const IS_PROD = nconf.get('IS_PROD');
+const FIREBASE_CONFIG = nconf.get('FIREBASE');
+const FIREBASE_ENABLED = IS_PROD && FIREBASE_CONFIG.ENABLED === 'true';
```
crookedneighbor, line 5: Do we want to make this a prod only thing? Can devs sign up for test accounts with Firebase?

website/src/libs/api-v3/firebase.js
```diff
@@ -0,0 +1,68 @@
...
+
+let firebaseRef;
```
crookedneighbor, line 7: I think this should just be 'firebase'. That's the normal convention for declaring instances of an object.

website/src/models/group.js
```diff
@@ -6,7 +6,7 @@ var _ = require('lodash');
 var async = require('async');
 var logging = require('../libs/api-v2/logging');
 var Challenge = require('./../models/challenge').model;
-var firebase = require('../libs/firebase');
+var firebase = require('../libs/api-v2/firebase');
```
crookedneighbor, line 9: Should this be v3?

crookedneighbor:
There is no actual content difference between 2 and 3, right? Should the models just reference v3 then?

crookedneighbor:
Looks good. Just had a few minor comments.

website/src/libs/api-v3/firebase.js
paglias, line 7: yeah but in this case it's not a normal instance, but a reference to a specific path in firebase

website/src/models/group.js
paglias, line 9: I plan to change this when we migrate the groups model

paglias:
Yeah no difference, I'm going to delete the v2 version later, we'll also delete the logging library and a few other things

website/src/libs/api-v3/firebase.js
paglias, line 5: Yes, you can signup for a dev account

crookedneighbor, line 5: I think the check should just be for FIREBASE_ENABLED.

paglias, line 5: ok, changing it

PR #6240, opened 2015-11-14
[API v3] Build manifest lib

paglias:


test/api/v3/unit/libs/buildManifest.test.js
```diff
@@ -0,0 +1,18 @@
+import {
+  getManifestFiles,
+} from '../../../../../website/src/libs/api-v3/buildManifest';
+
+describe('Build Manifest', () => {
+  describe('getManifestFiles', () => {
+    it('returns an html string', () => {
+      let htmlCode = getManifestFiles('app');
+
+      expect(htmlCode).to.be.a.String;
```
crookedneighbor: I think the one thing I'd change here is to use 'to.match()' which takes a regex. That way we can assert on the html without being to brittle.

paglias: any idea on which regex to use here?

PR #6256, opened 2015-11-17
API v3 - apidoc wip

paglias:


website/src/middlewares/api-v3/static.js
```diff
@@ -0,0 +1,18 @@
+import express from 'express';
+import nconf from 'nconf';
+import path from 'path';
+
+const IS_PROD = nconf.get('IS_PROD');
+const MAX_AGE = IS_PROD ? 31536000000 : 0;
+const PUBLIC_DIR = path.join(__dirname, '/../../../public');
+const BUILD_DIR = path.join(__dirname, '/../../../build');
```
crookedneighbor, line 8: We should probably have a const for `COMMON_DIR` rather than referencing it in relation to `PUBLIC_DIR`

website/src/libs/api-v3/errors.js
```diff
@@ -7,8 +7,17 @@ export class CustomError extends Error {
   }
 }
 
-// NotAuthorized error with a 401 http error code
-// used when a request is not authorized
+/**
```
crookedneighbor, line 10: APIDoc lets you define errors too? That's pretty cool.

website/src/middlewares/api-v3/static.js
paglias, line 8: yep, probably yes but I don't like the whole concept of how static files are handled here. Ideally before we go live all static files will belong to a single directory so that we can just zip it and upload it somewhere

website/src/libs/api-v3/errors.js
paglias, line 10: errors for sure, but I think you can also define permission and in general every function/class can be defined once and used in multiple places

PR #6261, opened 2015-11-17
Math helpers ES2015 module

SabreCat:
Moves several math helper functions into a JS module imported into `common/script/index.js`, using ES2015 conventions.

Also removes an unused math function `api.monod`, apparently an alternative diminishing returns solution that was written but never adopted.

Feedback needed in review:
- [x] The comments were pulled over without modification. Are we establishing any best practices for comments that we could put in place here?
- [x] Test file doesn't yet use ES2015. Anything worth refactoring there at this time?


common/script/statHelpers.js
```diff
@@ -0,0 +1,55 @@
...
+const maxHealth = 50;
+
+/*
+  ------------------------------------------------------
+  Scoring
+  ------------------------------------------------------
+ */
+
+function tnl (lvl) {
+  return Math.round((Math.pow(lvl, 2) * 0.25 + 10 * lvl + 139.75) / 10) * 10;
```
crookedneighbor, line 32: There are so many magic numbers here. Let's split them out and make them intelligible.

common/script/statHelpers.js
```diff
@@ -0,0 +1,55 @@
...
+  A hyperbola function that creates diminishing returns, so you can't go to infinite (eg, with Exp gain).
+  {max} The asymptote
+  {bonus} All the numbers combined for your point bonus (eg, task.value * user.stats.int * critChance, etc)
+  {halfway} (optional) the point at which the graph starts bending
+ */
+
+function diminishingReturns (bonus, max, halfway) {
+  if (!halfway) {
+    halfway = max / 2;
+  }
```
crookedneighbor: I believe you can do this to define the default for halfway:

```
function diminishingReturns (bonus, max, halfway=max/2) {
```

crookedneighbor: Though the linter might yell at you, so do:

``` js
function diminishingReturns (bonus, max, halfway = max/2) {
```

common/script/statHelpers.js
```diff
@@ -0,0 +1,55 @@
...
+
+const maxHealth = 50;
+
+/*
+  ------------------------------------------------------
+  Scoring
+  ------------------------------------------------------
+ */
+
+function tnl (lvl) {
```
crookedneighbor: Let's give this a full name: `toNextLevel`

We can keep it as `api.tnl` in the main file.

common/script/statHelpers.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+}
+
+/*
+  ------------------------------------------------------
+  Health cap
+  ------------------------------------------------------
+ */
+
+const maxHealth = 50;
```
crookedneighbor: the convention for constants is to use screaming snake case:

```
const MAX_HEALTH = 50;
```

crookedneighbor: It can still be `api.maxHealth` in the main file.

test/common/statHelpers.test.js
```diff
@@ -0,0 +1,58 @@
+var sinon = require('sinon');
+var chai = require("chai")
+chai.use(require("sinon-chai"))
+var expect = chai.expect
```
crookedneighbor: These four lines are now unnecessary. They are instantiated by https://github.com/HabitRPG/habitrpg/blob/8b7f50bfcf75f302a4e3963c804fdf524228d638/test/helpers/globals.helper.js

test/common/statHelpers.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+var shared = require('../../common/script/index.js');
```
crookedneighbor: I would do:

```
import {
  maxHealth,
  maxLevel,
  // ... etc
} from '../../common/scirpt/index';
```

test/common/statHelpers.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+  var HEALTH_CAP = 50;
+  var LEVEL_CAP = 100;
+  var LEVEL = 57;
+  var BONUS = 600;
+  var MAXIMUM = 200;
+  var HALFWAY = 75;
+
+  it('provides a maximum Health value', function() {
+    expect(shared.maxHealth).to.eql(HEALTH_CAP);
```
crookedneighbor: Health cap is only used here. I'd declare it in this it block since it doesn't get used in the rest of the tests.

test/common/statHelpers.test.js
```diff
@@ -0,0 +1,58 @@
...
+describe('helper functions used in stat calculations', function() {
+
+  var HEALTH_CAP = 50;
+  var LEVEL_CAP = 100;
+  var LEVEL = 57;
+  var BONUS = 600;
+  var MAXIMUM = 200;
+  var HALFWAY = 75;
+
+  it('provides a maximum Health value', function() {
```
crookedneighbor: Each thing that you're testing should have it's own describe block. IE:

```
describe('helper functions used in stat calculations', () => {

  describe('maxHealth', () => {
    it('provides a maximum Health value', () => {
      const MAX_HEALTH = 50;
      expect(maxHealth).to.eql(HEALTH_CAP);
    });
  });

  describe('maxLevel', () => {
    it('returns a maximum level for attribute gain', () => {
      const LEVEL_CAP = 100;
      expect(maxLevel).to.eql(LEVEL_CAP);
    });
  });
});
```

test/common/statHelpers.test.js
```diff
@@ -0,0 +1,58 @@
...
+    });
+  });
+
+  describe('Experience to next level', function() {
+    it('increases Experience target from one level to the next', function() {
+      expect(shared.tnl(LEVEL + 1)).to.be.greaterThan(shared.tnl(LEVEL));
+    });
+  });
+
+  describe('diminishing returns', function() {
```
crookedneighbor: Let's use the arrow function syntax: `() =>` instead of `function()`.

anonymous `function ()` should only be used if we need to reference `this` in the function.

test/common/statHelpers.test.js
```diff
@@ -0,0 +1,58 @@
...
+    it('provides a value under the maximum, given a bonus and maximum', function() {
+      expect(shared.diminishingReturns(BONUS,MAXIMUM)).to.be.lessThan(MAXIMUM);
+    });
+
+    it('provides a value under the maximum, given a bonus, maximum, and halfway point', function() {
+      expect(shared.diminishingReturns(BONUS,MAXIMUM,HALFWAY)).to.be.lessThan(MAXIMUM);
+    });
+
+    it('provides a different curve if a halfway point is defined', function() {
+      expect(shared.diminishingReturns(BONUS,MAXIMUM,HALFWAY)).to.not.eql(shared.diminishingReturns(BONUS,MAXIMUM));
```
crookedneighbor: Spaces between arguments.

test/common/statHelpers.test.js
crookedneighbor: Let's move the BONUS, MAXIMUM, HALFWAY constants into this describe block.

crookedneighbor:
Overall, :+1: just left nitpicky styling comments.

Not sure about the comment question. With the exception of the diminishing returns comment, they don't really provide anything useful, but just leave them for now until we decide for sure.

common/script/statHelpers.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+function diminishingReturns (bonus, max, halfway) {
+  if (!halfway) {
+    halfway = max / 2;
+  }
+  return max * (bonus / (bonus + halfway));
+}
+
+export default {
```
paglias: can we export those individually instead of a single export? The to import them you'll have to do `import * as name from file`

crookedneighbor: He means do:

``` js
export function diminishingReturns (bonus, max, halfway) {
```

Then you can do 

``` js
import * as statHelpers from './path/to/stathelpers';
```

or 

``` js
import {
  maxHealth,
  diminishingReturns,
  // ...
} from './path/to/stathelpers';
```

common/script/statHelpers.js
SabreCat, line 32: Yeah, there was a whole conversation about this formula a while back. Many faces were palmed. What would make it intelligible, though, given how arbitrary a lot of it is (139.75??)? Give them blandly descriptive constant names, like EXPONENT and DIVISOR and OFFSET or something?

crookedneighbor, line 32: I'll think on this done more. I don't like exponent and divisor.

SabreCat, line 32: Would it make sense to merge this PR and write a ticket?

common/script/statHelpers.js
```diff
@@ -0,0 +1,44 @@
...
+export const MAX_HEALTH = 50;
+
+/*
+  ------------------------------------------------------
+  Scoring
+  ------------------------------------------------------
+ */
+
+export function toNextLevel (lvl) {
+  return Math.round((Math.pow(lvl, 2) * 0.25 + 10 * lvl + 139.75) / 10) * 10;
```
crookedneighbor, line 32: At the very least, we should split this out into multiple declarations. Though, again we have the problem with creating variable names for processes we don't fully understand :/

@Alys Do you know what these numbers signify?

crookedneighbor:
I added linting to the files in `test/common`, so the linting errors need to be corrected.

test/common/statHelpers.test.js
```diff
@@ -0,0 +1,64 @@
...
+    });
+
+    it('returns level cap if above cap', () => {
+      expect(capByLevel(LEVEL_CAP + LEVEL)).to.eql(LEVEL_CAP);
+    });
+  });
+
+  describe('toNextLevel', () => {
+    it('increases Experience target from one level to the next', () => {
+      expect(tnl(LEVEL + 1)).to.be.greaterThan(shared.tnl(LEVEL));
```
crookedneighbor: I think we should import the module directly, not assert on shared.

crookedneighbor: regardless, shared is not defined here.

common/script/statHelpers.js
Alys, line 32: Magic.

It's lost to the mists of time, probably. Perhaps it's a question for TR, although the numbers might have been just determined experimentally with no plan.

test/common/statHelpers.test.js
crookedneighbor: If we wanted more test coverage, we could do something like:

```
it('increases Experience target from one level to the next', () => {
  _.times(110, (level) => {
    expect(tnl(level + 1)).to.be.greaterThan(tnl(level));
  });
});
```

SabreCat:
Addressed all comments except for the ongoing `toNextLevel` discussion. @crookedneighbor @Alys

PR #6262, opened 2015-11-17
SM modal fixes

### LINKED ISSUE #6253: "You do not own this pet" (or mount) message when displaying a pet / mount you have just hatched / raised
When you hatch a pet and click on the "display" button in the modal screen that appears, you sometimes see the message ":You do not own this pet" and the pet does not appear in your avatar. However when you go to your Pets page, you can see the pet there, and it remains after a sync.

screen shot 2015-11-17 at 8 20 10 pm

A similar thing can probably happen when raising a mount.
## What to do if this happens to you:

Find the pet in your [Inventory -> Pets](https://habitica.com/#/options/inventory/pets) page (or the mount in your [Inventory -> Mounts](https://habitica.com/#/options/inventory/mounts)  page) and click on it. It will then be displayed in your avatar. Click on it a second time if you decide you don't want it there.

The error message is almost always just a temporary glitch and it doesn't mean that you don't own the pet or mount. However in rare circumstances, a sync error might later cause the hatching action to be reversed, in which case the pet (or mount) will disappear, but the hatching potion and egg and (or food/saddle and pet) should reappear in your inventory and you can use them again to re-hatch the pet (or re-raise the mount).

You do **_not**_ need to report this error by commenting below. We know it can happen to any user.

SabreCat:
Fixes #6253, allows quest pets to appear in hatching and raising modals, and makes code for deciding whether or not to show a generic levelup modal a bit more readable per comments on #6042.

### DIFF
--- a/website/public/js/controllers/inventoryCtrl.js
+++ b/website/public/js/controllers/inventoryCtrl.js
@@ -116,7 +116,7 @@ habitrpg.controller("InventoryCtrl",
       if (!$window.confirm(window.env.t('hatchAPot', {potion: potName, egg: eggName}))) return;
       user.ops.hatch({params:{egg:egg.key, hatchingPotion:potion.key}});
       if (!user.preferences.suppressModals.hatchPet) {
-        $rootScope.hatchedPet = {egg: eggName, potion: potName, eggKey: egg.key, pet: 'Pet-' + egg.key + '-' + potion.key};
+        $rootScope.hatchedPet = {egg: eggName, potion: potName, potionKey:potion.key, eggKey: egg.key, pet: 'Pet-' + egg.key + '-' + potion.key};
         $rootScope.openModal('hatchPet', {controller: 'InventoryCtrl', size: 'sm'});
       }
       $scope.selectedEgg = null;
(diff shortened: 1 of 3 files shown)


website/public/js/controllers/inventoryCtrl.js
```diff
@@ -116,7 +116,7 @@ habitrpg.controller("InventoryCtrl",
       if (!$window.confirm(window.env.t('hatchAPot', {potion: potName, egg: eggName}))) return;
       user.ops.hatch({params:{egg:egg.key, hatchingPotion:potion.key}});
       if (!user.preferences.suppressModals.hatchPet) {
-        $rootScope.hatchedPet = {egg: eggName, potion: potName, eggKey: egg.key, pet: 'Pet-' + egg.key + '-' + potion.key};
+        $rootScope.hatchedPet = {egg: eggName, potion: potName, potionKey:potion.key, eggKey: egg.key, pet: 'Pet-' + egg.key + '-' + potion.key};
```
crookedneighbor, line 119: Can we expand this object into multiple lines? Having it all on one makes it hard to read.

crookedneighbor:
Other than my one comment, this looks reasonable.

PR #6264, opened 2015-11-18
Two-handed weapons no longer automatically equip unless desired. fixes #6231

### LINKED ISSUE #6231: Buying Mage Staff causes shield hand equipment to unequip with auto-equip off
I have "Auto-equip new gear" turned off, but whenever I buy a Mage staff, I get the "[staff name] is two handed" message, and my shield hand equipment (Mustaine's Milestone Mashing Morningstar) unequips itself. My weapon hand equipment remains as it was before the purchase.

UID: 809d8b43-07a1-49ab-9bb5-ad3ba461688d

ojeytonwilliams:
fixes #6231

### DIFF
--- a/common/script/index.js
+++ b/common/script/index.js
@@ -1558,9 +1558,9 @@ api.wrap = function(user, main) {
         } else {
           if (user.preferences.autoEquip) {
             user.items.gear.equipped[item.type] = item.key;
+            message = user.fns.handleTwoHanded(item, null, req);
           }
           user.items.gear.owned[item.key] = true;
-          message = user.fns.handleTwoHanded(item, null, req);
           if (message == null) {
             message = i18n.t('messageBought', {
               itemText: item.text(req.language)
--- a/test/common/user.fns.buy.test.js
+++ b/test/common/user.fns.buy.test.js
@@ -100,35 +100,35 @@ describe('user.fns.buy', () => {
 
       expect(user.items.gear.equipped).to.not.have.property('armor');
     });
-      
-   it('removes one-handed weapon and shield if auto-equip is on and a two-hander is bought', () => {
+
+    it('removes one-handed weapon and shield if auto-equip is on and a two-hander is bought', () => {
       user.stats.gp = 100;
       user.preferences.autoEquip = true;
       user.ops.buy({params: {key: 'shield_warrior_1'}});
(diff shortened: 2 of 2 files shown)


crookedneighbor:
We should include test coverage for this: https://github.com/HabitRPG/habitrpg/blob/553548046f44149508131ba4e2a9a8f889260311/test/common/user.fns.buy.test.js#L88-L104

SabreCat:
Ahoy @ojeytonwilliams - can you see about resolving the merge conflicts on the PR?

SabreCat:
Though if you could use a hand resolving the conflict, we can definitely help. Let us know!

ojeytonwilliams:
There we go!  Hopefully the final version should be okay.

Also, in the process of sorting this out, I noticed a problem with the messages that pop up when you unequip two-handed weapons.  Should I create another issue for that?

Alys:
What's the problem? I can't work it out. :)

ojeytonwilliams:
Oh, sorry, that was a bit vague..  The problem is that when you have a two-hander equipped and then try to equip a single-handed item all it says is "<two-hander> is two handed" rather than, say, "<two-hander> un-equipped".

What I think it should say is "<X> un-equipped" whenever that happens, rather than just when you directly click on that item.  So, if you have a sword and shield equipped and then grab a staff it tells you that you've unequipped both of them and vice versa.

Alys:
Oh yes, I thought that too when I started testing, but now I don't think it actually says "<two-hander> is two handed" when a two-handed item is unequipped. It does say that when you equip it, but then if you wait until the notification disappears, it doesn't come back when you equip something else.

But yes, we could probably make those messages more explicit. If you'd like to create an issue for it, please do! Thank you!

ojeytonwilliams:
Sure.  I'll get right on that :)

SabreCat:
Looking good! The fix will be live soon. @ojeytonwilliams, this will count toward a contributor tier! What's your Habitica UUID (from Settings > API)?

ojeytonwilliams:
Cool!  My UUID is 24103022-40d6-4888-939f-2778926c4ba5

SabreCat:
Noted and thank you! Since contributor tiers require increasing amounts of effort from one to the next, this PR hasn't taken you to 3 just yet, but keep helping out and we'll express our gratitude accordingly!

PR #6266, opened 2015-11-18
[API v3] User controller

paglias:
- [x] Local login
- [x] Local signup
- [x] Link local auth to existing facebook account (removed from signup route)
- [x] Social signup / login
- [x] Tests


website/src/models/emailUnsubscription.js
```diff
@@ -1,14 +1,20 @@
...
+import common from '../../../common';
+import validator from 'validator';
 
 // A collection used to store mailing list unsubscription for non registered email addresses
-var EmailUnsubscriptionSchema = new mongoose.Schema({
+export let schema = new mongoose.Schema({
   _id: {
     type: String,
-    'default': shared.uuid
+    default: common.uuid,
```
crookedneighbor, line 9: Should we be more careful about reserved words? That's the reason we had it quoted before.

paglias, line 9: it's not necessary anymore since es5

test/api/v3/unit/middlewares/getUserLanguage.test.js
```diff
@@ -12,6 +13,11 @@ import accepts from 'accepts';
 describe('getUserLanguage', () => {
   let res, req, next;
 
+  let checkReqT = (req) => {
+    expect(res.t).to.be.a('function');
+    expect(res.t('help')).to.equal(i18n.t('help', req.language));
```
crookedneighbor, line 18: If you're not passing in `res`, do you need to pass in `req`?

website/src/libs/api-v3/setupRoutes.js
```diff
@@ -14,7 +14,7 @@ fs
     let controller = require(CONTROLLERS_PATH + fileName); // eslint-disable-line global-require
 
     _.each(controller, (action) => {
-      let {method, url, middlewares, handler} = action;
+      let {method, url, middlewares = [], handler} = action;
```
crookedneighbor, line 17: does the `= []` works as a default value here?

website/src/models/user.js
```diff
@@ -328,6 +348,7 @@ export let schema = new Schema({
     orderAscending: {type: String, default: 'ascending'},
     quest: {
       key: String,
+      // TODO why are we storing quest progress here too and not only on party object?
```
crookedneighbor, line 341: Yes. This is the user's progress throughout the day, not the total progress. It gets applied and cleaned up at cron.

test/api/v3/unit/middlewares/getUserLanguage.test.js
paglias, line 18: Can you explain it better, I'm not getting it. (yeah, the name should be checkResT)

website/src/libs/api-v3/setupRoutes.js
paglias, line 17: yep, I didn't add a default for url, method and handler because those are not optional

PR #6269, opened 2015-11-19
Convenience functions ES2015 module

SabreCat:
Moves several convenience functions into a JS module imported into common/script/index.js, using ES2015 conventions.

Also removes a couple of said functions (`dotGet` and `dotSet`) superseded by new functions available in recent versions of Lodash.
# To Do
- [x] Fix lint failures
- [x] `refPush` test for specifying an existing UUID
# Feedback Requested
- [x] Making a `codeHelpers` module may be silly (aren't all functions code helpers? heh). Better name or organization ideas?
- [ ] The test for `refPush` is a little clumsy, because the function itself is somewhat opaque (at least to a junior dev like me). Ideas for better coverage or readability?
- [ ] If, in the test `increments sort position by 1 for each added item`, I continue to have `refPush` generate its own UUIDs instead of supplying my own, the object gets messed up. The second item in the object gets its `id` and `sort` values overwritten by the third item to be added. A bug in `refPush`, maybe?
- [x] https://github.com/HabitRPG/habitrpg/pull/6269#commitcomment-14537442
- [x] https://github.com/HabitRPG/habitrpg/pull/6269#commitcomment-14537475


test/common/codeHelpers.test.js
```diff
@@ -0,0 +1,70 @@
...
+      const KEY_1 = TEST_KEYS[1];
+      expect(testObject[KEY_1].id).to.eql(KEY_1);
+    });
+
+    it('increments sort position by 1 for each added item', () => {
+      const ITEM_WITH_UUID_2 = {id: UUID + '2'};
+      refPush(testObject, ITEM_WITH_UUID_2);
+      const TEST_KEYS = keys(testObject);
+      const KEY_1 = TEST_KEYS[1];
+      const KEY_2 = TEST_KEYS[2];
```
SabreCat: Oh, this is an artifact of when I was trying to use two generated UUIDs. I don't actually need a constant here. Will fix (pending overall feedback on these tests).

common/script/codeHelpers.js
```diff
@@ -0,0 +1,29 @@
+export function $w (s) {
+  return s.split(' ');
+}
+
+export function uuid () {
```
crookedneighbor: I don't think we should use this. I think we should outsource our uuid-ing needs to a node module. @paglias Do you have objections to that? I've used https://www.npmjs.com/package/uuid with the v4 function.

common/script/codeHelpers.js
```diff
@@ -0,0 +1,29 @@
...
+  Reflists are arrays, but stored as objects. Mongoose has a helluvatime working with arrays (the main problem for our
+  syncing issues) - so the goal is to move away from arrays to objects, since mongoose can reference elements by ID
+  no problem. To maintain sorting, we use these helper functions:
+ */
+
+export function refPush (reflist, item, prune) {
+  if (prune == null) {
+    prune = 0;
+  }
+  item.sort = _.isEmpty(reflist) ? 0 : _.max(reflist, 'sort').sort + 1;
```
crookedneighbor: You need to import lodash to use these.

common/script/codeHelpers.js
```diff
@@ -0,0 +1,29 @@
...
+  });
+};
+
+/*
+  Reflists are arrays, but stored as objects. Mongoose has a helluvatime working with arrays (the main problem for our
+  syncing issues) - so the goal is to move away from arrays to objects, since mongoose can reference elements by ID
+  no problem. To maintain sorting, we use these helper functions:
+ */
+
+export function refPush (reflist, item, prune) {
```
crookedneighbor: you can define prune's default in the argument list: `(reflist, item, prune = 0)`

common/script/index.js
```diff
@@ -13,48 +15,11 @@ api = module.exports = {};
...
-  if (prune == null) {
-    prune = 0;
-  }
-  item.sort = _.isEmpty(reflist) ? 0 : _.max(reflist, 'sort').sort + 1;
-  if (!(item.id && !reflist[item.id])) {
-    item.id = api.uuid();
-  }
-  return reflist[item.id] = item;
-};
+$w = api.$w = codeHelpers.$w;
```
crookedneighbor: I did a check and this is not used outside of the common file. Therefore, it does not need to be exposed on the api object.

crookedneighbor:
I think we can just call the file helpers.

common/script/codeHelpers.js
SabreCat: Actually, is `prune` even used? It gets set there if not given, and then... that's it. :confused:

test/common/helpers.test.js
```diff
@@ -0,0 +1,67 @@
+import {
+  refPush,
+} from '../../common/script/index';
+
+import {
+  $w,
+} from '../../common/script/helpers';
```
SabreCat, line 5: With this no longer on the exported `api` object, I had to import it directly thus. Legit? If so, would it be preferable to import everything from `helpers.js`, or does it still make sense to import from `index` where we can?

paglias:
@SabreCat what's the status on this? It just needs some changes to tests?

SabreCat:
@paglias: Awaiting final code review from you and/or @crookedneighbor. I've updated the checklist in the PR description to include all my outstanding questions.

test/common/helpers.test.js
```diff
@@ -0,0 +1,101 @@
...
+    it('appends additional items', () => {
+      refPush(testObject, {});
+      expect(size(testObject)).to.eql(2);
+    });
+
+    it('assigns a UUID to the newly added item if none was supplied', () => {
+      const GENERATED_UUID = keys(testObject)[1];
+      const UUID_LENGTH = 36;
+
+      expect(GENERATED_UUID.length).to.eql(UUID_LENGTH);
```
paglias, line 55: if you want here to validate the UUID you can use the `validator` module `isUUID` method

common/script/helpers.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/*
+  Reflists are arrays, but stored as objects. Mongoose has a helluvatime working with arrays (the main problem for our
+  syncing issues) - so the goal is to move away from arrays to objects, since mongoose can reference elements by ID
+  no problem. To maintain sorting, we use this helper function:
+ */
+
+export function refPush (reflist, item) {
+  item.sort = isEmpty(reflist) ? 0 : max(reflist, 'sort').sort + 1;
```
paglias, line 20: the only thing I don't like here is that we're adding the `sort` property not only to the item pushed inside the refPush but also changing the original `item` object, it's safer if we `_.clone` the object before

paglias, line 20: yep, I'd definitively use `_.clone` to clone the item at the beginning of this method. It might also fix the bug described in the PR description though I'm not sure

crookedneighbor:
This branch is now up to date with lodash v3.

I've done a little work on this as well to make things a bit more readable.

common/script/helpers.js
crookedneighbor, line 20: The only place refpush is used is in the members controller for inbox messages.

crookedneighbor:
I'll take over this PR. Looks like the only outstanding thing is to fully review what refPush is doing.

SabreCat:
Given the changes coming in #6834, this is getting headed off. We can revisit later in v3.

PR #6275, opened 2015-11-22
Tasks add push to bottom

### LINKED ISSUE #6248: enabling pushTaskToBottom
I'm a fairly new user (again) - I'm using pushTaskToTop regularly, but adding new ToDo's is a chore because my list is so long I have to plan on spending half a minute per entry, as I have to drag most tasks all the way to the bottom.  Having pushTaskToBottom working would save me quite a bit of time (and encourage good habits...).

I see that pushTaskToBottom is there but commented out:
https://github.com/HabitRPG/habitrpg/blob/2d862593ae4152e2b3a61c66f71dc203ea2341de/website/views/shared/tasks/meta_controls.jade#L18  
I didn't see any commits explaining why.  If it's for UI clutter reasons, I'd like to suggest binding ctrl-click on pushTaskToTop as a UI element.  I need to do my sorting on desktop anyway, so it's not a real problem that I can't ctrl-tap on mobile.

I don't have an instance setup here at this point.  I didn't want to go to the trouble of doing that and submitting a pull request if there's a reason pushTaskToBottom isn't currently enabled (like maybe it doesn't actually work yet).

TheHollidayInn:
This is should fix #6248 . When ever you click and hold ctrl or command (for mac) the pushToBottom button will invert. 
screen shot 2015-11-22 at 3 55 08 pm


common/locales/en/tasks.json
```diff
@@ -89,7 +89,7 @@
     "sureDelete": "Are you sure you want to delete this task?",
     "streakCoins": "Streak Bonus!",
     "pushTaskToTop": "Push task to top",
-    "pushTaskToBottom": "Push task to bottom",
+    "pushTaskToBottom": "Push task to bottom. Hold ctrl or cmd to push to top.",
```
Alys: Should the "Hold ctrl...." sentence be on the pushTaskToTop text?

website/public/css/tasks.styl
```diff
@@ -394,6 +394,11 @@ for $stage in $stages
   a.badge
     position: relative; top:-2px
 
+  a.push-down span
+    -ms-transform: rotate(180deg); /* IE 9 */
+    -webkit-transform: rotate(180deg); /* Chrome, Safari, Opera */
+    transform: rotate(180deg);
+
```
Alys: This method for changing the icon isn't working on my browser, so we can't rely on it working for all users. It might also produce odd results in future if we change the icon to one that doesn't look meaningful upside down (not likely, but you never know). Did you see crookedneighbor's suggestion in the issue about applying a class to the body when ctrl/cmd is pressed? The class could be named something generic so that it can be reused in cases other than pushToTop/Bottom.

TheHollidayInn: Gotcha. Is it okay for me to make an upside down version of the icon? 

I have added the `push-down` class when ctrl/cmd is clicked. The link doesn't have a class in the push up stated.

common/locales/en/tasks.json
TheHollidayInn: Actually, the pushTaskToTop isn't being used, I believe. I could be wrong though.

website/public/css/tasks.styl
crookedneighbor: The correlating glyphicon class is glyphicon-save

TheHollidayInn: Ah, right, it is just the bootstrap class. I'll post an update soon, then.

common/locales/en/tasks.json
SabreCat: It looks like you've got this backward. Push to top is still the default, and Ctrl or Cmd flips it to push-to-bottom. It's the `pushTaskToBottom` string that doesn't get used:
image

SabreCat:
The functionality looks good on Windows Chrome; I can also check Firefox and Microsoft Edge, but I'll need a Mac user to check the command button take. @TheHollidayInn, the string discussion is the only thing I see needing cleanup so far. If there's an unused string, it can be removed.

SabreCat:
Oh, never mind Microsoft Edge; you apparently can't even log in to Habitica there. o.O But this looks good in Firefox.

website/public/js/controllers/tasksCtrl.js
```diff
@@ -76,10 +76,23 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
...
+        $scope.pushLocation = "bottom";
+      }
+    });
+
+    $scope.$on("habit:keyup", function (e, keyEvent) {
+      if (ctrlKeys.indexOf(keyEvent.keyCode) !== -1) {
+        $scope.pushLocation = "top";
+      }
+    });
+
```
Alys: Can you change the words "pushLocation" and "top" and "bottom" to something more generic? If I understand your code correctly, then `$scope.pushLocation` is set based on whether the ctrl/cmd key is pressed or not. We're using that here for pushing tasks, but it could be used elsewhere in future for other purposes. A generic name like 'ctrlPressed' with values of 'true' or 'false' would make more sense in other contexts. Lines 83 t0 95 could then be moved to a more generic location in the file, since they aren't necessarily specific to pushing tasks.

Alys:
I can test on Mac after the next commit.

TheHollidayInn:
Okay, I fixed the strings, and I moved the ctrlPressed logic to the rootScope so that the $scope.ctrlPressed can be used anywhere the rootScope is.

SabreCat:
Bug: if you use the "Dated" view of tasks, you can push things to the top OK, but if you attempt a push to bottom, you get "Task not found."

Alys:
@SabreCat ordering tasks manually isn't supported in the dated view because they are sorted automatically by date. The "task not found" is something that the sorting code produces when there's an id mismatch during sorting - it's expected behaviour here. I've been meaning to log an issue to have that message suppressed in the dated view, and perhaps even in the non-dated view since there's not much the user can do about it - it basically means "sort failed, try again" (but don't try again in the dated view, manual sorting won't work).

Alys:
This works on Mac OS.

SabreCat:
Great, I'm going to merge this in! @lemoness, do we maybe want to drop a note about this into the next Bailey announcement, because it's a feature that folks might not notice right away?

lemoness:
Yes, I will slot it in! This announcement is pretty full, so it will be in the next one.

PR #6278, opened 2015-11-23
Api v3 user custom validators

crookedneighbor:


website/src/controllers/api-v3/user.js
```diff
@@ -30,31 +30,37 @@ api.registerLocal = {
...
-      {'auth.local.lowerCaseUsername': lowerCaseUsername},
-    ]}, {'auth.local': 1})
-    .exec()
-    .then((user) => {
-      if (user) {
-        if (email === user.auth.local.email) throw new NotAuthorized(res.t('emailTaken'));
-        // Check that the lowercase username isn't already used
-        if (lowerCaseUsername === user.auth.local.lowerCaseUsername) throw new NotAuthorized(res.t('usernameTaken'));
-      }
+    req.checkBody({
```
crookedneighbor, line 33: This all could be extracted out into a validations module. IE,

```
req.checkBody(userValidations.registration(req, res));
```

That would make the registration controller very simple and almost all related directly to creating a new user.

crookedneighbor, line 33: In case it wasn't clear, `userValidation.registration(req, res)` would return the object that is represented here.

website/src/libs/api-v3/customExpressValidators.js
```diff
@@ -0,0 +1,46 @@
...
+      if (user) {
+        reject();
+      } else {
+        resolve();
+      }
+    });
+  });
+}
+
+export function isUniqueEmail (email) {
```
crookedneighbor, line 28: Not sure how I feel about having to split this up into two looks ups.

website/src/libs/api-v3/customExpressValidators.js
```diff
@@ -0,0 +1,46 @@
...
+  if (!email) return true;
```
crookedneighbor, line 29: We have to return true here, otherwise it tries to call `toLowerCase()` on an undefined variable. If we return true, than this validation passes and is not added as a validation error. If we return false, than the "email is taken" message appears for not providing an email, which seemed wrong to me.

Issue #6272, opened 2015-11-21
When trying to hatch a pet that I already have, the popup still appears


sporeball:
**User ID:** ce3d1e2b-be53-4e5e-8101-273157548178
**Operating System:** Windows 10 (Google Chrome)

_Popup and alert_
![image](https://cloud.githubusercontent.com/assets/12767408/11315692/50e3b5bc-8fab-11e5-8af2-434e5e3b7af6.png)

_Proof that I have the Zombie Tiger_
![image](https://cloud.githubusercontent.com/assets/12767408/11315703/ac21c162-8fab-11e5-8ffc-dd454a337476.png)

Alys:
Confirmed. Thanks @MrEikono!

ojeytonwilliams:
I looked at this earlier and noticed that there is a fair bit of logic going on in user.ops.hatch to see if a pet should appear.  However you can't get the result back to the controller without calling hatch with a callback and _that_ stops the messages from appearing.

I feel like it should be possible to get a result back from user.ops functions, otherwise you'll end up having to duplicate their logic elsewhere.

Is this a reasonable change?

SabreCat:
@ojeytonwilliams: I had the same thought, looking at this and wishing we could get the existing-pet check up ahead of the `$window.confirm`. That's a broader issue and a larger refactoring effort than this fix, though, so if you'd like to spin off a new issue ticket, we can discuss further there!

PR #6281, opened 2015-11-23
Prevented modal from showing when user tries to hatch a pet they own

TheHollidayInn:
This is a fix for #6272 . I added a front-end check to see if the user already owns the pet before showing the hatch modal.


SabreCat:
Looking good, thank you! I've noted this in your progress toward contributor tier 7.

PR #6290, opened 2015-11-24
[API v3] [WIP] Base model plugin

paglias:


website/src/libs/api-v3/baseModel.js
```diff
@@ -0,0 +1,53 @@
+import _ from 'lodash';
+import { uuid } from '../../../../common';
```
crookedneighbor, line 1: This should just use the uuid module.

website/src/libs/api-v3/baseModel.js
```diff
@@ -0,0 +1,53 @@
...
+  schema.add({
+    _id: {
+      type: String,
+      default: uuid.v4,
+      validate: [validator.isUUID, 'Invalid uuid.'], // TODO check for UUID version
+    },
+  });
+
+  if (options.timestamps) {
+    schema.add({
```
crookedneighbor, line 15: Would it be beneficial to have this on every model? Ids there a case where we wouldn't want it?

website/src/libs/api-v3/baseModel.js
```diff
@@ -0,0 +1,53 @@
...
+    });
+
+    return objToSanitize;
+  };
+
+  if (Array.isArray(options.private)) privateFields.push(...options.private);
+  if (!schema.options.toObject) schema.options.toObject = {};
+  schema.options.toObject.transform = function transformToObject (doc, plainObj) {
+    privateFields.forEach((fieldPath) => {
+      _.set(plainObj, fieldPath, undefined); // TODO decide wheter to use delete here
```
crookedneighbor: Why not use delete here? I think it makes more semantic sense.

crookedneighbor: Sorry, meant to ask if there is some weird quirk of using delete.

crookedneighbor:
Overall, I think this looks great. Will look at the build tomorrow.

website/src/libs/api-v3/baseModel.js
paglias, line 1: I had the same idea but there are 2 modules, uuid and node-uuid with the first being a fork of the second and there are talks of merging the two so I didn't want to rely on one and then have to change it in 100 different places if it ever changed

website/src/libs/api-v3/baseModel.js
paglias, line 15: you're talking about timestamps? The fact is that user model has already auth.timestamps with createdAt and loggedIn so it would be a bit of a duplication

website/src/libs/api-v3/baseModel.js
paglias: Yeah it definitively makes more sense but:
1. lodash doesn't support it
2. they say it slow down things but shouldn't be a problem with such objects

So I used _.set to have something working quickly and that supported deep paths but I'll switch to delete soon

paglias:
Ok, don't worry about the build I can look at it

paglias:
Properties are being deleted instead of being set to undefined and it only transform toJSON instead of toObject

paglias:
Added the ability to pass additional transformation functions for sanitize and toJSON to be passed as options

crookedneighbor:
Looks reasonable to me.

PR #6305, opened 2015-11-26
Fixes #6252 - Guild description is too narrow on guild page

### LINKED ISSUE #6252: Guild description is too narrow on guild page
When looking at a guild page with a long description the site forces you to scroll very far down, even if you have a screen that has plenty of space too allow it to be wider.

image

It would be nice if this had a variable width, so it would just use the space that is available. For example like this: 
image

markhowson:
Attempted to fix #6252 by creating a new popover class.

### DIFF
--- a/website/public/css/alerts.styl
+++ b/website/public/css/alerts.styl
@@ -16,6 +16,9 @@
 .wide-popover
   max-width: 400px
 
+.fill-popover
+  max-width: max-content
+
 // Small alerts
 alert-sm-mixin()
   padding-top: 8px
--- a/website/public/css/alerts.styl
+++ b/website/public/css/alerts.styl
@@ -14,9 +14,6 @@
   display: inline-block
 
 .wide-popover
-  max-width: 400px
-
-.fill-popover
   max-width: max-content
 
 // Small alerts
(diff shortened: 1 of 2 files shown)


SabreCat:
This is a big improvement! One thing that does happen, though, is that a wide bubble runs close enough up against the Guild Gems sticker that it looks like it's overlapping:

image

Maybe we could include a bit of margin/padding so there's a visual break between those two elements?

markhowson:
Yeah, I did wonder what to do in that case, but I didn't want to make too many changes at once! The 'guild gems' sticker looks a little bit lonely sat out there anyway to my eye (there's a lot of left-aligned elements, and then this one floating out on the right). I'd maybe suggest having it on the left side anyway, as it seems like a property of the guild itself.

I'll see if I can sort the padding situation though, thanks!

markhowson:
image

SabreCat:
@markhowson: Looking good, thanks! Your changes will go live soon. This will count toward a contributor tier! What's your Habitica User ID (from Settings > Site > API)?

markhowson:
Thanks! : 22b8625c-ab67-4812-a832-978464d28049

SabreCat:
@markhowson: Tier awarded! Note that while your first tier is easy to get, further tiers require increasing amounts of effort from one to the next, so another PR of this scope may not bump you to 2. But keep helping out and we'll express our thanks accordingly!

PR #6309, opened 2015-11-27
[API v3] Port tasks and tags

paglias:
Missing:
- [x] Adapt user tests 
- [ ] add login tests and addLocalAuthToSocial that are missing
- [ ] Write tests for tasks
- [x] Delete task
- [x] Prevent challenge tasks from being deleted
- [x] Update task
- [x] Decide whether checklist should have their own routes, mainly for scoring and updating / removing items
- [x] Score checklist
- [x] Update checklist item
- [x] Remove checklist item
- [ ] Port score logic
- [x] Sort tasks and make sure tasks are added to correct user.tasksOrder list
- [x] Separate completed and due todos, making completed todos non sortable and not returned unless requested esplicitly
- [ ] Find a way to correct errors in user.tasksOrder (maybe in cron?)
- [x] Disable updates only for a few fields (_id, type, ...)


test/api/v3/integration/tasks/POST-create_task.test.js
```diff
@@ -0,0 +1,67 @@
...
+  let api;
+
+  before(() => {
+    return generateUser().then((generatedUser) => {
+      user = generatedUser;
+      api = requester(user);
+    });
+  });
+
+  context('checks "req.body.type"', () => {
```
crookedneighbor: I'd make this context something like "required params" or "param validation"

test/api/v3/integration/tasks/POST-create_task.test.js
```diff
@@ -0,0 +1,67 @@
...
+    it('returns an error if req.body.type is absent', () => {
+      expect(api.post('/tasks', {
```
crookedneighbor: This expect call is using chai-as-promised and needs to be returned to work properly. (otherwise mocha might finish the test before checking the expectation)

It's an async test, and the way to do that in Mocha is either to include a callback or to return a promise. One of the things that chai-as-promised does is turn the expect function into a promise.

test/api/v3/integration/tasks/POST-create_task.test.js
```diff
@@ -0,0 +1,67 @@
+import {
```
crookedneighbor: I think the structure for naming should actually be:

```
test/api/v3/integration/<route_namespace>/<REST_ACTION>-<route_path>.test.js
```

So this one would be

```
test/api/v3/integration/tasks/POST-tasks.test.js
```

The idea is to make the integration tasks as agnostic about the actual workings of the controllers as possible. All they know is that they give a route some data, and get some data back.

test/api/v3/integration/tasks/POST-tasks_score_id_direction.test.js
```diff
@@ -0,0 +1,121 @@
+import {
```
paglias, line 1: @crookedneighbor most of this aren't already tested in common when we test the score function? Would it be enough to just test that the right params are passed to the score function?

PR #6312, opened 2015-11-28
Added check for total points before skipping attribution

### LINKED ISSUE #6294: dying after level 100 permanently removes an attribute point
If you are above level 100 and you die, you lose an attribute point. It is then impossible to gain the point back because when you level up after level 100, you can't gain points.

("Impossible" unless you do a workaround which is: adjust your level down to below 100, level up, and then adjust your level up again to what it should be.)

There's two ways we could fix this:
1. Prevent an attribute point being lost if you die above level 100.
2. Allow an attribute point to be gained if you level up above level 100 and your points are less than 100.

My preference at the moment is for the second method because it would fix the problem for anyone who's suffered from this in the past and doesn't know about the workaround, however if the code to implement the first method was significantly cleaner, we could go with that and do a database migration to fix people who needed a fix.

The second method is also good, in my opinion, because it imposes an additional, temporary penalty for dying, in the same way as dying below level 100. Opposing opinions are welcome. :)

TheHollidayInn:
Fixes #6294 . Before, leveling up would skip stat attribution if the user was over level 100. I have added a new condition that the player must also have 100 total stat points.


common/script/index.js
```diff
@@ -2279,13 +2279,14 @@ api.wrap = function(user, main) {
           user.stats.lvl++;
           tnl = api.tnl(user.stats.lvl);
           user.stats.hp = 50;
-          if (user.stats.lvl > api.maxLevel) {
+          var totalStatPoints = user.stats.str + user.stats.int + user.stats.con + user.stats.per;
+          if (user.stats.lvl > api.maxLevel && totalPoints >= 100 ) {
```
Alys: 100 should be api.maxLevel
I think the check for user.stats.lvl is no longer needed - checking total points should be enough?

TheHollidayInn: Yep, i think you are right. The points seem to outweigh the level criteria. I'll make a change when I get home tonight unless anyone else thinks the check should stay.

TheHollidayInn:
Updated

common/script/index.js
```diff
@@ -2279,13 +2279,14 @@ api.wrap = function(user, main) {
           user.stats.lvl++;
           tnl = api.tnl(user.stats.lvl);
           user.stats.hp = 50;
-          if (user.stats.lvl > api.maxLevel) {
+          var totalStatPoints = user.stats.str + user.stats.int + user.stats.con + user.stats.per;
+          if (totalPoints >= 100) {
```
Alys: Please change 100 to api.maxLevel

Alys: and totalPoints should be totalStatPoints

TheHollidayInn:
Sorry, I misunderstood what you meant before. I just amended the last commit.

SabreCat:
This looks pretty sound. (I tweaked the PR description a bit because this isn't a cron change.) Can we add tests for this behavior?

Be advised there may be some delay in merging this even once we have tests, because with API v3 development underway, changes to backend behavior incur some extra overhead at the moment. (FYI @paglias)

TheHollidayInn:
Thanks for the update, @SabreCat . I can add tests. Can you point me to any examples of testing this kind of behavior? I'm looking through the tests now, but I have only written Angular and API tests. I think this is a bit outside of the API.

SabreCat:
@TheHollidayInn: You'll want to poke around in https://github.com/HabitRPG/habitrpg/tree/develop/test/common to get familiar with the `common` tests. `algos.mocha.js` is a bloated test file that we're slowly refactoring out into individual module tests, as exemplified by the `user.fns`, `user.ops` etc. tests in there. You'll want to model your test's style after the individual modules, but you may find tests in `algos` related to what you're doing.

TheHollidayInn:
Okay, I have added the tests! It wasn't too bad. 

I have some extra code I will remove in here, but I wanted to ask if this is correct so far. Basically, I ran into a problem, it seems that if I want to unit test `updateStats`, I need to resolve the `if (!user._tmp) user._tmp = {}` issue that is in another PR. I may be able to het past that by calling a full cron (which seems to be happening in other tests), but I figured that was against the spirit of unit tests.

So, if this looks good, I will add my tests to the other PR, let that get merged, then come back an clean this up.

crookedneighbor:
Please pay attention to the linting rules: 

```
    3:8   error  "sinon" is defined but never used                no-unused-vars
   88:7   error  Expected blank line after variable declarations  newline-after-var
   88:7   error  Unexpected var, use let or const instead         no-var
   89:18  error  Missing trailing comma                           comma-dangle
   90:8   error  Missing semicolon                                semi
   97:42  error  Missing semicolon                                semi
  101:7   error  Unexpected var, use let or const instead         no-var
  101:7   error  Expected blank line after variable declarations  newline-after-var
  102:18  error  Missing trailing comma                           comma-dangle
  103:8   error  Missing semicolon                                semi
  110:50  error  Missing semicolon                                semi
  114:2   error  Block must not be padded by blank lines          padded-blocks
```

TheHollidayInn:
Sorry, I just saw these errors and I was running `gulp test:common` before.

crookedneighbor:
No worries, :)

TheHollidayInn:
Okay, with my recent PR now merged, correct the ._tmp var error, this PR should be able to progress. I have made updates to the tests as well.

test/common/user.fns.updateStats.test.js
```diff
@@ -0,0 +1,50 @@
...
+      user.stats.lvl = 99;
+      user.stats.str = 25;
+      user.stats.int = 25;
+      user.stats.con = 25;
+      user.stats.per = 24;
+      user.fns.updateStats(stats);
+      expect(user.stats.points).to.eql(1);
+    });
+
+    it('Does not add an attibute point when user\'s stat points are equal to max level', () => {
```
SabreCat, line 36: Wait, shouldn't the user get a point _at_ level 100, just not for any levels thereafter?

TheHollidayInn, line 36: Yep, but the problem is articulated in the issue above. Basically, what happens is the user can lose a point if they level to a 100 (and sometimes passed) and then die. Thus, losing a skill point. There was code checking the level of 100, but that case is included if we simply check to ensure a user has all the stat points needed.

common/script/index.js
```diff
@@ -2281,13 +2281,14 @@ api.wrap = function(user, main) {
           user.stats.lvl++;
           tnl = api.tnl(user.stats.lvl);
           user.stats.hp = 50;
-          if (user.stats.lvl > api.maxLevel) {
+          var totalStatPoints = user.stats.str + user.stats.int + user.stats.con + user.stats.per;
+          if (totalStatPoints >= api.maxLevel) {
```
crookedneighbor: We should probably include a maxStatPoints constant instead of referencing maxLevel.

test/common/user.fns.updateStats.test.js
SabreCat, line 36: Ahh, I get it. I misread the test description. Carry on!

common/script/index.js
TheHollidayInn: I think it is a constant: `api.maxLevel = statHelpers.MAX_LEVEL;` 
https://github.com/TheHollidayInn/habitrpg/blob/leveling-ensure-user-gets-100-points/common/script/index.js#L23

crookedneighbor: Correct. I'm saying we should add an additional stat points constant. Something like:

``` js
// statHelpers.js
export const MAX_STAT_POINTS = MAX_LEVEL;

// index.js
var userTotalStatPoints = user.stats.str + user.stats.int + user.stats.con + user.stats.per;
if (userTotalStatPoints >= MAX_STAT_POINTS) {
```

I think it's more semantically meaningful to check against a max stat points constant than a max level constant.

TheHollidayInn: Ohhh. Sorry. I mixed up what you said with the maxLevel because of the comment's location. I'll set that up right now.

TheHollidayInn:
Okay, I made the update. I just imported the new constant like all the others. 

```
api.maxLevel = statHelpers.MAX_LEVEL;
api.maxStatPoints = statHelpers.MAX_STAT_POINTS;
api.capByLevel = statHelpers.capByLevel;
api.maxHealth = statHelpers.MAX_HEALTH;
```

Let me know if you want me to change this. Also, I rebased the branch with upstream.

crookedneighbor:
Made a couple tweaks (the notable one being to take the MAX_STAT_POINTS constant off of the public API).

TheHollidayInn:
Sweet. Thanks!

common/script/index.js
```diff
@@ -2281,13 +2282,14 @@ api.wrap = function(user, main) {
...
-          if (user.stats.lvl > api.maxLevel) {
+          var userTotalStatPoints = user.stats.str + user.stats.int + user.stats.con + user.stats.per;
+          if (userTotalStatPoints >= api.maxStatPoints) {
             continue;
           }
           if (user.preferences.automaticAllocation) {
             user.fns.autoAllocate();
           } else {
-            user.stats.points = user.stats.lvl - (user.stats.con + user.stats.str + user.stats.per + user.stats.int);
+            user.stats.points = user.stats.lvl - userTotalStatPoints;
```
Alys, line 2292: This needs to set user.stats.points to api.maxStatPoints if user.stats.points > api.maxStatPoints. 

user.fns.autoAllocate() should also be checked to ensure it can't increase points above maxStatPoints

PR #6314, opened 2015-11-29
Changed language when deleteing tasks to indicate type and text of task

### LINKED ISSUE #6146: Deleting a reward asks you if you want to delete the *task*
image

The text should read "Are you sure you want to delete this reward?"

_EDIT by Alys 2019-08-29:_
See https://github.com/HabitRPG/habitica/issues/6146#issuecomment-523241278 for the current status of this bug and the desired fix.

TheHollidayInn:
Fixes #6146 . This is just continuing the discussion in https://github.com/HabitRPG/habitrpg/pull/6153 . 
screen shot 2015-11-29 at 10 16 20 am
screen shot 2015-11-29 at 10 16 13 am
screen shot 2015-11-29 at 10 16 02 am
screen shot 2015-11-29 at 10 15 49 am


Alys:
`sureDelete` is also used here:

website/public/js/controllers/challengesCtrl.js:215:      if (!confirm(window.env.t('sureDelete'))) return;

TheHollidayInn:
Bam. Added. I see that via the Github search now.

screen shot 2015-11-29 at 2 29 25 pm

website/public/js/controllers/challengesCtrl.js
```diff
@@ -212,7 +212,7 @@ habitrpg.controller("ChallengesCtrl", ['$rootScope','$scope', 'Shared', 'User',
     };
 
     $scope.removeTask = function(task, list) {
-      if (!confirm(window.env.t('sureDelete'))) return;
+      if (!confirm(window.env.t('sureDelete', {taskType: task.type, taskText: task.text}))) return;
```
crookedneighbor: The task type should also use a translation string.

TheHollidayInn: As in, `env.t(task.type)` ? 

If so, I am only seeing the plural value here: https://github.com/HabitRPG/habitrpg/blob/develop/common/locales/en/tasks.json#L35 . Should I add the singular forms for each word?

crookedneighbor: Yes to both.

crookedneighbor:
```
if (!confirm(window.env.t('sureDelete', {taskType: window.env.t(task.type), taskText: task.text}))) return;
```

TheHollidayInn:
Updated! Can I add `common/dist/sprites/habitrpg-shared.css` to .gitignore?

crookedneighbor:
Unfortunately no, not until the old android app is fully deprecated.

crookedneighbor:
Do we not have tests around any of this behavior?

crookedneighbor:
Actually, since the CSS file gets created on build, it should be fine to gitignore it. I'll have to review it and get back to you.

paglias:
I'm pretty sure that that file was ignored then we were having problems
with it being built during the heroku build process and moved it to git

2015-11-30 19:53 GMT+01:00 Blade Barringer notifications@github.com:

> Actually, since the CSS file gets created on build, it should be fine to
> gitignore it. I'll have to review it and get back to you.
> 
> —
> https://github.com/HabitRPG/habitrpg/pull/6314#issuecomment-160723602.

## 

Matteo Pagliazzi - paglias.net

SabreCat:
Marked as needing tests per @crookedneighbor's comment above, but if we do in fact have this covered, just point it out and we'll keep rolling along!

TheHollidayInn:
It seems we have tests for removing task in the challengesCtrl. I have added tests for removing task in the taskCtrl. 

I'm starting to think we should abstract this logic to a service though, since it is duplicated. I believe the task logic in the challengesCtrl is less dependent on the User service, which would allow us some abstraction. I'm not sure. 

Also, we can do some rewriting to test the actual alert box text. Especially if we use a default confirmation modal. I'm unsure if there is currently a way to test the alert boxes text.

test/spec/controllers/tasksCtrlSpec.js
```diff
@@ -29,10 +34,36 @@ describe('Tasks Controller', function() {
...
+
+    it('asks user to confirm deletion', function() {
+      scope.removeTask(task);
+      expect(window.confirm).to.be.calledOnce;
+    });
+
+    it('does not remove task if not confirmed', function() {
+      window.confirm.returns(false);
+      scope.removeTask(task);
+      expect(User.user.ops.deleteTask).to.not.be.calledOnce;
```
SabreCat: This would pass if for some reason `User.user.ops.deleteTask` were called _more_ than once. You want `.to.not.be.called`.

test/spec/controllers/tasksCtrlSpec.js
```diff
@@ -1,12 +1,17 @@
 'use strict';
 
 describe('Tasks Controller', function() {
-  var $rootScope, shared, scope, user, ctrl;
+  var $rootScope, shared, scope, user, User, ctrl;
 
   beforeEach(function() {
-    user = specHelper.newUser();
```
crookedneighbor, line 7: You can leave this and just have:

```
User = {
  user: user
};
```

Then in the tests you can reference `user` instead of `User.user` for everything.

paglias:
Are we sure we want to make Habit, Daily, To-Do, ... translatable?

crookedneighbor:
The plural versions are already translated. Why wouldn't we want the singular version?

paglias:
Hmm. then it makes sense. it's just that lately I've seen a lot of names on
the web are being kept in english but in our case habits / dailies / ...
aren't just names so it probably makes sense to have them translatable

2015-12-05 22:28 GMT+01:00 Blade Barringer notifications@github.com:

> The plural versions are already translated. Why wouldn't we want the
> singular version?
> 
> —
> https://github.com/HabitRPG/habitrpg/pull/6314#issuecomment-162249260.

## 

Matteo Pagliazzi - paglias.net

test/spec/controllers/tasksCtrlSpec.js
TheHollidayInn, line 7: If I leave it that way then I have to call controllers with this notation `$controller('RootCtrl',  {$scope: scope, User: {user: user}});` . I just decided to clean up the notation at the top and keep it consistent with other tests files. That way, we don't have to call `User: {user: user}` at every controller initialization.

TheHollidayInn:
Updated!

test/spec/controllers/tasksCtrlSpec.js
crookedneighbor, line 7: No, you're misunderstanding.

``` js
user = specHelper.newUser();
User = {
  user: user,
};
user.ops = {
  deleteTask: sandbox.stub(),
};

module(function ($provide) {
  $provide.value('User', User);
  $provide.value('Guide', {});
});

// actual tests
expect(user.ops.deleteTask).to.be.calledOnce;
// is equivalent to `expect(User.user.ops.deleteTask).to.be.calledOnce;` but is easier to read
```

TheHollidayInn, line 7: Ah, yes. Sorry. I interpreted the other way because of the change I made. I have updated the code with this syntax.

SabreCat:
Looking good! Thanks for yet another useful fix, @TheHollidayInn!

TheHollidayInn:
Thanks for the help!

PR #6315, opened 2015-11-29
Initialized usr._temp when score function is skipped

### LINKED ISSUE #4986: Cannot level up with skill (burst of flames, backstab) to levels 15, 30, 40, 60 (from levels 14, 29, 39, 59) -- "Error Cannot set property 'drop' of undefined"
#### Level up using tasks or habits instead. You'll then be able to use your skills normally.

This is a temporary workaround until the bug is fixed. 
### If you experience this bug, you do not need to comment here. It happens for all players in this situation.

FYI, this bug only occurs when levelling up to a level that gives you a free quest scroll.

---

This is the error I get every time I use backstab. Furthermore, upon reload of the page or upon cron, my backstab is "undone" as it were and everything that would have happened had I not backstabbed is reversed.

Please let me know what I can do to fix this! :)
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10573752-cannot-level-up-with-skill-burst-of-flames-backstab-error-cannot-set-property-drop-of-undefined-see-chrome-console-for-more-details?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheHollidayInn:
Fixes #4986 . What seems to be happening is that the `score` function is skipped when either burst of flames or backstab is cast. In the `score` function,`user._tmp` is initialized. So, when that initialization is skipped, that caused the error to be produced.


common/script/index.js
```diff
@@ -2338,6 +2338,8 @@ api.wrap = function(user, main) {
           if (analytics != null) {
             analytics.track('acquire item', analyticsData);
           }
+          console.log(user._tmp)
```
paglias: you should remove the console statement

TheHollidayInn: Eh, my bad habit continues. Removed. Thanks!

paglias: hopefully we'll have a code style in place to prevent such things soon!

TheHollidayInn: Like a Gulp task that runs lint?? That would be nice!

paglias: actually we already have it https://github.com/HabitRPG/habitrpg/blob/develop/tasks/gulp-eslint.js, but it doesn't cover most files because we still have to port them, it'll come with api v3

TheHollidayInn: Ah, gotcha. Sweet!

SabreCat:
Nice simple change, not sure why we hadn't gotten around to this already. XD Maybe worth writing a test that would fail against the unfixed behavior?

TheHollidayInn:
Okay, tests are added. The second test for each spell fails without the change. Basically, the second test is the same as the first, but the user level is change to prevent breakage before reaching the user._tmp variable.

If these tests are good, I can begin writing more tests for other spells in a separate PR.

test/common/shared.spells.test.js
```diff
@@ -0,0 +1,128 @@
...
+
+  buffs = {
+    per: 0,
+    int: 0,
+    con: 0,
+    str: 0,
+    stealth: 0,
+    streaks: false,
+  };
+  user = {
```
SabreCat: There's a lot of data in here not actually used in the test. It doesn't need to be a complete user object, just enough to  get through the code paths being tested.

test/common/shared.spells.test.js
```diff
@@ -0,0 +1,128 @@
...
+  let user;
+
+  beforeEach(() => {
+    user = newUser();
+    shared.wrap(user);
+  });
+
+  context('Rogue Spells', () => {
+    it('Should add exp and gp to user when backstab is used', () => {
+      let previousExp = user.stats.exp;
```
SabreCat: Use `const` for static values. `let` is for when you're going to modify the variable along the way.

test/common/shared.spells.test.js
```diff
@@ -0,0 +1,128 @@
...
+      let previousGp = user.stats.gp;
+
+      shared.content.spells.rogue.backStab.cast(user, user.todos[0]);
+      expect(user.stats.exp).to.greaterThan(previousExp);
```
SabreCat: `to.be.greaterThan` would read a little more naturally.

test/common/shared.spells.test.js
```diff
@@ -0,0 +1,128 @@
...
+      expect(user.stats.gp).to.greaterThan(previousGp);
+    });
+
+    it('Should pass when user is leveled passed all breaks and backstab is used', () => {
```
SabreCat: I don't understand this description. "It should pass" - sure, all tests should pass, but what's actually being tested? What does "user is leveled passed all breaks" mean?

test/common/shared.spells.test.js
```diff
@@ -0,0 +1,128 @@
...
+      let previousExp = user.stats.exp;
+      let previousGp = user.stats.gp;
+
+      user.stats.lvl = 99;
```
SabreCat: If we're testing what happens when you Backstab past a level that awards a scroll or the like, shouldn't we set the user's level and Experience to replicate that scenario?

test/common/shared.spells.test.js
TheHollidayInn: It is to my understanding that most of this is needed during this test. When commenting out the data to find out what I did not need, I ran into errors. I will get you some more specific details soon.

test/common/shared.spells.test.js
TheHollidayInn: Gotcha. I mix up this notation with Swift's constants - which are `let` now.

test/common/shared.spells.test.js
TheHollidayInn: Well, I can rethink how I describe this test for sure. It was ambiguous to me as well. Basically, there are many return statements that prevent the code from reaching the undefined usr._temp variable. To circumvent those stops, I simply placed the user's level at 99 and called the update function. This ensure's the usr._tmp function is called and that the full function would complete with the updated PR above. So, simply, I am testing that the function can completely work.

Let me play around with this though. I was trying to stay in context of creating a function that fails without my change and passes with the change as requested above.

test/common/shared.spells.test.js
crookedneighbor: Ultimately this should be pulled out into a helper.

test/common/shared.spells.test.js
crookedneighbor: Also, code style for constants is screaming snake case: `PREVIOUS_EXP`

test/common/shared.spells.test.js
crookedneighbor: Yes.

test/common/shared.spells.test.js
crookedneighbor: :+1:

test/common/shared.spells.test.js
TheHollidayInn: I thought screaming case was only for global constants, not local constants, at least, that is what I do in C++. I will have an update in the next 30 mins or so.

test/common/shared.spells.test.js
TheHollidayInn: Okay, I left the comments to show how much I could remove and keep the tests passing. I think we might as well keep the full object and abstract this function to a helper class. This function was taken from another test.

TheHollidayInn:
Alright, I'm unsure of why I didn't think of this before, but I simply added the explicit tests for the issue, i.e., the level for each major milestone. I think they are slightly redundant because of edge cases, but with the current set up, I believe these are good tests. They also all fail without the change I have made in the initial PR.

TheHollidayInn:
I merged the PR, thanks @crookedneighbor . Couldn't I have rebased and forced pushed in place of merging the develop branch? That's what I usually do.

crookedneighbor:
The main point of the PR was to extract the new user function into a helper. I also merged down ~~master~~ develop to bring it up to date.

TheHollidayInn:
I understand. I was just checking for clarity. :)

crookedneighbor:
I did a little work to simplify the tests: https://github.com/HabitRPG/habitrpg/commit/ce06d8a873059d69e9a863f7e549842d6e02d711

Couple of notes:
- our code style for tests is to declare what the functions do, not what they should do. (I'm sure we're inconsistent with this)
- It was more meaningful to test that the user leveled up when crossing the exp threshold with the spells and attained the quest drop, rather than just having many tests for the level up milestones without any context for why they were significant. (in the future, we may want to include a loop around that test to programmatically test the different quest drop scenarios, but for now I think it's fine to include just the first)
- I split out the exp and gp backstab tests into separate it blocks

Thanks for your work on this, was sorely needed :)

TheHollidayInn:
Sweet. Thanks. This makes sense. I was moving towards this direction. I figured there was some general way to explain the milestones, but I am still becoming familiar with the leveling logic. Thanks for the revisions!

PR #6324, opened 2015-12-01
Unequip notification fix

### LINKED ISSUE #6298: Notifications when equipping and unequipping items lack clarity
When a user directly unequips an item by clicking on it on the equipment page, a message pops up saying "<%= itemText %> un-equipped".  However no message appears if the user unequips it indirectly (by equipping something that conflicts with it).

I think the messages should inform the user they've unequipped something in both cases, with an extra notification about two-handed weapons if the two-handedness matters.

I'd be happy to implement this.

ojeytonwilliams:
Fixes #6298

This simplifies the message output when two-handed weapons are involved, getting rid of some redundant notifications.

In order to let user.ops functions send more than one message I ended up having to change the ops wrapping in userServices, so that the ops functions can parse arrays and send a message for each element.  I think this is the cleanest way to handle both single and multiple messages.


SabreCat:
Cool, we'll take a look! Be advised though that we're deep in a major overhaul of the back end (API v3!), so bear with us as it may take more time/hoop-jumping to get changes to `common` files merged right now. FYI @paglias

paglias:
This can be merged, I still haven't touched that part of the codebase! Thanks for the ping @SabreCat

common/script/public/userServices.js
```diff
@@ -107,10 +107,14 @@ angular.module('habitrpg')
                       if (err) {
                         var message = err.code ? err.message : err;
                         console.log(message);
-                        if (MOBILE_APP) Notification.push({type:'text',text:message});
-                        else Notification.text(message);
-                        // In the case of 200s, they're friendly alert messages like "Your pet has hatched!" - still send the op
-                        if ((err.code && err.code >= 400) || !err.code) return;
+                        if (typeof message === 'string') 
+                          message = [message];  
+                        _.each(message, (msg) => {
```
SabreCat: This line fails during the `uglify:build` step of deploy, complaining of an `Unexpected token: operator (>)`. I think we're not using the arrow function syntax in this file yet, and it's getting confused by that.

common/script/public/userServices.js
```diff
@@ -107,10 +107,14 @@ angular.module('habitrpg')
                       if (err) {
                         var message = err.code ? err.message : err;
                         console.log(message);
```
SabreCat: Not an issue in this PR, but while we're stomping around here--is there a reason we spit all of these into the console? It's always struck me as kinda odd. @crookedneighbor

Alys: It's been like that since forever. I've never found it useful.

common/script/public/userServices.js
ojeytonwilliams: Ah, that didn't crop up when I ran the tests.  Is there some way I can catch these kinds of issues?

common/script/public/userServices.js
crookedneighbor: Remove the console log.

common/script/public/userServices.js
crookedneighbor: We're in the process of moving everything over to ES2015. This is one of the files that is still using es5

crookedneighbor: Also, I could be wrong, but I believe making the message returned from the equip call an array would be a breaking change in the API.

common/script/index.js
```diff
@@ -2063,23 +2063,33 @@ api.wrap = function(user, main) {
...
     handleTwoHanded: function(item, type, req) {
-      var message, ref, weapon;
+      var message, ref, weapon, shield;
       if (type == null) {
         type = 'equipped';
       }
       if (item.type === "shield" && ((ref = (weapon = content.gear.flat[user.items.gear[type].weapon])) != null ? ref.twoHanded : void 0)) {
         user.items.gear[type].weapon = 'weapon_base_0';
-        message = i18n.t('messageTwoHandled', {
+        message = [i18n.t('messageTwoHandled', {
```
crookedneighbor: Why not just change this message to include the information about un-equipping? That would eliminate the need to change this to an array.

common/script/public/userServices.js
paglias: Agree, this is a breaking change

ojeytonwilliams:
Right, that should be better.  I decided to go with two types of messages over one, since I couldn't find a nice way of gluing the existing messages together.

Hope that's okay!

common/locales/en/messages.json
```diff
@@ -19,7 +19,8 @@
   "messageAlreadyPet": "You already have that pet. Try hatching a different combination!",
   "messageHatched": "Your egg hatched! Visit your stable to equip your pet.",
   "messageNotEnoughGold": "Not Enough Gold",
-  "messageTwoHandled": "<%= gearText %> is two handed",
+  "messageTwoHandedEquip": "Wielding <%= twoHandedText%> takes two hands, so <%= offHandedText%> has been unequipped.",
```
crookedneighbor: There should be a space before the second `%`. So `<%= twoHandedText %>`

ojeytonwilliams:
I've corrected the templates.  Also, for consistency, I changed "un-equip" to "unequip".

SabreCat:
OK, looking good! This will go live soon. @ojeytonwilliams, I've added this to your progress toward contributor tier 3. Thank you!

ojeytonwilliams:
You're welcome!  Sorry it was a bit more of an adventure than was strictly necessary.

Issue #5924, opened 2015-09-13
Audio Volume


fragmental:
http://danielstern.github.io/ngAudio/#/
Would this be a good option for volume control?

The only reason i never use the audio is that there's no way for me to control the volume for just habitrpg alone, and end up getting blasted after listening to just about anything else first and then doing a task.

I've been using habitrpg for a 5 year old, and I thought the audio might make habitrpg more interesting.
##

crookedneighbor:
That looks good to me. :+1:

crookedneighbor:
*looks not lols

SabreCat:
There was also some discussion of normalizing the audio levels of the themes on the site. Are we tracking that issue already?

bennyty:
@SabreCat I'm looking to help with Habitica and I know Audacity well enough to use the audio normalizing filter. To what level do you want the sound normalized?

EDIT: I've been doing some reading and there are programs that analyze more than the peak audio. They look at how the whole audio piece will sound to the human ear and try to balance around that. While I realize this is less important for the short sound effects we have, I have just run the algorithms over the sound effects in common/audio and the results seem impressive enough.

SabreCat:
@bennyty: That sounds great, no pun intended! Could you submit the modified audio files in a pull request so we can all have a listen?

Alys:
I'm putting this on hold because we're most of the way through a project to change the website's front-end code and this might no longer be relevant afterwards, or if it does, there'll be new code that will need to be modified to fix it.

lemoness:
This isn't a current need, so I'm closing this for now. We can always reopen if necessary.

Etherfly:
I'd like to highlight that this is still a problem with the website that needs solving.
https://www.reddit.com/r/habitica/comments/n1mxvd/how_do_i_turn_the_sound_down/

I personally feel that the volume is too high for most sound themes, and the only theme that I currently find okay in terms of volume is the LuneFox's theme. All others blast twice as loud and way louder that everything else I might have in the browser.

PR #6334, opened 2015-12-04
Normalize theme audio volumes

bennyty:
I used a combination of mp3Gain and Vorbisgain to rebalance these audio files. I am confident that the .mp3 files sound more balanced across the board but am not confident that the .ogg files have been
modified correctly. Perhaps my audio player isn't compatible?

From the Vorbisgain website:

> It needs player support to work. Non-supporting players will play back the files without problems, but you'll miss out on the benefits. Nowadays most good players such as Winamp or Foobar2K or recent XMMS are already compatible. If your player isn't, consider upgrading to one that is.


bennyty:
@SabreCat mentioned wanting to do this in issue #5924

crookedneighbor:
If you include: 

fixes #5924

in a comment, or commit message, it will automatically close the issue if this PR gets merged in.

bennyty:
@crookedneighbor Thank you! I already knew this but thought that this PR does not address that specific issue. @fragmental's issue dealt with adding volume control while this PR only normalizes the existing audio files.

Alys:
For those of us not familiar with audio, what do rebalancing and normalising mean?

SabreCat:
I like how this sounds from what I've tested thus far! Can somebody with a better ear than me try Watts' Theme on production vs. this PR, though? I feel like the tempo's faster on the normalized version, somehow.

bennyty:
@Alys I would not call myself immensely "familiar" with audio but as I understand it what I have done is normalization not balancing (have updated the PR title).

Since I don't want to get this wrong this Wikipedia's definition:

> Audio normalization is the application of a constant amount of gain to an audio recording to bring the average or peak amplitude to a target level (the norm).

So since these audio files were (presumably) created by different people using different functions/sources, they are wildly different in volume(amplitude). The filter that I ran the audio through tries to make all the files the same reasonable volume.

SabreCat:
Never mind my comment about the speed. I just used that theme a bit on production and it was as rapid as when using this PR. It'd been a while since I'd used that theme (I prefer LuneFox), so I guess I just remembered it being a different tempo than it actually was. :cat::sweat_drops:

SabreCat:
@bennyty: Thanks much for this! It'll be live next time we deploy to the site. Your work will count toward a contributor tier! What's your Habitica User ID (from Settings > API)?

bennyty:
Thanks! I'm glad you found it useful! UserID: 97d3419d-5277-4e1f-aac8-77a1ac1b30a5

SabreCat:
@bennyty: Tier awarded, congratulations!

Note that contributor tiers require increasing amounts of effort from one to the next, so a PR of similar scope may not immediately bump you to Tier 2. But keep helping out and we'll express our thanks accordingly!

PR #6343, opened 2015-12-06
[API v3] Tasks 2

paglias:
- [x] finish writing tests
- [x] port algos test for scoring task and random drops
- [x] GET user (with tests)
- [ ] checklist tests
- [x] tags tests
- [x] tasks' tags tests
- [x] port score logic (and tests)
- [ ] port cron logic (and tests)
- [ ] port cron middleware (and tests)
- [ ] add support for Mongoose's VersionError to error handler


common/script/api-v3/scoreTask.js
```diff
@@ -0,0 +1,251 @@
...
+export default function scoreTask (options = {}, req = {}) {
+  let {user, task, direction, times = 1, cron = false} = options;
+  let delta = 0;
+  let stats = {
+    gp: user.stats.gp,
+    hp: user.stats.hp,
+    exp: user.stats.exp,
+  };
+
+  // TODO return or pass to cb, don't add to user object
```
crookedneighbor, line 178: Whatever we do, we need some way to pass it on to the API client in the response.

test/api/v3/integration/tasks/POST-tasks.test.js
```diff
@@ -50,21 +56,41 @@ describe('POST /tasks', () => {
...
+        createdAt: 'yesterday',
+        updatedAt: 'tomorrow',
+        challenge: 'no',
+        completed: true,
+        streak: 25,
+        dateCompleted: 'never',
+      }).then((task) => {
+        expect(task.userId).to.equal(user._id);
+        expect(task.history).to.eql([]);
+        expect(task.createdAt).not.to.equal('yesterday');
```
crookedneighbor, line 76: `.to.not.equal` reads more naturally than `.not.to.equal`

common/script/api-v3/scoreTask.js
paglias, line 178: I was thinking about  saving this to user.notifications

PR #6360, opened 2015-12-11
clear local storage page added

bkovacik:
This is a response to [issue 6337](https://github.com/HabitRPG/habitrpg/issues/6344).

I created a page that suggests to users to first try clearing local storage before reporting a bug, then offers the report bug link that was in its place.


website/src/routes/pages.js
```diff
@@ -18,7 +18,7 @@ router.get('/', i18n.getUserLanguage, locals, function(req, res) {
 
 // -------- Static Pages --------
 
-var pages = ['front', 'privacy', 'terms', 'api', 'features', 'videos', 'contact', 'plans', 'new-stuff', 'community-guidelines', 'old-news', 'press-kit', 'faq', 'overview', 'apps'];
+var pages = ['front', 'privacy', 'terms', 'api', 'features', 'videos', 'contact', 'plans', 'new-stuff', 'community-guidelines', 'old-news', 'press-kit', 'faq', 'overview', 'apps', 'bug'];
```
crookedneighbor: Not sold on the word "bug" for this.

website/views/static/bug.jade
```diff
@@ -0,0 +1,21 @@
...
+  .row
+    .col-md-12
+      .page-header
+        h1=env.t('communityBug')
+      p
+        =env.t('tryFirst')
+        button(onclick='localStorage.clear()') Clear local storage
+        br
+        =env.t('tryNext')
+        a(target='_blank', href='https://github.com/HabitRPG/habitrpg/issues/2760') Report a Bug
```
crookedneighbor: Shouldn't this use the `Report a Bug` translation string?

website/src/routes/pages.js
bkovacik: Sure. Would something like report-bug make more sense?

Alys: How about "clear-browser-data"? @darkmeep Wait a day or so to see if there's any objections to that, or any better ideas, and if not, you can go ahead with that name.

Alys:
The primary purpose of this page is to allow an easy way of clearing data, and we'll often be sending people to it when they've already reported a bug, so while it's great idea to have the bug report link there, it's not the main focus.

For the page title, try "Clear Your Browser's Stored Data", and for the paragraph above the button, use "If you are experiencing problems with Habitica, click the button below to clear cookies and local storage for this website (other websites will not be affected). You will need to log in again after doing this, so first be sure that you know your log-in details, which can be found at Settings -> Site."  Make Settings -> Site a link to that page. The exact words will probably change after the staff have reviewed this page, but this will do for a start.

The `tryNext` text can be "If the problem persists, please [report a bug] if you haven't already." That sentence will need to be in a single string so that translators can use the best word order for their language - see how `linkStart` and `linkEnd` are used in other translation strings.

website/views/static/bug.jade
```diff
@@ -0,0 +1,21 @@
...
+    =env.t('communityBug')
+
+block content
+  .row
+    .col-md-12
+      .page-header
+        h1=env.t('communityBug')
+      p
+        =env.t('tryFirst')
+        button(onclick='localStorage.clear()') Clear local storage
```
Alys: This should say "Clear data" and be a translatable string.

Please also include clearing cookies. Although I suppose if that's complex, we could remove mention of cookies from the text I've just suggested and go live with local storage only, and then add cookies later. Local storage is the main thing that I want to have cleared. Cookies tend to be easier to clear with built-in browser features.

website/src/routes/pages.js
crookedneighbor: I like clear-browser-data

website/views/static/bug.jade
crookedneighbor: If it's going to be multiple things, it's probably best to extract this into an angular function.

Alys:
For simplicity now, let's make this just clearing local storage. We very often see cases where players need to do that, and there isn't a consistent method across all browsers for it, whereas there is for clearing cookies. The sooner we have this page, the easier our support efforts will be. :) In a later stage, we can create an angular function as crookedneighbor suggests and add support for cookies.

@darkmeep Would you like to make the suggested text changes and push them to this PR? Please also add a screenshot to the PR so that the staff can review it easily.

bkovacik:
Hi, sorry about the delay. I tried to incorporate your suggestions into a new commit.

An example render looks like this:
Preview

crookedneighbor:
Merged in https://github.com/HabitRPG/habitrpg/commit/7df3b333f62b39212c170cf6936bcdf35bf72eeb

PR #6365, opened 2015-12-12
removed gymnasium dependency status

VirenMohindra:
broken link- no point in keeping it in the readme


crookedneighbor:
Thanks for noticing that!

Issue #4522, opened 2015-01-15
Part of the site doesn't use translated strings


Atomk:
As the title says, I have found parts of the site using English strings even if there is a translated string available. I checked only the Italian translation but I am pretty sure that such problems can be found also in other languages (it's not the first time I discover these issues in the site).
What I've found till now:
1.  **[solved]** Backgrounds descriptions
2. **[solved]** "Subscribe" form
3.  **[solved]** Task progress
4. **[solved]** "Submit" button in Coupon page
5. **[solved]** "Press Kit" page _("Press Kit" in the navbar is still untranslated, the rest of the page is ok)_
6.  **[solved]** "Rewards" in quest scroll form (when purchasing one in the Market)
7. **[solved]** "Last updated" in Community guidelines
8. **[solved]** "Buy Gems" form (some string are already translated but not used)
9. **[solved]** Set names in "Time Travelers" page
   ![2015-03-22 12_13_16-greenshot](https://cloud.githubusercontent.com/assets/6122017/6769144/8050ca82-d08f-11e4-8dd6-a61166620416.png)
10. **[solved]** "Send gift" (in the form that pops up when you click on username label)
11. **[solved]** "Invite Friends" button (in "Members" section in guilds and party)
12. **[solved]** https://habitrpg.com/static/contact
##

TheHollidayInn:
It seems that point 2 is fixed.

![screen shot 2015-03-02 at 5 53 01 pm](https://cloud.githubusercontent.com/assets/1253400/6453810/c885b624-c107-11e4-91db-74126a51745a.png)

I have created a pull request for number 4: https://github.com/HabitRPG/habitrpg/pull/4788 .

I'll be taking a look at number 5 now.

Atomk:
@TheHollidayInn "Card", "Apply" and "Promo Code" translations are still missing in your screenshot, my fault, I should have marked it in the picture I uploaded. Thank you anyway!

NotGonnaGitUs:
Added a pull request which fixes the following from the list
- 2. "Subscribe" form
- 4. "Submit" button in Coupon page
- 5. "Press Kit" page (Page Title)
- 7. "Last updated" in Community guidelines
- 8. "Buy Gems" form

Atomk:
@NotGonnaGitUs I have just checked but only the 5 and part of the 8 seem to be fixed, maybe I need to wait for a site/translation update?

NotGonnaGitUs:
[...]

PR #6367, opened 2015-12-13
Fix for #9 of Part of the site doesn't use translated strings #4522

NathanMaton:
Fix for #9 of Part of the site doesn't use translated strings #4522


paglias:
@NathanMaton thanks for the PR!

The build is broken which means some tests are failing after your changes, also why did you add a new package that is not used anywhere and those 2 files to gitignore? 

As for the code itself, you should move the env.t call from the jade file directly to the JavaScript one so that the texts always show he new strings, not only in the template

common/script/content/mystery-sets.js
```diff
@@ -4,122 +4,122 @@ let mysterySets = {
   201402: {
     start: '2014-02-22',
     end: '2014-02-28',
-    text: 'Winged Messenger Set',
+    text: 'wingedMessengerSet',
```
crookedneighbor: This the linting error that caused the build to fail. A trailing blank space.

common/locales/en/subscriber.json
```diff
@@ -63,6 +63,30 @@
     "timeTravelersPopover": "We see you have a Mystic Hourglass, so we will happily travel back in time for you! Please choose the pet, mount, or Mystery Item Set you would like. You can see a list of the past item sets <a href='http://habitica.wikia.com/wiki/Mystery_Item' target='_blank'>here</a>! If those don't satisfy you, perhaps you'd be interested in one of our fashionably futuristic Steampunk Item Sets?",
     "timeTravelersAlreadyOwned": "Congratulations! You already own everything the Time Travelers currently offer. Thanks for supporting the site!",
     "mysticHourglassPopover": "A Mystic Hourglass allows you to purchase certain limited-time items, such as monthly Mystery Item Sets and awards from world bosses, from the past!",
+    "wingedMessengerSet":"Winged Messenger Set",
```
crookedneighbor: I think these should be namespaced with the prefix "mysterySet" so you'd have:

``` json
"mysterySetWingedMessenger": "Winged Messenger Set",
"mysterySetForestWalker": "Forest Walker Set",
// ...
```

You could even do:

``` json
"mysterySet201402": "Winged Messenger Set",
"mysterySet201403": "Forest Walker Set",
// etc
```

And then you could remove the text attribute from the objects, and just loop over the mystery sets object (line 130-132) to set each text attribute to be:

```
// top of the file
let t = require('./translation.js');

// line 130
each(mysterySets, (value, key) => {
  value.key = key;
  value.text = t(`mysterySet${key}`);
});
```

website/views/options/inventory/time-travelers.jade
```diff
@@ -29,7 +29,7 @@
                 popover-placement='right', popover-append-to-body='true',
                 ng-click='clickTimeTravelItem("#{type}",item)')
       li.customize-menu.inventory-gear
-        menu.pets-menu(label='{{::set.text}}', ng-repeat='set in Content.timeTravelerStore(user.items.gear.owned)')
+        menu.pets-menu(label=env.t('{{::set.text}}'), ng-repeat='set in Content.timeTravelerStore(user.items.gear.owned)')
```
crookedneighbor: If you set the translation function in the content object, then all you need to do is: `label='{{::set.text()}}'`

common/locales/en/subscriber.json
Alys: +++1 for the "mysterySet201402" format. Much preferred when working out problems related to people not getting their items.

Alys: Also excellent for sorting, which could be useful.

NathanMaton:
Thanks for the comments (this is my first PR so that is why a few things may be off).

The package seemed to come from another update on the /develop branch when I pulled it down. Should I just exclude that? For the .gitignore stuff I can exclude that.

For the code what's the right way to call the env.t command in the javascript file? I saw elsewhere in the app we use window.env.t('') but when I tried that on the mystery-sets.js file it didn't run locally.

common/script/content/mystery-sets.js
```diff
@@ -129,6 +132,7 @@ let mysterySets = {
 
 each(mysterySets, (value, key) => {
   value.key = key;
+  value.text = t('mysterySet${key}'');
```
crookedneighbor: These should be backticks (`), not single quotes ('). See http://babeljs.io/docs/learn-es2015/#template-strings

You also have an extra single quote in there.

package.json
```diff
@@ -43,6 +43,7 @@
     "in-app-purchase": "^0.2.0",
     "jade": "~1.11.0",
     "js2xmlparser": "~0.1.2",
+    "latest-version": "^2.0.0",
```
crookedneighbor: This should be removed.

crookedneighbor:
Correct the syntax error, remove the extraneous package from the package.json and removed the commented out code, and this should be good to go.

common/script/content/mystery-sets.js
NathanMaton: Thanks for the note!

NathanMaton:
Put in requested edits. Build still failing locally for me but out of time to debug tonight.

crookedneighbor:
Merged in with https://github.com/HabitRPG/habitrpg/commit/35a921bf43a1d33b0606e622d5ab8fd809d834b1

NathanMaton:
Thanks Blade. UiD is 2ab55d89-a3e5-4565-ba35-b6c9e145a60a

Definitely hope to contribute more.

crookedneighbor:

NathanMaton:
Thanks!

PR #6369, opened 2015-12-14
Collar icon fix

cassidypignatello:
Hi, I thought I'd open this PR for some feedback. After I updated the images and ran `npm run sprites`, some of the icons (such as eggs and quests) aren't showing up at all after visiting the Inventory page. I figured the CSS for the spritesheets wouldn't need to be updated if you're just replacing the same image, but please let me know if I'm over-simplifying things. Thanks!


cassidypignatello:
Hm, I see..it makes changes to the stylesheets based on the compile task itself. Should I just roll back my changes to when I updated the icons instead?

crookedneighbor:
Sorry it has taken me so long to respond. 

You are correct, only the icons need to be adjusted. Though, looking at the diff for the collar icons, they don't look like they are quite the right size.  

screen shot 2015-12-26 at 9 49 57 am

The new icons _should_ be the exact same dimensions and position of the old icons so they sit on the correct position on the avatar.

crookedneighbor:
I think the big problem is that you replaced the main collar sprite, and not the icon version. They're in this folder: https://github.com/HabitRPG/habitrpg/tree/develop/common/img/sprites/spritesmith/gear/events/wondercon/shop

@cassidypignatello Will you be able to replace the correct icons?

cassidypignatello:
@crookedneighbor thank you so much for clarifying, this makes much more sense now. I'll definitely be able to replace those icons sometime today!

cassidypignatello:
Just wanted to provide a quick update and mention that I'm currently working through a login issue with my local installation before I swap the icons. Worst case scenario, I might just have to re-clone the repo. Either way, I should have this settled pretty soon.

crookedneighbor:
Closed with https://github.com/HabitRPG/habitrpg/pull/6477

Issue #5801, opened 2015-08-16
[Vagrant] Error on 'npm start' after vagrant provisioning.


3onyc:
After running `vagrant up` and following instructions to start the server I get the following errorss:

``` bash
$ npm start

> habitrpg@0.0.0-152 start /vagrant
> grunt run:dev

Loading "browserify.js" tasks...ERROR
>> Error: Cannot find module 'concat-map'
Loading "uglify.js" tasks...ERROR
>> Error: Cannot find module 'ansi-regex'
Loading "cssmin.js" tasks...ERROR
>> Error: Cannot find module 'pako/lib/zlib/messages'
Loading "nodemon.js" tasks...ERROR
>> Error: Cannot find module 'minimatch'
Loading "watch.js" tasks...ERROR
>> Error: Cannot find module 'graceful-fs'
Loading "grunt-spritesmith.js" tasks...ERROR
>> Error: Cannot find module 'mustache'
Loading "imagemin.js" tasks...ERROR
>> Error: Cannot find module 'process-nextick-args'
Loading "grunt-karma.js" tasks...ERROR
>> Error: Cannot find module 'socket.io'
Warning: Task "browserify" not found.# Use --force to continue.

Aborted due to warnings.

npm ERR! habitrpg@0.0.0-152 start: `grunt run:dev`
npm ERR! Exit status 3
npm ERR! 
npm ERR! Failed at the habitrpg@0.0.0-152 start script.
npm ERR! This is most likely a problem with the habitrpg package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     grunt run:dev
npm ERR! You can get their info via:
npm ERR!     npm owner ls habitrpg
npm ERR! There is likely additional logging output above.
npm ERR! System Linux 3.2.0-23-generic
npm ERR! command "/usr/bin/node" "/usr/bin/npm" "start"
npm ERR! cwd /vagrant
npm ERR! node -v v0.10.37
npm ERR! npm -v 1.4.28
npm ERR! code ELIFECYCLE
npm ERR! 
npm ERR! Additional logging details can be found in:
npm ERR!     /vagrant/npm-debug.log
npm ERR! not ok code 0
```

Running `npm install` gives the following log:

``` bash
npm WARN package.json icalendar@0.6.4 No repository field.
npm WARN package.json karma-phantomjs-launcher@0.1.4 No README data
npm WARN package.json expect.js@0.2.0 No repository field.
[...]

PR #6375, opened 2015-12-16
Fix npm install error - Phantomjs dependency

mfdeveloper:
Like suggested by @Alys into issue #5801, was added the dependency `phantomjs` on `package.json` file. This problems happens in latest versions of **Karma** and **Node.js**

My User ID in habitica service is: _fa87aaf3-076e-4084-8f58-8cfa80422bb8_

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,6 +103,7 @@
     "event-stream": "^3.2.2",
     "expect.js": "~0.2.0",
     "istanbul": "^0.3.14",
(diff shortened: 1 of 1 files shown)


SabreCat:
Thanks for this! I've awarded you a contributor tier. Note that while your first tier is easy to get, further levels require increasing amounts of effort from one to the next. But keep helping out and we'll express our thanks accordingly!

PR #6393, opened 2015-12-22
Added chat post route and initial tests

TheHollidayInn:
I'm unsure of as to why, but the Get chat tests are blocking other tests for me. If you comment those out, you can run the tests. 

I have created the chat post and added some tests. Then, I can add the other chat routes.

### DIFF
--- a/test/api/v3/integration/chat/POST-chat.test.js
+++ b/test/api/v3/integration/chat/POST-chat.test.js
@@ -1,0 +1,98 @@
+import {
+  generateUser,
+  requester,
+  translate as t,
+} from '../../../../helpers/api-integration.helper';
+
+describe('POST /chat', () => {
+  let user;
+  let api;
+
+  before(() => {
+    return generateUser().then((generatedUser) => {
+      user = generatedUser;
+      api = requester(user);
+    });
+  });
+
+  it('Returns an error when no message is provided', () => {
+    let groupName = 'Test Guild';
+    let groupType = 'guild';
+    let groupPrivacy = 'public';
+    let testMessage = '';
+    let api2;
+
+    return generateUser({balance: 1}).then((generatedUser) => {
+      api2 = requester(generatedUser);
+      return api2.post('/groups', {
+        name: groupName,
+        type: groupType,
+        privacy: groupPrivacy,
+      });
+    })
+    .then((group) => {
+      return expect(api.post(`/groups/${group._id}/chat`, { message: testMessage}))
+        .to.eventually.be.rejected.and.eql({
+          code: 400,
+          error: 'BadRequest',
+          message: t('invalidReqParams'),
+        });
+    });
+  });
+
+  it('Returns an error when group is not found', () => {
+    let testMessage = 'Test Message';
+    return expect(api.post('/groups/nvalidID/chat', { message: testMessage})).to.eventually.be.rejected.and.eql({
+      code: 404,
+      error: 'NotFound',
+      message: t('groupNotFound'),
+    });
+  });
+
+  it('Returns an error when chat privileges are revoked', () => {
+    let groupName = 'Test Guild';
+    let groupType = 'guild';
+    let groupPrivacy = 'public';
+    let testMessage = 'Test Message';
+    let api2;
+
+    return generateUser({balance: 1, 'flags.chatRevoked': true}).then((generatedUser) => {
+      api2 = requester(generatedUser);
+      return api2.post('/groups', {
+        name: groupName,
+        type: groupType,
+        privacy: groupPrivacy,
+      });
+    })
+    .then((group) => {
+      return expect(api2.post(`/groups/${group._id}/chat`, { message: testMessage})).to.eventually.be.rejected.and.eql({
+        code: 404,
+        error: 'NotFound',
+        message: 'Your chat privileges have been revoked.',
+      });
+    });
+  });
+
+  it('creates a chat', () => {
+    let groupName = 'Test Guild';
+    let groupType = 'guild';
+    let groupPrivacy = 'public';
+    let testMessage = 'Test Message';
+    let api2;
+
+    return generateUser({balance: 1}).then((generatedUser) => {
+      api2 = requester(generatedUser);
+      return api2.post('/groups', {
+        name: groupName,
+        type: groupType,
+        privacy: groupPrivacy,
+      });
+    })
+    .then((group) => {
+      return api2.post(`/groups/${group._id}/chat`, { message: testMessage});
+    })
+    .then((result) => {
+      expect(result.message.id).to.exist;
+    });
+  });
+});
--- a/website/src/controllers/api-v3/chat.js
+++ b/website/src/controllers/api-v3/chat.js
@@ -39,4 +39,60 @@ api.getChat = {
   },
 };
 
+/**
+ * @api {post} /groups/:groupId/chat Post chat message to a group
+ * @apiVersion 3.0.0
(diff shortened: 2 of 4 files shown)


website/src/controllers/api-v3/chat.js
```diff
@@ -39,4 +39,59 @@ api.getChat = {
...
+    .then((group) => {
+      if (!group) throw new NotFound(res.t('groupNotFound'));
+      if (group.type !== 'party' && user.flags.chatRevoked) {
+        throw new NotFound('Your chat privileges have been revoked.');
+      }
+
+      let lastClientMsg = req.query.previousMsg;
+      chatUpdated = lastClientMsg && group.chat && group.chat[0] && group.chat[0].id !== lastClientMsg ? true : false;
+
+      group.sendChat(req.query.message, user);  // FIXME this should be body, but ngResource is funky
```
paglias: you have to use the message passed in the body here

website/src/controllers/api-v3/chat.js
```diff
@@ -39,4 +39,59 @@ api.getChat = {
...
+    if (validationErrors) return next(validationErrors);
+
+    Group.getGroup(user, groupId)
+    .then((group) => {
+      if (!group) throw new NotFound(res.t('groupNotFound'));
+      if (group.type !== 'party' && user.flags.chatRevoked) {
+        throw new NotFound('Your chat privileges have been revoked.');
+      }
+
+      let lastClientMsg = req.query.previousMsg;
```
paglias, line 80: this param should be documented in the documentation comment at the top

test/api/v3/integration/chat/POST-chat.test.js
```diff
@@ -0,0 +1,102 @@
...
+    it('Returns an error when group is not found', () => {
+      let testMessage = 'Test Message';
+      return expect(api.post('/groups/' + 'invalidID' + '/chat', { message: testMessage})).to.eventually.be.rejected.and.eql({
+        code: 404,
+        error: 'NotFound',
+        message: t('groupNotFound'),
+      });
+    });
+
+    it('Returns an error when chat privileges are provoked', () => {
```
crookedneighbor: Revoked, not provoked.

test/api/v3/integration/chat/POST-chat.test.js
```diff
@@ -0,0 +1,102 @@
...
+  let user, api;
+
+  before(() => {
+    return generateUser().then((generatedUser) => {
+      user = generatedUser;
+      api = requester(user);
+    });
+  });
+
+  context("Describe Creating", () => {
```
crookedneighbor: This context block doesn't seem useful to me.

website/src/controllers/api-v3/chat.js
TheHollidayInn: Yea.... that's right. :P

TheHollidayInn:
Thanks! I have made the changes you requested. Also, I made some changes per eslint and I replaced the concatenation with template strings. eslint doesn't like our global use of expect, by the way :(

TheHollidayInn:
I have the like route and tests ready as well. Should I add those here or wait for the PR and rebase?

paglias:
@TheHollidayInn add them here! I'm sorry for the delay

TheHollidayInn:
Added! I'll have flags this weekend.

website/src/controllers/api-v3/chat.js
```diff
@@ -39,4 +40,113 @@ api.getChat = {
...
+      if (message.uuid === user._id) throw new NotFound(res.t('messageGroupChatLikeOwnMessage'));
+
+      if (!message.likes) message.likes = {};
+      if (message.likes[user._id]) {
+        delete message.likes[user._id];
+      } else {
+        message.likes[user._id] = true;
+      }
+
+      let messageIndex = group.chat.indexOf(message);
```
paglias: this and the next line aren't really necessary, you already changed the message object above at lines 132-137.

paglias:
ok! Except for the comment I left this seems quite good. At this point I'm going to wait for flags before merging

website/src/controllers/api-v3/chat.js
```diff
@@ -39,4 +40,113 @@ api.getChat = {
...
+
+      group.sendChat(req.body.message, user);
+
+      if (group.type === 'party') {
+        user.party.lastMessageSeen = group.chat[0].id;
+        user.save();
+      }
+      return group.save();
+    })
+    .then((group) => {
```
crookedneighbor, line 91: Can you explain a little what is happening here? Why are we sending back different objects?

website/src/controllers/api-v3/chat.js
```diff
@@ -39,4 +40,113 @@ api.getChat = {
...
+      }
+
+      let messageIndex = group.chat.indexOf(message);
+      group.chat[messageIndex].likes = message.likes;
+
+      return group.save();
+    })
+    .then((group) => {
+      if (!group) throw new NotFound(res.t('groupNotFound'));
+      res.respond(200, group.chat);
```
crookedneighbor: Should we be sending back the entire group.chat array for liking a message? Or just the message?

website/src/controllers/api-v3/chat.js
TheHollidayInn, line 91: I'm unsure to be honest. This was just ported: https://github.com/HabitRPG/habitrpg/blob/develop/website/src/controllers/api-v2/groups.js#L313 .

website/src/controllers/api-v3/chat.js
TheHollidayInn: Just porting here as well: https://github.com/HabitRPG/habitrpg/blob/develop/website/src/controllers/api-v2/groups.js#L455 . I think it is more efficient to change as well. Let me know.

website/src/controllers/api-v3/chat.js
TheHollidayInn: Gotcha. I'll test this out and make an update.

website/src/controllers/api-v3/chat.js
paglias, line 91: basically if you pass the id of the last message you have on the client using the `lastClientMessage` query param it'll return the whole chat if a message was posted in the meantime, otherwise just the new message

website/src/controllers/api-v3/chat.js
paglias: Agree with @crookedneighbor , there's no point in returning the whole chat

TheHollidayInn:
Okay, I made updates, but I realized after adding a group query to the like test, that the likes aren't actually being saved. So, I am trying to debug that now. I have the flag test, but the flags also aren't being saved.

TheHollidayInn:
Okay, basically there was a problem with using group.save(). I could have fixed it with this: http://stackoverflow.com/questions/24618584/mongoose-save-not-updating-value-in-an-array-in-database-document, but I decided to make the change to Group.update().

TheHollidayInn:
@paglias changing all the likes was some debugging code. It was unnecessary as you pointed out before, so I removed the code. Everything should be good now.

TheHollidayInn:
Alright, have a look at this. I apologize if the query is slightly verbose, but I was having issues with concatenating and the positional operator only goes one level in Mongodb.

paglias:
@TheHollidayInn except for the little things eslint is complaining it looks very good!

TheHollidayInn:
Okay, I changed the syntax a bit to be consistent. Either way, I made the updated the tests!

PR #6394, opened 2015-12-22
Fix e2e test in vagrant environment

KristianTashkov:
I saw a comment of @crookedneighbor mentioning that e2e tests don't work in a vagrant environment. I've been investigating why it crashes/hangs. I have found several issues causing it:
1. The xvfb package is not installed and it's part of the e2e test suite preparations
2. Protractor raises a selenium server which needs Java 7. We didn't have it installed before.
3. We don't have Firefox installed for the e2e tests and not all versions play nice with the selenium server. I managed to find that Firefox 40 is one of the newer versions that work. Since you can't apt-get a previous release we need to wget it and install it via dpkg.
4. Part of the e2e preparations are to set the DISPLAY environment variable, but they don't seem to propagate to the selenium server. I checked how the travis CI build passed despite this problem and it seems it exports it in a before script. So I added it to the .bashrc file. A better way would be to fix the test to set the DISPLAY variable properly, but I can't seem to find a way and setting the DISPLAY variable for a vagrant ssh'ed environment without a display is not a bad solution. I'm open to ideas.

The proposed fix installs all the needed dependencies and environment variables for the test to work.

### DIFF
--- a/vagrant.sh
+++ b/vagrant.sh
@@ -26,6 +26,8 @@ function autostart_habitrpg {
 # Main provisioning
 echo Setting up Habitica...
 echo cd /vagrant >> /home/vagrant/.bashrc
+# Needed for running e2e tests
+echo export DISPLAY=:99 >> /home/vagrant/.bashrc
 
 # Prevent warnings: "dpkg-preconfigure: unable to re-open stdin ..."
 export DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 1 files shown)


crookedneighbor:
This works beautifully. 

Note: The karma tests still don't work in a vagrant setup, but since that was the case before, I'm going to merge this in now.

KristianTashkov:
User ID: bb8db09b-5822-4608-bba3-1486964b8537
Thanks :)

crookedneighbor:

KristianTashkov:
Thanks!

I made some changes in the wiki for setting Habitica locally to reflect the changes in this PR. Removed the line that said you need to install java to run tests and added a note on the DISPLAY=:99 in .bashrc :)

crookedneighbor:
:+1:

Issue #5182, opened 2015-05-10
Allow cancel editing group infobox


SabreCat:
When editing the description of a Guild, there is no obvious way to cancel editing without saving. There's a "Save" button, the "Leave" button for quitting the Guild (which really shouldn't be visible at all at that point), but no way to back out of the edit other than hitting the Back or Refresh button on the browser.

Suggestions:
- Add a Cancel button. A good place for it might be where the "Leave" button now appears.
- Allow usage of the Esc key to bail out.

Using either of the above should discard all changes and return the group infobox to read-only state.
##

AnitaYCheng:
+1 for the cancel button, for sure! And of course, the Esc key is a good failsafe.

vinsonizer:
I've forked the code and written a simple implementation of this that is working, but I am not sure how to contribute it back (new to git/github).  Can someone give me hints on how to do this and also how to get contributor credit (if possible)?

crookedneighbor:
You can follow this guide to initiate a pull request: https://help.github.com/articles/using-pull-requests/

vinsonizer:
Do i need to do anything with my user id for contribution?

vinsonizer:
Thanks @crookedneighbor, got it worked out (i think).  If i don't get credit it is fine, but it would be neat to get the contributor badge.

Alys:
@vinsonizer When a pull request is accepted and merged, we always ask for the submitter's Habitica User ID and use that for recording the contribution to a tier. You can add your User ID to your pull request now if you wish, although be aware that not all requests can be accepted (it's rare that one is rejected outright though).

vinsonizer:
Ok, I am modifying it to use a angular.copy for a local working copy while editing.  This should be pretty light-weight and it would be all client side till the save button is used.  I will try to update later today and then see if i can figure out how to modify my pull request.

crookedneighbor:
@vinsonizer When you push to your branch, your PR gets updated automatically.

vinsonizer:
Should this issue be closed?

SabreCat:
Ahyes! I thought GitHub would have closed it automatically, but maybe the # tag wasn't in the right place.

vinsonizer:
No problem.  Perhaps on my next one @crookedneighbor won't have to re-write most of my tests.

PR #6400, opened 2015-12-26
Fix for issue #5182: Allow cancel editing group infobox

vinsonizer:
Added cancel button that is available only when editing a group, which reverts changes from the server when clicked.  Also hides the "leave" button when in "Edit Mode"

### DIFF
--- a/test/spec/controllers/groupCtrlSpec.js
+++ b/test/spec/controllers/groupCtrlSpec.js
@@ -23,69 +23,150 @@ describe('Groups Controller', function() {
     });
   });
 
-  describe("isMemberOfGroup", function() {
-    it("returns true if group is the user's party retrieved from groups service", function() {
-      var party = specHelper.newGroup({
-        _id: "unique-party-id",
-        type: 'party',
-        members: ['leader-id'] // Ensure we wouldn't pass automatically.
+  describe("Group Membership Behavior", function() {
+
+    var party, guild;
+
+    beforeEach(function() {
+      party = specHelper.newGroup({
+         _id: "unique-party-id",
+         type: 'party',
+         members: ['leader-id'] // Ensure we wouldn't pass automatically.
       });
-
-      var partyStub = sandbox.stub(groups,"party", function() {
-        return party;
-      });
-
-      expect(scope.isMemberOfGroup(user._id, party)).to.be.ok;
-    });
-
-    it('returns true if guild is included in myGuilds call', function(){
-
-      var guild = specHelper.newGroup({
+      guild = specHelper.newGroup({
         _id: "unique-guild-id",
         type: 'guild',
         members: [user._id]
       });
-
-      var myGuilds = sandbox.stub(groups,"myGuilds", function() {
-        return [guild];
-      });
-
-      expect(scope.isMemberOfGroup(user._id, guild)).to.be.ok;
-      expect(myGuilds).to.be.called;
     });
 
-    it('does not return true if guild is not included in myGuilds call', function(){
-
-      var guild = specHelper.newGroup({
-        _id: "unique-guild-id",
-        type: 'guild',
-        members: ['not-user-id']
+    describe("isMemberOfGroup behavior", function() {
+      it("returns true if group is the user's party retrieved from groups service", function() {
+        var partyStub = sandbox.stub(groups,"party", function() {
+          return party;
+        });
+    
+        expect(scope.isMemberOfGroup(user._id, party)).to.be.ok;
+      });
+    
+      it('returns true if guild is included in myGuilds call', function(){
+        var myGuilds = sandbox.stub(groups,"myGuilds", function() {
+          return [guild];
+        });
+    
+        expect(scope.isMemberOfGroup(user._id, guild)).to.be.ok;
+        expect(myGuilds).to.be.called;
+      });
+    
+      it('does not return true if guild is not included in myGuilds call', function(){
+        var myGuilds = sandbox.stub(groups,"myGuilds", function() {
+          return [];
+        });
+    
+        expect(scope.isMemberOfGroup('other-user-id', guild)).to.not.be.ok;
+        expect(myGuilds).to.be.calledOnce;
       });
 
-      var myGuilds = sandbox.stub(groups,"myGuilds", function() {
-        return [];
+      it('does not return true if party members is empty', function(){
+        var partyStub = sandbox.stub(groups, "party", function() {
+          return {};
+        });
+        expect(scope.isMemberOfGroup(user._id, party)).to.not.be.ok;
       });
-
-      expect(scope.isMemberOfGroup(user._id, guild)).to.not.be.ok;
-      expect(myGuilds).to.be.calledOnce;
     });
+
+    /**
+     * Why is this distinct from the isMemberOfGroup method?  Redundant?
+     *
+     * This coercion business is troublesome...
+     */
+    /*
+    describe("isMember behavior", function() { 
+      it("should return true dependent on whether the user id belongs to the group", function() {
+        expect(scope.isMember(user, guild)).to.eql(truthy);
+        expect(scope.isMember(user, party)).to.eql(falsy);
+      });
+    });
+    */
   });
 
+  describe("Quest Membership Behavior", function() { 
+
+    var runningQuestParty, pendingQuestParty;
+
+    beforeEach(function() {
+      runningQuestParty= specHelper.newGroup({
+         _id: "unique-party-id",
+         type: 'party',
+         members: ['leader-id'],
+         quest: {
+           progress: {},
+           active: true,
+           members: {}
+         }
+      });
+      pendingQuestParty= specHelper.newGroup({
+         _id: "unique-party-id",
+         type: 'party',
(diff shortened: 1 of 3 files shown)


website/public/js/controllers/groupsCtrl.js
```diff
@@ -47,6 +47,11 @@ habitrpg.controller("GroupsCtrl", ['$scope', '$rootScope', 'Shared', 'Groups', '
       group._editing = false;
     };
 
+    $scope.cancelSave = function(group) {
+      group = group.$get();
```
crookedneighbor: Does the `$get` method return the group object? I thought `group.$get` returned a promise.

vinsonizer: According to the resource docs (
https://docs.angularjs.org/api/ngResource/service/$resource) it returns an
empty object that is populated on result which will trigger rendering
changes. I guess this could create some interesting client experience, but
I wasn't sure if going with something like a shadow copy directive would
over complicate the change.

Also I am new to Angular and Jade, so there is likely an even better way to
handle this.

vinsonizer:
User ID b57a198e-d079-48ef-9fd9-631d99a628ce by the way

website/public/js/controllers/groupsCtrl.js
crookedneighbor: You shouldn't need to reassign the variable, just doing:

``` js
$scope.cancelSave = function(group) {
  group.$get();
  //...
```

Should work.

That said, I think making a local copy is probably preferable so we don't need to make an extra server call for the cancel activity to work. This will probably mean making an `$scope.editGroup` function.

crookedneighbor: That said, I'm not an expert on client side performance.

Also, thanks for the link. I really need to actually read the docs for ng-resource. :smile:

crookedneighbor:
In addition, we'll need to add tests for the new functions. [the group controller doesn't have a ton of coverage right now](https://github.com/HabitRPG/habitrpg/blob/develop/test/spec/controllers/groupCtrlSpec.js), but we should still add tests for new functionality.

vinsonizer:
I am not familiar with the testing framework used, but I will see if i can get something together that will work.

crookedneighbor:
If you need help, I'm happy to give it.

vinsonizer:
Just need a start, glad to do it if you can point me to references

vinsonizer:
@crookedneighbor I have written some test cases for the groups controller, but i am having issues actually executing the save method (undefined).  Is there some way i can share here without committing into the pull request?

vinsonizer:
Actually, you can see it in this gist, from line 74 on:

https://gist.github.com/vinsonizer/55d5a8cc444d77ccba85

I get errors on the last line of the test, which is the invocation of group.$save() in the controller.

crookedneighbor:
@vinsonizer looks like maybe you rebased and pulled in a merge commit? Could you remove that extraneous commit from this PR so it's easier to review?

vinsonizer:
@crookedneighbor how do I do that?  I've rebased twice trying to ensure i stay up to date, but I am getting duplicate merge conflicts.  I am not sure why that is...

vinsonizer:
@crookedneighbor Cleaning up now, hopefully have it all recommitted today

vinsonizer:
@crookedneighbor  Ok I reset (further back than I meant to) and reapplied my changes.  Please let me know if this is what you are looking for.

vinsonizer:
@crookedneighbor @SabreCat I hate to keep hitting you guys with new commits... I now understand the jasmine library and am restructuring the group controller tests (all of them) into more a modular setup.  Please wait to review the changes till I've completed this.  I think it will be much nicer.

crookedneighbor:
@vinsonizer No worries. Ping me when you're ready for me to look at it.

vinsonizer:
@crookedneighbor I think it is done.  There are only two sections I didn't add tests for: 1.) Modal methods and 2.) isMember, which appears to be using coercion and also basically looks like it should be superseded by isMemberOfGroup.

test/spec/controllers/groupCtrlSpec.js
```diff
@@ -23,51 +23,207 @@ describe('Groups Controller', function() {
...
+      expect(guild.description).to.eql(editGuild.description);
+      expect(guild._newLeader).to.eql(editGuild._newLeader);
+      expect(guildSave).to.be.calledOnce;
 
-      expect(scope.isMemberOfGroup(user._id, guild)).to.not.be.ok;
-      expect(myGuilds).to.be.calledOnce;
     });
   });
+  /* TODO: Modal testing
+  describe("deleteAllMessages", function() { });
```
crookedneighbor: The proper way to pend these is by doing:

``` js
describe.skip('deleteAllMessages', function () {

});
```

test/spec/controllers/groupCtrlSpec.js
```diff
@@ -23,51 +23,207 @@ describe('Groups Controller', function() {
...
+    it('should return false if my user isn\'t part of the quest members', function() {
+      pendingQuestParty.quest.members = {};
+      runningQuestParty.quest.members = {};
+      expect(scope.isMemberOfPendingQuest(user._id, pendingQuestParty)).to.eql(false);
+      expect(scope.isMemberOfRunningQuest(user._id, runningQuestParty)).to.eql(false);
+    });
+  });
+   
+
+  describe("Group Edit Behavior", function() {
```
crookedneighbor: This is describing the `editGroup` method, so it should be `describe('editGroup'`, function () {`

test/spec/controllers/groupCtrlSpec.js
```diff
@@ -23,51 +23,207 @@ describe('Groups Controller', function() {
...
+
```
crookedneighbor: No blank line here.

test/spec/controllers/groupCtrlSpec.js
```diff
@@ -23,51 +23,207 @@ describe('Groups Controller', function() {
...
+    it('should return false if my user isn\'t part of the quest members', function() {
+      pendingQuestParty.quest.members = {};
+      runningQuestParty.quest.members = {};
+      expect(scope.isMemberOfPendingQuest(user._id, pendingQuestParty)).to.eql(false);
+      expect(scope.isMemberOfRunningQuest(user._id, runningQuestParty)).to.eql(false);
+    });
+  });
+   
+
```
crookedneighbor: There's an extra line break here.

test/spec/controllers/groupCtrlSpec.js
```diff
@@ -23,51 +23,207 @@ describe('Groups Controller', function() {
...
+      var editGuild = scope.groupCopy;
+      expect(editGuild).to.eql({});
+      expect(guild._editing).to.eql(undefined);
+    });
+
+
+    it('should allow for editing without changing group resource', function() {
+      // switch to edit mode
+      scope.editGroup(guild);
+      var editGuild = scope.groupCopy;
```
crookedneighbor: Why is it important to assign this to a variable? Why can't you assert on `scope.groupCopy` directly?

website/public/js/controllers/groupsCtrl.js
```diff
@@ -12,7 +12,7 @@ habitrpg.controller("GroupsCtrl", ['$scope', '$rootScope', 'Shared', 'Groups', '
     $scope.isMemberOfRunningQuest = function(userid, group) {
       if (!group.quest || !group.quest.members) return false;
       if (!group.quest.active) return false; // quest is pending, not started
-      return group.quest.members[userid];
+      return userid in group.quest.members && group.quest.members[userid] != false;
```
crookedneighbor, line 15: Why are you working on this code? Seems like this should be a separate PR.

crookedneighbor:
The group leader message box is not part of this PR: https://github.com/HabitRPG/habitrpg/pull/6400/files#diff-093f1d61737be6ef20f580e7dcc66fe1R134

test/spec/controllers/groupCtrlSpec.js
vinsonizer: Which line specifically?  149?

website/public/js/controllers/groupsCtrl.js
vinsonizer, line 15: When adding tests, i found that the behavior of this method and isMemberOfPendingQuest was inconsistent.  This should be functionally equivalent to what was there before but also handles undefined.  It is a safer check overall right?

In reality these two methods should probably be refactored for code reuse, since they are almost identical other than the active check.

vinsonizer:
@crookedneighbor 

I will add the leader message box in later this evening and incorporate your comments.  If you can give specific feedback on the whitespace concerns, i will also correct those.

I already fixed the other comments (re: describe.skip, describe editGroup, local variable assignment)

crookedneighbor:
I've merged in the portion of your pr that deals with the editGroup. If you'd like to open a separate PR with your isMember changes, feel free.

Note: I changed your test to be in a more unit test style instead of a behavioral style: https://github.com/HabitRPG/habitrpg/commit/2989057dd3093247472010acb24d9b88b0905ee2

PR #6404, opened 2015-12-26
Fixed a minor css bug, where both min-width and max-width were set to 768.

proarunas:
Max-width was changed to 767 to prevent overlap, which produces artefacts when
screen width is exactly 768 pixels.

### DIFF
--- a/website/public/css/global-modules.styl
+++ b/website/public/css/global-modules.styl
@@ -1,3 +1,5 @@
+@require "./variables/screen-size.styl"
+
 // Clearfix (http://css-tricks.com/snippets/css/clear-fix/)
 $clearfix
   &:after
@@ -50,7 +52,7 @@ hrpg-button-color-mixin($hrpg-button-color, highlight=false)
     else
       border-left: 1px solid darken($hrpg-button-color, 3.82%) !important
   > div
-    @media screen and (min-width:768px)
+    @media screen and (min-width: $sm-min-screen-width)
       border:1px solid darken($hrpg-button-color, 16.18%)
     h4
       color: darken($hrpg-button-color, 61.8%)
@@ -85,7 +87,7 @@ hrpg-anchor-button-color-mixin($hrpg-button-color)
   background-color: $hrpg-button-color !important
   &:active
     background-color: darken($hrpg-button-color, 61.18%) !important
-  @media screen and (min-width:768px)
+  @media screen and (min-width: $sm-min-screen-width)
     &:hover
       background-color: darken($hrpg-button-color, 2.36%) !important
 $hrpg-button-master
@@ -221,7 +223,7 @@ $hrpg-modal-dropdown
     min-width:110px
     border-radius:4px
     background-color:#fff
-    @media screen and (max-width:767px)
+    @media screen and (max-width: $xs-max-screen-width)
       position:fixed
       z-index: 1000
       top: 2%
@@ -249,12 +251,12 @@ $hrpg-modal-dropdown
       overflow: auto
       border-top-width: 1px
       border-top-style: solid
-      @media screen and (min-width:768px)
+      @media screen and (min-width: $sm-min-screen-width)
         max-height: 320px
         max-height: 62.8vh
       li
         width: 100%
-        @media screen and (max-width:767px)
+        @media screen and (max-width: $xs-max-screen-width)
           width: auto
       a
         display: inline-block
@@ -277,14 +279,14 @@ $hrpg-modal-dropdown
       transform: rotate(45deg)
 $hrpg-modal-dropdown-left
   @extend $hrpg-modal-dropdown
-  @media screen and (min-width:768px)
+  @media screen and (min-width: $sm-min-screen-width)
     div
       right: 0px
       > ul:first-child:before, > div:first-child:before, h4:before
         right: .5em
 $hrpg-modal-dropdown-right
   @extend $hrpg-modal-dropdown
-  @media screen and (min-width:768px)
+  @media screen and (min-width: $sm-min-screen-width)
     div
       left: 0px
       > ul:first-child:before, > div:first-child:before, h4:before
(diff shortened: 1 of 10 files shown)


crookedneighbor:
Can that number be pulled out into a variable instead?

proarunas:
Well. 768 is bootstrap constant for switching between mobile and tablet UI, it is hardcoded in a number of .styl files.
One could extract it, two other constants like this (992 and 1200 which are used once in habitica) and the derived 767 into separate variables. Not sure where would be the best place to put them though.

proarunas:
Alternative fix with variables would look something like this:
https://github.com/proarunas/habitrpg/commit/afbb4eb5299670865001ddd2c1b3a062d0dc0ae6
If that is preferable.

crookedneighbor:
I would create a folder call variables and make a file in it called screen-size.styl

Can we prexif all the variables with a `$`? That's more of a sass convention, but it's valid in stylus, and helps to have a visual declaration that it is a variable.

I'd also prefer more verbose variable names: 

```
$xs-max-screen-size = 767px
$sm-min-screen-size = 768px
$md-min-screen-size = 992px
$lg-min-screen-size = 1200px
```

proarunas:
Done. Also have changed the import statements to require statement for less checking. (It might still be desirable to simply import the variables folder instead, but it does not really matter at this point, when there is only one variable file)

The screen window is still checked in two places that are not covered by these variables:
1) .avatar-window class in customizer.styl, which is not used by anything afaict.
2) .task-column class in tasks.styl to enable scrollbars for tablet-sized screens. This one makes navigation painful when tiling the browser across half of the screen (I imagine it's even worse for tablets). Addressing this was my original reason for messing with habitrpg.

I am inclined to remove them both before the merge (if not I will open another issue/merge for the second one at least).

crookedneighbor:
I'm struggling to understand what the problem is, precisely. Could you take some before and after screenshots to illustrate the issue?

proarunas:
Actually there is no need for before/after as the threshold is arbitrary and the effect can be seen from simply resizing the window.

If the window width between 971 and 992px - the tasks panes are displayed in two columns:
capture1

If the width is between 768 and 970 (tiling the window across half of the 1080p screen is ~952px), the task panes are still displayed in two columns, but they are restricted to a very small size and scrollbars are added to them:
capture2

The mileage for different resolutions/dpi will vary quite a bit as the size of panes is defined in em and not px, but I find it almost impossible to use habitica when tiling my screens (which I do all the time).

My suggestion was to remove the special case for adding the scrollbars. It would make the UI more uniform as well. 
If you disagree with removing the scrollbars entirely, I could instead increase the threshold so that they always appear with two column layout (and hack up my own browser to remove those pesky scrollbars client-side :)), or reduce the threshold <950 to fix the ui for those that use window tiling with 1080p resolution (which is very hacky approach imho).

In either case, I've tested the removal of scrollbars and noticed, that tiling has to be fixed first (if you have more habits than dailies it messes up), so that will have to be another merge anyway.

Also there is another issue with resizing (menu bar does not play nicely with it), but again that is yet another issue entirely, that I plan to tackle later.

P.S. Can still remove the unused avatar-window css class and/or the associated arbitrary window width checks for this merge.

SabreCat:
Additional scrollbars within scrolling regions are pretty terrible, so I wouldn't mind taking those out. Is the tiling fix something you could include in this pull request? It'd be preferable to do a complete fix rather than a partial one that introduces an additional problem.

proarunas:
Well, at this point this pull request simply fixes the tiny bug (where width = 768) and slightly re-factors the css files (as per Blade's request). I have not committed the removal of scrollbars as of yet.

I am looking into the tiling issue, but I need to familiarise myself with angular first. I can just create another pull request with removed scrollbars and fixed tiling when I am done with that. It's up to you.

proarunas:
The tiling problem turned out to be simply a standard issue with CSS floats. Added a class to "todo" list, which activates only on medium screen width and clears the floating for it, which fixes it.
Removed the scrollbars in the medium width view, as now it's safe to do so afaict.

SabreCat:
Attempting to access the site running this branch throws the following error:

> error: TypeError: /home/sabrecat/habitica/website/views/shared/avatar/appearance.jade:5
>    3| mixin avatar(opts)
>    4|   .character-sprites
>  > 5|     .addthis_native_toolbox(ng-if='::profile._id==user._id',
>    6|       data-url="#{env.BASE_URL}/static/front/#?memberId={{::profile._id}}",
>    7|       data-title=env.t('checkOutProgress'))
>    8|
> Cannot read property 'indexOf' of undefined

The same error does not occur when running against `develop`. Possibly the variables aren't getting included (need to add the file to `index.styl`?), or an apparently unused style actually was in use somewhere. `addthis` is rather a strange place to run into issues, but something's not right here in any case!

proarunas:
All is fine.

It's just that npm/db is no longer compatible between the branches. You have to actually merge in `develop`, if you want to test it with your (supposedly up-to-date-with-develop) npm installation. 
(No changes to those in these commits, it's all in `develop`.) I imagine unit tests would catch something that big in the first place :).

P.S. have not removed any of unused styles and it's all `CSS`, which has high tolerance for missing classes. For variables, the `requires` statements are placed in all files that use them, so adding them to `index.styl` would be redundant.

SabreCat:
@proarunas: OK, looks good! This will count toward a contributor tier. What's your Habitica User ID (from https://habitica.com/#/options/settings/api, or Settings in the mobile apps)?

proarunas:
8afffbf1-851e-4cef-92ba-9d79a49d4347 
Thanks.

SabreCat:
@proarunas: Tier awarded, congratulations!

Note that while your first tier is easy to get, subsequent tiers require increasing effort from one to the next. So reaching Tier 2 may take more than one PR, or a larger PR. But keep helping out and we'll express our thanks accordingly!

Alys:
Regarding the scroll bars that were removed in this PR, I think this comment in the Tavern from yesterday might be relevant:

"Oh no, did something change in the page layout? I used to be able to make Habitica not full screen and get four boxes with scroll bars, to line up my spells and the tasks I wanted to do the spells on without having to scroll up and down all the time, but now the little boxes no longer seem to have scroll bars."

Being able to use the scroll bars like that to make skill casting more convenient seems like a really good idea to me. What problems would occur if the scroll bars were put back?

SabreCat:
Additional scrollbars within a bigger scrolling region look really bad and are hard to work with, in the bigger picture of usability. I guess we could put it back, but I feel like the 80/20 yardstick suggests favoring general task interaction over spellcasting.

proarunas:
Having a mode it is more easy to use spells is nice (and there are some related possible improvements to spell usage in Trello). But making user interface almost unusable at certain widths again, just to act as a work-around for clunky spell casting sounds like a rather bad idea.

PR #6410, opened 2015-12-28
Screen reader labels

serverless-mom:
# what

Add a bunch of Aria-labels to various shop and equipment items
# why

User mentions that much of the site is unlabeled for screen-readers [on a Trello card mainly concerned with the chat ordering](https://trello.com/c/pxTcbKuD/381-accessibility-settings) and sure enough when using a screen readers several areas are impossible to navigate.

If your curious about how widely aria-label works as an option for screen-readers, check out [this handy article](https://www.nczonline.net/blog/2013/04/01/making-accessible-icon-buttons/) that does some testing with multiple browsers.
# known issues

In all cases the label is drawn from the popup title, and in a couple of instances there might be a more useful string to add as the label. Right now for example it doesn't read how many of an item you have, and I don't think the badge number will be picked up by any screen readers. When I tried grabbing the badge values... everything broke so I rolled that back.
# test

0)  turn on a screen-reading program (on OSX you can do this with CMD+f5) and see that there are no labels on many equipment and inventory items, they just say things like 'potion button' and 'equipment button'
1) switch to branch
2) go to the two pages I modified http://localhost:3000/#/options/inventory/drops and http://localhost:3000/#/options/inventory/equipment and see that nothing is messed up on these pages for a sighted user
3)  turn on a screen-reading program (on OSX you can do this with CMD+f5) and see those same things have labels!
3a) if you aren't familiar with screen readers (it took me some time to get in control of the Apple version), just inspect the buttons and see they now have reasonable 'aria-label's attached.

### DIFF
--- a/website/views/options/inventory/drops.jade
+++ b/website/views/options/inventory/drops.jade
@@ -9,12 +9,10 @@
           menu.pets-menu(label=(env.t('eggs') + ' ({{eggCount}})'))
             p.muted(ng-show='eggCount < 1')=env.t('noEggs')
             div(ng-repeat='(egg,points) in ownedItems(user.items.eggs)')
-              - var title = env.t("egg", {eggType: "{{::Content.eggs[egg].text()}}"})
               button.customize-option(class='Pet_Egg_{{::egg}}',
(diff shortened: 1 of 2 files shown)


serverless-mom:
I figured out a better way to do this. New PR inbooooound.

PR #6414, opened 2015-12-29
Api v3 groups remaining routes

TheHollidayInn:
Working on more group routes here. I also added some minor fixes for linting and for the test stops, but I can remove the test stop since it has been fixed. 

Also, I realized that I needed the invite and join routes for the next update test, so I marked it as pending and will pick up there tomorrow. Feel free to look over the code until then.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -119,9 +119,6 @@
         "mocha": true,
         "node": true
     },
-    "globals": {
-        "expect": true,
-    },
     "ecmaFeatures" : {
         "modules": true
     },
--- a/.eslintrc
+++ b/.eslintrc
@@ -119,6 +119,9 @@
         "mocha": true,
         "node": true
     },
+    "globals": {
+        "expect": true,
+    },
     "ecmaFeatures" : {
         "modules": true
     },
--- a/test/api/v3/integration/groups/GET-groups.test.js
+++ b/test/api/v3/integration/groups/GET-groups.test.js
@@ -1,22 +1,12 @@
 import {
   generateUser,
-  translate as t,
 } from '../../../../helpers/api-integration.helper';
 
-describe('GET /group', () => {
+describe('GET /groups', () => {
   let user;
 
   beforeEach(async () => {
     user = await generateUser({balance: 4});
-  });
-
-  it('returns an error when group is not found', async () => {
-    await expect(user.get('/groups/fakeGroupId', {}))
-    .to.eventually.be.rejected.and.eql({
-      code: 404,
-      error: 'NotFound',
-      message: t('groupNotFound'),
-    });
   });
 
   it('returns a list of groups', async () => {
@@ -57,57 +47,5 @@ describe('GET /group', () => {
 
     let groupsFound = await user.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
     expect(groupsFound.length).to.be.greaterThan(3);
-  });
-
-  context('Guilds', () => {
-    context('public guild', () => {
-      it('returns a group', async  () => {
-        let groupName = 'Test Public Guild';
-        let groupType = 'guild';
-        let createdGroup = await  user.post('/groups', {
-          name: groupName,
-          type: groupType,
-        });
-        let groupFound = await user.get(`/groups/${createdGroup._id}`);
-
-        expect(groupFound._id).to.exist;
-        expect(groupFound.name).to.equal(groupName);
-        expect(groupFound.type).to.equal(groupType);
-      });
-    });
-
-    context('private guild', () => {
-      it('returns a group', async () => {
-        let groupName = 'Test Private Guild';
-        let groupType = 'guild';
-        let groupPrivacy = 'private';
-        let createdGroup = await user.post('/groups', {
-          name: groupName,
-          type: groupType,
-          privacy: groupPrivacy,
-        });
-        let groupFound = await user.get(`/groups/${createdGroup._id}`);
-
-        expect(groupFound._id).to.exist;
-        expect(groupFound.name).to.equal(groupName);
-        expect(groupFound.type).to.equal(groupType);
-      });
-    });
-  });
-
-  context('Parties', () => {
-    it('returns a group', async () => {
-      let groupName = 'Test Party';
-      let groupType = 'party';
-      let createdGroup = await user.post('/groups', {
-        name: groupName,
-        type: groupType,
-      });
-      let groupFound = await user.get(`/groups/${createdGroup._id}`);
-
-      expect(groupFound._id).to.exist;
-      expect(groupFound.name).to.equal(groupName);
-      expect(groupFound.type).to.equal(groupType);
-    });
   });
 });
--- a/test/api/v3/integration/groups/GET-groups.test.js
+++ b/test/api/v3/integration/groups/GET-groups.test.js
@@ -36,12 +36,24 @@ describe('GET /group', () => {
       },
     ];
 
-    for (let group of groups) {
-      let tmpGroup = await user.post('/groups', {
-        name: group.name,
-        type: group.type,
-      });
-    }
+    let group1 = await user.post('/groups', {
+      name: groups[0].name,
+      type: groups[0].type,
+    });
+
+    let group2 = await user.post('/groups', {
+      name: groups[1].name,
+      type: groups[1].type,
+    });
+
+    let group3 = await user.post('/groups', {
+      name: groups[2].name,
+      type: groups[2].type,
+    });
+
+    expect(group1).toExist;
+    expect(group2).toExist;
+    expect(group3).toExist;
 
     let groupsFound = await user.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
     expect(groupsFound.length).to.be.greaterThan(3);
@@ -49,7 +61,7 @@ describe('GET /group', () => {
 
   context('Guilds', () => {
     context('public guild', () => {
-      it('returns a group',async  () => {
+      it('returns a group', async  () => {
         let groupName = 'Test Public Guild';
         let groupType = 'guild';
         let createdGroup = await  user.post('/groups', {
--- a/test/api/v3/integration/groups/GET-groups.test.js
+++ b/test/api/v3/integration/groups/GET-groups.test.js
@@ -1,0 +1,129 @@
+import {
+  generateUser,
+  requester,
+  translate as t,
+} from '../../../../helpers/api-integration.helper';
+
+describe('GET /group', () => {
+  let user;
+  let api;
+
+  beforeEach(() => {
+    return generateUser({balance: 4}).then((generatedUser) => {
+      user = generatedUser;
+      api = requester(user);
+    });
+  });
+
+  it('returns an error when group is not found', () => {
+    return expect(api.get('/groups/fakeGroupId', {}))
+    .to.eventually.be.rejected.and.eql({
+      code: 404,
+      error: 'NotFound',
+      message: t('groupNotFound'),
+    });
+  });
+
+  it('returns a list of groups', () => {
+    let groups = [
+      {
+        name: 'Test Party',
+        type: 'party',
+      },
+      {
+        name: 'Test Public Guild',
+        type: 'guild',
+      },
+      {
+        name: 'Test Private Guild',
+        type: 'guild',
+        privacy: 'private',
+      },
+    ];
+
+    return api.post('/groups', {
+      name: groups[0].name,
+      type: groups[0].type,
+    })
+    .then(() => {
+      return api.post('/groups', {
+        name: groups[1].name,
+        type: groups[1].type,
+      });
+    })
+    .then(() => {
+      return api.post('/groups', {
+        name: groups[2].name,
+        type: groups[2].type,
+        privacy: groups[2].privacy,
+      });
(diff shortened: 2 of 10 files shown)


test/api/v3/integration/groups/PUT-groups.test.js
```diff
@@ -0,0 +1,64 @@
...
+  let api;
+
+  beforeEach(() => {
+    return generateUser({balance: 1}).then((generatedUser) => {
+      user = generatedUser;
+      api = requester(user);
+    });
+  });
+
+  xit('returns an error when a non group leader tries to update', () => {
```
paglias: @TheHollidayInn why is this test skipped?

TheHollidayInn: I just stopped here for the day, and posted it in case you wanted to review. I'm about to reset and force push to overwrite the commit. 

The reason I stopped is because I realized I needed a second member in the group to try and make an edit. So, I am assuming I need to call the invite and join routes for the second user. But, from Slack, it seems there may be a better way. Do you have any suggestions?

.eslintrc
```diff
@@ -113,6 +113,9 @@
         "mocha": true,
         "node": true
     },
+    "globals": {
```
crookedneighbor: This is not necessary. 

It's checked here: https://github.com/HabitRPG/habitrpg/blob/5304b37102f7210b00f63fdefe4041f8ec08a720/tasks/gulp-eslint.js#L61-L65

Going to move these rules to their own `.eslintrc` in the tests directory soon so a normal `eslint path/to/file` will work correctly.

test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -0,0 +1,129 @@
...
+  generateUser,
+  requester,
+  translate as t,
+} from '../../../../helpers/api-integration.helper';
+
+describe('GET /group', () => {
+  let user;
+  let api;
+
+  beforeEach(() => {
```
crookedneighbor: This should be updated to use `async` and `await`

test/api/v3/integration/groups/PUT-groups.test.js
```diff
@@ -0,0 +1,57 @@
...
+    let groupName = 'Test Public Guild';
+    let groupType = 'guild';
+    let groupToUpdate = await user.post('/groups', {
+      name: groupName,
+      type: groupType,
+    });
+    let groupUpdatedName = 'Test Public Guild Updated';
+    let newUser = await generateUser();
+
+    return user.post(`/groups/${groupToUpdate._id}/invite`, {
```
crookedneighbor: All of these should be updated to use await instead of promise chains.

.eslintrc
TheHollidayInn: Sounds good. I was having this issue when running eslint at the time.

test/api/v3/integration/groups/GET-groups.test.js
TheHollidayInn: Yea, I had these posted before our discussion. I will be updating them today.

TheHollidayInn:
Okay, I made some changes. I still have the PUT reject test pending until I push my invite route. I am almost done with the invite route.

test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -0,0 +1,101 @@
+import {
+  generateUser,
+  translate as t,
+} from '../../../../helpers/api-integration.helper';
+
+describe('GET /group', () => {
```
crookedneighbor: This should be `Get /groups` (plural)

test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -0,0 +1,101 @@
...
+  let user;
+
+  beforeEach(async () => {
+    user = await generateUser({balance: 4});
+  });
+
+  it('returns an error when group is not found', async () => {
+    await expect(user.get('/groups/fakeGroupId', {}))
```
crookedneighbor: This file is for testing `GET /groups` and this test is testing `GET /groups/:id`

test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -0,0 +1,101 @@
...
+  context('Guilds', () => {
+    context('public guild', () => {
+      it('returns a group',async  () => {
+        let groupName = 'Test Public Guild';
+        let groupType = 'guild';
+        let createdGroup = await  user.post('/groups', {
+          name: groupName,
+          type: groupType,
+        });
+        let groupFound = await user.get(`/groups/${createdGroup._id}`);
```
crookedneighbor: All the rest of the tests in this file are testing `GET /groups/:id`. They should be moved to their own test file.

TheHollidayInn:
Alright, I fixed lint errors from the previous tests and moved the group id tests. Then, I added my progress on invites which I will be overwriting with force push soon. But, if you want to take a look at what I am doing so far, feel free.

website/src/controllers/api-v3/groups.js
```diff
@@ -391,145 +392,101 @@ api.removeGroupMember = {
...
-    });
-  }))
-  .then(() => {
-    if (usersAlreadyRegistered.length > 0){
-      return _inviteByUUIDs(usersAlreadyRegistered, group, inviter, req, res, next);
+  if (group.type === 'guild') {
+    if (_.contains(userToInvite.guilds, group._id)) {
+      throw new NotAuthorized(res.t('userAlreadyInGroup'));
     }
+    if (userToInvite.invitations && userToInvite.invitations.guilds && _.find(userToInvite.invitations.guilds, {id: group._id})) {
```
KristianTashkov: Aren't `userToInvite.invitations` and `userToInvite.invitations.guilds` part of the model schema? Can they ever be undefined?

paglias: good point, they'll always be defined

website/src/controllers/api-v3/groups.js
```diff
@@ -563,15 +520,25 @@ api.inviteToGroup = {
...
+      results.push(uuidResults);
+    }
+
+    if (Array.isArray(emails)) {
+      let emailInvites = emails.map((invite) => _inviteByEmail(invite, group, user, req, res));
+      let emailResults = await Q.all(emailInvites);
+      results.push(emailResults);
+    }
+
+    res.respond(200, {result: results});
```
paglias: @TheHollidayInn what are we sending to the client here exactly?

website/src/controllers/api-v3/groups.js
```diff
@@ -4,14 +4,15 @@ import _ from 'lodash';
...
+import { model as EmailUnsubscription } from '../../models/emailUnsubscription';
 import {
   NotFound,
   BadRequest,
   NotAuthorized,
 } from '../../libs/api-v3/errors';
 import * as firebase from '../../libs/api-v3/firebase';
-import { txnEmail } from '../../libs/api-v3/email';
-// import { encrypt } from '../../libs/api-v3/encryption';
+import { sendTxn } from '../../libs/api-v3/email';
```
crookedneighbor: Let's  make this `import { sendTxn as sendTxnEmail }` so it's clear what this function is doing when we use it.

test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -0,0 +1,51 @@
...
+      name: groups[1].name,
+      type: groups[1].type,
+    });
+
+    let group3 = await user.post('/groups', {
+      name: groups[2].name,
+      type: groups[2].type,
+    });
+
+    expect(group1).toExist;
```
crookedneighbor: These three expectations shouldn't be necessary. If there's an error, it'll get thrown and mocha will tell you about it. If it's not actually making  the groups, that's a concern of the post route, not this test.

test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -0,0 +1,51 @@
...
+        type: 'guild',
+      },
+      {
+        name: 'Test Private Guild',
+        type: 'guild',
+        privacy: 'private',
+      },
+    ];
+
+    let group1 = await user.post('/groups', {
```
crookedneighbor: We don't actually care about retaining a reference the the groups here, so why not:

```
groups.forEach(async (groupData) => {
  await user.post('/groups', groupData);
});
```

crookedneighbor: Updated to actually work.

.eslintrc
crookedneighbor: Please remove these.

test/api/v3/integration/groups/GET-groups.test.js
paglias: forEach won't block the main function execution.

Better do `await Q.all(groups.map(group => await user.post(...)))`

crookedneighbor: Good call.

crookedneighbor:
In my opinion, these should be separate PRs

test/api/v3/integration/groups/GET-groups.test.js
TheHollidayInn: Yea, I actually had forEach, but then that was an eslint error. I didn't know how to use Q.all at the time, but I will make the update.

TheHollidayInn:
Yea, I can see how this should be separate PRs. The issue for me is that a lot of the routes and tests depend on each other, so it is hard to start working on the next route without the previous code.

test/api/v3/integration/groups/GET-groups.test.js
paglias: @TheHollidayInn forEach did produce an error because when you use await the direct parent function must be `async function () {}` not a normal `function () {}`

TheHollidayInn: Hmm. I don't exactly remember my code, but the issue was the eslint said not to use await inside of a for loop. I'll take a screenshot if it comes up again, though.

website/src/controllers/api-v3/groups.js
TheHollidayInn: These just return the values that wen through the Qs up top. Basically, these just return the return values from _inviteByUUID and _inviteByEmail which are just the information from uuids, emails. So, it is probably useless. I was using the info to confirm that the information passed.

TheHollidayInn:
Okay, I made the updates.

test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -0,0 +1,35 @@
+import {
```
crookedneighbor, line 1: Shouldn't this test look more like: https://github.com/HabitRPG/habitrpg/blob/8a546c7ec3d41acd606991aca1f78c813f0ff64b/test/api/v2/groups/GET-groups.test.js

Adjusted of course for the way v3 differs from v2.

test/api/v3/integration/groups/GET-groups_id.test.js
```diff
@@ -0,0 +1,73 @@
+import {
```
crookedneighbor, line 1: See https://github.com/HabitRPG/habitrpg/blob/8a546c7ec3d41acd606991aca1f78c813f0ff64b/test/api/v2/groups/GET-groups_id.test.js

test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -0,0 +1,35 @@
...
+  generateUser,
+} from '../../../../helpers/api-integration.helper';
```
crookedneighbor, line 3: This will need to change to ../../../../helpers/api-v3-integration.helper';

test/api/v3/integration/groups/GET-groups_id.test.js
```diff
@@ -0,0 +1,73 @@
...
+  generateUser,
+  translate as t,
+} from '../../../../helpers/api-integration.helper';
```
crookedneighbor, line 4: This will need to change to ../../../../helpers/api-v3-integration.helper';

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,105 @@
+import {
+  generateUser,
+} from '../../../../helpers/api-integration.helper';
```
crookedneighbor, line 3: This will need to change to ../../../../helpers/api-v3-integration.helper';

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -0,0 +1,197 @@
+import {
+  generateUser,
+  translate as t,
+} from '../../../../helpers/api-integration.helper';
```
crookedneighbor, line 4: This will need to change to ../../../../helpers/api-v3-integration.helper';

Also, we need a line break.

test/api/v3/integration/groups/PUT-groups.test.js
```diff
@@ -0,0 +1,55 @@
+import {
+  generateUser,
+  translate as t,
+} from '../../../../helpers/api-integration.helper';
```
crookedneighbor, line 4: This will need to change to ../../../../helpers/api-v3-integration.helper';

PR #6415, opened 2015-12-29
Add user request class for integration tests

crookedneighbor:
I adjusted the user object that gets returned from the `generateUser` method to have the HTTP methods directly on it. This should make writing tests easier, since we no longer need a user && an api object. 

Will be merging this into the api-v3 branch as well shortly.

Pinging @TheHollidayInn @paglias @SabreCat @KristianTashkov since you're working on integration tests as well.

### DIFF
--- a/test/api/README.md
+++ b/test/api/README.md
@@ -4,7 +4,7 @@ That's great! This README will serve as a quick primer for style conventions and
 
 ## What is this?
 
-These are integration tests for the Habitica API. They are performed by making REST requests to the API's endpoints and asserting on the data that is returned.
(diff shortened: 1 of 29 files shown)


KristianTashkov:
Cool :) I will be ready with a PR into api-v3-groups soon. Should I wait for you to merge these changes there and fix it or just submit it when I'm ready and fix it if necessary?

crookedneighbor:
The old requester syntax will still work, so no need to wait.

TheHollidayInn:
Awesome, this is much better than managing a request object as well.

PR #6416, opened 2015-12-30
API v3 Fix basilist scroll giving logic when joining party

KristianTashkov:
While testing the guild code I noticed that `group.memberCount` is not set to 1 when we create a group, so the first member to join (after the creator) becomes the leader, so I fixed it in this commit. I can remove it and add it in a separate PR if you want.

UserID: bb8db09b-5822-4608-bba3-1486964b8537

### DIFF
--- a/test/api/v3/integration/groups/POST-groups_groupId_join.js
+++ b/test/api/v3/integration/groups/POST-groups_groupId_join.js
@@ -1,0 +1,72 @@
+import {
+  generateUser,
+} from '../../../../helpers/api-integration.helper';
+
+describe('POST /group/:groupId/join', () => {
+  let user;
+  let group;
+
+  context('Accepting invitation to a guild', () => {
+    it('does not give basilist quest to inviter when joining a guild', async () => {
+      user = await generateUser({balance: 1});
+      group = await user.post('/groups', {
+        name: 'Test Guild',
+        type: 'guild',
+      });
+
+      let invitedUser = await generateUser({
+        'invitations.guilds': [{ id: group._id}],
+      });
+      await invitedUser.post(`/groups/${group._id}/join`);
+
+      await expect(user.get('/user')).to.eventually.not.have.deep.property('items.quests.basilist');
+    });
+
+    it('does not increment basilist quest count to inviter with basilist when joining a guild', async () => {
+      user = await generateUser({balance: 1, 'items.quests.basilist': 1});
+      group = await user.post('/groups', {
+        name: 'Test Guild',
+        type: 'guild',
+      });
+
+      let invitedUser = await generateUser({
+        'invitations.guilds': [{ id: group._id}],
+      });
+      await invitedUser.post(`/groups/${group._id}/join`);
+
+      await expect(user.get('/user')).to.eventually.have.deep.property('items.quests.basilist', 1);
+    });
+  });
+
+  context('Accepting invitation to a party', () => {
+    it('gives basilist quest item to the inviter when joining a party', async () => {
+      user = await generateUser();
+      group = await user.post('/groups', {
+        name: 'Test Party',
+        type: 'party',
+      });
+
+      let joiningUser = await generateUser({
+        'invitations.party': { id: group._id, inviter: user._id },
+      });
+      await joiningUser.post(`/groups/${group._id}/join`);
+
+      await expect(user.get('/user')).to.eventually.have.deep.property('items.quests.basilist', 1);
+    });
+
+    it('increments basilist quest item count to inviter with basilist when joining a party', async () => {
+      user = await generateUser({'items.quests.basilist': 1 });
+      group = await user.post('/groups', {
+        name: 'Test Party',
+        type: 'party',
+      });
+
+      let joiningUser = await generateUser({
+        'invitations.party': { id: group._id, inviter: user._id },
+      });
+      await joiningUser.post(`/groups/${group._id}/join`);
+
+      await expect(user.get('/user')).to.eventually.have.deep.property('items.quests.basilist', 2);
+    });
+  });
+});
--- a/website/src/controllers/api-v3/groups.js
+++ b/website/src/controllers/api-v3/groups.js
@@ -201,6 +201,7 @@ api.joinGroup = {
   middlewares: [authWithHeaders(), cron],
   async handler (req, res) {
     let user = res.locals.user;
+    let inviter;
 
     req.checkParams('groupId', res.t('groupIdRequired')).notEmpty().isUUID();
 
@@ -213,6 +214,7 @@ api.joinGroup = {
     let isUserInvited = false;
 
     if (group.type === 'party' && group._id === (user.invitations.party && user.invitations.party.id)) {
+      inviter = user.invitations.party.inviter;
       user.invitations.party = {}; // Clear invite TODO mark modified?
 
       // invite new user to pending quest
@@ -242,11 +244,13 @@ api.joinGroup = {
 
     if (group.memberCount === 0) group.leader = user._id; // If new user is only member -> set as leader
 
-    await Q.all([
-      group.save(),
-      user.save(),
-      User.update({_id: user.invitations.party.inviter}, {$inc: {'items.quests.basilist': 1}}).exec(), // Reward inviter
-    ]);
+    let promises = [group.save(), user.save()];
+
+    if (group.type === 'party' && inviter) {
+      promises.push(User.update({_id: inviter}, {$inc: {'items.quests.basilist': 1}}).exec()); // Reward inviter
+    }
+
+    await Q.all(promises);
 
     firebase.addUserToGroup(group._id, user._id);
     res.respond(200, {}); // TODO what to return?
--- a/test/api/v3/integration/groups/POST-groups.test.js
+++ b/test/api/v3/integration/groups/POST-groups.test.js
@@ -45,6 +45,7 @@ describe('POST /group', () => {
           expect(result._id).to.exist;
           expect(result.name).to.equal(groupName);
           expect(result.type).to.equal(groupType);
+          expect(result.memberCount).to.equal(1);
         });
       });
     });
@@ -66,6 +67,7 @@ describe('POST /group', () => {
           expect(result._id).to.exist;
           expect(result.name).to.equal(groupName);
           expect(result.type).to.equal(groupType);
+          expect(result.memberCount).to.equal(1);
           expect(result.privacy).to.equal(groupPrivacy);
         });
       });
@@ -85,6 +87,7 @@ describe('POST /group', () => {
         expect(result._id).to.exist;
         expect(result.name).to.equal(groupName);
         expect(result.type).to.equal(groupType);
+        expect(result.memberCount).to.equal(1);
       });
     });
 
--- a/website/src/models/group.js
+++ b/website/src/models/group.js
@@ -35,7 +35,7 @@ export let schema = new Schema({
     challenges: {type: Boolean, default: false, required: true},
     // invites: {type:Boolean, 'default':false} // TODO ?
   },
-  memberCount: {type: Number, default: 0},
+  memberCount: {type: Number, default: 1},
   challengeCount: {type: Number, default: 0},
   balance: {type: Number, default: 0},
   logo: String,


website/src/controllers/api-v3/groups.js
```diff
@@ -34,6 +34,7 @@ api.createGroup = {
 
     let group = new Group(Group.sanitize(req.body)); // TODO validate empty req.body
     group.leader = user._id;
+    group.memberCount = 1;
```
crookedneighbor: In my opinion, this is something that should happen in the `new Group` call.

crookedneighbor: Maybe not. I'm of two minds about it.

website/src/controllers/api-v3/groups.js
```diff
@@ -257,11 +260,13 @@ api.joinGroup = {
...
-        user.save(),
-        User.update({_id: user.invitations.party.inviter}, {$inc: {'items.quests.basilist': 1}}).exec(), // Reward inviter
-      ]).then(() => {
+      let promises = [group.save(), user.save()];
+      if (group.type === 'party' && inviter) {
+        promises.push(User.update({_id: inviter}, {$inc: {'items.quests.basilist': 1}}).exec());
+      }
+
+      Q.all(promises)
+      .then(() => {
```
crookedneighbor: Let's put this `.then` On the same line as `Q.all`

crookedneighbor:
I think the controller code looks sound. I have some nitpicky thoughts about test structure that I will try to post tomorrow.

website/src/controllers/api-v3/groups.js
KristianTashkov: I have no strong opinion about it, the place where you add the first user and the leader seemed the most appropriate.

KristianTashkov:
I fixed the code style issue and added expects for the `group.memberCount = 1` in the appropriate tests.

website/src/controllers/api-v3/groups.js
paglias: We could add 1 as the default value for memberCount in the model

paglias:
@KristianTashkov this looks good, the only thing is that tests for invite route should be in a different file called `POST-groups_groupId_join`

test/api/v3/integration/groups/POST-groups_groupId_join.js
```diff
@@ -0,0 +1,63 @@
...
+      return generateUser({balance: 1, 'items.quests.basilist': 0 }).then((generatedUser) => {
+        user = generatedUser;
+        return user.post('/groups', {
+          name: groupName,
+          type: groupType
+        });
+      })
+      .then((guild) => {
+        return generateUser({ 'invitations.guilds': [{ id: guild._id}]}).then((generatedUser) => {
+          return generatedUser.post('/groups/' + guild._id + '/join');
```
TheHollidayInn: Is this okay to generate an invitation this way? Can we do the same for simply adding users to groups? This is the style I was talking about for skipping routes and only testing the route we need to call. @paglias @crookedneighbor

paglias: in this case it was done because we don't have the invite route ready yet, in general I don't know. To me it doesn't look like calling real routes makes things more difficult, rather it should help with routes that have a lot of side effects that are more difficult to replicate. But I'll leave the decision to @crookedneighbor

KristianTashkov: While writing it I changed my opinion on what's better like 5 times. On one hand calling the API is better, but sometimes you need to set a lot of other stuff to make an API call work, that will just set a variable for you. It slows down the unit test, but also makes it resilient to change in the implementation of said API route. If we don't care about unit tests running time and the API call can be executed without a lot of hassle, I think it's the better choice. If it's easier to mock the API route, than to set up the API call to actually run, maybe mocking is better.

TheHollidayInn: Yea, I agree. And, @crookedneighbor, I were discussing it. I couldn't think of the word, but Mocking was what I was thinking of. There is a difference between integration tests and unit tests though. Unit tests you definitely want to mock all this.

KristianTashkov:
@crookedneighbor  Rewrote using `await` syntax :heart:

test/api/v3/integration/groups/POST-groups_groupId_join.js
```diff
@@ -0,0 +1,51 @@
...
+  let user;
+
+  beforeEach(() => {
+    return generateUser().then((generatedUser) => {
+      user = generatedUser;
+    });
+  });
+
+  context('Guilds', () => {
+    it('inviting to guild doesn\'t give basilist to inviter', async () => {
```
crookedneighbor: This should read more naturally in English: `it inviting to guild doesn't give basilist to inviter`

Perhaps:

```
it('does not give basilist quest to inviter when joining a guild', async () => {
```

test/api/v3/integration/groups/POST-groups_groupId_join.js
```diff
@@ -0,0 +1,51 @@
+import {
+  generateUser,
+} from '../../../../helpers/api-integration.helper';
+
+describe('POST /group/:groupId/join', () => {
+  let user;
+
+  beforeEach(() => {
```
crookedneighbor: You can make this:

```
beforeEach(async () => {
  user = await generateUser();
});
```

test/api/v3/integration/groups/POST-groups_groupId_join.js
```diff
@@ -0,0 +1,51 @@
...
+describe('POST /group/:groupId/join', () => {
+  let user;
+
+  beforeEach(() => {
+    return generateUser().then((generatedUser) => {
+      user = generatedUser;
+    });
+  });
+
+  context('Guilds', () => {
```
crookedneighbor: I'd make this context something like: `Accepting an invitation to a guild`

test/api/v3/integration/groups/POST-groups_groupId_join.js
```diff
@@ -0,0 +1,51 @@
...
+
+      let invitedUser = await generateUser({ 'invitations.guilds': [{ id: guild._id}]});
+      await invitedUser.post('/groups/' + guild._id + '/join');
+      
+      let changedUser = await user.get('/user')
+      expect(changedUser.items.quests.basilist).to.equal(0);
+    });
+  });
+
+  context('Parties', () => {
```
crookedneighbor: I'd make this context something like: `Accepting an invitation to a party`

test/api/v3/integration/groups/POST-groups_groupId_join.js
```diff
@@ -0,0 +1,51 @@
...
+
+  beforeEach(() => {
+    return generateUser().then((generatedUser) => {
+      user = generatedUser;
+    });
+  });
+
+  context('Guilds', () => {
+    it('inviting to guild doesn\'t give basilist to inviter', async () => {
+      let groupName = 'Test Guild';
```
crookedneighbor: I'd move the guild creation into a beforeEach block in this context.

test/api/v3/integration/groups/POST-groups_groupId_join.js
KristianTashkov: I plan to add more tests here for the join route once this PR is merged.

test/api/v3/integration/groups/POST-groups_groupId_join.js
KristianTashkov: Same answer as before, when I add the new tests the plan is to use more beforeEach blocks for common logic.

test/api/v3/integration/groups/POST-groups_groupId_join.js
KristianTashkov: Same as party context.

test/api/v3/integration/groups/POST-groups_groupId_join.js
```diff
@@ -0,0 +1,51 @@
...
+      expect(changedUser.items.quests.basilist).to.equal(0);
+    });
+  });
+
+  context('Parties', () => {
+    it('gives basilist quest item to the inviter', async () => {
+      let groupName = "Test Party";
+      let groupType = "party";
+
+      user = await generateUser({'items.quests.basilist': 0 });
```
crookedneighbor: I don't think this user should be instantiated with `'items.quests.basilist': 0`. The test should cover that it handles users who do not have the scroll at all yet.

In addition, there should be an additional test that has a user that is instantiated with `'items.quests.basilist': 5` (the 5 is arbitrary, it should just be greater than 0) to show that it increments the value

test/api/v3/integration/groups/POST-groups_groupId_join.js
```diff
@@ -0,0 +1,51 @@
...
+      await invitedUser.post('/groups/' + guild._id + '/join');
+      
+      let changedUser = await user.get('/user')
+      expect(changedUser.items.quests.basilist).to.equal(0);
+    });
+  });
+
+  context('Parties', () => {
+    it('gives basilist quest item to the inviter', async () => {
+      let groupName = "Test Party";
```
crookedneighbor: The party created should be put into a beforeEach in this context.

test/api/v3/integration/groups/POST-groups_groupId_join.js
```diff
@@ -0,0 +1,51 @@
...
+      let party = await user.post('/groups', {
+        name: groupName,
+        type: groupType
+      })
+      
+      let invitedUser = await generateUser({ 'invitations.party': { id: party._id, inviter: user._id }} )
+      await invitedUser.post('/groups/' + party._id + '/join');
+      
+      let changedUser = await user.get('/user');
+      expect(changedUser.items.quests.basilist).to.equal(1);
```
crookedneighbor: For things like this, I'm unsure of which syntax we want to use. We could do it like you have it, or we could do it like this:

``` js
await expect(user.get('/user')).to.eventually.have.deep.property('items.quest.basilist', 1);
```

Does anyone have thoughts?

website/src/controllers/api-v3/groups.js
```diff
@@ -257,11 +259,12 @@ api.joinGroup = {
 
       if (group.memberCount === 0) group.leader = user._id; // If new user is only member -> set as leader
 
-      Q.all([
-        group.save(),
-        user.save(),
-        User.update({_id: user.invitations.party.inviter}, {$inc: {'items.quests.basilist': 1}}).exec(), // Reward inviter
-      ]).then(() => {
+      let promises = [group.save(), user.save()];
```
crookedneighbor: Can we add a new line here?

test/api/v3/integration/groups/POST-groups_groupId_join.js
crookedneighbor: I think it should be done now.

test/api/v3/integration/groups/POST-groups_groupId_join.js
crookedneighbor: I [wrote a helper](https://github.com/HabitRPG/habitrpg/blob/02f4a841463c87159041c2e66b2faec60caa140d/test/helpers/api-integration.helper.js#L132-L212) (for v2 routes) to generate a group with members and invitees.

Once the invite routes are done, that should be used instead. Until them, I'd be fine with adding a todo comment to impliment it.

test/api/v3/integration/groups/POST-groups_groupId_join.js
```diff
@@ -0,0 +1,51 @@
...
+      let groupType = 'guild';
+
+      user = await generateUser({balance: 1, 'items.quests.basilist': 0 })
+      let guild = await user.post('/groups', {
+        name: groupName,
+        type: groupType
+      });
+
+      let invitedUser = await generateUser({ 'invitations.guilds': [{ id: guild._id}]});
+      await invitedUser.post('/groups/' + guild._id + '/join');
```
crookedneighbor: Where possible, we should be using template strings:

```
`/groups/${guild._id}/join`
```

KristianTashkov:
@crookedneighbor I fixed most of your notes. When I added more tests to handle both cases where the user has scrolls and doesn't then we are changing the user that creates the group, hence his creation and the group's can't be in a `beforeEach`. For future tests we can add a context with a valid creator where we don't care about basilist, but for now I just don't see a way to include these two cases and still have the creation in a `beforeEach`.

test/api/v3/integration/groups/POST-groups_groupId_join.js
```diff
@@ -0,0 +1,84 @@
...
+  context('Accepting invitation to a guild', () => {
+    beforeEach(async () => {
+      groupName  = 'Test Guild';
+      groupType = 'guild';
+    });
+
+    it('does not give basilist quest to inviter when joining a guild', async () => {
+      user = await generateUser({balance: 1})
+      group = await user.post('/groups', {
+        name: groupName,
```
crookedneighbor: I don't think these need to be variables. There's no need to assert on them later. Why not just:

```
{
  name: 'Test Guild',
  type: 'guild',
}
```

crookedneighbor:
I've been toying with adding an update method to the API user class for scenarios like this.

crookedneighbor:
Updated it to use beforeEach and `user.update`

Closed with https://github.com/HabitRPG/habitrpg/commit/7025bbfa3244dc1da9477513b6db68f4e436bf17

Remind me of your Habitica UUID?

PR #6417, opened 2015-12-30
Add option to reverse chat order

### LINKED ISSUE #6006: need option to reverse the order of Tavern/guild/party chat for screen reader accessibility [$10]
As described in the [Accessibility Settings](https://trello.com/c/pxTcbKuD/381-accessibility-settings) Trello card, in recent comments from habitrpguser and Alice Harris, it is desirable to be able to reverse the order of chat messages so that the oldest are at the top. This would allow you to use a screen reader to read the messages to you in chronological order.

This should apply to the Tavern, guilds, and your party.

This is not just for visually impaired people but also for players with good vision who want to have chats read to them while they are commuting, doing housework, etc. It's a productivity setting as well as an accessibility one.

We should decide whether this option is controlled by a preference checkbox under Settings -> Site, or by a button on the Tavern/guild/party page (e.g., next to the "Send Chat" and "Fetch Recent Messages" buttons), or both. Points to consider are:
- Would the player want to change the sort order all / almost all the time or only some of the time?
- Which would be more annoying: having to click the button every time you wanted to change the order or having to go into Settings -> Site every time?
- Would a button be too confusing or cluttery for some users?
- Would a Settings -> Site preference checkbox be too obscure?

An alternative would be to have a Settings -> Site preference checkbox, and if it is set to on, then display a button on the chat page for toggling the sort order. That would give a quick-change option for those who wanted it while not cluttering up the chat interface for other players. I think that's my favourite alternative.

Comments are welcome from people who would use this feature.
## <bountysource-plugin>

There is a **[$10 open bounty](https://www.bountysource.com/issues/27152136-need-option-to-reverse-the-order-of-tavern-guild-party-chat-for-screen-reader-accessibility?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

GeorgeSG:
Fixes #6006.
- Added a user setting `reverseChatOrder` that reverses the order of chat messages in Party/Tavern/Guild chats.
- ~~Added the setting under **Settings -> Site** (screenshot attached)~~
- As discussed in the issue, I tried to add it in the chat box as well. (screenshots attached) It reorders the chat on-click. I'm not sure if it's okay like this, but I tried multiple variants of where/how to present the option and this seemed to me like the best way because it's consistent with the Site Settings and it's descriptive. Anyway, It's easy to change/remove if needed.

My user id is: `a187eae5-b3d2-479c-99c5-3e00b3857e24`

~~Settings -> Site~~ (change reverted)
options
**Chat Box**
chatbox
**Chat Box - popover**
chatbox_tooltip

### DIFF
--- a/common/locales/en/settings.json
+++ b/common/locales/en/settings.json
@@ -10,6 +10,8 @@
     "newTaskEditPop": "With this option set, new tasks will immediately open for you to add details like notes and tags.",
     "dailyDueDefaultView": "Set Dailies default to 'due' tab",
     "dailyDueDefaultViewPop": "With this option set, the Dailies tasks will default to 'due' instead of 'all'",
+    "reverseChatOrder": "Show chat messages in reverse order",
+    "reverseChatOrderPop": "Show the messages in the Tarvern, Guild, and Party chats in reverse order, so that the oldest are on top.",
     "startCollapsed": "Tag list in tasks starts collapsed",
     "startCollapsedPop": "With this option set, the list of task tags will be hidden when you first open a task for editing.",
     "startAdvCollapsed": "Advanced Options in tasks start collapsed",
--- a/website/src/models/user.js
+++ b/website/src/models/user.js
@@ -360,6 +360,7 @@ var UserSchema = new Schema({
     tagsCollapsed: {type: Boolean, 'default': false},
     advancedCollapsed: {type: Boolean, 'default': false},
     toolbarCollapsed: {type:Boolean, 'default':false},
+    reverseChatOrder: {type:Boolean, 'default':false},
     background: String,
     displayInviteToPartyWhenPartyIs1: { type:Boolean, 'default':true},
     webhooks: {type: Schema.Types.Mixed, 'default': {}},
--- a/website/views/options/social/chat-box.jade
+++ b/website/views/options/social/chat-box.jade
@@ -18,3 +18,10 @@ form.chat-form(ng-if='user.flags.communityGuidelinesAccepted' ng-submit='postCha
       input.btn(type='submit', value=env.t('sendChat'), ng-disabled='_sending')
       button.btn(type="button", ng-click='sync(group)', ng-disabled='_sending')=env.t('toolTipMsg')
     include ../../shared/formatting-help
+  .chat-controls.clearfix
+    .chat-buttons
+      .checkbox
+        label
+          input(type='checkbox', ng-model='user.preferences.reverseChatOrder', ng-change='set({"preferences.reverseChatOrder": user.preferences.reverseChatOrder?true: false})')
+          span.hint(popover-trigger='mouseenter', popover-placement='top', popover=env.t('reverseChatOrderPop'))=env.t('reverseChatOrder')
+
--- a/website/views/options/social/chat-message.jade
+++ b/website/views/options/social/chat-message.jade
@@ -1,6 +1,9 @@
 mixin chatMessages(inbox)
   ul.list-unstyled.tavern-chat
-    - var ngRepeat = inbox ? 'message in user.inbox.messages | toArray:true | orderBy:"sort":true' : 'message in group.chat track by message.id'
+    -
+      var ngRepeat = inbox ?
+        'message in user.inbox.messages | toArray:true | orderBy:"sort":true' :
+        'message in group.chat | orderBy:"timestamp":!user.preferences.reverseChatOrder track by message.id'
     li.chat-message(ng-repeat=ngRepeat, ng-class=':: {highlight: isUserMentioned(user,message) || message.uuid=="system", "own-message": user._id == message.uuid}')
       span.pull-right.text-danger(ng-if="user.contributor.admin && message.flagCount > 0")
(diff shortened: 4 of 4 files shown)


crookedneighbor:
Personally, I don't think that this needs to be in the settings page at all. I think it will be sufficient to have it in the chat menu.

GeorgeSG:
@crookedneighbor, I personally like it the way it is, mostly because it is a _persistent_ setting that is saved in the user's preferences. I think it'll be weirder and inconsistent to have a user preference only in the chat box, while all other preferences are in the settings. To me, the presence of it in the chat box is just for convenience and not required - if I have to leave it in one place only, I'd rather leave it only in the settings.

If you're suggesting that this shouldn't be a persistent setting altogether, then that's a different thing. It'll just be a sort option in the chat page.

Anyway, this was discussed in the original issue and I think everybody agreed it should be a setting. It's an easy change to make, but let's reach a final decision first. cc: @lemoness, @Alys, @habitrpguser

Alys:
Not all persistent settings are on the Settings page, for example, the Use Costume option is on the Equipment page.

Lately, I've been thinking that I'd prefer to have more settings on the pages they specifically apply to (e.g., I'd like the settings to control whether you see the Pet-Hatched and Mount-Tamed popups being on the Market and Pets pages, but I haven't got around to suggesting that anywhere yet). Too many settings on the Settings page could make it appear overwhelming for some users. There's also a lot of users who don't know the Settings page exists, but they would see settings that were on the pages they applied to.

The issue was talking about having a permanent setting on the Settings page and a non-persistent option controlled by a button on each chat page. But it's so easy to tick that checkbox on and off that I no longer think a non-persistent button would be useful. I like how you've done the checkbox on the chat pages, including the on-hover text. Now that I've seen it, I think the redundant checkbox is the one on the Settings page.

GeorgeSG:
@Alys, that's an interesting problem to tackle. Here are some of my thoughts on the topic.

**Related to sorting the chat messages:**
You're right about the "use costume" option, I didn't think of it. Also, I guess I didn't understand the original issue very well (describing the option in the chat to be non-persistent). I'm okay with having the option as a permanent preference and shown only on the chat-box. I still feel it's somewhat inconsistent with everything else, but it's a minor thing and it'll be very easy to find. In order to do that, we just have to revert the changes I made in`website/views/options/settings.jade`

**Related to settings in general:**
By putting settings on the pages they're related to, you run the risk of scattering too many settings on too many different places. There are benefits to having a centralized place for all settings - if you want to change a setting you know exactly where to look for it and don't have to go through multiple places to find it.

I especially like the current implementation of the pets & mounts popover, because you have the ability to disable the popup in the popup itself. As Steve Jobs said: _"Ask them. Ask them everytime. Make them tell you to stop asking them"_ IMHO, the checkbox in the settings page is mostly for re-enabling, not disabling the popups in the first place. Also, most of the people who choose to disable won't need to enable ever again, so they don't need to see the option every time on the pets & mounts page. And for the rare cases that a user disabled and then wants to re-enable, they have the option to do so from the settings.

The problems that you describe - that settings could become overwhelming in numbers and that some users don't know the settings exists at all - are very valid concerns, but that doesn't mean we should get rid of centralized settings to fix them. As Habitica grows, I think it's expected that we will have more and more settings - we should be able to handle that.

We could think of redesigning and restructuring the settings pages (including API, Export, etc). For example we can introduce categories of options (e.g. Appearance, Default Behavior, Popups). I don't like the "Registration" section being there in the second column - this is something that should probably be on the edit profile page.

If there are a lot of users that don't know settings exist, we could work on that too. For example, instead of only having a "cog" icon (settings icon in top rihgt), we could make it like the User/Social/Inventory buttons - have it say "Settings", which would directly open the site settings. And then have the right part with the cog to open the dropdown.

Anyway, I realize I'm talking about some pretty big changes. That's just my two cents :)

Alys:
Please do revert the changes to website/views/options/settings.jade. We'll try this setting on the chat pages alone first, and see how that goes.

I won't go into detail here about your other points (we can refer back to them later from elsewhere if we go further with settings changes), but one point about the pet and mount settings: yes, the checkbox in the settings page is mostly for re-enabling the popups, and that is something that some players will want to do often. E.g., you release then re-hatch a bunch of standard pets and you don't want to get the popup for each one of those so you turn it off. Then you turn it back on afterwards for when you are hatching just one or two quest pets at a time, which are more special and novel so you do want the popup. I've changed that popup setting several times since it was implemented.

GeorgeSG:
Reverted the settings page changes. PTAL :)

Alys:
I've tested this and it works well - thank you @GeorgeSG!

I believe it's ready to merge, but we're deploying some other, critical changes in a few hours, and I think this PR should wait until a day or two after that.

SabreCat:
Merged in and noted toward your next contributor tier! Thanks @GeorgeSG!

PR #6423, opened 2015-12-30
tests(api): Convert new group tests to use async and await (ES7)

crookedneighbor:
@TheHollidayInn @KristianTashkov @paglias @SabreCat

### DIFF
--- a/test/api/v2/groups/POST-groups.test.js
+++ b/test/api/v2/groups/POST-groups.test.js
@@ -5,14 +5,11 @@ import {
 } from '../../../helpers/api-integration.helper';
 
 describe('POST /groups', () => {
-
   context('All groups', () => {
     let leader;
 
-    beforeEach(() => {
-      return generateUser().then((user) => {
-        leader = user;
-      });
+    beforeEach(async () => {
+      leader = await generateUser();
     });
 
     xit('returns defaults? (TODO: it\'s possible to create a group without a type. Should the group default to party? Should we require type to be set?', () => {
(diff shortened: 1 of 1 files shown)


KristianTashkov:
:ship: it :) Tests like that are so much cleaner when you have to setup multiple objects one after another for the test.

TheHollidayInn:
Does eslint play nicely with this?

test/api/v2/groups/POST-groups.test.js
```diff
@@ -98,49 +92,52 @@ describe('POST /groups', () => {
...
-      }).then((group) => {
-        expect(group.balance).to.eql(1);
-        return leader.get('/user');
-      })).to.eventually.have.deep.property('balance', 1);
+      });
+
+      expect(guild.balance).to.eql(1);
+
+      let updatedUser = await leader.get('/user');
+
```
TheHollidayInn, line 139: Does it matter that there is a variable declaration down here?

test/api/v2/groups/POST-groups.test.js
```diff
@@ -24,61 +21,58 @@ describe('POST /groups', () => {
...
+      expect(group.description).to.eql(group.description);
+      expect(group.leaderMessage).to.eql(group.leaderMessage);
+      expect(group.leaderOnly).to.eql(group.leaderOnly);
+      expect(group.memberCount).to.eql(1);
     });
 
-    it('returns a populated members array', () => {
-      return leader.post('/groups', {
+    it('returns a populated members array', async () => {
+      let party = await leader.post('/groups', {
```
TheHollidayInn, line 43: So, should we not keep the return statements notation for the actual route we are testing? I see below you kept the return syntax for the errors.

test/api/v2/groups/POST-groups.test.js
crookedneighbor, line 139: No. It just scopes the variable to this block.

crookedneighbor, line 139: Oh, you mean for eslint. Hmmm... possibly.

test/api/v2/groups/POST-groups.test.js
crookedneighbor, line 43: Personally, I like this better.

For the errors, I thought you needed to do the `return <Promise>` syntax, but since it's a promise, you can just use `await` for those as well. I'll update the PR. Thanks for pointing it out.

crookedneighbor, line 43: It also makes the script I'm going to write to do this en masse much easier :stuck_out_tongue:

PR #6424, opened 2015-12-30
Port API v3 controllers to es7 async functions

paglias:
This PR is against `api-v3-groups` and currently implements the auth controller only.

Notes:
1. eslint decided not to support `async/await` at this point so I had to use [babel-eslint](https://github.com/babel/babel-eslint) and [eslint-plugin-babel](https://github.com/babel/eslint-plugin-babel)
2. ~~I had to rewrite `passport._strategies.facebook.userProfile` in `website/src/controllers/api-v3/auth.js` to a promise because eslint (even with the 2 plugins above) treat an async function inside a callback as a Syntax error. This may be an error in eslint, will test further~~

Do not merge (yet)


paglias:
Regarding point 2: it really fails to compile `await` inside a callback, I'll post an issue on babel to see if the behavior is normal or not. otherwise might be a deal breaker the impossibility to use callbacks with inside `away/async`.

opinions @crookedneighbor ?

website/src/libs/api-v3/setupRoutes.js
```diff
@@ -4,7 +4,11 @@ import express from 'express';
 import _ from 'lodash';
 
 const CONTROLLERS_PATH = path.join(__dirname, '/../../controllers/api-v3/');
-let router = express.Router(); // eslint-disable-line new-cap
+let router = express.Router(); // eslint-disable-line babel/new-cap
+
+// Wrapper function to handler `async` route handlers that return promises
+// It takes the async function, execute it and pass any error to next (args[2])
+let _wrapAsyncFn = fn => (...args) => fn(...args).catch(args[2]);
```
TheHollidayInn, line 11: Do we need to wait for this to be merged before we can start converting the controllers to the async syntax?

paglias, line 11: Yep, I'm trying to finish this in the next few hours

paglias:
This PR is ready to be merged unless we want to port middlewares to async functions. Personally I'd prefer not to:
1. We would have to write a custom method to run middlewares sequentially instead of using express' next(). At this point it would be easier to migrate directly to [Koa](http://koajs.com/) (which I'm not suggesting)
2. Some of the middlewares are very simple and synchronous code that would be slowed down considerably if it was transformed to an async function or a promise

crookedneighbor:
I'm fine with middleware being synchronous. 

This looks :+1: to me.

.eslintrc
```diff
@@ -64,7 +66,7 @@
         "no-dupe-class-members": 2,
         "no-this-before-super": 2,
         "no-var": 2,
-        "object-shorthand": 2,
+        "object-shorthand": 0,
```
crookedneighbor, line 69: Why don't we want to use object-shorthand?

crookedneighbor, line 69: Oh, I see, you added it as a babel rule.

.eslintrc
```diff
@@ -106,7 +108,11 @@
         "space-unary-ops": 2,
         "spaced-comment": [2, "always", { "exceptions": ["-"]}],
         "padded-blocks": [2, "never"],
-        "no-multiple-empty-lines": [2, {"max": 2}]
+        "no-multiple-empty-lines": [2, {"max": 2}],
+        "generator-star-spacing": 0,
+        "babel/new-cap": 2,
+        "babel/object-shorthand": 2,
+        "babel/no-await-in-loop": 2,
```
crookedneighbor, line 115: This is not valid JSON (ends in a comma)

crookedneighbor, line 115: I'll fix it.

paglias, line 115: thanks

.eslintrc
paglias, line 69: yeah it replaces some rules so they work better with babel

PR #6425, opened 2015-12-30
Add screenreader Labels to the 'plus' and 'minus' checkboxes

serverless-mom:
# What

Adds `aria-label =plus` (or `minus`) and `role=button` attributes to the checkboxes for habits
# Why

This is an accessibility issue: some screen readers currently won't select the <a> object and therefore call both the positive and negative buttons 'checkbox' with no other information. Changing the role to 'button` ensures that keyboard navigation tools will act correctly (ie letting you click it repeatedly).
# Test

0) before switching to this branch, turn on a screen reader with Cmd+f5 in OSX or on windows grab the [JAWS free trial](http://www.freedomscientific.com/Downloads/JAWS) and see on the Habits page that you can select and press a context-free 'checkbox'
0a) Alternatively, note that the buttons for habits with no readable label lack `aria-label` attributes which isn't best practices for accessibility

1) switch to this branch

2) hooray there are now labels on the plus or minus buttons as appropriate. You may notice that the button role makes them easier to press, but this varies a bit by navigation tool.
2a) alternatively, just inspect the buttons and see that `aria-label` attributes are there now!
# Known issues

Right now the 'rewards' at the far right are still unusable for some blind users: they can find the price of the buttons but not the associated button.  I'm still working on a best solution for this, since I don't think I can use `aria-labelledby` here. 
# open questions

Why are these checkboxes here? Why not just the links or, even better, buttons? I'm not a front-end pro so this may be a dumb question (does it maybe have something to do with wanting to prevent double-clicks?)

### DIFF
--- a/website/views/shared/tasks/task_view/index.jade
+++ b/website/views/shared/tasks/task_view/index.jade
@@ -7,12 +7,14 @@
     input.task-input.habit.visuallyhidden(
       ng-if='task.up',
       type='checkbox',
+      aria-label='plus',
       ui-keypress='{ 13:"score(task, \'up\')" }' )
     a(ng-if='task.up', ng-click='applyingAction || score(task,"up")')
       span.glyphicon.glyphicon-plus
     input.task-input.habit.visuallyhidden(
       ng-if='task.down',
       type='checkbox',
+      aria-label='minus',
       ui-keypress='{ 13:"score(task, \'down\')" }' )
     a(ng-if='task.down', ng-click='applyingAction || score(task,"down")')
       span.glyphicon.glyphicon-minus
--- a/website/views/shared/tasks/task_view/index.jade
+++ b/website/views/shared/tasks/task_view/index.jade
@@ -8,6 +8,7 @@
       ng-if='task.up',
       type='checkbox',
       aria-label='plus',
+      role='button', 
       ui-keypress='{ 13:"score(task, \'up\')" }' )
     a(ng-if='task.up', ng-click='applyingAction || score(task,"up")')
       span.glyphicon.glyphicon-plus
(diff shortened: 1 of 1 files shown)


serverless-mom:
This PR part of the same effort as https://github.com/HabitRPG/habitrpg/pull/6418 but the two PR's aren't interdependent. This PR is a way tinier change and I don't know any way it could break anything

GeorgeSG:
Looks good to me, but I would squash the commits together and reword the message :)

crookedneighbor:
Don't worry about squashing, I will do that once it's ready to be merged in.

serverless-mom:
@GeorgeSG see, you say I should smoosh them together but #6418 is done in a pretty janky way, and this part of the 'fix' is more critical.

crookedneighbor:
@tobyfee I think @GeorgeSG just meant squashing the two commits in this PR.

crookedneighbor:
Also :100: for working on this.

crookedneighbor:
RE: your question about the checkboxes. I think there _is_ a reason... but I can't remember what it is... I think something to do with using the keyboard to activate them.

crookedneighbor:
Whoops, did not mean to close it. Should not review PRs after midnight :)

SabreCat:
Reviewing PRs after midnight turns @crookedneighbor into a gremlin!

crookedneighbor:
Sorry it's taking me so long to get to this. I'm gonna try to get this merged in tomorrow (after I download a screen-reading application)

crookedneighbor:
Uhg, still haven't had time to test this. Sorry this is taking so long.

crookedneighbor:
Okay, I was finally able to test this a little with OSx's voiceover. Do you think you could give more specific instructions for how this improves usability? I couldn't really tell a difference between your changes and the site as it currently is, but that's probably just that I don't know how to use voiceover properly.

It's possible that the JAWS software you mentioned makes a difference, but I couldn't figure it out with OSx's voiceover.

serverless-mom:
Yeah I'm sorry I wasn't more clear on the testing plans, I'll send some
specific steps & the differences.

crookedneighbor:
@tobyfee going to lose this, but if you want to send the instructions, I'm happy to re-open it. 

We're going to do a site redesign in the near future and accessibility is one of the things I want to have baked in from the beginning.

PR #6428, opened 2015-12-31
Encode special characters when sending to paypal

### LINKED ISSUE #6422: Special character in the gift subscription message box causes error/blockage
Hello! So, a week ago I wanted to gift my sister a subscription. 

I tried twice through the credit card process, completing the card information and everything, and it didn’t work. I got a message saying {"err":"Unexpected end of input"}, and I didn’t know if the payment had been made.

It seemed it hadn’t cause I didn’t get any message, so after a while I tried again through Paypal: it opened a page with the same message, no access to Paypal possible.

I finally had an idea of what was causing the problem and had a fourth try, again with Paypal, and it worked. It seems that all the time, what was causing the problem was a special character in the message box (I had written "Joyeux Noël de Papa & Maman"). I first thought the "ë" was the problem but I just tried again and it would seem that it actually comes from the ampersand.

So, there, I’m telling you guys. If it’s too much of a pain to fix and I’m the only old-fashioned one using ampersands, maybe a simple message warning about special characters in the message box would be enough? (If this is indeed the problem: I haven’t tried with other special characters like ; or }.)

My ID is 23c9e7d5-5f62-41fa-9b0c-80d0e759ff80 and I think I used Firefox (sure) on an Android/Samsung tablet and a Windows 8 laptop.

GeorgeSG:
Fixes #6422.

Basically, we're sending the whole `gift` stringified object to paypal in the URL, but special characters (`&` in particular) mess with the GET parameters. Encoding it with `encodeURIComponent` fixes the issue.

Sorry for the unnecessary changes - my editor removes trailing whitespaces by default and I thought it'd be okay to leave it like this.

Screenshot of received message:
gift_message

My UserID in habitica is: `a187eae5-b3d2-479c-99c5-3e00b3857e24`

### DIFF
--- a/website/public/js/services/paymentServices.js
+++ b/website/public/js/services/paymentServices.js
@@ -107,7 +107,7 @@ function($rootScope, User, $http, Content) {
       // because no easy method to intercept those types of closings
       // and we need to make some cleanup
       keyboard: false,
-      backdrop: 'static' 
+      backdrop: 'static'
     });
 
     modal.rendered.then(function(){
@@ -161,9 +161,9 @@ function($rootScope, User, $http, Content) {
     if(Payments.amazonPayments.type === 'single'){
       return Payments.amazonPayments.paymentSelected === true;
     }else if(Payments.amazonPayments.type === 'subscription'){
-      return Payments.amazonPayments.paymentSelected === true && 
+      return Payments.amazonPayments.paymentSelected === true &&
               // Mah.. one is a boolean the other a string...
-              Payments.amazonPayments.recurringConsent === 'true'; 
+              Payments.amazonPayments.recurringConsent === 'true';
     }else{
       return false;
     }
@@ -179,7 +179,7 @@ function($rootScope, User, $http, Content) {
       onPaymentSelect: function() {
         $rootScope.$apply(function(){
           Payments.amazonPayments.paymentSelected = true;
-        });        
+        });
       },
 
       onError: amazonOnError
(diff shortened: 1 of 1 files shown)


Alys:
Removing trailing whitespace is great! When we want to see the diff without whitespace changes, we can append ?w=1 to the URL, like this: https://github.com/HabitRPG/habitrpg/pull/6428/files?w=1

GeorgeSG:
@Alys, that's so awesome! Thanks for the tip :)

crookedneighbor:
I :+1: removing the blank spaces. We have an `eslint` for no trailing white space, but the only a small fraction of the code has been cleaned up enough for us to use it.

crookedneighbor:
Normally I'd require tests to be written around a change in a service, but seeing as there are no existing tests (:disappointed:), the payments service is a bit of a mess to begin with, and this change seems sane... I'm fine with just [creating an issue for it](https://github.com/HabitRPG/habitrpg/issues/6429) and writing the tests later.

GeorgeSG:
Yeah, I was also kinda disappointed there are no tests. I can try to write
some after the holidays. I think it'll be nice to add some end-to-end tests
as well, if possible (not sure)

crookedneighbor:
The E2E tests are kinda borked and no one has had the bandwidth to really fix them. Right now, they assert that the login modal appears, but they hang when actually trying to write a test that logs in.

If you want to work on fixing them, go for it.

GeorgeSG:
@crookedneighbor, are you waiting for me to add tests? I'm not looking into e2e tests right now and from your previous comment I thought you were fine with merging this as-is. IMO it should be merged, because the bug is potentially stopping people from paying with paypal :)

If I add tests, I 'll track them in the issue you opened (#6429)

crookedneighbor:
I'm just busy and haven't had a chance to finish this up yet. I'll get to it soon.

crookedneighbor:
Just tested it with my sandbox account and it works great :+1:

crookedneighbor:
Whoops. Totally forgot to add this to your contributor tier notes.

crookedneighbor:
If we ever miss adding a PR to your contributor notes, feel free to just comment on the PR.

GeorgeSG:
This came as a surprise, thanks! :)

PR #6446, opened 2016-01-02
Add tests for group join route

KristianTashkov:
@crookedneighbor added awaits to `user.update` calls.
I tested most of the functionality in the join route. The ones left are:
1. When joining a party with a non-active quest, invite the new user. I believe the quest logic is still not implemented so I skipped it.
2. When joining a group(public guild?) with 0 member make the joining user the leader. I couldn't make a guild with zero members, the leave route kept crashing.

### DIFF
--- a/test/api/v3/integration/groups/POST-groups_groupId_join.js
+++ b/test/api/v3/integration/groups/POST-groups_groupId_join.js
@@ -1,8 +1,20 @@
 import {
   generateUser,
+  translate as t,
 } from '../../../../helpers/api-integration.helper';
+import { v4 as generateUUID } from 'uuid';
 
 describe('POST /group/:groupId/join', () => {
+  it('returns error when groupId is not for a valid group', async () => {
+    let joiningUser = await generateUser();
+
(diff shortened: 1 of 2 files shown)


test/api/v3/integration/groups/POST-groups_groupId_join.js
```diff
@@ -12,23 +23,62 @@ describe('POST /group/:groupId/join', () => {
         name: 'Test Guild',
         type: 'guild',
```
crookedneighbor, line 25: Let's make this explicitly a private guild. Right now private is the default, but let's not count on that for this test.

paglias:
@KristianTashkov about point 2 it should be limited to public guilds only, private guilds should be deleted when they reach 0 members (as part of the leave route).

If you are willing to you could do two things: make sure the Group.leave method works correctly and convert it to use the `async/await` syntax

KristianTashkov:
@crookedneighbor added explicit private to guild creation
@paglias Yeah I can try to look at the leave route and make tests for it, and then fill in the test here, but that sounds like a new PR right?

paglias:
@KristianTashkov yeah a new PR is probably better

crookedneighbor:
Notated.

PR #6448, opened 2016-01-02
fix: Corrects logic where it was possible to customize avatar

### LINKED ISSUE #6427: Able to select locked avatar skins / hairstyles / shirts without purchasing with gems.
User ID: f96fd5ef-3a77-4913-8d2f-ec8af90b77b2
Browser: Firefox
OS: Windows 7

I'm a programmer so I prod things. I noticed that one is able to "select" locked avatar styles.

I've also tested shirts and skin colors, so I assume the bug is general to avatar customization in general.

crookedneighbor:
with gem purchasable items without purchasing them.

Closes #6427

### DIFF
--- a/common/script/content/appearance.js
+++ b/common/script/content/appearance.js
@@ -1,0 +1,60 @@
+export let defaultAppearancePreferences = {
+  background: {},
+  hair: {
+    bangs: {
+      0: true,
+      1: true,
+      2: true,
+      3: true,
+    },
+    base: {
+      0: true,
+      1: true,
+      3: true,
+    },
+    beard: {
+      0: true,
+    },
+    color: {
+      white: true,
+      brown: true,
+      blond: true,
+      red: true,
+      black: true,
+    },
+    flower: {
+      0: true,
+      1: true,
+      2: true,
+      3: true,
+      4: true,
+      5: true,
+      6: true,
+    },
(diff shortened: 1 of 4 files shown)


common/script/content/appearance.js
```diff
@@ -0,0 +1,60 @@
...
+  },
+  skin: {
+    ddc994: true,
+    f5a76e: true,
+    ea8349: true,
+    c06534: true,
+    '98461a': true,
+    915533: true,
+    c3e1dc: true,
+    '6bd049': true,
```
Alys, line 59: there's a mix of quoted and unquoted values here

crookedneighbor, line 59: As stupid as it is, object attributes that begin with a number and contain letters must be quoted. Those that begin with a letter or properties that are all numbers do not. I can quote them all and disable the eslint rule that objects to unnecessarily quoted items, if you prefer. I was debating about it.

Alys, line 59: Ah, sorry. The inconsistency looks weird, but there's a good reason for it. Disabling that rule is probably undesirable. I'm happy with whatever you want there.

crookedneighbor, line 59: I think you're right though, it looks weird. I'll set the style rule to be ignored just for this section.

PR #6456, opened 2016-01-04
use example.com in the example json so as not to spam legit servers

querry43:
I won't want to inadvertently spam yours.com domain.com with my testing.  example.com exists for this purpose, so I have changed most of the references in config.json.example.


crookedneighbor:
I don't _think_ any of the emails code actually gets hit in development mode, but you're right that these should be using example.com.

querry43:
@crookedneighbor , bf56a7cf-0970-40f2-af1b-f4df3f9bda4d.  Thanks.

PR #6458, opened 2016-01-04
[API v3] Challenges Tasks

paglias:
WIP


paglias:
Logic implemented. Tests missing

test/api/v3/integration/tasks/DELETE-tasks_id.test.js
```diff
@@ -16,7 +16,7 @@ describe('DELETE /tasks/:id', () => {
     let task;
 
     beforeEach(() => {
-      return user.post('/tasks', {
+      return user.post('/tasks?tasksOwner=user', {
```
crookedneighbor: I don't like this. I think tasksOwner should default to user.

paglias: As we had discussed earlier my idea was to have /tasks/user and /tasks/challenge/userId. I had to use a query parameter because otherwise the route handler would intercept requests going to /tasks/taskId. I think the parameter name could be changed, I don't like it either. But regarding the user being default I don't understand why it should be privileged over challenges (and in future other types). In my opinion it also makes the API behavior clearer

crookedneighbor: I agree that a separate route would be better and cleaner. We should figure out how to make express respect that, rather than go the query parameter route.

website/src/controllers/api-v3/tasks.js
```diff
@@ -54,26 +82,32 @@ api.createTask = {
...
     }
+
+    // If adding tasks to a challenge -> sync users
+    if (challenge) challenge.addTasks(tasks); // TODO catch/log
   },
 };
 
 /**
- * @api {get} /tasks Get an user's tasks
+ * @api {get} /tasks/:tasksOwner/:challengeId Get an user's tasks
```
crookedneighbor, line 211: I assume this is incorrect, given that we're using query parameters here instead?

paglias, line 211: I forgot to update it

test/api/v3/integration/tasks/DELETE-tasks_id.test.js
paglias: Okay, I'll try to write a regex that matches /task/(user/challenge)/optional challenge id without matching /tasks/taskId. If you've more experience with regex feel free to do it, I'm not very good at them

website/src/controllers/api-v3/tasks.js
```diff
@@ -15,80 +16,220 @@ import { preenHistory } from '../../../../common/script/api-v3/preenHistory';
 
 let api = {};
 
+// challenge must be passed only when a challenge task is being created
+async function _createTasks (req, res, user, challenge) {
```
crookedneighbor, line 20: I think user and challenge should be part of an options object.

```
async function _createTasks (req, res, options) {
  let {
    user,
    challenge,
  } = options;

  // stuff
}
```

website/src/controllers/api-v3/tasks.js
```diff
@@ -15,80 +16,220 @@ import { preenHistory } from '../../../../common/script/api-v3/preenHistory';
...
+    if (challenge) challenge.addTasks(tasks); // TODO catch/log
+  },
+};
 
-      // Validate that the task is valid and throw if it isn't
-      // otherwise since we're saving user and task in parallel it could save the user with a tasksOrder that doens't match reality
-      let validationErrors = newTask.validateSync();
-      if (validationErrors) throw validationErrors;
+// challenge must be passed only when a challenge task is being created
+async function _getTasks (req, res, user, challenge) {
```
crookedneighbor, line 112: Same here, user and challenge should be part of an options object.

website/src/controllers/api-v3/tasks.js
crookedneighbor, line 20: In fact, you don't even need to pass the user in, since it's on the `res.locals.user`. You could just have the arguments be `req, res, challenge`

crookedneighbor, line 20: Though I think I'd still be in favor of making the last argument an object to make expanding this function later easier.

crookedneighbor:
I haven't had time the last few days to fully explore this and write tests. Will do that this week.

test/api/v3/integration/tasks/GET-tasks_user.test.js
```diff
@@ -0,0 +1,33 @@
+import {
+  generateUser,
+} from '../../../../helpers/api-integration.helper';
```
crookedneighbor, line 3: This will need to change to ../../../../helpers/api-v3-integration.helper'; Will do wo when I work on it.

PR #6466, opened 2016-01-05
update help menu and FAQ to jump to Wikia

taldin:
The Help > FAQ option no longer jumps to the Wiki.  Added a help menu
item to fix that.   I also added a link to the Wiki FAQ at the bottom of
the FAQ.


taldin:
One problem I'm having is that the updated FAQ (link at the bottom, changed date to January 5, 2016) doesn't seem to be applied on the test db I have.    Might it be because there's some trickery involved in getting files in 'static' to do something?

Alys:
@taldin It's deliberate that the wiki's FAQ is no longer linked to because it is very overwhelming for new users - the staff have done user testing that shows that. Also I think markdown links aren't working yet on the iOS app. However if you would like to include a screenshot showing the change that this PR makes, we can ask the staff what they think.

The links shouldn't be related to the database. Can you give us more information about the problem in your second comment?

taldin:
image  That is what I changed.   It doesn't go to the Wiki FAQ:  it goes to the Wiki front page.   The problem I had was that the only way to get to the Wiki at all if you didn't know it existed was to click on the _Contribute to Habitica_ link, which was very non-intuitive, or 'Overview for New Users' and click on other links in there. 

The second comment:
The FAQ page (the new one with the dropdown arrows at https://habitica.com/static/faq/) apparently can't be altered via changing the website/views/static/faq.jade and common/locales/en/faq.json files  -- I've made changes, but they don't show up in the GUI.    And I'm wondering if there's some trickery to get the server to recognize that I changed them.

Alys:
For changing the FAQ items, edit common/locales/en/faq.json but note that there's two or three entries for each item - e.g., iosFaqAnswer12 and webFaqAnswer12. Make sure you're editing the web one for the change to show up on your local copy of the site. That should be all that's needed. Of course, editing the iOS one to match would also sometimes be needed.

What change do you want to make to the FAQ? If you wish, we can ask if it will be approved before you do further work on it.
_(Note to self: also ask if world boss entry should be removed now.)_

taldin:
I changed both the "iosFaqStillNeedHelp" and  "webFaqStillNeedHelp" entries:
 "If you have a question that isn't on this list, come ask in the [Newbies Guild](https://habitica.com/#/options/groups/guilds/5481ccf3-5d2d-48a9-a871-70a7380cee5a)! We're happy to help."      

....to include a link to the Wiki FAQ (so you can have a FAQ in your FAQ) and then I tried changing the date from September 2015 to January 2016 to reflect that I'd changed something.   But when I recompiled, nothing changed at all.  It made me think of the way that you couldn't change something in the main habitrpg repository without adjusting it in habitrpg-shared back in the day, and I was wondering if there was something similar going on.

SabreCat:
I find that if I'm running a local server and change a string, I need to halt the server entirely with a Ctrl+C and rerun "npm start" to get it to pick up the change. Just restarting the server, as it does automatically with many changes, isn't sufficient.

taldin:
I've closed the window completely, restarted the server in a new window, and still no luck.   Any other suggestions?

Alys:
Odd. Running `npm start` should be enough for it to work.

Try again and show us the output of npm start.

But first, see some comments I'm about to make on a couple of lines.

website/views/shared/header/menu.jade
```diff
@@ -58,6 +58,8 @@ nav.toolbar(ng-controller='MenuCtrl')
               li
                 a(target='_blank' href='https://habitica.com/static/faq/')=env.t('FAQ')
               li
+                a(target='_blank' href='http://habitica.wikia.com/wiki/Habitica_Wiki')=env.t('HabiticaWiki')
```
Alys: Links to any wiki page can be placed in the locales files so that translators can specify non-English wiki links where available. For example:

common/locales/en/tasks.json line 111:
 `"toDoHelp4": "For inspiration, check out these <a href='http://habitica.wikia.com/wiki/Sample_To-Dos' target='_blank'>sample To-Dos</a>!",`

website/views/shared/header/menu.jade
```diff
@@ -157,6 +159,8 @@ nav.toolbar(ng-controller='MenuCtrl')
             li
               a(target='_blank' href='https://habitica.com/static/faq/')=env.t('FAQ')
             li
+              a(target='_blank' href='http://habitica.wikia.com/wiki/Habitica_Wiki')=env.t('HabiticaWiki')
```
Alys: Same here, of course.

That's my final comment. :)

taldin:
Okay, this is the first time someone's commented on my code.  :) How do i get those changes?

Alys:
You need to make them yourself. ;) 

If changes were made by others, they'd appear in a pull request in your own branch, which you'd need to accept by merging.

I've modified my first comment so that it shows the sample code I was referring to as raw code, instead of as a link.

taldin:
Yepyep!  I already got that part covered.  :)  It's why I have the edits to the 'generic.json' file in the en directory.   Or is there something else you need done there that I'm not seeing?

-Taldin

Alys:
No, the actual link can be placed in the locale file so the URL can be "translated" too. So instead of this: `a(target='_blank' href='http://habitica.wikia.com/wiki/Habitica_Wiki')=env.t('HabiticaWiki')` plus this: `"HabiticaWiki": "The Habitica Wiki",`

you would edit the generic.json file to have something like this example: `"toDoHelp4": "For inspiration, check out these <a href='http://habitica.wikia.com/wiki/Sample_To-Dos' target='_blank'>sample To-Dos</a>!",`

taldin:
Oh!  I think I see.  Hang on.

taldin:
Like so?

taldin:
Also, here's that screenshot.
image

Alys:
I've loaded it locally and what you've done works. You could also put the URL and visible text of the link into a single locale string, like in the `toDoHelp4` example. Although keeping them separate probably reduces the chance that a translator accidentally messes up the HTML code.

When you click on the Help > FAQ menu entry, you're redirected to the production site, because the full habitica.com URL is coded into the link. If you didn't notice the redirection, that's probably why it seems that your changes aren't taking effect. Try this link: http://localhost:3000/static/faq/

Alys:
@lemoness : This PR makes two changes:

It adds "or on the wiki FAQ" to the text below, on both the website FAQ and the iOS FAQ (and changes the "Last updated" date at the top of the FAQ):

screen shot 2016-01-10 at 7 20 14 pm

---

It adds "The Habitica Wiki" as the second item in the Help menu on the website:
screen shot 2016-01-10 at 7 23 50 pm

---

Do you want these changes?

Unrelated to this, should we remove the World Boss question and answer now? If we did, we'd save the text for use for the next Boss.

taldin:
Yep, that's exactly the text I was looking for, thank you.  Now I know why it didn't work, too. :) 
As far as the world boss bit goes, we're not going to have one for another few months.  But when we do, we'll have to put the answer back in.   It would probably serve us better if we just reworded this to be 'whenever a world boss is up' as a question/answer.

taldin:
Suggested rewording:
Q:  How do I battle that World Boss that's in the Tavern?  
A:   World Bosses are special monsters that sometimes appear in the Tavern during a World Event.... (rest is the same from there)

Alys:
I think it's better to remove the World Boss question. The FAQ has only the most important questions for new users and additional information is more likely to be confusing than helpful. I've been seeing occasional questions in the Tavern about where a world boss can be found, and it's probably this FAQ item that's causing them. Also editing the question now and then back again later will just make extra work for the translators (and translations might not be done promptly enough to be useful). BTW, don't work on removing it - I want to talk to Blade about hiding it programmatically if Lemoness approves its removal.

lemoness:
I think removing the World Boss question is fine, as long as we remember to
put it back!

Link to Wiki FAQ is ok. However, in the drop-down menu, I would like the
link to the wiki to be last in the list.

Alys:
@crookedneighbor For removing the world boss FAQ question and answer, I'm thinking the best way to do it is to put in an `if` statement that prevents these lines being executed when `heading === world-boss` :
https://github.com/HabitRPG/habitrpg/blob/develop/website/views/static/faq.jade#L23-L25
That allows us to keep the text in the locales files (no need for translators to re-add it later when the next boss arrives), and also means that we can add new faq entries at the end of the list (index 13, etc) and have it all still work.

crookedneighbor:
I am now regretting the choice to use indexed faq names for this very reason :(

I'll try to look at an alternate solution in the near future. Really, since the faq is via the API, we should not send out the faq about the world boss at all if it is not to be displayed.

taldin:
Wiki Link is now moved to the bottom in the menu.

SabreCat:
Merged! Thanks @taldin!

Quick tip for future: the merge conflict was `habitrpg-shared.css`. That gets updated every time the server runs, so it doesn't need to be included in commits. Leaving it out will make your PRs easier to merge!

taldin:
Ah, gotcha.  I'm still a bit of a novice at this, so now I know.  :)

PR #6475, opened 2016-01-06
Updated invite route and function for new standards, and added tests.

TheHollidayInn:
Here I have updated the private functions to use new syntax. I have also implemented the ability to invite by email and userId. Then, I added tests. 

This PR is what I need merged before adding tests and the other group routes.

### DIFF
--- a/test/api/v3/integration/groups/POST-groups_invite.test.js
+++ b/test/api/v3/integration/groups/POST-groups_invite.test.js
@@ -2,6 +2,9 @@ import {
   generateUser,
   translate as t,
 } from '../../../../helpers/api-integration.helper';
+import { v4 as generateUUID } from 'uuid';
+
+const INVITES_LIMIT = 100;
 
 describe('Post /groups/:groupId/invite', () => {
   let inviter;
@@ -18,7 +21,7 @@ describe('Post /groups/:groupId/invite', () => {
 
   describe('user id invites', () => {
     it('returns an error when invited user is not found', async () => {
-      let fakeID = '206039c6-24e4-4b9f-8a31-61cbb9aa3f66';
+      let fakeID = generateUUID();
 
       await expect(inviter.post(`/groups/${group._id}/invite`, {
         uuids: [fakeID],
@@ -31,7 +34,7 @@ describe('Post /groups/:groupId/invite', () => {
     });
 
     it('returns an error when uuids is not an array', async () => {
-      let fakeID = '206039c6-24e4-4b9f-8a31-61cbb9aa3f66';
+      let fakeID = generateUUID();
 
       await expect(inviter.post(`/groups/${group._id}/invite`, {
         uuids: {fakeID},
@@ -48,6 +51,23 @@ describe('Post /groups/:groupId/invite', () => {
         uuids: [],
       }))
       .to.eventually.be.empty;
+    });
+
+    it('returns an error when there are more than INVITES_LIMIT uuids', async () => {
+      let uuids = [];
+
+      for (let i = 0; i < 101; i += 1) {
+        uuids.push(generateUUID());
+      }
+
+      await expect(inviter.post(`/groups/${group._id}/invite`, {
+        uuids,
+      }))
+      .to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: t('canOnlyInviteMaxInvites', {maxInvites: INVITES_LIMIT}),
+      });
     });
 
     it('invites a user to a group by uuid', async () => {
@@ -85,10 +105,24 @@ describe('Post /groups/:groupId/invite', () => {
       await expect(userToInvite.get('/user')).to.eventually.have.deep.property('invitations.guilds[0].id', group._id);
       await expect(userToInvite2.get('/user')).to.eventually.have.deep.property('invitations.guilds[0].id', group._id);
     });
+
+    it('returns an error when inviting multiple users and a user is not found', async () => {
+      let userToInvite = await generateUser();
+      let fakeID = generateUUID();
+
+      await expect(inviter.post(`/groups/${group._id}/invite`, {
+        uuids: [userToInvite._id, fakeID],
+      }))
+      .to.eventually.be.rejected.and.eql({
+        code: 404,
+        error: 'NotFound',
+        message: t('userWithIDNotFound', {userId: fakeID}),
+      });
+    });
   });
 
   describe('email invites', () => {
-    let testInvite = {name: 'test', email: 'test@habitca.com'};
+    let testInvite = {name: 'test', email: 'test@habitica.com'};
 
     it('returns an error when invite is missing an email', async () => {
       await expect(inviter.post(`/groups/${group._id}/invite`, {
@@ -119,6 +153,23 @@ describe('Post /groups/:groupId/invite', () => {
       .to.eventually.be.empty;
     });
 
+    it('returns an error when there are more than INVITES_LIMIT emails', async () => {
+      let emails = [];
+
+      for (let i = 0; i < 101; i += 1) {
+        emails.push(`${generateUUID()}@habitica.com`);
+      }
+
+      await expect(inviter.post(`/groups/${group._id}/invite`, {
+        emails,
+      }))
+      .to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: t('canOnlyInviteMaxInvites', {maxInvites: INVITES_LIMIT}),
+      });
+    });
+
     it('invites a user to a group by email', async () => {
       await expect(inviter.post(`/groups/${group._id}/invite`, {
         emails: [testInvite],
@@ -127,7 +178,7 @@ describe('Post /groups/:groupId/invite', () => {
 
     it('invites multiple users to a group by email', async () => {
       await expect(inviter.post(`/groups/${group._id}/invite`, {
-        emails: [testInvite, {name: 'test2', email: 'test2@habitca.com'}],
+        emails: [testInvite, {name: 'test2', email: 'test2@habitica.com'}],
       })).to.exist;
     });
   });
@@ -142,11 +193,34 @@ describe('Post /groups/:groupId/invite', () => {
       });
     });
 
+    it('returns an error when there are more than INVITES_LIMIT uuids and emails', async () => {
+      let emails = [];
+      let uuids = [];
+
+      for (let i = 0; i < 50; i += 1) {
+        emails.push(`${generateUUID()}@habitica.com`);
+      }
+
+      for (let i = 0; i < 51; i += 1) {
+        uuids.push(generateUUID());
+      }
+
+      await expect(inviter.post(`/groups/${group._id}/invite`, {
+        emails,
+        uuids,
+      }))
+      .to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: t('canOnlyInviteMaxInvites', {maxInvites: INVITES_LIMIT}),
+      });
+    });
+
     it('invites users to a group by uuid and email', async () => {
       let newUser = await generateUser();
       let invite = await inviter.post(`/groups/${group._id}/invite`, {
         uuids: [newUser._id],
-        emails: [{name: 'test', email: 'test@habitca.com'}],
+        emails: [{name: 'test', email: 'test@habitica.com'}],
       });
       let invitedUser = await newUser.get('/user');
 
--- a/test/api/v3/integration/groups/POST-groups_invite.test.js
+++ b/test/api/v3/integration/groups/POST-groups_invite.test.js
@@ -1,0 +1,236 @@
+import {
+  generateUser,
+  translate as t,
+} from '../../../../helpers/api-integration.helper';
+
+describe('Post /groups/:groupId/invite', () => {
+  let inviter;
+  let group;
+  let groupName = 'Test Public Guild';
+
+  beforeEach(async () => {
+    inviter = await generateUser({balance: 1});
+    group = await inviter.post('/groups', {
+      name: groupName,
+      type: 'guild',
+    });
+  });
+
+  describe('user id invites', () => {
+    it('returns an error when invited user is not found', async () => {
+      let fakeID = '206039c6-24e4-4b9f-8a31-61cbb9aa3f66';
+
+      await expect(inviter.post(`/groups/${group._id}/invite`, {
+        uuids: [fakeID],
+      }))
+      .to.eventually.be.rejected.and.eql({
+        code: 404,
+        error: 'NotFound',
+        message: t('userWithIDNotFound', {userId: fakeID}),
+      });
+    });
+
+    it('returns an error when uuids is not an array', async () => {
+      let fakeID = '206039c6-24e4-4b9f-8a31-61cbb9aa3f66';
+
+      await expect(inviter.post(`/groups/${group._id}/invite`, {
+        uuids: {fakeID},
+      }))
+      .to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: t('uuidsMustBeAnArray'),
+      });
+    });
+
+    it('returns empty when uuids is empty', async () => {
+      await expect(inviter.post(`/groups/${group._id}/invite`, {
+        uuids: [],
+      }))
+      .to.eventually.be.empty;
+    });
+
+    it('invites a user to a group by uuid', async () => {
+      let userToInvite = await generateUser();
+
+      await expect(inviter.post(`/groups/${group._id}/invite`, {
+        uuids: [userToInvite._id],
+      })).to.eventually.deep.equal([{
+        id: group._id,
+        name: groupName,
+        inviter: inviter._id,
+      }]);
+      await expect(userToInvite.get('/user'))
+        .to.eventually.have.deep.property('invitations.guilds[0].id', group._id);
+    });
+
+    it('invites multiple users to a group by uuid', async () => {
+      let userToInvite = await generateUser();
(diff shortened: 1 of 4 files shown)


test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -0,0 +1,198 @@
...
+  translate as t,
+} from '../../../../helpers/api-integration.helper';
+
+describe('Post /groups/:groupId/invite', () => {
+  let user;
+  let group;
+
+  beforeEach(async () => {
+    user = await generateUser({balance: 1});
+    let groupName = 'Test Public Guild';
```
crookedneighbor: I don't think it's meaningful to pull these out into variables.

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -0,0 +1,198 @@
...
+      }))
+      .to.eventually.be.rejected.and.eql({
+        code: 404,
+        error: 'NotFound',
+        message: t('userWithIDNotFound', {userId: fakeID}),
+      });
+    });
+
+    it('invites a user to a group by uuid', async () => {
+      let newUser = await generateUser();
```
crookedneighbor: Let's be more meaningful with the variable name. Maybe something like `invitedUser`

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -0,0 +1,198 @@
...
+      let invite = await user.post(`/groups/${group._id}/invite`, {
+        uuids: [newUser._id],
+      });
+      let invitedUser = await newUser.get('/user');
```
crookedneighbor: This could instead be:

```
await expect(invitedUser.get('/user')).to.eventually.have.deep.property('invitedUser.invitations.guilds[0].id', group._id);
```

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -0,0 +1,198 @@
+import {
+  generateUser,
+  translate as t,
+} from '../../../../helpers/api-integration.helper';
+
+describe('Post /groups/:groupId/invite', () => {
+  let user;
```
crookedneighbor: Let's make this something like `inviter` or `invitingUser`

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -0,0 +1,198 @@
...
+    let groupType = 'guild';
+    group = await user.post('/groups', {
+      name: groupName,
+      type: groupType,
+    });
+  });
+
+  describe('user id invites', () => {
+    it('returns an error when invited user is not found', async () => {
+      let fakeID = 'fakeID';
```
crookedneighbor: This should probably be a uuid instead of the string `'fakeID'`.

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -0,0 +1,198 @@
...
+      let newUser2 = await generateUser();
+      let invite = await user.post(`/groups/${group._id}/invite`, {
+        uuids: [newUser._id, newUser2._id],
+      });
+      let invitedUser = await newUser.get('/user');
+      let invitedUser2 = await newUser2.get('/user');
+
+      expect(invite).to.exist;
+      expect(invitedUser.invitations.guilds[0].id).to.equal(group._id);
+      expect(invitedUser2.invitations.guilds[0].id).to.equal(group._id);
```
crookedneighbor: Same two notes as the last test, could have more meaningful variable names and you can write the expectation directly on the `get('/user')` call for each user.

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -0,0 +1,198 @@
...
+    });
+
+    it('invites a user to a group by uuid', async () => {
+      let newUser = await generateUser();
+      let invite = await user.post(`/groups/${group._id}/invite`, {
+        uuids: [newUser._id],
+      });
+      let invitedUser = await newUser.get('/user');
+
+      expect(invite).to.exist;
```
crookedneighbor: There should be a separate test that asserts on what is returned from the invite call.

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -0,0 +1,198 @@
...
+    });
+  });
+
+  describe('email invites', () => {
+    it('invites a user to a group by email', async () => {
+      let invite = await user.post(`/groups/${group._id}/invite`, {
+        emails: [{name: 'test', email: 'test@habitca.com'}],
+      });
+
+      expect(invite).to.exist;
```
crookedneighbor: We should decide explicitly what the invite call returns and assert on it.

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -0,0 +1,198 @@
...
+      await expect(user.post(`/groups/${group._id}/invite`, {
+        uuids: [newUser._id],
+      }))
+      .to.eventually.be.rejected.and.eql({
+        code: 401,
+        error: 'NotAuthorized',
+        message: t('userAlreadyInvitedToGroup'),
+      });
+    });
```
crookedneighbor, line 264: What happens when you invite multiple users and one is already in the group/invited to the group and the others are not?

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -0,0 +1,198 @@
...
+  });
+
+  describe('party invites', () => {
+    let party;
+    let partyName = 'Test Party';
```
crookedneighbor: Again, not meaningful to have these as variable names.

website/src/controllers/api-v3/groups.js
```diff
@@ -391,145 +392,101 @@ api.removeGroupMember = {
...
+    }
+    userToInvite.invitations.guilds.push({id: group._id, name: group.name, inviter: res.locals.user._id});
+  } else if (group.type === 'party') {
+    if (!_.isEmpty(userToInvite.invitations.party)) {
+      throw new NotAuthorized(res.t('userAlreadyPendingInvitation'));
+    }
+    if (userToInvite.party._id) {
+      throw new NotAuthorized(res.t('userAlreadyInAParty'));
+    }
+    // @TODO: Why was this here?
```
crookedneighbor, line 421: Can you explain your question better?

website/src/controllers/api-v3/groups.js
```diff
@@ -563,15 +520,25 @@ api.inviteToGroup = {
     let uuids = req.body.uuids;
     let emails = req.body.emails;
```
crookedneighbor, line 528: Should we have a cap on the number of uuids and emails a user can invite at once?

website/src/controllers/api-v3/groups.js
```diff
@@ -563,15 +520,25 @@ api.inviteToGroup = {
...
 
-    if (uuids && emails) { // TODO fix this, low priority, allow for inviting by both at the same time
-      throw new BadRequest(res.t('canOnlyInviteEmailUuid'));
-    } else if (Array.isArray(uuids)) {
-      // return _inviteByUUIDs(uuids, group, user, req, res, next);
-    } else if (Array.isArray(emails)) {
-      // return _inviteByEmails(emails, group, user, req, res, next);
-    } else {
+    if (!Array.isArray(uuids) && !Array.isArray(emails)) {
```
crookedneighbor: Could we do something like:

``` js
let hasUuids = !_.isEmpty(Array.isArray(uuids));
let hasEmails = !_.isEmpty(Array.isArray(emails));

if (!hasUuids && !hasEmails) {
  throw new BadRequest(res.t('canOnlyInviteEmailUuid'));
}

let results = [];

if (hasUuids) {
  // stuff
}

if (hasEmails) {
  // stuff
}
```

website/src/controllers/api-v3/groups.js
```diff
@@ -563,15 +520,25 @@ api.inviteToGroup = {
...
+      results.push(uuidResults);
+    }
+
+    if (Array.isArray(emails)) {
+      let emailInvites = emails.map((invite) => _inviteByEmail(invite, group, user, req, res));
+      let emailResults = await Q.all(emailInvites);
+      results.push(emailResults);
+    }
+
+    res.respond(200, {});
```
crookedneighbor: We should respond with something. Maybe: `{ message: "<number> invitations sent" }`

website/src/controllers/api-v3/groups.js
```diff
@@ -391,145 +392,101 @@ api.removeGroupMember = {
...
+  ]})
+  .select({_id: true, 'preferences.emailNotifications': true})
+  .exec();
+
+  if (userToContact) {
+    userReturnInfo = await _inviteByUUID(userToContact._id, group, inviter, req, res);
+  } else {
+    userReturnInfo = invite.email;
+    // yeah, it supports guild too but for backward compatibility we'll use partyInvite as query
+    // TODO absolutely refactor this horrible code
```
crookedneighbor, line 473: We should listen to this TODO :smile:

website/src/controllers/api-v3/groups.js
```diff
@@ -391,145 +392,101 @@ api.removeGroupMember = {
...
+    const partyQueryString = JSON.stringify({id: group._id, inviter, name: group.name});
+    const encryptedPartyqueryString = encrypt(partyQueryString);
+    let link = `?partyInvite=${encryptedPartyqueryString}`;
+
+    let variables = [
+      {name: 'LINK', content: link},
+      {name: 'INVITER', content: inviter || inviter.name},
```
crookedneighbor: This doesn't look quite right to me. User is getting passed in here, so it should be `inviter.profile.name`

website/src/controllers/api-v3/groups.js
```diff
@@ -391,145 +392,101 @@ api.removeGroupMember = {
...
+      {name: 'REPLY_TO_ADDRESS', content: inviter.email},
```
crookedneighbor, line 481: This is also not right.

website/src/controllers/api-v3/groups.js
```diff
@@ -391,145 +392,101 @@ api.removeGroupMember = {
...
+
+  if (userToContact) {
+    userReturnInfo = await _inviteByUUID(userToContact._id, group, inviter, req, res);
+  } else {
+    userReturnInfo = invite.email;
+    // yeah, it supports guild too but for backward compatibility we'll use partyInvite as query
+    // TODO absolutely refactor this horrible code
+    const partyQueryString = JSON.stringify({id: group._id, inviter, name: group.name});
+    const encryptedPartyqueryString = encrypt(partyQueryString);
+    let link = `?partyInvite=${encryptedPartyqueryString}`;
```
crookedneighbor, line 476: @paglias Is this `partyInvite` query something in our email server?

paglias, line 476: No it's not on the email server, it's used on the main server to show invitations to non-registered users. If you open a habitica.com?partyInvite=code link I think once you signup you automatically get invited to the group. This is the old code https://github.com/HabitRPG/habitrpg/blob/develop/website/src/controllers/api-v2/user.js#L547 I'll refactor this once we get to that point

website/src/controllers/api-v3/groups.js
paglias, line 528: Hmm probably a good idea. At some point we'll also have to enforce limits in other places too

website/src/controllers/api-v3/groups.js
paglias: why `_.isEmpty` and not just `hasUuids = Array.isArray(uuids)` and `if (uuids)`?

website/src/controllers/api-v3/groups.js
paglias: or return the invitations that are now on the users' profiles

website/src/controllers/api-v3/groups.js
TheHollidayInn, line 473: Yea, I move some of the strings out to make it more readable. I'll take a look to see if there is any more to change.

website/src/controllers/api-v3/groups.js
TheHollidayInn: How about this?

```
    let uuids = req.body.uuids;
    let emails = req.body.emails;

    if (!uuids && !emails) {
      throw new BadRequest(res.t('canOnlyInviteEmailUuid'));
    }

    let uuidsIsArray = Array.isArray(uuids);
    let emailsIsArray = Array.isArray(emails);

    let results = [];

    if (uuidsIsArray) {
      let uuidInvites = uuids.map((uuid) => _inviteByUUID(uuid, group, user, req, res));
      let uuidResults = await Q.all(uuidInvites);
      results.push(uuidResults);
    }

    if (emailsIsArray) {
      let emailInvites = emails.map((invite) => _inviteByEmail(invite, group, user, req, res));
      let emailResults = await Q.all(emailInvites);
      results.push(emailResults);
    }
```

website/src/controllers/api-v3/groups.js
TheHollidayInn, line 528: That's a good idea. What should I cap at?

test/api/v3/integration/groups/POST-groups_invite.test.js
TheHollidayInn, line 264: It should just throw an error as shown in a test above. I'll make an explicit test for multiple users. Throwing an error in the middle may not be the best though, because, some users will have invites. Hmm..

website/src/controllers/api-v3/groups.js
paglias, line 528: I'd say 100

website/src/controllers/api-v3/groups.js
paglias: @TheHollidayInn seems good but since we only support emails OR uuids there's no need to push emailResults to results, just do `results = await Q.all(emailInvited/uuidInvites)`

TheHollidayInn: Well, I updated the code to support emails and uuids.

paglias: @TheHollidayInn at the same time? In which case this check should be removed https://github.com/HabitRPG/habitrpg/pull/6475/files#diff-110878cdf614e13dc8eac3ea77ed6265R523

TheHollidayInn: Alright, I'm going to add two more tests for each email and uuids to test if there was an object passed and if the array is empty. Then, I'll test out Blades code to see if it passes. Does that sounds correct? I'll post an update soon if you need to see.

TheHollidayInn: And, to clarify, here is the logic I am about to refactor the code for:

//If we have neither, return error
//If we have uuids, but empty or object, return error
 //If we have emails, but empty or object, return error

paglias: Yeah it looks good, I misread the code. One thing that may change is the error string here https://github.com/HabitRPG/habitrpg/pull/6475/files#diff-110878cdf614e13dc8eac3ea77ed6265R524 

And when you push to results you may want to use the spread operator to push invites individually to `results`: `results.push(...uuidResults)` More infot https://developer.mozilla.org/it/docs/Web/JavaScript/Reference/Operators/Spread_operator

TheHollidayInn: Yea, I just saw that error again while refactoring. I'm going to change it for sure.

And, sweet. I'll look into that operator.

website/src/controllers/api-v3/groups.js
paglias: @TheHollidayInn regarding this one it would be very useful to return the invitations that are stored on the users so we can update the UI without a new request

TheHollidayInn: Will do!

TheHollidayInn:
Okay, updates:
-I removed extra abstract variables
-I added more meaningful variables names
-Added real UUIDs for examples
-Added limit and tests for invites - to be change per our slack discussion
-And, I added a test that when a user invites multiple users (and on user has an error) that the server will still respond with an error. I'm unsure if this is the functionality we want though.

website/src/controllers/api-v3/groups.js
```diff
@@ -391,145 +395,105 @@ api.removeGroupMember = {
...
-        Group.find({type: 'party', members: {$in: [uuid]}}, function(err, groups){
-          if (err) return cb(err);
-          if (!_.isEmpty(groups) && groups[0].members.length > 1) {
-            return cb({code: 400, err: "User already in a party."})
-          }
-          sendInvite();
-        });
-      }
+async function _inviteByUUID (uuid, group, inviter, req, res) {
+  // @TODO: Add Push Notifications
```
paglias, line 399: @TheHollidayInn there's a reason why the push notifications code was removed?

website/src/controllers/api-v3/groups.js
```diff
@@ -391,145 +395,105 @@ api.removeGroupMember = {
...
-    if (usersAlreadyRegistered.length > 0){
-      return _inviteByUUIDs(usersAlreadyRegistered, group, inviter, req, res, next);
+  if (group.type === 'guild') {
+    if (_.contains(userToInvite.guilds, group._id)) {
+      throw new NotAuthorized(res.t('userAlreadyInGroup'));
+    }
+    if (_.find(userToInvite.invitations.guilds, {id: group._id})) {
+      throw new NotAuthorized(res.t('userAlreadyInvitedToGroup'));
+    }
+    userToInvite.invitations.guilds.push({id: group._id, name: group.name, inviter: res.locals.user._id});
```
paglias: @TheHollidayInn instead of `inviter: res.locals.user._id` you could use `inviter: inviter._id`

Issue #6460, opened 2016-01-04
Upgrade amount of users on static page


lemoness:
We've hit another milestone, which means it's time to update the static page.

It should now say, "Join over a million people making it fun to achieve goals!"

:)

Alys:
advertised in Aspiring Coders

lemoness:
Another similar change that is needed: adding the Discover Magazine logo to
the list of press logos. Should I make a separate ticket for that, or can
it tag along with this one?

Alys:
Tag along, to reduce review effort. :)

NathanMaton:
how do you upload content to https://d2afqr2xdmyzvu.cloudfront.net/front/images/presslogos/ ?

crookedneighbor:
They go in here: https://github.com/HabitRPG/habitrpg/tree/develop/website/public/front/images/presslogos

We'll upload them manually to cloudfront if your PR gets merged.

NathanMaton:
Ok sounds good. I went ahead and fixed this locally but just saw now Sighne has a PR open. Next time I'll know. Thanks.

crookedneighbor:
Closed with #6476

PR #6476, opened 2016-01-06
#6460: upgrade amount of users

Sighne:
Changes the amount of users to a million and adds the Discovery Magazine logo on static page (https://github.com/HabitRPG/habitrpg/pull/6460).


Alys:
There's a second instance of "900,000" that needs to be replaced in front.jade around line 582:

```
      #footercall
        .container-fluid
          .row
              h3= env.t('joinOthers', {userCount:'900,000'})
```

It's also in website/views/social/landing-page.jade :

```
      li=env.t('landingCopy3', {userCount:'900,000'})
```

It would be great if this number could be inserted in one place in a variable.

Sighne:
Thank you, I'll take a look at it.

Sighne:
Changes have been made.
I've also added a variable in routes/pages.js that is passed to the templates in static and social.

website/src/routes/pages.js
```diff
@@ -5,6 +5,8 @@ var _ = require('lodash');
 var locals = require('../middlewares/locals');
 var i18n = require('../libs/i18n');
 
+var userCount = '1,000,000';
```
crookedneighbor: Let's make this a constant.

``` js
const TOTAL_USER_COUNT = '1,000,000';
```

Sighne:
Good idea! I changed it.

website/src/routes/pages.js
paglias: hmm @crookedneighbor is const supported in the api v2 environment? it's still on node 0.10

paglias: I forgot we were using babel here too, ignore the comment ^

SabreCat:
Looking good! This will count toward a contributor tier. @Sighne, what's your Habitica User ID? (See https://habitica.com/#/options/settings/api or Settings in the mobile apps.)

Sighne:
Thank you!
Here is my ID: b8c9e0c3-dfe7-4025-b141-d7289ed8e935

SabreCat:
Tier awarded! Note that while your first tier is easy to get, subsequent tiers require increasing amounts of effort from one to the next. So level 2 will require multiple PRs, or a larger PR, to attain. In any case, keep helping out and we'll express our thanks accordingly!

PR #6486, opened 2016-01-10
Use nvm to install node 4 to vagrant env

KristianTashkov:
Reasons for the changes:
- NVM needs `curl` to download node versions
- Node 4 needs `gcc 4.8` or above to build files
- I was having problem with make skipping files because my host OS was changing their modified times. The `ntp` server seems to fix it.

`install.sh` should be ran after ssh-ing to the vagrant machine.

Keep in mind, we are using `npm install --no-bin-links` because apparently there are problems with some host OSes and symlinks. But our test setup is made to search for mocha in `./node_modules/bin/mocha` where the symlinks are. So only `npm start` will work while we use `--no-bin-links` and tests will fail.

### DIFF
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -16,5 +16,5 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.hostname = "habitrpg"
   config.vm.network "forwarded_port", guest: 3000, host: 3000, auto_correct: true
   config.vm.usable_port_range = (3000..3050)
-  config.vm.provision :shell, :path => "vagrant.sh"
+  config.vm.provision :shell, :path => "vagrant_scripts/vagrant.sh"
 end
--- a/vagrant_scripts/install.sh
+++ b/vagrant_scripts/install.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+echo Installing nvm...
+curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.30.1/install.sh | bash
+source ~/.profile
+
+echo Setup node 4...
+nvm install 4
+nvm use 4
+nvm alias default 4
+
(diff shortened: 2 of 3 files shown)


paglias:
@KristianTashkov great work, regarding:

> Keep in mind, we are using npm install --no-bin-links because apparently there are problems with some host OSes and symlinks. But our test setup is made to search for mocha in ./node_modules/bin/mocha where the symlinks are. So only npm start will work while we use --no-bin-links and tests will fail.

What about adding installing those modules globally and adding some documentation on how to run them instead of using `npm test`? Maybe in the wiki

crookedneighbor:
@KristianTashkov This should be a PR into `develop` and the command to install the correct version of node should just be `nvm install`. There's an `.nvmrc` which dictates which version to use. In the `develop` branch it is set to `0.10` and in the `api-v3` branch it's set to `4.2`

crookedneighbor:
Regarding the `--no-bin-links` thing, that only affects users with older versions of virtualbox (some that cannot upgrade because their OS does not support the newer version of vritualbox).

There's a couple ways we could tackle this.
1. Add all the test commands as npm scripts and install the modules globally (pinning to the same version range as the package.json). Change the gulp tasks to run those npm scripts instead. If node can't find a local module, it'll bubble up to use the global module. 
2. Add a note to the wiki about the issue, and if you meet the version requirements, you can just install modules locally.

```
npm install mocha gulp istanbul karma 
```
1. Stop supporting development for versions of virtualbox that have trouble with symlinks.

I think option 1 is the way to go and the devs and staff should discuss internally when we will drop dev support for that version of virtualbox.

That said, I think these changes are outside of the scope of this PR.

KristianTashkov:
@crookedneighbor Should I close this PR and make a new one targeting develop?

PR #6489, opened 2016-01-10
Use nvm to install node 4 to vagrant env

KristianTashkov:
Same as [previous PR](https://github.com/HabitRPG/habitrpg/pull/6486) but targeting develop.

### DIFF
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -16,5 +16,5 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.hostname = "habitrpg"
   config.vm.network "forwarded_port", guest: 3000, host: 3000, auto_correct: true
   config.vm.usable_port_range = (3000..3050)
-  config.vm.provision :shell, :path => "vagrant.sh"
+  config.vm.provision :shell, :path => "vagrant_scripts/vagrant.sh"
 end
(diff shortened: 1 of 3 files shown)


crookedneighbor:
The install.sh file isn't executable and never gets called in the provisioning process.

KristianTashkov:
Hmm I thought I made it executable. For the provisioning process it isn't meant to be. I was gonna put it as a command you need to run after you `vagrant ssh` the first time. I'm not sure if it will work if I run it with `sudo -H -u vagrant` from `vagrant.sh`. Will test tonight.

KristianTashkov:
@crookedneighbor I fixed the executable problem.
I tried running `install.sh` from the provision script, but `nvm` is being picky when being installed for multiple users and I couldn't make it work. In the first couple of days while trying to make this PR work I had a solution which installed `nvm` from the provision script and made it useable from the vagrant user, but later scrapped it because installing `nvm` properly from the right user seemed cleaner and better than hacking path variables to make it work.
If you think running an extra command when you login for the first time(I can update the wiki to reflect this change) is too much, I can try to recreate the `nvm` installation but I don't really think it's worth it. Also having a handy install script that sets up nvm/node/the project, when you switch between develop (node 0.10) and api-v3 (node 4) is really handy.

crookedneighbor:
Tier awarded!

PR #6496, opened 2016-01-11
Converted posts tests to async/await syntax and updated tests

TheHollidayInn:
-Update the test syntax
-Added some new tests and functionality that were to-dos from version 2.

### DIFF
--- a/test/api/v3/integration/groups/POST-groups.test.js
+++ b/test/api/v3/integration/groups/POST-groups.test.js
@@ -6,21 +6,34 @@ import {
 describe('POST /group', () => {
   let user;
 
-  beforeEach(() => {
-    return generateUser().then((generatedUser) => {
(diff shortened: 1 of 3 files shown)


test/api/v3/integration/groups/POST-groups.test.js
```diff
@@ -6,21 +6,34 @@ import {
...
-  beforeEach(() => {
-    return generateUser().then((generatedUser) => {
-      user = generatedUser;
+  beforeEach(async () => {
+    user = await generateUser();
+  });
+
+  context('All Groups', () => {
+    it('it returns validation error when type is not provided', async () => {
+      let userToCreateGroup = await generateUser({balance: 1});
```
crookedneighbor, line 15: use `await user.update({balance: 1});` instead.

PR #6502, opened 2016-01-12
Fixes additive boss damage from checking/unchecking tasks - fixes #5648

### LINKED ISSUE #5648: Checked and unchecked tasks still cause boss quest damage and task-based attribute training
Just had a day where I completed a large to-do - 4 checklist items, task value -150.96. Over the course of the day I checked and unchecked it three times. Once as Medium difficulty, another as Easy, once as Medium with a 106% crit and lastly the final completed value as a Medium to-do with no crit.

After checking it three times - Medium, Easy, Medium Crit - I had an estimated 60 damage on the boss on Data Display. This stayed after I unchecked it.
Checking it the fourth and final time, the damage went up to around 85.

I had figured this would fix itself in cron and deal what should be around my general daily damage of 9 + the 25 damage from that one task but the next day I did 83 damage which implies it's stacking up each time.

I might have a look at testing this in a closed environment but I'll need to get some gems first for a boss with enough health.

User ID: 4352e75b-0c23-4aa0-b64c-ea82543aa2a3
Group ID: 3918426d-5a0a-42bf-8247-737e890fe32d

**Note** added June 2018: This issue also affect attribute training points for task-based automatic stat allocation.

doyler:
Note that there is no way to easily calculate the PREVIOUS delta, so while this
won't quite allow additive damage, it won't remove all of the damage
done by checking and unchecking.


crookedneighbor:
This should be covered by tests before it can be merged in.

crookedneighbor:
@doyler Will you be adding tests for this behavior?

doyler:
@crookedneighbor I will, sorry about that. Went away for holiday and now still need to figure out how in the world these test cases are setup and organized. Thanks!

doyler:
@crookedneighbor Is this where I should get an example of some of the tests/where I should be putting mine? - https://github.com/HabitRPG/habitrpg/blob/develop/test/common/dailies.js

This doesn't seem to cover anything but dailies, and has nothing related to damage as far as I can tell yet. Thanks again.

nasfarley88:
Does this still need (re?)doing?

PR #6504, opened 2016-01-12
[API v3] Members routes

paglias:
This PR implements the following route:
- getMember to return a member pubic info given its id
- get to return users members of a group (with pagination and option to return all public info in case of a party and not only `profile.name`)
- getInvitesForGroup to return users invited to a group (with pagination)
- get members for challenges too

Missing tests.


paglias:
Ready to be merged

Issue #6431, opened 2015-12-31
Convert API (v3) Integration tests to use the new `await` syntax


crookedneighbor:
[This PR illustrates](https://github.com/HabitRPG/habitrpg/pull/6423/files) how to do the conversion.

Notice that each `before/beforeEach/it` blocks callback function has been replaced with `async function`

This allows use to use the `await` keyword to wait for the promise to resolve and it returns the value that the promise resolves with.

tests in test/api/v3/integration in the api-v3-groups branch that need to be converted
#### chat
- [ ] `│   ├── GET-chat.test.js`
- [ ] `│   ├── POST-chat.flag.test.js`
- [ ] `│   ├── POST-chat.like.test.js`
- [ ] `│   └── POST-chat.test.js`
#### groups
- [ ] `│   └── POST-groups.test.js`
#### notFound
- [ ] `├── notFound.test.js`
#### tags
- [ ] `│   ├── DELETE-tags_id.test.js`
- [ ] `│   ├── GET-tags.test.js`
- [ ] `│   ├── GET-tags_id.test.js`
- [ ] `│   ├── POST-tags.test.js`
- [ ] `│   └── PUT-tags_id.test.js`
#### tasks
- [ ] `│   ├── DELETE-tasks_id.test.js`
- [ ] `│   ├── GET-tasks.test.js`
- [ ] `│   ├── GET-tasks_id.test.js`
- [ ] `│   ├── POST-tasks.test.js`
- [ ] `│   ├── POST-tasks_id_score_direction.test.js`
- [ ] `│   ├── PUT-tasks_id.test.js`
#### tasks/checklists
- [ ] `│   │   ├── DELETE-tasks_taskId_checklist_itemId.test.js`
- [ ] `│   │   ├── POST-tasks_taskId_checklist.test.js`
- [ ] `│   │   ├── POST-tasks_taskId_checklist_itemId_score.test.js`
- [ ] `│   │   └── PUT-tasks_taskId_checklist_itemId.test.js`
#### tasks/tags
- [ ] `│       ├── DELETE-tasks_taskId_tags_tagId.test.js`
- [ ] `│       └── POST-tasks_taskId_tags_tagId.test.js`
#### user
- [ ] `├── GET-user.test.js`
#### user/auth
- [ ] `└── POST-register_local.test.js`

TheHollidayInn:
I can get all the group and chat tests converted/created. I am already working on a PR with many group routes, so it would be a simple extension.

[...]

PR #6505, opened 2016-01-12
Refactor integration tests to use await syntax

GeorgeSG:
UserID: a187eae5-b3d2-479c-99c5-3e00b3857e24
See #6431.
- Refactored all `tasks`, `tags`, `user` and the `notFound` tests to use the `await` syntax instead of callbacks.
- Changed all uses of `return expect` to `await expect` for consistency

The only test I couldn't refactor is in `PUT-tasks_id.test.js`:23 - `ignores setting _id, type, userId, history, createdAt, updatedAt, challenge, completed, streak,  dateCompleted fields`. For some reason when I do this:

``` js
let savedTask = await user.put(`/tasks/${task._id}`, {
   // ...
});

// ...
expect(savedTask.completed).to.equal(false);
expect(savedTask.streak).to.equal(0);
expect(savedTask.streak).not.to.equal('never');
```

`savedTask.completed` and `savedTask.streak` are `undefined`, which makes the assertion fail. Not sure why this is happening, so I left the test as-is.

### DIFF
--- a/test/api/v3/integration/notFound.test.js
+++ b/test/api/v3/integration/notFound.test.js
@@ -1,10 +1,10 @@
 import { requester } from '../../../helpers/api-integration.helper';
 
 describe('notFound Middleware', () => {
-  it('returns a 404 error when the resource is not found', () => {
+  it('returns a 404 error when the resource is not found', async () => {
     let request = requester().get('/api/v3/dummy-url');
 
-    return expect(request).to.eventually.be.rejected.and.eql({
+    await expect(request).to.eventually.be.rejected.and.eql({
       code: 404,
       error: 'NotFound',
       message: 'Not found.',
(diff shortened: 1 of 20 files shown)


crookedneighbor:
Thanks! Tier awarded! You can read more about [contributor tiers and rewards here](http://habitrpg.wikia.com/wiki/Contributor_Rewards).

crookedneighbor:
The reason you couldn't fix the PUT test is because it was failing since the promise never got returned :facepalm:

crookedneighbor:
Failing without reporting that it was failing, that is.

paglias:
@GeorgeSG thanks for this PR! We've been very impressed, if you'd like to do something else you may take a look at #6509. If you have any question just ping me or @crookedneighbor

GeorgeSG:
Thanks @paglias, I'll take a look :)

PR #6506, opened 2016-01-13
Added intitial group leave tests

TheHollidayInn:
Challenge and quests tests are pending the routes. Other than that, I made some interesting changes, so feel free to ask away.

### DIFF
--- a/test/api/v3/integration/groups/POST-groups_groupId_leave.js
+++ b/test/api/v3/integration/groups/POST-groups_groupId_leave.js
@@ -66,7 +66,7 @@ describe('POST /groups/:groupId/leave', () => {
 
           let userWithoutChallenges = await leader.get('/user');
 
-          expect(userWithoutChallenges.challenges.indexOf(challenge._id)).to.equal(-1);
+          expect(userWithoutChallenges.challenges).to.not.include(challenge._id);
           expect(userWithoutChallenges.tasksOrder.habits.length).to.equal(0);
         });
 
@@ -75,73 +75,127 @@ describe('POST /groups/:groupId/leave', () => {
 
           let userWithChallenges = await leader.get('/user');
 
-          expect(userWithChallenges.challenges.indexOf(challenge._id)).to.equal(-1);
+          expect(userWithChallenges.challenges).to.not.include(challenge._id);
           expect(userWithChallenges.tasksOrder.habits.length).to.equal(1);
         });
       });
 
-      xit('prevents quest leader from leaving a groupToLeave', async () => {});
-      xit('prevents a user from leaving during an active quest', async () => {});
+      it('prevents quest leader from leaving a groupToLeave');
+      it('prevents a user from leaving during an active quest');
     });
   });
 
   context('Leaving a group as the last member', () => {
-    it('removes a group and invitations when the last member leaves a private guild', async () => {
-      let { group, groupLeader, invitees } = await createAndPopulateGroup({
-        groupDetails: {
-          name: 'Test Private Guild',
-          type: 'guild',
-        },
-        invites: 1,
+    context('private guild', () => {
+      let privateGuild;
+      let leader;
+      let invitedUser;
+
+      beforeEach(async () => {
+        let { group, groupLeader, invitees } = await createAndPopulateGroup({
+          groupDetails: {
+            name: 'Test Private Guild',
+            type: 'guild',
+          },
+          invites: 1,
+        });
+
+        privateGuild = group;
+        leader = groupLeader;
+        invitedUser = invitees[0];
       });
 
-      await groupLeader.post(`/groups/${group._id}/leave`);
+      it('removes a group when the last member leaves', async () => {
+        await leader.post(`/groups/${privateGuild._id}/leave`);
 
-      let groups = await groupLeader.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
-      let userWithoutInvitation = await invitees[0].get('/user');
+        let groups = await leader.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
 
-      expect(_.findIndex(groups, {_id: group._id})).to.equal(-1);
-      expect(userWithoutInvitation.invitations.guilds).to.be.empty;
-      await expect(checkExistence('groups', group._id)).to.eventually.equal(false);
+        expect(_.findIndex(groups, {_id: privateGuild._id})).to.equal(-1);
+        await expect(checkExistence('groups', privateGuild._id)).to.eventually.equal(false);
+      });
+
+      it('removes invitations when the last member leaves', async () => {
+        await leader.post(`/groups/${privateGuild._id}/leave`);
+
+        let userWithoutInvitation = await invitedUser.get('/user');
+
+        expect(userWithoutInvitation.invitations.guilds).to.be.empty;
+      });
     });
 
-    it('removes invitations but not the group when the last member leaves a public guild', async () => {
-      let { group, groupLeader, invitees } = await createAndPopulateGroup({
-        groupDetails: {
-          name: 'Test Public Guild',
-          type: 'guild',
-          privacy: 'public',
-        },
-        invites: 1,
+    context('public guild', () => {
+      let publicGuild;
+      let leader;
+      let invitedUser;
+
+      beforeEach(async () => {
+        let { group, groupLeader, invitees } = await createAndPopulateGroup({
+          groupDetails: {
+            name: 'Test Public Guild',
+            type: 'guild',
+            privacy: 'public',
+          },
+          invites: 1,
+        });
+
+        publicGuild = group;
+        leader = groupLeader;
+        invitedUser = invitees[0];
       });
 
-      await groupLeader.post(`/groups/${group._id}/leave`);
+      it('keeps the group when the last member leaves a public guild', async () => {
+        await leader.post(`/groups/${publicGuild._id}/leave`);
 
-      let groups = await groupLeader.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
-      let userWithoutInvitation = await invitees[0].get('/user');
+        let groups = await leader.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
 
-      expect(_.findIndex(groups, {_id: group._id})).to.be.above(-1);
-      expect(userWithoutInvitation.invitations.guilds).to.be.empty;
-      await expect(checkExistence('groups', group._id)).to.eventually.equal(true);
+        expect(_.findIndex(groups, {_id: publicGuild._id})).to.be.above(-1);
+        await expect(checkExistence('groups', publicGuild._id)).to.eventually.equal(true);
+      });
+
+      it('keeps the invitations when the last member leaves a public guild', async () => {
+        await leader.post(`/groups/${publicGuild._id}/leave`);
+
+        let userWithoutInvitation = await invitedUser.get('/user');
+
+        expect(userWithoutInvitation.invitations.guilds).to.not.be.empty;
+      });
     });
 
-    it('removes a group and invitations when the last member leaves a party', async () => {
-      let { group, groupLeader, invitees } = await createAndPopulateGroup({
-        groupDetails: {
-          name: 'Test Party',
-          type: 'party',
-        },
-        invites: 1,
+    context('party', () => {
+      let party;
+      let leader;
+      let invitedUser;
+
+      beforeEach(async () => {
+        let { group, groupLeader, invitees } = await createAndPopulateGroup({
+          groupDetails: {
+            name: 'Test Party',
+            type: 'party',
+          },
+          invites: 1,
+        });
+
+        party = group;
+        leader = groupLeader;
+        invitedUser = invitees[0];
       });
 
-      await groupLeader.post(`/groups/${group._id}/leave`);
+      it('removes a group when the last member leaves a party', async () => {
+        await leader.post(`/groups/${party._id}/leave`);
 
-      await groupLeader.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
-      let userWithoutInvitation = await invitees[0].get('/user');
+        await leader.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
 
-      expect(groupLeader.party._id).to.equal(undefined);
-      expect(userWithoutInvitation.invitations.party).to.be.empty;
-      await expect(checkExistence('party', group._id)).to.eventually.equal(false);
+        expect(leader.party._id).to.equal(undefined);
+        await expect(checkExistence('party', party._id)).to.eventually.equal(false);
+      });
+
+      it('removes invitations when the last member leaves a party', async () => {
+        await leader.post(`/groups/${party._id}/leave`);
+
+        let userWithoutInvitation = await invitedUser.get('/user');
+
+        expect(userWithoutInvitation.invitations.party).to.be.empty;
+      });
     });
   });
 });
--- a/test/api/v3/integration/groups/POST-groups_groupId_leave.js
+++ b/test/api/v3/integration/groups/POST-groups_groupId_leave.js
@@ -1,6 +1,7 @@
 import {
   generateUser,
-} from '../../../../helpers/api-integration.helper';
+  checkExistence,
+} from '../../../../helpers/api-v3-integration.helper';
 
 describe('POST /groups/:groupId/leave', () => {
   let user;
@@ -48,9 +49,9 @@ describe('POST /groups/:groupId/leave', () => {
       let groups = await user.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
       let userWithoutInvitation = await userToInvite.get('/user');
 
-      // @TODO: Is there a way we can check using an admin account to see if private groups were deleted?
       expect(_.findIndex(groups, {_id: guild._id})).to.equal(-1);
       expect(userWithoutInvitation.invitations.guilds).to.be.empty;
+      expect(await checkExistence('groups', guild._id)).to.equal(false);
     });
   });
 
@@ -78,7 +79,7 @@ describe('POST /groups/:groupId/leave', () => {
       await userToInvite.post(`/groups/${party._id}/leave`);
 
       let userWithoutParty = await userToInvite.get('/user');
-      expect(userWithoutParty.party._id).to.be.a('null');
+      expect(userWithoutParty.party._id).to.not.exist;
     });
 
     xit('prevents quest leader from leaving a guild', async () => {});
--- a/test/api/v3/integration/groups/POST-groups_groupId_leave.js
+++ b/test/api/v3/integration/groups/POST-groups_groupId_leave.js
@@ -2,6 +2,7 @@ import {
   generateUser,
   generateChallenge,
   checkExistence,
+  sleep,
 } from '../../../../helpers/api-v3-integration.helper';
 
 describe('POST /groups/:groupId/leave', () => {
@@ -18,7 +19,7 @@ describe('POST /groups/:groupId/leave', () => {
 
     beforeEach(async () => {
       guild = await user.post('/groups', {
-        name: 'Test Public Guild',
+        name: 'Test Guild',
         type: 'guild',
       });
       await user.post(`/groups/${guild._id}/invite`, {
@@ -65,6 +66,7 @@ describe('POST /groups/:groupId/leave', () => {
           text: 'test habit',
           type: 'habit',
         });
+        await sleep(2);
       });
 
       it('removes all challenge tasks when keep parameter is set to remove', async () => {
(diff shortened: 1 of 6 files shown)


test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,86 @@
...
+    group = await user.post('/groups', {
+      name: groupName,
+      type: groupType,
+    });
+    await user.post(`/groups/${group._id}/invite`, {
+      uuids: [userToInvite._id],
+    });
+  });
+
+  describe('leaving guilds', () => {
```
KristianTashkov: Shouldn't nested `describe`s be `context` blocks?

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,86 @@
...
+      await user.post(`/groups/${party._id}/invite`, {
+        uuids: [userToInvite._id],
+      });
+      await userToInvite.post(`/groups/${party._id}/join`);
+    });
+
+    it('lets user leave a party', async () => {
+      await userToInvite.post(`/groups/${party._id}/leave`);
+
+      let userWithOutGroup = await userToInvite.get('/user');
```
KristianTashkov: [nitpick] without is a single word, no need to uppercase the `O` [/nitpick]

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,86 @@
...
+
+      expect(userWithOutGroup.guilds).to.be.empty;
```
KristianTashkov: shouldn't this check `party` to be empty?

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,86 @@
+import {
+  generateUser,
+} from '../../../../helpers/api-integration.helper';
+
+describe('POST /groups/:groupId/leave', () => {
+  let user;
+  let userToInvite;
+  let group;
```
KristianTashkov: Why are these 3 `group` variables in this scope when they are used only in the guilds context? Shouldn't `group` be called `guild` as in the `party` tests? And do we need `groupName` and `groupType` to be variables when we are never asserting on them(the way the party name and type are just hardcoded in the `beforeEach`)?

test/api/v3/integration/groups/POST-groups_groupId_leave.js
TheHollidayInn: Yea, this should check party._id. I must have changed this by accident.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
TheHollidayInn: I'm unsure. I'm not familiar with context blocks vs describe. I'll look it up.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
TheHollidayInn: Well, variables should be abstracted when used in multiple places, not just when they are asserted on. But, to answer your questions I was using these variables when starting the pending test. (I stopped to wait in some changes.) I'm fine removing the global variables for now since the context doesn't make sense.

website/src/models/group.js
```diff
@@ -440,69 +440,56 @@ schema.statics.bossQuest = function bossQuest (user, progress) {
...
-      }
+  let challenges = await Challenge.find({
+    _id: {$in: user.challenges}, // Challenges I am in
+    group: group._id, // that belong to the group I am leaving
+  });
 
-      // If the leader is leaving (or if the leader previously left, and this wasn't accounted for)
-      let leader = group.leader;
+  await Challenge.update(
+    {_id: {$in: _.pluck(challenges, '_id')}},
```
paglias: @TheHollidayInn in this case the ids should be the same one stored on user.challenges, can we use those?

website/src/models/group.js
```diff
@@ -440,69 +440,56 @@ schema.statics.bossQuest = function bossQuest (user, progress) {
...
+    {$pull: {members: user._id}},
```
paglias: we don't store members anymore

website/src/models/group.js
```diff
@@ -440,69 +440,56 @@ schema.statics.bossQuest = function bossQuest (user, progress) {
...
-    firebase.removeUserFromGroup(group._id, user._id);
-    return; // TODO ok not to return promise?
-  }).catch(err => { // TODO do we have to catch err if we return the promise?
-    throw err;
-  });
+  // otherwise just remove a member TODO create User.methods.removeFromGroup?
+  if (group.type === 'guild') {
+    promises.push(User.update({_id: user._id}, {$pull: {guilds: group._id } }).exec());
+  } else {
+    promises.push(User.update({_id: user._id}, {$set: {'party._id': undefined } }).exec());
```
paglias: I'm unsure if here we should also cleanup the whole user.party object

TheHollidayInn: Is there a fast way of doing this rather than setting all the values back to default inside the $set object?

test/api/v3/integration/groups/POST-groups_groupId_leave.js
crookedneighbor: context is just an alias for describe, so functionally, they do the same thing.

In general, our style is to have a describe block describe a method (or in this case, a route), and context blocks describe a specific circumstance for differing behavior.

So yes, this should be a context block.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
crookedneighbor: In the case of tests, it only makes sense to extract the values into variables if it's meaningful to do so. These are just dummy values, so it'd be better to remove them.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,87 @@
+import {
+  generateUser,
+} from '../../../../helpers/api-integration.helper';
```
crookedneighbor: This will need to change to `../../../../helpers/api-v3-integration.helper';`

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,87 @@
...
+      expect(guildWithNewLeader.leader).to.equal(userToInvite._id);
+    });
+
+    it('removes a group and invitations when the last member leaves', async () => {
+      await user.post(`/groups/${guild._id}/leave`);
+
+      let groups = await user.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
+      let userWithoutInvitation = await userToInvite.get('/user');
+
+      // @TODO: Is there a way we can check using an admin account to see if private groups were deleted?
```
crookedneighbor: https://github.com/HabitRPG/habitrpg/blob/1f4e58e5cc44abbd3c866a4faefd8414f5a417f9/test/helpers/api-v3-integration.helper.js#L72-L91

website/src/models/group.js
crookedneighbor: Could we just set `party` to `{}`?

paglias: That's what I've been doing in other places, I'm unsure though if it requires to be `.markModified` or not, you should probably check that

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,88 @@
...
+
+    it('removes a group and invitations when the last member leaves', async () => {
+      await user.post(`/groups/${guild._id}/leave`);
+
+      let groups = await user.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
+      let userWithoutInvitation = await userToInvite.get('/user');
+
+      expect(_.findIndex(groups, {_id: guild._id})).to.equal(-1);
+      expect(userWithoutInvitation.invitations.guilds).to.be.empty;
+      expect(await checkExistence('groups', guild._id)).to.equal(false);
```
KristianTashkov: In other tests we write it as `await expect(checkExistence('groups', guild._id)).to.eventually.equal(false)`, but I'm not sure if there is a good reason for it, @crookedneighbor ?

website/src/models/group.js
```diff
@@ -439,69 +439,55 @@ schema.statics.bossQuest = function bossQuest (user, progress) {
...
-      } else {
-        user.party._id = undefined;
-      }
+  let challenges = await Challenge.find({
+    _id: {$in: user.challenges}, // Challenges I am in
+    group: group._id, // that belong to the group I am leaving
+  });
+
+  await Challenge.update(
+    {_id: {$in: _.pluck(user.challenges, '_id')}},
```
paglias: no need to use _.pluck, use.challenges is already an array of ids

website/src/models/group.js
```diff
@@ -439,69 +439,55 @@ schema.statics.bossQuest = function bossQuest (user, progress) {
...
+    {multi: true}
+  );
 
-      // If the leader is leaving (or if the leader previously left, and this wasn't accounted for)
-      let leader = group.leader;
+  //  remove user from challenges
+  let challengesUserWasRemovedFrom = challenges.map(chal => {
+    let i = user.challenges.indexOf(chal._id);
+    if (i !== -1) user.challenges.splice(i, 1);
+    return user.unlink({cid: chal._id, keep});
```
paglias: one this branch is up to date with api-v3-groups this should be user.unlinkChallengeTasks(chal._id, keep);

paglias:
@TheHollidayInn expect for merging in the new updates this is ready to be merged?

TheHollidayInn:
Yep! I'm going to merge and complete the challenge todos right now.

website/src/controllers/api-v3/tasks.js
```diff
@@ -102,10 +102,11 @@ api.createChallengeTasks = {
     if (challenge.leader !== user._id) throw new NotAuthorized(res.t('onlyChalLeaderEditTasks'));
 
     let tasks = await _createTasks(req, res, user, challenge);
-    res.respond(201, tasks.length === 1 ? tasks[0] : tasks);
 
     // If adding tasks to a challenge -> sync users
-    if (challenge) challenge.addTasks(tasks); // TODO catch/log
+    if (challenge) await challenge.addTasks(tasks); // TODO catch/log
```
paglias: this has to be placed after res.respond. In big challenges it may take a while so we want this to run in background

website/src/models/group.js
```diff
@@ -98,30 +98,30 @@ schema.statics.sanitizeUpdate = function sanitizeUpdate (updateObj) {
 }*/
 
 // TODO test
-schema.pre('remove', true, function preRemoveGroup (next, done) {
+schema.pre('remove', true, async function preRemoveGroup (next, done) {
```
paglias, line 101: Hmm I hadn't thought using async functions here would have worked... The only thing is that it's not inside a try/catch block which means errors will be swallowed. You should wrap the function body in a try/catch block and in case of errors pass the error to `done(err)`

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,114 @@
...
+
+        let userWithChallenges = await user.get('/user');
+
+        expect(userWithChallenges.challenges.indexOf(challenge._id)).to.equal(-1);
+        expect(userWithChallenges.tasksOrder.habits.length).to.equal(1);
+      });
+    });
+  });
+
+  context('leaving parties', () => {
```
paglias: for parties could you replicate the same tests we have for guilds so we're sure it works correctly? Mainly regarding challenges tasks and invitations being removed

website/src/controllers/api-v3/tasks.js
TheHollidayInn: Ah, okay. I thought it was a mistake.

website/src/controllers/api-v3/groups.js
```diff
@@ -303,8 +305,8 @@ api.leaveGroup = {
       }
     }
 
-    await group.leave(user, req.query.keep);
-    res.respond(200, {});
+    let leave = await group.leave(user, req.query.keep);
+    res.respond(200, leave);
```
paglias: what does leave return?

website/src/models/group.js
TheHollidayInn, line 101: Okay, I can do that.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
TheHollidayInn: Yep!

website/src/controllers/api-v3/groups.js
TheHollidayInn: Just the results of group.leave - which don't seem to have to much value. We don't really assert on this ever. (Doesn't mean we shouldn't)

website/src/controllers/api-v3/tasks.js
TheHollidayInn: So.. I actually need this to be before the response to make my challenge test pass.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
crookedneighbor: They should be functionally equivalent, but we should stick to a particular convention. Let's go with the `await` on the outside. I find it more readable.

website/src/controllers/api-v3/groups.js
crookedneighbor: Let's just return an empty object for now. That way we can add stuff later if it becomes useful.

TheHollidayInn:
Updated, but I think we will need to take a look at that challenge route for performance. I'll look into it tomorrow.

website/src/controllers/api-v3/tasks.js
crookedneighbor: No, @paglias is right, this should be running in the background. Probably best to thoroughly test the behavior with unit tests, but their should be some testing in integration tests as well. Probably need to add a promisified sleep function to the helpers.

Something like:

``` js
function sleep (seconds) {
  let milliseconds = seconds * 1000;

  return new Promise((resolve) => {
    setTimeout(resolve, milliseconds);
  });
}
```

Then in the tests, you can do 

``` js
await sleep(2);
// fetch tasks and assert
```

paglias:
@TheHollidayInn what are the perf problem?

website/src/controllers/api-v3/tasks.js
TheHollidayInn: Perfect. Thanks!

TheHollidayInn:
Updated!

paglias:
This looks good to me, thanks @TheHollidayInn !

@crookedneighbor I'll wait to merge in case you wanted to change something

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,167 @@
...
+    });
+
+    it('removes a group and invitations when the last member leaves', async () => {
+      await user.post(`/groups/${guild._id}/leave`);
+
+      let groups = await user.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
+      let userWithoutInvitation = await userToInvite.get('/user');
+
+      expect(_.findIndex(groups, {_id: guild._id})).to.equal(-1);
+      expect(userWithoutInvitation.invitations.guilds).to.be.empty;
```
crookedneighbor: I thought we _were_ leaving invitations on the user for public guilds?

crookedneighbor: Oh, it's not a public guild, but the guild name is 'test public guild' so I was confused.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,167 @@
...
+    user = await generateUser({balance: 2});
+    userToInvite = await generateUser();
+  });
+
+  context('leaving guilds', () => {
+    let guild;
+
+    beforeEach(async () => {
+      guild = await user.post('/groups', {
+        name: 'Test Public Guild',
```
crookedneighbor: This is creating a private guild, not a public guild.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
paglias: yeah invitations are removed in a pre('remove') middleware so only for private groups that gets deleted

TheHollidayInn:
@crookedneighbor Fixed the name.

website/src/models/group.js
```diff
@@ -98,30 +98,15 @@ schema.statics.sanitizeUpdate = function sanitizeUpdate (updateObj) {
...
-        user.invitations.guilds.splice(i, 1);
-      }
-
-      return user.save();
-    }));
-  })
-  .then(done)
-  .catch(done);
+  try {
+    await group.removeGroupInvitations();
```
crookedneighbor, line 105: I like this a lot. Much clearer what is happening in here.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,147 @@
...
+          });
+          await sleep(2);
+        });
+
+        it('removes all challenge tasks when keep parameter is set to remove', async () => {
+          await leader.post(`/groups/${groupToLeave._id}/leave?keep=remove-all`);
+
+          let userWithoutChallenges = await leader.get('/user');
+
+          expect(userWithoutChallenges.challenges.indexOf(challenge._id)).to.equal(-1);
```
crookedneighbor: This is better written as:

``` js
expect(userWithoutChallenges.challenges).to.not.include(challenge._id);
```

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,147 @@
...
+          expect(userWithoutChallenges.tasksOrder.habits.length).to.equal(0);
```
crookedneighbor, line 70: This should probably assert on the task id not being included in the habits array.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,147 @@
...
+        });
+
+        it('keeps all challenge tasks when keep parameter is not set', async () => {
+          await leader.post(`/groups/${groupToLeave._id}/leave`);
+
+          let userWithChallenges = await leader.get('/user');
+
+          expect(userWithChallenges.challenges.indexOf(challenge._id)).to.equal(-1);
+          expect(userWithChallenges.tasksOrder.habits.length).to.equal(1);
```
crookedneighbor, line 79: Same comments as above. Only, this should assert that the task id does exist in the habits array.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,147 @@
...
+        });
+      });
+
+      xit('prevents quest leader from leaving a groupToLeave', async () => {});
+      xit('prevents a user from leaving during an active quest', async () => {});
```
crookedneighbor: These are just not implemented yet, right?

Just so you know, you can write a pending test like this:

``` js
it('prevents quest leader from leaving a groupToLeave');
```

The way you've done it works just as well, but this is a little simpler to write.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,147 @@
...
+    });
+  });
+
+  context('Leaving a group as the last member', () => {
+    it('removes a group and invitations when the last member leaves a private guild', async () => {
```
crookedneighbor: These should be separate assertions. One that the group is removed, and another that the invitations are removed.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,147 @@
...
+
+      let groups = await groupLeader.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
+      let userWithoutInvitation = await invitees[0].get('/user');
+
+      expect(_.findIndex(groups, {_id: group._id})).to.equal(-1);
+      expect(userWithoutInvitation.invitations.guilds).to.be.empty;
+      await expect(checkExistence('groups', group._id)).to.eventually.equal(false);
+    });
+
+    it('removes invitations but not the group when the last member leaves a public guild', async () => {
```
crookedneighbor: @paglias @Alys Is this the desire behavior? Should the invitation remain on the user if it is for a public guild?

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -0,0 +1,147 @@
...
+
+      let groups = await groupLeader.get('/groups?type=party,privateGuilds,publicGuilds,tavern');
+      let userWithoutInvitation = await invitees[0].get('/user');
+
+      expect(_.findIndex(groups, {_id: group._id})).to.be.above(-1);
+      expect(userWithoutInvitation.invitations.guilds).to.be.empty;
+      await expect(checkExistence('groups', group._id)).to.eventually.equal(true);
+    });
+
+    it('removes a group and invitations when the last member leaves a party', async () => {
```
crookedneighbor: Again, these should be separate tests. One for the group removal, and one for the invitation removal.

website/src/controllers/api-v3/groups.js
```diff
@@ -251,6 +251,8 @@ api.joinGroup = {
 
     if (group.memberCount === 0) group.leader = user._id; // If new user is only member -> set as leader
 
+    group.memberCount += 1;
```
crookedneighbor, line 254: Good call.

website/src/models/group.js
```diff
@@ -150,6 +135,29 @@ schema.statics.getGroup = function getGroup (options = {}) {
   // TODO purge chat flags info? in tojson?
 };
 
+schema.methods.removeGroupInvitations = async function removeGroupInvitations () {
+  let group = this;
+
+  // Remove invitations when group is deleted
```
crookedneighbor: I don't think this comment is necessary.

website/src/models/group.js
```diff
@@ -150,6 +135,29 @@ schema.statics.getGroup = function getGroup (options = {}) {
...
+  // TODO verify it works for everything
```
crookedneighbor: Is this comment needed now that we have tests?

test/api/v3/integration/groups/POST-groups_groupId_leave.js
TheHollidayInn, line 70: The task id is changed because the task is cloned from the challenge. It is not the same challenge that is on the challenge. I was asserting on the task id before.

website/src/models/group.js
```diff
@@ -452,71 +460,53 @@ schema.statics.bossQuest = function bossQuest (user, progress) {
...
-        user.save(),
-      ]);
-    })(),
-  ]).then(() => {
-    firebase.removeUserFromGroup(group._id, user._id);
-    return; // TODO ok not to return promise?
-  }).catch(err => { // TODO do we have to catch err if we return the promise?
-    throw err;
-  });
+  if (group.memberCount === 1) {
```
crookedneighbor: This should be able to be simplified to:

``` js
if (group.memberCount === 1 && group.privacy === 'private') {
  return await group.remove();
}
```

website/src/models/group.js
```diff
@@ -452,71 +460,53 @@ schema.statics.bossQuest = function bossQuest (user, progress) {
...
+  // If user is the last one in group and group is private, delete it
 
-      update.$inc = {memberCount: -1};
-      return Q.all([
-        model.update({_id: group._id}, update).exec(), // eslint-disable-line no-use-before-define
-        user.save(),
-      ]);
-    })(),
-  ]).then(() => {
-    firebase.removeUserFromGroup(group._id, user._id);
```
crookedneighbor, line 515: Where did this logic go?

website/src/models/group.js
crookedneighbor: Since all parties are private.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
TheHollidayInn: Yea, quests aren't implemented. Looks good. I'll use those.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
TheHollidayInn: Separate tests?

website/src/models/group.js
TheHollidayInn, line 515: User's aren't stored on groups anymore, so I thought this was no needed.

crookedneighbor, line 515: That _might_ be true. firebase is the real time chat service we're going to outsource to in the (hopefully) near future, so it's possible this logic is still necessary. 

@paglias ?

TheHollidayInn, line 515: Ah, I see. I'm reading more docs about it now.

paglias, line 515: Hmm firebase logic will need to be rewritten but it's not something we want to do right now and also it's disabled so it won't do anything. I'd say let's keep the calls to the firebase lib anyway, they'll be useful in future

crookedneighbor, line 515: I've added it back in. Will merge shortly.

crookedneighbor:
Merged.

PR #6513, opened 2016-01-13
Add missing tests in group join route

KristianTashkov:
Note: must be merged after https://github.com/HabitRPG/habitrpg/pull/6506 since we need changes in leave route and `memberCount` increment from there.

### DIFF
--- a/test/api/v3/integration/groups/POST-groups_groupId_join.test.js
+++ b/test/api/v3/integration/groups/POST-groups_groupId_join.test.js
@@ -16,7 +16,39 @@ describe('POST /group/:groupId/join', () => {
     });
   });
 
-  context('Accepting invitation to a private guild', () => {
+  context('Joining a public guild', () => {
+    let user, joiningUser, publicGuild;
+
+    beforeEach(async () => {
+      let {group, groupLeader} = await createAndPopulateGroup({
+        groupDetails: {
+          name: 'Test Guild',
+          type: 'guild',
+          privacy: 'public',
+        },
+      });
+
+      publicGuild = group;
+      user = groupLeader;
+      joiningUser = await generateUser();
+    });
+
+    it('allows non-invited users to join public guilds', async () => {
+      await joiningUser.post(`/groups/${publicGuild._id}/join`);
+
+      await expect(joiningUser.get('/user')).to.eventually.have.property('guilds').to.include(publicGuild._id);
+    });
+
+    it('promotes joining member in a public empty guild to leader', async () => {
+      await user.post(`/groups/${publicGuild._id}/leave`);
+
+      await joiningUser.post(`/groups/${publicGuild._id}/join`);
+
+      await expect(joiningUser.get(`/groups/${publicGuild._id}`)).to.eventually.have.deep.property('leader._id', joiningUser._id);
+    });
+  });
+
+  context('Joining a private guild', () => {
     let user, invitedUser, guild;
 
     beforeEach(async () => {
@@ -44,21 +76,6 @@ describe('POST /group/:groupId/join', () => {
       });
     });
 
-    it('allows non-invited users to join public guilds', async () => {
-      let publicGuild = (await createAndPopulateGroup({
-        groupDetails: {
-          name: 'Test Guild',
-          type: 'guild',
-          privacy: 'public',
-        },
-      })).group;
-
-      let joiningUser = await generateUser();
-      await joiningUser.post(`/groups/${publicGuild._id}/join`);
-
-      await expect(joiningUser.get('/user')).to.eventually.have.property('guilds').and.to.include(publicGuild._id);
-    });
-
     context('User is invited', () => {
       it('allows invited user to join private guilds', async () => {
         await invitedUser.post(`/groups/${guild._id}/join`);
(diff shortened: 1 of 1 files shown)


test/api/v3/integration/groups/POST-groups_groupId_join.test.js
```diff
@@ -130,6 +155,14 @@ describe('POST /group/:groupId/join', () => {
         await expect(invitedUser.get('/user')).to.eventually.not.have.deep.property('invitations.party.id');
       });
 
+      it('increments memberCount when joining party', async () => {
+        let oldMemberCount = party.memberCount;
```
TheHollidayInn, line 159: I think you also need a separate test for testing a member is now in that party.

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
```diff
@@ -16,7 +16,39 @@ describe('POST /group/:groupId/join', () => {
...
+          privacy: 'public',
+        },
+      });
+
+      publicGuild = group;
+      user = groupLeader;
+      joiningUser = await generateUser();
+    });
+
+    it('allows non-invited users to join public guilds', async () => {
```
TheHollidayInn, line 36: You should probably add the test for member count to public guilds to remain consistent with the rest of the guild tests

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
```diff
@@ -44,21 +76,6 @@ describe('POST /group/:groupId/join', () => {
...
-      })).group;
-
-      let joiningUser = await generateUser();
-      await joiningUser.post(`/groups/${publicGuild._id}/join`);
-
-      await expect(joiningUser.get('/user')).to.eventually.have.property('guilds').and.to.include(publicGuild._id);
-    });
-
     context('User is invited', () => {
       it('allows invited user to join private guilds', async () => {
```
TheHollidayInn, line 80: You should add a test to make sure an error is thrown when the user tries to join and they are not invited.

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
```diff
@@ -16,7 +16,39 @@ describe('POST /group/:groupId/join', () => {
...
+        groupDetails: {
+          name: 'Test Guild',
+          type: 'guild',
+          privacy: 'public',
+        },
+      });
+
+      publicGuild = group;
+      user = groupLeader;
+      joiningUser = await generateUser();
```
TheHollidayInn, line 33: Should this use the invitees generate by the createAndPooulateGroup function?

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
```diff
@@ -90,7 +115,7 @@ describe('POST /group/:groupId/join', () => {
     });
   });
 
-  context('Accepting invitation to a party', () => {
+  context('Joining a party', () => {
     let user, invitedUser, party;
```
TheHollidayInn, line 119: Does this syntax pass lint? I'm pretty sure ES6 wants us to declare variables on new lines. I may be wrong.

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
KristianTashkov, line 33: The test bellow checks that the user can still join even if he wasn't invited. I'm not sure if I should add a test that you can still join a public guild if you were invited, thoughts?

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
KristianTashkov, line 36: If i'm checking for every type it makes sense, I will add it later.

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
KristianTashkov, line 159: We are keeping the membership information purely in `user.party._id` if I'm not mistaken? Shouldn't `'allows invited user to join party'` test be enough?

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
KristianTashkov, line 80: There are tests for both parties and private guilds that they reject non-invited users.. am I missing something?

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
KristianTashkov, line 119: Judging by the travis CI and my local lint run it passes. Also this is not even part of the PR and is commited code, so it already passed it :P

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
KristianTashkov, line 36: I guess that's a general question, @crookedneighbor @paglias  should I have duplicate tests for everything that is both true for private and public guilds, or only where there is a difference in the code that changes the behavior depending on the guild type?

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
TheHollidayInn, line 33: Oh, that's right. Never mind then!

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
TheHollidayInn, line 80: Oh, odd. On my phone it was condensed. My mistake.

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
crookedneighbor, line 119: I disabled the requirement for multiple lines in tests. (see https://github.com/HabitRPG/habitrpg/blob/023ee44ddcae865b5a52fc2f4f858b4346c82ae9/test/.eslintrc#L3)

It's not an es6 thing, it's just a style thing. Since tests regularly need a bunch of undefined variables, I thought it was appropriate to remove the rule for it. If y'all have strong feelings about it, I can re-enable and fix the tests.

PR #6518, opened 2016-01-14
New String Variable to avoid translation problem

luveluen:
"optOut" is already used to skip class selection
recently the variable was also used for the checkbox to disable ingame messages.

A phrase equal to "Opt Out" does not exist in all languages, we need two different strings for 
"No, I don't want to choose a class yet" and "Disable messages".
I created optOutMsg for this purpose. (they have the same value in English).


common/locales/en/character.json
```diff
@@ -126,6 +126,7 @@
     "rogueText": "Rogues love to accumulate wealth, gaining more Gold than anyone else, and are adept at finding random items. Their iconic Stealth ability lets them duck the consequences of missed Dailies. Play a Rogue if you find strong motivation from Rewards and Achievements, striving for loot and badges!",
     "healerText": "Healers stand impervious against harm, and extend that protection to others. Missed Dailies and bad Habits don't faze them much, and they have ways to recover Health from failure. Play a Healer if you enjoy assisting others in your Party, or if the idea of cheating Death through hard work inspires you!",
     "optOut": "Opt Out",
+    "optOutMsg: "Opt Out",
```
paglias: @luveluen there's a missing " before the semicolon `"optOutMsg: "Opt Out"` should be `"optOutMsg": "Opt Out"`

Alys:
@luveluen Would it be obvious to the translators which string was used for each purpose?

I'm not sure `optOutMsg` clearly indicates it's related to the private message system because you could interpret "Msg" as meaning this is a generic message about opting out of things. Use something like `optOutOfPMs` maybe? We might also want to change `optOut` to `optOutOfClassSystem` or similar. That would help avoid this same problem in future for other things you can opt out of.

luveluen:
"optOutOfClasses" and "optOutOfPMs" are now the key names.
The quotation mark mistake is fixed.

common/locales/en/character.json
```diff
@@ -125,7 +125,8 @@
...
     "rogueText": "Rogues love to accumulate wealth, gaining more Gold than anyone else, and are adept at finding random items. Their iconic Stealth ability lets them duck the consequences of missed Dailies. Play a Rogue if you find strong motivation from Rewards and Achievements, striving for loot and badges!",
     "healerText": "Healers stand impervious against harm, and extend that protection to others. Missed Dailies and bad Habits don't faze them much, and they have ways to recover Health from failure. Play a Healer if you enjoy assisting others in your Party, or if the idea of cheating Death through hard work inspires you!",
-    "optOut": "Opt Out",
+    "optOutOfClasses": "Opt Out",
+    "optOutOfPMs": "Opt Out",
     "optOutText": "Can't be bothered with classes? Want to choose later? Opt out - you'll be a warrior with no special abilities. You can read about the class system later on the wiki and enable classes at any time under User -> Stats.",
```
crookedneighbor, line 130: This should be changed to `optOutOfClassesText`

crookedneighbor:
I've corrected the `optOutText` string and merged this in.

luveluen:
be0d25d1-9e71-400d-9d78-7c06647f419f
thank you

crookedneighbor:

PR #6521, opened 2016-01-14
Remove tooltip for 'reverse chat messages' checkbox

### LINKED ISSUE #6517: "Show chat messages in reverse order" checkbox is easy to click accidentally
There's a few reports in the Tavern that the "Show chat messages in reverse order" checkbox can be accidentally selected when clicking on the "send chat" or "fetch recent messages" buttons.

citrusella says:

**It appears it happens most often when I approach the button from below; it seems sometimes it's catching the "hitbox" of the reverse link and not letting go when the mouse is moved to the fetch button.**

Ishmaeel says:

**I think it happens because of the tooltip. When you "approach from below", you trigger the tooltip popup (and its hitbox extends to the checkbox) and even after the tooltip appears to be disappeared (heh), it actually lingers for a while there, invisible. When you click the button, you're actually clicking the tooltip.  I can even toggle the checkbox by clicking the chat textbox: 1)hover over checkbox. 2)quickly move the mouse up and click in the textbox where the tooltip was.**

One possible fix is moving the checkbox to a region where there's no other UI elements above it.

GeorgeSG:
Fixes #6517.

### DIFF
--- a/common/locales/en/settings.json
+++ b/common/locales/en/settings.json
@@ -11,7 +11,6 @@
     "dailyDueDefaultView": "Set Dailies default to 'due' tab",
     "dailyDueDefaultViewPop": "With this option set, the Dailies tasks will default to 'due' instead of 'all'",
     "reverseChatOrder": "Show chat messages in reverse order",
-    "reverseChatOrderPop": "Show the messages in the Tarvern, Guild, and Party chats in reverse order, so that the oldest are on top.",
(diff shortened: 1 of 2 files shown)


crookedneighbor:
Merged in with https://github.com/HabitRPG/habitrpg/commit/e761315c74cb7442a103c77590960555ac6dbb9b

Noted toward your next tier.

PR #6527, opened 2016-01-15
added screen-reader labels from the popover title/text

serverless-mom:
# what

Add a bunch of Aria-labels to various shop and equipment items. This is a new version that had such significant changes from the previous PR I'm submitting a new one.
# why

User mentions that much of the site is unlabeled for screen-readers [on a Trello card mainly concerned](https://trello.com/c/pxTcbKuD/381-accessibility-settings) with the chat ordering and sure enough when using a screen readers several areas are impossible to navigate.

If your curious about how widely aria-label works as an option for screen-readers, check [out this handy article](https://www.nczonline.net/blog/2013/04/01/making-accessible-icon-buttons/) that does some testing with multiple browsers.
# known issues

While the labelling now uses a nice bit of Jquery to grab either the popover-title or popover text to make the aria label, that JS is added to the Options pages with the extremely hacky step of including a js file in its folder in `/views` (!!) 

I know this isn't the best approach in general but the site's pattern didn't present an obvious place to store this script otherwise. I'm not certain we want this labelling happening on other pages, so for now it's stored there.

In all cases the label is drawn from the popup title, and in a couple of instances there might be a more useful string to add as the label. Right now for example it doesn't read how many of an item you have, and I don't think the badge number will be picked up by any screen readers. When I tried grabbing the badge values... everything broke so I rolled that back.
# test

0) turn on a screen-reading program (on OSX you can do this with CMD+f5) and see that there are no labels on many equipment and inventory items, they just say things like 'potion button' and 'equipment button'
1) switch to branch
2) go to the inventory pages and see that nothing is messed up on these pages for a sighted user
3) turn on a screen-reading program (on OSX you can do this with CMD+f5) and see those same things have labels!
3a) if you aren't familiar with screen readers (it took me some time to get in control of the Apple version), just inspect the buttons and see they now have reasonable 'aria-label's attached.

### DIFF
--- a/website/views/options/inventory/drops.jade
+++ b/website/views/options/inventory/drops.jade
@@ -215,3 +215,6 @@
               p
                 | {{Content.spells.special[type.key].value}}
                 span(class='shop_gold')
+  script
+    include popoverToAriaLabel.js
+
--- a/website/views/options/inventory/equipment.jade
+++ b/website/views/options/inventory/equipment.jade
@@ -58,3 +58,5 @@
               popover='{{::item.notes()}}', popover-title='{{::item.text()}}',
               popover-trigger='mouseenter', popover-placement='right',
               popover-append-to-body='true')
+  script
+    include popoverToAriaLabel.js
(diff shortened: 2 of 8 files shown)


crookedneighbor:
I'm :+1: on doing this programmatically, but I'm not convinced that this is the best way to handle it.

crookedneighbor:
I'm going to think on this for a little while.

serverless-mom:
I completely agree. I really wasn't able to get a good answer on the best
approach to "run this JQuery on select views." I could much more elegantly
do this on every page, but that feels super sloppy.

I think the best approach may be "manual:" just declare the aria-label
whenever a popover or popover-title is being declared.

The advantage of the manual approach is I could tailor some body text,
label, and role for each menu button e.g. we could label each equipment
button with the items name and include "text" with its stats and whether
it's currently equipped.

It kinda makes sense there's no prescribed method to programmatically
change the views that _you just wrote._

I'll let this PR stay open for a bit in case anyone has a better plan, but
barring any revelations I'll submit a new version with labels that's done
the way I would have if I were writing these views from scratch.

crookedneighbor:
See my last comment. If you want to bring this up to date (now that the v3 api work is done), I'm happy to look at it.

PR #6528, opened 2016-01-15
tests(api): Create ApiDoc class for dealign with api objects

crookedneighbor:
This adds an ApiGroup class for group objects generated by the `generateGroup` method.

The `generateGroup` method now has some defaults, so if it really doesn't matter what the group details are, you can just run:

``` js
group = await generateGroup(leader);
```

If you need to override the defaults, you can run:

``` js
group = await generateGroup(leader, {
  name: 'a new name',
  type: 'guild',
  privacy: 'public',
});
```

If you need to update something about the group that can't be done on creation, you can do

``` js
group = await generateGroup(leader, { /* initial group data */ }, {
  balance: 0,
});
```

The last one should be used sparingly.

### DIFF
--- a/test/helpers/api-v3-integration.helper.js
+++ b/test/helpers/api-v3-integration.helper.js
@@ -16,20 +16,36 @@ i18n.translations = require('../../website/src/libs/api-v3/i18n').translations;
 
 const API_TEST_SERVER_PORT = 3003;
 
-class ApiUser {
+class ApiObject {
   constructor (options) {
     assign(this, options);
+  }
+
+  update (options) {
+    return new Promise((resolve) => {
+      _updateDocument(this._docType, this, options, resolve);
+    });
+  }
+}
+
+class ApiUser extends ApiObject {
+  constructor (options) {
+    super(options);
+
+    this._docType = 'users';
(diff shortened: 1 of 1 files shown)


test/helpers/api-v3-integration.helper.js
```diff
@@ -116,10 +132,16 @@ export async function generateUser (update = {}) {
 // will will become the groups leader. Takes an update
 // argument which will update group
 export function generateGroup (leader, details = {}, update = {}) {
+  details.type = details.type || 'party';
+  details.privacy = details.privacy || 'private';
+  details.name = details.name || 'test group';
```
crookedneighbor, line 135: Are these sane defaults for groups? Or should it be a public guild?

paglias, line 135: The only thing that comes to my mind with it being a party is that you can only have one for user. But I think it's more than :+1: since creating a guild is really easy

TheHollidayInn, line 135: Private is default.

crookedneighbor:
Merged via command line.

Issue #6509, opened 2016-01-13
API v3 Short-Term Roadmap 1


paglias:
This is a list of small items regarding the upcoming API v3. If you plan to make a Pull Request just announce it with a comment. We'd like to have _atomic_ PRs so each one should only implement 1 route and its tests, if you plan to do more (awesome!) just propose separate PRs. All PRs must be done against the `api-v3-groups` branch.

**Members**
- [x] tests for getMember route
- [x] option to always populate leader? (on hold)
- [x] route to get members for challenges and tests (assigned to @paglias)

**Groups**
- [x] tests for remove group member route (assigned to @TheHollidayInn, in progress)
- [x] tests for get group route (assigned to @TheHollidayInn, in progress)
- [x] tests for get groups route (assigned to @TheHollidayInn, in progress)
- [x] tests for leave group route (assigned to @TheHollidayInn, in progress)
- [x] tests for join group route (assigned to @KristianTashkov , in progress)

**Challenges**
- [x] Challenge.getChallenge (assigned to @paglias)
- [x] Get challenges for group and tests (assigned to @paglias)
- [x] Update challenge route and tests (assigned to @paglias)
- [x] Challenge export to CSV route and tests (assigned to @paglias)
- [x] Get member progress route and tests (assigned to @paglias)
- [x] join route and tests (assigned to @KristianTashkov)
- [x] leave route and tests (assigned to @KristianTashkov)
- [x] user.unlink (assigned to @paglias)
- [x] tests for select winner route (assigned to @KristianTashkov)
- [x] tests for delete route (assigned to @KristianTashkov)
- [x] tests for create route (assigned to @crookedneighbor)
- [x] tests for get single challenge route
- [x] populate leader and group info

**Chat**
- [x] clearFlagCount route and tests (assigned to @TheHollidayInn)
- [x] seenMessage route and tests (assigned to @TheHollidayInn)
- [x] deleteChatMessage route and tests (assigned to @TheHollidayInn)

**Quest**
- [x] quests/invite route and tests (wip pr #6565)
- [x] quests/accept route and tests (wip pr #6565)
- [x] quests/reject route and tests (wip pr #6640)
- [x] quests/cancel route and tests (wip pr #6654)
- [x] quests/abort route and tests (wip pr #6662)
- [x] quests/leave route and tests (wip pr #6655)

**Tasks**
- [x] clearCompleted todos route and tests
- [x] getTasks route: add ability to get only completed todos (assigned to @paglias)
[...]

PR #6531, opened 2016-01-16
Added initial get groups tests

TheHollidayInn:
These are the initial get groups tests for #6509

### DIFF
--- a/test/api/v3/integration/groups/GET-groups.test.js
+++ b/test/api/v3/integration/groups/GET-groups.test.js
@@ -1,0 +1,99 @@
+import {
+  generateUser,
+  resetHabiticaDB,
+  generateGroup,
+} from '../../../../helpers/api-v3-integration.helper';
+
+describe('GET /groups', () => {
+  let user;
+  const NUMBER_OF_PUBLIC_GUILDS = 3;
+  const NUMBER_OF_USERS_PRIVATE_GUILDS = 1;
+  const NUMBER_OF_GROUPS_USER_CAN_VIEW = 5;
+
+  before(async () => {
+    await resetHabiticaDB();
+
+    let leader = await generateUser({ balance: 10 });
+    user = await generateUser({balance: 4});
+
+    let publicGuildUserIsMemberOf = await generateGroup(leader, {
+      name: 'public guild - is member',
+      type: 'guild',
+      privacy: 'public',
+    });
+    await leader.post(`/groups/${publicGuildUserIsMemberOf._id}/invite`, { uuids: [user._id]});
+    await user.post(`/groups/${publicGuildUserIsMemberOf._id}/join`);
+
+    await generateGroup(leader, {
+      name: 'public guild - is not member',
+      type: 'guild',
+      privacy: 'public',
+    });
+
+    let privateGuildUserIsMemberOf = await generateGroup(leader, {
+      name: 'private guild - is member',
+      type: 'guild',
+      privacy: 'private',
+    });
+    await leader.post(`/groups/${privateGuildUserIsMemberOf._id}/invite`, { uuids: [user._id]});
+    await user.post(`/groups/${privateGuildUserIsMemberOf._id}/join`);
+
+    await generateGroup(leader, {
+      name: 'private guild - is not member',
+      type: 'guild',
+      privacy: 'private',
+    });
+
+    await generateGroup(leader, {
+      name: 'party - is not member',
+      type: 'party',
+      privacy: 'private',
+    });
+
+    await user.post('/groups', {
+      name: 'party - is member',
+      type: 'party',
+      privacy: 'private',
+    });
+  });
+
+  it('returns error when no query passed in', async () => {
+    await expect(user.get('/groups'))
+      .to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: 'Invalid request parameters.',
+      });
+  });
+
+  it('returns only the tavern when tavern passed in as query', async () => {
+    await expect(user.get('/groups?type=tavern'))
+      .to.eventually.have.a.lengthOf(1)
+      .and.to.have.deep.property('[0]')
+      .and.to.have.property('_id', 'habitrpg');
+  });
+
+  it('returns only the user\'s party when party passed in as query', async () => {
+    await expect(user.get('/groups?type=party'))
+      .to.eventually.have.a.lengthOf(1)
+      .and.to.have.deep.property('[0]')
+      .and.to.have.property('leader', user._id);
+  });
+
+  it('returns all public guilds when publicGuilds passed in as query', async () => {
+    await expect(user.get('/groups?type=publicGuilds'))
+      .to.eventually.have.a.lengthOf(NUMBER_OF_PUBLIC_GUILDS);
+  });
+
+  it('returns all private guilds user is a part of when privateGuilds passed in as query', async () => {
+    await expect(user.get('/groups?type=privateGuilds'))
+      .to.eventually.have.a.lengthOf(NUMBER_OF_USERS_PRIVATE_GUILDS);
+  });
+
+  it('returns a list of groups user has access to', async () => {
+    let groups = await user.get('/groups?type=privateGuilds,publicGuilds,party');
+
+    await expect(groups.length)
+      .to.eql(NUMBER_OF_GROUPS_USER_CAN_VIEW);
+  });
+});
--- a/test/api/v3/integration/groups/GET-groups.test.js
+++ b/test/api/v3/integration/groups/GET-groups.test.js
@@ -76,8 +76,7 @@ describe('GET /groups', () => {
   it('returns only the user\'s party when party passed in as query', async () => {
     await expect(user.get('/groups?type=party'))
       .to.eventually.have.a.lengthOf(1)
-      .and.to.have.deep.property('[0]')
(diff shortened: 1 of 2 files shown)


test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -0,0 +1,97 @@
...
+  const NUMBER_OF_USERS_PRIVATE_GUILDS = 1;
+  const NUMBER_OF_GUILDS_USER_CAN_VIEW = 6;
+
+  before(async () => {
+    await resetHabiticaDB();
+
+    let leader = await generateUser({ balance: 10 });
+    user = await generateUser({balance: 4});
+
+    let publicGuildUserIsMemberOf = await generateGroup(leader, {
```
KristianTashkov, line 19: Is there a reason why you are not using `createAndPopulateGroup` for creating a group with members in it?

test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -0,0 +1,97 @@
...
+  generateUser,
+  resetHabiticaDB,
+  generateGroup,
+} from '../../../../helpers/api-v3-integration.helper';
+
+describe('GET /groups', () => {
+  let user;
+  const NUMBER_OF_PUBLIC_GUILDS = 3;
+  const NUMBER_OF_USERS_PRIVATE_GUILDS = 1;
+  const NUMBER_OF_GUILDS_USER_CAN_VIEW = 6;
```
KristianTashkov: `NUMBER_OF_GROUPS_USER_CAN_VIEW` since the party is in this number as well?

test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -0,0 +1,97 @@
...
+
+  before(async () => {
+    await resetHabiticaDB();
```
paglias, line 14: why this?

test/api/v3/integration/groups/GET-groups.test.js
paglias: @TheHollidayInn I'm not sure why user has access to 6 groups: party, tavern, 2 public guilds and a private guild. It shouldn't be 5?

test/api/v3/integration/groups/GET-groups.test.js
paglias, line 14: I mean, it shouldn't be necessary

crookedneighbor, line 14: It is necessary. Any other test that creates a public guild before this test runs will have that included in the public guild query. Resetting the db to an empty (plus tavern) state ensures that only the newly created public guilds will be counted.

test/api/v3/integration/groups/GET-groups.test.js
crookedneighbor, line 19: Probably because it's not part of the interface to add an existing user in the `createAndPopulateGroup` helper. Should probably add that functionality.

TheHollidayInn, line 19: Yep ^^

test/api/v3/integration/groups/GET-groups.test.js
TheHollidayInn, line 14: We can remove the dependency on counting. I can use `indexOf()` and `_.findIndex` to check rather than depend on the number of groups. Thoughts??

test/api/v3/integration/groups/GET-groups.test.js
TheHollidayInn: Yea.. that's odd that it passes. I'll check it out.

website/src/controllers/api-v3/groups.js
```diff
@@ -123,6 +123,10 @@ api.getGroups = {
       return previousValue.concat(Array.isArray(currentValue) ? currentValue : [currentValue]); // otherwise concat the new results to the previousValue
     }, []);
 
+    results = _.uniq(results, (group) => {
```
paglias: @TheHollidayInn why uniq? When it can result in the same challenge returned twice?

TheHollidayInn: Never mind about this. For some reason the await-to-eventually-have-lengthOf function was not counting the return object correctly. I am about to push a revision.

test/api/v3/integration/groups/GET-groups.test.js
TheHollidayInn, line 14: @paglias or @crookedneighbor . Let me know if you have any thoughts on changing this. I'm leaning towards using the index, but they both seem to work.

test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -0,0 +1,99 @@
...
+  it('returns all private guilds user is a part of when privateGuilds passed in as query', async () => {
+    await expect(user.get('/groups?type=privateGuilds'))
+      .to.eventually.have.a.lengthOf(NUMBER_OF_USERS_PRIVATE_GUILDS);
+  });
+
+  it('returns a list of groups user has access to', async () => {
+    let groups = await user.get('/groups?type=privateGuilds,publicGuilds,party');
+
+    await expect(groups.length)
+      .to.eql(NUMBER_OF_GROUPS_USER_CAN_VIEW);
```
TheHollidayInn: When I put the await inside of the expect and called `to.eventually.be.lengthOf` this was counting incorrectly.

crookedneighbor: Why are you `await`ing at all? `groups.length` does not return a promise.

TheHollidayInn: Eh, that is left over syntax. Now, I am in the habit of adding await before my expects..

paglias:
@TheHollidayInn it's possible that when you fetch public guilds the tavern is returned too? Which means that if you fetch both tavern and publicGuilds the tavern could be duplicated?

Also I'm not sure if we should populate the leader for the party and the tavern but not for other guilds, it's doesn't make sense

TheHollidayInn:
....dang. Yep, that is what I was doing. I went to test that out and forgot to put back the tavern string. So, then, that is why I was using _.uniq. Should we modify something else to ensure that tavern isn't duplicated or should I just leave the string as it is?

And, I'm unsure what the populate leader code does. I was going to test it on the get group by id tests. Do you want me to look into it?

paglias:
@TheHollidayInn for the tavern I think the best thing is to do something like this:
1. When only the tavern is fetched and not other public guilds -> fetch it as a single doc
2. When public guilds are fetched and the tavern too, just fetch public guilds since the tavern will be included

We could use https://docs.mongodb.org/manual/reference/operator/query/ne/ to exclude the tavern id but according to MongoDB's docs it's very inefficient.

The populateLeader option should be removed, basically it populates the leader field with the leader profile name but we don't care when listing all the guilds

TheHollidayInn:
Sounds good. I can make that change in the route. I was doing that with _.uniq, but I bet if I just check for the param and skip the tavern push when we see public guilds, that would be more efficient. I'll post the solution. 

Ah, I see what you are saying. Yea, I will remove the populateLeader.

crookedneighbor:
Ahh, yeah, the tavern + public guilds thing is exactly what is happening. (I made a note about that in the v2 tests)

crookedneighbor:
Also. Here's a crazy idea. (Not for this PR) What if we made the tavern it's own kind of group? IE: `type: 'tavern'`?

We're already going to need to change the id from 'habitrpg' to a uuid, right? Does it make more sense to call out that it is, in fact, a different kind of group?

paglias:
@crookedneighbor hmm for the uuid I made [this](https://github.com/HabitRPG/habitrpg/pull/6374/files#diff-bda86c52880f0950a5eb0e2f0538b936R537) change and we should be fine now. At this point changing the type of group would require to change a lot of code and in fact the tavern works almost as a normal public guild except that everyone is always a members

Alys:
Making the Tavern a new kind of group makes sense to me. Do we want to do
the same for Party Wanted and Newbies?

paglias:
@Alys @crookedneighbor I think we can have an attribute to identify official groups but not a completely different type of groups :( It would really complicate the code and there's no reason to do it as long as the functionalities are the same. And also having it compatible with api-v2 would complicate things even more. I think it can be reviewed post api v3 but not now

crookedneighbor:
@paglias Good points.

TheHollidayInn:
Updated with the two changes @paglias requested.

website/src/controllers/api-v3/groups.js
```diff
@@ -109,7 +109,9 @@ api.getGroups = {
           }).select(groupFields).sort(sort).exec()); // TODO use lean?
           break;
         case 'tavern':
-          queries.push(Group.getGroup({user, groupId: 'habitrpg', fields: groupFields, populateLeader: true}));
+          if (types.indexOf('publicGuilds') === -1) {
```
paglias, line 112: @TheHollidayInn could you remove the populateLeader at line 96? thanks

paglias, line 112: and at line 113 too

TheHollidayInn, line 112: Eh, sorry about that. I didn't notice it was a parameter. I was thinking it was just in the projection.

TheHollidayInn:
Updated.

PR #6541, opened 2016-01-17
Challenges join route and tests

KristianTashkov:
Some notes:
- Changed `createAndPopulateGroup` to properly create users who are members.
- I added `hasAccess` which is a method from a @paglias branch but it should merge easily since he told me where to put it :smile_cat: 
- I found a couple of routes that were crashing due to bad validation `checkQuery` instead of `checkParams` and `res.local` which was `undefined` for me. Am I missing something or are all these copy paste errors?
- Added a default value of 1 for Challenge `memberCount` since the creator always enters the challenge.

Most of the api logic with the join route is executed in `syncToUser`. I didn't want to make a thousand tests for every branch in this method, because this should be a unit test. So I created a couple of tests to check for the most important logic of `syncToUser` to make sure it is called and correctly executes in this route.

### DIFF
--- a/test/api/v3/integration/challenges/POST-challenges_challengeId_join.test.js
+++ b/test/api/v3/integration/challenges/POST-challenges_challengeId_join.test.js
@@ -1,0 +1,113 @@
+import {
+  generateUser,
+  createAndPopulateGroup,
+  translate as t,
+} from '../../../../helpers/api-v3-integration.helper';
+import { v4 as generateUUID } from 'uuid';
+
+describe('POST /challenges/:challengeId/join', () => {
+  it('returns error when challengeId is not a valid UUID', async () => {
+    let user = await generateUser({ balance: 1});
+
+    await expect(user.post('/challenges/test/join')).to.eventually.be.rejected.and.eql({
+      code: 400,
+      error: 'BadRequest',
+      message: t('invalidReqParams'),
+    });
+  });
+
+  it('returns error when challengeId is not for a valid challenge', async () => {
+    let user = await generateUser({ balance: 1});
+
+    await expect(user.post(`/challenges/${generateUUID()}/join`)).to.eventually.be.rejected.and.eql({
+      code: 404,
+      error: 'NotFound',
+      message: t('challengeNotFound'),
+    });
+  });
+
+  context('Joining a valid challenge', () => {
+    let groupLeader;
+    let group;
+    let challenge;
+    let authorizedUser;
+
+    beforeEach(async () => {
+      let populatedGroup = await createAndPopulateGroup({
+        members: 1,
+      });
+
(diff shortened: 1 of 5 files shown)


website/src/controllers/api-v3/challenges.js
```diff
@@ -135,12 +171,12 @@ api.getChallenge = {
   url: '/challenges/:challengeId',
   middlewares: [authWithHeaders(), cron],
   async handler (req, res) {
-    req.checkQuery('challengeId', res.t('challengeIdRequired')).notEmpty().isUUID();
+    req.checkParams('challengeId', res.t('challengeIdRequired')).notEmpty().isUUID();
```
crookedneighbor, line 184: Good catch on this.

test/api/v3/integration/challenges/POST-challenges_challengeId_join.test.js
```diff
@@ -0,0 +1,113 @@
...
+      let oldMemberCount = challenge.memberCount;
+
+      await authorizedUser.post(`/challenges/${challenge._id}/join`);
+
+      challenge = await authorizedUser.get(`/challenges/${challenge._id}`); // Switch to sync syntax when it's implemented for challenges
+
+      expect(challenge).to.have.property('memberCount', oldMemberCount + 1);
+    });
+
+    it('syncs challenge tasks to joining user', async () => {
```
paglias, line 89: here I'd say to check that `task.challenge` is correct, so `task.challenge.id === challenge._id` and maybe that `task.challenge.taskId` correctly matches the original task challenge. But if we decide that `syncToUser` will be unit tested, then we can skip these tests. @crookedneighbor ?

website/src/controllers/api-v3/challenges.js
```diff
@@ -86,6 +87,41 @@ api.createChallenge = {
 };
 
 /**
+ * @api {post} /challenges/:challengeId/join Joins a challenge
+ * @apiVersion 3.0.0
+ * @apiName JoinChallenge
+ * @apiGroup Challenge
+ *
```
paglias, line 102: here you should document the challengeId param `* @apiParam {UUID} challengeId The challenge _id`. I know it's missing in some other routes but I already added it where it's missing in my branch

test/api/v3/integration/challenges/POST-challenges_challengeId_join.test.js
KristianTashkov, line 89: yeah my thought process was that I can assert a lot of things, but they are part of `syncToUser` more than the integration. I'm testing the route and that it does the basic things.

crookedneighbor:
Merged in with f37b5a7fac74b3ddc0ca3bd22faaa65899ec1997

crookedneighbor:
Notated toward your next tier

PR #6543, opened 2016-01-17
Added intial group remove member tests

TheHollidayInn:
This is for the remove member section on #6509.

### DIFF
--- a/test/api/v3/integration/groups/POST-groups_id_removeMember.test.js
+++ b/test/api/v3/integration/groups/POST-groups_id_removeMember.test.js
@@ -1,0 +1,119 @@
+import {
+  generateUser,
+  createAndPopulateGroup,
+  translate as t,
+} from '../../../../helpers/api-v3-integration.helper';
+
+describe('POST /groups/:groupId/removeMember/:memberId', () => {
+  let leader;
+  let invitedUser;
+  let guild;
+  let member;
+  let member2;
+
+  before(async () => {
+    let { group, groupLeader, invitees, members } = await createAndPopulateGroup({
+      groupDetails: {
+        name: 'Test Guild',
+        type: 'guild',
+        privacy: 'private',
+      },
+      invites: 1,
+      members: 2,
+    });
+
+    guild = group;
+    leader = groupLeader;
+    invitedUser = invitees[0];
+    member = members[0];
+    member2 = members[1];
+  });
+
+  context('All Groups', () => {
+    it('returns an error when user is not member of the group', async () => {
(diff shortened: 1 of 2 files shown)


paglias:
@TheHollidayInn I just realized that this method doesn't do anything about challenges which would could be a problem. What about using group.leave here too with `keep-all`?

@Alys @crookedneighbor when a member is booted from a group they should be removed from challenges too? I'm worried that if they will not be removed from challenges they could have problems accessing them

Alys:
@paglias Members removed from groups should stay in the challenges. You can access challenges you've joined even when you aren't a guild member - the guild still appears in the list of guilds on the challenges page.

You could make a case for removing people from challenges if they've been booted from a guild for bad behaviour, but people get booted for other reasons too (e.g., accidental booting if someone with a similar name was supposed to be removed).

paglias:
@Alys that's true about public guilds challenges that can be joined by anyone but when you leave a guild you also leave all its challenges https://github.com/HabitRPG/habitrpg/blob/develop/website/src/models/group.js#L404. Being booted is the same action to leave except that is made by someone else and not by yourself

TheHollidayInn:
I guess this question mostly depends on if progress is lost during an accidental boot.

Alys:
True, and we should change that. It should be permissible to stay in
challenges when leaving a guild.

It's already possible to join a challenge without joining the guild (which
is good).

paglias:
@TheHollidayInn we've decided to delay the feature to keep challenges when leaving to after api-v3. So when booting a user we should call group.leave to keep things consistent

Alys:
No, please don't remove challenges when booting a user. Consistency there
isn't that important. It's fine to not add new code yet to support staying
in challenges when you leave a group voluntarily, but I think we also
should not add code to remove challenges on booting. Users can be booted
for invalid reasons and they shouldn't be penalised in their challenge
progress.

paglias:
In this case we'll have to change the code to allow for users not part of
the group to see them, I'll do it once it's merged
Il 18 gen 2016 8:45 PM, "Alys" notifications@github.com ha scritto:

Alys:
@paglias You can already see challenges that you are in when you aren't in the group. Is that what you're referring to?

paglias:
@Alys yeah in the current site yes but in v3 I made a change that could prevent accessing a challenge if you're not part of the group anymore, I'll fix it once this PR is merged

TheHollidayInn:
Okay, so then we don't need to make a change? :)

paglias:
@TheHollidayInn no changes :)

Issue #5854, opened 2015-08-28
Tutorials don't save progress mid-stream


SabreCat:
Tutorials are supposed to save the user's progress to `user.flags.tour[name-of-specific-tour]`, so that if you step away and come back, clicking Justin's head takes you to where you left off. However, this doesn't happen; if you bail out, you always resume the tour from the beginning. It appears that nothing in the step-by-step flow actually saves `user.flags.tour` with an appropriate value unless you reach the end.

I took a few stabs at addressing this myself, but `guideServices.js` is pretty bad spaghetti. There's 0-indexes mixed with 1-indexes, methods getting rebuilt several places, etc. etc. Might be this whole feature needs sanity checking and refactoring.

If we can figure this out, it'd make #5852 much easier to address.
##

PR #6545, opened 2016-01-18
Adding persistence for progress and reward tracking to welcome tour

### LINKED ISSUE #5852: Easy to (even accidentally) abuse intro tutorial bonuses
The [recent tweaks to onboarding to have the tutorial award Experience and Gold](https://github.com/HabitRPG/habitrpg/pull/5851) have some basic code in place to prevent tutorial steps from awarding their bonuses twice. However, the protections get thrown off if the user refreshes the page, so it's not too difficult to end up with the tour awarding more bennies than intended. We could improve this functionality with more persistent tracking of what awards have been given out.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26284466-easy-to-even-accidentally-abuse-intro-tutorial-bonuses?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tylernotfound:
Fixes #5852 and #5854.

I opted for localStorage to handle the persistence. The one caveat I see with that is that I never clear the key manually. The best spot for that, if it is a concern, would probably be on first login. That would ensure the key did not exist before a user sees the welcome dialog and tour tips. (I'm not sure where that would be added, or I would have done it.)

I added the `checkStats()` function to index.js because I didn't want to add unintentional bloat in the `update()` function, which is used with `User.set()`. If there's a smarter way to handle this level checking already, let me know. :)

I couldn't find a way to break this on my end, but it'd be worth having some other people test I think.

### DIFF
--- a/common/script/index.js
+++ b/common/script/index.js
@@ -2056,6 +2056,10 @@ api.wrap = function(user, main) {
           cb(null, user);
         }
         return delta;
+      },
+      checkStats: function (req, cb) {
+        user.fns.updateStats(user.stats, req);
+        return typeof cb === "function" ? cb(null, user) : void 0;
       }
     };
   }
--- a/website/public/js/services/guideServices.js
+++ b/website/public/js/services/guideServices.js
@@ -196,22 +196,19 @@ function($rootScope, User, $timeout, $state, Analytics) {
       };
       step.onHide = function(){
         var ups = {};
-        // Persistantly track tour progress and rewards with localStorage
-        if (step.experience) {
-          var tourStep = localStorage.getItem('habitrpg-tourStep') || -1;
-          if (i > tourStep) {
-            localStorage.setItem('habitrpg-tourStep', i);
-            ups['stats.gp'] = User.user.stats.gp + (step.gold || 0);
-            ups['stats.exp'] = User.user.stats.exp + (step.experience || 0);
-          }
+        var lastKnownStep = User.user.flags.tour[k];
+
+        if (i > lastKnownStep) {
+          if (step.gold) ups['stats.gp'] = User.user.stats.gp + step.gold;
+          if (step.experience) ups['stats.exp'] = User.user.stats.exp + step.experience;
+          ups['flags.tour.'+k] = i;
         }
         if (step.final) { // -2 indicates complete
           ups['flags.tour.'+k] = -2;
           Analytics.track({'hitType':'event','eventCategory':'behavior','eventAction':'tutorial','eventLabel':k+'-web','eventValue':i+1,'complete':true})
         }
         User.set(ups);
-        // User.set() doesn't include a check for level changes,
-        // so manually check here as we expect the user to ding lvl2.
+        // User.set() doesn't include a check for level changes, so manually check here.
         if (step.experience) {
           User.user.ops.checkStats({});
         }
@@ -252,9 +249,6 @@ function($rootScope, User, $timeout, $state, Analytics) {
     if (page === -1) page = 0;
     var curr = User.user.flags.tour[chapter];
     if (page != curr+1 && !force) return;
-    var updates = {};
-    updates['flags.tour.'+chapter] = page;
-    User.set(updates);
     var chap = tour[chapter], opts = chap._options;
     opts.steps = [];
     _.times(page, function(p){
--- a/website/public/js/services/guideServices.js
+++ b/website/public/js/services/guideServices.js
@@ -198,6 +198,11 @@ function($rootScope, User, $timeout, $state, Analytics) {
         var ups = {};
         var lastKnownStep = User.user.flags.tour[k];
 
+        // Return early if user has already completed this tutorial
+        if (lastKnownStep === -2) {
+          return;
+        }
+
         if (i > lastKnownStep) {
           if (step.gold) ups['stats.gp'] = User.user.stats.gp + step.gold;
           if (step.experience) ups['stats.exp'] = User.user.stats.exp + step.experience;
--- a/website/public/js/services/guideServices.js
+++ b/website/public/js/services/guideServices.js
@@ -195,19 +195,26 @@ function($rootScope, User, $timeout, $state, Analytics) {
         }
       };
       step.onHide = function(){
-        var ups={};
-        if (!$rootScope.stepAwarded) $rootScope.stepAwarded = {};
-        if (!$rootScope.stepAwarded[i]) {
-          $rootScope.stepAwarded[i] = true;
-          ups['stats.gp'] = User.user.stats.gp + (step.gold || 0);
-          ups['stats.exp'] = User.user.stats.exp + (step.experience || 0);
+        var ups = {};
+        // Persistantly track tour progress and rewards with localStorage
+        if (step.experience) {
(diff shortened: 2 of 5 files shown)


Alys:
We unfortunately have a LOT of trouble with local storage (for example, https://github.com/HabitRPG/habitrpg/issues/4313 ). I'm concerned that it might cause problems for the tutorial too, which means new users would have a bad first experience of the website. 

Would you consider other forms of persistence? For example, we could store progress as a database flag (which has the advantage that it persists across browsers, which could be useful for users who initially skip the tutorial).

tylernotfound:
Hi @Alys!

I see. So, if I read that issue correctly, and localStorage is cleared out each night, this approach definitely wouldn't be effective. I'm still very new to this codebase, so I wasn't about to start setting database properties if the solution could be handled on the user's end (especially since it's transient).

It also looks like one of the tests isn't passing, so I'll need to sort that out too. I'll take another look at handling the localStorage as a database flag instead.

If there are any other issues in here, apart from those two, let me know.

tylernotfound:
I've updated the code to remove localStorage from the process, using a database flag instead per @Alys's suggestion. The tests were passing on my end, so hopefully that is addressed as well.

There's one specific change in this commit that's worth calling out. Around line 255 of `guideServices.js`, the updating process in the `goto()` function has been removed, in favor of handling all flag updating for the step in the `onHide()` function. IMO, this makes sense as we only update flags for a given step when the user has interacted with it.

Alys:
Just FYI: It's not that local storage is cleared out every night (it's not). The problem is that local storage for the habitica website can't be trusted to contain correct information.

tylernotfound:
Ah, word. My bad. I misinterpreted you suggesting someone clear their localStorage as something else, it seems. Haha :grin:

SabreCat:
Are there tests we can implement to get coverage here? The new function in particular would be great to have a case for.

tylernotfound:
@SabreCat I'm happy to try and add a test file for the new method. I'll make another commit with some working tests. It will be pretty simple, though, since the method is really just a helper to `updateStats()`.

tylernotfound:
@SabreCat I pushed another commit with the test file for `user.ops.checkStats.js`. Let me know if you think that covers it.

SabreCat:
@tylerxo: Excellent, thank you!

SabreCat:
This looks like quite an improvement! Paging back and forth through tutorial, refreshing the browser window, or closing the browser window and returning all work as expected, only ever awarding Gold and Experience once per tutorial step.

@vIiRuS -- do we need to check anything with the mobile app tutorial for these changes?
@crookedneighbor -- any suggestions for the tests here?

website/views/shared/header/menu.jade
```diff
@@ -272,7 +272,7 @@ nav.toolbar(ng-controller='MenuCtrl')
 
     ul.toolbar-bailey(ng-class='{inactive: !_expandedMenu.menu}')
       li.toolbar-bailey-container(ng-if='user.flags.tour.intro!=-2')
-        .npc_justin_head.npc_bailey_head(popover='Continue Tour', popover-trigger='mouseenter', popover-placement='bottom', ng-click='Guide.goto("intro", user.flags.tour.intro, true)')
+        .npc_justin_head.npc_bailey_head(popover='Continue Tour', popover-trigger='mouseenter', popover-placement='bottom', ng-click='Guide.goto("intro", user.flags.tour.intro + 1, true);')
```
crookedneighbor: This does not need the added semicolon.

crookedneighbor:
@SabreCat The mobile app tutorials are a different object on the user.

common/script/index.js
```diff
@@ -2056,6 +2056,10 @@ api.wrap = function(user, main) {
           cb(null, user);
         }
         return delta;
+      },
+      checkStats: function (req, cb) {
```
crookedneighbor, line 2060: Why do we need this wrapper?

tylernotfound, line 2060: > "...If there's a smarter way to handle this level checking already, let me know. :)"

I mentioned that this might not be the best approach in the original PR comment; totally open to what you think would work better here. :+1: 

With that said, my reasoning for creating this wrapper is that calling `user.fns.updateStats()` directly never produced the right result, and was causing other issues. I couldn't find another function designed to ensure the current state of a user was correct (such as being lvl1 with 151 exp). This is necessary because `User.set()` does not include a path for checking a user's level. So when `User.set()` is used to set exp, no leveling up occurs, and that's the issue with the initial tour.

website/public/js/services/guideServices.js
```diff
@@ -195,19 +195,23 @@ function($rootScope, User, $timeout, $state, Analytics) {
...
         }
         if (step.final) { // -2 indicates complete
           ups['flags.tour.'+k] = -2;
-          $rootScope.stepAwarded = null;
           Analytics.track({'hitType':'event','eventCategory':'behavior','eventAction':'tutorial','eventLabel':k+'-web','eventValue':i+1,'complete':true})
         }
         User.set(ups);
+        // User.set() doesn't include a check for level changes, so manually check here.
+        if (step.experience) {
+          User.user.ops.checkStats({});
```
crookedneighbor, line 218: I just feel like we could just do `User.user.fns.updateStats(User.user.stats);` and it would eliminate the need for a wrapper.

crookedneighbor:
I've been testing this out, but if I progress a few steps forward, refresh the page, and then click a few steps backward, I do get extra experience and gold.

website/public/js/services/guideServices.js
tylernotfound, line 218: I tried this route first, but was running into problems that I didn't know how to solve. Here's what happens when using the `fns` over the `ops` function:
1. Create new account
2. Step through all tutorial steps
3. Hit level 2 upon completion
4. Refresh the page
5. User is lvl1 with 151 exp

This is never the case when I use the wrapper in `ops`. Perhaps there needs to be an additional save call after using `fns.updateStats()`?

tylernotfound:
@crookedneighbor I'm not sure if you're saying this update isn't working at all for you, or just in a particular case, but I did find an edge case on my end.

If I advanced the final step, then refreshed, and start clicking to previous steps, I was being awarded extra exp. I just pushed a commit which catches this condition by simply returning early if the tour flag is -2 (complete). This also avoids extra Analytics and `User.set()` calls.

SabreCat:
Whoops, I misread the most recent comments. Apologies for the update spam.

tylernotfound:
I'm not really sure what the status is with this one. @crookedneighbor was having problems with it, but that may have been resolved with the edge case I found and fixed. There's also a question around whether we need the additional wrapper function, which I tried to outline exactly what was happening that lead me to create it.

SabreCat:
Since the upgrade to API v3, we've been removing the use of local storage wherever possible. Is this solution still viable? @Alys @crookedneighbor @paglias

crookedneighbor:
Sorry this took _soooooooo_ long, but your changes have been merged in. 

What's your user id?

PR #6549, opened 2016-01-18
Fix setting of environment variables in gulp test tasks in Windows

KristianTashkov:
This will make all test suits runnable on Windows.
Note: We should really add in the wiki that you need Firefox 40.0.3 specifically otherwise e2e tests can't connect to the selenium server.


Alys:
@KristianTashkov If you'd like to update the wiki with anything you think it needs, please do! We'd love the help there.

KristianTashkov:
@Alys I would really prefer someone else write it, I'm not that good with words and the wiki should be as clear as possible.

Alys:
No worries! I'll see that it gets done. Thank you for pointing it out!

SabreCat:
Gave this a try in my Windows environment:

`gulp test:common` works great!
`gulp test:karma` fails five tests against the Inventory Controller that pass in my Linux env.
`gulp test:e2e` hangs, even with Firefox 40.0.3 installed. Does Firefox need to be the default browser?
`gulp lint` throws thousands of errors `error Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style` that don't occur in my Linux env.

KristianTashkov:
@SabreCat the `LF` vs `CLRF` is because you have setup git to checkout files as `CLRF` and then commit them `LF`.
for `e2e` tests if it hangs, it means the selenium server isn't starting. Do you have Java 7? Try running `npm run test:e2e:webdriver` and see the error it probably produces.
For the karma tests I have no idea, I'm running them without problems.

SabreCat:
I'm running Java 8, by the look of it.

`seleniumProcess.pid: 10932
Error: Invalid or corrupt jarfile F:\Git\habitica\node_modules\protractor\selenium\selenium-server-standalone-2.47.1.jar
Selenium Standalone has exited with code 1`

SabreCat:
Ugh. Tried changing Git's LF/CLRF setting, cloned the repo anew, and it all took a step backward. Now even `gulp test:common` errors out:

> [15:56:14] 'test:prepare:build' errored after 4.94 s
> [15:56:14] Error: Command failed: C:\WINDOWS\system32\cmd.exe /s /c "set NODE_ENV=testing&&grunt build:test"
> js-bson: Failed to load c++ bson extension, using pure JS version
>    at ChildProcess.exithandler (child_process.js:751:12)
>    at ChildProcess.emit (events.js:110:17)
>    at maybeClose (child_process.js:1015:16)
>    at Socket.<anonymous> (child_process.js:1183:11)
>    at Socket.emit (events.js:107:17)
>    at Pipe.close (net.js:485:12)

Friggin' Windows `>.<`

KristianTashkov:
hmm do you have a c++ compiler? VS 2013 EXPRESS or above?

KristianTashkov:
@SabreCat I figured your error out. When you were installing moongose installation failed because you were missing some build essentials (look for the dependencies in [node-gyp readme](https://github.com/nodejs/node-gyp)). When you set them up (I had to run `npm install --msvs_version=2013` to make `node-gyp` use vs 2013 which I installed), you should reinstall the mongoose module.

KristianTashkov:
For more information: [here](http://stackoverflow.com/questions/21656420/failed-to-load-c-bson-extension) and [here](http://stackoverflow.com/questions/28651028/cannot-find-module-build-release-bson-code-module-not-found-js-bson)

crookedneighbor:
Merged in with a different commit

PR #6333, opened 2015-12-04
Proposed fix for issue #6332

damemi:
Fixes cosmetic bug where "add task" button jumps from being centered to not. Removed one line of CSS specifying "height:100%"

Issue #6332, opened 2015-12-04
Add task "+" positioning acts weird


damemi:
This is a small cosmetic bug, but when adding a task (at least for me on Linux Chrome 46.0.2490.86 64-bit) the plus sign seems to be positioned incorrectly (too high), because as soon as I start typing in the box it shifts down to a centered position. Screenshots below:

![plus-above](https://cloud.githubusercontent.com/assets/1839101/11597312/06566a2e-9a89-11e5-92a7-fed2b18784c4.png)
![plus-moved](https://cloud.githubusercontent.com/assets/1839101/11597315/0a9ce72a-9a89-11e5-9991-50f3068fa967.png)

PR #5327, opened 2015-06-01
Correct add task icon alignment

excentris:
Fixes #5148

PR #5304, opened 2015-05-29
fixes #5148 ("Add Task plus not vertically centered")

averrin:
just add
.task-add.button.glyphicon-plus {
      vertical-align: text-bottom;
}
maybe its a bad solution, but works

PR #6554, opened 2016-01-19
#6332, #5148: Fix 'add task' button jumping around

keysmashes:
Stops the plus icon jumping down when text is entered in the 'new task' textbox.

Related: #6333, #6332, #5327, #5304 and #5148.

[Comment 1 (crookedneighbor)](https://github.com/HabitRPG/habitrpg/pull/5327#issuecomment-111784268):

> That div was created to generate a tooltip when a user tries to press the plus button when no task information has been supplied. It's also what is causing the plus button to get pushed around.
> 
> The tooltip should appear when hovering over the button, and only if the input is empty. It seems to not be behaving correctly :frowning:. It was flaky before, but without that styling, it's extra flaky.

[Comment 2 (crookedneighbor)](https://github.com/HabitRPG/habitrpg/pull/6333#issuecomment-164201578):

> This has been suggested before, but that height style is required for the tooltip to show up. I'm going to add a comment to that effect.

I've tested on Chrome 47 and Firefox 43 (both on Ubuntu), and the behaviour with this change is exactly the same as on the live site, except the icon not jumping around. On Chrome, the tooltip appears and behaves the same on both live site and local server. On Firefox, I can't get the tooltip to show up at all on either my local server or the live site, but that seems like a separate issue.

Before:

image

After:

image

### DIFF
--- a/website/public/css/tasks.styl
+++ b/website/public/css/tasks.styl
@@ -193,11 +193,6 @@ for $stage in $stages
     border: 0
     float: right
     font-size: 0.85em
-  .empty-task-notification
-    // @TODO: This causes weird centering behavior (see #6332)
-    // however, this is required to show the tooltip on the task button
-    // Figure out how to fix the centering behavior while preserving the tooltip
-    height: 100%;
 
 // message in Dailies column when Resting in Inn
 // ------------------------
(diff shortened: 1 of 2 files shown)


SabreCat:
Merged and noted toward your next contributor tier! Thanks, @anowlcalledjosh!

PR #6557, opened 2016-01-20
Add avatar customizations to content API

phillipthelen:
This adds all avatar customizations to the content API. This makes them available for the mobile apps.
A future change should also make the website use this API data, instead of the current solution.

Are there any tests that should be added to this?


crookedneighbor:
Got yourself some linting errors :)

website/src/controllers/api-v2/auth.js
```diff
@@ -6,7 +6,6 @@ var async = require('async');
 var utils = require('../../libs/utils');
 var nconf = require('nconf');
 var request = require('request');
-var FirebaseTokenGenerator = require('firebase-token-generator');
```
crookedneighbor, line 9: Why was this stuff removed?

common/script/content/appearance/keys.js
```diff
@@ -0,0 +1,8 @@
+import {forOwn} from 'lodash';
+
+export default function(obj) {
```
paglias: @vIiRuS I think the forOwn here is not cloning incoming objects so it's probably attaching `key` everywhere, you should first clone (`_.cloneDeep`) the object and then attach the `key` property

crookedneighbor: I belie e that is the intention. To transform the object that is passed in.

paglias: Hmm.. then ok!

phillipthelen: Yes, that was the idea. :)

phillipthelen:
Got everything green :tada:

website/src/controllers/api-v2/user.js
```diff
@@ -319,9 +319,11 @@ let requiresPurchase = {
 
 let checkPreferencePurchase = (user, path, item) => {
   let itemPath = `${path}.${item}`;
-  let isDefaultPreference = _.get(shared.content.defaultAppearancePreferences, itemPath);
+  let appearance = _.get(shared.content.appearances, itemPath);
 
-  if (isDefaultPreference) return true;
+  if (appearance) {
```
crookedneighbor: I think part of the defaults should be to set the price to 0 if it does not exist and this should now be:

``` js
if (!appearance) return false;
if (appearance.price === 0) return true;
```

IE, if the appearance is not valid, return false early. If the price is 0, return true early.

phillipthelen:
Had to also add a check if it's a background. And added the price field to backgrounds too.

website/src/controllers/api-v2/user.js
```diff
@@ -319,9 +319,11 @@ let requiresPurchase = {
 
 let checkPreferencePurchase = (user, path, item) => {
   let itemPath = `${path}.${item}`;
-  let isDefaultPreference = _.get(shared.content.defaultAppearancePreferences, itemPath);
-
-  if (isDefaultPreference) return true;
+  if (path !== 'background') {
```
crookedneighbor: I don't like this. :(

phillipthelen: I'm open to better ideas. Couldn't think of a nicer solution for this.

phillipthelen:
Should I try to find a different solution for the check if the item is a background, or should be just merge this for now?

crookedneighbor:
Merged in with 8e8c28725a057a2620e490b978d0402a0458b381

PR #6558, opened 2016-01-20
Paypal sub button creates an alert when there are errors fix #6401

### LINKED ISSUE #6401: immediate cancellation of new PayPal subscription shows the user a raw error message
If you create a subscription with PayPal and then try to cancel it immediately, the notification that you can't do that yet appears as a raw error message in a page by itself, instead of in a dialog screen with an "ok" button to dismiss it.

```
{"err":"\"The plan hasn't activated yet (due to a PayPal bug). It will begin 2015-12-26T10:00:00Z, after which you can cancel to retain your full benefits\""}
```

screen shot 2015-12-26 at 12 20 44 pm

ojeytonwilliams:
Should fix #6401

The main body of the fix just handles any error messages coming back from Paypal.  I opted to pull out error_description, it looks to be the most informative message, and pop that into an alert.

In addition I changed a bit of paypalBillingSetup since that didn't appear to be working properly or give you the keys you'd want for testing.

Hope this is okay!

### DIFF
--- a/test/spec/services/paymentServicesSpec.js
+++ b/test/spec/services/paymentServicesSpec.js
@@ -1,0 +1,96 @@
+'use strict';
+
+import i18n from '../../common/script/i18n';
+
+describe('paymentServices', function(){
+  let $httpBackend, $window, user, Payments, getAnySubPattern
+
+  getAnySubPattern= /^\/paypal\/subscribe.*/
+ 
+  beforeEach(module('habitrpg'));
+ 
+  beforeEach(function(){
+    module(function($provide){
+      $window = {href: '', alert: sandbox.spy(), location: {search:   '', pathname: ''}};
+      $provide.value('$window', $window);
+    });
+    
+    inject(function($injector) {
+    // Set up the mock http service responses
+    $httpBackend = $injector.get('$httpBackend');
+    Payments = $injector.get('Payments');
+    
+   // console.log("PAYMENT SERVICES")
+   // console.log($httpBackend);
+  //  console.log(Payments);
+   // /\/user\/(.+)/
+
+ //   $httpBackend.when('GET', '/\/paypal\/subscribe.*/')    .respond({userId: 'userX'}, {'A-Token': 'xxx'});
+/*    $httpBackend.whenRoute('GET', '/paypal/subscribe') .respond(function(method, url, data, headers, params) {
+    return [200, "Whatever"];
+  });
+    
+    .respond(function(method, url, data, headers, params) {
+    return [200, MockUserList[Number(params.id)]];
+  }); */
+    /*.respond(function(){
+      return {userId: 'userX'}, {'A-Token': 'xxx'}
+    }); */
+   })});
+/*
+  beforeEach(function(){
+    module(function($provide){
+      $window = {href: '', alert: sandbox.spy(), location: {search: '', pathname: ''}};
+      $provide.value('$window', $window);
+    });
+    
+    
+
+    inject(function(_$httpBackend_, User, _Payments_) {
+      Payments = _Payments_;
+      $httpBackend = _$httpBackend_;
+    });
+ //   localStorage.removeItem(STORAGE_SETTINGS_ID);
+ //   localStorage.removeItem(STORAGE_USER_ID);
+  }); */
+  
+  afterEach(function() {
+    $httpBackend.verifyNoOutstandingExpectation();
+    $httpBackend.verifyNoOutstandingRequest();
+  }); 
+
+  
+  it('navigates to the correct url', function(){
+      let data = {_id: 7, apiToken: "anApiToken", sub: "subKey", coupon: "coupon"};
+      let subUrl = '/paypal/subscribe?_id=7&apiToken=anApiToken&sub=subKey&coupon=coupon';
+    
+      $httpBackend.when('GET', subUrl).respond({place : 'holder'});
+      $httpBackend.expectGET(subUrl);
+      Payments.showPaypal(data);
+      $httpBackend.flush();
+  })
+    
+    it('changes the browser location if paypal responds as expected', function(){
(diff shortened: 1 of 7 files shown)


SabreCat:
Taking a look at this, the functions added and modified could use test coverage. Let us know if you need a hand navigating the test framework!

ojeytonwilliams:
@SabreCat I might need a little help, actually!

Am I right in thinking that I should be writing two tests - one karma test for paymentServices.js and a mocha test for paypal.js?  The impression I got from looking through the tests was that the karma test should be similar to the tests in test/spec/services and the mocha test should be something like the tests in test/server_side/controllers

If that's broadly the right idea, I'll see what I can do from here.

SabreCat:
You've got it! Karma (test/spec/*) is for our Angular tests, and Mocha (test/server_side/* in this case) for server operations.

website/src/controllers/payments/paypal.js
```diff
@@ -26,9 +26,10 @@ paypal.configure({
 });
 
 var parseErr = function(res, err){
-  //var error = err.response ? err.response.message || err.response.details[0].issue : err;
-  var error = JSON.stringify(err);
-  return res.json(400,{err:error});
+  var error = err.response ? err.response.error_description || err.response.error : 'Paypal returned an invalid error message.';
+  return res.json(400, {err: 'Paypal error: ' + error});
```
Alys: Instead of "Paypal returned an invalid error message" I think it would be better to say just "PayPal reported an error". It's a simpler message and therefore might be less stressful to any users who see it. Also that and the "PayPal error" string should be translatable. Ask if you're not sure how to achieve that!

Please also add this string after the error message:
"Please email admin@habitica.com for assistance and paste that error message into your email."

ojeytonwilliams: Yeah, that was a bit awkward.  The idea was that that message was there in case Paypal responded in a weird way (didn't give an error or a normal response), but I agree that it looks a bit scary.

I'll implement those changes as soon as I've wrapped my brain around the test framework!

ojeytonwilliams:
One thing I've discovered is that the tests in sub directories of test/server_side are not being run, and when you run them a fair number of them fail.

Does anyone know if the tests are obsolete and should be deleted or are in dire need of fixing?

crookedneighbor:
Sorry, forgot to mention that the tests in the server_side folder should not be modelled after. They were my (unsuccessful) attempt at starting unit testing for the controllers and libs.

The server code is primarily tested by integration tests, but this kind of thing will be hard to test with that.

Just do client side testing for now and I'll write something for the server before merging it in.

website/src/controllers/payments/paypalBillingSetup.js
```diff
@@ -1,3 +1,5 @@
+require('babel-core/register');
```
crookedneighbor: Why did you add this here?

ojeytonwilliams: That I added because I needed to run that on the command line to generate the Paypal keys and node didn't like the ES6 code.

I'm not 100% sure that's the best way to go about it, but it seemed to work.

paglias: You should not use babel-core/register. When exactly this code will have to run from the command line? If only locally you can use https://babeljs.io/docs/usage/cli/

ojeytonwilliams: Could you explain why babel-core/register shouldn't be used?  As I understand it, it's discouraged for production code because of how slow it is, but this particular script should never appear in production (if it does, that seems problematic).

Anyways, if you don't want that line, I'll take it out and replace it with some notes that might help out others that run into the same issues I did.

ojeytonwilliams:
Oh, I deleted the wrong file.  Please ignore this while I sort that out!

ojeytonwilliams:
Well, that was surprisingly frustrating.  I ended up having to force push this commit, since it refused to accept any changes.

Sorry about that, I'll try to keep it a bit cleaner in future!

test/spec/services/paymentServicesSpec.js
```diff
@@ -0,0 +1,96 @@
...
+      $window = {href: '', alert: sandbox.spy(), location: {search:   '', pathname: ''}};
+      $provide.value('$window', $window);
+    });
+    
+    inject(function($injector) {
+    // Set up the mock http service responses
+    $httpBackend = $injector.get('$httpBackend');
+    Payments = $injector.get('Payments');
+    
+   // console.log("PAYMENT SERVICES")
```
crookedneighbor, line 23: What's all this for?

ojeytonwilliams, line 23: Let's see.  

The regex is just to catch any PayPal sub requests.

beforeEach(module('habitrpg')) I copied from userServicesSpec and may not be necessary.  I didn't think to check, sorry!

The next part stubs out $window, so we can see if Angular triggers an alert when things go wrong or redirects the user when they go right.

Finally, the injector gives us access to a mock $http service and the Payments service we want to test.

crookedneighbor, line 23: I meant, "why do you have all this commented out code"?

crookedneighbor:
Sorry for the delay. Now that v3 of the API is out, we can finally devote some time to reviewing these PRs. Are you able to sort out the merge conflicts?

crookedneighbor:
Going to close this, but please re-open it with the merge conflicts sorted out so we can review it.

crookedneighbor:
Adapting this for v3. You'll receive a contributor credit if it gets merged in.

ojeytonwilliams:
Hi.  I'll try to have a look at this as soon as I can, but I'm currently travelling.

If I can convince my windows machine to install Habitica I might be able to do something tomorrow, but sometime next week is more realistic.

PR #6563, opened 2016-01-21
Added initial get group tests

TheHollidayInn:
Most of the get group tests from V2 transferred. I removed some member tests, cleaned up some others, and added the new generating chat functions. 

This is part of #6509.

### DIFF
--- a/test/helpers/api-integration/api-classes.js
+++ b/test/helpers/api-integration/api-classes.js
@@ -59,6 +59,31 @@ export class ApiGroup extends ApiObject {
 
     this._docType = 'groups';
   }
+
+  async addChat (chat) {
+    let group = this;
+
+    if (!chat) {
+      chat = {
+        id: 'Test_ID',
+        text: 'Test message',
+        flagCount: 0,
+        timestamp: Date(),
+        likes: {},
+        flags: {},
+        uuid: group.leader,
+        contributor: {},
+        backer: {},
+        user: group.leader,
+      };
+    }
+
+    let update = { chat };
+
+    this.update(update, '$push');
+
+    return this;
+  }
 }
 
 export class ApiChallenge extends ApiObject {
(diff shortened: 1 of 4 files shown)


TheHollidayInn:
Okay, I just pushed a fix for a failing chat tests, because in this PR I removed the ability to view flags for a non-admin. I am going to work on converting the chat tests to await/async and adding the remaining tests after this, so I didn't make too many adjustments to the chat in this PR.

website/src/controllers/api-v3/groups.js
```diff
@@ -152,6 +152,13 @@ api.getGroup = {
     let group = await Group.getGroup({user, groupId: req.params.groupId, populateLeader: true});
     if (!group) throw new NotFound(res.t('groupNotFound'));
 
+    if (!user.contributor.admin) {
+      _.remove(group.chat, function removeChat (chat) {
```
paglias, line 157: @TheHollidayInn here you probably want to convert the group to a plain object with group.toJSON before removing flags information

test/helpers/api-integration/api-classes.js
```diff
@@ -59,6 +59,31 @@ export class ApiGroup extends ApiObject {
...
+        backer: {},
+        user: group.leader,
+      };
+    }
+
+    let update = { chat };
+
+    this.update(update, '$push');
+
+    return this;
```
paglias: this is wrong I think, this.update is an async function so you have to call `await this.update(...)` and since `this.update` already returns `this` you may just do `return await this.update(...)`

paglias: also update only takes one parameter https://github.com/TheHollidayInn/habitrpg/blob/api-v3-get-group-tests/test/helpers/api-integration/api-classes.js#L20, not an object

website/src/controllers/api-v3/groups.js
TheHollidayInn, line 157: What will that help with?

test/helpers/api-integration/api-classes.js
TheHollidayInn: Awesome, this was my problem. I was also working on implementing the push, but decided it wasn't needed.

TheHollidayInn:
Updated!

PR #6564, opened 2016-01-21
Return enable symlinks in vagrant and use npm@3

KristianTashkov:
We need symlinks enabled so we can have the symlinks in `./node_modules/.bin/` for binaries used during testing.

When trying to `npm install` with npm@2 I get a lot of errors and a failed installation. This doesn't occur with npm@3.

Also it's interesting to note that while these changes work for me, there seems to be a problem with npm that requires `npm install` to be called twice. I found other people with the same issue and the npm guys confirming it's a bug. [source](https://github.com/npm/npm/issues/10727)

### DIFF
--- a/vagrant_scripts/install_node.sh
+++ b/vagrant_scripts/install_node.sh
@@ -16,7 +16,7 @@ nvm use
 nvm alias default current
 
 echo Update npm...
-npm install -g npm@2
+npm install -g npm@3
 
 echo Installing global modules...
 npm install -g gulp bower grunt-cli mocha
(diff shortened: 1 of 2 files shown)


vagrant_scripts/install_node.sh
```diff
@@ -16,7 +16,7 @@ nvm use
 nvm alias default current
 
 echo Update npm...
-npm install -g npm@2
+npm install -g npm@3
```
crookedneighbor, line 19: Hmmm... I would have thought that using npm 3 would cause a lot of errors because of unmet peer dependencies. Is that not the case?

KristianTashkov, line 19: I have raised the box like 3 or 4 times and have no problems with it.

crookedneighbor:
Notated.

PR #6565, opened 2016-01-21
WIP: Quests for API v3

SabreCat:
Quest invite and accept routes, WIP. Corresponds to my assignments in #6509.


test/api/v3/integration/groups/POST-groups_groupId_quests_accept.test.js
```diff
@@ -0,0 +1,54 @@
+import {
+  createAndPopulateGroup,
+  translate as t,
+} from '../../../../helpers/api-v3-integration.helper';
+
+describe.skip('POST /groups/:groupId/quests/accept', () => {
```
SabreCat: Ignore this file for now, it's just got some placeholder garbage until `invite` is in good shape.

test/api/v3/integration/groups/POST-groups_groupId_quests_invite.test.js
```diff
@@ -0,0 +1,119 @@
...
+      });
+    });
+  });
+
+  context('successfully issuing a quest invitation', () => {
+    it('sends an invite to all party members', async () => {
+      leader.items.quests[PET_QUEST] = 1;
+
+      await expect(leader.post(`groups/${questingGroup._id}/quests/invite/${PET_QUEST}`)).to.eventually.deep.equal([{
+
```
SabreCat: I don't know what the return for this route looks like or should look like.

TheHollidayInn: I think we just returned the invites in the group invites, so that we can just update the object on the front end easily.

test/api/v3/integration/groups/POST-groups_groupId_quests_invite.test.js
```diff
@@ -0,0 +1,119 @@
+import {
+  createAndPopulateGroup,
+  translate as t,
+} from '../../../../helpers/api-v3-integration.helper';
+import { v4 as generateUUID } from 'uuid';
+
+describe.skip('POST /groups/:groupId/quests/invite', () => {
```
TheHollidayInn: Is this supposed to have the .skip function call?

SabreCat: Only for now. When I want to start running the test, I'll take it out.

TheHollidayInn: Gotcha. I thought we were just ignoring the top file.

test/api/v3/integration/groups/POST-groups_groupId_quests_invite.test.js
crookedneighbor: Agreed.

SabreCat: I don't think we have an invite object, as such, here. It rolls through adding a few properties to each party member and the party itself. Unless I'm not understanding it correctly, which is quite possible.

paglias: I'd return group.quest

crookedneighbor: I think you're right, group.quest makes the most sense.

website/src/controllers/api-v3/groups.js
```diff
@@ -616,4 +617,42 @@ api.inviteToGroup = {
...
+ *
+ * @apiParam {string} groupId The group _id (or 'party')
+ *
+ * @apiSuccess {Object} Quest Object
+ */
+api.inviteToQuest = {
+  method: 'POST',
+  url: '/groups/:groupId/quests/invite/:questKey',
+  middlewares: [authWithHeaders(), cron],
+  async handler (req, res) {
```
crookedneighbor: Added failure conditions logic @SabreCat

TheHollidayInn:
I think it would be better to separate the quest routes into their own file.

paglias:
@crookedneighbor I think @crookedneighbor is right and we should have quest(s).js as we have chat.js

crookedneighbor:
Makes sense to me.

website/src/models/group.js
```diff
@@ -168,6 +170,64 @@ schema.methods.isMember = function isGroupMember (user) {
...
+  });
+
+  // TODO Add emails to users that quest has started to background ops
+
+  // These operations should run in the background
+  // and not hold up the quest routes from resolving
+  // TODO: What here?
+  // Q.all(backgroundOperations).then(() => {
+  // }).catch(err => {
+  // TODO: How to handle errors?
```
paglias: @crookedneighbor In these cases I think out best bet is to use Q.allSettled that in case of an error (user doesn't exist) it doesn't block the other promises from completing but waits for all of them to complete and throws at the end. Specifically for this case I think the main user.update op (setting quest details) could be a foreground operation, with notifications (email and push) in the background. 

Usually parties don't have more than 10-20 members

crookedneighbor: TIL `Q.allSettled`

What's the benefit of it being a foreground operation? Other than the initiating user, is it important that the other users updates happen before the request completes?

website/src/models/group.js
```diff
@@ -168,6 +170,64 @@ schema.methods.isMember = function isGroupMember (user) {
   }
 };
 
+schema.methods.startQuest = function startQuest () {
+  if (this.type !== 'party') throw new BadRequest('Must be a party to use this method');
+  if (!this.quest.key) throw new BadRequest('Party does not have a pending quest');
+  if (this.quest.active) throw new BadRequest('Quest is already active');
```
crookedneighbor: Are these sanity checks important? These are mostly for programmers using this method, they're not intended to be passed back to the user. I could go either way on them.

website/src/models/group.js
paglias: @crookedneighbor at least the group should be saved in foreground, we don't want users to think their action didn't succeed and start clicking again and again the start quest / invite quest button

website/src/models/group.js
paglias: @crookedneighbor I think we need them on the controller probably but here yeah they don't make much sense, I'd keep them only if the startQuest would become an async function / promise so that they would be passed to the errorHandler middleware

website/src/models/group.js
crookedneighbor: Oh, definitely. I think the group and the current user request should be, which means the quest start method might need to take in a user for that to work.

website/src/models/group.js
crookedneighbor: I think it's going to have to be asyncronous, actually, because it needs to fetch the email details from the members.

website/src/controllers/api-v3/quests.js
```diff
@@ -0,0 +1,98 @@
...
+    let group = await Group.getGroup({user, groupId: req.params.groupId, fields: 'type quest'});
+
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+    if (group.type !== 'party') throw new NotAuthorized(res.t('guildQuestsNotSupported'));
+    if (!quest) throw new NotFound(res.t('questNotFound', { key: questKey }));
+    if (!user.items.quests[questKey]) throw new NotAuthorized(res.t('questNotOwned'));
+    if (user.stats.lvl < quest.lvl) throw new NotAuthorized(res.t('questLevelTooHigh', { level: quest.lvl }));
+    if (group.quest.key) throw new NotAuthorized(res.t('questAlreadyUnderway'));
+
+    let members = await User.find({ 'party._id': group._id }, 'auth.facebook auth.local preferences.emailNotifications').exec();
```
paglias: just select auth here, we'll also need to fetch preferences for pull notifications

paglias: hmm I don't get why we're fetching data here that is never used, also if you change member.party below you have to fetch that too. Actually there's no need to fetch members at all, just use User.update()

website/src/controllers/api-v3/quests.js
```diff
@@ -0,0 +1,98 @@
...
+    group.markModified('quest');
+    group.quest.key = questKey;
+    group.quest.leader = user._id;
+    group.quest.members = {};
+    group.quest.members[user._id] = true;
+
+    user.party.quest.RSVPNeeded = false;
+    user.party.quest.key = questKey;
+
+    _.each(members, (member) => {
```
paglias, line 87: hmm here it's replicating what startQuest does

website/src/models/group.js
```diff
@@ -168,6 +170,81 @@ schema.methods.isMember = function isGroupMember (user) {
...
+    member.party.quest.completed = null;
+    member.markModified('party.quest');
+
+    if (this.quest.leader === member._id) {
+      member.items.quests[this.quest.key] -= 1;
+      member.markModified('items.quests');
+    }
+
+    if (member._id !== user._id) {
+      backgroundOperations.push(member.save());
```
paglias: same as in the controller, no need to save, just use User.update

paglias: Even though you still have to fetch data to send emails I think the best thing here is not to save but to use a single User.update for all users

website/src/controllers/api-v3/quests.js
```diff
@@ -0,0 +1,98 @@
...
+        group.quest.members[member._id] = null;
+        member.party.quest.RSVPNeeded = true;
+        member.party.quest.key = questKey;
+        // TODO: Send Quest invite email
+        backgroundOperations.push(member.save());
+      }
+    });
+
+    if (canStartQuestAutomatically(group)) {
+      group.startQuest(user);
```
paglias: await?

crookedneighbor: You are right, this should be await.

website/src/controllers/api-v3/quests.js
crookedneighbor, line 87: It's not. This sets all the user's in the quest members object to null (not yet decided) and RSVPNeeded to true, except for the inviting user, which gets set to true (a couple lines above). I think the only duplicated logic is the `member.quest.key` setting, which I'm not sure is necessary in the start quest method. It gets used in the invite modal, but I don't think it's used elsewhere. 

Quest start does completely different logic.

website/src/models/group.js
crookedneighbor: I started doing that originally, but then it got a little weird to do `User.update` calls for the other members, and direct changes to the initiating user (the last to accept/reject the invitation or force starts it). Should the initiating user be included in the mass update? And just make that a blocking action before the route resolves? Or should we do two `User.update` calls? One for the other members which runs in the background, and one for the user which waits?

website/src/controllers/api-v3/quests.js
crookedneighbor: We do need to fetch members in order to send out the quest invite email. (Just haven't added that logic yet)

website/src/models/group.js
paglias: I'd say that as long as quests are only for parties just do one mass update that blocks the route

website/src/controllers/api-v3/quests.js
paglias, line 87: Ohh I see... I thought this was some sort of leftover before you moved the logic into quest start :stupid:

website/src/models/group.js
```diff
@@ -168,6 +170,81 @@ schema.methods.isMember = function isGroupMember (user) {
   }
 };
 
+schema.methods.startQuest = async function startQuest (user) {
+  if (this.type !== 'party') throw new BadRequest('Must be a party to use this method');
+  if (!this.quest.key) throw new BadRequest('Party does not have a pending quest');
+  if (this.quest.active) throw new BadRequest('Quest is already active');
+
```
TheHollidayInn: Shouldn't these be `r.t('stirng')` ? I have added some of these strings to common.

crookedneighbor: My intention for these is that they never reach the user. They're sanity checks for programmers to ensure that you don't use `startQuest` if the group is not a party, is not on a quest, or if the quest is already active. These are cases that should never happen in the routes.

In which case, maybe it's better to do InternalServer error so it 500s?

Either that, or just abandon these checks entirely. I think @paglias was in favor. I'm still undecided.

paglias: I don't have a strong opinion for this but since those errors can only be caused by a programmer error or lack of data I think InteralServer errors are better

TheHollidayInn: Actually, I believe in the specification for http error codes it states that 400 error codes you depict user errors and 500 should depict fault in our part. So, we shouldn't give 500s if it is the user's fault.

paglias: @TheHollidayInn the fact is that we already have checks in the controllers for the user input, the errors here can only happen if a developer calls the questStart method with in the wrong situazion or with the wrong parameters, it's our fault

TheHollidayInn: Ah, that's a good point. I agree.

PR #6566, opened 2016-01-22
Remove from array

crookedneighbor:
# Refactor

Add method on baseModel to remove items from arrays on a mongoose object. 

@paglias @TheHollidayInn @KristianTashkov


website/src/libs/api-v3/baseModel.js
```diff
@@ -45,6 +46,24 @@ export default function baseModel (schema, options = {}) {
     return options.sanitizeTransform ? options.sanitizeTransform(objToSanitize) : objToSanitize;
   };
 
+  schema.methods.removeFromArray = function removeFromArray (arrayPath, item) {
```
crookedneighbor: I feel like this could have a better name. Ideas?

KristianTashkov:
Just my opinion, but I find this very unnecessary. The arrayPath syntax is clunky and the method is something that you must know is there, new contributors will probably use splice anyway, because everyone does it in javascript. Also makes the code unfamiliar when reading for new contributors, while they understand indexOf/splice. Overall it's nice that it saves you one function call, but I'm not sure it's worth it for the _clunkyness_ it introduces.

DISCLAIMER: I'm not a big node/js coder so writing 'tasks.todos' as a string may be normal, but I find it weird.

paglias:
I think if we want this (don't have a strong preference) there's no point in having it linked to mongoose. Just drop it in the /libs folder and use it as `removeFromArray(array, item)`. No strings to pass and works with every array

KristianTashkov:
I agree with @paglias, but shouldn't we just use `_.pull` in that case?

Edit: if it triggers mongoose's modified check

KristianTashkov:
Also come to think of it, shouldn't all these tests save, get the document and then assert to see if works as expected?

paglias:
@KristianTashkov yeah I'm not sure if _.pull triggers mongoose check and even if it does now it may change in future, so we can try if it works in which case we may use it but should add a very quick test to make sure it keeps working

crookedneighbor:
My main goal with this is to make the functions more readable. I know what doing `indexOf` and `slice` on an array do, but it always takes me a second or two longer to parse when I'm reading the logic.

> DISCLAIMER: I'm not a big node/js coder so writing 'tasks.todos' as a string may be normal, but I find it weird.

I'd say it's pretty common when working with mongo. If you're not working with mongo, I agree, kind of weird. 

> Also makes the code unfamiliar when reading for new contributors, while they understand indexOf/splice.

I'd argue that contributors are more likely to see what code is already doing and emulate that. Having one function to do the array element removal ensures that it is done in a way that mongoose can handle.

Re: `_.pull`, what @paglias said is correct. Even if it works now, we can't guarantee it'll work in the future.

I'd be fine moving this off the baseModel plugin and into it's own lib. My main goal for adding it to the baseModel was learn how to use the plugin. Now that I know, I think you're right, there's no need to include it there. (I do think we should include a comment about why splice is being used specifically instead of `_.pull`) 

Maybe it could be a `removeElement` function in an `array-manipulators.js` file? What I don't want is for this to end up in a `uitls.js` file. Utils files are evil and tend to bloat. 

Again, ultimately, I want the code to be readable. If we decide that the `findIndex` and `if index !==1, splice` syntax is the most readable, that's what we should do. 

Right now I have it return the altered array if the element could be removed and return false if it couldn't. I think instead it should return the removedItem if it was removed. Thoughts?

paglias:
for _.pull we could simply write a test using it:

```
_.pull(modelInstance.array, item)
expect(modelInstance.modifiedPaths() === ['array']).to.equal(true)
```

TheHollidayInn:
Hmmm. Honestly, this seems like a lot of work for a cleaner remove, but, this idea is the Facade pattern at work. Looking at the actual function, you are handling some extra cases, which is good. This would be good as some utility library we have. 

Can we generalize this to remove from collection? Also, I don't understand the plugin stuff. Where can I read about that?

paglias:
@TheHollidayInn the plugin is a mongoose plugin http://mongoosejs.com/docs/plugins.html

I created it so that all models could have access to a standard set of features that are used everywhere, basically it adds an _id field with a default value and support blacklisting properties so they're not returned when calling toJSON and a few other utilities

paglias:
_.pull doesn't mark the path as modified. I personally don't have any problem with using indexOf/splice but adding a custom removeFromArray utility won't hurt. Let's just do it without getting too blocked by this :)

crookedneighbor:
This has been updated to move all logic to a separate lib instead of piggy-backing on the baseModel plugin.

I decided to change the behavior to return the removed item instead of the altered array.

test/api/v3/unit/libs/collectionManipulators.test.js
```diff
@@ -0,0 +1,71 @@
+import {
+ removeElementFromArray,
+} from '../../../../../website/src/libs/api-v3/collectionManipulators';
+
+describe('Collection Manipulators', () => {
+  describe('removeElementFromArray', () => {
+    it('removes item from specified array on document', () => {
```
KristianTashkov, line 8: Is this supposed to test mongo documents? If not, don't we need such a test, because we expect this function to mark the array as modified?

website/src/libs/api-v3/collectionManipulators.js
```diff
@@ -0,0 +1,20 @@
+import { findIndex } from 'lodash';
+
+export function removeElementFromArray (array, element) {
```
KristianTashkov: I would suggest skipping the `Element` and just leave it `removeFromArray`

crookedneighbor: I think you're right. I've been struggling to find a good name for this. I think `removeFromArray` does the job.

test/api/v3/unit/libs/collectionManipulators.test.js
crookedneighbor, line 8: I think you're right on this. I've added a test.

test/helpers/api-unit.helper.js
```diff
@@ -1,10 +1,27 @@
 import '../../website/src/libs/api-v3/i18n';
+import mongoose from 'mongoose';
 import { defaultsDeep as defaults } from 'lodash';
 import { model as User } from '../../website/src/models/user';
 import { model as Group } from '../../website/src/models/group';
 
-afterEach(() => {
+mongoose.Promise = require('q').Promise;
+
+mongoose.connect('mongodb://localhost/habitica-unit-tests');
```
crookedneighbor, line 9: I've added this to the unit helper so we can actually write unit tests for mongoose operations if we'd like.

test/api/v3/unit/libs/collectionManipulators.test.js
```diff
@@ -0,0 +1,88 @@
+import mongoose from 'mongoose';
+import {
+ removeFromArray,
+} from '../../../../../website/src/libs/api-v3/collectionManipulators';
+
+describe('Collection Manipulators', () => {
+  describe('removeFromArray', () => {
+    it('removes item from specified array on document', () => {
```
KristianTashkov, line 8: Shouldn't this just be `removes element from array`?

test/api/v3/unit/libs/collectionManipulators.test.js
```diff
@@ -0,0 +1,88 @@
...
+
+    it('returns false if item is not found', () => {
+      let array = ['a', 'b', 'c'];
+
+      let result = removeFromArray(array, 'z');
+
+      expect(result).to.eql(false);
+    });
+
+    it('removal of element persists when mongoose document is saved', async () => {
```
KristianTashkov, line 72: `it('persists removal of element when mongoose document is saved'` to make it sound like a sentence?

KristianTashkov:
Is there a reason `collectionManipulators.js` is in an api-v3 folder? This isn't something specific for the API or v3.

crookedneighbor:
Closed with e5f1c44a3e7aa24c3bddfa5ab37055bd546d9365

PR #6569, opened 2016-01-22
Color code quest accepted/rejected text

shellbot:
I noticed that in the iOS app, the list of quest participants is easier to scan as party members who accept the quest are highlighted in green. As I'm more of a visual person, thought I'd dip a toe into the codebase and try to implement the same thing on the website.

### DIFF
--- a/website/public/css/quests.styl
+++ b/website/public/css/quests.styl
@@ -8,3 +8,9 @@ quest-rewards
 
 .quest-icon
   margin: .6em
+
+.quest-participants .accepted
+  color: #2DB200
+
+.quest-participants .rejected
+  color: #B30409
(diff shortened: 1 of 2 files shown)


Alys:
@shellbot This sounds like a nice idea! Can you please attach a screenshot showing what it looks like? That will help the staff review it and provide feedback.

Alys:
That looks nice! I've run various filters over the screenshot to mimic various types of colour vision deficiencies and the words were clear for all of them. I'll repeat the tests on the actual webpage if I can get my local install to stop being broken... :)

Alys:
I've tested this on my local install with various vision filters and the red and green work well for all of them, on both white and grey backgrounds.

@shellbot What's your User ID?

Alys:
Tier awarded! Note that the higher tiers require more effort than the lower ones, so if you make any small changes from now on, you might find that they won't grant you an extra tier immediately, but they will be noted as partial credit towards a tier. There's a bit of information about that at [Contributor Rewards](http://habitrpg.wikia.com/wiki/Contributor_Rewards) but do ask if you have questions. 

Thank you @shellbot! I like this change; it's a nice little enhancement that we hadn't thought of.

PR #6571, opened 2016-01-22
Added initial clear chat flags tests

TheHollidayInn:
These are the clear flags tests for #6509. We might want to use my new generate chat function when that PR gets merged, but I feel pretty comfortable using these tests as well.

### DIFF
--- a/website/src/controllers/api-v3/chat.js
+++ b/website/src/controllers/api-v3/chat.js
@@ -4,6 +4,7 @@ import { model as Group } from '../../models/group';
 import { model as User } from '../../models/user';
 import {
   NotFound,
+  NotAuthorized,
 } from '../../libs/api-v3/errors';
 import _ from 'lodash';
 import { sendTxn } from '../../libs/api-v3/email';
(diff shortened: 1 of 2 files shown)


website/src/controllers/api-v3/chat.js
```diff
@@ -256,4 +257,51 @@ api.flagChat = {
...
+    }
+
+    message.flagCount = 0;
+
+    await Group.update(
+      {_id: group._id, 'chat.id': message.id},
+      {$set: {'chat.$.flagCount': message.flagCount}}
+    );
+
+    res.respond(204, message.flagCount);
```
paglias: 204 means no content. but for compatibility with mobile apps we prefer to always return an empty object so I'd do `res.respond(200, {})`

website/src/controllers/api-v3/chat.js
```diff
@@ -256,4 +257,51 @@ api.flagChat = {
...
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let group = await Group.getGroup({user, groupId});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    let message = _.find(group.chat, {id: chatId});
+    if (!message) throw new NotFound(res.t('messageGroupChatNotFound'));
+
+    if (!user.contributor.admin) {
```
paglias: I'd move it right after parameter validation so we don't fetch the group if user is not an admin

website/src/controllers/api-v3/chat.js
TheHollidayInn: Should I still return the flag count?

TheHollidayInn:
Updated!

website/src/controllers/api-v3/chat.js
```diff
@@ -256,4 +257,51 @@ api.flagChat = {
...
+/**
+ * @api {post} /groups/:groupId/chat/:chatId/clear-flags Clear a group chat message's flags
+ * @apiVersion 3.0.0
+ * @apiName ClearFlags
+ * @apiGroup Chat
+ *
+ * @apiParam {groupId} groupId The group _id
+ * @apiParam {chatId} chatId The chat message _id
+ *
+ * @apiSuccess {Array} chat An array of chat messages
```
crookedneighbor: We're not returning an array of chat messages in this route.

TheHollidayInn:
Fixed the note.

PR #6574, opened 2016-01-23
Fixed send message and added intiial seen chat tests

TheHollidayInn:
Seen chat tests and route for #6509. The v2 route skips the auth middleware. Should we continue to do that?

### DIFF
--- a/website/src/controllers/api-v3/chat.js
+++ b/website/src/controllers/api-v3/chat.js
@@ -105,7 +105,7 @@ api.postChat = {
  * @apiSuccess {Array} chat An array of chat messages
  */
 api.likeChat = {
-  method: 'Post',
+  method: 'POST',
   url: '/groups/:groupId/chat/:chatId/like',
   middlewares: [authWithHeaders(), cron],
   async handler (req, res) {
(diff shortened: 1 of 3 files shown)


website/src/models/group.js
```diff
@@ -205,7 +205,8 @@ schema.methods.sendChat = function sendChat (message, user) {
     // User.update({'profile.name':{$in:profileNames}},lastSeenUpdate,{multi:true}).exec();
   } else {
     User.update({
-      _id: {$in: this.members, $ne: user ? user._id : ''},
+      guilds: this._id,
```
paglias: parties are not supported?

TheHollidayInn:
Added! I need to test for send message in the post chat tests. I will add those tests when I rewrite them to have the await/async syntax.

website/src/controllers/api-v3/chat.js
```diff
@@ -256,4 +256,40 @@ api.flagChat = {
...
+
+    req.checkParams('groupId', res.t('groupIdRequired')).notEmpty();
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let group = await Group.getGroup({user, groupId});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    // Skip the auth step, we want this to be fast. If !found with uuid/token, then it just doesn't save
```
paglias, line 285: the auth middleware is used (which is :+1: for me) so we should remove this comment probably

PR #6579, opened 2016-01-25
Updated chat tests with new async/await syntax

TheHollidayInn:
This is a refactor of previous tests that were merged. I have updated the syntax and added a test or two.


crookedneighbor:
Looks :+1: to me.

Issue #6562, opened 2016-01-21
Tags not visible when turning off the Header


NicolasWebDev:
Hi,

When I turn the header off in FF 43.0.4, I don't have access to the tags list anymore (nor the search field for that matter).

Alys:
I can't replicate this in Firefox 43.0.4 on Mac OS 10.6.

@Sathors Can you please tell us your User ID? We'd like to be able to examine your exact settings to see if there's a particular combination that causes this. If you're not sure how to find your User ID, go to Help -> Report a Bug and see step 6.

Can you recall approximately when this bug first started?

NicolasWebDev:
@Alys , my User ID is 3f6ce5ca-c64c-47ea-ad1c-22d75aaa0da4.
I am using Linux Mint 13.

Actually I created my avatar yesterday.
Now that I am toying a little bit with it, I realized that if the window is full-screen, then I see the complete menus above on two lines. The tags do not appear.
If I resize the window to half-screen, then the top menu takes up only one line, and now the tags/search bar do appear.

Alys:
Could you show us screenshots of each case please? If you normally update
github issues by email, you'll need to go to the github website to add
images (drag into a new comment).

NicolasWebDev:
Here is the "full-screen" case:
![First case](http://s14.postimg.org/5w9womk01/Habitica_Tasks_Pentadactyl_004.png)

Here is the "small-screen" case:
![Second case](http://s12.postimg.org/xfnvoguzx/Habitica_Tasks_Pentadactyl_005.png)

shellbot:
Reproduced this in Chrome too, looks like the responsive styles leave a window where this occurs at a certain screen width.

I'll be happy to look at a CSS fix, if nobody else is on it?

Alys:
Hi @shellbot. That would be great, thank you!

There's a recent change that might be relevant: 
https://github.com/HabitRPG/habitrpg/pull/6404
It's possible that the values in website/public/css/variables/screen-size.styl might need tweaking.
However I haven't looked into this in any kind of detail, so that might be irrelevant.

shellbot:
Investigating!

Alys:
This seems to be fixed (see screenshot below from Chrome) but if anyone is still experiencing this, comment here with which browser and OS you're using, including the versions of both, and with a screenshot showing the menu and area where the tags should be.

[...]

PR #6581, opened 2016-01-26
Fix header overlap on small screens

shellbot:
Tiny CSS fix for #6562, where the navigation bar was covering tags and search on smaller screens.

Used specific widths instead of variables as the problem didn't really bear much relation to existing breakpoints. See before and after pic:

header-menu-css-fix

Please note: Only tested on Windows (Chrome, FF, IE11)

### DIFF
--- a/website/public/css/header.styl
+++ b/website/public/css/header.styl
@@ -12,6 +12,8 @@
   overflow-y: hidden
   overflow-x: auto
   // position relative
+  @media screen and (min-width: 768px) and (max-width: 1014px) //extra padding for smaller screen sizes
+    padding-top: 92px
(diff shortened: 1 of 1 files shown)


crookedneighbor:
Your change causes extra padding in circumstances when there shouldn't be:

screen shot 2016-01-25 at 10 18 51 pm

crookedneighbor:
Mac Chrome, but I don't think it's specific to that browser.

crookedneighbor:
@shellbot Any update on this?

Alys:
@shellbot I'm going to close this for now, but we can reopen it later. If you'd like advice about how to go further with this, just comment here - we'll see it even when this is closed!

PR #6582, opened 2016-01-26
Adds user details to the removal modal

### LINKED ISSUE #6578: Confirm name of user being booted from party
When booting the party member, there is a confirmation screen, but no username present. It's hard to tell if you're booting the right person or not. 
screenshot 2016-01-24 17 48 40
I imagine booting activity doesn't happen too often, but fixing it so that the username of the player being booted would be helpful.

tylernotfound:
Fixes #6578 

Adds the ability to see the username within the removal modal for parties and guilds.

### DIFF
--- a/website/public/css/index.styl
+++ b/website/public/css/index.styl
@@ -109,13 +109,6 @@ html, body
 /* Misc
 -------------------------------------------------- */
 //move
-
-.modal-subheading {
-  margin: 12px 0 8px;
-  font-size: 24px;
-  font-weight: bold;
-}
-
 .new-stuff-modal // a .modal-body
   h2, h3
     font-weight: 700
--- a/website/public/css/index.styl
+++ b/website/public/css/index.styl
@@ -110,6 +110,12 @@ html, body
 -------------------------------------------------- */
 //move
 
+.modal-subheading {
+  margin: 12px 0 8px;
+  font-size: 24px;
+  font-weight: bold;
+}
+
 .new-stuff-modal // a .modal-body
   h2, h3
     font-weight: 700
--- a/website/views/shared/modals/members.jade
+++ b/website/views/shared/modals/members.jade
@@ -116,7 +116,7 @@ script(type='text/ng-template', id='modals/abuse-flag.html')
 script(type='text/ng-template', id='modals/remove-member.html')
   .modal-header.text-center
     h4=env.t('sureKick')
-    p.modal-subheading {{::removeMemberData.member.profile.name}}
+    p.lead: strong {{::removeMemberData.member.profile.name}}
   .modal-body
     textarea.form-control(type='text',rows='5',placeholder=env.t('optionalMessage'),ng-model='removeMemberData.message')
   .modal-footer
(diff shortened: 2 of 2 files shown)


crookedneighbor:
Could you provide a before and after screenshot?

tylernotfound:
**Before:**
image

**After:**
image

tylernotfound:
@crookedneighbor There's a little bit more explanation as to this decision in that issue thread.

website/views/shared/modals/members.jade
```diff
@@ -114,8 +114,9 @@ script(type='text/ng-template', id='modals/abuse-flag.html')
     button.btn.btn-danger(ng-click='reportAbuse(user, abuseObject, groupId)')=env.t("abuseFlagModalButton")
 
 script(type='text/ng-template', id='modals/remove-member.html')
-  .modal-header
+  .modal-header(class='text-center')
```
crookedneighbor: This should be `.modal-header.text-center`

website/views/shared/modals/members.jade
```diff
@@ -114,8 +114,9 @@ script(type='text/ng-template', id='modals/abuse-flag.html')
...
     h4=env.t('sureKick')
+    p(class='modal-subheading') {{removeMemberData.member.profile.name}}
```
crookedneighbor: this should be `p.modal-subheading {{::removeMemberData.member.profile.name}}` (notice the `::` to signify one-time-binding)

tylernotfound:
Doh, sorry for the syntax miscues. ^__^

website/public/css/index.styl
```diff
@@ -110,6 +110,12 @@ html, body
 -------------------------------------------------- */
 //move
 
+.modal-subheading {
+  margin: 12px 0 8px;
+  font-size: 24px;
+  font-weight: bold;
+}
```
crookedneighbor: I feel like there is a bootstrap class we could leverage instead of creating a new class. Maybe a combination of a strong element nested in a [lead class](http://getbootstrap.com/css/#lead-body-copy)?

```
p.lead: strong {{::removeMemberData.member.profile.name}}
```

tylernotfound: `.lead` has 20px of bottom margin, which exaggerates the spacing. I'm not sure there's much advantage in changing it, but I'll defer to your preference. Screenshot of your suggested change, for reference:

image

Alys: It doesn't look too bad with the extra spacing from the .lead class, and keeping our css simpler will help with maintenance. Let's go with .lead, if that's okay with you, @tylerxo ?

tylernotfound: Yep, of course.

SabreCat:
"party/guild" is kinda awkward. Any chance we can get that to display the appropriate group type dynamically?

crookedneighbor:
Seems like a separate category concern to me @sabrecat

tylernotfound:
@SabreCat there's translation issues with that, from what I could tell. You'd have to translate those words individually across all languages, since they are set up as a single translation text right now.

crookedneighbor:
@tylerxo From a code perspective, you'd just change the english version. We'd update the translations on transifex, and it would automatically set those strings for retranslation in the other languages.

But, like I said, that change is separate from this PR.

Alys:
Thank you @tylerxo! Tier 2 awarded.

PR #6583, opened 2016-01-26
Fix laundromancer quest text

KristianTashkov:
I thought I was getting another type of potion, but got a base potion on completion. This should fix this problem for future adventurers :stuck_out_tongue:


Alys:
Thanks for noticing this! We don't actually accept changes to any of the non-`en` locales directories in pull requests (see the readme file in the `en_GB` directory for details), so I've push the change to the `en` file in commit https://github.com/HabitRPG/habitrpg/commit/6c106ad64429a7c897e5dc7a4273a79755e18287

KristianTashkov:
@Alys, @blade mention to comment when a PR was not added to contribution notes :blush:

Alys:
Sorry about that! It was deliberate but I should have said something. As your tier increases, it takes progressively more work to achieve the next one, so for the higher tiers, several PRs are required, with each one adding to your credit. This PR would have no significant impact on your credit for tier 4 since it was small and wasn't actually used directly. However since you've asked, I've added it to your list of contributions. :)

KristianTashkov:
Heh no problem, I like having it in the list :stuck_out_tongue: Thanks!

PR #6586, opened 2016-01-26
#6560: Add alt attributes for images on front page

### LINKED ISSUE #6560: Many images on front page have no alt attribute
According to the [HTML spec](https://www.w3.org/html/wg/drafts/html/master/semantics.html#alt), every image must have an alt attribute, but most of the images on the front page don't.

I'm happy to write alt-texts for the images, but presumably they will need translating; is `common/locales/en/front.json` the right place for them?

keysmashes:
Fixes #6560.

I'm open to suggestions about which images should have alt-text; for example, I left out alt-text for the section about spending gold on rewards, because the images purely reinforce the point made in the text, but if anyone can think of helpful alt-text to put there, I can add it in. The problem I found with the visual parts was that the arrows are font icons, not images, and so they can't have alt-text.


Alys:
Making the font icons accessible is definitely something we should do, but I suggest we do that in a separate piece of work, so that we can push this PR out sooner. Thank you for mentioning it!

I'm going to put your PR onto a test web server for the staff to review it. I'm experimenting with an extension that will let them toggle between image and alt text so that they can provide feedback.

Alys:
Sorry about the delay on this. I've found an extension to display alt attributes that works with the site (the first couple I tried didn't for some reason) and the attributes you've chosen look good to me. I'm passing information on to the staff about viewing them on a private test server.

Alys:
Thank you @anowlcalledjosh ! Tier awarded.

keysmashes:
Awesome, thanks @Alys! Which extension did you use? I couldn't find one that worked at all.

Alys:
The Web Developer Toolbar in both Chrome and Firefox
https://chrome.google.com/webstore/detail/web-developer/bfbameneiokkgbdmiekhjnmfkcnldhhm/related
https://addons.mozilla.org/en-US/firefox/addon/web-developer/developers

and Popup ALT Attribute in Firefox
https://addons.mozilla.org/en-US/firefox/addon/popup-alt-attribute/

keysmashes:
Aha, thank you!

PR #6591, opened 2016-01-27
Challenge leave route and tests

KristianTashkov:
Challenge leave route for https://github.com/HabitRPG/habitrpg/issues/6509
As with join I opted with testing that the main functionalities of `unlinkChallengeTasks` are called, and the more detailed ones with be left for unit tests.

### DIFF
--- a/test/api/v3/integration/challenges/POST-challenges_challengeId_leave.test.js
+++ b/test/api/v3/integration/challenges/POST-challenges_challengeId_leave.test.js
@@ -1,0 +1,123 @@
+import {
+  generateUser,
+  generateChallenge,
+  createAndPopulateGroup,
+  translate as t,
+} from '../../../../helpers/api-v3-integration.helper';
+import { v4 as generateUUID } from 'uuid';
+
+describe('POST /challenges/:challengeId/leave', () => {
+  it('returns error when challengeId is not a valid UUID', async () => {
+    let user = await generateUser();
+
+    await expect(user.post('/challenges/test/leave')).to.eventually.be.rejected.and.eql({
+      code: 400,
+      error: 'BadRequest',
+      message: t('invalidReqParams'),
+    });
+  });
+
+  it('returns error when challengeId is not for a valid challenge', async () => {
+    let user = await generateUser();
+
+    await expect(user.post(`/challenges/${generateUUID()}/leave`)).to.eventually.be.rejected.and.eql({
+      code: 404,
+      error: 'NotFound',
+      message: t('challengeNotFound'),
+    });
+  });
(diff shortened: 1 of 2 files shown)


test/api/v3/integration/challenges/POST-challenges_challengeId_leave.test.js
```diff
@@ -0,0 +1,123 @@
...
+      expect(tasksTexts).to.not.include(taskText);
+    });
+
+    it('doesn\'t unlink challenge tasks from leaving user when remove-all isn\'t passed', async () => {
+      await leavingUser.post(`/challenges/${challenge._id}/leave`, {
+        keep: 'test',
+      });
+
+      let tasks = await leavingUser.get('/tasks/user');
+      let testTask = _.find(tasks, (task) => {
```
paglias, line 115: here would you mind to check that task.challenge has been set to an empty object? (it's possible the api doesn't return task.challenge at all if it's an empty object but it's okay, in that case just test that task.challenge doesn't exist)

paglias, line 115: I'm stupid sorry :) I had to scroll down a little more and would have seen that you already tested it :)

website/src/controllers/api-v3/challenges.js
```diff
@@ -130,6 +130,44 @@ api.joinChallenge = {
...
+    let group = await Group.getGroup({user, groupId: challenge.groupId, fields: '_id type privacy'});
+    if (!group || !challenge.canView(user, group)) throw new NotFound(res.t('challengeNotFound'));
+
+    if (!challenge.isMember(user)) throw new NotAuthorized(res.t('challengeMemberNotFound'));
+
+    challenge.memberCount -= 1;
+
+    // Unlink challenge's tasks from user's tasks and save the challenge
+    await Q.all([user.unlinkChallengeTasks(challenge._id, keep), challenge.save()]);
+    res.respond(200, challenge);
```
paglias: @KristianTashkov the PR looks good, the only thing is that I'm unsure if returning the challenge is correct here since you just left it

KristianTashkov: I copied the behavior from api-v2 since I guessed it was used for something. If we don't need it I can return an empty object.

paglias: yeah an empty object feels more natural to me

KristianTashkov:
@paglias  updated, leave route now returns empty object on success.

paglias:
@KristianTashkov level 4 awarded :)

PR #6599, opened 2016-01-27
Fix failing unit tests on windows

KristianTashkov:
These two tests were failing on windows, because of the different path seperator.

We discussed with @blade how to fix the `is an absolute path to common/locales/` test, and arrived to the conclusion that the current test doesn't really test what it describes. Even if we write it to do what it says, it's just going to be a repetition of the code that creates the variable so there is no point.

For the second test I just added the `path.sep` in the regex.


crookedneighbor:
Noted.

PR #6600, opened 2016-01-27
Challenge select winner route tests

KristianTashkov:
The same as my previous challenge routes tests, I tested the main functionality and left the specifics of `_closeChal` for unit tests.

### DIFF
--- a/test/api/v3/integration/challenges/POST-challenges_challengeId_winner_winnerId.test.js
+++ b/test/api/v3/integration/challenges/POST-challenges_challengeId_winner_winnerId.test.js
@@ -44,6 +44,7 @@ describe('POST /challenges/:challengeId/winner/:winnerId', () => {
     let group;
     let challenge;
     let winningUser;
+    let taskText = 'A challenge task text';
 
     beforeEach(async () => {
       let populatedGroup = await createAndPopulateGroup({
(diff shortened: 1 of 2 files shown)


test/api/v3/integration/challenges/POST-challenges_challengeId_winner_winnerId.test.js
```diff
@@ -0,0 +1,114 @@
...
+
+      groupLeader = populatedGroup.groupLeader;
+      group = populatedGroup.group;
+      winningUser = populatedGroup.members[0];
+
+      challenge = await generateChallenge(groupLeader, group, {
+        prize: 1,
+      });
+
+      await groupLeader.post(`/tasks/challenge/${challenge._id}`, [
```
paglias, line 62: @KristianTashkov since you create a task here would you mind adding a small test to make sure that members' tasks are getting the correct `task.challenge.broken` and `task.challenge.winner` values? I know this type of checks belong to unit tests but since we're a bit lacking on them I'd like to be sure that at least it's not completely broken.

Other than that it looks good :)

KristianTashkov, line 62: Sure, I will update the PR tonight.

KristianTashkov:
@paglias updated with the test.

PR #6617, opened 2016-01-29
Add in missing word your to passMan phrase.

murraybd:
I noticed that the dialog when entering your username and password was missing a word so added it.

### DIFF
--- a/common/locales/en/front.json
+++ b/common/locales/en/front.json
@@ -124,7 +124,7 @@
     "motivate1": "Motivate yourself to do anything.",
     "motivate2": "Get Organized. Get Motivated. Get Gold.",
     "passConfirm": "Confirm Password",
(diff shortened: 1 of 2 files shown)


Alys:
Thanks! However we don't actually update the locales files through GitHub - see the _README_FIRST.md file in common/locales/en_GB for details. Can you modify the PR to remove that file?

Alys:
Thanks! What's your User ID (from Settings -> API)?

murraybd:
5b93ce83-2508-4ea4-9647-079899dffa96 is my user id.

Alys:
Tier awarded! You can read more about contributor tiers and rewards at http://habitica.wikia.com/wiki/Contributor_Rewards

Note that as your tier increases, it takes progressively more work to achieve the next one, so for the higher tiers, larger PRs or a few PRs are required. There's some information about that at that page, but feel free to ask here if you have questions!

PR #6629, opened 2016-01-31
Add .img-rendering-auto to several images

keysmashes:
Whilst things like avatars are pixel art, the fact that they get resized means that they look very jagged when rendered pixelated. With `image-rendering: auto`, they look significantly better, for example supermouse35's avatar in the Health section of the carousel.

Live site:

image

image

PR:

image

image

### DIFF
--- a/website/views/static/front.jade
+++ b/website/views/static/front.jade
@@ -214,13 +214,13 @@ html(ng-app='habitrpg', ng-controller='RootCtrl')
                   img.sample-img.visible-lg-inline-block(src='https://d2afqr2xdmyzvu.cloudfront.net/front/images/uses/coding.png')
                   .usetweet-groups.col-lg-5.col-md-6.visible-lg-inline-block.visible-md-inline-block
                     .usetweet-group
-                      img(data-toggle='tooltip', data-placement='top', title='frabjabulous', src='https://d2afqr2xdmyzvu.cloudfront.net/front/images/testimonials/frabjabulous.png')
+                      img.img-rendering-auto(data-toggle='tooltip', data-placement='top', title='frabjabulous', src='https://d2afqr2xdmyzvu.cloudfront.net/front/images/testimonials/frabjabulous.png')
                       .usetweet.tweet.popover.right
                         .arrow
                         .popover-content
                           = env.t('frabjabulousQuote')
                     .usetweet-group
-                      img(data-toggle='tooltip', data-placement='top', title='_AlexandraSo_', src='https://d2afqr2xdmyzvu.cloudfront.net/front/images/testimonials/AlexandraSo.png')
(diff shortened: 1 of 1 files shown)


Alys:
@anowlcalledjosh I'm not really certain that this is something that we want. The pixellated appearance of the avatars is deliberate, as far as I'm aware. To my eyes, the rendered ones seem out of focus.

@lemoness ?

keysmashes:
@Alys I think it definitely doesn't look as good as it could if the images were either used at their native resolution or recaptured at whatever they're currently being used at*, but IMHO having smooth but out-of-focus images looks better than jagged artifacts all along diagonal edges (like the last picture in the OP, although the screenshot doesn't really do the pixelation justice). Of course if you think that the pixelated ones look better please do just close this :)

Also, currently half the images on the front page are `image-rendering: auto`, whereas the others are pixelated – they should presumably be more consistent, especially the 'equipment and extras' and 'social play' images at the bottom, one of which is pixelated, the other is not.

*some of the images are width 80 (players use Habitica to manage), some are width 120 (what people say), some are width 100 (all the others)

Alys:
It's really a question for @lemoness so we'll wait to see what she says. :)

SabreCat:
I'm in favor of blurring rather than janky badly sized pixels.

lemoness:
The thing that will look the best is if we resize the art assets to the correct size using nearest neighbor (or non-PS equivalent) rather than resizing them automatically. If you can get me the size needed for the assets, as well as a list of which ones are being badly resized, it should be pretty easy to get new files.  Unless there is a reason that they need to be resized automatically?

keysmashes:
The only thing I can think of is if the images get different sizes based on screen size? Assuming that's not an issue, the following images (all in `website/public/front/images/`) currently look janky:
- `Mount.png` (100px tall)
- `testimonials/`
  - `frabjabulous.png` (80px wide)
  - `AlexandraSo.png` (80px wide)
  - `EvaGantz.png` (80px wide)
  - `supermouse35.png` (80px wide)
  - `Althaire.png` (80px wide)
  - `InfH.png` (80px wide)
  - `Drei-M.png` (80px wide)
  - `Kazui.png` (80px wide)
  - `autumnesquirrel.png` (80px wide)
  - `irishfeet123.png` (80px wide)
  - `Zelah_Meyer.png` (100px wide)
  - `skysailor.png` (100px wide)

The images for 16bitFil, Elmi (100px wide) and the four images in the 'What people say' carousel (Drag0nSilver, frabjabulous, AndeeLiao and AlexandraSo) are all currently `image-rendering: auto`, as well as `Party-Header.png` (100px tall). This works for the carousel, because the images resize, but the other three are inconsistent with the other images.

lemoness:
Thanks! I will add this to my list. It may take some time.

Alys:
Since the preference is to resize the art assets, I'm going to close this, but it can be reopened if other admins think it's still useful as a temporary change.

PR #6631, opened 2016-01-31
Make task help icons consistent across browsers

keysmashes:
The `zoom` CSS property isn't supported by Firefox, and nor is `vertical-align: -webkit-baseline-middle`. `font-size: 1.5em` has the same effect, and `vertical-align: middle` ensures that the tooltip doesn't overlap with the icon. I don't actually know what `-webkit-baseline-middle` does, but having a WebKit-specific value with no fallback seems odd, and frankly the difference is so slight that we may as well go for consistency.

Live site vs. PR, Chrome:

image

image

Live site vs. PR, Firefox:

image

image

### DIFF
--- a/website/public/css/tasks.styl
+++ b/website/public/css/tasks.styl
@@ -149,8 +149,8 @@ for $stage in $stages
     font-weight: 300
 
   .option-box .glyphicon
-    zoom: 1.5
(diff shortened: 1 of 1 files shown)


SabreCat:
Good idea. Merged, and thank you!

PR #6640, opened 2016-02-02
Added quest reject route and intial tests

TheHollidayInn:
This is the initial quest reject code for #6509 .
- I am unsure how we are going to track members, so I left a comment
- @crookedneighbor  is working on the questStart method, so I left that commented
- And, I had to use update methods on the Api classes, since the quest invite route is unfinished

### DIFF
--- a/test/api/v3/integration/quests/POST-groups_groupid_quests_reject.test.js
+++ b/test/api/v3/integration/quests/POST-groups_groupid_quests_reject.test.js
@@ -30,7 +30,7 @@ describe('POST /groups/:groupId/quests/invite/:questKey', () => {
 
   context('failure conditions', () => {
     it('returns an error when group is not found', async () => {
-      await expect(member.post(`/groups/${generateUUID()}/quests/reject/${PET_QUEST}`))
+      await expect(member.post(`/groups/${generateUUID()}/quests/reject`))
         .to.eventually.be.rejected.and.eql({
           code: 404,
           error: 'NotFound',
@@ -38,43 +38,10 @@ describe('POST /groups/:groupId/quests/invite/:questKey', () => {
         });
     });
 
-    it('returns an error when group is not a party', async () => {
-      let { group, groupLeader } = await createAndPopulateGroup({
-        groupDetails: { type: 'guild', privacy: 'private' },
-      });
-
-      await expect(groupLeader.post(`/groups/${group._id}/quests/reject/${PET_QUEST}`))
-        .to.eventually.be.rejected.and.eql({
-          code: 401,
-          error: 'NotAuthorized',
-          message: t('guildQuestsNotSupported'),
-        });
-    });
-
-    it('returns an error when quest is not found', async () => {
-      let questKey = 'fakeQuestName';
-
-      await expect(member.post(`/groups/${questingGroup._id}/quests/reject/${questKey}`))
-        .to.eventually.be.rejected.and.eql({
-          code: 404,
-          error: 'NotFound',
-          message: t('questNotFound', { key: questKey }),
-        });
-    });
-
-    it('returns an error when user is not on the quest', async () => {
-      await expect(member.post(`/groups/${questingGroup._id}/quests/reject/${PET_QUEST}`))
-        .to.eventually.be.rejected.and.eql({
-          code: 401,
-          error: 'NotAuthorized',
-          message: t('questNotOwned'),
-        });
-    });
-
     it('returns an error when group is not on a quest', async () => {
       await member.update(userQuestUpdate);
 
-      await expect(member.post(`/groups/${questingGroup._id}/quests/reject/${PET_QUEST}`))
+      await expect(member.post(`/groups/${questingGroup._id}/quests/reject`))
         .to.eventually.be.rejected.and.eql({
           code: 404,
           error: 'NotFound',
@@ -88,11 +55,18 @@ describe('POST /groups/:groupId/quests/invite/:questKey', () => {
       await member.update(userQuestUpdate);
       await questingGroup.update({'quest.key': PET_QUEST});
 
-      await member.post(`/groups/${questingGroup._id}/quests/reject/${PET_QUEST}`);
+      let questMembers = {};
+      questMembers[member._id] = true;
+      await questingGroup.update({'quest.members': questMembers});
 
+      let rejectResult = await member.post(`/groups/${questingGroup._id}/quests/reject`);
       let userWithRejectInvitation = await member.get('/user');
+      let updatedGroup = await member.get(`/groups/${questingGroup._id}`);
+
       expect(userWithRejectInvitation.party.quest.key).to.be.null;
       expect(userWithRejectInvitation.party.quest.RSVPNeeded).to.be.false;
+      expect(updatedGroup.quest.members[member._id]).to.be.false;
+      expect(updatedGroup.quest).to.deep.equal(rejectResult);
     });
   });
 });
--- a/test/api/v3/integration/quests/POST-groups_groupid_quests_reject.test.js
+++ b/test/api/v3/integration/quests/POST-groups_groupid_quests_reject.test.js
@@ -1,0 +1,98 @@
+import {
+  createAndPopulateGroup,
+  translate as t,
+} from '../../../../helpers/api-v3-integration.helper';
+import { v4 as generateUUID } from 'uuid';
+
+describe('POST /groups/:groupId/quests/invite/:questKey', () => {
+  let questingGroup;
+  let member;
+  const PET_QUEST = 'whale';
+  let userQuestUpdate = {
+    items: {
+      quests: {},
+    },
+    'party.quest.RSVPNeeded': true,
+    'party.quest.key': PET_QUEST,
+  };
+
+  before(async () => {
+    let { group, members } = await createAndPopulateGroup({
+      groupDetails: { type: 'party', privacy: 'private' },
+      members: 1,
+    });
+
+    questingGroup = group;
+    member = members[0];
+
+    userQuestUpdate.items.quests[PET_QUEST] = 1;
+  });
+
+  context('failure conditions', () => {
+    it('returns an error when group is not found', async () => {
+      await expect(member.post(`/groups/${generateUUID()}/quests/reject/${PET_QUEST}`))
+        .to.eventually.be.rejected.and.eql({
+          code: 404,
+          error: 'NotFound',
+          message: t('groupNotFound'),
+        });
+    });
+
+    it('returns an error when group is not a party', async () => {
+      let { group, groupLeader } = await createAndPopulateGroup({
+        groupDetails: { type: 'guild', privacy: 'private' },
+      });
+
+      await expect(groupLeader.post(`/groups/${group._id}/quests/reject/${PET_QUEST}`))
+        .to.eventually.be.rejected.and.eql({
+          code: 401,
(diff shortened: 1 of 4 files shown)


website/src/controllers/api-v3/groups.js
```diff
@@ -18,6 +18,7 @@ import * as firebase from '../../libs/api-v3/firebase';
 import { sendTxn as sendTxnEmail } from '../../libs/api-v3/email';
 import { encrypt } from '../../libs/api-v3/encryption';
 import { quests as questScrolls } from '../../../../common/script/content';
+import { mockAnalyticsService } from '../../libs/api-v3/analyticsService';
```
crookedneighbor: Why are we importing a mockAnalyticsService?

website/src/controllers/api-v3/groups.js
```diff
@@ -655,4 +656,58 @@ api.inviteToQuest = {
...
+ * @apiGroup Group
+ *
+ * @apiParam {string} groupId The group _id (or 'party')
+ * @apiParam {string} questKey The quest _id
+ *
+ * @apiSuccess {Object} Quest Object
+ */
+api.rejectQuest = {
+  method: 'POST',
+  url: '/groups/:groupId/quests/reject/:questKey',
```
crookedneighbor: Specifying `questKey` should not be here. There can only be one quest at a time.

website/src/controllers/api-v3/groups.js
```diff
@@ -655,4 +656,58 @@ api.inviteToQuest = {
...
+    req.checkParams('groupId', res.t('groupIdRequired')).notEmpty();
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let group = await Group.getGroup({user, groupId: req.params.groupId, fields: 'type quest'});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+    if (group.type !== 'party') throw new NotAuthorized(res.t('guildQuestsNotSupported'));
+    if (!quest) throw new NotFound(res.t('questNotFound', { key: questKey }));
+    if (!user.items.quests[questKey]) throw new NotAuthorized(res.t('questNotOwned'));
```
crookedneighbor: Why does it matter if the user rejecting the quest owns it?

website/src/controllers/api-v3/groups.js
```diff
@@ -655,4 +656,58 @@ api.inviteToQuest = {
...
+    let quest = questScrolls[questKey];
+
+    req.checkParams('groupId', res.t('groupIdRequired')).notEmpty();
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let group = await Group.getGroup({user, groupId: req.params.groupId, fields: 'type quest'});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+    if (group.type !== 'party') throw new NotAuthorized(res.t('guildQuestsNotSupported'));
```
crookedneighbor: Not sure this is necessary... It'll bomb out a few lines down from here when the group quest check happens.

website/src/controllers/api-v3/groups.js
```diff
@@ -655,4 +656,58 @@ api.inviteToQuest = {
...
+      category: 'behavior',
+      owner: false,
+      response: 'reject',
+      gaLabel: 'reject',
+      questName: group.quest.key,
+      uuid: user._id,
+    };
+    mockAnalyticsService.track('quest', analyticsData);
+
+    // @TODO: Are we tracking members this way?
```
crookedneighbor: I think this is something we should discuss. Right now we have:

```
true - user is participating
false - user is not participating
null / undefined - user has not chosen yet
```

Is that a pattern we want to continue in v3? It make a kind of sense.

website/src/controllers/api-v3/groups.js
```diff
@@ -655,4 +656,58 @@ api.inviteToQuest = {
...
+    // group.quest.members[user._id] = false;
+
+    user.party.quest.RSVPNeeded = false;
+    user.party.quest.key = null;
+    await user.save();
+
+    // questStart(req,res,next);
+
+    res.respond(200, {});
```
crookedneighbor: This should return the group quest object.

website/src/controllers/api-v3/groups.js
paglias: I'd say to keep tracking them this way

crookedneighbor: :+1:

website/src/controllers/api-v3/groups.js
TheHollidayInn: These is the invitation object, I believe. So, the language may be off, but I am pretty sure this is checking to make sure the user has an invitation.

website/src/controllers/api-v3/groups.js
TheHollidayInn: Make sense to me. There isn't a performance increase in stopping here either.

website/src/controllers/api-v3/groups.js
TheHollidayInn: Sounds good!

website/src/controllers/api-v3/groups.js
TheHollidayInn: Awesome. I'll add that in.

website/src/controllers/api-v3/groups.js
TheHollidayInn: That comments said to use this in a non production environment. I was thinking that that this was replaced somehow in production. Maybe I just misread that.

website/src/controllers/api-v3/groups.js
TheHollidayInn: Yea, just checked. I'm not sure that we need to check for this. I ported it over thinking that it was needed in apiv2, but I'm not sure we need to keep it.

TheHollidayInn:
Okay, updated! And, I fixed the merge conflict. Just waiting on Travis, as usual.

paglias:
Looks good :+1: Waiting for startQuest

website/src/controllers/api-v3/groups.js
```diff
@@ -616,4 +618,96 @@ api.inviteToGroup = {
...
+ * @api {post} /groups/:groupId/quests/invite Invite users to a quest
+ * @apiVersion 3.0.0
+ * @apiName InviteToQuest
+ * @apiGroup Group
+ *
+ * @apiParam {string} groupId The group _id (or 'party')
+ *
+ * @apiSuccess {Object} Quest Object
+ */
+api.inviteToQuest = {
```
crookedneighbor: This route was moved to a separate quests.js file. You should remove these lines and add your reject route to it.

website/src/controllers/api-v3/groups.js
```diff
@@ -616,4 +618,96 @@ api.inviteToGroup = {
...
+    let quest = questScrolls[questKey];
+
+    req.checkParams('groupId', res.t('groupIdRequired')).notEmpty();
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let group = await Group.getGroup({user, groupId: req.params.groupId, fields: 'type quest'});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+    if (!quest) throw new NotFound(res.t('questNotFound', { key: questKey }));
```
crookedneighbor: Why this check? Does it matter what is on `user.party.quest.key`? The `group.quest.key` is the source of truth here.

test/api/v3/integration/quests/POST-groups_groupid_quests_reject.test.js
```diff
@@ -0,0 +1,81 @@
...
+      await questingGroup.update({'quest.members': questMembers});
+
+      let rejectResult = await member.post(`/groups/${questingGroup._id}/quests/reject`);
+      let userWithRejectInvitation = await member.get('/user');
+      let updatedGroup = await member.get(`/groups/${questingGroup._id}`);
+
+      expect(userWithRejectInvitation.party.quest.key).to.be.null;
+      expect(userWithRejectInvitation.party.quest.RSVPNeeded).to.be.false;
+      expect(updatedGroup.quest.members[member._id]).to.be.false;
+      expect(updatedGroup.quest).to.deep.equal(rejectResult);
```
crookedneighbor, line 69: :+1:

crookedneighbor:
Aside from my two remaining comments, I think this looks good.

website/src/controllers/api-v3/groups.js
TheHollidayInn: This check is to make sure the quest exists: `let quest = questScrolls[questKey];`

website/src/controllers/api-v3/groups.js
TheHollidayInn: Awesome, can do.

website/src/controllers/api-v3/quests.js
```diff
@@ -65,4 +65,58 @@ api.inviteToQuest = {
...
+    let quest = questScrolls[questKey];
+
+    req.checkParams('groupId', res.t('groupIdRequired')).notEmpty();
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let group = await Group.getGroup({user, groupId: req.params.groupId, fields: 'type quest'});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+    if (!quest) throw new NotFound(res.t('questNotFound', { key: questKey }));
```
crookedneighbor: Isn't it already validated in the invite route?

TheHollidayInn: Sure. I was thinking there was some added security with this, but I don't really see that being the case. I think it is good to remove. Sound good to you?

crookedneighbor: Yes. Please remove it. If the party quest object is being modified with an invalid key, we have bigger problems and I'd prefer it to blow up so we can track it down.

website/src/controllers/api-v3/quests.js
```diff
@@ -65,4 +65,58 @@ api.inviteToQuest = {
...
+    };
+    track('quest', analyticsData);
+
+    group.quest.members[user._id] = false;
+    group.markModified('quest.members');
+    await group.save();
+
+    user.party.quest.RSVPNeeded = false;
+    user.party.quest.key = null;
+    await user.save();
```
crookedneighbor: The user and group save calls should probably be done together, since they can be parallelized.

``` js
let [savedGroup] = await Q.all([
  group.save(),
  user.save(),
]);

res.respond(200, savedGroup.quest);
```

TheHollidayInn: Good call.

TheHollidayInn:
Updated!

website/src/controllers/api-v3/quests.js
```diff
@@ -65,4 +67,58 @@ api.inviteToQuest = {
...
+      questName: group.quest.key,
+      uuid: user._id,
+    };
+    track('quest', analyticsData);
+
+    group.quest.members[user._id] = false;
+    group.markModified('quest.members');
+
+    user.party.quest.RSVPNeeded = false;
+    user.party.quest.key = null;
```
paglias, line 111: question: here we should use Group.cleanQuestProgress?

paglias:
Merging manually

paglias:
updated version at #6693

PR #6648, opened 2016-02-03
inventory sorting

querry43:
Adds inventory sorting buttons.

I'm not sure what makes sense for locales or testing.  Does it make sense to add "constitutionShort": "Cons" to en/character.json?

I tried to add spec/controller tests but wasn't sure how to run them.  I am in a docker container and things go awry when the e2e test tries to start a framebuffer.

user id: bf56a7cf-0970-40f2-af1b-f4df3f9bda4d

### DIFF
--- a/test/spec/controllers/sortableInventoryCtrlSpec.js
+++ b/test/spec/controllers/sortableInventoryCtrlSpec.js
@@ -1,0 +1,32 @@
+'use strict';
+
+describe('Sortable Inventory Controller', function() {
+  var scope;
+
+  beforeEach(inject(function($rootScope, $controller) {
+    scope = $rootScope.$new();
+    $controller('SortableInventoryController', {$scope: scope});
+  }));
+
+  it('has no default sort order', function(){
+    expect(scope.order).to.not.exist;
+  });
+
+  it('sets sort criteria for all standard attributes', function(){
+    var oldOrder = scope.order;
+
+    var attrs = ['con', 'int', 'per', 'str', 'name', 'set'];
+
+    attrs.forEach(function (attribute) {
+      scope.setOrder(attribute);
+      expect(scope.order).to.exist;
+      expect(scope.order).to.not.eql(oldOrder);
+      oldOrder = scope.order;
+    });
+  });
+
+  it('does nothing when missing sort criteria', function(){
+    scope.setOrder('foooo');
+    expect(scope.order).to.not.exist;
+  });
+});
--- a/website/public/js/controllers/sortableInventoryCtrl.js
+++ b/website/public/js/controllers/sortableInventoryCtrl.js
@@ -1,0 +1,19 @@
+habitrpg.controller('SortableInventoryController', ['$scope',
+  function ($scope) {
+
+    // doing this in a function causes 10 $digest iteration error
+    attributeSort = {
+      con: ['-con', '-(con+int+per+str)'],
+      int: ['-int', '-(con+int+per+str)'],
+      per: ['-per', '-(con+int+per+str)'],
+      str: ['-str', '-(con+int+per+str)'],
+      name: 'text()',
+      set: 'set'
+    }
+
+    $scope.setOrder = function(order) {
+      if (order in attributeSort) {
+        $scope.order = attributeSort[order];
+      }
+    };
+}]);
--- a/website/views/options/inventory/equipment.jade
+++ b/website/views/options/inventory/equipment.jade
@@ -17,8 +17,14 @@
 
       li.customize-menu.inventory-gear
         menu.pets-menu(label='{{::label}}', ng-show='gear[klass]',
-          ng-repeat='(klass,label) in {warrior:env.t("warrior"), wizard:env.t("mage"), rogue:env.t("rogue"), healer:env.t("healer"), special:env.t("special"), mystery:env.t("mystery"), armoire:env.t("armoireText")}')
-          div(ng-repeat='item in gear[klass]')
+          ng-repeat='(klass,label) in {warrior:env.t("warrior"), wizard:env.t("mage"), rogue:env.t("rogue"), healer:env.t("healer"), special:env.t("special"), mystery:env.t("mystery"), armoire:env.t("armoireText")}', ng-controller="SortableInventoryController")
+          span(class="sort-menu") {{env.t('orderBy')}}
+            span.btn-group.btn-group-xs
+              button.btn.btn-default(ng-click='setOrder("name")')=env.t('name')
+              button.btn.btn-default(ng-if='klass == "armoire"' ng-click='setOrder("set")')=env.t('set')
+            span.btn-group.btn-group-xs
+              button.btn.btn-default(ng-repeat='attr in ["con", "int", "per", "str"]', ng-click='setOrder(attr)') {{env.t(attr)}}
+          div(ng-repeat='item in gear[klass] | orderBy:order')
             button.customize-option(, class='shop_{{::item.key}}',
               ng-class='{selectableInventory: user.items.gear.equipped[item.type] == item.key}',
               ng-click='equip(item.key, "equipped")',
@@ -50,8 +56,14 @@
 
       li.customize-menu(ng-if='user.preferences.costume')
         menu.pets-menu(label='{{::label}}', ng-show='gear[klass]',
-          ng-repeat='(klass,label) in {warrior:env.t("warrior"), wizard:env.t("mage"), rogue:env.t("rogue"), healer:env.t("healer"), special:env.t("special"), mystery:env.t("mystery"), armoire:env.t("armoireText")}')
-          div(ng-repeat='item in gear[klass]')
+          ng-repeat='(klass,label) in {warrior:env.t("warrior"), wizard:env.t("mage"), rogue:env.t("rogue"), healer:env.t("healer"), special:env.t("special"), mystery:env.t("mystery"), armoire:env.t("armoireText")}', ng-controller="SortableInventoryController")
+          span(class="sort-menu") {{env.t('orderBy')}}
+            span.btn-group.btn-group-xs
+              button.btn.btn-default(ng-click='setOrder("name")')=env.t('name')
+              button.btn.btn-default(ng-if='klass == "armoire"' ng-click='setOrder("set")')=env.t('set')
+            span.btn-group.btn-group-xs
+              button.btn.btn-default(ng-repeat='attr in ["con", "int", "per", "str"]', ng-click='setOrder(attr)') {{env.t(attr)}}
+          div(ng-repeat='item in gear[klass] | orderBy:order')
             button.customize-option(class='shop_{{::item.key}}',
               ng-class='{selectableInventory: user.items.gear.costume[item.type] == item.key}',
               ng-click='equip(item.key, "costume")',
--- a/common/locales/en/gear.json
+++ b/common/locales/en/gear.json
@@ -1,4 +1,6 @@
 {
+  "set": "Set",
+
   "weapon": "weapon",
 
   "weaponBase0Text": "No Weapon",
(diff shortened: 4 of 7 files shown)


querry43:
screencap

Alys:
Can you write some words about how the sorting works, from the user's perspective?

For example, does selecting "INT" put all the gear that gives INT is at the top and the other gear is under it? Is there any sorting by the amount of INT it gives?

How does the "set" option sort the Armoire items and the non-Armoire items? E.g., do you see Armoire items that are part of a set at the top, then Armoire items that are not part of a set under that, and then all non-Armoire items under that?

The "Set" button should be at the end of the row of buttons so that all the attribute buttons are together.

For the attribute abbreviations, I think it's normal to have them in all upper case (so INT rather than Int) - at least that's how it's done on the wiki. I can't find an example on the site at the moment. 

It is important to allow the buttons to be translated. "constitutionShort" seems good, if there aren't already short versions of the attributes in the locales files.

SabreCat:
On the stats page, you see the attribute abbreviations in the area where the benefit from each piece of equipped gear is displayed. They are all-caps, but I'm not sure about translation status.

querry43:
Thanks for the feedback.

Here are the current sorting rules and I would appreciate feedback on improving them:
1. Each gear[klass] has its own sorting buttons and sorting state including costume classes
2. The attributes sort descending when selected
3. There is no default sort (current behavior)
4. The armoire also sorts by set which sorts by item[set] descending, it does not sort within a set (ex "1 of 3" vs "2 of 3").

Perhaps it makes sense to sort by the item name in addition to the primary sort criteria?

Maybe it also makes sense to sort by a class primary attribute by default?

All caps makes sense to me and I can add them to locales along with the stats page.  Separating the set button makes sense too.  I will add a new commit with this and other changes as we tease them out.

SabreCat:
Gave this a quick try. A couple of thoughts:
- How does it look with a small "Sort by" label in front of the row of buttons? When I first looked at it, I kinda expected it to be a filter rather than a sort. If it looks bad with a label, it may not be necessary (it becomes clear enough after a moment's experimentation), but it's worth a try.
- When sorting by an attribute, I'd suggest a secondary sort by total plusses across all attributes, descending. For instance, at present, if I sort by Constitution bonus, the Horned Iron Helm (9 CON 7 STR) sorts ahead of the Orange Cat Hat (9 CON 9 STR), though the latter is superior.
- I'm just as likely to want to sort by item slot (head, armor, etc.) as by attribute. That might be a nice option!

querry43:
@SabreCat, I like your ideas.  The first two bullets are in the just now commit, but I'm not sure there's a data source for slot information.  I could iterate through each slot at test inclusion but that seems inefficient.  Thoughts?

querry43:
Here are the new sorting rules:
1. Each gear[klass] has its own sorting buttons and sorting state including costume classes
2. The attributes sort descending when selected, then by sum of all attributes
3. There is no default sort (current behavior)
4. The armoire also sorts by set which sorts by item[set] descending, it does not sort within a set (ex "1 of 3" vs "2 of 3")
5. Name sorts by .text() ascending

SabreCat:
Upon further thought, item slot would do better as a filter or a grouping than a sort, so we can leave that aside for now!

querry43:
@SabreCat , @Alys , I have touched on each of the items you have mentioned in the most recent commit (to be squashed after review.)  I don't really like the duplication in equipment.jade.  Is there some prior art for how you like to manage this?  Partials vs directives with templates?

I am also short on tests.  Thoughts?

querry43:
I finally got around to some testing for the new controller.  The commits have been squashed and all current behavior is documented in the commit message.

capture

website/public/js/controllers/sortableInventoryCtrl.js
```diff
@@ -0,0 +1,14 @@
+habitrpg.controller('SortableInventoryController', ['$scope',
+  function ($scope) {
+
+    // doing this in a function causes 10 $digest iteration error
+    $scope.attributeSort = {};
+    $scope.attributeSort[env.t('con')] = ['-con', '-(con+int+per+str)'];
```
crookedneighbor: I still don't understand why this needs to be using the translated versions of the keys. I'd think that would be a wonky way to assign keys to an object since we can't fully control what those values are. IE, if someone made a mistake in a translation and used the same word for `con` as `int`, this would break.

website/views/options/inventory/equipment.jade
```diff
@@ -50,8 +56,14 @@
 
       li.customize-menu(ng-if='user.preferences.costume')
         menu.pets-menu(label='{{::label}}', ng-show='gear[klass]',
-          ng-repeat='(klass,label) in {warrior:env.t("warrior"), wizard:env.t("mage"), rogue:env.t("rogue"), healer:env.t("healer"), special:env.t("special"), mystery:env.t("mystery"), armoire:env.t("armoireText")}')
-          div(ng-repeat='item in gear[klass]')
+          ng-repeat='(klass,label) in {warrior:env.t("warrior"), wizard:env.t("mage"), rogue:env.t("rogue"), healer:env.t("healer"), special:env.t("special"), mystery:env.t("mystery"), armoire:env.t("armoireText")}', ng-controller="SortableInventoryController")
+          span(class="sort-menu") {{env.t('orderBy')}}
+            span.btn-group.btn-group-xs
+              button.btn.btn-default(ng-click='setOrder("text()")') {{env.t('name')}}
```
crookedneighbor: Does this need to use angular to set the name? Would probably be more performant to do:

```
button.btn.btn-default(ng-click='setOrder("text()")')=env.t('name')
```

website/views/options/inventory/equipment.jade
```diff
@@ -50,8 +56,14 @@
...
+              button.btn.btn-default(ng-if='klass == "armoire"' ng-click='setOrder("set")') {{env.t('set')}}
```
crookedneighbor: Same as the comment above:

``` jade
button.btn.btn-default(ng-if='klass == "armoire"' ng-click='setOrder("set")')=env.t('set')
```

crookedneighbor: Let's use `===` for the comparison.

test/spec/controllers/sortableInventoryCtrlSpec.js
```diff
@@ -0,0 +1,26 @@
...
+  it('has no default sort order', function(){
+    expect(scope.order).to.not.exist;
+  });
+
+  it('sets sort order', function(){
+    scope.setOrder('foooo');
+    expect(scope.order).to.eql('foooo');
+  });
+
+  it('sorts by all attributes', function(){
```
crookedneighbor: Not sure that this test is sufficient for the behavior.

website/public/js/controllers/sortableInventoryCtrl.js
```diff
@@ -0,0 +1,14 @@
...
+    $scope.attributeSort[env.t('int')] = ['-int', '-(con+int+per+str)'];
+    $scope.attributeSort[env.t('per')] = ['-per', '-(con+int+per+str)'];
+    $scope.attributeSort[env.t('str')] = ['-str', '-(con+int+per+str)'];
+
+    $scope.setOrder = function(order) {
+      $scope.order = order
```
crookedneighbor: Missing a semi-colon

querry43:
Thanks for the fantastic feedback, @crookedneighbor .  Regarding tests, could you give me an idea of what you have in mind?  It seems that the functionality in equipment.jade might need some testing but I am not certain how to approach that.  Everything in this stack is new (I mean everything) to me so a few pointers or examples would be very helpful.

crookedneighbor:
I don't think we need tests for logic in the jade files, especially our e2e browser tests are not in a great state right now. (Hangs after logging in) Once we figure out what the deal is with our e2e tests, we can put together a plan to do some browser testing. For now, I don't think it's entirely necessary. 

I don't have much else in mind for the tests, but just testing that an attribute exists on an object doesn't tell me much. I'll have to think about it a little to see if I can think of a more meaningful assertion.

website/public/js/controllers/sortableInventoryCtrl.js
querry43: Yea, you're right. I have also been thinking about testing the jade file and I think that there's just too much logic in there.  The sort key => sort order is split between the controller and view.  I should probably move all of that to the controller where it can be more easily tested.

querry43:
@crookedneighbor, great review.  This was super helpful.  I think I have touched on everything in the new commit including tests.

test/spec/controllers/sortableInventoryCtrlSpec.js
```diff
@@ -0,0 +1,32 @@
...
+  it('has no default sort order', function(){
+    expect(scope.order).to.not.exist;
+  });
+
+  it('sets sort criteria for all standard attributes', function(){
+    var oldOrder = scope.order;
+
+    var attrs = ['con', 'int', 'per', 'str', 'name', 'set'];
+
+    for (var i = 0; i < attrs.length; i++) {
```
crookedneighbor: How about

``` js
attrs.forEach(function (attribute) {
  scope.setOrder(attribute);
  expect(scope.order).to.exist;
  expect(scope.order).to.not.eql(oldOrder);
  oldOrder = scope.order;
});
```

website/public/js/controllers/sortableInventoryCtrl.js
```diff
@@ -0,0 +1,19 @@
+habitrpg.controller('SortableInventoryController', ['$scope',
+  function ($scope) {
+
+    // doing this in a function causes 10 $digest iteration error
+    attributeSort = {
+      con:['-con', '-(con+int+per+str)'],
```
crookedneighbor: Could you include a space between the `:` and the value?

crookedneighbor:
Overall, it's looking good. Left a few more nitpicky comments. Will try to check this out locally and double check things this week.

querry43:
Thanks and good comments. Do we have a style guide for things like spacing?

querry43:
@crookedneighbor, I followed your recommendations and then squashed all of the commits together.

querry43:
@crookedneighbor , @Alys , @SabreCat, I think you all have had a look at this.  AFAIK, there is no remaining feedback to touch on.  Is there anything I can do to help this along?  I would love to start sorting stuff.

website/public/js/controllers/sortableInventoryCtrl.js
```diff
@@ -0,0 +1,19 @@
+habitrpg.controller('SortableInventoryController', ['$scope',
+  function ($scope) {
+
+    // doing this in a function causes 10 $digest iteration error
+    attributeSort = {
```
crookedneighbor, line 5: Can we add a `var` here so this isn't on the global namespace?

website/public/js/controllers/sortableInventoryCtrl.js
```diff
@@ -0,0 +1,19 @@
...
+      con: ['-con', '-(con+int+per+str)'],
+      int: ['-int', '-(con+int+per+str)'],
+      per: ['-per', '-(con+int+per+str)'],
+      str: ['-str', '-(con+int+per+str)'],
+      name: 'text()',
+      set: 'set'
+    }
+
+    $scope.setOrder = function(order) {
+      if (order in attributeSort) {
```
crookedneighbor, line 15: TIL you can use `in` outside of a `for in` loop.

website/public/js/controllers/sortableInventoryCtrl.js
```diff
@@ -0,0 +1,19 @@
+habitrpg.controller('SortableInventoryController', ['$scope',
+  function ($scope) {
+
+    // doing this in a function causes 10 $digest iteration error
```
crookedneighbor, line 4: I don't think we need this comment. Declaring this object outside of the function has better performance implications by nature.

crookedneighbor:
Aside from some nitpicky comments I just left, I think this looks good. I'm deploying it to one of our staging sties so that the non-dev staff members can try it out.

crookedneighbor:
Rebased and merged in.

crookedneighbor:
This will count towards your next contributor tier.

crookedneighbor:
We had to revert this because the call to `text()` on name option over and over again was causing the site to slow to a crawl. I've removed the name option and we're making sure that all is well and will redeploy.

querry43:
@crookedneighbor oh no!  Bummer.  Sorry about that and thanks for taking care of it.

PR #6654, opened 2016-02-03
Added quest cancel route and initial tests

TheHollidayInn:
Here is the quest cancel route and initial tests for #6509 .

I'm unsure about the return object, and we have started a discussion in Slack as to whether or not we should verify that a user is a group or quest leader before cancelling.

### DIFF
--- a/website/src/controllers/api-v3/quests.js
+++ b/website/src/controllers/api-v3/quests.js
@@ -4,6 +4,9 @@ import {
   model as Group,
 } from '../../models/group';
 import {
+  model as User,
+} from '../../models/user';
+import {
   NotFound,
   NotAuthorized,
 } from '../../libs/api-v3/errors';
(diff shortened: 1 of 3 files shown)


TheHollidayInn:
Eh, don't review this yet. I have a few updates.

TheHollidayInn:
Okay, updated!

website/src/controllers/api-v3/quests.js
```diff
@@ -65,4 +68,50 @@ api.inviteToQuest = {
...
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let group = await Group.getGroup({user, groupId, fields: 'type quest'});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    if (group.quest.active) throw new NotAuthorized(res.t('cantCancelActiveQuest'));
+
+    group.quest = {key: null, progress: {}, leader: null, members: {}};
```
paglias, line 102: we may want a group.resetQuestObject method

paglias:
Updated version #6694

PR #6655, opened 2016-02-03
Added quest leave route and initial tests

TheHollidayInn:
This the the quest leave route and initial tests for #6509.

### DIFF
--- a/website/src/controllers/api-v3/quests.js
+++ b/website/src/controllers/api-v3/quests.js
@@ -8,6 +8,7 @@ import {
   NotAuthorized,
 } from '../../libs/api-v3/errors';
 import { quests as questScrolls } from '../../../../common/script/content';
+import Q from 'q';
 
 let api = {};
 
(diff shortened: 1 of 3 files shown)


website/src/controllers/api-v3/quests.js
```diff
@@ -65,4 +65,55 @@ api.inviteToQuest = {
...
+    group.quest.members[user._id] = false;
+    group.markModified('quest.members');
+
+    user.party.quest = Group.cleanQuestProgress();
+    user.markModified('party.quest');
+
+    await group.save();
+    await user.save();
+
+    res.respond(200, {});
```
crookedneighbor: This should return the group quest object.

TheHollidayInn: Sounds good. I figured that might be the case.

TheHollidayInn:
Updated!

paglias:
Updated version at #6695

PR #6683, opened 2016-02-09
Fixes #6356 (gryphon avatar clipping)

### LINKED ISSUE #6356: Avatar clipping in the wrong spots
Not sure if this is the right place for this issue, but it's just something that I noticed on Habitica.com
http://i.imgur.com/V5HoHAf.png
This was noted with the "Absurd Party Robes" used as part of the costume on the "Royal Purple Gryphon" mount.

User ID: 17bbc7b9-95bf-48b1-9bab-c550c2084c0b
OS: Windows 7 Pro
Browser: Firefox 42

taldin:
Updated gryphon mount graphics.

One odd problem popping up now that there are monkeys in there:
image

More than happy to fix this if you tell me how.


Alys:
I had thought that the number of spritesheets was determined automatically to prevent this (`calculateSpritesheetsSrcIndicies` in `tasks/gulp-sprites.js`) but yeah.. that doesn't seem to be working.

@taldin I think all the code should be in `tasks/gulp-sprites.js` so if you'd like to look at it, please do! 

If you end up changing anything in that file, please also change `Verifiying` to `Verifying` (that tripped me up when searching for the error message).

taldin:
Hmm.  The code isn't well commented, so I'd be guessing what needs to be done to fix that is to tinker with the code here: 
 if (totalPixels > MAX_SPRITESHEET_SIZE) {
      slices.push(index - 1);
      totalPixels = imageSize;
    }

That doesn't seem right to me.  I don't see how it would add a new spritesheet in there in the code -- is it supposed to add a new one automatically?

Alys:
I think all it's doing there is looping through all the individual images, and creating a new element in the `slices` array every time it identifies that a new spritesheet will be needed. The value of the element is the index of the individual image that starts the new spritesheet (so as a made-up example, 1, 25, 66, etc). The slices array is then used in `createSpritesStream()`, and `index` in that function is the index of the spritesheet (so, 0, 1, 2, etc). So I think all that is correct, and new spritesheets SHOULD be created when needed. 

It's as if the calculateImgDimensions() test is returning different results in the two places its used - in the test portion of that file, it measures the total size of each spritesheet image; in the creation portion, it adds up the sizes of the individual images. Somehow the total size does not match the added sizes.

I can't help wondering if the image minification is playing a role in this, although that's wild guesswork.

It might be helpful to put a `console.log(totalPixels)` just after `if (totalPixels > MAX_SPRITESHEET_SIZE) {` and `console.log(spriteSize)` just after `let spriteSize = calculateImgDimensions(img);`.

But I'm clutching at straws. :)

crookedneighbor:
I should really just take out that console log message, it's not actually useful :(

Basically, it's just really hard to guess the dimensions that a spritesmith sheet ends up having. Sometimes it creates images that are slightly to large (and then warns about it), but it's within a margin of error where it shouldn't be noticeable on ios devices.

I say just ignore it for now. Merge and re-run the sprite task and this should be ready to merge.

taldin:
All right.  Do I need to do anything on my end, or is this an Alys thing?

Alys:
I've vagued up the size warning so that it no longer sounds like a showstopper. I'm fixing merge conflicts locally and if all's well, I'll merge and close this.

Alys:
Merged! Thank you @taldin! The Gryphon mount changes will go live when we do our next deployment to production, which is likely to be within the next few days.

PR #6684, opened 2016-02-09
Add button to buy 5 gems.

taldin:
Adds additional gem button to buy gems in batches of 5.
(Feel free to request an Artisan make a better button.  I won't be offended.) 
image
Deducts the correct amount of gold and adds 5 gems:
image
Error checks to see if you try and buy 5 when you can only buy 4:
image


SabreCat:
Ooh, interesting! Maybe the image could be a crisscross of Gems, like the five-spots on a die.

taldin:
Something like this work?
pet_currency_gem5

Alys:
I like that five-spots icon! Could you do another screenshot like the first one in this PR to show it in context?

taldin:
Looks like this now:
image

Nice unintended side effect that the buttons grey out independently when you have > 0 but < 5 gems remaining:
image

Alys:
@taldin Surely you mean INTENDED side effect? ;)

@lemoness Are you okay with this new button for buying 5 gems at once?

common/script/index.js
```diff
@@ -1152,6 +1152,53 @@ api.wrap = function(user, main) {
             message: "+1 Gem"
           }, _.pick(user, $w('stats balance'))) : void 0;
         }
+/* buy 5 gems */
+        if (type === 'gems5' && key === 'gem') {
```
crookedneighbor, line 1156: The whole `purchase` function is terrible. We never should have lumped all these things together... gold -> gem purchases should have been their own separate route instead of being tacked onto `purchase`

crookedneighbor, line 1156: Wouldn't it be better to adjust the gems logic to take in a quantity instead of having totally separate logic for gem and gem5?

crookedneighbor, line 1156: Also, we need test coverage around this before it can be merged in.

crookedneighbor:
What would be really awesome is if we could detect if someone was clicking rapidly on the gem icon, then have a modal popup asking if the user would like to buy multiple gems and input how many.

taldin:
@crookedneighbor: True enough, but my Jade coding skills aren't at the level where I can build modal dialog boxes (yet) -- this is just bolting a relatively easy fix for something that's bugged me for awhile and extending an existing algorithm.

lemoness:
I think this is very convenient! Good idea.

crookedneighbor:
I'd like the following to happen before this can be merged in:
- [ ] extract the purchase function into a separate module
- [ ] test module (ideally fully tested, but at the very least, the gem functionality)
- [ ] refactor purchase to take a quantity parameter defaulting at 1 if not provided (should only factor into gems for now)

I still think that a better interface would be a modal that pops up when the gem icon is clicked multiple times in quick succession, but we can go with a 5 gems button for now.

taldin:
I think I need to defer this to someone with more skills than I have, unless there's a handy 'this is how you write test code' out there somewhere that you recommend.   Also, by 'extract the purchase function', do you mean that you want a new function file, or just a separate function?'

Alys:
I feel uneasy about the idea of a modal that pops up when the gem icon is clicked multiple times in quick succession. It's not an obvious way of getting to the modal, so we'll have people who trigger it without realising how it happened and who will then get frustrated later when they can't find it again. It will also train them to click on things quickly, which we usually tell them not to. I realise that "quickly" here doesn't necessarily mean fast enough to cause sync errors but the users won't realise that unless it's explained in detail.

taldin:
Heck, I didn't even know there was anything special with a double click.   :)   All I'm doing is reading existing code, doing a bunch of google-fu searches, and modifying existing code.   I'm more than happy to learn things (I'm a kinesthetic, which means I learn by doing), but I get all teeth-gnashy when I'm making unprepared changes to someone else's code because I'm afraid of breaking stuff.

crookedneighbor:
Now that v3 of the api is out, the code has changed significantly. Going to close this for now, but if you want to bring it up to date, I'm happy to help you get it into a merge-able state.

taldin:
Sure will.  :)   I still need to figure out how to make tests, too.

PR #6688, opened 2016-02-10
start quest tests

crookedneighbor:
@paglias

### DIFF
--- a/website/src/models/group.js
+++ b/website/src/models/group.js
@@ -210,23 +210,16 @@ schema.methods.startQuest = async function startQuest (user) {
     user.markModified('party.quest');
   }
 
-  // Remove the quest from the quest leader items (if he's the current user)
+  // Remove the quest from the quest leader items (if they are the current user)
   if (this.quest.leader === user._id) {
     user.items.quests[this.quest.key] -= 1;
     user.markModified('items.quests');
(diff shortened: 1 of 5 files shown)


crookedneighbor:
Whoops, this is going to fail because of linting errors in the test file. Fixing.

website/src/models/group.js
```diff
@@ -210,23 +210,16 @@ schema.methods.startQuest = async function startQuest (user) {
     user.markModified('party.quest');
   }
 
-  // Remove the quest from the quest leader items (if he's the current user)
+  // Remove the quest from the quest leader items (if they are the current user)
```
paglias, line 213: so the leader is updated with all the other members and the quest is removed separately?

crookedneighbor, line 213: Yeah, I know that's probably not as performative, but I think it's worth it for code clarity and reducing duplication. Plus it removes the need to pop the quest leader of the array, and then put them back on after the second update call so that they can receive the quest started email. 

Do you have strong feelings about it?

paglias, line 213: No, absolutely. Just wanted to be sure of what it did. Also the number of updates is the same so the difference should be minimal. For me it looks good if you want to merge

PR #6694, opened 2016-02-11
Updated quest cancel

paglias:


paglias:
Updated version of #6654

PR #6733, opened 2016-02-18
Store data for next mobile update

phillipthelen:
This PR adds  storage for two new things:
- a list of reminders associated with a task.
- different areas in which the user wants to improve their life. This is chosen during mobile signup and can be used to p.e. suggest guilds.

**Reminders:** Currently the iOS app supports reminders for dailies and todos. Daily reminders are triggered on each day the daily is active at the time specified in the _time_ property. Todos also use the _startDate_ property to determine the day when the reminder should be displayed. Currently a reminder for a todo is only displayed once, but I plan on adding repeat options in the future (hence _startDate_). The intention of saving the reminders on the server is to 1. have them persistently between app and device installs and 2. eventually have the reminders on different plattforms too.

**Improvement Categories/areas** during the setup of the iOS and android app, the user has the option to select one or multiple categories they want to improve in. Currently this only creates default tasks based on this. The guilds feature that will be available in the next update for the android and iOS app will also feature guild suggestions for people that aren't in many guilds yet. In order to tailor the suggestions to the user, the choices made during setup have to be persisted. Having those choices available can later also be used to expand on such suggestions.

### DIFF
--- a/common/script/index.js
+++ b/common/script/index.js
@@ -815,9 +815,12 @@ api.wrap = function(user, main) {
             message: i18n.t('messageTaskNotFound', req.language)
           }) : void 0;
         }
-        _.merge(task, _.omit(req.body, ['checklist', 'id', 'type']));
+        _.merge(task, _.omit(req.body, ['checklist', 'remidners', 'id', 'type']));
         if (req.body.checklist) {
           task.checklist = req.body.checklist;
+        }
+        if (req.body.reminders) {
+          task.reminders = req.body.reminders;
         }
         if (typeof task.markModified === "function") {
           task.markModified('tags');
--- a/common/script/index.js
+++ b/common/script/index.js
@@ -815,7 +815,7 @@ api.wrap = function(user, main) {
             message: i18n.t('messageTaskNotFound', req.language)
           }) : void 0;
         }
-        _.merge(task, _.omit(req.body, ['checklist', 'remidners', 'id', 'type']));
+        _.merge(task, _.omit(req.body, ['checklist', 'reminders', 'id', 'type']));
         if (req.body.checklist) {
           task.checklist = req.body.checklist;
         }
--- a/website/src/models/user.js
+++ b/website/src/models/user.js
@@ -395,7 +395,7 @@ var UserSchema = new Schema({
       type: Array,
       validate: (categories) => {
         const validCategories = ['work', 'exercise', 'healthWellness', 'school', 'teams', 'chores', 'creativity'];
-        let isValidCategory = categories.every(category => validValues.indexOf(category) !== -1);
+        let isValidCategory = categories.every(category => validCategories.indexOf(category) !== -1);
         return isValidCategory;
     }}
   },
--- a/website/src/models/user.js
+++ b/website/src/models/user.js
@@ -391,15 +391,15 @@ var UserSchema = new Schema({
       raisePet: {type: Boolean, 'default': false},
       streak: {type: Boolean, 'default': false}
     },
-    improvementCategories: {
-      "work": {type: Boolean, 'default': false},
-      "exercise": {type: Boolean, 'default': false},
-      "healthWellness": {type: Boolean, 'default': false},
-      "school": {type: Boolean, 'default': false},
-      "teams": {type: Boolean, 'default': false},
-      "chores": {type: Boolean, 'default': false},
-      "creativity": {type: Boolean, 'default': false},
-    }
+    improvementCategories: {type: Array, validate: function (categories) {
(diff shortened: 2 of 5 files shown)


common/script/index.js
```diff
@@ -815,10 +815,13 @@ api.wrap = function(user, main) {
             message: i18n.t('messageTaskNotFound', req.language)
           }) : void 0;
         }
-        _.merge(task, _.omit(req.body, ['checklist', 'id', 'type']));
+        _.merge(task, _.omit(req.body, ['checklist', 'remidners', 'id', 'type']));
```
paglias: typo?

website/src/models/user.js
```diff
@@ -390,6 +390,15 @@ var UserSchema = new Schema({
       hatchPet: {type: Boolean, 'default': false},
       raisePet: {type: Boolean, 'default': false},
       streak: {type: Boolean, 'default': false}
+    },
+    improvementCategories: {
```
paglias, line 394: can you use an array here where you store the categories instead of multiple fields?

phillipthelen, line 394: Jeah. Array of strings? What would the best way be to still restrict what strings can be inserted?

paglias, line 394:
```
improvementCategories: {type: Array, validate: function (categories) {
 var validValues = true;
 categories.forEach(function (cat) {
  if (['work', 'exercise', ...].indexOf(cat) === -1){
    validValues = false
  }
})

return validValues;
}}
```

something like ^

crookedneighbor, line 394: @vIiRuS What route will be used to update these categories? I don't think we should tack these onto the [User PUT route's acceptable paths](https://github.com/HabitRPG/habitrpg/blob/c95950c679c018a30f07d6430b3e9fda61fac11e/website/src/controllers/api-v2/user.js#L285-L299), and instead should have it's own route to update the properties.

phillipthelen, line 394: i had planned to use the user PUT route. But I can also create a new route for this if you think that is better. I just thought it would be a bit overkill.

paglias, line 394: I think that PUT route isn't that evil. If we start creating a route for all the boolean/string/number flags we'll never end

crookedneighbor, line 394: I forgot this was under the preferences object, which is _mostly_ fine to update via the put route.

The problem with the user PUT route is that the user has some protected attributes that should not be allowed to be changed (gem balance, for instance), and some that can be (preferences, for instance).

The tricky thing is that there are some preferences, like background choice, which need to be protected. 

It's probably fine to leave this on the put route, but I do think eventually (v3) we're going to want a separate preferences route and move anything off the preferences object that can't be freely updated.

website/src/models/user.js
```diff
@@ -390,7 +390,16 @@ var UserSchema = new Schema({
       hatchPet: {type: Boolean, 'default': false},
       raisePet: {type: Boolean, 'default': false},
       streak: {type: Boolean, 'default': false}
-    }
+    },
+    improvementCategories: {type: Array, validate: function (categories) {
```
crookedneighbor: Can we split this onto multiple lines? 

```
improvementCategories: {
  type: Array, 
  validate: function (categories) {
```

We could also use an arrow function since we don't care about `this`

```
validate: (categories) => {
```

website/src/models/user.js
```diff
@@ -390,7 +390,16 @@ var UserSchema = new Schema({
...
+      var validValues = true;
```
crookedneighbor: Can we do something like this with `every`:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

``` js
const validCategories = ['work', 'exercise', 'healthWellness', 'school', 'teams', 'chores', 'creativity'];

let isValidCategory = categories.every(category => validValues.indexOf(category) !== -1);

return isValidCategory;
```

With the `forEach`, it doesn't break early if there's an invalid category.

crookedneighbor:
If we just have a test around valid categories vs invalid categories, I think we're good to go.

website/src/models/user.js
```diff
@@ -390,7 +390,14 @@ var UserSchema = new Schema({
       hatchPet: {type: Boolean, 'default': false},
       raisePet: {type: Boolean, 'default': false},
       streak: {type: Boolean, 'default': false}
-    }
+    },
+    improvementCategories: {
```
SabreCat, line 394: This'll actually be nice for improved onboarding on the Web site later, too! :+1:

crookedneighbor:
I think this makes sense :+1:

PR #6735, opened 2016-02-19
[API v3] Data export

paglias:
WIP. Missing tests

@crookedneighbor this is just a refactor of the API v2 methods without any breaking change. So I think we're safe just swapping the old methods with these new ones. Question: these routes should be under /api/v3/export (with the old ones redirecting to them) or they should be top level routes directly under /export without the api namespace?

Also, I have no idea on how to tests a few methods that don't use api keys for authentication but instead authenticate with a cookie session.


website/src/controllers/api-v3/dataexport.js
```diff
@@ -0,0 +1,225 @@
...
+    let csvRes = await csvStringify(output);
+    res.status(200).send(csvRes);
+  },
+};
+
+// Convert user to json and attach tasks divided by type
+// at user.tasks[`${taskType}s`] (user.tasks.{dailys/habits/...})
+async function _getUserDataForExport (user) {
+  let userData = user.toJSON();
+  userData.tasks = {};
```
crookedneighbor, line 81: Do we want to attach everything to the user? Or return:

```
{
  user,
  dailys,
  habits,
  rewards,
  todos,
}
```

website/src/middlewares/api-v3/auth.js
```diff
@@ -41,7 +41,7 @@ export function authWithHeaders (optional = false) {
 // Authenticate a request through a valid session
 // TODO should use json web token
 export function authWithSession (req, res, next) {
-  let userId = req.session.userId;
+  let userId = req.session && req.session.userId;
```
crookedneighbor: Does our session middleware not guarantee that `req.session` exists?

package.json
```diff
@@ -29,6 +29,7 @@
     "firebase": "^2.2.9",
     "firebase-token-generator": "^2.0.0",
     "glob": "^4.3.5",
+    "got": "^6.1.1",
```
crookedneighbor, line 32: Do you want to switch everything over to use `got` eventually?

paglias, line 32: Yeah that was my idea. It just seems simpler and less bloated than request

website/src/middlewares/api-v3/auth.js
paglias: not entirely sure

website/src/controllers/api-v3/dataexport.js
paglias, line 81: I don't know, I thought attaching everything to user.tasks was a good idea but we can change it

package.json
paglias, line 32: and also it supports promises

paglias:
@crookedneighbor all the tests would pass, the only problem is that I don't know how to test authentication when authWithSession is used

Issue #6672, opened 2016-02-06
when buying block subscription through gem in header, perk amounts aren't shown


Alys:
When you buy a block subscription (three months or more) through the gem on the header, the "cap raised to" and "mystic hourglass" green texts don't contain values:

![screen shot 2016-02-07 at 6 16 57 am](https://cloud.githubusercontent.com/assets/1495809/12869147/74722fe6-cd62-11e5-80f0-7de56a5db09a.png)

When you buy through the "subscribe" button, they are displayed correctly:

![screen shot 2016-02-07 at 6 17 10 am](https://cloud.githubusercontent.com/assets/1495809/12869146/74715f76-cd62-11e5-8713-3d4c0bde095e.png)

kogelnikp:
@Alys I would like to work on a fix for this one.

Alys:
@kogelnikp Thank you! Please do!

PR #6736, opened 2016-02-19
Fixed perk amounts when buying subscription through gem

kogelnikp:
This change fixes issue #6672. 

The angular controller "SettingsCtrl" was missing in the modal buy-gems view. Therefore the gemGoldCap function couldn't be called. I just added the controller to this part of the view to get it work.

2016-02-19 12_39_19-fotos


Alys:
Sorry about the delay on this! This is good, thank you! 

What is your User ID (from Settings > API)?

kogelnikp:
User Id: 120bed2d-e59f-4adc-820b-9de529d24dc4

Alys:
Tier 1 awarded.

PR #6742, opened 2016-02-20
Display attribute info in Character Build

murraybd:
When I have unallocated attribute points I never know which attribute to allocate them to because I don't really know what the attributes do. I thought it would be useful to display information about the attributes in the "Character Build" section where one allocate points. Come to find out there is already detailed information about the attributes it just was off the screen in the Attribute section. I still think it make sense to have that information accessible right where you will be allocating the points though, rather than having to mouse over to the Attribute section and then back to the allocation part.

### DIFF
--- a/common/locales/en/character.json
+++ b/common/locales/en/character.json
@@ -136,7 +136,7 @@
     "respawn": "Respawn!",
     "youDied": "You Died!",
     "dieText": "You've lost a Level, all your Gold, and a random piece of Equipment. Arise, Habiteer, and try again! Curb those negative Habits, be vigilant in completion of Dailies, and hold death at arm's length with a Health Potion if you falter!",
-    "sureReset": "Are you sure? This will reset your character's class and allocated points (you'll get them all back to re-allocate), and costs 3 gems",
+    "sureReset": "Are you sure? This will reset your character's class and allocated points (you'll get them all back to re-allocate), and costs 3 gems.",
     "purchaseFor": "Purchase for <%= cost %> Gems?",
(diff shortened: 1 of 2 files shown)


Alys:
Can you please include screenshot(s) that show the changes you've made?

murraybd:
Here a couple.

constitution-hint
strength-hint

Alys:
That's a great idea! 

It looks like this might be a good opportunity to clean up the code and put all the repeated parts into a loop. Are you interested in doing that? Have a look at how it's done in this code:
https://github.com/HabitRPG/habitrpg/blob/develop/website/views/shared/profiles/stats/attributes.jade#L4-L10

murraybd:
Yeah, that makes sense. Here you go.

crookedneighbor:
:+1:

murraybd:
How long does it usually take to get things merged?

Alys:
It depends a lot on the PR, how much review and testing it needs, how many other PRs need to be reviewed and tested at the same time, how many other admin tasks need to be done.  Some of my PRs in the past went unreviewed for many weeks. :) These days though it's usually only a few days. We'll get to yours as soon as we can. Sorry about the delay!

Alys:
Thanks @murraybd , that's a nice little enhancement! What's your User ID (from Settings > API)?

murraybd:
Thanks for merging this! My User ID is: 5b93ce83-2508-4ea4-9647-079899dffa96

Alys:
Tier 2 awarded.

PR #6748, opened 2016-02-21
Added unsubscribe route and initial tests

TheHollidayInn:
Here is the updated unsubscribe route for Api V3.

### DIFF
--- a/website/src/controllers/api-v3/unsubscription.js
+++ b/website/src/controllers/api-v3/unsubscription.js
@@ -1,0 +1,57 @@
+import { model as User } from '../../models/user';
+import { model as EmailUnsubscription } from '../../models/emailUnsubscription';
+import { decrypt } from '../../libs/api-v3/encryption';
+import {
+  NotFound,
+} from '../../libs/api-v3/errors';
+
+let api = {};
+
+/**
+ * @api {post} /unsubscribe Unsubscribe an email or user from email notifications
+ * @apiVersion 3.0.0
+ * @apiName UnsubscribeEmail
+ * @apiGroup Unsubscribe
+ *
+ * @apiParam {String} code An unsubscription code
+ *
+ * @apiSuccess {String} okRes An message stating the user/email unsubscribed successfully
+ */
+api.unsubscribe = {
+  method: 'GET',
+  url: '/unsubscribe',
+  middlewares: [],
+  async handler (req, res) {
+    req.checkQuery({
+      code: {
+        notEmpty: {errorMessage: res.t('missingUnsubscriptionCode')},
+      },
+    });
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let data = JSON.parse(decrypt(req.query.code));
+
+    if (data._id) {
+      let userUpdated = await User.update(
+        {_id: data._id},
+        { $set: {'preferences.emailNotifications.unsubscribeFromAll': true}}
+      );
+
+      if (userUpdated.nModified !== 1) throw new NotFound(res.t('userNotFound'));
+
+      res.send(`<h1>${res.t('unsubscribedSuccessfully', null, req.language)}</h1> res.t('unsubscribedTextUsers', null, req.language)`);
+    } else {
+      let unsubscribedEmail = await EmailUnsubscription.findOne({email: data.email});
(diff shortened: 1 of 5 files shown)


crookedneighbor:
Are these really supposed to be gets?

crookedneighbor:
Not a post?

paglias:
It must be a get since it's used as a link in emails

website/src/controllers/api-v3/unsubscription.js
```diff
@@ -0,0 +1,57 @@
...
+
+    if (data._id) {
+      let userUpdated = await User.update(
+        {_id: data._id},
+        { $set: {'preferences.emailNotifications.unsubscribeFromAll': true}}
+      );
+
+      if (userUpdated.nModified !== 1) throw new NotFound(res.t('userNotFound'));
+
+      res.send(`<h1>${res.t('unsubscribedSuccessfully', null, req.language)}</h1> res.t('unsubscribedTextUsers', null, req.language)`);
```
paglias: passing the 2nd and 3rd parameters to res.t is not necessary

website/src/controllers/api-v3/unsubscription.js
```diff
@@ -0,0 +1,57 @@
...
+    } else {
+      let unsubscribedEmail = await EmailUnsubscription.findOne({email: data.email});
+      let okResponse = `<h1>${res.t('unsubscribedSuccessfully', null, req.language)}</h1> ${res.t('unsubscribedTextOthers', null, req.language)}`;
+      if (unsubscribedEmail) return res.send(okResponse);
```
paglias: I would write it a bit differently:

```
if (!unsubscribedEmail) await EmailUnsubscription.create({email: data.email});
res.send(okResponse);
```

website/src/controllers/api-v3/unsubscription.js
```diff
@@ -0,0 +1,57 @@
...
+ * @apiSuccess {String} okRes An message stating the user/email unsubscribed successfully
+ */
+api.unsubscribe = {
+  method: 'GET',
+  url: '/unsubscribe',
+  middlewares: [],
+  async handler (req, res) {
+    req.checkQuery({
+      code: {
+        notEmpty: {errorMessage: res.t('missingUnsubscriptionCode')},
```
paglias: Hmm we're not passing a translation for these errors anywhere else, should we start to?

website/src/controllers/api-v3/unsubscription.js
```diff
@@ -0,0 +1,57 @@
...
+ * @apiName UnsubscribeEmail
+ * @apiGroup Unsubscribe
+ *
+ * @apiParam {String} code An unsubscription code
+ *
+ * @apiSuccess {String} okRes An message stating the user/email unsubscribed successfully
+ */
+api.unsubscribe = {
+  method: 'GET',
+  url: '/unsubscribe',
```
paglias: @TheHollidayInn @crookedneighbor I don't like /unsubscribe, it's too generic. It should be something about emails. Do you have any idea?

paglias:
Except the comments it looks good!

crookedneighbor:
Ahh, that makes sense. The test file should be GET and not POST then.

website/src/controllers/api-v3/unsubscription.js
crookedneighbor: Agreed. I assumed this was about paid subscriptions. Could we do '/email/unsubscribe' instead?

paglias: Yeah it was pretty confusing. At this point I would rename the controller to emails.js and maybe have /emails/unsubscribe

website/src/controllers/api-v3/unsubscription.js
TheHollidayInn: It looks like the new auth.js controller does this in api.registerLocal. I would like to pass the translations, because the invalid parameters error seems to vague. Although, I can't get this to return a new message :P

website/src/controllers/api-v3/unsubscription.js
TheHollidayInn: I agree.

TheHollidayInn:
Updated!

TheHollidayInn:
Updated!

paglias:
@TheHollidayInn thanks! I'm merging it

PR #6760, opened 2016-02-23
Add button to add new empty checklist item

jeancl:
Enter button without explication isn't intuitive.

after and before


Alys:
I think the button should say "Add checklist item" to make it really obvious what it is for, and it's probably better at the bottom of the checklist, rather than at the top.

Could you also please put on-hover text on the button so that when you move your mouse over it to use it, you see a message explaining that hitting enter after typing an item will allow another item to be entered? I agree that it's not intuitive, but it's a useful shortcut and so we should have an explanation of it there. Something like "You can also add a new checklist item by hitting Enter after typing the previous item." Feel free to improve that.

Can you then show us screenshots that include more of the task edit screen so that we can see the new button in context? Also show us what it looks like with the hover text visible.
Please include English screenshots, although you're welcome to include Portuguese ones as well if you wish.  

Note that you'll need to edit `en/tasks.json` to add the new texts. You might also edit `pt/tasks.json` so that you can see the change in your own language, but please do not attach `pt/tasks.json` to this PR because translations are handled through Transifex.

jeancl:
Hello,

I made the suggested changes, but I can't change my location to show all in English.
This is the result with all the context:

changes

Alys:
@jeancl I just now loaded this into my local install to test it and the first thing I did almost by instinct was click the "Add checklist item" button a couple of times, to try to create a few blank items so that I could then add text to them all at once. :) It seemed to be a natural thing to do, but of course the button doesn't work until you've typed text into one blank item. I think we'll get other players trying to do the same thing, and we'll get bug reports from it because they won't realise that it's not supposed to work like that.

Is it possible to make the button greyed out until the user starts typing text?

jeancl:
hey @Alys here it is, take a look when have time

Alys:
@jeancl I've tested this on my local install and I really like how this works! However I won't merge it yet because others might like to look over the code, and we'll need to fix the merge conflicts.

If you're familiar with fixing merge conflicts, could you please do that now, and also while you're doing that, please remove the blank spaces from the ends of lines 8 and 24 in website/views/shared/tasks/edit/checklist.jade for code consistency? If you're not familiar with it, let us know and we'll either give you pointers or fix them ourselves before merging.

If anyone wants to review this before the conflicts are gone, these are the only files that are different than upstream/develop:
common/locales/en/tasks.json
website/public/css/tasks.styl
website/public/js/controllers/tasksCtrl.js
website/views/shared/tasks/edit/checklist.jade

Alys:
For anyone curious: I also tested in Habits and in the challenge edit and creation screens and this PR made no difference there (as expected and required).

jeancl:
Blank spaces removed and conflicts solved :)

crookedneighbor:
Looks like a rebase went wrong and you brought in a ton of other commits. Going to close this now, but if you want to sort that out and re-open the PR, I'm happy to look at it now that the API v3 work is done.

PR #6762, opened 2016-02-24
Add automatic note to Guilds with 5000+ members: Fix #6706

### LINKED ISSUE #6706: Add automatic message to Guilds with 5000+ Members
A new change to the site to promote stability has made it so that Guilds with more than 5000 members no longer support notifications. This change automatically takes effect when a Guild hits 5000 members.

When this happens, a message should automatically display at the top of the Guild stating that there will no longer be automatic notifications.

Suggested text (feel free to edit!):
"NOTE: This Guild is now too large to support notifications! Be sure to check back every day to see new messages."

khipkin:
This addresses issue #6706.

Adds an info alert to the top of the guild page if the guild has >5000 members that says:
"Note: This Guild is now too large to support notifications! Be sure to check back every day to see new messages."

@Alys suggested that the `5000` value should be pulled into a variable to be referenced throughout this code change, but I don't know if it's possible to reference a single variable across js and jade files.

Also, the use of `ng-if` is probably overkill.  Should I use `bo-if` instead?

---

To test this change, I created two guilds in my local database:
image

One of these guilds should have the note about notifications, and the other should not.

The large guild:
image

The small guild:
image

---

Is there a way I could automate these tests?

---

My Habitica User ID is f9c89bba-abf8-4f20-a974-cf1550ff603f


website/src/models/group.js
```diff
@@ -164,7 +164,8 @@ GroupSchema.methods.sendChat = function(message, user){
   // Kick off chat notifications in the background.
   var lastSeenUpdate = {$set:{}, $inc:{_v:1}};
   lastSeenUpdate['$set']['newMessages.'+group._id] = {name:group.name,value:true};
-  if (NO_CHAT_NOTIFICATIONS.indexOf(group._id) !== -1 || group.memberCount > 5000) {
+  var LARGE_GROUP_CUTOFF = 5000; // Guilds larger than 5000 cannot receive notifications.
```
crookedneighbor: Lets put this after the requires at the top of the file and make it a `const` instead of a `var`

website/views/options/social/group.jade
```diff
@@ -145,6 +145,7 @@ a.pull-right.gem-wallet(ng-if='group.type!="party"', popover-trigger='mouseenter
                 .popover-content
                   markdown(text='group._editing ? groupCopy.leaderMessage : group.leaderMessage')
       div(ng-controller='ChatCtrl')
+        .alert.alert-info.alert-sm(ng-if='group.memberCount > 5000')=env.t('largeGroupNote')
```
crookedneighbor: Not too keen about having the magic number in two different locations...

crookedneighbor:
What if instead we created a virtual property on the group model that returned true or false depending on whether or not it could receive notifications. IE,

```
const LARGE_GROUP_CUTOFF = 5000;
const NO_CHAT_NOTIFICATIONS = ['habitrpg'];

GroupSchema.virtual('canSendChatNotifications').get(function () {
  let ignoresAllChatNotifications = NO_CHAT_NOTIFICATIONS.indexOf(this._id) !== -1;
  if (ignoresAllChatNotifications) return false;

  if (this.memberCount > LARGE_GROUP_CUTOFF) return false;

  return true;
});
```

Then in both the server file and the view, you should be able to just check whether `group.canSendChatNotifications` is true or false.

Wrote this all on the fly, so who knows if it'll work as is. :)

Alys:
That would result in the "This Guild is now too large to support notifications" message being shown for the Tavern (which strictly speaking is accurate, although we typically don't refer to it as a guild), and it would also show that message for any guilds that we excluding in future for reasons other than size (e.g., we've considered excluding the Back Corner from having notifications, for reasons other than size).

khipkin:
How about this as a compromise:

```
GroupSchema.virtual('sizeCutoffForChatNotifications').get(function () {
  return 5000;
});
```

Then in both the server file and the view, you can call the function instead of using the hard-coded value in two separate places.

Another option:

```
GroupSchema.virtual('isTooLargeForChatNotifications').get(function () {
  return this.memberCount > 5000;
});
```

crookedneighbor:
turn `5000` into constant near the top of the model file and I'm :+1:

Alys:
Sorry @khipkin, we didn't notice you'd made a new commit. GitHub doesn't send us notifications of that, alas, so it's worth making a small comment when you've added a commit to a PR. (I've added a note about that to [Guidance for Blacksmiths > Pull Request](http://habitica.wikia.com/wiki/Guidance_for_Blacksmiths#Pull_Request) - we should have done that ages ago).

We'll try to get this reviewed soon.

khipkin:
@Alys, ahh that makes sense! I just figured I was back at the end of the line for review, and the line was pretty long. :stuck_out_tongue_winking_eye:

khipkin:
@Alys ping!

website/src/models/group.js
```diff
@@ -164,7 +175,7 @@ GroupSchema.methods.sendChat = function(message, user){
   // Kick off chat notifications in the background.
   var lastSeenUpdate = {$set:{}, $inc:{_v:1}};
   lastSeenUpdate['$set']['newMessages.'+group._id] = {name:group.name,value:true};
-  if (NO_CHAT_NOTIFICATIONS.indexOf(group._id) !== -1 || group.memberCount > 5000) {
+  if (NO_CHAT_NOTIFICATIONS.indexOf(group._id) !== -1 || group.isTooLargeForChatNotifications) {
```
cridenour: The NO_CHAT_NOTIFICATIONS variable was deleted above, right? Should probably remove the indexOf or keep the variable above.

Alys: @cridenour The NO_CHAT_NOTIFICATIONS was moved to a const further up the file.

@khipkin Sorry, we still haven't reviewed this. :( Thanks for keeping it up to date with develop. I'll try to get to this this weekend if no one else does it sooner.

Alys:
@khipkin I'm so sorry (again) about the delay on this. My local install was broken for a while.

I've tested this and it all works well! The message appears when it should and not when it shouldn't, the Tavern never sends notifications, nor do guilds that are too large, but other guilds do.

I'm happy to merge this and note it as partial credit towards tier 3, but @paglias : This modifies code that has been changed for API v3. How would you like us to handle this? Merge now and modify API v3 now/later, or merge this (with appropriate changes) after API v3 goes live?

paglias:
can we put this on hold for a few weeks until v3 is out?

Alys:
@khipkin When API v3 is released, there'll probably be code conflicts with this, but we'll help as much as needed to resolve them. I'm sorry about the extra delay. We definitely want this change you've made, it's just been a very busy time for us recently.

crookedneighbor:
I decided against using virtuals for now and just moved the count into a constant. We may want to move toward using virtuals in the future, but for now this will work. 

khipkin:
:cry:

PR #6763, opened 2016-02-24
TEST_DB_URI env override

querry43:
Allow individuals to override TEST_DB_URI when running tests.  This facilitates development under docker or other environments where mongo may not be running on localhost.

Tested with:
    TEST_DB_URI=mongodb://habitrpg_mongo_1/habitrpg_test gulp test:content

user id: bf56a7cf-0970-40f2-af1b-f4df3f9bda4d

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -6,6 +6,7 @@
     "FACEBOOK_KEY":"123456789012345",
     "FACEBOOK_SECRET":"aaaabbbbccccddddeeeeffff00001111",
     "NODE_DB_URI":"mongodb://localhost/habitrpg",
+    "TEST_DB_URI":"mongodb://localhost/habitrpg_test",
     "NODE_ENV":"development",
     "SESSION_SECRET":"YOUR SECRET HERE",
     "ADMIN_EMAIL": "you@example.com",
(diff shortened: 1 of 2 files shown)


Alys:
Good idea! But would it be better to make `gulp-tests.js` use `nconf`, in the same way that `gulp-console.js` does? The custom host name and database name would already be specified in config.json. Unless I'm missing something that makes that not possible!

querry43:
I thought about that. There is one problem in that nconf has the full uri
including the DB name which will differ. Would it make sense to add
TEST_DB_URI to config.json as a separate entry?

Alys:
Yes, I think so. That would keep all the customisation in one place.

crookedneighbor:
I'm +1 on this change.

querry43:
Done, @Alys , @crookedneighbor .  This does mean that people will have to update config.json tho, think that will be a problem?

Alys:
No, not a problem. Changes are made to config.json every so often, and they'll have the example one to refer to.

Alys:
Thank you @querry43! Could you please post to the Aspiring Coders guild to advise the other devs that a change has been made to config.json.example?

Tier awarded. Note that the higher tiers require more effort than the lower ones, so if you make any small changes from now on, you might find that they won't grant you an extra tier immediately, but they will be noted as partial credit towards a tier. There's a bit of information about that at [Contributor Rewards](http://habitrpg.wikia.com/wiki/Contributor_Rewards) but do ask if you have questions.

PR #6765, opened 2016-02-24
makes "Battle Monsters With Friends" Button persist until four members in party - fixes #6754

### LINKED ISSUE #6754: "Battle Monsters With Friends" Button should persist
We've discovered that users stay much more accountable to their habits and engaged with each other if they have at least 4 people in their party, so we want to encourage that! An easy way to do this is to edit the "Battle Monsters with Friends" button so that it persists until there are four users in the party, at which point it should disappear as usual.

colwem:
fixed maybe?

---

_edit by admin:_

**To Do:**
- [ ] Alys to arrange a migration from `user.preferences.displayInviteToPartyWhenPartyIs1` to whatever the new name of the preference becomes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,9 @@ common/dist/scripts/*
 common/dist/sprites/habitrpg-shared.css
 .elasticbeanstalk/config.yml
 
+
 test/spec/mocks/translations.js
+
+#vim
+.tern-project
+tags
--- a/.gitignore
+++ b/.gitignore
@@ -30,9 +30,4 @@ common/dist/scripts/*
 common/dist/sprites/habitrpg-shared.css
 .elasticbeanstalk/config.yml
 
-
 test/spec/mocks/translations.js
-
-#vim
-.tern-project
-tags
--- a/website/views/shared/header/header.jade
+++ b/website/views/shared/header/header.jade
@@ -31,9 +31,10 @@
           | {{Math.floor(user.stats.mp)}} / {{user._statsComputed.maxMP}}
     // party
     .party(ng-controller='PartyCtrl')
-      button.party-invite.btn.btn-primary(ng-click="inviteOrStartParty(group)",
-        ng-if="(!party.members || party.memberCount === 1) && user.preferences.displayInviteToPartyWhenPartyIs1",
-        popover="{{!party.members ? env.t('startAParty') : env.t('addToParty')}}", popover-placement="left", popover-trigger="mouseenter")
-        span=env.t("battleWithFriends")
       .herobox-wrap(ng-repeat='profile in partyMinusSelf')
         +herobox()
+      .party-invite-wrap
+        button.party-invite.btn.btn-primary(ng-click="inviteOrStartParty(group)",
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -30,4 +30,9 @@ common/dist/scripts/*
 common/dist/sprites/habitrpg-shared.css
 .elasticbeanstalk/config.yml
 
+
 test/spec/mocks/translations.js
+
+#vim
+.tern-project
```
crookedneighbor: These should not be included in the project repo.

crookedneighbor: Your global gitignore for your user should have these configurations.

website/views/shared/header/header.jade
```diff
@@ -31,9 +31,10 @@
...
     .party(ng-controller='PartyCtrl')
-      button.party-invite.btn.btn-primary(ng-click="inviteOrStartParty(group)",
-        ng-if="(!party.members || party.memberCount === 1) && user.preferences.displayInviteToPartyWhenPartyIs1",
-        popover="{{!party.members ? env.t('startAParty') : env.t('addToParty')}}", popover-placement="left", popover-trigger="mouseenter")
-        span=env.t("battleWithFriends")
       .herobox-wrap(ng-repeat='profile in partyMinusSelf')
         +herobox()
+      .party-invite-wrap
+        button.party-invite.btn.btn-primary(ng-click="inviteOrStartParty(group)",
+          ng-if="(!party.members || party.memberCount < 4 ) && user.preferences.displayInviteToPartyWhenPartyIs1",
```
crookedneighbor, line 38: Whelp, this preference name is now inaccurate. Not sure there's much to do about that unless we want to rename it and do a migration.

Alys, line 38: I'm in favour of changing it and I'm willing to do the migration.  The more accurate we make our names, the less pain we'll suffer later.

Perhaps `displayInviteToPartyWhenPartyIsSmall` to avoid future problems of this nature.

.gitignore
colwem: How do I remove it? I'm really new to this.  If I do a rebase --onto and then push origin --force will that work or will I mess everything up?  Or do I make a new commit that basically undoes that commit?  or do I fix it then create a new PR?

crookedneighbor: I wouldn't worry try rebasing unless you know what you're doing. This should be sufficient:

```
git revert d85410607229a5b7541112a1bd32abaab5d9d041
git push
```

We can squash the commits when we merge it in.

website/views/shared/header/header.jade
Alys, line 38: I've looked into what's required for a migration to change the preference name from `displayInviteToPartyWhenPartyIs1` and it should be simple. I can run one migration script shortly before this change goes live to create the new preference option in all user accounts, and then after go-live, run another to remove the old option. I've updated the first comment in the PR to add a reminder that I need to do this.

crookedneighbor:
Sorry this took a million years.

@Alys I don't think there's a need to run a migration right now. I'd like to discuss moving these website specific properties out of the database and into localstorage, so let's hold off on a migration.

PR #6768, opened 2016-02-25
Remove  query from group invites, replaced with two queries

gisikw:
Definitely need some eyes on this. Avoids an `$or` query by replacing it with two (or in some cases, one) query. The test suite passes, but I'm not certain what coverage looks like on this section of the API, so please review with caution!


crookedneighbor:
We determined that this wasn't the magic bullet. Closing since this will soon be replaced with v3 anyway.

PR #6775, opened 2016-02-26
Implement Party Cap - fixes #6756

### LINKED ISSUE #6756: Implement Party Cap
We've found that very large parties result in a lot of issues, both for the party (see: Quest rewards bug) and for the site as a whole. Therefore, we should probably implement a party size cap.

The number that we were tentatively thinking about as a cap was 30, but that is open to discussion! Parties that are currently larger than the cap should be grandfathered in, but prevented from adding any additional members. 

There should also definitely be a pop-up warning parties when they are approaching the cap, as well as announcing when they have reached it.

colwem:
fixes #6756

Not a complete solution but the best I know how to do.  

The trouble is that after giving an invite the group model is not updated to reflect the new invite.  The invite doesn't show up in the invites list and a new invite that puts the group over the 30 member + invite limit does not trigger the invite button to hide.  After a page refresh however the invite button will be hidden.  I also didn't implement any way of preventing a user from accepting an invite if the group is over it's limit but I think that would be a bad idea anyways.  If a user has an invite then let him join regardless. So what if the party goes 1 or 2 users over the limit.  At least the ability to invite more users is taken away.  These changes also don't in any way affect a "grandfathered" party except that they can't invite new users until they get below 30 again.

I also didn't put in any kind of warning but I don't really know people would like that done.  Should we just pop up a regular notification every time they invite someone after they reach 25 users?  Or should it be in the invite modal?  or near the invite button?

**EDIT:**  I figured out how to update the invites list.  I really don't know what I'm doing here so I have no idea whether what I did will cause problems in some way.  Also as before I can't run the tests and I didn't write any tests.  I also created a warning notification.

### DIFF
--- a/common/locales/en/groups.json
+++ b/common/locales/en/groups.json
@@ -121,6 +121,7 @@
     "inviteAlertInfo": "If you have friends already using Habitica, invite them by <a href='http://habitica.wikia.com/wiki/API_Options' target='_blank'>User ID</a> here.",
     "inviteExistUser": "Invite Existing Users",
     "byColon": "By:",
+    "partyCapWarning":  "Parties have a maximum size of <%= partyCap %>.  Your party currently has <%= memberAndInviteCount %> members and  invites.",
     "inviteNewUsers": "Invite New Users",
     "sendInvitations": "Send Invitations",
     "invitationsSent": "Invitations sent!",
--- a/website/public/js/controllers/inviteToGroupCtrl.js
+++ b/website/public/js/controllers/inviteToGroupCtrl.js
@@ -26,11 +26,9 @@ habitrpg.controller('InviteToGroupCtrl', ['$scope', 'User', 'Groups', 'injectedG
       _inviteByMethod(inviteMethod);
     };
 
-    $scope.getMemberAndInviteCount = function(group) {
-      return (group && group.memberCount) 
-              ? group.memberCount + (group.invites ?  group.invites.length: 0)
-              : 0;
-    }
+    $scope.isOverPartyCap = Groups.isOverPartyCap;
+    $scope.isOverPartyCapWarning = Groups.isOverPartyCapWarning;
+    $scope.memberAndInviteCount = Groups.memberAndInviteCount;
 
     function _inviteByMethod(inviteMethod) {
       var invitationDetails;
@@ -47,7 +45,7 @@ habitrpg.controller('InviteToGroupCtrl', ['$scope', 'User', 'Groups', 'injectedG
 
       Groups.Group.invite({gid: $scope.group._id}, invitationDetails, function(group){
         Notification.text(window.env.t('invitationsSent'));
-        if (group.invites) $scope.group.invites = group.invites;
+        if (group && group.invites) $scope.group.invites = group.invites;
         _resetInvitees();
       }, function(){
         _resetInvitees();
--- a/website/public/js/controllers/inviteToGroupCtrl.js
+++ b/website/public/js/controllers/inviteToGroupCtrl.js
@@ -26,6 +26,10 @@ habitrpg.controller('InviteToGroupCtrl', ['$scope', 'User', 'Groups', 'injectedG
       _inviteByMethod(inviteMethod);
     };
 
+    $scope.getMemberAndInviteCount = function(group) {
+      return group.memberCount + group.invites.length;
+    }
+
     function _inviteByMethod(inviteMethod) {
       var invitationDetails;
 
@@ -39,8 +43,9 @@ habitrpg.controller('InviteToGroupCtrl', ['$scope', 'User', 'Groups', 'injectedG
         return console.log('Invalid invite method.')
       }
 
-      Groups.Group.invite({gid: $scope.group._id}, invitationDetails, function(){
+      Groups.Group.invite({gid: $scope.group._id}, invitationDetails, function(group){
         Notification.text(window.env.t('invitationsSent'));
+        if (group.invites) $scope.group.invites = group.invites;
         _resetInvitees();
       }, function(){
         _resetInvitees();
--- a/website/public/js/controllers/inviteToGroupCtrl.js
+++ b/website/public/js/controllers/inviteToGroupCtrl.js
@@ -27,6 +27,7 @@ habitrpg.controller('InviteToGroupCtrl', ['$scope', 'User', 'Groups', 'injectedG
     };
 
     $scope.getMemberAndInviteCount = function(group) {
+
       return group.memberCount + group.invites.length;
     }
 
--- a/website/public/js/controllers/inviteToGroupCtrl.js
+++ b/website/public/js/controllers/inviteToGroupCtrl.js
@@ -27,8 +27,7 @@ habitrpg.controller('InviteToGroupCtrl', ['$scope', 'User', 'Groups', 'injectedG
     };
 
     $scope.getMemberAndInviteCount = function(group) {
-
-      return group.memberCount + group.invites.length;
+      return group.memberCount + (group.invites ? group.invites.length: 0);
     }
 
     function _inviteByMethod(inviteMethod) {
--- a/website/public/js/controllers/inviteToGroupCtrl.js
+++ b/website/public/js/controllers/inviteToGroupCtrl.js
@@ -27,7 +27,9 @@ habitrpg.controller('InviteToGroupCtrl', ['$scope', 'User', 'Groups', 'injectedG
     };
 
     $scope.getMemberAndInviteCount = function(group) {
-      return group.memberCount + (group.invites ? group.invites.length: 0);
+      return (group && group.memberCount) 
+              ? group.memberCount + (group.invites ?  group.invites.length: 0)
+              : 0;
     }
 
     function _inviteByMethod(inviteMethod) {
--- a/website/public/js/services/groupServices.js
+++ b/website/public/js/services/groupServices.js
@@ -77,6 +77,21 @@
         $location.path("/options/groups/party");
       }
     }
+    
+    function isOverPartyCap(group) {
+      return group.memberCount + group.invites.length >= group.partyCap;
+    }
+
+    function isOverPartyCapWarning(group) {
+      var partyCapWarningBoundary = 5;
+      return group.memberCount + group.invites.length > group.partyCap - partyCapWarningBoundary;
+    }
+
+    function memberAndInviteCount(group) {
+      if(!group) return 0;
+      if(!group.invites) return group.memberCount;
+      return group.memberCount + group.invites.length;
+    };
 
     return {
       party: party,
@@ -84,6 +99,9 @@
       myGuilds: myGuilds,
       tavern: tavern,
       inviteOrStartParty: inviteOrStartParty,
+      isOverPartyCap: isOverPartyCap,
+      isOverPartyCapWarning: isOverPartyCapWarning,
+      memberAndInviteCount: memberAndInviteCount,
 
       data: data,
       Group: Group
--- a/website/src/controllers/api-v2/groups.js
+++ b/website/src/controllers/api-v2/groups.js
@@ -693,7 +693,7 @@ api.invite = function(req, res, next){
   if (group.privacy === 'private' && !_.contains(group.members,res.locals.user._id)) {
     return res.json(401, {err: "Only a member can invite new members!"});
   }
-  if (group.memberCount + group.invites.length >= 30) {
+  if (group.isOverPartyCap) {
     return res.json(400, {err: "Over 30 member party cap"});
   } 
 
--- a/website/src/controllers/api-v2/groups.js
+++ b/website/src/controllers/api-v2/groups.js
@@ -693,6 +693,10 @@ api.invite = function(req, res, next){
   if (group.privacy === 'private' && !_.contains(group.members,res.locals.user._id)) {
     return res.json(401, {err: "Only a member can invite new members!"});
   }
+  if (group.memberCount + group.invites.length >= 30) {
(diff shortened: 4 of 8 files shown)


colwem:
Oh this travis thing is great.  Right now I can't actually run these tests on my computer.  Is there a way for me to push my code to travis before I submit the PR rather than after?

paglias:
@colwem unfortunately to use travis you have to submit a PR or to push a commit to an existing one

colwem:
So it just now occurs to me that this would need translation right?  Will that be required before it's merged or can that be done later?  Do I need to extract out the English text?

Alys:
You don't need to do translations - see the _README_FIRST.md file in any one of the locales directories (apart from en). In brief, it's done in transifex.

colwem:
What happened to travis.  Is there a limit to the number of checks it will do?

paglias:
@colwem not sure why it didn't run... can you try pushing another commit?

colwem:
The errors make it seem like that was run on a previous version of the code.  What I did was I ran 

```
git commit --amend
git push origin issue-6756 -f
```

each time I thought I had it fixed.  Could that be the problem?  Should I not amend commits after I've created the PR?  Should I instead just make new commits every time?

website/public/js/controllers/inviteToGroupCtrl.js
```diff
@@ -26,6 +26,11 @@ habitrpg.controller('InviteToGroupCtrl', ['$scope', 'User', 'Groups', 'injectedG
       _inviteByMethod(inviteMethod);
     };
 
+    $scope.getMemberAndInviteCount = function(group) {
+
+      return group.memberCount + group.invites.length;
```
paglias: @colwem it's possible group.invites is undefined here? Try adding a check for it.

`return group.memberCount + (group.invites ? group.invites.length : 0)`

colwem: True, but why does the error reference line 9?  A similar line to this was on line 9 in a previous commit and it produced the same exact error.  But now that line is gone.

Alys:
I don't think that amending commits and then force pushing would cause any problems with Travis, but it's best to avoid that for other reasons. Have a look at these pages when you have time: https://www.atlassian.com/git/tutorials/rewriting-history/git-commit--amend and https://help.github.com/articles/changing-a-commit-message/ 

Each time you make a code change, even if it's small, it's worth doing a new commit because that gives an accurate history of your work. The only time I ever amend a commit is when I have made the commit in the past couple of minutes and I realise that I forgot to include a file or a tiny code change (e.g., updating a version number), or I see a typo in my commit comment. Any larger changes are basically different pieces of work (albeit related) and so they deserve their own commits. That can make it easier for yourself and others to review your work and can also help to pinpoint and revert bugs.

Also, amending a commit after pushing it to a repo isn't a good idea for the reasons listed in those links. That's why it's necessary to use the force option to git push - git is trying to tell you that maybe you don't want to do what you're doing. ;)

website/src/controllers/api-v2/groups.js
```diff
@@ -691,8 +691,9 @@ var inviteByEmails = function(invites, group, req, res, next){
 
 api.invite = function(req, res, next){
   var group = res.locals.group;
-
-  if (req.body.uuids) {
+  if (group.memberCount + group.invites.length >= 30) {
+    return res.json(400, {err: "Over 30 member party cap"});
```
Alys, line 697: Could you please put `30` into a variable or similar that can be used in all places it's needed throughout the code base? There's an idea for how to do that here: https://github.com/HabitRPG/habitrpg/pull/6762#issuecomment-188060614
but you might have your own ideas. Feel free to ask if you'd like feedback or clarifications.

Alys, line 697: Also the error message (in fact, any text anywhere that the user can see) should be translatable. I'm happy to give more details about how to do this, if you wish, including how to include `30` as a variable in it - just ask if you want suggestions (and make me write that up in the Blacksmiths wiki page!)

colwem, line 697: Yup, that 30 as a variable thing sounds nice.  To be clear I mean how do I include 30 as a variable in the env.t('messageToTranslate') function and the associated templates.

paglias, line 697: @colwem you can add a variable in a string like it's done here https://github.com/HabitRPG/habitrpg/blob/develop/common/locales/en/challenge.json#L41, then to pass the value you'll do `env.t('stringName', {varName: 30})`

colwem, line 697: Oh, I thought this was a different line.  I was asking about doing the translation in `invite-friends.jade`.  As for the 400 error message in `groups.js` I don't know how that's done.  I just copied that from line 702 which is also not translated.

website/views/shared/modals/invite-friends.jade
```diff
@@ -54,6 +54,10 @@ script(type='text/ng-template', id='modals/invite-party.html')
     div(ng-if='!group._id')
       h4(style='margin-top:0')=env.t('nameYourParty')
       input.form-control#new-group-name.input-medium.option-content(required, type='text', ng-model='group.name')
+    p.alert.alert-warning(ng-if="getMemberAndInviteCount(group) > 25") 
+      | You are approaching the 30 member party cap!
```
colwem: How do I include a variable in a translation for this situation?

website/src/controllers/api-v2/groups.js
paglias, line 697: @colwem I think for the api it can be skipped. I'll add the translation in the upcoming api v3

website/views/shared/modals/invite-friends.jade
paglias: to split the translation string in multiple lines you can use the `\n` character (for example `"unorderedListMarkdown": "+ First item\n+ Second item\n+ Third item"
`). The `{{getMemberAndInviteCount(group)}}` code instead can be passed as a variable like `button.btn.btn-primary.btn-block(ng-show='selectedEgg', ng-click='sellInventory()')=env.t('sellForGold', {itemType: "{{selectedEgg.text()}}", gold: "{{selectedEgg.value}}"})
`

colwem:
Why do the tests have groups that don't have an invites member?  Do I really have to worry about that or is this something that should be fixed in the tests?

Alys:
@colwem I'm actually working on the groups tests at the moment for another reason, and I'll be adding members to the test groups. You might want to hold off on your changes until I post here again. It should only be a day or two. Feel free to nag me aggressively if it's longer than that!

Alys:
@colwem I've finished the work on the groups that I was talking about. I actually didn't have to adjust the existing tests because the problem I was having with them turned out to be not related to my new code - sorry about being misleading there. I added new tests related to inviting members to a group: test/api/v2/groups/POST-groups_id_invite.test.js
Is that what you were asking about? There's other tests related to joining groups in test/api/v2/groups/POST-groups_id_join.test.js

If you're looking for tests to copy and modify for new purposes, use the latter file rather than the former because it would have most likely been written by someone more familiar with our test code than I am.

colwem:
Ok, so now I have conflicts... How do I resolve this?

colwem:
@Alys no I was referring to a series of tests that were repeatedly failing this code because they were creating groups that didn't have `invites` members on them.  They were from unrelated tests that had nothing to do with invites so I guess they were using mock group objects that didn't have `invites` members.  But yes, I would love to add actual tests to this PR but what's the point in me doing it when I can barely get any tests to pass in my setup.

Alys:
@colwem For resolving conflicts, the best information I can give you is this link: https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging
I know that seems like a cop-out, but that is an excellent online book that will give you a solid understanding of what has happened when there's a conflict and how you can resolve it.  There's also a lot of other, simpler advice on the web (google for "git resolve merge conflict") that will probably help you fix the conflict rapidly, but I do recommend reading that page if you have the time. If you have specific questions about this particular merge, you're welcome to ask! Also see the information that GitHub provides here of course (the command-line instructions below this comment) - that will give you specific commands suitable for this particular commit (assuming you can see them; I hope they're not an admin-only feature).

colwem:
@Alys Oh, no I totally understand how merging works, I've read that book through chapter 5.  What I don't know about is this message in this PR.  Is it just telling us that we will need to resolve the conflicts when this PR is merged or is it telling me that I need to resolve these conflicts right now?.  Should I be continually rebasing my PR on upstream/develop?  Or do I leave it where it is until it gets merged?  Or should I be merging upstream/develop into it?

colwem:
@Alys 

> (assuming you can see them; I hope they're not an admin-only feature).

Nope, can't see them

Alys:
@colwem In general, if your pull request has merge conflicts and you need to make more changes that will be pushed to the pull request, then you'll have to fix the conflicts. If you don't need to make more changes, then the conflicts will still have to be fixed before the PR can be merged, but we'll either fix them ourselves when we merge the PR or ask you to fix them.

colwem:
@Alys I don't know if I need to make changes.  I just want to see if the travis tests will pass.  If they don't then yeah, I'll make changes.  

So normally the instructions here are to rebase your work onto upstream/develop.  But as I understand it you should never rebase published work.  So does that mean I shouldn't rebase branches that have been PRed?  So I should do a regular pull or fetch-merge this time?

crookedneighbor:
@Alys I have to disagree a little bit about your git philosophy. There's no problem with amending/rebasing and force pushing if you're the only one working on the branch. In @colwem's case, since no one else has write access to their fork of the repo, this is certainly the case. 

Git commits should be atomic, that is, each commit encompasses a single change. If I'm the one to  merge this in, I will probably rebase and squash all of these commits into a single commit before merging. This will also make it much easier to revert if an urgent problem is discovered with it than reverting 7-10 commits. 

So @colwem, if you're comfortable rebasing, go for it. Many users that have rebased on a Habit PR in the past have not known what they were doing and ended up adding a bunch of duplicate commits instead of properly rebasing. (Full disclosure: I have been guilty of this before) If not, merging with upstream and fixing the conflicts is fine.

Alys:
I wouldn't call it my git philosophy, as such. :) It's a fairly common recommendation, but your point about colwem being the only one to write to their fork is very pertinent. I agree.

colwem:
Ok, rebased and passing finally.  What's the next step?  Ready for review? Is there more to do? Do I need write tests now?

website/src/controllers/api-v2/groups.js
```diff
@@ -693,6 +693,10 @@ api.invite = function(req, res, next){
   if (group.privacy === 'private' && !_.contains(group.members,res.locals.user._id)) {
     return res.json(401, {err: "Only a member can invite new members!"});
   }
+  if (group.type === 'party' && group.isOverPartyCap) {
+    return res.json(400, {err: "Over 30 member party cap"});
```
Alys, line 697: Can this error message either use the PARTY_CAP variable (best if possible) or not mention the number at all?

website/src/models/group.js
```diff
@@ -8,6 +8,8 @@ var logging = require('../libs/logging');
 var Challenge = require('./../models/challenge').model;
 var firebase = require('../libs/firebase');
 
+const PARTY_CAP = 30;
```
crookedneighbor, line 11: I believe you can export this like:

``` js
module.exports.PARTY_CAP = const PARTY_CAP = 30;
```

website/public/js/services/groupServices.js
```diff
@@ -77,6 +77,21 @@
         $location.path("/options/groups/party");
       }
     }
+    
+    function isOverPartyCap(group) {
```
crookedneighbor, line 81: Isn't this just `group.isOverPartyCap` as defined on the virtual attribute for the group instance? Or am I missing something?

website/src/controllers/api-v2/groups.js
crookedneighbor, line 697: Doesn't `group.isOverPartyCap` check that the group is a party?

You can insert the number by doing:

```
"Over " + group.partyCap + " member party cap"
```

website/src/models/group.js
crookedneighbor, line 11: Actually, there's no need to export it since it's a virtual property on the group.

website/public/js/services/groupServices.js
colwem, line 81: This is client side.  Unless I'm missing something the server side function is exported as a property which will become out of date if users are invited or removed.  So as the model changes the value needs to be updated.  I'm not happy about it either though.  What I guess I would like to do is completely update `group` every time `removeMember` or `invite` is run.  This is doable with the `invite` function because it returns the model, but the `removeMember` function doesn't.

website/src/controllers/api-v2/groups.js
colwem, line 697: Yeah, I've actually already done this.  I have a commit ready that I haven't pushed.

common/locales/en/groups.json
```diff
@@ -121,6 +121,7 @@
     "inviteAlertInfo": "If you have friends already using Habitica, invite them by <a href='http://habitica.wikia.com/wiki/API_Options' target='_blank'>User ID</a> here.",
     "inviteExistUser": "Invite Existing Users",
     "byColon": "By:",
+    "partyCapWarning":  "Parties have a maximum size of <%= partyCap %>.  Your party currently has <%= memberAndInviteCount %> members and  invites.",
```
khipkin, line 124: It looks like there's an extra space in this string, "members and *\* invites".

paglias:
I'm going to take over this as part of #7664. @colwem thanks for your help!

PR #6780, opened 2016-02-27
Improve styling of markdown blockquotes - fixes #6774

### LINKED ISSUE #6774: Markdown rendering of blockquotes could be improved.
The text just gets bigger which makes little sense for a block quote.  Should have a marker of some sort down the side to indicate that it's a quote and if anything the text should get less strong rather than more.  I'm a reddit user so I use 

> quotes like this

All the time to remind people of what comment I'm referring to.  Instead of being rendered like it is in github above it's rendered more

colwem:
Fixes issue #6774

More sensible blockquotes in chat.

example

UserID: 041abe75-7ebc-4e11-a32d-9d54f77d74f8

### DIFF
--- a/website/public/css/game-pane.styl
+++ b/website/public/css/game-pane.styl
@@ -52,7 +52,7 @@
     color: #999
 
 .tavern-chat, .party-chat
-  
+
   li
     padding-top:15px
     padding-bottom:15px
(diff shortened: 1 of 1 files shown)


Alys:
Looks okay in chat and doesn't change any of the other places where blockquotes are used, that I can find (website/views/shared/modals/members.jade; website/views/static/front.jade and associated css files).

Alys:
Tier 1 awarded (041abe75-7ebc-4e11-a32d-9d54f77d74f8). Congratulations @colwem!

PR #6781, opened 2016-02-27
allow bulk task creation in challenges - fixes #6723

### LINKED ISSUE #6723: Cannot add multiple tasks when creating or editing a challenge
When editing a challenge clicking Add Multiple does nothing.

UserID 8da9ca27-8fdd-4612-936e-0ed56241d9d7
Chrome Version 48.0.2564.109 m
Windows 10 version 1511

colwem:
fixes issue #6723

`challengesCtrl.js` didn't have any of the functionality implemented for bulk task creation.  I basically just copied over what was in `tasksCtrl.js` and modified it to work with challenges.

By the way, is there a way to link references to code from a file that's not in this PR?  In this case I would like to just point to the lines of code in `tasksCtrl.js` that I copied.

UserID: 041abe75-7ebc-4e11-a32d-9d54f77d74f8

### DIFF
--- a/website/public/js/controllers/challengesCtrl.js
+++ b/website/public/js/controllers/challengesCtrl.js
@@ -214,6 +214,7 @@ habitrpg.controller("ChallengesCtrl", ['$rootScope','$scope', 'Shared', 'User', 
 
     $scope.toggleBulk = Tasks.toggleBulk;
 
+
     $scope.removeTask = function(task, list) {
       if (!confirm(window.env.t('sureDelete', {taskType: window.env.t(task.type), taskText: task.text}))) return;
       //TODO persist
--- a/website/public/js/controllers/challengesCtrl.js
+++ b/website/public/js/controllers/challengesCtrl.js
@@ -204,32 +204,15 @@ habitrpg.controller("ChallengesCtrl", ['$rootScope','$scope', 'Shared', 'User', 
     // Tasks
     //------------------------------------------------------------
 
-    $scope.addTask = function(addTo, listDef) {
-      if (listDef.bulk) {
-        var tasks = listDef.newTask.split(/[\n\r]+/);
-        tasks.reverse();
-        _.each(tasks, function(t) {
-          var task = Shared.taskDefaults({text: t, 
-                                         type: listDef.type});
-          addTo.unshift(task);
-        });
-        listDef.bulk = false;
-      } else {
-        var task = Shared.taskDefaults({text: listDef.newTask, 
-                                       type: listDef.type});
-        addTo.unshift(task);
-      }
-      delete listDef.newTask;
-      delete listDef.focus;
+    $scope.addTask = function(list, listDef) {
+      Tasks.addTask(listDef, function(t) {
+        var task = Shared.taskDefaults({text: t, 
+                                        type: listDef.type});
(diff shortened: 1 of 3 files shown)


crookedneighbor:
Sounds like these functions should be extracted into a service so they can be used bothe places.

colwem:
@crookedneighbor ok, extracted.

colwem:
I don't know what to make of this error.

website/public/js/services/taskServices.js
```diff
@@ -15,6 +15,30 @@
 
   function tasksFactory($rootScope, Shared, User) {
 
+    function addTask(listDef, fn) {
```
crookedneighbor, line 18: These functions needs tests.

website/public/js/controllers/challengesCtrl.js
```diff
@@ -204,13 +204,17 @@ habitrpg.controller("ChallengesCtrl", ['$rootScope','$scope', 'Shared', 'User',
...
+      Tasks.addTask(listDef, function(t) {
+        var task = Shared.taskDefaults({text: t, 
+                                        type: listDef.type});
+        list.unshift(task);
+      });
     };
 
+    $scope.toggleBulk = Tasks.toggleBulk;
+
+
```
crookedneighbor, line 217: Extra space here.

website/public/js/services/taskServices.js
```diff
@@ -15,6 +15,30 @@
...
+      } else {
+        var task = listDef.newTask;
+        fn(task);
+      }
+      delete listDef.newTask;
+      delete listDef.focus;
+    }
+
+    function toggleBulk(list) {
+      if (typeof list.bulk === 'undefined') {
```
crookedneighbor, line 35: I know you didn't write this part of the code, but this if is unnecessary.

`!undefined --> true` which is the same as `!false`

crookedneighbor:
I'm not happy with how the extraction into the service is looking. Will have to think about it a little longer.

website/public/js/services/taskServices.js
colwem, line 35: I'm sure this doesn't matter but I don't think that's true.  I mean `typeof list.bulk === 'undefined'` is only true if `list.bulk` is undef. But `!list.bulk` would be true if it were 0,null, '', false, undefined and maybe others.  I'm probably misunderstanding you though.

colwem:
Without knowing why the `addTask` implementation differed between the two controllers I think this is the best extraction that can be done.  However if we can just make both controllers.  between challenges and tasks I think this is best solution available.  But if we can make both controllers create and save tasks in the same way then that would be better.

I'm a bit busy at the moment I won't be able to get back to this for a little while, maybe someone else should add tests and fix whatever it is you don't like about the extraction.  If not I'll be back to this in a week or so.

website/public/js/services/taskServices.js
colwem, line 35: So... nevermind, I made that comment without actually re-reading the code, you're right, that line is pointless.

crookedneighbor:
It's been a million years, but I was finally able to incorporate your changes into the codebase. Thanks so much for your work on this.

I've notated this PR toward your next tier.

PR #6786, opened 2016-02-28
Use level style in modals

murraybd:
Upon leveling up one is presented with a modal dialog indicating that they have gained a level. The individual's level is not styled properly though and subsequently if one is a contributor or a backer this is not indicated in the modal dialog. After a lot of digging (hours really!) I discovered this is because the userLevelStyle was set as a scope and not a root scope.  Since it was not a root scope it can not be used in a modal dialog. This fixes that issue.

Here's a screnshot of the fix.

habitica-fix2


murraybd:
This is what the dialog looks like without my change.

level-not-colored

Alys:
@murraybd Thanks! This will count towards a contributor tier. What's your User ID from Settings > API?

murraybd:
Thanks! I'll be excited to level up and see this change.  My User ID is: 5b93ce83-2508-4ea4-9647-079899dffa96

Alys:
I've noted this as being credit towards tier 3.

Issue #6776, opened 2016-02-26
API v3 Short-Term Roadmap 2


paglias:
This is a list of work items regarding the upcoming API v3. 

If you plan to make a Pull Request please post a comment to this issue before starting.

We'd like to have atomic PRs as much as possible so each one should implement only one route with the related tests. All PRs must be done against the api-v3` branch.

Items already assigned are marked with the username of the person assigned.

---
# Routes
### Authentication
- [x] Tests for login route (assigned to @piousbox)
- [x] Tests for attaching local authentication to a Facebook user
- [x] Change email / username / password (assigned to @piousbox)
- [x] Reset password (assigned to @piousbox)
- [x] Tests for deleting social authentication (assigned to @piousbox)
- [x] getFirebaseToken (assigned to @paglias)
- [x] setupPassport (assigned to @paglias)
### Other routes
- [x] delete user (along with his tasks and remove from challenges, groups, ...) (assigned to @TheHollidayInn)
- [x] getUserAnonymized (assigned to @piousbox)
- [x] update user (assigned to @crookedneighbor)
- [x] cast (assigned to @paglias)
- [x] sessionPartyInvite (refactor and tests) (assigned to @paglias)
- [x] getBuyList
- [x] getModelPaths (assigned to @paglias)
- [x] user.tmp and user.filters (assigned to @paglias)
### Dev routes (in separate controller)
- [x] add ten gems (assigned to @TheHollidayInn)
- [x] add hour glass (assigned to @TheHollidayInn)
## Content
- [x] getContent (was in user controller in api v2)
## Groups
- [x] reject group invite (assigned to @TheHollidayInn)
## Members (needs refactoring!)
- [x] sendMessage (assigned to @TheHollidayInn)
- [x] sendPrivateMessage (move to inbox routes?) (assigned to @TheHollidayInn)
- [x] sendGift (assigned to @TheHollidayInn)
## Challenges
- [x] tests for /unlink (assigned to @TheHollidayInn)
- [x] unit tests for sync to members (assigned to @TheHollidayInn)
## Tasks
- [x] tests for routes used to manage tasks for challenges (assigned to @TheHollidayInn)
- [x] test task scoring for challenges tasks (assigned to @TheHollidayInn)

TheHollidayInn:
I can take the following if no one else is working on them.
- test task scoring for challenges tasks
- reject group invite
- add ten gems
-  add hour glass

paglias:
[...]

PR #6787, opened 2016-02-28
Added initial challenge model tests

TheHollidayInn:
Added the challenge model test to complete the unit tests for #6776

### DIFF
--- a/test/api/v3/unit/models/challenge.test.js
+++ b/test/api/v3/unit/models/challenge.test.js
@@ -1,0 +1,121 @@
+import { model as Challenge } from '../../../../../website/src/models/challenge';
+import { model as Group } from '../../../../../website/src/models/group';
+import { model as User } from '../../../../../website/src/models/user';
+import * as Tasks from '../../../../../website/src/models/task';
+import { each } from 'lodash';
+
+describe('Challenge Model', () => {
+  let guild, leader, challenge, task;
+  let tasksToTest = {
+    habit: {
+      text: 'test habit',
+      type: 'habit',
+      up: false,
+      down: true,
+      notes: 1976,
+    },
+    todo: {
+      text: 'test todo',
+      type: 'todo',
+      notes: 1976,
+    },
+    daily: {
+      text: 'test daily',
+      type: 'daily',
+      notes: 1976,
+      frequency: 'daily',
+      everyX: 5,
+      startDate: new Date(),
+    },
+    reward: {
+      text: 'test reward',
+      type: 'reward',
+      notes: 1976,
+    },
+  };
+
+  beforeEach(async () => {
+    guild = new Group({
+      name: 'test party',
+      type: 'guild',
+    });
+
+    leader = new User({
+      guilds: [guild._id],
+    });
+
+    guild.leader = leader._id;
+
+    challenge = new Challenge({
+      name: 'Test Challenge',
+      shortName: 'Test',
+      leader: leader._id,
+      group: guild._id,
+    });
+
+    leader.challenges = [challenge._id];
+
+    await Promise.all([
+      guild.save(),
+      leader.save(),
+      challenge.save(),
+    ]);
+  });
+
+  each(tasksToTest, (taskValue, taskType) => {
+    context(`${taskType}`, () => {
+      before(async() => {
+        task = new Tasks[`${taskType}`](Tasks.Task.sanitizeCreate(taskValue));
+      });
+
+      it('adds tasks to challenge and challenge members', async () => {
+        await challenge.addTasks([task]);
+
+        let updatedLeader = await User.findOne({_id: leader._id});
+
+        expect(updatedLeader.tasksOrder[`${taskType}s`].length).to.be.above(0);
+      });
+
+      it('syncs a challenge to a user', async () => {
+        await challenge.addTasks([task]);
+
+        let newMember = new User({
+          guilds: [guild._id],
+        });
+        await newMember.save();
+
+        await challenge.syncToUser(newMember);
+
+        let updatedNewMember = await User.findById(newMember._id);
+
+        expect(updatedNewMember.challenges).to.contain(challenge._id);
+        expect(updatedNewMember.tags[3]._id).to.equal(challenge._id);
+        expect(updatedNewMember.tags[3].name).to.equal(challenge.shortName);
+        expect(updatedNewMember.tasksOrder[`${taskType}s`].length).to.be.above(0);
+      });
+
+      it('updates tasks to challenge and challenge members', async () => {
+        let updatedTaskName = 'Updated Test Habit';
+        await challenge.addTasks([task]);
+
+        _.assign(task, _.merge(task.toObject(), Tasks.Task.sanitizeUpdate({ text: updatedTaskName })));
+        await challenge.updateTask(task);
+
+        let updatedLeader = await User.findOne({_id: leader._id});
+        let updatedUserTask = await Tasks.Task.findById(updatedLeader.tasksOrder[`${taskType}s`][0]);
+
+        expect(updatedUserTask.text).to.equal(updatedTaskName);
+      });
+
+      it('removes a tasks to challenge and challenge members', async () => {
+        await challenge.addTasks([task]);
+        await challenge.removeTask(task);
+
+        let updatedLeader = await User.findOne({_id: leader._id});
+        let updatedUserTask = await Tasks.Task.findOne({_id: updatedLeader.tasksOrder[`${taskType}s`][0]}).exec();
+
+        expect(updatedUserTask.challenge.broken).to.equal('TASK_DELETED');
+      });
+    });
+  });
+});
--- a/test/api/v3/unit/models/challenge.test.js
+++ b/test/api/v3/unit/models/challenge.test.js
@@ -122,6 +122,33 @@ describe('Challenge Model', () => {
 
         expect(updatedUserTask.challenge.broken).to.equal('TASK_DELETED');
       });
+
+      it('unlinks and deletes challenge tasks for a user when remove-all is specified', async () => {
+        await challenge.addTasks([task]);
+        await challenge.unlinkTasks(leader, 'remove-all');
+
+        let updatedLeader = await User.findOne({_id: leader._id});
+        let updatedLeadersTasks = await Tasks.Task.find({_id: { $in: updatedLeader.tasksOrder[`${taskType}s`]}});
+        let syncedTask = find(updatedLeadersTasks, function findNewTask (updatedLeadersTask) {
+          return updatedLeadersTask.type === taskValue.type && updatedLeadersTask.text === taskValue.text;
+        });
+
+        expect(syncedTask).to.not.exist;
+      });
+
+      it('unlinks and keeps challenge tasks for a user when keep-all is specified', async () => {
+        await challenge.addTasks([task]);
+        await challenge.unlinkTasks(leader, 'keep-all');
+
+        let updatedLeader = await User.findOne({_id: leader._id});
+        let updatedLeadersTasks = await Tasks.Task.find({_id: { $in: updatedLeader.tasksOrder[`${taskType}s`]}});
+        let syncedTask = find(updatedLeadersTasks, function findNewTask (updatedLeadersTask) {
+          return updatedLeadersTask.type === taskValue.type && updatedLeadersTask.text === taskValue.text;
+        });
+
+        expect(syncedTask).to.exist;
+        expect(syncedTask.challenge._id).to.be.empty;
+      });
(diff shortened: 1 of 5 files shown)


website/src/models/challenge.js
```diff
@@ -97,7 +97,6 @@ schema.methods.syncToUser = async function syncChallengeToUser (user) {
   let [challengeTasks, userTasks] = await Q.all([
     // Find original challenge tasks
     Tasks.Task.find({
-      userId: {$exists: false},
```
paglias, line 100: why this was removed? we want to select the original challenge tasks, this way it will also get the ones that are already synced to the users

test/api/v3/unit/models/challenge.test.js
```diff
@@ -0,0 +1,121 @@
...
+      before(async() => {
+        task = new Tasks[`${taskType}`](Tasks.Task.sanitizeCreate(taskValue));
+      });
+
+      it('adds tasks to challenge and challenge members', async () => {
+        await challenge.addTasks([task]);
+
+        let updatedLeader = await User.findOne({_id: leader._id});
+
+        expect(updatedLeader.tasksOrder[`${taskType}s`].length).to.be.above(0);
```
paglias: can we check for the exact number here?

paglias: mostly because some tasks are added on user creation so this test may pass even if the challenge tasks haven't been added

TheHollidayInn: That make sense. I think I know a better way to check this.

website/src/models/challenge.js
TheHollidayInn, line 100: Hmm. I thought it would be okay to grab all. I was having a problem with the userId being set. If we can't grab all the tasks, I have a another idea.

paglias, line 100: The point was that we want to fetch the challenge tasks and sync them to the user so there's no point in fetching the challenge tasks that are already synced to the users.

TheHollidayInn, line 100: Actually, doesn't the logic below check to see if the task is already synced? 

`let matchingTask = _.find(userTasks, userTask => userTask.challenge.taskId === chalTask._id);`

It handles tasks already syncs and checks for updates. The problem with the previous query is that the userId is set. I also don't see how that property relates to the user already having the task synced. The userId property only holds on user.

test/api/v3/unit/models/challenge.test.js
```diff
@@ -0,0 +1,130 @@
...
+
+  each(tasksToTest, (taskValue, taskType) => {
+    context(`${taskType}`, () => {
+      before(async() => {
+        task = new Tasks[`${taskType}`](Tasks.Task.sanitizeCreate(taskValue));
+      });
+
+      it('adds tasks to challenge and challenge members', async () => {
+        // Resetting the todos because at todo is added when a new user is saved
+        leader.tasksOrder.todos = [];
```
TheHollidayInn: @paglias is there a better way to prevent the default todo from being added? I used this rather than conditional logic during the expect tests. But, it is still pretty bad..

paglias: I usually count manually for the number of todos are added automatically. It'll have to be changed if add a new todo but shouldn't be a big deal

website/src/models/challenge.js
paglias, line 100: The userId property can be found on all tasks and it's used to link a task to a user and in case it's missing to say that this task doesn't belong to a specific user (but to a challenge).

What we're doing here are 2 queries:
- one to fetch all the tasks belonging to a challenge and not to any user (the original challenge tasks, `chalTasks`)
- the second to fetch all tasks belonging to our user AND to the challenge.

This way we can see if any new task has been added to the challenge but not yet synced to the user (and add it to the user), deleted from the challenge but not yet synced to the user (and remove it from the user) and if any challenge task has been edited and needs to be synced to the user.

If you remove the `userId: {$exists: false}` query modified all challenge tasks will be fetched, the ones belonging to the challenge (which we use to compare to the user's tasks) but also all the challenge tasks from all the users that are members of the challenged which we don't need.

Let me know if I'm not being very clear

TheHollidayInn, line 100: Oh... eh, I get it now. I'm sorry for misunderstanding. I'll add it make and correct the tests as I should have before.

test/api/v3/unit/models/challenge.test.js
TheHollidayInn: Well, since I am wrapping the test in an `each` function, should I just do something like:

```
if (task.type === 'todo') {
   expect(length).to.be.(2); 
} else {
   expect(length).to.be.(1);
}
```

crookedneighbor: I don't see why the number of tasks is important. Shouldn't you be asserting that the array includes the specified task id?

TheHollidayInn: It isn't important. I did write assertions to compare the task. You can't check for id because the id is not the same. The task is cloned from the challenge to the user, but is not a direct reference.

TheHollidayInn:
Okay, how does this look?

paglias:
Looks good to me! @TheHollidayInn do you plan to add the unlinkChallengeFromUser method in this PR (moving it to the challenge model) or in a different one?

TheHollidayInn:
Yea, I can do that! I haven't looked at the /unlink route yet. I'll get it posted tomorrow.

test/api/v3/unit/models/challenge.test.js
```diff
@@ -0,0 +1,154 @@
...
+        await challenge.unlinkTasks(leader, 'keep-all');
+
+        let updatedLeader = await User.findOne({_id: leader._id});
+        let updatedLeadersTasks = await Tasks.Task.find({_id: { $in: updatedLeader.tasksOrder[`${taskType}s`]}});
+        let syncedTask = find(updatedLeadersTasks, function findNewTask (updatedLeadersTask) {
+          return updatedLeadersTask.type === taskValue.type && updatedLeadersTask.text === taskValue.text;
+        });
+
+        expect(syncedTask).to.exist;
+        // expect(syncedTask.challenge).to.be.empty;
```
TheHollidayInn: This is an empty object, but I am getting an error in tests:
`AssertionError: expected {} to be empty`

Mocha documentation shows {} as empty. So, I'm confused. :/

paglias: @TheHollidayInn It may be that the syncedTask object is a class and so has some hidden properties? Actually syncedTask.challenge is not a class... if you do `expect(syncedTask.challenge).to.eql({});` or Object.keys(syncedTask).length === 0

crookedneighbor: Hmmm.... `.to.be.empty` should work. I would try `.to.eql({});` (not `.to.equal({}`) and see if the assertion passes.

TheHollidayInn: Hmmm. This is what is shows:

```
AssertionError: expected {} to deeply equal {}
      + expected - actual

      -{
      -  "$__": {
      -    "_id": [undefined]
      -    "activePaths": {
      -      "paths": {
      -        "__v": "init"
      -        "_id": "init"
      -        "attribute": "init"
      -        "createdAt": "init"
      -        "notes": "init"
      -        "priority": "init"
      -        "tags": "init"
      -        "text": "init"
      -        "type": "init"
      -        "updatedAt": "init"
      -        "userId": "init"
      -        "value": "init"
      -      }
      -      "stateNames": [
      -        "require"
      -        "modify"
      -        "init"
      -        "default"
      -        "ignore"
      -      ]
      -      "states": {
      -        "default": {}
      -        "ignore": {}
      -        "init": {
      -          "__v": true
      -          "_id": true
      -          "attribute": true
      -          "createdAt": true
      -          "notes": true
      -          "priority": true
      -          "tags": true
      -          "text": true
      -          "type": true
      -          "updatedAt": true
      -          "userId": true
      -          "value": true
      -        }
      -        "modify": {}
      -        "require": {}
      -      }
      -    }
      -    "adhocPaths": [undefined]
      -    "emitter": {
      -      "_events": {}
      -      "_eventsCount": 0
      -      "_maxListeners": 0
      -      "domain": [null]
      -    }
      -    "fullPath": [undefined]
      -    "getters": {
      -      "challenge": [Circular]
      -    }
      -    "inserting": [undefined]
      -    "ownerDocument": [undefined]
      -    "populate": [undefined]
      -    "populated": [undefined]
      -    "removing": [undefined]
      -    "saveError": [undefined]
      -    "scope": {
      -      "$__": [Circular]
      -      "$__original_save": {
      -        "numAsyncPres": 1
      -      }
      -      "_doc": {
      -        "__v": 0
      -        "_id": "dd667e40-e464-4fd3-9919-f77f0d597c28"
      -        "attribute": "str"
      -        "challenge": {}
      -        "createdAt": [Date: 2016-03-04T00:23:17.163Z]
      -        "notes": ""
      -        "priority": 1
      -        "tags": []
      -        "text": "test reward"
      -        "type": "reward"
      -        "updatedAt": [Date: 2016-03-04T00:23:17.163Z]
      -        "userId": "8efdc19e-ae90-4e4c-aeb7-278a2afcf46a"
      -        "value": 0
      -      }
      -      "_posts": {
      -        "$__original_save": []
      -      }
      -      "_pres": {
      -        "$__original_save": [
      -          {
      -            "isAsync": false
      -          }
      -          {
      -            "isAsync": true
      -          }
      -          {
      -            "isAsync": false
      -          }
      -        ]
      -      }
      -      "errors": [undefined]
      -      "isNew": false
      -      "save": [Function]
      -    }
      -    "selected": [undefined]
      -    "shardval": [undefined]
      -    "strictMode": true
      -    "validationError": [undefined]
      -    "version": [undefined]
      -    "wasPopulated": false
      -  }
      -  "$__isNested": true
      -  "$__original_save": {
      -    "numAsyncPres": 1
      -  }
      -  "_doc": {
      -    "__v": 0
      -    "_id": "dd667e40-e464-4fd3-9919-f77f0d597c28"
      -    "attribute": "str"
      -    "challenge": {}
      -    "createdAt": [Date: 2016-03-04T00:23:17.163Z]
      -    "notes": ""
      -    "priority": 1
      -    "tags": []
      -    "text": "test reward"
      -    "type": "reward"
      -    "updatedAt": [Date: 2016-03-04T00:23:17.163Z]
      -    "userId": "8efdc19e-ae90-4e4c-aeb7-278a2afcf46a"
      -    "value": 0
      -  }
      -  "_posts": {
      -    "$__original_save": []
      -  }
      -  "_pres": {
      -    "$__original_save": [
      -      {
      -        "isAsync": false
      -      }
      -      {
      -        "isAsync": true
      -      }
      -      {
      -        "isAsync": false
      -      }
      -    ]
      -  }
      -  "broken": [undefined]
      -  "errors": [undefined]
      -  "id": [undefined]
      -  "isNew": false
      -  "save": [Function]
      -  "taskId": [undefined]
      -  "toJSON": [Function]
      -  "toObject": [Function]
      -  "winner": [undefined]
      -}
      +{}
```

I don't see those properties when I log to console. Are they part of Mongo?

crookedneighbor: Mongoose stuff, actually. Could you assert against the `_id` instead?

TheHollidayInn:
Sweet, thanks @crookedneighbor . I checked the _id field.

website/src/models/challenge.js
```diff
@@ -213,4 +215,37 @@ schema.methods.removeTask = async function challengeRemoveTask (task) {
...
+    let tasks = await Tasks.Task.find(findQuery).select('_id type completed').exec();
+    let taskPromises = tasks.map(task => {
+      // Remove task from user.tasksOrder and delete them
+      if (task.type !== 'todo' || !task.completed) {
+        removeFromArray(user.tasksOrder[`${task.type}s`], task._id);
+      }
+
+      return task.remove();
+    });
+    user.markModified('tasksOrder');
```
paglias, line 244: @TheHollidayInn last thing before I can merge (btw, very clean PR!): it the markModified necessary here? removeFroArray uses `array.splice` which should be proxied by mongoose

TheHollidayInn, line 244: Sorry for the delay. I believe it is. The tasksOrder array was not being updated when I removed the markModified.

paglias, line 244: Don't worry, in this case it's ok!

TheHollidayInn, line 244: Sweet. So, do I need to change anything?

paglias, line 244: Sorry forgot this PR, I'll review it once again and merge once I finish resolving the conflicts

PR #6790, opened 2016-02-29
gem gift message is always plural

murraybd:
I received the following message:

`Hello murraybd, basepair has sent you 1 gems!`

Even though it is only 1 gem, there is an 's' after gems.  This fixes that.

### DIFF
--- a/website/src/controllers/api-v2/members.js
+++ b/website/src/controllers/api-v2/members.js
@@ -38,7 +38,15 @@ api.sendMessage = function(user, member, data){
     msg = data.message
   } else {
     msg = "`Hello " + member.profile.name + ", " + user.profile.name + " has sent you ";
-    msg += (data.type=='gems') ? data.gems.amount + " gem" + ((data.gems.amount>1)?"s":"") + "!`" : shared.content.subscriptionBlocks[data.subscription.key].months + " month" + ((shared.content.subscriptionBlocks[data.subscription.key].months>1)?"s":"") + " of subscription!`";
+    if (data.type == 'gems') {
+      var gemAmount = data.gems.amount;
+      var gemLabel = gemAmount > 1 ? "gems" : "gem";
+      msg += gemAmount + " " + gemLabel + "!`";
+    } else {
+      var monthAmount = shared.content.subscriptionBlocks[data.subscription.key].months;
+      var monthLabel = monthAmount > 1 ? "months" : "month";
+      msg += monthAmount + " " + monthLabel + " of subscription!`";
+    }
     msg += data.message;
   }
   shared.refPush(member.inbox.messages, groups.chatDefaults(msg, user));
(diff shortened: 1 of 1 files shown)


website/src/controllers/api-v2/members.js
```diff
@@ -38,7 +38,7 @@ api.sendMessage = function(user, member, data){
     msg = data.message
   } else {
     msg = "`Hello " + member.profile.name + ", " + user.profile.name + " has sent you ";
-    msg += (data.type=='gems') ? data.gems.amount + " gems!`" : shared.content.subscriptionBlocks[data.subscription.key].months + " months of subscription!`";
+    msg += (data.type=='gems') ? data.gems.amount + " gem" + ((data.gems.amount>1)?"s":"") + "!`" : shared.content.subscriptionBlocks[data.subscription.key].months + " month" + ((shared.content.subscriptionBlocks[data.subscription.key].months>1)?"s":"") + " of subscription!`";
```
crookedneighbor: Let's take this as an opportunity to clean up this code a little. 

These lines should probably also be translatable. But in this case, we'd need to make sure that it used the receiver's language, not the sender's.

Alys: When gems are gifted, the message appears in the sender's mailbox as well as the receiver's, so the sender needs to be able to read it, so sometimes two languages would be needed.

I'm wondering if both players should see the message in both languages, so that they still get identical copies.

I think the translation should be done in a separate issue since it's not trivial.

Alys: @crookedneighbor What code cleanup would you like here?

crookedneighbor: Mostly we're using a multiple ternaries on this line, which makes it very very difficult to read. I'd advocate for a more verbose `if {} else {}` logic. IE:

``` js
if (data.type === 'gems') {
  var gemAmount = data.gems.amount;
  var gemLabel = gemAmount > 1 ? "gems" : "gem";
  msg += gemAmount + " " + gemLabel + "!";
} else {
  var monthAmount = shared.content.subscriptionBlocks[data.subscription.key].months;
  var monthLabel = monthAmount > 1 ? 'months' : 'month';
  msg += months + " " + monthLabel + " of subscription!`"
}
```

IMO, that's much easier to parse than what we currently have.

colwem: And do we or should we have any style rules about long lines.  I'm really not a fan of having to scroll to read the code.  I always stick to 80 columns myself.

crookedneighbor:
Merged in with https://github.com/HabitRPG/habitrpg/commit/eaea55687b5c525a026df4ccf1c17bdc91daa131

murraybd:
Thanks!  My User ID is 5b93ce83-2508-4ea4-9647-079899dffa96.

crookedneighbor:

Issue #6783, opened 2016-02-28
Change Menu Names


lemoness:
We have a few changes that we need to make to the names of items on the drop-down menus, for the sake of clarity.

Here is a list of changes:
Avatar > Customize Avatar
Stats > Avatar Stats
Tavern > Tavern Chat
Hall > Hall of Heroes

This is probably suitable for an Aspiring Coder.

keysmashes:
NB 'Hall' contains both the Hall of Heroes and the Hall of Patrons; the UI starts to look a bit weird (IMHO) when you have something like this:

![image](https://cloud.githubusercontent.com/assets/9433472/13380034/65e8a670-de2f-11e5-869b-0398009579f5.png)

I don't really have any ideas for how to make it better, apart from making the Hall of Patrons a top-level section:

![image](https://cloud.githubusercontent.com/assets/9433472/13380106/ec43331e-de31-11e5-9981-82f4e9be3ffa.png)

![image](https://cloud.githubusercontent.com/assets/9433472/13380109/f6910ca6-de31-11e5-9126-37b2a78e4a71.png)

I'm not sure if I think that looks better or worse.

Alys:
pinging @lemoness

nrbunn:
I agree the duplicate hall of heroes made it look weird. Another option would to be change some wording if you want to keep format. 
![menu_rename](https://cloud.githubusercontent.com/assets/8887847/13382253/83d784d4-de35-11e5-8d28-2c1478178dec.PNG)

querry43:
This might be an opportunity for something new.  The Halls are kind of a dead end that most players would never visit, right?  I wonder if there is a way to build more community around it.

What if the hall was sorted not by tier but by the most recent contribution.  That would make it quite a bit more dynamic.  Then we could perhaps involve the community with "thank you" buttons that would send thank you card to the contributor.  I don't know about others, but I like to get feedback as an opportunity for improvement.

Since getting my badge, several people have asked about what they could do.  I get the impression that some additional dialog could energize parts of the community to contribute.  Aside from a little design work, including menu adjustments, this doesn't seem like much work given the existing thank you card mechanism.

lemoness:
Why don't we call the top menu "Hall of Heroes" and then split it into "Hall of Contributors" and "Hall of Patrons."

[...]

PR #6793, opened 2016-03-01
Change Menu Names 

nrbunn:
Changed menu names based off requirements of  #6783


Alys:
Thank you, @Sinble! What's your User ID from Settings > API?

nrbunn:
25479d8c-45e1-482c-877e-6af2ccaba643

Alys:
Tier awarded! Congratulations! :)

taldin:
Grats on your new tier!  :D

PR #6794, opened 2016-03-01
[Api v3] Cast

paglias:
WIP
- [ ] tests
- [x] challenge tasks


website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +39,98 @@ api.getUser = {
   },
 };
 
+let partyMembersFields = 'profile.name stats achievements items.special';
```
crookedneighbor: This should be a const

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +39,98 @@ api.getUser = {
...
+    let spellId = req.params.spellId;
+
+    // optional because not required by all targetTypes, presence is checked later if necessary
+    req.checkQuery('targetId', res.t('targetIdUUID')).optional().isUUID();
+    req.checkQuery('targetType', res.t('invalidTargetType')).isIn(['task', 'self', 'user', 'party']);
+
+    let reqValidationErrors = req.validationErrors();
+    if (reqValidationErrors) throw reqValidationErrors;
+
+    let klass = common.content.spells.special[spellId] ? 'special' : user.stats.class;
```
crookedneighbor, line 71: Would now be a good time to discuss separating out the proper spells from spooky sparkle-type items and greeting cards? I think we could abstract most of the shared logic into a separate function, and I see no reason to keep them under the same route.

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +39,98 @@ api.getUser = {
...
+    let spell = common.content.spells[klass][spellId];
+
+    if (!spell) throw new NotFound(res.t('spellNotFound', {spell: req.params.spellId}));
+    if (spell.mana > user.stats.mp) throw new BadRequest(res.t('notEnoughMana'));
+
+    if (targetType === 'task') {
+      if (!targetId) throw new BadRequest(res.t('targetIdUUID'));
+
+      // TODO what about challenge tasks? should casting be disabled on them?
```
crookedneighbor, line 81: I think spells should still apply to challenge tasks, but I'd love to hear @Alys's thoughts on them. 

The only thing I can think of is challenges that depend on task value for determining a winner. That could be tricky.

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +39,98 @@ api.getUser = {
...
+ *
+ * @apiSuccess {Object|Array} mixed Will return the modified targets. For party members only the necessary fields will be populated.
+ */
+api.castSpell = {
+  method: 'POST',
+  middlewares: [authWithHeaders(), cron],
+  url: '/user/class/cast/:spell',
+  async handler (req, res) {
+    let user = res.locals.user;
+    let targetType = req.query.targetType;
```
crookedneighbor: Why are we using query params in a post request?

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +39,98 @@ api.getUser = {
...
+
+      // TODO what about challenge tasks? should casting be disabled on them?
+      let task = await Tasks.Task.findOne({
+        _id: targetId,
+        userId: user._id,
+      }).exec();
+      if (!task) throw new NotFound(res.t('taskNotFound'));
+
+      spell.cast(user, task);
+      let savedTask = await task.save();
```
crookedneighbor: I think you can just do `await task.save();` because that changes the object in place.

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +39,98 @@ api.getUser = {
...
+          partyMembers = await User.findOne({_id: targetId, 'party._id': party._id}).select(partyMembersFields).exec();
+        }
+
+        if (!partyMembers) throw new NotFound(res.t('userWithIDNotFound', {userId: targetId}));
+        spell.cast(user, partyMembers);
+        savedPartyMembers = await partyMembers.save();
+      }
+      res.respond(200, savedPartyMembers);
+
+      // TODO save group concurrently with users?
```
crookedneighbor: I think it's fine to do it after the response is sent.

website/src/controllers/api-v3/user.js
paglias: good point, I copied the api v2 structure but it doesn't make sense

website/src/controllers/api-v3/user.js
paglias: :+1:

website/src/controllers/api-v3/user.js
paglias, line 81: yeah my concern was the task value, will wait for @Alys

website/src/controllers/api-v3/user.js
paglias: :+1:

website/src/controllers/api-v3/user.js
paglias: :+1:

website/src/controllers/api-v3/user.js
paglias: actually I'm not sure about this... I think a query parameter is easier to send

paglias: In the newest commit I removed targetType since we store it in the spell object. So at this point for targetId only it doesn't make sense for me to use req.body

website/src/controllers/api-v3/user.js
paglias, line 71: It's fine for me

paglias:
Updated

website/src/controllers/api-v3/user.js
Alys, line 81: Skills shouldn't apply to challenge tasks because it makes it harder for challenge owners to judge winners.

We'll get repeated newbies questions about why they can't cast a skill on some tasks, but we'll just have to handle them. I think the challenge owners are more important here.

It would be nice if casting on a challenge task produced an informative message rather than just not working - that would avoid questions.

Note though that none of this would have to be done for v3 go-live. I'd be happy to see just the question marks removed from that comment and a note added about the error message.

common/script/content/spells.js
```diff
@@ -0,0 +1,499 @@
+import t from './translation';
```
crookedneighbor, line 1: Any reason moving the spells into their own content module shouldn't happen in `develop`?

common/script/content/spells.js
```diff
@@ -0,0 +1,499 @@
...
+let spells = {};
+
+spells.wizard = {
+  fireball: {
+    text: t('spellWizardFireballText'),
+    mana: 10,
+    lvl: 11,
+    target: 'task',
+    notes: t('spellWizardFireballNotes'),
+    cast (user, target) {
```
crookedneighbor, line 43: We should have tests around each spell.

paglias, line 43: yep, wasn't sure if we already had them in the common tests

paglias, line 43: but for now i'm going to assume the cast functions are working and postpone the tests for each single spell

common/script/content/spells.js
paglias, line 1: no reason. I was ready to do it directly in develop then I had to make a change for https://github.com/HabitRPG/habitrpg/pull/6794/files#diff-24cac23322db42dce4ea110a133c9ea3R221 so I did it into this branch. I'll make sure to move the changes to develop before merging

website/src/controllers/api-v3/user.js
paglias, line 81: I'm going to remove challenge tasks and provide an useful error message tomorrow

PR #6802, opened 2016-03-02
[Api v3] Cast

paglias:
Missing tests

### DIFF
--- a/common/locales/en/api-v3.json
+++ b/common/locales/en/api-v3.json
@@ -73,7 +73,7 @@
   "guildQuestsNotSupported": "Guilds cannot be invited on quests.",
   "questNotFound": "Quest \"<%= key %>\" not found.",
   "questNotOwned": "You don't own that quest scroll.",
-  "questLevelTooHigh": "You must be Level <%= level %> to begin this quest.",
+  "questLevelTooHigh": "You must be level <%= level %> to begin this quest.",
   "questAlreadyUnderway": "Your party is already on a quest. Try again when the current quest has ended.",
   "questAlreadyAccepted": "You already accepted the quest invitation.",
   "noActiveQuestToLeave": "No active quest to leave",
@@ -90,5 +90,11 @@
   "noAdminAccess": "You don't have admin access.",
   "pageMustBeNumber": "req.query.page must be a number",
   "missingUnsubscriptionCode": "Missing unsubscription code.",
-  "userNotFound": "User not Found"
+  "userNotFound": "User not found.",
+  "spellNotFound": "Spell \"<%= spellId %>\" not found.",
+  "partyNotFound": "Party not found",
+  "targetIdUUID": "\"targetId\" must be a valid UUID.",
+  "challengeTasksNoCast": "Casting a spell on challenge tasks is not supported.",
+  "spellNotOwned": "You don't own this spell.",
+  "spellLevelTooHigh": "You must be level <%= level %> to use this spell."
 }
--- a/common/script/api-v3/errors.js
+++ b/common/script/api-v3/errors.js
@@ -1,9 +1,37 @@
-import CustomError from './customError';
+// Base class for custom application errors
+// It extends Error and capture the stack trace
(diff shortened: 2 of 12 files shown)


common/script/api-v3/errors.js
```diff
@@ -0,0 +1,9 @@
+import CustomError from './customError';
+
+export class NotAuthorized extends CustomError {
```
paglias, line 12: @crookedneighbor what if we used the errors with the httpCode in shared code too? I know it may seem a little weird but it becomes simpler to use them from the API, especially if they're more or less the same: NotAuthorized, NotFound and BadRequest as a catch all for the rest

Otherwise I can make a map of shared errors to http codes

crookedneighbor, line 12: I think it's fine to use the http code errors. It's the website that's using the shared code, so we can just ignore the http code part of it.

common/locales/en/api-v3.json
```diff
@@ -90,5 +90,11 @@
   "noAdminAccess": "You don't have admin access.",
   "pageMustBeNumber": "req.query.page must be a number",
   "missingUnsubscriptionCode": "Missing unsubscription code.",
-  "userNotFound": "User not Found"
+  "userNotFound": "User not found.",
+  "spellNotFound": "Spell \"<%= spellId %>\" not found.",
+  "partyNotFound": "Party not found",
+  "targetIdUUID": "\"targetId\" must be a valid UUID.",
+  "challengeTasksNoCast": "Casting a spell on challenge tasks is not supported.",
```
crookedneighbor, line 97: This seems a little transactional to me.  @Alys @lemoness thoughts on better wording?

paglias, line 97: all these strings will require some review before we deploy, I'm not very good at writing them :)

test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
```diff
@@ -0,0 +1,172 @@
...
+    await groupLeader.post(`/user/class/cast/earth`);
+    await sleep(1);
+    await group.sync();
+    expect(group.chat[0]).to.exists;
+    expect(group.chat[0].uuid).to.equal('system');
+  });
+
+  // TODO find a way to have sinon working in integration tests
+  // it doesn't work when tests are running separately from server
+  it('passes correct target to spell when targetType === \'task\'');
```
paglias, line 165: @crookedneighbor the PR is ready, the only thing that bother me is that we can't have these tests...

The only idea is to substitute them with ones that call spell.cast for real, one for each type of targetType

lemoness:
I've tried to put together some strings for these, but in some cases I
will need more context about when/where they will display. In
particular, if I've guessed about something and it's incorrect, let me
know so I can change it.

"You don't have admin access." >

"Whoops! You can't see this because you don't have admin access."
(Assuming that the user is trying to access something like granting
contributor levels.)
,
"Missing unsubscription code.", >

"It looks like we're missing your unsubscription code! Email
admin@habitica.com and we will help you." (Assuming this is the error
that displays that prevents people from unsubscribing.)

,
"Party not found" >

"Yikes! That party doesn't exist." (I'd like to have some helpful
instructions here, but I am not sure what they should be. For example,
are they entering the party ID incorrectly?)

 "\"targetId\" must be a valid UUID." >

 "Whoops, looks like you didn't enter a valid UUID! Be sure to enter
the long string of numbers and letters that serves as a Unique User
ID, not a username." (Assuming that it displays when users enter an
incorrect UUID, for example when inviting people to a Party. Also, do
these only display on the website? If so, we can get more specific in
our instructions.)

"Casting a spell on challenge tasks is not supported." >

"Hey, you can't use a Skill on a Challenge task! Try casting it on a
normal task instead."

---

I need more context about the following strings before I can rewrite them:

"User not Found" (first version)

"User not found." (second version)

"req.query.page must be a number"

"Spell \"<%= spellId %>\" not found."

Issue #6758, opened 2016-02-23
Leaving group button bug : pressing leave multiple times causes multiple confirm/cancel popups


lalx:
i miss-click on "leave group" button, so i clicked to cancel it, but the message dont dessapeared, and appaear another message ahead (sorry for bad english)

Group ID: 50f8c890-5959-41e2-a35d-6111474a72a2
Lalx ID: a0ade8da-16dc-41a1-8521-742d5b6098c8
TheLalx ID: 6be384fc-f6af-4f07-ab7e-eb572c16d565

browser: google chrome
operating system: windows 8

![bug habitica](https://cloud.githubusercontent.com/assets/17428451/13249813/3b9f3236-da05-11e5-93b4-538c4b955abe.png)

Alys:
I haven't been able to replicate this, although I'm using Chrome on Mac OS, not Windows.

@lalx Does the problem still happen if you try to do it again? If so, does reloading the page make the problem stop?

crookedneighbor:
I was able to reproduce this bug by pressing the "leave" button multiple times without choosing confirm or cancel.

Alys:
It must be browser-specific. That still doesn't happen in Chrome on Mac - I can see the popup flicker briefly as the old one is closed and the new one open to replace it.

I would guess that this happens in other buttons around the site, so it would be great to have a fix for this that was applied site-wide.

@lalx When you tried to click cancel, is it possible that you misclicked again on the leave button? Maybe there's a problem with your mouse that causes misclicks like that?

lalx:
yes, it happens again and again, every time i reload the page 

when i click Sair(Leave), appaears the correct popup with the message "Deixar Equipe"("leave team") and its ok there
but sometimes appaears the popup without the message, and the bug occurs when that happens

![bug habitica2](https://cloud.githubusercontent.com/assets/17428451/13396015/fd16e1e6-ded0-11e5-9234-765924504d00.png)

lalx:
![bug habitica3](https://cloud.githubusercontent.com/assets/17428451/13396360/e31748c4-ded2-11e5-9b5b-37b9bb64c67f.png)

lalx:
![bug habitica4](https://cloud.githubusercontent.com/assets/17428451/13396602/4fd39246-ded4-11e5-8d47-b19fbdeaddc9.png)

lalx:
i think there is 2 button with 2 different popups with the same function, and when i click on both, occurs a bug.
[...]

PR #6805, opened 2016-03-02
Fixed issue #6758  - prevent multiple confirm/cancel popups from group Leave button

ltorroba:
Closes HabitRPG/habitrpg#6758

Problem was caused by allowing the span element, containing the
glyphicon in the "Leave" button of the Party page, to add the popup
inside the a element, instead of outside it. This lead to the
possibility of having multiple popups simultaneously, which was not
originally intended.


website/public/js/controllers/partyCtrl.js
```diff
@@ -51,6 +51,11 @@ habitrpg.controller("PartyCtrl", ['$rootScope','$scope','Groups','Chat','User','
           Analytics.track({'hitType':'event','eventCategory':'button','eventAction':'click','eventLabel':'Leave Party'});
           $scope.selectedGroup = group;
           $scope.popoverEl = $($event.target);
+
+          if ($scope.popoverEl[0].localName == "span") {
+              $scope.popoverEl = $($scope.popoverEl[0].parentNode);
+          }
```
colwem: Instead of checking the tag can we check the el has the `.btn` class. I think this would make it a lot more obvious what the point of this block of code is.

crookedneighbor: I don't think checking the `.btn` class will work in this instance, since the span portion does not have a button class. I agree that this isn't exactly the correct solution, but it's on the right track. I'll need to investigate this a little and see if there's a more meaningful way of accomplishing this. Especially in a way that won't break if we change that span to a div :)

One other thing. Do we need to do the same thing in the guild controller?

ltorroba: Alright - sorry, I'm quite new to this! From what I've tested, this indeed also happens when clicking the leave button for the guilds:



There's a TODO on the file suggesting the function to be refactored and used for both leaving parties and guilds. I suppose this would be a great time to do that - I'd be happy to do it but knowing so little about the structure of the code, I'd need some guidance on where the new method should be placed (is there a global JS file for this sort of thing?).

I went ahead and tested what @colwem suggested, along with some better code, and I came up with this - seems a lot better to me:

```
if (!$scope.popoverEl.hasClass("btn")) {
    $scope.popoverEl = $scope.popoverEl.parent();
}
```

What do you think?

colwem: You could just do

```
$scope.popoverEl = $($event.target).closest('.btn');
```

`closest` starts at the element and searches up the tree till it finds a match.  In this case it will just be the el itself or it's parent. I tried it out and it seems to work.

crookedneighbor:
Did you intend to close this?

ltorroba:
Sorry - didn't mean to. Tried to fix the code and seems I deleted the original pull request. I opened a new one with @colwem's suggested solution here: #6819.

Issue #68, opened 2013-01-15
Export Data


ghost:
The ability to export your data maybe be able to show it to a doctor for the habit of tracking blood pressure would be really useful.

silverymoon:
I would love that option as well.

ghost:
So would my doctor.

yene:
+1 for this option for when the website is unstable. I would love export format to be a simple plain text.

ghost:
I was thinking exporting to an excel spreadsheet

Andrew565:
It would be relatively straight-forward to support multiple output formats. Plain text and HTML could be done with a simple template, just like creating a standard view. JSON is pretty-much a no-brainer to output too.

For excel, you can use [jQuery-CSV](https://code.google.com/p/jquery-csv/) to output to CSV, which is usable in Excel.

Once support for these has been prototyped, could wrap in an API.

hookang:
Added to Master Feature List #169. Closed for Organization.

wc8:
Vote for [Import/Export request here](https://trello.com/c/29qEE1DF)

PR #6806, opened 2016-03-02
[API v3] Change email

piousbox:
Running this command:

`vagrant@vagrant-ubuntu-trusty-64:~/projects/habitrpg$ mocha test/api/v3/integration/emails/POST-email-update.test.js`

I see that in test, the user has the expected empty auth.local:

```
+++ +++ TEST fbUser is ApiUser {
  _id: '367c745a-7b2c-417c-b53a-e5ade140855f',
```

whereas on the server run with `npm start`, the user with the same _id has an unexpected auth.local:

```
+++ +++ user._id is 367c745a-7b2c-417c-b53a-e5ade140855f
+++ +++ user.auth.local is { hashed_password: '238c1160afd5ef4fd8d7a2344caacc3ab19c02c6',
  salt: '280bb7da92',
  email: '50ad1b97-d7c9-4d30-9787-20445ea28d5c@example.com',
  lowerCaseUsername: '50ad1b97-d7c9-4d30-9787-20445ea28d5c',
  username: '50ad1b97-d7c9-4d30-9787-20445ea28d5c' }

```


.gitignore
```diff
@@ -32,3 +32,7 @@ common/dist/sprites/habitrpg-shared.css
 .elasticbeanstalk/config.yml
 
 test/spec/mocks/translations.js
+
+# emacs
```
crookedneighbor: These should be in your global gitignore, not the project.

test/api/v3/integration/emails/POST-email-update.test.js
```diff
@@ -0,0 +1,86 @@
...
+    });
+  });
+*/
+  describe("facebook user", async () => {
+
+    beforeEach(async () => {
+      fbUser = await generateUser();
+      await fbUser.update({"auth.local": {ok:true} });
+      console.log("+++ +++ TEST fbUser is", fbUser);
+      fbUser = new ApiUser( fbUser );
```
paglias: why this line? It's not necessary, generateUser at line #68 will take care of everything

paglias: I think the problem is that you're basically rebuilding the user on line 71 so it gets the default local authentication. If you remove line 71 it should work

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +38,45 @@ api.getUser = {
...
+
+    if (!user.auth.local.email) throw new PreconditionFailed(res.t('userHasNoLocalRegistration'));
+
+    req.checkBody('newEmail', res.t('newEmailRequired')).notEmpty().isEmail();
+    req.checkBody('password', res.t('missingPassword')).notEmpty();
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+ 
+    // check password
+    let temp_user = await User.findOne({ _id: user._id}, {"auth.local": 1}).exec();
```
paglias: we use single quotes for strings, this will fail when you run `gulp lint`

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +38,45 @@ api.getUser = {
...
+    let candidate_password = passwordUtils.encrypt(req.body.password, temp_user.auth.local.salt);
+    if (candidate_password !== temp_user.auth.local.hashed_password) throw new BadRequest();
```
paglias: NotAuthorized here? Also I'd supply an useful message like for the password not matching

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +38,45 @@ api.getUser = {
...
+ * @apiGroup User
+ *
+ * @apiSuccess {Object} { status: 'ok' }
+ **/
+api.updateEmail = {
+  method: 'POST',
+  middlewares: [authWithHeaders(), cron],
+  url: '/email/update',
+  async handler (req, res) {    
+    let user = res.locals.user.toJSON();
```
paglias: why toJSON?

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +38,45 @@ api.getUser = {
...
+    
+    console.log("+++ +++ user._id is", user._id);
+    user = await User.findOne({ _id: user._id }).exec();
```
paglias: the user is already loaded and stored in the user variable

website/src/controllers/api-v3/user.js
paglias: you're reloading the user for the second time here

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +38,45 @@ api.getUser = {
...
+    if (validationErrors) throw validationErrors;
+ 
+    // check password
+    let temp_user = await User.findOne({ _id: user._id}, {"auth.local": 1}).exec();
+    let candidate_password = passwordUtils.encrypt(req.body.password, temp_user.auth.local.salt);
+    if (candidate_password !== temp_user.auth.local.hashed_password) throw new BadRequest();
+    
+    // save new email
+    temp_user.auth.local.email = req.body.newEmail;
+    let userUpdated = await User.update({ _id: temp_user._id },
```
paglias: User.update doesn't return the updated user, it just sends an update command to mongoose. You should use `user.save()` that will save any change you made to the user object automatically. You may want to give a look at Mongoose's docs http://mongoosejs.com/docs/guide.html

test/api/v3/integration/emails/POST-email-update.test.js
piousbox: I removed this line and still the same behavior of the test.

test/api/v3/integration/emails/POST-email-update.test.js
```diff
@@ -0,0 +1,85 @@
...
+    });
+  });
+*/
+  describe("facebook user", async () => {
+
+    beforeEach(async () => {
+      fbUser = await generateUser();
+      await fbUser.update({"auth.local": {ok:true} });
+      console.log("+++ +++ TEST fbUser is", fbUser);
+    });
```
piousbox: without re-creating fbUser the problem is the same.

paglias: you mean that line 70 outputs a user with an email?

website/src/controllers/api-v3/user.js
piousbox: Yes... I specifically would like to make the authentication errors not precise. If password doesn't match, I'd like the error to be the same as username not found (because otherwise you can run an attack to find all the usernames). That's why authentication errors can be made less clear.

paglias: yeah but consider that this route requires authentication so we have to assume the requester has already access to all the user info (except the password because it's not stored in the database)

website/src/controllers/api-v3/user.js
piousbox: Yes... I think, specifically hashed_password isn't available in res.local.user so I reload the user here.

paglias: If you didn't call toJSON at the beginning (which is meant to produce a public version of the user document and so remove some sensitive data) all fields would be available without the need to reload the user ever

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +37,39 @@ api.getUser = {
...
+ * @apiGroup User
+ *
+ * @apiSuccess {Object} { status: 'ok' }
+ **/
+api.updateEmail = {
+  method: 'POST',
+  middlewares: [authWithHeaders(), cron],
+  url: '/user/update-email',
+  async handler (req, res) {
+    let user = res.locals.user.toJSON();
```
paglias: @piousbox you should not call toJSON here but just do `let user = res.locals.user;`. This way `user` will be a mongoose document that you can `.save()`, `.remove()`, ... `.toJSON()` instead convert the document to a plain javascript object ready to be used as a response and will remove some private fields like `hashed_password`.

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +37,39 @@ api.getUser = {
...
+    if (!user.auth.local.email) throw new PreconditionFailed(res.t('userHasNoLocalRegistration'));
+
+    req.checkBody('newEmail', res.t('newEmailRequired')).notEmpty().isEmail();
+    req.checkBody('password', res.t('missingPassword')).notEmpty();
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    // check password
+    let candidatePassword = passwordUtils.encrypt(req.body.password, user.auth.local.salt);
+    if (candidatePassword !== user.auth.local.hashed_password) throw new NotAuthorized();
```
paglias: I still think here we should return an informative message.

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +37,39 @@ api.getUser = {
...
+
+    // save new email
+    user.auth.local.email = req.body.newEmail;
+    let flag = await user.save();
+    if (!flag) throw new BadRequest();
```
paglias: `user.save` is a promise, if an error happens it'll throw an error so there's no need to check if the result is ok or not, it's handled automatically

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +37,39 @@ api.getUser = {
...
+
+    return res.respond(200, { status: 'ok' });
```
paglias: Here I'd return the new email: `{email: user.auth.local.email}`

website/src/controllers/api-v3/user.js
paglias: Also you don't have to assign the result to a new variable as the original `user` will already have the updated fields

website/src/controllers/api-v3/user.js
```diff
@@ -31,4 +37,38 @@ api.getUser = {
...
+
+    // check password
+    let candidatePassword = passwordUtils.encrypt(req.body.password, user.auth.local.salt);
+    if (candidatePassword !== user.auth.local.hashed_password) throw new NotAuthorized(res.t('wrongPassword'));
+
+    // save new email
+    user.auth.local.email = req.body.newEmail;
+    await user.save();
+
+    return res.respond(200, { status: 'ok', email: user.auth.local.email });
```
paglias: @piousbox can you remove the `status: ok` part? We have discussed adding a success: false|true property but if we add it it'll be done automatically

test/api/v3/integration/user/POST-user-update-email.test.js
```diff
@@ -0,0 +1,77 @@
...
+        message: t('wrongPassword'),
+      });
+    });
+
+    it('changes email if new email and existing password are provided', async () => {
+      let response = await user.post(endpoint, {
+        newEmail,
+        password: thePassword,
+      });
+      expect(JSON.stringify(response)).to.equal(JSON.stringify({ status: 'ok', email: 'some-new-email_2@example.net' }));
```
paglias: here you should use `expect(response).eql(...expected result...)`. `.eql()` is used in place of `.equal()` when comparing an array/object that can't be compared using `===`

PR #6811, opened 2016-03-03
chore: upgrade babel to v6

crookedneighbor:
This is part 1, the upgrade process.

Part 2 is to pre-compile the server code before the server runs. It's pretty tricky though, as the common code needs to be pre-compiled as well, which is a separate process from the browserify step.


paglias:
@crookedneighbor I would precompile the server only in production, locally we can still run it normally

crookedneighbor:
Makes sense.

common/script/content/appearance/backgrounds.js
```diff
@@ -320,4 +320,4 @@ forOwn(backgrounds, function prefillBackgroundSet (value) {
   });
 });
 
-export default backgrounds;
+module.exports = backgrounds;
```
crookedneighbor, line 323: Babel changed how it treats exports from v5 to v6.

``` js
// babel 5

// default is equivalent with module.exports
export default backgrounds;
module.exports = backgrounds;

import backgrounds from './backgrounds';
let backgrounds = require('./backgrounds');

// also equivalent
export foo;
export bar;
module.exports.foo = foo;
module.exports.bar = foo;

import { foo, bar } from './foobar';
let { foo, bar } = require('./foobar');

// also equivalent
export default {
  foo,
  bar,
}
module.exports = {
  foo,
  bar,
}

import { foo, bar } from './foobar';
let { foo, bar } = require('./foobar');
```

``` js
// babel 6

// default is not equivalent with module.exports
export default backgrounds;
module.exports = {
  default: backgrounds,
};

import backgrounds from './backgrounds';
let backgrounds = require('./backgrounds').default;

// these are the same as before, I believe
export foo;
export bar;
module.exports.foo = foo;
module.exports.bar = foo;

import { foo, bar } from './foobar';
let { foo, bar } = require('./foobar');

// this is not the same
export default {
  foo,
  bar,
}
module.exports = {
  default: {
    foo,
    bar,
  },
}

import foobar from './foobar';
import { foo, bar } from './foobar'; // this will not work
let { foo, bar } = require('./foobar').default;
```

I chose to go with `module.exports` instead of `export default`, because it was easier for me to wrap my head around (and required fewer changes to files). It's a little weird to mix import syntax with module.exports, but as long as we're consistent, I think it's okay.

crookedneighbor:
Did part 2. It's kind of weird and hacky, but it does the job.

.gitignore
```diff
@@ -1,6 +1,8 @@
 .DS_Store
 website/public/gen
 website/public/common
+website/transpiled-babel/
```
crookedneighbor, line 4: Are these good names for these folders?

PR #6819, opened 2016-03-04
Closed issue #6758 - prevent multiple confirm/cancel popups from group Leave button

ltorroba:
After some discussion on the thread, this is probably the neatest
solution to issue #6758. I have also modified the code on the guild
controller, which displayed the very same problem when pressing the
'Leave' button to leave a guild.

### DIFF
--- a/website/public/js/controllers/guildsCtrl.js
+++ b/website/public/js/controllers/guildsCtrl.js
@@ -54,7 +54,7 @@ habitrpg.controller("GuildsCtrl", ['$scope', 'Groups', 'User', 'Challenges', '$r
 
       $scope.clickLeave = function(group, $event){
           $scope.selectedGroup = group;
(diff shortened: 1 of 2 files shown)


Alys:
Thanks @zesme ! This will count towards a contributor tier. What's your User ID from Settings > API?

I really liked your description of the problem in the issue. It made it much easier to work out how to replicate the problem and hence confirm that this fixed it.

ltorroba:
No worries @Alys! My User ID is: 8ac1be04-4201-4eed-a2a4-436eada5c0e9.

Got quite a lot of free time lately, so I'll see if I can help around getting more of these issues before I tackle some of the bigger ones. Feel free to tag me if you find some boring bug that requires fixing!

Alys:
Best offer ever. :)  We will certainly keep you in mind!

I've given you your first tier. Congratulations!

PR #6821, opened 2016-03-04
Upgrade babel v3

crookedneighbor:
Merges in the Babel 6 upgrade from `develop` and changes `export default` to use `module.exports`

The way we're importing and exporting the error codes will not work with this setup and still need to be fixed. 

@paglias


paglias:
Going to finish this and merge

PR #6828, opened 2016-03-05
chore: Upgrade node

crookedneighbor:


colwem:
So should we be using nvm?  It makes sense but there's no mention of this in the wiki instructions for setting up your env.

paglias:
@colwem it's not necessary but many in the team use it because it makes installing and switching between node versions very easy

colwem:
@paglias should the app only be run with 0.10.xx or can we use any version?  Because I was using my systems 5.1.xx for a couple days and it seemed to be working.

paglias:
@colwem we're using 0.10 in production and hopefully we'll switch to 4.x in a few days. So _officially_ we advise to run Habitica on 0.10.x but we've encountered no problems so far with 4.x except for a few native modules falling back to JS versions because they aren't compiled for 4.x (or any version > 0.10). But this shouldn't affect functionalities

colwem:
@paglias 

> except for a few native modules falling back to JS versions because they aren't compiled for 4.x

Is that what causes the bson error I always get?

paglias:
@colwem exactly, but don't worry about it, it shouldn't be a big performance hit

PR #6834, opened 2016-03-07
Split shared ops

paglias:
WIP

Todo:
- [x] review all ops after clearPMs
- [x] check user is always the first argument
- [x] run tests

cc @SabreCat @crookedneighbor


SabreCat:
There are a few things I'd love to do here (e.g. rename `reroll` to `fortify`), but given this is strictly a splitting step with no additional changes, this looks OK to me. 

I'll delete my extant refactoring PR, looks like that won't be happening just yet.

paglias:
@SabreCat I tried to keep this as simple as possible so we can do the real refactor in v3 while not getting crazy every time a file change in shared

Issue #6679, opened 2016-02-08
Bottom and right of XP graph is clipped


keysmashes:
Adding a few more pixels to the SVG element's height/width fixes it, but also makes it overlap with the tags; this is fixable by increasing the size of the (grand)parent div, but that seems like a nasty solution. Anyone know how the SVGs work? (i.e. how they're generated, why the height gets mangled)

![image](https://cloud.githubusercontent.com/assets/9433472/12893727/bf142b3e-ce8a-11e5-80f0-60acd0413bfc.png)

ltorroba:
Seems like this is a problem to do with the chartArea option on the `toggleChart` function of `website/public/js/controllers/rootCtrl.js`. I'll be doing some tests tomorrow and see what can be done about it.

ltorroba:
So I did some tests and this boils down to the fact that the size of the container object of the SVG, defined through the `width` and `height` parameters, is not sufficiently big enough to contain the chart, the legend and the axes.

The fix to this issue was simply a matter of adjusting the `chartArea` property so that the legend and axes are given a fixed, minimum space to be drawn, and then the chart is drawn in the remaining space.

I've addressed this issue as part of a series of fixes when it came to the problems that came up when drawing charts. Here's a preview of the fix, running on Safari @ Mac OS X:

New alignment, 1000’s (red background for illustration):
![](http://i.imgur.com/3ckQEq7.png)

New alignment, decimals (red background for illustration):
![](http://i.imgur.com/KA1J6wt.png)

SabreCat:
Client issue. Pending for frontend redesign.

librarianmage:
Should this be closed and considered during re-implementation?

SabreCat:
Yes, we don't currently have the graphs at all, so this issue isn't in evidence.

PR #6836, opened 2016-03-08
Improvements to the display of the experience, todo and task charts

ltorroba:
1) Fixes issue #6679, which referred to the EXP chart getting cropped.
2) Added some bottom padding to the EXP chart so that the tags are
further apart from the chart itself (requires rebuilding stylus)
3) Added event handlers for the dynamic resizing of the charts. The
event handlers use the same data every time and are only created when a
chart is first toggled, as to reduce overhead.
4) Moved the location of the tasks chart into the .task-text, so that
the habit +/- buttons are now fully triggerable.

I’ll add more details on the implementation below soon.

### DIFF
--- a/website/client-old/css/tasks.styl
+++ b/website/client-old/css/tasks.styl
@@ -705,3 +705,7 @@ form
     background-color:inherit !important
   .task-checker label:hover:after
     content: ''
+
+// Add padding to exp-chart
+.exp-chart
+  padding-bottom: 10px
--- a/website/client-old/js/controllers/rootCtrl.js
+++ b/website/client-old/js/controllers/rootCtrl.js
@@ -187,7 +187,13 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
 
     $rootScope.charts = {};
     $rootScope.toggleChart = function(id, task) {
-      var history = [], matrix, data, chart, options;
+      if($rootScope.charts[id] === undefined) {
+        $(window).resize(function() { 
+          $rootScope.drawChart(id, data);
+        });
+      }
+
+      var history = [], matrix, data, options, container;
       switch (id) {
         case 'exp':
           history = User.user.history.exp;
@@ -202,23 +208,62 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
           $rootScope.charts[id] = (history.length == 0) ? false : !$rootScope.charts[id];
           if (task && task._editing) task._editing = false;
       }
+
       matrix = [[env.t('date'), env.t('score')]];
       _.each(history, function(obj) {
         matrix.push([moment(obj.date).format(User.user.preferences.dateFormat.toUpperCase().replace('YYYY','YY') ), obj.value]);
       });
       data = google.visualization.arrayToDataTable(matrix);
+
+      $rootScope.drawChart(id, data);
+    };
+
+    $rootScope.drawChart = function(id, data, options) {
+      var chart, width;
+
+      switch(id) {
+        case 'exp':
+          width = $(".row").width() - 5;
+          break;
+        case 'todos':
+          width = $(".task-column.todos").width() - 5;
+          break;
+        default:
+          width = $(".task-text").width() - 5;
+          break;
+      }
+
       options = {
         title: window.env.t('history'),
         backgroundColor: {
           fill: 'transparent'
         },
-        hAxis: {slantedText:true, slantedTextAngle: 90},
-        height:270,
-        width:300
+        hAxis: {
+          slantedText: true, 
+          slantedTextAngle: 90,
+          textStyle: {
+            fontSize: 12
+          }
+        },
+        vAxis: {
+          format: 'short',
+          textStyle: {
+            fontSize: 12
+          }
+        },
+        width: width,
+        height: 270,      
+        chartArea: {
+          left: 50,
+          top: 30,  
+          right: 65,
+          bottom: 65
+        }
       };
+
       chart = new google.visualization.LineChart($("." + id + "-chart")[0]);
       chart.draw(data, options);
-    };
+    }
 
     $rootScope.getGearArray = function(set){
       var flatGearArray = _.toArray(Content.gear.flat);
--- a/website/client-old/js/controllers/rootCtrl.js
+++ b/website/client-old/js/controllers/rootCtrl.js
@@ -225,16 +225,12 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
     function drawChart(id, data, options) {
       var chart, width;
 
-      switch(id) {
-        case 'exp':
-          width = $(".row").width() - 20;
-          break;
-        case 'todos':
-          width = $(".task-column.todos").width();
-          break;
-        default:
-          width = $(".task-text").width() - 15;
-          break;
+      if(id === "exp") {
+        width = $(".row").width() - 20;
+      } else if(id === "todos") {
+        width = $(".task-column.todos").width();
+      } else {
+        width = $(".task-text").width() - 15;
       }
 
       options = {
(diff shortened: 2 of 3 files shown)


ltorroba:
I picked up issue #6679, and noticed that alongside fixing the clipping problem this would be a good chance to fix a few other bugs related to charts. They were listed above. On the whole, I believe this PR would make the chart somewhat prettier, especially when it comes to dynamic/flexible container sizes.

## Charts currently at desktop resolution (Safari @ Mac):

### EXP



### Todos



### Task

Portion of the `+` sign becomes unclickable as the chart floats above it, See next section for the specific region,

_The same behaviour was observed on Chrome on the original issue, and confirmed by tests._

## Charts currently, at reduced resolution (Chrome @ Mac):

### EXP



### Todos



### Task

_Moreover, bounded region becomes unclickable_

_I also verified this on Firefox and Safari._

## Charts after these commits, at desktop resolution (Chrome @ Mac)

### EXP



### Todos



### Task


_Also tested on Firefox and Safari._

## Charts after these commits, at reduced resolution (Chrome @ Mac)

### EXP



### Todos



### Task



# The only problem

Only exception is at very specific screen sizes (e.g.: task column div width is between 271 and 273 pixels) on Safari and Firefox, where the following may occur:


I have tried different workarounds but it seems to be a library and browser issue, so we are not talking anything neat. Any ideas? Should we fix this, and/or try to report it as an issue with the Google Charts API itself?

keysmashes:
Is it even worth showing the legend? It seems like it should be self-evident, and 'score' isn't very informative anyway.

website/public/js/controllers/rootCtrl.js
```diff
@@ -169,7 +169,13 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
 
     $rootScope.charts = {};
     $rootScope.toggleChart = function(id, task) {
-      var history = [], matrix, data, chart, options;
+      if($rootScope.charts[id] === undefined) {
+        $(window).resize(function() {
```
crookedneighbor: Window resize functions should be debounced.

website/public/js/controllers/rootCtrl.js
```diff
@@ -169,7 +169,13 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
 
     $rootScope.charts = {};
     $rootScope.toggleChart = function(id, task) {
-      var history = [], matrix, data, chart, options;
+      if($rootScope.charts[id] === undefined) {
```
crookedneighbor: A simple `if (!$rootScope.charts[id])` should suffice here.

website/views/options/social/tavern.jade
```diff
@@ -83,7 +83,7 @@
                 a.label.label-contributor-1(ng-click='toggleUserTier($event)')=env.t('tier') + ' 1 (' + env.t('friend') + ')'
                 div
                   p
-                    span.achievement.achievement-firefox
+                    span.achievement.achievement-boot
```
crookedneighbor, line 86: This doesn't look related to this PR.

website/public/js/controllers/rootCtrl.js
```diff
@@ -184,23 +190,62 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
...
       matrix = [[env.t('date'), env.t('score')]];
       _.each(history, function(obj) {
         matrix.push([moment(obj.date).format(User.user.preferences.dateFormat.toUpperCase().replace('YYYY','YY') ), obj.value]);
       });
       data = google.visualization.arrayToDataTable(matrix);
+
+      $rootScope.drawChart(id, data);
+    };
+
+    $rootScope.drawChart = function(id, data, options) {
```
crookedneighbor: Does this need to be exposed onto `$rootScope`? Looks like it is only called internally.

ltorroba:
@anowlcalledjosh @crookedneighbor Sorry for the delay, but here are the commits associated with your suggestions. Let me know if there's anything else to be done. 😄

Alys:
@zesme Thanks for that! Just so you know, we won't be able to review this until late this month because we have a code freeze now while we finish the final stages of version 3 of Habitica's API. If you'd like more information, see the new "API Changes" section in Guidance for Blacksmiths at http://habitica.wikia.com/wiki/Guidance_for_Blacksmiths#API_Changes and you're welcome to ask questions in the Aspiring Coders guild.

ltorroba:
@Alys No worries, I had heard of the code freeze but, ironically, this To-Do had gotten quite red in my list, so I felt it was time to get it over with. As soon as the new API gets released I will get working on the merging.

website/views/shared/tasks/task.jade
```diff
@@ -14,4 +14,4 @@ li(id='task-{{::task.id}}',
 
     include ./edit/index
 
-  div(class='{{obj._id}}{{task.id}}-chart', ng-show='charts[obj._id+task.id]')
+
```
awitherow: is this necessary?

website/public/js/controllers/rootCtrl.js
```diff
@@ -184,23 +184,73 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
           $rootScope.charts[id] = (history.length == 0) ? false : !$rootScope.charts[id];
           if (task && task._editing) task._editing = false;
       }
+
+      if($rootScope.charts[id]) {
```
awitherow: I see a lot of cases with spaces between the if and the parenthesis, and also a lot of cases not with. Is there a code styleguide that is being followed?

crookedneighbor: We've instituted eslint for the server code. We're about to do a rewrite of the client code, so we haven't bothered including it in the linting, since it will be replaced soon.

website/views/shared/tasks/task.jade
ltorroba: Good point! Not quite sure what was going on in my head when I removed that. I just made a push to bring it back.

ltorroba:
Fixed conflicts with base branch

website/client/js/controllers/rootCtrl.js
```diff
@@ -185,23 +185,73 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
           $rootScope.charts[id] = (history.length == 0) ? false : !$rootScope.charts[id];
           if (task && task._editing) task._editing = false;
       }
+
+      if($rootScope.charts[id]) {
+        var handleResize = _.debounce(function() {
+          drawChart(id, data);
+        }, 300);
+
+        $(window).resize(handleResize);
```
crookedneighbor: This is adding a new event listener to the window resize every time the chart is toggled. Meaning if you toggle back and forth a few times, you add a new listener each time and the old listeners do not get cleaned up. This will be a performance issue.

website/client/js/controllers/rootCtrl.js
```diff
@@ -185,23 +185,73 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
...
       });
       data = google.visualization.arrayToDataTable(matrix);
+
+      drawChart(id, data);
+    };
+
+    function drawChart(id, data, options) {
+      var chart, width;
+
+      switch(id) {
```
crookedneighbor: Let's use `if/else if/else` here instead of a switch.

crookedneighbor:
Sorry for the super long delay on reviewing this. It looks 95% of the way there. The one big problem with it is that it's setting window resize listeners without cleaning up after itself.

ltorroba:
@crookedneighbor No worries! I am kinda busy this week, but as soon as I get a chance I will get everything sorted.

crookedneighbor:
@zesme Status report on this?

ltorroba:
@crookedneighbor I'll be taking a look at this today or within the next two days by the latest. Sorry about the delay!

ltorroba:
@crookedneighbor Made the changes you asked for. 66922a6 seems kinda hacky to me, but I couldn't figure out a better way to implement it. If you (or anyone else) has any ideas I'll happily improve it!

crookedneighbor:
Did a tiny adjustment: e3b9636c42d2641fe038df6226ad548e88d4f1b3

Basically, just set one listener on the window and looped through the resize functions for each active chart.

PR #6837, opened 2016-03-08
Changed tavern tier 1 text icon

ltorroba:
The icon for the text/description of the 1st tier contributor was
incorrect. It should be the boot icon, but the firefox icon was being
displayed.

This issue was reported through the wiki here: http://habitica.wikia.com/wiki/Thread:41876


ltorroba:
Here's the fix for the problem you posted in the guild chat @Alys

Alys:
Thank you! I've noted that as credit towards your second tier.

PR #6842, opened 2016-03-08
Added initial get challenge task by id

TheHollidayInn:

### DIFF
--- a/website/src/controllers/api-v3/tasks.js
+++ b/website/src/controllers/api-v3/tasks.js
@@ -257,7 +257,7 @@ api.getTask = {
     if (!task) {
       throw new NotFound(res.t('taskNotFound'));
     } else if (!task.userId) { // If the task belongs to a challenge make sure the user has rights
-      let challenge = await Challenge.find().selec({_id: task.challenge.id}).select('leader').exec();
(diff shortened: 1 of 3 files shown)


paglias:
@TheHollidayInn I noticed the get tasks for user tests are under at /tasks/GET-tasks_user.test.js, let's move these ones to /tasks/GET-tasks_challenge_challengeId.test?

paglias:
I'd add a test to check that a wrong id is handled correctly then I think we can merge!

website/src/controllers/api-v3/tasks.js
```diff
@@ -257,7 +257,7 @@ api.getTask = {
     if (!task) {
       throw new NotFound(res.t('taskNotFound'));
     } else if (!task.userId) { // If the task belongs to a challenge make sure the user has rights
-      let challenge = await Challenge.find().selec({_id: task.challenge.id}).select('leader').exec();
+      let challenge = await Challenge.find({_id: task.challenge.id}).select('leader').exec();
```
paglias, line 260: ops...! thanks for catching this

TheHollidayInn:
Are you saying to move the tests out of the challenges directory and rename the file to `/tasks/GET-tasks_challenge_challengeId.test` ?

And, I'll get that test added!

PR #6843, opened 2016-03-09
Fix #6766 - New party badges

### LINKED ISSUE #6766: New Party Badges!
We've got two new achievement badges to implement! Hooray :)

The first is called "Party Up" and is awarded to anyone who has formed a party with at least two total people.

partybadge1b

Description: Joined a Party with another person! Have fun battling monsters and supporting each other.

The second is called "Party On" and is awarded to anyone who has formed a party with at least four total people.

partybadge2

Description: Was in a Party with at least four people. Enjoy your increased accountability as you unite with your friends to vanquish your foes!

A note that's open to debate -- we should decide whether or not leaving the party causes you to lose the badge. My instinct is that no, once you've earned it you should get to keep it, but I'd like to hear discussion from others.

khipkin:
This PR addresses issue #6766.

(Edit: the comment that was here is no longer relevant.)

Here are @Alys's acceptance criteria for this issue:
- [x] You gain the achievement as soon as the party you are in gains a new member to bring it up to two/four people.
- [x] All players in the party gain the achievement when the party reaches the correct size, including the newest player.
- [x] Once you have the achievement, you cannot lose it.
- [x] There's no special treatment for WikiFAQ_bot - inviting it does the same thing as inviting a normal player.

I followed the same pattern as the pet/mount achievements.

---

In addition to those acceptance criteria, I added the following of my own:
- [x] The achievements are listed in the achievement section of the user profile page.

On another user's profile, they are visible if earned, otherwise not visible.  On your user profile, they are displayed normally if earned, otherwise they are grayed-out.
They precede the pet and mount achievements in the listing because, as @lemoness pointed out, pets don't unlock until level 3 but you can form a party as soon as you join the site.

---

There aren't any tests for this yet, but I'm going to try to write some.


lemoness:
The images were attached to the first GH ticket -- are you unable to access them? Here they are again:

partybadge2
partybadge1b

khipkin:
@lemoness Oh I didn't realize I was supposed to add them to the repo.  How do I do that?

Alys:
You can put them into the common/img/sprites/spritesmith/achievements directory, and then use `git add` to allow them to be committed.

khipkin:
Okay, I figured out how to add those two images and recompile the spritesheets! Turns out it was in the wiki after all. :smiling_imp: 

For the 2x images, can I just scale the provided images up and save new copies?  Do I have to worry about distortion?

SabreCat:
Scaling up the images works fine. If you tell your image editor not to use any interpolation when upsizing, it should preserve the pixelated look we want!

khipkin:
PartyUp modal
image

Earned vs Unearned Achievs on Your Profile
image

khipkin:
@SabreCat I have a question! I'm going to speak in terms of only the PartyUp achievement, for simplicity's sake.

When a user goes to their party page (for example, after the auto refresh after joining a party), if the party has 2+ players, the PartyUp achievement modal pops up correctly.  Then, if you go to the user's profile (the current user's profile), the achievement shows up there, too.

However, the achievement doesn't persist.  If I log out and back in, the achievement goes away again.  I think my code [here](https://github.com/khipkin/habitrpg/commit/d2d3b9deab4c65ecafe8c83a402a740b43fff4a4#diff-d264a0a600ac16c2ba1bf7bb86dd1dd3R536) isn't working correctly.  Or maybe it's because I changed the user schema?  I'm lost.

SabreCat:
@khipkin: You may need a `user.save()` call in the party controller. I'll take a look at this from my side today!

website/public/js/controllers/partyCtrl.js
```diff
@@ -11,6 +14,20 @@ habitrpg.controller("PartyCtrl", ['$rootScope','$scope','Groups','Chat','User','
 
       if ($state.is('options.social.party')) {
         $scope.group.$syncParty(); // Sync party automatically when navigating to party page
+
+        // Checks if user's party has reached 2 players for the first time.
+        if(!user.achievements.partyUp
+            && $scope.group.memberCount >= 2) {
+          user.achievements.partyUp = true;
```
SabreCat: OK, here's the trouble. Simply changing the property in the `user` object here only updates it on the client. You need a `User.set()` call to get it to the server--see, for instance, https://github.com/khipkin/habitrpg/blob/new_party_badges/website/public/js/controllers/partyCtrl.js#L122.

SabreCat:
The 2x badge for "Party Up" appears to have gotten blurry, while the "Party On" looks OK. Used a different upscale method, maybe?

khipkin:
@SabreCat Nice! Thanks!

That seems to have done it.  Here's what another user's profile looks like once they've earned the "Party Up" badge:
image
That didn't show up before!

Also, I'll look at the images again and see if I can get the second one to stop being blurry...

SabreCat:
@khipkin: Thank you so much for this! We will move the logic into the API later so that the achievements can be awarded even when a user doesn't visit the web party page, but at present API changes are frozen while we work on version 3. So this will go live soon!

This will count toward a contributor tier! What's your Habitica User ID (from https://habitica.com/#/options/settings/api )?

khipkin:
@SabreCat ahh yayy! :smile_cat: f9c89bba-abf8-4f20-a974-cf1550ff603f

Also, I've written a handful of tests for these changes, but I haven't been able to run them because the test command errs out. :cry: I'll create a new pull request for them and see if they work.

SabreCat:
@khipkin: Tier awarded!

Note that while your first tier is relatively easy to get, each subsequent tier requires greater effort from one to the next. So your next contribution may or may not take you to 2, depending on scope. But keep helping out and we'll express our thanks accordingly!

PR #6853, opened 2016-03-10
Added initial challenge task update tests

TheHollidayInn:

### DIFF
--- a/website/src/controllers/api-v3/tasks.js
+++ b/website/src/controllers/api-v3/tasks.js
@@ -301,7 +301,7 @@ api.updateTask = {
     if (!task) {
       throw new NotFound(res.t('taskNotFound'));
     } else if (!task.userId) { // If the task belongs to a challenge make sure the user has rights
-      challenge = await Challenge.find().selec({_id: task.challenge.id}).select('leader').exec();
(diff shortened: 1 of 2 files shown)


website/src/controllers/api-v3/tasks.js
```diff
@@ -301,9 +301,9 @@ api.updateTask = {
     if (!task) {
       throw new NotFound(res.t('taskNotFound'));
     } else if (!task.userId) { // If the task belongs to a challenge make sure the user has rights
-      challenge = await Challenge.find().selec({_id: task.challenge.id}).select('leader').exec();
+      challenge = await Challenge.findOne({_id: task.challenge.id}).populate('leader').exec();
       if (!challenge) throw new NotFound(res.t('challengeNotFound'));
-      if (challenge.leader !== user._id) throw new NotAuthorized(res.t('onlyChalLeaderEditTasks'));
```
paglias, line 306: is it necessary to populate the leader here? We should just need the simple `leader` property to check the id

TheHollidayInn, line 306: You right. I removed the populate.

PR #6858, opened 2016-03-10
Add fix for running in dev and test mode

crookedneighbor:


crookedneighbor:
babel-polyfill was needed when starting up the server in dev or test mode.

Issue #6859, opened 2016-03-10
Inconsistent html in locale source strings


ghost:
Bringing up the conversation from the [i18n guild](https://habitica.com/#/options/groups/guilds/7732f64c-33ee-4cce-873c-fc28f147a6f7):

> **_Blablux**_ [...] The string is in `groups.json` line 103 [on Transifex], and it reads (excerpt): [...]`but are not limited to: <p></p></br><ul style='margin-left: 10px;'><li>swearing`[...]. What's bothering me here is the `<p></p>`. While fundamentaly valid, this html paragraph doesn't seem to be of any purpose. You can get this text in the site by trying to report any user message, and while it adds some space in that message, I'm not sure that it's either intended, neither the way to do it.
> 
> **_Blablux**_ And now that I wrote about it, I'm just seeing the `</br>` tag which, as far as I know, is invalid. It should be either `<br>` or much better `<br />`...

For reference, in the locale file, this goes to line 104 `"abuseFlagModalBody"`.

> **_Sphinx**_ I don't know the reason, but the pages of Habitica will have empty paragraphs from time to time. In this case, my guess would be that this paragraph was introduced by something in the [jade source ](https://github.com/HabitRPG/habitrpg/blob/develop/website/views/shared/modals/members.jade#L103)and was simply transferred to plain html when this part was made translatable. Deleting this empty paragraph doesn't seem to influence the way the modal looks.
> Line break does work, but I agree with you that it may be not the best practice to have it in such a way
> 
> **_Blade**_ Hmm.... that looks really wrong to me. I'll investigate it later today.
> 
> **_Blablux**_ While we're at it, I checked if there was some other parts like this in Transifex. I didn't found any like `<p></p>`, but I noticed that the text in `contentQuest.json` is usually just plain text, but from time to time, it has html formating. Here also, this may be the intended behaviour (though I don't always see why), but there are times where you have some text like `<p>[...]</p><br/><p>[...]</p>`. And while it is valid html, the line break between the two paragraphs seems redundant and not that useful.
> 
> **_Blade**_ In my opinion, all the quest content should be in markdown. Not html. Obviously, that's a change that needs to happen in the source strings.

For example, see `"questMoonstone1Notes"` in `questcontent.json` locale file.
### Long story short:

There are some html inconsistencies in the source strings.  
As far as I understand it, all the html should be written in markdown instead in the source. If that's indeed the case, I may be able to help tracking them and correcting them.  
[...]

PR #6864, opened 2016-03-11
Fix html code in locale files (see #6859)

ghost:
Fixed some html code discussed in issue #6859 and some other x/html inconsistencies.


keysmashes:
I would suggest that we continue to use `<br>` instead of `<br/>` or `<br />`, because that is the style used throughout the codebase for the most part*, and it makes no difference to browsers.

https://stackoverflow.com/questions/1946426/html-5-is-it-br-br-or-br

*searching for `<br>` in the Tasks page gave 220 results, `<br/>` gave 11 and `<br />` gave 3

Alys:
The tasks page isn't a good indicator of the codebase because most of those br tags will be automatically generated by the markdown parser or templating engine or other things.

If you do a git grep for br ignoring the locales files, you find:

```
<br/> 8 lines
<br /> 4 lines
<br> 2 lines
```

Those numbers are too small to indicate a convention for any of them.

In the locales/en directory (not including this PR's changes) you find:

```
<br/> 6 lines
<br /> 0 lines
<br> 34 lines
```

This PR probably changes all of those to `<br />`; it looks thorough at a glance.

Normally, my personal preference is for `<br />` because having valid XHTML doesn't hurt and can be useful. You could make a case for leaving them as `<br>` for the sake of not making extra work now for the translators. However if we find that later we'd benefit from `<br />` being consistently used, then changing later would be more work than changing now. It's not inconceivable that we'd want XHTML later so that we could do automatic processing/conversion of the text. I think I'm in favour of moving to `<br />` now.

Other opinions?

crookedneighbor:
I'd prefer `<br>` as it is what the jade templating language renders `br` as and is valid HTML5 (current standard over XHTML).

keysmashes:
OTOH, all three are valid HTML5.

> Start tags consist of the following parts, in exactly the following order:
> 1. A "<" character.
> 2. The element's tag name.
> 3. Optionally, one or more attributes, each of which must be preceded by one or more space characters.
> 4. **Optionally, one or more space characters.**
> 5. **Optionally, a "/" character, which may be present only if the element is a void element.**
> 6. A ">" character.

https://www.w3.org/TR/html-markup/syntax.html#syntax-start-tags

ghost:
My tendency was obviously to go toward valid xhtml, but I can change back to `<br>` if we can agree on any convention. As far as I'm concerned, @crookedneighbor made a strong point with jade. You just tell me.

ghost:
I've reverted to `<br>` as it seems to be the best idea. All that's left is the html corrections on the few files that were explicitly mentioned on the original issue.

common/locales/en/questsContent.json
```diff
@@ -69,7 +69,7 @@
...
+  "questVice1Notes": "They say there lies a terrible evil in the caverns of Mt. Habitica. A monster whose presence twists the wills of the strong heroes of the land, turning them towards bad habits and laziness! The beast is a grand dragon of immense power and comprised of the shadows themselves: Vice, the treacherous Shadow Wyrm. Brave Habiteers, stand up and defeat this foul beast once and for all, but only if you believe you can stand against its immense power. <br><h3>Vice Part 1: </h3><br>How can you expect to fight the beast if it already has control over you? Don't fall victim to laziness and vice! Work hard to fight against the dragon's dark influence and dispel his hold on you!",
```
ghost: Have you checked that there would be no visual difference between
`<p>text</p>` and just `text`
for the strings in this file?

Alys: @Blablux When paragraphs and headings are mixed, `<p>` tags should be used. 

This is better HTML:
`<p>... power. </p><h3>Vice Part 1: </h3><p>How ... </p>`

than this:
`... <br><h3>Vice Part 1: </h3><br>How ...`

ghost: I reverted to `<p>` for this part. But this leads me to another question. _Vice Part 1_ is the only quest using a heading tag. I did not find another one on any part of the _Vice_ quests or any other quest. While it is certainly not a problem, this seems odd to me. Should I open an issue about it to discuss what we can do ? Or should I just leave it like that as it is not a real problem ?

crookedneighbor:
Squashed your commits and merged them in. 

ghost:
Wow, I did not thought it was worth a contributor tier ! So, well, thank you !
My User ID is a75eb6f8-5a29-482f-a971-7aff7438da86

crookedneighbor:
I've notated this toward your next tier and added Blacksmith to your title.

PR #6868, opened 2016-03-12
E2E test suit for protractor

kogelnikp:
I've seen that there are some experiments with protractor in the repository. I think protractor is a great tool for end to end testing of AngularJS sites. Therefore, I started to create a test suit for habitica which follows the [page object pattern](http://martinfowler.com/bliki/PageObject.html). The tests are based on the approach of [this blog post](http://avandeursen.com/2015/06/25/beyond-page-objects/) and the github project which is referenced in the post. The suit provides objects for the different states of the system. These objects map the action which can be performed if the application is in this state.

I implemented three states for the system, one for the public habitica page, one for the welcome tour and one for the tasks page. In the test suit itself I just perform the actions of the state and after the actions I check if everything worked correctly.

If there are some questions, feel free to ask.

### DIFF
--- a/test/e2e/tasks.test.js
+++ b/test/e2e/tasks.test.js
@@ -1,0 +1,161 @@
+'use strict';
+
+var fs = require('fs');
+
+
+
+
+
+//State: public welcome page
+var HabiticaPublic = (function () {
+    
+    function generateUserid(length) {
+        var text = "";
+        var possible = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+        for( var i=0; i < length; i++ ) {
+            text += possible.charAt(Math.floor(Math.random() * possible.length));
+        }
+        return text;
+    }
+    
+    return {
+        
+        get: function () {
+            return browser.get('/');
+        },
+        selfCheck: function () {
+            expect(browser.getLocationAbsUrl()).toBe('');
+        },
+        register: function(done) {
+            var userId = generateUserid(10);    
+            element(by.id('play-btn')).click();
+            element(by.linkText('Register')).click();
+            element(by.model('registerVals.username')).sendKeys(userId);
+            element(by.model('registerVals.email')).sendKeys(userId + '@example.com');
+            element(by.model('registerVals.password')).sendKeys('pass');
+            element(by.model('registerVals.confirmPassword')).sendKeys('pass');
+            element(by.css('#registrationForm input[type="submit"]')).click();     
+            browser.sleep(1000);
+        }
+    };
+}());
+
+
+//State: welcome tour after registering
+var WelcomeTour = (function () {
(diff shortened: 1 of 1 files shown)


crookedneighbor:
There's a bunch of linting errors in your test file. Could you clean those up first?

protractor.conf.js
```diff
@@ -10,7 +10,7 @@ exports.config = {
 
   // Spec patterns are relative to the current working directly when
   // protractor is called.
-  specs: ['test/e2e/e2e.js'],
+  specs: ['test/e2e/tasks.test.js', 'test/e2e/e2e.js'],
```
crookedneighbor: This should be a glob pattern, I believe:

``` js
specs: ['test/e2e/**/*.js'],
```

crookedneighbor:
Thanks for your interest in improving the end to end tests with protractor. I'm currently updating our protractor config to use protractor v3 and to use the mocha test framework. You can see the progress here: https://github.com/HabitRPG/habitrpg/compare/develop...crookedneighbor:update_protractor?expand=1

I think the only thing left to do is get it to run in headless mode on CI. I'll ping you when it's done so you can adjust your PR.

As for the style of the tests, I'm not convinced that it's the best way to go for these. I'll read the post and think about it though and get back to you.

crookedneighbor:
Protractor update is done.

kogelnikp:
Ok, I'll have a look on it today. Let me know if you have some suggestions for the test suite.

kogelnikp:
Should I keep the registration test in my test file? At the moment the user is already logged in when the test arrives at my file (because of the front page tests). I would suggest I remove the registration and HabiticaPublic page tests from my file and the front-page tests are only done in front-page.test.js.

kogelnikp:
@crookedneighbor Now the tests are working. I had a few problems with the notifications, because they hide some buttons in the tests. For now I just wait 8 seconds until the notifications are invisible and then I perform the clicks. Would be great if you know a better solution.

crookedneighbor:
One problem with these tests is that they are not autonomous. They all rely on the welcome test to go through first to ensure that the user has already gone through the tutorial. This means you can't run any of the other tests in isolation.

Before this can be merged in, we need to create a helper function that creates a user and marks all tutorial and welcome message flags so they don't appear for the user when loading the site. I may not get to this for a little while, so this PR may be on hold till then.

crookedneighbor:
Okay. So, here's the thing. We're going to do a pretty drastic redesign of the site soon, meaning most of these tests will need to be rewritten anyway. 

When we do the redesign, we'll make sure to include protractor tests from the beginning for the important functionality. If you'd like to help with that when that starts to happen, you are very welcome to. I will help to write up some infrastructure then to make the tests autonomous. 

But for now, I'm going to close this.

Jarda8:
OK, we will wait for that. :wink:

PR #6869, opened 2016-03-12
Added initial test for removing challenge checklist item

TheHollidayInn:


website/src/controllers/api-v3/tasks.js
```diff
@@ -703,7 +703,7 @@ api.removeChecklistItem = {
     if (!task) {
       throw new NotFound(res.t('taskNotFound'));
     } else if (!task.userId) { // If the task belongs to a challenge make sure the user has rights
-      challenge = await Challenge.find().selec({_id: task.challenge.id}).select('leader').exec();
+      challenge = await Challenge.findOne({_id: task.challenge.id}).exec();
```
paglias, line 706: wow... I think I copy-pasted this error everywhere :)

PR #6872, opened 2016-03-13
Added initial delete challenge tasks tests

TheHollidayInn:

@paglias I saw this in a TODO: "cannot delete active challenge tasks". Is this still a functionality we want? If so, I think this PR is a good place to add it. Just let me know, and I'll add the code.

### DIFF
--- a/test/api/v3/integration/tasks/challenges/DELETE-tasks_id_challenge_challengeId.test.js
+++ b/test/api/v3/integration/tasks/challenges/DELETE-tasks_id_challenge_challengeId.test.js
@@ -2,6 +2,7 @@ import {
   generateUser,
   generateGroup,
   generateChallenge,
+  sleep,
   translate as t,
 } from '../../../../../helpers/api-integration/v3';
 import { v4 as generateUUID } from 'uuid';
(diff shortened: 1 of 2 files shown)


paglias:
@TheHollidayInn you mean that it shouldn't be possible for a user to delete a challenge task (not the original challenge task but a synced one)? (I can't find the TODO comment),

If that's what you meant I'm not too sure, it's a good question for @Alys

Alys:
When a user has joined a challenge, they must not be able to delete any tasks from the challenge.

If they leave the challenge and choose to keep the tasks, then they can be allowed to delete any of them.

TheHollidayInn:
Here is the line: https://github.com/HabitRPG/habitrpg/blob/api-v3/test/api/v3/integration/tasks/DELETE-tasks_id.test.js#L57 .

From reading the todo, it made me think there is a time when people can join and the challenge is inactive. But, I may just be thinking about it incorrectly.

Alys:
I think that just means "TODO after some piece of code for challenges is implemented". `shrug`

Regardless, when a player is still in a challenge (i.e., after they have joined and before they have left) they must not be able to delete any of the challenge tasks in any circumstances.

Challenges are active as soon as they are created. A player is an active participant as soon as they have joined.

(You might sometimes see a challenge owner say that a challenge starts on a particular date even though the challenge already exists on an earlier date, but that "start date" is not something that is ever coded into the challenge itself. It's purely something that the owner has to keep track of themself in some way.)

TheHollidayInn:
Okay. That makes sense. Well, the current functionality in this PR prevents anyone who is not the leader from deleting. So, I think we should be good to go if you agree.

paglias:
I think what Alys is talking about (and what I meant with that pending test) is that every user that has joined a challenge should not be able to delete the tasks associated with that challenge until they leave it. I think you're confusing that with the ability of challenge leaders to delete directly challenge tasks (and so have them deleted from all the users that joined that challenge). Right now tasks with challengeId but no userId can be deleted by the leader and that's ok. We also want tasks with userId and challengeId not being deletable unless the task is broken (user left, challenge closed,...

TheHollidayInn:
Ohh. Yea, I understand now. Okay, I'll get that updated tonight.

TheHollidayInn:
Alright, updated! I was forced to use the sleep function because there are tasks after the response on this route.

TheHollidayInn:
Ah, okay, I guess I was just over thinking it. Updating now.

PR #6875, opened 2016-03-13
update username, password

piousbox:


website/src/controllers/api-v3/user.js
```diff
@@ -43,6 +43,95 @@ api.getUser = {
...
+    // check username already exists
+    let candidateUser = await User.findOne({
+      'auth.local.username': req.body.username,
+    }, {'auth.local': 1}).exec();
+    if (candidateUser) throw new BadRequest(res.t('usernameTaken'));
+
+    // save username
+    user.auth.local.lowerCaseUsername = req.body.username.toLowerCase();
+    user.auth.local.username = req.body.username;
+    user.save();
```
paglias: await

website/src/controllers/api-v3/user.js
```diff
@@ -43,6 +43,95 @@ api.getUser = {
...
+    if (validationErrors) throw validationErrors;
+
+    if (!user.auth.local.username) throw new BadRequest(res.t('userHasNoLocalRegistration'));
+
+    let oldPassword = passwordUtils.encrypt(req.body.password, user.auth.local.salt);
+    if (oldPassword !== user.auth.local.hashed_password) throw new NotAuthorized(res.t('wrongPassword'));
+
+    // check username already exists
+    let candidateUser = await User.findOne({
+      'auth.local.username': req.body.username,
```
paglias: you want to search for `lowercaseUsername` here not for `username`

website/src/controllers/api-v3/user.js
```diff
@@ -43,6 +43,95 @@ api.getUser = {
...
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    if (!user.auth.local.username) throw new BadRequest(res.t('userHasNoLocalRegistration'));
+
+    let oldPassword = passwordUtils.encrypt(req.body.password, user.auth.local.salt);
+    if (oldPassword !== user.auth.local.hashed_password) throw new NotAuthorized(res.t('wrongPassword'));
+
+    // check username already exists
+    let candidateUser = await User.findOne({
```
paglias: I'd use a count query instead of a findOne here

website/src/controllers/api-v3/user.js
```diff
@@ -43,6 +43,95 @@ api.getUser = {
...
+        notEmpty: {errorMessage: res.t('missingPassword')},
+      },
+    });
+
+    if (req.body.newPassword !== req.body.confirmPassword) throw new NotAuthorized(res.t('passwordConfirmationMatch'));
+
+    user.auth.local.hashed_password = passwordUtils.encrypt(req.body.newPassword, user.auth.local.salt); // eslint-disable-line camelcase
+    user.save();
+
+    res.send(200, { message: res.t('passwordSaved') });
```
paglias: @piousbox usually we don't return specific success messages... probably it's better to just do `res.respond(200, {})` for now

website/src/controllers/api-v3/user.js
```diff
@@ -43,6 +43,95 @@ api.getUser = {
...
+      },
+      newPassword: {
+        notEmpty: {errorMessage: res.t('missingPassword')},
+      },
+    });
+
+    if (req.body.newPassword !== req.body.confirmPassword) throw new NotAuthorized(res.t('passwordConfirmationMatch'));
+
+    user.auth.local.hashed_password = passwordUtils.encrypt(req.body.newPassword, user.auth.local.salt); // eslint-disable-line camelcase
+    user.save();
```
paglias: await

website/src/controllers/api-v3/user.js
```diff
@@ -43,6 +43,91 @@ api.getUser = {
...
+
+    let count = await User.count({ 'auth.local.lowerCaseUsername': req.body.username.toLowerCase() });
+    if (count > 0) throw new BadRequest(res.t('usernameTaken'));
+
+    // save username
+    user.auth.local.lowerCaseUsername = req.body.username.toLowerCase();
+    user.auth.local.username = req.body.username;
+    await user.save();
+
+    res.send(200, { username: req.body.username });
```
paglias: you should use `res.respond` here, res.send is different and also passing the status code to it is not supported anymore http://expressjs.com/en/4x/api.html#res.send

website/src/controllers/api-v3/user.js
```diff
@@ -43,6 +43,91 @@ api.getUser = {
...
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    if (!user.auth.local.username) throw new BadRequest(res.t('userHasNoLocalRegistration'));
+
+    let oldPassword = passwordUtils.encrypt(req.body.password, user.auth.local.salt);
+    if (oldPassword !== user.auth.local.hashed_password) throw new NotAuthorized(res.t('wrongPassword'));
+
+    let count = await User.count({ 'auth.local.lowerCaseUsername': req.body.username.toLowerCase() });
```
paglias, line 117: This isn't something very important but there's no reason to call toLowerCase twice, just to `let lowercaseUsername = req.body.username.toLowerCase()` and use that

website/src/controllers/api-v3/user.js
```diff
@@ -43,6 +43,91 @@ api.getUser = {
...
+      newPassword: {
+        notEmpty: {errorMessage: res.t('missingPassword')},
+      },
+    });
+
+    if (req.body.newPassword !== req.body.confirmPassword) throw new NotAuthorized(res.t('passwordConfirmationMatch'));
+
+    user.auth.local.hashed_password = passwordUtils.encrypt(req.body.newPassword, user.auth.local.salt); // eslint-disable-line camelcase
+    await user.save();
+    res.send(200, {});
```
paglias: res.respond

website/src/controllers/api-v3/user.js
```diff
@@ -43,6 +43,91 @@ api.getUser = {
...
+    req.checkBody({
+      password: {
+        notEmpty: {errorMessage: res.t('missingNewPassword')},
+      },
+      newPassword: {
+        notEmpty: {errorMessage: res.t('missingPassword')},
+      },
+    });
+
+    if (req.body.newPassword !== req.body.confirmPassword) throw new NotAuthorized(res.t('passwordConfirmationMatch'));
```
paglias, line 76: you can use something like this inside checkBody:

```
      password: {
        notEmpty: {errorMessage: res.t('missingPassword')},
        equals: {options: [req.body.confirmPassword], errorMessage: res.t('passwordConfirmationMatch')},
      },
```

it'll also check that confirmPassword matches

common/locales/en/api-v3.json
```diff
@@ -4,7 +4,9 @@
   "missingEmail": "Missing email.",
   "missingUsername": "Missing username.",
   "missingPassword": "Missing password.",
+  "missingNewPassword": "Missing newPassword.",
   "wrongPassword": "Wrong password.",
+  "passwordSaved": "New password has been saved.",
```
paglias: this is not used anymore and should be removed

test/api/v3/integration/user/POST-user_update_password.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+  it('successfully changes the password', async () => {
+    let previousHashedPassword = user.auth.local.hashed_password;
+    let response = await user.post(endpoint, {
+      password,
+      newPassword,
+      confirmPassword: newPassword,
+    });
+    expect(response).to.eql({});
+    user = await User.findOne({ _id: user._id });
```
paglias: to update the user object you can use `await user.sync()`

website/src/controllers/api-v3/user.js
piousbox, line 76: I tried that first, but it didn't work for some reason.

website/src/controllers/api-v3/user.js
piousbox, line 117: I thought of that, but didn't think it a good idea (it introduces another variable and another line of code).

paglias, line 117: @piousbox I think a variable is less expensive than a function call, also all nested property accesses are slower (`obj.property`) since the property must be looked up. But it's not a big deal

PR #6876, opened 2016-03-13
E2E test suit for protractor - User pages

Jarda8:
Bunch of e2e tests for pages under the "User" menu item. The tests don't cover that much now. I want to see your opinions on this before I add more of them. The tests are working correctly on Chrome, not so much on Firefox (getting some weird errors there). The style of tests is described in PR #6868 , we work together on this.


Jarda8:
There are some selections based on a visible element text (linkText). Because of this the tests work only for the English version of Habitica.

crookedneighbor:
See this comment: https://github.com/HabitRPG/habitrpg/pull/6868#issuecomment-196058272

Jarda8:
OK. I hope it will be soon, so I can continue with the testing. :)

crookedneighbor:
Update is done.

Jarda8:
These tests need another test before them to log in and go through the welcome tour. (e.g. the static-pages/front-page.test and tasks/test, without the log off at the end).

Jarda8:
I am not sure, why "protractor" is not defined. Works fine on my computer.

khipkin:
The error is from linting.  I think `protractor` is in the global namespace, which is why eslint is saying it's undefined.

From [the eslint documentation](http://eslint.org/docs/rules/no-undef), you need to put the following before the "undefined" function:

```
/*global someFunction*/
/*eslint no-undef: 1*/
```

Jarda8:
You are right, it is ESLint. But I think that better solution will be to add the whole protractor environment to the ESLint config file. Or at least "protractor" to the globals.

khipkin:
Oh yeah, that's much better.  I didn't know about that!

crookedneighbor:
I would add it to the eslintrc in test/e2e since its specific to those tests.

crookedneighbor:
See this comment. It applies to this PR as well: https://github.com/HabitRPG/habitrpg/pull/6868#issuecomment-199828156

crookedneighbor:
See https://github.com/HabitRPG/habitrpg/pull/6868#issuecomment-207889557

PR #6878, opened 2016-03-14
Update protractor

crookedneighbor:
- Updates protractor dependency
- Converts protractor conf to use mocha framework instead of default jasmine
- Restructures e2e folder to be more organized


crookedneighbor:
Also fixes registration test that has been pended for-e-ver

pj7naal

PR #6888, opened 2016-03-15
[Api v3] Port shared.ops.allocate

paglias:


test/common/constants.js
```diff
@@ -0,0 +1,11 @@
+import {
```
crookedneighbor: I'm not sure a test against the values of constants is particularly useful. Just means you have to update this information in two places without any tangible benefit. If this test breaks, it is literally because you changed the constants file.

paglias: :+1:

PR #6889, opened 2016-03-15
feat: Add debug scripts to easilly get all pets/mounts/gear

crookedneighbor:
Is this useful? Makes it easier to test out gear, pets, and mount related things in development. 

@Alys @SabreCat @paglias


crookedneighbor:
Alternatively, these could be gulp tasks.

Alys:
I think that's good!

I like that it's run as a script with node because that will help devs work out how to write and execute their own scripts for other modifications they might want. Using node is simple.

SabreCat:
Coool. That's a great point, I've been wanting some models from which to build that sort of thing.

PR #6890, opened 2016-03-15
add tests for awarding party {up,on} achievs

khipkin:
~~For the life of me, I couldn't get the test suite to run locally.  So this PR is to check if the tests actually work...~~

Edit: got the test to run, just can't get them to pass.


khipkin:
@SabreCat I think the tests are failing because [`$state.is('options.social.party')`](https://github.com/HabitRPG/habitrpg/pull/6843/files#diff-2890f7e889237421a66b710ae1213522L12) is falsey when the tests are run, meaning none of my new party achievement code is actually executed, but I can't confirm and can't figure out how to force it to be truthy.

I've tried adding `_$state_` to the [`inject` function in the first `beforeEach`](https://github.com/HabitRPG/habitrpg/pull/6890/files#diff-c4bf8441da2dba76a23f7694922b0cc7L18) and `sandbox.stub`ing that, but that doesn't work.  I've tried passing in my own nonsense as `state` in [`ctrl = $controller('PartyCtrl', {$scope: scope, User: User, $state: state});`](https://github.com/HabitRPG/habitrpg/pull/6890/files#diff-c4bf8441da2dba76a23f7694922b0cc7L32) and that didn't work either...

I keep getting assertion errors in all of my tests that `User.set` was expected to be called exactly once/twice but that it was called 0 times.

Can you shed some light on this issue?

khipkin:
The karma tests for [challengesCtr](https://github.com/HabitRPG/habitrpg/blob/ad347090c30c45537a64b299458b7241ebb97435/test/spec/controllers/challengesCtrlSpec.js#L17) seem to successfully utilize the `inject(function(..., _$state_, ...)` method.  I think this may be because the only calls to `$state` in the [actual controller](https://github.com/HabitRPG/habitrpg/blob/ad347090c30c45537a64b299458b7241ebb97435/website/public/js/controllers/challengesCtrl.js) are when functions in its scope are called, whereas the party controller accesses `$state` during initialization.  That's the only significant difference as far as I can tell.

SabreCat:
I had some of the same difficulties when trying to add tests myself. @crookedneighbor @paglias, any ideas?

crookedneighbor:
Okay, so here's the issue. The [code you wrote](https://github.com/HabitRPG/habitrpg/blob/develop/website/public/js/controllers/partyCtrl.js#L15-L31) executes when the controller is instantiated, but in your tests, you're setting the group data _after_ the controller has been fully instantiated.

khipkin:
@crookedneighbor The `beforeEach` doesn't happen before instantiation?  How can I set things prior to the controller being instantiated?

crookedneighbor:
As you can see, the controller is instabilities in another beforeEach that happens _before_ your change to scope.

I'm not sure what the best course of action is going to be. We might have to break with the pattern for how we're testing controllers to catch this behavior. Let me research a little and get back to you.

test/spec/controllers/partyCtrlSpec.js
```diff
@@ -8,9 +8,12 @@ describe("Party Controller", function() {
     user._id = "unique-user-id";
     User = {
       user: user,
-      sync: sandbox.spy
+      sync: sandbox.spy,
```
crookedneighbor: These both should just be `sandbox.spy()`. Not sure why sync was set up wrong.

crookedneighbor:
Woah, clearly wrote that last comment on my phone. Tons of spelling mistakes. 

This is kinda gross, but you could just call `$controller` again after you've set up your scenario:

``` javascript
it('awards "Party On" achievement', function() {
  sandbox.stub(groups, 'party').returns({
    $syncParty: function () {},
    memberCount: 4
  });
  sandbox.stub(state, 'is').returns(true);
  user.achievements.partyUp = true;
  $controller('PartyCtrl', { $scope: scope });

  expect(User.set).to.be.calledOnce;
  expect(User.set).to.be.calledWith(
    { 'achievements.partyOn': true}
  );
  expect(rootScope.openModal).to.be.calledOnce;
  expect(rootScope.openModal).to.be.calledWith(
    'achievements/partyOn'
  );
});
```

test/spec/controllers/partyCtrlSpec.js
```diff
@@ -33,6 +36,101 @@ describe("Party Controller", function() {
...
+      context('user does not have "Party Up" achievement', function() {
+        it('awards "Party Up" achievement', function() {
+          expect(User.set).to.be.calledOnce;
+          expect(User.set).to.be.calledWith(
+            { 'achievements.partyUp': true}
+          );
+          expect(rootScope.openModal).to.be.calledOnce;
+          expect(rootScope.openModal).to.be.calledWith(
+            'achievements/partyUp',
+            { controller: 'UserCtrl' }
```
crookedneighbor: This will fail because it's being called with `{controller:'UserCtrl', size:'sm'}`. You could do:

``` js
expect(rootScope.openModal).to.be.calledWith(
  'achievements/partyUp',
  sinon.match({ controller: 'UserCtrl' })
)
```

or just

```
expect(rootScope.openModal).to.be.calledWith('achievements/partyUp');
```

I don't think it's particularly important to know that the modal was opened with a particular controller, and, in fact, all the achievement modal stuff should probably be abstracted into a service. (not for this pr)

test/spec/controllers/partyCtrlSpec.js
```diff
@@ -33,6 +36,101 @@ describe("Party Controller", function() {
...
+    });
+
+    context('party has 4 members', function() {
+      beforeEach(function() {
+        scope.group = { memberCount: 4 };
+      });
+
+      context('user has "Party Up" but not "Party On" achievement', function() {
+        beforeEach(function() {
+          user.achievements.partyUp = true;
```
crookedneighbor, line 95: I don't find doing this kind of setup in a beforeEach very useful when there is only one it block.

test/spec/controllers/partyCtrlSpec.js
```diff
@@ -33,6 +36,101 @@ describe("Party Controller", function() {
...
+    context('party has 2 members', function() {
+      beforeEach(function() {
+        scope.group = { memberCount: 2 };
+      });
+
+      context('user does not have "Party Up" achievement', function() {
+        it('awards "Party Up" achievement', function() {
+          expect(User.set).to.be.calledOnce;
+          expect(User.set).to.be.calledWith(
+            { 'achievements.partyUp': true}
```
crookedneighbor: The curly braces spacing does not match. `true}` should be `true }`

test/spec/controllers/partyCtrlSpec.js
```diff
@@ -33,6 +36,101 @@ describe("Party Controller", function() {
...
+          user.achievements.partyUp = true;
+          user.achievements.partyOn = true;
+        });
+
+        it('awards no new achievements', function() {
+          expect(User.set).to.not.be.called;
+          expect(rootScope.openModal).to.not.be.called;
+        });
+      });
+
```
crookedneighbor: extra line break here.

khipkin:
@crookedneighbor I did what you suggested, but the tests still fail.

crookedneighbor:
That's because I forgot about the bit where you need to inject the $state service and stub the is method. :)

Will share a code snippet tomorrow. Heading to bed now.

khipkin:
@crookedneighbor You're the bomb dot com!  The tests all pass!

SabreCat:
Wonderful! Thank you both!

PR #6894, opened 2016-03-16
Added development controller, middleware, and initial tests

TheHollidayInn:
These are for the "Dev routes (in separate controller)" items in #6776.

I took some liberty with the naming on these methods/controller, so let me know if those need to be changed.

Also, I am interested in adding more Middleware/Model Binding to our routes to reduce code reuse, so I hope this was a good scenario to add a middleware.

### DIFF
--- a/test/api/v3/integration/debug/POST-debug_addTenGems.test.js
+++ b/test/api/v3/integration/debug/POST-debug_addTenGems.test.js
@@ -3,15 +3,19 @@ import {
   generateUser,
 } from '../../../../helpers/api-v3-integration.helper';
 
-describe('POST /development/addTenGems', () => {
+describe('POST /debug/add-ten-gems', () => {
   let userToGainTenGems;
 
   before(async () => {
     userToGainTenGems = await generateUser();
   });
 
+  after(() => {
+    nconf.set('IS_PROD', false);
+  });
+
   it('adds ten gems to the current user', async () => {
-    await userToGainTenGems.post('/development/addTenGems');
+    await userToGainTenGems.post('/debug/add-ten-gems');
 
     let userWithTenGems = await userToGainTenGems.get('/user');
 
@@ -21,15 +25,11 @@ describe('POST /development/addTenGems', () => {
   it('returns error when not in production mode', async () => {
     nconf.set('IS_PROD', true);
 
-    await expect(userToGainTenGems.post('/development/addTenGems'))
+    await expect(userToGainTenGems.post('/debug/add-ten-gems'))
     .eventually.be.rejected.and.to.deep.equal({
       code: 404,
       error: 'NotFound',
       message: 'Not found.',
     });
-  });
-
-  after(() => {
-    nconf.set('IS_PROD', false);
   });
 });
--- a/test/api/v3/integration/development/POST-development_addHourglass.test.js
+++ b/test/api/v3/integration/development/POST-development_addHourglass.test.js
@@ -1,0 +1,35 @@
+import nconf from 'nconf';
+import {
+  generateUser,
+} from '../../../../helpers/api-v3-integration.helper';
+
+describe('POST /development/addHourglass', () => {
+  let userToGetHourGlass;
+
+  before(async () => {
+    userToGetHourGlass = await generateUser();
+  });
+
+  it('adds Hourglass to the current user', async () => {
+    await userToGetHourGlass.post('/development/addHourglass');
+
+    let userWithHourGlass = await userToGetHourGlass.get('/user');
+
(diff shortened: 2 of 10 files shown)


crookedneighbor:
I think this is great! What do you think about namespacing them as `debug` instead of `development`?

crookedneighbor:
:100: for use of middleware.

website/src/middlewares/api-v3/developmentMode.js
```diff
@@ -0,0 +1,12 @@
+import nconf from 'nconf';
```
crookedneighbor: This file should have a unit test. Should be able to just stub the result of `nconf.get` for it.

test/api/v3/integration/development/POST-development_addTenGems.test.js
```diff
@@ -0,0 +1,35 @@
...
+
+    await expect(userToGainTenGems.post('/development/addTenGems'))
+    .eventually.be.rejected.and.to.deep.equal({
+      code: 404,
+      error: 'NotFound',
+      message: 'Not found.',
+    });
+  });
+
+  after(() => {
```
crookedneighbor: This after should be at the top with the before.

test/api/v3/integration/development/POST-development_addHourglass.test.js
```diff
@@ -0,0 +1,35 @@
...
+
+    await expect(userToGetHourGlass.post('/development/addHourglass'))
+      .eventually.be.rejected.and.to.deep.equal({
+        code: 404,
+        error: 'NotFound',
+        message: 'Not found.',
+      });
+  });
+
+  after(() => {
```
crookedneighbor: This after should be at the top with the before.

website/src/middlewares/api-v3/ensureDevelpmentMode.js
```diff
@@ -0,0 +1,12 @@
+import nconf from 'nconf';
+import {
+  NotFound,
+} from '../../libs/api-v3/errors';
+
+module.exports = function ensureDevelpmentMode (req, res, next) {
+  if (nconf.get('IS_PROD')) {
```
paglias, line 7: let's define IS_PROD as a const at the beginning of the file instead of calling nconf.get at every request

test/api/v3/unit/middlewares/ensureDevelpmentMode.js
```diff
@@ -0,0 +1,35 @@
...
+
+    expect(next).to.be.calledWith(new NotFound());
+  });
+
+  it('passes when not in production', () => {
+    sandbox.stub(nconf, 'get').withArgs('IS_PROD').returns(false);
+
+    ensureDevelpmentMode(req, res, next);
+
+    expect(next).to.be.calledOnce;
```
crookedneighbor, line 33: This doesn't test that it wasn't called with `new NotFound()`

test/api/v3/integration/debug/POST-debug_addTenGems.test.js
```diff
@@ -0,0 +1,35 @@
...
+
+    await expect(userToGainTenGems.post('/debug/add-ten-gems'))
+    .eventually.be.rejected.and.to.deep.equal({
+      code: 404,
+      error: 'NotFound',
+      message: 'Not found.',
+    });
+  });
+
+  after(() => {
```
crookedneighbor: Move this after up with the before.

TheHollidayInn: Eh, I thought I moved both of these. Sorry.

test/api/v3/unit/middlewares/ensureDevelpmentMode.js
TheHollidayInn, line 33: Is there a calls without?

crookedneighbor, line 33: I do not see one in the documentation. There's probably a helper for it. Or you could assert on the args array. I'll look at it a little later.

test/api/v3/unit/middlewares/ensureDevelpmentMode.js
```diff
@@ -0,0 +1,36 @@
...
+    expect(next.args[0]).to.be.empty;
```
TheHollidayInn, line 34: I added this expectation, @crookedneighbor . I'm unsure if it is the best way, but it does confirm that NotFound was not called.

crookedneighbor, line 34: Looks fine to me.

PR #6896, opened 2016-03-16
[V3] WIP - User update routes

crookedneighbor:
Noticed that the update routes we have written so far are set up as POST routes, when they should be PUT routes. 

Also, shouldn't the stuff related to username and password be namespaced under `/user/auth`?

Will continue to work on the actual User update routes, but wanted to get this out there first for some some opinions.

### DIFF
--- a/test/api/v3/integration/user/PUT-user_update_email.test.js
+++ b/test/api/v3/integration/user/PUT-user_update_email.test.js
@@ -4,10 +4,10 @@ import {
 } from '../../../../helpers/api-v3-integration.helper';
 import { model as User } from '../../../../../website/src/models/user';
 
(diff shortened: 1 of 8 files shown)


paglias:
@crookedneighbor : 

> Also, shouldn't the stuff related to username and password be namespaced under /user/auth?

Yes, I think we should move it

> Noticed that the update routes we have written so far are set up as POST routes, when they should be PUT routes.

:+1: 

One thing I want to point out about the user update routes: to avoid having to rewrite all the client side we'll want to have a catch all update route like we did in v2 that'll handle all the `user.set()` ops

crookedneighbor:
Going to merge this in now. Just a few refactors for how things are organized.

PR #6898, opened 2016-03-16
[v3] Port several shared ops and their linked fns

paglias:
WIP

This PR rewrites:
- ops/buy
- ops/buyMysterySet
- ops/buyQuest
- buySpecialSpell
- fns/handleTwoHanded
- fns/randomVal
- fns/predictableRandom
- fns/ultimateGear
- fns/autoAllocate
- ops/allocateNow

Missing tests and server part.


common/script/fns/ultimateGear.js
```diff
@@ -1,33 +1,35 @@
...
         return soFarGood && (!found || owned[found.key] === true);
       }, true);
     }
   });
-  if (typeof user.markModified === "function") {
-    user.markModified('achievements.ultimateGearSets');
-  }
+
+  // TODO
+  if (user.markModified) user.markModified('achievements.ultimateGearSets');
```
paglias: @crookedneighbor I'm probably changing my stance about moving markModified out of the shared code... it's a mess when it's not called directly by an ops but from a nested function (ops.buy -> ultimateGear -> markModified)

test/common/ops/buy.js
```diff
@@ -1,14 +1,28 @@
...
+import randomVal from '../../../common/script/fns/randomVal';
+import content from '../../../common/script/content/index';
+import {
+  NotAuthorized,
+} from '../../../common/script/libs/errors';
+import i18n from '../../../common/script/i18n';
+
+describe('shared.ops.buy', () => {
   let user;
+  let fns = {
```
crookedneighbor: When stubbing, this is not the same object that is being used in the source files. So it's not actually stubbing `predictableRandom` or `randomVal`. What I would do is create an `index.js` in the `fns` and `ops` folder that just export all the `fns` and `ops`, make those your entry point in the code and the tests. Then you can stub as needed.

test/common/ops/buy.js
```diff
@@ -59,7 +68,11 @@ describe('user.fns.buy', () => {
     it('does not purchase if not enough gp', () => {
       user.stats.hp = 45;
       user.stats.gp = 5;
-      user.ops.buy({params: {key: 'potion'}});
+      try {
```
crookedneighbor, line 66: Pretty sure all these assertions don't actually work as you expect them to. 

The expect...throw syntax supposed to be written like:

``` js
// function without arguments
let aFunctionWithoutArgsThatThrowsAnError = () => {/* do stuff, throw an error */};
let error = new NotAuthorized(); // might need message as well, not sure
expect(aFunctionWithoutArgsThatThrowsAnError).to.throw(error);

// function with arguments
let aFunctionWithArgsThatThrowsAnError = (arg1, arg2) => {/* do stuff, throw an error */};
let error = new NotAuthorized(); // might need message as well, not sure
expect(() => {
  aFunctionWithArgsThatThrowsAnError('foo', 'bar');
}).to.throw(error);
```

And it stops the error from being thrown. Meaning you don't need the try catch.

The way you have it written, I'm pretty sure the `expect...throw` assertion is never being evaluated. It's just throwing normally, hence why you needed the `try...catch`

common/script/ops/buy.js
```diff
@@ -3,117 +3,133 @@ import i18n from '../i18n';
 import _ from 'lodash';
 import count from '../count';
 import splitWhitespace from '../libs/splitWhitespace';
+import {
+  BadRequest,
+  NotAuthorized,
+  NotFound,
+} from '../libs/errors';
+import predictableRandom from '../fns/predictableRandom';
```
crookedneighbor, line 11: The reason it's still not working is because you're referencing the object this way instead of `shared.fns.predictableRandom` or `fns.predictableRandom`. Sinon is wrapping the the parent object and replacing the specific function. In this case you're calling the function directly, so it's not being wrapped. 

If these were imported from a `fns` object, it should work.

paglias:
Added routes and integration tests

paglias:
Added 
- fns/autoAllocate
- ops/allocateNow

with passing tests

PR #6907, opened 2016-03-17
[v3] Add Party Up / Party On achievements to server in V3

SabreCat:
Logic to award the `partyUp` and `partyOn` achievements via the party join route instead of the heavily client-dependent current implementation.

To dos / questions:
- [ ] Remove current modal logic and create an Angular watch to display the achievement
- [x] Fix failing tests. It looks to me like the achievements aren't available on the user model for some reason?
- [x] This first draft logic is inefficient in that it will `$set` the achievement for party members who already have it. But I'm not sure if GETting the whole party's achievements to check them first would be a great move either. Thoughts?

### DIFF
--- a/test/api/v3/integration/groups/POST-groups.test.js
+++ b/test/api/v3/integration/groups/POST-groups.test.js
@@ -189,6 +189,17 @@ describe('POST /group', () => {
       expect(updatedUser.party._id).to.eql(party._id);
     });
 
+    it('does not award Party Up achievement to solo partier', async () => {
+      await user.post('/groups', {
+        name: partyName,
+        type: partyType,
+      });
+
+      let updatedUser = await user.get('/user');
+
+      expect(updatedUser.achievements.partyUp).to.not.eql(true);
+    });
+
     it('prevents user in a party from creating another party', async () => {
       await user.post('/groups', {
         name: partyName,
--- a/test/api/v3/integration/groups/POST-groups_groupId_join.test.js
+++ b/test/api/v3/integration/groups/POST-groups_groupId_join.test.js
@@ -225,4 +225,53 @@ describe('POST /group/:groupId/join', () => {
       });
     });
   });
+
+  context('Party incentive achievements', () => {
+    let leader, member, party;
+
+    beforeEach(async () => {
+      leader = await generateUser();
+      member = await generateUser();
+      party = await leader.post('/groups', {
+        name: 'Testing Party',
+        type: 'party',
+      });
+      await leader.post(`/groups/${party._id}/invite`, {
+        uuids: [member._id],
+      });
+      await member.post(`/groups/${party._id}/join`);
+    });
+
+    it('awards Party Up achievement to party of size 2', async () => {
+      await member.sync();
+      await leader.sync();
+
+      expect(member).to.have.deep.property('achievements.partyUp', true);
+      expect(leader).to.have.deep.property('achievements.partyUp', true);
+    });
+
+    it('does not award Party On achievement to party of size 2', async () => {
+      await member.sync();
+      await leader.sync();
+
+      expect(member).to.not.have.deep.property('achievements.partyOn');
+      expect(leader).to.not.have.deep.property('achievements.partyOn');
+    });
+
+    it('awards Party On achievement to party of size 4', async () => {
+      let addlMemberOne = await generateUser();
+      let addlMemberTwo = await generateUser();
+      await leader.post(`/groups/${party._id}/invite`, {
+        uuids: [addlMemberOne._id, addlMemberTwo._id],
+      });
+      await addlMemberOne.post(`/groups/${party._id}/join`);
+      await addlMemberTwo.post(`/groups/${party._id}/join`);
+
+      await member.sync();
+      await leader.sync();
+
+      expect(member).to.have.deep.property('achievements.partyOn', true);
+      expect(leader).to.have.deep.property('achievements.partyOn', true);
+    });
+  });
 });
--- a/website/src/controllers/api-v3/groups.js
+++ b/website/src/controllers/api-v3/groups.js
@@ -295,6 +295,12 @@ api.joinGroup = {
 
     if (group.type === 'party' && inviter) {
       promises.push(User.update({_id: inviter}, {$inc: {'items.quests.basilist': 1}}).exec()); // Reward inviter
+      if (group.memberCount > 1) {
+        promises.push(User.update({$or: [{'party._id': group._id}, {_id: user._id}], 'achievements.partyUp': {$ne: true}}, {$set: {'achievements.partyUp': true}}, {multi: true}).exec());
+      }
+      if (group.memberCount > 3) {
+        promises.push(User.update({$or: [{'party._id': group._id}, {_id: user._id}], 'achievements.partyOn': {$ne: true}}, {$set: {'achievements.partyOn': true}}, {multi: true}).exec());
+      }
     }
 
     await Q.all(promises);


crookedneighbor:
Regarding the note about the angular stuff, the client doesn't run on the api-v3 branch.

test/api/v3/integration/groups/POST-groups.test.js
```diff
@@ -189,6 +189,17 @@ describe('POST /group', () => {
...
 
+    it('does not award Party Up achievement to solo partier', async () => {
+      await user.post('/groups', {
+        name: partyName,
+        type: partyType,
+      });
+
+      let updatedUser = await user.get('/user');
+
+      expect(updatedUser.achievements.partyUp).to.not.be(true);
```
crookedneighbor: `eql(true)` not `be(true)`

website/src/controllers/api-v3/groups.js
```diff
@@ -292,9 +292,20 @@ api.joinGroup = {
...
 
     if (group.type === 'party' && inviter) {
       promises.push(User.update({_id: inviter}, {$inc: {'items.quests.basilist': 1}}).exec()); // Reward inviter
+      if (group.memberCount > 3) {
+        partyAchievements = {$set: {'achievements.partyUp': true, 'achievements.partyOn': true}};
+      } else if (group.memberCount > 1) {
+        partyAchievements = {$set: {'achievements.partyUp': true}};
+      }
+      if (partyAchievements) {
+        _.forEach(group.members, (member) => {
```
crookedneighbor: Members aren't on the group object anymore.

SabreCat: Shoot, OK. Should I call GET/groups/:groupId/members then?

paglias: I think you'll want to do something like `promises.push(User.update({'party.id': group._id}, partyAchievements, {multi: true}).exec());`, check the syntax because I'm not sure it's correct

SabreCat: Oho, of course! Actually, with a bit of rearranging, I could use the Mongo query to only update the users who don't have the achievement yet, too.

SabreCat:
Made the suggested changes (though I might further improve on the achievement award bit based on @TheHollidayInn's suggestion), but still failing. This is the line that has me puzzled about why most achievements aren't showing up on the test user: https://travis-ci.org/HabitRPG/habitrpg/builds/117550688#L6932

test/api/v3/integration/groups/POST-groups.test.js
```diff
@@ -189,6 +189,17 @@ describe('POST /group', () => {
       expect(updatedUser.party._id).to.eql(party._id);
     });
 
+    it('does not award Party Up achievement to solo partier', async () => {
+      await user.post('/groups', {
+        name: partyName,
+        type: partyType,
+      });
+
+      let updatedUser = await user.get('/user');
```
paglias, line 198: instead of updatedUser you can do `await user.sync()` and the original user object will get updated

paglias:
@SabreCat the last failing test is not your fault, try rebasing to the api-v3 branch

test/api/v3/integration/groups/POST-groups.test.js
SabreCat, line 198: This is the same pattern used in tests throughout the file, though. Do we want to update the rest of the tests to use that approach?

paglias, line 198: No, too much work, just to let you know you could do it this way

paglias:
@SabreCat now it's failing because of two linting errors

SabreCat:
Fixed the linting and rebased, but the tests are still failing.

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
```diff
@@ -225,4 +225,46 @@ describe('POST /group/:groupId/join', () => {
...
+    });
+
+    it('awards Party Up achievement to party of size 2', async () => {
+      await invitedUsers[0].post(`/groups/${party._id}/join`);
+
+      await expect(invitedUsers[0].get('/user')).to.eventually.have.deep.property('achievements.partyUp', true);
+      await expect(user.get('/user')).to.eventually.have.deep.property('achievements.partyUp', true);
+    });
+
+    it('does not award Party On achievement to party of size 2', async () => {
```
paglias, line 253: I think this one is failing because it expects the property to be there but not to be true and in this case it's missing completely

paglias:
@SabreCat @crookedneighbor can confirm this but I think it's failing because the `createAndPopulateGroup` doesn't use the server routes but acts directly on Mongo

crookedneighbor:
This is the code for creating and populating a group. https://github.com/HabitRPG/habitrpg/blob/api-v3/test/helpers/api-integration/v3/object-generators.js#L66-L108

It uses the route to create the group, but mongo to update the member count and membership of the users.

website/src/controllers/api-v3/groups.js
```diff
@@ -295,6 +295,12 @@ api.joinGroup = {
 
     if (group.type === 'party' && inviter) {
       promises.push(User.update({_id: inviter}, {$inc: {'items.quests.basilist': 1}}).exec()); // Reward inviter
+      if (group.memberCount > 3) {
+        promises.push(User.update({'party._id': group._id, 'achievements.partyOn': false}, {$set: {'achievements.partyOn': true}}, {multi: true}).exec());
```
paglias: multi: true here is not needed

crookedneighbor: Can you break down when multi is necessary and when it is not?

paglias: Stupid me... I thought it was searching for a group not for multiple groups.

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
SabreCat, line 253: Yeah, that's what I keep saying (or trying to say but failing to communicate?). For some reason, the user doesn't have all the achievement properties.

paglias, line 253: it's normal that it doesn't have them because you don't specify any value. They're available when loading the user with mongoose but then when they're sent down as json if there's no value they get discarded. The only way it to add a default value here https://github.com/HabitRPG/habitrpg/blob/api-v3/website/src/models/user.js#L88

paglias:
@SabreCat If you don't use the createAndPopulate helper but make manual requests to the API it'll work

SabreCat:
OK, I think I fixed it! I did change the tests to use direct API requests rather than `createAndPopulateGroup`, but the problems were functional in the changes to the join party route, not in the test.
1. I was checking for a `false` value for the achievements to limit the Mongo update query, but given the lack of a default, I needed to check for non-true, not equal-false.
2. At that point in the invite process, the newly joining user doesn't have the party ID in their data yet, so the query was finding the existing folks but not the newcomer.

I also changed the `POST-groups_groupId_join` tests to use the `sync()` approach @TheHollidayInn suggested; I do find it a little more readable that way.

website/src/controllers/api-v3/groups.js
```diff
@@ -295,6 +295,12 @@ api.joinGroup = {
 
     if (group.type === 'party' && inviter) {
       promises.push(User.update({_id: inviter}, {$inc: {'items.quests.basilist': 1}}).exec()); // Reward inviter
+      if (group.memberCount > 1) {
+        promises.push(User.update({$or: [{'party._id': group._id}, {_id: user._id}], 'achievements.partyUp': {$nin: [true]}}, {$set: {'achievements.partyUp': true}}, {multi: true}).exec());
```
paglias: instead of $nin I think here you could use $ne https://docs.mongodb.org/manual/reference/operator/query/ne/

PR #6908, opened 2016-03-17
api-v3 password reset

piousbox:


test/api/v3/integration/user/auth/POST-user_reset_password.test.js
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  it('same message on error as on success', async () => {
+    let response = await user.post(endpoint, {
+      email: 'nonExistent@email.com',
+    });
+    expect(response).to.eql({code: 200, message: t('passwordReset')});
+  });
+
+  it('errors is email is not provided', async () => {
```
crookedneighbor: s/is/if

website/src/controllers/api-v3/user.js
```diff
@@ -82,6 +84,50 @@ api.updatePassword = {
...
+      sendEmail({
+        from: 'Habitica <admin@habitica.com>',
+        to: email,
+        subject: 'Password Reset for Habitica',
+        text: `Password for ${user.auth.local.username} has been reset to ${newPassword} . Important! Both username and password are case-sensitive -- you must enter both exactly as shown here. We recommend copying and pasting both instead of typing them. Log in at ${nconf.get('BASE_URL')}. After you have logged in, head to ${nconf.get('BASE_URL')}/#/options/settings/settings and change your password.`,
+        html: `Password for <strong>${user.auth.local.username}</strong> has been reset to <strong>${newPassword}</strong><br /><br />Important! Both username and password are case-sensitive -- you must enter both exactly as shown here. We recommend copying and pasting both instead of typing them.<br /><br />Log in at ${nconf.get('BASE_URL')}. After you have logged in, head to ${nconf.get('BASE_URL')}/#/options/settings/settings and change your password.`,
+      });
+      await user.save();
+    }
+    res.respond(300, { message: res.t('passwordReset') });
```
paglias: 200?

website/src/controllers/api-v3/user.js
```diff
@@ -82,6 +84,50 @@ api.updatePassword = {
...
+  async handler (req, res) {
+    req.checkBody({
+      email: {
+        notEmpty: {errorMessage: res.t('missingEmail')},
+      },
+    });
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let email = req.body.email && req.body.email.toLowerCase();
```
paglias: no need to check for req.body.email, you already did it at line 100

website/src/controllers/api-v3/user.js
```diff
@@ -82,6 +84,50 @@ api.updatePassword = {
...
+      user.auth.local.hashed_password = hashedPassword; // eslint-disable-line camelcase
+      sendEmail({
+        from: 'Habitica <admin@habitica.com>',
+        to: email,
+        subject: 'Password Reset for Habitica',
+        text: `Password for ${user.auth.local.username} has been reset to ${newPassword} . Important! Both username and password are case-sensitive -- you must enter both exactly as shown here. We recommend copying and pasting both instead of typing them. Log in at ${nconf.get('BASE_URL')}. After you have logged in, head to ${nconf.get('BASE_URL')}/#/options/settings/settings and change your password.`,
+        html: `Password for <strong>${user.auth.local.username}</strong> has been reset to <strong>${newPassword}</strong><br /><br />Important! Both username and password are case-sensitive -- you must enter both exactly as shown here. We recommend copying and pasting both instead of typing them.<br /><br />Log in at ${nconf.get('BASE_URL')}. After you have logged in, head to ${nconf.get('BASE_URL')}/#/options/settings/settings and change your password.`,
+      });
+      await user.save();
+    }
```
paglias: if not user return an error

website/src/controllers/api-v3/user.js
```diff
@@ -82,6 +84,50 @@ api.updatePassword = {
...
+    let user = await User.findOne({ 'auth.local.email': email }, { 'auth.local': 1 });
+
+    if (user) {
+      user.auth.local.salt = salt;
+      user.auth.local.hashed_password = hashedPassword; // eslint-disable-line camelcase
+      sendEmail({
+        from: 'Habitica <admin@habitica.com>',
+        to: email,
+        subject: 'Password Reset for Habitica',
+        text: `Password for ${user.auth.local.username} has been reset to ${newPassword} . Important! Both username and password are case-sensitive -- you must enter both exactly as shown here. We recommend copying and pasting both instead of typing them. Log in at ${nconf.get('BASE_URL')}. After you have logged in, head to ${nconf.get('BASE_URL')}/#/options/settings/settings and change your password.`,
```
paglias: let's move these strings to locales/en/api-v3.json

crookedneighbor:
This route should be in the auth file.

website/src/controllers/api-v3/user.js
piousbox: If we do that, it will allow anyone to get emails of all the users in the system. If the user is not found, the response should be the same as if the user if found in this case. You agree?

piousbox:
@crookedneighbor  Re: "This route should be in the auth file." - can we talk about that? The user file, not the auth file, currently has the methods for managing username, email, and password. The auth file has methods for creating registrations (new user) and sessions (login). This route, password reset, assumes the user already exists. I'd say it belongs to user file.

paglias:
@piousbox we moved the username, emails and password routes to the auth file

website/src/controllers/api-v3/user.js
paglias: :+1:

paglias:
Looking good to me :+1:

PR #6918, opened 2016-03-18
Added tests for challenge task scoring and moved scoring code to model

TheHollidayInn:
These are for the "test task scoring for challenges tasks" items in #6776.

I only tested the section @paglias referred to after the `try` statement. Which is where the challenge scoring logic lies.

I also move the challenge score logic to the Task model. And, I took the generateHistory function from the preening tests and placed it in the helpers library.

### DIFF
--- a/website/src/models/task.js
+++ b/website/src/models/task.js
@@ -4,6 +4,7 @@ import validator from 'validator';
 import moment from 'moment';
 import baseModel from '../libs/api-v3/baseModel';
 import _ from 'lodash';
+import { preenHistory } from '../libs/api-v3/preening';
 
 let Schema = mongoose.Schema;
 let discriminatorOptions = {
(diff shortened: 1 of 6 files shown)


website/src/models/task.js
```diff
@@ -75,6 +76,39 @@ TaskSchema.statics.sanitizeReminder = function sanitizeReminder (reminderObj) {
...
+    } else {
+      chalTask.history.push({
+        date: Number(new Date()),
+        value: chalTask.value,
+      });
+
+      // Only preen task history once a day when the task is scored first
+      if (chalTask.history.length > 365) {
+        chalTask.history = preenHistory(chalTask.history, true); // true means the challenge will retain as much entries as a subscribed user
+        chalTask.markModified(`history.${chalTask.history.length - 1}`);
```
paglias: here the markModified is probably not necessary since the whole array is replaced, if it's needed then it should mark modified the whole array not the last item only

TheHollidayInn: That's true. Turns out we don't need it! Update coming.

PR #6923, opened 2016-03-19
[v3] Added group reject invite route and initial tests

TheHollidayInn:
These are for the "reject group invite" items in #6776.

I added a todo. Not sure if we want to send a notification when someone rejects an invitation. If not, I can just amend that commit.

### DIFF
--- a/website/src/controllers/api-v3/groups.js
+++ b/website/src/controllers/api-v3/groups.js
@@ -308,6 +308,53 @@ api.joinGroup = {
 };
 
 /**
+ * @api {post} /groups/:groupId/reject Reject a group invitation
+ * @apiVersion 3.0.0
+ * @apiName RejectGroupInvite
+ * @apiGroup Group
+ *
+ * @apiParam {UUID} groupId The group _id
+ *
+ * @apiSuccess {Object} group The group
+ */
+api.rejectGroupInvite = {
+  method: 'POST',
+  url: '/groups/:groupId/reject-invite',
+  middlewares: [authWithHeaders(), cron],
+  async handler (req, res) {
+    let user = res.locals.user;
+
+    req.checkParams('groupId', res.t('groupIdRequired')).notEmpty().isUUID();
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let group = await Group.getGroup({user, groupId: req.params.groupId, optionalMembership: true}); // Do not fetch chat and work even if the user is not yet a member of the group
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    let isUserInvited = false;
+
+    if (group.type === 'party' && group._id === user.invitations.party.id) {
+      user.invitations.party = {};
+      user.markModified('invitations.party');
+      isUserInvited = true;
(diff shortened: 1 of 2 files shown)


website/src/controllers/api-v3/groups.js
```diff
@@ -308,6 +308,57 @@ api.joinGroup = {
...
+ * @api {post} /groups/:groupId/reject Reject a group invitation
+ * @apiVersion 3.0.0
+ * @apiName RejectGroupInivite
+ * @apiGroup Group
+ *
+ * @apiParam {UUID} groupId The group _id
+ *
+ * @apiSuccess {Object} group The group
+ */
+api.rejectGroupInivite = {
```
paglias: typo here, `inivIte`

website/src/controllers/api-v3/groups.js
```diff
@@ -308,6 +308,57 @@ api.joinGroup = {
...
+  middlewares: [authWithHeaders(), cron],
+  async handler (req, res) {
+    let user = res.locals.user;
+
+    req.checkParams('groupId', res.t('groupIdRequired')).notEmpty().isUUID();
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let group = await Group.getGroup({user, groupId: req.params.groupId, optionalMembership: true}); // Do not fetch chat and work even if the user is not yet a member of the group
```
paglias: to not fetch chat here you'll have to specify `{fields:`-chat`}`

TheHollidayInn: Thanks. I'm unsure if we need to fetch the group at all. I think if someone rejects, they don't need the group info. I just used the return code from the accept invite. Any thoughts?

paglias: you're right, let's not fetch the group at all and return an empty object. I'm not even sure if this route belongs in the groups controller, what about moving it to the user controller?

paglias: Let's keep the route here, with this url it would be confusing to move it to the user controller. But let's not fetch the group and just return an empty object

TheHollidayInn: Yea, definitely keep the route here. I know we notify the user, but the concept is that we are modifying the group. So, from a UX point of view. We should stay here.

TheHollidayInn:
Updated!

website/src/controllers/api-v3/groups.js
paglias: I think you removed the group from the response but we're still loading it at line 332 :)

TheHollidayInn: Oh, I was loading some group data before the response that I removed. I actually need to here to make sure the group exists. In fact, I think we should add a middleware to check for group existence when calling a :groupId route.

paglias: In this case I think we don't need to check that the group exists, just loop through all invitations and when you find a group id that match the one in the parameter, remove that invite. If no matching invite is returned than you can throw an error

TheHollidayInn: That make sense too. ha

Issue #4888, opened 2015-03-19
Popup after class select (change class) shows wrong class - reports that I've equipped old class's gear instead of new


gkjt:
I just hit level 10, and chose my class, Rogue, but the popup after tells me I am now wearing my Warrior gear. It appears to have applied the rogue equipment, but the popup is incorrect. You can see this in [this](http://puu.sh/gGY47/a541c9b2ad.png) screenshot, also attached below:

![Incorrect class on popup](http://puu.sh/gGY47/a541c9b2ad.png)
##

TheHollidayInn:
Technical: I believe this happens because the user class is not synced before the Guide is called after this line: https://github.com/HabitRPG/habitrpg/blob/de67f44e1f0b6027871397610a37458a87cfdc91/website/public/js/controllers/userCtrl.js#L29 .

I'm unsure of how to use a callback to call the guide after the user has synced.

relsqui:
Adding a couple of notes from my (duplicate) #5430:

The message is this one: https://github.com/HabitRPG/habitrpg/blob/4042c1ce7ae07c84da927501d7915cd828a10be3/common/locales/en/npc.json#L31
It seems to be getting its variable here: https://github.com/HabitRPG/habitrpg/blob/1743899dd49adae7887f73c300f9802a35307b9b/website/public/js/services/guideServices.js#L60

The reported class seems to be the one the user had at last refresh, which is not necessarily the class they're changing from. (It's different if you change classes twice in one pageload.)

ghost:
In addition

`content: window.env.t('classGearText', {klass: User.user.stats.class})`

has `klass` untranslatable

( https://github.com/HabitRPG/habitrpg/blob/develop/website/public/js/services/guideServices.js#L70 )

ghost:
Oh, the ticket is almost year-old... But is it really necessary to point out the class in that message?

Maybe simplify it and instead of 

> First: don't panic! Your old gear is in your inventory, and you're now wearing **_your apprentice <strong><%= klass %></strong> equipment**_. Wearing your class's gear grants you a 50% bonus to stats. However, feel free to switch back to your old gear."

just say something general like:

> First: don't panic! Your old gear is in your inventory, and you're now wearing **_the apprentice equipment of the class you have just joined**_. Wearing your class's gear grants you a 50% bonus to stats. However, feel free to switch back to your old gear."

GitHubSphinx:
[...]

PR #6924, opened 2016-03-19
Modal explaining new class gear is now class-agnostic

khipkin:
This PR is regarding Issue #4888.

The `classGearText` string is now class-agnostic and the reference to "klass" has been removed.

image


crookedneighbor:

khipkin:
@crookedneighbor cool, thanks!

PR #6927, opened 2016-03-19
[v3] Ported sendPrivateMessage route and added initial tests

TheHollidayInn:
These are for the "sendPrivateMessage and sendMessage" items in #6776.

It was recommended to place this in the inbox routes, but I am unsure of where those are currently located. I think adding these to the user routes might be a good option. one last option would be to separate out and generalize chat.

### DIFF
--- a/common/locales/en/api-v3.json
+++ b/common/locales/en/api-v3.json
@@ -117,5 +117,9 @@
   "missingPetFoodFeed": "\"pet\" and \"food\" are required parameters.",
   "invalidPetName": "Invalid pet name supplied.",
   "missingEggHatchingPotionHatch": "\"egg\" and \"hatchingPotion\" are required parameters.",
-  "invalidTypeEquip": "\"type\" must be one of 'equipped', 'pet', 'mount', 'costume'."
+  "invalidTypeEquip": "\"type\" must be one of 'equipped', 'pet', 'mount', 'costume'.",
+  "cannoyBuyItem": "You can't buy this item",
+  "messageRequired": "A message is required.",
+  "toUserIDRequired": "A toUserId is required",
+  "notAuthorizedToSendMessageToThisUser": "Can't send message to this user."
 }
--- a/common/locales/en/api-v3.json
+++ b/common/locales/en/api-v3.json
@@ -124,5 +124,7 @@
   "notAuthorizedToSendMessageToThisUser": "Can't send message to this user.",
   "privateMessageGiftIntro": "Hello <%= receiverName %>, <%= senderName %> has sent you ",
   "privateMessageGiftGemsMessage": "<%= gemAmount %> gems! ",
-  "privateMessageGiftSubscriptionMessage": "<%= numberOfMonths %> months of subscription! "
+  "privateMessageGiftSubscriptionMessage": "<%= numberOfMonths %> months of subscription! ",
+  "cannotSendGemsToYourself": "Cannot send gems to yourself. Try a subscription instead.",
+  "notEnoughGemsToSend": "Amount must be within 0 and your current number of gems."
 }
--- a/common/locales/en/api-v3.json
+++ b/common/locales/en/api-v3.json
@@ -121,5 +121,8 @@
   "cannoyBuyItem": "You can't buy this item",
   "messageRequired": "A message is required.",
   "toUserIDRequired": "A toUserId is required",
-  "notAuthorizedToSendMessageToThisUser": "Can't send message to this user."
+  "notAuthorizedToSendMessageToThisUser": "Can't send message to this user.",
+  "privateMessageGiftIntro": "Hello <%= receiverName %>, <%= senderName %> has sent you ",
+  "privateMessageGiftGemsMessage": "<%= gemAmount %> gems! ",
+  "privateMessageGiftSubscriptionMessage": "<%= numberOfMonths %> months of subscription! "
 }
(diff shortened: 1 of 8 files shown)


website/src/controllers/api-v3/members.js
```diff
@@ -231,4 +241,76 @@ api.getChallengeMemberProgress = {
...
+ * @apiParam {UUID} toUserId The toUser _id
+ *
+ * @apiSuccess {} Object Returns an empty object
+ */
+api.sendPrivateMessage = {
+  method: 'POST',
+  url: '/send-private-message',
+  middlewares: [authWithHeaders(), cron],
+  async handler (req, res) {
+    req.checkBody('message', res.t('messageRequired')).notEmpty();
```
TheHollidayInn, line 256: Can anyone tell me how to return this custom message? I think testing and receiving these specific messages is crucial to a good API.

paglias, line 256: you can find the messages in response.errors

common/locales/en/api-v3.json
```diff
@@ -102,5 +102,9 @@
   "spellNotOwned": "You don't own this spell.",
   "spellLevelTooHigh": "You must be level <%= level %> to use this spell.",
   "invalidAttribute": "\"<%= attr %>\" is not a valid attribute.",
-  "notEnoughAttrPoints": "You don't have enough attribute points."
+  "notEnoughAttrPoints": "You don't have enough attribute points.",
+  "messageRequired": "A message is required.",
+  "toUserIDRequired": "A toUserId is required",
+  "userNotFound": "The user you are trying to send to can't be found",
```
paglias: I'm pretty sure we already have this message somewhere above

website/src/controllers/api-v3/members.js
```diff
@@ -9,8 +9,18 @@ import { model as Group } from '../../models/group';
 import { model as Challenge } from '../../models/challenge';
 import {
   NotFound,
+  NotAuthorized,
 } from '../../libs/api-v3/errors';
 import * as Tasks from '../../models/task';
+import { chatDefaults } from '../../models/group';
+import refPush from '../../../../common/script/libs/refPush.js';
```
paglias: we've started precompiling the server and shared code so you should only import the main `shared` module and then access the submodules from it, like https://github.com/HabitRPG/habitrpg/blob/api-v3/website/src/controllers/api-v3/tasks.js#L400

website/src/controllers/api-v3/members.js
paglias, line 256: which is populated here https://github.com/HabitRPG/habitrpg/blob/api-v3/website/src/middlewares/api-v3/errorHandler.js#L37

website/src/controllers/api-v3/members.js
```diff
@@ -231,4 +241,76 @@ api.getChallengeMemberProgress = {
   },
 };
 
+async function sendMessage (sender, userToReceiveMessage, messageData) {
```
paglias: this function ideally should be in a model, since we don't have an `inbox` model we could put it into the user one maybe?

website/src/controllers/api-v3/members.js
paglias: in this case you can access refPush using `shared.refPush`

website/src/controllers/api-v3/members.js
TheHollidayInn: Yea, that sounds like a good place for it now. It might be good as a Mail/Messenger micro-service in the future.

TheHollidayInn:
Updated!

website/src/controllers/api-v3/members.js
```diff
@@ -231,4 +236,53 @@ api.getChallengeMemberProgress = {
...
+ * @apiGroup Members
+ *
+ * @apiParam {String} message The message
+ * @apiParam {UUID} toUserId The toUser _id
+ *
+ * @apiSuccess {} Object Returns an empty object
+ */
+api.sendPrivateMessage = {
+  method: 'POST',
+  url: '/send-private-message',
```
paglias: I think the route here should be under /members -> /members/send-private-message

website/src/models/user.js
```diff
@@ -702,6 +704,30 @@ schema.methods.getGroups = function getUserGroups () {
...
 };
 
+schema.methods.sendMessage = async function sendMessage (userToReceiveMessage, messageData) {
+  let msg;
+  let sender = this;
+
+  if (!messageData.type) {
+    msg = messageData.message;
+  } else {
+    msg = `Hello ${userToReceiveMessage.profile.name }, ${sender.profile.name} has sent you `;
```
paglias: @TheHollidayInn can you move this to a localized string? using the language in userToReceiveMessage.preferences.language

paglias:
Except for the two comments it looks good!

website/src/models/user.js
TheHollidayInn: Oh, yea.. String params. Good idea :P

tasks/gulp-eslint.js
```diff
@@ -74,7 +74,7 @@ const COMMON_FILES = [
   '!./common/script/libs/planGemLimits.js',
   '!./common/script/libs/preenHistory.js',
   '!./common/script/libs/preenTodos.js',
-  '!./common/script/libs/refPush.js',
+  './common/script/libs/refPush.js',
```
paglias: it's enough to remove the line, the file will be added to linting automatically

TheHollidayInn:
Okay, I removed the line in eslint and fixed the merge conflict.

PR #6934, opened 2016-03-20
[v3] Ported user delete route and added initial tests

TheHollidayInn:
This is for "delete user" in #6776.

### DIFF
--- a/website/src/controllers/api-v3/user.js
+++ b/website/src/controllers/api-v3/user.js
@@ -15,6 +15,7 @@ import { model as User } from '../../models/user';
 import Q from 'q';
 import _ from 'lodash';
 import * as firebase from '../../libs/api-v3/firebase';
+import * as passwordUtils from '../../libs/api-v3/password';
 
 let api = {};
 
@@ -61,6 +62,18 @@ api.deleteUser = {
     let user = res.locals.user;
     let plan = user.purchased.plan;
 
+    req.checkBody({
+      password: {
+        notEmpty: {errorMessage: res.t('missingPassword')},
+      },
+    });
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let oldPassword = passwordUtils.encrypt(req.body.password, user.auth.local.salt);
+    if (oldPassword !== user.auth.local.hashed_password) throw new NotAuthorized(res.t('wrongPassword'));
+
     if (plan && plan.customerId && !plan.dateTerminated) {
       throw new NotAuthorized(res.t('cannotDeleteActiveAccount'));
     }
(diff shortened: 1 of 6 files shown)


website/src/controllers/api-v3/user.js
```diff
@@ -41,6 +45,47 @@ api.getUser = {
...
+    let user = res.locals.user;
+    let plan = user.purchased.plan;
+
+    if (plan && plan.customerId && !plan.dateTerminated) {
+      throw new NotAuthorized(res.t('cannotDeleteActiveAccount'));
+    }
+
+    let types = ['party', 'publicGuilds', 'privateGuilds'];
+    // @TODO: The group leave route doesn't work unless it has these fields. We should probably force the group to get these
+    let groupFields = basicGroupFields.concat(' leader memberCount');
```
paglias, line 83: what do you mean by `We should probably force the group to get these`? also why is the leader populated?

website/src/models/group.js
```diff
@@ -151,6 +151,50 @@ schema.statics.getGroup = async function getGroup (options = {}) {
   return group;
 };
 
+schema.statics.getGroups = async function getGroups (options = {}) {
+  let {user, types, groupFields = basicFields, sort = '-memberCount', populateLeader = false} = options;
+  let queries = [];
+
+  types.forEach(type => {
```
paglias, line 158: I don't like the idea, in this case, of making 4 different queries to get all the groups, it's not really necessary.

paglias:
Except for the comments it's look good. I think in this case we'll want to ask the user for the password (unless they're facebook users) before deleting the account

website/src/models/group.js
paglias, line 158: but it seems that we've been doing the same thing when getting all groups so using different queries is probably okay for now since this route won't be used much

website/src/controllers/api-v3/user.js
TheHollidayInn, line 83: It took some tinkering, but I discovered that you need to populate the leader field for this check to work: https://github.com/HabitRPG/habitrpg/blob/api-v3/website/src/models/group.js#L613 .

It is the same for memberCount and this check: https://github.com/HabitRPG/habitrpg/blob/api-v3/website/src/models/group.js#L597 .

I think it should be the model responsibility to enforce the fields are available for checking. I'm not exactly sure how that would work though.

website/src/models/group.js
TheHollidayInn, line 158: Yea, this can probably be refactored, but I took the query from the getGroups route. I'm not sure who wrote it. But, it might need to be a separate request to update the function.

website/src/controllers/api-v3/user.js
paglias, line 83: are you sure about https://github.com/HabitRPG/habitrpg/blob/api-v3/website/src/models/group.js#L613 ? it's comparing a string to a string, if you populate the leader it'll compare a string to an object and always fail

website/src/models/group.js
paglias, line 158: I think I wrote it, definitely not my best piece of code

paglias:
Last thing before I merge, can you make so the user has to input it's current password before deleting the account (not for facebook-only accounts)

website/src/controllers/api-v3/user.js
TheHollidayInn, line 83: Eh, let me see if I can print out an example. I don't mean populate as in the mongo way. I mean, including the leader in the projection.

TheHollidayInn:
Yea, I haven't made updates to this yet - just replied. I thought asking for password would be something handled on the front-end. Is there an example route for how to ask on the server side?

paglias:
an example is https://github.com/HabitRPG/habitrpg/blob/api-v3/website/src/controllers/api-v3/auth.js#L318

TheHollidayInn:
Added the password requirement. I see what you were saying about populating the leader. I didn't mean to populate the leader. But, I still need to include the leader and memberCount fields in the query project for group.leave to work.

Issue #6912, opened 2016-03-18
Put achievement modal logic in its own service


khipkin:
@crookedneighbor mentioned [here](https://github.com/HabitRPG/habitrpg/pull/6890#discussion_r56442362) that all the achievement modal logic should probably be abstracted into its own service.  I would like to do that!

Alys:
It looks like there's no objections, so please go ahead! Thank you!

PR #6943, opened 2016-03-22
Improvement #6912: New achievement modal service and more achievement tests

khipkin:
(Edited to remove future plans and instead discuss work completed.)

I have added passing tests for awarding some achievements, added an achievement modal service (issue #6912), and updated the achievement tests to make sure they still pass.

My Habitica User ID is f9c89bba-abf8-4f20-a974-cf1550ff603f


khipkin:
Update!  I added the service and the tests still/now pass.

Is there any other functionality I should add or move to the achievement service?

website/public/js/services/achievementServices.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+/**
+ * Services that handle achievement logic.
+ */
+
+angular.module("habitrpg").factory("Achievement",
+['$rootScope', function($rootScope) {
+
+	function openModal(achievementName) {
```
crookedneighbor: Our code style is 2 space indentation, not 8.

website/public/js/services/achievementServices.js
```diff
@@ -0,0 +1,18 @@
...
+		$rootScope.openModal('achievements/' + achievementName,
+			{controller:'UserCtrl', size:'sm'});
+	}
+
+	return {
+		openModal: openModal
```
crookedneighbor: Let's call this function something like `displayAchievement`. That way, if we change the inner works of how we're displaying the achievements, the function name isn't tied to the modal process.

crookedneighbor:
It's awesome that the tests still pass, that's a great indication that your refactor didn't break anything there.

Ultimately we're going to want tests for the new service itself and each of the other tests should test that the achievement service gets called, not that the modal opens. That will allow the controllers to be agnostic about what precisely happens when the service's function gets called. So if we ever changed from modals to some other form of notification, we wouldn't have to update the controller tests because all they care about is that the achievement service was called.

website/public/js/services/achievementServices.js
khipkin: I thought this file looked weird, but couldn't put my finger on _why_...

khipkin:
It looks like `npm install` failed on Travis after the max number of tries.... I don't think that's a result of my changes.

crookedneighbor:
It is not related. Travis and npm are having difficulties right now.

crookedneighbor:
Whelp, this is why: https://medium.com/@azerbike/i-ve-just-liberated-my-modules-9045c06be67c#.tlcdr5ieg

khipkin:
Wow that's extremely unfortunate. Thanks a lot, skeevy corporate lawyers! :disappointed:

crookedneighbor:
I think the only thing left on this PR is to swap out the assertions on the modal opening with assertions that `Achievement.displayAchievement` was called and write tests for the `Achievement` service itself. Correct?

khipkin:
@crookedneighbor correct!  I'm working on it.

test/spec/controllers/notificationCtrlSpec.js
```diff
@@ -84,4 +85,111 @@ describe('Notification Controller', function() {
       expect(rootScope.openModal).to.not.be.called;
     });
   });
+
+  describe('User challenge won modal watch', function() {
```
khipkin: Note that `wonChallenge` is [not treated like a normal "achievement"](https://github.com/khipkin/habitrpg/blob/new_achievement_modal_service/website/public/js/controllers/notificationCtrl.js#L33).  If this is unintentional, I can fix it.

khipkin:
Also, `npm install` is failing again.

khipkin:
@crookedneighbor This is ready for review.

test/spec/controllers/notificationCtrlSpec.js
```diff
@@ -1,33 +1,38 @@
 'use strict';
 
 describe('Notification Controller', function() {
-  var user, scope, rootScope, ctrl;
+  var user, scope, rootScope, fakeBackend, achievement, ctrl;
 
   beforeEach(function() {
     user = specHelper.newUser();
     user._id = "unique-user-id";
+    user.preferences.suppressModals = {};
```
crookedneighbor: This default should be added to the specHelper.newUser function.

test/spec/controllers/notificationCtrlSpec.js
```diff
@@ -1,33 +1,38 @@
...
       $provide.value('User', {user: user});
       $provide.value('Guide', {});
     });
 
-    inject(function(_$rootScope_, _$controller_) {
+    inject(function(_$rootScope_, $httpBackend, _$controller_, Achievement) {
       scope = _$rootScope_.$new();
       rootScope = _$rootScope_;
 
+      fakeBackend = $httpBackend;
```
crookedneighbor, line 29: Is this still needed?

crookedneighbor, line 29: Never mind, looking at the rest of the tests, I can see that it is.

khipkin:
Done.

crookedneighbor:
I don't see this in the manifest. Is it actually getting loaded into the site?

test/spec/controllers/notificationCtrlSpec.js
crookedneighbor: Yes, please adjust this to match the others.

khipkin:
@crookedneighbor I don't know what you mean by "this" and "manifest".

khipkin:
Now that the tests are all passing again, I just need to make `wonChallenge` consistent with the other achievements.

khipkin:
**Mount Master Modal**
image

**Won Challenge Modal**
image

**Ultimate Gear Modal**
image

**Streak Modal**
image

khipkin:
This is ready for review again.

website/client/js/services/achievementServices.js
```diff
@@ -0,0 +1,24 @@
...
+['$rootScope', function($rootScope) {
+
+  function displayAchievement(achievementName) {
+    $rootScope.openModal('achievements/' + achievementName,
+      {controller:'UserCtrl', size:'sm'});
+  }
+
+  function displayBulkyAchievement(achievementName) {
+    $rootScope.openModal('achievements/' + achievementName,
+      {controller:'UserCtrl'});
```
khipkin: I would like some input on whether or not this is the best way to handle passing vs. not passing the `size` value.  I wasn't sure if it was desirable for the caller to specify the modal size, or if it would be better to abstract that away.  Since the goal of this whole PR  is to abstract stuff away, I went with the less transparent option.

However, I think it would be preferable if one of these methods called the other under the hood with somehow modified parameters.  This would make unit testing cleaner and reduce duplicate functionality.

test/spec/controllers/notificationCtrlSpec.js
```diff
@@ -1,33 +1,50 @@
...
 describe('Notification Controller', function() {
-  var user, scope, rootScope, ctrl;
+  var user, scope, rootScope, fakeBackend, achievement, ctrl;
 
   beforeEach(function() {
     user = specHelper.newUser();
     user._id = "unique-user-id";
 
+    var userSync = sinon.stub().returns({
+      then: function then (f) { f(); }
```
khipkin, line 11: I couldn't find a better way to handle this.  Any suggestions?

khipkin:
This is ready for review again.

test/spec/controllers/notificationCtrlSpec.js
crookedneighbor, line 11:
``` js
sinon.stub().returns(Promise.resolve());
```

khipkin, line 11: @crookedneighbor I tried that, but I need the function that gets passed into the `then` to actually get called, otherwise the `wonChallenge` test fails.

khipkin: @crookedneighbor Ping!

khipkin:
@crookedneighbor ?

test/spec/controllers/notificationCtrlSpec.js
khipkin: @crookedneighbor I still can't find a better way to do this.  Except for this weirdness, I believe the PR is done.

website/client/js/services/achievementServices.js
crookedneighbor: I think I'd prefer for it to take an optional second argument that is an object

``` js
function displayAchievement(achievementName, options) {
  options = options || {};

  $rootScope.openModal('achievements/' + achievementName, {
    controller: 'UserCtrl', 
    size: options.size || 'sm'
  });
}
```

crookedneighbor:
Sorry it's been a million years. Left one last comment. After that, I think it's done.

website/client/js/services/achievementServices.js
khipkin: That doesn't work though because the alternative to 'sm' is just to NOT specify a size in the `openModal` call at all.

crookedneighbor: No, the openModal function defaults to `size: 'lg'`

See modal section here https://angular-ui.github.io/bootstrap/#/modal

crookedneighbor:
Wonderful!

website/client/js/services/achievementServices.js
khipkin: In that documentation, they list 'lg' as a possible value, not the default value.  Specifying the 'lg' parameter makes things look huge and weird:

image

khipkin: In the example modal buttons in the documentation, the first modal is with no size param, which is what the larger achievement modals originally used.

Edit:
image

crookedneighbor: Sorry, you're right that the default isn't `lg`. All size does is add an extra class to the modal `modal-${size}`. If you don't pass a size, it doesn't pass anything in. The equivalent size is `md`. So either you can pass `md` as the size or make no size the default and have the small achievements pass in `sm`

``` js
function displayAchievement(achievementName, options) {
  options = options || {};

  $rootScope.openModal('achievements/' + achievementName, {
    controller: 'UserCtrl', 
    size: options.size // if undefined, will be md size
  });
}
```

crookedneighbor: Either one will work for me.

khipkin: Okay things look nice again :smile: :

image

khipkin:
@crookedneighbor This is ready for review again. :smile_cat:

khipkin:
@crookedneighbor :tada: :tada: Thanks! Did I go up a contribution tier?

crookedneighbor:
Just bumped you!

PR #6945, opened 2016-03-23
[v3] working on attaching local to social auth

piousbox:
Fixes put_issue_url_here
### Changes

---

UUID:

### DIFF
--- a/website/src/controllers/api-v3/auth.js
+++ b/website/src/controllers/api-v3/auth.js
@@ -70,7 +70,7 @@ api.registerLocal = {
   url: '/user/auth/local/register',
   async handler (req, res) {
     let fbUser = res.locals.user; // If adding local auth to social user
-    // TODO check user doesn't have local auth
+
     req.checkBody({
       email: {
         notEmpty: {errorMessage: res.t('missingEmail')},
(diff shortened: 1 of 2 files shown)


website/src/controllers/api-v3/auth.js
```diff
@@ -124,7 +123,8 @@ api.registerLocal = {
 
     if (fbUser) {
       if (!fbUser.auth.facebook.id) throw new NotAuthorized(res.t('onlySocialAttachLocal'));
-      fbUser.auth.local = newUser;
+      fbUser.auth.local = newUser.auth.local;
+      fbUser.preferences = newUser.preferences;
```
paglias: why this?

piousbox: `preferences` are defined on line 119 of this same file. (let newUser = ... ). I cannot do fbUser.auth.local = newUser because newuser is mostly the auth.local part, not the whole user. So to conform to the schema (of user model), I put what is newUser here, into the fbUser.

paglias: I see, in this case I think you can just do `fbUser.auth.local = newUser.auth.local` and skip assigning the preferences, that object only contains the language preference and it's not important for returning users also we would risk overwriting any information in fbUser.preferences

PR #6949, opened 2016-03-23
[v3] porting user/anonymized with test

piousbox:
Fixes put_issue_url_here
### Changes

---

UUID:

### DIFF
--- a/website/src/controllers/api-v3/user.js
+++ b/website/src/controllers/api-v3/user.js
@@ -41,6 +41,69 @@ api.getUser = {
   },
 };
 
+function _cleanChecklist (task) {
+  let checklistIndex = 0;
+  _.forEach(task.checklist, (c) => {
+    c.text = `item ${checklistIndex++}`;
+  });
+}
+
+/**
+ * @api {get} /user/anonymized
+ * @apiVersion 3.0.0
+ * @apiName UserGetAnonymized
+ * @apiGroup User
+ * @apiSuccess {Object} object The object { user, tasks }
+ **/
+api.getUserAnonymized = {
+  method: 'GET',
+  middlewares: [authWithHeaders(), cron],
+  url: '/user/anonymized',
+  async handler (req, res) {
+    let user = res.locals.user.toJSON();
+    user.stats.toNextLevel = common.tnl(user.stats.lvl);
+    user.stats.maxHealth = common.maxHealth;
+    user.stats.maxMP = res.locals.user._statsComputed.maxMP;
+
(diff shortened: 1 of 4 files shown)


website/src/controllers/api-v3/user.js
```diff
@@ -41,6 +41,65 @@ api.getUser = {
...
+    delete user.webhooks;
+    delete user.achievements.challenges;
+    _.forEach(user.inbox.messages, (msg) => {
+      msg.text = res.t('inboxMessageText');
+    });
+    _.forEach(user.tags, (tag) => {
+      tag.name = res.t('tag');
+      tag.challenge = res.t('challenge');
+    });
+    // HEREHERE
```
paglias: this should be removed

website/src/controllers/api-v3/user.js
```diff
@@ -41,6 +41,65 @@ api.getUser = {
...
+    let query = { userId: user._id, $or: [{ type: 'todo', completed: false },
+                                           { type: { $in: ['habit', 'daily', 'reward'] } }],
```
paglias: can you fix the indentation here?

website/src/controllers/api-v3/user.js
```diff
@@ -41,6 +41,65 @@ api.getUser = {
...
+                };
+    let tasks = await Tasks.Task.find(query).exec();
+    function cleanChecklist (task) {
```
paglias: I'd move this function outside of the router handler, it doesn't make sense to recreate it every time

common/locales/en/api-v3.json
```diff
@@ -29,6 +29,8 @@
   "taskNotFound": "Task not found.",
   "invalidTaskType": "Task type must be one of \"habit\", \"daily\", \"todo\", \"reward\".",
   "cantDeleteChallengeTasks": "A task belonging to a challenge can't be deleted.",
+  "taskText": "task text",
```
paglias: this route will be used by devs so it's not necessary to translate the texts

website/src/controllers/api-v3/user.js
```diff
@@ -41,6 +41,69 @@ api.getUser = {
   },
 };
 
+function _cleanChecklist (task) {
+  let checklistIndex = 0;
+  _.forEach(task.checklist, (c) => {
```
paglias, line 46: when you use _.forEach the index is passed as the second argument, so you could do 

```
_.forEach(task.checklist, (c, i) => {
     c.text = `item i`;
});
```

PR #6955, opened 2016-03-24
Wheelchair

### LINKED ISSUE #6515: Add Wheelchair to Avatar Customization Page
**UDPATE**: _See the comments below. This won't be done as a back accessory._

Very exciting! We now have the art assets for adding a wheelchair to the avatar customization page. (Thank you, Balduranne!)

This should be a free back accessory available to all avatars, in its own section (since we will probably have other colors eventually).

icon-wheelchair
wheelchair

crookedneighbor:
Fixes #6515
Closes https://github.com/HabitRPG/habitrpg/pull/6816
### Changes

This is an extension of @Sinble's work on https://github.com/HabitRPG/habitrpg/pull/6816

Takes those commits, makes one adjustment to how the wheelchair sprite is shown in the avatar customization view, and brings it up to date with `develop` so it can be merged when we're ready to deploy it.


common/locales/en/character.json
```diff
@@ -34,6 +34,7 @@
     "beard": "Beard",
     "mustache": "Mustache",
     "flower": "Flower",
+    "wheelChair": "Wheel Chair",
```
Alys: I think this should be "Wheelchair" rather than "Wheel Chair". The former is the common spelling.

crookedneighbor: :+1:

PR #6960, opened 2016-03-25
Api v3 Get Content

paglias:


website/src/controllers/api-v3/content.js
```diff
@@ -10,6 +22,43 @@ function walkContent (obj, lang) {
   });
 }
 
+// After the getContent route is called the first time for a certain language
+// the response is saved on disk and subsequentially served directly from there to reduce computation.
+// Example: if `cachedContentResponses.en` is true it means that the response is cached
+let cachedContentResponses = {};
+
+// Language key set to true while the cache file is being written
+let cacheBeingWritten = {};
```
crookedneighbor, line 31: Good call here.

website/src/controllers/api-v3/content.js
```diff
@@ -24,12 +73,33 @@ api.getContent = {
   method: 'GET',
   url: '/content',
   async handler (req, res) {
-    let language = req.query.language ? req.query.language.toString() : 'en';
-    let content = _.cloneDeep(common.content);
+    let language = 'en';
+    let proposedLang = req.query.language && req.query.language.toString();
+
+    if (proposedLang && _.isBoolean(cachedContentResponses[proposedLang])) {
```
crookedneighbor: You can also do:

``` js
if (proposedLang && proposedLang in cachedContentResponses) {
```

Which I think is more readable.

In fact, I think you _could_ just do:

``` js
if (proposedLang in cachedContentResponses) {
```

But I'd want to double check that.

crookedneighbor:
Looks reasonable to me.

website/src/controllers/api-v3/content.js
paglias: `in` works, I'll use it

PR #6962, opened 2016-03-26
shared-code-user-score-task

piousbox:


test/common/ops/scoreTask.test.js
```diff
@@ -0,0 +1,185 @@
+/* eslint-disable no-shadow */
+import scoreTask from '../../../common/script/ops/scoreTask';
+import {
+  generateUser,
```
paglias, line 3: these are unit tests, you should not import helpers from the integration tests. These will save the objects to the server which is not necessary.

You can use the helpers here https://github.com/piousbox/habitrpg/blob/shared-code-user-score-task/test/helpers/api-unit.helper.js#L16 which are not async and just create the necessary objects

test/common/ops/scoreTask.test.js
```diff
@@ -0,0 +1,185 @@
...
+describe('shared.ops.scoreTask', () => {
+  let ref;
+
+  beforeEach(async () => {
+    ref = await beforeAfter();
+  });
+
+  it('throws error not enough gold', async () => {
+    let reward = await generateReward({ userId: ref.after._id, text: 'some reward', value: 100 });
+    expect(scoreTask.bind(_, { user: ref.after, task: reward })).to.throw('NotAuthorized: Not Enough Gold');
```
paglias: you should use i18n.t like here https://github.com/HabitRPG/habitrpg/blob/api-v3/test/common/ops/allocate.js#L6.

Also why you're binding to `_`? When testing for an error you want to do like here https://github.com/HabitRPG/habitrpg/blob/api-v3/test/common/ops/allocate.js#L19

expect().to.throw is a bit confusing (you can read more here http://chaijs.com/api/bdd/#method_throw), you should pass a function (without executing it) and then expect.js will run it and see if it throws, so we used try/catch instead

tasks/gulp-tests.js
```diff
@@ -358,6 +358,16 @@ gulp.task('test:api-v3:integration', (done) => {
   pipe(runner);
 });
 
+gulp.task('test:api-v3:ops', (done) => {
```
crookedneighbor: Do we need this extra task?

test/common/ops/scoreTask.test.js
crookedneighbor: s/expect.js/chai expect

Yeah, expect throw should be constructed something like this:

``` js
expect(() => {
  scoreTask({ user: ref.after, task: reward });
}).to.throw(new NotAuthorized(i18n.t('stringName')));
```

test/common/ops/scoreTask.test.js
```diff
@@ -0,0 +1,185 @@
...
+};
+
+describe('shared.ops.scoreTask', () => {
+  let ref;
+
+  beforeEach(async () => {
+    ref = await beforeAfter();
+  });
+
+  it('throws error not enough gold', async () => {
```
crookedneighbor: This should be a sentence and should fully explain the scenario it is testing.

`it throws an error when scoring a reward if user does not have enough gold`

If there are a couple of reward related assertions, you can wrap it all in a `context('Rewards')` and then the it would be `it('thows error if the user does not have enough gold')`

test/common/ops/scoreTask.test.js
```diff
@@ -0,0 +1,185 @@
...
+    let reward = await generateReward({ userId: ref.after._id, text: 'some reward', value: 100 });
+    expect(scoreTask.bind(_, { user: ref.after, task: reward })).to.throw('NotAuthorized: Not Enough Gold');
+  });
+
+  it('streak', async () => {
```
crookedneighbor: This is not an informative test description. I have no idea, based on this it block, what the test is supposed to do.

test/common/ops/scoreTask.test.js
paglias: @crookedneighbor I don't think this will work. Chai expect you to pass a class not an instance, that's why I prefer to use try/catch

crookedneighbor: It works with both, if I recall.

piousbox: I get `AssertionError: expected [Function] to throw 'NotAuthorized: Not Enough Gold' but 'NotAuthorized: Not Enough Gold' was thrown`, which is the reason why I wrote it like that - because that way the assertion worked. The _ is a placeholder scope because which scope to use does not matter for this test.

test/helpers/api-unit.helper.js
```diff
@@ -64,3 +65,32 @@ export function generateHistory (days) {
 
   return history;
 }
+
+export async function generateHabit (update = {}) {
```
paglias: Since these are unit tests you don't need to save the objects to the database and subsequentially to use async functions. Just remove all the await task.save() and convert the function to normal sync functions

PR #6964, opened 2016-03-26
chore: upgrade express to version 4

crookedneighbor:
This does the minimal work needed to upgrade to express 4. 

There are some deprecation warnings being logged, but everything _should_ work just fine. 

Will do some work to clear out the deprecated methods before merging in.


crookedneighbor:
Found bug. Cannot export csv history.

Alys:
I just received the "entity too large" error when trying to start a quest in a party of four. I don't recall getting that error before in that situation and I start quests a lot.

I reloaded and it happened again. I cleared local storage and logged back in, and it happened again.

I went to the `upstream/develop` version of the site and the error did not happen.

crookedneighbor:
@Alys Should be fixed now. Care to doublecheck?

Alys:
It's good now. Worked first time with the new code.

crookedneighbor:
As far as I know, this is ready to be merged in. Will let us do some further QA testing before merging in though.

crookedneighbor:
Merge in on the command line

PR #6978, opened 2016-03-28
Ported Hourglass Purchase, add unit tests. Created new hourglass purc…

TheHollidayInn:
https://github.com/HabitRPG/habitrpg/issues/6887 User hourglass purchase
### Changes

-Ported the hourglass purchase function
-Added unit tests
-Created purchase hourglass route
-Added integration tests

---

### DIFF
--- a/common/locales/en/api-v3.json
+++ b/common/locales/en/api-v3.json
@@ -128,5 +128,6 @@
   "privateMessageGiftSubscriptionMessage": "<%= numberOfMonths %> months of subscription! ",
   "cannotSendGemsToYourself": "Cannot send gems to yourself. Try a subscription instead.",
   "notEnoughGemsToSend": "Amount must be within 0 and your current number of gems.",
-  "mustPurchaseToSet": "Must purchase <%= val %> to set it on <%= key %>."
+  "mustPurchaseToSet": "Must purchase <%= val %> to set it on <%= key %>.",
+  "notAllowedHourglass": "Pet/Mount not available for purchase with Mystic Hourglass."
 }
--- a/common/script/ops/hourglassPurchase.js
+++ b/common/script/ops/hourglassPurchase.js
@@ -3,53 +3,58 @@ import i18n from '../i18n';
 import _ from 'lodash';
 import splitWhitespace from '../libs/splitWhitespace';
 import pickDeep from '../libs/pickDeep';
+import {
+  BadRequest,
+  NotAuthorized,
+} from '../libs/errors';
 
-module.exports = function(user, req, cb, analytics) {
-  var analyticsData, key, ref, type;
-  ref = req.params, type = ref.type, key = ref.key;
+module.exports = function purchaseHourglass (user, req = {}, analytics) {
+  let key = _.get(req, 'params.key');
+  if (!key) throw new BadRequest(i18n.t('missingKeyParam', req.language));
+
+  let type = _.get(req, 'params.type');
+  if (!type) throw new BadRequest(i18n.t('missingTypeParam', req.language));
+
   if (!content.timeTravelStable[type]) {
-    return typeof cb === "function" ? cb({
-      code: 403,
-      message: i18n.t('typeNotAllowedHourglass', req.language) + JSON.stringify(_.keys(content.timeTravelStable))
-    }) : void 0;
+    throw new NotAuthorized(i18n.t('typeNotAllowedHourglass', {allowedTypes: _.keys(content.timeTravelStable).toString()}, req.language));
   }
+
   if (!_.contains(_.keys(content.timeTravelStable[type]), key)) {
-    return typeof cb === "function" ? cb({
-      code: 403,
-      message: i18n.t(type + 'NotAllowedHourglass', req.language)
-    }) : void 0;
+    throw new NotAuthorized(i18n.t('notAllowedHourglass', req.language));
   }
+
   if (user.items[type][key]) {
-    return typeof cb === "function" ? cb({
-      code: 403,
-      message: i18n.t(type + 'AlreadyOwned', req.language)
-    }) : void 0;
+    throw new NotAuthorized(i18n.t(`${type}AlreadyOwned`, req.language));
   }
-  if (!(user.purchased.plan.consecutive.trinkets > 0)) {
-    return typeof cb === "function" ? cb({
-      code: 403,
-      message: i18n.t('notEnoughHourglasses', req.language)
-    }) : void 0;
+
+  if (user.purchased.plan.consecutive.trinkets <= 0) {
+    throw new NotAuthorized(i18n.t('notEnoughHourglasses', req.language));
   }
+
   user.purchased.plan.consecutive.trinkets--;
+
   if (type === 'pets') {
     user.items.pets[key] = 5;
   }
+
   if (type === 'mounts') {
     user.items.mounts[key] = true;
   }
-  analyticsData = {
-    uuid: user._id,
-    itemKey: key,
-    itemType: type,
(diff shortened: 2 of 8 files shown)


common/script/ops/hourglassPurchase.js
```diff
@@ -3,53 +3,59 @@ import i18n from '../i18n';
...
-module.exports = function(user, req, cb, analytics) {
-  var analyticsData, key, ref, type;
-  ref = req.params, type = ref.type, key = ref.key;
   if (!content.timeTravelStable[type]) {
-    return typeof cb === "function" ? cb({
-      code: 403,
-      message: i18n.t('typeNotAllowedHourglass', req.language) + JSON.stringify(_.keys(content.timeTravelStable))
-    }) : void 0;
+    throw new NotAuthorized(i18n.t('typeNotAllowedHourglass', req.language));
+    //  JSON.stringify(_.keys(content.timeTravelStable))
```
crookedneighbor: Is this supposed to be here?

test/common/ops/hourglassPurchase.js
```diff
@@ -0,0 +1,138 @@
...
+  BadRequest,
+  NotAuthorized,
+} from '../../../common/script/libs/errors';
+import i18n from '../../../common/script/i18n';
+
+describe('user.ops.hourglassPurchase', () => {
+  let user;
+
+  beforeEach(() => {
+    user = {
```
crookedneighbor: I believe there's a helper for creating a new user.

common/locales/en/api-v3.json
```diff
@@ -127,5 +127,6 @@
   "privateMessageGiftGemsMessage": "<%= gemAmount %> gems! ",
   "privateMessageGiftSubscriptionMessage": "<%= numberOfMonths %> months of subscription! ",
   "cannotSendGemsToYourself": "Cannot send gems to yourself. Try a subscription instead.",
-  "notEnoughGemsToSend": "Amount must be within 0 and your current number of gems."
+  "notEnoughGemsToSend": "Amount must be within 0 and your current number of gems.",
+  "NotAllowedHourglass": "Pet/Mount not available for purchase with Mystic Hourglass."
```
paglias: I know I'm being picky but the first letter should be lowercase :)

common/script/ops/hourglassPurchase.js
```diff
@@ -3,53 +3,59 @@ import i18n from '../i18n';
...
-    }) : void 0;
+    throw new NotAuthorized(i18n.t('NotAllowedHourglass', req.language));
   }
+
   if (user.items[type][key]) {
-    return typeof cb === "function" ? cb({
-      code: 403,
-      message: i18n.t(type + 'AlreadyOwned', req.language)
-    }) : void 0;
+    throw new NotAuthorized(i18n.t('AlreadyOwned', req.language));
```
paglias: is this correct? the old i18n string use seems to change based on the type

common/script/ops/hourglassPurchase.js
```diff
@@ -3,53 +3,59 @@ import i18n from '../i18n';
...
   }
+
   if (!(user.purchased.plan.consecutive.trinkets > 0)) {
```
paglias: this should be user.purchased.plan.consecutive.trinkets <= 0, it's more readable IMHO

test/common/ops/hourglassPurchase.js
paglias: yeah, you can import generateUser from unit tests helpers and pass any change to the original object there

common/script/ops/hourglassPurchase.js
crookedneighbor: Agreed.

common/script/ops/hourglassPurchase.js
crookedneighbor: I think you're right, this should be: 

``` js
`${type}AlreadyOwned`
```

TheHollidayInn:
Updated!

common/script/ops/hourglassPurchase.js
```diff
@@ -3,53 +3,58 @@ import i18n from '../i18n';
...
 
-module.exports = function(user, req, cb, analytics) {
-  var analyticsData, key, ref, type;
-  ref = req.params, type = ref.type, key = ref.key;
   if (!content.timeTravelStable[type]) {
-    return typeof cb === "function" ? cb({
-      code: 403,
-      message: i18n.t('typeNotAllowedHourglass', req.language) + JSON.stringify(_.keys(content.timeTravelStable))
-    }) : void 0;
+    throw new NotAuthorized(i18n.t('typeNotAllowedHourglass', {allowedTypes: JSON.stringify(_.keys(content.timeTravelStable))}, req.language));
```
paglias: why is JSON.stringify needed? toString doesn't work?

TheHollidayInn:
Added to string. Thanks :)

crookedneighbor:
I'm a little concerned about not having the error conditions in the integration tests. I know we technically  have them covered in unit tests, but it's strange not to include those in the integration tests. I'd feel better if we had a happy path (like you do) and at least one sad path.

TheHollidayInn:
Yea, I understand. I think @paglias said to just focus on the unit tests, though. I don't mind adding any extra tests, of course.

test/common/ops/hourglassPurchase.js
```diff
@@ -0,0 +1,136 @@
...
+describe('user.ops.hourglassPurchase', () => {
+  let user;
+
+  beforeEach(() => {
+    user = generateUser();
+  });
+
+  context('failure conditions', () => {
+    it('return error when key is not provided', () => {
+      try {
```
crookedneighbor, line 21: If we have tests with a try catch, we _need_ to use a done callback, because if this function fails to thrown an error, the assertions will never get run. Having the done callback assures the tests will fail because it won't run. 

I still think we should be able to do this with expect to.throw. I'll need to investigate why it wasn't working for @paglias

paglias, line 21: yeah I was wondering how to avoid skipping tests if an error wasn't thrown, using `done` is a good idea! Regarding `expect.throw` it works but IMO it's very confusing

common/script/ops/hourglassPurchase.js
```diff
@@ -3,53 +3,58 @@ import i18n from '../i18n';
 import _ from 'lodash';
 import splitWhitespace from '../libs/splitWhitespace';
 import pickDeep from '../libs/pickDeep';
+import {
+  BadRequest,
+  NotAuthorized,
+} from '../libs/errors';
+
+module.exports = function purchaseHourglass (user, req = {}, analytics) {
+  let key = _.get(req, 'body.key');
```
paglias: same thing as purchase here, it should be req.parameters https://github.com/HabitRPG/habitrpg/blob/develop/website/src/routes/api-v2/swagger.js#L170

TheHollidayInn:
This should be good now

website/src/controllers/api-v3/user.js
```diff
@@ -676,4 +676,27 @@ api.disableClasses = {
...
+*
+* @apiSuccess {Object} data `items purchased.plan.consecutive`
+*/
+api.userPurchaseHourglass = {
+  method: 'POST',
+  middlewares: [authWithHeaders(), cron],
+  url: '/user/purchase-hourglass/:type/:key',
+  async handler (req, res) {
+    let user = res.locals.user;
+    let purchaseHourglassResponse = common.ops.purchaseHourglass(user, req);
```
paglias: you should pass res.analytics here as the third parameter

Issue #6887, opened 2016-03-15
[API v3] Shared code port guide and roadmap


paglias:
Shared code has been split into multiple files. It has to be adapted to v3 and ported to ES6.

For now we're only going to port _ops_, starting from _user_ ones and ignoring those which handle tasks that'll be done later.

_Libs_ and _fns_ should be modified only when they're used by the op that's being ported, the remaining ones will be reviewed at the end.

The changes should be the following:
- Port code to ES6
- Explicitly declare routes in the API controllers (we don't import all of them at once automatically)
- Only really shared code should stay in /common, anything else (markModified, ...) should be moved where appropriate (server or client)
- Ops/fns shouldn't use callbacks but return values. _NOTE:_ responses should be where possible the same ones as API v2, new properties can be added to existing objects but nothing should be removed. If something needs to be removed post a comment before doing so
- Errors should be thrown (using the classes that can be found into /common/script/libs/errors)
- Where access to the request object (to get parameters or return localized messafeg) is necessary it can be passed as the last object
- The user should be passed like any other parameter, it won't be automatically available anymore
- Fns should behave like ops (throwing errors, returning values, accepting `req` and `user` parameters...)
- Both unit (for shared code) and integration (for api endpoints) tests should be written trying not to duplicate too many tests (the preference is for unit tests)
- Where possible avoid to make breaking changes the API (query parameters, ...)
  An example can be found here https://github.com/HabitRPG/habitrpg/pull/6886
### Ops

Can be found in `/common/script/ops`.
#### User
- [x] `allocate.js` -
- [x] `allocateNow.js` -
- [x] `buy.js` -
- [x] `buyMysterySet.js` -
- [x] `buyQuest.js` -
- [x] `buySpecialSpell.js` -
- [x] `changeClass.js` -
- [x] `disableClasses.js` -
- [x] `equip.js` -
- [x] `feed.js` -
- [x] `hatch.js` -
- [x] `hourglassPurchase.js` - assigned to @TheHollidayInn 
- [x] `openMysteryItem.js` -
- [x] `purchase.js` -
- [x] `readCard.js` -
- [x] `rebirth.js` - Changes tasks
- [x] `releaseBoth.js` -
- [x] `releaseMounts.js` -
- [x] `releasePets.js` -
- [x] `reroll.js` - Changes tasks
- [ ] `reset.js` - Changes tasks
- [x] `revive.js` -
- [x] `sell.js` -
- [x] `sleep.js` -
[...]

PR #6979, opened 2016-03-29
Ported open mystery item, added unit tests, added route, and integration tests

TheHollidayInn:
#6887  User open mystery item
### Changes

-Ported the open mystery item function
-Added unit tests
-Created open mystery item route
-Added integration tests

---

### DIFF
--- a/common/locales/en/api-v3.json
+++ b/common/locales/en/api-v3.json
@@ -127,6 +127,8 @@
   "privateMessageGiftGemsMessage": "<%= gemAmount %> gems! ",
   "privateMessageGiftSubscriptionMessage": "<%= numberOfMonths %> months of subscription! ",
   "cannotSendGemsToYourself": "Cannot send gems to yourself. Try a subscription instead.",
-  "notEnoughGemsToSend": "Amount must be within 0 and your current number of gems.",
-  "mustPurchaseToSet": "Must purchase <%= val %> to set it on <%= key %>."
+  "notEnoughGemsToSend": "Amount must be within 0 and your current number of gems.",,
+  "mustPurchaseToSet": "Must purchase <%= val %> to set it on <%= key %>.",
+  "mysteryItemIsEmpty": "Mystery items are empty",
+  "mysteryItemOpened": "Mystery item opened."
 }
--- a/common/script/ops/openMysteryItem.js
+++ b/common/script/ops/openMysteryItem.js
@@ -1,32 +1,40 @@
 import content from '../content/index';
+import i18n from '../i18n';
+import {
+  BadRequest,
+} from '../libs/errors';
+import _ from 'lodash';
 
-module.exports = function(user, req, cb, analytics) {
-  var analyticsData, item, ref, ref1;
-  item = (ref = user.purchased.plan) != null ? (ref1 = ref.mysteryItems) != null ? ref1.shift() : void 0 : void 0;
+module.exports = function openMysteryItem (user, req = {}, analytics) {
+  let item = user.purchased.plan.mysteryItems.shift();
+
   if (!item) {
-    return typeof cb === "function" ? cb({
-      code: 400,
-      message: "Empty"
-    }) : void 0;
+    throw new BadRequest(i18n.t('mysteryItemIsEmpty', req.language));
   }
-  item = content.gear.flat[item];
-  user.items.gear.owned[item.key] = true;
-  if (typeof user.markModified === "function") {
-    user.markModified('purchased.plan.mysteryItems');
-  }
+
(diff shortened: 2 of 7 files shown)


common/script/ops/openMysteryItem.js
```diff
@@ -1,32 +1,41 @@
...
-    }) : void 0;
+    throw new BadRequest(i18n.t('mysteryItemIsEmpty', req.language));
   }
+
   item = content.gear.flat[item];
+  item.notificationType = 'Mystery';
   user.items.gear.owned[item.key] = true;
-  if (typeof user.markModified === "function") {
+
+  if (typeof user.markModified === 'function') {
```
paglias: you can call markModified here, it'll be an empty function on the client

common/script/ops/openMysteryItem.js
```diff
@@ -1,32 +1,41 @@
...
-  item = (ref = user.purchased.plan) != null ? (ref1 = ref.mysteryItems) != null ? ref1.shift() : void 0 : void 0;
   if (!item) {
-    return typeof cb === "function" ? cb({
-      code: 400,
-      message: "Empty"
-    }) : void 0;
+    throw new BadRequest(i18n.t('mysteryItemIsEmpty', req.language));
   }
+
   item = content.gear.flat[item];
```
paglias: this object should be cloned or notificationType will be added to it

common/script/ops/openMysteryItem.js
TheHollidayInn: So, what is wrong with this line then? It check for the markModified and will call it if the function exists.

TheHollidayInn:
Fixed!

common/script/ops/openMysteryItem.js
crookedneighbor: The original line was there to only run the markmodified code on the server. To cut down on unnecessary if statements, we're going to add a noop markmodified function to the client user object. So you don't need the if statement.

TheHollidayInn: Oh, okay, cool!

common/script/ops/openMysteryItem.js
```diff
@@ -1,32 +1,41 @@
...
   if (!item) {
-    return typeof cb === "function" ? cb({
-      code: 400,
-      message: "Empty"
-    }) : void 0;
+    throw new BadRequest(i18n.t('mysteryItemIsEmpty', req.language));
   }
-  item = content.gear.flat[item];
+
+  item = Object.assign(content.gear.flat[item]);
```
paglias: as far as I know Object.assign works with 2 arguments not one (see https://developer.mozilla.org/it/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) , also I'd use lodash's cloneDeep here in case `item` has some nested objects (or if they get added later)

common/locales/en/api-v3.json
```diff
@@ -127,6 +127,8 @@
   "privateMessageGiftGemsMessage": "<%= gemAmount %> gems! ",
   "privateMessageGiftSubscriptionMessage": "<%= numberOfMonths %> months of subscription! ",
   "cannotSendGemsToYourself": "Cannot send gems to yourself. Try a subscription instead.",
-  "notEnoughGemsToSend": "Amount must be within 0 and your current number of gems.",
-  "mustPurchaseToSet": "Must purchase <%= val %> to set it on <%= key %>."
+  "notEnoughGemsToSend": "Amount must be within 0 and your current number of gems.",,
```
piousbox, line 130: extra second comma?

TheHollidayInn, line 130: ...haha. Thanks!

test/common/ops/openMysteryItem.js
```diff
@@ -0,0 +1,38 @@
...
+} from '../../helpers/common.helper';
+import {
+  BadRequest,
+} from '../../../common/script/libs/errors';
+import i18n from '../../../common/script/i18n';
+
+describe('shared.ops.openMysteryItem', () => {
+  let user;
+
+  beforeEach(() => {
```
piousbox, line 13: async / await is missing? `
beforeEach(async () => {
  user = await generateUser();
});`

paglias, line 13: no it's not missing, in this case generateUser is from the unit tests helpers so it's a plain function that return an object

paglias:
Rebased to api-v3, let's see if it helps. I've opened a new PR at https://github.com/HabitRPG/habitrpg/pull/6995

TheHollidayInn:
@paglias @piousbox found that I had an extra comma. I fixed and forced pushed.

PR #6981, opened 2016-03-29
cron timezone fix (stage 1): cron records current timezone in preferences.timezoneOffsetAtLastCron

Alys:
First stage in a fix for https://github.com/HabitRPG/habitrpg/issues/3806
### Changes
- Adds a new preference to the user model: `preferences.timezoneOffsetAtLastCron`
- Adjusts cron so that when the daily reset happens, `timezoneOffsetAtLastCron` is set to the user's current timezone.
- Adds `timezoneOffsetAtLastCron` to test/common/algos.mocha.js but only enough to stop the current tests failing (proper test coming soon).

This is being pushed now so that `timezoneOffsetAtLastCron` can be populated for active players before cron code that uses it is introduced. Any players that don't pick up the new setting in advance won't have anything bad happen to them; they just won't benefit from a new cron feature.

### DIFF
--- a/common/script/fns/cron.js
+++ b/common/script/fns/cron.js
@@ -36,6 +36,9 @@ module.exports = function(user, options) {
   }
   user.auth.timestamps.loggedin = new Date();
   user.lastCron = now;
+  if (_.isFinite(+user.preferences.timezoneOffset)) {
+    user.preferences.timezoneOffsetAtLastCron = user.preferences.timezoneOffset;
+  }
   if (user.items.lastDrop.count > 0) {
     user.items.lastDrop.count = 0;
   }
--- a/test/common/algos.mocha.js
+++ b/test/common/algos.mocha.js
@@ -53,6 +53,7 @@ let beforeAfter = (options = {}) => {
     before.preferences.dayStart = after.preferences.dayStart = options.dayStart;
   }
   before.preferences.timezoneOffset = after.preferences.timezoneOffset = options.timezoneOffset || moment().zone();
+  before.preferences.timezoneOffsetAtLastCron = after.preferences.timezoneOffsetAtLastCron = before.preferences.timezoneOffset;
   if (options.limitOne) {
     before[`${options.limitOne}s`] = [before[`${options.limitOne}s`][0]];
     after[`${options.limitOne}s`] = [after[`${options.limitOne}s`][0]];
--- a/website/src/models/user.js
+++ b/website/src/models/user.js
@@ -350,6 +350,7 @@ var UserSchema = new Schema({
     skin: {type:String, 'default':'915533'},
     shirt: {type: String, 'default': 'blue'},
     timezoneOffset: Number,
+    timezoneOffsetAtLastCron: Number,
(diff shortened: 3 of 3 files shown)


common/script/fns/cron.js
```diff
@@ -36,6 +36,9 @@ module.exports = function(user, options) {
   }
   user.auth.timestamps.loggedin = new Date();
   user.lastCron = now;
+  if (_.isFinite(+user.preferences.timezoneOffset)) {
```
crookedneighbor, line 39: I don't understand this check. Can you elaborate?

Alys, line 39: It returns true it it's a number (positive, negative, or zero). Returning true on zero is important so that the GMT timezone isn't ignored.  https://lodash.com/docs#isFinite

Alys, line 39: Oh and it doesn't set user.preferences.timezoneOffsetAtLastCron to anything at all if user.preferences.timezoneOffset doesn't exist because the only time those two values should be different is if the user's timezone has changed.

crookedneighbor, line 39: But why `_.isFinite`? Wouldn't `_.isNumber` make more sense? Checking against the number not being infinity is a little confusing to me. 

Also, is there any case in which the value is set, but it is _not_ a number? I'd guess that the two value types are undefined or Number, right? In which case, I think `if (timezoneOffset in user.preferences)` would do just as well.

Alys, line 39: _.isNumber returns true on NaN, which enrages me. :)  _.isFinite is the easiest check I found to ensure that the value is an actual "number" in any real sense of the term.

> Also, is there any case in which the value is set, but it is not a number? 

At the moment, no, not that I'm aware of, but code changes in future might let that happen. _.isFinite protects against that. Paranoia I suppose.

I was basing it on this code and comment (and I'll be modifying that code similarly to "check and clean timezoneOffset"):

```
function sanitizeOptions (o) {
  let ref = Number(o.dayStart || 0);
  let dayStart = !_.isNaN(ref) && ref >= 0 && ref <= 24 ? ref : 0;
  let timezoneOffset = o.timezoneOffset ? Number(o.timezoneOffset) : Number(moment().zone());
  // TODO: check and clean timezoneOffset as for dayStart (e.g., might not be a number)
```

Alys, line 39: But sorry, I should have said, I'll change it to `if (timezoneOffset in user.preferences)`. I don't feel strongly about my current method and I'm happy to use yours.

paglias, line 39: `if (timezoneOffset in user.preferences)` won't work correctly. Because this isn't a normal object but a mongoose one, that property will always be present even if undefined

paglias, line 39: also the + is not necessary, the value is always casted to a number. In general you can just check for the value existing and not being NaN

Alys, line 39: Then I'll leave it as _.isFinite because that checks for both existence and not being NaN. I'll get rid of the +.

PR #6989, opened 2016-03-31
Fix: typo in the response for Hatching Potions

a-ayyash:
Fixes put_issue_url_here
### Changes

---

UUID:

### DIFF
--- a/common/script/fns/randomDrop.js
+++ b/common/script/fns/randomDrop.js
@@ -69,7 +69,7 @@ module.exports = function(user, modifiers, req) {
         base2[name2] = 0;
       }
       user.items.hatchingPotions[drop.key]++;
-      drop.type = 'HatchingPotion';
+      drop.type = 'Hatching Potion';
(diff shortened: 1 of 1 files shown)


common/script/fns/randomDrop.js
```diff
@@ -69,7 +69,7 @@ module.exports = function(user, modifiers, req) {
         base2[name2] = 0;
       }
       user.items.hatchingPotions[drop.key]++;
-      drop.type = 'HatchingPotion';
```
crookedneighbor, line 72: I'm not sure this is actually a typo.

Alys, line 72: Yes, it's not a typo. @a-ayyash It's a key that is used in other parts of the code to control behaviour depending on what type of thing is dropped. It's deliberately written as a single word because that's common in programming for keys.

Generally in Habitica (with some exceptions that are getting rarer), all user-visible text is defined in files under the `locales` directory. Any text elsewhere is likely to be program code and usually shouldn't be modified when it seems like a typo.

PR #6990, opened 2016-03-31
Ported read card, added unit tests, added read card route and integration tests

TheHollidayInn:
#6887 User read card
### Changes

-Ported read card
-Added unit tests
-Added read card route
-Add integration tests

---

### DIFF
--- a/common/script/ops/readCard.js
+++ b/common/script/ops/readCard.js
@@ -1,10 +1,28 @@
-module.exports = function(user, req, cb) {
-  var cardType;
-  cardType = req.params.cardType;
-  user.items.special[cardType + "Received"].shift();
-  if (typeof user.markModified === "function") {
-    user.markModified("items.special." + cardType + "Received");
+import splitWhitespace from '../libs/splitWhitespace';
+import _ from 'lodash';
+import i18n from '../i18n';
+import {
+  BadRequest,
+  NotAuthorized,
+} from '../libs/errors';
+import content from '../content/index';
+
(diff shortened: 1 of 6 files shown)


common/script/ops/readCard.js
```diff
@@ -1,10 +1,23 @@
...
+
+module.exports = function readCard (user, req = {}) {
+  let cardType = _.get(req.params, 'cardType');
+
+  if (!cardType) {
+    throw new BadRequest(i18n.t('cardTypeRequired', req.language));
   }
+
+  user.items.special[`${cardType}Received`].shift();
+  user.markModified(`items.special.${cardType}Received`);
```
paglias: markModified in this case was not necessary because mongoose will proxy the shift operation on the array, it was here before but it can be removed

test/api/v3/integration/user/POST-user_read_card.test.js
```diff
@@ -0,0 +1,38 @@
...
+
+  beforeEach(async () => {
+    user = await generateUser();
+  });
+
+  it('returns an error when cardType not provided', async () => {
+    await expect(user.post('/user/read-card'))
+      .to.eventually.be.rejected.and.to.eql({
+        code: 404,
+        error: 'NotFound',
```
paglias: hmm I don't see any 404 error in the readCard code

paglias: I think it returns a 404 because the user/read-card route without a parameter doesn't exist

PR #7007, opened 2016-04-02
Inserting split into boss attacks to separate boss attacks and user attacks on two different lines.

### LINKED ISSUE #3453: Party questing: damage dealt text doesn't wrap
I'm sure this is not a priority thing but I wanted to drop it in here just in case it's a quick fix or something?

When my party is questing, damage dealt by party and boss is written in red in the chat. Stuff like "harpereaves attacks Vice, the Shadow Wyrm for 13.3 damage, Vice, the Shadow Wyrm attacks party for 0.0 damage."

My first complaint is that the comma in the middle should be a semi-colon. "... Vice, the Shadow Wyrm for 13.3 damage; Vice, the Shadow Wyrm attacks ..."

See! Very minor. But also, for some reason that red text doesn't wrap, and when the boss' name is long you have to scroll to the side to find out how much damage the boss dealt. This bugs me and other people in my party. Also if you're scrolling up and down sometimes it gets in the way for me because I'm on a Mac and use two-finger trackpad scrolling; sometimes it tries to scroll sideways when the cursor is bang in the middle of the chat stuff while I'm trying to scroll up and down. I don't mind but it's a thing that happens.

I'm on OS X 10.9.2, with the latest version of Chrome.

screen shot 2014-05-12 at 11 17 48

So there we are. Some totally unimportant stuff. :) Thanks for being awesome!
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2116794-party-questing-damage-dealt-text-doesn-t-wrap?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Hus274:
fixes #3453 
### Changes

Splits the markdown text notification to the party into two code tags split by a newline instead of one code tag. The code tag prevents wrap-text css styles from being applied which was the cause of the scrolling.

The newline can be removed so that it only splits into two lines if needed by the browser size which might be slightly more ideal as Markdown doesn't recommend counting on line break escapes. This would be easy to change by simply updating the '  \n' to a comma or semicolon but the newline seem to be the preferred solution in the issue.

---

Related issue: https://github.com/HabitRPG/habitrpg/issues/3453

### DIFF
--- a/website/src/models/group.js
+++ b/website/src/models/group.js
@@ -355,7 +355,7 @@ GroupSchema.statics.bossQuest = function(user, progress, cb) {
     var down = progress.down * quest.boss.str; // multiply by boss strength
 
     group.quest.progress.hp -= progress.up;
-    group.sendChat("`" + user.profile.name + " attacks " + quest.boss.name('en') + " for " + (progress.up.toFixed(1)) + " damage, " + quest.boss.name('en') + " attacks party for " + Math.abs(down).toFixed(1) + " damage.`"); //TODO Create a party preferred language option so emits like this can be localized
+    group.sendChat("`" + user.profile.name + " attacks " + quest.boss.name('en') + " for " + (progress.up.toFixed(1)) + " damage.` `" + quest.boss.name('en') + " attacks party for " + Math.abs(down).toFixed(1) + " damage.`"); //TODO Create a party preferred language option so emits like this can be localized
 
     // If boss has Rage, increment Rage as well
     if (quest.boss.rage) {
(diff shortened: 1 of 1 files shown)


crookedneighbor:
Seems reasonable to me.

Alys:
I like the idea of not using the newline, but a comma or semicolon would look out of place. Is a space enough to make it split when the browser window is small?

Hus274:
Yep, as long as they're in two different code blocks, it should automatically wrap the text to a new line.

Hus274:
Changed to a space instead of a newline addressing Alys's comment above and updated the remote branch to avoid annoying merge commit necessity. Doing this resulting in this pull request closing. New pull request with updated code can be found here: https://github.com/HabitRPG/habitrpg/pull/7013

PR #7009, opened 2016-04-03
shared-code-random-drop

piousbox:


common/script/fns/predictableRandomNamespaced.js
```diff
@@ -0,0 +1,5 @@
+import predictableRandom from './predictableRandom';
```
crookedneighbor: I don't like this pattern. We should either export predictableRandom like this in the first place or attach all the functions to an object and import them that way whenever they are used.

piousbox: That's right. I vote for exporting an object and having functions inside it.

common/script/fns/randomDrop.js
```diff
@@ -3,80 +3,129 @@ import content from '../content/index';
...
+    quest = content.quests[user.party.quest.key];
+  if (quest && quest.collect && predictableRandom(user, user.stats.gp) < chance) {
+    dropK = randomVal(user, quest.collect, {
+      key: true,
     });
+    if (!user.party.quest.progress.collect[dropK])
+      user.party.quest.progress.collect[dropK] = 0;
     user.party.quest.progress.collect[dropK]++;
-    if (typeof user.markModified === "function") {
+    if (typeof user.markModified === 'function') {
```
paglias: this check is not necessary

common/script/fns/randomDrop.js
```diff
@@ -3,80 +3,129 @@ import content from '../content/index';
...
-  }), 0) || 0));
+
+  chance = _.min([Math.abs(task.value - 21.27), 37.5]) / 150 + 0.02;
+  chance *= task.priority *
+    (1 + (task.streak / 100 || 0)) *
+    (1 + user._statsComputed.per / 100) *
+    (1 + (user.contributor.level / 40 || 0)) *
+    (1 + (user.achievements.rebirths / 20 || 0)) *
+    (1 + (user.achievements.streak / 200 || 0)) *
+    (user._tmp.crit || 1) * (1 + 0.5 * (_.reduce(task.checklist, (m, i) => {
```
paglias, line 35: here we should make so it works if task.checklist doesn't exists, maybe passing an empty array?

common/script/fns/randomDrop.js
```diff
@@ -3,80 +3,129 @@ import content from '../content/index';
...
+      return m + (i.completed ? 1 : 0);
+    }, 0) || 0));
   chance = diminishingReturns(chance, 0.75);
-  quest = content.quests[(ref = user.party.quest) != null ? ref.key : void 0];
-  if ((quest != null ? quest.collect : void 0) && user.fns.predictableRandom(user.stats.gp) < chance) {
-    dropK = user.fns.randomVal(quest.collect, {
-      key: true
+
+  /* chance *=
```
paglias: these comments should be added next to the code above

common/script/fns/predictableRandomNamespaced.js
paglias: We can't export an object of functions because there are fns that require other fns, libs that require other libs, ... so they would cause circular dependencies

paglias:
I'm going to merge this without the stubbing

piousbox:
Sounds good.

PR #7013, opened 2016-04-03
Splitting party and boss damage notifications into two code blocks.

Hus274:
Fixes https://github.com/HabitRPG/habitrpg/issues/3453
### Changes

```
Splitting party and boss damage notifications into two code blocks.

Splitting the party and boss damage notifications into two code blocks will allow the page to push them to newlines if needed instead of added a scroll bar and making the user scroll to see the entire message.

Related issue: https://github.com/HabitRPG/habitrpg/issues/3453
```

---


crookedneighbor:
Works well!

Issue #6462, opened 2016-01-05
Tavern / Guild / Party chat lag / slowness


MHDante:
Platform: Win10, Chrome 47.0.2526.106 m

Take a look at this example of the tavern's chat:
![image](https://cloud.githubusercontent.com/assets/4652122/12110172/5c02a85c-b357-11e5-9f20-3e753b775e8a.png)
The Ss were typed one at a time and the Js were typed 5 at a time.

It seems that there is a listener happening every time that the content of the textbox changes. While this doesn't seem to be a problem in-of-itself, It appears that the digest function it is calling is quite expensive.(I'm not sure if this is an AngularJS problem, or something regarding your backend). For reference, my account has ~20 tasks.

It seems that the lag is present throughout the site, (probably on some task iteration logic), and other Issue threads have tried to address it. #6215 #3486 However, I'd recommend fixing the issue at the chat level.  The entire autocomplete feature could be disabled, or perhaps changing it so that it waits a few seconds from last input before it calls the digest handler. ( I think it already works this way? but the delay is really short?)

I know @lefnire mentioned addressing the chat system in another performance thread, but I'm not sure if it came through. Some good places to start would probably be:
habitrpg/website/public/js/controllers/autoCompleteCtrl.js

I would try to fix it myself, but I'm unfamiliar with the software stack. (As of now)

testsubject1987:
Thank you.

colwem:
Is any work on this being done?  Is there a reason it can't or shouldn't be done?  If not I'd like to work on this issue.

nrbunn:
Ok looking at a different angle into this as debounce seems to help but after a while it seems to generate the same problem. Plus the issue with the crtl-enter not working because of super typers @crookedneighbor. 

i found removing the below from habitrpg\website\views\index.jade sped up the chat overall. Not sure yet why its there as there is a commented line right before it without it.
`, ng-keydown="$broadcast('habit:keydown', $event)", ng-keyup="$broadcast('habit:keyup', $event)"`

crookedneighbor:
Yep, I think that's it.

That was added to detect `ctrl` being pressed on the tasks page and escape being pressed on on the inventory page to deselect an item.

crookedneighbor:
Just deployed a release that uses some of the changes @Sinble suggested. There's still a _bit_ of lag, but it's much better than it was before and since we're doing a rewrite of the client soon, I think I'm going to say this is good enough.

robertpateii:
[...]

Issue #6482, opened 2016-01-09
incorrect attribute / stats points from various ways of changing levels (e.g. FCV) - need generic function to do allocation / deallocation


Alys:
Staff/Admins: Adjust anything here as you see fit. 

The bugs described here happen only under non-standard conditions and aren't critical to fix. I recommend that we don't work on this issue until after the API v3 changes are live.

A player's allocated plus unallocated attribute points should add up to their current level (or to 100 if their level is over 100, with exceptions listed below). Under some circumstances, this does not occur - e.g., from past bugs or from using Fix Character Values to adjust level. Once the points have got out of sync with the level, they aren't always brought back into sync during levelling up or dying.

To fix this, we could have a generic function that performs the following actions:
- adds up the allocated and unallocated attribute points
- compares them to the current level OR to 100 if the level is over 100
- if there are too few points, EITHER:
  - run the `autoAllocate` function if the user has chosen auto-allocation
  - add unallocated attribute points (`stats.points`)
- if there are too many points:
  - first subtract from the unallocated points, as much as necessary or until the unallocated points are zero
  - if still too many points AND if the user is not above level 100, subtract from allocated points, choosing the affected stat(s) in the same manner they are currently chosen after death
- never let any attribute have negative points
- never let unallocated points be negative

That function could then be run under the following circumstances:
- levelling up
- death
- rebirth
- reset
- Fix Character Values when the level is changed
- class change
- anything else?

There are special conditions for players who are above level 100, and special treatment for those who already have more than 100 points. Note the decision that is documented in "Attribute Points Change" in [Skills Rebalancing and Other Changes](http://habitica.wikia.com/wiki/User_blog:LadyAlys/Skills_Rebalancing_and_Other_Changes):
_"Players who already have attribute points above 100 will not have those points removed."_
- If they level up, they are not given a point unless their points are less than 100.
- If they rebirth or reset, their points drop to 1 as for any other player, and from that time on, they cannot ever gain more than 100 points.
- If they do a class change, their points drop to 100 and from that time on, they cannot ever gain more than 100 points.
- If they die and they have **un**allocated points, they lose all those unallocated points. If they die and have no unallocated points, they lose only **one** allocated point.
- Under any other circumstances, if they happen to have more **allocated** points than 100, the function described above does **not** reduced them to 100.

[...]

PR #7015, opened 2016-04-03
added debounce option of 1000ms to delay digest()

nrbunn:
#6462
### Changes

Added 500+ users and 500+ messages to tarvern chat to reproduce slowness when typing. Found that angular was spending a lot of time in digest. Found article related to slowness [http://www.aaron-gray.com/delaying-the-digest-cycle-in-angularjs/](url). Added 1000ms delay and saw improvement might need to play around with number some more.

---

UUID: 25479d8c-45e1-482c-877e-6af2ccaba643

### DIFF
--- a/website/views/options/social/chat-box.jade
+++ b/website/views/options/social/chat-box.jade
@@ -8,7 +8,7 @@ div.chat-form.guidelines-not-accepted(ng-if='!user.flags.communityGuidelinesAcce
 
 form.chat-form(ng-if='user.flags.communityGuidelinesAccepted' ng-submit='postChat(group,message.content)')
   div(ng-controller='AutocompleteCtrl')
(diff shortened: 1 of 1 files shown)


crookedneighbor:
Seems reasonable to me.

crookedneighbor:
While this fixes the textarea delay, it also adds a little delay between when someone can type a message and press enter to send it so that it creates a blank line _instead_ of sending it. Kind of an edge case, as someone would have to be typing fast and have very little delay between typing a message and pressing enter, but I would find that change confusing.

Redmega:
I type >100wpm when I'm not being lazy so I can see this happening quite often for me.

nrbunn:
Interesting... Thanks for testing. Didn't even think that was a use case since I type then hit the send button. Didn't know enter was also send. Not sure of a quick fix for that. Thoughts?

Alys:
There is this: https://github.com/HabitRPG/habitrpg/issues/4460 "return key causes chat messages to be submitted - should we change that?"  The consensus was that Enter to send the message is not desirable. Maybe we could change that at the same time as fixing the chat delay? Faster chat would be popular and would help to smooth over any complaints about the new behaviour for Enter.

nrbunn:
I updated my pull request with the comments from @Alys. If we change are minds no big deal. Also, when testing i had to add ctrl-enter to the meta-enter line i'm not sure if that was a issue on my side or just my limited knowledge. @crookedneighbor thoughts?

crookedneighbor:
It still has the same problem. There's a delay between when you type and when you can press "meta - enter" to send the message.

nrbunn:
Closing this pull request as I believe I found the root of the issue for #6482 and will need a different fix

Issue #6994, opened 2016-04-01
cannot "Copy message as To-Do" from the Inbox


meganstrickland:
General Info
- UUID: 73e5125c-2c87-4004-8ccd-972aeac4f17a
- Browser: Chrome (49.0.2623.110 m) and Firefox (40.0.3)
- OS: Windows 7
### Description

The "Copy message as To-Do" feature does not work from the Inbox. It works in the Tavern (and I assume guilds, etc.), but in the Inbox there is no confirmation dialog. I see the same behavior in both Chrome and Firefox.

Alys:
Confirmed. No javascript console messages. The button does literally nothing (except for the on-hover explanatory text).

Redmega:
If nobody picks this up by tonight I'd like to give it a go

Redmega:
I've got my local environment set up and running, I'm looking into this now.

Redmega:
Still performing root cause analysis but it looks like from a cursory look the copyToDo method is never being hit on click of the inbox "Copy to Todos" button. I don't know enough about angular to fathom why, though.

Alys:
@Redmega Maybe the method is in a scope that isn't being used for the code that does the inbox? 
`throws buzzwords around while trying to look intelligent and knowledgable`

nrbunn:
@Redmega I wanted to see if I could solve this problem as well. I believe i have a fix but i don't want to steal your issue. Give you a hint to **Part** of the problem. There is a missing`ng-controller='ChatCtrl'`

Redmega:
@Sinble Thanks for the tip. Been reading into angular so not having that ChatCtrl controller attached would cause it to not be able to talk with the controller layer, right? So it doesn't even know what the copyToDo function is...

Guessing there's more to it than just adding the ng-controller attribute, though, since you're saying thats only _part_ of the problem. You can go ahead and solve this, I haven't gotten any headway with it and I'm going to be busy with work the next few days.

PR #7017, opened 2016-04-04
Bug/inbox todo

nrbunn:
#6994
### Changes

missing `ng-controller='ChatCtrl'` so it was unable to find copyToDo. Next copyToDo uses group.name in scope so created inboxCtrl to supply it.

---

UUID: 25479d8c-45e1-482c-877e-6af2ccaba643


nrbunn:
hmm noticed my debounce hash is in there. let me know if that causes any issues.

crookedneighbor:
I'm hesitant to just slap the chat controller onto the inbox, as it's really just this function that is missing: https://github.com/HabitRPG/habitrpg/blob/develop/website/public/js/controllers/chatCtrl.js#L93-L109

Doesn't seem right to expose all the other functionality of the chat controller onto the inbox.

In fact, the notes field should probably be updated to be something more specific to the inbox. Why not just add a similar function to the new inbox controller? Would also eliminate the need to create a hacky group object on `$scope`?

nrbunn:
You know didn't even think of that. Got blinding by the issue of not having group in scope when I added the chat controller. Sorry still a noob. I will revise and send a update.

nrbunn:
Created new copytodo function in new inbox controller based on @crookedneighbor feedback.

Might need some better word smithing for default info?
inboxtodo

crookedneighbor:
The bracketed text before was creating a link to the party/user. This bracketted text doesn't really do anything. 

Also, I don't think the "Task from <user>" text conveys useful information about the task. I'd like to hear from @lemoness and @Alys about what they want out of copying messages to tasks.

Alys:
If I were using this feature, I'd probably rewrite the task title myself each time, or move the entire text from the note into the title. I haven't seen any comments to indicate how others would use it.

Maybe "Message from username" as the default title, with no brackets. I don't think we can predict what the majority's preferred text would be so anything simple would do. I'd prefer "Message..." rather than "Task..." because it's a more accurate description of where the material came from, and sometimes you might be copying the message to a To-Do just to think about it, rather than to perform a task.

But feel free to ignore this if there's conflicting opinions. I don't feel strongly about it.

lemoness:
I also like "Message From Username." @veeeeeee may have options as well, though.

nrbunn:
inboxtodo

common/locales/en/groups.json
```diff
@@ -110,6 +110,8 @@
     "copyMessageAsToDo": "Copy message as To-Do",
     "messageAddedAsToDo": "Message copied as To-Do.",
     "messageWroteIn": "<%= user %> wrote in <%= group %>",
+    "taskFromInbox": "<%= from %> wrote, '<%= message %>'",
```
crookedneighbor: No need for a comma here.

website/public/js/controllers/inboxCtrl.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+habitrpg.controller("InboxCtrl", ['$scope', '$rootScope',
+    function($scope, $rootScope) {
+      $scope.copyToDo = function(message) {
+        var taskNotes = env.t("taskFromInbox",  {
+          from: message.uuid == 'system'
```
crookedneighbor: It's confusing to have a ternary here. I think this logic should be outside of the object.

crookedneighbor: Especially given the fact that we do the exact same call below.

crookedneighbor:
This should also have a karma test before it can be merged in.

crookedneighbor:
Overall, looking good!

nrbunn:
Updated. Had a fun time with tests so please review to make sure I did it correctly.

veeeeeee:
Eek, just saw this in my email. "Message from [X]" is fine by me. Other possible variations: "Request from X" "Action item via X" or even "ToDo via X"

crookedneighbor:
Sorry it is taking me so long to review this. I'm slowly going through the PRs now.

crookedneighbor:
Great work! I've notated this toward your next tier.

PR #7018, opened 2016-04-04
Shared code webhooks

piousbox:


common/script/ops/addWebhook.js
```diff
@@ -1,15 +1,22 @@
 import refPush from '../libs/refPush';
+import validator from 'validator';
+import i18n from '../i18n';
+import {
+  NotFound,
+  BadRequest,
+} from '../libs/errors';
 
-module.exports = function(user, req, cb) {
+module.exports = function(user, req) {
```
paglias: missing space between function and (, linting will fail

paglias: also req should have a default value of `{}`

website/src/controllers/api-v3/user.js
```diff
@@ -769,7 +769,64 @@ api.userOpenMysteryItem = {
...
+ **/
+api.addWebhook = {
+  method: 'POST',
+  middlewares: [authWithHeaders()],
+  url: '/user/webhook',
+  async handler (req, res) {
+    let user = res.locals.user;
+    let result = common.ops.addWebhook(user, req);
+    await user.save();
+    res.respond(200, {id: result.id});
```
paglias: I'd return the full result here

website/src/controllers/api-v3/user.js
```diff
@@ -769,7 +769,64 @@ api.userOpenMysteryItem = {
...
+ **/
+api.updateWebhook = {
+  method: 'PUT',
+  middlewares: [authWithHeaders()],
+  url: '/user/webhook/:id',
+  async handler (req, res) {
+    let user = res.locals.user;
+    common.ops.updateWebhook(user, req);
+    await user.save();
+    res.respond(200, {});
```
paglias: return the result

common/script/ops/addWebhook.js
```diff
@@ -1,15 +1,22 @@
...
   var wh;
   wh = user.preferences.webhooks;
-  refPush(wh, {
+
+  if(!validator.isURL(req.body.url)) throw new BadRequest(i18n.t('invalidUrl', req.language));
```
paglias: to make sure the server doesn't crash if req.body is missing I'd use `_.get()` (https://lodash.com/docs#get) to get the value from `req.body`

common/script/ops/deleteWebhook.js
```diff
@@ -1,7 +1,5 @@
-module.exports = function(user, req, cb) {
+
+module.exports = function(user, req) {
   delete user.preferences.webhooks[req.params.id];
```
paglias: same thing here about using _.get

PR #7019, opened 2016-04-04
Changing level up attribute allocation division from collapsable to ng-show

### LINKED ISSUE #6961: Clicking 'show stat allocation' button rapidly will desync text and collapse state
General Info
- UUID: n/a
- Browser: Chrome 49
- OS: Ubuntu 14.04.3
### Description

The Bootstrap collapse will not toggle when partway through its animation. The button to show/hide it, however, changes its text on every click – if you double-click the show/hide button, you end up with this:

image

The collapse fires several [events](https://getbootstrap.com/javascript/#collapse-events), which would presumably help to fix this.
#### Console Errors

n/a

Hus274:
Fixes https://github.com/HabitRPG/habitrpg/issues/6961
### Changes

Changing level up attribute allocation division from collapsable to ng-show

Changing this dialog from a collapsible element to one that uses ng-show takes care of this issue. The only change in functionality from the user perspective (outside of fixing the bug) is it no longer has a slow open effect where the table slides down to be displayed. Now it just appears.

Example:
Old Collapse Style - http://i.imgur.com/iKqar66.gif
New ng-show style - http://i.stack.imgur.com/z94GJ.gif

---

### DIFF
--- a/website/views/shared/modals/level-up.jade
+++ b/website/views/shared/modals/level-up.jade
@@ -17,11 +17,11 @@ script(type='text/ng-template', id='modals/levelUp.html')
       p=env.t('fullyHealed')
       br
       div(ng-if='user.flags.classSelected && !user.preferences.disableClasses && !user.preferences.automaticAllocation')
-        a.btn.btn-default(ng-click='statsAllocationBoxIsOpen = !statsAllocationBoxIsOpen',
-          data-toggle='collapse', data-target='#stat-allocation', aria-expanded=false, aria-controls='stat-allocation')
+        a.btn.btn-default(ng-click='statsAllocationBoxIsOpen = !statsAllocationBoxIsOpen', 
+          aria-expanded=false, aria-controls='stat-allocation')
           | {{statsAllocationBoxIsOpen  ? env.t('hideQuickAllocation') : env.t('showQuickAllocation')}}
         p &nbsp;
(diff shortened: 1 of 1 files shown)


crookedneighbor:
This should be used instead: http://angular-ui.github.io/bootstrap/#/collapse

Hus274:
Ahh nice, I was looking for something like that but couldn't find it. Will update and repost.

Hus274:
I looked into using the angular bootstrap collapse that you attached but it's not currently functional. The link you posted is for Angular bootstrap 1.4 and it looks like Habitica is using 1.3. It looks like there's a bug (feature?) around the collapse function in 1.3 that prevents the slide effect from working (http://stackoverflow.com/questions/30178573/collapse-transition-not-working-with-angulars-ui-bootstrap) which results in the same behavior as ng-show which is what I posted initially.

It looks like it could work on this version but it would require including the ng-animate module (https://github.com/angular-ui/bootstrap/issues/3676)

What do you think is the best way to proceed?

crookedneighbor:
Hmmm... That's tough. Will have to think about that.

Hus274:
Did you find anything you think could be useful here?

crookedneighbor:
I think we're just going to pass on this for now. There's a redesign happening soon, so all this UI code will be moot anyway.

PR #7030, opened 2016-04-06
[API v3] Payments refactor

SabreCat:
### Changes
- _Temporary:_ Copies files from `/website/src/controllers/payments/` to `/website/src/controllers/api-v3/payments/`
- Updates code style of payments modules to current standards

---

UUID: mememe


website/src/controllers/api-v3/payments/amazon.js
```diff
@@ -0,0 +1,277 @@
...
+let amzPayment = amazonPayments.connect({
+  environment: amazonPayments.Environment[IS_PROD ? 'Production' : 'Sandbox'],
+  sellerId: nconf.get('AMAZON_PAYMENTS:SELLER_ID'),
+  mwsAccessKey: nconf.get('AMAZON_PAYMENTS:MWS_KEY'),
+  mwsSecretKey: nconf.get('AMAZON_PAYMENTS:MWS_SECRET'),
+  clientId: nconf.get('AMAZON_PAYMENTS:CLIENT_ID'),
+});
+
+api.verifyAccessToken = function verifyAccessToken (req, res) {
+  if (!req.body || !req.body.access_token) {
```
paglias: here the check for `req.body` can be removed, now it defaults to an empty object

test/api/v3/unit/libs/amazonPayments.test.js
```diff
@@ -0,0 +1,103 @@
...
+
+  describe('#createOrderReferenceId', () => {
+    it('verifies billingAgreementId', async (done) => {
+      try {
+        let inputSet = {};
+        delete inputSet.Id;
+        await amzLib.createOrderReferenceId(inputSet);
+      } catch (e) {
+
+        /* console.log('error!', e);
```
paglias: this should be removed

website/src/controllers/api-v3/auth.js
```diff
@@ -2,6 +2,8 @@ import validator from 'validator';
 import moment from 'moment';
 import passport from 'passport';
 import nconf from 'nconf';
+import setupNconf from '../../libs/api-v3/setupNconf';
+setupNconf();
```
paglias: these calls shouldn't be necessary as long as the whole server is loaded. If you're requiring the controllers separately then setupNconf should be called from the tests

website/src/controllers/top-level/payments/amazon.js
```diff
@@ -0,0 +1,280 @@
...
+import { authWithHeaders } from '../../../middlewares/api-v3/auth';
+var payments = require('./index');
+
+let api = {};
+
+/**
+ * @api {post} /api/v3/payments/amazon/verifyAccessToken verify access token
+ * @apiVersion 3.0.0
+ * @apiName AmazonVerifyAccessToken
+ * @apiGroup Payments
```
paglias, line 24: blank line before before apiParam(s)

website/src/controllers/top-level/payments/amazon.js
```diff
@@ -0,0 +1,280 @@
...
+ * @apiParam {string} access_token the access token
+ * @apiSuccess {} empty
```
paglias: blank line before apiSuccess

website/src/controllers/top-level/payments/amazon.js
```diff
@@ -0,0 +1,280 @@
...
+  },
+};
+
+/**
+ * @api {post} /api/v3/payments/amazon/createOrderReferenceId create order reference id
+ * @apiVersion 3.0.0
+ * @apiName AmazonCreateOrderReferenceId
+ * @apiGroup Payments
+ * @apiParam {string} billingAgreementId billing agreement id
+ * @apiSuccess {object} object containing { orderReferenceId }
```
paglias: when using res.respond the response object will always be in the {data, success, message} formate so instead of using `object` you should use `data` as the apiSuccess name

website/src/controllers/top-level/payments/amazon.js
```diff
@@ -0,0 +1,280 @@
...
+ **/
+api.createOrderReferenceId = {
+  method: 'POST',
+  url: '/payments/amazon/createOrderReferenceId',
+  // middlewares: [authWithHeaders()],
```
paglias: this should not be disabled

website/src/controllers/top-level/payments/amazon.js
```diff
@@ -0,0 +1,280 @@
...
+ * @apiGroup Payments
+ * @apiParam {string} access_token the access token
+ * @apiSuccess {} empty
+ **/
+api.verifyAccessToken = {
+  method: 'POST',
+  url: '/payments/amazon/verifyAccessToken',
+  async handler (req, res) {
+    await amzLib.getTokenInfo(req.body.access_token)
+    .then(() => {
```
paglias: why are you mixing await with `then` and `catch`? `await` expects a promise and will block the code until it's resolved. If you need to catch the error you can wrap it into a try/catch block

website/src/controllers/top-level/payments/amazon.js
```diff
@@ -0,0 +1,280 @@
...
+  url: '/payments/amazon/createOrderReferenceId',
+  // middlewares: [authWithHeaders()],
+  async handler (req, res) {
+
+    try {
+      let response = await amzLib.createOrderReferenceId({
+        Id: req.body.billingAgreementId,
+        IdType: 'BillingAgreement',
+        ConfirmNow: false,
+        AWSAccessKeyId: 'something',
```
paglias: why this?

website/src/controllers/top-level/payments/amazon.js
paglias: in this case for example it should be `@apiSuccess {object} data.orderReferenceId The order reference id`. The format for @apiSuccess is `{data type} name description`

website/src/controllers/top-level/payments/amazon.js
```diff
@@ -0,0 +1,280 @@
...
+
+    if (gift) {
+      if (gift.type === 'gems') {
+        amount = gift.gems.amount / 4;
+      } else if (gift.type === 'subscription') {
+        amount = shared.content.subscriptionBlocks[gift.subscription.key].price;
+      }
+    }
+
+    /* if (!req.body || !req.body.orderReferenceId) {
```
paglias: why is this commented out? Also it should use `req.check{Body|Params|Query}` as in the other controllers

paglias: other params should be validated too where necessary

website/src/controllers/top-level/payments/amazon.js
```diff
@@ -0,0 +1,280 @@
...
+        },
+        SellerAuthorizationNote: 'HabitRPG Payment',
+        TransactionTimeout: 0,
+        CaptureNow: true,
+      });
+
+      await amzLib.closeOrderReference({ AmazonOrderReferenceId: orderReferenceId });
+
+      // execute payment
+      let giftUser = await User.findById(gift ? gift.uuid : undefined);
```
paglias: if `gift` is falsy you can skip the user.findById part

paglias: I mean, just put it inside the `if (gift)`block

website/src/controllers/top-level/payments/index.js
```diff
@@ -0,0 +1,238 @@
...
+import nconf from 'nconf';
+import pushNotify from '../../../libs/api-v3/pushNotifications';
+import shared from '../../../../../common' ;
+
+import amazon from './amazon';
+import iap from './iap';
+import paypal from './paypal';
+import stripe from './stripe';
+
+const IS_PROD = nconf.get('NODE_ENV') === 'production';
```
paglias: you can use nconf.get('IS_PROD')

website/src/controllers/top-level/payments/index.js
```diff
@@ -0,0 +1,238 @@
...
+  // Block sub perks
+  let perks = Math.floor(months / 3);
+  if (perks) {
+    plan.consecutive.offset += months;
+    plan.consecutive.gemCapExtra += perks * 5;
+    if (plan.consecutive.gemCapExtra > 25) plan.consecutive.gemCapExtra = 25;
+    plan.consecutive.trinkets += perks;
+  }
+  revealMysteryItems(recipient);
+  if (IS_PROD) {
```
paglias: blank line before and after if block

website/src/libs/api-v3/amazonPayments.js
```diff
@@ -0,0 +1,93 @@
...
+  return amazonPayments.connect({
+    environment: amazonPayments.Environment[IS_PROD ? 'Production' : 'Sandbox'],
+    sellerId: nconf.get('AMAZON_PAYMENTS:SELLER_ID'),
+    mwsAccessKey: nconf.get('AMAZON_PAYMENTS:MWS_KEY'),
+    mwsSecretKey: nconf.get('AMAZON_PAYMENTS:MWS_SECRET'),
+    clientId: nconf.get('AMAZON_PAYMENTS:CLIENT_ID'),
+  });
+}
+
+api.getTokenInfo = (token) => {
```
paglias: in this file, using `Promise` directly seems a bit of an overkill to me. Can't we use Q's method for this? This way there's simply more code

website/src/libs/api-v3/amazonPayments.js
```diff
@@ -0,0 +1,93 @@
+import amazonPayments from 'amazon-payments';
+import nconf from 'nconf';
+import common from '../../../../common';
+let t = common.i18n.t;
+const IS_PROD = nconf.get('NODE_ENV') === 'production';
+
+let api = {};
+
+function connect (amazonPayments) { // eslint-disable-line no-shadow
```
paglias: this function shouldn't be called every time. Just when the server starts and then store the return value somewhere

website/src/controllers/top-level/payments/amazon.js
piousbox: This looks to be required now by amazon? I get the specific error 'awsaccesskeyid is required'

website/src/controllers/top-level/payments/amazon.js
piousbox: I believe the `amazon-payments` library will actually cover these validations and throw errors if any of those parameters are missing.

website/src/libs/api-v3/amazonPayments.js
```diff
@@ -0,0 +1,93 @@
...
+    });
+  });
+};
+
+api.createOrderReferenceId = (inputSet) => {
+  let amzPayment = connect(amazonPayments);
+  return new Promise((resolve, reject) => {
+    amzPayment.offAmazonPayments.createOrderReferenceForId(inputSet, (err, response) => {
+      if (err) return reject(err);
+      if (!response.OrderReferenceDetails || !response.OrderReferenceDetails.AmazonOrderReferenceId) {
```
paglias: this line gets ignored since all errors are catched by the line before

website/src/controllers/top-level/payments/amazon.js
paglias: that's true... we should still validate our own parameters like `gift` in this case

piousbox: `gift` looks entirely optional here.

website/src/libs/api-v3/amazonPayments.js
piousbox: I'll rewrite in Q. I think Promise has cleaner semantics and is easier to understand. I believe if we run benchmarks Q will be equal performance to Promise. But, I'll rewrite in Q if that's the preference throughout the codebase.

website/src/controllers/top-level/payments/amazon.js
paglias: you're right, sorry

website/src/libs/api-v3/amazonPayments.js
paglias: hmm but in this case writing promises manually is just longer. What I'm talking about is using one lines like https://github.com/kriskowal/q#adapting-node (specifically `Q.nbind`)

PR #7042, opened 2016-04-08
Ported reroll. Added unit tests. Added reroll route. Added integration tests

TheHollidayInn:
#6887 User ops reroll
### Changes

Ported reroll. 
Added unit tests. 
Add reroll route. 
Added integration tests

---

UUID:  206039c6-24e4-4b9f-8a31-61cbb9aa3f66


common/script/ops/releaseBoth.js
```diff
@@ -4,6 +4,7 @@ import {
   NotAuthorized,
 } from '../libs/errors';
 import splitWhitespace from '../libs/splitWhitespace';
+import _ from 'lodash';
```
paglias, line 7: what... how didn't we notice this?

website/src/controllers/api-v3/user.js
```diff
@@ -974,4 +974,33 @@ api.userRebirth = {
...
+  async handler (req, res) {
+    let user = res.locals.user;
+    let query = {
+      userId: user._id,
+      type: {$in: ['daily', 'habit', 'todo']},
+    };
+    let tasks = await Tasks.Task.find(query).exec();
+    let rerollResponse = common.ops.reroll(user, tasks, req, res.analytics);
+
+    await user.save();
```
paglias: in parallel with the tasks below?

TheHollidayInn: Ah, yea. Can do.

common/script/ops/reroll.js
```diff
@@ -1,29 +1,36 @@
...
+      uuid: user._id,
+      acquireMethod: 'Gems',
+      gemCost: 4,
+      category: 'behavior',
+    });
   }
-  return typeof cb === "function" ? cb(null, user) : void 0;
+
+  let response = {
+    data: user,
```
paglias: this probably doesn't make too much sense now... we could return both the user or the tasks

TheHollidayInn: This would help us update the user without reloading or sending another ajax request on the front-end.

paglias: yeah, let's return everything, it's a lot of data but in this case it's useful

PR #7043, opened 2016-04-08
Ported reset. Added unit tests. Added reset route. Added integration tests

TheHollidayInn:
#6887 User ops reset
### Changes

Ported reset. 
Added unit tests. 
Add reroll reset. 
Added integration tests

---


common/script/ops/reset.js
```diff
@@ -1,35 +1,22 @@
...
-  if (typeof user.markModified === "function") {
-    user.markModified('items.gear.owned');
-  }
-  user.preferences.costume = false;
-  return typeof cb === "function" ? cb(null, user) : void 0;
+
+  resetGear(user);
+
+  let response = {
+    data: user,
```
paglias: here too let's return an array of [user, tasks]

TheHollidayInn: Tasks aren't modified here.

paglias: you're right!

TheHollidayInn: Also, tasks are delete here.. so yea :P

tasks/gulp-eslint.js
```diff
@@ -14,7 +14,19 @@ const COMMON_FILES = [
   '!./common/script/ops/blockUser.js',
   '!./common/script/ops/clearPMs.js',
   '!./common/script/ops/deletePM.js',
-  '!./common/script/ops/reset.js',
+  '!./common/script/ops/deleteTag.js',
```
paglias: these should be removed

TheHollidayInn:
Sorry. I messed up the lint during rebase. It should be good now.

common/script/ops/reset.js
```diff
@@ -1,35 +1,22 @@
...
-    return true;
-  });
-  gear.owned.weapon_warrior_0 = true;
-  if (typeof user.markModified === "function") {
-    user.markModified('items.gear.owned');
-  }
-  user.preferences.costume = false;
-  return typeof cb === "function" ? cb(null, user) : void 0;
+
+  resetGear(user);
```
paglias, line 20: can you move resetGear.js to /fns? it's confusing in /ops

common/script/ops/reset.js
```diff
@@ -1,35 +1,22 @@
...
+import i18n from '../i18n';
 
-module.exports = function(user, req, cb) {
-  var gear;
-  user.habits = [];
-  user.dailys = [];
-  user.todos = [];
-  user.rewards = [];
+module.exports = function reset (user) {
+  user.tasksOrder.habits = [];
```
paglias: these 4 lines don't make sense as we remove the necessary ids from tasksOrder using `removeNonChallengeTasks` and anyway not all ids should be removed. 

I'd just keep `removeNonChallengeTasks` and add a comment here to do the same on the client side

paglias: Otherwise you can pass tasks to `reset` and remove the ids from `tasksOrder` here in the shared code. Then you return also the removed ids and use them to executed the server query

TheHollidayInn:
Okay, I decided to pass the task ids that need to be removed to the function so that it will remove those tasks. I think it is a good solution. Let me know your thoughts.

paglias:
@TheHollidayInn Hmm the problem now is that you run the same code (to remove tasks ids from tasksOrder) twice: first on the server and then in common code

TheHollidayInn:
Yea, I see what you mean. Well, doesn't calling removeNonChallengeTasks first solve the issue? Doing that gives the response a user with the TaskOrder tasks removed so the client can update without a second request.

TheHollidayInn:
Updated.

PR #7047, opened 2016-04-09
[v3] Merge in develop

crookedneighbor:
Getting a test failure related to only showing the first 50 patrons in the hall. Not sure what that is about. 

Opening this PR so we can sort it out together.


crookedneighbor:
Of course, now it passes :/

May be a flaky test. We'll have to investigate that some more.

PR #7049, opened 2016-04-09
feat: Add eb extension to increase deploy timeout to 30 minutes

crookedneighbor:
This _should_ prevent the EB CLI from timing out. 

Pinging @SabreCat


Alys:
When I was deploying several days ago, I found that 30 minutes wasn't enough. I don't have a better number though because I didn't need to deploy again after using 30.

However this might be specific to Australia's atrocious internet access. :rage:

PR #7053, opened 2016-04-09
fix: Change from moment.zone to moment.utcOffset to fix warnings

crookedneighbor:
### Problem

Moment deprecated the use of `moment().zone` in favor of `moment().utcOffset`. So we're getting deprecation warnings on the server and client. HOWEVER, it is not just a rename. See this note from the `utcOffset` docs:

> **NOTE:** Unlike `moment.fn.zone` this function returns the real offset from UTC, not the reverse offset (as returned by `Date.prototype.getTimezoneOffset`).

So we may need to test and make some adjustments before merging this in. 

@Alys @SabreCat @paglias

### DIFF
--- a/common/script/cron.js
+++ b/common/script/cron.js
@@ -27,7 +27,7 @@ function sanitizeOptions (o) {
   let dayStart = !_.isNaN(ref) && ref >= 0 && ref <= 24 ? ref : 0;
 
   let timezoneOffset;
-  let timezoneOffsetDefault = Number(moment().zone());
+  let timezoneOffsetDefault = Number(moment().utcOffset());
   if (_.isFinite(o.timezoneOffsetOverride)) {
     timezoneOffset = Number(o.timezoneOffsetOverride);
   } else if (_.isFinite(o.timezoneOffset)) {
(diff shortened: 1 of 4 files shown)


paglias:
I think there's another deprecation notice that sometimes appears regarding moment: https://github.com/moment/moment/issues/1407

Alys:
We should definitely change from `zone()` to `utcOffset()`, but I think it might need to be introduced more slowly with a cutover period where cron can run using either function (with each associated zone value stored in separate `preferences` attributes). If functions return different values, then we'll get bugs if cron is using only `utcOffset()` but if the user's stored `preferences.timezoneOffset` value was set using `zone()`.

This won't be a problem when the new cron code runs immediately after the user reloads or logs in to Habitica on a web browser (because the new code in common/script/public/userServices.js will pick up the correct utcOffset value), but if the new cron code runs before `preferences.timezoneOffset` has been adjusted, I think we'll get bugs - this could happen if cron runs on a browser that hasn't been reloaded recently, or runs when a third-party tool connects to the server, or runs from an older version of the mobile apps.

In fact I think the apps will need to be adjusted to match, since they send the timezone to the server, and their current code presumably uses the same values that `zone()` gives.

I'm wondering if it's worth delaying this until after we've introduced some cron-related changes that are being considered, because they might give us more control over when cron is run.

Alys:
Cron can convert from the old zone format to the new, and should probably store the new in a different field and delete the old, and then keep on doing the same conversion every time it sees that the old field has been repopulated (e.g., from a non-upgraded app). It should also check if the converted old format is different that the previously stored new format (in the same way it currently checks for the user having changed zones since the last cron).

crookedneighbor:
Going to redo this later. Not urgent until the next version of moment comes out and we want to upgrade.

PR #7056, opened 2016-04-09
shared-code-tags-gold-silver

piousbox:


test/common/libs/taskDefaults.test.js
```diff
@@ -0,0 +1,15 @@
...
+  it('applies defaults', () => {
+    let task = taskDefaults();
+    expect(task.type).to.eql('habit');
+    expect(task._id).to.exist;
+    expect(task.text).to.eql(task._id);
+    expect(task.tags).to.eql([]);
+    expect(task.value).to.eql(0);
+    expect(task.priority).to.eql(1);
+    expect(task.up).to.eql(true);
+    expect(task.down).to.eql(true);
```
paglias, line 13: can you add tests also for when type is daily, reward and todo?

common/script/libs/appliedTags.js
```diff
@@ -6,15 +6,11 @@ Are there tags applied?
...
-  arr = [];
-  _.each(userTags, function(t) {
-    if (t == null) {
-      return;
-    }
-    if (taskTags != null ? taskTags[t.id] : void 0) {
-      return arr.push(t.name);
+module.exports = function appliedTags (userTags, taskTags = {}) {
+  let arr = [];
+  _.each(userTags, (t) => {
```
crookedneighbor: Would this be equivalent:

``` js
let arr = userTags
  .filter(tag => { return taskTags[tag.id]; })
  .map(tag => { return tag.name; });
```

common/script/libs/silver.js
```diff
@@ -4,10 +4,10 @@ Silver amount from their money
 
 // TODO move to client
 
-module.exports = function(num) {
+module.exports = function silver (num) {
   if (num) {
-    return ("0" + Math.floor((num - Math.floor(num)) * 100)).slice(-2);
+    return `0${Math.floor((num - Math.floor(num)) * 100)}`.slice(-2);
```
crookedneighbor: The calculations should be pulled out so this is easier to read.

PR #7058, opened 2016-04-10
Fixed reward sound playing after gold subtracted (fixes #6963)

### LINKED ISSUE #6963: Unreliable sound when clicking Armoire
General Info
- UUID: 2d6ef231-50b4-4a22-90e7-45eb97147a2c
- Browser: Firefox 45.0.1
- OS: Windows 7 Service Pack 1 (via Bootcamp on 2011 Macbook Pro, if that's relevant)
### Description

Sound (in LuneFox's theme at least) does not play reliably when clicking Armoire or Health Potion when it used to do so. Not sure if this is a bug or some sort of intended change (I'm leaning toward unintended because the sound is unreliable), so thought I'd report it to be sure if it was unintended that you guys caught wind of it. (There's also no sound when casting spells, but I don't cast spells enough to remember if that's normal.)

If I had to give a percentage to how much the sound is playing, I'd have to say it's generally not playing a sound (maybe 90-95% of the clicks?) but occasionally is doing so. I only tested the health potion once and got no sound for that. I don't generally buy health potions.
#### Console Errors

No JS errors to speak of.

Redmega:
Fixes #6963 
### Problem

Reward sound would not play if the gold left was less than the cost of the item, even if the gold before the purchase was enough to purchase it (and the item was actually bought).
### Changes

Made reward sound play before purchase (basing reward sound on current gold, instead of gold left after purchase)

---

UUID: e2d15ef1-fb48-429d-ba54-646da01f4da7


crookedneighbor:
Seems sensible to me.

Redmega:
One issue I can see is if the `User.user.ops.buy({params:{key:item.key}});` operation fails for some reason other than not enough gold, then the reward sound will have played but the item will not have been purchased. Then again we are not considering that in the previous code, so I'm not sure where to proceed with regards to that.

crookedneighbor:
I think that is an edge case that we shouldn't worry too much about.

website/public/js/controllers/tasksCtrl.js
```diff
@@ -220,8 +220,8 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
     $scope.armoire = Content.armoire;
 
     $scope.buy = function(item) {
-      User.user.ops.buy({params:{key:item.key}});
-      playRewardSound(item);
+        playRewardSound(item);
+        User.user.ops.buy({params:{key:item.key}});
```
Alys: Redmega, I'm sorry but could you reduce the indent here by two spaces? We're standardising on two spaces per indent level across the code base.

Redmega: Whoops! Those were inserted by my IDE. Slipped my notice. I'll fix that tonight.

Redmega:
Ugh, I made a mess of the commit. Somehow checked out a new branch by accident and made a bit of a mess. How do I fix it? =/

crookedneighbor:
I've cherry-picked the relevant commits and merged them in. 

PR #7065, opened 2016-04-11
Separated out buy functions into buyGear, buyArmoire, and buyPotion

TheHollidayInn:
#6887 User ops buy refactor
### Changes

Move Buy to:
BuyGear
BuyPotion
BuyArmoire

---

### DIFF
--- a/common/script/index.js
+++ b/common/script/index.js
@@ -91,6 +91,9 @@ import scoreTask from './ops/scoreTask';
 import sleep from './ops/sleep';
 import allocate from './ops/allocate';
 import buy from './ops/buy';
+import buyGear from './ops/buyGear';
+import buyPotion from './ops/buyPotion';
+import buyArmoire from './ops/buyArmoire';
 import buyMysterySet from './ops/buyMysterySet';
 import buyQuest from './ops/buyQuest';
 import buySpecialSpell from './ops/buySpecialSpell';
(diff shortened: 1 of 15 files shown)


paglias:
@TheHollidayInn thanks for this! there's a change we can rename the original buy to buyGear, keep the 3 new routes but also a more general `buy` route that'll work like the old one (call the specific buyGear/buyPotion/buyArmoire from the general buy route?)

TheHollidayInn:
Updated!

common/script/index.js
```diff
@@ -259,7 +263,7 @@ api.wrap = function wrapUser (user, main = true) {
       releasePets: _.partial(importedOps.releasePets, user),
       releaseMounts: _.partial(importedOps.releaseMounts, user),
       releaseBoth: _.partial(importedOps.releaseBoth, user),
-      buy: _.partial(importedOps.buy, user),
```
paglias, line 262: buyPotion and buyArmoire are missing

paglias:
Except for the comment, the only thing I feel like we should add is a dummy `common.ops.buy` op that simply call the new ops so we can keep using common.ops.buy on the client if we want to

TheHollidayInn:
So, the buy route that I have added back will call the common.ops.buy and do the "switch" logic there?

paglias:
I'd say so

TheHollidayInn:
Updated!

PR #7067, opened 2016-04-11
shared-code-statsComputed

piousbox:


common/script/fns/statsComputed.js
```diff
@@ -0,0 +1,28 @@
+import _ from 'lodash';
+import content from '../content/index';
+import splitWhitespace from '../libs/splitWhitespace';
+const $w = splitWhitespace;
+import * as statHelpers from '../statHelpers';
+
+module.exports = function statsComputed (user) {
+  let thisArr = $w('stats stats.buffs items.gear.equipped.weapon items.gear.equipped.armor items.gear.equipped.head items.gear.equipped.shield');
```
paglias: let's transform this into an array and on multiple lines

common/script/fns/statsComputed.js
```diff
@@ -0,0 +1,28 @@
...
+  let computed = _.reduce(['per', 'con', 'str', 'int'], (m, stat) => {
+    m[stat] = _.reduce(thisArr, (m2, path) => {
+      let val = user.fns.dotGet(path);
```
paglias: in place of dotGet you can use _.get(user, path)

common/script/fns/statsComputed.js
```diff
@@ -0,0 +1,28 @@
...
+      let item = content.gear.flat[val];
+      if (!item) item = {};
+      if (!item[stat])
```
paglias: parenthesis

common/script/fns/statsComputed.js
```diff
@@ -0,0 +1,28 @@
...
+      else
+        item[stat] = Number(item[stat]);
+      let thisMultiplier = item.klass === user.stats.class || item.specialClass === user.stats.class ? 1.5 : 1;
+      let thisReturn = path.indexOf('items.gear') !== -1 ? item[stat] * thisMultiplier : Number(val[stat]);
+      return m2 + thisReturn || 0;
+    }, 0);
+    m[stat] += Math.floor(statHelpers.capByLevel(user.stats.lvl) / 2);
+    return m;
+  }, {});
+  computed.maxMP = computed.int * 2 + 30;
```
paglias: a black line above this one?

common/script/fns/statsComputed.js
piousbox: But it becomes an array anyway. $w() is the array shorthand. This syntax looks very clean to me.

common/script/fns/statsComputed.js
```diff
@@ -0,0 +1,28 @@
+import _ from 'lodash';
+import content from '../content/index';
+import splitWhitespace from '../libs/splitWhitespace';
+const $w = splitWhitespace;
+import * as statHelpers from '../statHelpers';
+
+module.exports = function statsComputed (user) {
```
paglias: let's move this as a normal function inside /libs not as a fns

common/script/fns/statsComputed.js
piousbox: Explain please? Looks ok to me. you mean add curly brackets {} ?

common/script/fns/statsComputed.js
paglias: I feel like this line is too long, with an array you can split it into multiple lines at least. Right now I have to scroll to see the whole line

common/script/fns/statsComputed.js
paglias: yeah it should be `if () { } else {}`. Otherwise linting will fail and if it doesn't I still don't like it :P

common/script/fns/statsComputed.js
crookedneighbor: Agreed.

piousbox: Updated.

2016-04-11 15:45 GMT-07:00 Blade Barringer notifications@github.com:

> In common/script/fns/statsComputed.js
> https://github.com/HabitRPG/habitrpg/pull/7067#discussion_r59294744:
> 
> > @@ -0,0 +1,28 @@
> > +import _ from 'lodash';
> > +import content from '../content/index';
> > +import splitWhitespace from '../libs/splitWhitespace';
> > +const $w = splitWhitespace;
> > +import \* as statHelpers from '../statHelpers';
> > +
> > +module.exports = function statsComputed (user) {
> > -  let thisArr = $w('stats stats.buffs items.gear.equipped.weapon items.gear.equipped.armor items.gear.equipped.head items.gear.equipped.shield');
> 
> Agreed.
> 
> —
> You are receiving this because you authored the thread.
> https://github.com/HabitRPG/habitrpg/pull/7067/files/586749613f859d411458dc96cc9b07c64cfb1748#r59294744

## 

Victor Pudeyev
piousbox@gmail.com

common/script/fns/index.js
```diff
@@ -20,6 +21,7 @@ module.exports = {
   randomDrop,
   autoAllocate,
   updateStats,
+  statsComputed,
```
crookedneighbor: This is weird. We should look in the code and see where it's being used. If it doesn't need to be on the fns object, it shouldn't be.

common/script/index.js
```diff
@@ -284,23 +280,14 @@ api.wrap = function wrapUser (user, main = true) {
...
-          let item;
-          let val = user.fns.dotGet(path);
-          return m2 + (path.indexOf('items.gear') !== -1 ? (item = content.gear.flat[val], (Number(item ? item[stat] : undefined) || 0) * ((item ? item.klass : undefined) === user.stats.class || (item ? item.specialClass : undefined) === user.stats.class ? 1.5 : 1)) : Number(val[stat]) || 0);
-        }, 0);
-        m[stat] += Math.floor(api.capByLevel(user.stats.lvl) / 2);
-        return m;
-      }, {});
-      computed.maxMP = computed.int * 2 + 30;
-      return computed;
+      return user.fns.statsComputed();
```
crookedneighbor: This should call the library directly.

common/script/libs/statsComputed.js
```diff
@@ -0,0 +1,30 @@
+import _ from 'lodash';
+import content from '../content/index';
+import splitWhitespace from '../libs/splitWhitespace';
+const $w = splitWhitespace;
+import * as statHelpers from '../statHelpers';
+
+module.exports = function statsComputed (user) {
+  let thisArr = $w('stats stats.buffs items.gear.equipped.weapon items.gear.equipped.armor items.gear.equipped.head items.gear.equipped.shield');
```
crookedneighbor: The comment for this was lost because you force pushed, but this should be an array so it's easier to read.

common/script/libs/statsComputed.js
```diff
@@ -0,0 +1,30 @@
+import _ from 'lodash';
+import content from '../content/index';
+import splitWhitespace from '../libs/splitWhitespace';
+const $w = splitWhitespace;
```
crookedneighbor: Probably not necessary since we're removing the place that uses this, but you could have just done: `import $w from '../libs/splitWhitespace';`

common/script/libs/statsComputed.js
```diff
@@ -0,0 +1,30 @@
...
+import * as statHelpers from '../statHelpers';
+
+module.exports = function statsComputed (user) {
+  let thisArr = $w('stats stats.buffs items.gear.equipped.weapon items.gear.equipped.armor items.gear.equipped.head items.gear.equipped.shield');
+  let computed = _.reduce(['per', 'con', 'str', 'int'], (m, stat) => {
```
crookedneighbor, line 8: Any chance we'd just want to use the native Array `reduce` method?

common/script/libs/statsComputed.js
crookedneighbor: Let's get a better name for this variable.

common/script/libs/statsComputed.js
```diff
@@ -0,0 +1,30 @@
...
+      return m2 + thisReturn || 0;
+    }, 0);
+    m[stat] += Math.floor(statHelpers.capByLevel(user.stats.lvl) / 2);
+    return m;
+  }, {});
+
+  computed.maxMP = computed.int * 2 + 30;
+  return computed;
+};
+
```
crookedneighbor: Extra line.

test/common/fns/statsComputed.test.js
```diff
@@ -0,0 +1,28 @@
...
+} from '../../helpers/common.helper';
+
+describe('common.fns.statsComputed', () => {
+  let user;
+
+  beforeEach(() => {
+    user = generateUser();
+  });
+
+  it('is sane', () => {
```
crookedneighbor: This doesn't tell me anything about the test.

test/common/fns/statsComputed.test.js
```diff
@@ -0,0 +1,28 @@
...
+    let result = statsComputed(user);
+    let result2 = user._statsComputed;
+    let result3 = user.fns.statsComputed();
+    expect(result).to.eql(result2);
+    expect(result).to.eql(result3);
+  });
+
+  it('computes properties', () => {
```
crookedneighbor: This also doesn't tell me anything about what the function is doing.

common/script/index.js
piousbox: What do you mean? `return _.partial(importedFns.statsComputed, user);`?

common/script/fns/index.js
piousbox: I removed that I'm getting regressions like 

```
  4) shared.ops.scoreTask scores dailys up:
     AssertionError: expected NaN to be above 0
      at expectGainedPoints (scoreTask.test.js:37:37)
      at Context.<anonymous> (scoreTask.test.js:172:9)

```

So I'll take a look in depth on this in a bit.

common/script/index.js
paglias: I think he means that the statsComputed file should be moved to /libs, imported here and called like `statsComputed(user)` - when not using user.{fns|ops} you should always pass `user`

common/script/libs/statsComputed.js
piousbox, line 8: Almost everything in lodash can be written using native javascript. _.reduce() is pretty much the same as Array.prototype.reduce(). By the way, I prefer underscore library instead of lodash because the former is semantically cleaner. http://underscorejs.org/ In this particular case, using lodash or native js I think makes no difference.

common/script/index.js
```diff
@@ -33,10 +33,6 @@ api.capByLevel = statHelpers.capByLevel;
 api.tnl = statHelpers.toNextLevel;
 api.diminishingReturns = statHelpers.diminishingReturns;
 
-// TODO under api.libs?
-import splitWhitespace from './libs/splitWhitespace';
-const $w = api.$w = splitWhitespace;
```
paglias, line 38: hmm I think this is used somewhere on the client, it should not be removed

piousbox, line 38: Are you sure? I can add it back with a linter exception. If it's used on the frontend in another file, wouldn't linter catch it with "$w is undefined"?

paglias, line 38: No it wouldn't because the client files have are not being linted. To avoid the linter exception just do api.$w = splitWhitespace; without assign it to a local const

PR #7072, opened 2016-04-12
Fixes: cursor as pointer on hero-box class in level up dialog that is not clickable by the user.

Hus274:
Fixes https://github.com/HabitRPG/habitrpg/issues/7066
### Changes

Updates the style of the herobox to use the default cursor on this page instead of the pointer cursor that is defaulted by the hero-box class.

---

…


crookedneighbor:
Thanks for this! I've notated this toward your next tier.

PR #7073, opened 2016-04-13
moved broadcast to task page. changed up key press binding

nrbunn:
#6462
### Changes

Removed 
`, ng-keydown="$broadcast('habit:keydown', $event)", ng-keyup="$broadcast('habit:keyup', $event)"`
from html element and placed on task div. Had to use mouseover and mouseout to find key down and key up due to ng-keypress not being able to work on a div. The biggest change of how this would work is someone would have to have there mouse over the task area vs the whole page being able to detect the keypress event. 

That being said chat seems to be way faster :0) So need someone to review and see what they think.. if we are good i will do the same for inventory. if not i will need some other ideas :0)

---

UUID:

### DIFF
--- a/website/public/js/controllers/tasksCtrl.js
+++ b/website/public/js/controllers/tasksCtrl.js
@@ -258,4 +258,26 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
         $rootScope.playSound('Reward');
       }
     }
+
+    var ctrlKeys = [17, 224, 91];
+    $scope.hoverIn = function () {
+      $(document).keydown(function (keyEvent) {
+        if (ctrlKeys.indexOf(keyEvent.keyCode) !== -1) {
+                $scope.ctrlPressed = true;
+                $scope.$apply();
+              }
+      });
+      $(document).keyup(function (keyEvent) {
+        console.log('blah');
+        if (ctrlKeys.indexOf(keyEvent.keyCode) !== -1) {
+          $scope.ctrlPressed = false;
+          $scope.$apply();
+        }
+      });
+    }
+    $scope.hoverout = function () {
+      $(document).unbind('keydown');
+      $(document).unbind('keyup');
+      $scope.ctrlPressed = false;
+    }
   }]);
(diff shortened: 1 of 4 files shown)


crookedneighbor:
Finally got a chance to look at this, but the chat _does not_ seem noticeably faster to me. :(

nrbunn:
Just so I know we are testing apples to apples. I'm testing in tavern chat. How Many chat entries are you using. I think I made 100 test ones. Because it gets slower with the more messages on the page.

crookedneighbor:
I also tested in the tavern. Don't remember the exact number of messages, but it was substantial. It's possible the problem is on my end.

website/public/js/controllers/tasksCtrl.js
```diff
@@ -258,4 +258,26 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
...
+    var ctrlKeys = [17, 224, 91];
+    $scope.hoverIn = function () {
+      $(document).keydown(function (keyEvent) {
+        if (ctrlKeys.indexOf(keyEvent.keyCode) !== -1) {
+                $scope.ctrlPressed = true;
+                $scope.$apply();
+              }
+      });
+      $(document).keyup(function (keyEvent) {
+        console.log('blah');
```
crookedneighbor, line 271: This console log shouldn't be here.

website/public/js/controllers/tasksCtrl.js
```diff
@@ -258,4 +258,26 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
         $rootScope.playSound('Reward');
       }
     }
+
+    var ctrlKeys = [17, 224, 91];
+    $scope.hoverIn = function () {
+      $(document).keydown(function (keyEvent) {
```
crookedneighbor, line 264: Let's use addEventListener instead of jquery

nrbunn:
I will test again. It's been awhile maybe something has changed

paglias:
The tavern in production will always keep 200 hundreds messages, when a new one is added the oldest one gets deleted

crookedneighbor:
@Sinble I took your basic idea and implemented it without jQuery. 

nrbunn:
Sorry you had to make the fix. I was going to get around to it but i went on vacation. 
25479d8c-45e1-482c-877e-6af2ccaba643

crookedneighbor:
No worries. I've notated this toward your next tier.

PR #7081, opened 2016-04-15
Added initial tests for cron

TheHollidayInn:
Fixes N/A
### Changes

Add initial tests for cron
-Middleware tests to come

---

UUID:

### DIFF
--- a/test/api/v3/unit/libs/cron.test.js
+++ b/test/api/v3/unit/libs/cron.test.js
@@ -57,6 +57,12 @@ describe('cron', () => {
     expect(user.items.lastDrop.count).to.equal(0);
   });
 
+  it('increments user cron count', () => {
+    let cronCountBefore = user.flags.cronCount;
+    cron({user, tasksByType, daysMissed, analytics});
+    expect(user.flags.cronCount).to.be.greaterThan(cronCountBefore);
+  });
+
   describe('end of the month perks', () => {
     beforeEach(() => {
       user.purchased.plan.customerId = 'subscribedId';
@@ -127,6 +133,70 @@ describe('cron', () => {
     });
   });
 
+  describe('end of the month perks when user is not subscribed', () => {
+    it('does not reset plan.gemsBought on a new month', () => {
+      user.purchased.plan.gemsBought = 10;
+      cron({user, tasksByType, daysMissed, analytics});
+      expect(user.purchased.plan.gemsBought).to.equal(10);
+    });
+
+    it('does not reset plan.dateUpdated on a new month', () => {
+      cron({user, tasksByType, daysMissed, analytics});
+      expect(user.purchased.plan.dateUpdated).to.be.empty;
+    });
+
+    it('does not increment plan.consecutive.count', () => {
+      user.purchased.plan.consecutive.count = 0;
+      cron({user, tasksByType, daysMissed, analytics});
+      expect(user.purchased.plan.consecutive.count).to.equal(0);
+    });
+
+    it('does not decrement plan.consecutive.offset when offset is greater than 0', () => {
+      user.purchased.plan.consecutive.offset = 1;
+      cron({user, tasksByType, daysMissed, analytics});
+      expect(user.purchased.plan.consecutive.offset).to.equal(1);
+    });
+
+    it('does not increment plan.consecutive.trinkets when user has reached a month that is a multiple of 3', () => {
+      user.purchased.plan.consecutive.count = 5;
+      cron({user, tasksByType, daysMissed, analytics});
+      expect(user.purchased.plan.consecutive.trinkets).to.equal(0);
+    });
+
+    it('doest not increment plan.consecutive.gemCapExtra when user has reached a month that is a multiple of 3', () => {
+      user.purchased.plan.consecutive.count = 5;
+      cron({user, tasksByType, daysMissed, analytics});
+      expect(user.purchased.plan.consecutive.gemCapExtra).to.equal(0);
+    });
+
+    it('does not increment plan.consecutive.gemCapExtra when user has reached the gemCap limit', () => {
+      user.purchased.plan.consecutive.gemCapExtra = 25;
+      user.purchased.plan.consecutive.count = 5;
+      cron({user, tasksByType, daysMissed, analytics});
+      expect(user.purchased.plan.consecutive.gemCapExtra).to.equal(25);
+    });
+
+    it('does nothing to plan stats if we are before the last day of the cancelled month', () => {
+      user.purchased.plan.dateTerminated = moment(new Date()).add({days: 1});
+      cron({user, tasksByType, daysMissed, analytics});
+      expect(user.purchased.plan.customerId).to.not.exist;
+    });
+
+    xit('does nothing to plan stats when we are after the last day of the cancelled month', () => {
+      user.purchased.plan.dateTerminated = moment(new Date()).subtract({days: 1});
+      user.purchased.plan.consecutive.gemCapExtra = 20;
+      user.purchased.plan.consecutive.count = 5;
+      user.purchased.plan.consecutive.offset = 1;
+
+      cron({user, tasksByType, daysMissed, analytics});
+
+      expect(user.purchased.plan.customerId).to.exist;
+      expect(user.purchased.plan.consecutive.gemCapExtra).to.exist;
+      expect(user.purchased.plan.consecutive.count).to.exist;
+      expect(user.purchased.plan.consecutive.offset).to.exist;
+    });
+  });
+
   describe('user is sleeping', () => {
     beforeEach(() => {
       user.preferences.sleep = true;
@@ -165,9 +235,12 @@ describe('cron', () => {
       tasksByType.dailys.push(task);
       tasksByType.dailys[0].completed = true;
 
+      let healthBefore = user.stats.hp;
+
       cron({user, tasksByType, daysMissed, analytics});
 
       expect(tasksByType.dailys[0].completed).to.be.false;
+      expect(user.stats.hp).to.equal(healthBefore);
     });
   });
 
@@ -263,13 +336,18 @@ describe('cron', () => {
 
     it('should do less damage for missing a daily with partial completion', () => {
       daysMissed = 1;
+      let hpBefore = user.stats.hp;
       tasksByType.dailys[0].startDate = moment(new Date()).subtract({days: 1});
+      cron({user, tasksByType, daysMissed, analytics});
+      let hpDifferenceOfFullyIncompleteDaily = hpBefore - user.stats.hp;
+
+      hpBefore = user.stats.hp;
       tasksByType.dailys[0].checklist.push({title: 'test', completed: true});
       tasksByType.dailys[0].checklist.push({title: 'test2', completed: false});
-
       cron({user, tasksByType, daysMissed, analytics});
+      let hpDifferenceOfPartiallyIncompleteDaily = hpBefore - user.stats.hp;
 
-      expect(user.stats.hp).to.equal(49);
+      expect(hpDifferenceOfPartiallyIncompleteDaily).to.be.lessThan(hpDifferenceOfFullyIncompleteDaily);
     });
 
     it('should decrement quest progress down for missing a daily', () => {
@@ -310,6 +388,16 @@ describe('cron', () => {
       cron({user, tasksByType, daysMissed, analytics});
 
       expect(tasksByType.habits[0].value).to.be.lessThan(1);
+    });
+
+    it('should do nothing to habits with both up and down', () => {
+      tasksByType.habits[0].value = 1;
+      tasksByType.habits[0].up = true;
+      tasksByType.habits[0].down = true;
+
+      cron({user, tasksByType, daysMissed, analytics});
+
+      expect(tasksByType.habits[0].value).to.equal(1);
     });
   });
 
@@ -481,11 +569,5 @@ describe('cron', () => {
 
       expect(user.inbox.messages[messageId]).to.not.exist;
     });
-  });
-
-  it('increments user cron count', () => {
-    let cronCountBefore = user.flags.cronCount;
-    cron({user, tasksByType, daysMissed, analytics});
-    expect(user.flags.cronCount).to.be.greaterThan(cronCountBefore);
   });
 });
--- a/test/api/v3/unit/libs/cron.test.js
+++ b/test/api/v3/unit/libs/cron.test.js
@@ -1,0 +1,491 @@
+/* eslint-disable global-require */
+import moment from 'moment';
+import { cron } from '../../../../../website/src/libs/api-v3/cron';
+import { model as User } from '../../../../../website/src/models/user';
+import * as Tasks from '../../../../../website/src/models/task';
+import { clone } from 'lodash';
+import common from '../../../../../common';
+
+// const scoreTask = common.ops.scoreTask;
+
+describe('cron', () => {
+  let user;
+  let tasksByType = {habits: [], dailys: [], todos: [], rewards: []};
+  let daysMissed = 0;
+  let analytics = {
+    track: sinon.spy(),
+  };
+
+  beforeEach(() => {
+    user = new User({
+      auth: {
+        local: {
+          username: 'username',
+          lowerCaseUsername: 'username',
+          email: 'email@email.email',
+          salt: 'salt',
+          hashed_password: 'hashed_password', // eslint-disable-line camelcase
+        },
+      },
+    });
+
+    user._statsComputed = {
+      mp: 10,
+    };
+  });
+
+  it('updates user.auth.timestamps.loggedin and lastCron', () => {
+    let now = new Date();
+
+    cron({user, tasksByType, daysMissed, analytics, now});
+
+    expect(user.auth.timestamps.loggedin).to.equal(now);
+    expect(user.lastCron).to.equal(now);
+  });
+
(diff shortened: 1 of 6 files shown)


website/src/libs/api-v3/cron.js
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  // After all is said and done, progress up user's effect on quest, return those values & reset the user's
+  let progress = user.party.quest.progress;
+  let _progress = _.cloneDeep(progress);
+  _.merge(progress, {down: 0, up: 0});
+  progress.collect = _.transform(progress.collect, (m, v, k) => m[k] = 0);
+
+  // Clean PMs - keep 200 for subscribers and 50 for free users
+  let maxPMs = 200;
```
crookedneighbor: Maker this a const at the top of the file

website/src/libs/api-v3/cron.js
```diff
@@ -0,0 +1,274 @@
...
+  let numberOfPMs = Object.keys(user.inbox.messages).length;
+  if (Object.keys(user.inbox.messages).length > maxPMs) {
```
paglias: there's a duplicate call to Object.keys(user.inbox.messages).length. Also to be sure of not having problems with the special mongoose objects I'd do `user.inbox.message.toObject()` apply the changes to the new object and then replace user.inbox.messages with it

website/src/libs/api-v3/cron.js
```diff
@@ -0,0 +1,274 @@
...
+    _(user.inbox.messages)
+      .sortBy('timestamp')
+      .takeRight(numberOfPMs - maxPMs)
+      .each(pm => {
+        user.inbox.messages[pm.id] = undefined;
```
paglias: let's delete the object, not only set it to undefined

website/src/libs/api-v3/cron.js
TheHollidayInn: Like this? `delete user.inbox.messages[pm.id].toObject();`

TheHollidayInn:
Alright, updates have been made.

website/src/middlewares/api-v3/cron.js
```diff
@@ -399,11 +138,11 @@ module.exports = function cronMiddleware (req, res, next) {
     tasks.forEach(task => {
       if (task.isModified) toSave.push(task.save());
```
paglias: isModified() ?

test/helpers/api-unit.helper.js
```diff
@@ -70,3 +71,33 @@ export function generateHistory (days) {
...
+  let todo = {
+    text: 'test todo',
+    type: 'todo',
+    value: 0,
+    completed: false,
+  };
+
+  let task = new Tasks.todo(Tasks.Task.sanitize(todo)); // eslint-disable-line babel/new-cap
+  task.userId = user._id;
+  task.save();
```
paglias, line 85: save shouldn't be called for unit tests

paglias, line 85: you can simply return the task object

TheHollidayInn, line 85: Hmmm. If I don't save the object, I can't access it from the database after the middleware modifies it.

paglias, line 85: well since you pass an object that contains tasks objects you can use it after the middleware has run. But i See the point with completed todos since those are not loaded

paglias, line 85: so it's okay

test/api/v3/unit/middlewares/cronMiddleware.js
```diff
@@ -0,0 +1,170 @@
...
+    cronMiddleware(req, res, function callback () {
+      Tasks.Task.findOne({_id: task}, function (err, taskFound) {
+        expect(err).to.not.exist;
+        expect(taskFound).to.not.exist;
+        done();
+      });
+    });
+  });
+
+  it('should call next is user was not modified after cron', (done) => {
```
crookedneighbor, line 109: This test assertions says "it should call next", but never asserts that in the test.

crookedneighbor, line 109: Ahh, never mind, you swapped out passing in next _for_ a callback, so it's called implicitly by calling done(); Carry on.

test/api/v3/unit/middlewares/cronMiddleware.js
```diff
@@ -0,0 +1,170 @@
...
+    let hpBefore = res.locals.user.stats.hp;
+    res.locals.user.lastCron = moment(new Date()).subtract({days: 2});
+    generateDaily(res.locals.user);
+
+    cronMiddleware(req, res, function callback () {
```
crookedneighbor: All of these `function callback ()` things should be arrow functions:

``` js
cronMiddleware(req, res, () => {
```

You should need to access `this` in those functions, so there's no reason to make them named functions.

website/src/middlewares/api-v3/cron.js
```diff
@@ -5,274 +5,13 @@ import * as Tasks from '../../models/task';
...
-    progressDown: _progress.down,
-  });
-
-  return _progress;
-}
 
 module.exports = function cronMiddleware (req, res, next) {
   let user = res.locals.user;
+
   if (!user) return next(); // User might not be available when authentication is not mandatory
```
crookedneighbor, line 15: I don't see a test for this line.

test/api/v3/unit/middlewares/cronMiddleware.js
```diff
@@ -0,0 +1,170 @@
...
+    cronMiddleware(req, res, next);
+    expect(next).to.be.calledOnce;
+  });
+
+  it('should clear todos older than 30 days for free users', (done) => {
+    res.locals.user.lastCron = moment(new Date()).subtract({days: 2});
+    let task = generateTodo(res.locals.user);
+    task.dateCompleted = moment(new Date()).subtract({days: 31});
+    task.completed = true;
+    task.save();
```
crookedneighbor: Isn't this an asynchronous action?

test/api/v3/unit/middlewares/cronMiddleware.js
```diff
@@ -0,0 +1,170 @@
...
+
+    cronMiddleware(req, res, function callback () {
+      expect(res.locals.user.stats.hp).to.be.equal(hpBefore);
+      done();
+    });
+  });
+
+  it('does damage for missing dailies', (done) => {
+    let hpBefore = res.locals.user.stats.hp;
+    res.locals.user.lastCron = moment(new Date()).subtract({days: 2});
```
crookedneighbor: You realize that you can just do `user.lastCron = moment(new Date()).subtract({days: 2});` since `res.locals.user` is just a reference to `user`

test/api/v3/unit/middlewares/cronMiddleware.js
```diff
@@ -0,0 +1,170 @@
...
+    generateDaily(res.locals.user);
+
+    cronMiddleware(req, res, function callback () {
+      expect(res.locals.user.stats.hp).to.be.equal(hpBefore);
+      done();
+    });
+  });
+
+  it('does damage for missing dailies', (done) => {
```
crookedneighbor, line 120: Is this a concern of cron middleware? Doesn't this effect happen inside the cron lib?

test/api/v3/unit/middlewares/cronMiddleware.js
```diff
@@ -0,0 +1,170 @@
...
+    daily.startDate = moment(new Date()).subtract({days: 2});
+    daily.save();
+
+    cronMiddleware(req, res, function callback () {
+      expect(res.locals.user.stats.hp).to.be.lessThan(hpBefore);
+      done();
+    });
+  });
+
+  it('updates tasks', (done) => {
```
crookedneighbor, line 133: I thin this is cron's concern, not the middleware.

test/api/v3/unit/middlewares/cronMiddleware.js
```diff
@@ -0,0 +1,170 @@
...
+    cronMiddleware(req, res, function callback () {
+      Tasks.Task.findOne({_id: todo._id}, function (err, todoFound) {
+        expect(err).to.not.exist;
+        expect(todoFound.value).to.be.lessThan(todoValueBefore);
+        done();
+      });
+    });
+  });
+
+  it('applies quest progress', (done) => {
```
crookedneighbor: I think this is a cron concern.

test/api/v3/unit/middlewares/cronMiddleware.js
```diff
@@ -0,0 +1,170 @@
...
+    party.quest.key = questKey;
+    party.save();
+
+    res.locals.user.party._id = party._id;
+    res.locals.user.save();
+
+    party.startQuest(res.locals.user);
+
+    cronMiddleware(req, res, function callback () {
+      expect(res.locals.user.stats.hp).to.be.lessThan(hpBefore);
```
crookedneighbor: this does not actually assert on anything related to the quest. How do we know that this was boss damage on not regular daily damage?

test/api/v3/unit/middlewares/cronMiddleware.js
```diff
@@ -0,0 +1,170 @@
...
+    res.locals.user.party.quest.key = questKey;
+
+    let party = new Group({
+      type: 'party',
+      name: generateUUID(),
+      leader: res.locals.user._id,
+    });
+    party.quest.members[res.locals.user._id] = true;
+    party.quest.key = questKey;
+    party.save();
```
crookedneighbor: Asynchronous action?

test/api/v3/unit/middlewares/cronMiddleware.js
```diff
@@ -0,0 +1,170 @@
...
+
+    res.locals.user.party._id = party._id;
+    res.locals.user.save();
```
crookedneighbor: Asynchronous action?

website/src/libs/api-v3/cron.js
```diff
@@ -0,0 +1,274 @@
+import moment from 'moment';
+import common from '../../../../common/';
+import { preenUserHistory } from '../../libs/api-v3/preening';
+import _ from 'lodash';
+
+const shouldDo = common.shouldDo;
+const scoreTask = common.ops.scoreTask;
+// const maxPMs = 200;
+
+let clearBuffs = {
```
crookedneighbor: Should this be `const CLEAR_BUFFS = {` ?

website/src/libs/api-v3/cron.js
```diff
@@ -0,0 +1,274 @@
...
+  user.auth.timestamps.loggedin = now;
+  user.lastCron = now;
+  user.preferences.timezoneOffsetAtLastCron = timezoneOffsetFromUserPrefs;
+  // Reset the lastDrop count to zero
+  if (user.items.lastDrop.count > 0) user.items.lastDrop.count = 0;
+
+  // "Perfect Day" achievement for perfect-days
+  let perfect = true;
+
+  // end-of-month perks for subscribers
```
crookedneighbor: I don't thin this comment is necessary. Doesn't the code say exactly that?

crookedneighbor:
I could be reading this wrong, but I think there are quite a few places where the cron middleware tests are testing what `cron` is doing. I think most of what cron middleware tests _should_ be doing is going through the various paths that would cause `cron` to run or not.

test/api/v3/unit/libs/cron.test.js
```diff
@@ -0,0 +1,491 @@
...
+    expect(user.preferences.timezoneOffsetAtLastCron).to.equal(timezoneOffsetFromUserPrefs);
+  });
+
+  it('resets user.items.lastDrop.count', () => {
+    user.items.lastDrop.count = 4;
+    cron({user, tasksByType, daysMissed, analytics});
+    expect(user.items.lastDrop.count).to.equal(0);
+  });
+
+  describe('end of the month perks', () => {
```
crookedneighbor, line 66: Probably should be a test that this stuff doesn't happen when user is not a subscriber.

test/api/v3/unit/libs/cron.test.js
```diff
@@ -0,0 +1,491 @@
...
+        startDate: new Date(),
+      };
+
+      let task = new Tasks.daily(Tasks.Task.sanitize(daily)); // eslint-disable-line babel/new-cap
+      tasksByType.dailys.push(task);
+      tasksByType.dailys[0].completed = true;
+
+      cron({user, tasksByType, daysMissed, analytics});
+
+      expect(tasksByType.dailys[0].completed).to.be.false;
```
crookedneighbor, line 242: I don't see an expectation about user's hp.

test/api/v3/unit/libs/cron.test.js
```diff
@@ -0,0 +1,491 @@
...
+      expect(tasksByType.dailys[0].history).to.be.lengthOf(1);
+    });
+
+    it('should set tasks completed to false', () => {
+      tasksByType.dailys[0].completed = true;
+      cron({user, tasksByType, daysMissed, analytics});
+      expect(tasksByType.dailys[0].completed).to.be.false;
+    });
+
+    it('should set task checklist to completed for completed dailys', () => {
```
crookedneighbor, line 301: Is this accurate? I may be misunderstanding, but this seems wrong to me.

crookedneighbor, line 301: Hmm... I see now that you're saying that completed dailies with incomplete checklists should be treated as if all their checklist items were completed. 

I guess the actual saving of the tasks and resetting them to their uncompleted state happens elsewhere?

test/api/v3/unit/libs/cron.test.js
```diff
@@ -0,0 +1,491 @@
...
+
+    it('should do less damage for missing a daily with partial completion', () => {
+      daysMissed = 1;
+      tasksByType.dailys[0].startDate = moment(new Date()).subtract({days: 1});
+      tasksByType.dailys[0].checklist.push({title: 'test', completed: true});
+      tasksByType.dailys[0].checklist.push({title: 'test2', completed: false});
+
+      cron({user, tasksByType, daysMissed, analytics});
+
+      expect(user.stats.hp).to.equal(49);
```
crookedneighbor: This seems dangerous to me. If we tweak the algorithm even the tinies bit, this will fail...

Also, does this happen _in_ the cron code? Or as part of the score task method?

test/api/v3/unit/libs/cron.test.js
```diff
@@ -0,0 +1,491 @@
...
+    });
+
+    it('should decrement only down value', () => {
+      tasksByType.habits[0].value = 1;
+      tasksByType.habits[0].up = false;
+
+      cron({user, tasksByType, daysMissed, analytics});
+
+      expect(tasksByType.habits[0].value).to.be.lessThan(1);
+    });
```
crookedneighbor, line 391: Shouldn't you have a test that has both up and down buttons?

test/api/v3/unit/libs/cron.test.js
```diff
@@ -0,0 +1,491 @@
...
+        flagCount: 0,
+      };
+
+      cron({user, tasksByType, daysMissed, analytics});
+
+      expect(user.inbox.messages[messageId]).to.not.exist;
+    });
+  });
+
+  it('increments user cron count', () => {
```
crookedneighbor: This should probably be up at the top of the file with the other tests like it.

crookedneighbor:
Left lots of nitpicky comments. Overall, 👍

test/api/v3/unit/middlewares/cronMiddleware.js
TheHollidayInn, line 120: Yes, but what I am testing here is that the cron middleware is making it through the full call and saving the data. The saving is the big part.

test/api/v3/unit/middlewares/cronMiddleware.js
TheHollidayInn, line 133: Same comment as above.

test/api/v3/unit/middlewares/cronMiddleware.js
TheHollidayInn: And again.

website/src/libs/api-v3/cron.js
TheHollidayInn: Sounds good. I didn't really remove any comments or refactor much.

test/api/v3/unit/libs/cron.test.js
TheHollidayInn, line 301: Eh, I'm not sure to be honest.

test/api/v3/unit/middlewares/cronMiddleware.js
TheHollidayInn: I agree. There is not explicit correlation being asserted. However, cron does not seem to reward and the only quest progress saved is the damage done - after reviewing Group.bossQuest and Group.collectQuest. So, I'm not sure what to do. But, I think if we eventually refactor this, we can remove this function from the middleware. I think responsibility of saving this data should be handled elsewhere.

test/api/v3/unit/libs/cron.test.js
Alys, line 301: I haven't yet looked at the code, but from a functionality point of view:

When a Daily has been marked completed, the status of the checklist items is irrelevant and the only thing that cron needs to do with the checklist items is untick them. For a completed Daily, the number of checklist items ticked should not have any effect on the change in the task's value or the user's stats.

TheHollidayInn:
This has been updated.

PR #7085, opened 2016-04-15
chore(lint): lint files in preparation for eslint v2

crookedneighbor:
This preps api-v3 branch for the eslint upgrade in `develop`.

### DIFF
--- a/website/src/controllers/api-v3/chat.js
+++ b/website/src/controllers/api-v3/chat.js
@@ -224,7 +224,7 @@ api.flagChat = {
     if (group._id === TAVERN_ID) {
       groupUrl = '/#/options/groups/tavern';
     } else if (group.type === 'guild') {
-      groupUrl = `/#/options/groups/guilds/{$group._id}`;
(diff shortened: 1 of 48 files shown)


website/src/controllers/api-v3/chat.js
```diff
@@ -224,7 +224,7 @@ api.flagChat = {
     if (group._id === TAVERN_ID) {
       groupUrl = '/#/options/groups/tavern';
     } else if (group.type === 'guild') {
-      groupUrl = `/#/options/groups/guilds/{$group._id}`;
+      groupUrl = `/#/options/groups/guilds/${group._id}`;
```
crookedneighbor, line 227: Glad `eslint` caught this :/

crookedneighbor:
Summary of changes:
- If a string used backtickes (`), but did not use string interpolation, it was changed to a single quote.
- Corrected 3 instances where strings were formatted like `foo {$var}` instead of `foo ${var}`
- cases in switch statement got braces
- Added a few disables for process.env and global require

PR #7096, opened 2016-04-21
V3 payments 5

piousbox:
.


website/src/controllers/top-level/payments/amazon.js
```diff
@@ -130,31 +131,22 @@ api.checkout = {
 
       await amzLib.closeOrderReference({ AmazonOrderReferenceId: orderReferenceId });
 
+      // execute payment
+      let giftUser = await User.findById(gift ? gift.uuid : undefined);
```
paglias, line 135: this line can be moved directly into `if (gift)` because if gift is undefined it won't fetch any user

PR #7099, opened 2016-04-21
fix(markdown): open md links in new tab

SabreCat:
### Changes

A recent change made for security reasons incidentally changed the handling of URLs in Markdown to open in the existing window instead of a new tab. This PR uses a feature of the new Markdown library to add the `target='_blank'` to all Markdown URLs, causing them to open in a new tab.

_Needs investigation/discussion_: Is it in fact desirable to have all Markdown links open new tabs? This could be a nuisance for links opening onto pages within Habitica.

---

UUID: wut

### DIFF
--- a/common/script/public/directives.js
+++ b/common/script/public/directives.js
@@ -9,7 +9,8 @@
     var remarkable = new Remarkable({
       // TODO: Add in code highlighting?
       // highlight: function (#<{(|str, lang|)}>#) { return ''; }
-      linkify: true
+      linkify: true,
+      linkTarget: '_blank',
     });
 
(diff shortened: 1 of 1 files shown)


Alys:
I think it's more likely that markdown links will point to other websites than to Habitica pages. I'm firmly on the site of opening new tabs for all markdown links.

SabreCat:
@Alys: The most common case I can think of is "Check out the Party Wanted Guild!" with a link to the Guild. (Or any other Guild for that matter.) But that's probably not terrible to have open in a new tab anyway!

crookedneighbor:
@SabreCat I'm working on an alternate solution for this, so I'm going to close this.

PR #7102, opened 2016-04-22
payments api: cancelSubscription

piousbox:


website/src/libs/api-v3/payments.js
```diff
@@ -130,33 +128,34 @@ api.createSubscription = function createSubscription (data, cb) {
...
   let remaining = data.nextBill ? moment(data.nextBill).diff(new Date(), 'days') : 30;
+  let nowStr = `${now.format('MM')}/${moment(plan.dateUpdated).format('DD')}/${now.format('YYYY')}`;
+  let nowStrFormat = 'MM/DD/YYYY';
 
   plan.dateTerminated =
-    moment(`${now.format('MM')}/${moment(plan.dateUpdated).format('DD')}/${now.format('YYYY')}`)
+    moment(nowStr, nowStrFormat)
     .add({days: remaining}) // end their subscription 1mo from their last payment
-    .add({months: Math.ceil(plan.extraMonths)})// plus any extra time (carry-over, gifted subscription, etc) they have. FIXME: moment can't add months in fractions...
+    .add({days: Math.ceil(30 * plan.extraMonths)}) // plus any extra time (carry-over, gifted subscription, etc) they have.
```
paglias, line 141: why this was changed to months?

website/src/libs/api-v3/payments.js
```diff
@@ -130,33 +128,34 @@ api.createSubscription = function createSubscription (data, cb) {
...
     .toDate();
   plan.extraMonths = 0; // clear extra time. If they subscribe again, it'll be recalculated from p.dateTerminated
 
-  data.user.save(cb);
+  data.user.save();
```
paglias, line 145: we have to handle async ops like this somehow even if we're not using callbacks anymore, either return data.user.save(); ( a promise ) or transform this whole function in an async one

website/src/libs/api-v3/payments.js
piousbox, line 141: This was changed from months to days to accommodate partial months, acting on the comment " FIXME: moment can't add months in fractions"

website/src/libs/api-v3/payments.js
piousbox: ok, convert to async it is.

paglias:
What's the status of this?

piousbox:
This was superseded by `v3-payments-7-stripe`, and an await was added on that `data.user.save()` call.

PR #7104, opened 2016-04-22
V3 payments 6

piousbox:
.


test/api/v3/unit/libs/amazonPayments.test.js
```diff
@@ -3,33 +3,43 @@ import * as amzLib from '../../../../../website/src/libs/api-v3/amazonPayments';
...
     let amzOldConnect;
 
     beforeEach(() => {
       amzOldConnect = amazonPayments.connect;
       amazonPayments.connect = () => {
-        let api = { getTokenInfo: (token, cb) => {
-          return cb(undefined, thisToken);
-        } };
-        return { api };
+        let api = {
```
paglias: can't you stub the amazon payments library that use promises instead of stubbing the ones that use callbacks and we don't use directly (so we may be missing some error in our own implementation). Also for stubbing you should use sinon.js (http://sinonjs.org/) not replace objects directly

common/script/cron.js
```diff
@@ -28,7 +28,7 @@ function sanitizeOptions (o) {
   let dayStart = !_.isNaN(ref) && ref >= 0 && ref <= 24 ? ref : 0;
 
   let timezoneOffset;
-  let timezoneOffsetDefault = Number(moment().zone());
+  let timezoneOffsetDefault = Number(moment().utcOffset());
```
paglias: let's not change these. In the deprecation warning it says that the new value can be different from the previous one

website/src/controllers/api-v3/chat.js
```diff
@@ -12,8 +12,9 @@ import _ from 'lodash';
 import { removeFromArray } from '../../libs/api-v3/collectionManipulators';
 import { sendTxn } from '../../libs/api-v3/email';
 import nconf from 'nconf';
-import setupNconf from '../../libs/api-v3/setupNconf';
-setupNconf();
+// @TODO: move to test?
+// import setupNconf from '../../libs/api-v3/setupNconf';
+// setupNconf();
```
paglias: remove

website/src/controllers/api-v3/auth.js
```diff
@@ -2,8 +2,11 @@ import validator from 'validator';
 import moment from 'moment';
 import passport from 'passport';
 import nconf from 'nconf';
+
+// @TODO: move to test?
 import setupNconf from '../../libs/api-v3/setupNconf';
```
paglias: these shall be removed

website/src/controllers/top-level/payments/amazon.js
```diff
@@ -213,68 +211,57 @@ api.subscribe = function subscribe (req, res, next) {
...
+api.subscribeCancel = {
+  method: 'GET',
+  url: '/payments/amazon/subscribeCancel',
+  middlewares: [authWithHeaders()],
+  async handler (req, res) {
+    let user = res.locals.user;
+    let billingAgreementId = user.purchased.plan.customerId;
 
-  let billingAgreementId = user.purchased.plan.customerId;
+    if (!user.purchased.plan.customerId) throw new BadRequest(res.t('missingSubscription'));
```
paglias: you can check for the billingAgreementId variable

website/src/libs/api-v3/amazonPayments.js
```diff
@@ -70,24 +46,15 @@ api.authorize = (inputSet) => {
...
-  let amzPayment = connect(amazonPayments);
-  return new Promize((resolve, reject) => {
-    amzPayment.offAmazonPayments.closeOrderReference(inputSet, (err, response) => {
-      if (err) return reject(err);
-      return resolve(response);
-    });
-  });
-};
-
-api.executePayment = (inputSet) => {
```
paglias, line 83: this is missing from the exported object

website/src/libs/api-v3/payments.js
```diff
@@ -192,21 +181,12 @@ api.buyGems = function buyGems (data, cb) {
...
-        data.gift.member.save(cb2);
-      } else {
-        cb2(null);
-      }
-    },
-  ], cb);
+  await data.user.save();
 };
 
+// @TODO: this shouldn't be here or should not be a middleware
```
paglias: this can be removed, I ported it separately

website/src/libs/api-v3/amazonPayments.js
piousbox, line 83: the same functionality is `closeOrderReference()` function. Further, I searched for executePayment and only found it for paypal:

``````
vagrant@dhcppc6:~/projects/habitrpg/website$ grep -R executePayment .
./src/libs/api-v3/payments.js:api.paypalCheckoutSuccess = paypal.executePayment;
./src/libs/api-v3/payments.js~:api.paypalCheckoutSuccess = paypal.executePayment;
./src/controllers/top-level/payments/paypal.js~:exports.executePayment = function(req, res) {
./src/controllers/top-level/payments/paypal.js:exports.executePayment = function(req, res) {```
``````

Comment (paglias) [inline website/src/libs/api-v3/amazonPayments.js:83]: @piousbox the executePayment code was a route and is used by an external job queue we use to bill amazon payments subscribers

File: common/script/cron.js [hunk 8]
```diff
@@ -41,7 +41,7 @@ function sanitizeOptions (o) {
     timezoneOffset = timezoneOffsetDefault;
   }
 
-  let now = o.now ? moment(o.now).zone(timezoneOffset) : moment().zone(timezoneOffset);
+  let now = o.now ? moment(o.now).utcOffset(timezoneOffset) : moment().utcOffset(timezoneOffset);
```
paglias: these code should not be changed

SabreCat: This looks like it's meant to address the fact that Moment has deprecated `zone` and keeps blaring that fact into our server logs. Is now not a good time to do that?

website/src/libs/api-v3/amazonPayments.js
piousbox, line 83: Great, I can add it back in. What is the form of that route, `POST /amazon/executePayment`? I do not find it in this router: https://github.com/HabitRPG/habitrpg/blob/develop/website/src/routes/payments.js

common/script/cron.js
paglias: @SabreCat the problem is that our data could not be compatible since utcOffset doesn't always return the same data as zone()

I think @Alys wanted to change this but wasn't able to

website/src/libs/api-v3/amazonPayments.js
paglias, line 83: it's not a route.... this is the library

paglias, line 83: anyway it's used here https://github.com/HabitRPG/habitrpg/blob/develop/website/src/controllers/payments/amazon.js#L123

piousbox, line 83: So, that's part of the checkout route, that object isn't exported, it's labeled (for readability) and executed as async.series. We're getting rid of async, right? that `executePayment()` function finds user, does something with gifts, and calls `payments[method](data ...` the equivalent is here: https://github.com/piousbox/habitrpg/blob/v3-payments-6/website/src/controllers/top-level/payments/amazon.js#L141 and the label became a comment for readability: https://github.com/piousbox/habitrpg/blob/v3-payments-6/website/src/controllers/top-level/payments/amazon.js#L132

paglias, line 83: yeah... sorry I'm a bit tired and didn't see it. Just undo the changes to cron and I'll merge

piousbox, line 83: Done, zone() is back in.

common/script/cron.js
Alys: The old and new functions never return the same value.

I have not tried to change it. I know how I want to do it but it should be done later. To much chance of chaos if it's bundled into the v3 changes. See https://github.com/HabitRPG/habitrpg/pull/7053#issuecomment-207950843

PR #7108, opened 2016-04-23
Ported groups service to user new api v3 and ported dependent controllers

TheHollidayInn:
Fixes put_issue_url_here
### Changes

Updated Groups service with api-v3
Ported functionality in Party, Guilds, and Group ctrl.

---

UUID:

### DIFF
--- a/test/spec/services/groupServicesSpec.js
+++ b/test/spec/services/groupServicesSpec.js
@@ -2,6 +2,7 @@
 
 describe('groupServices', function() {
   var $httpBackend, $http, groups, user;
+  var groupApiUrlPrefix = '/api/v3/groups';
 
   beforeEach(function() {
     module(function($provide) {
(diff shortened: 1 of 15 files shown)


website/public/js/controllers/guildsCtrl.js
```diff
@@ -4,91 +4,110 @@ habitrpg.controller("GuildsCtrl", ['$scope', 'Groups', 'User', 'Challenges', '$r
...
+            $rootScope.hardRedirect('/#/options/groups/guilds/' + joinedGroup._id);
+          });
+      }
+
+      $scope.reject = function(guild) {
+        Groups.Group.rejectInvite(guild.id);
+        var i = _.findIndex(User.user.invitations.guilds, {id: guild.id});
+        if (i !== -1) {
+          User.user.invitations.guilds.splice(i, 1);
+          User.set({'invitations.guilds':User.user.invitations.guilds});
```
paglias: this should use the new reject invitation route

paglias: which seems to be used at line 60 and should take care of removing the invite from the user document

website/public/js/services/groupServices.js
```diff
@@ -1,80 +1,164 @@
...
-          transformResponse: function(data) {
-            data = angular.fromJson(data);
-            _.each(data && data.challenges, function(c) {
-              angular.extend(c, Challenges.Challenge.prototype);
-            });
-            return data;
-          }
+angular.module('habitrpg')
+.factory('Groups', [ '$location', '$resource', '$rootScope', '$http', 'Analytics', 'ApiUrl', 'Challenges','User',
+  function($location, $resource, $rootScope, $http, Analytics, ApiUrl, Challenges, User) {
```
paglias: $resource is not used anymore, right? it should be removed

test/spec/services/groupServicesSpec.js
```diff
@@ -16,27 +17,121 @@ describe('groupServices', function() {
...
 
+  it('returns tavern if cached', function (done) {
+    var uid = 'abc';
+    var tavern = {
+      _id: uid,
+    };
+    groups.data.tavern = tavern;
+    groups.tavern()
+      .then(function (tavern) {
+        expect(tavern).to.eql(tavern);
```
crookedneighbor: This is not testing anything since the `tavern` variable in this function will reference the argument in the function, and not the variable you instantiated earlier. Should be something like this:

``` js
groups.tavern().then(function (result) {
  expect(result).to.eql(tavern);
});
```

test/spec/services/groupServicesSpec.js
```diff
@@ -16,27 +17,121 @@ describe('groupServices', function() {
...
 
+  it('returns public guilds if cached', function (done) {
+    var uid = 'abc';
+    var publicGuilds = [
+      {_id: uid},
+    ];
+    groups.data.publicGuilds = publicGuilds;
+
+    groups.publicGuilds()
+      .then(function (publicGuilds) {
```
crookedneighbor: Same problem here. This isn't actually testing anything. `publicGuilds` could be `"foo"` and it would still pass.

website/public/js/services/groupServices.js
```diff
@@ -1,80 +1,180 @@
...
 
     function party(cb) {
-      if (!data.party) return (data.party = Group.get({gid: 'party'}, cb));
-      return (cb) ? cb(party) : data.party;
+      if (!data.party) {
+        Group.get('party')
+          .then(function successCallback(response) {
+            data.party = response.data.data;
+          }, function errorCallback(response) {
+
```
crookedneighbor: Are we going to do anything with the error callback? If not, why have it?

website/src/controllers/api-v3/groups.js
```diff
@@ -103,7 +103,7 @@ api.getGroups = {
     let results = await Group.getGroups({user, types, groupFields, sort});
 
     // If no valid value for type was supplied, return an error
-    if (results.length === 0) throw new BadRequest(res.t('groupTypesRequired'));
+    // if (results.length === 0) throw new BadRequest(res.t('groupTypesRequired'));
```
crookedneighbor: Why was this commented out?

crookedneighbor:
Left a few comments. Overall, seems reasonable, assuming it works.

TheHollidayInn:
Updated!

website/public/js/services/groupServices.js
```diff
@@ -1,80 +1,183 @@
...
+    function party () {
+      var deferred = $q.defer();
+
+      if (!data.party) {
+        Group.get('party')
+          .then(function (response) {
+            data.party = response.data.data;
+            deferred.resolve(data.party);
+          });
```
paglias, line 114: if there's an error here it should reject the deferred?

website/public/js/services/groupServices.js
```diff
@@ -1,80 +1,183 @@
...
 
-    function publicGuilds() {
+    function publicGuilds () {
+      var deferred = $q.defer();
+
+      if (!data.publicGuilds) {
+        Group.getGroups('publicGuilds')
+          .then(function (response) {
+            data.publicGuilds = response.data.data;
+            deferred.resolve(data.publicGuilds);
```
paglias, line 129: same here

website/public/js/services/groupServices.js
```diff
@@ -1,80 +1,183 @@
...
-      if (!data.myGuilds) data.myGuilds = Group.query({type:'guilds'});
-      return data.myGuilds;
+    function myGuilds () {
+      var deferred = $q.defer();
+
+      if (!data.myGuilds) {
+        Group.getGroups('privateGuilds')
+          .then(function (response) {
+            data.myGuilds = response.data.data;
+            deferred.resolve(data.myGuilds);
```
paglias, line 148: and here

website/public/js/services/groupServices.js
```diff
@@ -1,80 +1,183 @@
...
-      return data.tavern;
+    function tavern () {
+      var deferred = $q.defer();
+
+      if (!data.tavern) {
+        Group.get('habitrpg')
+          .then(function (response) {
+            data.tavern = response.data.data;
+            deferred.resolve(data.tavern);
+          });
```
paglias, line 169: here too

TheHollidayInn:
Updated!

PR #7110, opened 2016-04-24
Disabling Frost and Stealth when all dailies are already effected  [merged in b97d7cc]

Hus274:
Fixes https://github.com/HabitRPG/habitrpg/issues/6867
### Changes

This enables new functionality for the stealth and frost skills. It updates the skill to disable it when the benefit of the skill has already been gained and there's no additional benefit to casting it again. 

This change was limited to the jade file in order to not interfere with the api v3 changes that are currently going on. Pictures of the changes are in the issue linked here.

---

### DIFF
--- a/website/views/shared/tasks/task_view/skills.jade
+++ b/website/views/shared/tasks/task_view/skills.jade
@@ -7,27 +7,21 @@ ul.items.rewards
       +specialSkill(dispel,skill)
 
 // Class Skills
+- var frostNoAdditionalBenefitText = '{{k=="frost" && user.stats.buffs.streaks ? env.t("spellWizardFrostAlreadyCast") : ""}}'
+- var stealthNumberOfCastsText = '{{k=="stealth" ? env.t("spellRogueStealthDaliesAvoided", { number: user.stats.buffs.stealth }) : ""}}'
+- var stealthNoAdditionalBenefitText = '{{k=="stealth" && user.stats.buffs.stealth >= user.dailys.length ? env.t("spellRogueStealthMaxedOut") : ""}}'
 ul.items(ng-if='main && list.type=="reward" && user.stats.class && !user.preferences.disableClasses')
   li.task.reward-item.list-group-item(ng-repeat='(k,skill) in Content.spells[user.stats.class]', ng-if='user.stats.lvl >= skill.lvl',popover-trigger='mouseenter', popover-placement='top',
-  popover='{{skill.notes()}}{{k=="frost" && user.stats.buffs.streaks ? " You have already cast this today. Your streaks are frozen, and there\'s no need to cast this again." : ""}}}{{k=="stealth" ? " Number of dailies avoided: " + user.stats.buffs.stealth + "." : ""}}{{k=="stealth" && user.stats.buffs.stealth >= user.dailys.length ? " You have already avoided all your dailies, there\'s no need to cast this again." : ""}}')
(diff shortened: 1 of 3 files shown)


website/views/shared/tasks/task_view/skills.jade
```diff
@@ -8,16 +8,26 @@ ul.items.rewards
 
 // Class Skills
 ul.items(ng-if='main && list.type=="reward" && user.stats.class && !user.preferences.disableClasses')
-  li.task.reward-item(ng-repeat='(k,skill) in Content.spells[user.stats.class]', ng-if='user.stats.lvl >= skill.lvl',popover-trigger='mouseenter', popover-placement='top', popover='{{skill.notes()}}')
+  li.task.reward-item.list-group-item(ng-repeat='(k,skill) in Content.spells[user.stats.class]', ng-if='user.stats.lvl >= skill.lvl',popover-trigger='mouseenter', popover-placement='top',
+  popover='{{skill.notes()}}{{k=="frost" && user.stats.buffs.streaks ? " You have already cast this today. Your streaks are frozen, and there\'s no need to cast this again." : ""}}}{{k=="stealth" ? " Number of dailies avoided: " + user.stats.buffs.stealth + "." : ""}}{{k=="stealth" && user.stats.buffs.stealth >= user.dailys.length ? " You have already avoided all your dailies, there\'s no need to cast this again." : ""}}')
```
crookedneighbor: These should be made into translatable strings:

```
"You have already cast this today. Your streaks are frozen, and there\'s no need to cast this again."
" Number of dailies avoided: <%= number %>."
" You have already avoided all your dailies, there\'s no need to cast this again."
```

I understand that this spaghetti code is only here until we can actually do this work on the backend, but can we at least extract these pieces into Jade variables so it's easier to read?

website/views/shared/tasks/task_view/skills.jade
```diff
@@ -8,16 +8,26 @@ ul.items.rewards
...
       span.task-notes
         span.glyphicon.glyphicon-comment
     //left-hand size commands
     .task-controls.task-primary
-      a.money.btn-buy.item-btn(ng-click='castStart(skill)', ng-class='{active: skill.key == spell.key}')
+      a.money.btn-buy.item-btn(ng-click='castStart(skill)', ng-if='!( k=="frost" || k=="stealth" )')
+        span.reward-cost
+          strong {{::skill.mana}}
+          =env.t('mp')
+      a.btn.money.btn-buy.item-btn(ng-click='castStart(skill)', ng-class='{"disabled": user.stats.buffs.streaks}', ng-if='k=="frost"')
```
crookedneighbor: Instead of repeating this three times, you could just make a function that determines if the disabled class should be applied or not. IE:

``` js
// if is frost and streaks buff is applied - disabled
// if is stealth and stealth buff greater than dailies - disabled
// else not disabled
```

website/views/shared/tasks/task_view/skills.jade
Hus274: Definitely a good call. I was able to make this work so now the popover line reads as follows:
  popover='{{skill.notes()}}#{frostNoAdditionalBenefitText}#{stealthNumberOfCastsText}#{stealthNoAdditionalBenefitText}')

Hus274:
Where were you thinking the function that determines if the disabled class should be applied should go?

crookedneighbor:
I assume in the task controller.

Hus274:
Ahh gotcha. I'm new to front end development so I wasn't really sure what the controller was as I didn't see ng-controller anywhere.

Hus274:
Posted updated version after addressing comments.

crookedneighbor:

PR #7113, opened 2016-04-25
Updated chat service to use api v3

TheHollidayInn:
Fixes put_issue_url_here
### Changes

Ported chat service to use api v3

---

UUID:

### DIFF
--- a/website/public/js/controllers/menuCtrl.js
+++ b/website/public/js/controllers/menuCtrl.js
@@ -26,7 +26,7 @@ angular.module('habitrpg')
         }
       }
 
-      $scope.clearMessages = Chat.seenMessage;
+      $scope.clearMessages = Chat.markChatSeen;
       $scope.clearCards = Chat.clearCards;
 
       $scope.iconClasses = function() {
--- a/website/public/js/services/chatServices.js
+++ b/website/public/js/services/chatServices.js
@@ -1,8 +1,8 @@
 'use strict';
 
 angular.module('habitrpg')
-.factory('Chat', ['$resource', '$http', 'ApiUrl', 'User',
-  function($resource, $http, ApiUrl, User) {
+.factory('Chat', ['$http', 'ApiUrl', 'User',
+  function($http, ApiUrl, User) {
     var apiV3Prefix = '/api/v3';
 
     function getChat (groupId) {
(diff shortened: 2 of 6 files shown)


website/public/js/services/chatServices.js
```diff
@@ -1,33 +1,85 @@
...
+      getChat: getChat,
+      postChat: postChat,
+      deleteChat: deleteChat,
+      like: like,
+      flagChatMessage: flagChatMessage,
+      clearFlagCount: clearFlagCount,
+      markChatSeen: markChatSeen,
+    }
+
+    function clearCards() {
```
crookedneighbor, line 85: This function isn't exposed.

paglias, line 85: I think it's not exported because it's an User operation and they'll be ported later

website/public/js/services/chatServices.js
```diff
@@ -1,33 +1,85 @@
...
-
-  function seenMessage(gid) {
-    // On enter, set chat message to "seen"
-    $http.post(ApiUrl.get() + '/api/v2/groups/'+gid+'/chat/seen');
-    if (User.user.newMessages) delete User.user.newMessages[gid];
-  }
-}]);
+angular.module('habitrpg')
+.factory('Chat', ['$resource', '$http', 'ApiUrl', 'User',
+  function($resource, $http, ApiUrl, User) {
```
paglias: $resource is not used

website/public/js/services/chatServices.js
```diff
@@ -1,33 +1,85 @@
...
-  return chatService;
-
-  function clearCards() {
-    User.user.ops.update && User.set({'flags.cardReceived':false});
-  }
-
-  function seenMessage(gid) {
-    // On enter, set chat message to "seen"
-    $http.post(ApiUrl.get() + '/api/v2/groups/'+gid+'/chat/seen');
-    if (User.user.newMessages) delete User.user.newMessages[gid];
```
paglias, line 31: this line is missing in the new code

website/public/js/controllers/menuCtrl.js
```diff
@@ -26,7 +26,7 @@ angular.module('habitrpg')
         }
       }
 
-      $scope.clearMessages = Chat.seenMessage;
+      $scope.clearMessages = Chat.markChatSeen;
       $scope.clearCards = Chat.clearCards;
```
crookedneighbor, line 30: This will be undefined and throw an error anyone tries to call `$scope.clearCards` because `clearCards` wasn't exported. If this needs to be ported somewhere else, you should at least put a TODO comment here.

TheHollidayInn:
Updated

PR #7116, opened 2016-04-26
add static login page

librarianmage:
Fixes https://github.com/HabitRPG/habitrpg/issues/2413
### Changes
- Adds a static log in page at /static/login
- Adds a message with a link to /static/login in the login modal

---


librarianmage:
screenshot 2016-04-26 at 10 42 58 am

librarianmage:
EDIT: I have added a link in the modal

Alys:
Could you please include a screenshot of the modal with the link?

librarianmage:
@Alys 
screenshot 2016-04-26 at 6 33 15 pm
I am not quite sold on the wording...

crookedneighbor:
Is that supposed to be a link/button? Does not look like one. I wouldn't guess by looking at it that it was supposed to be interacted with.

librarianmage:
Oh, it is supposed to be a link. I can try changing it into a button

librarianmage:
@Alys @crookedneighbor How is this?
screenshot 2016-04-26 at 8 02 05 pm

website/views/static/login.jade
```diff
@@ -0,0 +1,83 @@
...
+block content
+  .row
+    .col-md-12
+      .page-header
+        .row
+          .col-md-8
+            h1=env.t('loginAndReg')
+          .col-md-4
+            a.zocial.facebook(alt=env.t('loginFacebookAlt'), ng-click='socialLogin("facebook")')=env.t('loginFacebookAlt')
+            //-ul.list-inline
```
khipkin: Why not just delete this?

website/views/static/login.jade
```diff
@@ -0,0 +1,83 @@
...
+            .form-group
+              input.form-control(type='password', ng-model='loginPassword', placeholder=env.t('password'), name='password', required)
+            .form-group
+              input.btn.btn-block.btn-lg.btn-primary(type='submit', ng-disabled='loginForm.$invalid', value=env.t('login'))
+
+            small=env.t('passMan')
+            br
+            br
+
+          // good god accordions have html ceremony
```
khipkin: Should remove this since it'll show up in the html. :stuck_out_tongue:

website/views/static/login.jade
```diff
@@ -0,0 +1,83 @@
...
+          .panel-group#forgot-password
+            .panel.panel-default
+              .panel-heading
+                h3.panel-title
+                  a(data-toggle="collapse" data-parent="#forgot-password" href="#forgot-password-panel")=env.t('forgotPass')
+
+              .panel-body.collapse#forgot-password-panel
+                form(name='passwordResetForm', ng-submit='passwordReset(passwordResetEmail)', novalidate)
+                  //.alert.alert-success {.success.passwordReset}
```
khipkin: Remove.

website/views/static/login.jade
librarianmage: It was in the code I found for the modal

audiolion:
MathWhiz the button text could be `Trouble Logging In or Registering?` vs. `Having Trouble Logging In / Registering?`

librarianmage:
@audiolion It could also be along the lines of 'Login / Registration Form Not Working?'

audiolion:
@MathWhiz for sure that seems good, I was thinking about whether a '/' was too casual, maybe substitute `/` -> `or` but I like that

khipkin:
@audiolion That last thing is a good point.  Does `/` mean the same thing in all languages?  Because whatever text you choose needs to be translatable.

ghost:
If I were to reverse-translate it from my language, I'd just say something like "Open the form in a separate page".

librarianmage:
That would make more sense

website/views/static/login-modal.jade
```diff
@@ -21,51 +21,62 @@ mixin registrationForm
...
+          //span.help-inline {.errors.passwordReset}
           .form-group
-            input.btn.btn-block.btn-lg.btn-primary(type='submit', ng-disabled='loginForm.$invalid', value=env.t('login'))
+            input.btn.btn-default(type='submit', value=env.t('submit'))
 
-          small=env.t('passMan')
-          br
-          br
+mixin loginModal
+  script(id='modals/login.html', type='text/ng-template')
```
khipkin: This should not go inside the mixin.  Instead, use the mixin inside the script tag:

```
script(id='modals/login.html', type='text/ng-template')
    +loginModal
```

Hkly:
If someone was having trouble logging in or registering, would they understand/realize that opening the form in a new page would help?

ghost:
@Hkly may be a good point. I believe it depends on the kind of issues that are "fixed" when opening a dedicated login page and a particular user.

IMHO experienced users should understand that they have troubles logging in because some (mobile) browsers can have troubles rendering the pop-over login modal. However newbies might get confused and in this case the short links can be misleading:
- with the original version `Having Trouble Logging In / Registering?` or the later one `Login or Registration Form Not Working?` - I can imagine users trying this link e.g. when they get an error about incorrect password - and it won't help
- with the current version `Open Form in Separate Page` - I'm not sure, but once again I can imagine that for someone it can really be not quite clear why would they want to do it (if they haven't experienced similar issues before).

In this regard maybe consider something even more detailed like `Form not working or displayed correctly? Try using the dedicated page` (I'm not suggesting exact wording, just a concept)

Alys:
I think spelling it out very clearly as @GitHubSphinx suggests would be best.

librarianmage:
Ping to reviewers

MathWhiz:
Pinging

Ping to reviewers!

librarianmage:
Ping to reviewers!

crookedneighbor:
Thanks for your work on this, but I don't think we should merge this. I think this should be part of the redesign effort.

librarianmage:
I understand 😔

PR #7124, opened 2016-04-28
V3 payments 7 stripe

piousbox:


test/api/v3/unit/models/group.test.js
```diff
@@ -48,6 +49,10 @@ describe('Group Model', () => {
         ]);
       });
 
+      afterEach(() => {
+        fakeEmail.restore();
```
crookedneighbor: If you're using sandbox, it gets restored after each run automatically.

piousbox:
This error is the reason I include setupNconf() in chat.js and auth.js controllers: https://travis-ci.org/HabitRPG/habitrpg/builds/126268571#L2980

paglias:
have you tried using require instead of import and putting setupConf in the test file? If it doesn't work either add it and we'll figure out how to fix it later

website/src/controllers/top-level/payments/stripe.js
```diff
@@ -1,137 +1,163 @@
...
+        user,
+        customerId: response.id,
+        paymentMethod: 'Stripe',
+        sub,
+      });
+    } else {
+      let method = 'buyGems';
       let data = {
         user,
-        nextBill: results.get_cus.subscription.current_period_end * 1000, // timestamp is in seconds
```
paglias: why was this removed and customerId added?

paglias, line 100: sorry it's okay

website/src/middlewares/api-v3/auth.js
```diff
@@ -55,3 +55,25 @@ export function authWithSession (req, res, next) {
...
+ **/
+export function authWithUrl (req, res, next) {
+  let userId = req.query._id;
+  let apiToken = req.query.apiToken;
+
+  if (!userId || !apiToken) {
+    throw new NotAuthorized(res.t('missingAuthParams'));
+  }
+
+  User.findOne({ _id: req.query._id, apiToken }).exec()
```
paglias: _id: userId

website/src/middlewares/api-v3/auth.js
```diff
@@ -55,3 +55,25 @@ export function authWithSession (req, res, next) {
   })
   .catch(next);
 }
+
+/**
+ * @apiParam {string} _id       The user _id in query
```
paglias: docs like these only works for routes, not for middlewares

piousbox:
I tried `require('../../libs/api-v3/setupNconf')();` and putting it in the test file, didn't work. The error happens before the test file is loaded. I'll add the `setupNconf()` call to those two files again and we'll figure it out later.

PR #7126, opened 2016-04-28
Updated quest service to use new api-v3

TheHollidayInn:
Fixes put_issue_url_here
### Changes

Updated the Quest service to use api-v3

---

UUID:

### DIFF
--- a/test/spec/services/questServicesSpec.js
+++ b/test/spec/services/questServicesSpec.js
@@ -1,7 +1,7 @@
 'use strict';
 
 describe('Quests Service', function() {
-  var groupsService, quest, questsService, user, content, resolveSpy, rejectSpy;
+  var groupsService, quest, questsService, user, content, resolveSpy, rejectSpy, state;
 
   beforeEach(function() {
     user = specHelper.newUser();
@@ -16,10 +16,11 @@ describe('Quests Service', function() {
       $provide.value('User', {sync: sinon.stub(), user: user});
     });
 
-    inject(function(Quests, Groups, Content) {
+    inject(function(Quests, Groups, Content, _$state_) {
       questsService = Quests;
       groupsService = Groups;
       content = Content;
+      state = _$state_;
     });
 
     sandbox.stub(groupsService, 'inviteOrStartParty');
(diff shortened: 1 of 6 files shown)


test/spec/services/questServicesSpec.js
```diff
@@ -335,13 +336,36 @@ describe('Quests Service', function() {
...
 
     it('returns a promise', function() {
-      var promise = questsService.initQuest('whale');
+      var promise = questsService.initQuest(key);
       expect(promise).to.respondTo('then');
     });
 
-    it('accepts quest');
+    it('accepts quest', function(done) {
+      fakeBackend.expectPOST( '/api/v3/groups/party-id/quests/invite/' + key);
```
paglias, line 359: this starts a quest, doesn't accept it

website/public/js/services/groupServices.js
```diff
@@ -94,10 +94,10 @@ angular.module('habitrpg')
       });
     };
 
-    Group.startQuest = function(gid) {
+    Group.startQuest = function(gid, key) {
       return $http({
         method: "POST",
-        url: groupApiURLPrefix + '/' + gid + '/questAccept',
+        url: groupApiURLPrefix + '/' + gid + '/quests/invite/' + key,
```
paglias, line 100: hmm not sure this is correct. If it's only used to invite to a quest it should be called Group.inviteToQuest

website/public/js/services/questServices.js
```diff
@@ -106,20 +97,20 @@
...
-        party.$startQuest({key:key}, function(){
-          party.$syncParty();
-          $state.go('options.social.party');
-          resolve();
-        });
+        Analytics.track({'hitType':'event', 'eventCategory':'behavior', 'eventAction':'quest', 'owner':true, 'response':'accept', 'questName': key});
+        Analytics.updateUser({'partyID': party._id, 'partySize': party.memberCount});
+        Groups.Group.startQuest(party._id, key)
+          .then(function() {
+            Groups.Group.syncParty();
```
paglias: instead of syncing, can't you use the return data from this route? Most quest routes return party.quest so you can just substitute the cached party.quest object with the one returned by the route call.

website/public/js/services/groupServices.js
TheHollidayInn, line 100: Yea, the invite route used to be in the accept route. So, I didn't realize that for a while and did not update the function names.

TheHollidayInn:
This is updated!

PR #7141, opened 2016-05-03
Turned on client side tests. Fixed broken tests

TheHollidayInn:
Fixes put_issue_url_here
### Changes

Add client tests to npm run
Fixed broken client tests. 

---

UUID:


website/public/js/services/taskServices.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 (function(){
-  var TASK_KEYS_TO_REMOVE = ['_id', 'completed', 'date', 'dateCompleted', 'dateCreated', 'history', 'id', 'streak'];
+  var TASK_KEYS_TO_REMOVE = ['_id', 'completed', 'date', 'dateCompleted', 'dateCreated', 'history', 'id', 'streak', 'createdAt'];
```
paglias: dateCreated is not stored anymore in v3, it now uses createdAt

crookedneighbor:
Why do you need setTimeout in the tests? Does the save call not provide a callback or a promise to show when it's done? If it doesn't, why not?

website/public/js/controllers/partyCtrl.js
```diff
@@ -46,7 +46,9 @@ habitrpg.controller("PartyCtrl", ['$rootScope','$scope','Groups','Chat','User','
         }
       }
 
-      Chat.markChatSeen($scope.group._id);
+      if ($scope.group) {
```
crookedneighbor: About time :)

TheHollidayInn:
Well, this is the thing I have been confused about.

For example: The createChallenge function is a promise. But, in the controller tests, I don't just need to test after the callback is returned, I need to test after the callback returns then runs some actions - like User.sync. 

So, these seem to be examples of background calls, so I need to wait for the background calls to finish.

website/public/js/controllers/partyCtrl.js
TheHollidayInn: lol, it seemed odd to me as well.

website/public/js/services/taskServices.js
TheHollidayInn: Okay, I changed createAt in the tests. I wasn't sure if I should remove it from this array for any v2 support.

TheHollidayInn:
Updated!

website/public/js/services/taskServices.js
paglias: Nah, you can remove it

TheHollidayInn:
Updated!

PR #7145, opened 2016-05-05
[WIP] Ported User Serivce to client side and to api v3

TheHollidayInn:
Fixes put_issue_url_here
### Changes

This is still a work in progress. I changed the flow of the User Service quite a bit, so I hope that is okay. I think this flow can simply the readability. I want to remove the wrapping as well. It is only there for the task and tag routes at the moment. 
- Moved User Service to public
- Update User service with API v3 routes
- Update most op functions

---

UUID:


common/script/ops/clearPMs.js
```diff
@@ -1,6 +1,6 @@
 module.exports = function clearPMs (user) {
   user.inbox.messages = {};
-  user.markModified('inbox.messages');
+  // user.markModified('inbox.messages');
```
paglias: why?

website/public/js/controllers/rootCtrl.js
```diff
@@ -21,7 +21,8 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
         if (!!fromState.name) Analytics.track({'hitType':'pageview','eventCategory':'navigation','eventAction':'navigate','page':'/#/'+toState.name});
         // clear inbox when entering or exiting inbox tab
         if (fromState.name=='options.social.inbox' || toState.name=='options.social.inbox') {
-          User.user.ops.update && User.set({'inbox.newMessages':0});
+          //@TODO: Protected path. We need a url
+          User.set({'inbox.newMessages': 0});
```
paglias: for this a route might be the best choice. Otherwise we can unprotect inbox.newMessages

website/public/js/services/userServices.js
```diff
@@ -0,0 +1,416 @@
...
+
+      user._wrapped = false;
+
+      function sync() {
+        $http({
+          method: "GET",
+          url: 'api/v3/user/',
+        })
+        .then(function (response) {
+          Notification.text(response.data.message);
```
paglias: if (response.data.message) ?

website/public/js/services/userServices.js
```diff
@@ -0,0 +1,416 @@
...
+        var queryString = '';
+        if (opData.query) queryString = '?' + $.param(opData.query)
+
+        $http({
+          method: method,
+          url: url + queryString,
+          body: body,
+        })
+        .then(function (response) {
+          Notification.text(response.data.message);
```
paglias: if (response.data.message) ?

website/public/js/services/userServices.js
```diff
@@ -0,0 +1,416 @@
...
+          save();
+        })
+      }
+
+      function setUser(updates) {
+        for (var key in updates) {
+          user[key] = updates[key];
+        }
+
+        sync();
```
paglias: why sync() here? the changes aren't applied directly to the client data? it should not be necessary to get the user from the sever

website/public/js/services/userServices.js
```diff
@@ -0,0 +1,416 @@
...
+        },
+
+        hourglassPurchase: function (data) {
+          var type = data.params.type;
+          var key = data.params.key;
+          callOpsFunctionAndRequest('hourglassPurchase', 'purchase-hourglass', "POST", type + '/' + key, data);
+        },
+
+        unlock: function (data) {
+          $window.habitrpgShared.ops['unlock'](user, data);
```
paglias: this will call ops.unlock twice?

website/public/js/services/userServices.js
```diff
@@ -0,0 +1,416 @@
...
+          } else {
+            settings.sync.queue.push(action);
+          }
+
+          save();
+          // syncQueue(cb);
+        },
+
+        sync: function(){
+          user._v--;
```
paglias: the v-- was necessary for batchUpdate, in v3 it's not used anymore

website/public/js/services/userServices.js
```diff
@@ -0,0 +1,416 @@
...
+      //create and load if not
+      } else {
+        localStorage.setItem(STORAGE_SETTINGS_ID, JSON.stringify(defaultSettings));
+        _.extend(settings, defaultSettings);
+      }
+
+      //If user does not have ApiID that forward him to settings.
+      if (!settings.auth.apiId || !settings.auth.apiToken) {
+
+        if (MOBILE_APP) {
```
paglias: this IF can be removed, it was used for the old mobile app based on phonegap/ionic/angular

common/script/ops/clearPMs.js
TheHollidayInn: markModified is undefined on the client, so this line breaks the script.

paglias: for that case I added https://github.com/HabitRPG/habitrpg/blob/api-v3/common/script/index.js#L236-L239

TheHollidayInn: Hmm it was still breaking, but I was doing some changes with the wrap function. Let me test it again.

paglias: it's possible that the function is called before the user is wrapped.. but anyway we can't really remove the markModified calls

TheHollidayInn: Sounds good. I will make sure it works then!

TheHollidayInn:
Updated!

website/public/js/services/userServices.js
```diff
@@ -259,12 +266,19 @@ angular.module('habitrpg')
...
 
         set: function(updates) {
           setUser(updates);
+          $http({
+            method: "PUT",
+            url: 'api/v3/user',
+            data: updates,
+          })
+          .then(function (response) {
+            sync();
```
paglias: syncing is not necessary, the changes are already applied client side

website/src/controllers/api-v3/user.js
```diff
@@ -1175,6 +1175,26 @@ api.clearMessages = {
...
 
+/**
+ * @api {post} /api/v3/user/mark-pms-read Marks Private Messages as read
+ * @apiVersion 3.0.0
+ * @apiName markPmsRead
+ * @apiGroup User
+ *
+ * @apiSuccess {object} data user.inbox.messages
+**/
+api.markPmsRead = {
```
paglias, line 1186: a test for this wouldn't hurt

website/public/js/services/userServices.js
```diff
@@ -201,6 +198,16 @@ angular.module('habitrpg')
           })
         },
 
+        clearNewMessages: function () {
+          $http({
+            method: "POST",
+            url: 'api/v3/user/mark-pms-read',
+          })
+          .then(function (response) {
+            sync();
```
paglias: maybe this should be a common op so we don't have to sync the whole user but just use the same code on the client and on the server?

website/public/js/services/userServices.js
TheHollidayInn: Eh, I looked into this, and I believe I needed to sync for some reason, but can't remember. I'll try it again.

website/public/js/services/userServices.js
TheHollidayInn: Yea, I think the add hourglass and 10 gems should be the same, but I was trying to stray away from adding to much. I can definitely do it though.

paglias: I think in this case a common op is worth it, in the case of debug options that'll be used only by devs syncing is okay

website/public/js/services/userServices.js
TheHollidayInn: So, it looks like the for user loop doesn't work. You can't set a object's sub object properties with string notation without a work around. 

I can add this workaround: http://stackoverflow.com/questions/6491463/accessing-nested-javascript-objects-with-string-key .

Or, I can update all the User.set calls to actually set the user object before calling User.set.

For example:

```
$scope.addGold = function(){
      User.user.stats.gp = User.user.stats.gp + 500;
      User.set({
        'stats.gp': User.user.stats.gp + 500,
      });
    };
```

website/public/js/controllers/authCtrl.js
```diff
@@ -29,7 +29,7 @@ angular.module('habitrpg')
         if (status === 0) {
           $window.alert(window.env.t('noReachServer'));
         } else if (!!data && !!data.err) {
-          $window.alert(data.err);
+          $window.alert(data.data.message);
```
paglias: Errors return {error, success, message} as response, data is only used for successful responses ( {data, success} )

website/public/js/controllers/authCtrl.js
```diff
@@ -98,9 +99,9 @@ angular.module('habitrpg')
...
       $scope.socialLogin = function(network){
         hello(network).login({scope:'email'}).then(function(auth){
-          $http.post(ApiUrl.get() + "/api/v2/user/auth/social", auth)
+          $http.post(ApiUrl.get() + "/api/v3/user/auth/social", auth)
             .success(function(data, status, headers, config) {
-              runAuth(data.id, data.token);
+              runAuth(data.data.id, data.data.apiToken);
             }).error(errorAlert);
         }, function( e ){
           alert("Signin error: " + e.error.message );
```
paglias: e.message instead of e.error.message

website/public/js/controllers/authCtrl.js
TheHollidayInn: This is an $http standard right?

TheHollidayInn: Nevermind

website/public/js/controllers/authCtrl.js
```diff
@@ -29,7 +29,7 @@ angular.module('habitrpg')
         if (status === 0) {
           $window.alert(window.env.t('noReachServer'));
         } else if (!!data && !!data.err) {
```
crookedneighbor: I know you didn't write this, but this should be `(!(data && data.err))`

website/public/js/controllers/authCtrl.js
```diff
@@ -46,10 +46,10 @@ angular.module('habitrpg')
 
         $scope.registrationInProgress = true;
 
-        var url = ApiUrl.get() + "/api/v2/register";
+        var url = ApiUrl.get() + "/api/v3/user/auth/local/register";
         if($rootScope.selectedLanguage) url = url + '?lang=' + $rootScope.selectedLanguage.code;
         $http.post(url, scope.registerVals).success(function(data, status, headers, config) {
-          runAuth(data.id, data.apiToken);
+          runAuth(data.data.id, data.data.apiToken);
```
crookedneighbor: Every time I see `data.data`, I die a little inside.

crookedneighbor: When we do the client rewrite, we'll want these to be `res` or something so it becomes `res.data`

TheHollidayInn: Couldn't agree more. I have change this to response.data in many places. It depends on the rewrite, but it definitely needs to be addressed.

PR #7154, opened 2016-05-08
Corrected reset checklist logic

TheHollidayInn:
This fixed the Carryover at Cron bug reported in slack
### Changes

---

UUID:

### DIFF
--- a/website/src/libs/api-v3/cron.js
+++ b/website/src/libs/api-v3/cron.js
@@ -181,7 +181,7 @@ export function cron (options = {}) {
     });
     task.completed = false;
 
-    if (completed || scheduleMisses > 0) {
+    if (!completed && scheduleMisses === 0) {
       task.checklist.forEach(i => i.completed = true); // FIXME this should not happen for grey tasks unless they are completed
     }
   });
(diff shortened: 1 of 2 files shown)


website/src/libs/api-v3/cron.js
```diff
@@ -181,7 +181,7 @@ export function cron (options = {}) {
     });
     task.completed = false;
 
-    if (completed || scheduleMisses > 0) {
+    if (!completed && scheduleMisses === 0) {
```
paglias: I think the problem here is that you're marking the items as completed while instead they should be un-marked. See how it works in v2 https://github.com/HabitRPG/habitrpg/blob/develop/common/script/fns/cron.js#L276-L281

TheHollidayInn: Hmm. That makes sense. Yea, that makes more sense. I was trying to find this version  of the cron.

website/src/libs/api-v3/cron.js
```diff
@@ -181,8 +181,8 @@ export function cron (options = {}) {
     });
     task.completed = false;
 
-    if (completed || scheduleMisses > 0) {
-      task.checklist.forEach(i => i.completed = true); // FIXME this should not happen for grey tasks unless they are completed
+    if (completed || (scheduleMisses > 0)) {
```
paglias: eslint complains about the extra parenthesis here

PR #7184, opened 2016-05-11
change opacity of emoji in completed tasks

### LINKED ISSUE #2377: Change .completed color to opacity to support emoji?
I'm working on my first commit, but it wasn't as straightforward as I thought so I wanted to run it by people for discussion. Once we started adding emoji to tasks, it bugged me that they were still brightly colored even if the task was completed or disabled for the day. So I thought I'd reduce the opacity on the emoji accordingly.

However, it looks like Chrome and IE create `<span class="emoji">` but Firefox and Safari don't set the emoji off in any way.

So, are there any objections if instead I set the color of `.task-text` back to #333 whether completed or not, and then for `.completed` set both the emoji and the text to `opacity: .4`? (.4 gets the text pretty darn close to #999). Any disadvantages to this? It means older versions of IE will be using a filter. Do we care about older versions of IE? 

If people don't like this solution, then I can just set the opacity on the emoji span and Chrome and IE get this feature but Firefox and Safari don't, too bad for them.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1386046-change-completed-color-to-opacity-to-support-emoji?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

librarianmage:
Fixes #2377
### Changes

image

---


crookedneighbor:

PR #7198, opened 2016-05-12
fix: Only apply captureStackTrace if it exists on the error object

crookedneighbor:
Fixes js error in non-Chrome browsers.


crookedneighbor:
Feel free to merge when the build passes (if I don't do it first).

PR #7215, opened 2016-05-14
[API v3] Correct checklist on v3 client

crookedneighbor:
Linting is failing for me locally. It's like the babel eslint plugins aren't being applied :(


crookedneighbor:
Confirmed that it's a regression in eslint. Pinned to 2.9.

https://github.com/eslint/eslint/issues/6158

PR #7228, opened 2016-05-14
[API v3] allow players to send gems to each other

Alys:
Fixes https://github.com/HabitRPG/habitrpg/issues/7227
### Changes
- finishes the `/members/transfer-gems` route (for transferring gems from one player's balance to another)
- makes the English locale string for `toUserIDRequired` more user-friendly -- **BUT** maybe it's meant to have the `toUserId` parameter name in it, if it's intended to be seen by only developers? If so, we should do something to stop `toUserId` being incorrectly modified by translators.
- renames locale key `notEnoughGemsToSend` to `badAmountOfGemsToSend` to more accurately reflect the text
- edits the text for `badAmountOfGemsToSend` (must be at least 1 not 0)


Alys:
Ooops. Fixing tests....

Alys:
NOTE: The build is still running, and if it fails, I might need to change the code again, so don't bother reviewing this yet. (Sorry, test:karma's not working locally; I'll look into that soon.)

I've removed `giftType` from the tests because there's no support for it in the code, however the code hadn't been completed before I did this PR, so maybe I completed the code in a different way than it was intended to be done. If you want it fixed differently, tell me and I'll do it.

Possibly giving gems and subscriptions was meant to be done by the same route, so that a giftType would then be needed??

Also, in the tests, do you want `userToSendMessage` to be renamed to `userToSendGems` ? It looks like `userToSendMessage` was copied from `POST-send_private_message.test.js`

website/client/js/controllers/memberModalCtrl.js
```diff
@@ -39,7 +39,7 @@ habitrpg
       };
 
       $scope.sendGift = function (uuid, gift) {
```
crookedneighbor, line 41: Looks like the gift variable is never used.

crookedneighbor:
I'm making a few adjustments and then I'll merge it in.

PR #7245, opened 2016-05-14
Upstream/api v3 fixes 2

TheHollidayInn:
7194
### Changes

Fixed:

(the one about the timezone is needed)
Changed site language doesn't reload page
Tags selection is broken (adding / removing tags from tasks)
Battle monsters with friends button (when not in a party or partying solo) doesn't work
load completed todos if Completed tab is clicked

---

UUID:

### DIFF
--- a/test/spec/controllers/inviteToGroupCtrlSpec.js
+++ b/test/spec/controllers/inviteToGroupCtrlSpec.js
@@ -1,7 +1,7 @@
 'use strict';
 
 describe('Invite to Group Controller', function() {
-  var scope, ctrl, groups, user, guild, $rootScope;
+  var scope, ctrl, groups, user, guild, rootScope, $controller;
 
   beforeEach(function() {
     user = specHelper.newUser({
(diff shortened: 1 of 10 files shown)


test/spec/controllers/inviteToGroupCtrlSpec.js
```diff
@@ -93,6 +97,10 @@ describe('Invite to Group Controller', function() {
     });
 
     context('email', function() {
+      beforeEach(function () {
+        sinon.stub(rootScope, 'hardRedirect');
```
crookedneighbor, line 101: Shouldn't this be sandbox.stub?

test/spec/controllers/inviteToGroupCtrlSpec.js
```diff
@@ -166,6 +174,10 @@ describe('Invite to Group Controller', function() {
     });
 
     context('uuid', function() {
+      beforeEach(function () {
+        sinon.stub(rootScope, 'hardRedirect');
```
crookedneighbor, line 178: Same here. Shouldn't this be sandbox.stub?

website/client/js/controllers/rootCtrl.js
```diff
@@ -293,11 +293,11 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
       User.save();
 
       var spell = $scope.spell;
-      var targetId = (type == 'party' || type == 'self') ? '' : type == 'task' ? target.id : target._id;
+      var targetId = target._id;
       $scope.spell = null;
       $rootScope.applyingAction = false;
 
-      $http.post(ApiUrl.get() + '/api/v2/user/class/cast/'+spell.key+'?targetType='+type+'&targetId='+targetId)
```
crookedneighbor, line 300: Does this indicate a problem with v2?

test/spec/controllers/inviteToGroupCtrlSpec.js
crookedneighbor, line 101: I guess it doesn't matter that much since rootScope gets overwritten on every test.

crookedneighbor:
Merged in.

PR #7263, opened 2016-05-15
Debug update user

crookedneighbor:
Added a debug route to set any property on a user object.
- Fixes +{1,2} Missed day buttons
- Adds a Make Admin button

### DIFF
--- a/website/client/js/services/userServices.js
+++ b/website/client/js/services/userServices.js
@@ -244,6 +244,21 @@ angular.module('habitrpg')
           })
         },
 
+        setCron: function (numberOfDays) {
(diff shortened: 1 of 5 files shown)


website/client/js/services/userServices.js
```diff
@@ -244,6 +244,35 @@ angular.module('habitrpg')
           })
         },
 
+        setCron: function (numberOfDays) {
```
crookedneighbor, line 247: It's a little weird that these methods are just on the User object (following the pattern we had in place before for addTenGems, etc). When we rewrite the client, these should be isolated so no module other than the debug footer can use it.

PR #7264, opened 2016-05-15
remove userServices.js from karma.conf

Alys:
### Changes

removes userServices.js from karma.conf to prevent this warning:

```
> karma start --single-run

14 05 2016 23:12:48.766:WARN [watcher]: Pattern
"..../common/script/public/userServices.js" does not match any file.
```

I believe userServices.js was moved to website/client/js/services/userServices.js and so it's covered in karma.conf by `website/client/js/services/**/*.js`


crookedneighbor:
Yep, you're right!

PR #7286, opened 2016-05-15
[API v3] add CRON_SAFE_MODE

Alys:
Partial work for https://github.com/HabitRPG/habitrpg/issues/7161
### Changes
- Adds "safe mode" for cron.
- Requires a **new environment variable**: `CRON_SAFE_MODE` (set to "false" normally; "true" when we want safe mode).
- Removes an unrelated TODO that's already been done.
- Removes duplicated code (this was causing twice as much mana as appropriate to be gained overnight in API v3 only, so if anyone reports that as a bug, this fixes it).

I suggest reviewing with `w=1` because a bunch of unchanged code is now wrapped in an `else`.

The quest damage and chat message portions still need to be tested (can't start a quest due to an unrelated bug).

I'm writing automated tests for this.

### DIFF
--- a/website/server/libs/api-v3/cron.js
+++ b/website/server/libs/api-v3/cron.js
@@ -152,33 +152,37 @@ export function cron (options = {}) {
         }
       }
 
-      if (scheduleMisses > EvadeTask && !CRON_SAFE_MODE) {
+      if (scheduleMisses > EvadeTask) {
         // The user did not complete this due Daily (but no penalty if cron is running in safe mode).
-        perfect = false;
-
-        if (task.checklist && task.checklist.length > 0) { // Partially completed checklists dock fewer mana points
-          let fractionChecked = _.reduce(task.checklist, (m, i) => m + (i.completed ? 1 : 0), 0) / task.checklist.length;
-          dailyDueUnchecked += 1 - fractionChecked;
-          dailyChecked += fractionChecked;
+        if (CRON_SAFE_MODE) {
+          dailyChecked += 1; // allows full allotment of mp to be gained
         } else {
-          dailyDueUnchecked += 1;
+          perfect = false;
+
+          if (task.checklist && task.checklist.length > 0) { // Partially completed checklists dock fewer mana points
+            let fractionChecked = _.reduce(task.checklist, (m, i) => m + (i.completed ? 1 : 0), 0) / task.checklist.length;
+            dailyDueUnchecked += 1 - fractionChecked;
+            dailyChecked += fractionChecked;
+          } else {
+            dailyDueUnchecked += 1;
+          }
+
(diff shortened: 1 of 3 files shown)


website/server/models/group.js
```diff
@@ -515,8 +515,10 @@ schema.statics.bossQuest = async function bossQuest (user, progress) {
   let down = progress.down * quest.boss.str; // multiply by boss strength
 
   group.quest.progress.hp -= progress.up;
-  // TODO Create a party preferred language option so emits like this can be localized
-  group.sendChat(`\`${user.profile.name} attacks ${quest.boss.name('en')} for ${progress.up.toFixed(1)} damage.\` \`${quest.boss.name('en')} attacks party for ${Math.abs(down).toFixed(1)} damage.\``);
+  // TODO Create a party preferred language option so emits like this can be localized. Suggestion: Always display the English version too. Or, if English is not displayed to the players, at least include it in a new field in the chat object that's visible in the database - essential for admins when troubleshooting quests!
+  let playerAttack = `${user.profile.name} attacks ${quest.boss.name('en')} for ${progress.up.toFixed(1)} damage.`;
+  let bossAttack = nconf.get('CRON_SAFE_MODE') ? `${quest.boss.name('en')} did not attack the party because it was asleep while maintenance was happening.` : `${quest.boss.name('en')} attacks party for ${Math.abs(down).toFixed(1)} damage.`;
```
paglias: nconf.get('CRON_SAFE_MODE') === 'true'

website/server/libs/api-v3/cron.js
```diff
@@ -231,14 +239,9 @@ export function cron (options = {}) {
...
   // Adjust for fraction of dailies completed
-  user.stats.mp += _.max([10, 0.1 * user._statsComputed.maxMP]) * dailyChecked / (dailyDueUnchecked + dailyChecked);
-  if (user.stats.mp > user._statsComputed.maxMP) user.stats.mp = user._statsComputed.maxMP;
-
   if (dailyDueUnchecked === 0 && dailyChecked === 0) dailyChecked = 1;
   user.stats.mp += _.max([10, 0.1 * user._statsComputed.maxMP]) * dailyChecked / (dailyDueUnchecked + dailyChecked);
-  if (user.stats.mp > user._statsComputed.maxMP) {
-    user.stats.mp = user._statsComputed.maxMP;
-  }
+  if (user.stats.mp > user._statsComputed.maxMP) user.stats.mp = user._statsComputed.maxMP;
```
paglias, line 244: 👍

paglias:
Looking good!

website/server/models/group.js
Alys: Ugh, good point, thanks. Fixed. This is ready for review again.

PR #7293, opened 2016-05-16
[API v3] fix bug in a test for Rebirth, add new tests, adjust apidocs

Alys:
### Changes
- Fix minor bug in Rebirth test (changed to.be.emtpy to to.be.empty) - that typo didn't seem to have any effect on the success of the test though in previous builds (see below from previous build https://travis-ci.org/HabitRPG/habitrpg/builds/130502946 line 3393) - maybe we should look into that?
- Add more tests for Rebirth
- Adjust apidocs for Rebirth to remove typos and use normal terminology ("rebirth" not "reset")

```
  shared.ops.rebirth
    ✓ returns an error when user balance is too low and user is less than max level
    ✓ rebirths a user with enough gems
    ✓ rebirths a user with not enough gems but max level
    ✓ resets user's taks values except for rewards to 0
    ✓ resets user's daily streaks to 0
    ✓ resets a user's buffs
    ✓ resets a user's health points
    ✓ resets a user's class
    ✓ resets a user's stats
    ✓ resets a user's gear
    ✓ resets a user's gear owned
    ✓ resets a user's current pet
    ✓ resets a user's current mount
    ✓ resets a user's flags
    ✓ does not reset rebirthEnabled if user has beastMaster
    ✓ sets rebirth achievement
    ✓ increments rebirth achievemnts
```

### DIFF
--- a/test/common/ops/rebirth.js
+++ b/test/common/ops/rebirth.js
@@ -3,7 +3,9 @@ import i18n from '../../../common/script/i18n';
 import { MAX_LEVEL } from '../../../common/script/constants';
 import {
   generateUser,
+  generateHabit,
   generateDaily,
+  generateTodo,
   generateReward,
 } from '../../helpers/common.helper';
 import {
@@ -19,7 +21,7 @@ describe('shared.ops.rebirth', () => {
   beforeEach(() => {
     user = generateUser();
     user.balance = 2;
-    tasks = [generateDaily(), generateReward()];
+    tasks = [generateHabit(), generateDaily(), generateTodo(), generateReward()];
   });
 
   it('returns an error when user balance is too low and user is less than max level', (done) => {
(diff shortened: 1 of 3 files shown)


Alys:
Renamed to add [API v3] to title

test/common/ops/rebirth.js
```diff
@@ -156,7 +171,7 @@ describe('shared.ops.rebirth', () => {
     expect(user.flags.dropsEnabled).to.be.false;
     expect(user.flags.classSelected).to.be.false;
     expect(user.flags.rebirthEnabled).to.be.false;
-    expect(user.flags.levelDrops).to.be.emtpy;
+    expect(user.flags.levelDrops).to.be.empty;
```
paglias, line 174: good catch! probably chai do not throw when using a wrong type

crookedneighbor, line 174: The problem is that chai can't detect when you misspell a property. That's one advantage that expect.js has over chai, all assertions are functions so it throws an error if the function doesn't exist, so it's very obvious when you make a mistake. 

With Node 6, we could wrap chai in a Proxy and create a getter to throw an error if the property is undefined. :/

I'm looking into solutions for this.

Issue #7280, opened 2016-05-15
[API v3] Can't load user info by links to static/front/#?memberId=


aviskase:
General Info
- UUID: 266104ca-889e-4a7a-956d-3bd9a94244d3
- Browser: Google Chrome 50
- OS: Ubuntu Mate 16.04
### Description

This type of links can be generated when adding a message as a task. 

For example, I've added a message as a task. By default, there is a link to the user in the notes.

![0515_59f](https://cloud.githubusercontent.com/assets/3819473/15275663/7b8fa4d2-1af3-11e6-9dfc-951c22a2e4d9.png)

After clicking on it a new page will open with no info in user pop-up: `https://v3.habitica.com/static/front/#?memberId=bc10ec41-4213-4482-bbb8-558e3c3ea8d5`

![0515_70a](https://cloud.githubusercontent.com/assets/3819473/15275676/b93303f6-1af3-11e6-8b7f-a9ace9bc3b6a.png)
#### Console Errors

```
Failed to load resource: the server responded with a status of 401 (Unauthorized)
angular.js:11594 TypeError: Cannot read property 'stats' of undefined
    at Object.expDisplay (statServices.js:57)
    at $parseFunctionCall (angular.js:12332)
    at oneTimeInterceptedExpression (angular.js:12836)
    at Object.oneTimeWatch (angular.js:12771)
    at Scope.$digest (angular.js:14218)
    at Scope.$apply (angular.js:14489)
    at done (angular.js:9646)
    at completeRequest (angular.js:9836)
    at XMLHttpRequest.requestLoaded (angular.js:9777)(anonymous function) @ angular.js:11594
angular.js:11594 TypeError: Cannot read property 'weapon' of undefined
    at statServices.js:45
    at tt (habitrpg-shared.js:12)
    at Function.<anonymous> (habitrpg-shared.js:13)
    at rr (habitrpg-shared.js:13)
    at _.$a (habitrpg-shared.js:13)
    at Object.equipmentStatBonus (statServices.js:51)
    at $parseFunctionCall (angular.js:12332)
    at extend.> (angular.js:11729)
    at $parseBinaryFn (angular.js:12058)
    at Object.oneTimeWatch (angular.js:12771)(anonymous function) @ angular.js:11594
angular.js:11594 TypeError: Cannot read property 'fns' of undefined
    at Object.classBonus (statServices.js:25)
    at $parseFunctionCall (angular.js:12332)
[...]

PR #7299, opened 2016-05-16
Fixed member modal on front page

TheHollidayInn:
#7280
### Changes

Updated static.js to use Member service
Made getMember route public

---

UUID:

### DIFF
--- a/website/server/controllers/api-v3/members.js
+++ b/website/server/controllers/api-v3/members.js
@@ -33,7 +33,7 @@ let api = {};
 api.getMember = {
   method: 'GET',
   url: '/members/:memberId',
-  middlewares: [authWithHeaders()],
+  middlewares: [],
   async handler (req, res) {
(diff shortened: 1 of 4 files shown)


TheHollidayInn:
@paglias I made the getMember route public. I think it was like that before, but I'm unsure if that is okay.

website/server/controllers/api-v3/members.js
```diff
@@ -33,7 +33,7 @@ let api = {};
 api.getMember = {
   method: 'GET',
   url: '/members/:memberId',
-  middlewares: [authWithHeaders()],
+  middlewares: [],
```
paglias, line 36: @crookedneighbor what do you think?

crookedneighbor, line 36: This is fine. It's what is supposed to be used for the front page when sharing the member modal.

PR #7321, opened 2016-05-17
[API v3] adjust locales strings for Rebirth, Reset, and Fortify / ReRoll

Alys:
### Changes

Adjusts locales strings only (no other code changes):
- adjusts text for Rebirth, Reset, and Fortify / ReRoll to describe new behaviour for challenge tasks
- changes key names containing `reroll` to contain `fortify` for consistency with existing strings
- moves those `fortify` strings to the same locales file as existing strings
- moves a new `reset` string to the same locales file as old `reset` strings

The keys that I changed were not used in v2.

My thinking is that it will be easier for coders and translators to work with the strings if they have consistent names and files. However please tell me if you want me to revert any of this.

If you don't mind what I've done here, can I do similar things to at least some of the other strings in `api-v3.json`? I think it would help us in the future if the strings were in files related to their topics.


paglias:
It would be great if you can do it, I had meant to ask
Il 17 mag 2016 10:11 AM, "Alys" notifications@github.com ha scritto:

Alys:
Fantastic! Will do. :)

PR #7337, opened 2016-05-17
Legacy Id for v2

crookedneighbor:
First pass at supporting _legacyId in v2 routes.

If we want to keep `_legacyId` as `legacyId`, we should remove it from the object when `toJSON` is called.


crookedneighbor:
From @paglias 

> I think they should also check for legacyId if id doesn’t match anything
> most v2 ids where valid uuids but not unique

paglias:
Looking good to me if you want to merge 👍

PR #7354, opened 2016-05-18
feat: Add v3 strings to develop for translating

crookedneighbor:
@paglias


Alys:
Oh... one thing: I made some changes to the text for new v3 features, for example (new text in bold):
-    "rebirthAdvList3": "Your Habits, Dailies, and To-Dos reset to yellow, and streaks reset, **except for challenge tasks**.",

If this ends up getting deployed to production significantly before AP v3 goes live, we'll need to make some adjustments.

Issue #7229, opened 2016-05-14
[API v3] Chat does not keep updates since last refresh, even when updated or synced


citrusella:
General Info
- UUID: 2d6ef231-50b4-4a22-90e7-45eb97147a2c
- Browser: Mozilla Firefox 46.0.1
- OS: Windows 7 Professional Service Pack 1 (via Bootcamp on a late-2011 Macbook Pro)
### Description

Fetching recent messages updates the chat... but only for as long as you are on it. Going to another page on the site (without a reload) and coming back puts the chat back in the state it was in when the site was last reloaded. Syncing does not bring the chat up to date in regards to this; only a full reload seems to change how recent the messages are without fetching them. I only tested this in the Tavern but would assume it could be happening in other chats, too.

I think it may be related: Chat also isn't updating with new messages right after I post a message. (Or it isn't updating reliably. XP)
#### Console Errors

None to speak of.

paglias:
@TheHollidayInn new group object is not being sinced?

PR #7360, opened 2016-05-18
Added force cache update when fetching new messages

TheHollidayInn:
#7229
### Changes

---

UUID:

### DIFF
--- a/website/client/js/controllers/chatCtrl.js
+++ b/website/client/js/controllers/chatCtrl.js
@@ -114,12 +114,19 @@ habitrpg.controller('ChatCtrl', ['$scope', 'Groups', 'Chat', 'User', '$http', 'A
       });
     };
 
+    function handleGroupResponse (response) {
+      $scope.group = response;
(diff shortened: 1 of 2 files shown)


website/client/js/controllers/chatCtrl.js
```diff
@@ -114,12 +114,19 @@ habitrpg.controller('ChatCtrl', ['$scope', 'Groups', 'Chat', 'User', '$http', 'A
...
+      if (!$scope.group._id) $scope.group = response.data.data;
+    };
+
     $scope.sync = function(group) {
-      //@TODO: We need to use chat service here
-      Groups.Group.get(group._id)
-        .then(function (response) {
-          $scope.group = response.data.data;
-        })
+      if (group.name === Groups.TAVERN_NAME) {
```
crookedneighbor, line 123: Why are we testing against the name instead of the id constant that @paglias created?

https://github.com/HabitRPG/habitrpg/blob/api-v3/website/client/js/app.js#L29

Issue #7291, opened 2016-05-16
[API v3] Custom Day Start time will not set


Memry:
- UUID: e495bd8b-f2f1-48e8-8f5f-9a52710999cf
- Browser: Chrome Version 50.0.2661.102 m
- OS: Windows 7

Tried to set a custom cron.  It returned the error message "path lastCron was not saved" and when I refresh the old custom cron time shows.
![image](https://cloud.githubusercontent.com/assets/7385423/15280408/e4c2a74c-1ae4-11e6-94b9-4a89f5c348ba.png)

Alys:
I'm about to rename this to "Custom Day Start time will not set".

I believe setting lastCron is an important part of this feature as it protects the user from a surprise cron at the time of changing or too soon afterwards.

TheHollidayInn:
@crookedneighbor I think you created the new route `/debug/set-cron`. I think we should move the route to the user controller to fix this issue.

paglias:
@crookedneighbor do you want to use debug/set-cron (moved to user controller) or the old way of updating the user with PUT /user whitelisting lastCron?

crookedneighbor:
I think the real question is do we want to give clients the ability to set lastCron to _any_ date? If so, yeah, I'm fine moving it to the user controller or making it adjustable. 

If we want to control how lastCron is set, I'd make setting a custom day start a separate route.

paglias:
Well in v3 we tried to limit the paths freely updatable but I've been forced to undo many changes like streaks on tasks. The option here is to make lastCron settable or create a new route to change the custom day start

TheHollidayInn:
I think we move the debug route to the user controller. Then it becomes a "setter" and we can add restrictions.

paglias:
I agree with @crookedneighbor , let's create a new route to set custom day start

TheHollidayInn:
@crookedneighbor finished this up. So this should be fixed now.

crookedneighbor:
I don't think this has been deployed to the test site yet.

PR #7363, opened 2016-05-18
Added new route for setting last cron and updated front end

TheHollidayInn:
#7291
### Changes

---

UUID:

### DIFF
--- a/test/spec/controllers/settingsCtrlSpec.js
+++ b/test/spec/controllers/settingsCtrlSpec.js
@@ -17,6 +17,7 @@ describe('Settings Controller', function () {
         releasePets: sandbox.stub(),
         releaseMounts: sandbox.stub(),
         releaseBoth: sandbox.stub(),
+        setLastCron: sandbox.stub(),
         user: user
       };
 
@@ -97,8 +98,8 @@ describe('Settings Controller', function () {
       expect(User.set).to.be.calledOnce;
       expect(User.set).to.be.calledWith({
         'preferences.dayStart': 5,
-        'lastCron': expectedTime
       });
+      expect(User.setLastCron).to.be.calledWith(expectedTime);
     });
   });
 
--- a/website/client/js/services/userServices.js
+++ b/website/client/js/services/userServices.js
@@ -310,6 +310,19 @@ angular.module('habitrpg')
           });
         },
 
+        setLastCron: function (date) {
+          $http({
+            method: "POST",
+            url: 'api/v3/user/set-cron',
+            data: {
+              lastCron: date
+            }
+          })
+          .then(function (response) {
+            Notification.text('Last cron updated. Remember to refresh');
+          });
+        },
+
         makeAdmin: function () {
           $http({
             method: "POST",
--- a/website/client/js/services/userServices.js
+++ b/website/client/js/services/userServices.js
@@ -310,16 +310,16 @@ angular.module('habitrpg')
           });
         },
 
-        setLastCron: function (date) {
+        setCustomDayStart: function (dayStart) {
           $http({
             method: "POST",
-            url: 'api/v3/user/set-cron',
+            url: 'api/v3/user/set-custom-day-start',
             data: {
-              lastCron: date
+              dayStart: dayStart
             }
           })
           .then(function (response) {
-            Notification.text('Last cron updated. Remember to refresh');
+            Notification.text('Day start updated. Remember to refresh');
           });
         },
 
--- a/website/server/controllers/api-v3/user.js
+++ b/website/server/controllers/api-v3/user.js
@@ -1311,22 +1311,22 @@ api.userReset = {
 };
 
 /**
-* @api {post} /api/v3/user/set-cron Sets lastCron for user
+* @api {post} /api/v3/user/set-custom-day-start Sets preferences.dayStart for user
 * @apiVersion 3.0.0
-* @apiName UserSetCron
+* @apiName UserSetCustomDayStart
 * @apiGroup User
 *
 * @apiSuccess {Object} data An empty Object
 */
-api.userSetCron = {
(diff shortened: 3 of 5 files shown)


crookedneighbor:
Per our conversation, the route should set the custom day start. 

Also, you need tests for the API route.

crookedneighbor:
That is to say that this PR creates a route to set user's last cron value when it should instead set the custom day start (which should also automatically set the lastCron value to the current date).

TheHollidayInn:
Okay, I didn't really understand the difference. Also, I don't see a field or function on the user model. How do I set the custom day start on the user on the server side?

paglias:
The field is preferences.dayStart

website/server/controllers/api-v3/user.js
```diff
@@ -1311,22 +1311,22 @@ api.userReset = {
...
   method: 'POST',
   middlewares: [authWithHeaders()],
-  url: '/user/set-cron',
+  url: '/user/set-custom-day-start',
   async handler (req, res) {
     let user = res.locals.user;
-    let cron = req.body.lastCron;
+    let dayStart = req.body.dayStart;
 
-    user.lastCron = cron;
```
paglias: it should also set user.lastCron as it did here https://github.com/HabitRPG/habitrpg/pull/7363/commits/5b7a56d28dd391311c513118d5240fd497bd9c86#diff-b4d231a8557edc4f769a50fa794b0bbaL81

TheHollidayInn: Okay, Blade said it should automatically set the lastCron. I figured that was from calling the route. Does the user service need to pass the date as well?

paglias: Hmm no, you can set it to Date.now() from the server

website/client/js/services/userServices.js
```diff
@@ -310,6 +310,19 @@ angular.module('habitrpg')
...
+        setCustomDayStart: function (dayStart) {
+          $http({
+            method: "POST",
+            url: 'api/v3/user/set-custom-day-start',
+            data: {
+              dayStart: dayStart
+            }
+          })
+          .then(function (response) {
+            Notification.text('Day start updated. Remember to refresh');
```
crookedneighbor, line 322: This should be translatable. Does the user need to refresh?

website/server/controllers/api-v3/user.js
```diff
@@ -1310,4 +1310,28 @@ api.userReset = {
...
+*
+* @apiSuccess {Object} data An empty Object
+*/
+api.userSetCustomDayStart = {
+  method: 'POST',
+  middlewares: [authWithHeaders()],
+  url: '/user/set-custom-day-start',
+  async handler (req, res) {
+    let user = res.locals.user;
+    let dayStart = req.body.dayStart;
```
crookedneighbor, line 1327: Do we need to do any validation on this?

test/spec/controllers/settingsCtrlSpec.js
```diff
@@ -17,6 +17,7 @@ describe('Settings Controller', function () {
         releasePets: sandbox.stub(),
         releaseMounts: sandbox.stub(),
         releaseBoth: sandbox.stub(),
+        setLastCron: sandbox.stub(),
```
crookedneighbor, line 20: Is this still correct?

website/server/controllers/api-v3/user.js
```diff
@@ -1310,4 +1310,28 @@ api.userReset = {
...
+* @apiVersion 3.0.0
+* @apiName UserSetCustomDayStart
+* @apiGroup User
+*
+* @apiSuccess {Object} data An empty Object
+*/
+api.userSetCustomDayStart = {
+  method: 'POST',
+  middlewares: [authWithHeaders()],
+  url: '/user/set-custom-day-start',
```
crookedneighbor, line 1324: Nitpick, but it's it implied that you're setting it by being a `POST`?

website/client/js/services/userServices.js
TheHollidayInn, line 322: I just copied from the debug function... Seems to me the concept is the similar. I don't think they need to refresh, but I figured you added the message for a reason.

website/server/controllers/api-v3/user.js
TheHollidayInn, line 1327: Again, I was just mimicking the debug route. I can put an is empty check on this. We seemed to want some other validation, but I didn't see clarification.

website/client/js/services/userServices.js
Alys, line 322: That debug function needs to say "refresh" (although "sync" would be better) because cron only happens when the user syncs, and the purpose of the debug function is to force an immediate cron.

This feature shouldn't need the browser to reload (and if it does for any reason, the reload should happen automatically because users often don't read advice).

TheHollidayInn, line 322: This function will set the last cron as well according to Paglias note above. So, it will do the same thing as the debug and some extra.

Alys:
@TheHollidayInn Do you need any information about how this is working on the current production site? The interaction between cron and changes to the Custom Day Start time can be complex and not intuitive if you've never used it.

website/server/controllers/api-v3/user.js
paglias, line 1327: There is validation for this on the user model but I'd add an is empty check anyway because it's not marked as required

website/client/js/services/userServices.js
paglias, line 322: @TheHollidayInn the refresh for the debug route was needed because it set lastCron to a past date so refreshing was needed so that cron would run. But in this case cron is set to now so it won't run, you can remove the request to refresh

crookedneighbor:
I'm going to take over this PR.

PR #7365, opened 2016-05-19
Skill fixes

crookedneighbor:
Fixes `heallAll -> healAll` and `spookDust -> spookySparkles`

Provides backwards compatibility for v2.


migrations/api_v3/users.js
```diff
@@ -113,6 +113,16 @@ function processUsers (afterId) {
 
       delete oldUser.id;
 
+      // spookDust -> spookySparkles
+      oldUser.achievments.spookySparkles = oldUser.achievements.spookDust;
```
SabreCat: Typo `achievments`

crookedneighbor: Good catch, thanks!

Issue #7312, opened 2016-05-16
[API v3] User.sortTag is not a function


paglias:
General Info
- UUID: 
- Browser: 
- OS:

TheHollidayInn:
@paglias The front end is an easy fix, however, the tags don't have any tagOrder field like tasks do. I think we need to add a index key to the tag model and update the api to use this. Does that sound correct to you?

paglias:
Hmm. The tags are saved as an array so we can directly reorder it
Il 20 mag 2016 8:20 AM, "Keith Holliday" notifications@github.com ha
scritto:

TheHollidayInn:
Oh, well never mind then :P

TheHollidayInn:
I still think I will need to add a new route. Maybe User update tags? Or should I add it to the tag controller?

paglias:
Oh yeah a new route is necessary, I'd say in the tags controller
Il 20 mag 2016 1:20 PM, "Keith Holliday" notifications@github.com ha
scritto:

PR #7381, opened 2016-05-20
Added tags update route. Added sort to user service

TheHollidayInn:
#7312
### Changes

Update user service to use sort tags
Update tag service to send update tags route
Added update tags route and tests

---

UUID:


paglias:
Hmm not sure about this one.. I think a moveTag route would be better like we have for tasks

TheHollidayInn:
Sounds good. I figured it could go either way. Ill update tonight.

paglias:
Sorry if i'm being picky on this but for tasks we only ask for the task id (tag id in this case) and `to` parameter and I'd like it to work the same way. Then you get the original position by searching for the position of the id in the array

TheHollidayInn:
No worries! I was just staying consistent with the user ops, but that makes more sense to be consistent with the tasks.

website/server/controllers/api-v3/tags.js
```diff
@@ -114,6 +114,41 @@ api.updateTag = {
...
+    req.checkBody('to', res.t('toRequired')).notEmpty();
+    req.checkBody('tagId', res.t('tagIdRequired')).notEmpty();
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let tagIndex = _.findIndex(user.tags, function findTag (tag) {
+      return tag.id === req.body.tagId;
+    });
+    if (tagIndex === !1) throw new NotFound(res.t('tagNotFound'));
```
paglias: hmm === !1? I think you mean === -1

TheHollidayInn: Meh.. Late nights. Let me try again with and add a test.

paglias:
The failing test is the usual one that always fails

PR #7383, opened 2016-05-20
Fix day for end of maintenance

keysmashes:
### Changes

End of maintenance changed to Sunday morning (from Saturday) – assuming typo, as maintenance runs from Saturday 21st 8pm to Sunday 22nd 5am UTC but ends on the Saturday (at 10pm) Pacific Time.

---

UUID: 0e1d15f7-f36b-4f15-a9e1-0c31b732dac9


crookedneighbor:
Thanks, but we decided not to use the config variables after all.

PR #7386, opened 2016-05-20
add dateCreated to all tasks; add empty challenge object to tasks tha…

Alys:
### Changes

adjusts API v2:
- put dateCreated back into all tasks
- add an empty challenge object to any tasks that don't have a challenge object (I don't know of any integrations that require this, but it can't hurt)


crookedneighbor:
Looks good to me. Let's wait for the build to finish and then merge it.

Alys:
build finished. Thanks Blade.

PR #7400, opened 2016-05-21
Update tasks.json

Atomk:
### Changes

Added a period for constistency with other strings. (hope this is the right way to do it)

UUID: d8fc8d8b-a10f-416a-b089-587db2af9e4d


crookedneighbor:
Thanks! Notated toward your next tier.

PR #7401, opened 2016-05-21
Docs changes

crookedneighbor:
Make some adjustments to the documentation.

Notably, mark webhook routes as beta.

I plan to expand functionality of webhook routes so you can set webhooks for multiple events.


crookedneighbor:
Alternatively we can set them to be ignored for now: http://apidocjs.com/#param-api-ignore

paglias:
I prefer BETA. If we ignore them people will think we don't have webhooks

PR #7403, opened 2016-05-22
Updated shortname validation to support multiple browsers

### LINKED ISSUE #7382: [API v3] Path shortName is required - if challenge tag isn't at least 3 symbols long
General Info
- UUID: -
- Browser: ff 46.0.1
- OS: win10
### Description

Creating a new challenge...
- set no tag - get notification in your default (system?) language (not the one Habitica is set to use) - looks strange
- set tag of e.g. two symbols - get error "Path `shortName` is required" instead of a specified error for this case
#### Console Errors

scr

TheHollidayInn:
Fixes #7382
### Changes

minLength is not supported in many browsers, so I added an alert message
I completely removed minlength because ng-model does not play well with it even when the browser does not support the attribute.

---

UUID:

### DIFF
--- a/common/locales/en/challenge.json
+++ b/common/locales/en/challenge.json
@@ -78,5 +78,6 @@
     "userTasksNoChallengeId": "When \"tasksOwner\" is \"user\" \"challengeId\" can't be passed.",
     "onlyChalLeaderEditTasks": "Tasks belonging to a challenge can only be edited by the leader.",
     "userAlreadyInChallenge": "User is already participating in this challenge.",
-    "cantOnlyUnlinkChalTask": "Only broken challenges tasks can be unlinked."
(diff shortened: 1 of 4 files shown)


paglias:
Looking good! One thing I would add is server side validation too, shortName is required but the minimum length isn't really enforced

TheHollidayInn:
Hmm. I thought it was. According to the issue, the server returns shortName required. The message was incorrect though, which is why I added the front end message. Let me take a look.

crookedneighbor:
It's not validated on the server currently. I will finish this.

crookedneighbor:
https://github.com/HabitRPG/habitrpg/commit/8cf4f9f41469eb161de97727b1d998c74a861d01

PR #7462, opened 2016-05-24
implement cron semi-safe mode - cron behaves as normal except boss does no damage

Alys:
As requested by Lemoness, this introduces a "semi-safe" mode for cron. Cron behaves as it normally does in all ways except:
- a quest boss does not cause damage
- the message in party chat when you attack the quest boss explains that the boss has not caused damage


crookedneighbor:
Assuming no tests need to be updated, 👍

Alys:
I can state absolutely that there are no tests to be updated.

PR #7475, opened 2016-05-24
Fix Armoire returns wrong equipment bug

crookedneighbor:
The armoire now does all calculations on the server. No chance for it to guess the wrong thing on the client. 

@Alys

### DIFF
--- a/website/client/js/controllers/tasksCtrl.js
+++ b/website/client/js/controllers/tasksCtrl.js
@@ -259,6 +259,11 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
       User.buy({params:{key:item.key}});
     };
 
+    $scope.buyArmoire = function () {
+      playRewardSound($scope.armoire);
+      User.buyArmoire();
+    }
+
     /*
      ------------------------
      Hiding Tasks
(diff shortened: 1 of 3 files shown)


paglias:
I'm  not sure this will sync changes correctly. You should use callOpsFunctionAndRequest

crookedneighbor:
Isn't the whole source of this bug that we're calling stuff on the client as well as the server?

crookedneighbor:
Looks like you may have already fixed the general bug then? https://github.com/HabitRPG/habitrpg/commit/7ea581bb38881f36f8b651172c871fb99b0fde5e

crookedneighbor:
How to reproduce some buggy behavior on the current `develop` branch:
- Make sure you have over 100gp
- Open two tabs
- in one, select the armoire - get a prize
- in the other, select the armoire - it tells you that you get a prize and then immediately sends a message that you don't have enough gold

The running the code on both the client and the server is what is causing the buggy behavior. This does the armoire calculations _only_ on the server and passes the response back.

paglias:
@crookedneighbor you're totally right. I thought I had already fixed this but it seems to only apply to random drops, armoire is still broken

What I meant with my comment earlier is that only calling the server function will show the notification message but won't add the item to the local user until sync

crookedneighbor:
There's a sync call once the response comes back.

paglias:
Didn't see it :( looks good to me!

paglias:
Actually it may be syncing twice because of the mechanism I added a few days ago to automatically sync when server and client are not up to date

crookedneighbor:
I'll check.

crookedneighbor:
Just checked, it does not sync twice.

PR #7476, opened 2016-05-24
Updated logic to not assign armoire incorrectly

### LINKED ISSUE #7418: Enchanted Armoire appears too early and without Ultimate Gear achievement
For a new player, the Enchanted Armoire appears before you have bought all the gear for one class.

Example:
1. Create a new account
2. Buy all the Warrior swords (must be all of them).
3. Sync after buying the final sword.
4. The Armoire appears but it tells you that you can't use it: "You can't buy this item."

At this point you don't have the Ultimate Gear achievement and none of the flags associated with the Armoire have been changed, so this is purely a display problem: the Armoire is displayed too early.

TheHollidayInn:
Fixes #7418
### Changes

Updated the ultimate gear check to use include and to grab the keys.
Forced the array to and object to remove the hidden fields

---

UUID:

### DIFF
--- a/common/script/fns/ultimateGear.js
+++ b/common/script/fns/ultimateGear.js
@@ -15,7 +15,9 @@ module.exports = function ultimateGear (user) {
     }
   });
 
-  if (_.contains(user.achievements.ultimateGearSets, true) && user.flags.armoireEnabled !== true) {
+  let hasFullSet = _.includes(Object.values(user.achievements.ultimateGearSets.toObject()), true);
+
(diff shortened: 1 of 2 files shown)


crookedneighbor:
You should update the tests to catch this error in the future.

crookedneighbor:
This creates a new problem:

screen shot 2016-05-24 at 9 11 43 pm

TheHollidayInn:
How did you make the error appear?

TheHollidayInn:
Nevermind. I forgot this was shared code so it runs on the client and server. toObject is not defined on the client.

PR #7478, opened 2016-05-24
Vagrantfile and scripts changes for API v3

### LINKED ISSUE #7456: Vagrant scripts for local installs need to be updated
The vagrant scripts for installing Habitica on a virtual machine need to be updated before new local installs can be created on vagrant. For example, it's now essential to use npm 3 instead of npm 2 and the current scripts install npm 2. Other changes might also be needed.

There's also this comment from sanserifs, from the Aspiring Coders guild:
"I'd even go so far as to suggest that the example Vagrantfile should have more memory added, with a warning to users with less RAM on their systems to try a swap file instead."

I think it's highly likely that that suggestion is correct. Swap files of 4 GB seem to be necessary for non-VM Ubuntu installs, so the RAM on the vagrant machines is probably far too small now.

If someone could investigate and fix, that would be lovely. Please test your fixes before submitting a PR by running up a Vagrant VM and confirming that a local install of Habitica can done without errors and with the website working. Save the full output of "vagrant up" and "npm install" and "npm start" and upload them to Gists or similar so that we can confirm that there's no bugs we need to fix in the npm install process.

See the "API Changes" section and its "Upgrading Your Local Installation to use API v3" subsection in [Guidance for Blacksmiths](http://habitica.wikia.com/wiki/Guidance_for_Blacksmiths#API_Changes), and [Setting up Habitica Locally](http://habitica.wikia.com/wiki/Setting_up_Habitica_Locally#Important_Notes).

Also if you find things that need to be updated in those documents, please do so! Or comment here with your suggested changes if you prefer.

shervinmathieu:
Fixes https://github.com/HabitRPG/habitrpg/issues/7456
### Changes

I have a new vagrant install running for API v3 after a very minor change (npm 2 to npm 3) and the local website is running.

There are some quirks with npm 3 and links, as reported [here](https://github.com/npm/npm/issues/10800) and [here](https://github.com/npm/npm/issues/10343). I've yet to be able to complete the npm install process in a single run of the command. Packages will go missing during the install process causing errors.

Here are the logs:

[Vagrant up](https://gist.github.com/velezsarain/65a923f7d66f9ae9c610c504e0533154)
[First run of npm install](https://gist.github.com/velezsarain/f9fb6e269b10d22c44d326153d1564e9)
[Second run of npm install with cache cleared](https://gist.github.com/velezsarain/0389cd7f854e67ec6a42787a48d0aeaf)
[npm start](https://gist.github.com/velezsarain/86f8dc35ff727c8c0f83db40b0d2cf03)

However I can't run any of the tests as I end up with weird errors. I've tried fixing them but was unable to (mostly because I can't find the files that are causing these errors). This should be looked into.

Here are the [test errors](https://gist.github.com/velezsarain/1299e4c0f94fcf6ba5738091da9f1123). No debug log is created for the npm test command. Here is the [debug log for npm run lint](https://gist.github.com/velezsarain/d2c94e52ebb1a423be0b6e144c432511). Error is the same (eslint .) in both cases. npm run test:karma causes another error, this time karma not found.

I would also ask how much RAM would be acceptable as the default for the future example file. I've set it to 4GB, but maybe more is necessary.

---

### DIFF
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -6,7 +6,7 @@ VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provider "virtualbox" do |v|
-    v.memory = 768
+    v.memory = 4096
     v.cpus = 1
     v.customize ["setextradata", :id, "VBoxInternal2/SharedFoldersEnableSymlinksCreate/vagrant", "1"]
   end
--- a/vagrant_scripts/install_node.sh
+++ b/vagrant_scripts/install_node.sh
@@ -16,7 +16,7 @@ nvm use
 nvm alias default current
 
 echo Update npm...
-npm install -g npm@2
+npm install -g npm@3
 
 echo Installing global modules...
 npm install -g gulp bower grunt-cli mocha


Alys:
At first glance, this is looking great! Very thorough, thank you.

> I've yet to be able to complete the npm install process in a single run of the command. Packages will go missing during the install process causing errors.

That's not uncommon. I've experienced it a lot with both Vagrant and non-VM Ubuntu installs. I've found it becomes much less common on Ubuntu as the swap file increases (my current one is 4GB), so maybe increasing RAM further would help. Perhaps try 5?

I'm wondering if those test errors are from the same problem, causing eslint to not be installed.

shervinmathieu:
eslint is installed, it is present in the node-packages directory. I've also ran additional runs of npm install command to make sure of it. I'm going to look further into it today, see if I can reach a solution.

shervinmathieu:
I found the cause of these problems, many packages installed are missing their executables in the node-modules/.bin directory.

I'm trying to figure out if it is an issue related to my computer or setup because I can't imagine npm would break the installs this bad.

Edit: Uninstalling and reinstalling culprit packages fixes this. It seems something other than missing packages is going horribly wrong with the npm install process. I have another computer I can try this on.

Edit 2: I've successfully ran the test suite after reinstalling the packages with missing executables on my main computer. I'll do the same on my laptop tomorrow.

shervinmathieu:
Having run the install process on my laptop without any issues whatsoever, it seems there is some quirk with my main computer that is hindering npm. Maybe permissions are not set properly on that drive, as Windows likes to restrict access to the main hard drive even when logged in as the administrator.

I believe the box is ready to go. I can update the wiki if that is okay with you.

Alys:
That sounds great! I'm happy to accept this. However some other commits have got caught up in it. If you're comfortable with git, would you like to fix it? Otherwise I'll clean it up when I merge it on the weekend.

shervinmathieu:
It is now cleaned up. I probably should have worked in a branch to avoid that mess. Sorry about that.

Alys:
@velezsarain No worries. :)

If you'd like to edit the wiki, that would be great, thank you! You could also mention in the Aspiring Coders guild that you've fixed the Vagrant configuration, if you'd like to (otherwise I will).

I've awarded you two tiers for this because we often award tier 1 for small changes and I know how much work is involved in testing vagrant installs! Thank you for this!

PR #7496, opened 2016-05-25
Apidoc Adjustments

crookedneighbor:
- Sets package.json to correct version number (I'll update our deploy documentation to bump the package.json version when releasing)
- Move apidoc config to separate file
- Turn off [apidoc versioning](http://apidocjs.com/#param-api-version)

To do versioning correctly, we'd need to copy the different API documentation for the previous versions to a separate file, which seems like a lot of work for only a little payoff (especially since we don't have a versioned api).

Open to other opinions though. 

@SabreCat @paglias @Alys

### DIFF
--- a/apidoc.json
+++ b/apidoc.json
@@ -1,5 +1,8 @@
 {
   "name": "Habitica V3 API Documentation",
   "title": "Habitica",
-  "url": "https://habitica.com"
+  "url": "https://habitica.com",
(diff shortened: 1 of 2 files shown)


crookedneighbor:
screen shot 2016-05-25 at 11 17 00 am

paglias:
Looks good to me 👍

PR #7498, opened 2016-05-25
fix(crit): notify from server response

SabreCat:
Fixes https://github.com/HabitRPG/habitrpg/issues/3166 and https://github.com/HabitRPG/habitrpg/issues/3048
### Changes

Removes notification for critical hits from a `$watch` in the notification controller, and instead displays it based on the data returned from the server.

We may also be able to deal with streak bonuses and drops in similar fashion, but those will require a bit more refactoring because the logic in the notification controller is more tangled.

---

UUID: meh


crookedneighbor:
Nice.

PR #7501, opened 2016-05-26
Mystery Items May 2016

SabreCat:
### Changes

Adds the Marching Bard Set, the subscriber reward for May 2016!

---

UUID: badum pum pum


mandyzhou:
Not sure if this is the right place to report this, but the Marching Bard Uniform says "April 2016 Subscriber Item" in its hovertext description.

crookedneighbor:
Thanks, I've fixed it. Will go live on the next deploy.

Issue #7509, opened 2016-05-26
Sporadic failing tests on score task routes


SabreCat:
### Description

When builds run on Travis CI, errors occur idiosyncratically in the tests for task scoring routes. They are of the format `POST /tasks/:id/score/:direction habits it scores and adds history`, suggesting they come from https://github.com/HabitRPG/habitrpg/blob/develop/test/api/v3/integration/tasks/challenges/POST-tasks_challenges_challengeId_tasks_id_score_direction.test.js.

Restarting the build sometimes causes it to run clean, so it doesn't appear to be a test failing outright, but perhaps some race condition that throws errors some times and not others.
#### Console Errors

The error thrown is `invalid request parameters`. I'm not seeing any further detail in the Travis log.

paglias:
Fixed

PR #7513, opened 2016-05-26
Fix #7493 - prevent duplicate items when editing checklists

### LINKED ISSUE #7493: Checklist - item duplicates
General Info
- UUID: 022bf9e4-e8fe-43d4-afb2-f367d79cf64b
- Browser: Firefox (up to date)
- OS: W7
### Description

Hello. When adding an item to an old checklist ... Edit the task, cursor on last filled item of the checklist, enter, type in the new item in the blank field, enter. The typed value has disappeared  and the created item is filled with the previous one. No hurry :-) Have a good day.

ltorroba:
Fixes #7493
### Changes

This PR has two commits.
1. The first commit closes issue #7493. This was done by changing the API call to `addChecklistItem` for a call to `saveTask`. The problem arose from the fact that, when an existing checklist was being edited, the first time the user pressed enter, the very last item would be re-added. This would cause the locally displayed checklist to be rewritten when the API response was received, and the newly added item's name was updated. An alternative fix, though doubtful, would be removing the sync. This seemed to work on my tests, but it would still effectively duplicate the checklist item upon pressing enter on the server-side. Alternatively, adding an extra textbox at the very end of the list upon entering edit mode started spawning even more bugs that I could initially foresee. I finally settled for `saveTask`.
2. The second commit cleans up the function, as the logic of the last two conditionals became virtually the same. I moved the `TODO` to an else statement, although I think it would be interesting to reevaluate whether or not UI feedback for that action is justified, and if so, what it would consist of (e.g.: nor a modal nor a popover seems right to me personally... perhaps changing the outline to red of the textbox to red?).

---

UUID: 8ac1be04-4201-4eed-a2a4-436eada5c0e9

### DIFF
--- a/website/client/js/controllers/tasksCtrl.js
+++ b/website/client/js/controllers/tasksCtrl.js
@@ -193,10 +193,7 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
         // Don't allow creation of an empty checklist item
         // TODO Provide UI feedback that this item is still blank
       } else if ($index == task.checklist.length - 1) {
-        Tasks.addChecklistItem(task._id, task.checklist[$index])
-          .then(function (response) {
-            task.checklist[$index] = response.data.data.checklist[$index];
-          });
+        $scope.saveTask(task, true);
         task.checklist.push({completed:false, text:''});
         focusChecklist(task, task.checklist.length - 1);
       } else {
--- a/website/client/js/controllers/tasksCtrl.js
+++ b/website/client/js/controllers/tasksCtrl.js
@@ -189,16 +189,13 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
     }
 
     $scope.addChecklistItem = function(task, $event, $index) {
-      if (!task.checklist[$index].text) {
-        // Don't allow creation of an empty checklist item
-        // TODO Provide UI feedback that this item is still blank
-      } else if ($index == task.checklist.length - 1) {
+      if (task.checklist[$index].text) {
         $scope.saveTask(task, true);
-        task.checklist.push({completed:false, text:''});
-        focusChecklist(task, task.checklist.length - 1);
-      } else {
-        $scope.saveTask(task, true);
+        if ($index == task.checklist.length - 1)
+          task.checklist.push({ completed: false, text: '' });
         focusChecklist(task, $index + 1);
+      } else {
+        // TODO Provide UI feedback that this item is still blank
       }
     }
 


Alys:
For point 2, I think not justified. At least it's certainly not urgent. Something to consider later maybe, so we can get this fix out sooner.

ltorroba:
I've been trying to find out what's wrong, but it seems like the failing tests have arisen out of #7509 and the recent 792aafd commit. Let's hope they get fixed soon.

crookedneighbor:
@sabrecat

crookedneighbor:
I've kicked off the build again. Should be fine now.

paglias:
I'm not sure changing the request from addChecklistItem to save is the correct way to go. Probably not syncing at [this](https://github.com/HabitRPG/habitrpg/pull/7513/files#diff-97708febaa4f3f9ac239cbdb2c5d7830L198) line is enough but it should be tested.

ltorroba:
@paglias I did some more tests, removing that line and keeping the function as it was. Indeed, removing that line does solve the issue, but not completely.

Let's say I create a new task, with two checklist items. I proceed to save it, and later on, I open it up to edit it and add more items:



I place my cursor in the checkbox for "Item two" and press enter, to add a new item. This triggers a call to _addChecklistItem_, and thus the server responds with the following object.



Notice how the last item, "Item Two", was duplicated as a result of pressing enter. As long as we remove the syncing line, this doesn't lead to any problems, but only because [the `$scope.saveTask` function indirectly overwrites the duplicates](https://github.com/zesme/habitrpg/blob/b20f5545b9acd429025f53bb0b6bb47b9e0770e4/website/client/js/controllers/tasksCtrl.js#L103-L107) when we press the checkmark to save the task.

Side-effects from this specific fix arise whenever adding an item to the checklist without subsequently calling `$scope.saveTask` to sync the server with the client. E.g.: Adding an item and quitting the browser before pressing save, bad internet connectivity, running Habitica on two tabs, etc.

Here is an example of what you'd see when running Habitica on two tabs:

_Tab 1: Editing checklist, I've pressed enter to add a new item, but have not typed anything yet_


_Tab 2: I refresh/open a new Habitica tab logged on to the same account - the checklist has not been fixed because `$scope.saveTask` has not been called yet_


Pressing save on Tab 1 and refreshing Tab 2 fixes the issue.

On the whole, I have to agree the margin for this to happen is extremely narrow, but I can't see any disadvantages to using saveTask instead - I might be very wrong though! 

Either way, if you guys feel it's best to use addChecklistItems instead, just let me know and I'll commit the fix right away. 😁

SabreCat:
Commit https://github.com/HabitRPG/habitrpg/commit/792aafd737608a5cb4a1f76c866576ac6d2cd19c was broken, yeah. It should be fixed now, though, if you haven't rebased already!

paglias:
@zesme good points!

crookedneighbor:

PR #7529, opened 2016-05-28
Use npm 3 in Dockerfile

### LINKED ISSUE #7525: Docker configuration for local installs need to be updated
The Docker configuration files for installing Habitica on a virtual machine need to be updated before new local installs can be created on docker. For example, it's now essential to use npm 3 instead of npm 2 and the current configuration probably installs npm 2. Other changes might also be needed.

Please test your fixes before submitting a PR by running up a docker container and confirming that a local install of Habitica can done without errors and with the website working. Save the full output from starting the docker container and "npm start" and upload them to Gists or similar so that we can confirm that there's no bugs we need to fix in the npm install process.

See the "API Changes" section and its "Upgrading Your Local Installation to use API v3" subsection in [Guidance for Blacksmiths](http://habitica.wikia.com/wiki/Guidance_for_Blacksmiths#API_Changes), and [Setting up Habitica Locally](http://habitica.wikia.com/wiki/Setting_up_Habitica_Locally#Important_Notes).

Also if you find things that need to be updated in those documents, please do so! Or comment here with your suggested changes if you prefer.

hairlessbear:
Fixes #7525. [Gist of `docker-compose up -d` output](https://gist.github.com/hairlessbear/674026813231c6eecec796672acd3c2b). I shut down `habitrpg_web` and restarted it with `docker start -a habitrpg_web_1` to capture [the output of `npm start`](https://gist.github.com/hairlessbear/6d223f83e849a65465e1038ba3386c5c). Tested creating a new user, logging in, and creating new to-dos to ensure it installed properly.

Surprisingly, the only change that was needed was updating to npm3.

Maybe worth mentioning: I had some errors initially due to severely outdated cookies stored in my browser from when I last worked on Habitica (~1 year ago). Might be worth adding instructions to the wiki to tell people to delete any localhost cookies they have if they experience certain errors.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ RUN apt-get clean
 RUN rm -rf /var/lib/apt/lists/*
 
 # Install global packages
-RUN npm install -g gulp grunt-cli bower
+RUN npm install -g gulp grunt-cli bower npm@3
 
 # Clone Habitica repo and install dependencies
 WORKDIR /habitrpg


Alys:
Fantastic, thank  you! I've noted that towards your tier 7.

Would you like to update the wiki to remove the warnings about not using Docker and mention it in Aspiriing Coders? I think it's definitely worth adding a note about cookies to Guidance for Blacksmiths too.

hairlessbear:
Certainly. Currently doing errands, but once I get back home I'll get right
on it

PR #7531, opened 2016-05-29
add CRON_SLIGHTLY_SAFE_MODE to prevent damage from an offset double cron

Alys:
Partial workaround for https://github.com/HabitRPG/habitrpg/issues/2805 - see https://github.com/HabitRPG/habitrpg/issues/2805#issuecomment-222336424

**WIP! DO NOT MERGE!** Submitting now in case anyone wants to review my approach. This needs more code and manual testing, and automatic tests that will be included before this is merged.
### Changes
- Adds a new environment variable, CRON_SLIGHTLY_SAFE_MODE (feel free to suggest a new name).
- If that variable is true and if the user's damage to the boss is very small, then we assume that the current cron is an incorrect duplicate which has occurred after a real cron (i.e., the real cron has already unticked all the user's Dailies and has applied the user's real damage to the boss). In that case:
  - prevents any damage being done to the party by the boss
  - prints a special message to party chat

Still to be done:
- prevent personal damage to the player from their own Dailies
- prevent streak loss
- maybe other things
- tests

### DIFF
--- a/website/server/libs/api-v3/cron.js
+++ b/website/server/libs/api-v3/cron.js
@@ -8,7 +8,6 @@ import nconf from 'nconf';
 
 const CRON_SAFE_MODE = nconf.get('CRON_SAFE_MODE') === 'true';
 const CRON_SEMI_SAFE_MODE = nconf.get('CRON_SEMI_SAFE_MODE') === 'true';
-const CRON_SLIGHTLY_SAFE_MODE = nconf.get('CRON_SLIGHTLY_SAFE_MODE') === 'true';
 const shouldDo = common.shouldDo;
 const scoreTask = common.ops.scoreTask;
 // const maxPMs = 200;
@@ -21,7 +20,6 @@ let CLEAR_BUFFS = {
   stealth: 0,
   streaks: false,
 };
-let MIN_PROGRESS_UP = 2; // if damage to boss is not more than this, assume duplicate cron, when running in CRON_SLIGHTLY_SAFE_MODE https://github.com/HabitRPG/habitrpg/issues/2805#issuecomment-222336424
 
 function grantEndOfTheMonthPerks (user, now) {
   let plan = user.purchased.plan;
@@ -135,10 +133,6 @@ export function cron (options = {}) {
   // If the user does not log in for two or more days, cron (mostly) acts as if it were only one day.
   // When site-wide difficulty settings are introduced, this can be a user preference option.
 
-  // If running in CRON_SLIGHTLY_SAFE_MODE, assume that small damage to a boss means this is a double cron
-  let progress = user.party.quest.progress;
-  let progressUpIsTiny = CRON_SLIGHTLY_SAFE_MODE && progress.up <= MIN_PROGRESS_UP;
-
   // Tally each task
   let todoTally = 0;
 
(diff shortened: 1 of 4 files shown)


website/server/libs/api-v3/cron.js
```diff
@@ -20,6 +21,7 @@ let CLEAR_BUFFS = {
   stealth: 0,
   streaks: false,
 };
+let MIN_PROGRESS_UP = 2; // if damage to boss is not more than this, assume duplicate cron, when running in CRON_SLIGHTLY_SAFE_MODE https://github.com/HabitRPG/habitrpg/issues/2805#issuecomment-222336424
```
crookedneighbor: Make this a const instead?

website/server/models/group.js
```diff
@@ -27,6 +27,7 @@ export const TAVERN_ID = shared.TAVERN_ID;
 
 const CRON_SAFE_MODE = nconf.get('CRON_SAFE_MODE') === 'true';
 const CRON_SEMI_SAFE_MODE = nconf.get('CRON_SEMI_SAFE_MODE') === 'true';
+// const CRON_SLIGHTLY_SAFE_MODE = nconf.get('CRON_SLIGHTLY_SAFE_MODE') === 'true';
```
crookedneighbor: Commented out on purpose?

website/server/models/group.js
```diff
@@ -519,7 +520,17 @@ schema.statics.bossQuest = async function bossQuest (user, progress) {
...
   let playerAttack = `${user.profile.name} attacks ${quest.boss.name('en')} for ${progress.up.toFixed(1)} damage.`;
-  let bossAttack = CRON_SAFE_MODE || CRON_SEMI_SAFE_MODE ? `${quest.boss.name('en')} does not attack, because it respects the fact that there are some bugs\` \`post-maintenance and it doesn't want to hurt anyone unfairly. It will continue its rampage soon!` : `${quest.boss.name('en')} attacks party for ${Math.abs(down).toFixed(1)} damage.`;
+  let bossAttack = `${quest.boss.name('en')} `;
+  if (CRON_SAFE_MODE || CRON_SEMI_SAFE_MODE) {
+    bossAttack += `does not attack, because it respects the fact that there are some bugs\` \`post-maintenance and it doesn't want to hurt anyone unfairly. It will continue its rampage soon!`;
+  } else if (CRON_SLIGHTLY_SAFE_MODE && down === 0) {
+    // ALYS TODO TODAY pass something in from cron() in website/server/libs/api-v3/cron.js so that we only run this when we really are making the double-cron assumtion (i.e., not just when the user got a perfect day)
+    bossAttack += `does not attack, because it thinks a bug made you swing and miss.`;
```
crookedneighbor, line 532: LOVE THIS.

website/server/models/group.js
Alys: Sort of. I'm still working on that file and had forgotten I'd left it commented out. :) I'm hopinng to avoid using it altogether though, in that file. I want to pass in a flag from cron() instead.

website/server/libs/api-v3/cron.js
Alys: Ah, of course. Thanks!

crookedneighbor:
I don't want this to be a permanent thing, but I think while we're trying to figure this out, it's fine. Maybe log out when this happens so we can track it down?

Alys:
The code is nicer now but it still isn't even manually tested because my laptop's been nearly unusable for a few hours due to a slow database import. I'm giving up for today. I'll try to get this finished tomorrow, although if anyone's looking for something to do, feel free to play with it;

Alys:
Closing because paglias has a much better idea of what's going wrong and how to fix it.

PR #7534, opened 2016-05-29
set lastCron at the start of cron instead of the end

Alys:
possible partial workaround for https://github.com/HabitRPG/habitrpg/issues/2805
### Changes

This sets lastCron when cron starts running (in api.cron()) instead of setting it when cron ends (in middleware cronAsync()) to see if that helps prevent the new double cron bug: https://github.com/HabitRPG/habitrpg/issues/2805#issuecomment-222314260

auth.timestamps.loggedin is still set when cron ends because I'm curious about the difference between the timestamps. I can't see any code that could possibly break from its value being different to lastCron

@paglias what do you think? As with the cron change in getGroup() this shouldn't be necessary but maybe it will make a difference.

### DIFF
--- a/website/server/libs/api-v3/cron.js
+++ b/website/server/libs/api-v3/cron.js
@@ -110,6 +110,7 @@ export async function recoverCron (status, locals) {
 export function cron (options = {}) {
   let {user, tasksByType, analytics, now = new Date(), daysMissed, timezoneOffsetFromUserPrefs} = options;
 
+  user.lastCron = now; // TODO does putting this here help prevent this double-cron bug: https://github.com/HabitRPG/habitrpg/issues/2805#issuecomment-222314260
   user.preferences.timezoneOffsetAtLastCron = timezoneOffsetFromUserPrefs;
   // User is only allowed a certain number of drops a day. This resets the count.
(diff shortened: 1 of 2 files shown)


website/server/libs/api-v3/cron.js
```diff
@@ -110,6 +110,7 @@ export async function recoverCron (status, locals) {
 export function cron (options = {}) {
   let {user, tasksByType, analytics, now = new Date(), daysMissed, timezoneOffsetFromUserPrefs} = options;
 
+  user.lastCron = now; // TODO does putting this here help prevent this double-cron bug: https://github.com/HabitRPG/habitrpg/issues/2805#issuecomment-222314260
```
paglias, line 113: this is not setting lastCron at the beginning of the process but only when the user is saved

paglias:
What you could do is setting it here https://github.com/Alys/habitrpg/blob/5dbfc0b2a2b4304d380db89b9108b3dd9c0f993e/website/server/middlewares/api-v3/cron.js#L112

paglias:
Included in https://github.com/HabitRPG/habitrpg/pull/7563

PR #7538, opened 2016-05-29
fix: Prevent user from inivting self to group

crookedneighbor:
Prevents user from inviting self to a group.

### DIFF
--- a/common/locales/en/groups.json
+++ b/common/locales/en/groups.json
@@ -177,6 +177,7 @@
     "inviteMissingEmail": "Missing email address in invite.",
     "partyMustbePrivate": "Parties must be private",
     "userAlreadyInGroup": "User already in that group.",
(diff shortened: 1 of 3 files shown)


common/locales/en/groups.json
```diff
@@ -177,6 +177,7 @@
     "inviteMissingEmail": "Missing email address in invite.",
     "partyMustbePrivate": "Parties must be private",
     "userAlreadyInGroup": "User already in that group.",
+    "cannotInviteSelfToGroup": "You cannot invite yourself to a group.",
```
crookedneighbor, line 180: We _could_ use the `userAlreadyInGroup` string, but I think being more specific is probably better.

crookedneighbor:
I don't think this is controversial, so I'm merging it in.

PR #7543, opened 2016-05-30
Fix challenge sorting

### LINKED ISSUE #7507: Challenge Order: Desired Behavior
In general, Challenges should display in order of newest-oldest, to offer exposure for newly-created Challenges.

The only exception to this rule are Official Habitica Challenges, which should stick to the top of the list, marked with the "Official Challenge" tag. These Challenges are designated by a staff member marking the corresponding checkbox during Challenge creation.

@crookedneighbor

shervinmathieu:
Fixes #7507
### Changes

This one was surprisingly easy to fix, as it was only an issue with the .sort() method being passed a wrong argument (-timestamp instead of -createdAt, as it is stored in the database).

However it took me a while to find the issue as I did not know how the Habitica code works (as you can see in my comment in the issue thread), so I spent most of my time figuring out the location of the controllers and such.

---

### DIFF
--- a/test/api/v3/integration/challenges/GET-challenges_group_groupid.test.js
+++ b/test/api/v3/integration/challenges/GET-challenges_group_groupid.test.js
@@ -79,8 +79,7 @@ describe('GET challenges/group/:groupId', () => {
       expect(foundChallengeIndex).to.eql(0);
     });
   });
-
-
+  
   context('Private Guild', () => {
     let privateGuild, user, nonMember, challenge, challenge2;
 
--- a/test/api/v3/integration/challenges/GET-challenges_group_groupid.test.js
+++ b/test/api/v3/integration/challenges/GET-challenges_group_groupid.test.js
@@ -76,7 +76,7 @@ describe('GET challenges/group/:groupId', () => {
         official: true,
       });
 
-      let challenges = await nonMember.get(`/challenges/groups/${publicGuild._id}`);
+      let challenges = await user.get(`/challenges/groups/${publicGuild._id}`);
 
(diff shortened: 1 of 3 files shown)


crookedneighbor:
Great work!

Two things:
1. Does this put the official challenges at the top as the issue specifies? (On my phone, so I can't check myself)
2. We'll need to add an integration test for this behavior. (Let me know if you need help)

shervinmathieu:
In my local install official challenges are put on top, and sorted newest first.

I am going to give a shot at the integration test.

shervinmathieu:
I finished writing the tests, at the end of the public group context for both files, there's 2 new tests:
-  the first test modifies the group leader to be an admin and creates a new official challenge. I then pull this challenge's index position in the array returned by the api. As it is the only official challenge, its position should be 0.
-  the second one created a brand new challenge and tests its index position to be 1, as it is the latest challenge to be created but is not the official one. I create this new challenge to make sure that the sorting is working as intended and is not just a byproduct of the order the challenges were put into the database.

I am not sure about modifying the user to be an admin, as it could break future tests, but as it is the only way to add an official challenge, I did it anyway. Maybe in the future "before async" will change to "beforeEach async".

shervinmathieu:
I wasn't quite happy with the tests I wrote before so I moved some of them to new contexts I made (for official challenge sorting) and improved them a little bit by adding some more checks to make sure the challenges are in order and stay as such when new ones are added in.

shervinmathieu:
Also my IDE likes to add a lot of trailing spaces...

crookedneighbor:
Good work. I've noted this as partial credit towards tier 3.

PR #7547, opened 2016-05-30
Don't $watch drops

SabreCat:
### Changes

Removes the Angular `$watch` from random drops and opening Mystery Item boxes, moving handling of those notifications instead to the User service.

Also brings back individualized graphics for the mystery item you obtained when opening the box.

---

UUID: =^.,.^=

### DIFF
--- a/common/script/ops/openMysteryItem.js
+++ b/common/script/ops/openMysteryItem.js
@@ -31,9 +31,8 @@ module.exports = function openMysteryItem (user, req = {}, analytics) {
     return user.items.gear.owned;
   } else {
     return [
-      user.items.gear.owned,
-      i18n.t('mysteryItemOpened', req.language),
       item,
+      i18n.t('mysteryItemOpened', req.language),
     ];
   }
 };
--- a/common/script/ops/openMysteryItem.js
+++ b/common/script/ops/openMysteryItem.js
@@ -13,7 +13,6 @@ module.exports = function openMysteryItem (user, req = {}, analytics) {
   }
 
   item = _.cloneDeep(content.gear.flat[item]);
-  item.notificationType = 'Mystery';
   user.items.gear.owned[item.key] = true;
 
   user.markModified('purchased.plan.mysteryItems');
@@ -28,17 +27,13 @@ module.exports = function openMysteryItem (user, req = {}, analytics) {
     });
   }
 
-  if (typeof window !== 'undefined') {
-    if (!user._tmp) user._tmp = {};
-    user._tmp.drop = item;
-  }
-
   if (req.v2 === true) {
     return user.items.gear.owned;
   } else {
     return [
       user.items.gear.owned,
       i18n.t('mysteryItemOpened', req.language),
+      item,
     ];
   }
 };
--- a/website/client/js/services/userServices.js
+++ b/website/client/js/services/userServices.js
@@ -14,8 +14,8 @@ angular.module('habitrpg')
 /**
  * Services that persists and retrieves user from localStorage.
  */
-  .factory('User', ['$rootScope', '$http', '$location', '$window', 'STORAGE_USER_ID', 'STORAGE_SETTINGS_ID', 'Notification', 'ApiUrl', 'Tasks', 'Tags',
-    function($rootScope, $http, $location, $window, STORAGE_USER_ID, STORAGE_SETTINGS_ID, Notification, ApiUrl, Tasks, Tags) {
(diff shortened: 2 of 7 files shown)


common/script/ops/openMysteryItem.js
```diff
@@ -28,17 +27,13 @@ module.exports = function openMysteryItem (user, req = {}, analytics) {
...
-    user._tmp.drop = item;
-  }
-
   if (req.v2 === true) {
     return user.items.gear.owned;
   } else {
     return [
       user.items.gear.owned,
       i18n.t('mysteryItemOpened', req.language),
+      item,
```
paglias, line 34: I don't think this will work... You can only return [data, message] in common ops

paglias, line 34: I see that not returning the item in first place wasn't very smart on my side... but now changing the response to [{item, owned}, message] would break compatibility

website/client/js/services/userServices.js
```diff
@@ -231,7 +236,49 @@ angular.module('habitrpg')
...
+                Notification.drop(env.t('messageDropFood', {dropArticle: drop.article, dropText: text, dropNotes: notes}), drop);
+              } else if (drop.type === 'Quest') {
+                $rootScope.selectedQuest = Content.quests[drop.key];
+                $rootScope.openModal('questDrop', {controller:'PartyCtrl', size:'sm'});
+              } else {
+                // Keep support for another type of drops that might be added
+                Notification.drop(drop.dialog);
+              }
+
+              // Analytics.track({'hitType':'event','eventCategory':'behavior','eventAction':'acquire item','itemName':after.key,'acquireMethod':'Drop'});
```
SabreCat, line 280: This is the one part I'm not sure how to deal with. If I inject Analytics here, I get a circular dependency: `User -> Analytics -> User`. Can we move this tracking server side?

paglias, line 280: I think it makes sense since the code must run on the server anyway

common/script/ops/openMysteryItem.js
SabreCat, line 34: Hmm, really? In manual tests, it behaved just fine. Will it break other API consumers?

paglias, line 34: I'm pretty sure that the res.respond code only works with 3 arguments (statusCode, data, message) https://github.com/HabitRPG/habitrpg/blob/develop/website/server/middlewares/api-v3/response.js#L3 and yeah it would probably break API consumers. @crookedneighbor what do you think?

crookedneighbor, line 34: @SabreCat You manually tested mystery items? I'm with @paglias, I think the item is just being ignored and not actually being passed on. 

I say we call it a bug and fix the data being passed back. Better to do it now than down the line. Let's check with @vIiRuS and make sure the mobile apps are not using the mystery box route.

crookedneighbor, line 34: Could we return this instead:

``` js
{
  gearOwned: user.items.gear.owned,
  item,
}
```

crookedneighbor, line 34: That way, if people want a quick an easy way to overwrite the `gear.owned` user prop, they can.

I'm fine with just returning the item though. Returning an object with properties is more forwards compatible.

paglias: we could but wouldn't it be a breaking change since the top level object won't be `user.items.gear.owned ` anymore

crookedneighbor, line 34: It would be a breaking change, but of all the routes, this one is least likely to be used outside the site.

SabreCat, line 34: I did test mystery items, aye. But I am wholly in favor of making a minor breaking change for better forward compatibility!

common/script/ops/openMysteryItem.js
```diff
@@ -28,16 +27,11 @@ module.exports = function openMysteryItem (user, req = {}, analytics) {
...
-    if (!user._tmp) user._tmp = {};
-    user._tmp.drop = item;
-  }
-
   if (req.v2 === true) {
     return user.items.gear.owned;
   } else {
     return [
-      user.items.gear.owned,
+      item,
```
paglias, line 34: this will require an update to the API docs in the controller file

SabreCat, line 34: Right right! Also tests, it turns out.

SabreCat, line 34: Done.

paglias:
Looks good to me 👍

crookedneighbor:
Assuming the build passes, 👍

paglias:
Looks like you missed one integration test :)

SabreCat:
Indeed. Someday, maybe, I will not feel like a bull in a china shop when committing...

PR #7553, opened 2016-05-30
Fix tag removal functionality

### LINKED ISSUE #7412: Unselected tags are being added to new tasks and tags cannot be removed from new/existing tasks
# System

UUID: fedc44cb-088a-4bc8-8f4f-3e6a1882f042
Browsers: Safari, FF, Chrome 
OS: OS X 10.11.4

The browser seems to 'accumulating' tags that were previously selected in the tag filter on the page so that new tasks created have all the tags that were previously selected, even if they are no longer highlighted. Refreshing the browser clears the accumulation so that new tasks added don't have multiple tags. But it is not possible to remove a tag from a task once it has been added.
# How to reproduce:

1) Clear cache and reload the page.
2) Select multiple tags.

screen shot 2016-05-22 at 3 13 18 am

3) Unselect the tags.
4) Select only one tag, and create a new task.

screen shot 2016-05-22 at 3 11 50 am

5) Open the task to see the tags and note that all of the tags previously selected are on the task (there should be only one tag, but instead there are 5)

screen shot 2016-05-22 at 3 13 31 am

6) Deselect the unwanted tags and save the task.
7) Reopen the task and note that the unselected tags are still selected.
8) Refresh the page. Note that no tags are selected in the tag filter at the top of the page (this is a new behavior -- previously it was possible to reload the page and not lose the tags that were selected)
9) Select one tag and create a new task.
10) Open the task to see the tags and note that now it has only the one tag that was selected.
11) Deselect the tag and save the task.
12) Reopen the task and note that the unselected tag is still selected.
## Console Errors:

None, but there are several unrelated warnings
screen shot 2016-05-22 at 3 21 12 am
screen shot 2016-05-22 at 3 22 18 am

ltorroba:
Fixes #7412
### Changes

Changing second argument for splice from `0` to `1` sufficed.

---

UUID: 8ac1be04-4201-4eed-a2a4-436eada5c0e9

Tag removal functionality was broken. Turns out the issue was simply
that we weren't removing the tag from the local `task` object, so the
tag was removed, but readded upon sync/pressing save.

Closes #7412

### DIFF
--- a/website/client/js/controllers/tasksCtrl.js
+++ b/website/client/js/controllers/tasksCtrl.js
@@ -305,7 +305,7 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
         task.tags.push(tagId);
       } else {
         Tasks.removeTagFromTask(task._id, tagId);
-        task.tags.splice(tagIndex, 0);
+        task.tags.splice(tagIndex, 1);
       }
     }
   }]);


SabreCat:
Gave this a try locally. It appears to be a partial fix: you can successfully remove tags, now, but the unintended stickiness of the tag buttons that leads to having lots of erroneous tags in the first place still occurs.

SabreCat:
Merged and noted toward your next contributor tier. Thanks @zesme!

Issue #7535, opened 2016-05-29
when moderator unflags a post, email must be sent


Alys:
Currently, when any player flags a post, an email is sent to some of the moderators.

We need a similar email to be sent when a moderator removes the flags from a post.

The recipient list should be the same as for flag emails.

The subject should be the same so that the unflag email is threaded with the matching flag email.

The body can be very similar, but with moderator name and unflagging date instead of the name etc of the person who flagged, and with a clear message near the top that the post has been unflagged.

shervinmathieu:
I'm going to work on this.

shervinmathieu:
However, is there any way to enable emails in development? I can't test changes if there is no way to do so.

quin-jay:
@velezsarain You may have already figured this out, but in your local install there's a config.json file that includes a line where you can specify an email address. I believe any tests you do with your local account will send a real email to that address (and it has nothing to do with login information, so you can use the same email address as your real Habitica account).

quin-jay:
Although, if the email is to be sent to mods only, config.json won't help. Oops, disregard previous message! :relaxed:

paglias:
@lynnyliz the email setting you see in config.json is only used for password recovery that for 100% delivery have to be sent directly from our server. All other emails are sent through Mandrill

PR #7580, opened 2016-06-01
send email when admin unflags post

shervinmathieu:
Implements #7535
### Changes

New block at the en of the unflag API call that hopefully sends an email to admins to notify them an admin has unflagged a post. 

However since apparently the email service does not work in a dev environment I was unable to test my code. I am unsure if it is the correct syntax for sendTxn or not.

The mail recipients and title _should_ be the same. The mail would begin by a 'Unflagged by admin' text. This one is hardcoded, as I am not sure where to put the string in the locale directory (I'm thinking front.json for now) The details of the admin who performed the action are also sent.

---

### DIFF
--- a/test/api/v3/unit/libs/email.test.js
+++ b/test/api/v3/unit/libs/email.test.js
@@ -49,23 +49,6 @@ function getUser () {
   };
 }
 
-function getGroup (type) {
-  if (type === 'tavern') {
-    return {
-      _id: TAVERN_ID,
-      type: 'guild',
-    };
-  } else if (type === 'guild') {
-    return {
-      _id: 'random _id',
-      type: 'guild',
-    };
-  } else return {
-    _id: 'random _id',
-    type: 'group',
-  };
-}
-
 describe('emails', () => {
   let pathToEmailLib = '../../../../../website/server/libs/api-v3/email';
 
@@ -157,24 +140,18 @@ describe('emails', () => {
 
   describe('getGroupUrl', () => {
     it('returns correct url if group is the tavern', () => {
-      let attachEmail = requireAgain(pathToEmailLib);
-      let getGroupUrl = attachEmail.getGroupUrl;
-      let group = getGroup('tavern');
-      expect(getGroupUrl(group)).to.eql('/#/options/groups/tavern');
+      let getGroupUrl = require(pathToEmailLib).getGroupUrl;
+      expect(getGroupUrl({_id: TAVERN_ID, type: 'guild'})).to.eql('/#/options/groups/tavern');
     });
 
     it('returns correct url if group is a guild', () => {
-      let attachEmail = requireAgain(pathToEmailLib);
-      let getGroupUrl = attachEmail.getGroupUrl;
-      let group = getGroup('guild');
-      expect(getGroupUrl(group)).to.eql(`/#/options/groups/guilds/${group._id}`);
+      let getGroupUrl = require(pathToEmailLib).getGroupUrl;
+      expect(getGroupUrl({_id: 'random _id', type: 'guild'})).to.eql('/#/options/groups/guilds/random _id');
     });
 
     it('returns correct url if group is a party', () => {
-      let attachEmail = requireAgain(pathToEmailLib);
-      let getGroupUrl = attachEmail.getGroupUrl;
-      let group = getGroup('party');
-      expect(getGroupUrl(group)).to.eql('party');
+      let getGroupUrl = require(pathToEmailLib).getGroupUrl;
+      expect(getGroupUrl({_id: 'random _id', type: 'party'})).to.eql('party');
     });
   });
 
--- a/test/api/v3/unit/libs/email.test.js
+++ b/test/api/v3/unit/libs/email.test.js
@@ -5,6 +5,7 @@ import nodemailer from 'nodemailer';
 import Bluebird from 'bluebird';
 import requireAgain from 'require-again';
 import logger from '../../../../../website/server/libs/api-v3/logger';
+import { TAVERN_ID } from '../../../../../website/server/models/group';
 
 function defer () {
   let resolve;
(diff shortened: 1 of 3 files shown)


website/server/controllers/api-v3/chat.js
```diff
@@ -300,6 +300,52 @@ api.clearChatFlags = {
...
+    if (group._id === TAVERN_ID) {
+      groupUrl = '/#/options/groups/tavern';
+    } else if (group.type === 'guild') {
+      groupUrl = `/#/options/groups/guilds/${group._id}`;
+    } else {
+      groupUrl = 'party';
+    }
+
+    sendTxn(FLAG_REPORT_EMAILS, 'flag-report-to-mods', [
+      {name: 'STATUS', content: 'UNFLAGGED BY ADMIN'},
```
paglias: this should be removed, I'll update the email text separately

website/server/controllers/api-v3/chat.js
```diff
@@ -300,6 +300,52 @@ api.clearChatFlags = {
...
+    let groupUrl;
+    if (group._id === TAVERN_ID) {
+      groupUrl = '/#/options/groups/tavern';
+    } else if (group.type === 'guild') {
+      groupUrl = `/#/options/groups/guilds/${group._id}`;
+    } else {
+      groupUrl = 'party';
+    }
+
+    sendTxn(FLAG_REPORT_EMAILS, 'flag-report-to-mods', [
```
paglias: the email name (`flag-report-to-mods`) should be changed to something like `unflag-report-to-mods`

website/server/controllers/api-v3/chat.js
```diff
@@ -300,6 +300,52 @@ api.clearChatFlags = {
       {$set: {'chat.$.flagCount': message.flagCount}}
     );
 
+    let adminEmailContent;
+    if (user.auth.local) {
```
paglias: here to get the email addresses could you use the `email.getUserInfo(user, ['email'])` function from https://github.com/HabitRPG/habitrpg/blob/develop/website/server/libs/api-v3/email.js#L46? It would be great if you could change it here too https://github.com/HabitRPG/habitrpg/blob/develop/website/server/controllers/api-v3/chat.js#L212

paglias:
@velezsarain emails are not really sent from the server, the server makes an HTTP request to an email server (the code for which can found here https://github.com/HabitRPG/habitrpg-email-server) that handles the sending. Unfortunately the emails templates are hosted directly on our email provider and so can't be edited by contributors.

I've left a few comments on the PR but overall it looks good. When this goes live I'll update the email text

shervinmathieu:
I've pushed a new commit with the changes you requested.

website/server/controllers/api-v3/chat.js
```diff
@@ -300,6 +291,41 @@ api.clearChatFlags = {
...
     );
 
+    let adminEmailContent = getUserInfo(user, ['email']).email;
+
+    let author = await User.findOne({_id: message.uuid}, {auth: 1});
+
+    let authorEmailContent = getUserInfo(author, ['email']).email;
+
+    let groupUrl;
+    if (group._id === TAVERN_ID) {
```
crookedneighbor: This logic is used in the flag route too, right? This should be abstracted out into a function that both can use then.

crookedneighbor: IE: `let groupUrl = getGroupUrl(group);`

shervinmathieu:
getGroupUrl is now its own method. I put it into email.js because I didn't find a better place to put it in.

paglias:
This is looking good to me 👍 can you add a test for `getGroupUrl`? Then I'll merge it

shervinmathieu:
There's now unit test coverage for the new function (method? I never know)

shervinmathieu:
I also pressed the wrong button. oops

test/api/v3/unit/libs/email.test.js
```diff
@@ -137,6 +155,29 @@ describe('emails', () => {
     });
   });
 
+  describe('getGroupUrl', () => {
+    it('returns correct url if group is the tavern', () => {
+      let attachEmail = requireAgain(pathToEmailLib);
```
crookedneighbor: requireAgain shouldn't be needed because you're not doing anything that necessitates clearing the require cache.

crookedneighbor: And the getGroup function seems like overkill to me. I'd do:

``` js
describe('getGroupUrl', () => {
  let getGroupUrl = require(pathToEmailLib).getGroupUrl;

  it('returns correct url if group is the tavern', () => {
    let url = getGroupUrl({_id: TAVERN_ID, type: 'guild'});

    expect(url).to.eql('/#/options/groups/tavern');
  });
```

test/api/v3/unit/libs/email.test.js
```diff
@@ -137,6 +155,29 @@ describe('emails', () => {
...
+    });
+
+    it('returns correct url if group is a guild', () => {
+      let attachEmail = requireAgain(pathToEmailLib);
+      let getGroupUrl = attachEmail.getGroupUrl;
+      let group = getGroup('guild');
+      expect(getGroupUrl(group)).to.eql(`/#/options/groups/guilds/${group._id}`);
+    });
+
+    it('returns correct url if group is a party', () => {
```
crookedneighbor, line 152: Technically speaking, this `returns the party url for any group that is not a guild`

website/server/libs/api-v3/email.js
```diff
@@ -64,6 +65,19 @@ export function getUserInfo (user, fields = []) {
   return info;
 }
 
+export function getGroupUrl (group) {
+  let groupUrl;
+  if (group._id === TAVERN_ID) {
+    groupUrl = '/#/options/groups/tavern';
+  } else if (group.type === 'guild') {
+    groupUrl = `/#/options/groups/guilds/${group._id}`;
+  } else {
```
crookedneighbor: Let's make this `else  if (group.type === 'party') {`, just to be clear about it.

test/api/v3/unit/libs/email.test.js
crookedneighbor, line 152: Although, see https://github.com/HabitRPG/habitrpg/pull/7580/files#r65692745

shervinmathieu:
Beginners mistakes!

crookedneighbor:
No worries. You were just following patterns in the test file. I wrote [require-again](https://www.npmjs.com/package/require-again) to tackle a very specific problem.

This looks good to me. When @paglias is ready, we'll do some additional testing to make sure it works and then merge it in.

paglias:
Sorry for the delay. The email template has been correctly published, so this is ready to be merged 👍

crookedneighbor:

PR #7583, opened 2016-06-01
fix: Cleanup quest progress for non-members when quest starts

### LINKED ISSUE #7353: Reject button should always delete quest invitation (was: Keep getting invitations for an already completed quest)
_edit by admin:_

The immediate problem described below has been fixed (the bad invitation no longer exists).

**However we should change the reject button so that it always works, even if the invitation is for a quest or party that are no longer relevant. I.e., clicking the reject button should ALWAYS completely remove the invitation.**

This issue is now for implementing that change.

---

General Info
- UUID: ffa26f90-f4cf-40b5-8b3b-aca988ab8ed2
- Browser: FF 46.0.1
- OS: win10
### Description

After a quest had been started and completed by another party member for the sake of testing for https://github.com/HabitRPG/habitrpg/issues/7332, I keep getting invitation modals each time I refresh a page. Accepting or rejecting gives an error: "No quest invitation found" and "No invitation has been sent out yet" respectively.

I haven't tried the Ask Later button yet.

Screenshot: sc
#### Console Errors

404 POST accept https://v3.habitica.com/api/v3/groups/e3565aa7-2917-4460-9716-c930e4fd778d/quests/accept
or
404 POST reject https://v3.habitica.com/api/v3/groups/e3565aa7-2917-4460-9716-c930e4fd778d/quests/reject

and then
304 GET api/v3/user/ https://v3.habitica.com/api/v3/user/
304 GET user https://v3.habitica.com/api/v3/tasks/user

crookedneighbor:
Fixes #7353

This cleans up the RSVPNeeded property when the quest starts. Since the client does a user sync after any quest action, it should correct the value on the client as well.

### DIFF
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -110,6 +110,27 @@ schema.post('remove', function postRemoveGroup (group) {
   firebase.deleteGroup(group._id);
 });
 
+// return a clean object for user.quest
+function _cleanQuestProgress (merge) {
+  let clean = {
+    key: null,
+    progress: {
+      up: 0,
+      down: 0,
+      collect: {},
+    },
+    completed: null,
+    RSVPNeeded: false,
+  };
+
+  if (merge) {
+    _.merge(clean, _.omit(merge, 'progress'));
+    if (merge.progress) _.merge(clean.progress, merge.progress);
+  }
+
+  return clean;
+}
+
 schema.statics.getGroup = async function getGroup (options = {}) {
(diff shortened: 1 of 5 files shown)


website/server/models/group.js
```diff
@@ -110,6 +110,27 @@ schema.post('remove', function postRemoveGroup (group) {
   firebase.deleteGroup(group._id);
 });
 
+// return a clean object for user.quest
```
crookedneighbor, line 113: Just moved this so it would be defined earlier.

PR #7593, opened 2016-06-02
fix tag sorting

### LINKED ISSUE #7541: Tags cannot be sorted / reordered
Tags cannot be sorted. If you change the order of the tags, the next time a sync occurs, they change back.

shervinmathieu:
Fixes #7541
### Changes

A sneaky bug, frustratingly hard to find.
The call to user.ops.sortTag causes the client side tag data to be sorted before the call to the API to update the database. The call to the API then receives the ID of the tag that occupied the slot of the intended tag to be sorted.

Since the tag id cannot be pulled from ui.item.data, I've instead added a variable to the code to store the id if the tag to be sorted.

There is no karma coverage for sorting tags, and I don't know how to write karma tests (maybe I should learn to do so). However, this bug would not have been detected by a karma test either.

---

### DIFF
--- a/website/client/js/services/userServices.js
+++ b/website/client/js/services/userServices.js
@@ -319,7 +319,7 @@ angular.module('habitrpg')
         },
 
         sortTag: function (data) {
-          let fromId = user.tags[data.query.from].id;
+          var fromId = user.tags[data.query.from].id;
           user.ops.sortTag(data);
           Tags.sortTag(fromId, data.query.to);
         },
--- a/website/client/js/services/userServices.js
+++ b/website/client/js/services/userServices.js
@@ -319,8 +319,9 @@ angular.module('habitrpg')
         },
 
         sortTag: function (data) {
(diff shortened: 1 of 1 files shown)


shervinmathieu:
Other fixes can be attempted, for example swapping the 2 functions around so that they do not interfiere with each other, or modifying the API so that it takes the tag position instead of its id as a parameter

website/client/js/services/userServices.js
```diff
@@ -319,8 +319,9 @@ angular.module('habitrpg')
         },
 
         sortTag: function (data) {
+          let fromId = user.tags[data.query.from].id;
```
crookedneighbor: This should be `var`

crookedneighbor:
Looks good other than my one comment. Great investigation!

shervinmathieu:
Done.

crookedneighbor:
Good work. I've notated this toward your next tier!

PR #7594, opened 2016-06-03
Make wheelchairs public

SabreCat:
Fixes an issue reported by Shine Augustine in the Report a Bug guild.
### Changes

Previously, `preferences.chair` was not considered public data, so it was not returned using the GET members/:memberId route. That led in turn to wheelchairs not appearing outside of one's own profile. Now the `preferences.chair` path is marked as public, so it is visible across the site.

---

UUID: [Bulky Platypus Hive](https://map.what3words.com/bulky.platypus.hive)

### DIFF
--- a/test/api/v3/integration/members/GET-members_id.test.js
+++ b/test/api/v3/integration/members/GET-members_id.test.js
@@ -35,7 +35,7 @@ describe('GET /members/:memberId', () => {
     ]);
     expect(Object.keys(memberRes.auth)).to.eql(['timestamps']);
     expect(Object.keys(memberRes.preferences).sort()).to.eql(['size', 'hair', 'skin', 'shirt',
-      'costume', 'sleep', 'background'].sort());
+      'chair', 'costume', 'sleep', 'background'].sort());
   });
 
   it('handles non-existing members', async () => {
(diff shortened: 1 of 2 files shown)


test/api/v3/integration/members/GET-members_id.test.js
```diff
@@ -35,7 +35,7 @@ describe('GET /members/:memberId', () => {
     ]);
     expect(Object.keys(memberRes.auth)).to.eql(['timestamps']);
     expect(Object.keys(memberRes.preferences).sort()).to.eql(['size', 'hair', 'skin', 'shirt',
```
crookedneighbor: You can remove the sort call, the eql check shouldn't care about order.

crookedneighbor:
Other than my one comment, :+1:

test/api/v3/integration/members/GET-members_id.test.js
crookedneighbor: Judging by the build, it looks like I was wrong 😛

crookedneighbor:
Cherry picked the first commit and merge with 7108b02f4df501ea5f3dd2b2efb29c834c5e35cf

PR #7596, opened 2016-06-03
Correctly award 3-month benefits

### LINKED ISSUE #4819: Subscription renewal hourglasses
Hello,

Thank you
## <bountysource-plugin>

</bountysource-plugin>

SabreCat:
Fixes #4819.
### Changes

Previously, the check for whether or not to award 3-, 6-, or 12-month subscribers their new Mystic Hourglass and Gem-to-Gold cap increase would be skipped when their `purchased.plan.consecutive.offset` value was anything above 0 when their subscription was updated. This meant that the first time they should get their benefits--i.e., when the `offset` was 1 then fell to 0--they would receive nothing.

Now, the perk award code handles the case of an `offset` of 1 correctly, by reducing it to 0 then proceeding to give out the Hourglass and Gem cap increase.

---

UUID: [Subtle Toaster Model](https://map.what3words.com/subtle.toaster.model)

### DIFF
--- a/test/api/v3/unit/libs/cron.test.js
+++ b/test/api/v3/unit/libs/cron.test.js
@@ -80,21 +80,25 @@ describe('cron', () => {
     });
 
     it('decrements plan.consecutive.offset when offset is greater than 0', () => {
-      user.purchased.plan.consecutive.offset = 1;
+      user.purchased.plan.consecutive.offset = 2;
       cron({user, tasksByType, daysMissed, analytics});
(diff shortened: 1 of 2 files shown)


crookedneighbor:
I don't fully understand the logic of the payments stuff, but your explanation seems sound. 👍

SabreCat:
This could stand to be refactored further. There's probably a better way to handle "award bennies immediately, then on every subscription interval thereafter" than the `offset` rigmarole. But for now, this fixes a long-running and important problem for our present use cases!

PR #7598, opened 2016-06-03
(WIP) New discount coupon

SabreCat:
### Changes

Adds a new subscription coupon for an upcoming promotion. The configuration is basically identical to the existing `google_6mo` coupon, but is set up so we can track the success of that promotion separately.

---

UUID: [Rower Cool Bouncy](https://map.what3words.com/rower.cool.bouncy)

### DIFF
--- a/scripts/paypalBillingSetup.js
+++ b/scripts/paypalBillingSetup.js
@@ -73,7 +73,7 @@ switch(OP) {
     });
     break;
   case "create":
-    paypal.billingPlan.create(blocks["google_6mo"].definition, function(err,plan){
+    paypal.billingPlan.create(blocks["focusme_6mo"].definition, function(err,plan){
       if (err) return console.log(err);
       if (plan.state == "ACTIVE")
         return console.log({err:err, plan:plan});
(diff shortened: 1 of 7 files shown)


scripts/paypalBillingSetup.js
```diff
@@ -73,7 +73,7 @@ switch(OP) {
     });
     break;
   case "create":
-    paypal.billingPlan.create(blocks["google_6mo"].definition, function(err,plan){
+    paypal.billingPlan.create(blocks["focusme_6mo"].definition, function(err,plan){
```
paglias, line 76: the google_6mo coupon should be removed?

SabreCat, line 76: As far as I know, Google employees can still take advantage of that discount (@lemoness, confirm?), so we're not removing the google_6mo plan.

This particular bit, however, doesn't affect existing plans. It's a bit of script to make a call to the PayPal API to create a new plan. I could've just made the change locally before running the script, but wanted it in the commit for people to see in case I am mistaken about how it works!

paglias, line 76: @SabreCat sounds good!

lemoness, line 76: Quickly confirming that yes, we want to keep the preexisting discount.

paglias:
This file will have to be updated too https://github.com/HabitRPG/habitrpg-email-server/blob/master/workers/amazonPayments.js#L12

SabreCat:
Thanks to some ugly hardcoding on the client, we can't really do this right now. Thankfully, the existing Google coupon type matches what we want for the Focus Me coupon.

PR #7607, opened 2016-06-05
refactor common/script/ops/scoreTask.js

Alys:
### Changes

Refactors `common/script/ops/scoreTask.js` because when I was doing https://github.com/HabitRPG/habitrpg/pull/7370 I found the code very hard to follow.

If you want to review this in detail, it might be easiest to do it one commit at a time because I tried to commit in a logical fashion. The commit comments are also supposed to help with reviewing.

To ensure that no behaviour has changed, we'd need to have tests that looked for precise numbers in user and task stats, but I notice that that's not usually done in the test files. Is that just because it's not been done yet, or is there another reason? I think it would make sense to do it for the `test/common/ops/scoreTask.test.js` tests.
- [ ] Note to myself: must test for the effect of checklists on To-Dos.

### DIFF
--- a/common/script/ops/scoreTask.js
+++ b/common/script/ops/scoreTask.js
@@ -106,45 +106,34 @@ function _gainMP (user, val) {
   }
 }
 
-// HP modifier
-// ===== CONSTITUTION =====
-// TODO Decreases HP loss from bad habits / missed dailies by 0.5% per point.
 function _subtractPoints (user, task, stats, delta) {
-  let conBonus = 1 - user._statsComputed.con / 250;
+  let conBonus = 1 - user._statsComputed.con / 250; // CON decreases HP loss
   if (conBonus < 0.1) conBonus = 0.1;
 
   let hpMod = delta * conBonus * task.priority * 2; // constant 2 multiplier for better results
-  stats.hp += Math.round(hpMod * 10) / 10; // round to 1dp
+  stats.hp += Math.round(hpMod * 10) / 10;
   return stats.hp;
 }
 
 function _addPoints (user, task, stats, direction, delta) {
-  // ===== CRITICAL HITS =====
   // allow critical hit only when checking off a task, not when unchecking it:
-  let _crit = delta > 0 ? crit(user) : 1;
+  let _crit = direction === 'up' ? crit(user) : 1;
   // if there was a crit, alert the user via notification
   if (_crit > 1) user._tmp.crit = _crit;
 
-  // Exp Modifier
-  // ===== Intelligence =====
-  // TODO Increases Experience gain by .2% per point.
-  let intBonus = 1 + user._statsComputed.int * 0.025;
+  let intBonus = 1 + user._statsComputed.int * 0.025; // INT increases XP gain
   stats.exp += Math.round(delta * intBonus * task.priority * _crit * 6);
 
-  // GP modifier
-  // ===== PERCEPTION =====
-  // TODO Increases Gold gained from tasks by .3% per point.
-  let perBonus = 1 + user._statsComputed.per * 0.02;
+  let perBonus = 1 + user._statsComputed.per * 0.02; // PER increases GP gain
   let gpMod = delta * task.priority * _crit * perBonus;
 
-  if (task.streak) {
+  if (task.streak) {  // streak increases GP gain
     let currStreak = direction === 'down' ? task.streak - 1 : task.streak;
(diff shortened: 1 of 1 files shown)


common/script/ops/scoreTask.js
```diff
@@ -245,14 +241,10 @@ module.exports = function scoreTask (options = {}, req = {}) {
     }
   } else if (task.type === 'reward') {
     // Don't adjust values for rewards
-    delta += _changeTaskValue(user, task, direction, times, cron);
+    // If they're trying to purchase a too-expensive reward, don't allow them to do that.
+    if (task.value > user.stats.gp) throw new NotAuthorized(i18n.t('messageNotEnoughGold', req.language));
     // purchase item
     stats.gp -= Math.abs(task.value);
-    // hp - gp difference
-    if (stats.gp < 0) {
```
paglias, line 252: we do want to remove this?

common/script/ops/scoreTask.js
```diff
@@ -141,27 +157,34 @@ function _changeTaskValue (user, task, direction, times, cron) {
...
   // If multiple days have passed, multiply times days missed
   _.times(times, () => {
-    // Each iteration calculate the nextDelta, which is then accumulated in the total delta.
-    let nextDelta = !cron && direction === 'down' ? _calculateReverseDelta(task, direction) : _calculateDelta(task, direction, cron);
-
-    if (user.preferences.automaticAllocation === true && user.preferences.allocationMode === 'taskbased' && !(task.type === 'todo' && direction === 'down')) {
-      user.stats.training[task.attribute] += nextDelta;
-    }
+    // Each iteration calculates a change in task value, which is then accumulated in the total delta.
+    addToDelta += !cron && direction === 'down' ? _calculateReverseDelta(task) : _calculateDelta(task, direction, cron);
```
paglias, line 159: since _calculateDelta has always direction === 'up', the `direction` parameter should be removed?

common/script/ops/scoreTask.js
Alys, line 252: Definitely. It hasn't actually been a feature for a long time because code further up in the file stops execution when the user doesn't have enough gold to buy a Reward, and if we reintroduced this feature, we'd get confused users complaining about their health decreasing when Rewards were purchased. It's not a good idea any more. That block is just ancient, unused code that should have been cleaned up years ago. :)

crookedneighbor:
@Alys when you get a chance, could you bring this up to date?

Alys:
Sure. 

Do you have any comments about my questions about tests?

common/script/ops/scoreTask.js
```diff
@@ -90,79 +109,80 @@ function _gainMP (user, val) {
...
 
-// HP modifier
-// ===== CONSTITUTION =====
-// TODO Decreases HP loss from bad habits / missed dailies by 0.5% per point.
+// Decrease user's health when:
+// - user does negative Habits
+// - cron finds uncompleted Dailies
 function _subtractPoints (user, task, stats, delta) {
-  let conBonus = 1 - user._statsComputed.con / 250;
+  let conBonus = 1 - user._statsComputed.con / 250; // CON decreases HP loss
```
khipkin, line 116: Make this `* 0.004` instead.  As-is, it doesn't fit with the way `intBonus` and `perBonus` are calculated (below).

khipkin, line 116: Also, changing this will make it easy to pull out the bonus-per-point values for con, int, and per and add them as `consts` at the top of the file.

common/script/ops/scoreTask.js
```diff
@@ -223,35 +240,37 @@ module.exports = function scoreTask (options = {}, req = {}) {
...
     stats.gp -= Math.abs(task.value);
-    // hp - gp difference
-    if (stats.gp < 0) {
-      stats.hp += stats.gp;
-      stats.gp = 0;
+  }
+
+  // User gains / loses MP for their 'up' / 'down' actions on tasks
+  // Must be done after other stat changes, because INT affects maxMP
+  // TODO refactor - put all/most in _gainMP; pass in task / task.type; extract numbers out to variables
```
khipkin, line 264: I agree that these numbers should be extracted out. :+1:

Alys:
I am still working on the tests for this but it's taking a while because I want to do a lot, and I also need to find some way to avoid random critical hits from breaking the tests (if anyone has any ideas about that, I'd love to hear them!). I'll reopen this when it's ready to be reviewed again.

PR #7612, opened 2016-06-05
change user quest.progress.collect to quest.progress.collectedItems

Alys:
@crookedneighbor

### DIFF
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -117,9 +117,7 @@ function _cleanQuestProgress (merge) {
     progress: {
       up: 0,
       down: 0,
-      collect: {type: Schema.Types.Mixed, default: () => {
-        return {};
-      }},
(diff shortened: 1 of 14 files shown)


website/server/models/group.js
```diff
@@ -117,7 +117,10 @@ function _cleanQuestProgress (merge) {
     progress: {
       up: 0,
       down: 0,
-      collect: 0,
+      collect: {type: Schema.Types.Mixed, default: () => {
```
crookedneighbor: This is wrong. Leaving a note to elaborate in a minute.

crookedneighbor: This should be changed to:

``` js
collect: {},
collectedItems: 0,
```

Alys: Thanks! Fixed.

PR #7622, opened 2016-06-08
Fix webhooks

librarianmage:
Fixes https://github.com/HabitRPG/habitrpg/issues/7481

### DIFF
--- a/common/script/ops/addWebhook.js
+++ b/common/script/ops/addWebhook.js
@@ -10,7 +10,7 @@ module.exports = function addWebhook (user, req = {}) {
   let wh = user.preferences.webhooks;
 
   if (!validator.isURL(_.get(req, 'body.url'))) throw new BadRequest(i18n.t('invalidUrl', req.language));
-  if (!validator.isBoolean(_.get(req, 'body.enabled'))) throw new BadRequest(i18n.t('invalidEnabled', req.language));
+  if (!validator.isBoolean(_.get(req, 'body.enabled') + '')) throw new BadRequest(i18n.t('invalidEnabled', req.language));
 
   user.markModified('preferences.webhooks');
 
--- a/common/script/ops/addWebhook.js
+++ b/common/script/ops/addWebhook.js
@@ -10,7 +10,7 @@ module.exports = function addWebhook (user, req = {}) {
   let wh = user.preferences.webhooks;
 
   if (!validator.isURL(_.get(req, 'body.url'))) throw new BadRequest(i18n.t('invalidUrl', req.language));
-  if (!validator.isBoolean(String(_.get(req, 'body.enabled')))) throw new BadRequest(i18n.t('invalidEnabled', req.language));
+  if (!validator.isBoolean(_.get(req, 'body.enabled'))) throw new BadRequest(i18n.t('invalidEnabled', req.language));
 
   user.markModified('preferences.webhooks');
 
--- a/common/script/ops/addWebhook.js
+++ b/common/script/ops/addWebhook.js
@@ -10,7 +10,7 @@ module.exports = function addWebhook (user, req = {}) {
   let wh = user.preferences.webhooks;
 
   if (!validator.isURL(_.get(req, 'body.url'))) throw new BadRequest(i18n.t('invalidUrl', req.language));
-  if (!validator.isBoolean(_.get(req, 'body.enabled') + '')) throw new BadRequest(i18n.t('invalidEnabled', req.language));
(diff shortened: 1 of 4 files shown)


librarianmage:
Ooops @paglias! Just realized you created a PR for the same thing.

common/script/ops/addWebhook.js
```diff
@@ -10,7 +10,7 @@ module.exports = function addWebhook (user, req = {}) {
   let wh = user.preferences.webhooks;
 
   if (!validator.isURL(_.get(req, 'body.url'))) throw new BadRequest(i18n.t('invalidUrl', req.language));
-  if (!validator.isBoolean(_.get(req, 'body.enabled'))) throw new BadRequest(i18n.t('invalidEnabled', req.language));
+  if (!validator.isBoolean(String(_.get(req, 'body.enabled')))) throw new BadRequest(i18n.t('invalidEnabled', req.language));
```
crookedneighbor: Why are you wrapping this in a string object?

librarianmage: Validate.js has deprecated testing anything but strings (it warned me while
testing), so to be preemptive I changed it before it became an error.

crookedneighbor: Gotcha. What a weird quirk.

paglias: Yeah I tried to upgrade to version 5 but this blocked me. You cannot even pass `undefined` or `null` as it refuses to validate them...

website/client/js/services/userServices.js
```diff
@@ -151,7 +151,7 @@ angular.module('habitrpg')
         $http({
           method: method,
           url: url + queryString,
-          body: body,
```
paglias, line 154: good catch!

paglias, line 154: Hmm actually I'm not sure about this.. I mean, it makes sense but it's used for a lof of ops and we should have noticed if they didn't work.. @TheHollidayInn

paglias:
@MathWhiz thanks for the PR, my PR also had several other fixes, I'll probably cherry pick some parts of this and merge it in my own.

website/client/js/services/userServices.js
```diff
@@ -512,15 +512,15 @@ angular.module('habitrpg')
         },
 
         addWebhook: function (data) {
-          callOpsFunctionAndRequest('addWebhook', 'webhook', "POST", '', data, data.body);
```
paglias, line 515: I think data.body is correct here as `{body: {}}` will be passed as `data` and we only want to send data.body to the server

paglias:
Don't worry about updating this PR, sorry for the confusion, I meant to merge my own sooner but got distracted by other things

librarianmage:
@paglias For some reason, body was not being sent on the request, and
renaming it to data fixed the problem

website/client/js/services/userServices.js
```diff
@@ -512,15 +512,15 @@ angular.module('habitrpg')
...
+          callOpsFunctionAndRequest('addWebhook', 'webhook', "POST", '', data);
```
paglias, line 515: @MathWhiz you were right, passing `data` is the way to go!

librarianmage, line 515: @paglias Ok! I was confused that `data.body` was being passed in, considering `callOpsFunctionAndRequest` only takes 5 args.

website/client/js/services/userServices.js
librarianmage, line 154: I noticed that it needed to be `data` instead of `body` when looking up [$http](https://docs.angularjs.org/api/ng/service/$http#usage) on AngularJS

website/client/js/controllers/settingsCtrl.js
```diff
@@ -247,7 +247,7 @@ habitrpg.controller('SettingsCtrl',
       $scope.hasWebhooks = _.size(webhooks);
     })
     $scope.addWebhook = function(url) {
-      User.addWebhook({body:{url:url, id:Shared.uuid()}});
```
paglias, line 250: let's keep the id here

paglias:
@MathWhiz okay, if you remove the coercion to String (going to do that when / if we upgrade) I'll merge this!

librarianmage:
@paglias Done!

paglias:
@MathWhiz thanks! UUID?

crookedneighbor:
No need for the UUID, I've notated it toward your next tier.

librarianmage:
@crookedneighbor @paglias

PR #7624, opened 2016-06-08
allow "Your password" to be translated in account deletion confirmation

Alys:
When you're deleting your account, the confirmation screen instructs you to enter your password, but the phrase "Your password" is untranslatable. Example:

screenshot from 2016-06-08 21-10-15

This code that does that is left over from before API v3 when you had to type the word "DELETE" into the confirmation box but there's no need for that any more.

This PR removes the special handling for "Your password" and makes it a normal part of the `deleteText` locale string. 

@GitHubSphinx This PR removes the old `deleteText` locale string and adds a new `deleteAccountText` string. They're nearly identical so translators could copy the old translation, but will that still be possible in Transifex when this PR is merged? Should I adjust it to leave the old string in place until later?

### DIFF
--- a/common/locales/en/settings.json
+++ b/common/locales/en/settings.json
@@ -63,7 +63,7 @@
     "dangerZone": "Danger Zone",
     "resetText1": "WARNING! This resets many parts of your account. This is highly discouraged, but some people find it useful in the beginning after playing with the site for a short time.",
     "resetText2": "You will lose all your levels, gold, and experience points. All your tasks (except those from challenges) will be deleted permanently and you will lose all of their historical data. You will lose all your equipment but you will be able to buy it all back, including all limited edition equipment or subscriber Mystery items that you already own (you will need to be in the correct class to re-buy class-specific gear). You will keep your current class and your pets and mounts. You might prefer to use an Orb of Rebirth instead, which is a much safer option and which will preserve your tasks.",
-    "deleteText": "Are you sure? This will delete your account forever, and it can never be restored! You will need to register a new account to use Habitica again. Banked or spent Gems will not be refunded. If you're absolutely certain, type <strong><%= deleteWord %></strong> into the text box below.",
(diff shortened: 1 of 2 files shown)


ghost:
@Alys thank you for the fix and for thinking about translators! I'd apply
the changes replacing (deleting) the previous strings: previous
translations won't be available in search directly on Transfix in this
case, but we'll (a) avoid confusion of "why are both version still there
and what's the difference in use", (b) Trsnsifex translation memory should
provide suggestions showing previous similar strings (may not work for long
strings), (c) translators can now use the output of your dif script if
needed.

common/locales/en/settings.json
```diff
@@ -63,7 +63,7 @@
...
-    "deleteText": "Are you sure? This will delete your account forever, and it can never be restored! You will need to register a new account to use Habitica again. Banked or spent Gems will not be refunded. If you're absolutely certain, type <strong><%= deleteWord %></strong> into the text box below.",
+    "deleteAccountText": "Are you sure? This will delete your account forever, and it can never be restored! You will need to register a new account to use Habitica again. Banked or spent Gems will not be refunded. If you're absolutely certain, type your password into the text box below.",
```
crookedneighbor: Could this be `"deleteLocalAccountText"`? I ask, because we have an outstanding issue for deleting a facebook based account, where there is no password to input. And that may require a separate string.

crookedneighbor:
I'm 👍 on changing the key. If I recall, Transifex _should_ suggest similar translations in the history, even if the other key is deleted.

ghost:
@crookedneighbor Yes, sorry for possible confusion. By "replace" I meant "feel free to delete the old one" not "please reuse the same key"

common/locales/en/settings.json
Alys: Certainly! That's done and pushed.

PR #7627, opened 2016-06-08
Turn automaticAllocation to false when resetting or rebirthing

### LINKED ISSUE #7620: Automatic allocation (task-based, etc) continues working for first ten levels after Orb or Rebirth or Reset
General Info
- UUID: 2d6ef231-50b4-4a22-90e7-45eb97147a2c
- Browser: Firefox 46.0.1
- OS: Windows 7
### Description

I rebirthed with task-based allocation on but intended to turn it off before level 10 when I should have gotten my first batch of attribute points. Then I remembered/found out I couldn't because that section of the stats page is inaccessible before 10. So I waited for level 10, and when I got there I went to turn it off and noticed I had nine points allocated (6 INT, 1 CON, 2 PER) with no unallocated points.This is strange for two reasons:
1. I don't recall it happening at any time (especially not level 10) during the v3 tests when I rebirthed with allocation on there. I know the site isn't entirely equivalent to the v3 tests, but I thought this important to bring up because it was the only other time I'd leveled to ten with allocation on.
2. I'm pretty sure from the past that I should have 10 (or 11?) points to allocate which means one is missing. Theoretically, that would mean it would give me the extra points to allocate next level (I've turned allocation off), but it's strange they weren't already granted...

I'm also a little bummed because I probably would have distributed those differently, but... eh, I can wait until the next time I class cycle during the summer festival.

I said in the Report a Bug guild that (from my inexperienced perspective) it's looks like the auto-allocation kept working after rebirth but didn't grant a point for the first level (for whatever rebirth-related reasons that could have happened). Alys asked me to report here.
#### Console Errors

Nothing.

librarianmage:
Fixes #7620
### Changes

Sets `user.preferences.automaticAllocation` to false when rebirthing or resetting.

---

### DIFF
--- a/common/script/ops/rebirth.js
+++ b/common/script/ops/rebirth.js
@@ -51,6 +51,9 @@ module.exports = function rebirth (user, tasks = [], req = {}, analytics) {
   stats.hp = 50;
   stats.lvl = 1;
   stats.class = 'warrior';
+  
+  let preferences = user.preferences;
+  preferences.automaticAllocation = false;
 
   _.each(USERSTATSLIST, function resetStats (value) {
     stats[value] = 0;
--- a/common/script/ops/rebirth.js
+++ b/common/script/ops/rebirth.js
@@ -51,9 +51,8 @@ module.exports = function rebirth (user, tasks = [], req = {}, analytics) {
   stats.hp = 50;
   stats.lvl = 1;
   stats.class = 'warrior';
-  
-  let preferences = user.preferences;
-  preferences.automaticAllocation = false;
+
+  user.preferences.automaticAllocation = false;
 
   _.each(USERSTATSLIST, function resetStats (value) {
     stats[value] = 0;
--- a/test/api/v3/integration/user/POST-user_reset.test.js
+++ b/test/api/v3/integration/user/POST-user_reset.test.js
@@ -101,4 +101,30 @@ describe('POST /user/reset', () => {
 
     expect(userChallengeTask).to.eql(task);
   });
+
+  it('keeps automaticAllocation false', async () => {
+    await user.update({
+      preferences: {
+        automaticAllocation: false,
+      },
+    });
+
+    await user.post('/user/reset');
+    await user.sync();
+
+    expect(user.preferences.automaticAllocation).to.be.false;
+  });
+
+  it('sets automaticAllocation to false when true', async () => {
+    await user.update({
+      preferences: {
+        automaticAllocation: true,
+      },
+    });
+
+    await user.post('/user/reset');
+    await user.sync();
+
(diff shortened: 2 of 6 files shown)


crookedneighbor:
Needs tests.

librarianmage:
Not sure how to do that, but I'll try!

common/script/ops/rebirth.js
```diff
@@ -51,6 +51,9 @@ module.exports = function rebirth (user, tasks = [], req = {}, analytics) {
   stats.hp = 50;
   stats.lvl = 1;
   stats.class = 'warrior';
+  
+  let preferences = user.preferences;
+  preferences.automaticAllocation = false;
```
crookedneighbor: Since you're only changing one property, creating a variable is not needed.

crookedneighbor:
there should be some testing around these functions already. First step will be to fix your linting errors.

librarianmage:
Ok!

librarianmage:
@crookedneighbor Done!

test/api/v3/integration/user/POST-user_reset.test.js
```diff
@@ -101,4 +101,30 @@ describe('POST /user/reset', () => {
 
     expect(userChallengeTask).to.eql(task);
   });
+
+  it('keeps automaticAllocation false', async () => {
+    await user.update({
+      preferences: {
+        automaticAllocation: false,
```
paglias: should be true? to check that later is changed?

paglias: Hmm I see you made two different tests... okay

paglias:
@MathWhiz the tests are okay but it would be better if you could add them to the common code tests https://github.com/HabitRPG/habitrpg/blob/develop/test/common/ops/reset.js. The integration ones are more to make sure that the common code is correctly called and handled

crookedneighbor:
I agree.

librarianmage:
Oh! I will move them

librarianmage:
@crookedneighbor The tests have beed moved!

test/common/ops/rebirth.js
```diff
@@ -233,4 +233,20 @@ describe('shared.ops.rebirth', () => {
     expect(user.achievements.rebirths).to.equal(2);
     expect(user.achievements.rebirthLevel).to.equal(MAX_LEVEL);
   });
+
+  it('keeps automaticAllocation false', async () => {
```
crookedneighbor, line 237: No need for this to be async. There's no async behavior happening in the test.

crookedneighbor:
I removed the async keywords from the tests and squashed the commits.

PR #7631, opened 2016-06-09
prevent sync from clearing selected tags

### LINKED ISSUE #7413: Clicking sync (or doing other activities) removes selected tags in the tag filter
_edit by admin: This also occurs when clicking sync._
# System

UUID: fedc44cb-088a-4bc8-8f4f-3e6a1882f042
Browsers: Safari, FF, Chrome
OS: OS X 10.11.4

Doing multiple activities in short succession -- e.g., adding several new tasks or completing several tasks in a row, without waiting for the response to come back from Habitica -- causes the display to re-render/refresh. The refresh is expected behavior; however when the refresh occurs it unselects any tags that were previously selected in the tag filter at the top of the page. This deselection is new and unwanted behavior.
# How to reproduce

1) Clear the cache and reload the page.
2) Select multiple tags in the tag filter at the top.

screen shot 2016-05-22 at 3 35 27 am

3) Create three dummy tasks

screen shot 2016-05-22 at 3 36 59 am

4) Check them off in short succession, so that the response from Habitica does not have time to return before the next one is clicked. (This may take a few tries to reproduce a lag, or trying other close-succession behaviors like Brutal Smashing several tasks in a row)

5) Notice that the display re-renders/refreshes (expected behavior). Notice that tags that were previously selected are now unselected (new, unexpected and undesired behavior).

screen shot 2016-05-22 at 3 44 22 am
## Console Errors

None, but there are unrelated warnings that show up.
screen shot 2016-05-22 at 3 21 12 am
screen shot 2016-05-22 at 3 22 18 am

shervinmathieu:
Fixes #7413
### Changes

These are the changes pointed out by @paglias to fix the unintended clearing of selected tags when a sync occurs, plus one additional change to prevent user.filters from becoming permanently undefined, which breaks tag functionality.

---

### DIFF
--- a/website/client/js/controllers/filtersCtrl.js
+++ b/website/client/js/controllers/filtersCtrl.js
@@ -25,9 +25,6 @@ habitrpg.controller("FiltersCtrl", ['$scope', '$rootScope', 'User', 'Shared',
     };
 
     $scope.toggleFilter = function(tag) {
-      if (user.filters === undefined) {
-        user.filters = {};
-      }
       if (!user.filters[tag.id]) {
         user.filters[tag.id] = true;
       } else {
--- a/website/client/js/controllers/filtersCtrl.js
+++ b/website/client/js/controllers/filtersCtrl.js
@@ -25,6 +25,9 @@ habitrpg.controller("FiltersCtrl", ['$scope', '$rootScope', 'User', 'Shared',
     };
 
     $scope.toggleFilter = function(tag) {
+      if (user.filters === undefined) {
+        user.filters = {};
+      }
       if (!user.filters[tag.id]) {
         user.filters[tag.id] = true;
       } else {
--- a/website/server/models/user/hooks.js
+++ b/website/server/models/user/hooks.js
@@ -13,7 +13,7 @@ schema.plugin(baseModel, {
   private: ['auth.local.hashed_password', 'auth.local.salt', '_cronSignature'],
   toJSONTransform: function userToJSON (plainObj, originalDoc) {
     plainObj._tmp = originalDoc._tmp; // be sure to send down drop notifs
-    plainObj.filters = undefined;
+    delete plainObj.filters;
 
     return plainObj;
   },
--- a/website/server/models/user/hooks.js
+++ b/website/server/models/user/hooks.js
@@ -13,6 +13,7 @@ schema.plugin(baseModel, {
   private: ['auth.local.hashed_password', 'auth.local.salt', '_cronSignature'],
   toJSONTransform: function userToJSON (plainObj, originalDoc) {
     plainObj._tmp = originalDoc._tmp; // be sure to send down drop notifs
+    plainObj.filters = undefined;
 
     return plainObj;
   },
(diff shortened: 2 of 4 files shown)


website/client/js/controllers/filtersCtrl.js
```diff
@@ -25,6 +25,9 @@ habitrpg.controller("FiltersCtrl", ['$scope', '$rootScope', 'User', 'Shared',
     };
 
     $scope.toggleFilter = function(tag) {
+      if (user.filters === undefined) {
```
paglias: the PR looks good! Maybe `if (!user.filter) user.filters = {}` should be added here https://github.com/HabitRPG/habitrpg/blob/develop/website/client/js/services/userServices.js#L82 right after `_.extend` instead than in `toggleFilter`. So if something else relies on `user.filters` it will work even if `toggleFilter` is not called before

crookedneighbor: Agreed.

website/server/models/user/hooks.js
```diff
@@ -13,6 +13,7 @@ schema.plugin(baseModel, {
   private: ['auth.local.hashed_password', 'auth.local.salt', '_cronSignature'],
   toJSONTransform: function userToJSON (plainObj, originalDoc) {
     plainObj._tmp = originalDoc._tmp; // be sure to send down drop notifs
+    plainObj.filters = undefined;
```
crookedneighbor: If filters is removed from the model, this should be unnecessary, right?

website/server/models/user/schema.js
```diff
@@ -108,9 +108,9 @@ let schema = new Schema({
 
   balance: {type: Number, default: 0},
   // Not saved on the user right now
-  filters: {type: Schema.Types.Mixed, default: () => {
-    return {};
-  }},
+  // filters: {type: Schema.Types.Mixed, default: () => {
```
crookedneighbor: @paglias do we even need this here if the filters property is client side only?

crookedneighbor: I mean to say, why have it commented out? Why not just delete it?

paglias: @crookedneighbor no we don't. I asked for this to be commented out because we always left it in the code in case we decided to start persisting them again... but yeah it can be removed

website/server/models/user/hooks.js
paglias: Not really because some users may have an empty object saved and while mongoose doesn't save new data that is not in the database it keeps anything that is in mongo but not in the model

crookedneighbor: I see. Personally, I prefer `delete plainObject.filters` over setting it to `undefined`. But I'm fine either way.

shervinmathieu: If filters was deleted, would `if (!user.filters)` still evaluate to true?

paglias: @velezsarain yes, in the client `user.filters` doesn't exists anyway because JSON doesn't have an `undefined` type and so it gets discarded anyway

shervinmathieu:
I've updated the code with your input.

paglias:
@velezsarain thanks! I've notated this in your list of contributions

PR #7638, opened 2016-06-10
feat: Add shortName property to task model

crookedneighbor:
Adds a `shortName` property to tasks that can be used to get the task or score it. This `shortName` value must be unique to the user. 

This allows tasks to have the v2-like ability to be named human readable things like `productivity` 

Left to do:
- [x] Determine if we really need to prevent updating short-name after the fact
- [x] Add short name field with explanation to client
- [x] Add locales for messages
- [x] Disallow short names in challenge task creation and updating


crookedneighbor:
Adds a `shortName` property to tasks that can be used to get the task or score it. This `shortName` value must be unique to the user. 

This allows tasks to have the v2-like ability to be named human readable things like `productivity`

test/common/ops/updateTask.js
```diff
@@ -24,6 +24,7 @@ describe('shared.ops.updateTask', () => {
         text: 'updated',
         id: '123',
         _id: '123',
+        shortName: 'short-name',
```
paglias: here you should provide a different short name and check that it isn't updated

paglias: ignore it, sorry

website/server/controllers/api-v3/tasks.js
```diff
@@ -16,6 +16,27 @@ import logger from '../../libs/api-v3/logger';
 
 let api = {};
 
+async function _checkShortNameUniqueness (user, tasks) {
+  let shortNames = tasks.map(task => task.shortName).filter(name => name);
+
+  if (shortNames.length !== [...new Set(shortNames)].length) {
```
paglias, line 26: why two checks? one with local data and one with mongodb query

paglias, line 26: you could condense them in a single mongodb query that matches the provided short name

paglias, line 26: it took me a moment to understand what the `Set` was used for :) A comment here would help

paglias:
I was thinking that we could re-use the existing routes and if `validator.isUUID(req.params.taskId)` -> search by `_id`, otherwise search by `shortName`. We would just have to disallow UUIDs for short names

website/server/controllers/api-v3/tasks.js
crookedneighbor, line 26: Yeah, the local data checks against creating multiple tasks with the same shortName. There's no way to catch that with a mongo query, since the tasks haven't been saved yet.

I'll leave a comment about Set.

crookedneighbor:
Sounds good to me. That would definitely simplify the logic. I'll make those changes.

crookedneighbor:
We also need to prevent shortNames from being applied to challenge tasks. (the original challenge task, not the user owned one)

test/api/v3/integration/tasks/PUT-tasks_id.test.js
```diff
@@ -176,6 +177,42 @@ describe('PUT /tasks/:id', () => {
...
+
+    it('can set a shortName if the shortName does not already exist', async () => {
+      let savedDaily = await user.put(`/tasks/${daily._id}`, {
+        shortName: 'short-name',
+      });
+
+      expect(savedDaily.shortName).to.eql('short-name');
+    });
+
+    it('ignores shortName if it is already set', async () => {
```
crookedneighbor, line 189: The behavior I decided on was to just ignore the shortName property if it is already set, rather than error. 

Ultimately, I still think it should be fine to update it.

paglias, line 189: There's a part of me that only wants a UUID / shortName to be usable for a single task. So if you delete that task you won't be able to recreate a new one with the same shortName but I see how it can be a little extreme

crookedneighbor, line 189: Can you explain _why_ you want that?

The way I see it, the short name is like the url of a blog post. Something that is human readable and unique to the post. If you _do_ change it, any link to the post will be broken, just like any integrations that use the short name will be broken. So we should provide warnings to that effect on the website when someone tries to change it, but not totally prevent them from doing so. 

I can see someone making a simple typo that they want to correct, and being unable to do so. I'd find that very frustrating.

If you feel very strongly about it, I'm willing to hear why, and ultimately, I'll do what you want to do in this regard, but it seems very strange to me to put that limit on it.

paglias, line 189: because I see it as a unique identifier similar to the UUID: it should always point to the same resource. If I have an integration that points to the task with short name `my-integration` I'd like to know that it would always point to that task and that if I use it one year later it'll always be an habit and not have been changed to a daily for example

crookedneighbor, line 189: But changing the short name would be _completely_ in your control. It's not like you would have to keep an eye on it to make sure it _didn't_ change. You would have to complete the action to make it change.

Here's the use case I'm thinking of.

Let's say I use [GitHabit](http://habitica.wikia.com/wiki/GitHabit_-_GitHub_Webhooks) to score a habit I've given the shortName `github` every time I push commits to Github. This is great. But let's say that at some point in the future, I want to add my Gitlab and Bitbucket accounts to do the same thing on the same habit. I go to try and change the shortName to `git` or `git-score` or something, but I cannot. I now have 2 choices:

1) Delete `github` entirely and create a new habit, thus losing my history
2) Continue using `github` even though it doesn't accurately reflect what it encompasses 

All this because my needs changed slightly.

I don't think it hurts anything to allow the user to change it, especially if we warn them of the consequences before doing so.

crookedneighbor, line 189: That said, if we allowed short names to be edited up front and then changed our minds and decided that they can't be edited, that would be a breaking change. But deciding that they can't be edited and then allowing them to be edited later, that would be a new feature and not a breaking change. 

So if you're still not convinced, I'll just carry on with the version of this feature where you can only set it once. (And hope that eventually you come around 😛  ) 

Ultimately, this is intended for use by very simple integrations that require id + api-token + short-name, so I want to make this as easy for users as possible.

paglias, line 189: Okay, I'm convinced :)

crookedneighbor:
screen shot 2016-06-12 at 7 38 51 am

website/server/controllers/api-v3/tasks.js
```diff
@@ -12,10 +12,34 @@ import {
 import common from '../../../../common';
 import Bluebird from 'bluebird';
 import _ from 'lodash';
+import validator from 'validator';
 import logger from '../../libs/api-v3/logger';
 
 let api = {};
 
+async function _checkShortNameUniqueness (user, tasks, res) {
+  let shortNames = tasks.map(task => task.shortName).filter(name => name);
```
khipkin, line 21: Why is the `filter` needed?

website/server/controllers/api-v3/tasks.js
```diff
@@ -374,19 +407,24 @@ api.scoreTask = {
...
 
     let user = res.locals.user;
-    let direction = req.params.direction;
+    let {taskId} = req.params;
 
-    let task = await Tasks.Task.findOne({
-      _id: req.params.taskId,
-      userId: user._id,
-    }).exec();
+    let taskQuery = { userId: user._id};
```
khipkin, line 418: asymmetrical spacing

website/server/models/task.js
```diff
@@ -25,6 +26,13 @@ export let TaskSchema = new Schema({
   type: {type: String, enum: tasksTypes, required: true, default: tasksTypes[0]},
   text: {type: String, required: true},
   notes: {type: String, default: ''},
+  shortName: {
+    type: String,
+    validate: [function validateShortName (val) {
+      return validator.toBoolean(this.userId) && !validator.isUUID(val) && val.match(SHORT_NAME_VALIDATION_REGEX);
+    },
+    'shortName can only contain alphanumeric characters, underscores and dashses. They must be owned by a user.'],
```
khipkin, line 34: "dashes"

Also, ambiguous pronoun "they".  The task?

website/server/controllers/api-v3/tasks.js
crookedneighbor, line 21: Because it drops any falsy values. So undefined and empty string get removed.

website/server/controllers/api-v3/tasks.js
crookedneighbor, line 418: Good catch. I'm surprised our linter isn't configured for this.

PR #7649, opened 2016-06-11
fix training values not resetting at level up

### LINKED ISSUE #7542: In automatic task-based attribute point allocation, training points not reset to zero after leveling up
General Info
- UUID: bc612705-7cb4-4f3e-ba39-0871980b1a11
- Browser: Google Chrome (Version 50.0.2661.102 m)
- OS: Windows 7 
### Description

I'm using the automatic allocation mode for my attribute points, based on the tasks I've performed.I'm also using the official user data display tool to track my training points. 
From my experience so far, only the training points for the attribute of the last task I performed before leveling up is resetted to zero when I level up. The attribute point is allocated to the attribute with the most training points. 

_**Example :** 
Training points at level 19 : 
CON 16.7
STR 7.2
PER 15.4
INT 33.2 

Attribute point allocation : 
STR 4
INT 5
CON 3
PER 7

I do a physical task and I level up. New data is : 
Training points at level 20 : 
CON 16.7
STR 0
PER 15.4
INT 33.2 

Attribute point allocation : 
STR 4
INT 6
CON 3
PER 7

The point has correctly been attributed to Intelligence, where I have the most training points, but only training points for Strength have been reset. _

From my understanding, this is a bug, since everything should be reset to zero when I level up.

shervinmathieu:
Fixes #7542
### Changes

I am not sure why the previous code needed to be fixed, or why this is a fix, but in my test environment all values are reset to zero upon level up as expected.

This is curious since at least one value is reset in the current production version, so I can't figure out why the others wouldn't be.

---

### DIFF
--- a/common/script/fns/autoAllocate.js
+++ b/common/script/fns/autoAllocate.js
@@ -56,11 +56,10 @@ function getStatToAllocate (user) {
     case 'taskbased': {
       suggested = _.invert(statsObj.training)[_.max(statsObj.training)];
 
-      let training = statsObj.training;
-      training.str = 0;
-      training.int = 0;
-      training.con = 0;
-      training.per = 0;
+      user.stats.training.str = 0;
+      user.stats.training.int = 0;
+      user.stats.training.con = 0;
+      user.stats.training.per = 0;
 
       return suggested || 'str';
     }


crookedneighbor:
This should include a test that illustrates the problem.

crookedneighbor:
I've adjusted the tests to account for this behavior. Notated toward your next tier.

shervinmathieu:
Sorry about the delay, I would have done it myself, but I've been very busy these days, not much time to use my computer.

crookedneighbor:
No worries. Turns out that test file was in a pretty confusing state, so I rewrote it anyway.

Issue #7636, opened 2016-06-10
Add "Equipment Type" to Equipment Sorting


lemoness:
To match the mobile apps, there should be an "Equipment Type" sorting option based on the body location of the item (helm, armor, etc).

Hus274:
I can give this a shot

PR #7651, opened 2016-06-11
Created new grouping by function for equipment page. #7636

Hus274:
#7636
### Changes

This commit creates a new button on the equipment page that allows the user to group their equipment type (body location) instead of simply by class.

screenshot from 2016-06-11 14-22-03

---

This commit creates a new button on the equipment page that allows the user to group their equipment type (body location) instead of simply by class.


khipkin:
@Hus274 Should there be a matching button on the "Battle Gear" side of the page?  I think it would be useful over there as well.

common/locales/en/gear.json
```diff
@@ -1,7 +1,10 @@
 {
   "set": "Set",
+  "equipmentType" : "Equipment Type",
+  "klazz": "Class",
+  "groupingBy": "Grouping By ",
```
khipkin: Extra space at the end?

website/client/js/controllers/inventoryCtrl.js
```diff
@@ -54,11 +54,16 @@ habitrpg.controller("InventoryCtrl",
 
     $scope.$watch('user.items.gear', function(gear){
       $scope.gear = {};
+      $scope.gearByType = {};
       _.each(gear.owned, function(v,key){
         if (v === false) return;
         var item = Content.gear.flat[key];
-        if (!$scope.gear[item.klass]) $scope.gear[item.klass] = [];
+        if (!$scope.gear[item.klass])
```
khipkin: You should put brackets around the consequences of the `if` block, even if it's just one line.  That way, if more logic is added at a later date, it won't be accidentally outside of the logic block.

Still though, not everyone cares about this guideline.  I'm not sure what the rules are for this repo.

website/client/js/controllers/inventoryCtrl.js
```diff
@@ -54,11 +54,16 @@ habitrpg.controller("InventoryCtrl",
...
+          $scope.gear[item.klass] = [];
         $scope.gear[item.klass].push(item);
+        if (!$scope.gearByType[item.type])
+          $scope.gearByType[item.type] = [];
```
khipkin, line 68: Same here.

website/client/js/controllers/inventoryCtrl.js
Hus274: I simply moved it to a new line for clarity but left out the brackets simply in favor of leaving it alone because I don't know what the convention is for this repository. I'm fine either way.

Hus274:
There is a matching button on the battle gear side.

website/client/js/controllers/inventoryCtrl.js
khipkin: Oh I see!  In my opinion, having everything on one line makes the scope obvious enough that brackets aren't needed.  However, moving it to a new line makes the scope of the `if` a lot less clear, so there should probably be brackets.  Better safe than sorry. :+1:

khipkin:
Oh my bad.  Just not in the picture!

Hus274:
It's in the picture, above battle gear it has a grouping by button.

khipkin:
Ohhh okay now I really see it.  My bad!

Hus274:
Haha no worries, I was just confused there for a minute wondering if I missed something!

common/locales/en/gear.json
```diff
@@ -1,7 +1,10 @@
...
 
-  "weapon": "weapon",
+  "weapon": "Weapon",
```
crookedneighbor: Won't this change every other instance that uses the `weapon` string? Have you audited that every other instance that uses it is appropriate to use the capitalized version?

Hus274:
@crookedneighbor Ahh that's a great catch. I just did a search now and didn't find any other instances of this being used, but I would think that it's not possible to effectively audit every one as it could be passed through a variable (ie env.t(weapon1).
The capitalization also appears to be inconsistent across the different types. weapon was lower case but eyewear was upper case. It seemed to be split down the middle, half were uppercase and half weren't.

What do you think of for the ones that were lowercase making a new locale entry for weapon_capitalized and using that? The other option I could think of is leaving them half and half in terms of capitalization but I'm not a huge fan of that.

Hus274:
Another option would be to write some javascript in to just perform the letter capitalization on the page but I don't know if this makes sense in all locales.

crookedneighbor:
It definitely does not make sense to do the capitalization via javascript. Too many edge cases with various languages.

website/client/js/controllers/inventoryCtrl.js
```diff
@@ -54,11 +54,23 @@ habitrpg.controller("InventoryCtrl",
 
     $scope.$watch('user.items.gear', function(gear){
       $scope.gear = {};
+      $scope.gearByType = {};
       _.each(gear.owned, function(v,key){
-        if (v === false) return;
+        if (v === false)
```
crookedneighbor: Our style is to have the braces on the same line:

``` js
if (v === false) { 
  return;
}
```

common/locales/en/gear.json
```diff
@@ -1,7 +1,10 @@
 {
   "set": "Set",
+  "equipmentType" : "Equipment Type",
+  "klazz": "Class",
```
crookedneighbor: I'm going to veto `klazz` right now 😉

Hus274: lol that's totally my bad, I was trying to be consistent and apparently combined the java "clazz" convention with the "klass" that was used here.

website/client/js/controllers/inventoryCtrl.js
Hus274: sounds good, I'll update all these.

crookedneighbor:
Made a few slight modification, squashed and merged. Well done!

Hus274:
Just noticed in the diff that weapon capitalized is misspelled, it's actually waeaponCapitalized. This should definitely be fixed, sorry about that! I can submit a quick PR if you want?

crookedneighbor:
Sure.

PR #7655, opened 2016-06-12
Shortname

crookedneighbor:
Github was behaving oddly in https://github.com/HabitRPG/habitrpg/pull/7638, so I've re-opened this PR. 

Adds an `alias` property to tasks that can be used to get the task or score it. This `alias` value must be unique to the user. 

This allows tasks to have the v2-like ability to be named human readable things like `productivity` 

Left to do:
- [x] Determine if we really need to prevent updating alias after the fact
- [x] Add alias field with explanation to client
- [x] Add locales for messages
- [x] Disallow aliases in challenge task creation and updating
- [x] Confirm locales
- [x] Add [note to wiki](http://habitica.wikia.com/wiki/Challenges#Challenge_Participant.27s_Permissions) about alias property on user's challenge tasks
- [x] Add wiki page about task aliases


common/locales/en/tasks.json
```diff
@@ -20,6 +20,8 @@
     "extraNotes": "Extra Notes",
     "direction/Actions": "Direction/Actions",
     "advancedOptions": "Advanced Options",
+    "taskShortName": "Short Name",
```
crookedneighbor: @Alys @lemoness Would love your opinions on the wording for these strings.

Alys: Sounds good to me.

website/server/models/task.js
```diff
@@ -25,6 +26,13 @@ export let TaskSchema = new Schema({
   type: {type: String, enum: tasksTypes, required: true, default: tasksTypes[0]},
   text: {type: String, required: true},
   notes: {type: String, default: ''},
+  shortName: {
+    type: String,
+    validate: [function validateShortName (val) {
+      return Boolean(this.userId) && !validator.isUUID(val) && val.match(SHORT_NAME_VALIDATION_REGEX);
+    },
+    'Task short names can only contain alphanumeric characters, underscores and dashses. They must be owned by a user.'],
```
crookedneighbor: @lemoness @Alys Is this a good error message for setting an invalid value for the short name?

The intention is that 

1) it can only contain letters, numbers, dashes and underscores.
2) it will error if it is a valid uuid (so we don't get conflicts when looking up tasks by id)
3) it will error if the task is not owned by a user (IE, a challenge task)

crookedneighbor:
### First time setting short name on a task

screen shot 2016-06-12 at 8 17 03 am

### Editing a short name on a task

screen shot 2016-06-12 at 8 17 31 am

crookedneighbor:
Assuming Travis passes, I think this feature is done.

@paglias just need one final review from you and confirmation that the locale strings are good.

website/server/models/task.js
Alys: dashses -> dashes

"They must be owned by a user" will cause confusion.

A participant in an active challenge CAN give a short name to their own copies of the challenge tasks, yes? The message is saying that a challenge owner cannot set a short name on the tasks in the challenge itself? If so, it's probably best to state that explicitly.

Alys: Oh and it's probably worth stating that they can't be valid UUIDs. If you already have scripts set up for old tasks using a UUID, you might try to copy the UUID to the short name field, and then be confused about why it doesn't work because it does match the "alphanum, underscore, dash" restriction.

crookedneighbor: Oh! I think it's possible to add multiple validators for mongoose properties. Going to look into that, and that should help make the error messages more helpful. 

And yes, a challenge owner cannot set a short name on a task. And, in the future, if we set up tasks shared by groups, short names will not be set for those. IE, it requires a user to have a short name.

crookedneighbor: So if we call out challenges specifically, we'll just have to remember to add group tasks if that feature gets built out.

website/server/controllers/api-v3/tasks.js
```diff
@@ -374,19 +408,24 @@ api.scoreTask = {
...
+    let taskQuery = { userId: user._id};
+
+    if (validator.isUUID(taskId)) {
+      taskQuery._id = taskId;
+    } else {
+      taskQuery.shortName = taskId;
+    }
+
+    let task = await Tasks.Task.findOne(taskQuery).exec();
+    let direction = req.params.direction;
```
paglias, line 398: since you're destructuring req.params at line 417 it makes sense to also define direction there?

website/server/models/task.js
```diff
@@ -93,6 +101,15 @@ TaskSchema.statics.sanitizeReminder = function sanitizeReminder (reminderObj) {
   return reminderObj;
 };
 
+TaskSchema.statics.checkShortNameAvailability = async function checkShortNameAvailability (userId, shortName) {
+  let taskWithSameShortname = await this.findOne({
+    userId,
+    shortName,
+  }).exec();
```
paglias: here you should only fetch the task `_id` (only because you cannot fetch no data)

khipkin:
@crookedneighbor I awkwardly commented on the diffs from #7638 because I hadn't refreshed since you closed the issue. :scream: 

Also, I think the wall of hover text when you change the short-name is pretty overwhelming.  If _everything_ is bold like that, nothing really stands out.

khipkin:
Old text:

> **This short name can be
> used when integrating
> with 3rd party
> integrations. Only
> dashes, underscores,
> and alphanumeric
> characters are
> supported. The short
> name must be unique
> among all your tasks.
> WARNING: Changing
> this value will break any
> 3rd party integrations
> that rely on the short
> name.**

Maybe something more succinct for the help text, like:

> **This can be used when
> integrating with 3rd party
> apps.  Only dashes,
> underscores, and
> alphanumeric characters
> are supported.  The short
> name must be unique among
> all your tasks. WARNING:
> Changing this value may
> break 3rd party
> integrations.**

Not sure what the actual char limit is in the little hover box, but I stayed <= 27 char line length for both.  (The original from the pictures is <= 26 char line length.)

I think maybe the "allowed chars" info could be offered up in a validation error if necessary, but otherwise left out?  Also, the warning still blends in with the rest of the text.  Idk.

website/server/models/task.js
Alys: How about:

'Task short names can only contain alphanumeric characters, underscores, and dashes. They cannot be UUIDs. They can only be applied to tasks in a user's own task list.'

website/server/models/task.js
```diff
@@ -25,6 +25,21 @@ export let TaskSchema = new Schema({
...
+        return Boolean(this.userId);
+      },
+      msg: 'Task short names can only be applied to tasks in a user\'s own task list.',
+    }, {
+      validator: function validateShortNameIsNotUuid (val) {
+        return !validator.isUUID(val);
+      },
+      msg: 'Task short names cannot be uuids.',
+    }],
+  },
```
crookedneighbor, line 43: @Alys Was able to split these out into multiple validations, which makes a lot more sense to me.

website/server/models/task.js
```diff
@@ -167,6 +182,20 @@ TaskSchema.statics.fromJSONV2 = function fromJSONV2 (taskObj) {
 
 export let Task = mongoose.model('Task', TaskSchema);
 
+Task.schema.path('shortName').validate(function valiateShortNameNotTaken (shortName, respond) {
```
crookedneighbor: @paglias I was able to move almost all the duplicate shortName checking to the model, but because I needed access to `findOne` to do it, I had to put it after the model was defined. Is that acceptable? Or can you think of a better way to do it?

website/server/models/task.js
```diff
@@ -167,6 +182,20 @@ TaskSchema.statics.fromJSONV2 = function fromJSONV2 (taskObj) {
...
+  Task.findOne({
+    _id: { $ne: this._id },
+    userId: this.userId,
+    shortName,
+  }).exec().then((task) => {
+    let shortNameAvailable = !task;
+
+    respond(shortNameAvailable);
+  }).catch(() => {
```
crookedneighbor, line 214: Not sure about this. Is it necessary? Or should it bubble up to the global catcher?

crookedneighbor:
I agree that the popover is too verbose. :/ 

Ideally, we'd have a short popover and a link to a wiki page explaining more about it.

website/server/models/task.js
paglias: I'd say it's okay 👍

website/server/models/task.js
paglias, line 214: It's necessary in this case. We want the error to be handled by mongoose not by the global catcher

website/server/models/task.js
Alys, line 43: Looks great!

crookedneighbor:
Just had a thought. Is shortName the best title for this? We already use shortName in challenges to denote the tag name (though I'd argue that it should be tagName or tag instead)

Alys:
How about `taskAlias`? I think that's nicely meaningful, although I've thought of it so I'm biased. :)

I'd also like to see the challenge shortName changed to tagName.

crookedneighbor:
I like task alias.

website/server/controllers/api-v3/tasks.js
```diff
@@ -458,19 +467,17 @@ api.moveTask = {
...
 
     let user = res.locals.user;
+    let taskId = req.params.taskId;
     let to = Number(req.params.position);
 
-    let task = await Tasks.Task.findOne({
-      _id: req.params.taskId,
-      userId: user._id,
-    }).exec();
+    let task = await Tasks.Task.findByIdOrShortName(taskId, user._id);
```
paglias: here for example additional queries should be passed {userId: user._id}

crookedneighbor: Thanks, just missed this one. I'll double check the rest too.

crookedneighbor:
I elected to use `alias` over `taskAlias` since `task.taskAlias` seemed redundant.

crookedneighbor:
Okay, now it looks like this:

screen shot 2016-06-16 at 7 39 04 am

And it includes a link here: http://habitica.wikia.com/wiki/Using_a_Task_Alias_in_3rd_Party_Integrations

I'm satisfied with that for now. We can refine both the wiki page and the popover in the future if needed.

crookedneighbor:
@paglias @Alys Any final thoughts?

paglias:
Looks good to me!

Alys:
I'll rename the wiki page for naming style consistency so wait a little bit
before merging.

Alys:
http://habitica.wikia.com/wiki/Task_Alias will be the URL, just as soon as
my laptop stops crashing.

Alys:
The wiki page is renamed. I have no other feedback.

I really like this feature - it will be very convenient! Good calling on naming it just "alias". That makes sense.

crookedneighbor:
Thanks for the feedback. Will merge this when the build completes.

PR #7657, opened 2016-06-12
Removing accepted guild invitations from the local copy of the user o…

### LINKED ISSUE #7600: after accepting a guild/party invitation, the invitation persists
If you accept a guild invitation, the invitation is correctly deleted from your user document in the database, but it still appears in the browser until you reload. If you try to accept it again, you receive an error message (which is expected and correct, but not user-friendly).

The desired behaviour is that as soon as you've accepted the invitation, the notification for it should be removed from the notification icon in the menu bar.

Hus274:
closes #7600 
### Changes

Removed the accepted guild request from the pending invitations for a user so the notification goes away without having to refresh.

---


crookedneighbor:
Notated toward your next tier

Issue #7633, opened 2016-06-09
Disallow @habitica.com email addresses for new accounts


SabreCat:
### Description

It's possible when signing up for a Habitica account to specify a habitica.com email address. That causes all notification emails meant for the new user to arrive in the administrative inbox.

Pending a more thorough email verification workflow, we should at least disallow @habitica.com email addresses when validating the field, so people are clear that they should specify an email address they already own rather than a hypothetical new Habitica address. (If we add any new administrators to the list of legitimate habitica.com addresses, we can associate those with Habitica accounts by hand!)
#### Console Errors

N/A

librarianmage:
I can try this! Should it be server-side, client-side, or both?

crookedneighbor:
Server side for sure. We can add client side later if needed.

PR #7659, opened 2016-06-14
reject registering users using habitica.com or habitrpg.com emails

librarianmage:
# Changes:

Rejects registering users that use a habitrpg.com or a habitica.com email account (fixing #7633)

---


website/server/controllers/api-v3/auth.js
```diff
@@ -75,6 +75,10 @@ api.registerLocal = {
       email: {
         notEmpty: {errorMessage: res.t('missingEmail')},
         isEmail: {errorMessage: res.t('notAnEmail')},
+        matches: {
+          options: ['.+@(?!(?:(?:habitrpg\.com$)|(?:habitica\.com$))).+', 'i'],
```
crookedneighbor: This is a messy regex. Would prefer a custom validate method.

librarianmage: Not sure if that can be done with the schema, but I can look.

crookedneighbor: http://mongoosejs.com/docs/validation.html

librarianmage:
@crookedneighbor Can you help me find why all of the integration tests broke?

librarianmage:
@crookedneighbor if it helps, I've reasonably determined that it is occurring in the generateUser function

website/server/controllers/api-v3/auth.js
```diff
@@ -75,6 +75,10 @@ api.registerLocal = {
       email: {
         notEmpty: {errorMessage: res.t('missingEmail')},
         isEmail: {errorMessage: res.t('notAnEmail')},
+        notEmailDomains: {
```
crookedneighbor: This is why all the tests are failing.

crookedneighbor: Why did you move the validation here instead of using a validate function on the model like I asked for?

librarianmage: @crookedneighbor I guess I misunderstood. I'm sorry.

librarianmage: Do you want it in the `middlewares` section?

crookedneighbor: Take a look at how it's done here: https://github.com/HabitRPG/habitrpg/pull/7655/files#diff-1f63728ba8da6291c25b8acf11aab121R32

librarianmage: Thanks!

website/server/models/user/schema.js
```diff
@@ -24,7 +24,21 @@ let schema = new Schema({
     local: {
       email: {
         type: String,
-        validate: [validator.isEmail, shared.i18n.t('invalidEmail')],
+        validate: [{
+          validator: validator.isEmail,
+          msg: shared.i18n.t('invalidEmail'),
+        }, {
+          validator: function notEmailDomains (email) {
+            let domains = ['habitica\\.com', 'habitrpg\\.com'];
```
crookedneighbor: I'd put these as a const near the top of the file:

``` js
const INVALID_DOMAINS = ['habitica.com', 'habitrpg.com'];
```

And then in the validator, I'd just do this:

``` js
validator (email) {
  return INVALID_DOMAINS.every(domain => email.indexOf(domain) === -1);
}
```

librarianmage: Okay, I just didn't want to exclude an email like `habitica.com@example.com` (even though that is a really contrived example)

crookedneighbor: Good point, I had intended to make it:

``` js
return INVALID_DOMAINS.every(domain => email.indexOf(`@${domain}`) === -1);
```

And just forgot. That should take care of that edge case.

librarianmage: that might work. Would you still want an email like `example@test.habitica.com` to fail?

librarianmage: If you do, would

``` javascript
return INVALID_DOMAINS.every(domain => new RegExp(`${domain}$`, 'i').test(email) === false);
```

work?

librarianmage: that would also fix failing an email like `not.fail@habitica.com.au`

website/server/models/user/schema.js
```diff
@@ -9,6 +9,8 @@ import {
 
 const Schema = mongoose.Schema;
 
+const INVALID_DOMAINS = ['habitica.com', 'habitrpg.com'];
```
awitherow: Do you think there will be more invalid domains eventually? If the list here grows, perhaps exporting stuff to a config file of sorts would be useful. Perhaps also over-engineering 😉

librarianmage: Perhaps, if Habitica goes through another rebranding 😉

crookedneighbor: Full disclosure: I was going to convert this to an env variable while merging.

librarianmage: Makes sense

website/server/models/user/schema.js
```diff
@@ -24,7 +26,15 @@ let schema = new Schema({
     local: {
       email: {
         type: String,
-        validate: [validator.isEmail, shared.i18n.t('invalidEmail')],
+        validate: [{
+          validator: validator.isEmail,
+          msg: shared.i18n.t('invalidEmail'),
+        }, {
+          validator: function notEmailDomains (email) {
```
crookedneighbor: Let's use the object shorthand here:

``` js
validator (email) {
```

librarianmage:
Finished!

librarianmage:
:bell: This should be finished.

crookedneighbor:
I elected to remove the client side portions of this after fixing a bug that prevented the custom messages from being displayed on the client. https://github.com/HabitRPG/habitrpg/commit/0150b355cb25067b23969b13b1048756548f8927

Issue #4840, opened 2015-03-11
Canceling subscription inconsistent with keeping block subscriptions


lefnire:
From a user: 

> I was/am currently in possession of subscriber perks (nearing the end of a 3 month block).
> I decided to buy more time; in a 3 month block.
> 
> I picked the 3 month option (cost: $15), the payment was made, and I then unsubscribed (because I don't like automated recurring payments unless absolutely necessary).
> 
> Unfortunately, instead of calculating me as having paid for 3 months, it says my subscription is due to expire in 1 month (+ the credit I have remaining from the previous block).
> 
> This is the second time this has happened. 
> I believe a similar issue cropped up when I bought one of the new 3 month subscriptions just as my old subscription (ie: before the new subscriptions setup) expired.
> 
> Is this perhaps some sort of old code that assumes any subscription being canceled only lasts 1 month?

I've seen this from time to time, but haven't pegged the issue (plus it's inconsistent) 
##

lefnire:
One thing that seems to be at play, if `user.purchased.plan.dateTerminated` is in the past (they had a plan before, re-upping now) then it might be adding to _that_ date and not the current date?

zelq:
This just happened to me. I cancelled the recurring, but only got 1 month of my 3.

Alys:
@zelq We need to know your Habitica account to be able to help you. Please go to Help -> Report a Bug and see step 6 (or just tell us your UUID if you already know how to find it; it is different than your user name).

zelq:
Grrr, meant to include that. Sorry.
It's be67f33e-abb0-4242-8974-50b4704df771

Alys:
@zelq I can't find a record of your payment. Can you please email admin@habitica.com and tell me which payment option you chose, and either which email address you entered into the payment form (for a credit card payment) or which email address you use to log into the payment processor website.

Jolanda5:
Is this still an issue? Is somebody working on this? Can be worked on this at the moment?

Alys:
@Jolanda5 This issue is best not worked on at the moment. Habitica's API and related code are undergoing redevelopment at the moment, and that includes the code related to subscriptions. Any changes made for this issue would conflict with the new API code.

I've actually just made a post about that in the [Aspiring Coders](https://habitica.com/#/options/groups/guilds/68d4a01e-db97-4786-8ee3-05d612c5af6f) guild. Drop in there and if you have any questions, you're welcome to ask them there.

[...]

PR #7660, opened 2016-06-14
Fixing Subscription Cancellation Bug

Hus274:
#4840
### Changes

Added functional default behavior to subscription cancellation to calculate the terminated date and migrated amazon payment method to use this default. This now makes the default behavior of cancelling a subscription give you 30 days for every month you subscribed for from the time you last paid for a subscription.

---


website/server/libs/api-v3/payments.js
```diff
@@ -117,7 +118,8 @@ api.createSubscription = async function createSubscription (data) {
 api.cancelSubscription = async function cancelSubscription (data) {
   let plan = data.user.purchased.plan;
   let now = moment();
-  let remaining = data.nextBill ? moment(data.nextBill).diff(new Date(), 'days') : 30;
+  let defaultRemaining = plan.subscriptionLengthMonths ? plan.subscriptionLengthMonths * 30 : 30;
+  let remaining = data.nextBill ? moment(data.nextBill).diff(new Date(), 'days') : defaultRemaining;
```
paglias: why is nextBill still used here?

test/api/v3/unit/libs/payments.test.js
```diff
@@ -55,6 +61,21 @@ describe('payments/index', () => {
       expect(difference - 60).to.be.lessThan(3); // the difference is approximately two months, +/- 2 days
     });
 
+    it('data.nextBill is undefined', () => {
```
crookedneighbor: This it block doesn't tell me the expected behavior

test/api/v3/unit/libs/payments.test.js
```diff
@@ -55,6 +61,21 @@ describe('payments/index', () => {
...
+      data.user.purchased.plan.subscriptionLengthMonths = 3;
+      api.cancelSubscription(data);
+      let terminated = data.user.purchased.plan.dateTerminated;
+      let difference = moment(terminated).diff(data.user.purchased.lastBillingDate, 'days');
+      expect(difference).to.be.eql(89);
+    });
+
+    it('data.nextBill is undefined and plan.subscriptionLengthMonths is undefined', () => {
```
crookedneighbor: Same with this one

test/api/v3/unit/libs/payments.test.js
crookedneighbor: Uhg, looking at the rest of the file, this test file did not model good it blocks.

Hus274: If you can point me at a good example file or let me know what the standard is, I can update this file to match it.

website/server/libs/api-v3/payments.js
Hus274: The reason I left nextBill in is because it still makes sense in some contexts. Payment methods that have subscription functionality look like they maintain this information for you so it makes sense that the subscription should end when they say you were supposed to pay next. Amazon doesn't have this feature and therefore shouldn't be sending a made up next date. 

The idea was to allow the payment method to override the default behavior if it had the information needed but to provide smart default behavior that should be capable of computing an accurate dateTerminated.

The main reason I would prefer not to remove this for all payment methods is that the default behavior depends on a new subscription being created with the new field. If we removed the nextBill from all payment methods, it would break anyone who tried to cancel an already existing subscription because the subscription wouldn't have a subscriptionLengthMonths field and would therefore default to 30 days even if paypal said that the next bill was a year from now.

test/api/v3/unit/libs/payments.test.js
khipkin: https://github.com/habitrpg/habitrpg/blob/develop/test/api/v3/unit/libs/cron.test.js perhaps?

crookedneighbor: For instance, the test you created above, `it('sets subscription length'`, is great. It reads like a sentence and tells me exactly what you expect to happen. `it('data.nextBill is undefined'` does not do either of those things.

Hus274: Ok, I'll update the tests in this file and resubmit!

Hus274:
Fixed up the tests according to the comments

crookedneighbor:
I think this looks good to me. 👍

crookedneighbor:
Actually, I'm not so sure anymore. Looking at it again, wouldn't this cause a 3 month amazon subscription that was cancelled 2 months in to have an additional 3 months on the termination date? IE, this solution only works for someone who purchases the subscription and immediately cancels it. If you cancel it any time in the future, you get extra time subscribed.

crookedneighbor:
Yep, just confirmed that is the case.

Hus274:
Yeah that was a good catch. I've reverted back to the original model of amazon setting a nextBill field. This will allow amazon to calculate the nextBill from the lastBilledDate ( an amazon specific variable ) without the general method being responsible for figuring a payment specific nextBilled date.

Thanks for catching that!

crookedneighbor:
Looks like you pushed an unrelated commit about challenges.

Hus274:
That was strange. not quite sure what happened there but my git branches got a bit messed up. Should be cleaned up now.

website/server/models/user/schema.js
```diff
@@ -137,6 +137,7 @@ let schema = new Schema({
       gemsBought: {type: Number, default: 0},
       mysteryItems: {type: Array, default: () => []},
       lastBillingDate: Date, // Used only for Amazon Payments to keep track of billing date
+      subscriptionLengthMonths: Number, // Length of Subscription
```
crookedneighbor, line 140: Is this even really necessary? Don't we get the number of months from the subscription block, which we can find from the plan id?

crookedneighbor:

Issue #2466, opened 2014-01-18
Opening Tavern Tab Doesn't Refresh


deilann:
When you go into Guilds and click on their tab, it automatically refreshes the chat, so that you can see new messages. Is there a logistical reason clicking on the Tavern tab doesn't do the same?
##

wc8:
I can't answer the question, but I can tell you this: if you add Tavern to your browser bookmark bar, I think it does refresh when you access the page that way.

deilann:
Yes. If you open a new tab or do a force refresh or do anything but
navigate to the tavern tab through a funnel, it'll refresh. I'm just not
sure what triggers a Guild refresh upon opening the tab and why it's not
implemented for the Tavern, if there's a reason.

PaulSainsbury:
I think I can fix this one.  I'll send a pull request shortly.

paglias:
If I recall correctly guilds are not cached because they use a lot of memory so each time you open one it's fetched again from the server, while the tavern is cached.

PaulSainsbury:
Yeah, that seems to be it.  I've basically made the controller for '/groups/tavern' update the cached copy before assigning it to the scope.  I'm kinda new to Angular, Node and Mongo so I'm not yet sure if I've broken some best practice or not though.

paglias:
I think it's ok, you should just check one thing: when opening the tavern for the second, third... time can you check that the memory usage doesn't increase too much? Because last time we implemented an auto update for the tavern chat it had this problem, you can use the chrome task manager for that.

PaulSainsbury:
Cool, I'll do that.  How well is the code covered by tests and how do I run them before I submit my pull request? (I'm running on Ubuntu)

paglias:
Actually when you "sync" the chat we have a 100mb surge in memory, for some reason the old chat object is not garbage collected correctly.

About tests, if you installed habitrpg (npm install) and its global dependencies (npm install -g grunt mocha bower) then you can run tests using `npm test`. Anyway I don't think this is covered by tests (they are [here](https://github.com/HabitRPG/habitrpg/tree/develop/test)) and when you send a pull request it's automatically tested remotely using TravisCI

PaulSainsbury:
Yeah, I have very little data in my chat and the sync was adding 2-3mb per sync. I'm trying to spot what could be referencing the "old" object still that might cause the browser to not garbage collect it... but likely that's going to mean I need a far better understanding of the Angular $scope usage. :(

[...]

PR #7664, opened 2016-06-15
Real-time Chat

paglias:
WIP
### Implementation
- [ ] Tweaks to interface and notifications on the client
- [ ] What to do with flags
- [ ] Spam blocker
- [ ] Limit concurrent messages
- [ ] Security when booting an user from a group (we must be sure it's been disconnected)
- [ ] Batch updates to `user.newMessages.{groupId}`?
- [ ] is `user.lastMessageSeen` still necessary?
### Technical steps
- [x] Add pusher library to the server 
- [x] Limit to private parties / guilds for now
- [x] Add channel authentication
- [ ] Authenticate multiple channels with 1 HTTP call (only needed for guilds)
- [x] Make sure the user has access to the channels when authenticating
- [x] Add real-time updates to the client
- [ ] Implement party cap
- [ ] Mark a message as seen when the user is on the party page and a new message comes in
- [ ] Implement tweaks to client UI
- [x] use only presence channel for parties
- [x] remove user when is booted or leaves group
- [x] Remove old Firebase code if we definitely decide to not use it anymore
- [x] Tests


website/server/controllers/api-v3/auth.js
```diff
@@ -282,6 +283,55 @@ api.loginSocial = {
...
 
+/*
+ * @apiIgnore
+ * @api {post} /api/v3/user/auth/pusher Pusher.com authentication
+ * @apiDescription Authentication for Pusher.com private and presence channels, private route
+ * @apiVersion 3.0.0
+ * @apiName UserAuthPusher
+ * @apiGroup User
+ *
+ * @apiParam {String} socker_id Body parameter
```
khipkin: socket_id

paglias: fixed

website/server/controllers/api-v3/auth.js
```diff
@@ -312,10 +314,30 @@ api.pusherAuth = {
...
+      throw new BadRequest('Invalid Pusher channel type.');
+    }
+
+    if (resourceType !== 'group') { // only groups are supported
+      throw new BadRequest('Invalid Pusher resource type.');
+    }
+
+    resourceId = resourceId.join('-'); // the split at the beginning had split resourceId too
+    if (!validator.isUUID(resourceId)) {
+      throw new BadRequest('Invalid Pusher resource id, must be a UUID.');
```
paglias, line 327: @crookedneighbor do you think these strings should be localized?

crookedneighbor, line 327: These all look like implementation details, so as long as the API is private, I think it's fine to not localize them. A user should never see these errors.

test/api/v3/integration/user/auth/POST-user_auth_pusher.test.js
```diff
@@ -0,0 +1,102 @@
...
+  it('returns an error if the passed resource id doesn\'t match the user\'s party', async () => {
+    await expect(user.post(endpoint, {
+      channel_name: `private-group-${generateUUID()}`,
+      socket_id: '123',
+    })).to.eventually.be.rejected.and.eql({
+      code: 404,
+      error: 'NotFound',
+      message: 'Resource id must be the user\'s party.',
+    });
+  });
```
khipkin, line 101: success case(s)?

paglias, line 101: The problem is that these are integration tests and we cannot stub the return value of `pusher.authenticate`

bower.json
```diff
@@ -42,6 +42,8 @@
     "swagger-ui": "wordnik/swagger-ui#v2.0.24",
     "smart-app-banner": "78ef9c0679723b25be1a0ae04f7b4aef7cbced4f",
     "habitica-markdown": "1.2.1"
```
khipkin: Missing comma at EOL

paglias: good catch! it was why the build was failing

khipkin:
Can probably close #2466 when you finish this and merge it in. :smiley_cat:

paglias:
Going to merge it as it's, will implement the UI separately

Issue #7653, opened 2016-06-12
a player can incorrectly remain in a quest in a party that they have left


Alys:
greeniezona (bf3cdf52-0b87-448a-81dd-459d74323230) reported in the Report a Bug guild:

"I keep randomly losing HP for no reason. I completed all dailies, there was no damage from my party. My suspicion is this damage is from a quest in my old party, a quest and party that I left over a week ago. It shouldn't still be doing damage to me, right?"

That shouldn't be possible but I checked anyway and found this:

greeniezona's current party:

```
{
    "_id": "84bf6664-6bfb-4844-b77a-a5da6dff01c0",
    "quest": {
        "leader": "bf3cdf52-0b87-448a-81dd-459d74323230",
        "members": {
            "bf3cdf52-0b87-448a-81dd-459d74323230": true,
            [and others]
        },
```

greeniezona's previous party:

```
{
    "_id": "c52de31f-2b8d-4da2-8690-148a8d51b944",
    "quest": {
        "leader": "[someone else]",
        "members": {
            [other people]
            "bf3cdf52-0b87-448a-81dd-459d74323230": false,
            [other people]
        },
```

That quest in the previous party was active, even though it had that one false entry in it (must have been force-started by the quest owner).

I have NOT yet had confirmation that the damage from that previous-party quest matches the damage that greeniezona received, but that `false` member value shouldn't be there, regardless.

FYI for Socialites: If you see a player reporting unexplained damage, it could be from:
- negative Habits that they clicked on by accident (mouse might have twitched)
- their own Dailies that weren't marked completed, possibly due to sync errors (Data > Data DIsplay Tool > Dailies History will show if that's happened)
- boss damage from their current party (they can go to the party chat and click "fetch recent messages" to see damage reports)
- this bug (there is no way for a player to work out if they are still in a previous party's quest; an admin will have to search the database to find out)

Actions:
- [x] remove greeniezona from their previous party's quest
- [x] find out how widespread this is - EDIT: 239 players affected
- [x] fix all affected players (Alys will do this today)
- [x] PR #7848 - check that the code for force-starting a quest removes any players from quest.members that don't have a `true` value
- [x] PR #7848 - check that the code for force-starting a quest removes any players from quest.members that aren't in the party
[...]

PR #7665, opened 2016-06-15
prevents damage to quest participants with false or null values - #7653

Alys:
Partial fix for #7653 - addresses this part: "adjust the code that allocates quest damage so that it doesn't affect people who don't have a true value in quest.members."
### Changes

Makes _processBossQuest() not apply damage to quest participants who have a `false` or `null` value, which can happen if a bug prevents them from leaving a quest correctly.

Also adds a test that does fail against the current `develop` code but passes with this PR.


Alys:
If there's no complaints, I'll merge this in a couple of hours and do a deployment to production.

crookedneighbor:
👍

Issue #7629, opened 2016-06-09
Possible inconsistency with naming quest lines: Moonstone chain, Golden Knight chain


ghost:
An issue raised on Transifex and in the i18n guild 

> **_Niika**_  I have a problem to understand a title: "The Moonstone Chain, Part 1: The Moonstone Chain". The Chain in first part means a set of quests and in the second part it means chain made of moonstones or not?

 -> about `questMoonstone1Text`

> **_lyubomirv**_ What's with the 'chain'? There is nothing about any chains in the text…

 -> about `questGoldenknight1DropGoldenknight2Quest` _The Golden Knight Chain Part 2: Gold Knight (Scroll)_ and `questGoldenknight2DropGoldenknight3Quest` _The Golden Knight Chain Part 3: The Iron Knight (Scroll)_

While I think it's OK to have a quest name repeated in the name of the quest line (The Moonstone Chain, Part 1: The Moonstone Chain) - even if it may look a bit unnatural, I'm convinced that calling a quest line a "chain" in case of the Knight is more misleading.

@lemoness could you please share you opinion:
- may the two lines about the Golden Knight be changed to exclude the "chain"?
- (mainly for possible future reference) is it OK to have a quest name repeating the name of the quest line?

lemoness:
So, any references to chains in the Golden Knight quest line titles are a mistake. The Moonstone Chain literally refers to the line "Without a chain of moonstones, nothing can harm me" and is not meant to be a different name for a quest line. 

As for the repetition, let's change the name of that quest line to Recidivate, and the second quest to Recidivate Pt. 2: The Necromancer. That should remove the worst repetition. There's still a little in Recidivate Pt. 3: Recidivate Transformed, but its analogous to Vice Pt 3: Vice Awakens, and I think that's acceptable.

ghost:
@lemoness thank you! Since it means major renaming, I believe another minor inconsistency could be fixed as well:

> **_walderich**_ There seems to be another inconsistency regarding the quest scrolls. They are usually suffixed with "(Scroll)", but The SnackLess Monster and The Laundromancer uses "(Quest Scroll)".

crookedneighbor:
Closed by https://github.com/HabitRPG/habitrpg/pull/7672

PR #7672, opened 2016-06-16
Fixing quest naming according to @lemoness #7629

Hus274:
#7629
### Changes

Fixing naming of the golden knight by removing the word chain and changing the moonstone chain quest name to recidivate.

---


crookedneighbor:

Issue #7392, opened 2016-05-21
[API v3] Recent completed To-Dos are not displayed on website


Alys:
### Description

No completed To-Dos are displayed in the "Done" column on the api v3 test website, even when ones have been completed very recently.
If you complete a To-Do and then immediately view the "Done" column, you see the To-Do, but it disappears as soon as you sync.
The To-Dos DO exist in the database and ARE returned by API calls. It's just a website display problem.

---

_edit by Alys, 1 June:_

**The bug behaviour and the workaround are:**

When you go to the "Done" tab for To-Dos, you see all your completed To-Dos from the past three days (as intended, and as happened before the Maintenance). However while you're looking at the completed tab, if you sync or if something causes a sync to occur, then all the completed To-Dos will disappear. Syncing again won't bring them back. What you have to do to see them again is reload the site (using your browser's reload/refresh button) and then go back to the Done tab.

The fix should be that when the "Done" tab is active, all completed To-Dos should always be visible.

Alys:
#### Console Errors

none


#### Extra Description

The To-Dos DO exist in the database and ARE returned by API calls. It's just a website display problem.

paglias:
@Alys can't replicate this, if works for me. Note that completed todos are loaded only when you open the Completed Tab so if they don't appear immediately that's the reason. But when I open the Completed tab for the first time a request is immediately dispatched to the site and they appear really quickly

SabreCat:
I saw this in action on production just now. I brought up the Completed To-Dos list; it showed more items than I'm used to, but worked. Then I took another action (a Brutal Smash, in this case), and the entire list vanished.

Hard refresh brings back the list.

paglias:
This happens when the user is synced because user.todos is replaced completely

Hus274:
The api call that is being made when sync is called only calls for the uncompleted todos and then overwrites the entire todos object as paglias said. I've updated the sync method to also call for completed todos. Just want to write some tests before committing.

Hus274:
Actually I'm not seeing any tests around the client side JS files so I'm not sure quite how I would further test this in addition to the manual validation. I'll submit the pull request for now, I can go back and add tests if needed.

crookedneighbor:
We're putting off tests on the client side js until the client rewrite. Just do a lot of manual testing for now.

PR #7673, opened 2016-06-16
Making sync method call for the completed todos in addition to other tasks

Hus274:
#7392
### Changes

Added a call to the api to get the completed todos while syncing.

---

Issue #7392


website/client/js/services/userServices.js
```diff
@@ -100,11 +100,13 @@ angular.module('habitrpg')
               }
             });
           }
-
-          return Tasks.getUserTasks();
+          let completedTodos = Tasks.getUserTasks(true);
+          let nonCompletedTasks = Tasks.getUserTasks();
+          return Promise.all([completedTodos, nonCompletedTasks]);
```
crookedneighbor: Native promises are not supported in every browser we support: http://caniuse.com/#feat=promises

Hus274: Sad day, it was a lot nicer with the promises :p

website/client/js/services/userServices.js
```diff
@@ -103,9 +104,19 @@ angular.module('habitrpg')
 
           return Tasks.getUserTasks();
         })
+        // refresh all tasks but completed todos
         .then(function (response) {
           var tasks = response.data.data;
           syncUserTasks(tasks);
+          return Tasks.getUserTasks(true)
```
paglias: this way there are 2 calls to Tasks.getUserTasks, one at line 105 and one at 111

paglias: Just ignore what I said

paglias:
The problem with this is that completed todos will be synced automatically at every `sync()` but we only want them to load if the completed todos tab is accessed

crookedneighbor:
Why not store the completedTodos on a separate object that gets ignored when calling sync?

Hus274:
@crookedneighbor do you mean a different user property? Like user.completedTodos?

@paglias The concern I have there is that the sync method would then need to know if the Done tab is currently selected (if it is, the refresh needs to happen when syncing). This seemed like it was tying the sync method too closely to the done tab which is why I avoided that.

Hus274:
So should the expected behavior of the sync method depend on if the user is currently viewing completed todos? And then only load the completed todos when syncing if the user is already on that page? If they're not on the page just clear the completed todos from local memory and re pull them in when they click on the tab.

paglias:
Yeah, it's the best way
Il 17 giu 2016 3:36 PM, "Travis" notifications@github.com ha scritto:

Hus274:
@paglias Ok, I'll make that change and resubmit.

Hus274:
Ok resubmitted and merged back into develop branch

website/client/js/services/userServices.js
```diff
@@ -103,8 +104,28 @@ angular.module('habitrpg')
 
           return Tasks.getUserTasks();
         })
+        // refresh all but completed todos
         .then(function (response) {
-          var tasks = response.data.data;
+          tasks = response.data.data;
+
+          // only refresh completed todos if the user has the completed tabs list open
+          var taskButton = document.getElementById('completedTasksButton');
```
paglias: If possible I'd like to avoid relying on the DOM. Can you see if you can get access to the `list.view` value used here https://github.com/HabitRPG/habitrpg/pull/7673/files#diff-52e526ae61c107ffe0964693be9534faL34? Otherwise it's okay this way

website/client/js/services/userServices.js
```diff
@@ -103,8 +104,28 @@ angular.module('habitrpg')
...
+          if (taskButton && taskButton.classList.contains('active')) {
+            return Tasks.getUserTasks(true)
+          }
+        })
+        // refresh completed todos
+        .then(function (response) {
+          if (response) {
+            tasks = tasks.concat(response.data.data);
```
paglias, line 119: `tasks.push.apply(tasks, response.data.data)` so you avoid creating a new array

website/client/js/services/userServices.js
Hus274: I was able to do this but using the list.view == 'complete' in the expression results in it evaluating to true even when on a different page (like Social or Inventory). This results in the call for loading completed tasks being made anytime the user left the task page with the "Done" todos list item selected.

Hus274: was able to get around this by also checking to make sure that the current page title was tasks. I'll be pushing the new revision here shortly.

website/client/js/services/userServices.js
```diff
@@ -106,11 +106,11 @@ angular.module('habitrpg')
...
         // refresh all but completed todos
         .then(function (response) {
-          tasks = response.data.data;
+          tasks.push.apply(tasks, response.data.data);
+
 
           // only refresh completed todos if the user has the completed tabs list open
-          var taskButton = document.getElementById('completedTasksButton');
-          if (taskButton && taskButton.classList.contains('active')) {
+          if ($rootScope.lists && $rootScope.pageTitle == 'Tasks' && _.find($rootScope.lists, {'type':'todo'}).view == 'complete') {
```
crookedneighbor: Will asserting on the page title work if you're using a non-English locale? Probably better to use the $stare module.

website/client/js/services/userServices.js
```diff
@@ -103,8 +104,28 @@ angular.module('habitrpg')
...
           return Tasks.getUserTasks();
         })
+        // refresh all but completed todos
         .then(function (response) {
-          var tasks = response.data.data;
+          tasks.push.apply(tasks, response.data.data);
+
+
+          // only refresh completed todos if the user has the completed tabs list open
+          if ($rootScope.lists && $rootScope.pageTitle == 'Tasks' && _.find($rootScope.lists, {'type':'todo'}).view == 'complete') {
```
crookedneighbor: I meant $state

Hus274:
Makes sense, adjusted accordingly.

website/client/js/services/userServices.js
```diff
@@ -103,8 +104,27 @@ angular.module('habitrpg')
 
           return Tasks.getUserTasks();
         })
+        // refresh all but completed todos
         .then(function (response) {
-          var tasks = response.data.data;
+          tasks.push.apply(tasks, response.data.data);
```
paglias, line 109: in this case you have an empty array before adding the fetched data, just replace it

paglias:
Apart from the comment I left it looks good to me 👍

crookedneighbor:
Made the final adjustment @paglias asked for and squashed it. 

PR #7674, opened 2016-06-16
Fix auth if localStorage is missing or corrupted

paglias:


website/client/js/app.js
```diff
@@ -330,13 +330,19 @@ window.habitrpg = angular.module('habitrpg',
...
 
-      if (settings && settings.auth) {
-        $httpProvider.defaults.headers.common['Content-Type'] = 'application/json;charset=utf-8';
+      try {
+        settings = JSON.parse(localStorage.getItem(STORAGE_SETTINGS_ID));
+      } catch (e) {
+        settings = {};
+      }
+
+      if (settings && settings.auth && settings.auth.apiId && settings.auth.apiToken) {
```
khipkin, line 341: `settings` will always be truthy here

paglias, line 341: Not if localStorage has null stored

website/client/js/services/userServices.js
```diff
@@ -613,7 +621,7 @@ angular.module('habitrpg')
       }
 
       //If user does not have ApiID that forward him to settings.
-      if (!settings.auth.apiId || !settings.auth.apiToken) {
+      if (!settings || !settings.auth || !settings.auth.apiId || !settings.auth.apiToken) {
```
khipkin: `settings` will always be truthy here

website/client/js/app.js
khipkin, line 341: ahh true

paglias, line 341: It's also true that null.auth won't break anything so I could remove it...

khipkin, line 341: Nahh I think it's better like this since `settings` can be falsey

website/client/js/services/userServices.js
```diff
@@ -625,7 +633,7 @@ angular.module('habitrpg')
           var isStaticOrSocial = $window.location.pathname.match(/^\/(static|social)/);
           if (!isStaticOrSocial){
             localStorage.clear();
-            $location.path('/logout');
+            window.location.href = '/logout';
```
crookedneighbor: This is why the build is failing. Karma doesn't like full page redirects. you could use $window instead.

website/client/js/services/userServices.js
```diff
@@ -633,7 +633,7 @@ angular.module('habitrpg')
           var isStaticOrSocial = $window.location.pathname.match(/^\/(static|social)/);
           if (!isStaticOrSocial){
             localStorage.clear();
-            window.location.href = '/logout';
+            $window.location.href = '/logout';
```
crookedneighbor, line 636: You may still need to stub `$window` in the test, but Karma provides a fake window for that kind of thing.

paglias, line 636: yeah I'm trying to find where it happens.. they could provide some more info with that error

crookedneighbor, line 636: Problem is, I suspect it's happening in multiple places.

paglias, line 636: eh okay but this way I have to check every test file to find if it's happening or not... it takes ages

paglias:
Thanks @crookedneighbor !

PR #7676, opened 2016-06-16
Correct GA unsub info

SabreCat:
Fixes an issue spotted in server error logs (trimmed):

> Error: Please provide at least an event category (ec) and an event action (ea)
>    at Object.Visitor._handleError (/var/app/current/node_modules/universal-analytics/lib/index.js:391:24)
>  in libs/api-v3/payments.js
### Changes

We were passing `gaCategory` in the `data` of an analytics service call, when the expected property name was simply `category`. This led to us trying to send Google Analytics information without required fields. This commit corrects the error.

---

UUID: [Money Label Error](https://map.what3words.com/money.label.error)

### DIFF
--- a/website/server/libs/api-v3/analyticsService.js
+++ b/website/server/libs/api-v3/analyticsService.js
@@ -144,7 +144,7 @@ let _generateValueForGoogleAnalytics = (data) => {
 
 let _sendDataToGoogle = (eventType, data) => {
   let eventData = {
-    ec: data.category || data.gaCategory || 'behavior',
+    ec: data.gaCategory || data.category || 'behavior',
     ea: eventType,
   };
 
--- a/website/server/libs/api-v3/analyticsService.js
+++ b/website/server/libs/api-v3/analyticsService.js
@@ -116,20 +116,6 @@ let _sendDataToAmplitude = (eventType, data) => {
   });
 };
 
-let _generateCategoryForGoogleAnalytics = (data) => {
-  let category;
-
-  if (data.category) {
-    category = data.category;
-  } else if (data.gaCategory) {
-    category = data.gaCategory;
-  } else {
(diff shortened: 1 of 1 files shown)


paglias:
@SabreCat I serched in the server code and there are many other places where `gaLabel` is used. Also in the analytics service code `server/libs/api-v3/analyticsService.js` and `server/libs/api-v2/analytics.js`

crookedneighbor:
The analytics service should convert gaLabel to label for the universal analytics module. Is that not the case?

SabreCat:
Ah, looks like the fix needs to be a little different. @crookedneighbor: I think the trouble is that we convert `gaLabel` but not `gaCategory`? https://github.com/HabitRPG/habitrpg/blob/develop/website/server/libs/api-v3/analyticsService.js#L147

SabreCat:
OK, might 8631a3b be a better approach, @paglias @crookedneighbor ?

SabreCat:
@khipkin: Yay, the fix is a one-liner now :stuck_out_tongue:

website/server/libs/api-v3/analyticsService.js
```diff
@@ -144,7 +144,7 @@ let _generateValueForGoogleAnalytics = (data) => {
 
 let _sendDataToGoogle = (eventType, data) => {
   let eventData = {
-    ec: data.category,
+    ec: data.category || data.gaCategory || 'behavior',
```
crookedneighbor: Shouldn't this give preference to gaCategory if both are passed in?

SabreCat: Sure, good point. I did it this way for best consistency with the behavior where it only looked at `data.category`, but I'll turn it around!

PR #7682, opened 2016-06-17
Push notifications

paglias:
Original PR https://github.com/HabitRPG/habitrpg/pull/7642
- [x] Unit testing
- [ ] Remove old devices


paglias:
Missing tests

paglias:
Rebased

test/api/v3/integration/user/POST-user_push_device.test.js
```diff
@@ -0,0 +1,62 @@
...
+    await user.post('/user/push-devices', {type, regId});
+    await expect(user.post('/user/push-devices', {type, regId}))
+      .to.eventually.be.rejected.and.eql({
+        code: 401,
+        error: 'NotAuthorized',
+        message: t('pushDeviceAlreadyAdded'),
+      });
+  });
+
+  // More tests in common code unit tests
```
crookedneighbor: This line should be removed.

test/api/v3/unit/libs/pushNotifications.js
```diff
@@ -0,0 +1,184 @@
...
+    });
+
+    sendPushNotification = requireAgain(pathToPushNotifications);
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it('returns if user is not supplied', () => {
```
crookedneighbor: Shouldn't these be `it('throws an error if....`

website/server/controllers/api-v3/user.js
```diff
@@ -1267,32 +1267,6 @@ api.userReroll = {
 };
 
 /**
-* @api {post} /api/v3/user/addPushDevice Add a push device to a user
```
crookedneighbor, line 1270: Hmmm.... we're sure this could not actually be used, right? Otherwise, this is a breaking change.

paglias, line 1270: I wouldn't worry about this, push notifications are disabled

test/api/v3/unit/libs/pushNotifications.js
paglias: Yep, forgot to updated them!

website/server/controllers/api-v3/user.js
paglias, line 1270: I mean even if a 3rd party used this route (and I don't see why they would) it would just have not done anything. Now they'll get a 404

crookedneighbor, line 1270: :+1:

PR #7685, opened 2016-06-18
Local deletion of completed tasks, updating to api v3 call.

### LINKED ISSUE #7549: Can't delete completed tasks
UUID: 84947daf-c445-4bf6-b49c-e62adcd2043d
Browser: Safari 
OS: OS X El Capitan

On my TODOs, I try to "Delete Completed" tasks and sync but after I refresh my browser, the completed tasks come back. I've tried to delete them more than three times but they always come back, even if I sync right after deleting. 

---

_edit by Alys, 2016-06-16:_

I'm seeing different behaviour now. Clicking the "Delete Completed" button makes the blue sync line appear but the completed To-Dos are still visible, however if you reload, then the completed To-Dos disappear, so it's probably related to https://github.com/HabitRPG/habitrpg/issues/7392 . There's also no confirmation step; I thought there used to be but I could well be wrong.

The desired behaviour is:
1. Click on the "Delete Completed" button.
2. See a confirmation step of some kind.
3. After choosing to proceed:
   - all completed To-Dos are deleted from the database 
   - they disappear immediately from the current view (if essential, this could be done by forcing a reload; that might be the best approach until https://github.com/HabitRPG/habitrpg/issues/7392 is fixed)

Hus274:
#7549
### Changes

Updating call from User api v2 to Tasks v3, adding deletion verification step, and clearing the local list upon deletion. closes #7549

---

### DIFF
--- a/common/locales/en/tasks.json
+++ b/common/locales/en/tasks.json
@@ -94,6 +94,7 @@
     "confirmFortify": "Are you sure?",
     "fortifyComplete": "Fortify complete!",
     "sureDelete": "Are you sure you want to delete the <%= taskType %> with the text \"<%= taskText %>\"?",
+    "sureDeleteCompleted": "Are you sure you want to delete your completed todos?",
     "streakCoins": "Streak Bonus!",
     "pushTaskToTop": "Push task to top.  Hold ctrl or cmd to push to bottom.",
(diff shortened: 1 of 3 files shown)


crookedneighbor:

PR #7686, opened 2016-06-18
Adding check to see if user is in a party before allowing the user to…

### LINKED ISSUE #7684: "Start Quest" button does nothing when not in a party
If you are not in a party, and you click on a quest scroll that you own, the "Start Quest" button does nothing when you click on it.

It should show you a message explaining that you need to create or join a party before you can start a quest.

Hus274:
#7684
### Changes

Adding a new check when the user tries to start a quest that shows a pop up if they're not in a quest so they know why the quest won't start.

screenshot from 2016-06-18 14-45-04

closes #7684

---

### DIFF
--- a/website/client/js/controllers/inventoryCtrl.js
+++ b/website/client/js/controllers/inventoryCtrl.js
@@ -1,6 +1,6 @@
 habitrpg.controller("InventoryCtrl",
-  ['$rootScope', '$scope', 'Shared', '$window', 'User', 'Content', 'Analytics', 'Quests', 'Stats', 'Social',
-  function($rootScope, $scope, Shared, $window, User, Content, Analytics, Quests, Stats, Social) {
+  ['$rootScope', '$scope', 'Shared', '$window', 'User', 'Content', 'Analytics', 'Quests', 'Stats', 'Social', 'Notification',
+  function($rootScope, $scope, Shared, $window, User, Content, Analytics, Quests, Stats, Social, Notification) {
 
     var user = User.user;
 
@@ -44,7 +44,7 @@ habitrpg.controller("InventoryCtrl",
           $scope.$close();
         });
       } else {
-        $window.alert(window.env.t('needPartyToStartQuest'))
+        Notification.text((env.t('needPartyToStartQuest')));
       }
     };
 
--- a/website/client/js/controllers/inventoryCtrl.js
+++ b/website/client/js/controllers/inventoryCtrl.js
@@ -38,10 +38,14 @@ habitrpg.controller("InventoryCtrl",
     $scope.questInit = function() {
       var key = $rootScope.selectedQuest.key;
 
-      Quests.initQuest(key).then(function() {
-        $rootScope.selectedQuest = undefined;
-        $scope.$close();
-      });
+      if( User.user.party._id ) {
+        Quests.initQuest(key).then(function() {
+          $rootScope.selectedQuest = undefined;
+          $scope.$close();
+        });
+      } else {
+        $window.alert(window.env.t('needPartyToStartQuest'))
+      }
     };
 
     // count egg, food, hatchingPotion stack totals
--- a/website/client/js/services/questServices.js
+++ b/website/client/js/services/questServices.js
@@ -5,12 +5,6 @@ angular.module('habitrpg')
   function questsFactory($http, $state, $q, ApiUrl, Content, Groups, User, Analytics) {
 
     var user = User.user;
-    var party;
-
-    Groups.party()
-      .then(function (partyFound) {
-        party = partyFound;
-      });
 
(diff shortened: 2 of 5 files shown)


Alys:
Could you please describe what effect this has from the user's point of view? I.e., how the site's behaviour changes? For example, if they click the "Start Quest" button, they see a notification saying .....  That's a good idea for any PR because it helps non-technical players follow the progress of fixes for any bugs that affect them. We also sometimes give Socialites or wiki editors links to PRs so that they can inform players about changes or adjust wiki pages, and so a short written explanation of the PR helps them.
(I've been meaning to add a guide to the wiki's Blacksmith pages with tips for creating PRs, and I'll add this to it.)

Hus274:
For sure! I'll update it now

crookedneighbor:
This should use the notification service, not an alert box.

Hus274:
Updated picture in post and repushed.

khipkin:
Why still have the "Start" button enabled if we know it won't work?  Wouldn't it be better to disable it and offer an explanation as hover/help text?

Hus274:
My concern with that is the user might just see that it's not enabled so they won't know to hover over it to see the error message. If that's what the desired behavior is though, I'd be happy to update it!

crookedneighbor:
Well, they'd have to hover over it to click it. 

Though I do recall tooltips not working very well with disabled buttons.

khipkin:
Could there be an info modal that pops up on top of the quest modal?  And the user would have to read the text and dismiss it before viewing the quest modal with the disabled button?

Alys:
@khipkin There's a reasonable chance that they'd dismiss that info modal without reading it carefully, since they'd see the thing that they were interested in underneath it and want to get straight to it.

@Hus274 You're correct that users wouldn't always see hover text on a disabled button - we notice that elsewhere.

Perhaps make the button grey but still clickable - if that's possible with bootstrap without doing clunky workarounds? Put the message on the modal under the buttons, always visible when they aren't in a party (never visible when they are). Also put the message in a notification when they click on the disabled button (as in your current screenshot) - that handles the case where the player doesn't read the text under the buttons, which I know will occur. :)

If it's not easy to make a button look disabled but still be clickable, then leave the button blue. It's not ideal but it's more important to show the notification when they click than to not give any feedback for an attempt to click. Clicking on form controls is what a player will do when they're in too much of a hurry to read all the text on the modal.

Don't worry too much about getting this perfect. We'll be redesigning the website's front end and we'll come up with a consistent method for cases like this.

Hus274:
@Alys @khipkin what about a warning button? http://v4-alpha.getbootstrap.com/components/buttons/

Alys:
"Indicates caution should be taken with this action", so perhaps that's not quite right. Maybe an info button?

Hus274:
That makes sense to me!

Hus274:
Here's the updated picture:
final

Alys:
I like that! 

However the `.text-info` style on the p tag makes it look like a link. I think the best fix would be to make it a link. :) http://habitica.wikia.com/wiki/Party

In case you don't know, we typically do links as `<a href="...">text goes here</a>` in the locales file so that translators can replace the URL with a version from the appropriate language wiki.

Hus274:
@Alys good call, I think that looks really good. Also the notification link is clickable as well!

finalv2

website/client/js/services/questServices.js
```diff
@@ -5,12 +5,6 @@ angular.module('habitrpg')
   function questsFactory($http, $state, $q, ApiUrl, Content, Groups, User, Analytics) {
 
     var user = User.user;
-    var party;
-
-    Groups.party()
```
paglias, line 10: are the changes in this file needed? They seem unrelated to this PR

website/client/js/controllers/inventoryCtrl.js
```diff
@@ -38,10 +38,14 @@ habitrpg.controller("InventoryCtrl",
     $scope.questInit = function() {
       var key = $rootScope.selectedQuest.key;
```
khipkin, line 39: Why leave this line outside the `if/else` block?

website/client/js/services/questServices.js
crookedneighbor, line 10: I agree. This refactor is outside of the scope of the PR.

website/client/js/controllers/inventoryCtrl.js
crookedneighbor, line 39: It's fine to have it here.

crookedneighbor:
I went with this instead:

screen shot 2016-06-19 at 9 20 12 am

crookedneighbor:
Seems simpler to me to remove the button and the information about what happens with a message instead. 

I've notated your work toward your next tier.

Hus274:
I like it, looks great!

Issue #7688, opened 2016-06-18
Change still shows after canceling task edit (until next sync/reload)


habitrpguser:
General Info
- UUID: 1aff9116-34b0-438c-9877-06038bdce161
- Browser: several (Firefox, Chrome)
- OS: Windows
### Description

When I edit a task, then click "Cancel", I see the new text I entered instead of the old one (as if I had clicked "Save" instead). After clicking "Sync" or reloading the page in my browser, however, the correct unaltered text is shown as it should be.

I already reported this minor bug in the "Report a Bug" guild and was asked to please write a GitHub issue. I myself have no problem with the workaround, I'm just reporting the bug for the coders. So, as far as I am concerned, please don't feel obliged to write an answer quickly, or at all!

PR #7692, opened 2016-06-19
Cancel task edit resets the task to the original values without refreshing

Hus274:
#7688
### Changes

Updating the cancel functionality of editing a task in order to reset the task to the original values if the user cancels the edit.

---


test/spec/controllers/tasksCtrlSpec.js
```diff
@@ -30,11 +30,89 @@ describe('Tasks Controller', function() {
...
+
+    it('sets _tags to false if preference for collapsed tags is turned on', function() {
+      user.preferences.tagsCollapsed = true;
+      scope.editTask(task, user);
+
+      expect(task._tags).to.eql(false);
+    });
+
+    it('sets _advanced to true by default', function(){
+      user.preferences.advancedCollapsed = true;
```
khipkin: You should update the `specHelper.newUser()` function to include this default setting.  [Here](https://github.com/HabitRPG/habitrpg/pull/6943/files#diff-ecd2e89a7b19d9ac70c6d589bb181bd9R39) is an example of where this has been done before.

test/spec/controllers/tasksCtrlSpec.js
```diff
@@ -30,11 +30,89 @@ describe('Tasks Controller', function() {
...
+      scope.editTask(task, user);
+
+      expect(task._advanced).to.eql(false);
```
khipkin: Test description says `true` but you assert `false`?

khipkin: I think your test descriptions for this test and the next one are mixed up.

test/spec/controllers/tasksCtrlSpec.js
khipkin: Whatever the default setting should actually be, not necessarily this setting. :stuck_out_tongue_winking_eye: (See below comments.)

test/spec/controllers/tasksCtrlSpec.js
```diff
@@ -30,11 +30,89 @@ describe('Tasks Controller', function() {
...
+    });
+
+    it('sets _advanced to false if preference for collapsed advance menu is turned on', function() {
+      user.preferences.advancedCollapsed = false;
+      scope.editTask(task, user);
+
+      expect(task._advanced).to.eql(true);
     });
+
```
khipkin: Extra space.

test/spec/controllers/tasksCtrlSpec.js
```diff
@@ -30,11 +30,89 @@ describe('Tasks Controller', function() {
...
+    var task;
+
+    beforeEach(function(){
+      task = specHelper.newTask();
+    });
+
+    it('resets the task text', function() {
+      let originalText = task.text;
+      scope.editTask(task, user);
+      task.text = 'test';
```
khipkin: Sanity check: assert that `originalText` is not `'test'`.

test/spec/controllers/tasksCtrlSpec.js
```diff
@@ -30,11 +30,89 @@ describe('Tasks Controller', function() {
...
+      scope.cancelTaskEdit(task)
```
khipkin: missing semicolon

test/spec/controllers/tasksCtrlSpec.js
```diff
@@ -30,11 +30,89 @@ describe('Tasks Controller', function() {
...
+      expect(task.text).to.be.eql(originalText);
+    });
+
+    it('resets the task notes', function() {
+      let originalNotes = task.notes;
+      scope.editTask(task, user);
+      task.notes = 'test';
```
khipkin: Same sanity check as above

test/spec/controllers/tasksCtrlSpec.js
```diff
@@ -30,11 +30,89 @@ describe('Tasks Controller', function() {
...
+      scope.cancelTaskEdit(task)
+      expect(task.notes).to.be.eql(originalNotes);
+    });
+
+    it('resets the task alias', function() {
+      task.alias = 'alias';
+      let originalAlias = task.alias;
+      scope.editTask(task, user);
+      task.alias = 'test';
```
khipkin: Same sanity check as above

website/client/js/controllers/tasksCtrl.js
```diff
@@ -100,8 +99,17 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
       }
     };
 
+    $scope.editTask = function(task, user) {
+      originalTasks.push(_.cloneDeep(task));
+      task._editing = !task._editing;
+      task._tags = !user.preferences.tagsCollapsed;
```
khipkin: Make sure `specHelper.newUser()` sets default values for both of these settings.

website/client/js/controllers/tasksCtrl.js
```diff
@@ -100,8 +99,17 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
...
+      task._advanced = !user.preferences.advancedCollapsed;
+      if($rootScope.charts[task._id]) $rootScope.charts[task.id] = false;
```
khipkin: Is this check necessary?

test/spec/controllers/tasksCtrlSpec.js
crookedneighbor: That sanity check doesn't make sense. originalText would be a string, so it doesn't hold onto a reference like an object would.

website/client/js/services/taskServices.js
```diff
@@ -150,13 +150,6 @@ angular.module('habitrpg')
       });
     };
 
-    function editTask(task, user) {
```
crookedneighbor, line 153: Why are you removing this from the task service? The challenges controller also uses this function.

website/client/js/controllers/tasksCtrl.js
```diff
@@ -100,8 +99,17 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
       }
     };
 
+    $scope.editTask = function(task, user) {
+      originalTasks.push(_.cloneDeep(task));
```
crookedneighbor: This seems like a weird way to deal with this. I'd model it after how groups are edited.

test/spec/controllers/tasksCtrlSpec.js
khipkin: But the originalText string might also be 'test'?

crookedneighbor: Hmm... that's true, from the spec helper.

khipkin: Exactly!  The sanity check is to make sure you aren't "changing" a string to the exact same string.

crookedneighbor:
Basically, I'd use the angular copy function to copy the task and save it to `task._edit`. Then change the edit views to reference `task._edit.whatever` instead of the properties themselves. 

When editing is cancelled, you just close the edit lane as normal and the task shows as it should. 

If it's saved, update the task with the response back from the server (might already be don't that)

Hus274:
Ahh that's a great idea, and it eliminates the possibility of the memory leak which I didn't love. I can work on getting that up and cleaning up based on the other comments and then I'll repost. Thanks for the assistance!

Hus274:
Updated according to Blade's comments. The tests are failing but they are failing for me on the main branch as well so I'm not quite sure what's wrong.

Hus274:
Looks like the test is failing because it's missing the apn npm module. After installing this module locally, the tests all passed.

test/spec/services/taskServicesSpec.js
```diff
@@ -161,8 +161,6 @@ describe('Tasks Service', function() {
     it('toggles the _editing property', function() {
```
crookedneighbor: This is no longer accurate. You made it so that it doesn't toggle it. It just sets it to true.

Alys:
@Hus274 There's a feature where a task being edited stays visible while it's open for editing even if it wouldn't normally be visible otherwise. E.g., if it's tagged "morning" and you have the "morning" tag filter active, and you edit the task to deselect that tag, the task does not disappear from view until after you've saved the edit. Another example, if there's a Daily active/grey filter active, and you change the due status of the Daily, the Daily stays visible while editing is happening. Can you test that that still works? I haven't seen anything in your code to indicate that it won't, but a quick test will be more reliable than me reading the code. :)

Hus274:
@Alys I checked both these cases, the dailies filter and the tags filter, and they both worked as expected! Definitely worth checking it out and doing the test, thanks for bringing it up! :)

Alys:
@Hus274 Awesome, thanks for testing! Trying to edit tasks was a bit of a nightmare before that feature was put in place so it's good to know it's still there. :)

test/spec/services/taskServicesSpec.js
```diff
@@ -150,6 +150,56 @@ describe('Tasks Service', function() {
...
+
+    it('sets _tags to false if preference for collapsed tags is turned on', function() {
+      user.preferences.tagsCollapsed = true;
+      tasks.editTask(task, user);
+
+      expect(task._tags).to.eql(false);
+    });
+
+    it('sets _advanced to true by default', function(){
+      user.preferences.advancedCollapsed = true;
```
khipkin, line 180: add default preference value to `SpecHelper.newUser()`

website/client/js/controllers/tasksCtrl.js
```diff
@@ -196,16 +191,15 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
...
 
     $scope.addChecklistItem = function(task, $event, $index) {
-      if (task.checklist[$index].text) {
-        $scope.saveTask(task, true);
-        if ($index === task.checklist.length - 1)
-          task.checklist.push({ completed: false, text: '' });
-        focusChecklist(task, $index + 1);
+      if (task._edit.checklist[$index].text) {
+        if ($index === task._edit.checklist.length - 1)
+          task._edit.checklist.push({ completed: false, text: '' });
```
khipkin, line 201: add curly braces

test/spec/services/taskServicesSpec.js
crookedneighbor, line 180: Not necessary.

test/spec/services/taskServicesSpec.js
```diff
@@ -200,6 +198,21 @@ describe('Tasks Service', function() {
     });
   });
 
+  describe('cancelTaskEdit', function() {
+
```
crookedneighbor: Remove this extra line break.

Hus274:
Updated formatting according to comments.

crookedneighbor:

PR #7700, opened 2016-06-20
Add licensing info to readme & fix small grammar mistakes in licensing

ExcaliburZero:
### Changes

Adds information on the licensing of the contents of the repo to the `README.md` file, as that information is currently only held within the `LICENSE` file which is generally used to contain the full contents of the license of a project, rather than the general licensing information. This makes it easier for contributors to see what license the project is under.

Also fixes some small grammar mistakes in the `LICENSE` file.

---

UUID: dea0a1fd-6cc3-4511-bf6f-9957c233e792

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -2,8 +2,9 @@
 This Source Code is subject to the terms of the GNU General Public License, v. 3.0.
 If a copy of the GPL was not distributed with this file, You can obtain one at http://www.gnu.org/licenses/gpl-3.0.txt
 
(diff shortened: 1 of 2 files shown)


crookedneighbor:
The license type is also listed in the package.json. For a node based project, including the license file and listing the type in the package.json is the standard. 

I don't think it's necessary to include that info on the README as well.

crookedneighbor:
I've extracted just the fixes to the license file itself and awarded you a contributor tier.

You can read more about [contributor tiers and rewards here](http://habitrpg.wikia.com/wiki/Contributor_Rewards).

PR #7706, opened 2016-06-22
Add tests around finishing quests

crookedneighbor:
Also adds a virtual method to return an array of participating quest members.

### DIFF
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -90,6 +90,10 @@ schema.plugin(baseModel, {
   noSet: ['_id', 'balance', 'quest', 'memberCount', 'chat', 'challengeCount'],
 });
 
+schema.virtual('quest.participatingMembers').get(function getParticipatingQuestMembers () {
+  return Object.keys(this.quest.members).filter(member => this.quest.members[member]);
+});
+
 // A list of additional fields that cannot be updated (but can be set on creation)
 let noUpdate = ['privacy', 'type'];
 schema.statics.sanitizeUpdate = function sanitizeUpdate (updateObj) {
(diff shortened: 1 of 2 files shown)


website/server/models/group.js
```diff
@@ -478,7 +485,7 @@ schema.methods.finishQuest = async function finishQuest (quest) {
     }
   });
 
-  let q = this._id === TAVERN_ID ? {} : {_id: {$in: _.keys(this.quest.members)}};
```
crookedneighbor, line 481: We were awarding quest rewards to all members, not just participating ones :(

website/server/models/group.js
```diff
@@ -439,11 +443,14 @@ schema.statics.cleanGroupQuest = function cleanGroupQuest () {
...
   let questK = quest.key;
-  let updates = {$inc: {}, $set: {}};
-
-  updates.$inc[`achievements.quests.${questK}`] = 1;
-  updates.$inc['stats.gp'] = Number(quest.drop.gp);
-  updates.$inc['stats.exp'] = Number(quest.drop.exp);
+  let updates = {
+    $inc: {
+      [`achievements.quests.${questK}`]: 1,
+      ['stats.gp']: Number(quest.drop.gp),
```
paglias: no need for square brackets here

website/server/models/group.js
```diff
@@ -90,6 +90,10 @@ schema.plugin(baseModel, {
   noSet: ['_id', 'balance', 'quest', 'memberCount', 'chat', 'challengeCount'],
 });
 
+schema.virtual('quest.participatingMembers').get(function getParticipatingQuestMembers () {
```
paglias: can we transform this in a method? I find virtual getters and setters confusing, with a method you immediately now that you're computing a result

website/server/models/group.js
khipkin, line 481: Semi-related to #7653

website/server/models/group.js
crookedneighbor: Good catch.

website/server/models/group.js
crookedneighbor: I could go either way on this, but I'm fine making it an instance method.

PR #7709, opened 2016-06-23
Deduping completed tasks on load. closes #7708

### LINKED ISSUE #7708: Duplicate completed todos
### Description

Currently, if a user checks off a todo before loading the completed todos and then proceeds to load the completed todos, the task that the user has completed in the current session is duplicated under the done todos tab.

To replicate this bug:
1. Refresh the page
2. Complete a todo
3. Click Done tab under todos
4. Look for the todo that was completed, it should have two entries in the list.
#### Console Errors

None

Hus274:
#7708
### Changes

This will prevent the user from seeing duplicate completed tasks. This would happen if the user completed tasks locally before loading the completed tasks tab.

---

### DIFF
--- a/website/client/js/controllers/tasksCtrl.js
+++ b/website/client/js/controllers/tasksCtrl.js
@@ -156,11 +156,15 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
     $scope._today = moment().add({days: 1});
 
     $scope.loadedCompletedTodos = function () {
-      if (Tasks.loadedCompletedTodos === true) return;
+      if (Tasks.loadedCompletedTodos === true) {
+        return;
+      }
+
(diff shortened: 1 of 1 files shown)


paglias:
I think this is something we missed from your earlier PR. Instead of filtering duplicate tasks it isn't simpler and faster to just substitute all local completed todos with the ones from loaded from the server?

Hus274:
I think this would have been a bug before that because I didn't change the default loading logic of loading completed todos and I didn't change the fact that non completed todos were saved locally and then completed.

Either way, I think that's a good call, just remove the completed todos from the list on a load from the server. I can update this and repost.

crookedneighbor:
Can confirm, it was a bug from before.

Hus274:
I've updated this to clear the local copy of completed messages before loading the completed messages from the server.

crookedneighbor:

PR #7710, opened 2016-06-23
Accounting for class bonus when sorting equipment by stats. closes #7701

### LINKED ISSUE #7701: Should sorting equipment on an attribute while grouped by type consider the class bonus?
As an example, when I group by type and sort by PER, the Jester Hat (PER+15) sorts before the Umbral Hood (PER+12).  But with my Rogue's class bonus, the Umbral Hood actually gives me PER+18 (12 + 6) which is better than the Jester Hat.  When grouped by class, it's not an issue since my class-specific equipment is all together.

Should the sorting take class-bonus into consideration?

Hus274:
#7701
### Changes

Updated the equipment sorting so that ordering the equipment by stats will take into account the class bonus gained by having those items equipped according to the [wiki](http://habitica.wikia.com/wiki/Class_System).

---

### DIFF
--- a/website/client/js/controllers/inventoryCtrl.js
+++ b/website/client/js/controllers/inventoryCtrl.js
@@ -62,6 +62,14 @@ habitrpg.controller("InventoryCtrl",
 
         var item = Content.gear.flat[key];
 
+        if (item.klass == User.user.stats.class)
+        {
+          item.str = item.str * 1.5;
+          item.con = item.con * 1.5;
+          item.per = item.per * 1.5;
+          item.int = item.int * 1.5;
+        }
+
         if (!$scope.gearByClass[item.klass]) {
           $scope.gearByClass[item.klass] = [];
         }
--- a/website/client/js/controllers/inventoryCtrl.js
+++ b/website/client/js/controllers/inventoryCtrl.js
@@ -62,18 +62,15 @@ habitrpg.controller("InventoryCtrl",
 
         var item = Content.gear.flat[key];
 
-        if (item.klass == User.user.stats.class || item.specialClass == User.user.stats.class) {
-          item._effectiveStr = item.str * 1.5;
-          item._effectiveCon = item.con * 1.5;
-          item._effectivePer = item.per * 1.5;
-          item._effectiveInt = item.int * 1.5;
+        var bonusMultiplier = 1;
+        if (item.klass === User.user.stats.class || item.specialClass === User.user.stats.class) {
+          bonusMultiplier = 1.5;
         }
-        else {
-          item._effectiveStr = item.str;
-          item._effectiveCon = item.con;
-          item._effectivePer = item.per;
-          item._effectiveInt = item.int;
-        }
+
+        item._effectiveStr = item.str * bonusMultiplier;
+        item._effectiveCon = item.con * bonusMultiplier;
+        item._effectivePer = item.per * bonusMultiplier;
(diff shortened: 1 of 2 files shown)


website/client/js/controllers/inventoryCtrl.js
```diff
@@ -62,6 +62,19 @@ habitrpg.controller("InventoryCtrl",
 
         var item = Content.gear.flat[key];
 
+        if (item.klass == User.user.stats.class || item.specialClass == User.user.stats.class) {
```
crookedneighbor: Use triple equals.

website/client/js/controllers/inventoryCtrl.js
```diff
@@ -62,6 +62,19 @@ habitrpg.controller("InventoryCtrl",
...
+          item._effectiveStr = item.str * 1.5;
+          item._effectiveCon = item.con * 1.5;
+          item._effectivePer = item.per * 1.5;
+          item._effectiveInt = item.int * 1.5;
+        }
```
crookedneighbor: Put the else on the same line.

website/client/js/controllers/inventoryCtrl.js
```diff
@@ -62,6 +62,19 @@ habitrpg.controller("InventoryCtrl",
 
         var item = Content.gear.flat[key];
```
crookedneighbor, line 64: How about:

``` js
var bonusMultiplier = 1;

if (item.klass === User.user.stats.class || item.specialClass === User.user.stats.class) {
  bonusMultiplier = 1.5;
}

item._effectiveStr = item.str * bonusMultiplier;
item._effectiveCon = item.con * bonusMultiplier;
item._effectivePer = item.per * bonusMultiplier;
item._effectiveInt = item.int * bonusMultiplier;
```

website/client/js/controllers/inventoryCtrl.js
crookedneighbor: Stupid autocorrect. Put the else on the same line.

website/client/js/controllers/inventoryCtrl.js
Hus274, line 64: Good call, this is much cleaner.

website/client/js/controllers/sortableInventoryCtrl.js
```diff
@@ -1,10 +1,10 @@
 habitrpg.controller('SortableInventoryController', ['$scope',
   function ($scope) {
     var attributeSort = {
-      constitution: ['-con', '-(con+int+per+str)'],
-      intelligence: ['-int', '-(con+int+per+str)'],
-      perception: ['-per', '-(con+int+per+str)'],
-      strength: ['-str', '-(con+int+per+str)'],
+      constitution: ['-(_effectiveCon)', '-(_effectiveCon+_effectiveInt+_effectivePer+_effectiveStr)'],
```
khipkin, line 4: Maybe assign the value `'-(_effectiveCon+_effectiveInt+_effectivePer+_effectiveStr)'` to a var and then use that var four times instead

crookedneighbor, line 4: I think this is fine.

crookedneighbor, line 4: This is verbose, but it easily conveys what is happening. I don't think it's worth over engineering the client side code before we do the rewrite anyway.

PR #7711, opened 2016-06-23
Elastic Beanstalk: update NPM to version 3

paglias:


crookedneighbor:
Seems reasonable to me.

paglias:
After this goes live we should try again to use NPM Shrinkwrap

PR #7733, opened 2016-06-28
Add two unit tests for CRON_SAFE_MODE

### LINKED ISSUE #7464: tests are needed for cron safe mode and cron semi-safe mode
This is so that I'm kept accountable for doing this, or at least guiding a contributor through doing it.

In PR https://github.com/HabitRPG/habitrpg/pull/7286, "cron safe mode" was implemented. It causes cron to behave differently than normal, with the differences described here: https://github.com/HabitRPG/habitrpg/issues/7161#issuecomment-218096217

In PR https://github.com/HabitRPG/habitrpg/pull/7462, "cron semi-safe mode" was implemented. It allows cron to behave normally in all ways except that a quest boss does not cause damage, and the the message in party chat when you attack the quest boss explains that the boss has not caused damage.

We need tests for all of that. Current cron tests are in:
- ./test/api/v3/integration/debug/POST-debug_set-cron.test.js
- ./test/api/v3/unit/libs/cron.test.js
- ./test/api/v3/unit/middlewares/cronMiddleware.js

cmcarthur:
Fixes #7464
### Changes

This introduces two simple unit tests for `CRON_SAFE_MODE`. It's my first PR so I wanted to keep it small. I copied the style of this test from `email.test.js`, but this style of reloading strikes me as hacky, so I also wanted to clear that with a maintainer before pushing forward. If this gets through, I'll do another pass and add more than two tests :D

Tests are green.

---

UUID: 859e77a3-cb68-4cf0-8cc2-b84c9adf18ce

### DIFF
--- a/test/api/v3/unit/libs/cron.test.js
+++ b/test/api/v3/unit/libs/cron.test.js
@@ -2,6 +2,7 @@
 import moment from 'moment';
 import nconf from 'nconf';
 import Bluebird from 'bluebird';
+import requireAgain from 'require-again';
 import { recoverCron, cron } from '../../../../../website/server/libs/api-v3/cron';
 import { model as User } from '../../../../../website/server/models/user';
 import * as Tasks from '../../../../../website/server/models/task';
@@ -9,6 +10,8 @@ import { clone } from 'lodash';
 import common from '../../../../../common';
 
 // const scoreTask = common.ops.scoreTask;
+
+let pathToCronLib = '../../../../../website/server/libs/api-v3/cron';
 
 describe('cron', () => {
   let user;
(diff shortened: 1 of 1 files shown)


crookedneighbor:
I agree that it is a little hacky. We should probably make the cron safe and semi safe mode options be options you pass into cron, rather than environmental variables. I.e., have the file that calls cron define those variables and pass them in. 

Not a concern for this PR though.

crookedneighbor:
Also, you have some lint errors.

Alys:
@crookedneighbor Certainly not a concern for this PR, but to add my thoughts to your comment:
I'd be worried about needing to modify code to turn cron safe mode on and off. In future we might need safe mode after an unexpected outage, and then we'd want to enable it very quickly to limit the number of players who took unfair damage from cron. An environment variable lets us do that, but editing code and deploying can take many minutes, during which some players would cron. It might also be at a time when we're doing other fire-fighting and wouldn't want the extra work involved in the code change and deployment.

crookedneighbor:
Sorry, I wasn't clear. We'd still use an environmental variable, but it would be defined in cron middleware and passed into the cron function.

cmcarthur:
@crookedneighbor just my 2 cents, but isn't it more straightforward to call

```
let CRON_SAFE_MODE = nconf.get('CRON_SAFE_MODE') === 'true';
```

at the beginning of the `cron` function? then we don't need to go out and modify all the callers of `cron`, we don't need to add arguments, and each call to `cron` will have its own local `CRON_SAFE_MODE` flag.

cmcarthur:
I can work on adjusting CRON_SAFE_MODE next. Should I make an issue for that?

cmcarthur:
Build is passing now -- this is ready for another look.

crookedneighbor:
Sorry for the _loooooong_ delay.

PR #7742, opened 2016-06-30
update documentation for marking chat seen

eshapard:
Fixes put_issue_url_here
### Changes

Changed documentation for marking group/guild chat as read. The documentation included the chatId in the url, but this is not used by the api and will cause a 404 error.

---

UUID: 9173198f-4777-4a35-9f03-8aea24f51782

Documentation included the chatId in the url, but the api does not use it. Adding the chatId will cause a 404 error. The api marks all chats in the guild as read; not individual chats.

### DIFF
--- a/website/server/controllers/api-v3/chat.js
+++ b/website/server/controllers/api-v3/chat.js
@@ -305,7 +305,7 @@ api.clearChatFlags = {
 };
 
 /**
- * @api {post} /api/v3/groups/:groupId/chat/:chatId/seen Seen a group chat message
+ * @api {post} /api/v3/groups/:groupId/chat/seen Seen a group chat message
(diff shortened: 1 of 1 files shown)


khipkin:
Since you're already fixing documentation in this file, can you also change `PostCat`->`PostChat` [on this line](https://github.com/eshapard/habitrpg/blob/fdf95121f7f531180fa0ba045d9918d53c046a90/website/server/controllers/api-v3/chat.js#L55)? :smile_cat:

crookedneighbor:
I made a slight adjustment to the description of that route and squashed the commit. I've notated this toward your next tier. 

@khipkin Thanks for finding that typo. Fixed too.

PR #7746, opened 2016-07-01
Trimming username while registering

### LINKED ISSUE #7725: login usernames can include spaces at the end - adjust API's login route to prevent that
It's possible to register for Habitica with a login name that has a space at the end and this space is saved in the database as part of the login name.

This means that you can't log in unless you realise that the space was added and enter it along with the rest of your login name.

You can't register like that on the website (I've just tested and the space is removed) so I'm guessing that it's happening on one of the mobile apps, and it might also happen with third-party applications that allow registrations.

I suggest that the API's `user/auth/local/register` route be adjusted so that it removes any leading and trailing whitespace on any of the input parameters before it creates the account. Note that internal whitespace on login names should NOT be removed.

I'll leave this open for a day and if there's no complaints or better ideas, I'll mark it as help welcome.

sougat818:
Fixes #7725 
### Changes

Trimmed the username input while registering

---

UUID: 7a502714-1e43-4345-89f6-ec4248d2b761

### DIFF
--- a/test/api/v3/integration/user/auth/POST-register_local.test.js
+++ b/test/api/v3/integration/user/auth/POST-register_local.test.js
@@ -33,6 +33,23 @@ describe('POST /user/auth/local/register', () => {
       expect(user.auth.local.username).to.eql(username);
     });
 
+    it('registers a new user with space in username', async () => {
+      let username = generateRandomUserName();
+      let email = `${username}@example.com`;
+      let password = 'password';
+      let usernameWithSpaces = "  " + username + "  ";
+      let user = await api.post('/user/auth/local/register', {
+        username: usernameWithSpaces,
+        email,
+        password,
+        confirmPassword: password,
+      });
+
+      expect(user._id).to.exist;
+      expect(user.apiToken).to.exist;
+      expect(user.auth.local.username).to.eql(username);
+    });
+
     it('requires password and confirmPassword to match', async () => {
       let username = generateRandomUserName();
       let email = `${username}@example.com`;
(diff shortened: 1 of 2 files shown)


crookedneighbor:
This needs a test around this behavior.

test/api/v3/integration/user/auth/POST-register_local.test.js
```diff
@@ -33,6 +33,23 @@ describe('POST /user/auth/local/register', () => {
       expect(user.auth.local.username).to.eql(username);
     });
 
+    it('registers a new user with space in username', async () => {
```
crookedneighbor: This should be more explicit to say that it removes the spaces.

crookedneighbor:
Sorry this took so long for me to look at again.

This is tricky, because:

A) when someone goes to login and they expect their trailing username to have trailing spaces it will fail
B) we probably need to do some db querying to see if there are duplicate usernames with extra spaces in the beginning or end
C) I bet this can be done on the model level http://mongoosejs.com/docs/api.html#schema_string_SchemaString-trim

paglias:
I think the best thing to do is to send a validation error if the username contains spaces so the user will have to fix it before proceeding.

@sougat818 if you have time, we would like new usernames to allow only letters, numbers, hyphens and dashes. All the old users should be grandfathered which means the validation should be in the signup route and not in the model. If you can't do it, don't worry, we'll do it later

crookedneighbor:
I'm going to close this for now, but re-open it if you still want to work on it and have made the requested changes.

sougat818:
@paglias So I suppose there is no need to trim the username now.

crookedneighbor:
Please see @paglias's comment

> if you have time, we would like new usernames to allow only letters, numbers, hyphens and dashes. All the old users should be grandfathered which means the validation should be in the signup route and not in the model. If you can't do it, don't worry, we'll do it later

PR #7760, opened 2016-07-05
Added FCM push notifications

TheHollidayInn:
Fixes put_issue_url_here
### Changes

This adds FCM push notifications to the push notification library. This was suggested by Google to upgrade to Firebase.

---

UUID:


website/server/libs/api-v3/pushNotifications.js
```diff
@@ -92,6 +96,25 @@ module.exports = function sendNotification (user, details = {}) {
             data: payload,
           });
         }
+
+        if (fcm) {
+          var message = {
```
paglias: let

website/server/libs/api-v3/pushNotifications.js
```diff
@@ -92,6 +96,25 @@ module.exports = function sendNotification (user, details = {}) {
...
+              data: payload,
+              notification: {
+                  title: details.title,
+                  body: details.message,
+              }
+          };
+
+          fcm.send(message, function(err, response){
+              if (err) {
+                  console.log(err, "Something has gone wrong!");
```
paglias: should use the logger module

website/server/libs/api-v3/pushNotifications.js
```diff
@@ -92,6 +96,25 @@ module.exports = function sendNotification (user, details = {}) {
...
+              } else {
+                  console.log("Successfully sent with response: ", response);
```
paglias: this should be removed, if the notification is sent correctly this isn't really necessary

paglias:
I see some problems here:
- registration ids used for gcm can be used also for fcm?
- the code right now will send the notification twice, once for GCM and once for FCM

also there are some conflicts

paglias:
Also it looks like our current library uses node-gcm that supports FCM too https://github.com/ToothlessGear/node-gcm/issues/255. But push-notify doesn't currently have the latest version of node-gcm

website/server/libs/api-v3/pushNotifications.js
TheHollidayInn: Cool, I wasn't sure where to log this information.

TheHollidayInn:
Yes, regIds work for both. And, I think I should remove the GCM code since there is nothing implemented to use it on the Android side.

test/api/v3/unit/libs/pushNotifications.js
```diff
@@ -17,9 +19,12 @@ describe('pushNotifications', () => {
   beforeEach(() => {
     user = new User();
     gcmSendSpy = sinon.spy();
+    fcmSendSpy = sinon.spy();
     apnSendSpy = sinon.spy();
 
-    sandbox.stub(nconf, 'get').returns('true');
+    sandbox.stub(nconf, 'get').returns('true-key');
+
+    sandbox.stub(FCM.prototype, "send", fcmSendSpy)
```
khipkin: missing semicolon

TheHollidayInn:
Updated!

package.json
```diff
@@ -31,6 +31,9 @@
     "express": "~4.13.3",
     "express-csv": "~0.6.0",
     "express-validator": "^2.18.0",
+    "fcm-push": "^1.0.7",
+    "firebase": "^2.2.9",
```
paglias: Are these firebase modules necessary?

website/server/libs/api-v3/pushNotifications.js
```diff
@@ -7,8 +7,12 @@ import Bluebird from 'bluebird';
...
   S3,
 } from './aws';
+import FCM from 'fcm-push';
 
 const GCM_API_KEY = nconf.get('PUSH_CONFIGS:GCM_SERVER_API_KEY');
+const FCM_API_KEY = nconf.get('PUSH_CONFIGS:FCM_SERVER_API_KEY');
+
+let fcm = FCM_API_KEY ? new FCM(FCM_API_KEY) : undefined;
 
 let gcm = GCM_API_KEY ? pushNotify.gcm({
```
paglias: gcm is still being setup here even if it's not used

TheHollidayInn:
Updated.

test/api/v3/unit/libs/pushNotifications.js
```diff
@@ -40,14 +45,16 @@ describe('pushNotifications', () => {
 
   it('throws if user is not supplied', () => {
     expect(sendPushNotification).to.throw;
-    expect(gcmSendSpy).to.not.have.been.called;
+    // expect(gcmSendSpy).to.not.have.been.called;
```
crookedneighbor: Why keep this around?

website/server/libs/api-v3/pushNotifications.js
```diff
@@ -7,19 +7,23 @@ import Bluebird from 'bluebird';
...
-  retries: 3,
-}) : undefined;
+let fcm = FCM_API_KEY ? new FCM(FCM_API_KEY) : undefined;
 
-if (gcm) {
-  gcm.on('transmissionError', (err, message, registrationId) => {
-    logger.error('GCM Error', err, message, registrationId);
-  });
-}
+// let gcm = GCM_API_KEY ? pushNotify.gcm({
```
crookedneighbor: Why are we keeping this code here?

TheHollidayInn: @paglias mentioned above that the push notify library supports FCM but doesn't have the update version of node-gcm. So, I figured if we would leave this code incase they update. I also suggested removing it - so either way is fine with me.

TheHollidayInn:
Updated

TheHollidayInn:
Updated

PR #7761, opened 2016-07-05
Deprecate API v2

paglias:


tasks/gulp-tests.js
```diff
@@ -404,100 +322,4 @@ gulp.task('test:api-v3', (done) => {
...
-
-  if (totals[1] > 0) {
-    console.error('ERROR: There are failing tests!');
-    process.exit(1);
-  } else {
-    console.log('\n\x1b[36mThanks for helping keep Habitica clean!\x1b[0m');
-    process.exit();
-  }
-});*/
+});
```
khipkin, line 325: nit: add nl

website/views/shared/footer.jade
```diff
@@ -63,8 +63,6 @@ footer.footer(ng-controller='FooterCtrl')
             a(target='_blank', href='http://devs.habitica.com')=env.t('devBlog') + ' - The Forge'
           li
             a(target='_blank', href='/apidoc')=env.t('APIv3')
-          li
-            a(target='_blank', href='/static/api-v2')=env.t('APIv2')
```
khipkin, line 67: This seems to be the only use of this localized string in the repo.  Can it be removed from settings.json?

tasks/gulp-tests.js
crookedneighbor, line 325: I don't think we have a policy on trailing new lines on file endings. I added an editorconfig file that will auto format if you have the plugin for your preferred text editor, but it never really caught on.

khipkin, line 325: Oh my bad.  I think it's generally considered best-practice to have trailing newlines though.  Hence GitHub's angry little red thing when there's no trailing newline.

website/views/shared/footer.jade
paglias, line 67: removed

Issue #7728, opened 2016-06-26
Fill in tests for payments lib


crookedneighbor:
The [test file for the payments lib](https://github.com/HabitRPG/habitrpg/blob/81ac45f2a7ccabc8b31ae5856592894ff3a24109/test/api/v3/unit/libs/payments.test.js) is under tested. I've filled in the it blocks I believe need to be added.

Hus274:
I can work on filling these in.

PR #7770, opened 2016-07-10
Completing tests for payments. Wasn't able to mock out pushNotificati…

Hus274:
#7728
### Changes

Filled in the payments test library. I haven't been able to figure out how to properly spy on the push notifications method. I tried importing \* and then using notifications.default but that still wasn't able to properly inject the spy into the payments module.

Can someone provide some advice on how to finish those up and properly spy on them? I've left //TODO above the two tests that aren't working.

---

### DIFF
--- a/test/api/v3/unit/libs/payments.test.js
+++ b/test/api/v3/unit/libs/payments.test.js
@@ -1,7 +1,6 @@
 import * as sender from '../../../../../website/server/libs/api-v3/email';
 import * as api from '../../../../../website/server/libs/api-v3/payments';
 import analytics from '../../../../../website/server/libs/api-v3/analyticsService';
-import notify from '../../../../../website/server/libs/api-v3/pushNotifications';
 import { model as User } from '../../../../../website/server/models/user';
 import nconf from 'nconf';
 import moment from 'moment';
@@ -61,8 +60,8 @@ describe('payments/index', () => {
           member: recipient,
           subscription: {
             key: 'basic_3mo',
-            months: 3
-          }
+            months: 3,
+          },
         };
       });
 
@@ -79,22 +78,18 @@ describe('payments/index', () => {
         recipient.purchased.plan.dateTerminated = dateTerminated;
 
         await api.createSubscription(data);
-        expect(recipient.purchased.plan.dateTerminated).to.eql( moment(dateTerminated).add(3, 'months').toDate());
+        expect(recipient.purchased.plan.dateTerminated).to.eql(moment(dateTerminated).add(3, 'months').toDate());
       });
 
       it('sets a dateTerminated date for a user without an existing subscription', async () => {
         expect(recipient.purchased.plan.dateTerminated).to.not.exist;
-
         await api.createSubscription(data);
-
         expect(recipient.purchased.plan.dateTerminated).to.exist;
       });
 
       it('sets plan.dateUpdated if it did not previously exist', async () => {
         expect(recipient.purchased.plan.dateUpdated).to.not.exist;
-
         await api.createSubscription(data);
-
         expect(recipient.purchased.plan.dateUpdated).to.exist;
       });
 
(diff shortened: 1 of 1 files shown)


test/api/v3/unit/libs/payments.test.js
```diff
@@ -17,38 +53,125 @@ describe('payments/index', () => {
...
+        await api.createSubscription(data);
 
-      it('tracks subscription purchase as gift (if prod)');
+        expect(user.purchased.txnCount).to.eql(1);
+      });
+
+      it('sends a private message about the gift', async () => {
+        await api.createSubscription(data);
+        expect(user.sendMessage).to.be.calledOnce;
+        expect(user.sendMessage).to.be.calledWith(recipient, `\`Hello recipient, sender has sent you 3 months of subscription!\``);
```
khipkin: Would here be the right place to check that the message is properly localized? (i.e., if the sender speaks Japanese and the recipient speaks French, the recipient should be able to read the message in French.)

paglias: To check that the message is properly localized is not really necessary as we test the underlying translation function. But instead of testing for the exact string you should use `t('stringKey')` so it won't be necessary to change the test when we change the string's content

crookedneighbor:
You have a lot of linting errors to clean up. Ping me when those are sorted out and then I'll review this.

Hus274:
My mistake, I was running just this test file and missed these. I'll clean them up, sorry.

test/api/v3/unit/libs/payments.test.js
```diff
@@ -216,20 +404,72 @@ describe('payments/index', () => {
...
-      it('sends a message from purchaser to recipient');
+        data.gift = {
+          gems: {
+            amount: 4,
+          },
+          member: recipient,
+        };
+      });
+
+      it('calculates balance from gem amount if gift', async () => {
```
khipkin, line 426: inconsistent use of "if gift" phrasing in the Gift context block

Hus274:
I've updated these to pass the lint tests. I've noticed that the tests pass when I run mocha on the sub-directory but it's failing inconsistently when I run npm test or gulp test:api-v3:unit. My guess is that it's due to some asynchronous interference but I can't really find anything on it. 

Do you have an idea on why this would be happening?

test/api/v3/unit/libs/payments.test.js
```diff
@@ -17,38 +52,111 @@ describe('payments/index', () => {
...
-      it('sets a dateTerminated date for a user without an existing subscription');
+      it('adds extra months to an existing subscription', async () => {
+        recipient.purchased.plan = plan;
+        expect(recipient.purchased.plan.extraMonths).to.eql(0);
+        await api.createSubscription(data);
+        expect(recipient.purchased.plan.extraMonths).to.eql(3);
+      });
 
-      it('sets plan.dateUpdated if it did not previously exist');
+      it('udpates date terminated for an existing plan with a terminated date', async () => {
```
crookedneighbor, line 75: udpates -> updates

khipkin:
Is it possible for tests to fail due to underspecification?

Some observations:
- You have four tests with `expect(sender.sendTxn).to.be.calledOnce`.
- In one of those tests, you follow that up with `expect(sender.sendTxn).to.be.calledWith(...)`.
- The other three are the only tests that are failing.

crookedneighbor:
No. That's irrelevant. If you look at the test results, you'll see that it expected the function to be called once, but it wasn't called at all.

crookedneighbor:
My best guess is that some other test isn't mocking sendTxn correctly and causing issues for these tests. The other two test files that import it are:

```
test/api/v3/unit/libs/email.test.js
test/api/v3/unit/models/group.test.js
```

And, indeed, if I set those test files to skip, it works.

crookedneighbor:
Looks like the problem is in `test/api/v3/unit/libs/email.test.js`

crookedneighbor:
I'm not sure it's even necessary to check for IS_PROD in the payments library. It only prevents sending an email and tracking analytics, which are already handled within those libraries.

crookedneighbor:
Okay, I think I've figured out the issues. Will try to get this finished and merged in today.

crookedneighbor:
Notated.

PR #7773, opened 2016-07-11
"allCompletedTodos" option in API for returning all completed To-Dos for when 30 isn't enough

Alys:
**EDIT:   An important note for anyone thinking of using this new feature: We will be changing how it works in a few days. The `_allCompletedTodos` option will be removed. It will be replaced by a different method of specifying that you want to return all completed To-Dos at once. It is recommended that you do NOT use `_allCompletedTodos`**

Fixes https://github.com/HabitRPG/habitrpg/issues/7301
### Changes

Adds a new option to the `tasks/user` route to allow returning all completed To-Dos at once, as suggested by paglias at https://github.com/HabitRPG/habitrpg/issues/7301#issuecomment-223800764

This is NOT ready for merging. There's a lint error in the test at `await user.post(`/tasks/${id}/score/up`);` because it's in a loop and I suspect you'll want the final test `it('returns only some completed todos if req.query.type is "completedTodos" or "allCompletedTodos"'` broken up into two, one for each option. However I want to see if the basic approach is okay before I finish off the details.

### DIFF
--- a/test/api/v3/integration/tasks/GET-tasks_user.test.js
+++ b/test/api/v3/integration/tasks/GET-tasks_user.test.js
@@ -64,7 +64,7 @@ describe('GET /tasks/user', () => {
     expect(todos.every(task => task.completed === false));
   });
 
-  it('returns completed todos sorted by reverse completion date if req.query.type is "completeTodos"', async () => {
+  it('returns completed todos sorted by reverse completion date if req.query.type is "completedTodos"', async () => {
     let todo1 = await user.post('/tasks/user', {text: 'todo to complete 1', type: 'todo'});
     let todo2 = await user.post('/tasks/user', {text: 'todo to complete 2', type: 'todo'});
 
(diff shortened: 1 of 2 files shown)


test/api/v3/integration/tasks/GET-tasks_user.test.js
```diff
@@ -81,4 +81,50 @@ describe('GET /tasks/user', () => {
...
+      todosInput[i] = {text: 'todo to complete ${i}', type: 'todo'};
+    }
+    let todos = await user.post('/tasks/user', todosInput);
+    await user.sync();
+    let initialTodoCount = user.tasksOrder.todos.length;
+
+    for (let i = 0; i < numberOfTodos; i++) {
+      let id = todos[i]._id;
+
+      await user.post(`/tasks/${id}/score/up`);
```
paglias: the linting error is because code written this way will be executed sequentially and not in parallel, but in this case it's fine, just add `// eslint-disable-line no-await-in-loop` next to the line with `await`

website/server/controllers/api-v3/tasks.js
```diff
@@ -204,6 +209,7 @@ api.getUserTasks = {
   async handler (req, res) {
     let types = Tasks.tasksTypes.map(type => `${type}s`);
     types.push('completedTodos');
+    types.push('allCompletedTodos');
```
paglias: this can be condensed in a single call `types.push('completedTodos', 'allCompletedTodos')`

paglias:
Looks good to me!

Alys:
Thanks! Those two changes are made.

Alys:
@crookedneighbor How do you feel about the `returns only some completed todos if req.query.type is "completedTodos" or "allCompletedTodos"` test? Do you want it split into two tests, one for each query.type option? If so, what's the best way to avoid duplicated code and duplicated creation of the 31 tasks?

crookedneighbor:
I'm :+1: on the beta stuff

Alys:
@crookedneighbor Awesome, thanks! I was about to ask. :) I'll tidy it up in a few days and use your limit idea.

For my reference: `tasks/user?type=completedTodos&limit=0` Make limit a number, 0 for returning all. Limit only applies to completed todos.  Might need to be coerced into a number (querystrings are always strings).

PR #7781, opened 2016-07-13
Leaving a group or guild no longer removes you from challenges associated with those groups (merged in #8517)

### LINKED ISSUE #7764: Leaving the guild removes you from the challenges even though it appears you are still participating
### General Info

Opera 37.0.2178.43
Windows 8.1 32-bit
### Description

Leaving the guild removes you from list of participants of the challenges even though it appears you are still participating. I just checked with [this challenge](https://habitica.com/#/options/groups/challenges/0e44b906-86bf-46ef-9733-b99d4fd6404c). States 2 Participants but only @R@chel states is there and can win the challenge. I have rejoin the guild to no affect. I have not received a broken megaphone to remove the tasks.

_Update:_ @R@chel  has left the challenge but I have asked her to rejoin.
### Desired Affect

You don't want to be automatically removed from challenges when leaving the guild. There are some regular challenges I complete but aren't actually apart of the guild as I have no desire to stay in the chat. At the very least  it should warn you if you will be removed from the challenges.

Hus274:
_edit by Alys 2017-02-28: This was accepted and was merged in https://github.com/HabitRPG/habitica/pull/8517_

closes #7764
### Changes

Leaving a group or a guild no longer removes the user from the challenges of that group or guild.

---


website/server/models/challenge.js
```diff
@@ -240,13 +240,14 @@ schema.methods.unlinkTasks = async function challengeUnlinkTasks (user, keep) {
   };
 
   removeFromArray(user.challenges, challengeId);
+  this.memberCount--;
```
crookedneighbor, line 240: Good catch.

website/server/controllers/api-v3/groups.js
```diff
@@ -329,7 +329,6 @@ api.rejectGroupInvite = {
  * @apiGroup Group
  *
  * @apiParam {string} groupId The group _id ('party' for the user party and 'habitrpg' for tavern are accepted)
- * @apiParam {string="remove-all","keep-all"} keep Query parameter - Whether to keep or not challenges' tasks. Defaults to keep-all
```
crookedneighbor: Removing this behavior is a breaking change. So we can't do it. I think it would be better to correct this behavior and have `keep-all` keep challenges _and_ tasks, while `remove-all` should remove challenges _and_tasks. Does that make sense?

Hus274: Yeah that makes sense, I can look into that.

crookedneighbor:
@Hus274 What's the status on this?

Hus274:
In progress. Going to try and get it finished up this week.

Hus274:
I've updated this to not be a breaking change and migrated to client-old.

website/server/models/group.js
```diff
@@ -750,15 +750,18 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
     $inc: {memberCount: -1},
   };
 
-  let challenges = await Challenge.find({
-    _id: {$in: user.challenges},
-    group: group._id,
-  });
+  // only remove user from challenges if it's set to remove-all
+  if (keep === 'remove-all') {
```
crookedneighbor: I think the API docs for the group leave route need to be updated https://github.com/HabitRPG/habitrpg/blob/b54441a637b91ddd80e370cd78885514c8056b3e/website/server/controllers/api-v3/groups.js#L362-L372

Since `keep` no longer defaults to anything, right? We should also clarify better what the keep queryParam means when passing it in.

crookedneighbor (changes requested):
Looks good, other than my one comment about documentation. :+1:

website/server/models/group.js
Hus274: I've updated the documentation here. I left the "defaults to keep-all" section because that's still true: https://github.com/HabitRPG/habitrpg/pull/7781/files#diff-b6f239870392ed0ec2a2b73cbebaa699R747

Hus274:
I've updated the documentation, let me know if this works.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -93,7 +93,7 @@ describe('POST /groups/:groupId/leave', () => {
 
           let userWithChallengeTasks = await leader.get('/user');
 
-          expect(userWithChallengeTasks.challenges).to.not.include(challenge._id);
+          expect(userWithChallengeTasks.challenges).to.include(challenge._id);
```
crookedneighbor: I think the it block of this test should specify that the user remains in the challenge.

website/server/controllers/api-v3/groups.js
```diff
@@ -366,7 +366,7 @@ api.rejectGroupInvite = {
  * @apiGroup Group
  *
  * @apiParam {String} groupId The group _id ('party' for the user party and 'habitrpg' for tavern are accepted)
- * @apiParam {String="remove-all","keep-all"} keep Query parameter - Whether to keep or not challenges' tasks. Defaults to keep-all
+ * @apiParam {String="remove-all","keep-all"} keep Query parameter - Whether or not to keep challenges belonging to the group being left. Defaults to keep-all
```
crookedneighbor: This should follow the APIDoc way of designating a default param: http://apidocjs.com/#param-api-param and use the Query param group (newly created)

``` js
* @apiParam (Query) {String="remove-all","keep-all"} keep=keep-all  Whether or not to keep challenges belonging to the group being left.
```

website/server/models/group.js
crookedneighbor: Hmm... See, in prod now, you always leave the challenge and the `keep` query param determines if you keep the tasks or have them removed. I _think_ the change you're doing is technically a breaking change.

I think it makes _more_ sense to provide a body parameter for keeping the challenges (in which case the query param is ignored). Then we just update the web app's calls to provide that parameter. 

3rd parties using this route will still remove all the challenges, but it won't be a breaking change.

crookedneighbor (changes requested):

khipkin:
@Hus274 What's the status of this?

Hus274:
Ugh, I definitely lost track of this one. I might be able to look into this this week but most likely won't be able to until next week as work is crazy now currently. If you'd like to pick it up, feel free as it looks like there's a fair bit of refactoring to do. Otherwise I'll take a look when I get a chance.

khipkin:
@Hus274 Nah it's totally yours! I just wanted to check that you hadn't forgotten about it, since this is a PR that I def already thought was merged into dev. 😊

Hus274:
I've updated this according to @crookedneighbor 's comments so it's not a breaking change by using a body parameter that the client side now sends that keeps the user in the group's challenges. I've also updated the test coverage to be slightly more specific and to cover the new case.

Let me know if anything else is needed here!

website/server/models/group.js
```diff
@@ -833,21 +833,24 @@ schema.statics.tavernBoss = async function tavernBoss (user, progress) {
...
-  let challenges = await Challenge.find({
-    _id: {$in: user.challenges},
-    group: group._id,
-  });
+  // only remove user from challenges if it's set to leave-challenges
+  if (keepChallenges === 'leave-challenges') {
+    let challenges = await Challenge.find({
+      _id: {$in: user.challenges},
+      group: group._id,
+    });
```
paglias: missing .exec() (Note: I'm commenting on all PRs that have missing `.exec()` calls, not reviewing entire PRs)

paglias (changes requested):

Hus274:
Fixed the merge conflicts here and added the exec() method to the end of the mongoose call.

khipkin:
Ping to reviewers!

TheHollidayInn:
Noted towards your Tier!

Alys:
FYI for the record:

We're having some trouble on our test server with the PRs that have been merged in recently, so I've started testing each one individually.  I tested this one on my local install by fetching upstream, then checking out 68a042cdb97bf00b496328a79660be97d150b660 and doing `npm start`. My testing has been VERY brief because I just wanted to follow on from my comments at https://github.com/HabitRPG/habitica/pull/8469#issuecomment-283013476

Testing locally with 68a042cdb97bf00b496328a79660be97d150b660 shows that tasks CAN be edited so the edit bug described in https://github.com/HabitRPG/habitica/pull/8469#issuecomment-283013476 does not come from this PR.

Alys:
I've had two reports that this bug is still happening so I'm reopening this. I'll test it myself later when I get a chance.

Alys:
Ugh, sorry, that was meant to go on the relevant issue.

PR #7785, opened 2016-07-13
Added initial code for creating and reading group tasks

TheHollidayInn:
Fixes put_issue_url_here
### Changes

This is the initial code to allow groups to create tasks. We can then start on assigning, claiming, approving.

A few things to note:
- We may need to add a property to groups to verify that the group has been paid for (or maybe a subscription model like users)
- I added in the comments a suggestion to use a TaskFactory. I can show some examples, but we should probably move the lingering code out
- I think we should also separate the task controller to separate files (group, challenge, user)

---

UUID:

### DIFF
--- a/common/locales/en/groups.json
+++ b/common/locales/en/groups.json
@@ -192,5 +192,6 @@
     "uuidsMustBeAnArray": "User ID invites must be an array.",
     "emailsMustBeAnArray": "Email address invites must be an array.",
     "canOnlyInviteMaxInvites": "You can only invite \"<%= maxInvites %>\" at a time",
-    "onlyCreatorOrAdminCanDeleteChat": "Not authorized to delete this message!"
+    "onlyCreatorOrAdminCanDeleteChat": "Not authorized to delete this message!",
+    "onlyGroupLeaderEditTasks": "Not authorized to manage tasks"
 }
--- a/common/locales/en/groups.json
+++ b/common/locales/en/groups.json
@@ -193,5 +193,5 @@
     "emailsMustBeAnArray": "Email address invites must be an array.",
     "canOnlyInviteMaxInvites": "You can only invite \"<%= maxInvites %>\" at a time",
     "onlyCreatorOrAdminCanDeleteChat": "Not authorized to delete this message!",
-    "onlyGroupLeaderEditTasks": "Not authorized to manage tasks"
+    "onlyGroupLeaderCanEditTasks": "Not authorized to manage tasks!"
 }
--- a/test/api/v3/integration/tasks/groups/GET-tasks_group_id.test.js
+++ b/test/api/v3/integration/tasks/groups/GET-tasks_group_id.test.js
@@ -1,0 +1,83 @@
+import {
+  generateUser,
+  generateGroup,
+  translate as t,
+} from '../../../../../helpers/api-integration/v3';
+import { v4 as generateUUID } from 'uuid';
+import { each } from 'lodash';
+
+describe('GET /tasks/group/:groupId', () => {
+  let user;
+  let group;
+  let task;
+  let tasks = [];
+  let groupWithTask;
+  let tasksToTest = {
+    habit: {
+      text: 'test habit',
+      type: 'habit',
+      up: false,
+      down: true,
+    },
+    todo: {
+      text: 'test todo',
+      type: 'todo',
+    },
+    daily: {
+      text: 'test daily',
+      type: 'daily',
+      frequency: 'daily',
+      everyX: 5,
+      startDate: new Date(),
+    },
+    reward: {
+      text: 'test reward',
+      type: 'reward',
+    },
+  };
+
+  before(async () => {
+    user = await generateUser();
+    group = await generateGroup(user);
+  });
+
+  it('returns error when group is not found', async () => {
+    let dummyId = generateUUID();
+
+    await expect(user.get(`/tasks/group/${dummyId}`)).to.eventually.be.rejected.and.eql({
+      code: 404,
+      error: 'NotFound',
+      message: t('groupNotFound'),
+    });
+  });
+
+  each(tasksToTest, (taskValue, taskType) => {
+    context(`${taskType}`, () => {
+      before(async () => {
+        task = await user.post(`/tasks/group/${group._id}`, taskValue);
+        tasks.push(task);
+        groupWithTask = await user.get(`/groups/${group._id}`);
+      });
+
+      it('gets group tasks', async () => {
+        let getTask = await user.get(`/tasks/group/${groupWithTask._id}`);
+        expect(getTask).to.eql(tasks);
+      });
+
+      it('gets group tasks filtered by type', async () => {
+        let groupTasks = await user.get(`/tasks/group/${groupWithTask._id}?type=${task.type}s`);
+        expect(groupTasks).to.eql([task]);
+      });
+
+      it('cannot get a task owned by someone else', async () => {
+        let anotherUser = await generateUser();
+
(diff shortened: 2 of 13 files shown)


website/server/controllers/api-v3/tasks.js
```diff
@@ -55,7 +57,7 @@ async function _createTasks (req, res, user, challenge) {
     if (validationErrors) throw validationErrors;
 
     // Otherwise update the user/challenge
-    (challenge || user).tasksOrder[`${taskType}s`].unshift(newTask._id);
+    if (!group) (challenge || user).tasksOrder[`${taskType}s`].unshift(newTask._id);
```
paglias: groups are not going to have an order?

website/server/controllers/api-v3/tasks.js
```diff
@@ -134,9 +136,53 @@ api.createChallengeTasks = {
...
+ * @apiParam {UUID} groupId The id of the group the new task(s) will belong to
+ *
+ * @apiSuccess data An object if a single task was created, otherwise an array of tasks
+ */
+api.createGroupTasks = {
+  method: 'POST',
+  url: '/tasks/group/:groupId',
+  middlewares: [authWithHeaders()],
+  async handler (req, res) {
+    // req.checkParams('groupId', res.t('groupIdRequired')).notEmpty().isUUID();
```
paglias: why commented out?

paglias:
> We may need to add a property to groups to verify that the group has been paid for (or maybe a subscription model like users)

Absolutely

> I think we should also separate the task controller to separate files (group, challenge, user)

You mean having a controller for user tasks, challenge tasks and group tasks? If possible I'd like to be able to share most of the code between these 3 types of tasks

website/server/controllers/api-v3/tasks.js
```diff
@@ -33,7 +33,7 @@ async function _validateTaskAlias (tasks, res) {
 }
 
 // challenge must be passed only when a challenge task is being created
-async function _createTasks (req, res, user, challenge) {
+async function _createTasks (req, res, user, challenge, group) {
```
crookedneighbor: This should be refactored to:

``` js
async function _createTasks (req, res, options = {}) {
let {
  user,
  challenge,
  group,
} = options;
```

website/server/controllers/api-v3/tasks.js
```diff
@@ -134,9 +136,53 @@ api.createChallengeTasks = {
...
+
+    return null;
+  },
+};
+
 // challenge must be passed only when a challenge task is being created
-async function _getTasks (req, res, user, challenge) {
-  let query = challenge ? {'challenge.id': challenge.id, userId: {$exists: false}} : {userId: user._id};
+// @TODO: Factory - GetTasksQuery(type) - where type is user, challenge, group
+async function _getTasks (req, res, user, challenge, group) {
```
crookedneighbor: This should be refactored to:

``` js
async function _getTasks (req, res, options = {}) {
let {
  user,
  challenge,
  group,
} = options;
```

common/locales/en/groups.json
```diff
@@ -189,5 +189,6 @@
     "uuidsMustBeAnArray": "User ID invites must be an array.",
     "emailsMustBeAnArray": "Email address invites must be an array.",
     "canOnlyInviteMaxInvites": "You can only invite \"<%= maxInvites %>\" at a time",
-    "onlyCreatorOrAdminCanDeleteChat": "Not authorized to delete this message!"
+    "onlyCreatorOrAdminCanDeleteChat": "Not authorized to delete this message!",
+    "onlyGroupLeaderEditTasks": "Not authorized to manage tasks"
```
khipkin: onlyGroupLeaderCanEditTasks

khipkin: Also, add an `!` at the end.

test/api/v3/integration/tasks/groups/GET-tasks_group_id.test.js
```diff
@@ -0,0 +1,83 @@
+import {
+  generateUser,
+  generateGroup,
+  translate as t,
+} from '../../../../../helpers/api-integration/v3';
+import { v4 as generateUUID } from 'uuid';
+import { each } from 'lodash';
+
+describe('GET /tasks/group/:groupId', () => {
+  let user;
```
khipkin: `let user, group, task, groupWithTask;`

crookedneighbor: :+1: to that suggestion.

test/api/v3/integration/tasks/groups/POST-tasks_group_id.test.js
```diff
@@ -0,0 +1,119 @@
+import {
+  generateUser,
+  generateGroup,
+  translate as t,
+} from '../../../../../helpers/api-v3-integration.helper';
+
+describe('POST /tasks/group/:groupid', () => {
+  let user;
```
khipkin: `let user, guild;`

website/server/controllers/api-v3/tasks.js
```diff
@@ -9,71 +9,17 @@ import {
   NotAuthorized,
   BadRequest,
 } from '../../libs/api-v3/errors';
+import {
+  _createTasks,
```
crookedneighbor: If these are being exported to a lib, they should not be designated as private methods.

website/server/libs/api-v3/taskManager.js
```diff
@@ -0,0 +1,140 @@
+import * as Tasks from '../../models/task';
+import {
+  NotFound,
+  NotAuthorized,
+  BadRequest,
+} from '../api-v3/errors';
+import Bluebird from 'bluebird';
+import _ from 'lodash';
+
+export async function _validateTaskAlias (tasks, res) {
```
crookedneighbor: Since these are no longer private methods, they need unit test coverage.

TheHollidayInn: Do they need that in this PR? Just after a move?

TheHollidayInn:
Updated!

paglias:
tasksOrder should be added here https://github.com/TheHollidayInn/habitrpg/blob/40419ac29dbbc14bf30a02e4bde24920e8aaf742/website/server/models/group.js#L94

and group here https://github.com/TheHollidayInn/habitrpg/blob/40419ac29dbbc14bf30a02e4bde24920e8aaf742/website/server/models/task.js#L83

TheHollidayInn:
Updated!

website/server/libs/taskManager.js
```diff
@@ -0,0 +1,158 @@
...
+  });
+}
+
+
+/**
+ * Creates tasks for a user, challenge or group.
+ *
+ * @param  req  The Express req variable
+ * @param  res  The Express res variable
+ * @param  options  Should be a user, user and challenge to create a challenge task or user and group to create a group task
```
paglias, line 30: commented on https://github.com/HabitRPG/habitrpg/pull/7887/files#r75236820

website/server/libs/taskManager.js
```diff
@@ -0,0 +1,158 @@
...
+  });
+
+  // tasks with aliases need to be validated asyncronously
+  await _validateTaskAlias(toSave, res);
+
+  toSave = toSave.map(task => task.save({ // If all tasks are valid (this is why it's not in the previous .map()), save everything, withough running validation again
+    validateBeforeSave: false,
+  }));
+
+  toSave.unshift((challenge || user).save());
```
paglias, line 75: https://github.com/HabitRPG/habitrpg/pull/7887/files#r75349402

PR #7787, opened 2016-07-15
Fixing Duplicate tasks showing up after joining a challenge

### LINKED ISSUE #7730: challenges being joined twice simultaneously; challenge tasks duplicated
I've seen some reports of players having duplicate challenge tasks and I've also seen challenge IDs appearing twice in users' `challenge` objects. I'm going to start reporting all instances here so that we can track it and work out what's causing it.

This is distressing for the players because it makes their task lists feel out of control, and it can be damaging for them and their party if they think they're not allowed to tick a duplicated Daily.

@paglias @crookedneighbor  Do you have any idea what might be causing this?

Joining challenges doesn't seem to be recorded in Amplitude. Is it worth adding it?

Natusha (0f3823fd-38a4-477a-ad4a-6d0ce458ef6f) in the Report a Bug guild
**When I click in the "Join" on the challenge area, all the quests (daily, to do, etc) appear duplicated, that happened in the 2 challenges I took, and I'm scared of taking more and continuing duplicating things**

```
    "challenges": [
        "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e",
        "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e",
        "2650b24f-7097-4789-adcd-dad9cae5d74b",
        "2650b24f-7097-4789-adcd-dad9cae5d74b",
        "c4004b63-e868-4e6e-a7e4-3cd30d02a3e8"
    ],
```

For challenge [63a21fcf-7ab8-4483-a70d-d4fb61d88f6e](https://habitica.com/#/options/groups/challenges/63a21fcf-7ab8-4483-a70d-d4fb61d88f6e):

```
{
    "_id": "4bb04ea8-2332-4d4f-9aa6-1d60fe6c7315",
    "tags": [],
    "challenge": {
        "taskId": "576cb003-394e-49ad-ae23-e35fe28fb320",
        "id": "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e"
    },
    "createdAt": "2016-06-26T18:20:13.464Z"
}

{
    "_id": "5b25e9bf-55c7-4d1d-bf6e-3886fa61920d",
    "tags": [
        "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e"
    ],
    "challenge": {
        "taskId": "576cb003-394e-49ad-ae23-e35fe28fb320",
        "id": "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e"
    },
    "createdAt": "2016-06-26T18:20:13.491Z"
}

{
    "_id": "513c4947-99d5-4433-8553-9eb4b8d71ad3",
    "tags": [
        "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e"
    ],
    "challenge": {
        "taskId": "69acb6fe-2d2d-4a96-b5d3-847ff17b0e8e",
        "id": "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e"
    },
    "createdAt": "2016-06-18T15:12:21.551Z"
}

{
    "_id": "e58fef95-5dfa-4840-9f28-a6528436cb1a",
    "tags": [
        "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e"
    ],
    "challenge": {
        "taskId": "2231e1e5-ed97-429f-a96e-2e400bd89e4d",
        "id": "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e"
    },
    "createdAt": "2016-06-26T18:20:13.493Z"
}

{
    "_id": "e7eacac8-422e-471c-a8cc-4e6d288cd432",
    "tags": [
        "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e"
    ],
    "challenge": {
        "taskId": "2231e1e5-ed97-429f-a96e-2e400bd89e4d",
        "id": "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e"
    },
    "createdAt": "2016-06-26T18:20:13.465Z"
}

{
    "_id": "e35f5786-4c30-4947-90ba-79597f2af4da",
    "tags": [
        "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e"
    ],
    "challenge": {
        "taskId": "ae9a0b12-810e-4364-8bc5-78aecfa1ce9d",
        "id": "63a21fcf-7ab8-4483-a70d-d4fb61d88f6e"
    },
    "createdAt": "2016-06-18T15:12:21.553Z"
}
```

For challenge [2650b24f-7097-4789-adcd-dad9cae5d74b](https://habitica.com/#/options/groups/challenges/2650b24f-7097-4789-adcd-dad9cae5d74b):

```
{
    "_id": "abc8d9dd-5bae-4f9e-89e4-77287c566a3a",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "a93cd6ef-811d-4c11-9458-ee8341fefff3",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.113Z"
}

{
    "_id": "64fe1e81-1ad6-4178-87ef-1188ca0666a2",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "a93cd6ef-811d-4c11-9458-ee8341fefff3",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.111Z"
}


{
    "_id": "1de2604a-3bb0-459b-a89d-33f823d23258",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "1b6fe224-d20b-41f8-a745-70ae3af8f0c1",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.106Z"
}

{
    "_id": "d7023082-bae6-416b-872b-ac6a07274cd1",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "1b6fe224-d20b-41f8-a745-70ae3af8f0c1",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.101Z"
}


{
    "_id": "16a27440-8910-4da8-bdcb-810dc2f5f41f",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "2bbf82a9-d080-46ff-8b1a-e16dba03d355",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.105Z"
}

{
    "_id": "1251dc3c-fa27-4a7d-940e-a8b44821a8e0",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "2bbf82a9-d080-46ff-8b1a-e16dba03d355",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.099Z"
}


{
    "_id": "fd1e9fb8-5f27-4560-9d5a-74adf9d8a409",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "d83864fd-e442-4a36-86fa-b67b7253cd99",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.103Z"
}

{
    "_id": "75f339ec-13d5-4a93-a8ee-c923c3c76ec5",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "d83864fd-e442-4a36-86fa-b67b7253cd99",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.107Z"
}


{
    "_id": "3476d193-d141-4ef5-ba7f-eb2074ffca1b",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "b4fda3e5-b071-4ce4-ac35-8504b0d9c238",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.108Z"
}

{
    "_id": "b06b8e2c-2bcf-4711-9963-c5f14df29621",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "b4fda3e5-b071-4ce4-ac35-8504b0d9c238",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.104Z"
}



{
    "_id": "8bc52e97-9708-402c-9aeb-0a9aeca6d198",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "c4628b97-6784-48fb-a263-8edcd79d63a4",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.105Z"
}

{
    "_id": "d20e577e-785d-497a-ae1f-4f02893d1653",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "c4628b97-6784-48fb-a263-8edcd79d63a4",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.109Z"
}


{
    "_id": "f6a639a0-ef41-4e9d-83c9-482ae9e047a8",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "07f03a8f-e420-4e00-bfaf-1667d45ff135",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.110Z"
}

{
    "_id": "dfbbeb23-cd78-4340-89ea-c78c76688ef7",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "07f03a8f-e420-4e00-bfaf-1667d45ff135",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.106Z"
}


{
    "_id": "e7212581-c400-4978-9622-7e4fc238b1e1",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "221d286c-f496-4ef0-8cf4-0de69d340dac",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.111Z"
}

{
    "_id": "5f90d4bc-a6c3-421e-aa9b-c5d04e7de02d",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "221d286c-f496-4ef0-8cf4-0de69d340dac",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.108Z"
}


{
    "_id": "d611e017-e291-4253-9aba-b6e463a8b1b6",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "3d23f8bf-df9d-4feb-a0fb-024a4ab9b88a",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.112Z"
}

{
    "_id": "0e42003b-a6fd-432b-a809-98f062ffad83",
    "tags": [
        "2650b24f-7097-4789-adcd-dad9cae5d74b"
    ],
    "challenge": {
        "taskId": "3d23f8bf-df9d-4feb-a0fb-024a4ab9b88a",
        "id": "2650b24f-7097-4789-adcd-dad9cae5d74b"
    },
    "createdAt": "2016-06-26T20:49:22.109Z"
}
```

Hus274:
Fixes #7730 
### Changes

Double clicking the join button on a challenge will no longer result in the user showing duplicate tasks.

Switched from mongoose's save method to an update method which should prevent the save from using a diff which was the cause of the new item being saved twice.

---

closes #7730

### DIFF
--- a/website/server/models/challenge.js
+++ b/website/server/models/challenge.js
@@ -79,8 +79,11 @@ schema.methods.syncToUser = async function syncChallengeToUser (user) {
   challenge.shortName = challenge.shortName || challenge.name;
 
   // Add challenge to user.challenges
-  if (!_.contains(user.challenges, challenge._id)) user.challenges.push(challenge._id);
-
+  if (!_.contains(user.challenges, challenge._id)) {
+    // using concat because mongoose's protection against concurrent array modification isn't working as expected.
+    // see https://github.com/HabitRPG/habitrpg/pull/7787#issuecomment-232972394
+    user.challenges = user.challenges.concat([challenge._id]);
+  }
   // Sync tags
   let userTags = user.tags;
   let i = _.findIndex(userTags, {id: challenge._id});
--- a/website/client-old/js/controllers/challengesCtrl.js
+++ b/website/client-old/js/controllers/challengesCtrl.js
@@ -320,7 +320,7 @@ habitrpg.controller("ChallengesCtrl", ['$rootScope','$scope', 'Shared', 'User', 
     */
 
     $scope.join = function (challenge) {
-      challenge._joiningInProgress = false;
+      challenge._joiningInProgress = true;
       Challenges.joinChallenge(challenge._id)
         .then(function (response) {
           User.user.challenges.push(challenge._id);
@@ -330,8 +330,8 @@ habitrpg.controller("ChallengesCtrl", ['$rootScope','$scope', 'Shared', 'User', 
         .then(function (response) {
           var tasks = response.data.data;
           User.syncUserTasks(tasks);
+          challenge._joiningInProgress = false;
         });
-      challenge._joiningInProgress = true;
     }
 
     $scope.leave = function(keep, challenge) {
--- a/website/client-old/js/controllers/challengesCtrl.js
+++ b/website/client-old/js/controllers/challengesCtrl.js
@@ -320,6 +320,7 @@ habitrpg.controller("ChallengesCtrl", ['$rootScope','$scope', 'Shared', 'User', 
     */
 
     $scope.join = function (challenge) {
+      challenge._joiningInProgress = false;
       Challenges.joinChallenge(challenge._id)
         .then(function (response) {
           User.user.challenges.push(challenge._id);
@@ -330,6 +331,7 @@ habitrpg.controller("ChallengesCtrl", ['$rootScope','$scope', 'Shared', 'User', 
           var tasks = response.data.data;
           User.syncUserTasks(tasks);
         });
+      challenge._joiningInProgress = true;
     }
 
     $scope.leave = function(keep, challenge) {
--- a/website/views/options/social/challenges.jade
+++ b/website/views/options/social/challenges.jade
@@ -202,7 +202,7 @@ script(type='text/ng-template', id='partials/options.social.challenges.html')
                   a.btn.btn-sm.btn-danger(ng-show='isUserMemberOf(challenge)', ng-click='clickLeave(challenge, $event)')
                     span.glyphicon.glyphicon-ban-circle
                     =env.t('leave')
-                  a.btn.btn-sm.btn-success(ng-hide='isUserMemberOf(challenge)', ng-click='join(challenge)')
+                  a.btn.btn-sm.btn-success(ng-hide='isUserMemberOf(challenge)', ng-click='join(challenge)', ng-disabled='challenge._joiningInProgress')
                     span.glyphicon.glyphicon-ok
                     =env.t('join')
               a.accordion-toggle(id="{{challenge._id}}" ng-click='toggle(challenge._id)')


website/server/models/challenge.js
```diff
@@ -149,7 +154,7 @@ schema.methods.syncToUser = async function syncChallengeToUser (user) {
     }
   });
 
-  toSave.push(user.save());
+  toSave.push(User.collection.update({_id: user._id}, user.toObject()));
```
crookedneighbor: This is a lot of data to update on the user object. Why not just the things that are being added to it?

website/server/models/challenge.js
```diff
@@ -149,7 +154,7 @@ schema.methods.syncToUser = async function syncChallengeToUser (user) {
...
   return Bluebird.all(toSave);
```
Hus274, line 148: That was what I had done initially, but then I had three statements for the user modifications in this file and then found that there are modifications in the server library that need to be saved to the user object as well so I thought this would be cleaner. 

Would you rather have the update statements for the improved performance?

Alys, line 148: We do have problems periodically with multiple updates that are supposed to happen at the same time but only some of them are executed or saved, leaving the user's data in an inconsistent state. From that point of view, it would be better to have one update instead of three.

website/server/models/challenge.js
paglias: Also, this should not be `User.update({_id: user._id}, {$set: attributesToUpdate})`?

paglias:
@Hus274 have you tested that this works?

This problem is happening because when a user clicks very fast on the join button 2 or more times the server will execute the requests concurrently so the check for `user is in challenge?` returns `false` in two or more occasions causing the code to add the challenge and its tasks to the user to be executed multiple times.

I'm not sure that simply using `update` instead of `save` is going to make a difference since `save` under the hood simply executes an `update` command

website/server/models/challenge.js
paglias: Another thing: `update` doesn't return a promise unless you call `.exec()` at the end http://mongoosejs.com/docs/promises.html

Hus274:
User.collections.save from my understanding will actually use the mongo implementation of the update method which will overwrite the existing document from the one from memory. This will cause the updates to act independently and allow one to overwrite the other instead of combining the results.

So in this case if the button is clicked very fast twice, it will simply call the method twice and get the result of whichever executes last as the user state in the database which will only have the challenge added once. I was able to confirm that this was the behavior on my developer instance. Thinking about it more this morning, I should be able to write a test around this two by just making two async calls in parallel and confirming the data isn't duplicated.

On the other hand, I was not able to replicate the duplicate tasks issue, just the challenge appearing twice in the user object. Do you have any idea on how I could reproduce this other one?

paglias:
@Hus274 unfortunately that's not how it works. `Model.collection.update` and `Model.update` execute exactly the same command which is a mongodb's `update`.

What's happening here is that MongoDB doesn't have transactions which allow multiple queries to be executed in an atomic way making sure that the data being updated is exactly the same that the application read a few milliseconds before.

So what happens is that parallel requests will read that user is not in a challenge and they'll all try to add the user to the challenge.

The way mongoose works is that it proxies array methods (push, pop, ...) and instead of telling mongoose to replace the entire array it used mongodb's atomic operations to change them. So for example two parallel requests will both send the `$push` op to mongodb and two challenges will be added to the `user.challenges` array.

The problem is that mongoose has built-in protection for this sort of issues: basically it has a field `__v` used as a counter that's increased whenever an array is modified and used to issue the update query when saving. 

So what should happen in theory is that the 2 parallel queries will read `__v = 2` and send these two update commands:

(the syntax may be wrong)

```
{_id: user._id, __v:2}, {$inc: {__v: 1}, $push: {challenges: challenge._id}}}
{_id: user._id, __v:2}, {$inc: {__v: 1}, $push: {challenges: challenge._id}}}
```

Since updates on a single doc are atomic one of the two updates will happen before the other setting __v to 3 and thus should cause the second query to fail because a doc that satisfy `{_id: user._id, __v:2}` doesn't exists anymore.

More info can be found here:
http://aaronheckmann.tumblr.com/post/48943525537/mongoose-v3-part-1-versioning

One thing that we could do and is similar to @Hus274 's fix without having to update the entire doc  is to replace the array instead of adding an element to it:

```
user.challenges.push(challengeId) -> user.challenge = user.challenges.concat([challengeId])
```

But going ahead and doing this I'd like to hear from @crookedneighbor

Hus274:
Ahh ok, that makes sense to me. Thanks for explaining and drawing it out @paglias, it was really helpful!

paglias:
It's possible this is a bug in mongoose https://github.com/Automattic/mongoose/pull/1942 but I'm not sure if it's the same thing we're experiencing

website/server/models/challenge.js
crookedneighbor: @Hus274 Are you going to make this change?

Hus274: From what @paglias said below he didn't sound convinced that this would actually fix the problem so I didn't know if I should keep going. I can make this change and update it if that's the way we want to go

paglias: I think the best thing to do is to use `concat` instead of `push` on the challenges array as outlined here https://github.com/HabitRPG/habitrpg/pull/7787#issuecomment-232972394 at the end.

I would like to point out, though, that mongoose should protect against this type of issues and I'm not sure why we have this issues.

So while the change is okay I'd add a comment above it to explain while we're doing it (that mongoose's protection against concurrent updates of arrays using the __v field isn't working as expected)

Hus274:
Switched to using concat like paglias suggested.

website/server/models/challenge.js
```diff
@@ -79,7 +79,10 @@ schema.methods.syncToUser = async function syncChallengeToUser (user) {
   challenge.shortName = challenge.shortName || challenge.name;
 
   // Add challenge to user.challenges
-  if (!_.contains(user.challenges, challenge._id)) user.challenges.push(challenge._id);
+  if (!_.contains(user.challenges, challenge._id)) {
+    // using concat because mongoose's protection against concurrent array modification isn't working as expected. see #7787
```
crookedneighbor: Let's link to the specific comment.

crookedneighbor:
I think other than my one comment. It's looking good. Is there a test we can add that reproduces the bug and shows that it's now fixed?

Hus274:
I linked the specific comment in the code and repushed. I tried to replicate the bug in a test but wasn't able to do it by calling sync to user multiple times and then awaiting all of them. Do you have another idea on how to replicate the bug?

crookedneighbor:
Do we do any kind of client side disabling of the button once it is clicked once? That should be done in addition to this adjustment to the API.

Hus274:
Added client side disabling of the button after the first click so the user can't hit the server twice for the same request as requested. This should be good for review again, let me know if you would like any other changes made.

TheHollidayInn:
Ah, just saw you had trouble replicating. The client side should solve most of it for now.

Hus274:
@TheHollidayInn I was able to replicate it but it just took some time, it was a matter of double clicking as fast as I could until I finally got it. After the fix, I was not ever able to make the issue occur. 

Do you want me to update this pr to only include the client side changes?

TheHollidayInn:
No, you can keep the server side changes. I'm confused about the client logic though. 

The placement of `challenge._joiningInProgress = true` doesn't make sense and seems to run in parallel to the `challenge._joiningInProgress = false`. I'm sure you don't notice and issue because once the user joins, the button is hidden or there is a refresh. 

But, when the join function is called, you should immediately set `challenge._joiningInProgress = true` not to false. Then inside of the lass `then` callback, you should set `challenge._joiningInProgress = false`. Because the promises are now over.

Hus274:
Yeah that's a good catch, not sure what happened there. I'll update it.

Hus274:
Updated to put the variable setting in the last promise and switched the assignments.

TheHollidayInn:
Added to your Tier progress. Thanks for the PR!

PR #7797, opened 2016-07-16
Send a notification to a user when their group invite is accepted

### LINKED ISSUE #7788: When an invitation is accepted, the inviter should be notified.
When a Party or Guild invitation is accepted, the person who sent the invitation should receive a notification that it was accepted.

Sample text:

[NAME OF USER] accepted your invitation to [NAME OF GUILD/PARTY]!

Hus274:
Fixes #7788 
### Changes

This change sends notifications to users when another user has accepted their invite to join a party.

acceptednotification

---


common/script/ops/clearInviteAcceptedNotification.js
```diff
@@ -0,0 +1,10 @@
+import _ from 'lodash';
+
```
khipkin: delete extra nl

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
```diff
@@ -134,6 +134,14 @@ describe('POST /group/:groupId/join', () => {
 
         await expect(user.get('/user')).to.eventually.have.deep.property('items.quests.basilist', 1);
       });
+
+      it('pushes the accepted invitation to invitations.accepted of the inviting user', async () => {
+        await invitedUser.post(`/groups/${guild._id}/join`);
+
+        await user.sync();
+
+        expect(user.invitations.accepted).to.include({id: guild._id, groupName: guild.name, invitedUsername: invitedUser.profile.name, type: 'guild'});
```
khipkin: break this up to reduce line length

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
```diff
@@ -232,6 +240,14 @@ describe('POST /group/:groupId/join', () => {
         expect(invitedUser).to.have.deep.property('party.quest.key', party.quest.key);
         expect(party.quest.members[invitedUser._id]).to.be.null;
       });
+
+      it('pushes the accepted invitation to invitations.accepted of the inviting user', async () => {
+        await invitedUser.post(`/groups/${party._id}/join`);
+
+        await user.sync();
+
+        expect(user.invitations.accepted).to.include({id: party._id, groupName: party.name, invitedUsername: invitedUser.profile.name, type: 'party'});
```
khipkin: also break this up to reduce line length

test/api/v3/integration/user/POST-user_clearInviteAcceptedNotification.js
```diff
@@ -0,0 +1,23 @@
+import {
+  generateUser,
+} from '../../../../helpers/api-integration/v3';
+import { v4 as generateUUID } from 'uuid';
+
+describe('POST /user/clerInviteAcceptedNotification', () => {
+  let user;
+  let groupId = generateUUID();
+  let acceptedInvitationNotification = {id: groupId, groupName: 'group', invitedUsername: 'test', type: 'party'};
```
khipkin: line length

test/api/v3/integration/user/POST-user_clearInviteAcceptedNotification.js
```diff
@@ -0,0 +1,23 @@
+import {
+  generateUser,
+} from '../../../../helpers/api-integration/v3';
+import { v4 as generateUUID } from 'uuid';
+
+describe('POST /user/clerInviteAcceptedNotification', () => {
```
khipkin: clear

test/common/ops/clearInviteAcceptedNotification.js
```diff
@@ -0,0 +1,23 @@
+import clearInviteAcceptedNotification from '../../../common/script/ops/clearInviteAcceptedNotification';
+import {
+  generateUser,
+} from '../../helpers/common.helper';
+import { v4 as generateUUID } from 'uuid';
+
+describe('shared.ops.clearInviteAcceptedNotification', () => {
+  let user;
+  let groupId = generateUUID();
+  let acceptedInvitationNotification = {id: groupId, groupName: 'group', invitedUsername: 'test', type: 'party'};
```
khipkin, line 10: line length

test/spec/controllers/menuCtrlSpec.js
```diff
@@ -8,12 +8,13 @@ describe('Menu Controller', function() {
...
         $provide.value('Chat', { seenMessage: function() {} });
+        $provide.value('User', { clearInviteAcceptedNotification: function() {} });
       });
 
-      inject(function(_$httpBackend_, $rootScope, $controller) {
+      inject(function(_$httpBackend_, $rootScope, $controller, User) {
         scope = $rootScope.$new();
 
-        ctrl = $controller('MenuCtrl', {$scope: scope, $window: $window, User: user});
+        ctrl = $controller('MenuCtrl', {$scope: scope, $window: $window});
```
khipkin, line 17: Why remove `User` here?

website/client/js/controllers/menuCtrl.js
```diff
@@ -13,7 +13,8 @@ angular.module('habitrpg')
         var user = $scope.user;
         if (user.purchased && user.purchased.plan && user.purchased.plan.mysteryItems && user.purchased.plan.mysteryItems.length) {
           return mysteryValue;
-        } else if ((user.invitations.party && user.invitations.party.id) || (user.invitations.guilds && user.invitations.guilds.length > 0)) {
+        } else if ((user.invitations.party && user.invitations.party.id) || (user.invitations.guilds && user.invitations.guilds.length > 0)
```
khipkin, line 16: inconsistent formatting of `if` statement; also, line length.

website/server/controllers/api-v3/user.js
```diff
@@ -1326,4 +1326,33 @@ api.setCustomDayStart = {
...
+*
+* @apiParam {String} invitedUsername The username of the user that accepted the invitation
+* @apiParam {UUID} groupId The id of the group that was joined
+*
+* @apiSuccess {Object} data The updated user.invitations.accepted array
+*/
+api.clearInviteAcceptedNotification = {
+  method: 'POST',
+  middlewares: [authWithHeaders()],
+  url: '/user/clearInviteAcceptedNotification/:invitedUsername/:groupId',
```
khipkin, line 1337: this url does not match the one above

website/server/controllers/api-v3/user.js
```diff
@@ -1326,4 +1326,33 @@ api.setCustomDayStart = {
...
+  async handler (req, res) {
+    let user = res.locals.user;
+    let invitedUsername = req.params.invitedUsername;
+    let groupId = req.params.groupId;
+
+    _.remove(user.invitations.accepted, {invitedUsername, id: groupId});
+
+    await User.update(
+      {_id: user._id},
+      {$set: {'invitations.accepted': user.invitations.accepted}}).exec();
```
khipkin: move `).exec();` to next line

test/spec/controllers/menuCtrlSpec.js
Hus274, line 17: Because the user is now being injected above.

test/common/ops/clearInviteAcceptedNotification.js
paglias, line 10: Line length is fine

website/client/js/controllers/menuCtrl.js
paglias, line 16: line was longer before this change, it's okay and we're not enforcing a strong code style on client side code

khipkin: Sorry, just trying to help.

website/server/controllers/api-v3/user.js
```diff
@@ -1326,4 +1326,33 @@ api.setCustomDayStart = {
   },
 };
 
+/**
+* @api {post} /user/notifications/clear/:invitedUsername/:groupId Clear a notification of an accepted invite
```
Alys: Our convention for these `@api` lines is that the path starts with `/api/v3/`  (see the previous route in this file, for example). Also, @khipkin's comment below is correct - there's a mismatch between this comment and the `url` in the code.

Hus274: Sorry, when these comments were made I was waiting for Blade to look through it as well since it looked like all the comments made above were minor spelling or wrong comment fixes but I should've said something here. I'm working on this now and will resubmit again once I've fixed this.

Hus274:
I've fixed the things that were commented on here. Should be ready to be reviewed now!

paglias:
So, while this PR was open we added a new way to store notifications for the user in user.notifications, that is more general and and doesn't require to add a field for every new notifications. I think we should use it but I'm sorry for all the work you've already done :( @crookedneighbor what do you think?

crookedneighbor:
I agree with @paglias

Hus274:
Ok, I'll update this to use the new model.

crookedneighbor:
@Hus274 What's the status on this?

Hus274:
In progress. Going to try and get it finished up this week.

Hus274:
I don't see user.notifications being used anywhere. Is there any particular format or style or ideas on how this is intended to be used in the future? Or is there somewhere that this is being used that I just haven't found yet?

paglias:
@Hus274 there's a method on the user model called `addNotification`, you have to add the notification to the list in models/userNotification and then you use addNotification like this `user.addNotification(NOTIF_NAME, data)`.

If you search for `addNotification(` in the code it's used in several spots

Hus274:
So it looks like the current usage of user.notifications is used to bring up a popup that displays the notification which is different than the initial way I did this with a notification in the messages box in the upper right. Should this notification be a pop up as well?

crookedneighbor:
I think a popup makes sense for now.

paglias:
Most notifications are using modals but that's only because they substituted modals, you can display a message if you want. I think a message in the top right could be more appropriate than a modal for this use case

Hus274:
@paglias The problem I had trying to do that was the the user.notifications object in header.jade was always an empty array even when it wasn't in the database.

paglias:
@Hus274 I was thinking more about a one time blue message in the top right that you can trigger from the Notification service. 

Otherwise a popup is fine too. The problem you're seeing is that user.notifications is cleared as soon as a notification is received and so it disappear immediately (see https://github.com/HabitRPG/habitrpg/blob/develop/website/client-old/js/controllers/notificationCtrl.js#L134). We'll want to change it but not before the redesign.

Hus274:
This one still needs to be refactored after the code was move and sounds like the implementation has changed based on how it sounds we want to handle this. I'm going to go ahead and close this for now and likely re-open a new one if I can work on this to minimize the diff issues.

PR #7800, opened 2016-07-17
Fixing Exponential Quest Reward Scrolls

### LINKED ISSUE #2715: Exponential Quest Reward Scrolls
So... as far as I can tell, if the second scroll is a reward of the first, everyone's going to get it. So a party of 7 finishes Vice pt 1. They get 7 scrolls of Vice pt 2. After that, they have 49 scrolls of Vice pt. 3.

And then another member hits level 30...

---

_edit by Alys:_

This issue applies to these four quest lines (which contain three quests each; the first only is linked to below):
- [Vice](http://habitica.wikia.com/wiki/Free_Yourself_of_the_Dragon's_Influence) - `vice1`, `vice2`, `vice3`
- [Attack of the Mundane](http://habitica.wikia.com/wiki/Dish_Disaster) - `atom1`, etc
- [Recidivate](http://habitica.wikia.com/wiki/The_Moonstone_Chain) - `moonstone1`, etc
- [The Golden Knight](http://habitica.wikia.com/wiki/A_Stern_Talking-To) - `goldenknight1`, etc

The way it is supposed to work is that when the first and second quest finish, the scroll for the next quest is given to only the person who had started the quest. All other quest rewards are given to all participants, as normal.

There's some old code in these two PRs, but they developed significant merge conflicts before they were reviewed, and I never found more time to finish it:
- https://github.com/HabitRPG/habitrpg-shared/pull/303
- https://github.com/HabitRPG/habitrpg/pull/4004

The code might be so outdated now that it could be better to start from scratch.

Here are screenshots of what we had in mind then. They were made a long time ago, so they might not exactly match the current site appearance - if so, the current site appearance should be retained. These are just a guideline for how this might be implemented now. If you want to handle it in some other way, that's fine!

**Quest Scroll when you click on it in Market or Inventory:**

1 - quest scroll when you click on it in market or inventory

**Quest Invitation:**

2 - quest invitation

**In-Progress Quest on Party page:**

3 - in-progress quest on party page

**Quest Completed pop-up:**

4 - quest completed pop-up
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1398569-exponential-vice?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Hus274:
Fixes #2715 
### Changes

Updating Recidivate, Golden Knight, Attack of the Mundane, and Vice quests to only reward the next quest scroll in the story line to the quest owner. This should remove the exponential awarding of quest scrolls in a party with multiple people.

Hover Over Quest in Market:
fullpicture

Quest Start:
queststart

Quest Completed:
questcompleted

---


common/script/content/index.js
```diff
@@ -1251,7 +1251,7 @@ api.quests = {
       str: 1.5
     },
     drop: {
-      items: [
+      itemsForOwner: [
```
paglias: why this name change? It's probably a breaking change that we cannot make without breaking mobile apps

Hus274: The idea was to split the awards of a quest into two different arrays, items for all participants and items for just the quest owner. It's not a rename so much as moving some of the rewards into a new array titled itemsForOwner. 

That being said, it probably will be a breaking change in terms of correctly displaying the quest object on the mobile app. The only other way I can think of distinguishing these is to update the item model and add another field to it that says who it applies to. So something like item.awardsOnlyQuestOwner. The reason I didn't do this in the first place was because it seemed to be less efficient as then getting the list of rewards for allParticipants and for just the quest owner would both require a mapping function on the full list to turn it into the two arrays.

If this would be a breaking change, it's probably worth sacrificing the efficiency in favor of not breaking the apps. Should I update this to the item property model?

paglias: Yeah having `item.onlyOwner` or something similar seems the best solution. These are small arrays and it's not going to be a perf problem having to filter them

Hus274: Makes sense to me! I'll update this and resubmit

Hus274:
Updated the model so this won't be a breaking change.

common/locales/en/quests.json
```diff
@@ -10,7 +10,13 @@
     "questDetails": "Quest Details",
     "invitations": "Invitations",
     "completed": "Completed!",
+    "rewardsAllParticipants": "Rewards for all Quest Participants",
+    "rewardsQuestOwner": "Additional Rewards for Quest Owner",
+    "questOwnerReceived": "The Quest Owner has also Received",
```
khipkin: You have this string twice.

Hus274: Good catch, I was doing some pretty interesting git stash stuff and I must have missed this on one of the pops. I'll fix this when the rest is reviewed.

website/views/shared/modals/quest-rewards.jade
```diff
@@ -1,8 +1,8 @@
 script(id='partials/options.social.party.quest-rewards.html', type='text/ng-template')
   hr
-  h5 {{::header}}
+  h5 {{::headerParticipant}}
   table.table.table-striped
-    tr(ng-repeat='drop in quest.drop.items')
+    tr(ng-repeat='_.reject(quest.drop.items, \'onlyOwner\')')
```
khipkin: no more `drop in`?

common/locales/en/quests.json
paglias: Is Received meant to start with an upper case letter?

paglias:
Looking good to me

common/locales/en/quests.json
Hus274: Originally they were all capitalized as I based them on the image in #2715. I can update this to be lowercase though

crookedneighbor: It's supposed to be capitalized as it is a title.

Hus274: Should it be "Quest Owner Has Also Received"?

Hus274:
Capitalization added to the title

Hus274:
Rebased on top of develop and the tests are passing again

test/api/v3/unit/models/group.test.js
```diff
@@ -929,9 +929,19 @@ describe('Group Model', () => {
```
crookedneighbor, line 1264: Let's specify in the it block that it awards quest scrolls to the owner.

website/client/js/services/questServices.js
```diff
@@ -68,9 +68,10 @@ angular.module('habitrpg')
         }
       }
       text += '---\n\n';
-      text += '**' + window.env.t('rewards') + ':**\n\n';
-      if(quest.drop.items) {
-        for (var item in quest.drop.items) {
+      text += '**' + window.env.t('rewardsAllParticipants') + ':**\n\n';
+      var participantRewards = quest.drop.items ? _.reject(quest.drop.items, 'onlyOwner') : quest.drop.items;
```
crookedneighbor: I think:

``` js
var participantRewards = quest.drop.items ? _.reject(quest.drop.items, 'onlyOwner') : null;
```

Would have a little more clarity here.

crookedneighbor: Also, I believe `_.reject` works with falsy values:

``` js
var participantRewards = _.reject(undefined, 'onlyOwner') // []
if (particpantRewards.length > 0) {
  // logic
}
```

Double check that, but it could make the logic more readable.

website/client/js/services/questServices.js
```diff
@@ -68,9 +68,10 @@ angular.module('habitrpg')
...
+      if(participantRewards) {
+        for (var item in participantRewards) {
```
crookedneighbor: Let's use `_.each` here. `for in` loops have some bad behavior when the Object.prototype is modified by a third party. (I know you didn't write the original `for in`, but we should still clean it up.

website/client/js/services/questServices.js
```diff
@@ -79,6 +80,14 @@ angular.module('habitrpg')
       if(quest.drop.gp)
         text += quest.drop.gp + ' ' + window.env.t('gold') + '\n\n';
 
+      var ownerRewards = quest.drop.items ? _.filter(quest.drop.items, 'onlyOwner') : quest.drop.items;
```
crookedneighbor: same comments as above

website/client-old/js/services/questServices.js
```diff
@@ -68,17 +68,26 @@ angular.module('habitrpg')
         }
       }
       text += '---\n\n';
-      text += '**' + window.env.t('rewards') + ':**\n\n';
-      if(quest.drop.items) {
-        for (var item in quest.drop.items) {
+      text += '**' + window.env.t('rewardsAllParticipants') + ':**\n\n';
+      var participantRewards = _.reject(quest.drop.items, 'onlyOwner');
+      if(participantRewards.length > 0) {
+        _.each( participantRewards, function(item) {
```
crookedneighbor: remove the space after `_.each(`, so that it is `_.each(participantRewards`

website/client-old/js/services/questServices.js
```diff
@@ -68,17 +68,26 @@ angular.module('habitrpg')
...
       }
       if(quest.drop.exp)
         text += quest.drop.exp + ' ' + window.env.t('experience') + '\n\n';
       if(quest.drop.gp)
         text += quest.drop.gp + ' ' + window.env.t('gold') + '\n\n';
 
+      var ownerRewards = _.filter(quest.drop.items, 'onlyOwner');
+      if(ownerRewards.length > 0) {
+        text += '**' + window.env.t('rewardsQuestOwner') + ':**\n\n';
+        _.each( ownerRewards, function(item){
```
crookedneighbor: Same here.

Hus274:
Finished up fixing this according to the comments here. Also tested the _.reject(undefined, ...) behavior to make sure that it worked as @crookedneighbor said and everything looks good!

website/server/models/group.js
```diff
@@ -507,38 +539,33 @@ schema.methods.finishQuest = async function finishQuest (quest) {
...
+  let questOwnerUpdates = {
+    $inc: {},
+    $set: {},
+  };
+  let questLeader = this.quest.leader;
+
+  _.each(_.filter(quest.drop.items, 'onlyOwner'), (item) => {
+    _.merge(questOwnerUpdates, _getUserUpdateForQuestReward(item, quest.drop.items));
   });
+  questOwnerUpdates = _.omit(questOwnerUpdates, _.isEmpty);
```
crookedneighbor: I feel like this should be done in the `_getUserUpdateForQuestReward`  function.

crookedneighbor (changes requested):

website/client-old/js/services/questServices.js
Hus274: done.

website/client-old/js/services/questServices.js
Hus274: done

website/server/models/group.js
Hus274: I'm not sure I understand. The method right now creates the update object that gets pushed to all users participating in the quest and then comes back and generates the update specifically for the owner of the quest.

What specifically did you want moved into the method above and what would the return value be?

Hus274 (commented):
Addressed comments and left a question on one

Hus274:
Re-added the missing only owner field and fixed console errors.

website/server/models/group.js
crookedneighbor: Maybe I don't understand what this is doing. I thought it was stripping out any update key that was an empty object. IE, if `$set` never gets a value, it removes it. Is there any reason to do it out here rather than in the function itself?

https://runkit.com/crookedneighbor/lodash-3-omit-test

crookedneighbor (changes requested):

website/server/models/group.js
Hus274: Ah gotcha, that totally makes sense. When I was looking at this initially, i thought this was a comment on a block of code not on this single line which is why I didn't understand. What you suggested makes perfect sense, my mistake. I've updated the pull request accordingly.

Hus274:
Added change to move omit into the method.

Hus274:
Sorry, I totally missed that these were failing the test suite. I've made the final adjustments to get them to pass.

This should be totally ready to go now! Let me know if there's anything else you need me to change here.

website/views/shared/modals/quest-rewards.jade
```diff
@@ -12,3 +12,8 @@ script(id='partials/options.social.party.quest-rewards.html', type='text/ng-temp
         =env.t('gold')
     tr(ng-if='::quest.drop.unlock()')
       td {{::quest.drop.unlock()}}
+  div(ng-if='_.filter(quest.drop.items, \'onlyOwner\').length > 0')
```
TheHollidayInn: One last comment for this PR. Can you move `_.filter(quest.drop.items, \'onlyOwner\')` to a variable on the controller. OR, since this is a model, move it to the Modal scope?

Hus274: Yep, no problem. I'll try to be better about keeping this sort of logic out of the display layer in the future.

TheHollidayInn: Sounds good. I'm trying to get your older PRs merged in, so no worries :P

Hus274: I really appreciate that! Just trying to do my part to make your job easier! 😄

Hus274:
I've fixed the merge conflicts, added a new test case, and moved the code on the client side into a controller method. Let me know if there's anything else needed here

TheHollidayInn:
Noted towards your tier!

PR #7808, opened 2016-07-19
fix: removing party invitations

### LINKED ISSUE #7793: party invitations can't be removed
Reported by Kingkrimmson (cea63e37-cf21-4296-a2d5-416dd8664e9a) in the Report a Bug guild:

**Hey, I'm a party leader and noticed that when I try to "Boot Member" for someone who has been invited but hasn't join yet (effectively to rescind invite), they are removed from the list, but when I come back to the party screen, they're still listed. Is this a known issue, or am I missing something?**

I've seen similar comments before but looks like we don't have an issue for it. I haven't tested yet to see if it's reproducable. If anyone seeing this would like to test and comment, please do, otherwise I will later when I have some free time.

Hus274:
closes #7793 
### Changes

Fixed mistake party invitation removal code. Also fixed bug in object generator that prevented the invitees of a group from being correctly populated. 

Also did some test driven development to figure out why the existing test didn't catch this behavior before. It looks like the test was expecting the party invitations to be an array instead of an object.

In looking in the invite groups method in groups of the server controllers, I noticed that the post request awaits the return value of _inviteToGroup but it wasn't returning anything before so I fixed that as well.

---

### DIFF
--- a/test/helpers/api-integration/v3/object-generators.js
+++ b/test/helpers/api-integration/v3/object-generators.js
@@ -128,9 +128,7 @@ export async function createAndPopulateGroup (settings = {}) {
 
   await Bluebird.all(invitationPromises);
 
-  await Bluebird.all(invitees.map((invitee) => {
-    return invitee.sync();
-  }));
+  await Bluebird.map(invitees, (invitee) => invitee.sync());
 
   return {
     groupLeader,
--- a/test/helpers/api-integration/v3/object-generators.js
+++ b/test/helpers/api-integration/v3/object-generators.js
@@ -128,6 +128,10 @@ export async function createAndPopulateGroup (settings = {}) {
 
   await Bluebird.all(invitationPromises);
 
+  await Bluebird.all(invitees.map((invitee) => {
+    return invitee.sync();
+  }));
+
   return {
     groupLeader,
     group,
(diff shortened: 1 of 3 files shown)


test/helpers/api-integration/v3/object-generators.js
```diff
@@ -128,6 +128,10 @@ export async function createAndPopulateGroup (settings = {}) {
 
   await Bluebird.all(invitationPromises);
 
+  await Bluebird.all(invitees.map((invitee) => {
```
crookedneighbor: You could do 

``` js
await Bluebird.map(invitees, (invitee) => invitee.sync())
```

website/server/controllers/api-v3/groups.js
```diff
@@ -514,6 +514,7 @@ async function _inviteByUUID (uuid, group, inviter, req, res) {
     }
 
     userToInvite.invitations.party = {id: group._id, name: group.name, inviter: inviter._id};
+    return userToInvite.save();
```
crookedneighbor: Returning here skips all the email stuff that happens in the background.

Hus274: Meant to put it at the bottom of the function, that was my mistake. I'll fix this.

Hus274:
Updated that according to the comments. Should be good to go now!

audiolion:
tests and changes look good to me!

crookedneighbor:

PR #7810, opened 2016-07-19
Add API calls for shop inventories

phillipthelen:
### Changes

This pull request adds 4 new api calls, for the different shops. Each API call returns what the user currently can purchase.
Additionally it changes the webclient code for these shops to use the same code.
This also includes some tests for the content of the shops. Since the content can easily change, it only tests for the structure of the returned items. If I should add other tests, let me know.


common/script/libs/shops.js
```diff
@@ -0,0 +1,219 @@
...
+}
+
+shops.getMarketCategories = function getMarket (user, language) {
+  let categories = [];
+  let eggsCategory = {
+    identifier: 'eggs',
+    text: i18n.t('eggs', language),
+    notes: i18n.t('dropsExplanation', language),
+  };
+  eggsCategory.items = _(content.questEggs).values().filter(egg => egg.canBuy(user)).concat(_.values(content.dropEggs)).map(egg => {
```
crookedneighbor: This line is super long. Can we pull some of this out into variables?

common/script/libs/shops.js
```diff
@@ -0,0 +1,219 @@
...
+        currency: 'gems',
+        purchaseType: 'gear',
+      };
+    }).value();
+    categories.push(category);
+  }
+
+  return categories;
+};
+
```
crookedneighbor: Extra line break.

common/script/libs/shops.js
```diff
@@ -0,0 +1,219 @@
...
+      locked: false,
+      currency: 'gems',
+      purchaseType: 'food',
+    };
+  }).sortBy('key').value();
+  categories.push(foodCategory);
+
+  return categories;
+};
+
```
khipkin, line 85: remove extra nl

common/script/libs/shops.js
```diff
@@ -0,0 +1,219 @@
...
+        purchaseType: 'gear',
+      };
+    }).value();
+    categories.push(category);
+  }
+
+  return categories;
+};
+
+
```
khipkin: remove extra nl

website/views/options/inventory/drops.jade
```diff
@@ -110,61 +110,19 @@
...
+                popover='{{item.notes}}', popover-append-to-body='true',
+                popover-title!='{{item.text}}',
                 popover-trigger='mouseenter', popover-placement='top',
-                popover-append-to-body='true',
-                ng-click='purchase("food", food)')
-              p
-                |  {{::food.value}}&nbsp;
-                span.Pet_Currency_Gem1x.inline-gems
-
-        li.customize-menu
```
khipkin, line 165: What about this section about where to buy quests?

common/script/libs/shops.js
paglias: I think splitting this into multiple lines will be enough, without creating new variables

common/script/libs/shops.js
```diff
@@ -0,0 +1,219 @@
...
+shops.getQuestShopCategories = function getQuestShopCategories (user, language) {
+  let categories = [];
+
+  _.each(content.userCanOwnQuestCategories, type => {
+    let category = {
+      identifier: type,
+      text: i18n.t(`${type}Quests`, language),
+    };
+
+    category.items = _(content.questsByLevel).filter(quest => quest.canBuy(user) && quest.category === type).map(quest => {
```
paglias: I'd split this into multiple lines

website/server/controllers/api-v3/shops.js
```diff
@@ -0,0 +1,121 @@
...
+  url: '/shops/time-travelers',
+  middlewares: [authWithHeaders()],
+  async handler (req, res) {
+    let user = res.locals.user;
+
+    let resObject = {
+      identifier: 'timeTravelersShop',
+      text: res.t('timeTravelers'),
+      notes: user.purchased.plan.consecutive.trinkets > 0 ? res.t('timeTravelersPopover') : res.t('timeTravelersPopoverNoSubMobile'),
+      imageName: user.purchased.plan.consecutive.trinkets > 0 ? 'npc_timetravelers_active' : 'npc_timetravelers',
```
paglias: user.purchased.plan.consecutive.trinkets > 0 should go in a variable since it's used multiple times

common/script/libs/shops.js
```diff
@@ -36,7 +40,10 @@ shops.getMarketCategories = function getMarket (user, language) {
     text: i18n.t('hatchingPotions', language),
     notes: i18n.t('dropsExplanation', language),
   };
```
khipkin: Missing closing curly brace and closing paren?

website/views/options/inventory/drops.jade
crookedneighbor, line 165: It's been long enough now that I think we can remove it.

website/server/controllers/api-v3/shops.js
```diff
@@ -0,0 +1,122 @@
...
+  method: 'GET',
+  url: '/shops/market',
+  middlewares: [authWithHeaders()],
+  async handler (req, res) {
+    let user = res.locals.user;
+
+    let resObject = {
+      identifier: 'market',
+      text: res.t('market'),
+      notes: res.t('welcomeMarketMobile'),
```
crookedneighbor, line 26: Why are we passing this over the normal welcome? Space constraints? Should we consider just shortening the normal welcome text?

website/server/controllers/api-v3/shops.js
```diff
@@ -0,0 +1,122 @@
...
+      imageName: 'npc_alex',
```
crookedneighbor, line 27: Just thinking about the API for this, should it return the image URL instead of the image name?

website/server/controllers/api-v3/shops.js
phillipthelen, line 26: different capabilities of the markets. mobile doesn't have selling at that place and also doesn't offer some special items at the moment

common/script/libs/shops.js
```diff
@@ -0,0 +1,227 @@
...
+
+  let categories = [];
+
+  for (let key in availableSets) {
+    let category = {
+      identifier: key,
+      text: availableSets[key],
+    };
+
+    let flatGearArray = _.toArray(content.gear.flat);
```
crookedneighbor: Why recreate this variable over and over again? Shouldn't it be done outside the loop?

common/script/libs/shops.js
```diff
@@ -0,0 +1,227 @@
...
+
+    category.items = _(flatGearArray).filter((gear) => {
```
crookedneighbor: Is there a reason to use lodash over the native filter?

crookedneighbor: Never mind, you're doing it to chain the where.

common/script/libs/shops.js
```diff
@@ -0,0 +1,227 @@
...
+    });
+
+    categories.push(category);
+  }
+
+  return categories;
+};
+
+shops.getSeasonalShopCategories = function getSeasonalShopCategories (user, language) {
+  let availableSets = {
```
crookedneighbor, line 194: Is the idea to switch these out for whatever season we're in? Could you put a comment above this function to state that.

crookedneighbor:
There aren't any tests for the routes themselves. Most of the functionality is already covered in the common tests, but we should have a simple test for each route that asserts against the response object that is returned.

crookedneighbor:
All the for ins should have an object has own property check. I'm surprised that the linter didn't get angry about that. 

``` js
for (let key in content.timeTravelStable[type]) {
  if (!content.timeTravelerStable[type].hasOwnProperty(key)) {
    continue;
  }
```

test/api/v3/integration/shops/GET-shops_market.test.js
```diff
@@ -0,0 +1,20 @@
...
+  let user;
+
+  before(async () => {
+    user = await generateUser();
+  });
+
+  it('returns a valid shop object', async () => {
+    let shop = await user.get('/shops/market');
+    expect(shop.identifier).to.equal('market');
+    expect(shop.text).to.be.a('string');
```
crookedneighbor: These can test against the actual strong using the translation utility. The utility also automatically checks for missing translation strings, so it's doubly good for that.

common/script/libs/shops.js
```diff
@@ -0,0 +1,241 @@
...
+      }
+    }
+  }
+
+  let sets = content.timeTravelerStore(user.items.gear.owned);
+  for (let setKey in  sets) {
+    if (sets.hasOwnProperty(setKey)) {
+      let set = sets[setKey];
+      let category = {
+        identifier: set.key,
```
crookedneighbor, line 162: I'm not sure about this API. Do we really want to return each set as a separate category?

I mean, it makes a kind of sense, because the thing that is being sold is a set of items... but it still feels weird to me.

crookedneighbor:
Of the four routes, my only real concern is with the time traveler route. I think we should discuss how it sends down the categories for the mystery sets. We might end up just going with what we have here, but since changing it would constitute a breaking change, I want to be absolutely sure it's what we want to do before we do it.

If this stuff needs to go out imminently, I'd be fine releasing the other three routes and tabling this one until we've figured out the API for it. 

I also want to refactor some of the lodash looping stuff, but that shouldn't change the API of the route, so it can wait.

test/common/libs/shops.js
```diff
@@ -0,0 +1,72 @@
...
+} from '../../helpers/common.helper';
+
+describe('shops', () => {
+  let user = generateUser();
+
+  describe('market', () => {
+    let shopCategories = shared.shops.getMarketCategories(user);
+    it('contains at least the 3 default categories', () => {
+      expect(shopCategories.length).to.be.greaterThan(2);
+    });
```
crookedneighbor, line 14: line breaks between it blocks.

test/common/libs/shops.js
```diff
@@ -0,0 +1,72 @@
...
+    it('does not contain an empty category', () => {
+      _.each(shopCategories, (category) => {
+        expect(category.items.length).to.be.greaterThan(0);
+      });
+    });
+    it('items contain required fields', () => {
+      _.each(shopCategories, (category) => {
+        _.each(category.items, (item) => {
+          expect(item).to.have.keys(['key', 'text', 'notes', 'value', 'currency', 'locked', 'purchaseType', 'class']);
```
crookedneighbor: If these are required keys for each shop, they can be extracted into a variable at the top of the file.

crookedneighbor: Never mind, I see that is not the case.

test/common/libs/shops.js
```diff
@@ -0,0 +1,72 @@
...
+    });
+    it('items contain required fields', () => {
+      _.each(shopCategories, (category) => {
+        _.each(category.items, (item) => {
+          expect(item).to.have.keys('key', 'text', 'notes', 'value', 'currency', 'locked', 'purchaseType', 'boss', 'class', 'collect', 'drop', 'unlockCondition', 'lvl');
+        });
+      });
+    });
+  });
+  describe('timeTravelers', () => {
```
crookedneighbor, line 49: These tests should cover the logic for what happens when a user owns a mystery set, pet or mount.

test/common/libs/shops.js
```diff
@@ -0,0 +1,72 @@
...
+    });
+    it('items contain required fields', () => {
+      _.each(shopCategories, (category) => {
+        _.each(category.items, (item) => {
+          expect(item).to.have.keys(['key', 'text', 'notes', 'value', 'currency', 'locked', 'purchaseType', 'class']);
+        });
+      });
+    });
+  });
+  describe('questShop', () => {
```
crookedneighbor, line 31: There should be a test covering locked quests.

common/script/libs/shops.js
phillipthelen, line 162: My idea was/is that the json the shop returns should reflect how the shop is presented to the user, so that the client side code for the shop structure can be as simple as possible. Since the time travelers shop has the mystery sets in different categories, that's also the way the json is. The reason it's displayed like this is probably because you always purchase a full set instead of a single set item. I wouldn't change the code without then also adjusting the representation in the UI accordingly.

PR #7811, opened 2016-07-19
WIP: Monthlies

SabreCat:
### Changes

Implementation of much more sophisticated scheduling for when Dailies are active. (more thorough description coming soon)
### To Do
- [ ] Save off Dailies schedule in `_createTasks` of website/server/controllers/api-v3/tasks.js
- [ ] Cap non-subscribers' evaluation period for "Y times in X interval" at 30 days / 4 weeks / 1 month
- [ ] Check new schedule logic in `shouldDo` of common/script/cron.js

---

UUID: [Daily Monthly Yearly](https://map.what3words.com/daily.monthly.yearly)

### DIFF
--- a/common/script/cron.js
+++ b/common/script/cron.js
@@ -5,6 +5,7 @@
   ------------------------------------------------------
  */
 import _ from 'lodash';
+import reverse from 'lodash.reverse';
 import moment from 'moment';
 
 export const DAY_MAPPING = {
@@ -88,17 +89,17 @@ export function daysSince (yesterday, options = {}) {
   Should the user do this task on this date, given the task's repeat options and user.preferences.dayStart?
  */
 
-function _lastCompletedCheck (dailyHistory) {
-  _.reverse(dailyHistory); // Work in reverse chronological order
+function _missedLastDueCheck (dailyHistory) {
+  reverse(dailyHistory); // Work in reverse chronological order
   let nextIndex = 1;
   for (let entry of dailyHistory) {
     if (!dailyHistory[nextIndex]) {
-      return false; // History is empty or value static throughout; Daily was never completed or failed
+      return true; // History is empty or value static throughout; Daily was never completed or failed
     }
     if (entry.value < dailyHistory[nextIndex].value) {
-      return false; // We've found a value lower than the next older one in dailyHistory; Daily was failed
+      return true; // We've found a value lower than the next older one in dailyHistory; Daily was failed
     } else if (entry.value > dailyHistory[nextIndex].value) {
-      return true; // We've found a value higher than the next older one in dailyHistory; Daily was completed
+      return false; // We've found a value higher than the next older one in dailyHistory; Daily was completed
     } else nextIndex++; // Values are equal; Daily was skipped that day. Keep looking
   }
 }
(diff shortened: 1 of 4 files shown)


website/server/models/task.js
```diff
@@ -243,14 +243,19 @@ export let HabitSchema = new Schema(_.defaults({
 export let habit = Task.discriminator('habit', HabitSchema);
 
 export let DailySchema = new Schema(_.defaults({
-  frequency: {type: String, default: 'weekly', enum: ['daily', 'weekly']},
+  frequency: {type: String, default: 'weekly', enum: ['daily', 'weekly', 'sinceCompletion', 'timesInInterval']}, // 'Every X Interval', 'Certain Days of the Week', 'Every X Interval Since Completion', 'Y Times in X Interval'
```
crookedneighbor: This field is getting unwieldy. Let's expand it out:

``` js
frequency: {
  type: String,
  default: 'weekly',
  enum: [
    'daily', // Every X Interval
    'weekly', // Certain Days of the Week
    'sinceCompletion', // Every X Interval Since Completion
    'timesInInterval', // Y Times in X Interval
  ],
},
```

SabreCat: Got it. I'll do the same for `intervalUnit`, which is pretty stretchy too.

common/script/cron.js
```diff
@@ -88,6 +88,21 @@ export function daysSince (yesterday, options = {}) {
   Should the user do this task on this date, given the task's repeat options and user.preferences.dayStart?
  */
 
+function _lastCompletedCheck (dailyHistory) {
+  _.reverse(dailyHistory); // Work in reverse chronological order
+  let nextIndex = 1;
```
khipkin: What about 0?

common/script/cron.js
```diff
@@ -88,6 +88,21 @@ export function daysSince (yesterday, options = {}) {
...
+  for (let entry of dailyHistory) {
+    if (!dailyHistory[nextIndex]) {
+      return false; // History is empty or value static throughout; Daily was never completed or failed
```
khipkin: What about index 0?

website/server/models/task.js
```diff
@@ -247,10 +247,10 @@ export let DailySchema = new Schema(_.defaults({
     type: String,
     default: 'weekly',
     enum: [
-      'daily', // Every X Interval
+      'daily', // Every X Intervals
       'weekly', // Certain Days of the Week
       'sinceCompletion', // Every X Interval Since Completion
-      'timesInInterval', // Y Times in X Interval
+      'timesInInterval', // Y Times in X Intervals
```
khipkin, line 253: Y Times in Interval?  (just one interval)

SabreCat, line 253: What's the question? Valid setups for this might include:

3 times in 1 Month
5 times in 3 Weeks
1 time in 2 Years
1 time in 1 Day (equivalent to current default active-every-day Dailies)

Not valid (will need to catch and prevent these):
2 times in 1 Day 
90 times in 2 Months

common/script/cron.js
SabreCat: Thus `nextIndex`, not current index, which is handled within the iterator. We want to quit out if there's nothing to compare our current index against, which is the case when:

a) There is no history data at all
b) There is only one history entry
c) We've run to the end of the history without finding any entries whose values differ

I think the logic I wrote catches those? Though I am very likely wrong, and happy to be shown how, ha.

website/server/models/task.js
khipkin, line 253: Oh I see.  I didn't realize X could be >1.

SabreCat:
@TheHollidayInn has taken this over in #8444, so I'm closing this version!

PR #7818, opened 2016-07-22
Remove need for Bailey migrations

SabreCat:
### Changes

Creates a timestamp- and environment-variable-based method of determining whether or not to display a Bailey announcement. This would remove the need to make database updates to tens of thousands of users whenever a new announcement is published, changing that step to instead change a publication date in a server variable.

---

UUID: [News Clock Stamp](https://map.what3words.com/news.clock.stamp)

### DIFF
--- a/website/client/js/controllers/notificationCtrl.js
+++ b/website/client/js/controllers/notificationCtrl.js
@@ -180,7 +180,7 @@ habitrpg.controller('NotificationCtrl',
     });
 
     // Show new-stuff modal on load
-    if (User.user.flags.newStuff)
+    if (moment(User.user.flags.newStuffSeenDate).toDate() < moment(env.NEWS_DATE).toDate())
       $rootScope.openModal('newStuff', {size:'lg'});
   }
 ]);
--- a/website/client/js/controllers/notificationCtrl.js
+++ b/website/client/js/controllers/notificationCtrl.js
@@ -180,7 +180,8 @@ habitrpg.controller('NotificationCtrl',
     });
 
     // Show new-stuff modal on load
-    if (moment(User.user.flags.newStuffSeenDate).toDate() < moment(env.NEWS_DATE).toDate())
+    if (moment(User.user.flags.newStuffSeenDate).isBefore(env.NEWS_DATE)) {
       $rootScope.openModal('newStuff', {size:'lg'});
+    }
   }
 ]);
--- a/website/client/js/controllers/notificationCtrl.js
+++ b/website/client/js/controllers/notificationCtrl.js
@@ -180,8 +180,7 @@ habitrpg.controller('NotificationCtrl',
     });
 
     // Show new-stuff modal on load
-    if (moment(User.user.flags.newStuffSeenDate).isBefore(env.NEWS_DATE)) {
+    if (User.user.flags.newStuff)
       $rootScope.openModal('newStuff', {size:'lg'});
-    }
   }
 ]);
--- a/website/client/js/controllers/rootCtrl.js
+++ b/website/client/js/controllers/rootCtrl.js
@@ -28,7 +28,7 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
     $rootScope.$on('$stateChangeSuccess',
       function(event, toState, toParams, fromState, fromParams){
         $rootScope.pageTitle = $state.current.title;
-        
+
         if (!($state.current.name in IGNORE_SCROLL_PAGES)) {
             $window.scrollTo(0, 0);
         }
@@ -159,7 +159,7 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
     }
 
     $rootScope.dismissAlert = function() {
-      $rootScope.set({'flags.newStuffSeenDate':moment()});
+      $rootScope.set({'flags.newStuffSeenDate':new Date()});
     }
 
     $rootScope.acceptCommunityGuidelines = function() {
--- a/website/client/js/controllers/rootCtrl.js
+++ b/website/client/js/controllers/rootCtrl.js
@@ -28,7 +28,7 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
     $rootScope.$on('$stateChangeSuccess',
       function(event, toState, toParams, fromState, fromParams){
         $rootScope.pageTitle = $state.current.title;
-
+        
         if (!($state.current.name in IGNORE_SCROLL_PAGES)) {
             $window.scrollTo(0, 0);
         }
@@ -159,7 +159,7 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
     }
 
     $rootScope.dismissAlert = function() {
-      $rootScope.set({'flags.newStuffSeenDate':new Date()});
+      $rootScope.set({'flags.newStuff':false});
     }
 
     $rootScope.acceptCommunityGuidelines = function() {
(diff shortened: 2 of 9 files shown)


audiolion:
Hey, learning the codebase still -- where is the removal of the line that calls the bailey migration? Is the code to do the bailey migration still required to stay in the system with a deprecation warning to give api users time to migrate to the new solution, that is if this is even an api callable item! :P

SabreCat:
@audiolion: Migrations are run directly against the database; there isn't a trigger to make it happen, exactly. The script we've historically run for announcements can be found at /migrations/new_stuff.js.\* Marking that as deprecated somehow might be a good idea, aye!

~~*Though it hasn't actually worked for me since an upgrade to the database a few months ago. I've been running the script from within the Mongo shell. @Alys and I need to get that straightened out by end of month so we can do the awards for Take This and Habitica Naming Day.~~

crookedneighbor:
💯  💯  💯  💯  💯  💯  💯  💯  💯

website/client/js/controllers/notificationCtrl.js
```diff
@@ -178,7 +178,7 @@ habitrpg.controller('NotificationCtrl',
     });
 
     // Show new-stuff modal on load
-    if (User.user.flags.newStuff)
+    if (moment(User.user.flags.newStuffSeenDate).toDate() < moment(env.NEWS_DATE).toDate())
```
crookedneighbor: Couldn't we use

``` js
if (moment(User.user.flags.newStuffSeenDate).isBefore(env.NEWS_DATE)) {
```

http://momentjs.com/docs/#/query/is-before/

website/client/js/controllers/rootCtrl.js
```diff
@@ -152,7 +152,7 @@ habitrpg.controller("RootCtrl", ['$scope', '$rootScope', '$location', 'User', '$
     }
 
     $rootScope.dismissAlert = function() {
-      $rootScope.set({'flags.newStuff':false});
+      $rootScope.set({'flags.newStuffSeenDate':moment()});
```
crookedneighbor: Does this new `toDate()`?

website/client/js/controllers/settingsCtrl.js
```diff
@@ -64,7 +64,7 @@ habitrpg.controller('SettingsCtrl',
     }
 
     $scope.showBailey = function(){
-      User.set({'flags.newStuff':true});
+      User.set({'flags.newStuffSeenDate':moment(env.NEWS_DATE).subtract(1,'days').toDate()});
```
crookedneighbor: Can we pull out this value into a variable?

website/client/js/controllers/rootCtrl.js
paglias: no, it needs a call to .toDate(). Better to use new Date

paglias:
Can we re-use flags.newStuff instead of adding a new field?

crookedneighbor:
@vIiRuS the mobile apps use that flag?

phillipthelen:
Yes, both apps use the newStuff flag already and expect it to be a boolean

crookedneighbor:
Changing the value type of the newStuff flag would be a breaking change. I don't think we document it anywhere, so we technically can do it. However, integrations may still use it (including the mobile apps)

crookedneighbor:
We could do the calculation on the server as middleware for the get user route. Mark the newFlagDate as private and then we're free to change how it's implemented as we see fit.

paglias:
I like that 👍 less stuff on the client is better

SabreCat:
Addressed line comments and rebased to resolve merge conflicts. Will take a stab at the middleware refactor next.

SabreCat:
OK, first stab at moving the logic server side. The idea is that there will be zero changes to clients; they continue to say "set `flags.newStuff` to true" and so on, but the server then translates that into manipulations of `newStuffSeenDate` relative to the env variable. Is that a step in the right direction?

paglias:
I'd do it this way!

Il 09 set 2016 23:27, "Sabe Jones" notifications@github.com ha scritto:

crookedneighbor:
@SabreCat I don't think the client needs to have any knowledge about the date string.

``` js
// psuedo code for should newStuff flag be true

// model
// set flags._newStuffSeenDate as private

// middleware
let nconf = require('nconf');
const NEWS_DATE = nconf.get('NEWS_DATE');

function checkNewsFlag (req, res, next) {
  let user = res.locals.user;
  if (user.flags.newStuff) {
    next(); // exit early if user already has newStuff flag as true
    return;
  }

  user.flags.newStuff = moment(user.flags._newStuffSeenDate).isBefore(NEWS_DATE);
  // unsure if we need to save the user. Probably do.
  next();
}

// GET user route
middlewares: [authWithHeaders(), checkNewsFlag],

// client
// no change to show bailey logic
// for the show bailey button in settings, you can just open up the new stuff modal instead or make no changes do like we've done
```

paglias:
I think user.flags.newStuff should be removed from the model and simply
returned as a virtual property like we do for computed stats

Il 10 set 2016 00:26, "Blade Barringer" notifications@github.com ha
scritto:

crookedneighbor:
Me too.

website/server/controllers/api-v3/user.js
```diff
@@ -27,16 +29,18 @@ let api = {};
...
   async handler (req, res) {
     let user = res.locals.user;
     let userToJSON = user.toJSON();
+    let env = res.locals.habitrpg;
 
     // Remove apiToken from response TODO make it private at the user level? returned in signup/login
     delete userToJSON.apiToken;
 
     user.addComputedStatsToJSONObj(userToJSON);
+    userToJSON.flags.newStuff = moment(user.flags.newStuffSeenDate).isBefore(env.NEWS_DATE);
```
TheHollidayInn, line 43: Shouldn't you just use `nconf.get('NEWS_DATE')`?

website/server/controllers/api-v3/user.js
```diff
@@ -137,6 +142,16 @@ let checkPreferencePurchase = (user, path, item) => {
   return _.get(user.purchased, itemPath);
 };
 
+function _setNewStuffFlag (val, res) {
+  if (val === true) {
+    return moment(res.locals.habitrpg.NEWS_DATE).subtract(1, 'days').toDate();
```
TheHollidayInn, line 147: Again, probably best to use nconf here. Maybe we can move this function to the user model?

website/server/controllers/api-v3/user.js
```diff
@@ -148,7 +163,7 @@ let checkPreferencePurchase = (user, path, item) => {
  */
 api.updateUser = {
   method: 'PUT',
-  middlewares: [authWithHeaders()],
+  middlewares: [authWithHeaders(), locals],
```
TheHollidayInn, line 166: I don't see where you are using locals here.

website/server/controllers/api-v3/user.js
```diff
@@ -161,6 +176,10 @@ api.updateUser = {
       }
 
       if (acceptablePUTPaths[key]) {
+        if (key === 'flags.newStuff') {
+          val = _setNewStuffFlag(val, res);
+          key = 'flags.newStuffSeenDate';
```
TheHollidayInn, line 181: Shouldn't we also be setting newStuff to false. I'm not sure where we would use newStuffSeendDate if we just toggle newStuff. 

Also, maybe we should add analytics here to tracking them looking?

website/server/controllers/api-v3/user.js
paglias: 👍 I don't think we want the entire locals middleware just for this, since it includes a lot of things only used for Jade views

SabreCat:
Closing this for now as a stale project. We can definitely revisit this, but at the moment updating the `newStuff` flag is a trivial amount of overhead.

PR #7821, opened 2016-07-24
Xp bar overfilled clarification

### LINKED ISSUE #2433: Failed to Level Up after gaining Experience (XP) from Completing Quest
_edit by admin:_
# What to do if you notice this bug:

**You will level up as soon as you complete a Daily or To-Do, or use a Habit or Reward. None of your Experience will be lost.**

**You don't need to report it because we know that it happens for all players if they gain enough Experience to put them over the limit for levelling up from any of these causes:**
- **completing a quest**
- ~using the Enchanted Armoire~ - fixed in https://github.com/HabitRPG/habitica/pull/11884
- ~repeating steps in the tutorial that new players receive~ - tutorial works differently now and can't cause this bug

This issue is currently on hold for the reasons given here: https://github.com/HabitRPG/habitica/issues/2433#issuecomment-635062474

_end of edit by admin_

---

My currently experience is 358/350 (Level 15). I probably should have leveled ... to 16.

The last experience I got was 100 XP from the Trapper Santa party quest. I got the 20 gold just fine, already had the polar bear mount, and while I'm not sure if I got the right amount of experience, I'm fairly certain 100XP would have put me over to the next level.

I refreshed the page and I was still at 358/350 XP without level.

habitrpg-did-not-level-from-quest

I haven't tried checking off another box, but I'll update when I get around to that.

audiolion:
Fixes #2433 
### Changes

Adds a text label to exp bar informing user to complete a task to level up when they have enough exp to level.

Media queries for xs screens used to move label below exp bar so it is not cut off.

Refactored jade template to allow the UserCtrl to apply to the .hero-stats div as well, they are all logically contained in the same unit of displaying user info.

The refactor allowed me to replace some of the jade math in the template for displaying HP, EXP, and MP to use the built in functions in statServices service. The benefit is code is being reused rather than having multiple implementations, so refactoring is easier if changes need to be made and lesser chance bugs are introduced. Furthermore it reduces the number of watchers on the header.jade section. Finally, the functions have tests against them and cover additional edge cases for displaying hp, mp and exp.
#### >XS Screen Sizes

Media Query Example
#### <=XS Screen Size

Media Query Example

---

UUID: 236e74bc-986d-41ad-9fa1-d4842ad72334

### DIFF
--- a/test/spec/services/statServicesSpec.js
+++ b/test/spec/services/statServicesSpec.js
@@ -139,22 +139,23 @@ describe('Stats Service', function() {
   });
 
   describe('canLevelUp', function() {
-  	it('returns true if a user has enough exp to level up', function() {
-  		user.stats.exp = 150;
-  		user.stats.lvl = 1;
-  		var canLevelUp = statCalc.canLevelUp(user);
+    it('returns true if a user has enough exp to level up', function() {
+      user.stats.exp = 150;
+      user.stats.lvl = 1;
+      var canLevelUp = statCalc.canLevelUp(user);
 
-  		expect(canLevelUp).to.eql(true);
-  	});
+      expect(canLevelUp).to.eql(true);
+    });
 
-  	it('returns false if a user doesn\'t have enough exp to level up', function() {
-  		user.stats.exp = 149;
-  		user.stats.lvl = 1;
-  		var canLevelUp = statCalc.canLevelUp(user);
+    it('returns false if a user doesn\'t have enough exp to level up', function() {
+      user.stats.exp = 149;
+      user.stats.lvl = 1;
+      var canLevelUp = statCalc.canLevelUp(user);
 
-  		expect(canLevelUp).to.eql(false);
-  	});
+      expect(canLevelUp).to.eql(false);
+    });
   });
+
   describe('equipmentStatBonus', function() {
     it('tallies up stats from equipment that is equipped', function() {
       var equippedGear = {
--- a/test/spec/services/statServicesSpec.js
+++ b/test/spec/services/statServicesSpec.js
@@ -138,6 +138,23 @@ describe('Stats Service', function() {
     });
   });
 
+  describe('canLevelUp', function() {
+  	it('returns true if a user has enough exp to level up', function() {
+  		user.stats.exp = 150;
+  		user.stats.lvl = 1;
+  		var canLevelUp = statCalc.canLevelUp(user);
+
+  		expect(canLevelUp).to.eql(true);
+  	});
+
+  	it('returns false if a user doesn\'t have enough exp to level up', function() {
+  		user.stats.exp = 149;
+  		user.stats.lvl = 1;
+  		var canLevelUp = statCalc.canLevelUp(user);
+
+  		expect(canLevelUp).to.eql(false);
+  	});
+  });
   describe('equipmentStatBonus', function() {
     it('tallies up stats from equipment that is equipped', function() {
       var equippedGear = {
--- a/website/client/js/services/statServices.js
+++ b/website/client/js/services/statServices.js
@@ -131,6 +131,6 @@
       mountMasterProgress: mountMasterProgress,
       mpDisplay: mpDisplay,
       totalCount: totalCount
-    }
+    };
   }
 }());
--- a/website/client/js/services/statServices.js
+++ b/website/client/js/services/statServices.js
@@ -61,6 +61,13 @@
       return display;
     }
 
+    function canLevelUp(user) {
+      var exp = Math.floor(user.stats.exp);
+      var toNextLevel = Shared.tnl(user.stats.lvl);
+
+      return (exp - toNextLevel) >= 0;
+    }
+
     function goldDisplay(gold) {
       var display = Math.floor(gold);
       return display;
@@ -117,6 +124,7 @@
       classBonus: classBonus,
       equipmentStatBonus: equipmentStatBonus,
       expDisplay: expDisplay,
+      canLevelUp: canLevelUp,
       goldDisplay: goldDisplay,
       hpDisplay: hpDisplay,
       levelBonus: levelBonus,
--- a/website/views/shared/header/header.jade
+++ b/website/views/shared/header/header.jade
@@ -23,7 +23,6 @@
             a(ng-click='toggleChart("exp")').glyphicon.glyphicon-signal
           span
           | {{statCalc.expDisplay(user)}}
-      span(ng-show='statCalc.canLevelUp(user)' class="visible-xs pull-right")=env.t('canLevelUp')
       .meter-label(tooltip=env.t('mana'), ng-if='user.flags.classSelected && !user.preferences.disableClasses')
         span.glyphicon.glyphicon-fire
       .meter.mana(ng-if='user.flags.classSelected && !user.preferences.disableClasses', tooltip='{{Math.round(user.stats.mp * 100) / 100}}')
@@ -31,6 +30,7 @@
         span.meter-text.value
           span
           | {{statCalc.mpDisplay(user)}}
+      span(ng-show='statCalc.canLevelUp(user)' class="visible-xs pull-right")=env.t('canLevelUp')
     // party
     .party(ng-controller='PartyCtrl')
       button.party-invite.btn.btn-primary(ng-click="inviteOrStartParty(party)",
--- a/website/views/shared/header/header.jade
+++ b/website/views/shared/header/header.jade
@@ -1,9 +1,9 @@
 .header-wrap(ng-controller='HeaderCtrl')
   a.label.label-default.undo-button(x-bind='click:undo', ng-show='_undo')=env.t('undo')
 
-  header.site-header(ng-hide='user.preferences.hideHeader', role='banner', data-partysize='{{party.members.length>1 ? truarr(party.members.length) : 0}}')
+  header.site-header(ng-controller='UserCtrl', ng-hide='user.preferences.hideHeader', role='banner', data-partysize='{{party.members.length>1 ? truarr(party.members.length) : 0}}')
     // avatar
-    .herobox-wrap.main-herobox(ng-controller='UserCtrl')
+    .herobox-wrap.main-herobox
       +herobox({main:1})
     // stat bars
     .hero-stats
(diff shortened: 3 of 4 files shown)


test/spec/services/statServicesSpec.js
```diff
@@ -138,6 +138,23 @@ describe('Stats Service', function() {
     });
   });
 
+  describe('canLevelUp', function() {
+  	it('returns true if a user has enough exp to level up', function() {
+  		user.stats.exp = 150;
```
audiolion: The other tests use magic numbers -- as a side note these should really be lookups based on a dictionary of how much exp is needed to get to level 2. If it ever changes these tests will all fail and to anyone making new or more nuanced tests it is a lot harder to figure out what magic numbers to use.

website/views/shared/header/header.jade
```diff
@@ -12,23 +12,25 @@
       .meter.health(tooltip='{{Math.round(user.stats.hp * 100) / 100}}')
         .bar(ng-style='{"width": Shared.percent(user.stats.hp, Shared.maxHealth)+"%"}')
         span.meter-text.value
-          | {{Math.ceil(user.stats.hp)}} / {{::Shared.maxHealth}}
+          | {{statCalc.hpDisplay(user.stats.hp)}}
```
audiolion, line 15: the hpDisplay function is the only one out of the the statServices functions that takes in the specific current hp value rather than a `user` representation. I wanted to refactor this but it was outside the scope of this PR.

website/views/shared/header/header.jade
```diff
@@ -12,23 +12,25 @@
...
       .meter-label(tooltip=env.t('experience'))
         span.glyphicon.glyphicon-star
       .meter.experience(tooltip='{{Math.round(user.stats.exp * 100) / 100}}')
         .bar(ng-style='{"width": Shared.percent(user.stats.exp,Shared.tnl(user.stats.lvl))+"%"}')
         span.meter-text.value
+          span(ng-show='statCalc.canLevelUp(user)' class="hidden-xs") Complete a Task to Level Up!
```
audiolion: I tried putting this in a `bootstrap label`, and tried a few variations: `label-success`, `label-info`, `label-default` etc. but they didn't look good so I stuck with just plain text.

audiolion:
I placed some comments throughout the committed files to explain though process.

On a side note I really think this text should be a button that appears, upon clicking calls a `database sync` with no changes, what if a user wants to level up but doesn't have a task to complete, they have to fake a task to level up when they should just level up.

Steps to reproduce xp bar overfill in production:
1. Register for a new account
2. After load finishes, complete the `Productive Day` default task
3. Click through the rest of the tutorial menus which grant total `+145 exp`

Result is you will have `151 / 150 exp`

ghost:
For what it's worth  I'd suggest making the text translatable and testing the way it looks like with the mana bar

crookedneighbor:
It should definitely be translatable. 

Can we make the text replace the experience numbers instead of being pre pended to it?

Alys:
I think it's important to still show the experience numbers so they know why they can level up and because their actual experience will be of interest to some players.

test/spec/services/statServicesSpec.js
```diff
@@ -138,6 +138,23 @@ describe('Stats Service', function() {
...
+  	});
+
+  	it('returns false if a user doesn\'t have enough exp to level up', function() {
+  		user.stats.exp = 149;
+  		user.stats.lvl = 1;
+  		var canLevelUp = statCalc.canLevelUp(user);
+
+  		expect(canLevelUp).to.eql(false);
+  	});
+  });
```
crookedneighbor, line 157: Add a line break here.

test/spec/services/statServicesSpec.js
```diff
@@ -138,6 +138,23 @@ describe('Stats Service', function() {
     });
   });
 
+  describe('canLevelUp', function() {
+  	it('returns true if a user has enough exp to level up', function() {
```
crookedneighbor: The indentation is off. We use 2 space indentation.

audiolion:
Ah, apologies on the translation, first time working on this project and also first time working on software that has translations! I have seen the env.t('text goes here that should be translated'), just gotta figure out how that works (does it auto translate or do I have to find translations?).

Weird that the eslint didn't pick up the indentation error, ill fix those items.

audiolion:
If anyone can help me figure out where `env.t('some text to translate')` points to in the codebase and/or how to implement translations that would be great. It looks like we integrate with transifex? I am just not sure how it works.

edit:

I think I figured it out, added default en/us english translation and I believe transifex takes care of the rest.

crookedneighbor:
There are json files in common/locales/en. Should be obvious once you take a peek in there. Add only to the en folder. Do not mainly add to any other folders.

audiolion:
@GitHubSphinx You were right, mana bar got muddled, here are updated screenshots:

XP Message with Mana Bar >=SM Screen Size

XP Message with Mana Bar <=XS Screen Size

The without mana bar view looks exactly the same.

and uh, I took the first screenshot with the wrong text in the translations json file, it was fixed to be the same as prior screenshots in the media query XS screen size screenshot. You can see this from the git commit.

website/client/js/services/statServices.js
```diff
@@ -131,6 +131,6 @@
       mountMasterProgress: mountMasterProgress,
       mpDisplay: mpDisplay,
       totalCount: totalCount
-    }
+    };
```
audiolion, line 134: this was a missing semicolon from previous versions of the file, changed to improve consistency and code climate rating

Alys:
@audiolion A useful tip for finding out how existing code works is `git grep "some text to translate"` or `git grep "somePieceOfCode"`

`git grep "some text to translate" | grep "locales/en/"` makes text searches less spammy.

crookedneighbor:
One treason I thought it better to replace the numbers is for screen real estate on small screens.

audiolion:
I posted a pic on XS screen size. Ill do another tomorrow to show what it looks like on mobile web browser

ghost:
@crookedneighbor :+1: 
@audiolion as far as I remember the bars also shrink when a user is in a party.

Added: BTW, from what I've seen in this regard mobile browsers usually behave the same way as you can get on a computer by changing the width of the window to a minimum just  by dragging vertical borders of the window. It's not necessarily true, but can save some effort (not to connect a mobile device to your local Habitica)

ghost:
Suggestion for discussion: is it possible to add a special mark/emoji (like upwards arrow or question mark in circle or anything) to the bar when it's overfilled? In this case there could be a mouseover balloon tooltip with the explanation.

Alys:
@GitHubSphinx It would be possible, but I'm not certain it would be meaningful enough to all players. Some of them miss the question mark icons that are already on the site. Very occasionally, we are even asked what the question mark icons mean.

ghost:
@Alys I've looked at the screenshot in the original ticket - the one in the opening post with the bar being very short in the "party layout".

pic 
It just feels that no meaningful text could fit in there

audiolion:
@githubsphinx that is how I normally do mobile screenshots :P

Ill see what happens with party members and add them in

audiolion:
@GitHubSphinx good catch with the party ui interfering.

Here is what it looks like when you add an overflow of party members (it starts clipping at ~5 on smaller screen sizes before hitting the XS breakpoint)

clipped

New suggestion to accommodate both small and large screens, party or no party: display it below always. Variations:

#### Without Label

without label

#### With bootstrap label

with normal bootstrap label

#### With bootstrap label overriding some defaults

with label overriding bootstrap defaults

The benefit of the bootstrap label is that it spans to fit the size of the container without wrapping. This view is exactly how it looks like on mobile, with more of the party on the right cut off.

ghost:
Just to make sure, will it be below mana bar?

audiolion:
@GitHubSphinx Yeah, here is a ss

#### With mana bar, bootstrap default overridden label

with mana bar and label overridden

edit:

#### Mobile view, for edification, 400px wide

mobile view

ghost:
I personally like the way it looks. I'm a little bit worried about the no
wrapping part. For the sake of testing could you please duplicate the text
(or just invent anything twice as long) to see how it will look like in
case a localized text is longer.

audiolion:
Good idea to check what happens with longer strings.

#### Longer string at big screen size

longer string

The longer string and bootstrap label forces the character window to be larger to fit the text

#### On mobile screen size

mobile with longer string

Continues to span the necessary display size to display on one line. The mobile user can scroll right to see the rest of the text

mobile longer string scrolled right

The problem with wrapping text is that it starts to break other UI elements

#### Wrapping text overflows into other elements

example of wrapping text

If I enforce wrapping on the label it looks a little better but the UI's are still overlapping.

label with wrapping text

#### UI Overlap

The source of the UI Overlap is this html code

``` html
<div id=#sticky-wrapper class="sticky-wrapper" style="height: 195px;">
```

This div wraps the entire header, and it enforces it to be 195px even though #sticky-wrapper sets the element to be "height: 215px;" by default. The issue is that the 195px isn't enough so the text overflows and the other elements get lumped together, when I increase the size to be `height: 215px;` it is almost there, a slightly larger height makes it fit well e.g. `height: 230px`.

230px height

I am not sure if I should be overriding this default. I don't know how long translations can be, this string is over twice the length so it should _probably_ encapsulate all translation sizes, but I am just guessing here. It would seem like it would be better if the header size just adapted to the max size it needs as `container-fluid` will automatically do, but I also understand that the app then has a different look and feel across translations as the header is slightly different heights.

Alys:
I think it's okay to have it not wrapping and pushing the stats section to be wider because this message won't be shown often or for long. The most common time to see it will be for a new user who's just done the tutorial and they often won't be in a party, or not in a large party.

Is it possible to put the message directly below the XP bar (above the mana bar)? It will be more obvious there when people are looking at too-much XP and wondering why it's happening.

audiolion:
Ok, here is what it looks like placed above mana bar and below exp bar

alys suggestion

We are getting a bit of clipping, the issue being each of the bar's have css 'margin-bottom: 1em;' applied to them, the label to look centered also needs 'margin-bottom: 1em;' and with that it starts to get pushed out of the way.

I could apply a negative margin to the label which would also squeeze it in and maybe that works well enough:

alys suggestion2

Realistically though, we would push the tasks container farther away, I am just not sure how to accomplish that. What is happening here is the divs are overlapping in such a way that you can't access the horizontal slider.

slider blocked

As you can see from above, the padding is completely overlapping the horizontal slider making it impossible to click and grab it unless you are within that 1-2px tolerance.

What I need to do is push the tasks pane farther away if we go with the above, and figure out how to do that 😉 

edit:

It is worth noting this padding issue also happens on live, so maybe this is a separate issue and I should leave it as-is for this PR/Issue?

audiolion:
any direction on this? the extra label exacerbates the scrollbar padding overlap issue as shown above, that issue I believe occurs on live as well. I can in this PR add in code that helps fix the overlap though css styling.

If everything is ok with how it looks ill remove the inline styles in the template and create css classes or use existing ones that meet the same need.

khipkin:
@ESWAT seems to be a CSS wizard, so maybe they could offer some input?

ESWAT:
Hmm to be honest it’s hard for me to tell what the issue is by reading the comments. Though here’s my two cents based on my uneducated idea of the problem:
1. Maybe align elements in the `.hero-stats` element using flexbox and let it worry about aligning things properly (might have to set a `max-height` on that element to make sure it doesn’t leak into the content below?
2. Instead of showing text under the bars, why not make the XP icon glow noticeably and put a tooltip directive on it with the same explanation text? Given there aren’t too many things in the UI currently that animate this would probably be noticeable enough (pure assumption the user would hover or tap on the icon after noticing this to bring up the tooltip though).

khipkin:
@ESWAT Just my opinion, but I think something as subtle as "glowing" might not work for visually-impaired users.  Although, if the added text is still somewhere in the html where screenreaders can get to it, that may not matter.

ESWAT:
@khipkin that's totally fair.

audiolion:
Yeah I can do a screenreader available text and a glowing animation with a tooltip.

crookedneighbor:
Just a thought. Could we try calling the updateStats function on sync? https://github.com/HabitRPG/habitrpg/blob/develop/common/script/fns/updateStats.js#L9

If that works, I don't think we need this message.

audiolion:
we could, it is probably the most sane way to combat this problem, it was deemed [too expensive](https://github.com/HabitRPG/habitrpg/issues/2433#issuecomment-217822716) of an operation and so we started down this alternate route, if this is the way you want to go I am fine with that

crookedneighbor:
I could be wrong, but I think that was referring to doing the checjI on the server for each member of a party. This would be using your should level up function on the client, and if the user should level up, using the update state function.

crookedneighbor:
@audiolion what's the status on this?

audiolion:
hey, sorry been really busy as of late, just got married and am on honeymoon, I can start looking at this next week, if you want to we can close this PR and open a new PR as we are going a completely different route and I don't know as if having the history on the eventual merge will be helpful, and the closed pr will still be linked to the issue for reference

crookedneighbor:
Going to close this. Open a new PR when you're ready.

PR #7823, opened 2016-07-24
Added new user flag when user registers with facebook

TheHollidayInn:
Fixes put_issue_url_here
### Changes

When a new user registers with Facebook, the api will not return a newUser value of true. This is used in the app to tell when a user is registering or logging in via social media. 

---

UUID:

### DIFF
--- a/website/server/controllers/api-v3/auth.js
+++ b/website/server/controllers/api-v3/auth.js
@@ -162,10 +162,8 @@ api.registerLocal = {
 };
 
 function _loginRes (user, req, res) {
-  var newUser = false;
-  if (user.newUser) newUser = true;
   if (user.auth.blocked) throw new NotAuthorized(res.t('accountSuspended', {userId: user._id}));
-  return res.respond(200, {id: user._id, apiToken: user.apiToken, newUser});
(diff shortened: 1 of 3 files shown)


website/server/controllers/api-v3/auth.js
```diff
@@ -162,8 +162,10 @@ api.registerLocal = {
 };
 
 function _loginRes (user, req, res) {
+  var newUser = false;
```
crookedneighbor: You can just do `var isNewUser = user.newUser === true`

crookedneighbor:
There should be test coverage around this. Other than that, seems sound to me.

website/server/controllers/api-v3/auth.js
paglias: Should be let not var

paglias: Actually you don't need to create a new car at all. You can just pass user.newUser || false

crookedneighbor: +1 to that

TheHollidayInn:
Updated!

paglias:
@TheHollidayInn looks good to me 👍

PR #7829, opened 2016-07-25
Align guild gems box with other page elements

### LINKED ISSUE #7796: Guild gems box doesn't align with anything else
General Info
- UUID: `0e1d15f7-f36b-4f15-a9e1-0c31b732dac9`
- Browser: Chrome 51
- OS: Ubuntu 14.04
### Description

image

IMO the guild gems box looks kind of odd stuck to the right-hand side like that. I don't know what it looked like before (or even if it changed, I might just be going mad), but I think it might look better if it 
aligned with the infobox on the left and speech bubble below and had centred text and icon:

image

If people think this looks better, I'd like to work on it.
#### Console Errors

n/a

keysmashes:
Fixes #7796
### Changes

Aligns right side with speech bubble below, and aligns top with description box to the left. Also centres text and gem icon.

Old:

image

New:

image

One side-effect of these changes is that the 'guild gems' text is now underlined on hover. Hopefully that doesn't matter? I've no idea why it happens.

I still need to work on this to fix issues created by this change in other places, like the Fortify modal - applying `display: flex` to align the text vertically causes the green plus box visible on hover to be misplaced on top of the gem count, rather than to the side [solution - `display: inline-flex`]. Also, the '+' icon is underlined, which is a problem [fixed with inline-flex].

---

UUID: `0e1d15f7-f36b-4f15-a9e1-0c31b732dac9`

### DIFF
--- a/website/client/css/tasks.styl
+++ b/website/client/css/tasks.styl
@@ -517,6 +517,8 @@ form
   border-radius: 4px
   display: inline-flex
   align-items: center // vertically align text
+  .Gems
+    margin-top: 0px
 
 .task-options
   .task-action-btn.tile
--- a/website/client/css/tasks.styl
+++ b/website/client/css/tasks.styl
@@ -509,13 +509,15 @@ form
   background-color: lighten($best, 20%)
   &:hover, &:focus
     background-color: lighten($best, 40%)
+// this affects the guild gems box, maybe other stuff?
 .tile.flush
   margin-left: 0
   border: 1px solid rgba(0,0,0,0.2)
   outline: 0
   line-height: 2em
-  &:first-child
-    border-right: 0
+  border-radius: 4px
+  display: flex
+  align-items: center // vertically align text
 
 .task-options
   .task-action-btn.tile
--- a/website/client/css/tasks.styl
+++ b/website/client/css/tasks.styl
@@ -516,7 +516,7 @@ form
   outline: 0
   line-height: 2em
   border-radius: 4px
-  display: flex
+  display: inline-flex
   align-items: center // vertically align text
 
 .task-options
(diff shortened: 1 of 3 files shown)


keysmashes:
I fixed places like the Fortify modal so the green box appears in its proper place, and I stopped the text being underlined on hover:

image

image

image

and on hover:

image

image

image

The Purchase Gems modal looks just as it did:

image

and the original point of the PR looks just the same:

image

website/client/css/tasks.styl
```diff
@@ -509,13 +509,15 @@ form
   background-color: lighten($best, 20%)
   &:hover, &:focus
     background-color: lighten($best, 40%)
+// this affects the guild gems box, maybe other stuff?
```
khipkin: Remove this comment.

keysmashes: Done.

crookedneighbor:
Personally, I think the rounded gem box with the rounded plus button looks weird. Since we're about to do a redesign of the site, I'm not going to sweat small graphical things like that, so if @Alys and @lemoness are fine with it, I'm fine with merging this.

keysmashes:
@crookedneighbor I also wasn't sure about that - I can change it so that both boxes always have squared corners in modals with that green box (or revert to squared corners throughout for consistency), or I can have the corners that touch between the two boxes square off when the green box is visible, which would look something like this:

image

I can nudge the gem icon up (really, stop it being pushed down) to align it properly as well:

image

I think that looks better, but I guess you might not want tons more styling if there's about to be a redesign; it's your call.

crookedneighbor:
^ I like that

keysmashes:
It now looks like this:

image

image

Other things look the same as they did.

lemoness:
I agree that we can revisit any changes post-redesign, so Blade, if you're
comfortable, go ahead and merge.

crookedneighbor:
Notated toward your next tier!

PR #7840, opened 2016-07-27
Enable guild challenge prize to use guild bank gems

### LINKED ISSUE #7361: can't use guild bank gems for challenge prize
General Info
- UUID: 
- Browser: opera
- OS: Android
### Description

When a guild leader creates a guild challenge and sets a challenge prize the site tells the maximum amount is the personal gem count and it doesn't include the 4 gems from the guild bank. In other words challenges with prize exceeding your own gems won't save even if your guild bank has more of them.

Ozzystrasza:
Fixes https://github.com/HabitRPG/habitrpg/issues/7361
### Changes

The problem here was that on the function [_calculateAvailableGroupBalance(gid)](https://github.com/HabitRPG/habitrpg/blob/develop/website/client/js/controllers/challengesCtrl.js#L431) the `if (group && group.balance && group.leader === User.user._id)` is never true because `group.leader` is an unidentified object. So I tried to find the root of the problem. The object group comes from $scope.groups, which comes from Groups.Group.getGroups which is not returning the leader property from the query.

So I went to [`schema.statics.getGroups`](https://github.com/HabitRPG/habitrpg/blob/develop/website/server/models/group.js#L167) on group.js to see what I could do. I saw `populateLeader = false` and changed it to `true` but it didn't fix the problem. So I went to `basicFields` and added 'leader' and that made the query return the leader property so that `if (group && group.balance && group.leader === User.user._id)` could work properly again. I think that change shouldn't affect somewhere else in a bad way, but if it does plz let me know!

Here are some tests:

Guild leader creating a challenge with the sum of his gems plus guild gems:
test1

Guild leader creating a challenge only with guild gems:
test2

Guild member (not leader) creating a challenge:
test3

---

### DIFF
--- a/test/api/v3/integration/challenges/GET-challenges_challengeId.test.js
+++ b/test/api/v3/integration/challenges/GET-challenges_challengeId.test.js
@@ -46,7 +46,14 @@ describe('GET /challenges/:challengeId', () => {
         id: groupLeader._id,
         profile: {name: groupLeader.profile.name},
       });
-      expect(chal.group).to.eql(_.pick(group, ['_id', 'id', 'name', 'type', 'privacy']));
+      expect(chal.group).to.eql({
+        _id: group._id,
+        id: group.id,
+        name: group.name,
+        type: group.type,
+        privacy: group.privacy,
+        leader: groupLeader.id,
+      });
     });
   });
 
@@ -91,7 +98,14 @@ describe('GET /challenges/:challengeId', () => {
         id: groupLeader._id,
         profile: {name: groupLeader.profile.name},
       });
-      expect(chal.group).to.eql(_.pick(group, ['_id', 'id', 'name', 'type', 'privacy']));
+      expect(chal.group).to.eql({
+        _id: group._id,
+        id: group.id,
+        name: group.name,
+        type: group.type,
+        privacy: group.privacy,
+        leader: groupLeader.id,
+      });
     });
   });
 
@@ -136,7 +150,14 @@ describe('GET /challenges/:challengeId', () => {
         id: groupLeader.id,
         profile: {name: groupLeader.profile.name},
       });
-      expect(chal.group).to.eql(_.pick(group, ['_id', 'id', 'name', 'type', 'privacy']));
+      expect(chal.group).to.eql({
+        _id: group._id,
+        id: group.id,
+        name: group.name,
+        type: group.type,
+        privacy: group.privacy,
+        leader: groupLeader.id,
+      });
     });
   });
 });
--- a/test/api/v3/integration/challenges/GET-challenges_user.test.js
+++ b/test/api/v3/integration/challenges/GET-challenges_user.test.js
@@ -45,6 +45,7 @@ describe('GET challenges/user', () => {
         type: publicGuild.type,
         privacy: publicGuild.privacy,
         name: publicGuild.name,
+        leader: publicGuild.leader._id,
       });
     });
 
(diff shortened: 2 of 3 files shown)


crookedneighbor:
There should be test coverage added for this.

Ozzystrasza:
I'm sorry @crookedneighbor , but I don't really know what that is or how to do that. I will try to look it up and see if I can come up with something.

crookedneighbor:
No worries. If you look at the bottom of this PR, you'll see that the Travis build failed. That's a system that runs our automated test suite and the change you made caused it to fail because now some of those tests are getting back different results from that function you changed. If you inspect it, scroll down far enough and you'll see:

screen shot 2016-07-27 at 8 53 22 pm

That tells you that the [GET /challenges/:challengeid test](https://github.com/HabitRPG/habitrpg/blob/7c579bf850402ac43c146002e475b9baac4f66a2/test/api/v3/integration/challenges/GET-challenges_challengeId.test.js#L38-L51) failed because the result it got back from the request was different than the test expected (wasn't expecting to see the leader property).

Those tests that failed will need to be updated for this new behavior of getting the leader data. Does that make sense?

Ozzystrasza:
@crookedneighbor Yes, it actually makes perfect sense! I will work on it right away. If you could just clarify one thing for me. The test says to expect an object named leader with some properties, but the query is returning only a property named leader with the id as its value. 

```
  expect(chal.leader).to.eql({
        _id: groupLeader._id,
        id: groupLeader._id,
        profile: {name: groupLeader.profile.name},
      });
```

Should I change what is expected or change the query to return the whole object?

crookedneighbor:
The part that's failing is here: https://github.com/HabitRPG/habitrpg/blob/7c579bf850402ac43c146002e475b9baac4f66a2/test/api/v3/integration/challenges/GET-challenges_challengeId.test.js#L49

For that particular test, it should be as easy as adding `'leader'` to the array of properties to expect from the group.

crookedneighbor:
There are 5 other failing tests that will likely be similar. Ping me if you get stuck.

Ozzystrasza:
Yes, that's what I did, but this was the result:

```
 1) GET /challenges/:challengeId public guild should return challenge data:

      AssertionError: expected { Object (_id, leader, ...) } to deeply equal { Object (_id, id, ...) }
      + expected - actual

       {
         "_id": "f5f53ac8-345b-45cf-84c4-9ee5fcb112b9"
         "id": "f5f53ac8-345b-45cf-84c4-9ee5fcb112b9"
      -  "leader": "3c7c1ee9-2acc-4d51-b03b-1d4bfbda1739"
      +  "leader": {
      +    "_id": "3c7c1ee9-2acc-4d51-b03b-1d4bfbda1739"
      +    "profile": {
      +      "name": "cc022041-01a6-423e-aede-e65922fd46e6"
      +    }
      +  }
         "name": "test group"
         "privacy": "public"
         "type": "guild"
       }
```

crookedneighbor:
Ahh, the problem is that we do some magic to the leader property on groups and populate the actual data for the leader.

You'll have to rewrite that expectation to be something like:

``` js
expect(chal.group).to.eql({
  _id: group._id,
  id: group.id,
  name: group.name,
  type: group.type,
  privacy: group.privacy,
  leader: group.leader.id,
});
```

Ozzystrasza:
I think I got it all fixed now :smile:

paglias:
Looks good to me!

crookedneighbor:
Tier awarded!

Ozzystrasza:
Thanks!

PR #7844, opened 2016-07-29
Hide user paths via environment variable

SabreCat:
Fixes put_issue_url_here
### Changes

Introduces a new environment variable, USER_PRIVATE_PATHS, that allows server administrators to "hide" paths within the user object for the purposes of the API. This would support the ability to put a piece of content into production, but not display it to end users until all the users who should have the item have it--at which point the distribution can be announced.

Also includes partial build for Naming Day content, which will be the first use of the above feature.

---

UUID: [Hidden Pets Present](https://map.what3words.com/hidden.pets.present)


website/server/models/user/hooks.js
```diff
@@ -1,19 +1,38 @@
...
+    if (HIDE_LADDER_ITEMS !== '' && plainObj.auth) {
+      let hideLadderList = HIDE_LADDER_ITEMS.split(',');
+      let ladderHideArray;
+      let ladderHidePath;
+      let ladderHideDate;
+
+      for (let item of hideLadderList) {
+        ladderHideArray = item.split(':');
+        ladderHidePath = ladderHideArray[0];
+        ladderHideDate = new Date(ladderHideArray[1]);
```
paglias, line 29: we have to pay attention here and always supply strings that are interpreted correctly by `new Date()`

SabreCat, line 29: Definitely. Bringing `moment` in is also an option. Though we should probably have an escape clause for typo situations.

website/server/models/user/hooks.js
```diff
@@ -1,19 +1,38 @@
...
 schema.plugin(baseModel, {
   // noSet is not used as updating uses a whitelist and creating only accepts specific params (password, email, username, ...)
   noSet: [],
   private: ['auth.local.hashed_password', 'auth.local.salt', '_cronSignature'],
   toJSONTransform: function userToJSON (plainObj, originalDoc) {
     plainObj._tmp = originalDoc._tmp; // be sure to send down drop notifs
     delete plainObj.filters;
+    if (HIDE_LADDER_ITEMS !== '' && plainObj.auth) {
+      let hideLadderList = HIDE_LADDER_ITEMS.split(',');
+      let ladderHideArray;
```
khipkin, line 22: `let ladderHideArray, ladderHidePath, ladderHideDate;`

SabreCat, line 22: That's against our linting rules.

crookedneighbor:
I think we've decided not to go down this path. (Correct me if I'm mistaken)

PR #7845, opened 2016-07-30
 voluntarily leaving a party removes the user from the quest the party is involved in

Hus274:
part of #7653 
### Changes

Removes user from party.quest.members on leaving a party.

---

### DIFF
--- a/test/api/v3/unit/models/group.test.js
+++ b/test/api/v3/unit/models/group.test.js
@@ -442,13 +442,26 @@ describe('Group Model', () => {
           [nonParticipatingMember._id]: false,
           [undecidedMember._id]: null,
         };
+        party.memberCount = 4;
+        await party.save();
 
         await party.leave(participatingMember);
+
+        party = await Group.findOne({_id: party._id});
         expect(party.quest.members).to.eql({
           [questLeader._id]: true,
           [nonParticipatingMember._id]: false,
           [undecidedMember._id]: null,
         });
+      });
+
+      it('deletes group if user is only member', async () => {
+        party.memberCount = 1;
+
+        await party.leave(participatingMember);
+
+        party = await Group.findOne({_id: party._id});
+        expect(party).to.not.exist;
       });
     });
 
--- a/test/api/v3/unit/models/group.test.js
+++ b/test/api/v3/unit/models/group.test.js
@@ -455,13 +455,23 @@ describe('Group Model', () => {
         });
       });
 
-      it('deletes group if user is only member', async () => {
+      it('deletes a private group when the last member leaves', async () => {
         party.memberCount = 1;
 
         await party.leave(participatingMember);
 
         party = await Group.findOne({_id: party._id});
         expect(party).to.not.exist;
+      });
+
+      it('does not delete a public group when the last member leaves', async () => {
+        party.memberCount = 1;
+        party.privacy = 'public';
+
+        await party.leave(participatingMember);
+
+        party = await Group.findOne({_id: party._id});
+        expect(party).to.exist;
       });
     });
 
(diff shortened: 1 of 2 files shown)


website/server/models/group.js
```diff
@@ -724,6 +724,11 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
     pusher.trigger(`presence-group-${group._id}`, 'user-left', {
       userId: user._id,
     });
+
+    // remove member from quest
+    group.quest.members = _.omit(group.quest.members, user._id);
```
crookedneighbor: No need to use lodash here. Let's just do:

``` js
delete group.quest.members[user._id];
group.markModified('quest.members');
promises.push(group.save());
```

website/server/models/group.js
```diff
@@ -724,6 +724,11 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
     pusher.trigger(`presence-group-${group._id}`, 'user-left', {
       userId: user._id,
     });
+
+    // remove member from quest
```
crookedneighbor: This comment is not necessary. It's obvious from the code what is happening.

Hus274:
Makes sense, I've pushed a new revision

website/server/models/group.js
```diff
@@ -724,6 +724,10 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
     pusher.trigger(`presence-group-${group._id}`, 'user-left', {
       userId: user._id,
     });
+
+    delete group.quest.members[user._id];
+    group.markModified('quest.members');
+    promises.push(group.save());
```
crookedneighbor: It's a little weird to save the group and then possibly update or remove it below. I'm going to have to think on this a little.

Hus274: Yeah I would agree. What about moving the creation of the update object to above, appending the update.$set value here with the new quest.members value, and having the $set portion below just append to the update.$set object?

crookedneighbor: Best thing will probably to instantiate the update variable at the top of the function:

``` js
let update = {
  $inc: {memberCount: -1},
};
```

And then here, $unset the user id from the members object.

``` js
update.$unset = {[`quest.members.${user._id}`]: 1};
```

Then at the bottom, if the group is just removed, the update variable will get garbage collected and discarded. If it is updated, it'll include the call to unset the user from the quest.

Hus274: Makes sense, I'll update it here.

crookedneighbor: Beat me to it, but yes to what you said. Only use `$unset` instead of `$set`. We don't need to overwrite the whole `quest.members` object, just the one member.

Hus274:
Repushed using update and unset here.

test/api/v3/unit/models/group.test.js
```diff
@@ -434,6 +434,37 @@ describe('Group Model', () => {
...
+
+        party = await Group.findOne({_id: party._id});
+        expect(party.quest.members).to.eql({
+          [questLeader._id]: true,
+          [nonParticipatingMember._id]: false,
+          [undecidedMember._id]: null,
+        });
+      });
+
+      it('deletes group if user is only member', async () => {
```
crookedneighbor: This should specify private group. Public groups don't get deleted.

crookedneighbor: In fact, for bonus points you could add a test for it not deleting a public guild.

Hus274: sure, can do!

crookedneighbor:
I think this looks sound to me. If the build passes, I think it can be merged in.

Hus274:
added the other test and updated the it statement of the old one to specify the private group.

crookedneighbor:
Notated toward your next tier!

PR #7848, opened 2016-07-30
Adding tests to prove expected behavior around force starting a quest

Hus274:
part of #7653 
### Changes

Added tests to prove the following occurs as it is the expected behavior:
- check that the code for force-starting a quest removes any players from quest.members that don't have a true value
- check that the code for force-starting a quest removes any players from quest.members that aren't in the party

---

### DIFF
--- a/test/api/v3/integration/quests/POST-groups_groupId_quests_force-start.test.js
+++ b/test/api/v3/integration/quests/POST-groups_groupId_quests_force-start.test.js
@@ -149,5 +149,64 @@ describe('POST /groups/:groupId/quests/force-start', () => {
       expect(partyMemberThatIgnores.party.quest.key).to.not.exist;
       expect(partyMemberThatIgnores.party.quest.completed).to.not.exist;
     });
+
+    it('removes users who have not accepted the quest from quest.members', async () => {
+      let partyMemberThatRejects = partyMembers[1];
+      let partyMemberThatIgnores = partyMembers[2];
+
+      await leader.post(`/groups/${questingGroup._id}/quests/invite/${PET_QUEST}`);
+      await partyMembers[0].post(`/groups/${questingGroup._id}/quests/accept`);
+      await partyMemberThatRejects.post(`/groups/${questingGroup._id}/quests/reject`);
+
+      await leader.post(`/groups/${questingGroup._id}/quests/force-start`);
+
+      await Bluebird.delay(500);
+
+      await questingGroup.sync();
+
+      expect(questingGroup.quest.members[partyMemberThatRejects._id]).to.not.exist;
+      expect(questingGroup.quest.members[partyMemberThatIgnores._id]).to.not.exist;
+      expect(questingGroup.quest.members[partyMembers[0]._id]).to.exist;
+      expect(questingGroup.quest.members[leader._id]).to.exist;
+    });
+
+    it('removes users who are not in the party from quest.members', async () => {
+      await leader.post(`/groups/${questingGroup._id}/quests/invite/${PET_QUEST}`);
+
+      let notInPartyUser = await generateUser();
+      questingGroup.quest.members[notInPartyUser._id] = true;
+
+      expect(questingGroup.quest.members[notInPartyUser._id]).to.eql(true);
+
+      await leader.post(`/groups/${questingGroup._id}/quests/force-start`);
+
+      await Bluebird.delay(500);
+
+      await questingGroup.sync();
+
+      expect(questingGroup.quest.members[notInPartyUser._id]).to.not.exist;
+    });
+
+    it('removes users who don\'t have true value in quest.members from quest.members', async () => {
+      let partyMemberThatRejects = partyMembers[1];
+      let partyMemberThatIgnores = partyMembers[2];
+
+      await leader.post(`/groups/${questingGroup._id}/quests/invite/${PET_QUEST}`);
+      await partyMembers[0].post(`/groups/${questingGroup._id}/quests/accept`);
+
+      questingGroup.quest.members[partyMemberThatRejects._id] = false;
+      questingGroup.quest.members[partyMemberThatIgnores._id] = null;
+
+      await leader.post(`/groups/${questingGroup._id}/quests/force-start`);
+
+      await Bluebird.delay(500);
+
+      await questingGroup.sync();
+
+      expect(questingGroup.quest.members[partyMemberThatRejects._id]).to.not.exist;
+      expect(questingGroup.quest.members[partyMemberThatIgnores._id]).to.not.exist;
+      expect(questingGroup.quest.members[partyMembers[0]._id]).to.exist;
+      expect(questingGroup.quest.members[leader._id]).to.exist;
+    });
   });
 });
--- a/test/api/v3/integration/quests/POST-groups_groupId_quests_force-start.test.js
+++ b/test/api/v3/integration/quests/POST-groups_groupId_quests_force-start.test.js
@@ -172,9 +172,12 @@ describe('POST /groups/:groupId/quests/force-start', () => {
 
     it('removes users who are not in the party from quest.members', async () => {
       await leader.post(`/groups/${questingGroup._id}/quests/invite/${PET_QUEST}`);
-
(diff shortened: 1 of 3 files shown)


test/api/v3/integration/quests/POST-groups_groupId_quests_force-start.test.js
```diff
@@ -149,5 +149,64 @@ describe('POST /groups/:groupId/quests/force-start', () => {
...
+      expect(questingGroup.quest.members[partyMemberThatIgnores._id]).to.not.exist;
+      expect(questingGroup.quest.members[partyMembers[0]._id]).to.exist;
+      expect(questingGroup.quest.members[leader._id]).to.exist;
+    });
+
+    it('removes users who are not in the party from quest.members', async () => {
+      await leader.post(`/groups/${questingGroup._id}/quests/invite/${PET_QUEST}`);
+
+      let notInPartyUser = await generateUser();
+      questingGroup.quest.members[notInPartyUser._id] = true;
```
crookedneighbor: This is only changing the local variable in the test. It's not adding the id to the object in the db.

test/api/v3/integration/quests/POST-groups_groupId_quests_force-start.test.js
```diff
@@ -149,5 +149,64 @@ describe('POST /groups/:groupId/quests/force-start', () => {
...
+    });
+
+    it('removes users who don\'t have true value in quest.members from quest.members', async () => {
+      let partyMemberThatRejects = partyMembers[1];
+      let partyMemberThatIgnores = partyMembers[2];
+
+      await leader.post(`/groups/${questingGroup._id}/quests/invite/${PET_QUEST}`);
+      await partyMembers[0].post(`/groups/${questingGroup._id}/quests/accept`);
+
+      questingGroup.quest.members[partyMemberThatRejects._id] = false;
```
crookedneighbor: Same comment as above.

test/api/v3/integration/quests/POST-groups_groupId_quests_force-start.test.js
crookedneighbor: I want to say you should be able to do

``` js
await questingGroup.update({
  [`quest.members.${notInPartyUset._id}`]: true,
});
```

crookedneighbor: ^ does a direct db update

Hus274: I was trying to do this earlier but couldn't get it to work and then wasn't sure I needed it. It seems like I was missing the array brackets [ ] around the quest members thing in the update. Why are those needed? I thought you just needed the string defining the path

Hus274:
Fixed up the tests so they actually talk to the database correctly and added functionality to remove any users who aren't in the party from the party's quest.

website/server/models/group.js
```diff
@@ -366,6 +366,17 @@ schema.methods.startQuest = async function startQuest (user) {
   let nonUserQuestMembers = _.keys(this.quest.members);
   removeFromArray(nonUserQuestMembers, user._id);
 
+  // remove any users from quest.members who aren't in the party
+  let partyId = this._id;
+  let questMembers = this.quest.members;
+  await Bluebird.map(Object.keys(this.quest.members), async (memberId) => {
+    let member = await User.findById(memberId, 'party._id').exec();
```
crookedneighbor: I'd do: 

``` js
let member = await User.findOne({_id: memberId, 'party.id': partyId});

if (!member) {
  delete questMembers[memberId];
}
```

This will catch users that don't exist.

crookedneighbor: In fact, let's add a test for that case. The case where a user deleted their account and it wasn't cleaned up properly.

Hus274:
added code to handle the deleted user and a test around that case.

Hus274:
the merge fixed the tests.

website/server/models/group.js
```diff
@@ -366,6 +366,17 @@ schema.methods.startQuest = async function startQuest (user) {
   let nonUserQuestMembers = _.keys(this.quest.members);
   removeFromArray(nonUserQuestMembers, user._id);
 
+  // remove any users from quest.members who aren't in the party
+  let partyId = this._id;
+  let questMembers = this.quest.members;
+  await Bluebird.map(Object.keys(this.quest.members), async (memberId) => {
+    let member = await User.findOne({_id: memberId, 'party._id': partyId});
```
paglias: this way the entire user will be fetched. I'd do `User.findOne({_id: memberId, 'party._id': partyId}).select('_id').lean()` so only the _id will be returned and using `.lean()` will prevent a mongoose object from being build instead returning a plain js object which is much faster

Hus274: Makes sense to me, I'll update this and resubmit

test/api/v3/integration/quests/POST-groups_groupId_quests_force-start.test.js
```diff
@@ -149,5 +149,86 @@ describe('POST /groups/:groupId/quests/force-start', () => {
       expect(partyMemberThatIgnores.party.quest.key).to.not.exist;
       expect(partyMemberThatIgnores.party.quest.completed).to.not.exist;
     });
+
+    it('removes users who have not accepted the quest from quest.members', async () => {
+      let partyMemberThatRejects = partyMembers[1];
+      let partyMemberThatIgnores = partyMembers[2];
+
+      await leader.post(`/groups/${questingGroup._id}/quests/invite/${PET_QUEST}`);
+      await partyMembers[0].post(`/groups/${questingGroup._id}/quests/accept`);
```
khipkin: `partyMemberThatAccepts` for consistency and clarity

test/api/v3/integration/quests/POST-groups_groupId_quests_force-start.test.js
```diff
@@ -149,5 +149,86 @@ describe('POST /groups/:groupId/quests/force-start', () => {
...
+      await partyMemberThatRejects.post(`/groups/${questingGroup._id}/quests/reject`);
+
+      await leader.post(`/groups/${questingGroup._id}/quests/force-start`);
+
+      await Bluebird.delay(500);
```
khipkin, line 164: Why is this necessary?

test/api/v3/integration/quests/POST-groups_groupId_quests_force-start.test.js
```diff
@@ -149,5 +149,86 @@ describe('POST /groups/:groupId/quests/force-start', () => {
...
+    });
+
+    it('removes users who are not in the party from quest.members', async () => {
+      await leader.post(`/groups/${questingGroup._id}/quests/invite/${PET_QUEST}`);
+      let notInPartyUser = await generateUser();
+
+      await questingGroup.update({
+        [`quest.members.${notInPartyUser._id}`]: true,
+      });
+      questingGroup.sync();
```
khipkin: await?

test/api/v3/integration/quests/POST-groups_groupId_quests_force-start.test.js
Hus274, line 164: If you look at the force start code it's not all synchronous, it kicks things off purposefully without waiting for them. If you look at the other tests in this file, the majority of them also have this.

Hus274:
addressed comments

website/server/models/group.js
```diff
@@ -366,6 +366,17 @@ schema.methods.startQuest = async function startQuest (user) {
   let nonUserQuestMembers = _.keys(this.quest.members);
   removeFromArray(nonUserQuestMembers, user._id);
 
+  // remove any users from quest.members who aren't in the party
+  let partyId = this._id;
+  let questMembers = this.quest.members;
+  await Bluebird.map(Object.keys(this.quest.members), async (memberId) => {
```
crookedneighbor, line 372: Since you've created the `questMembers` variable, we should probably use it here.

crookedneighbor: Never mind, it's fine.

crookedneighbor:

PR #7849, opened 2016-07-30
booting a member from party removes member from quest

Hus274:
part of #7653 
### Changes

Adds test around expected behavior of quests when booting a user from a party when the party is on an active quest or has a quest invite pending.

---

### DIFF
--- a/website/server/controllers/api-v3/groups.js
+++ b/website/server/controllers/api-v3/groups.js
@@ -436,7 +436,7 @@ api.removeGroupMember = {
         group.quest.leader = undefined;
       } else if (group.quest && group.quest.members) {
         // remove member from quest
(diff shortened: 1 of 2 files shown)


website/server/controllers/api-v3/groups.js
```diff
@@ -436,7 +436,7 @@ api.removeGroupMember = {
         group.quest.leader = undefined;
       } else if (group.quest && group.quest.members) {
         // remove member from quest
-        group.quest.members[member._id] = undefined;
+        delete group.quest.members[member._id];
```
crookedneighbor, line 439: These should be roughly equivalent, but I do prefer the delete syntax.

crookedneighbor:
This looks great to me.

crookedneighbor:

PR #7850, opened 2016-07-30
Don't show items in the seasonal shop that can't be purchased

### LINKED ISSUE #7843: Seasonal Shop shouldn't display "broken" equipment
General Info
- UUID: 7f14ed62-5408-4e1b-be83-ada62d504931
- Browser: Vivaldi
- OS: Windows 10
### Description

When you purchase a gear item from the Seasonal Shop, it disappears. If you lose a current Seasonal Shop gear item to zeroing out your Health or to using the Orb of Rebirth, though, the item then appears on the page, faded out. You can click it, and be informed that you already own the item and need to buy it with Gold instead of Gems.

This is confusing. If you can't buy a thing from the Shop, it shouldn't show up on the page at all.

Some things to keep in mind while fixing this:
- Currently we have a banner that displays in the event that you've used an Orb of Rebirth, which explains how to get those items back. That may need to be modified to show up if you have any "broken" Seasonal Shop items, even if you don't have any Rebirth achievements.
- This is a separate concern from the possibility of not removing these items during a Rebirth at all. That can be changed, and this fixed, independently.
#### Console Errors

None

Hus274:
Fixes #7843 
### Changes

Stops showing transparent items in the seasonal shop and always shows the banner that describes how to repurchase season items. Reads:

```
"You can repurchase Seasonal Items that you've lost in the Rewards Column. You'll only be able to purchase the items for your current class (Warrior by default), but fear not, the other class-specific items will become available if you switch to that class."
```

---


crookedneighbor:
This should be an API change so the mobile apps also show the right equipment.

Hus274:
I'm not quite sure how to make this a API change because right now it looks like the api is returning all the seasonal shop items and the ui filters out ones that the user owns or has owned which is strictly a display layer change. Changing the API to only return unowned items would be a pretty big change and I'm not sure about the blast radius. 

Is that what I should do here, change the api to only return unowned items for the seasonal shop?

crookedneighbor:
The shop API is new and still in beta. 

@vIiRuS was it your intention to display items that had been previously owned but were broken do to death? I assumed not and this was an oversight.

phillipthelen:
No, that wasn't my intention. The API shouldn't return the items in that case, since they are already made available in the rewards column. The issue is probably this line [here](https://github.com/HabitRPG/habitrpg/blob/develop/common/script/libs/shops.js#L220), that does the wrong comparison.

Hus274:
Awesome, I can fix this in the API then and remove the unnecessary filtering code out of the website's UI then which definitely seems better. I'll fix this up and recommit!

Hus274:
Cleaned up the git tree by deleting this remote branch and creating a new one. The new pr is referenced above and makes this an api change.

PR #7852, opened 2016-07-31
Clear invites from user when the request fails because the invite is invalid

Hus274:
part of #7653 
### Changes

This should clear a user invite to a quest when the user tries to accept or decline the quest when the invite has gotten into a bad or invalid state. Should help prevent the need for admin intervention.

---

### DIFF
--- a/website/server/controllers/api-v3/quests.js
+++ b/website/server/controllers/api-v3/quests.js
@@ -162,7 +162,7 @@ api.acceptQuest = {
     if (validationErrors) throw validationErrors;
 
     user.party.quest.RSVPNeeded = false;
-    user.save();
+    await user.save();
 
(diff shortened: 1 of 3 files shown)


website/server/controllers/api-v3/quests.js
```diff
@@ -161,6 +161,9 @@ api.acceptQuest = {
     let validationErrors = req.validationErrors();
     if (validationErrors) throw validationErrors;
 
+    user.party.quest.RSVPNeeded = false;
+    user.save();
```
paglias: await?

Hus274: good catch. I'll fix this

Hus274:
Added missing await statement.

crookedneighbor:

PR #7858, opened 2016-08-01
Challenges: Assigning Location as Tavern only sets prize to 1 if 0 or empty

Ozzystrasza:
Fixes https://github.com/HabitRPG/habitrpg/issues/3998
### Changes

Added a test to check the current value of `$scope.newChallenge.prize` so it doesn't always change to 1 when you choose tavern as location for the challenge. 
Now changing the location to Tavern will only change `$scope.newChallenge.prize` to 1 if the current value is `null`, `undefined`, 0, or negative.

Tests:

---


crookedneighbor:
Thanks! Notated toward your next tier.

Ozzystrasza:
Thank you!

PR #7859, opened 2016-08-02
Don't show items in the seasonal shop that can't be purchased 

### LINKED ISSUE #7843: Seasonal Shop shouldn't display "broken" equipment

Hus274:
Fixes #7843 
### Changes

Stops showing transparent items in the seasonal shop and always shows the banner that describes how to repurchase season items. Reads:

```
"You can repurchase Seasonal Items that you've lost in the Rewards Column. You'll only be able to purchase the items for your current class (Warrior by default), but fear not, the other class-specific items will become available if you switch to that class."
```

recreated from: #7850 
Updated to be an api change so this would be fixed on the mobile app as well.

---


website/views/options/inventory/seasonal-shop.jade
```diff
@@ -22,9 +22,7 @@
         p {{::Content.spells.special.seafoam.value}}
           span(class='shop_gold')
     menu.pets-menu(label='{{category.text}}', ng-repeat='category in seasonalShopCategories')
-      div(ng-repeat='item in category.items',
-        ng-class="{transparent: user.items.gear.owned[item.key] !== undefined}",
```
crookedneighbor, line 26: I think this will make the item remain on the page. Keeping the `ng-if` should suffice to make it disappear.

Hus274: makes sense, good catch

Hus274:
Fixed to not show already owned items in the seasonal shop.

PR #7865, opened 2016-08-03
Improve server side analytics

phillipthelen:
### Changes

This PR changes two things.
1. Added the following information to server side amplitude events:
   - set platform to "Web", "iOS", "Android" or "3rd Party" depending on the client that sent the request
   - Set "OS" to either the browser or the mobile OS
2. Moved the following events from the mobile client to the server
   - login
   - Creating a guild/party
   - joining a guild/party

### DIFF
--- a/common/script/ops/rebirth.js
+++ b/common/script/ops/rebirth.js
@@ -30,8 +30,7 @@ module.exports = function rebirth (user, tasks = [], req = {}, analytics) {
   }
 
   if (analytics) {
-    analyticsData.client = req.headers['x-client'];
-    analyticsData.useragent = req.headers['user-agent'];
+    analyticsData.headers = req.headers;
     analytics.track('Rebirth', analyticsData);
   }
 
--- a/common/script/ops/rebirth.js
+++ b/common/script/ops/rebirth.js
@@ -30,6 +30,9 @@ module.exports = function rebirth (user, tasks = [], req = {}, analytics) {
   }
 
   if (analytics) {
+
+    analyticsData.client = req.headers['x-client'];
+    analyticsData.useragent = req.headers['user-agent'];
     analytics.track('Rebirth', analyticsData);
   }
 
--- a/common/script/ops/rebirth.js
+++ b/common/script/ops/rebirth.js
@@ -30,7 +30,6 @@ module.exports = function rebirth (user, tasks = [], req = {}, analytics) {
   }
 
   if (analytics) {
-
     analyticsData.client = req.headers['x-client'];
     analyticsData.useragent = req.headers['user-agent'];
     analytics.track('Rebirth', analyticsData);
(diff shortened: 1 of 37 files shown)


crookedneighbor:
Since the headers are passed into every session, would it make more sense to pass the `req.headers` in and let the analytics service pull out the data?

paglias:
I agree with @crookedneighbor 's comment

common/script/ops/rebirth.js
```diff
@@ -30,6 +30,8 @@ module.exports = function rebirth (user, tasks = [], req = {}, analytics) {
   }
 
   if (analytics) {
+    analyticsData.client = req.headers['x-client'];
```
crookedneighbor: Any reason this isn't just passing in headers?

website/server/controllers/api-v3/groups.js
```diff
@@ -66,6 +66,30 @@ api.createGroup = {
       _id: user._id,
       profile: {name: user.profile.name},
     };
+
+    if (savedGroup.privacy === 'public') {
```
crookedneighbor, line 80: How about 

``` js
let analyticsObject = {
  uuid: user._id,
  hitType: 'event',
  category: 'behavior',
  owner: true,
  groupType: savedGroup.type,
  privacy: savedGroup.privacy,
  headers: req.headers
};

if (savedGroup.privacy === 'public') {
  analyticsObject.groupName = savedGroup.name;
}

res.analytics.track('join group', analyticsObject);
```

website/server/controllers/api-v3/groups.js
```diff
@@ -273,6 +297,30 @@ api.joinGroup = {
     if (leader) {
       response.leader = leader.toJSON({minimize: true});
     }
+
+    if (group.privacy === 'public') {
```
crookedneighbor, line 304: Same here.

website/server/libs/analyticsService.js
```diff
@@ -81,13 +82,45 @@ let _formatUserData = (user) => {
   return properties;
 };
 
+let _formatPlatformForAmplitude = (platform) => {
+  switch (platform) {
```
crookedneighbor: I'd prefer an if/else if/else logic here. Or a map:

``` js
const PLATFORM_MAP = Object.freeze({
  'habitica-web': 'Web',
  'habitica-ios': 'iOS',
  'habitica-android': 'Android',
});

function _formatPlatformForAmplitude (platform) {
  if (platform in PLATFORM_MAP) {
    return PLATFORM_MAP[platform];
  }

  return platform;
}

```

website/server/libs/analyticsService.js
```diff
@@ -100,10 +133,10 @@ let _formatDataForAmplitude = (data) => {
   if (itemName) {
     event_properties.itemName = itemName;
   }
-
   return ampData;
 };
 
+
```
crookedneighbor: Extra line break here.

phillipthelen:
made the requested changes

PR #7870, opened 2016-08-04
API: return computed stats for members routes

### LINKED ISSUE #7863: /members/{uid} not returning calculated stats like maxHealth / maxMP / toNextLevel
this would be useful for API Users

as @paglias @crookedneighbor  pointed out this could be done using virtual properties

paglias:
Fixes #7863 

Tests are missing

### DIFF
--- a/test/api/v3/integration/groups/GET-groups_groupId_members.test.js
+++ b/test/api/v3/integration/groups/GET-groups_groupId_members.test.js
@@ -66,7 +66,7 @@ describe('GET /groups/:groupId/members', () => {
     expect(res[0].profile).to.have.all.keys(['name']);
   });
 
-  it('populates only some fields if req.query.includeAllPublicFields === true and it is a guild', async () => {
+  it('req.query.includeAllPublicFields === true only works with parties', async () => {
     let group = await generateGroup(user, {type: 'guild', name: generateUUID()});
     let res = await user.get(`/groups/${group._id}/members`);
     expect(res[0]).to.have.all.keys(['_id', 'id', 'profile']);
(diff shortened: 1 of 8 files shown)


paglias:
Tests added, this is ready to be merged

test/api/v3/integration/groups/GET-groups_groupId_members.test.js
```diff
@@ -57,6 +66,30 @@ describe('GET /groups/:groupId/members', () => {
     expect(res[0].profile).to.have.all.keys(['name']);
   });
 
+  it('populates only some fields if req.query.includeAllPublicFields === true and it is a guild', async () => {
```
crookedneighbor: `req.query.includeAllPublicFields === true` -> `req.query.includeAllPublicFields !== true`

paglias: it shows that req.query.includeAllPublicFields === true doesn't work if it's not a party

paglias: I'll change it because it's not clear

crookedneighbor: If that's the case, you'll need to add the query in your test case:

``` js
let res = await user.get(`/groups/${group._id}/members`);
```

Does not include the query parameter.

paglias: Ops... Forgot about it!

crookedneighbor:
👍

SabreCat:
Tests aren't passing yet, though?

paglias:
@SabreCat fixing it, the new tests are causing another one to fail

PR #7872, opened 2016-08-04
Fix #6708 - guild message maximizes width in Firefox

### LINKED ISSUE #6708: The bubble of the leader's opening post in chat doesn't stretch horizontally
If the bubble of the opening post by the leader in the guild/party chat is meant to stretch horizontally, it doesn't seem to do so for me but rather has a fixed width.

I'm Using Firefox 44.0.2 for Windows (Win10 Home). The console shows a critical warning about loading mixed content via secured connection for
 view-source:https://habitica.com/app-c14655c7.js
 but I can't judge if it's related.

Screenshots:
Pic1
Pic2
Pic3

Reporting this as suggested in the Aspiring Coders guild and not having found related threads.

To quote @DrStrangepork from similar issue:

> The https://habitica.com/app-bf28709d.css line 5 says "table {max-width: 100%;}" but in my Firefox browser the table always renders at width 194. I copied the markdown of the leader message from Aspiring Linguists to my local install and was able to duplicate the problem there as well: width is rendered to some hard number relative to zoom level in Firefox, but in Chrome it renders to 100% width (up to 1095px, at which point it remains at 1095px even if the zoom level is so small that it could be wider).

ESWAT:
Fixes https://github.com/HabitRPG/habitrpg/issues/6708
### Changes

Guild messages, or anything that uses `.wide-popover` class really, won’t stretch to let its contents spread naturally in Firefox, due to the `max-content` value still being experimental in that browser.

This fix adds a `-moz` browser-prefix for that value so it works in Firefox and browsers that support it without a browser-prefix.

Firefox 47.0.1 screenshot:
screen shot 2016-08-04 at 18 20 56

---

### DIFF
--- a/website/client/css/alerts.styl
+++ b/website/client/css/alerts.styl
@@ -14,6 +14,7 @@
   display: inline-block
 
 .wide-popover
+  max-width: -moz-max-content
   max-width: max-content
 
 // Small alerts


audiolion:
Interesting. Shouldnt we be using a css preprocessor that adds in browser related prefixes for all applicable css?

ESWAT:
That would be the ideal, yes, assuming we use [autoprefixer](https://github.com/postcss/autoprefixer) or the [Stylus equivalent](https://github.com/jescalan/autoprefixer-stylus) since they choose whether to apply browser-prefixes based on an updated source, not just hard-coded choices like some of the other preprocessors for Stylus.

Either way, since the app doesn’t have such a preprocessor right now and there are other occurrences where we hard-code browser-prefixes, I think this PR is good to merge if the admins are ready. Adding non-manual prefixing should be done in another branch/PR.

crookedneighbor:
We're also on the brink of a redesign and we'll likely switch to Sass, so I'm hesitant to add more stylus extensions.

crookedneighbor:

PR #7877, opened 2016-08-06
feat: allowing modification of challenge tasks after challenge creation

### LINKED ISSUE #7320: [API v3] Challenges can not be edited by their owner or an admin
### Description

Most challenge attributes cannot be edited by their owner or by admins. 

Note that I'm not referring to editing the tasks that have been added to the player's own task lists. I'm talking about the challenge itself.

When a challenge owner or admin tries to edit the challenge:
- They ARE able to change the title (this is correct).
- They seem to be able to change the tag but on a reload, the old tag name is in place.
- They are not able to edit the titles of the tasks - it's greyed out.
- They are not able to change the difficulty - the buttons don't respond to clicks.
#### Console Errors

none

Hus274:
Fixes #7320 
### Changes

This change allows challenge owners to modify challenge tasks after the initial creation of the challenge. 

Allowed Changes:
- Edit the title of a task
- Edit the extra notes (NB: the participants can also edit)
- Change the difficulty
- Edit the Habit's negative and positive buttons
- Edit the repeat fields and start date for a Daily
- Edit the due date for a To-Do
- Add and delete tasks

Notes:
- Removing a task from a challenge doesn't remove the task from the user, only results in the user task having a broken challenge link and can be removed that way. This isn't a change in behavior.
- Changes (except for notes) will be propagated to all users of that challenge.
- Changes to the notes of a challenge won't effect anyone who's already in the challenge, only users who join the challenge after the update will have the new notes.
- I've remove the ability to open a task edit window for challenge tasks when just viewing challenges by double clicking. You can replicate this behavior in prod by going to challenges, expanding one, and double clicking a task. This functionality will be removed by this pull request.

---

### DIFF
--- a/test/api/v3/unit/models/challenge.test.js
+++ b/test/api/v3/unit/models/challenge.test.js
@@ -158,6 +158,7 @@ describe('Challenge Model', () => {
       });
     });
   });
+
   context('type specific updates', () => {
     it('updates habit specific field to challenge and challenge members', async () => {
       task = new Tasks.habit(Tasks.Task.sanitize(tasksToTest.habit)); // eslint-disable-line babel/new-cap
@@ -195,7 +196,7 @@ describe('Challenge Model', () => {
     });
 
     it('updates daily specific field to challenge and challenge members', async () => {
-      task = new Tasks.daily(Tasks.Task.sanitize(tasksToTest.todo)); // eslint-disable-line babel/new-cap
+      task = new Tasks.daily(Tasks.Task.sanitize(tasksToTest.daily)); // eslint-disable-line babel/new-cap
       task.challenge.id = challenge._id;
       await task.save();
 
--- a/test/api/v3/unit/models/challenge.test.js
+++ b/test/api/v3/unit/models/challenge.test.js
@@ -158,4 +158,56 @@ describe('Challenge Model', () => {
       });
     });
   });
+  context('type specific updates', () => {
+    it('updates habit specific field to challenge and challenge members', async () => {
+      task = new Tasks.habit(Tasks.Task.sanitize(tasksToTest.habit)); // eslint-disable-line babel/new-cap
+      task.challenge.id = challenge._id;
+      await task.save();
+
+      await challenge.addTasks([task]);
+
+      task.up = true;
+      task.down = false;
+
+      await challenge.updateTask(task);
+
+      let updatedLeader = await User.findOne({_id: leader._id});
+      let updatedUserTask = await Tasks.Task.findById(updatedLeader.tasksOrder.habits[0]);
+
+      expect(updatedUserTask.up).to.equal(true);
+      expect(updatedUserTask.down).to.equal(false);
+    });
+
+    it('updates todo specific field to challenge and challenge members', async () => {
+      task = new Tasks.todo(Tasks.Task.sanitize(tasksToTest.todo)); // eslint-disable-line babel/new-cap
+      task.challenge.id = challenge._id;
+      await task.save();
+
+      await challenge.addTasks([task]);
+
+      task.date = new Date();
+      await challenge.updateTask(task);
+
+      let updatedLeader = await User.findOne({_id: leader._id});
+      let updatedUserTask = await Tasks.Task.findById(updatedLeader.tasksOrder.todos[0]);
+
+      expect(updatedUserTask.date).to.exist;
+    });
+
+    it('updates daily specific field to challenge and challenge members', async () => {
+      task = new Tasks.daily(Tasks.Task.sanitize(tasksToTest.todo)); // eslint-disable-line babel/new-cap
+      task.challenge.id = challenge._id;
+      await task.save();
+
+      await challenge.addTasks([task]);
+
+      task.everyX = 2;
+      await challenge.updateTask(task);
+
+      let updatedLeader = await User.findOne({_id: leader._id});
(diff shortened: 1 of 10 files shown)


website/server/models/challenge.js
```diff
@@ -209,8 +209,22 @@ schema.methods.updateTask = async function challengeUpdateTask (task) {
     updateCmd.$set[key] = syncableAttrs[key];
   }
```
crookedneighbor, line 211: Why not

``` js
let taskSchema = Tasks[task.type] ? Tasks[task.type] : Tasks.Task;
```

Hus274, line 211: That's so much better, I didn't like this and then didn't think better of fixing it after I saw how the tests did it...

crookedneighbor:
We don't have linting for the client code yet, but our style is to not put a space between the `!` and the variable.

website/views/shared/tasks/task_view/index.jade
```diff
@@ -37,7 +37,8 @@
     label(for='box-{{::obj._id}}_{{::task._id}}')
 
 // main content
-.task-text(ng-dblclick='task._editing ? saveTask(task) : editTask(task)')
+// don't allow a double click to edit if the object is locked
+.task-text(ng-dblclick='obj._locked ? {} : task._editing ? saveTask(task) : editTask(task, obj)')
```
crookedneighbor: Double ternaries make me sad. Extract this into a controller function.

Hus274: So that's what I tried to do initially but I don't know what the obj is and the controller doesn't appear to have access to it somehow? The method signature is already (task, user) and we're only sending task so adding obj didn't seem ideal. Do you know what this obj is?

crookedneighbor: For challenges obj is defined here: https://github.com/HabitRPG/habitrpg/blob/develop/website/views/options/social/challenges.jade#L54

For users, obj is defined here:
https://github.com/HabitRPG/habitrpg/blob/develop/website/views/main/index.jade#L4

So I'd create a function in the challenges controller and the tasks controller (it's okay that we have duplication for it, we're about to do a rewrite to the frontend code which will make these much more component based).

``` js
$scope.doubleClickTask = function (obj, task) {
  if (obj._locked) {
    return false;
  }

  if (task._editing) {
    $scope.saveTask(task);
  } else {
    $scope.editTask(task);
  }
}
```

Then in the view:

``` jade
.task-text(ng-dblclick='doubleClickTask(obj, task)')
```

test/api/v3/unit/models/challenge.test.js
```diff
@@ -158,4 +158,56 @@ describe('Challenge Model', () => {
...
+      await challenge.updateTask(task);
+
+      let updatedLeader = await User.findOne({_id: leader._id});
+      let updatedUserTask = await Tasks.Task.findById(updatedLeader.tasksOrder.todos[0]);
+
+      expect(updatedUserTask.date).to.exist;
+    });
+
+    it('updates daily specific field to challenge and challenge members', async () => {
+      task = new Tasks.daily(Tasks.Task.sanitize(tasksToTest.todo)); // eslint-disable-line babel/new-cap
```
crookedneighbor: Was this supposed to be `tasksToTest.daily`?

test/api/v3/unit/models/challenge.test.js
```diff
@@ -158,4 +158,56 @@ describe('Challenge Model', () => {
       });
     });
   });
```
crookedneighbor, line 160: line break here.

test/api/v3/unit/models/challenge.test.js
Hus274: yep, good catch. fixed.

website/views/shared/tasks/task_view/index.jade
Hus274: This is much cleaner, thanks. I don't know why it didn't cross my mind to make a new method as a pass through to only effect this call. Apparently I'm off my game today

website/server/models/challenge.js
paglias, line 211: Should it ever happen that `Tasks[task.type]` is undefined?

Hus274, line 211: Yeah that's a good point, it shouldn't ever be undefined and if it is, it's better to fail to save so we can investigate instead of just not saving the particular fields. I would agree that it's probably better to fail fast than to fail silently

Hus274, line 211: @crookedneighbor do you agree this is best?

crookedneighbor, line 211: Agreed.

Hus274:
updated according to paglias' comment

khipkin:
@Hus274 Don't forget to remove the spaces after your !s.

Hus274:
fixed the formatting

paglias:
Looks good to me 👍

crookedneighbor:
Notated toward your next tier. Thank you so much for this!

ShulamitLH:
This still hasn't been fixed. I just tried editing an existing challenge that I created and was unable to change name or difficulty of any task within the challenge. I tried this with one I created long ago and also a test challenge I created this morning. This is true for to-do tasks. I haven't tried dailies or habits.

Alys:
@ShulamitLH  My comment at https://github.com/HabitRPG/habitrpg/issues/7320#issuecomment-239669635 applies to this too.

PR #7878, opened 2016-08-07
fix: allows user to leave party/guild that doesn't exist

### LINKED ISSUE #7724: cannot leave a party that no longer exists
If a player's account contains a `party._id` for a party that no longer exists, they are unable to leave the party, or create a new one, or be invited to a new one. I found a case of this today - confirmed that the `party._id` in their account did  not correspond to any existing group.

I don't know how this situation happened, but I imagine that any action to leave or be removed from a party could time out, resulting in the ID not being deleted. Also it could arise from this bug: https://github.com/HabitRPG/habitrpg/issues/7723

This situation triggers the `groupNotFound` error ("Group not found or you don't have access."). 

I think it would be good if the code that finds that error in this particular situation would delete the party ID from the player's account, ideally without reporting an error to the user. I'm going to wait a day for feedback about this, and if there's no complaints or better ideas, I'll mark it as help welcome for that solution.

Hus274:
Fixes #7724 
### Changes

This change allows a user to leave a party or group that doesn't exist to avoid the poison pill situation that currently exists.

---

### DIFF
--- a/website/server/controllers/api-v3/groups.js
+++ b/website/server/controllers/api-v3/groups.js
@@ -348,7 +348,13 @@ api.leaveGroup = {
     if (validationErrors) throw validationErrors;
 
     let group = await Group.getGroup({user, groupId: req.params.groupId, fields: '-chat', requireMembership: true});
-    if (!group) throw new NotFound(res.t('groupNotFound'));
+    if (!group) {
+      delete user.party._id;
+      removeFromArray(user.guilds, req.params.groupId);
+      await user.save();
+
+      throw new NotFound(res.t('groupNotFound'));
+    }
 
     // During quests, checke wheter user can leave
     if (group.type === 'party') {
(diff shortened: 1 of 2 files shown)


website/server/controllers/api-v3/groups.js
```diff
@@ -348,7 +348,13 @@ api.leaveGroup = {
     if (validationErrors) throw validationErrors;
 
     let group = await Group.getGroup({user, groupId: req.params.groupId, fields: '-chat', requireMembership: true});
-    if (!group) throw new NotFound(res.t('groupNotFound'));
+    if (!group) {
+      delete user.party._id;
```
crookedneighbor: Seems to me like this would remove a user from their party of the guild they tried to leave didn't exist. This should do a check first to see if the party id equals the id in question. If it does, delete it. If it doesn't, remove it from the guilds array.

Hus274: great catch!

crookedneighbor:
I moved the logic you wrote into the Groups.getGroup method https://github.com/HabitRPG/habitrpg/commit/bba56b8c29caee7a22091944fae75ac1a817819c

That way we'll remove the non-existant group any time the getGroup call is made with a group that doesn't exist. 

This way, when a user loads the site, if they have a non-existant party, it'll give them the group not found error once, but it'll delete the group from the user.

crookedneighbor:

PR #7484, opened 2016-05-24
Webhooks fixes

paglias:
WIP Missing tests and new i18n strings

PR #7879, opened 2016-08-07
Webhook improvements

crookedneighbor:
Currently we have one type of webhook, which is fired when a task is scored. This PR adds two additional webhooks. One that is fired when a task is created and one that is fired when a chat is received in a group.

This allows for better integration with services like Zapier (so we don't have to resort to a polling method which could get costly for us).

Includes validation improvements from #7484

I put the webhooks ops into a single file since there was some shared validation stuff. This breaks the pattern for how ops stuff has been set up, but I do not think it is a problem.

I moved the webhook routes to their own file and apidoc group and flushed out the documentation more than we've done with the other routes. I'd like to see us do the same with the other routes in time.
### Consolidate task webhooks?

Instead of a `taskCreated` and a `taskScored` webhook, should we have a `taskActivity` webhook that fires when the task is created, updated, scored, or deleted? (You could have options that tell it to ignore specific actions). This would make it easier sync services to stay up to date. For instance, if you had an integration with Todoist, this would make it much easier to keep everything in sync. 

The migration file would need to be changed to update all existing webhooks to include the options object:

``` js
options: {
  created: false,
  updated: false,
  scored: true,
  deleted: false
}
```

I think I like this a lot better, but would love to hear some thoughts about it before implementing. 
### Allow multiple group ids in groupChatReceieved Webhooks?

Currently I have the webhook set to require an `options.groupId` parameter. Alternatively, we could have options object look like this:

``` js
options: {
  allGroups: true,
  groupIds: {
    'id-here': true,
    'another-id-here': false
  }
}
```

If `allGroups` is true, then it sends out the webhook for any group you are a part of. If it is false, then it would check to see that the group id is truthy in the `groupIds` object. 

This has the benefit of only requiring one webhook configured to send out data from multiple groups, but I think it probably makes more sense to require a webhook to be set up for each group independently. 

In addition, this only fires when a user sends a chat message. A notification is not fired when a system message happens. This can be added in the future as a configuration option if desired. Something like:

``` js
includeSystemMessages: true
```
## For the Future

I'd like to add a questActivity webhook (fired off at similar times to the push notifications), but figured that could wait.


crookedneighbor:
Note: I'm intentionally leaving the beta tag on and not providing an interface for creating other types of webhooks on the client. This is so that we can try these things out internally first (and through a beta test with Zapier) before providing a way for users to add webhooks themselves. (Obviously they can still do it through the API)

Once we've solidified how we want to do this, we can remove the beta tag and add an interface for adding additional webhook types.

khipkin:
Configurable, consolidated task webhook: :+1: 

Separate webhooks for each group (with future `includeSystemMessages` option): :+1:

website/server/libs/webhook.js
```diff
@@ -1,31 +1,112 @@
...
+    });
+  }
+}
+
+export let taskScoredWebhook = new WebhookSender({
+  type: 'taskScored',
+  transformData (data) {
+    let { user, task, direction, delta } = data;
+
+    let extendedStats = Object.assign({
```
paglias: we should merge this code with https://github.com/HabitRPG/habitrpg/blob/develop/website/server/models/user/methods.js#L47

common/script/ops/webhooks.js
```diff
@@ -0,0 +1,143 @@
...
+    enabled,
+    type: body.type || DEFAULT_WEBHOOK_TYPE,
+    options: body.options,
+  }, req);
+
+  user.markModified('preferences.webhooks');
+
+  webhook = refPush(webhooks, webhook);
+
+  if (req.v2 === true) {
```
crookedneighbor: Note to self. Remove these v2 logic blocks.

website/server/libs/webhook.js
crookedneighbor: Agreed. Do you think it better to convert the whole user object to json or modify the method to take a stats object instead?

paglias: I don't love either but from a pure perf point of view it's better to change the model's method to take a stats object

crookedneighbor:
Consolidated the task webhooks. This should be ready for review now.

common/script/ops/webhooks.js
```diff
@@ -0,0 +1,152 @@
+import refPush from '../libs/refPush';
+import validator from 'validator';
+import i18n from '../i18n';
+import { v4 as uuid } from 'uuid';
+import {
+  BadRequest,
+  NotFound,
+} from '../libs/errors';
+
+const TASK_ACTIVITY_DEFAULT_OPTIONS = Object.freeze({
```
paglias: we're not doing this anywhere else, it's okay for me if we decide to use it for config objects but it should be standard.

common/script/ops/webhooks.js
```diff
@@ -0,0 +1,152 @@
...
+  created: false,
+  updated: false,
+  deleted: false,
+  scored: true,
+});
+// Enumerates webhook types and provides functions for sanitizing options
+const WEBHOOK_TYPES = {
+  taskActivity (options = {}, language) {
+    options = Object.assign({}, TASK_ACTIVITY_DEFAULT_OPTIONS, options);
```
paglias: Is supported only in Firefox, maybe use lodash since we load it anyway?

common/script/ops/webhooks.js
```diff
@@ -0,0 +1,152 @@
...
+    type,
+    options,
+  };
+}
+
+function addWebhook (user, req = {}) {
+  let webhooks = user.preferences.webhooks;
+  let body = req.body || {};
+  let id = body.id || uuid();
+  let enabled = 'enabled' in body ? body.enabled : true;
```
paglias: this line shouldn't go in generateWebhookObject?

common/script/ops/webhooks.js
```diff
@@ -0,0 +1,152 @@
...
+    webhook.enabled = body.enabled;
+  } else {
+    webhook.enabled = oldWebhook.enabled;
+  }
+
+  webhook = generateWebhookObject(webhook, req);
+
+  user.markModified('preferences.webhooks');
+
+  user.preferences.webhooks[id].url = webhook.url;
```
paglias: assign?

paglias: or at least cache the result of `user.preferences.webhooks[id]`

migrations/20160806_add_missing_webhook_type.js
```diff
@@ -0,0 +1,106 @@
...
+  $gte: new Date("2014-01-01T00:00:00.000Z"),
+  $lte: new Date("2014-06-01T00:00:00.000Z"),
+};
+
+let Users;
+
+connectToDb(DB_URI).then((db) => {
+  Users = db.collection('users');
+
+  return Promise.resolve();
```
paglias: I think the Promise will use anything you return as the resolved value

migrations/20160806_add_missing_webhook_type.js
```diff
@@ -0,0 +1,106 @@
...
+.then(() => {
+  timer.stop();
+  closeDb();
+}).catch(reportError);
+
+function reportError (err) {
+  logger.error('Uh oh, an error occurred');
+  closeDb();
+  timer.stop();
+  throw err;
```
paglias: since it's called by `catch` I think the error will get swallowed because you could recover it with another catch statement. The workaround is to use setTimeout inside the catch

test/helpers/api-integration/v2/index.js
```diff
@@ -0,0 +1,8 @@
+// Import requester function, set it up for v2, export it
```
paglias: is this a leftover from v2?

website/server/libs/webhook.js
```diff
@@ -1,31 +1,122 @@
...
+  }
+
+  static defaultWebhookFilter () {
+    return true;
+  }
+
+  send (webhooks, data) {
+    let hooks = Object.keys(webhooks)
+      .map(hook => webhooks[hook])
+      .filter(isValidWebhook)
```
paglias: all these `filter` are nice but it's basically iterating 3 times over an array. I'd combine the three or use lodash chaining

website/server/models/group.js
```diff
@@ -446,6 +447,28 @@ schema.methods.startQuest = async function startQuest (user) {
   });
 };
 
+schema.methods.sendGroupChatReceivedWebhooks = function sendGroupChatReceivedWebhooks (chat) {
+  let query = {
+    'preferences.webhooks': { $ne: {} },
```
paglias: we have to monitor this query, it could get expensive for some guilds

paglias:
👍 there are only 2-3 comments important, the rest are code choices I'm not too sure about

common/script/ops/webhooks.js
```diff
@@ -0,0 +1,152 @@
...
+
+  if ('enabled' in body) {
+    webhook.enabled = body.enabled;
+  } else {
+    webhook.enabled = oldWebhook.enabled;
+  }
+
+  webhook = generateWebhookObject(webhook, req);
+
+  user.markModified('preferences.webhooks');
```
khipkin: I think this should be moved down so that it isn't called until after you finish modifying `user.preferences.webhooks`.

common/script/ops/webhooks.js
```diff
@@ -0,0 +1,152 @@
...
+  let webhook = generateWebhookObject({
+    id,
+    url: body.url,
+    enabled,
+    type: body.type,
+    options: body.options,
+  }, req);
+
+
+  user.markModified('preferences.webhooks');
```
khipkin: I think this should be moved down so that it isn't called until after you finish modifying `user.preferences.webhooks` (`webhooks`).

test/api/v3/integration/chat/POST-chat.test.js
```diff
@@ -54,6 +57,44 @@ describe('POST /chat', () => {
...
+    await member.post('/user/webhook', {
+      url: `http://localhost:${server.port}/webhooks/${memberUuid}`,
+      type: 'groupChatReceived',
+      enabled: true,
+      options: {
+        groupId: groupWithChat.id,
+      },
+    });
+
+    let message = await user.post(`/groups/${groupWithChat._id}/chat`, { message: testMessage});
```
khipkin: asymmetrical spacing around contents of {...}

crookedneighbor: Good catch. I should add that to the linter.

common/script/ops/webhooks.js
crookedneighbor: I think we should make using `Object.freeze` standard anywhere we are creating constant that is an object or an array.

common/script/ops/webhooks.js
crookedneighbor: Hmm... there's no support for IE at all, so I'll switch it out for lodash.

migrations/20160806_add_missing_webhook_type.js
crookedneighbor: You are right. :+1:

crookedneighbor:
Talked to Paglias and we decided to try out webhooks as an array to make the group chat query more performant. Since the webhook routes are still in beta and the mobile apps don't map the webhook object, this shouldn't be a problem.

common/script/ops/webhooks.js
crookedneighbor: It doesn't really matter when this gets called. It just tells mongoose to pay attention to that property when saving the document. 

I concede though that just reading the code it makes more logical sense to have it at the end. Will be unnecessary entirely if we make these arrays of objects though.

crookedneighbor:
Travis is conspicuously absent, but the build passed on my fork's copy of the branch https://travis-ci.org/crookedneighbor/habitrpg

This is ready for a final (hopefully) review

Main changes are:
- webhooks now live at the root level of the user instead of in preferences. 
- most of the validation logic happens inside the user via a webhook schema
- everything is server side, no common code (mostly so that the model can do the validation)
- groupChatReceived webhooks only look up users that have groupChatReceived webhooks
- webhooks are filtered in a single function instead of 3

website/client/js/services/userServices.js
```diff
@@ -532,15 +532,26 @@ angular.module('habitrpg')
...
-          callOpsFunctionAndRequest('updateWebhook', 'webhook', "PUT", data.params.id, data);
-        },
-
-        deleteWebhook: function (data) {
-          callOpsFunctionAndRequest('deleteWebhook', 'webhook', "DELETE", data.params.id, data);
+          return $http({
+            method: 'POST',
+            url: '/api/v3/user/webhook',
+            data: data,
+          }).then(sync);
```
paglias: here instead of syncing can we use the result of the API call to update the local user?

website/server/controllers/api-v3/webhook.js
```diff
@@ -0,0 +1,193 @@
...
+*/
+api.addWebhook = {
+  method: 'POST',
+  middlewares: [authWithHeaders()],
+  url: '/user/webhook',
+  async handler (req, res) {
+    let user = res.locals.user;
+    let webhook = new Webhook(req.body);
+
+    await webhook.validate();
```
paglias: I _think_ that being webhooks under user they get validated when the user is validated so this would be a repetition. If it's needed I'd at least use validateSync that is much faster

website/server/models/webhook.js
```diff
@@ -0,0 +1,77 @@
...
+  deleted: false,
+  scored: true,
+});
+
+export let schema = new Schema({
+  id: {
+    type: String,
+    required: true,
+    validate: [validator.isUUID, shared.i18n.t('invalidWebhookId')],
+    default () {
```
paglias: `default: uuid`? same functionality, only a stylistic change

crookedneighbor: Good call.

website/client/js/services/userServices.js
crookedneighbor: :+1:

website/server/controllers/api-v3/webhook.js
crookedneighbor: I think you're right we can just let user.save do the validation. At first I thought it was required for formatOptions t work correctly, but looking at it now, that is not the case.

paglias:
Looks good to me, only what about update and delete still syncing the entire user?

Then for me it's ready to get merged

paglias:
👍 let's remember to add the index on users. Should probably be a sparse index so that it's not built for users without the chat webhooks

website/server/models/group.js
```diff
@@ -446,6 +447,33 @@ schema.methods.startQuest = async function startQuest (user) {
   });
 };
 
+schema.methods.sendGroupChatReceivedWebhooks = function sendGroupChatReceivedWebhooks (chat) {
+  let query = {
+    webhooks: {
+      $elemMatch: {
```
crookedneighbor, line 529: Talked in slack with @paglias about using the shorthand for this instead, but I think for this case it needs to be `$elemMatch`, because its matching on both the `type` and the `options.groupId`

paglias: 👍

paglias:
comments https://github.com/HabitRPG/habitrpg/pull/7879#discussion_r74612951 and https://github.com/HabitRPG/habitrpg/pull/7879#discussion_r74613205 are still valid

crookedneighbor:
Thanks for the reminder.

ystyle:
I would like to ask webhooks now still only send `score change` information?

PR #7881, opened 2016-08-07
fix: shows user correct challenge when showing won challenge modal

### LINKED ISSUE #7716: Winner of challenge sees a notification for a previous challenge that they've won
Amazon_Warrior (4273f332-21e2-419a-8a35-e462ce341f5c) reports:
**I just won a Challenge-'Help the Poet'-but the alert that came up was for the very first Challenge I won 'A Weekend Fully Booked: 24-Hour Readathon'. The alert said I'd just won the Readathon Challenge, even though that happened ages ago.**

I've seen one other report of a similar problem and I was able to replicate the bug just now.

However I don't know if it happens always or just sometimes. It's probably the sort of thing that could happen always but few people report it.

_edit by Alys 2017-12-01:_

This bug is back. DCL ( 3bd49b39-283b-4d03-8117-7b87e9547e91 ) wrote in the Report a Bug guild: "when I closed a challenge (in the private guild I use to keep weekly to-do lists as challenges) and listed myself as the winner, the pop-up notification said I had won a different challenge! The challenge I closed was 'Weekly Work Tasks,' and the challenge it told me I won was 'Reflect on the Day...' which I have both in my personal guild and my party (and was the last challenge I won). The problem is just the pop-up, as the proper achievement is listed in my profile, and I got the gems I had associated with the challenge."

The code related to this will be different since the new website was released but the behaviour seems the same and the same style of fix might work, even though the tech is different, so I'm reopening this rather than creating a new issue. If other admins want to close an open a new one, feel free. :)

_edit by Alys 2020-04-14:_

It's back again. https://github.com/HabitRPG/habitica/issues/7716#issuecomment-613309165

Hus274:
Fixes #7716 
### Changes

Updating challenge pop up to sync the user object from the database before showing the user the challenge won modal to make sure that the most recent challenge won is the one that is displayed.

---

### DIFF
--- a/website/client/js/controllers/notificationCtrl.js
+++ b/website/client/js/controllers/notificationCtrl.js
@@ -97,7 +97,9 @@ habitrpg.controller('NotificationCtrl',
             $rootScope.openModal('rebirthEnabled');
             break;
           case 'WON_CHALLENGE':
-            $rootScope.openModal('wonChallenge', {controller: 'UserCtrl', size: 'sm'});
+            User.sync().then( function() {
+              $rootScope.openModal('wonChallenge', {controller: 'UserCtrl', size: 'sm'});
+            });
             break;
           case 'STREAK_ACHIEVEMENT':
             Notification.streak(User.user.achievements.streak);


crookedneighbor:
Makes sense to me.

crookedneighbor:

PR #7883, opened 2016-08-09
Collection quest progress reports "found no [item names]" fixes #7816

### LINKED ISSUE #7816: collection quest progress message for zero items should be "... found no [item names]"
_This is something that was discussed in Aspiring Socialites 45 days ago. The messages are still there if anyone wants to review the conversation (search for "found nothing")._

Currently, for collection quests, on any days when you find no items, the quest progress message is `[person's name] found nothing`

We'd like it changed to `[person's name] found no [item name(s)]` (for example, "Alys found no Moonstones" or "Alys found no Blue Fins and no Fire Coral").

The reasons are:
- It would help when troubleshooting quests (an admin could grep through the party log for "found .. Moonstones" and be sure to find all messages from that quest). 
- It might also help new players understand what's going on; if a quest starts amongst players with low perception, they could easily not find anything for the first couple of messages, and "found no Moonstones" is probably more meaningful than "found nothing" - they will already have seen the term "Moonstones" in the quest description and so will have some idea what it's talking about.

abatula:
Fixes #7816
### Changes

Collection quest progress reports "found 0 [item names] instead of "found nothing" if no items are found.
- For single item quests: "found 0 [item names]"
- For two or more item quests: "found 0 [item names 1], 0  [item names 2], ... etc"
  - If only one type of item is found, it will still report all items: "found 0 [item names 1], 2 [item names 2], ... etc"

Also updated collection quest tests.

Test 'sends a chat message if no progress is made' 
- checks for: '`Participating Member found 0 Bars of Soap.`'
- previously checked for '`Participating Member found nothing.`'

Added new test 'sends a chat message if no progress is made on quest with multiple items'
- checks that the chat message contains the strings:
  -  '`Participating Member found`'
  - '`0 Blue Fins`'
  - '`0 Fire Coral`'

Old Functionality:
Old Functionality

New Functionality:
newfunctionality

---


website/server/models/group.js
```diff
@@ -596,7 +596,21 @@ schema.methods._processCollectionQuest = async function processCollectionQuest (
     return m;
   }, []);
 
-  foundText = foundText.length > 0 ? foundText.join(', ') : 'nothing';
+  // Text for no found quest items
+  let questItems = Object.keys(this.quest.progress.collect);
+  for (let i = 0; i < questItems.length; i++) {
```
crookedneighbor: Let's use `map`

``` js
let questItems = Object.keys(this.quest.progress.collect).map((item) => {
  return quest.collect[item].text('en');
});
```

crookedneighbor: s/ma/map

crookedneighbor:
I'm not sure about this change. Eventually we want to make these system messages translatable, and formatting it this way would make it quite difficult.

abatula:
I wasn't sure if I should consider translations since the original didn't use them. I would imagine not bothering with contractions would make it easier to translate in the future. And it would make the code cleaner/simpler as well. 

And thanks for the tip on using map.

crookedneighbor:
My concern is the `and no... and no... and no....` part. With a variable number of items, that would be difficult to create a translation string. We could have a different translation string for every number of items, but as soon as we add a quest with 4 items, we'll also have to remember to add a new translation string for it.

ghost:
@crookedneighbor do you mean that would be difficult to code?

For what it's worth, from linguistic point of view I believe that some languages won't support negative sentence like this at all. I mean that in some languages we could say "didn't find any items" (or even "didn't find no items") but not "found no items" because the negative particle must be before/after the verb. So I think that when it comes to making this string translatable, many translators will choose to replace "no" with zero to reintroduce the old _buggy_ "Found 10 items and 0 items" as this will be the only way possible. In this regard I see no major issues with having separate lines for "no " and ", and no ".

Another way which may seem more natural might be refactoring the string to separate items found and not found to have something like "User found 10 these items, 20 these items and found no those items". Thus we'll have two complex strings to translate "found X", "found no Y" and connective ", and ".

Alys:
It's sounding like it would be better to use 0 rather than "no" (i.e., "found 0 Soap Bars" or "found 0 Coral and 0 Fins"). It's not perfect English, but it's easily understood, will require a bit less code, and, I think, won't have the same translation problems - @GitHubSphinx please tell us if I'm wrong about that!

ghost:
@Alys I believe you're right that "found 0 items" will be easier to code and translate, but leaving "found nothing" as is will be even easier than that  :smiley: However in general I'm not sure what is best here, and I really feel that I'm not the one to judge.

I've thought about this as a simple replacement "0"->"no" at first just because I imagined it would look better in English. I haven't considered translations mainly because I wasn't aware of the plans to make system messages translatable. In any case when this feature becomes translatable, some languages will have problems because of the need to ensure agreement (concordance) of particles, prepositions, postfixes/endings etc. 

Having this in mind, I would be slightly supporting "no" in English which can be turned into zero in translations if the phrase can't be structured the same way - but only if it can be implemented without serious efforts.

@Alys @Blade I wish I didn't make any suggestions from the very beginning of the discussion, I'm really sorry if they are misleading, interfering with the development plans or otherwise unwanted.

abatula:
With the caveat that I still know next to nothing about the HabitRPG code structure (and web programming in general), it seems to me that this problem only really exists because some part of the following code:

`let foundText = _.reduce(itemsFound, (m, v, k) => {
    m.push(`${v} ${quest.collect[k].text('en')}`);
    return m;
  }, []);`

returns a null or void object instead of the number 0 when no objects are found. Would the better solution be to edit this code so that 0 items found is treated like any other number of found items? It would requre the use of "found 0 [items]", rather than "found no [items]", but should be no harder to translate than any other number.

abatula:
I updated the progress report to "[user] found 0 [item1], 0 [item2]" based on suggestions for simplifying the code itself, and the future need for translations.

Another possible alternative is to change the original "found nothing" to "found no quest items" or "found no items for the [quest name] quest". This would keep a similar level of ease of future translation to "found nothing", while also providing more information to new users who might not understand what "found nothing' was referring to. Which, I believe, was the original reason for this change to be requested (ease of understanding for new users).

Alys:
Ease of understanding for new users was one of the reasons, but I also want to be able to grep through party logs for text like "found .\* moonstones" and be sure of finding all entries. That's useful in troubleshooting reports of quest problems. Because of that, I'd prefer "found 0 [item1]..." instead of "found no items for the [quest name] quest".

website/server/models/group.js
```diff
@@ -596,7 +596,16 @@ schema.methods._processCollectionQuest = async function processCollectionQuest (
...
 
-  foundText = foundText.length > 0 ? foundText.join(', ') : 'nothing';
+  // Text for no found quest items
+  let questItems = Object.keys(this.quest.progress.collect).map((item) => {
+    return quest.collect[item].text('en');
+  });
+  questItems.sort(); // Make sure message is always reported in the same order
+
+  let noFoundQuestItems = '0 ';
+  noFoundQuestItems += questItems.join(', 0 ');
```
khipkin: This logic does not work for translations because it assumes a similar word order to English.  Also, it looks like the questItems are all in English (i.e. not translated) anyway?

abatula:
Another attempt, this one is simpler and should be less of an issue for translations. Instead of creating a custom string when no items are found, all items that aren't found are added with a quantity of 0. This also means all items are always listed, even if only one type is found on a multiple-item quest.

If this version is approved, the remaining task is to fix the test for multiple items to allow the items to be reported in either order.

abatula:
I fixed the test so that the order in which the items are listed in the group chat message does not matter. I think the code fully resolves the issue now, but let me know if there is anything I should change.

crookedneighbor:
I made one adjustment to use a forEach instead of a for loop ed4039c08c2859b67836c2ba6659e81219024574

PR #7887, opened 2016-08-10
Groups assign tasks

TheHollidayInn:
Fixes put_issue_url_here
### Changes

This depends on #7785.

This adds the ability for a guild leader to assign and unassign users.

This will also create a separate task for assigned users and sync the task when it is updated or deleted.

---

UUID:


test/api/v3/integration/dataexport/GET-export_history.csv.test.js
```diff
@@ -39,12 +39,13 @@ describe('GET /export/history.csv', () => {
...
     let res = await user.get('/export/history.csv');
     let splitRes = res.split('\n');
+
     expect(splitRes[0]).to.equal('Task Name,Task ID,Task Type,Date,Value');
     expect(splitRes[1]).to.equal(`habit 1,${tasks[0]._id},habit,${moment(tasks[0].history[0].date).format('YYYY-MM-DD HH:mm:ss')},${tasks[0].history[0].value}`);
     expect(splitRes[2]).to.equal(`habit 1,${tasks[0]._id},habit,${moment(tasks[0].history[1].date).format('YYYY-MM-DD HH:mm:ss')},${tasks[0].history[1].value}`);
-    expect(splitRes[3]).to.equal(`daily 1,${tasks[1]._id},daily,${moment(tasks[1].history[0].date).format('YYYY-MM-DD HH:mm:ss')},${tasks[1].history[0].value}`);
-    expect(splitRes[4]).to.equal(`habit 2,${tasks[2]._id},habit,${moment(tasks[2].history[0].date).format('YYYY-MM-DD HH:mm:ss')},${tasks[2].history[0].value}`);
-    expect(splitRes[5]).to.equal(`habit 2,${tasks[2]._id},habit,${moment(tasks[2].history[1].date).format('YYYY-MM-DD HH:mm:ss')},${tasks[2].history[1].value}`);
+    expect(splitRes[3]).to.equal(`habit 2,${tasks[2]._id},habit,${moment(tasks[2].history[0].date).format('YYYY-MM-DD HH:mm:ss')},${tasks[2].history[0].value}`);
```
paglias, line 46: why you had to change this test?

website/server/controllers/api-v3/tasks.js
```diff
@@ -341,6 +234,14 @@ api.updateTask = {
     // see https://github.com/Automattic/mongoose/issues/2749
 
     let savedTask = await task.save();
+
+    if (task.group.id && task.assignedUsers.length > 0) {
+      let group = await Group.getGroup({user, groupId: task.group.id, populateLeader: false});
```
paglias: you should limit the fields fetched here or at least exclude the chat because it's very large

website/server/controllers/api-v3/tasks.js
```diff
@@ -951,6 +852,14 @@ api.deleteTask = {
       throw new NotAuthorized(res.t('cantDeleteChallengeTasks'));
     }
 
+    if (task.group.id) {
+      //  @TODO: Abstract this access snippet
+      let group = await Group.getGroup({user, groupId: task.group.id, populateLeader: false});
```
paglias: same here, limit the fetched fields

website/server/controllers/api-v3/tasks.js
```diff
@@ -951,6 +852,14 @@ api.deleteTask = {
...
+      if (!group) throw new NotFound(res.t('groupNotFound'));
+      if (group.leader !== user._id) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
+      group.removeTask(task);
```
paglias: await?

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,187 @@
...
+    let assignedUser = await User.findById(assignedUserId);
+
+    let taskId = req.params.taskId;
+    let task = await Tasks.Task.findByIdOrAlias(taskId, user._id);
+
+    if (!task) {
+      throw new NotFound(res.t('taskNotFound'));
+    }
+
+    if (!task.group || !task.group.id) {
```
paglias: checking for task.group should not be necessary as it'll always exist, even if empty

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,187 @@
...
+      throw new NotAuthorized(res.t('onlyGroupTasksCanBeAssigned'));
+    }
+
+    let group = await Group.getGroup({user, groupId: task.group.id, populateLeader: false});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    if (group.leader !== user._id) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
+
+    group.syncTask(task, assignedUser);
```
paglias: await?

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,187 @@
...
+    let assignedUser = await User.findById(assignedUserId);
+
+    let taskId = req.params.taskId;
+    let task = await Tasks.Task.findByIdOrAlias(taskId, user._id);
+
+    if (!task) {
+      throw new NotFound(res.t('taskNotFound'));
+    }
+
+    if (!task.group || !task.group.id) {
```
paglias: same as above about checking for task.group

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,187 @@
...
+      throw new NotAuthorized(res.t('onlyGroupTasksCanBeAssigned'));
+    }
+
+    let group = await Group.getGroup({user, groupId: task.group.id, populateLeader: false});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    if (group.leader !== user._id) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
+
+    group.unlinkTask(task, assignedUser);
```
paglias: await?

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,187 @@
...
+
+    res.respond(201, task);
+
+    return null;
```
paglias: this is not necessary

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,187 @@
...
+    let group = await Group.getGroup({user, groupId: task.group.id, populateLeader: false});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    if (group.leader !== user._id) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
+
+    group.syncTask(task, assignedUser);
+
+    res.respond(201, task);
+
+    return null;
```
paglias: not necessary

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,187 @@
...
+    let group = await Group.getGroup({user, groupId: req.params.groupId, populateLeader: false});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    if (group.leader !== user._id) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
+
+    let tasks = await createTasks(req, res, {user, undefined, group});
+
+    res.respond(201, tasks.length === 1 ? tasks[0] : tasks);
+
+    return null;
```
paglias: not necessary

website/server/libs/api-v3/taskManager.js
```diff
@@ -0,0 +1,155 @@
...
+
+  // Compares the short names in tasks against
+  // a Set, where values cannot repeat. If the
+  // lengths are different, some name was duplicated
+  if (aliases.length !== [...new Set(aliases)].length) {
+    throw new BadRequest(res.t('taskAliasAlreadyUsed'));
+  }
+
+  await Bluebird.map(tasksWithAliases, (task) => {
+    return task.validate();
```
paglias: This is not strictly related to this PR but since we don't have async mongoose validators (as far as i remember) we could change this to use `task.validateSync()` which should be considerably faster as it's only sync code

website/server/libs/api-v3/taskManager.js
```diff
@@ -0,0 +1,155 @@
...
+    return tasks;
+  }
+}
+
+// Takes a Task document and return a plain object of attributes that can be synced to the user
+
+export function syncableAttrs (task) {
+  let t = task.toObject(); // lodash doesn't seem to like _.omit on Document
+  // only sync/compare important attrs
+  let omitAttrs = ['_id', 'userId', 'challenge', 'history', 'tags', 'completed', 'streak', 'notes', 'updatedAt'];
```
paglias: this should probably include `group` too now

website/server/models/group.js
```diff
@@ -748,6 +756,113 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
...
+    _.merge(matchingTask, syncableAttrs(taskToSync));
+    // Make sure the task is in user.tasksOrder
+    let orderList = user.tasksOrder[`${taskToSync.type}s`];
+    if (orderList.indexOf(matchingTask._id) === -1 && (matchingTask.type !== 'todo' || !matchingTask.completed)) orderList.push(matchingTask._id);
+  }
+
+  if (!matchingTask.notes) matchingTask.notes = taskToSync.notes; // don't override the notes, but provide it if not provided
+  if (matchingTask.tags.indexOf(group._id) === -1) matchingTask.tags.push(group._id); // add tag if missing
+
+  toSave.push(matchingTask.save());
```
paglias: toSave.push(matchingTask.save(), taskToSync.save(), ...)

website/server/models/group.js
```diff
@@ -748,6 +756,113 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
...
+    linkedTaskId: unlinkingTask._id,
+    userId: user._id,
+  };
+
+  let assignedUserIndex = unlinkingTask.assignedUsers.indexOf(user._id);
+  unlinkingTask.assignedUsers.splice(assignedUserIndex, 1);
+
+  if (keep === 'keep-all') {
+    await Tasks.Task.update(findQuery, {
+      $set: {'group.id': undefined},
```
paglias: $unset group.id?

website/server/models/task.js
```diff
@@ -57,6 +57,8 @@ export let TaskSchema = new Schema({
   },
   attribute: {type: String, default: 'str', enum: ['str', 'con', 'int', 'per']},
   userId: {type: String, ref: 'User', validate: [validator.isUUID, 'Invalid uuid.']}, // When not set it belongs to a challenge
+  assignedUsers: [{type: String, ref: 'User', validate: [validator.isUUID, 'Invalid uuid.']}],
```
paglias: these two should be added to the list of not settable attributes

website/server/libs/taskManager.js
```diff
@@ -0,0 +1,170 @@
...
+  });
+}
+
+
+/**
+ * Creates tasks for a user, challenge or group.
+ *
+ * @param  req  The Express req variable
+ * @param  res  The Express res variable
+ * @param  options  Should be a user, user and challenge to create a challenge task or user and group to create a group task
```
crookedneighbor: I like this, but let's be more specific:

``` js
* @param  options 
* @param  options.user  The user that these tasks belong to
* @param  options.challenge  The challenge that these tasks belong to
* @param  options.group  The group that these tasks belong to
```

website/server/libs/taskManager.js
```diff
@@ -0,0 +1,170 @@
...
+  toSave = toSave.map(taskData => {
+    // Validate that task.type is valid
+    if (!taskData || Tasks.tasksTypes.indexOf(taskData.type) === -1) throw new BadRequest(res.t('invalidTaskType'));
+
+    let taskType = taskData.type;
+    let newTask = new Tasks[taskType](Tasks.Task.sanitize(taskData));
+
+    if (challenge) {
+      newTask.challenge.id = challenge.id;
+    } else {
```
crookedneighbor, line 58: Is this logic correct? Do you always needs to pass in a user when passing in a group?

paglias, line 58: Hmm yeah... if it works like challenges tasks do then it there should be no userId when it's the original group task

paglias, line 58: About this comment, there's this code https://github.com/HabitRPG/habitrpg/pull/7887/files#diff-7e3afbc9a950c0819d20e0d1c440ee4aR209 that loads the challenge if there's no `task.userId` it'll load the challenge automatically which should not happen if the task is linked to a group instead

website/server/controllers/api-v3/tasks.js
```diff
@@ -962,6 +851,14 @@ api.deleteTask = {
       throw new NotAuthorized(res.t('cantDeleteChallengeTasks'));
     }
 
+    if (task.group.id) {
+      //  @TODO: Abstract this access snippet
+      let group = await Group.getGroup({user, groupId: req.params.groupId, populateLeader: false, fields: '_id'});
```
paglias: populateLeader is false by default

test/api/v3/unit/libs/taskManager.js
```diff
@@ -0,0 +1,171 @@
...
+    expect(task.down).to.equal(testHabit.down);
+    expect(task.createdAt).isNotEmtpy;
+    expect(task.challenge.id).to.equal(challenge._id);
+  });
+
+  it('returns syncable attibutes', async () => {
+    req.body = testHabit;
+    res.t = i18n.t;
+
+    let tasks = await createTasks(req, res, {user, challenge, undefined});
```
paglias: undefined

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,183 @@
...
+    if (reqValidationErrors) throw reqValidationErrors;
+
+    let user = res.locals.user;
+
+    let group = await Group.getGroup({user, groupId: req.params.groupId, populateLeader: false, fields: '_id'});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    if (group.leader !== user._id) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
+
+    let tasks = await createTasks(req, res, {user, undefined, group});
```
paglias: undefined

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,183 @@
...
+  middlewares: [authWithHeaders()],
+  async handler (req, res) {
+    req.checkParams('groupId', res.t('groupIdRequired')).notEmpty().isUUID();
+
+    let reqValidationErrors = req.validationErrors();
+    if (reqValidationErrors) throw reqValidationErrors;
+
+    let user = res.locals.user;
+
+    let group = await Group.getGroup({user, groupId: req.params.groupId, populateLeader: false, fields: '_id'});
```
paglias: populateLeader: false by default

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * @apiSuccess {Array} data An array of tasks
+ */
+api.getGroupTasks = {
+  method: 'GET',
+  url: '/tasks/group/:groupId',
+  middlewares: [authWithHeaders()],
+  async handler (req, res) {
+    req.checkParams('groupId', res.t('groupIdRequired')).notEmpty().isUUID();
+    let types = Tasks.tasksTypes.map(type => `${type}s`);
```
paglias: I think this can go outside the controller so it's not calculated anytime

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,183 @@
...
+    req.checkQuery('type', res.t('invalidTaskType')).optional().isIn(types);
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let user = res.locals.user;
+
+    let group = await Group.getGroup({user, groupId: req.params.groupId, populateLeader: false, fields: '_id'});
```
paglias: populateLeader: false by default

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,183 @@
...
+
+    if (!task) {
+      throw new NotFound(res.t('taskNotFound'));
+    }
+
+    if (!task.group.id) {
+      throw new NotAuthorized(res.t('onlyGroupTasksCanBeAssigned'));
+    }
+
+    let group = await Group.getGroup({user, groupId: req.params.groupId, populateLeader: false, fields: '_id'});
```
paglias: populateLeader: false by default

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,183 @@
...
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    if (group.leader !== user._id) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
+
+    await group.syncTask(task, assignedUser);
+
+    res.respond(201, task);
```
paglias: 200 not 201

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    let group = await Group.getGroup({user, groupId: req.params.groupId, populateLeader: false, fields: '_id'});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    if (group.leader !== user._id) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
+
+    await group.unlinkTask(task, assignedUser);
+
+    res.respond(201, task);
```
paglias: 200

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -0,0 +1,183 @@
...
+
+    if (!task) {
+      throw new NotFound(res.t('taskNotFound'));
+    }
+
+    if (!task.group.id) {
+      throw new NotAuthorized(res.t('onlyGroupTasksCanBeAssigned'));
+    }
+
+    let group = await Group.getGroup({user, groupId: req.params.groupId, populateLeader: false, fields: '_id'});
```
paglias: populateLeader: false by default

website/server/libs/taskManager.js
paglias: If we decide to use this doc style we should try to be consistent on how and where we use it

website/server/libs/taskManager.js
```diff
@@ -0,0 +1,170 @@
...
+  });
+
+  // tasks with aliases need to be validated asyncronously
+  await _validateTaskAlias(toSave, res);
+
+  toSave = toSave.map(task => task.save({ // If all tasks are valid (this is why it's not in the previous .map()), save everything, withough running validation again
+    validateBeforeSave: false,
+  }));
+
+  toSave.unshift((challenge || user).save());
```
paglias: group is missing?

website/server/libs/taskManager.js
```diff
@@ -0,0 +1,170 @@
...
+    challenge,
+    group,
+  } = options;
+
+  let query = {userId: user._id};
+
+  if (challenge) {
+    query =  {'challenge.id': challenge.id, userId: {$exists: false}};
+  } else if (group) {
+    query =  {'group.id': group._id};
```
paglias: userId: {$exists: false}

website/server/libs/taskManager.js
```diff
@@ -0,0 +1,170 @@
...
+      {type: 'todo', completed: false},
+      {type: {$in: ['habit', 'daily', 'reward']}},
+    ];
+  }
+
+  let tasks = await Tasks.Task.find(query).exec();
+
+  // Order tasks based on tasksOrder
+  if (type && type !== 'completedTodos' && type !== '_allCompletedTodos') {
+    let order = (challenge || user).tasksOrder[type];
```
paglias: group

website/server/libs/taskManager.js
```diff
@@ -0,0 +1,170 @@
...
+    return tasks;
+  }
+}
+
+// Takes a Task document and return a plain object of attributes that can be synced to the user
+
+export function syncableAttrs (task) {
+  let t = task.toObject(); // lodash doesn't seem to like _.omit on Document
+  // only sync/compare important attrs
+  let omitAttrs = ['_id', 'userId', 'challenge', 'history', 'tags', 'completed', 'streak', 'notes', 'updatedAt', 'group', 'assignedUsers', 'linkedTaskId'];
```
paglias: linkedTaskId should be under task.group?

website/server/models/group.js
```diff
@@ -760,6 +770,168 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
   return await Bluebird.all(promises);
 };
 
+schema.methods.updateTask = async function updateTask (taskToSync) {
+  let group = this;
+  let listOfPromises = [];
+
+  let usersToSync = await User.find({_id: { $in: taskToSync.assignedUsers} });
+
+  usersToSync.forEach(function syncAssignedUser (user) {
```
paglias: usersToSync can be really heavy with many users so it should only be used when there's only one user involved. When you have to update the tasks for many users you should directly issue an update command to mongo like we do for challenges https://github.com/HabitRPG/habitrpg/blob/develop/website/server/models/challenge.js#L202

website/server/models/group.js
```diff
@@ -760,6 +770,168 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
...
+    linkedTaskId: unlinkingTask._id,
+    userId: user._id,
+  };
+
+  let assignedUserIndex = unlinkingTask.assignedUsers.indexOf(user._id);
+  unlinkingTask.assignedUsers.splice(assignedUserIndex, 1);
+
+  if (keep === 'keep-all') {
+    await Tasks.Task.update(findQuery, {
+      $unset: {'group.id': ''},
```
paglias: maybe this should set task.group to an empty object? for challenges we do `$set: {challenge: {}},`

website/server/models/group.js
```diff
@@ -760,6 +770,168 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
...
+    userId: {$exists: true},
+    'group.id': group.id,
+    linkedTaskId: task._id,
+  }, {
+    $set: {'group.broken': 'TASK_DELETED'},
+  }, {multi: true}).exec();
+};
+
+// API v2 compatibility methods
+schema.methods.getTransformedData = function getTransformedData (options) {
```
paglias: This is a leftover from when API v2 was still there, should be removed

website/server/models/task.js
```diff
@@ -64,6 +66,11 @@ export let TaskSchema = new Schema({
     winner: String, // user.profile.name of the winner
   },
 
+  group: {
+    id: {type: String, ref: 'Group', validate: [validator.isUUID, 'Invalid uuid.']},
+    broken: {type: String, enum: ['GROUP_DELETED', 'TASK_DELETED', 'UNSUBSCRIBED', 'GROUP_TASK_NOT_FOUND']}, // CHALLENGE_TASK_NOT_FOUND comes from v3 migration
```
paglias: I think `GROUP_TASK_NOT_FOUND` doesn't really apply for groups but only for challenges

website/server/models/group.js
TheHollidayInn: You told me to change this to $unset above.

website/server/controllers/api-v3/tasks/groups.js
TheHollidayInn: Why is this 201 then: https://github.com/HabitRPG/habitrpg/blob/develop/website/server/controllers/api-v3/tasks.js#L92 Is 201 for content created?

website/server/models/group.js
paglias: Yeah I should have suggested to use $set before, sorry. The $unset was only for group.id but it's better to reset the entire object

website/server/controllers/api-v3/tasks/groups.js
paglias: yeah in theory 201 means "CREATED"

website/server/models/group.js
TheHollidayInn: Yea, I just copied from the Challenge function before, no worries.

website/server/controllers/api-v3/tasks/groups.js
TheHollidayInn: Okay, I'll keep that in mind.

website/server/libs/taskManager.js
TheHollidayInn: I think it would be useful to have this for challenge and group - and in the future other items. Not sure if you agree.

website/server/models/task.js
TheHollidayInn: Even though these are linked to group tasks?

website/server/libs/taskManager.js
paglias: Hmm it could make sense but the problem is that the linked task for challenges it's stored under task.challenge.taskId and it would be difficult to migrate it now

website/server/models/task.js
paglias: yeah because we don't don't really use CHALLENGE_TASK_NOT_FOUND but I messed up with the v3 migration so some task got unlinked even though it shouldn't have. But it's not really used in the code, it was just to identify what needed to be fixed from the migration

website/server/libs/taskManager.js
crookedneighbor: Agreed.

website/server/libs/taskManager.js
paglias: same thing for assignedUsers

website/server/libs/taskManager.js
crookedneighbor: Instead of passing `(group || challenge || user)` everywhere, we could do

``` js
let owner = group || challenge || user;
```

website/server/libs/taskManager.js
TheHollidayInn, line 58: I was thinking we don't depend on not having a task.userId. I think depending on the task.linkTaskId is more reliable and understandable. Challenge has a linkedId as well, so the idea of assuming a task is a challenge task if it doesn't have a user is implicit when the logic could be explicit.

website/server/libs/taskManager.js
TheHollidayInn: Okay, but I think we should think more generally though. What if we want assigned users from challenges or something else in the future. The linkedTaskID is an example. We didn't generalize the challenge linked id, now we have two fields that are essentially the same.

crookedneighbor:
Killed your build because it failed on the task manager tests. `owner.save` is not a function

paglias:
Can we call task.group.linkedTaskId task.group.taskId so that it matches task.challenge.taskId?

crookedneighbor:
I agree with @paglias

TheHollidayInn:
That makes sense. I'll update that

website/server/libs/api-v3/taskManager.js
TheHollidayInn: I had to revert this change because it breaks https://github.com/HabitRPG/habitrpg/blob/develop/website/server/models/task.js#L164

TheHollidayInn:
Updated.

paglias:
I think it's ready to be merged. @crookedneighbor opinions?

paglias:
Maybe if we merge it we should block the routes from being usable in production?

Also, there are currently no checks on whether the group is part of a group plan and thus can use the task assigning feature. Like a flag on the group model

TheHollidayInn:
Updated with middleware to prevent use on the live site. The payments will be in a separate PR.

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -1,5 +1,5 @@
 import { authWithHeaders } from '../../../middlewares/auth';
-import { ensureDevelpmentMode } from '../../../middlewares/ensureDevelpmentMode';
+import ensureDevelpmentMode from '../../../middlewares/ensureDevelpmentMode';
```
paglias, line 2: hmm we're missing an `o` in the file name. ensureDevelOpmentMode

TheHollidayInn: Ah, yea. Looks like it was merged in like that. That was my bad. I can fix that Ina separate PR. Other files will be affected.

paglias: Sounds good

website/server/controllers/api-v3/tasks.js
```diff
@@ -634,9 +634,18 @@ api.addTagToTask = {
     if (validationErrors) throw validationErrors;
 
     let taskId = req.params.taskId;
-    let task = await Tasks.Task.findByIdOrAlias(taskId, user._id, { userId: user._id });
+    let task = await Tasks.Task.findByIdOrAlias(taskId, user._id);
+
+    if (!task) {
```
paglias: `if (!task || task.userId !== user._id)` since the last `else if` can be added to the first one

website/server/controllers/api-v3/tasks.js
```diff
@@ -674,9 +683,17 @@ api.removeTagFromTask = {
     if (validationErrors) throw validationErrors;
 
     let taskId = req.params.taskId;
-    let task = await Tasks.Task.findByIdOrAlias(taskId, user._id, { userId: user._id });
+    let task = await Tasks.Task.findByIdOrAlias(taskId, user._id);
 
-    if (!task) throw new NotFound(res.t('taskNotFound'));
+    if (!task) {
```
paglias: same here

website/server/controllers/api-v3/tasks.js
```diff
@@ -634,9 +634,18 @@ api.addTagToTask = {
     if (validationErrors) throw validationErrors;
 
     let taskId = req.params.taskId;
-    let task = await Tasks.Task.findByIdOrAlias(taskId, user._id, { userId: user._id });
+    let task = await Tasks.Task.findByIdOrAlias(taskId, user._id);
```
paglias: will this work with tasks that do not have an userId?

website/server/controllers/api-v3/tasks.js
paglias: Hmm actually... this way the leader will be able to edit tags for all tasks linked to the group, even the ones in the users' accounts. Is it okay?

website/server/controllers/api-v3/tasks.js
TheHollidayInn: No, I don't believe this is correct. This will now allow a check for challenge or group tasks that do not have a user._id. You can see the previous logic for how this worked with challenges.

website/server/controllers/api-v3/tasks.js
TheHollidayInn: I'm not seeing wherein the group.syncTask function that the tags get synced. This should be exactly like challenges. Am I missing something?

website/server/controllers/api-v3/tasks.js
```diff
@@ -312,9 +204,15 @@ api.updateTask = {
 
     let taskId = req.params.taskId;
     let task = await Tasks.Task.findByIdOrAlias(taskId, user._id);
+    let group;
 
     if (!task) {
       throw new NotFound(res.t('taskNotFound'));
+    } else if (task.group.id) {
```
paglias: @TheHollidayInn continuing from previous discussion because it's not outdated, I see one problem here: you're passing `user._id` to `Tasks.Task.findByIdOrAlias` so a task with no `userId`(like the original ones for challenges and group) will never be found. So lines 217-219 of this file will never be reached.

As for challenges original tasks there should be some special logic for when `task.userId` doesn't exist. If task.userId exists task.challenge or task.groups should not be considered at all except for sanitization (see line 231)

website/server/controllers/api-v3/tasks.js
```diff
@@ -166,7 +166,7 @@ api.getTask = {
 
     if (!task) {
       throw new NotFound(res.t('taskNotFound'));
-    } else if (!task.userId) { // If the task belongs to a challenge make sure the user has rights
+    } else if (task.challenge.id && !task.userId) { // If the task belongs to a challenge make sure the user has rights
```
khipkin, line 169: Maybe pull this boolean logic into a helper function that takes a task object?  It seems you make this check often enough that it would be beneficial.

paglias: I think the check is simple enough that it doesn't require a specific function

PR #7889, opened 2016-08-10
fix: Correct purchase call to use purchaseType, not identifier

crookedneighbor:
The view for purchasing things in the market was using the category identifier instead of the item's purchaseType to make the purchase. This fixes that and removes the logic to accept premiumHatchingPotions since the purchase type for those is just hatchingPotions.

Since we just added the premiumHatchingPotions check and didn't document in the api docs that premiumHatchingPotions are an accepted type, I don't consider this a breaking change.

@SabreCat 

@vIiRuS sanity check. The mobile apps don't use the category identifier to make the purchase, right? They use purchaseType?

### DIFF
--- a/common/script/ops/purchase.js
+++ b/common/script/ops/purchase.js
@@ -60,11 +60,10 @@ module.exports = function purchase (user, req = {}, analytics) {
     ];
   }
 
-  let acceptedTypes = ['eggs', 'hatchingPotions', 'premiumHatchingPotions', 'food', 'quests', 'gear'];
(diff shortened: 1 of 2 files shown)


SabreCat:
Good call! I knew there had to be some way to achieve what you're up to here, but didn't spot it in last night's emergency situation.

crookedneighbor:
Messaged @vIiRuS about this. Confirmed that mobile apps use `purchaseType`

PR #7890, opened 2016-08-10
Fix preferences typeError

SabreCat:
Fixes an issue where double-clicking a task to edit it would throw `TypeError: Cannot read property 'preferences' of undefined`.
### Changes

In most places where we call `editTask`, we provide the task and the user object. However, in the double-click action for tasks, we only sent the task, which resulted in the function being unable to find the user's editing preferences. Now this bit also provides the user object.


crookedneighbor:
Good catch!

PR #7899, opened 2016-08-14
Update email check exist

cching0L2:
https://github.com/HabitRPG/habitrpg/issues/7847 Update email route should check that proposed email is not already being used 
### Changes
- throw an error when user tries to update his/her email in setting -> site to an email that is already used by another user
- can't explicitly show "email is taken" message, because this hurts users privacy (as suggested by @Alys 
- added a generic error message that goes "Your request cannot be fulfilled.  Email admin@habitica.com if this error persists.", hoping that this message can be used somewhere else.

screen shot 2016-08-13 at 11 20 01 pm

Note: the updated message has a period at the end.  This is a slightly outdated screenshot.

---

UUID: 62d73118-8bef-4085-b2ff-81f522b4a3e2

### DIFF
--- a/common/locales/en/front.json
+++ b/common/locales/en/front.json
@@ -252,5 +252,6 @@
     "onlySocialAttachLocal": "Local authentication can be added to only a social account.",
     "invalidReqParams": "Invalid request parameters.",
     "memberIdRequired": "\"member\" must be a valid UUID.",
-    "heroIdRequired": "\"heroId\" must be a valid UUID."
+    "heroIdRequired": "\"heroId\" must be a valid UUID.",
+    "cannotFulfillReq":"Your request cannot be fulfilled.  Email admin@habitica.com if this error persists"
 }
--- a/common/locales/en/front.json
+++ b/common/locales/en/front.json
@@ -253,5 +253,5 @@
     "invalidReqParams": "Invalid request parameters.",
     "memberIdRequired": "\"member\" must be a valid UUID.",
     "heroIdRequired": "\"heroId\" must be a valid UUID.",
-    "cannotFulfillReq":"Your request cannot be fulfilled.  Email admin@habitica.com if this error persists"
+    "cannotFulfillReq":"Your request cannot be fulfilled.  Email admin@habitica.com if this error persists."
 }
--- a/website/server/controllers/api-v3/auth.js
+++ b/website/server/controllers/api-v3/auth.js
@@ -532,6 +532,12 @@ api.updateEmail = {
     let validationErrors = req.validationErrors();
     if (validationErrors) throw validationErrors;
 
+    let userSameEmail = await User.findOne({'auth.local.email': req.body.newEmail}).exec();
+    if (userSameEmail) {
+        console.log("email already taken by another user");
+        throw new NotAuthorized(res.t('emailTaken'));
+    }
+
     let candidatePassword = passwordUtils.encrypt(req.body.password, user.auth.local.salt);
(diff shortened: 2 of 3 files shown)


website/server/controllers/api-v3/auth.js
```diff
@@ -532,6 +532,9 @@ api.updateEmail = {
     let validationErrors = req.validationErrors();
     if (validationErrors) throw validationErrors;
 
+    let userSameEmail = await User.findOne({'auth.local.email': req.body.newEmail}).exec();
```
paglias: this should be `await User.findOne({'auth.local.email': req.body.newEmail}).select({_id: 1}.lean().exec();` so that only the _id field is loaded and not the entire user

common/locales/en/front.json
```diff
@@ -252,5 +252,6 @@
     "onlySocialAttachLocal": "Local authentication can be added to only a social account.",
     "invalidReqParams": "Invalid request parameters.",
     "memberIdRequired": "\"member\" must be a valid UUID.",
-    "heroIdRequired": "\"heroId\" must be a valid UUID."
+    "heroIdRequired": "\"heroId\" must be a valid UUID.",
+    "cannotFulfillReq":"Your request cannot be fulfilled.  Email admin@habitica.com if this error persists."
```
paglias, line 256: we don't want to show a more explicative error message?

paglias, line 256: Just saw the reason in the PR description but as far as I know during registration we tell explicitly if an email is already taken

Alys, line 256: There's a discussion about it in the Aspiring Coders guild, but in brief, since we've configured the forgot password form to not leak information about which email addresses are in use, it seems sensible to do the same as best we can for changing your email address.

Also, if someone is trying to use an email address that already exists, they would probably benefit from admin assistance. Copying from one of my posts in that guild:

The situation would arise if they'd forgotten that they already had another account, possibly created a long time ago (we'd want them to know about it because it would have good loot for them!), or because they were mistyping their own email address and might need someone to point that out to them :), or because someone else was incorrectly using their email address either maliciously or by accident (definitely a situation that admins should investigate!).

website/server/controllers/api-v3/auth.js
crookedneighbor: Ditto. Also I'd prefer the variable to be called `emailAlreadyInUse`

website/server/controllers/api-v3/auth.js
```diff
@@ -532,6 +532,9 @@ api.updateEmail = {
...
+    if (userSameEmail) throw new NotAuthorized(res.t('cannotFulfillReq'));
```
crookedneighbor: Is not authorized the right error for this?

paglias: Bad Request?

common/locales/en/front.json
paglias, line 256: Hmm.. I'm a little concerned that we show a different message when registering but okay

website/server/controllers/api-v3/auth.js
cching0L2: I noticed that we throw a not authorized error in the registration route when the email is taken.  But I'm not sure in this case should we maintain consistency or throw a bad request error instead because we want to protect privacy (?)

cching0L2:
Incorporated the changes suggested above

Alys:
@cching0L2 I've awarded you a tier. You can read more about [contributor tiers and rewards here](http://habitrpg.wikia.com/wiki/Contributor_Rewards).

PR #7900, opened 2016-08-15
Analytics fix

crookedneighbor:
This should fix the regression caused by 3.32.0

@Alys @SabreCat

### DIFF
--- a/test/api/v3/unit/libs/analyticsService.test.js
+++ b/test/api/v3/unit/libs/analyticsService.test.js
@@ -94,7 +94,19 @@ describe('analyticsService', () => {
           amplitudeNock
             .filteringPath(/httpapi.*platform.*3rd\%20Party.*/g, '');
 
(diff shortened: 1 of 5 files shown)


test/api/v3/unit/libs/analyticsService.test.js
```diff
@@ -101,6 +101,18 @@ describe('analyticsService', () => {
               amplitudeNock.done();
             });
         });
+
+        it('logs unknown if headers are not passed in', () => {
+          amplitudeNock
+            .filteringpath(/httpapi.*platform.*unkown.*/g, '');
```
khipkin: unkown -> unknown

crookedneighbor: Thanks, that's what I get for writing this in the 5 minutes before I go to bed without actually running the tests

PR #7901, opened 2016-08-15
(server) Add parties array to store invites

### LINKED ISSUE #7792: Allow Multiple Invites
Right now, there is a problem where if you try to send someone an invite but they have another pending, you cannot send the invitation. Users should be allowed to have multiple invites pending, and when they choose a Party, the other Party invites should be automatically rejected.  (Note that this should not happen with Guilds.)

eternal44:
Fixes #7792 
### Changes

Allows users to receive multiple party invites by adding a 'parties' array to the user's 'invitation' object in the schema.

We still need to figure out how we want multiple invites to render on the client side.

### DIFF
--- a/website/server/controllers/api-v3/groups.js
+++ b/website/server/controllers/api-v3/groups.js
@@ -231,8 +231,8 @@ api.joinGroup = {
 
     if (group.type === 'party' && group._id === user.invitations.party.id) {
       inviter = user.invitations.party.inviter;
-      user.invitations.party = {}; // Clear invite
-      user.markModified('invitations.party');
+
+      clearPartyInvitation(user, group._id);
 
       // invite new user to pending quest
       if (group.quest.key && !group.quest.active) {
@@ -340,7 +340,7 @@ api.rejectGroupInvite = {
     let isUserInvited = false;
 
     if (groupId === user.invitations.party.id) {
-      user.invitations.party = {};
+      clearPartyInvitation(user, group._id);
       user.markModified('invitations.party');
       isUserInvited = true;
     } else {
@@ -542,9 +542,6 @@ async function _inviteByUUID (uuid, group, inviter, req, res) {
     }
     userToInvite.invitations.guilds.push({id: group._id, name: group.name, inviter: inviter._id});
   } else if (group.type === 'party') {
-    if (userToInvite.invitations.party.id) {
-      throw new NotAuthorized(res.t('userAlreadyPendingInvitation'));
-    }
 
     if (userToInvite.party._id) {
       let userParty = await Group.getGroup({user: userToInvite, groupId: 'party', fields: 'memberCount'});
@@ -553,7 +550,7 @@ async function _inviteByUUID (uuid, group, inviter, req, res) {
       if (userParty && userParty.memberCount !== 1) throw new NotAuthorized(res.t('userAlreadyInAParty'));
     }
 
-    userToInvite.invitations.party = {id: group._id, name: group.name, inviter: inviter._id};
+    userToInvite.invitations.parties.push({id: group._id, name: group.name, inviter: inviter._id});
   }
 
   let groupLabel = group.type === 'guild' ? 'Guild' : 'Party';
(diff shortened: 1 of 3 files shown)


website/server/models/user/schema.js
```diff
@@ -357,6 +357,7 @@ let schema = new Schema({
     party: {type: Schema.Types.Mixed, default: () => {
       return {};
     }},
+    parties: {type: Array, default: () => []},
```
crookedneighbor: Since it's an array, we can be more explicit about the contents of the array.

website/server/controllers/api-v3/groups.js
```diff
@@ -340,7 +340,7 @@ api.rejectGroupInvite = {
     let isUserInvited = false;
 
     if (groupId === user.invitations.party.id) {
-      user.invitations.party = {};
+      clearPartyInvitation(user, group._id);
       user.markModified('invitations.party');
```
crookedneighbor, line 345: You're calling this twice.

website/server/controllers/api-v3/groups.js
```diff
@@ -718,4 +715,19 @@ api.inviteToGroup = {
   },
 };
 
+// Transitioning invitation storage to an array of objects (see issue #7792 for reference)
+function clearPartyInvitation(user, groupID) {
+  user.invitations.parties.forEach(function(party, index, partiesArr) {
```
crookedneighbor: I think you can use the `removeFromArray` utility here. Then you won't even need to mark `user.invitations.parties` as modified.

crookedneighbor:
You have a number of linting errors and I imagine you'll have to update some tests.

website/server/controllers/api-v3/groups.js
eternal44: I read through the removeFromArray utility and didn't see a markModified call in it.  If it's not being handled elsewhere let's keep this bit as-is.

eternal44:
Thanks I saw that earlier!  I'll make the changes in my next push.

website/server/controllers/api-v3/groups.js
crookedneighbor: It doesn't need markModified because it uses array methods that allow mongoose to keep track of what needs to be updated.

crookedneighbor: We wrote that utility specifically for that reason.

eternal44: Nice.  It's been a while since I used Mongoose but I just read if it's a mixed schema we need a markModified call ([click for link and see 'Mixed'](http://mongoosejs.com/docs/schematypes.html)).  Let me know if I missed something!

eternal44:
I linted the files but noticed that there were failing tests for cron jobs before this PR.  All other tests pass and if my recent changes are ok I'll squash the commits before an admin merges.

website/server/models/user/schema.js
```diff
@@ -357,6 +358,7 @@ let schema = new Schema({
     party: {type: Schema.Types.Mixed, default: () => {
       return {};
     }},
+    parties: {type: [{}], default: () => []},
```
crookedneighbor, line 361: This should be an array of objects with specific values for the properties instead.

eternal44, line 361: What was the reason for specific values for the properties?  I was thinking of leaving the schema description as a generic array of objects so future developers could make changes to the User model easily.  If your comment was regarding syntax for the type description I pulled it from the docs [(see under 'array')](http://mongoosejs.com/docs/schematypes.html)

Assuming the syntax is correct - I wanted to keep this a generic array of objects to make it easier for future changes to the User model.  Plus the Guild model seems to have taken a more generic pattern so I figured it'd be better to assume a similar pattern.

eternal44:
I'm looking to update the routes & methods that reference user.invitations.party to reference it's new parties property instead but am having trouble pinpointing the user.get('/groups/party/invites') route.  Any suggestions?

eternal44:
@crookedneighbor - hey just wanted to check in to see how you were doing - any thoughts?.  I'd love to continue to contribute!

website/server/models/user/schema.js
crookedneighbor, line 361: We try to be as specific as possible with _most_ attributes. Helps us keep data in a known good state. So in this case, it would be:

``` js
parties: [{
  id: {
    type: String, 
    ref: 'Group', 
    required: true,
    validate: [validator.isUUID, 'Invalid uuid.'],
  }, 
  name: {
    type: String, 
    required: true,
  }, 
  inviter: {
    type: String, 
    ref: 'User', 
    required: true,
    validate: [validator.isUUID, 'Invalid uuid.'],
  },
}],
```

crookedneighbor:
The test is here: test/api/v3/integration/groups/GET-groups_groupId_invites.test.js

Almost all of the integration tests use the route as their top most describe block, so I found it by doing:

``` bash
git grep "groups/:groupId/invites" test/
```

crookedneighbor:
@eternal44 What's the status on this? Are you working on the tests?

crookedneighbor:
Going to close this now since I haven't heard back, but if you're still working on it, feel free to re-open it when you're ready.

PR #7902, opened 2016-08-15
add route for QR codes

phillipthelen:
This adds a route, that redirects a special URL to the profiles page for a user. The idea/reason is that we can then change the profiles url to something nicer, while still keeping existing qr codes usable.

### DIFF
--- a/test/api/v3/integration/qrcodes/GET-qrcodes_user.test.js
+++ b/test/api/v3/integration/qrcodes/GET-qrcodes_user.test.js
@@ -6,7 +6,7 @@ import superagent from 'superagent';
 import nconf from 'nconf';
 
 const API_TEST_SERVER_PORT = nconf.get('PORT');
-describe.only('GET /qr-code/user/:memberId', () => {
+describe('GET /qr-code/user/:memberId', () => {
   let user;
 
   before(async () => {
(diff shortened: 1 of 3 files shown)


website/server/controllers/top-level/qrcodes.js
```diff
@@ -0,0 +1,17 @@
+let api = {};
+
+api.redirectProfileQRCode = {
+  method: 'GET',
+  url: '/qr-code/user/:memberid',
+  runCron: false,
+  async handler (req, res) {
+    req.checkParams('memberid', res.t('memberIdRequired')).notEmpty();
```
paglias: .isUUID()

crookedneighbor:
Are we considering this a private part of the API? If not, it should be documented.

Could we get a test that it redirects?

paglias:
It's outside the API, I'd say it's better to keep it private

test/api/v3/integration/qrcodes/GET-qrcodes_user.test.js
```diff
@@ -0,0 +1,32 @@
+import {
+  generateUser,
+  translate as t,
+} from '../../../../helpers/api-v3-integration.helper';
+import superagent from 'superagent';
+import nconf from 'nconf';
+
+const API_TEST_SERVER_PORT = nconf.get('PORT');
+describe.only('GET /qr-code/user/:memberId', () => {
```
crookedneighbor: `.only`

paglias:
Looks good to me 👍

PR #7904, opened 2016-08-16
Achievement list renovation & Achievements API

### LINKED ISSUE #2893: Achievement List Renovation
The achievement list is getting pretty big. Should we consider a new way to start showing achievements to avoid clutter?

I was thinking maybe a grid of the badges with a pop-up of the achievement name and description when you hover over it. Not sure if that's a good fix, but something should probably be done eventually. (I imagine we'll need to do something similar with the quests in the store at some point. Perhaps a separate quests page. One part is for gem-bought quests and the other for random encounters)
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1419720-achievement-list-renovation?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

khipkin:
Fixes #2893
## Changes
- [x] Divided all possible achievements into three categories: Basic, Seasonal, and Special.
  1. **Basic**: Streaks, Perfect Days, Party Up, Party On, Beast Master, Mount Master, Triad Bingo, Healer Ultimate Gear, Rogue Ultimate Gear, Warrior Ultimate Gear, Mage Ultimate Gear, Rebirth
  2.  **Seasonal**: Habitica Day, Habit Birthday, Annoying Friends (Snowballs), Alarming Friends (Spooky Sparkles), Agricultural Friends (Shiny Seeds), Aquatic Friends (Seafoam), Dilatory World Quest, Stressbeast World Quest, Burnout World Quest, Bewilder World Quest, Costume Contest, Greeting Cards, Thank You Cards, NYE Cards, Valentine Cards, Birthday Cards
  3.  **Special**: Habitica Surveys, Contributor, _NPC Backer_, _Backer (Any Tier)_, _Veteran_, _Original User_

_(Italicized achievements are only shown if they are achieved; otherwise they are hidden so that users aren't disheartened by visible-yet-unachievable achievements.)_
- [x] Unachieved achievements are now always shown, regardless of whether the profile being viewed is yours or someone else's.  (Note: this does not include the special-case achievements mentioned above.)
- [x] Made achievement strings more consistent.  (See discussion on #2893 for details.)
  - Replaced manual string concatenation with parameterized string interpolation.
  - Achievements with singular-vs-plural structures now consistently use the singular form when the count is exactly 1 or when the achievement is not achieved (in which case, the achievement is muted), otherwise the plural form is used.
  - Renamed "Helped Habit Grow" achievement to "Helped Habitica Grow". (requested by @lemoness)
- [x] Added count bubbles for applicable achievements when they are achieved.  This only applies to non-binary achievements.  (For example, party up/on, ultimate gear, etc. would not have counts.)
- [x] Show achievements in a grid display (similar to the Pets/Mounts grids).
  - Each achievement uses a 2x (aka large) badge as its image.  The achievement title and description appear in a popup on mouseover.
    - [x] Added any 2x images that were missing.
  - Each achievement shows the normal (colored) badge if it is achieved; otherwise, it shows the special, dark "unachieved" badge.
    - [x] Added a 2x "unachieved" badge (credit to @tricksyfox) to be displayed for unearned achievements (except the hidden-if-unachieved special ones).
- [x] @negue and I added an achievement API!  This makes it possible to list a user's achievements using the following route: `/member/{memberId}/achievements`.  The returned object will look something like this, with the format of each achievement over on the right:
  image
## Remaining Questions
- [x] Some of the achievement descriptions contain links to pages with additional information.  How can we make the popovers stick around long enough for users to click on the links?  Maybe clicking on the achievement icon makes its popover persist until clicked again?
  
  > This is a good plan for now; we can revisit this in the upcoming website redesign, if desired.
- [x] Should the formatting of the popover text change for achieved vs unachieved achievements?  For example: previously, unachieved titles/descriptions were previously `muted`, meaning the text was gray.  Maybe that could still be the case, with the text in the popover being `muted` in addition to the dark "unearned" badge?  (See discussion on #2893 for examples and screenshots.)
  
  > ~~Ideally, for unachieved achievements, the text would change to reflect what the user must do. ("To earn this badge, blah blah blah...") If that's tricky from a technical perspective, we can just mute the text instead.~~
  > Just mute the text for now, and then we'll make a separate ticket to change the strings later.
- [x] _Some achievements (notably the Contributor and Backer achievements) previously used colored spans in the achievement titles to highlight the special status of the user's contributions.  Should these colors be used somehow in the new layout?  If so, how?_
  
  > Not really worth it for now; we can revisit this in the upcoming website redesign, if desired.
## Screenshots

_The layout of the achievements with the survey and contributor achievements in the "Special" section, with the hidden-if-unachieved achievements both shown and hidden.  This screenshot does not include counts._

image

_An example achievement grid for a user, with both achieved and unachieved achievements shown.  Applicable achieved achievements show the count bubbles._

image

_Sample popovers for the earned "Perfect Days" achievement and the unearned "Mount Master" achievement._

image

image

_The new achievements grid as seen in a member modal._

image


website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
   },
 };
 
+function _addAchievement(achievementsResult, type, key, title, text, icon, category, value, earned) {
```
crookedneighbor: All of these functions should take in an options object instead of positioned arguments. So when called, you would do:

``` js
_addAchievement({
  achievementsResult: 'value', 
  type: 'something', 
  key: 'hey', 
  title: 'impressive thing', 
  text: 'owowow', 
  icon: 'hey', 
  category: 'yay', 
  value: 'foo',
  earned: true
});
```

Instead of 

``` js
_addAchievement('value',  'something', 'hey', 'impressive thing', 'owowow', 'hey', 'yay',  'foo', true);
```

This has the benefit of providing named arguments, so you're never confused about what you're passing in and you don't have to worry about the position in which you pass it.

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+function _addSimpleAchievement (res, memberObj, achievementsResult, key, icon, category, keySuffixOrLabels, alreadyParsed) {
+  let value = memberObj.achievements[key];
+  let isEarned = !!value;
+
+  let labels = {
+    title: key,
+    text: key + 'Text',
+  };
+
+  if (keySuffixOrLabels instanceof Object) {
```
crookedneighbor: Just so you know, in JS: `null instanceof Object === true`

crookedneighbor: So it's probably better to do the `!keySuffixOrLabels` check first.

crookedneighbor:
@negue I see a number of linting errors. Once the merge conflicts are sorted out, the build will fail.

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+      text: key + 'Text',
+    };
+  }
+
+  _addAchievement(achievementsResult, 'simple', key, res.t(labels.title), res.t(labels.text, {count: value}), icon, category, value, isEarned);
+}
+
+function _addSimpleAchievementWithMasterCount (res, memberObj, achievementsResult, key, icon, category) {
+  let value = memberObj.achievements[key + 'Count'];
+  let isEarned = !!value;
```
crookedneighbor: All these `!!value` coercions should be written as `Boolean(value)` instead.

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+  },
+};
+
+// copy from rootCtrl => $scope.contribText
+function contribText(contrib, backer, res){
+  if (!contrib && !backer) return;
+  if (backer && backer.npc) return backer.npc;
+  var l = contrib && contrib.level;
+  if (l && l > 0) {
+    var level = (l < 3) ? res.t('friend') : (l < 5) ? res.t('elite') : (l < 7) ? res.t('champion') : (l < 8) ? res.t('legendary') : (l < 9) ? res.t('guardian') : res.t('heroic');
```
crookedneighbor: I know you just copy pasted this, but this is awful. Let's do a big if/else if chain instead of nested ternaries.

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+  _addPluralAchievement(res, member, achievements, 'costumeContest', 'achievements-costumeContest', 'Seasonal', 'costumeContests', {
+    singularTitle: 'costumeContest',
+    singularText: 'costumeContestText',
+    pluralTitle: 'costumeContest',
+    pluralText: 'costumeContestTextPlural',
+  });
+
+  let cardAchievements = ['greeting', 'thankyou', 'nye', 'valentine', 'birthday'];
+
+  for (let index in cardAchievements) {
```
crookedneighbor:
``` js
cardAvhievements.forEach((key) => {
```

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+
+  // Seasonal-Achievements
+  _addPluralAchievement(res, member, achievements, 'habiticaDay', 'achievements-habiticaDay', 'Seasonal', 'habiticaDays', {
+    singularTitle: 'habiticaDay',
+    singularText: 'habiticaDaySingularText',
+    pluralTitle: 'habiticaDay',
+    pluralText: 'habiticaDayPluralText',
+  });
+
+  _addPluralAchievement(res, member, achievements, 'habiticaDay', 'achievements-habiticaDay', 'Seasonal', 'habitBirthdays', {
```
khipkin: Copy-paste errors? (`'habiticaDay'` and `'achievements-habiticaDay'`)

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+    if (!member) throw new NotFound(res.t('userWithIDNotFound', {userId: memberId}));
+
+    let achievements = fillAchievements(member, res);
+
+    res.respond(200, achievements);
+  },
+};
+
+// copy from rootCtrl => $scope.contribText
+function contribText(contrib, backer, res){
```
khipkin: Is there any way to reuse this code instead of copying it?

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+  }
+}
+
+function fillAchievements (member, res) {
+  var achievements = {
+    _count: 0,
+  };
+
+  // Basic Achievements
+  _addPluralAchievement(res, member, achievements, 'streak', 'achievements-thermometer', 'Basic');
```
khipkin: I think most (if not all) of these sprite names should be `'achievement-{achiev}'`, singular "achievement" instead of plural "achievements".

website/server/controllers/api-v3/members.js
khipkin: Confirmed.  Should be `'habitBirthdays'` and `'achievement-habitBirthday'`.

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+
+  _addUltimateGearAchievement(res, member, achievements, 'healer', 'achievement-ultimate-healer', 'Basic');
+  _addUltimateGearAchievement(res, member, achievements, 'rogue', 'achievement-ultimate-rogue', 'Basic');
+  _addUltimateGearAchievement(res, member, achievements, 'warrior', 'achievement-ultimate-warrior', 'Basic');
+  _addUltimateGearAchievement(res, member, achievements, 'mage', 'achievement-ultimate-mage', 'Basic', 'wizard');
+
+  // TODO Rebirth Achievement
+
+  // Seasonal-Achievements
+  _addPluralAchievement(res, member, achievements, 'habiticaDay', 'achievements-habiticaDay', 'Seasonal', 'habiticaDays', {
```
khipkin: Should be `'habiticaDays'` and `'achievement-habiticaDay'`.

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+  _addSimpleAchievementWithCount(res, member, achievements, 'shinySeed', 'achievements-shinySeed', 'Seasonal', {
+    title: 'agriculturalFriends',
+    text: 'agriculturalFriendsText',
+  });
+  _addSimpleAchievementWithCount(res, member, achievements, 'seafoam', 'achievements-seafoam', 'Seasonal', {
+    title: 'aquaticFriends',
+    text: 'aquaticFriendsText',
+  });
+
+  _addSimpleAchievement(res, member, achievements, 'dilatory', 'achievements-dilatory', 'Seasonal', {
```
khipkin: `'quests.dilatory'`

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+    title: 'achievementDilatory',
+    text: 'achievementDilatoryText',
+  });
+
+  _addSimpleAchievement(res, member, achievements, 'stressbeast', 'achievements-stoikalm', 'Seasonal', {
```
khipkin: `'quests.stressbeast'`

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+    title: 'achievementStressbeast',
+    text: 'achievementStressbeastText',
+  });
+
+  _addSimpleAchievement(res, member, achievements, 'burnout', 'achievements-burnout', 'Seasonal', {
```
khipkin: `'quests.burnout'`

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+    title: 'achievementBurnout',
+    text: 'achievementBurnoutText',
+  });
+
+  _addSimpleAchievement(res, member, achievements, 'bewilder', 'achievements-bewilder', 'Seasonal', {
```
khipkin: `'quests.bewilder'`

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+    title: 'achievementBewilder',
+    text: 'achievementBewilderText',
+  });
+
+  _addPluralAchievement(res, member, achievements, 'costumeContest', 'achievements-costumeContest', 'Seasonal', 'costumeContests', {
```
khipkin: `'costumeContests'` and `'achievement-costumeContest'`

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +57,288 @@ api.getMember = {
...
+    singularTitle: 'helped',
+    singularText: 'surveysSingle',
+    pluralTitle: 'helped',
+    pluralText: 'surveysMultiple',
+  });
+
+
+  if(member.contributor.level) {
+    _addAchievement(achievements, 'simple', 'contributor', contribText(member.contributor, member.backer, res), 
+      res.t('contribText'), 'achievements-boot', 'Seasonal', member.contributor.level, true);
```
khipkin: Special for all of these

website/server/controllers/api-v3/members.js
negue: yeah, I'll change it, but it would also increase wrongly used parameters, don't you think?

website/server/controllers/api-v3/members.js
negue: oh nice! didn't know about Boolean(value), thx

website/server/controllers/api-v3/members.js
negue: Also I get this from eslint 
`
[eslint] 
Expected property shorthand. (babel/object-shorthand)
(property) value: any
`

so that I have to use 
`{
type,
key,
title
}` ?

crookedneighbor: It the variable is the same name as the property, then yes.

website/server/controllers/api-v3/members.js
```diff
@@ -57,47 +57,81 @@ api.getMember = {
...
+    icon: data.icon,
+    category: data.category,
+    key: data.key,
+    value: data.value,
+    earned: data.earned,
+    index: achievementsResult._count++,
+  };
+}
+
+function _addQuestAchievement (res, memberObj, achievementsResult, key, icon, category, labels) {
```
khipkin: Use the same pattern as above (i.e., pass all the params in using one `options` or `data` param) for the rest of these methods as well, for the same reason(s) mentioned before.

website/server/controllers/api-v3/members.js
```diff
@@ -106,16 +140,40 @@ function _addSimpleAchievementWithCount (res, memberObj, achievementsResult, key
     };
   }
 
-  _addAchievement(achievementsResult, 'simple', key, res.t(labels.title), res.t(labels.text, {count: value}), icon, category, value, isEarned);
+  if (!value) {
+    value = 0;
+  }
+
+  console.log(key + ' ' + labels.text + ' ' + value);
```
khipkin: Remove this.

website/server/controllers/api-v3/members.js
```diff
@@ -194,12 +270,27 @@ api.getMemberAchievements = {
 };
 
 // copy from rootCtrl => $scope.contribText
-function contribText(contrib, backer, res){
+function contribText (contrib, backer, res) {
   if (!contrib && !backer) return;
   if (backer && backer.npc) return backer.npc;
   var l = contrib && contrib.level;
```
khipkin: Just a nit, but the variable name `l` looks uncomfortably similar to the number `1`, which is confusing. :stuck_out_tongue:

website/server/controllers/api-v3/members.js
negue: `res`, `memberObj` and `achievementsResult` also inside the `data` object? I'd rather have those as separate params. @khipkin / @crookedneighbor  your opinion please ^^

khipkin: No, I think those are fine being standalone, since they are required and must always be passed in.

negue:
Extracted the whole achievement building to a separate class

khipkin:
@tricksyfox's "unachieved" badge art has been approved by @lemoness! I have updated the PR description accordingly.

khipkin:
I just need to add tests, and then this should be ready for review.

khipkin:
@Alys @lemoness Any thoughts on the remaining questions described above?

khipkin:
@Alys @lemoness All of my remaining questions are small-ish UI questions, so they may not even matter, given the upcoming UI overhaul.

Alys:
A related point from the Bug guild, from Daniel Widrew (9682e84a-17b0-4ad6-8cb8-5277d6800155) 
"it's not a bug so much as a typo, but on the stats & achievements page, the medal you get for filling out surveys and etc is called "Helped Habit Grow" instead of Helped Habitica Grow"

My response:
"Hmmmmmmmm. Good point, although we sort of used Habit as an alternative name a bit in the past (some users still do occasionally) and that achievement is an old one. Possibly should stay the same for historical reasons??"

But if anyone thinks it should be changed, go ahead! It does sound odd now.

khipkin:
@crookedneighbor What work still needs to be done?  I haven't received any feedback except for @Alys's question/non-request above. :stuck_out_tongue_winking_eye:

crookedneighbor:
I saw that there were unchecked boxes in your list of tasks and assumed it wasn't done yet. I'll update the labels.

khipkin:
@crookedneighbor The unchecked boxes are unanswered UI questions that I still need feedback on.

mandyzhou:
My two cents on the unanswered UI questions:
- I like the idea of clicking on an achievement icon to make its pop-up persist until clicked again -- similar to what happens currently when clicking a task's Note icon.
- I think it would make sense to mute the text for unearned achievements.
- For the achievements with colored spans, maybe they can have something similar to the count bubbles? The bubbles could follow the color scheme from the colored spans and could contain the tier number for the contributor achievements and maybe "NPC" for the backer achievement.

website/views/shared/profiles/achievements.jade
```diff
@@ -1,267 +1,31 @@
...
-    small(ng-if='::profile.achievements.mountMasterCount')
-      =env.t('mountMasterText2', {count: "{{::profile.achievements.mountMasterCount}}"})
-  hr
-
-div(ng-if='::profile.achievements.triadBingo || user._id == profile._id')
-  .achievement.achievement-triadbingo(ng-if='::profile.achievements.triadBingo')
-  div(ng-class='::{muted: !profile.achievements.triadBingo}')
-    h5=env.t('triadBingoName')
+mixin simpleAchiev(achiev)
+  div(popover-title='{{achiev.title}}',
```
crookedneighbor: Any reason we're not using the one time binding mark for these? `{{::acheiv.title}}`

I assume they won't be changing once instantiated on controller load.

khipkin: Probably just got lost in the chaos... I'll bring back the magic colons. :+1:

common/script/index.js
```diff
@@ -97,6 +97,9 @@ api.statsComputed = statsComputed;
 import shops from './libs/shops';
 api.shops = shops;
 
+import achievs from './libs/achievs';
```
crookedneighbor: Let's not abbreviate this here. I'd prefer it to be achievements.

test/api/v3/integration/members/GET-members_id.test.js
```diff
@@ -54,3 +54,42 @@ describe('GET /members/:memberId', () => {
     });
   });
 });
+
+describe('GET /members/:memberId/achievements', () => {
```
crookedneighbor: This should be a separate test file.

crookedneighbor:
Could you remove the compiled sprites from the PR? Having them in there makes it a tad difficult to review through the website interface. (On my MacBook Air, the page just crawls examining the diff)

``` bash
git checkout develop common/dist/sprites/
```

should do it.

The reviewers can compile them locally.

common/locales/en/character.json
```diff
@@ -60,8 +60,11 @@
     "gearAchievement": "You have earned the \"Ultimate Gear\" Achievement for upgrading to the maximum gear set for a class! You have attained the following complete sets:",
     "moreGearAchievements": "To attain more Ultimate Gear badges, change classes on <a href='/#/options/profile/stats' target='_blank'>your stats page</a> and buy up your new class's gear!",
     "armoireUnlocked": "You've also unlocked the <strong>Enchanted Armoire!</strong> Click on the Enchanted Armoire Reward for a random chance at special Equipment! It may also give you random XP or food items.",
-    "ultimGearName": "Ultimate Gear",
-    "ultimGearText": "Has upgraded to the maximum weapon and armor set for the following classes:",
+    "ultimGearName": "Ultimate Gear - <%= ultClass %>",
+    "warriorUltimGearText": "Has upgraded to the maximum weapon and armor set for the Warrior class.",
```
crookedneighbor: This should probably be 

``` json
"ultimGearText": "Has upgraded to the maximum weapon and armor set for the <%= class(locale) %> class.",
```

And you'd pass in `i18n.t('warrior')` as the parameter:

``` js
let text = i18n.t('ultimGearText', { class: i18n.t(data.path) }, data.language);
```

Assuming `data.path` is the class.

common/locales/en/generic.json
```diff
@@ -86,15 +89,15 @@
     "originalUserText": "One of the <em>very</em> original early adopters. Talk about alpha tester!",
     "habitBirthday": "Habitica Birthday Bash",
     "habitBirthdayText": "Celebrated the Habitica Birthday Bash!",
-    "habitBirthdayPluralText": "Celebrated <%= number %> Habitica Birthday Bashes!",
+    "habitBirthdayPluralText": "Celebrated <%= count %> Habitica Birthday Bashes!",
```
crookedneighbor: :+1: on standardizing this.

common/script/libs/achievements.js
```diff
@@ -0,0 +1,507 @@
+import i18n from '../i18n';
+
+let achievs = {};
+let index;
+
```
crookedneighbor: Is there a good reason to construct these achievements this way instead of enumerating them in the content files?

khipkin: Hmm not sure. @negue? What were your thoughts initially?

negue: I don't really understand what do you mean by content files, since there are no achievements in the content api result.

At first I had this logic inside a class, which would be created on every request.

Or do you mean the functions in this file? Anyway ping me on Slack ^^

crookedneighbor: Copying my note to @negue in slack here:

> Could the different achievements instead be a part of the content API?
> 
> And then the route would just loop through the users achievements, check the key against the content api's achievements object (which would contain the name, notes and type, etc)
> 
> And return the achievements grouped by type

khipkin: Ohh @crookedneighbor I see what you're saying!  Yeah, that makes sense.  I can do that.

khipkin: So all of the logic for figuring out if a given user earned certain achievements would be OUTSIDE of the content api, correct?

The only info inside the content api would be the localization string keys, sprite name string, and maybe the achievement types?

crookedneighbor: Exactly.

common/script/content/achievements.js
```diff
@@ -0,0 +1,163 @@
...
+  habitSurveys: {
+    icon: 'achievement-tree',
+    singularTitleKey: 'helped',
+    singularTextKey: 'surveysSingle',
+    pluralTitleKey: 'helped',
+    pluralTextKey: 'surveysMultiple',
+  },
+};
+
+let ultimateGearTypes = ['healer', 'rogue', 'warrior', 'mage'];
```
crookedneighbor: Should probably use this constant: https://github.com/HabitRPG/habitrpg/blob/a4fd687510160fc54e5730266f4f921f63f49c53/common/script/content/constants.js#L3-L8

common/script/content/achievements.js
```diff
@@ -0,0 +1,163 @@
+import { each } from 'lodash';
+
+let achievementsData = {
```
crookedneighbor: Would it make sense to group some of these together into separate objects and then assign them all into the exported module at the end?

``` js
let worldQuestAchievements = {
  dilatoryQuest: { /* data */ },
  stressbeastQuest: { /* data */ },
  burnoutQuest: { /* data */ },
  bewilderQues: { /* data */ },
};

let transformBuffAchievements = {
  snowball: { /* data */ },
  spookySparkles: { /* data */ },
  shinySeed: { /* data */ },
}

// more achievement objects

let achievements = Object.assign({}, worldQuestAchievements, transformBuffAchievements, /* more variables */)
```

I think that would make it easier to navigate around this file

common/script/content/achievements.js
```diff
@@ -0,0 +1,163 @@
...
+let ultimateGearTypes = ['healer', 'rogue', 'warrior', 'mage'];
+each(ultimateGearTypes, (type) => {
+  achievementsData[`${type}UltimateGear`] = {
+    icon: `achievement-ultimate-${type}`,
+    titleKey: 'ultimGearName',
+    textKey: 'ultimGearText',
+  };
+});
+
+let cardTypes = ['greeting', 'thankyou', 'nye', 'valentine', 'birthday'];
```
crookedneighbor: If we go with the setting up separate variables I mentioned above, you could do:

``` js
let cardAchievements = ['greeting', 'thankyou', 'nye', 'valentine', 'birthday'].reduce((achievements, type) => {
  achievements[`${type}Cards`] = {
    icon: `achievement-${type}`,
    titleKey: `${type}CardAchievementTitle`,
    textKey: `${type}CardAchievementText`,
  };
  return achievements;
}, {});
```

Even if we decided to still do just one big object, you could do:

``` js
['greeting', 'thankyou', 'nye', 'valentine', 'birthday'].forEach((type) => {
  achievementsData[`${type}Cards`] = {
    icon: `achievement-${type}`,
    titleKey: `${type}CardAchievementTitle`,
    textKey: `${type}CardAchievementText`,
  };
});
```

website/client-old/js/controllers/userCtrl.js
```diff
@@ -83,5 +83,9 @@ habitrpg.controller("UserCtrl", ['$rootScope', '$scope', '$location', 'User', '$
       }).join(',');
     }
 
+    $scope.achievements = Shared.achievements.getAchievementsForProfile($scope.profile);
+    $scope.achievPopoverPlacement = 'right';
+    $scope.achievAppendToBody = 'true'; // append-to-body breaks popovers in modal windows
+
```
crookedneighbor: Extraneous line break here.

website/client-old/js/controllers/memberModalCtrl.js
```diff
@@ -65,5 +69,6 @@ habitrpg
             $scope.$close();
           });
       }
+
```
crookedneighbor: Extra line break here.

website/server/controllers/api-v3/members.js
```diff
@@ -57,6 +58,41 @@ api.getMember = {
   },
 };
 
+/**
+ * @api {get} /api/v3/members/:memberId/achievements Get member achievements object
```
crookedneighbor, line 63: Let's document this API route better including what exactly is contained on the success object and any errors you might encounter. See how the webhook routes are documented in this PR https://github.com/HabitRPG/habitrpg/pull/7879/files#diff-cf049c3d45b117fee868e17845f1f130R1

common/script/libs/achievements.js
```diff
@@ -0,0 +1,322 @@
...
+    value = 0;
+  }
+
+  let key = data.key || data.path;
+  let thisContent = achievsContent[key];
+
+  // If value === 0, use singular versions of strings.
+  // If value !== 0, use plural versions of strings.
+
+  let title = i18n.t(value === 0 ? thisContent.singularTitleKey : thisContent.pluralTitleKey, {count: value}, data.language);
```
crookedneighbor: Can we split this out a little?

``` js
let titleKey, textKey;

if (value === 0) {
  titleKey = thisContent.singularTitleKey;
  textKey = thisContent.singularTextKey;
} else {
  titleKey = thisContent.pluralTitleKey;
  textKey = thisContent.pluralTextKey;
}

let title = i18n(titleKey, {count: value}, data.language);
let text = i18n(textKey, {count: value}, data.language);
```

common/script/libs/achievements.js
```diff
@@ -0,0 +1,322 @@
...
+    key,
+    value,
+    optionalCount: value,
+    earned: Boolean(value),
+  });
+}
+
+function _addPlural (result, user, data) {
+  let value = user.achievements[data.path];
+  if (!value) {
```
crookedneighbor: I prefer `let value = user.achievements[data.path] || 0;` for this.

common/script/libs/achievements.js
negue: Adding the achievements to the content api is a good choice, but now the achievements result just has some string-keys not the actual values (like every other property in the content), I guess they should have the localized values right? @crookedneighbor 

@khipkin good job already extracting those keys :+1: 

Any 3rd party application would still rather use the /members/{mid}/achievements api call to get the achievements, instead of enumerating through the content array and trying to recreate the achievements client-wise.

@khipkin gonna test the changes tomorrow, but on the first look it probably still works as before^^

common/script/libs/achievements.js
```diff
@@ -0,0 +1,322 @@
...
+  if (user.achievements.originalUser) {
+    _addSimple(result, user, {path: 'originalUser', language});
+  }
+
+  return result;
+}
+
+// Build and return the given user's achievement data.
+achievs.getAchievementsForProfile = function getAchievementsForProfile (user, language) {
+  index = 0;
```
crookedneighbor: Pretty sure this won't work. If this function gets run a second time before the first time is finished, it'll reset the index.

common/script/libs/achievements.js
```diff
@@ -0,0 +1,322 @@
...
+  let result = {
```
crookedneighbor: Maybe this is too much too dream, but I feel like we could do:

``` js
achievs.getAchievementsForProfile = function getAchievementsForProfile (user, language) {
  let achievements = {
    basic: {
      label: 'Basic',
      achievements: [],
    },
    seasonal: {
      label: 'Seasonal',
      achievements: [],
    },
    special: {
      label: 'Special',
      achievements: [],
    },
  };

  assignAchievements(user, achievements);

  return achievements;
}

let achievementGenerators = {
  plural() { /* add plural logic */ },
  simple() { /* add simple logic */ },
  // other generators
}

function assignAchievements (user, result) {
  _.each(user.achievements, (data, key) => {
     let achievement = achievementContent[key];
     let kind = achievement.kind;
     let type = achievement.type;
     let achievementData = achievementGenerators[type](achievement, data);

     result[kind].achievements.push(achievementData);
  });
}
```

Then this script does not need to have _any_ knowledge of the achievements that are getting passed in. Just that if it's an achievement of this type, do this and add it to this kind.

crookedneighbor: Or is the logic just so wonky that it's not worth doing this kind of meta-programming?

common/script/content/achievements.js
```diff
@@ -0,0 +1,163 @@
+import { each } from 'lodash';
+
+let achievementsData = {
+  dilatoryQuest: {
+    icon: 'achievement-dilatory',
+    titleKey: 'achievementDilatory',
```
crookedneighbor: Why not wrap this with the `t` function? https://github.com/HabitRPG/habitrpg/blob/a4fd687510160fc54e5730266f4f921f63f49c53/common/script/content/eggs.js#L6

``` js
title: t('achievementDilatory'),
```

Then to render the translation you'd do: `dilatoryQuest.title(language)`

common/script/libs/achievements.js
```diff
@@ -0,0 +1,322 @@
...
+    index: index++,
+    optionalCount: data.optionalCount,
+  };
+}
+
+function _addSimpleWithCustomPath (result, user, data) {
+  let value = user;
+
+  // traverse the custom path into the user object
+  data.path.split('.').forEach(pathSeg => {
```
crookedneighbor: Why not use `_.get` here?

crookedneighbor (changes requested):
Left a bunch of comments. Let me know what you think!

common/script/libs/achievements.js
crookedneighbor: > Adding the achievements to the content api is a good choice, but now the achievements result just has some string-keys not the actual values (like every other property in the content), I guess they should have the localized values right? @crookedneighbor

Yeah, I just made a note about that when I reviewed it.

> Any 3rd party application would still rather use the /members/{mid}/achievements api call to get the achievements, instead of enumerating through the content array and trying to recreate the achievements client-wise.

Correct. Though you could use the content API to enumerate the available achievements without having them be attached to a user.

common/script/content/achievements.js
khipkin: Sure, I can do that.

website/server/controllers/api-v3/members.js
khipkin: Sure! Wasn't done previously because the exact stuff the API would return hadn't been decided yet.

common/script/libs/achievements.js
khipkin: True.  Not thread safe... so should I manually assign each index?  The achievements need a way to be sorted consistently (aka in the order in which they're added).

common/script/content/achievements.js
khipkin: functional programming ftw :tada:

common/script/content/achievements.js
khipkin: That constant uses 'wizard' instead of 'mage'.

common/script/content/achievements.js
khipkin: Should I add card types to content/constants.js?

common/script/libs/achievements.js
khipkin: Didn't know it existed!

common/script/libs/achievements.js
khipkin: Or I could just never have the index reset.  Then, even if the method is running more than once at the same time, everything returned will be orderable, the numbers just won't necessarily be sequential.

common/script/libs/achievements.js
khipkin: If the achievements in each category can be an array, then the `index` field isn't needed at all!

khipkin: Note to self: look into this.

common/script/content/achievements.js
khipkin: Note to self: look into this.

common/script/libs/achievements.js
khipkin: I think the logic is already as simplified as it can possibly be. :/

common/script/content/achievements.js
khipkin: This would only work for strings that don't take a `{ some: 'thing' }` parameter.

khipkin:
@crookedneighbor I think this is ready for you to take another look.

lemoness:
Q1: "Some of the achievement descriptions contain links to pages with additional information. How can we make the popovers stick around long enough for users to click on the links? Maybe clicking on the achievement icon makes its popover persist until clicked again?"

I think that's a good stopgap for now! We can revisit that question during the website redesign.

Q2: " Should the formatting of the popover text change for achieved vs unachieved achievements? For example: previously, unachieved titles/descriptions were previously muted, meaning the text was gray. Maybe that could still be the case, with the text in the popover being muted in addition to the dark "unearned" badge? (See discussion on #2893 for examples and screenshots.)"

Ideally, the text would change to reflect what the user must do. Example:

"Joined a Party with another person! Have fun battling monsters and supporting each other" 
can become
"To earn this badge, join a Party with another person! Have fun battling monsters and supporting each other."

The changes should be pretty straightforward but I'm happy to look over any individual strings you have questions about if you tag me.

If that's tricky from a technical perspective, we can just mute the text.

Q3: "Some achievements (notably the Contributor and Backer achievements) previously used colored spans in the achievement titles to highlight the special status of the user's contributions. Should these colors be used somehow in the new layout? If so, how?"

I'm not sure that's worth it, unless there's a straightforward way to tackle it. That's another question that can be revisited in the redesign if needed.

khipkin:
Thanks, @lemoness! I've updated the PR description with your (paraphrased) answers to the UI questions, and I'll try to address them at some point this week.  Also, thanks @mandyzhou for your input!

lemoness:
Also, just saw the question about the "Helped Habit Grow" badge -- it
should be "Helped Habitica Grow."

lemoness:
@khipkin I've just found out that this PR is a blocker for getting achievements on Android, so for now, don't worry about the edited strings. Just gray them out, and then we'll make a separate ticket to change the strings later.

khipkin:
@lemoness Okay, sounds good! I'll take care of that tomorrow.

khipkin:
I spent forever trying to figure out how to get the popover title to respect the html I passed into it, to no avail. Eventually, I settled on getting rid of the title attribute altogether and just passing the title text (and html) as part of the popover content instead.

**Here's an earned achievement:**
image

**And here's an unearned achievement:**
image

Edit: **Here's an old popover for comparison:**
image

khipkin:
I haven't forgotten about this!  I've been super busy with school and traveling for job interviews.  Also, I got a new computer, so I'm working on getting my local install running (_again_).  I expect to get to this during the next few days (for real this time). 😁

website/common/script/libs/achievements.js
```diff
@@ -0,0 +1,306 @@
...
+  let value = user.achievements[data.path] || 0;
+
+  let key = data.key || data.path;
+  let thisContent = achievsContent[key];
+
+  let titleKey;
+  let textKey;
+  // If value === 0, use singular versions of strings.
+  // If value !== 0, use plural versions of strings.
+  if (value === 0) {
```
khipkin: @negue I'm looking at this again, and I think this should actually be `=== 1`.

That way, both zero count and multiple count use the plural versions of strings, and ONLY single count uses the singular versions:
- 0 thing`s`
- 1 thing`_`
- 2+ thing`s`

website/server/controllers/api-v3/members.js
```diff
@@ -58,6 +59,58 @@ api.getMember = {
...
+ * @apiSuccess {String} data.*.achievements.title The localized title string
+ * @apiSuccess {String} data.*.achievements.text The localized description string
+ * @apiSuccess {Boolean} data.*.achievements.earned Whether the user has earned the achievement
+ * @apiSuccess {Number} data.*.achievements.index The unique index assigned to the achievement (only for sorting purposes)
+ * @apiSuccess {Anything} data.*.achievements.value The value related to the achievement (if applicable)
+ * @apiSuccess {Number} data.*.achievements.optionalCount The count related to the achievement (if applicable)
+ *
+ * @apiError MemberIdRequired The `id` param is required and must be a valid `UUID`
+ * @apiError UserWithIdNotFound The `id` param did not belong to an existing member
+ */
```
khipkin, line 143: @negue This API route documentation should be updated to be in line with the standards established in #8087.

negue:
Pushed my changes @khipkin @crookedneighbor 

Don't know if I need to add more to the apidocs, any opinions?

website/views/shared/profiles/achievements.jade
```diff
@@ -1,5 +1,8 @@
 mixin simpleAchiev(achiev)
-  div(ng-init='earnedClass = achiev.earned ? "" : "muted"')&attributes(attributes)
+  div(ng-init='earnedClass = achiev.earned ? "" : "muted"', 
+      data-popover-html='<div class="{{earnedClass}}">{{achiev.title}}<hr>{{achiev.text}}</div>',
```
khipkin: Suggestion: pull this string into a var

khipkin: Also, should probably use the same popover placement as below, otherwise the popover might jump around when clicked vs hovered over?

website/server/controllers/api-v3/members.js
```diff
@@ -82,9 +83,34 @@ api.getMember = {
  * @apiSuccess {Number} data.*.achievements.index The unique index assigned to the achievement (only for sorting purposes)
  * @apiSuccess {Anything} data.*.achievements.value The value related to the achievement (if applicable)
  * @apiSuccess {Number} data.*.achievements.optionalCount The count related to the achievement (if applicable)
+
+ * @apiSuccessExample {json} Successful Response
```
khipkin, line 86: All of this would be nested within one of the achievement categories, as described above.

website/views/shared/profiles/achievements.jade
khipkin: FYI: GitHub is hiding this, but it's still relevant.

website/server/controllers/api-v3/members.js
khipkin: FYI: GitHub is hiding this, but it's still relevant.

negue: I kinda, can't follow you on this one..., should the example have ALL the achievemts? or 2 achievements per category?

website/views/shared/profiles/achievements.jade
negue: I had the popover on the same placement, but the clicked one is about 3 pixel lower than the "mouseenter", but I could change it back

khipkin: Can you please post screenshots? 🙏

website/server/controllers/api-v3/members.js
khipkin: I think the example should have the full depth of the actual response that would be returned, but not the full breadth.  Just a few entries would be enough.  Might not even need to do all the categories.  But I think omitting levels in the response object is potentially very misleading.

website/views/shared/profiles/achievements.jade
negue: Clicked and mouseenter popover

habitica__achievements_-_google_chrome_2016-11-07_18-49-03

website/server/controllers/api-v3/members.js
negue: I kinda forgot to add "basic", "seasonal" and "special" to the example 😅 , what about now?

khipkin:
Thanks, @negue!

@crookedneighbor This is ready for you to take another look!

website/server/controllers/api-v3/members.js
khipkin: Beautiful!

khipkin:
Ping @Alys @crookedneighbor @lemoness

khipkin:
Meow meow meow. 😸

khipkin:
YIPPEE! Thanks, @negue and @TheHollidayInn!!! Super pumped this finally got merged in. 🎉 🎉 🎉 😹

khipkin:
Also reminder that the sprites have to be recompiled!

TheHollidayInn:
Ah, @SabreCat should I submit a PR with recompiled sprites then?

negue:
@crookedneighbor @TheHollidayInn don't forget to publish the many new images to s3 😁

khipkin:
@TheHollidayInn My Hall of Heroes blurb has messed up formatting in the link markdown. 😱 Also it says tier 5 but I'm still that pretty tangerine color.

SabreCat:
@TheHollidayInn Compiled sprites will go up with the next deploy, planned for as soon as iOS review gets its act in gear!

SabreCat:
@khipkin: Hall of Heroes issues fixed! Welcome to tier 5.

TheHollidayInn:
@SabreCat Thanks!

PR #7913, opened 2016-08-19
Update faq.json - description of mechanic outdated

luveluen:
Update faq.json - description of mechanic outdated
Changed line 30
--->Click on a type of food, and then select the pet you want to feed!

because clicking on the pet first does not work anymore, clicking on the pet first selects it for use.
As described by MeikeMertsch here: https://www.transifex.com/habitrpg/habitrpg/translate/#de/faqjson/60117132

second commit removed changes for iOS because I can not confirm how the mechanic works in the app. If the problem exists for the app, then it has to be resolved separately.

no need to give me any contrib levels anymore, thank you, I have been at this for years. =P
be0d25d1-9e71-400d-9d78-7c06647f419f


Alys:
@luveluen Thank you! That's gone unnoticed for ages! The current text for iOS is correct as it is.

I have noted this towards your next tier. ;)

luveluen:
@Alys All credit for spotting it goes to MeikeMertsch, one of the newer reviewers =)

PR #7924, opened 2016-08-24
allows leader of challenge to create new challenge tasks even when not a participant

### LINKED ISSUE #7918: new tasks cannot be added to an existing challenge
When a challenge has already been created and saved, it is not possible to add new tasks to it (it should be possible). When you try, you see "Challenge not found or you don't have access." and the console reports:
`POST https://habitica.com/api/v3/tasks/challenge/0b63cb69-8942-4a17-a506-9c3e4855616e?userV=13904 404 (Not Found)`
That ID is the ID of the challenge.

Hus274:
Fixes #7918 
### Changes

Allows the leader of a challenge to add new tasks to the challenge even if they are not participating in the challenge.

---

UUID:

### DIFF
--- a/test/api/v3/integration/tasks/challenges/POST-tasks_challenge_id.test.js
+++ b/test/api/v3/integration/tasks/challenges/POST-tasks_challenge_id.test.js
@@ -33,19 +33,14 @@ describe('POST /tasks/challenge/:challengeId', () => {
     });
   });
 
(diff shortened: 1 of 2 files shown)


crookedneighbor:
Made a very minor change to your test: 2308e14d3e3bfa212ae45164c67eea1d556511e7

PR #7930, opened 2016-08-25
Flag improvements

### LINKED ISSUE #7925: API does not allow admin to delete or flag private posts
### Description

Version 3 of the API does not allow administrative users to delete or flag posts found in private groups (such as parties), even if the post has been flagged by a group member. The route returns "Group not found or you don't have access."

The trouble appears to be that we don't pass the `optionalMembership` option to the `getGroup` call used to fetch the group before deleting a message from it. Given there's a check a few lines later that allows only the message creator or an admin to perform the delete, should `optionalMembership: true` be sent here?
#### Console Errors

`{"success":false,"error":"NotFound","message":"Group not found or you don't have access."}`

crookedneighbor:
Closes #7925

@SabreCat

Also fleshed out the flag docs:

screen shot 2016-08-24 at 9 34 21 pm

screen shot 2016-08-24 at 9 34 29 pm

### DIFF
--- a/website/server/controllers/api-v3/chat.js
+++ b/website/server/controllers/api-v3/chat.js
@@ -182,7 +182,11 @@ api.flagChat = {
     let validationErrors = req.validationErrors();
     if (validationErrors) throw validationErrors;
 
-    let group = await Group.getGroup({user, groupId});
+    let group = await Group.getGroup({
+      user,
+      groupId,
+      optionalMembership: user.contributor.admin,
+    });
     if (!group) throw new NotFound(res.t('groupNotFound'));
     let message = _.find(group.chat, {id: req.params.chatId});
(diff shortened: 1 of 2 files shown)


website/server/controllers/api-v3/chat.js
```diff
@@ -182,7 +195,11 @@ api.flagChat = {
     let validationErrors = req.validationErrors();
     if (validationErrors) throw validationErrors;
 
-    let group = await Group.getGroup({user, groupId});
+    let group = await Group.getGroup({
+      user,
+      groupId,
+      optionalMembership: user.contributor.admin,
```
SabreCat, line 201: Oh! That's a clever way of going about it. Excellent.

PR #7934, opened 2016-08-26
GP will not be deducted if health is max

ujjwalgulecha:
https://github.com/HabitRPG/habitrpg/issues/6384
### Changes

Now if the user tries to buy potion if he/she has max health, it will not deduct any GP

Note: Need to update message.json for all languages. 

---

UUID: 
2d7c17ad-3f7e-4783-b697-b77820fff31d

### DIFF
--- a/common/script/ops/buyHealthPotion.js
+++ b/common/script/ops/buyHealthPotion.js
@@ -15,6 +15,10 @@ module.exports = function buyHealthPotion (user, req = {}, analytics) {
     throw new NotAuthorized(i18n.t('cannotBuyItem', req.language));
   }
 
+  if (user.stats.hp >= 50) {
+    throw new NotAuthorized(i18n.t('messageHealthAlreadyMax', req.language));
+  }
+
   user.stats.hp += 15;
   if (user.stats.hp > 50) {
     user.stats.hp = 50;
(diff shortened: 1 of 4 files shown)


crookedneighbor:
I would expect this to return an error instead of just nooping and sending back a message. Anyone have thoughts on that?

Please remove the change you made to en_GB. Translations are managed by Transifex.

SabreCat:
Agreed. We `throw` for the case of trying to buy the potion when you don't have enough Gold; this seems like a similar situation.

ujjwalgulecha:
I have addressed the changes. It throws an error now. I have removed translation for en_gb as well.

common/script/ops/buyHealthPotion.js
```diff
@@ -15,6 +15,10 @@ module.exports = function buyHealthPotion (user, req = {}, analytics) {
     throw new NotAuthorized(i18n.t('cannotBuyItem', req.language));
   }
 
+  if (user.stats.hp >= 50) {
+    throw new NotAuthorized(i18n.t('messageHealthAlreadyMax', req.language));
```
crookedneighbor, line 19: Let's do BadRequest instead of NotAuthorized.

crookedneighbor:
Other than the comment I just left, looks good!

crookedneighbor:
Actually, looks like you were just following the error pattern we have set up. I think we should revisit what errors we're sending down, but that doesn't have to be done in this PR.

crookedneighbor:

PR #7938, opened 2016-08-27
Send flags to slack

crookedneighbor:
When a message is flagged, we send an email to the moderators. This adds functionality to send a slack message as well.

The method will noop if no slack url is set in the config.json.

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -74,5 +74,9 @@
         "APP_ID": "appId",
         "KEY": "key",
         "SECRET": "secret"
+    },
+    "SLACK": {
+        "FLAGGING_URL": "https://hooks.slack.com/services/id/id/id",
+        "FLAGGING_FOOTER_LINK": "https://crookedneighbor.github.io/flag-o-rama/"
     }
 }
--- a/config.json.example
+++ b/config.json.example
@@ -77,6 +77,6 @@
     },
     "SLACK": {
         "FLAGGING_URL": "https://hooks.slack.com/services/id/id/id",
-        "FLAGGING_FOOTER_LINK": "https://crookedneighbor.github.io/flag-o-rama/"
+        "FLAGGING_FOOTER_LINK": "https://habitrpg.github.io/flag-o-rama/"
     }
 }
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "version": "3.37.0",
   "main": "./website/server/index.js",
   "dependencies": {
+    "@slack/client": "slackhq/node-slack-sdk#2ee794cd31326c54f38c518eef2b9d223327d939",
     "accepts": "^1.3.2",
     "amazon-payments": "0.0.4",
     "amplitude": "^2.0.3",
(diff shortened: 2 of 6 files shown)


config.json.example
```diff
@@ -74,5 +74,9 @@
         "APP_ID": "appId",
         "KEY": "key",
         "SECRET": "secret"
+    },
+    "SLACK": {
+        "FLAGGING_URL": "https://hooks.slack.com/services/id/id/id",
+        "FLAGGING_FOOTER_LINK": "https://crookedneighbor.github.io/flag-o-rama/"
```
crookedneighbor: This is a small webpage I created to make flagging easier. 

The slack message will contain a link to this page with query params for the group id and the chat id. If the mod has entered their api details (creds saved in localstorage), it will automatically flag them. Now that it's possible to flag and unflag messages in private groups, this will make it easier to flag inappropriate content that has been flagged once in a private group.

If we decide to merge this in, I'll copy this repo over to the HabitRPG org.

package.json
```diff
@@ -4,6 +4,7 @@
   "version": "3.34.1",
   "main": "./website/server/index.js",
   "dependencies": {
+    "@slack/client": "slackhq/node-slack-sdk#2ee794cd31326c54f38c518eef2b9d223327d939",
```
crookedneighbor, line 7: I [wrote the Incoming Webhook feature](https://github.com/slackhq/node-slack-sdk/pull/232) for this module, but it hasn't been published to npm yet. When it is, we can switch this out with a version number instead.

We'll need to update shrinkwrap.

website/server/libs/slack.js
```diff
@@ -0,0 +1,58 @@
...
+import nconf from 'nconf';
+
+const SLACK_FLAGGING_URL = nconf.get('SLACK:FLAGGING_URL');
+const SLACK_FLAGGING_FOOTER_LINK = nconf.get('SLACK:FLAGGING_FOOTER_LINK');
+const BASE_URL = nconf.get('BASE_URL');
+
+let flagSlack;
+
+try {
+  flagSlack = new IncomingWebhook(SLACK_FLAGGING_URL);
```
crookedneighbor, line 14: This error catching is so the app doesn't crash if someone forgets to set the slack flagging url.

We could alternatively set the channel and bot info in the send call, rather than relying on the default values set for the slack endpoint.

config.json.example
crookedneighbor: Should these be nested?

``` json
"SLACK": {
    "FLAGGING": {
        "WEBHOOK_ENDPOINT": "https://hooks.slack.com/services/id/id/id",
        "FOOTER_LINK": "https://crookedneighbor.github.io/flag-o-rama/"
    }
}
```

crookedneighbor:
screen shot 2016-08-26 at 11 39 51 pm

crookedneighbor:
The travis build failure was do to the logger tests. I've rewritten the logger tests to not cause side effects in other tests. 

Going to look into adding a method to requireAgain that will clean up it's state after it's done to prevent this kind of leakage.

test/api/v3/unit/libs/logger.js
```diff
@@ -1,57 +1,174 @@
 import winston from 'winston';
-import requireAgain from 'require-again';
+import logger from '../../../../../website/server/libs/logger';
+import {
+  NotFound,
+} from '../../../../../website/server/libs//errors';
```
khipkin, line 5: Extra "/"

crookedneighbor: Doh.

PR #7941, opened 2016-08-27
When blocking a user, make them rest at the inn - fixes #7935

### LINKED ISSUE #7935: blocking a user should cause them to rest in the inn
The [API](https://habitica.com/apidoc/) (see also [on the wiki](http://habitica.wikia.com/wiki/Application_Programming_Interface)) has a route for blocking a user, which prevents them using Habitica in any way. The route should also cause the user to be resting in the inn ("sleeping").  (EDIT: the API `block` route is not related to this issue. See one of my comments below for details.)

Unblocking the user should NOT stop them resting in the inn. 

This is being done so that non-malicious users who need to be blocked for various reasons do not suffer unfair damage when they are later unblocked.

gambtho:
Fixes https://github.com/HabitRPG/habitrpg/issues/7935
### Changes

Blocking a user will make the user rest at an inn, the user will remain resting when unblocked

---

### DIFF
--- a/test/api/v3/integration/hall/PUT-hall_heores_heroId.test.js
+++ b/test/api/v3/integration/hall/PUT-hall_heores_heroId.test.js
@@ -68,6 +68,7 @@ describe('PUT /heroes/:heroId', () => {
     expect(hero.contributor.level).to.equal(1);
     expect(hero.purchased.ads).to.equal(true);
     expect(hero.auth.blocked).to.equal(true);
+    expect(hero.preferences.sleep).to.equal(true);
     expect(hero.notifications.length).to.equal(1);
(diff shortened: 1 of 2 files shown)


test/api/v3/integration/hall/PUT-hall_heores_heroId.test.js
```diff
@@ -62,12 +62,14 @@ describe('PUT /heroes/:heroId', () => {
     expect(heroRes.contributor.level).to.equal(1);
     expect(heroRes.purchased.ads).to.equal(true);
     expect(heroRes.auth.blocked).to.equal(true);
+    expect(heroRes.preferences.sleep).to.equal(true);
```
crookedneighbor: Pretty sure this data isn't passed back in the response, so this expectation shouldn't be here.

crookedneighbor:

Issue #7922, opened 2016-08-24
show quest progress notification when completing task


Alys:
If the player is participating in an active boss quest (i.e., not while the quest is in the invitation stage), when a Daily or To-Do is completed or a positive Habit is clicked, there should be a notification of how much boss damage has been just been obtained (from `party.quest.progress.up`). Suggested wording: "+12.3 damage to boss"

When in an active collection quest, if the player has just found a quest drop (`party.quest.progress.collectedItems`), there should be a notification of that. Suggested wording: "+1 quest item found".

We need to decide on the colour of the notification. If no other suggestions are made, I suggest using the same green as for the drop notifications.

This feature was requested by @lemoness

lemoness:
This is linked to this issue as well: https://github.com/HabitRPG/habitrpg/issues/7928

tricksyfox:
Can the item notifications also show the item image, like other drops do?
What colors do we have in use right now?

Green - Item Drop
Yellow - gold/silver
Purple - XP
Red - Health
Blue - Mana

...So maybe quest stuff should be orange?

Alys:
Orange is typically an indicator of errors, warnings, or other bad/semi-bad things so it's not ideal for a positive message.

gambtho:
leave it as is -- or?

khipkin:
Suggestion: [help welcome now] --> [in progress]

Alys:
This was implemented in https://github.com/HabitRPG/habitica/pull/7951

PR #7951, opened 2016-08-31
Shows quest progress notification when completing task #7922

gambtho:
Feature https://github.com/HabitRPG/habitrpg/issues/7922
### Changes

Shows quest progress notification when completing task #7922

### DIFF
--- a/test/common/ops/scoreTask.test.js
+++ b/test/common/ops/scoreTask.test.js
@@ -142,6 +142,21 @@ describe('shared.ops.scoreTask', () => {
       expect(ref.beforeUser._id).to.eql(ref.afterUser._id);
     });
 
+    it('and increments quest progress', () => {
+      expect(ref.afterUser.party.quest.progress.up).to.eql(0);
+      ref.afterUser.party.quest.key = 'gryphon';
+
+      scoreTask({ user: ref.afterUser, task: habit, direction: 'up', cron: false });
+      let firstTaskDelta = ref.afterUser.party.quest.progress.up;
+      expect(firstTaskDelta).to.be.greaterThan(0);
+      expect(ref.afterUser._tmp.quest.progressDelta).to.eql(firstTaskDelta);
+
+      scoreTask({ user: ref.afterUser, task: habit, direction: 'up', cron: false });
+      let secondTaskDelta = ref.afterUser.party.quest.progress.up - firstTaskDelta;
+      expect(secondTaskDelta).to.be.greaterThan(0);
+      expect(ref.afterUser._tmp.quest.progressDelta).to.eql(secondTaskDelta);
+    });
+
     context('habits', () => {
       it('up', () => {
         options = { user: ref.afterUser, task: habit, direction: 'up', times: 5, cron: false };
--- a/website/client-old/js/services/userServices.js
+++ b/website/client-old/js/services/userServices.js
@@ -247,11 +247,22 @@ angular.module('habitrpg')
             var tmp = res.data.data._tmp || {}; // used to notify drops, critical hits and other bonuses
             var crit = tmp.crit;
             var drop = tmp.drop;
+            var quest = tmp.quest;
 
             if (crit) {
               var critBonus = crit * 100 - 100;
               Notification.crit(critBonus);
             }
+
+            if (quest && user.party.quest) {
+              var userQuest = Content.quests[user.party.quest.key];
+              if (quest.progressDelta && userQuest.boss) {
+                Notification.quest('questDamage', quest.progressDelta.toFixed(1));
+              } else if (quest.collection && userQuest.collect) {
+                Notification.quest('questCollection', quest.collection);
+              }
+            }
+
             if (drop) {
               var text, notes, type;
               $rootScope.playSound('Item_Drop');
--- a/test/common/fns/randomDrop.test.js
+++ b/test/common/fns/randomDrop.test.js
@@ -29,8 +29,10 @@ describe('common.fns.randomDrop', () => {
     };
     randomDrop(user, { task, predictableRandom });
     expect(user.party.quest.progress.collectedItems).to.eql(1);
+    expect(user._tmp.quest.collection).to.eql(1);
     randomDrop(user, { task, predictableRandom });
     expect(user.party.quest.progress.collectedItems).to.eql(2);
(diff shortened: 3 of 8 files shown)


gambtho:
sorry for the pull request spam....'let' doesn't work so well in the client code 👎 

this is ready for review

gambtho:
es6 habits burn me again -- validated the full test suite locally this time

crookedneighbor:
We'll be using Babel on the client code soon enough, then you won't have to worry about that. :)

gambtho:
any concerns on this one? @crookedneighbor ?

common/locales/en/quests.json
```diff
@@ -26,6 +26,8 @@
     "pending": "Pending",
     "questStart": "Once all members have either accepted or rejected, the quest begins. Only those that clicked \"accept\" will be able to participate in the quest and receive the drops. If members are pending too long (inactive?), the quest owner can start the quest without them by clicking \"Begin\". The quest owner can also cancel the quest and regain the quest scroll by clicking \"Cancel\".",
     "questStartBroken": "Once all members have either accepted or rejected, the quest begins... Only those that clicked \"accept\" will be able to participate in the quest and receive the drops... If members are pending too long (inactive?), the quest owner can start the quest without them by clicking \"Begin\"... The quest owner can also cancel the quest and regain the quest scroll by clicking \"Cancel\"...",
+    "questCollection": "+<%= val %> quest item found",
```
khipkin: items

gambtho: not sure on this one -- you only get one collection drop at a time?

Alys: Currently, yes, just one at a time. That might change in future but if it does, then we should introduce programmatic pluralisation. One of our users will be sure to remind us if we forget. :) "item" is good for now.

@khipkin It definitely seems wrong though at first glance, because of the variable. :)

website/client/js/services/userServices.js
```diff
@@ -247,11 +247,23 @@ angular.module('habitrpg')
...
             if (crit) {
               var critBonus = crit * 100 - 100;
               Notification.crit(critBonus);
             }
+
+            if (quest && user.party.quest) {
+              var userQuest = Content.quests[ user.party.quest.key ];
+              if (quest.progressDelta && userQuest.boss) {
+                Notification.quest('questDamage', quest.progressDelta.toPrecision(2));
+              }
```
crookedneighbor: No line break here.

``` js
} else if (quest.collection && userQuest.collect) {
```

common/script/ops/scoreTask.js
```diff
@@ -150,14 +150,17 @@ function _changeTaskValue (user, task, direction, times, cron) {
...
+        let prevProgress = user.party.quest.progress.up;
 
         if (task.type === 'todo' || task.type === 'daily') {
           user.party.quest.progress.up += nextDelta * (1 + user._statsComputed.str / 200);
         } else if (task.type === 'habit') {
           user.party.quest.progress.up += nextDelta * (0.5 + user._statsComputed.str / 400);
         }
-      }
 
+        if (!user._tmp.quest) user._tmp.quest = {};
```
crookedneighbor: This should be reflected in the score task tests.

common/script/fns/randomDrop.js
```diff
@@ -39,6 +39,8 @@ module.exports = function randomDrop (user, options, req = {}) {
   if (predictableRandom(user, user.stats.gp) < chance) {
     if (!user.party.quest.progress.collectedItems) user.party.quest.progress.collectedItems = 0;
     user.party.quest.progress.collectedItems++;
+    if (!user._tmp.quest) user._tmp.quest = {};
```
crookedneighbor: This should be reflected in the randomDrop tests.

common/locales/en/quests.json
khipkin: Oh okay!  But if `val` can only ever be 1, why not just use 1 in the string?

gambtho: in case it changes in the future, i went back and forth on that, and left it like this :)

could make it "item(s)" I guess?

gambtho:
pushed updated version with requested changes/tests

gambtho:
pushed version merged with other recent changes, please let me know if anything else is needed

test/common/ops/scoreTask.test.js
```diff
@@ -121,6 +121,30 @@ describe('shared.ops.scoreTask', () => {
     });
   });
 
+  describe('increments', () => {
```
crookedneighbor: This should just be an it block.

``` js
it('increments quest progress', () => {
  // test
});
```

website/client-old/js/services/userServices.js
```diff
@@ -247,11 +247,22 @@ angular.module('habitrpg')
...
             var drop = tmp.drop;
+            var quest = tmp.quest;
 
             if (crit) {
               var critBonus = crit * 100 - 100;
               Notification.crit(critBonus);
             }
+
+            if (quest && user.party.quest) {
+              var userQuest = Content.quests[ user.party.quest.key ];
```
crookedneighbor: no need for the spaces surrounding `user.party.quest.key`

website/client-old/js/services/userServices.js
```diff
@@ -247,11 +247,22 @@ angular.module('habitrpg')
...
+              if (quest.progressDelta && userQuest.boss) {
+                Notification.quest('questDamage', quest.progressDelta.toPrecision(2));
```
crookedneighbor: Let's match what's used in the marker in the menu bar. https://github.com/HabitRPG/habitrpg/blob/fefcb0f4aca56379a9734aec485ca9e8827c1ea0/website/client-old/js/controllers/menuCtrl.js#L56

We changed it from `toPrecision` to `toFixed` so that it wouldn't show as scientific notation.

crookedneighbor (changes requested):
There should probably also be an alert box when using a skill that damages the boss.

crookedneighbor:
Left some nitpicky comments. The only real blocking one I think is that skills that damage the boss should also display a notification (people will assume it's a bug if they don't)

gambtho:
ok -- fixed the issues mentioned, and added a notification that displays when skills damage the boss

crookedneighbor:
Thanks! Notated this toward your next tier.

PR #7957, opened 2016-08-31
Create .editorconfig

librarianmage:
Fixes https://github.com/HabitRPG/habitrpg/issues/4376
### Changes

Adds an `.editorconfig` file

---

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,15 +1,10 @@
 # top level config
 root = true
 
-# All files:
-# utf-8
-# Unix-style newlines
-# Ensure end-of-file newline
-# Soft indents (2 spaces)
-# Trim trailing whitespace
+# all files
 [*]
 charset = utf-8
-end_of_line = lf
+end_of_line = lf # Unix-style
 insert_final_newline = true
 indent_style = space
 indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,7 +10,7 @@ root = true
 [*]
 charset = utf-8
 end_of_line = lf
-insert_final_newline = false
+insert_final_newline = true
 indent_style = space
 indent_size = 2
 trim_trailing_whitespace
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,11 +6,11 @@ root = true
 # Unix-style newlines
 # Remove end-of-file newlines
 # Soft indents (2 spaces)
-# Trim tailing whitespace
+# Don't trim trailing whitespace
 [*]
 charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 indent_style = space
 indent_size = 2
-trim_trailing_whitespace
+trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,8 +14,3 @@ insert_final_newline = true
 indent_style = space
 indent_size = 2
 trim_trailing_whitespace = true
-
-# Jade/Pug files:
-# Don't trim trailing whitespace
-[*.jade]
-trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,4 +13,4 @@ end_of_line = lf
 insert_final_newline = true
 indent_style = space
 indent_size = 2
-trim_trailing_whitespace = false
+trim_trailing_whitespace = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# top level config
+root = true
+
+# All files:
+# utf-8
+# Unix-style newlines
+# Remove end-of-file newlines
+# Soft indents (2 spaces)
+# Trim tailing whitespace
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = false
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,9 +4,9 @@ root = true
 # All files:
 # utf-8
 # Unix-style newlines
-# Remove end-of-file newlines
+# Ensure end-of-file newline
 # Soft indents (2 spaces)
-# Don't trim trailing whitespace
+# Trim trailing whitespace
 [*]
 charset = utf-8
 end_of_line = lf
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+# All files:
+# utf-8
+# Unix-style newlines
+# Remove end-of-file newlines
+# Soft indents (2 spaces)
+# Trim tailing whitespace
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = false
```
khipkin: I don't think this is standardized for the repo.

crookedneighbor: ^ only because we don't have an editor config to enforce a standard. This is what I prefer.

khipkin: My vote goes the other way. :stuck_out_tongue_closed_eyes:

crookedneighbor: Actually, I reviewed the editorconfig I usually use and I go back on what I said. I prefer true as well 😄

librarianmage: Updated

.editorconfig
```diff
@@ -0,0 +1,16 @@
+# top level config
+root = true
+
+# All files:
+# utf-8
+# Unix-style newlines
+# Remove end-of-file newlines
+# Soft indents (2 spaces)
+# Trim tailing whitespace
```
paglias: trim trailing (missing r)

.editorconfig
```diff
@@ -6,11 +6,11 @@ root = true
...
-# Trim tailing whitespace
+# Don't trim trailing whitespace
 [*]
 charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 indent_style = space
 indent_size = 2
-trim_trailing_whitespace
+trim_trailing_whitespace = false
```
khipkin: why false?

librarianmage: Nvm, made a mistake (should be true)

.editorconfig
```diff
@@ -0,0 +1,16 @@
+# top level config
+root = true
+
+# All files:
+# utf-8
+# Unix-style newlines
+# Remove end-of-file newlines
```
keysmashes: Given line 13, shouldn't this be something like "ensure file ends with a newline"?

librarianmage: Yes, my mistake

.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+# Soft indents (2 spaces)
+# Don't trim trailing whitespace
```
keysmashes: Disagrees with line 16. Also, it's occasionally useful to be able to end a line with whitespace in Jade/Pug - I don't know if that's been done in this repo, but I've certainly found myself doing it.

crookedneighbor:
I would like to hear from the other maintainers before deciding anything on this. @alys @sabrecat @thehollidayinn @paglias @viirus is this something you'd use? I don't want to include it if no one else would install the editorconfig plugin for their editor.

librarianmage:
Should we also advocate for text editors with .editorconfig plugins on the
wiki? Then there is a higher chance of it being useful.

Alys:
I might use this myself if my editor supports it (I haven't checked that yet), but regardless, I don't think it's a bad thing to have. If nothing else, we can say to contributors, "our coding style is described in [this file] which you can read even if your editor doesn't support it". It might save some discussions about what's the appropriate style.

I'd be cautious about advocating on the wiki for specific editors. I don't want contributors to feel they are obliged to use them. We could however say something like "There's a .editorconfig file to help with consistency and it will work in some editors (e.g., Abc, Def, etc)."

.editorconfig
Alys: @anowlcalledjosh Can you explain the circumstances under which it would be useful to end a jade/pug line with whitespace? If we do allow that, we'll want to document why on the wiki so that we don't get questions about the inconsistency.

paglias:
I'd use if for sure! It basically matches my own coding style for personal projects so my text editor is already configured but I'll install it in case we change something for Habitica

.editorconfig
keysmashes: @Alys I can't find the code I was thinking of when I wrote that, but I suspect I meant something like this:

``` pug
| Here is some text with a
a(href="#") link
| in it
```

which renders as `Here is some text with alinkin it`, and needs a space at the end of the first line and at the start of the third line to render properly. However, the proper way to do the above is

``` pug
| Here is some text with a #[a(href="#") link] in it
```

and Habit barely uses inline text anyway, so maybe the special case for Jade/Pug files shouldn't be included. If it turns out that having trailing whitespace is useful after all, we can always change the editorconfig.

Alys: Ah yes, all our text should be using locales strings, with variables when necessary, so this wouldn't occur, at least in an ideal world. :) I think there's still some text that's not translated but if we ran into that problem with whitespace, we'd fix it by making it translatable. Thanks for the explanation!

@MathWhiz Sorry, could you please remove the exception for jade/pug?

SabreCat:
I'm fine with adding it. I like @Alys's take--it is basically human-readable, so it's almost like having a little code style README even if you don't plug it in anywhere.

librarianmage:
@Alys We can also link to the [.editorconfig download list](http://editorconfig.org/#download), which has a list of the editors that they know work with it defaultly, and a list of ones they know have plugins for it.

librarianmage:
Pinging

This should be done, although if anyone thinks there are any exceptions, please let me know!

.editorconfig
```diff
@@ -0,0 +1,16 @@
+# top level config
+root = true
+
+# All files:
+# utf-8
+# Unix-style newlines
+# Ensure end-of-file newline
+# Soft indents (2 spaces)
+# Trim trailing whitespace
```
Alys: Is it normal in .editorconfig files to have comments that say the same thing as the config itself, and that are separated this much from the associated config lines? This seems like the sort of situation where the comments and code could get out of sync really easily.

I'm not sure the comments are necessary, except perhaps `# Unix-style newlines` which to a new coder might not be obvious from `end_of_line = lf` (although a new coder should read the editorconfig docs if they are uncertain about anything they see here, so perhaps even that comment isn't needed).

The `# top level config` and `# All files` comments could stay as subheadings.

crookedneighbor: Agreed.

librarianmage:
Ping to reviewers

crookedneighbor:

librarianmage:
🎉 Yay!

PR #7965, opened 2016-09-01
delete unread messages when a user leaves a group - fixes 7955

gambtho:
https://github.com/HabitRPG/habitrpg/issues/7955
### Changes

Removes unread group/party messages when a user leaves, or is removed from a group.   This includes when the final user leaves the group, preventing the issue described in 7955.   Also, added a migration script to remove any existing null messages.

---

### DIFF
--- a/test/api/v3/integration/groups/POST-groups_groupId_leave.js
+++ b/test/api/v3/integration/groups/POST-groups_groupId_leave.js
@@ -65,6 +65,21 @@ describe('POST /groups/:groupId/leave', () => {
         expect(groupToLeave.leader).to.equal(member._id);
       });
 
+      it('removes new messages', async () => {
+        await member.post(`/groups/${groupToLeave._id}/chat`, { message: 'Some message' });
+        await member.post(`/groups/${groupToLeave._id}/leave`);
+
+        await groupToLeave.sync();
+        await member.sync();
+        await leader.sync();
+        expect(leader.newMessages[groupToLeave._id]).to.not.be.empty;
+
+        await leader.post(`/groups/${groupToLeave._id}/leave`);
+        await leader.sync();
+        await groupToLeave.sync();
+        expect(leader.newMessages[groupToLeave._id]).to.be.empty;
+      });
+
       context('With challenges', () => {
         let challenge;
 
@@ -122,6 +137,8 @@ describe('POST /groups/:groupId/leave', () => {
         privateGuild = group;
         leader = groupLeader;
         invitedUser = invitees[0];
+
+        await leader.post(`/groups/${group._id}/chat`, { message: 'Some message' });
       });
 
       it('removes a group when the last member leaves', async () => {
--- a/test/api/v3/integration/groups/POST-groups_id_removeMember.test.js
+++ b/test/api/v3/integration/groups/POST-groups_id_removeMember.test.js
@@ -87,6 +87,7 @@ describe('POST /groups/:groupId/removeMember/:memberId', () => {
     let partyLeader;
     let partyInvitedUser;
     let partyMember;
+    let removedMember;
 
     beforeEach(async () => {
       let { group, groupLeader, invitees, members } = await createAndPopulateGroup({
@@ -96,13 +97,14 @@ describe('POST /groups/:groupId/removeMember/:memberId', () => {
           privacy: 'private',
         },
         invites: 1,
-        members: 1,
+        members: 2,
       });
 
       party = group;
       partyLeader = groupLeader;
       partyInvitedUser = invitees[0];
       partyMember = members[0];
+      removedMember = members[1];
     });
 
     it('can remove other members', async () => {
@@ -127,6 +129,20 @@ describe('POST /groups/:groupId/removeMember/:memberId', () => {
       let invitedUserWithoutInvite = await partyInvitedUser.get('/user');
 
       expect(invitedUserWithoutInvite.invitations.party).to.be.empty;
+    });
+
+    it('removes new messages from a member who is removed', async () => {
+      await partyLeader.post(`/groups/${party._id}/chat`, { message: 'Some message' });
+      await partyLeader.sync();
+      await removedMember.sync();
+      await party.sync();
+      expect(removedMember.newMessages[party._id]).to.not.be.empty;
+
+      await partyLeader.post(`/groups/${party._id}/removeMember/${removedMember._id}`);
+      await party.sync();
+      await removedMember.sync();
+
+      expect(removedMember.newMessages[party._id]).to.be.empty;
     });
 
     it('removes user from quest when removing user from party after quest starts', async () => {
--- a/website/server/controllers/api-v3/groups.js
+++ b/website/server/controllers/api-v3/groups.js
@@ -359,6 +359,13 @@ api.rejectGroupInvite = {
   },
 };
 
+function _removeMessagesFromMember (member, groupId) {
+  if (member.newMessages[groupId]) {
+    delete member.newMessages[groupId];
+    member.markModified('newMessages');
+  }
+}
+
 /**
  * @api {post} /api/v3/groups/:groupId/leave Leave a group
  * @apiVersion 3.0.0
@@ -376,7 +383,6 @@ api.leaveGroup = {
   middlewares: [authWithHeaders()],
   async handler (req, res) {
     let user = res.locals.user;
-
     req.checkParams('groupId', res.t('groupIdRequired')).notEmpty();
     // When removing the user from challenges, should we keep the tasks?
     req.checkQuery('keep', res.t('keepOrRemoveAll')).optional().isIn(['keep-all', 'remove-all']);
@@ -390,7 +396,7 @@ api.leaveGroup = {
       throw new NotFound(res.t('groupNotFound'));
     }
 
-    // During quests, checke wheter user can leave
+    // During quests, check if user can leave
     if (group.type === 'party') {
       if (group.quest && group.quest.leader === user._id) {
         throw new NotAuthorized(res.t('questLeaderCannotLeaveGroup'));
@@ -402,6 +408,14 @@ api.leaveGroup = {
     }
 
     await group.leave(user, req.query.keep);
+
+    _removeMessagesFromMember(user, group._id);
+
+    await Bluebird.all([
+      user.save(),
+      group.save(),
+    ]);
+
     res.respond(200, {});
   },
 };
@@ -493,10 +507,7 @@ api.removeGroupMember = {
         member.party._id = undefined; // TODO remove quest information too? Use group.leave()?
       }
 
-      if (member.newMessages[group._id]) {
-        member.newMessages[group._id] = undefined;
-        member.markModified('newMessages');
-      }
+      _removeMessagesFromMember(member, group._id);
 
       if (group.quest && group.quest.active && group.quest.leader === member._id) {
         member.items.quests[group.quest.key] += 1;


khipkin:
@gambtho You accidentally linked to the wrong issue.

gambtho:
@khipkin yep...thanks for the heads-up 👍

crookedneighbor:
This needs test coverage.

website/server/controllers/api-v3/groups.js
```diff
@@ -494,7 +500,7 @@ api.removeGroupMember = {
       }
 
       if (member.newMessages[group._id]) {
```
crookedneighbor: Seems like you could extract this logic into a function and use it both places:

``` js
function removeMessagesFromMember (member, groupId) {
  if (member.newMessages[groupId]) {
    delete member.newMessages[groupId];
    member.markModified('newMessages');
  }
}
```

website/server/models/group.js
```diff
@@ -741,9 +741,13 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
 
     update.$unset = {[`quest.members.${user._id}`]: 1};
   }
-
   // If user is the last one in group and group is private, delete it
   if (group.memberCount <= 1 && group.privacy === 'private') {
+    // this may be redundant, remove any remaining newMessages for the final user
```
crookedneighbor: I don't think this is necessary. You have it covered in the leave and kick routes.

migrations/20160901_remove_null_notifications.js
```diff
@@ -0,0 +1,13 @@
...
+db.users.find({}).forEach(function(user){
+  var newNewMessages = {};
+
+  _.each(user.newMessages, function(val, key){
+    if(key != "undefined" && val!=null){
+      newNewMessages[key] = val;
+    }
+  });
+
+  db.users.update({_id: user._id}, {$set: {'newMessages': newNewMessages}});
```
crookedneighbor: Wouldn't it be better to `$unset` the values that are null?

gambtho: not super familiar with mongo syntax -- but didn't think i could unset a specific element in an array?   i see how to unset all of the new messages, or how to set the value of the null message to an empty string.  was reading through this  http://stackoverflow.com/a/19946487

Alys: @gambtho This command:
`db.users.update({_id: "d904bd62-da08-416b-a816-ba797c9ee265"}, {$unset:{ 'newMessages.4fc50931-0d1c-45bf-b658-077af46c55ad': 1 }},{multi:true});`

will turn this:

```
    "newMessages": {
        "4fc50931-0d1c-45bf-b658-077af46c55ad": null,
        "85dd1be6-336b-4e53-9949-2c09150b4035": {
            "name": "Alys's private guild",
            "value": true
        }
    },
```

into this:

```
    "newMessages": {
        "85dd1be6-336b-4e53-9949-2c09150b4035": {
            "name": "Alys's private guild",
            "value": true
        }
    },
```

gambtho: So, for this specific example? -- 

```
var _ = require('lodash');

db.users.find({}).forEach(function(user){
  _.each(user.newMessages, function(val, key){
    if(!val){
      message = `newMessages.${key}`;
      db.users.update({_id: user._id}, {$unset:{message: 1 }},{multi:true});
    }
  });
});
```

website/server/controllers/api-v3/groups.js
```diff
@@ -402,6 +409,11 @@ api.leaveGroup = {
     }
 
     await group.leave(user, req.query.keep);
+
+    if (user.newMessages[group._id]) {
```
crookedneighbor: Is there a reason to have this if check both here and in the removeMessages function?

crookedneighbor:
Let me know if you need help with the tests.

website/server/controllers/api-v3/groups.js
gambtho: no :)

gambtho:
thanks -- just got busy last week -- submitting updated version with tests now

website/server/controllers/api-v3/groups.js
crookedneighbor: Is there a good reason to keep the `if check` out of the function?

migrations/20160901_remove_null_notifications.js
```diff
@@ -0,0 +1,10 @@
+var _ = require('lodash');
+
+db.users.find({}).forEach(function(user){
```
crookedneighbor: This migration file looks like it would work for a small database, but I bet it times out running across a database of over a million users. 

Let's just remove this file for now and one of the DB admins will write the migration file.

website/server/controllers/api-v3/groups.js
gambtho: my thinking was that the function removes messages associated with a specific group_id, the if check remaining where it was allows the function to be more generic, and the logic for validating if it's needed is kept at that call site.   prefer i flip the location?

migrations/20160901_remove_null_notifications.js
gambtho: removed, and pushed changes

test/api/v3/integration/groups/POST-groups_id_removeMember.test.js
```diff
@@ -129,6 +132,11 @@ describe('POST /groups/:groupId/removeMember/:memberId', () => {
       expect(invitedUserWithoutInvite.invitations.party).to.be.empty;
     });
 
+    it('removes new messages from a member who is removed', async () => {
+      await partyLeader.post(`/groups/${party._id}/removeMember/${removedMember._id}`);
+      expect(removedMember.newMessages).to.be.empty;
```
crookedneighbor: Let's move the send message logic here. 

You'll need to use the sync method to bring this member up to date. You're asserting on the version of the data before the message was sent.

gambtho: done

website/server/controllers/api-v3/groups.js
crookedneighbor: To me, the only point of the if check is to save a call to markModified if it isn't needed. You could arguably do the logic without the check, and it would make no difference except the call to markModified. 

So let's put the if check in the function.

test/api/v3/integration/groups/POST-groups_id_removeMember.test.js
```diff
@@ -129,6 +131,16 @@ describe('POST /groups/:groupId/removeMember/:memberId', () => {
...
     });
 
+    it('removes new messages from a member who is removed', async () => {
+      await partyLeader.post(`/groups/${party._id}/chat`, { message: 'Some message' });
+      await party.sync();
+
+      await partyLeader.post(`/groups/${party._id}/removeMember/${removedMember._id}`);
+      await party.sync();
+
+      expect(removedMember.newMessages).to.be.empty;
```
crookedneighbor: It's the removedMember object that needs to sync. Not the party. In fact, I'd sync after the message is posted and assert that the member has the newMessages property and then sync again after leaving and check that it is empty.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -122,11 +122,15 @@ describe('POST /groups/:groupId/leave', () => {
...
         invitedUser = invitees[0];
+
+        await leader.post(`/groups/${group._id}/chat`, { message: 'Some message' });
       });
 
-      it('removes a group when the last member leaves', async () => {
+      it('removes a group and any new messages when the last member leaves', async () => {
         await leader.post(`/groups/${privateGuild._id}/leave`);
 
+        expect(leader.newMessages).to.be.empty;
```
crookedneighbor: You don't get notified for your own messages, so this isn't testing anything. Also, the leader object needs to sync.

gambtho: thanks -- will fix that now :)

crookedneighbor: Still waiting on this.

website/server/controllers/api-v3/groups.js
crookedneighbor: Still waiting on this.

crookedneighbor (changes requested):
I think other than my two comments still waiting on action, this looks good.

gambtho:
ok -- i think i got everything, sorry for the delay :)

test/api/v3/integration/groups/POST-groups_id_removeMember.test.js
```diff
@@ -129,6 +131,20 @@ describe('POST /groups/:groupId/removeMember/:memberId', () => {
       expect(invitedUserWithoutInvite.invitations.party).to.be.empty;
     });
 
+    it('removes new messages from a member who is removed', async () => {
+      await partyLeader.post(`/groups/${party._id}/chat`, { message: 'Some message' });
+      await partyLeader.sync();
+      await removedMember.sync();
+      await party.sync();
+      expect(removedMember.newMessages).to.not.be.empty;
```
crookedneighbor: Just to be clear about what's happening, can we assert on `removedMember.newMessages[party._id]` instead?

test/api/v3/integration/groups/POST-groups_groupId_leave.js
```diff
@@ -65,6 +65,21 @@ describe('POST /groups/:groupId/leave', () => {
         expect(groupToLeave.leader).to.equal(member._id);
       });
 
+      it('removes new messages when the last member leaves', async () => {
```
crookedneighbor: Is the `when the last member leaves` bit relevant for this assertion?

crookedneighbor (changes requested):
Left some nit-picky comments, but it looks great.

test/api/v3/integration/groups/POST-groups_groupId_leave.js
gambtho: fair enough -- removed it

test/api/v3/integration/groups/POST-groups_id_removeMember.test.js
gambtho: got this one also

website/server/controllers/api-v3/groups.js
```diff
@@ -402,6 +408,14 @@ api.leaveGroup = {
     }
 
     await group.leave(user, req.query.keep);
+
+    _removeMessagesFromMember(user, group._id);
+
+    await Bluebird.all([
+      user.save(),
+      group.save(),
```
crookedneighbor, line 416: What has changed in the group that requires the group to be saved? Doesn't `group.leave` handle the save logic?

crookedneighbor (commented):
Just left one last question. Other than that, I think this is good to go.

website/server/controllers/api-v3/groups.js
gambtho: apparently not -- in the absence of that save, the tests don't pass

crookedneighbor:
@gambtho Couldn't find a user with that id you posted. 

You can find it in [Settings > API](https://habitrpg.com/#/options/settings/api). Be sure to only copy the User ID and _not_ your API Token.

gambtho:
@crookedneighbor 349ea7be-f378-4593-8702-672f75fa83f9

crookedneighbor:

PR #7966, opened 2016-09-02
Email feedback when given on account deletion

### LINKED ISSUE #7942: optional feedback in account deletion confirmation screen
When a player goes to Settings > Site to delete their Habitica account, they are given a confirmation modal. The modal should contain a textarea for **optional** feedback about Habitica (e.g., what they didn't like, why they decided to stop using it).

The API route ("Delete an authenticated user's account") should also accept an optional feedback message.

The feedback should be emailed to admin@habitica.com. It does not need to be stored in the database.

The words describing the textarea should ask if the player would like to give any feedback, should emphasise that it's optional, and should say that it can be in whichever language the player feels comfortable with (does not have to be English).

JSchneidler:
Fixes #7942 
### Changes

Sends an automated email to `admin@habitica.com` when feedback is provided during account deletion.
Added corresponding text area for this to modal that handles account deletion.

Mail options were left very rudimentary as I wasn't sure what should go there.

---


JSchneidler:
Will fix the testing issue tomorrow, ignore this for now.

paglias:
# Before anyone merges this, ping me because we'll want to use Mandrill to send the email

website/views/shared/modals/settings.jade
```diff
@@ -66,6 +66,12 @@ script(type='text/ng-template', id='modals/delete.html')
     .row
       .col-md-6
         input.form-control(type='password', ng-model='_deleteAccount')
+    br
+    .row
+      .col-xs-12
+        .form-group
+          label(for='feedback') Feedback(optional)
```
crookedneighbor: This should be a translation string.

Alys:
@JSchneidler From the issue: "The words describing the textarea should ask if the player would like to give any feedback, should emphasise that it's optional, and should say that it can be in whichever language the player feels comfortable with (does not have to be English)."

Please add a sentence or two along those lines and then show us a screenshot of what it looks like.

website/server/controllers/api-v3/user.js
```diff
@@ -223,6 +224,17 @@ api.deleteUser = {
 
     await user.remove();
 
+    if (req.body.feedback) {
+      let mailOptions = {
+        from: '"Habitica" <noreply@habitica.com>',
+        to: 'admin@habitica.com',
+        subject: 'Feedback received from account deletion',
+        html: `<b>Feedback:</b> ${req.body.feedback}`,
```
crookedneighbor, line 232: Should we include any information about the user, such as an email address? If not, should we state that the feedback is anonymous?

Alys: I like the idea of anonymous feedback. If they want it not anonymous, they can write their contact details in their message.

crookedneighbor: Cool. Then we should state that it is anonymous.

common/locales/en/settings.json
```diff
@@ -33,6 +33,7 @@
     "resetAccPop": "Start over, removing all levels, gold, gear, history, and tasks.",
     "deleteAccount": "Delete Account",
     "deleteAccPop": "Cancel and remove your Habitica account.",
+    "feedback": "Feedback(optional)",
```
crookedneighbor, line 36: Missing a space before `(optional)`

JSchneidler:
Screwed up branching on my copy of the repo. Closing this and will open new PR when fixed and new changes incorporated.

Issue #7929, opened 2016-08-25
Change Drop Calculation To Not Rely on Seeded Stats from User


dancerinsox:
General Info
- UUID: c1d236c4-cc77-443a-9830-265af88bd7fb
- Browser: Chome
- OS: Windows 7
### Description

I have a grey wishing well task, and I was using it. I got a lion cub egg about 16 times in a row (I overused since I was curious about the behavior :P). I then paused, and tried a different task. I then got recurring shade potions...without a pause, I went to the first task I was using. Still shade potions...didn't seem to be any sync errors, and I wasn't fixing my gold value in between clicks to keep it the same (which Alys tells me would give me the same drop).

crookedneighbor:
This is actually expected behavior right now with the way the random values are generated. That said, we no longer need to seed the random value because all drop logic is now done on the server. I'm going to change the title of this issue to reflect the work that actually needs to be done. Namely, to select the drop without relying on a seed.

Original title: _Same drop ~20 times in row on a wishing well daily_

gambtho:
@crookedneighbor not entirely sure my PR is what you intended --- please validate 🔢

khipkin:
Suggestion: [help welcome now] --> [in progress]

PR #7967, opened 2016-09-03
changes drop calculation to not use seeded stats #7929

gambtho:
Fixes https://github.com/HabitRPG/habitrpg/issues/7929
### Changes

Uses math.random instead of predictable random from drop calculations

---


common/script/fns/randomDrop.js
```diff
@@ -21,7 +20,7 @@ module.exports = function randomDrop (user, options, req = {}) {
   let rarity;
   let task;
 
-  let predictableRandom = options.predictableRandom || _predictableRandom;
+  let random = options.predictableRandom || Math.random;
```
paglias: is this option used? if yes, what's being passed to it?

gambtho: i believe it's only used for testing.

paglias:
randomVal is also used in `revive`, we should check that it's only used on the server and never on the client. Also we should be able to remove `predictableRandom` at all from the code

paglias:
or better, never used anywhere where we want to run the same op and expect the same result between server and client

common/script/fns/randomDrop.js
paglias: 👍

gambtho:
removing predictable random completely will require changes to a significant amount of tests and functions, not sure it's worth the trouble

paglias:
well, if it's not used I'd like to remove it, we don't want to keep unused code that months from now we won't remember why it's there

gambtho:
Totally fair point -- just a bigger change.    Adjusting removing predictableRandom from randomVal   has an impact to a number of the tests that rely on it.

paglias:
Yeah I understand... Maybe we could tweak the predictable random function
directly to allow the use of Math.random as an option?

Il 03 set 2016 20:04, "Tom Gamble" notifications@github.com ha scritto:

gambtho:
Ooh, that's a cool idea.   Easy way to maintain the existing structure, but add randomness.   I'll revert my original change and try that.

gambtho:
@paglias thanks for the feedback, updated PR makes drops random, but minimizes other side effects

paglias:
I think you can keep the old tests, they're still valid for the predictable behavior.

@crookedneighbor opinions?

crookedneighbor:
I don't like this. Defaulting to `Math.random` in a function named `predicatableRandom` doesn't make sense to me. 

I think instead we should do some dependency injection into the randomVal function. One of the options should be a function to produce a random value, and if none is passed in, it should default to `Math.random`.

gambtho:
Totally fair point -- i'll update and resubmit

gambtho:
tried a few different ways to do this, please let me know what you think

crookedneighbor:
Did you intend to close this?

gambtho:
got concerned with the number of commits above -- wanted to validate I hadn't pulled the wrong thing

PR #7969, opened 2016-09-04
Add Google Signin

phillipthelen:
Adds Google Login and adds the option to add a social authentication method to existing accounts.


common/locales/en/front.json
```diff
@@ -248,9 +248,9 @@
...
     "invalidLoginCredentialsLong": "Uh-oh - your username or password is incorrect.\n- Make sure your username or email is typed correctly.\n- You may have signed up with Facebook, not email. Double-check by trying Facebook login.\n- If you forgot your password, click \"Forgot Password\".",
     "invalidCredentials": "There is no account that uses those credentials.",
-    "accountSuspended": "Account has been suspended, please contact admin@habitica.com with your User ID \"<%= userId %>\" for assistance.",
-    "onlyFbSupported": "Only Facebook is supported currently.",
-    "cantDetachFb": "Account lacks another authentication method, can't detach Facebook.",
+    "accountSuspended": "Account has been suspended, please contact leslie@habitica.com with your User ID \"<%= userId %>\" for assistance.",
```
crookedneighbor: This should still be `admin@`

package.json
```diff
@@ -75,6 +75,8 @@
     "pageres": "^4.1.1",
     "passport": "~0.2.1",
     "passport-facebook": "2.0.0",
+    "passport-google-oauth20": "1.0.0",
+    "passport-twitter": "1.0.4",
```
crookedneighbor: If we don't have twitter enabled yet, can we just remove the twitter stuff?

website/views/static/login-modal.jade
```diff
@@ -26,14 +26,14 @@ script(id='modals/login.html', type='text/ng-template')
     button.close(type='button', ng-click='$close()') ×
     h4.modal-title=env.t('loginAndReg')
   .modal-body(ng-controller='AuthCtrl')
-    a.zocial.facebook(alt=env.t('loginFacebookAlt'), ng-click='socialLogin("facebook")')=env.t('loginFacebookAlt')
-    //-ul.list-inline
+    //a.zocial.facebook(alt=env.t('loginFacebookAlt'), ng-click='socialLogin("facebook")')=env.t('loginFacebookAlt')
```
crookedneighbor: Any reason to keep this?

website/views/static/login-modal.jade
```diff
@@ -26,14 +26,14 @@ script(id='modals/login.html', type='text/ng-template')
...
+    ul.list-inline
       li
         a.zocial.icon.facebook(alt=env.t('loginFacebookAlt'), ng-click='socialLogin("facebook")')
       li
         a.zocial.icon.googleplus(alt="Google", ng-click='socialLogin("google")') Google+
```
crookedneighbor: Does this need a label? The Facebook one does not.

Also, it's technically not a google plus login.

website/server/libs/setupPassport.js
```diff
@@ -1,8 +1,10 @@
 import passport from 'passport';
 import nconf from 'nconf';
 import passportFacebook from 'passport-facebook';
+import passportGoogle from 'passport-google-oauth20';
 
 const FacebookStrategy = passportFacebook.Strategy;
+const GoogleStrategy = passportGoogle.Strategy;
```
crookedneighbor: Why not:

``` js
import { Strategy as GoogleStrategy } from 'passport-google-oauth20';
```

Same with the Facebook one, unless `passportFacebook` is used elsewhere.

website/server/controllers/api-v3/auth.js
```diff
@@ -227,6 +227,8 @@ api.loginLocal = {
   },
 };
 
+const supportedSocialNetworks = ['facebook', 'google'];
```
crookedneighbor: Let's put this at the top of the file under the imports. Our convention is to use screaming snake case for these. Also, let's freeze it.

``` js
const SUPPORTED_SOCIAL_NETWORKS = Object.freeze(['facebook', 'google']);
```

website/client/js/controllers/settingsCtrl.js
```diff
@@ -287,6 +293,37 @@ habitrpg.controller('SettingsCtrl',
       return Math.floor(numberOfHourglasses);
     };
 
+    $scope.hasSocialAuth = function(user) {
+      for (var id in SOCIAL_AUTH_NETWORKS) {
```
crookedneighbor: You should use forEach here:

``` js
SOCIAL_AUTH_NETWORKS.forEach(function (network) {
  if (user.auth.hasOwnProperty(network.key)) {
    return Boolean(user.auth[network.key].id)
  }
});
```

website/client/js/controllers/settingsCtrl.js
```diff
@@ -287,6 +293,37 @@ habitrpg.controller('SettingsCtrl',
...
+        }
+      }
+      return false;
+    };
+
+    $scope.hasBackupAuthOption = function(user, checkedNetworkKey) {
+      if (user.auth.local.username) {
+        return true;
+      }
+      for (var id in SOCIAL_AUTH_NETWORKS) {
```
crookedneighbor: Same as above. A `forEach` should be used. `for in`s can have weird behavior if a browser extension has added something to the object or array prototypes.

website/client/js/controllers/settingsCtrl.js
```diff
@@ -2,14 +2,20 @@
...
-  function($scope, User, $rootScope, $http, ApiUrl, Guide, $location, $timeout, Content, Notification, Shared, $compile) {
+  ['$scope', 'User', '$rootScope', '$http', 'ApiUrl', 'Guide', '$location', '$timeout', 'Content', 'Notification', 'Shared', 'Social', '$compile',
+  function($scope, User, $rootScope, $http, ApiUrl, Guide, $location, $timeout, Content, Notification, Shared, Social, $compile) {
     var RELEASE_ANIMAL_TYPES = {
       pets: 'releasePets',
       mounts: 'releaseMounts',
       both: 'releaseBoth',
     };
 
+    var SOCIAL_AUTH_NETWORKS = [
```
crookedneighbor: Should this be added in the common files so both the server and the client use the same enumerated list of available login methods?

test/api/v3/integration/user/auth/POST-user_auth_social.test.js
```diff
@@ -22,38 +23,101 @@ describe('POST /user/auth/social', () => {
...
-    let response = await api.post(endpoint, {
-      authResponse: {access_token: randomAccessToken}, // eslint-disable-line camelcase
-      network,
+      expect(response.apiToken).to.exist;
+      expect(response.id).to.exist;
+      expect(response.newUser).to.be.true;
+    });
+
+    it('logs an existing user in', async () => {
+      await user.update({ 'auth.facebook.id': facebookId });
```
crookedneighbor: Facebook or google here?

website/client/js/services/socialServices.js
```diff
@@ -34,8 +36,27 @@
...
     }
 
+    hello.init({
+      facebook : window.env.FACEBOOK_KEY,
+      google : window.env.GOOGLE_CLIENT_ID
+    });
+
+    function socialLogin(network){
+      hello(network).login({scope:['email', 'profile']}).then(function(auth){
+        $http.post(ApiUrl.get() + "/api/v3/user/auth/social", auth)
```
crookedneighbor: How does this work? Is `ApiUrl` on the window?

website/client/js/services/socialServices.js
```diff
@@ -34,8 +36,27 @@
...
+          .success(function(res, status, headers, config) {
+            runAuth(res.data.id, res.data.apiToken);
+          }).error(errorAlert);
```
crookedneighbor: Where is `errorAlert` defined?

website/client/js/services/socialServices.js
```diff
@@ -34,8 +36,27 @@
...
+      }, function( e ){
```
crookedneighbor: `(err)` here

website/client/js/services/socialServices.js
```diff
@@ -34,8 +36,27 @@
...
+        alert("Signin error: " + e.message );
```
crookedneighbor: Is this the best way to handle this?

website/server/controllers/api-v3/auth.js
```diff
@@ -239,17 +241,39 @@ function _passportFbProfile (accessToken) {
   });
 }
 
+function _passportGoogleProfile (accessToken) {
+  return new Bluebird((resolve, reject) => {
```
crookedneighbor: Since it's just a normal promise here (nothing Bluebird specific), just use `Promise`

crookedneighbor: There's also a bluebird method to convert callbacks like this to promises automatically. That's probably what should actually be done for both of these.

website/server/controllers/api-v3/auth.js
```diff
@@ -239,17 +241,39 @@ function _passportFbProfile (accessToken) {
...
+        reject(err);
+      } else {
+        resolve(profile);
+      }
+    });
+  });
+}
+
+function _passportProfile (network, accessToken) {
+  if (network === 'facebook') {
```
crookedneighbor: You could also attach these functions to an object like:

``` js
let passportProfiles = {
  facebook: _passportFbProfile,
  google: _passportGoogleProfile,
}

function _passportProfile (network, accessToken) {
  if (network in passportProfiles) {
    return passportProfiles[network](accessToken);
  }
}
```

Then adding additional logins is as trivial as updating the object mapping.

website/server/controllers/api-v3/auth.js
```diff
@@ -239,17 +241,39 @@ function _passportFbProfile (accessToken) {
...
 api.loginSocial = {
   method: 'POST',
   url: '/user/auth/social', // this isn't the most appropriate url but must be the same as v2
   async handler (req, res) {
+    let existingUser = res.locals.user;
     let accessToken = req.body.authResponse.access_token;
     let network = req.body.network;
 
-    if (network !== 'facebook') throw new NotAuthorized(res.t('onlyFbSupported'));
+    if (supportedSocialNetworks.indexOf(network) === -1) throw new NotAuthorized(res.t('unsupportedNetwork'));
```
crookedneighbor: Lets go with `BadRequest` for these instead.

website/server/controllers/api-v3/auth.js
```diff
@@ -574,11 +609,11 @@ api.deleteSocial = {
...
     let user = res.locals.user;
     let network = req.params.network;
-
-    if (network !== 'facebook') throw new NotAuthorized(res.t('onlyFbSupported'));
-    if (!user.auth.local.username) throw new NotAuthorized(res.t('cantDetachFb'));
-
-    await User.update({_id: user._id}, {$unset: {'auth.facebook': 1}}).exec();
+    if (supportedSocialNetworks.indexOf(network) === -1) throw new NotAuthorized(res.t('unsupportedNetwork'));
+    if (!user.auth.local.username) throw new NotAuthorized(res.t('cantDetachSocial'));
+    let unset = {};
```
crookedneighbor:
``` js
let unset = {
  [`auth.${network}`]: 1,
};
```

website/server/controllers/api-v3/auth.js
```diff
@@ -574,11 +609,11 @@ api.deleteSocial = {
   async handler (req, res) {
     let user = res.locals.user;
     let network = req.params.network;
-
-    if (network !== 'facebook') throw new NotAuthorized(res.t('onlyFbSupported'));
-    if (!user.auth.local.username) throw new NotAuthorized(res.t('cantDetachFb'));
-
-    await User.update({_id: user._id}, {$unset: {'auth.facebook': 1}}).exec();
+    if (supportedSocialNetworks.indexOf(network) === -1) throw new NotAuthorized(res.t('unsupportedNetwork'));
+    if (!user.auth.local.username) throw new NotAuthorized(res.t('cantDetachSocial'));
```
crookedneighbor: So if you've attached via Google _and_ Facebook, you can't detach one of them if you don't also have a local?

crookedneighbor:
I left a lot of comments, but I think this is most of the way there. Great job!

crookedneighbor:
Looks like you forgot to update the delete social login tests.

website/server/controllers/api-v3/auth.js
khipkin: Oh my god screaming snake case

website/server/controllers/api-v3/auth.js
phillipthelen: using `Promise` instead of `Bluebird` results in an error.

```
TypeError: A value undefined was yielded that could not be treated as a promise
```

What would the right way to do this be?

website/client/js/services/socialServices.js
phillipthelen: Oh. That's only defined in authCtrl.js. so it's not actually accessible here. What would the best way be to make that accessible in both places?

website/client/js/services/socialServices.js
paglias: it's injected at line 9

website/client/js/services/socialServices.js
paglias: a service?

phillipthelen:
I moved the errorAlert method to an `alertServices` service. When I manually test it, it seems to work. But when I run the tests, the test for incorrect login data fails because the spy for `$window.alert` is not called. Although there is an alert when I manually try this.is the spy maybe working with the wrong thing, because of the new service?

website/server/controllers/api-v3/auth.js
```diff
@@ -231,7 +231,7 @@ api.loginLocal = {
 
 function _passportProfile (network, accessToken) {
   if (network in SUPPORTED_SOCIAL_NETWORKS) {
-    return new Promise.promisify((resolve, reject) => {
+    return new Bluebird((resolve, reject) => {
```
paglias: just `new Promise` should work

phillipthelen: That always results in `TypeError: A value undefined was yielded that could not be treated as a promise`

paglias: I think that happens because later you're `await`ing passportProfile but it when `network in SUPPORTED_SOCIAL_NETWORKS` returns false it doesn't return a promise that is necessary when using await .

Also, I don't know if we misunderstood but I meant that you should use `new Promise(...)` not `new Promise.promisify`

phillipthelen:
Got the tests fixed!

common/locales/en/front.json
```diff
@@ -249,8 +249,8 @@
     "invalidLoginCredentialsLong": "Uh-oh - your username or password is incorrect.\n- Make sure your username or email is typed correctly.\n- You may have signed up with Facebook, not email. Double-check by trying Facebook login.\n- If you forgot your password, click \"Forgot Password\".",
     "invalidCredentials": "There is no account that uses those credentials.",
     "accountSuspended": "Account has been suspended, please contact admin@habitica.com with your User ID \"<%= userId %>\" for assistance.",
-    "onlyFbSupported": "Only Facebook is supported currently.",
-    "cantDetachFb": "Account lacks another authentication method, can't detach Facebook.",
+    "unsupportedNetwork": "This network is currently not supported.",
+    "cantDetachSocial": "Account lacks another authentication method, can't detach this authentication method.",
```
khipkin: Grammar nit: comma splice.

common/locales/en/settings.json
```diff
@@ -90,8 +90,8 @@
     "passwordSuccess": "Password successfully changed",
     "usernameSuccess": "Login Name successfully changed",
     "emailSuccess": "Email successfully changed",
-    "detachFacebook": "De-register Facebook",
-    "detachedFacebook": "Successfully removed Facebook from your account",
+    "detachSocial": "De-register <%= network %>",
+    "detachedSocial": "Successfully removed authentication method from your account",
```
khipkin: Why parameterize one but not the other?

test/api/v3/integration/user/auth/POST-user_auth_social.test.js
khipkin: This is in the Facebook `describe` block.

test/api/v3/integration/user/auth/POST-user_auth_social.test.js
```diff
@@ -11,49 +11,109 @@ describe('POST /user/auth/social', () => {
...
 
-  it('logs an existing user in', async () => {
-    await user.update({ 'auth.facebook.id': facebookId });
+  describe('google', () => {
+    before(async () => {
+      let expectedResult = {id: googleId};
+      let passportGoogleProfile = sinon.stub(passport._strategies.google, 'userProfile');
+      passportGoogleProfile.yields(null, expectedResult);
+      network = 'google';
+    });
```
khipkin: nit: add nl after this

website/client/js/controllers/settingsCtrl.js
khipkin: :+1: to this

website/client/js/controllers/settingsCtrl.js
```diff
@@ -287,6 +293,33 @@ habitrpg.controller('SettingsCtrl',
       return Math.floor(numberOfHourglasses);
     };
 
+    $scope.hasSocialAuth = function(user) {
+      SOCIAL_AUTH_NETWORKS.forEach(function (network) {
+        if (user.auth.hasOwnProperty(network.key)) {
+          return Boolean(user.auth[network.key].id)
```
khipkin: If this can ever return `false`, it would be possible to return `false` before a potential `true` is reached.

website/client/js/services/alertServices.js
```diff
@@ -0,0 +1,32 @@
...
+  ];
+
+  function alertFactory($window) {
+
+    function authErrorAlert(data, status, headers, config) {
+      if (status === 0) {
+        $window.alert(window.env.t('noReachServer'));
+      } else if (status === 400 && data.errors && _.isArray(data.errors)) { // bad requests
+        data.errors.forEach(function (err) {
+          $window.alert(err.message);
```
khipkin: Will this spawn multiple alert windows?  If so, maybe it would be better to combine all of the error messages in a single alert window.

website/server/controllers/api-v3/auth.js
```diff
@@ -121,10 +122,10 @@ api.registerLocal = {
       },
     };
 
-    if (fbUser) {
-      if (!fbUser.auth.facebook.id) throw new NotAuthorized(res.t('onlySocialAttachLocal'));
-      fbUser.auth.local = newUser.auth.local;
-      newUser = fbUser;
+    if (existingUser) {
+      if (!existingUser.auth.facebook.id && !existingUser.auth.google.id) throw new NotAuthorized(res.t('onlySocialAttachLocal'));
```
khipkin: Suggestion: loop through SUPPORTED_SOCIAL_NETWORKS instead of manually referencing facebook and google.

website/client/js/controllers/settingsCtrl.js
crookedneighbor: You should use `find` not `forEach`

``` js
$scope.hasSocialAuth = function (user) {
  return SOCIAL_AUTH_NETWORKS.find(function (network) {
    if (user.auth.hasOwnProperty(network.key)) {
      return user.auth[network.key].id;
    }
  })
};
```

This will return true if it can find the social auth and return undefined if it cannot.

The way you have it, this method will _always_ return false because `forEach`s don't break like `for loop`s do.

crookedneighbor: s/true/truthy

website/client/js/services/alertServices.js
paglias: It could but since it's only used for authentication is should never, also it's been this way for a long time. We'll change it with the redesign

website/views/options/settings.jade
```diff
@@ -131,9 +131,12 @@ script(type='text/ng-template', id='partials/options.settings.settings.html')
...
           .panel-body
-            div(ng-if='user.auth.facebook.id')
-              button.btn.btn-primary(disabled='disabled', ng-if='!user.auth.local.username')=env.t('registeredWithFb')
-              button.btn.btn-danger(ng-click='http("delete", "/api/v3/user/auth/social/facebook", null, "detachedFacebook")', ng-if='user.auth.local.username')=env.t('detachFacebook')
+            div
+              ul.list-inline
+                li(ng-repeat='network in SOCIAL_AUTH_NETWORKS')
+                  button.btn.btn-primary(ng-if='!user.auth[network.key].id', ng-click='socialLogin(network.key)')=env.t('registerWithSocial', {network: '{{network.name}}'})
+                  button.btn.btn-primary(disabled='disabled', ng-if='!hasBackupAuthOption(user, network.key) && user.auth[network.key].id')=env.t('registeredWithSocial', {network: '{{network.name}}'})
+                  button.btn.btn-danger(ng-click='http("delete", "/api/v3/user/auth/social/"+network.key, null, "detachedSocial")', ng-if='hasBackupAuthOption(user, network.key) && user.auth[network.key].id')=env.t('detachSocial', {network: '{{network.name}}'})
```
khipkin: Suggestion: move this ApiUrl logic out of the jade file.

website/views/static/login-modal.jade
```diff
@@ -26,14 +26,11 @@ script(id='modals/login.html', type='text/ng-template')
...
-    a.zocial.facebook(alt=env.t('loginFacebookAlt'), ng-click='socialLogin("facebook")')=env.t('loginFacebookAlt')
-    //-ul.list-inline
+    ul.list-inline
       li
         a.zocial.icon.facebook(alt=env.t('loginFacebookAlt'), ng-click='socialLogin("facebook")')
       li
-        a.zocial.icon.googleplus(alt="Google", ng-click='socialLogin("google")') Google+
-      li
-        a.zocial.icon.twitter(alt="Twitter", ng-click='socialLogin("twitter")') Twitter
+        a.zocial.icon.google(alt="Google", ng-click='socialLogin("google")')
```
khipkin: Should "Google" be localizable?

paglias: I don't think so

website/client/js/services/alertServices.js
crookedneighbor: Agreed. In the redesign, we won't use alerts or confirm dialogs at all.

common/locales/en/front.json
crookedneighbor: `, -> ;`

test/api/v3/integration/user/auth/POST-user_auth_social.test.js
crookedneighbor: I think the line commit got shifted around since I made this comment. If I recall, it _was_ in the Google section. (or I am just completely mistaken)

website/client/js/services/socialServices.js
crookedneighbor: It wasn't when I made the comment :)

website/server/controllers/api-v3/auth.js
crookedneighbor: :+1:

website/views/static/login-modal.jade
crookedneighbor: Definitely not.

website/server/models/user/schema.js
```diff
@@ -22,6 +22,12 @@ let schema = new Schema({
     facebook: {type: Schema.Types.Mixed, default: () => {
       return {};
     }},
+    google: {type: Schema.Types.Mixed, default: () => {
+      return {};
+    }},
+    twitter: {type: Schema.Types.Mixed, default: () => {
```
crookedneighbor: This should be removed until twitter is actually a method to add.

website/views/static/login-modal.jade
khipkin: :+1:

common/locales/en/settings.json
paglias: Agreed, let's pass the network in here

website/client/js/controllers/settingsCtrl.js
paglias: 👍

website/server/models/user/hooks.js
```diff
@@ -95,14 +95,30 @@ function _populateDefaultsForNewUser (user) {
...
+  let google = user.auth.google;
 
   let localUsername = user.auth.local && user.auth.local.username;
-  let facebookUsername = fb && (fb.displayName || fb.name || fb.username || `${fb.first_name && fb.first_name} ${fb.last_name}`);
+  let facebookUsername = _getFacebookName(user.auth.facebook);
+  let googleUsername = google && google.displayName;
   let anonymous = 'Anonymous';
 
-  return localUsername || facebookUsername || anonymous;
+  return localUsername || facebookUsername || googleUsername || anonymous;
```
paglias: maybe this should be `localUsername || _getFacebookName() || googleUsername || anonymous` so we don't execute `_getFacebookName()` if not necessary

paglias:
Few small comments, otherwise looks good to me 👍

website/server/models/user/hooks.js
crookedneighbor: I agree.

common/locales/en/settings.json
phillipthelen: This is passed as an argument to the `http` function. ( `$http("delete", ApiUrl.get() + "/api/v3/user/auth/social/"+networkKey, null, "detachedSocial");` ) Is it even possible to pass in the network in this case?

website/client/js/controllers/settingsCtrl.js
```diff
@@ -287,6 +293,33 @@ habitrpg.controller('SettingsCtrl',
...
+    $scope.hasSocialAuth = function (user) {
+      return SOCIAL_AUTH_NETWORKS.find(function (network) {
+        if (user.auth.hasOwnProperty(network.key)) {
+          return user.auth[network.key].id;
+        }
+      });
+    };
+
+    $scope.deleteSocialAuth = function (networkKey) {
+      $http("delete", ApiUrl.get() + "/api/v3/user/auth/social/"+networkKey, null, "detachedSocial");
```
khipkin: @vIiRuS So does the "detachedSocial" string get used for localization within the `$http` module?  Is there any way to pass in a pre-localized message?

common/locales/en/front.json
```diff
@@ -250,8 +250,8 @@
     "invalidLoginCredentialsLong": "Uh-oh - your username or password is incorrect.\n- Make sure your username or email is typed correctly.\n- You may have signed up with Facebook, not email. Double-check by trying Facebook login.\n- If you forgot your password, click \"Forgot Password\".",
     "invalidCredentials": "There is no account that uses those credentials.",
     "accountSuspended": "Account has been suspended, please contact admin@habitica.com with your User ID \"<%= userId %>\" for assistance.",
-    "onlyFbSupported": "Only Facebook is supported currently.",
-    "cantDetachFb": "Account lacks another authentication method, can't detach Facebook.",
+    "unsupportedNetwork": "This network is currently not supported.",
```
crookedneighbor: The wording on this is kinda clunky (was clunky before). How about: `This network is not currently supported.`

website/server/controllers/api-v3/auth.js
```diff
@@ -121,10 +121,15 @@ api.registerLocal = {
       },
     };
 
-    if (fbUser) {
-      if (!fbUser.auth.facebook.id) throw new NotAuthorized(res.t('onlySocialAttachLocal'));
-      fbUser.auth.local = newUser.auth.local;
-      newUser = fbUser;
+    if (existingUser) {
+      let hasSocialAuth = common.constants.SUPPORTED_SOCIAL_NETWORKS.find(network => {
```
crookedneighbor, line 137: This exact logic is used both on the client and the server. Good case for extracting it into a common lib?

website/server/controllers/api-v3/auth.js
```diff
@@ -259,36 +269,49 @@ api.loginSocial = {
...
           [network]: profile,
         },
         preferences: {
           language: req.language,
         },
-      });
-      user.registeredThrough = req.headers['x-client'];
+      };
+
+
```
crookedneighbor: lots of extra lines here.

website/client-old/js/services/alertServices.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
```
crookedneighbor, line 1: I'm not thrilled about perpetuating our use of alert windows, but since this will (hopefully) be short-lived code, I'm not going to require that you do anything different.

website/client-old/js/controllers/settingsCtrl.js
```diff
@@ -2,14 +2,17 @@
...
-  function($scope, User, $rootScope, $http, ApiUrl, Guide, $location, $timeout, Content, Notification, Shared, $compile) {
+  ['$scope', 'User', '$rootScope', '$http', 'ApiUrl', 'Guide', '$location', '$timeout', 'Content', 'Notification', 'Shared', 'Social', '$compile',
+  function($scope, User, $rootScope, $http, ApiUrl, Guide, $location, $timeout, Content, Notification, Shared, Social, $compile) {
     var RELEASE_ANIMAL_TYPES = {
       pets: 'releasePets',
       mounts: 'releaseMounts',
       both: 'releaseBoth',
     };
 
+    var SOCIAL_AUTH_NETWORKS = window.habitrpgShared.constants.SUPPORTED_SOCIAL_NETWORKS
```
crookedneighbor: This should be accessible as `Shared.constants.SUPPORTED_SOCIAL_NETWORKS`

Also, missing a `;`

crookedneighbor (changes requested):
Just a few minor things to clean up.

website/server/controllers/api-v3/auth.js
crookedneighbor: The linter will fail the build for this.

config.json.example
```diff
@@ -7,6 +7,8 @@
     "FACEBOOK_ANALYTICS":"1234567890123456",
     "FACEBOOK_KEY":"123456789012345",
     "FACEBOOK_SECRET":"aaaabbbbccccddddeeeeffff00001111",
+    "GOOGLE_CLIENT_ID":"123456789012345",
```
crookedneighbor, line 10: Not something we have to do right away (so it doesn't block this PR), but we should add a description to the wiki for how to create creds for a local install. Same with Facebook.

website/server/libs/email.js
```diff
@@ -47,8 +48,12 @@ export function getUserInfo (user, fields = []) {
   if (fields.indexOf('email') !== -1) {
     if (user.auth.local && user.auth.local.email) {
       info.email = user.auth.local.email;
-    } else if (user.auth.facebook && user.auth.facebook.emails && user.auth.facebook.emails[0] && user.auth.facebook.emails[0].value) {
-      info.email = user.auth.facebook.emails[0].value;
+    } else {
+      common.constants.SUPPORTED_SOCIAL_NETWORKS.forEach(network => {
+        if (user.auth[network.id] && user.auth[network.id].emails && user.auth[network.id].emails[0] && user.auth[network.id].emails[0].value) {
+          info.email = user.auth.facebook.emails[0].value;
```
crookedneighbor: Is this right? shouldn't it be `user.auth[network.id].emails[0].value;`

Also, would it make sense to clean up this logic by using `_.get` here?

crookedneighbor: Also, the test for this is now failing.

website/client-old/js/controllers/settingsCtrl.js
```diff
@@ -287,6 +290,39 @@ habitrpg.controller('SettingsCtrl',
       return Math.floor(numberOfHourglasses);
     };
 
+    $scope.hasBackupAuthOption = function(user, checkedNetworkKey) {
+      if (user.auth.local.username) {
+        return true;
+      }
+      return SOCIAL_AUTH_NETWORKS.find(function (network) {
```
crookedneighbor: I didn't realize until yesterday that `find` is not supported in IE. I'll fix this. https://github.com/vIiRuS/habitrpg/pull/6

website/client-old/js/services/socialServices.js
```diff
@@ -34,8 +36,27 @@
       }
     }
 
+    hello.init({
+      facebook : window.env.FACEBOOK_KEY,
+      google : window.env.GOOGLE_CLIENT_ID
+    });
+
+    function socialLogin(network){
+      hello(network).login({scope:['email', 'profile']}).then(function(auth){
```
crookedneighbor: I'm getting an out of scope error when I register with Facebook. Should this be: `hello(network).login({scope:'email'})` instead?

crookedneighbor (changes requested):
Just not sure about the Facebook error.

PR #7972, opened 2016-09-06
Real Time Chat v1 (parties only)

paglias:
WIP
- [x] Sync party on system message
- [x] Change button from "Fetch recent messages" to "Sync party"?
- [x] Notifications (menu)
- [x] Notification (interactive)
- [x] Sync party on reconnection after inactivity
- [x] Remove limit of 1 tab connected per browser?
- [x] lastSeen (throttle) - later maybe
- [x] throttle markChatSeen (later if messages volume explodes)
- [x] members status
- [x] disconnect when leaving a party (and reconnect correctly when joining)
- [x] user._v gets increased every time a chat message in their party is sent, causing a sync the next time they interact with the site...
- [x] audio


website/client/js/services/pusherService.js
```diff
@@ -59,29 +59,49 @@ angular.module('habitrpg')
...
-        // If we just reconnected after some inactivity, sync the party
-        if (reconnecting === true) {
-          Groups.party(true).then(function (syncedParty) {
-            // Assign and not replace so that all the references get the modifications
-            if ($rootScope.party) {
-              _.assign($rootScope.party, syncedParty);
-              $rootScope.loadingParty = false; // make sure the party is set as loaded
+      partyChannel.bind('pusher:subscription_succeeded', function(pusherMembers) {
+        // Wait for the party to be loaded
+        Groups.party(reconnecting ? true : false).then(function (party) {
```
khipkin: Suggestion: `Boolean(reconnecting)`

crookedneighbor: Agreed. Any time a ternary returns `true : false`, it should use `Boolean` instead.

crookedneighbor: Though, to be honest, I don't think `Groups.party` cares that it is a `Boolean` just whether or not it's truthy. So `Groups.party(reconnecting)` should work.

paglias: I'll change it, actually i think it does check if (forceSync === true)

crookedneighbor: :+1: Though in this case, reconnecting will only ever equal `true` or `undefined`, right? 

Doesn't hurt to `Boolean(reconnecting)` though.

paglias: yeah only true or undefined

paglias: Actually I've decided not to use Boolean but just pass reconnecting since it can only be true or undefined

paglias:
Should be ready for review :)

PR #7978, opened 2016-09-07
API Warning message

### LINKED ISSUE #3049: API could use clarification/warning
- [x] add a warning/clarification to the API token, e.g. _This token should only be used for synchronization or integration between other software programs and your habit account. This API token should not be given out to others or posted on a site such as Github._\* 
- [ ] add User ID explanation, e.g _This is the User ID you will need to provide to join a party or when seeking help for account or gem issues._
- [ ] @lefnire's suggestion on #2592: ability to change the API token in case it is compromised
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1437466-api-could-use-clarification-warning?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

librarianmage:
Fixes #3049 (ish)
### Changes

Adds a warning to the API screen advising users to post to the Report a Bug guild if a new token is needed, but warning them that it won't happen instantly:

capture

---

### DIFF
--- a/common/locales/en/settings.json
+++ b/common/locales/en/settings.json
@@ -68,6 +68,7 @@
     "APIv3": "API v3",
     "APIText": "Copy these for use in third party applications. However, think of your API Token like a password, and do not share it publicly. You may occasionally be asked for your User ID, but never post your API Token where others can see it, including on Github.",
     "APIToken": "API Token (this is a password - see warning above!)",
+    "APITokenWarning": "<strong>Warning!</strong> If you need a new API token, post a request on the <a href='https://habitica.com/#/options/groups/guilds/a29da26b-37de-4a71-b0c6-48e72a900dac' class='alert-link'>Report A Bug</a> guild. This may take time, and once it is reset you will need to re-authorize everything.",
     "thirdPartyApps": "Third Party Apps",
     "dataToolDesc": "A webpage that shows you certain information from your Habitica account, such as statistics about your tasks, equipment, and skills.",
     "beeminder": "Beeminder",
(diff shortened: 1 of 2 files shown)


Alys:
I'm not sure it needs to be a warning because it's not something the users need to be careful of. We should also explain the reason why you'd want a new token (if someone else accidentally found out what yours is) otherwise we'll have users asking us why they might want one. 

Leave out the part about how it may take time. We'd try to get that done as quickly as possible for security. And that has made me realise that this is unfortunately one case where they should not post to the guild. As soon as a malicious player sees someone requesting a new token, they'd go searching the web to try to find where the token was posted. Please change it to something like "email admin@habitica.com and tell them your User ID and your current API Token" (FYI, I want both in the email as partial proof that they own the account). Sorry about that - I should have thought of that to begin with.

For "reauthorise everything", players might not realise every implication of that (e.g., we've had even coders who've had their tokens reset be surprised that the mobile app stopped working). Can you explain that they should log out and back in for the website and official mobile apps, and for any third-party integrations, they'll need to provide the new token.

librarianmage:
screenshot 2016-09-07 at 8 44 42 pm

How does this look?

Alys:
I like how it looks! Let's go with something like this:

"If you need a new API Token (e.g., if you accidentally shared it), email <a href='mailto:admin@habitica.com'>admin@habitica.com</a> with your User ID and current Token. Once it is reset you will need to re-authorize everything by logging out of the website and mobile app and by providing the new Token to any other Habitica tools that you use."

That keeps consistency with the naming and capitalisation used elsewhere on the page and avoids people thinking that the last part applies to only browser extensions. It doesn't specify logging back in (for brevity) but they should be able to work that part out. :)

librarianmage:
Done!

Alys:
Looks good! Thanks! I've notated this towards your next tier.

PR #7979, opened 2016-09-08
Fix infinite build error

### LINKED ISSUE #7861: Fix super long error output when test fails in Travis build
I'm not sure why, but lately my builds have been throwing the same error over and over again, sometimes running for more than 45 minutes.

For example, [this build](https://travis-ci.org/HabitRPG/habitrpg/builds/147853560) only stopped because the output file exceeded the limit of 4MB, with over 80k lines of output.  And this only happened after more than 48 minutes.

It's just this failure over and over again:

> 'build:dev' errored after 28 s
> Error in plugin 'run-sequence(test:karma)'
> Message:
>     Command failed: /bin/sh -c NODE_ENV=test  npm run test:karma

I think we should limit the number of retries for running karma tests to prevent this from happening again and hogging TravisCI resources.

crookedneighbor:
Fixes #7861

Here's a build that illustrates the problem: https://travis-ci.org/crookedneighbor/habitrpg/builds/158311228

Here's the same failing build, but with these changes added: https://travis-ci.org/crookedneighbor/habitrpg/builds/158312916

### DIFF
--- a/tasks/gulp-tests.js
+++ b/tasks/gulp-tests.js
@@ -90,7 +90,10 @@ gulp.task('test:sanity', (cb) => {
   let runner = exec(
     testBin(SANITY_TEST_COMMAND),
     (err, stdout, stderr) => {
-      cb(err);
+      if (err) {
+        process.exit(1);
+      }
+      cb();
     }
   );
   pipe(runner);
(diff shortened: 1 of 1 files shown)


SabreCat:
Hallelujah!

PR #7984, opened 2016-09-08
A/B Testing, Round 1

SabreCat:
### Changes

Implements a user property `ABtest` that is then sent as part of user-level properties for analytics calls. This will be used to sort new users into cohorts to test different setups for improving retention.

The first of these tests is included as part of this PR: do users retain better if the site uses sound effects from the beginning? ~50% of new users during the test will have the "Rosstavo Theme" enabled by default, while the rest continue with the current setup of no starting sounds.

---

UUID: [Sound Test Retention](https://map.what3words.com/sound.test.retention)

### DIFF
--- a/test/api/v3/integration/user/auth/POST-register_local.test.js
+++ b/test/api/v3/integration/user/auth/POST-register_local.test.js
@@ -64,8 +64,8 @@ describe('POST /user/auth/local/register', () => {
         confirmPassword: password,
       });
 
-      expect(user.ABtest).to.exist;
-      expect(user.ABtest).to.be.a('string');
+      expect(user._ABtest).to.exist;
+      expect(user._ABtest).to.be.a('string');
     });
 
     it('requires password and confirmPassword to match', async () => {
--- a/test/api/v3/integration/user/auth/POST-register_local.test.js
+++ b/test/api/v3/integration/user/auth/POST-register_local.test.js
@@ -3,6 +3,7 @@ import {
   requester,
   translate as t,
   createAndPopulateGroup,
+  getProperty,
 } from '../../../../../helpers/api-integration/v3';
 import { v4 as generateRandomUserName } from 'uuid';
 import { each } from 'lodash';
@@ -64,8 +65,8 @@ describe('POST /user/auth/local/register', () => {
         confirmPassword: password,
       });
 
-      expect(user._ABtest).to.exist;
-      expect(user._ABtest).to.be.a('string');
+      await expect(getProperty('users', user._id, '_ABtest')).to.eventually.exist;
+      await expect(getProperty('users', user._id, '_ABtest')).to.eventually.be.a('string');
     });
 
     it('requires password and confirmPassword to match', async () => {
--- a/test/api/v3/integration/user/auth/POST-register_local.test.js
+++ b/test/api/v3/integration/user/auth/POST-register_local.test.js
@@ -52,6 +52,22 @@ describe('POST /user/auth/local/register', () => {
       expect(user.tasksOrder.habits).to.have.a.lengthOf(0);
     });
 
+    it('enrolls new users in an A/B test', async () => {
+      let username = generateRandomUserName();
+      let email = `${username}@example.com`;
+      let password = 'password';
+
+      let user = await api.post('/user/auth/local/register', {
+        username,
+        email,
+        password,
+        confirmPassword: password,
+      });
+
+      expect(user.ABtest).to.exist;
+      expect(user.ABtest).to.be.a('string');
+    });
+
     it('requires password and confirmPassword to match', async () => {
       let username = generateRandomUserName();
       let email = `${username}@example.com`;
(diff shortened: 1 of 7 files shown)


website/server/models/user/schema.js
```diff
@@ -521,6 +521,7 @@ let schema = new Schema({
     return {};
   }},
   pushDevices: [PushDeviceSchema],
+  ABtest: {type: String},
```
paglias: should be added to the list of private and noSet properties

crookedneighbor: User is different than other models. Everything is noset by default. It should be made private though.

SabreCat: Like so? 3ffde38

SabreCat: Shoot, that makes it show up `undefined` during the integration test. Is the test wrong, or the way I found to make the property private?

paglias: It's the way private properties work, they don't show up in the API and so
in integration tests. I'd say either move it to the pre save hook and use a
unit test or just make the property public, it doesn't contain sensitive
data anyway. It was nice to have but not necessary

Il 08 set 2016 23:52, "Sabe Jones" notifications@github.com ha scritto:

crookedneighbor: Rule of thumb:
- If the client should not be able to see it, mark it as private
- If the client should be able to see it but third party integrations should not use it, mark it as internally private by giving it a leading `_`
- If the client should be able to see and integrations should use it, leave it as is

crookedneighbor: Are we only ever going to have a single AB test going on? Should we at least do:

``` js
ABtest: {
  newUserRetention: String
}
```

As I assume there are other kinds of AB testing we could do.

khipkin: @crookedneighbor But I think it would make sense for any given user to be in **maximally one** AB test. Otherwise, the retention data wouldn't be particularly useful because there would be no way to tell which test/reason had an effect.  And as far as I can tell, the string wouldn't be checked in the code, as all of the code differences would take place during account/user creation.  Therefore, the field might as well be left as a (descriptive) string.

SabreCat:
Rebased to resolve merge conflicts, changed from private to internally private property.

paglias:
👍 to me, the only change could be to make `_ABTest` private, deleting it in `controllers/user/getUser`

test/api/v3/integration/user/auth/POST-register_local.test.js
```diff
@@ -52,6 +52,22 @@ describe('POST /user/auth/local/register', () => {
...
+      let password = 'password';
+
+      let user = await api.post('/user/auth/local/register', {
+        username,
+        email,
+        password,
+        confirmPassword: password,
+      });
+
+      expect(user._ABtest).to.exist;
```
crookedneighbor: I would still make this attribute private and write a helper function to query the db directly to assert that it exists. There's no need for the client to have this.

website/server/controllers/api-v3/auth.js
```diff
@@ -130,6 +130,14 @@ api.registerLocal = {
...
     }
 
+    // A/B Test 2016-09-12: Start with Sound Enabled?
+    if (Math.random() < 0.5) {
+      newUser.preferences.sound = 'rosstavoTheme';
+      newUser._ABtest = '20160912-soundEnabled';
+    } else {
+      newUser._ABtest = '20160912-soundDisabled';
+    }
+
```
khipkin, line 140: Suggestions:
- Move the ABtest logic out of here.
- Have some sort of collection of possible ABtests (right now, it would only have this one), where each test has a set of possible options.
- Choose one test at random, then choose one option at random.

That way, this logic would allow for multiple ABtests at once, and the logic would be modular, hidden away elsewhere.

paglias: I think for now this is fine. If we decide to have multiple tests at the
same time or with more options we can expand it later. Let's not over
engineer

Il 11 set 2016 17:47, "Kaitlin Hipkin" notifications@github.com ha
scritto:

crookedneighbor: Agreed. Since this is all private, we don't have to worry about breaking changes. (Before, when we were going to expose it to the API, we would have needed to worry about it)

SabreCat:
Added a new helper function based on `checkExistence` and put the property private again. How's that?

paglias:
LGTM 👍

test/api/v3/integration/user/auth/POST-register_local.test.js
```diff
@@ -52,6 +53,22 @@ describe('POST /user/auth/local/register', () => {
...
+      let password = 'password';
+
+      let user = await api.post('/user/auth/local/register', {
+        username,
+        email,
+        password,
+        confirmPassword: password,
+      });
+
+      await expect(getProperty('users', user._id, '_ABtest')).to.eventually.exist;
```
crookedneighbor, line 68: I'd do:

``` js
let ABTest = await getProperty('users', user._id, '_ABtest');

expect(ABTest).to.be.a('string');
```

crookedneighbor, line 68: Actually, this is the only thing I'd change, so I'll just make the change and merge it in. (two async calls means it takes twice as long, which is why I'd keep it at one async call)

PR #7985, opened 2016-09-08
Optional feedback on account deletion  {continued in #8750}

### LINKED ISSUE #7942: optional feedback in account deletion confirmation screen

JSchneidler:
Fixes #7942 
### Changes

Added text area to account deletion modal for user to submit optional feedback.

Mail options on back-end were left basic and not sent as requested.

Here is what it looks like:  Feedback Modal

---


website/server/controllers/api-v3/user.js
```diff
@@ -223,6 +224,17 @@ api.deleteUser = {
...
 
+    if (req.body.feedback) {
+      let mailOptions = {
+        from: '"Habitica" <noreply@habitica.com>',
+        to: 'admin@habitica.com',
+        subject: 'Feedback received',
+        html: '<b>Feedback:</b> ${req.body.feedback}',
+      };
+
+      //send(mailOptions);
```
crookedneighbor: This is supposed to be commented out?

crookedneighbor:
We should still state that feedback is anonymous.

website/server/controllers/api-v3/user.js
JSchneidler: Yes, I just commented out the sending part since I know that will be changed.

JSchneidler:
Would this be good? "Feedback (optional) Note: Can be in whichever language you feel comfortable with. All feedback is anonymous unless contact or identifying information is provided."

Alys:
Let's try something like this, which I think is a bit more in the tone Lemoness would use (other suggestions welcome):

"If you'd like to give us feedback, please enter it below - we'd love to know what you liked or didn't like about Habitica! It will be anonymous unless you choose to enter your contact details. Don't speak English well? No problem! Use the language you prefer."

It shouldn't be bolded because that draws attention away from the deletion instructions, which are more important. In fact, now that that screen is bigger, adding bold around the first paragraph is probably a good idea.

Can you please use admin@habitica.com instead of noreply@habitica.com? That way if we accidentally reply, our own email will come back to us instead of getting lost, and we'll realise what we've done. :)

crookedneighbor:
@JSchneidler What's your status on this?

JSchneidler:
Almost done, trying to figure out how to add a stylus file and have it included.

JSchneidler:
Made the changes mentioned above. Only thing not working is the CSS rule for `#feedback`, I imported `settings.styl` into `index.styl` but it is not being included in the final `app.css` file for some reason.

website/views/shared/modals/settings.jade
```diff
@@ -66,6 +66,11 @@ script(type='text/ng-template', id='modals/delete.html')
     .row
       .col-md-6
         input.form-control(type='password', ng-model='_deleteAccount')
+    br
+    .row
+      .col-xs-12.form-group
+        label(for='feedback')=env.t('feedback')
+        textarea#feedback.form-control(ng-model='feedback')
```
crookedneighbor: Feedback is a textarea, so your bold styles will not apply here. I think maybe you meant to put the `feedback` id on the surrounding `.col-xs-12`?

crookedneighbor:
You've got a few linting errors as well.

JSchneidler:
The linting errors are caused by the unused mail send. Should I just remove this entirely? Also, not sure why the new CSS isn't working still..

crookedneighbor (changes requested):

website/views/shared/modals/settings.jade
JSchneidler: Haha yeah, its been a rough transition back into web dev. Thanks for pointing that out and taking the time to make this easier on me, I really appreciate it.

JSchneidler:
This is just about done. My Stylus changes just don't seem to be getting copied to `app.css`. Most likely something simple..

crookedneighbor:
@paglias What's the progress on updating the email template for this?

website/server/controllers/api-v3/user.js
```diff
@@ -223,6 +223,10 @@ api.deleteUser = {
 
     await user.remove();
 
+    if (req.body.feedback) {
+      // Send feedback via email
```
TheHollidayInn, line 227: @crookedneighbor  Does this just need a call to `sendTxnEmail`?

TheHollidayInn: For reference, this is what needs to be added: https://github.com/HabitRPG/habitica/blob/develop/website/server/controllers/api-v3/chat.js#L266 . I'll add it if this isn't updated soon.

website/server/controllers/api-v3/user.js
```diff
@@ -223,6 +223,10 @@ api.deleteUser = {
 
     await user.remove();
 
+    if (req.body.feedback) {
```
paglias, line 226: This new parameter needs to be documented in the comment block and some validation would not hurt too (limit length, ..)

paglias (changes requested):

website/server/controllers/api-v3/user.js
paglias: yes that's the call that is needed

Alys: The code used to send the email can be this:
```
      txnEmail(TECH_ASSISTANCE_EMAIL, 'admin-feedback', [
        {name: 'PROFILE_NAME', content: user.profile.name},
        {name: 'UUID', content: user._id},
        {name: 'EMAIL', content: getUserInfo(user, ['email']).email},
        {name: 'FEEDBACK_SOURCE', content: 'from deletion form')},
        {name: 'FEEDBACK', content: req.body.feedback},
      ]);
```
That's based on similar code in `website/server/libs/payments.js`  (search in that file for `admin-user-subscription-details`) so you can copy related code from that file, e.g., for defining `txnEmail` and `TECH_ASSISTANCE_EMAIL`.

I've created a template in Mandrill called `admin-feedback`, which expects the parameters listed in that code block.

The FEEDBACK_SOURCE parameter should have just the text I've shown in that code block, although if you want to configure it as a constant, that's fine. The string will only ever be used in that one place though. The idea of the FEEDBACK_SOURCE parameter is that in future we can use the same email template in other locations, and adjust only the FEEDBACK_SOURCE value so that the admins who receive the email know what's triggered the feedback.

website/common/locales/en/settings.json
```diff
@@ -62,8 +63,8 @@
...
+    "resetText2": "You will lose all your levels, gold, and experience points. All your tasks (except those from challenges) will be deleted permanently and you will lose all of their historical data. You will lose all your equipment but you will be able to buy it all back, including all limited edition equipment or subscriber Mystery items that you already own (you will need to be in the correct class to re-buy class-specific gear). You will keep your current class and your pets and mounts. You might prefer to use an Orb of Rebirth instead, which is a much safer option and which will preserve your tasks.",
+    "deleteLocalAccountText": "<b>Are you sure? This will delete your account forever, and it can never be restored! You will need to register a new account to use Habitica again. Banked or spent Gems will not be refunded. If you're absolutely certain, type your password into the text box below.</b>",
```
Alys, line 67: @JSchneidler Can you please add "and equipment" back on to the end of `resetText2`? The Orb of Rebirth no longer removes gear.

For `deleteLocalAccountText`, can you do the bolding in the Jade file that uses that string instead of having a tag in the locales files? That will avoid any possibility of the translators modifying the bold tag, and also means we can change the formatting later without needing to duplicate the change in all the language files.

How are you going with this PR? Do you have any questions about any of it or need any advice?

SabreCat:
Continued in #8750.

PR #7986, opened 2016-09-08
Use npm package instead of github for @slack/client.

JSchneidler:
### Changes

Changed `@slack/client` dependency in `package.json` to use npm instead of github. Package is now released on npm and fixed a bug for me and possibly others not being able to download the module through github.

---


crookedneighbor:
:+1:

crookedneighbor:

PR #7988, opened 2016-09-09
Avatar's eyes can be set to open or closed in modals while in the Inn fixes #6282

### LINKED ISSUE #6282: Avatar's eyes are closed in modals while in the Inn
If you're checked in to the Inn, the various modals for congratulations and social media sharing show the closed-eyes version of the avatar:

image

There's a parameter when displaying `generatedAvatar` to force the avatar's eyes closed regardless of Inn status, for the purposes of the "knocked out" look in the 0 Health modal, but it might be valuable to also manage the opposite (always show open eyes). Perhaps `null` for the parameter shows the avatar in whatever their current state actually is, while `true` and `false` force open/closed.

abatula:
Fixes 6282
### Changes

Avatar's eyes can be specified to show as open, closed, or to match their current Inn status by setting the `sleep` variable in `generatedAvatar` to false, true, or null (respectively). Now the avatar can be specified to be displayed with open eyes even if the user is in the Inn (e.g. when levelling up).

I did not change any calling functions to display the modal with eyes open in this commit, only local modifications to generate the images below. But I can also make those changes if desired.

Old Functionality: User in the Inn
old11

Old Functionality: User not in the Inn
old00

New Functionality: User in the Inn, `sleep=true` and `sleep=null`
new1n

New Functionality: User in the Inn, `sleep=false`
new10

New Functionality: User not in the Inn, `sleep=true`
new01

New Functionality: User not in the Inn, `sleep=false`  and `sleep=null`
## new0n

### DIFF
--- a/website/views/shared/avatar/generated_avatar.jade
+++ b/website/views/shared/avatar/generated_avatar.jade
@@ -10,6 +10,8 @@ mixin generatedAvatar(options)
   +costumeSetting('back')
   if options.sleep
     span(ng-class="'skin_' + profile.preferences.skin + '_sleep'")
+  else if options.sleep == false
+    span(ng-class="'skin_' + profile.preferences.skin")
   else
     span(ng-class="profile.preferences.sleep ? 'skin_' + profile.preferences.skin + '_sleep' : 'skin_' + profile.preferences.skin")
   span(class='{{profile.preferences.size}}_shirt_{{profile.preferences.shirt}}')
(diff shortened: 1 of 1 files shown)


website/views/shared/avatar/generated_avatar.jade
```diff
@@ -10,6 +10,8 @@ mixin generatedAvatar(options)
   +costumeSetting('back')
   if options.sleep
```
crookedneighbor: Let's be exploit. Have the first option check for true. Also, use triple equals.

crookedneighbor: S/exploi/explicit

abatula:
Explicitly check if `sleep` parameter is true, and switched to triple equals for comparisons.

crookedneighbor:
I applied this to a few of the generatedAvatar mixins b8878df6bdc3080ae54ed8eb482c32e7d262828f

PR #7989, opened 2016-09-10
Pull out animal content into separate files

crookedneighbor:
Pulls out eggs, hatching potions, pets, and mounts content into separate files.

Adds a new `petsInfo` and `mountsInfo` which are objects with the full name of the pet/mount with data about the pet/mount. IE, you can now use the currentPet attribute with this file to get the translated name of your pet. You don't need to know what type of animal it is to do this. No more needing to do data munging.

Should make the pet and mount logic in this PR significantly easier: https://github.com/HabitRPG/habitrpg/pull/7768/files

Ping @MathWhiz


librarianmage:
Yay! How should I use it?

crookedneighbor:
Well, you should wait until it's merged in. :smile:

I must have a typo or accidentally deleted something because the tests around pets se failing. Will check tomorrow.

librarianmage:
I know, I was just wondering how I would use it after it is merged in.

common/script/content/eggs.js
```diff
@@ -0,0 +1,262 @@
...
+
+function hasQuestAchievementFunction (key) {
+  return (user) => {
+    return user.achievements.quests &&
+      user.achievements.quests[key] > 0;
+  };
+}
+
+let drops = {
+  Wolf: {
```
khipkin, line 31: Why are some `*Cub` but not all?

crookedneighbor: It was a decision made when this feature was first introduced (before my time) and now we can't change it because of backwards compatibility.

khipkin: Idea: Backward compatibility could be maintained if new fields with a consistent naming convention were used to just map to the old fields.

crookedneighbor: The problem is the way the content api works. Let's say we decided to drop the `Cub` convention, but maintain the keys for the ones that had them before, then you'd then have this:

``` js
let eggs = Object.keys(content.eggs);
eggs; // ['Wolf', 'TigerCub', 'Tiger', 'PandaCub', 'Panda', 'LionCub', 'Lion', 'Fox', 'FlyingPig', 'Dragon', 'Cactus', 'BearCub', 'Bear']
```

So if you were to loop over the eggs to display the different kind of drop eggs available, you'd suddenly list duplicates of a few.

crookedneighbor: It's something we _should_ have addressed in the v2 -> v3 switch, but ultimately, I don't think that it is that big of a deal.

khipkin: Just gonna have to wait until api v4. :laughing:

crookedneighbor: Pretty much. :)

librarianmage:
So how do I use this?

crookedneighbor:
I found a bug right after merging this in, trying to fix it.

PR #7994, opened 2016-09-10
drops are randomly selected, not based on user values fixes #7929

### LINKED ISSUE #7929: Change Drop Calculation To Not Rely on Seeded Stats from User

gambtho:
Fixes https://github.com/HabitRPG/habitrpg/issues/7929
### Changes

Uses math.random instead of predictable random from drop calculations

Previous discussion - https://github.com/HabitRPG/habitrpg/pull/7967#issuecomment-246126330

---

UUID:  349ea7be-f378-4593-8702-672f75fa83f9

### DIFF
--- a/test/common/fns/randomVal.js
+++ b/test/common/fns/randomVal.js
@@ -1,10 +1,14 @@
 import randomVal from '../../../website/common/script/fns/randomVal';
+
 import {
   generateUser,
 } from '../../helpers/common.helper';
 
 describe('shared.fns.randomVal', () => {
   let user;
+  let predictableRandom = () => {
+    return 0.5;
+  };
   let obj = {
     a: 1,
     b: 2,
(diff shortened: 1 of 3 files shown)


common/script/fns/randomVal.js
```diff
@@ -1,12 +1,15 @@
...
 
 // Get a random property from an object
 // returns random property (the value)
 
-module.exports = function randomVal (user, obj, options = {}) {
+function randomGenerator (user, seed, providedRandom) {
+  return providedRandom ? providedRandom(user, seed) : Math.random();
+}
+
+module.exports = function randomVal (user, obj, options = {}, randomFunc) {
```
crookedneighbor: randomFunc should be in the options object, not a fourth argument.

test/common/fns/randomVal.test.js
```diff
@@ -0,0 +1,83 @@
+import randomVal from '../../../common/script/fns/randomVal';
```
crookedneighbor: It's correct to rename this file to `randomVal.test.js`, but to make reviewing the changes easier, let's leave it as is.

gambtho:
any other changes needed here?

crookedneighbor:
There's one thing I want to check first. Stand by.

crookedneighbor:
I thought this issue might be blocking this https://github.com/HabitRPG/habitrpg/issues/7851, but after re-reading it, I don't think it is.

gambtho:
fixed merge conflict

crookedneighbor:
Thanks for your work on this. I've noted this toward your next tier.

PR #7996, opened 2016-09-11
Group tasks ui picked

TheHollidayInn:
Fixes put_issue_url_here
### Changes

This adds the task ui to groups page
This also adds the assign field to task
_note_ The UI is still a little large and odd, but I would like feedback before I make any major changes. Also, I am unsure if all four tasks list should be required. 

---

UUID:


crookedneighbor:
Is this UI ready to be seen by users?

paglias:
Can we have screenshots?

Il 11 set 2016 17:18, "Blade Barringer" notifications@github.com ha
scritto:

bower.json
```diff
@@ -43,7 +43,9 @@
     "smart-app-banner": "78ef9c0679723b25be1a0ae04f7b4aef7cbced4f",
     "habitica-markdown": "1.2.2",
     "pusher-js-auth": "^2.0.0",
-    "pusher-websocket-iso": "pusher#^3.1.0"
+    "pusher-websocket-iso": "pusher#^3.1.0",
+    "ng-tags-input": "^3.1.1",
```
crookedneighbor: I'm wary about adding additional dependencies so close to the client redesign. :/

TheHollidayInn: We still have a while, imo. But, I can remove this ng dependency. I don't need it anymore. The taggle is requried - yet it is a js/jquery plugin.

Alys: All four tasks lists should definitely be included.

TheHollidayInn:
I posted the screenshots in Slack so we can discuss changes in there. I figured there would be many comments.

TheHollidayInn:
Updated.

website/views/shared/tasks/index.jade
```diff
@@ -6,7 +6,7 @@ include ./task_view/mixins
 script(id='templates/habitrpg-tasks.html', type="text/ng-template")
   .tasks-lists.container-fluid
     .row
-      .col-md-3.col-sm-6(ng-repeat='list in lists', ng-class='::{ "rewards-module": list.type==="reward", "new-row-md": list.type==="todo" }')
+      .col-md-3.col-sm-6(ng-repeat='list in lists', ng-class='::{ "rewards-module": list.type==="reward", "new-row-md": list.type==="todo", "col-md-2": obj.type }')
```
paglias, line 9: what is this for?

website/server/controllers/api-v3/tasks.js
```diff
@@ -629,7 +642,7 @@ api.addTagToTask = {
     if (validationErrors) throw validationErrors;
 
     let taskId = req.params.taskId;
-    let task = await Tasks.Task.findByIdOrAlias(taskId, user._id, { userId: user._id });
```
paglias, line 631: removing the third argument this route can be used by anyone that knows the taskId, even if they don't have the right because the query will be `{_id: taskId}` when taskId is an UUID

paglias (changes requested):

website/client-old/merch/teespring-logo.svg
```diff
@@ -1,63 +1,63 @@
-<?xml version="1.0" encoding="utf-8"?>
```
paglias, line 1: Did you want to change this file?

website/client-old/manifest.json
```diff
@@ -30,9 +30,12 @@
 
             "bower_components/jquery-ui/ui/minified/jquery.ui.core.min.js",
             "bower_components/jquery-ui/ui/minified/jquery.ui.widget.min.js",
+            "bower_components/jquery-ui/ui/minified/jquery.ui.menu.min.js",
```
paglias: how big is all the jquery ui stuff? Also I think we already have an autocomplete widget, the one we're using in chat, is is necessary to add this one?

paglias (changes requested):

website/views/shared/tasks/index.jade
TheHollidayInn: This makes the columns fit on the group page.

website/server/controllers/api-v3/tasks.js
TheHollidayInn: Don't we check permissions outside of the task query? And, either way, is the better practice to leave the user id blank for group and challenge tasks? Because I check for group permission below.

website/client-old/merch/teespring-logo.svg
TheHollidayInn: Eh, no. this must be when I was trying to fix my line endings.

website/server/controllers/api-v3/tasks.js
paglias: Where are you checking for permission in this route? Also tags can be added only to user tasks, not to challenge and groups original tasks since it wouldn't make sense and the values come from user.tags

website/client-old/manifest.json
TheHollidayInn: The other autocomplete has a bunch of logic based on the chat. These two jquery items are small since I didn't have the full jQuery. And it is probably better to have a separate directive for this autocomplete vs the chat autocomplete.

paglias (changes requested):

website/client-old/manifest.json
paglias: Hmm I see it's not the whole thing, it's fine then

TheHollidayInn:
Actually, your right. I don't think I have that in this route. I asked in Slack if we should have tasks on group original, and the answer was yes. We can remove it though, it doesn't make much sense except for it provides default.

website/client-old/js/app.js
```diff
@@ -84,11 +84,6 @@ window.habitrpg = angular.module('habitrpg',
           templateUrl: "partials/options.profile.stats.html",
           title: env.t('titleStats')
         })
-        .state('options.profile.achievements', {
```
paglias, line 87: is this normal?

website/client-old/js/app.js
```diff
@@ -102,9 +97,8 @@ window.habitrpg = angular.module('habitrpg',
         })
 
         .state('options.social.inbox', {
-          url: '/inbox',
```
paglias, line 105: same here

website/client-old/js/components/groupMembersAutocomplete/groupMembersAutocompleteDirective.js
```diff
@@ -0,0 +1,73 @@
...
+            allowedTags: currentTags,
+            allowDuplicates: false,
+            onBeforeTagAdd: function(event, tag) {
+              return confirm('You really wanna add ' + tag + '?');
+            },
+            onTagAdd: function(event, tag) {
+              $rootScope.$broadcast('addedGroupMember', memberProfileNameToIdMap[tag]);
+            },
+            onBeforeTagRemove: function(event, tag) {
+              return confirm('You really wanna remove ' + tag + '?');
```
paglias: needs a string

website/client-old/js/components/groupTasks/groupTasksController.js
```diff
@@ -0,0 +1,147 @@
...
+      }
+      task._edit = undefined;
+
+      if (task.checklist) {
+        task.checklist = _.filter(task.checklist, function (i) {
+          return !!i.text
+        });
+      }
+
+      User.updateTask(task, {body: task});
```
paglias: is User right here?

website/client-old/js/components/groupTasks/groupTasksController.js
```diff
@@ -0,0 +1,147 @@
+habitrpg.controller('GroupTasksCtrl', ['$scope', 'Shared', 'Tasks', 'User', function ($scope, Shared, Tasks, User) {
```
paglias, line 1: The code in this file seems mostly duplicate, there's a way to share it with the tasks controller user for users?

website/views/options/social/group.jade
```diff
@@ -147,11 +147,26 @@ a.pull-right.gem-wallet(ng-if='group.type!="party"', popover-trigger='mouseenter
                 h3.popover-title {{group.leader.profile.name}}
                 .popover-content
                   markdown(text='group._editing ? groupCopy.leaderMessage : group.leaderMessage')
-      div(ng-controller='ChatCtrl')
+                  
+      ul.options-menu(ng-init="groupPane = 'chat'", ng-show="group.purchased.active")
+        li
+          a(ng-click="groupPane = 'chat'")
```
paglias, line 153: I think these two panes need a route, group/chat and group/tasks

website/views/options/social/groups/group-tasks-actions.jade
```diff
@@ -0,0 +1,5 @@
+script(type='text/ng-template', id='partials/groups.tasks.actions.html')
+  div(ng-if="group.leader._id === user._id", class="col-md-12")
+    strong Assign Task
```
paglias: string

website/views/options/social/groups/group-tasks.jade
```diff
@@ -0,0 +1,5 @@
+include ./group-tasks-actions
+include ./group-members-autocomplete
+
+script(type='text/ng-template', id='partials/groups.tasks.html')
+  habitrpg-tasks(main=false, obj='newChallenge')
```
paglias: is newChallene correct?

website/client-old/manifest.json
```diff
@@ -30,9 +30,12 @@
```
paglias, line 30: Since these changes aren't going online I'd like all these new files not to be loaded, maybe you can remove them and then re-add them in the next PR?

paglias (changes requested):

website/client-old/js/components/groupTasks/groupTasksController.js
TheHollidayInn: Yes, I was going to abstract everything to the Task Service which would be shared between the challenge, user and group. But, since we are refactoring, I tried not to change much.

website/views/options/social/groups/group-tasks.jade
TheHollidayInn: I don't think that does anything. It looks like we set the obj via the route, but I will remove it here for clarification.

website/client-old/manifest.json
TheHollidayInn: Can do.

website/views/options/social/group.jade
TheHollidayInn: I thought about that. But, chat is the default, so that would make the default group url always have chat. Plus, groups with out a subscription would always have the chat url as well. I think maybe in the future it would be a good idea.

website/client-old/js/components/groupTasks/groupTasksController.js
TheHollidayInn: It works, but I will use the Tasks service for clairty

website/client-old/manifest.json
TheHollidayInn: Actually, I do have 3 more PRs before this will go live. Do you want me to remove and add the dependencies throughout each PR? It does seem a little odd :P

website/views/options/social/group.jade
paglias: okay, let's revisit for the redesign

website/client-old/js/components/groupTasks/groupTasksController.js
paglias: I agree about not doing big refactors but this file is really almost the same and we'll have to remember to do any change in two places until the redesign. Let's move what's possible in a service

website/client-old/manifest.json
paglias: I don't know... I don't like to load code that's unused @crookedneighbor what do you think?

TheHollidayInn: Alright, I thought of an easy way to do this.

website/client-old/js/components/groupTasks/groupTasksController.js
TheHollidayInn: Cool, cool. I can do that.

paglias (approved):

PR #7997, opened 2016-09-11
Disallow interactions by blocked users; some interaction API refactoring (fixes #7971) {Continued in #8755}

### LINKED ISSUE #7971: flags.chatRevoked does not prevent user sending private messages (PMs)
When a user's chat permissions need to be revoked, they are given a true value for `flags.chatRevoked`. This stops them posting to any guild or the Tavern.

It should also stop them sending private messages to any player.

CrystallineCat:
Fixes #7971

This PR changes the way objections to user-to-user interactions are handled (e.g. chat revocation or blocks). It includes some refactoring to the API to make it more flexible and allows the UI to query the estimated success of interactions. No UI changes are done at this time.

API refactoring:
- Factors out the checks in `api.sendPrivateMessage` and `api.transferGems` to see if the interaction is allowed.
- The checks are now in `User.getObjectionsToInteractionIfAny`; the method is designed to be easily extensible to new checks or new interactions.
- To allow the UI to check whether an interaction is allowed without actually preforming it, there is a new function `api.getObjectionsToInteractionIfAny`; it only applies interactions initiated by the logged in user and can not be used to query interactions by other users.
- All objections to an interaction now use NotAuthorized instead of NotFound.
- All objection strings are now translatable.

Additional checks:
- Both `api.sendPrivateMessage` and `api.transferGems` now are now blocked by `sender.flags.chatRevoked`.
- `api.transferGems` is now blocked by either user blocking the other.

Tests:
- Tests for the additional checks and the new API have been added.

---

UUID: fea892e5-ac99-4b03-83ac-d01a2154e2af

### DIFF
--- a/test/api/v3/integration/members/GET-objections_to_interaction.test.js
+++ b/test/api/v3/integration/members/GET-objections_to_interaction.test.js
@@ -38,14 +38,13 @@ describe('GET /members/:toUserId/objections-to/:interaction', () => {
     await expect(
       user.get(`/members/${receiver._id}/objections-to/hug-a-whole-forest-of-trees`)
     ).to.eventually.be.rejected.and.eql({
-      code: 404,
-      error: 'NotFound',
-      message: 'Unknown kind of interaction: "hug-a-whole-forest-of-trees", expected one of send-private-message, transfer-gems',
-      // FIXME: Very fragile message, not sure what to do here
+      code: 400,
+      error: 'BadRequest',
+      message: t('invalidReqParams'),
     });
   });
 
-  it('returns nothing if there are no objections', async () => {
+  it('returns an empty array if there are no objections', async () => {
     let receiver = await generateUser();
 
     await expect(
--- a/test/api/v3/integration/members/GET-objections_to_interaction.test.js
+++ b/test/api/v3/integration/members/GET-objections_to_interaction.test.js
@@ -3,7 +3,6 @@ import {
   translate as t,
 } from '../../../../helpers/api-v3-integration.helper';
 import { v4 as generateUUID } from 'uuid';
-// import common from '../../../../../common';
 
 describe('GET /members/:toUserId/objections-to/:interaction', () => {
   let user;
--- a/test/api/v3/integration/members/GET-objections_to_interaction_if_any.test.js
+++ b/test/api/v3/integration/members/GET-objections_to_interaction_if_any.test.js
@@ -1,0 +1,56 @@
+import {
+  generateUser,
+  translate as t,
+} from '../../../../helpers/api-v3-integration.helper';
+import { v4 as generateUUID } from 'uuid';
+// import common from '../../../../../common';
+
+describe('GET /members/:toUserId/objections-to/:interaction', () => {
+  let user;
+
+  before(async () => {
+    user = await generateUser();
+  });
+
+  it('validates req.params.memberId', async () => {
+    await expect(
+      user.get('/members/invalidUUID/objections-to/send-private-message')
+    ).to.eventually.be.rejected.and.eql({
+      code: 400,
+      error: 'BadRequest',
+      message: t('invalidReqParams'),
+    });
+  });
+
+  it('handles non-existing members', async () => {
+    let dummyId = generateUUID();
+    await expect(
+      user.get(`/members/${dummyId}/objections-to/send-private-message`)
+    ).to.eventually.be.rejected.and.eql({
+      code: 404,
+      error: 'NotFound',
+      message: t('userWithIDNotFound', {userId: dummyId}),
+    });
+  });
+
+  it('handles non-existing interactions', async () => {
+    let receiver = await generateUser();
+
+    await expect(
+      user.get(`/members/${receiver._id}/objections-to/hug-a-whole-forest-of-trees`)
+    ).to.eventually.be.rejected.and.eql({
+      code: 404,
+      error: 'NotFound',
+      message: 'Unknown kind of interaction: "hug-a-whole-forest-of-trees", expected one of send-private-message, transfer-gems',
+      // FIXME: Very fragile message, not sure what to do here
+    });
+  });
+
+  it('returns nothing if there are no objections', async () => {
+    let receiver = await generateUser();
+
+    await expect(
+      user.get(`/members/${receiver._id}/objections-to/send-private-message`)
+    ).to.eventually.be.fulfilled.and.eql([]);
+  });
+});
--- a/test/api/v3/integration/members/POST-transfer_gems.test.js
+++ b/test/api/v3/integration/members/POST-transfer_gems.test.js
@@ -87,21 +87,6 @@ describe('POST /members/transfer-gems', () => {
     });
   });
 
-  it('returns error when to user has opted out of messaging', async () => {
-    // FIXME: Not sure this is wanted
-    let receiverOptOut = await generateUser({'inbox.optOut': true});
-
-    await expect(userToSendMessage.post('/members/transfer-gems', {
-      message,
-      gemAmount,
-      toUserId: receiverOptOut._id,
-    })).to.eventually.be.rejected.and.eql({
-      code: 401,
-      error: 'NotAuthorized',
-      message: t('notAuthorizedToSendMessageToThisUser'),
-    });
-  });
-
   it('returns an error when chat privileges are revoked', async () => {
     let userWithChatRevoked = await generateUser({'flags.chatRevoked': true});
 
--- a/test/api/v3/integration/members/POST-transfer_gems.test.js
+++ b/test/api/v3/integration/members/POST-transfer_gems.test.js
@@ -59,6 +59,63 @@ describe('POST /members/transfer-gems', () => {
     });
   });
 
+  it('returns error when to user has blocked the sender', async () => {
+    let receiverBlocks = await generateUser({'inbox.blocks': [userToSendMessage._id]});
+
+    await expect(userToSendMessage.post('/members/transfer-gems', {
+      message,
+      gemAmount,
+      toUserId: receiverBlocks._id,
+    })).to.eventually.be.rejected.and.eql({
+      code: 401,
+      error: 'NotAuthorized',
+      message: t('notAuthorizedToSendMessageToThisUser'),
+    });
+  });
+
+  it('returns error when sender has blocked to user', async () => {
+    let sender = await generateUser({'inbox.blocks': [receiver._id]});
+
+    await expect(sender.post('/members/transfer-gems', {
+      message,
+      gemAmount,
+      toUserId: receiver._id,
+    })).to.eventually.be.rejected.and.eql({
+      code: 401,
+      error: 'NotAuthorized',
+      message: t('notAuthorizedToSendMessageToThisUser'),
+    });
+  });
+
+  it('returns error when to user has opted out of messaging', async () => {
+    // FIXME: Not sure this is wanted
+    let receiverOptOut = await generateUser({'inbox.optOut': true});
+
+    await expect(userToSendMessage.post('/members/transfer-gems', {
+      message,
+      gemAmount,
+      toUserId: receiverOptOut._id,
+    })).to.eventually.be.rejected.and.eql({
+      code: 401,
+      error: 'NotAuthorized',
+      message: t('notAuthorizedToSendMessageToThisUser'),
+    });
+  });
+
+  it('returns an error when chat privileges are revoked', async () => {
+    let userWithChatRevoked = await generateUser({'flags.chatRevoked': true});
+
+    await expect(userWithChatRevoked.post('/members/transfer-gems', {
+      message,
+      gemAmount,
+      toUserId: receiver._id,
+    })).to.eventually.be.rejected.and.eql({
+      code: 401,
+      error: 'NotAuthorized',
+      message: t('chatPrivilegesRevoked'),
+    });
+  });
+
   it('returns error when there is no gemAmount', async () => {
     await expect(userToSendMessage.post('/members/transfer-gems', {
       message,
--- a/test/api/v3/integration/members/POST-transfer_gems.test.js
+++ b/test/api/v3/integration/members/POST-transfer_gems.test.js
@@ -60,12 +60,12 @@ describe('POST /members/transfer-gems', () => {
   });
 
   it('returns error when to user has blocked the sender', async () => {
-    let receiverBlocks = await generateUser({'inbox.blocks': [userToSendMessage._id]});
+    let receiverWhoBlocksUser = await generateUser({'inbox.blocks': [userToSendMessage._id]});
 
     await expect(userToSendMessage.post('/members/transfer-gems', {
       message,
       gemAmount,
-      toUserId: receiverBlocks._id,
+      toUserId: receiverWhoBlocksUser._id,
     })).to.eventually.be.rejected.and.eql({
       code: 401,
       error: 'NotAuthorized',
@@ -99,6 +99,22 @@ describe('POST /members/transfer-gems', () => {
       error: 'NotAuthorized',
       message: t('chatPrivilegesRevoked'),
     });
+  });
+
+  it('works when only the receiver\'s chat privileges are revoked', async () => {
+    let receiverWithChatRevoked = await generateUser({'flags.chatRevoked': true});
+
+    await expect(userToSendMessage.post('/members/transfer-gems', {
+      message,
+      gemAmount,
+      toUserId: receiverWithChatRevoked._id,
+    })).to.eventually.be.fulfilled;
+
+    let updatedReceiver = await receiverWithChatRevoked.get('/user');
+    let updatedSender = await userToSendMessage.get('/user');
+
+    expect(updatedReceiver.balance).to.equal(gemAmount / 4);
+    expect(updatedSender.balance).to.equal(0);
   });
 
   it('returns error when there is no gemAmount', async () => {
--- a/website/server/controllers/api-v3/members.js
+++ b/website/server/controllers/api-v3/members.js
@@ -269,21 +269,43 @@ api.getChallengeMemberProgress = {
   },
 };
 
-// Throws an exception if for any reason the sender is not allowed to interact with
-// the receiver.
-function _throwUnlessInteractionAllowedBetween (sender, receiver, res) {
-  if (sender.flags.chatRevoked) {
-    throw new NotAuthorized(res.t('chatPrivilegesRevoked'));
-  }
+/**
+ * @api {get} /api/v3/members/:toUserId/objections-to/:interaction Get the message of any errors that would occur if the given interaction was attempted
+ * @apiVersion 3.0.0
+ * @apiName GetObjectionsToInteractionIfAny
+ * @apiGroup Member
+ *
+ * @apiParam {UUID} toUserId The user to interact with
+ * @apiParam {String} interaction Name of the interaction to query, e.g. "send-private-message"
+ *
+ * @apiSuccess {Array} data Return an array of error messages, if the interaction would be blocked; otherwise an empty array
+ */
+api.getObjectionsToInteractionIfAny = {
+  method: 'GET',
+  url: '/members/:toUserId/objections-to/:interaction',
+  middlewares: [authWithHeaders()],
+  async handler (req, res) {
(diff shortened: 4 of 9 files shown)


test/api/v3/integration/members/GET-objections_to_interaction_if_any.test.js
```diff
@@ -0,0 +1,56 @@
+import {
+  generateUser,
+  translate as t,
+} from '../../../../helpers/api-v3-integration.helper';
+import { v4 as generateUUID } from 'uuid';
+// import common from '../../../../../common';
```
khipkin: Suggestion: delete this line. :)

website/server/models/user/methods.js
```diff
@@ -20,6 +20,49 @@ schema.methods.getGroups = function getUserGroups () {
...
+      // Unlike private messages, gems can't be sent to oneself
+      (sndr, rcvr) => rcvr._id === sndr._id && 'cannotSendGemsToYourself',
+    ],
+  };
+  /* eslint-enable no-unused-vars */
+
+  let knownInteractions = Object.keys(checks).filter((k) => k !== 'always');
+
+  if (!knownInteractions.includes(interaction)) {
+    throw new Error(`Unknown kind of interaction: "${interaction}", expected one of ${knownInteractions.join(', ')}`);
```
khipkin: Suggestion: add this string to a file in the common/locales/en directory with an "interaction" parameter.  The known interactions can be manually added (probably) because there are a known, finite number of allowed interactions.

(You should verify this with someone on the Habitica staff before implementing though. :smile: )

test/api/v3/integration/members/POST-transfer_gems.test.js
```diff
@@ -59,6 +59,48 @@ describe('POST /members/transfer-gems', () => {
     });
   });
 
+  it('returns error when to user has blocked the sender', async () => {
+    let receiverBlocks = await generateUser({'inbox.blocks': [userToSendMessage._id]});
```
khipkin: Maybe it's just me, but this variable name is confusing me.

test/api/v3/integration/members/POST-transfer_gems.test.js
```diff
@@ -59,6 +59,48 @@ describe('POST /members/transfer-gems', () => {
...
+      gemAmount,
+      toUserId: receiver._id,
+    })).to.eventually.be.rejected.and.eql({
+      code: 401,
+      error: 'NotAuthorized',
+      message: t('notAuthorizedToSendMessageToThisUser'),
+    });
+  });
+
+  it('returns an error when chat privileges are revoked', async () => {
```
khipkin, line 90: What about if someone sends gems to someone with chat privileges revoked?

website/server/controllers/api-v3/members.js
```diff
@@ -270,6 +270,44 @@ api.getChallengeMemberProgress = {
...
+    let sender = res.locals.user;
+    let receiver = await User.findById(req.params.toUserId).exec();
+    if (!receiver) throw new NotFound(res.t('userWithIDNotFound', {userId: req.params.toUserId}));
+
+    let response;
+
+    try {
+      response = sender.getObjectionsToInteractionIfAny(req.params.interaction, receiver);
+    } catch (e) {
+      throw new NotFound(e.message);
```
khipkin: Maybe allow the callee to throw this message ifneedbe? And don't catch it here (in the caller)?

website/server/models/user/methods.js
```diff
@@ -20,6 +20,49 @@ schema.methods.getGroups = function getUserGroups () {
   return userGroups;
 };
 
+// Get an array of error message keys that would be thrown if the given interaction was attempted
```
khipkin: Technically, if the given interaction was attempted, only one error would be thrown, since it would be pointless to keep trying once a failure condition was found.  Is it really necessary to return ALL issues with an interaction?  Why not just the first one found (if a problem exists) or none (if the interaction is allowed)?

website/server/models/user/methods.js
```diff
@@ -20,6 +20,49 @@ schema.methods.getGroups = function getUserGroups () {
...
+      // Direct user blocks prevent all interactions
+      (sndr, rcvr) => rcvr.inbox.blocks.includes(sndr._id) && 'notAuthorizedToSendMessageToThisUser',
+      (sndr, rcvr) => sndr.inbox.blocks.includes(rcvr._id) && 'notAuthorizedToSendMessageToThisUser',
+    ],
+
+    'send-private-message': [
+      // Private messaging has an opt-out, which does not affect other interactions
+      (sndr, rcvr) => rcvr.inbox.optOut && 'notAuthorizedToSendMessageToThisUser',
+
+      // NB: We allow a player to message themselves so they can test how PMs work or send their own notes to themselves
```
khipkin: What does NB stand for?

Alys: Nota Bene ("note well"). Let's remove "NB" since it's not necessary and few of us are ancient Romans. :) Sorry, CrystallineCat - entirely my fault. I hadn't thought that it wasn't widely known.

test/api/v3/integration/members/POST-transfer_gems.test.js
Alys: If the recipient has chat permissions revoked, sending gems to them is allowed.

If the sender has site-wide chat privileges revoked, or has been personally blocked by the recipient, then they should not be able to send gems.

Does that cover what you're asking and/or what the code is doing / not doing?  (I haven't checked the code in detail, sorry. :( )

khipkin: I was just suggesting that test coverage for the above situation be added. :smile_cat:

website/server/models/user/methods.js
```diff
@@ -20,6 +20,49 @@ schema.methods.getGroups = function getUserGroups () {
   return userGroups;
 };
 
+// Get an array of error message keys that would be thrown if the given interaction was attempted
+schema.methods.getObjectionsToInteractionIfAny = function getObjectionsToInteractionIfAny (interaction, receiver) {
+  let sender = this;
+
+  /* eslint-disable no-unused-vars */
```
crookedneighbor: Why disable eslint here?

website/server/models/user/methods.js
```diff
@@ -20,6 +20,49 @@ schema.methods.getGroups = function getUserGroups () {
   return userGroups;
 };
 
+// Get an array of error message keys that would be thrown if the given interaction was attempted
+schema.methods.getObjectionsToInteractionIfAny = function getObjectionsToInteractionIfAny (interaction, receiver) {
```
crookedneighbor: Can we remove the `IfAny` from the name? I don't think that's necessary.

website/server/models/user/methods.js
```diff
@@ -20,6 +20,49 @@ schema.methods.getGroups = function getUserGroups () {
...
+    ],
+
+    'transfer-gems': [
+      // Unlike private messages, gems can't be sent to oneself
+      (sndr, rcvr) => rcvr._id === sndr._id && 'cannotSendGemsToYourself',
+    ],
+  };
+  /* eslint-enable no-unused-vars */
+
+  let knownInteractions = Object.keys(checks).filter((k) => k !== 'always');
```
crookedneighbor: Lets avoid one character variables. `filter(check => check !== 'aways');`  should do it.

website/server/models/user/methods.js
crookedneighbor: This would be a developer error, not something that should be exposed to the user, so it shouldn't be added to the locale files.

khipkin: In that case, how should it be handled?

crookedneighbor: Just like this. Throwing a normal error.

website/server/models/user/methods.js
CrystallineCat: Actually, it's not your fault, the NB wasn't in your original line. I added it because the comment looked a bit out of place to me without any code to follow it. I didn't consider it might confuse people, so I agree it's better to remove it. :)

website/server/models/user/methods.js
CrystallineCat: The checks are all of the form `(sndr, rcvr) => ...`, due to being iterated over and called with the same signature, but not all of them use both parameters. This leads to spurious `no-unused-vars` issues.

I think following the linter here would be wrong and make the code worse, not better. I've added a comment stating the reason for the disable.

website/server/models/user/methods.js
Alys: Oh! Huh. :) "NB" is something I use myself so often I just assumed I'd specified that it should be used. :)

website/server/models/user/methods.js
CrystallineCat: I've thought about that, but I'm not sure it's the way to go. I'd rather change the error reporting code to be able to report multiple issues -- the new API function already does that.

Reporting only one when there are multiple seems to me to give a skewed view of the situation, kinda like "fix this one error and it will work", while reporting all gives the full view -- but maybe I'm reading too much into things here.

website/server/controllers/api-v3/members.js
CrystallineCat: The callee is throwing a normal `Error` instead of a HTTP `NotFound` to be independent of HTTP. Unfortunately, an unhandled `Error` will be turned into a generic 500 Internal Server Error and the message lost. This is fine for most API functions, but when the cause of the error can be an API parameter directly, I think the error message needs to be exposed to the user.

website/server/controllers/api-v3/members.js
```diff
@@ -270,6 +270,45 @@ api.getChallengeMemberProgress = {
 };
 
 /**
+ * @api {get} /api/v3/members/:toUserId/objections-to/:interaction Get the message of any errors that would occur if the given interaction was attempted
+ * @apiVersion 3.0.0
+ * @apiName GetObjectionsToInteraction
+ * @apiGroup Member
+ *
+ * @apiParam {UUID} toUserId The user to interact with
+ * @apiParam {String} interaction Name of the interaction to query, e.g. "send-private-message"
```
crookedneighbor: We should enumerate in the API docs all the interactions that can be queried.

CrystallineCat:
@Alys @crookedneighbor @khipkin: Is there anything else I need to do to get this merged? I think I addressed all the comments. Apologies if I missed anything.

website/server/controllers/api-v3/members.js
```diff
@@ -270,6 +270,45 @@ api.getChallengeMemberProgress = {
...
+ * @apiGroup Member
+ *
+ * @apiParam {UUID} toUserId The user to interact with
+ * @apiParam {String="send-private-message","transfer-gems"} interaction Name of the interaction to query
+ *
+ * @apiSuccess {Array} data Return an array of error messages, if the interaction would be blocked; otherwise an empty array
+ */
+api.getObjectionsToInteraction = {
+  method: 'GET',
+  url: '/members/:toUserId/objections-to/:interaction',
```
crookedneighbor, line 288: I'm not thrilled about the name of this url (the `objections-to` part), but I can't think of a better name :(

crookedneighbor: Can we mark this route as beta so we can change it without worrying about breaking 3rd party integrations?

website/server/controllers/api-v3/members.js
```diff
@@ -270,6 +270,45 @@ api.getChallengeMemberProgress = {
 };
 
 /**
+ * @api {get} /api/v3/members/:toUserId/objections-to/:interaction Get the message of any errors that would occur if the given interaction was attempted
```
crookedneighbor: Can we fill out this documentation a bit? See https://github.com/HabitRPG/habitrpg/pull/7879/files#diff-cf049c3d45b117fee868e17845f1f130R8 for an example.

test/api/v3/integration/members/GET-objections_to_interaction.test.js
```diff
@@ -0,0 +1,55 @@
...
+
+  it('handles non-existing interactions', async () => {
+    let receiver = await generateUser();
+
+    await expect(
+      user.get(`/members/${receiver._id}/objections-to/hug-a-whole-forest-of-trees`)
+    ).to.eventually.be.rejected.and.eql({
+      code: 404,
+      error: 'NotFound',
+      message: 'Unknown kind of interaction: "hug-a-whole-forest-of-trees", expected one of send-private-message, transfer-gems',
```
crookedneighbor: This should be a translation string with interpolations:

``` json
"uknownInteraction": "Uknown kind of interaction: \"<%= unknownInteraction %>\". Expected one of <%= interactions %>"
```

``` js
let unknownInteraction = req.params.interaction;
let interactions = 'send-private-message, transfer-gems';
t('uknownInteraction', { unknownInteraction, interactions });
```

test/api/v3/integration/members/GET-objections_to_interaction.test.js
```diff
@@ -0,0 +1,55 @@
...
+      // FIXME: Very fragile message, not sure what to do here
+    });
+  });
+
+  it('returns nothing if there are no objections', async () => {
```
crookedneighbor: Technically it returns an empty array, not nothing.

website/server/controllers/api-v3/members.js
```diff
@@ -270,6 +270,45 @@ api.getChallengeMemberProgress = {
...
+    let receiver = await User.findById(req.params.toUserId).exec();
+    if (!receiver) throw new NotFound(res.t('userWithIDNotFound', {userId: req.params.toUserId}));
+
+    let response;
+
+    try {
+      response = sender.getObjectionsToInteraction(req.params.interaction, receiver);
+    } catch (e) {
+      // Rethrow, so that the message gets passed to the client
+      throw new NotFound(e.message);
```
crookedneighbor: I feel weird about this. Feels like the check for valid params should happen much earlier, before we look up the receiver.

website/server/models/user/methods.js
```diff
@@ -20,6 +20,49 @@ schema.methods.getGroups = function getUserGroups () {
...
+    ],
+
+    'transfer-gems': [
+      // Unlike private messages, gems can't be sent to oneself
+      (sndr, rcvr) => rcvr._id === sndr._id && 'cannotSendGemsToYourself',
+    ],
+  };
+  /* eslint-enable no-unused-vars */
+
+  let knownInteractions = Object.keys(checks).filter((key) => key !== 'always');
```
crookedneighbor: The `checks` and `knownInteractions` variables can be pull out of this function since they are constant and do not need to be recreated for every time this method is called. 

Maybe:

``` js
const INTERACTION_CHECKS = Object.freeze({ /* functions here */ });
const KNOWN_INTERACTIONS = Object.freeze(Object.keys(INTERACTION_CHECKS).filter(key => key !== 'always'));

This means you can also export `KNOWN_INTERACTIONS` for a params check in the route instead of throwing an error here.
```

test/api/v3/integration/members/GET-objections_to_interaction.test.js
crookedneighbor: If using my thought to export the known interactions:

``` js
t('uknownInteraction', { unknownInteraction, interactions: KNOWN_INTERACTIONS.join(',') });
```

website/server/models/user/methods.js
```diff
@@ -20,6 +20,49 @@ schema.methods.getGroups = function getUserGroups () {
...
+
+  if (!knownInteractions.includes(interaction)) {
+    throw new Error(`Unknown kind of interaction: "${interaction}", expected one of ${knownInteractions.join(', ')}`);
+  }
+
+  let checksToRun = checks.always.concat(checks[interaction]);
```
crookedneighbor: Might be better to use lodash chaining for the following lines so we're not creating a ton of extra arrays

website/server/controllers/api-v3/members.js
```diff
@@ -270,6 +270,45 @@ api.getChallengeMemberProgress = {
...
 /**
+ * @api {get} /api/v3/members/:toUserId/objections-to/:interaction Get the message of any errors that would occur if the given interaction was attempted
+ * @apiVersion 3.0.0
+ * @apiName GetObjectionsToInteraction
+ * @apiGroup Member
+ *
+ * @apiParam {UUID} toUserId The user to interact with
+ * @apiParam {String="send-private-message","transfer-gems"} interaction Name of the interaction to query
+ *
+ * @apiSuccess {Array} data Return an array of error messages, if the interaction would be blocked; otherwise an empty array
```
crookedneighbor, line 284: I think it should return an array of objection messages, rather than error messages.

crookedneighbor (changes requested):

website/server/models/user/methods.js
crookedneighbor: ^ How is this going?

website/server/controllers/api-v3/members.js
CrystallineCat: I'm not sure what you mean by "objection messages".

website/server/controllers/api-v3/members.js
CrystallineCat: I added "- BETA" to the comment; is anything else required to mark it as beta?

website/server/controllers/api-v3/members.js
CrystallineCat: I'm not sure there is anything to add. The example is of a method with a much more complicated parameter structure; the current documentation for the new method already includes everything about its parameters. The only missing information is the exception, and I don't think committing to a specific exception here is a good idea.

test/api/v3/integration/members/GET-objections_to_interaction.test.js
CrystallineCat: I decided to move this to the parameter checking with req.checkParams. This provides a less useful error message, but is consistent with the other methods and I'm feeling uneasy about handling this check separately. Ideally, all the parameter checking would be done data-driven, but that's out of scope for this, I think.

website/server/models/user/methods.js
CrystallineCat: I've converted it to lodash. I'm fairly new to JS and haven't used that library before, so I hope I did it the right way. :)

TBH, I'm not sure I'm that concerned about the arrays; the number is constant per call and I don't think the method is going to be a bottleneck, as it is likely only to be called in response to relatively infrequent UI events.

I realize I've gone against some of your requests, I've considered them over and over again and couldn't get myself to agree with them. Apologies if this causes any issues and thank you for your reviews. Also apologies for the huge delay... it took me being shocked from recent events to numb my fear of this conversation enough... :/ I'll gladly step back if you'd prefer someone else to work on this.

test/api/v3/integration/members/GET-objections_to_interaction.test.js
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('returns an empty array if there are no objections', async () => {
+    let receiver = await generateUser();
+
+    await expect(
+      user.get(`/members/${receiver._id}/objections-to/send-private-message`)
+    ).to.eventually.be.fulfilled.and.eql([]);
+  });
+});
```
TheHollidayInn, line 54: You should added a test here to show that the route returns objections

website/server/controllers/api-v3/members.js
TheHollidayInn: I think his is saying that since this route returns a list of objects, you should change your wroding to say "Return an array of objections"

website/server/controllers/api-v3/members.js
TheHollidayInn: Why don't we just called the route `objections`? You are getting objections right?

website/server/models/user/methods.js
```diff
@@ -4,8 +4,7 @@ import {
   chatDefaults,
   TAVERN_ID,
 } from '../group';
-import { defaults } from 'lodash';
-
+import { defaults, map, flatten, flow, compact, unique, partialRight } from 'lodash';
```
paglias, line 7: Heads up, we're migrating to lodash v4 (https://github.com/HabitRPG/habitica/pull/8495) which includes many breaking changes (https://github.com/lodash/lodash/wiki/Changelog#v400). Before merging this PR should be checked and updated to make sure all lodash methods are compatible with v4.

paglias (changes requested):

SabreCat:
Continued in #8755!

PR #7999, opened 2016-09-12
Export user's private messages as HTML from inbox, #7939

### LINKED ISSUE #7939: Private Messages (PMs) should be exportable from Inbox
The Social > Inbox screen needs an option for exporting all PMs in an easy-to-read format.

I suggest that the exported file be an HTML page so that the formatting from markdown can be retained (exporting as markdown itself isn't so good because many players won't have an easy way to display it). However other suggestions are welcome.

There should also be a note to explain that if the player wants to export their messages in a code-friendly format (JSON or XML) they should use the Data > Export Data > User Data options.

That Data > Export Data screen should also have a brief message saying that private messages can be exported in a readable format from the Inbox.

akortesmaa:
Fixes #7939
### Changes

**Added translation string to groups.json:**
- "exportInbox": "Export Messages",
- "exportInboxMessages": "Export your messages as HTML? To export your messages in a code-friendly format, use Data > Export Data > User Data"

**Appended to translation string in settings.json:**
- "saveData": "Here are a few options for saving your data. You can export your private messages from your inbox."

screen shot 2016-09-12 at 14 50 43
screen shot 2016-09-12 at 14 50 27

**Added 'Export Messages' button to user's inbox next to 'Delete All Messages'.** 

screen shot 2016-09-12 at 14 49 51
screen shot 2016-09-12 at 14 50 03

The button has a confirmation dialog explaining it's use and also noting that user data can be exported from Data > Export Data in code-friendly format. 
screen shot 2016-09-12 at 14 50 16

The downloadable file is generated in groupsCtrl. File is generated as Blob and named with env.t('inbox')

The implementation is quite simple, meaning that it's just pure HTML so formatting from markdown is not retained (for now, at least).

Also, the way that time is formatted might need some changes as I wasn't sure how to do it properly. It seems to work for my timezone, but it might look wrong for other locales. Moment function used to format time (timezoneOffset is passed as a parameter from view): 
moment.utc(arr[i].timestamp).zone(timezoneOffset).format('MM/DD/YYYY HH:mm:ss') 

---

UUID: 5abc818d-baeb-4951-a66f-d6c1dba21305


crookedneighbor:
Can you move that method to the (newly created) inbox controller?

akortesmaa:
Sure, I'll look into it.

website/client/js/controllers/groupsCtrl.js
```diff
@@ -71,6 +71,35 @@ habitrpg.controller("GroupsCtrl", ['$scope', '$rootScope', 'Shared', 'Groups', '
...
+        var messages = '<html><head></head><body>';
+          for (var i = 0; i < arr.length; i++) {
+            messages += '<p><strong>' + arr[i].user + '</strong> ' + 
+              moment.utc(arr[i].timestamp).zone(timezoneOffset).format('MM/DD/YYYY HH:mm:ss') + 
+              ' (' + (i + 1) + '/' + arr.length + ')<br />' + arr[i].text + '</p>';     
+          }
+          
+        messages += '</body></html>';
+        
+        var blob = new Blob([messages], {type: "text/html"});
```
crookedneighbor: We can't really use this way of constructing it as `Blob` isn't supported in IE9 and doesn't really have any support in mobile browsers. https://developer.mozilla.org/en-US/docs/Web/API/Blob

We may drop support for IE9 soon, but as of right now, we support it.

Also, doing this entirely on the client means that no other integrations can export the inbox messages without creating their own logic. If the API handles creating it, we don't have to worry about either issue I've brought up.

paglias: Not sure we support IE9

Il 13 set 2016 16:02, "Blade Barringer" notifications@github.com ha
scritto:

crookedneighbor: At this moment, we support the browsers that angular, angular ui and lodash@3 support, which is IE9+. Semantic UI supports IE10+ and lodash@4 supports IE11+.

Althenis:
Backwards compatibility or doing this on the backend didn't really even cross my mind, but it makes perfect sense. I'll be moving this to server side soon. It should be implemented in "website/server/controllers/top-level/dataexport.js", right?

website/client/js/controllers/groupsCtrl.js
akortesmaa: Backwards compatibility or doing this on the backend didn't really even cross my mind, but it makes perfect sense. I'll be moving this to server side soon. It should be implemented in "website/server/controllers/top-level/dataexport.js", right?

crookedneighbor: 👍

crookedneighbor: Actually, looks like lodash@4 dropped support for ie 6-8, but only tests against ie11.

Alys:
I'm not sure that a confirmation popup is necessary since it's a safe operation (selecting it by accident just means you have to delete a downloaded file). That style of popup also causes trouble with users ticking the checkbox and then not being able to get confirmations for any other actions.

How about clicking on the button starts the export immediately, but there's an on-hover effect that explains what the button does and gives the information about the other export options? Something like how it's done for pet food:

image

The title could be "Export your messages as HTML" (no question mark) and then the message could be something like "HTML allows easy reading of messages in a browser. For a code-friendly format, use Data > ...."  BTW, if you can think of a better term than "code-friendly" please use it! 

How does this handle markdown, especially markdown links? I see that you said formatting is not retained, but does that mean that you see the raw markdown code (including URLs)? Or is formatting dropped completely (meaning that you wouldn't see URLs that were inserted in a `[]()` link)? I'd be concerned about going live with this if it removed links or displayed formatting in a confusing or unattractive way.

akortesmaa:
Hover certainly seems like a better option. The confirmation dialog just was the first thing that came to my mind for showing the message and allowing the download.

I guess "Machine-readable" instead of "code-friendly" is the correct term in this case.

As suggested by @crookedneighbor, the file creation is now handled by the API. The message text is taken directly from the user JSON, so markdown code is displayed raw (`[Habitica is great!](https://habitica.com)`). Is there any examples on how to use the markdown library on the server side? I can probably get this finalized during the weekend.

khipkin:
@Althenis I think something went wrong with your commits in this PR... GitHub says you changed 2,966 files?

crookedneighbor:
Looks like you merged and then tried to rebase, but you rebased against your remote repo instead of HabitRPG's repo.

I would do this on the command line:

``` bash
# set up your upstream remote, can skip if you already have it set up
git remote add upstream git@github.com:HabitRPG/habitrpg.git

# reset your changes
git fetch upstream develop
git reset upstream/develop # resets state to what upstream's version of develop is, but leaves your changes as uncomitted
git add -Av # adds all your changes
git commit -m "your commit message here"
git push --force # overwrites your branch with these commits, removing the extra commits you added from your bad rebase
```

akortesmaa:
@khipkin Yes, I managed to commit the bower_components that was in the directory for the new client. Should be (mostly) fixed now.

khipkin:
@Althenis You still have some unrelated changes in there, but 9 files instead of 2,966 is a huge improvement! 😛

website/server/controllers/top-level/dataexport.js
```diff
@@ -243,4 +243,49 @@ api.exportUserAvatarPng = {
...
+    });
+
+    let messages = '<!DOCTYPE html><html><head></head><body>';
+
+    let recipient;
+
+    for (let i = 0; i < inbox.length; i++) {
+      recipient = inbox[i].sent ? '=>' : '<=';
+      messages += `<p>${recipient}<strong>${inbox[i].user}</strong> ${moment.utc(inbox[i].timestamp).zone(timezoneOffset).format(`${dateFormat } HH:mm:ss`)} 
+        (${i + 1}/${inbox.length})<br />${inbox[i].text}</p>`;
```
crookedneighbor: @Althenis When this PR (https://github.com/HabitRPG/habitrpg/pull/8030) is merged in, you'll be able to do this to render the markdown as html:

``` js
import md from 'habitica-markdown';

let text = md.render(inbox[i].text);
```

website/server/controllers/top-level/dataexport.js
```diff
@@ -243,4 +243,49 @@ api.exportUserAvatarPng = {
...
+    // Sort array in descending order
+    inbox = _.sortBy(inbox, function sortBy (num) {
+      return num.sort * -1;
+    });
+
+    let messages = '<!DOCTYPE html><html><head></head><body>';
+
+    let recipient;
+
+    for (let i = 0; i < inbox.length; i++) {
```
crookedneighbor: Since inbox is an array, you can just use `forEach`

``` js
inbox.forEach((message) => {
  // logic
});
```

website/server/controllers/top-level/dataexport.js
```diff
@@ -243,4 +243,49 @@ api.exportUserAvatarPng = {
...
+      recipient = inbox[i].sent ? '=>' : '<=';
+      messages += `<p>${recipient}<strong>${inbox[i].user}</strong> ${moment.utc(inbox[i].timestamp).zone(timezoneOffset).format(`${dateFormat } HH:mm:ss`)}
```
crookedneighbor: This is kind of messy looking. Can we pull some of the logic out into variables and do some line breaks?

``` js
let recipientIcon = inbox[i].sent ? '=>' : '<=';
let { user, text } = inbox[i];
let timestamp = moment.utc(inbox[i].timestamp).zone(timezoneOffset).format(`${dateFormat } HH:mm:ss`);
let index = `(${i + 1}/${inbox.length});
messages = `
<p>
  ${recipientIcon}<strong>${user}</strong> ${timestamp}
  ${index}
  <br />
  ${text}
</p>`;
```

website/server/controllers/top-level/dataexport.js
```diff
@@ -243,4 +243,49 @@ api.exportUserAvatarPng = {
   },
 };
 
+/**
+ * @api {get} /export/inbox.html Export user private messages as HTML document
```
crookedneighbor, line 248: We're going to need a test for this route.

website/client-old/js/controllers/inboxCtrl.js
```diff
@@ -27,5 +27,11 @@ habitrpg.controller('InboxCtrl', ['$scope', '$rootScope',
         scope: newScope
       });
     };
+
+    $scope.exportAllMessages = function() {
+      if(confirm(env.t('exportInboxMessages'))) {
```
crookedneighbor: We don't need a confirm dialogue for this.

website/views/options/social/index.jade
```diff
@@ -17,6 +17,8 @@ script(type='text/ng-template', id='partials/options.social.inbox.html')
         .form-inline
           a.btn.btn-xs.btn-danger(ng-click='deleteAllMessages()')=env.t('clearAll')
           | &nbsp; &nbsp; &nbsp;
+          a.btn.btn-xs.btn-default(ng-click='exportAllMessages()')=env.t('exportInbox')
```
crookedneighbor: This can just be a link since we don't need a confirm dialogue.

crookedneighbor (changes requested):

crookedneighbor:
Looks like you accidentally deleted more files than you meant to when you deleted the bower components. You'll probably want to run those git commands again.

crookedneighbor:
That PR I mentioned is now merged.

akortesmaa:
@crookedneighbor I've ran the above git commands again, but the changes don't look like what they are supposed to. Also, I can try to write the test for the API endpoint. I've written some similar tests before, but as of now, I have no idea on how to validate the html document. Any tips for this?

Aside from those, the export functionality should be working as intended.

### Changes
- The 'Export Messages' button now has a popover instead of a confirm dialog.
  screen shot 2016-09-18 at 23 20 16
- The resulting inbox.html now renders markdown. The overall look is quite crude though.
  screen shot 2016-09-18 at 23 43 01

website/server/controllers/top-level/dataexport.js
akortesmaa: Done.

website/server/controllers/top-level/dataexport.js
akortesmaa: Done.

website/server/controllers/top-level/dataexport.js
akortesmaa: Done.

website/client-old/js/controllers/inboxCtrl.js
akortesmaa: Removed, now has a popover instead.

website/views/options/social/index.jade
akortesmaa: Done.

crookedneighbor:
I'll sort out the git woes when I merge it. 

As far as asserting on the HTML that's returned, as far as the client is concerned, it's just text. So just some assertions that the response includes some strings about how the messages are sent down is good enough for me.

crookedneighbor:
@Alys @lemoness Are you happy with how this looks?

screen shot 2016-09-19 at 8 40 14 pm

crookedneighbor:
Where the bold `asdfasdf` would be an actual username. And the messages would be real markdown messages.

crookedneighbor:

PR #8006, opened 2016-09-13
Players able to flag their own messages - Fixes #8005

### LINKED ISSUE #8005: Players should be able to flag their own messages
Currently, it is not possible to flag a message that you posted.

It should be possible. (E.g., when hiding posts you made in response to a user who has revealed themself as under-aged.)

This is probably an easy matter of just removing an `if` clause from the code so I'm marking this as entry level coding.

cincospenguinos:
Fixes Issue #8005 - Players should be able to flag their own messages
### Changes

Players can now flag their own messages. Previously if a player sent a message in chat, they could not flag that message. Now all players can flag all messages that they personally have sent.

This is the previous front end behavior:
old_flag

This is the new front end behavior:
flag_change

This is the front end behavior when you flag your own post:
flag_myself

---

UUID: 59197564-6563-4a98-9a35-7391033dd87a

### DIFF
--- a/website/server/controllers/api-v3/chat.js
+++ b/website/server/controllers/api-v3/chat.js
@@ -223,7 +223,7 @@ api.flagChat = {
     if (!message) throw new NotFound(res.t('messageGroupChatNotFound'));
 
     // Your own message should be flaggable
-    //if (message.uuid === user._id) throw new NotFound(res.t('messageGroupChatFlagOwnMessage'));
+    // if (message.uuid === user._id) throw new NotFound(res.t('messageGroupChatFlagOwnMessage'));
 
     let update = {$set: {}};
 
--- a/website/server/controllers/api-v3/chat.js
+++ b/website/server/controllers/api-v3/chat.js
@@ -222,7 +222,8 @@ api.flagChat = {
 
     if (!message) throw new NotFound(res.t('messageGroupChatNotFound'));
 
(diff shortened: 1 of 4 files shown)


website/server/controllers/api-v3/chat.js
```diff
@@ -222,7 +222,8 @@ api.flagChat = {
 
     if (!message) throw new NotFound(res.t('messageGroupChatNotFound'));
 
-    if (message.uuid === user._id) throw new NotFound(res.t('messageGroupChatFlagOwnMessage'));
+    // Your own message should be flaggable
+    //if (message.uuid === user._id) throw new NotFound(res.t('messageGroupChatFlagOwnMessage'));
```
crookedneighbor: Please just remove this. Also, I bet the build will fail because there is a test that covers not being able to flag your own message.

The test should be updated to illustrate that you _can_ flag your own messages.

Also, the `messageGroupChatFlagOwnMessage` translation string should be removed.

``` bash
git grep "messageGroupChatFlagOwnMessage" common/locales/en
```

^ to find the file it is in

cincospenguinos:
I removed what you gave me and modified the test case, although the test case just checks if the response is ok, not the contents of the response. Should I have it test the object returned instead?

crookedneighbor:
It's fine as is. It'll mostly serve as a reminder that being able to flag your own message is intentional.

cincospenguinos:
Cool! I'm glad I could help out.

Issue #8008, opened 2016-09-13
Deleted tags are not removed from tasks


JSchneidler:
General Info
- UUID: b1f8093a-76b7-4f80-95ce-5c00bec74c0a
- Browser: Firefox 48.0
- OS: Ubuntu 16.04
### Description

When a tag is assigned to a task and that tag is then deleted, the task still contains a reference to the deleted tag. This causes tasks to falsely display the tagged icon even if they have no existing tags. Once deleted, the tag cannot be removed from the task either.
#### Steps to reproduce:
1. Create task "test task"
2. Create tag "test tag" and save changes
3. Assign tag "test tag" to task "test task" and save changes
4. Delete tag "test tag", save changes.
#### Expected results:

Icon notifying that "test task" is tagged should disappear, tag id should be removed from corresponding entry in db.
#### Actual results:

Icon notifying that "test task" is tagged remains. In database, "test task" contains reference to the id of the deleted tag.

Alys:
Can you please click the sync button on the top right of the website (the double curved arrows icon) and tell us if that causes the tag icon to be removed from the task?

It does when I tested this and also the database did not contain reference to the ID of the deleted tag (it was removed even before I did the sync).

If you can confirm that syncing fixes this for you, then all we need to do is force a sync after tag editing is finished, which we can add to https://github.com/HabitRPG/habitrpg/issues/7820

JSchneidler:
Yes, I should have mentioned syncing does fix it. That sounds good, I'll include the fix in a PR for #7820.

Alys:
I'm closing this since we have https://github.com/HabitRPG/habitrpg/issues/7820

PR #8009, opened 2016-09-14
Run sync when tags are done being edited or a tag is added #7820

### LINKED ISSUE #7820: Unsaved tags appear in list of tags to apply and can't be moved without reload
General Info
- UUID: d814e580-fa5c-49e3-a40e-edc225329740
- Browser: Chrome 51
- OS: Windows 7
### Description

When I'm editing tags and create a new one and haven't saved it yet, it already appears in the list of tags to apply to a To-Do. When I select this option,  I get this error:

"tagId" must be a valid UUID corresponding to a tag belonging to the user.

As I understand it, the problem here is that the tag should not be appearing as an option until it has been saved.
#### Console Errors

https://habitica.com/api/v3/tasks/8ec9bc5d-a0ea-41d0-9c63-bd2c69efa0e9/tags/4bb3dc20-e43c-4c3d-9463-8e036d78915a?userV=1583 Failed to load resource: the server responded with a status of 400 (Bad Request)

---

**edit by Alys:**

The fix for this should also fix https://github.com/HabitRPG/habitrpg/issues/7806 or if that's not possible, https://github.com/HabitRPG/habitrpg/issues/7806 should be reopened.

JSchneidler:
Fixes #7820
### Changes

User is synced when a tag is added or user clicks save.

---


website/client/js/controllers/filtersCtrl.js
```diff
@@ -17,6 +17,7 @@ habitrpg.controller("FiltersCtrl", ['$scope', '$rootScope', 'User', 'Shared',
             User.updateTag({params:{id:tag.id}, body:{name:tag.name}});
         })
         $scope._editing = false;
+        User.sync();
```
crookedneighbor: Syncing is a pretty weighty operation. You should be able to pull the necessary tag info from the response of both the update and the create calls without the need to sync.

crookedneighbor: In fact, doing it this way will cause a race condition to occur. If the user sync completes before the updateTag operation completes, it'll remove the update from the ui until _another_ sync.

JSchneidler: Okay, I'll go ahead and try that.

website/client/js/services/userServices.js
```diff
@@ -319,7 +319,9 @@ angular.module('habitrpg')
 
         addTag: function(data) {
           user.ops.addTag(data);
-          Tags.createTag(data.body);
+          Tags.createTag(data.body).then(function(res) {
+            sync();
```
crookedneighbor: All you'd need to do here to return the newly created tag is:

``` js
return Tags.createTag(data.body).then(function(res) {
  return res.data.data;
});
```

That will make the `User.addTag` function a promise and you could use that to apply the id to the corresponding tag on the webpage.

I _think_ something like this would work.

``` js
User.addTag(data).then(function (tag) {
  data.id = tag.id;
});
```

crookedneighbor (changes requested):

crookedneighbor:
You will also need to merge with the upstream branch because we just moved the client folder to client-old, but it should merge without a problem.

website/client/js/controllers/filtersCtrl.js
JSchneidler: When tags are deleted though, it takes a sync for the tagged icon to disappear from tasks that no longer have a tag, described in #8008. Is there a less expensive way to update these?

crookedneighbor: I might have to look at it more closely, but you should just be able to remove it from the array.

website/client/js/services/userServices.js
JSchneidler: After looking into #7820 a bit more, I noticed the tag is being added just fine to the users tags, but the id seems to be incorrect until it syncs. Would it be better to wait for the response with the correct id before adding it to the users tags since it isn't usable before then anyways?

website/client/js/controllers/filtersCtrl.js
JSchneidler: Doesn't the deleteTag op in `common/script/ops/deleteTag.js` do that?

website/client-old/js/services/userServices.js
```diff
@@ -318,9 +318,8 @@ angular.module('habitrpg')
         },
 
         addTag: function(data) {
-          user.ops.addTag(data);
           Tags.createTag(data.body).then(function(res) {
-            sync();
+            user.ops.addTag({body: res.data.data});
```
JSchneidler, line 322: Would this be better for adding tags? This way they are not available to use until the server responds with the correct id.

JSchneidler:
On a side note, did I screw up the rebase/merge? There seem to be a lot of files changed that were part of the merge. I'm a bit new to this so let me know if I should just redo the PR. Not much was changed. Also, my solution seems to have broken a test :(

crookedneighbor:
Yeah, looks like you merged and then tried to rebase, but you rebased against your origin instead of HabitRPGs repo.

I would do this:

``` bash
# set up your upstream remote, can skip if you already have it set up
git remote add upstream git@github.com:HabitRPG/habitrpg.git

# reset your changes
git fetch upstream develop
git reset upstream/develop # resets state to what upstream's version of develop is, but leaves your changes as uncomitted
git add -Av # adds all your changes
git commit -m "your commit message here"
git push --force # overwrites your branch with these commits, removing the extra commits you added from your bad rebase
```

website/client-old/js/services/userServices.js
crookedneighbor: Probably. Sort out your merge issue and then I'll take another look.

crookedneighbor:
@JSchneidler What's the progress on this?

JSchneidler:
Sorry, was out of town. My working directory looks good after fixing it but this PR seems to be messed up. Why is it showing the 11,000+ files added as part of my PR? Haven't those changes already been done as part of the rebase?

crookedneighbor:
If you can't sort out the rebase, you're probably better off closing this PR, copying over the changes you made to a fresh branch, and opening new PR.

Alys:
@JSchneidler I'm closing this now, however you're welcome to open a new PR. If you'd like any advice about how to do @crookedneighbor's suggestion ("If you can't sort out the rebase, you're probably better off closing this PR, copying over the changes you made to a fresh branch, and opening new PR"), please ask here and we'll help!

PR #8015, opened 2016-09-15
Fix Magic Mount Spacing

Ozzystrasza:
Fixes https://github.com/HabitRPG/habitrpg/issues/7888
### Changes

I did it so the magic mount section will only appear if the user has 1 or more magic pets. Also it will show only the pawprint of the mounts that the user has an equivalent magic pet.

---

### DIFF
--- a/website/client-old/js/controllers/inventoryCtrl.js
+++ b/website/client-old/js/controllers/inventoryCtrl.js
@@ -382,6 +382,24 @@ habitrpg.controller("InventoryCtrl",
       });
     };
 
+    $scope.shouldShowPremiumMountRow = function (potion) {
+      potion = Content.premiumHatchingPotions[potion].key;
+      var pets = Object.keys(user.items.pets);
+      return pets.find(function (pet) {
+        return pet.indexOf(potion) !== -1;
+      });
+    };
+
+    $scope.shouldShowPremiumMountSection = function () {
+      var potions = Object.keys(Content.premiumHatchingPotions);
+      var pets = Object.keys(user.items.pets);
+      return pets.find(function (pet) {
+        return potions.find(function (potion) { 
+          return pet.indexOf(potion) !== -1;
+        });
+      }).length >= 1;
+    };
+
     function _updateDropAnimalCount(items) {
       $scope.petCount = Shared.count.beastMasterProgress(items.pets);
       $scope.mountCount = Shared.count.mountMasterProgress(items.mounts);
--- a/website/client-old/js/controllers/inventoryCtrl.js
+++ b/website/client-old/js/controllers/inventoryCtrl.js
@@ -355,7 +355,7 @@ habitrpg.controller("InventoryCtrl",
     $scope.seasonalShopCategories = Shared.shops.getSeasonalShopCategories(user);
 
     $scope.shouldShowPremiumPetRow = function (potion) {
-      premiumPotion = Content.premiumHatchingPotions[potion];
+      var premiumPotion = Content.premiumHatchingPotions[potion];
 
       if (!premiumPotion) {
         return false;
--- a/website/client-old/js/controllers/inventoryCtrl.js
+++ b/website/client-old/js/controllers/inventoryCtrl.js
@@ -367,25 +367,13 @@ habitrpg.controller("InventoryCtrl",
         return true;
       }
 
-      var pets = Object.keys(user.items.pets);
-      return pets.find(function (petKey) {
-        var pet = Content.petInfo[petKey];
-        return pet.potion === potion;
-      });
+      return $scope.hasAPetOfPotion(potion);
     };
 
     $scope.shouldShowPremiumPetSection = function () {
       var potions = Content.premiumHatchingPotions;
       return Object.keys(potions).find(function (potion) {
         return $scope.shouldShowPremiumPetRow(potions[potion].key);
-      });
-    };
-
-    $scope.shouldShowPremiumMountRow = function (potion) {
-      var pets = Object.keys(user.items.pets);
-      return pets.find(function (petKey) {
-        var pet = Content.petInfo[petKey];
-        return pet.potion === potion;
       });
     };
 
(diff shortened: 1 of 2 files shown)


website/client-old/js/controllers/inventoryCtrl.js
```diff
@@ -382,6 +382,24 @@ habitrpg.controller("InventoryCtrl",
       });
     };
 
+    $scope.shouldShowPremiumMountRow = function (potion) {
+      potion = Content.premiumHatchingPotions[potion].key;
```
crookedneighbor: I added some data to the Content api that should make this much simpler:

``` js
$scope.shouldShowPremiumMountRow = function (potion) {
  var pets = Object.keys(user.items.pets);
  return pets.find(function (petKey) {
    var pet = Content.petInfo[petKey];
    return pet.potion === potion;
  });
};
```

crookedneighbor: A similar simplification can be done to the pet equivalent.

website/client-old/js/controllers/inventoryCtrl.js
```diff
@@ -382,6 +382,24 @@ habitrpg.controller("InventoryCtrl",
...
+      var pets = Object.keys(user.items.pets);
+      return pets.find(function (pet) {
+        return pet.indexOf(potion) !== -1;
+      });
+    };
+
+    $scope.shouldShowPremiumMountSection = function () {
```
crookedneighbor: The logic here should basically be if the user has ever had a premium pet, right? 

You can use a similar method to what I showed you above, asserting on the `pet.type === 'premium'` property. Let me know if you need help.

crookedneighbor (changes requested):
It's all function as is, but we can clean up the logic a bit.

website/client-old/js/controllers/inventoryCtrl.js
Ozzystrasza: Made the changes, including on the `shouldShowPremiumPetRow` function. Hope I did the way you wanted!
On a side note, what did you mean by "It's all function as is"?

website/client-old/js/controllers/inventoryCtrl.js
```diff
@@ -383,21 +382,19 @@ habitrpg.controller("InventoryCtrl",
     };
 
     $scope.shouldShowPremiumMountRow = function (potion) {
-      potion = Content.premiumHatchingPotions[potion].key;
       var pets = Object.keys(user.items.pets);
-      return pets.find(function (pet) {
-        return pet.indexOf(potion) !== -1;
+      return pets.find(function (petKey) {
```
crookedneighbor, line 382: Looks like we do this exact same check twice. Pull it out into a function?

crookedneighbor (changes requested):

crookedneighbor:
It was a typo. All `functional` as is.

website/client-old/js/controllers/inventoryCtrl.js
Ozzystrasza: Does it need to be `$scope.functionName`, or should it be `var functionName` ? Sorry for the noob question, but I don't quite get the difference yet.

Ozzystrasza: also, all the `shouldShowPremiumMountRow` does now is call this new function, should I remove it and use the new function directly in the mounts.jade? 
And if that's the case it needs to be `$scope.newFunction` right?

Ozzystrasza: Did some changes, let me know if they are ok.

khipkin: The difference is that the functions in the `$scope` can be used by anything else with access to that `$scope` (i.e., any views that use this controller), whereas the `var functionName` functions are helper functions that are not visible outside this file.

crookedneighbor: It's worth noting that `$scope` is an Angular thing. Not a general JavaScript thing.

website/client-old/js/controllers/inventoryCtrl.js
```diff
@@ -355,24 +355,19 @@ habitrpg.controller("InventoryCtrl",
     $scope.seasonalShopCategories = Shared.shops.getSeasonalShopCategories(user);
 
     $scope.shouldShowPremiumPetRow = function (potion) {
-      potion = Content.premiumHatchingPotions[potion];
+      premiumPotion = Content.premiumHatchingPotions[potion];
```
crookedneighbor: var premiumPotion

crookedneighbor (changes requested):
Other than my one comment, it looks good to me.

website/client-old/js/controllers/inventoryCtrl.js
Ozzystrasza: Damn, how did I miss that!
Just pushed the fix. Thanks for the help and teachings @crookedneighbor and @khipkin .

crookedneighbor: Probably because I just reassigned the existing variable before :)

crookedneighbor (approved):

crookedneighbor:
I made one minor edit that I noticed while making one last review: https://github.com/HabitRPG/habitrpg/commit/9c0998c29ab44e5370871a16732c0eb1ef5dc87f

Extracts the find pet logic into it's own function that takes a function for how to compare the pet, rather than repeating the logic.

Notated this PR to your next tier.

Ozzystrasza:
No code repetitions on your watch! 😄 Thanks!

mandyzhou:
Looks like this was released in v3.41.3, but I'm still seeing the large gap on my Mounts page. The Pets page looks fine, though.

SabreCat:
@mandyzhou: The release hasn't gone live yet thanks to a hiccup I've attempted to address [here](https://github.com/HabitRPG/habitrpg/pull/8029/files).

PR #8024, opened 2016-09-15
Handle social auth in A/B testing

SabreCat:
### Changes

Previously,
1. It was possible for a user who registered prior to the implementation of A/B tests to get assigned to an A/B test, which could in turn modify their preferences undesirably. This happened if they signed up using a Facebook account, then later added an email/username based sign-in method.
2. Users registering via Facebook authentication were not being added to A/B tests.

This PR corrects both of the above, hopefully in a futureproof fashion--as we move from one A/B test to the next, all we need to do is tweak the block in the pre save hook to make the needed user updates.

---

UUID: [Authentic Test Correction](https://map.what3words.com/authentic.test.correction)

### DIFF
--- a/test/api/v3/integration/user/auth/POST-user_auth_social.test.js
+++ b/test/api/v3/integration/user/auth/POST-user_auth_social.test.js
@@ -2,6 +2,7 @@ import {
   generateUser,
   requester,
   translate as t,
+  getProperty,
 } from '../../../../../helpers/api-integration/v3';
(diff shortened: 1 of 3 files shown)


crookedneighbor:
Mentioned this in slack, but this should be move to the `pre('save'` hook for new users.

paglias (approved):
LGTM 👍

crookedneighbor (approved):

PR #8025, opened 2016-09-16
Common reorg

crookedneighbor:
- Moves common -> website/common
- Extracts common/img/sprites -> website/assets/sprites
- Extracts common/img -> website/assets/img
- Extracts common/audio -> website/assets/audio
- Browserifies common scripts into website/client-old/js
- Concats sprite css into website/client-old/css


crookedneighbor:
This should be squashed when it's ready to be merged.

paglias (commented):
@crookedneighbor only thing that concerns me is that if an external website or app referenced static files in `/common/audio|img|...` they'll break since the path changed.

Also, but it's less important, with the Vue PR we're going to have `website/client/assets` -> assets going through Webpack, `website/static` -> assets copied to `/dist` (webpack output folder) without being modified and `website/assets` -> the ones from this PR. I can merge `website/static` with `website/assets` but I'm slightly worried about the confusion of having 2 assets folder. 

But anyway I think it's easier to change these things into my PR than here. So if we're not worried about paths changing you can merge for me.

crookedneighbor:
I am not worried about the file paths changing. Any asset that we intend to expose to third parties (sprites, audio, etc) should be put on S3. If anyone is relying on a particular website path, they should not be doing so.

PR #8031, opened 2016-09-17
Fix "All" button to select all guilds in the filter list

Ozzystrasza:
Fixes https://github.com/HabitRPG/habitrpg/issues/7763
### Changes

Changed the `selectAll` and `selectNone` functions to use the `groupsFilter` array, which is the same used to fill the list of guilds to filter. Also for `filterInitialChallenges` function so the same thing happens by default when you open the challenges page.

Let me know if I got the tests right.

Before / After

---


crookedneighbor:
:+1: Good work.

crookedneighbor:

Ozzystrasza:
Thanks!

PR #8040, opened 2016-09-19
Updating a task now sends only data that's actually needed (Fixes #7846)

### LINKED ISSUE #7846: Habits with large history cause "Request entity too large" errors
A Habit that's used very frequently develops a large history and can cause "Request entity too large" errors when the player tries to keep using the task.

We found one such task today with about 700 entries for this month and 1600 for the previous month (before that preening had been happening). There were about 2440 history entries in all.

The error stopped happening when I set the history to an empty array (with the player's permission).

We could do special preening for huge history entries, although that's problematic since we promise that history will be kept for a certain time.

Might not be a priority since it might not happen often. This is the first time I've seen it.

rbreu:
Fixes #7846
### Changes

Updating a task with a huge history lead to 'request entity too large' errors since the client attempted to send all data associated with a task, regardless of whether that data could be changed or not. With this change, the post request in updateTask is now omitting the following fields:

challenge
completed
createdAt
group
history
id
reminders
tags
type
updatedAt
userId

---

UUID: 7b45f003-63bf-4bbe-9015-88d74753595e

### DIFF
--- a/website/client-old/js/services/taskServices.js
+++ b/website/client-old/js/services/taskServices.js
@@ -72,27 +72,16 @@ angular.module('habitrpg')
     };
 
     function updateTask (taskId, taskDetails) {
+      var taskDetailsToSend = _.omit(
+        taskDetails,
+        ['challenge', 'completed', 'createdAt', 'group', 'history', 'id',
+         'reminders', 'tags', 'type', 'updatedAt', 'userId']
+      )
+
       return $http({
         method: 'PUT',
         url: '/api/v3/tasks/' + taskId,
-        data: {
-          alias: taskDetails.alias,
-          attribute: taskDetails.attribute,
-          checklist: taskDetails.checklist,
-          collapseChecklist: taskDetails.collapseChecklist,
-          date: taskDetails.date,
-          down: taskDetails.down,
-          everyX: taskDetails.everyX,
-          frequency: taskDetails.frequency,
-          notes: taskDetails.notes,
-          priority: taskDetails.priority,
-          repeat: taskDetails.repeat,
-          startDate: taskDetails.startDate,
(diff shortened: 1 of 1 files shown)


paglias:
@rbreu it would be better to blacklist items we don't want to send instead of specifying the ones to send. So if we add a new field we don't have to remember to include it. You can use Lodash's omit function to achieve this https://lodash.com/docs/4.15.0#omit

crookedneighbor:
Was thinking the same thing

rbreu:
Done!

website/client-old/js/services/taskServices.js
```diff
@@ -72,10 +72,16 @@ angular.module('habitrpg')
     };
 
     function updateTask (taskId, taskDetails) {
+      var taskDetailsToSend = _.omit(
+        taskDetails,
+        ['challenge', 'completed', 'createdAt', 'group', 'history', 'id',
```
crookedneighbor, line 77: Question for the room: Should we just prevent updating complex fields (`challenge`, `group`, `history`, `reminders`, `tags`) and just let the other negligible properties be ignored? The problem is that we're sending too much data along with the request, so if we cut out the non-string stuff, it would make the code simpler.

``` js
var taskDetailsToSend = _.omit(taskDetails, ['challenge', 'group', 'history', 'reminders', 'tags']);
```

crookedneighbor: Though perhaps it wouldn't be obvious why we're keeping just those properties off of the update call.

crookedneighbor: I'm just going to make that adjustment and merge this in. We can change it later if necessary.

crookedneighbor:

PR #8044, opened 2016-09-20
Send user to analytics on register

SabreCat:
### Changes

Previously, we only sent the full user object to the analytics service when tracking the `cron` event. While that makes sense from an efficiency perspective (we don't want to ship the whole user object around more often than necessary), it meant that we didn't capture some important user data until their second day of activity. Now, we send user data to Analytics when the user completes registration, which allows us to collect e.g. A/B test assignment information that was often missing before.

Also fixes an issue that could prevent item name data from being correctly sent to Amplitude when tracking drops and similar acquisitions.

### DIFF
--- a/website/server/libs/analyticsService.js
+++ b/website/server/libs/analyticsService.js
@@ -144,7 +144,7 @@ let _formatDataForAmplitude = (data) => {
   let itemName = _lookUpItemName(data.itemKey);
 
   if (itemName) {
-    event_properties.itemName = itemName;
+    ampData.event_properties.itemName = itemName;
   }
   return ampData;
 };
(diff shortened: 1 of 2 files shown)


website/server/libs/analyticsService.js
```diff
@@ -144,7 +144,7 @@ let _formatDataForAmplitude = (data) => {
   let itemName = _lookUpItemName(data.itemKey);
 
   if (itemName) {
-    event_properties.itemName = itemName;
+    ampData.event_properties.itemName = itemName;
```
crookedneighbor, line 147: This is referencing the same object so wasn't broken (or if it was broken, this won't fix it). 👍 for clarity though.

SabreCat: Oh, does it? That must be a convention/trick I'm not familiar with, oops.

PR #8051, opened 2016-09-22
fix broken links in locales readme files

mleko:
### Changes

Fixes broken _README_FIRST link

---

UUID:


Alys:
Ah, sorry about that, I'd missed the fact that those links would be broken.

Let's future-proof this in case of further changes. Can you please make each file say just this:

```
 Do not edit any files in this directory!

 For more information read README.md in the parent `locales` directory.
```

Alys:
That's great, thank you! 

If you'd like this to count towards a contributor tier, please tell us your User ID (that's what the UUID field is for in the Pull Request template, but please add it to a new comment now so that we get emailed automatically when you've added it).

mleko:
UUID: 59543ced-98ce-4d5c-8226-aa82f896b928

Alys:
I've awarded you a tier. You can read more about [contributor tiers and rewards here](http://habitrpg.wikia.com/wiki/Contributor_Rewards).

PR #8054, opened 2016-09-23
Vuex Structure

paglias:
- Adds loading screen
- Add actions to fetch user and set page title
- Add vue-resource


paglias:
Merging so that it goes on the staging site, if we decide to go without Vuex we can always remove it. It has some good improvements anyway

PR #8064, opened 2016-09-25
Fixes #7958 - do not remove Battle Gear equipment when changing class

### LINKED ISSUE #7958: do not remove Battle Gear equipment when changing class
When a player changes class, their Battle Gear equipment should NOT be removed (nor their Costume gear of course). 

They should also NOT be directed to the Equipment page, and should NOT see the "Class Gear" popup:
image
(We might want to put an explanation somewhere  about the class bonus, but that's probably something for the upcoming website improvement project, which will introduce more help tips on the site.)

Staff/mods/admins have talked about this somewhere else and I thought it was on GitHub but I can't find it.

The reason is that unless you're a fairly new user, you are likely to be wearing Legendary equipment, or customised gear from other classes and having it change is annoying. For new-ish users, they'll likely need to buy new gear anyway, and can change their equipment as they do that.

cmajames:
#7958
### Changes

website/client-old/js/controllers/userCtrl.js#38: removed to keep equipt equipment constant
website/common/common/script/ops/changeClass.js#33:  removed to stop 'classes' introduction


website/common/script/ops/changeClass.js
```diff
@@ -35,8 +35,6 @@ module.exports = function changeClass (user, req = {}, analytics) {
         }
       }
 
-      user.items.gear.equipped[type] = foundKey;
```
crookedneighbor: Looks like all of this `foundKey` logic can be removed as it's only used for this line.

You'll need to update the corresponding test as well.

crookedneighbor (changes requested):

crookedneighbor:
What's the status on this PR?

cmajames:
@crookedneighbor Sorry that I haven't been able to follow up - I run a tight schedule.  I've gone ahead and pushed the changes you suggested.

crookedneighbor (approved):

crookedneighbor (dismissed):
@Alys was your intention to have the class tour stuff removed entirely (even when first achieving level 10)? Because that's what happens in this PR. I would think you should go through the tour when you first choose a class, but not after.

Alys:
The intention is to never remove the gear that the player is wearing when they change class, not even on the first class change (because they are almost certain to be wearing better gear than the default gear for the new class; also keeping their current gear equipped when they change class for the first time emphasises that the gear is still usable).

If the tour is kept for the first class change, it will need to be modified since it currently says that your gear has been removed. Once this change is in place, the only useful part of the tour is "Wearing your class's gear grants you a 50% bonus to stats" which is also stated (more or less) when you hover over class gear. I'm not opposed to us having a different class tour screen for the first change.

cmajames:
@crookedneighbor is there a reason why userCtrl#changeClass handles both the prompting dialogue and actually setting the new class? If not, what sort of terrible things would happen if I wanted to refactor that segment of code?

crookedneighbor:
Go ahead and refactor it.

cmajames:
The test for api-v3 fails, however no tests are shown as being invalid; rather there appears to be an issue with the testing harness.

Alys:
@cmajames I've restarted the Travis CI build for your PR, so check back in a bit to see if it was only a temporary problem.

cmajames:
@Alys, yep, it appears the bug was some weird fluke.

cmajames:
@crookedneighbor I'd like to hear your insight on the current state of things if you have the time to take a look at things.

website/server/models/user/schema.js
```diff
@@ -233,6 +233,7 @@ let schema = new Schema({
     armoireEmpty: {type: Boolean, default: false},
     cardReceived: {type: Boolean, default: false},
     warnedLowHealth: {type: Boolean, default: false},
+    shownClassIntro: {type: Boolean, defalut: false},
```
crookedneighbor: Not a fan of adding another user flag. 😞

cmajames: "Unfortunate But Necessary"? Or can you think of another solution that doesn't involve adding flags?

cmajames: User.user.flags.tour.classes?

website/client-old/js/services/guideServices.js
```diff
@@ -64,11 +64,15 @@ function($rootScope, User, $timeout, $state, Analytics) {
...
+          content: window.env.t('classGearText'),
+          final: true,
           state: 'options.inventory.equipment',
           element: '.equipment-tab',
           title: window.env.t('classGear'),
-          content: window.env.t('classGearText')
-        }, {
+          hideNavigation: true
+        }
+        /*, {
```
cmajames, line 75: @crookedneighbor Do you know if any of the following tour elements are still used in the code?

cmajames: As far as I can tell, no; a quick git grep only shows the one call to 'classes' in userCtrl. See [Gist](https://gist.github.com/cmajames/91d9469ed080589ebeba004f7c78b0e6).

website/client-old/js/controllers/userCtrl.js
```diff
@@ -21,16 +21,17 @@ habitrpg.controller("UserCtrl", ['$rootScope', '$scope', '$location', 'User', '$
...
       Shared.updateStore(User.user);
-      Guide.goto('classes', 0,true);
+      Guide.goto('classes', 0, true);
+    }
+
+    $scope.enableClasses = function(){
+      return User.changeClass({});
+    }
+
+    $scope.payForNewClass = function(){
```
TheHollidayInn: Why is this called `payForNewClass?`

website/client-old/js/services/guideServices.js
TheHollidayInn: I think not since it looks like this was already commented out.

website/server/controllers/api-v3/user.js
```diff
@@ -77,6 +77,7 @@ let updatablePaths = [
   'flags.cardReceived',
   'flags.warnedLowHealth',
   'flags.newStuff',
+  'flags.shownClassIntro',
```
TheHollidayInn: Where is this flag used?

TheHollidayInn:
Everything here looks good to me and works locally. I just had a question about the flag, but I think we can merge this in.

website/server/controllers/api-v3/user.js
cmajames: It isn't; I'll squash.

cmajames:
I put the "flake" in "special snowflake" ;)
I've purged the last remnants of the additional flag code and rebased onto the latest code as cleanly as I could ^^;

test/common/ops/changeClass.js
```diff
@@ -70,13 +70,9 @@ describe('shared.ops.changeClass', () => {
 
       expect(user.stats.class).to.equal('rogue');
       expect(user.flags.classSelected).to.be.true;
-      expect(user.items.gear.equipped.weapon).to.equal('weapon_rogue_0');
       expect(user.items.gear.owned.weapon_rogue_0).to.be.true;
-      expect(user.items.gear.equipped.armor).to.equal('armor_rogue_1');
       expect(user.items.gear.owned.armor_rogue_1).to.be.true;
-      expect(user.items.gear.equipped.shield).to.equal('shield_rogue_0');
       expect(user.items.gear.owned.shield_rogue_0).to.be.true;
-      expect(user.items.gear.equipped.head).to.equal('head_base_0');
```
Alys: You've removed the old tests which showed that the base level equipment of the new class was equipped, which is great, but can you please also add tests to show that the player's original choice of gear was not unequipped? For example, set the test account to own some mid-level Healer items and be wearing them as Battle Gear, then after the test does the class change, check that they're still equipped.

website/common/locales/en/npc.json
```diff
@@ -61,7 +61,7 @@
     "paymentMethods": "Purchase using",
 
     "classGear": "Class Gear",
-    "classGearText": "First: don't panic! Your old gear is in your inventory, and you're now wearing the apprentice equipment of your new class. Wearing your class's gear grants you a 50% bonus to stats. However, feel free to switch back to your old gear.",
+    "classGearText": "Congratulations on choosing a class! I've gone ahead and added your new basic weapon to your inventory. Take a look below to equip it!",
```
Alys: Can you remove "gone ahead and"? The sentence still makes sense without it and keeping it short is good.

Alys: A question for the other admins:

The text in `classGearText` now says almost the opposite of what it used to say. If we keep using that key, anyone using a language other than `en` will get the wrong message until the translators update it, and for some languages that could be months. What's our usual process in that situation? Should we use a new key for this new text and delete the old one?

website/common/script/ops/changeClass.js
```diff
@@ -44,26 +43,6 @@ module.exports = function changeClass (user, req = {}, analytics) {
     user.flags.classSelected = true;
 
     each(['weapon', 'armor', 'shield', 'head'], (type) => {
```
Alys: `armor` and `head` can be removed from this `each`

It might even be clearer to get rid of the `each` and just have lines like below, but I'm not certain this is better so disagreements are welcome. :)

    user.items.gear.owned[`weapon_${klass}_0`] = true;
    if (klass === 'rogue')  user.items.gear.owned[`shield_${klass}_0`] = true;

test/common/ops/changeClass.js
```diff
@@ -70,13 +70,9 @@ describe('shared.ops.changeClass', () => {
 
       expect(user.stats.class).to.equal('rogue');
       expect(user.flags.classSelected).to.be.true;
-      expect(user.items.gear.equipped.weapon).to.equal('weapon_rogue_0');
       expect(user.items.gear.owned.weapon_rogue_0).to.be.true;
-      expect(user.items.gear.equipped.armor).to.equal('armor_rogue_1');
       expect(user.items.gear.owned.armor_rogue_1).to.be.true;
```
Alys: I think this `expect` for armor_rogue_1` could be removed, along with `user.items.gear.owned.armor_rogue_1 = true;` above. I believe their only purpose was to test that the armor was equipped when it was already owned, which is no longer relevant.

Alys (dismissed):
@cmajames Sorry it's taken us a while to look at this. :(

Note to other admins: I haven't done a thorough code review.
I've done some testing and it looks good so far.

website/common/locales/en/npc.json
paglias: When a new english string is uploaded to Transifex all old translations are removed, so there should be no problems

website/common/script/ops/changeClass.js
paglias: agreed the code now only runs for `(type === 'weapon' || (type === 'shield' && klass === 'rogue')`

cmajames:
Okay, that should handle all the requested changes that I've seen :)

SabreCat:
The fix is working, but the equipment-related notification shows up with the wrong timing now:

image

I think this is related to the changes made in the client `userCtrl.js`?

SabreCat (changes requested):
Need to address timing of the Justin popover, then this will be ready methinks!

SabreCat (approved):

SabreCat:
Upon revisiting, I am no longer able to reproduce that issue. `¯\_(ツ)_/¯` So this is good to go! @cmajames, what's your Habitica UUID from https://habitica.com/#/options/settings/api ? I can get you Blacksmith credit for this!

PR #8065, opened 2016-09-25
Random val refactor

### LINKED ISSUE #7929: Change Drop Calculation To Not Rely on Seeded Stats from User

crookedneighbor:
Piggybacking on #7994 (includes a commit from it)

Adjusts the server side only random values to use `Math.random` instead of seeded random values.

Ping @paglias @gambtho 

closes #7994
fixes #7929

### DIFF
--- a/website/common/script/index.js
+++ b/website/common/script/index.js
@@ -298,7 +298,6 @@ api.wrap = function wrapUser (user, main = true) {
     handleTwoHanded: _.partial(importedFns.handleTwoHanded, user),
     predictableRandom: _.partial(importedFns.predictableRandom, user),
     crit: _.partial(importedFns.crit, user),
-    randomVal: _.partial(importedFns.randomVal, user),
     randomDrop: _.partial(importedFns.randomDrop, user),
     autoAllocate: _.partial(importedFns.autoAllocate, user),
     updateStats: _.partial(importedFns.updateStats, user),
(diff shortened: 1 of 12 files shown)


website/common/script/index.js
```diff
@@ -298,7 +298,6 @@ api.wrap = function wrapUser (user, main = true) {
     handleTwoHanded: _.partial(importedFns.handleTwoHanded, user),
     predictableRandom: _.partial(importedFns.predictableRandom, user),
     crit: _.partial(importedFns.crit, user),
-    randomVal: _.partial(importedFns.randomVal, user),
```
paglias, line 301: why removed from here?

crookedneighbor: Because the user is no longer passed in. `randomVal` should probably be moved to the libs folder, since it is no longer a user function.

paglias (approved):

PR #8074, opened 2016-09-26
Adding spam prevention - fixes #8060 - continued in PR #8687

### LINKED ISSUE #8060: Spam prevention - prevent a player from posting too many chat messages in a short time
We want the Tavern, guild, and party chats to have a timed limit preventing people from posting many messages in a row. It should allow only X posts per Y time, with a rolling window, where X and Y are specified as environment variables or constants (constants might be fine because we probably won't want to change this often).

Moderators are exempt. We might also want to exempt Socialites of a high enough level, so perhaps make an exemption for all players who have `contributor.level` greater than perhaps 3 (moderators are level 8). Specify that number as a constant or environment variable.

FYI @lemoness

colingm:
Fixes #8060 
### Changes

I added a loop inside the POST chat api function to loop over the most recent messages to see if the user has posted too many messages within the allotted time. There are a few constants added that we can change to alter the length of the window and how many messages are allowed in that window. Also this spam detection will not be run on anyone that has a contributor level higher than 3 (this can also be changed).

The loop will only look at messages that fit within the time window so as to not look at more messages than needed. A translatable error will be thrown as soon as the spam message limit has been reached to alert as quickly as possible.

Note: Please give input if the constant values look okay or if we need to adjust the length of the window or the amount of messages allowed. Also I am sure the message isn't perfect so if you have a better idea for an error message I can quickly change that as well. All input is needed since this is my first PR for Habitica.

---

UUID: 9c877aed-a20b-4aa0-b4b3-85e4456f934e


website/server/controllers/api-v3/chat.js
```diff
@@ -101,6 +111,22 @@ api.postChat = {
...
 
+    // Check for spam by seeing if the current user has posted too many messages in a short window
+    if (!(user.contributor.level >= SPAM_CONTRIBUTOR_LEVEL) && (group.type === 'party' || group.type === 'guild')) {
+      let currentTime = Number(new Date());
+      let userMessages = 0;
+      for (let i = 0; i < group.chat.length; i++) {
+        let message = group.chat[i];
+        if (message.uuid === user._id && currentTime - message.timestamp <= SPAM_WINDOW_LENGTH) {
+          if (++userMessages >= SPAM_MESSAGE_LIMIT) {
+            throw new NotFound(res.t('messageGroupChatSpam'));
```
crookedneighbor: Throwing a 404 is probably not the right error here.

website/server/controllers/api-v3/chat.js
```diff
@@ -101,6 +111,22 @@ api.postChat = {
...
     chatUpdated = lastClientMsg && group.chat && group.chat[0] && group.chat[0].id !== lastClientMsg ? true : false;
 
+    // Check for spam by seeing if the current user has posted too many messages in a short window
+    if (!(user.contributor.level >= SPAM_CONTRIBUTOR_LEVEL) && (group.type === 'party' || group.type === 'guild')) {
+      let currentTime = Number(new Date());
+      let userMessages = 0;
+      for (let i = 0; i < group.chat.length; i++) {
+        let message = group.chat[i];
+        if (message.uuid === user._id && currentTime - message.timestamp <= SPAM_WINDOW_LENGTH) {
+          if (++userMessages >= SPAM_MESSAGE_LIMIT) {
```
crookedneighbor: For clarity sake, let's move the number incrementation outside of the if block.

website/server/controllers/api-v3/chat.js
```diff
@@ -101,6 +111,22 @@ api.postChat = {
     let lastClientMsg = req.query.previousMsg;
     chatUpdated = lastClientMsg && group.chat && group.chat[0] && group.chat[0].id !== lastClientMsg ? true : false;
 
+    // Check for spam by seeing if the current user has posted too many messages in a short window
+    if (!(user.contributor.level >= SPAM_CONTRIBUTOR_LEVEL) && (group.type === 'party' || group.type === 'guild')) {
```
crookedneighbor: Why the group.type check? There are only two types, guild and party.

website/server/controllers/api-v3/chat.js
```diff
@@ -101,6 +111,22 @@ api.postChat = {
     let lastClientMsg = req.query.previousMsg;
     chatUpdated = lastClientMsg && group.chat && group.chat[0] && group.chat[0].id !== lastClientMsg ? true : false;
 
+    // Check for spam by seeing if the current user has posted too many messages in a short window
```
crookedneighbor: I don't think this comment is necessary. I think it would probably be better to pull this logic into an instance method on the group model.

test/api/v3/integration/chat/POST-chat.test.js
```diff
@@ -78,4 +79,31 @@ describe('POST /chat', () => {
...
+      } catch (error) {
+        spamError = error;
+        break;
+      }
+    }
+
+    expect(spamError).to.eql({
+      code: 404,
+      error: 'NotFound',
+      message: 'You have been sending messages too quickly, please wait briefly and try again.',
```
crookedneighbor: You should use the imported `t` function to assert on a translation string.

test/api/v3/integration/chat/POST-chat.test.js
```diff
@@ -78,4 +79,31 @@ describe('POST /chat', () => {
     expect(message.message.id).to.exist;
     expect(memberWithNotification.newMessages[`${group._id}`]).to.exist;
   });
+
+  it('Returns an error when the user has been posting too many messages', async () => {
+    let spamError = {};
+
+    // this loop should only actually run 10 times if the test would fail, else it is cut short at first error
+    for (let i = 0; i < 5; i++) {
+      try {
```
crookedneighbor: I think it would be better to not do this in a for loop, to better illustrate the threshold.

test/api/v3/integration/chat/POST-chat.test.js
```diff
@@ -20,6 +20,7 @@ describe('POST /chat', () => {
     user = groupLeader;
     groupWithChat = group;
     userWithChatRevoked = await members[0].update({'flags.chatRevoked': true});
+    userSocialite = await members[1].update({'contributor.level': 5});
```
crookedneighbor: You only need this for one test, do the update in the test where it is needed.

crookedneighbor (changes requested):
Overall, I like where you're going with it, since it doesn't require any writing to the database. 

When someone hits the spam limit on the client, is there message lost? Or does it remain?

website/server/controllers/api-v3/chat.js
colingm: yeah not found doesn't really make sense, I had just done that because it seemed to me like a similar situation as chat privileges being revoked that did use a 404 not found. Would a 403 be more appropriate for the application? Is that one that is handled in the code base well enough?

website/server/controllers/api-v3/chat.js
colingm: Very simple comment yeah, okay I can look at moving this function into the group model, that should make it more reusable too.

website/server/controllers/api-v3/chat.js
colingm: I didn't know there were only two types so that is why :) now that I know that I can remove this check

website/server/controllers/api-v3/chat.js
colingm: sounds good, I will get this in the next commit

test/api/v3/integration/chat/POST-chat.test.js
colingm: makes sense, I will move it.

test/api/v3/integration/chat/POST-chat.test.js
colingm: okay, I had only done this to make it so the test wouldn't have to be changed every time we change the threshold but an explicit amount of calls would be a better test I will change that.

test/api/v3/integration/chat/POST-chat.test.js
colingm: oh cool I forgot to do that here, I will change that.

colingm:
In my tests their message remained in the chat box as if they hadn't tried to send it yet and they got the little popup in the corner with the error message.

colingm:
Okay I will make those quick changes. I forgot about checking if there were unit tests for group after I moved it to a function. I will add those as well.

website/server/models/group.js
```diff
@@ -905,6 +915,28 @@ schema.methods.removeTask = async function groupRemoveTask (task) {
   }, {multi: true}).exec();
 };
 
+// Returns true if the user has reached the spam message limit
+schema.methods.checkChatSpam = function groupCheckChatSpam (user) {
+  let group = this;
+  if (!(user.contributor.level >= SPAM_CONTRIBUTOR_LEVEL)) {
+    let currentTime = Number(new Date());
```
paglias: Date.now() works too

website/server/models/group.js
```diff
@@ -36,6 +36,16 @@ const LARGE_GROUP_COUNT_MESSAGE_CUTOFF = shared.constants.LARGE_GROUP_COUNT_MESS
...
 
+/*
+#  Spam constants to limit people from sending too many messages too quickly
+#    MESSAGE_LIMIT - The amount of messages that can be sent in a time window
+#    WINDOW_LENGTH - The window length for spam protection in milliseconds
+#    CONTRIBUTOR_LEVEL - Anyone at or above this level is exempt
+*/
+export const SPAM_MESSAGE_LIMIT = 2;
+export const SPAM_WINDOW_LENGTH = 60000; // 1 minute
+export const SPAM_CONTRIBUTOR_LEVEL = 4;
```
crookedneighbor: I feel like this could have a better name.

website/server/models/group.js
```diff
@@ -985,6 +995,30 @@ schema.methods.removeTask = async function groupRemoveTask (task) {
   }, {multi: true}).exec();
 };
 
+// Returns true if the user has reached the spam message limit
+schema.methods.checkChatSpam = function groupCheckChatSpam (user) {
+  if (user.contributor && user.contributor.level >= SPAM_CONTRIBUTOR_LEVEL) {
+    return false;
+  }
+
+  let group = this;
```
crookedneighbor: We don't need to assign this to a variable. We never go into a context where the reference to `this` changes and the logic is not so complex that it's confusing to do `this.chat` instead.

If you wanted to do:

``` js
let chat = this.chat;
```

I would be fine with that.

crookedneighbor (changes requested):
Just talked with @lemoness and confirmed that this should only affect posts made in the Tavern. You should be able to use the `TAVERN_ID` constant for that check.

website/common/locales/en/messages.json
```diff
@@ -56,6 +56,7 @@
   "messageGroupChatFlagAlreadyReported": "You have already reported this message",
   "messageGroupChatNotFound": "Message not found!",
   "messageGroupChatAdminClearFlagCount": "Only an admin can clear the flag count!",
+  "messageGroupChatSpam": "You have been sending messages too quickly, please wait briefly and try again.",
```
khipkin: Grammar nit: comma splice

website/server/models/group.js
khipkin: Some ideas: `SPAM_MIN_EXCUSED_CONTRIB_LEVEL`, `SPAM_MIN_EXEMPT_CONTRIB_LEVEL`, `MIN_CONTRIB_LEVEL_NO_MESSAGE_LIMIT`, `MIN_CONTRIB_LEVEL_SPAM_EXEMPT`

website/server/models/group.js
khipkin: Even then, `chat` is only used twice below.

website/common/locales/en/messages.json
colingm: Thank you! Please let me know if there are any other ideas for how this message could be better

lemoness:
I'd tweak the message that shows up for using who are posting lots of
messages to make it friendlier and more informative. How about:

"Whoops, looks like you're posting too many messages! Please wait a minute
and try again. The Tavern chat only holds 200 messages at a time, so
Habitica encourages posting longer, more thoughtful messages and
consolidating replies. Can't wait to hear what you have to say :)"

Feedback very welcome!

test/api/v3/unit/models/group.test.js
```diff
@@ -607,6 +613,90 @@ describe('Group Model', () => {
       });
     });
 
+    describe('#checkChatSpam', () => {
+      let testMessage, testUser, testTime;
+      let testUserID = '1';
+      beforeEach(async () => {
+        party._id = TAVERN_ID;
```
crookedneighbor: I'd rather create a new variable than use `party` for this. It's confusing to look at.

test/api/v3/unit/models/group.test.js
```diff
@@ -607,6 +613,90 @@ describe('Group Model', () => {
...
+          uuid: testUserID,
+          timestamp: testTime,
+        };
+        testUser = {
+          _id: testUserID,
+        };
+      });
+
+      it('group that is not the tavern returns false', async () => {
+        party._id = '2';
```
crookedneighbor: For better coverage, this should include a condition that if it _was_ the Tavern, it would return true.

test/api/v3/unit/models/group.test.js
```diff
@@ -607,6 +613,90 @@ describe('Group Model', () => {
       });
     });
 
+    describe('#checkChatSpam', () => {
+      let testMessage, testUser, testTime;
+      let testUserID = '1';
+      beforeEach(async () => {
+        party._id = TAVERN_ID;
+        testTime = Date.now();
+        testMessage = {
```
crookedneighbor: Might be better to create a function that returns a test message object. Something like:

``` js
function generateTestMessage (overrides = {}) {
  return Object.assign({}, overrides, {
    text: 'test message',
    uuid: testUserID,
    timestamp: testTime,
  });
}
```

That way, you create a new object every time, not reference the same object over and over.

website/server/models/group.js
```diff
@@ -985,6 +995,29 @@ schema.methods.removeTask = async function groupRemoveTask (task) {
   }, {multi: true}).exec();
 };
 
+// Returns true if the user has reached the spam message limit
+schema.methods.checkChatSpam = function groupCheckChatSpam (user) {
+  if (this._id !== TAVERN_ID || user.contributor && user.contributor.level >= SPAM_MIN_EXEMPT_CONTRIB_LEVEL) {
```
crookedneighbor: For clarity of the logic, let's put these in separate `if` blocks:

``` js
if (this._id !== TAVERN_ID) {
  return false;
} else if (user.contributor && user.contributor.level >= SPAM_MIN_EXEMPT_CONTRIB_LEVEL) {
  return false;
}
```

crookedneighbor (changes requested):

TheHollidayInn:
Looks good to me. @paglias @crookedneighbor any additional thoughts?

test/api/v3/integration/chat/POST-chat.test.js
```diff
@@ -155,4 +161,28 @@ describe('POST /chat', () => {
     expect(message.message.id).to.exist;
     expect(memberWithNotification.newMessages[`${group._id}`]).to.exist;
   });
+
+  it('Returns an error when the user has been posting too many messages', async () => {
```
khipkin, line 183: Unless a mocked timer is used, I believe these tests may fail if they are delayed (for whatever reason) in the middle of a test.  I'm not sure if the conditions under which each test is run are guaranteed, or if a context switch could occur in the middle of a test run, potentially making the timings significantly longer than expected.

test/api/v3/unit/models/group.test.js
```diff
@@ -607,6 +613,99 @@ describe('Group Model', () => {
...
+
+      function generateTestMessage (overrides = {}) {
+        return Object.assign({}, {
+          text: 'test message',
+          uuid: testUserID,
+          timestamp: testTime,
+        }, overrides);
+      }
+
+      it('group that is not the tavern returns false, while tavern returns true', async () => {
```
khipkin, line 630: Suggestion: break this into two separate tests.

website/common/locales/en/messages.json
```diff
@@ -56,6 +56,7 @@
   "messageGroupChatFlagAlreadyReported": "You have already reported this message",
   "messageGroupChatNotFound": "Message not found!",
   "messageGroupChatAdminClearFlagCount": "Only an admin can clear the flag count!",
+  "messageGroupChatSpam": "Whoops, looks like you're posting too many messages! Please wait a minute and try again. The Tavern chat only holds 200 messages at a time, so Habitica encourages posting longer, more thoughtful messages and consolidating replies. Can't wait to hear what you have to say :)",
```
khipkin, line 59: Suggestion: add some sort of punctuation at the end, before the smiley face. :)

website/server/models/group.js
```diff
@@ -36,6 +36,16 @@ const LARGE_GROUP_COUNT_MESSAGE_CUTOFF = shared.constants.LARGE_GROUP_COUNT_MESS
 const CRON_SAFE_MODE = nconf.get('CRON_SAFE_MODE') === 'true';
 const CRON_SEMI_SAFE_MODE = nconf.get('CRON_SEMI_SAFE_MODE') === 'true';
 
+/*
+#  Spam constants to limit people from sending too many messages too quickly
+#    MESSAGE_LIMIT - The amount of messages that can be sent in a time window
+#    WINDOW_LENGTH - The window length for spam protection in milliseconds
+#    CONTRIBUTOR_LEVEL - Anyone at or above this level is exempt
```
khipkin, line 53: Suggestion: Make the constant names in this comment block match those below.

website/server/models/group.js
```diff
@@ -985,6 +995,31 @@ schema.methods.removeTask = async function groupRemoveTask (task) {
   }, {multi: true}).exec();
 };
 
+// Returns true if the user has reached the spam message limit
+schema.methods.checkChatSpam = function groupCheckChatSpam (user) {
+  if (this._id !== TAVERN_ID) {
+    return false;
+  } else if (user.contributor && user.contributor.level >= SPAM_MIN_EXEMPT_CONTRIB_LEVEL) {
```
khipkin, line 1216: Nit: `else` not needed here because of the previous return statement

test/api/v3/integration/chat/POST-chat.test.js
TheHollidayInn: I don't think you need to mock a timer. You can just hard code the timestamp. I think this is better covered in a unit test. The integration test shouldn't need to be so specific.

website/server/models/group.js
TheHollidayInn: Yea, these can be just broken into two if statements, but it is a style preference.

TheHollidayInn:
This work locally for me. @colingm Looks like there is a merge conflict and a few minor test changes to update. Then we can merge

TheHollidayInn:
Continuing in separate PR since this is stale.

TheHollidayInn:
Congratz on Tier 1 @colingm! Thank you for the PR :D

PR #8076, opened 2016-09-26
Add semantic-ui

paglias:
Notes:
- not sure website/client/semantic-ui is the most appropriate location for the theme
- the framework is pretty big so to process everything Webpack takes ~15s, so when components/app.vue (entry point where semantic is loaded) or a css file is edited it takes 15s to rebuild everything


paglias:
All the files added in website/client/semantic-ui/site aren't required for the code to work but are simply a list of the things that can be changed, we can remove them

One thing that might help with build times is only loading the components that we use in client/semantic-ui/semantic.less

paglias:
Build is failing because of https://github.com/HabitRPG/habitrpg/blob/paglias/semantic-ui/test/client/unit/index.js#L10 trying to load all files including those into /semantic-ui without a .less extension that is not understood

paglias:
Tests and perf should be solved. I'd like some input on the directory structure as placing /semantic-ui next to all the rest of JS is not something that I like at all

PR #8086, opened 2016-09-30
added balance to analytics

### LINKED ISSUE #8057: Amplitude should record the user's balance (amount of gems)
Our  Amplitude analytics data should include the user's number of gems (the user's `balance` object) to help us track payment / purchase problems.

Ideally it should report it as something like this `Balance   0.75 (3 Gems)` to avoid us making mistakes with the mental arithmetic - I know it's not hard, but it's even easier for the computer. ;)

If for any reason the balance doesn't exist or isn't a finite number, the code should report that (I have no idea if that's possible with our current code base or any future code, but it wouldn't be good to report a balance of 0 if there was actually some bug that prevented the balance existing at all).

**EDIT 2016-12-27: The work remaining to do is:**

- Leave the existing "balance" property as it is now.
- Create a new property called "balance gem amount" or similar (spaces are optional) which works in the same way as the "balance" property but displays the balance divided by 4. I want the word "balance" (lower case) to be at the start of the property name so that in Amplitude it appears immediately under the existing "balance" field (sorting is alphabetical there).

dumindux:
Fixes #8057
### Changes

---

UUID:

### DIFF
--- a/website/server/libs/analyticsService.js
+++ b/website/server/libs/analyticsService.js
@@ -66,6 +66,10 @@ let _formatUserData = (user) => {
     properties.Mana = Math.floor(user.stats.mp);
   }
 
+  if (user.balance) {
+    properties.balance = user.balance;
+  }
+
   properties.tutorialComplete = user.flags && user.flags.tour && user.flags.tour.intro === -2;
 
   if (user.habits && user.dailys && user.todos && user.rewards) {
--- a/website/server/libs/analyticsService.js
+++ b/website/server/libs/analyticsService.js
@@ -66,9 +66,7 @@ let _formatUserData = (user) => {
     properties.Mana = Math.floor(user.stats.mp);
   }
 
-  if (user.balance) {
-    properties.balance = user.balance;
-  }
+  properties.balance = user.balance;
 
   properties.tutorialComplete = user.flags && user.flags.tour && user.flags.tour.intro === -2;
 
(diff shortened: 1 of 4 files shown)


website/server/libs/analyticsService.js
```diff
@@ -66,6 +66,10 @@ let _formatUserData = (user) => {
     properties.Mana = Math.floor(user.stats.mp);
   }
 
+  if (user.balance) {
```
paglias: User.balance should Always exist. Also this check will Always returns false for a balsnce of 0 but we want to store it even if 0

paglias (changes requested):

website/server/libs/analyticsService.js
crookedneighbor: Agreed.

dumindux: Thanks for the suggestions. I will add the changes

crookedneighbor:
Don't forget to provide your user id.

You can find it in [Settings > API](https://habitrpg.com/#/options/settings/api). Be sure to only copy the User ID and _not_ your API Token.

paglias (approved):

paglias:
LGTM 👍

crookedneighbor:
Let me know your user id so I can credit you for your work!

dumindux:
Hi, my user id is 5fc49b3a-752c-4d35-9f73-287e19fdc96b

crookedneighbor:

dumindux:
Wow great!!. Thank you :).

Issue #8087, opened 2016-09-30
Improve API Documentation   (apidocs)


crookedneighbor:
_EDIT by Alys 2018-06-30:_
 
The only remaining action for this PR is to document the member route. It's been started (probably mostly done) by @madpink in https://github.com/HabitRPG/habitica/pull/9760 so that needs to be taken over by someone and finished (see the staff comments in that PR).

----

The docs for the API routes are, for the most part, very sparse. Often they don't provide enough information for actually using them.

There are a bunch of parameters that we can use with [API Doc](http://apidocjs.com/#params).

A good example of how each route should look is the [invite to group route](https://habitica.com/apidoc/#api-Group-InviteToGroup). Here it is in the [source code](https://github.com/HabitRPG/habitrpg/blob/ee15e29ba4d34812c4fd4e5836dc310d713a7db9/website/server/controllers/api-v3/groups.js#L643-L683).

When opening a PR, you should do it a little at a time. NOT ALL AT ONCE. For instance, one PR for each section.
- [x] Challenge -  @MathWhiz -- _(a mostly complete solution by JTorr is at https://github.com/HabitRPG/habitica/pull/8112 and if MathWhiz uses it in a PR that is merged, JTorr will get partial credit towards a tier)_  https://github.com/HabitRPG/habitica/pull/8626
- [x] Chat - @MathWhiz https://github.com/HabitRPG/habitica/pull/8277
- [x] Content (just document the top level parts of the object) - @MathWhiz #8098
- [x] Coupon - @MathWhiz https://github.com/HabitRPG/habitica/pull/8109
- [x] Data Export - @MathWhiz https://github.com/HabitRPG/habitica/pull/8120
- [x] Development - @MathWhiz https://github.com/HabitRPG/habitica/pull/8480
- [x] Group - @MathWhiz https://github.com/HabitRPG/habitica/pull/8522
  - [x] Invite to Group - @JTorr #8091
- [x] Hall - @Althenis #8536
- [ ] Member - continue from https://github.com/HabitRPG/habitica/pull/9760
- [x] Meta - @MathWhiz https://github.com/HabitRPG/habitica/pull/8167
- [x] Status - @MathWhiz #8165 
- [x] Tag - @madpink #8400 
- [x] Task - @madpink https://github.com/HabitRPG/habitica/pull/8447
- [x] Unsubscribe - @MathWhiz #8187 
- [x] User - a maximum of 10 routes per PR:
[...]

PR #8091, opened 2016-10-02
Improve API Docs for Invite to Group, Iss#8087

JTorr:
https://github.com/HabitRPG/habitica/issues/8087
### Changes

Update documentation for Invite to Group to more accurately reflect current structure of parameters and error types.

screen shot 2016-10-02 at 7 25 18 pm
screen shot 2016-10-02 at 7 25 28 pm

---

UUID: 1bf2957d-600e-4952-9e7b-d796cf99763d

### DIFF
--- a/website/server/controllers/api-v3/groups.js
+++ b/website/server/controllers/api-v3/groups.js
@@ -22,6 +22,11 @@ import { sendNotification as sendPushNotification } from '../../libs/pushNotific
 import pusher from '../../libs/pusher';
 
 /**
+ * @apiDefine GroupBodyInvalid
+ * @apiError (400) {BadRequest} GroupBodyInvalid A parameter in the group body was invalid.
+ */
+
+/**
  * @apiDefine GroupNotFound
  * @apiError (404) {NotFound} GroupNotFound The specified group could not be found.
  */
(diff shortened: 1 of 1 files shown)


website/server/controllers/api-v3/groups.js
```diff
@@ -648,38 +648,64 @@ async function _inviteByEmail (invite, group, inviter, req, res) {
  *
  * @apiParam {String} groupId The group _id ('party' for the user party and 'habitrpg' for tavern are accepted)
  *
- * @apiParam {Array} [emails] Body parameter - An array of emails addresses to invite
+ * @apiParam {Array} [emails] Body parameter - An array of objects, each
```
crookedneighbor: I think the preferred way of writing this is:

``` js
/*
 * // other stuff
 * @apiParam {Object[]} [emails] Body parameter - An array of objects, each representing one email address to invite
 * @apiParam {String} emails.email The email address of the user being invited
 * @apiParam {String} [emails.name] The name of the user being invited
*/
```

website/server/controllers/api-v3/groups.js
```diff
@@ -648,38 +648,64 @@ async function _inviteByEmail (invite, group, inviter, req, res) {
...
+ * email address. Key 'name' is optional, value is invitee's name.
  * @apiParam {Array} [uuids] Body parameter - An array of uuids to invite
  *
  * @apiParamExample {json} Emails
- *   {
- *     "emails": ["user-1@example.com", "user-2@exmaple.com"]
- *   }
+ * {
+ *    "emails": [
+ *      {"email": "user-1@example.com"},
```
crookedneighbor, line 685: Good catch on this.

website/server/controllers/api-v3/groups.js
```diff
@@ -648,38 +648,64 @@ async function _inviteByEmail (invite, group, inviter, req, res) {
...
  *
- * @apiSuccess {Array} data The invites
+ * @apiSuccess {Array} data The invitees
  * @apiSuccess {Object} data[0] If the invitation was a user id, you'll receive back an object. You'll recieve one Object for each succesful user id invite.
  * @apiSuccess {String} data[1] If the invitation was an email, you'll receive back the email. You'll recieve one String for each successful email invite.
  *
- * @apiSuccessExample {json} Successful Response
+ * @apiSuccessExample {json} Successful Response with Emails
+ * {
+ *   "success": true,
```
crookedneighbor: I think we should provide some boiler plate at the top describing the success and notifications properties. No need to provide them in every response example.

website/server/controllers/api-v3/groups.js
```diff
@@ -648,38 +648,64 @@ async function _inviteByEmail (invite, group, inviter, req, res) {
...
  *   }
  *
- * @apiError GroupNotFound The group could not be found
- * @apiError InvalidInvitationParams An error relating to the data sent in `emails` and/or `uuids`
- * @apiError TooManyInvites A max of 100 invites (combined emails and user ids) can be sent out at a time
+ * @apiError BadRequest Request body has missing or invalid content.
+ * @apiError NotAuthorized Missing API user or API key, or user already invited
+ * @apiError NotFound The group could not be found
+ *
+ * @apiErrorExample
```
crookedneighbor: Let's not include `apiErrorExample`s unless there's some really tricky case.

crookedneighbor (changes requested):

website/server/controllers/api-v3/groups.js
JTorr: Ah, thank you. Will fix.

JTorr:
@crookedneighbor Please review changes. Thanks!

website/server/controllers/api-v3/groups.js
crookedneighbor: Looks like this still needs to be addressed.

JTorr: I see that I missed the Notification property. Just to clarify, do you mean the success and notification properties should be at the top of the file (representing all of the Groups API calls), or together with the other properties (ex. line 676)

crookedneighbor: I'm saying that every request, regardless of group, has these properties. We don't need to put them in every success example. We should adjust the template to include some introductory information about requests. Not something that needs to be done in this PR, but you should remove the success and notifications keys from this example.

crookedneighbor:
Just pushed a commit that illustrates how errors should be structured. https://github.com/HabitRPG/habitica/commit/2dc09586784366fcb139c377d3993c52ea5ac9da

Any error that would be repeated should be defined at the top of the appropriate file. For instance, the `GroupNotFound` error is defined here: https://github.com/HabitRPG/habitica/commit/2dc09586784366fcb139c377d3993c52ea5ac9da#diff-b950a7c557b35ea58bc2685532c46170R25

```
/**
 * @apiDefine GroupNotFound
 * @apiError (404) {NotFound} GroupNotFound The specified group could not be found.
 */
```

And it's used like this:

```
* @apiUse GroupNotFound
```

If it's an error that only occurs in a single route, it should be defined in that route:

```
 * @apiError (statusCode) {ErrorType} NameForError A description of the error
```

JTorr:
@crookedneighbor Thanks for the example. Very helpful. LMK if these changes are what you were looking for.

crookedneighbor:
What I'm saying regarding the `success` and `notifications` params returned in a successful response is that we _should not_ include them on every request. Instead, we should add information in the header of the API doc (see http://apidocjs.com/#headerfooter) about them. We should also include information about authenticating.

Those seem like separate concerns from this PR. So for now, just remove the information about `success` and `notifications`.

For the errors, I don't think we should generalize them. IE, let's be specific to the resource:

``` js
// only applicable to the invite route, so we just add it there
* @apiError (400) {BadRequest} NoEmailProvided An email address was not provided in the `emails` body param
// for group update and creation, so we define it at the top and use `@apiUse GroupBodyInvalid` in the routes
* @apDefine GroupBodyInvalid
* @apiError (400) {BadRequest} GroupBodyInvalid A parameter in the group body was invalid
```

Also, just opened a PR to correct our use of `NotAuthorized` which should only be used for authentication. Here's a link to the changes to the `invite` route: https://github.com/HabitRPG/habitica/pull/8103/files#diff-b950a7c557b35ea58bc2685532c46170R550

JTorr:
@crookedneighbor How's it look now?

website/server/controllers/api-v3/groups.js
```diff
@@ -669,38 +674,68 @@ async function _inviteByEmail (invite, group, inviter, req, res) {
...
- *     "emails": ["user-1@example.com", "user-2@exmaple.com"],
- *     "uuids": ["user-id-of-existing-user", "user-id-of-another-existing-user"],
- *   }
+ * {
+ *   "emails": [
+ *       {"email": "user-1@example.com"},
+ *       {"email": "user-2@example.com"}
+ *   ],
+ *   "uuids": ["user-id-of-existing-user"]
+ *}
```
crookedneighbor: Space here between `*` and `}`

website/server/controllers/api-v3/groups.js
```diff
@@ -669,38 +674,68 @@ async function _inviteByEmail (invite, group, inviter, req, res) {
...
+ * @apiError (400) {BadRequest} UuidOrEmailOnly The `emails` and `uuids` params
+ * were both missing and/or a key other than `emails` or `uuids` was provided in
+ * the body param.
+ * @apiError (400) {BadRequest} CannotInviteSelf User id or email of invitee matches that
+ * of inviter.
+ * @apiError (400) {BadRequest} UserAlreadyInvited The user has already been
+ * invited to the group.
+ * @apiError (400) {BadRequest} UserAlreadyInGroup The user is already a member
+ * of the group.
+ * @apiError (400) {BadRequest} MustBeArray The `uuids` or emails body param is
```
crookedneighbor: emails should be surrounded in back ticks like uuids is.

website/server/controllers/api-v3/groups.js
```diff
@@ -669,38 +674,68 @@ async function _inviteByEmail (invite, group, inviter, req, res) {
...
+ * not an array.
+ * @apiError (400) {BadRequest} UserNotFound One or more of the `uuids` does not
```
crookedneighbor: It's actually a `404 NotFound` error. You can just use the `UserNotFound` error defined in the user controller.

```
@apiUse UseNotFound
```

crookedneighbor (changes requested):
We decided that those changes I made in the PR I linked would be a breaking change. So the errors need to stay as `NotAuthorized`.

Many of the `BadRequest` errors you have listed should actually be listed as `NotAuthorized`.

JTorr:
@crookedneighbor Issues addressed.

crookedneighbor (approved):

crookedneighbor:
Thanks for your patience. Notated toward your next tier.

PR #8092, opened 2016-10-03
Critical Hits now affect boss damage fixes #5429

### LINKED ISSUE #5429: Critical hits don't add to quest damage
While poking around in the algorithms for doing damage to quest monsters, I noticed that critical hits don't actually do any more damage to boss monsters than normal clicks. Crits are assessed during `addPoints()`, which occurs after `changeTaskValue()` where boss damage is applied.

This strikes me as very sad! Warriors ought to dish out extra damage via their mighty critical blows, but alas, they avail them nothing.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/22501904-critical-hits-don-t-add-to-quest-damage?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

abatula:
Fixes #5429
### Changes

Critical hit bonus now increases damage done to quest bosses.

Calculations for critical hits is now done in in `_changeTaskValue()` instead of `_addPoints()`, so that it is included when boss damage is calculated. The value stored in `user._tmp.crit` is used in `_addPoints()` to access the value of critical hits.

The `crit()` function for calculating critical hits is now exported within an object so that it can be stubbed for testing. References to the `crit()` function were updated to refer to `crit.crit()`.

Added a test function to verify that critical hits increase gold, mana, experience, and boss damage but do not affect the task value.

---


test/api/v3/integration/tasks/challenges/POST-tasks_challenges_challengeId_tasks_id_score_direction.test.js
```diff
@@ -78,7 +78,7 @@ describe('POST /tasks/:id/score/:direction', () => {
       previousTaskHistory = task.history[0];
 
       expect(task.history).to.have.lengthOf(1);
-      expect(task.value).to.equal(1);
+      expect(task.value).to.equal(1.5);
```
crookedneighbor: Let's save the original task value and just assert that the new task value is larger.

crookedneighbor:
``` js
expect(task.value).to.be.greaterThan(originalTaskValue);
```

website/common/script/ops/scoreTask.js
```diff
@@ -40,6 +40,13 @@ function _calculateDelta (task, direction, cron) {
     }
   }
 
+  // ===== CRITICAL HITS =====
```
crookedneighbor: So critical hits now affect the value of tasks? I'm not sure that's something we want.

Also, what happens when a task is unchecked? Since critical hits can only happen on task completion, the task value will never go back down to what it was before.

crookedneighbor (changes requested):
Is there a reason we can's just apply the value in `user._temp.crit` when calculating the boss damage?

crookedneighbor:
How is this going?

abatula:
As long as tests pass, I should be pushing a fix very soon.

crookedneighbor:
:+1:

abatula:
Critical hits no longer affect task value calculations. Critical hit calculation is now in `_changeTaskValue()`, and `user._tmp.crit` value is used in `_addPoints()` for gold and mp calculations.

website/common/script/ops/scoreTask.js
```diff
@@ -103,11 +103,8 @@ function _subtractPoints (user, task, stats, delta) {
 }
 
 function _addPoints (user, task, stats, direction, delta) {
-  // ===== CRITICAL HITS =====
-  // allow critical hit only when checking off a task, not when unchecking it:
-  let _crit = delta > 0 ? crit(user) : 1;
-  // if there was a crit, alert the user via notification
-  if (_crit > 1) user._tmp.crit = _crit;
+  // Check for a critical hit from _changeTaskValue()
```
crookedneighbor: I don't think this comment is necessary.

crookedneighbor (dismissed):
I feel like this behavior should be reflected in the tests, but I'm not sure how tricky that will be. (have not looked at the tests yet

abatula:
Test of critical hit behavior created, except I can't convince it to make a stub out of the `crit` function. I have also tried `import * as crit from <fileName>` and `sinon.stub(crit,'crit').returns(2)` with no luck.

crookedneighbor:
Unfortunately, to stub something, it needs to be a function on an object. You can't stub a function by itself.

crookedneighbor:
So crit may need to be changed to export like this:

``` js
function crit () { /* stuff */ }

module.exports = {
  crit,
};
```

Then you should be able to stub it.

test/common/ops/scoreTask.test.js
```diff
@@ -1,4 +1,7 @@
+import sinon from 'sinon'; // eslint-disable-line no-shadow
```
crookedneighbor: Pretty sure `sinon` and `sandbox` are already test globals.

test/common/ops/scoreTask.test.js
```diff
@@ -142,6 +146,31 @@ describe('shared.ops.scoreTask', () => {
...
+      let critUser = ref.afterUser;
+      expect(critUser.party.quest.progress.up).to.eql(0);
+      critUser.party.quest.key = 'gryphon';
+      let normalTask = todo;
+      let critTask = freshTodo;
+
+      scoreTask({ user: normalUser, task: normalTask, direction: 'up', cron: false });
+      let normalTaskDelta = normalUser.party.quest.progress.up;
+
+      sinon.stub(crit).returns(2);
```
crookedneighbor: Better to use `sandbox` so it's restored automatically after the test.

abatula:
Completed test of critical hits. 

Test function verifies that critical hits increase gold, mana, experience, and boss damage but do not affect the task value.

The `crit()` function for calculating critical hits is now exported within an object so that it can be stubbed for testing. References to the `crit()` function were updated to refer to `crit.crit()`.

abatula:
Does this pull request need any other tests or changes?

abatula:
Are there any other changes needed for this pull request, or is it ready to merge?

TheHollidayInn:
Looks good to me.

TheHollidayInn:
Noted towards your Tier. Thank you!

Issue #4460, opened 2015-01-01
return key causes chat messages to be submitted - should we change that?


Alys:
When you're typing a chat message, the keyboard's enter key causes it to be sent instead of producing a new line. That's common behaviour for HTML forms when you're typing in text boxes, but not in textareas, which the chat message box appears visually to be.

When writing long messages, new lines are beneficial and it's annoying to have to type shift-enter (also easy to forget).

We get quite a few users who submit multiple messages when really they should be in a single one, and I suspect that at least sometimes that's because of the enter key behaviour. 

So, should we change the chat form so that enter in the textarea will not submit it?

Users who do want to submit with keystrokes could still hit tab twice to put the focus on the submit button (and we could make that more convenient by moving the markdown link out of the tab path, or using tabindex to change the focus order).
##

crookedneighbor:
I'm in favor of this. 

I just did a version that changes it from `ui-keypress='{13:"postChat(group,message.content)"}'` to `ui-keydown='{"meta-enter":"postChat(group,message.content)"}'` which submits the form on ctrl/cmd + enter, just like Github has for comments.

We could also move the markdown link for those used to tabbing to submit forms.

Regardless, I think there would need to be a Bailey announcement to go along with this, I know I would be confused if the chat system suddenly changed. We'll stick get people reporting the "bug", but I think far fewer if we include it in an announcement.

SabreCat:
:+1:

crookedneighbor:
If we enable `meta-enter` for the chats, we should do the same for private messages.

UnexpectedMushroom:
Please do, that would be very helpful.

paglias:
Closing in favor of #8066

PR #8096, opened 2016-10-03
Hitting enter no longer sends a chat message, instead inserts a new line

### LINKED ISSUE #8066: Enter/Return key should not post chat messages
Currently, when typing a message in the Tavern, guilds, or party, hitting the keyboard's Enter key will post the message. This causes users who aren't familiar with this behaviour to post multiple messages when they meant to post only one.

**This should be changed so that Enter does nothing but make a new line (as Shift-Enter does now).**

This is important because we want to introduce an anti-spam limit on posts (https://github.com/HabitRPG/habitrpg/issues/8060) and we don't want users to accidentally exceed the limit by posting messages when they don't mean to.

Users who want to use keyboard controls to post messages can do Tab-Enter or Tab-Space (Tab moves focus to the send button and Enter or Space "clicks" the button). This behaviour must keep working. We can inform the Socialites about that so that they can share it to people who aren't happy with the removal of the Enter key's function.

(We could change Shift-Enter so that it posts the message, but many users are already used to using Shift-Enter to insert line breaks, so I personally think that changing that isn't a good idea at the moment. Perhaps a few months after Enter stops posting messages we could consider making Shift-Enter post them.)

FYI @lemoness

Hus274:
Fixes #8066 
### Changes

changing behavior so hitting enter in a chat box only now inserts a newline instead of submitting the form.

The user can now tab to send chat and click enter or use Control(command) + enter to submit thanks to Blade's feature here: https://github.com/HabitRPG/habitica/issues/4460

### DIFF
--- a/website/common/locales/en/groups.json
+++ b/website/common/locales/en/groups.json
@@ -46,6 +46,7 @@
     "chat": "Chat",
     "sendChat": "Send Chat",
     "toolTipMsg": "Fetch Recent Messages",
+    "sendChatToolTip": "You can send a chat from the keyboard by tabbing to the 'Send Chat' button and pressing Enter or by pressing Control (Command on a Mac) + Enter.",
     "syncPartyAndChat": "Sync Party and Chat",
     "guildBankPop1": "Guild Bank",
     "guildBankPop2": "Gems which your guild leader can use for challenge prizes.",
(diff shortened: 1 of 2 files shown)


crookedneighbor:
The send chat button should have a tooltip that describes how to send a chat from the keyboard. We should also be sure that we can tab to the send button from the text area.

Hus274:
I tested the ability to tab to the send button from the text area and it worked for me. I'll add in the tool tip for the submit button here.

Hus274:
Added the tool tip message. It's a bit wordy and might need a bit of word smithing. Just let me know and I can update it!

I was also thinking that it could potentially be a loading screen tip!

crookedneighbor:

PR #8099, opened 2016-10-04
[WIP] Group tasks claim

TheHollidayInn:
Fixes put_issue_url_here
### Changes

---

UUID:


website/client-old/js/services/taskServices.js
```diff
@@ -106,6 +106,21 @@ angular.module('habitrpg')
       });
     };
 
+    function getGroupTasks (groupId) {
```
paglias: I don't remember, these client side files are from the old PR or are new?

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -122,7 +122,7 @@ api.assignTask = {
     let group = await Group.getGroup({user, groupId: task.group.id, fields: requiredGroupFields});
     if (!group) throw new NotFound(res.t('groupNotFound'));
 
-    if (group.leader !== user._id) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
+    if (group.leader !== user._id && user._id !== assignedUserId) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
```
paglias, line 125: this is correct? Does it mean that the user who assign must be the assigned one?

paglias:
For the client side files, if you're adding them to every PR I think it's enough to just remove them from the manifest.json file so they're not loaded in production

website/client-old/js/services/taskServices.js
TheHollidayInn: Eh, these were duplicated during my merge. Let me remove the extra functions.

website/server/controllers/api-v3/tasks/groups.js
TheHollidayInn: This means that the user has to be the leader or they have to be assigning themselves. 

assignedUserId is the user that is being assigned. So, if the user._id is equal to the user being assigned, they are assigning themselves. 

You can also look at the opposite logic: (!condition1 && !condition2) = (condition1 || condition2) if that helps...

TheHollidayInn:
What is the correct way to add them for every PR? I assumed I would add them to the manifest - let the review happen- then remove them from the manifest.

website/server/controllers/api-v3/tasks/groups.js
paglias: You're right, sorry I misread that

paglias:
It's fine the way you did it, I got confused by the duplicate functions and thought that you removed all the client side PR and added it back

website/views/shared/tasks/edit/advanced_options.jade
```diff
@@ -13,6 +13,9 @@ div(ng-if='::task.type!="reward"')
     group-tasks-actions(ng-if="obj.type == 'guild'", task='task', group='obj')
   
   div(ng-show='task._edit._advanced')
+    group-tasks-actions
+
+  div(ng-show='task._edit._advanced')
```
paglias, line 15: it's already at line 15, it's needed to duplicate it?

TheHollidayInn: Nope! Updated.

TheHollidayInn:
Not sure why the branch failed. Local test run and there is no output.

PR #8102, opened 2016-10-05
Fixing debug scripts

Hus274:
### Changes

The debug scripts were broken when the common package was moved into the website folder. This pull request simply fixes the path variables so the scripts work again.

### DIFF
--- a/debug-scripts/grant-all-equipment.js
+++ b/debug-scripts/grant-all-equipment.js
@@ -13,7 +13,7 @@ if (!userId) {
   return;
 }
 
-let gearFlat = require('../common').content.gear.flat;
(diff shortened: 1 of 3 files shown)


crookedneighbor:
Actually, I've been meaning to remove these since I added the functionality to the footer in dev mode.

crookedneighbor:
screen shot 2016-10-04 at 8 39 44 pm

Hus274:
Ahh sweet, that's way better. We should update the readme file to remove the references to those scripts and remove those then.

crookedneighbor:
Done. 😄

Hus274:
Awesome!

PR #8103, opened 2016-10-05
chore: Correct use of NotAuthorized Errors to Forbidden

crookedneighbor:
@paglias you were right that a 401 NotAuthorized error should only be used in the case that a user needs to authenticate. I've switched all of the NotAuthorized errors to 403 Forbidden errors (except for the auth ones)

I caught some some BadRequest errors that should actually be 403 Forbidden, and I'm sure there are more, but we can do those later.


crookedneighbor:
Since we're starting to document the errors returned from routes, I wanted us to use the correct ones.

crookedneighbor:
We determined that changing the error type like this would be a breaking change. Closing.

Issue #7722, opened 2016-06-25
Gem Receiving text is in native language of sender


cTheDragons:
_Admin note: Edits made to bring original ticket up to date with conversation_
## General Info

bc10ec41-4213-4482-bbb8-558e3c3ea8d5
Opera 37.0.2178.43
Windows 8.1 32-bit
## Description
- Text for the receiver of a gem gift appears in the native language of the sender. _(See Example below)_.  **The text for the sender of the gem should be in the native language of the sender and for the receiver the gem receive text would be in the native language of the receiver.**
- ~~The @ symbol is missing in front of the usernames. (This occurs regardless if the sender is sending in a different language.~~

For example a French  User sending to myself, an English speaker, I myself see this:

> `Bonjour cTheDragons, FrenchHabitican vous a envoyé1 gemmes!` Extra user comment here

It should appear as:

> `Hello cTheDragons, FrenchHabitican has sent you 1 gems!`  Extra user comment here

habitrpguser:
Just confirming this (though only by hearsay): Someone mentioned in guild talk on 29 July 2016:

> "@XYZ Spreek je Nederlands? (Your message came up as "XYZ heeft je gezonden 4 edelstenen", which I assume means XYZ has given you 4 gems)"

So I went looking if this is already a known bug - and here it is! :)

khipkin:
Suggestion: add [translations] label.

crookedneighbor:
We've decided that the leading @ should not be added.

khipkin:
Suggestion: update issue title to prevent future wasted effort

librarianmage:
Should this be marked as In Progress?

PR #8104, opened 2016-10-05
Add @-prefix to usernames in gift messages

mrhmouse:
This addresses part of #7722

---

UUID: e345f9e6-81d2-4494-8a4a-9c263e174dac

### DIFF
--- a/website/common/locales/bg/groups.json
+++ b/website/common/locales/bg/groups.json
@@ -111,7 +111,7 @@
     "toUserIDRequired": "Нужен е потребителски идентификатор",
     "gemAmountRequired": "Нужен е брой диаманти",
     "notAuthorizedToSendMessageToThisUser": "Не можете да изпратите съобщение до този потребител",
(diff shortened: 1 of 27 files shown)


khipkin:
@mrhmouse Please undo your changes in locale directories other than "en".  See the [locales README](https://github.com/HabitRPG/habitica/blob/develop/website/common/locales/README.md) for more details.

mrhmouse:
Would this be better fixed by just prefixing the names with '@' prior to handing them to the template? Or should the '@' go in the template like it is now..?

crookedneighbor:
Personally, I don't think we need to add the `@` sign at all. It doesn't have any meaning within private messages like it does in the chat.

crookedneighbor:
@Alys @lemoness @veeeeeee thoughts?

lemoness:
I don't think we need the @ symbol. We don't display it in most system messages.

PR #8105, opened 2016-10-05
Subscriptions Fixes

### LINKED ISSUE #4846: negative subscription credit
It says I have "-2.95 months of subscription credit." I use to donate monthly, then I canceled, and now I am donating again monthly.
snapshot2

Shows same thing in Firefox and Chrome.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9478666-negative-subscription-credit?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SabreCat:
Fixes https://github.com/HabitRPG/habitica/issues/4846
### Changes
- Subscription credits can no longer go negative when resubscribing after a lapse.
- Gem cap no longer resets when a user purchases or is gifted an additional month of subscription during an ongoing subscription.
- When converting month credits to subscription duration, use a factor of 30.5 days/month instead of 30 to avoid shorting the user a day for each 31-day month.
- A user with continuous subscription who logs in after more than a month's absence correctly receives consecutive subscription benefits for the paid-for but unused month or months.
- Some code refactored to normalize `moment`s to days without sub-day data, and store repeated uses of `new Date()` in variables, to eliminate possible race scenarios.

I need to add/update some tests to cover the fixed scenarios, and more fixes are forthcoming, but the code thus far can be reviewed in the meantime.

### DIFF
--- a/website/server/libs/cron.js
+++ b/website/server/libs/cron.js
@@ -8,7 +8,6 @@ import nconf from 'nconf';
 
 const CRON_SAFE_MODE = nconf.get('CRON_SAFE_MODE') === 'true';
 const CRON_SEMI_SAFE_MODE = nconf.get('CRON_SEMI_SAFE_MODE') === 'true';
-const YEAR_MONTH_FORMAT = 'YYYY-MM';
 const shouldDo = common.shouldDo;
 const scoreTask = common.ops.scoreTask;
 // const maxPMs = 200;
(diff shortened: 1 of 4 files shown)


website/server/libs/cron.js
```diff
@@ -46,24 +46,27 @@ let CLEAR_BUFFS = {
 
 function grantEndOfTheMonthPerks (user, now) {
   let plan = user.purchased.plan;
+  let subscriptionEndDate = moment(plan.dateTerminated).isBefore() ? moment(plan.dateTerminated).format('YYYY-MM') : moment(now).format('YYYY-MM');
```
khipkin: Suggestion: store this date format string in a variable somewhere.

website/server/libs/cron.js
```diff
@@ -121,6 +124,10 @@ export function cron (options = {}) {
   // "Perfect Day" achievement for perfect days
   let perfect = true;
 
+  // Reset Gold-to-Gems cap if it's the start of the month
+  if (user.purchased && user.purchased.plan && moment(user.purchased.plan.dateUpdated).format('MMYYYY') !== moment().format('MMYYYY')) {
```
khipkin: If `user.purchased.plan` exists, will `user.purchased.plan.dateUpdated` always exist?

crookedneighbor: If it's undefined, `moment` will use the current date.

website/server/libs/cron.js
SabreCat: Done!

PR #8106, opened 2016-10-05
fix quest progress reporting whitespace

kees:
Notifications of other things (HP, GP, etc) have a regular format of
"+/- NUM THING". For example:

  function gp(val, bonus) {
    _notify(_sign(val) + " " + coins(val - bonus), 'gp');
  }

However, the recent quest collection/damage notifications do not. This
attempts to regularize the reporting by adding in the "missing" space.

---

UUID: 4c122516-014d-4490-a433-1df4bafb844f


crookedneighbor (changes requested):
Please undo your changes in locale directories other than "en".  The translation work is done on Transifex, not directly in the repo. See the [locales README](https://github.com/HabitRPG/habitica/blob/develop/website/common/locales/README.md) for more details.

kees:
Thanks for the correction! I've dropped all the other locale changes now.

crookedneighbor:

Issue #8072, opened 2016-09-25
when a quest starts, a `system` message should be written to party chat


Alys:
When a quest starts (i.e., after the invitation phase is finished) a message should be posted to party chat saying something like:
`Your quest, The Call of Octothulu, has started.`

It should  be posted by the system user, as for other automated quest messages.

This would help players be aware of when they need to watch their Dailies carefully, and would help admins with troubleshooting quest bugs.

In addition, I'd like a new `info` field to be added to the system message, listing the UIDs of the participants (see below for an example). This is solely to help with troubleshooting for completed quests so it doesn't need to be visible to the players or accessible through the API (unless any other admins disagree). I've chosen `info` as the field name so that in future it can be used for other purposes too.

```
        {
            "id": "2498af5a-cf78-4e22-93b4-3f41d79f517c",
            "text": "`Your quest, The Call of Octothulu, has started.`",
            "info": "participants: 9264741a-e7df-4202-8eaf-5e32bba01f1c, f58f8ff8-8a7f-4a49-8cda-789354b5be92, 7b55f43a-49b5-4f05-bff1-fb3994311dc0",
            "timestamp": 1474102683475,
            "likes": {},
            "flags": {},
            "flagCount": 0,
            "uuid": "system"
        },
```

lemoness:
This looks good to me! Go ahead and mark it "help welcome."

dumindux:
Hi, I would like to work on this issue. Its great if you can point me where to get started. Thanks.

NicoSiebler:
@dumindux I already tried to solve this for the first part of the issue in a short time without luck yet and I didn't mention this as this was my first task try. So I hope this could help:
My guess was to adjust the `startQuest()` method in **server/models/group.js** directly above the `sendTxnEmail()` with a `this.sendChat()` but the object was missing a usable chat to send the message and I didn't look deep enough what I did wrong or what was missing. Maybe it's simply the wrong spot or approach. I would be interested in the proper solution.

dumindux:
@NicoSiebler Take this issue if you are already working on this :) .

PR #8107, opened 2016-10-06
added message saying quest is started

dumindux:
https://github.com/HabitRPG/habitica/issues/8072
### Changes

---

UUID: dumindux

### DIFF
--- a/test/api/v3/integration/quests/POST-groups_groupid_quests_reject.test.js
+++ b/test/api/v3/integration/quests/POST-groups_groupid_quests_reject.test.js
@@ -180,5 +180,15 @@ describe('POST /groups/:groupId/quests/reject', () => {
       expect(rejectingMember.party.quest.key).to.not.exist;
       expect(rejectingMember.party.quest.completed).to.not.exist;
     });
+
+    it('starts the quest when the last user reject and verifies chat', async () => {
+      await leader.post(`/groups/${questingGroup._id}/quests/invite/${PET_QUEST}`);
+      await partyMembers[0].post(`/groups/${questingGroup._id}/quests/accept`);
+      await partyMembers[1].post(`/groups/${questingGroup._id}/quests/reject`);
+      await questingGroup.sync();
+
+      expect(questingGroup.chat[0].text).to.exist;
+      expect(questingGroup.chat[0]._meta).to.exist;
+    });
   });
 });
--- a/test/api/v3/unit/models/group.test.js
+++ b/test/api/v3/unit/models/group.test.js
@@ -366,7 +366,7 @@ describe('Group Model', () => {
 
           party = await Group.findOne({_id: party._id});
 
-          expect(Group.prototype.sendChat).to.be.calledOnce;
+          expect(Group.prototype.sendChat).to.be.calledTwice;
           expect(Group.prototype.sendChat).to.be.calledWithMatch(/`Participating Member found/);
           expect(Group.prototype.sendChat).to.be.calledWithMatch(/0 Blue Fins/);
           expect(Group.prototype.sendChat).to.be.calledWithMatch(/0 Fire Coral/);
@@ -384,7 +384,7 @@ describe('Group Model', () => {
 
           party = await Group.findOne({_id: party._id});
 
-          expect(Group.prototype.sendChat).to.be.calledOnce;
+          expect(Group.prototype.sendChat).to.be.calledTwice;
           expect(Group.prototype.sendChat).to.be.calledWithMatch(/`Participating Member found/);
           expect(Group.prototype.sendChat).to.be.calledWithMatch(/\d* (Tracks|Broken Twigs)/);
         });
--- a/test/api/v3/unit/models/group.test.js
+++ b/test/api/v3/unit/models/group.test.js
@@ -360,13 +360,14 @@ describe('Group Model', () => {
           party.quest.active = false;
 
           await party.startQuest(questLeader);
+          Group.prototype.sendChat.reset();
           await party.save();
 
           await Group.processQuestProgress(participatingMember, progress);
 
           party = await Group.findOne({_id: party._id});
 
-          expect(Group.prototype.sendChat).to.be.calledTwice;
+          expect(Group.prototype.sendChat).to.be.calledOnce;
           expect(Group.prototype.sendChat).to.be.calledWithMatch(/`Participating Member found/);
           expect(Group.prototype.sendChat).to.be.calledWithMatch(/0 Blue Fins/);
           expect(Group.prototype.sendChat).to.be.calledWithMatch(/0 Fire Coral/);
@@ -378,13 +379,14 @@ describe('Group Model', () => {
           party.quest.active = false;
 
           await party.startQuest(questLeader);
+          Group.prototype.sendChat.reset();
           await party.save();
 
           await Group.processQuestProgress(participatingMember, progress);
 
           party = await Group.findOne({_id: party._id});
 
-          expect(Group.prototype.sendChat).to.be.calledTwice;
+          expect(Group.prototype.sendChat).to.be.calledOnce;
           expect(Group.prototype.sendChat).to.be.calledWithMatch(/`Participating Member found/);
           expect(Group.prototype.sendChat).to.be.calledWithMatch(/\d* (Tracks|Broken Twigs)/);
         });
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -418,6 +418,7 @@ schema.methods.startQuest = async function startQuest (user) {
   let userIsParticipating = this.quest.members[user._id];
   let quest = questScrolls[this.quest.key];
   let collected = {};
+  let group = this;
   if (quest.collect) {
     collected = _.transform(quest.collect, (result, n, itemToCollect) => {
       result[itemToCollect] = 0;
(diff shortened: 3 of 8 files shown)


dumindux:
I was looking around the codebase and added the message mentioned in the issue. But it does not get posted with the system user. I couldnt figure out how to do so. Any help on this is really appreciated.

crookedneighbor:
You can see here in the group model that there is a method on the group object that can send chats: https://github.com/HabitRPG/habitica/blob/develop/website/server/models/group.js#L672

You will want to add a call like that to the `startQuest` method. https://github.com/HabitRPG/habitica/blob/b83ef872c95d7e93b917edd55ab202a2110f5bad/website/server/models/group.js#L416

crookedneighbor (changes requested):
You implemented this client side on the app. But we actually want this to happen through server.

dumindux:
@crookedneighbor Thanks so much for the help. I have added the chat as you mentioned. But it throws an error when the quest starts saying the following.
https://gist.github.com/dumindux/be1649d90aa2b1ee01442753f506285c
do you see why?
I think is is due to group.chat not being initialized. But I cant see why.

Thanks.

dumindux:
I reinitilized the array to empty array when this.chat is undefined. But this erases the previous chats.

dumindux:
@crookedneighbor @Alys I have found where the issue was and fixed it. Please review the pull request. Thanks.

dumindux:
I cant think why the travis build fails. I think it is due to **`** in the chat message I send. But why does it give an error?

test/api/v3/unit/models/group.test.js
```diff
@@ -366,7 +366,7 @@ describe('Group Model', () => {
 
           party = await Group.findOne({_id: party._id});
 
-          expect(Group.prototype.sendChat).to.be.calledOnce;
+          expect(Group.prototype.sendChat).to.be.calledTwice;
```
crookedneighbor: Can we instead reset the spy after the `startQuest` call?

``` js
await party.startQuest(questLeader);
Group.prototype.sendChat.reset();
```

That will put the spy in a clean state for the assertion we're actually looking to make here.

website/server/models/group.js
```diff
@@ -418,6 +418,7 @@ schema.methods.startQuest = async function startQuest (user) {
   let userIsParticipating = this.quest.members[user._id];
   let quest = questScrolls[this.quest.key];
   let collected = {};
+  let group = this;
```
crookedneighbor: Do we need to assign a variable here? Can't we just do 

``` js
this.sendChat(`Your quest, ${quest.text('en')}, has started.`);
```

crookedneighbor (changes requested):
You're missing this part of the issue:

> In addition, I'd like a new info field to be added to the system message, listing the UIDs of the participants (see below for an example). This is solely to help with troubleshooting for completed quests so it doesn't need to be visible to the players or accessible through the API (unless any other admins disagree). I've chosen info as the field name so that in future it can be used for other purposes too.

When that's done, it should be an underscored property to designate that it is private. Probably it should also be hidden from the end user (but we can take care of that)

dumindux:
Incorporated the requested changes :)

website/server/models/group.js
```diff
@@ -349,7 +349,7 @@ schema.methods.isMember = function isGroupMember (user) {
   }
 };
 
-export function chatDefaults (msg, user) {
+export function chatDefaults (msg, user, participants) {
```
crookedneighbor: This is not something that should be added in chatDefaults. This is behavior particular to starting a quest. IMO, it doesn't make sense to add the logic here.

website/server/models/group.js
```diff
@@ -370,11 +370,23 @@ export function chatDefaults (msg, user) {
...
+    message._info = infoStr;
+  }
+
   return message;
 }
 
-schema.methods.sendChat = function sendChat (message, user) {
-  let newMessage = chatDefaults(message, user);
+schema.methods.sendChat = function sendChat (message, user, participants) {
+  let newMessage = chatDefaults(message, user, participants);
```
crookedneighbor: I'd prefer to see something like:

``` js
schema.methods.sendChat = function sendChat (message, user, metaData) {
let newMessage = chatDefaults(message, user)
newMessage._meta = metaData;
```

An underscored property, to designate that it is private and should not be relied on (as it could change at any time). Also, an object so that it's extensible and more properties can be added to it if needed.

Then when it is used in the quest start method:

``` js
this.sendChat(`Your quest, ${quest.text('en')}, has started.`, null, {
  participatingMembers: this.getParticipatingQuestMembers().join(', '),
});
```

@alys, thoughts?

crookedneighbor (changes requested):

website/server/models/group.js
Alys: Agreed. I like that much better than my idea.

dumindux:
Incorporated the requested changes

dumindux:
Fixed the test case. But not sure if it is the best way.

website/server/models/group.js
```diff
@@ -373,8 +373,9 @@ export function chatDefaults (msg, user) {
   return message;
 }
 
-schema.methods.sendChat = function sendChat (message, user) {
+schema.methods.sendChat = function sendChat (message, user, metaData) {
   let newMessage = chatDefaults(message, user);
+  newMessage._meta = metaData;
```
crookedneighbor: Let's do

``` js
if (metaData) {
  newMessage._meta = metaData;
}
```

Then you can delete the `message._meta = null;` in the test.

crookedneighbor (changes requested):
You still need tests around the behavior that starting a quest now sends a system message.

dumindux:
Thanks for the tip. you mean we need new test cases?

crookedneighbor:
Yes. You've changed the behavior of the method, so we need to test that behavior.

dumindux:
Ok. Can you point me to which files the new tests should go?

crookedneighbor:
At an integration test level, you could add an assertion to all the routes that can start a quest (accept, invite, reject, force start). https://github.com/HabitRPG/habitica/tree/develop/test/api/v3/integration/quests

dumindux:
Thank you. Should I assert the chat to be there in group.chat ?

dumindux:
Added new tests

dumindux:
any idea why travis fails? npm test ran fine in my local setup. Thank you.

Alys:
@dumindux It might have been a temporary failure. I've restarted the tests.

dumindux:
Ah.. its ok now.

dumindux:
@Alys @crookedneighbor Any feedback on this PR? :)

website/server/models/group.js
```diff
@@ -520,6 +523,9 @@ schema.methods.startQuest = async function startQuest (user) {
         });
     });
   });
+  this.sendChat(`Your quest, ${quest.text('en')}, has started.`, null, {
```
TheHollidayInn, line 526: I believe we should make this string translatable. That will allow you to remove the meta data addition as well. Unless you are using the meta data for something else. 

add this to common/locals
`"questStarted": "Your quest, <%= questText %>, has started. <%= participatingMembers %>."`

replace the first param of sent chat with
```
i18n.t('questStarted', { questText: quest.text('en'), participatingMembers: this.getParticipatingQuestMembers().join(', ')});
```

test/api/v3/integration/quests/POST-groups_groupid_quests_reject.test.js
```diff
@@ -180,5 +180,15 @@ describe('POST /groups/:groupId/quests/reject', () => {
...
       expect(rejectingMember.party.quest.completed).to.not.exist;
     });
+
+    it('starts the quest when the last user reject and verifies chat', async () => {
+      await leader.post(`/groups/${questingGroup._id}/quests/invite/${PET_QUEST}`);
+      await partyMembers[0].post(`/groups/${questingGroup._id}/quests/accept`);
+      await partyMembers[1].post(`/groups/${questingGroup._id}/quests/reject`);
+      await questingGroup.sync();
+
+      expect(questingGroup.chat[0].text).to.exist;
```
TheHollidayInn, line 190: All of these expects new to verify the contents of the text. So, something like:
`expect(questinGroup.chat[0].text).equals("Your quest has started....etc")`

website/server/models/group.js
Alys: But this is going in to party chat, not to a specific user, and we don't know what language the majority of the party speak. If the person who starts the quest speaks a language that none of the others do, then the translation won't be helpful (assuming it's that person's language that's used). Until we have a solution for party system messages to be translated into whatever language the reader is using, they should stay as English.

TheHollidayInn:
@dumindux I need your UUID. I don't see your user name in the hall list.

test/api/v3/integration/quests/POST-groups_groupid_quests_reject.test.js
khipkin: @TheHollidayInn This change was never made.

TheHollidayInn: I know. I've noted it for myself to do.

khipkin:
The quest-started message doesn't use backticks, so it stands out from the regular system messages:

image

TheHollidayInn:
Ah, You can open a separate issue for this if you want.

dumindux:
hi my uuid is 5fc49b3a-752c-4d35-9f73-287e19fdc96b

TheHollidayInn:
Added to your tier!

PR #8111, opened 2016-10-08
Group tasks approval

TheHollidayInn:
Fixes put_issue_url_here
### Changes

If a task is set to require approval, the following flow happens:
1. An assigned user receives a task that states approval is required
2. When a user attempts to score, approval will be requested via notification to group leader
3. The group leader can then query approvals need via the api
4. The group leader can then approve
5. The user will then be able to score the task

---

UUID:

### DIFF
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -316,13 +316,13 @@ api.scoreTask = {
 
     if (!task) throw new NotFound(res.t('taskNotFound'));
 
-    if (task.requiresApproval && !task.approved) {
-      if (task.approvalRequested) {
+    if (task.group.requiresApproval && !task.group.approved) {
+      if (task.group.approvalRequested) {
         throw new NotAuthorized(res.t('taskRequiresApproval'));
       }
 
-      task.approvalRequested = true;
-      task.approvalRequestedDate = new Date();
+      task.group.approvalRequested = true;
+      task.group.approvalRequestedDate = new Date();
 
       let group = await Group.getGroup({user, groupId: task.group.id, fields: requiredGroupFields});
       let groupLeader = await User.findById(group.leader); // Use this method so we can get access to notifications
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -316,13 +316,13 @@ api.scoreTask = {
 
     if (!task) throw new NotFound(res.t('taskNotFound'));
 
-    if (task.group.requiresApproval && !task.group.approved) {
-      if (task.group.approvalRequested) {
+    if (task.group.approval.required && !task.group.approval.approved) {
+      if (task.group.approval.requested) {
         throw new NotAuthorized(res.t('taskRequiresApproval'));
       }
 
-      task.group.approvalRequested = true;
-      task.group.approvalRequestedDate = new Date();
+      task.group.approval.requested = true;
+      task.group.approval.requestedDate = new Date();
 
       let group = await Group.getGroup({user, groupId: task.group.id, fields: requiredGroupFields});
       let groupLeader = await User.findById(group.leader); // Use this method so we can get access to notifications
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -316,7 +316,13 @@ api.scoreTask = {
     if (!task) throw new NotFound(res.t('taskNotFound'));
 
     if (task.requiresApproval && !task.approved) {
-      throw new NotAuthorized(res.t('taskRequiresApproval'));
+      if (task.approvalRequested) {
+        throw new NotAuthorized(res.t('taskRequiresApproval'));
+      }
+      task.approvalRequested = true;
+      task.approvalRequestedDate = new Date();
+      await task.save();
+      return res.respond(200, {message: res.t('taskRequiresApproval'), task});
     }
 
     let wasCompleted = task.completed;
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -325,7 +325,7 @@ api.scoreTask = {
       task.approvalRequestedDate = new Date();
 
       let group = await Group.getGroup({user, groupId: task.group.id, fields: requiredGroupFields});
-      let groupLeader = await User.findById(group.leader); //Use this method so we can get access to notifications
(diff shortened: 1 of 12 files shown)


website/server/models/task.js
```diff
@@ -71,6 +71,13 @@ export let TaskSchema = new Schema({
     taskId: {type: String, ref: 'Task', validate: [validator.isUUID, 'Invalid uuid.']},
   },
 
+  requiresApproval: {type: Boolean, default: false},
```
TheHollidayInn: @paglias Can I get your thoughts here. I wanted to either put these properties under group or under a approval object. However, I am having a trouble allow a user to set sub documents via the `user.post('tasks')` route. For example:

```
task = await user.post(`/tasks/group/${guild._id}`, {
      text: 'test todo',
      type: 'todo',
      'group.requiresApproval': true,
    });
```

Will not work, even if I remove group from the noSet. Do you have any idea why?

paglias: @TheHollidayInn this is not working because you cannot pass single nested props but have to pass the entire object. I would suggest to pass `{text, type, groupRequiresApproval}` and then based on the value of `groupRequiresApproval` adapt group.requiresApproval in the task. Or you could use a query parameter

paglias:
@TheHollidayInn can we try to get this fixed and merged in? I totally forgot about it but it's been more than a month and I'd like the branches not too divert too much

TheHollidayInn:
Definitely. I think it has only been 2 weeks, but I want to finish this and the other PR by the end of the weekend. I also forgot about this one.

website/server/models/task.js
```diff
@@ -69,6 +69,12 @@ export let TaskSchema = new Schema({
     broken: {type: String, enum: ['GROUP_DELETED', 'TASK_DELETED', 'UNSUBSCRIBED']},
     assignedUsers: [{type: String, ref: 'User', validate: [validator.isUUID, 'Invalid uuid.']}],
     taskId: {type: String, ref: 'Task', validate: [validator.isUUID, 'Invalid uuid.']},
+    requiresApproval: {type: Boolean, default: false},
+    approved: {type: Boolean, default: false},
+    approvedDate: {type: Date},
+    approvingUser: {type: String, ref: 'User', validate: [validator.isUUID, 'Invalid uuid.']},
+    approvalRequested: {type: Boolean, default: false},
+    approvalRequestedDate: {type: Date},
```
khipkin: Suggestion: create one new field here to house all of these approval-related fields. That way, there will be less clutter in the main `task.group` namespace.

TheHollidayInn: Yea, I thought about separating this into a approval sub-doc. Let's see what others think as well.

paglias: yeah i think using an object is the best thing here

website/server/controllers/api-v3/tasks.js
```diff
@@ -315,6 +316,28 @@ api.scoreTask = {
...
+      }
+
+      task.group.approvalRequested = true;
+      task.group.approvalRequestedDate = new Date();
+
+      let group = await Group.getGroup({user, groupId: task.group.id, fields: requiredGroupFields});
+      let groupLeader = await User.findById(group.leader); // Use this method so we can get access to notifications
+      groupLeader.addNotification('GROUP', {
+        message: res.t('userHasRequestedTaskApproval', {
+          user: user.auth.local.username,
```
paglias: user.profile.name because auth.local won't work with Facebook/Google users

website/server/controllers/api-v3/tasks.js
```diff
@@ -315,6 +316,28 @@ api.scoreTask = {
...
+      if (task.group.approvalRequested) {
+        throw new NotAuthorized(res.t('taskRequiresApproval'));
+      }
+
+      task.group.approvalRequested = true;
+      task.group.approvalRequestedDate = new Date();
+
+      let group = await Group.getGroup({user, groupId: task.group.id, fields: requiredGroupFields});
+      let groupLeader = await User.findById(group.leader); // Use this method so we can get access to notifications
+      groupLeader.addNotification('GROUP', {
```
paglias: let's rename this notification to something more specific than GROUP

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -178,4 +179,97 @@ api.unassignTask = {
...
+
+    let user = res.locals.user;
+    let groupId = req.params.groupId;
+
+    let group = await Group.getGroup({user, groupId, fields: requiredGroupFields});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    if (group.leader !== user._id) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
+
+    let approvals = await Tasks.Task.find({
```
paglias, line 265: is this meant to return the entire tasks or just an info on their approval status?

website/server/libs/taskManager.js
```diff
@@ -51,6 +51,10 @@ export async function createTasks (req, res, options = {}) {
     let taskType = taskData.type;
     let newTask = new Tasks[taskType](Tasks.Task.sanitize(taskData));
 
+    if (taskData.requiresApproval) {
```
paglias: let's put this in the if (group) at line 60 so that it does nothing when passed to a task that is not part of a group. Also this parameter needs to be documented

website/server/models/task.js
paglias: I think the best place is task.group.approval or something similar. It has the added benefit of denying changes to all these fields from normal users

paglias (changes requested):

website/server/models/task.js
TheHollidayInn: Hmm. I'm hesitant to make approval a subdoc of group. It doesn't really allow for any future flexibility. Can't we just mark approval as unsettable like group?

paglias: Why no flexibility? For subdivision I meant an object not a separate model in a separate file

TheHollidayInn: Because you limit the approval to only be part of the group. If you add the approval to the top level of the task, i.e. `tasks.approval.[property]` This allows for clarity in the future if we want other orgs, challenges, etc to use the approval object. It is the same argument I was trying to make about have a linked ID. We now have duplicate fields in the challenge and group subdoc.

website/server/controllers/api-v3/tasks/groups.js
TheHollidayInn: Yea, we can limit this to just return the task and userid.

paglias (approved):

PR #8112, opened 2016-10-09
Improve API Documentation for Challenges API, Iss#8087

JTorr:
### Fixes

https://github.com/HabitRPG/habitica/issues/8087
#### Update API documentation for /challenges

---

UUID: 1bf2957d-600e-4952-9e7b-d796cf99763d

### DIFF
--- a/website/server/controllers/api-v3/challenges.js
+++ b/website/server/controllers/api-v3/challenges.js
@@ -21,18 +21,143 @@ import csvStringify from '../../libs/csvStringify';
 let api = {};
 
 /**
+ * @apiDefine ChallengeLeader Challenge Leader
+ * The leader of the challenge can use this route.
+ */
+
+/**
  * @apiDefine ChallengeNotFound
  * @apiError (404) {NotFound} ChallengeNotFound The specified challenge could not be found.
  */
 
 /**
+ * @apiDefine SuccessfulChallengeRequest
+ * @apiSuccess {UUID} challenge.group._id The group id.
+ * @apiSuccess {String} challenge.group.type Group type: `Guild` or `Party`.
+ * @apiSuccess {String} challenge.group.privacy Group privacy: `Public` or `Private`.
+ * @apiSuccess {String} challenge.name Full name of challenge.
+ * @apiSuccess {String} challenge.shortName Short name of challenge.
+ * @apiSuccess {Object} challenge.leader User details of challenge leader.
+ * @apiSuccess {UUID} challenge.leader._id User id of challenge leader.
+ * @apiSuccess {Object} challenge.leader.profile Profile information of leader.
+ * @apiSuccess {Object} challenge.leader.profile.name Username of leader.
+ * @apiSuccess {String} challenge.updatedAt Timestamp of last update.
+ * @apiSuccess {String} challenge.createdAt Timestamp of challenge creation.
+ * @apiSuccess {UUID} challenge.id Id number of newly created challenge.
+ * @apiSuccess {UUID} challenge._id Same as `challenge.id`.
+ * @apiSuccess {String} challenge.prize Number of gems offered as prize to winner.
+ * @apiSuccess {String} challenge.memberCount Number users participating in challenge.
+ * @apiSuccess {Object} challenge.tasksOrder Object representing associated tasks and rewards.
+ * @apiSuccess {Array} challenge.tasksOrder.rewards Array of `reward` objects associated with challenge.
+ * @apiSuccess {Array} challenge.tasksOrder.todos Array of `todo` objects associated with challenge.
+ * @apiSuccess {Array} challenge.tasksOrder.dailys Array of `dailys` objects associated with challenge.
+ * @apiSuccess {Array} challenge.tasksOrder.habits Array of `habits` objects associated with challenge.
+ * @apiSuccess {Boolean} challenge.official Boolean indicating if this is an official Habitica challenge.
+ *
+ */
+
+/**
+ * @apiDefine ChallengeSuccessExample
+ * @apiSuccessExample {json} Sucessfull response with single challenge
+ {
+   "data": {
+     "group": {
+      "_id": "group-id-associated-with-challenge",
+      "name": "MyGroup",
+      "type": "guild",
+      "privacy": "public"
+     },
+     "name": "Long Detailed Name of Challenge",
+     "shortName": "my challenge",
+     "leader": {
+       "_id": "user-id-of-challenge-creator",
+       "profile": {
+         "name": "MyUserName"
+       }
+     },
+     "updatedAt": "timestamp,
+     "createdAt": "timestamp",
+     "_id": "challenge-id",
+     "prize": 0,
+     "memberCount": 1,
+     "tasksOrder": {
+       "rewards": [
+         "uuid-of-challenge-reward"
+       ],
+       "todos": [
+         "uuid-of-challenge-todo"
+       ],
+       "dailys": [
+         "uuid-of-challenge-daily"
+       ],
+       "habits": [
+         "uuid-of-challenge-habit"
+       ]
+     },
+     "official": false,
+     "id": "challenge-id"
+   }
+ }
+*/
+
+/**
+ * @apiDefine ChallengeArrayExample
+ * @apiSuccessExample {json} Sucessfull response with array of challenges
+ {
+   "data": [{
+     "group": {
+       "_id": "group-id-associated-with-challenge",
+       "name": "MyGroup",
+       "type": "guild",
+       "privacy": "public"
+     },
+     "name": "Long Detailed Name of Challenge",
+     "shortName": "my challenge",
+     "leader": {
+       "_id": "user-id-of-challenge-creator",
+       "profile": {
+         "name": "MyUserName"
+       }
+     },
+     "updatedAt": "timestamp,
+     "createdAt": "timestamp",
+     "_id": "challenge-id",
+     "prize": 0,
+     "memberCount": 1,
+     "tasksOrder": {
+       "rewards": [],
+       "todos": [],
+       "dailys": [],
+       "habits": []
+     },
+     "official": false,
+     "id": "challenge-id"
+   }]
+ }
+*/
+
+/**
  * @api {post} /api/v3/challenges Create a new challenge
  * @apiName CreateChallenge
  * @apiGroup Challenge
+ * @apiDescription Creates a challenge. Cannot create associated tasks with this route. See <a href="#api-Task-CreateChallengeTasks">CreateChallengeTasks</a>.
  *
- * @apiSuccess {Object} data The newly created challenge
+ * @apiParam (Body) {Object} challenge An object representing the challenge to be created
+ * @apiParam (Body) {UUID} challenge.groupId The id of the group to which the challenge belongs.
+ * @apiParam (Body) {String} challenge.name The full name of the challenge.
+ * @apiParam (Body) {String} challenge.shortName A shortened name for the challenge, to be used as a tag.
+ * @apiParam (Body) {Number} [challenge.prize=0] Number of gems offered as a prize to challenge winner. Default is 0.
+ *
+ * @apiSuccess {Object} challenge The newly created challenge.
+ * @apiUse SuccessfulChallengeRequest
+ *
+ * @apiUse ChallengeSuccessExample
+ *
+ * @apiError (401) {NotAuthorized} CantAffordPrize User does not have enough gems to offer this prize.
+ * @apiError (400) {BadRequest} ChallengeValidationFailed Invalid or missing parameter in challenge body.
  *
  * @apiUse GroupNotFound
+ * @apiUse UserNotFound
  */
 api.createChallenge = {
   method: 'POST',
@@ -120,11 +245,15 @@ api.createChallenge = {
  * @api {post} /api/v3/challenges/:challengeId/join Join a challenge
  * @apiName JoinChallenge
  * @apiGroup Challenge
- * @apiParam {UUID} challengeId The challenge _id
+ * @apiParam (Path) {UUID} challengeId The challenge _id
  *
- * @apiSuccess {Object} data The challenge the user joined
+ * @apiSuccess {Object} challenge The challenge the user joined
+ * @apiUse SuccessfulChallengeRequest
  *
  * @apiUse ChallengeNotFound
+ * @apiUse UserNotFound
+ *
+ * @apiUse ChallengeSuccessExample
  */
 api.joinChallenge = {
   method: 'POST',
@@ -168,11 +297,12 @@ api.joinChallenge = {
  * @api {post} /api/v3/challenges/:challengeId/leave Leave a challenge
  * @apiName LeaveChallenge
  * @apiGroup Challenge
- * @apiParam {UUID} challengeId The challenge _id
+ * @apiParam (Path) {UUID} challengeId The challenge _id
  *
  * @apiSuccess {Object} data An empty object
  *
  * @apiUse ChallengeNotFound
+ * @apiUse UserNotFound
  */
 api.leaveChallenge = {
   method: 'POST',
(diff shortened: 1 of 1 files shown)


website/server/controllers/api-v3/challenges.js
```diff
@@ -410,13 +547,18 @@ api.exportChallengeCsv = {
 };
 
 /**
- * @api {put} /api/v3/challenges/:challengeId Update a challenge
+ * @api {put} /api/v3/challenges/:challengeId Update a challenge. Only reflects
+ * changes to the full name of the challenge. Other challenge details cannot be
+ * updated through this route.
```
JTorr: With this route, I was only able to update the 'name' field of the challenge. Updates to 'shortName', 'group', and 'prize' failed silently. Let me know if this should be accepting updates to other fields.

crookedneighbor: That is the correct behavior. You can't change the prize, group or tag name after creation.

website/server/controllers/api-v3/challenges.js
```diff
@@ -26,13 +26,126 @@ let api = {};
...
+        "id": "challenge-id"
+        }]
+ }
+*/
+
+/**
  * @api {post} /api/v3/challenges Create a new challenge
  * @apiName CreateChallenge
  * @apiGroup Challenge
+ * @apiDescription Creates a challenge. Required params are `group`, `name` and
```
crookedneighbor: Params should be assumed to be required unless you mark them as optional with the `[]` notation. http://apidocjs.com/#param-api-param

website/server/controllers/api-v3/challenges.js
```diff
@@ -172,7 +289,13 @@ api.joinChallenge = {
  *
  * @apiSuccess {Object} data An empty object
  *
+ * @apiSuccessExample {json} Successfully left challenge
```
crookedneighbor: Hmm.... I don't think it's useful to provide an example for an empty object. Let's just nix these.

crookedneighbor (changes requested):

website/server/controllers/api-v3/challenges.js
JTorr: Good point. I removed this and the existing examples that were just empty objects.

JTorr:
@crookedneighbor Comments addressed.

website/server/controllers/api-v3/challenges.js
```diff
@@ -26,13 +26,125 @@ let api = {};
...
+        "id": "challenge-id"
+     }
+ }
+*/
+
+/**
+ * @apiDefine ChallengeArrayExample
+ * @apiSuccessExample {json} Sucessfull response with array of challenges
+ {
+    "data": [{
```
crookedneighbor: Let's go with `2` spaces instead of `4` for all of these examples.

website/server/controllers/api-v3/challenges.js
```diff
@@ -26,13 +26,125 @@ let api = {};
...
+        "memberCount": 1,
+        "tasksOrder": {
+            "rewards": [],
+            "todos": [],
+            "dailys": [],
+            "habits": []
+        },
+        "official": false,
+        "id": "challenge-id"
+        }]
```
crookedneighbor: Spacing is off here.

website/server/controllers/api-v3/challenges.js
```diff
@@ -26,13 +26,125 @@ let api = {};
...
+ }
+*/
+
+/**
  * @api {post} /api/v3/challenges Create a new challenge
  * @apiName CreateChallenge
  * @apiGroup Challenge
+ * @apiDescription Creates a challenge. Cannot create associated tasks with this route.
+ *
+ * @apiParam {Object} challenge Body parameter - An object representing the challenge to be created
```
crookedneighbor: Instead of `Body parameter -`, use the newly created param group (https://github.com/HabitRPG/habitica/commit/afb5b473a37325a82151bf48b0c7d54b4979e6eb)

``` js
* @apiParam (Body) {Object} challenge An object representing the challenge to be created
```

website/server/controllers/api-v3/challenges.js
```diff
@@ -26,13 +26,125 @@ let api = {};
...
+ * @apiParam {String} challenge.shortName A shortened name for the challenge, to be used as a tag.
+ * @apiParam {Number} [challenge.prize] Number of gems offered as a prize to challenge winner. Default is 0.
+ *
+ * @apiSuccess {Object} challenge The newly created challenge.
+ * @apiUse SuccessfulChallengeRequest
  *
- * @apiSuccess {Object} data The newly created challenge
+ * @apiUse ChallengeSuccessExample
+ *
+ * @apiError (401) {Unauthorized} CantAffordPrize User does not have enough gems to offer this prize.
```
crookedneighbor: `NotAuthorized`

website/server/controllers/api-v3/challenges.js
```diff
@@ -204,11 +321,20 @@ api.leaveChallenge = {
 };
 
 /**
- * @api {get} /api/v3/challenges/user Get challenges for a user
+ * @api {get} /api/v3/challenges/user Get challenges for a user.
  * @apiName GetUserChallenges
  * @apiGroup Challenge
+ * @apiDescription Get challenges that the user is a member, public challenges and the ones from the user's groups.
```
crookedneighbor: I find this phrasing a little awkward.

website/server/controllers/api-v3/challenges.js
```diff
@@ -204,11 +321,20 @@ api.leaveChallenge = {
...
  *
- * @apiSuccess {Array} data An array of challenges sorted with official challenges first, followed by the challenges in order from newest to oldest
+ * @apiSuccess {Object[]} challenges An array of challenges sorted with official challenges first, followed by the challenges in order from newest to oldest
+ * @apiSuccess {Object} challenges.challenge An object representing the details
```
crookedneighbor: I don't think this is necessary. (especially since this makes it look like the success looks like:

```
[
  {
    challenge: { /* challenge data */ },
  }, {
    challenge: { /* challenge data */ },
  }
]
```

When it's more like:

``` js
[
  {
    /* challenge data */
  }, {
    /* challenge data */
  }
]
```

website/server/controllers/api-v3/challenges.js
```diff
@@ -248,14 +374,17 @@ api.getUserChallenges = {
 
 /**
  * @api {get} /api/v3/challenges/group/:groupId Get challenges for a group
- * @apiDescription Get challenges that the user is a member, public challenges and the ones from the user's groups.
+ * @apiDescription Get challenges belonging to the given group.
  * @apiName GetGroupChallenges
  * @apiGroup Challenge
  *
  * @apiParam {UUID} groupId The group _id
```
crookedneighbor:
``` js
 * @apiParam (Path) {UUID} groupId The group id
```

website/server/controllers/api-v3/challenges.js
```diff
@@ -410,13 +541,17 @@ api.exportChallengeCsv = {
...
  * @apiName UpdateChallenge
  * @apiGroup Challenge
  *
  * @apiParam {UUID} challengeId The challenge _id
+ * @apiParam {String} [challenge.name] The full name of the challenge.
  *
  * @apiSuccess {Object} data The updated challenge
+ * @apiUse ChallengeSuccessExample
  *
  * @apiUse ChallengeNotFound
```
crookedneighbor, line 570: This should include an `@apiPermission` section since it's limited on what users can use this route.

website/server/controllers/api-v3/challenges.js
```diff
@@ -410,13 +541,17 @@ api.exportChallengeCsv = {
 };
 
 /**
- * @api {put} /api/v3/challenges/:challengeId Update a challenge
+ * @api {put} /api/v3/challenges/:challengeId Update a challenge. Only reflects
```
crookedneighbor: I don't think this is true. According to the group model, you should be able to update `name`, `description`, `tasksOrder`, `leader`. Is that not the case?

crookedneighbor (changes requested):
Overall, I think it's looking pretty good.

One thing we'll want to do throughout this file is add `(Body)` to the `@apiParam` body parameters, `(Path)` to the url path parameters and `(Query)` to the query params.

website/server/controllers/api-v3/challenges.js
JTorr: Fixed

website/server/controllers/api-v3/challenges.js
JTorr: Fixed

website/server/controllers/api-v3/challenges.js
JTorr: Fixed

website/server/controllers/api-v3/challenges.js
JTorr: Fixed

website/server/controllers/api-v3/challenges.js
JTorr: Fixed

website/server/controllers/api-v3/challenges.js
JTorr: Fixed

website/server/controllers/api-v3/challenges.js
JTorr: Fixed

website/server/controllers/api-v3/challenges.js
JTorr: Fixed

website/server/controllers/api-v3/challenges.js
JTorr: @crookedneighbor Good catch. I've added `description` and `leader`, but I'm having trouble with `tasksOrder`. I'm not sure if I should be sending a task id or an object that contains the attributes to create a new task. I've tried both with no success. I'm sure it's probably the format of the request, but since it fails silently (updating only the valid fields), I can't figure out what I'm missing. Would appreciate any help you can provide.

website/server/controllers/api-v3/challenges.js
JTorr: @crookedneighbor Added the `@apiPermission`. Also added a `MustBeChallengeLeader` error on line 560. LMK if I should leave that in, or if it's redundant.

website/server/controllers/api-v3/challenges.js
crookedneighbor: I might be wrong about task order. I'll have to look into it more.

JTorr:
@crookedneighbor I think the only thing left for this PR is to figure out whether tasksOrder allows updates. I haven't been able to successfully add tasks. Here's a Gist with what I'm seeing: https://gist.github.com/JTorr/cbb4f853005cc31d53a767f743143f7c

paglias:
@JTorr tasksOrders is in the `noSet` list so it cannot be updated nor set on creation

JTorr:
Thanks @paglias. I've fixed the params for the `update` route. I believe that should be the final requested change, but please lmk if I've missed anything.

website/server/controllers/api-v3/challenges.js
```diff
@@ -26,13 +26,125 @@ let api = {};
...
+         "name": "MyUserName"
+       }
+     },
+     "updatedAt": "timestamp,
+     "createdAt": "timestamp",
+     "_id": "challenge-id",
+     "prize": 0,
+     "memberCount": 1,
+     "tasksOrder": {
+       "rewards": [],
```
crookedneighbor: I think these should have dummy ids:

``` js
"tasksOrder": {
  "rewards": [
    "uuid-of-challenge-reward"
  ],
  "todos": [
     "uuid-of-challenge-todo"
  ],
  // etc
```

website/server/controllers/api-v3/challenges.js
```diff
@@ -26,13 +26,125 @@ let api = {};
...
+     "id": "challenge-id"
+   }]
+ }
+*/
+
+/**
  * @api {post} /api/v3/challenges Create a new challenge
  * @apiName CreateChallenge
  * @apiGroup Challenge
+ * @apiDescription Creates a challenge. Cannot create associated tasks with this route.
```
crookedneighbor: We should link to the documentation for the tasks creation route.

website/server/controllers/api-v3/challenges.js
```diff
@@ -26,13 +26,125 @@ let api = {};
...
+ *
+ * @apiParam (Body) {Object} challenge An object representing the challenge to be created
+ * @apiParam (Body) {String} challenge.groupId The id of the group to receive the challenge.
```
crookedneighbor: `to receive the challenge` doesn't sound quite right to me. Maybe `The id of the group to which the challenge belongs.`

 Also, the type should be `{UUID}` instead of `{STRING}`

website/server/controllers/api-v3/challenges.js
```diff
@@ -26,13 +26,125 @@ let api = {};
...
+ * @apiParam (Body) {String} challenge.name The full name of the challenge.
+ * @apiParam (Body) {String} challenge.shortName A shortened name for the challenge, to be used as a tag.
+ * @apiParam (Body) {Number} [challenge.prize] Number of gems offered as a prize to challenge winner. Default is 0.
```
crookedneighbor: Use the APIDoc default syntax here instead: `[challenge.prize=0]`

website/server/controllers/api-v3/challenges.js
```diff
@@ -26,13 +26,125 @@ let api = {};
...
+ *
+ * @apiSuccess {Object} challenge The newly created challenge.
+ * @apiUse SuccessfulChallengeRequest
  *
- * @apiSuccess {Object} data The newly created challenge
+ * @apiUse ChallengeSuccessExample
+ *
+ * @apiError (401) {Not Authorized} CantAffordPrize User does not have enough gems to offer this prize.
```
crookedneighbor: `Not Authorized` -> `NotAuthorized`

website/server/controllers/api-v3/challenges.js
```diff
@@ -410,15 +539,25 @@ api.exportChallengeCsv = {
...
  * @apiGroup Challenge
  *
- * @apiParam {UUID} challengeId The challenge _id
+ * @apiParam (Path) {UUID} challengeId The challenge _id
+ * @apiParam (Body) {String} [challenge.name] The new full name of the challenge.
+ * @apiParam (Body) {String} [challenge.description] The new challenge description.
+ * @apiParam (Body) {String} [challenge.leader] The UUID of the new challenge leader.
  *
  * @apiSuccess {Object} data The updated challenge
+ * @apiPermission ChallengeLeader
```
crookedneighbor, line 566: Is this defined somewhere? I don't see it in this file.

website/server/controllers/api-v3/challenges.js
JTorr: Updated

website/server/controllers/api-v3/challenges.js
JTorr: Fixed

JTorr:
@crookedneighbor All new comments addressed.

JTorr:
@crookedneighbor Anything else I need to do on this before merge?

crookedneighbor:
I was looking at it this morning and spotted a few things. 

The top level group object is missing from the challenges response, which orphans the group properties from the rest of it. 

You're missing description and official in the create challenge route body params. 

The prize parameter in create challenge has text about the default value, but that's redundant. Apidoc already takes care of that by providing the default.

TheHollidayInn:
@JTorr Are you still working on this?

website/server/controllers/api-v3/challenges.js
```diff
@@ -168,11 +297,12 @@ api.joinChallenge = {
  * @api {post} /api/v3/challenges/:challengeId/leave Leave a challenge
  * @apiName LeaveChallenge
  * @apiGroup Challenge
- * @apiParam {UUID} challengeId The challenge _id
+ * @apiParam (Path) {UUID} challengeId The challenge _id
  *
  * @apiSuccess {Object} data An empty object
  *
  * @apiUse ChallengeNotFound
+ * @apiUse UserNotFound
```
Alys, line 305: The `keep` parameter should be described in this comment block.

`let keep = req.body.keep === 'remove-all' ? 'remove-all' : 'keep-all';`

Alys:
@JTorr Would you like to keep working on this or would you prefer someone else to take it over?

Alys:
@JTorr I'm closing this however if you'd like to keep working on it before someone else takes it over, you'd be welcome to add commits and we'll reopen it.

If someone else takes it up and uses part of your work, and their PR is merged, we'll give you partial credit towards a tier.

Alys:
Partial credit towards a tier was given as described at https://github.com/HabitRPG/habitica/pull/8626#issuecomment-301204946

PR #8117, opened 2016-10-10
fix(api): Allow x-client to be set in cors middleware

crookedneighbor:
Allow 3rd party extensions in browsers to set `x-client` header.

Currently it seems like protocols that don't use CORS (such as the mobile apps) can set the `x-client` header, but browser based extensions cannot.

@paglias Is there a reason not to do this?


paglias:
sounds good to me

crookedneighbor:
:+1: I'll wait for the build to finish and merge it in then.

crookedneighbor:
Whoops. Forgot a test. Fixing

PR #8120, opened 2016-10-11
Api doc dataexport

librarianmage:
#8087
### Changes

Improves documentation of Data Export functions

---

UUID:

### DIFF
--- a/website/server/controllers/top-level/dataexport.js
+++ b/website/server/controllers/top-level/dataexport.js
@@ -30,6 +30,10 @@ let api = {};
  */
 
 /**
+ * @apiDefine pathParam Path Parameters
+ */
+
+/**
  * @api {get} /export/history.csv Export user tasks history in CSV format
  * @apiDescription History is only available for habits and dailies so todos and rewards won't be included. NOTE: Part of the private API that may change at any time.
  * @apiName ExportUserHistory
@@ -152,7 +156,7 @@ api.exportUserDataJson = {
  * @apiGroup DataExport
  * @apiDescription NOTE: Part of the private API that may change at any time.
  *
- * @apiSuccess {String} data A xml file
+ * @apiSuccess {String} data A xml file. See /export/userdata.json for structure info.
  */
 api.exportUserDataXml = {
   method: 'GET',
@@ -175,7 +179,9 @@ api.exportUserDataXml = {
  * @apiGroup DataExport
  * @apiDescription NOTE: Part of the private API that may change at any time.
  *
- * @apiSuccess {String} An html page
+ * @apiParam (pathParam) {string} uuid The User ID of the user
+ *
+ * @apiSuccess {String} data An html page
  *
  * @apiUse UserNotFound
  */
@@ -209,7 +215,9 @@ api.exportUserAvatarHtml = {
  * @apiGroup DataExport
  * @apiDescription NOTE: Part of the private API that may change at any time.
  *
- * @apiSuccess {String} A png file
+ * @apiParam (pathParam) {string} uuid The User ID of the user
+ *
+ * @apiSuccess {String} image A png file
  */
 api.exportUserAvatarPng = {
   method: 'GET',
--- a/website/server/controllers/top-level/dataexport.js
+++ b/website/server/controllers/top-level/dataexport.js
@@ -25,12 +25,23 @@ const BASE_URL = nconf.get('BASE_URL');
 let api = {};
 
 /**
+ * @apiDefine authError
+ * @apiError NotAuthorized There was an error in authorizing the User ID / API Token pair. Check the <code>message</code> for more information.
+ */
+
+/**
  * @api {get} /export/history.csv Export user tasks history in CSV format
  * @apiDescription History is only available for habits and dailies so todos and rewards won't be included. NOTE: Part of the private API that may change at any time.
  * @apiName ExportUserHistory
  * @apiGroup DataExport
  *
- * @apiSuccess {String} A csv file
+ * @apiSuccess {String} returns A csv file
+ *
+ * @apiSuccessExample {csv} Example
+ * Task Name,Task ID,Task Type,Date,Value
+ * ...
+ *
+ * @apiUse authError
  */
 api.exportUserHistory = {
   method: 'GET',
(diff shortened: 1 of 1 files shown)


website/server/controllers/top-level/dataexport.js
```diff
@@ -25,12 +25,27 @@ const BASE_URL = nconf.get('BASE_URL');
 let api = {};
 
 /**
+ * @apiDefine authError
```
crookedneighbor: Let's not use this in the documentation for each route. We should instead have some info in the header of the API doc about authorization requirements. (not for this PR)

website/server/controllers/top-level/dataexport.js
```diff
@@ -25,12 +25,27 @@ const BASE_URL = nconf.get('BASE_URL');
...
+ * @apiError NotAuthorized There was an error in authorizing the User ID / API Token pair. Check the <code>message</code> for more information.
+ */
+
+/**
+ * @apiDefine pathParam Path Parameters
```
crookedneighbor: This is defined here, so you can remove it: https://github.com/HabitRPG/habitica/commit/afb5b473a37325a82151bf48b0c7d54b4979e6eb

website/server/controllers/top-level/dataexport.js
```diff
@@ -25,12 +25,27 @@ const BASE_URL = nconf.get('BASE_URL');
...
  * @apiDescription History is only available for habits and dailies so todos and rewards won't be included. NOTE: Part of the private API that may change at any time.
  * @apiName ExportUserHistory
  * @apiGroup DataExport
  *
- * @apiSuccess {String} A csv file
+ * @apiSuccess {String} returns A csv file
+ *
+ * @apiSuccessExample {csv} Example
+ * Task Name,Task ID,Task Type,Date,Value
+ * ...
```
crookedneighbor, line 39: Provide some sample data here.

website/server/controllers/top-level/dataexport.js
```diff
@@ -97,7 +112,26 @@ async function _getUserDataForExport (user) {
  * @apiGroup DataExport
  * @apiDescription NOTE: Part of the private API that may change at any time.
  *
- * @apiSuccess {String} A json file
```
crookedneighbor, line 100: This returns a json file, not an object.

crookedneighbor: It should probably stat that it's a json file of the user object and link to the get user route.

website/server/controllers/top-level/dataexport.js
```diff
@@ -122,7 +156,7 @@ api.exportUserDataJson = {
  * @apiGroup DataExport
  * @apiDescription NOTE: Part of the private API that may change at any time.
  *
- * @apiSuccess {String} A xml file
+ * @apiSuccess {String} data A xml file. See /export/userdata.json for structure info.
```
crookedneighbor: `a xml` -> `an xml`

It should probably stat that it's an xml file of the user object and link to the get user route.

website/server/controllers/top-level/dataexport.js
```diff
@@ -145,7 +179,9 @@ api.exportUserDataXml = {
  * @apiGroup DataExport
  * @apiDescription NOTE: Part of the private API that may change at any time.
  *
- * @apiSuccess {String} An html page
+ * @apiParam (pathParam) {string} uuid The User ID of the user
```
crookedneighbor: `(Path) {String}`

website/server/controllers/top-level/dataexport.js
crookedneighbor: Also, I elected to name the group `Path` instead of `pathParam`

website/server/controllers/top-level/dataexport.js
```diff
@@ -179,7 +215,9 @@ api.exportUserAvatarHtml = {
  * @apiGroup DataExport
  * @apiDescription NOTE: Part of the private API that may change at any time.
  *
- * @apiSuccess {String} A png file
+ * @apiParam (pathParam) {string} uuid The User ID of the user
+ *
+ * @apiSuccess {String} image A png file
```
crookedneighbor: It's a PNG file, not a `{String}`. Also, specify that it is the User's avatar

website/server/controllers/top-level/dataexport.js
```diff
@@ -179,7 +215,9 @@ api.exportUserAvatarHtml = {
  * @apiGroup DataExport
  * @apiDescription NOTE: Part of the private API that may change at any time.
  *
- * @apiSuccess {String} A png file
+ * @apiParam (pathParam) {string} uuid The User ID of the user
```
crookedneighbor: `{String}`

crookedneighbor (changes requested):

website/server/controllers/top-level/dataexport.js
librarianmage: @crookedneighbor That might be a good idea, but it also adds tasks organized by type, which is not part of the get user route

crookedneighbor: That's true. For this route (and the XML one) I'm comfortable with just describing it and not including extensive information about it.

librarianmage:
@crookedneighbor Made the changes!

website/server/controllers/top-level/dataexport.js
librarianmage: @crookedneighbor Got it. I have included the link in both (although get user doesn't currently have info)

website/server/controllers/top-level/dataexport.js
```diff
@@ -122,7 +128,7 @@ api.exportUserDataJson = {
  * @apiGroup DataExport
  * @apiDescription NOTE: Part of the private API that may change at any time.
  *
- * @apiSuccess {String} A xml file
+ * @apiSuccess {XML} returns An xml file of the user object. See <code><a href='#api-User-UserGet'>GET /user</a></code>.
```
crookedneighbor, line 131: For future reference, markdown is preferred here.

crookedneighbor:
I made a few adjustments and merged this in. :+1:

librarianmage:
@crookedneighbor Awesome! Did you notate it?

crookedneighbor:
Thanks for the reminder. Just did.

librarianmage:
Thank you 😊!

PR #8135, opened 2016-10-13
Remove cookies on clearing browser data

### LINKED ISSUE #6490: Clear cookies in Clear Browser Data page

librarianmage:
Fixes #6490 
### Changes

_Should_ remove cookies when clearing local storage (seemed to work during testing)

---

UUID: [Leave Cookie Remove](https://map.what3words.com/leave.cookie.remove)


website/client-old/js/controllers/authCtrl.js
```diff
@@ -11,6 +11,7 @@ angular.module('habitrpg')
 
       $scope.logout = function() {
         localStorage.clear();
+        window.document.cookie.split(";").forEach(function(c) { window.document.cookie = c.replace(/^ +/, "").replace(/=.*/, "=;expires=" + new Date().toUTCString() + ";path=/"); });
```
crookedneighbor: This doesn't look right to me.

crookedneighbor: Can you provide documentation about where you got this snippet? Overwriting `window.focument.cookie` over and over again seems wrong to me.

librarianmage: @crookedneighbor Can't find where I found it, but looking at it again, it does seem off. I will look and see if I can find a different one.

librarianmage:
@crookedneighbor Should be fixed now, I used http://www.quirksmode.org/js/cookies.html as the basis of the cookie removal.

website/client-old/js/controllers/authCtrl.js
```diff
@@ -11,6 +11,13 @@ angular.module('habitrpg')
 
       $scope.logout = function() {
         localStorage.clear();
+
+        var cookies = document.cookie.split(';');
```
crookedneighbor, line 17: Please include the link where you got this solution as a comment.

website/client-old/js/controllers/authCtrl.js
```diff
@@ -11,6 +11,13 @@ angular.module('habitrpg')
...
+        for (var i = 0; i < cookies.length; i++) {
+          var cookieName = cookies[i].match(/([^=]+)(?=\=)/)[0];
+          document.cookie = cookieName + '= ; expires=Thu Jan 1 00:00:00 1970 GMT; path=/' +
```
crookedneighbor: I still don't see how this is correct. 

1) You're just setting `document.cookie` over and over again
2) You have a trailing `+`

librarianmage: Oops, the trailing `+` should not be there, [but](http://www.quirksmode.org/js/cookies.html) [every](http://www.java2s.com/Tutorial/JavaScript/0080__Development/Deleteallcookies.htm) [solution](http://stackoverflow.com/questions/179355/clearing-all-cookies-with-javascript) [I](http://stackoverflow.com/questions/595228/how-can-i-delete-all-cookies-with-javascript) [have](http://stackoverflow.com/questions/28804564/delete-all-cookies-of-browser-using-javascript) [found](https://www.reddit.com/r/javascript/comments/2vafdu/how_to_remove_cookies_with_javascript/) either only clears a cookie (meaning it would need to be looped) or has a loop in it already.

crookedneighbor:
TIL `document.cookie = whatever` sets a new cookie. https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie

What a dumb API.

librarianmage:
I have removed the trailing `+` and included the link I adapted it from.

crookedneighbor (approved):
I'm still not convinced this is necessary, because the only cookies that Habitica uses directly are the session cookies, which are not accessible via the browser's JavaScript. They should already be cleared here on the server: https://github.com/HabitRPG/habitica/blob/a4fd687510160fc54e5730266f4f921f63f49c53/website/server/controllers/top-level/auth.js#L10-L11

@Alys what cookies are you trying to get clear?

librarianmage:
Perhaps in case the other cookies haven't been properly removed, such as:
- __atuvc
- _atshc
- _ga
- amplitude_idhabitica.com
- session-set

crookedneighbor:
Those shouldn't affect the user's experience though. I'm just saying I'm pretty ambivalent about adding this code.

Alys:
@crookedneighbor My reasons are here: https://github.com/HabitRPG/habitica/issues/6490#issuecomment-249354073

website/client-old/js/controllers/authCtrl.js
```diff
@@ -11,6 +11,14 @@ angular.module('habitrpg')
 
       $scope.logout = function() {
         localStorage.clear();
+
+        // Adapted from http://www.quirksmode.org/js/cookies.html
+        var cookies = document.cookie.split(';');
+        for (var i = 0; i < cookies.length; i++) {
+          var cookieName = cookies[i].match(/([^=]+)(?=\=)/);
```
Alys: If this line is meant to return only the name of the cookie, then it's not working as desired; it returns an array. You can see what's in it by putting `console.log(cookieName);` just after this line. `match` returns an array, so perhaps you want the first element of that array?

crookedneighbor: The match method will also return null if a match could not be found, so you'll need to write a condition for that.

librarianmage: @crookedneighbor @Alys Those should be fixed

Alys: @MathWhiz That's working now, thanks! Please rename the variable `cookieName` since it contains more than the name.

website/client-old/js/controllers/authCtrl.js
```diff
@@ -11,6 +11,16 @@ angular.module('habitrpg')
 
       $scope.logout = function() {
         localStorage.clear();
+
```
Alys, line 14: Add another comment here saying that this deletes cookies, except for those that have the HttpOnly flag set.

Since this code is client side, it can't delete HttpOnly cookies, but we'll just have to live with that.

website/client-old/js/controllers/authCtrl.js
```diff
@@ -11,6 +11,16 @@ angular.module('habitrpg')
...
+        var cookies = document.cookie.split(';');
+        for (var i = 0; i < cookies.length; i++) {
+          var cookieName = cookies[i].match(/([^=]+)(?=\=)/);
+          if(cookieName != null){
+            document.cookie = cookieName[0] + '= ; expires=Thu Jan 1 00:00:00 1970 GMT; path=/';
+          }
+        }
+
         $window.location.href = '/logout';
       };
```
Alys, line 27: Please also edit the texts that are on the clear-browser-data page. 

Change this: "... click the button below to clear local storage for this website ..." 
to this: "... click the button below to clear local storage and most cookies for this website ..." 

Change the button to say "Clear Data" and its on-hover text to say "This button will clear local storage and most cookies, and log you out." (note the new full stop at the end).

The pop-up message that appears after clicking the button should be changed to:

"Clearing Data"

"Habitica's stored data is being cleared from your browser. You will be logged out and redirected to the home page. Please wait."

Show us screenshots of what the page and the pop-up message look like after that, and then it should be good to merge!

Alys (changes requested):

librarianmage:
@Alys Everything should be updated, working on picture.

librarianmage:
screenshot 2016-10-24 at 5 42 54 pm
screenshot 2016-10-24 at 5 42 51 pm
screenshot 2016-10-24 at 5 42 46 pm

@Alys Got the images!

Alys (approved):

Alys:
Thanks @MathWhiz, that's good! I've noted it towards you next tier.

Unrelated to that, but noting it here for public record, I've awarded you the Comrade title for your [Inactive Guild Leader Script](https://gist.github.com/MathWhiz/bdcdbf7122d2b8af4d0af8f2fa3c4d48), used to create the [Guilds with Inactive Leaders: Upcoming](http://habitica.wikia.com/wiki/The_Keep:Pirate_Cove/Guilds_with_Inactive_Leaders:_Upcoming) wiki page, to assist the Pirates in helping orphaned guilds find new leaders. If we ever make a decision to award tiers for Pirate activities, the script will also count towards a tier for you.

PR #8138, opened 2016-10-13
Add special spells to Seasonal Shop API

SabreCat:
### Changes
- Adds the ability to list special seasonal transformation spells to be returned via the Seasonal Shop list API.
- If the `purchase` route is called with a `type` of `spells`, call the `buySpecialSpell` op instead of `purchase`.

---

UUID: [Purchase Special Spell](https://map.what3words.com/purchase.special.spell)

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "less-loader": "^2.2.3",
     "lodash": "^3.10.1",
     "lodash.setwith": "^4.2.0",
+    "lodash.pickby": "^4.2.0",
     "merge-stream": "^1.0.0",
     "method-override": "^2.3.5",
     "moment": "^2.13.0",
--- a/website/common/script/libs/shops.js
+++ b/website/common/script/libs/shops.js
@@ -209,12 +209,12 @@ shops.getTimeTravelersCategories = function getTimeTravelersCategories (user, la
   return categories;
 };
 
-// To switch seasons/available inventory, edit the availableSets object to whatever should be sold.
-// let availableSets = {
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -70,6 +70,7 @@
     "less-loader": "^2.2.3",
     "lodash": "^3.10.1",
     "lodash.setwith": "^4.2.0",
+    "lodash.pickby": "^4.2.0",
```
crookedneighbor, line 73: Is there a reason to use the v4 version?

website/common/script/libs/shops.js
```diff
@@ -224,10 +225,40 @@ shops.getSeasonalShopCategories = function getSeasonalShopCategories (user, lang
     fall2015Warrior: i18n.t('scarecrowWarriorSet', language),
   };
 
+  let availableSpells = [
```
crookedneighbor: Should this be a const

``` js
const AVAILABLE_SPELLS = [
```

package.json
SabreCat: It's only been available since v4: https://lodash.com/docs/4.16.4#pickBy

website/common/script/libs/shops.js
SabreCat: Can do! I was following the pattern established by the existing `availableSets`, but that works too.

SabreCat: Actually, I can change `availableSets` to a constant too, if that'd make sense.

PR #8139, opened 2016-10-14
Cleaned up some CSS

bcpletcher:
Fixes put_issue_url_here
### Changes

---

UUID:


crookedneighbor:
What's your User ID? You can find it in [Settings > API](https://habitrpg.com/#/options/settings/api). Be sure to only copy the User ID and _not_ your API Token.

PR #8150, opened 2016-10-19
Send a message to the party chat when a quest is aborted, fixes #4879

### LINKED ISSUE #4879: Aborting a Quest Should Add System Message to Party Chat
Currently, when a quest is aborted, no message is put in party chat. This means that there is no explanation for where a quest went to the rest of the party, which can be rather confusing. When a quest is completed, there is a message added. When a quest is aborted, a system message should be posted, saying:

`Alys aborted the party quest The Spirit of Spring.`

or something similar.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9685689-aborting-a-quest-should-add-system-message-to-party-chat?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

abatula:
Fixes #4879
### Changes

When a user aborts an active quest, print a message to the party. Added checks to the test for aborting a quest to check that the message is sent .

questabort

---


website/server/controllers/api-v3/quests.js
```diff
@@ -412,12 +412,16 @@ api.abortQuest = {
...
-    let group = await Group.getGroup({user, groupId, fields: 'type quest leader'});
+    let group = await Group.getGroup({user, groupId, fields: 'type quest leader chat'});
+
     if (!group) throw new NotFound(res.t('groupNotFound'));
     if (group.type !== 'party') throw new NotAuthorized(res.t('guildQuestsNotSupported'));
     if (!group.quest.active) throw new NotFound(res.t('noActiveQuestToAbort'));
     if (user._id !== group.leader && user._id !== group.quest.leader) throw new NotAuthorized(res.t('onlyLeaderAbortQuest'));
 
+    let questName = questScrolls[group.quest.key].text('en');
+    group.sendChat(`\`${user.profile.name} aborted the party quest ${questName}.\``);
```
crookedneighbor, line 423: We should probably have tests around this behavior.

crookedneighbor (changes requested):

abatula:
Added two test cases to check that party message is sent.

abatula:
I'm stumped on why the check is failing. 

It looks like the failing check is a line in a [test for checking a cast spell](https://github.com/HabitRPG/habitica/blob/f1bb2db73b5dc8ff54114662e9a86647f3bba320/test/api/v3/integration/user/POST-user_class_cast_spellId.test.js#L159). I don't know why this becomes a problem, does something about creating a stub for the chat function affect later files?

I also tried using a similar check to the previously mentioned function:
`expect(questingGroup.chat[0]).to.exist;
  expect(questingGroup.chat[0].uuid).to.equal('system');`
But this causes an issue with coupon codes and payment, as shown [here.](https://gist.github.com/abatula/a364a13386d3cd8098dd3ad699bd776d)

I can't see how any of these issues are related to my test code.

abatula:
Fixed the issue with the test, just needed to restore the `sendChat` function after testing.

abatula:
Are there any other changes needed for this pull request before it can be merged?

Alys:
I've tested this locally and confirmed that the message is inserted in party chat as desired, and that the test works correctly (i.e., changing the message that the test expects to see causes the test to fail; setting it to the expected value allows the test to pass). 

I'm not good enough with our code base to properly review the quality of the new code but it looks okay to me. I'll merge this PR in about four days if there's no other comments. @abatula Feel free to post a reminder here if none of us have merged or made comments after five days.

abatula:
Checking in to see if there are any requested changes to this pull request, or if it's ready to go.

Alys:
@abatula Thanks for the reminder. I was overloaded with real-live work the past couple of days. I'll merge it later today since no one else has commented.

Alys:
@abatula Thank you! I've noted this as credit towards your next tier.

PR #8166, opened 2016-10-25
Moderator edit challenge fixes #8149

### LINKED ISSUE #8149: moderators can't edit challenge tasks
This issue relates to editing the tasks belonging to a challenge itself (i.e., in the challenge's own page under Social > Challenge). It does not relate to editing any copies of challenge tasks that are added to any player's own list when the player joins a challenge.

Moderators must be able to edit a challenge to change or add or delete any of the challenge's tasks, in the same way that the owner of the challenge can.

Currently, moderators can't do this - they receive the error "Tasks belonging to a challenge can only be edited by the leader."

A moderator is defined as anyone with `contributor.admin` set to true.

An easy way to find the code that needs to be changed for this is to search for `onlyChalLeaderEditTasks`. You'll find code like this:
https://github.com/HabitRPG/habitica/blob/1eb1fe76a8917cf5cd67d317fb91f577e7470d51/website/server/controllers/api-v3/tasks.js#L1344

Instead of using a leader ID check, the `canModify` method could be used:
https://github.com/HabitRPG/habitica/blob/1eb1fe76a8917cf5cd67d317fb91f577e7470d51/website/server/models/challenge.js#L75-L77

Note there's no need to change the `onlyChalLeaderEditTasks` string to make it say that mods can edit too. It's best to leave it as it is now for simplicity. Mods don't edit challenge tasks often enough for it to matter that the message is a little inaccurate and we'd typically inform the challenge owner that we've done it.

_FYI for anyone curious: Moderators occasionally need to edit challenges if the challenge owner has put a swear word or other inappropriate content into one of the challenge's tasks, or under other rare special circumstances. Under normal circumstances, they do not edit other player's challenges._

tribblx:
https://github.com/HabitRPG/habitica/issues/8149
### Changes

Changes made to allow moderators to edit and delete challenge tasks.

Also updated Unit tests to programmatically test this ability. 

---

UUID: 
75c44842-dac9-4f08-bf22-98c6fa943ca8

### DIFF
--- a/test/api/v3/integration/tasks/challenges/PUT-tasks_challenge_challengeId_tasksId_checklist_itemId.test.js
+++ b/test/api/v3/integration/tasks/challenges/PUT-tasks_challenge_challengeId_tasksId_checklist_itemId.test.js
@@ -8,11 +8,13 @@ import { v4 as generateUUID } from 'uuid';
 
 describe('PUT /tasks/:taskId/checklist/:itemId', () => {
   let user;
+  let moderator;
   let guild;
   let challenge;
 
   before(async () => {
     user = await generateUser();
+    moderator = await generateUser({ 'contributor.admin': true });
     guild = await generateGroup(user);
     challenge = await generateChallenge(user, guild);
   });
@@ -82,6 +84,29 @@ describe('PUT /tasks/:taskId/checklist/:itemId', () => {
     expect(savedTask.checklist[0].completed).to.equal(true);
     expect(savedTask.checklist[0].id).to.not.equal('123');
   });
+  // TODO
+  it('moderator updates a checklist item on dailies', async () => {
+    let task = await user.post(`/tasks/challenge/${challenge._id}`, {
+      type: 'daily',
+      text: 'Daily with checklist',
+    });
+
+    let savedTask = await user.post(`/tasks/${task._id}/checklist`, {
+      text: 'Checklist Item 1',
+      completed: false,
+    });
+
+    savedTask = await moderator.put(`/tasks/${task._id}/checklist/${savedTask.checklist[0].id}`, {
+      text: 'updated',
+      completed: true,
+      _id: 123, // ignored
+    });
+
+    expect(savedTask.checklist.length).to.equal(1);
+    expect(savedTask.checklist[0].text).to.equal('updated');
+    expect(savedTask.checklist[0].completed).to.equal(true);
+    expect(savedTask.checklist[0].id).to.not.equal('123');
+  });
 
   it('updates a checklist item on todos', async () => {
     let task = await user.post(`/tasks/challenge/${challenge._id}`, {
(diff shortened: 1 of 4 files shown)


test/api/v3/integration/tasks/challenges/PUT-tasks_challenge_challengeId_tasksId_checklist_itemId.test.js
```diff
@@ -82,6 +84,29 @@ describe('PUT /tasks/:taskId/checklist/:itemId', () => {
     expect(savedTask.checklist[0].completed).to.equal(true);
     expect(savedTask.checklist[0].id).to.not.equal('123');
   });
+  // TODO
```
TheHollidayInn, line 87: What is the todo message for?

website/views/shared/tasks/meta_controls.jade
```diff
@@ -45,7 +45,7 @@
         span.glyphicon.glyphicon-bullhorn(tooltip=env.t('challenge'))
         | &nbsp;
     // delete
-    a(ng-if='!task.challenge.id || obj.leader._id === User.user._id', ng-click='removeTask(task, obj)', tooltip=env.t('delete'))
+    a(ng-if='!task.challenge.id || obj.leader._id === User.user._id || user.contributor.admin', ng-click='removeTask(task, obj)', tooltip=env.t('delete'))
```
TheHollidayInn, line 48: Not a big deal, but this is probably more clear if moved to a function in the controller called `canEditTask`.

TheHollidayInn:
Do we have integration tests for editing and deleting the challenge tasks? It looks to me that you only wrote tests for the checklists.

test/api/v3/integration/tasks/challenges/PUT-tasks_challenge_challengeId_tasksId_checklist_itemId.test.js
tribblx: that's ...just a dumb mistake - - forgot to remove that

website/views/shared/tasks/meta_controls.jade
tribblx: as it turns out, I'm not all that familiar with how angular is laid out - and I'm definitely not very familiar with this code base yet.  The closest I think I've come is client-old/js/controllers/tasksCtrl.js

Could you point me in the right direction here, I'm a little lost?

tribblx:
I didn't see existing tests for the challenge tasks - - and also wasn't quite sure about the semantic difference between the two?  Maybe you could shed some light on that for me.

TheHollidayInn:
Between a checklist item and a task? A checklist item is a subdoc (Mongo term) of a Task doc. 

Here are the tests for deleting a challenge task : https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/tasks/challenges/DELETE-tasks_id_challenge_challengeId.test.js .

Here is for updating: 
https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/tasks/challenges/PUT-tasks_challenge_challengeId.test.js

You have actually added five new features
1. Mod can edit challenge tasks
2. Mod can delete challenge tasks
3. Mod can add checklist to challenge task
4. Mod can edit checklist on challenge tasks
5. Mod can delete checklist item on challenge tasks

So, you will want _at least_ 2 tests for each feature. 1 for success and 1 for failure.
For example:
1. Fails to edit challenge when user is not leader or mod
2. Edits challenge when user is mod

Sorry if that is too much detail. Also, we should probably abstract these access checks to a library. And add unit tests. @paglias , thoughts? (Not in this PR though).

paglias:
Yeah in future we may want to move the checks to a library if we keep adding things to check, it's easy to forget something

Alys:
I haven't had time to look at the code yet, but these three features sound
like they shouldn't be there:
1. Mod can add checklist to challenge task
2. Mod can edit checklist on challenge tasks
3. Mod can delete checklist item on challenge tasks

I suggest not writing any test or more code for them until we've confirmed
we want them. Challenge owners can't add checklists to the challenge's own
tasks, so mods shouldn't be able to.

tribblx:
Yeah sorry about that - - I wasn't quite catching the semantic differences between tasks and checklists.. I can change that easily to not allow moderators to update checklists.

SabreCat:
Looks like the changes made here substantially missed the scope of the issue, and there hasn't been any activity on the PR in some time. I'm going to close this for now and reopen the issue ticket for help. @btribble, if you'd still like to tackle this, feel free to submit a new PR!

PR #8168, opened 2016-10-25
New A/B test: Delay Damage from Dailies

SabreCat:
### Changes
- Completes the preceding A/B test, enabling the Armoire from the beginning for better Day 4+ retention.
- Begins a new A/B test: is retention improved if Dailies don't do any damage until level 4?
- In order to implement client-side behavior for this test, the `_ABtest` property is no longer hidden from the client, and a new flag `dailyDamageEnabled` was implemented. If this functionality doesn't end up being beneficial for retention, these modifications can be rolled back.

---

UUID: [Daily Damage Test](https://map.what3words.com/daily.damage.test)


paglias:
If we rollback the new flag will it create problems with mobile apps?

SabreCat:
Closing this for now. We will attempt this test again at a later date.

PR #8170, opened 2016-10-26
Allow user to save the checklist collapse and alias properties of a challenge task

### LINKED ISSUE #7875: Checklists on challenge tasks always expand when site syncs
General Info
- UUID: 8389891a-ba52-4580-b08b-2cb706c7e0b4
- Browser: Chrome
- OS: Windows 10
### Description

If you add checklists to a challenge task and click on expand/collapse to keep them hidden, they always expand back when the site syncs.

Hus274:
Fixes #7875 
### Changes

This change allows the user to save an alias or the checklist collapse properties of a challenge task. Prior to this fix, if a user assigned an alias or collapsed a checklist on a challenge task, the data would be lost on a sync or refresh. After this update, a change to either (collapsing a checklist or creating an alias) will be saved to the database and preserved when the user refreshes.

I made this change for both alias and checklist collapse because the alias was listed in the existing test and just wasn't being tested for and is documented [here](http://habitica.wikia.com/wiki/Challenges#Challenge_Participant.27s_Permissions) that it should be modifiable. I've updated the test to make sure both these fields are save-able by the user.

### DIFF
--- a/test/api/v3/integration/tasks/PUT-tasks_id.test.js
+++ b/test/api/v3/integration/tasks/PUT-tasks_id.test.js
@@ -74,6 +74,7 @@ describe('PUT /tasks/:id', () => {
         checklist: [
           {text: 123, completed: false},
         ],
+        collapseChecklist: false,
       });
       await sleep(2);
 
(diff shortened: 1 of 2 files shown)


crookedneighbor:
Looks right to me. 👍

crookedneighbor (approved):

crookedneighbor:

PR #8173, opened 2016-10-27
transferGems: use receiver language translation for PM strings #7722

### LINKED ISSUE #7722: Gem Receiving text is in native language of sender

lynxlynxlynx:
Fixes #7722 

If a preferred language isn't set, it will automatically fall back to the default 'en' locale.

UUID: 10c530d0-a796-476b-8a31-6f048c6e8a8c


crookedneighbor (changes requested):
One problem with this is that this will send the message using the receiver's language to both the sender and the receiver. I've opened a PR to your branch that adds a test (https://github.com/lynxlynxlynx/habitrpg/pull/1) to illustrate the problem.

Let me know if you need help implementing a solution.

lynxlynxlynx:
I'll take a look. Judging from the strings, it'd be odd if the sender received the same message, but on the other hand I don't doubt your tests (haven't read them yet).

> "privateMessageGiftIntro": "Hello <%= receiverName %>, <%= senderName %> has sent you",
> "privateMessageGiftGemsMessage": "<%= gemAmount %> gems!",

Edit: Ah, I see, the carbon copy for the sent archive.

lynxlynxlynx:
Ok, I think the best approach is to change sendMessage to take an additional param, rather than hacking one of the inboxes after the fact. That's what the last commit does. I'll go change the caller in question now.

Looking at other users, there are two candidates that could benefit from a similar change in payments.js (gifted sub, gifted (paid?) gems), but they currently use hardcoded English strings, so more work is needed first.

website/server/models/user/methods.js
```diff
@@ -20,10 +20,11 @@ schema.methods.getGroups = function getUserGroups () {
   return userGroups;
 };
 
-schema.methods.sendMessage = async function sendMessage (userToReceiveMessage, message) {
+schema.methods.sendMessage = async function sendMessage (userToReceiveMessage, translatedMessage, message) {
```
crookedneighbor: I think this should be a two argument method. The first argument being the user and the second to be an options object with two properties: receiverMsg and senderMsg. If senderMsg is undefined, set it to receiverMsg. 

``` is
let senderMsg = options.senderMsg || options.receiverMsg;
```

crookedneighbor (changes requested):

website/server/models/user/methods.js
lynxlynxlynx: Ok. About the default value: || would catch anything that resolves to false, not just undefined. It's unlikely to get passed, but isn't it better to be on the safe side?

lynxlynxlynx: Ah well, both translations would be the same in that case, so the point is moot.

test/api/v3/integration/members/POST-transfer_gems.test.js
```diff
@@ -4,6 +4,14 @@ import {
 } from '../../../../helpers/api-v3-integration.helper';
 import { v4 as generateUUID } from 'uuid';
 
+function findMessage (messages, receiverId) {
+  let message = _.find(messages, (inboxMessage) => {
```
khipkin, line 8: Suggestion: just move the `return` to this line

website/server/controllers/api-v3/members.js
```diff
@@ -384,18 +384,27 @@ api.transferGems = {
...
+    // generate the message in both languages, so both users can understand it
+    let receiverLang = receiver.preferences.language;
+    let senderLang = sender.preferences.language;
+    let [receiverMsg, senderMsg] = [receiverLang, senderLang].map((lang) => {
+      let messageContent = res.t('privateMessageGiftIntro', {
+        receiverName: receiver.profile.name,
+        senderName: sender.profile.name,
+      }, lang);
+      messageContent += res.t('privateMessageGiftGemsMessage', {gemAmount}, lang);
+      messageContent = `\`${messageContent}\` `;
```
khipkin, line 396: Suggestion: remove the extra whitespace at the end here and instead add it below if necessary

test/api/v3/integration/members/POST-transfer_gems.test.js
lynxlynxlynx: It's a bit cleaner this way, so I agree with crookedneighbor's choice. It should have no noticeable overhead either.

website/server/controllers/api-v3/members.js
lynxlynxlynx: It preserves the current behaviour, but sure, could be changed. I don't think it would be any nicer if moved below though. What would do it in general would be a utility function to wrap strings in code tags, so such ugly packaging can be hidden behind an interface. It would make changing them later a bit easier too (eg. for party chat).

lynxlynxlynx:
I've made the two other strings translatable now, but they're not doubled yet (waiting for confirmation for previous work). I've changed the string to use `byUserName`, which means it's less likely to be empty.

website/common/locales/en/settings.json
```diff
@@ -113,8 +113,10 @@
     "sentGems": "Sent gems!",
     "giftedGems": "Gifted Gems",
     "giftedGemsInfo": "<%= name %> gifted you <%= amount %> Gems",
+    "giftedGemsFull": "Hello <%= username %>, <%= sender %> has sent you <%= gemAmount %> gems!"
```
khipkin: Missing a comma at the end of this line.

website/common/locales/en/settings.json
```diff
@@ -113,8 +113,10 @@
...
     "giftedSubscription": "Gifted Subscription",
     "giftedSubscriptionInfo": "<%= name %> gifted you a <%= months %> Subscription",
+    "giftedSubscriptionFull": "Hello <%= username %>, <%= sender %> has sent you <%= monthCount %> months of subscription!"
```
khipkin: Missing a comma at the end of this line.

website/common/locales/en/settings.json
lynxlynxlynx: Thanks, CI caught it too.

lynxlynxlynx:
Since there was no further comment, I went ahead, split out the double translator and then used it for the other cases as well. The chat syntax wrapping could be done inside, since all users need it, but then the function would be doing two unrelated things at once (ugly).

The new function isn't written defensively. It's not part of the public API, so I wasn't sure how much, if any, input validation it should do. I guess it also needs a few tests of its own?

lynxlynxlynx:
Welp, I'm going to need help on this bit. The transfer-gems tests end up erroring with 500 and no useful message:

> POST /api/v3/members/transfer-gems **500** 93.739 ms - 89
>     1) sends a private message about gems to a user

I see nothing particularly wrong. Should translateTwice be exported also in some other fashion somewhere?

Alys:
I'm sorry we haven't got back to you about this. I had a quick look at the code and I couldn't see anything  wrong at a glance. Have you had a chance to do any debugging? It might help to temporarily remove almost all of the code from the first failing test and then put it back bit by bit until you see what part causes the 500 error. Let us know how that goes!

website/server/controllers/api-v3/members.js
```diff
@@ -408,8 +416,8 @@ api.transferGems = {
     if (receiver.preferences.pushNotifications.giftedGems !== false) {
       sendPushNotification(receiver,
         {
-          title: res.t('giftedGems'),
-          message: res.t('giftedGemsInfo', {amount: gemAmount, name: byUsername}),
+          title: res.t('giftedGems', langs[0]),
```
paglias: res.t automatically passes the user language to the translation function. If you want to pass a different value you'll want to use `i18n.t`

website/common/script/i18n.js
```diff
@@ -60,4 +61,26 @@ function t (stringName) {
   }
 }
 
+// translates passed strings to two (passed) languages and concatenates them
+// returns an array
+function translateTwice (languages, messages) {
```
paglias: this function needs tests, but to be honest I think it would be better to just call `i18n.t` twice, with same parameters except for the language. I find this function to be a little confusing and not really necessary

paglias (changes requested):

website/common/script/i18n.js
lynxlynxlynx: It's there for DRY, which is most evident in the cases where more than one translatable string is needed in construction. Example for explicit duplication: http://sprunge.us/iGaR?diff
But it's easy to go either way.

paglias: I don't know. Because we're creating a completely new function with its own logic and test requirements for a specific situation that it's probably not going to be used elsewhere

lynxlynxlynx: Currently it has 3 users (one more if a test is converted). I took a glance if any similar reports are open, but there are none. Since this handles both gifted subs and gems, I think all automated two user interactions are handled, so you'd be correct.

paglias: @crookedneighbor opinions on this?

crookedneighbor: I agree. This is needlessly complex. We should just call i18n.t twice.

lynxlynxlynx:
I've unrolled the translations and cleaned up the resulting history. It's ready for a final review.

lynxlynxlynx:
Can I get some eyes on this, before it becomes unmergable again?

TheHollidayInn:
@lynxlynxlynx I'll put this on my list to review tonight and tomorrow.

website/server/models/user/methods.js
```diff
@@ -20,15 +20,16 @@ schema.methods.getGroups = function getUserGroups () {
   return userGroups;
 };
 
-schema.methods.sendMessage = async function sendMessage (userToReceiveMessage, message) {
+schema.methods.sendMessage = async function sendMessage (userToReceiveMessage, options) {
```
TheHollidayInn, line 33: Let's document this method. Especially if you are going to add an options parameter. We will want to know what the options are without looking at the source.

website/server/libs/payments.js
```diff
@@ -259,9 +277,28 @@ api.buyGems = async function buyGems (data) {
...
+      gemAmount,
+    }, languages[1]);
+    receiverMsg = `\`${receiverMsg}\``;
+
+    if (data.gift.message) {
+      receiverMsg += ` ${data.gift.message}`;
+      senderMsg += ` ${data.gift.message}`;
+    }
+
+    data.user.sendMessage(data.gift.member, { receiverMsg, senderMsg });
```
TheHollidayInn, line 301: Do you have a test for the buy gems messages?

TheHollidayInn (commented):
Overall, looks good. I left a few comments.

Honestly, I think that your logic for translation a message into two messages of different languages can be abstracted to our translation library. But, I won't enforce it.

website/server/models/user/methods.js
lynxlynxlynx: Do you use any particular format for these, like with the public API or should I just throw an ad-hoc comment above?

website/server/libs/payments.js
lynxlynxlynx: There's the existing test, but it uses the same languages. I could add one akin to the transfer-gems one if you think it is needed.

lynxlynxlynx:
Thanks. As for the abstraction — indeed it can and at first it was, but both paglias and crookedneighbor preferred an explicit approach.

website/server/models/user/methods.js
TheHollidayInn: Okay, use the format in this file above the  creating tasks function: https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/taskManager.js.

TheHollidayInn:
Looks like your last commit broke the build.

lynxlynxlynx:
Yeah, forgot a utility function.

lynxlynxlynx:
Not sure what's wrong with the test. The line number is off, so I guess the user update or refetch is failing for some reason. Any ideas?

test/api/v3/unit/libs/payments.test.js
```diff
@@ -689,14 +691,51 @@ describe('payments/index', () => {
...
+      it('sends gem donation message in each participant\'s language', async () => {
+        await recipient.update({
+          'preferences.language': 'es',
+        });
+        await user.update({
+          'preferences.language': 'cs',
+        });
+        await api.buyGems(data);
+
+        let updatedRecipient = await recipient.get('/user');
```
TheHollidayInn: In unit tests, we don't use "GET" because we are not making api requests. You should use `.sync()` I believe. https://github.com/HabitRPG/habitica/blob/develop/test/helpers/api-integration/api-classes.js#L32

crookedneighbor: @thehollidayinn that method you linked to is only used in integration tests. 

If you need to look up the user, you're better off using Mongoose methods to get the object.

lynxlynxlynx: This was an adaptation of the (working) transfer-gems test you added, but it is an integration one. Is there really a need to refetch the users?

TheHollidayInn: Ah, sorry. I mixed it up myself. 

Just use the tests above for examples. So, you can use this expectation instead of fetching the users again: `expect(user.sendMessage).to.be.calledWith(recipient, { receiverMsg: msg, senderMsg: msg });` You already have separate tests for sendMessage. And it looks like this method is already stubbed for you.

lynxlynxlynx:
Great, now it passes again. :)

TheHollidayInn:
Good work! Added to your Tier progress. :D

PR #8176, opened 2016-10-29
added message for challenges that could not be found - fixes #5538

### LINKED ISSUE #5538: Challenge page for a challenge that is not found may have a more noticeable message
Currently if you click a link to the challenge which no longer exists the page will show two buttons "Back to all challenges" and "Create challenge". [An example](https://habitrpg.com/#/options/groups/challenges/someid).

A standard notification on the right that the challenge is not found may be not quite enough, as it disappears and can be easily missed while the page is loading. And an empty page itself looks like something that shouldn't have happened.

So the page may need some kind of a blurb: `Unfortunately, the challenge you are looking for cannot be found.` If it's difficult to tell if there's no such challenge, I believe a phrase might be used like the one with the party: `Challenge #id not found or it takes time to load`
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23469806-challenge-page-for-a-challenge-that-is-not-found-may-have-a-more-noticeable-message?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

corinnaj:
Fixes https://github.com/HabitRPG/habitica/issues/5538
### Changes

If a user tries to access a challenge that no longer exists, a message will be displayed.

screenshot from 2016-10-29 23-08-14

---

UUID:  712c8649-991b-479a-aed2-ff09f803e592

### DIFF
--- a/website/client-old/js/controllers/challengesCtrl.js
+++ b/website/client-old/js/controllers/challengesCtrl.js
@@ -493,7 +493,7 @@ habitrpg.controller("ChallengesCtrl", ['$rootScope','$scope', 'Shared', 'User', 
           .then(function (response) {
             var challenge = response.data.data;
             $scope.challenges = [challenge];
-          });
+          }).catch(function() { $scope.challengeNotFound = true; });
       } else {
         Challenges.getUserChallenges()
           .then(function(response){
--- a/website/views/options/social/challenges.jade
+++ b/website/views/options/social/challenges.jade
@@ -210,3 +210,6 @@ script(type='text/ng-template', id='partials/options.social.challenges.html')
             .panel-body(ng-class='{collapse: !$stateParams.cid == challenge._id}')
               .accordion-inner(ng-if='$stateParams.cid == challenge._id')
                 div(ui-view)
+
+  .container-fluid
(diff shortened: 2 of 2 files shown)


website/client-old/js/controllers/challengesCtrl.js
```diff
@@ -493,7 +493,7 @@ habitrpg.controller("ChallengesCtrl", ['$rootScope','$scope', 'Shared', 'User',
           .then(function (response) {
             var challenge = response.data.data;
             $scope.challenges = [challenge];
-          });
+          }).catch(function() { $scope.challengeNotFound = true; });
```
crookedneighbor, line 496: The catch function should pass an error. Would it be better to pass on the error message from the API call? That would provide better reporting if the request failed for another reason.

crookedneighbor (changes requested):

crookedneighbor:
I adjusted your commit and made the change I requested: https://github.com/HabitRPG/habitica/commit/b094fb1e527588ddec7a874a5c458ff39ee7d452

PR #8177, opened 2016-10-30
Automatically mute users who attempt to post a slur, fixes #8062

### LINKED ISSUE #8062: Automatic muting (revoking of chat privileges) for players who attempt to post a slur
Related to the [swearword blocker](https://github.com/HabitRPG/habitrpg/issues/8061):

When a user attempts to post a slur that violates the Terms of Service, a notification should be sent to moderators and the user should have `flags.chatRevoked` set to `true` so that they cannot make any posts until moderators have determine whether or not the user needs to be blocked.

This is only for slurs, not for less offensive swearwords.

FYI @lemoness

abatula:
Fixes #8062
### Changes

Work in progress. 
- [x]  Block user from posting a message containing  a word contained in an array of blacklisted words
- [x] Permanently revoke chat privileges
- [x] Faster/more elegant method for checking all banned words
- [x] Load in the official list of banned slurs
- [x] Is there only one file, or are there ones for different languages?
- [x] Email the moderators
- [x] Send slack message to moderators

Tests:
- [x] Blocks posting chat with a slur
- [x] Blocks posting in private group
- [x] Revokes chat privileges
- [x] Sends email to moderators
- [x] Sends slack notification to moderators

---


abatula:
Progress update:
- Does a simple check for banned words in a list
- Revokes chat privileges and prevents posting if a match is found in the message
- Created report to send to the moderators (written but email is untested)

website/server/controllers/api-v3/chat.js
```diff
@@ -100,6 +100,32 @@ api.postChat = {
...
+    let message = req.body.message;
+    let noPunctMessage = " " + message.replace(/[.,\/#!$%\^&\*;:{}=\-_`~()]/g," ") + " ";
+
+    // Check for any slurs in the message
+    for (let i=0; i<slurList.length; i++){
+      let tmp = noPunctMessage.search(" " + slurList[i] + " "); // Don't match partial words
+      if (tmp > -1) {
+        // Need a way to make revoking chat permanent
+        user.muteUser(message, groupId);
+        throw new NotFound('Your message contained inapropriate language, and your chat privileges have been revoked.');
```
TheHollidayInn: This should be moved to a local. Use `res.t()`.

Also, you should move this check below, because you need to check the privacy setting of the group. Private groups can use banned language.

Alys:
@abatula I'm sorry but can you change this back so that it **does** apply to private guilds and parties?

It's true that people can swear in a private group if the group allows it, however slurs that violate the Terms of Service are prohibited everywhere.

abatula:
Progress update:
- Restored blocking of slurs in private groups
- Created generic function to check for banned words
  - Needs to coordinate with #8197 
  - Is the chat.js file an appropriate place for this function? 
- List of banned slurs is loaded from a file in `website/common` 
- Initial code to send a slack message

abatula:
I incorporated the textContainsBannedWords() function from #8197, and added an additional input to specify the list of banned words to test. 

I'm not sure of the best way to share one specific change between two in-progress pull requests without cluttering one of them up with all of the other's changes. A third pull request that these two both pull in/push to? Make sure we're using the same function, and when it's finished delete one of them or hope git is smart enough to realize we both added the exact same code?

abatula:
@TheHollidayInn Are you planning to push your changes to this pull request? Or should I pull in your branch? Our two pull requests are similar but not totally the same, and I want to make sure you get the credit for your code.

TheHollidayInn:
I believe you moved the code over, correct?

If you want, it looks like you could still pull over my integration tests. Don't worry about crediting.

abatula:
Oh, no! Sorry for the misunderstanding. I only copied over your `textContainsBannedWords()` function, because both of our code sections are supposed to use it. I intended to delete that from my code once yours was finished and merged. My code only looks for totally-banned slurs, which automatically revoke chat privileges, while yours is looking for swear words which are acceptable in some private areas and uses different logic. Your test functions and code for checking for swear words are still separate.

I was trying to simplify things and I think I made them more confusing. Sorry about that. :-/

TheHollidayInn:
@abatula I believe I understand. I guess the best way to do this would be to wait for my PR to be merged, then for you to rebase on top of develop. Thanks for explaining.

abatula:
@TheHollidayInn I agree, that sounds like the best way to handle this.

abatula:
Does the code to email and send a slack message need to be in an api file (like the code to flag a message)? I'm trying to figure out if that's the cause of the `error: Error: Slack url is required` failure.

abatula:
I moved the code to email/slack the mods to the chat.js api file, which seems to have gotten rid of the `slack url required` error, at least.  Also switched from NotFound to BadRequest errors. 

I tried restoring chat privileges after each test so that the following tests don't fail, but that doesn't seem to be working properly yet.

abatula:
The current code successfully prevents users from posting chats containing slurs and removes their chat privileges (sets `flags.chatRevoked` to `true`). But when testing, the test fails on `expect(user.flags.chatRevoked).to.be.true;`, saying that `user.flags.chatRevoked` is undefined. Any idea how it could be defined (and used) in the `api.postChat` but not be defined in the test?

test/api/v3/integration/chat/POST-chat.test.js
```diff
@@ -94,10 +94,10 @@ describe('POST /chat', () => {
       error: 'BadRequest',
       message: 'Your message contained inapropriate language, and your chat privileges have been revoked.',
     });
+
     expect(user.flags.chatRevoked).to.be.true;
 
-    user.flags.chatRevoked = false;
-    await user.save();
+    await user.update({'flags.chatRevoked': false});
```
paglias: consider that this only sends a command to mongo, so the database will have `flags.chatRevoked` set to false but the local representation of the user will not be updated, so `user.flags.chatRevoked` won't be `false` but still the old value

abatula:
Tests for blocking slurs in public and private groups pass locally. Still can't directly check that the chatRevoked flag is true. It either doesn't exist, or if I set it to false at the top of the test via:
`user.flags.chatRevoked = false;`
`wait user.update({'flags.chatRevoked': false});`
it fails because it is still set to false. But if I don't use that code at the end of the test, the remaining tests will fail because chat privileges have been revoked due to the slur tests. Not sure what to make of that issue yet.

Other remaining tasks are notifications via email and slack - still in progress.

website/common/bannedSlurs.js
```diff
@@ -0,0 +1,3 @@
+let bannedSlurs = ['TEST_PLACEHOLDER_SLUR_WORD_HERE'];
```
paglias: since this code is only used server side it should go in website/server not website/common, otherwise it'll be included in the client too

paglias (dismissed):

abatula:
When the code detects a slur, it revokes chat privileges and blocks the message in all locations (public and private), but the user can continue to post messages in private groups (or at least the party chat) as long as those new messages don't get blocked for inappropriate language. Is this the desired behavior, or do we want them to be fully muted in all areas for trying to post a slur?

Alys:
@abatula what you've described is the desired behaviour for this issue.

At some future time we'll implement stricter levels of chat blocking and can maybe change the behaviour then. Thank you for asking! I'm sorry about the delay in our reply.

abatula:
Finally got the tests of public and private groups to pass. Tests that the message is blocked and chat privileges are revoked. Also made some changes to more closely mirror the recently merged code for blocking banned non-slur words.

Next on the list is sending email and slack messages to the mods.

abatula:
Is there any information on how sending emails works? I'm trying to make an email with the subject "slur-report-to-mods", based on the email for flagged messages (flag-report-to-mods). But when I test, it says that the subject of the emails is "welcome" instead of what I told it to send (e.g. "slur-report-to-mods"). Even more strangely, if I tell it to use the "flag-report-to-mods", which already exists, the first test ('errors and revokes privileges when chat message contains a banned slur') says that the subject was flag-report-to-mods, but the second test ('does not allow slurs in private groups') still gets a "welcome" message.
Any pointers on where to look or what may be causing this?

Alys:
@abatula Is the test picking up an earlier email that's sent to the user when the user's account is created? You might need to test the second email instead of the first. (`expect(email.sendTxn.args[1][1])` instead of `expect(email.sendTxn.args[0][1])`). That's a wild guess though. :)  If nothing else, adding a (temporary?) test for a second email as well will at least let you see if there is a second email and if so, what it is. If that doesn't help, can you post here the output from the failing test?

abatula:
@Alys Thanks, that fixed it!

The basic functionality is now finished. Tests now check that an email is sent to the moderators when a slur is posted. It also sends and tests  the slack message. 

I have two more questions about this code:
* Some of the variables (like the user email and display name) cause errors when I try to define them in the `slack.sendSlurNotification()` function instead of passing them as arguments from the `chat.js` file. Specifically, it will cause a the following error:
 ``"code": 500
   "error": "InternalServerError"
   "message": "An unexpected error occurred."``
I'm not sure why it does this, but is the way it's set up now a problem?
* What changes would you like made to the message text/formatting of the mesages?

website/server/controllers/api-v3/chat.js
```diff
@@ -141,6 +158,49 @@ api.postChat = {
...
     let group = await Group.getGroup({user, groupId});
 
+    // Check message for banned slurs
+    if (textContainsBannedSlur(req.body.message)) {
+      let message = req.body.message;
+      user.flags.chatRevoked = true;
+      await user.save();
+
+      // Email the mods
+      let authorEmail = user.auth.local.email;
```
Alys: Try using the `getUserInfo` function to get the email address, from `website/server/libs/email.js`. It will fetch addresses for Facebook and Google sign-in users when needed.

This might stop the InternalServerError you were getting (e.g., when there's no `local` authentication email address).

website/server/libs/slack.js
```diff
@@ -93,3 +93,51 @@ module.exports = {
   sendFlagNotification,
   sendSubscriptionNotification,
 };
+
+function sendSlurNotification ({
+  // authorEmail,
+  // author,
```
Alys: This `author` line is failing because when you pass data to this function, you are using `user` as the object's name. If you change this line to `author: user,` it should work.

abatula:
@Alys Thanks! I believe that fixed it.

I think this is no longer a work-in-progress, as the basic functionality (and corresponding tests) are complete. I'm not sure exactly what information you want contained in the emails and slack notifications, so I just took my best guess.

abatula:
Checking in to see what edits or changes are needed to make this PR ready.

SabreCat:
Ah, you're quite right, this should be switched to ready for review. I will take a look later today!

SabreCat:
I cleaned up a typo, but this looks *fantastic* now. @Alys, how would you like to handle adding the canonical list of banned words to website/server/bannedSlurs.js?

SabreCat (approved):

SabreCat:
Checked with @Alys offline--we'll commit adjustments to the banned word lists directly on `develop`. @abatula, thanks so much for all your work here! I've awarded you contributor tier 5 for the combination of this and recent Pixelist contributions.

PR #8194, opened 2016-11-06
replace candy food with normal food and enhance canBuy / canDrop code

Alys:
- replaces candy food with normal food
- adds the `canDrop` and `canBuy` attributes to cakes
- replaces the true/false values on all food types with variable for easier switching next season

FYI @SabreCat

### DIFF
--- a/website/common/script/content/index.js
+++ b/website/common/script/content/index.js
@@ -163,96 +163,101 @@ api.premiumMounts = stable.premiumMounts;
 api.specialMounts = stable.specialMounts;
 api.mountInfo = stable.mountInfo;
 
+// For seasonal events, change these booleans:
+let canUseNormalFood = true;
+let canUseCandyFood = false;
+let canUseCakeFood = false;
+
 api.food = {
   Meat: {
     text: t('foodMeat'),
     target: 'Base',
     article: '',
     canBuy: (function() {
-      return false;
+      return canUseNormalFood;
     }),
-    canDrop: false,
+    canDrop: canUseNormalFood,
   },
   Milk: {
     text: t('foodMilk'),
     target: 'White',
     article: '',
     canBuy: (function() {
-      return false;
+      return canUseNormalFood;
     }),
-    canDrop: false,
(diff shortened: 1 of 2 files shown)


website/common/script/content/index.js
```diff
@@ -163,96 +163,101 @@ api.premiumMounts = stable.premiumMounts;
 api.specialMounts = stable.specialMounts;
 api.mountInfo = stable.mountInfo;
 
+// For seasonal events, change these booleans:
+let canUseNormalFood = true;
```
crookedneighbor: These should be canBuy instead?

crookedneighbor (changes requested):

khipkin:
Suggestion: add food types to the api in chunks and then set `canBy` using an `each` or something on each chunk (sort of like [here](https://github.com/HabitRPG/habitica/pull/7904/files#diff-ec2140f575a660415f495e3edeffb669R184)) to establish uniform behavior that stands the test of time.

website/common/script/content/index.js
Alys: I don't believe so because the variable is used for canBuy and canDrop

Alys: so to clarify, I chose canUse as a generic description of both.

If I saw canBuyNormalFood being used in canDrop, I might think it was a mistake.

Alys:
@khipkin Seems like a good idea! Perhaps something for a new PR though since the main purpose of this is to get rid of the candies now that the Gala is over. The extra code I did was just a quick and easy improvement I saw at the same time. However if anyone posts another comment here saying that it should be done in this PR, I will submit. :) It's not devastating that we still have to eat candy. ;)

website/common/script/content/index.js
Alys: Actually, I've thought better of it and put in both `canBuyCandyFood` and `canDropCandyFood` for naming consistency and so we can still control buying and dropping separately if we ever want to do that.

PR #8195, opened 2016-11-06
Convert test UUID to string to avoid test warning

rrgayhart:
When running unit tests for blockUser, the following warning would be posted in the test logs:

```
validator deprecated you tried to validate a number but this library (validator.js) validates strings only. Please update your code as this will be an error soon. website/common/script/ops/blockUser.js:18:28
```

### Changes

Converts the test fake user id to a string to avoid the error.

May be advisable to change the code in `blockUser` to ignore non-string input, but it seems like an edge case.

----
UUID: 4f82317e-6f0c-4855-9512-3b9b8ca0017d

### DIFF
--- a/test/common/ops/blockUser.test.js
+++ b/test/common/ops/blockUser.test.js
@@ -18,7 +18,7 @@ describe('shared.ops.blockUser', () => {
 
   it('validates uuid', (done) => {
     try {
-      blockUser(user, { params: { uuid: 1 } });
+      blockUser(user, { params: { uuid: '1' } });
     } catch (error) {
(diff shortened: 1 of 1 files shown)


crookedneighbor:
> May be advisable to change the code in blockUser to ignore non-string input, but it seems like an edge case.

There are quite a few places where we use validator and assume string input, which will be a problem if we ever want to update the validator module. 

I think the better course of action would be to wrap the few validator methods we do use and add explicit string checking to all of them.

Not something for this PR though.

crookedneighbor:

PR #8200, opened 2016-11-08
Added note sync when user adds task to challenge, tests, and fixed ch…

### LINKED ISSUE #8189: Participant joins challenge, then creator adds more tasks with Notes -> new user tasks don't get Notes
If a challenge creator edits the challenge to add a new task and at the same time adds a note to the task, the existing challenge participants get a copy of the new task but with the note missing. The desired behaviour is that when a new task is added to a challenge, the notes field should appear in the users' copies of the task.

Any participants who join the challenge after the new task has been added do receive the notes (that is correct behaviour).

Note that this is NOT referring to the situation where a challenge creator edits an EXISTING task in the challenge's edit screen to add a note to it. In that case the note should NOT be copied to the tasks that are already owned by the participants because the participants might have already added their own notes. I believe this feature works correctly (but have not tested it recently) - it's probably the code for that which is causing the bug for new tasks.

Related issue: https://github.com/HabitRPG/habitica/issues/8004 "Participant joins challenge, then creator adds more tasks -> new tasks aren't tagged"

TheHollidayInn:

Fixes #8189

### Changes

- This will sync task notes to challenge members when a task is added to a challenge
- This PR also fixes an issue where challenge model tests were being skipped

----
UUID: 

…allenge tests

### DIFF
--- a/test/api/v3/integration/tasks/challenges/POST-tasks_challenge_id.test.js
+++ b/test/api/v3/integration/tasks/challenges/POST-tasks_challenge_id.test.js
@@ -5,11 +5,16 @@ import {
   translate as t,
 } from '../../../../../helpers/api-v3-integration.helper';
 import { v4 as generateUUID } from 'uuid';
+import { find } from 'lodash';
 
 describe('POST /tasks/challenge/:challengeId', () => {
   let user;
   let guild;
   let challenge;
+
+  function findUserChallengeTask (memberTask) {
+    return memberTask.challenge.id === challenge._id;
+  }
 
   beforeEach(async () => {
     user = await generateUser({balance: 1});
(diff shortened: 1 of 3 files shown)


paglias (approved):

paglias:
LGTM, will wait for someone else to give a look but we can merge for me

Alys:
I realise this is a very late comment and I'm sorry about that, but I'm not sure that this fix works. When I use one account to join a challenge that's owned by a  second account, and then use the second account to edit the challenge to add a task with a note, the task is added to the first account but not the note. It's because the task is copied to the participant account as soon as it's created, before the owner has had a chance to add the note. I haven't reopened the issue though in case I'm missing something. @TheHollidayInn Is there a way of editing challenges that lets this work?

Alys:
... other than by adding a task through the API which isn't a possibility for most challenge creators.

PR #8206, opened 2016-11-11
Fix: PM-blocking a player does not stop them inviting you to a party or guild, fixes #8190

### LINKED ISSUE #8190: PM-blocking a player does not stop them inviting you to a party or guild
When a player uses the "block" feature on another user's profile to prevent that user from sending a private message to the player, the user is still able to invite the player to a party or guild. This is undesirable because blocking a user from sending you PMs should prevent all contact from that user.

This is the PM block feature that I'm referring to (see the icon at the bottom with the "Block" on-hover text):

image

rrgayhart:
Fixes #8190 

### Changes

- Fixes issue where blocking a player does not stop them from inviting you to a party of guild
- Moves shared logic of detecting a blockedUser to it's own function
- Adds unit and integration testing

----
UUID: 4f82317e-6f0c-4855-9512-3b9b8ca0017d

### DIFF
--- a/test/api/v3/integration/groups/POST-groups_invite.test.js
+++ b/test/api/v3/integration/groups/POST-groups_invite.test.js
@@ -97,6 +97,19 @@ describe('Post /groups/:groupId/invite', () => {
       });
     });
 
+    it('returns error when to receiver has blocked the inviter', async () => {
+      let receiver = await generateUser({'inbox.blocks': [inviter._id]});
+
+      await expect(inviter.post(`/groups/${group._id}/invite`, {
+        uuids: [receiver._id],
+      }))
+      .to.eventually.be.rejected.and.eql({
+        code: 401,
+        error: 'NotAuthorized',
+        message: t('notAuthorizedToSendMessageToThisUser'),
+      });
+    });
+
     it('invites a user to a group by uuid', async () => {
       let userToInvite = await generateUser();
 
--- a/test/api/v3/integration/groups/POST-groups_invite.test.js
+++ b/test/api/v3/integration/groups/POST-groups_invite.test.js
@@ -97,7 +97,7 @@ describe('Post /groups/:groupId/invite', () => {
       });
     });
 
-    it('returns error when to receiver has blocked the inviter', async () => {
+    it('returns an error when the receiver has blocked the inviter', async () => {
       let receiver = await generateUser({'inbox.blocks': [inviter._id]});
 
       await expect(inviter.post(`/groups/${group._id}/invite`, {
--- a/website/common/script/libs/userBlockPresent.js
+++ b/website/common/script/libs/userBlockPresent.js
@@ -1,5 +1,0 @@
-module.exports = function userBlockPresent (sender, receiver) {
-  let userBlockedSender = receiver.inbox.blocks.indexOf(sender._id) !== -1;
-  let senderBlockedReceiver = sender.inbox.blocks.indexOf(receiver._id) !== -1;
-  return userBlockedSender || senderBlockedReceiver;
-};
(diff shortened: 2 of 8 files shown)


website/common/script/libs/userBlockPresent.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function userBlockPresent (sender, receiver) {
+  let userBlockedSender = receiver.inbox.blocks.indexOf(sender._id) !== -1;
```
crookedneighbor: Let's make this a bit more DRY

``` js
function isBlocked(user1, user2) {
  return user1.inbox.blocks.indexOf(user2._id) !== -1
}

module.exports = function userBlockPresent (sender, receiver) {
  let userBlockedSender = isBlocked(receiver, sender);
  let senderBlockedReceiver = isBlocked(sender, receiver);
  return userBlockedSender || senderBlockedReceiver;
};
```

crookedneighbor (changes requested):

crookedneighbor:
Overall, seems reasonable.

website/common/script/libs/userBlockPresent.js
khipkin: I agree! But to be fair, the code was like this originally. 😛

crookedneighbor: Picard Nodding

crookedneighbor:
Looks to me like this isn't used anywhere on the client. We may want to implement this logic later on the client, but for now it should be moved to a server lib instead.

website/common/script/libs/userBlockPresent.js
rrgayhart: 😭 ❤️

rrgayhart:
@crookedneighbor Changes implemented! LMK if you see additional adjustments!

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -97,6 +97,19 @@ describe('Post /groups/:groupId/invite', () => {
       });
     });
 
+    it('returns error when to receiver has blocked the inviter', async () => {
```
khipkin: "the" receiver 😈

rrgayhart: 😱 Haha - will fix right now!

rrgayhart:
@crookedneighbor - Quick ping reminder that this PR is ready to be re-looked at when you're able!

website/server/controllers/api-v3/members.js
```diff
@@ -306,11 +307,11 @@ api.sendPrivateMessage = {
...
     if (!receiver) throw new NotFound(res.t('userNotFound'));
 
-    let userBlockedSender = receiver.inbox.blocks.indexOf(sender._id) !== -1;
-    let userIsBlockBySender = sender.inbox.blocks.indexOf(receiver._id) !== -1;
     let userOptedOutOfMessaging = receiver.inbox.optOut;
 
-    if (userBlockedSender || userIsBlockBySender || userOptedOutOfMessaging) {
+    let userBlock = userBlockPresent(sender, receiver);
+
+    if (userBlock || userOptedOutOfMessaging) {
```
TheHollidayInn, line 314: You should add a new test for this route.

TheHollidayInn:
I'm a little late to the review, but I added one comment. Overall, this looks really good. Your coding style is nice.

SabreCat:
This covers a lot of the same ground as https://github.com/HabitRPG/habitica/commit/018976a7231683420acd982bf9705b0046846840 ; I think for this we should expand that functionality to cover these scenarios rather than implement a parallel check. Closing this for now.

PR #8212, opened 2016-11-14
Remove party joined option

### LINKED ISSUE #7442: Date Joined Party Changed? (sort) -- "sort by date joined the party" needs to be removed
General Info
- UUID: 2fb6b0c7-a7eb-4903-a2a9-2f5a0916de47
- Browser: Chrome
- OS: Win 7
### Description

Did the date joined party attribute change in last weekend's maintenance?  My party is sorted differently this morning.  I'm still sorting my party by date joined, ascending, which I was doing before the maintenance.  For example, UUID 7b4987fd-9c6e-402e-8266-22871d1b3f11 was the last person to join our party, but for me they're showing up in the middle of the list rather than at the end.

Group ID: f3fd093a-09d6-407a-ba05-67facbd2b029

librarianmage:

Fixes #7442 

### Changes
Remove the option to sort the party by the date they joined the party (which does not work)

@Alys 

----

### DIFF
--- a/website/client-old/js/controllers/headerCtrl.js
+++ b/website/client-old/js/controllers/headerCtrl.js
@@ -52,7 +52,6 @@ habitrpg.controller("HeaderCtrl", ['$scope', 'Groups', 'User',
                 return member.auth.timestamps.loggedin;
                 break
               default:
-                // party date joined
                 return member.profile.name;
           }
         }
--- a/website/client-old/js/controllers/headerCtrl.js
+++ b/website/client-old/js/controllers/headerCtrl.js
@@ -53,7 +53,7 @@ habitrpg.controller("HeaderCtrl", ['$scope', 'Groups', 'User',
                 break
               default:
                 // party date joined
-                return true;
+                return member.profile.name;
           }
         }
       )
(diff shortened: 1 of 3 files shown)


khipkin:
Food for thought: Maybe the default case in [this switch statement](https://github.com/HabitRPG/habitica/blob/a4fd687510160fc54e5730266f4f921f63f49c53/website/client-old/js/controllers/headerCtrl.js#L55) should be changed to log an error or something since "party joined" will no longer exist?

Alys:
@khipkin Ah, great catch! 

To avoid errors, we'll do a migration before this is merged to change the setting for anyone who has selected `party_date_joined`.

We just need to decide on an appropriate default. We could leave it as just `return true` and then its basically a random sort from the user's point of view. Or we could choose `name` or `habitrpg_date_joined`. I wouldn't recommend the others as a default since it would result in the party order changing often which would look like a bug to anyone who hadn't selected that option deliberately.

librarianmage:
I think that using `return true` might make sense

khipkin:
I think that the default sort should be meaningful.

librarianmage:
Perhaps not meaningful, but it could be the User ID

librarianmage:
@Alys Updated!

website/client-old/js/controllers/headerCtrl.js
```diff
@@ -53,7 +53,7 @@ habitrpg.controller("HeaderCtrl", ['$scope', 'Groups', 'User',
                 break
               default:
                 // party date joined
```
khipkin: Sorry to keep nit-picking, but this comment can be deleted now.

librarianmage:
Done!

Alys:
Tested, all good. Having the removed option set in your account doesn't create errors, and the default name sort then takes effect.

Thanks @MathWhiz ! Notated.

librarianmage:
@Alys :tada:

PR #8219, opened 2016-11-15
Group plans user based subscription

TheHollidayInn:

Fixes put_issue_url_here

### Changes
This adds per month per user subscriptions to group plans

### Todo
- [x]   The “subscription” tab is too small and easy to gloss over. I think the call to action should actually be “upgrade,” to encourage people to investigate it, and to differentiate it from normal subscriptions. (We can A/B test later.) 
- [x]   It says the cost of a group plan is 8/mo + 3/user/month, but it’s charging me $9 for a 1 person guild? I'm confused.
- [x]   The names of the tabs are a bit too vague. Here’s what I would recommend: “Tasks” > “Group Task List”; “Approvals” > “Tasks Awaiting Approval.” (There also needs to be some explanation when the Approvals page is blank. Possibly “When your group completes tasks that need your approval, they’ll appear here! Adjust approval requirement settings under task editing.” (Feedback/other language welcome.)
- [x]   A weird bug: when I add new tasks to the list, it doesn’t display them until I add another task or do a full refresh.
- [x] In popup confirmation: “Do you really want to add ‘NAME’?” > “Do you want to assign this task to NAME?”
-  [x]  Also, I just thought of this now: is there a notification when you get your rewards and it’s approved, or do you just randomly gain points? There should be some kind of notification, such as “Your task “NAME” has been approved!” **From Keith: There is a notification when your task is approved in the top right**
- [x]   “Are you sure you want to approve this task” > “Do you want to reward NAME for completing this task?”
- [x]   Is there any indication for the rest of the team when somebody has claimed a task? Not seeing one aside from in the group leader’s view under task edit. Maybe display it in the chat as a system message and/or in the task notes?
- [x] The tag for the group tasks is green, which is the color reserved for tags that you made yourself. The tags should be purple, like Challenges, or a different color entirely.
-  [x]  It’s confusing to add and then remove the XP and GP for a task pending approval. It should just say, “This task needs to be approved by your group leader before you gain your rewards!”
-  [x]  Also, the task then reappears in my list, exactly as before. There should definitely be some kind of indication that it’s pending approval, not just respawning.
- [x] There’s a choppy cutoff of the “Claim” button (see upcoming screenshot)

### Need Feedback
- [x]   I’m still running into slow loading times and app errors
-  [x]  It looks like the tasks that I tried to assign to a group member didn’t successfully get assigned. **From Keith: Is this happening consistently?**
- [ ]   As I briefly discussed with Keith, we need a different entry point into Group Plans aside from just a button appearing when you create a Guild. (Does it also appear when you already own a group? It should, to upsell you.)
- [ ]   There definitely needs to be a better benefits page. I can do a rough mockup in the next few days. 
-  [x]  It took me a while to find the approval options under Settings, and then I thought it was already enabled because it said “Approval Required.” Seems like we’ll need to change the copy and/or the selection method to make it more clear. That’s something we’ll have to cover in the tutorial, but I feel like it needs additional pointers. Also, it’s very stretched-out, which looks odd. (See screenshot.)
-  [x]  I made a To-Do with checklists, but when I used a different account to claim it, the checklists disappeared. (Possibly because it’s modeled on a Challenge To-Do, which doesn’t support checklists?) - **From Keith: Yes, should I add logic to copy the checklists?**

----
UUID:


website/client-old/js/controllers/menuCtrl.js
```diff
@@ -28,7 +29,7 @@ angular.module('habitrpg')
           } else if (groupNotificationTypes.indexOf('GROUP_TASK_APPROVED') !== -1) {
             return groupApproved;
           }
-          return noneValue;
+          return noneValue;s
```
khipkin: oopsie! assuming this is a typo 😛

TheHollidayInn: Yep :P Thanks. I noticed it locally. Will push a fix soon and any other feedback I received by tomorrow.

website/server/controllers/top-level/payments/stripe.js
```diff
@@ -59,12 +60,21 @@ api.checkout = {
...
+      let customerObject = {
         email: req.body.email,
         metadata: { uuid: user._id },
         card: token,
         plan: sub.key,
-      });
+      };
+
+      if (groupId) {
+        customerObject.plan = 'group_monthly_single_user';
```
TheHollidayInn: This needs to be moved to the plan in common.

website/server/libs/payments.js
```diff
@@ -66,6 +70,7 @@ api.createSubscription = async function createSubscription (data) {
     itemPurchased = 'Group-Subscription';
     purchaseType = 'group-subscribe';
     groupId = group._id;
+    recipient.purchased.plan.quantity = 3;
```
TheHollidayInn: Moved to plan in common

website/common/script/content/subscriptionBlocks.js
```diff
@@ -0,0 +1,36 @@
+import _ from 'lodash';
+
+let subscriptionBlocks = {
+  basic_earned: { // eslint-disable-line camelcase
```
paglias: if you don't want to repeat the esline-disable-... you can write `/* eslint-disable camelcase */` at the top of the file once

website/views/options/settings.jade
```diff
@@ -397,7 +397,7 @@ script(id='partials/options.settings.subscription.html',type='text/ng-template')
         div(ng-if='!user.purchased.plan.customerId || (user.purchased.plan.customerId && user.purchased.plan.dateTerminated)')
           h4(ng-if='(user.purchased.plan.customerId && user.purchased.plan.dateTerminated)')= env.t("resubscribe")
           .form-group.reduce-top-margin
-            .radio(ng-repeat='block in Content.subscriptionBlocks | toArray | omit: "discount==true" | orderBy:"months"')
+            .radio(ng-repeat='block in Content.subscriptionBlocks | toArray | omit: "discount==true" | orderBy:"months"', ng-if="block.key !== 'group_monthly'")
```
paglias: we could use something like `block.key.indexOf('group-') === -1` so that we don't have update this every time a new group plan is added

paglias: or use a new property on `block` to distinguish between users and groups plans like `group.type === 'user'`. Or even split the subscriptionBlocks obj in two different objects

website/server/controllers/api-v3/groups.js
```diff
@@ -301,6 +302,8 @@ api.joinGroup = {
 
     group.memberCount += 1;
 
+    if (group.purchased.plan.customerId) await payments.updateGroupPlan(group);
```
paglias: question: when the group is deleted or created the stripe plan is automatically deleted / created?

website/server/libs/payments.js
```diff
@@ -173,6 +182,21 @@ api.createSubscription = async function createSubscription (data) {
   if (data.gift) await data.gift.member.save();
 };
 
+api.updateGroupPlan = async function updateGroupPlan (group, stripeInc) {
```
paglias: what about renaming this to `updateStripeGroupPlan`?

paglias (changes requested):

paglias:
Looks good!

website/common/script/content/subscriptionBlocks.js
```diff
@@ -0,0 +1,39 @@
...
+  },
+  basic_12mo: {
+    months: 12,
+    price: 48,
+  },
+  group_monthly: {
+    type: 'group',
+    months: 1,
+    price: 9,
+    quantity: 3, // Default quantity for Stripe - The same has having 3 user subscriptions
```
khipkin: Spelling: "The same **as** having"

website/server/libs/payments.js
```diff
@@ -65,7 +70,9 @@ api.createSubscription = async function createSubscription (data) {
     recipient = group;
     itemPurchased = 'Group-Subscription';
     purchaseType = 'group-subscribe';
+    emailType = 'group-subscription-begins';
     groupId = group._id;
+    recipient.purchased.plan.quantity = 3;
```
TheHollidayInn: Change to plan.quantity

website/server/libs/payments.js
```diff
@@ -173,12 +184,29 @@ api.createSubscription = async function createSubscription (data) {
   if (data.gift) await data.gift.member.save();
 };
 
+api.updateStripeGroupPlan = async function updateStripeGroupPlan (group, stripeInc) {
+  if (group.purchased.plan.paymentMethod !== 'Stripe') return;
+  let stripeApi = stripeInc || stripe;
+  let plan = shared.content.subscriptionBlocks;
```
TheHollidayInn: subscriptionBlocks['group']

website/server/models/group.js
```diff
@@ -867,6 +868,10 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
 
   // If user is the last one in group and group is private, delete it
   if (group.memberCount <= 1 && group.privacy === 'private') {
+    let plan = group.purchased.plan;
```
paglias: this should be moved up in the code because some actions have already happened like await `Bluebird.all(challengesToRemoveUserFrom);` and `User.update`. It's possible some has already happened in the controller that calls group.leave, in that case the check should be moved there

website/server/models/group.js
```diff
@@ -867,6 +868,10 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
...
+    if (plan && plan.customerId && !plan.dateTerminated) {
+      throw new NotAuthorized(shared.i18n.t('cannotDeleteActiveAccount'));
```
paglias: I think it deserves a new string specific for groups. Also, just thought about the case when the leader deleted its account, what happens to the group plan?

paglias (changes requested):

website/server/libs/payments.js
```diff
@@ -189,10 +218,13 @@ api.cancelSubscription = async function cancelSubscription (data) {
       throw new NotFound(shared.i18n.t('groupNotFound'));
     }
 
-    if (!group.leader === data.user._id) {
+    let allowedManagers = [group.leader, group.purchased.plan.owner];
+
+    if (allowedManagers.indexOf(data.user._id) === -1) {
```
paglias, line 223: this check should also be made when the subscription is edited?

lemoness:
@TheHollidayInn Re: the checklists -- yes, I think they should copy over.

paglias:
moved to #8263

PR #8220, opened 2016-11-15
Fixed popover translations and moved view to use Angluarjs

### LINKED ISSUE #8210: pet and mount names are not being translated
In the Pets and Mounts pages, the colour (`potion`) and breed (`egg`) of a pet or a mount are not translated, even though translations exist in the locales files for many languages. See screenshot below.

The problem seems to lie within these bits of code:

https://github.com/HabitRPG/habitica/blob/9aa4b8aa64ccdfa3af4a6832cc3dded4b102b73e/website/common/locales/en/pets.json#L74-L75

https://github.com/HabitRPG/habitica/blob/9aa4b8aa64ccdfa3af4a6832cc3dded4b102b73e/website/common/script/content/stable.js#L35-L42

The `(locale)` part in the strings (first link) doesn't seem to be taking effect.

----

### What to do if you notice this for your language:

We know that this issue affects all languages so you do NOT need to comment below or report it anywhere else. When this is fixed, the fix will work for all languages at once.

----

image

TheHollidayInn:

Fixes #8210

### Changes
I moved the server logic to Angular for clarity and because the translations were not finished loading, thus preventing the popovers from being translated. 

----
UUID:

### DIFF
--- a/website/views/options/inventory/mounts.jade
+++ b/website/views/options/inventory/mounts.jade
@@ -7,7 +7,7 @@ mixin mountList(eggSource, potionSource)
             each potion in potionSource
               - var mountKey = egg.key+"-"+potion.key
               - var mount = env.Content.mountInfo[mountKey]
-              div(popover-trigger='mouseenter', popover=mount.text(), popover-placement='bottom')
+              div(popover-trigger='mouseenter', popover=mount.text(env.language.code), popover-placement='bottom')
                 button(class="pet-button Mount_Icon_#{mountKey}", ng-show='user.items.mounts["#{mountKey}"]', ng-class='{active: user.items.currentMount == "#{mountKey}"}', ng-click='chooseMount("#{egg.key}", "#{potion.key}")')
                 button(class="pet-button mount-not-owned", ng-if='!user.items.mounts["#{mountKey}"] && (#{potion.canBuy()} || user.items.hatchingPotions["#{potion.key}"] || user.items.pets["#{mountKey}"])')
                   .PixelPaw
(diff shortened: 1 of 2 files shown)


paglias:
Isn't this going to be quite slow given angular perf? Also what do you mean with translations were not finished loading? All the translations should be available when Jade renders

TheHollidayInn:
As far as translations loading - maybe the are ready, but the local is returning the english version inside of jade. You can verify this way:
1. Go to pets.jade and add the following at the top

```
mixin petList
  - var text = env.Content.petInfo['BearCub-Base'].text()
  #{text}
```
1. Check your browser, you will see that Angular has loaded the correct translation (using the same function call) and displayed value (viewed from web inspector) is the English version.

As for performance, what concerns do you have? I can measure it out - but the only thing we are adding is the loop. The functions were already being called via Angluar after loading.

paglias:
Translations:

the problem is that `env.Content.petInfo['BearCub-Base'].text()` needs to be passed `locale` as a parameter in order to use the correct language, otherwise it'll fallback to English by default. I think it broke after #7995 but I don't know exactly why. It works if you use angular because the client side only load the necessary locale files so it can only use them and doesn't require to specify the language.

Perf:

I'm worried about rendering so many elements with Angular which is one of the reasons why some loops are written in Jade, because Angular is quite slow with these long lists. With the rewrite we're ditching server side rendering and doing everything on the client because Vue is much more performant but I'm not sure it's possible to have good perfs with Angular especially on lower end computers and smartphones. So if possible I would like to keep using the server side rendering for this since it's worked until recently

TheHollidayInn:
Hmm. Looks like it is because the local is set at a different point. How do I access the locale via the `env` variable? Is it just `env.locale`?

crookedneighbor:
Should be env.language if I'm reading the locals middleware correctly.

TheHollidayInn:
Updated

PR #8223, opened 2016-11-17
Added party sync and request sync events

### LINKED ISSUE #8046: health in party doesn't reflect actual health.
General Info
- UUID: 48794933-9b73-4b05-8efd-f44fde5aedf3
- Browser: Google Chrome Version 55.0.2866.0 canary (64-bit)
- OS: Windows 10 64-bit build 14393
### Description

As shown in the picture below, the displayed health in my party is not the same as my health. I did just level up and got my health reset. Just a cosmetic change, not super important.
habitica-bug

TheHollidayInn:

Fixes #8046

### Changes

There are two ways to fix issue 8046 (and other sync issues). One, is to request a sync every time a page is loaded or removed the cache - but this is inefficient. The second, is to request a sync when necessary.

This provides an example of how to solve sync problems. It starts with solving 8046 by requesting a sync when the user's hp has changed because we know that the members in the party will need to be updated as well. This sort of request can be optimized further, but this is a good start. 

----
UUID:

### DIFF
--- a/website/client-old/js/controllers/partyCtrl.js
+++ b/website/client-old/js/controllers/partyCtrl.js
@@ -18,6 +18,11 @@ habitrpg.controller("PartyCtrl", ['$rootScope','$scope','Groups','Chat','User','
       var partyMessageNumber = Math.floor(Math.random() * PARTY_LOADING_MESSAGES) + 1;
       $scope.partyLoadingMessage = window.env.t('partyLoading' + partyMessageNumber);
 
+      $rootScope.$on('partySynced', function (group) {
+        _.assign($rootScope.party, group);
+        $scope.group = $rootScope.party;
+      });
+
       function handlePartyResponse (group) {
         // Assign and not replace so that all the references get the modifications
         _.assign($rootScope.party, group);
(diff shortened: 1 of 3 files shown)


paglias:
For this specific case wouldn't it be better to manually update your HP in the party page instead of making a full party load? Because a party sync request loads 3 things: the party, it's members and all the challenges and would be done every time you score a task or somehow modify your health. Which is not super efficient

TheHollidayInn:
Yep, that would be the more efficient. I just wanted to make sure this method was accepted before getting to details like that. I'll submit an update.

website/client-old/js/services/groupServices.js
```diff
@@ -107,8 +107,12 @@ angular.module('habitrpg')
       });
     };
 
-    $rootScope.$on('syncPartyRequest', function () {
-      party(true);
+    $rootScope.$on('syncPartyRequest', function (event, options) {
+      if (options.type === 'user_update') {
+        var index = _.findIndex(data.party.members, function(user) { return user._id === options.user._id; });
+        var memberToUpdate = data.party.members[index];
+        _.assign(memberToUpdate, User.user);
```
paglias: Can we change this so that it only syncs the fields that are fetched from the server in first place? The members array only contains a few fields, not the entire user object which is very big (contains tasks too on client) and will slow down Angular's change detection

paglias (changes requested):

website/client-old/js/controllers/partyCtrl.js
```diff
@@ -18,6 +18,11 @@ habitrpg.controller("PartyCtrl", ['$rootScope','$scope','Groups','Chat','User','
       var partyMessageNumber = Math.floor(Math.random() * PARTY_LOADING_MESSAGES) + 1;
       $scope.partyLoadingMessage = window.env.t('partyLoading' + partyMessageNumber);
 
+      $rootScope.$on('partySynced', function (group) {
```
paglias: isn't this handled by `function handlePartyResponse` a few lines below?

paglias (changes requested):

PR #8224, opened 2016-11-17
added ability to edit the challenge by the owner

### LINKED ISSUE #8144: challenge owner can't change Reward value when editing challenge
When a player creates a challenge they can add a Reward to it and change the Reward's value to any number (this is correct behaviour).

However when they try to edit the challenge and edit that Reward, they are not able to adjust the value. They should be able to.

Note that this applies to editing the challenge itself, not to editing the copy of the Reward that appears in their own personal task list. The value of their own copy of the Reward must remain uneditable.

dumindux:
@Alys This might be hacky. If so, more input is appreciated. Thanks.
Fixes #8144

### Changes

----
UUID:

### DIFF
--- a/website/views/shared/tasks/edit/rewards/pricing.jade
+++ b/website/views/shared/tasks/edit/rewards/pricing.jade
@@ -1,4 +1,4 @@
-fieldset.option-group.option-short(ng-if='task.type=="reward" && !task._edit.challenge.id')
+fieldset.option-group.option-short(ng-if='task.type=="reward" && challenge_edit')
   legend.option-title=env.t('price')
   input.option-content(type='number', size='16', min='0', step='any', ng-model='task._edit.value', required)
   .money.input-suffix
--- a/website/client-old/js/app.js
+++ b/website/client-old/js/app.js
@@ -230,6 +230,7 @@ window.habitrpg = angular.module('habitrpg',
           title: env.t('titleChallenges'),
           controller: ['$scope', 'Challenges', '$stateParams', 'Tasks',
             function ($scope, Challenges, $stateParams, Tasks) {
+              $scope.challenge_edit = true;
               Challenges.getChallenge($stateParams.cid)
                 .then(function (response) {
                   $scope.obj = $scope.challenge = response.data.data;


website/views/shared/tasks/edit/rewards/pricing.jade
```diff
@@ -1,4 +1,4 @@
-fieldset.option-group.option-short(ng-if='task.type=="reward" && !task._edit.challenge.id')
+fieldset.option-group.option-short(ng-if='task.type=="reward" && challenge_edit')
```
crookedneighbor, line 1: Pretty sure this will prevent users from editing their rewards on their task page.

TheHollidayInn:
Editing tasks is becoming quite and issue. I'd like to this this logic abstracted to the three task controllers: TaskCtrl, ChallengeCtrl, and GroupTaks ctrl into a function called canEdit (which I believe exists, but is ot always used).

Then, I think we should have unit tests written around the test for each case. For example:
 - Owner can edit original challenge task 
 - User can not edit original challenge task
 - User can edit cloned challenge task

Alys:
In case this assists with creating the tests, the following sections in wiki pages will describe the permissions allowed for participants and creators / owners in challenges and group plans.

- Challenge Participant's Permissions http://habitica.wikia.com/wiki/Challenges#Joining_and_Participating_in_Challenges
- Challenge Creator's Permissions http://habitica.wikia.com/wiki/Running_a_Challenge
-  Moderators and staff should have identical permissions to challenge creators.
- Group Plans - http://habitica.wikia.com/wiki/Group_Plans  (this page doesn't yet have clearly-defined lists of permissions but will in future).
- Moderators and staff do **not** have special permissions for editing group  plans tasks.

If anyone has any questions about permissions that aren't answered in those wiki pages or if the pages are unclear, comment here.

Also, in case the terminology isn't clear, "cloning" refers to creating a copy of a complete challenge. Cloning can be done by only the challenge owner or moderators or staff.

SabreCat:
Pulling and running this locally, it doesn't even seem to fix the issue; you can't edit the Reward value _anywhere_. I'm going to close this; if @dumindux or anyone else wants to take a stab at this, step on up!

PR #8226, opened 2016-11-17
[WIP] Incentives rewarding

TheHollidayInn:

Fixes put_issue_url_here

### Changes

Added the rewards for the new login incentives. 

----
UUID:


website/server/libs/cron.js
```diff
@@ -347,7 +393,7 @@ export function cron (options = {}) {
 
   // Login Incentives
   user.loginIncentives++;
-  user.addNotification('LOGIN_INCENTIVE', {});
+  awardLoginIncentives(user);
```
Alys: I'd like this to be moved to earlier in this function to minimise the chance of it not being executed when bugs occur (e.g., it's not uncommon for `_.merge(progress, {down: 0, up: 0, collectedItems: 0}` to be skipped). People are really going to scream if they miss a count for their incentives.

It should also execute for people who are resting in the inn because that includes those who have chronic illnesses and they shouldn't be disadvantaged.

I'd also like `user.loginIncentives` to be added to analytics since this is the sort of thing people will log bug reports about if they think something's gone wrong. This might also be a good time to make `analytics.track('Cron' ...})` run when resting in the inn (see the TODO comment above).

paglias: As long as it's before the line where the user is saved it doesn't really matter if it's before or after

Alys: I've seen cases where cron exits prematurely, for example if corrupted data exists in the user's tasks. It's rare but it happens. When the only effect is that the tasks don't get adjusted and quest progress isn't made, it's not a big problem, but this feature will be popular and highly visible. Good point though about it not mattering since there's only one save location (sorry, I should have thought of that), but can we consider putting the most important cron actions together and having a save for them, then a second save for everything else? It might be more trouble than it's worth I suppose, but I'm afraid of this being another feature that gets bug reports and needs manual investigations and fixes when unexpected conditions occur. :(

website/server/libs/cron.js
```diff
@@ -113,6 +113,87 @@ function performSleepTasks (user, tasksByType, now) {
...
+      user.balance = 4 * 5;
+      break;
+    case 7:
+      user.items.quests.moon1 = 1;
+      break;
+    case 10:
+      user.items.hatchingPotions.Purple = 1;
+      break;
+    case 14:
+      user.balance = 4 * 3;
```
paglias: this is going to override the entire balance, even if the user already had its own gems

website/server/libs/cron.js
paglias: I see what you're saying, I think the reason why I put the user saving after the tasks processing was because if the user is saved, saying that cron was done, the tasks won't be processed anymore. Unfortunately these bugs will keep happening, we just have to decide which part is more important

website/server/libs/cron.js
TheHollidayInn: 😳 You are right. Thanks!

website/server/libs/cron.js
TheHollidayInn: I moved the incentives code up. I can look into adding a second save if we want.

Alys: It's more complex than I originally thought it would be, so if paglias doesn't think it's a good idea, I withdraw my request.

paglias: @Alys I'm not saying that it's not a good idea but just that if we add another save location for the user then we either save the lastCron date the first time we save the user, which means that the parts that didn't run will be skipped or we do not save lastCron which means we could have a partial double cron in case of errors. Unfortunately there's no 100% correct way to do this without transactions

Alys: Yeah, too complex. Let's forget it. Thanks for the explanations.

TheHollidayInn:
Here is my progress on the daily modal

website/common/script/content/loginIncentives.js
```diff
@@ -85,7 +85,7 @@ module.exports = function getLoginIncentives (api) {
       },
     },
     30: {
-      rewardKey: 'balance',
+      rewardKey: 'Pet_Currency_Gem2x Gems',
       reward: '',
       assignReward: function assignReward (user) {
         user.balance += 4 * 3;
```
paglias: I thought we decided the gems incentives would come in v2?

TheHollidayInn: Yea, we did. I haven't no updated this with the new reward list.

khipkin:
Are all of the purple pet/mount sprites done for the base animals?

TheHollidayInn:
They are, but they are being added in a separate PR.

website/client-old/js/config.js
```diff
@@ -35,7 +35,7 @@ angular.module('habitrpg')
           if (isUserLoaded && notification.type === 'CRON') {
             // If the user is already loaded, do not show the notification, syncing will show it
             // (the user will be synced automatically)
-            $rootScope.User.readNotification(notification.id);
+            // $rootScope.User.readNotifications([notification.id]);
```
paglias, line 38: what about this?

website/client-old/js/controllers/notificationCtrl.js
```diff
@@ -140,14 +141,33 @@ habitrpg.controller('NotificationCtrl',
...
+            modalScope.user = User.user;
+            modalScope.loadWidgets = Social.loadWidgets;
+
+            var modalKey = 'login-incentives';
+            if (notification.data.rewardKey) {
+              modalKey = 'login-incentives-reward-unlocked';
+            }
+
+            $rootScope.openModal(modalKey, {
+              scope: modalScope
```
paglias, line 158: question: cannot you just pass $rootScope here instead of creating a new one?

website/server/controllers/api-v3/notifications.js
```diff
@@ -43,4 +43,44 @@ api.readNotification = {
...
+ * @api {post} /api/v3/notifications Mark notifications as read
+ * @apiName ReadNotifications
+ * @apiGroup Notification
+ *
+ *
+ * @apiSuccess {Object} data user.notifications
+ */
+api.readNotifications = {
+  method: 'POST',
+  url: '/notifications',
```
paglias: This url should be changed to something else like `/notifications/read` because usually POST /notifications would mean a route to create a notification

paglias (changes requested):

website/client-old/js/config.js
TheHollidayInn: From manual testing, I have seen that this is duplicate. The Cron notification is read by the noitificationCtrl .

website/client-old/js/controllers/notificationCtrl.js
TheHollidayInn: That would be inefficient, right? The $rootScope is large.

website/client-old/js/config.js
paglias: I think the point of this was to avoid showing a double notification for cron, reading it here makes sure it's not handled by noitificationCtrl

website/client-old/js/controllers/notificationCtrl.js
paglias: what does exactly $rootScope.new() does?

website/client-old/js/config.js
TheHollidayInn: Well, CRON has tests to combine notifications. And, this breaks the code by creating a race condition of sorts. So, I'm not sure what id does or if it is needed from tests.

website/client-old/js/controllers/notificationCtrl.js
TheHollidayInn: Creates a new child scope. Also, attaching the notification data to the root scope doesn't make sense. Creating a new scope for these bootstrap modals is pretty normal.

paglias: 👍

website/client-old/js/config.js
paglias: No I think the point of this code was another. If you're online and your cron time passes, the next time you complete and action two things happens:

1. The notification about cron is sent to the browser
2. the browser notices that the user has been updated (user._v changed) and issue a sync request and if there's a difference between the local user data and what the server sent (for example because hp changed after cron) it'll show another notification for the HP change

So this code is there to avoid showing the same notification twice. Also if it's there it's for a reason and before removing existing code I would like us to check whether it's used or not. Also because commenting it leaving the comments above in place and an empty IF condition doesn't make sense and only increase confusion for the next person looking at the code

website/client-old/js/controllers/menuCtrl.js
```diff
@@ -116,7 +116,7 @@ angular.module('habitrpg')
       };
 
       $scope.viewGroupApprovalNotification = function (notification, $index) {
-        User.readNotification(notification.id);
+        User.readNotifications([notification.id]);
```
paglias: if we're marking only one notification as read, let's keep the old readNotification function

website/common/script/content/loginIncentives.js
```diff
@@ -0,0 +1,169 @@
...
+      reward: [api.food.Saddle],
+      assignReward: function assignReward (user) {
+        if (!user.items.food.Saddle) user.items.food.Saddle = 0;
+        user.items.food.Saddle += 1;
+      },
+    },
+  };
+
+  // Add refence link to next reward and add filler days so we have a map to refernce the next reward from any day
+    // We could also, use a list, but then we would be cloning each of the rewards.
```
paglias: indentation

website/server/libs/cron.js
```diff
@@ -113,9 +112,45 @@ function performSleepTasks (user, tasksByType, now) {
...
+  notificationData.message = i18n.t('checkinEarned');
+
+  let loginIncentive = loginIncentives[user.loginIncentives];
+
+  if (loginIncentive.rewardKey) {
+    loginIncentive.assignReward(user);
+    notificationData.reward = loginIncentive.reward;
+    if (loginIncentive.reward[0].text) notificationData.rewardText = loginIncentive.reward[0].text();
+    notificationData.rewardKey = loginIncentive.rewardKey;
+    notificationData.message = i18n.t('unlockedCheckInReward');
```
paglias: can we pass user.preferences.language to this function so it gets translated?

website/server/libs/cron.js
```diff
@@ -113,9 +112,45 @@ function performSleepTasks (user, tasksByType, now) {
...
+    progressDown: _progress.down,
+    headers: options.headers,
+    loginIncentives: user.loginIncentives,
+  });
+}
+
+function awardLoginIncentives (user) {
+  if (user.loginIncentives > 50) return;
+  let notificationData = {};
+  notificationData.message = i18n.t('checkinEarned');
```
paglias: pass user.preferences.language

paglias (changes requested):

PR #8242, opened 2016-11-29
Login incentives remove multiple notifications

TheHollidayInn:

Fixes put_issue_url_here

### Changes
This removes previous login incentive notifications so that there is always at max, one. 

----
UUID:

### DIFF
--- a/test/api/v3/unit/libs/cron.test.js
+++ b/test/api/v3/unit/libs/cron.test.js
@@ -704,14 +704,14 @@ describe('cron', () => {
 
       cron({user, tasksByType, daysMissed, analytics});
 
-      expect(user.notifications.length - notifsBefore2).to.equal(1);
-      expect(user.notifications[1].type).to.not.equal('CRON');
-      expect(user.notifications[2].type).to.equal('CRON');
-      expect(user.notifications[2].data).to.eql({
+      expect(user.notifications.length - notifsBefore2).to.equal(0);
+      expect(user.notifications[0].type).to.not.equal('CRON');
+      expect(user.notifications[1].type).to.equal('CRON');
+      expect(user.notifications[1].data).to.eql({
         hp: user.stats.hp - hpBefore2 - (hpBefore2 - hpBefore1),
         mp: user.stats.mp - mpBefore2 - (mpBefore2 - mpBefore1),
       });
-      expect(user.notifications[1].type).to.not.equal('CRON');
+      expect(user.notifications[0].type).to.not.equal('CRON');
     });
   });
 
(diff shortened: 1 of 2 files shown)


website/server/libs/cron.js
```diff
@@ -130,6 +130,14 @@ function trackCronAnalytics (analytics, user, _progress, options) {
 
 function awardLoginIncentives (user) {
   if (user.loginIncentives > 50) return;
+
+  //  Remove old noitification if it exists
```
crookedneighbor: Spelling error

website/server/libs/cron.js
```diff
@@ -130,6 +130,14 @@ function trackCronAnalytics (analytics, user, _progress, options) {
...
+  user.notifications
+    .toObject()
+    .find((notif, index) => {
```
crookedneighbor, line 137: To make use of find to short circuit the loop, you should return a truthy value when you've done the required operation. The way you have it now, it will continue to search through the array after it has found the notification.

test/api/v3/unit/libs/cron.test.js
```diff
@@ -780,6 +780,18 @@ describe('cron', () => {
       expect(user.notifications[0].type).to.eql('LOGIN_INCENTIVE');
     });
 
+    it('replaces previous notifications', () => {
+      cron({user, tasksByType, daysMissed, analytics});
+      cron({user, tasksByType, daysMissed, analytics});
+      cron({user, tasksByType, daysMissed, analytics});
+
+      let filteredNotifications = filter(user.notifications, function filterNotifications (notification) {
```
crookedneighbor: Why not use the array filter method and an arrow function?

website/server/libs/cron.js
khipkin: "notification"

crookedneighbor (changes requested):

website/server/libs/cron.js
TheHollidayInn: I want it to remove multiple notifications, not the first. So, I don't want to short circuit.

test/api/v3/unit/libs/cron.test.js
TheHollidayInn: I wasn't aware of it. I can change it.

website/server/libs/cron.js
```diff
@@ -130,6 +130,14 @@ function trackCronAnalytics (analytics, user, _progress, options) {
 
 function awardLoginIncentives (user) {
   if (user.loginIncentives > 50) return;
+
+  //  Remove old notifications if they exists
+  user.notifications
+    .toObject()
+    .find((notif, index) => {
```
crookedneighbor, line 137: If you don't want it to exit early, why are you using find? You should be using forEach instead. 

Also, why not exit early? If this is working as intended, there should just be the one login incentives notification. The only exception is how it is currently, where we've allowed multiples to stack up. But that should be resolved as soon as a client accepts them. 

I could see the argument to keep it like this for a week to prune most of the duplicates, then switch to a version where it exits early.

website/server/libs/cron.js
```diff
@@ -130,6 +130,14 @@ function trackCronAnalytics (analytics, user, _progress, options) {
 
 function awardLoginIncentives (user) {
   if (user.loginIncentives > 50) return;
```
crookedneighbor, line 132: Something for the future, but this should be a constant instrwad of a magic number.

PR #8243, opened 2016-11-29
Quest Panel Reload (Issue #7697)

### LINKED ISSUE #7697: Inviting party to quest should hard-refresh party page
I invited my party to quest, but the only visual feedback I got was the quest modal going away.  Fetching recent messages and pressing the "sync" button didn't make any difference.  The quest invitations panel only appeared after I manually refreshed the whole page.

This is super weird to me and doesn't seem like desirable behavior.

Thoughts?

_edit by Alys, 16 October 2016:_

See [this comment](https://github.com/HabitRPG/habitica/issues/7697#issuecomment-254025148) and below it for the most recent status of this issue.

marceloyuela:

Fixes #7697 

### Changes
Forces a reload on the quest panel state, after a quest has been initialized.

### User ID
c10b82f6-6d0d-4d1d-94af-6dbbe35f7f2b
----

### DIFF
--- a/website/client-old/js/services/questServices.js
+++ b/website/client-old/js/services/questServices.js
@@ -104,7 +104,7 @@ angular.module('habitrpg')
             Groups.data.party = party;
             $state.go('options.social.party');
             resolve();
-            if ($state.current.name == "options.social.party") {
+            if ($state.current.name === "options.social.party") {
               $state.reload();
             }
           });
--- a/website/client-old/js/services/questServices.js
+++ b/website/client-old/js/services/questServices.js
@@ -104,6 +104,8 @@ angular.module('habitrpg')
             Groups.data.party = party;
             $state.go('options.social.party');
             resolve();
+          }).then(function() {
+            $state.reload();
           });
       });
     }
--- a/website/client-old/js/services/questServices.js
+++ b/website/client-old/js/services/questServices.js
@@ -104,8 +104,9 @@ angular.module('habitrpg')
             Groups.data.party = party;
             $state.go('options.social.party');
             resolve();
-          }).then(function() {
-            $state.reload();
+            if ($state.current.name == "options.social.party") {
+              $state.reload();
+            }
           });
       });
     }


marceloyuela:
Is there something I should do to get someone to review this?

Alys:
Ah, I'm sorry, no, you've done the right thing in posting it here. We're overloaded at the moment. :( I'll test this locally today and then if there's no comments from other admins in the next couple of days, I'll merge it.

Alys:
When a quest is started from the Inventory > Quests page, this change prevents the website from redirecting the user to the party page. Can you do something to still allow that to happen?

marceloyuela:
Hey, sorry I didn't notice that before. I completely forgot that was an option. What was happening was that I was forcing the state to reload always, but that would just reload the inventory (since the initial state was the inventory).

Fixed that and removed the unnecessary .then.

website/client-old/js/services/questServices.js
```diff
@@ -104,6 +104,9 @@ angular.module('habitrpg')
             Groups.data.party = party;
             $state.go('options.social.party');
             resolve();
+            if ($state.current.name == "options.social.party") {
```
crookedneighbor: Triple equals.

crookedneighbor (changes requested):

marceloyuela:
Fixed to === instead of ==.

website/client-old/js/services/questServices.js
```diff
@@ -104,6 +104,9 @@ angular.module('habitrpg')
             Groups.data.party = party;
             $state.go('options.social.party');
             resolve();
+            if ($state.current.name === "options.social.party") {
+              $state.reload();
```
crookedneighbor, line 108: It's been a while since I looked at the code, but is there a reason to have this logic here instead of in the controller that uses this service? Wouldn't need special logic then.

marceloyuela: I'm unsure if you can do this.

The chain of command from the party screen starts in the [PartyCtrl](https://github.com/HabitRPG/habitica/blob/develop/website/client-old/js/controllers/partyCtrl.js#L157): If we have quests we load the InventoryCtrl, if not we switch to the Inventory state (see other chain of command). From the InventoryCtrl, the [ownedQuests jade file](https://github.com/HabitRPG/habitica/blob/a4fd687510160fc54e5730266f4f921f63f49c53/website/views/shared/modals/quests.jade#L126) is constructed and placed into the quest popup. Then we deal with the logic that I changed.

In the chain of command from the Inventory side, it's similar but we don't have the Party Ctrl.

If we move the "$state.reload" into the party controller, it doesn't reload after starting a quest( I tried this out). Probably due to the fact that the chain of command doesn't finish before it reloads, due to the popup quest selection window.

Moving it into the InventoryCtrl doesn't quite make sense/work, since it's still used by both chains of command so it would still require a conditional.

I may be wrong about some of this but I do think the logic should be there. I am also moderaely unfamiliar with the codebase as a whole though.

crookedneighbor: I bet this is being called from the modal, so it's out of the context from the party controller. Doing the logic here is fine. I'd prefer to just fetch the part object though, rather than do a reload.

marceloyuela: To be honest I'm very unsure how to do this. Any suggestions?

marceloyuela: While poking around for doing something like this I figured out that the $scope.go line can be changed to
`$state.go('options.social.party', {}, {reload: true});`

and then you don't need the if statement at all, as it "forces" a reload. Is this preferred or should I keep the if statement?

crookedneighbor: I'm ambivalent about it. What you have already is fine with me as a stop gap until the front end redesign is complete.

crookedneighbor (approved):

website/client-old/js/services/questServices.js
marceloyuela: I'll keep it as is- better to only have reload when needed.

marceloyuela:
Anything else I should look at?

website/client-old/js/services/questServices.js
```diff
@@ -104,6 +104,9 @@ angular.module('habitrpg')
             Groups.data.party = party;
             $state.go('options.social.party');
             resolve();
+            if ($state.current.name === "options.social.party") {
```
TheHollidayInn, line 107: Is there a reason this is comes after the promise is resolved? I have this PR on my list to try.

marceloyuela: No. It works either way. Is it preferable to have it before the resolve?

TheHollidayInn: Eh, it seems odd, but having it before the resolve doesn't resolve the oddness. Let me test this locally, then get back with you.

TheHollidayInn:
Awarded towards your Tier. Congrats on the first Tier!!

PR #8244, opened 2016-11-29
New feature that notifies a user when their group invite is accepted.

### LINKED ISSUE #7788: When an invitation is accepted, the inviter should be notified.

Hus274:

Fixes #7788 

### Changes
Created a new feature that notifies the inviting user when another use accepts their invitation to join a group. This works for both parties and guilds. I've also added tests to support this new behavior.

screen shot 2016-11-29 at 3 20 30 pm

This is in replacement of the previous pull request #7797 which I closed due to the website refactoring.


crookedneighbor:
This might have been discussed in the previous PR, but why use the easy to miss blue notification instead of the modal notifications stored in user.notifications?

Hus274:
That's because it seems to be what paglias was suggesting in the last one:
> I was thinking more about a one time blue message in the top right that you can trigger from the Notification service.

I get what you're saying about it being easy to miss and may make more sense as a modal and I'm alright with that direction as well.

Hus274:
I've updated this to use a modal instead:

screen shot 2016-11-29 at 5 57 24 pm

I figured the diff is pretty minimal so i can just revert if we don't want to use a modal.

website/client-old/js/controllers/notificationCtrl.js
```diff
@@ -144,6 +144,11 @@ habitrpg.controller('NotificationCtrl',
           case 'LOGIN_INCENTIVE':
             Notification.showLoginIncentive(User.user, notification.data, Social.loadWidgets);
             break;
+          case 'GROUP_INVITE_ACCEPTED':
```
crookedneighbor: Do we not have a generic notification modal we use when the type isn't found on the client?

paglias: I don't think so, we could add it though

crookedneighbor: I think we should, rather than continually adding new modals.

paglias:
Yeah, it'll also make it easier to handle these notifications in other
clients like the mobile apps since the message will be provided by the
server instead of being specified in each modal source code

Il 30 Nov 2016 15:38, "Blade Barringer" <notifications@github.com> ha
scritto:

Hus274:
I'm not sure I totally understand how to design what you're looking for here. Would the server just return the header and message for the modal and create a default case statement that takes any undefined notification type and assume it has a header and body and then just pass that into a generic modal that displays those fields?

paglias:
@Hus274 I think for generic notifications notification.data should have `{title, text}` and when a notification type hasn't a special handler a modal is shown with `data.title` in the header and `data.header` in the body

Hus274:
@paglias Ok cool, that's about what I was understanding. I can take a look into making this generic then. Thanks!

Hus274:
I added a new generic modal template and made the invite accept notification use that instead. It allows the server to pass header and body text keys in addition to header and body data fields that can be used to fill in the translatable text to make it a little more flexible and reusable.

Let me know what you think.

Hus274:
Updated this for @paglias' comments. I've moved the translation to the server side and required that the generic modal popup only need the data object passed to it from the notification to have the translated headerText and bodyText in order for the modal to properly work.

Thanks for all the feedback @paglias, it was super helpful! Let me know if there's anything else that needs to get updated here!

website/server/controllers/api-v3/groups.js
```diff
@@ -303,6 +305,18 @@ api.joinGroup = {
 
     let promises = [group.save(), user.save()];
 
+    if (inviter) {
+      inviter = await User.findById(inviter).select('notifications preferences.language').exec();
+
+      let data = {
+        headerText: common.i18n.t('invitationAcceptedHeader', inviter.preferences.language),
+        bodyText: common.i18n.t('invitationAcceptedBody', {groupName: group.name, username: user.auth.local.username},
+          inviter.preferences.language),
```
paglias: I would put this back on the upper line and instead split the `{groupName: group.name, username: user.auth.local.username}` on multiple lines, like

```
data = {
  headerText: common.i18n.t('invitationAcceptedHeader', inviter.preferences.language),
  bodyText: common.i18n.t('invitationAcceptedBody', {
    groupName: group.name, 
    username: user.auth.local.username
  }, inviter.preferences.language),
}
```

website/server/controllers/api-v3/groups.js
```diff
@@ -303,6 +305,18 @@ api.joinGroup = {
...
+      };
+      inviter.addNotification('GROUP_INVITE_ACCEPTED', data);
+      promises.push(inviter.save());
+    }
+
     if (group.type === 'party' && inviter) {
       promises.push(User.update({_id: inviter}, {$inc: {'items.quests.basilist': 1}}).exec()); // Reward inviter
```
paglias: since we load the inviter, it's better to do the modification on the inviter object and then `save` instead of a separate update (you'll need to load `items.quests.basilist` other than `notifications preferences.language `

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
```diff
@@ -134,6 +134,14 @@ describe('POST /group/:groupId/join', () => {
 
         await expect(user.get('/user')).to.eventually.have.deep.property('items.quests.basilist', 1);
       });
+
+      it('notifies inviting user that their invitation was accepted', async () => {
+        await invitedUser.post(`/groups/${guild._id}/join`);
+
+        let inviter = await user.get('/user');
```
paglias, line 141: can we add expectation for the notification data too?

paglias (changes requested):

paglias:
@Hus274 I left 3 comments, but they're minor, otherwise it's very good and I'll merge it as soon as these last things are addressed. Thanks!

Hus274:
Made those final changes. Definitely a good call on expecting on the data output from the join method. Let me know if there's anything else that needs to be updated!

paglias:

PR #8249, opened 2016-12-02
feature: adding hp notification for boss damage

### LINKED ISSUE #7749: Post cron damage report does not include boss damage
General Info:
- UUID: 66902097-c79b-496f-acd9-b78cc5867b28
- Browser: FF (47)
- OS: Windows 7

Description:
I noticed that after shutting down my machine for the evening pre-cron, when I wake the next day and load habitica, damage for incomplete dailies is given but that damage does not include damage (also from incomplete dailies) given by a quest boss you're fighting.

---Example---
Before I went to bed full health:
habitica-nightbefore2

Next day when I got up (see damage is different than cron damage overnight report):
habitica-dayafter2

The boss damage (added together they total what I took correctly):
habitica-damage

I spoke to @Alys about this in the Report a Bug Guild (now scrolled off) and was suggested that I post screens if I saw a discrepancy and I felt this was the right place (rather than there again).  Not convinced I explained it very well there; hopefully this is a bug.  Let me know if still unclear.

If it's intentional, I'm curious if a suggestion to include it would be accepted (if one doesn't already exist) or if that'd be a waste of my time.  It seems like seeing all the damage you've taken would be useful rather than partial.  This seems to be how it works when I switch between machines (gives a summary of gains/losses since last accessed rather than just the latest entry or individual entries).

@Alys also mentioned when helping me in the Report a Bug Guild that my health was different at that time then post-cron which is also currently true.  If it helps to keep my health the same after this occurs I probably can (at least for a while=)).  I wasn't sure the relavence so decided to go about my normal day.

PS: I don't believe related but when I mentioned I was doing this to my party member today, they mentioned that they had not taken any damage that day despite me taking it from the boss (and their cron going off; I checked and indeed they had damaged the boss after me).  I am mentioning in the off chance it's related (and maybe specific to my group somehow).  I've never noticed this myself but probably because they are kind and never damage me.

Hus274:

Fixes #7749 

### Changes
I've added a new notification that notifies the user of boss damage taken at cron. This is a separate notification from damage. The notification will not be triggered if the boss does 0 damage and the notification aggregates the damage across multiple notifications if multiple instances of boss damage occur between user logins.

screen shot 2016-12-02 at 1 22 18 pm

Let me know any feedback!


Alys:
I love it! I think though that we might get some questions about whether it's damage to or from the boss - it should be obvious since it's reported in the same way as other damage to the player, but we do get questions for things that seem obvious to us. :)  Can you change the English text to "Damage from Boss"?

Hus274:
@Alys that makes total sense and it's definitely worth avoiding the confusion! I'll add it quick

Hus274:
Here's a screenshot with the new text:

screen shot 2016-12-02 at 2 47 23 pm

TheHollidayInn:
Looks good! If you fix the merge conflict, I'll merge

Hus274:
Fixed the merge conflicts, should be good to go

TheHollidayInn:
Noted towards your tier. Thanks!

joedragons:
Thanks @Hus274 !!

PR #8251, opened 2016-12-04
Implementing retries on failed user updates when finishing a quest

### LINKED ISSUE #8035: code change needed for "Quest completed but rewards and completion not recorded for quest participant(s)"
As described in https://github.com/HabitRPG/habitrpg/issues/4252, when quests finish, they sometimes do not give the rewards and achievement to all participants. I don't believe there's anything buggy with the code; I think it's just timeouts or similar, so we need code that works in a way that can't be broken by slow / overloaded web or database servers.

I proposed an idea in this comment: https://github.com/HabitRPG/habitrpg/issues/4252#issuecomment-226402321 (copied below), and there were no complaints or other suggestions when I posted about it again a few days ago (https://github.com/HabitRPG/habitrpg/issues/4252#issuecomment-246515501), so I'm opening this issue for that fix to be done.

But if anyone does have better ideas, definitely tell us! Posting to https://github.com/HabitRPG/habitrpg/issues/4252 would be better than commenting here (I'll close this issue if something better is proposed there).

---

When the quest ends, we collate the participant's IDs and the rewards they are due, and store all that in a `questRewardsPending` object in the party's `groups` document - write all that in a single update. If the update fails, retry until it succeeds. The update would need to NOT overwrite any data from previous quests, so have the object keyed with timestamps.

Then we have a separate process that reads that object and handles the users one by one. When it has **successfully** updated a user, it deletes that user's data from the `questRewardsPending` object (and if the update fails, it tries again until it succeeds). That way the worst bug we can get is that a user doesn't get the update immediately, or they get the update twice if the update does succeed but the deletion from `questRewardsPending` fails, causing the update to later be applied again - we can put up with that bug!

That separate process could be run after the quest had ended. To cover the cases where it timed out, the same process (or a similar per-user process) could also be run at other times - e.g., when a user loaded the party page or cronned, the process could look in the user's party's `questRewardsPending` object and apply anything it finds there for the user.

There'd be edge cases where the end-of-quest process had timed out for a user and the user left the party before any future processing gave them the rewards, but at least there'd be a record in the database of what they were owed and it would be fairly easy for an admin to find it and apply it. Also, the code that's run when a user leaves the party could look for any pending rewards they were owned and apply them.

Hus274:

Fixes #8035 

### Changes
Added a finite retry system for individual user updates when finishing a quest. This should help prevent users not getting the quest rewards when their party finishes a quest. The retry system currently allows up to 30 seconds of retries but this can be adjusted as needed depending on the load and timeouts seen by mongo.

Let me know if you have any comments!

### DIFF
--- a/test/api/v3/unit/models/group.test.js
+++ b/test/api/v3/unit/models/group.test.js
@@ -1173,11 +1173,12 @@ describe('Group Model', () => {
         it('updates participating members with rewards', async () => {
           await party.finishQuest(quest);
 
-          expect(User.update).to.be.calledOnce;
+          expect(User.update).to.be.calledTwice;
           expect(User.update).to.be.calledWithMatch({
-            _id: {
-              $in: [questLeader._id, participatingMember._id],
-            },
+            _id: questLeader._id,
+          });
+          expect(User.update).to.be.calledWithMatch({
+            _id: participatingMember._id,
           });
         });
 
--- a/test/api/v3/unit/models/group.test.js
+++ b/test/api/v3/unit/models/group.test.js
@@ -1071,7 +1071,7 @@ describe('Group Model', () => {
         };
         let failedMock = {
           exec: () => {
-            return Promise.reject('err');
+            return Promise.reject(new Error('error'));
           },
         };
 
@@ -1096,7 +1096,7 @@ describe('Group Model', () => {
         it('retries failed updates at most five times per user', async () => {
           sandbox.stub(User, 'update').returns(failedMock);
 
-          await party.finishQuest(quest);
+          await expect(party.finishQuest(quest)).to.eventually.be.rejected;
 
           expect(User.update.callCount).to.eql(10);
         });
--- a/test/api/v3/unit/models/group.test.js
+++ b/test/api/v3/unit/models/group.test.js
@@ -1061,8 +1061,45 @@ describe('Group Model', () => {
           [nonParticipatingMember._id]: false,
           [undecidedMember._id]: null,
         };
+      });
 
-        sandbox.spy(User, 'update');
+      describe('user update retry failures', () => {
+        let successfulMock = {
+          exec: (callback) => {
+            callback(null, 'sucess');
+          },
+        };
+        let failedMock = {
+          exec: (callback) => {
+            callback('error', null);
+          },
+        };
+
+        it('doesn\'t retry successful operations', async () => {
+          sandbox.stub(User, 'update').returns(successfulMock);
+
+          await party.finishQuest(quest);
+
+          expect(User.update).to.be.calledTwice;
+        });
+
+        it('stops retrying when a successful update has occurred', async () => {
+          let updateStub = sandbox.stub(User, 'update');
+          updateStub.onCall(0).returns(failedMock);
+          updateStub.returns(successfulMock);
+
+          await party.finishQuest(quest);
+
+          expect(User.update).to.be.calledThrice;
+        });
+
+        it('retries failed updates at most five times per user', async () => {
+          sandbox.stub(User, 'update').returns(failedMock);
+
+          await party.finishQuest(quest);
+
+          expect(User.update.callCount).to.eql(10);
+        });
       });
 
       it('gives out achievements', async () => {
@@ -1171,6 +1208,7 @@ describe('Group Model', () => {
 
       context('Party quests', () => {
         it('updates participating members with rewards', async () => {
+          sandbox.spy(User, 'update');
           await party.finishQuest(quest);
 
           expect(User.update).to.be.calledTwice;
@@ -1205,6 +1243,7 @@ describe('Group Model', () => {
         });
 
         it('updates all users with rewards', async () => {
+          sandbox.spy(User, 'update');
           await party.finishQuest(tavernQuest);
 
           expect(User.update).to.be.calledOnce;
--- a/test/api/v3/unit/models/group.test.js
+++ b/test/api/v3/unit/models/group.test.js
@@ -1065,13 +1065,13 @@ describe('Group Model', () => {
 
       describe('user update retry failures', () => {
         let successfulMock = {
-          exec: (callback) => {
-            callback(null, 'sucess');
+          exec: () => {
+            return Promise.resolve({raw: 'sucess'});
           },
         };
         let failedMock = {
-          exec: (callback) => {
-            callback('error', null);
+          exec: () => {
+            return Promise.reject('err');
           },
         };
 
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -586,7 +586,7 @@ async function _updateUserWithRetries (userId, updates, numTry = 1) {
       if (numTry < MAX_UPDATE_RETRIES) {
         return _updateUserWithRetries(userId, updates, ++numTry);
       } else {
-        return err;
+        throw err;
       }
     });
 }
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -582,7 +582,7 @@ async function _updateUserWithRetries (userId, updates, numTry = 1) {
   return await User.update({_id: userId}, updates).exec()
     .then((raw) => {
       return raw;
-    }, (err) => {
+    }).catch((err) => {
       if (numTry < MAX_UPDATE_RETRIES) {
         return _updateUserWithRetries(userId, updates, ++numTry);
       } else {
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -579,15 +579,16 @@ schema.statics.cleanGroupQuest = function cleanGroupQuest () {
 };
 
 async function _updateUserWithRetries (userId, updates, numTry = 1) {
-  return User.update({_id: userId}, updates).exec((err, raw) => {
-    if (!err) {
+  return await User.update({_id: userId}, updates).exec()
+    .then((raw) => {
       return raw;
-    } else if (numTry < MAX_UPDATE_RETRIES) {
-      return _updateUserWithRetries(userId, updates, ++numTry);
-    } else {
-      return err;
-    }
-  });
+    }, (err) => {
+      if (numTry < MAX_UPDATE_RETRIES) {
+        return _updateUserWithRetries(userId, updates, ++numTry);
+      } else {
+        return err;
+      }
+    });
 }
 
 // Participants: Grant rewards & achievements, finish quest.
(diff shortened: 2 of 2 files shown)


website/server/models/group.js
```diff
@@ -623,11 +623,45 @@ schema.methods.finishQuest = async function finishQuest (quest) {
...
   });
 
-  let q = this._id === TAVERN_ID ? {} : {_id: {$in: this.getParticipatingQuestMembers()}};
+  let participants = this._id === TAVERN_ID ? {} : this.getParticipatingQuestMembers();
   this.quest = {};
   this.markModified('quest');
 
-  return await User.update(q, updates, {multi: true}).exec();
+  if (this._id === TAVERN_ID) {
+    return await User.update(participants, updates, {multi: true}).exec();
```
crookedneighbor: If we're going to return early here for tavern quests, we don't need the ternary above.

crookedneighbor (changes requested):
I don't see any tests around the retry logic. 

This whole thing makes me nervous. To be honest, if there's an error when saving the user, I'd prefer to be notified avout the error so we can actually figure out what is casusing this. 

If we must do retries, I'd prefer them to be a set number, rather than a time period.

Hus274:
> I don't see any tests around the retry logic.

I did some manual testing and it worked as expected. I agree the behavior should have automated tests around it but I'm not familiar with a way to do that here. Would there be a way to mock out the mongo call and have it return an error on the first call and succeed the second time?

> This whole thing makes me nervous. To be honest, if there's an error when saving the user, I'd prefer to be notified about the error so we can actually figure out what is causing this.

From my understanding of the problem described in the issue, it sounds like it's believed that this is due to load and timeouts on the database which is generally handled with retries and spacing the calls out. Is there a way to log the error if it happens the first time so it can be looked up? Otherwise, if we don't want to do retries, I can do what @Alys suggested and simply write it to the groups object.

>If we must do retries, I'd prefer them to be a set number, rather than a time period.

Yeah that makes sense to me. I used time cause it allows you to control the total time you're willing to spend retrying across any number of failed users but I'm ok with switching this over.

Hus274:
I refactored this to have the retries use the same code path as the original mongoose call meaning that the retry logic code path is now tested. I've also updated the retries to be count based instead of time based. Let me know your thoughts now with these changes.

website/server/models/group.js
```diff
@@ -623,11 +642,20 @@ schema.methods.finishQuest = async function finishQuest (quest) {
     }
   });
 
-  let q = this._id === TAVERN_ID ? {} : {_id: {$in: this.getParticipatingQuestMembers()}};
+  let participants = this.getParticipatingQuestMembers();
```
crookedneighbor: This variable should be after the tavern logic.

website/server/models/group.js
```diff
@@ -623,11 +642,20 @@ schema.methods.finishQuest = async function finishQuest (quest) {
...
   this.quest = {};
   this.markModified('quest');
 
-  return await User.update(q, updates, {multi: true}).exec();
+  if (this._id === TAVERN_ID) {
+    return await User.update({}, updates, {multi: true}).exec();
+  }
+
+  let promises = [];
+  _.each(participants, userId => {
```
crookedneighbor: I don't remember, is participants an array? If so, why not use map?

```js
let promises = participants.map(userId => _updateUserWithRetries(userId, updates))

File: test/api/v3/unit/models/group.test.js [hunk 4]
```diff
@@ -1173,11 +1173,12 @@ describe('Group Model', () => {
         it('updates participating members with rewards', async () => {
           await party.finishQuest(quest);
 
-          expect(User.update).to.be.calledOnce;
```
crookedneighbor, line 1176: There's still no test for the retry logic. You can use the sinon spy library to provide the results you need for failures to update.

website/server/models/group.js
```diff
@@ -577,6 +577,25 @@ schema.statics.cleanGroupQuest = function cleanGroupQuest () {
   };
 };
 
+async function _updateUserWithRetries (userId, updates) {
+  let data = {
+    saved: false,
+    retries: 0,
+  };
+  while (!data.saved && data.retries < 5) {
```
crookedneighbor: Extract 5 into a variable like 

```js 
const MAX_UPDATE_RETRIES = 5;
``` 

So we don't have a magic number.

website/server/models/group.js
```diff
@@ -577,6 +577,25 @@ schema.statics.cleanGroupQuest = function cleanGroupQuest () {
...
+    await User.update({_id: userId}, updates).exec((err, raw) => { // eslint-disable-line babel/no-await-in-loop
```
crookedneighbor: I'm not confident that this will even work in a loop, to be honest.

test/api/v3/unit/models/group.test.js
Hus274: Makes sense, thanks for the pointer! I've updated this to include tests with no failures, test with one failure followed by success, and then test for failing infinitely. Let me know if there are any other tests you think I should add.

website/server/models/group.js
Hus274: Good call, updated.

website/server/models/group.js
Hus274: I agree this was kind of messy and made the scopes hard to follow. I've updated this to a recursive method that I think is much cleaner.

website/server/models/group.js
Hus274: Makes sense, I don't know why I never think to use the map function in javascript. I've updated this.

website/server/models/group.js
Hus274: The problem is that participants depends on this.quest not being empty which happens on the following line and before the tavern logic. Because this.quest = {}; needs to be before the tavern logic and participants needs to be before that, I think this needs to remain above the tavern logic. Let me know if you have other thoughts here or if I'm misinterpreting something.

Hus274:
I've updated this according to your review. I've added tests around the retry logic and moved the retry logic to a recursive method as it seems to be a good bit cleaner. Let me know any more feedback you have

website/server/models/group.js
```diff
@@ -650,9 +644,8 @@ schema.methods.finishQuest = async function finishQuest (quest) {
     return await User.update({}, updates, {multi: true}).exec();
   }
 
-  let promises = [];
-  _.each(participants, userId => {
-    promises.push(_updateUserWithRetries(userId, updates));
+  let promises = _.map(participants, userId => {
```
crookedneighbor: Is there a reason to use lodash over the native implementation of map?

website/server/models/group.js
crookedneighbor: Hmm. I think this will not work with tavern quests then. I'd have to look at the getParticipatingQuestMembers method again, because I'm not sure how it interacts with a tavern quest.

website/server/models/group.js
```diff
@@ -577,6 +578,18 @@ schema.statics.cleanGroupQuest = function cleanGroupQuest () {
   };
 };
 
+async function _updateUserWithRetries (userId, updates, numTry = 1) {
+  return User.update({_id: userId}, updates).exec((err, raw) => {
```
crookedneighbor: Is there a reason to use the callback syntax instead of the promise syntax? If I'm reading this correctly, returning these values isn't actually doing anything.

website/server/models/group.js
Hus274: That's why I left the ternary operator in the beginning because I wasn't sure of the impact of calling this with a tavern quest. The getParticipatingMembers method depends on this.quest.members which I'm not sure the state of for tavern quests as I don't know how they are started or populated.

I can add the ternary operation back or make the getParticipatingQuestMembers method explicitly handle tavern quests to return an empty object. What do you think?

website/server/models/group.js
Hus274: Sorry, this is just due to my lack of experience with Javascript. I've adjusted it to the promise syntax now and it's way better. I'm not using the .catch() method because it seems it doesn't work well with mongoose's promise library and so instead I'm using the .then(function(success), function(error)) syntax. Let me know if that works for you.

website/server/models/group.js
```diff
@@ -623,11 +636,19 @@ schema.methods.finishQuest = async function finishQuest (quest) {
...
+  let participants = this.getParticipatingQuestMembers();
   this.quest = {};
   this.markModified('quest');
 
-  return await User.update(q, updates, {multi: true}).exec();
+  if (this._id === TAVERN_ID) {
+    return await User.update({}, updates, {multi: true}).exec();
+  }
+
+  let promises = _.map(participants, userId => {
```
Hus274: Nope, I just thought the standard was to use lodash functions. My mistake, I've fixed this.

website/server/models/group.js
Hus274: Nope, I just thought the standard was to use lodash functions. My mistake, I've fixed this.

website/server/models/group.js
Hus274: http://stackoverflow.com/questions/30672265/why-can-i-not-chain-catch-when-calling-mongoose-model-create-in-node

Hus274:
I moved the getParticipatingQuestMembers back to a ternary operator for now and addressed your other comments. I can definitely re-remove the ternary operator if you want, just let me know. Thanks for taking all the time to review this, I appreciate it.

website/server/models/group.js
```diff
@@ -577,6 +578,19 @@ schema.statics.cleanGroupQuest = function cleanGroupQuest () {
   };
 };
 
+async function _updateUserWithRetries (userId, updates, numTry = 1) {
+  return await User.update({_id: userId}, updates).exec()
+    .then((raw) => {
```
paglias, line 583: you should be able to use catch, it's used in multiple places with mongoose. Because we're plugging Bluebird in place of their promise implementation that doesn't support it

paglias (changes requested):

website/server/models/group.js
Hus274: Ahh ok that makes sense. I'll update it, thanks!

Hus274:
Updated this to use .catch() functionality, thanks @paglias!

website/server/models/group.js
```diff
@@ -577,6 +578,19 @@ schema.statics.cleanGroupQuest = function cleanGroupQuest () {
...
+      return raw;
+    }).catch((err) => {
+      if (numTry < MAX_UPDATE_RETRIES) {
+        return _updateUserWithRetries(userId, updates, ++numTry);
+      } else {
+        return err;
```
paglias: here you should throw err, not return it, otherwise it'll be treated as a success return value

paglias (changes requested):

Hus274:
Updated to throw an error instead of returning it.

paglias (approved):

crookedneighbor (approved):

PR #8256, opened 2016-12-05
Adds a free-text filter (search) to the equipment page.

### LINKED ISSUE #8241: Equipment Page Needs Search Function
Despite the sorting options, the Equipment page still gets quite unwieldy. There should be a way to easily search for a specific piece of equipment.

a2lin:
Fixes #8241 

### Changes
Added a search bar to the equipment page to allow filtering of items. The search filter is implemented as a regex match on name.

Photos:
search-full
search-500px
search-cn

---
UUID: 7b8d3ea1-d46a-4e4e-8927-0dcaca995e36

### DIFF
--- a/website/client-old/js/controllers/inventoryCtrl.js
+++ b/website/client-old/js/controllers/inventoryCtrl.js
@@ -8,6 +8,7 @@ habitrpg.controller("InventoryCtrl",
 
     $scope.selectedEgg = null; // {index: 1, name: "Tiger", value: 5}
     $scope.selectedPotion = null; // {index: 5, name: "Red", value: 3}
+    $scope.equipmentFilterQuery = {'query': ''};
 
     _updateDropAnimalCount(user.items);
 
@@ -85,7 +86,7 @@ habitrpg.controller("InventoryCtrl",
       })
     }, true);
 
-    var equipmentSearch = function(equipment, term) {
+    $scope.equipmentSearch = function(equipment, term) {
       if (!equipment) return;
       if (!angular.isString(term) || term.length == 0) {
         return equipment;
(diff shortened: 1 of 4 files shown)


Alys:
I've tested this locally and it works well! It doesn't seem to interfere with any of the other functions on the page. I like it.

I'm not good enough with Angular to comment on the quality of the code or the tests needed, but other admins are. If this hasn't been commented on by any of them in about a week, I'll merge it.

Copying your question about tests to here from the issue:
"**If the solution seems acceptable i'd appreciate a pointer to what tests are needed for something like this.**"

website/client-old/js/controllers/inventoryCtrl.js
```diff
@@ -420,3 +456,12 @@ habitrpg.controller("InventoryCtrl",
     }
   }
 ]);
+
+habitrpg.controller("InventorySearchCtrl",
+  ['$scope', 'User',
+  function($scope, User) {
+    $scope.updateEquipmentQuery = function() {
+      User.user.equipmentQuery = $scope.equipmentQuery;
```
TheHollidayInn: Why are you attaching the equipment query to the user?

TheHollidayInn:
Looks good to me. Technically, you would want to turn this into a directive and load your controller via the directive. 

For testing: Here is the InventoryCtrlSpec which has some good examples: https://github.com/HabitRPG/habitica/blob/develop/test/client-old/spec/controllers/inventoryCtrlSpec.js .

Essentially, you would want to:

1. populated 'gearByClass', 'gearByType', 'user.equipmentQuery'
2. Change the values
3. Check that $scope.filteredGearByClass and $scope.filteredGearByType have the correct values

Then, you should test that equipmentSearch will take a list of equipment and a term then return the filtered equipment.

a2lin:
@TheHollidayInn I've fixed the issue with the User object; I made an object to hold the query in the InventoryCtrl, which I think makes more sense.

Can you explain a bit more about what you expect with regards to turning it into a directive?

Thanks for reviewing!

TheHollidayInn:
Looking good. Here is a good article: https://teropa.info/blog/2015/10/18/refactoring-angular-apps-to-components.html#replace-ng-controller-with-component-directive

Read the section 'Replace ng-controller with Component Directive'

a2lin:
@TheHollidayInn 
I actually seem to have discovered that no controller is needed, and have removed it. Is this better?

TheHollidayInn:
Yes, this looks good. Let me test locally later tonight, and then I will merge.

TheHollidayInn:
Noted towards your next Tier. Congrats on Tier 1!

a2lin:
Thanks for the CR and tips!

Alys:
@a2lin Finding my new Mystery Items in my huge Equipment list was SO EASY THIS MONTH!!! Thank you! :heart:

PR #8257, opened 2016-12-06
fix: prevents blank messages from being posted to chat

### LINKED ISSUE #8255: Blank messages can be posted to chat
### Description

I posted this to the Report a Bug guild:

> "The Android app allows blank messages to be posted to chat (well, it posts a blank message if you post a space). Should that be reported via the app? (The website doesn't allow a message with only a space in it to post to a chat... though it does allow a message with nothing but `&nbsp;`.)"

I tested it out after seeing a blank message in the Tavern this morning; the app won't post a message if you leave the field blank but will post a blank message if it includes a space character. This doesn't happen on the website (except in the case I mentioned above with `&nbsp;`). I don't know if it also happens on the iOS app.

@Alys replied to me with this:

> "@citrusella, the serving wench : It's probably best that we prevent blank messages being posted in the API, then they're fixed for all interfaces, including third-party tools, so that means we'd make an issue on the normal GitHub repo instead of reporting in the app. Would you like to copy your message and this reply to GitHub? If not, I'll do it later."

Hus274:

Fixes #8255 

### Changes
I've updated the server to reject requests to postChat when the message from the user contains only whitespace. This should prevent chat messages from being posted via the mobile or third party apps that do not have any text.

### DIFF
--- a/test/api/v3/integration/chat/POST-chat.test.js
+++ b/test/api/v3/integration/chat/POST-chat.test.js
@@ -35,6 +35,15 @@ describe('POST /chat', () => {
       });
   });
 
+  it('Returns an error when an empty message is provided', async () => {
+    await expect(user.post(`/groups/${groupWithChat._id}/chat`, { message: '    '}))
+      .to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: t('invalidReqParams'),
+      });
+  });
+
   it('Returns an error when group is not found', async () => {
     await expect(user.post('/groups/invalidID/chat', { message: testMessage})).to.eventually.be.rejected.and.eql({
       code: 404,
--- a/test/api/v3/integration/chat/POST-chat.test.js
+++ b/test/api/v3/integration/chat/POST-chat.test.js
@@ -44,6 +44,15 @@ describe('POST /chat', () => {
       });
   });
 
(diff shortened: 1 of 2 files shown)


Alys:
Looks good to me. I'll merge this in a few days if there's no other comments (I'm being cautious because I'm not strong in our test code).

For the record, this won't fix the problem with people deliberately posting blank messages by typing `&nbsp;` but that will be very rare and any repeat offenders can be handled by the moderators, so I definitely don't think it's worth writing code to try to catch that case.

test/api/v3/integration/chat/POST-chat.test.js
```diff
@@ -35,6 +35,15 @@ describe('POST /chat', () => {
       });
   });
 
+  it('Returns an error when an empty message is provided', async () => {
+    await expect(user.post(`/groups/${groupWithChat._id}/chat`, { message: '    '}))
```
crookedneighbor, line 39: Can we add a test for a message with line breaks?

crookedneighbor (commented):
Looks good to me. Just need the test for line breaks.

Hus274:
Good call, I've added a test around messages containing only newlines.

Hus274:
Accidentally had left an `.only` in the new test causing it to fail linting. I've updated it now and it should be good to go.

TheHollidayInn:
Noted towards your tier! Thanks :D

PR #8267, opened 2016-12-08
fix: confirm no user objects reference a group before deleting it when the member count reaches 0

### LINKED ISSUE #7723: non-empty private group (guild/party) can be deleted if memberCount is wrong
If a private group contains multiple members but the memberCount is 1, then when one player leaves the group, the group will be deleted (i.e., the group is treated as empty).

This situation arose recently when I was manually creating a private guild for a huge party that needed to be split up, and I added all the members to the guild but forgot to set the memberCount to the correct number, so when I left the guild, the guild was deleted. That was clearly my error, but it's  possible for the memberCount to be inaccurate for other reasons. For example, https://github.com/HabitRPG/habitrpg/issues/7454 and also I imagine if a player joins a group and a timeout happens, the memberCount might not be updated -- a two-party group would then have a memberCount of 1.

Since deletion of a group is a drastic action, resulting in lost chats, quest progress, and group description (which might contain important information), and possibly lost access to challenges, I think it's worth checking for this situation before deleting a group with a memberCount of zero. I.e., the code that deletes an "empty" private guild or party should first search for players that still have that group's ID in their accounts. If the code finds any such players, it should update the memberCount and not delete the group.

I'll wait a day or two for feedback, and if there's no objections or better ideas, I'll mark this as help wanted for that suggested fix.

Hus274:

Fixes #7723 
### Changes
This code queries for any user objects that reference the group that's at risk of being deleted to confirm there are no members remaining before actually deleting the group. This helps remove the risk of accidentally deleting a group due to an inaccurate member count field. I've also added tests around this new behavior.

### DIFF
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -868,9 +868,9 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
     // double check the member count is correct so we don't accidentally delete a group that still has users in it
     let members;
     if (group.type === 'guild') {
-      members = await User.find({guilds: group._id}, '_id');
+      members = await User.find({guilds: group._id}).select('_id').exec();
     } else {
-      members = await User.find({'party._id': group._id}, '_id');
+      members = await User.find({'party._id': group._id}).select('_id').exec();
     }
     _.remove(members, {_id: user._id});
     if (members.length === 0) {
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -835,9 +835,7 @@ schema.statics.tavernBoss = async function tavernBoss (user, progress) {
 
 schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
   let group = this;
-  let update = {
-    $inc: {memberCount: -1},
-  };
+  let update = {};
 
   let challenges = await Challenge.find({
     _id: {$in: user.challenges},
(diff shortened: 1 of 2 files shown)


website/server/models/group.js
```diff
@@ -867,18 +865,30 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all') {
...
-      let query = group.type === 'party' ? {'party._id': group._id} : {guilds: group._id};
-      query._id = {$ne: user._id};
-      let seniorMember = await User.findOne(query).select('_id').exec();
-
-      // could be missing in case of public guild (that can have 0 members) with 1 member who is leaving
-      if (seniorMember) update.$set = {leader: seniorMember._id};
+    // double check the member count is correct so we don't accidentally delete a group that still has users in it
+    let members;
+    if (group.type === 'guild') {
+      members = await User.find({guilds: group._id}, '_id');
```
paglias: .find doesn't return a promise unless you call `.exec` at the end. Also you should use .select to specify which fields to load, it's cleaner

paglias: .find only returns an object that is similar to a promise (has a .then property) but no a full promise

paglias (changes requested):

Hus274:
Updated mongo queries to use better syntax with .select() and .exec()

Hus274:
Looks like the tests are failing for an unrelated reason.

PR #8272, opened 2016-12-09
Adding new user.addNotification method with Mongodb Update

### LINKED ISSUE #8264: Create a method to add user notifications through MongoDB's updates
Right now notifications can be added to users through `user.addNotification` that requires a `.save()`, we should have an equivalent method that adds notifications using `User.update` thus not requiring to load and save the entire user

Hus274:

Fixes #8264 

### Changes
Added a new method to the user schema that uses mongo's update functionality to push the notification to the database.

Currently the method is called addNotificationUpdate which I don't love. If you have a suggestion let me know. Other things I thought of were 
* addNotificationAndSave
* pushNewNotification
* updateNotificationsWithNew

Let me know what you think.

----


test/api/v3/unit/models/user.test.js
```diff
@@ -60,7 +60,22 @@ describe('User Model', () => {
       expect(userToJSON.notifications[0].data).to.eql({});
     });
 
-    it('can add notifications without data', () => {
+    it('can add notifications using update', async() => {
```
khipkin: Suggestion: "without data", just for clarity's sake.

Hus274: yep, meant to have that, updated.

website/server/models/user/methods.js
```diff
@@ -42,6 +42,15 @@ schema.methods.addNotification = function addUserNotification (type, data = {})
   });
 };
 
+schema.methods.addNotificationUpdate = async function addUserNotificationUpdate (type, data = {}) {
+  let newNotification = {
+    type,
+    data,
+  };
+  this.notifications.push(newNotification);
```
paglias: What we want to do with this method is to provide a way to add notifications without having to save the user. So this should probably be a static method of the user model (User.statics.addUserNotificationUpdate) that takes the notification data and one or more ids of users. If you issue the update command and add the data to the user doc (this.notifications.push) the next time the user is saved, the notification will be added again to it.

paglias (changes requested):

TheHollidayInn:
I suggest pushNotification. The function doesn't need to detail anything about the internals like your other options do.

website/server/models/user/methods.js
TheHollidayInn: Yea, you don't need to push this at all. Your update query doesn't use it.

Hus274: Yeah that was a mistake on my part. I was thinking about keeping the local version in sync with the database version and forgot that the database won't handle the deduping on a later save.

Hus274:
I believe I've made all the changes that were requested here. Let me know if you have any more feedback!

website/server/models/user/methods.js
```diff
@@ -35,13 +35,38 @@ schema.methods.sendMessage = async function sendMessage (userToReceiveMessage, m
   await Bluebird.all(promises);
 };
 
+/**
+ * Creates a notification and based on the input parameters and adds it to the local user notifications array.
```
TheHollidayInn: Change "Creates a notification and based" to "Creates a notification based"

website/server/models/user/methods.js
```diff
@@ -35,13 +35,38 @@ schema.methods.sendMessage = async function sendMessage (userToReceiveMessage, m
...
     type,
     data,
   });
 };
 
+/**
+ * Creates a notification based on the type and data input parameters and saves that new notification
+ * to the database directly using an update statement. The local copy of these users are not updated by
+ * this operation.
+ *
```
TheHollidayInn, line 68: Can you add a note for use case on this? For example: "Use this function when you want to add a notification to a user(s), but do not have the document(s) opened.

Hus274: Yeah that's a good call, I'll update it.

website/server/models/user/methods.js
Hus274: Good catch!

Hus274:
Updated the documentation according to the comments here

website/server/models/user/methods.js
```diff
@@ -35,13 +35,39 @@ schema.methods.sendMessage = async function sendMessage (userToReceiveMessage, m
...
+ * @param  userIds An array containing the ids of the users to add this notification to
+ * @param  type  The type of notification to add to the user. Possible values are defined in the UserNotificaiton Schema
+ * @param  data  The data to add to the notification
+ */
+schema.statics.pushNotification = async function pushNotification (userIds, type, data = {}) {
+  let newNotification = new UserNotification({type, data});
+  let promises = [];
+  userIds.forEach(userId => {
+    promises.push(this.update({_id: userId}, {$push: {notifications: newNotification}}).exec());
```
paglias: @Hus274 I think the only missing change here is that instead of updating each user separately we should use `{_id: {$in: userIds}}` so only one update command is used.

paglias (changes requested):

website/server/models/user/methods.js
```diff
@@ -35,13 +35,39 @@ schema.methods.sendMessage = async function sendMessage (userToReceiveMessage, m
...
+ * to the database directly using an update statement. The local copy of these users are not updated by
+ * this operation. Use this function when you want to add a notification to a user(s), but do not have
+ * the user document(s) opened.
+ *
+ * @param  userIds An array containing the ids of the users to add this notification to
+ * @param  type  The type of notification to add to the user. Possible values are defined in the UserNotificaiton Schema
+ * @param  data  The data to add to the notification
+ */
+schema.statics.pushNotification = async function pushNotification (userIds, type, data = {}) {
+  let newNotification = new UserNotification({type, data});
```
paglias, line 74: let's validate this to be sure all values are valid, `newNotification.validateSync()`, it returns `undefined` if the doc is valid or the error if it's not, http://mongoosejs.com/docs/api.html#document_Document-validateSync

website/server/models/user/methods.js
paglias: and to allow a single id to be passed, so the userIds param can be either an array of ids or a single user id string

paglias (changes requested):

Hus274:
Switched this to use the multi update functionality so that a single userId or a list of user Ids can be used. Also added in mongoose validation before saving. Let me know if anything else should be updated here!

paglias:
@Hus274 just wanted to say that we'll merge this soon. We're trying to debug some issues with notifications and would prefer to fix that before merging new code.

Hus274:
@paglias sounds good, that makes perfect sense. Thanks for letting me know!

TheHollidayInn:
@paglias I believe we have resolved the Notification issue, correct. Should we merge this PR in now?

website/server/models/user/methods.js
```diff
@@ -35,13 +35,39 @@ schema.methods.sendMessage = async function sendMessage (userToReceiveMessage, m
...
+ * Creates a notification based on the type and data input parameters and saves that new notification
+ * to the database directly using an update statement. The local copy of these users are not updated by
+ * this operation. Use this function when you want to add a notification to a user(s), but do not have
+ * the user document(s) opened.
+ *
+ * @param  userIds An array containing the ids of the users to add this notification to
+ * @param  type  The type of notification to add to the user. Possible values are defined in the UserNotificaiton Schema
+ * @param  data  The data to add to the notification
+ */
+schema.statics.pushNotification = async function pushNotification (userIds, type, data = {}) {
```
paglias: since we haven't yet merged this, would it be possible to change the function to allow an entire query to be passed instead of ids only? It'll make the code much easier to use in situations where we want to target users based on some properties instead of ids. I'll do it tomorrow otherwise

paglias (changes requested):

website/server/models/user/methods.js
Hus274: yeah, I should be able to make this change today.

Hus274:
@paglias updated to have the user pass a query instead of an array of user ids.

paglias:
Thanks! Notated towards your next tier, and welcome to tier 6 :)

Hus274:
@paglias that's awesome!!! Thank you!!!

PR #8278, opened 2016-12-12
(WIP) Thehollidayin/front end updates

TheHollidayInn:

Fixes put_issue_url_here

### Changes

----
UUID:


website/client/router.js
```diff
@@ -26,7 +29,15 @@ export default new VueRouter({
       component: ParentPage,
       children: [
         {path: 'tavern', component: Page},
-        {path: 'inbox', component: Page},
+        {
+          path: 'inbox',
+          component: InboxPage,
+        },
+        {
```
TheHollidayInn: I'm pretty sure this should be a child route, but the child routes seem to keep the html of the parent. Still looking into this.

paglias: I think, but double check, that the solution here is to have a main `/inbox` route with two (or more) children routes, the first one having `''` (empty path) as its `path` that will be loaded when the main `/inbox` route is loaded.

Something like this:

- Parent route with shared html `/inbox`
    - Main child route with path `''` with the html to be shown only in the parent route but not in the child routes
    - Other child routes

website/client/components/Inbox/InboxConversationPage.vue
```diff
@@ -0,0 +1,58 @@
...
+    reply: function reply () {
+      // @TODO: This will be default values based on the conversation and current user
+      let messageToSend = {
+        from: 'TheHollidayInn',
+        fromUserId: 3211,
+        to: 'Paglias',
+        message: this.newMessage,
+        date: new Date(),
+      };
+      this.messages.push(messageToSend);
```
paglias, line 53: this logic should become an action inside the store

website/client/components/Inbox/InboxConversationPage.vue
```diff
@@ -0,0 +1,58 @@
...
+<script>
+export default {
+  data () {
+    // @TODO: Abstract to Store
+    let messages = [
+      {
+        from: 'Paglias',
+        fromUserId: 1234,
+        to: 'TheHollidayInn',
+        message: 'I love the Gang of Four',
```
paglias, line 32: this is psychological terrorism 😛

website/client/components/Inbox/InboxConversationPage.vue
TheHollidayInn: Yea, added note at the top :D

PR #8279, opened 2016-12-12
Moderator hall and previously blocked users

### LINKED ISSUE #8254: moderator Hall bug: adjusting a previously-blocked user causes them to rest in the inn
When a moderator uses Social > Hall, they have extra features there for modifying user accounts (e.g., to give pets or gems). When they use those features to make any change to a user's account, if the user has ever been blocked in the past, the change that the moderator makes causes the user to be resting in the inn (not blocked, just resting).

This is because of this code: https://github.com/HabitRPG/habitica/blob/9aa4b8aa64ccdfa3af4a6832cc3dded4b102b73e/website/server/controllers/api-v3/hall.js#L180-L183

The code should be checking if auth.blocked is true, not if it's a boolean (which can also mean false).

The test should be changed too: https://github.com/HabitRPG/habitica/blob/9aa4b8aa64ccdfa3af4a6832cc3dded4b102b73e/test/api/v3/integration/hall/PUT-hall_heores_heroId.test.js#L43
It shouldn't be checking blocking + sleep at the same time as the other changes.

I suggest that the tests should be:
+ Block a user > They become blocked and asleep.
+ Unblock a user > The stop being blocked. They remain asleep.
+ Change contributor level, balance, ads > Those items change. The user is not blocked. The user is not asleep.

_(FYI for anyone curious: The "ads" code and associated user data are left over from a time when we had small ads on the site pointing to things like books about productivity. There are no plans to reintroduce ads.)_

mdakan:

Fixes https://github.com/HabitRPG/habitica/issues/8254

### Changes

Changed the logic for moderators updating user accounts so that previously blocked users are not put to sleep.
Changed three tests to reflect this:
- change contributor level, balance, ads: No longer checks blocking at the same time. Checks if block/sleep haven't been changed.
- block a user: checks that users are blocked and asleep
- unblock a user: checks that users are unblocked and don't changed sleep state

This last test fails and I need some feedback on how to fix it. I get an AssertionError that in line 99 heroRes.auth.blocked is undefined rather than the boolean that I expect it to be. I'm particularly confused as to why this test fails and the block a user test does not.

----
UUID: e48c8f5d-6ec9-49a0-9713-354447be2d17


website/server/controllers/api-v3/hall.js
```diff
@@ -177,8 +177,8 @@ api.updateHero = {
       _.set(hero, updateData.itemPath, updateData.itemVal); // Sanitization at 5c30944 (deemed unnecessary)
     }
 
-    if (updateData.auth && _.isBoolean(updateData.auth.blocked)) {
-      hero.auth.blocked = updateData.auth.blocked;
+    if (updateData.auth && _.isBoolean(updateData.auth.blocked) && updateData.auth.blocked === true) {
```
haroldtreen: It looks like this line could be:

```js
if (updateData.auth && updateData.auth.blocked === true) {
// ...
}
```

All the cases where `_.isBoolean(updateData.auth.blocked)` is `false`, will also be caught by `updateData.auth.blocked === true`.

haroldtreen: If the hero was already blocked, do you still want to change their sleep state?
This just checks what update was sent, but not what the current state of the hero is. 

(Maybe not, just a thought!)

TheHollidayInn:
@mylesdakan Do you have an update on this PR? Looks like you have a failing test and I agree with the style change above if you want to add that in as well.

mdakan:
Ack, sorry for adding all these irrelevant commits. I'll open a new PR.

PR #8297, opened 2016-12-18
Docker improvements

### LINKED ISSUE #8288: Habitica does not build as a docker container after upgrade to npm@4
General Info
  * UUID: 30983c37-52c3-415d-977c-38af6c7db91c
  * OS: docker

### Description
Because each line of a Dockerfile runs as a separate container in cache, the environment that installs npm@4 does not carry over to subsequent commands, resulting in a failure of npm commands.

#### Console Errors
    Step 11 : RUN npm install -g gulp grunt-cli bower
     ---> Running in 1a82edfe5612
    module.js:471
        throw err;
        ^

    Error: Cannot find module 'semver'
        at Function.Module._resolveFilename (module.js:469:15)
        at Function.Module._load (module.js:417:25)
        at Module.require (module.js:497:17)
        at require (internal/module.js:20:19)
        at Object.<anonymous> (/usr/lib/node_modules/npm/lib/utils/unsupported.js:2:14)
        at Module._compile (module.js:570:32)
        at Object.Module._extensions..js (module.js:579:10)
        at Module.load (module.js:487:32)
        at tryModuleLoad (module.js:446:12)
        at Function.Module._load (module.js:438:3)
    The command '/bin/sh -c npm install -g gulp grunt-cli bower' returned a non-zero code: 1

DrStrangepork:
#8288

### Changes
Fixes #8288 
Dockerfile was missing node module `mocha`
Configuration/Documentation updates

----
UUID: 30983c37-52c3-415d-977c-38af6c7db91c


.dockerignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+.git
+website
```
paglias, line 3: why was this ignored but not other project folders? Like /gulp, /webpack, ...

bower.json
```diff
@@ -1,10 +1,19 @@
 {
-  "name": "HabitRPG",
-  "version": "0.1.1",
```
paglias, line 3: isn't the version number required? Also I'd just skip changes to bower.json, thanks but we're going to ditch it with the redesign that is in progress

paglias (changes requested):

.dockerignore
DrStrangepork: I added what I knew was not needed, and this list resulted in a significant speed increase, so I felt it safe to stop there. If there are other folders that you know are also unneeded, I'd be happy to add them.

paglias: @DrStrangepork I have to say I'm no Docker expert but website is the folder that contains the entire source code, why isn't it needed?

bower.json
DrStrangepork: According to https://github.com/bower/spec/blob/master/json.md, version has been deprecated.
But if bower has a timeline itself to be deprecated in Habitica, then I will bower.json out of the PR entirely.

.dockerignore
DrStrangepork: The .dockerignore file manages the build context that gets sent to the docker service during the build process. In Habitica's case, we're building a stand-alone image that pulls its own`git clone` of Habitica, builds it and serves it, so there is little that _must_ be passed to the docker service (theoretically, everything except Dockerfile could be excluded). The /website directory gets created within the image as part of its local Habitica repo and build process, so it doesn't need to be passed on to docker.

DrStrangepork: In my local install, the 3 directories included consist of >98% of the total size of my habitica folder, so there would be very little speed increase in adding any additional folders.

paglias: If only Dockerfile is needed what about

```
# Ignore everything
*

# But not Dockerfile
!Dockerfile
```

DrStrangepork: I can test that out

DrStrangepork: `docker-compose up` with above `.dockerignore`:

    real	1m27.618s
    user	0m23.492s
    sys	0m9.614s

`docker-compose up` with `.dockerignore` from this PR:

    real	0m12.910s
    user	0m0.856s
    sys	0m0.414s

paglias: Sounds good!

paglias:
Notated towards your next tier

PR #8319, opened 2016-12-27
Add 'Save & Close' and 'Cancel' buttons to top of task edit modal

### LINKED ISSUE #8316: add "Save & Close" and "Cancel" buttons to the top of the task edit screen
As described in https://github.com/HabitRPG/habitica/issues/8307, now that task editing is done in a modal, reaching the "Save & Close" and "Cancel" buttons requires scrolling to the bottom, which is especially inconvenient on small screens or when the tags and advanced options are expanded.

We should add "Save & Close" and "Cancel" buttons at the top of the modal (as well as having them at the bottom).

My personal suggestion is to add the two buttons in a horizontal row at the top right of the modal, however other suggestions are welcome.

I believe this is definitely worth doing now for usability, but we're currently rewriting the website's front end to make many significant improvements, so it's not worth putting too much time into making the buttons look perfect. I think a quick fix now for usability would be fine.

maximumtiu:

Fixes #8316 

### Changes
Implements 'Save & Close' and 'Cancel' buttons at the top of the edit task modal to prevent users from having to scroll all the way to the bottom of the modal to use these buttons. I found that having the two buttons centered inline looked best, but feedback is welcome.

top

I've also adjusted the styles of the buttons at the bottom of the modal to be consistent with the top buttons.

bottom

----
UUID: 
bd794f6e-3845-49e5-88cc-a140b24cdea4

### DIFF
--- a/website/client-old/css/tasks.styl
+++ b/website/client-old/css/tasks.styl
@@ -556,7 +556,11 @@ form
         min-width: 2.5em
   .save-close
     text-align: center
+    display: inline
+    padding: 0 0.5em
     @extend $hrpg-button
+  h2#task-edit-title
+    margin-top: 1em
 
 // Dailies
 .dailies
--- a/website/views/shared/tasks/edit/index.jade
+++ b/website/views/shared/tasks/edit/index.jade
@@ -1,7 +1,14 @@
 div(ng-if='task._editing')
   .task-options
-    h2 {{task._edit.text}}
-  
+    .save-close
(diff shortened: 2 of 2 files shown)


TheHollidayInn:
Looks good to me! I'll give other's a chance to look, but I'll merge it tomorrow if no one objects.

TheHollidayInn:
Noted towards your tier. Congrats on Tier 1!

PR #8323, opened 2016-12-28
Add 'balanceGemAmount' to Amplitude #8057

### LINKED ISSUE #8057: Amplitude should record the user's balance (amount of gems)

cailinglu:

Fixes #8057

### Changes
- Added a 'balanceGemAmount' property
- ~~'balanceGemAmount' is equal to the 'balance' property divided by 4~~
- 'balanceGemAmount' is equal to the 'balance' property times by 4
- Added a check for the correct 'balanceGemAmount' in api unit test and client-old services test

----

### DIFF
--- a/test/api/v3/unit/libs/analyticsService.test.js
+++ b/test/api/v3/unit/libs/analyticsService.test.js
@@ -303,7 +303,7 @@ describe('analyticsService', () => {
                 contributorLevel: 1,
                 subscription: 'foo-plan',
                 balance: 12,
-                balanceGemAmount: 3,
+                balanceGemAmount: 48,
                 loginIncentives: 1,
               },
             });
--- a/test/api/v3/unit/libs/analyticsService.test.js
+++ b/test/api/v3/unit/libs/analyticsService.test.js
@@ -303,6 +303,7 @@ describe('analyticsService', () => {
                 contributorLevel: 1,
                 subscription: 'foo-plan',
                 balance: 12,
+                balanceGemAmount: 3,
                 loginIncentives: 1,
               },
             });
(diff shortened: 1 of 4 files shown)


TheHollidayInn:
@SabreCat Look good to you?

SabreCat:
I believe it should be _times_ 4, not _divided by_ 4. The `balance` property is in USD (leaving aside any discounts like the bonus Gems from buying on mobile), so to get a count of Gems, you need to multiply the user's balance by 4. 

`balance` of 5 means 20 Gems
`balance` of 0.25 means 1 Gem

Under the PR as written, a `balance` of 5 will give us 1.25 Gems.

TheHollidayInn:
Added to your Tier progress! Congratz on Tier 1 :D

PR #8334, opened 2016-12-30
Fixes pressing enter to confirm character choice in IME creates new checklist line

### LINKED ISSUE #8326: Pressing enter to confirm character choice in IME creates new checklist line
_edit by Alys 2020-01-14:_

For the current status of this issue, see this comment below: https://github.com/HabitRPG/habitica/issues/8326#issuecomment-574092129

The fix for this issue was made on the old website so it might have stopped working when the new website was created.

_end of edit_

---

### Description

I was asked to create a new issue here after a discussion in the Report a Bug guild about this; I think the most succinct way I can describe it is a modified version of the long post I created in the guild explaining the issue after I tested what I thought user chan_gami was talking about:

(I'll use Microsoft's IME as it works on Windows 7 as an example, since it's what I have at my disposal that I know how to use, but most IMEs work in similar ways. The images below have tooltips further explaining them when they are hovered over.) When you use an [IME](https://en.wikipedia.org/wiki/Input_method) (a typing method that allows you to use one input format (i.e. a latin keyboard) to type in a different one (i.e. a syllabic/logographic language)) to enter Japanese, you type in kana (hiragana, usually) and it shows up underlined.

えいが with a black squiggly line under it

Then you hit space to use the IME's dictionary to select what kanji you want based on the hiragana/words you typed in. 

映画 with a dropdown under it displaying various kanji choices; the displayed term, 映画, is highlighted as the selection

Enter confirms your choice and sort of "locks in" that character sequence, I guess (you're not editing the kana in a way that you can select kanji for it anymore and the line under the characters disappears, allowing you to start the process over again if you're not finished). 

を has a squiggly line under it but 映画 is no longer underlined

**In checklist editing (but not elsewhere that I can find), hitting enter to confirm the IME choice is also being interpreted as the "enter" that creates a new checklist line**, which isn't ideal if you were just partway through the message since you have to return to the previous line to write more. Here's an example if you're only writing part of a line at a time (I'll go word by word in the examples below (though an IME can technically handle multiple words at a time) to show how it should look vs. how it'd be interpreted for checklists if you didn't return to the original checklist item.)

Here is the sentence in full, along with a table splitting it into chunks based on how I typed it for use in this example:

Full sentence: メアリーさん、映画を見に行きましょう。

Romaji (from QWERTY typing)|meari-san,|eiga|wo|mi|ni|ikimashou.
----|----|----|----|----|----|----
Kana (pre-space-press) |めありーさん、|えいが|を|み|に|いきましょう。
Final characters chosen |メアリーさん、|映画|を|見|に|行きましょう。

Below is how the sentence would look if the enter to confirm the character choices (separated like above) also created a new line:

メアリーさん、  
映画  
を  
見  
に  
行きましょう。

That is to say, if you didn't return to the first checklist item, a new one would be created for each thing you typed and pressed enter to confirm. I'm not sure I've seen that as a usual behavior in most places I've used an IME (granted, most places I've used an IME are language-learning sites). The usual behavior I've seen is that the enter for the IME character choice is ignored (somehow) and only non-IME-choice-confirming enter presses result in the action pressing enter is supposed to achieve.

So the behavior one might expect to see in terms of how IME choice usually functions is that the IME kana-to-kanji-conversion enter press would do nothing and only a typical enter press (i.e. outside of the IME/with intent to create a new line, the way it works in English) would create a new checklist item. But instead, with checklist items on Habitica, currently IME-confirm presses and "normal" presses do it. I think this probably happens across various systems and browsers (and IMEs if they function similarly, even if they're non-Microsoft or non-Japanese, i.e. a Google Chinese/pinyin IME), even though I've only tested on Windows 7 with Microsoft IME in Firefox. (For instance, chan_gami reported it on Mac OS Sierra with Chrome, Firefox, *and* Safari via Google's Japanese IME.)

...It was very confusing to write this (or, well, to modify it to post here), so if something needs to be clarified, please let me know.

changami:
Fixes #8326 

### Changes

I changed the event from "ui-keyup" to "ui-keydown".

First, this issue was happened on OS X El Capitan and macOS Sierra.
The cause of the issue was that Enter pressing event was acquired only for "ui-keyup" **at character conversion**. ("ui-keydown", "ui-keypress" do not work in this case. I think this case should **not** trigger any key-event, including "ui-keyup", like on Windows.)

The behavior of the IME is well explained in that issue by @citrusella.

▼Before
habitica_bug_report

▼After
habitica_bug_report_after

Feedback is welcome, thank you.

----
UUID: 1c980e87-ef12-471b-a812-4a020cc226c1

### DIFF
--- a/website/views/shared/tasks/edit/checklist.jade
+++ b/website/views/shared/tasks/edit/checklist.jade
@@ -18,6 +18,6 @@
           //-,ng-blur='saveTask(task,true)')
           span.checklist-icon.glyphicon.glyphicon-resize-vertical
           input(type='text', ng-model='item.text',
-            ui-keyup="{'13':'addChecklistItem(task,$event,$index)','38 40':'navigateChecklist(task,$index,$event)'}")
+            ui-keydown="{'13':'addChecklistItem(task,$event,$index)','38 40':'navigateChecklist(task,$index,$event)'}")
           a(ng-click='removeChecklistItem(task,$event,$index,true)')
             span.glyphicon.glyphicon-trash(tooltip=env.t('delete'))


TheHollidayInn:
Seems straightforward to me. Thanks for the details!

Noted towards your tier. Congratz on Tier 1!

PR #8353, opened 2017-01-02
Fixes on-hover notes on tasks do not update after you edit the notes, until a sync occurs

### LINKED ISSUE #8352: on-hover notes on tasks do not update after you edit the notes, until a sync occurs
When you edit the notes on a task, close the editing window, and then hover over the task to view the notes, the changes you have just made to the notes do not appear (even though the changes were saved - you can confirm that by opening the edit screen again). You have to sync before the changes will appear in the on-hover notes view.

I think this is a recent change because I've seen it reported only in the past few days (two or three times though and I've confirmed it now). Possibly it started when the new edit modal was introduced.

declanramsay:
Fixes #8352 

### Changes
I removed the one-time-binding syntax from the logic that displays notes, this one-time-binding meant the notes' value was only being computed at initialisation and sync, not reflecting local updates to the task data.

----
UUID: f301baf5-b41b-44f0-8978-ee3e7764b8e1

### DIFF
--- a/website/views/shared/tasks/task.jade
+++ b/website/views/shared/tasks/task.jade
@@ -5,7 +5,7 @@ li(id='task-{{::task._id}}',
   ng-click='spell && (list.type != "reward") && castEnd(task, "task", $event)',
   ng-show='shouldShow(task, list, user.preferences)',
   popover-trigger='mouseenter', popover-placement="top", popover-append-to-body='{{::modal ? "false":"true"}}',
-  data-popover-html="{{::taskPopover(task) | markdown}}")
+  data-popover-html="{{taskPopover(task)}}")
 
   ng-form(name='taskForm')
     include ./meta_controls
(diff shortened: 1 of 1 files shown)


website/views/shared/tasks/task.jade
```diff
@@ -5,7 +5,7 @@ li(id='task-{{::task._id}}',
   ng-click='spell && (list.type != "reward") && castEnd(task, "task", $event)',
   ng-show='shouldShow(task, list, user.preferences)',
   popover-trigger='mouseenter', popover-placement="top", popover-append-to-body='{{::modal ? "false":"true"}}',
-  data-popover-html="{{::taskPopover(task) | markdown}}")
+  data-popover-html="{{taskPopover(task)}}")
```
TheHollidayInn: You will still need to send this through the markdown filter.

declanramsay:
D'oh! Re-added the markdown filter.

TheHollidayInn:
Noted towards your tier! Congratz on Tier 1!!

Also, you should update your author details locally:
```
$ git config --global user.name "John Doe"
$ git config --global user.email johndoe@example.com
```

Issue #8338, opened 2016-12-31
users had profile name change to @profile name not found & "BOSS_DAMAGE" notification errors occur


cTheDragons:
### Description
With the last release a number of users names were change to *profile name not found*. It appears users who are affected also removes their blurb and image associated with them.

With current reports it appears not to be device specific or the users interacting with the API in any way.

Reports from Report a Bug Guild:
***From @Erik Angeles UID: 36904466-cfa0-4743-8f6c-201a1f3b5728***
> After purchasing a Coif, my name changed to profile name not found.

> @cTheDragons I purchased it on my android device. Chain Coif to be specific. Thanks for your quick reply! 

***From @tehkittehkat UID: d84a7025-05d9-4ba6-8f88-9cf55e3bc3bd***
> @cTheDragons I'm having the same problem as @Erik Angeles my name is showing as "profile name not found". This happened after I logged in after a couple of hours away. I hadn't performed any particular action, it was there at login. All members of my party are also showing as "profile name not found" so it's not just an issue with my profile.

> @cTheDragons I'm using the website, but it shows up on my Android app too. Thanks :)

> @cTheDragons Our party ID is ffb062fb-b161-4a1e-84c0-9a3f501fb4e3. No one else in my party is online, so I'm guessing they haven't logged in at all since the issue occurred and I'm the only one who's spotted it. 

One other report from the Tavern with no specific detail. (UID eb72f978-9f6f-430a-adf9-c0f8014642c2)

etaiklein:
![](https://i.imgur.com/eRlKGPI.png)

etaiklein:
Coming across this. https://github.com/HabitRPG/habitica/commit/0b8def555b01fc105368e4edb160838c5401803f
it...  explains why we're named this way but not how it became so. I lol'd.
Still digging.

cTheDragons:
OK I just got hit by said bug. Been switching between the Android app and Party app. I cast a skill last before I notice but it could of occurred before that. 
UID : 95640e89-24f5-41c7-86ab-533d4ad41cc7

Party being casting Snowballs but not at me. Usual buffing activity.

cTheDragons:
Appears 14 users of my 54 member party have been hit. All been active since the code release.

tehkittehkat:
The issue has reoccurred for me (after I fixed it by manually changing
profile name under User>Profile). Also, now my avatar has turned into a
snowman?! Very strange. Android app screens hot attached.

[...]

PR #8357, opened 2017-01-02
Avoid setting profile name to not found

paglias:
The `profile name not found` bug happens when the user is loaded without the `profile` part: when the user is saved the code check for existence of `profile.name` and if it cannot find it (even if it exists but has not been loaded) it set it to (in order): username, google and fb profile names and at last fallbacks to `profile name not found`.

This PR makes sure the `profile.name` always exists by adding a required validator on it that prevents it to be set to an empty string and only set the profile name automatically when the user is created.

### DIFF
--- a/website/server/models/user/hooks.js
+++ b/website/server/models/user/hooks.js
@@ -131,6 +131,16 @@ function _setProfileName (user) {
   return localUsername || _getFacebookName(user.auth.facebook) || googleUsername || anonymous;
 }
 
+schema.pre('validate', function preValidateUser (next) {
+  // Populate new user with profile name, not running in pre('save') because the field
+  // is required and validation fails if it doesn't exists like for new users
+  if (this.isNew) {
+    this.profile.name = _setProfileName(this);
+  }
+
+  next();
+});
+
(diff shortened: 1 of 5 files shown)


paglias:
cc @Alys @SabreCat @TheHollidayInn

website/server/models/user/hooks.js
```diff
@@ -134,7 +134,7 @@ function _setProfileName (user) {
 schema.pre('validate', function preValidateUser (next) {
   // Populate new user with profile name, not running in pre('save') because the field
   // is required and validation fails if it doesn't exists like for new users
-  if (this.isNew) {
+  if (this.isNew && !this.profile.name) {
```
TheHollidayInn, line 137: Do you have a test to ensure the profile doesn't get overwritten if their is a new user with a profile name?

paglias: Right now when a new user is created it cannot have a profile name, it's set automatically to the username or google/fb name. I've only added ` && !this.profile.name` because a lot of tests relied on creating a user with a profile name already set but it's not a behavior that we have on the website

TheHollidayInn:
Fixes #8338

PR #8361, opened 2017-01-03
Always use .exec() for .find*() and .update()

paglias:
As specified at http://mongoosejs.com/docs/promises.html mongoose's `.find*()` and `.update*()` methods don't return real promises until `.exec()` is called. We're not doing that everywhere and it's exposing us to serious problems when using `await` because failures aren't handled correctly.

WIP, for now I've gone through the `/models` folder


paglias:
I've also reviewed all pending PRs for the same problem

PR #8367, opened 2017-01-05
Fixes #8315 - Background of task editing modal is white if you double-click the task

### LINKED ISSUE #8315: Background of task editing modal is white if you double-click the task
**UPDATE**: The issue occurs when you double-click the task to edit it instead of clicking the pencil icon, regardless of browser/platform.

---

In the Report a Bug guild, derpderp (a7d65869-2f12-433a-a111-b4811ab89555) has told us that when editing a task, the background is always white and there is no highlighting to indicate which buttons have been selected (e.g., for due days, difficulty, etc) - see screenshot below. No browser extensions were installed.

This seems to be a problem with Edge, Internet Explorer, and Safari. derpderp writes "the white background persists for me across different browsers (edge, internet explorer) and 3 different computers (all running windows). I asked a friend to check their Habitica, and they reported the same thing: white background with no highlights of days or difficulty. They are on the latest Mac OS on Safari."

I can't replicate this on Chrome or Firefox on Linux.

It's possible that the method of applying the background colour and button highlights isn't supported by all browsers.

white

## What to do if this happens to you

We know this happens equally to everyone, so you do not need to comment here or report it elsewhere. 

You can avoid this problem by clicking on the pencil icon when you want to edit a task, instead of double-clicking on the task.

beejunk:

Fixes https://github.com/HabitRPG/habitica/issues/8315

### Changes

When tasks were double-clicked, task-editing modals would not be given the correct HTML class names, and would therefore use the wrong CSS. This would cause the modals to render with a flat white background, as seen here:

habitica_task_wrong

This is due to the editTask() function being called without the expected taskStatus parameter (see website/client-old/js/controllers/taskCtrl.js:L80). This parameter should be a list of class names as generated by the taskClasses() function exported by website/common/script/libs/taskClasses.js.

This fix simply calls taskClasses(), passing the result directly to editTask(). Most of the parameters to taskClasses() are left as their defaults, since few of them appear to be relevant to rendering the task editing modal. This displays the modal with the appropriate CSS, as seen in this image which shows a task that has been opened with a double-click after the fix has been applied

habitica_task_right

[//]: # 

----
UUID: a654bd1f-e925-438b-8d98-9ac4462b973f

### DIFF
--- a/website/client-old/js/controllers/tasksCtrl.js
+++ b/website/client-old/js/controllers/tasksCtrl.js
@@ -77,7 +77,7 @@ habitrpg.controller("TasksCtrl", ['$scope', '$rootScope', '$location', 'User','N
       if (task._editing) {
         $scope.saveTask(task);
       } else {
-        $scope.editTask(task, User.user);
+        $scope.editTask(task, User.user, Shared.taskClasses(task, [], User.user.preferences.dayStart));
       }
     }
 


TheHollidayInn:
Looks good to me!

TheHollidayInn:
Noted towards your tier. Congratz on Tier 1!

PR #8380, opened 2017-01-07
Fixes #8227 (buying gems with gold gives two notifications when language is not 'en')

### LINKED ISSUE #8227: buying gems with gold gives two notifications when language is not 'en'
If you set your language to some of the non-English languages and then go to the Market to buy a gem with gold, you receive two notifications for `+1 gem`, one in your chosen language and then after a short delay, another one in English:

image

I've tested a few languages and it happens with only those that have a translation for the `plusOneGem`  text that is different than the English version. For example, it happens with German and French, but not with `en_GB`, `en@pirate`, or `zh_TW` all of which still use the default text: `"plusOneGem": "+1 Gem"`

myoshuGO:

Fixes #8227 

### Changes

- Added request language to i18n.t call

Previously, two notifications were shown: one with the correct translation, from the return of the purchase function, and one untranslated, from the API call. When the translated string matched the 'en' string (as in 'en_GB' or 'en@pirate'), only one notification was shown. But if both strings were different, two notifications were shown:

captura de pantalla 2017-01-05 a las 20 06 25

With this modification, the API call returns the translated string, so both string matches and only one notification is shown. Some examples follow: 

For Spanish:

captura de pantalla 2017-01-07 a las 13 43 16

For German:

captura de pantalla 2017-01-07 a las 13 43 46

For French:

captura de pantalla 2017-01-07 a las 13 44 11

For English:

captura de pantalla 2017-01-07 a las 13 44 31

----
UUID: bbbb9c50-f373-4528-ac4d-c1f9925a611d

### DIFF
--- a/test/common/ops/purchase.js
+++ b/test/common/ops/purchase.js
@@ -147,6 +147,15 @@ describe('shared.ops.purchase', () => {
       expect(user.stats.gp).to.equal(goldPoints - planGemLimits.convRate);
     });
 
+    it('purchases gems with a different language than the default', () => {
+      let [, message] = purchase(user, {params: {type: 'gems', key: 'gem'}, language: 'de'});
+
+      expect(message).to.equal(i18n.t('plusOneGem', 'de'));
+      expect(user.balance).to.equal(userGemAmount + 0.5);
+      expect(user.purchased.plan.gemsBought).to.equal(2);
+      expect(user.stats.gp).to.equal(goldPoints - planGemLimits.convRate*2);
+    });
+
     it('purchases eggs', () => {
       let type = 'eggs';
       let key = 'Wolf';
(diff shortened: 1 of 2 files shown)


TheHollidayInn (commented):
Can you add a Test for this?

myoshuGO:
@TheHollidayInn I thought it wasn't necessary since there is already a test for a successful gem purchase message in test/common/ops/purchase.js, lines 141-148:

```javascript
it('purchases gems', () => {
      let [, message] = purchase(user, {params: {type: 'gems', key: 'gem'}});

      expect(message).to.equal(i18n.t('plusOneGem'));
      expect(user.balance).to.equal(userGemAmount + 0.25);
      expect(user.purchased.plan.gemsBought).to.equal(1);
      expect(user.stats.gp).to.equal(goldPoints - planGemLimits.convRate);
    });
```
Do you mean a test to check that messages are not duplicated?

TheHollidayInn:
I think it would be good to have a test where a user starts with a language other than the default and that the gem message is translated.

myoshuGO:
Ok, I'll try!

myoshuGO:
Sorry about the second commit, I forgot to run the lint tests.

TheHollidayInn:
Looks good to me!

TheHollidayInn:
Noted towards your tier! Thank you.

myoshuGO:
Nice! Thanks @TheHollidayInn

Issue #8341, opened 2016-12-31
Subscribe page needs text describing how to give a gift subscription


Alys:
I've seen several people ask how to give gift subscriptions. They go to the Settings > Subscription page and try to find the option there. Occasionally they even buy a subscription for themself, thinking that they'll then be able to give it to someone else.

To fix this, the Settings > Subscription page (https://habitica.com/#/options/settings/subscription ) needs a sentence explaining that you can give a subscription to someone else by clicking on their avatar in the party header or their name in any chat area to see their profile, then at the bottom left click on the gift box icon.

@lemoness Do you have any preference for the wording used or the exact position on the [Settings > Subscription page](https://habitica.com/#/options/settings/subscription)?

lemoness:
Something like "Want to gift a subscription to someone?

1. Open their profile! You can do this by clicking on their avatar in your
party header, or by clicking on their name in chat.

2. Click on the gift icon in the bottom left of their profile.

3. Select "subscription" and enter your payment information.

Thanks for supporting Habitica!"

mleah:
I'd like to take a look at this issue.  Anything else I need to know or any reason to wait on starting?

Alys:
@mleah Thanks! Please start whenever you're ready. 

I think the best location for it might be on the left hand side of the page under the Benefits information, but we might change our minds on that once we've seen what it looks like (apologies in advance if that happens!). I don't think there's anything else you'd need to know.

PR #8386, opened 2017-01-08
Adding "How to gift subscription" copy to Settings > Subscription Page:  Fixes Issue 8341

mleah:

Fixes https://github.com/HabitRPG/habitica/issues/8341

### Changes

This PR adds in "How to Gift Subscription" copy into the Settings > Subscription page, below the Benefits heading and copy.  

Copy was taken from the issue https://github.com/HabitRPG/habitica/issues/8341 .  I added the copy into the locales/en/subscriber.json file.

Here is the copy:

"Want to gift a subscription to someone?
1. Open their profile! You can do this by clicking on their avatar in your
party header, or by clicking on their name in chat.
2. Click on the gift icon in the bottom left of their profile.
3. Select "subscription" and enter your payment information.

Thanks for supporting Habitica!"

#### Screenshots

Current
screen shot 2017-01-08 at 12 35 29 pm

Change
screen shot 2017-01-08 at 12 23 21 pm

----
UUID: 2fd8e68b-4105-4bba-9378-82e4f64c858c


Alys:
I love the look of the new section, but is it possible to move the list of prices and coupon code box upwards so that it starts at the top of the right hand side and then have the three payment options buttons directly underneath it, all of them also on the right hand side? The row-column structure might need to be adjusted for this.

It might even be best if the new section starts under both the Benefits and payment options sections, to make it really clear that the payment options aren't related to it. So perhaps something like this (but without visible table borders!):

<table><tr><td valign="top">Benefits<br>Buy gems with gold<br>etc</td><td>Recurring $5 every....<br>Recurring $15 every....<br>etc<br>[coupon code form goes here]<br>Subscribe using:<br>[the three buttons go here]</td></tr><tr><td colspan="2">New section goes here, stretching across both columns if that's easy to do</td></tr></table>

mleah:
@Alys  I've updated the layout of the Settings > Subscription page based on your comment.  Please let me know if this is what you had in mind!

#### Screenshot

screen shot 2017-01-08 at 6 33 33 pm

Alys:
I like that! @lemoness what do you think?

lemoness:
Looks good to me!

website/common/locales/en/subscriber.json
```diff
@@ -14,6 +14,11 @@
     "mysteryItemText": "Each month you will receive a unique cosmetic item for your avatar! Plus, for every three months of consecutive subscription, the Mysterious Time Travelers will grant you access to historic (and futuristic!) cosmetic items.",
     "supportDevs": "Supports the developers",
     "supportDevsText": "Your subscription helps keep Habitica thriving and helps fund the development of new features. Thank you for your generosity!",
+    "giftSubscription": "Want to gift a subscription to someone?",
+    "giftSubscriptionText1": "Open their profile! You can do this by clicking on their avatar in your party header, or by clicking on their name in chat.",
```
khipkin: Grammar nit: maybe remove this comma since it does not separate 2 complete sentences?

mleah:
@khipkin   Removed the extra comma.  Let me know if it looks good!

website/views/options/settings/subscription.jade
```diff
@@ -50,14 +49,26 @@ script(id='partials/options.settings.subscription.html',type='text/ng-template',
...
+    h3= env.t('giftSubscription')
+    .row
+      .col-lg-12
+        ol
+          li
+            span= env.t('giftSubscriptionText1')
+          li
+            span= env.t('giftSubscriptionText2')
+          li
+            span= env.t('giftSubscriptionText3')
```
Alys: I'm not sure those three span tags are needed. I think doing the list items like this would work:

`    li= env.t('giftSubscriptionText1')`

@mleah What do you think? Do the spans add anything to the layout?

mleah:
@Alys   The layout looks the same without the spans, so I went ahead and removed them.  Let me know if you have any other feedback!

TheHollidayInn:
Noted towards your Tier! Congrats on Tier 1!

PR #8388, opened 2017-01-09
Group plans misc fixes

TheHollidayInn:

Fixes put_issue_url_here

### Changes
Fixes:
- Approval request notifications are now in the leader's language
- Tasks are synced after claiming
- Assigned users are correctly synced after edit
- Approval required is correctly synced after edit
- Member's with commas in their name can be assigned
- Approvals while away are now correctly synced on the UI

----
UUID:


website/server/models/group.js
```diff
@@ -392,7 +392,18 @@ schema.methods.sendChat = function sendChat (message, user) {
...
   this.chat.unshift(newMessage);
-  this.chat.splice(200);
+
+  const MAX_CHAT_COUNT = 200;
+  const MAX_SUBBED_GROUP_CHAT_COUNT = 400;
+
+  let maxCount = MAX_CHAT_COUNT;
+
+  let plan = this.purchased.plan;
+  if (plan && plan.customerId && !plan.dateTerminated) {
```
paglias: should also check `if plan.dateTerminated && plan.dateTerminated < now`

TheHollidayInn: Okay. I think what I have here is exactly what we use to check an active user subscription. Should I update that as well?

Alys: It is definitely possible to have a value for purchased.plan.dateTerminated in the past - that indicates an inactive subscription. If there's any checks for active user subscriptions that rely on dateTerminated not existing, that would be a bug.

A dateTerminated in the past could also indicate a newly-created subscription - one that has just been made one or more days after an old one terminated. When the new one is created,  dateTerminated should be set to null, but I'm pretty sure there's still bug(s) that prevent that happening.

TheHollidayInn: So the active sub check should be:

`if (plan && plan.customerId && (!plan.dateTerminated || plan.dateTerminated < now)) {`

Alys: plan.dateTerminated > now

It is possible for an active subscription to have a dateTerminated < now, but it's a bug situation (and the subscription will end as soon as the next cron runs). For normal code, we'd have to assume that plan.dateTerminated < now means that the subscription is no longer active.

paglias: What I'm saying is that a plan.dateTerminated > now doesn't mean the user has no more subscription benefit, it's the opposite: dateTerminated < now means they've expire > now they still haven't expired and !dateTerminated they still have an active sub (with customerId) or never had one (without customerId)

TheHollidayInn:
Yea... my bad :P

TheHollidayInn:
Okay, I think this is ready. @Alys you marked this as notify helpers. What do we need to notify?

Alys:
@TheHollidayInn When this is live, I'll notify the Socialites who were doing testing in my private guild. They'll be interested in the changes and will help to update the wiki.

PR #8389, opened 2017-01-10
gulp/gulp-transifex-test.js: fixed variable name typo

lynxlynxlynx:
Trivial typo fix, noticed while reading the droparticle discussion.


Alys:
Thanks @lynxlynxlynx !

PR #7987, opened 2016-09-09
reduce number of required items for collection quests (fixes #5842) {continued in #8754}

sebastienplat:
Fixes https://github.com/HabitRPG/habitrpg/issues/5618
### Changes

Reduce the number of required items for collection quests, with values proposed by [lemoness](https://github.com/HabitRPG/habitrpg/issues/5618#issuecomment-157479872):

| Quest Name | Quest Object Name | Old value(s) | new value(s) |
| :-- | :-- | :-- | :-- |
| Dish Disaster | atom1 | 20 | no change |
| Find the Cub | evilsanta2 | 30 (10 branches + 20 tracks) | no change |
| Find the Lair of the Wyrm | vice2 | 45 | 30 |
| Egg Hunt | egg | 100 | 40 |
| Message in a Bottle | dilatoryDistress1 | 50 (25 fireCoral + 25 blueFins) | 40 (20 fireCoral + 20 blueFins) |
| A Stern Talking-To | goldenknight1 | 300 | 60 |
| The Moonstone Chain | moonstone1 | 500 | 100 |

---

UUID: b6d7a7ca-98e1-4f11-b895-47c2b3667b9d

PR #8392, opened 2017-01-10
create script to insert message into party chat when collection quests become easier {continued in #8754}

Alys:
Some of the collection quests are being made easier in  https://github.com/HabitRPG/habitrpg/pull/7987 

This PR adds a migration script that finds parties currently engaged in one of those collection quests and puts a message into party chat to explain why the quest has changed.

I have not tested this recently because something's broken with mongod on my dev box, but you might want code changes for it anyway. For example, it doesn't currently use monk.

### DIFF
--- a/migrations/20170111_announce_collection_quest_change_in_parties.js
+++ b/migrations/20170111_announce_collection_quest_change_in_parties.js
@@ -1,0 +1,115 @@
+'use strict';
+
+/****************************************
+ * Author: @Alys
+ *
+ * Reason: Collection quests are being changed
+ * to require fewer items collected:
+ * https://github.com/HabitRPG/habitrpg/pull/7987
+ * This will cause existing quests to end sooner
+ * than the party is expecting.
+ * This script inserts an explanatory `system`
+ * message into the chat for affected parties.
+ ***************************************/
+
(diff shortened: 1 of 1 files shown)


TheHollidayInn:
for #7987

migrations/20170111_announce_collection_quest_change_in_parties.js
```diff
@@ -0,0 +1,115 @@
...
+  var newMessage = {
+    'id' : uuid.v4(),
+    'text' : message,
+    'timestamp': Date.now(),
+    'likes': {},
+    'flags': {},
+    'flagCount': 0,
+    'uuid': 'system'
+  };
+  return Groups.findOneAndUpdate({_id: group._id}, {$push:{"chat" :{$each: [newMessage], $position:0}}}, {returnOriginal: false});
```
paglias, line 92: why not just Groups.update?

Alys: No solid reason. I'd copied from another migration script and left it mostly unchanged since it worked,  but I'll try to get on to changing it as soon as the group plans subscription testing is over.

SabreCat:
Continued in #8754!

PR #8394, opened 2017-01-11
Group plans subs to all

TheHollidayInn:

Fixes put_issue_url_here

### Changes

General adds subscription to users who are in group plan. More specifically:
- On creation of group plan - all existing users get sub
- On joining a group plan - user receives sub
- On leaving a group plan - user loses sub
- On remove from a group plan - user loses sub
- On cancellation of a group plan - users are removed from sub

Todo
 - [x] User already has one-month once-off gift subscription (expected effect might be that the remaining fraction of the month already paid for is added to extraMonths).
- [x] User already has multi-month once-off gift subscription (the remaining full months and fraction of month already paid for is added to extraMonths).
- [x] User already has one-month recurring subscription (expected effect might be that the ongoing payments are cancelled and the remaining fraction of the month already paid for is added to extraMonths).
- [x] User already has multi-month recurring subscription (the ongoing payments are cancelled and the remaining full months and fraction of month already paid for is added to extraMonths).
- [x] User already has a cancelled but not yet terminated one-month recurring subscription (the remaining fraction of the month already paid for is added to extraMonths and the termination date is erased).
- [x] User already has a cancelled but not yet terminated multi-month recurring subscription (the remaining full months and fraction of month already paid for is added to extraMonths and the termination date is erased).
- [x] User already has a cancelled and terminated subscription - i.e., the termination date is in the past (the termination date is erased).
- [x] In any case where the user already had a POSITIVE value in `purchased.plan.extraMonths`, the additional extraMonths should be added to that.
- [x] In any case where the user already had a negative value in `purchased.plan.extraMonths`, the negative value should be ignored (those values might not exist any more, but a test for this would be good in case future bugs bring back negative values).
- [x] In any case where the user already has a non-zero value for `purchased.plan.gemsBought`, that value should be either NOT overwritten (my suggestion) OR set to zero, depending on what the business requirement is for this (keeping the existing count of gemsBought is consistent with how subscriptions currently work; if we change it for group plans, we might be queries about why it's not changed for normal subscriptions).
- [x] There could also be tests to ensure that the following data is not changed:
  - `purchased.plan.mysteryItems`
  - `purchased.plan.consecutive.count`
  - `purchased.plan.consecutive.offset`
  - `purchased.plan.consecutive.gemCapExtra`
  - `purchased.plan.consecutive.trinkets`
  - `purchased.plan.dateCreated` (troubleshooting complaints about missing Hourglasses is much easier if this is kept unchanged in every case except when a subscription lapses)

- [x] When ongoing payments are cancelled, it would be desirable if there were a test for all three payment methods (with a TODO comment to remind us to add tests later when in-app Apple and Android subscription payments are implemented).
- [x] Tests would also be needed for the case where a person is in one group plan and then also joins another - their group plan subscription should probably not be changed.
- [ ] Existing tests for subscriptions might need to be expanded to ensure they test for group plans subscriptions too (e.g., if there's a test to ensure that a subscriber is given `purchased.plan.mysteryItems` each month).
- [x] Tests would also be needed for the case where a person is in two group plans and then leaves one - their group plan subscription should not be removed.
- [x] Test for users with unlimited free sub
- [x] Add emails for each step

###Next Set
 - [x] Ensure user gains sub from the /join route
 - [x] Ensure user is removed from sub via /remove /leave routes
 - [x] Ensure users are cancelled when group plan is cancelled

----
UUID:


TheHollidayInn:
@paglias I'm hoping the way I created the sub for group plan members doesn't interfere with  the queue.

@SabreCat Can you help me verify that we are giving a "grace" period after cancellation so that users can sub an not lose hour glass.

paglias:
@TheHollidayInn for amazon as long as payment method is not "Amazon Payments" it simply skip the users

website/server/libs/payments.js
```diff
@@ -51,6 +51,95 @@ function _dateDiff (earlyDate, lateDate) {
...
+    consecutive: {
+      trinkets: 0,
+      offset: 0,
+      gemCapExtra: 0,
+    },
+  };
+
+  member.purchased.plan = plan;
+  data.user = member;
+  await this.createSubscription(data);
```
paglias: what if the user has an existing suscription?

Alys:
Can there be tests for all the different situations that might occur for pre-existing subscriptions? For example:

- User already has one-month once-off gift subscription (expected effect might be that the remaining fraction of the month already paid for is added to extraMonths).
- User already has one-month recurring subscription (expected effect might be that the ongoing payments are cancelled and the remaining fraction of the month already paid for is added to extraMonths).
- User already has multi-month once-off gift subscription (the remaining full months and fraction of month already paid for is added to extraMonths).
- User already has multi-month recurring subscription (the ongoing payments are cancelled and the remaining full months and fraction of month already paid for is added to extraMonths).
- User already has a cancelled but not yet terminated one-month recurring subscription (the remaining fraction of the month already paid for is added to extraMonths and the termination date is erased).
- User already has a cancelled but not yet terminated multi-month recurring subscription (the remaining full months and fraction of month already paid for is added to extraMonths and the termination date is erased).
- User already has a cancelled and terminated subscription - i.e., the termination date is in the past (the termination date is erased).
- In any case where the user already had a POSITIVE value in `purchased.plan.extraMonths`, the additional extraMonths should be added to that.
- In any case where the user already had a negative value in `purchased.plan.extraMonths`, the negative value should be ignored (those values might not exist any more, but a test for this would be good in case future bugs bring back negative values).
- In any case where the user already has a non-zero value for `purchased.plan.gemsBought`, that value should be either NOT overwritten (my suggestion) OR set to zero, depending on what the business requirement is for this (keeping the existing count of gemsBought is consistent with how subscriptions currently work; if we change it for group plans, we might be queries about why it's not changed for normal subscriptions).
- There could also be tests to ensure that the following data is not changed:
  - `purchased.plan.mysteryItems`
  - `purchased.plan.consecutive.count`
  - `purchased.plan.consecutive.offset`
  - `purchased.plan.consecutive.gemCapExtra`
  - `purchased.plan.consecutive.trinkets`
  - `purchased.plan.dateCreated` (troubleshooting complaints about missing Hourglasses is much easier if this is kept unchanged in every case except when a subscription lapses)

- Existing tests for subscriptions might need to be expanded to ensure they test for group plans subscriptions too (e.g., if there's a test to ensure that a subscriber is given `purchased.plan.mysteryItems` each month).
- When ongoing payments are cancelled, it would be desirable if there were a test for all three payment methods (with a TODO comment to remind us to add tests later when in-app Apple and Android subscription payments are implemented).
- Tests would also be needed for the case where a person is in one group plan and then also joins another - their group plan subscription should probably not be changed.
- Tests would also be needed for the case where a person is in two group plans and then leaves one - their group plan subscription should not be removed.
- There might be other cases that I've missed.

Alys:
One other thing about the tests: Because this feature involves potentially several dozen dollars for each group, we should be really sure that the tests are solid. As well as making sure the tests pass, we should temporarily break the website's code in various ways and make sure each test fails (e.g., for the dateTerninated tests, remove the website's code that sets dateTerminated to null and check that the tests pick up the bug).

Alys:
An extra thought about this:

> In any case where the user already has a non-zero value for `purchased.plan.gemsBought`, that value should be either NOT overwritten (my suggestion) OR set to zero, depending on what the business requirement is for this (keeping the existing count of gemsBought is consistent with how subscriptions currently work; if we change it for group plans, we might be queries about why it's not changed for normal subscriptions).

If gemsBought is set to zero whenever a group plan subscription is given, it could be used to farm gems. You create a group plan and invite a friend. The friend buys all their allowed gems. You remove the friend from the group and then reinvite them. The friend again buys all their allowed gems....

To prevent that, the only time that `purchased.plan.gemsBought` is ever set to zero for any type of subscription should be at the start of a new calendar month (i.e., the current behaviour for normal subscriptions should be applied to group plan subscriptions).

paglias:
Another missing thing, the code doesn't cancel the existing subscriptions
calling api.cancelSubscription

Il 11 Gen 2017 11:30 AM, "Alys" <notifications@github.com> ha scritto:

Alys:
The grace period after cancellation of a group plan subscription should not apply when the user has a positive value for extraMonths unless that value happens to be less than the grace period.

One possibility for the grace period: don't have special code for it at all. Instead, when a user is given a group plan sub, immediately set extraMonths to the grace period length (unless extraMonths is already larger than that), and then the grace period is automatically there for whenever it's needed in future. Less code is needed, no extra tests are needed; less chance for future bugs. Disadvantage: we can't easily extend the grace period in future for existing group plan subs, but if we choose a good value now, we might not need to.

TheHollidayInn:
From Alys in Slack: Test for users with unlimited free sub

website/server/controllers/api-v3/groups.js
```diff
@@ -21,7 +21,8 @@ import { encrypt } from '../../libs/encryption';
 import { sendNotification as sendPushNotification } from '../../libs/pushNotifications';
 import pusher from '../../libs/pusher';
 import common from '../../../common';
-import payments from '../../libs/payments';
+import payments from '../../libs/stripePayments';
```
TheHollidayInn: should be payments

website/server/libs/payments.js
```diff
@@ -44,6 +45,158 @@ function _dateDiff (earlyDate, lateDate) {
...
+
+  if (member.isSubscribed()) {
+    let memberPlan = member.purchased.plan;
+    let customerHasCancelledGroupPlan = memberPlan.dateTerminated && memberPlan.customerId === this.constants.GROUP_PLAN_CUSTOMER_ID && moment().isBefore(memberPlan.dateTerminated);
+    if (customerIdsToIgnore.indexOf(memberPlan.customerId) !== -1 && !customerHasCancelledGroupPlan) return;
+
+    await member.cancelSubscription();
+
+    let today = new Date();
+    plan = _.clone(member.purchased.plan.toObject());
```
paglias: Heads up, we're migrating to lodash v4 (https://github.com/HabitRPG/habitica/pull/8495) which includes many breaking changes (https://github.com/lodash/lodash/wiki/Changelog#v400). Before merging this PR should be checked and updated to make sure all lodash methods are compatible with v4.

website/server/libs/payments.js
```diff
@@ -44,6 +45,158 @@ function _dateDiff (earlyDate, lateDate) {
...
+    let extraMonths = Number(plan.extraMonths);
+    if (plan.dateTerminated) extraMonths += _dateDiff(today, plan.dateTerminated);
+
+    _(plan).merge({ // override with these values
```
paglias, line 121: Heads up, we're migrating to lodash v4 (https://github.com/HabitRPG/habitica/pull/8495) which includes many breaking changes (https://github.com/lodash/lodash/wiki/Changelog#v400). Before merging this PR should be checked and updated to make sure all lodash methods are compatible with v4.

paglias (changes requested):

paglias (commented):
If we decide to add yarn.lock we should keep it up to date

TheHollidayInn:
Eh, didn't mean to add that.

website/server/models/user/methods.js
```diff
@@ -91,3 +98,17 @@ schema.methods.addComputedStatsToJSONObj = function addComputedStatsToUserJSONOb
 
   return statsObject;
 };
+
+schema.methods.cancelSubscription = async function cancelSubscription () {
+  let plan = this.purchased.plan;
+
+  if (plan.paymentMethod === amazonPayments.constants.PAYMENT_METHOD_AMAZON) {
```
paglias: why is amazon's constant called PAYMENT_METHOD_AMAZON while paypal's and stripe PAYMENT_METHOD?

website/server/models/user/methods.js
```diff
@@ -91,3 +98,17 @@ schema.methods.addComputedStatsToJSONObj = function addComputedStatsToUserJSONOb
...
+    return await amazonPayments.cancelSubscription({user: this});
+  } else if (plan.paymentMethod === stripePayments.constants.PAYMENT_METHOD) {
+    return await stripePayments.cancelSubscription({user: this});
+  } else if (plan.paymentMethod === paypalPayments.constants.PAYMENT_METHOD) {
+    return await paypalPayments.subscribeCancel({user: this});
+  }
+
+  return await payments.cancelSubscription({user: this});
```
paglias, line 124: so payments. cancelSubscription should not be called directly anymore? I'm a little confused by this: we have user.cancelSubscription, payments.cancelSubscrition({user}) and all the payments specific ones and all do a different thing. For example what's the difference between `paypalPayments.subscribeCancel({user: this});` and `payments.cancelSubscription()` for a user that uses paypal?

website/server/libs/payments.js
```diff
@@ -44,6 +45,158 @@ function _dateDiff (earlyDate, lateDate) {
...
+    user: {},
+    sub: {
+      key: 'group_plan_auto',
+    },
+    customerId: 'group-plan',
+    paymentMethod: 'Group Plan',
+    headers: {},
+  };
+
+  let plan = {
```
paglias, line 91: most of these values are already defined as defaults in models/subscriptionPlan

website/server/libs/payments.js
```diff
@@ -44,6 +45,158 @@ function _dateDiff (earlyDate, lateDate) {
...
+    consecutive: {
+      trinkets: 0,
+      offset: 0,
+      gemCapExtra: 0,
+    },
+  };
+
+  if (member.isSubscribed()) {
+    let memberPlan = member.purchased.plan;
+    let customerHasCancelledGroupPlan = memberPlan.dateTerminated && memberPlan.customerId === this.constants.GROUP_PLAN_CUSTOMER_ID && moment().isBefore(memberPlan.dateTerminated);
```
paglias: can use the new hasCancelled method?

website/server/libs/payments.js
paglias: once the new version of mongoose is out (should be a few days) this can become member.purchased.plan.toObject() but until then _.clone is necessary otherwise it's not really cloned

website/server/libs/payments.js
```diff
@@ -44,6 +45,158 @@ function _dateDiff (earlyDate, lateDate) {
...
+  if (user.purchased.plan.customerId !== this.constants.GROUP_PLAN_CUSTOMER_ID) return;
+
+  let userGuilds = _.clone(user.guilds);
+  userGuilds.push('party');
+
+  let index = userGuilds.indexOf(group._id);
+  userGuilds.splice(index, 1);
+
+  let groupPlansQuery = {
+    type: {$in: ['guild', 'party']},
```
paglias: why match parties if then it limits to _ids in userGuilds?

paglias: oh I see, can we change the name of userGuilds to something else, otherwise it's super confusing

website/server/libs/payments.js
```diff
@@ -44,6 +45,158 @@ function _dateDiff (earlyDate, lateDate) {
...
+    return this.cancelGroupSubscriptionForUser(member, group);
+  });
+
+  await Promise.all(promises);
+};
+
+api.cancelGroupSubscriptionForUser = async function cancelGroupSubscriptionForUser (user, group) {
+  if (user.purchased.plan.customerId !== this.constants.GROUP_PLAN_CUSTOMER_ID) return;
+
+  let userGuilds = _.clone(user.guilds);
```
paglias: toObject

website/server/libs/payments.js
```diff
@@ -44,6 +45,158 @@ function _dateDiff (earlyDate, lateDate) {
...
+  userGuilds.push('party');
```
paglias: this should push the party _id not `party`?

website/server/libs/payments.js
```diff
@@ -289,7 +435,7 @@ api.cancelSubscription = async function cancelSubscription (data) {
     await data.user.save();
   }
 
-  txnEmail(data.user, emailType);
+  if (customerId !== this.constants.GROUP_PLAN_CUSTOMER_ID) txnEmail(data.user, emailType, emailMergeData);
```
paglias, line 445: the email is only sent if this is a user subscription? because emailMergeData is not empty only `if data.groupId`

website/server/controllers/api-v3/groups.js
```diff
@@ -907,10 +921,15 @@ async function _inviteByEmail (invite, group, inviter, req, res) {
     userReturnInfo = await _inviteByUUID(userToContact._id, group, inviter, req, res);
   } else {
     userReturnInfo = invite.email;
+
+    let cancelledPlan = false;
+    if (group.isSubscribed() && group.purchased.plan.dateTerminated) cancelledPlan = true;
```
paglias: !groupHasNotCancelled?

website/server/controllers/api-v3/groups.js
```diff
@@ -844,7 +856,9 @@ async function _inviteByUUID (uuid, group, inviter, req, res) {
       if (userParty && userParty.memberCount !== 1) throw new NotAuthorized(res.t('userAlreadyInAParty'));
     }
 
-    userToInvite.invitations.party = {id: group._id, name: group.name, inviter: inviter._id};
+    let partyInvite = {id: group._id, name: group.name, inviter: inviter._id};
+    if (group.isSubscribed() && group.purchased.plan.dateTerminated) partyInvite.cancelledPlan = true;
```
paglias: !group.hasNotCancelled?

website/server/controllers/api-v3/groups.js
```diff
@@ -831,7 +840,10 @@ async function _inviteByUUID (uuid, group, inviter, req, res) {
     if (_.find(userToInvite.invitations.guilds, {id: group._id})) {
       throw new NotAuthorized(res.t('userAlreadyInvitedToGroup'));
     }
-    userToInvite.invitations.guilds.push({id: group._id, name: group.name, inviter: inviter._id});
+
+    let guildInvite = {id: group._id, name: group.name, inviter: inviter._id};
+    if (group.isSubscribed() && group.purchased.plan.dateTerminated) guildInvite.cancelledPlan = true;
```
paglias: !group.hasNotCancelled?

website/server/controllers/api-v3/groups.js
```diff
@@ -763,7 +769,10 @@ api.removeGroupMember = {
 
     if (isInGroup) {
       group.memberCount -= 1;
-      if (group.purchased.plan.customerId) await payments.updateStripeGroupPlan(group);
+      if (group.isSubscribed() && !group.purchased.plan.dateTerminated)  {
```
paglias: group.hasNotCancelled?

website/server/controllers/api-v3/groups.js
```diff
@@ -676,7 +682,7 @@ api.leaveGroup = {
 
     await group.leave(user, req.query.keep, req.body.keepChallenges);
 
-    if (group.purchased.plan && group.purchased.plan.customerId) await payments.updateStripeGroupPlan(group);
+    if (group.isSubscribed() && !group.purchased.plan.dateTerminated)  await group.updateGroupPlan(true);
```
paglias: group.hasNotCancelled?

website/server/controllers/api-v3/groups.js
```diff
@@ -493,7 +496,10 @@ api.joinGroup = {
 
     group.memberCount += 1;
 
-    if (group.purchased.plan.customerId) await payments.updateStripeGroupPlan(group);
+    if (group.isSubscribed() && !group.purchased.plan.dateTerminated)  {
```
paglias: group.hasNotCancelled?

website/server/models/group.js
```diff
@@ -1165,6 +1174,19 @@ schema.methods.isSubscribed = function isSubscribed () {
   return plan && plan.customerId && (!plan.dateTerminated || moment(plan.dateTerminated).isAfter(now));
 };
 
+schema.methods.hasNotCancelled = function hasNotCancelled () {
```
paglias, line 1181: I think this is missing unit tests

paglias (changes requested):

website/server/models/user/methods.js
TheHollidayInn: I changed my mind after refactor amazon. I can update amazon as well.

website/server/models/user/methods.js
TheHollidayInn: Yea, it is confusing. The problem is there is a recursive relationship between payments and the other payment providers. Which means we need a third entity. Here the third entity is the User.

You do not want to cal payments.cancelSubscription, because that will not call the api functions with the respective payment provider api. So, to answer you paypal question, if you call the payments.cancelSubscription, you will not correctly cancel the billing date because paypal.getBillingDate will not be called.

There is a better way to organize this, but that should wait until the next refactor. I can put it on my list for right after this PR, but I don't want to hold up the PR because it is already large.

website/server/libs/payments.js
TheHollidayInn: Only five keys are assigned by default. I'll double check what is redundant though.

website/server/libs/payments.js
TheHollidayInn: No, the email is only sent if the plan is not a user group plan (the free plan that all users get if they are members of a group with a group plan)

website/server/models/user/methods.js
paglias: Okay, can we add a comment about payments.cancelSubscription explaining the situation?

PR #8407, opened 2017-01-14
fixed recoverCron tests failure when using mocha v3+

FedyuninV:

Fixes some failing tests

### Changes
Callbacks were removed from `RecoverCron` tests in `test:api-v3:unit` part. Now it returns a promises, as it shown in  [README](https://github.com/HabitRPG/habitica/blob/develop/test/api/README.md).

REASON: Some tests are failing when using Mocha v3+, because of  `Error: Resolution method is overspecified. Specify a callback *or* return a Promise; not both.`  You can find an example [here](https://mochajs.org/#asynchronous-code) in section "Working with promises".

P.S. this is my first pull request and it is small because I want to be sure that this way of test repairing is correct. If it would be accepted or explained why this PR is incorrect, I will try to repair remaining tests (there are 11 more failing tests in `test:api-v3:unit` section).

----
UUID: b78f667c-eb2b-416d-866b-65e75221772a


TheHollidayInn:
Thanks for submitting this! I'm not sure it is completely correct though. `await/async` is a way to work with promises. You shouldn't have to revert back to `.then/.catch`

FedyuninV:
Agree, turned back to async/await, but "done" removed.

paglias:
LGTM, thanks!

paglias:
I've notated this towards your next tier

Issue #7715, opened 2016-06-25
Start Date can be set to null / empty string on a Daily


Alys:
When editing a Daily on the website, you can delete the Start Date. After saving, the task has a Start Date of `null`.

This is confusing because it's not obvious to the player how cron will treat it (immediately due versus never due) and because the official and unofficial site/apps/tools will display the due-ness of the Daily in different ways (e.g., website says not due, Android app says due).

I believe the API should prohibit a non-existent or nonsensical Start Date. It should always be a valid date, and if a valid date isn't supplied, then default to the current date, which is what cron does anyway:
common/script/cron.js:  `let taskStartDate = moment(dailyTask.startDate).zone(o.timezoneOffset);`

The website should also handle deletion of the Start Date in a more informative way. For example, if there's no Start Date then don't let the Daily be saved, or save it but set the Start Date to today so that you see the date if you edit the task again, or set the Start Date to today as soon as you move focus out of the Start Date field. I think the latter is my favourite option at the moment.

Alys:
oops, saved prematurely. I've changed the final paragraph to this:


The website should also handle deletion of the Start Date in a more informative way. For example, if there's no Start Date then don't let the Daily be saved, or save it but set the Start Date to today so that you see the date if you edit the task again, or set the Start Date to today as soon as you move focus out of the Start Date field. I think the latter is my favourite option at the moment.

Alys:
Ha, I'm an idiot. Cron treats Dalies with no Start Date as starting on the day that cron runs (`moment()`), which means they were NOT due on  the previous day.

So I guess if there's no Start Date supplied, the default could be `tomorrow`, although I still sort of like the idea of setting it to today, for consistency with new Dailies. I don't have a strong preference though.

ghost:
So in other words, if no start date is set, it'll always be tomorrow i.e. never, won't it?

Alys:
@GitHubSphinx That's correct, with the current implementation. I'd encourage anyone reading this to not rely on that always being the case.

I can see that that could be considered a useful feature (an easy way to make a Daily permanently not due) but I don't think it's intuitive, and it relies upon all possible clients (official and unofficial; current and future) handling a missing Start Date in the same manner (that won't make any difference to cron but it will affect how the user perceives which Dailies are due).

paglias:
Can be fixed by setting `required: true`, making sure it's a default value and running a migration to set the missing dates to today (?)

Alys:
[...]

PR #8414, opened 2017-01-15
Change type of date on Todo task to Date

### LINKED ISSUE #8405: API: dates may be invalid
General Info
  * Browser: Safari

### Description

#### Console Errors

### Notes

sprusr:

Fixes #8405 

### Changes

Was previously set to String, with no validation. Schema now checks for
correct date, and stores as a Date. Also includes a migration to
convert valid date strings to Dates in the database, and remove invalid
strings.

----
UUID: db711097-c5d0-4d25-8e5e-d955da4b79fd

### DIFF
--- a/migrations/20170115_api_date_validation.js
+++ b/migrations/20170115_api_date_validation.js
@@ -1,0 +1,65 @@
+var migrationName = '20170115_api_date_validation.js';
+var authorName = 'sprusr'; // in case script author needs to know when their ...
+var authorUuid = 'db711097-c5d0-4d25-8e5e-d955da4b79fd'; //... own data is done
+
+/*
+ * API date validation
+ */
+
+var monk = require('monk');
+
+var connectionString = 'mongodb://localhost:27017/habitrpg?auto_reconnect=true'; // FOR TEST DATABASE
+
+var dbTasks = monk(connectionString).get('tasks', { castIds: false });
+
+var query = {
+  date: { $exists: true }
+};
+
+console.warn('Updating tasks...');
+
+dbTasks.find(query, 'date').each(function(task, cursor) {
+  var updateQuery = {};
+  var newDate = new Date(task.date);
+
+  if(isValidDate(newDate)) {
+    updateQuery = {
+      $set: {
+        'date': newDate
+      }
+    }
+  } else {
+    updateQuery = {
+      $unset: {
+        'date': ''
+      }
+    }
+  }
+
+  console.log(task, updateQuery);
+
+  cursor.pause();
+
+  dbTasks.update(task._id, updateQuery).then(function() {
+    cursor.resume();
+  });
+}).then(function() {
+  console.warn('All appropriate tasks found and modified.');
+  return exiting(0);
+}).catch(function(err) {
+  return exiting(1, 'ERROR! ' + err);
+});
+
+function isValidDate(d) {
+  return Object.prototype.toString.call(d) === "[object Date]" && !isNaN(d.getTime());
+}
+
+function exiting(code, msg) {
+  code = code || 0; // 0 = success
+  if (code && !msg) { msg = 'ERROR!'; }
+  if (msg) {
+    if (code) { console.error(msg); }
+    else      { console.log(  msg); }
+  }
+  process.exit(code);
+}
--- a/test/api/v3/integration/tasks/POST-tasks_user.test.js
+++ b/test/api/v3/integration/tasks/POST-tasks_user.test.js
@@ -205,6 +205,18 @@ describe('POST /tasks/user', () => {
         message: t('taskAliasAlreadyUsed'),
       });
     });
+
+    it('errors if todo due date supplied is not a valid date', async () => {
+      await expect(user.post('/tasks/user', {
+        type: 'todo',
+        text: 'todo text',
+        date: 'not a valid date',
+      })).to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: 'todo validation failed',
+      });
+    });
   });
 
   context('sending task activity webhooks', () => {
--- a/website/server/models/task.js
+++ b/website/server/models/task.js
@@ -241,8 +241,7 @@ export let daily = Task.discriminator('daily', DailySchema);
 
 export let TodoSchema = new Schema(_.defaults({
   dateCompleted: Date,
-  // TODO we're getting parse errors, people have stored as "today" and "3/13". Need to run a migration & put this back to type: Date see http://stackoverflow.com/questions/1353684/detecting-an-invalid-date-date-instance-in-javascript
-  date: String, // due date for todos
+  date: Date, // due date for todos
 }, dailyTodoSchema()), subDiscriminatorOptions);
 export let todo = Task.discriminator('todo', TodoSchema);
 


migrations/20170115_api_date_validation.js
```diff
@@ -0,0 +1,73 @@
...
+console.warn('Updating tasks...');
+
+dbTasks.findEach(query, { batchSize: 250 }, function(err, task) {
+  if (err) { return exiting(1, 'ERROR! ' + err); }
+  if (!task) {
+    console.warn('All appropriate tasks found and modified.');
+    return exiting(0);
+  }
+
+  console.log(task);
```
khipkin: Suggestion: if this is just for testing the script, put a "TODO" comment so you remember to remove it.

migrations/20170115_api_date_validation.js
```diff
@@ -0,0 +1,73 @@
...
+        'date': ''
+      }
+    }
+  }
+
+  dbTasks.update({_id: task._id}, updateQuery);
+});
+
+function isValidDate(d) {
+  if (Object.prototype.toString.call(d) === "[object Date]" && !isNaN(d.getTime())) {
```
khipkin: function isValidDate(d) {
        return Object.prototype.toString.call(d) === "[object Date]" && !isNaN(d.getTime());
    }

migrations/20170115_api_date_validation.js
sprusr: Ah yes, well spotted. This was indeed for testing purposes, will remove. Thanks.

sprusr:
For the sake of cleanliness, I've amended that to the previous commit. Should be ready to go now @khipkin

khipkin:
You should probably add a test that non-date values now fail.

librarianmage:
Is this related to #7715? If so, would you be able to fix that here as well? I've completely forgotten to work on it, and this seems to be related/the same problem.

migrations/20170115_api_date_validation.js
```diff
@@ -0,0 +1,67 @@
+var migrationName = '20161230_nye_hats.js';
```
librarianmage: Does this need to be changed?

khipkin:
@MathWhiz #7715 seems to include UI changes as well (disallow saving tasks with invalid start dates).

sprusr:
@MathWhiz does seem to be out of the scope of this fix, but thanks for the heads up about the migration name

librarianmage:
@sprusr Ok. I will wait until your PR gets merged in to start working on it.

Alys:
@sprusr I like this change and I love that you did the migration script!

I think though that we need to delay merging it because it could count as a breaking change for the API. There might be third-party tools using that rely on the date field being free-text and/or that rely on a specific date format in the users' tasks. I think we definitely should make this change though - it's clearly an improvement. We can advertise the change and the data migration in the coders guilds in Habitica and in our dev blog and then wait a while before merging.

@paglias @crookedneighbor Tell us if you disagree!

paglias:
I'm in favor of the change, I consider it a bug fix instead of a breaking change

sprusr:
@khipkin got round to adding a test. Not very experienced with writing tests, so please let me know if it can be improved (edit: oops I did that wrong, fixed now).

@Alys @paglias I would think it would depend on what the API docs say about setting the due date. I just tried to have a look, but didn't find [the docs](https://habitica.com/apidoc/#api-Task-CreateUserTasks) very descriptive. Is the schema for a todo task documented somewhere? Otherwise, I reckon those docs could be an area of improvement and might have a look into that myself 😁

But yeah, merge it when you feel appropriate. Thanks to all for the guidance in contributing to this.

migrations/20170115_api_date_validation.js
paglias: yes

migrations/20170115_api_date_validation.js
```diff
@@ -0,0 +1,67 @@
...
+
+var mongo = require('mongoskin');
+
+var connectionString = 'mongodb://localhost:27017/habitrpg?auto_reconnect=true'; // FOR TEST DATABASE
+
+var dbTasks = mongo.db(connectionString).collection('tasks');
+
+var query = {
+  $nor: [
+    { date: { $type: "date" } },
```
paglias: I don't think this is going to work as $nor matches docs whose date type is not a date AND where the date fields doesn't exist. And i don't think is necessary as all date fields are going to be strings if they exist. You can simply match `date: {$exists: true}`

migrations/20170115_api_date_validation.js
```diff
@@ -0,0 +1,67 @@
...
+    return exiting(0);
+  }
+
+  var updateQuery = {};
+  var newDate = new Date(task.date);
+
+  if(isValidDate(newDate)) {
+    updateQuery = {
+      $set: {
+        'migration': migrationName,
```
paglias: the migration field doesn't exits on tasks, let's not set it

migrations/20170115_api_date_validation.js
```diff
@@ -0,0 +1,67 @@
...
+        'date': newDate
+      }
+    }
+  } else {
+    updateQuery = {
+      $set: {
+        'migration': migrationName
```
paglias: same here

migrations/20170115_api_date_validation.js
```diff
@@ -0,0 +1,67 @@
+var migrationName = '20161230_nye_hats.js';
+var authorName = 'sprusr'; // in case script author needs to know when their ...
+var authorUuid = 'db711097-c5d0-4d25-8e5e-d955da4b79fd'; //... own data is done
+
+/*
+ * API date validation
+ */
+
+var mongo = require('mongoskin');
```
paglias: the migration should be updated to use `monk` instead of mongoskin that doesn't support anymore our database version, see https://github.com/HabitRPG/habitica/blob/develop/migrations/20170120_missing_incentive.js#L10

paglias (changes requested):

sprusr:
@paglias fingers crossed, that should sort everything

migrations/20170115_api_date_validation.js
```diff
@@ -0,0 +1,65 @@
...
+    updateQuery = {
+      $unset: {
+        'date': ''
+      }
+    }
+  }
+
+  console.log(task, updateQuery);
+
+  cursor.pause();
```
paglias, line 41: what is this for? To make sure only one task is processed at a certain time? There are more than 20 milions tasks that have this field so even with a low processing time of 5ms per task (and with latency it's going to be a lot more) it'll take more than a day to run the migration. I think we want multiple tasks processed at the same time. For example fetch 100 tasks, process them while we fetch the next 10 and when the first 10 have been processed repeat the process

migrations/20170115_api_date_validation.js
```diff
@@ -0,0 +1,65 @@
...
+
+  dbTasks.update(task._id, updateQuery).then(function() {
```
paglias, line 43: this should be returned in order for errors to be catched by the `.catch` below

paglias (changes requested):

SabreCat:
@sprusr, this PR has been awaiting your input for a while. Are you still planning to address the comments here?

SabreCat:
No response to our query, so I'm closing this and reopening the source issue as help-welcome.

PR #8427, opened 2017-01-19
spells: searing brightness should not affect challenge tasks

### LINKED ISSUE #8423: Searing Brightness affects challenge tasks
Skills are not meant to adjust the value of challenge tasks but when a Healer casts Searing Brightness, all their tasks are made less red, including challenge tasks.

The desired behaviour is that challenge tasks should not be affected by Searing Brightness.

lynxlynxlynx:
Just added a few checks for various challenge states.

Fixes #8423 
----
UUID: 10c530d0-a796-476b-8a31-6f048c6e8a8c

### DIFF
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -901,11 +901,19 @@ api.clearCompletedTodos = {
       userId: user._id,
       type: 'todo',
       completed: true,
-      $or: [
-        {'challenge.id': {$exists: false}},
-        {'challenge.broken': {$exists: true}},
-        {'group.id': {$exists: false}},
-        {'group.broken': {$exists: true}},
+      $and: [ // exclude challenge and group tasks
+        {
+          $or: [
+            {'challenge.id': {$exists: false}},
+            {'challenge.broken': {$exists: true}},
+          ],
+        },
+        {
+          $or: [
+            {'group.id': {$exists: false}},
+            {'group.broken': {$exists: true}},
+          ],
+        },
       ],
     }).exec();
 
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -896,7 +896,7 @@ api.clearCompletedTodos = {
     let user = res.locals.user;
 
     // Clear completed todos
-    // Do not delete challenges completed todos unless the task is broken
+    // Do not delete completed todos from challenges or groups, unless the task is broken
     await Tasks.Task.remove({
       userId: user._id,
       type: 'todo',
--- a/test/api/v3/integration/tasks/POST-tasks_clearCompletedTodos.test.js
+++ b/test/api/v3/integration/tasks/POST-tasks_clearCompletedTodos.test.js
@@ -43,7 +43,7 @@ describe('POST /tasks/clearCompletedTodos', () => {
     let completedTodos = await user.get('/tasks/user?type=completedTodos');
     let todos = await user.get('/tasks/user?type=todos');
     let allTodos = todos.concat(completedTodos);
-    expect(allTodos.length).to.equal(initialTodoCount + 4); // + 6 - 3 completed (but one is from challenge)
-    expect(allTodos[allTodos.length - 1].text).to.equal('todo 7');
+    expect(allTodos.length).to.equal(initialTodoCount + 5); // + 7 - 3 completed (but one is from challenge)
+    expect(allTodos[allTodos.length - 1].text).to.equal('todo 6'); // last completed todo
   });
 });
--- a/website/server/controllers/api-v3/user.js
+++ b/website/server/controllers/api-v3/user.js
@@ -392,11 +392,19 @@ api.castSpell = {
     } else if (targetType === 'tasks') { // new target type in v3: when all the user's tasks are necessary
       let tasks = await Tasks.Task.find({
         userId: user._id,
-        $or: [ // exclude challenge tasks
-          {'challenge.id': {$exists: false}},
-          {'challenge.broken': {$exists: true}},
-          {'group.id': {$exists: false}},
-          {'group.broken': {$exists: true}},
+        $and: [ // exclude challenge and group tasks
+          {
+            $or: [
+              {'challenge.id': {$exists: false}},
+              {'challenge.broken': {$exists: true}},
+            ],
+          },
+          {
+            $or: [
+              {'group.id': {$exists: false}},
+              {'group.broken': {$exists: true}},
+            ],
+          },
         ],
       }).exec();
 
--- a/website/server/controllers/api-v3/user.js
+++ b/website/server/controllers/api-v3/user.js
@@ -1063,11 +1063,19 @@ api.userRebirth = {
     let tasks = await Tasks.Task.find({
       userId: user._id,
       type: {$in: ['daily', 'habit', 'todo']},
-      $or: [ // exclude challenge tasks
-        {'challenge.id': {$exists: false}},
-        {'challenge.broken': {$exists: true}},
-        {'group.id': {$exists: false}},
-        {'group.broken': {$exists: true}},
+      $and: [ // exclude challenge and group tasks
+        {
+          $or: [
+            {'challenge.id': {$exists: false}},
+            {'challenge.broken': {$exists: true}},
+          ],
+        },
+        {
+          $or: [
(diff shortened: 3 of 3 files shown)


TheHollidayInn:
We will also need this to skip group tasks: https://github.com/HabitRPG/habitica/blob/develop/website/common/script/ops/reset.js#L15 .

Also, can you add a test for this? Example: https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/user/POST-user_reset.test.js#L90

lynxlynxlynx:
Why exclude group tasks? That seems arbitrary.

I didn't see any tests for spells — is there a preferred place? Since most of the code will be identical, I'd just stick into that file, but it doesn't really belong.

TheHollidayInn:
Group tasks are handled like challenge tasks. 

I believe this is the correct place to write spell tests: https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/user/POST-user_class_cast_spellId.test.js

lynxlynxlynx:
Hmpf, after fixing it for the various spell casting prerequisites, it now fails with the wrong task value. It didn't get incremented, but something seems to reset it back to the default 0. It can't be the spell.

Am I missing something obvious? The schema matches and other tests construct them the same way.

paglias:
Aren't challenge tasks already excluded https://github.com/HabitRPG/habitica/blob/develop/website/server/controllers/api-v3/user.js#L392-L401?

test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
```diff
@@ -186,6 +187,42 @@ describe('POST /user/class/cast/:spellId', () => {
...
+    await user.post(`/tasks/challenge/${challenge._id}`, {
+      text: 'test challenge habit',
+      type: 'habit',
+      value: 1,
+    });
+
+    let groupTask = await user.post(`/tasks/group/${guild._id}`, {
+      text: 'todo group',
+      type: 'todo',
+      value: 1,
```
paglias: you cannot set an arbitrary `value` number when using the api, it's not among the fields that you're allowed to modify

paglias (changes requested):

lynxlynxlynx:
Ah, thanks! I guess it's cleaner then to just add checks for challenge and group to the general list in castSpell().

paglias:
@lynxlynxlynx what i was saying is that i don't understand why this is happening since challenge tasks already seem to be excluded

lynxlynxlynx:
Yes, but from all other checks, I gather that there may be some tasks that can have "challenge" defined, but not match the rest of the criteria (and same for group).

paglias:
I would like to have a test that fails before your changes to show that the current behavior isn't working correctly

lynxlynxlynx:
I'll open another PR with just the test. It's fine with the default value too.

Alys:
Challenge tasks are no longer automatically excluded because of this code:

https://github.com/HabitRPG/habitica/blob/develop/website/server/controllers/api-v3/user.js#L396-L399

It's saying that the skill should find tasks that are not challenge tasks OR that are not group plans tasks.

Any challenge task is not a group plan task so all challenge tasks are found and processed by the skill. Also vice versa - all group plans tasks are processed (and should not be).

Alys:
So what we probably want for this issue is a fix for that piece of broken code, rather than new code.

website/server/controllers/api-v3/tasks.js
```diff
@@ -901,11 +901,19 @@ api.clearCompletedTodos = {
       userId: user._id,
```
Alys, line 901: Could you please also modify the comment a couple of lines above here so that it says "Do not delete challenge and group completed ..."? Sorry, I realise that's not your omission!

website/server/controllers/api-v3/tasks.js
```diff
@@ -901,11 +901,19 @@ api.clearCompletedTodos = {
...
       type: 'todo',
       completed: true,
-      $or: [
-        {'challenge.id': {$exists: false}},
-        {'challenge.broken': {$exists: true}},
-        {'group.id': {$exists: false}},
-        {'group.broken': {$exists: true}},
+      $and: [ // exclude challenge and group tasks
```
Alys, line 904: I think that this outer $and might  not be necessary since $and is used by default when multiple conditions are given. Just check that though before taking my word for it. :)

Alys (commented):
Thanks very much for finding those other bits of code to fix!

website/server/controllers/api-v3/tasks.js
lynxlynxlynx: Apparently not: https://docs.mongodb.com/manual/reference/operator/query/and/ (at the end)

Alys: Oh great point! Thank you! :+1:

lynxlynxlynx:
The failing test looks unrelated, maybe a temporary external problem:
```
 1) payments : amazon #subscribeCancel success cancels a user subscription:
     Error: timeout of 8000ms exceeded. Ensure the done() callback is being called in this test.
```

Alys:
I've restarted the test build.

lynxlynxlynx:
It was the same an hour ago, so I doubt it will make a difference. Digging around, especially seeing that this area was worked on recently, it is clear that it's a new problem, but not from this PR. Last release commit already had it, if not introduced it: https://travis-ci.org/HabitRPG/habitica/jobs/193943628

paglias:
@lynxlynxlynx the current `develop` branch is passing so try to update it in your local install and merge into this PR or rebase, let me know if you need help

lynxlynxlynx:
Right, seems I forgot to update in the meanwhile.

lynxlynxlynx:
No change with HEAD, the Amazon test still times out.

paglias:
Ok, it seems that sometimes it fails, we're looking into it

Il 22 Gen 2017 12:12 PM, "Jaka Kranjc" <notifications@github.com> ha
scritto:

paglias:
If I can ask one last thing, can you abstract 

```
+        $and: [ // exclude challenge and group tasks
 +          {
 +            $or: [
 +              {'challenge.id': {$exists: false}},
 +              {'challenge.broken': {$exists: true}},
 +            ],
 +          },
 +          {
 +            $or: [
 +              {'group.id': {$exists: false}},
 +              {'group.broken': {$exists: true}},
 +            ],
 +          },
```

in a static method inside the Task model, so that it's shared and not copied around?

paglias (approved):

paglias:
also we should understand why the existing tests aren't working and failing as expected like https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/user/POST-user_reset.test.js#L90

lynxlynxlynx:
I'll be in another country the whole week, so I can't do the refactor yet. However, the second question is easily answered: user/reset manually filters out challenge and group tasks. So perhaps one of the sets of filters is redundant — if you don't consider it an optimisation.

Alys:
I agree that abstracting the code to a method would be great and should be done, and I wouldn't normally suggest putting that off, but I would like to see this PR merged as soon as possible before the abstracting is done, since @lynxlynxlynx will be busy for a week. This issue is creating problems for challenge creators which will be affecting them for months to come for any long-running challenges that exist now - the challenge participants will be inadvertently adjusting challenge task values and deleting challenge To-Dos, which will make it much harder for challenge creators to assess winners. Since there are already so many bugs related to challenges, I'd like to get this one closed off asap.

If there are no significant complaints in the next day, I'll merge this in its current state.and create an important-priority issue for abstracting the code (which I can mark as being claimed by @lynxlynxlynx if they'd like that).

lynxlynxlynx:
Sure, you can assign the follow up assign it to me. Now off to conferencing ...

paglias:
@Alys yeah, you're right, I'm merging this in.

@lynxlynxlynx thanks for the help and yes, one of the filters is unnecessary unless it's used as a client side filter (but anyway it should be moved outside the /common code). Thanks for the help and happy conferencing :) I've notated this PR in your profile

Alys:
New issue created and marked as in progress: https://github.com/HabitRPG/habitica/issues/8452 ("abstract the code that checks for a task belonging to a challenge or a group plan")

@lynxlynxlynx Have a great time at the conference!
After you get back, let us know if something crops up so that you don't have time to work on it and we can open it up as help welcome.

Issue #7417, opened 2016-05-22
can't delete account if you use Facebook or Google authentication


Alys:
The delete account button now requires that you enter your password.

Players who use only Facebook authentication don't have a password. The delete feature needs to have an exception for them where they need to enter only "DELETE" (localised) as previously.

ghost:
For what it's worth, I believe "DELETE" hasn't been localised yet in this very case...

khipkin:
This may be addressed by #7969.

paglias:
@khipkin I don't think #7969 is addressing this (the ability to delete an account that only has social authentication)

khipkin:
@paglias Oh oops, my bad!

pllearns:
Just double checking, could I go ahead and work on this issue? I don't see a PR for it, and any recent updates. Thanks! I'll also update the Blacksmith Guild Page.  I will be working with @rachel-ftw on this.

TheHollidayInn:
Go for it. :D

pllearns:
Hey! We are making progress, we have created a condition that allows the user to enter the word 'DELETE' and that deletes their account. We intend for this to be the command for anyone who has logged in with Facebook and Google without creating a local user. We will write a test for this and should have a PR for review either later today or tomorrow. Thanks again for letting us take this on!

pllearns:
Oh, we did have a question, how have you mock logged in through the local dev site a Facebook or Google login? We would like to be able to test all edge cases. I didn't see anything on this in the wikia. Thank you!

paglias:
@pllearns the only way that comes to my mind is to register an app with google and / or facebook in order to get api keys that then you can use. But probably it's easier to just change the values into your local database to simulate the existence of google / facebook auth

Alys:
@pllearns Below are examples of the `auth` object for user accounts created using the Facebook and Google authentication methods. The `auth` object is at the top level of the player's document in the "users" collection in the mongodb database. You can create two user accounts in your test database using the username and password method, and then change the login method for each of them by using an "update" command like this (the "_id" value is the account's User ID):

db.users.update({"_id":"12345678-336b-4e53-9949-2c09150b4035"}, {$set:{   

    "auth": {
        "timestamps": {
            "loggedin": "2017-01-22T06:40:18.321Z",
            "created": "2015-10-21T22:59:01.582Z"
        },
			..... etc .....
    }

}}); 



Facebook authentication:
```
    "auth": {
        "timestamps": {
[...]

PR #8429, opened 2017-01-20
WIP: Wrote tests for deleting user with FB/Google auth. Still need to… { Continued in #8796 }

pllearns:
WIP: Wrote two tests for FB and Google user deletion, still need to update object generators correctly, these tests need a little more work. 

[//]: # (Put Issue #7417  or URL here, if applicable. This will automatically close the issue if your PR is merged in)

### Changes
[//]: # Added two tests for Google and Facebook authorized user deletion. Still working on the matching component/methods to test against. 

[//]: # 0976d018-2433-410c-8c6e-8dbb42160978

----
UUID:

### DIFF
--- a/test/api/v3/integration/user/DELETE-user.test.js
+++ b/test/api/v3/integration/user/DELETE-user.test.js
@@ -30,6 +30,20 @@ describe('DELETE /user', () => {
     });
   });
 
+  it('does not need a password if logged in through Facebook', async () => {
+    await user.auth.facebook.id.del('/user', {
+      user,
+    });
+    await expect(checkExistence('users', user._id)).to.eventually.eql(false);
+  });
+
+  it('does not need a password if logged in through Google', async () => {
+    await user.auth.google.id.del('/user', {
+      user,
+    });
+    await expect(checkExistence('users', user._id)).to.eventually.eql(false);
+  });
+
   it('returns an error if user has active subscription', async () => {
     let userWithSubscription = await generateUser({'purchased.plan.customerId': 'fake-customer-id'});
 


test/api/v3/integration/user/DELETE-user.test.js
```diff
@@ -30,6 +30,20 @@ describe('DELETE /user', () => {
     });
   });
 
+  it('does not need a password if logged in through Facebook', async () => {
+    await user.auth.facebook.id.del('/user', {
```
khipkin, line 34: > **`DELETE /user does not need a password if logged in through Facebook:`**
>     TypeError: Cannot read property 'del' of undefined
>      at _callee3$ (test/api/v3/integration/user/DELETE-user.test.js:34:11)

It looks like `user.auth.facebook.id` is undefined here.

test/api/v3/integration/user/DELETE-user.test.js
```diff
@@ -30,6 +30,20 @@ describe('DELETE /user', () => {
...
+      user,
+    });
+    await expect(checkExistence('users', user._id)).to.eventually.eql(false);
+  });
+
+  it('does not need a password if logged in through Google', async () => {
+    await user.auth.google.id.del('/user', {
```
khipkin, line 41: > **`DELETE /user does not need a password if logged in through Google:`**
>     TypeError: Cannot read property 'del' of undefined
>      at _callee4$ (test/api/v3/integration/user/DELETE-user.test.js:41:11)

It looks like `user.auth.google.id` is also undefined.

test/api/v3/integration/user/DELETE-user.test.js
pllearns: yep, I figured that, (hence the WIP) I'm still looking at how I should use the requester to get the right id. I was looking at something else when I made this test. I'll update. Thanks!

test/api/v3/integration/user/DELETE-user.test.js
pllearns: same comment as above. Thanks again for looking!

test/api/v3/integration/user/DELETE-user.test.js
TheHollidayInn: I think what you want to do here is just call `user.del('/user')`

And before that, you will want to generate a user with the auth.facebook.id. Which I think you said you are working on.

pllearns: Thanks, we are trying that and the test comes back with this error:
`path `auth.facebook.id` was not saved, as it's a protected path.`

Here is the updated test code: 

```js

  it('does not need a password if logged in through Social Network to delete account', async () => {
    await user.update({
      'auth.facebook.id': 'some-fb-id',
      'auth.google.id': 'some-google-id',
      'auth.local': { ok: true },
    });

    await user.del('/user', {
    });

    await user.sync();
    await expect(checkExistence('users', user._id)).to.eventually.eql(false);
  });
```

TheHollidayInn: Yea, you will have to alter the generator (which uses the Mongoose Model directly, I believe). You will not be able to update the id via the user update route.

pllearns: Thanks for that. Sorry for the late reply to this. I will go ahead and work on the generator. I'll update soon.

SabreCat:
This WIP hasn't seen any updates in some time. I'm going to close it for now; @pllearns, if you'd like to resume, you need only say so!

PR #8434, opened 2017-01-20
added test case for #8423 (searing brightness should not affect challenge tasks)

lynxlynxlynx:
split off #8427

### DIFF
--- a/test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
+++ b/test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
@@ -2,6 +2,7 @@ import {
   generateUser,
   translate as t,
   createAndPopulateGroup,
+  generateGroup,
   generateChallenge,
   sleep,
 } from '../../../../helpers/api-integration/v3';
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
There you have it, the test fails, 0.5957446808510638 vs 0.

paglias:
Can this test go into #8427? The fact that current tests weren't failing means they weren't written correctly to spot these bugs and I'd like to have one working once #8427 is merged

paglias:
actually never mind, I'll merge both at the same time, thanks a lot for this!

TheHollidayInn:
IF I am reading this correctly, this test only fails without the change. The test won't fail once the change is made (which is already tested for in the other PR). So, we won't merge this in, right?

paglias:
We'll merge it to be sure this case is covered by tests and in future if
this bug comes back we'll notice it immediately. The problem is that
current tests around this functionality weren't working as expected (I'm
still unsure why)

Il 23 Gen 2017 16:15, "Keith Holliday" <notifications@github.com> ha
scritto:

PR #8435, opened 2017-01-20
Incentives, Batch 2

SabreCat:

### Changes

Implements check-in incentive rewards from 55 days through 100.

Because more than 55 days have passed since incentives were implemented, also includes a migration to award the missed incentive. The migration also cuts down excessive check-in counts (likely resulting from custom day start abuse) to the maximum temporally possible. It depends on @TheHollidayInn's work in #8436.


website/common/script/content/loginIncentives.js
```diff
@@ -165,7 +243,7 @@ module.exports = function getLoginIncentives (api) {
   });
 
   let prevRewardKey;
-  _.range(51).forEach(function addPrevRewardLink (index) {
+  _.range(MAX_INCENTIVES).forEach(function addPrevRewardLink (index) {
```
TheHollidayInn: Thank you :P for fixing my bad magic number.

website/common/script/content/loginIncentives.js
```diff
@@ -145,13 +147,89 @@ module.exports = function getLoginIncentives (api) {
         user.items.food.Saddle += 1;
       },
     },
+    55: {
+      rewardKey: ['Pet_HatchingPotion_Purple'],
+      reward: [api.hatchingPotions.RoyalPurple],
+      assignReward: function assignReward (user) {
+        if (!user.items.hatchingPotions.RoyalPurple) user.items.hatchingPotions.RoyalPurple = 0;
```
khipkin, line 153: This logic is repeated many, many times.

website/common/script/content/loginIncentives.js
```diff
@@ -1,5 +1,7 @@
 import _ from 'lodash';
 
+const MAX_INCENTIVES = 101;
```
khipkin: cron.js has another const with the same name, but set as 100 instead. Should this be pulled out into a more globally-available constant?

TheHollidayInn: Yea, let's export this and use it in Cron.

website/common/script/content/loginIncentives.js
TheHollidayInn: Yes, but let's delay the abstraction. I want to fix this in a more generic way. I'm postponing this until we redo some of the way we add content.

SabreCat:
How's that for tidying up the constant @TheHollidayInn @khipkin? I'm not sure about the 100 vs. 101 discrepancy (which I preserved with an add-one); THI may need to explain what's up there.

website/common/script/content/loginIncentives.js
SabreCat: We could in theory remove the check from all of them past the first, on the assumption that once you've got at least one potion you'll have the 0 value already. But that may not be safe.

website/common/script/content/loginIncentives.js
```diff
@@ -243,7 +242,7 @@ module.exports = function getLoginIncentives (api) {
   });
 
   let prevRewardKey;
-  _.range(MAX_INCENTIVES).forEach(function addPrevRewardLink (index) {
+  _.range(MAX_INCENTIVES + 1).forEach(function addPrevRewardLink (index) {
```
TheHollidayInn, line 245: Looks good, although I am weary about this +1 now, but I don't think it should hold us up.

SabreCat:
Can't import from `common` into a server-side-only script. So I put it back to the way it was previously.

website/common/script/content/loginIncentives.js
khipkin: Delaying would be safer IMO.

khipkin:
@SabreCat Much tidier! Thanks. 😊

PR #8446, opened 2017-01-22
Migrate to bcrypt

paglias:
- [x] Migrate schema
- [x] Migrate register route
- [x] Migrate login route
- [x] Migrate update password route
- [x] Migrate update username route
- [x] Migrate update email route
- [x] Migrate reset password route
- [x] Migrate delete user route
- [x] Set all current users' auth.local.passwordHashMethod to `sha1` (or default to it)
- [x] Migrate existing users password to bcrypt when they login
- [x] Tests
- [x] Use native bcrypt, see https://github.com/kelektiv/node.bcrypt.js/issues/476


website/server/controllers/api-v3/auth.js
```diff
@@ -584,8 +608,14 @@ api.updateEmail = {
 
     if (emailAlreadyInUse) throw new NotAuthorized(res.t('cannotFulfillReq'));
 
-    let candidatePassword = passwordUtils.encrypt(req.body.password, user.auth.local.salt);
-    if (candidatePassword !== user.auth.local.hashed_password) throw new NotAuthorized(res.t('wrongPassword'));
+    let password = req.body.password;
```
khipkin, line 612: Suggestion: pull this logic into a `verifyPassword` helper method that checks the password, throws an error if the password is wrong, and converts to the new encryption schema if needed.

website/server/libs/password.js
```diff
@@ -1,18 +1,63 @@
...
+import bcrypt from 'bcrypt';
+
+const BCRYPT_SALT_ROUNDS = 10;
+
+// Hash a plain text password
+export function bcryptHash (passwordToHash) {
+  return bcrypt.hash(passwordToHash, BCRYPT_SALT_ROUNDS); // returns a promise
+}
+
+// Check if a plain text password matches an hash
```
khipkin: a hash

website/server/controllers/api-v3/auth.js
paglias: I'm not sure I agree with this:

1. `Throwing the error in the compare function`:  we may want to show a different error in some situations and right now there's a piece of code where such a change would result in a duplication of the code that throws the error

```
    let isValidPassword;

    if (!user) {
      isValidPassword = false;
    } else {
      isValidPassword = await passwordUtils.compare(user, password);
    }

    if (!isValidPassword) throw new NotAuthorized(res.t('invalidLoginCredentialsLong'));
```

2. Converting to bcrypt while comparing: I would prefer not because we're talking about two very different actions: one is checking for the validity of the password and the other is converting data to something different (and requires a save afterwards). And for example when deleting an user account we don't care about converting the password to bcrypt before it's destroyed

website/server/libs/password.js
paglias: 👍 thanks!

website/server/controllers/api-v3/auth.js
khipkin: I meant a (private) helper method within auth.js, not in the hash lib.

paglias: Ohh that could make sense

paglias: I've tried this approach but at the end decided against it because it was even less straightforward because there are maximum two places where the exact same code could be used, the others needed some tweaks

paglias:
This is done, added some integration tests. Will wait for the build to pass and then merge

website/server/controllers/api-v3/auth.js
khipkin: Okay! :)

PR #8400, opened 2017-01-12
Updating API Documentation for Tasks and Tags

madpink:
[//]: # (Note: See http://habitica.wikia.com/wiki/Using_Habitica_Git#Pull_Request for more info)

https://github.com/HabitRPG/habitica/issues/8087

### Changes
Beginning the processes of updating Tasks
Added Body parameters
Added Request example
Added Success example
Added Failure example
Expanded error types

[//]: # (Put User ID in here - found in Settings -> API)

----
UUID:

PR #8447, opened 2017-01-22
Updating Tasks and Tags API Doc

madpink:
Okay, this was originally:  https://github.com/HabitRPG/habitica/pull/8400
But I royally messed up my local repo so I started fresh.

The issue: https://github.com/HabitRPG/habitica/issues/8087

I've updated the API documentation for Tasks and Tags
----
UUID: b0413351-405f-416f-8787-947ec1c85199

### DIFF
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -774,6 +774,7 @@ api.scoreCheckListItem = {
  *
  * @apiUse TaskNotFound
  * @apiUse ChecklistNotFound
+ * @apiUse ChallengeNotFound
  */
 api.updateChecklistItem = {
   method: 'PUT',
@@ -1159,6 +1160,9 @@ api.clearCompletedTodos = {
  *
  * @apiUse TaskNotFound
  * @apiError (400) {NotAuthorized} Challenge A task belonging to a challenge can't be deleted.
+ * @apiError (400) {NotAuthorized} Group Can't delete group tasks that are assigned to you
+ * @apiError (400) {NotAuthorized} ChallengeLeader Tasks belonging to a challenge can only be edited by the leader.
+ * @apiError (400) {NotAuthorized} GroupLeader Not authorized to manage tasks!
  */
 api.deleteTask = {
   method: 'DELETE',
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -35,7 +35,7 @@ import logger from '../../libs/logger';
 
 /**
  * @apiDefine NotAuthorized
- * @apiError (400) {NotAuthorized} There is no account that uses those credentials.
+ * @apiError (401) {NotAuthorized} There is no account that uses those credentials.
  */
 
 let api = {};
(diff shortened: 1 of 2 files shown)


website/server/controllers/api-v3/tasks.js
```diff
@@ -578,11 +765,15 @@ api.scoreCheckListItem = {
...
+ * @apiParam (body) {string} text The text that will replace the current checkitem text.
+ *
+ * @apiParamExample {json} Example body:
+ * {"text":"Czech 1"}
+ *
  * @apiSuccess {Object} data The updated task
  *
  * @apiUse TaskNotFound
  * @apiUse ChecklistNotFound
- * @apiUse ChallengeNotFound
```
paglias, line 585: why was challenge not found removed?

website/server/controllers/api-v3/tasks.js
```diff
@@ -920,10 +1144,13 @@ api.clearCompletedTodos = {
...
  * @apiParam {String} taskId The task _id or alias
  *
+ * @apiExample {json} Example call:
+ * curl -X "DELETE" https://habitica.com/api/v3/tasks/3d5d324d-a042-4d5f-872e-0553e228553e
+ *
  * @apiSuccess {Object} data An empty object
  *
  * @apiUse TaskNotFound
- * @apiUse ChallengeNotFound
+ * @apiError (400) {NotAuthorized} Challenge A task belonging to a challenge can't be deleted.
```
paglias: this should also mention `NotAuthorized(res.t('cantDeleteAssignedGroupTasks'));` and that original group and challenge tasks can only be deleted by the leader

website/server/controllers/api-v3/tasks.js
madpink: I wasn't able to produce the error, do you have a recipe for it?

website/server/controllers/api-v3/tasks.js
madpink: agreed, will add

madpink: added errors for unauthorized group, unauthorized challenge and can't delete group task

website/server/controllers/api-v3/tasks.js
TheHollidayInn: Challenge not found is triggered on line 804

madpink: I see the code, but I can't produce the error. In order to throw the error, challengeId must have data in it, but turn out to have an invalid Id... I've tried creating a challenge, creating the task, then deleting the challenge, but it still passes. Under what circumstances would there be a challengeId in a task that does not return a valid challenge?

paglias: It should never happen but it could in case the challenge was removed from the database but its task were not

madpink: To be honest, this raises more questions for me... if a task was linked to a challenge and that challenge no longer exists, that means there is no way to delete that task. Wouldn't it make more sense to allow the task to be deleted instead of throwing an error that can't be corrected? 

I will add it back in, but I am adding a note to try and explain it better because I do not think that the error message makes sense.

madpink: added back

website/server/controllers/api-v3/tasks.js
paglias: thanks, NotAuthorized is 401 non 400

madpink: Updated

paglias:
This look good to me. @SabreCat @TheHollidayInn

paglias:
Thanks for the help, I've awarded your first tier :)

Issue #8385, opened 2017-01-08
Email addresses should not probably be translatable


ghost:
Following the discussion in the i18 guild...

> ***Alys***  Also for Czech translators ... : In the translations for the Community Guidelines (communityguidelines.json), there's some problems with email addresses and URLs. In at least two places, the leslie@habitica.com email address has been written incorrectly in both the mailto: portion and visible portion of email links (for example, <a href=\"mailto:leslie@\">leslie@habitica</a>)...

Although I haven't managed to find mistakes in the mailto part of the strings, for consistency and security reasons may we have email addresses as tokens so that translators could not change them?

E.g.: instead of
`If you have an issue or concern about a particular Mod, please send an email to Lemoness (<a href="mailto:leslie@habitica.com">leslie@habitica.com</a>).`
introduce something like
`If you have an issue or concern about a particular Mod, please send an email to Lemoness (<%= email %>).`

FYI @Alys @paglias 

----

_edit by Alys 2017-01-16:_

We'll use environment variables (i.e., new items in config.json) to specify the email addresses and then use tokens in the locales strings to specify where the addresses should be placed. See the comments below for more details.

Alys:
That's a really good idea!

The <%= email %> token could be used to insert a value from an environment
variable so that we could change it easily when needed (e.g., <%=
communityManagerEmail %> and change the value when/if Lemoness goes on
holiday).

If there's no disagreements for this idea, I can make a list of all current
addresses in the strings and then we can plan suitable variable names for
them.

Also I'll track down the incorrect mailto links that (I think) I found.

khipkin:
Suggestion: [translations] and not sure if medium- or entry-level-coding?

Alys:
Below are all the locales strings that contain email addresses, separated into sections based on what the email addresses are used for, with my suggestions for the environment variable names that we use for each address.

@paglias @SabreCat  Do you have any objections to using ENV variables for this? Or better ideas?

----
suggested ENV variable: `PRESS_ENQUIRY_EMAIL`
current value: leslie@habitica.com

```
front.json:    "presskitText": "Thanks for your interest in Habitica! The following images can be used for articles or videos about Habitica. For more information, please contact Siena Leslie at leslie@habitica.com.",
```

----

suggested ENV variable: `TECH_ASSISTANCE_EMAIL`
current value: admin@habitica.com

```
[...]

PR #8448, opened 2017-01-23
Remove email addresses from translatable strings

gsacavdm:
Fixes https://github.com/HabitRPG/habitica/issues/8385

### Changes
Replace emails in strings for replacements and updated jade/js/tests to fetch the email from the environment variables and pass it in for replacement.

----
UUID:  5783fd4a-dac9-466e-afea-15134a7365d9


gsacavdm:
Need some help with the test, @SabreCat, @TheHollidayInn  maybe you can help or point me the right direction.
I've got three failures:

**Failure 1- test:api-v3:paymets: amazon #subscribeCancel success cancels a user subscription**
**Assessment thus far: Unstable test and ignore?**
> Error: timeout of 8000ms exceeded. Ensure the done() callback is being called in this test.

When I run locally, it generally passes, it takes about ~7000 ms to run, it sporadically fails, so it just seems like a slow test. This guy also fails for me sporadically on the unaltered develop branch.

**Failure 2 - test:karma**
**Assessment thus far: No clue yet, still investigating**
> TypeError: undefined is not a function (evaluating 'n.readdirSync(r+'/nconf/stores')')
  at /home/travis/build/HabitRPG/habitica/website/client-old/js/habitrpg-shared.js:9
>
>✔ 0 tests completed
PhantomJS 2.1.1 (Linux 0.0.0) ERROR
  TypeError: undefined is not an object (evaluating 'window.habitrpgShared.i18n')
  at /home/travis/build/HabitRPG/habitica/website/client-old/js/env.js:9
✖ Error while running the tests! Exit code: 1
PhantomJS 2.1.1 (Linux 0.0.0) ERROR
  TypeError: undefined is not an object (evaluating 'window.habitrpgShared.TAVERN_ID')
  at /home/travis/build/HabitRPG/habitica/website/client-old/js/app.js:9
✖ Error while running the tests! Exit code: 1

Reproduces locally but I'm not sure what's going on here. This is happening with the unaltered develop branch too.

**Failure 3 - client:unit**
**Assessment thus far: Need a way to use nconf from faqs.js without breaking these tests, still investigating options.**
> PhantomJS 2.1.1 (Linux 0.0.0) ERROR
>  Error: Cannot find module "fs"
>  at webpack:///~/nconf/lib/nconf.js:8:0 <- index.js:50425

This seems to be a result of me adding *import nconf from 'nconf'* in faq.js. Any ideas on how I can leverage nconf in faq.js without this issue?

website/common/script/content/faq.js
```diff
@@ -1,4 +1,5 @@
 import t from './translation';
+import nconf from 'nconf';
```
TheHollidayInn: I don't think you can use nconf here, because it will not work. You need to include config variables on the window object if it is use in the client code: https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/locals.js

TheHollidayInn:
The first test should be fixed now if you merge in develop.

For the other two, you are trying to use server code on shared and it will not work on the client side.

website/common/script/content/faq.js
gsacavdm: I removed it and reverted back to hardcoded values for this file for now. The challenge is that this file, being common, is used both by the client and the server and I'm having a hard time figuring out how to access values from config.json for each of those scenarios for a single file.

gsacavdm:
@TheHollidayInn would you guys be open to merging this and creating a separate issue just for faq.js. 

As it stands, there's value in having emails excluded from localization and managed by a consistent set of config values. The only value email that breaks this rule (because it's hardcoded) is the faq.js which, from what I can tell is used from the Vue client implementation and api/v3/content.

I'm happy to keep iterating on this, but I'd like to minimize how many files I'm modifying and have keep merging.

test/api/v3/integration/user/auth/POST-login-local.test.js
```diff
@@ -43,7 +45,7 @@ describe('POST /user/auth/local/login', () => {
     })).to.eventually.be.rejected.and.eql({
       code: 401,
       error: 'NotAuthorized',
-      message: t('accountSuspended', { userId: user._id }),
+      message: t('accountSuspended', { communityManagerEmail: nconf.get('EMAILS:COMMUNITY_MANAGER_EMAIL'), userId: user._id }),
```
paglias, line 48: to avoid calling nconf.get too many times you should store the result in a constant at the beginning of the file and then always re-use it

paglias (changes requested):

TheHollidayInn:
I'm fine with a separate PR for Faqs. Just make sure to remove your FAQ changes from this PR.

test/api/v3/integration/user/auth/POST-login-local.test.js
gsacavdm: I can certainly do that, though as it stands it's only called once in this whole test, or am I missing something?

If you want me to do this regardless, does adding the following in line 18 suffice (and modifying this line to use that variable instead)?
```javascript
let communityManagerEmail = nconf.get('EMAILS:COMMUNITY_MANAGER_EMAIL');
```

Lastly, I was only planning on doing this for the two test files:
* POST-login-local.test.js
* PUT-user_update_email.test.js

Or did you also want this in the two auth.js files? Doesn't seem like would add much value doing it there.

website/common/script/content/faq.js
```diff
@@ -14,7 +14,11 @@ for (let i = 0; i <= NUMBER_OF_QUESTIONS; i++) {
   let question = {
     question: t(`faqQuestion${i}`),
     ios: t(`iosFaqAnswer${i}`),
-    web: t(`webFaqAnswer${i}`),
+    web: t(`webFaqAnswer${i}`, {
+      //  TODO: Need to pull these values from nconf
```
gsacavdm, line 18: @TheHollidayInn do you want me to revert this as well as the changes in faq.json?

I was thinking it was better as I have it since you still get benefit of decoupling the emails from the localization strings, the gotcha being that while I sort out how to pull from nconf here, these two emails (techAssistance and wikiTech) will be managed in two places: here and in the config.json file.

If I revert the changes here and the ones in faq.json, those two emails will still need to be managed in two places: config.json file and faq.json where they're hardcoded but with the added drawback that faq.json will still not have the emails decoupled from the localization strings.

So, what would you rather me do? Keep this and faq.json as is, or revert changes on those two files?

test/api/v3/integration/user/auth/POST-login-local.test.js
paglias: sorry, I meant to post this comment in the two auth.js files, doing it in the tests is not necessary indeed

website/common/script/content/faq.js
TheHollidayInn: You can keep these. Are those the correct emails though?

Alys: techassistance@habitica.com isn't correct. It should be admin@habitica.com.

I've checked the use of the env variables through all these changes and they're all good.

gsacavdm: Good catch. I've updated these.

test/api/v3/integration/user/auth/POST-login-local.test.js
gsacavdm: Done!

gsacavdm:
@paglias , @TheHollidayInn ready for approval (hopefully).

gsacavdm:
Ping @paglias , @TheHollidayInn

website/server/middlewares/auth.js
```diff
@@ -4,6 +4,9 @@ import {
 import {
   model as User,
 } from '../models/user';
+import nconf from 'nconf';
+
+const COMMUNITY_MANAGER_EMAIL = nconf.get('MAINTENANCE_MODE');
```
paglias: this should be `nconf.get('MAINTENANCE_MODE');`

website/common/script/content/faq.js
paglias: Just a question: why we're not using env vars here?

paglias (changes requested):
@gsacavdm so sorry about this, but I just noticed a place where the email isn't loaded correctly from nconf and it'll have to be fixed before I can merge it

website/server/middlewares/auth.js
gsacavdm: I assume you meant `nconf.get('EMAILS:COMMUNITY_MANAGER_EMAIL');`
Good catch. I fixed this.

website/common/script/content/faq.js
gsacavdm: That's the goal, but given that these files are shared between the client and the server, I'm having a hard time figuring out a single way to access the env variables.

process.env doesn't have them.
env is not defined.
nconf is for server side only.

Seems like I need to pass these in when loading common... My ask was to get this checked in as is as it adds value and then proceed to figure out how to handle these.

gsacavdm:
@paglias , fixed the incorrect  nconfg call and answered your question. Good to merge?

website/common/script/content/faq.js
paglias: Ah okay... with the current client side the only way to access env vars is in Jade templates but not here in a JS file. I'll be sure to add it to the rewrite of the client side that we're doing

paglias:

PR #8451, opened 2017-01-24
fix(nextRewardUnlocksIn): Check-in prize message was always plural --…

### LINKED ISSUE #8428: Check-in prize message shouldn't always be plural
The check-in prize message for when the next prize will be awarded is always pluralised, even when there's only one day until the next prize. When there's only one day left, it should be singular, e.g. "Your next prize will unlock in 1 more check-in!"

Apologies if this is already known or if a fix is planned.

image

deonna:
Fixes #8428

### Changes
[//]: # 

Check-in prize message was always plural -- moving to a non-sentence like structure to fix incorrect grammar.

----
UUID:

### DIFF
--- a/website/common/locales/cs/loginincentives.json
+++ b/website/common/locales/cs/loginincentives.json
@@ -1,7 +1,7 @@
 {
   "unlockedReward": "Obdržel/a jsi <%= reward %>",
   "earnedRewardForDevotion": "Získal/a jsi <%= reward %> za snahu zlepšit se ve svém životě.",
-  "nextRewardUnlocksIn": "Your next prize will unlock in <%= numberOfCheckinsLeft %> more check-ins!",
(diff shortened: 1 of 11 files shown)


khipkin:
@deonna Localization should be done through Transifex, not Github. Please revert all changes not in the `en` locale files! You can find more info in the "locales" directory's README.

Other than that, LGTM. 👍 😄

deonna:
@khipkin -- will do! I've never worked with Transifex before. Any ideas for getting started and what directories or config files I should be looking at?

Alys:
@deonna All you need to do is update files under under the /en/ directory. The changes will  be semi-automatically deployed to Transifex by admins and the translators will take it from there. :) If you're curious about the process or would like to be a translator, see http://habitica.wikia.com/wiki/Guidance_for_Linguists

Alys:
@deonna Let us know if you'd like advice about how to revert the non-English locales files! :)

PR #8456, opened 2017-01-25
Added double columns to edit modal

### LINKED ISSUE #8307: New Edit functions requires more scrolling on smaller screens
## General Info

User Id: bc10ec41-4213-4482-bbb8-558e3c3ea8d5
Opera 42.0.2393.85
Windows 10 32-bit
Screen Size: 1368 X 768

## Description

With the new edit function I now I have to scroll much more. As 
+ the text is much bigger and therefore I can't see all my task details without scrolling 
+ There is no save option at the top of the edit task now. {EDIT by Alys: a save button will be added in https://github.com/HabitRPG/habitica/issues/8316}

I realise you have the defaults for tags and advance options hidden, however 60% of the time I am editing the advance options.

Could we please :
+ make the text smaller? and possible use 2 columns. Most website screens are wide. (except on phones) and 
+ put a save delete option at the top again?  {EDIT see https://github.com/HabitRPG/habitica/issues/8316}

Thanks

----

_edit by Alys: This issue is now for discussing whether we should make the text smaller and/or use two columns._

needtoscrolltoedit

TheHollidayInn:

Fixes #8307

@Alys 

### Changes
double-column-layout

----
UUID:

### DIFF
--- a/website/client-old/css/shared.styl
+++ b/website/client-old/css/shared.styl
@@ -115,3 +115,6 @@ li.spaced
 
 .slight-right-margin
   margin-right: 1em
+
+.modal-footer
+  border: none
--- a/website/client-old/css/tasks.styl
+++ b/website/client-old/css/tasks.styl
@@ -51,6 +51,8 @@ taskContainerStyles($stage)
     background-color: $stage[1]
   .option-group:not(.task-checklist)
     border-bottom: 1px solid darken($stage[1], 15%)
+  hr
+    border-bottom: 1px solid darken($stage[1], 15%)
   .option-content
     border-color: darken($stage[1], 16.18%) !important
     &:hover
@@ -727,3 +729,7 @@ form
 // Add padding to exp-chart
 .exp-chart
   padding-bottom: 10px
+
+.task-extra-notes
+  z-index: 10
+  position: relative
--- a/website/views/shared/modals/tasks-edit.jade
+++ b/website/views/shared/modals/tasks-edit.jade
@@ -1,5 +1,5 @@
 script(type='text/ng-template', id='modals/task-edit.html')
-  .modal-content.task-modal(style='min-width:28em', class='{{taskStatus}}', id="task-{{task._id}}")
+  .modal-content.task-modal(style='min-width:22em', class='{{taskStatus}}', id="task-{{task._id}}")
     .modal-body.text-center(style='padding-bottom:0')
       | {{scope}}
       include ../tasks/edit/index
--- a/website/views/shared/tasks/edit/advanced_options.jade
+++ b/website/views/shared/tasks/edit/advanced_options.jade
@@ -9,8 +9,8 @@ div(ng-if='(task.type !== "reward") || (!obj.auth && obj.purchased && obj.purcha
       a.hint(href='http://habitica.wikia.com/wiki/Task_Alias', target='_blank', popover-trigger='mouseenter', popover="{{::env.t('taskAliasPopover')}} {{::task._edit.alias ? '\n\n\' + env.t('taskAliasPopoverWarning') : ''}}")=env.t('taskAlias')
     input.form-control(ng-model='task._edit.alias' type='text' placeholder=env.t('taskAliasPlaceholder'))
 
-  fieldset.option-group.advanced-option(ng-show="task._edit._advanced")
-    group-tasks-actions(ng-if="!obj.auth && obj.purchased && obj.purchased.active", task='task', group='obj')
+  fieldset.option-group.advanced-option(ng-show="task._edit._advanced", ng-if="!obj.auth && obj.purchased && obj.purchased.active")
+    group-tasks-actions(task='task', group='obj')
 
   div(ng-show='task._edit._advanced')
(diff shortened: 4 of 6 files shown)


Alys:
Looks pretty good!

For anyone curious, here's what it looks like when Advanced Options is collapsed:

image

Alys:
One problem though is that when you expand the note field, it sits underneath the Advanced Options fields (see below). Can that be fixed with a z-index or something in the styles?

image

Alys:
@cTheDragons @Castironowl @mandyzhou @GrayDays : I've put this change on a test server that I own at http://192.241.249.166:3000

Can you please test there on your mobile browsers and give us any feedback that you have? You won't see the two-column format there because your screen likely won't be wide enough but the fields in the single column should be a little narrower. Put any comments in THIS pull request, not in the original issue because you'll be discussing the specific change made in this pull request.

To use that website you will have to create a NEW account (you will  NOT have access to your normal account from there). IMPORTANT: That is NOT a secure server! Do not use the same username and password that you use on the real Habitica site and use a fake email address (e.g., alskkdjf@lkskdjf.com). Do not put real tasks or other personal information into that website. However it's completely safe for testing with a test account! You will  not be able to use Google sign-in or Facebook sign-in to create an account.

mandyzhou:
@Alys Thanks for setting up the test server! Unfortunately, I'm getting an ERR_CONNECTION_TIMED_OUT error when I try to access it.

Alys:
@mandyzhou Sorry about that - it had stopped running while I was away from home. It's back up now and has the latest fix from @TheHollidayInn (an expanded notes field is no longer under the Advanced Options and the grey line that was above the Title field is gone).

mandyzhou:
Thanks, @Alys! Getting a ERR_CONNECTION_REFUSED now but I was able to test it briefly a few hours ago.

I can confirm that the field widths are slightly narrower, but horizontal scrolling is still required for the entire width of the fields to appear on my screen.

I also have a few thoughts on the two columns:
- I think they seem a little unbalanced. Maybe moving the Tags section to the right would help? 
- The two columns don't save that much vertical space for me, maybe because some elements now take up more vertical space (e.g., Repeat Every and Difficulty options take up more rows). 
- There seem to be further opportunities to save on vertical space by removing double horizontal lines and reducing vertical padding (e.g., there's a lot of extra padding under the Repeat Every options).

Alys:
@mandyzhou Thanks for the feedback and sorry - the server had fallen over again. It's up now.

cTheDragons:
Finally got around to checking it myself and agree with @mandyzhou that it is slightly unbalanced.
It would be best the tags be moved to the left.

Here are some screen shots from my computer.
tagsfurled
tagscollapse

As you can see even with the tags collaspse still takes up quite of bit of room.

On the mobile the text is still too wide.
Here is the visual from my samsung s4 galaxy active.

screenshot_2017-01-29-16-17-51

cTheDragons:
Another thought I had, and not sure if it is possible, if we move the tags across maybe we can add an extra line to Extra Notes so it 4 lines down

TheHollidayInn:
I'm trying to focus on the major issues atm so we can get this out. Thanks for the feedback, everyone!

Just posted an updated that allows the modal to be shrunk more and moves the tags to the right.

mandyzhou:
Looks great on my phone now! Thank you! 

Another (non-major/non-urgent) thought: the daily modal has the Repeat Every option in both columns. When it's just one column, it looks okay, but now that they're almost side by side, the second one looks more redundant.

TheHollidayInn:
I merged in to get the main fix out. We can leave the other items to discussion - and I'll add them if they are agree on. Also, I think the duplicated Repeat Every was mentioned somewhere else as well.

GrayDays:
Just got the chance to try the test server – sorry for the wait.

Having the same issue as cTheDragons where the edit modal still isn't sized quite right on mobile. It did let me zoom out a bit this time, but this was as small as I could make it: http://i.imgur.com/zojA502.jpg

Otherwise, it looks good.

I'm also still noticing the slingshotting issue that I mentioned in the other thread, though it seems not to be as bad with the modal at a more suitable size. 

A weird thing: if I add checklist items in edit mode and then try to scroll down, it scrolls the background instead of the edit modal until I zoom out again, at which point it scrolls properly.

After testing on both my current phone and my old one, it seems that all of these issues occur on both iOS 10.1.1 and 10.2.

TheHollidayInn:
I'm not sure the test server has my latest commit.

SabreCat:
Yeah, that's not a standard test server--looks like it was something @Alys threw together on a short-term basis. This PR is live at production habitica.com now, though, so you can see there if you're able to reproduce your difficulties.

Alys:
Yeah, the test server was updated for @TheHollidayInn's change 22 hours ago (from [his comment above](https://github.com/HabitRPG/habitica/pull/8456#issuecomment-276226707)) but not since then. It's turned off now. 

Any future comments about this should be put in the issue (https://github.com/HabitRPG/habitica/issues/8307) if it's not fully fixed.

Issue #7469, opened 2016-05-24
Challenge Extra Notes Missing in creator's copies of tasks after creating a challenge


omtigger:
General Info
- UUID: 84c14a15-4d8a-4bf7-a10b-e6bd3ddf0fa1
- Browser: Firefox 
- OS: Windows 10
### Description

Extra notes made in challenge creation are not showing when I go to my "Tasks" screen. Tried reloading tasks, and edited challenge (needed to fix tag anyway) and saved again.

Alys:
As far as I can tell from my testing, this happens only when a challenge creator first saves the challenge - the tasks that are added to the creator's task list at that time don't have anything in the "Extra Notes" field. 

If the challenge creator leaves the challenge, choosing to delete the task, and then rejoins the challenge, the new copies of the tasks do have notes.

If anyone else joins the challenge, their own copies of the tasks do have notes.

Alys:
related: https://github.com/HabitRPG/habitrpg/issues/7449 "Challenge Tags are not automatically associated with Challenge Tasks when creating new challenges"

Alys:
I'm closing this because we've decided to do this instead: https://github.com/HabitRPG/habitica/issues/9248 "challenge creator should not automatically join their own challenge"

Issue #7449, opened 2016-05-23
Challenge Tags are not automatically associated with Challenge Tasks when creating new challenges


cTheDragons:
### Description

When you create new challenge the tasks that appear in the your own, that is the challenge owner's, task list due to the challenge are not automatically associated with the challenge tag.

The tasks themselves are still connected to the challenge are marked as challenge tasks. They are still removed automatically if you choose to remove the tasks when leaving or if the challenge is finished.
## What to do if this happens to you:

This occurs only for the challenge creator when they first create a challenge - their own copies of the challenge tasks are not associated with the challenge tag. The creator can be certain that when other players join the challenge, those players will find that the tag works correctly.

The challenge creator can work around this problem by leaving their own challenge (they'll still remain as the owner!), choosing to delete the challenge tasks. They can then re-join the challenge and they'll find that the tag will then work for them.

makkoi:
I am also experiencing this error. 

user ID: b3e66733-ab6f-432f-a052-6fa64585da5a
Browser: Google Chrome Version 50.0.2661.102 (64-bit)
OS: Mac OS X Version 10.9.5

cTheDragons:
Should note if you are joining a challenge the tags appear to work correctly.

Altariel3:
I am also experiencing this error for challenges that I joined, that I don't own.  If a task is added after I joined the challenge than the challenge tag isn't associated with the task.

user ID: be317a32-c027-4032-bbd5-25ccb21eb912
Browser: Google Chrome Version 50.0.2661.102 m
OS: Windows 10, 64-bit

kawunngg:
same thing for me: in all challenges i create, the tags aren't connected to the tasks. this only happens in my own challenges. if i leave the challenge and join again, they are connected again. (tried once). challenges by other people seem to behave normal.

ce935e7f-00e6-44bc-a44a-5280553810f0
Win7 / Firefox 46.0.1

Habitica user aust said they have the same problem, don't know what system they use.
90ad1ff2-b48b-406a-b8a9-2b2415a6a3af

kawunngg:
oh, another thing i noticed, don't know if this is connected: when i create challenges the tasks appear on top of my to do list, if i join other people's challenges, they are added on the bottom of the list as usual.

Alys:
related: https://github.com/HabitRPG/habitrpg/issues/7469 "Challenge Extra Notes Missing in creator's copies of tasks after creating a challenge"

[...]

PR #8469, opened 2017-02-02
Added support for grouping tasks by challenge

TheHollidayInn:

### Changes
Adds a user setting if the user wants to group tasks by challenge name

group-tasks-by-challenges-setting
tasks-grouped-by-challenge

----
UUID:


khipkin:
Suggestion: make the casing of the options strings consistent (either don't use title case for the new option, or make them all use title case).

website/server/models/task.js
```diff
@@ -58,6 +58,7 @@ export let TaskSchema = new Schema({
   userId: {type: String, ref: 'User', validate: [validator.isUUID, 'Invalid uuid.']}, // When not set it belongs to a challenge
 
   challenge: {
+    name: {type: String, default: 'No Category'},
```
khipkin: The "No Category" is weird here, IMO. Maybe use "None" or "Private" or something instead?

Alys: "Personal"? "No challenge"?

I wonder if someone would create a challenge called "Private" or "Personal". I guess they might if they felt that was a meaningful title, or they might use a title that started with one of those words. "None" or "No challenge" might be better options to minimise the chance of confusion.

Alys:
@TheHollidayInn  Exciting idea! I think this will be popular! Is this PR still a work in progress though? I'm getting some odd behaviour locally but I'll stop playing around with it if you're still working on the feature. :)

TheHollidayInn:
@khipkin I don't understand what you are referencing when talking about casing.

@Alys Feel free to tell me the odd behavior :P

Alys:
@TheHollidayInn Sorry, I didn't want to list problems if it was still in development! :) When I try it on my local install with the latest `develop`, challenge tasks aren't put into their own groups (they're all under "No Category") and when I go to Social > Challenges to view individual challenges that I've created, all tasks are invisible there in both view mode and edit mode. There might be recent commits that are interfering with the new code - anything from after that screenshot was taken.

Regarding casing, it should be "Group tasks by challenge title" for consistency with the other options.

Alys:
Ah, this problem:
"challenge tasks aren't put into their own groups (they're all under "No Category")"
is because of this bug: 
https://github.com/HabitRPG/habitica/issues/7449 "Challenge Tags are not automatically associated with Challenge Tasks when creating new challenges"
When I leave and rejoin my challenge, then its tasks are categorised properly. 

However I'm not sure using challenge tags is the best way to do this because players can remove the tag from  one or more challenge tasks or delete the challenge tag, and they're more likely to do that if they like this new categorisation method and don't like having challenge tags messing up their own tag lists. Can the `challenge` object be used to categorise tasks instead of the tags?

Alys:
Sorry, another edit to my bug report. Deleting or unselecting tags doesn't stop this feature from working, which is great, but there's still a problem with the challenge owner not having the tasks correctly categorised when they first create the challenge, and it would be nice to see that fixed before this goes live to stop more bug reports. It can't be the issue I linked to above but it looks like it's similar. Perhaps that issue would be handled by the same fix (also https://github.com/HabitRPG/habitica/issues/7469 for challenge notes).

To summarise: When a challenge owner creates a challenge, they automatically join the challenge and are given the challenge tasks. However the tasks are not categorised in the owner's own task list. The owner has to leave the challenge and rejoin it to make the tasks be categorised.


Challenge tasks are still invisible in the Social > Challenges screen.

TheHollidayInn:
The issue is that I have added a new field to the task's challenge items. Yes, leaving and rejoining will fix the issue, but the true issue is that user's will have to resync with the challenge for it to be categorized. This will basically only work for future challenges. 

Side note: this feature is for an enterprise client, but I figured some users would want it as well - hence this PR.

Alys:
> The issue is that I have added a new field to the task's challenge items. Yes, leaving and rejoining will fix the issue, but the true issue is that user's will have to resync with the challenge for it to be categorized. This will basically only work for future challenges.

Ah, I see! Thanks for the explanation.

We'll need to do a migration to fix all existing challenge tasks before this goes live or immediately after. Otherwise we'll be getting bug reports about it for months to come.

TheHollidayInn:
OKay, let me write a migration for this as well then!

TheHollidayInn:
@alys I am assuming my best option here would be to go through all active challenges and resync each task with the challenge members?

website/server/models/task.js
```diff
@@ -58,6 +58,7 @@ export let TaskSchema = new Schema({
   userId: {type: String, ref: 'User', validate: [validator.isUUID, 'Invalid uuid.']}, // When not set it belongs to a challenge
 
   challenge: {
+    name: {type: String, default: 'None'},
```
Alys: Could this be `shortName` so that it's really clear to future programmers? It would also need changing in website/server/models/challenge.js above.

One benefit would be that when you were grepping the source code for all places where shortName is used, this would show up.

Alys:
> go through all active challenges and resync each task with the challenge members?

Oh, is that possible?! That would be great. It would fix other problems with tasks too, although we'd need to check that it had actually handled all the tasks since we often get timeouts from challenge task syncing (hence the other problems that exist). How is it done?

If that's how we do it, I'd actually like us to do it before this is released as well, so that the other problems get fixed a couple of days ahead of this release, so that we can handle any questions that arise from that before we get new questions about this feature.

website/server/models/task.js
khipkin: `shortName` or `label` or `nickname`

Alys: `shortName` because that's what it's called elsewhere in challenges code.

Alys:
The tasks on a Group Tasks List page are invisible in the same way challenge tasks are.

Alys:
There's also something that's stopping the user's tasks page from updating promptly. For example:

1. click a task's delete button, confirm the deletion
2. the task remains visible as if it's not deleted
3. try to delete it again and you see "Task not found"
4. reload - the task disappears

Another example:
1. create a task - it doesn't appear in the task list
2. reload - the task appears. You can also move to another page and back again to make it appear. Clicking the sync button doesn't make it appear.

If you create a task and then make it visible by moving to another page and back, it's categorised in an "undefined" category. You have to reload before it appears under "None".


When you click the sync button, it will update your stats if they've been changed in another browser, but it doesn't make the blue bar appear at the top of the screen. This makes it look as if it's not working, unless there are visible changes that appear when you sync.

website/client-old/js/directives/task-list.directive.js
```diff
@@ -0,0 +1,78 @@
...
+          if (!$scope.obj) return;
+          $scope.groupedList = {};
+          ['habit', 'daily', 'todo', 'reward'].forEach(function (listType) {
+            groupTasksByChallenge($scope.obj[listType + 's'], listType);
+          });
+        }
+        setUpGroupedList();
+
+        function groupTasksByChallenge (taskList, type) {
+          $scope.groupedList[type] = _.groupBy(taskList, 'challenge.shortName');
```
paglias, line 50: Heads up, we're migrating to lodash v4 (https://github.com/HabitRPG/habitica/pull/8495) which includes many breaking changes (https://github.com/lodash/lodash/wiki/Changelog#v400). Before merging this PR should be checked and updated to make sure all lodash methods are compatible with v4.

website/views/shared/tasks/task_view/index.jade
```diff
@@ -16,13 +16,13 @@
...
       span.glyphicon.glyphicon-minus
-
+      
   // Rewards
   span(ng-if='::task.type=="reward"')
     input.task-input.reward.visuallyhidden(
       type='checkbox',
       ui-keypress='{ 13:"score(task, \'down\')" }' )
-    a.money.btn-buy(ng-class='{highValue: task.value >= 1000}', ng-click='score(task, "down")')
+    a.money.btn-buy(ng-class='{highValue: task.value >= 1000}', ng-click='score(task, "down"); console.log("SDF")')
```
paglias: let's remember to remove this

paglias (changes requested):

Alys:
When you select the option to turn this feature on, the following bugs appear:

On the "Group Tasks List" screen of a guild with a group plan, all the tasks are under "None" headings. There should be no headings.

The first bug in [this comment](https://github.com/HabitRPG/habitica/pull/8469#issuecomment-279210478] happens in the "Group Tasks List" screen (deleted tasks don't disappear until you reload).

A less severe version of that bug appears on the user's own tasks page:
1. click a task's delete button, confirm the deletion
1. the task remains visible as if it's not deleted BUT the title of the task disappears
1. sync - the task disappears (a reload is not necessary)
1. if you don't sync and try to delete it again and you see "Task not found"

[EDIT by Alys: this is fixed (and was not from this PR] There's also a new bug on the user's tasks list page: tasks can't be edited (the pencil icon can be clicked, and the tasks checkbox or plus-minus buttons become invisible, but no edit modal appears). HOWEVER, that bug happens even when you don't turn on the option for this new feature, so it might come from one of the other "pending testing" PRs - I'll be looking into that. Almost certainly related is a bug if you have "Open new tasks in edit mode" selected: when you create a new task the task appears in your list but does not have a checkbox or +/- buttons and the edit modal does not appear. I'll post again here if I find that this is more likely caused by another pending PR (and if anyone else finds out first, that please post here!)

_(Also tasks can't be sorted manually while this feature is on OR off - I'll post more details later after more testing with other pending PRs. This is just a reminder to myself that I need to do that. I suspect this bug isn't from this PR because I'm pretty sure I had tested sorting successfully earlier.)_

Alys:
As mentioned in other PRs, because we're having some trouble on our test server with the PRs that have been merged recently, I've started testing each one individually.  I tested this one on my local install by fetching upstream, then checking out 30954fe7c58966852c054e094abc10f010e0cb0a and doing `npm start`. That includes the three other PRs below, but I've tested them briefly over the past hour or so and found no problems:
- 6d0df7844168916c2a948e521eff84e73b2be9fa Habits v2: adding counter to habit
- 705a78e835ded7643ed6655d61d9c1fe3d555f4a Debounce `$scope` updates when typing in chat
- 44f23a76751f1fcf5860db9f0ab7ec7dfa8d3a28 fix(nextRewardUnlocksIn): Check-in prize message was always plural (_the only thing I tested here was task sorting, which works, but the PR is just a text change so it can't have any significant bugs_)

Testing locally on 30954fe7c58966852c054e094abc10f010e0cb0a shows:
- The delete bug described above comes from this PR  (i.e., you have to sync before a deleted task disappears). The behaviour where the task title becomes invisible after "deletion" does not happen with this PR, but that's a minor variation of the delete bug and can be ignored.
- Tasks CAN be edited - the edit modal does appear. The edit bug described above does NOT come from this PR.
- When the "Open new tasks in edit mode" option is selected, tasks CAN be added successfully and the edit modal opens correctly.
- Tasks can NOT be sorted manually. That bug comes from this PR (I forgot to test task sorting in 6d0df7844168916c2a948e521eff84e73b2be9fa and 705a78e835ded7643ed6655d61d9c1fe3d555f4a but I've just now gone back to 44f23a76751f1fcf5860db9f0ab7ec7dfa8d3a28, which includes those two, and sorting works correctly there).

Details of the sorting bug:
- When you try to drag one task, your mouse picks up the entire column of tasks and tries to move them all together. 
- As you're doing that, the task text and icons go all freaky and misaligned and squished. That is TOTALLY not a good bug report :) but it's hard to describe the behaviour and even harder to screenshot it since you have to be moving the mouse to see it. It's easy to replicate though for anyone with access to our test server.
- The sorting problem happens regardless of whether the "Group tasks by challenge title" setting for this new feature is turned on or off.

Issue #8351, opened 2017-01-01
Greetings cards have a confusing notification when they are read by the recipient


Alys:
When you read a Greeting Card, Birthday Card, Thank-You Card, or New Year's Card, you see a notification saying that it's been read but it uses the key for the notification rather than its user-friendly name. For example, "birthday has been read" or "nye has been read" (screenshot below).

~It should read "Birthday Card has been read" (and similar text for the other cards). Translatable of course.~

EDIT: The "... has been read" notification should not be shown.

![nye](http://savepic.net/8767290.png)

vincecampanale:
Hi @Alys, I'll take this one as well!

Alys:
@vincecampanale Thank you!

SabreCat:
Another possibility: don't show the notification at all. It's not useful information to the user. They already know they read a card!

Alys:
@SabreCat I would totally prefer that! I've been assuming that the less useful notifications we see had to be there because the API enforced it. I'll start making low priority issues to hide them when I find them.

SabreCat:
@Alys They are provided by the API, but the client can choose not to display them with no ill effect. (We still want to show any _error_ that occurs, however.)

vincecampanale:
Is the cardType variable required to be one word/lowercase? 

Unrelated, where would be a good place to add a function that translates cardType into a readable name: (e.g. 'nye' => New Year's Eve) before displaying it to the user? I'm thinking about adding a function like this `if(cardType === 'nye'){ displayCardType = "New Year's Eve" }` and so on for each cardType. Let me know what you think.

Thanks!

Alys:
@vincecampanale Ah, I'm sorry, we should have updated the top post. The desired change now is that we don't show a notification to tell the user that they've read the card. See the above comments for details. I'll update the top post now.

joedragons:
@Alys @SabreCat Tangential but in case deserving of a note/issue per the discussion above, I find the the fact healers casting protective aura generating a notification weird.  For me (as a healer currently) it's an additional thing I need to click to clear after casting the spell.  In other words, I know I cast protective aura and it lists it in the party things done, so getting a notification and clicking to clear seem unnecessary to me.  Seems similar sentiment to this.
(to be clear, I am not suggesting a dual fix in this issue but there seemed to be discussion above about making additional notes/issues if there could be further cleanup based around facts discovered here)

Alys:
[...]

PR #8473, opened 2017-02-03
Preventing cardRead from notifying user.

vincecampanale:

Fixes Issue #8351 

### Changes
Added a condition to line 159 in userServices.js that checks if the op is 'cardRead.' If it is, then the user will not receive a notification saying that they read the card. 

----
UUID: 08c2b5db-70bd-4677-8a95-d84cdfec2afb

### DIFF
--- a/website/client-old/js/services/userServices.js
+++ b/website/client-old/js/services/userServices.js
@@ -158,7 +158,8 @@ angular.module('habitrpg')
 
         var clientMessage = clientResponse[1];
 
-        if (clientMessage && opName !== "readCard") {
+        const opNamesWithoutNotification = ["readCard"];
+        if (clientMessage && opNamesWithoutNotification.indexOf(opName) === -1) {
           Notification.text(clientMessage);
         }
 
--- a/website/client-old/js/services/userServices.js
+++ b/website/client-old/js/services/userServices.js
@@ -158,7 +158,7 @@ angular.module('habitrpg')
 
         var clientMessage = clientResponse[1];
 
-        const opNamesWithoutNotification = ["readCard"];
+        var opNamesWithoutNotification = ["readCard"];
         if (clientMessage && opNamesWithoutNotification.indexOf(opName) === -1) {
           Notification.text(clientMessage);
         }
--- a/website/client-old/js/services/userServices.js
+++ b/website/client-old/js/services/userServices.js
@@ -158,7 +158,7 @@ angular.module('habitrpg')
 
         var clientMessage = clientResponse[1];
 
-        if (clientMessage) {
+        if (clientMessage && opName !== "readCard") {
           Notification.text(clientMessage);
         }
 


SabreCat:
This would do the trick, though it's a peculiar bit of special-case hard coding. Maybe acceptable pending the frontend rewrite, even so. Thoughts @TheHollidayInn @paglias ?

khipkin:
It would probably be better to just never create the notification to begin with, rather than selectively ignoring it after it's created.

Alys:
When using Habitica through the API, the notification might be useful for feedback.

vincecampanale:
So what's the verdict? Will this work for now, pending the front end rewrite?

TheHollidayInn:
1. Where does this notification originate from? API? 

2. If we go with this fix (which I am fine with), let's do two things:
 - Moved the string "readCard" to a constant.
 - Add a test for this condition (I know it is mundane but it will ensure consistency in the redesign)

Alys:
I suggest that the constant should be an array or object listing the types of notifications that we want to suppress. At the moment there'll be just one type in it but there might be others in future and an array / object will be easier to update.

vincecampanale:
@TheHollidayInn 
1. Unsure where the notification originates from.
2. Put "readCard" in an array which contains opNames we want to suppress notifications for (per @Alys recommendation). 
If the `opName` is not in the `opNamesWithoutNotifaction` array, it will get a notification.

vincecampanale:
Any further input/suggestions?

SabreCat:
We could still add a test per @TheHollidayInn's second bullet point. We'd need to stub `$http`, spy on `Notification`, and check that `Notification.text` doesn't get called when you run `readCard`. Does that sound about right THI?

TheHollidayInn:
Yea, just stub and spy, then call `callOpsFunctionAndRequest ` with the `"readCard"` in the param then ensure `Notification.text` is not called.

vincecampanale:
@SabreCat @TheHollidayInn 
Hey guys, working on the test now. What exactly do you mean by "stub and spy"? Is there an example in the tests I can look at? Also will this test go in POST-user_read_card.test.js or /test/common/ops/readCard.js? 
Thanks for the help and feedback - I'm learning a lot.

vincecampanale:
Getting a couple of errors when trying to test. 
`npm test` outputs: 
npm test output

`npm start` outputs:
npm start output

Alys:
@vincecampanale this might not help but as a first step, merge/rebase the latest `develop`, check that your config.json is up to date with config.json.example, and run `npm install` to ensure that you have the latest libraries (check the output for errors). Post the test errors again after that if they're still happening and if they're not identical to before.

TheHollidayInn:
The slack url error can be ignored. The first error just needs to be installed via npm. I would just run it manually.

vincecampanale:
How would I run it manually? I tried doing `npm install -g eslint` and `npm install -g esnext` and a bunch of other things, but got the same error.

Alys:
@vincecampanale You wouldn't use the -g flag for installing it in a specific project, like the habitica site, but really, I'd still do `npm install`. If you're missing that one, you might be missing others. The Aspiring Blacksmiths guild lists changes that are made to package.json and advises when npm install should be done again, so if you haven't noticed a few of them, definitely do the full `npm install`.

TheHollidayInn:
`npm install eslint-config-habitrpg/esnext`

vincecampanale:
@Alys I did try `npm install` a couple of times, and did some troubleshooting on my own. I'm especially confused because I've had this working before.

@TheHollidayInn That command is returning this error: 

screen shot 2017-02-10 at 2 19 05 pm

Sorry for being a hassle >.<

Alys:
No worries. :) The setup is complex until you're used to it.

There's no sign of a "habitica" directory in the output. You need to run all install commands in the top level habitica directory (the one that contains package.json and config.json).  Move into that directory and run `npm install`. Examine the output carefully for errors and if there is anything that you're not at all sure of, show us the command and full output. The best way to do that is by pasting a link here to a GitHub Gist or similar (see [Setting up Habitica Locally >  Prepare for Troubleshooting](http://habitica.wikia.com/wiki/Setting_up_Habitica_Locally#Prepare_for_Troubleshooting) for details).

After you've done `npm install`, if the tests still fail, do `ls node_modules/eslint-config-habitrpg/` (still in the habitica directory). It should contain a few files. Show us that command and its output, as well as the full output of `npm install` and any other troubleshooting data that you think might be relevant..

TheHollidayInn:
Approved!

TheHollidayInn:
Noted towards your Tier 2 progress :D

PR #8479, opened 2017-02-07
Send gem-purchased item data to Amplitude

### LINKED ISSUE #8365: purchases of backgrounds are not shown in Amplitude
I can't find any records in Amplitude of my account purchasing backgrounds. We need this information to help troubleshoot bugs.

I tested on the production website by buying first a single background then a set. UUID 3e595299-3d8a-4a10-bfe0-88f555e4aa0c

cailinglu:
### Changes
- In the analytics services files, sends the user's animal ears, backgrounds, shirts, skins, and hairs to Amplitude
- Added tests for the above
- Added two filter functions: One removes the plain backgrounds and only sends gem-purchased backgrounds to Amplitude. The other filters out animal ears from `items.gear.owned` to send to Amplitude.
- Added a function to format hair purchases, see below for an example.
- Gem-purchased items are sent to Amplitude in arrays, so something like this:
```javascript
{
  // ...
  animalEars: [ 'headAccessory_special_foxEars' ],
  background: [ 'blizzard' ],
  // hair is sent as an object of arrays:
  hair: {
    base: [ '1' ],
    beard: [ '3' ],
    color: [ 'snowy' ],
    mustache: [ '2' ],
  },
  shirt: [ 'horizon' ],
  skin: [ 'snowy' ]
  // ...
}
```

EDIT (2017-04-29): The filter function that removes free backgrounds from being sent to Amplitude now grabs them from the common backgrounds.js file.

----

### DIFF
--- a/test/api/v3/unit/libs/analyticsService.test.js
+++ b/test/api/v3/unit/libs/analyticsService.test.js
@@ -279,6 +279,29 @@ describe('analyticsService', () => {
           rewards: [{_id: 'reward'}],
           balance: 12,
           loginIncentives: 1,
+          background: {
+            blizzard: true,
+          },
+          shirt: {
+            horizon: true,
+          },
+          hair: {
+            base: {
+              1: true,
+            },
+            beard: {
+              3: true,
+            },
+            color: {
+              snowy: true,
+            },
+            mustache: {
+              2: true,
+            },
+          },
+          skin: {
+            snowy: true,
+          },
         };
 
         data.user = user;
@@ -305,6 +328,29 @@ describe('analyticsService', () => {
                 balance: 12,
                 balanceGemAmount: 48,
                 loginIncentives: 1,
+                background: {
+                  blizzard: true,
+                },
+                shirt: {
+                  horizon: true,
+                },
+                hair: {
+                  base: {
+                    1: true,
+                  },
+                  beard: {
+                    3: true,
+                  },
+                  color: {
+                    snowy: true,
+                  },
+                  mustache: {
+                    2: true,
+                  },
+                },
+                skin: {
+                  snowy: true,
+                },
               },
             });
           });
@@ -523,6 +569,29 @@ describe('analyticsService', () => {
           dailys: [{_id: 'daily'}],
           todos: [{_id: 'todo'}],
           rewards: [{_id: 'reward'}],
+          background: {
+            blizzard: true,
+          },
+          shirt: {
+            horizon: true,
+          },
+          hair: {
+            base: {
+              1: true,
+            },
+            beard: {
+              3: true,
+            },
+            color: {
+              snowy: true,
+            },
+            mustache: {
+              2: true,
+            },
+          },
+          skin: {
+            snowy: true,
+          },
         };
 
         data.user = user;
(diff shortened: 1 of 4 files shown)


cailinglu:
Added some checks to make sure that the user has backgrounds, shirts, etc. before trying to format them for Amplitude.

website/client-old/js/services/analyticsServices.js
```diff
@@ -98,6 +98,38 @@
...
+          animalEars.push(item);
+        }
+        return animalEars;
+      }, []);
+    }
+
+    // filter out gem-purchased backgrounds from backgrounds
+    function _filterBackgrounds (owned) {
+      // current free backgrounds
+      var free = ['blue', 'green', 'purple', 'red', 'yellow'];
```
Alys: The free backgrounds now include `violet` too.

Is it possible to write this code so that it identifies the free backgrounds as being all those in `backgrounds.incentiveBackgrounds` from website/common/script/content/appearance/backgrounds.js ?

Alys: Although since the violet one is given when you join the site, it can't ever be purchased, so it doesn't matter that the code won't prevent it being reported in Amplitude. It also won't be a disaster if other free backgrounds are added in future and we forget to update this code. Changing it to identify free backgrounds automatically might be more trouble than it's worth.

cailinglu: Sounds like a good plan for the long-term though, I'll make the edit.

cailinglu:
Not sure if commits send messages to the mods, so: Ready for review! :)

test/api/v3/unit/libs/analyticsService.test.js
```diff
@@ -279,6 +279,38 @@ describe('analyticsService', () => {
...
+            beard: {
+              3: true,
+            },
+            color: {
+              snowy: true,
+            },
+            mustache: {
+              2: true,
+            },
+          },
```
SabreCat: These are set up on the wrong level of the user object. They should be under `user.purchased`

SabreCat (changes requested):
This is on the right track, but I don't think it will work in production. The data you're trying to send is in the user's `purchased` property, not on the root of the user object. See https://github.com/HabitRPG/habitica/blob/develop/website/server/models/user/schema.js#L141-L161

SabreCat:
Actually, hold off on this. I think the original issue ticket may be in error and there isn't actually any change needed here. :\ If that turns out to be the case, I'm sorry for the wasted effort...

SabreCat:
OK, very sorry. See my comments on the original issue: https://github.com/HabitRPG/habitica/issues/8365#issuecomment-300600450. Adding more user property data isn't really what we're looking for re the linked issue. It could be useful for people who have Amplitude access but not database access, so if you'd like to clean up this PR and continue with it, it can still be incorporated! But it isn't a fix for #8365.

cailinglu:
Made the changes to the tests to put background, hair, skin, shirt on `user.purchased`, if this PR is of any use. I'm ok if it's not!

Currently taking a look at the unlock functions and why the customizable gem purchases aren't going through it as assumed.

SabreCat:
The properties are going through, but without any data in the arrays. This for a user with several of each of the items in question (except maybe backgrounds):

image

Are you able to get data for these properties in your own environment?

SabreCat:
Closing this for now; if you'd like to keep trying to get it to work, @caillu, feel free to reopen it, but your efforts may be better spent elsewhere!

PR #8481, opened 2017-02-07
Option to disable login incentive modal fixes #8478

### LINKED ISSUE #8478: Check-in Incentives popup should have an opt-out option

### Description

Most repetitious pop-ups (like Level Up and Streak Achievement modals) have an option in site settings, mirrored on the popup itself, to disable them. This is not the case for Check-in Incentive notifications, however, which has proven to be a nuisance to users who aren't particularly interested in or motivated by those incentives. We should add another entry to the list of "Don't show popup" checkboxes at /#/options/settings/settings to disable the incentives modal.

ghost:

#8478 

### Changes
* Checkbox to toggle preference in settings/settings
* Checkbox to toggle preference in the standard modal
* Checkbox to toggle preference in the reward unlocked modal
* Suppression preference added to schema
* Test for suppression of modal before displaying it

----
UUID: d59a4ef9-21dc-438e-a52e-c8dfd55822bb

### DIFF
--- a/website/common/locales/en/settings.json
+++ b/website/common/locales/en/settings.json
@@ -16,6 +16,7 @@
     "startAdvCollapsed": "Advanced Options in tasks start collapsed",
     "startAdvCollapsedPop": "With this option set, Advanced Options will be hidden when you first open a task for editing.",
     "dontShowAgain": "Don't show this again",
+    "suppressLoginIncentiveModal": "Don't show popup when checking in",
     "suppressLevelUpModal": "Don't show popup when gaining a level",
     "suppressHatchPetModal": "Don't show popup when hatching a pet",
     "suppressRaisePetModal": "Don't show popup when raising a pet into a mount",
--- a/website/client-old/js/controllers/notificationCtrl.js
+++ b/website/client-old/js/controllers/notificationCtrl.js
@@ -147,7 +147,9 @@ habitrpg.controller('NotificationCtrl',
             scoreTaskNotification = notification;
             break;
           case 'LOGIN_INCENTIVE':
-            Notification.showLoginIncentive(User.user, notification.data, Social.loadWidgets);
+            if (!User.user.preferences.suppressModals.loginIncentive) {
+              Notification.showLoginIncentive(User.user, notification.data, Social.loadWidgets);
+            }
             break;
           default:
             if (notification.data.headerText && notification.data.bodyText) {
--- a/website/server/models/user/schema.js
+++ b/website/server/models/user/schema.js
@@ -460,6 +460,7 @@ let schema = new Schema({
       invitedQuest: {type: Boolean, default: true},
     },
     suppressModals: {
+      loginIncentive: {type: Boolean, default: false},
       levelUp: {type: Boolean, default: false},
       hatchPet: {type: Boolean, default: false},
       raisePet: {type: Boolean, default: false},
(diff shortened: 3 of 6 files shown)


TheHollidayInn:
Looks good. Can you add a test for this?

ghost:
I could use some help with testing what would trigger this modal. Could you or someone point me in the right direction?

website/common/locales/en/settings.json
```diff
@@ -16,6 +16,7 @@
     "startAdvCollapsed": "Advanced Options in tasks start collapsed",
     "startAdvCollapsedPop": "With this option set, Advanced Options will be hidden when you first open a task for editing.",
     "dontShowAgain": "Don't show this again",
+    "suppressLoginIncentiveModal": "Don't show popup when checking in",
```
khipkin, line 19: Suggestion: Have @lemoness approve this phrasing.

TheHollidayInn:
This example should help: https://github.com/HabitRPG/habitica/blob/develop/test/client-old/spec/controllers/notificationCtrlSpec.js#L119

website/common/locales/en/settings.json
ghost: I do imagine "checking in" to not mean anything to a user.

Alys: Maybe 'Don't show "Check-In Counter" popup each morning' ?

lemoness: Yes, I prefer "Don't show "Check-In Counter" popup each morning" because it is more specific. Does this hide all of them, or only the ones between prizes?

ghost: All of them. I can certainly differentiate between the two if necessary.

Alys: @kevinsearle That's a good idea! I think some would like to be able to see the ones where you've earned an item but hide the others.

ghost: Should the reward text read 'Don't show "Check-In prize" popup' or similar?

Alys: Yes, I guess something like this:
- **Don't show "Check-In Prize" popup**
- **Don't show "Check-In Counter" popup**

They'd probably benefit from an on-hover effect to give more explanations about what each are, as we do for some of the other settings (see underlines in the screenshot below). Perhaps:

- **Each day when you use Habitica, your "Check-In" counter is increased by one. At certain numbers, you receive [prizes](http://habitica.wikia.com/wiki/Daily_Check-In_Incentives)! This setting controls whether you receive a notification each time you get a prize.**
- **Each day when you use Habitica, your "Check-In" counter is increased by one. At certain numbers, you receive [prizes](http://habitica.wikia.com/wiki/Daily_Check-In_Incentives)! This setting controls whether you receive a notification each time your counter has increased but you haven't got a prize yet.**

Both of those could probably do with improvement though.


However.... 

I'm wondering if we can reverse the way this works and make the messages say "Show .... popup" with the checkbox ticked by default. That feels more natural. Below is a screenshot showing the current settings as they are for a new account. Making this "Show..." would be inconsistent with the other items at the bottom of the list, but consistent with the "Show Header" option at the top. The inconsistency is undesirable but we could move towards doing it for the other settings too. I feel it's a nicer way of presenting it. @lemoness what do you think? The wording would then be
- **Show "Check-In Prize" popup**
- **Show "Check-In Counter" popup**
with the on-hover text being the same as above. If we put these two new settings directly under "Set Dailies default to 'due' tab" then the underlining would be consistent with the items above it, which might help to reduce the impact of the "Show / Don't show" inconsistency.

image

khipkin: 👍 👍 to using "show" instead of "don't show".

ghost:
I had to take some time away from this due to other obligations but jumping back on it.

ghost:
My fork got messed up, so had to recreate it.

Alys:
@kevinsearle Have you had a chance to do more on this? Do you need any advice about fixing any commit problems?

Or if you've run out of time for this, would you prefer someone else to take it over? If so, just let us know - we completely understand about real life taking precedence!

ghost:
@Alys to be honest, the issue's scope creeped so I put it off. It seemed like some of the discussion could have been broken out into a different issue altogether. At this point, I don't really have the time unfortunately so it might be best for someone else taking over.

Alys:
@kevinsearle No worries, I'll open it up for someone else to work on. You're always welcome to tell us if you can no longer work on an issue!

PR #8483, opened 2017-02-07
[WIP] Added undo directive and undo for task scoring

TheHollidayInn:

Fixes put_issue_url_here

### Changes

----
UUID:


website/client-old/js/services/undoService.js
```diff
@@ -0,0 +1,50 @@
...
+
+    undoApi.undoAction = function () {
+      var pop = undoApi.history.pop();
+      if (!pop) return;
+
+      // @TODO: We need to add a pop interface for each action
+      if (pop.action === 'scoreTask') {
+        var direction = pop.data.direction;
+
+        if (direction === "up")  {
```
paglias, line 22: the undo for task scoring works by scoring in the opposite direction? If yes I don't think it's the right approach:

1. It'll pollute task.history
2. It won't revert drops or critical hits, ...

paglias (changes requested):

website/client-old/js/services/undoService.js
TheHollidayInn: Is this not how we uncheck a todo?

paglias: yes, but I think we would get more questions if the undo didn't work exactly as an undo. But at the same time the only other option I see is not sending the request for 10 seconds and in that timeframe it's undoable, which would have the problem of not showing drops for 10 seconds so not ideal

Alys: Another option: Each time a task is scored, record the change in value of the task and the things gained from it (stats, quest progress, drops). The record would be in an object in the user's document, not in the task's document. The record would include the task's ID. When the undo button is clicked, use that data to reverse the effects on the user's account and on the task (subtract the change in task value and also remove the most recent history entry when appropriate).

It could also be an array of objects if we wanted multi-undo. The array could have a "type" field so you could record skills, purchases, etc, not just task clicks. But really, I think a single undo would be enough.

Erase the object on cron so that we don't try to revert quest progress after it's already been applied to the quest.

Alys: If you did decide to do this by storing the most recent change in an object, and if you wanted to allow just one level of undo, I think it would still be best to make the data structure an array containing the object, so that in future we could more easily expand it if we decided to allow multi-level undos.

That would also allow us to use the same data structure to provide a visible history of the user's recent actions (at some time in the future; not now). That's a feature request that we see fairly often.

To prevent the array becoming too large, it could be emptied on cron and/or when it had more than X objects in it then the oldest one could be deleted as the newest was added. 

@TheHollidayInn Would it be possible to add a few words to the PR's description to describe the extent of this feature? E.g., one level of undo or multiple levels; only tasks or tasks plus skills plus ...?  That will help Socialites and wiki editors be prepared for answering questions and writing documentation when it goes live, and it might help us in the future when we're looking back on changes.

website/server/models/task.js
```diff
@@ -243,7 +243,7 @@ export let TodoSchema = new Schema(_.defaults({
   dateCompleted: Date,
   // TODO we're getting parse errors, people have stored as "today" and "3/13". Need to run a migration & put this back to type: Date see http://stackoverflow.com/questions/1353684/detecting-an-invalid-date-date-instance-in-javascript
   date: String, // due date for todos
-}, dailyTodoSchema()), subDiscriminatorOptions);
+}, dailyTodoSchema(), habitDailySchema()), subDiscriminatorOptions);
```
paglias, line 246: this will add all todos fields to habits, is what you wanted?

website/server/controllers/api-v3/tasks.js
```diff
@@ -1215,4 +1242,52 @@ api.deleteTask = {
...
+ *
+ * @apiUse TaskNotFound
+ */
+api.undoTask = {
+  method: 'POST',
+  url: '/tasks/:taskId/undo',
+  middlewares: [authWithHeaders()],
+  async handler (req, res) {
+    let user = res.locals.user;
+    let taskId = req.params.taskId;
```
paglias, line 1265: validation

website/server/controllers/api-v3/tasks.js
```diff
@@ -1215,4 +1242,52 @@ api.deleteTask = {
...
+    if (!task) {
+      throw new NotFound(res.t('taskNotFound'));
+    }
+
+    let latestHistory = task.history[task.history.length - 1];
+    user.stats.hp -= latestHistory.stats.hp;
+    user.stats.mp -= latestHistory.stats.mp;
+    user.stats.exp -= latestHistory.stats.exp;
+    user.stats.gp -= latestHistory.stats.gp;
+    if (latestHistory.quest && latestHistory.quest.progressDelta) user.party.quest.progress.up -= latestHistory.quest.progressDelta;
```
paglias, line 1277: what if the quest ended in the meantime?

website/server/controllers/api-v3/tasks.js
```diff
@@ -1215,4 +1242,52 @@ api.deleteTask = {
...
+
+    let drop = latestHistory.drop;
+    if (drop && latestHistory.direct === 'up') {
+      let dropType = drop.type.toLowerCase();
+      user.items[dropType][drop.key] -= 1;
```
paglias, line 1282: what if the items has been used to hatch a pet in the meantime?

website/server/controllers/api-v3/tasks.js
```diff
@@ -1215,4 +1242,52 @@ api.deleteTask = {
...
+    }
+
+    await user.save();
```
paglias, line 1285: shouldn't the undone task be modified to? at least regarding completeness, task.value

website/server/controllers/api-v3/tasks.js
```diff
@@ -542,10 +542,37 @@ api.scoreTask = {
...
+      stats: {},
+    };
+    userUpdates.stats.hp = newStats.hp - previousStats.hp;
+    userUpdates.stats.mp = newStats.mp - previousStats.mp;
+    userUpdates.stats.exp = newStats.exp - previousStats.exp;
+    userUpdates.stats.gp = newStats.gp - previousStats.gp;
+    if (user._tmp.drop) userUpdates.drop = user._tmp.drop;
+    if (user._tmp.quest) userUpdates.quest = user._tmp.quest;
+
+    if (!task.history || task.history.length === 0) {
```
paglias, line 563: task.history should always exist

website/server/controllers/api-v3/tasks.js
```diff
@@ -542,10 +542,37 @@ api.scoreTask = {
 
     let wasCompleted = task.completed;
 
+    let previousStats = _.clone(user.stats);
```
paglias, line 545: clone doesn't work well on mongoose objects (especially in v4), it's better to call _.clone(user.stats.toObject())

paglias (changes requested):

website/server/models/task.js
TheHollidayInn: Ah, no. This seems to had the habitDaily fields todos. I thought.

website/server/controllers/api-v3/tasks.js
TheHollidayInn: Ah, right, thanks!

website/server/controllers/api-v3/tasks.js
TheHollidayInn: This was lingering from when todos didn't have history. I'll remove it.

website/server/controllers/api-v3/tasks.js
TheHollidayInn: Hmm. There is much to consider if we keeps these undos for a long time. I think if now we reset undos every time the user refreshes we can prevent worrying about this. 

Or, I can check if the user quest still exists.

website/server/controllers/api-v3/tasks.js
TheHollidayInn: Same as above with the breif undo stack, or I can simply add undoing sells, hatching to the undo stack.

paglias: hmm I fear handling every possible outcome will be difficult, maybe this is the least bad solution. What if we allow undoing for X seconds only like Gmail where you can undo an email send for 10 seconds?

website/server/controllers/api-v3/tasks.js
```diff
@@ -1215,4 +1242,52 @@ api.deleteTask = {
...
+    let drop = latestHistory.drop;
+    if (drop && latestHistory.direct === 'up') {
+      let dropType = drop.type.toLowerCase();
+      user.items[dropType][drop.key] -= 1;
+    }
+
+    await user.save();
+
+    res.respond(200, {
+      user,
```
paglias, line 1288: probably it's enough to return user.stats and items

website/server/models/task.js
paglias: sorry, I got confused by the name. It'll add the fields in `habitDailySchema` but right now is only history. So it's fine to add them but the name of the object should be renamed or it's confusing

website/server/controllers/api-v3/tasks.js
TheHollidayInn: Yea, that is where I was going above to only allow undoing until a refresh. But we can add a timer too.

paglias: I was thinking of not executing the action at all for X seconds on the server but the problem is that you would not be able to get drops immediately since they're only server side and we cannot simulate them on the client

Alys: We could just not remove drops when we undo a task. That's no worse than the current situation of manually unticking a task, which lets you keep the drop (and we want to keep that behaviour for manually unticking because it's used as a Wishing Well feature). It's not completely ideal to keep a drop when undoing, but I think that's better for UI than introducing a delay, especially since undoing won't be that common, and undoing after getting  a drop will be less common still.

website/server/controllers/api-v3/tasks.js
```diff
@@ -1215,4 +1242,52 @@ api.deleteTask = {
...
+    let latestHistory = task.history[task.history.length - 1];
+    user.stats.hp -= latestHistory.stats.hp;
+    user.stats.mp -= latestHistory.stats.mp;
+    user.stats.exp -= latestHistory.stats.exp;
+    user.stats.gp -= latestHistory.stats.gp;
+    if (latestHistory.quest && latestHistory.quest.progressDelta) user.party.quest.progress.up -= latestHistory.quest.progressDelta;
+
+    let drop = latestHistory.drop;
+    if (drop && latestHistory.direct === 'up') {
+      let dropType = drop.type.toLowerCase();
```
Alys, line 1281: I don't think toLowerCase should be here. For example, user.items includes `hatchingPotions` (capital P).

website/server/controllers/api-v3/tasks.js
TheHollidayInn: I think the best case is to just keep with refreshing the undo stack after reload. Then, simply only remove the drop if it is in the user's inventory. That way, if they sell or use the drop before the undo - then they keep the drop. Which seems to be okay with us anyway.

SabreCat:
Recent discussion has led us in a different direction for this. Closing the PR for now; we can reopen it easily enough when we resume work here!

PR #8198, opened 2016-11-08
Habits v2: adding counter to habits (cleaned up branch) - fixes #8113

shalott:
Fixes #8113 
### Changes

The original PR for this was here:
https://github.com/HabitRPG/habitica/pull/8175

Unfortunately while fixing a conflict in tasks.json, I messed up the rebase and wound up pulling in too many commits and making a giant mess, and I can't face the wrestling with git so am making a clean branch, sorry. :P

The only actual change here is to the test in cron.test.js, where it turns out that moment() and sinon.useFakeTimers() do not play well together. I should note that this means that the other tests that use the two together (eg, preening.test.js) are also not actually working as expected -- when using moment, it seems that sinon.useFakeTimers() is setting the time to the start of the UNIX epoch regardless of the day specified with moment. The fix for this is to use Date.new() instead. 

ORIGINAL BRANCH:

First pass at adding a counter to habits per discussion on the Trello board and in issue #8113. 

- For discussion: how to display the +/- counter. Currently, the counter on onlies is unmarked, and the counter on +/- habits is displayed as `+3|-2` I considered other options which are screenshotted in the issue description but this one looks like the best to me so far. 

- For discussion: I went with Monday as the start of the week per ISO standard and also because it seems more useful for separating weekend/weekday tasks. I could alternatively add a user pref for start of the week, however it occurs to me that this might be confusing to people who would expect it to be relevant in Dailies? 

Here's how the new feature works:

Habits have counterUp/counterDown and a frequency that is either daily, weekly, or monthly. The default is per day. The frequency can be set in the advanced options of habits. 

In tasks view, the user sees a counter on the habit box (to the left of the "top" icon, same place the streak counter goes on a daily) showing the number of times the habit has been ticked within the period.

If it is a +/- habit then it shows +#positive|-# negative per suggestion from @Alys. 

screenshot 2016-10-27 15 40 03

screenshot 2016-10-27 18 30 36

In the cron after the end of the counter period, the counter resets to 0.

---

UUID: 21a3965c-9203-4f83-b819-6b63902e03e2

PR #8485, opened 2017-02-09
Debounce `$scope` updates when typing in chat.

### LINKED ISSUE #6462: Tavern / Guild / Party chat lag / slowness

applmak:

Fixes #6462.

### Changes
Debounce `$scope` updates when typing in chat. Saves a bunch of time per frame when the user is typing, making the chat widget feel more responsive.

Profile before:

Profile after:

----
UUID: 
912546cc-28cf-4c3a-b01e-c8dccfd6cbd9


paglias:
@applmak I have a question about this, what happens if a) debounce runs and save the value in the model variable b) user types more chars c) user press enters 

will the chars entered between the end of a) and the start of c) be lost?

applmak:
@paglias According to the [docs on NgModelOptions](https://docs.angularjs.org/api/ng/directive/ngModelOptions#triggering-and-debouncing-model-updates), a `submit` event will force the model to sync with the view. This means that the action attached to the form via `ng-submit` will work, but as currently implemented the `keyDownListener` will not. I'll update this PR to include a test for this case, while changing the `keyDownListener` to submit the form, which should address your concern.

applmak:
@paglias Updated with test that shows that data is not lost when requesting the form to submit. I introduced a new directive submit-on-meta-enter to encapsulate this behavior. PTAL.

TheHollidayInn:
This is pretty awesome. The profile report looks great.

paglias:
this look good to me, @TheHollidayInn feel free to merge if it's good for you

TheHollidayInn:
The integration test can be ignored, but there are failing builds on the common and karma.

applmak:
Those errors seem to be related to the test environment, rather than the test or code itself. Is there some way to punch the build to get it to retry?

TheHollidayInn:
Restarted!

applmak:
Strict mode failure reported in the karma test. Fixed, presubmit checks are running.

applmak:
Okay, tests are green. Is there any manual QA that happens before this goes live?

TheHollidayInn:
We will add this to our release branch. After some testing it will go live! Thanks. :D

applmak:
No problem. I know you're re-writing your frontend, but if there are other quick performance improvements I could do, lmk.

TheHollidayInn:
We are actually hoping to improve the test coverage of the current front end if you would like to help.

TheHollidayInn:
Noted towards your tier! Congrats on Tier 1 :D

Alys:
Just FYI for the record:

We're having some trouble on our test server with the PRs that have been merged in recently, so I've started testing each one individually. I tested this one on my local install by fetching upstream, then checking out 705a78e835ded7643ed6655d61d9c1fe3d555f4a and doing `npm start` My testing has been VERY brief but everything looks good here so far. My tests show:

- Chat messages can be posted in the Tavern
- People's names can be selected in Tavern chats by typing @ then selecting from the list
- Habits can be scored up and down (with the new counters incrementing correctly https://github.com/HabitRPG/habitica/pull/8198)
- Tasks can be edited
- Tasks can be deleted
- Tasks can be added, and if the "edit mode" setting is turned on, the edit modal appears as soon as you add the task

PR #8489, opened 2017-02-14
Automatically add a 'two handed' suffix (#8441)

### LINKED ISSUE #8441: two-handed weapons/shields aren't easily identifiable as such
Most two-handed weapons aren't easily identified as such because they don't have anything in their text to state that they are.

We sometimes see bug reports from mobile app users when equipping a two-handed item unequips a shield. There are issues to give notifications about that when equipping them on the apps (https://github.com/HabitRPG/habitica-ios/issues/433 and https://github.com/HabitRPG/habitica-android/issues/688) but we should also include a message in each two-handed item's description on website and apps - i.e., the message should be in the data returned from the API's `content` route.

For consistency, the message should be automatically added to the end of the item's description when the item has a true value for the `twoHanded` attribute (e.g., the rancherLasso below). I.e., the PR for this change will NOT change any of the `common/locales/en/` json files (with one exception listed below) but instead will modify the API's code to insert the message.

The message should be something like "**Two-handed item.**" We won't use "Two-handed weapon" because future items might have more of a shield-like feel to them.

The PR should change `weaponSpecialCandycaneNotes` in `common/locales/en/gear.json` to remove the hard-coded two-handed description. I.e., change this:
`A powerful mage's staff. Powerfully DELICIOUS, we mean! Two-handed weapon. Increases Intelligence by <%= int %> and Perception by <%= per %>. Limited Edition 2013-2014 Winter Gear.`
to this:
`A powerful mage's staff. Powerfully DELICIOUS, we mean! Increases Intelligence by <%= int %> and Perception by <%= per %>. Limited Edition 2013-2014 Winter Gear.`

Example of a two-handed item:
```
  rancherLasso: {
    twoHanded: true,
    text: t('weaponArmoireRancherLassoText'),
    notes: t('weaponArmoireRancherLassoNotes', { str: 5, per: 5, int: 5 }),
    value: 100,
    str: 5,
    per: 5,
    int: 5,
    set: 'rancher',
    canOwn: ownsItem('weapon_armoire_rancherLasso'),
  },
```

AndrzejDoyle:
…ded weapons (similar to the class bonus suffix).

Fixes https://github.com/HabitRPG/habitica/issues/8441

### Changes

Items with the `twoHanded` property will now automatically have the suffix "Two-handed item" appended to their description on the Equipment page.  

----
UUID:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ test/client-old/spec/mocks/translations.js
 .elasticbeanstalk/*
 !.elasticbeanstalk/*.cfg.yml
 !.elasticbeanstalk/*.global.yml
+/nbproject/private/
--- a/website/common/locales/en/gear.json
+++ b/website/common/locales/en/gear.json
@@ -4,6 +4,7 @@
   "klass": "Class",
   "groupBy": "Group By <%= type %>",
   "classBonus": "(This item matches your class, so it gets an additional 1.5 stat multiplier.)",
+  "twoHandedItem": "Two-handed item.",
 
   "weapon": "weapon",
   "weaponCapitalized" : "Weapon",
@@ -101,7 +102,7 @@
   "weaponSpecialSkiText": "Ski-sassin Pole",
   "weaponSpecialSkiNotes": "A weapon capable of destroying hordes of enemies! It also helps the user make very nice parallel turns. Increases Strength by <%= str %>. Limited Edition 2013-2014 Winter Gear.",
   "weaponSpecialCandycaneText": "Candy Cane Staff",
-  "weaponSpecialCandycaneNotes": "A powerful mage's staff. Powerfully DELICIOUS, we mean! Two-handed weapon. Increases Intelligence by <%= int %> and Perception by <%= per %>. Limited Edition 2013-2014 Winter Gear.",
+  "weaponSpecialCandycaneNotes": "A powerful mage's staff. Powerfully DELICIOUS, we mean! Increases Intelligence by <%= int %> and Perception by <%= per %>. Limited Edition 2013-2014 Winter Gear.",
   "weaponSpecialSnowflakeText": "Snowflake Wand",
   "weaponSpecialSnowflakeNotes": "This wand sparkles with unlimited healing power. Increases Intelligence by <%= int %>. Limited Edition 2013-2014 Winter Gear.",
 
--- a/website/client-old/js/controllers/inventoryCtrl.js
+++ b/website/client-old/js/controllers/inventoryCtrl.js
@@ -365,6 +365,12 @@ habitrpg.controller("InventoryCtrl",
       }
     };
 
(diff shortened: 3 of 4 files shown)


website/common/locales/en/gear.json
```diff
@@ -101,7 +102,7 @@
   "weaponSpecialSkiText": "Ski-sassin Pole",
   "weaponSpecialSkiNotes": "A weapon capable of destroying hordes of enemies! It also helps the user make very nice parallel turns. Increases Strength by <%= str %>. Limited Edition 2013-2014 Winter Gear.",
   "weaponSpecialCandycaneText": "Candy Cane Staff",
-  "weaponSpecialCandycaneNotes": "A powerful mage's staff. Powerfully DELICIOUS, we mean! Two-handed weapon. Increases Intelligence by <%= int %> and Perception by <%= per %>. Limited Edition 2013-2014 Winter Gear.",
```
paglias, line 104: This is only going to work on the website but will remove completely any info about the item being two-handled in mobile apps for example

paglias (changes requested):

website/common/locales/en/gear.json
AndrzejDoyle: That's a good point (though note that we're already in that situation with every other two-handed weapon).

I'll look at what we can do to add this in the mobile app as well.

Alys: We do want "Two-handed weapon." to be removed from that text now, for consistency with all the other two-handed gear. This is the only item that specifies the handedness in its Notes. See the issue for details (https://github.com/HabitRPG/habitica/issues/8441).

The purpose of the issue is to have "Two-handed item."  automatically added to the item descriptions returned by the API, so the mobile apps should pick up that extra text automatically. However @AndrzejDoyle it looks like you've added it in the website's code only.

.gitignore
```diff
@@ -41,3 +41,4 @@ test/client-old/spec/mocks/translations.js
 .elasticbeanstalk/*
 !.elasticbeanstalk/*.cfg.yml
 !.elasticbeanstalk/*.global.yml
+/nbproject/private/
```
khipkin, line 44: What is this for?

Alys: @AndrzejDoyle  Could you please undo your change to the .gitignore file? We use it for only files that will appear in everybody's local installs, rather than for files specific to certain set-ups. There's some other options for ignoring your own files here: https://help.github.com/articles/ignoring-files/

AndrzejDoyle: Agreed - this was not intended to make the pull request, I'll remove it from my branch.

SabreCat:
Given this PR will need to be entirely redone to be server-side instead of Web client side, and there hasn't been activity on it in months, I'm closing this for now. @AndrzejDoyle, if you'd still like to tackle this, submit a new PR!

PR #8492, opened 2017-02-15
Aded challenge migration sync

TheHollidayInn:

Fixes put_issue_url_here

### Changes

This script will iterate through all challenges, find users in the challenges, then sync the challenge with those users. 

- [x] fix createdAt
- [x] fix task attribute

----
UUID:


TheHollidayInn:
This is pre: #8469

migrations/challenges/sync-all-challenges.js
```diff
@@ -0,0 +1,45 @@
+import Bluebird from 'Bluebird';
+
+import { model as Challenges } from '../../website/server/models/challenge';
+import { model as User } from '../../website/server/models/user';
+
+async function syncChallengeToMembers (challenges) {
+  challenges.forEach(async function (challenge) {
+    let users = await User.find({challenges: challenge._id}).exec();
+
+    users.forEach(async function (user) {
```
paglias: forEach doesn't work well with async because it has no return value, so nothing here will be waited for. You should transform this code to 

```
    await users.map(function () {
        .... 
        return Bluebird.all(promises);
    })
```

migrations/challenges/sync-all-challenges.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+async function syncChallenges (lastChallengeDate) {
+  let query = {};
+
+  if (lastChallengeDate) {
+    query.createdOn = { $lte: lastChallengeDate };
+  }
+
+  await Challenges.find(query)
```
paglias: `let challengesFound = await Challenges.find().limit().sort(...).exec()`

then remove the `.then` and do

```
      let syncedChallenges = await syncChallengeToMembers(challengesFound);
      let lastChallenge = challengesFound[challengesFound.length - 1];
      if (lastChallenge) await syncChallenges(lastChallenge.createdAt);
      return syncedChallenges;
```

migrations/challenges/sync-all-challenges.js
```diff
@@ -0,0 +1,45 @@
...
+    .exec()
+    .then(async function (challengesFound) {
+      let syncedChallenges = await syncChallengeToMembers(challengesFound);
+      let lastChallenge = challengesFound[challengesFound.length - 1];
+      if (lastChallenge) syncChallenges(lastChallenge.createdAt);
+      return syncedChallenges;
+    });
+};
+
+module.exports = async function syncAllChallenges() {
```
paglias: this is not necessary, you can just export syncChallenge which when called will return a promise

migrations/migration-runner.js
```diff
@@ -17,5 +17,5 @@ function setUpServer () {
 setUpServer();
 
 // Replace this with your migration
-var processUsers = require('./new_stuff');
-processUsers();
+var syncAllChallenges = require('./challenges/sync-all-challenges');
+syncAllChallenges();
```
paglias: here probably you want to add a `.catch` otherwise errors will be swallowed

paglias (dismissed):

migrations/challenges/sync-all-challenges.js
```diff
@@ -0,0 +1,40 @@
+import Bluebird from 'Bluebird';
+
+import { model as Challenges } from '../../website/server/models/challenge';
+import { model as User } from '../../website/server/models/user';
+
+async function syncChallengeToMembers (challenges) {
+  challenges.forEach(async function (challenge) {
```
paglias: this will still do all the challenges in parallel and it'll return at line 19 immediately because forEach doesn't work with await, it'll kickstart the async function and then return

TheHollidayInn: So, what is wrong with that?

TheHollidayInn: I guess if a user is in two challenges that would be an issue?

paglias: well the problem is that any error inside these async functions will get swallowed as they're not part of a chain with a `catch` statement or must be wrapped inside a try/catch `async function () { try {code} catch (e) {handle}` and also that some challenges are really big and the database would crash. Consider that the onboarding emails process only 10 users in parallel and the cpu on the database grows to more than 20%

paglias: and while you have a limit(10) in the query it'll start a new one as soon as the forEach has run without waiting for the first 10 challenges to be processed

Alys:
Has this been tested on sample data since the most recent changes were made?

If not, I can set up test challenge tasks again, to check that the script correctly updates them.

Alys:
I'm sorry about that revert commit there. :( I didn't submit it in a new PR so it hasn't gone into `develop`, but be careful if you're still using the TheHollidayInn:challenge-sync-migration branch. It's been a while since I tried to do a reversion and I'd forgotten how it worked. Reversion isn't really important for this anyway - it has some bugs that I've documented in our google doc but since it's a script that's only used manually, we can just avoid using it until it's fixed.

PR #8493, opened 2017-02-16
Implement iOS subscriptions

phillipthelen:
Adds API calls, so that the iOS app can offer subscriptions.

### DIFF
--- a/test/api/v3/integration/payments/apple/GET-payments_apple_cancelSubscribe.js
+++ b/test/api/v3/integration/payments/apple/GET-payments_apple_cancelSubscribe.js
@@ -2,7 +2,7 @@ import {generateUser} from '../../../../../helpers/api-integration/v3';
 import applePayments from '../../../../../../website/server/libs/applePayments';
 
 describe('payments : apple #cancelSubscribe', () => {
-  let endpoint = '/iap/ios/subscribe/cancel';
+  let endpoint = '/iap/ios/subscribe/cancel?noRedirect=true';
   let user;
 
   beforeEach(async () => {
--- a/test/api/v3/integration/payments/apple/GET-payments_apple_cancelSubscribe.js
+++ b/test/api/v3/integration/payments/apple/GET-payments_apple_cancelSubscribe.js
@@ -23,6 +23,7 @@ describe('payments : apple #cancelSubscribe', () => {
     it('cancels the subscription', async () => {
       user = await generateUser({
         'profile.name': 'sender',
+        'purchased.plan.paymentMethod': 'Apple',
         'purchased.plan.customerId': 'customer-id',
         'purchased.plan.planId': 'basic_3mo',
         'purchased.plan.lastBillingDate': new Date(),
--- a/test/api/v3/integration/payments/apple/GET-payments_apple_cancelSubscribe.js
+++ b/test/api/v3/integration/payments/apple/GET-payments_apple_cancelSubscribe.js
@@ -1,0 +1,40 @@
+import {generateUser} from '../../../../../helpers/api-integration/v3';
+import applePayments from '../../../../../../website/server/libs/applePayments';
+
+describe('payments : apple #cancelSubscribe', () => {
+  let endpoint = '/iap/ios/subscribe/cancel';
+  let user;
+
+  beforeEach(async () => {
+    user = await generateUser();
+  });
+
+  describe('success', () => {
+    let cancelStub;
+
+    beforeEach(async () => {
+      cancelStub = sinon.stub(applePayments, 'cancelSubscribe').returnsPromise().resolves({});
+    });
+
+    afterEach(() => {
+      applePayments.cancelSubscribe.restore();
+    });
+
+    it('cancels the subscription', async () => {
+      user = await generateUser({
+        'profile.name': 'sender',
+        'purchased.plan.customerId': 'customer-id',
+        'purchased.plan.planId': 'basic_3mo',
+        'purchased.plan.lastBillingDate': new Date(),
+        balance: 2,
+      });
+
+      await user.get(endpoint);
+
+      expect(cancelStub).to.be.calledOnce;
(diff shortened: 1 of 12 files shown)


website/server/libs/applePayments.js
```diff
@@ -0,0 +1,172 @@
...
+  } else {
+    throw new NotAuthorized(api.constants.RESPONSE_INVALID_RECEIPT);
+  }
+};
+
+
+api.cancelSubscribe = async function cancelSubscribe (user, headers) {
+  let data = user.purchased.plan.additionalData;
+
+  if (!data) throw new NotAuthorized(shared.i18n.t('missingSubscription'));
```
paglias: I think purchased.plan.additionalData always exist

test/api/v3/integration/payments/apple/GET-payments_apple_cancelSubscribe.js
```diff
@@ -0,0 +1,40 @@
+import {generateUser} from '../../../../../helpers/api-integration/v3';
+import applePayments from '../../../../../../website/server/libs/applePayments';
+
+describe('payments : apple #cancelSubscribe', () => {
+  let endpoint = '/iap/ios/subscribe/cancel';
```
paglias: the enpoint should include the noRedirect query parameter because the home takes too long to load an may cause a test timeout

website/server/controllers/top-level/payments/iap.js
```diff
@@ -92,88 +87,51 @@ api.iapiOSVerify = {
   url: '/iap/ios/verify',
   middlewares: [authWithHeaders()],
   async handler (req, res) {
+
+    let appleRes = await applePayments.verifyGemPurchase(res.locals.user, req.body.transaction.receipt, req.headers);
```
paglias: if req.body.transaction doesn't exist it'll crash, should probably have a validation of the body

website/server/controllers/top-level/payments/iap.js
```diff
@@ -92,88 +87,51 @@ api.iapiOSVerify = {
...
+ * @api {post} /iap/android/subscription iOS Subscribe
+ * @apiName IapiOSSubscribe
+ * @apiGroup Payments
+ **/
+api.iapSubscriptioniOS = {
+  method: 'POST',
+  url: '/iap/ios/subscribe',
+  middlewares: [authWithUrl],
+  async handler (req, res) {
+    if (!req.body.sku) throw new BadRequest(res.t('missingSubscriptionCode'));
```
paglias, line 109: req.body.receipt should be checked too?

website/server/libs/applePayments.js
```diff
@@ -0,0 +1,172 @@
...
+  let purchaseDataList = iap.getPurchaseData(appleRes);
+  if (purchaseDataList.length === 0) throw new NotAuthorized(api.constants.RESPONSE_NO_ITEM_PURCHASED);
+  let correctReceipt = false;
+
+  // Purchasing one item at a time (processing of await(s) below is sequential not parallel)
+  for (let index in purchaseDataList) {
+    let purchaseData = purchaseDataList[index];
+    let token = purchaseData.transactionId;
+
+    let existingReceipt = await IapPurchaseReceipt.findOne({ // eslint-disable-line no-await-in-loop
```
paglias, line 37: is this working? I'm not 100% sure that await works correctly inside a for in loop, it works in a `for of` but I'm not sure about a for in. I know this has been used before but I just got this doubt

paglias:
Overall 👍

website/server/libs/applePayments.js
phillipthelen: We've been using this await in a for loop since the gem purchases were added and it worked so far. During my testing now, I also didn't notice any issues. The reason for the eslint warning seems to rather be, that using await like this is not taking advantage of parallelization: http://eslint.org/docs/rules/no-await-in-loop

paglias: yeah I was concerned because that page only mentioned for-of loops and not for-in, but I checked the source and it's fine

website/server/libs/applePayments.js
Alys: This might be irrelevant since you're talking about a new subscription payment method, but I've never seen purchased.plan.additionalData in any of the subscriptions I've ever looked at.

phillipthelen: That field was added when I added support for google/android subscriptions. So only a fee have it so far

website/server/libs/googlePayments.js
```diff
@@ -131,13 +131,13 @@ api.subscribe = async function subscribe (sku, user, receipt, signature, headers
 
 
 api.cancelSubscribe = async function cancelSubscribe (user, headers) {
-  let data = user.purchased.plan.additionalData;
+  let plan = user.purchased.plan;
 
-  if (!data) throw new NotAuthorized(shared.i18n.t('missingSubscription'));
+  if (plan.paymentMethod !== 'Google') throw new NotAuthorized(shared.i18n.t('missingSubscription'));
```
paglias: constant here too?

website/server/libs/applePayments.js
```diff
@@ -142,13 +142,13 @@ api.subscribe = async function subscribe (sku, user, receipt, headers, nextPayme
 
 
 api.cancelSubscribe = async function cancelSubscribe (user, headers) {
-  let data = user.purchased.plan.additionalData;
+  let plan = user.purchased.plan;
 
-  if (!data) throw new NotAuthorized(shared.i18n.t('missingSubscription'));
+  if (plan.paymentMethod !== 'Apple') throw new NotAuthorized(shared.i18n.t('missingSubscription'));
```
paglias: should we use the constan `applePayments.constants.PAYMENT_METHOD_APPLE ` since it exists?

paglias (changes requested):

PR #8503, opened 2017-02-22
Use backticks in 'quest started' system message

### LINKED ISSUE #8445: "Quest started" system message should use backticks.
The quest-started message (added in #8107) is not wrapped in backticks, so it stands out from the other system messages:

image

maximumtiu:
Fixes #8445

### Problem

All automated system messages in party chat are written with backticks to format as a `<code>` HTML element. Except the message to signify a quest starting, which is unfortunately quite inconsistent.

### Solution

Make it match the other messages! Format with backticks.

screen shot 2017-02-21 at 9 38 37 pm

----
UUID: bd794f6e-3845-49e5-88cc-a140b24cdea4

### DIFF
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -554,7 +554,7 @@ schema.methods.startQuest = async function startQuest (user) {
         });
     });
   });
-  this.sendChat(`Your quest, ${quest.text('en')}, has started.`, null, {
+  this.sendChat(`\`Your quest, ${quest.text('en')}, has started.\``, null, {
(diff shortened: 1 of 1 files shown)


Alys:
I've tested this locally and it looks good. Thank you @megantiu!  Tier 2 awarded. 

Note that as your tier increases, it takes progressively more work to achieve the next one, so for the higher tiers, larger PRs or a few PRs are required. There's some information about that at [Contributor Rewards](http://habitrpg.wikia.com/wiki/Contributor_Rewards) on the wiki but feel free to ask here if you have questions!

PR #8507, opened 2017-02-23
Tasks score notes

TheHollidayInn:

Fixes put_issue_url_here

### Changes
This adds a user preference that will prompt them to add notes to a task when they score. This is a feature requested by groups/organizations to have notes on details for completing a task. 

Right now, this works on habits and is a user preference. But this can be expanded as a full group/org option so they can have more reviewing details. 

- [x] limit notes

----
UUID:


website/server/controllers/api-v3/tasks.js
```diff
@@ -509,11 +509,14 @@ api.scoreTask = {
     if (validationErrors) throw validationErrors;
 
     let user = res.locals.user;
+    let scoreNotes = req.body.scoreNotes;
```
paglias, line 515: can we add some validation limiting how big the note can be? Also we should document the field

Alys:
When the setting for "Confirm task scoring with notes" is turned on, the Habits' plus and minus buttons can't be used. When you move your mouse over the buttons, the mouse pointer changes to indicate the buttons are clickable and the buttons change colour slightly as normal, but clicking on them has no effect. The buttons work correctly when the setting is turned off, so I think the bug is due to this PR rather than another pending one. _[**EDIT**: I think that bug is NOT from this PR. See one of my later comments.]_

Also, when a user turns this feature on, what happens if they use Habits in the mobile apps? Or in the Android app's widget? If the new setting doesn't work but Habits can still be scored as if the feature wasn't turn on, I personally think that's acceptable (we can warn users about that when we release the feature and also put a warning on the feature's option on the Settings page), but if it's going to cause errors, that would be bad. FYI @vIiRuS 

Also what happens when a user turns on the feature and scores a Habit using a third-party tool that doesn't have support built in for the feature. Again, if the Habits can still be scored as if the feature wasn't turn on, I think that's acceptable, but if it's going to cause errors, we should give all third-party devs a warning (also the users, as above). Perhaps we should wait a few extra weeks before merging this to production, to give devs time to provide support (that's just a suggestion).

phillipthelen:
As far I can tell, the server side does not enforce this. So with this feature enabled, the apps would still work the same until we implement the feature.

Am I understanding it correctly, that you can only see the last scoreNote? Doesn't that limit the usefulness especially for habits?

Alys:
@TheHollidayInn Another question: are the notes visible in the user's exported data (using the options from the website's Data menu)? Also are they visible when you fetch the user's tasks using the API? It might help third-party devs integrate this feature if you could edit the top post to include some brief information about how the data is enhanced, perhaps with a couple of sample sections of JSON data.

Alys:
Regarding this note from me above: _"When the setting for "Confirm task scoring with notes" is turned on, the Habits' plus and minus buttons can't be used...."_
I'm no longer sure that bug is from this PR. I've tested this PR on a local install that was based on 6f0d0b1fb37ba2c9756d52e0fb3c2c7ff42edff6 and then had 6f0d0b1fb37ba2c9756d52e0fb3c2c7ff42edff6 cherry picked onto it, and the Habit buttons work when this new feature is turned on. I'm still looking for the source of the bug.

This feature does allow me to add a note when I mark a task as complete, however I can't find where to see the note afterwards. It doesn't seem to be visible from the tasks page, and isn't in the Data > Export Data > JSON file, and I can't find it in the database's task or user document.

Alys:
I tested this locally (very briefly) by checking out 93befcebcc7508d7106996b6c11b117708dbd7fe and running `npm start`. That includes all previous PRs mentioned in my comments at https://github.com/HabitRPG/habitica/pull/7781#issuecomment-283015166 and https://github.com/HabitRPG/habitica/pull/8469#issuecomment-283013476

The edit bug mentioned in my latter comment does not occur in this PR.

PR #8444, opened 2017-01-22
Repeatables - more repeat options for Dailies (weeklies, monthlies, yearlies, etc)

TheHollidayInn:
[//]: # (Note: See http://habitica.wikia.com/wiki/Using_Habitica_Git#Pull_Request for more info)

[//]: # (Put Issue # or URL here, if applicable. This will automatically close the issue if your PR is merged in)
Fixes put_issue_url_here

### Changes
[//]: # (Describe the changes that were made in detail here. Include pictures if necessary)
This adds the business logic for the following:
1. Every X weeks
2. Every X Months
3. Every month on a day of the month
4. Every month on a day of the week
5. Every X years

Todo:
- [x] Combine every X with day of the week, month, etc
- [x] Add model fields
- [x] Add UI

[//]: # (Put User ID in here - found in Settings -> API)

----
UUID:

PR #8509, opened 2017-02-24
Add isDue prop to tasks

crookedneighbor:
### Changes

First pass at adding `task.isDue` to tasks. This will make API users (such as the mobile apps) not need to duplicate complex and ever changing shouldDo logic.

We need to have access to the user's preferences when calculating if a daily is do, so, unfortunately, we have to call a method on each route that sends back the user's task object.

Open to suggestions in regard to naming these methods and property.

### DIFF
--- a/test/api/v3/integration/tasks/GET-tasks_challenge_challengeId.test.js
+++ b/test/api/v3/integration/tasks/GET-tasks_challenge_challengeId.test.js
@@ -61,6 +61,12 @@ describe('GET /tasks/:taskId', () => {
         expect(getTask).to.eql(task);
       });
 
+      it('does not include isDue property', async () => {
+        let getTask = await user.get(`/tasks/${task._id}`);
+
+        expect(getTask.isDue).to.not.exist;
+      });
+
       it('returns error when user is not a member of the challenge', async () => {
         let anotherUser = await generateUser();
 
--- a/test/api/v3/integration/tasks/GET-tasks_id.test.js
+++ b/test/api/v3/integration/tasks/GET-tasks_id.test.js
@@ -34,6 +34,12 @@ describe('GET /tasks/:id', () => {
       expect(getTask).to.eql(task);
     });
 
+    it('includes isDue property', async () => {
+      let getTask = await user.get(`/tasks/${task.alias}`);
+
+      expect(getTask.isDue).to.be.a('boolean');
+    });
+
     // TODO after challenges are implemented
     it('can get active challenge task that user does not own'); // Yes?
   });
--- a/test/api/v3/integration/tasks/GET-tasks_user.test.js
+++ b/test/api/v3/integration/tasks/GET-tasks_user.test.js
@@ -15,6 +15,16 @@ describe('GET /tasks/user', () => {
     expect(tasks.length).to.equal(createdTasks.length + 1); // + 1 because 1 is a default task
   });
 
+  it('includes isDue properties on user tasks', async () => {
+    await user.post('/tasks/user', [{text: 'test habit', type: 'habit'}, {text: 'test todo', type: 'todo'}]);
+    let tasks = await user.get('/tasks/user');
+    expect(tasks.length).to.be.greaterThan(0);
+
+    tasks.forEach((task) => {
+      expect(task.isDue).to.be.a('boolean');
+    });
+  });
+
   it('returns only a type of user\'s tasks if req.query.type is specified', async () => {
     let createdTasks = await user.post('/tasks/user', [
       {text: 'test habit', type: 'habit'},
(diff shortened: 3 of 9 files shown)


TheHollidayInn:
Doesn't this move us farther from an offline mode?

crookedneighbor:
Comments from @vIiRuS: 

> issue with this: the apps cache their data locally. When the field is just a boolean, we don't know if the data is still up to date or if it has to be fetched again
Having a date field for when it's next due would be more convenient (or even just returning the current date instead of a boolean

The problem with this is that the `shouldDo` function returns a boolean. We use this on the frontend to determine if we should show the daily as due and on the backend during cron.

We'd in order to have logic to return a next due date and have it be consistent with shouldDo, we'd basically have to rewrite the [shouldDo function](https://github.com/HabitRPG/habitica/blob/dfcd32d54ac36787edfe650a6b5b384651ffa62b/website/common/script/cron.js#L91-L124) as follows:

```js
// Currently, DAY_MAPPING is an object with numbers 
// from 0-6 as keys. For ease of use in nextDue, I've
// made it an array instead. Should be functionally the
// same for our uses
export const DAY_MAPPING = [
  'su',
  'm',
  't',
  'w',
  'th',
  'f',
  's',
];

export function nextDue (today, dailyTask, options = {}) {
  if (dailyTask.type !== 'daily') {
    return false;
  }

  // The time portion of the Start Date is never visible to or modifiable by the user so we must ignore it.
  // Therefore, we must also ignore the time portion of the user's day start (today), otherwise the date comparison will be wrong for some times.
  // NB: The user's day start date has already been converted to the PREVIOUS day's date if the time portion was before CDS.
  let taskStartDate = moment(dailyTask.startDate).zone(options.timezoneOffset);

  taskStartDate = moment(taskStartDate).startOf('day');
  if (taskStartDate > today) {
    return taskStartDate; // Daily starts in the future
  }
  if (dailyTask.frequency === 'daily') { // "Every X Days"
    if (!dailyTask.everyX) {
      return false; // error condition
    }
    let daysSinceTaskStart = today.diff(taskStartDate, 'days');

    let daysUntilNextDue = daysSinceTaskStart % dailyTask.everyX;
    return today.add(daysUntilNextDue, 'days');
  } else if (dailyTask.frequency === 'weekly') { // "On Certain Days of the Week"
    if (!dailyTask.repeat) {
      return false; // error condition
    }
    let dayOfWeekNum = today.day(); // e.g., 0 for Sunday
    let weekLength = DAY_MAPPING.length;

    // this is a bit complicated, but it basically 
    // does the math for what day we're on starting
    // at the current day of the week and going into the future
    // 7 days
    for (let i = dayOfWeekNum; i < weekLength + dayOfWeekNum; i++) {
      let day = i >= weekLength ? i - weekLength : i;

      if (dailyTask.repeat[DAY_MAPPING[day]]) {
        return today.add(i - dayOfWeekNum, 'days');
      }
    }
  }

  return false; // error condition - unexpected frequency string or weekly repeats were all not set
}

export function shouldDo (day, dailyTask, options = {}) {
  if (dailyTask.type !== 'daily') {
    return false;
  }
  let o = sanitizeOptions(options);
  let startOfDayWithCDSTime = startOfDay(_.defaults({ now: day }, o)).startOf('day');

  return nextDue(startOfDayWithCDSTime, dailyTask, o) === startOfDayWithCDSTime;
}
```

I made this change locally and all tests pass, but I noticed that we don't actually have any unit test coverage around `shouldDo`, so I think this kind of change may be blocked until we add unit test coverage around it to ensure that the behavior does not change with this refactor.

That said, I think @vIiRuS is right and we should go this route. 

One lingering question I have about all this is the proper thing to return when the daily is never due in the future. IE, a weekly with no days of the week set.

crookedneighbor:
@TheHollidayInn 

I think the parameters of what is desired with offline mode are not clear to me. For instance, how does it interact with cron? 

Ultimately, I think what we want is a way for API consumers to know if a daily is due or not without having to replicate the logic from the website. In many cases, in a language other than JavaScript. We also want to be able to add new logic to dailys (monthlies, etc) without requiring API consumers to update that logic.

IMO, whether or not the mobile apps actually use this is irrelevant. It's still a good addition to make for 3rd party API consumers.

Alys:
"It's still a good addition to make for 3rd party API consumers." -- Totally totally totally agree! I have two third-party tools that would use this and would immediately become more accurate, and it would be even more important in future when monthlies are added.

TheHollidayInn:
Before rewriting the shouldDo function, please note my PR: https://github.com/HabitRPG/habitica/pull/8444

Alys:
A bug report today meminded me of this edge case:

A Daily that has no days of the week selected or has "every X days" set to zero should be treated as never due. They are both legitimate settings for creating Grey Dailies.

The same will apply to "every 0 months/years" when monthly and yearly repeats exist.

TheHollidayInn:
What is the progress on this @crookedneighbor ?

TheHollidayInn:
Closing in favor of #8711

PR #8513, opened 2017-02-25
set _cronSignature to current time instead of uuid when cron is running fixes #8163

### LINKED ISSUE #8163: cron does not run when _cronSignature is permanently not set to NOT_RUNNING
When cron starts, it adjusts the user's document to set `_cronSignature`  to the user's UID. When cron ends, it sets the value of `_cronSignature` to `NOT_RUNNING`. This is correct behaviour and is used to prevent cron running more than once at the same time.

However if something stops cron from completing (e.g., a timeout or other error), then `_cronSignature` is never reset to `NOT_RUNNING`. This prevents the user from being able to use Habitica in any way - they are not able to log in, or load the website if they are already logged in, or access their account in the mobile app.

In the past few days, we've had emails from three users affected by this. I've done a search in the database and found another 57 users affected by it, with last loggedin dates ranging from 2016-07-15 to today (for today, I checked that they weren't actually in the middle of a real cron). 16 of those users had last loggedin dates of 2016-10-19T17:31 to 2016-10-19T17:33, so they were probably affected by the CPU spike we had then ([ref for admins](https://habitica.slack.com/archives/server-health/p1476898552000013)). Another 3 users had dates of 2016-10-23T00:10 and 2016-10-23T00:11, correlating with today's server alarms ([admin ref](https://habitica.slack.com/archives/server-health/p1477181591000058)).

I've fixed all those users by setting their `_cronSignature` to `NOT_RUNNING` so they can all log in again, although I expect that some have given up trying. :(  I also made them all be resting in the inn so they won't take damage from missed Dailies.

_EDIT 2017-02-01:_
**The desired fix is that instead of setting `_cronSignature` to the User ID, set it to the current datetime, and then allow cron to run if the value is more than an hour old.**

<strike>We should adjust cron so that either it's impossible for `_cronSignature` to ever be left in an inappropriate state (might not be possible to make the code that robust) or it can recognise when it should run even though `_cronSignature` isn't set to NOT_RUNNING. Perhaps instead of setting `_cronSignature` to the User ID, we could set it to the current datetime, and then allow cron to run if the value is more than, say, an hour old. Other suggestions are welcome of course. :)</strike>

KyleVerma:

Fixes https://github.com/HabitRPG/habitica/issues/8163

### Changes

_cronSignature now set to now.getTime() instead of uuid()

created const cronTimeout, which is set to 1hour (in ms) can be changed easily later

where the code used to make sure _cronSignature was 'NOT_RUNNING' is now an $or that also checks to see if _cronSignature is cronTimeout ms smaller than the current time

All _cronSignature time calculations are in ms since epoch time to hopefully avoid issues with timezones and end of day

----
UUID: 0036df29-70e5-4700-8e38-d7dd964a6401


website/server/middlewares/cron.js
```diff
@@ -6,9 +6,10 @@ import Bluebird from 'bluebird';
 import { model as Group } from '../models/group';
 import { model as User } from '../models/user';
 import { recoverCron, cron } from '../libs/cron';
-import { v4 as uuid } from 'uuid';
 
 const daysSince = common.daysSince;
+// Wait this length of time in ms before attempting another cron
+const cronTimeout = new Date(3600000).getTime();
```
paglias: we try to name constants with all upper case letters, so this would be CRON_TIMEOUT. Also for the length of time maybe a simple, 1000 * 60 * 60 is easier to read (with a comment saying that it's 1h)

paglias (changes requested):

website/server/middlewares/cron.js
Alys: I suggest that the comment should say something like just `// milliseconds` with no value specified (the existing comment might be enough). If the comment says specifically `1 h` there's a good chance we'll forget to update it if we change the value, and comments that get out of sync with the code are evil incarnate. :)

KyleVerma: Name changed. I kept the comment the same so that it will be accurate if a different time is selected in the future

KyleVerma:
What is the cron middleware: recovers from failed cron and does not error when user is already cronning test actually doing? I can't get a handle on what's actually happening in that test so I can't tell if there's a problem with my code or the test just needs to be changed to match the new code.

In a more general sense is just learning how mocha works sufficient to begin understanding how the tests work or are there other things I should learn as well?

paglias:
@KyleVerma mocha is mostly a wrapper to make testing easier, its knowledge isn't super important and anyway quite simple. That tests makes sure that when cron fails (setting nMatched to 0) causes [this error](https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/cron.js#L121) to be thrown, the recover cron function is called and the cron process continues without erroring.

paglias:
@KyleVerma the test is failing because it expects to be called with

```
{ _id: user._id, _cronSignature: 'NOT_RUNNING' }
```

but with your code changes it'll be called with something different, similar to

```
{ _id: user._id, _cronSignature: $or: [....] }
```

so the call won't be matched here https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/unit/middlewares/cronMiddleware.js#L210 and it won't return the result with nMatched = 0.

What we're doing here is using a `stub` with [sinon.js](http://sinonjs.org/) that allows us to replace a piece of code for testing purposes so that in this case `User.update` will return the result that we want to see and that is needed to test a specific behavior.

paglias:
@KyleVerma looks like there could have been some problem when merging develop into your branch

KyleVerma:
As far as I can tell it's still the same failing test. If that's not true where do I look to figure out the problem?

Also I've been looking into fixing the stub but I'm having trouble figuring out how. How do I represent the timeoutThreshold in the $or? Or am I thinking about this the wrong way?

(sorry for the long delay on this I was on holiday)

paglias:
@KyleVerma no worries, I was referring to the fact that the PR shows a lot of unrelated commits. To fix it see this comment about a similar situation https://github.com/HabitRPG/habitica/pull/8539#issuecomment-284984787

KyleVerma:
Here's the new pull request: https://github.com/HabitRPG/habitica/pull/8565

PR #8531, opened 2017-03-04
invalid class change throws error (#8496)

### LINKED ISSUE #8496: API Issue: user/change-class doesn't validate class listed
General Info
  * UUID:  b0413351-405f-416f-8787-947ec1c85199

### Description

If the class given as a parameter is invalid in an API call, a success is still returned

for example:
`user/change-class?class=cellist`

will still return:
`{"success":true,"data":{etc...}}`
(although the class will still remain the same as it was)

lulock:

Fixes #8496

### Changes

If an invalid class is specified during API call to change-class, an error is thrown. 

for example:
> user/change-class?class=cellist

will return:
> {"success":false,"error":"BadRequest","message":"Invalid class. Please specify warrior, rogue, wizard, or healer."}

If no class is specified, this is perceived as a "class reset" and returns true. Class points are re-allocated, 3 gems are paid, and the user is prompted to choose a class in the GUI (just as would occur upon levelling up to 10 to begin with).  

**To clarify**, this is also how the class system is enabled after the user has choosen to opt out:

image

as can be seen in the _**User.changeClass**_ function call below:

image

without charging gems again:

image

----
UUID: 0bff3694-938f-4e8c-8acc-ba0726f58a26


website/common/locales/en_GB/character.json
```diff
@@ -112,6 +112,7 @@
     "mystery": "Mystery",
     "changeClass": "Change Class, Refund Attribute Points",
     "lvl10ChangeClass": "To change class you must be at least level 10.",
+    "invalidClass":"Invalid class. Please specify warrior, rogue, wizard, or healer.",
```
paglias: can you revert the changes to the en_GB locales? On github you should only change the American English files, the other languages are managed using Transifex

website/common/script/ops/changeClass.js
```diff
@@ -47,32 +71,19 @@ module.exports = function changeClass (user, req = {}, analytics) {
...
-    user.stats.int = 0;
-    user.stats.points = capByLevel(user.stats.lvl);
-    user.flags.classSelected = false;
+        // if invalid class is specified, throw an error.
+    throw new BadRequest(i18n.t('invalidClass', req.language));
   }
-
   return [
-    pick(user, splitWhitespace('stats flags items preferences')),
+        pick(user, splitWhitespace('stats flags items preferences')),
```
paglias: indentation seems wrong here and in a few other places (lines 87, 83, 74-79)

website/common/script/ops/changeClass.js
```diff
@@ -6,16 +6,40 @@ import pick from 'lodash/pick';
...
+
 module.exports = function changeClass (user, req = {}, analytics) {
   let klass = get(req, 'query.class');
-
-  // user.flags.classSelected is set to false after the user paid the 3 gems
+    // user.flags.classSelected is set to false after the user paid the 3 gems
   if (user.stats.lvl < 10) {
     throw new NotAuthorized(i18n.t('lvl10ChangeClass', req.language));
-  } else if (!user.flags.classSelected && (klass === 'warrior' || klass === 'rogue' || klass === 'wizard' || klass === 'healer')) {
+  } else if (typeof klass === 'undefined') {
```
paglias: if (!klass) maybe?

paglias (changes requested):
Left a few comments, overall it looks good!

website/common/locales/en_GB/character.json
lulock: I was wondering about that ... good to know, thanks!

website/common/script/ops/changeClass.js
paglias: not sure why the build hasn't failed since this indentation should not be considered valid by ESLint

lulock: Testing locally at first gave me a bunch of errors from ESLint for indentation and so I corrected them all before making my pull request. Or so I thought ...

lulock: I just fixed up your requests and I am running the tests again now :) Do I make a new pull request once I'm done or is there some way to update this one with my new commit?

paglias: If you push another commit to your fork (lulock:develop) it'll show up here automatically :) 

```
git commit ...
git push origin develop
```

(`origin` might differ based on your setup)

lulock:
pushed new commits. hope it's ok now!

website/common/script/ops/changeClass.js
```diff
@@ -6,16 +6,40 @@ import pick from 'lodash/pick';
...
+  user.stats.int = 0;
+  user.stats.points = capByLevel(user.stats.lvl);
+  user.flags.classSelected = false;
+}
+
 module.exports = function changeClass (user, req = {}, analytics) {
   let klass = get(req, 'query.class');
-
-  // user.flags.classSelected is set to false after the user paid the 3 gems
+    // user.flags.classSelected is set to false after the user paid the 3 gems
```
paglias: indentation

paglias (changes requested):

website/common/script/ops/changeClass.js
lulock: fixed

test/common/ops/changeClass.js
```diff
@@ -27,6 +28,19 @@ describe('shared.ops.changeClass', () => {
...
+    user.flags.classSelected = false;
+    user.preferences.disableClasses = false;
+
+    try {
+      changeClass(user, {query: {class: 'cellist'}});
+    } catch (err) {
+      expect(err).to.be.an.instanceof(BadRequest);
+      expect(err.message).to.equal(i18n.t('invalidClass'));
+      done();
+    }
```
khipkin, line 41: As it is right now, if this functionality is broken again in the future, causing no error to be thrown due to a nonsense class value, this test will give a false positive. No assertions will fail, but none will pass either. A test no-op.

Suggestion: add an `assert(false)` or something after the `try`/`catch` block in such a way that it will never be reached if the expected error is thrown. This will make the test more robust and catch any future regressions.

website/common/locales/en/character.json
```diff
@@ -112,6 +112,7 @@
     "mystery": "Mystery",
     "changeClass": "Change Class, Refund Attribute Points",
     "lvl10ChangeClass": "To change class you must be at least level 10.",
+    "invalidClass":"Invalid class. Please specify warrior, rogue, wizard, or healer.",
```
khipkin: Suggestion: maybe put quotation marks or backticks around the acceptable values?

website/common/script/ops/changeClass.js
```diff
@@ -56,23 +80,10 @@ module.exports = function changeClass (user, req = {}, analytics) {
...
-    user.flags.classSelected = false;
+    // if invalid class is specified, throw an error.
+    throw new BadRequest(i18n.t('invalidClass', req.language));
   }
-
   return [
     pick(user, splitWhitespace('stats flags items preferences')),
   ];
-};
+};
```
khipkin: Nit: add this nl back.

test/common/ops/changeClass.js
lulock: thanks for these suggestions! 

I'm confused about this, though. How does this differ from all other error tests in this same file (or in other test files for that matter). All expect an error to be thrown and only call done() in the catch block. From what I understand the test will return a timeout error anyway, and all error tests seem to behave that way. Should they all be modified in that case?

khipkin: Oh thanks for the reply! I didn't know that's what `done()` did. That makes sense.

Since that's apparently what the other tests do, it's probably fine. However, it seems sub-optimal to sit and wait for a timeout for every possible failure case. But it might not be a big deal. Can someone else clarify please? @staff

paglias: Yes, I can confirm that if done isn't called in 8s the test will fail. It would probably be a little more efficient not to wait for this bug given that tests usually don't fail it's not a big deal if one day a build takes 8s more than usual

paglias (approved):

paglias:
This LGTM! @lulock we've switched to a new merging and testing process where PRs get merged and tested every two weeks, so it won't be merged immediately but in a few days. When it's merged I'll also note the contribution in your Habitica profile. Thanks for your help!

SabreCat:
Merged for testing! @lulock, thank you, and welcome to Blacksmith tier 1! Further tiers require increasing amounts of effort from one to the next, so Tier 2 will take multiple PRs or a larger PR. But keep helping out and we will express our gratitude accordingly!

Issue #5736, opened 2015-08-04
Negative experience in dashboard...


ghost:
<img width="158" alt="screen shot 2015-08-04 at 9 16 33 am" src="https://cloud.githubusercontent.com/assets/577460/9061421/a616d554-3a89-11e5-8809-26fd0ae4930a.png">

The score graph is positive.
<img width="313" alt="screen shot 2015-08-04 at 9 18 28 am" src="https://cloud.githubusercontent.com/assets/577460/9061463/d182ca18-3a89-11e5-9c8c-e4a52403e920.png">

deilann:
Did you un-check a task after leveling up?

If so, this is expected. We've considered reverting the level up, but have come to the conclusion that the ramifications of deleveling someone aren't worth having to work up through some negative XP.

crookedneighbor:
Another option would be to prevent experience deduction from lowering below zero.

deilann:
Pretty sure we considered that as well, but didn't think it was a good
solution because the full value should be deducted.

On Tue, Aug 4, 2015 at 10:07 AM, Blade Barringer notifications@github.com
wrote:

> Another option would be to prevent experience deduction from lowering
> below zero.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/HabitRPG/habitrpg/issues/5736#issuecomment-127677647.

## 

Currently Reading
https://www.goodreads.com/user/email_signature_destination/35215676?utm_medium=reading_link&utm_source=email_signature
[image: Book Cover]
https://www.goodreads.com/user/email_signature_destination/35215676?utm_medium=cover&utm_source=email_signature

crookedneighbor:
It's such an edge case that I'd be okay with it.

Alys:
If you level up by accidentally ticking off a red To-Do with a checklist, the XP gain can be large, and if if was deducted only down to zero, you could end up with a significant amount of unearned XP. Negative XP is better. I've seen more users disturbed by unearned things in their accounts than by negative experience. @Archimage This isn't intended to dismiss your bug report. Let us know if deilann's explanation isn't relevant to your situation.

Gondulf:
Looks similar to issue #663

Alys:
Thanks @Gondulf.

PR #8541, opened 2017-03-08
Fix character values - do not allow values less than 1 - fixes #8512

### LINKED ISSUE #8512: Fix Character Values allows values less than 1

General Info
  * UUID: e8277c2b-d1a0-47f8-9a08-31211ee578c8
  * Browser: Chrome
  * OS: Windows 10

### Description

The "Fix Character Values" feature on the Site page disallows decimal values for Level, which is correct, but it still allows you to fill in 0 or a negative number. Levels below 1 are not supported and could cause unexpected behavior with other features related to game statistics.

image

#### Console Errors

None

cailinglu:

Fixes #8512 

### Changes
- ~~Display an error message if invalid character values are entered in the 'fix character values' section~~
- ~~I put an invalidValue message in the en locale settings.json file that says: "Invalid value: Level must be 1 or greater. All other values must be 0 or greater."~~
  - ~~Let me know if this should be modified. This was what I have as a general message to display, rather than displaying a separate message depending on which values are invalid.~~
- Display an error message if a level less than 1 is entered in the 'fix character values' section
- Fix character values modal only closes when the user clicks on 'save and close' and the entered values are valid (can still be closed if they hit the 'cancel' button)

levelerror

----

### DIFF
--- a/test/client-old/spec/controllers/settingsCtrlSpec.js
+++ b/test/client-old/spec/controllers/settingsCtrlSpec.js
@@ -18,6 +18,7 @@ describe('Settings Controller', function () {
         releaseMounts: sandbox.stub(),
         releaseBoth: sandbox.stub(),
         setCustomDayStart: sandbox.stub(),
+        restore: sandbox.stub(),
         user: user
       };
 
@@ -27,6 +28,7 @@ describe('Settings Controller', function () {
         releasePets: sandbox.stub(),
         releaseMounts: sandbox.stub(),
         releaseBoth: sandbox.stub(),
+        restore: sandbox.stub(),
       };
 
       Notification = {
@@ -331,6 +333,44 @@ describe('Settings Controller', function () {
 
         expect(Notification.error).to.not.be.called;
         expect(Notification.text).to.be.calledWith('Coupon applied!');
+      });
+    });
+  });
+
+  context('Fixing character values', function () {
+    describe('#restore', function () {
+      it('updates character values only when valid values are entered', function () {
+        scope.restoreValues = {
+          stats: {
+            hp: -1,
+            exp: -1,
+            gp: -1,
+            lvl: 0,
+            mp: -1,
+          },
+          achievements: {
+            streak: -1,
+          },
+        };
+        scope.restore();
+        expect(User.set).to.not.be.called;
+      });
+
+      it('doesn\'t update character values when invalid values are entered', function () {
+        scope.restoreValues = {
+          stats: {
+            hp: 0,
+            exp: 0,
+            gp: 0,
+            lvl: 1,
+            mp: 0,
+          },
+          achievements: {
+            streak: 0,
+          },
+        };
+        scope.restore();
+        expect(User.set).to.be.called;
       });
     });
   });
--- a/test/client-old/spec/controllers/settingsCtrlSpec.js
+++ b/test/client-old/spec/controllers/settingsCtrlSpec.js
@@ -18,7 +18,6 @@ describe('Settings Controller', function () {
         releaseMounts: sandbox.stub(),
         releaseBoth: sandbox.stub(),
         setCustomDayStart: sandbox.stub(),
-        restore: sandbox.stub(),
         user: user
       };
 
@@ -28,7 +27,6 @@ describe('Settings Controller', function () {
         releasePets: sandbox.stub(),
         releaseMounts: sandbox.stub(),
         releaseBoth: sandbox.stub(),
-        restore: sandbox.stub(),
       };
 
       Notification = {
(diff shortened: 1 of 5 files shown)


Alys:
@caillu This looks like a good fix for the level issue, and I like that you've expanded the fix to other values as well. However Experience is actually allowed to be negative in normal game play and so it should be allowed to be negative in Fix Character Values (FCV) too (reasons below). Health can be negative briefly in normal game play, but the case for allowing that in FCV is weaker.

Experience can become negative if you level up from completing a task and then realise that you'd accidentally marked the wrong task as complete. When you uncomplete the task, the XP gained is subtracted, sometimes leading to a negative value. This doesn't do any harm because the XP increases normally as you do other tasks and rises above zero once you've earned back enough of it, and [we made a decision](https://github.com/HabitRPG/habitica/issues/5736#issuecomment-127677025) to allow negative XP in that situation because it was the clearest, simplest behaviour when unticking a task that had caused a level-up. Because negative XP can be obtained from normal usage of Habitica, FCV should allow it. I can think of a couple of uses for setting XP negative in FCV: a Socialite or wiki editor might want to set their XP negative temporarily to obtain a screenshot for illustrating that situation on the wiki. Or a player who had obtained negative XP through normal gameplay might then have had their XP increased dramatically due to overenthusiastic buffing from a party member, and might want to set the XP back to the real value so as to avoid feeling like they were cheating. Also, players who are using FCV to make the game harder for themselves might want to set their XP negative each time they level up - "hard mode" tweaks like that are not uncommon (we actually have at least one guild dedicated to them).

Health can be negative if you take enough damage from missed Dailies or a quest boss. In that situation, it becomes 50 again as soon as you take any action on the site (you die and then revive immediately). The only time you ever see negative health is very briefly on the Experience bar and then in the death modal. I imagine that a wiki editor _might_ possibly want to get a screenshot of that, but really a screenshot with a health of exactly zero would be good enough, so if we want to prohibit negative values of Health in FCV, I'm okay with that.

I think it's okay to have a single message for all error conditions in FCV, although it will get a little wordy since XP can be negative. I guess something like "Invalid value: Level must be 1 or greater. Experience may be negative or 0 or greater. All other values must be 0 or greater."

cailinglu:
I went back and made changes. Decided to only check if the level is valid since that was the original issue, and if other values being negative aren't impacting the game, then perhaps I should just leave them as is for now.

website/client-old/js/controllers/settingsCtrl.js
```diff
@@ -164,14 +164,17 @@ habitrpg.controller('SettingsCtrl',
...
         achievements = $scope.restoreValues.achievements;
-      User.set({
-        "stats.hp": stats.hp,
-        "stats.exp": stats.exp,
-        "stats.gp": stats.gp,
-        "stats.lvl": stats.lvl,
-        "stats.mp": stats.mp,
-        "achievements.streak": achievements.streak
-      });
+      if (_validateValues(stats.lvl)) {
```
crookedneighbor: 1) this method is called `_validateValues` but only validates a single value, level.

2) Wouldn't this be cleaner:

```
if (stats.lvl < 1) {
  Notification.error(env.t('invalidLevel'), true);
  return;
}

User.set({ /* stuff */ });
```

crookedneighbor:
TBH, we should be validating this on the server if we want to prevent invalid values: https://github.com/HabitRPG/habitica/blob/164177f010a9791046d300f5053ddae664bcf251/website/server/models/user/schema.js#L493-L497

cailinglu:
Should I also be editing code on the server side then?

crookedneighbor:
If the problem you're trying to solve is getting the user into a bad state, then yes, I think we should be validating on the server since a user could use the API to make this same call and put themselves in a bad state.

crookedneighbor:
`getting the user into a bad state` -> `preventing the user from getting into a bad state`

cailinglu:
I looked at the schema linked and just added a `min: 1` for the user's level. Not 100% sure if I should be doing more.

TheHollidayInn:
Looks good to me

website/server/models/user/schema.js
```diff
@@ -494,7 +494,7 @@ let schema = new Schema({
     mp: {type: Number, default: 10},
     exp: {type: Number, default: 0},
     gp: {type: Number, default: 0},
-    lvl: {type: Number, default: 1},
+    lvl: {type: Number, default: 1, min: 1},
```
paglias, line 497: should we run a migration for users who have it set to a value < 1?

Alys: No need in my opinion. The site behaves normally with negative levels as far as I can tell from a quick test, and even if there were problems, there'd be no need for a migration before merging this (and we'd probably already have heard from any users affected by it).

paglias:
Well with this change users with a level < 1 won't be able to use the site,
saving the user will fail. Hmm.. I think it depends on whether the
validation is done always or only when the field changes

Il 09 Apr 2017 21:20, "Alys" <notifications@github.com> ha scritto:

Alys:
Fair point. I'll do a search for users with level < 1 now and another after we merge. I'd expect the numbers to be so low it's not worth writing a dedicated migration script.

Alys:
I've tested locally as listed below and it's good.  I have not reviewed the code.

1. Used FCV to set level to 0 - failed: level was not changed and message appeared "Invalid value: Level must be 1 or greater."
1. Used FCV to set level to -1 - failed: as above
1. Used FCV to set level to -10 - failed: as above
1. Used FCV to set level to 1 - succeeded: level was set correctly
1. Used API (example call below) to set level to -1 - failed: level was not changed and the error shown below was given
1. Used API to set level to 0 - failed: as above
1. Used API to set level to 1 - succeeded: level was set correctly

curl -X PUT -H "x-api-user: d2c49eab-dae4-4ffb-9d40-cccb4d793cdf" -H "x-api-key: b6b6b446-d5eb-468e-a467-d640118c9661" http://localhost:3001/api/v3/user -d "stats.lvl=-1"  

{"success":false,"error":"BadRequest","message":"User validation failed","errors":[{"message":"Path `stats.lvl` (-1) is less than minimum allowed value (1).","path":"stats.lvl","value":-1}]}

test/client-old/spec/controllers/settingsCtrlSpec.js
```diff
@@ -334,4 +336,42 @@ describe('Settings Controller', function () {
...
+            exp: 0,
+            gp: 0,
+            lvl: 0,
+            mp: 0,
+          },
+          achievements: {
+            streak: 0,
+          },
+        };
+        scope.restore();
```
TheHollidayInn: I'm not sure this is correct. You have stubbed restore above, so the function won't be called, thus, you expectation below will never be true leading to this test not doing anything.

TheHollidayInn:
Noted towards you tier! Thanks, @caillu !

PR #8542, opened 2017-03-08
Client: Guilds: infinite loading

paglias:
WIP

Instead of loading all public guilds at once we only load a certain number (30, 50) and load the others only when the user scroll down.

Issues it introduces:
- Caching of public guilds on the client becomes more difficult
- Filtering and searching will have to be done on the server

TODO
- [x] stop loading if we reached the end


paglias:
@TheHollidayInn @SabreCat what do you think about the apiMessages change for API messages in english only?

Issue #8497, opened 2017-02-20
API Issue: release of pets/mounts can give undue achievements


madpink:
[//]: # (Fill out relevant information - UUID is found in Settings -> API)
General Info
  * UUID:  b0413351-405f-416f-8787-947ec1c85199

### Description
[//]: # (Describe bug in detail here. Include pictures if helpful.)
Sending release to pets, mounts or both returns a success and grants achievements regardless of whether the user earned it. That is, even if the user has gotten 0 pets, the call will return a success and give the "Beast Master" achievement.

Sending "release-both" will give the user all 3 achievements regardless of how many pets and or mounts they've gotten.

rakesh-tirupathi:
Hi guys! I would like to contribute to this. I am new to this, could any one please suggest how to get started and what needs to be done.
Thanks.

Alys:
@rakesh-tirupathi Thank you!

Have a look at http://habitica.wikia.com/wiki/Guidance_for_Blacksmiths for an introduction to coding for the website. It links off to a Setting Up Habitica Locally page which will help you create a local install for development and testing.  Post to the Blacksmith's guild if you need help (the wiki pages contain links).

After that it's just a matter of finding the code that releases the pets and mounts (https://habitica.com/apidoc/ will show you the command that's used and you can grep for that command in the code base) and then working out how the code needs to be changed. Let us know when you've got the local install working and are ready to start fixing this issue and then we'll change the issue's label to in progress.

jerell-mendoza:
Hi @Alys, I've gotten the local instance set up and running after following the guide. Can I contribute to this issue? It'll be my first!

Alys:
@rakesh-tirupathi Have you made a start on any code for this? If not, that's fine! We understand about real life getting in the way of coding. :) Let us know if you're not working on it yet and if you're not, @jerellmendoodoo can take it over.

rakesh-tirupathi:
I am sorry @Alys, I am having troubles having the environment set up. ya
sure @jerellmendoodoo <https://github.com/jerellmendoodoo> can take over
this issue. I will let you know after setting up the local instance and i
will take up other tasks.

Thank you

jerell-mendoza:
Hi, just as an update, I've fixed the issue and tested it locally through the UI and it appears to be working. Edited the releaseBoth, releaseMounts, releasePets .js files. I've read a bit through the wiki and I see that it's recommended that I write some unit tests regarding this? Is that a good next step?

[...]

PR #8545, opened 2017-03-09
Fixed release pets mounts

jerell-mendoza:

[//]: # https://github.com/HabitRPG/habitica/issues/8497#issuecomment-282994335

### Changes
[//]: # The current release pets, release mounts, and release both functions award achievements regardless of whether the user has any pets or mounts. There is now checking for each pet and mount before giving any of the achievements.

[//]: # 4b9dc32d-ad7d-4d2e-aaad-0543472f87f6

----
UUID:

### DIFF
--- a/website/common/script/ops/releaseBoth.js
+++ b/website/common/script/ops/releaseBoth.js
@@ -14,6 +14,8 @@ module.exports = function releaseBoth (user, req = {}, analytics) {
   }
 
   let giveTriadBingo = true;
+  let giveBeastMasterAchievement = true;
+  let giveMountMasterAchievement = true;
 
   if (!user.achievements.triadBingo) {
     if (analytics) {
@@ -45,20 +47,30 @@ module.exports = function releaseBoth (user, req = {}, analytics) {
     if (user.items.pets[animal] === -1) {
       giveTriadBingo = false;
     }
+    else if(user.items.pets[animal] === 0 || user.items.pets[animal] === null) {
+      giveBeastMasterAchievement = false;
+    }
+    if(user.items.mounts[animal] === null) {
+      giveMountMasterAchievement = false;
+    }
 
     user.items.pets[animal] = 0;
     user.items.mounts[animal] = null;
   }
 
-  if (!user.achievements.beastMasterCount) {
-    user.achievements.beastMasterCount = 0;
+  if(giveBeastMasterAchievement) {
+    if (!user.achievements.beastMasterCount) {
+      user.achievements.beastMasterCount = 0;
+    }
+    user.achievements.beastMasterCount++;
   }
-  user.achievements.beastMasterCount++;
 
-  if (!user.achievements.mountMasterCount) {
-    user.achievements.mountMasterCount = 0;
+  if(giveMountMasterAchievement) {
+    if (!user.achievements.mountMasterCount) {
+      user.achievements.mountMasterCount = 0;
+    }
+    user.achievements.mountMasterCount++;
   }
-  user.achievements.mountMasterCount++;
 
   if (giveTriadBingo) {
     if (!user.achievements.triadBingoCount) {
--- a/website/common/script/ops/releaseBoth.js
+++ b/website/common/script/ops/releaseBoth.js
@@ -46,11 +46,10 @@ module.exports = function releaseBoth (user, req = {}, analytics) {
   for (animal in content.pets) {
     if (user.items.pets[animal] === -1) {
       giveTriadBingo = false;
-    }
-    else if(user.items.pets[animal] === 0 || user.items.pets[animal] === null) {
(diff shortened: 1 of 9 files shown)


paglias:
This LGTM! @jerellmendoodoo we've switched to a new merging and testing process where PRs get merged and tested every two weeks, so it won't be merged immediately but in a few days. When it's merged I'll also note the contribution in your Habitica profile. Thanks for your help!

jerell-mendoza:
@paglias Yay! Thanks for your help with getting this in!

website/common/script/ops/releaseBoth.js
```diff
@@ -44,21 +46,30 @@ module.exports = function releaseBoth (user, req = {}, analytics) {
   for (animal in content.pets) {
     if (user.items.pets[animal] === -1) {
       giveTriadBingo = false;
+    } else if (user.items.pets[animal] === 0 || user.items.pets[animal] === null) {
```
crookedneighbor: Should this account for `undefined` as well?

jerell-mendoza: Hi, I'm not 100% sure, maybe @paglias or someone else would have some input on this? My thoughts are that we shouldn't include undefined, because we only want to check the pets that exist within the user.items.pets object. If a pet is missing from the object then it'd fail if we checked for undefined. Also, the only valid values should be -1, 0, null or positive int. 

Lastly, I did try it before using else if(!user.items.pets[animal]) - which checks for null, 0 or undefined - but this ended up failing previously written unit tests for releaseBoth.js - which is also another reason I didn't check for undefined

paglias: which tests failed when checking with `if(!user.items.pets[animal])`?

Now that I think about it, by default pets not owned are not stored at all inside `user.items.pets`, so the only way to check for them would be with `undefined`...

jerell-mendoza:
If you take a look at the releasePets.js and releaseBoth.js test files in
the test/common/ops folder, they're both initialized with just one pet.
We'd have to make sure that the test initialization is done correctly with
the object fully filled out with all the animals.

One of the tests that fail is: 'incremenets beastMasterCount' in the
releaseBoth.js tests.

Let me know what you'd want to do. We can modify the releasePets and
releaseBoth to include undefined checking, but then we'd have to modify the
tests accordingly. Perhaps on constructing the tests we could loop through
the content.pets object and add every pet to the users.items.pets object.
This way we wouldn't have a hard coded users.items.pets obj in the tests
(this could lead to failing tests in the future if any new pets are
added/removed).

Let me know what you think. Thanks!

paglias:
I see. The way the test is passing right now is because it's built to only consider one pet but it's not what we want. I think the best solution is to have the successful tests build a user.items.pets object with all the pets (built as in website/server/models/user/schema) with the right value. So for example to simulate the beast master achievement it should have all the pets set to -1 or > 0.

Does it make sense?

jerell-mendoza:
Hi @paglias, sorry it's been a while, but yeah that makes sense. I can go ahead and fix those tests and then change the releaseBoth and releasePets to use `if(!user.items.pets[animal])`

jerell-mendoza:
Hi @paglias, I've fixed this, but now I'm wondering how I should commit this to the same branch..?

paglias:
If the commit is in fix-release-pets-mounts just push it to GitHub and it'll appear here. Git push origin jerellmendoodoo:fix-release-pets-mounts

paglias:
If the commit is in fix-release-pets-mounts just push it to GitHub and it'll appear here. Git push origin fix-release-pets-mounts

jerell-mendoza:
Hi @paglias, I think I finally figured out how to fix it and also how to push to the branch, thanks for the help! Let me know if you have any feedback

paglias:
@jerellmendoodoo thanks, it looks good but some integration tests including `POST /user/release-both grants triad bingo with gems:`are failing

jerell-mendoza:
Hi, I think I'll have to update those tests as well. I wasn't aware of
those ones.

I'll take a look and do another commit with those updated.

Thanks!

SabreCat:
@jerellmendoodoo, this PR has been awaiting your updates for a while. Are you still planning to address the comments?

jerell-mendoza:
Hi, @SabreCat  - totally sorry lost track of this issue. Let me try to clean this up today, I think I just need to make sure the tests are passing

jerell-mendoza:
Thanks @SabreCat , let me know if there are any issues.

SabreCat (approved):

SabreCat:
Thank you for this, @jerellmendoodoo! I've awarded you Blacksmith tier 1 for your efforts. Note that subsequent tiers require increasing amounts of effort from one to the next, so tier 2 may require multiple PRs or a larger PR to achieve. But keep helping out and we will express our gratitude accordingly!

PR #8555, opened 2017-03-10
Group migration fixes

SabreCat:

### Changes

* Changes the migration used to award individual subscriptions to users in group plans to use a cursor-based stream, to avoid Mongo timeouts.
* Adds more error catches in payment libraries, so cancel operations don't fail out if the subscription is already gone on the payment processor end.
* Adds a migration to award Jackalope mounts to users who ought to qualify for them but are skipped over in current subscription code thanks to their having lifetime subscriptions.

### DIFF
--- a/website/server/libs/amazonPayments.js
+++ b/website/server/libs/amazonPayments.js
@@ -196,11 +196,11 @@ api.cancelSubscription = async function cancelSubscription (options = {}) {
 
   let details = await this.getBillingAgreementDetails({
     AmazonBillingAgreementId: billingAgreementId,
-  }).catch(function (err) {
-    console.log(err);
+  }).catch(function errorCatch (err) {
+    return err;
   });
 
-  let badBAStates = ['Canceled', 'Closed', 'Suspended']
+  let badBAStates = ['Canceled', 'Closed', 'Suspended'];
   if (details && details.BillingAgreementDetails && details.BillingAgreementDetails.BillingAgreementStatus &&
       badBAStates.indexOf(details.BillingAgreementDetails.BillingAgreementStatus.State) === -1) {
     await this.closeBillingAgreement({
--- a/website/server/libs/amazonPayments.js
+++ b/website/server/libs/amazonPayments.js
@@ -196,13 +196,18 @@ api.cancelSubscription = async function cancelSubscription (options = {}) {
 
   let details = await this.getBillingAgreementDetails({
     AmazonBillingAgreementId: billingAgreementId,
+  }).catch(function (err) {
+    console.log(err);
   });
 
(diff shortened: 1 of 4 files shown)


website/server/libs/stripePayments.js
```diff
@@ -226,20 +227,27 @@ api.cancelSubscription = async function cancelSubscription (options, stripeInc)
   if (!customerId) throw new NotAuthorized(i18n.t('missingSubscription'));
 
   // @TODO: Handle error response
-  let customer = await stripeApi.customers.retrieve(customerId);
+  let customer = await stripeApi.customers.retrieve(customerId).catch(function errorCatch (err) {
+    return err;
```
paglias, line 231: why this catch? it basically swallows the error, without it it would have bubbled up and returned an api error. This way it's ignored

website/server/libs/amazonPayments.js
```diff
@@ -196,14 +196,19 @@ api.cancelSubscription = async function cancelSubscription (options = {}) {
 
   let details = await this.getBillingAgreementDetails({
     AmazonBillingAgreementId: billingAgreementId,
+  }).catch(function errorCatch (err) {
+    return err;
```
paglias, line 200: same here, the error gets ignored

website/server/libs/stripePayments.js
SabreCat: I'd like to log it somewhere, but do we have a good way of doing that at present? (console logging is bad for Lint)

By catching the error, though, if the customer ID doesn't exist on the payment processor end--such as when the subscription has been remotely canceled due to failed payments--we will go ahead and cancel the subscription on the Habitica side. That fixes users seeing a bare API error when they try to cancel a dead Stripe sub, and helps us prune unpaid-for subscriptions pending better Webhook handling. @TheHollidayInn may have further insight!

website/server/libs/amazonPayments.js
TheHollidayInn: Do we have a log library?

website/server/libs/stripePayments.js
TheHollidayInn: Canceing and skipping this error is fine. We just need to log in a library.

TheHollidayInn: Winston is pretty good for example.

website/server/libs/amazonPayments.js
paglias: yes, libs/logger and it uses winston

website/server/libs/stripePayments.js
paglias: okay! We can use the logger in libs/logger to log it

PR #8558, opened 2017-03-11
Attempt to fix achievements awarding

paglias:
@Alys @SabreCat @TheHollidayInn

### DIFF
--- a/website/server/models/user/hooks.js
+++ b/website/server/models/user/hooks.js
@@ -153,31 +153,42 @@ schema.pre('validate', function preValidateUser (next) {
 schema.pre('save', true, function preSaveUser (next, done) {
   next();
 
-  if (_.isNaN(this.preferences.dayStart) || this.preferences.dayStart < 0 || this.preferences.dayStart > 23) {
-    this.preferences.dayStart = 0;
-  }
+  // VERY IMPORTANT NOTE: when only some fields from an user document are selected
+  // using `.select('field1 field2')` when the user is saved we must make sure that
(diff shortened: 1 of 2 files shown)


website/server/models/user/hooks.js
```diff
@@ -153,31 +153,51 @@ schema.pre('validate', function preValidateUser (next) {
...
+  // document.isSelected('items') will return true even if only a sub field (like 'items.mounts')
+  // was selected. So this fix only works as long as the entire subdoc is selected
+  // For example in the code below it won't work if only `achievements.beastMasterCount` is selected
+  // which is why we should only ever select the full paths and not subdocs,
+  // or if we really have to do the document.isSelected() calls should check for
+  // every specific subpath (items.mounts, items.pets, ...) but it's better to avoid it
+  // since it'll break as soon as a new field is added to the schema but not here.
+
+  // do not calculate achievements if items or achievements are not selected
+  if (!this.isSelected('items') || !this.isSelected('achievements')) {
```
TheHollidayInn: So when we are in this block with items selected but achievements not select, the achievements can be overwritten?

paglias: I think so: the first time achievements are accessed if they're not selected it creates a new achievements object with all the default values, so for example achievements.quests has a default of `{}` that will override the saved object. 

But if we could verify that this fix works it would be better. I'm pretty sure it won't create bugs as it's simply limits when the code can run

TheHollidayInn: A unit test around this could help. I'm still having trouble getting this to recreate locally.

Issue #8432, opened 2017-01-20
Check-in Incentives progress bar inconsistency


SabreCat:
[//]: # (Before logging this issue, look through common problems at https://github.com/HabitRPG/habitrpg/issues If you find your issue there, read at least the first post to see if there is a workaround for you)

[//]: # (Github is primarily used for reporting bugs. If you have a feature request, use "Help > Request a Feature" so that the feature request can be vetted by the larger Habitica community)

[//]: # (To report a bug in one of the mobile apps, please report it in the correct repository. Android: https://github.com/HabitRPG/habitrpg-android, iOS: https://github.com/HabitRPG/habitrpg-ios)

[//]: # (For more guidelines see https://github.com/HabitRPG/habitrpg/issues/2760)

[//]: # (Fill out relevant information - UUID is found in Settings -> API)
General Info
  * UUID: N/A
  * Browser: Chrome
  * OS: Windows 10

### Description
[//]: # (Describe bug in detail here. Include pictures if helpful.)

When you're in a stretch of Check-in Incentives that reward every 5 days, you'll first see "Progress until next 1/5" on the Profile page, then 2/5 and so on. Once you hit 5 and earn a reward, it'll show "Progress until next 5/10". The next day it'll be 1/5 again.

Reward days should probably show 0/5 instead of 5/10, for a sensible progression.

TheHollidayInn:
This is the way I original displayed it, but was asked to change.

SabreCat:
That's really odd. I understand why dropping to 0 could be off-putting, but having it be out of 10 one day and out of 5 the next is at least as puzzling as that.

eastwood:
I'm happy to fix this for the team. Let me know if this is okay!

SabreCat:
@eastwood: Go for it, with our thanks! Let us know if you need any help as you go.

eastwood:
PR should be done pending review :)

librarianmage:
The issue has been fixed; should this Issue be closed?

PR #8559, opened 2017-03-11
Issue 8432: Incentives Progress Fix

eastwood:
Issue #8432 

### Changes
The `loginIncentives.js` file describes a keyed map of rewards, detailing the `prevAward` and `nextAward`. Using the number of consecutive logins as the key, the map was able to be searched to determine *when* the next rewards would be and *when* the previous award was. 

On the profile page, if you had just received a reward, this calculation did not take into consideration if the current login was issuing a reward; therefore it was displaying values of `5/10` instead of `0/5`. 

While I believe there could be a better data structure to store the rewards, for now the fix, adds a conditional to ensure that if the given login received an award, then it is accounted for during the progress component.

The fix also fixed up typos found during investigation.

----
UUID: 4df45790-08ba-4e11-8fcd-0d56c3343cd1

### DIFF
--- a/test/client-old/spec/controllers/userCtrlSpec.js
+++ b/test/client-old/spec/controllers/userCtrlSpec.js
@@ -1,0 +1,37 @@
+'use strict';
+
+describe('User Controller', function() {
+  var $rootScope, shared, scope, user, User, ctrl, content;
+
+  beforeEach(function() {
+    user = specHelper.newUser();
+    User = {
+      user: user
+    };
+
+    module(function($provide) {
+      $provide.value('User', User);
+      $provide.value('Guide', {});
+    });
+
+    inject(function($rootScope, $controller, Shared, Content){
+      scope = $rootScope.$new();
+      Shared.wrap(user);
+      shared = Shared;
+      content = Content
+      $controller('RootCtrl',  {$scope: scope, User: User, Shared: Shared});
+      ctrl = $controller('UserCtrl', {$scope: scope, User: User});
+    });
+  });
+
+  describe.only('getProgressDisplay', function() {
+    it('should return initial progress', function() {
+      sinon.stub(content, 'loginIncentives').onFirstCall().returns({
+        prevRewardKey: 0,
+        nextRewardKey: 1
+      });
+      var actual = scope.getProgressDisplay();
+      expect(actual).to.eql('');
+    });
+  });
+});
--- a/test/client-old/spec/controllers/userCtrlSpec.js
+++ b/test/client-old/spec/controllers/userCtrlSpec.js
@@ -27,30 +27,29 @@ describe.only('User Controller', function() {
     inject(function($rootScope, $controller, User, Content) {
       scope = $rootScope.$new();
       content = Content;
-      $window = {
-        env: {
-          t: sandbox.stub(),
-        },
-      };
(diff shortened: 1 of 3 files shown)


TheHollidayInn:
Thanks!

Would you mind adding a test around this while you are working on it.

Also, what is your suggestion for the new data structure?

eastwood:
No worries :) I've attempted to add a test in, but wow Angular 1 test DI is interesting... Would somebody mind giving a hand scaffolding the test harness? I'm having difficulty getting the right dependencies injected.

test/client-old/spec/controllers/userCtrlSpec.js
```diff
@@ -0,0 +1,55 @@
...
+        suppressModals: {}
+      },
+      purchased: {
+        plan: {
+          mysteryItems: [],
+        },
+      },
+    });
+
+    User = {
```
TheHollidayInn: You will want to provide this: https://github.com/HabitRPG/habitica/blob/develop/test/client-old/spec/controllers/authCtrlSpec.js#L7

test/client-old/spec/controllers/userCtrlSpec.js
```diff
@@ -0,0 +1,55 @@
...
+      user: user
+    };
+
+    inject(function($rootScope, $controller, Shared, User, Achievement, Guide) {
+      scope = $rootScope.$new();
+      Shared.wrap(user);
+      shared = Shared;
+      achievement = Achievement;
+      User.setUser(user);
```
TheHollidayInn: You shouldn't need this if you use provide.

TheHollidayInn (commented):
Sorry for the delay. You are close. I left a few comments to help.

eastwood:
Thanks for the help. That was enough for me to realise what method is for injecting into test and what's for hooking into the test.

test/client-old/spec/controllers/userCtrlSpec.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+describe.only('User Controller', function() {
```
crookedneighbor: Remove this only.

crookedneighbor (dismissed):

test/client-old/spec/controllers/userCtrlSpec.js
eastwood: Gets me everytime :) Thanks

TheHollidayInn: You can run `npm run lint` and/or add a js lint to your text editor

eastwood: I have linting support, but I believe these files are ignored:
```
//.eslintignore
# Not linted
website/client-old/
test/client-old/spec/**/*
```

TheHollidayInn: Ah, yes. My mistake :P

TheHollidayInn (approved):

paglias:
Merged for testing! @eastwood , thank you, and welcome to Blacksmith tier 1! Further tiers require increasing amounts of effort from one to the next, so Tier 2 will take multiple PRs or a larger PR. But keep helping out and we will express our gratitude accordingly!

PR #8563, opened 2017-03-12
use correct wording and date format for cancelled Group Plan in Payment Details screen

Alys:
Currently, when you have cancelled a group plan, the group's Payment Details screen shows a cancellation message that refers to subscriptions instead of group plans. Also, the date format is the USA format, regardless of the user's preferences.

image

This PR changes the wording to specify "Group Plan" instead of "Subscription" and uses the user's preferred date format:

image

Once this is reviewed, tested, and approved, if possible, I'd like it to be merged to the `release` branch and released fairly soon to avoid confusion (the group owner might think that only the free subscriptions had been cancelled, not the group plan that they tried to cancel). It's a small change with a low chance of errors that can't be picked up by testing.


TheHollidayInn:
Looks good to me

PR #8564, opened 2017-03-12
add new feature descriptions to Social > Group Plans page

Alys:
### Changes

This adjusts the Social > Group Plans page to include two new points for the new Group Plans features (free subscriptions and a Jackalope mount, seen at the bottom of the screenshot below). It also removes Windows line breaks at the end of one of the files.

image


paglias:
Can we also show the price?

Alys:
@lemoness would you like to give me words for the price too?

lemoness:
It should be the same as what displays later: 

$9 every month + $3 a month for every additional group member

Alys:
I've added the price under the button:

image

PR #8565, opened 2017-03-13
Set _cronSignature to current time instead of uuid

### LINKED ISSUE #8163: cron does not run when _cronSignature is permanently not set to NOT_RUNNING

KyleVerma:

Fixes #8163

### Changes

_cronSignature now set to now.getTime() instead of uuid()

created const CRON_TIMEOUT, which is set to 1hour (in ms) can be changed easily later

where the code used to make sure _cronSignature was 'NOT_RUNNING' is now an $or that also checks to see if _cronSignature is cronTimeout ms smaller than the current time

All _cronSignature time calculations are in ms since epoch time to hopefully avoid issues with timezones and end of day

----
UUID: 0036df29-70e5-4700-8e38-d7dd964a6401


website/server/middlewares/cron.js
```diff
@@ -6,9 +6,10 @@ import Bluebird from 'bluebird';
 import { model as Group } from '../models/group';
 import { model as User } from '../models/user';
 import { recoverCron, cron } from '../libs/cron';
-import { v4 as uuid } from 'uuid';
 
 const daysSince = common.daysSince;
+// Wait this length of time in ms before attempting another cron
+const CRON_TIMEOUT = new Date(60 * 60 * 1000).getTime();
```
khipkin: Suggestion: rename variable to something more descriptive.

Ideas: CRON_TIMEOUT_WINDOW, CRON_TIMEOUT_MS, CRON_TIMEOUT_WAIT, CRON_TIMEOUT_MAX_WAIT, etc.

website/server/middlewares/cron.js
```diff
@@ -101,12 +102,18 @@ async function cronAsync (req, res) {
       return null;
     }
 
-    let _cronSignature = uuid();
+    // set _cronSignature to current time in ms since epoch time so we can make sure to wait at least an hour before trying again even if there is a timezoned
```
khipkin: "even if there is a timezoned"?

Also, I suggest making this comment agnostic of the specific wait time (1 hr currently).

website/server/middlewares/cron.js
```diff
@@ -101,12 +102,18 @@ async function cronAsync (req, res) {
...
+    let _cronSignature = now.getTime();
+    // Calculate how long ago cron must have been attempted to try again
+    let timeoutThreshold = _cronSignature - CRON_TIMEOUT;
 
     // To avoid double cron we first set _cronSignature and then check that it's not changed while processing
     let userUpdateResult = await User.update({
       _id: user._id,
-      _cronSignature: 'NOT_RUNNING', // Check that in the meantime another cron has not started
+      $or: [ // Make sure cron was successful or failed at least CRON_TIMEOUT ms ago
```
khipkin: Comment says CRON_TIMEOUT, but timeoutThreshold is actually used.

website/server/middlewares/cron.js
```diff
@@ -101,12 +102,18 @@ async function cronAsync (req, res) {
       return null;
     }
 
-    let _cronSignature = uuid();
+    // set _cronSignature to current time in ms since epoch time so we can make sure to wait at least an hour before trying again even if there is a timezoned
+    let _cronSignature = now.getTime();
+    // Calculate how long ago cron must have been attempted to try again
+    let timeoutThreshold = _cronSignature - CRON_TIMEOUT;
```
khipkin: Suggestion: rename variable to something more descriptive. Can't think of any suggestions right now.

website/server/middlewares/cron.js
KyleVerma: Gonna go with CRON_TIMEOUT_WAIT if nobody has a better suggestion, it seems most accurate in my mind to what the constant is actually storing

website/server/middlewares/cron.js
KyleVerma: how does cronRetryTime sound to you as a more accurate/descriptive variable name?

khipkin: @KyleVerma Ohh that sounds perfect!

KyleVerma:
Am I supposed to change the stub for the failing test or is changing tests handled by admins?

I can't figure out how to change stubs in sinon.js in a way that will work nicely with the new update call so if I'm supposed to change it can someone point me in the direction of something that will be more helpful than sinon.js documentation? Thanks.

paglias:
@KyleVerma The test is failing because the stub uses `.withArgs()` https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/unit/middlewares/cronMiddleware.js#L238 which means it'll return the value in `.returns()` only when the arguments passed to `User.update` match the ones in `.withArgs`. To fix it you should make sure `withArgs` is passed the new arguments that are now passed to User.update() which are now

```
       _id: user._id,
       _cronSignature: 'NOT_RUNNING', // Check that in the meantime another cron has not started
       $or: [ // Make sure last cron was successful or failed before cronRetryTime
         {_cronSignature: 'NOT_RUNNING'},
         {_cronSignature: {$lt: cronRetryTime}},
       ],
```

KyleVerma:
Sorry if I'm being dense but how do I handle cronRetryTime being a variable? Is there somewhere else where the stub is being called that I can't find where I can put in 'fake' times or something?

paglias:
@KyleVerma no worries. You can use sinon.js matchers http://sinonjs.org/releases/v2.0.0/matchers/
that make it possible to match a value without knowing it completely.

Something like this could work

```
withArgs({
       _id: user._id,
       $or: [ // Make sure last cron was successful or failed before cronRetryTime
         {_cronSignature: 'NOT_RUNNING'},
         {_cronSignature: {$lt: sinon.match.number}},
       ],
});
```

KyleVerma:
Worked like a charm thanks!

Should a test be created for the new functionality?

I was thinking something along the lines of:
- Simulate a failed cron; either just set _cronSignature to some time or force cron to timeout somehow
- Try an update call less than CRON_TIMEOUT_WAIT after the failure
- Try an update call more than CRON_TIMEOUT_WAIT after the failure (unless we want tests that take an hour I'm guessing with fake timestamps)

paglias:
I think it's enough to have these two tests:

- Set cron_signature (using User.update()) to more than 1 hour ago and verify that calling `cron()` re-execute it
- Set cron to 59 minutes ago and verify that calling cron() doesn't re-execute it

KyleVerma:
A few questions concerning the commit I just pushed:

- What is the naming convention for new unit tests?
- What are the standards/guidelines for creating/throwing new errors in unit tests? (concerning the first test I added)
- Is there a nicer way to handle setting cronSignature to 1 hour ago? (Ideally it would somehow reference the variable in cron.js so if it's changed there these tests don't also need to be changed)
- In the first test should I check the error message to verify that cron is failing for the right reason? Or check something else before resolving?

Alys:
@KyleVerma I'm really sorry, we never answered your questions. :( Do you still need answers for any/all of them? Or is this ready for us to review?

KyleVerma:
I think I have tests that correctly test the functionality added, so a code review couldn't hurt. However my questions are still potentially necessary since my tests might not be up to standards without them.

SabreCat:
Naming convention: The naming convention for unit tests is pretty loose vs. the more regimented one we have for integration tests. You could name this one with `.test.js` instead if you'd like, but this is fine.

Creating/throwing errors: That's interesting--I don't know that we've structured any other tests this way (without an explicit `expect`). @TheHollidayInn or @paglias, any thoughts?

Constants: We would typically put magic numbers like that in a constant at the top of the test file and reference it that way (so at least it's clear where you need to update if the functionality does change). The alternative would be to export a constant somewhere and import it into both the middleware and the test, but I don't know that we use that pattern outside of `/website/common`.

Check error message: Ideally we should check that the right error is thrown, yes. And that could be an answer for the second question--go ahead and set up the test with an `expect` that the cron will error out a specific way.

TheHollidayInn:
You should be able to `await` and expectation which will resolve promises. Here is an example: https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/unit/libs/applePayments.test.js#L52

SabreCat:
@KyleVerma, are you still working on this? If not, let us know and we can open this up for someone else to tackle.

KyleVerma:
So sorry for the delay I had some work deadlines and an currently on vacation. Should be able to wrap this up when I get home.

SabreCat:
@KyleVerma New ping. Back from vacation yet?

KyleVerma:
Yes I just got back recently and was planning to do some work on this today.

SabreCat:
Some linting errors, shouldn't be too hard to fix!

KyleVerma:
Gonna try and get to them by the end of the day :)

Is it a problem that one of the tests I added takes 1.5s to run?

KyleVerma:
Two tests are failing due to timeouts now. Did I accidentally cause this? I'll look into it further soon but didn't want to waste time if someone could easily look at the tests and determine that I'm not the culprit. Thanks.

Alys:
@KyleVerma I've restarted the tests. If it times out again we'll look into it in more detail, but from a quick glance it didn't seem to be related to your code.

KyleVerma:
Is there anything that still needs to be done on my end for this issue?

SabreCat:
No, we mainly just set this aside because we've been prioritizing PRs related to the recent frontend redesign! We'll take a look at this soon.

KyleVerma:
Makes sense

SabreCat:
@KyleVerma Thank you for your effort (and patience!). I've awarded you your first contributor tier--welcome to the ranks of the Blacksmiths!

KyleVerma:
Awesome, Thanks!

PR #8572, opened 2017-03-15
Enabled repeatables

### LINKED ISSUE #4173: Repeat Every X Days (Monthlies) [$140]
This is a Github ticket for this much-requested Trello card feature so that people can put Bounties on it!

From card:
Tasks that repeat every "n" days

Examples:
Every 5 days after last Clean Bathroom
On Aug 1st every year, Send birthday card
Every 2 weeks ...
Every 3 days ...

The original card is here: https://trello.com/c/XuVokK2V/108-repeat-every-days-weeklies-monthlies
## <bountysource-plugin>

There is a **[$140 open bounty](https://www.bountysource.com/issues/5413688-repeat-every-x-days-monthlies-125?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheHollidayInn:

Fixes #4173

### Changes

----
UUID:


Alys:
I'd like us to take a very cautious approach to testing and merging this. We had bugs last time that weren't picked up by our automated tests, so I'd like to see the tests improved first, and then this merged after that.

The bugs were from people who had non-zero Custom Day Start times and were viewing their Dailies after midnight but before the CDS time, so I suspect that the tests don't cover that case. There might be other cases too. I'll try to make time this weekend to think about all the cases that might need to be tested and see if we have tests for them.

TheHollidayInn:
Yea, this is because we had no automated tests before - so refactoring becomes an issue. We will have to try and think of all the manual tests/features.

TheHollidayInn:
Yea, this is because we had no automated tests before - so refactoring becomes an issue. We will have to try and think of all the manual tests/features.

website/common/script/cron.js
```diff
@@ -4,8 +4,10 @@
   Cron and time / day functions
   ------------------------------------------------------
  */
-import _ from 'lodash'; // eslint-disable-line lodash/import-scope
```
paglias, line 7: I hadn't noticed this but we should really avoid importing the entire lodash in common or in client. Adding one exception defats the entire purpose of only loading individual utilities

paglias:
This need to wait for the isNextDue PR for the mobile apps anyway?

website/common/script/cron.js
TheHollidayInn: I'll update it! This was before I saw the lodash update.

TheHollidayInn:
The UI needs to wait, but nextDue depends on this shouldDo

Alys:
I've listed all the `shouldDo` tests that I think are necessary for the current behaviour at https://github.com/HabitRPG/habitica/issues/8585

paglias:
This is waiting on some mobile changes ans the isDue PR right?

Alys:
Also waiting on new tests for the current repeat behaviour. Our tests weren't sufficient to pick up a bug when this first went live.

website/common/script/cron.js
```diff
@@ -93,32 +98,70 @@ export function shouldDo (day, dailyTask, options = {}) {
...
-    return false; // error condition - unexpected frequency string
+    schedule = schedule.every(daysOfTheWeek).daysOfWeek();
+
+    if (options.nextDue) return schedule.next(3, 'L');
+
+    return schedule.matches(startOfDayWithCDSTime);
+  } else if (dailyTask.frequency === 'monthly') {
+    let schedule = moment(startDate).recur();
+
+    let differenceInMonths = moment(day).month() + 1 - moment(startDate).month() + 1;
```
paglias: why plus 2 months here?

website/server/controllers/api-v3/tasks.js
```diff
@@ -457,7 +458,13 @@ api.updateTask = {
     }
 
     if (sanitizedObj.type === 'daily') {
-      task.isDue = common.shouldDo(Date.now(), sanitizedObj, user.preferences);
+      let optionsForShouldDo = cloneDeep(user.preferences.toObject());
```
paglias: toObject already clones objects so calling deepClone is not necessary, see https://github.com/Automattic/mongoose/issues/5008

paglias (changes requested):

website/common/script/cron.js
TheHollidayInn: Yea, I guess we don't need this since I am adding it to both sides :P

website/server/controllers/api-v3/tasks.js
TheHollidayInn: Ah, nice.

SabreCat:
Non-blocking question. There are two more Daily features we would eventually like, to finish out the use cases:

* "Active until completed" option. Example: "Pay rent" comes due on the first of each month. On June 1 the Daily is active. If I check it off, it will be active again July 1. But if I _don't_ complete it, it's still active on June 2, 3, etc. Once I get around to taking care of it, it'll go inactive until July 1.
* "X times in Y interval" method. For things like "work out three times a week" or "call parents twice a month" or whatever. This would necessitate a distinction between Dailies _appearing active_ vs. _causing penalties when missed_; if there's still enough days left in the interval to satisfy the condition, it doesn't hurt you. But if you get to the point where there are X or fewer days left in the interval, missing it kicks in the usual damage, degrades the Daily's color, and so on.

These features needn't be in scope for this. But how much refactoring would be necessary to incorporate them later? Would the Repeatables code here need to be substantially rewritten to accommodate the above, or could we extend/tweak the logic without too much waste?

SabreCat:
Issue: the summary line "Repeats [frequency] every [X] [intervals]" is hardcoded in English. Should be an I18N string with variables.

https://github.com/TheHollidayInn/habitrpg/blob/enable-repeatables/website/client-old/js/services/taskServices.js#L355

SabreCat:
Non-blocking existing issue we could maybe clean up here: Every X can go below 1. Should stop at 1.

PR #8588, opened 2017-03-20
Added end date option to group plan migration

TheHollidayInn:

Fixes put_issue_url_here

### Changes

This allows for the addition of an end date when giving a free group plan.

----
UUID:


Alys:
As shown in the screenshot below, there should be more space to the left of the "Remove" button and less space before the managers' names in that section. There should also be more space to the left of the "Add Manager" button.

image

Also as shown there, the leader can add themself as manager. Is that intentional? On the one hand, it's undesirable because they leader might think they need to and that wastes their time. On the other hand, if the leader is about to transfer leadership to another member, they might want to make themself a manager first. However that would only happen after the group plan had been cancelled, and it should probably then be up to the new leader whether the old one is allowed to be a manager. I'm leaning towards making it not possible for the leader to declare themself a manger too.

When a task requiring approval is completed, only the group leader gets a notification that an approval is waiting; managers do not get approvals. If that's deliberate, that's fine but can you let us know if it's deliberate so that we can document it? If there's anything else non-obvious that you think should be documented about this, can you edit the top post to give details about the new features?

When a task awaiting approval is approved by a manager, the notification that approval is needed is not removed from the group leader's notification list. This is going to cause wasted time and annoyance for the leader. Possible solutions include:
- Give all the group plans notifications "x" buttons so that they can be more easily dismissed - if none of the other options are done, I think this will be essential to reduce the complaints. Even if one of the other options is done, this will still be a good idea now.
- Remove the notification when the task is approved by someone else.
- Don't have individual notifications for any tasks awaiting approval, but instead have one notification for each group plan stating that there is at least one task to be approved. As soon as all tasks are approved, that notification could be automatically removed (or at the very least could be dismissable with an "x" button).

If a member completes a task requiring approval and it's approved by both a manager and the leader, the member receives twice as much XP and GP for it (i.e., both approvals give rewards to the user). They also receive two notifications that it was approved. This can only happen if the second approver had not refreshed the list of approvals before approving, but that could happen in real life when two approvers happened to be working through the list of pending approvals at the same time.

Alys:
Ah, heck I'm sorry, that went into the wrong PR. :( I'm moving it now.

PR #8589, opened 2017-03-21
Limit party size to 30 members

### LINKED ISSUE #8561: limit party size to 30 members
Currently, a party can have an unlimited number of members. This isn't good for performance and it can also give players a bad experience of Habitica for reasons related to accountability and interpersonal interactions.

We want parties to be limited to 30 members. However we don't want to immediately disadvantage the parties that are currently larger, so the limitation will be that you cannot invite new members when you already have 30 or more.

Parties that already have more than 30 can keep all their members. If they choose to remove members, they cannot add more until their member count is lower than 30.

Some suggestions for the code:
- 30 should be stored in a variable so it's easy to change later
- The API must enforce the limit. The API route is [https://habitica.com/api/v3/groups/:groupId/invite](https://habitica.com/apidoc/#api-Group-InviteToGroup)
- The limit must apply to ONLY parties, not public or private guilds.
   - There must be tests to prove that:
      - the limit applies to parties
      - the limit does not apply to public guilds
      - the limit does not apply to private guilds. (Also any other tests that are necessary of course.)
- The party's `memberCount` attribute can be used to find the current size of the party. The `memberCount` is not always accurate (https://github.com/HabitRPG/habitica/issues/7506) but for the purposes of this feature it will be close enough. (We'll be fixing the inaccurate memberCounts eventually, and any players negatively affected by an inaccurate count can get it fixed for their party by reporting it in the Report a Bug guild.)
- The party page should have a sentence stating that once a party has 30 or more members, no more can be invited. It should probably be placed under the "Members" header and "Invite Friends" button: image
- When the memberCount is 30 or greater, the "Invite Friends" button should not appear. The message described just above should be changed to say that because the party has more than 30 members, you can't invite more. It should explain that when the party size is less than 30, invitations will be possible again.

Yutsuten:
_note from Alys to Alys: when notifying the wiki editors about this new feature, include this: https://github.com/HabitRPG/habitica/pull/8589#issuecomment-289715055_

----

Fixes #8561 

### Changes
To-do list:
- [x] Add a field that will contain the information about the current number of members in the party and the limit
- [x] Create a global variable with the maximum value of members in a party
- [x] Hide the button when the limit is reached
- [x] Enforce the limit in the server (API) considering:
  - [x] Current number of members
  - [x] Current pending invitations in the party
  - [x] Number of emails and UUIDs provided for invitations
- [x] Create tests to prove that the limit:
  - [x] Applies to parties
  - [x] Does not apply to guilds

### Current results
Party with 29 members, and trying to invite more 2:

Invite2

Result (notification with error):

Error

After inviting only one (button to invite is hidden):

Invited1

----

### DIFF
--- a/test/api/v3/integration/groups/POST-groups_invite.test.js
+++ b/test/api/v3/integration/groups/POST-groups_invite.test.js
@@ -331,7 +331,7 @@ describe('Post /groups/:groupId/invite', () => {
       let generatedInvites = await Promise.all(invitesToGenerate);
       // Invite users
       expect(await inviter.post(`/groups/${group._id}/invite`, {
-        uuids: generatedInvites.map(invite => invite._id)
+        uuids: generatedInvites.map(invite => invite._id),
       }))
       .to.be.an('array');
     });
@@ -435,7 +435,7 @@ describe('Post /groups/:groupId/invite', () => {
       let generatedInvites = await Promise.all(invitesToGenerate);
       // Invite users
       expect(await inviter.post(`/groups/${party._id}/invite`, {
-        uuids: generatedInvites.map(invite => invite._id)
+        uuids: generatedInvites.map(invite => invite._id),
       }))
       .to.be.an('array');
     });
@@ -449,7 +449,7 @@ describe('Post /groups/:groupId/invite', () => {
       let generatedInvites = await Promise.all(invitesToGenerate);
       // Invite users
       await expect(inviter.post(`/groups/${party._id}/invite`, {
-        uuids: generatedInvites.map(invite => invite._id)
+        uuids: generatedInvites.map(invite => invite._id),
       }))
       .to.eventually.be.rejected.and.eql({
         code: 400,
--- a/test/api/v3/integration/groups/POST-groups_invite.test.js
+++ b/test/api/v3/integration/groups/POST-groups_invite.test.js
@@ -322,6 +322,20 @@ describe('Post /groups/:groupId/invite', () => {
       });
     });
 
+    it('allow 30+ members in a guild', async () => {
+      let invitesToGenerate = [];
+      // Generate 30 users to invite (30 + leader = 31 members)
+      for (let i = 0; i < PARTY_LIMIT_MEMBERS; i++) {
+        invitesToGenerate.push(generateUser());
+      }
+      let generatedInvites = await Promise.all(invitesToGenerate);
+      // Invite users
+      expect(await inviter.post(`/groups/${group._id}/invite`, {
+        uuids: generatedInvites.map(invite => invite._id)
+      }))
+      .to.be.an('array');
+    });
+
     // @TODO: Add this after we are able to mock the group plan route
     xit('returns an error when a non-leader invites to a group plan', async () => {
       let userToInvite = await generateUser();
@@ -423,7 +437,7 @@ describe('Post /groups/:groupId/invite', () => {
       expect(await inviter.post(`/groups/${party._id}/invite`, {
         uuids: generatedInvites.map(invite => invite._id)
       }))
-      .to.be.a('array');
+      .to.be.an('array');
     });
 
     it('do not allow 30+ members in a party', async () => {
--- a/test/api/v3/integration/groups/POST-groups_invite.test.js
+++ b/test/api/v3/integration/groups/POST-groups_invite.test.js
@@ -420,15 +420,10 @@ describe('Post /groups/:groupId/invite', () => {
       }
       let generatedInvites = await Promise.all(invitesToGenerate);
       // Invite users
-      await inviter.post(`/groups/${party._id}/invite`, {
+      expect(await inviter.post(`/groups/${party._id}/invite`, {
         uuids: generatedInvites.map(invite => invite._id)
-      });
-      // Users accept invites
-      for (let i = 0; i < PARTY_LIMIT_MEMBERS - 1; i++) {
-        await generatedInvites[i].post(`/groups/${party._id}/join`);
-      }
-      // Verify the number of members
-      expect((await inviter.get(`/groups/${party._id}`)).memberCount).to.equal(PARTY_LIMIT_MEMBERS);
+      }))
+      .to.be.a('array');
     });
 
     it('do not allow 30+ members in a party', async () => {
@@ -438,7 +433,7 @@ describe('Post /groups/:groupId/invite', () => {
         invitesToGenerate.push(generateUser());
       }
       let generatedInvites = await Promise.all(invitesToGenerate);
-      // Invite users - must fail
+      // Invite users
       await expect(inviter.post(`/groups/${party._id}/invite`, {
         uuids: generatedInvites.map(invite => invite._id)
       }))
--- a/test/api/v3/integration/groups/POST-groups_invite.test.js
+++ b/test/api/v3/integration/groups/POST-groups_invite.test.js
@@ -430,5 +430,23 @@ describe('Post /groups/:groupId/invite', () => {
       // Verify the number of members
       expect((await inviter.get(`/groups/${party._id}`)).memberCount).to.equal(PARTY_LIMIT_MEMBERS);
     });
+
+    it('do not allow 30+ members in a party', async () => {
+      let invitesToGenerate = [];
+      // Generate 30 users to invite (30 + leader = 31 members)
+      for (let i = 0; i < PARTY_LIMIT_MEMBERS; i++) {
+        invitesToGenerate.push(generateUser());
+      }
+      let generatedInvites = await Promise.all(invitesToGenerate);
+      // Invite users - must fail
+      await expect(inviter.post(`/groups/${party._id}/invite`, {
+        uuids: generatedInvites.map(invite => invite._id)
+      }))
+      .to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: t('partyExceedsMembersLimit', {maxMembersParty: PARTY_LIMIT_MEMBERS}),
+      });
+    });
   });
 });
--- a/test/api/v3/integration/groups/POST-groups_invite.test.js
+++ b/test/api/v3/integration/groups/POST-groups_invite.test.js
@@ -322,7 +322,7 @@ describe('Post /groups/:groupId/invite', () => {
       });
     });
 
-    it('allow 30+ members in a guild', async () => {
+    it('allows 30+ members in a guild', async () => {
       let invitesToGenerate = [];
       // Generate 30 users to invite (30 + leader = 31 members)
       for (let i = 0; i < PARTY_LIMIT_MEMBERS; i++) {
@@ -332,8 +332,7 @@ describe('Post /groups/:groupId/invite', () => {
       // Invite users
       expect(await inviter.post(`/groups/${group._id}/invite`, {
         uuids: generatedInvites.map(invite => invite._id),
-      }))
-      .to.be.an('array');
+      })).to.be.an('array');
     });
 
     // @TODO: Add this after we are able to mock the group plan route
@@ -426,7 +425,7 @@ describe('Post /groups/:groupId/invite', () => {
       expect((await userToInvite.get('/user')).invitations.party.id).to.equal(party._id);
     });
 
-    it('allow 30 members in a party', async () => {
+    it('allows 30 members in a party', async () => {
       let invitesToGenerate = [];
       // Generate 29 users to invite (29 + leader = 30 members)
       for (let i = 0; i < PARTY_LIMIT_MEMBERS - 1; i++) {
@@ -436,11 +435,10 @@ describe('Post /groups/:groupId/invite', () => {
       // Invite users
       expect(await inviter.post(`/groups/${party._id}/invite`, {
         uuids: generatedInvites.map(invite => invite._id),
-      }))
-      .to.be.an('array');
+      })).to.be.an('array');
     });
 
-    it('do not allow 30+ members in a party', async () => {
+    it('does not allow 30+ members in a party', async () => {
       let invitesToGenerate = [];
       // Generate 30 users to invite (30 + leader = 31 members)
       for (let i = 0; i < PARTY_LIMIT_MEMBERS; i++) {
--- a/test/api/v3/integration/groups/POST-groups_invite.test.js
+++ b/test/api/v3/integration/groups/POST-groups_invite.test.js
@@ -6,6 +6,7 @@ import {
 import { v4 as generateUUID } from 'uuid';
 
 const INVITES_LIMIT = 100;
+const PARTY_LIMIT_MEMBERS = 30;
 
 describe('Post /groups/:groupId/invite', () => {
   let inviter;
@@ -409,6 +410,25 @@ describe('Post /groups/:groupId/invite', () => {
         uuids: [userToInvite._id],
       });
       expect((await userToInvite.get('/user')).invitations.party.id).to.equal(party._id);
+    });
+
+    it('allow 30 members in a party', async () => {
+      let invitesToGenerate = [];
+      // Generate 29 users to invite (29 + leader = 30 members)
+      for (let i = 0; i < PARTY_LIMIT_MEMBERS - 1; i++) {
+        invitesToGenerate.push(generateUser());
+      }
+      let generatedInvites = await Promise.all(invitesToGenerate);
+      // Invite users
+      await inviter.post(`/groups/${party._id}/invite`, {
+        uuids: generatedInvites.map(invite => invite._id)
+      });
+      // Users accept invites
+      for (let i = 0; i < PARTY_LIMIT_MEMBERS - 1; i++) {
+        await generatedInvites[i].post(`/groups/${party._id}/join`);
+      }
+      // Verify the number of members
+      expect((await inviter.get(`/groups/${party._id}`)).memberCount).to.equal(PARTY_LIMIT_MEMBERS);
     });
   });
 });
--- a/test/api/v3/unit/models/group.test.js
+++ b/test/api/v3/unit/models/group.test.js
@@ -537,7 +537,7 @@ describe('Group Model', () => {
         expect(res.t).to.be.calledWith('canOnlyInviteMaxInvites', {maxInvites: INVITES_LIMIT });
       });
 
-      it('does not throw error if number of invites matches max invite limit', () => {
+      it('does not throw error if number of invites matches max invite limit', async () => {
         let uuids = [];
         let emails = [];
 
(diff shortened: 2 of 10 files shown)


website/server/controllers/api-v3/groups.js
```diff
@@ -1066,6 +1067,31 @@ api.inviteToGroup = {
     let uuids = req.body.uuids;
     let emails = req.body.emails;
 
+    // If party, check the limit of members
+    if (group.type === 'party') {
+      let partyLimitMembers = 5;
```
paglias: I'd move this to a `const PARTY_LIMIT_MEMBERS` outside of the function

paglias: actually since the limit is used in the jade file below it should probably be put in common/script/constants so that it can be used in both the server and the client

paglias:
This will need tests but it looks okay overall.

website/server/controllers/api-v3/groups.js
```diff
@@ -1066,6 +1067,31 @@ api.inviteToGroup = {
...
+        .select('_id')
+        .exec();
+      memberCount += groupInvites.length;
+      // Counting the members that are going to be invited by email and uuids
+      if (uuids)
+        memberCount += uuids.length;
+      if (emails)
+        memberCount += emails.length;
+
+      if (memberCount > partyLimitMembers)
```
paglias: let's add parenthesis around the `if`, it matches our code style better

paglias (changes requested):

website/server/controllers/api-v3/groups.js
Yutsuten: Thank you for the feedback! I'll change that.

website/server/controllers/api-v3/groups.js
Yutsuten: Do you mean something like this?
``` JavaScript
if (emails) {
   memberCount += emails.length;
}
```

Yutsuten:
@paglias I have no experience with tests, but I would like to try it. Do you know how I could learn and create the necessary tests to validate the changes I'm doing here?

website/server/controllers/api-v3/groups.js
paglias: yes

paglias:
@mateus-etto all the tests for controllers can be found in `test/api/v3/integration`. You can get a good understanding from there but if you have any question feel free to post here and I'll answer.

The software we use for testing is mocha.js, chai (expect version) and sinon.js

You'll want to test in particular that the limit is enforced with correct error codes, ...

Yutsuten:
@paglias I finished my changes in the code, so now I'm going to the tests.

I installed mocha.js, chai and sinon.js, and was able to run ``npm test`` successfully. But the complete test takes about 10 minutes to run, and I would like to run only a single script or test.
I tried ``mocha <path/to/test.js>``, ``mocha --grep <pattern>`` and ``mocha *test*`` but I'm getting a lot of errors...

Error when using path to file: 
"error: Error: Slack url is required"

Error when using ``--grep <patterns>`` (on folder test/api/v3/integration/groups): 
"Warning: Could not find any test files matching pattern: user id invites"

Error when using ``mocha *test*`` (on folder test/api/v3/integration/groups): 
"(function (exports, require, module, __filename, __dirname) { **import** {
SyntaxError: Unexpected token import"

Could you help me?

paglias:
@mateus-etto that's great! For tests you can run integration tests only with `gulp test:api-v3:integration` and unit tests with `gulp test:api-v3:unit`. If you want to start the server separately from the integration tests you can do `NODE_ENV=test npm start` and then in a different terminal window `gulp test:api-v3:integration:separate-server`.

To run only a single test or only a single file you'll have to change `describe('section name', ...)` or `it('test name', ...)` to `describe.only(...)` or `test.only(..)`.

Yutsuten:
I "fixed" the test "returns only first 30 invites" that was being done in a party. As a party can have up to only 30 members now, I changed the check to a guild instead.

I also added 3 tests:
1. Party with 1 leader and 29 invites, total 30 members (must finish with success)
2. Party with 1 leader and 30 invites, total 31 members (must finish with fail)
3. Guild with 1 leader and 30 invites, total 31 members (must finish with success)

With this, I finished my to-do list for this PR. I'm ready for code/test reviews.

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -321,6 +322,20 @@ describe('Post /groups/:groupId/invite', () => {
...
+      // Generate 30 users to invite (30 + leader = 31 members)
+      for (let i = 0; i < PARTY_LIMIT_MEMBERS; i++) {
+        invitesToGenerate.push(generateUser());
+      }
+      let generatedInvites = await Promise.all(invitesToGenerate);
+      // Invite users
+      expect(await inviter.post(`/groups/${group._id}/invite`, {
+        uuids: generatedInvites.map(invite => invite._id),
+      }))
+      .to.be.an('array');
```
paglias: this should be on the previous line

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -410,5 +425,37 @@ describe('Post /groups/:groupId/invite', () => {
...
+      // Generate 29 users to invite (29 + leader = 30 members)
+      for (let i = 0; i < PARTY_LIMIT_MEMBERS - 1; i++) {
+        invitesToGenerate.push(generateUser());
+      }
+      let generatedInvites = await Promise.all(invitesToGenerate);
+      // Invite users
+      expect(await inviter.post(`/groups/${party._id}/invite`, {
+        uuids: generatedInvites.map(invite => invite._id),
+      }))
+      .to.be.an('array');
```
paglias: previous line

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -410,5 +425,37 @@ describe('Post /groups/:groupId/invite', () => {
...
+    });
+
+    it('do not allow 30+ members in a party', async () => {
```
paglias: does not allow

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -410,5 +425,37 @@ describe('Post /groups/:groupId/invite', () => {
       });
       expect((await userToInvite.get('/user')).invitations.party.id).to.equal(party._id);
     });
+
+    it('allow 30 members in a party', async () => {
```
paglias: allows

test/api/v3/integration/groups/POST-groups_invite.test.js
```diff
@@ -321,6 +322,20 @@ describe('Post /groups/:groupId/invite', () => {
       });
     });
 
+    it('allow 30+ members in a guild', async () => {
```
paglias: allows

website/server/controllers/api-v3/groups.js
```diff
@@ -1066,6 +1067,31 @@ api.inviteToGroup = {
     let uuids = req.body.uuids;
     let emails = req.body.emails;
 
+    // If party, check the limit of members
+    if (group.type === 'party') {
+      let memberCount = 0;
+
+      // Counting the members that already joined the party
+      memberCount += group.memberCount;
```
paglias: can you add a blank line after this one? It helps with readability

website/server/controllers/api-v3/groups.js
```diff
@@ -1066,6 +1067,31 @@ api.inviteToGroup = {
...
+      // Count how many invitations currently exist in the party
+      let query = {};
+      query['invitations.party.id'] = group._id;
+      let groupInvites = await User
+        .find(query)
+        .select('_id')
+        .exec();
+      memberCount += groupInvites.length;
```
paglias: blank line

website/server/controllers/api-v3/groups.js
```diff
@@ -1066,6 +1067,31 @@ api.inviteToGroup = {
...
+      // Counting the members that are going to be invited by email and uuids
+      if (uuids) {
+        memberCount += uuids.length;
+      }
+      if (emails) {
+        memberCount += emails.length;
+      }
```
paglias: blank line

website/server/controllers/api-v3/groups.js
```diff
@@ -1066,6 +1067,31 @@ api.inviteToGroup = {
...
+      if (memberCount > common.constants.PARTY_LIMIT_MEMBERS)
```
paglias: parenthesis

website/server/controllers/api-v3/groups.js
```diff
@@ -1066,6 +1067,31 @@ api.inviteToGroup = {
...
+      // Count how many invitations currently exist in the party
+      let query = {};
+      query['invitations.party.id'] = group._id;
+      let groupInvites = await User
+        .find(query)
+        .select('_id')
+        .exec();
+      memberCount += groupInvites.length;
+      // Counting the members that are going to be invited by email and uuids
+      if (uuids) {
```
paglias: can you move this if and the one after (`if (email)`) to after the call to `Group.validateInvitations`? Because at this point the uuids and emails params have not been validated and may not be arrays

paglias (changes requested):
I've let a lot of comments but they're mostly style issues. The code is very good

website/server/controllers/api-v3/groups.js
```diff
@@ -1068,6 +1069,36 @@ api.inviteToGroup = {
...
+      let query = {};
+      query['invitations.party.id'] = group._id;
+      let groupInvites = await User
+        .find(query)
+        .select('_id')
+        .exec();
+      memberCount += groupInvites.length;
+      
+      // Counting the members that are going to be invited by email and uuids
+      if (uuids) {
```
paglias: last edit then I think it's good to go: since we already have a function to validate invites ( `Group.validateInvitations` ) I think we should use that for validating the number of invites too. It's only used here so feel free to change it if you need to pass any other parameter that you might need

paglias (changes requested):

paglias:
Other than my last commit, can you post a screenshot of the changes when you've reached the maximum amount of members?

website/server/controllers/api-v3/groups.js
Yutsuten: Do you mean that I could put all my code that validates the size of party inside ``Group.validateInvitations``?

paglias:
Yes

Il 24 Mar 2017 19:28, "Mateus Etto" <notifications@github.com> ha scritto:

Yutsuten:
Something like this?

InvitedAll

paglias:
@mateus-etto that looks good, thanks!

website/server/controllers/api-v3/groups.js
paglias: yes, exactly

mateus-etto:
Looks like that changing the ``Group.validateInvitations`` function broke some other tests...
I'll see if I can make it work again.

paglias:
They're probably failing because the validation function is now
asynchronous so the tests will need to be changed accordingly

Il 24 Mar 2017 20:42, "Mateus Etto" <notifications@github.com> ha scritto:

Yutsuten:
@paglias I did as you suggested, thank you for the suggestions.
I think I fixed the tests now, locally they worked without problems...

deltaidea:
I've been randomly following this conversation over the last couple days and I've gotta say, this is the most friendly and open to contributions project I've ever seen.

test/api/v3/unit/models/group.test.js
```diff
@@ -460,73 +460,67 @@ describe('Group Model', () => {
         };
       });
 
-      it('throws an error if no uuids or emails are passed in', (done) => {
+      it('throws an error if no uuids or emails are passed in', async () => {
```
paglias: there's a problem with these tests, when you have an `async` function and you call something async like `Group.validateInvitations` you need to put `await` in front of it otherwise `Group.validateInvitations` will be called but not waited for and the function will continue to the next step regardless of what happens in `Group.validateInvitations`. So the tests are passing but actually the expec

Also this test was written to make sure that if `Group.validateInvitations ` didn't throw an error (the expected behavior) the test would timeout because `done()` wouldn't be called. Doing the same with `async/await` is less straightforward but you can try something like this

```
      it('throws an error if no uuids or emails are passed in', () => {
       let isSuccessful = false;
        try {
           await Group.validateInvitations(null, null, res);
           isSuccessful = true;
        } catch (err) {
            expect(err).to.be.an.instanceof(BadRequest);
            expect(res.t).to.be.calledOnce;
            expect(res.t).to.be.calledWith('canOnlyInviteEmailUuid');
        });

        if (isSuccessfull) throw new Error('Group.validateInvitations should have throw an error but it did not.')
   })
```

paglias (changes requested):

Yutsuten:
I see... I tried something different:

``` JavaScript
it.only('throws an error if no uuids or emails are passed in', async () => {
  await expect(Group.validateInvitations(null, null, res)).to.eventually.be.rejected.and.eql({
    httpCode: 400,
    message: 'Bad request.',
    name: 'BadRequest',
  });
  expect(res.t).to.be.calledOnce;
  expect(res.t).to.be.calledWith('canOnlyInviteEmailUuid');
});
```

The test finishes with success, but is this code okay in your opinion?

Yutsuten:
I also can change:

``` JavaScript
.to.eventually.be.rejected.and.eql({
    httpCode: 400,
    message: 'Bad request.',
    name: 'BadRequest',
  });
```

to: ``.to.eventually.be.rejected.and.to.be.an.instanceof(BadRequest);``

The code would get even more cleaner. Tested here and looks ok.

paglias:
Oh yes, I forgot about that format but it's much more readable. Let's keep it like your first example that is the same we use in all other tests, without the `and.to.be.an.instanceof(BadRequest); `

test/api/v3/unit/models/group.test.js
```diff
@@ -556,47 +546,47 @@ describe('Group Model', () => {
           emails.push(`user-${i}@example.com`);
         }
 
-        expect(function () {
-          Group.validateInvitations(uuids, emails, res);
+        expect(async function () {
```
paglias: these `expect()...to.not.throw()` aren't working correctly with `async` functions. If you change the code to

```
      it.only('does not throw an error if emails are passed in and uuids are an empty array', async () => {
        expect(async function () {
          throw new Error('error');
          await Group.validateInvitations([], ['user1@example.com', 'user2@example.com'], res);
        }).to.not.throw();

        expect(res.t).to.not.be.called;
      });
```

the test passes even if it should not.

But you don't need to expect the function not to throw because if an error is thrown inside an async function the function and the test fails. So this will work

```
      it.only('does not throw an error if emails are passed in and uuids are an empty array', async () => {
        await Group.validateInvitations([], ['user1@example.com', 'user2@example.com'], res);
        expect(res.t).to.not.be.called;
      });
```

paglias (changes requested):

test/api/v3/unit/models/group.test.js
```diff
@@ -460,73 +457,67 @@ describe('Group Model', () => {
...
-        } catch (err) {
-          expect(err).to.be.an.instanceof(BadRequest);
-          expect(res.t).to.be.calledOnce;
-          expect(res.t).to.be.calledWith('canOnlyInviteEmailUuid');
-          done();
-        }
+      it('throws an error if no uuids or emails are passed in', async () => {
+        await expect(Group.validateInvitations(null, null, res)).to.eventually.be.rejected.and.eql({
+          httpCode: 400,
+          message: 'Bad request.',
```
paglias, line 463: these messages shouldn't be the proper i18n strings like `canOnlyInviteEmailUuid ` in this case?

paglias (changes requested):

test/api/v3/unit/models/group.test.js
Yutsuten: The message returned in the call to ``Group.validateInvitations`` is ``Bad request.``.

The only expectation about the call of this function is that it will fail with Bad Request. The reason of fail (``canOnlyInviteEmailUuid``) is on the ``res`` variable.

So, I can't see a way to do what you're requesting. Sorry.

test/api/v3/unit/models/group.test.js
Yutsuten: Makes sense. I'll fix it.

test/api/v3/unit/models/group.test.js
paglias: Yes, I'm wondering why the message is `Bad request` and not the real reason, it should not be `Bad message`. Does the test fails if `message: 'Bad request.'` becomes `message: shared.i18n.t('canOnlyInviteEmailUuid')`? 

I can't see where `Bad request.` as a message would come from. It should be like this https://github.com/HabitRPG/habitica/pull/8589/files#diff-ee4d06fc0df454c1e75431b5abb56e39R455

paglias (changes requested):

test/api/v3/unit/models/group.test.js
Yutsuten: Yes, the test fails...
```
AssertionError: expected { Object (name, httpCode, ...) } to deeply equal { Object (httpCode, message, ...) }
+ expected - actual

 {
   "httpCode": 400
-  "message": "Bad request."
+  "message": "Can only invite using uuids or emails."
   "name": "BadRequest"
 }
```

paglias: Ok, the `Bad request.` string seems to come from https://github.com/HabitRPG/habitica/blob/develop/website/common/script/libs/errors.js#L31 and it's used as a default when the message isn't provided and it happens because the `res` used in tests (https://github.com/HabitRPG/habitica/pull/8589/files/414e316e832f398b9a695aeb20f9223aa2f556f8#diff-548695dbee775dda2626c88e15cbcec6L459) doesn't return anything but it's just a stub. So false alarm :)

Yutsuten:
If there is anything else that I should fix, please tell me! :)

paglias:
Merged :) I've also awarded your first contributor tier on Habitica. Really good work with the PR, thanks! 

Note that the PR being merged doesn't mean it's live on Habitica, it'll go there in a few days when it's also merged into the `release` branch

Alys:
Just as a note for our reference, paglias has pointed out that party invitations made by email to users outside of Habitica are not stored anywhere within Habitica, so they’re not considered when calculating the limit. This would allow parties to grow beyond 30 but it's unlikely to be a large problem and not worth the extra code and database storage that would be needed to keep track of emailed invitations.

paglias:
I've made two small changes in the group view https://github.com/HabitRPG/habitica/commit/daf9421f4f9e3b1ad15a675c9e06c69e2c57f7e8

They make sure the invite button isn't hidden in guilds and completely hide the panel used for the members limit message in guilds

Yutsuten:
Oh, those are very big mistakes of mine... I need to be more careful in my tests...

PR #8591, opened 2017-03-21
Group managers

TheHollidayInn:

Fixes put_issue_url_here

### Changes
This allows users to be assigned as managers and allows those managers to:
 - create, edit, delete group tasks
 - Assign and unassign users
 - Approve approval requests

----
UUID:


Alys:
- [x] Before this is merged, I'd like to get feedback from a couple of the Socialites about the UI aspects. I can do that by putting it on a local install that I have on a public server (uses its own database, not the production database). I'll get that set up today.

Alys:
- [x] As shown in the screenshot below, there should be more space to the left of the "Remove" button and less space before the managers' names in that section. There should also be more space to the left of the "Add Manager" button.
[screenshot](https://cloud.githubusercontent.com/assets/1495809/24175784/dd0deb26-0ee3-11e7-85bb-e86f58274ee9.png)

- [x] Also as shown there, the leader can add themself as manager. Is that intentional? On the one hand, it's undesirable because they leader might think they need to and that wastes their time. On the other hand, if the leader is about to transfer leadership to another member, they might want to make themself a manager first. However that would only happen after the group plan had been cancelled, and it should probably then be up to the new leader whether the old one is allowed to be a manager. I'm leaning towards making it not possible for the leader to declare themself a manger too.

- [x] When a task requiring approval is completed, only the group leader gets a notification that an approval is waiting; managers do not get approvals. If that's deliberate, that's fine but can you let us know if it's deliberate so that we can document it? If there's anything else non-obvious that you think should be documented about this, can you edit the top post to give details about the new features?

- [x] When a task awaiting approval is approved by a manager, the notification that approval is needed is not removed from the group leader's notification list. This is going to cause wasted time and annoyance for the leader. Possible solutions include:
- Give all the group plans notifications "x" buttons so that they can be more easily dismissed - if none of the other options are done, I think this will be essential to reduce the complaints. Even if one of the other options is done, this will still be a good idea now.
- Remove the notification when the task is approved by someone else.
- Don't have individual notifications for any tasks awaiting approval, but instead have one notification for each group plan stating that there is at least one task to be approved. As soon as all tasks are approved, that notification could be automatically removed (or at the very least could be dismissable with an "x" button).

- [x] If a member completes a task requiring approval and it's approved by both a manager and the leader, the member receives twice as much XP and GP for it (i.e., both approvals give rewards to the user). They also receive two notifications that it was approved. This can only happen if the second approver had not refreshed the list of approvals before approving, but that could happen in real life when two approvers happened to be working through the list of pending approvals at the same time.

Alys:
Oh, this comment might be irrelevant: "However that would only happen after the group plan had been cancelled". That's from the part where I was saying that the leader can add themself as manager.

- [x] I see now that managers also be added to public guilds and parties. Assuming that's deliberate, what functions can they do?

website/client-old/js/components/groupTasks/groupTasksController.js
```diff
@@ -158,10 +158,12 @@ habitrpg.controller('GroupTasksCtrl', ['$scope', 'Shared', 'Tasks', 'User', '$ro
        Tasks.updateTask(task._id, task);
      };
 
-    $scope.checkGroupAccess = function (group) {
+    $scope.checkGroupAccess = function (group, sup) {
```
Alys: Can `sup` be given a more meaningful name? I can't work out from the code what it's for so a longer name might help.

website/server/controllers/api-v3/tasks.js
```diff
@@ -413,9 +413,13 @@ api.updateTask = {
       throw new NotFound(res.t('taskNotFound'));
     } else if (task.group.id && !task.userId) {
       //  @TODO: Abstract this access snippet
-      group = await Group.getGroup({user, groupId: task.group.id, fields: requiredGroupFields});
+      let fields = requiredGroupFields.concat(' managers');
+      group = await Group.getGroup({user, groupId: task.group.id, fields});
       if (!group) throw new NotFound(res.t('groupNotFound'));
-      if (group.leader !== user._id) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
+      let isNotGroupLeader = group.leader !== user._id;
```
Alys: Can `isNotGroupLeader` be changed to `isGroupLeader`? That would be consistent with `isManager`.

website/client-old/js/components/groupTasks/groupTasksController.js
```diff
@@ -158,10 +158,12 @@ habitrpg.controller('GroupTasksCtrl', ['$scope', 'Shared', 'Tasks', 'User', '$ro
...
       if (!group || !group.leader) return true;
-      if (User.user._id !== group.leader._id) return false;
-      return true;
+      var userId = User.user._id;
+      var leader = group.leader._id === userId;
+      var userIsManager = !!group.managers[userId];
```
Alys: Can this be changed to be more like `isManager = Boolean(group.managers[user._id]);` for consistency with that code?

Alys (commented):
I haven't actually done a code review. These are just a couple of things that I noticed when I was looking through the code to work out how the new features work. :)

website/client-old/js/components/groupTasks/groupTasksController.js
TheHollidayInn: Yea... That is an accident lol

website/server/controllers/api-v3/groups.js
```diff
@@ -1086,4 +1086,100 @@ api.inviteToGroup = {
...
+ */
+api.addGroupManager = {
+  method: 'POST',
+  url: '/groups/:groupId/add-manager',
+  middlewares: [authWithHeaders()],
+  async handler (req, res) {
+    let user = res.locals.user;
+    let managerId = req.body.managerId;
+
+    req.checkParams('groupId', res.t('groupIdRequired')).notEmpty(); // .isUUID(); can't be used because it would block 'habitrpg' or 'party'
```
paglias: since these errors are not meant for end users, can we use the new `apiMessages` lib like here https://github.com/HabitRPG/habitica/commit/939712ad1f8e8c8e228907b30844fcd3bb5dae51#diff-b950a7c557b35ea58bc2685532c46170R285 So the i18n strings to load on the client don't grow in size. It should be used for all API messages meant for devs that don't need to be translated

website/server/controllers/api-v3/groups.js
```diff
@@ -1086,4 +1086,100 @@ api.inviteToGroup = {
...
+    req.checkBody('managerId', res.t('managerIdRequired')).notEmpty();
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let newManager = await User.findById(managerId).exec();
```
paglias: selecting `guilds` should be enough in this case

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -171,11 +175,15 @@ api.assignTask = {
       throw new NotAuthorized(res.t('onlyGroupTasksCanBeAssigned'));
     }
 
-    let groupFields = `${requiredGroupFields} chat`;
+    let groupFields = `${requiredGroupFields} chat managers`;
     let group = await Group.getGroup({user, groupId: task.group.id, fields: groupFields});
     if (!group) throw new NotFound(res.t('groupNotFound'));
 
-    if (group.leader !== user._id && user._id !== assignedUserId) throw new NotAuthorized(res.t('onlyGroupLeaderCanEditTasks'));
+    let isNotGroupLeader = group.leader !== user._id;
```
paglias: can we abstract these lines into a function since it's used in 4 places in this file?

website/server/models/group.js
```diff
@@ -105,6 +105,9 @@ export let schema = new Schema({
       return {};
     }},
   },
+  managers: {type: Schema.Types.Mixed, default: () => {
```
paglias, line 108: this field should be added to the `noSet` list

paglias (changes requested):

website/server/controllers/api-v3/groups.js
Alys: Parties can have group plans too. Is that relevant here?

paglias: maybe... I said guilds because it only checks for them 7 lines below this one

website/server/controllers/api-v3/groups.js
```diff
@@ -1086,4 +1086,100 @@ api.inviteToGroup = {
...
+    let groupFields = basicGroupFields.concat(' managers');
+    let group = await Group.getGroup({user, groupId: req.params.groupId, fields: groupFields});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    if (group.leader !== user._id) throw new NotAuthorized(res.t('messageGroupOnlyLeaderCanUpdate'));
+
+    let isMember = newManager.guilds.indexOf(group._id) !== -1;
```
paglias: group.isMember should be used here https://github.com/HabitRPG/habitica/blob/develop/website/server/models/group.js#L369

Alys:
- [x] When a party leader upgrades their party, they cannot use this new feature to make a party member a Manager. When they try, they see the message "User must be a member" (I checked that the user I was trying to make a manager really was a member).

website/server/controllers/api-v3/tasks/groups.js
TheHollidayInn: Yea, we haven't been abstracted our permission checks, but I'm fine getting that started.

website/server/libs/apiMessages.js
```diff
@@ -8,6 +8,8 @@ const messages = {
   guildsOnlyPaginate: 'Only public guilds support pagination.',
   guildsPaginateBooleanString: 'req.query.paginate must be a boolean string.',
   guildsPageInteger: 'req.query.page must be an integer greater than or equal to 0.',
+  groupIdRequired: 'req.params.groupId must contain a groupId.',
```
paglias, line 11: can you remove the strings from the english locale now that they've been added here?

paglias (changes requested):

website/server/libs/apiMessages.js
TheHollidayInn: This local is used in multiple places not just my routes.

paglias: Ah right

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -16,6 +16,14 @@ import {
 
 let requiredGroupFields = '_id leader tasksOrder name';
 let types = Tasks.tasksTypes.map(type => `${type}s`);
+
+function canNotEditTasks(group, user, assignedUserId) {
+  let isNotGroupLeader = group.leader !== user._id;
```
Alys: Can `isNotGroupLeader` be changed to `isGroupLeader` ?

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -16,6 +16,14 @@ import {
...
+  let isManager = Boolean(group.managers[user._id]);
+  let userIsAssigningToSelf = Boolean(assignedUserId && user._id === assignedUserId);
+  let canNotEditTasks = isNotGroupLeader && !isManager && !userIsAssigningToSelf;
```
Alys: This `canNotEditTasks` code looks like there's a new feature to stop the group leader doing something they used to be able to. Is that right? If so, can you please write something about that in the PR so that we have it documented? If a behaviour has changed, we'll need to update the wiki and make sure the socialites know about it so that they can answer questions (also Sabe and I need to know so that we can handle emailed questions).

website/client-old/js/components/groupTasks/groupTasksController.js
```diff
@@ -160,8 +160,10 @@ habitrpg.controller('GroupTasksCtrl', ['$scope', 'Shared', 'Tasks', 'User', '$ro
 
     $scope.checkGroupAccess = function (group) {
       if (!group || !group.leader) return true;
-      if (User.user._id !== group.leader._id) return false;
-      return true;
+      var userId = User.user._id;
+      var leader = group.leader._id === userId;
+      var userIsManager = Boolean(group.managers[userId]);
```
Alys: Can `userIsManger` be just `isManager` for consistency with other code? If they all have the same name, finding all places where the check is made will be easier, if we ever need to do that. (Also in the other places where this appears.)

website/common/locales/en/limited.json
```diff
@@ -92,5 +92,8 @@
     "winterEventAvailability": "Available for purchase until January 31.",
     "springEventAvailability": "Available for purchase until May 31.",
     "summerEventAvailability": "Available for purchase until August 31.",
-    "valentinesEventAvailability": "Available for purchase until February 28."
+    "valentinesEventAvailability": "Available for purchase until February 28.",
+    "managerIdRequired": "req.body.managerId is required",
```
Alys: I'm not sure about this but it looks like `managerIdRequired` should be in apiMessages.json. Althernatively, if it is meant to be seen by the users and translated, we should make the message more user-friendly.

Alys (changes requested):

website/server/controllers/api-v3/tasks/groups.js
TheHollidayInn: Nope, this is not a change. If they are a group leader they can edit tasks. In English this says,

If you are not a group leader or a manager or assigning yourself, then you cannot assign this task.

website/common/locales/en/limited.json
TheHollidayInn: Paglias said to put it here in a comment above.

TheHollidayInn: It is not meant to be translated.

TheHollidayInn:
" I see now that managers also be added to public guilds and parties. Assuming that's deliberate, what functions can they do?"

@Alys Parties can have group plans. And managers will be able to do other things as well in the future. I can hide this options on the UI if they are not a group plan, I guess.

website/server/controllers/api-v3/tasks/groups.js
Alys: Ah, sorry, that makes sense. I was confused by the mix of editing and assigning terminology here.

Is this line a concern here?
`let userIsAssigningToSelf = Boolean(assignedUserId && user._id === assignedUserId);`
Since the canNotEditTasks() function is used to control editing as well as assigning, will that line allow an ordinary member to edit the task if they are assigning it to themself? Even if the code is safe now, is it possible it could introduce that bug in future if the function is used in ways that it's not being used now? It feels odd that a function labelled as being for controlling editing is also being used to allow other permissions to users who should never be allowed to edit.

Would it be safer to have two functions, one to control editing and the other to control assigning? The function to control assigning could use canNotEditTasks() for most of its checks. At the least, I think this function needs to be renamed or commented to accurately describe what it's for.

Alys:
Yes, please hide the options for guilds and parties that don't have group plans. If they're visible, we'll get a lot of questions about them and complaints that they don't do anything.

website/common/locales/en/limited.json
Alys: It should go in the apiMessages file. Some translators would try to convert `req.body.managerId` to their own language.

website/server/controllers/api-v3/tasks/groups.js
TheHollidayInn: Yea, that makes sense. The abstraction was completely correct. I'll split the functions.

TheHollidayInn:
I just pushed an update that has the following additions (mangers include leader):
1. All managers will get notifications
2. Only one manger will be able to approve a task
3. All managers will have the notification remove if anyone approves
4. Parties can now use managers correctly

website/common/locales/en/limited.json
paglias: Yeah it should go into `apiMessages`, I had only looked at the other file previously

website/client-old/js/components/groupTasks/groupTasksController.js
TheHollidayInn: Not exactly user what you mean. I'm fine changing the variable, but this function is abstracted so the variable definition is only here.

website/common/locales/en/limited.json
```diff
@@ -88,6 +88,14 @@
     "grandMalkinSet": "Grand Malkin (Mage)",
     "cleverDogSet": "Clever Dog (Rogue)",
     "braveMouseSet": "Brave Mouse (Warrior)",
+    "fallEventAvailability": "Available for purchase until October 31.",
+    "winterEventAvailability": "Available for purchase until January 31.",
+    "springEventAvailability": "Available for purchase until May 31.",
+    "summerEventAvailability": "Available for purchase until August 31.",
+    "valentinesEventAvailability": "Available for purchase until February 28.",
+    "managerIdRequired": "req.body.managerId is required",
```
paglias: this should be removed now that it's been moved to apiMessages

website/common/locales/en/limited.json
```diff
@@ -88,6 +88,14 @@
     "grandMalkinSet": "Grand Malkin (Mage)",
     "cleverDogSet": "Clever Dog (Rogue)",
     "braveMouseSet": "Brave Mouse (Warrior)",
+    "fallEventAvailability": "Available for purchase until October 31.",
+    "winterEventAvailability": "Available for purchase until January 31.",
+    "springEventAvailability": "Available for purchase until May 31.",
+    "summerEventAvailability": "Available for purchase until August 31.",
+    "valentinesEventAvailability": "Available for purchase until February 28.",
```
paglias: all these strings don't seem to be used anywhere in the PR

paglias (changes requested):

website/common/locales/en/limited.json
TheHollidayInn: Not sure what happened here. These appeared when I merged in Devleop.

website/common/locales/en/limited.json
TheHollidayInn: I have no idea why these strings are here. I must have put them here by accident I guess :P

website/common/locales/en/limited.json
paglias: I think these should be removed too, probably some glitch with develop

Alys:
Once the code changes are done, I'd like to test again locally before this is merged. @TheHollidayInn Should I do that now or are there more changes to come?

TheHollidayInn:
Should be ready for more testing! Thanks 👍

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -296,7 +310,29 @@ api.approveTask = {
...
 
-    await Bluebird.all([assignedUser.save(), task.save()]);
+    let managerIds = Object.keys(group.managers);
+    managerIds.push(group.leader);
+    let managers = await User.find({_id: managerIds}).exec(); // Use this method so we can get access to notifications
+
+    let managerPromises = [];
+    managers.forEach((manager) => {
+      let notificationIndex =  findIndex(manager.notifications, function findNotification (notification) {
+        let notificationMessage = res.t('userHasRequestedTaskApproval', {
```
paglias: there isn't a better way to identify the notification? Relaying on the translated text doesn't seem optimal and would fail if the user changes its language

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -296,7 +310,29 @@ api.approveTask = {
       scoreTask: task,
     });
 
-    await Bluebird.all([assignedUser.save(), task.save()]);
+    let managerIds = Object.keys(group.managers);
+    managerIds.push(group.leader);
+    let managers = await User.find({_id: managerIds}).exec(); // Use this method so we can get access to notifications
```
paglias: can we just fetch notifications?

website/server/controllers/api-v3/tasks/groups.js
paglias: Actually this won't work as soon as one manager has a language different than the others

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -13,9 +14,18 @@ import {
   getTasks,
   moveTask,
 } from '../../../libs/taskManager';
+import apiMessages from '../../../libs/apiMessages';
 
 let requiredGroupFields = '_id leader tasksOrder name';
 let types = Tasks.tasksTypes.map(type => `${type}s`);
+
+function canNotEditTasks (group, user, assignedUserId) {
```
paglias, line 21: this function (just statements outside of ay function) seems also used in  website/server/controllers/api-v3/tasks.js, can we abstract it somewhere and add some unit tests?

website/server/controllers/api-v3/tasks.js
```diff
@@ -530,18 +534,28 @@ api.scoreTask = {
...
-          taskName: task.text,
-        }, groupLeader.preferences.language),
-        groupId: group._id,
+      let fields = requiredGroupFields.concat(' managers');
+      let group = await Group.getGroup({user, groupId: task.group.id, fields});
+
+      // @TODO: we can use the User.pushNotification function because we need to ensure notifications are translated
+      let managerIds = Object.keys(group.managers);
+      managerIds.push(group.leader);
+      let managers = await User.find({_id: managerIds}).exec(); // Use this method so we can get access to notifications
```
paglias: let's just load preferences and notifications?

paglias (changes requested):

website/server/controllers/api-v3/tasks/groups.js
TheHollidayInn: Yep, I added a comment about that.

website/server/controllers/api-v3/tasks/groups.js
TheHollidayInn: If we start abstracting permissions for routes, I'd rather have a full separate PR on creating "policy" services in a unified way.

paglias: In general I'd say okay but this PR is adding the same code just written differently in two different places. It won't require much effor to create a function like `user.isSubscribed()` and avoid confusion later

website/server/controllers/api-v3/tasks/groups.js
paglias: so what happens if two managers have different languages? the notification cannot be removed?

TheHollidayInn: Oh, wrong section. I see what you mean. Let me add the task_id instead.

website/server/controllers/api-v3/tasks/groups.js
TheHollidayInn: IT would be easy to abstract, just incorrect. user.isLeader, user.isManager make sense, but user.canNotEditTasks doesn't really make sense for the owning object to be the approver. 

So, you could add the function to the group and have group.userCanEditTask, but then we see that this code is duplicated as well: https://github.com/HabitRPG/habitica/blob/develop/website/server/controllers/api-v3/tasks.js#L1192 . So, why are we abstracting one and not the other? 

Then, we run into a SOLID design issue on the group model (which we already have). But we can see that a better (first step) would be to create a TaskPolicy class to begin handling these permissions. Which would be out of scope of this PR.

So, sure, it is easy, but it would be support bad code that I have already brought up as an issue.

test/api/v3/integration/tasks/groups/POST-group_tasks_id_score_direction.test.js
```diff
@@ -63,6 +64,40 @@ describe('POST /tasks/:id/score/:direction', () => {
     expect(updatedTask.group.approval.requestedDate).to.be.a('string'); // date gets converted to a string as json doesn't have a Date type
   });
 
+  it('sends notificatoins to all managers', async () => {
```
Alys: Typo: notificatoins > notifications

website/server/libs/apiMessages.js
```diff
@@ -8,6 +8,9 @@ const messages = {
...
+  managerIdRequired: 'req.body.managerId must contain a user ID.',
+  canOnlyApproveTaskOnce: 'You can only approve a task once',
```
Alys: Should `canOnlyApproveTaskOnce` be translatable? It sounds like a message the group leader will see.

The text would be better as "This task has already been approved." 
("You can only approve a task once" will confuse a user who thinks it's saying that they themself have tried to approve it twice.)

website/views/options/social/group.jade
```diff
@@ -64,7 +64,18 @@ a.pull-right.gem-wallet(ng-if='group.type!="party"', popover-trigger='mouseenter
 
               h4=env.t('assignLeader')
               select#group-leader-selection(ng-model='groupCopy._newLeader', ng-options='member.profile.name for member in group.members')
-
+
```
Alys: This blank line has whitespace in it (its a really minor, picky point but I think we try to avoid that, and git shows it as a red mark). Line 78 at the end of this block also has whitespace.

Alys (changes requested):

Alys:
This works:
	"All managers will have the notification remove if anyone approves"
except for two things:
- [x] The notification disappears only when you reload, not when you sync. Can it be made to disappear on sync? Reloading isn't a common action at the moment.
- [x] The notification doesn't disappear if the manager has been demoted before approval was done. Steps to reproduce:
   1. The leader makes user Bob a manager.
   2. Someone completes a task requiring approval.
   3. All managers, including Bob, get a notification that approval is needed.
   4. Before the task is approved, the leader removes Bob's management status. Bob still has the notification.
   6. The task is approved.
   7. The notification is removed from all **current** managers. Bob still has the notification.
   - This could be fixed by removing the notification at step 4 (probably best) or at step 7. It's an edge case though, so if both of those are difficult to achieve, we could ignore it.

I'd really like it if "x" buttons could be added to the notifications in this PR, if it's not very difficult. In groups of peers, potentially all group members could be managers, and even with the notifications disappearing automatically when approvals are done, the notifications are still going to seem spammy if they can't also be deleted manually.

----

I think to make the manager feature easy for group members to follow, the group membership list needs to state who is the leader and who are the managers. For example, something like this:

image

The "leader" indicator should appear on all parties and guilds, not just those that use group plans or that have managers.

I suggest that the "manager" indicator should also be coded so that it could work in all groups, not just those that have group plans (even though managers currently can't be assigned in non-group-plan groups). That way the indicators will work immediately once managers are enabled for all groups.

Alys:
- [x] The leader and manager indicators in the group membership list would be better at the right of the HP indicator, otherwise mischievous players can make themselves look like a leader or manager by changing their name, and I am certain we have some players who would do that. For example in the image below, Canasta is not the leader. It would probably be best if the leader/manager indicators had a different appearance as well, for example, normal text rather than being part of the link.

image


@TheHollidayInn Are you still working through the previous suggestions or is this ready to be tested again?

paglias (approved):

Alys:
Ah heck I'm so sorry. I tried to push a commit with some minor typo fixes and one text change and I've screwed it up by pushing recent commits from develop. :( 

@TheHollidayInn I think if you force push from your local group-managers branch to your remote group-managers branch it will clean everything up.

The commit I was trying to push was this one: https://github.com/HabitRPG/habitica/pull/8591/commits/595441ab59c10822088de35c43d5439d41e217e1

I've tested and everything else looks good except for a small problem with notifications:

When a completed task is approved, all managers and leader are supposed to be able to sync to remove the notification about the task needing approval. I've found that doesn't work all the time though. I'm not sure this is consistent but it seems that usually when there's only one notification about a pending task approval, then the notification is often not removed by sync. If there's two or more notifications, each one is removed by a sync.
If you click a notification that should have been removed, you get a "notification" not found message and then the notification disappears, so this isn't a critical error. If it's complex to fix, I don't think it needs to be fixed for this PR to be merged, but maybe an issue should be logged for it. It might be part of a wider problem with our notification system - that might be what one of the new comments in the code is talking about. :)

PR #8592, opened 2017-03-22
Redirection fixes

SabreCat:

### Changes

**Previously**, the `redirects` middleware treated the `IGNORE_REDIRECT` environment variable as a Boolean. When retrieved in environments like Heroku, however, this variable is a string; therefore, `redirects` always acted as if redirection was ignored in these scenarios. That caused some undesirable behavior, such as users encountering the login page when given a www.habitica.com link, even if they were properly logged in to habitica.com.

**Now**, when `redirects` looks up `IGNORE_REDIRECT`, it correctly converts it to a Boolean by checking for the `'true'` string, allowing for redirection to proceed as expected.

---

**Previously**, Habitica used an extra load balancing layer to handle the habitrpg.com to habitica.com transition that needed to be accounted for in the `redirects` middleware.

**Now**, that architecture layer is no longer in use (see instead https://github.com/HabitRPG/habitrpg-redirect), so the checks for it in `redirects` have been removed.

---

**Previously**, the `redirects` middleware did not distinguish between request methods, despite Express `res.redirect` only being applicable to `GET` requests.

**Now**, `redirects` does not attempt to act on requests using methods other than `GET`.

---

**Previously**, the `redirects` middleware had no unit tests.

**Now**, tests have been added covering the current logic.

### DIFF
--- a/test/api/v3/unit/middlewares/redirects.js
+++ b/test/api/v3/unit/middlewares/redirects.js
@@ -1,0 +1,182 @@
+import {
+  generateRes,
+  generateReq,
+  generateNext,
+} from '../../../../helpers/api-unit.helper';
+import nconf from 'nconf';
+import requireAgain from 'require-again';
+
+describe('redirects middleware', () => {
+  let res, req, next;
+  let pathToRedirectsMiddleware = '../../../../../website/server/middlewares/redirects';
+
+  beforeEach(() => {
+    res = generateRes();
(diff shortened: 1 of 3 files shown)


paglias (approved):

paglias:
@SabreCat looks good to me! I've not merged it myself in case you wanted to add something else but for me it can be merged whenever you want 👍

PR #8597, opened 2017-03-24
New default background

SabreCat:

### Changes

To support revisions to the onboarding flow on Android, we are adding a new Violet background.

* Adds `violet` to the plain background set
* Grants this background to all new users and equips it by default
* ~Adds a migration to award the above to existing users~
* Tweaks the description of the Plain Background Set to account for the fact that it now starts with one background unlocked
* Hides the description of the Plain Background Set once all of its components are unlocked, like other background sets
* Changes line endings in `background.jade` to Unix style


migrations/violet_background.js
```diff
@@ -0,0 +1,35 @@
+var migrationName = 'violet_backgrounds.js';
+
+/*
+ * This migration will find users with unlimited subscriptions who are also eligible for Jackalope mounts, and award them
```
khipkin: Update comment?

website/common/script/content/appearance/backgrounds.js
```diff
@@ -480,6 +480,11 @@ let backgrounds = {
     },
   },
   incentiveBackgrounds: {
+    violet: {
```
TheHollidayInn, line 483: Ian this the appropriate place for this background?

website/server/models/user/hooks.js
```diff
@@ -96,6 +96,8 @@ function _setUpNewUser (user) {
     user._ABtests.checkInModals = '20161221_showCheckInPreviews'; // show 'preview' check-in modals
   }
   user.items.quests.dustbunnies = 1;
+  user.purchased.background.violet = true;
```
TheHollidayInn, line 99: We should add a test for this

website/views/options/profile/backgrounds.jade
```diff
@@ -1,13 +1,13 @@
...
-      button.customize-option(ng-repeat='bg in set.items', type='button', class='background_{{bg.key}}', ng-class='user.purchased.background.{{bg.key}} ? "background-unlocked" : "background-locked"', ng-click='unlock("background." + bg.key)', popover-title='{{bg.text}}', popover='{{bg.notes}}', popover-trigger='mouseenter')
-        i.glyphicon.glyphicon-lock(ng-if='!user.purchased.background[bg.key]')
-
-script(type='text/ng-template', id='partials/options.profile.backgrounds.html')
-  +backgrounds()
+mixin backgrounds(mobile)
+  div(class=mobile ? 'padding' : 'container-fluid')
+    li.customize-menu(ng-repeat='set in backgroundShopSets')
+      menu(label='{{set.text}}')
+      div(ng-if='set.identifier === "incentiveBackgrounds" && !ownsSet("background", set.items)')=env.t('incentiveBackgroundsUnlockedWithCheckins')
```
TheHollidayInn: We should abstract this if logic to the controller

TheHollidayInn (commented):
Looks good. I think the migration is still in progress?

migrations/violet_background.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+import Bluebird from 'bluebird';
+
+import { model as User } from '../website/server/models/user';
+
+async function addVioletBGs () {
+  let userCount = 0;
+  let promises = [];
+  let cursor = User.find({
+    'purchased.background.violet': {$ne: true},
```
Alys: I think this search will be very slow because there's no index for it. Is a query needed here? If the script has to be run more than once, it won't hurt to reassign the background to any users who have it (apart from the time take to assign it unnecessarily, but that might be less than from running a slow query). Or you could query for the migration name, as long as no other migrations are run until you're certain all users have the background.

paglias: I agree with alys, it seems the query is not necessary, you could simply update all users with 

```
User.update({}, {$set: ...}).exec()

**Review by paglias — DISMISSED** [review]

File: migrations/violet_background.js (hunk 1 above)
Comment (SabreCat) [reply migrations/violet_background.js]: Hah, yup. Done locally, will be updated next push.

File: website/common/script/content/appearance/backgrounds.js (hunk 2 above)
Comment (SabreCat) [reply website/common/script/content/appearance/backgrounds.js:483]: Arguably? Note that the set is labeled "Plain Background Set", which this is very much another of.

File: website/server/models/user/hooks.js (hunk 3 above)
Comment (SabreCat) [reply website/server/models/user/hooks.js:99]: Aye aye!

File: migrations/violet_background.js (hunk 5 above)
Comment (SabreCat) [reply migrations/violet_background.js]: OK, got it!

File: website/views/options/profile/backgrounds.jade (hunk 4 above)
Comment (SabreCat) [reply website/views/options/profile/backgrounds.jade]: Good call. I'll see what I can come up with.

File: website/common/script/content/appearance/backgrounds.js (hunk 2 above)
Comment (TheHollidayInn) [reply website/common/script/content/appearance/backgrounds.js:483]: Hmm, but the key is incentives. Meh, this is the problem with hierarchy.

File: migrations/violet_background.js (hunk 5 above)
Comment (TheHollidayInn) [reply migrations/violet_background.js]: Wouldn't you still need the index for User.update?

File: migrations/violet_background.js (hunk 1 above)
Comment (SabreCat) [reply migrations/violet_background.js]: Actually, I'm deleting the migration entirely, so ¯\_(ツ)_/¯

File: test/api/v3/integration/user/auth/POST-register_local.test.js [hunk 6]
```diff
@@ -71,6 +71,23 @@ describe('POST /user/auth/local/register', () => {
...
+
+      let user = await api.post('/user/auth/local/register', {
+        username,
+        email,
+        password,
+        confirmPassword: password,
+      });
+
+      expect(user.items.quests).to.not.be.empty;
+      expect(user.purchased.background).to.not.be.empty;
```
TheHollidayInn: These checks should not just check for not empty, they should check for equality.

SabreCat: I was trying to avoid having to update the tests if we tweak what gets handed out, but OK, simple enough!

TheHollidayInn (approved):

PR #8599, opened 2017-03-27
Fixed text of check-in prize when it is a set of backgrounds

### LINKED ISSUE #8580: backgrounds Check-In prize has no name in the notification modal
The check-in prize for your second day in Habitica is a set of backgrounds. When you're given that prize, the modal does not show the name of the prize:

image

It should show the name in a way similar to this:

image

I expect that this is because the backgrounds aren't content in the same sense at the other prizes so the code can't fetch their name. 

One possible fix would be to hard-code a name into `website/common/script/content/loginIncentives.js` in a new field called something like `rewardName` (the value should be translatable of course). That field would be used for only the backgrounds (but could be reused in future if we needed to give a custom name to any other prizes). The code that inserts the name into the modal could first look for the `rewardName` and if it's not specified, it could then fetch the name from the content in the same way it currently does.

_After this is fixed for the website, we should test the Android and iOS apps to see if they pick up the fix automatically (I'm not sure whether the name lookup happens in the API or in the apps). Currently in the Android app, the backgrounds are described as "null"._

Yutsuten:

Fixes #8580 

### Changes
BgReward

----

### DIFF
--- a/website/server/libs/cron.js
+++ b/website/server/libs/cron.js
@@ -165,6 +165,11 @@ function awardLoginIncentives (user) {
       count += 1;
     }
 
+    // Overwrite notificationData.rewardText if rewardName was explicitly declared
+    if (loginIncentive.rewardName) {
+      notificationData.rewardText = i18n.t(loginIncentive.rewardName);
+    }
+
     notificationData.rewardKey = loginIncentive.rewardKey;
     notificationData.message = i18n.t('unlockedCheckInReward', user.preferences.language);
   }
--- a/website/server/libs/cron.js
+++ b/website/server/libs/cron.js
@@ -157,7 +157,7 @@ function awardLoginIncentives (user) {
           notificationData.rewardText = i18n.t('potion', {potionType: notificationData.rewardText}, user.preferences.language);
         }
       } else if (loginIncentive.rewardKey[0] === 'background_blue') {
-        notificationData.rewardText = i18n.t('incentiveBackgrounds');
+        notificationData.rewardText = i18n.t('incentiveBackgrounds', user.preferences.language);
       }
 
       if (loginIncentive.reward.length > 0 && count < loginIncentive.reward.length - 1) notificationData.rewardText += ', ';
@@ -167,7 +167,7 @@ function awardLoginIncentives (user) {
 
     // Overwrite notificationData.rewardText if rewardName was explicitly declared
     if (loginIncentive.rewardName) {
-      notificationData.rewardText = i18n.t(loginIncentive.rewardName);
+      notificationData.rewardText = i18n.t(loginIncentive.rewardName, user.preferences.language);
     }
 
     notificationData.rewardKey = loginIncentive.rewardKey;
--- a/website/common/locales/en/loginIncentives.json
+++ b/website/common/locales/en/loginIncentives.json
@@ -12,5 +12,6 @@
   "totalCheckinsTitle": "Total Check-Ins",
   "checkinProgressTitle": "Progress until next",
   "incentiveBackgroundsUnlockedWithCheckins": "These backgrounds will unlock with Daily Check-Ins.",
-  "moreIncentivesComingSoon": "New check-in incentives will be added soon!"
+  "moreIncentivesComingSoon": "New check-in incentives will be added soon!",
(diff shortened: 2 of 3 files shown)


TheHollidayInn:
Thanks for submitting. I don' think we want to solve this problem by adding the rewardName explicitly. The solution should determine why the text from this section is not being called: https://github.com/Yutsuten/habitica/blob/1538326f076292ae7a9cfd76eb334ca3b9264506/website/common/script/content/appearance/backgrounds.js#L482

Yutsuten:
Okay, so we have something to discuss here.

Before continuing, I think its important to determine what is the objective of this PR. So, in this case, what is the desired output.
I can see 2 solutions. If you can see any other, please tell me.

1. Use a new and customized name created for the set of backgrounds
2. Show the name of each background received

So the output would be, respectively:

1. "You have received Plain Background Set"
2. "You have received Blue, Green, Purple, Red, Yellow"

About the code:

Currently, the reward for the second check-in is ``[api.backgrounds.incentiveBackgrounds]``. It is an object that has the 5 backgrounds inside it. The difference of this reward and the others is that the ``api.backgrounds.incentiveBackgrounds`` DON'T have a field ``text`` (which has the name of the item). It's a big object with 5 "small" objects.

For the first solution, we don't have a item that is called "Plain Background Set". It is a set, and the code can't take such "complex" name from nowhere. That's the situation where ``rewardName`` is used. If ``rewardName`` was declared, we use it as a name.

For the second solution, a common way to do this is already implemented for other rewards: put each reward in the ``reward`` array. For example, the reward for the 5th check-in is ``[api.food.Chocolate, api.food.Meat, api.food.CottonCandyPink]``. For the output, it is shown the name of each item. We can do something similar for the backgrounds.

I chose the first solution for this PR because in my opinion its better, and because it was suggested (at least this is the way I understood) by @Alys in the issue.
> One possible fix would be to hard-code a name into ``website/common/script/content/loginIncentives.js`` in a new field called something like ``rewardName`` (the value should be translatable of course). That field would be used for only the backgrounds (but could be reused in future if we needed to give a custom name to any other prizes). The code that inserts the name into the modal could first look for the rewardName and if it's not specified, it could then fetch the name from the content in the same way it currently does.

Alys:
@Yutsuten Yes, you understood my suggestion correctly.

I think it's a clean approach for a situation where an accurate name can't be pulled from the content data. 

Another approach would be to list all the backgrounds individually as we do for food, but then we'd need code to suppress the display of the images of all but one background, and the names wouldn't make a lot of sense ("Blue, Green, Purple, Red, Yellow") so we'd have to manually add the word "background" (the new users who are receiving this would not recognise the images as backgrounds), and we'd also have to suppress listing the new Violet background that will be part of the incentives set in the content data when that feature is released. `rewardName` is simpler.

TheHollidayInn:
Okay, I agree with the use case. Let's continue with this approach.

website/server/libs/cron.js
```diff
@@ -165,6 +165,11 @@ function awardLoginIncentives (user) {
       count += 1;
     }
 
+    // Overwrite notificationData.rewardText if rewardName was explicitly declared
+    if (loginIncentive.rewardName) {
+      notificationData.rewardText = i18n.t(loginIncentive.rewardName);
```
paglias: can you pass user.preferences.language as the second parameter here? Otherwise it'll always be in English. And since you're there can you also add it to the i18n.t call at line 160?

paglias (changes requested):

website/server/libs/cron.js
Yutsuten: Sure

paglias (approved):

Alys:
I've tested this as described below and it's working. I have not reviewed the code.

1. Create new account. Leave its language set to `en`. 
2. Use the +1 missed days debug option to cron a few times, examining the check-in prize modal each time.
3. Prizes names were correct, including the new name for the background set.
4. Create a new account. Change language to `fr` (which has a translation for the set name).
5. Repeat 2 but in French.
6. Prize names were correct, including the French translation of the background set.

TheHollidayInn:
Code looks fine to me.

Alys:
Thank you @Yutsuten! I've noted this towards your tier progression.

PR #8605, opened 2017-03-28
adjust gem gift modal to encourage reporting players who beg for gems

Alys:
This PR adds a message to the gem gift modal to state that gem gifting is never required and that begging is a violation of the guidelines and should be reported.

It is something that the mods and @lemoness and @veeeeeee were discussing after we discovered that a player had harassed others into giving him gems. Some of the players had thought that giving gems was mandatory to join his party.

I'll be asking the mods for feedback so don't merge yet (I'm also happy to take feedback from the other admins of course).

In the screenshot below, the only new thing is the grey message just below the "From Balance" and "Purchase" buttons.
 
image

This PR also changes the config.json.example file to replace the example email addresses with the real ones. This allows us to easily tell from local install screenshots whether the correct address variable has been used. I can't see any disadvantage to it - emails aren't sent from local installs, and even if they ever are for a specific testing purpose, I think the benefit of seeing the correct address in screenshots outweighs the small inconvenience from occasional test email spam.

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -94,8 +94,8 @@
     },
     "ITUNES_SHARED_SECRET": "aaaabbbbccccddddeeeeffff00001111",
     "EMAILS" : {
-        "COMMUNITY_MANAGER_EMAIL" : "community@example.com",
-        "TECH_ASSISTANCE_EMAIL" : "tech@example.com",
-        "PRESS_ENQUIRY_EMAIL" : "press@example.com"
+        "COMMUNITY_MANAGER_EMAIL" : "leslie@habitica.com",
+        "TECH_ASSISTANCE_EMAIL" : "admin@habitica.com",
+        "PRESS_ENQUIRY_EMAIL" : "leslie@habitica.com"
     }
 }
--- a/website/common/locales/en/groups.json
+++ b/website/common/locales/en/groups.json
@@ -155,6 +155,7 @@
   "sendGiftPurchase": "Purchase",
   "sendGiftMessagePlaceholder": "Personal message (optional)",
   "sendGiftSubscription": "<%= months %> Month(s): $<%= price %> USD",
+  "gemGiftsAreOptional": "Please note that Habitica will never require you to gift gems to other players. Begging people for gems is a <strong>violation of the Community Guidelines</strong>, and all such instances should be reported to <%= hrefTechAssistanceEmail %>.",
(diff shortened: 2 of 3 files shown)


paglias (approved):

Alys:
The mods and staff haven't requested any changes for this so it's good to merge. I'll do that in a couple of days if no one does it sooner or reports problems with it.

config.json.example
```diff
@@ -94,8 +94,8 @@
     },
     "ITUNES_SHARED_SECRET": "aaaabbbbccccddddeeeeffff00001111",
     "EMAILS" : {
-        "COMMUNITY_MANAGER_EMAIL" : "community@example.com",
-        "TECH_ASSISTANCE_EMAIL" : "tech@example.com",
-        "PRESS_ENQUIRY_EMAIL" : "press@example.com"
+        "COMMUNITY_MANAGER_EMAIL" : "leslie@habitica.com",
+        "TECH_ASSISTANCE_EMAIL" : "admin@habitica.com",
+        "PRESS_ENQUIRY_EMAIL" : "leslie@habitica.com"
```
SabreCat, line 99: I'm not a fan of this. The example file isn't supposed to hold production configs, but placeholders for local development. And if we're checking to see if strings use appropriate variables, it's better that these be something we'd never have hardcoded in production. With it this way, in a screenshot there's no way to distinguish between a string that was written with "email leslie@habitica.com" vs. one that's written with "email <%= communityManager %>" or what have you -- the configuration would make them look identical.

Alys: Yes, in the screenshot you can't distinguish how it's been written in the code, but we have to do code reviews anyway for all PRs so we can check for hard-coded email addresses then. Viewing a screenshot is never a fully reliable way to ensure code has been written correctly.

It's more important that the screenshots show exactly what the text will look like in production. The screenshots will sometimes need to be approved by moderators or staff who don't have experience with reviewing the code, and they will want to see the real email address to do a full visual review and to be certain the text is correct. The screenshots will also sometimes also be used by Socialites and wiki editors for understanding new features and documenting them, including copying the screenshots to the wiki, and having fake email addresses in them will be confusing and could lead to misunderstandings.

SabreCat: OK, I get the reasoning now. I still think it's a misuse of the example config file, though. If we need contributors to furnish screenshots with canonical addresses, they can update their local config.json.

PR #8606, opened 2017-03-28
add recent Grand Gala seasonal special equipment set names

Alys:
This PR adds the set names for recent Grand Gala seasonal gear. These names won't be needed on the website for 3 to 12 months, but they are in the wikis already. If they're added to Transifex now, it will help the translators keep the wiki consistent with the future website content.

@SabreCat Is this okay with you? I've added the season and year to the string keys, which isn't consistent with what's been done for earlier entries, but I was having a heck of a time keeping all the gear straight in my head until I started doing that. I suspect it will help the translators too.


SabreCat:
Yup, that'll work! And good call on the year/season tags. I wish we could update the old names without making a mess in Transifex, but alas :\

Issue #8627, opened 2017-03-30
approved Group Plans task gives XP,GP,MP, then subtracts it, then re-gives it


Alys:
When a complete Group Plan task requiring approval is approved, and the user who completed the task syncs, they see notifications for XP, GP, MP being given, then notifications for the same amounts being taken away, and then a third set of notifications for the same amounts being given again. They end up with the right stats but it looks buggy.

It's possible that this does not happen every time but in my recent testing it's been consistent.

Steps to reproduce:

1. Join a group plan
2. Claim a task requiring approval
3. Complete the task
4. Just before the leader approves the task, sync to be sure there's no pending notifications for XP, etc
4. Wait for the leader to approve the task
5. Sync. You'll see notifications for XP, GP, and MP being gained and at the same time notifications for the same amounts being lost.
6. Sync again. You'll see notifications for XP, GP, and MP being gained.

Yutsuten:
I think it is related to #8582, I'll check it...

Yutsuten:
I guess this issue can be closed now.

PR #8629, opened 2017-03-31
Fix bugs with approved tasks in a Group Plan

### LINKED ISSUE #8582: multiple Group Plan To-Dos approved together don't give rewards automatically
If a group plan member completes two or more To-Dos that require approval and the group leader approves them all at the same time, only one of the tasks is correctly handled in the member's task list (it gives XP, etc and disappears). The other approved tasks remain in the member's list and have to be marked completed again manually to make them give XP, etc and disappear.

Steps to reproduce:

1. Group leader creates two or more To-Dos in the group plan, all requiring approval.
2. Group member claims the To-Dos.
3. Member marks all To-Dos as complete (To-Dos correctly remain visible and are labelled as awaiting approval).
4. Leader approves all the tasks at the same time.
5. AFTER all are approved, the member syncs with these results:
   - The member sees notifications for XP, GP, etc for just one task.
   - Sometimes the member sees a "notification not found" error message.
   - Only one task disappears from the active To-Do list.
   - The remaining approved tasks are no longer marked as awaiting approval.
   - The member can tick the remaining approved tasks to gain the XP, etc and to make them disappear from the active list. (It should not be necessary for the member to do that.)

Yutsuten:

Fixes #8582 #8627

### Changes
This PR aims to fix bugs with approved tasks in a Group Plan. I haven't finished it yet.

- [x] Score **all** approved tasks
- [x] Show notifications (gold, xp, mp) for each completed task
- [x] Do not remove earned gold/xp/mp (earn, lose, earn again)
- [x] Fix "Notification not found" error message

---

I'm having a hard time with this issue... Can someone help me, please?

The leader of a group plan approves some tasks that need approval, and the member syncs Habitica.
I found out that only one of the approved tasks were being stored to call ``User.score(...)``. I changed it to an array and it looks great, I can have any quantity of approved tasks in this array.

And then I created a ``for`` loop to call ``User.score(...)`` for all approved tasks in this array, which for me makes all sense. However, only the first one finishes with success. The second approved task fails with an Internal Server Error. I checked if the task id used is correct, and it is.

This is the server output:

For the first task

```
POST /api/v3/tasks/{ID of first task}/score/up?userV=59 200 167.910 ms - -
```

For the second task

```
2017-03-31T03:25:49.910Z - error: Error
at (etc)
{ method: 'POST',
  originalUrl: '/api/v3/tasks/{ID of second task}/score/up?userV=59',
  headers: 
   { host: 'localhost:3000',
     connection: 'keep-alive',
     'content-length': '2',
     origin: 'http://localhost:3000',
     'x-user-timezoneoffset': '180',
     'user-agent': 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/56.0.2924.76 Chrome/56.0.2924.76 Safari/537.36',
     'content-type': 'application/json;charset=UTF-8',
     accept: 'application/json, text/plain, */*',
     'x-api-user': '4be3f433-5d29-424f-b409-66adeaabeec6',
     'x-client': 'habitica-web',
     referer: 'http://localhost:3000/',
     'accept-encoding': 'gzip, deflate, br',
     'accept-language': 'en,en-US;q=0.8,ja;q=0.6' },
  body: {},
  httpCode: 500,
  isHandledError: false,
  fullError: 
   { message: 'No matching document found for id "{ID of the member who had the task approved}"',
     name: 'VersionError' } }
POST /api/v3/tasks/{ID of second task}/score/up?userV=59 500 257.342 ms - 89
```

I can't understand this error message... Why the server wouldn't find a user ID when it just did it on the first task?

----


TheHollidayInn:
You won't be able to for loop the User.score, because it sends concurrent requests to update a user which leads to the version issue. 

You will need to abstract the score logic and create a bulk score route.

paglias:
Or send each score request after the previous one is complete

Yutsuten:
I see... I tried to send a score after the previous request, but I can't find a way to know when the first score finished...
@paglias Will I break many things if I change the ``User.score`` to async? Do you see any other way to only continue my loop when the first score have finished?

paglias:
@Yutsuten the problem is that you cannot use `async` in client-old files because that's not yet supported by most browsers (it's a new feature) and we don't compile client-old with Babel.

What you can do is have User.score accept a callback that gets called when it'd done or return a promise (using angular.js $q https://code.angularjs.org/1.3.20/docs/api/ng/service/$q as the native Promise too isn't supported in some browsers)

Yutsuten:
Thank you for the suggestions @paglias !! I'll try a callback...

Yutsuten:
Oh, feels like I ignored @TheHollidayInn , sorry for that. I'll consider your idea too! >.<

Yutsuten:
Now it is mostly working, tested 2 to-dos, 2 dailies and 1 habit at once (all approved at the same time), and when I synchronized, all notifications and tasks scoring worked perfectly.

But sometimes I see an alert of "Notification not found.", or gold/xp "removed" from the user (if sync again, it is added back).

I need to investigate why those still happens sometimes...
Any guess is very welcome! :)

Yutsuten:
I've been looking the code for hours, but I can't figure out why sometimes appears a "Notification not found" error message, and why gold/xp is added and removed simultaneously sometimes... Sorry.

This image is a test I did by approving 8 tasks at once:

approvedTasks

Is the PR the way it is now acceptable? If not, I'll need help again... >.<

TheHollidayInn:
It is because you are hitting a race condition and the notification is already marked read when you are trying to read it again. You're requests need to be chained together via promises not callbacks. They are currently asynchronous and are trying to modify notifications at the same time. 

Again, the correct solution to this is to create a bulk score route.

Yutsuten:
Well, with callbacks I only call the next score when the previous finished... I see no race condition here, it is a queue waiting for resources.
And what is the difference between callbacks and promises? Both return when a function finished. By using callbacks, I just followed what @paglias suggested...

About the notifications, I haven't touched them. I just created an array of tasks to be scored. Could you tell me where I try to read notifications that are already read?

About the "bulk score route" you're telling me since the beginning, what exactly does it mean? I should create a function in the API to deal with an array of tasks to score? Something like ``api.readNotifications`` in ``website/server/controllers/api-v3/notifications.js`` line 55?

TheHollidayInn:
Every time a response is returned, angular has an interceptor and the notifications process anything in the response object. I'm assuming that is the issue you are hitting, although, I'd have to try locally to see what is happening.

For the bulk route, yes, you would need to create an api route (kind of similar) to the bulk notifications. I mean, the theory is the same, nothing else. I would just copy the score route, and update it to handle multiple scores.

Yutsuten:
Got it. I'll try doing as you say. Thank you for the patience.

TheHollidayInn:
Not a problem. Thanks for working on this!

Yutsuten:
@TheHollidayInn I'm creating an ``api.scoreTasks`` in the server, which to begin I copied ``api.scoreTask``.
But I'm not sure about what to do with the end of its code...　To be more exact, I'm not sure about what should be inside the loop (of scoring tasks) and what should be outside the loop.

There are ``user.save()``, ``task.save()``, ``res.respond(200, resJsonData)``, ``taskScoredWebhook.send`` and some code for ``task.challenge`` from line 568 to 610: https://github.com/HabitRPG/habitica/blob/develop/website/server/controllers/api-v3/tasks.js#L568-L610

I think that everything related to the task (``task.save()``, ``taskScoredWebhook.send`` and the big ``if`` statement for ``task.challenge``) must be run every iteration of the loop, but is it alright? Some of those code are run after the success response in the ``api.scoreTask``.

And how about the ``user.save()``? May I run it once outside the loop? Or it should be inside the loop with the ``task.save()``?

And what is the expected ``res.respond`` value to be sent to the client? An array with every JSON created from ``results[0]``, or a single JSON calculated only after the loop, using the last output of the ``results`` array?

Yutsuten:
I created an ``api.scoreTasks`` which scores an array of tasks.

But when I approve 2+ tasks simultaneously, all the gold and xp are summed into a single notification... "Notification not found" error message still appears sometimes.

When I approved 5 tasks simultaneously, the gold and xp was given, subtracted, and given again...

Those values are being updated when ``$window.habitrpgShared.ops.scoreTask`` is run, but looks like I'm running it too fast. I think that changing xp/gp values too fast is the main problem. There are whatches for xp/gp in ``notificationCtrl``, but they're slow.

---

What should I do now @TheHollidayInn ? Creating a bulk score route looks more efficient, but the one I did with callbacks was nearer to the expected behaviour. What do you think?

Yutsuten:
I rolled back my last commit locally, and was able to solve the problem with xp/gp being added, removed, and added again. I just added a setTimeout with 200ms delay between each call to ``user.score``.

With the delay I noticed that the "Notification not found" isn't related to my loop of scoring tasks. The problem happens just before it runs my code, at line 167 of notificationCtrl.js.
https://github.com/HabitRPG/habitica/pull/8629/files#diff-293d30fc259e779cb3655c7ce5b91ed0R167

---

I was looking the reason my last commit failed on the tests, and looks like I cannot use ``await`` inside a loop...

website/server/controllers/api-v3/tasks.js
```diff
@@ -611,6 +611,111 @@ api.scoreTask = {
   },
 };
 
+api.scoreTasks = {
+  method: 'POST',
+  url: '/tasks/score',
+  middlewares: [authWithHeaders()],
+  async handler (req, res) {
+    req.checkBody('tasks', res.t('tasksRequired')).notEmpty();
```
paglias: this should use an apiMessage (see website/server/libs/apiMessages)

website/server/controllers/api-v3/tasks.js
```diff
@@ -611,6 +611,111 @@ api.scoreTask = {
...
+
+    let user = res.locals.user;
+    let resJsonData = [];
+
+    for (let i = 0; i < req.body.tasks.length; i++) {
```
paglias: all this code is mostly copied from the scoreTask route, it should be abstracted to avoid duplication

paglias (changes requested):

paglias:
@Yutsuten you can use await in a for loop adding an eslint exception, the error is there so that people realize await in a for loop run each iteration sequentially, not in parallel

Yutsuten:
To abstract the scoreTask, I would create a function on ``website/server/libs/taskManager.js`` with the code that is similar to scoreTask and scoreTasks routes? And then in that function I should use an eslint exception for all ``await`` I need?

Sorry for making a lot of questions, lately I'm learning a lot about the habitica code and some coding concepts that I wasn't familiar.

website/server/controllers/api-v3/tasks.js
Yutsuten: I should add my message in the ``const messages`` object?

Yutsuten:
@paglias I'm having big problems with this bulk score route...

In the client, I cannot send separated notifications of gold/xp for each task, they appear summed into one no matter what I do... I tried calling ``$window.habitrpgShared.ops.scoreTask`` for each task with delay, but it isn't working. I think it's because the ``user`` object gets updated after bulk scoring the tasks.

With all gold/xp showing together, how will I deal with critical hits, drops or boss damage? Is this the desired output?

In my opinion this is getting way too complicated... May I undo the bulk score route and continue the solution that I was calling user.score for each task using callback? That one is working almost perfectly...

TheHollidayInn:
I think the aggregating exp and gold is fine. I'd prefer it over the many separate notifications.

TheHollidayInn:
Abstracting to the taskManger is a good choice. 

You should abstract to the task manager, add unit tests, then create use the multiple score function in the score task rotue or create a separate route. It is definitely a difficult fix, but it is correct over the front end hack.

Yutsuten:
On the server side, this solution looks okay.

But how about the client? There I only have functions for completing single tasks, not multiple of them at the same time.
Those are:

``$window.habitrpgShared.ops.scoreTask``
``taskScored`` (I created it from user.score: https://github.com/HabitRPG/habitica/pull/8629/files#diff-ac0d6eec6b07fc4c8ae5d5015a4ad488R199 )

The ``$window.habitrpgShared.ops.scoreTask`` receives the user object and the task as parameters. Even if 5 tasks were scored, the first one that I use to run here already give the xp/gold for all the 5 tasks (which makes no sense). When I finish scoring all the 5 scores, I have the earn-remove-earnAgain xp/gold problem. How should I deal with it? I wasn't able to find a solution for this, and I garantee you that I tried for hours. I just can't understand this behaviour.

And how about critical hits, drops or boss damage? Should it show a notification of critical hit for each task or the whole? Should I sum the quest drops into a single notification? Should I sum the boss damage? The current client don't have support for almost nothing of those.

If we're summing the xp and gold, should do it for everything else.

I do not mind creating the bulk score route in the server, but it is a waste to make so many modifications in the code of client-old, which will be discarded later. Want me to do it on the new client? I would be happy to do so, but not on client-old.

I have the "front end hack" almost working perfectly. How about finishing with it as a temporary solution and creating another issue and PR to improve it with the bulk score route? At least the (paid) users that are being affected by issues #8582 and #8627 would be happy to see the problems were fixed.

paglias:
I think @Yutsuten is right and we should consider the front end hack for now. What do you think @TheHollidayInn ?

Yutsuten:
I've got pretty busy lately, but the only bug I didn't resolved yet is the "Notification not found" error message that appears sometimes...

Yutsuten:
I found out the reason that was making the "Notification not found" error to appear.

The function ``handleUserNotifications`` is called a lot of times in a short period of time, and sometimes it is called twice before the command ``User.readNotifications`` finishes in the first call. As a result, the probability of the "Notification not found" error to appear is bigger if there are a lot of notifications about approved tasks to be read.

The best solution, in my opinion, is to not let ``handleUserNotifications`` be called before ``User.readNotifications`` has finished, but in the client I have found not way to do this...

So for now I created a local array named ``alreadyReadNotification``. Everytime that there is a notification of type "SCORED_TASK", I add the ID of the notification into this array. If the ID already exists in the array, I change the ``markAsRead`` variable to false, so it won't ask the server to mark it as read (again).

During my tests, I added a code to log the IDs of the notifications being read, and a message when my code found repeated IDs (would be the moment that the "Notification not found" would appear).

AlreadyRead

The disadvantage of this solution is that this array will only get bigger... If it is not acceptable, do anyone have ideas of not letting ``handleUserNotifications`` function to be called before ``User.readNotifications`` finishes?

If it is fine, I think I finished fixing the issues.

Alys:
I've tested locally and this fixes both #8582 and #8627. To test, I followed the steps to reproduce that I wrote in both issues and the results were that the incorrect behaviour no longer occurs. 

While testing, I discovered this bug: https://github.com/HabitRPG/habitica/issues/8676 but I've confirmed that it happens in production too and does not need to be fixed as part of this issue. I'm mentioning it here only in case someone else tests and also finds it.

I have not reviewed the code.

paglias:
This looks good to me, @TheHollidayInn ?

TheHollidayInn:
Yea, I think it is a fine intermediate step. I really appreciate your work on this @Yutsuten!!

TheHollidayInn:
Thanks again! Added to your tier progress :D

PR #8635, opened 2017-04-02
minor spelling error fixed

rbaumert:

Fixes put_issue_url_here

### Changes

----
UUID:

### DIFF
--- a/website/common/locales/en/challenge.json
+++ b/website/common/locales/en/challenge.json
@@ -17,7 +17,7 @@
     "selectWinner": "Select a winner and close the challenge:",
     "deleteOrSelect": "Delete or select winner",
     "endChallenge": "End Challenge",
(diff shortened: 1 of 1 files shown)


Alys:
I'm sorry but I'm afraid we can't accept this change. The `challengeDiscription` key is certainly spelled incorrectly but it's only visible in the code, not to the end users, so it's not critical to have it fixed. It's a key in a locales string, and the same key is used in the other languages that Habitica supports, which means that if this one were changed, the translators would have to change it in their versions too. The program code that uses this key would also have to be modified. It's not really worth making that many changes to fix this spelling mistake. 

However if you're looking for other simple changes to make, there's four in our [entry level coding](https://github.com/HabitRPG/habitica/labels/type%3A%20entry%20level%20coding) label. If you have questions about any of them or would like to work on one, comment in the issue.

PR #8637, opened 2017-04-02
use new email template when joining a group plan for customisation of subscription cancellation information

Alys:
This PR makes various changes related to sending emails when subscriptions are cancelled and/or group plans are created / joined:

- Adds the `cancellationReason` option to the cancelSubscription() / subscribeCancel() functions to pass information through to the code that controls what kind of email is sent when a user's subscription is cancelled.
- Uses that option to prevent sending of a subscription cancellation email when a user's subscription is replaced by the free subscription from a group plan.
- Adds the `PREVIOUS_SUBSCRIPTION_TYPE` Mandril "conditional merge tag" to the config for the email that the user receives when they join a group plan (the `group-plan-join` template). This allows the email template to give the user precise information about what has happened to their previous subscription. (The plain-text Mandrill template for that email is included below for comparison.)
- Uses that conditional merge tag to make the `group-plan-join` email tell the user to cancel their own iOS / Android subscription.
- Adjusts the text that is sent to an admin when the user has an iOS / Android subscription since the admin no longer needs to email instructions to the user.
- Fixes typos in comments and test descriptions; makes other small text changes for consistency.

For reference, below is the plain-text version of the Mandrill template for the `group-plan-join` email.

```
Dear *|RECIPIENT_NAME|*,

We just wanted to let you know that *|LEADER|* has enrolled you in the group "*|GROUP_NAME|*"!

As a member of a premium group plan, you'll have access to the following benefits:
- Access to the group's shared task dashboard and chatroom
- The ability to claim a task that you're working on
- The ability to be notified when a task has been assigned to you
- The ability to see which tasks have been assigned to other group members
- An exclusive Jackalope Mount for your avatar
- A free subscription with all the normal subscription benefits!

*|IF:PREVIOUS_SUBSCRIPTION_TYPE=normal_subscription|*
Since you were already an awesome subscriber, we wanted to let you know that we've switched your existing subscription over to a free Group Plan subscription and cancelled your ongoing payments. Don't worry, your benefits for consecutive subscriptions will seamlessly transfer over, and any unused subscription credits will be saved for your use if you leave the group plan.

*|ELSEIF:PREVIOUS_SUBSCRIPTION_TYPE=Google_subscription|*
Since you were already an awesome subscriber through Google Play, there's an extra action that you'll need to take to get your free subscription. Habitica can't automatically cancel your preexisting subscription for you, so you'll need to:
- go to the "My apps &amp; games" > "Subscriptions" section of the Google Play Store app to cancel your subscription.
- respond to this email so we know to set you up with your free subscription. Don't worry, your benefits for consecutive subscriptions will seamlessly transfer over, and any unused subscription credits will be saved for your use if you leave the group plan!

*|ELSEIF:PREVIOUS_SUBSCRIPTION_TYPE=iOS_subscription|*
Since you were already an awesome subscriber through the App Store, there's an extra action that you'll need to take to get your free subscription. Habitica can't automatically cancel your preexisting subscription for you, so you'll need to follow Apple's official instructions at https://support.apple.com/en-us/HT202039 and then respond to this email so we know to set you up with your free subscription. Don't worry, your benefits for consecutive subscriptions will seamlessly transfer over, and any unused subscription credits will be saved for your use if you leave the group plan!

*|ELSEIF:PREVIOUS_SUBSCRIPTION_TYPE=lifetime_free_subscription|*
Because you already had a lifetime free subscription, your subscription has not been changed.

*|ELSEIF:PREVIOUS_SUBSCRIPTION_TYPE=unknown_type_of_subscription|*
Due to the nature of your subscription, it has not been changed. Please reply to this email to ask for futher details.

*|END:IF|*
More benefits will be available in the future as we continue to develop Group Plans. Definitely feel free to reach out if you have any feedback!

Warmly,
{etc}
```

### DIFF
--- a/test/api/v3/unit/libs/stripePayments.test.js
+++ b/test/api/v3/unit/libs/stripePayments.test.js
@@ -683,6 +683,7 @@ describe('Stripe Payments', () => {
           groupId: undefined,
           nextBill: currentPeriodEndTimeStamp * 1000, // timestamp in seconds
           paymentMethod: 'Stripe',
+          cancellationReason: undefined,
         });
       });
 
@@ -702,6 +703,7 @@ describe('Stripe Payments', () => {
           groupId,
           nextBill: currentPeriodEndTimeStamp * 1000, // timestamp in seconds
           paymentMethod: 'Stripe',
+          cancellationReason: undefined,
         });
       });
     });
--- a/website/server/libs/payments.js
+++ b/website/server/libs/payments.js
@@ -131,7 +131,7 @@ api.addSubToGroupUser = async function addSubToGroupUser (member, group) {
 
     if ((ignorePaymentPlan || ignoreCustomerId) && !customerHasCancelledGroupPlan) return;
 
-    if (member.hasNotCancelled()) await member.cancelSubscription();
+    if (member.hasNotCancelled()) await member.cancelSubscription({reason:'joined_group_plan'}); // TODO make joined_group_plan a CONST
 
     let today = new Date();
     plan = member.purchased.plan.toObject();
@@ -414,7 +414,7 @@ api.cancelSubscription = async function cancelSubscription (data) {
   let emailType = 'cancel-subscription';
   let emailMergeData = [];
 
-  //  If we are buying a group subscription
+  //  If we are cancelling a group subscription
   if (data.groupId) {
     let groupFields = basicGroupFields.concat(' purchased');
     group = await Group.getGroup({user: data.user, groupId: data.groupId, populateLeader: false, groupFields});
@@ -435,6 +435,7 @@ api.cancelSubscription = async function cancelSubscription (data) {
     await this.cancelGroupUsersSubscription(group);
   } else {
     plan = data.user.purchased.plan;
+    if (data.reason && data.reason === 'joined_group_plan') emailType = 'move-subscription-to-free-group-plan'; // TODO email template requires RECIPIENT_NAME and GROUP_NAME
   }
 
   let customerId = plan.customerId;
--- a/website/server/libs/payments.js
+++ b/website/server/libs/payments.js
@@ -82,11 +82,21 @@ api.addSubscriptionToGroupUsers = async function addSubscriptionToGroupUsers (gr
  * @return undefined
  */
 api.addSubToGroupUser = async function addSubToGroupUser (member, group) {
+  // These EMAIL_TEMPLATE constants are used to pass strings into templates that are
+  // stored externally and so their values must not be changed.
+  const EMAIL_TEMPLATE_SUBSCRIPTION_TYPE_GOOGLE = 'Google_subscription';
+  const EMAIL_TEMPLATE_SUBSCRIPTION_TYPE_IOS = 'iOS_subscription';
+  const EMAIL_TEMPLATE_SUBSCRIPTION_TYPE_GROUP_PLAN = 'group_plan_free_subscription';
+  const EMAIL_TEMPLATE_SUBSCRIPTION_TYPE_LIFETIME_FREE = 'lifetime_free_subscription';
+  const EMAIL_TEMPLATE_SUBSCRIPTION_TYPE_NORMAL = 'normal_subscription';
+  const EMAIL_TEMPLATE_SUBSCRIPTION_TYPE_UNKNOWN = 'unknown_type_of_subscription';
+  const EMAIL_TEMPLATE_SUBSCRIPTION_TYPE_NONE = 'no_subscription';
+
   // When changing customerIdsToIgnore or paymentMethodsToIgnore, the code blocks below for
   // the `group-member-join` email template will probably need to be changed.
   let customerIdsToIgnore = [this.constants.GROUP_PLAN_CUSTOMER_ID, this.constants.UNLIMITED_CUSTOMER_ID];
   let paymentMethodsToIgnore = [this.constants.GOOGLE_PAYMENT_METHOD, this.constants.IOS_PAYMENT_METHOD];
-  let previousSubscriptionType = 'none';
+  let previousSubscriptionType = EMAIL_TEMPLATE_SUBSCRIPTION_TYPE_NONE;
   let leader = await User.findById(group.leader).exec();
 
   let data = {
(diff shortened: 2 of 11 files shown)


website/server/models/user/methods.js
```diff
@@ -110,16 +110,17 @@ schema.methods.addComputedStatsToJSONObj = function addComputedStatsToUserJSONOb
 // To negotiate between the payment providers and the payment helper (which probably has too many responsiblities)
 // In summary, currently is is best practice to use this method to cancel a user subscription, rather than calling the
 // payment helper.
-schema.methods.cancelSubscription = async function cancelSubscription () {
+schema.methods.cancelSubscription = async function cancelSubscription (options = {}) {
```
TheHollidayInn, line 124: Can you add docs for the options

website/server/libs/payments.js
```diff
@@ -414,6 +414,7 @@ api.cancelSubscription = async function cancelSubscription (data) {
     await this.cancelGroupUsersSubscription(group);
   } else {
     plan = data.user.purchased.plan;
+    if (data.reason && data.reason === 'joined_group_plan') emailType = 'move-subscription-to-free-group-plan'; // TODO email template requires RECIPIENT_NAME and GROUP_NAME
```
TheHollidayInn: let's add a test for this

website/server/models/user/methods.js
Alys: Yes, and also tests, but as I said, this is a WIP and incomplete. I was posting it just to see if it's the best approach. I'm guessing though that you think it's okay to proceed with doing it this way?

TheHollidayInn:
Ah, sorry. I have been conditional checking for different emails, but this approach is good too. I wrote a todo that would should use the factory pattern here. So, basically any approach will be pretty hacky until we remove some of the responsibility from these functions.

Alys:
This is mostly done. I just need to add tests for iOS and Android subscriptions and I also want to do some manual tests with actual subscriptions.

In the same file that I want to add my tests to, these two iOS and Android tests haven't been filled in:
https://github.com/HabitRPG/habitica/blob/dd5a9aa6ccc85ef89c24d1dbe4d019bd0086986e/test/api/v3/unit/libs/payments/group-plans/group-payments-create.test.js#L335-L336
Just checking: there isn't anything that would stop a test for iOS or Android working, is there?

Alys:
The iOS and Android tests I've been trying to write were to make sure that a subscription cancellation email was not sent to iOS and Android subscribers, but now that https://github.com/HabitRPG/habitica/pull/8646 is done, that's no longer necessary.

EDIT: I've updated the code for the latest changes in #8646 but some tests are failing so I'm still working on them.

Alys:
This is ready for review. I've updated the top post significantly to better describe what this does now.

test/api/v3/unit/libs/stripePayments.test.js
```diff
@@ -702,6 +703,7 @@ describe('Stripe Payments', () => {
           groupId,
           nextBill: currentPeriodEndTimeStamp * 1000, // timestamp in seconds
           paymentMethod: 'Stripe',
+          cancellationReason: undefined,
```
TheHollidayInn, line 706: I'm pretty sure you don't need to define cancellation reason when it is undefined.

website/server/libs/payments.js
```diff
@@ -125,13 +130,39 @@ api.addSubToGroupUser = async function addSubToGroupUser (member, group) {
...
+        // TODO User won't get email instructions if they've opted out of all emails. See if we can make this email an exception and if not, report here whether they've opted out.
       ]);
     }
 
-    if ((ignorePaymentPlan || ignoreCustomerId) && !customerHasCancelledGroupPlan) return;
+    if ((ignorePaymentPlan || ignoreCustomerId) && !customerHasCancelledGroupPlan) {
+      // member has been added to group plan but their subscription will not be changed
+      // automatically so they need a special message in the email
+      if (memberPlan.paymentMethod === this.constants.GOOGLE_PAYMENT_METHOD) {
+        previousSubscriptionType = 'Google';
```
TheHollidayInn: Can you just used the constants here instead of setting strings?

Alys: I am nervous about the constants being changed for any reason - for example, we might change UNLIMITED_CUSTOMER_ID from `habitrpg` to `habitica` at some point. That would then cause the Mandrill email templates to stop working correctly, unless we remembered to update them as well. Specifying the strings separately feels safer, however I don't feel very strongly about it so if you want me to use the constants, I will, and I'll add a comment to the code about updating the templates if the constants are ever changed.

The code would still be a bit ugly though. We'd have to do something like this to ensure that unexpected values were never passed through to the email templates:

```
      if (memberPlan.paymentMethod === this.constants.GOOGLE_PAYMENT_METHOD || memberPlan.paymentMethod === this.constants.IOS_PAYMENT_METHOD) {
        previousSubscriptionType = memberPlan.paymentMethod;
      } else if (memberPlan.customerId === this.constants.UNLIMITED_CUSTOMER_ID || memberPlan.customerId === this.constants.GROUP_PLAN_CUSTOMER_ID) {
        previousSubscriptionType = memberPlan.customerId;
      } else {
        // this triggers a generic message in the email template in case we forget
        // to update this code for new special cases
        previousSubscriptionType = 'unknown';
      }
```

Do you want me to go ahead with that?

test/api/v3/unit/libs/stripePayments.test.js
Alys: It complains with a message like this when undefined values aren't specified:

```
  1) Stripe Payments cancel subscription success cancels a group subscription:
     expected cancelSubscription to have been called with arguments {
  groupId: "07ba459a-729a-4867-8f05-6bf84535a9b1",
  nextBill: 1493290022147000,
  paymentMethod: "Stripe",
  user: { _id: ....
    }
}
    cancelSubscription({
  cancellationReason: undefined,
  groupId: "07ba459a-729a-4867-8f05-6bf84535a9b1",
  ...
```

TheHollidayInn: Ah, I see, becase these are the call arguments. My mistake.

website/server/libs/payments.js
TheHollidayInn: Hmm. I have the same fear of the strings changing, that is why I want the constants :P

I was thinking this:
```
 if (memberPlan.paymentMethod === this.constants.GOOGLE_PAYMENT_METHOD || memberPlan.paymentMethod === this.constants.IOS_PAYMENT_METHOD) {
        previousSubscriptionType = this.constants.GOOGLE_PAYMENT_METHOD;
      } 
```

And, in theory, if we correctly use the `this.constants.GOOGLE_PAYMENT_METHOD` constant everywhere, then when it changes it one place it will change in all. Unless I am missing something.

Alys: The email template needs a distinction between iOS and Google so it would have to be `previousSubscriptionType = memberPlan.paymentMethod;`

The templates are stored in Mandrill so if the constants are ever changed, the templates would have to be updated manually.

TheHollidayInn: OKay, I see I was mixed up because you added an ||. 
 
I meant to suggest something like this:
```
if (memberPlan.paymentMethod === this.constants.GOOGLE_PAYMENT_METHOD) {
    previousSubscriptionType = this.constants.GOOGLE_PAYMENT_METHOD;
  } else if (memberPlan.paymentMethod === this.constants.IOS_PAYMENT_METHOD) {
    previousSubscriptionType = this.constants.IOS_PAYMENT_METHOD;
  } else if (memberPlan.customerId === this.constants.UNLIMITED_CUSTOMER_ID) {
    previousSubscriptionType = this.constants.UNLIMITED_CUSTOMER_ID';
  } else if (memberPlan.customerId === this.constants.GROUP_PLAN_CUSTOMER_ID) {
    previousSubscriptionType = this.constants.GROUP_PLAN_CUSTOMER_ID';
  } else {
    // this triggers a generic message in the email template in case we forget
    // to update this code for new special cases
    previousSubscriptionType = 'unknown';
  }
```
Since the previousSubscriptionType is the value of the content then it does not affect the constant `PREVIOUS_SUBSCRIPTION_TYPE` (merge tag) on Mandril.

But, it doesn't seem that you are using the previousSubscriptionType for anything other than the email content, so never mind. It is find as is for now. We will probably need to abstract these email stuff to the factory patterns I suggested above later on.

Alys: That's correct, `previousSubscriptionType` is solely to pass a specific string that Mandrill will recognise into the email content. But I do see your point about constants. I've added new ones specifically for this so it won't matter if we change the existing constants. I've also made the values that are passed to Mandrill more complex so there's less chance we'll change them by accident when changing other constants. I've pushed a commit for that.

I've adjusted the Mandrill template to match. It's not in use yet with production code so this change won't affect current emails.

TheHollidayInn:
I think we can merge this in unless anyone has further thoughts.

PR #8649, opened 2017-04-04
Require Dailies to have a Start Date

### LINKED ISSUE #7715: Start Date can be set to null / empty string on a Daily

librarianmage:

Fixes #7715 

### Changes

----
UUID:


website/views/shared/tasks/edit/repeatables.jade
```diff
@@ -1,34 +1,34 @@
...
-      label
-        input(type="radio", ng-model='task._edit.repeatsOn', value='dayOfMonth')
-        =env.t('dayOfMonth')
-      label
-        input(type="radio", ng-model='task._edit.repeatsOn', value='dayOfWeek')
-        =env.t('dayOfWeek')
-
-    //- .form-group
-    //-   legend.option-title=env.t('summary')
+fieldset.option-group.advanced-option(ng-show="task.type === 'daily'")
```
paglias: what changed in this file? Can't see from git diff

website/server/models/task.js
```diff
@@ -229,6 +229,7 @@ export let DailySchema = new Schema(_.defaults({
     default () {
       return moment().startOf('day').toDate();
     },
+    required: true,
```
paglias, line 232: Test needed

paglias (changes requested):

website/views/shared/tasks/edit/repeatables.jade
MathWhiz: I

librarianmage: I can't tell either

khipkin: Append `?w=1` to the URL to ignore whitespace changes, and you'll see that only whitespace changed.

librarianmage:
@paglias Ping!

paglias:
This looks good but it'll probably require a migration for tasks that don't have the field unless we want to give it a default value. Ideas?

paglias (approved):

website/views/shared/tasks/edit/repeatables.jade
Alys: The file has Windows line breaks in it in `develop`, so I'm guessing this change removes them and uses just Unix line breaks, which is consistent with the majority of other files in the repo.

Alys:
There's some discussion of the default value in the issue although I'm thinking now that for simplicity and speed we should just go with a default value of "today".

librarianmage:
@paglias @Alys What should I change?

Alys:
@MathWhiz sorry about the delay on your question. My feeling is that with the code that you have now, you probably don't need to change anything regarding the default date. When a new Daily is created, it is always given the correct default date (today) so that part is fine as it is. The `required: true` that you've added should prevent the date being removed. I believe the `type: Date` code that was already there should prevent someone saving a value that isn't a date. So with this PR I think the API has everything it needs.

Ideally, there'd be code on the website to give a friendly error message to the user if they try to delete the start date, but since we're rewriting the website and since deleting the start date is probably rare, I don't think it's a critical thing to add now.

librarianmage:
Does that meat that everything looks good?

paglias (approved):
Can be merged for me!

librarianmage:
Ping? I am really sorry for bugging you all, but I would really like to know if I am missing anything.

paglias:
Notated towards your next tier! Thanks and sorry again for the super long delay!

librarianmage:
Cool, thanks!

PR #8656, opened 2017-04-05
Fix android setup process

phillipthelen:
This fixes two things that caused issues with the setup process in the android app.


SabreCat (changes requested):
Looks good but for that failing lint test.

TheHollidayInn:
Didn't we have this value before? I believe I added this somewhere a while back.

TheHollidayInn:
We should also add a  test

SabreCat (approved):

PR #8660, opened 2017-04-08
Add tests for shouldDo - fixes #8585

### LINKED ISSUE #8585: more tests are needed to ensure shouldDo performs correctly for all current Dailies repeat options
We are adding new repeat options to Dailies so we need to be sure that the tests for the current options are thorough. This issue describes all the conditions that should be tested (not including tests for the new repeat behaviour, which will also be needed but they can be added with the new code).

Some of these tests will already exist but not all of them do. Please add any that I've missed or give any other comments you have.

The code for determining whether a Daily is due is in the `shouldDo` function in `website/common/script/cron.js`

When a Daily is due today but not yesterday or tomorrow:
- If the user's Custom Day Start time (CDS) is midnight and the current time is before midnight (i.e., it's not today yet), the Daily is not due.
- If the CDS is not midnight and the current time is between midnight and the CDS, the Daily is not due.
- If the current time is after the CDS but is not yet tomorrow, the Daily is due.
- If the CDS is not midnight and the current time is after midnight on tomorrow but before the CDS tomorrow, the Daily is due.
- If the CDS is not midnight and the current time is after the CDS tomorrow, the Daily is not due.
- If the CDS is midnight and the current time is tomorrow, the Daily is not due.

Tests for the Daily's Start Date:
- NB this assumes that the current time and CDS are such that the Daily would be due if the Start Date had an appropriate value.
- If the Start Date is before today, the Daily is due.
- If the Start Date is today, the Daily is due.
- If the Start Date is tomorrow, the Daily is not due.
- The Start Date is saved as a date-time in UTC but when it's converted to the user's time zone, we care about only the date portion (e.g., if the Start Date when coverted to the user's time zone is today at 9am, the Daily is due now even if the current time is not yet 9am).
- However in the above example, if the CDS is 10am, the Daily is not due now (it will be due after the current time passes 10am).

If the Daily uses the Every X Days repeat:
- it's due on the Start Date
- it's due X days after the Start Date
- it's due X*2 days after the Start Date
- but if X = 0:
   - it's not due on the Start Date
   - it's not due on the day before the Start Date
   - it's not due on the day after the Start Date
   - it's not due today

If the Daily uses the days of the week repeat:
- it's due on the days selected
- it's not due on the other days
- those should be tested with a CDS of midnight and a non-midnight CDS
- if no days of the week are selected:
   - it's not due on the Start Date
   - it's not due on the day before the Start Date
   - it's not due on the day after the Start Date
   - it's not due today

Can anyone see any cases I've missed for the current repeat behaviour?

gracesusanchen:
Fixes #8585 

### Changes

- New unit tests to cover both CDS and timezone variations while using shouldDo to check if Daily Tasks are due.
- Update `.zone` (deprecated) to `.utcOffset`

* TODO:
Need feedback on the two failing tests while testing "CDS is midnight, and should return false when current time is before CDS"

----
UUID: 8814b00a-0c50-41b1-9451-a5ad57ec3870


paglias:
@gracesusanchen thanks for this! 

For anyone interested these are the two failing tests
schermata 2017-04-11 alle 10 18 02

test/common/shouldDo.test.js
```diff
@@ -38,6 +39,136 @@ describe('shouldDo', () => {
...
+      });
+      context('Current Date is yesterday', () => {
+        it('should not be due yesterday', () => {
+          day = moment(day).subtract(1, 'days').toDate();
+          expect(shouldDo(day, dailyTask, options)).to.equal(false);
+        });
+      });
+      context('Current Date is today', () => {
+        it('returns false if current time is before midnight', () => {
+          day = moment(day).startOf('day').add(1, 'hours').toDate();
```
TheHollidayInn: Shouldn't this be `subtract(1, 'hours')`? To be before?

test/common/shouldDo.test.js
```diff
@@ -134,6 +316,115 @@ describe('shouldDo', () => {
...
+        });
+        context('Current Date is one day before the matching day', () => {
+          it('should return false', () => {
+            day = moment(day).subtract(1, 'days').toDate();
+            expect(shouldDo(day, dailyTask, options)).to.equal(false);
+          });
+        });
+        context('Current Date is on the matching day', () => {
+          it('returns false if current time is before midnight', () => {
+            day = moment(day).startOf('day').add(1, 'hours').toDate();
```
TheHollidayInn: Same here, shouldn't this be subtract to be before?

TheHollidayInn (commented):
I'm am working with this locally, but from reading everything looks good. 

I commented on the failing tests. I think "before midnight" is the same as "the previous day" right?

Also, as a style note, please add spaces at the end of your functions. So, you should have a space between each `it` function for example.

Looking great overall! Thanks so much :D

test/common/shouldDo.test.js
gracesusanchen: Ah, I think I had a misunderstanding of what CDS is midnight means.

This is what I originally thought is the CDS (today's midnight, at hour 24).


When in fact the midnight is actually at hour 0, like this


So if the task has the start date of today, it should be due today at 1am, which was what my test case was testing. Is this correct?

TheHollidayInn:
Awesome! Thanks for the update :D

TheHollidayInn:
Added to your tier list! Congratz on tier 1 :D

SabreCat:
Reverted in https://github.com/HabitRPG/habitica/commit/15d4f7d6ab35e5194424ae183f1a00720528405b . Testing in the staging environment revealed that in production-like use cases, the change from `zone` to `utcOffset` caused miscalculation of the user's current date, leading to cron occurring unexpectedly and Dailies showing due on the wrong days. @TheHollidayInn is investigating why the test suite didn't reveal the issue!

TheHollidayInn:
To be clear though, we aren't reliable able to recreate this issue, so we need to investigate to be sure. It would be save to simply use .zone for now.

gracesusanchen:
Hello, just wanted to drop by and mentioned that I would be happy to help with any questions about the test suite itself. Thanks!

TheHollidayInn:
Okay, I believe I found the issue. Basically, we use `.zone` in a few places. Mainly here: https://github.com/HabitRPG/habitica/blob/8c76ccd39bd44f247f376a5b2bb54989c8bba687/website/client-old/js/services/userServices.js#L616.

So, this is causing the issue with calculations. @gracesusanchen Can you update the PR to use `.zone` instead and make a note that we need to update the depreciation.

Then, in a separate PR, we will write integration tests and refactor `.zone.`

TheHollidayInn:
Actually, I realized that my repeatables PR has been rebased on this, so I will fix the zone issue in that PR and you will still receive commit credit. Thank you.

PR #8661, opened 2017-04-08
Fix for automatic allocation not persisting #8641

### LINKED ISSUE #8641: Automatic Allocation of attribute points cannot be switch on / off
On the User > Stats page, the checkbox for turning Automatic Allocation on and off doesn't save the user's choice.

Steps to reproduce:

1. Use a database update command to set this: `"preferences.automaticAllocation": true`
2. Sync the website and confirm that the "Automatic Allocation" checkbox is on
3. Untick the "Automatic Allocation" checkbox.
4. Observer that a sync occurs (the blue sync line appears) and the checkbox is still unticked afterwards.
5. Use a database find command to see that your setting change was NOT saved - this is still showing: `"preferences.automaticAllocation": true`
6. Sync the website. Observer that the checkbox becomes ticked.

The same thing happens in the opposite direction:

1. Use a database update command to set this: `"preferences.automaticAllocation": false`
2. Sync the website and confirm that the "Automatic Allocation" checkbox is off
3. Tick the "Automatic Allocation" checkbox.
4. Observer that a sync occurs (the blue sync line appears) and the checkbox is still ticked afterwards.
5. Use a database find command to see that your setting change was NOT saved: `"preferences.automaticAllocation": false`
6. Sync the website. Observer that the checkbox becomes unticked.

I think this bug  was introduced in https://github.com/HabitRPG/habitica/commit/98c019a0b658df1fb5f0c6db5b09b379d6cb58e2 ("Upgrade lodash to v4 and lint more files") because I can replicate it in a local install with that commit but not with the one before it (https://github.com/HabitRPG/habitica/commit/ef02e595908bcafb9b15c95383a125293f636e69).

### What to do if this happens to you

Go to the [Report a Bug](https://habitica.com/#/options/groups/guilds/a29da26b-37de-4a71-b0c6-48e72a900dac) guild (from Help > Report a Bug on the website or from the Guilds > Public Guilds on the mobile app) and state whether you want automatic allocation to be on or off. An admin will set that for you. If you need attribute points to be unallocated, say that as well.

TNychka:

https://github.com/HabitRPG/habitica/issues/8641

### Changes
- Changed stats.jade so on selection of Automatic Allocation Checkbox, user preference changes are sent in one request. 
- Commented issue on setUser function

The issue was when Automatic allocation is selected on a user's stats page, the changes are not saved. This issue was caused during an update of lodash from version 3 to 4, the root cause being previously the jade profile saved user.preferences.automaticAllocation and _then_ user.preferences.allocationMode one after the other. Some issue in lodash caused the User object not to persist between each call, so only the final call was saved (In this case automaticAllocation was lost but allocationMode was saved). The fix was to send all the changes at once, so when the Automatic allocation checkbox is selected, autoAllocate _and_ allocMode are both sent at the same time. 

---

### DIFF
--- a/website/client-old/js/services/userServices.js
+++ b/website/client-old/js/services/userServices.js
@@ -189,7 +189,7 @@ angular.module('habitrpg')
           }
         })
       }
-      
+
       function setUser(updates) {
         for (var key in updates) {
           _.set(user, key, updates[key]);
--- a/website/client-old/js/services/userServices.js
+++ b/website/client-old/js/services/userServices.js
@@ -189,9 +189,7 @@ angular.module('habitrpg')
           }
         })
       }
-
-      //Calling this multiple times rapidly will not persist changes
-      //All changes must either be sent at once, or spaced in time (0.1-0.5 sec)
+      
       function setUser(updates) {
         for (var key in updates) {
           _.set(user, key, updates[key]);
--- a/website/client-old/js/services/userServices.js
+++ b/website/client-old/js/services/userServices.js
@@ -190,6 +190,8 @@ angular.module('habitrpg')
         })
       }
 
+      //Calling this multiple times rapidly will not persist changes
+      //All changes must either be sent at once, or spaced in time (0.1-0.5 sec)
       function setUser(updates) {
         for (var key in updates) {
           _.set(user, key, updates[key]);
--- a/website/views/options/profile/stats.jade
+++ b/website/views/options/profile/stats.jade
@@ -28,7 +28,7 @@ script(id='partials/options.profile.stats.html', type='text/ng-template')
               fieldset.auto-allocate
                 .checkbox
                   label
-                    input(type='checkbox', ng-model='user.preferences.automaticAllocation', ng-change='set({"preferences.automaticAllocation": user.preferences.automaticAllocation?true: false})', ng-click='set({"preferences.allocationMode":"taskbased"})')
+                    input(type='checkbox', ng-model='user.preferences.automaticAllocation', ng-change='set({"preferences.automaticAllocation": user.preferences.automaticAllocation?true: false, "preferences.allocationMode" : "taskbased"})')
                     span.hint(popover-trigger='mouseenter', popover-placement='right', popover=env.t('autoAllocationPop'))=env.t('autoAllocation')
                 form(ng-show='user.preferences.automaticAllocation',style='margin-left:1em')
                   .radio
--- a/website/views/options/profile/stats.jade
+++ b/website/views/options/profile/stats.jade
@@ -28,7 +28,7 @@ script(id='partials/options.profile.stats.html', type='text/ng-template')
               fieldset.auto-allocate
                 .checkbox
                   label
-                    input(type='checkbox', ng-model='user.preferences.automaticAllocation', ng-change='set({"preferences.automaticAllocation": user.preferences.automaticAllocation?true: false, "preferences.allocationMode" : "taskbased"})')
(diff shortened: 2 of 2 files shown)


website/client-old/js/services/userServices.js
```diff
@@ -190,6 +190,8 @@ angular.module('habitrpg')
         })
       }
 
+      //Calling this multiple times rapidly will not persist changes
```
paglias: Are changes not persisted locally or on the server? Locally it shouldn't happen as this is a normal sync function

TNychka: What I saw was that calling it twice somehow didn't save the user object locally which was then synced. Calling the function with attribute A and then B only saves B for some reason. However if you call the function with A and B then both save.

paglias: That's strange.. it's possible that between one call and the other the server responded with the updated user that then overwrote the second call?

TNychka: That is a possibility, the two calls issue the setUser call, then an ajax request. If the ajax request overwrites the local version, then that would be the cause. I noticed that I could not replicate the issue while debugging with a breakpoint. If there was a delay (like a breakpoint) the calls probably had time to resolve before the next one was made. If the call could issue a promise or create a queue of requests, that would probably be the best solution. The comment might be unnecessary then.

EDIT: One possible solution to this might be to queue the calls 
http://stackoverflow.com/questions/12472174/asynchronous-calls-inside-loop

website/client-old/js/services/userServices.js
```diff
@@ -189,7 +189,7 @@ angular.module('habitrpg')
           }
         })
       }
-
+
```
SabreCat: There don't seem to be any functional changes committed to `userServices`, just this bit of whitespace.

website/views/options/profile/stats.jade
```diff
@@ -1,51 +1,51 @@
...
+                strong.inline
+                  |{{user.stats.points}}&nbsp;
+                strong.hint(popover-trigger='mouseenter', popover-placement='right', popover=env.t('levelPopover'))=env.t('unallocated')
+            td
+          tr
+            td(colspan=2)
+              fieldset.auto-allocate
+                .checkbox
+                  label
+                    input(type='checkbox', ng-model='user.preferences.automaticAllocation', ng-change='set({"preferences.automaticAllocation": user.preferences.automaticAllocation?true: false, "preferences.allocationMode" : "taskbased"})')
```
SabreCat: It's a little hard to tell thanks to the line-ending change (@TheHollidayInn's Windows dev environment messing with us, heh), but this should be the only line substantially altered, correct?

SabreCat (changes requested):
Remove `userServices.js` if it wasn't meant to be changed, or commit the changes it should have if it was. After that, we should be OK to merge this!

TNychka:
@SabreCat alright, cleaned up the whitespacing so now only the stats.jade file has been modified. 
And the only major change to the stats file was 
`ng-change='set({"preferences.automaticAllocation": user.preferences.automaticAllocation, "preferences.allocationMode":"taskbased"})'` at line 31
Which sends the two changes along in one request so there is no issue with the syncing to server.

SabreCat (approved):

Alys:
I've tested this locally and it's working well. Toggling the "Automatic Allocation" checkbox causes the `preferences.automaticAllocation` field to be toggled in the user's database document.

@TNychka thank you! I've awarded you tier 1. You can read more about [contributor tiers and rewards here](http://habitrpg.wikia.com/wiki/Contributor_Rewards).

PR #8664, opened 2017-04-11
Added email invite limit

TheHollidayInn:

Fixes put_issue_url_here

### Changes

This prevents users from sending more than 200 emails (and prevents spammers)

----
UUID:

### DIFF
--- a/website/common/locales/en/groups.json
+++ b/website/common/locales/en/groups.json
@@ -270,5 +270,6 @@
   "confirmCancelGroupPlan": "Are you sure you want to cancel the group plan and remove its benefits from all members, including their free subscriptions?",
   "canceledGroupPlan": "Canceled Group Plan",
   "groupPlanCanceled": "Group Plan will become inactive on",
-  "purchasedGroupPlanPlanExtraMonths": "You have <%= months %> months of extra group plan credit."
+  "purchasedGroupPlanPlanExtraMonths": "You have <%= months %> months of extra group plan credit.",
+  "inviteLimitReached": "You have already sent the max number of email invites."
 }
--- a/website/common/locales/en/groups.json
+++ b/website/common/locales/en/groups.json
@@ -148,6 +148,7 @@
   "invitationsSent": "Invitations sent!",
   "invitationSent": "Invitation sent!",
   "inviteAlertInfo2": "Or share this link (copy/paste):",
+  "inviteLimitReached": "You have already sent the maximum number of email invitations. We have a limit to prevent spamming, however if you would like more, please contact us at <%= techAssistanceEmail %> and we'll be happy to discuss it!",
   "sendGiftHeading": "Send Gift to <%= name %>",
   "sendGiftGemsBalance": "From <%= number %> Gems",
   "sendGiftCost": "Total: $<%= cost %> USD",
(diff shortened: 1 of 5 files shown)


website/common/locales/en/groups.json
```diff
@@ -270,5 +270,6 @@
   "confirmCancelGroupPlan": "Are you sure you want to cancel the group plan and remove its benefits from all members, including their free subscriptions?",
   "canceledGroupPlan": "Canceled Group Plan",
   "groupPlanCanceled": "Group Plan will become inactive on",
-  "purchasedGroupPlanPlanExtraMonths": "You have <%= months %> months of extra group plan credit."
+  "purchasedGroupPlanPlanExtraMonths": "You have <%= months %> months of extra group plan credit.",
+  "inviteLimitReached": "You have already sent the max number of email invites."
```
Alys: Can this be changed to something like:

"You have already sent the maximum number of email invitations. We have a limit to prevent spamming, however if you would like more, please contact us at admin@habitica.com and we'll be happy to discuss it!"

The email address to use is `TECH_ASSISTANCE_EMAIL`

@lemoness Do you want different wording?

Alys (changes requested):

website/server/controllers/api-v3/groups.js
```diff
@@ -27,6 +27,8 @@ import amzLib from '../../libs/amazonPayments';
 import shared from '../../../common';
 import apiMessages from '../../libs/apiMessages';
 
+const MAX_EMAIL_INVITES_BY_USER = 200;
```
khipkin, line 31: Seems fishy to establish this constant in 2 different places in the code.

khipkin:
I think it would be good to have the limit reset after a reasonable amount of time... maybe every month or so? Wouldn't even require expensive/annoying migrations to reset the counts; we could just reset them lazily:

if a user tries to send an email but has reached their cap, check the last time the cap was reset; if it was more than X time ago, reset again and update the date.

But idk. Might not be worth the hassle since 200 email invites is A LOT.

Alys:
Yes, or we could add it to cron to reset at the start of the month - there's already some monthly processing happening so it would fit in neatly there. But I'm fine with the manual email solution. We might even get some interesting feedback as part of it. We might have users we don't know about who are legitimately inviting lots of people, and flushing them out of the woodwork to thank them would be cool. :) I'm not anticipating a lot of extra work from this and a manual reset will be fast to do.

website/server/controllers/api-v3/groups.js
paglias: 👍

paglias:
Since this is pretty urgent let's add the resetting later.

website/server/controllers/api-v3/groups.js
paglias: Actually the other place where it's used is a test so I'd say it's fine

paglias:
The code LGTM, maybe let's change the wording as @Alys suggested and deploy this

Alys (approved):

Alys:
This is ready for someone to review my additions, and then if it's all okay, merge and deploy as soon as we can.

website/server/models/group.js
```diff
@@ -35,7 +35,7 @@ import stripePayments from '../libs/stripePayments';
 const questScrolls = shared.content.quests;
 const Schema = mongoose.Schema;
 
-export const INVITES_LIMIT = 100;
+export const INVITES_LIMIT = 100; // must not be greater than MAX_EMAIL_INVITES_BY_USER
```
TheHollidayInn, line 38: Why must this not be greater than the user limit? This is a limit for the group.

Alys: There's more details in the commit message and in slack but if you haven't yet reached MAX_EMAIL_INVITES_BY_USER, you can send as many invitations as you want in a single API call, and so if INVITES_LIMIT is greater than MAX_EMAIL_INVITES_BY_USER, it's a once-off way to send more than MAX_EMAIL_INVITES_BY_USER. It's a minor exploit though so not worth fixing (or at least not worth fixing now when we have to get this deployed quickly).

TheHollidayInn: Ah, yea, I saw the slack message after posting this. Thanks :D

PR #8681, opened 2017-04-18
minor text changes: Shield-Hand Item; space in quest completion modal; Rebirth clarification

Alys:
This makes unrelated changes but they're all so small I didn't think they were worth their own PRs. Each is in its own commit if you want to review them separately.

### replace Shield with Shield-Hand Item where appropriate for consistency and clarity

For example, previously, this Horn was referred to as just a shield, which it isn't. We use "Shield-Hand Item" elsewhere in similar situations.
delete

### add space in quest completion modal heading between name of quest and "Completed!"

Previously, the quest name and "Completed" were touching. Now they aren't.

image

### DIFF
--- a/website/common/locales/en/gear.json
+++ b/website/common/locales/en/gear.json
@@ -988,7 +988,7 @@
   "offhandCapitalized": "Shield-Hand Item",
 
   "shieldBase0Text": "No Shield-Hand Equipment",
-  "shieldBase0Notes": "No shield or second weapon.",
+  "shieldBase0Notes": "No shield or shield-hand item.",
 
   "shieldWarrior1Text": "Wooden Shield",
   "shieldWarrior1Notes": "Round shield of thick wood. Increases Constitution by <%= con %>.",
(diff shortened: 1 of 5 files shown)


TheHollidayInn:
Looks good to me.

Alys:
I've removed the translation fixes since they've been done in Transifex now (and I've joined Transifex myself so I can fix variable errors there now directly to speed up future bug fixes). 

I've added a minor change to the popup message for the Orb of Rebirth to clarify that it doesn't give you a new character but restarts your current one. CakeKnight pointed out in the Commonwealth of i18n guild that '"Begin a new character" makes me think I'll lose my current avatar, that I'll also have to choose a new name, when in fact you lose almost nothing.'

image

Alys:
I've added a minor text change to the latest Bailey to indicate that group managers can be added by clicking the group's edit button (not pencil icon).

image

I'll merge this on the weekend if there's no complaints (but do tell me if you want more time to look at it).

Alys:
Added one-word change for "correct typo in questVice3Completion: breath > breathe"

khipkin (approved):

PR #8699, opened 2017-04-25
Added new web profile for checkout

### LINKED ISSUE #4676: Remove shipping field from paypal process
The paypal checkout / subscription process currently asks for users' shipping address, which is unecessary. Steps for removal at https://developer.paypal.com/webapps/developer/docs/api/#payment-experience
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8537167-remove-shipping-field-from-paypal-process?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheHollidayInn:

Fixes #4676

### Changes

This allows for us to have a custom web profile for Paypal checkout: https://developer.paypal.com/docs/api/payment-experience/. 

Unfortunately, this is not yet supported with subscriptions: https://github.com/paypal/PayPal-REST-API-issues/issues/4. 

This could be a good case for Braintree since it is listed as an alternative.

----
UUID:


SabreCat:
How would you recommend running this script? I can't seem to get it to load, because some of the required files use ES2015 syntax that straight Node doesn't like. Do we need to transpile it first?

SabreCat (changes requested):
See question in comments--we should probably put a comment block at the top giving instructions on how to run this. Moreover, as written line 10 fails out because the path to `common` is no longer correct.

paglias:
@SabreCat specifically about which syntax is it complaining? If you use Node 7 (we use 6 on the server) everything should be supported except for `import/export ` which still require Babel

TheHollidayInn:
The script needs a CLI library for better interaction. I added a todo for that, but it should be a separate PR.

You just need to change the `OP` create to create-webprofle then run the script: `node [location].js`

SabreCat:
@TheHollidayInn Yup, that's what I'm doing when I hit the error.

@paglias Yeah, it's complaining about an `import`.

TheHollidayInn:
Ah, yea. Node doesn't support that syntax. I'll just update the syntax then.

SabreCat:
There's no `import` in this file, though. It's running into it at least one dependency deep, somewhere we may still want the `import`.

TheHollidayInn:
Yea, it is in common

TheHollidayInn:
I can fix this similar to migrations. (without starting up the server)

TheHollidayInn:
Here you go. Babel has some nice interpreters for dev items.

SabreCat (approved):

PR #8702, opened 2017-04-26
[WIP] Donations

### LINKED ISSUE #3995: add one-time "donate" option that doesn't grant in-game purchase
Just as important as a code issue: if I donate to developers, I can just buy all of the things I would have otherwise earned on the site. Because of this, I don't think I'd ever want to donate- it would make rewards feel meaningless. I know it feels like a good idea but don't you think there are ways to get people to donate without jeopardizing the entire point of the game? 

Aside from my personal incentive to be able to use this site meaningfully (it's SO COOL!!), I'm also concerned about that option taking away donations from developers. Easiest solution is to at least add a paypal donation link at the bottom of your page so members can at least have the option of a click donation that DOESN'T compromise the incentives system HabitRPG is built on.

Thanks!

edit: oops I'm very sorry- did not realize there was a suggestions board...
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4256412-add-one-time-donate-option-that-doesn-t-grant-in-game-purchase?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheHollidayInn:

Fixes #3995

### Changes

This adds the ability to donate a one time fee to Habitica. I am assuming this will replace the current donate button that pulls up the gems modal. 

I am looking for approval on direction before adding in Amazon and Stripe as well as before needing a code review.

screen shot 2017-04-26 at 11 15 46 am

----
UUID:

### DIFF
--- a/test/api/v3/integration/payments/paypal/GET-payments_paypal_checkout.test.js
+++ b/test/api/v3/integration/payments/paypal/GET-payments_paypal_checkout.test.js
@@ -36,5 +36,21 @@ describe('payments : paypal #checkout', () => {
       expect(checkoutStub).to.be.calledOnce;
       expect(checkoutStub.args[0][0].gift).to.eql(undefined);
     });
+
+    it('creates a purchase link with a donation', async () => {
+      user = await generateUser({
+        'profile.name': 'sender',
+        'purchased.plan.customerId': 'customer-id',
+        'purchased.plan.planId': 'basic_3mo',
+        'purchased.plan.lastBillingDate': new Date(),
+        balance: 2,
+      });
+      let donation = '5';
+
+      await user.get(`${endpoint}?noRedirect=true&donation=${donation}`);
+
+      expect(checkoutStub).to.be.calledOnce;
+      expect(checkoutStub.args[0][0].donation).to.eql(donation);
+    });
   });
 });
--- a/test/api/v3/unit/libs/payments/donation.test.js
+++ b/test/api/v3/unit/libs/payments/donation.test.js
@@ -1,0 +1,81 @@
+import nconf from 'nconf';
+
+const BASE_URL = nconf.get('BASE_URL');
+
+import payments from '../../../../../../website/server/libs/payments';
+import paypalPayments from '../../../../../../website/server/libs/paypalPayments';
+
+describe.only('One Time Donations', () => {
+  describe('Paypal', () => {
+    let approvalHerf;
+    let user = {};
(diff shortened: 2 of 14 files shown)


paglias:
The current donations also give gems, is this changing?

TheHollidayInn:
Yea, I saw that. Maybe that is the preferred way to gather gems? There is a Trello card and issue for donations, soooo idk.

paglias:
Hmm. In the original issue it seems they want a way to donate without getting gems, some maybe it could just be a checkbox under the existing donation procedure to opt out of gems?

Alys:
Do we really want this? The issue is still marked as suggestion-discussion.

People asking to donate without getting gems is rare; might be even rarer now that there's many more things to buy with gems. Anyone who really wants to donate for no benefit could buy gems and give them to a staff member. I'm not sure this code would be used often enough to make it worth writing and supporting, especially since anyone who uses this feature mistakenly would then demand a refund or to be given gems manually.

See also my comments on that issue and the ideas on the donations Trello card. Would it be better to add a gem-purchasable "item" as a way of making a donation? That wouldn't require another integration with payment processors and so would have less chance of problems.

TheHollidayInn:
I agree we probably don't need this. But, this does allow for a custom donation amount and this would solve the "Donate to contributor" problem.

I'm not sure what you mean by a gem purchasable item. If they still purchase gems what is the difference from the current flow?

Alys:
Very little difference - that's what makes it so easy. :)  No need for a new real-world-money purchase mechanism (no new payment code and tests,, no need to support all payment methods). There'd be an "item" that you'd buy with gems but nothing would be added to your account. It's just a way of deleting gems. Basically just a button that says "donate gems back to Habitica".

But my main question is would any donation mechanism be used often enough to make it worth having? 

Also would it look weird that we'd have somethng on our site that's basically asking people to give us money?

TheHollidayInn:
AH, okay. A donate gems item. I get it. 

I think it is common to have a donation button for free sites, but yea.. I'm not seeing a need for this with out current gem donation.

TheHollidayInn:
Closing because we now have a donation button. Further discussion can be continued on Trello.

PR #8705, opened 2017-04-27
Show task alias advanced option for Rewards - fixes #8668

### LINKED ISSUE #8668: Task Alias field is not visible in Rewards
When you edit a Reward, you cannot see the Task Alias field. You should be able to see it (and modify it).

You can see the Task Alias field when editing Habits, Dailies, and To-Dos, so they are all working correctly.

The Task Alias field is part of the Advanced Options, so this bug has probably happened because all Advanced Options have been hidden for Rewards. If so, the fix would be to un-hide the Advanced Options heading and Task Alias field but hide all the other advanced options that aren't relevant for Rewards.

This applies only to tasks in the user's own tasks list. The Task Alias field should not be visible when editing a challenge task or a group plan task.

savoie:

Fixes https://github.com/HabitRPG/habitica/issues/8668

### Changes

This has got to be one of my shortest PRs ever:
* Show advanced options when editing a task if the task alias form entry should be shown
* No tests because I couldn't find examples of any that describe how task modals should look; please let me know if I've missed something!

Screenshots:

**Editing reward in task list:**

editing reward in task list

**Editing reward while editing challenge:**

editing reward in challenge

**Editing reward while editing private group tasks:**

editing reward in private group

----
UUID: b58c50cd-d497-4f53-88a6-69f076ea1114

### DIFF
--- a/website/views/shared/tasks/edit/advanced_options.jade
+++ b/website/views/shared/tasks/edit/advanced_options.jade
@@ -1,4 +1,4 @@
-div(ng-if='(task.type !== "reward") || (!obj.auth && obj.purchased && obj.purchased.active)')
+div(ng-if='(task.type !== "reward") || task.userId || (!obj.auth && obj.purchased && obj.purchased.active)')
   button.advanced-options-toggle.option-title.mega(type='button',
     ng-class='{active: task._edit._advanced}',
     ng-click='task._edit._advanced = !task._edit._advanced', tooltip=env.t('expandCollapse'))


savoie:
Minor thing I noticed while testing--when editing a reward, the price disappears off of the reward entry. Is that intended? (I'll admit I didn't go digging through the code for it.) I'd expect it to stay visible and for any changes to price in the modal to apply on "Save & Close".

If that's not intended behaviour, I can look into changing it and just expand the scope of this PR ever so slightly to roll 2 rewards edit fixes into one.

TheHollidayInn:
Are you saying that when you type in the reward value the input box disappears?

savoie:
@TheHollidayInn I should just have not been lazy and added screenshots from the start.

**Modal closed**
modal closed

**Modal open**
modal open

I was suprised that the coin sprite + amount disappear when I edit the reward.

Alys:
That's deliberate and also happens with the other task types (the checkbox or +- buttons are hidden). It's to stop a task being used while it's being edited, which used to cause confusion before the edit modal existed. It's no longer necessary because we have the edit modal but it's not worth fixing now because we're making some changes to the website that will significantly modify the code that does that.

savoie:
@Alys ah that makes sense. It also sounds like it was designed to combat users like me--"Hmm I wonder if I can click this thing and break something." Thanks for the clarification! (Sorry for going pretty off-topic in this PR)

Alys:
Ha ha, pretty close to that. :) And no need to apologise - always better to know about potential bugs than to not know!

TheHollidayInn:
Congratz on Tier 1!

PR #8706, opened 2017-04-27
travis: don't pollute the log with npm output

lynxlynxlynx:
npm install causes about 7k lines of output. While travis folds that out of view almost immediately, it means not all tests fit into its web view, so you have to look at the raw log, with no color info. This should alleviate the problem and probably insignificantly speed up the install itself.


lynxlynxlynx:
eg., now test output starts at line 792, while before it did at 7239.

TheHollidayInn:
Awesome! This is great :D

TheHollidayInn:
Can you post your UUID so I can give you credit on habitica?

lynxlynxlynx:
Sure, here you go: 10c530d0-a796-476b-8a31-6f048c6e8a8c

TheHollidayInn:
Noted towards your tier!

PR #8712, opened 2017-05-02
Payments gem reset

TheHollidayInn:
### Changes

Checks for the existence of the dateUpdate in a plan. Fixes the following issue:

"If an account is missing the `purchased.plan.dateUpdated` value, then at the start of each month, cron is unable to reset their subscription. The result is that their monthly gem cap doesn't reset and the consecutive month code isn't run. "

----
UUID:


Alys:
Were a couple of those commits added accidentally? One has an "it.only" in it too.

TheHollidayInn:
Yea, it was an accident. I'll remove them

SabreCat (approved):
Tested and confirmed. Merging!

PR #8719, opened 2017-05-06
update / delete outdated README.md files

Alys:
This makes some small changes to a couple of README.md files that were outdated.

It also deletes two READMEs from the test directories. Material from them that is still relevant is on the wiki at http://habitica.wikia.com/wiki/Using_Your_Local_Install_to_Modify_Habitica's_Website_and_API where it will be easier for all contributors to help with keeping it up to date.

This PR also adds a README.md to the top-level tests directory to point to that wiki page.

If you see things on that wiki page that should be changed, either tell me what to change or edit it yourself, as you prefer.


Alys:
I'll merge this on the weekend if there's no complaints (but do tell me if you want more time to look at it).

PR #8722, opened 2017-05-09
Client/equipment

negue:
- extract strings to newClient.json
- add popovers to `use costume` and `auto equip`
- disable item selection with deactivated costume
- add error message on deactivated costume mode


website/client/components/inventory/drawer.vue
```diff
@@ -75,6 +77,24 @@
...
   overflow-x: scroll;
   white-space: nowrap;
+  position: relative;
+
+  & .message {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+
+    top: calc(50% - 30px);
```
paglias, line 100: ohh this is cool, it's going to work in all browsers?

website/client/components/inventory/equipment.vue
```diff
@@ -151,7 +162,9 @@ export default {
   },
   methods: {
     equip (item) {
-      this.$store.dispatch('common:equip', {key: item.key, type: this.costume ? 'costume' : 'equipped'});
+      if (!this.costume || this.user.preferences.costume) {
```
paglias: the starts are still going to show up with this I suppose, just not working. Maybe we should not show them at all?

paglias (changes requested):

website/client/components/inventory/equipment.vue
```diff
@@ -28,28 +28,37 @@
         b-dropdown-item(@click="groupBy = 'type'", :class="{'dropdown-item-active': groupBy === 'type'}") Type
         b-dropdown-item(@click="groupBy = 'class'", :class="{'dropdown-item-active': groupBy === 'class'}") {{ $t('class') }}
 
-    drawer(:title="$t('equipment')")
+    drawer(:title="$t('equipment')", :errorMessage="(costume && !user.preferences.costume) ? $t('costumeDisabled') : ''")
```
paglias: I would pass null instead of empty string

website/client/components/inventory/drawer.vue
```diff
@@ -6,6 +6,8 @@
       slot(name="drawer-header")
       .drawer-slider
         slot(name="drawer-slider")
+        div.message(v-if="errorMessage != ''")
```
paglias: !errorMessage should work too

paglias (changes requested):

website/client/components/inventory/drawer.vue
negue: it should http://caniuse.com/#feat=calc , back then the browsers need to have their own prefix but now all supports it

website/client/components/inventory/equipment.vue
negue: I could hide it, but then maybe users ask "why are there isn't a star" -> "bug"

website/client/components/inventory/drawer.vue
negue: `!errorMessage` didn't work, had to use ` != null `

website/client/components/inventory/equipment.vue
negue: star is now hidden

Issue #8540, opened 2017-03-07
When editing a Challenge, tasks save instantly instead of awaiting Save


SabreCat:
[//]: # (Before logging this issue, look through common problems at https://github.com/HabitRPG/habitrpg/issues If you find your issue there, read at least the first post to see if there is a workaround for you)

[//]: # (Github is primarily used for reporting bugs. If you have a feature request, use "Help > Request a Feature" so that the feature request can be vetted by the larger Habitica community)

[//]: # (To report a bug in one of the mobile apps, please report it in the correct repository. Android: https://github.com/HabitRPG/habitrpg-android, iOS: https://github.com/HabitRPG/habitrpg-ios)

[//]: # (For more guidelines see https://github.com/HabitRPG/habitrpg/issues/2760)

[//]: # (Fill out relevant information - UUID is found in Settings -> API)

### Description
[//]: # (Describe bug in detail here. Include pictures if helpful.)

When you open a Challenge for editing, there are "Save" and "Cancel" buttons at the top, suggesting you've begun an editing session that will be finalized or rejected upon clicking one of those two. However, that appears only to apply to the miscellaneous fields at the top of the Challenge (description and so on)--if you manipulate the challenge tasks at all, those changes propagate instantly.

This was some embarrassment to me when I accidentally added a nonsense Habit to an official Habitica x Take This challenge, syncing it out to all ~4000 participants despite my attempting to cancel the action. :\

Alys:
I've marked this as entry level coding because the easiest "fix" is to just as a sentence of explanation to the form, however feel free to override me and change it and make other suggestions!

I think changing the actual behaviour would be nightmare level coding. :)

Alys:
I think changing the actual behaviour would be nightmare level coding. :)

librarianmage:
Would the warning be best placed by [the Save/Cancel](https://github.com/HabitRPG/habitica/blob/develop/website/views/options/social/challenges.jade#L40) buttons?

Alys:
@MathWhiz Yeah, I think that would work. Give it a try if you wish. Just wait a day or so to see if SabreCat disagrees with my assessment of the issue. :)

librarianmage:
@Alys Oh, I wasn't planning on completing this, I was hoping to clarify

teqnickels:
@Alys, I'd like to take this if it's ok.

Alys:
@teqnickels Thank you! Comment here if you have questions or problems.

[...]

PR #8723, opened 2017-05-10
Bug in Challenge: when editing any task that edit will save instantly instead of awaiting Save or Cancel

helenyau0:

[//]: #8540 
Partial fix for https://github.com/HabitRPG/habitica/issues/8540

### Changes
When a user creates a new challenge they are able to edit any of the above fields (title, description). However, if they add or make changes to the tasks below (habits, dailies, todos, and rewards) those changes will save instantly and clicking the cancel button will have no effect in discarding them because the "+" button does its own functionality when clicking it it will make a POST. The easiest fix to this problem was to add a warning message. That warning message is placed inside all the task templates warning users right away that adding any new task will get saved automatically and clicking the cancel button above will not revert any changes.   

screen shot 2017-05-10 at 9 55 18 am

fda98da6-862b-4583-80ef-5ab81671875e
----
UUID:

### DIFF
--- a/website/views/shared/tasks/index.jade
+++ b/website/views/shared/tasks/index.jade
@@ -7,7 +7,7 @@ include ./task-list
 include ./task
 
 script(id='templates/habitrpg-tasks.html', type="text/ng-template")
-  .tasks-lists.container-fluid    
+  .tasks-lists.container-fluid
     .row
       .col-sm-6.col-md-3(
         ng-repeat='list in lists',
@@ -15,14 +15,14 @@ script(id='templates/habitrpg-tasks.html', type="text/ng-template")
         .task-column(class='{{::list.type}}s')
           include ./task_view/graph
           h2.task-column_title(class='{{::list.type}}-title') {{::list.header}}
-          h5.tasks-warning When you add a new task, it will be immediately saved to the challenge and given to all participants. Clicking the cancel button above will not revert your change.
+          h5.tasks-message=env.t('challengeTaskMessage')
 
           include ./task_view/help
 
           .todos-chart(ng-if='::list.type == "todo"', ng-show='charts.todos')
 
           include ./task_view/add_new
-          
+
           alert.alert-warning.dailiesRestingInInn(ng-if='list.type == "daily" && user.preferences.sleep && !$state.includes("options.social.challenges")')
             i.glyphicon.glyphicon-warning-sign &nbsp;
             =env.t('dailiesRestingInInn')
--- a/website/views/shared/tasks/index.jade
+++ b/website/views/shared/tasks/index.jade
@@ -7,22 +7,22 @@ include ./task-list
 include ./task
 
 script(id='templates/habitrpg-tasks.html', type="text/ng-template")
-  .tasks-lists.container-fluid
+  .tasks-lists.container-fluid    
     .row
       .col-sm-6.col-md-3(
         ng-repeat='list in lists',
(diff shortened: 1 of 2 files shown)


SabreCat:
I like it. That would certainly have stopped me from the goof that prompted my posting that issue ticket. XD @Alys @lemoness any thoughts on the wording?

Alys:
Suggested wording:
"When you add a new task, it will be immediately saved to the challenge and given to all participants. Clicking the cancel button above will not revert your change."

I don't think the word "Warning" is needed. It makes it sound a bit like an error condition, rather than just advising the user of the behaviour.

helenyau0:
@SabreCat @Alys thank you guys for responding back :) I will make those word changes now.

website/views/shared/tasks/index.jade
```diff
@@ -7,22 +7,22 @@ include ./task-list
...
+  .tasks-lists.container-fluid    
     .row
       .col-sm-6.col-md-3(
         ng-repeat='list in lists',
         ng-class='::{ "rewards-module": list.type==="reward", "new-row-sm": list.type==="todo" }')
         .task-column(class='{{::list.type}}s')
           include ./task_view/graph
-
           h2.task-column_title(class='{{::list.type}}-title') {{::list.header}}
+          h5.tasks-warning When you add a new task, it will be immediately saved to the challenge and given to all participants. Clicking the cancel button above will not revert your change.
```
librarianmage: Shouldn't this be localized instead of put directly into the file?

librarianmage (changes requested):

website/views/shared/tasks/index.jade
Alys: Oh yes, good point. @helenyau0 If you're not sure how to do localised text (i.e., text that can be translated into the user's own language), have a look at line 28 inn this file, where it says `=env.t('dailiesRestingInInn')`. That's an example of how to insert text in a way that can be translated. `dailiesRestingInInn` is a string in [website/common/locales/en/tasks.json](https://github.com/HabitRPG/habitica/blob/develop/website/common/locales/en/tasks.json). For this string that you'd need to add, the `challenges.json` file in the same directory would be appropriate.

Also, can you please remove the blank spaces or tabs at the end of lines 10 and 25? You'll see them highlighted in dark green in [GitHub's diff](https://github.com/HabitRPG/habitica/pull/8723/files) for your change. It's a minor point but it helps to keep the code clean.

TheHollidayInn:
Looks good and works locally for me.

SabreCat (changes requested):
Problem: under this PR, the warning appears on ALL task screens, not just Challenges. The warning should only appear in the Challenge create/edit context.

Suggestion: It'd look much better if the warning appeared once across the task columns, not repeated per every task type. Put a background behind it to draw attention.

TheHollidayInn:
Ah, I checked on challenges and group plans, my mistake. Should this appear on group plan tasks as well?

SabreCat:
You're only ever editing individual tasks in the group plan list, not a larger collection with top-level data, so I don't think it's applicable. (at least with current features!)

TheHollidayInn:
Makes sense!

SabreCat:
It's been a while since we've heard anything on this, and given this is a frontend change which will likely be invalidated by the Vue redesign in progress, I'm going to mark the source issue as On Hold and close this PR. @helenyau0, if you do get a chance to tidy things up, feel free to add commits and reopen the PR, and we can squeeze it in!

PR #8731, opened 2017-05-16
Discount Bundled Quests

SabreCat:

### Changes

Adds a new type of purchase: item bundles. The user pays a single Gem fee and receives several items of a normally _a la carte_ purchase category at a discounted rate. The first such bundle to be released packages three bird-themed pet quests for 7 Gems, a 5-Gem discount.

Also performs a minor refactoring of the Content folder, moving the `quests` object and related exports to its own JS file instead of living within /content/index.js.

### TO DO

* [x] Add tests


website/views/shared/modals/quests.jade
```diff
@@ -61,7 +61,8 @@ script(type='text/ng-template', id='modals/buyQuest.html')
     +questInfo
   .modal-footer
     button.btn.btn-default(ng-click='closeQuest(); $close()')=env.t('neverMind')
-    button.btn.btn-primary(ng-if='::selectedQuest.category !== "gold"', ng-click='purchase("quests", quest); closeQuest(); $close()')=env.t('buyQuest') + ': {{::selectedQuest.value}} ' + env.t('gems')
+    button.btn.btn-primary(ng-if='::selectedQuest.bundleKeys', ng-click='purchase("bundles", selectedQuest); closeQuest(); $close()')=env.t('buyQuestBundle') + ': {{::selectedQuest.value}} ' + env.t('gems')
+    button.btn.btn-primary(ng-if='::selectedQuest.category && selectedQuest.category !== "gold"', ng-click='purchase("quests", selectedQuest); closeQuest(); $close()')=env.t('buyQuest') + ': {{::selectedQuest.value}} ' + env.t('gems')
```
SabreCat, line 65: For some reason, this stopped working when passing `quest` instead of `selectedQuest`. It seems to work OK this way, but how did I break it...?

phillipthelen (approved):
Tested the PR with the mobile app. works great!
@TheHollidayInn will handle the code review.

website/common/script/ops/purchase.js
```diff
@@ -101,6 +102,14 @@ module.exports = function purchase (user, req = {}, analytics) {
 
   if (type === 'gear') {
     user.items.gear.owned[key] = true;
+  } else if (type === 'bundles') {
```
TheHollidayInn, line 105: Is the price set in the content file?

TheHollidayInn: This will need tests.

website/common/script/libs/shops.js
```diff
@@ -114,6 +114,27 @@ shops.getMarketCategories = function getMarket (user, language) {
 shops.getQuestShopCategories = function getQuestShopCategories (user, language) {
   let categories = [];
 
+  let bundleCategory = {
+    identifier: 'bundle',
+    text: i18n.t('questBundles', language),
+  };
+
+  bundleCategory.items = sortBy(values(content.bundles)
```
TheHollidayInn, line 174: I know we haven't started doing this, but I think it would be good to add a doc code with an example of this data structure. As in, what it looks like after fitlered and mapped.

website/client-old/js/services/questServices.js
```diff
@@ -82,33 +85,35 @@ angular.module('habitrpg')
...
-      if(quest.drop.gp)
-        text += quest.drop.gp + ' ' + window.env.t('gold') + '\n\n';
-
-      var ownerRewards = _.filter(quest.drop.items, 'onlyOwner');
-      if(ownerRewards.length > 0) {
-        text += '**' + window.env.t('rewardsQuestOwner') + ':**\n\n';
-        _.each(ownerRewards, function(item){
-          text += item.text() + '\n\n';
-        });
+      if (quest.drop) {
```
TheHollidayInn, line 88: Why is this now surrounded by quest.drop?

TheHollidayInn (commented):
Looks good overall!

website/common/script/ops/purchase.js
SabreCat: Added tests! See https://github.com/HabitRPG/habitica/pull/8731/commits/6feba59d4b995b0996e5ac204bcc0feef8c75e5b

SabreCat: Price is in Content, aye.

website/client-old/js/services/questServices.js
SabreCat: Previously, we assumed that if the quest data passed did not include `quest.boss`, then it must have `quest.drop`. This was not an extensible assumption--when I introduced bundles, which are neither bosses nor collection quests, this function would error out as soon as it tried to find `quest.drop.items`.

website/common/script/libs/shops.js
SabreCat: I might be able to add such a thing; do you have an example or recommended template I could work from?

test/common/ops/purchase.js
```diff
@@ -200,5 +202,24 @@ describe('shared.ops.purchase', () => {
...
+        'harpy',
+        'owl',
+      ];
+
+      purchase(user, {params: {type, key}});
+
+      forEach(questList, (bundledKey) => {
+        expect(user.items.quests[bundledKey]).to.equal(1);
+      });
+
```
TheHollidayInn, line 223: You should add a check for amount on the user as well.

test/common/ops/purchase.js
```diff
@@ -200,5 +202,24 @@ describe('shared.ops.purchase', () => {
 
       expect(user.items.gear.owned[key]).to.be.true;
     });
+
+    it.only('purchases quest bundles', () => {
+      let clock = sandbox.useFakeTimers(+moment('2017-05-20'));
```
TheHollidayInn: Why are you using a timer?

website/common/script/libs/shops.js
TheHollidayInn: I don't know what the content looks like :P I would add something like this:

```
The bundle model must follow the following structure:
{
  items: [Content Item],
  price: 50,
}
```

test/common/ops/purchase.js
SabreCat: Oh, checking that the Gems were deducted? Good point.

test/common/ops/purchase.js
SabreCat: This ensures that we won't need to update the test when the specific bundle stops being available May 31.

test/common/ops/purchase.js
SabreCat: Added in 5197375

website/common/script/libs/shops.js
SabreCat: Ah, OK, I slightly misunderstood what you were talking about. I'll see what I can come up with!

TheHollidayInn (approved):

PR #8733, opened 2017-05-17
First OAuth 2.0 implementation {continued in #8747}

sreckosmodis:

### Changes

#### OAuth 2.0 implementation
I have created an simple interface in user settings (API), where a user can create an OAuth client/application. There the user gets the client_id and client_secret for the OAuth flow.
Then a user can invoke the authorization for the application.
The implemented method is the grant_type "code".
App invokes the endpoint /api/v3/oauth/authorization with client_id, response_type and redirect_uri
User needs to login - local passport strategy
User grants access and app gets auth code
App exchanges auth code for access token


website/client-old/js/services/userServices.js
```diff
@@ -613,7 +633,7 @@ angular.module('habitrpg')
 
         authenticate: function (uuid, token, cb) {
           if (uuid && token) {
-            var offset = moment().zone(); // eg, 240 - this will be converted on server as -(offset/60)
+            var offset = moment().utcOffset(); // eg, 240 - this will be converted on server as -(offset/60)
```
paglias: can you revert the changes to `utcOffset` back to `zone`? we already tried to change them but caused issues with out code so now we've a more broad open PR that should be merged soon that deals with all the problems

website/common/locales/en/settings.json
```diff
@@ -184,5 +184,10 @@
     "timezone": "Time Zone",
     "timezoneUTC": "Habitica uses the time zone set on your PC, which is: <strong><%= utc %></strong>",
     "timezoneInfo": "If that time zone is wrong, first reload this page using your browser's reload or refresh button to ensure that Habitica has the most recent information. If it is still wrong, adjust the time zone on your PC and then reload this page again.<br><br> <strong>If you use Habitica on other PCs or mobile devices, the time zone must be the same on them all.</strong> If your Dailies have been resetting at the wrong time, repeat this check on all other PCs and on a browser on your mobile devices.",
-    "push": "Push"
+    "push": "Push",
+    "applications":"Applications",
```
paglias: Space between `:` and `"`. Also these strings are developers oriented so English only should be fine. But let's see later

website/server/controllers/api-v3/oauth.js
```diff
@@ -0,0 +1,151 @@
...
+
+api.createClient = {
+  method: 'POST',
+  url: '/oauth/client',
+  middlewares: [authWithHeaders()],
+  async handler (req, res) {
+    let user = res.locals.user;
+
+    req.checkBody({
+      redirectUri: {notEmpty: {errorMessage: res.t('missingRedirectUris')}},
```
paglias, line 35: these strings don't seem to exist in the locales files

website/server/controllers/api-v3/oauth.js
```diff
@@ -0,0 +1,151 @@
...
+  }
+};
+
+api.loginPost = {
+  method: 'POST',
+  url: '/oauth/login',
+  middlewares: [
+    passport.authenticate('local',{ successReturnToOrRedirect: '/', failureRedirect: '/api/v3/oauth/login' })
+  ],
+  async handler (req, res) {
```
paglias: this route doesn't do anything except for `passport.authenticate('local',{ successReturnToOrRedirect: '/', failureRedirect: '/api/v3/oauth/login' })`?

website/server/controllers/api-v3/oauth.js
```diff
@@ -0,0 +1,151 @@
...
+  }
+};
+
+api.authorization = {
+  method: 'GET',
+  url: '/oauth/authorization',
+  middlewares: [
+    ensureLoggedIn('/api/v3/oauth/login'),
+    server.authorization(async (clientId, redirectUri, done) => {
+      let user = await User.findOne({'oauth.clients':{$elemMatch: {clientId: clientId}}}).exec();
```
paglias: `{'oauth.clients.clientId': clientId}` should also work if I remember correctly

website/server/controllers/api-v3/oauth.js
```diff
@@ -0,0 +1,151 @@
...
+    
+  }
+};
+
+api.authorization = {
+  method: 'GET',
+  url: '/oauth/authorization',
+  middlewares: [
+    ensureLoggedIn('/api/v3/oauth/login'),
+    server.authorization(async (clientId, redirectUri, done) => {
```
paglias: this code should go in its own file under /middlewares? Otherwise if it's only used here I would just move it to the handler

website/server/libs/oauth.js
```diff
@@ -0,0 +1,103 @@
...
+// stored in the session.
+//
+// An application must supply serialization functions, which determine how the
+// client object is serialized into the session. Typically this will be a
+// simple matter of serializing the client's ID, and deserializing by finding
+// the client by ID from the database.
+
+server.serializeClient((client, done) => done(null, client.clientId));
+
+server.deserializeClient(async (clientId, done) => {
```
paglias: mixing async/await and callbacks usually causes problems, would you mind turning the `await User...` to a standard promise? `User.find(...).exec().then(...)`?

website/server/libs/routes.js
```diff
@@ -17,6 +17,7 @@ module.exports.readController = function readController (router, controller) {
     // If an authentication middleware is used run getUserLanguage after it, otherwise before
     // for cron instead use it only if an authentication middleware is present
     let authMiddlewareIndex = _.findIndex(middlewares, middleware => {
+      if(_.isUndefined(middleware.name)) return false;
```
paglias: let's just use `typeof middleware.name === 'undefined'`

website/server/libs/setupPassport.js
```diff
@@ -2,6 +2,70 @@ import passport from 'passport';
...
+} from './errors';
+
+/**
+ * LocalStrategy
+ *
+ * This strategy is used to authenticate users based on a username and password.
+ * Anytime a request is made to authorize an application, we must ensure that
+ * a user is logged in before asking them to approve the request.
+ */
+passport.use(new LocalStrategy(
```
paglias: Two things: would it be possible not to use LocalStrategy but doing like on the homepage where we make an AJAX call and then redirect based on the result of it? I know it's not optimal but I would prefer to avoid adding extra complexity to our code where possible. You should be able to reuse most of the code from the home page (it's a simple Angular app https://github.com/HabitRPG/habitica/blob/develop/website/client-old/js/static.js)

paglias: Otherwise the verification code for username and passwords should be abstracted somewhere so it's not duplicated

paglias (changes requested):

website/common/script/cron.js
```diff
@@ -1,124 +1,124 @@
-// TODO what can be moved to /website/server?
```
paglias: I think there's been some problem with git where it shows entire files modified even though there have been only small changes

website/server/libs/auth.js
```diff
@@ -0,0 +1,41 @@
...
+    let login;
+
+    if (validator.isEmail(username)) {
+      login = {'auth.local.email': username.toLowerCase()}; // Emails are stored lowercase
+    } else {
+      login = {'auth.local.username': username};
+    }
+
+    // load the entire user because we may have to save it to convert the password to bcrypt
+    User.findOne(login).exec().then((user)=>{
```
paglias: Two advices for Promises:

1. You can return the promise from this method so that we can choose to use either the promise or the callback to deal with results and errors

paglias: You can chain promises like this instead of nesting them, with a single `.catch` and easier code:

```
return User.findOne().....then(user => {
 if (!user) throw...
 return user;
}).then(user => {
    return passwordUtils.compare(...)
}).then(resultOfPasswordComparing => {

}).catch(err => //catches any error in previous promises)
```

paglias (changes requested):

paglias:
I'm going to move this code in a separate branch so that I can make the final changes before we can merge

paglias:
Closed in favor of https://github.com/HabitRPG/habitica/pull/8747. Thanks @sreckosmodis

PR #8737, opened 2017-05-17
[WIP] client/inventory/stable

negue:
- [x] Welcome to the Stable - Dialog
- [x] Hatch pet
   - [x] Hold to hatch (for 5 seconds with progress)
- [x] Sidebar Filter
   - [x] Pets
   - [x] Mounts
   - [x] Hide Missing
   - [x] Search
- [x] Sort
  - [x] Color
  - [x] A-Z
  - [x] Hatchable
- [x] Select a
  - [x] pet
  - [x] mount
- [x] feeding progress
- [x] food drawer
  - [x] drag food
  - [x] custom scrolling with two left/right buttons

Popovers:
- [x] Hatchable pet
- [x] normal pet 
- [x] food item

**mentionable changes:**
- ResizeDirective - get a callback if an element was resized
- DragDropDirective - easier implementation 
- ~~item component has now `popoverContent` and `itemContent` slots #8738~~

PR #8738, opened 2017-05-18
Client/item content

negue:
item component has now popoverContent and itemContent slots

extracted from #8737 

Ping @paglias

### DIFF
--- a/website/client/components/inventory/equipment.vue
+++ b/website/client/components/inventory/equipment.vue
@@ -30,7 +30,7 @@
           b-dropdown-item(@click="groupBy = 'class'", :class="{'dropdown-item-active': groupBy === 'class'}") {{ $t('class') }}
 
     drawer(
-      :title="$t('equipment')", 
+      :title="$t('equipment')",
       :errorMessage="(costume && !user.preferences.costume) ? $t('costumeDisabled') : null",
     )
       div(slot="drawer-header")
@@ -72,8 +72,11 @@
           :starVisible="!costume || user.preferences.costume",
           @click="equip",
         )
-          div(slot="itemContent", :class="'shop_' + flatGear[activeItems[group]].key")
-          equipmentAttributesPopover(:item="flatGear[activeItems[group]]", slot="popoverContent")
+          template(slot="itemContent", scope="ctx")
+            div(:class="'shop_' + ctx.item.key")
+          template(slot="popoverContent", scope="ctx")
+            equipmentAttributesPopover(:item="ctx.item")
+
     div(
       v-for="group in itemsGroups",
       v-if="viewOptions[group.key].selected",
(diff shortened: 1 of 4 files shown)


website/client/components/inventory/equipment.vue
```diff
@@ -65,12 +65,15 @@
           v-for="(label, group) in gearTypesToStrings",
           :key="group",
           :item="flatGear[activeItems[group]]",
+          :showPopover="flatGear[activeItems[group]] && flatGear[activeItems[group]].key.indexOf('_base_0') === -1",
```
paglias: Ah I get what you meant before, actually here `item` will already be avaible or at least it should be but I'm not super secure

paglias: If it doesn't work you might be able to use this feature to access item in the slot https://vuejs.org/v2/guide/components.html#Scoped-Slots

website/client/components/inventory/equipment.vue
```diff
@@ -65,12 +65,15 @@
...
           :label="$t(label)",
           :selected="true",
           :popoverPosition="'top'",
           :starVisible="!costume || user.preferences.costume",
           @click="equip",
         )
+          div(slot="itemContent", :class="'shop_' + flatGear[activeItems[group]].key")
```
paglias: Like here (referring to https://vuejs.org/v2/guide/components.html#Scoped-Slots)

website/client/components/inventory/equipment.vue
negue: so, item wasn't available, then after some issues with the scoped context it works, just the popover-issue now also happen in side the drawer (without the scope it didn't)

paglias: Sometimes for me it also happened in the drawer, depends on its position

website/client/components/inventory/equipment.vue
```diff
@@ -65,12 +65,18 @@
...
           :item="flatGear[activeItems[group]]",
+          :showPopover="flatGear[activeItems[group]] && flatGear[activeItems[group]].key.indexOf('_base_0') === -1",
           :label="$t(label)",
           :selected="true",
           :popoverPosition="'top'",
           :starVisible="!costume || user.preferences.costume",
           @click="equip",
         )
+          template(slot="itemContent", scope="ctx")
+            div(:class="'shop_' + ctx.item.key")
```
paglias: What if instead of a slot the item content was just a prop passed to the item component since the only thing that changes is the class? But maybe a full slot gives more flexibility

negue: as long every item content will be just a css class, then sure thats a bit easier. Should I use a property instead of a slot?

paglias: Yeah I think it'll always be just a class that changes

negue: done

PR #8744, opened 2017-05-20
adjust postinstall command so it works in Windows as well as *nix

Alys:
**Do not merge without considering the impact on our production server!**

I'm proposing this change to package.json to make Habitica easier to install on Windows machines. The current `postinstall` command fails because it contains `;` which Windows interprets differently than Unix/Linux.

This PR replaces the `;` with `&&` which works in *nix and in at least some versions of Windows.

This changes the meaning of the postinstall line slightly because now the later commands won't run if the earlier ones failed but I don't see that being a problem and in fact is probably desirable to give faster, clearer feedback about failures.

I've tested this locally on Ubuntu 16.04 and Guy Geva (add37003-c9da-40cd-9eb6-803579c06b40) has tested it on Windows (see today's posts in the Aspiring Blacksmiths guild). If any admins run Windows, them testing this would be advisable.

Before merging, we should ensure this will work on our test, staging, and production servers.

Another alternative would be to move the postinstall commands to a script (with line breaks instead of `;`) and call the script from package.json - tell me if you'd prefer that.


TheHollidayInn:
This works for me. And, I believe && is the preferred syntax. You can see the use of them here: https://docs.npmjs.com/getting-started/using-a-package.json.

As for the CI error, if you merge develop in here, it will be fixed.

Alys:
develop has been merged in and tests are passing

PR #8755, opened 2017-05-25
Disallow interactions by blocked users; new "get objections" Members API route -- chatRevoked users cannot send private messages

### LINKED ISSUE #7971: flags.chatRevoked does not prevent user sending private messages (PMs)

SabreCat:

Fixes #7971 

Continuation of #7997, rebased onto `develop` as of 2017-05-25. @CrystallineCat, if you'd like to resume work on this, you need only say so!

### Changes

>This PR changes the way objections to user-to-user interactions are handled (e.g. chat revocation or blocks). It includes some refactoring to the API to make it more flexible and allows the UI to query the estimated success of interactions. No UI changes are done at this time.
>
>API refactoring:
>- Factors out the checks in `api.sendPrivateMessage` and `api.transferGems` to see if the interaction is allowed.
>- The checks are now in `User.getObjectionsToInteractionIfAny`; the method is designed to be easily extensible to new checks or new interactions.
>- To allow the UI to check whether an interaction is allowed without actually preforming it, there is a new function `api.getObjectionsToInteractionIfAny`; it only applies interactions initiated by the logged in user and can not be used to query interactions by other users.
>- All objections to an interaction now use NotAuthorized instead of NotFound.
>- All objection strings are now translatable.
>
>Additional checks:
>- Both `api.sendPrivateMessage` and `api.transferGems` now are now blocked by `sender.flags.chatRevoked`.
>- `api.transferGems` is now blocked by either user blocking the other.
>
>Tests:
>- Tests for the additional checks and the new API have been added.

---

UUID: fea892e5-ac99-4b03-83ac-d01a2154e2af

### DIFF
--- a/test/api/v3/integration/members/POST-transfer_gems.test.js
+++ b/test/api/v3/integration/members/POST-transfer_gems.test.js
@@ -43,7 +43,7 @@ describe('POST /members/transfer-gems', () => {
     });
   });
 
-  it('returns error when to user is not found', async () => {
+  it('returns error when recipient is not found', async () => {
     await expect(userToSendMessage.post('/members/transfer-gems', {
       message,
       gemAmount,
@@ -55,7 +55,7 @@ describe('POST /members/transfer-gems', () => {
     });
   });
 
-  it('returns error when to user attempts to send gems to themselves', async () => {
+  it('returns error when user attempts to send gems to themselves', async () => {
     await expect(userToSendMessage.post('/members/transfer-gems', {
       message,
       gemAmount,
@@ -67,7 +67,7 @@ describe('POST /members/transfer-gems', () => {
     });
   });
 
-  it('returns error when to user has blocked the sender', async () => {
+  it('returns error when recipient has blocked the sender', async () => {
     let receiverWhoBlocksUser = await generateUser({'inbox.blocks': [userToSendMessage._id]});
 
     await expect(userToSendMessage.post('/members/transfer-gems', {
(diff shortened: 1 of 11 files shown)


SabreCat:
@khipkin @paglias @TheHollidayInn, does this cover your concerns from the prior PR?

paglias:
Looks good to me

test/api/v3/integration/members/POST-transfer_gems.test.js
```diff
@@ -67,6 +67,64 @@ describe('POST /members/transfer-gems', () => {
...
+      gemAmount,
+      toUserId: receiverWhoBlocksUser._id,
+    })).to.eventually.be.rejected.and.eql({
+      code: 401,
+      error: 'NotAuthorized',
+      message: t('notAuthorizedToSendMessageToThisUser'),
+    });
+  });
+
+  it('returns error when sender has blocked to user', async () => {
```
TheHollidayInn: should be blocked the receiver

TheHollidayInn (commented):
Works locally for me. What is the get objects route use case? I'm not seeing where it is used.

test/api/v3/integration/members/POST-transfer_gems.test.js
SabreCat: We use that "to user" phrasing (as distinct from the "from user", from/to...) throughout this test file. I agree it's more readable as "recipient"; should we then change it in the rest of the tests?

SabreCat:
I think the idea is that we can run this prior to firing off a chat message, and if it's going to be problematic, stop there. It's extensible so we could add in the swearword or slur blocking, and so on. It's not actually hooked in yet. But if we don't want it, we can close the PR.

TheHollidayInn:
Sounds good to me. I was just making sure we didn't need to be using it in this PR

PR #8749, opened 2017-05-23
Weekly/Monthly Habit reset counters resetting early - fixes #8570

joe-salomon2:
[//]: # (Note: See http://habitica.wikia.com/wiki/Using_Habitica_Git#Pull_Request for more info)

[//]: # (Put Issue # or URL here, if applicable. This will automatically close the issue if your PR is merged in)
Fixes https://github.com/HabitRPG/habitica/issues/8570

### Changes
[//]: # (Describe the changes that were made in detail here. Include pictures if necessary)

Previously, weekly/monthly habit reset counters were using server time to calculate "Monday" or "First of the month" to reset. This caused counters to reset too often if a user's timezone was already the next day, but the server was still on a Monday or "First day". Also, counters would reset early if the user was behind UTC.

For habit reset logic, changed day check calculation to use user’s timezone instead of server time.
Added unit tests to check following cases:
- Weekly habit reset: Server tz is Sunday, User tz is Monday
- Weekly habit reset: Server tz is Monday, User tz is Sunday
- Monthly habit reset: Server tz is 1st of month, User tz is 2nd of month
- Monthly habit reset: Server tz is end of prev month, User tz is 1st of month

[//]: # (Put User ID in here - found in Settings -> API)

----
UUID: fc28e9dc-9c84-4ee3-be6e-b4dffa54f558

PR #8767, opened 2017-05-30
Dailies performance fix - fixes #8756

### LINKED ISSUE #8756: editing Dailies is slow or freezes / crashes website (calculating next due dates)
Three reports below from the Report a Bug guild of problems when editing Dailies because the calculation for "Next Due Dates" is slowing things down. I've noticed it myself now when testing Yearlies with large frequencies - initial loading of the edit form is slow and when I made a change to the repeat settings, the website hung while recalculating the next due dates.

Some initial thoughts about how to handle this (other ideas most welcome!):
- only show the next due date and have a button for showing others after that (so only one calculation is done)
- when editing the repeat settings, don't automatically change the next due dates; perhaps have a button that appears as soon as a change is detected and you click the button if you want to see the updated dates (useful if you're changing two or more settings at once - no need to wait for the recalculation after each step)
- maybe do these fixes only when we can detect that the calculation is likely to be slow - e.g., for "daily" or "weekly" repeats with a small frequency number, just show all next due dates at once and refresh as soon as there's changes since there's unlikely to be any delay for them

yamo25000 (242241ad-cbf2-4f1d-8549-ed1f24a3476b) :
"Hello, I noticed a bug when making a yearly Daily. It seems that if one makes the value of X too high, the website will start to lag, and eventually crash. I think this is probably due to the algorithm that predicts the Daily's next active date, but that's just a guess."

Laura D (fbd1f053-a460-4b34-a893-2d5934b30e69) :  
"... dear goodness, don't switch "every 365 days" to "yearly" before changing the 365. It froze for so long calculating the next run dates. Should have set that one to 1, then yearly, instead of the other way around. ;) LOL"

thebat137 (dea8a9e8-c084-438a-9a14-7fe96a73fd0a) :
"I'm really thrilled to see the new task frequency settings (daily/weekly/monthly/yearly), but ever since they were added, editing dailies seems to have gotten incredibly slow for me (in the browser). Opening a daily for editing takes a while, and then the response to clicking on different settings changes is somewhere between slow and non-existent...  sometimes it slows things down badly enough that it locks up the whole browser."

joe-salomon2:

Fixes https://github.com/HabitRPG/habitica/issues/8756

### Changes

For recurring Daily tasks, the [moment-recur](https://github.com/c-trimm/moment-recur) plugin was slowing down calculations for the next due date. This is a known issue with the plugin:

> Important Note: These functions may be very inefficient/slow. They work by attempting to match every date from the start of a range until the desired number of dates have been generated. So if you attempt to get 10 dates for a rule that matches once a year, it will run the match function for ~3650 days.

The logic was rewritten to use [momentjs'](https://momentjs.com/docs/#/manipulating/add/) adding functions to calculate the upcoming due dates.

----
UUID: fc28e9dc-9c84-4ee3-be6e-b4dffa54f558


TheHollidayInn:
Thanks for the PR! There isn't an issue with matching and checking for recurrence - just for calculation the next due. So, we don't want to check the "matching" logic. Please only change the nextDue calculation.

Also, can you post your time measurements to show performance improvements?

Great work!

website/common/script/cron.js
```diff
@@ -110,81 +110,95 @@ export function shouldDo (day, dailyTask, options = {}) {
...
-    let schedule = moment(startDate).recur()
-      .every(dailyTask.everyX).days();
 
-    if (options.nextDue) return schedule.fromDate(startOfDayWithCDSTime).next(6);
-
-    return schedule.matches(startOfDayWithCDSTime);
-  } else if (dailyTask.frequency === 'weekly') {
-    let schedule = moment(startDate).recur();
+    let i = 1;
+    while (schedule.length < 6) {
```
TheHollidayInn: Can you change this while loops to for loops to be more concise.

joe-salomon2:
By "matches" logic, do you mean this previous code:
`    return schedule.matches(startOfDayWithCDSTime) && matchEveryX;
`

TheHollidayInn:
Exactly.

joe-salomon2:
So the matches() function is actually part of the moment-recur plugin. Since I am generating the "schedule" manually as an array, rather than a moment-recur object, I won't be able to use the previous matching check.

TheHollidayInn:
Alright, let me think it through a bit. We have forked the moment-reccurr library and I have plans on enhancing it to meeting more needs. I'd rather keep logic abstracted on that path.

Can you link your performance measurements/benchmarks to show this change is better?

joe-salomon2:
Oh wow, I was preparing the timing tests for you guys, and it actually wasn't the .recur().every() calculation function that was the culprit, it was the matches() function. I guess that makes sense, since it has to do a match check on all the generated dates.

Let me know how you want to proceed.

TheHollidayInn:
The issue should be the next, prev and all functions. How are you measuring? I'll be posting some local tests as well.

joe-salomon2:
I was grabbing timestamps at certain points in the code like below and monitoring the values in Chrome. Local unit tests didn't seem to produce any time issues though.

```
} else if (dailyTask.frequency === 'yearly') {
    let t0 = performance.now();
    let schedule = moment(startDate).recur();

    schedule = schedule.every(dailyTask.everyX).years();
    
    if (options.nextDue) {
      let dates = schedule.fromDate(startOfDayWithCDSTime).next(6);
      let filterDates = dates.filter((momentDate) => {
        let monthDiff = momentDate.years() - moment(startDate).years();
        let matchX = monthDiff % dailyTask.everyX === 0;
        return matchX;
      });
      let t1 = performance.now();
      console.log(t1-t0);
      return filterDates;
    }

    return schedule.matches(startOfDayWithCDSTime);
  }
```
and
```
function generateNextDue (task, user) {
      var options = angular.copy(user);
      options.nextDue = true;
      var t0 = performance.now();
      var nextDueDates = Shared.shouldDo(new Date, task, options);
      var t1 = performance.now();
      console.log(t1 - t0);
```

TheHollidayInn:
All the below are averages of 10 tests. 

Matching (doesn't seem to be the problem):
 - everyX 300, 3.0ms
 - everyX 30, 2.7ms
 - everyX 3, 2.6ms

Generating Dates:
 - everyX 300, 25375.0ms
 - everyX 30, 2325.7ms
 - everyX 3, 253.6ms

Local unit tests to show the issue. I am sure Chrome times the tab out after a delay and tries to close.
So, the only fix we needs is generating the next dues. It also seems this issue only happens on yearlies.

TheHollidayInn:
Using this code: 
```
if (options.nextDue) {
      let filterDates = [];
      for (let i = 1; i < 6; i += 1) {
        let calcDate = moment(startDate).add(dailyTask.everyX * i, 'years');
        if (calcDate >= startOfDayWithCDSTime) filterDates.push(calcDate);
      }
      return filterDates;
    }
```

Reduces the time to 1ms. So, I think that is all we need for this fix. We can take this idea and apply it to the moment recurr repo to improve performance as well.

joe-salomon2:
I get what you're saying now. Will fix

joe-salomon2:
Writing additional tests now to make sure the calculated values are valid.

test/api/v3/unit/libs/cron.test.js
```diff
@@ -392,6 +392,118 @@ describe('cron', () => {
       expect(tasksByType.dailys[0].nextDue.length).to.eql(6);
     });
 
+    it('should compute daily nextDue values', () => {
```
TheHollidayInn: These tests should go in the common tests directory. Thanks for writing them.

website/common/script/cron.js
```diff
@@ -207,6 +219,6 @@ export function shouldDo (day, dailyTask, options = {}) {
 
     return schedule.matches(startOfDayWithCDSTime);
   }
-
+  console.log(2);
```
paglias: this should be removed

joe-salomon2: The latest checkin has it removed. Thanks for reviewing

website/common/script/cron.js
```diff
@@ -110,11 +110,9 @@ export function shouldDo (day, dailyTask, options = {}) {
   }
 
   let daysOfTheWeek = [];
-
   if (dailyTask.repeat) {
     for (let [repeatDay, active] of Object.entries(dailyTask.repeat)) {
-      if (!isFinite(DAY_MAPPING_STRING_TO_NUMBER[repeatDay])) continue; // eslint-disable-line no-continue
-      if (active) daysOfTheWeek.push(parseInt(DAY_MAPPING_STRING_TO_NUMBER[repeatDay], 10));
+      if (active && !isNaN(DAY_MAPPING_STRING_TO_NUMBER[repeatDay])) daysOfTheWeek.push(parseInt(DAY_MAPPING_STRING_TO_NUMBER[repeatDay], 10));
```
TheHollidayInn: Why did you change this line?

TheHollidayInn:
Looking good. I just had one comment. I'll test locally later and check the performance as well.

website/common/script/cron.js
joe-salomon2: For some reason, the Object.entries() function was mapping additional functions/properties for dailyTask.repeat, which pushed unnecessary values to daysOfTheWeek. I was using the daysOfTheWeek.length property later on.

I don't have the output handy, but I can generate it again for you if you'd like in a few.

TheHollidayInn: Yea, the isFinite function should fix that though

joe-salomon2: you're right. I can quickly fix

TheHollidayInn:
The CI build broke. I am rerunning. Any ideas why the coverage went down?

joe-salomon2:
Weird, let me check on my end.

joe-salomon:
Here's the output just from a fresh develop branch. Do you notice anything odd?
https://gist.github.com/joe-salomon/3aa92a874eb8a3f8ae0f45fe6786edd4

joe-salomon:
When comparing my two coveralls below, the one for this PR doesn't have the job for 'TEST="test:api-v3" REQUIRES_SERVER=true COVERAGE=true'. Maybe that is why the coverage dropped?
This PR #8767 : https://coveralls.io/builds/11855993
PR #8749: https://coveralls.io/builds/11855636

joe-salomon2:
Fixed the issues :)

TheHollidayInn:
Great work. This is super good. Eventually, I'd like to move this too our fork of moment recur too. :D

TheHollidayInn:
Noted towards your Tier!

Alys:
@joe-salomon You now have tier 2 and are part way to tier 3.

PR #8769, opened 2017-05-30
Set isDue and NextDue during sleep

### LINKED ISSUE #8768: due / not due status and "Next Due Dates" for Dailies are not adjusted by cron when resting in the inn
When a player is resting in the inn (`preferences.sleep` set to true), cron will not run the code that recalculates the `isDue` and `nextDue` attributes for all Dailies.

This means that for the entire time a player is resting in the inn, each of their Dailies when viewed on the mobile apps will remain looking due {or not due} if the Daily was due {or not due} when they started resting, regardless of what the Daily's actual schedule should be. Also, the list of "Next Due Dates" in the Daily will not be updated.

If the player happens to mark a Daily as complete while resting, then it will be rescheduled properly, but since the point of resting is to avoid your Dailies, we can't rely on that happening often.

This is an important bug to fix quickly, because even while resting, you can still look at your Dailies to see which important ones have come due. For example, someone with a severe illness could be resting in the inn all the time and could be relying on their Dailies to take urgent medication on specific days.

TheHollidayInn:

Fixes #8768

### Changes

Sets isDue and nextDue when user is sleeping at the Inn.

----
UUID:

### DIFF
--- a/test/api/v3/unit/libs/cron.test.js
+++ b/test/api/v3/unit/libs/cron.test.js
@@ -313,6 +313,24 @@ describe('cron', () => {
       expect(tasksByType.dailys[0].completed).to.be.false;
       expect(user.stats.hp).to.equal(healthBefore);
     });
+
+    it('sets isDue for daily', () => {
+      let daily = {
+        text: 'test daily',
+        type: 'daily',
+        frequency: 'daily',
+        everyX: 5,
+        startDate: new Date(),
+      };
+
+      let task = new Tasks.daily(Tasks.Task.sanitize(daily)); // eslint-disable-line new-cap
+      tasksByType.dailys.push(task);
+      tasksByType.dailys[0].completed = true;
+
+      cron({user, tasksByType, daysMissed, analytics});
+
+      expect(tasksByType.dailys[0].isDue).to.be.exist;
+    });
   });
 
   describe('todos', () => {
--- a/website/server/libs/cron.js
+++ b/website/server/libs/cron.js
@@ -16,6 +16,16 @@ const i18n = common.i18n;
 const loginIncentives = common.content.loginIncentives;
 // const maxPMs = 200;
 
+function setIsDueNextDue (task, user, now) {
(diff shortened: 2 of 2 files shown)


SabreCat (approved):

test/api/v3/unit/libs/cron.test.js
```diff
@@ -313,6 +313,24 @@ describe('cron', () => {
...
+        text: 'test daily',
+        type: 'daily',
+        frequency: 'daily',
+        everyX: 5,
+        startDate: new Date(),
+      };
+
+      let task = new Tasks.daily(Tasks.Task.sanitize(daily)); // eslint-disable-line new-cap
+      tasksByType.dailys.push(task);
+      tasksByType.dailys[0].completed = true;
```
Alys, line 328: This isn't a thorough test for this bug because the Daily is being set to completed. The bug occurs with incomplete Dailies. Testing for both cases would be ideal.

TheHollidayInn: From looking at the performSleep function, setting the isDue is independent of the the completed status, so this line can be removed. The bug happened to both completed and incomplete.

Alys: In the manual tests I did, the bug did not apply to completed Dailies. Possibly the act of uncompleting it triggers the re-calculation of the due status? If you want me to, I'll confirm that with more manual tests locally using an old commit. If I'm not misremembering what I saw, then I think a test for each case (completed and not completed) might be safest.

TheHollidayInn: Yea, scoring the task computes the isDue status. So that is probably why it seemed to not take effect. I can add the test, but the outcomes won't affect the code because there is no branching statement.

Alys: Oh yes of course. Being completed doesn't matter when it's assigned like that because it's not actually being scored. Makes perfect sense now. Thanks! You're right that the test is fine as it is.

PR #8772, opened 2017-05-31
Mount Collector achievemt [WIP]- adds #8619

### LINKED ISSUE #8619: New Achievements and Badges: Grew X Mounts
These new achievements and badges are ready to be implemented! These should be received when a player grows the number of mounts indicated by each badge. The badge images and the names/mouseover text for the user achievement page can be found here: 

https://trello.com/c/ZP7TlViF/11-grew-x-mounts-badges

_edit by Alys:_  See the [comment below](https://github.com/HabitRPG/habitica/issues/8619#issuecomment-317213127) by @guyguy2001 which contains a lot of useful information for implementing this feature.

guyguy2001:
Fixes #8619 

I was told to send a pull request to show that I am working on it, so here it is. So far I have the achievement working, although I didn't put the achievement message, and it doesn't yet update automatically on mystic hourglass mounts (only risen ones). I finish those things and saving the achievement's data and update the PR.


Alys:
Just to clarify I suggested submitting a PR when you were asking for help because it's always easier for us to help when we can see the code. :) There's generally no need to submit one just to prove progress. 

Can you put "fixes (url of issue)" in the title and top post? That creates an automatic link to the issue.

SabreCat:
Cool, thanks for this! To make this easier for us to review, can you remove the compiled spritesheets (`website/assets/sprites/dist/spritesmith-*.*`) from the PR? Those changes will just get overwritten when the PR is incorporated, and it may save you having to deal with merge conflicts.

SabreCat:
Whoops--looks like you deleted the files and committed that, rather than reverting the changes to them. Sorry, I guess I should've been more clear! Let me see if I can come up with a set of steps you can use to correct the situation...

SabreCat:
This should do it:
```
git reset 9de118f0d
git add package.json website/common/locales/en/pets.json website/common/locales/en/tasks.json website/common/script/content/achievements.js website/common/script/libs/achievements.js website/common/script/ops/feed.js website/client-old/js/controllers/inventoryCtrl.js website/views/shared/modals/achievements.jade
git commit
git push --force
```

guyguy2001:
My problem is that I can't find the way to save the achievement's data. I thought that it is because I forgot to add it to schema.js, but even after I added it it starts as 0 whenever I reset the server; It only gets a value when I calculate it when recieveing a mount.

website/client-old/js/controllers/inventoryCtrl.js
```diff
@@ -253,20 +253,40 @@ habitrpg.controller("InventoryCtrl",
...
 
       // Selecting Pet
       } else {
         User.equip({params:{type: 'pet', key: pet.key}});
       }
     }
+    function handleMountAchievements() {
+        //Mount collector
+        var retrospectiveCount = (user.achievements.mountMaster || 0) * 90 + $rootScope.countExists(user.items.mounts);
+        user.achievements.mountCollector = Math.max(retrospectiveCount, user.achievements.mountCollector || 0);
```
SabreCat: It might be that you're only saving to the client-local `user` instead of the server `User`. Try changing out `user.achievements.mountCollector` for `User.user.achievements.mountCollector` when updating the value. (When reading the value, as in the latter part of this line, `user.` is fine.)

package.json
```diff
@@ -159,7 +159,7 @@
     "client:e2e": "node test/client/e2e/runner.js",
     "client:test": "npm run client:unit && npm run client:e2e",
     "start": "gulp run:dev",
-    "postinstall": "bower --config.interactive=false install -f; gulp build; npm run client:build"
+    "postinstall": "bower --config.interactive=false install -f && gulp build && npm run client:build"
```
SabreCat, line 162: What's this for?

website/common/locales/en/tasks.json
```diff
@@ -1,4 +1,4 @@
-{
+﻿{
```
SabreCat, line 1: Looks like an extraneous change. Not a big deal, but it would be a cleaner PR if this file were left out.

website/common/script/content/achievements.js
```diff
@@ -50,6 +50,18 @@ let seasonalSpellAchievs = {
 };
 Object.assign(achievementsData, seasonalSpellAchievs);
 
+let multiiconAchievs = {
```
SabreCat, line 53: This would be a little more readable as `multiIconAchievs`

website/common/script/ops/feed.js
```diff
@@ -1,4 +1,4 @@
-import content from '../content/index';
+﻿import content from '../content/index';
```
SabreCat, line 1: Another altered file with no relevance to the PR.

SabreCat (changes requested):

website/client-old/js/controllers/inventoryCtrl.js
guyguy2001: Oh, now I get it. I was sure that the user that I had there was the User.user collector used (Actually I think I was right there- but then I moved this piece of code to this file and forgot to change it. Not sure about it, but it doesn't matter).

package.json
guyguy2001: Forgot to leave it out of the commit. ';' doesn't work on windows so I changed it to '&&'

guyguy2001:
I changed it like you said, but User.user and user are apperently the same, and changing neither saves the achievement in the server. So, what could be the problem?

website/common/script/libs/achievements.js
```diff
@@ -147,6 +147,42 @@ function _addPlural (result, user, data) {
   });
 }
 
+function _addMultiIcon(result, user, data) {
```
TheHollidayInn, line 150: Please use curly braces here :D

website/common/locales/en/pets.json
```diff
@@ -1,101 +1,105 @@
...
+  "petKeyInfo4": "There are three Keys to the Kennels: Release Pets Only (4 Gems), Release Mounts Only (4 Gems), or Release Both Pets and Mounts. Using a Key lets you stack the Beast Master and Mount Master achievements. The Triad Bingo achievement will only stack if you use the \"Release Both Pets and Mounts\" key and have collected all 90 pets a second time. Show the world just how much of collection master you are! But choose wisely, because once you use a Key and open the kennel or stable doors, you won't be able to get them back without collecting them all again...",
+  "petKeyPets": "Release My Pets",
+  "petKeyMounts": "Release My Mounts",
+  "petKeyBoth": "Release Both",
+  "confirmPetKey": "Are you sure?",
+  "petKeyNeverMind": "Not Yet",
+  "petsReleased": "Pets released.",
```
TheHollidayInn, line 99: I think you have have change all the line endings here. PLease check that they are not CLFR

website/client-old/js/controllers/inventoryCtrl.js
```diff
@@ -253,20 +253,36 @@ habitrpg.controller("InventoryCtrl",
...
+            case 25:
+                break;
+            case 50:
+                break;
+        }
 
+        // Checks if mountmaster has been reached for the first time
+        if (!user.achievements.mountMaster
+            && $scope.mountCount >= 90) {
+            User.user.achievements.mountMaster = true;
```
TheHollidayInn: So, the issue here is that you are changing the users value on the client. You need to move this code to the server so the user gets saved to the DB. Probably here: https://github.com/HabitRPG/habitica/blob/develop/website/common/script/ops/feed.js#L77 or in the evolve function.

website/common/script/ops/feed.js
```diff
@@ -21,34 +19,25 @@ function evolve (user, pet, req) {
...
+        if (_user.items.mounts[i]) mountCount++;
     }
-module.exports = function feed (user, req = {}) {
+    //tracks mounts raised on mobile, or before the achievement was added
+    console.log("MC0 " + _user.achievements.mountCollector);
+    console.log(mountCount);
+    //var retrospectiveCount = (user.achievements.mountMasterCount || 0) * 90 + (mountsCount || 0);
+    //console.log(retrospectiveCount);
+    console.log("MC1 " + _user.achievements.mountCollector);
+    _user.achievements.mountCollector = Math.max(mountCount, (_user.achievements.mountCollector || 0) + 1);
```
guyguy2001, line 36: For some reason, user.achievements.mountCollector sometimes jumps  +91 either **between** calls to evolve, or after restarting the server. It seems to have something to do with line 36, as if it is Math.max(0, (...) + 1) it doesn't do it (at least not between function calls, I am not sure about the server reset yet). Any ideas as for why?

SabreCat: Is it always +91? Have you tried it with different accounts that have different numbers of mounts accumulated?

guyguy2001: Sorry, I didn't rxplain it well at all :P
When I get my first mount it's 1. Then it's 92 at the beginning of the next evolve, and then it's 93 as it is incrumented in the function

SabreCat: I wonder if there's something funky going on with the iteration in lines 27-29 vs. the way the Mongo schema for the mounts list works. What does it look like if you log out not just `mountCount`, but `_user.items.mounts`?

guyguy2001: _user.items.mounts is acting like it should. Also, when all the code I keep there is the one that inits it to zero, and then _user.items.mounts++ it works just fine. Do you know where is the code that reads the data from mongo? Or does it save and load schema as it is?

guyguy2001: Also, that 91 is added to the amount of mounts, not to achievements.mountCollector (so if I use debug tools to give myself mounts it will make it larger). I checked in choosePet (the UI function which is called when I click a pet), so it is +91 since before calling evolve (not only after the key for the mount was set true, and then it would have been +90 which makes more sense)

guyguy2001: First of all, I am really sorry for not being active on this at all in the last 2 weeks. I didn't both because life got in the way, speciffically a CS test, and terrible time managment.

I implemented the pet collector achievement (credit to @corinnaj for the UI, strings I think and making me understand what/where to put testing in / in testing). I did get the same bug, and it seems that it is always about 91 (will double check that tomorrow). I think that if I can't fix that in a few days I'll just finish the UI, share my code on the threads and let someone else continue, as I don't think I am getting anywhere with the server side code,

SabreCat:
@guyguy2001 Are you still trying to get this one sorted out? If not, let us know and we can open this up for another contributor to tackle.

guyguy2001:
Sorry. I meant to finish up some UI stuff, give an overview of what i know of the bug and of improvements that have to be made to the retrospective system, but I didn't really get into it. I will upload a comment on the issue and maybe commit some finish ups tomorrow (It is night here).

SabreCat:
Anything further on this, @guyguy2001?

guyguy2001:
No. I commented a week or so ago on the issue itself, in which I said that I quit it and someone else took it (it also contained all of the info that is needed to work on this)

PR #8774, opened 2017-05-31
Added repeat last functions

### LINKED ISSUE #8759: "5th Xday of the week" repeat confusion for Dailies
This is not exactly a bug and we already knew about it, but it's likely to keep being reported as a bug every so often, so perhaps adding some explanatory text to the Summary in this situation might help to clarify it.

As reported by Quilynn (f36d8801-177f-4a86-936e-35aead16c9ca) :
 
another minor issue with the "monthly" repeating dailies: if I choose "monthly" and "Day of the week" but the starting date is the 5th of the month, it repeats inconsistently. It only works when there are 5 of that day-of-the-week in the month.

So for instance, I have a daily with a start date of 1/31/2017 (The 5th sunday of the month.) If I choose "Monthly", "day of the week" and "repeat every 1 month" then the daily will only show up in May (5), Aug (8) and Oct (10) this year. If I select "repeat every 2 months" then the daily will activate May of this year and January of next year.

Maybe not a bad design. I used to go to a church that had a special event on the 5th Sunday of every month. So maybe it could be a very handy tool in edge cases when users only want to repeat on a daily on months with 5 Sundays (or any week day).

TheHollidayInn:

Fixes #8759

### Changes
This adds a repeat last functionality. The generating next due is being fixed in a separate PR, and I believe that function will allow for generating the last days easier, so this PR ignores that.


TheHollidayInn:
reapeat-last
As per requested, here is the select options for repeating last or 4th.

Alys:
Can this also give the option of last or 5th for the 5th week of the month? There's been one person saying they'd like to have literally the "5th {weekday}" so that the Daily is due only on months that have 5 of those days. It's a small use case but the code would be very similar to 4th, so this might be a good time to add it.

TheHollidayInn:
Actually, I just need to change the wording. The "4th week" option is the every 5 day of the week. 0-4 -> so in code 4 equals 5 :P

TheHollidayInn:
@TODO: if last week of month, show option

TheHollidayInn:
Closing this for now. I'll open a new PR after the redesign.

PR #8780, opened 2017-06-02
Fixes #8752

### LINKED ISSUE #8752: Wording on Weekly Dailies set to Repeat Every 0 weeks confusing on website.
(Though this is strictly speaking not a bug, it has caused a lot of confusion after the implementation of Weeklies, Monthlies and Yearlies, so Alys asked me to make an issue for it here. :) )

I have a.. wording suggestion regarding the weeklies and such. I realized after some questions in the Tavern and some testing myself.
When you create a Weekly task, that is active on e.g. Monday, you still have to set the Repeat Every to something else than 0 for it to ever be active. On the app (at least iOS) it says clearly that if the Repeat Every is set to 0, it will never repeat/be active. On the website, however, it says that it Repeats every 0 weeks on Mondays. Even though it does mean the same thing, I think it's a bit more clear wording in the apps - "Repeats never".

Sylvir9:

Issue #8752 

### Changes

1. If Repeat Every is 0 the summary is changed to "Repeats never."
2. Subheading added for "Repeat Type" 

issue8752

----
UUID:


website/client-old/js/services/taskServices.js
```diff
@@ -373,6 +373,8 @@ angular.module('habitrpg')
 
         summary += ' on the ' + (week + 1) + ' ' + longDay;
       }
+      
+      if (task._edit.everyX === 0) summary = 'Repeats never.';
```
TheHollidayInn, line 377: Can you change this to a translatable string. Also, you can but this condition at the top and simple return the repeats never string.

PR #8783, opened 2017-06-03
Never block admin messages fixes #8142

### LINKED ISSUE #8142: allow moderators to send a private message (PM) to any user - ignore opt-out / blocking
A moderator (anyone with `contributor.admin` set to true) should be able to send a PM to any player, even if the player has opted out of receiving PMs.

In the player's inbox, the on-hover text that explains the opt out feature must state that it does not prevent mods from sending them PMs.

The moderator should also be able to send a PM if the player has blocked the mod. Note that this should not be done by preventing the player from blocking the mod because the player might have blocked them before they became a mod.  Instead, this should be done by simply ignoring the block if the person sending the PM is a mod.

In the profile screen that you see when you click on a player's name, the on-hover text for the block button at the top right should state that it will have no effect if the player is a mod now or becomes a mod in future.

_FYI, this is because moderators sometimes need to contact players with important information, often for the player's own benefit (e.g., to explain that the player had to have their chat permissions temporarily removed and to describe what the player can do to have the permissions restored)._

mczuchnowski:
Fixes https://github.com/HabitRPG/habitica/issues/8142

### Changes
A very simple condition change that ignores blocking/opting out of private messages if the sender is a moderator (admin). In addition, the popover messages for `Block` button and `Opt Out` checkbox now inform the users about this special case. Tests for this have their own dedicated context.

----
UUID: a5966c55-f839-49c8-8462-5400d2ee4166


website/server/controllers/api-v3/members.js
```diff
@@ -417,8 +417,9 @@ api.sendPrivateMessage = {
     let userBlockedSender = receiver.inbox.blocks.indexOf(sender._id) !== -1;
     let userIsBlockBySender = sender.inbox.blocks.indexOf(receiver._id) !== -1;
     let userOptedOutOfMessaging = receiver.inbox.optOut;
+    let senderIsNotAdmin = !sender.contributor.admin;
```
TheHollidayInn: Can you change this to `Boolean(sender.contributor.admin)`. I think that will be safer and more readable.

TheHollidayInn (commented):
Looks good and works locally! Just requested one change.

website/server/controllers/api-v3/members.js
mczuchnowski: @TheHollidayInn you mean `Boolean(!sender.contributor.admin)` (with the bang)?

TheHollidayInn: Yea, that should work 👍

mczuchnowski:
@TheHollidayInn fixed, but now I see something big has changed while this PR waited for changes, so not sure if this PR is still relevant

mczuchnowski:
So I'm guessing this should now get re-implemented in the new interactions flow that SabreCat has introduced ;)

TheHollidayInn:
Ah, yes, we did just merge in some objections code. I think the code can just be abstracted to the objections, but I'm fine doing that separately. This still works correctly.

mczuchnowski:
The tests claim it doesn't work ;) I think the objections currently block this all the way, even for admins, I'll see what I can do with the objections flow.

TheHollidayInn:
Thank you! Sorry about that. We merged in quite a few yesterday.

mczuchnowski:
Not a problem, let's be agile ;)

SabreCat:
@mczuchnowski Are you still planning to work on this? Let us know, or we can open it up for another contributor to tackle.

mczuchnowski:
@SabreCat yeah, you can open it up again, I'm currently swamped with work and side projects, sorry.

PR #8795, opened 2017-06-07
New client members

TheHollidayInn:


website/client/components/guilds/membersModal.vue
```diff
@@ -17,13 +34,37 @@ div
     button(@click='confirmRemoveMember(member)', v-once) {{$t('remove')}}
 </template>
 
+<style lang='scss'>
+
+header {
+  background-color: #edecee;
```
paglias: since I'm cherry picking this, just a reminder to use variables for colors here and a few lines below

PR #8438, opened 2017-01-21
Ability to delete account as a social user { Continued in #8796 }

rae-ralston:
[//]: # (Note: See http://habitica.wikia.com/wiki/Using_Habitica_Git#Pull_Request for more info)

[//]: # (Put Issue # or URL here, if applicable. This will automatically close the issue if your PR is merged in)
Fixes #7417

### Changes
[//]: # (Describe the changes that were made in detail here. Include pictures if necessary)

Updated so that users without a local log in information can still delete their account by typing DELETE instead of a password. When the settings page is loaded if `user.auth.local.email` is present then the modal requiring the password pops up. If it's not present then it's assumed that the person is logged in via social and they're given the DELETE modal.

social delete modal

The user route handles verification, and checks if both `user.auth.local.email` & `user.auth.local.hashed_password` are true and if so, uses the previous written flow to delete a user. 

If `user.auth.google.id` or `user.auth.facebook.id` are true then the logic checks that the user wrote DELETE (case sensitive).

#### File Changes
- **front.json & settings.json**: added phrases that match the updated UI elements.
- **website/server/controllers/api-v3/user.js**: added an `else/if` statement that checks if the user has a local or social log in and check the respective expected input entry.
- **website/views/options/settings.jade**: added another button and if logic to both so the correct button will display depending on whether or not the user has local auth.
- **website/views/shared/modals/settings.jade**: added the social modal content.

I'd love it if someone could give me some reflections on the best things to test the validity of a local user account. Can someone get an email into local if they don't have a local registration? I couldn't get a real idea of what social user data is stored in `user.auth.google/facebook`, is looking for `.id` specific enough?

Looking forward to reflections 💃 

big ups to @pllearns for writing the testing and @deonna for help and support <3 

[phil's testing PR](https://github.com/HabitRPG/habitica/pull/8429)

----
rachel-ftw's UUID: 97f4775f-0e99-47cd-8002-63a062972bf6

PR #8796, opened 2017-06-07
Delete Account with Social Auth

### LINKED ISSUE #7417: can't delete account if you use Facebook or Google authentication

SabreCat:
Continuation of #8429 and #8438. @rachel-ftw, @pllearns, @deonna, if you'd like to resume work on this, just speak up!

Fixes #7417

### Changes

> Updated so that users without a local log in information can still delete their account by typing DELETE instead of a password. When the settings page is loaded if `user.auth.local.email` is present then the modal requiring the password pops up. If it's not present then it's assumed that the person is logged in via social and they're given the DELETE modal.
>
>social delete modal
>
>
>The user route handles verification, and checks if both `user.auth.local.email` & `user.auth.local.hashed_password` are true and if so, uses the previous written flow to delete a user. 
>
>If `user.auth.google.id` or `user.auth.facebook.id` are true then the logic checks that the user wrote DELETE (case sensitive).

#### Misc

* Fixed line endings in website/views/options/settings.jade
* Removed validation errors check without any preceding `req.check` calls

TO DO:

- [x] Add tests

----
rachel-ftw's UUID: 97f4775f-0e99-47cd-8002-63a062972bf6


TheHollidayInn (approved):

SabreCat:
@rachel-ftw and @pllearns, I've awarded you Blacksmith tier 1 for your help here! Thank you!

PR #8799, opened 2017-06-08
Dailies new subheading and summary change

Sylvir9:

Issue #8752 
Fixes put_issue_url_here

### Changes
This is the same as #8780 and is only here due to my git-related ineptitude.  
Also, moved "Repeats never." to tasks.json which I'm assuming makes it translatable probably.
1. Summary for zero repeats now says "Repeats never."
2. New subheading for "Repeat Type"
untitled

----
UUID:


SabreCat:
This looks pretty good! I'm not sure I agree with the conversation on the original issue ticket about the suggested phrasing, though. "Repeats never" is rather awkward English for this scenario. @Alys @Lalaitha, what would you think about one of the following instead?

"Does not repeat."
"Daily is inactive."
"Daily is disabled."

Alys:
"Daily is never due"

Maybe even something like "Daily is never due (it's a [Grey Daily](http://habitica.wikia.com/wiki/Dailies#Grey_Dailies))" since that's a specific term for it and it has uses as listed on that wiki page.

Disabled / inactive imply that it can't be used.

SabreCat:
Let's go with "Daily is never due." @LlamaLad, care to make that quick change?

Lalaitha:
Sounds good to me. Makes it very clear. :)
Maybe the wording should be adjusted in the apps as well, then? It says "repeats never" there.

Sylvir9:
Yup it's been changed to "Daily is never due"

SabreCat (approved):

SabreCat:
Thanks, @LlamaLad! What's your User ID from https://habitica.com/#/options/settings/api ? Let me know and I can give you contributor credit on the site!

PR #8803, opened 2017-06-08
Client: Header Party Members

paglias:
- [ ] Expand
- [ ] Smaller stats
- [ ] Exclude yourself
- [ ] View party if on your own or invite friends?


paglias:
@TheHollidayInn pull develop and we should be good to go

PR #8805, opened 2017-06-09
Yesterdailies 2

TheHollidayInn:

### Changes
Replaces: #8717

Documentation of a discussion:

The way yesterdailies and cron will be changed. Essentially the following changes will be made:

- Cron is not checked/run on API calls automatically anymore. 
- Instead there will be a special API call /cron that runs it.
- all dailies are handled the same on the server regardless if they are yesterdailies or not.
- On the client side, 
     - the client checks if there are any uncompleted yesterdailies, 
     - displays those yesterdailies (if available), 
     - lets the user score the yesterdailies 
     - then runs the cron API call.

This change is made to later enable an offline mode for the mobile clients, so that they can score all tasks that were checked off when they were offline before cron is run.


website/server/models/task.js
```diff
@@ -244,6 +244,7 @@ export let DailySchema = new Schema(_.defaults({
   weeksOfMonth: {type: [Number], default: []}, // Weeks of the month that the daily should repeat on
   isDue: {type: Boolean},
   nextDue: [{type: String}],
+  yesterDaily: {type: Boolean, default: true},
```
paglias: let's add a required flag here

paglias (changes requested):

website/common/locales/en/tasks.json
```diff
@@ -169,5 +169,10 @@
     "yearlyRepeatHelpContent": "This task will be due every X years",
     "resets": "Resets",
     "summaryStart": "Repeats <%= frequency %> every <%= everyX %> <%= frequencyPlural %> ",
-    "nextDue": "Next Due Dates"
+    "nextDue": "Next Due Dates",
+    "yesterDailiesTitle": "You left these Dailies unchecked yesterday! Do you want to check off any of them now?",
+    "yesterDailiesSubTitle": "",
+    "yesterDailiesCallToAction": "Start My New Day!",
+    "yesterDailiesOptionTitle": "Confirm that this Daily wasn't done before applying damage",
+    "yesterDailiesDescription": "Hover: If this setting is applied, Habitica will ask you if you meant to leave the Daily undone before calculating and applying damage to your avatar. This can protect you against unintentional damage."
```
Alys: I think the word "Hover:" isn't meant to be there. :)

TheHollidayInn: Lol copy and paste fail

Alys:
I've added some notes to the wiki about this to help third-party developers understand the changes and use them appropriately. It's in the "Information for Developers" section at the bottom of the [Dailies](http://habitica.wikia.com/wiki/Dailies#Information_for_Developers) page. By default that section will be hidden - there's a button to unhide it. @TheHollidayInn if you have time, would you like to review it in case I have some of the details wrong?

PR #8703, opened 2017-04-26
Allow leaving a challenge without having access to the challenge - fixes #8693

savoie:
(e.g. after leaving a party or guild)

[//]: # (Note: See http://habitica.wikia.com/wiki/Using_Habitica_Git#Pull_Request for more info)

[//]: # (Put Issue # or URL here, if applicable. This will automatically close the issue if your PR is merged in)
Fixes https://github.com/HabitRPG/habitica/issues/8693

### Changes
[//]: # (Describe the changes that were made in detail here. Include pictures if necessary)
* Exactly as @Alys described in the issue description; removed getting the challenge group and checking that the group exists, since the user might not have access.
* Note that the `challenge.canView(user, group)` check that is also removed is redundant, since that will return true if `challenge.isMember(user)` is true, and that's the next check anyways.

[//]: # (Put User ID in here - found in Settings -> API)

----
UUID: b58c50cd-d497-4f53-88a6-69f076ea1114

(Seemed fitting that since I reported the bug in the first place in the Report a Bug guild, I should finally get around to setting up my dev env to fix it.)

PR #8811, opened 2017-06-11
Fix allowing users to quit challenges after leaving party: fixes https://github.com/HabitRPG/habitica/issues/8693

### LINKED ISSUE #8693: can't leave challenges from a private guild or party
If you join a private guild or party, then join a challenge in that group, then leave the group, you are then not able to leave the challenge.

It's because of [these two lines](https://github.com/HabitRPG/habitica/blob/91438aff904ed127bd8f60d349625da07ed2d16f/website/server/controllers/api-v3/challenges.js#L193-L194) in the API's `leaveChallenge` function, which prevent the challenge being left when you don't have permission to view the group. That's an important check to perform when you are joining a challenge but not when leaving a challenge you've been permitted to see in the past.

I suggest this bug be fixed by removing those two lines. I'll leave this as suggestion-discussion for a couple of days and then if there's no objections, mark it as help wanted.

JoshSears:

Fixes https://github.com/HabitRPG/habitica/issues/8693

### Changes
Two lines commented out in challenges.js (347 and 348) which prevented the user from leaving a challenge after leaving the party.  Due to the discussed potential security risk found on the URL, I left the code commented out in case there was a need to alter to deal with the security issue at a later time.
If user chooses to keep tasks when leaving challenge, the challenge will automatically disappear from page.  If they remove the tasks, the challenge will stay - with 'leave' turned to 'join' (which doesn't work), and will be gone upon refresh.

----
UUID: 348b9f3a-83ef-4865-b4d1-8f64870c9734

### DIFF
--- a/website/server/controllers/api-v3/challenges.js
+++ b/website/server/controllers/api-v3/challenges.js
@@ -343,8 +343,8 @@ api.leaveChallenge = {
     let challenge = await Challenge.findOne({ _id: req.params.challengeId }).exec();
     if (!challenge) throw new NotFound(res.t('challengeNotFound'));
 
-    //Prevented users from leaving challenge after quitting party.
-    //let group = await Group.getGroup({user, groupId: challenge.group, fields: '_id type privacy'});
+   // Prevented users from leaving challenge after quitting party.
+   // let group = await Group.getGroup({user, groupId: challenge.group, fields: '_id type privacy'});
    // if (!group || !challenge.canView(user, group)) throw new NotFound(res.t('challengeNotFound'));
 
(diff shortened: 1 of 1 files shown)


JoshSears:
@Alys Error here in travis was that message changed since we removed the two lines in challenges.js that checked to see if user was still in group before leaving challenge.  Therefore, the error would change.

TheHollidayInn:
@Dagrik If the intention is for the error to change, you will need to change the test as well.

JoshSears:
Will complete!

JoshSears:
Still working on this.  Life had to take precedence first.

SabreCat:
This was fixed in #8703.

PR #8818, opened 2017-06-14
WIP - Subscription renew perks - fixes #4819

### LINKED ISSUE #4819: Subscription renewal hourglasses

joe-salomon2:

Fixes https://github.com/HabitRPG/habitica/issues/4819

### Changes

Previously, when a 6 / 12-month subscription was renewed, the user did not get their perks right away (hourglass, gem cap).

There is additional logic added to cron to check at the beginning of a new month, if a 6 / 12-month subscription was renewed, to award the user with the subscription perks right away.

----
UUID: fc28e9dc-9c84-4ee3-be6e-b4dffa54f558


website/server/libs/cron.js
```diff
@@ -66,20 +66,37 @@ function grantEndOfTheMonthPerks (user, now) {
...
-
-      if (plan.consecutive.offset > 1) {
-        plan.consecutive.offset--;
-      } else if (plan.consecutive.count % 3 === 0) { // every 3 months
-        if (plan.consecutive.offset === 1) plan.consecutive.offset--;
-        plan.consecutive.trinkets++;
-        plan.consecutive.gemCapExtra += 5;
-        if (plan.consecutive.gemCapExtra > 25) plan.consecutive.gemCapExtra = 25; // cap it at 50 (hard 25 limit + extra 25)
+
+    let dateCreatedMoment = moment(plan.dateCreated).startOf('month');
```
Alys, line 70: I'm afraid you can't rely on plan.dateCreated to be the start date of the subscription that the user currently has. If they had a previous subscription (e.g., a one-month recurring subscription) and they changed it to a different kind of subscription (e.g., six-month recurring), then dateCreated will be the date their first subscription was created. It can't be used to get an accurate calculation for whether the six-month period is finished.

joe-salomon2: Thanks for the insight. Is there a reason as to why we can't update dateCreated on a new subscription?

So based on what it looks like, createSubscription isn't getting called on a renewed subscription payment from the source (paypal, stripe, etc). Most likely because the webhook isn't being processed, or isn't even configured. Had createSubscription been called on a renew, the proper amount of perks would have been awarded, as can be proved by a unit test. Thus, I was hoping to use something like dateCreated (and other condition checks) to make sure an active 6 or 12-mo subscription existed.

SabreCat: Can you elaborate on what you mean by "why we can't update dateCreated on a new subscription"?

joe-salomon2: Based on @Alys' comment, I thought we couldn't update plan.dateCreated whenever a new subscription type was chosen. I.e, if you guys used that data for some type of calculation or info elsewhere. If there's no issue with me updating it whenever a new subscription is made, then I would propose it as part of this PR. 
E.g: 
User initially has a 3mo subscription, dateCreated is set for start of the 3mo sub. 
User then changes to a 6mo subscription, dateCreated gets updated for the start of the 6mo sub.

Alys: As far as I'm aware, there's no issue with it being updated when a new subscription is started, but that hasn't been done for all the existing subscriptions, so any code that uses dateCreated to calculate consecutive-month benefits will be inaccurate for many existing subscriptions and in a way that's less predictable than the current bug and therefore harder for us to do individual fixes for. I'm not saying we can't use this method - if it works for new subscriptions that will be great - but we might need to have some way of marking old subscriptions where we know the dateCreated is inaccurate to help us do manual fixes. E.g., we could do a binary flag that's set to true on all relevant subscriptions before this PR goes live and then is never set to true on any new subscriptions.

SabreCat:
@joe-salomon Are you still planning to work on this? If not, let us know and we can open this back up for another contributor to tackle.

Alys:
@joe-salomon2 How about instead of using the value of `dateCreated`, you base it on only `purchased.plan.planId`, which can have the values `basic_3mo`, `basic_6mo`, `google_6mo`, `basic_12mo`, and others. If `plan.consecutive.offset` has reached zero and the planId has a value containing `Nmo` where `N` is any integer, then you apply the consecutive month benefits for `N` months, including increasing the offset.

Let us know if you'd still like to work on this, or if we should open it up to other contributors.

SabreCat:
Given no response to queries for almost two weeks, I'm going to close this PR and re-flag the source issue as Help Welcome. @joe-salomon2, if you do come back to this, it's no problem to reopen the PR!

Issue #8598, opened 2017-03-25
the "Delete Completed" on-hover message needs to be changed


Alys:
_<strike>I'm logging this initially for suggestion/discussion, but once a decision is made, a PR will be needed to change some text on the website (regardless of the exact decision).</strike>_

<strike>When you click the "Delete Completed" button in the To-Dos "Done" tab, completed To-Dos from Group Plans are **not** deleted. Is this what we want? It's consistent with To-Dos from active challenges (we definitely don't want them to be deleted because that would make it harder for the challenge owner to judge the challenge), but I'm not sure there's any need to keep completed Group Plans To-Dos and they'll build up over time. To-Dos that required approvals would already have been approved before they become marked as completed. I don't think that Group Plans leaders can see the members' completed To-Dos, so there's probably no point keeping them.</strike>

<strike>Are there any opposing viewpoints? In about a week, if no one has expressed a desire for completed Group Plans To-Dos to be kept, I'll mark this as help welcome for allowing them to be deleted.</strike>

At the bottom of the To-Dos "Done" tab, there's a "Delete Completed" button. The text that appears when you hover over that button needs to be changed. Currently it says "If you click the button below, all of your completed To-Dos and archived To-Dos will be permanently deleted. Export them first if you want to keep a record of them." That should be changed to something like:

"If you click the button below, all of your completed To-Dos and archived To-Dos will be permanently deleted, except for To-Dos from active challenges and Group Plans. Export them first if you want to keep a record of them."

TheHollidayInn:
Completed todos get archived, correct? Why would we want to delete information? I think we should not delete todos.

Alys:
There isn't exactly an archiving process; it's just that the completed To-Dos on the website don't show more than 30 (the most recently completed ones). All completed To-Dos are stored in the database in the same way as other tasks until deletion. We delete old completed To-Dos for performance and database space ([Task History Compression](http://habitica.wikia.com/wiki/Task_History_Compression) for details).

Regardless of that though, even if we never deleted completed To-Dos ourselves, we must still allow the users to delete them when they want to. For example, they might realise there were privacy or security concerns with some of their completed To-Dos.

TheHollidayInn:
If we don't allow for deleting in challenge, we should not here either. Currently, the user's task is the only indication of approval for group leaders to verify that a member completed it. The task can be deleted if the group leader unassigns them.

CutTheCode:
[...]

PR #8826, opened 2017-06-17
minor text fixes: accurate flavour text for Golden Knight testimonies collection quest, etc

Alys:
This PR does three small, unrelated things:

- Adjusts the flavour text for the Golden Knight collection quest so that it mentions dozens of Testimonies not hundreds (thanks to Atanasi for reporting that in the Report a Bug guild).
- Improves accuracy of a couple of comments in the code for removing group members (members are removed, not banned, because they can always rejoin with an invitation in a private group or at any time in a public group).
- Standardises line breaks in a few files (view with ?w=1 in the URL to remove those whitespace fixes)

EDIT: See my comments below for a couple more things.


mandyzhou:
Should it also be "Habiticans'", not "Habitican's"?

Alys:
@mandyzhou Great point, well spotted, thank you! I've added a fix for that.

I've also made another unrelated change and added it to this PR. When you try to send a PM to a user who has either opted out of PMs or blocked you specifically, you used to see the message:
"Can't send message to this user."
We fairly often saw people report this as a bug. I've changed it to:
"You can't send a message to this player because they have chosen to block messages."
I kept it vague about whether they'd blocked all messages or just you for politeness.

Alys:
I've added one more change to update the instructions for unsubscription from Google subscriptions.

Alys:
I'll merge this in a day if there's no complaints.

Alys:
I've added a fix for #8598 (improves the Delete Completed on-hover message)

Alys:
Also added an improvement to the Orb of Rebirth text.

PR #8832, opened 2017-06-21
fix bug that prevented sending of emails to admin addresses

Alys:
Emails from the new feature to send feedback on account deletion haven't been getting through because the `sendTxn` / `txnEmail` call had a wrong parameter. 

This PR fixes that, and also fixes the same bug in a feature that sends emails to admins about certain subscription problems.

Admin slack discussion: https://habitica.slack.com/archives/C02V10HPE/p1498049915383928


paglias:
Can be merged for me!

paglias (approved):

PR #8840, opened 2017-06-26
Client/stable fixes - part 1

negue:
Fixes:
- disable feeding on pets which already exist as mount
- disabled filter instead of hiding them
- don't hide special pets
- avatar refreshes on selecting a pet / mount
- unfocus the first dropdown-item

### DIFF
--- a/website/client/components/inventory/stable/index.vue
+++ b/website/client/components/inventory/stable/index.vue
@@ -64,7 +64,7 @@
             b-dropdown-item(
               v-for="sort in sortByItems",
               @click="selectedSortBy = sort",
-              :active="selectedSortBy === sort",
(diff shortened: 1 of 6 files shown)


website/client/components/inventory/stable/index.vue
```diff
@@ -630,7 +678,7 @@
 
         let animalRows = [];
 
-        let itemsPerRow = Math.floor(availableSpace / (94 + 24));
+        let itemsPerRow = Math.floor(availableSpace / (94 + 20));
```
paglias, line 681: this can be used in the equipment too?

paglias (approved):

website/client/components/inventory/stable/index.vue
negue: sorry for this late response, yes!

PR #8843, opened 2017-06-29
add missing string for mountNowOwned error message

Alys:
When you try to equip a mount you don't own, you're supposed to see an error message saying that you don't own it, but that string was never created. This PR creates it.


SabreCat (approved):

SabreCat:
Failing test is unrelated and was recently fixed on `develop`. Merging!

PR #8864, opened 2017-07-12
Added catch for cron error

TheHollidayInn:


paglias (commented):
I'm not entirely sure if this is needed. All the code is wrapped in a try/catch so any await that throws is already logged

paglias:
this is the global error handler https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/errorHandler.js

TheHollidayInn:
This will prevent the 500 error if there is a version error. And will log the previous error better.

TheHollidayInn:
The issue is we never see the error on line 117 because the update is throwing the error.

SabreCat (approved):

PR #8865, opened 2017-07-12
Make responce codes uniform

asmfreak:
This PR and issue is about #8087 on improving API docs.
Post request return `200 Success` in the documentation, whereas in code and on actual site they return `201 Created`. I've fixed this issue in `tasks.js`, but it should be checked through all other api endpoints.

### DIFF
--- a/website/server/controllers/api-v3/challenges.js
+++ b/website/server/controllers/api-v3/challenges.js
@@ -158,7 +158,7 @@ let api = {};
  * @apiParam (Body) {Boolean} [official=false] Whether or not a challenge is an official Habitica challenge (requires admin)
  * @apiParam (Body) {Number} [challenge.prize=0] Number of gems offered as a prize to challenge winner
  *
- * @apiSuccess {Object} challenge The newly created challenge.
+ * @apiSuccess (201) {Object} challenge The newly created challenge.
  * @apiUse SuccessfulChallengeRequest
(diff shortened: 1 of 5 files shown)


SabreCat:
Thanks for this, good catch! What's your UUID in Habitica (from https://habitica.com/#/options/settings/api)? We can give you Blacksmith contributor credit for this!

asmfreak:
My Habitica user uuid is `1f740057-8e11-4ca3-a9b2-54886615dca6`.

asmfreak:
Offtopic question.
Is there a habitica testing server, on which one can test API? Or maybe API should be mocked?
I could have registered a testing user on main habitica server, but I just want a clean environment for each test run. I could also run habitica locally, but my niether my personal server nor my pc can run habitica.

SabreCat:
Great, thanks and welcome to Blacksmith tier 1! Note that further tiers require increasing amounts of effort from one to the next, so tier 2 may require multiple PRs or a larger PR to achieve. But keep helping out and we'll express our gratitude accordingly!

There aren't any public-facing test servers, no. You might check in with the Aspiring Blacksmiths guild--there may be some good free or minimal-cost server solutions out there you could avail yourself of! I know folks have run Habitica servers on Heroku, AWS EC2, and DigitalOcean, probably others.

PR #8870, opened 2017-07-13
Groups can prevent members from getting gems

paglias:
This PR adds the ability for groups with group plans to prevent their members from obtaining gems in Habitica.

The UI is missing for the setting, groups that need this feature will contact the admins who will then add a field in the database to enabled it.

In particular this becomes blocked for members of such groups:
- Buying gems
- Being gifted gems
- Converting gold to gems: due to the way this is implemented users will briefly see the conversion as successful but it'll be reverted as soon as the request hit the server, because the verification process is made on the server so the client doesn't know if the user can buy gems.
- Winning a challenge prize: in this case the winner creator is refunded

Leaders are excluded.

Tests are coming.


SabreCat:
Looks like this is causing some tests to fail? (Not sure if that's still WIP / you're already aware)

TheHollidayInn (approved):

PR #8884, opened 2017-07-17
[WIP] New Client - Shops/Market

negue:


website/client/assets/scss/icon.scss
```diff
@@ -1,8 +1,6 @@
 .svg-icon {
   display: block;
   stroke-width: 0;
-  stroke: currentColor;
```
paglias: can we keep these two lines? it's needed for many icons

website/client/assets/scss/modal.scss
```diff
@@ -0,0 +1,33 @@
+@mixin habitModal() {
```
paglias: why this file?

paglias: i mean, why it's called habitModal?

website/client/assets/scss/modal.scss
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.modal-dialog {
+  .title {
+    height: 24px;
+    margin-top: 24px;
+    font-family: Roboto;
+    font-size: 20px;
+    font-weight: bold;
+    font-stretch: condensed;
```
paglias: this doesn't work, you have to use `font-family: 'Roboto Condensed'`

website/client/components/shops/market/balanceInfo.vue
```diff
@@ -0,0 +1,55 @@
...
+    .svg-icon(v-html="icons.gem")
+    span {{userGems | roundBigNumber}}
+    .svg-icon(v-html="icons.gold")
+    span {{userGold | roundBigNumber}}
+</template>
+
+<style lang="scss" scoped>
+.span {
+  font-weight: normal;
+  font-family: Roboto;
```
paglias: the text is by default roboto

website/server/controllers/api-v3/user.js
```diff
@@ -1495,7 +1496,7 @@ api.userReleaseMounts = {
 api.userSell = {
   method: 'POST',
   middlewares: [authWithHeaders()],
-  url: '/user/sell/:type/:key',
+  url: '/user/sell/:type/:key/:amount',
```
paglias: will this break old calls to /user/sell/:type/:key since it now expect a longer path?

website/client/assets/scss/icon.scss
negue: those didn't do anything, thats why I moved them to 

```scss
&.color {
   svg path {
       stroke: currentColor;
       fill: currentColor;
     }
 }
```

but I can add them also back, if you want

website/client/assets/scss/modal.scss
negue: good question ... , I'll rename to `centeredModal`

website/server/controllers/api-v3/user.js
negue: I didn't thought about that, so what would be the best practice for this? since the `common.ops.sell` can work without the amount parameter.

- copy the api.userSell with a different url
- or use amount as query-param?

paglias: I'd say query param

website/client/assets/scss/icon.scss
paglias: I'm sure this is working but you first have to remove any fill="xxx" from the svg files before you can use it, otherwise it cannot be overriden

negue: re added it back to .svg-icon

website/client/assets/scss/modal.scss
```diff
@@ -0,0 +1,32 @@
...
+    font-size: 20px;
+    font-weight: bold;
+    line-height: 1.2;
+    text-align: center;
+    color: #4e4a57;
+  }
+
+  .text {
+    height: 60px;
+    font-family: Roboto;
```
paglias, line 29: this is not necessary, also all the height values for text css from Zeplin are not necessary, they're already going to be okay with line-heigth+font-size

PR #8905, opened 2017-07-31
Many updates on our large list

TheHollidayInn:


website/client/components/achievements/welcome.vue
```diff
@@ -0,0 +1,48 @@
+<template lang="pug">
+  b-modal#welcome(:title="$t('welcomeToHabit')", size='lg', :hide-footer="true")
+    .modal-body.container-fluid
+      .row
+        .col-4.col-centered
+          span(style='display:flex')
```
paglias, line 6: can we try not to use inline styles?

website/client/components/appFooter.vue
```diff
@@ -1,46 +1,72 @@
-<template lang="pug">
+<template lang='pug'>
```
paglias, line 1: I don't remember exactly why but initally we had problems with vue + pug when using single quotes for attributes

website/client/components/appFooter.vue
```diff
@@ -99,125 +142,103 @@
...
 
 <script>
+import axios from 'axios';
+import moment from 'moment';
+import { mapState } from 'client/libs/store';
 import gryphon from 'assets/svg/gryphon.svg';
 
+import modifyInventory from './modifyInventory';
+
 // const IS_PRODUCTION = process.env.NODE_ENV === 'production'; // eslint-disable-line no-process-env
```
paglias, line 152: we should import this with `import()` when calling modifyInventory() so it's not loaded for all users

website/client/components/appFooter.vue
TheHollidayInn: Yea, when there are single quotes, the syntax highlighter doesn't really work

website/client/components/achievements/welcome.vue
TheHollidayInn: Just copied from the live site. I must have missed this.

PR #8911, opened 2017-08-02
Fix Dockerfile's to Run App (Fixes https://github.com/HabitRPG/habitica/issues/8909 & 8910)

### LINKED ISSUE #8909: Docker container cannot start the application after being built
### General Info
  * UUID: 26d0ea00-bad1-4ec2-8531-918d522e6e47
  * Browser: Firefox
  * OS: macOS Sierra 10.12.6

### Description
I will open a PR after creating this as I have a local fix. I'm citing the [Docker section of Setting Up Locally](http://habitica.wikia.com/wiki/Setting_up_Habitica_Locally#Docker) wiki where appropriate in the description.

One of the instructions in the general wiki is to copy config.json.example to config.json in order to run the website. That's all well and good, but the Dockerfile on its initial build checks out the source code and then immediately starts the application, which will crash every time. If you're going to immediately stop the container and mount your local copy, that's fine, but if you want to test something else out (e.g., an iOS fix), Docker will be broken.

In the Dockerfile, after the line `RUN git clone https://github.com/HabitRPG/habitica.git /usr/src/habitrpg`, add `RUN cp config.json.example config.json`. This will run the cp command on the newly checked out code in the image itself, and then any container made from the image should start properly.

#### Steps to Reproduce
0. Install [Docker](https://www.docker.com/) for your system
1. Clone the source repo somewhere
2. From the repo, run `docker-compose up -d` from the command line and let it do its thing (takes a bit of time because it does a complete checkout)
3. After you regain control of your CLI (that's what the -d flag is for), try to navigate to where the site is running (localhost:3000) and get some errors
4. You can jump into the container with `docker-compose exec web sh` and then run `npm start` to see the output

#### Console Errors
I can add some if you'd like. The error I remember seeing was a complaint about there not being a Slack URL because there was no config file.

### LINKED ISSUE #8910: Docker container does not mount local source correctly to run the app
### General Info
  * UUID: 26d0ea00-bad1-4ec2-8531-918d522e6e47
  * Browser: Firefox
  * OS: macOS Sierra 10.12.6

### Description
I will open a PR after creating this as I have a local fix. I'm citing the [Docker section of Setting Up Locally](http://habitica.wikia.com/wiki/Setting_up_Habitica_Locally#Docker) wiki where appropriate in the description.

The wiki instructions to mount a local copy of the source code into the container are all appropriate, but the container won't actually run the source code. The Dockerfile assumes the application is in /usr/src/habitrpg when you start it up, and the steps to mount your local drive behave the same way, but the docker-compose.dev.yml file will instead mount your local copy in /habitrpg instead of /usr/src/habitrpg when you run the master container with `docker-compose -f docker-compose.yml -f docker-compose.dev.yml up -d`. It may be a change with Docker since the file was written (I think the intent was to base the docker-compose.dev.yml file off of the Dockerfile WORKDIR, but that doesn't work now at least).

The solution is, in docker-compose.dev.yml, to change the last line `'.:/habitrpg'` to `'.:/usr/src/habitrpg'`. That will mount your local copy in a place where just running the docker-compose command will start your copy of the app.

#### Steps to Reproduce
Running all of the steps in the [Docker section of the setup wiki](http://habitica.wikia.com/wiki/Setting_up_Habitica_Locally#Docker) up until the `docker-compose -f docker-compose.yml -f docker-compose.dev.yml up -d` command will get you there. To see the issue, create a new file in your local copy that should be mounted, and then either hop into your container and view the home directory (`docker-compose exec web sh` and then `ls`) or output the directory contents (`docker-compose exec web ls`) to see your new file isn't there. Bonus points: check out the /habitrpg directory in the container to see all your stuff.

#### Console Errors
There aren't really any errors with this approach unless you're also affected by #8909.

BennyHarv3:
Fixes #8909 https://github.com/HabitRPG/habitica/issues/8909
Fixes #8910 https://github.com/HabitRPG/habitica/issues/8910

### Changes
For 8909, I added the command RUN cp config.json.example config.json so that the Docker container will have a config.json file when it tries to run the app. Previously, it would only check out the source code into the container and not copy the file, so the app would crash without intervention.

For 8910, I changed the volume in docker-compose.dev.yml to mount the local source drive to the location the Docker container runs the application from. Previously, it mounted the directory to a different location, so the user would have to enter the container and run the app manually rather than call `docker-compose up -d`.

----
UUID: 26d0ea00-bad1-4ec2-8531-918d522e6e47

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,7 @@ RUN npm install -g gulp grunt-cli bower mocha
 RUN mkdir -p /usr/src/habitrpg
 WORKDIR /usr/src/habitrpg
 RUN git clone https://github.com/HabitRPG/habitica.git /usr/src/habitrpg
+RUN cp config.json.example config.json
 RUN npm install
 RUN bower install --allow-root
 
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,3 +1,3 @@
 web:
   volumes:
-    - '.:/habitrpg'
+    - '.:/usr/src/habitrpg'


Alys:
The new cp command would overwrite config.json if it already existed. Are there circumstances in which that could happen? For example, in #8909 you write "... If you're going to immediately stop the container and mount your local copy, that's fine..."  -- is that a situation where you would not want an automatic copy? If so, can you change it so it does the cp command only if the file doesn't already exist?

BennyHarv3:
The Dockerfile only describes building the image, not the container. So the cp command will only run the first time and then never again. My language was imprecise, and I've updated the issue with the correct wording.

Since the cp command follows immediately after a clean checkout, it would only ever wipe out a config.json file that's committed to source (which I think will never happen?). Note that this is also an issue in the Dockerfile-Production file, but from your response in the [talk portion of the setup wiki](http://habitica.wikia.com/wiki/Talk:Setting_up_Habitica_Locally), Docker isn't used for production pushes, so I left it as-is because I wasn't sure of the far-reaching issues. We don't run the risk of wiping out a pre-existing config.json on the local disk or in a data container since we only include those in the container when starting the container.

My line of "If you're going to stop... that's fine" is meant to highlight the fact that maybe we don't really care if we always expect Docker users to only be running in development and mounting a local source drive rather than run the image straight up. But I think fixing this makes Docker the best option for a quick run of running the app locally.

BennyHarv3:
Related to the Travis build failure, it's failing on 3 tests for GET /tasks/user. Is that expected/known? Should I try and fix those in this branch?

SabreCat:
@BennyHarv3 Re Dockerfile-Production, correct, `config.json` is not needed there. The env variables are supplied by the `docker service` that manages the production swarm.

As for the /tasks/user failures, nope, that's not your doing nor anything you need to address here. It's a sporadic issue with those tests we need to figure out, unrelated.

Thanks for submitting this, I'll take a look at it shortly!

SabreCat (approved):

SabreCat:
Thanks much for this, @BennyHarv3! In conjunction with some of your iOS work, I've awarded you contributor tier 2.

PR #8918, opened 2017-08-02
[WIP] item pinning

negue:
- [x] list items in task rewards
   - [x] open buy dialog for all types
      - [ ] combined purchase api (for gold, gem & hourglass)
- [ ] pin items from
   - [x] shops - market
   - [x] shops - quest
   - [x] shops - seasonal
   - [ ] shops - time travelers
      - [x] item sets
      - [ ] animals
   - [x] buy item modal dialogs
- [x] default pinned items
   - [x] new items on class change
- [x] select a new item if a suggested item was bought
- [x] remove pinned item if bought
- [x] unpin show notification
 ~~- [ ] pin new seasonal items~~ ***moved to v2***
- [x] pin health potion and armoire
    - [x] when buying them do not unpin
    - [x] show notes in shop item popover
- [x] custom suggestions from habitica
- [x] migration to add pinned items to existing users
- [x] api should say if the item is suggested or pinned by the user itself
   - [x] apply different css

bugs:
- [ ] unknown origin in seasonal of `[Vue warn]: Error in render function: "TypeError: Cannot read property 'items' of undefined"`
- [ ] unknown origin in seasonal of `[Vue warn]: Error in render function: "TypeError: Cannot read property 'text' of undefined"`
- [ ] buyModal - pin not refreshing after toggle

### Matteo's TODO
- [x]  type from path / item?
- [x] time travelers
- [ ] mobile apps


website/client/store/actions/user.js
```diff
@@ -46,3 +46,13 @@ export async function deleteWebhook (store, payload) {
   let response = await axios.delete(`/api/v3/user/webhook/${payload.webhook.id}`);
   return response.data.data;
 }
+
+
+export async function togglePinnedItemAsync (store, params) {
+  let response = await axios.get(`/api/v3/user/toggle-pinned-item/${params.key}`);
```
paglias: can we also remove the pinned item immediately without waiting for the server response? so we don't have to wait for a roundtrip to the server before updating the UI

website/server/models/user/methods.js
```diff
@@ -299,3 +300,24 @@ schema.methods.canGetGems = async function canObtainGems () {
     return !g.isSubscribed() || g.leader === user._id || g.leaderOnly.getGems !== true;
   });
 };
+
+const officialPinnedItems = common.content.officialPinnedItems;
+
+schema.methods.togglePinnedItem = function togglePinnedItem (key) {
```
paglias: like this could go in a common ops file so that it can be used both on server and client

website/common/script/libs/shops.js
```diff
@@ -408,6 +418,7 @@ shops.getSeasonalShopCategories = function getSeasonalShopCategories (user, lang
 
       if (category.items.length > 0) {
         category.specialClass = category.items[0].specialClass;
+        category.pinKey = 'seasonal!TODO';
```
negue: any idea? sets don't have a content property

website/common/script/libs/shops.js
```diff
@@ -251,6 +257,7 @@ shops.getTimeTravelersCategories = function getTimeTravelersCategories (user, la
               notes: '',
               locked: false,
               currency: 'hourglasses',
+              pinKey: 'TODO',
```
negue: mounts / pets also don't have a content path, just a bool property

paglias: Are mounts and pets pinnable? I think you can only pin gem or gold purchasable items

negue: every item in the shops should be pinnable

PR #8919, opened 2017-08-02
Refactor api description with @apiParam groups to make it more explicit

asmfreak:
apiDoc has a neat feature of `@apiParam` groups. They are use in several places throughout the apiDoc, but some APIs are not consistent. This somewhat fuzzes the apiDoc an make it less readable.

### Changes
The goal of this PR is to make those groups consistent in all of existing APIs. I will slowly refactor all api modules.

----
UUID: 1f740057-8e11-4ca3-a9b2-54886615dca6

### DIFF
--- a/website/server/controllers/api-v3/auth.js
+++ b/website/server/controllers/api-v3/auth.js
@@ -436,8 +436,8 @@ api.pusherAuth = {
  * @apiName UpdateUsername
  * @apiGroup User
  *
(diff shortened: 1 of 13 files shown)


SabreCat:
Very tidy! Thanks for this @ASMfreaK, I've added it as credit toward your Blacksmith tier 2.

PR #8921, opened 2017-08-02
New client tour

TheHollidayInn:


website/client/components/notifications.vue
```diff
@@ -223,14 +224,21 @@ export default {
       if (after !== true) return;
       this.$root.$emit('show::modal', 'quest-invitation');
     },
-    userDailies () {
+    userTasks () {
+      // @TODO: Is this the best way to check for loaded?
```
paglias, line 228: you can just do Promise.all(['user.fetch', 'tasks.fetchUserTasks']).then(), it's the most safe way

paglias: or watch for tasks.loadingStatus === 'LOADED'

PR #8949, opened 2017-08-14
Added avatars to chat

TheHollidayInn:


website/client/components/chat/chatMessages.vue
```diff
@@ -145,7 +154,29 @@ export default {
...
+      let promises = [];
+
+      this.messages.forEach(message => {
+        let uuid = message.uuid;
+        if (!this.cachedProfileData[uuid]) {
+          promises.push(axios.get(`/api/v3/members/${uuid}`));
+        }
+      });
+
+      let results = await Bluebird.all(promises);
```
paglias, line 174: You should be able to use Promise.all, without loading Bluebird

TheHollidayInn: Is there a difference? I'll switch to Promise.all, but I am just curious.

paglias: No difference but I think it's the first time I've seen bluebird used in the client so if we avoid loading it is one less dependency

TheHollidayInn: Sounds good to me!

PR #8952, opened 2017-08-16
add an Earnable Quests group for Basi-List and Dust Bunnnies (new client fix)

Alys:
This modifies `website/common/script/content/quests.js` to allow Basi-List and Dust Bunnies to be automatically placed in their own group on the Quests Shop page in the new client. I.e., it fixes this bug from our testing doc:

image

Like so:

image

It doesn't seem to have any effect on the old client:

image

@vIiRuS Will this have a negative effect on either of the mobile apps?


Alys:
Also adjusts the names of the Vice and Recidivate quest groups:

image

PR #8956, opened 2017-08-16
[WIP] Client/multiple fixes shops

negue:
- reposition FlyingPigs & Hydra images
- hide count-badge if zero
- fix sorting of hatchable pets
- always show an animal name
- correct featuredItems in market / quests
- fix egg / potion names
- buyModals mark if the value of a currency is not enough / change opacity / open purchase gems (once available)
- save itemRows state during the session
- able to pin featured items
- show bordered items in market buy-modal
- fix popover margins


TheHollidayInn:
Failed on client tests, but I have removed those now.

PR #8957, opened 2017-08-16
new client router changes: rename guild.vue to group.vue and GuildPage to GroupPage

Alys:
For the new client, this renames guild.vue to group.vue and in router.js it changes GuildPage to GroupPage.

It also makes a small text change for the challenge creation button.


negue:
LGTM 👍  , could you update (and merge) it

PR #8958, opened 2017-08-17
new client register & login screen wording changes

Alys:
Do not merge until approved by @lemoness / Tressley

This changes wording on register and login screens as in screenshot below.

The directions from Tressley were "change this field to 'Username or Email' for the login screen." However I've made two changes to that:
- I've used "Login Name" instead of "Username" because "Login Name" is what we use in the Settings screen where the users can change their details, and it also avoids confusion with the profile name (users typically use "username" to mean either profile or login name, but "Login Name" is unambiguous).
- I've emphasised the Email option over the Login Name option because down the line we'll probably be making changes to names and that will be less confusing for any users who are used to logging in with their email address.

Tressley, the two links below are mostly white-on-white so are very hard to read. Any suggestions?
"Already have a Habitica account? Log in"
"Don't have an account? Join Habitica!"

NB: The link colours in "By clicking the button below, you are indicating that you have read and agree to the Terms of Service and Privacy Policy" are still wrong. Will fix later (doesn't seem to be a simple fix because the links are in the locales files and aren't picking up styles in the vue file).

This PR also makes a minor change to the challenge creation button, as requested by Lemoness.

image

image


Tressley:
I think the label updates are good. Still need to brainstorm on how to prevent the white links from overlaying on the scenery.

lemoness:
Agreed about the updated labels!

Atomk:
Some thoughts looking at the screenshots:
- In the sign in buttons, the F of the Facebook logo is not aligned with the Google logo, breaking the symmetry
- The space between the upper sign in buttons and the first label ("login name") is not the same space between each field of the form, it looks like the first field is related more to the buttons than to the other fields
- The Habitica logo layout is a bit confusing, between the gryphon and "habitica" there's enough margin to make it look like they are two different things and not part of the same entuty. If they both make the logo, they should be closer to each other or even on the same line, or at least less close to other elements of the UI
- Wouldn't be simpler to use just the email instead of a login name? Too many fields look intimidating to a new user willing to register, and this mechanic (login name vs. display name) looks unnecessary complicated to me
- To make the sentence at the bottom of the page more visible you could remove the white mountains from the center of the page, to avoid adding thext shadow or a darker background

Other than these little details, this looks really good, I can't wait to see the rest of the redesign :smile:

Tressley:
Good to go @Alys .

PR #8959, opened 2017-08-17
new client various fixes: Equipment shop empty; Skills not Spells; no Saddles message

Alys:
Various small fixes for the new client.

* when you've bought all class gear, you see a message instead of an empty space
* the Skills tray on the tasks page is now labelled "Skills" not "Spells"
* when you don't have any Saddles, the tray in the Stable says that, instead of saying you have no food

I've done the change from "Spells" to "Skills" in the locales strings that are used by production, so if this is pushed to prod, it will change in the live website. However that's fine - we should have done that years ago. The only thing it affects in production is the Justin tour that introduces Skills to the user when they reach level 10/11.

image

image

image


negue:
LGTM 👍

PR #8967, opened 2017-08-18
New client tavern fixes

TheHollidayInn:


website/client/components/chat/chatMessages.vue
```diff
@@ -27,7 +27,7 @@
           span.action( @click='copyAsTodo(msg)')
             .svg-icon(v-html="icons.copy")
             | {{$t('copyAsTodo')}}
-          span.action(v-if='user.contributor.admin || (!msg.sent && user.flags.communityGuidelinesAccepted)', @click='report(msg)')
+          span.action(v-if='user.contributor.admin || (msg.uuid !== user._id && user.flags.communityGuidelinesAccepted)', @click='report(msg)')
```
Alys: It needs to be possible for a user to report their own message. It's useful occasionally. I had to report one of mine a couple of days ago beccause I replied to a user who later was revealed to be underaged without permission, and it would have been appropriate for me to flag my own message then even if I wasn't a mod. There can also be other  reasons.

TheHollidayInn: It looks like this logic was on live though, correct? I just updated it for Vue

Alys: In prod you can flag your own messages (even if you aren't a mod). I might have misread your code - does it stop a player flagging their own message?

TheHollidayInn: Then there must be something wrong with the msg.sent variable, because this line `user.contributor.admin || (!msg.sent && user.flags.communityGuidelinesAccepted)` is supposed to prevent you from flagging your own message. I'm fine allow for users to report their own message, but maybe there is a better flow rather than flagging? Either way, I just need to make sure everyone else knows.

Alys:
The current flagging flow is fine for this. When we report our own
messages, we want all the same actions to happen as when we report someone
else's (email to mods, post to Slack chat, hiding or partial hiding of the
message). This is the PR it was implemented in:
https://github.com/HabitRPG/habitica/pull/8006

Alys:
I've worked out what's happening with that original code. `!msg.sent`
applies only to the player's own inbox, so it's preventing a player from
seeing the report icon on their own sent PMs. It doesn't apply to
guild/party/tavern chat messages, which have no `sent` field.

And in fact, when flagging is enabled for PMs, we would want people to be
able to report their own sent messages because it's a reliable, unhackable
way of sending both sides of an argument to moderators, so the `!msg.sent`
part would need to be deleted for that. However since flagging in PMs is
currently broken, it doesn't really matter whether it stays or goes for now.

TheHollidayInn:
Awesome! Thanks for looking into that. I will update the code then :D

PR #8971, opened 2017-08-21
Fixes multiple complete and uncomplete for todos and daily (Fixes #8669)

### LINKED ISSUE #8669: Dailies and To-Dos can be scored up or down (completed or uncompleted) multiple times
Using the API's route for scoring a task (https://habitica.com/api/v3/tasks/:taskId/score/:direction ), you can mark a Daily or To-Do complete multiple times (i.e., without toggling it back to incomplete each time). 

If you test it, you see that all the effects of completing the task are applied each time: you gain XP for each completion, the task's value changes each time, for Dailies, the streak increases, etc.

Similarly, you can score a task down (mark it as incomplete) when it is already incomplete.

You can also test this bug by opening the website in two tabs, completing the same task in both, and then syncing to see that your XP has increased twice (the second gain in XP is *not* removed after syncing). A Daily is the best thing to use for this because then you can also see the streak increase by two.

I think the best fix for this would be for the API's score route to check to see if the task is complete / uncomplete and not allow the score to happen if it's not appropriate. I.e., if the task is complete, you cannot score up; if the task is uncompleted, you cannot score down.

Any change made for this should NOT affect Habits or Rewards, for which multiple scores in the same direction are appropriate.

asifmallik:
Fixes https://github.com/HabitRPG/habitica/issues/8669

### Changes
Fixed the issue and also added tests for it. Edited existing code that relied on this bug.
----
UUID: bc6a5bb0-b458-4526-b822-a1e63c581ce4


test/api/v3/integration/tasks/challenges/POST-tasks_challenges_challengeId_tasks_id_score_direction.test.js
```diff
@@ -82,14 +82,14 @@ describe('POST /tasks/:id/score/:direction', () => {
...
     it('should update the history', async () => {
+      await user.post(`/tasks/${usersChallengeTaskId}/score/down`);
       await user.post(`/tasks/${usersChallengeTaskId}/score/up`);
 
       let tasks = await user.get(`/tasks/challenge/${challenge._id}`);
       let task = find(tasks, {_id: daily._id});
 
       expect(task.history).to.have.lengthOf(1);
       expect(task.history[0].date).to.not.equal(previousTaskHistory.date);
-      expect(task.history[0].value).to.not.equal(previousTaskHistory.value);
```
asifmallik: https://www.youtube.com/watch?v=buLE7kIu-xI

asifmallik: Possibly, the only way to test this is by somehow changing the time to tomorrow so that the dailies are unchecked. Can cron be used?

SabreCat: Possibly you could use the approach the debug menu does, by changing `lastCron` into the past and rerunning cron?

website/common/locales/en/tasks.json
```diff
@@ -178,5 +178,6 @@
     "yesterDailiesCallToAction": "Start My New Day!",
     "yesterDailiesOptionTitle": "Confirm that this Daily wasn't done before applying damage",
     "yesterDailiesDescription": "If this setting is applied, Habitica will ask you if you meant to leave the Daily undone before calculating and applying damage to your avatar. This can protect you against unintentional damage.",
-    "repeatDayError": "Please ensure that you have at least one day of the week selected."
+    "repeatDayError": "Please ensure that you have at least one day of the week selected.",
+    "sessionOutdated": "Your session is outdated. Please reload the page."
```
SabreCat: Given this comes from the API, the user might not be interacting with a "page" at all. Maybe "Please refresh or sync"?

SabreCat:
@asifmallik Here too, will you have a chance soon to address comments and fix merge conflicts?

test/api/v3/integration/tasks/challenges/POST-tasks_challenges_challengeId_tasks_id_score_direction.test.js
asifmallik: I will try that

asifmallik: I attempted to replace

```JavaScript
await user.post(`/tasks/${usersChallengeTaskId}/score/down`);
```

with

```JavaScript
let newCron = new Date(2015, 11, 20);
await user.post('/debug/set-cron', {
    lastCron: newCron,
});

await user.sync();
expect(user.lastCron).to.eql(newCron);
```

but for some reason, this fails to update the `lastCron`. Then, I tried running the tests in `test/api/v3/integration/debug` but they fail too. Do I have to carry out any extra step before running those tests?

SabreCat: Hmm, that's peculiar. Do the tests fail for you locally even without the above change? That suggests a dev environment issue, given that the tests are passing on CI.

Does it behave any differently if you set `lastCron` directly as part of the test rather than using the debug API?

asifmallik: I made two mistakes. 1) Did not send a request to /cron 2) Forgot to quit the server before running the tests.

After fixing those two, the tests have passed successfully.

SabreCat:
Thanks for this, @asifmallik! Given your tier 1 PR was a pretty substantial one, I've gone ahead and awarded you tier 2 Blacksmith credit for your continuing effort. Congratulations!

PR #8976, opened 2017-08-21
Choose Class modal

SabreCat:

### Changes

Implements the level 10 Choose Class modal in the new client, to spec but for some fancy borders I'm still figuring out.

Also corrects the placement of the avatar when previewing equipment to be worn.


website/client/components/achievements/chooseClass.vue
```diff
@@ -1,119 +1,206 @@
...
+    markdown: markdownDirective,
   },
   methods: {
     close () {
       this.$root.$emit('hide::modal', 'choose-class');
     },
-    disableClasses () {
-      // @TODO:
+    async disableClasses (user) {
+      await axios.post('/api/v3/user/disable-classes');
```
paglias: both this and changeClass should be actions on the user (store/actions/user) not defined directly in the component, also you shouldn't use await before calling the local disableClass or it'll wait for the server response and defeat the purpose of applying the changes locally so they appear immediately and not after the server response

website/client/components/achievements/chooseClass.vue
```diff
@@ -1,119 +1,206 @@
...
+    bPopover,
     Avatar,
   },
+  computed: {
+    ...mapState({user: 'user.data'}),
+  },
   data () {
     return {
-      maxHealth,
+      heroClasses: ['warrior', 'wizard', 'rogue', 'healer'],
```
paglias: the classes are already defined as constants in common/content/constants and you should be able to access them inside the store's state as `constants.CLASSES`

website/common/locales/en/newClient.json
```diff
@@ -288,6 +288,11 @@
   "haveHatchablePet": "You have a <%= potion %> hatching potion and <%= egg %> egg to hatch this pet! <b>Click</b> the paw print to hatch.",
   "welcomeBack": "Welcome back!",
   "checkOffYesterDailies": "Check off any Dailies you did yesterday:",
+  "wizardText": "Mages learn swiftly, gaining Experience and Levels faster than other classes. They also get a great deal of Mana for using special abilities. Play a Mage if you enjoy the tactical game aspects of Habitica, or if you are strongly motivated by leveling up and unlocking advanced features!",
+  "chooseClass": "Choose your Class",
+  "chooseClassLearnMarkdown": "[Learn more about Habitica's class system](http://habitica.wikia.com/wiki/Class_System)",
+  "selectClass": "Select <%= heroClass %>",
+  "wizard": "Mage",
```
paglias, line 295: this already exists in the `mage` string

website/common/locales/en/newClient.json
```diff
@@ -288,6 +288,11 @@
   "haveHatchablePet": "You have a <%= potion %> hatching potion and <%= egg %> egg to hatch this pet! <b>Click</b> the paw print to hatch.",
   "welcomeBack": "Welcome back!",
   "checkOffYesterDailies": "Check off any Dailies you did yesterday:",
+  "wizardText": "Mages learn swiftly, gaining Experience and Levels faster than other classes. They also get a great deal of Mana for using special abilities. Play a Mage if you enjoy the tactical game aspects of Habitica, or if you are strongly motivated by leveling up and unlocking advanced features!",
```
paglias, line 291: exist as mageText

website/client/components/inventory/equipment/equipGearModal.vue
```diff
@@ -65,6 +65,11 @@
       cursor: inherit;
     }
 
+    .character-sprites {
```
paglias: cannot this be passed as the margin prop to avatar since you added that?

website/client/components/avatar.vue
```diff
@@ -1,6 +1,6 @@
 <template lang="pug">
 .avatar(:style="{width, height, paddingTop}", :class="backgroundClass", @click.prevent='castEnd()')
-  .character-sprites
+  .character-sprites(:style='{margin}')
```
paglias: i think this margin prop is confusing as it's passed to avatar but applies to an element inside

website/client/components/achievements/chooseClass.vue
```diff
@@ -1,119 +1,206 @@
...
+          .d-inline-flex(v-if='selectedClass === heroClass')
+            .class-badge.d-flex.justify-content-center
+              .svg-icon.align-self-center(v-html='icons[heroClass]')
+            .class-name(:class='`${heroClass}-color`') {{ $t(heroClass) }}
+        div(v-for='heroClass in heroClasses')
+          .class-explanation.text-center(v-if='selectedClass === heroClass') {{ $t(`${heroClass}Text`) }}
+        .text-center(v-markdown='$t("chooseClassLearnMarkdown")')
+        .modal-actions.text-center
+          button.btn.btn-primary.d-inline-block(v-if='!selectedClass', :disabled='true') {{ $t('select') }}
+          button.btn.btn-primary.d-inline-block(v-if='selectedClass', @click='changeClass(selectedClass); selectedClass = undefined; close();') {{ $t('selectClass', {heroClass: $t(selectedClass)}) }}
```
paglias: imho the @click has too much code, it should be in a function

paglias: you can use v-else if you want since it's the opposite of an adjacent v-if

paglias (changes requested):

website/common/locales/en/newClient.json
SabreCat: Would it be such a bad thing to replace it? This way we can iterate over the classes list without having to hardcode an exception for mage/wizard.

website/client/components/avatar.vue
SabreCat: Yeah, that's tricky. Would it be sufficient to rename it to `inner-margin` or `sprites-margin`?

website/common/locales/en/newClient.json
SabreCat: see question below

website/common/locales/en/newClient.json
paglias: I'm fine with renaming the strings to wizard but we should make sure it's used everywhere in the new client. In particular for the string below that is quite long, so that we don't have to remember to update it in two places

website/client/components/avatar.vue
paglias: Yeah I think so

PR #8977, opened 2017-08-22
new client - various fixes: Earnable Quests removed, equipment type names and sort order changed

Alys:
Various fixes for the new client (other commits will be added).

- Removes Earnable Quests from Quests Shop because there's currently no easy/neat way to explain how you can buy them. NB We don't have approval from Lemoness/designers for this approach but it's a tier 1 issue.
image
- Changes name of some equipment types and changes sort order:
img

Regarding the renaming of the equipment types:
The `...Capitalized` strings all have lower-case versions (e.g., "weapon" and "weaponCapitalized"). I suspect we should change or get rid of the uncapitalized version but I'm not sure what other effects that might have (e.g., haven't searched for exactly where it's used; don't know what it would do to the mobile apps). I suggest looking into that after launch. If the old, uncapitalized strings are noticed elsewhere during our site testing, we should probably change them to the Capitalized version then, or if lowercase is essential, create new strings (e.g., "weaponUncapitalized" - that will be much easier to search for than just "weapon" for future code changes when we want to know where it's used).

### DIFF
--- a/website/client/components/inventory/equipment/index.vue
+++ b/website/client/components/inventory/equipment/index.vue
@@ -183,12 +183,12 @@ export default {
       costume: false,
       groupBy: 'type', // or 'class'
       gearTypesToStrings: Object.freeze({ // TODO use content.itemList?
-        headAccessory: i18n.t('headAccessoryCapitalized'),
-        head: i18n.t('headgearCapitalized'),
-        eyewear: i18n.t('eyewear'),
         weapon: i18n.t('weaponCapitalized'),
         shield: i18n.t('offhandCapitalized'),
+        head: i18n.t('headgearCapitalized'),
         armor: i18n.t('armorCapitalized'),
+        headAccessory: i18n.t('headAccessoryCapitalized'),
+        eyewear: i18n.t('eyewear'),
         body: i18n.t('body'),
         back: i18n.t('back'),
       }),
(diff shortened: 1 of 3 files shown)


Alys:
DO NOT MERGE. Waiting for approval for gear order. Skils change isn't complete yet.

- Changes name of some equipment types and changes sort order:
img
- [WIP] Skill/spell fixes such as making Searing Brightness apply to all tasks.

Regarding the renaming of the equipment types:
The `...Capitalized` strings all have lower-case versions (e.g., "weapon" and "weaponCapitalized"). I suspect we should change or get rid of the uncapitalized version but I'm not sure what other effects that might have (e.g., haven't searched for exactly where it's used; don't know what it would do to the mobile apps). I suggest looking into that after launch. If the old, uncapitalized strings are noticed elsewhere during our site testing, we should probably change them to the Capitalized version then, or if lowercase is essential, create new strings (e.g., "weaponUncapitalized" - that will be much easier to search for than just "weapon" for future code changes when we want to know where it's used).

Alys:
I've removed my commit related to skills from this PR because the skill problem I was trying to fix needs more than a small change. I'm submitting it into a new PR: https://github.com/HabitRPG/habitica/pull/8978

Alys:
Sara has approved the sort order for the Equipment types.

Alys:
Lemoness has approved the removal of the Earnable Quests. We'll put them back after go-live when we have more time to design a neat explanation for them.

PR #8978, opened 2017-08-22
[WIP] new client: allow Searing Brightness to act on all skills and make other skills fixes

Alys:
For the new client, this PR:

- starts to fix the problem reported by SL: "I believe that Searing Brightness should affect ALL tasks automatically, not a single task. Right now it’s telling me to cast it on a specific task. [tier 1]"
- adds TODO comments for some other things that still need fixing related to skills
- comments out some incomplete quests code that was throwing errors into the javascript console
- adjusts other code a little to help prevent bugs from typos in parameters being passed around (use `spell.target` directly instead of specifying its value)

@TheHollidayInn Can you take this over? It requires more knowledge of how data is being passed around, especially related to tasks, than I currently have, so you'd be much faster at it than I would. I'll go back to the easy things. :)

### DIFF
--- a/website/client/components/tasks/spells.vue
+++ b/website/client/components/tasks/spells.vue
@@ -5,6 +5,7 @@ div(v-if='user.stats.lvl > 10')
       .col-8.details
         p.title {{spell.text()}}
         p.notes {{ `Click on a ${spell.target} to cast!`}}
+        // @TODO make that translatable. We'll need locales strings for the targets: `task` and `user` ("party member")
       .col-4.mana
         .img(:class='`shop_${spell.key} shop-sprite item-img`')
 
@@ -236,35 +237,39 @@ export default {
       this.spell = spell;
 
       if (spell.target === 'self') {
-        this.castEnd(null, 'self');
+        this.castEnd(null, spell.target);
       } else if (spell.target === 'party') {
         if (!this.user.party._id) {
           let party = [this.user];
-          this.castEnd(party, 'party');
+          this.castEnd(party, spell.target);
           return;
         }
-
-        // @TODO: do we need to fetcht the party everytime? We should probably just check store
+        // @TODO: do we need to fetch the party everytime? We should probably just check store. But what if a new member has arrived recently?
         let party = await this.$store.dispatch('guilds:getGroup', {groupId: 'party'});
         party = isArray(party) ? party : [];
         party = party.concat(this.user);
         this.$store.state.party.data = party;
-        this.castEnd(party, 'party');
-      } else if (spell.target === 'tasks') {
+        this.castEnd(party, spell.target);
+      } else if (spell.target === 'tasks' || spell.target === 'task') {
+        // cast on all 'tasks' or one 'task' (Rewards are not allowed)
+        if (spell.target === 'tasks') this.potionClickMode = false;
+        // @TODO make this refer to tasks correctly. Currently getting `Unhandled promise rejection TypeError: this.$store.state.tasks.habits is undefined`
         let tasks = this.$store.state.tasks.habits.concat(this.user.dailys)
-          .concat(this.$store.state.tasks.rewards)
           .concat(this.$store.state.tasks.todos);
         // exclude challenge tasks
         tasks = tasks.filter((task) => {
           if (!task.challenge) return true;
           return !task.challenge.id || task.challenge.broken;
         });
-        this.castEnd(tasks, 'tasks');
+        this.castEnd(tasks, spell.target);
+      // } else if (spell.target === 'user') {
+        // avatar transformation buffs and greeting cards (e.g., shinySeed, valentine)
+        // @TODO need to do anything here?
       }
     },
     async castEnd (target, type) {
       if (!this.$store.state.spellOptions.castingSpell) return;
-      let beforeQuestProgress = this.questProgress();
+      // let beforeQuestProgress = this.questProgress();
 
       if (!this.applyingAction) return 'No applying action';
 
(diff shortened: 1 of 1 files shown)


website/client/components/tasks/spells.vue
```diff
@@ -236,35 +237,39 @@ export default {
       this.spell = spell;
 
       if (spell.target === 'self') {
-        this.castEnd(null, 'self');
+        this.castEnd(null, spell.target);
```
Alys, line 240: @TheHollidayInn  This change from `'self''` to `spell.target` (and the similar changes below) aren't strictly necessary but make the code a bit clearer I believe, since it makes it obvious you're just passing the target through to the next stage. It also avoids a chance of a future bug if code is copy-pasted to make new code for a similar purpose -- no need to change '`self'` in two places.

website/client/components/tasks/spells.vue
```diff
@@ -236,35 +237,39 @@ export default {
...
       } else if (spell.target === 'party') {
         if (!this.user.party._id) {
           let party = [this.user];
-          this.castEnd(party, 'party');
+          this.castEnd(party, spell.target);
           return;
         }
-
-        // @TODO: do we need to fetcht the party everytime? We should probably just check store
+        // @TODO: do we need to fetch the party everytime? We should probably just check store. But what if a new member has arrived recently?
```
Alys, line 247: I added the question on the end of this comment. I don't know if it's relevant though - if it's not, just ignore it.

website/client/components/tasks/spells.vue
```diff
@@ -236,35 +237,39 @@ export default {
...
         let party = await this.$store.dispatch('guilds:getGroup', {groupId: 'party'});
         party = isArray(party) ? party : [];
         party = party.concat(this.user);
         this.$store.state.party.data = party;
-        this.castEnd(party, 'party');
-      } else if (spell.target === 'tasks') {
+        this.castEnd(party, spell.target);
+      } else if (spell.target === 'tasks' || spell.target === 'task') {
+        // cast on all 'tasks' or one 'task' (Rewards are not allowed)
```
Alys, line 254: You definitely need that. Skills that act on one task have a target of `task`, skills that act on all, like Searing Brightness, have a target of `tasks`. Search for "brightness" and "fireball" in website/common/script/content/spells.js for examples. I suggest including the comment to make that clearer for future programmers too, but feel free to improve the comment because it's clearly not clear enough. :)

website/client/components/tasks/spells.vue
```diff
@@ -236,35 +237,39 @@ export default {
...
+        if (spell.target === 'tasks') this.potionClickMode = false;
```
Alys, line 255: This line stops the mouse cursor changing to have a message about clicking on a single task when you're casting Searing Brightness. However there might be a better way to do that.

website/client/components/tasks/spells.vue
```diff
@@ -236,35 +237,39 @@ export default {
...
+        // @TODO make this refer to tasks correctly. Currently getting `Unhandled promise rejection TypeError: this.$store.state.tasks.habits is undefined`
         let tasks = this.$store.state.tasks.habits.concat(this.user.dailys)
-          .concat(this.$store.state.tasks.rewards)
```
Alys, line 255: Skills that act on tasks should not act on Rewards. If the current prod site lets them act on Rewards, it's wrong.

website/client/components/tasks/spells.vue
```diff
@@ -236,35 +237,39 @@ export default {
...
         party = isArray(party) ? party : [];
         party = party.concat(this.user);
         this.$store.state.party.data = party;
-        this.castEnd(party, 'party');
-      } else if (spell.target === 'tasks') {
+        this.castEnd(party, spell.target);
+      } else if (spell.target === 'tasks' || spell.target === 'task') {
+        // cast on all 'tasks' or one 'task' (Rewards are not allowed)
+        if (spell.target === 'tasks') this.potionClickMode = false;
+        // @TODO make this refer to tasks correctly. Currently getting `Unhandled promise rejection TypeError: this.$store.state.tasks.habits is undefined`
```
Alys, line 256: This just needs your fix to use `userTasks.habits` etc.

website/client/components/tasks/spells.vue
```diff
@@ -236,35 +237,39 @@ export default {
...
         let tasks = this.$store.state.tasks.habits.concat(this.user.dailys)
-          .concat(this.$store.state.tasks.rewards)
           .concat(this.$store.state.tasks.todos);
         // exclude challenge tasks
         tasks = tasks.filter((task) => {
           if (!task.challenge) return true;
           return !task.challenge.id || task.challenge.broken;
         });
-        this.castEnd(tasks, 'tasks');
+        this.castEnd(tasks, spell.target);
```
Alys, line 264: Using `spell.target` here instead of `'tasks'` is essential since this block is using for both `tasks` and `task`.

Another reason to use `spell.target` in the other locations is for consistency with this line.

NB I believe `tasks` and `task` should be handled in the same block because they have very similar code. E.g., **not** replicating the code to find and filter tasks ensures that if we ever change that code, we can't change it in one place and forget the other.

website/client/components/tasks/spells.vue
```diff
@@ -236,35 +237,39 @@ export default {
...
+      // } else if (spell.target === 'user') {
+        // avatar transformation buffs and greeting cards (e.g., shinySeed, valentine)
+        // @TODO need to do anything here?
```
Alys, line 267: Delete these comments if we don't need to handle `user` spells here. I wasn't sure how widely this code is used.

website/client/components/tasks/spells.vue
```diff
@@ -236,35 +237,39 @@ export default {
...
       }
     },
     async castEnd (target, type) {
       if (!this.$store.state.spellOptions.castingSpell) return;
-      let beforeQuestProgress = this.questProgress();
+      // let beforeQuestProgress = this.questProgress();
```
Alys, line 272: I commented this out, and the quest code below, just to stop it spewing errors into the browser's javascript console. Uncomment when the code for handling quest progress is finished.

website/client/components/tasks/spells.vue
```diff
@@ -324,16 +329,18 @@ export default {
       }
 
       this.markdown(msg); // @TODO: mardown directive?
+      // @TODO: wrap the quest code below in an `if` for only the two boss attack skills? Collection quests aren't directly affected by skills.
+      // I've commented it out to stop console errors (also `let beforeQuestProgress` above). -- Alys
```
Alys, line 333: Brutal Smash and Burst of Flames increase boss damage (i.e., increase user.party.quest.progress.up). None of the other skills directly change user.party.quest.

Alys (commented):
notes for @TheHollidayInn

PR #8994, opened 2017-08-27
new client: fix warning: "component lists rendered with v-for should have explicit keys"

Alys:
My local install has been outputting the warning below for a couple of days. This PR prevents the warning but I don't know if this is the best way to do it so feel free to close if it's not, or tell me how to fix it better.

```
WARNING in ./~/vue-loader/lib/template-compiler?{"id":"data-v-78f7211c"}!./~/vue-loader/lib/template-compiler/preprocessor.js?raw&engine=pug!./~/vue-loader/lib/selector.js?type=template&index=0!./website/client/components/snackbars/notifications.vue
(Emitted value instead of an instance of Error) <notification v-for="notification in notifications">: component lists rendered with v-for should have explicit keys. See https://vuejs.org/guide/list.html#key for more info.
 @ ./website/client/components/snackbars/notifications.vue 9:2-284
 @ ./~/babel-loader/lib!./~/vue-loader/lib/selector.js?type=script&index=0!./website/client/app.vue
 @ ./website/client/app.vue
 @ ./website/client/main.js
 @ multi ./webpack/dev-client ./website/client/main.js
```

### DIFF
--- a/website/client/components/snackbars/notifications.vue
+++ b/website/client/components/snackbars/notifications.vue
@@ -1,6 +1,6 @@
 <template lang="pug">
 .notifications
-  notification(v-for='notification in notifications', :notification='notification')
+  notification(v-for='notification in notifications', :notification='notification', :key='notification.id')
 </template>
 
 <style lang="scss" scoped>


paglias:
Hmm I'm unsure if the notifications we're talking about are from user.notifications and have an id property or are just client side notifications with no id

Alys:
Ah, of course, you're right! I was thinking of the wrong kind of notification. 

These ones have notification.type and notification.text (and other attributes). I wonder if we can use type or text just to make the warning message go away. The :key seems to be used by vue primarily to help when the order of items changes, which isn't so relevant for notifications. 

Or we could put up with the warning always being there, but I'd like to avoid that if possible since clutter in the output can make real problems harder to see.

This is probably better looked at later though since it's not as trivial as I thought.

paglias:
We should definitely fix this but the key must be unique

PR #9004, opened 2017-08-29
Miscing groups

TheHollidayInn:
- Fixed party invite accepting
- Fixed inbox sending and scrolling
- Fixed profile click to open modal
- Hid needs work from group tasks
- Added group plan page images
- Added categories to editing groups
- Fixed party quest border
- Added tier styling to chat


webpack/config/index.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
   },
   dev: {
     env: devEnv,
-    port: 8080,
+    port: 8081,
```
paglias: is this necessary?

TheHollidayInn: No, this was an accident.

PR #9006, opened 2017-08-30
More fixes

TheHollidayInn:
- Fixed issue with showing wrong user in profile modal
- Added editing profile
- Added scroll to conversations 
- Added scroll to bottom after adding new message in inbox
- Changed notification when messaging user from their detail modal
- Fixed editing challenge loading values from selected challenge
- Updated guild list to return summary
- Updated tier icons in chat
- Removed extra loading of part members on modal show
- Removed avatar background in tutorial
- Fixed task adding when clicking done on tutorial

### DIFF
--- a/website/client/components/creatorIntro.vue
+++ b/website/client/components/creatorIntro.vue
@@ -282,7 +282,7 @@ b-modal#avatar-modal(title="", size='lg', :hide-header='true', :hide-footer='tru
           .next(v-once) {{$t('next')}}
           .next-arrow
         div(v-if='modalPage === 3', @click='done()')
-          button.btn.btn-primary.next(v-once) {{$t('done')}}
+          button.btn.btn-primary.next(v-once, v-if='!loading') {{$t('done')}}
 </template>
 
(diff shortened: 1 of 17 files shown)


website/client/components/creatorIntro.vue
```diff
@@ -905,7 +908,7 @@ export default {
       let tasks = response.data.data;
       tasks.forEach(task => {
         this.$store.state.user.data.tasksOrder[`${task.type}s`].unshift(task._id);
-        Object.assign(this.$store.state.tasks.data[`${task.type}s`][0], task);
+        this.$store.state.tasks.data[`${task.type}s`].unshift(task);
```
paglias, line 911: can we use the tasks create action here? the code is the same

paglias (changes requested):

website/client/components/creatorIntro.vue
TheHollidayInn: It is a little different, but I can modify the store.

PR #9018, opened 2017-09-02
[WIP] new client - seasonal-shop

negue:
- [x] Config-File for all needed settings `shops-seasonal.config.js`:
   - `opened`: false | true
   - `currentSeason`: 'Closed'| 'Name'
   - `availableSets`: [ `setKey` ]
   - `availableSpells`: [ `spellItemId` ]
   - `availableQuests`: [ `questId` ]
   - `pinnedSets`: { `class`: `setKey` }
   - `featuredSet`: `setKey`
- [x] Refactor to use `setKey` instead of event-name to setKey map
- [x] Show featured set `The Bat-tle Rogue set`
- [x] `During a Seasonal Event, the new class items will appear pinned to the top of the Rewards column as suggested Habitica items.`
- [x] Styling of Suggested Pinned Items 
- [x] `Seasonal Event Items display their limited status.`
- [x] return featured items with the shops api `featured`: { `text`, `items` }
  - [x] market
  - [x] quest 
  - [x] seasonal
  - [x] time-travelers
- [x] add the set-key to each special equipment
- [x] fix: buy seasonal spell
- [x] filter: by gear types
- [x] purchase the current seasonal gear with gold instead of gem
- [x] FALL FESTIVAL Sets @SabreCat
  - [x] text / notes of new gear


website/common/script/libs/getItemInfo.js
```diff
@@ -40,10 +39,15 @@ function getDefaultGearProps (item, language) {
     per: item.per,
     con: item.con,
     klass: item.klass,
+    event: item.event,
   };
 }
 
-module.exports = function getItemInfo (user, type, item, language = 'en') {
+module.exports = function getItemInfo (user, type, item, officialPinnedItems, language = 'en') {
```
paglias, line 47: why pass the official pinned items manually when you can get them with a call?

negue: I just thought that `getItemInfo` is called sometimes for a lot of items, it would be bad for the performance to recreate that array each time for each item

paglias: 👍

paglias: Actually I just noticed that a few lines below if you do not pass the official pinned items they're calculated here so it doesn't make any difference to pass them manually (lines 47-49)

website/common/script/content/gear/sets/special/index.js
```diff
@@ -8,7 +8,7 @@ import takeThisGear from './special-takeThis';
 import wonderconGear from './special-wondercon';
 import t from '../../../translation';
 
-const CURRENT_SEASON = 'None';
+const CURRENT_SEASON = 'Fall';
```
negue: shouldn't this be `fall` in lowercase ?

website/common/script/libs/getItemInfo.js
negue: yes some calls to `getItemInfo` don't add the officialPinned-Array (which only needed for performance reasons on multiple items), so if the parameter `officialPinnedItems` is undefined the items get calculated as a fallback

PR #9027, opened 2017-09-12
Redesign: Static Pages Wrapper

SabreCat:

Addresses this comment by @paglias regarding static pages in the redesign:

> All those pages are including `@import '~client/assets/scss/static.scss';` in the component’s CSS: I’m quite sure that the css inside that file is duplicate right now with one version included for every static page. I think you should change the rules there that are very broad (all h1, ..) to some classes to load inside `index.scss` like `.static-h1` and use those
>^ another option, maybe it’s quicker, is to create a wrapper component called [`staticWrapper.vue`] that includes that scss file and renders the correct page based on the route

...using the latter of the suggestions to address the situation.

### Changes

* Implements a `staticWrapper.vue` file that imports and displays the various static pages according to Vue routing. This cuts down on repetition of SCSS elements and importing of header and footer components.
* Comments out some static pages we may not even want or need at this time; easy enough to add them back in when relevant.


website/client/components/static/staticWrapper.vue
```diff
@@ -0,0 +1,88 @@
+<template lang='pug'>
+  div
+    static-header
+
+    app-page(v-if='$route.name === "app"')
```
paglias: instead of adding all these `v-if`s you should use something like this https://github.com/HabitRPG/habitica/blob/develop/website/client/router.js#L107-L115, then the wrapper will just need to contain a `router-view` element like here https://github.com/HabitRPG/habitica/blob/develop/website/client/components/inventory/index.vue#L8. The only thing is that this way /home will have to be a children of /static so it'll become /static/home and the link the in the static header will have to be changed

paglias: instead of adding all these `v-if`s you should use something like this https://github.com/HabitRPG/habitica/blob/develop/website/client/router.js#L107-L115, then the wrapper will just need to contain a `router-view` element like here https://github.com/HabitRPG/habitica/blob/develop/website/client/components/inventory/index.vue#L8. The only thing is that this way /home will have to be a children of /static so it'll become /static/home and the link the in the static header will have to be changed

website/client/components/static/staticWrapper.vue
```diff
@@ -0,0 +1,88 @@
...
+    merch-page(v-if='$route.name === "merch"')
+    overview-page(v-if='$route.name === "overview"')
+    press-kit(v-if='$route.name === "pressKit"')
+    privacy-page(v-if='$route.name === "privacy"')
+    terms-of-use(v-if='$route.name === "terms"')
+
+    app-footer
+</template>
+
+<style lang='scss'>
```
paglias: this should be `scoped` but probably won't work when you include child components so I'd say give a class `.static-wrapper` to the root div and put all the css rules under it like

```
.static-wrapper {
    h1, h2 { ... }
}
```

Otherwise this rules will become global and could affect all pages of the site

paglias (changes requested):
Looks good apart from the two comments!

PR #9033, opened 2017-09-12
new client misc fixes

negue:
- extracted `questDialogDrops`and `questDialogContent` to have the same content/drops in both quest-dialogs
- add quests to items inventory
- fix market search
- reset pinned gear on revive


website/client/components/groups/startQuestModal.vue
```diff
@@ -197,6 +142,10 @@ export default {
   mounted () {
     let questKeys = Object.keys(this.user.items.quests);
     this.selectedQuest = questKeys[0];
+
+    this.$root.$on('selectQuest', (quest) => {
```
paglias: this listener should be removed when the component is removed (this.$root.$off

paglias (changes requested):

PR #9034, opened 2017-09-13
(WIP) Allow user to save manual timezone (partial fix for #8794)

### LINKED ISSUE #8794: allow users to specify what time zone they are in
~This is currently marked as suggestion-discussion and we are still considering it. It might not be implemented. In a couple of days if there's no objections, I'll probably change it to help welcome.~

Currently, the Habitica website reads your time zone from your browser (which gets it from your OS). The mobile apps read it from your mobile device. 

We will keep those mechanisms but we are considering also allowing users to specify a time zone which would override the values that Habitica gets automatically. This might be useful in cases where people are travelling but want to keep their Dailies on their home time schedule or choose the exact time at which the zone changes. It would also be useful for people who use VPNs that are hosted in different locations than they live in (especially on Windows 10 which by default will change your time zone automatically based on where it thinks you are).

The disadvantage is that some uses would use this feature incorrectly when they don't understand time zones and would then find that cron was running at the wrong time. For example, we sometimes see users report a bug in the automatic time zone mechanism because they don't understand that their zone changes when daylight savings starts or stops. They would be likely to try to set the zone manually and choose the wrong zone, leading to further bug reports and wasted time for them and us.

Because of that, if we add this feature, it would need to have a clear warning that it should only be used when essential and that it's usually best to let Habitica detect your zone automatically. 

The user must still be able to see the zone that Habitica detects and if it's different than the one the user has selected, there should be an extra message advising the user to check if they really do want to use their manual setting.

The feature would need a new preference setting stored in the user document: 
`preferences.timezoneOffsetManual`
The users would change that setting in the [Settings > Site page](https://habitica.com/#/options/settings/settings).

----

_edit by Alys 2017-09-05:_

It's possible that Firefox's "resist fingerprinting" feature can also prevent Habitica obtaining a user's correct time zone from the browser.

mrbeardy:
### To-Dos
- [x] Settings to allow user to set their timezone.
- [x] Use `moment().utcOffset` instead of `moment().zone`. `zone` is deprecated, and is causing warnings to appear in the console. The major difference between the two functions is that `utcOffset` uses positive values, while `zone` uses negative. For instance, `zone` treats `UTC+1` as `-60`, but `utcOffset` treats it as `60`.
- [ ] Fix failing tests after switching over to use `utcOffset`.
- [ ] Use manual timezone in place of `user.preferences.timezoneOffset`, if `user.preferences.manualTimezoneId` is set to any value other than `-1`.

---
### Possible Refactoring

1. `AVAILABLE_TIMEZONES` might be best stored somewhere else, possibly in a `.json` file.
2. The `findTimezoneById` method in `site.vue` could be moved into a helper file.
3. The `availableTimezones` computed method in `site.vue`, which is returning a list of timezones and prepending the "Automatic" timezone, could be moved into a helper file.

---
#### Automatic timezone (or using a timezone that matches the System)
image

#### Timezone does not match System timezone
image

----
UUID: 20f6d92e-6421-41e0-9602-5b96550eba79


paglias:
@mrbeardy thanks for this PR! I just wanted to let you know that since we're in the finishing phases of the redesign we're not going to merge new features until that's fully launched. So we'll review the PR but it'll have to wait until the new site is live. I hope you understand :)

If you want to help with the new redesign let me know and I can find something for you to work on!

mrbeardy:
Hey @paglias, thanks for letting me know :) I'm happy to wait, I pretty much just created the PR to keep track of things more easily, and since time is such a huge part of the way the site works it definitely won't hurt to get more time to work on it properly :)

> If you want to help with the new redesign let me know and I can find something for you to work on!

Thanks, that'd be great! Is there an issue label for all of the redesign work, or are they listed on Trello?

paglias:
@mrbeardy 

> Thanks, that'd be great! Is there an issue label for all of the redesign work, or are they listed on Trello?

They're on a private google doc right now because we didn't want to clutter too much github but I'll make a list available here for this evening

Alys:
@mrbeardy Changing to moment().utcOffset instead of moment().zone is definitely something we want to do but it needs to be done cautiously so that there's no chance of cron running at inappropriate times. There's some comments about that at https://github.com/HabitRPG/habitica/pull/7053 although they might be outdated in some respects due to recent and upcoming changes to cron.

Because of the potential for problems, we'd want the zone -> utcOffset change to go live by itself, rather than being included in other changes to cron or the time zone code. That would make it easier to do testing and (if necessary) reversion for both your manual time zone change and the utcOffset change. Would you like to move your code for the utcOffset change out of this PR and put it into a new one? If you do, please reference https://github.com/HabitRPG/habitica/pull/7053 in your PR and if you have any responses to the comments in https://github.com/HabitRPG/habitica/pull/7053, you'd be welcome to add them!

mrbeardy:
Thanks for the heads up @Alys. I was debating whether to remove those changes from this PR, but decided to keep them in for the time-being since it became a WIP PR and probably won't go live for a while. 

I'll remove them once I pick this back up, and convert the current timezone offset mappings to work with the existing `.zone()`.

SabreCat:
Whoops, thought this addressed a POST-REDESIGN issue but was mistaken.

SabreCat:
@mrbeardy What's the status on this? We're working our way through the backlog now that redesign bug fixing has slowed down a bit, so we can get this reviewed once it's ready!

librarianmage:
@mrbeardy How are you doing on this?

(@SabreCat)

lemoness:

Alys:
I'm going to close this but I'll make a note of it in the issue so that some of the code can be reused if appropriate.

PR #9038, opened 2017-09-14
Redesign: Subscribe/Buy Gems modal style

SabreCat:

### Changes

* Implements styles from Zeplin for the modal that appears when the user clicks the Gem icon in the header, with two tabs: one for subscription, one for Gem purchasing.
* Adds a new navigation block to the bottom of the user menu to open the subscription side of the above modal.
* Inserts some dividers in the user dropdown to better match organization as laid out in Zeplin.


website/client/components/appMenu.vue
```diff
@@ -353,13 +373,16 @@ export default {
...
+      this.$store.state.gemModalOptions.startingPage = startingPage;
+
       Analytics.track({
         hitType: 'event',
         eventCategory: 'button',
         eventAction: 'click',
         eventLabel: 'Gems > Toolbar',
       });
+
       this.$root.$emit('show::modal', 'buy-gems');
```
paglias: not completely related to this PR but this $emit should be `this.$root.$emit('show::modal', 'buy-gems', {alreadyTracked: true})` or it'll be tracked twice, see https://github.com/HabitRPG/habitica/blob/develop/website/client/app.vue#L220 (I had to add that because not all modals are shown with `this.$root.$emit('show::modal', 'buy-gems');` so I cannot track them like here)

PR #9041, opened 2017-09-15
Show error message in empty market inventory drawer

mrbeardy:
### Changes
- Market: 
  - Hide drawer sliders if number of items for type is empty
  - Move drawer-help-text icon down to line up vertically with text
  - Add TODO for error thrown when trying to sell special item types
- Groups locale: `religous` typo

---

#### before
image

#### after
image

----
UUID: 20f6d92e-6421-41e0-9602-5b96550eba79


paglias:
Regarding market search, I think it's been fixed in another PR. cc @negue https://github.com/HabitRPG/habitica/pull/9033/commits/ecdedb14cd5772832390b21a4ef1feac63ca7481

mrbeardy:
:+1: @paglias, I've merged in the latest `develop`

website/client/components/shops/market/index.vue
```diff
@@ -187,6 +187,7 @@
                 .popover-content-text(v-html="$t('petLikeToEatText')", v-once)
 
         drawer-slider(
+          v-if="ownedItems(selectedDrawerItemType).length > 0"
```
paglias: @negue this looks good to you?

website/client/components/tasks/user.vue
```diff
@@ -135,7 +135,7 @@
       }
   }
 
-  .dropdown-icon-item:hover .svg-icon, .dropdown-item.active .svg-icon {
```
paglias: isn't this going to break the divs using the `dropdown-icon-item` class?

website/common/script/ops/sell.js
```diff
@@ -9,6 +9,7 @@ import {
   BadRequest,
 } from '../libs/errors';
 
+// @TODO: 'special' type throws NotAuthorized error
```
paglias, line 12: isn't this ok since ACCEPTEDTYPES doesn't list it?

website/client/components/shops/market/index.vue
negue: hmm @mrbeardy  could you also add `errorMessage` (like in this [example](https://github.com/HabitRPG/habitica/blob/develop/website/client/components/inventory/stable/index.vue#L165) ) to `drawer` that way the drawer isn't that empty ^^

mrbeardy: Can do :+1:

website/common/script/ops/sell.js
mrbeardy: I was trying to sell a Saddle on local that appeared in the `special` tab and it threw this error. 

Maybe this could be re-worded, or moved into the market view and state that `special` types currently aren't in the `ACCEPTEDTYPES` const and will break.

website/client/components/tasks/user.vue
mrbeardy: It shouldn't in this component since their parent element has the `.dropdown-item` class, but it might break any other places this class was being used. I got confused and thought vue styles were namespaced to their components :grin:

I'll have a look at the other places these classes are used and see how they look.

paglias: @mrbeardy they're scoped indeed (when defined as `<style scoped>` but in this case `.dropdown-item` comes from a child component and since the css here it's scoped it won't be applied. It's only applied to elements that appear in this component, not in children's html

mrbeardy:
#### Market inventory drawer with error message

image

website/client/components/tasks/user.vue
mrbeardy: :thinking: strange, it seems to be allowing me to style the child element when using `.dropdown-item`.

I'll remove this change for now, it's not a huge change and I'd rather not have it accidentally break something somewhere since it's not a game-changing issue.

paglias:
@mrbeardy I finally got the design! If you email me at matteopagliazzi AT gmail DOT com I'll send you the instructions on how to access them

SabreCat:
Thanks again for your help, @mrbeardy! I've noted this toward your next contributor tier!

PR #9042, opened 2017-09-16
New client 2017 09 16 - Fox_town, adjust URLs in links, etc

Alys:
- add Fox_town to mods in Tavern
- change URLs in links from old client format to new client format (e.g., change /#/options/groups/tavern t
+o /groups/tavern)
   - in code
   - in English locales files
   - [ ] non-English locales files should be done in Transifex close to go-live
- change "Username" to "Login Name" on login and registration form (see #8958)
- unrelated to the new client, add some TODO comments to party invitation code because it's buggy


Alys:
Also:
- fixes gem benefits text (previously the first text was repeated four times)
- adjusts the popover for achievements to have the title as a heading and the description under it (previously was all merged on a single line)
- text changes: spells to skills; shield-hand to off-hand; weapon to main-hand

PR #9069, opened 2017-09-24
new client: allow on-hover text to work for all achievements

Alys:
Fixes this item from the testing doc:

**Tester (Chrome OS):   Some achievement hovertexts don't show (streaks, rebirths) [Tier 2a]**

The problem was that the achievements were using their keys as HTML id attributes and some keys were already in use as ids for other elements (`streak` and `rebirth`). This stopped the on-hover effect from triggering.


Alys:
I'm closing this because it was fixed elsewhere in much the same way while this was waiting for review.

TheHollidayInn:
Sorry about that! I didn't see you were waiting.

PR #9075, opened 2017-09-25
sep 25 fixes

negue:
there are multiple popovers that just won't hide, example: 
- the equipment gear drawer, once you unpin an item the popover stays active 
- or if you open the `What does my pet like to eat? 
` popover & hide the drawer in the stable the popover stays active.

there is an update of `bootstrap-vue` but if we upgrade it could break a lot of our code again ...

what do you think - upgrade the package or do you have a better idea / fix for this?

@paglias  @TheHollidayInn

### DIFF
--- a/website/client/components/inventory/equipment/attributesPopover.vue
+++ b/website/client/components/inventory/equipment/attributesPopover.vue
@@ -2,7 +2,7 @@
 div
   h4.popover-content-title {{ itemText }}
   .popover-content-text {{ itemNotes }}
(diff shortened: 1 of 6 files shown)


paglias:
Is the upgrade a major version with breaking changes?

Il 25 Set 2017 22:46, "negue" <notifications@github.com> ha scritto:

negue:
@paglias If I update it to 1.0.0-beta.9 I get this https://github.com/bootstrap-vue/bootstrap-vue/issues/1103 

I guess we'll wait for newer version then ?

paglias:
Yeah

PR #9077, opened 2017-09-25
Add support for `multiDaysCountAsOneDay == false` to evasion

zags:

There is no corresponding issue.  I found this bug while reading the cron code.

### Changes

`if (dailiesDaysMissed > 1) dailiesDaysMissed = 1;` causes the evasion for-loop to only evaluate once even if `multiDaysCountAsOneDay == false`.  This statement isn't necessary because `if (multiDaysCountAsOneDay) break;` will cause the for-loop to evaluate only once if `multiDaysCountAsOneDay == true`.  Removing this statement makes the `dailiesDaysMissed` variable unnecessary.

----
UUID:  5c08b0b0-dfc9-4bdb-b02a-b1d735bab587

### DIFF
--- a/website/server/libs/cron.js
+++ b/website/server/libs/cron.js
@@ -321,8 +321,8 @@ export function cron (options = {}) {
             user.stats.buffs.stealth--;
             EvadeTask++;
           }
-          if (multiDaysCountAsOneDay) break;
         }
+        if (multiDaysCountAsOneDay) break;
       }
 
       if (scheduleMisses > EvadeTask) {
--- a/website/server/libs/cron.js
+++ b/website/server/libs/cron.js
@@ -301,10 +301,6 @@ export function cron (options = {}) {
     let EvadeTask = 0;
     let scheduleMisses = daysMissed;
 
-    // Only check one day back
-    let dailiesDaysMissed = daysMissed;
-    if (dailiesDaysMissed > 1) dailiesDaysMissed = 1;
-
     if (completed) {
       dailyChecked += 1;
       if (!atLeastOneDailyDue) { // only bother checking until the first thing is found
@@ -315,7 +311,7 @@ export function cron (options = {}) {
       // dailys repeat, so need to calculate how many they've missed according to their own schedule
       scheduleMisses = 0;
 
-      for (let i = 0; i < dailiesDaysMissed; i++) {
+      for (let i = 0; i < daysMissed; i++) {
         let thatDay = moment(now).subtract({days: i + 1});
 
         if (shouldDo(thatDay.toDate(), task, user.preferences)) {


SabreCat:
@zags Thanks for submitting this! It looks like your change is causing a test to fail, though. Can you investigate why the user's HP is decreasing more than expected here? https://travis-ci.org/HabitRPG/habitica/jobs/279717583#L1891

zags:
Had to move the break check out of the conditional.  Tests now pass

paglias:
@zags the PR looks good, can you try adding some tests for this functionality?

paglias:
Let me know if you have any question or need help

zags:
The cron function currently has `let multiDaysCountAsOneDay = true;` (https://github.com/zags/habitica/blob/71729850ab318395c0c6ad84033dc6735cfb6390/website/server/libs/cron.js#L273) and doesn't yet have a way to set this variable to false as this is intended in a future set of changes (the comment is, "When site-wide difficulty settings are introduced, this can be a user preference option.").

How should I set this variable to `false` to test this given site-wide difficulty settings haven't been implemented yet?

TheHollidayInn:
Hey @zags. Sorry for the delay here. 

Can you explain a little more about what you are trying to fix. Your title says that you are adding support for setting multiDaysCountAsOneDay to false for stealth, but we haven't implemented that yet. So, I'm not sure how to answer your questions of setting the variable to false.

zags:
I was looking through the code for how the cron works and noticed that there was code for the yet-to-be-implemented user setting multiDaysCountAsOneDay.  This code didn't do anything differently if multiDaysCountAsOneDay was `false`.  This hasn't broken anything yet, since this is not yet available as a user setting, and is currently hardcoded here https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/cron.js#L273

This change adds support for this setting without changing the current behavior of the code.  However, as this setting is not yet implemented, I do not know how to add tests for this.

TheHollidayInn:
Ah, I see. It looks like this was a hack for setting multiDaysCountAsOneDay: https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/cron.js#L306. So maybe two people tried to implement the multi -> one day. 

Moving the break out of the for loop seems odd though. The break is supposed stop the for loop after 1. Any ideas why you needed to move it out?

There definitely is redundant code here :P Thanks for the PR.

zags:
The previous code capped the iterations of the for loop at 1 (if multiDaysCountAsOneDay), so the break statement needs to always run (if multiDaysCountAsOneDay) to preserve this functionality.  At least one test fails if the break statement is conditional.

TheHollidayInn:
I see. Can you tell me which test failed?

zags:
the test "should do damage for only yesterday's dailies" https://github.com/HabitRPG/habitica/blob/bba2e71af37c6fe5738b12f727d52ea22088be63/test/api/v3/unit/libs/cron.test.js#L553 fails if the break statement is in the conditional.

TheHollidayInn:
Sweet. I see what is happening now, thank you! If @paglias agrees, I'm good with this.

paglias:
LGTM 👍

SabreCat:
Thanks for this, @zags, and welcome to the ranks of the Blacksmiths at tier 1!

Note that further tiers require increasing amounts of effort from one to the next, so tier 2 may require multiple PRs or a larger PR to attain. But keep helping out and we'll express our gratitude accordingly!

PR #9085, opened 2017-09-27
fix link to Data Display Tool

Alys:
This fixes the insertion of the user.id into the DDT link, which had been broken.

I've already replaced the data.habitrpg.com domain name wiith a direct link since that domain is outdated anyway. We can put back a habitica domain later if desired.

### DIFF
--- a/website/client/components/settings/api.vue
+++ b/website/client/components/settings/api.vue
@@ -23,7 +23,7 @@
           br
           | {{ $t('chromeChatExtensionDesc') }}
         li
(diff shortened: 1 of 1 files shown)


mandyzhou:
There's also a link to the DDT in the footer (under the Developers section).

Alys:
@mandyzhou Thank you!! I'd missed that completely. It's not as broken as the other one (which was inserting code into the login form) so we'll fix that shortly after go-live.

PR #9098, opened 2017-09-29
Hide the Block button in the user's own profile screen

Alys:
Hide the Block button in the user's own profile screen. This prevents a user from accidentally blocking themself from sending PMs.
User's own profile:
image
Another user's profile:
image


Alys:
ready for review and merge

crookedneighbor (approved):

PR #9100, opened 2017-09-29
Enable browser spell-checking for task titles

q--:
This is disabled by default in Firefox for `<input>` elements.

It can also be enabled document-wide by adding `spellcheck=true` to the `<html>` tag, but I haven't done so because I don't know if there might be elements on the page where spell checking is not warranted.

See also: https://developer.mozilla.org/docs/Web/HTML/Global_attributes/spellcheck


SabreCat:
Looks good! Thanks for submitting this PR. @q--, what's your User ID on Habitica? Let me know and I'll get you contributor credit for this!

PR #9105, opened 2017-09-29
fix allocate() function in userMenu/profile.vue component (fixes #9096)

### LINKED ISSUE #9096: allocated stat points appear to go to Strength
When a user allocates a stat point, on the server it is allocated correctly every time,  but on the client it looks like the point is always allocated to Strength. As soon as you reload you see the correct allocations.

I've tracked it down to this code:
https://github.com/HabitRPG/habitica/blob/4495539e8c34c16b00b3f3c4bb7721428ae72478/website/client/components/userMenu/profile.vue#L281
https://github.com/HabitRPG/habitica/blob/4495539e8c34c16b00b3f3c4bb7721428ae72478/website/client/components/userMenu/profile.vue#L797-L800
https://github.com/HabitRPG/habitica/blob/4495539e8c34c16b00b3f3c4bb7721428ae72478/website/common/script/ops/allocate.js#L12

Note in that third piece of code, `str` is a default for `query.stat`. If you change it so that `int` is the default, then the same change occurs in this bug: all points appear to be assigned to Intelligence. So that default is definitely indirectly responsible for this bug.

I guess that when the allocation happens on the server, everything works perfectly because the user's chosen stat is correctly passed on to the code that assigns the point, and so the default in that third piece of code is never used.
On the client, it looks like the user's chosen stat is not passed, and so the default IS used for the client-side allocation.

I'm not familiar enough with vue to work out how to make the user's choice be correctly passed client-side.

trvrfrd:

Fixes #9096 

### Changes
Fixes the `allocate()` function in `website/client/components/userMenu/profile.vue`. When a user is allocating stat points, this function delegates to the `allocate()` function in `website/common/script/ops/allocate.js` but calls it incorrectly with a String instead of an Object, causing the latter function to default to the (potentially incorrect) Strength stat. As a result, the UI always displays the points being allocated to Strength, as described in the issue.

I didn't see an obvious place to write a new test for this bug, so I didn't write one. Sorry! Happy to add one if someone can show me the right place to do so. For what it's worth, no new test breakage seems to result from this change.

Thanks! Let me know if I can improve this in any way.

\- Trevor

----
UUID: 7620c72a-bb5a-4798-b3c0-846dacb2b44d

### DIFF
--- a/website/client/components/userMenu/profile.vue
+++ b/website/client/components/userMenu/profile.vue
@@ -795,7 +795,7 @@ export default {
       return display;
     },
     allocate (stat) {
-      allocate(this.user, stat);
+      allocate(this.user, {query: { stat }});
       axios.post(`/api/v3/user/allocate?stat=${stat}`);
     },
     allocateNow () {


TheHollidayInn:
Works great! Thanks and congratz on tier 1!

PR #9121, opened 2017-10-01
Oct 1 fixes

TheHollidayInn:


website/client/components/shops/seasonal/index.vue
```diff
@@ -363,7 +364,22 @@
       },
 
       seasonal () {
-        return shops.getSeasonalShop(this.user);
+        let seasonal = shops.getSeasonalShop(this.user);
+
+        let itemsNotOwned = seasonal.featured.items.filter(item => {
```
negue, line 369: I think this could be added inside of ` shops.getSeasonalShop`

TheHollidayInn: Hmm maybe. I added it here because It wanted it to be reactive with Vue. I'm sure there is another way to do it.

negue: ok good point! yeah its easier for the reactive part 👍 , isn't there a good way to have the shared ops act in the same reactive way as everything else?

PR #9147, opened 2017-10-03
Local filters

TheHollidayInn:
## TODO

- [x] Get testing on beta
- [x] Get review from Negue to ensure I'm not doing anything crazy
- [x] Move filter logic to library and share


website/client/components/inventory/equipment/index.vue
```diff
@@ -223,14 +224,41 @@ export default {
...
     }
+
+    this.loadFilters();
   },
   methods: {
+    loadFilters () {
+      let filters = getLocalSetting(this.$route.name);
+      if (!filters) {
+        // @TODO: Should we watch groups? This shouldn't be a side affect in a map like it was before
+        this.groups.forEach((label, group) => {
```
negue: what if we add more groups, then the `viewOptions` would had the old values (before the addition of a group) and probably throw an exception if the user changes the filter for the group

negue: also first try viewing it, just broken with:

```
TypeError: this.groups.forEach is not a function
    at VueComponent.loadFilters (index.vue:250)
    at VueComponent.boundFn [as loadFilters] (vue.runtime.esm.js:183)
    at VueComponent.mounted (index.vue:243)
    at callHook (vue.runtime.esm.js:2661)
    at Object.insert (vue.runtime.esm.js:3522)
    at invokeInsertHook (vue.runtime.esm.js:5465)
    at VueComponent.patch [as __patch__] (vue.runtime.esm.js:5635)
    at VueComponent.Vue._update (vue.runtime.esm.js:2420)
    at VueComponent.updateComponent (vue.runtime.esm.js:2528)
    at Watcher.get (vue.runtime.esm.js:2877)
```

website/client/components/inventory/stable/index.vue
```diff
@@ -602,6 +605,19 @@
 
         this.searchTextThrottled = search;
       }, 250),
+      viewOptions: {
+        handler (newVal) {
+          if (!newVal) return;
+          if (!this.viewOptionsLoaded) return;
+          setLocalSetting(this.$route.name, JSON.stringify(newVal));
```
negue: this won't save the  `Hide Missing` filter

negue: all `hide ...` filters should be moved to `viewOptions` then those would be saved too

negue (changes requested):

negue:
Could you resolve the conflicts?

PR #9148, opened 2017-10-03
Oct 3 fixes

TheHollidayInn:


website/client/components/tasks/column.vue
```diff
@@ -252,8 +263,30 @@ export default {
     },
     inAppRewards () {
       let watchRefresh = this.forceRefresh; // eslint-disable-line
+      let rewards = inAppRewards(this.user);
+
+
+      // Add season rewards if user is affected
+      // @TODO: Add buff coniditional
+      const seasonalSkills = {
```
negue, line 271: those would have to be added inside `inAppRewards` so that the mobile apps get the spells too

TheHollidayInn: Sounds good. I might need to ensure that this is reactive based on buff changes.

TheHollidayInn: The mobile apps need to prepare for this, so for now, we will leave it out.

PR #9178, opened 2017-10-11
[WIP] bootstrap-vue upgrade

negue:
Upgrading to v1.0.0-beta.9:
- include (`Vue.use`) bootstrap-vue instead of importing each component individually 
- change syntax of showing modals `bv::show::modal`

Upgrade already fixed:
- auto hiding popover of the drawer or on un-equipping gear

Current issues:
- [ ] modals are wider

Upgrade bootstrap to `4.0.0-beta.2`, current issues:
- [x] tags selection width (too close)
- [x] Tasks filter button height
- [x] Header Menus popovers are clipped (to the right)
- [x] collapsed header (smaller screens) menu broken 
- [x] button pressed color switched to blue


paglias:
let's upgrade bootstrap too!

negue:
Tried that too but the navigation broke (couldn't test others pages after that) - I'd suggest to upgrade bootstrap in a different PR

paglias:
The reason why I suggested that is that probably the boostrap-vue beta 9 is based on the newest bootstrap version. Here's a list of changes https://github.com/twbs/bootstrap/releases/tag/v4.0.0-beta

PR #9190, opened 2017-10-14
#9184 case insensitive search

### LINKED ISSUE #9184: Search Bar shouldn't be case-sensitive
When you try to search for a task via the task page search bar, it does not work if you use capital letters. For example, if you have a task named "Running" and I enter "Running" in the search bar, it finds nothing, but when you enter "running" the search correctly identifies the task. Search should not be case-sensitive.

geocine:
Fixes #9184 

### Changes
Convert searchText to lowercase using .toLowerCase()

### DIFF
--- a/website/client/components/tasks/user.vue
+++ b/website/client/components/tasks/user.vue
@@ -368,7 +368,7 @@ export default {
   },
   watch: {
     searchText: throttle(function throttleSearch () {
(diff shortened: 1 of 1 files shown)


SabreCat:
Works great! Thanks, @geocine! What's your Habitica User ID? This deserves contributor credit!

PR #9196, opened 2017-10-15
Bulk purchasing

TheHollidayInn:


website/common/script/ops/purchaseWithSpell.js
```diff
@@ -1,9 +1,12 @@
...
+import buy from './buy';
 import purchaseOp from './purchase';
 import get from 'lodash/get';
 
 module.exports = function purchaseWithSpell (user, req = {}, analytics) {
   const type = get(req.params, 'type');
 
-  return type === 'spells' ? buySpecialSpellOp(user, req) : purchaseOp(user, req, analytics);
+  // Set up type for buy function - different than the above type.
+  req.type = 'special';
```
paglias, line 9: I think we should try not to modify the `req` object directly as it's from Express, if we have to change req.params/query/... or add a private object to keep local data like req._habiticaData

TheHollidayInn: This was a hack because the function is an odd bridge. I’d like to move away from passing the request object and pass an params or a specified and documented options so we don’t have to worry about it being from Express

PR #9202, opened 2017-10-16
API Token Changes

paglias:
- [x] Reset API Token
- [x] Hide by default
- [ ] Route tests
- [ ] Finish UI
- [ ] What about social users?
- [ ] Reset on password change
- [ ] Abstract API Token generation (even if it's just uuid.v4())


Alys:
A note for myself for when this goes live: The "notify helpers" label is for me to tell the Socialites that the API Token will change every time a player changes their password, so if a player reports that an extension or other third party tool has suddenly stopped working, they should check to see if their API Token has changed. If it's reported often, we should clarify it further to the users.

paglias:
Going to merge just the part about the token being hidden by default

PR #9207, opened 2017-10-17
Bug Fix on ScoreTask: Gp added instead of append [fixes #9180]

### LINKED ISSUE #9180: Using FIx Character Values to adjust stats temporarily makes the website append not add
Using FIx Character Values to increase your gold by a certain number temporarily makes the website **append** that number to your gold balance, not **add** it. To replicate the bug:

1. Go to User Icon > Settings > Fix Character Values
2. Adjust your gold to have any non-zero value (e.g. 42).
3. Click "Save & Close".
4. Go to your task list and click a positive Habit.
5. Observe your gold balance: the gold you just earned has been **appended** to your gold balance, not **added** to it. E.g., if you earned 1.5 gold, your balance will change from 42 to 421.5  (it should be 43.5)
6. Click a positive Habit again.
7. Your gold balance drops to zero.
8. Reload the website. Your gold balance goes back to what it should be (e.g., about 45).

Examining my values in the database during that test showed that my actual gold balance was always correct in my account; it's just a display issue in the website.

Something similar happens for Experience and presumably other stats.

LulaV14:
Fixes #9180 

### Changes
Added parseInt() to stats.gp in scoreTask, in order for the value to be added instead of appending.

User ID: 996f6501-e620-42ca-9b71-c897bf281c9a

### DIFF
--- a/website/common/script/ops/scoreTask.js
+++ b/website/common/script/ops/scoreTask.js
@@ -130,7 +130,7 @@ function _addPoints (user, task, stats, direction, delta) {
 
     stats.gp += afterStreak;
   } else {
-    stats.gp = parseInt(stats.gp) + gpMod;
+    stats.gp = parseInt(stats.gp, 10) + gpMod;
   }
 }
 
--- a/website/common/script/ops/scoreTask.js
+++ b/website/common/script/ops/scoreTask.js
@@ -130,7 +130,7 @@ function _addPoints (user, task, stats, direction, delta) {
 
     stats.gp += afterStreak;
   } else {
-    stats.gp += gpMod;
+    stats.gp = parseInt(stats.gp) + gpMod;
   }
 }
 
--- a/website/common/script/ops/scoreTask.js
+++ b/website/common/script/ops/scoreTask.js
@@ -130,7 +130,7 @@ function _addPoints (user, task, stats, direction, delta) {
 
     stats.gp += afterStreak;
   } else {
-    stats.gp = parseInt(stats.gp, 10) + gpMod;
+    stats.gp += gpMod;
   }
 }
 
--- a/website/client/components/settings/restoreModal.vue
+++ b/website/client/components/settings/restoreModal.vue
@@ -94,12 +94,12 @@ export default {
       this.user.achievements.streak = clone(this.restoreValues.achievements.streak);
 
       let settings = {
-        'stats.hp': this.restoreValues.stats.hp,
-        'stats.exp': this.restoreValues.stats.exp,
-        'stats.gp': this.restoreValues.stats.gp,
-        'stats.lvl': this.restoreValues.stats.lvl,
-        'stats.mp': this.restoreValues.stats.mp,
-        'achievements.streak': this.restoreValues.achievements.streak,
+        'stats.hp': Number(this.restoreValues.stats.hp),
+        'stats.exp': Number(this.restoreValues.stats.exp),
+        'stats.gp': Number(this.restoreValues.stats.gp),
+        'stats.lvl': Number(this.restoreValues.stats.lvl),
+        'stats.mp': Number(this.restoreValues.stats.mp),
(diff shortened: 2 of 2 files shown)


website/common/script/ops/scoreTask.js
```diff
@@ -130,7 +130,7 @@ function _addPoints (user, task, stats, direction, delta) {
 
     stats.gp += afterStreak;
   } else {
-    stats.gp += gpMod;
+    stats.gp = parseInt(stats.gp) + gpMod;
```
paglias: this change is made into scoreTask.js, if I read correctly the issue is about the Fix Character Values. The code for that modal can be found here https://github.com/HabitRPG/habitica/blob/develop/website/client/components/settings/resetModal.vue

paglias (changes requested):

website/common/script/ops/scoreTask.js
LulaV14: Yes but the core problem is in scoreTask because the real problem is not in the Settings component but in the Task component when you click the positive button in a Habit. Then a method called 'score' is called which calls scoreTask.js.
In scoreTask the user stats object is passed down, which is the one used globally at this point. That's the one being updated and it's the reason why the gp is being updated like that in the user menu.

paglias: I've never heard of problems when scoring an habit and I'm sure we would have got a lot of reports if this was the case. Also the settings component doesn't use scoreTask to adjust the gold points. The problem here is that probably the number is treated as  string so doing numb += numb append the new value to the string

LulaV14: I'm new to this repo, but the changes I made fix the error. I don't  know why you say that if that was the case it will 'cause more problems, I might have to investigate more on that. But the append thing is actually what's happening. In scoreTask.js the method in which I made the change is using a 'stats' object which has 'gp' and the value is a string, that's why the parseInt() fixes the problem.

Here's a gif showing the actual problem.
habitica-error


And here's one with the same steps but applying my changes.
habitica-fix


Can you please guide me on why that's not the fix and what can I do to actually contribute?

paglias:
@LulaV14 the reason why this bug is happening is that the `input` field used to fix your GP values return a string that gets assigned to `user.stats.gp` when you score a task `+=` on a string appends the new value to it instead of adding it like it would do if `user.stats.gp` was a number. The way to fix this is to make sure the value applied to user.stats.gp is a number, for example this way:

```
--- a/website/client/components/settings/restoreModal.vue
+++ b/website/client/components/settings/restoreModal.vue
@@ -96,7 +96,7 @@ export default {
       let settings = {
         'stats.hp': this.restoreValues.stats.hp,
         'stats.exp': this.restoreValues.stats.exp,
-        'stats.gp': this.restoreValues.stats.gp,
+        'stats.gp': Number(this.restoreValues.stats.gp),
         'stats.lvl': this.restoreValues.stats.lvl,
         'stats.mp': this.restoreValues.stats.mp,
         'achievements.streak': this.restoreValues.achievements.streak,
```

The same problem seems to happen with any `number` value in the Fix Character Values modal, so if you want you can try to fix the others in the same way (`stats.hp`, `achievements.streak`, `stats.lvl`, `stats.exp` and `stats.mp`)

The reason why the bug is only temporary is because as soon as a request to the server is made the server will make sure that the field is correctly casted to a number and will transform any string that it was supplied by the browser.

paglias:
Hope this explains what's going on, let me know if you have other questions.

LulaV14:
Hi @paglias I'm doing the requested change but before I did a rebase for all the commits that were added after this PR was created, but they seemed to be added to the whole commit history :S ... so should I close this PR and open another one? or I just go on with all this commits that were added?

EDIT: I have fixed it, so sorry. I will now revert the previous changes and will update the PR.

paglias:
@LulaV14 looks good to me! I'm merging the PR and awarding your contributor tier.

PR #9212, opened 2017-10-17
Replace raw HTML internal hyperlinks with router links #9209

### LINKED ISSUE #9209: Replace raw HTML internal hyperlinks with router links
### Description

From @paglias in our internal bug tracking doc:

> I’ve noticed that sometimes we use `<a href="..."></a>` to link internal pages, we shouldn’t do that as it causes a full reload of the website. We should only use the `router-link` component.

Example of a hardcoded `href` we could get rid of: https://github.com/HabitRPG/habitica/blob/develop/website/client/components/appFooter.vue#L16

Example of a proper `router-link`: https://github.com/HabitRPG/habitica/blob/develop/website/client/components/appFooter.vue#L37

cezarypiatek:
Fixes #9209

### Changes
Hi. I search throught all *.vue files and replaces all "a" with "router-link".

### DIFF
--- a/website/client/components/appFooter.vue
+++ b/website/client/components/appFooter.vue
@@ -13,21 +13,21 @@
         h3 Company
         ul
           li
-            a(href='/static/features') {{ $t('companyAbout') }}
+           router-link(to='/static/features') {{ $t('companyAbout') }}
           li
             a(href='https://habitica.wordpress.com/', target='_blank') {{ $t('companyBlog') }}
           li
             a(href='http://blog.habitrpg.com/', target='_blank') {{ $t('tumblr') }}
           li
-            a(href='/static/faq') {{ $t('FAQ') }}
+           router-link(to='/static/faq') {{ $t('FAQ') }}
           li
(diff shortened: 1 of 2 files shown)


website/client/components/appFooter.vue
```diff
@@ -53,7 +53,7 @@
             h3 Developers
             ul
               li
-                a(href='/apidoc', target='_blank') {{ $t('APIv3') }}
+               router-link(to='/apidoc', target='_blank') {{ $t('APIv3') }}
```
paglias: links with `target="_blank"` shouldn't be changed

paglias (dismissed):

website/client/components/appFooter.vue
cezarypiatek: Fixed in the next commit

SabreCat (approved):

SabreCat:
Thanks much, @cezarypiatek! What's your user ID in Habitica? Let us know and we can get you contributor credit for this!

Issue #9225, opened 2017-10-18
Scheduled To-dos remain at top


lemoness:
When you switch you To-do list from the 'Active' view to the 'Scheduled' view and back again, all To-dos with dates remain at the top of the list in due date order instead of returning to their original (manual) sort order. Refreshing the page puts them back in their proper positions, but they should return to their original sort order immediately.

SimplyComplexable:
I'm looking at this one.

lemoness:
I've updated the labels on this ticket.

SimplyComplexable:
@lemoness There is an active PR for this. Not sure if that means it should have a different label, but just FYI.

paglias:
This seems to be fixed now

Issue #7645, opened 2016-06-11
tasks become unsorted - task list sort order is not preserved (tasksOrder errors)


Alys:
**Sometimes when you sort your To-Dos, some or all of them will become unsorted, either immediately after you have sorted them, or later when you add new To-Dos.**

**The workaround is to post to the [Report a Bug guild](https://habitica.com/#/options/groups/guilds/a29da26b-37de-4a71-b0c6-48e72a900dac) to say that your To-Do sort order is not being saved. An admin will then cleanse some data in your account which will fix the problem. You cannot fix this problem yourself. In particular, deleting completed To-Dos will NOT help (it used to help for a previous version of this bug, but it won't make any difference now).**

It is likely that this bug occurs only during periods of heavy site use, or if you mark several To-Dos as completed very quickly, or if you mark one as completed at the same time as taking other actions on the site very quickly. To minimise the chance of this bug happening (and certain other bugs and sync errors):
- on the website, each time you take an action, wait for the blue sync line at the top of the site to disappear before you take another action
- on the mobile app, each time you mark a task as complete, don't take any other actions until you've seen the notification at the bottom of the screen that tells you how much XP and GP you earned
### Technical Details

---

_EDIT by @paglias_: the best place to fix this is at cron. It can also happen that some tasks ids are missing from user.tasksOrder, this should be fixed too.

---

When you mark a To-Do as completed, Habitica has to modify the To-Do's record and also remove the To-Do's ID from a list that stores the sort order of your incomplete To-Dos (`tasksOrder.todos`). If Habitica is overloaded at the time, it might not do that second action. Thus the sort-order list can contain IDs for completed To-Dos, which interfere when you try to sort your incomplete To-Dos.

This was found when PaperCrystals (52e31d82-e731-4713-81cb-6ef8c5d5d9bc) reported problems with sorting To-Dos in the Report a Bug guild. PaperCrystals's `tasksOrder.todos` array contained the IDs of five To-Dos that had been completed. Their completion dates were all within 64 seconds, indicating that there might have been problems with the user document being updated multiple times in a short period:

```
"dateCompleted" : ISODate("2016-05-28T19:48:28.021Z")
"dateCompleted" : ISODate("2016-05-28T19:48:46.384Z")
"dateCompleted" : ISODate("2016-05-28T19:49:00.962Z")
"dateCompleted" : ISODate("2016-05-28T19:49:23.729Z")
[...]

PR #9238, opened 2017-10-19
Implement Default Sort for Tasks

SimplyComplexable:
Resolves: Issue #9225 

### Changes
Add default sort on to-do tasks to avoid confusion caused by switching
between to-do categories. Perform default sort on Column component mount
so that tasks are correctly sorted on page load.

----
UUID: 0c24cfdd-41ad-4a2b-aca0-f247e01f93cf

### DIFF
--- a/website/client/components/tasks/column.vue
+++ b/website/client/components/tasks/column.vue
@@ -11,7 +11,7 @@
       ) {{ $t(filter.label) }}
   .tasks-list(ref="tasksWrapper")
     input.quick-add(
-      v-if="isUser", :placeholder="quickAddPlaceholder", 
+      v-if="isUser", :placeholder="quickAddPlaceholder",
       v-model="quickAddText", @keyup.enter="quickAdd",
       ref="quickAdd",
     )
@@ -243,7 +243,7 @@ export default {
       todo: {
         label: 'todos',
         filters: [
-          {label: 'remaining', filter: t => !t.completed, default: true}, // active
+          {label: 'remaining', filter: t => !t.completed, default: true, sort: t => t.createdAt}, // active
           {label: 'scheduled', filter: t => !t.completed && t.date, sort: t => t.date},
           {label: 'complete2', filter: t => t.completed},
         ],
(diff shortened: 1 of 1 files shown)


paglias:
Shouldn't we sort based on user.tasksOrder.todos that's the order the user chose?

TheHollidayInn:
Yea, I think the confusion here is that one of your filters automatically sorts your todos and doesn't base off of the user.taskOrder. IMO, we just have a bad UX here. Not exactly sure how to fix it.

SimplyComplexable:
@paglias @TheHollidayInn I'm pretty new to habitica. I thought that might be an issue, but I couldn't see a way to specify a default order. How do I go about choosing that?

TheHollidayInn:
Yea! I think your intuition is correct. We just need to decide on a UX solution, first. I've asked in our internal Slack. I'll post here after we discuss. Thanks for the PR :D

paglias:
If we only want to fix the issue where tasks don't go back to the original position I think, but I haven't checked, that if you call the tasks:order action they should go back to the correct sorting order

SimplyComplexable:
Is there any update on the desired behavior here? 
Also just so I'm certain I understand the concern here: The debate is that currently selecting the scheduled filter applies a sort which overrides the user's preferred sort, and the solution provided is also overriding the user's preferred sort. Is that correct?

SabreCat:
@SimplyComplexable That's correct! Users are very particular about getting their tasks in the order they want them (thus #7645 being such a hotbed of activity right now), so we want to use the user-controlled `tasksOrder` whenever possible. Sorting by date created could be useful, but it'd be an additional option, not the default.

SimplyComplexable:
@SabreCat I noticed after I made the comment that @TheHollidayInn had done work related to this issue. Not sure if it's completely resolved. I'm happy to keep working on this, but I don't want to double work, so just to confirm; is there more work that needs to be done or should the issue be resolved?

SabreCat:
If I'm understanding the different issues correctly, you should be able to proceed with this fix independent of @TheHollidayInn's work. The causes and symptoms are somewhat different between issues #7645 and #9225.

paglias:
@SimplyComplexable I just checked and with the most recent code the issue seems to have been solved so I'm closing the PR. Thanks for your help :)

PR #9239, opened 2017-10-20
Force refresh is server version is updated

TheHollidayInn:


website/client/app.vue
```diff
@@ -213,6 +213,15 @@ export default {
         }
       }
 
+      // Verify the client is updated
+      const serverAppVersion = response.data.appVersion;
+      let serverAppVersionState = this.$store.state.serverAppVersion;
+      if (isApiCall && !serverAppVersionState) {
+        this.$store.state.serverAppVersion = serverAppVersion;
+      } else if (isApiCall && serverAppVersionState !== serverAppVersion) {
+        location.reload();
```
paglias: I think we should open a modal or something here, so that we don't refresh in the middle of someone doing things which would be annoying. Also let's use `location.reload(true);`, otherwise i've seen browsers reload a cached version of the site, with `true` it makes sure no cache is used

paglias (changes requested):

Issue #9244, opened 2017-10-21
API must enforce maximum length for summaries for Guilds and Challenges


Alys:
Guilds and Challenges have a summary field and the website enforces a maximum length for it:
https://github.com/HabitRPG/habitica/blob/6838b7d0a60c1b5b089ae5dab17f9cf10e09b937/website/common/script/constants.js#L9-L10

The API's routes for creating and editing Guilds and Challenges should enforce the same maximum.

Note that currently the API does not make the summary field mandatory - that is desired behaviour at the moment because a mandatory summary would be a breaking change for the API. Because of this, the API should enforce the maximum length only if a summary is provided. Don't make the summary mandatory.

There should also be tests that ensure that IF a summary is provided, it is not too long.

Alys:
There should also be tests that ensure that IF a summary is provided, it is not too long.

jyotman:
Hi, may I take this up?

From what I understand, there needs to be a validation check while creating/updating (APIs) challenges or guilds using *express-validator*. I noticed that the validation is already present at the mongoose schema level.

Also should the check be on all *groups* or only on *groups* of type *guild*?

jyotman:
@Alys ☝

lemoness:
@jyotman Sorry for the delay! We'd love to have you work on this ticket. I'll leave your more technical questions for one of our devs. 

I've updated the labels on this issue.

MathWhiz:
@jyotman How are you doing with this? Do you need any assistance with anything?

jyotman:
@MathWhiz yes, can you can answer the doubt I posted above?

MathWhiz:
Sadly, I am unable to. @Alys @Sabrecat What do you think?

TheHollidayInn:
Sorry for the delay! The validation for mongoose and express are different. So, I think the best start here is to validate length during the api request. You are welcome to add validation to the model as well.

This should be added to guilds and challenges. I’m not sure about parties (the other type of group from guilds). @alys what do you think about parties?

Alys:
Yes, parties too. They can be treated identically to guilds for this issue.

lemoness:
@jyotman Hi there! How is progress going on this fix? Let us know if you have any questions.

jyotman:
@lemoness sorry I forgot about this. I will work on this on the weekend. If its urgent then feel free to assign it someone else.

lemoness:
[...]

PR #9246, opened 2017-10-21
Restrict Challenge Summary String in API

calasada:
Uses substring to cleave off the end of a challenge summary if it's over the constant defining the size limit for the summary.

https://github.com/HabitRPG/habitica/issues/9244

### Changes

Just *hopefully* adds a string length limit for the summary to a challenge on the API level, which shouldn't happen anyway because the website restricts the length.

----
UUID: 4a58daa4-52de-49ff-b768-38bfbc84a563

### DIFF
--- a/website/server/controllers/api-v3/challenges.js
+++ b/website/server/controllers/api-v3/challenges.js
@@ -223,6 +223,8 @@ api.createChallenge = {
 
     if (!req.body.summary) {
       req.body.summary = req.body.name;
+    } else {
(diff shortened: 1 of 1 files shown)


Alys:
I think it would be better to return an error message and not change the summary, rather than use a truncated version. The person editing the summary might not notice the truncation.

Even though the website will give good error handling, the API needs to do something similar because it's used by the mobile apps and third-party tools.

paglias:
Yes, I agree with @Alys, this should be an error at the schema level in models/group.js

librarianmage:
@Calasada How are you doing with this?

lemoness:

PR #9251, opened 2017-10-22
Modify relative dueIn time on tasks to be in days

### LINKED ISSUE #9224: Relative due dates should be in days, not hours
On the tasks page, we now use relative due dates. However, they are erroneously displaying in hours instead of displaying in days. Because there is no way to select hours, instead of saying something like "due in 5 hours" it should simply say "due today" (or "due tomorrow", depending on the time of day).

aquibm:

Fixes #9224 

### Changes
Normalizes the current time and task due time to the ends of their respective days. Returns 'today' if the dates are the same day else uses moment's humanize function to allow for weeks, months, years and so on.

@Alys at present this will render values like `Due in a day` rather than `Due tomorrow` as we don't have `tomorrow` or `yesterday` as localized strings. We could change the behaviour fairly easily to allow for this if required.

Example:
image

----
UUID: c342ccd8-f5e3-4ad6-b366-e343db63a01a


mandyzhou:
Should "Due in a day" be grey instead of red, since it's not due yet? On the current website, To-Dos that are actually due in a day show up as either red or grey, depending on how many hours show up (e.g., "Due in 6 hours" might be in red, but "Due in 18 hours" might be in grey).

aquibm:
Updated based on @mandyzhou's suggestion, here's the same list now

image

website/client/components/tasks/task.vue
```diff
@@ -417,12 +417,21 @@ export default {
...
     },
+    timeTillDue () {
+      const endOfToday = moment().endOf('day');
+      const endOfDueDate = moment(this.task.date).endOf('day');
+
+      return moment.duration(endOfDueDate.diff(endOfToday));
+    },
     isDueOverdue () {
-      return moment().diff(this.task.date, 'days') >= 0;
+      return this.timeTillDue.asDays() <= 0;
```
crookedneighbor, line 605: I don't see how this works. You've defined a `timeTillDue` function, but you're not calling it as a function.

Shouldn't it be:

```js
return this.timeTillDue().asDays() <= 0;
```

crookedneighbor (dismissed):
Is there no test file for this component? This behavior should really have some coverage.

website/client/components/tasks/task.vue
paglias: @crookedneighbor this is a Vue computed property so it's accessed as a normal property not a function

paglias: but @aquibm i don't think moment plays very well with Vue reactivity and since you're always using `timeTillDue.asDays()` I think you should rename `timeTillDue` to `daysTillDue` and just return a number from it calling `asDays` on the moment duration

aquibm: @paglias I need a `moment.duration()` in order to use moment's humanize function. If we change `timeTillDue` to days, we're going to have to convert it back into a duration on line `432` before being able to humanize it.

aquibm:
@crookedneighbor I couldn't find any tests for this or most of the other Vue components. We could pull this behaviour out into a helper somewhere to make it more testable though.

paglias:
We don't have test coverage for the client yet

website/client/components/tasks/task.vue
paglias: Ah I see, I missed that line, in that case continue this way 👍

paglias:
@aquibm this looks good, can you resolve the conflicts so we can merge it?

aquibm:
@paglias Fixed 👍

paglias (approved):

paglias:
@aquibm thanks for this! I'm merging the PR and awarding your contributor tier.

Issue #8813, opened 2017-06-11
certain kinds of swear words / oaths should be blocked in guilds


Alys:
~This is on hold until the PR for https://github.com/HabitRPG/habitica/issues/8062 is completed (automatic muting of slurs), since there'll probably be code in that PR that can be repurposed for this issue.~ EDIT: We're now considering which words to block where.

In https://github.com/HabitRPG/habitica/pull/8197 , we prevented swear words, slurs, etc being posted to the Tavern and that's been working very well. We want it extended to public guilds, although with a more limited set of words. For example, in the Tavern, it's not permitted to mention legal addictive substances such as certain drinks, however those words would not be blocked in public guilds. 

We'll give more details about what should be blocked elsewhere after https://github.com/HabitRPG/habitica/issues/8062  is fixed. For now, this is just a placeholder issue so that we don't forget about this.

abatula:
I can work on this if you have more details on what you need done. As originally written, my fix (#8177) to #8062 should also block those words in guilds and private groups. Is this a case of more extensive tests needed (e.g. test both tavern chat and a regular guild to make sure the words are blocked)?

Alys:
@abatula Thank you for offering! However we'll leave this on hold for a little longer. As you might have seen from a [Bailey message](https://habitica.com/static/old-news), a new design for the website will be released soon. One aspect of it will be tags for guilds, to categorise them by topic. @shanaqui had the idea to use those tags to control swearword blocking in guilds. For example, most guilds would not be allowed to use religious oaths, however a guild tagged as being about religion would be allowed to use them because it would be more likely that the words were not being used as oaths.

So after the new site is released, moderators will group the current swearwords into sections associated with guild tags and then we'll take this issue off hold. If you'd like to work on it then, you'd be welcome to! If we remember, we'll post here to ask you about it then.

Alys:
Labels updated.

@abatula I've marked this as in progress in case you still want to work on it, but if you don't no worries at all - just let us know and we'll put it back to help wanted.

We're actually doing a quick fix to introduce swearword blocking to most guilds, requested by Lemoness: https://github.com/HabitRPG/habitica/pull/9253 however that's intended to be just a short term solution until a better fix is done for this issue. We still like shanaqui's idea from my previous comment here. What I'm thinking is we'll have a **new** file to replace `website/server/libs/bannedWords.js` and it will have different arrays in it, something like this:

```
let bannedWordsReligious = [
[...]

PR #9253, opened 2017-10-22
allow banned word / swearword blocker to apply to most public guilds

Alys:
(This PR is **not** related to our list of slurs, which are still prohibited in all public and private areas of Habitica.)

Currently, our `bannedWords` list (swear words, religious oaths, adult topics, etc) applies only to the Tavern (#8197).

This PR applies it to all public guilds, except for some hard-coded exceptions. Hard-coding is normally not desirable but this is for a quick, short-term solution to our need for swearword blocking in guilds. Later, it will be replaced with a better solution that allows each public guild to have an automatically customised list of banned words based on the guild's tags.

The public guilds that the swearword blocker does not apply to include:
- religious/atheist/philosophy guilds (legitimate use of words that would normally be banned as oaths);
- food/drink/lifestyle/perfume guilds (alcohol can be mentioned in appropriate ways);
- guilds dealing with traumatic life events (people must be allowed to describe them);
- foreign language guilds for languages using the Roman alphabet (to avoid accidental banning of non-English words).

The code in this PR is complete and ready to be reviewed but the PR should not be merged until other moderators have reviewed the list of guild exceptions. I'll post here when that's been done. EDIT: That's done but see my comment below about making the error message more readable - that should be done before merging this PR.

### DIFF
--- a/website/server/libs/guildsAllowingBannedWords.js
+++ b/website/server/libs/guildsAllowingBannedWords.js
@@ -33,6 +33,7 @@ let guildsAllowingBannedWords = {
   'df71c7e9-acbd-4333-81a7-699b34c7ef56': true, // Challenge Translation Guild
   '0f2c28ff-35db-47eb-a459-5466dcd7d958': true, // Chef's Special
   '3f88cb57-e491-471e-90a3-b5a6cdb6a3d4': true, // Classical Scholars
+  'f8eb26cd-42ab-4485-bfa1-0bc59c10609e': true, // Cocktail Crusaders
   '15d9aeca-50d9-4564-b955-41333ce60d68': true, // Coffee Enthusiasts
   '22e83aae-6004-4513-b3c9-66e5ff738f66': true, // Coisas Boas
   '7732f64c-33ee-4cce-873c-fc28f147a6f7': true, // Commonwealth of i18n
--- a/website/server/libs/guildsAllowingBannedWords.js
+++ b/website/server/libs/guildsAllowingBannedWords.js
@@ -1,0 +1,134 @@
+// Specify guilds that the automatic swearword blocker does not apply to:
+// - religious/atheist/philosophy guilds (legit use of words banned as oaths);
+// - food/drink/lifestyle/perfume guilds (alcohol allowed);
+// - guilds dealing with traumatic life events (must be allowed to describe them);
+// - foreign language guilds using the Roman alphabet (avoid accidental banning of non-English words).
+//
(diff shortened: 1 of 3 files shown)


Alys:
The other mods have reviewed the list of exceptions and are happy with it, except for some adjustments that I've made now. We can make more later too if needed - no need to keep this PR waiting for that.

However we should fix this problem before this PR is merged (copied from internal admin chat):

When you try to post a swear word in the Tavern, the error message disappears too quickly to be read. You can test this by pasting TEST_PLACEHOLDER_SWEAR_WORD_HERE into the Tavern. The same is also true of any other error message, I assume. Can we find some other way to display error messages, or customise the notification for errors so that it's visible for longer?

Having the word blocker on most guilds will be a surprising change so it's important that the error message is readable.

website/server/libs/guildsAllowingBannedWords.js
```diff
@@ -0,0 +1,139 @@
+// Specify guilds that the automatic swearword blocker does not apply to:
+// - religious/atheist/philosophy guilds (legit use of words banned as oaths);
+// - food/drink/lifestyle/perfume guilds (alcohol allowed);
+// - guilds dealing with traumatic life events (must be allowed to describe them);
+// - foreign language guilds using the Roman alphabet (avoid accidental banning of non-English words).
+//
+// This is for a short-term, partial solution to the need for swearword blocking in guilds.
+// Later, it will be replaced with customised lists of disallowed words based on the guilds' tags.
+
+let guildsAllowingBannedWords = {
```
paglias, line 10: I'm a bit concerned about this list, can we instead add an option to groups to allow banned words? It shouldn't take much and I can write the code an enable the option for the IDs below

Alys: This is a short-term fix for rapid control of swear word blocking, until a proper solution is written and there's already a contributor interested in that issue. `guildsAllowingBannedWords` won't be in the code base permanently. I won't have time this week to make code changes and we'd really like to get this live soon. If you want to make all the changes that you're suggesting then I have no objections at all but I'm really not sure it's worth the effort. What concerns you about the list?

paglias: that we'll forget about it but okay as otherwise there's no time for today's deploy to make the necessary changes :)

Alys: Forgetting is a valid concern! I've added a note about deleting the temporary code here: https://github.com/HabitRPG/habitica/issues/8813#issuecomment-345061987

PR #9255, opened 2017-10-23
Task order fixes

TheHollidayInn:


website/client/components/tasks/column.vue
```diff
@@ -394,6 +400,14 @@ export default {
       }
       this.activeFilters[type] = filter;
 
+      if (['scheduled', 'due'].indexOf(filter.label) === -1) {
+        let sortedTasks = this.$store.dispatch('tasks:order', [
+          this.tasks[`${type}s`],
+          this.user.tasksOrder,
+        ]);
+        this.tasks[`${type}s`] = sortedTasks[`${type}s`];
```
paglias, line 408: I'm a little concerned about this line: since we're changing the array I'm not sure the data will propagate correctly to the store.

```
let a = {b: [1, 2 ,3]}
let ref = a.b;
ref = [4, 5, 6]
```

`ref` and `a.b` are now different.

Ideally this line shouldn't be necessary or if it's it should be into the store action so that we can update the store state directly and keep the reference

website/client/components/tasks/column.vue
```diff
@@ -394,6 +400,14 @@ export default {
       }
       this.activeFilters[type] = filter;
 
+      if (['scheduled', 'due'].indexOf(filter.label) === -1) {
```
Alys, line 403: Does this stop the Dailies sort order being changed by the user when they're using the "Due" view? If so, that's not the ideal behaviour.

TheHollidayInn: Nope. This just prevents an ordering refresh when they click "due"

PR #9259, opened 2017-10-24
Add ESLint to gulp scripts

### LINKED ISSUE #8022: Lint /gulp and /webpack
- [x] /webpack
- [ ] /gulp

lachieh:
Fixes #8022 

### Changes
Add ESLint to `gulpfile.js` and `gulp/*.js` files.


lachieh:
This isn't quite complete. There are a few issues remaining that I wanted to check in on before making these changes.

* `gulp/gulp-build.js` - Uses `process.env.` while the default rules define `no-process-env`
* `gulp/gulp-console.js` - Uses non-global require statements (`global-require`)
* `gulp/gulp-tests.js` - Expects errors to be handled (`handle-callback-error`)

These are all things that could be ignored locally in the `gulp/.eslintrc` file, but wanted to get some input from the team.

paglias:
@lachieh this looks great, thanks! The only thing is that I'd like to avoid having exceptions defined into a `.eslintrc` file, instead I'd prefer if the `gulp` folder used the same `eslintrc` files as the other folders. When you need to add an exception you can disable the rule for the specific line only like this https://github.com/HabitRPG/habitica/blob/51ffe2c8c2cfc0ae6437bcfc151010759ee365f8/website/common/index.js#L5

gulp/.eslintrc
```diff
@@ -6,8 +6,5 @@
   "extends": [
     "habitrpg/server",
     "habitrpg/esnext",
-  ],
```
paglias: I think this file can be completely removed as it's exactly the same we have in the root folder of the project, so it's unnecessary

paglias (changes requested):

gulp/.eslintrc
lachieh: yep, not a problem.

lachieh:
I've gone and added the exceptions for all the rules that couldn't be fixed through simple es6 features (arrow funcs, let/const, etc). There are a few that should probably be refactored, but this PR now successfully builds and fixes the original issue.

paglias:
@lachieh thanks for the work on this PR! If you provide me your Habitica UUID I'll be happy to award your contributor tier for this :)

lachieh:
f1d51921-a042-450a-8279-147e086f77fb :D

paglias:
Your contributor tier has been awarded.

PR #9268, opened 2017-10-25
Validate that everyX values in dailies are integers bounded by 0 and 9999 fixes #8782

### LINKED ISSUE #8782: everyX values for Dailies should not be allowed to be non-integers
The <strike>website</strike> API accepts non-integer values for `everyX` for a Daily, such as 0.5. This causes errors such as the one below.

Suggested fixes:
- The API and/or task schema must prohibit anything other than an integer (including 0, which is currently allowed and should remain allowed until/if we implement a different way of making a Daily never due). It should also ensure that there IS a value for `everyX` - `null` is not permitted.
- _Already fixed:_ <strike>The website should not allow you to type anything other than digits (i.e., the `.` character and all other non-digit characters should not be able to be entered). I recommend still allowing people to type numbers rather than select from a drop-down because drop-downs are annoying for anything more than a few numbers.</strike>

The Android app doesn't have this bug because you can't manually type the everyX value. I suspect but do not know that the iOS is the same and therefore also wouldn't need a fix, but someone should check that to be thorough.

Example of a task affected by this bug (I've fixed this task and I'm looking for all others affected to fix them too): 
``` 
{
    "_id": "f489a688-01e7-43ae-b11e-fe8df7f6ee35",
   [snip irrelevant data]
    "userId": "b14c2d02-563a-4ba4-aa9d-efd20c183eb9",
    "updatedAt": "2017-05-24T07:25:31.105Z",
    "completed": false,
    "weeksOfMonth": [],
    "daysOfMonth": [
        1
    ],
    "streak": 0,
    "repeat": { "su": true, "s": true, "f": true, "th": true, "w": true, "t": true, "m": true },
    "startDate": "2017-05-01T04:00:00.000Z",
    "everyX": 0.5,
    "frequency": "daily",
    "isDue": true
}
```

Resulting error:

```

2017-06-01T04:06:54.695Z - error: Error: Intervals must be greater than zero
[snip]
170601_212551/i-00fcee367bbd7b16a/var/log/nodejs/rotated/nodejs.log-1496307661:    at processImmediate [as _immediateCallback] (timers.js:582:5) method=POST, originalUrl=/api/v3/tasks/236a33c9-0f04-4888-9ba1-197c4bdf1fb6/score/up?userV=367, host=habitica.com, content-length=2, accept=application/json, text/plain, */*, accept-encoding=gzip, deflate, br, accept-language=en-US,en;q=0.8, content-type=application/json;charset=UTF-8, dnt=1, origin=https://habitica.com, referer=https://habitica.com/, x-api-user=b14c2d02-563a-4ba4-aa9d-efd20c183eb9, x-client=habitica-web, x-forwarded-port=443, x-forwarded-proto=https, , httpCode=500, isHandledError=false, 
```

TNychka:
https://github.com/HabitRPG/habitica/issues/8782

### Changes
Added validation to daily schema, prevents non integer values from being used for everyX.

Note: also added bounding to everyX values, now all values must be 0 to 9999 (inclusive). This was not requested in the issue, but looking at cron.js:99 all dailies with values above 9999 or below 1 are treated as 0, thus the bounding makes sense in this case. In addition added the client side check in editing a task to make sure values are bounded by 9999 as well.

**(Manual) Test cases:**
```
"banana" 
834758923475890234752038945723409857234980572394805728930475390284671827385
-8347589234758902347520389457234098572349805723948057289304753902846718273
10000 
-1
2.5
2.333333
```
all return 400 response code with the following
```
"errors": [{
            "message": "Valid everyX values are integers from 0 to 9999",
            "path": "everyX",
            "value": <SENT-VALUE>
} ]

```
These test cases
```
0
1
2
9998
9999
```
all return 200 response code

@alys would like to remind any admins that before this PR is merged to remove existing non-int values. Not sure how to link to a git comment but it's in the comments on the issue.
"Don't worry about existing non-int values in the database - admins will remove those before your PR is merged. If you could remind us of that in your PR, we'd be grateful. :)"

----

### DIFF
--- a/migrations/tasks/tasks-set-everyX.js
+++ b/migrations/tasks/tasks-set-everyX.js
@@ -56,7 +56,7 @@ function updateTasks (tasks) {
 
   return Promise.all(taskPromises)
   .then(function () {
-    processtasks(lasttask._id);
+    processTasks(lasttask._id);
   });
 }
 
--- a/migrations/tasks/tasks-set-everyX.js
+++ b/migrations/tasks/tasks-set-everyX.js
@@ -1,0 +1,88 @@
+var migrationName = 'tasks-set-everyX';
+var authorName = ''; // in case script author needs to know when their ...
+var authorUuid = ''; //... own data is done
+
+/*
+ * Iterates over all tasks and sets invalid everyX values (less than 0 or more than 9999 or not an int) field to 0
+ */
+
+var monk = require('monk');
+var connectionString = 'mongodb://localhost:27017/habitrpg?auto_reconnect=true'; // FOR TEST DATABASE
+var dbTasks = monk(connectionString).get('tasks', { castIds: false });
+
+function processTasks(lastId) {
+  // specify a query to limit the affected tasks (empty for all tasks):
+  var query = {
+    type: "daily",
+    everyX: {
+      $not: {
+        $gte: 0,
+        $lte: 9999,
+        $type: "int",
+      }
+    },
+  };
+
+  if (lastId) {
+    query._id = {
+      $gt: lastId
+    }
+  }
+
+  dbTasks.find(query, {
+    sort: {_id: 1},
+    limit: 250,
+    fields: [],
+  })
+  .then(updateTasks)
+  .catch(function (err) {
+    console.log(err);
+    return exiting(1, 'ERROR! ' + err);
+  });
+}
+
+var progressCount = 1000;
+var count = 0;
+
+function updateTasks (tasks) {
+  if (!tasks || tasks.length === 0) {
+    console.warn('All appropriate tasks found and modified.');
+    displayData();
+    return;
+  }
+
+  var taskPromises = tasks.map(updatetask);
+  var lasttask = tasks[tasks.length - 1];
+
+  return Promise.all(taskPromises)
+  .then(function () {
+    processtasks(lasttask._id);
+  });
+}
+
+function updatetask (task) {
+  count++;
+  var set = {'everyX': 0};
+
+  dbTasks.update({_id: task._id}, {$set:set});
+
+  if (count % progressCount == 0) console.warn(count + ' ' + task._id);
+  if (task._id == authorUuid) console.warn(authorName + ' processed');
+}
+
+function displayData() {
+  console.warn('\n' + count + ' tasks processed\n');
+  return exiting(0);
+}
+
+function exiting(code, msg) {
+  code = code || 0; // 0 = success
+  if (code && !msg) { msg = 'ERROR!'; }
+  if (msg) {
+    if (code) { console.error(msg); }
+    else      { console.log(  msg); }
+  }
+  process.exit(code);
+}
+
+module.exports = processTasks;
--- a/test/api/v3/integration/tasks/POST-tasks_user.test.js
+++ b/test/api/v3/integration/tasks/POST-tasks_user.test.js
@@ -628,6 +628,43 @@ describe('POST /tasks/user', () => {
       });
     });
 
+    it('returns an error if everyX is a non int', async () => {
+      await expect(user.post('/tasks/user', {
+        text: 'test daily',
+        type: 'daily',
+        everyX: 2.5,
+      })).to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: 'daily validation failed',
+      });
+    });
+
+    it('returns an error if everyX is negative', async () => {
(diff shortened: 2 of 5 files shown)


TNychka:
I'm going to add some test cases to sanitation before this is merged.

TNychka:
Added test cases, should be good for review now. PTAL

TNychka:
Not sure why the travis build is failing and the coverage is failing because the travis build is failing...

test/api/v3/unit/models/task.test.js
```diff
@@ -151,6 +151,169 @@ describe('Task Model', () => {
       });
     });
 
+    describe('sanitizeTasks', () => {
```
paglias: what's the reasons for this new tests? they seem to validate that valid data does not error so it doesn't say anything about whether validation works (and I think we already have tests for those behaviors)... Can you instead add tests for this new validation?

paglias (changes requested):

test/api/v3/unit/models/task.test.js
TNychka: @paglias Sorry, I don't think I understand correctly. Many of these tests didn't exist before so I added them to increase coverage. For todo/reward/habit they only check valid inputs, only because they aren't directly related to this ticket, I can add more tests for them if you would like. For dailies which is what this issue relates to I have tests against negative ints, non ints, and valid inputs. One I do see I'm missing is large ints which I can add. But what kind of test are you thinking of?

paglias: > For dailies which is what this issue relates to I have tests against negative ints, non ints, and valid inputs. One I do see I'm missing is large ints which I can add. But what kind of test are you thinking of?

Sorry I didn't notice you already added them , they're enough but could you move them to the integration folder since most validation tests are there (under `tasks/POST_tasks_user.test.js`)?

>  Many of these tests didn't exist before so I added them to increase coverage. For todo/reward/habit they only check valid inputs, only because they aren't directly related to this ticket,

What I don't think is quite right is that in these new tests ("sanitize x") you're checking that validation doesn't fail if the data is correct which is not what we should test for two reasons IMHO:

- there should be a clear thing that is tested like "fails if input is X" not not fail if input is different from X 
- these tests are difficult to maintain as we should pass new data to the task creation every time a new field is added to the task model

TNychka: Alright sure, I'll update these. Thanks!

TNychka:
Moved and reorganized tests. PTAL

paglias (approved):

paglias:
@Alys shall we make a migration before merging this?

Alys:
Yes, we should change all invalid values to 0.

paglias:
@TNychka do you think you'll be able to write a migration (see the /migrations folder) for users who have invalid values? If it's a problem, no worries, someone from the team can do it

TNychka:
I'll give it a go

TNychka:
Added a migration, PTAL. I wasn't able to get it (or any other migrations) running locally due to node errors, specifically `require is not defined`.

paglias:
@SabreCat this PR has a migration that will have to be run before it's deployed

Alys:
Ideally also after it's deployed just in case any new tasks have bad data
in them (unlikely but being sure would be good)

SabreCat:
Did a trivial fix to the migration, looks good now. Thanks @TNychka, this will go into next week's merge batch! I've noted it toward your Blacksmith tier 3!

TNychka:
Thanks!

Alys:
This seems to have been pushed to production, although I can't find mention of it in our [Releases](https://github.com/HabitRPG/habitica/releases). 

A player has just reported cron failing for two days in a row and it seems to be from one of their Dailies having an incorrect everyX value (I've fixed it), so I'm not sure the migration has been run. I'm going to look into this more later after getting through the Bug guild.

Until that's resolved, if any other admins see reports of cron failing, check for Dailies for non-integer everyX values.

Alys:
I'm pretty sure all Dailies now have valid values.  For ones that had values that were invalid but otherwise sensible (e.g.., 2.5), I set them to have matching valid values (e.g., 2). For everything else (e.g., 1e+189 or null) I set them to zero. It didn't take long because most were clearly invalid and not sensible.

However it would be good to run the migration script just in case I missed anything, if it wasn't run previously. If it was run and failed, we should work out why for future reference.

SabreCat:
Odd, I don't know what happened there. I tried rerunning the migration after today's deploy, but it's timing out when trying to connect to the database. :\

Issue #9285, opened 2017-10-27
Clone Tasks Button Label Change


lemoness:
When you're cloning a Challenge,  the button at the bottom of the popup should say “Clone Tasks” instead of "Add Tasks."

EmilyOng:
I would like to take this up.

PR #9300, opened 2017-10-27
changed to 'createChallengeCloneTasks'

EmilyOng:

#9285

### Changes
`button.btn.btn-primary(v-once, v-if='creating', @click='createChallenge()') {{$t('createChallengeAddTasks')}}`
//instead of 'createChallengeAddTasks', i have changed it to 'createChallengeCloneTasks'

### DIFF
--- a/website/common/locales/en/challenge.json
+++ b/website/common/locales/en/challenge.json
@@ -30,6 +30,7 @@
   "notParticipating": "Not Participating",
   "either": "Either",
   "createChallenge": "Create Challenge",
+  "createChallengeAddTasks": "Add Challenge Tasks",
   "createChallengeCloneTasks": "Clone Challenge Tasks",
   "addTaskToChallenge": "Add Task",
   "discard": "Discard",
--- a/website/common/locales/en/challenge.json
+++ b/website/common/locales/en/challenge.json
@@ -30,7 +30,7 @@
   "notParticipating": "Not Participating",
   "either": "Either",
   "createChallenge": "Create Challenge",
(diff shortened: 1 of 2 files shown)


Alys:
@EmilyOng I think you might have forgotten to commit one file. :) The `createChallengeCloneTasks` string doesn't currently exist.

EmilyOng:
Is it something to do with the [i18n translation](https://github.com/EmilyOng/habitica/blob/issue9285/website/client/libs/i18n.js)? How could i then define the string as an i18n.string? Thanks :)

EmilyOng:
oh its ok i think i have figured it out

website/common/locales/en/challenge.json
```diff
@@ -30,7 +30,7 @@
   "notParticipating": "Not Participating",
   "either": "Either",
   "createChallenge": "Create Challenge",
-  "createChallengeAddTasks": "Add Challenge Tasks",
+  "createChallengeCloneTasks": "Clone Challenge Tasks",
```
Alys: @EmilyOng If you do `git grep createChallengeAddTasks` in your local install, you'll see that there's other code that still uses createChallengeAddTasks. Because of that, the string should not be deleted or changed. You'll want to leave createChallengeAddTasks as it is and just add the new createChallengeCloneTasks string. :)

EmilyOng: Okay :)

paglias:
@EmilyOng thanks for this, I'm merging the PR, if you can provide us your Habitica UUID I'll be happy to award your contributor tier :)

Issue #9186, opened 2017-10-14
Subscription termination date should obey date format preference


lemoness:
Right now,  the subscription termination date found under User icon > Settings > Subscription page is not displayed according to the user’s date format preference. This can lead to confusion.

tim1234ltp:
I guess the subscription date should be using date format set by users in setting->site?
I would like to take this issue on hand.

Alys:
@tim1234ltp Yes, although the setting is in User Icon > Settings now. Thanks for taking it on! Post here if you have questions or find problems.

tim1234ltp:
@Alys Is there any options to give my test account subscription in my local Habitica? I couldn't find it in debug menu.

Alys:
@tim1234ltp Use a mongodb command to set the account's `purchased.plan` to this:
```
{
	"owner": "the-UUID-of-your-test-account", // e.g., "owner": "7bedfefa-d0c6-4cd0-b89f-b039b50a95c4",
	"dateCreated"    : new Date("2017-10-11T01:00:00.000Z"),
	"dateUpdated"    : new Date("2017-10-11T01:00:00.000Z"),
	"dateTerminated" : new Date("2018-12-13T01:00:00.000Z"),
	"gemsBought" : 0,
	"extraMonths" : 0,
	"quantity" : 1,
	"mysteryItems" : [ ],
	"consecutive" : {
		"count" : 0,
		"offset":0,  "gemCapExtra":0,  "trinkets":0,
	},
	"customerId" : "Gift",
},
```

tim1234ltp:
@Alys  I have been struggling with TravisCI check. The error shown is said to be with the unit/libs/payments/group-plans/group-payments-create.test.js.
Yet, what I have done is only changing subscription.vue in website/client/component/setting. And I am quite sure my repo is up to date.

The only thing I suspect is that termination date depends on total month of the group and personal plan. And my implementation only cares personal plan, same as the original code's logic.

I may need some tips with this.

PR #9313, opened 2017-10-28
Bug fixes on Subscription termination date format [Fixes Issues #9186]

tim1234ltp:
Fixes https://github.com/HabitRPG/habitica/issues/9186

### Changes
Created a computed value of Subscription termination date which is formatted according to user preference.
The client will show this computed value instead of using a filter on raw value.

### Result (sorry for the non-english strings)
default

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "habitica",
-  "version": "4.6.2",
+  "version": "4.6.3",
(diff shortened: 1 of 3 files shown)


tim1234ltp:
@Alys The commit is all fine in my local environment but the auto checks have failed. Could you shed some lights on those errors as I am not familiar with travis-ci.

PR #9321, opened 2017-10-30
[WIP] Add initial fixes for concurrency

TheHollidayInn:


website/server/controllers/api-v3/tags.js
```diff
@@ -216,8 +217,13 @@ api.deleteTag = {
     let validationErrors = req.validationErrors();
     if (validationErrors) throw validationErrors;
 
-    let tag = removeFromArray(user.tags, { id: req.params.tagId });
-    if (!tag) throw new NotFound(res.t('tagNotFound'));
+    let tagFound = find(user.tags, (tag) => {
+      return tag.id === req.params.tagId;
+    });
+    if (!tagFound) throw new NotFound(res.t('tagNotFound'));
+    await user.update({
```
paglias: there's a remaining user.save() at the bottom of this file, is it needed ?

paglias: at line 237

website/server/controllers/api-v3/tasks.js
```diff
@@ -581,12 +581,18 @@ api.scoreTask = {
...
+          $pull: { 'tasksOrder.todos': task._id }
+        }).exec();
       } else if (wasCompleted && !task.completed) {
-        let hasTask = removeFromArray(user.tasksOrder.todos, task._id);
-        if (!hasTask) {
-          user.tasksOrder.todos.push(task._id);
-        } // If for some reason it hadn't been removed previously don't do anything
+        // @TOOD: Why do we pull and push
+        // let hasTask = removeFromArray(user.tasksOrder.todos, task._id);
+        await user.update({
```
paglias: we're still saving the user below

paglias (changes requested):
I'm a little concerned about places where we'll have both an `update´ op and a `save` op, this means doubling the number of database ops and two roundtrips to the db server. I'm all for these changes but we should only save/update each object once per route.

Another problem is here https://github.com/HabitRPG/habitica/pull/9321/files#diff-08839b15ad090f22b2c8fa5e105c976bR237, we're removing tags from the document but not updating the local version of the user that is then saved but we're saving a user with still the tags that have only been removed by the update which doesn't update the in memory version of the document.

But doing the opposite would also have problems as we would have updated the db version and also the local one so when we `.save()` the user it will send once again the changes made to the tags.

So we should really make sure that:

1. only `save` or `update` is used in a route, not both
2. the local version of the doc should be updated too (update doesn't change the in memory doc)

Other problems:

1. We have to make sure the local version of the user gets the `_v` field at least updated since it's used by the client to identify if we're out of date
2. What about all the code in the pre save hook? That's skipped by the update op

website/server/controllers/api-v3/tasks.js
TheHollidayInn: Yea, I figured I'd need to remove the save's from most of it. I posted this to make sure I am heading in the write direction. I didn't realize using update skips much of the mongoose features.

TheHollidayInn: Hmm. I actually don't know if I can prevent this right now. This function probably needs to good refactoring to see.

paglias:
There's a small problem with these, we update the user version when updating a doc https://github.com/HabitRPG/habitica/blob/develop/website/server/models/user/hooks.js#L276, the problem is that this new version isn't returned by the response's `userV` field because it's only updated in mongo but not in memory. This in turn causes the next action on the site to trigger a full sync of the user and their tasks on the website

The problem is that `pre('update'` hooks don't have access to the in memory doc (since it could be missing at all from memory) so we cannot update it there. We can hardcode the update but it's easy to forget it

paglias:
There's a small problem with these, we update the user version when updating a doc https://github.com/HabitRPG/habitica/blob/develop/website/server/models/user/hooks.js#L276, the problem is that this new version isn't returned by the response's `userV` field because it's only updated in mongo but not in memory. This in turn causes the next action on the site to trigger a full sync of the user and their tasks on the website

The problem is that `pre('update'` hooks don't have access to the in memory doc (since it could be missing at all from memory) so we cannot update it there. We can hardcode the update but it's easy to forget it

paglias:
For now I've added a quick fix here https://github.com/HabitRPG/habitica/pull/9716/commits/b06a4b0274c4fed94431b4aa4100130d2487b893 but it would be better to revisit

PR #9358, opened 2017-11-01
(ISSUE-9353) Fix pinned item alignment

### LINKED ISSUE #9353: Pinned Items misaligned
Pinned items are not aligning to the grid properly when there are only two items in a row (ex. 2, 5, 8...)
screen shot 2017-10-31 at 7 15 57 pm

rholinshead:

Fixes https://github.com/HabitRPG/habitica/issues/9353

### Changes

- Get rid of justify-content: space-between to allow flex-start default to be used
- Add margin to direct children (item-wrappers)

----
UUID: 9135bc92-bf74-4c15-82c2-64f1128fe950


website/client/components/tasks/column.vue
```diff
@@ -69,7 +69,9 @@
   .reward-items {
     display: flex;
     flex-wrap: wrap;
-    justify-content: space-between;
+    > {
```
paglias: what is this selector for? maybe you meant `& > *` that selects all direct children?

paglias (changes requested):

website/client/components/tasks/column.vue
rholinshead: My intention (and what just using `>` seems to be doing) was to select all direct children. Since each `shopItem` is constructed with a wrapping `div` I think I'll change this to `& > div` for a more explicit selection of those `divs`. It doesn't seem like the `&` is actually needed here but if that's the convention that should be used I will stick with it.

paglias (approved):

paglias:
LGTM but I haven't live tested the PR

rholinshead:
Is there some way to re-trigger the builds for this? Not sure why the tests are failing but I doubt it would be caused by this small ui change ...

paglias:
@rholinshead don't worry about the failing tests, we're having some problems with the build not passing right now but it's not related to your changes :)

paglias:
before
schermata 2017-11-14 alle 23 35 23

after
schermata 2017-11-14 alle 23 35 35

@rholinshead can you make sure the spacing remains the same?

rholinshead:
@paglias I will certainly try and fix the spacing so that it's closer to what it was, but there is only one way (next paragraph) to make sure it is the exact same (since it is dynamically sized due to the `justify-content`).

This is a very common issue with flexbox, and some SO answers rely on using javascript to dynamically add "blank" divs to the list of items so that the last row is always full. Personally I find that implementation very ugly due to constant calculations needed to see how many items fit in the current screen size's rewards-list (what do you think)? 

If you think that invisible divs is the way to go I can change it. Otherwise, I will spend some time cleaning up the margins to try and match the current spacing as much as possible.

paglias:
@rholinshead I agree with you that we should avoid blank divs if possible!

rholinshead:
@paglias, sorry it's taken me so long to get back to this - have been pretty busy lately. I checked out the Zeplin design specs for the items and it shows them spaced at 16px apart. I've changed the margins here to reflect that. I also added a media query with `grid` styling so that when `grid` is supported, the item spacing looks exactly like the mockup:
pinneditemspacing

If `grid` isn't supported, it'll keep on using `flex` with the fixed spacing:
pinneditemspacingflex


In this case, the right edge of items in the last row don't quite match up exactly with the grey "Add a Reward" box before being pushed to the next row, since doing so would require dynamically setting the margins on the items so that if they are the last one in a row they don't have the right margin. This seems like an unrealistic thing to do though and the use of `grid` will solve that

paglias:
@rholinshead this looks good, thanks! It'll be merged next week :)

SabreCat:
Doing a fresh merge of `develop` to see if it straightens out the CI failure.

SabreCat:
There we go!

PR #9411, opened 2017-11-06
Fix regex for chat highlighting

### LINKED ISSUE #9405: special characters in profile names cause chat message highlighting to fail (regex error)
If a player's name contains a closing bracket `)` and if a message to that player exists in any chat area, then a regular expression error happens when the player views the chat area and they are not able to see the message to them, or any messages that appear after it. This was reported by the player "Stay Whelmed :)" 7d8263bd-7b74-4d71-8da2-95ee147b14fe but I've replicated it with a test user.

Below is a screenshot of a party containing the test user when the user's name was set to `name with bracket :)`  Party chat contains a message saying `Hi @name with bracket :)` however as you can see in that screenshot, the message doesn't appear and the javascript console shows a regex error `SyntaxError: Invalid regular expression: /name with bracket :)([^w]|$){1}/: Unmatched ')'(…)`

image

When I changed the test user's name to remove the bracket, the regex error stopped happening and the player was able to see all chat messages:

image

This indicates that the regex error is happening when chat is being parsed to highlight all messages to the player. Special characters in the player's name should be escaped to allow the regex to work.

librarianmage:

Fixes #9405 

### Changes

Escapes usernames so chat matching works with users that have special characters in their display name.

----
UUID:

### DIFF
--- a/website/client/components/chat/chatMessages.vue
+++ b/website/client/components/chat/chatMessages.vue
@@ -353,7 +353,7 @@ export default {
       let messagetext = message.text.toLowerCase();
       let username = user.profile.name;
       let mentioned = messagetext.indexOf(username.toLowerCase());
-      let escapedUsername = escapeRegExp(username)
+      let escapedUsername = escapeRegExp(username);
       let pattern = `${escapedUsername}([^\w]|$){1}`;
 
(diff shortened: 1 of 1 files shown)


paglias:
thanks, notated towards your next tier!

PR #9466, opened 2017-11-13
Check previous gear owned before purchasing next level gear (fixes #9071)

### LINKED ISSUE #9071: API should block purchase of high-level basic class equipment until low-level gear is bought
Currently, the API allows you to purchase any of the basic class-specific gear, regardless of whether you've bought lower-level pieces. For example, using the API you can buy the Warrior's Golden Sword before you've bought the Warrior's Training Sword.

The API should prevent this.

matchajune:
Hello everyone!

I would like to have another set of eyes review my code before submitting my PR for review!

Thanks!

[x]: Fixes https://github.com/HabitRPG/habitica/issues/9071

### Changes
[x]: Add validation if gear level isn't 0 AND user has the previous gear. It gear level is 0, user can purchase the gear and no check for previous gear is necessary.

----


paglias:
@matchajune sorry for the delay here! Is this PR ready or still WIP? Could you add some tests for this functionality? Let me know if you need any help

matchajune:
@paglias I finished adding an integration test and everything is working with tests passing. PR ready.

test/api/v3/integration/user/buy/POST-user_buy_gear.test.js
```diff
@@ -25,12 +25,23 @@ describe('POST /user/buy-gear/:key', () => {
...
   });
+
+  it('tries to buy subsequent, level gear', async () => {
+    let key = 'armor_warrior_2';
+
+    return expect(user.post(`/user/buy-gear/${key}`))
+      .to.eventually.be.rejected.and.eql({
+        code: 401,
+        error: 'NotAuthorized',
+        message: 'String \'previousGearNotOwned\' not found.',
```
paglias: can you add the `previousGearNotOwned` string to website/common/locales/en?

paglias (changes requested):

test/api/v3/integration/user/buy/POST-user_buy_gear.test.js
matchajune: It's been fixed. I understand the message now :)

website/common/script/ops/buyGear.js
```diff
@@ -35,6 +35,13 @@ module.exports = function buyGear (user, req = {}, analytics) {
     throw new NotAuthorized(i18n.t('equipmentAlreadyOwned', req.language));
   }
 
+  let gearNumber = Number(item.key.slice(-1));
```
paglias: Why are you using in the first place .splice with a fixed number of chars selected and in the second a regexp that can match a different number of charachters?

Isn't the regexp going to cause problems if the key includes other numbers? Since all keys follow the same form (`type_class_number`) can you take the gearNumber from `key.split('_')[2]`, covert it to a number and check that it's not Nan as there are also gear items without a number part like `eyewear_special_summerWarrior`

paglias: And use the same method to generate the previous gear key

paglias (changes requested):

website/common/script/ops/buyGear.js
```diff
@@ -35,6 +35,15 @@ module.exports = function buyGear (user, req = {}, analytics) {
     throw new NotAuthorized(i18n.t('equipmentAlreadyOwned', req.language));
   }
 
+  let itemIndex = item.index;
```
paglias: This new check seems right but I think there should be a check for the cases where item.index is not a number, like here 
schermata 2017-11-27 alle 13 01 46

Maybe we could simply exclude from the check items when `Number.isNaN(Number(item.index)) === true` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN)

paglias (changes requested):

website/common/script/ops/buyGear.js
matchajune: I tried both ways to check for NaN and I keep getting that `isNan()` is not a function. Since NaN is the only object that doesn't equal itself, I checked for NaN this way.

Let me know what you think.

matchajune:
Hello @paglias ,

The check is now only being done for indices that are integers. May you please rerun the TravisCI test since the reason why it failed was independent to the PR changes?

Thanks!!

paglias (approved):

SabreCat:
Merged! @matchajune, we will use this in a staging environment for a bit to ensure no unexpected side effects have occurred, but you will see this live in a few days. I've awarded you your first Blacksmith contributor tier for this effort, thank you! Note that subsequent tiers require increasing amounts of effort from one to the next, so tier 2 may require multiple PRs or a larger PR to attain. But keep helping out and we will express our gratitude accordingly!

matchajune:
@SabreCat Sounds good! Feel free to ping me if side effects do occur!

Thank you for the award! 👍

PR #9516, opened 2017-11-16
Fix regex for message highlighting

### LINKED ISSUE #9506: Username notifying is too broad
The purple dot that notifies people when they've been tagged in a post is too widely applied. For example, a user named "The H" sees the purple dot appears every time the word "the" precedes the letter "h", for example in the phrase "the Habiticans."

librarianmage:
Fixes #9506 

### Changes

I accidentally mixed up the highlighting in my last PR, so here is the fix!

----
UUID:


SabreCat:
Thanks for sending this along, @MathWhiz !

Issue #9286, opened 2017-10-27
My Challenges should include all Owned Challenges


lemoness:
The My Challenges sections of the site should include:

-  all of a user's Owned Challenges, even if they aren't currently part of those Challenges. 
- Challenges that they own which are located in a Private Guild.
- Challenges that they have joined which are located in a Private Guild.

pcvonz:
I can try and take care of this one.

pcvonz:
Did some digging into the ownership issue:

On the My Challenges page the filter will always check if the user is a member of the challenge since the `participating` role is always passed. The easy solution is to not pass the `participating` role when `Owned` is checked in the sidebar. But that'd mean that the user wouldn't be able to see every challenge they own but aren't participating in and all participating challenges at once, which I don't think is what we want. So, I'm thinking of writing a new filter function specifically for the My Challenges page.

Another thing I noticed is that the sidebar doesn't reflect what is actually being passed to the filter. For instance, the participating option should be checked on the My Challenges page.

lemoness:
Good point about the filter! Thanks very much for tackling this.

I've updated the labels on this ticket.

TheHollidayInn:
Yea, the issue is we first wanted to only included participating challenges. To fix this, you'll probably need to:
1. Remove this: https://github.com/HabitRPG/habitica/blob/develop/website/client/components/challenges/myChallenges.vue#L115
2. Replace this: https://github.com/HabitRPG/habitica/blob/develop/website/client/components/challenges/myChallenges.vue#L130 with an "Owned" filter because our API doesn't return only owned challenges if I remember. 

I don't think there is a need to write a custom filter for my challenges, but I'm open to seeing what you have. Maybe just go with the fix first though.

marvinrabe:
I will do this.

PR #9535, opened 2017-11-18
Show not participating and owned challenges on MyChallenges page.

pcvonz:

#9286

### Changes

* Remove Participating/Not Participating filter. I believe if a user is viewing their challenges, they of course want to see only participating challenges (and owned and not participating challenges). If they are on the discover page, they will want to see only challenges they aren't already participating in.
* Separate the ownership variable in the `challengeUtility` into two variables `owned` and `notOwned`. Necessary for filtering `owned` challenges. Before you couldn't see both Not Owned and Owned challenges because the `ownerShip` variable would be flipped twice, only displaying `owned` challenges. 
* MyChallenges page now displays owned and not participating and participating challenges by default. 
* Removed some unused variables from the `filterChallenges` function. 

NOTE: I referenced issue #9286, but It'd be nice to get a second look into the displaying private challenges issue. I set up a private guild and created some challenges with my test user and had another test user create a few challenges as well. Both challenges showed up fine, so something to look out for!

----
UUID: b16e0c46-c20b-4aea-9f1f-6e9dd80ad61b

### DIFF
--- a/website/client/components/challenges/findChallenges.vue
+++ b/website/client/components/challenges/findChallenges.vue
@@ -108,6 +108,10 @@ export default {
       if (!this.filters.ownership || this.filters.ownership.length === 0) {
         this.filters.ownership = ['owned', 'not_owned'];
       }
+      // If roles is not defined, show all tasks
+      if (!this.filters.roles || this.filters.roles.length === 0) {
+        this.filters.roles = ['participating', 'not_participating'];
+      }
 
       // @TODO: Move this to the server
       return this.challenges.filter((challenge) => {
--- a/website/client/components/challenges/findChallenges.vue
+++ b/website/client/components/challenges/findChallenges.vue
@@ -103,6 +103,12 @@ export default {
       let search = this.search;
       let filters = this.filters;
       let user = this.$store.state.user.data;
+
+      // If ownership is not defined, show all tasks
+      if (!this.filters.ownership || this.filters.ownership.length === 0) {
+        this.filters.ownership = ['owned', 'not_owned'];
+      }
+
       // @TODO: Move this to the server
       return this.challenges.filter((challenge) => {
         return this.filterChallenge(challenge, filters, search, user);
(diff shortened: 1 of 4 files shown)


Alys:
I might be misunderstanding this, but you're changing the My Challenges page to contain all of these:
- challenges you own and are participating in
- challenges you own and are not participating in
- challenges someone else owns that you are participating in

Is that right? If so, I don't think this is desirable: "Remove Participating/Not Participating filter"
When I have a page that contains all the challenges I'm involved in, I want to be able to filter it based on my participation. Currently I own about 20 challenges I think (the spread out design of the page makes it too annoying to count them exactly) but I'm participating in only 3, so that filter will be useful.

pcvonz:
Ah, I can see where that'd be useful. I can can add the participating filter back in. I wasn't aware of how many challenges a person would create but not participate in.

pcvonz:
Alright, added the `participating` and `not_participating` filter back in. Let me know if I should change anything else. Thanks!

website/client/components/challenges/myChallenges.vue
```diff
@@ -126,8 +126,14 @@ export default {
       let filters = this.filters;
```
paglias, line 126: since we have filters defined here as a var, we should use it all the times or never (and instead use `this.filters`), both are okay but we should stick to one :)

website/client/components/challenges/findChallenges.vue
```diff
@@ -103,6 +103,16 @@ export default {
       let search = this.search;
       let filters = this.filters;
       let user = this.$store.state.user.data;
+
```
paglias, line 106: This entire `filteredChallenges` function seems to be duplicate here and in `website/client/components/challenges/myChallenges.vue`, would you mind moving it to client/store/getters/challenges.js and then use just add it to this file with `mapGetters(...)`  like here https://github.com/HabitRPG/habitica/blob/cd92f44365d6dff4fe265b850a671e379369fe29/website/client/components/tasks/taskModal.vue#L599

paglias (changes requested):
Left two comments but the code looks good!

paglias:
hey @pcvonz how are you doing on this?

lemoness:
Since we haven't had any response on this since December, this PR should be closed or taken over. @paglias @TheHollidayInn @SabreCat

lemoness:

SabreCat:
Addressed in a different PR

Issue #9393, opened 2017-11-03
Close Challenge Popup / modal appears erroneously


lemoness:
When editing a Challenge after closing a previous Challenge, the Challenge close dialog pops up after saving the edits to the Challenge. This shouldn't happen.

aalsehly86:
Hello @lemoness I can take care of this.

lemoness:
@aalsehly86 Thank you so much! Let us know if you have any questions.

I've updated the labels on this ticket.

lemoness:
@aalsehly86 Hi there! How are things going with this fix?

aalsehly86:
The @lemoness! I've submitted some changes to modal stack on PR #9545. I need you guys to test it and let me know please. Hopefully it will take care of this annoying issue because I saw some more users reporting it. Thanks!

lemoness:
Hurrah, thanks for letting us know! We'll take a look at that PR as soon as we can :)

Issue #9531, opened 2017-11-18
hatching a pet gives a second, empty hatching dialog


Alys:
From AbelianGrape ( e723ec48-2de6-4560-8f89-ed505a3efdf3 ) in the Report a Bug guild:

"I hatched some pets today and after clicking the Onward! response to [this dialogue](https://pasteboard.co/GTYF7eC.png), a nearly empty hatching dialogue containing only the response buttons kept popping up again at the top of the page. It looked like [this](https://pasteboard.co/GTYFsxk.png). I tried clicking on the Hatch button when I took these screenshots just now, but it didn't seem to do anything."

It was also  reported by jaolsn ( 04fd7a1a-1964-49c6-a2cb-59d2d231c7f6 ).

aalsehly86:
@Alys I'm thinking this is related to issues #9393 and #9536!

Alys:
@aalsehly86 Could be! There are definitely problems related to inappropriate modals reappearing, but there might be a few different cases. It's known that modals you've seen before can reappear when they're no longer relevant but that's not easily reproducible as far as I'm aware, whereas this issue is which made me think it had its own specific cause. 

I'd be interested to know if you can reproduce this issue on your local install when you are using the `develop` branch and if  you can then not reproduce it when using your fix in https://github.com/HabitRPG/habitica/pull/9545. If your PR fixes this one as well as #9393, we can close this one when your PR is merged.

aalsehly86:
@Alys The method used in PR #9545 fixes this problem, and I was able to reproduce it.
![hatch-modal-error](https://user-images.githubusercontent.com/30061021/33046573-8e723160-ce1f-11e7-9c5c-88bc00b1d6be.gif)

> Clicking `cancel` will exit the empty hatching modal

Now this behavior is stopped once I change what triggers that modal in `website/client/components/inventory/stable/hatchedPetDialog.vue` 

That is changing `this.$root.$emit('bv::show::modal', 'hatchedPet-modal')` to `this.$root.$emit('bv::show::modal', 'hatchedPet-modal', '$event.target')`

Notice that in #9393, passing `'bv::hide::modal'` as a third argument stopped the wrong challenge modal from reappearing, and here `'$event.target'` also does the same job for hatching modals! Apparently passing a third argument, no matter what, will stop emitting the wrong modal!

Now after changing the code, it works fine
[...]

PR #9545, opened 2017-11-18
WIP workaround to stop the random behavior mentioned in issue #9393 (Close Challenge Popup / modal appears erroneously)

aalsehly86:
Fixes https://github.com/HabitRPG/habitica/issues/9393

**### Changes**
This will stop the random behavior of the challenge modals. Apparently the problem is caused not only when deleting a challenge, but also when creating one. 

In addition to details given in issue #9393, to reproduce the error when creating a challenge: 
1- Create a new challenge
2- Once created and redirected to the new challenge, click on **Edit Challenge** button but don't change anything
3- Exit the modal by clicking outside the modal dialog
4- Click on **End Challenge** button and exit the same way

Now you'll notice the **Editing Challenge** modal pop up after exiting the **End Challenge** one. This behavior will persist now every time **End Challenge** is clicked until the page is refreshed. 

Notice this behavior will change if **Clone** button is clicked instead of **Edit Challenge** in _step 2_. The only difference now is the modal's title changes from **Editing Challenge** to **Create Challenge**.

Also notice if **Add Task** button is clicked instead of **End Challenge** in _step 4_, the **Editing Challenge** modal will pop up after exiting the **Add Task** modal.

See [Returning focus to the triggering element](https://bootstrap-vue.js.org/docs/components/modal/) for more details on this behavior. In this case, it seems that whatever is being passed as a third argument in each `$emit` statement, the behavior stops.

Thanks.

----
UUID: 4998c3fd-1c32-4ad2-bd2d-cc09823d83c2

### DIFF
--- a/website/client/components/challenges/challengeDetail.vue
+++ b/website/client/components/challenges/challengeDetail.vue
@@ -329,7 +329,7 @@ export default {
       this.workingTask = this.editingTask;
       // Necessary otherwise the first time the modal is not rendered
       Vue.nextTick(() => {
-        this.$root.$emit('bv::show::modal', 'task-modal');
+        this.$root.$emit('bv::show::modal', 'task-modal', 'bv::hide::modal');
       });
     },
     createTask (type) {
@@ -338,7 +338,7 @@ export default {
       this.workingTask = this.creatingTask;
       // Necessary otherwise the first time the modal is not rendered
       Vue.nextTick(() => {
-        this.$root.$emit('bv::show::modal', 'task-modal');
+        this.$root.$emit('bv::show::modal', 'task-modal', 'bv::hide::modal');
       });
     },
     cancelTaskModal () {
@@ -365,7 +365,7 @@ export default {
       this.$store.state.memberModalOptions.groupId = 'challenge'; // @TODO: change these terrible settings
       this.$store.state.memberModalOptions.group = this.group;
       this.$store.state.memberModalOptions.viewingMembers = this.members;
-      this.$root.$emit('bv::show::modal', 'members-modal');
+      this.$root.$emit('bv::show::modal', 'members-modal', 'bv::hide::modal');
     },
     async joinChallenge () {
       this.user.challenges.push(this.searchId);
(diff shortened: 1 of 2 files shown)


website/client/components/challenges/challengeDetail.vue
```diff
@@ -365,32 +365,32 @@ export default {
...
+      this.$root.$emit('bv::show::modal', 'challenge-modal', 'bv::hide::modal');
     },
     // @TODO: view members
     updatedChallenge (eventData) {
       Object.assign(this.challenge, eventData.challenge);
     },
     openMemberProgressModal (memberId) {
       this.progressMemberId = memberId;
-      this.$root.$emit('bv::show::modal', 'challenge-member-modal');
+      this.$root.$emit('bv::show::modal', 'challenge-member-modal', 'bv::hide::modal');
```
paglias, line 393: Can we add a comment above these calls to explain what's happening and why the third argument is necessary? Have you figured out why this is happening only with certain modals?

@TheHollidayInn you might want to give a look at this

aalsehly86: @paglias, unfortunately I don't have that much of experience to figure out why some modals appear randomly. Could it be caused by the recent bootstrap-vue upgrade? 

Do you think this have anything to do with this issue? 
[ **Auto Return Focus**
When < b-modal > is opened via the < v-b-modal > directive on an element, focus will be returned to this element automatically when < b-modal > closes, unless an element has been specified via the return-focus prop.](https://bootstrap-vue.js.org/docs/components/modal/)

@Alys suggested we include the fix for issue #9531 with this PR. I'll wait for @TheHollidayInn to review this first. Thanks.

TheHollidayInn:
I apologize for taking so long to respond. So, the fix you propose may work, but the true issue is that when you create a challenge, then the user is redirected without the `bv:hidden` event launched: https://github.com/HabitRPG/habitica/blob/develop/website/client/app.vue#L343. 

That event needs to be thrown so that our modal stack can remove the modal. Something in the redirect must be preventing the event from emitting. Since the create and edit modal share the same component, that is why you see the edit modal.

aalsehly86:
@TheHollidayInn, thank you so much for your review. As you mentioned: 

> Since the create and edit modal share the same component, that is why you see the edit modal.

Now what about the other random modals that still appear randomly? If I'm not mistaken, the delete modal and other modals (i.e. hatching & purchase) across the website also appear randomly so not only create & edit.
I have no idea how to deal with this modal stack though. I tried to debug and get better understanding on what's going on using the Vue Devtools on my browser but my big old laptop keeps crashing once I use it:(

I apologize I'm not an expert in web applications and don't have the right tools yet, just a decent student trying to help out and learn! Doesn't mean I'm giving up on this issue but will try my best to get better understanding. It might take me longer time though as I'm so busy with our engineering system design project for a showcase in April. Y'all welcome to attend, Friday April 6 at Florida Tech.

Your help is always appreciated so thank you again for your review and sorry everyone for the inconvenience!

TheHollidayInn:
No worries! The modal stack is something we added - so it is domain specific. It's not something in Vue and therefore not as easy to learn about.

I believe the other modals are having the same issues. I'd have to check on them one by one, but it seems to me the issue is their hide event is not allowed time to emit.

Since the third parameter is for return focus, I believe there may just be an error preventing the modals from being shown again. So it stops them from showing as a side effect rather than intentional.

paglias:
How is this going @aalsehly86 ? Let me know if you need any help

aalsehly86:
@paglias I apologize for taking so long to respond. I tested something and I think it worked. However, when I updated the repo something went wrong and the server won't start. I don't really have time to look into this now. I would appreciate it if you can test the `app.vue` file and let me know if it works! Thanks:)

lemoness:

Issue #9371, opened 2017-11-02
Opening the party page makes too many requests


paglias:
Every time the party page is opened all these requests are made to the server:

<img width="461" alt="schermata 2017-11-02 alle 10 01 53" src="https://user-images.githubusercontent.com/589911/32317672-e1e1390a-bfb4-11e7-9cbf-14d87ff117cb.png">

There's an extra party request and one for the public guilds that shouln't be there at all

gilmoreg:
I'm looking at this, the public guilds call is made when the ChallengeModal mounts (so you can attach a challenge to a Public Challenges if you have the gems). Unless there's a way to stop it from mounting until it is actually displayed that call will have to stay there.

The duplicate call to /api/v3/groups/party is due to the ChallengeModal dispatching the guilds:getGroup action in addition to Group. The obvious solution would seem to be to save the results of the API call in the store the first time and then fetch it from there afterwards. However I don't know the codebase well enough to know what side effects this might have, and I can't see that there are a lot of client side tests that would make it easier to change this confidently. Any thoughts?

paglias:
> I'm looking at this, the public guilds call is made when the ChallengeModal mounts (so you can attach a challenge to a Public Challenges if you have the gems). Unless there's a way to stop it from mounting until it is actually displayed that call will have to stay there.

we can pass an option to ChallengeModal modal to prevent it from doing that request?

> The duplicate call to /api/v3/groups/party is due to the ChallengeModal dispatching the guilds:getGroup action in addition to Group. The obvious solution would seem to be to save the results of the API call in the store the first time and then fetch it from there afterwards. However I don't know the codebase well enough to know what side effects this might have, and I can't see that there are a lot of client side tests that would make it easier to change this confidently. Any thoughts?

I think the party is already saved into the store since the party member are always in the header. There aren't many tests indeed but if you want to give it a shot I can help you if you have any question

gilmoreg:
Here's the code I think would be changed:
[https://github.com/HabitRPG/habitica/blob/develop/website/client/store/actions/guilds.js](https://github.com/HabitRPG/habitica/blob/develop/website/client/store/actions/guilds.js
)
```
export async function getGroup (store, payload) {
  let response = await axios.get(`/api/v3/groups/${payload.groupId}`);

  // @TODO: should we store the active group for modifying?
  // let guilds = response.data.data;
  // store.state.myGuilds = guilds;

  // @TODO: Populate wiht members, challenges, and invites

[...]

PR #9546, opened 2017-11-18
Challenge modal optimization - remove unnecessary API call  - partial fix for #9371

gilmoreg:
Partial fix for #9371 

### Changes
ChallengeModal was calling the API to get party data even though that data is already available from having loaded the Group component. Attached the 'party' state from the store to the ChallengeModal so that it could attempt to find it there first; if it is not found, it calls the API for it.

ChallengeModal still calls getMyGuilds so the issue described in 9371 may not be entirely fixed.

Also fixed a template syntax error in menu.vue to allow local tests to pass. There is already a PR in from a different user to fix this issue but I wanted the tests to pass on my install before that gets merged in.

I'd love to add component tests to test this change but there's not much to go on yet in terms of a testing strategy. Would love some feedback on how to test this out.

----
UUID: 74e5e6f8-cb86-45c3-a993-872cfc407094


website/client/components/challenges/challengeModal.vue
```diff
@@ -233,7 +233,10 @@ export default {
   async mounted () {
     this.groups = await this.$store.dispatch('guilds:getMyGuilds');
     if (this.user.party._id) {
-      let party = await this.$store.dispatch('guilds:getGroup', {groupId: 'party'});
+      let party;
```
paglias: two things: can you fix the formatting of the `if`? But mainly I'd moev the code to inside the `guilds:getGroup` action so that when it's called the group is fetched from the server only if necessary, so that this code doesn't have to be added everywhere. And add an optional `forceReload` parameter to the action that if true reloads the data from the server even if it's already cached

paglias (dismissed):

website/client/components/challenges/challengeModal.vue
```diff
@@ -233,7 +233,10 @@ export default {
   async mounted () {
     this.groups = await this.$store.dispatch('guilds:getMyGuilds');
```
paglias: we could move this code from the mounted event that is called every time to only when the modal is actually shown for example by calling a method when the modal is shown. You can attach a method to the shown event by adding `@shown="methodName"` at line 2 in this file

paglias: more info here https://bootstrap-vue.js.org/docs/components/modal

website/client/components/challenges/challengeModal.vue
gilmoreg: Great idea - though I can't figure out how to return store data from inside an action, because inside actions things like store.state.user.data.party are Observers rather than data that can be returned. I tried reading the Vuex docs but Habitica does things differently (Vuex advises against returning anything from actions at all).

If there's no way to access store data in a returnable format from inside actions, where else could this function go that would make it reusable?

Also what was wrong with the format of the if? I tried to copy the code style I saw elsewhere.

website/client/components/challenges/challengeModal.vue
gilmoreg: I made and pushed this change to the PR, but it isn't finished until I resolve the first issue you commented on. But this should work as proof of concept - the API call isn't made until the modal is shown. Thanks for the input - I knew there had to be something like this (and I should have just looked it up!)

website/client/components/challenges/challengeModal.vue
paglias: what do you need to return?

You can also not return anything and take the updated value using `this.$store...` but in general Observers are fine to return, it just means that Vue is observing them but if you convert them to JSON with JSON.stringify() you'll see that they're actually normal data, so don't worry about them.

Let me know if you have questions and if you get stuck feel free to push the code you have and I'll give a look :)

The if on a single line is fine if it's on its own but when there's also an else it should follow this form:

```
if (condition) {

} else {

}
```

gilmoreg: Awesome. Let me try to explain what I'm after (trying to put together all of your suggestions) and then this should be good to go.

You suggest quite rightly that we move the if statement to the ```guilds:getGroup``` action so that it doesn't have to be repeated in every component. For that to work though, I think, we'd need to be able to do just this in the component:
```
const party = await this.$store.dispatch('guilds:getGroup', {groupId: 'party'});
```
And the action would then check the store and return that or else return the results of the API call. I tried code like this:
```
export async function getGroup (store, payload) {
  // Unless forceReload is true, if we are looking for the party and it exists in the store, return that
  // Otherwise call the API
  if (payload.groupId === 'party' && store.state.party.data && !payload.forceReload) {
    return store.state.party;
  } else {
    let response = await axios.get(`/api/v3/groups/${payload.groupId}`);
    // @TODO: should we store the active group for modifying?
    // let guilds = response.data.data;
    // store.state.myGuilds = guilds;

    // @TODO: Populate with members, challenges, and invites

    return response.data.data;
  }
}
```
Here's the problem I have right now: if I return ```store.state.party.data```, I can just leave the code in ```challengeModal.vue``` as is. However, this breaks other components (like ```questDetailsModal.vue``` (somehow party.members is no longer an array in the store if I do that).  If I return just ```store.state.party``` then I don't get any errors from other components, but I have to change the lines in challengeModal to this:
```
this.groups.push({
    name: party.data.name,
    _id: party.data._id,
    privacy: 'private',
  });
```
This is the code I've pushed by this point. It seems to work without any errors, but I can't help feeling that it could be done better somehow.

paglias: Okay I think one problem could be when the party data is not cached you're returning the response but not updating the store, also `store.party` already exists (see client/store/index.js)

```
      party: {
        quest: {},
        members: asyncResourceFactory(),
      },
```

so it looks like you're not returning the correct party data but quest and members, try adding a new `data` object under store.party where you can save the actual party data

gilmoreg: Ok. While I'm at it, I've noticed that party data is saved in two different overlapping places: ```store.state.party``` and ```store.state.user.data.party```. The latter has an exact copy of the data stored in ```store.state.party.members.data```. That is, a subset of available data about the user's party is stored in two places. If these were consolidated that would make it a lot easier to reason about what is happening here - is there a reason the data is stored twice? Are there cases where this data would be different?

Also, the more I scratch my head with this code, the more I feel like fetching party data should be an action in ```party.js``` rather than ```guilds.js```. That way we wouldn't have to write code to handle the special case of a party in getGroup. Any thoughts?

Edit: some further thoughts. The party data in the store comes from calling ```getMembers``` in ```party.js``` and this is called by ```index.vue``` when the page loads. This calls the same API endpoint that getGroup does (with an additional parameter). So theoretically the data should always already be there and could just be referenced from the store, or if it needed to be refreshed it could just dispatch getMembers instead. I'll try pushing this and see what you think.

gilmoreg: All right, I just pushed another change. I simply take it for granted that the party data will be available in the store, and just omit the option to attach a challenge to a party if it isn't. This means no changes are required to the actions and no additional API calls need to be made. I think this is pretty clean, except that mapping to ```party.members.data``` returns an array, so you'll see at line 301 I had to index it, which is a bit ugly but it didn't look like ```mapState``` could be coerced into returning an index. Let me know what you think!

website/client/components/challenges/challengeModal.vue
```diff
@@ -243,7 +243,7 @@ export default {
     },
   },
   computed: {
-    ...mapState({user: 'user.data', party: 'party'}),
+    ...mapState({user: 'user.data', party: 'party.members.data'}),
```
paglias: I'm confused as to why the party data is inside `party.members.data`?

website/client/components/challenges/challengeModal.vue
paglias: > I simply take it for granted that the party data will be available in the store

What makes it available in the store?

Anyway when you fetch the party again the store should be updated

gilmoreg: I did an XHR breakpoint on calls to /group/party and the first one is called by ```header/index.vue```, which loads no matter which page you are on, and calls ```party:fetchMembers``` which itself calls the API, and this is how the data ends up in the store (through ```loadAsyncResource```). All further calls are to actions that just return the API response and do not modify the store, so far as I can tell.

This pattern is confusing to me - unless I'm missing something. The pattern followed by party:fetchMembers seems to be the right one - actions should modify the store, and the data should be gotten from the store. But that isn't how the majority of the actions I've been looking at are written.

website/client/components/challenges/challengeModal.vue
gilmoreg: Me too. Looking at ```party:fetchMembers```, I'd expect the API response to be plugged into ```store.state.party```. But when you log out the contents of the store, that is where the data is. I haven't changed anything about how this data was stored. I've looked around and haven't seen anywhere else yet that could be changing it.

website/client/components/challenges/challengeModal.vue
paglias: >  The pattern followed by party:fetchMembers seems to be the right one - actions should modify the store, and the data should be gotten from the store.

Yes that's the correct pattern, we're slowly refactoring actions to make them work that way but it'll take some time

> I did an XHR breakpoint on calls to /group/party and the first one is called by header/index.vue, which loads no matter which page you are on, and calls party:fetchMembers which itself calls the API, and this is how the data ends up in the store (through loadAsyncResource)

Okay, but `store.state.party.members` only contains the party members data, not the party data itself. So you should create a new action (yeah, let's put that into actions/party.js) that loads an async resource into `party.data` (or anothe sub property of `store.state.party`) like this

```
export function getParty (store, forceLoad = false) {
  return loadAsyncResource({
    store,
    path: 'party.data',
    url: '/api/v3/groups/party',
    deserialize (response) {
      return response.data.data;
    },
    forceLoad,
  });
}
```

and in `store/index.js`:

```
      party: {
        data: asyncResourceFactory(),
        quest: {},
        members: asyncResourceFactory(),
      },
```

I think you're on the right path but the party data should be under party.data, party.members only include the profile of party members :) If, for any reason, `party.members.data[0]` includes the party data it's a bug and somehow it's leaked from another part of the code but I don't think this is the case or viewing party members would be broken by now

website/client/components/challenges/challengeModal.vue
paglias: `party:fetchMembers` has `party.members` as the `path` option so it's normal that the data ends up there but we're talking about the party members profile data not the party group data (chat, leader, description, ...)

website/client/components/challenges/challengeModal.vue
gilmoreg: Ok this does look like the best plan. One more question then - when should we call ```party:getParty```? In my mind doing it in ```challengeModal``` seems wrong, as the parent page should already have that info. Maybe in ```group.vue```? Or somewhere higher up in the tree so that every component that might need party data can simply access it?

Either way, this code in ```group.vue``` would need to be changed, I think:
```
let group = await this.$store.dispatch('guilds:getGroup', {groupId: this.searchId});

if (this.isParty) {
  this.$store.state.party.data = group;
  this.group = this.$store.state.party.data;
  this.checkForAchievements();
  return;
}

this.$set(this, 'group', group);
```

I'm thinking of something along the following lines:
```
if (this.isParty) {
  const group = await this.$store.dispatch('party:getParty');
  this.group = this.$store.state.party.data;
  this.checkForAchievements();
} else {
  const group = await this.$store.dispatch('guilds:getGroup', {groupId: this.searchId});
  this.$set(this, 'group', group);
}
```

Edit: I'll also add a check to ```challengeModal``` to dispatch fetchParty if for some reason the data doesn't exist in the store.

paglias: Yeah, that sounds good, fetching the party in the challengeModal shouldn't be necessary as it's already loaded in a different place but since we're caching the value calling `await this.$store.dispatch('party:getParty');` once again is not going to cause problems + it could help if one day we decide to stop loading the party on page load. So I'd keep it

gilmoreg: All right I just pushed these changes. Let me know what you think!

gilmoreg: Anything look off so far?

website/client/components/challenges/challengeModal.vue
```diff
@@ -313,6 +296,24 @@ export default {
     },
   },
   methods: {
+    async shown () {
+      this.groups = await this.$store.dispatch('guilds:getMyGuilds');
+      const party = await this.$store.dispatch('party:getParty');
+      if (party.data._id) {
```
paglias: is party.data necessary here? Ideally `party:getParty` should return the party data directly

gilmoreg: Hrm, I did get the same feeling writing that code. I set it up that way because of ```asyncResourceFactory```. If ```party:getParty``` returns the party data directly, my understanding is that it also has to stash the party data directly. Looking at this code in ```store/index.js```:
```
party: {
        data: asyncResourceFactory(),
        quest: {},
        members: asyncResourceFactory(),
      },
```
How would ```party:getParty``` store the data directly without interfering with ```quest``` and ```members```? We can't just overwrite it with ```party:asyncResourceFactory(),```.

If we don't change any of this I could just do a destructuring assignment, a la:
```
const { partyId, partyName } = await this.$store.dispatch('party:getParty').data;
```
Or something along those lines, if you think that is cleaner.

paglias: ok, I think this is super confusing because we're using the asyncResourceFactory on a sub property.

The way it works, if you had it defined like this
```
party: asyncResourceFactory(),
```

the generated structure would look like this

```

Comment (paglias) [reply website/client/components/challenges/challengeModal.vue]: ok, I think this is super confusing because we're using the asyncResourceFactory on a sub property.

The way it works, if you had it defined like this
```
party: asyncResourceFactory(),
```

the generated structure would look like this

```
party: {data: ..., loadingStatus: ...},
```

so the party data would be available under `store.party.data`, but since we're storing the async resource under `party.data` it'll look like this

```
party: {
    quests: ...,
    members: ...,
    data: { data: ...., loadingStatus: ... }
}
```

so the party data will be under store.party.data.data (which is indeed confusing)

but party:getParty should return directly the party data, and I think it does since the getParty action is written exactly like the fetch user action. But as you said it's not the best practice to return data from an action so instead of doing this

```
const party = await this.$store.dispatch('party:getParty');
```

You can simply do 

```
await this.$store.dispatch('party:getParty');
const party = this.$store.state.party.data.data
```

and it'll be fine :)

Comment (gilmoreg) [reply website/client/components/challenges/challengeModal.vue]: I suppose I still don't understand what you mean by ```party:getParty``` returning the data directly. The last change you suggested seems to be the best option and I just pushed that. The alternative I thought of would be to use ```mapState``` for ```party.data```, await the ```party:getParty``` call and then look inside the component state. But I opted not to go that route since the party data isn't used anywhere else in the component.

**Comment by TheHollidayInn** [comment]
Sorry I'm coming in late to this, but is it possible to not have the extra data field under the party? Other code assumes the party object has the details, not a sub data field.

**Comment by gilmoreg** [comment]
@TheHollidayInn the way @paglias explains it covers pretty much what I would say. It's all pretty confusing the way it works right now - there's some data that is stored *both* in ```store.state.user.party``` and in ```store.state.party.members```, and as long as ```store.state.party``` has ```quest``` and ```members``` as immediate children the data can't be stored in the party object directly without overwriting those. And all of that seems to follow from how the API is designed, which (I think) would involve a huge rework of a lot of moving parts to smooth over.

It might be possible to store this data elsewhere, perhaps in ```store.state.user.party``` if that makes things easier. I'm not sure yet what else would have to be reworked to make that happen.

**Comment by gilmoreg** [comment]
I won't have much time to work on this starting next week, so if there are any other changes or suggestions now would be the best time. I may still be misunderstanding how this state is handled in the app (I don't see how the party object can have the details given how its defined in the store initialization). Please let me know what I can do to finish this PR!

**Comment by TheHollidayInn** [comment]
Sorry, I think the issue here is that members is not part of the party object on the server. Quest is.

So I think the solution would be to move party.members out of the party object on the Vue Store. So we would have 
```
{
 ...
 party: { ... },
 partyMembers: [...],
 ...
}
```
In the store.

Does that make sense? Sorry about this, it was definitely an issue when I implemented the members onto the party object.

**Comment by gilmoreg** [comment]
Ok I pushed some code that makes changes like this. However, I'm getting an error I don't understand:
```
[Vue warn]: Property or method "userIsOnQuest" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.

found in

---> <QuestSidebarSection> at website/client/components/groups/questSidebarSection.vue
       <Group> at website/client/components/groups/group.vue
         <App> at website/client/app.vue
           <Root>
```

```userIsOnQuest``` is a computed property, nothing seems to be wrong with it, the data that it checks for is all there (I'm really surprised I didn't have to mess with anything having to do with how the quest data was stored!). Everything I googled for on this error describes a situation in which a property is defined on a root component out of scope, which doesn't seem to be the case here.

Also no promises that I didn't miss some place where changes to the store might have unwanted side effects. This is the only warning I'm getting and the parts of the site I tested seem to work.

TheHollidayInn:
The `userIsOnQuest` issue is fixed on develop now and was separate. This looks good to me! Thanks for working on the partyMember change :D @paglias what do you think?

paglias:
Looks good to me but I haven't live tested it but we should, also https://github.com/HabitRPG/habitica/pull/9546/files#diff-cd03d18d2ec40ed535f74c334c9bf0e2L75 the party.quest object isn't needed? @gilmoreg

gilmoreg:
@paglias I checked and ```store.state.party.quest``` is never referenced, and as far as I can tell quests are working properly (they show in the sidebar, can be started and canceled, etc). There may be something I haven't thought to test though. It may be that quest data is just being retrieved straight from the API.

SabreCat:
Performed a local test and can confirm that parties load OK and that additional request does not appear. Thanks for this, @gilmoreg! I've noted it as credit toward your third contributor tier!

PR #9548, opened 2017-11-19
Added account transfer migration

TheHollidayInn:


Alys:
When you migrate pets and mounts from one account to another it's important to see if the account you're migrating to is newer than the other one and if any of the migrated pets/mounts are from special events - if they are, you also need to migrate the associated achievements and set the creation date of the newer account to be the same as the older one that you're migrating from, otherwise the user's data gets into an inconsistent state that can make it harder later to troubleshoot problems (e.g., if the user later reports that they didn't get the right Thanksgiving pet for future Thanksgiving events).

I can adjust that for the account that had the pets and mounts moved recently, but for ongoing use should we modify this script or just put a comment in it to say that it would need to be modified if used again? Or should we turn it into a more general purpose script for merging any two accounts?

TheHollidayInn:
Sounds good. I'd like to just update this script for that purpose if you don't mind.

Alys:
That's fine with me!

PR #9553, opened 2017-11-20
Fix comparisons for Guild badge / number

### LINKED ISSUE #9511: Silver tier Guilds use wrong color
Silver tier Guilds erroneously use the bronze color for their number in the crest. They should use their silver color.

### LINKED ISSUE #9512: Incorrect badge for Guilds with 100 members
Guilds with exactly 100 members seem to have a misaligned number in the Guilds list and are missing their badge, both there and in the Guild's "member list" box inside the Guild itself.

librarianmage:

Fixes #9511 
Fixes #9512

### Changes

Fixes the comparisons regarding the coloring of the guild badges and member count text.

It assumes the following breakdown for coloring:

* Bronze: less than 100 members
* Silver: less than 1,000 members, but more than 99
* Gold: 1,000+ members

----
UUID:


negue:
could you extract the two checks as a computed property so that if another fix (or different amounts are needed) we'd only need to change one method

librarianmage:
@negue Like this?

negue:
Sorry for the late response! 

You had a small syntax error `)` was missing. Once that was fixed, I had an error because `Object.freeze` wanted to create the object but `guild` is still undefined at this time. After that vue seems to be unable to find isTier or the inner value like `gold`.

I kinda forgot about the same checks in `group.vue` I think we should refactor this too.

This would be last request from me, sorry 😋 .

@MathWhiz  could you extract the `isTier`-object creation to the mixin `groupsUtilities.js` as a simple method? 

That way we could use the result in `publicGuildItem.vue` and `group.vue` (still as computed property, but without `Object.freeze`) and also inside of `groupsUtilities.js` in `filterGuild`.

I think this would be all occurrences

paglias:
hey @MathWhiz how are you doing on this?

librarianmage:
@paglias I haven't been able to push my changes yet because I haven't tested it. I will see if I have time this weekend.

lemoness:
@MathWhiz Would you like to finish this, or would you like to cede your claim?

librarianmage:
@lemoness I think I finished this one

@negue Any thoughts?

website/client/components/groups/publicGuildItem.vue
```diff
@@ -147,6 +147,9 @@ export default {
     isMember () {
       return this.isMemberOfGroup(this.user, this.guild);
     },
+    isTier (tier) {
+      return this.isTier(tier, this.guild);
```
paglias: hmm, there's already an isTier method available from the mixin, so you can remove this and just call isTier from the code, otherwise don't you risk overwriting the isTier from the mixin?

librarianmage: Maybe? I'll try that.

librarianmage:
@paglias Better?

negue:
Could you also replace the `isTier`-checks in `group.vue` ? 

And don't forget to fix the lint issues :)

website/client/mixins/groupsUtilities.js
```diff
@@ -28,6 +28,21 @@ export default {
     },
   },
   methods: {
+    isTier(tier, group) {
```
paglias, line 31: can you remove the group parameter from this function since it's not used?

website/client/mixins/groupsUtilities.js
```diff
@@ -28,6 +28,21 @@ export default {
...
+      switch (tier) {
+        case "gold":
+          return guild.memberCount >= 1000;
+          break;
+        case "silver":
+          return guild.memberCount < 1000 && guild.memberCount >= 100;
+          break;
+        case "bronze":
+          return guild.memberCount < 100;
```
paglias, line 40: where is the guild variable coming from? should it be this.guild?

paglias (commented):
there are two issues, otherwise it looks good

lemoness:
@MathWhiz - Hi there! How is this fix going?

librarianmage:
I have not had much time to work on it, unfortunately :(

PR #9565, opened 2017-11-21
Allow users to Batch Buy Quests

### LINKED ISSUE #9507: Errors with batch-buying gold purchasable quests
Batch-buying gold purchasable quests seems to result in errors where the number the user attempted to purchase don’t show up in inventory, but gold is sometimes deducted correctly for the amount they attempted to purchase anyway.

kipraske:

Fixes https://github.com/HabitRPG/habitica/issues/9507

### Changes

You can buy multiple quests from the buy modal, but the quantity is never passed to the server. So the client thinks that you are buying the number you said you did and deducts gold for everything, but the server only buys one regardless of how many wanted to buy. This can lead to syncing problems down the road.

Unfortunately, I never was able to create the sync problems that were described in the issue: where you would be permanently deducted gold for quests that you didn't receive, but hopefully if we don't cause the discrepancy in the first place, we won't give it a chance to readjust back to the server values incorrectly.

Since the server and the client use the same code for adjusting the user values, all we needed was that quantity variable to be defined on the server and everything else just fell into place.

----
UUID: f5d12242-5b66-471d-bfc2-6c3358861d9c

### DIFF
--- a/website/client/store/actions/shops.js
+++ b/website/client/store/actions/shops.js
@@ -33,7 +33,7 @@ export function buyQuestItem (store, params) {
 
   return {
     result: opResult,
-    httpCall: axios.post(`/api/v3/user/buy/${params.key}`, {type: 'quest'}),
+    httpCall: axios.post(`/api/v3/user/buy/${params.key}`, {type: 'quest', quantity}),
   };
 }
 


paglias:
Thanks, the PR works well, it'll be merged next week :)

SabreCat:
Thank you @kipraske, and welcome to the ranks of the Blacksmiths! Note that contributor tiers require increasing amounts of effort from one to the next, so tier 2 may require multiple PRs or a larger PR to attain. But keep helping out and we'll express our gratitude accordingly!

PR #9569, opened 2017-11-22
Ensured user is saved after validation checks

TheHollidayInn:


Alys:
I'm worried this might create a bug for the users. Sometimes quest invitations persist in a user's account after the quest has started (due to other bugs; it's not meant to happen but it does), and the user will keep seeing the invitation for as long as `user.party.quest.RSVPNeeded` is true. With this change, the user will never be able to set RSVPNeeded to false because the `throw` lines will stop the `save()` occuring won't they?

TheHollidayInn:
That may be true, but this will let us see the other bugs and fix them. This change ensures the group is updated at the same time the user is, which we want - although the group save could still error. But I haven’t seen any in Loggly

PR #9575, opened 2017-11-24
Changed search to regex

### LINKED ISSUE #9502: Must enter full Guild name for search results
Currently, searching for just a partial word in the Guild search bar does not bring up the Guild in questions. For example, searching for "wolf" still does not bring up Wolfsrudel, but if you enter "Wolfsrudel" in its entirety you see the Guild in results. This should also be the case for other search bars.

TheHollidayInn:
fixes #9502

### DIFF
--- a/website/server/controllers/api-v3/groups.js
+++ b/website/server/controllers/api-v3/groups.js
@@ -336,7 +336,11 @@ api.getGroups = {
     }
 
     if (req.query.search) {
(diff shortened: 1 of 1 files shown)


paglias (commented):
What's filters.$or for?

TheHollidayInn:
To the $or is so we can search the keywords in the name or the description.

PR #9579, opened 2017-11-24
Set default empty array when party members haven't loaded

### LINKED ISSUE #9396: Details button for Pending Quests doesn't always work
Sometimes the Details button on a pending quest doesn’t do anything when you click it. Reloading doesn’t fix it, but switching to a different browser does. This error was in the console: "TypeError: null is not an object (evaluating 'this.partyMembers.map')"

TheHollidayInn:
Fixes #9396

### DIFF
--- a/website/client/components/groups/questDetailsModal.vue
+++ b/website/client/components/groups/questDetailsModal.vue
@@ -180,6 +180,7 @@ export default {
       return quests.quests[this.group.quest.key];
     },
     members () {
+      let party = this.partyMembers || [];
       return this.partyMembers.map(member => {
         return {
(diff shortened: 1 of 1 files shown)


website/client/components/groups/questDetailsModal.vue
```diff
@@ -180,6 +180,7 @@ export default {
       return quests.quests[this.group.quest.key];
     },
     members () {
+      let party = this.partyMembers || [];
```
paglias: I think this should be `let partyMembers = this.partyMembers || [];`and then use `partyMembers` instead of `this.partyMembers` below

PR #9583, opened 2017-11-26
Bug fixes on Subscription termination date format [Fixes Issues #9186]

tim1234ltp:
Fixes https://github.com/HabitRPG/habitica/issues/9186

### Changes
Created a computed value of Subscription termination date which is formatted according to user preference.
The client will show this computed value instead of using a filter on raw value.

### Result (sorry for the non-english strings)
default

### DIFF
--- a/website/client/components/settings/subscription.vue
+++ b/website/client/components/settings/subscription.vue
@@ -232,7 +232,7 @@ export default {
     },
     dateTerminated () {
       if (!this.user.preferences || !this.user.preferences.dateFormat) return this.user.purchased.plan.dateTerminated;
-      return this.user.purchased.plan.dateTerminated.format(this.user.preferences.dateFormat.toUpperCase());
+      return moemnt(this.user.purchased.plan.dateTerminated).format(this.user.preferences.dateFormat.toUpperCase());
     },
     canCancelSubscription () {
       return (
--- a/website/client/components/settings/subscription.vue
+++ b/website/client/components/settings/subscription.vue
@@ -33,7 +33,7 @@
             span.noninteractive-button.btn-danger {{ $t('canceledSubscription') }}
             i.glyphicon.glyphicon-time
             |  {{ $t('subCanceled') }} &nbsp;
-            strong {{user.purchased.plan.dateTerminated | date}}
+            strong {{dateTerminated}}
           tr(v-if='!hasCanceledSubscription'): td
             h4 {{ $t('subscribed') }}
             p(v-if='hasPlan && !hasGroupPlan') {{ $t('purchasedPlanId', purchasedPlanIdInfo) }}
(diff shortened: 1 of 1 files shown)


website/client/components/settings/subscription.vue
```diff
@@ -237,6 +237,10 @@ export default {
         amount: this.numberOfMysticHourglasses,
       };
     },
+    dateTerminated () {
+      if (!this.user.preferences || !this.user.preferences.dateFormat) return moment(this.user.purchased.plan.dateTerminated);
+      return this.user.purchased.plan.dateTerminated.format(this.user.preferences.dateFormat.toUpperCase());
```
paglias: `this.user.purchased.plan.dateTerminated` shouldn't be `moemnt(this.user.purchased.plan.dateTerminated)`?

website/client/components/settings/subscription.vue
```diff
@@ -33,7 +33,7 @@
             span.noninteractive-button.btn-danger {{ $t('canceledSubscription') }}
             i.glyphicon.glyphicon-time
             |  {{ $t('subCanceled') }} &nbsp;
-            strong {{user.purchased.plan.dateTerminated | date}}
```
paglias, line 36: can you remove the code for the code for the date filter below in this file since it's not used anymore? https://github.com/HabitRPG/habitica/pull/9583/files#diff-c55b95935ec2f85da360edbe1360852dL146

paglias (changes requested):

website/client/components/settings/subscription.vue
paglias: I think there's still an issue, line 242 should be `return moment(this.user.purchased.plan.dateTerminated).format(this.user.preferences.dateFormat.toUpperCase());`, otherwise you'll be calling `.format` on a string, which doesn't work

tim1234ltp: You are right. Only line 241 should return normal string object?

paglias: yes!

website/client/components/settings/subscription.vue
```diff
@@ -237,6 +230,10 @@ export default {
         amount: this.numberOfMysticHourglasses,
       };
     },
+    dateTerminated () {
+      if (!this.user.preferences || !this.user.preferences.dateFormat) return this.user.purchased.plan.dateTerminated;
+      return moemnt(this.user.purchased.plan.dateTerminated).format(this.user.preferences.dateFormat.toUpperCase());
```
paglias: `moemnt` should be `moment` then I think we'll be good to go!

tim1234ltp:
@paglias The Travis CI check cannot be passed just like last time.
  1 failing
  1) GET /groups/:groupId/members supports using req.query.lastId to get more members:

Perhaps, some codes regarding group features are still not settled?

paglias:
@tim1234ltp don't worry about that error, it happens sometimes but it's not related to your PR :)

paglias (approved):

paglias:
@tim1234ltp  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #9592, opened 2017-11-28
Fixing the pig flying too high in the stable square bug

### LINKED ISSUE #9364: Flying Pig soars too high, must have its glory contained
When you tame a flying pig pet, the greyed out version hovers about a half-centimeter too high, so that they're actually flying out of the top of the box. This should be adjusted.

screen shot 2017-11-01 at 6 16 25 pm

kipraske:

Fixes https://github.com/HabitRPG/habitica/issues/9364

### Changes

Because the flying pig sprite is offset upwards a bit in the sprite sheet (so it can be "flying"), there is some CSS put in `website/client/components/inventory/stable/index.vue` to manually adjust the flying pig:

``` css
  .stable .item .item-content.Pet:not(.FlyingPig) {
    top: -28px;
  }

  .stable .item .item-content.FlyingPig {
    top: 7px;
  }
```

Well, the issue is that when you grey out the pets it doesn't include the type of pet anymore as a class and so the `FlyingPig` class is lost. Thus the flying pig doesn't get its special treatment from this CSS when it is greyed out, and it ends up going upwards to it's natural flying pig state.

So this pull request is just me adding the name of the pet back into the class when it is greyed out as well. This name is stored in `${pet.eggKey}`. Now, it is important to note then that we add the name of the pet class to all of the pets, so there is a potential to change the appearance of the other pets. However, I only found custom CSS to adjust offsets the Flying Pig and the Hydra (which has a large sprite). And we shouldn't have to worry about the Hydra because it never gets greyed out. So I am pretty confident that we will be fine. Feel free to check more thoroughly than I did for offsets in the quest pets or magic potion pets. I didn't see any, but I didn't check them all.

----
UUID: f5d12242-5b66-471d-bfc2-6c3358861d9c

### DIFF
--- a/website/client/components/inventory/stable/index.vue
+++ b/website/client/components/inventory/stable/index.vue
@@ -867,7 +867,7 @@
         }
 
         if (pet.mountOwned()) {
-          return `GreyedOut Pet Pet-${pet.key}`;
+          return `GreyedOut Pet Pet-${pet.key} ${pet.eggKey}`;
         }
 
         if (pet.isHatchable()) {


SabreCat:
Thanks much, @kipraske! I've noted this toward your second Blacksmith contributor tier!

PR #9600, opened 2017-11-29
Removed the from az sort

### LINKED ISSUE #9361: Quest Titles should not include "The" to accomodate A-Z sort and translations
~In the Quests Shop, sorting from A-Z includes "The" as a criterion. Sorting should be tweaked to ignore "the" when it's the first word of a Quest name.~

Edited: The simplest solution for this issue would be to remove "the" from all quest titles. This preserves the functionality of the A-Z sort as well as our ability to allow for translations.

TheHollidayInn:
Fixes #9361

### DIFF
--- a/website/client/components/shops/quests/index.vue
+++ b/website/client/components/shops/quests/index.vue
@@ -432,8 +432,10 @@ export default {
 
         switch (sortBy) {
           case 'AZ': {
-            result = _sortBy(result, ['text']);
-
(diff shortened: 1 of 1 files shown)


negue:
This would only remove the english stopword. Maybe we should, like you mentioned, add the `stopwords` package for this.

lemoness:
@TheHollidayInn What's the status on this?

PR #9606, opened 2017-11-29
add subscription cancellation instructions for free Group Plans subscriptions

### LINKED ISSUE #8874: cancel subscription message is incorrect for free subscriptions from group plans
When you have a free subscription from a group plan, the Settings > Subscription page displays this error message just below the blue "Subscribed" box:
`String 'cancelSubInfoGroup Plan' not found.`

It should show instead show a message explaining that the owner of the Group Plan can cancel the Group Plan by going to the guild or party that contains the plan, clicking on the "Payment Details" tab, and then clicking on the "cancel" button. We fairly often see questions from plan owners about how to cancel and they're often looking in the Subscription page for that option, so having the message there would be helpful.

~@lemoness This will require website changes. Do you want this fixed before the website improvement project is finished or should it wait?~ EDIT: this is ready to be worked on

Thanks to DudleyPippin (dbe40e7b-e9b6-48b1-8fc4-124ab64239e2) for reporting this in the Report a Bug guild!

Alys:
fixes https://github.com/HabitRPG/habitica/issues/8874

When you have a free subscription from a Group Plan, the cancellation section in the Subscription page currently has a missing string error. This PR inserts a message instead.

Current website with error at the bottom of this screenshot:

This PR:

The new wording is:
_Because you have a free subscription from a Group Plan, you cannot cancel it. It will end when you are no longer in the Group. If you are the Group leader and want to cancel the entire Group Plan, you can do that from the group's "Payment Details" tab._

This PR also fixes a related bug that doesn't have an issue (I've only just noticed it). If you have a Stripe, Amazon, or PayPal subscription which has already been cancelled, the current website is showing a missing string error (at the bottom of the screenshot below). This PR prevents the error. There's not meant to be any text there for this situation.


SabreCat:
All checks out!

PR #9617, opened 2017-11-30
Rewards permit decimal value. Fixes https://github.com/HabitRPG/habitica/issues/9513

### LINKED ISSUE #9513: Rewards should be allowed to have decimal values
It should be possible to set a Reward's value to have decimal points. E.g., it's allowed to create a Reward that costs 1.75 gold.

However when you try to do that on the new website, it doesn't allow decimal points and tells you that you have to choose between nearest values of 1 and 2. That restriction should be removed.

Feywood:

Fixes https://github.com/HabitRPG/habitica/issues/9513

### Changes
taskModal.vue modified so that where task.type === reward, the input type for cost is now type number.toFixed(2) to permit/force two decimal places.
Tested in chrome/firefox. Do want to add that my npm test keeps complaining of crlf line endings despite modification to autocrlf true. Apologies if this causes any issues.

----
UUID: 9c4776eb-7330-465b-85cc-e4ccef6c1ad1

### DIFF
--- a/website/client/components/groups/group.vue
+++ b/website/client/components/groups/group.vue
@@ -30,7 +30,7 @@
         h3(v-once) {{ $t('chat') }}
 
         .row.new-message-row
-          textarea(:placeholder="!isParty ? $t('chatPlaceholder') : $t('partyChatPlaceholder')", v-model='newMessage', @keydown='updateCarretPosition')
+          textarea(:placeholder="!isParty ? $t('chatPlaceholder') : $t('partyChatPlaceholder')", v-model='newMessage', @keydown='updateCarretPosition', @keyup.ctrl.enter='sendMessage()')
           autocomplete(:text='newMessage', v-on:select="selectedAutocomplete", :coords='coords', :chat='group.chat')
 
         .row
@@ -38,7 +38,7 @@
             button.btn.btn-secondary.float-left.fetch(v-once, @click='fetchRecentMessages()') {{ $t('fetchRecentMessages') }}
             button.btn.btn-secondary.float-left(v-once, @click='reverseChat()') {{ $t('reverseChat') }}
           .col-6
-            button.btn.btn-secondary.send-chat.float-right(v-once, @click='sendMessage()', @keyup.ctrl.enter='sendMessage()') {{ $t('send') }}
+            button.btn.btn-secondary.send-chat.float-right(v-once, @click='sendMessage()') {{ $t('send') }}
 
         .row.community-guidelines(v-if='!communityGuidelinesAccepted')
           div.col-8(v-once, v-html="$t('communityGuidelinesIntro')")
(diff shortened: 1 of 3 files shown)


Alys:
Thanks for the warning about the line endings but there's no problem with them. You'd notice if there were: every line in the file would be marked as changed. :)

negue:
I'm on Mac, tested on Vivaldi and Firefox, but both just let me enter characters or more decimal digits.

The change to `number.toFixed(2)` kinda acts just like `text` (as fallback).

Can anyone confirm that it works on a different browser / OS ?

Feywood:
@negue just retested and you're right, Windows/Chrome gives the same issue with toFixed allowing text (which then gives a validation error) and more decimals. Feeling both baffled and annoyed.

TheHollidayInn:
I think it would be better to add the fix to the value rather than on the html text attribute.

Feywood:
I think this is where my inexperience outs itself. I have found many instances of task value, but mostly it seems to apply to redness of dailies or to-dos. I can't seem to pinpoint where, in case of a reward, the value is actually defined. I suppose this pr can be dropped.

TheHollidayInn:
You are on the right track! The task.value is domain knowledge, so don't worry about not knowing. task. 

Task.value comes directly from the task property : https://github.com/Feywood/habitica/blob/a2ebcc6373a1b0ec7ac1845432ae69445c716f77/website/client/components/tasks/taskModal.vue#L550

Actually, it turns out I've misread the issue. Looking at the docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number. You might want to try using `step` `<input type="number" placeholder="1.0" step="0.01" min="0" max="10">`

I think that may have better support that the `toFixed` in the attribute.

Alys:
@Feywood These sorts of hiccups are common for new contributors! We've all been here. :)

This probably isn't needed for solving this issue, but since you mentioned the value:
Rewards are actually defined in much the same ways as Habits, Dailies, and To-Dos. All four types are considered "tasks". They're all stored in Habitica's database in the same collection, but with slightly different sets of fields for each type (for example, all four have "text", "note", and "value" attributes, but only Habits, Dailies, and To-Dos have an attribute for difficulty).
So `task.value` is the same attribute for all four task types. For Habits, Dailies, and To-Dos, it does control colour (see [Task Value](http://habitica.wikia.com/wiki/Task_Value) for details). For Rewards, it's used as the price. The `website/server/models/task.js` file defines task attributes, and goes in to a lot more detail than is relevant here, but glancing through it will give you some idea about how the same structures are used for all four task types.

Feywood:
Thanks all for the help. Put the step and placeholder in and that seems to have done it. Two decimals, more is not accepted and no more text can be entered.

SabreCat:
There are a few alignment issues with the coins etc. that this makes easier to run into, but not caused by the PR. Looks good here!

Thanks for the contribution, @Feywood! I've added "Blacksmith" to your contributor titles, and noted this as credit toward attaining tier 2 contributor status.

PR #9621, opened 2017-12-01
Fix install node permissions

luanmuniz:
Signed-off-by: Luan <luan@luanmuniz.com.br>

### Changes
This problem was happening:
screen shot 2017-12-01 at 09 09 37

Just fixed the permissions on the install_node.sh file since it was breaking for the vagrant setup.

----
UUID: 0d8208a5-9485-4a2f-9be9-009d9ed5679f


Alys:
Thanks @luanmuniz!  I've tested this and it gives the `install_node.sh` file the same permissions as the other scripts in its directory, as expected.

Normally we merge PRs only once a week as part of our testing and review process but since this is a simple change that doesn't affect the website or API and will be of immediate use to Blacksmiths, I'm merging it now.

I've noted it as credit towards Tier 2.

PR #9627, opened 2017-12-01
Cron age habits todos sleeping

### LINKED ISSUE #5232: Allow Habits and To-Dos to "age" while Resting in the Inn
When you are **not** Resting in the Inn, each time cron runs, it makes uncompleted To-Dos more red and it adjust plus-only and minus-only Habits to move towards yellow.

We would like the same thing happen when you are Resting. 

Ref: https://github.com/HabitRPG/habitrpg/pull/5073#issuecomment-97277606 with approval from Lemoness at the bottom of that issue.

_edit by Alys 2017-10-22:_
When this is implemented, the wording on the banner from https://github.com/HabitRPG/habitica/issues/9188 will need to be changed. Either change it on the website as part of the fix for this issue if https://github.com/HabitRPG/habitica/issues/9188 has already been completed, or put a comment in https://github.com/HabitRPG/habitica/issues/9188 to say that the wording needs to change.

_edit by Alys 2018-08-02:_
Habit counters should also reset when resting.

TheHollidayInn:
Fixes #5232


Alys:
I have promised to test this in the next day or two.

test/api/v3/unit/libs/cron/cron.habits.test.js
```diff
@@ -0,0 +1,308 @@
...
+  let tasksByType = {habits: [], dailys: [], todos: [], rewards: []};
+  let daysMissed = 0;
+
+  beforeEach(() => {
+    user = new User({
+      auth: {
+        local: {
+          username: 'username',
+          lowerCaseUsername: 'username',
+          email: 'email@email.email',
```
Alys: This is pretty minor and picky so it's just a suggestion: perhaps we should use example.com rather than email.email. ".email" is a gTLD so it's not impossible that that email address would actually exist and the owner might be peevish about it appearing in github.

website/server/libs/cron.js
```diff
@@ -266,6 +293,7 @@ export function cron (options = {}) {
   // On cron, buffs are cleared and all dailies are reset without performing damage
   if (user.preferences.sleep === true) {
     performSleepTasks(user, tasksByType, now, daysMissed);
+    ageTodos(user, tasksByType, true, daysMissed, 0);
```
Alys: Should this call to `ageTodos` be moved into `performSleepTasks` for consistency? Maybe also move `trackCronAnalytics` (from the line just below).

Although see also my point below about how performSleepTasks troubles me. :)

I think the `true` value in the call to `ageTodos` should be replaced by `multiDaysCountAsOneDay` in case we make that a user preference setting later.

website/server/libs/cron.js
```diff
@@ -277,17 +305,7 @@ export function cron (options = {}) {
...
-      task,
-      user,
-      direction: 'down',
-      cron: true,
-      times: multiDaysCountAsOneDay ? 1 : daysMissed,
-    });
-
-    todoTally += task.value;
-  });
+  todoTally = ageTodos(user, tasksByType, multiDaysCountAsOneDay, daysMissed, todoTally);
```
Alys, line 308: I think this `todoTally` variable and the code below that uses it is also important to do while the user is sleeping. Since we're changing task values now while sleeping, we'd want the tallied history data to be adjusted too.

That would mean copying yet another thing into the `performSleepTasks` function.  There's also a bunch of other things that we really should be doing when the user is sleeping, such as tallying experience (`expTally`), the `tasksOrder` code around line 418 (there'll still be tasksOrder bugs for any user who's sleeping!), `preenUserHistory`, and probably almost everything else except actually doing damage from Dailies.

Can we consider getting rid of `performSleepTasks` and just putting an `if !sleeping` condition around the few pieces of code that should not be executed while sleeping? Although I realise this is getting away from the point of the issue!

If we keep using `performSleepTasks` for now, this comment in website/server/libs/cron.js at line 426 will need to be changed in this PR:
_// TODO also do while resting in the inn. Note that later we'll be allowing the value/color of tasks to change while sleeping (https://github.com/HabitRPG/habitica/issues/5232), so the code in performSleepTasks() might be best merged back into here for that. Perhaps wait until then to do preen history for sleeping users._

It could be changed to something like this:
_// TODO also run preenUserHistory while resting in the inn._

website/server/libs/cron.js
TheHollidayInn: Eh, maybe not. I think `performSleepTasks` has too many responsibilities right now. All the arguments are off. I'm slowly making some refactoring as you can see which will allow for better argument passing.

TheHollidayInn: I have moved up the scope of `multiDaysCountAsOneDay` and passed it!

website/server/libs/cron.js
TheHollidayInn: Okay :P How about we map out some Cron flow offline and I'll just do some refactoring while allow these to happen. There are a few more conditions here, and I'd like to prevent adding hacking flow logic.

lemoness:
@TheHollidayInn What's the status on this?

PR #9643, opened 2017-12-03
Added client side logging

### LINKED ISSUE #6216: Automatic logging of client side errors
This app doesn't currently seem to have any sort `onerror` hook to capture and send to us browser errors that might be happening.

Adding a tool such as [Rollbar](https://rollbar.com) or [Raygun](https://raygun.io/javascript) could help us solve this problem and fix bugs pre-emptively.

Worst case scenario, we should at least add a manual `onerror` event handler ourselves and then use it to log the error to LocalStorage so if they want to report a bug, then the UI can show them exactly what they should paste in the description of the Github issue.

--------

NOTE from @paglias: if possible let's use Loggly for this

TheHollidayInn:
Fixes #6216


website/client/index.html
```diff
@@ -20,6 +20,7 @@
     <div id="app"></div>
     <!-- built files will be auto injected -->
 
+    <script type="text/javascript" src="//cloudfront.loggly.com/js/loggly.tracker-latest.min.js" async></script>
```
paglias, line 23: isn't this available as a npm package?

paglias (changes requested):

TheHollidayInn:
Yea.. Kind of. This library is for the front end. They have this: https://www.npmjs.com/package/loggly. But that is meant to be used with Winston which doesn't work on client side because of the fs module. 

Their docs says to use this on the front end (although it could just be old school). But the above allows to catch console errors globally.

TheHollidayInn:
I'm going to merge this for now, but if you think the npm package will work, let me know and I'll update it.

Issue #9265, opened 2017-10-25
Both eggs and potions should hatch pets from Items


lemoness:
When hatching pets from the Items page instead of using the new rapid hatch mode in the Stable, there is a bug. When you click on a hatching potion, it correctly highlights all the eggs that you can use it on and lets you click on an egg to hatch a pet. However, if you click on an egg, nothing happens. The same behavior should occur if you click on an egg: it should highlight all the possible potion combinations and let you click on one to hatch it.

petmat:
Hi! I would like to take a look at this if it's ok.

lemoness:
@petmat Thank you so much! Feel free to get started, and let us know if you have any questions.

I'm updating the labels on this issue.

Alys:
@petmat How are you going with this? Do you have any questions for us or need any advice? Or if real-life has got in the way and you no longer have any free time, just let us know! We understand what that's like. :)

petmat:
Sorry about the delay, yeah things have been crazy busy lately, but I was thinking about taking another look at this on the weekend if that's ok.

Alys:
@petmat Sure! Thanks! :)

PR #9645, opened 2017-12-03
Both eggs and potions should hatch pets from Items

petmat:
Fixes https://github.com/HabitRPG/habitica/issues/9265

### Changes
Makes it possible to hatch a pet also by clicking on an egg and then choosing the hatching potion. Previously it was only possible to hatch a pet by first clicking on the potion and the on the egg.

----
UUID: 70a004a6-3020-4907-9705-c6dddaa8aaf2

### DIFF
--- a/website/client/components/inventory/items/index.vue
+++ b/website/client/components/inventory/items/index.vue
@@ -48,6 +48,8 @@
             :item="context.item",
             :key="context.item.key",
             :itemContentClass="context.item.class",
+            :showPopover="currentDraggingEgg == null",
+            :active="currentDraggingEgg == context.item",
             :highlightBorder="isHatchable(currentDraggingPotion, context.item.key)",
             v-drag.drop.hatch="context.item.key",
 
@@ -82,6 +84,7 @@
             :itemContentClass="context.item.class",
             :showPopover="currentDraggingPotion == null",
             :active="currentDraggingPotion == context.item",
+            :highlightBorder="isHatchable(context.item.key, currentDraggingEgg)",
             v-drag.hatch="context.item.key",
 
             @itemDragEnd="onDragEnd($event, context.item)",
--- a/website/client/components/inventory/items/index.vue
+++ b/website/client/components/inventory/items/index.vue
@@ -54,6 +54,7 @@
             @itemDragOver="onDragOver($event, context.item)",
             @itemDropped="onDrop($event, context.item)",
             @itemDragLeave="onDragLeave()",
+            @itemDragStart="onDragStart($event, context.item)",
 
             @click="onEggClicked($event, context.item)",
           )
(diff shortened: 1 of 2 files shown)


website/client/components/inventory/items/index.vue
```diff
@@ -48,12 +48,15 @@
...
             :itemContentClass="context.item.class",
+            :showPopover="currentDraggingEgg == null",
+            :active="currentDraggingEgg == context.item",
             :highlightBorder="isHatchable(currentDraggingPotion, context.item.key)",
             v-drag.drop.hatch="context.item.key",
 
             @itemDragOver="onDragOver($event, context.item)",
             @itemDropped="onDrop($event, context.item)",
             @itemDragLeave="onDragLeave()",
+            @itemDragStart="onDragStart($event, context.item)",
```
negue: the real dragging feature is currently disabled, so you currently can't test the `onDragStart`.

I'd suggest to you remove this property and leave a comment in `onDragStart` that it needs to be added for the egg-items  -- or, if you want -- change the `onDragStart` /  `OnDrop` / `OnDragOver` /  `OnDragLeave` methods since those are currently dragging-potion specific

website/client/components/inventory/items/index.vue
```diff
@@ -81,6 +84,7 @@
             :itemContentClass="context.item.class",
             :showPopover="currentDraggingPotion == null",
             :active="currentDraggingPotion == context.item",
+            :highlightBorder="isHatchable(context.item.key, currentDraggingEgg)",
```
negue: `isHatchable` currently takes `potion` (object) and `eggKey` (string) as parameter, I'd suggest to change both to accept an object that way we don't need to pass `context.item.key` and can just use the object itself --- in `isHatchable` then needs to be another check if `egg` isn't null

website/client/components/inventory/items/index.vue
```diff
@@ -129,6 +133,18 @@
 
   hatchedPetDialog()
 
+  div.eggInfo(ref="draggingEggInfo")
+    div(v-if="currentDraggingEgg != null")
+      div.potion-icon(:class="'Pet_Egg_'+currentDraggingEgg.key")
+      div.popover
+        div.popover-content {{ $t('dragThisEgg', {eggName: currentDraggingEgg.text }) }}
```
negue, line 139: could you add to `.popover-content` a `margin: 15px;` its not your fault just would be nice if this is in the same PR :)  

cc @Tressley currently the "dragging" text is completely at the border, so I just think 15px looks good, what do you think ?

negue (changes requested):
Thanks for the PR 😃 , Code is LGTM

Just a few changes then its good to go.

petmat:
I made the changes you requested, but for some reason a test named "checks error message has the banned words used" is failing now. I don't quite understand why the test failed since it doesn't seem related to the changes. Can you help me with what I should do to make the test succeed?

negue (approved):

negue:
@petmat sorry for the wait! sometimes travis tests produce errors even if everything is ok

@paglias could we restart this test for the PR?

TheHollidayInn:
Restarted

website/client/components/inventory/items/index.vue
Tressley: @negue -- Can you attach a screenshot for reference? 👍

negue: @Tressley  
egg-hatching-drag

SabreCat:
Thank you for this, @petmat, and welcome to the ranks of the Blacksmiths at contributor tier 1! Note that further tiers require increasing amounts of effort from one to the next, so tier 2 may require multiple PRs or a larger PR to attain. But keep helping out and we'll express our gratitude accordingly!

website/client/components/inventory/items/index.vue
Tressley: Oh, wow. I'm just now seeing this @negue! Sorry for the delay! I think there is some style work needed before merging this. See attached screen below for better direction on positioning and placement.

image

negue: will fix this in a separate PR

PR #9653, opened 2017-12-04
Ensured admin can always PM user

### LINKED ISSUE #8142: allow moderators to send a private message (PM) to any user - ignore opt-out / blocking

TheHollidayInn:
Fixes #8142

### DIFF
--- a/test/api/v3/integration/members/POST-send_private_message.test.js
+++ b/test/api/v3/integration/members/POST-send_private_message.test.js
@@ -118,4 +118,56 @@ describe('POST /members/send-private-message', () => {
     expect(sendersMessageInReceiversInbox).to.exist;
     expect(sendersMessageInSendersInbox).to.exist;
   });
+
+  it('allows admin to send when sender has blocked to user', async () => {
+    const userToSendMessage = await generateUser({
+      'backer.admin': 1,
+    });
+    const receiver = await generateUser({'inbox.blocks': [userToSendMessage._id]});
+
+    await userToSendMessage.post('/members/send-private-message', {
+      message: messageToSend,
+      toUserId: receiver._id,
+    });
+
+    const updatedReceiver = await receiver.get('/user');
+    const updatedSender = await userToSendMessage.get('/user');
+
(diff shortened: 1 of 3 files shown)


test/api/v3/integration/members/POST-send_private_message.test.js
```diff
@@ -118,4 +118,56 @@ describe('POST /members/send-private-message', () => {
     expect(sendersMessageInReceiversInbox).to.exist;
     expect(sendersMessageInSendersInbox).to.exist;
   });
+
+  it('allows admin to send when sender has blocked to user', async () => {
+    userToSendMessage = await generateUser({
+      'backer.admin': 1,
```
Alys: `contributor.admin` rather than `backer.admin` :) Also in other places in the tests and code.

test/api/v3/integration/members/POST-send_private_message.test.js
```diff
@@ -118,4 +118,56 @@ describe('POST /members/send-private-message', () => {
     expect(sendersMessageInReceiversInbox).to.exist;
     expect(sendersMessageInSendersInbox).to.exist;
   });
+
+  it('allows admin to send when sender has blocked to user', async () => {
```
Alys: should this be "the user" rather than "to user"?

PR #9657, opened 2017-12-04
Added float rounding

### LINKED ISSUE #9275: Round task difficulty value
The iOS app currently has an issue where it changes the trivial difficulty (value of 0.1) to a string representation of "0.10000000000005" (or something similar). This causes the api to reject the value, because it is not a valid difficulty. Unfortunately fixing this in the app would be non-trivial, so it would probably be easier to change the validation on the server side to round the task difficulty to one digit after the comma before checking if it is a valid difficulty value.

TheHollidayInn:
Fixes #9275

### DIFF
--- a/website/server/libs/taskManager.js
+++ b/website/server/libs/taskManager.js
@@ -87,8 +87,8 @@ export async function createTasks (req, res, options = {}) {
     let newTask = new Tasks[taskType](Tasks.Task.sanitize(taskData));
 
     // Attempt to round priority
-    if (newTask.priority && Number.parseFloat(newTask.priority) !== NaN) {
-      newTask.priority = Number((newTask.priority).toFixed(1));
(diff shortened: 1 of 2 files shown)


website/server/libs/taskManager.js
```diff
@@ -86,6 +86,11 @@ export async function createTasks (req, res, options = {}) {
     let taskType = taskData.type;
     let newTask = new Tasks[taskType](Tasks.Task.sanitize(taskData));
 
+    // Attempt to round priority
+    if (newTask.priority && Number.parseFloat(newTask.priority) !== NaN) {
```
paglias: I don't think you can compare NaN with `===`, it looks like you have to use `Number.isNaN` https://developer.mozilla.org/it/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

paglias (changes requested):

PR #9663, opened 2017-12-05
Don't allow Saddles to be sold (fixes #9596)

### LINKED ISSUE #9596: Saddles Should Not Be Sellable
**EDIT 2018-12-08 by Alys: For the current behaviour and new fix required, see my comment further down in the isssue. This should be a pretty easy fix since it likely just needs small changes to existing code, based on code that was working in the past (links to both sets of code are in my comment) so I'm marking this as "good first issue" although it might be a little harder than such issues normally are.**

### Description
Currently the UI shows that you can sell Saddles in the Market if they are in your inventory. We should not allow this

from @Tressley "If they try to sell a Saddle, we could prompt them with a "Hey! This is a pretty useful item! Are you familiar with how to use a Saddle with your Pets?"

This is also an alternative solution.

matchajune:
[x]: Fixes https://github.com/HabitRPG/habitica/issues/9596

### Changes
[x]: Disallow selling of Saddle with conditional rendering with `sellWarningNote` property on Saddle object.

----

### DIFF
--- a/website/client/components/shops/market/sellModal.vue
+++ b/website/client/components/shops/market/sellModal.vue
@@ -7,9 +7,16 @@
     div.close
       span.svg-icon.inline.icon-10(aria-hidden="true", v-html="icons.close", @click="hideDialog()")
 
-    div.content(v-if="item != null")
+    div.content(v-if="item")
 
-      div.inner-content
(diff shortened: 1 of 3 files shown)


SabreCat:
And that's tier 2! Thanks for all the PRs, @matchajune!

Issue #9629, opened 2017-12-01
Challenge Winner Text is misleading


lemoness:
The Challenge winner declaration text implies that you can select more than one winner, which is not the case. It should say “Select a winner from the Challenge participants” and “Award Winner.” (Also, there should be some space between the winner’s name and “Selected”.)

avirgil:
I can take a look at this, but it would be my first issue. :)

TheHollidayInn:
Go for it :D

avirgil:
Ok so I've fixed the first two issues in en and en_GB but how are text updates for other languages handled?

Also I can't for the life of me figure out where the winner's name and "Selected" appear. @lemoness can you clarify?

Alys:
Thanks @avirgil !

Only the /en/ language should be edited in the repository. Other languages are handled in Transifex - see the _README_FIRST.md file in the locales directories for details. Can you reverse the changes to the en_GB files before creating a pull request? You can do that with `git checkout <path/to/file.json>`

avirgil:
Will do, thanks. Does anyone have insight on the `Name "Selected"` part?

lemoness:
The spacing question is something for @Tressley to answer!

Alys:
The "selected" spacing part is no longer relevant so I'm closing this.

PR #9664, opened 2017-12-05
Fixed Challenge Winner Text (partial fix for #9629)

avirgil:

partial fix for #9629 

### Changes
Changed challenge winner text. There is an issue noted in the ticket: 

> (Also, there should be some space between the winner’s name and “Selected”.)

That I couldn't find in the code. That part has not been fixed.

----
UUID: c4192f1b-fa23-43aa-b278-1f6b2618949d


avirgil:
Ugh. Text should read : "select A winner." I'll commit a fix.

avirgil:
Not sure why the build failed. Looks like an issue in the build itself? The last build failed with the message: 

> The command "npm install &> npm.install.log || (cat npm.install.log; false)" failed and exited with 1 during .

But the same command in the previous jobs ran fine.

Alys:
@avirgil Sometimes the tests time out or fail for other reasons unconnected to the PR's code changes. That message you quoted is a typical example of a transient failure. I restarted that test and it's passing now. :) Posting to the PR when you notice an unexplained failure is the right thing to do.

avirgil:
@Alys Does it matter that coverage went down?

TheHollidayInn:
The coverage is not a problem for this code. It was just an issue with Travis.

I don't think the space between select and the user is an issue anymore. This PR looks good!

SabreCat:
Thank you for this, @avirgil, and welcome to the ranks of the Blacksmiths at contributor tier 1! Note that further tiers require increasing amounts of effort from one to the next, so Tier 2 may require multiple PRs or a larger PR to attain. But keep helping out and we'll express our gratitude accordingly!

PR #9665, opened 2017-12-05
Added support for party invites by email

### LINKED ISSUE #9450: emailed party invitations don't give the invited member an invitation to the party
When a person who doesn't have a Habitica account receives an emailed invitation to join Habitica, they are not given an invitation to the inviter's party.

Steps to reproduce:

1. Go to your party page.
2. Use the email invitation feature to send an email to an email address that does not currently have a Habitica account.
3. Click on the "join their Party on Habitica" link in the email.
4. Create a Habitica account using the exact email address used for the invitation.
5. The new account should have an invitation to your party but does not.

Here's a sample link from the email:
https://mandrillapp.com/track/click/30295795/habitica.com?p=eyJzIjoiMmanyothercharactersZDRlZjhjY1wiXX0ifQ

Part of the testing for the fix for this should ensure that when the invited player joins the party, the person who invited them should get the `invitedFriend` achievement.

TheHollidayInn:
Fixes #9450

### DIFF
--- a/website/client/components/static/home.vue
+++ b/website/client/components/static/home.vue
@@ -598,11 +598,17 @@
           return;
         }
 
+        let partyInvite = '';
+        if (this.$route.query && this.$route.query.p) {
+          partyInvite = this.$route.query.p;
+        }
+
         await this.$store.dispatch('auth:register', {
           username: this.username,
           email: this.email,
           password: this.password,
           passwordConfirm: this.passwordConfirm,
+          partyInvite,
         });
 
         let redirectTo;
--- a/website/client/components/static/home.vue
+++ b/website/client/components/static/home.vue
@@ -598,9 +598,9 @@
           return;
         }
 
-        let partyInvite = '';
-        if (this.$route.query && this.$route.query.p) {
-          partyInvite = this.$route.query.p;
+        let groupInvite = '';
+        if (this.$route.query && this.$route.query.groupInvite) {
+          groupInvite = this.$route.query.groupInvite;
         }
 
         await this.$store.dispatch('auth:register', {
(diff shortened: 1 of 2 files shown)


website/client/components/static/home.vue
```diff
@@ -598,11 +598,17 @@
           return;
         }
 
+        let partyInvite = '';
+        if (this.$route.query && this.$route.query.p) {
```
paglias: @Alys @TheHollidayInn I don't understand where this p parameter is coming from... In the code I only found a reference to ?groupInvite, are we sure it's correct?

paglias (changes requested):

website/client/components/static/home.vue
TheHollidayInn: If you looks at issue 9450, the link posted has the p parameter. Where are you looking?

paglias: Here https://github.com/HabitRPG/habitica/blob/develop/website/server/controllers/api-v3/groups.js#L1028 (we'll have to change the static/front part too). That link in the issue doesn't work so I'm not sure if it's just wrong

TheHollidayInn: Ah, nice. Easy enough to change. Also see here: https://github.com/HabitRPG/habitica/blob/develop/website/server/controllers/api-v3/auth.js#L163

paglias: So initially we used partyInvite, then expanded the feature to all groups so I created ?groupInvite but kept partyInvite there in case someone used an old invite link but all the emails we're sending out right now should only use `?groupInvite`

PR #9671, opened 2017-12-05
Ensured quest drops are only from incomplete progress (users don't earn excess collection quest items)

### LINKED ISSUE #8248: Collection Quest - Users earn "excess" collection items in multi-category collection quests
General Info
  * UUID: 6d9d36a8-d91f-4737-8e8c-4fe9a4d632fe
  * Browser:  Chome 54.0.2840.98 
  * OS: Mac OSX 

### Description
My party just completed the *Terror in the Taskwoods, Part 2: Finding the Flourishing Fairies* collection quest where the party must collect 15 Brownies, 10 Dryads, and 25 Pixies.  

The problem was that we quickly found the 10 Dryads, but it obviously took longer to get to 25 pixies.  However, as we completed tasks, we kept on earning Dryads, even thought that objective was completed.  So the progress counter on the *Party* page would say stuff like **~14/10~**.

The **expected behavior** should be to only earn collection items in *categories that are not yet satisfied*.  In the end, we had like double the Dryads necessary by the time we finished it out.

It seems like the logic is basically just assigning a "found item" event to a random category without checking to see if any more items of that type are required.  It was a bit annoying to be earning "wasted" items on a collection quest, and it seems like a simple fix should be possible.

-----

_edit by Admin:_

### Current Situation:

When cron runs for a participant of a collection quest, it gets the number of items that the user has found (that's just a single number with no distinction between the types of items). It randomly chooses the type of item for each item found and applies the results to the quest.

### Desired Fix:

The same as above except that when randomly choosing items, it should ignore any item that the party has already found enough of.

There's a suggested fix at this comment: https://github.com/HabitRPG/habitica/issues/8248#issuecomment-561012497

TheHollidayInn:
Fixes #8248

### DIFF
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -874,8 +874,17 @@ schema.methods._processCollectionQuest = async function processCollectionQuest (
   let quest = questScrolls[group.quest.key];
   let itemsFound = {};
 
+  const possibleItemKeys = Object.keys(quest.collect).filter((key) => {
+    return group.quest.progress.collect[key] !== quest.collect[key].count;
+  });
+
+  const possibeItemsToCollect = possibleItemKeys.reduce((accumulator, current, index) => {
+    accumulator[possibleItemKeys[index]] = quest.collect[current];
+    return accumulator;
+  }, {});
+
   _.times(progress.collectedItems, () => {
-    let item = shared.randomVal(quest.collect, {key: true});
+    let item = shared.randomVal(possibeItemsToCollect, {key: true});
 
     if (!itemsFound[item]) {
       itemsFound[item] = 0;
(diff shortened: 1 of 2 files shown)


website/server/models/group.js
```diff
@@ -874,8 +874,17 @@ schema.methods._processCollectionQuest = async function processCollectionQuest (
...
+  });
+
+  const possibeItemsToCollect = possibleItemKeys.reduce((accumulator, current, index) => {
+    accumulator[possibleItemKeys[index]] = quest.collect[current];
+    return accumulator;
+  }, {});
+
   _.times(progress.collectedItems, () => {
-    let item = shared.randomVal(quest.collect, {key: true});
+    let item = shared.randomVal(possibeItemsToCollect, {key: true});
```
Alys: possible not possibe :)

TheHollidayInn: I prefer my version :P Will update!

PR #9677, opened 2017-12-05
Change focused state of introjs button

### LINKED ISSUE #9577: button in modal is white until focus is given to modal
When a new player visits their party for the first time, they see this modal:

image

The button is white so it can't be read, which is incorrect.

When you click on the modal, the correct colour is applied:

image

The desired behaviour is that the button has the correct colour as soon as the screen appears.

This might occur for other modals too.

AlyxMoon:
Fixes #9577

### Changes
Changed the introduction modal button to have intended styling (purple color with white text) when page is loaded by grouping it with the hover and active states.
The bug was caused by the button sometimes being in focus when the page loaded, and the default introjs styling being applied. 

----
UUID: cc212fcc-5f96-403b-a785-17ba66c7b11a

### DIFF
--- a/website/client/components/notifications.vue
+++ b/website/client/components/notifications.vue
@@ -56,7 +56,7 @@ div
     }
   }
 
-  .introjs-button:hover, .introjs-button:active {
+  .introjs-button:hover, .introjs-button:active, .introjs-button:focus {
     background-image: none;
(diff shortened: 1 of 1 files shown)


SabreCat:
Thank you for this, @AlyxMoon! I've noted it as progress toward your next contributor tier!

PR #9679, opened 2017-12-06
Added cache for user styles on chat

TheHollidayInn:


Alys:
I think it would make sense for this to be released in two stages. Firstly we could release the code that saves the avatar details in chat messages - that could go live any time now after brief testing. Then after a few days or longer, we could release the code to display the avatars. That way the Tavern and most popular guilds would have all the historical avatars appearing immediately, which would be a nice display of the new feature and would reduce questions and "bug" reports about why it's not working for older posts.

I'm wondering if we also want this for private messages. @lemoness ?

paglias:
Can we limit the amount of stuff we cache? I fear it'll occupy a lot of memory if you browse a lot of chats and keep Habitica running for days

paglias:
Nevermind, I thought this was a different type of caching :)

Alys:
I've done some brief testing and this is displaying `items.gear.equipped` instead of `items.gear.costume` when `preferences.costume` is `true`. The `equipped` gear should be displayed only when "use costume" is turned off.

If we want to store the entire `items.gear` object in each chat message, we'd need to also store `preferences.costume`, and then `chatMessages.vue` would need to have logic to decide which set of gear to display.

Alternatively, when the message is written to the database, we could store only the set of gear that's appropriate to the user's value of `preferences.costume` at the time they wrote the message. We wouldn't need to store that preference value itself, and we wouldn't need to have anything in the message's data to distinguishes whether the gear is `costume` or `equipped` (i.e., the same data structure could be used regardless of the costume setting). Then `chatMessages.vue` wouldn't need any `costume` vs `equipped` logic - it would just display the single set of data stored in it. That seems simpler to me, and would also save bytes, and would make it easier for third-party tools to also display the avatar.

TheHollidayInn:
The chatMessage.vue doesn't handle display logic. The avatar component does which already looks at the user's costume preference. But, yea. I just forgot to store the costume preference, and to check for that on the server when saving the message.

Alys:
Fair enough! Reusing the avatar component makes more sense than my idea!

Alys:
In each chat message, `userStyles` includes `items.gear.owned`, which is more than 20,000 characters in my account (and will grow larger over time). That's a large increase in data stored and fetched for each guild. Can it be limited to only `items.gear.costume` and `items.gear.equipped`?

This feature is SO MUCH FUN!!!

One thing I especially like is that for the old messages that were posted before this feature existed (the messages that don't contain their own costume information), the costume shown is the oldest one that *does* have cached costume info, rather than being your current costume. I think that's good! It's more historically accurate.

The profile names and tier colours and titles are still working well.

Apart from my concerns with the bytes used for each message, I'd be fine with this going live.

I wrote in a comment above about doing a two-stage release. We could still do that but I don't feel it's as necessary now because of how old messages are handled. If you'd rather release it all at once, I think that's okay.

I'm posting this screenshot totally as proof that it works and not at all because I'm just so darned pretty:
screenshot_2018-01-26_11-52-16

paglias:
Oh yeah let's not store un-needed data here or we'll have serious perf problem :)

TheHollidayInn:
Eh, yea... Good point :P Pushing an update now.

negue:
I don't know if its important, but this won't show if the user is sleeping at that time of the message.

If thats not important then this is ready :+1: otherwise please add the `sleep` preference^^

negue:
Do we also need those properties in each `userStyles` object?:

- `"currentMount" : "", "currentPet" : ""`
- `stats.training`
- `stats.buffs.` `con`, `per`, `int`, `str` `stealth` `streaks`
- everything else from `stats.` except `class` ?

if not this would save us quite a lot of unneeded / unused data in our database^^

Alys:
My opinion is that we shouldn't show the sleep indicator on the avatars in chat. Avatars are a bit prettier without it. I don't feel strongly about it though.

The original thought on currentPet and currentMount was to include them in case in future the chat avatars display the entire avatar. It would also allow a third-party tool to display the full avatar even if Habitica never does, which would make for a very nice integration. It's only a small bit of data so I think it's worth keeping them.

`stats.buffs` is needed for displaying the correct avatar when a transformation buff has been used. You could limit it to `"buffs.seafoam", "buffs.shinySeed", "buffs.spookySparkles", "buffs.snowball"`

I think everything else in `stats` could be removed, but I might be missing a use-case.

lemoness:
I think it's fine to not display the sleeping Zzzs on avatars in chat, so
that shouldn't be a blocker!

TheHollidayInn:
Sweet! Updated.

TheHollidayInn:
I'm planning on merging this next Monday to go through our release cycle unless anyone objects.

PR #9684, opened 2017-12-06
Clone challenges api

### LINKED ISSUE #8116: clone challenge: tasks were not cloned
hey all!

i run a daily challenge in the Pomodoro Playground and today a member pointed out that there were no tasks in the challenge. i clone it everyday and therefore don't pay too much attention while doing so, i only change the name and the tag, therefore I'm not sure _when_ the tasks disappeared, whether it was while cloning or even after that? but deleting 4 to dos and 2 habits by accident within 10 seconds does not really seem possible to me. 

i cloned the challenge again and this time the tasks are all there. 

original challenge: https://habitica.com/#/options/groups/challenges/2efe79bf-17a9-44ec-af49-8941c9a917f7

cloned, missing tasks: https://habitica.com/#/options/groups/challenges/2817172c-8896-4276-be0a-40689fdc34b1

EDIT: firefox 49.0.1 / Win 7

### LINKED ISSUE #9628: When cloning Challenge, Guild selection dropdown defaults to blank
When you are cloning a Challenge, the dropdown where you select a Guild defaults to blank. (This may be an issue with other such dropdown menus, such as the Guild leader/Party leader selection dropdown, but that would need to be investigated separately.) Note that this is a display issue: if you go ahead and create the cloned Challenge, it defaults to being in the same Guild as before. However, the dropdown should show the correct default Guild (or "Public", if that corresponded with the original Challenge).
image_720

TheHollidayInn:
Fixes #8116
Fixes #9628


Alys:
I've promised to test this locally.

website/server/controllers/api-v3/challenges.js
```diff
@@ -719,4 +733,70 @@ api.selectChallengeWinner = {
...
+
+  return cleansedTask;
+}
+
+/**
+ * @api {post} /api/v3/challenges/:challengeId/clone Clone a challenge
+ * @apiName CloneChallenge
+ * @apiGroup Challenge
+ *
+ * @apiParam (Path) {UUID} challengeId The _id for the challenge to close with a winner
```
Alys: This should be "... The _id for the challenge to clone"

website/server/controllers/api-v3/challenges.js
```diff
@@ -719,4 +733,70 @@ api.selectChallengeWinner = {
...
+    if (validationErrors) throw validationErrors;
+
+    const challengeToClone = await Challenge.findOne({_id: req.params.challengeId}).exec();
+    if (!challengeToClone) throw new NotFound(res.t('challengeNotFound'));
+
+    const {savedChal} = await createChallenge(user, req, res);
+
+    const challengeTasks = await Tasks.Task.find({
+      'challenge.id': challengeToClone._id,
+      userId: {$exists: true},
```
Alys: This should be `{$exists: false}`

website/server/controllers/api-v3/challenges.js
```diff
@@ -17,9 +20,81 @@ import {
 import * as Tasks from '../../models/task';
 import Bluebird from 'bluebird';
 import csvStringify from '../../libs/csvStringify';
+import {
+  createTasks,
+} from '../../libs/taskManager';
+
+const TASK_KEYS_TO_REMOVE = ['_id', 'completed', 'date', 'dateCompleted', 'history', 'id', 'streak', 'createdAt', 'challenge'];
```
Alys: I think `'date'` should not be in this list - i.e., it should be cloned into the new challenge's tasks. It's the due date for To-Dos, which the challenge creator is allowed to set.

Alys (changes requested):

website/server/controllers/api-v3/challenges.js
TheHollidayInn: This is copied directly from the old site, so I'm surprised we didn't notice that before, but I'm totally fine changing it!

website/server/controllers/api-v3/challenges.js
TheHollidayInn: Ah, correct. Good catch.

Alys:
Oh, I forgot to mention! When you clone a challenge, the guild selection drop-down menu should have the previous guild selected as the default (while still letting you choose a different guild).

Alys:
Test results so far are below. See the bolded parts for problems. Tell me if this isn't clear!

In a guild that has no gems left:

- a challenge with no tasks but all other fields filled in:
	- [x] original has no gem prize; challenge cloned with no edits made -> new challenge is identical to old one; no gems subtracted from user
	   - [x] original created in private guild -> new challenge is in same guild automatically
	- [x] original has a gem prize; challenge cloned with no edits made -> new challenge is identical to old one; correct gems subtracted from user
	   - [x] original created in tavern -> new challenge is in tavern
	- [x] original has no gem prize; challenge cloned with edits made to all fields and *gem prize added* -> new challenge has expected data; correct gems subtracted from user
	   - [x] original created in public guild but new challenge is edited to be in different guild -> new challenge exists in different guild
	- [x] original has a gem prize; challenge cloned with edits made to all fields and gem prize changed -> new challenge has expected data; correct gems subtracted from user
	- [x] original has a gem prize; challenge cloned with no edits made except that gem prize is changed -> new challenge is identical to old one except for prize; correct gems subtracted from user
- [x] a challenge with one unused task is cloned (default value; contains notes) -> new challenge has identical task with new ID, and has correct tasksOrder array
   - [x] original created in public guild -> new challenge is in same guild
- [x] a challenge with one task (with notes) used by participants is cloned (value is more "blue"; user has added note to their own copy) -> new challenge has identical task including default note, except with default value and new ID, and has correct tasksOrder array
   - [x] original created in tavern -> new challenge is in tavern
- [x] a challenge with three tasks in each of the four task types is cloned -> new challenge has identical tasks with new ID, and has correct tasksOrder arrays
   - [x] original created in party -> new challenge is in party

In a guild that has 2 gems, cloned by the leader:
- [x] cloning a challenge with a 2 gem prize takes gems from the guild

In a guild that has 3 gems, cloned by the leader:
- [x] cloning a challenge with a 1-gem prize takes gems from the guild 
- [x] cloning a challenge with a 6 gem prize takes gems from the guild and from the player 
   - [x] the player's gem balance displayed in the header is decreased immediately (you don't need to click sync to see that the gems have been removed)

In a guild that has 4 gems, cloned by a player who is NOT the leader:
- [x] cloning a challenge with a 1-gem prize takes gems from the player
   - [x] **The player's gem balance displayed in the header does not decrease immediately - you have to click sync to see that the gems have been removed.**

TheHollidayInn:
Thanks for the list!! The gem prize change is interesting. I'm assuming the clone is keeping the original prize rather than updating. 

I'll look into the group selection as well.

Alys:
The clone actually does have the new gem prize! It just doesn't subtract the gem from the user on the server-side. :scream: :)

TheHollidayInn:
Alright, I'll write a test around it, but since it is using the same function as createChallenge, I'm pretty confused on how that would happen. I'm sure I am just missing something subtle!

TheHollidayInn:
Actually, I can't recreate the gem issue at all. The gems are subtracted client side for me. Then on refresh, they are still gone.

TheHollidayInn:
I have updated the default group selection.

Alys:
The gem prize bug is still happening for me. When I start with a challenge that has no gem prize, clone it and add a gem prize, then the new challenge has the prize but no gems are permanently subtracted from my test account.

Alys:
To give @TheHollidayInn full credit, he's realised that the gem prize was not subtracted from my test account on the server side because it was quite correctly coming out of the guild's gem bank. :) The only real problem was that the gem prize was temporarily subtracted from the test account client-side.

Alys:
I've finished all the tests in my [comment above](https://github.com/HabitRPG/habitica/pull/9684#issuecomment-351415832). The only remaining problem is the one at the end, which I've bolded. When a challenge is being cloned by a player who isn't the leader of the guild, the gem prize IS correctly taken from the player's account but the gem balance displayed in the header isn't updated immediately. You have to sync to see that your gems have decreased.

TheHollidayInn:
I'm not sure I understand the test case. 

- You are cloning a challenge in to a guild that you are not the leader of? From the api, non leaders can not create challenges for guilds

- If you are cloning a guild challenge into a guild that you are the leader of, then the previous guild's balance doesn't matter and the results depend on the new guilds balance. 

- If you are cloning a guild challenge into "Public" then the gem is removed.

Alys:
Guild or party members can create challenges in the group (or clone challenges into the group) for any group, unless the group leader has selected the group's option to prevent other group members from creating challenges - i.e., any member can clone when this option is turned off:
screenshot_2018-01-16_18-51-00

TheHollidayInn:
OH wow... I promise I work here.. :P

Alys:
There's a small client-side problem now for this situation:

1. Be the leader of a guild that has 4 gems in its bank.
2. Create a challenge in that guild and give it a 2 gem prize.
3. Check that the guild still has 2 gems in its bank.
4. Make note of how many gems you own (ensure it's at least 2).
5. Clone the challenge.
6. **Observe that the website's display of your personal gems decreases by 2, even though the gems should have come from the guild's bank.** The expected behaviour is that your gems do not appear to decrease.
7. Sync and observe that the 2 gems are returned to your account (i.e., there's no bug server-side).
8. Refetch data for the guild and observe that the guild's gems are now 0 (as expected; no bug here either).

When you've pushed  a fix for this, nag me every day until I've tested it again! I REALLY should be able to test within a day or at the very most two days.

TheHollidayInn:
Updated!

Alys:
It's perfect. :)  I've labelled it ready for merging.

Below are the setup actions I took and the tests. All passed. Gem balances were always correct client-side and server-side.

- guild has 4 gems, guild leader has 20 gems
   - [ ] guild leader creates challenge with 2 gem prize - guild now has 2 gems
   - [ ] guild leader clones challenge - guild now has 0 gems, guild leader has 20 gems
- guild has 4 gems, guild leader has 20 gems
   - [ ] guild leader creates challenge with 3 gem prize - guild now has 1 gems
   - [ ] guild leader clones challenge - guild now has 0 gems, guild leader has 18 gems
- guild has 0 gems, guild leader has 18 gems
   - [ ] guild leader clones challenge that had 2 gem prize - guild leader has 16 gems
- guild has 4 gems, guild member (not leader) has 20 gems
   - [ ] guild member creates challenge with 2 gem prize - guild has 4 gems, member has 18 gems
   - [ ] guild leader clones challenge - guild now 4 gems, guild leader has 16 gems

PR #9687, opened 2017-12-06
Given back header the priority over snackbar with z-index value and a padding to avoid collision.

### LINKED ISSUE #9678: snackbar notifications should not cover the navigation / menu bar
The snackbar notifications that appear at the top right corner of the website and then disappear after a few seconds should never cover the menu bar. They should start appearing just under the menu bar.

This is has been created for jostein s to work on this fix, based on their comment on this Trello card: https://trello.com/c/684Dv1py/320-in-game-notifications-quiet-notifications-bell-click-to-view-notifications

josteins1:

Fixes https://github.com/HabitRPG/habitica/issues/9678

### Changes

Adjusted the top padding of the snackbar notification to be higher
than the main header so the two doesn't overlap.
The header is set at 56px and the old snackbar was set at 10px, now it's set to 65px.
This gives the two elements some space and avoids the overlap.

At the same time the z-index value for snackbar is changed from 99999 (random number?) to 999 (1 less than the header and its drop down menu)

----
UUID: 5e7df578-43e2-43aa-ac2e-8ed10d8f2fd5
snackbar vs header test

### DIFF
--- a/website/client/components/snackbars/notifications.vue
+++ b/website/client/components/snackbars/notifications.vue
@@ -8,7 +8,7 @@
   .notifications {
     position: fixed;
     right: 10px;
-    top: 10px;
+    top: 65px;
     width: 350px;
     z-index: 99999;
   }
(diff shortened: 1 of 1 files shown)


SabreCat:
Thank you for this, @josteins1! I've added "Blacksmith" to your contributor titles, and noted this as credit toward your second tier!

Issue #9031, opened 2017-09-12
VersionError prevents quest from ending (participants received rewards but quest was not removed from party)


Alys:
If a VersionError occurs when a party's `group` object is being saved, it can leave a quest in an inconsistent state.

The report below from Gumnos (43e2413a-f440-43c8-b96c-85b93024488c) in the Report a Bug guild describes a quest that should have ended normally but didn't. From the user's description and from what I've seen in the data I have access to, it looks like all of Habitica's code for ending the quest did execute except that the final `group.save();` didn't work, meaning that the participants' accounts were updated to indicate that the quest had finished (they were given the rewards and had the quest data from from their own `party` objects) but the party itself didn't have the quest data removed. 

Loggly shows a VersionError for what I believe was the event that should have saved the change to the party.

Gumnos's words are in bold and I've added comments in italics.

**Just had a weird quest experience (Mandrill/monkey quest) where my lvl 98 warrior did a one-shot kill on the boss (repeated use of Brutal Smash).**

_I confirmed in Amplitude that enough damage had been done to kill the boss (health 400). There was a non-resting cron at 2017-09-08 08:36:56 UTC with a progressUp of 429.6083716046._

**It completed the quest and my party received the eggs as spoils.**

_Gumnos posted later to say that they're pretty sure the normal quest completion modal appeared. I confirmed that one quest participant had three eggs and the other had three Monkey pets. I'm confident they were received from that specific quest because they were at the end of the `item.pets` and `items.eggs` objects (which indicates that they'd most likely been added after all other eggs/pets). The previous pet quests had been Horse, Whale, and Seahorse, and the Monkey items can be seen after all of those eggs/pets. The participants have only one achievement for the Monkey quest._

**I attempted to start another quest, but couldn't because it for some reason the system thought we were still on a quest. Checking the Party page showed we were still on the Mandrill quest (the boss still had an almost-full health bar, likely just the couple points of damage done by the other person in my party).**

_This is consistent with the database still containing the data from that quest._

**Completing additional tasks didn't show the little "compass" in the top right where pending damage appears, so it was like I was simultaneously on the quest but unable to do damage. To start another quest, I had to abort the Mandrill quest that was "in progress" . Upon doing that, the Mandrill quest appeared back in my inventory such that I could attempt it again, even though I'd already completed it.**

_That's all consistent with the participants' quest data being correctly removed from their own accounts._

[...]

PR #9703, opened 2017-12-09
Move Chat to Model

TheHollidayInn:


librarianmage:
If you don't mind me asking, what will this affect?

TheHollidayInn:
This will affect all chat in groups. But this is backwards compatible as I query the chat and apply it to the group model. 

I still have to fix the other routes though before this is done.

librarianmage:
So now chat messages will be forced to conform to this model?

TheHollidayInn:
Yea, I’ll keep what we currently implement and we can add fields. Are you concerned?

librarianmage:
No, just wondering what it will do, since when I did the documentation
about chat, the only info I could go off of was the commented part in the
group model.

website/server/models/chat.js
```diff
@@ -0,0 +1,13 @@
...
+const schema = new mongoose.Schema({
+  timestamp: Date,
+  user: String,
+  text: String,
+  contributor: {type: mongoose.Schema.Types.Mixed},
+  uuid: String,
+  id: String,
+  groupId: {type: String, ref: 'Group'},
+});
+
```
paglias: We should add the BasePlugin here before merging, like the other models

TheHollidayInn: Will do!

website/server/middlewares/errorHandler.js
```diff
@@ -16,7 +16,7 @@ module.exports = function errorHandler (err, req, res, next) { // eslint-disable
   // Otherwise try to identify the type of error (mongoose validation, mongodb unique, ...)
   // If we can't identify it, respond with a generic 500 error
   let responseErr = err instanceof CustomError ? err : null;
-
+  console.log(err)
```
crookedneighbor: I assume this is unintentional.

TheHollidayInn: Yep. Lint caught it too. I'll remove it during my next updates.

TheHollidayInn:
Fixes #9031

negue:
(I thought first I'd have to run a migration, but it worked out of the box 🎉 )

Will the old messages stay in the group objects or will those be converted to the new model?

TheHollidayInn:
Sweet! The plan is to have the old group messages eventually be removed as new messages come in. But we might just need to migrate them.

lemoness:
@TheHollidayInn What is the status of this?

website/server/models/chat.js
```diff
@@ -0,0 +1,26 @@
...
+  flagCount: {type: Number, default: 0},
+  likes: {type: mongoose.Schema.Types.Mixed},
+  userStyles: {type: mongoose.Schema.Types.Mixed},
+}, {
+  minimize: false, // Allow for empty flags to be saved
+});
+
+schema.plugin(baseModel, {
+  noSet: ['_id'],
+  timestamps: true,
```
paglias: do we care about timestamps here? The message date is already saved in timestamp and I don't think we allow edits

website/server/libs/chat/group-chat.js
```diff
@@ -0,0 +1,18 @@
+import { model as Chat } from '../../models/chat';
+
+// @TODO: Don't use this method when the group can be saved.
+export async function getGroupChat (group) {
+  const groupChat = await Chat.find({groupId: group._id}).limit(200).sort('-timestamp').exec();
```
paglias: is the limit 400 for some groups?

website/server/models/chat.js
TheHollidayInn: Ah, true. I can remove this.

website/server/libs/chat/group-chat.js
TheHollidayInn: Not that I know of. We talked about doing it for Group Plans, but after we merged this PR.

Alys: I think it's already in place.
website/server/models/group.js:537:  const MAX_SUBBED_GROUP_CHAT_COUNT = 400;

TheHollidayInn: Ah, okay. I can change that.

paglias:
When this gets merged, we should post a message in #10019

Alys:
When a quest starts on production, the message in chat is like this:
```
        {
            "id": "c2819113-708e-4c85-b74e-64f19118c15b",
            "text": "`Your quest, Attack of the Mundane, Part 1: Dish Disaster!, has started.`",
            "timestamp": 1498284774350,
            "likes": {},
            "flags": {},
            "flagCount": 0,
            "uuid": "system",
            "_meta": {
                "participatingMembers": "d904bd62-da08-416b-a816-ba797c9ee265"
            }
        },
```

In beta it's like this:
```
{
	"_id" : "a1e756e8-3bdf-4eef-8d36-80f8847dd7b6",
	"flagCount" : 0,
	"flags" : {
		
	},
	"id" : "45a931d1-05ed-40ab-9e6d-2f2a7e25bf42",
	"text" : "`Your quest, The Basi-List, has started.`",
	"timestamp" : ISODate("2018-04-19T11:00:30.485Z"),
	"likes" : {
		
	},
	"uuid" : "system",
	"groupId" : "ae9bf3a2-8050-4253-85ba-14eab6b19e72",
	"__v" : 0
}
```

I.e., it's missing the `_meta` object.

Also, there's tests for that object in `develop`, but they haven't told us about this error, so those tests must be broken in beta. That's worrying so maybe they could be fixed before the `_meta` object is added back into the chat messages so that we can be sure those tests are working?

TheHollidayInn:
I have that fixed on the PR. I'll update Beta in a sec

Alys:
Old guild messages are not being purged when there's more than 200 of them. If you'd like to test, you can use guild 4e5f91ea-692e-44d4-bd8d-82201ddc1313 which has slightly over 200 messages now, most of them numbered - feel free to post whatever you want there. The same should apply to parties and the Tavern. For any guild or party that has a group plan, the number of messages kept must be 400.

Alys:
For the migration script that will convert messages from the old storage system to the new, can I get a copy of it to test on a local install before it's run over the beta database? It will be easier to confirm that it's working correctly locally where I have only a small number of guilds and parties. Once that's tested, it could be run over the beta database and then I can do some final tests (e.g., to confirm that the console errors I've been seeing are just from messages existing in the old format).

Alys:
Mostly for my own reference so I don't forget:

We must remember to make sure the Blacksmiths know about that migration script too. Many of them are likely to have local installs set up with a lot of test data (I know I do) and it would be convenient for them to be able to convert their database rather than having to start from scratch.

TheHollidayInn:
- Just pushed the limit fix. The vars weren't being exported - I missed that :P  
- I pushed the migration. However, I have already run it on Beta Thursday or Friday.
- Console message like this `TypeError: Cannot read property 'classSelected' of undefined` and some other user errors are because I moved old prod chat over but didn't move every user over. Beta has a migration to grab userIds and move a lot of that connected user data, but not all. So that leads to chat messages with non existing users.

migrations/groups/migrate-chat.js
```diff
@@ -0,0 +1,53 @@
+// @migrationName = 'MigrateGroupChat';
+// @authorName = 'TheHollidayInn'; // in case script author needs to know when their ...
+// @authorUuid = ''; // ... own data is done
+
+
+/*
+ * This migrations will iterate through all groups with a group plan a subscription and resync the free
+ * subscription to all members
```
Alys: Can this comment be updated?

Alys:
The console log shows a lot of `TypeError: e.flags is undefined` errors in the Tavern and parties. To see them, visit one of those pages and then reload it. They don't happen in production. There might be other errors as well but they're hard to find so even if the e.flags errors are benign, it would be good if they could be prevented.

In the 5481ccf3-5d2d-48a9-a871-70a7380cee5a guild, I saw two examples of "t.response is undefined" in the console after one reload but I can't replicate it so it might have been a temporary problem and if you want to ignore it, that's fine with me. :)
```
Unhandled promise rejection 
TypeError
columnNumber: 1
fileName: "https://habitrpg-beta.herokuapp.com/static/js/app.d2e6be8e3bb70e4bebb7.js"
lineNumber: 1
message: "t.response is undefined"
stack: "created/<@https://habitrpg-beta.herokuapp.com/static/js/app.d2e6be8e3bb70e4bebb7.js:1:1\nT/</<@https://habitrpg-beta.herokuapp.com/static/js/vendor.b91cfcde844f24cdc5d8.js:6:211261\nT/<@https://habitrpg-beta.herokuapp.com/static/js/vendor.b91cfcde844f24cdc5d8.js:6:211131\nl@https://habitrpg-beta.herokuapp.com/static/js/vendor.b91cfcde844f24cdc5d8.js:6:74862\n"
```

I'm getting poor performance in my browser when viewing guilds/party/tavern on beta, with my browser really struggling when I reload the site, but that might just be because it's the beta server. Let me know if you want any more details.

Can this line https://github.com/HabitRPG/habitica/blob/develop/migrations/migration-runner.js#L20 be changed as part of this PR? All of the contributors will need to run that script over their own databases and it will be much easier to explain to them how if we don't have to tell them to edit that script and then revert the edit when they're finished. :)
FYI I'll document on the wiki how that migration-runner script works and how normally you do need to edit it to run any migration in general, but for this one specific case it would be good for them to not have to edit.

Alys:
I still can't get old messages to automatically delete. Guild 4e5f91ea-692e-44d4-bd8d-82201ddc1313 has more than 200 now and when I post more, the old ones don't disappear.

website/server/controllers/api-v3/chat.js
```diff
@@ -164,35 +166,35 @@ api.postChat = {
...
+      user.party.lastMessageSeen = newChatMessage.id;
       toSave.push(user.save());
     }
 
-    let [savedGroup] = await Promise.all(toSave);
+    await Promise.all(toSave);
 
-    // realtime chat is only enabled for private groups (for now only for parties)
-    if (savedGroup.privacy === 'private' && savedGroup.type === 'party') {
+    // real-time chat is only enabled for private groups (for now only for parties)
```
Alys: Not exactly related to this PR, but should this comment be changed to something like "real-time chat is not currently running" or "TODO make real-time chat work" or something along those lines?

PR #9706, opened 2017-12-11
clarify that "Leave" refers to guild/party; fix pluralisation in keep/remove challenge tasks

Alys:
This changes the "Leave" button on groups to "Leave Guild" or "Leave Party" as discussed with @Tressley and @lemoness last week. This clarifies that the button is not referring to the challenges that are just above it.

image
image

This also changes the "Keep It" / "Remove It" button when leaving a challenge to "Keep/Remove Them" because the buttons are referring to all the challenge tasks.
current production site:
image
New code:
image


website/client/components/groups/group.vue
```diff
@@ -105,7 +105,7 @@
       .section(v-if="sections.challenges")
         group-challenges(:groupId='searchId')
     div.text-center
-      button.btn.btn-danger(v-if='isMember', @click='clickLeave()') {{ $t('leave') }}
```
paglias, line 108: I think this string is only used there so we can remove it

paglias (changes requested):

Tressley:
Small addition, the button strings for leaving a Challenge should be "Keep Tasks" and "Remove Tasks."

website/client/components/groups/group.vue
Alys: It's still used in buttons on the guild list pages:
img

Alys (commented):
I've changed  "Keep Them" and "Remove Them" to "Keep Tasks" and "Remove Tasks".

paglias (approved):

PR #9707, opened 2017-12-11
Fix password reset when querying for emails with upcase characters (fixes #9059)

### LINKED ISSUE #9059: password reset process fails if a user's email address has upper-case letters
If a user's email address in Habitica contains any upper-case letters, when they use the password reset form, their account is not given a value for `auth.local.passwordResetCode` and they do not receive a password reset email. This happens even if they enter their email address into the reset form using the exact case that it's stored in in Habitica.

If their email address in their account is changed to all lower case, the password reset form works correctly, no matter whether they enter their email address into the form in the same case as it's stored in Habitica or a different case.

This looks like the code to be changed:
https://github.com/HabitRPG/habitica/blob/136947169b466ed02b12a151c4576c791154a069/website/server/controllers/api-v3/auth.js#L563-L564
The find command should be case-insensitive and the `toLowerCase` won't be needed.

matchajune:
Fixes #9059

### Changes
[x]: When APIv3 was released 2 years ago, email was sanitized with `.toLowerCase()` upon creating a user. However, an inconsistency is currently happening when updating an email: emails submitted to update aren't sanitized.

Therefore, instead of checking for emails consisting of upcase characters before sending a password reset, I fixed updating an email to save as lowercase.

In addition, there is a migration script `20171211_sanitize_emails.js` to update existing Users' emails to lowercase for consistency.

----


migrations/20171211_sanitize_emails.js
```diff
@@ -0,0 +1,14 @@
+/*
+  User creation saves email as lowercase, but updating an email did not.
+  Run this script to ensure all lowercased emails in db AFTER fix for updating emails is implemented.
+  This will fix inconsistent querying for an email when attempting to password reset.
+*/
+
+db.users.find().forEach(user => {
```
Alys: This would need to filter for users who already have a value for auth.local.email (accounts that use only social media login don't have that), and would also need to update the users in batches to avoid overloading requests to the database. See some of the other migrations scripts for examples (just the more recently updated ones). However you might want to wait for feedback about the overall method first. :)

test/api/v3/integration/user/auth/POST-register_local.test.js
```diff
@@ -288,6 +288,21 @@ describe('POST /user/auth/local/register', () => {
       });
     });
 
+    it('sanitizes email params to a lowercase string before creating the user', async () => {
+      let username = generateRandomUserName();
+      let email = 'ISANEmAiL@GMAIL.coM';
```
Alys: best to use @example.net here rather than @gmail.com, just in case someone has that email address and doesn't like it being used in github

Alys (changes requested):

Alys:
Interesting approach! There's 17 email addresses that are currently associated with two Habitica accounts each, using different case variants of the addresses. If we go with this approach, we'd need to handle those in some way. I'm happy to look into that further if needed. I'll leave a decision about whether we use this approach to the staff.

matchajune:
Thanks. I will look forward to your reply. In the meantime, I will push a new commit with the change requests!

migrations/20171211_sanitize_emails.js
matchajune: I will wait for feedback as suggested!

paglias:
@Alys did we reach any decision on this?

Alys:
I didn't hear anything about a decision.

It's occurred to me though that no matter which solution we use for case-insensitive password resets, those 17 users would have trouble if they tried to use the reset, so we'll have to sort them out regardless. I'll pull up the list of users again and I and the other mods will discuss contacting them.

TheHollidayInn:
What issues would those 17 users have? Could we still merge this PR while waiting to contact them?

Alys:
There's only one user left who needs to be considered for this. They have two accounts but have not used Habitica recently from either account. Mods will probably email them to help them with this.

This PR could be merged if it's reviewed, finished, and tested before that one user's case has been resolved.

TheHollidayInn:
Thanks, @alys!

@matchajune looks like Travis found a lint error. Can you fix that?

matchajune:
Thanks for catching the lint error. I will be able to fix that tonight!

migrations/20171211_sanitize_emails.js
Alys: @matchajune Your suggested method is approved, so this change will be needed before the PR is merged.

Alys:
As a note to admins: The migration script in this PR will need to be run shortly before the change is deployed to production and also shortly afterwards to catch any accounts that were created in the meantime (or just search recent accounts for uppercase in emails - that might be faster than a second run). FYI @SabreCat @TheHollidayInn since you're like to be the ones deploying.

migrations/20171211_sanitize_emails.js
Alys: It would also be good if the update was performed only when the email address was not already fully lower case. That would speed up the migration.

matchajune: I took your comments and combined them to check for emails that don't have at least one capital letter. I'm not 100% certain I've written it correctly and would love feedback before proceeding. Thank you for your patience!

migrations/20171211_sanitize_emails.js
```diff
@@ -0,0 +1,14 @@
+/*
+  User creation saves email as lowercase, but updating an email did not.
+  Run this script to ensure all lowercased emails in db AFTER fix for updating emails is implemented.
+  This will fix inconsistent querying for an email when attempting to password reset.
+*/
+
+db.users.find({'auth.local.email':{$not: /[A-Z]/}}).forEach(user => {
```
Alys: Will this actually find addresses that have no capital letters?

migrations/20171211_sanitize_emails.js
Alys: We should still update the users in batches as mentioned above, rather than trying to do them all at once - that can cause loading problems. Have a look at migrations/20180130_habit_birthday.js - I think that would be a decent example, although if other admins give you a better one, listen to them. :)  

The `query` variable can contain the search for addresses that have upper-case letters.

The `fields` parameter in the `dbUsers.find` command could limit the returned fields to just the email address.

The logic for the `set` variable can be simplified for this - all you need is setting the email address.
As a note to whichever admin runs the script: any errors output by the set command should be saved for examination - errors would happen if a new account had been created with a lower-cased version of an email address in an old account (we've already checked for examples of that but one or two more might be created before this migration is run). 

The `inc` variable and the code that uses it can be deleted.

migrations/20171211_sanitize_emails.js
matchajune: I fixed the query in order to find all emails that have any capital letters.

paglias:
@SabreCat the migration attached will have to be run when this is deployed

paglias:
@matchajune  thanks for this! I'm merging the PR and awarding your contributor tier.

SabreCat:
Huh, the migration seems to have been deleted from `develop`? I'll run it from the branch.

SabreCat:
Ah, never mind, found it! Moved to the `archive/2017` directory.

Issue #9638, opened 2017-12-02
Quick Inventory bug: "Special" appears empty after scrolling in "Food"


Alys:
In the Report a Bug guild, Aerilita ( 3b724a41-48d3-4048-ae27-70b258cf5a72 ) reported:

"On Stable page (website), if Food in Quick Inventory is scrolled (by clicking an arrow at the side), then switching to Special gives me an empty panel (like "second page" of Special, when there's only one page's worth of items). Scared me for a moment when I thought I lost two saddles. Scrolling on the Special tab fixes it (I see them now)."

I've replicated it.

The correct behaviour is that when you switch in either direction between Food and Special in the Quick Inventory, you should be automatically placed in the first "page" for that section.

matchajune:
@Alys May I take this?

TheHollidayInn:
Go for it!

lemoness:
@matchajune This may be related to this problem, since there shouldn't be "pages" at all: https://github.com/HabitRPG/habitica/issues/9366  So you may want to look at both of them at once.

matchajune:
@lemoness I'll grab the other and will do. Thank you!

thebat137:
I've found a possibly related issue in the Food tab of the Quick Inventory.  If I:

1.  Have more than one screen's worth of food (e.g., some of each color),
2.  Scroll away from the beginning of the food list,
3.  Begin feeding pets, and
4.  Run out of enough types of food that I no longer have more than one screen's worth,

Then some of my other types of food disappear from the screen as well and don't reappear until I switch back and forth from the Special tab or refresh the window.  I just did this now and it looks like I only have strawberries, pink cotton candy, and honey, even though, opening the stable page in another window, i can see that I also have meat and potatoes.  I think the food that seems to disappear when I do this is usually the food at the beginning of the list.

veeeeeee:
Closing due to inability to replicate after multiple attempts and test accounts. If someone else does manage to replicate this issue, we can reopen.

Alys:
I believe this was fixed in https://github.com/HabitRPG/habitica/pull/9711 by matchajune but the autoclose wasn't triggered.

PR #9711, opened 2017-12-12
fix drawer-slider carousel function && convert to pointer logic (fixes #9366 AND #9638)

### LINKED ISSUE #9366: Incorrect Divider in Food Quick Inventory
No matter how many times you cycle through the food in the quick inventory, the item on the right-hand side is always separated out slightly. The idea is that there is a separation that loops around like a conveyor belt. So if you have 10 items, the separator only shows up between item 10 and item 1. Right now, it just separates out whichever item is on the right-hand side. 

We had envisioned it working more like the "Multiple Items" slider demo seen here: http://kenwheeler.github.io/slick/

The divider should only appear between the last item and the first item as a visual identifier that you're back to the beginning, and there should not be a gap.

The divider:
screen shot 2017-11-01 at 9 42 34 pm

What it currently does:

screen shot 2017-11-01 at 9 42 41 pm

It should cycle through like this:

screen shot 2017-11-01 at 9 42 50 pm

matchajune:
[x]: Fixes https://github.com/HabitRPG/habitica/issues/9366 && https://github.com/HabitRPG/habitica/issues/9638

### Changes
[x]: Delete logic for paging for `drawerSlider.vue` and implement pointer logic. The pointer will act as the index of the array `this.items` and take `this.itemsPerPage()` elements.
[x]: Handle logic to display divider correctly.
[x]: Evenly space out `this.items` for the items being shown.
[x]: Create a new prop in `drawerSlider.vue` called `itemType` and watch for its value to change and reset `this.pointer` to 0 if it changes. This will remedy any false-negatives of not showing a user's items.

----

### DIFF
--- a/website/client/components/ui/drawerSlider.vue
+++ b/website/client/components/ui/drawerSlider.vue
@@ -130,7 +130,7 @@
         if (firstSlice.length === itemsPerPage) {
           return firstSlice;
         } else {
-          let getRemainderItems = Math.abs(itemsPerPage - firstSlice.length);
+          let getRemainderItems = itemsPerPage - firstSlice.length;
           let secondSlice = items.slice(0, getRemainderItems);
 
           return firstSlice.concat(secondSlice);
--- a/website/client/components/ui/drawerSlider.vue
+++ b/website/client/components/ui/drawerSlider.vue
@@ -4,13 +4,13 @@
   )
     div.slider-button-area.left-button(
       v-if="scrollButtonsVisible",
-      @mousedown.left="lastPage"
+      @mousedown.left="shiftLeft"
     )
       a.slider-button
         .svg-icon(v-html="icons.previous")
     div.slider-button-area.right-button(
       v-if="scrollButtonsVisible",
-      @mousedown.left="nextPage"
+      @mousedown.left="shiftRight"
     )
       a.slider-button
         .svg-icon(v-html="icons.next")
@@ -18,8 +18,11 @@
     // 120 = width of the left/right buttons
     div.sliding-content(v-resize="500", @resized="currentWidth = $event.width - 120")
       .items.items-one-line
-        template(v-for="item in pages[currentPage]")
-          div.vertical-divider(v-if="item.ofNextPage")
+        template(v-for="item in showItems")
+          div.vertical-divider(
+            v-if="shouldAddVerticalLine(item)"
+            v-bind:style="dividerMargins"
+          )
 
           slot(
             name="item",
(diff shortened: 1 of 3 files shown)


website/client/components/ui/drawerSlider.vue
```diff
@@ -115,45 +117,58 @@
...
+      showItems () {
+        let items = this.items;
+        let pointer = this.pointer;
+        let itemsPerPage = this.itemsPerPage();
+        let firstSlice = items.slice(pointer, pointer + itemsPerPage);
+
+        if (firstSlice.length === itemsPerPage) {
+          return firstSlice;
+        } else {
+          let getRemainderItems = Math.abs(itemsPerPage - firstSlice.length);
```
negue: `itemsPerPage` and `firstSlice.length` shouldn't be less than 0, so I guess there is no need for `Math.abs`

website/client/components/ui/drawerSlider.vue
```diff
@@ -115,45 +117,58 @@
...
-          this.currentPage = 0;
+          this.pointer = this.items.length - 1;
         }
       },
-
       itemsPerPage () {
         return Math.floor(this.currentWidth / (this.itemWidth + this.itemMargin));
       },
+      shouldAddVerticalLine (item) {
+        return this.items[this.itemsPerPage() - 1] === item && this.pointer !== 5;
```
negue, line 170: why the `pointer !== 5` check ?

negue (changes requested):
its LGTM :+1: 

only have the two requests ^^

website/client/components/ui/drawerSlider.vue
matchajune: The checker is to not show the pointer as so:

screen shot 2017-12-14 at 4 28 10 pm

negue (approved):

SabreCat:
Thank you, @matchajune! Noted toward your third contributor tier!

paglias:
@matchajune hey, I think we found some issues with this PR, in particular some items like saddles are shown twice and the direction of the scrolling in the quick inventory seems reversed, so clicking on the forward button actually does the opposite

website/client/components/ui/drawerSlider.vue
```diff
@@ -115,45 +117,58 @@
...
-        });
+      showItems () {
+        let items = this.items;
+        let pointer = this.pointer;
+        let itemsPerPage = this.itemsPerPage();
+        let firstSlice = items.slice(pointer, pointer + itemsPerPage);
+
+        if (firstSlice.length === itemsPerPage) {
+          return firstSlice;
+        } else {
```
paglias, line 132: I think this else code shouldn't run if the total number of items is < firstSlice.length

paglias:
I think I've fixed them here https://github.com/HabitRPG/habitica/pull/9819

matchajune:
Thanks for fixing the issues @paglias

PR #9714, opened 2017-12-13
Refactor Purchase API - Part 1

negue:
Merge all purchase / buy operations into one `buy`-op, except for the special ones like fortify / orb of rebirth

Todo:
- ~~[ ] only use the only `buy`-op in units tests~~
- [x] move purchase of `quantity` inside each op-method
- [x] replace calls `purchase` and `purchaseWithSpell`  to `buy`


website/common/script/ops/buy/buySpecialSpell.js
```diff
@@ -9,19 +9,34 @@ import {
...
 
   let item = content.special[key];
   if (!item) throw new NotFound(i18n.t('spellNotFound', {spellId: key}, req.language));
 
-  if (user.stats.gp < item.value) {
+  if (user.stats.gp < item.value * quantity) {
     throw new NotAuthorized(i18n.t('messageNotEnoughGold', req.language));
   }
-  user.stats.gp -= item.value;
+  user.stats.gp -= item.value * quantity;
```
TheHollidayInn: Are you removing the loop? I'd rather not have the quantity code in multiple places

negue: yes I removed the loop, I think its much cleaner having the `quantity` inside of each op, instead of having a loop which calls the method per iteration. Its also the better way for performance

TheHollidayInn: Yea, I can see that. But, this does show a lot of duplication and now is a case of "Shotgun Surgery." So, it would be good to remove that.

negue: What do you think about a shared method of checking the currency by type, value and quantity and the same for removing the currency ?

TheHollidayInn: You can check by type, but I think what we should have is an object or class with the value method. Then using one of the various methods of Javascript inheritance, we can return value *quantity on a `getValue` method and subtract that from a user in the buy method. Is the what you are also thinking?

negue: yes that is planned too :), I just wanted to have the first refactoring to be api save, once thats ok, the inner refactoring starts in a separate PR

TheHollidayInn: Hmm. Alright :P Well, before we go to the "inner" section, we should get this up to 100% coverage for the api and the common ops.

negue: Yeah you are right, gonna focus on the coverage first

website/common/script/ops/buy/buyMysterySet.js
```diff
@@ -1,14 +1,18 @@
...
 import each from 'lodash/each';
 import {
   BadRequest,
   NotAuthorized,
   NotFound,
-} from '../libs/errors';
+} from '../../libs/errors';
+
+module.exports = function buyMysterySet (user, req = {}, analytics, _window) {
+  if (typeof window !== 'undefined') {
```
paglias: is this used for tests only? What about definind a dummy window under `global.window` in tests?

website/common/script/ops/buy/buyArmoire.js
```diff
@@ -23,10 +23,6 @@ module.exports = function buyArmoire (user, req = {}, analytics) {
     throw new NotAuthorized(i18n.t('messageNotEnoughGold', req.language));
   }
 
-  if (item.canOwn && !item.canOwn(user)) {
```
paglias: why was this check removed? I cannot find it anywhere else

paglias (changes requested):
Left two comments, otherwise looks good!

website/common/script/ops/buy/buyMysterySet.js
negue: oh thanks for the hint!, thats much better

website/common/script/ops/buy/buyArmoire.js
negue: I removed it because `armoire`'s 

```
  canOwn () {
    return true;
  },
```

should I add the check back?

paglias: yeah I'd keep it in case we change it in future

website/common/script/ops/buy/buyMysterySet.js
```diff
@@ -27,8 +23,8 @@ module.exports = function buyMysterySet (user, req = {}, analytics, _window) {
     throw new NotFound(i18n.t('mysterySetNotFound', req.language));
   }
 
-  if (typeof _window !== 'undefined' && !req.noConfirm && _window.confirm) { // TODO move to client
-    if (!_window.confirm(i18n.t('hourglassBuyEquipSetConfirm'))) return;
+  if (!req.noConfirm && window.confirm) { // TODO move to client
```
paglias: shouldn't this be window && window.confirm for the node env where window is undefined?

paglias (changes requested):

website/common/script/ops/buy/buyMysterySet.js
negue: thats what I tried too, but:
`if (window && window.confirm) ` results in
`2018-01-28T16:07:20.949Z - error: ReferenceError: window is not defined`

paglias: try 

```
if (typeof window !== 'undefined') {
    window.confirm...
}
```

negue: I thought about it a second after I wrote my message haha, but now I think this is obselete, because every `hourglass` purchase has already a `purchase for 1 hourglass` confirmation on the client before even this code is called / executed, I think it can be removed completely

paglias: ah okay!

website/common/script/ops/buy/buyMysterySet.js
```diff
@@ -23,10 +23,6 @@ module.exports = function buyMysterySet (user, req = {}, analytics) {
     throw new NotFound(i18n.t('mysterySetNotFound', req.language));
   }
 
-  if (typeof window !== 'undefined' && !req.noConfirm && window.confirm) { // TODO move to client
```
TheHollidayInn, line 26: Why is this removed?

website/common/script/libs/getItemInfo.js
```diff
@@ -237,7 +237,7 @@ module.exports = function getItemInfo (user, type, item, officialPinnedItems, la
         notes: item.notes(language),
         value: item.value,
         currency: 'gold',
-        purchaseType: 'potions',
```
TheHollidayInn, line 240: Will this string change affect the client checks at all? I think these hard coded strings are used elsewhere

website/client/store/actions/shops.js
```diff
@@ -82,7 +82,7 @@ export function purchase (store, params) {
 
 export function purchaseMysterySet (store, params) {
   const user = store.state.user.data;
-  let opResult = buyOp(user, {params, noConfirm: true, type: 'mystery'});
```
TheHollidayInn, line 104: Why did you remove he noConfirm default?

TheHollidayInn:
This is looking good and working well for me.

website/common/script/ops/buy/buyMysterySet.js
negue: see discussion at https://github.com/HabitRPG/habitica/pull/9714/files/89f1ad2b07e0115ab494c3529518cb96ebc8d6d3..dd5870903f41d10531c4eca77c6ac4d6815d8055#diff-e92347dc0fb9e7ddffb5912653a501df

website/common/script/libs/getItemInfo.js
negue: it was changed to potion because `purchaseType` is assigned to `type` in 

```
makeGenericPurchase (item, type = 'buyModal', quantity = 1) {
      this.$store.dispatch('shops:genericPurchase', {
        pinType: item.pinType,
        type: item.purchaseType,
        key: item.key,
        currency: item.currency,
        quantity,
      });
```

then `type` is used for the op-method in `buy.js`


TL;DR no purchaseType is never used in a translation or something

paglias (approved):

paglias:
Marked as tested an approved, but at this point it'll be merged next week

SabreCat:
@paglias @negue I'm not 100% certain how the current conflict in `website/common/script/index.js` should be resolved here. Can one of you take a look?

PR #9723, opened 2017-12-16
Fixed pending damage display and nav size

TheHollidayInn:


Alys:
I'm still getting wrapping problems with this PR locally. The nav bar below was on one line with a browser window width of 1259. It went to two lines as you see below at about 1254. I couldn't get the cutover point narrowed down better than that.

The screenshot shows a reasonable maximum "size" for Hourglasses (2 digits), Gems (3 digits), and Gold (7 digits with a decimal point). Some players will have a lot more gold but I think will be in a minority small enough to not design for.

img

Alys:
Pending damage is working though!  I reckon this could be merged for that fix and then the nav sizing could be tweaked again later, if you want to handle it that way.

TheHollidayInn:
Eh, good point. I really should have approached this from a min/max perspective. Okay, let’s jsut merge this, then I can ask the designers about handling large numbers. Maybe we can introduce a K and M like the apps.

Issue #9043, opened 2017-09-16
dist client


billy-tutch:
{"success":false,"error":"Error","message":"ENOENT: no such file or directory, stat 'C:\\Users\\Administrator\\Desktop\\habitica\\dist-client\\index.html'"}

What's missing, i don't have a dist-client folder

paglias:
When are you getting this message?

billy-tutch:
When loading the localhost.

NPM Start runs fine, installation went through without any errors. I don't have a dist-client folder anywhere?

paglias:
I see, we've to update the docs. Basically you can run the client in a separate process running `npm run client:dev` and find it at `localhost:8080`, with the api at `localhost:3000` with automatic rebuilding when files change or build the client only once with `npm run client:build` and access it alongside the api at `localhost:3000` but it takes awhile to build it everytime and it's only meant for production usage

luanmuniz:
@paglias Closing the issue before updating the docs! :(
Just lost an hour looking for this

How can we update the docs?

paglias:
@luanmuniz the docs are up to date here http://habitica.wikia.com/wiki/Setting_up_Habitica_Locally#Run_Habitica

luanmuniz:
I had to run `npm run client:build`, not `npm run client:dev` to make it work. And actually keep rerunning it with `npm start` everytime i did a change =/

paglias:
@luanmuniz Try this:

- Open two terminal windows
- Into one run `npm start`
- Into the other `npm run client:dev`
- The second command will take some time for the initial build
- Then you'll have the website available at `localhost:8080` and the api at `localhost:3000`, but you can ignore that as all the api requests to `localhost:8080/api` are proxies to the api server at localhost:3000
- Any change made to /website files will not require any new command to be run, if you change a client file the site will reload automatically with the new changes, if you make a change to /website/common/locales instead you'll have to reload the site to see the changes

luanmuniz:
@paglias Here is what happen:

If i run `npm start` and `npm run client:dev`, there is no `dist-client` folder. So when i access `localhost:3000` to access the website, i receive a message 

```
2017-12-06T12:48:50.474Z - warn: Error: ENOENT: no such file or directory, stat '/vagrant/dist-client/index.html'
    at Error (native)
{ method: 'GET',
  originalUrl: '/',
  headers:
   { host: 'localhost:3000',
     connection: 'keep-alive',
[...]

PR #9724, opened 2017-12-16
Add client Dockerfile and change docker compose process

BradLugo:

Fixes are based on discussion from [#9043](https://github.com/HabitRPG/habitica/issues/9043)

### Changes

- Added another Dockerfile for the client side development
- Updated the docker-compose.yml file to accommodate the aforementioned Dockerfile
- Updated the version for the docker-compose.yml and docker-compose.dev.yml files

----
UUID: b20e27a5-e552-4247-9e91-65442a04f6d4

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,3 +1,10 @@
-web:
-  volumes:
-    - '.:/usr/src/habitrpg'
+version: "3"
+services:
+
+  client:
+    volumes:
+      - '.:/usr/src/habitrpg'
+
+  server:
+    volumes:
+      - '.:/usr/src/habitrpg'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,13 +1,36 @@
-web:
-  build: .
-  ports:
-    - "3000:3000"
-  links:
-    - mongo
-  environment:
-    - NODE_DB_URI=mongodb://mongo/habitrpg
+version: "3"
+services:
 
-mongo:
-  image: mongo
-  ports:
-    - "27017:27017"
+  client:
+    build: .
+    networks:
+      - habitica
+    environment:
+      - BASE_URL=http://server:3000
+    ports:
+        - "8080:8080"
+    command: ["npm", "run", "client:dev"]
+    depends_on:
+      - server
+
+  server:
+    build: .
+    ports:
+      - "3000:3000"
+    networks:
+      - habitica
+    environment:
+      - NODE_DB_URI=mongodb://mongo/habitrpg
+    depends_on:
+      - mongo
+
+  mongo:
+    image: mongo
+    ports:
+      - "27017:27017"
+    networks:
+      - habitica
+
+networks:
+  habitica:
+    driver: bridge


paglias:
Is it possible to have a single docker file that runs both the client and server?

BradLugo:
Yes we could but we'd be abusing containers at that point for the same reason we don't have Mongo in the same container as this project's source code. If you want other's opinion on the matter, [here's a DevOps StackExchange post](https://devops.stackexchange.com/questions/447/why-it-is-recommended-to-run-only-one-process-in-a-container). But if ya'll don't agree or don't see an issue with that, I can change it to one docker file

TheHollidayInn:
I support having two. That is the recommended way by docker and Docker Compose is there to help have a single command.

luanmuniz:
Can we have the same Dockerfile for both but overwrite the command in the docker-compose file? They are both the exact same container, the only thing that change is the command. That way would be just one file to maintain 

Reference: https://docs.docker.com/compose/compose-file/#command

BradLugo:
I just made a change but you're right, these containers would be the same if the EXPOSE statements weren't there (admittedly, the EXPOSE doesn't really do anything when using the docker-compose.yml). The reasons I'd argue for keeping them different are:

- It's not obvious that you need both containers to run the app code successfully (not including the database). This could include running unit tests without having to remember that the client side needs to have a different entry point
- If the client side and server side had different environmental dependencies that need to be resolved at build time instead of run time. For example, if we wanted to have just the client side artifact in the client container and just the server side artifact in the server container (as opposed to having both build artifacts in both containers)

All of that being said, I don't have a preference. I can change it if we want it that way

TheHollidayInn:
I can see the issue. We currently couple the client and the server together tightly. There is definitely a case to remove that coupling - which would require two environments, but that is out of scope.

lex111:
Hi all. 

This works fine for me, followed the instructions for the Docker on the wiki page, it did not work. Let's merge this already?

harakuri:
Hey all,

Really like the change! Quick question though: Should the mongo image be specified to version 3.4 in the compose file?

paglias:
I'm fine with specifying the mongo version but we're not using 3.4 in production. We're still on mongo 3.2, 3.4 works anyway, 3.6 will once we upgrade to mongoose 5 (which has some bug currently)

SabreCat:
Thanks for this, @BradLugo! Noted as credit toward Blacksmith tier 2!

Alys:
@BradLugo Thanks for this! Are there any changes needed to the Docker section of [Setting up Habitica Locally](http://habitica.wikia.com/wiki/Setting_up_Habitica_Locally)?

BradLugo:
@Alys Yes but nothing in the process changes. The only thing that needs to change is describing what's going on at the end of the Docker section. For example, I can see how "Habitica is available on port 3000 and mongo is on port 27017" can be misleading because the server is running on 3000 and the client is running on 8080. I don't know how y'all want to phrase that since the two are tightly coupled but it's mainly semantics. This documentation change also relates to the Vagrant change (#9732) so whatever we decide here, I can make that change in the Vagrant section as well

Alys:
@BradLugo Would you like to edit those sections in whatever way you think will be most helpful to people using Docker and Vagrant?

paglias:
I was able to install docker and the habitica image today but then I'm stuck when trying to run tests following the wiki instructions which are likely out of date. @BradLugo how do you run tests now?

Issue #9731, opened 2017-12-19
Equipment shop errors: some gear is mis-filed, some doesn't appear


Alys:
The Equipment shop is showing more problems than previously and I think it's a new bug from just the past few days.

Base and seasonal equipment for my current class appears correctly when that class is selected from the "Class" drop down.

Base and seasonal equipment for other classes also appears correctly when those classes are selected. BUT the base equipment also appears under "None" (it shouldn't appear there). 

Armoire gear that I've purchased and lost appears under "None" which is correct.

At least some other types of gear don't appear anywhere. For example, the Nameless Helm (head_special_2) doesn't appear anywhere after I've lost it.

There might be other problems too; I haven't done thorough tests.

This is high priority because you can't buy back items that don't appear there, and because any problems affecting seasonal gear are going to cause trouble when the Grand Gala starts this month.

Alys:
In addition, from kawunngg (ce935e7f-00e6-44bc-a44a-5280553810f0) in the Report a Bug guild:

"i can see all equipment of the other classes without being able to buy it back (i know this is intended). i can pin the items to my reward column, and if i have enought gold, i even get the message 'you bought xyz' but i don't lose gold and don't get the items. same thing with equipment of my own class i should not be able to get because i have to buy the beginner items first."

Alys:
fox_town says:
My gala gear is in my rewards column, but not in the market nor the seasonal shop?  Is that as intended? (User in the bug guild has unpinned something from their column and can't get it back, and I don't see a workaround.)
It's not under "None"

kawunngg:
> At least some other types of gear don't appear anywhere. For example, the Nameless Helm (head_special_2) doesn't appear anywhere after I've lost it.

after resetting my account all equipment won in quests don't show and i don't see my crystal armor anymore, so i guess all the gear i would get with contribution tiers would not be visible. other things like my "take this" set or the unconventional armor and subscriber items are still there.

Alys:
For the benefit of anyone working on this issue, I **think** the list of gear below will be sufficient to test all possibilities for lost equipment.  I've tried to choose a thorough selection that should test for all possibilities for code bugs, while not being too large to make testing difficult, but I might have missed an important example so if anyone can see anything that should be included, please comment here. Once this issue is sorted out, we can put the list into the wiki's Equipment page in an "Information for Developers" section to help with future testing.

[...]

PR #9739, opened 2017-12-20
Include seasonal gear in Market class lists

SabreCat:

### Changes

**Previously**, the gear section of the regular Market did not display seasonal class equipment. This proved a bit confusing, as users would expect to see currently buyable, class-specific gear there.
**Now**, Gold-buyable seasonal gear shows up in the Market under the expected class heading.

**Previously**, some special items did not show up in the Market if they were broken due to running out of Health. This affected, for example, "legendary" quest items and Take This challenge gear.
**Now**, these "classless" special items do show up in the Market. This is a partial fix for #9731.

### DIFF
--- a/website/common/script/libs/shops.js
+++ b/website/common/script/libs/shops.js
@@ -163,7 +163,7 @@ shops.getMarketGearCategories = function getMarketGear (user, language) {
           class: classType,
         },
       };
-      if (gearItem.specialClass === classType) return gearItem.canOwn(classShift);
+      if (gearItem.specialClass === classType && user.items.gear.owned[gearItem.key] !== false) return gearItem.canOwn(classShift);
     });
 
     category.items = map(result, (e) => {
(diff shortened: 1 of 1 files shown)


website/common/script/libs/shops.js
```diff
@@ -155,7 +155,16 @@ shops.getMarketGearCategories = function getMarketGear (user, language) {
...
-    let result = filter(content.gear.flat, ['klass', classType]);
+    let result = filter(content.gear.flat, function findClassGear (gearItem) {
+      if (gearItem.klass === classType) return true;
+      let classShift = {
+        items: user.items,
+        stats: {
+          class: classType,
+        },
+      };
+      if (gearItem.specialClass === classType) return gearItem.canOwn(classShift);
```
negue: this should also merge the `specialGear` https://github.com/HabitRPG/habitica/blob/c1bd762ff159112e5b735b387f3500af94c60280/website/common/script/libs/shops.js#L175
which adds the special gear that has been already owned


once both merged it should also check about the current season gear. Cause I don't think we should list gear that can't be purchased at the moment (which also never been owned)

SabreCat: OK, do you suggest chaining (or following up with) a Lodash `.merge` calling that function then?

This won't list gear that can't be purchased. `canOwn` takes care of that part.

Issue #8368, opened 2017-01-05
rounding inconsistencies


Alys:
This issue is to list inconsistencies in rounding numbers within the website and between the website and apps, and to discus which ones we want rounded up, or rounded down, or rounded off. It's not a complete list yet - if you notice any others, please comment here.

Items have been marked complete if they are correct or if an issue or PR has been created for them (even if the issue has not been fixed yet).

- Player's HP:
   - **desired rounding**: round **down** for health above 1 so that you don't think you have more health than you really have **but** we can't allow the situation where it looks like you have 0 health when you actually have slightly more than that because that produces bug reports ("I have zero health but I haven't died"), so round **up** for health less than 1
   - **desired decimal places**: zero decimal places for health greater than 1 but one decimal place for health less than 1.
  - examples of desired behaviour: 
      - 49.9 -> 49
      - 9.9 -> 9
      - 1.9 > 1
      - 0.99 -> 1
      - 0.9 -> 0.9
      - 0.19 -> 0.2
      - 0.09 -> 0.1
      - 0.0001 -> 0.1 
      - 0 -> 0
      - less than zero -> rounded down, showing negative sign and always zero decimal places (e.g., -0.1 -> -1)
  - [x] website: correct using this code: https://github.com/HabitRPG/habitica/blob/3b4b459e6861fd5e1c1abc5471fff95007559ff1/website/client/components/memberDetails.vue#L237-L243
  - [x] **Android app: rounds up** (49.9 -> 50, etc): https://github.com/HabitRPG/habitica-android/issues/914
  - [x] **Android widget: not tested, unlikely to be correct since app needs changes**: included in issue above
  - [ ] **iOS app: unsure, needs testing**: mentioned in Android issue above but a new issue should be created if needed

- Boss HP
  - **desired: rounds up** so that if you're aiming to kill the boss, you don't underestimate the damage you need to do.
      - However it would be nice if there was an on-hover effect on the website to show the actual value (I've seen two bug reports in the past couple of months related to this).
  - [ ] **website: truncates to two decimal places**  https://github.com/HabitRPG/habitica/pull/9749
  - [x] **Android: rounds down** https://github.com/HabitRPG/habitica-android/issues/915
  - [x] **iOS: rounds down** https://github.com/HabitRPG/habitica-ios/issues/442

- Pending damage to boss:
  - **desired**: shows one decimal place and **rounds down** so that if you're aiming to kill the boss, you don't overestimate the damage you have already done.
[...]

PR #9749, opened 2017-12-29
improves rounding for boss hp and player pending damage - partial fix for #8368

Alys:
partial fix for #8368 

### Changes
- The boss's current HP is rounded up so that when you're aiming to kill the boss, you don't underestimate the damage you need to do. 
- The player's current pending damage is rounded down so that you don't overestimate the damage you have already done.
- Adds comments about that so that future coders don't think it's an error that one uses ceil and the other uses floor.
- Adds translatable string for "pending damage"
- Fixes typo in an existing comment

screenshot_2017-12-29_23-01-52

That screenshot is from this data:
```
From party's group document:

	"quest.progress": {
		"hp": 30.011,
		"collect": {}
	},

From user document:
	"party.quest.progress": {
		"collectedItems": 0,
		"collect": {},
		"down": 0,
		"up": 50.999
	}
```


website/client/components/groups/questSidebarSection.vue
```diff
@@ -54,10 +54,12 @@ div
...
                 .col-6
                   span.float-left
-                    | {{parseFloat(group.quest.progress.hp).toFixed(2)}} / {{parseFloat(questData.boss.hp).toFixed(2)}}
+                    | {{ Math.ceil(parseFloat(group.quest.progress.hp) * 100) / 100 }} / {{ parseFloat(questData.boss.hp).toFixed(2) }}
+                    // current boss hp uses ceil so you don't underestimate damage needed to end quest
                 .col-6(v-if='userIsOnQuest')
-                  // @TODO: Why do we not sync quset progress on the group doc? Each user could have different progress
-                  span.float-right {{parseFloat(user.party.quest.progress.up).toFixed(1) || 0}} pending damage
+                  // @TODO: Why do we not sync quest progress on the group doc? Each user could have different progress.
+                  span.float-right {{ Math.floor(parseFloat(user.party.quest.progress.up) * 10) / 10 }} {{ $t('pendingDamage') }}
```
paglias: could this use the floor filter we already have in client/filters/floor? Maybe with an optional parameter to use *10 instead of *100

Also I don't think that parseFloat is necessary, it's already a number

paglias (changes requested):

paglias:
@Alys do you want me to finish this?

Alys:
@paglias if you you want to, sure. Otherwise I'll make time for it when I
can. Sorry there's been too much bug support lately. :(

lemoness:
@Alys What is the status on this?

Alys:
I need to modify a line that I changed to use a filter instead and adjust the filter to let it do the same thing as that line. It's still on my task list just hasn't been as high priority as other things. :(

website/client/components/groups/questSidebarSection.vue
Alys: @paglias  Since I made this PR, the `pendingDamage` locales text has been implemented to contain the damage variable in it:
https://github.com/HabitRPG/habitica/blob/70500d7c98d67f1edd4a91aa26e176c5952ebae8/website/common/locales/en/quests.json#L125

So I could change the pending damage line in my code to be like this:
`span.float-right {{ $t('pendingDamage', {damage: user.party.quest.progress.up}) }}`
and use the floor function in it (how would I do that though?)

Or I could add a different text string, `pendingDamage1` or something, and use code like this:
`span.float-right {{ user.party.quest.progress.up | floor(10) }} {{ $t('pendingDamage1') }}`
Which would you prefer?

paglias: Let's use the second option

Alys: That's done - the floor function is now used instead of calling Math.floor. I also made the same change to the user's pending damage in the World Boss box in the Tavern.

This is ready for review again.

PR #8930, opened 2017-08-08
Fixes stealth to have an upper limit.

guyguy2001:
Fixes #5279. I didn't close it so it could be used to discuss it- will change the title if it shuold be.

First of all I'm sorry it took me until not to send a PR, I was in a seminar so I couldn't submit it.

I made 3 versions of it, which is how I refer to it in the code:
0- The current option.
1- TNychka's idea in the comment to #5279.
2- My idea, which uses isDue. I used the number of casts as a perception buff, but it could be used in a similiar way to 1. I think using isDue is a bad idea, because it would be very confusing to players that they get less defense the more dailies they tick. (Also I don't know how the "number of avoided dailies" on mouseHover would be phrased. note- I couldn't get it to show the current number anyway as I didn't understand how to access the daily list).

Here is a desmos graph I made for the different options:
https://www.desmos.com/calculator/txwic7jqaa
s=the number of casts.
h=halfway.
p=perception.
m=the number of dailies.
d=the number of due dailies.

I couldn't get f1 to be a function of s, so just switch the active one when you change s.

From what it seems, the tasks aren't explicitely sorted by difficulty. What are they sorted by then? If not difficulty, isn't it exploitable? (ie the best option is quancky and not what was intended, like deleting tasks and recreating them in order of difficulty, or rearranging them by difficulty).

PR #9751, opened 2017-12-29
Fixes stealth to have an upper limit

### LINKED ISSUE #5279: Stealth Not Acting As Planned
According to the DDT's Task Statistics, I have 98 Dailies. 36 of them are inactive today.

My Perception is 358. However, Stealth will allow me to avoid **55** Dailies.

**Reasons why this is not as planned:**

At max, one cast of Stealth should only allow me to skip half of my Dailies, or 49.

Furthermore, as Stealth now no longer counts inactive tasks, it really should only be allowing me to skip 50% of my _active_ tasks.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/17500408-stealth-not-acting-as-planned?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

guyguy2001:
[EDIT] Repost of https://github.com/HabitRPG/habitica/pull/8930 from a different repository.
Fixes #5279.

I made 3 versions of it, which is how I refer to it in the code:
0- The current option.
1- TNychka's idea in the comment to #5279.
2- My idea, which uses isDue. I used the number of casts as a perception buff, but it could be used in a similiar way to 1. I think using isDue is a bad idea, because it would be very confusing to players that they get less defense the more dailies they tick. (Also I don't know how the "number of avoided dailies" on mouseHover would be phrased. note- I couldn't get it to show the current number anyway as I didn't understand how to access the daily list).

Here is a desmos graph I made for the different options:
https://www.desmos.com/calculator/txwic7jqaa
s=the number of casts.
h=halfway.
p=perception.
m=the number of dailies.
d=the number of due dailies.

[NEW]: currently x is set to the amount of due dailies you have. You can write `d=4` instead of `d=x` and put another value to be x.

I couldn't get f1 to be a function of s, so just switch the active one when you change s.

From what it seems, the tasks aren't explicitely sorted by difficulty. What are they sorted by then? If not difficulty, isn't it exploitable? (ie the best option is quancky and not what was intended, like deleting tasks and recreating them in order of difficulty, or rearranging them by difficulty).

[NEW]: I am sorry it took me so long to fix that. I got finals coming up and I had some tests and a lot of other stuff going on, so I completely forgot about this fix. Not good reasons at all, but that's what happened. I didn't have time to run it, however I did change what was needed to the new vue semantics. I will make sure everything runs correctly on Sunday.

Also, it seems that sometimes the code is spelled with dailies, and sometimes it is spelled like dailys. Should I change the wrong ones?

EDIT-  could you change the title to `Fixes stealth to have an upper limit.` like the original one? I forgot to do so myself.

### DIFF
--- a/website/client/components/tasks/spells.vue
+++ b/website/client/components/tasks/spells.vue
@@ -224,10 +224,10 @@ export default {
       if (skill === 'frost' && this.user.stats.buffs.streaks) {
         return true;
       }
-      // @TODO: Implement
-      // } else if (skill === 'stealth' && this.user.stats.buffs.stealth >= this.user.dailys.length) {
-      //   return true;
-      // }
+
+      else if (skill === 'stealth' && (this.user.stats.buffs.stealth >= this.user.dailys.length /*0*/ && true || /*1*/false && this.user.stats.buffs.stealth >= this.user.dailys.length / 2)) {
+        return true;
+      }
 
       return false;
     },
@@ -239,14 +239,34 @@ export default {
       } else if (skill.key === 'stealth' && this.spellDisabled(skill.key)) {
         notes = this.$t('spellRogueStealthMaxedOut');
       } else if (skill.key === 'stealth') {
-        notes = this.$t('spellRogueStealthDaliesAvoided', {
-          originalText: notes,
-          number: this.user.stats.buffs.stealth,
-        });
+        if(true) { //0,1
+          notes = this.$t('spellRogueStealthDaliesAvoided', {
+            originalText: notes,
+            number: this.user.stats.buffs.stealth,
+          });
+        } else { //2
+            notes = this.$t('spellRogueStealthDaliesAvoided', {
+                originalText: notes,
+                number: this.diminishingReturns(this.user._statsComputed.per * this.user.stats.buffs.stealth * 3, this.countDues() / 2, 55),
+            });
+        }
+        
       }
 
       return notes;
     },
+    countDues() {
+      var dues = 0;
+      this.user.dailys.forEach((task) => {
+        if (task.isDue)
+          dues++;
+      });
+    },
+    diminishingReturns(bonus, max, halfway) {
+      if (!halfway) halfway = max / 2;
+      return max * (bonus / (bonus + halfway));
+    },
+
     questProgress () {
       let user = this.user;
       if (!user.party.quest) return 0;
--- a/website/common/script/content/spells.js
+++ b/website/common/script/content/spells.js
@@ -196,9 +196,22 @@ spells.rogue = {
     lvl: 14,
     target: 'self',
     notes: t('spellRogueStealthNotes'),
-    cast (user) {
-      if (!user.stats.buffs.stealth) user.stats.buffs.stealth = 0;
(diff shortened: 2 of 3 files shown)


Alys:
@guyguy2001 Thanks for this! I've edited the title as you requested. No worries about the delay! We understand when real life gets in the way and finals are definitely more important, so that is a good reason!

"dailys" is not necessarily wrong. It's an incorrect English spelling of course but can make sense in code. For example, if you're grepping code to find variables with "daily" in them, you'd also want to find the pluralised version, which "dailys" allows but "dailies" doesn't. There's also consistency with the four task types having singular and plural versions created by adding "s" in every case. My personal feeling is that any occurrences of "dailys" shouldn't be changed to "dailies" unless there's a good reason apart from just "correcting" the spelling, and when adding new code sometimes it might be appropriate to use "dailys" instead of "dailies".

From what I've seen, tasks are processed in the cron code in the default order in the database which seems to be order of creation, although I wouldn't consider that to be fully reliable and it might change in future. We're not concerned about players exploiting this by creating their most difficult tasks first - if they want to try to ensure that Stealth protects their hardest Dailies, that's their decision. As a minor point, personally I'd actually like us to go back to how it used to be when Dailies were processed by cron (and hence by Stealth) in the same order that they appeared in the user's task list because that made Stealth's behaviour more predictable, but that's not something that needs to be considered now.

I haven't looked at your other comments in detail or your code, but one of us will when we can. Thanks again for this - you've obviously put a lot of thought into it!

TheHollidayInn:
Thanks for the PR! Can you add tests to cover the scenarios in `libs/cron`?

guyguy2001:
I'll do that in the next few days; I got some other things I need to finish, but it should be ready soon.

paglias:
Updated the tags to `needs tests`

lemoness:
@guyguy2001 Hi there! Are you still working on adding these tests?

guyguy2001:
Sorry, I couldn't find time for it yet (Didn't realize I didn't touch it for a month...).
I will try to do it in the next few days, but if I won't I'll just give it to someone else.

lemoness:
@guyguy2001 Wonderful! Keep us posted :)

guyguy2001:
I will probably have time for it after next week (as I'll be entering the passover school-break), however until them I doubt I'll get to it. If anyone else wants to do tests, I'll give it to them. If not, I'll finish it then.

lemoness:
@guyguy2001 Hi there! How is this fix going?

guyguy2001:
In the end I didn't have time for it, both because of the amount of things I have to finish and because of poor time management; If someone else could add the tests, I'm willing to re-open it.
If not, I'll add it when I'll have less things that I need to study for / finish, but I can't promise on when. If that will be needed however I will do as much as I can to finish it in a reasonable time.

I'm really sorry that I messed up with doing it in a reasonable time; I'll make sure that I'll take the next habitica project when I have less other projects on my head.

lemoness:
@guyguy2001 Thank you very much for letting us know! Good luck with your other projects.

PR #9752, opened 2017-12-29
refund gems when deleting a "Public Challenge" (Tavern challenge) - fixes #7379

### LINKED ISSUE #7379: Gems are not refunded when deleting a Tavern challenge
_edit by Alys 2017-04-07:_

When a Tavern challenge is deleted by either the challenge owner or a moderator, the gem prize must be returned to the challenge owner.

Alys:
Fixes https://github.com/HabitRPG/habitica/issues/7379

### Changes
* refunds a challenge prize to the challenge owner when deleting a "Public Challenge" (Tavern challenge)

### DIFF
--- a/website/server/models/challenge.js
+++ b/website/server/models/challenge.js
@@ -7,7 +7,6 @@ import * as Tasks from './task';
 import { model as User } from './user';
 import {
   model as Group,
-  TAVERN_ID,
 } from './group';
 import { removeFromArray } from '../libs/collectionManipulators';
 import shared from '../../common';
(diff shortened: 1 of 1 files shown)


paglias:
Code is 👍 do we want to add a test here?

Alys:
I'm thinking that we don't need a new test because we have this test for the generic case of any challenge being deleted:
https://github.com/HabitRPG/habitica/blob/9545f692efdbea7ff4980c0c7ce20534585363e2/test/api/v3/integration/challenges/DELETE-challenges_challengeId.test.js#L70-L78

Now that there's no special code for the Tavern, I don't think there's any reason to have a special test for it. All it would be checking for is whether we reintroduced the `if` statement for the Tavern, which isn't likely.

But if you'd feel more comfortable with a test for it, tell me and I'll write it.

TheHollidayInn:
I think we are fine without the test. You removed a branch (which was untested) but if it was tested, you would be removing a test in this PR. So we are covered.

paglias:
👍

PR #9762, opened 2018-01-03
Fix search in inventory and equipment by making it case insensitive

paglias:


Alys (approved):
Tested locally. Works in both Inventory > Items and Inventory > Equipment. All good. Since this is such a simple fix, I think we can merge immediately rather than waiting for Monday.

(FYI for the curious, the shops don't have the same bug so this fix isn't needed for them.)

PR #9766, opened 2018-01-04
Unpin Regained Items from Enchanted Armoire

### LINKED ISSUE #9746: Enchanted Arrmoire equipment lost to death doesn't automatically unpin when regained
As reported by Il Mago Azzurro ( c073342f-4a65-4a13-9ffd-9e7fa5410d6b ) in the Report a Bug guild:

"when you have an Enchanted Armoire item pinned to rewards and you gain it when buying the EA, the pinned item doesn't unpin on its own"

labels updated

SheepD:
Fixes #9746 

### Changes
Unpin regained items from enchanted armoire

----
UUID: 0e321fa4-c6b0-45e2-b80a-d007f10bfad2

### DIFF
--- a/website/common/script/ops/buyArmoire.js
+++ b/website/common/script/ops/buyArmoire.js
@@ -9,6 +9,7 @@ import {
   NotAuthorized,
 } from '../libs/errors';
 import randomVal from '../libs/randomVal';
+import { removeItemByPath } from './pinnedGearUtils';
 
 // TODO this is only used on the server
 // move out of common?
@@ -55,6 +56,8 @@ module.exports = function buyArmoire (user, req = {}, analytics) {
     if (count.remainingGearInSet(user.items.gear.owned, 'armoire') === 0) {
       user.flags.armoireEmpty = true;
     }
+
+    removeItemByPath(user, `gear.flat.${drop.key}`);
 
     armoireResp = {
       type: 'gear',


website/common/script/ops/buyArmoire.js
```diff
@@ -56,6 +58,16 @@ module.exports = function buyArmoire (user, req = {}, analytics) {
       user.flags.armoireEmpty = true;
     }
 
+    const officialPinnedItems = getOfficialPinnedItems(user);
+    if (isPinned(user, drop, officialPinnedItems)) {
+      const dropInfo = {
+        item: drop,
```
negue: no need for `item` if you have the correct `type` and `path`

website/common/script/ops/buyArmoire.js
```diff
@@ -56,6 +58,16 @@ module.exports = function buyArmoire (user, req = {}, analytics) {
...
+        type: 'gear',
```
negue: with `gear`(gem purchasable) it doesn't seem to work for me, `marketGear` is the correct one (purchase by gold)

negue:
Thanks for your contribution, just a little fix and then its good to merge

SheepD:
thanks for the info @negue 
i've update the PR with the suggested changes.

though for `togglePinnedItem`, is it better to just use `item: drop` instead of `type` and `path`? using type and path seems to imply it has to search for the item first instead of just using the item that was determined from the drop

negue:
You're right, but you'd still need to pass the `type` too. 

Could you test if it still work with only `type` and `item`?

SheepD:
@negue you are correct, the `type` field is still needed. and upon further inspection of the code,  passing in `type` and `item` runs faster(as it runs through far less conditionals than if we passed path and type instead)

I've updated the code to use type and item instead.

negue:
I've tested it again, but it still doesn't work (the change to `item` and `type` is still correct) but it never passes `isPinned(user, drop, officialPinnedItems)` because there the `drop` needs a `path` property, which it normally doesn't have. (only after calling `getItemInfo`)

The best solution would be just to call `removeItemByPath (user, path)` (also from `pinnedGearUtils`) this way we can skip the isPinned and toggle methods and just remove it.

If an item isn't pinned it just continues, no exceptions here.


Sorry for going back and forth 😅

SheepD:
i was hoping to have it play a bit safe to prevent possible errors hence the safety conditional statement.

anyway I have updated the code as suggested.

negue:
Works, thank you! 🎉

SabreCat:
Thanks for this, @SheepD, and welcome to the ranks of the Blacksmiths at tier 1! It will run in a staging environment for a few days before going live on the public site.

Note that subsequent contributor tiers require increasing amounts of effort from one to the next, so tier 2 may require multiple PRs or a larger PR to attain. But keep helping out and we will express our gratitude accordingly!

PR #9773, opened 2018-01-08
Updated plan updated date if user has cancelled

TheHollidayInn:
If a user with a cancelled subscription receives a gift, this PR will ensure their updated field is updated.

### DIFF
--- a/test/api/v3/unit/libs/payments.test.js
+++ b/test/api/v3/unit/libs/payments.test.js
@@ -153,6 +153,16 @@ describe('payments/index', () => {
         expect(recipient.purchased.plan.dateUpdated).to.exist;
       });
 
+      it('sets plan.dateUpdated if it did exist but the user has cancelled', async () => {
+        recipient.purchased.plan.dateUpdated = moment().subtract(1, 'days').toDate();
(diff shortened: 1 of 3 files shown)


test/api/v3/unit/libs/payments.test.js
```diff
@@ -153,6 +153,16 @@ describe('payments/index', () => {
...
+      it('sets plan.dateUpdated if it did exist but the user has cancelled', async () => {
+        recipient.purchased.plan.dateUpdated = moment().subtract(1, 'days').toDate();
+        recipient.purchased.plan.dateTerminated = moment().subtract(1, 'days').toDate();
+        recipient.purchased.plan.customerId = 'testing';
+
+        await api.createSubscription(data);
+
+        expect(moment(recipient.purchased.plan.dateUpdated).date()).to.eql(moment().date());
+      });
+
```
Alys, line 165: We'd also benefit from a test for the situation where there's a dateUpdated value but no dateTerminated or customerId. That's the situation for the current bug.

TheHollidayInn: Gotcha. I was going based off this `if ( !activeSubscription || !plan.dateUpdated) plan.dateUpdated = today;` But I can see that `activeSubscription` includes malformed plan fields all together.

PR #9780, opened 2018-01-10
Added redirects for old url styles

### LINKED ISSUE #9586: old-style URLs to guilds and challenges should be automatically redirected to new-style URLs
On the old website, URLs to guilds and challenges were like this:
https://www.habitica.com/#/options/groups/guilds/a29da26b-37de-4a71-b0c6-48e72a900dac
https://www.habitica.com/#/options/groups/challenges/e2816060-cb5c-46b3-9763-adf2ea20a72b

The new website uses URLs like this:
https://www.habitica.com/groups/guild/a29da26b-37de-4a71-b0c6-48e72a900dac
https://habitica.com/challenges/e2816060-cb5c-46b3-9763-adf2ea20a72b

We want automatic redirection from old to new. Other old URLs could be redirected too but those ones are the most important.

Some notes from paglias:

The redirect can't be done on the webserver; it has to be done fully client side, partly because the site is now routed only on the client (only the API is routed from the server) but mainly because the hash part (`#...`) is never sent to the server but is only accessible from JavaSscript in the browser.

Going to `/#/test` is recognized by Vue as a route change towards this:
```
{
    "name": "tasks",
    "meta": {},
    "path": "/",
    "hash": "#/test",
    "query": {},
    "params": {},
    "fullPath": "/#/test",
    "matched": [
        {
            "path": "",
            "regex": {
                "keys": []
            },
            "components": {},
            "instances": {},
            "name": "tasks",
            "meta": {},
            "props": {}
        }
    ]
}
```
So we can intercept and redirect those requests though the hash part is simply matched as the `tasks` page so it'll require some manual matching of paths.

TheHollidayInn:
Fixes #9586

### DIFF
--- a/website/client/router.js
+++ b/website/client/router.js
@@ -316,6 +316,32 @@ router.beforeEach(function routerGuard (to, from, next) {
     return next({name: 'tasks'});
   }
 
+  // Redirect old guild urls
+  if (to.hash.indexOf('#/options/groups/guilds/') !== -1) {
+    const splits = to.hash.split('/');
+    const guildId = splits[4];
+
+    return next({
+      name: 'guild',
+      params: {
+        groupId: guildId,
+      },
+    });
+  }
+
+  // Redirect old challenge urls
+  if (to.hash.indexOf('#/options/groups/challenges/') !== -1) {
+    const splits = to.hash.split('/');
+    const challengeId = splits[4];
+
+    return next({
+      name: 'challenge',
+      params: {
+        challengeId,
+      },
+    });
+  }
+
   Analytics.track({
     hitType: 'pageview',
     eventCategory: 'navigation',


Joseph67843:
One suggestion for this one and #9586 -
 
There are a number of other old-style URLs that broke with the redesign. (All links below are preceded by habitica.com

The hardest ones are /#/options/profile/achievements, /#/options/profile/stats, /#/options/profile/avatar - I don't know of any link to load those screens in the new web client, and have left those in the wiki.
 
The other broken addresses that I know of (I've swapped most of these from the wiki, but it's possible there are external links out there):

Existing URL | Replacement
--- | ---
/#/options/groups/tavern | /groups/tavern
/#/options/settings/api | /user/settings/api
/#/options/settings/settings | /user/settings/site
/#/options/inventory/mounts | /inventory/stable
/#/options/inventory/pets | /inventory/stable
/#/options/settings/promo | /user/settings/promo-code
/#/options/inventory/drops | /shops/market
/#/options/inventory/seasonalshop | /shops/seasonal
/#/options/inventory/equipment | /inventory/equipment
/#/options/settings/subscription | /user/settings/subscription
/#/options/inventory/quests | /shops/quests (I think!)
/#/options/groups/hall/patrons| /hall/patrons
/#/options/groups/hall/heroes | /hall/contributors

PR #9781, opened 2018-01-10
Pushed zindex of progress bar above modals

### LINKED ISSUE #9464: Sync bar is not visible for Record Yesterday's Activity screen (RYA)
When you're completing Dailies in the Record Yesterday's Activity (RYA) screen, there's no way to tell when each tick mark has been synced to the server and so you can take unfair damage from some of your clicks not being recorded if you click too quickly.

~Possible solutions include making the normal sync bar that appears at the top of the website visible when the RYA screen is up, or adding a customised sync bar to the RYA screen.~

The solution is to have the sync bar above the overlay in the stack (as decided in the discussion below).

~Labels updated. I've marked this as needing design. @Tressley~

TheHollidayInn:
Fixes #9464

### DIFF
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -101,6 +101,11 @@ div
     opacity: 1 !important;
     background-color: rgba(67, 40, 116, 0.9) !important;
   }
+
+  /* Push progress bar above modals */
+  #nprogress .bar {
+    z-index: 1041;
+  }
 </style>
 
 <script>
--- a/website/client/components/snackbars/notifications.vue
+++ b/website/client/components/snackbars/notifications.vue
@@ -10,7 +10,7 @@
     right: 10px;
     top: 65px;
     width: 350px;
-    z-index: 999;
+    z-index: 1041; // 1041 is above modal backgrounds
   }
 </style>
 


paglias:
Can we also push error notifications above modals? Or maybe all snackbars

TheHollidayInn:
Sure!

josteins1:
Hi, I left a suggestion in trello and was told to mention it here in the relevant pull request by @Alys .
It seems this have accidentally pushed up the snackbar to cover up the dropdown menus. https://github.com/HabitRPG/habitica/commit/8b5933177a172948169d9a54f42c2dd970620772 This commit was supposed to change that. Idk if best approach is to either increase or decrease the z-value of the things that's colliding.
Snackbar is set at z-index value 1041
Dropdown is set at z-index value 1000

TheHollidayInn:
Ah, sorry about that. I think if we want the drop downs higher than the snackbar we should move them up. We want the snackbars to be higher than the modals.

josteins1:
No worries. Could you take a look at it? I have a feeling you know this code better than me. Should be quick to do and I don't remember where the dropdown css is located rn.

paglias:
I have a feeling that snackbars should be above dropdowns, why do you want them below?

Alys:
It drives you wild if you want to use a menu while the snackbars are getting in your way. :) It is only a couple of seconds until they disappear or you can make them disappear immediately by clicking on them, but it's still an annoyance.

I don't think that having the menus block the snackbars is a problem because if you wanted to read them, you wouldn't be using the menu.

paglias:
Pinging @Tressley about this

Il 23 Gen 2018 08:50, "Alys" <notifications@github.com> ha scritto:

Tressley:
Menus should be above snackbars. An example of @Alys frustration would be trying to use the User menu with a queue of snackbars completing.

image

In addition -- I think I can clear up some of the confusion in terminology being used. What we actually want is the snackbars above the **overlay** of the modal, not the modal itself.

josteins1:
Yes exactly! That's the haunting picture I'm faced with every day. Now I might be wrong, but it should be as easy as to locate the css of the user menu and increase the z-index to higher than 1041. Now I've only used github once before and it would probably be faster if someone could take the responsibility to do this.

TheHollidayInn:
I'm on it :D I'll open a separate issue and push a fix.

PR #9791, opened 2018-01-12
Fixed update query to revert challenge tags when broken

### LINKED ISSUE #7569: Challenge Tags not Reverting to Ordinary Tags after Challenge Ends
General Info
- UUID: be317a32-c027-4032-bbd5-25ccb21eb912
- Browser: Google Chrome Version 50.0.2661.102 m
- OS: Windows 10, 64-bit
### Description

When a challenge I'm in ends, the challenge tag is still sorted into the Challenge tags section of the tags dropdown. It should be converted to an regular tag and sorted appropriately.

TheHollidayInn:
Fixes #7569

### DIFF
--- a/website/server/models/challenge.js
+++ b/website/server/models/challenge.js
@@ -340,7 +340,7 @@ schema.methods.closeChal = async function closeChal (broken = {}) {
     // Set the challenge tag to non-challenge status and remove the challenge from the user's challenges
     User.update({
       challenges: challenge._id,
-      'tags._id': challenge._id,
+      'tags.id': challenge._id,
     }, {
       $set: {'tags.$.challenge': false},
       $pull: {challenges: challenge._id},
--- a/test/api/v3/integration/challenges/POST-challenges_challengeId_winner_winnerId.test.js
+++ b/test/api/v3/integration/challenges/POST-challenges_challengeId_winner_winnerId.test.js
@@ -149,13 +149,19 @@ describe('POST /challenges/:challengeId/winner/:winnerId', () => {
 
       await sleep(0.5);
 
-      let tasks = await winningUser.get('/tasks/user');
-      let testTask = _.find(tasks, (task) => {
+      const tasks = await winningUser.get('/tasks/user');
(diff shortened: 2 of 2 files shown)


website/server/models/challenge.js
```diff
@@ -340,7 +340,7 @@ schema.methods.closeChal = async function closeChal (broken = {}) {
     // Set the challenge tag to non-challenge status and remove the challenge from the user's challenges
     User.update({
       challenges: challenge._id,
-      'tags._id': challenge._id,
+      'tags.id': challenge._id,
```
Alys, line 343: This update query will fail to remove the challenge from the user's challenges if the user has deleted the challenge tag. That's something that users are able to do in case they prefer to use their own tag system and don't want the challenge tags cluttering their tag list.

paglias: Can we just remove the tags.id part of the query here?

Alys: We'd then be executing this for the case where the challenge tag had already been deleted:
`$set: {'tags.$.challenge': false},`
I.e., the challenge tag would be recreated, which would be a bug from the user's perspective. I'm on the fence about which is the worse bug. They're both pretty minor in my opinion. 

I could put up with the at worst very rare bug reports we'd get if the challenge tag was recreated. We might well never hear about it from any users. I'd like a comment in the code though just documenting that this situation could happen. I think that would be an easier "fix" than having two updates.

paglias: In that case we may simply split this update into two: one that remove the challenge from user.challenges, the other that removes the tag where it exists

TheHollidayInn: If the user updates the tag then it would not be a challenge tag right? Then we don't have the issue this PR fixes

paglias: Can you explain? What do you mean by `if the user updates`?

TheHollidayInn: Oh nvm. I see that the issue is the users challenge array, not the tags.

PR #9792, opened 2018-01-12
Added analytics to drop

### LINKED ISSUE #7581: Add back drop tracking for analytics
### Description
#7547 moved some drop-related code around in such a way that we no longer send drop data to our analytics providers. It's not one of our top needs, data-wise, but it would still be nice to have again. We should implement it in the server code for drops instead of on the client.

TheHollidayInn:
Fixes #7581

### DIFF
--- a/website/common/script/fns/randomDrop.js
+++ b/website/common/script/fns/randomDrop.js
@@ -25,7 +25,7 @@ function trueRandom () {
   return Math.random();
 }
 
-module.exports = function randomDrop (user, options, req = {}) {
+module.exports = function randomDrop (user, options, req = {}, analytics) {
   let acceptableDrops;
   let drop;
   let dropMultiplier;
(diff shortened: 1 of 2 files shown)


Alys:
This is tracking drops as an 'acquire item' event. If that's what's needed for analytics reports to be run as required, I accept it, but if it's not essential, can I ask that it be given a different event type? When players report bugs with items they've tried to buy or with missing gems, I use the 'acquire item' events to track what they've bought - this means clicking down through all their 'acquire item' events until I've collected the information I need. If a lot of those events are from drops, it's going to increase the time it takes to analyse their actions.

If that's a necessary consequence of whatever @veeeeeee and @SabreCat need for tracking drops, I can cope with the extra effort, but if it's not necessary it would help a lot with troubleshooting if drops were tracked separately.

TheHollidayInn:
I just added the old tracking event back, and I'm pretty sure the type can be changed. What do you suggest we change it to? I will wait for confirmation before changing, though.

SabreCat:
Relabeling an existing event deserves some forethought (because there will be a breakpoint in the data), but since that part of the analytics has been inactive for a while, it's safe to reboot it under a new heading.

Alys:
Thanks @SabreCat!

I'd suggest a type like "dropped item".

PR #9803, opened 2018-01-14
Purchase Card: fixes

### LINKED ISSUE #9484: Card purchase should have confirmation notification + Card purchase modal should not reopen itself
When casting a card on someone, no confirmation is given ~and gold doesn't decrease until the page is synced or refreshed~. There should be a confirmation notification in the corner, ~and the gold should be subtracted immediately~.

_Edit by Alys 2018-01-13 to summarise the comments below:_

The desired fix for this issue is both of these:
1. When you buy a card, you should see a notification saying "You sent a card to [User Name]". It should be a snack notification (the kind that appears in the top right of the website and then disappears quickly).
2. The buy modal should NOT reopen itself as soon as the first purchase has been finished.

I've crossed out the parts about gold not being subtracted because that bug has now been fixed.

We need a fix for this to be live before Valentine's Day because card purchases will be happening often then, so someone should try to fix this **by the end of January** to give time for testing, review, and deployment.

negue:
fixes #9484

- Prevent showing `buy-modal` again
- Show notification after buying

Also fixed the class-badge in member-selection (was bugging me^^)


Alys:
I've marked this and the associated issue as "priority: important" because we'll want to have this working on the production site before Valentines day, when a lot more players than usual will be sending cards. Previous card-sending event(s) since the website was redesigned have caused a lot of confusion and bug reports so this PR will be a welcome improvement.

website/client/app.vue
```diff
@@ -399,6 +408,7 @@ export default {
       if (this.selectedSpellToBuy.pinType === 'card') {
         const newUserGp = castResult.data.data.user.stats.gp;
         this.$store.state.user.data.stats.gp = newUserGp;
+        this.text(this.$t('sentCardToUser', { userName: member.profile.name }));
```
Alys: Could this be changed from `userName` to `profileName` here and in the translatable string? That would avoid any uncertainty when someone is viewing the locales file about whether it's the profile name or login name. Sorry about this - I realise I wrote "User Name" in the issue. I should have been using better terminology.

website/common/locales/en/spells.json
```diff
@@ -77,5 +77,6 @@
   "challengeTasksNoCast": "Casting a skill on challenge tasks is not allowed.",
   "groupTasksNoCast": "Casting a skill on group tasks is not allowed.",
   "spellNotOwned": "You don't own this skill.",
-  "spellLevelTooHigh": "You must be level <%= level %> to use this skill."
+  "spellLevelTooHigh": "You must be level <%= level %> to use this skill.",
+  "sentCardToUser": "You sent a card to <%= userName %>"
```
Alys: Could the new string be moved to the `generic.json` file, just under the `Cards` line (about liine 175)? That would keep it with similar strings which would help us if we were modifying them all in some future change.

Alys (commented):
I've tested this locally and it's working just the way we want it to! Thank you!

I haven't done a full review because I haven't actually learnt Vue properly yet :scream: but I've made a couple of comments about minor points of naming.

website/client/app.vue
```diff
@@ -388,7 +388,16 @@ export default {
       if (item.purchaseType === 'card') {
         this.selectedSpellToBuy = item;
 
+        // hide the dialog
         this.$root.$emit('bv::hide::modal', 'buy-modal');
+        // remove the dialog from our modal-stack,
+        // the default hidden event is delayed
+        this.$root.$emit('bv::modal::hidden', {
```
paglias, line 400: what happens when this even is fired again by the lib and picked up by the listener in app.vue?

If the event is fired in the next tick can't we just delay the next one?

```
Vue.nextTick(() => this.$root.$emit('bv::show::modal', 'select-member-modal'));
```

website/common/locales/en/spells.json
negue: will do! I always try to find the right spot, but sometimes it's not possible haha, good catch :)

website/client/app.vue
negue: tried it right now, it still too early before the normal `hidden` event arrives. I guess bootstrap-vue uses a callback once the modal is really closed and then fires the event, which definitely takes longer than a vue-tick

paglias: Ah okay, but just to be sure, this means the listener in app.vue will be called twice? Can this cause any problem?

negue: once the first `hidden` event passes, the `modalStack` is empty, and on the second call it tries to get the next modal to show again, 

but `let modalOnTop = this.$store.state.modalStack[modalStackLength - 1];` the position would be `-1` 

and `if (modalOnTop) this.$root.$emit('bv::show::modal', modalOnTop, {fromRoot: true});` doesn't do anything with an undefined `modalOnTop`

so I guess it's safe

paglias: i'm just wondering about the case were something else is in the stack

negue: I can't get any other modal inside the stack, so should I add any extra checks ?

paglias: @TheHollidayInn opinions on this?

TheHollidayInn: Talked to @negue offline, and I think this is the correct choice for now. Maybe we can resist the modal stack, but I believe we have to this for now.

TheHollidayInn (approved):

paglias:
So this is ready? If someone has tested it and it works, let's merge

PR #9812, opened 2018-01-16
Do not show visual buffs on choose class avatars

### LINKED ISSUE #9798: when changing class, you can't see which class is which if you have an avatar transformation buff (Snowball, Shiny Seed, Seafoam, Spooky Sparkles)
When you are affected by an avatar transformation buff (e.g., Snowball), and you try to change class, this happens unhelpfully:

img

What should happen is that the avatars you see are not affected by the buff.

jzhou69:
Fixes #9798

----
UUID: 40c49c2b-1929-4e53-9c95-b2cdddaa01d6

### DIFF
--- a/website/client/components/achievements/chooseClass.vue
+++ b/website/client/components/achievements/chooseClass.vue
@@ -21,6 +21,7 @@
                 :hideClassBadge='true',
                 :spritesMargin='"1.8em 1.5em"',
                 :overrideTopPadding='"0px"',
+                :showVisualBuffs='false',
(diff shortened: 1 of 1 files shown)


SabreCat:
Thanks for this, @jzhou69, and welcome to the ranks of the Blacksmiths at tier 1! This will run in a staging environment for a few days, then you should see it go live on the public site late this week.

Note that contributor tiers require increasing amounts of effort from one to the next, so tier 2 may require multiple PRs or a larger PR to attain. But keep helping out and we will express our gratitude accordingly!

PR #9813, opened 2018-01-17
Feature - Store Exact Completion Date For Dailies

### LINKED ISSUE #8070: Store Exact Completion Date for Dailies
The habitica-user-data.json file does not store the user action update data on dailies. Instead what we have is the cron update data for dailies. This leads to some rather stage data export since we lose the timestamp of user input. I understand that a certain amount of preening is necessary for performance and database management. That said, I don't think we should delete this user history data or, in fact, fail to store it for a certain period. I propose storing maybe a month (or week) worth of data for non-subscribers and maybe a bit longer for subscribers. 

This change will lead to much better data that can be used for personal study and for broader community research on when people log their dailies and habits. This could create some very interesting data science research as well. In some ways this is a follow on to #8069.  

Look forward to hearing thoughts on this. Thanks.

LulaV14:

Fixes #8070 

### Changes
Changed the way cron was saving history for dailies. 
History is now being saved in scoreTask.js
Added tests to scoreTask.test.js and deleted a test from cron.test.js

----
UUID: 4a5d7afb-d65e-485b-8e46-c9414a43b0bb

### DIFF
--- a/website/common/script/ops/scoreTask.js
+++ b/website/common/script/ops/scoreTask.js
@@ -249,10 +249,9 @@ module.exports = function scoreTask (options = {}, req = {}) {
         task.history = task.history || [];
         let historyEntry = {
           date: Number(new Date()),
-          value: task.value
+          value: task.value,
         };
         task.history.push(historyEntry);
-
       } else if (direction === 'down') {
         // Remove a streak achievement if streak was a multiple of 21 and the daily was undone
         if (task.streak % 21 === 0) user.achievements.streak = user.achievements.streak ? user.achievements.streak - 1 : 0;
@@ -260,16 +259,15 @@ module.exports = function scoreTask (options = {}, req = {}) {
         task.completed = false;
 
         // Delete history entry when daily unchecked
-        if(task.history || task.history.length > 0) {
-          task.history.forEach((history, index) => {
-            let history_date = moment(history.date);
+        if (task.history || task.history.length > 0) {
+          task.history.forEach((entry, index) => {
+            let historyDate = moment(entry.date);
             let today = moment(Number(new Date()));
-            if(history_date.startOf('day').isSame(today.startOf('day'))) {
+            if (historyDate.startOf('day').isSame(today.startOf('day'))) {
               task.history.splice(index, 1);
             }
           });
         }
-
       }
     }
   } else if (task.type === 'todo') {
--- a/website/common/script/ops/scoreTask.js
+++ b/website/common/script/ops/scoreTask.js
@@ -8,6 +8,7 @@ import i18n from '../i18n';
 import updateStats from '../fns/updateStats';
 import crit from '../fns/crit';
 import statsComputed from '../libs/statsComputed';
+import moment from 'moment';
 
 const MAX_TASK_VALUE = 21.27;
 const MIN_TASK_VALUE = -47.27;
@@ -243,11 +244,32 @@ module.exports = function scoreTask (options = {}, req = {}) {
           if (user.addNotification) user.addNotification('STREAK_ACHIEVEMENT');
         }
         task.completed = true;
+
+        // Save history entry for daily
+        task.history = task.history || [];
+        let historyEntry = {
+          date: Number(new Date()),
+          value: task.value
+        };
+        task.history.push(historyEntry);
+
       } else if (direction === 'down') {
         // Remove a streak achievement if streak was a multiple of 21 and the daily was undone
         if (task.streak % 21 === 0) user.achievements.streak = user.achievements.streak ? user.achievements.streak - 1 : 0;
         task.streak -= 1;
         task.completed = false;
+
+        // Delete history entry when daily unchecked
+        if(task.history || task.history.length > 0) {
+          task.history.forEach((history, index) => {
+            let history_date = moment(history.date);
+            let today = moment(Number(new Date()));
+            if(history_date.startOf('day').isSame(today.startOf('day'))) {
+              task.history.splice(index, 1);
+            }
+          });
+        }
+
       }
     }
   } else if (task.type === 'todo') {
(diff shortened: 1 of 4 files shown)


website/common/script/ops/scoreTask.js
```diff
@@ -243,11 +244,30 @@ module.exports = function scoreTask (options = {}, req = {}) {
...
+        task.history.push(historyEntry);
       } else if (direction === 'down') {
         // Remove a streak achievement if streak was a multiple of 21 and the daily was undone
         if (task.streak % 21 === 0) user.achievements.streak = user.achievements.streak ? user.achievements.streak - 1 : 0;
         task.streak -= 1;
         task.completed = false;
+
+        // Delete history entry when daily unchecked
+        if (task.history || task.history.length > 0) {
+          task.history.forEach((entry, index) => {
```
paglias: Is it necessary to iterate over all the history entries here? Can't you check just the last entry since dailies have only one entry per day?

paglias (changes requested):

website/common/script/ops/scoreTask.js
LulaV14: Ok sure, wasn't sure about that but yes, I believe it's better to just get the last one then.
I'll change that 😊

Alys:
@LulaV14 cron still  needs to write to the Daily's history in the case where the user did not complete the Daily. The history object should remain the same as it is now except for your change that in completed Dailies the date-time records the time completed.

LulaV14:
@Alys noted, sorry for the delay.
I'm updating it right now.

paglias:
Removing the changes to the cron code entirely won't cause the daily history to be recorded twice? Once when the daily is scored and once at cron?

Alys:
@LulaV14 We will need some changes to the cron code. Currently in production, cron writes to the task's history regardless of whether the Daily has been completed or not. With your change, cron doesn't need to write to history for a completed task, but still needs to write to the Daily's history in the case where the user did not complete the Daily.

LulaV14:
@Alys @paglias  Oh ok got it, sorry for missing that.
So I'll just add a condition to check if task.completed is false, then it will be added by cron.

I was trying to test something that came up today, I was checking the history for each task after cron ran and I saw that some dailies had the same day repeated not two but three times for one single day, have you had that same problem before? I was thinking that it should be coming from cron because I'm testing on scoreTask and it doesn't happen there. 
Btw how can I run cron several times so I can test it more than once a day?

Alys:
@LulaV14 No worries. :) The condition you describe sounds good.

The same day repeated is odd! I don't recall seeing other reports of it, but it's not something that's easily spotted. Can you show us the history data for one of those Dailies?

To run cron at any time on your local install, the Debug menu in the footer of the website has a "+1 Missed Day" entry. Click that then click the sync button in the header and cron will run. What that item does is it changes the account's `lastCron` date to be 24 hours in the past. You can use it as many times as you need each day.

LulaV14:
Thanks @Alys! I'm uploading the changes right now.
I lost track of the task for the same day multiple entries because I'm using the  "+1 Missed Day" button and now I don't know if it's only because of that. But I haven't seen any other problem.

LulaV14:
@Alys @paglias One question, as I understand scoreTask.js is for the client side but I also modified cron.js for the server side.. shouldn't we wait to merge this PR until changes are also done for the mobile app? Btw I'll be glad if I can make those changes for both Android and iOS.

Alys:
scoreTask.js is run on the server side too. :)  It's what the API uses to score a task up or down and so the website, official mobile apps, and third party tools all use it. This means that there won't need to be changes to the mobile apps but thank you for thinking about it and offering!

LulaV14:
@Alys Oh ok awesome 😄 ! So if there's anything else to change I'm here to help 😊

paglias:
The code is ok, but I haven't been able to test this locally

website/server/libs/cron.js
```diff
@@ -308,6 +308,12 @@ export function cron (options = {}) {
         atLeastOneDailyDue = shouldDo(thatDay.toDate(), task, user.preferences);
       }
     } else {
+      // add history entry when task was not completed
+      task.history.push({
+        date: Number(new Date()),
+        value: task.value,
+      });
+
```
Alys: This still needs to be done after all the code below because some of that code will affect the value. For example, if Stealth doesn't protect the Daily, then scoreTask around line 340 will make the Daily redder, and it's that redder value that we'd want to record. Probably putting this code at the bottom of the `else` block is what we want.

Alys (changes requested):

website/server/libs/cron.js
LulaV14: Oh okok got it. 
Sorry for all the changes, I hope to catch up quicker next time.

Alys (approved):

website/server/libs/cron.js
Alys: No worries! :) The cron code is large and complex.

Alys:
I've tested this locally and it seems good! I'm going to do some more brief testing over the next couple of days so that I can watch how ti behaves with cron running normally but my impression at the moment is that it's doing what we need!

LulaV14:
Thank you so much @Alys 😊 !

Alys:
This has been working exactly as expected over the past three days on my local install so I'm marking it as ready for merging, which will probably happen next Monday and then it will probably go live a few days after that ([release cycle](http://habitica.wikia.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API#Acceptance_of_Your_PR_and_the_Release_Cycle). Thank you @LulaV14! It will count towards your next tier when it's merged.

paglias:
@LulaV14  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #9823, opened 2018-01-18
Allow user to buy numbered special gear

### LINKED ISSUE #9731: Equipment shop errors: some gear is mis-filed, some doesn't appear

SabreCat:

Fixes https://github.com/HabitRPG/habitica/issues/9731#issuecomment-357417784

### Changes

**Previously**, logic to determine if a user was eligible to purchase a piece of equipment mistakenly treated `special` equipment with numeric names (i.e., early-released legendary quest rewards) as sequential, requiring the numbered pieces to be bought in order.
**Now**, this correctly only applies to class gear.

### DIFF
--- a/test/common/ops/buy/buyGear.js
+++ b/test/common/ops/buy/buyGear.js
@@ -138,5 +138,27 @@ describe('shared.ops.buyGear', () => {
         done();
       }
     });
+
+    it('does not buyGear equipment if user does not own prior item in sequence', (done) => {
+      user.stats.gp = 200;
+
+      try {
+        buyGear(user, {params: {key: 'armor_warrior_2'}});
+      } catch (err) {
+        expect(err).to.be.an.instanceof(NotAuthorized);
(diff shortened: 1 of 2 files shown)


paglias:
Restarted the build, the code LGTM 👍

PR #9824, opened 2018-01-18
fix user avatar transformation - change number of buffs owned when you use them - fixes #9795

### LINKED ISSUE #9795: no change in number of avatar transformation buffs owned when you use them (Snowball, Shiny Seed, Seafoam, Spooky Sparkles)
When you use an avatar transformation buff items (e.g., Snowball) the number owned shown in your inventory doesn't decrease until you click the sync button.

It should decrease as soon as you use one item.

negue:
Cast the spell also on the website, to change the `stats.buffs` properties.

Fixes #9795


TheHollidayInn:
Fixes #9795

website/client/mixins/spells.js
```diff
@@ -73,12 +79,9 @@ export default {
...
       this.applyingAction = false;
 
-      let spellUrl = `/api/v3/user/class/cast/${spell.key}`;
-      if (targetId) spellUrl += `?targetId=${targetId}`;
-
       let spellText = typeof spell.text === 'function' ? spell.text() : spell.text;
 
-      await axios.post(spellUrl);
+      let apiResult = await this.$store.dispatch('user:castSpell', {key: spell.key, targetId});
       let msg = this.$t('youCast', {
```
paglias: can we move this to a default branch in the switch just below?

paglias (approved):

TheHollidayInn (approved):

paglias:
There's an issue with this @negue, when you buy a card and try to cast it on a party member the modal with the list of members appears but you can only cast if you click on the avatar, clicking on `Select` doesn't wok. Also the cursor becomes a cross which is not currently happening on staging

negue:
huh odd, I think I only used the select button never the avatar itself, will check again

negue:
@paglias  I couldn't reproduce the cast on avatar , it only oped the info modal , but I found a different issues which caused only to cast the spell every second try

paglias:
@negue working now!

PR #9835, opened 2018-01-20
checkout of the Inn - banner

### LINKED ISSUE #9188: "Check out of Inn" Banner
We need to add the following banner to the Tasks page and the Party page when you're resting in the inn, to enable you to easily check out of the inn without leaving the page that you're on.

screen shot 2017-10-13 at 6 51 22 pm

The designs for this feature are available in Zeplin. The person who volunteers to work on this feature should let us know an email address that we can use to invite them to our Zeplin, where they'll find all the information that they need to build the banner. Feel free to ask questions here!

negue:
closes #9188


website/client/app.vue
```diff
@@ -105,6 +112,61 @@ div
...
+    height: 40px;
+    background-color: #2995cd;
+    position: fixed;
+    top: 0;
+    z-index: 1030;
+    display: flex;
+
+    .content {
+      height: 24px;
+      font-family: Roboto;
```
paglias: is this necessary?

website/client/app.vue
```diff
@@ -105,6 +112,61 @@ div
...
+    .closepadding {
+      margin: 11px 24px;
+      display: inline-block;
+      position: absolute;
+      right: 0;
+      top: 0;
+      cursor: pointer;
+
+      span svg path {
+        stroke: #A9DCF6;
```
paglias: color variable?

website/client/app.vue
```diff
@@ -105,6 +112,61 @@ div
...
+      }
+    }
+
+    .separator {
+      color: #50b5e9;
```
paglias: color variable?

website/client/app.vue
```diff
@@ -105,6 +112,61 @@ div
...
+    z-index: 1030;
+    display: flex;
+
+    .content {
+      height: 24px;
+      font-family: Roboto;
+      font-size: 14px;
+      line-height: 1.71;
+      text-align: center;
+      color: #ffffff;
```
paglias: $white

website/client/app.vue
```diff
@@ -105,6 +112,61 @@ div
...
+    #app-header {
+      margin-top: 96px !important;
+    }
+
+  }
+
+  .resting-banner {
+    width: 100%;
+    height: 40px;
+    background-color: #2995cd;
```
paglias: color variable?

paglias (changes requested):

website/client/app.vue
```diff
@@ -417,6 +488,12 @@ export default {
     hideLoadingScreen () {
       this.loading = false;
     },
+    hideBanner () {
+      this.bannerHidden = true;
```
paglias, line 508: do we want to persist this in localstorage? @Alys what do you think?

paglias (changes requested):

website/client/app.vue
Alys: I'm really not keen on using localstorage. We had so many problems with it in the past!

What's the intention of the dismiss button for that banner? Is it meant to hide the banner for just the current session (e.g., until you close the browser window or log out) or permanently?

If it's meant to hide it permanently, we'd probably want it to do that across all browsers (e.g., home and work), which means a preference setting in the user's document. The setting could be removed automatically when the user stops resting in the inn, so that they'd see the banner again the next time they started resting.

How important is the dismiss button? We could consider going live without the dismiss button and then adding it later if there's requests for it.

paglias: If we want to keep it dismissed until the windows is closed there's no need for localstorage or a server setting, it depens from the behavior we want

paglias: Anyway we've been using localstorage again for several machine-specific settings and hadn't had problems. For stuff like this is perfect

negue: @lemoness  @Tressley Should 

- this banner be only visible during each browser session (browser closed => it appears next time again) , 
- or should this be hidden for ever 
   -  a) for the current browser 
or 
   - b) for every other opened instance (saved into the users settings)

Tressley: Each browser session is how I imagined it working.

paglias (approved):

TheHollidayInn:
@paglias You approved, so is this ready for merge or does it need more testing?

paglias:
The code is ok but haven't tested it locally

Alys:
The banner is currently appearing on all pages, not just on the Tasks and Party page mentioned in the issue. Personally, I'm on the fence about whether that's good or bad, so I'm not saying it has to be changed, just mentioning it. Maybe consistency across all pages is better. It might be annoying if the banner keeps flicking on and off as you move between pages.

There's merge conflicts in groups.json so when someone fixes that, you might also want to delete the strings `innText` and `innTextBroken`, which are part of the conflict. They're not used anywhere now.

I've tested this locally and the behaviour seems good:
- [x] The banner appears when you rest in the inn, including if you rest on another platform and then sync on the website.
- [x] It disappears when you click x and then reappears when you reload, as desired. 
- [x] It disappears permanently when you stop resting, either from the banner's button, or from the Tavern's button, or from another platform then a sync on the website.

However the appearance isn't ideal on a small screen or normal screen with large font (e.g., for someone with poor vision). The text wraps as you'd expect but the blue banner doesn't expand downwards to allow all the text to be seen. We will get bug reports for that so it would be nice if it could be fixed before going live rather than having an issue opened for it.

screenshot_2018-02-04_16-27-09

Alys:
Also I think I'm finding the sync bar harder to see when the banner is in place, although it's difficult to get a good feel for it when syncs are so fast locally. I think the sync bar is a bit thinner than it should be, maybe as if the banner is covering half of it? The colour might also be less good against the blue of the banner than against the dark purple.

paglias:
Pinging @Tressley about https://github.com/HabitRPG/habitica/pull/9835#issuecomment-362884958 and setting tags back to `needs work`

Tressley:
Sorry for the delay. Here are my thoughts:
- I agree that it feels weird to have the banner only appear on select pages, so let's just have it appear everywhere. Users have the ability to dismiss it and it will reappear for their next session.
- For smaller screens, we can bring the font-size down to ~12px and shorten the text to, "You are currently checked into the Inn | Resume Damage." Wrapping text, creating a taller banner, is not an ideal experience for mobile users.
- Regarding @Alys comment about the visibility of the sync bar. I think it's fine as is. While being checked into the inn and syncing is definitely an understandable use case, it's not primary and the sync bar shouldn't be visible for very long/attention grabbing. Confirmation of a completed sync tends to have a visible identifier in the content changing.

Alys:
I'm not sure forcing the font to be small is the best solution because the wrapping can occur when you're on a larger screen and have used zoom to make the font bigger, which people with visual difficulties can do. The text will also be longer in some translations which means that a font size of 12px might not be enough for it to all be visible even on a larger screen (and smaller than that will be unreadable to more people). Having the blue banner expand downwards to fit all the text is a more responsive solution.

I don't think we should remove words from the text because it's designed to prevent confusion about why you aren't taking damage or making quest progress - without that explanation there's less point to it. 

Regarding the sync bar: It's worth noting that resting in the inn for days or weeks at a time while still doing tasks is a use case for people with chronic illnesses or anyone who is too busy for their normal routine (e. g., I did that in the weeks after the new website launch). The sync bar needs to be just as usable as normal for those people. 

The sync bar is essential for correct use of Habitica. Confirmation of a completed sync almost never has a visible identifier in the content changing. The visible change happens **before** the sync is done - that's why the sync bar is needed. Remember that the sync bar was left out of the original new website design but had to be brought back as one of the early urgent changes because it was so necessary. I'm definitely not saying that we have to improve the appearance of the sync bar over the Resting banner because it might be just my perception that it's harder to see. However if others also think it's less visible, then I believe we should make a change to improve that. I know that might be something we'd have to do after this PR has gone live if that's when other feedback comes in.

deilann:
I am regularly in the inn due to illness. When I am in the inn, I need to sync quite a bit because I'm still using my task lists! I'm just needing to pause damage. Frex, when I tick many items of a long checklist, I have to sync, else I might get less XP due to sync issues. That's just one thing I use it for. If I had to refresh every time I needed to sync, I would probably not use Habitica while in a flare up or when injured, which would vastly impact my life in a negative direction.

Alys:
Actually, sorry, I'm wrong about this part: _"Remember that the sync bar was left out of the original new website design but had to be brought back as one of the early urgent changes because it was so necessary."_
I was thinking of the sync button. It took us longer to bring back the bar.

Alys:
@Tressley I've probably made this more complicated than it needs to be. Sorry about that! Let's ignore the sync bar visibility for now because we have only my feeling that it might not be visible enough and  that might change when I'm seeing it on a server that's slower than my local install. We can deal with that later if we need to.

So basically all we need is:
- This text displayed when someone is resting: 
"You are currently checked into the Inn. Your Dailies won't damage you and you won't make progress towards Quests. | Resume Damage".
It can't be significantly shortened because then it won't explain the common questions that users have (why they're not taking damage from their Dailies or making Quest progress). 
- That text visible in full under all these conditions:
   - small browser window for people on small laptops / mobile devices
   - large browser zoom level / font size for people with visual difficulties
   - non-English languages, some of which will make the text noticeably longer, for example, in Russian using Google Translate:
"Вы в настоящее время зарегистрированы в отеле. Ваши Dailies не повредят вам, и вы не достигнете прогресса в отношении квестов. | Возобновить повреждение"
     
Forcing a small font size won't be good for users with visual difficulties and might not help enough for the longer non-English languages.

You're right that wrapping text and creating a taller banner isn't ideal for mobile users but the mobile website has other, more significant problems - this will be a minor one by comparison. It will also be easy for people to close the banner on mobile since the close button will be close to the site's menu where their finger is likely to be anyway. A banner that expands itself to the right height for whatever amount of text that it needs to contain will give a better experience for all the other use cases.

negue:
So can this go back to `ready for review` ?  @Tressley  @Alys

Alys:
@negue Sorry! I didn't notice that you'd added a commit. I'm changing this back to ready for review and I'll test it locally again today.

Alys:
I can't test today because local installs are broken at the moment. I'll test as soon soon as I can once they're working again.

Alys:
I've tested and there's no negative effects from the most recent commit. :+1:

website/client/app.vue
```diff
@@ -13,34 +13,44 @@ div
...
-          )
-          selectMembersModal(
-            :item="selectedSpellToBuy || {}",
-            :group="user.party",
-            @memberSelected="memberSelected($event)",
-          )
-
-          div(:class='{sticky: user.preferences.stickyHeader}')
-            router-view
+        template(v-if="isUserLoaded")
```
paglias: can you fix the indentation here? Then this will be ready

paglias (changes requested):

website/client/app.vue
paglias: I mean, there seem to be an extra indent used here

negue: oh you are right! fixing..

PR #9837, opened 2018-01-20
Removed display of users personal checklist on challenge tasks

### LINKED ISSUE #9834: Checklists on Challenge tasks should be private to the participant
When a participant adds a checklist to a task from a Challenge, it should not be possible for the challenge owner or any other player to see the checklist. The checklist shouldn't appear in the CSV export file or in the "View Progress" screen.

The checklists can also be seen when using the API directly, which should also not be possible.

TheHollidayInn:
Fixes #9834

### DIFF
--- a/website/server/controllers/api-v3/members.js
+++ b/website/server/controllers/api-v3/members.js
@@ -413,7 +413,10 @@ api.getChallengeMemberProgress = {
     // manually call toJSON with minimize: true so empty paths aren't returned
     let response = member.toJSON({minimize: true});
     delete response.challenges;
-    response.tasks = chalTasks.map(chalTask => chalTask.toJSON({minimize: true}));
+    response.tasks = chalTasks.map(chalTask => {
+      chalTask.checklist = []; // Clear checklists as they are private
+      return chalTask.toJSON({minimize: true});
+    });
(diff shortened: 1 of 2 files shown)


website/server/controllers/api-v3/members.js
```diff
@@ -413,7 +413,10 @@ api.getChallengeMemberProgress = {
     // manually call toJSON with minimize: true so empty paths aren't returned
     let response = member.toJSON({minimize: true});
     delete response.challenges;
-    response.tasks = chalTasks.map(chalTask => chalTask.toJSON({minimize: true}));
+    response.tasks = chalTasks.map(chalTask => {
```
paglias, line 416: I think it's easier to just edit line 410 here that already contains some data to remove

paglias (changes requested):

website/server/controllers/api-v3/members.js
TheHollidayInn: I'd rather not because that would remove the value from the response and any integrations expecting the existence of the checklists would break. So returning an empty array is safer.

paglias: 👍

PR #9839, opened 2018-01-21
Added two handed message to weapons

### LINKED ISSUE #8441: two-handed weapons/shields aren't easily identifiable as such

TheHollidayInn:
Fixes #8441

### DIFF
--- a/website/common/script/content/gear/weapon.js
+++ b/website/common/script/content/gear/weapon.js
@@ -1,3 +1,5 @@
+import t from '../translation';
+
 import {weapon as baseWeapon} from './sets/base';
 
 import {weapon as healerWeapon} from './sets/healer';
@@ -21,5 +23,17 @@ let weapon = {
   mystery: mysteryWeapon,
   armoire: armoireWeapon,
 };
+
+// Add Two Handed message to all weapons
+for (let key in weapon) {
+  const set = weapon[key];
+  for (let weaponKey in set) {
+    const item = set[weaponKey];
+    item.notes = (lang) => {
+      const twoHandedText = item.twoHanded ? t('twoHandedItem')(lang) : '';
+      return `${t('armorHealer2Notes', { con: 9 })(lang)} ${twoHandedText}`;
+    };
+  }
+}
 
 module.exports = weapon;
--- a/website/common/script/content/gear/weapon.js
+++ b/website/common/script/content/gear/weapon.js
@@ -25,13 +25,40 @@ let weapon = {
 };
 
 // Add Two Handed message to all weapons
+const rtlLanguages = [
+  'ae',	/* Avestan */
+  'ar',   /* 'العربية', Arabic */
+  'arc',  /* Aramaic */
+  'bcc',  /* 'بلوچی مکرانی', Southern Balochi */
+  'bqi',  /* 'بختياري', Bakthiari */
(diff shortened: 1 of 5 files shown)


website/common/script/content/gear/weapon.js
```diff
@@ -22,4 +24,16 @@ let weapon = {
...
 };
 
+// Add Two Handed message to all weapons
+for (let key in weapon) {
+  const set = weapon[key];
+  for (let weaponKey in set) {
+    const item = set[weaponKey];
+    item.notes = (lang) => {
+      const twoHandedText = item.twoHanded ? t('twoHandedItem')(lang) : '';
+      return `${t('armorHealer2Notes', { con: 9 })(lang)} ${twoHandedText}`;
```
Alys: armorHealer2Notes ? :)

I'm not sure that putting `${twoHandedText}` at the end of the string will work perfectly for every language. For RTL languages, it will make that string be read first, followed by the rest of the description.  However probably the only(?) proper alternative is replacing all current notes text with versions that have a new variable in them, which will be very disruptive to all the existing translations. I think it might be best to put up with the RTL problem unless there's significant complaints. I think though that it is worth putting a comment in the code here to indicate that we're aware of the problem. Maybe in future there'll be other changes that will make it easier to handle RTL languages better here.

TheHollidayInn: :P Woops. Okay, I'll fix that.

We pass the user language, so if there is an indicator as to the language direction there, I can prepend the info.

PR #9848, opened 2018-01-21
Keys kennel fixes

### LINKED ISSUE #9243: Key to the Kennels is not showing for users who are eligible
Key to the Kennels is not showing for users who are eligible. It should be restored to the Market.

### LINKED ISSUE #9491: API route for Key to the Kennels missing checks to see if user is eligible

### Description

Via cTheDragons in the Report a Bug Guild:

> The API Call https://habitica.com/api/v3/user/release-both does not check if the user has all the pets and mounts for Standard (Gen 1) before awarding Triad Bingo. It also does not charge gems if the uers does not have all the pets and mounts for Standard (Gen 1).

TheHollidayInn:
Fixes #9243
Fixes #9491

- Add display of keys when you acquire all pets or mounts without refresh
- Remove keys after use without refresh
- Added redirect to stable after use
- Only show keys when you meat count requirements rather than achievement
- Added api calls

## Test Cases
- [x] Keys to Pets appears when getting your 90th pet 
- [x] Keys to Pets removes all pets immediately and the key disappears
- [x] Keys to Mounts appears when getting your 90th mount 
- [x] Keys to Mounts removes all mounts immediately and the key disappears
- [x] Keys to Kennel appears when getting your 90th pet while having all 90 mounts
- [x] Keys to Kennel removes all pets and mounts immediately and the key disappears
- [x] Pet key doesn't show when you have raised 90 pets


website/client/components/shops/market/keysToKennel.vue
```diff
@@ -109,10 +113,12 @@ export default {
           if (!confirm(this.$t('releaseMountsConfirm'))) return;
           try {
             releaseMounts(this.user);
+            axios.post('/api/v3/user/release-mounts');
```
paglias: can we make these actions?

TheHollidayInn: Sure.

SabreCat:
Is there supposed to be a modal announcement when you reach the threshold / gain the achievement? Because none turned up for me.

SabreCat:
Price for Key to the Pet Kennels shows as 3 Gems, but with 3 Gems I'm prompted to buy Gems instead of being allowed to activate the Key.

SabreCat:
User's Beastmaster achievement does not gain its "and has released their pets" addendum until after a hard refresh. Should we force a hard refresh after use of a Key?

TheHollidayInn:
There is no modal.

Can you screenshot what the addendum looks like? I'm not sure I addressed that in this PR. I thought the achievements were gained before this.

SabreCat:
image

The achievement is awarded correctly--you're right, that's on the server side and unaffected by this. But the client doesn't show it until after a reload.

SabreCat:
Master Key is showing up prematurely. When I raise my last pet into a mount, both the Mount and Master keys appear, and the Pets key remains. (That last might be intentional, but the Master key behavior doesn't match test cases.)

SabreCat:
Price of each key should probably appear in the modal that shows up on click. I keep mistaking the Gems balance in the modal footer for the price. Could also follow the practice of including the Gem price in the JS confirm popup ("Are you sure you want to release your standard Mounts for 4 Gems?")

SabreCat:
Key to the Pet Kennels is still available for me after releasing mounts, despite my not having any standard pets remaining.

SabreCat (changes requested):
See comments for functional issues

TheHollidayInn:
Updated!

SabreCat (approved):

DazoTrove:
I saw this was deployed today, but the test cases above do not match how these used to work before the redesign. The Release Both (Master Keys to the Kennels) option became available for 6 gems after gaining 90 mounts, and it used to be free when you had 90 pets and 90 mounts (which grants the Triad Bingo achievement).

I think the test cases should be changed to the following, and changes should be made to support the free Master Keys to the Kennels when a player has all 90 pets and 90 mounts.

- Master Keys to the Kennels appears when getting your 90th mount (for 6 gems)
- Master Keys to the Kennels becomes free when getting your 90th pet while having all 90 mounts
- Master Keys to the Kennels removes all pets and mounts immediately and the key disappears

PR #9849, opened 2018-01-21
Remove unused packages

paglias:


paglias:
I used https://github.com/dylang/npm-check to find unused packages and then verified them manually since it has some problems with webpack loaders / .vue files

PR #9851, opened 2018-01-22
Hide stats points allocation on locked or disabled class system (#9826)

### LINKED ISSUE #9826: Stats Points allocation should not be available when the class system is locked / opted out
Before a player unlocks the class system (at level 10) or when they have opted out of the class system, they should not see the Stat Points and Allocation method section at the bottom of the User Icon > Stats screen. I.e., these elements should be hidden:

screenshot_2018-01-20_07-26-49

This is because Stat Points and the manual/automatic allocation methods are part of the class system.

For the benefit of anyone working on this, a player who has not yet unlocked the class system has this in their data:
`"flags.classSelected": false`

A player who has opted out of the class system has this:
`"preferences.disableClasses": true`

nmassaro:
Fixes #9826.

### Changes

The stats points allocation section is now hidden from the Avatar/User Modal if the user hasn't selected a class, or has classes disabled (i.e. user flag 'classSelected' is false, 'disableClasses' is true under preferences).

Screenshots:

screen shot 2018-01-21 at 8 05 22 pm

Above (before changes), for a user who has not selected a class.

screen shot 2018-01-21 at 8 00 52 pm

And after changes.

:)

----
UUID: 35086b7e-ac9c-4c40-9737-2c5bdcd17325

### DIFF
--- a/website/client/components/userMenu/profile.vue
+++ b/website/client/components/userMenu/profile.vue
@@ -255,7 +255,7 @@ div
                 li
                   strong {{$t('buffs')}}:
                   | {{user.stats.buffs[stat]}}
-      #allocation(v-if='user._id === userLoggedIn._id')
+      #allocation(v-if='user._id === userLoggedIn._id && user.flags.classSelected && !user.preferences.disableClasses')
         .row.title-row
           .col-12.col-md-6
             h3(v-if='userLevel100Plus', v-once, v-html="$t('noMoreAllocate')")


paglias (approved):

SabreCat:
Thanks for this, @nmassaro, and welcome to the ranks of the Blacksmiths at tier 1! This will run in a staging environment for a few days, then go live on the public site around the end of the week.

Note that subsequent contributor tiers require increasing amounts of effort from one to the next, so tier 2 may require multiple PRs or a larger PR to attain. But keep helping out and we will express our gratitude accordingly!

PR #9856, opened 2018-01-22
Added message when there will be no notification for group

### LINKED ISSUE #9327: Warning about no notifications for Guilds with too many members
For performance reasons, Guilds that have over 5000 members do not receive notifications. Those Guilds should have the following warning in them:

"This Guild is too big to receive notifications for performance reasons. Be sure to check back often for replies to your messages!"

The style of the design should be as follows, albeit with the updated copy:

screen shot 2017-10-30 at 6 41 32 pm

If you'd like to work on this ticket, we'll invite you to our Zeplin so that you can get the specifics of implementing this design, so be sure to let us know which email address we should invite.

TheHollidayInn:
Fixes #9327


website/client/components/groups/group.vue
```diff
@@ -418,6 +432,9 @@ export default {
       if (!this.group.challenges) return false;
       return this.group.challenges.length === 0;
     },
+    showNoNotificationsMessage () {
+      return this.group.memberCount > 5000;
```
paglias: we should use the constant defined in the common code instead of the number directly, $store.state.constants.LARGE_GROUP_COUNT_MESSAGE_CUTOFF

website/common/locales/en/groups.json
```diff
@@ -415,5 +415,6 @@
   "wouldYouParticipate": "Would you like to participate?",
   "managerAdded": "Manager added successfully",
   "managerRemoved": "Manager removed successfully",
-  "leaderChanged": "Leader has been changed"
+  "leaderChanged": "Leader has been changed",
+  "groupNoNotifications": "This Guild does not have notifications due to member size. Be sure to check back often for replies to your messages!"
```
paglias, line 424: there's the string `  "largeGroupNote": "Note: This Guild is now too large to support notifications! Be sure to check back every day to see new messages.",
`

paglias (changes requested):

website/common/locales/en/groups.json
TheHollidayInn: Okay, I'll remove the old string as the string I've added is from Zeplin.

PR #9857, opened 2018-01-22
Added p parameter to group party query

### LINKED ISSUE #9450: emailed party invitations don't give the invited member an invitation to the party

TheHollidayInn:
Fixes #9450

### DIFF
--- a/website/client/components/static/home.vue
+++ b/website/client/components/static/home.vue
@@ -612,6 +612,10 @@
         }
 
         let groupInvite = '';
+        if (this.$route.query && this.$route.query.p) {
+          groupInvite = this.$route.query.p;
+        }
+
         if (this.$route.query && this.$route.query.groupInvite) {
           groupInvite = this.$route.query.groupInvite;
         }


paglias:
Regarding this, I'm honestly confused as to where the ?p param is coming from, but anyway since it seems to be happening can we try to remove the code that creates this param instead of supporting it in the client side? My fear is that ?p is way too generic and we risk having conflicts

TheHollidayInn:
We don't add query params to our emails/code a lot, so I'm not concerned about conflicts for now. But I'd be fine removing p as it is not a good parameter name. The param might be coming from mandrill or the email server.

paglias:
I'm quite sure it's a problem with mandrill or the way we encode parameters because there'a absolutely no place in the code where ?p is used, we only use ?groupInvite here https://github.com/HabitRPG/habitica/blob/develop/website/server/controllers/api-v3/groups.js#L1028

PR #9858, opened 2018-01-22
Added sync when won challenge

### LINKED ISSUE #7716: Winner of challenge sees a notification for a previous challenge that they've won

TheHollidayInn:
Fixes #7716

### DIFF
--- a/website/client/components/achievements/wonChallenge.vue
+++ b/website/client/components/achievements/wonChallenge.vue
@@ -1,5 +1,5 @@
 <template lang="pug">
-  b-modal#won-challenge(:title="$t('wonChallenge')", size='md', :hide-footer="true")
+  b-modal#won-challenge(:title="$t('wonChallenge')", size='md', :hide-footer="true", @shown='syncUser')
     .modal-body.text-center
       h4(v-markdown='user.achievements.challenges[user.achievements.challenges.length - 1]')
       .row
@@ -58,6 +58,12 @@ export default {
     };
   },
   methods: {
+    async syncUser () {
+      // @TODO: There are times when the user is not updated when this modal is shown
+      // It would be nice to not need to fetch to full user, maybe just the achievements?
+      // See https://github.com/HabitRPG/habitica/issues/7716
+      await this.$store.dispatch('user:fetch', {forceLoad: true});
+    },
     close () {
       this.$root.$emit('bv::hide::modal', 'won-challenge');
     },


paglias:
I'm not sure this will fix the problem, because the notification is shown when a request is made and new notifications are returned, but when notifications are returned we also return the `userV` from the server and if the user is out of date a sync request is made automatically so this way we'll only be doing two

TheHollidayInn:
I was about to ping you for this :P 

Yea, my only thinking is if you look at the PR from the old client, it did fix this problem by syncing. I still think there could be a delay in loading the user with the scenario you are describing, so the user could see the popup before the new version is loaded.

Although, maybe I should try check if the user:fetch is in progress?

paglias:
Yeah i think the problem is the delay in syncing, but this won't change and actually the sync will start just slightly later since the modal has to be shown. But yeah I would try to reproduce this locally to see if the user is synced or the problem is another

website/client/components/achievements/wonChallenge.vue
```diff
@@ -58,6 +58,12 @@ export default {
     };
   },
   methods: {
+    async syncUser () {
+      // @TODO: There are times when the user is not updated when this modal is shown
+      // It would be nice to not need to fetch to full user, maybe just the achievements?
+      // See https://github.com/HabitRPG/habitica/issues/7716
+      await this.$store.dispatch('user:fetch', {forceLoad: true});
```
negue, line 65: should something be executed after the dispatch? if not you can remove the `async`/`await` and it'll still work ^^

TheHollidayInn: Only if I return the promise and await this function.

PR #9868, opened 2018-01-23
Added initial account banned modal

### LINKED ISSUE #8636: better message for blocked users
Currently, when a moderator blocks a user, the user sees this message if they are already logged in or are trying to log in:

image

We want the message to be replaced with this text:

**"This account, Profile Name {insert profile name here}, User ID {insert User ID here}, has been blocked for breaking the [Community Guidelines](https://habitica.com/static/community-guidelines) or [Terms of Service](https://habitica.com/static/terms). For details or to ask to be unblocked, please email Lemoness at leslie@habitica.com or ask your parent or guardian to email her. Please copy your Profile Name and User ID into the email."**

The links for the Community Guidelines and Terms of Service are https://habitica.com/static/community-guidelines and https://habitica.com/static/terms

The email address should be a mailto link and must use the environment variable `EMAILS.COMMUNITY_MANAGER_EMAIL` instead of having "leslie@habitica.com" hard-coded. To see examples of how to do that, use `git grep COMMUNITY_MANAGER_EMAIL`.

We also want the format of the messages to be changed.  Instead of seeing the red notification shown above, they should see a modal that displays the message. The links must be clickable and the text must be selectable for copying. The modal does not need a button for closing it since the website can't be used when you're blocked, but if you want to include a button (or if anyone else here asks that one be included), that's fine.

If it's difficult to stop the red notification from appearing, then it's okay if they see that as well as the modal.

For testing, you can set a user account to be blocked by editing their document in the database to set `auth.blocked` to `true`, or set one account to be an admin user (using the Debug menu in the footer of your test website), and then go to Social > Hall to block a different account. The interface is shown below. Enter the User ID, click submit, then when the details open up, click the "Auth" header. 

image

### LINKED ISSUE #8672: Account suspended/blocked error lacks detail; lead to hanging website on refresh

### General Info
  * UUID: 54d0305c-fbc0-4fe8-a941-948204625f08
  * Browser: Firefox 52.0.2
  * OS: Windows 10

### Description

When a user is logged in and they become blocked (i.e. the "auth.blocked" property becomes set to "true"), a refresh of the page causes the website to hang on the loading screen.  Screenshot is below.

Related to this (I think) is a block of code in website/client-old/js/config.js (line 84-87) that never runs due to the response.data object from the server never having a "code" property (legacy?).  I added a console.log statement which you can see in the screenshot (returned response.data object is in the bottom right pane).  As you can see, there is no way to determine that the error is an account suspended error unless you look at the locale-dependent message.

githubissue

I think the easiest/concise/maintainable way to fix this would be to extend the "NotAuthorized" class with an "AccountSuspended" class (website/common/script/libs/errors.js).  Editing the "NotAuthorized" class would be relatively difficult because the constructor only accepts a string, and we probably don't want to mess with the constructor or do weird hacky logic with message.  I think it also makes sense from an abstraction standpoint; an AccountSuspended error is an extension of a NotAuthorized error, and there is already logic in config.js that is supposed to handle the specific case of a suspended account.  I.e. something like:

githubissue2

What do you guys think?

### LINKED ISSUE #9841: raw code error message shown for various errors (e.g., sessionOutdated,  challenge not found, delete Group Plan tasks)
When a Group Plan task has been assigned to you and you try to delete it, you are correctly shown a red snack notification saying you can't do that, but the error message contains all of this text:
```
{ "success": false, "error": "NotAuthorized", "message": "Can't delete group tasks that are assigned to you." }
```
I.e., you see the json object not just the `message`

We should either make that message be formatted correctly, or (probably better) prevent anyone trying to delete active Group Plan task by not showing the delete buttons for them.

TheHollidayInn:
Fixes #8636 
Fixes #9841
Fixes #8672


website/client/app.vue
```diff
@@ -370,6 +377,19 @@ export default {
     if (loadingScreen) document.body.removeChild(loadingScreen);
   },
   methods: {
+    checkForBannedUser (error) {
+      const AUTH_SETTINGS = localStorage.getItem('habit-mobile-settings');
```
paglias, line 423: looks good, just one thing: why are we parsing the auth settings again if these are stored into `store.state.credentials`? Because they cannot login at all when banned so the credentials are never stored?

TheHollidayInn: Hmmm. I figured a user would be banned while logged in but I'm sure both cases would happen. For testing locally, I banned my user while he was logged in.

Alys:
Users are most often banned while they are logged in.

TheHollidayInn:
Sweet. So is this PR good then besides the need to remove the profile name?

paglias:
LGTM 👍

website/common/locales/en/front.json
```diff
@@ -285,6 +285,8 @@
   "invalidLoginCredentialsLong": "Uh-oh - your email address / login name or password is incorrect.\n- Make sure they are typed correctly. Your login name and password are case-sensitive.\n- You may have signed up with Facebook or Google-sign-in, not email so double-check by trying them.\n- If you forgot your password, click \"Forgot Password\".",
   "invalidCredentials": "There is no account that uses those credentials.",
   "accountSuspended": "Account has been suspended, please contact <%= communityManagerEmail %> with your User ID \"<%= userId %>\" for assistance.",
+  "accountSuspendedTitle": "Account has been suspended",
+  "accountSuspendedBody": "This account, Profile Name <%= profileName %>, User ID \"<%= userId %>\", has been blocked for breaking the [Community Guidelines](https://habitica.com/static/community-guidelines) or [Terms of Service](https://habitica.com/static/terms). For details or to ask to be unblocked, please email Lemoness at <%= communityManagerEmail %> or ask your parent or guardian to email her. Please copy your Profile Name and User ID into the email.",
```
Alys: Suggested new text, which removes the profile name and also Lemoness's name so the message still works at times when she's on leave:

```
"This account, User ID \"<%= userId %>\", has been blocked for breaking the [Community Guidelines](https://habitica.com/static/community-guidelines) or [Terms of Service](https://habitica.com/static/terms). For details or to ask to be unblocked, please email our Community Manager at <%= communityManagerEmail %> or ask your parent or guardian to email them. Please copy your User ID into the email and include your Profile Name.", 
```

Also can the new text be copied into the `accountSuspended` key and have the code use that instead of `accountSuspendedBody`? That means the website and API/mobile apps will be using the same message and it also keeps the history of the message in one place in git's log, which will be useful if we ever need to look back through previous versions of the message (complaints from blocked users might occasionally make that useful).

Alys (changes requested):
This is looking great for the case where the user is already logged in! I love the new modal - very clear and easy for the user to copy out their User ID.

However if the user is already logged out and tries to log in after they've been blocked, no message comes up - the "Login" button lets you click it and causes a sync but there's no feedback about why you can't log in.

TheHollidayInn:
Okay, updated the copy and added a check for not logged in. 

We don't have the UUID when the user is not logged in, so I can't reliably check for the error message comparison. There is a 'growl' message that shows the normal response error although it isn't the best.

Alys:
These parts are working as required:
- If you become blocked while you are using Habitica, you see the modal.
- If you reload, you see the modal again.

There's still the problem where if you try to log in, you don't see a modal, you only see the growl notification, but I see your point above about not being able to get the User ID. It would be okay if that modal didn't have the User ID in it. For example, it could say:
"This account has been blocked for breaking the Community Guidelines or Terms of Service. For details or to ask to be unblocked, please email our staff at admin@habitica.com or ask your parent or guardian to email them. Please include your Profile Name."

However if you prefer, this PR could be merged as it is and the login modal could be added later. This PR is definitely an improvement on the current situation!

There's also a problem where if you're logged in at the time you are blocked, you can't get to the clear-browser-data page, which means it's difficult for another user to use Habitica on the same PC, but that problem also exists in `develop` so it could be fixed later too. We might want to discuss putting a logout button on the modal, although that might give the user too much of a hint about creating a new account to bypass the ban.

TheHollidayInn:
If it is fine with you, I think we can start with this and iterate. This is a pretty good improvement.

 And I agree about the logout. I'd rather them email us rather than pushing them to create a new account.

Alys:
I'm fine with it being included in our next batch of PRs to be merged! I've marked it ready for review in case anyone wants to do a code review again.

PR #9884, opened 2018-01-25
Added copy as todo

### LINKED ISSUE #9302: add "Copy As To-Do" to Tavern, guilds, party
The "Copy As To-Do" feature needs to be added back to **all** messages in the Tavern, all guilds, and the user's party. 

`website/client/components/chat/chatMessages.vue` is where you'd start looking into this; you'll see that there's already code there for it but it's been commented out because the copyAsTodo button wasn't working. The contributor who works on this would need to work out why and fix the problem. The feature was working on the old website so it's possible that the site's JavaScript code and/or API have support for it.

We do also want this feature to work for private messages in the Inbox, however that might need to be fixed in a different / more complex way. I'm not sure if that ever worked on the old website. If the contributor who works on this issue would like to fix it in the Inbox too, that would be great, but it's also okay if you just get it working in the Tavern, guilds, and party for now. We can open a new issue for the inbox later. If you don't get it working in the inbox, make sure that the copyAsTodo icon doesn't appear there.

TheHollidayInn:
Fixes #9302

### DIFF
--- a/website/client/components/chat/chatCard.vue
+++ b/website/client/components/chat/chatCard.vue
@@ -18,11 +18,9 @@ div
         .svg-icon(v-html="icons.like")
         span(v-if='!msg.likes[user._id]') {{ $t('like') }}
         span(v-if='msg.likes[user._id]') {{ $t('liked') }}
-      // @TODO make copyAsTodo work in Tavern, guilds, party (inbox can be done later)
-        span.action(v-if='!inbox', @click='copyAsTodo(msg)')
-          .svg-icon(v-html="icons.copy")
-          | {{$t('copyAsTodo')}}
-          // @TODO make copyAsTodo work in the inbox
+      span.action(v-if='!inbox', @click='copyAsTodo(msg)')
+        .svg-icon(v-html="icons.copy")
+        | {{$t('copyAsTodo')}}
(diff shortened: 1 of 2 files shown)


Alys:
I think this might have been merged by accident before it was finished? It has a couple of problems:

- link to user's profile doesn't work (best to not include the link until it is working because we'll get repeated bug reports)
- "group name" appears in the link to the group rather than the actual group name

I've done a reversion PR so that we don't have to rush to finish this before staging can be deployed. https://github.com/HabitRPG/habitica/pull/9959

Alys:
Actually no, sorry, I'm not reverting. The PR that github's "revert" button made looks too messy. I guess we might as well just fix this in staging. Can that be done before staging is deployed?

Alys:
This is good now!

I've found only one problem: if you copy a system message from party chat (and I assume also from the Tavern during a World Boss event), the Note says:
`undefined wrote in [PartyName](https://habitica.com/groups/guild/ae9bf3a2-8050-4253-85ba-14eab6b19e72)`
It would be better to phrase it something like:
`system message in [PartyName](https://habitica.com/groups/guild/ae9bf3a2-8050-4253-85ba-14eab6b19e72)`

PR #9885, opened 2018-01-25
Analytics: track challenge and task events

SabreCat:

### Changes

Adds tracking for the following analytics events:

* For new users (account age up to a week), creating personal tasks
* For new users (account age up to a week), scoring tasks
* Creating challenge tasks
* Creating group tasks
* Creating challenges
* Joining challenges
* Deleting challenges
* Awarding challenge winners

### DIFF
--- a/website/server/controllers/api-v3/challenges.js
+++ b/website/server/controllers/api-v3/challenges.js
@@ -760,7 +760,7 @@ api.selectChallengeWinner = {
     // Close channel in background, some ops are run in the background without `await`ing
     await challenge.closeChal({broken: 'CHALLENGE_CLOSED', winner});
 
-    res.analytics.track('challenge award', {
+    res.analytics.track('challenge close', {
       uuid: user._id,
       hitType: 'event',
       category: 'behavior',
--- a/website/server/controllers/api-v3/challenges.js
+++ b/website/server/controllers/api-v3/challenges.js
@@ -257,10 +257,12 @@ api.createChallenge = {
       privacy: group.privacy,
     };
 
-    res.analytics.track('create challenge', {
+    res.analytics.track('challenge create', {
       uuid: user._id,
       hitType: 'event',
       category: 'behavior',
+      challengeID: response._id,
+      groupID: group._id,
       groupName: group.privacy === 'private' ? null : group.name,
       groupType: group._id === TAVERN_ID ? 'tavern' : group.type,
     });
@@ -323,10 +325,12 @@ api.joinChallenge = {
     let chalLeader = await User.findById(response.leader).select(nameFields).exec();
     response.leader = chalLeader ? chalLeader.toJSON({minimize: true}) : null;
 
-    res.analytics.track('join challenge', {
+    res.analytics.track('challenge join', {
       uuid: user._id,
       hitType: 'event',
       category: 'behavior',
+      challengeID: challenge._id,
+      groupID: group._id,
       groupName: group.privacy === 'private' ? null : group.name,
       groupType: group._id === TAVERN_ID ? 'tavern' : group.type,
     });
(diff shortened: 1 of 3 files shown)


paglias (approved):

website/server/controllers/api-v3/tasks.js
```diff
@@ -241,6 +254,17 @@ api.createChallengeTasks = {
 
     // If adding tasks to a challenge -> sync users
     if (challenge) challenge.addTasks(tasks);
+
+    tasks.forEach((task) => {
+      res.analytics.track('create task', {
```
TheHollidayInn: Do you need a 7 days old check here?

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -58,6 +58,17 @@ api.createGroupTasks = {
     let tasks = await createTasks(req, res, {user, group});
 
     res.respond(201, tasks.length === 1 ? tasks[0] : tasks);
+
+    tasks.forEach((task) => {
+      res.analytics.track('create task', {
```
TheHollidayInn: 7 day age check?

website/server/controllers/api-v3/tasks.js
SabreCat: I'm thinking because adding challenge tasks is a much less frequent flow than users adding their own tasks, leaving them on would be a negligible event volume and wouldn't need such a cutoff.

That said, maybe adding tasks to challenges isn't useful information in addition to creating the challenge in the first place?

website/server/controllers/api-v3/tasks/groups.js
SabreCat: Similar to above, I think group plans are uncommon enough at the moment that we don't need to limit it. In fact, restricting to 7 days age would probably exclude the vast majority of group plan task creation activity.

website/server/controllers/api-v3/tasks.js
TheHollidayInn: Sounds good to me! Just making sure. I would like to keep these and see. We can always change.

website/server/controllers/api-v3/challenges.js
```diff
@@ -257,6 +257,14 @@ api.createChallenge = {
       privacy: group.privacy,
     };
 
+    res.analytics.track('create challenge', {
+      uuid: user._id,
+      hitType: 'event',
+      category: 'behavior',
+      groupName: group.privacy === 'private' ? null : group.name,
+      groupType: group._id === TAVERN_ID ? 'tavern' : group.type,
+    });
```
Alys: Should this record the group ID? The name alone isn't fully reliable since it can change plus its annoying to search for if we want to find the guild.
Can it have the challenge ID too? That could be useful occasionally for troubleshooting.

website/server/controllers/api-v3/challenges.js
```diff
@@ -315,6 +323,14 @@ api.joinChallenge = {
     let chalLeader = await User.findById(response.leader).select(nameFields).exec();
     response.leader = chalLeader ? chalLeader.toJSON({minimize: true}) : null;
 
+    res.analytics.track('join challenge', {
+      uuid: user._id,
+      hitType: 'event',
+      category: 'behavior',
+      groupName: group.privacy === 'private' ? null : group.name,
+      groupType: group._id === TAVERN_ID ? 'tavern' : group.type,
```
Alys: Challenge ID?

Also as a suggestion, if the analytics titles (? keys?) were "challenge create", "challenge John", etc, they'd all sort together in the Amplitude interface which would be convenient when doing several challenge investigations.

Alys: Similarly below for using "task create" rather than "create task" in case we ever add "task delete" too (or "task copy" if we build that requested feature}.

It might be too late to use "task score" since "score task" is already in use.

website/server/controllers/api-v3/challenges.js
SabreCat: Sure! I tend to think of these bits from the behavioral data side rather than troubleshooting utility, so that's a good perspective to get.

website/server/controllers/api-v3/challenges.js
SabreCat: Good points! Given we're rebooting "score task" and haven't had any data in that event for some time, I think it'll be OK to have a rename.

I'll add another commit with these changes, maybe also add "challenge delete" and "challenge award" (challenge conclude?)

Alys: Thanks!! :)

The usual term for ending a challenge with a winner is "close" (as opposed to "delete" which is ending with no winner).

SabreCat: ah, I just missed that. I'll make it "close"

TheHollidayInn (approved):

PR #9889, opened 2018-01-26
Update copyright year 2017->2018

dreeves:

Fixes put_issue_url_here

### Changes

----
UUID:

### DIFF
--- a/website/client/components/appFooter.vue
+++ b/website/client/components/appFooter.vue
@@ -83,7 +83,7 @@
         hr
     .row
       .col-12.col-md-5
-        | © 2017 Habitica. All rights reserved.
+        | © 2018 Habitica. All rights reserved.
         .debug.float-left(v-if="!IS_PRODUCTION && isUserLoaded")
           button.btn.btn-primary(@click="debugMenuShown = !debugMenuShown") Toggle Debug Menu
           .debug-group(v-if="debugMenuShown")


dreeves:
Was my last comment auto-generated or did I accidentally do something?

Anyway, for y'all's amusement, here's us Beeminder folks arguing about this:

Danny and Bee and Andy argue about automatically setting the copyright year to <%=Time.now.strftime('%Y')%>:

D: boo to automatically filling in copyright! that's how we prove we're still at the wheel :)
A: it only proves that if it's publicly visible how the copyright is being created
A: like if beeminder were open source. otherwise no one can tell the difference
D: actually i think when the copyright changes at a random time in january, that's a good signal that you're at the wheel...
D: if it changes at the stroke of midnight then obviously you cheated and could all be dead or on a beach somewhere
B: but everyone cheats
B: so it can only be an indicator in our absence
D: boo to cheaters
bystander: it's the best way to score GitHub points too: open `2017` -> `2018` pull requests
A: but who is going around checking exactly when copyrights change??
A: bots, that's who
A: okay nevermind I'm on board with this as a precaution to confuse future overlords
B: anyhow, it also makes us look stupid. like we don't know how to use string format, and causes people to write us emails like "Umm. so your copyright is out of date. I don't know what language you guys use, but in php it's really easy you.."
A: maybe that's a positive, identify users who can write code
A: "oh, thanks! here's another nerd sniping problem. can you rewrite beebrain?"
D: clearly the optimal solution is to set it to `year(now-random(5,10)*86400)` :)
D: (simulates a fairly attentive human at the wheel updating the copyright 5-10 days after new years)
D: i can't tell what fraction serious i am about doing it manually. i guess if, counterfactually, we ever weren't at the wheel it would be lies and that feels icky
D: like there's something inherently dishonest about literally being like "copyright... uh, whatever year it is when you're reading this!"
B: ok, so part of the checklist for shuttering beeminder is "update footer copyright date"
B: you clearly didn't understand my argument for why we're doing this, because you seem to remain unconvinced. go read my email again and then i'm putting you in timeout so you can think about what you've done.
D: [pastes this in the html source so we can show it to people in <%=(Time.now+86400*365.25).strftime('%Y')%> who see "<%=Time.now.strftime('%Y')%>" and think we don't know how to insert a dynamic date]
PS: Bee's argument in the email with the user who called us out:
I think Danny had some convoluted reasoning for why we shouldn't just make that programmatic, like "it's a signal that the site is still active", but I think that's kind of incorrect -- anyone writing a website can make the copyright date be programmatically generated, so it doesn't actually serve as any indicator of currency, it can only serve as a anti-activity signal by going out of date.

website/client/components/appFooter.vue
```diff
@@ -83,7 +83,7 @@
         hr
     .row
       .col-12.col-md-5
-        | © 2017 Habitica. All rights reserved.
+        | © 2018 Habitica. All rights reserved.
```
lex111, line 86: How about something like this?

```js
| © {{ new Date().getFullYear() }} Habitica. All rights reserved.
```

SabreCat:
Thanks for this, @dreeves -- do you have a Habitica UUID (see https://habitica.com/user/settings/api)? If so, post it here and we can get you contributor credit!

dreeves:
Ha, sure -- 14136054-8bd2-4f9d-b588-7ea920e8d7c0

I feel silly getting contributor credit for the most trivial possible contribution though! :)

SabreCat:
@dreeves: We tend to give Blacksmith tier 1 for a merged PR regardless of the scope, because that includes the effort you went to in cloning the repo and so on! Rest assured tier 2 won't be quite that easy. ;)

PR #9891, opened 2018-01-26
Updated notes on armoire when purchasing

### LINKED ISSUE #9342: Armoire equipment pieces remaining do not update automatically
As you buy pieces from the Armoire, the number of equipment pieces remaining does not automatically update itself in the on-hover message on the Armoire’s icon on the tasks page. You have to do a full refresh to see the new number. It should update automatically

TheHollidayInn:
Fixes #9342


website/client/components/shops/shopItem.vue
```diff
@@ -29,15 +29,14 @@ div
...
-      )
+        :item="item")
       div.questPopover(v-else-if="item.purchaseType === 'quests'")
         h4.popover-content-title {{ item.text }}
         questInfo(:quest="item")
       div(v-else)
-        h4.popover-content-title(v-once) {{ item.text }}
-        .popover-content-text(v-if="showNotes", v-once) {{ item.notes }}
-
+        h4.popover-content-title(v-once) {{ item.text }} {{item.key}}
```
paglias: do two different {{ }} block work? I thought only was permitted

paglias: do two different {{ }} block work? I thought only was permitted

TheHollidayInn: It works, but that is not intended. It was meant for debugging.

PR #9895, opened 2018-01-27
add length and character limitations for login name (username)

Alys:
Adds a limit on the length of the user's login name (`auth.local.username`) and limits the characters to a-z, A-Z, 0-9.

This applies only when creating a new Habitica account or adding local authentication to an existing account that uses social media auth.

Needs automated tests but we might want to add them later.


Alys:
I've tested this manually on a local install against `upstream/develop` and it works: it's not possible to choose a Login Name that's greater than 20 characters or that contains . or _ (I didn't test other characters since the `isAlphanumeric` function should be reliable). It does permit digits and lower case a-z letters and upper case A-Z letters.

BUT the automated tests are failing because we're generating login names with `let username = generateUUID();` which contains hyphens. Allowing hyphens should be fine for our spam-prevention purposes so I'll adjust the new code to permit them.

Alys:
This is now manually tested on my local install against `upstream/release` and it's looking good.

I can register with `123456789-abcde-FGHI` as a login name but not anything longer and not anything containing . or _ (I didn't test other characters).

I'm looking at the broken tests now.

paglias:
Changed limit to 20 chars, will update the other PR if travis passes

test/api/v3/integration/user/auth/POST-register_local.test.js
```diff
@@ -37,7 +41,8 @@ describe('POST /user/auth/local/register', () => {
       expect(user.newUser).to.eql(true);
     });
 
-    it('remove spaces from username', async () => {
+    xit('remove spaces from username', async () => {
```
TheHollidayInn, line 44: Why remove this test? We still trim the username.

TheHollidayInn: nvm. I see we invalidate at the beginning

PR #9896, opened 2018-01-27
describe Login Name limitations on Registration form and Add Local Auth form

Alys:
This modifies two forms to add a description of what characters and length are permitted for the Login Name:
- the registration form (creating a new account)
- the "Add local authentication" form when adding a Login Name (and password, email) to an existing account that uses social media login.

Here's what the registration form looks like:
2-zoomed-in
This link shows the full page:
[full registration page with login name limitations text](https://user-images.githubusercontent.com/1495809/35471362-887e2438-03a5-11e8-9ac4-6e27fd325cfc.png)

Here's what the "add local auth" form looks like:
4-zoomed
This link shows the full page:
[full settings page with login name limitations text](https://user-images.githubusercontent.com/1495809/35471360-88175898-03a5-11e8-9d40-fb4f3a9fdf45.png)
Ignore the fact that both of the social auth buttons are purple (usually one would be greyed-out). That's just a side-effect of how I tested this visual change and won't occur in the production site.

### DIFF
--- a/website/client/components/settings/site.vue
+++ b/website/client/components/settings/site.vue
@@ -118,6 +118,7 @@
           hr
           div(v-if='!user.auth.local.username')
             p {{ $t('addLocalAuth') }}
+            p {{ $t('usernameLimitations') }}
             .form(name='localAuth', novalidate)
               //-.alert.alert-danger(ng-messages='changeUsername.$error && changeUsername.submitted') {{ $t('fillAll') }}
               .form-group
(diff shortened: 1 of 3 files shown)


lex111 (approved):

Alys:
Because our test code currently uses login names that are 36 characters long and contain hyphens, I've changed the login name limitations in https://github.com/HabitRPG/habitica/pull/9895 to match. We can adjust it back later to 20 characters when we have more time, and can adjust and test the test suite to match.
I've also added underscores as permitted characters because as paglias pointed out they are commonly used, and if we're allowing hyphens, we might as well allow them too. If desired, we can later change our test code and the login name restrictions to prohibit underscores and/or hyphens.

This PR now has text changes to the screenshots shown above:

Issue #9276, opened 2017-10-26
Move from PhantomJS to Chrome Headless for client tests


paglias:
PhantomJS is giving us a lot of problem due to the fact that it's using an outdated JS engine.

We should move to Chrome Headless for this.

In a PR that's currently in progress but that's going to be merged soon we had to disable unit tests because of this, so when this issue is fixed client tests should be enabled again by changing this line in package.json (under "scripts")

```
"test": "npm run lint && gulp test && gulp apidoc",
```

to 

```
"test": "npm run lint && gulp test && npm run client:unit && gulp apidoc",
```

Important links:
- https://github.com/HabitRPG/habitica/blob/develop/test/client/unit/karma.conf.js#L19
- https://github.com/karma-runner/karma-chrome-launcher
- https://github.com/HabitRPG/habitica/pull/9236/commits/50c156d9d047198a7fb9bffc90c6069bfb456f12

Asuza:
I'd like to try this.

Alys:
@Asuza Thank you! Please comment here if you have questions or find problems.

Asuza:
Thanks, @Alys. I'm having a lot of trouble getting this working. I'm trying to use Docker (as mentioned here: http://habitica.wikia.com/wiki/Setting_up_Habitica_Locally#Docker). I've removed the phantomjs karma modules and added the one I need for Chrome. I've installed puppeteer, and had it trying to launch Chrome inside the "web" container. However, after struggling with trying to get missing dependencies installed for Chrome to function, I went looking for another container to use.

I ended up changing the _Dockerfile_'s first line to be `FROM alekzonder/puppeteer:0.12.0`... but that didn't seem to work (my knowledge of Docker is limited). Despite removing phantomjs from _package.json_ and removing the _package-lock.json_, doing a `docker-compose up --build` is **still trying to install phantomjs**.

Any suggestions on what to try now?

Alys:
@Asuza I'm afraid I'm not very familiar with Docker but some of the other admins are. Are you still having trouble with it?

Asuza:
Yes, I still don't know what's going on. I will need to learn a lot more about Docker to get any further, so hopefully the other admins know exactly what the problem is.

gilmoreg:
Has there been any progress on this? I'd really love to start writing tests for the client. Have you tried doing this without Docker? Would you like me to give it a shot?

TheHollidayInn:
[...]

PR #9898, opened 2018-01-27
Task page : task filters

### LINKED ISSUE #9582: sorting Dailies in "All" filter causes "Due" filter to be applied
As reported in the Report a Bug guild by Il Mago Azzurro (c073342f-4a65-4a13-9ffd-9e7fa5410d6b):

"Website, trying to reorder my dailies with the "All" filter active results in having the "Due" filter (the default one) activating everytime the page sync after I move something, I had to move only 2 things, but it was really longer than it should be..."

I've confirmed that in my account, although it didn't happen every time (just most of the time).

The correct behaviour is that when you are sorting Dailies using any of the All/Due/NotDue filters, the filter should never be changed automatically.

kartikadur:

Fixes https://github.com/HabitRPG/habitica/issues/9582

### Changes

All changes made to reduce response / processing times, improve data reactivity, and reduce data state ambiguity. I have run tests and checked things multiple time, but I don't expect them to be exhaustive. 

#### Existing files:
TaskColumn (column.vue):
+ store.getters access where possible instead of store.state
+ can only access its own task type instead of all tasks
+ task Labels and Filters assigned to own file (client/libs/store/helpers/filterTasks.js)
  - task label, filter labels, and active filter function assigned when component is created
+ functions adjusted where needed to incorporate above changes
+ unit tests for functions where possible

Task-Getters (store/getters/tasks.js)
+ returns all tasks of requested type
+ returns all tasks of requested type filtered by categories (all, due, notDue for dailies, etc.)
+ unit tests

User-Getters (store/getters/user.js)
+ return user buffs, preferences, and task order by task type

#### New files:
Filter-Tasks (client/libs/store/helpers/filterTasks.js)
+ contains all task filter data and return task label, filter labels, and active filter
+ unit tests

Order-Tasks (client/libs/store/helpers/orderTasks.js)
+ assigns task order by single task type for efficiency
+ unit tests

----
UUID:


TheHollidayInn:
This is looking pretty great! Thank you!

Before I review this, am I correct in seeing that you have removed the ability to set a filter based on the user's preference for now? I see your todos.

kartikadur:
The `taskFilters` object in filterTasks.js and `getFilteredTaskList()` function in tasks.js, in the code below, show how a user's preferences are passed to the filters. Is this what you are referencing?

```javascript
daily: {
  label: 'dailies',
  filters: [
    { label: 'all', filterFn: () => true, default: true },
    { label: 'due', filterFn: userPrefs => t => !t.completed && shouldDo(new Date(), t, userPrefs) },
    { label: 'notDue', filterFn: userPrefs => t => t.completed || !shouldDo(new Date(), t, userPrefs) },
  ],
},
```

```javascript
if (type === 'daily' && (filterType === 'due' || filterType === 'notDue')) {
  selectedFilter = {
    ...selectedFilter,
    filterFn: selectedFilter.filterFn(userPreferences),
  };
}
```

As far as I know only `dailies - due and notDue` required them and not `todos`.

Let me know if there were any changes to the filter functions for `todos` and I can update them accordingly before the review.

TheHollidayInn:
Oh yep! That looks good to me :D

kartikadur:
oh good, I thought I had messed up and forgotten something. :)

paglias:
Going to test this now @kartikadur . Thanks a lot for the tests, although we have them currently disabled for the client, pending a fix for https://github.com/HabitRPG/habitica/issues/9276

website/client/components/tasks/column.vue
```diff
@@ -591,35 +567,36 @@ export default {
...
+      // filter requested tasks by search text
+      if (!isEmpty(searchText)) {
+        // to ensure broadest case insensitive search matching
+        let searchTextLowerCase = searchText.toLowerCase();
+        filteredTaskList = taskList.filter(
+          task => {
+            // eslint rule disabled for block to allow nested binary expression
+            /* eslint-disable no-extra-parens */
+            return (
+              (!isEmpty(task.text) && task.text.toLowerCase().indexOf(searchTextLowerCase) > -1) ||
```
paglias: why ar you using `!isEmpty` instead of `task.note && ...` these are simple strings, not objects

paglias: And the check for `task.text` isn't necessary since we enforce its existence

website/client/libs/store/helpers/filterTasks.js
```diff
@@ -0,0 +1,83 @@
...
+      } else {
+        return filterFunction;
+      }
+    }
+  };
+}
+
+const getActiveFilter = activeFilter(taskFilters);
+
+export {
```
paglias: these should be exported separately not as a single object

website/client/libs/store/helpers/filterTasks.js
```diff
@@ -0,0 +1,83 @@
...
+
+const getFilterLabels = filterLabel(taskFilters);
+
+function activeFilter (filterList) {
+  return (type, filterType = '') => {
+    let filterListByType = filterList[type].filters;
+    if (isEmpty(filterType)) {
+      return filterListByType.find(f => f.default === true);
+    } else {
+      // check if filter type is available, else send default filter
```
paglias: is this necessary? I mean, we control which filters are activable it should never happen for the user to select a filter that's unavaible

website/client/libs/store/helpers/orderTasks.js
```diff
@@ -0,0 +1,34 @@
+import compact from 'lodash/compact';
+import isEmpty from 'lodash/isEmpty';
+
+// sets task order for single task type only.
+// Accepts task list and corresponding taskorder for its task type.
+export function orderSingleTypeTasks (rawTasks, taskOrder = []) {
+  // if there is no predefined task order return task list as is.
+  if (isEmpty(taskOrder)) return rawTasks;
+  // // eslint-disable-next-line no-console
```
paglias: let's remove thse two lines

website/client/libs/store/helpers/orderTasks.js
```diff
@@ -0,0 +1,34 @@
+import compact from 'lodash/compact';
+import isEmpty from 'lodash/isEmpty';
+
+// sets task order for single task type only.
+// Accepts task list and corresponding taskorder for its task type.
+export function orderSingleTypeTasks (rawTasks, taskOrder = []) {
+  // if there is no predefined task order return task list as is.
+  if (isEmpty(taskOrder)) return rawTasks;
```
paglias: at this point it's not necessary to pass a default to taskOrder, let's remove that and simply do `if (!taskOrder)`

website/client/store/getters/tasks.js
```diff
@@ -109,3 +116,51 @@ export function getTaskClasses (store) {
...
+  return ({
+    type,
+    filterType = '',
+  }) => {
+    // get requested tasks
+    // check if task list has been passed as override props
+    // assumption: type will always be passed as param
+    let requestedTasks = getters['tasks:getUnfilteredTaskList'](type);
+
+    let userPreferences = state.user.data.preferences || {};
```
paglias: these defaults aren't necessary, the api always return these fields

website/client/store/getters/tasks.js
```diff
@@ -109,3 +116,51 @@ export function getTaskClasses (store) {
...
+        filterFn: selectedFilter.filterFn(userPreferences),
+      };
+    }
+
+    requestedTasks = requestedTasks.filter(selectedFilter.filterFn);
+    if (selectedFilter.sort) {
+      requestedTasks = sortBy(requestedTasks, selectedFilter.sort);
+    }
+
+    // // eslint-disable-next-line no-console
```
paglias: let's remove these lines

website/client/components/tasks/column.vue
```diff
@@ -339,45 +310,49 @@ export default {
...
     }),
-    onUserPage () {
-      let onUserPage = Boolean(this.taskList.length) && (!this.taskListOverride || this.taskListOverride.length === 0);
+    ...mapGetters({
+      getFilteredTaskList: 'tasks:getFilteredTaskList',
+      getUnfilteredTaskList: 'tasks:getUnfilteredTaskList',
+      getUserPreferences: 'user:preferences',
+      getUserBuffs: 'user:buffs',
+    }),
+    // Temporarily commented till definitely known if onUserPage() is needed or not
```
paglias: Looks like it's never used (or was before this PR) so we can remove this

paglias (changes requested):
Left some comments, in general it seems to work well. Let's just not use `isEmpty` when dealing with strings :)

website/client/components/tasks/column.vue
kartikadur: I want to remove this bit of code, but wasn't sure if it was being used anywhere else. If it isn't I'll take it out.

website/client/components/tasks/column.vue
kartikadur: Force of habit trying to be thorough.

kartikadur: I tend to go overboard sometimes, thanks for the heads up

website/client/libs/store/helpers/filterTasks.js
kartikadur: activeFilter (and by extention getActiveFilter) takes two params. param 1 is the task type, while the second is optional filter type (e.g. all, due, notDue for dailies, etc.). I'll simplify the structure and remove the unit test checks for boundary conditions for unavailable filters.

website/client/libs/store/helpers/filterTasks.js
kartikadur: updated as separate exports

website/client/libs/store/helpers/orderTasks.js
kartikadur: updated as suggested

website/client/store/getters/tasks.js
kartikadur: updated as suggested

website/client/store/getters/tasks.js
kartikadur: updated as suggested

website/client/libs/store/helpers/orderTasks.js
kartikadur: updated as suggested

kartikadur:
@paglias thanks for the updates. I have a bad habit of overthinking boundary cases / worst case scenarios. So its really helpful getting pointers like these.

website/client/components/tasks/column.vue
paglias: well, it's a computed property, isn't it? I just searched if it was used anywhere in the component but let me know if I'm missing something

kartikadur: It is neither used in the same component nor in the direct parent.

onUserPage() resets the dailies filter to default and rewards filter to empty array if the user is not on their own page. Not sure if this type of functionality was designed.

paglias: cc @TheHollidayInn

paglias:
@kartikadur thanks for this, as for the other PRs if you have an Habitica UUID let me know

paglias:
Hi @kartikadur we've noticed that the user's own tasks are now appearing in the group plan tasks page. I think this is due to this PR, can you give a look? Thanks!

kartikadur:
@paglias I'm looking into it. How urgent is this required change?

paglias:
We have to find a solution by tomorrow, I can look into it this evening or
will revert the pr in the meantime

Il 21 Feb 2018 16:08, "kartik adur" <notifications@github.com> ha scritto:

kartikadur:
@paglias is there a way to do group plan tasks in a dev instance? I had accounted for the tasklistoverride in my code but I may need it to check how the code change works to be certain of where the problem is.

> the user's own tasks are now appearing in the group plan tasks page

are the user's tasks showing up instead of group-plan tasks or are they combined with the group-plan tasks?

paglias:
They're combined. You should be able to create a free group plan by trying
to create a new one and when you go to pay Stripe should have a yello test
badge somewhere, then you can use 4242 4242 4242 4242 with and CVC and
expiration date and it should work

Il 21 Feb 2018 16:18, "kartik adur" <notifications@github.com> ha scritto:

kartikadur:
Thanks.

For the moment reverting the PR till I have a chance to test and update the code would be prudent

kartikadur:
A preliminary check indicates that there should not be a problem of tasks merging in the frontend since group tasks and user tasks are treated as mutually exclusive. Any group tasks passed as props overwrite the user's tasks.

Could you provide any conditions you used for running this test / scenario where you ended up with merged group and user tasks? is there a possibility that the group + user tasks are retreived in a merged state from the API?

paglias:
Whenever you open a group page I see group tasks + user tasks (dailys always, habits appear briefly and then disappear) I checked and the api is not returning the user tasks with the group ones

schermata_2018-02-21_alle_17_34_37
schermata_2018-02-21_alle_17_34_57

paglias:
It looks like that:

1. If you have no group dailies, the user dailies are rendered
2. If you have group habits (for example) the user habits are renderend for just 1 sec and then replaced by the group habits

kartikadur:
Thanks for the update. I think I may have figured out where the problem might be. It'll take me a day or two to update the code and submit changes.

Should I submit the changes to branch that created this PR or should I make a new PR?

paglias:
New PR, i can also look into it if you don't have time

paglias:
Could the problem be in this line https://github.com/HabitRPG/habitica/pull/9898/files#diff-55116909dfc8ad199c0b7a35ba675675R334 ? I feel like it should have a condition in case we're on a group

kartikadur:
I'm attaching the changes I'll be making in the next 48 hours. 
change 1: Line 283
```JavaScript
props: ['type', 'isUser', 'searchText', 'selectedTags', 'taskListOverride', 'group'], // @TODO: maybe we should store the group on state?
```
set 'isUser' prop to have a default value. I would suggest setting it to default `false` as it would require fewer changes in the overall code.


change 2: Line 334 - 339
```JavaScript
let filteredTaskList = isEmpty(this.taskListOverride) ?
  this.getFilteredTaskList({
    type: this.type,
    filterType: this.activeFilter.label,
  }) :
  this.taskListOverride;
```
in these lines check for 'this.isUser' as `true` or `false` instead of 'this.taskListOverride' as empty or not.

kartikadur:
Let me know if its faster for you to make the changes, in which case I will leave it to you.

Alys:
@kartikadur We'll revert the PR as you suggest so that this doesn't have to be rushed. Thank you for looking into it! Please keep going if you'd like to!

kartikadur:
Ok, I'll have the new PR up by Friday.

paglias:
Reverted the PR in https://github.com/HabitRPG/habitica/commit/d5e4be85e991740e6ec622c41f7c38a307b802ea

PR #9908, opened 2018-01-29
Refactored spell code to functions

TheHollidayInn:


website/server/controllers/api-v3/user.js
```diff
@@ -520,6 +520,101 @@ api.getUserAnonymized = {
 
 const partyMembersFields = 'profile.name stats achievements items.special';
 
+async function castTaskSpell (res, req, targetId, user, spell) {
```
negue, line 523: can we move those to `ops/spells` ?

website/server/controllers/api-v3/user.js
```diff
@@ -593,101 +688,26 @@ api.castSpell = {
...
-            user.save(),
-            partyMembers.save(),
-          ]);
-        } else {
-          await partyMembers.save(); // partyMembers is user
-        }
+        await castUserSpell(res, req, party, partyMembers, targetId, user, spell);
       }
 
       let partyMembersRes = Array.isArray(partyMembers) ? partyMembers : [partyMembers];
```
negue: error on casting a transformation potion on a user in party 

```2018-02-07T20:55:03.650Z - error: TypeError: Cannot read property 'toJSON' of undefined
    at user.js:717:32
    at Array.map (native)
```

website/server/controllers/api-v3/user.js
paglias: ops in /common? that should only include stuff used both on client and server

negue: hmm I see, well sooner or later we'd need to clean up / split `user.js` since its getting tooo long ^^ , what about `api-v3/user/spell.ops.js`  or something

TheHollidayInn: My plan is to move stuff to libraries and create modules we can share. Common is bloated and unorganized so I'd like to keep stuff out of there.

I'll be working splitting the code. I'm on Challenges right now.

PR #9909, opened 2018-01-29
Logout url change fixes #9438

### LINKED ISSUE #9438: Going directly to /logout doesn't clear the local login status
If an user goes directly to /logout without passing from the button on the website they'll be logged out from the server but the client will still think that they're logged in.

The problem is that the /logout link must be reachable on the server, one option could be to rename the logout route on the server to something difficult to type or to require some weird query parameter to be passed that cannot be guessed.

pollaro:

[//]: # 9438
Fixes https://github.com/HabitRPG/habitica/issues/9438

### Changes
[//]: # changed /logout to /logoutServer in website/client/store/actions/auth.js, server/controllers/top-level/auth.js, webpack/config/index.js (this one may have only been needed for development)

----
UUID:


website/client/store/actions/auth.js
```diff
@@ -68,5 +68,5 @@ export async function socialAuth (store, params) {
 export function logout () {
   localStorage.removeItem(LOCALSTORAGE_AUTH_KEY);
   localStorage.removeItem(LOCALSTORAGE_SOCIAL_AUTH_KEY);
-  window.location.href = '/logout';
+  window.location.href = '/logoutServer';
```
TheHollidayInn: Can you turn this in an axios request? We don't want to redirect the users to the server page, right @paglias?

TheHollidayInn:
Looking good so far :D

website/client/store/actions/auth.js
pollaro: Ok. This shouldn't be too hard.

website/client/store/actions/auth.js
```diff
@@ -68,5 +68,5 @@ export async function socialAuth (store, params) {
 export function logout () {
   localStorage.removeItem(LOCALSTORAGE_AUTH_KEY);
   localStorage.removeItem(LOCALSTORAGE_SOCIAL_AUTH_KEY);
-  window.location.href = '/logout';
+  axios.get('/logoutServer').then(() => window.location.href = '/');
```
TheHollidayInn: Sorry, one more pick :P We use async/await instead of `.then` so you will want to change this function to `async`
Then 
```
await axios.get('/logoutServer');
window.location.href = '/';
```

website/client/store/actions/auth.js
paglias: actually it should be a redirect , the server will redirect to the home page

TheHollidayInn:
It looks like maybe you had an issue with rebasing? I'd suggest merging in develop if you have conflicts rather than rebasing. If you can revert and force push that would help, otherwise it might be best to just create a new PR :P

pollaro:
Yes. Sorry. I made a mistake in order of operations.

TheHollidayInn:
No worries!

PR #9915, opened 2018-01-30
Logout changes

pollaro:

[//]: # 9438/9909
https://github.com/HabitRPG/habitica/pull/9909
### Changes
[//]: # 
Changed logout address on server to '/logoutServer'
Changed from a DOM call to an axios call
Still need to redirect DOM to home in Vue because server is not doing it

----
UUID:


website/client/store/actions/auth.js
```diff
@@ -65,8 +65,9 @@ export async function socialAuth (store, params) {
   localStorage.setItem(LOCALSTORAGE_AUTH_KEY, userLocalData);
 }
 
-export function logout () {
+export async function logout () {
   localStorage.removeItem(LOCALSTORAGE_AUTH_KEY);
   localStorage.removeItem(LOCALSTORAGE_SOCIAL_AUTH_KEY);
-  window.location.href = '/logout';
+  await axios.get('/logoutServer');
```
paglias, line 71: I think this should be reverted to a redirect to `logoutServer ` instead of an axios call.

paglias (changes requested):
I'm not sure this is the right solution because we're just shifting the problem to a different route

website/client/store/actions/auth.js
TheHollidayInn: Okay, sorry @pollaro. @paglias will chat offline. I may have miscommunicated the solution.

paglias:
@pollaro talked with @TheHollidayInn and we think this is a good direction but there is one more thing to do: there should be a route on the client (see client/router.js) that calls the logout function so that if you type habitica.com/logout it still works

paglias:
how is this going @pollaro ? let me know if you need any help here

pollaro:
Hey. Sorry, I've been busy this week. I'll get on this either today or over the weekend.

pollaro:
Actually, I am having issues with this. I can't figure out how to clear the local storage from the router.js file.

paglias:
@pollaro you can either create a component linked to /logout that clears localstorage in the `created` function or use a function that gets executed before the route changes https://router.vuejs.org/en/advanced/navigation-guards.html

pollaro:
@paglias I'm adding a function before the route changes, but Express isn't finding a route I created. What am I missing? It looks like it is supposed to automatically scan the routes. Right?

paglias:
@pollaro that's right, routes should be recognized automatically. Can you post the code you're working on so that I can give alook?

pollaro:
/home/pollaro/Projects/habitica/website/server/controllers/top-level/auth.js
/home/pollaro/Projects/habitica/website/client/router.js
/home/pollaro/Projects/habitica/website/client/store/actions/auth.js

These are the only things I've changed. At least on this edit

pollaro:
I did a new pull request since I messed this one up

PR #9919, opened 2018-02-01
Changed mage and wizard's name as it was using the same name and tran…

rhiroshi:


Alys:
Thank you for wanting to help with the translations! However we can't accept translation work in GitHub because it's done in Transifex. Please see the _README_FIRST.md file in the locales/pt-br directory for details.  If you'd like to join the translation team on Transifex, that would be great!

One thing to watch out for is that translations should never be based on the `key` (`wizard` in this case), but instead should be based on the English version of the string. You can see in the English file that both strings are meant to be identical:
https://github.com/HabitRPG/habitica/blob/develop/website/common/locales/en/character.json#L125-L126
Transifex's interface displays the English strings with the translated strings to make accurate translations easier.

FYI, `wizard` was the old name for the `mage` class in the early versions of Habitica and so `wizard` still exists in some places in the code and content, which is why the `wizard` string is still needed, but with the modern "Mage" translation.

PR #9921, opened 2018-02-01
Refactor spells

paglias:
Create listeners only when cast start and user has to select target, remove them when the spell is casted


paglias:
Will move the base to develop when @TheHollidayInn 's PR is merged back into develop

PR #9744, opened 2017-12-25
WIP Partial fix for #8735 Failed Invites

JoshSears:
[//]: # (Note: See http://habitica.wikia.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API for more info)

[//]: # (Put Issue # or URL here, if applicable. This will automatically close the issue if your PR is merged in)
Partial Fix #8735

### Changes
[//]: # (Describe the changes that were made in detail here. Include pictures if necessary)
Altered /test/api/v3/integration/groups/POST-groups_invite.test.js to reflect new output
Altered /website/common/locales/en/groups.json - Added username and UUID outputs to failed invites messages
Altered /website/server/controllers/api-v3/groups.js - sends username and uuid to error messages

[//]: # (Put User ID in here - found on the Habitica website at User Icon > Settings > API)

----
UUID: 348b9f3a-83ef-4865-b4d1-8f64870c9734

PR #9939, opened 2018-02-03
Partial Fix for #8735 ("Failed party invitations do not indicate which invitations failed")

### LINKED ISSUE #9841: raw code error message shown for various errors (e.g., sessionOutdated,  challenge not found, delete Group Plan tasks)

JoshSears:

[//]: #8735  (Put Issue # or URL here, if applicable. This will automatically close the issue if your PR is merged in)
Partial fix for https://github.com/HabitRPG/habitica/issues/8735

### Changes
Altered /test/api/v3/integration/groups/POST-groups_invite.test.js to reflect new output
Altered /website/common/locales/en/groups.json - Added username and UUID outputs to failed invites messages
Altered /website/server/controllers/api-v3/groups.js - sends username and uuid to error messages
Updated branch from main to fix merge conflicts from #9744 (orig PR) 

----
UUID: 348b9f3a-83ef-4865-b4d1-8f64870c9734

-----

Edit by paglias:

- Fixes #9841


website/common/locales/en/groups.json
```diff
@@ -223,11 +223,11 @@
...
   "partyMustbePrivate": "Parties must be private",
-  "userAlreadyInGroup": "User already in that group.",
+  "userAlreadyInGroup": "User \"<%= username %>\" already in that group. \"<%= userId %>\"",
   "cannotInviteSelfToGroup": "You cannot invite yourself to a group.",
-  "userAlreadyInvitedToGroup": "User already invited to that group.",
-  "userAlreadyPendingInvitation": "User already pending invitation.",
-  "userAlreadyInAParty": "User already in a Party.",
+  "userAlreadyInvitedToGroup": "User \"<%= username %>\" already invited to that group.\"<%= userId %>\"",
+  "userAlreadyPendingInvitation": "User \"<%= username %>\" already pending invitation. \"<%= userId %>\"",
+  "userAlreadyInAParty": "User \"<%= username %>\"  already in a party.\"<%= userId %>\"",
```
negue: could you remove the extra space in the sentence, and add the user id in this way ` User-ID: {userId}`, that way people know it's the UserID and not a random number

negue (changes requested):
Thanks for your fix 👍 

It works but it doesn't look that great for me, I mean the error json that is visible as toast.

I think we should also change the way our generic api errors are visible to the website.

See: https://github.com/HabitRPG/habitica/blob/develop/website/client/app.vue#L232

if `error.response.data.message` contains a value you could call `this.error(error.response.data.message);` otherwise use the current code.

That way all messages would be displayed  in a more readable kind

JoshSears:
@negue Thank you for the review.  I am playing with the error messages right now, but could you explain more about the "this.error" portion of your suggested change?  Or point out a portion in the code where that is used, so I have something to compare that to?

negue:
`this.error` is a method from `mixins/notifications.js` which is mixed in with `vue`, that way you can use one of its method, here `this.error` .

 in `https://github.com/HabitRPG/habitica/blob/develop/website/client/app.vue#L233` you only have `error.response.data` but `.data` may have a message that why you also need to check if a message exists.

Once everything work we would see `User "test" already invit..`  instead of `{ok: false, message: "...` in an error notification

JoshSears:
So app.vue is where the error interceptor is, and I see where line 233 has the error.response.data.  Is this where we would check for, then possibly display any info in data.message?  Or would that be where the return is going back to?

paglias:
@Malcorr I would try to change line 233 so that to this:

```
          text: error.response.data && error.response.data.messages ? error.response.data.messages : error.response.data,
```

paglias:
It would be great if you could fix this too but actually it's a more generalized problem, see https://github.com/HabitRPG/habitica/issues/9841, not related to this PR in particular

paglias:
```
        const errorData = error.response.data;
        const errorMessage = errorData.message || errorData;

        this.$store.dispatch('snackbars:add', {
          title: 'Habitica',
          text: errorMessage,
          type: 'error',
          timeout: true,
        });
```

^ a bit better

JoshSears:
@paglias That fix seems to do it.  For readability, I am moving the UserID first, because it is about the length of the error display, so the first line will be "UserID:..." second line will read "User ""name" already..."

paglias:
Sounds good 👍

JoshSears:
Got a new error found by TravisCI, working with local build to find and fix source.  I am getting an error regarding signaling for async completion in the unit tests.  Am going to update from origin, then try the test again without the app.vue changes.

paglias:
It happens sometimes, I've restarted the build, not a problem with the pr

JoshSears:
Oh, good.  I couldn't figure out what I did.

JoshSears:
Had to fix merge conflict: app.vue line 421: "await this.castStart(this.selectedSpellToBuy, member);"

negue:
@Malcorr I guess you already fixed the conflict right? cause I can't see any in this PR

paglias:
@Malcorr  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #9959, opened 2018-02-06
Revert "Added copy as todo"

Alys:
Reverts HabitRPG/habitica#9884

- link to user's profile doesn't work (best to not include the link until it is working because we'll get repeated bug reports)
- "group name" appears in the link to the group rather than the actual group name


Alys:
I'm closing this, not merging. Git's "revert" button on the original PR produced this, and even though it shows code changes for only that PR, it seems to be saying that it's reverted a lot of other commits too. I don't trust it enough to merge this.

PR #9960, opened 2018-02-06
Remove user wrapping

### LINKED ISSUE #9146: Remove user wrapping from common code
`user.{fns/ops}...` is not used anymore by the new client so the code for it can be removed: there are some tests that still use the wrapping code and an occurence in the code itself.

paglias:
Fixes #9146


website/server/libs/cron.js
```diff
@@ -428,8 +428,8 @@ export function cron (options = {}) {
   // Add 10 MP, or 10% of max MP if that'd be more. Perform this after Perfect Day for maximum benefit
   // Adjust for fraction of dailies completed
   if (dailyDueUnchecked === 0 && dailyChecked === 0) dailyChecked = 1;
-  user.stats.mp += _.max([10, 0.1 * user._statsComputed.maxMP]) * dailyChecked / (dailyDueUnchecked + dailyChecked);
-  if (user.stats.mp > user._statsComputed.maxMP) user.stats.mp = user._statsComputed.maxMP;
+  user.stats.mp += _.max([10, 0.1 * common.statsComputed(user).maxMP]) * dailyChecked / (dailyDueUnchecked + dailyChecked);
```
negue, line 431: I think this `common.statsComputed(user).maxMP` could be saved to a local variable first, so that we need to calculate it three times

paglias: yeah I was wondering the same but didn't want to change the behavior here since before _statsComputed called internally the fn statsComputed

negue: hmmm well `maxMP` won't change in the two lines but sure improve the execution speed

PR #9964, opened 2018-02-06
Challenge small refactor

TheHollidayInn:
This is a small refactor to fix some issues listed in CodeClimate


website/server/controllers/api-v3/challenges.js
```diff
@@ -264,12 +200,7 @@ api.createChallenge = {
       _id: user._id,
       profile: {name: user.profile.name},
     };
-    response.group = { // we already have the group data
-      _id: group._id,
-      name: group.name,
-      type: group.type,
-      privacy: group.privacy,
-    };
+    response.group = getChallengeGroupResponse();
```
negue: shouldn't you add `group` as argument?

PR #9967, opened 2018-02-07
Fix typo

einomi:


paglias:
Thanks for the PR @einomi but we only accept changes to the English locales in the repo, any other language change should go through Transifex. You can find more info here https://github.com/HabitRPG/habitica/blob/develop/website/common/locales/README.md

PR #9977, opened 2018-02-09
Loaded profile when not cached

TheHollidayInn:


website/client/components/chat/chatMessages.vue
```diff
@@ -197,8 +197,14 @@ export default {
 
       return false;
     },
-    showMemberModal (memberId) {
-      const profile = this.cachedProfileData[memberId];
+    async showMemberModal (memberId) {
+      let profile = this.cachedProfileData[memberId];
```
paglias, line 201: what if profile is undefined? Like for old messages before the avatar is fetched?

TheHollidayInn: You can't click the user if it isn't loaded or if msg.userStyles doesn't exist. I can add the check if you want

TheHollidayInn: https://github.com/TheHollidayInn/habitrpg/blob/c924fa9446293a0c18bab575a6745d6b5d3165f9/website/client/components/chat/chatMessages.vue#L14

paglias:
Ah perfect!

Il 09 Feb 2018 16:39, "Keith Holliday" <notifications@github.com> ha
scritto:

PR #9983, opened 2018-02-10
task types: add new error messages for invalid task types

### LINKED ISSUE #10027: Tasks Docs API Typo
**Reported in the bug guild**

There is a little bug in the API. If you use [Get user data](https://habitica.com/apidoc/#api-Task-GetUserTasks) with the param type and use "daily", it will not work, even if the answer says, that you can use "daily", "todo", etc. , but in reality it is always the plural "habits", "dailys":
```
{
    "success": false,
    "error": "BadRequest",
    "message": "Invalid request parameters.",
    "errors": [
        {
            "message": "Task type must be one of \"habit\", \"daily\", \"todo\", \"reward\".",
            "param": "type",
            "value": "daily"
        }
    ]
}
```

imrehg:
* `/task/user` `GET` endpoint takes plurals of task types as allowed parameters, plus  another custom `completedTodos` type
* `/tasks/group/:groupId` and `/tasks/challenge/:challengeId` `GET` endpoints task plurals of task types as allowed parameters

### Changes

Updating the message with the current accepted task types for the relevant API calls

----
UUID: 677a7368-89f8-48e7-98f6-77795efd3937

-------

Fixes #10027

### DIFF
--- a/website/common/locales/en/tasks.json
+++ b/website/common/locales/en/tasks.json
@@ -149,6 +149,8 @@
   "taskAliasAlreadyUsed": "Task alias already used on another task.",
   "taskNotFound": "Task not found.",
   "invalidTaskType": "Task type must be one of \"habit\", \"daily\", \"todo\", \"reward\".",
+  "invalidTasksType": "Task type must be one of \"habits\", \"dailys\", \"todos\", \"rewards\".",
+  "invalidTasksTypeExtra": "Task type must be one of \"habits\", \"dailys\", \"todos\", \"rewards\", \"completedTodos\".",
   "cantDeleteChallengeTasks": "A task belonging to a challenge can't be deleted.",
   "checklistOnlyDailyTodo": "Checklists are supported only on Dailies and To-Dos",
   "checklistItemNotFound": "No checklist item was found with given id.",
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -287,7 +287,7 @@ api.getUserTasks = {
   async handler (req, res) {
     let types = Tasks.tasksTypes.map(type => `${type}s`);
     types.push('completedTodos', '_allCompletedTodos'); // _allCompletedTodos is currently in BETA and is likely to be removed in future
-    req.checkQuery('type', res.t('invalidTaskType')).optional().isIn(types);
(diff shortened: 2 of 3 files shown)


paglias:
Which route is returning this error? because I'm pretty sure the task type as stored in `task.type` is the singular version

imrehg:
@paglias `/api/v3/tasks/user` returns this error:

```
$ curl "https://habitica.com/api/v3/tasks/user?type=todo" \
  -s -X GET --compressed -H "Content-Type:application/json" \
  -H "x-api-user: ${USER}" \
  -H "x-api-key: ${TOKEN}" | jq .
...
{
  "success": false,
  "error": "BadRequest",
  "message": "Invalid request parameters.",
  "errors": [
    {
      "message": "Task type must be one of \"habit\", \"daily\", \"todo\", \"reward\".",
      "param": "type",
      "value": "todo"
    }
  ]
}
```

And they are definitely plural in the API:

https://github.com/HabitRPG/habitica/blob/037280601a91d5b13f9614a16847f587135df9ff/website/server/controllers/api-v3/tasks.js#L268-L273

paglias:
@imrehg you're right sorry. But there's a small problem because that error is also used in server/libs/taskManager#85 where it should use the singular types. So maybe we can create a new error for the plural version and use that

imrehg:
Okay, I'll take a look from that angle too, and update this PR with more info.

imrehg:
@paglias I've updated the patch and followed your recommendation, adding the new messages as needed for the relevant calls. What do you think this way?

paglias:
This PR fixes #10027

paglias:
@imrehg could you also fix the docs as reported in #10027?

imrehg:
@paglias I checked out the docs, and those are correct, am I missing something?

screenshot_2018-02-21_17-10-33

The issue you link to seems to report the discrepancy of the API and the API error message - exactly the thing that this PR is fixing up.

Basically the API docs seem to be correct (as are the tests as I checked), only the the error messages provided by the API were out of step.

paglias:
Sorry, you're completely right, this is working well and will be merged next week!

SabreCat:
Thanks for this, @imrehg, and welcome to the ranks of the Blacksmiths at tier 1!

Note that further tiers require increasing amounts of work from one to the next, so tier 2 may require multiple PRs or a larger PR to attain. But keep helping out and we'll express our gratitude accordingly!

PR #9988, opened 2018-02-12
If message uuid is set as system append as user 'system message' for Copy as To-Do #9985

### LINKED ISSUE #9985: Copy As To-Do for system messages shouldn't say "undefined"
When you use the Copy As To-Do feature on a system message (e.g., one of the red-on-white messages showing a player's quest progress), the To-Do's Note says:

`undefined wrote in [PartyName](https://habitica.com/groups/guild/12345678-8050-4253-1234567890abc)`

It should instead be phrased as:

`system message in [PartyName](https://habitica.com/groups/guild/12345678-8050-4253-1234567890abc)`

someoneigna:
Fixes #9985 

### Changes
When using "Copy As Todo" from a chat Card if the user uuid set in the message is system, add "system message" as the user name for the todo text.

UUID: a120835b-5382-44eb-b6d7-5c597d329c0e

### DIFF
--- a/website/client/components/chat/copyAsTodoModal.vue
+++ b/website/client/components/chat/copyAsTodoModal.vue
@@ -38,7 +38,7 @@ export default {
   },
   mounted () {
     this.$root.$on('habitica::copy-as-todo', message => {
-      const notes = `${message.user} wrote in [${this.groupName}](${baseUrl}/groups/guild/${this.groupId})`;
+      const notes = `${message.user || 'system message'}${message.user ? ' wrote' : ''} in [${this.groupName}](${baseUrl}/groups/guild/${this.groupId})`;
       const newTask = {
         text: message.text,
         type: 'todo',


Alys (commented):
I like it! Nice and simple.

One thing though: can you make the condition `if (!message.user && message.uuid === 'system')`
At the moment there's no need to do the extra check but in future it's possible that we'd assign names to some system messages (e.g., if we introduce messages for new purposes - mods are considering one such case now) and then this extra check would prevent a bug that overwrites the name.

someoneigna:
Left message.user untouched this time and put the formatting right in the "notes" template. Didn't know string templates allowed conditions inside.

paglias:
is this ok now @Alys ?

Alys:
I'll test locally tomorrow.

Alys:
This works as required. I've tested on system messages in a Party and the Tavern. Thank you @someoneigna! This will be merged soon and will be noted towards a contributor tier then.

paglias:
@someoneigna  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #9972, opened 2018-02-08
Replace PhantomJS with ChromeHeadless for unit test support

kartikadur:
[//]: # (Note: See http://habitica.wikia.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API for more info)

[//]: # (Put Issue # or URL here, if applicable. This will automatically close the issue if your PR is merged in)
Fixes #9276 

### Changes
[//]: # (Describe the changes that were made in detail here. Include pictures if necessary)
#### karma.conf.js
+ browser from PhantomJS to ChromeHeadless
#### package.json (package.lock.json by extension)
+ remove phantomjs-prebuilt
+ remove karma-phantomjs-launcher
+ add karma-chrome-launcher
#### .travis.yml
+ chrome addon support
+ unit test support

_todo: add docker support for chrome browser_

[//]: # (Put User ID in here - found on the Habitica website at User Icon > Settings > API)

----
UUID:

PR #9992, opened 2018-02-12
HOTFIX: purchase marketGear

negue:
While working on the purchase-refactoring part 2 - I discovered a bug where marketGear wasn't able to be purchased in the website (the API still working as expected)

its just the needed `type` was `gear` instead of `marketGear`

I hope to refactor it so something like this shouldn't happen again.

- `potion` still work after this change

This should be merged before the next deploy.

### DIFF
--- a/test/client/unit/specs/store/actions/shops.js
+++ b/test/client/unit/specs/store/actions/shops.js
@@ -1,0 +1,55 @@
+import axios from 'axios';
+import generateStore from 'client/store';
+
+import content from 'common/script/content';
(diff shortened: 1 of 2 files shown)


TheHollidayInn:
Can you add a test to cover this issue?

negue:
@TheHollidayInn there is now a test, but none of those are tested in the travis builds :/

paglias:
@negue they'll be enabled when https://github.com/HabitRPG/habitica/pull/9972 is merged

PR #9996, opened 2018-02-14
Upgrade client build tools

paglias:


paglias:
Problem seems to be that the viewBox attribute is being removed from svgs

PR #9999, opened 2018-02-14
World boss client

SabreCat:

Adds client components for the 2018 World Boss, enabling display of the WB in the Tavern.

### Changes

* Adds various image assets for the Dysheartener
* Adds a section to the Tavern that appears when a world boss quest is active and displays its quest info


website/client/assets/svg/information-red.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
```
paglias, line 1: for chevron red and this can't we use the existing files and change the color property using css? (Will have to remove the fill from the first path to  be able to override it)

SabreCat: Ah, possibly. I did try that, but ended up with a double stroke using both colors--probably due to what you said here, "have to remove the fill from the first path".

PR #10004, opened 2018-02-15
[WIP] refactor purchase api - part 2

negue:
Goal:
- extract / abstract every purchasable operation into classes
- use a factory to get the needed class
- each operation has its own checks but also the inner checks for each currency-type
- checks to prevent bulk purchases on types that can't be purchased in bulk

Additional Changes:
- [x] hide quantity selection for hourglass pet/mount purchases
- [x] always show currency (was hidden on non bulk purchases) 

Todo: 
- [x] Armoire [Gold]
- [x] Market Gear [Gold]
- [x] Health Potion [Gold]
- [x] Quest [Gold]
- [x] Special Spell [Gold]
- [x] Mount [Hourglass]
- [x] Pets [Hourglass]
- [x] Quest [Gem]
- [x] Mystery Set [Hourglass]
- [ ] Spell [Gem]
- [ ] Eggs [Gem]
- [ ] Hatching Potions [Gem]
- [ ] Food [Gem]
- [ ] Gear [Gem]
- [ ] Bundles [Gem]
- [ ] Buy Gem [Gold]
- [ ] create / use factory
- [ ] fix / add new **TESTS!**


TheHollidayInn:
This looking really good! I'd really like to do two things when we overhaul this much though:

1. Break this down into much smaller PRs. For example just do one function in common code
2. Ensure that there is 100% test coverage if not on the client.

This is what I've been doing for chat and flagging.

negue:
I see, ok. Will do!

I'll create an overview issue to catch the current state of the refactoring.

PR #10006, opened 2018-02-16
Updated membercount checks

TheHollidayInn:


website/server/libs/amazonPayments.js
```diff
@@ -270,10 +270,8 @@ api.subscribe = async function subscribe (options) {
   let priceOfSingleMember = 3;
 
   if (groupId) {
-    let groupFields = basicGroupFields.concat(' purchased');
-    let group = await Group.getGroup({user, groupId, populateLeader: false, groupFields});
-
-    amount = sub.price + (group.memberCount - leaderCount) * priceOfSingleMember;
+    const membersCount = await User.count({guilds: this._id}).exec();
```
paglias: What if the group is a party?

TheHollidayInn: Good point :P

paglias: might be a good idea to extract this check in a reusable function, in particular if it becomes more complex when you have to distinguish between parties and guilds

TheHollidayInn: Yea, I was planning on doing that after you suggested the party

PR #10018, opened 2018-02-18
Modals reappear after being closed #9986

### LINKED ISSUE #9986: modals reappear after being closed
EDIT May 2018: New reports start here: https://github.com/HabitRPG/habitica/issues/9986#issuecomment-377483866

This has been a well-known problem for some time now but I've just realised we don't have an issue for it (unless I'm just not able to find it).

A modal will often re-open itself after you've closed it. We see a lot of reports of this each time there's a new Bailey message and it also happens for other modals. For example, create a challenge, then before you add any tasks, click on a party member to view their profile. When you close the profile screen, the challenge creation modal will reappear.

Sometimes the only way to get a modal to stop reappearing seems to be reloading the website. 

I'm marking this as a high priority since it's a significant annoyance, especially for the case of the Bailey messages. 

We do already have https://github.com/HabitRPG/habitica/issues/9969 for the class change modal reappearing, which is probably just one example of this bug, but I'm leaving it open too just in case it does need a separate fix (and it's definitely high priority since it can cause a user to waste gems).

someoneigna:
Fixes #9986 

### Changes

Prevent most of the modals reappearing after dismissing unrelated modals.

UUID: a120835b-5382-44eb-b6d7-5c597d329c0e

### DIFF
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -321,12 +321,12 @@ export default {
     }
 
     // Manage modals
-    this.$root.$on('bv::show::modal', (modalId, data = {}) => {
-      if (data.fromRoot) return;
+    this.$root.$on('bv::show::modal', ({modalId, fromRoot = false, alreadyTracked = false}) => {
+      if (fromRoot) return;
 
       // Track opening of gems modal unless it's been already tracked
       // For example the gems button in the menu already tracks the event by itself
-      if (modalId === 'buy-gems' && data.alreadyTracked !== true) {
+      if (modalId === 'buy-gems' && alreadyTracked !== true) {
         Analytics.track({
           hitType: 'event',
           eventCategory: 'button',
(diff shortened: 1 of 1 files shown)


paglias:
Hi @someoneigna thanks for this PR, can you comment on the changes you made, how they solve the problem, etc?

paglias:
@TheHollidayInn what do you think?

website/client/app.vue
```diff
@@ -321,12 +321,12 @@ export default {
     }
 
     // Manage modals
-    this.$root.$on('bv::show::modal', (modalId, data = {}) => {
-      if (data.fromRoot) return;
+    this.$root.$on('bv::show::modal', ({modalId, fromRoot = false, alreadyTracked = false}) => {
```
paglias, line 324: isn't modalId passed as the first parameter when bootstrap-vue emits this event?

TheHollidayInn:
This cleans some code up, but I don't think it fixes any issues. I'm pretty sure the issue with the modals is that we dismiss them incorrectly and the events don't fire. But if you were able to recreate the issues consistently and can show how this fixes, I'd be happy to see :D

paglias:
@someoneigna can you explain you solution? thanks!

paglias:
Going to close this, @someoneigna if you'd like to reopen this just leave a comment

PR #10022, opened 2018-02-19
Logout Changes #9915

### LINKED ISSUE #9438: Going directly to /logout doesn't clear the local login status

pollaro:
Fixes #9438

#9915 
[//]: # I originally changed the URL the logout went to, but this still was not ideal. Now, I'm checking the server, but the route is not being found

[//]: # 9915
### Changes
[//]: # Change auth on both the client and server to check server side for login info. Express is not finding the new route to check this.

----
UUID:


website/client/router.js
```diff
@@ -286,6 +299,8 @@ const store = getStore();
 router.beforeEach(function routerGuard (to, from, next) {
   const isUserLoggedIn = store.state.isUserLoggedIn;
   const routeRequiresLogin = to.meta.requiresLogin !== false;
+  const isUserLoggedServer = store.dispatch('auth:loginCheck');
```
paglias: this function is async, you should wait for it with await but it also means a full request will be made for every route change on the client which is indesiderable. You should add a route guard function just to the speicific logout route

paglias (changes requested):

paglias:
The route on the server is correct but you're not awaiting the function correctly. But honestly I'm not sure why you added that route

paglias:
Looking better! Can you remove the commented out code?

pollaro:
Sure, I thought I did that. I'll get to it, shouldn't take long...

website/client/components/logout.vue
```diff
@@ -0,0 +1,21 @@
+<template lang="pug"></template>
+
+<style>
```
paglias: can you remove these empty blocks?

website/client/components/logout.vue
```diff
@@ -0,0 +1,21 @@
...
+
+</style>
+
+<script>
+export default {
+  components: {},
+  data () {},
```
paglias: data and computed should be removed as they're not used

website/client/store/actions/auth.js
```diff
@@ -65,8 +65,8 @@ export async function socialAuth (store, params) {
   localStorage.setItem(LOCALSTORAGE_AUTH_KEY, userLocalData);
 }
 
-export function logout () {
+export async function logout () {
   localStorage.removeItem(LOCALSTORAGE_AUTH_KEY);
   localStorage.removeItem(LOCALSTORAGE_SOCIAL_AUTH_KEY);
-  window.location.href = '/logout';
+  await axios.get('/logoutServer');
```
paglias: this shouldn't also redirect the user to the homepage? So instead of doing axios.get why not use `window.location.href = '/logoutServer';	 `?

website/server/controllers/top-level/auth.js
```diff
@@ -24,7 +24,7 @@ api.resetPasswordSetNewOne  = {
 // Logout the user from the website.
 api.logout = {
   method: 'GET',
-  url: '/logout',
+  url: '/logoutServer',
```
paglias: to better follow our urls conventions this should be `/logout-server`

paglias (changes requested):

pollaro:
Ok, I'll get to this in the next few days.

lemoness:
@pollaro Hi there! How is this fix going?

pollaro:
Slowly. Sorry, I was sick and in bed for a week, and then out of town. I'll have time tomorrow and Thursday

pollaro:
I applied the changes paglias suggested. Everything worked ok on my local

Yutsuten:
@pollaro Just notifying you: there are some conflicts to be resolved.

pollaro:
Sorry, I didn't see these. I've been trying to find a job.

website/server/libs/spells.js
```diff
@@ -1,9 +1,6 @@
 import { model as User } from '../models/user';
 import * as Tasks from '../models/task';
-import {
```
paglias: can you revert the changes to this file since they seem unrelated?

website/client/router.js
```diff
@@ -7,45 +7,46 @@ import getStore from 'client/store';
 
 // TODO Dummy elements used as placeholder until real components are implemented
 import ParentPage from './components/parentPage';
-
+import Logout from './components/logout';
```
paglias: are all these changes necessary?

website/client/app.vue
```diff
@@ -210,6 +211,7 @@ export default {
     BuyModal,
     SelectMembersModal,
     amazonPaymentsModal,
+    Logout,
```
paglias: we're importing the logout component here but not using it anywhere, can you remove it?

paglias (dismissed):

pollaro:
Alright, lets give this one a go.

website/server/middlewares/notFound.js
```diff
@@ -14,6 +14,7 @@ const TOP_LEVEL_ROUTES = [
   '/export',
   '/email',
   '/qr-code',
+  '/logout'
```
paglias: why was this added here? This way if you go to habitica.com/logout the client won't be loaded, maybe you wanted to add /logout-server here?

beffymaroo:
@paglias @pollaro How's this going? Any label changes or status changes needed?

pollaro:
Sorry, I've been busy and this got lost in the shuffle. I'll get it in the next day or two

beffymaroo:
@pollaro Great! Thanks for checking in.

SabreCat:
app.vue appears to have been committed with merge conflict markers, and there's a bunch of odd irrelevant changes. I'll see if I can figure out how to reconcile them from my end, but @pollaro, you might know better what was an intended change and not!

SabreCat:
OK, that was messy and a bit forceful, but I think I've got it. @pollaro and @paglias, how does this look? Local test seemed to check out, both visiting /logout and using the Log Out menu option take the user out of app and back to static page. And the only unrelated change is a negligible but valid code style fix in website/server/controllers/top-level/auth.js.

paglias:
Looks good @SabreCat , if it works let's merge it!

SabreCat:
Thanks for this, @pollaro, and welcome to the ranks of the Blacksmiths at tier 1! This will undergo some additional testing in a staging environment, then if all goes well be live around the end of next week.

For contributor tiers, note that tiers require increasing amounts of effort from one to the next, so tier 2 will require multiple PRs or a larger PR to attain. But keep helping out and we'll express our gratitude accordingly!

PR #10029, opened 2018-02-20
raise coverage for tasks api calls

negue:
- updates a group task => approval is required
=> https://coveralls.io/builds/15703336/source?filename=website%2Fserver%2Fcontrollers%2Fapi-v3%2Ftasks.js#L480
- updates a group task with checklist
 => https://coveralls.io/builds/15703336/source?filename=website%2Fserver%2Fcontrollers%2Fapi-v3%2Ftasks.js#L489


test/api/v3/integration/tasks/groups/PUT-group_task_id.test.js
```diff
@@ -50,6 +50,58 @@ describe('PUT /tasks/:id', () => {
...
+        {
+          text: 'checklist 1',
+        },
+      ],
+    });
+
+    await user.post(`/tasks/${task._id}/assign/${member._id}`);
+
+    // change the checklist text
+    task = await user.put(`/tasks/${task._id}`, {
```
paglias, line 86: shouldn't there be a check after this action? otherwise why we're doing it?

test/api/v3/integration/tasks/groups/PUT-group_task_id.test.js
```diff
@@ -50,6 +50,58 @@ describe('PUT /tasks/:id', () => {
...
+    // score up to trigger approval
+    await expect(member.post(`/tasks/${syncedTask._id}/score/up`))
+      .to.eventually.be.rejected.and.to.eql({
+        code: 401,
+        error: 'NotAuthorized',
+        message: t('taskApprovalHasBeenRequested'),
+      });
+  });
+
+  it('updates a group task with checklist', async () => {
```
paglias: There are no tests here

paglias (changes requested):

website/server/controllers/api-v3/tasks.js
```diff
@@ -464,8 +464,6 @@ api.updateTask = {
 
     if (!challenge && task.userId && task.challenge && task.challenge.id) {
       sanitizedObj = Tasks.Task.sanitizeUserChallengeTask(updatedTaskObj);
-    } else if (!group && task.userId && task.group && task.group.id) {
```
TheHollidayInn: We still need this condition. You can remove the `!group` check, but we still need to run that satanize check if the tasks is a group task. I see it is slightly confusing because of the `sanitizeUserChallengeTask` name, but basically the function is shared between challenges and groups.

You can add a different check `if ((!challenge && task.isChallengeTask) || task.isGroupTask)` which would be an improvement.

negue: ended up using `if (!challenge && task.isChallengeTask || task.group && task.group.id)`

negue:
could someone restart travis ?

paglias:
Restarted

paglias:
Can you pull in the most recent develop? Seems to have some problems otherwise

paglias (approved):

paglias:
@negue merged in develop but the tests are still failing

negue:
Hmm weird, seems that a group leader can't even change the task name, was something changed lately?

Also when I wrote the test `updates a group task - approval is required` it did work haha, now not anymore :/

paglias:
Not sure what's going on here, aspecially it looks like something broke that was working previously. No idea, really. @TheHollidayInn can you give a quick look?

kirsty-tortoise:
If people are still looking at this, it seems the change in `website/server/controllers/api-v3/task.js` to the condition for running `sanitizeUserChallengeTask` is what is causing the tests to fail and reverting that line fixes the tests.

In particular, on this branch when you run the failing tests, `sanitizeUserChallengeTask` is run, and this seems to stop any user from changing the task name. Previously, the condition meant `sanitizeUserChallengeTask` was not run, and the tests passed. But possibly it should be running and should be checking who the leader is somewhere later, or the condition may need changing to check who the leader is, or the tests should be sending data differently to update the group task, I'm not sure.

paglias:
thanks for the investigation @kirsty-tortoise !

website/server/models/task.js
```diff
@@ -169,9 +169,14 @@ TaskSchema.statics.findByIdOrAlias = async function findByIdOrAlias (identifier,
...
-  let initialSanitization = this.sanitize(taskObj);
-
-  return _.pick(initialSanitization, ['streak', 'checklist', 'attribute', 'reminders', 'tags', 'notes', 'collapseChecklist', 'alias', 'yesterDaily', 'counterDown', 'counterUp']);
+  const initialSanitization = this.sanitize(taskObj);
+
+  return _.pick(initialSanitization, [
+    'streak', 'checklist', 'attribute', 'reminders',
+    'tags', 'notes', 'text', 'collapseChecklist',
+    'alias', 'yesterDaily', 'counterDown', 'counterUp',
+    'up', 'down',
```
paglias: is it necessary to add the up and down to this whitelist?

paglias:
Left one comment, but the tests seem to be failing anyway

website/server/models/task.js
negue: at first I thought this would fix all the issues, but it is by-design to not allow changes to `text` / `up`/`down` and I assumed it was allowed, so when I wrote the tests ~(these sanitization weren't available back then) it worked right away.~ I didn't had the changed condition to sanitize the task obj, the issues started after that change.

Now that I revert this change, removed all others changes of text,up,down and reverted the sanitization conditions, all tests works fine again.

paglias:
thanks!

PR #10043, opened 2018-02-23
Marked new stuff complete when clicking later

TheHollidayInn:
When a user clicks new stuff later, we add a notification to their account. However, we don't mark newstuff completed as false which can cause the bailey to show multiple times.

### DIFF
--- a/website/client/components/achievements/newStuff.vue
+++ b/website/client/components/achievements/newStuff.vue
@@ -50,7 +50,6 @@
     methods: {
       tellMeLater () {
         this.$store.dispatch('user:newStuffLater');
-        this.$store.dispatch('user:set', {'flags.newStuff': false});
         this.$root.$emit('bv::hide::modal', 'new-stuff');
       },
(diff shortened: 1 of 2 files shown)


website/client/components/achievements/newStuff.vue
```diff
@@ -50,6 +50,7 @@
     methods: {
       tellMeLater () {
         this.$store.dispatch('user:newStuffLater');
```
paglias: Hmm this route on the server marks user.flags.newStuff as false, so maybe the only thing to do is to make sure it's marked as false on the client too? We can just add it to the user:newStuffLater action here

paglias (changes requested):

website/client/components/achievements/newStuff.vue
TheHollidayInn: Ah, okay. That makes sense. I'll move it to the action then.

paglias (approved):

PR #10046, opened 2018-02-23
Chat approval

### LINKED ISSUE #8063: Optional moderator approval for chat posts from all new users
We want to have a setting to temporarily turn off the ability of new users to post before allowed to do so by a moderator. This would be very helpful in situations where a single user (or several trolls) tries to make many accounts in a row to dodge a ban and spam chat.

Details:
1. All existing users are always allowed to post.
2. Under normal conditions, all new users are allowed as soon as they've created an account (just as they can now) BUT:
3. There should be a flag (an environment variable?) that mods can use to cause every new user account to be unable to post until a mod enables posting for that user. Mods could turn on that flag when a troll keeps creating new accounts. Mods would turn the flag off later when they believed that the troll might have moved on.
4. When a user who is not yet allowed to post tries to post, their message appears in the chat area but visible only to mods (and it's sent to mods in an email). There'd be an icon on the message that mods could use to make the message visible to everyone and to enable normal posting for that user. This behaviour would occur only while the flag was turned on.

EDIT 2021-11-10: One possible solution is in https://github.com/HabitRPG/habitica/pull/12499 but note the comments there from paglias!

TheHollidayInn:
Fixes #8063


website/common/locales/en/groups.json
```diff
@@ -468,5 +468,9 @@
   "howToRequireApprovalDesc2": "Group Leaders and Managers can approve completed Tasks directly from the Task Board or from the Notifications panel.",
   "whatIsGroupManager": "What is a Group Manager?",
   "whatIsGroupManagerDesc": "A Group Manager is a user role that do not have access to the group's billing details, but can create, assign, and approve shared Tasks for the Group's members. Promote Group Managers from the Group’s member list.",
-  "goToTaskBoard": "Go to Task Board"
+  "goToTaskBoard": "Go to Task Board",
+  
+  "approvalNotRequired": "Approval not required",
+  "messagedFlaggedNotHidden": "Message flagged once, not hidden",
+  "messageHidden": "Message hidden"
```
Alys: These three new strings could go into website/server/libs/apiMessages.js since they don't need to be translated.

website/server/models/group.js
```diff
@@ -539,6 +540,17 @@ function setUserStyles (newMessage, user) {
   newMessage.markModified('userStyles');
 }
 
+function checkForChatApproval (group, message, user) {
+  if (!config.isChatApprovalRequired()) return;
+  if (group.privacy !== 'public') return;
+
+  const userCreatedDate = moment(user.auth.timestamps.created);
+  const oneDayAgo = moment().startOf('day').subtract(1, 'days');
+  if (userCreatedDate.isBefore(oneDayAgo)) return;
```
Alys: Should oneDayAgo be given a more generic name in case we change the number of days later? Perhaps something like cutOffDate?

Also should 1 be a constant? That would allow us to reference the constant in the tests - for example, for this line, instead of `2` we could use the constant + 1:
` +    const twoDaysAgo = moment().startOf('day').subtract(2, 'days').toDate();`

It might even be good to use an ENV var instead of a constant so that we can change it easily to cover the case in the second paragraph at https://github.com/HabitRPG/habitica/issues/8063#issuecomment-353176169 (a group of trolls joins a few days in the past then starts spamming).

Although this is making the PR more complex so if you want to just leave it as it is we can add customisation later. :)

website/client/components/chat/chatCard.vue
```diff
@@ -1,8 +1,9 @@
 <template lang="pug">
 div
   .mentioned-icon(v-if='isUserMentioned')
-  .message-hidden(v-if='msg.flagCount === 1 && user.contributor.admin') Message flagged once, not hidden
-  .message-hidden(v-if='msg.flagCount > 1 && user.contributor.admin') Message hidden
+  .message-hidden(v-if='msg.flagCount === 1 && user.contributor.admin') {{ $t('messagedFlaggedNotHidden') }}
+  .message-hidden(v-if='msg.flagCount > 1 && user.contributor.admin') {{ $t('messageHidden') }}
+  .message-hidden(v-if='approvalRequired', @click='approve()') {{ $t('approvalRequired') }}
```
Alys: I'm not sure we should reuse the `approvalRequired` string. It's used for task approvals in group plans and if requirements ever change for that, we'd want to be able to change the text without affecting chat message approvals. Maybe the string could be `approvalIsRequired` for consistency with the `approvalNotRequired` string?

If that is changed, it might also be helpful for future coding if the functions and variables called `approvalRequired` are changed to be the same as the string, so that we can git grep for that one phrase and find all the relevant code, without the group task approval code appearing too. But that's a pretty minor point so ignore it if you wish. :)

website/server/models/group.js
```diff
@@ -547,6 +559,8 @@ schema.methods.sendChat = function sendChat (message, user, metaData) {
 
   if (user) setUserStyles(newChatMessage, user);
 
+  checkForChatApproval(this, newMessage, user);
```
Alys: This should be `newChatMesage` not `newMessage`. With `newMessage` the `true` value is never assigned to the `approvalRequired` field in the message. 

Also that field needs to be added to website/server/models/chat.js

website/server/controllers/api-v3/chat.js
```diff
@@ -480,4 +480,58 @@ api.deleteChat = {
...
+
+    const validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    if (!user.isAdmin()) throw new NotAuthorized(res.t('noAdminAccess'));
+
+    const group = await Group.getGroup({user, groupId});
+    if (!group) throw new NotFound(res.t('groupNotFound'));
+
+    const message = find(group.chat, {id: req.params.chatId});
```
Alys: This is throwing `ReferenceError: find is not defined` when a mod approves a message.

website/client/components/chat/chatMessages.vue
```diff
@@ -135,6 +137,9 @@ export default {
       if (!message.flagCount || message.flagCount < 2) return true;
       return this.user.contributor.admin;
     },
+    canViewUnapprovedChat (message) {
+      return !this.approvalRequired(message) || this.user.contributor.admin;
+    },
```
Alys, line 142: I'm checking with the other mods to see if we also want to let the user see their own message. That wasn't specified in the issue but I've realised from my testing that it's confusing when you make a post and don't see it appear in the chat.

website/server/controllers/api-v3/chat.js
```diff
@@ -480,4 +480,58 @@ api.deleteChat = {
...
+/**
+ * @api {post} /api/v3/groups/:groupId/chat/:chatId/approve Approve a group chat message
+ * @apiName ApproveChat
+ * @apiGroup Chat
+ * @apiDescription Approvates a chat message from a group
+ *
+ * @apiParam (Path) {UUID} groupId The group _id ('party' for the user party and 'habitrpg' for tavern are accepted)
+ * @apiParam (Path) {UUID} chatId The chat message _id
+ *
+ * @apiSuccess {Object} data The approved <a href='https://github.com/HabitRPG/habitica/blob/develop/website/server/models/group.js#L51' target='_blank'>chat message</a>
```
Alys: Should that URL be replaced with https://github.com/HabitRPG/habitica/blob/develop/website/server/models/chat.js now?

config.json.example
```diff
@@ -113,5 +113,6 @@
       "CLOUDKARAFKA_PASSWORD": "",
       "CLOUDKARAFKA_TOPIC_PREFIX": ""
     },
-    "STACK_IMPACT_KEY": "aaaabbbbccccddddeeeeffffgggg111100002222"
+    "STACK_IMPACT_KEY": "aaaabbbbccccddddeeeeffffgggg111100002222",
+    "CHAT_APPROVAL_REQUIRED": "false"
```
Alys: Is there any way we could turn this feature on and off using a setting in the Hall? That would allow moderators to manage it fully. If that's possible but difficult, it wouldn't need to be done in this PR.

I guess it would pretty much need to be a setting in the database, which would complicate things. A new collection? A value in the Tavern's document? I suppose we don't really have support for this sort of thing at the moment.

website/common/locales/en/groups.json
TheHollidayInn: I can move the server side one. The other two are on the client so we don't have access to that lib there.

website/client/components/chat/chatCard.vue
TheHollidayInn: I can change it to approvalIsRequired. Not really sure about the functions. Seems like organizing the code into the chat libs would be better, and I'll probably skip that to not create more work.

config.json.example
TheHollidayInn: Yea, we don't have any good support for this.

website/common/locales/en/groups.json
Alys: For those two client-side strings, it might be best to have them hard-coded into the vue files. That would save the translators translating them unnecessarily and would slightly reduce the volume in Transifex. Alternatively, perhaps we could have a `doNotTranslate.json` locales file that's never uploaded to Transifex.

website/client/components/chat/chatMessages.vue
Alys: We've decided that the user should be able to see their own post before it's approved, and should see a message on it indicating that it's pending approval. That will avoid confusion about why their post doesn't appear / why no one is replying and will hopefully prevent them posting the same message over and over again to get attention. Details are here in the issue: https://github.com/HabitRPG/habitica/issues/8063#issuecomment-385624004

Alys (changes requested):

PR #10047, opened 2018-02-23
Don't include seasonal class gear in classless category

### LINKED ISSUE #9971: Seasonal class-specific equipment lost to Death is shown under Class and under None
_From Il Mago Azzurro (c073342f-4a65-4a13-9ffd-9e7fa5410d6b) in the Report a Bug guild_

When you loose a seasonal equipment item for death in the market it appears both under your class and under None/Classless... This happened to me with the Candy Corn Shield (2017 Autumn Gear).

Screenshots: [here](https://www.awesomescreenshot.com/image/3159617/21e79e3f5a5cfd97e2610a3133ba90e1) you can see the piece under the Class:None section; [here](https://www.awesomescreenshot.com/image/3159619/2dd747e11aa92e7cc62bda4344ac75bf) under Class:Warrior (my class).

SabreCat:

Fixes #9971 

### Changes

Removes some logic from the Market library that was adding broken seasonal equipment to the "None" category in the Market. Not only was this needless and confusing, but it posed problems for upcoming functionality planned for Habitica's mobile apps.

Prompts the question of why this logic was in there in the first place, since it looks so deliberate... o.O

### DIFF
--- a/website/common/script/libs/shops.js
+++ b/website/common/script/libs/shops.js
@@ -189,8 +189,8 @@ shops.getMarketGearCategories = function getMarketGear (user, language) {
 
   let specialNonClassGear = filter(content.gear.flat, (gear) => {
     return !user.items.gear.owned[gear.key] &&
-      content.classes.indexOf(gear.klass) < 0 &&
-      content.classes.indexOf(gear.specialClass) < 0 &&
+      content.classes.indexOf(gear.klass) === -1 &&
+      content.classes.indexOf(gear.specialClass) === -1 &&
       (gear.canOwn && gear.canOwn(user));
   });
 
--- a/website/common/script/libs/shops.js
+++ b/website/common/script/libs/shops.js
@@ -187,16 +187,14 @@ shops.getMarketGearCategories = function getMarketGear (user, language) {
     text: i18n.t('none', language),
   };
 
-  let falseGear = filter(content.gear.flat, (gear) => {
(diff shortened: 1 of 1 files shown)


website/common/script/libs/shops.js
```diff
@@ -187,16 +187,14 @@ shops.getMarketGearCategories = function getMarketGear (user, language) {
...
-  let falseGear = filter(content.gear.flat, (gear) => {
-    let prevOwnedFalseGear = user.items.gear.owned[gear.key] === false && gear.klass !== user.stats.class;
-    let specialNonClassGear = !user.items.gear.owned[gear.key] &&
-                              content.classes.indexOf(gear.klass) < 0 &&
-                              content.classes.indexOf(gear.specialClass) < 0 &&
-                              (gear.canOwn && gear.canOwn(user));
-    return  prevOwnedFalseGear || specialNonClassGear;
+  let specialNonClassGear = filter(content.gear.flat, (gear) => {
+    return !user.items.gear.owned[gear.key] &&
+      content.classes.indexOf(gear.klass) < 0 &&
```
paglias: what does indexOf < 0 means? if we're matching for not found, you should use === -1, same in the line below.

(This works but I think it's quite confusing, === -1 is what we use elesewhere)

paglias (dismissed):

website/common/script/libs/shops.js
SabreCat: Ah, yeah. This was done by the previous author of these lines--I'll happily refactor that to standard.

PR #10053, opened 2018-02-26
Task page - task filters v2

### LINKED ISSUE #9582: sorting Dailies in "All" filter causes "Due" filter to be applied

kartikadur:

Fixes #9582, #9898

### Changes
Task Column Component:
+ getters instead of state
+ each column instance only gets data for one task type
+ differentiates between group-plan and user tasks
+ dailies default filter only set when component is created / first loaded

Task Getters:
+ returns all tasks or filtered list of tasks by type

User Getters:
+ returns user buffs, user preferences, and task order by task type

Task Filters
+ data structure for task filters
+ helper methods to get filter names, functions, etc. by task type

Order Tasks
+ returns task order by task type

Unit tests
+ vue/test-utils pkg as dev-dependency

----
UUID:


paglias:
Sorry for the delay here, can you fix the conflicts and merge the most recent develop in?

kartikadur:
hope thats all it needs :crossed_fingers:

PR #10058, opened 2018-02-27
Removed uri for client side oauth

### LINKED ISSUE #8127: Some pages prevent oauth through Google and Facebook
General Info
- UUID: 
- Browser: Chrome (53)
- OS: Windows 10 (1607)
### Description

It would seem that there are pages on the site from which you cannot authenticate. /static/plans being at least one of them.
error 400 oauth2 error 1 - google chrome 2016-10-12 11 26 11
### Update from @paglias

The solution here is to always redirect to the same page when logging in so we don't have to track which URLs can be used for login, and adding one entry for the settings page for when trying to add Google authentication to an existing account.

This is going to affect Facebook by Februrary / March 2018

TheHollidayInn:
Fixes #8127

### DIFF
--- a/website/client/components/settings/site.vue
+++ b/website/client/components/settings/site.vue
@@ -241,7 +241,9 @@ export default {
     hello.init({
       facebook: process.env.FACEBOOK_KEY, // eslint-disable-line no-process-env
       google: process.env.GOOGLE_CLIENT_ID, // eslint-disable-line no-process-env
-    }, {redirect_uri: ''});
+    }, {
+      redirect_uri: '', // eslint-disable-line
+    });
   },
   computed: {
     ...mapState({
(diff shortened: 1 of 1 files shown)


website/client/components/settings/site.vue
```diff
@@ -241,7 +241,7 @@ export default {
     hello.init({
       facebook: process.env.FACEBOOK_KEY, // eslint-disable-line no-process-env
       google: process.env.GOOGLE_CLIENT_ID, // eslint-disable-line no-process-env
-    });
+    }, {redirect_uri: ''});
```
paglias: what does this do? If it redirects to the current page it's going to have the same problems as before. I think we should redirect always to the same page which we know is authorized (so either the home page or a specific page used only for this)

TheHollidayInn: You don't need redirects for client side oauth. This prevents that and fixes the problem. You can test it locally without this and with.

paglias: 👍

Issue #9797, opened 2018-01-13
when buying new contributor gear, item does not disappear from Shop


Alys:
_This bug is currently only visible on our private staging server (or on anyone's local test install). It can't be seen in production because of issue https://github.com/HabitRPG/habitica/issues/9179, but once PR https://github.com/HabitRPG/habitica/pull/9306 is deployed to production, I imagine we'd see this bug there too._

When you are given a contributor tier, the associated gear is available for you to purchase in the Shop. The purchase is successful (the item appears in your Inventory immediately) but the bought item does not disappear from the Shop until you reload the website (a sync is not enough). I.e., it looks as if you haven't been able to buy the item, which will be confusing and frustrating to the new contributors.

Note that if you are buying the item because you previously owned it and lost it due to death, then it DOES disappear from the Shop as soon as you buy it. So this bug only happens if the item was not already stored in your account with a `false` value.

_BTW, when testing a fix for this bug, you should be aware that you can't buy the contributor sword at all because of this bug:  https://github.com/HabitRPG/habitica/issues/9731#issuecomment-354382219 . You'd need to test with the other contributor gear pieces._

PR #10064, opened 2018-02-28
[Input Needed] Filter out owned items from User Rewards

### LINKED ISSUE #10049: equipment does not become unpinned when purchased; next gear does not become pinned
As posted by Kajakas (40e9985b-8acb-4110-aeb9-a5946a83a229) in the Report a Bug guild:

**When I buy a piece of class equipment, it stays pinned in the rewards column (and looks as if it's locked) while the next piece of equipment in the series does not automatically get pinned.**

I've confirmed that. If you reload, the bought item disappears and the next one becomes pinned, however the expected behaviour is that as soon as you've bought it, it should become unpinned and then next one should appear without a reload being needed.

Hus274:
Adding an if statment to user rewards to filter out any already owned rewards from the Rewards section of the page when an item is purchased.

Fixes #10049
### Changes
I've added an additional condition to the loop of items displayed in the rewards section to filter out any items that are already owned.

----

### DIFF
--- a/website/client/components/tasks/column.vue
+++ b/website/client/components/tasks/column.vue
@@ -52,6 +52,7 @@
       .reward-items
         shopItem(
           v-for="reward in inAppRewards",
+          v-if="!reward.owned",
           :item="reward",
           :key="reward.key",
           :highlightBorder="reward.isSuggested",


Alys:
Does this also result in the next item in the sequence appearing?

Hus274:
Ahh it does not. I apologize, I've been struggling to get everything up to date a running and overlooked that section. I'll take a look at that and make the addition here when I'm done.

Sorry for the unneeded back and forth here.

Alys:
No worries, it's cool. :)

Hus274:
So after scratching my head over this for a good chunk of time, it looks like the root of the problem is surfacing from here:
https://github.com/HabitRPG/habitica/blame/develop/website/client/store/actions/shops.js#L21-L24 and is looks like an extension of the issue reported in #9797. 

From my research, it actually looks like the core of the problem is that the API isn't returning the new view of the user.pinnedItems from the server to the client. I can only see three options remaining at the moment:
1. Duplicate [this server side logic](https://github.com/HabitRPG/habitica/blob/develop/website/common/script/ops/buy/buyGear.js#L55) on the client side to keep the client view in sync
2. Force a refresh of the page after a purchase
3. Update the server to returned the pinnedItems view [here](https://github.com/HabitRPG/habitica/blob/develop/website/common/script/ops/buy/buyGear.js#L79)

I think that option three is the right solution here to minimize the code duplication and prevent the client view and client code from getting out of sync with the server but I wanted to make sure there weren't any major concerns with adding an additional value to the return of the request on the API before I moved forward with that.

Let me know your thoughts!

Hus274:
I've reworked this problem in #10085. Closing this one in favor of the other. Had to close this because I needed to get a clean clone from github.

PR #10078, opened 2018-03-01
Prettify background overview

### LINKED ISSUE #10076: Background overview could use some work

### General Info
  * UUID: 6d7d6651-604d-4e7c-adff-a040770a6768
  * Browser: Chrome
  * OS: Linux

### Description
Background sets that the user has purchased line out differently from sets not yet owned, as seen below. 
image

Also, the backgrounds page became very bloated since the recent overhaul. Totally pretty and responsive and awesome, but bloated. You have to scroll for half an hour to find the background you're looking for, and 90% of the page is white space. 

Some simple fixes, like boxes for the sets, and consolidating the title and controls, should spruce things up nicely. 

#### Console Errors

probaton:

Fixes #10076
### Changes
- Set the width of background sets to 100% to ensure both purchased and non-purchased sets align vertically.
- A variety of css tweaks to make the backgrounds overview clearer, more concise, and, if I may say so myself, prettier. 
- Bonus feature: a checkbox at the top of the backgrounds overview to filter out the backgrounds the user hasn't unlocked/purchased yet! Having to scroll through a million backgrounds in four different tabs to find the one you know you have is a thing of the past! This has been a thorn in my side ever since I bought my very first background, so I'd love for this to make the review. However, if there's some intrinsic reason no one has ever done this then I can easily revert the last commit and leave you with a prettier overview that is functionally identical to the one we have now. 

#### Prettier!
image
#### Bonus feature: now you see me.
image
#### Bonus feature: now you don't!
image

----

### DIFF
--- a/website/client/components/creatorIntro.vue
+++ b/website/client/components/creatorIntro.vue
@@ -261,12 +261,9 @@ b-modal#avatar-modal(title="", :size='editing ? "lg" : "md"', :hide-header='true
         .col-2.text-center.sub-menu-item(@click='changeSubPage("2014")', :class='{active: activeSubPage === "2014"}')
           strong(v-once) 2014
       .row.customize-menu(v-for='(sets, key) in backgroundShopSetsByYear')
-        .row(v-for='set in sets', v-if='activeSubPage === key')
+        .row.background-set(v-for='set in sets', v-if='activeSubPage === key')
           .col-8.offset-2.text-center.set-title
             strong {{set.text}}
-            button.buy-set-btn.btn.btn-secondary(v-if='!ownsSet("background", set.items) && set.identifier !== "incentiveBackgrounds"' @click='unlock(setKeys("background", set.items))') Purchase Set
-              .svg-icon.gem(v-html='icons.gem')
-              span 15
           .col-4.text-center.customize-option.background-button(v-for='bg in set.items',
             @click='!user.purchased.background[bg.key] ? backgroundSelected(bg) : unlock("background." + bg.key)',
             :popover-title='bg.text',
@@ -274,16 +271,19 @@ b-modal#avatar-modal(title="", :size='editing ? "lg" : "md"', :hide-header='true
             popover-trigger='mouseenter')
             .background(:class='[`background_${bg.key}`, backgroundLockedStatus(bg.key)]')
             i.glyphicon.glyphicon-lock(v-if='!user.purchased.background[bg.key]')
-            .purchase-single(v-if='!user.purchased.background[bg.key]')
+            .purchase-background.single(v-if='!user.purchased.background[bg.key]')
               .svg-icon.gem(v-html='icons.gem')
-              span 7
+              span.price 7
             span.badge.badge-pill.badge-item.badge-svg(
               :class="{'item-selected-badge': isBackgroundPinned(bg), 'hide': !isBackgroundPinned(bg)}",
               @click.prevent.stop="togglePinned(bg)",
               v-if='!user.purchased.background[bg.key]'
             )
               span.svg-icon.inline.icon-12.color(v-html="icons.pin")
-
+          .purchase-background.set(v-if='!ownsSet("background", set.items) && set.identifier !== "incentiveBackgrounds"' @click='unlock(setKeys("background", set.items))')
+            span.label Purchase Set
+            .svg-icon.gem(v-html='icons.gem')
+            span.price 15
 
   .container.interests-section(v-if='modalPage === 3 && !editing')
     .section.row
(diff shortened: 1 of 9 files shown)


paglias:
cc @Tressley

Tressley:
@probaton -- Thanks for taking a stab at this! I like the direction you're heading and think that this is great start. My only feedback is to remove the box-shadow from the gray containers, update the padding-bottom for a set you can no longer purchase, and to use a toggle instead of a checkbox.

Here's a mock-up to give you a better idea:

image

probaton:
@Tressley Thanks for the feedback. I've bumped the bottom margin, as well as restricting the box-shadow to the buttons. Also replaced the checkbox; I was sort of on the fence myself but now that I'm looking at the end result the toggle-switch definitely looks better. 

In a stroke of mischief I did remove the dashed line from toggle-switch labels. They always felt really random to me. The only possible reason I could think of for a dashed underline was to indicate a link (perhaps to an information page?) but that isn't the case. Obviously I'm totally new here though, so if I'm overstepping my bounds feel free to call me on it. I specifically added it in it's own commit, so removing it is two minutes work. 

Not sure how precisely you were hoping I would re-create your mock-up. In the end I went with my gut left in what I felt was prettier. Lemme know if I left anything out that you felt strongly on.

Tressley:
@probaton --Good eye! The underline on labels associated with a toggle was a carry over from one of our first toggle use cases/implementations. It's there to highlight a contextual tooltip around the control. I'd just be sure you're not overwriting where it is necessary (Stable/Equipment).

negue:
LGTM 👍

probaton:
@Tressley You convinced me to take a second look and I did actually miss an instance of a toggle-switch with a tool-tip. Exactly the one you mentioned in fact; the one for auto-equipping equipment in the equipment page drawer. The reason other toggle-switches aren't affected by this is because they implement their own label instead of using the built-in one. 

That left me with two choices that made sense to me. 
1) We can live with the fact that the auto-equip toggle-switch has no visual indicator for the fact that it has a hover. That's the simple solution; it means we can simple merge as is. 
2) We pull the hover element into the toggle-switch component. Always applying the underline, which it used to do, makes no sense at all. As such, we need to make the component aware of the fact that there may be a hover and only apply the underline when there is. I actually built this in a separate branch if you want to take a look. The code doesn't really suffer, but I'm still hesitant to implement it this way because it means extending a generic component to support a single use case. 

Lemme know which option you'd prefer!

Link to 'fix': https://github.com/probaton/habitica/commit/efc7a8d9bf831b7ab00a5065a042b413d8fbfa33

paglias:
@probaton I'd say adding support the hover text is okay

librarianmage:
@probaton @paglias @Tressley I've fiddled around with the toggle-switch in #10098; maybe there's something in there that might help?

paglias:
I think @probaton 's solution is okay, we just need to be sure it can be generalized and applied to every use case

probaton:
@paglias @Tressley Pushed my toggle-hover fix. I've checked out all usages of the toggle-switch and it covers all of them. The only scenario that isn't covered is if anyone ever wants a toggle-tool-tip in any other orientation than top, as the orientation is currently hard-coded. However, from a 'build for today' perspective it didn't seem reasonable to add a bunch of complexity to configure orientation when we might never need it. In any case adding an extra prop to the toggle-switch, or replacing `hoverText` with a `hoverOpts` object, isn't rocket surgery. 

I also took the liberty of squeezing in a little extra fix: toggle-switch had a *required* `label` prop, which made absolutely no sense as 8/10 instances were passing `""`. `label` is no longer required, which means I could get rid of a bunch of code clutter, as well as pruning the DOM of a bunch of empty divs.

probaton:
I feel like my last fix may have fallen between the cracks. Is this story still on someone's radar?

negue:
I rechecked the toggle switches too, code looks good, but the toggles itself are now a bit lower than before:
toggleswitchmargin

if you check the current toggles in market  you'll see a difference in the top padding/margin

probaton:
@negue Arg! Apparently there was a sneaky `margin-top` to `toggle-switch` that was only relevant in a single use-case and was being removed manually everywhere else. 

The bright side is that I could get rid of a bunch of redundant classes that were only being used to correct something that should never have been there in the first place. The slightly-less-bright-side is that  fixing the exception (the equipment drawer header) was way more of a pain than it should have been. 

I ended up adding `display: inline-block` to the equipment drawer tabs, which was the prettiest ugly fix I could come up with. A neater solution would have been to use the `DrawerHeaderTabs` component, which someone very helpfully built, but I would have to overhaul the component to make it more generic for that to work. I don't particularly mind doing that, but I don't really have the time right now, so I'd prefer just pushing this as is. However, if you feel strongly about it I can take a look later this week.

probaton:
Pushed a rebase to solve a conflict with my own fix. No changes to the code were made.

SabreCat:
This is lovely, @probaton! Thank you and welcome to Blacksmith tier 2!

PR #10082, opened 2018-03-02
Adding the ability for admins to revoke/reinstate chat privileges and block/unblock users from the profile page.

### LINKED ISSUE #10073: Mod Tool: Block/mute button directly in user profiles
Right now, users who are problematic can only be blocked or muted via the Hall of Heroes, which takes a lot of time. Usually this is okay, but sometimes there are situations of urgency where we need to  move as quickly as possible to protect the community.  In those situations, it would be helpful to have a block/mute button in the profile box. This button should ONLY be visible to moderators and staff.

This doesn't need to have a particularly polished design, as only staff and moderators see it, but they should be easily visible.

Hus274:
This change adds buttons to the profile viewer modal that allow the admins to control block/unblock and chat privilege status quickly and easily. I've attached pictures below that show these new icons.

fixes #10073

### Changes
I think there are two key decisions that I thought through that could change here. One is I didn't regionalize the tool tips for the admin buttons because only the admins should have access to them and the hall of heroes messages for the same don't appear to be regionalized. This should be pretty trivial to fix if we want these to be regionalized.

The second thing to consider is the fact that the profile page shows the block and unblock buttons simultaneously instead of only showing the button that changes the user's status (eg. only show the blocked button if the user in unblocked). The reason I did this is because the user object on the client side here doesn't have the user.flags or the user.auth.blocked fields of the user object and I don't know that it's something you want the clients to be able to access with some minor javascript work (mostly an information security thing) so I simply allow the property to be set with these idempotent operations regardless of the current state.

### Non-admin view
screen shot 2018-03-01 at 4 58 22 pm
### New Admin View
screen shot 2018-03-01 at 4 58 28 pm
### Tooltips
screen shot 2018-03-01 at 4 59 11 pm
screen shot 2018-03-01 at 4 59 13 pm
screen shot 2018-03-01 at 4 59 16 pm
screen shot 2018-03-01 at 4 59 18 pm

Let me know if you have any feedback!


Alys:
The tooltips don't need to be regionalised, but thank you for asking!

For myself personally, I don't mind that both sets of buttons show and I'd prefer that to having more complex code that's needed for only the sake of this moderator feature. I'll point the other mods at this and ask how they feel. Thanks for the screenshots!

Alys:
Oh, although... I was assuming that the buttons would be greyed out or not depending on the current status of the user as in the "New Admin View" screenshot, but I've just realised that that's not possible without knowing the state of the user, is it? Mods will need to see if the user has already been blocked/muted or not.

Hus274:
Yep, that makes sense to me. I can look at loading that section of the user in to make it so the buttons can react to the current status of the user.

Hus274:
Ok, I've updated the page to have a button that basically puts the page into admin editting mode and then dynamically creates the appropriate button based on the current state of the user. @Alys does this work for what you were thinking?

### Loading the tools
screen shot 2018-03-01 at 8 14 39 pm

### Dynamic Buttons
screen shot 2018-03-01 at 8 15 24 pm
screen shot 2018-03-01 at 8 15 34 pm
screen shot 2018-03-01 at 8 15 42 pm
screen shot 2018-03-01 at 8 19 16 pm

Hus274:
This should be ready for review. Let me know if you have any comments!

paglias:
@Hus274 thanks! This week the team's going to be a bit busy with meetings so it might have to wait until next week, but we'll get to it :)

website/client/components/userMenu/profile.vue
```diff
@@ -866,6 +898,40 @@ export default {
...
+          chatRevoked: true,
+        };
+      }
+      this.hero.flags.chatRevoked = true;
+
+      this.$store.dispatch('hall:updateHero', { heroDetails: this.hero });
+    },
+    adminReinstateChat () {
+      if (!this.hero.flags) {
+        this.hero.flags = {
```
paglias: both here and in the previous fiunction can you just do

```
if (!this.hero.flags) {
        this.hero.flags = {}
}
```

since the value is set to true anyway in the next line

paglias (changes requested):

website/client/components/userMenu/profile.vue
Hus274: Sounds good, I'll fix this up

Hus274:
Fixed this up, should be ready for review now.

SabreCat:
On a roll, man! Thanks for all this!

paglias:
Hey @Hus274 looks like there's a small problem with this, when you open someone's profile and are an admin the upper half of the profile modal contains your info instead of the user's

schermata_2018-03-27_alle_20_09_59

Hus274:
hmm ok, let me take a look.

Hus274:
I'm not able to replicate what you've described here. I've opened the profile modal by clicking on a member in my party and on a player icon in the tavern but both times it correctly showed that persons modal. Are you accessing the profile from somewhere else?

screen shot 2018-03-28 at 6 18 30 am

paglias:
Oh, the pr isn't live yet, I was able to reproduce it on a testing site we
use internally and uses the same database as the production site. But it
doesn't seem to happen all the times

Il Mer 28 Mar 2018, 15:20 Travis <notifications@github.com> ha scritto:

SabreCat:
Will revert this on `release` for today's Staging deploy, but leave it as-is on `develop` for further testing!

Hus274:
@SabreCat makes sense, thanks!

@paglias Do you think that this change introduced this defect? Looking through the code here, I don't see how this could be causing the behavior you're describing. Is this something you can reproduce consistently? If so could you share the steps?

I haven't been able to reproduce this on the local develop instance that I'm running but I might not doing the right things to replicate the behavior.

The reason that I don't think this has something to do with this change is the top of the Profile modal simply [takes the user object as it's input](https://github.com/HabitRPG/habitica/blob/develop/website/client/components/userMenu/profile.vue#L34) and the changes in this commit don't update the user object at all. Is it possible there was a different commit that introduced this issue or is there anything you see that I'm missing here?

Sorry, I know this doesn't provide a ton of useful information. I'm having a tough time figuring out where to look without being able to reproduce the issue on my side.

paglias:
@Hus274 you're it's probably not related to this one, sorry I just got confused because I got the bug report when someone was checking what caused this. I'll investigate

Hus274:
No worries! Thanks for checking!

PR #10089, opened 2018-03-03
Quest completion modal acknowledge on close

### LINKED ISSUE #10083: quest completion modal reappears if you click the "x" button to close
When your party finishes a quest and you see the quest completion modal, if you click the "x" button on the top right corner, the screen closes but then comes back again as soon as you take another action on the site. Reloading doesn't stop it so this isn't an example of the common bug where modals reappear inappropriately. Instead, it seems that the "x" button doesn't clear your `party.quest.completed` value.

The correct behaviour is for the "x" button to do the same thing as the "ok" button at the bottom of the screen.

Hus274:
Updated the quest completion modal to consider the `x` button as acknowledgement of the quest closing. This change also prevents the user from closing the quest completion modal by clicking outside the modal or clicking the escape key.

Fixes #10083 

### Changes

----


website/client/components/achievements/questCompleted.vue
```diff
@@ -61,6 +62,9 @@ export default {
       this.$store.dispatch('user:set', {'party.quest.completed': ''});
       this.close();
     },
+    hide (evt) {
+      if (evt.trigger === 'header-close') this.setQuestCompleted();
```
paglias: why do we check for this specific trigger?

paglias (commented):
Just one question

website/client/components/achievements/questCompleted.vue
Hus274: We check for this specific trigger because as @Alys mentioned on  the issue, we don’t want to dismiss the modal on an ‘esc’ Key press or on clicking off the modal because they’d make it too easy to miss the modal entirely. This check differentiates the close from clicking the x vs the other two. Let me know if that doesn’t make sense or you’d like me to change it. 

I could also add a comment here for visibility because I’m assuming others will have the same question in the future.

paglias: I guessed that was the reason, I asked because you already disabled close on esc or backdrop in the modal props so this shouldn't be necessary

Hus274: Ahh gotcha, my bad. I’ll clean it up and resubmit.

Hus274:
Updated this pull request, ready for review.

Issue #10016, opened 2018-02-17
Copy as To-Do has wrong URLs for messages copied from Party and Tavern


Alys:
When you use Copy as To-Do on a message from your Party or the Tavern, the URLs in the To-Do are wrong.

Currently the links are:
* `[Party name goes here](https://habitica.com/groups/guild/12345678-8050-4253-1234567890abc)`
* `[Tavern](https://habitica.com/groups/guild/00000000-0000-4000-A000-000000000000)`

They should be:
* `[Party name goes here](https://habitica.com/party)`
* `[Tavern](https://habitica.com/groups/tavern)`

Feywood:
I'll pick this up if that's okay

Alys:
@Feywood Thanks! Go ahead.

Feywood:
Okay, from what I've seen so far, it seems that the copyastodomodal is used for parties, tavern and guilds and applies the same logic for each of these with the group ID. 
So that gives two options, as far as I see it:
- create a separate modal for each location you can copy from
- add in an if structure that differentiates based on the group you're copying from with guilds retaining the current structure, tavern being hardcoded with its fixed ID and parties... well I'm sure there's some form of user.party variable? 
The second seems to me the more flexible solution or at least more practical in case of future changes, but I await feedback before hopping along :rabbit:

TheHollidayInn:
I can't remember, but if we aren't already, you can just pass the full group object to the component. Then create a factory function called `generateUrl` does:
1. if group.type === guild, return guild url
2. if group.type === party, return party url
3. if group._id === TAVERN_ID (a constant you can import) return tavern url.

FergusonSean:
I'll give this a shot if that's cool with people :)

paglias:
sure @FergusonSean :)

PR #10090, opened 2018-03-03
WIP fix for copy as to do with wrong url

Feywood:
I've added in a function to return a different formatted note depending on group type being party/guild/tavern, but I'm having trouble inserting it into that notes constant.
Popping in the function call breaks the copy as to do button and I can't concatenate it with other text.
My gut is telling me I should be working async, but I don't have much of a handle on that, so any feedback is welcome.

WIP for https://github.com/HabitRPG/habitica/issues/10016

### Changes
Have created function to return correct message format

----
UUID: 9c4776eb-7330-465b-85cc-e4ccef6c1ad1


website/client/components/chat/copyAsTodoModal.vue
```diff
@@ -36,9 +36,18 @@ export default {
...
+    } else if (this.group.Type === 'guild') {
+      return `${message.user || 'system message'}${message.user ? ' wrote' : ''} in [${this.groupName}]([${this.groupName}](${baseURL}/groups/guild/${this.groupId})`;
+    } else if (this.groupId === 'tavern') {
+      return `${message.user || 'system message'}${message.user ? ' wrote' : ''} in [${this.groupName}](${baseURL}/groups/tavern)`;
+    }
+  },
   mounted () {
     this.$root.$on('habitica::copy-as-todo', message => {
-      const notes = `${message.user || 'system message'}${message.user ? ' wrote' : ''} in [${this.groupName}](${baseUrl}/groups/guild/${this.groupId})`;
+      const notes = generateURL();
```
negue: this should be `this.generateURL(message)`

website/client/components/chat/copyAsTodoModal.vue
```diff
@@ -36,9 +36,18 @@ export default {
       task: {},
     };
   },
+  generateURL () {
```
negue: you should add `message` to `generateURL` or else you won't have a message to use ^^

website/client/components/chat/copyAsTodoModal.vue
```diff
@@ -36,9 +36,18 @@ export default {
...
+    if (this.group.Type === 'party') {
+      return `${message.user || 'system message'}${message.user ? ' wrote' : ''} in [${this.groupName}](${baseURL}/party)`;
+    } else if (this.group.Type === 'guild') {
+      return `${message.user || 'system message'}${message.user ? ' wrote' : ''} in [${this.groupName}]([${this.groupName}](${baseURL}/groups/guild/${this.groupId})`;
+    } else if (this.groupId === 'tavern') {
+      return `${message.user || 'system message'}${message.user ? ' wrote' : ''} in [${this.groupName}](${baseURL}/groups/tavern)`;
```
negue: `$baseURL` needs to be `$baseUrl`

website/client/components/chat/copyAsTodoModal.vue
```diff
@@ -36,9 +36,18 @@ export default {
       task: {},
     };
   },
+  generateURL () {
+    if (this.group.Type === 'party') {
+      return `${message.user || 'system message'}${message.user ? ' wrote' : ''} in [${this.groupName}](${baseURL}/party)`;
```
negue: `${message.user || 'system message'}${message.user ? ' wrote' : ''} in [${this.groupName}]` is three times in one method, I'd suggest to extract it to a local variable and then use that variable for each of the urls

negue (changes requested):
the direction is correct, just some issues in the code. 

Also there is no need to get `async` on this change, its just a string concat

Feywood:
Thanks so much for your suggestions, it did clarify a few things for me.
However, it's not working. Either I misunderstood what you meant or it's just...
In the current state, my console in chrome tells me generateURL is not a function. Defining it as a function with function:{} did not change this, nor did adding it onto the methods (this just got me can't resolve type of undefined).
So clearly I'm still missing something somewhere.
I'm learning, I guess?

paglias:
@Feywood generateUrl is a Vue method, so you can't access it from the global scope. When you do `console.log(functionName)` it's actually doing `window.functionName` but in this case the function is registered under the vue component, not the window (global scope). To log it you can do `console.log(this.functionName)` here https://github.com/HabitRPG/habitica/pull/10090/files#diff-59ae0c35bb3ef4d17ed8cb499eaee8cfL41 inside `mounted()`, `this` means you're referring to the Vue component scope

website/client/components/chat/copyAsTodoModal.vue
```diff
@@ -36,9 +36,19 @@ export default {
       task: {},
     };
   },
+  generateURL (message) {
```
negue: this needs to be placed inside `methods: { }`, or else it won't be recognized as a vue method

lemoness:
@Feywood - Hi there! How is this fix going?

Feywood:
@lemoness I'm still blocked despite the assistance above. Will post a more substantial update about my issue in next couple days.

beffymaroo:
@Feywood How's it going? Are you still blocked? Let us know when you have a chance. Thank you!

Feywood:
Having finally opened my code again: yes, moving the method into methods has it recognised as such, but now there is a new scope issue in that the group object the logic is based on is considered undefined. 
Not knowing vue well enough, I'm not sure how to move it between the two scopes as simply adding it into the method variables does not seem to be  doing the trick. @negue if you have any pointers, I would very much welcome them.

negue:
@Feywood either you add a new `prop` like `groupType` which you could use instead of `this.group.Type` or switch from `groupName` and `groupId` to a property which contains the full group, but then you'd have to change all occurrences

Feywood:
Alright, there is actual progress this time.
Having found the existing props groupId and groupName, the code is now functional for the tavern, hurrah hurrah.
However (as always), not yet for party/guild. There the issue is that of course groupNames are entirely up to users so I can't use that variable, and groupId is changable as well, so not so much either. I cannot seem to derive a group type from ID or name, so no luck.
The other option seems to be generating a prop myself with the grouptype, but while I can see the actual prop in my dev console, I cannot seem to get it to populate and it remains 'undefined'. I'm guessing this info has to come from chatMessages.vue or group.vue but how?

negue:
In your recent commit you removed the support for party & group messages, this was a step backwards^^

the issue was, like I wrote in my prior message, you either needed a new `prop`erty for `groupType` or pass the full `group` , in the code the `this.group` was undefined

For debugging steps is always helpful just to use `console.info`, just put all variables in it and then you'll see quite fast which one is `undefined`

beffymaroo:
@Feywood How's this fix going? Let us know if you're still working on it when you get a chance. Thank you!

PR #10107, opened 2018-03-07
Update Staff list in the tavern to make staff names clickable to pull up staff profiles.

### LINKED ISSUE #9290: Mod/Staff Names in Tavern should be Clickable
The Tavern has a list of all the staff and mods in the side bar. You should be able to click each name of a mod/staff member to bring up their user profile. This is important because it ensures that users have a quick way to get in touch with staff and mods.

Hus274:
This change updates the staff list in the tavern to make the staff names clickable. When the name of a staff member in this list is clicked, it will bring up that staff members profile. This will ensure habitica users have a quick and easy way to reach out to moderators and admins when needed.

fixes #9290

Please let me know if you have any feedback.


website/client/components/groups/tavern.vue
```diff
@@ -764,6 +782,22 @@ export default {
         this.$root.$emit('bv::show::modal', 'world-boss-rage');
       }
     },
+    async viewStaffProfile (staffId) {
+      const result = await this.$store.dispatch('members:fetchMember', { memberId: staffId });
```
paglias: This looks good but I think the code an be simplified, using something like this

```
    async clickMember (hero) {
      let heroDetails = await this.$store.dispatch('members:fetchMember', { memberId: hero._id });
      this.$root.$emit('habitica:show-profile', {
        user: heroDetails.data.data,
        startingPage: 'profile',
      });
    },
```

handling errors is not necessary as we have interceptors that manage them in case something doesn't work

paglias (changes requested):

Hus274:
Comments make sense to me, I've modeled the method after what you've listed here. I did want to note though, one of the reasons I did do the error handling was because during testing I noticed that if the profile couldn't be found, nothing was shown to the user and it simply caused a console error without giving the user any indication that happened which I didn't think was great. If this is the expected behavior, that's totally fine, this edge case shouldn't really happen.

Should be ready for review again.

paglias:
I see @Hus274 , but let's keep the error handling outside of this PR anyway, the right place to have it is in the interceptor in app.vue

paglias (approved):

paglias:
Thanks! Will merge in the next few days :)

Hus274:
Makes perfect sense to me, thanks!

paglias:
Welcome to tier 7 @Hus274 !

PR #10114, opened 2018-03-08
fix: Fixing xml data export 500 error.

### LINKED ISSUE #10100: Data Export XML option does not work  (export/userdata.xml file not found)
In the [Data Export](https://habitica.com/user/settings/data-export) screen, the "Export User Data (XML)" option can't be used. You see this message (wording depends on browser):

`can't find the file at https://habitica.com/export/userdata.xml

The JSON export option does work.

_Thanks to Cyrus Krieger (a4626bcc-2bff-4c39-a1ce-cf015ee1ba50) for reporting this in the Report a Bug guild._

Hus274:
The xml export api call was intermittently throwing a 500 failure due to trying to interpret user.pinnedItems and user.unpinnedItems `id` fields as hexadecimal values and converting them into invalid unicode xml characters. This change removes the automatically added and unnecessary `id` field from these user parameters and reinstates the xml export test.

Additional information about the diagnosis of this issue can be found [here](#10100).

fixes #10100


paglias:
I'm not sure but this might need a migration for users with already _id(s) stored

test/api/v3/integration/dataexport/GET-export_userdata.xml.test.js
```diff
@@ -8,7 +8,7 @@ let parseStringAsync = Bluebird.promisify(xml2js.parseString, {context: xml2js})
 
 describe('GET /export/userdata.xml', () => {
   // TODO disabled because it randomly causes the build to fail
```
paglias: can you remove this comment?

Hus274: Yep, missed this in my first pass. Removed.

Hus274:
I've removed the outdated comment. The xml test failed the first time because the id of the user item didn't match what was returned by the xml output. I'm unable to replicate this manually but I'm concerned about introducing more tests that fail intermittently to the test base. If someone can restart the tests, if the xml fails again, I can investigate it.

@paglias I think leaving the _id on the pinned items for the user won't cause any change and will slowly be removed as each user's array is updated. That being said, if you think it's best to do the migration to remove this id just to keep everything consistent, that seems reasonable as well.

Hus274:
Fixed the tests, they should pass consistently now.

paglias:
@Hus274 thanks a lot for all this PRs! Welcome to tier 7

Hus274:
Awesome, thank you!!

PR #10120, opened 2018-03-10
 fix: fix quest shop to not use string addition when buying quests

### LINKED ISSUE #10115: Quest Shop arithmetic follows its own rules. Laws of Mathematics are not the Habitica way.
Bulk purchasing doesn't work correctly in quest shop. Number purchased is treated as integer for 1 and as string for 2 or more (i.e., string concatenation occurs when you buy the items, instead of integer arithmetic). The number is also treated as a string if you change the number to anything and then change it back to 1.

To reproduce:

- Give yourself a chunk of gold.
- Go to the quest shop. Pick a gold-purchasable quest that you don't have any of. 
- Buy 1. You now own 1.
- Buy another 1. You now own 2. This is good.
- Use the bulk purchase feature to buy 2 at once. You now own 22. This is wrong.
- Buy 10 at once. You now own 2210. This is getting silly.
- Pick another quest that you own none of.
- Buy 1000. You now own 1000.
- Buy another 1000. You now own 10001000. This is not how maths is meant to work.

Hus274:
I've fixed the quest purchasing quantities being interpretting as strings and therefore doing string addition. At the beginning of my research, I found that the view controller was interpretting the user input as a string causing the error. I followed the documentation on vue's site to correctly interpret the user input as a number: https://vuejs.org/v2/guide/forms.html#number

I was about to submit a pull request when I thought through what happens if people are using this API incorrectly. I then also updated the server side code to interpret the user input as a string to make sure that this doesn't happen to any applications that use the API directly. I've also added a regression test for this case.

Let me know if you have any feedback.

fixes #10115


website/common/script/ops/buy/buyQuest.js
```diff
@@ -10,7 +10,7 @@ import get from 'lodash/get';
 // buy a quest with gold
 module.exports = function buyQuest (user, req = {}, analytics) {
   let key = get(req, 'params.key');
-  let quantity = req.quantity || 1;
+  let quantity = Number(req.quantity) || 1;
```
paglias: this works but I feel like we should also throw an error if the param couldn't be converted to a number correctly, so add another check after this like `Number.isInteger(quantity)`and if it's false, throw an error

paglias (changes requested):
Left one comment, also I think the same issue applies to other functions in /common. Would you mind adding the same check there? Thanks

Hus274:
I've updated the pull request to address the comments you've left here. Let me know if you have any additional comments!

website/common/script/ops/buy/buyQuest.js
```diff
@@ -10,7 +10,9 @@ import get from 'lodash/get';
 // buy a quest with gold
 module.exports = function buyQuest (user, req = {}, analytics) {
   let key = get(req, 'params.key');
-  let quantity = req.quantity || 1;
+
+  if (req.quantity && isNaN(req.quantity)) throw new BadRequest(i18n.t('invalidQuantity', req.language));
+  let quantity = req.quantity ? Number(req.quantity) : 1;
```
paglias: both here and in the other file, can you convert to Number before checking for isNan?

paglias (changes requested):

website/common/script/ops/buy/buyQuest.js
Hus274: IsNan from my understanding works correctly for strings that just have numbers: https://www.w3schools.com/jsref/jsref_isnan.asp

My concern with casting first is that the cast will fail and throw an error before the isNan check. I might be misunderstanding what you’re asking for though.

paglias: I see, but don't worry, casting an invalid value won't throw any error but just convert the input to a NaN value which will then get caught by the isNaN check

Hus274:
Updated the check here, let me know if there's anything else! Ready for review.

paglias (approved):

PR #9385, opened 2017-11-03
No Ethereal Surge on Mages (fixes #8860) - continued in PR #10121

LunarGem:
There might be a leftover comment or two from another fix I'm doing, but look it over and see if I'm doing this right...
 
fixes #8860

PR #10121, opened 2018-03-10
No Ethereal Surge on Mages

### LINKED ISSUE #8860: Make Ethereal Surge work only on non-Mages in the party

### Description

Intrepid gamebreakers have noticed it's possible to use multiple Mages in a party to cast skills indefinitely without checking off any tasks or crossing from one day to the next. They can give the rest of the party enough Mana to cast their skills without limits, increase INTelligence to stratospheric limits, etc. All it takes is Mages using Ethereal Surge to give one another Mana. The Surge doesn't raise one's own Mana, but it does increase the Mana of other Mages who can then use Ethereal Surge to give back the Mana you spent, and so on!

Normally we don't care overmuch when people discover "cheats" in Habitica, as the core concept of checking off real-life tasks to earn in-game rewards is something that only works on an honor system to begin with. As folks in the Tavern typically put it, "the only person you're cheating is yourself." For this exploit, however, it makes it a bit too simple to invalidate the rest of Habitica's game elements: damage never matters if you can cast unbounded Blessings, you never have to check off a task to attack a monster if you can Brutal Smash and Burst of Flames indefinitely, and so on. We've even ended up with server problems due to people setting up automated scripts to chain-cast Ethereal Surge.

One simple tweak should make infinite Ethereal Surge loops impossible: **make Ethereal Surge affect only non-Mages**. You can give up your Mana to help party members of other classes use their own skills, but you can't use Ethereal Surge to create a perpetual motion machine among a group of Mages.

Yutsuten:

Fixes #8860

### Changes
This is a continuation of the Pull Request #9385.
The feature itself was already done by @LunarGem, so I'm just completing it.

#### TODO:
- [x] When there are other mages in the party, show notification telling that the Ethereal Surge does not affect other mages
- [x] Create automated test

### Notification margin fix
I had a problem with the new notification text being too big, and the text was getting out of the div.
So I added some margins to left and right to fix it.
Still not perfect, but better than before.

#### Before
before
#### Now
now

----


Yutsuten:
The unit and integration tests failed for some random reason... Can someone restart it please?

Alys:
restarted

Yutsuten:
Thank you! Unit tests are now passing, but 3 integration tests failed by timeout.
I'm afraid that another restart will be needed... Sorry.

Alys:
No worries, restarted. :)

Yutsuten:
Now its 1 integration test failing by timeout.
Should we ignore it?

Yutsuten:
Nice, now it is passing.
Thank you for the patience! >.<

test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
```diff
@@ -187,6 +187,30 @@ describe('POST /user/class/cast/:spellId', () => {
     expect(group.chat[0].uuid).to.equal('system');
   });
 
+  it('Etheal Surge does not recover mp of other mages', async () => {
```
khipkin: typo: Ethereal

Yutsuten: Thank you for noticing it! I'll fix it right now.

website/client/mixins/spells.js
```diff
@@ -129,6 +129,28 @@ export default {
 
 
       this.markdown(msg); // @TODO: mardown directive?
+
+      // If using mpheal and there are other mages in the party, show extra notification
+      if (type === 'party' && spell.key === 'mpheal') {
+        // Get party
+        let members = await this.$store.dispatch('members:getGroupMembers', {
```
paglias: why are we loading this data? isn't it already loaded since it's the user's own party?

test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
```diff
@@ -187,6 +187,30 @@ describe('POST /user/class/cast/:spellId', () => {
     expect(group.chat[0].uuid).to.equal('system');
   });
 
+  it('Ethereal Surge does not recover mp of other mages', async () => {
+    let group = await createAndPopulateGroup({
+      groupDetails: { type: 'party', privacy: 'private' },
+      members: 4,
+    });
+    await group.groupLeader.update({'stats.mp': 200, 'stats.class': 'wizard', 'stats.lvl': 20});
```
paglias: this 5 updates can be made in parallel with `Promise.all(arrayofUpdates)` which will help with speed, same for the sync calls below. Is the sleep necessary?

website/common/script/content/spells.js
```diff
@@ -62,7 +62,7 @@ spells.wizard = {
     cast (user, target) {
       each(target, (member) => {
         let bonus = statsComputed(user).int;
-        if (user._id !== member._id) {
+        if (user._id !== member._id && member.stats.class !== 'wizard') {
```
paglias, line 65: I guess we don't want to throw an error, but this way we're just ignoring the mages and on mobile there won't be any notification about this. What do you think @Alys ?

paglias (changes requested):

website/client/mixins/spells.js
Yutsuten: Is it? Can you tell me the variable name?
Sorry, I really looked for it, but didn't find.

website/common/script/content/spells.js
Yutsuten: There is any way to send notifications to mobile?

test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
Yutsuten: I saw sleep on a lot of tests in that script, so I considered it was necessary.
Sure I can remove it.

website/client/mixins/spells.js
Yutsuten: Now I understand why I wasn't finding this variable. There was a mistake on the code that passes the party members to the `castEnd` method.
`this.$store.state.partyMembers` returns an object, but after it there was a check if it was an array.
```javascript
let party = this.$store.state.partyMembers;
party = isArray(party) ? party : [];
party = party.concat(this.user);
this.castEnd(party, spell.target);
```
At the end, `party` was always an array with `this.user`.

I fixed it changing this code to:
```javascript
let party = this.$store.state.partyMembers.data;
if (!isArray(party)) {
  party = [this.user];
}
this.castEnd(party, spell.target);
```

website/common/script/content/spells.js
paglias: not really, I mean yes but without proper handling it would be displayed.

@phillipthelen @TheHollidayInn  what do you think?

website/client/mixins/spells.js
```diff
@@ -32,9 +32,10 @@ export default {
           return this.castEnd(party, spell.target);
         }
 
-        let party = this.$store.state.partyMembers;
-        party = isArray(party) ? party : [];
-        party = party.concat(this.user);
+        let party = this.$store.state.partyMembers.data;
```
paglias: can we rename this variable to partyMembers, it's quite confusing

paglias (changes requested):
Left one small comment but this looks good! I'm just worried about the fact that mobile won't show anything so users won't know why their spell didn't affect some users. @Alys

website/client/mixins/spells.js
Yutsuten: Sure

Alys:
I agree with paglias's concern about mobile users not being told about this. My suggestion is that the skill's description could be changed from this:
"You sacrifice mana so the rest of your Party gains MP! (Based on: INT)"
to something like this:
"You sacrifice Mana so the rest of your Party gains MP! Does not give MP to other Mages. (Based on: INT)"
Note that Mana should be title case not lower case. That wording can probably be improved.

That makes it clear to all users how the skill will work before they cast the skill (even those using third-party clients if the client is already configured to show the skill description).

I'm a bit concerned about how users will interpret this message:
"Your Ethereal Surge backfires when mixed with another's magic. Only the non-magic users gain mana."
It's a great message for gamification but we've had trouble in the past with users taking gamified messages too literally, and that one sounds a bit like an error notification. I'd expect it to cause questions like "How can the Mages in our party ensure that their skills don't conflict?" or bug reports like "There's only one other Mage in my party and they are certain they didn't cast any skills today so why did my skill fail for them?"

Yutsuten:
@Alys This suggestion is nice, changing the description of the skill allows everyone notice that the skill does not work for other mages before using it. Maybe we can improve it to something like:

`You sacrifice Mana so the rest of your Party, except Mages, gains MP! (Based on: INT)`

Having "Does not give MP to other Mages." in a separated sentence somehow worries me. But that's something that should be decided with more opinions.

About the extra notification, how about if we change to something like: "*Your Ethereal Surge backfires when mixed with another's magic. Only non Mages gains MP*". This way we have a nice gamification description and a direct explanation of what it affected.

Reading both skill description and the extra notification, it should become very obvious that other mages don't receive MP by that skill, no matter what.

1. `You sacrifice Mana so the rest of your Party, except Mages, gains MP! (Based on: INT)`
2. `Your Ethereal Surge backfires when mixed with another's magic. Only non Mages gains MP.`

Alys:
"You sacrifice Mana so the rest of your Party, except Mages, gains MP! (Based on: INT)" - Yes, much better than my version!

"Your Ethereal Surge backfires when mixed with another's magic. Only non-Mages gain MP." - I like that! I made a couple of minor changes there (a hyphen after non which I _think_ is correct usage and gain rather than gains).

Yutsuten:
@paglias What do you think about the new strings for the Ethereal Surge's description and notification?

1. `You sacrifice Mana so the rest of your Party, except Mages, gains MP! (Based on: INT)`
2. `Your Ethereal Surge backfires when mixed with another's magic. Only non-Mages gain MP.`

paglias:
Works for me!

Yutsuten:
Long string
`Your Ethereal Surge backfires when mixed with another's magic. Only non-Mages gain MP.`

This string was a little too long, so I changed to:

Short string
`Your spell backfires when mixed with another's magic. Only non-Mages gain MP.`

Is it fine?

Alys:
@Yutsuten Shortening it is good, thank you! Can you please use "Skill" rather than "spell" though? "Skill" is the term that should be visible to the players; "spell" is just used in the code - a holdover from long ago when they were referred to as spells publicly.

Yutsuten:
Done!

paglias:
Not sure this is a bug but was just testing this and noticed that if I create a party with me being a mage and no other members and cast Ethereal Surge I don't get any blue snackbar messages. Is it the wanted behavior or a bug?

Yutsuten:
The message is only shown if there are any other mages in the party. Do you think this shouldn't be the behaviour?

paglias:
It makes sense, merging now!

Yutsuten:
Nice!

SabreCat:
Thanks for this! Noted toward your next contributor tier!

PR #10122, opened 2018-03-10
fix: you are already in group message - fixes #10119

### LINKED ISSUE #10119: userAlreadyInGroup should not be used when a user joins a group
The `userAlreadyInGroup` error message is designed for use when a group member tries to invite another player, as shown in the two code snippets below. That works well and should not be changed.
https://github.com/HabitRPG/habitica/blob/d5e4be85e991740e6ec622c41f7c38a307b802ea/website/common/locales/en/groups.json#L225
https://github.com/HabitRPG/habitica/blob/d5e4be85e991740e6ec622c41f7c38a307b802ea/website/server/controllers/api-v3/groups.js#L938

However the message is also being used when a player tries to join a guild that they are already a member of. They see an error message because the correct parameters aren't being passed to that locales string. `Error processing the string "userAlreadyInGroup"`
https://github.com/HabitRPG/habitica/blob/d5e4be85e991740e6ec622c41f7c38a307b802ea/website/server/controllers/api-v3/groups.js#L558

The fix is to create a new string for this case, called something like `youAreAlreadyInGroup`, with the text "You are already a member of this Guild". Please add that string to groups.json immediately below the `userAlreadyInGroup` to keep the two similar messages together.

You'll then need to adjust the code that throws the error when the user tries to join the guild:
https://github.com/HabitRPG/habitica/blob/d5e4be85e991740e6ec622c41f7c38a307b802ea/website/server/controllers/api-v3/groups.js#L557-L559

And you'll need to adjust the associated test to make it use the new string:
https://github.com/HabitRPG/habitica/blob/d5e4be85e991740e6ec622c41f7c38a307b802ea/test/api/v3/integration/groups/POST-groups_groupId_join.test.js#L47-L54

tmkuba:

[//]: # 10119
Fixes https://github.com/HabitRPG/habitica/issues/10119

### Changes
[//]: # display `youAreAlreadyInGroup` message when user attempts to join an already-joined group.

[//]: # b1cf5a63-61c5-46e5-807d-656415fde46b

----
UUID:

### DIFF
--- a/test/api/v3/integration/groups/POST-groups_groupId_join.test.js
+++ b/test/api/v3/integration/groups/POST-groups_groupId_join.test.js
@@ -44,12 +44,12 @@ describe('POST /group/:groupId/join', () => {
       expect(res.leader.profile.name).to.eql(user.profile.name);
     });
 
-    it('returns an error is user was already a member', async () => {
+    it('returns an error if user was already a member', async () => {
       await joiningUser.post(`/groups/${publicGuild._id}/join`);
       await expect(joiningUser.post(`/groups/${publicGuild._id}/join`)).to.eventually.be.rejected.and.eql({
         code: 401,
         error: 'NotAuthorized',
-        message: t('userAlreadyInGroup'),
+        message: t('youAreAlreadyInGroup'),
       });
     });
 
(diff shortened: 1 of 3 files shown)


paglias (commented):
This looks good, I'm only wondering if this change should also cover parties. cc @Alys

Alys:
I don't think it needs to mention parties. There's  a comment above that
code that says parties have already been handled.

paglias:
@tmkuba  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #10126, opened 2018-03-11
Fix/mana bar unselected class - fix: #10026

### LINKED ISSUE #10026: mana bar should not show in profile when user has not yet selected a class
When a user has not yet selected a class, currently if you view their profile you see a mana bar. The correct behaviour is that no mana bar should be shown (the same as for a user who has opted out of the class system).

Below is data showing the problem. `flags.classSelected` is `false` when a player has not yet selected a class so that is what you can use to prevent the mana bar being displayed.
```
{
    "stats": {
        "class": "warrior",
        "lvl": 13
    },
    "flags": {
        "classSelected": false
    },
}
```
screenshot_2018-02-20_09-26-24

tmkuba:

Fixes #10026

### Changes
[//]: # updates members.hasClass() to account for users who have not selected a class

[//]: # 
----
UUID: b1cf5a63-61c5-46e5-807d-656415fde46b

### DIFF
--- a/test/client/unit/specs/store/getters/members/hasClass.js
+++ b/test/client/unit/specs/store/getters/members/hasClass.js
@@ -13,7 +13,7 @@ describe('hasClass getter', () => {
         classSelected: true,
       },
     };
-    expect(hasClass(member)).to.equal(false);
+    expect(hasClass()(member)).to.equal(false);
   });
 
   it('returns false if member has disabled classes', () => {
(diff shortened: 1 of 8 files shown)


paglias:
This works @tmkuba  but I noticed that the same check is used in several places in the code and it would be great if we could use the getter everywhere, for example https://github.com/HabitRPG/habitica/blob/4108a22d7822027c29a1b5ed9843cc3844caae63/website/client/components/achievements/levelUp.vue#L152 (but there are other places too)

tmkuba:
Just an update, I am currently investigating the other places in the code that refers to the referenced flags

tmkuba:
Hi @paglias , I have updated the PR and used the hasClass() getter where similar checks are made.  Please review when you have a chance.

SabreCat:
Thanks for this, @tmkuba! Noted toward your second contributor tier!

PR #10127, opened 2018-03-11
Fix npm version on Dockerfile to 5.6.0

jfcorsini:
npm@5.7.1 is having some issue during extraction of large projects (see https://github.com/npm/npm/issues/19989). So while this is not fixed, new contributors were not able to start a new local project, because the containers were updating npm to the newest version. In order to fix this issue I fixed the version to 5.6.0, which is more stable and it works on Habitica project.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM node:boron
 
 # Upgrade NPM to v5 (Yarn is needed because of this bug https://github.com/npm/npm/issues/16807)
 # The used solution is suggested here https://github.com/npm/npm/issues/16807#issuecomment-313591975
-RUN yarn global add npm@5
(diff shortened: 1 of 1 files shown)


jfcorsini:
And for some reason testing are timing out as well.. is there a way that I can restart them?

paglias:
I'll restart them

paglias:
@jfcorsini thanks for this and sorry for the delay. I'm going to close this in favor of https://github.com/HabitRPG/habitica/commit/0f339d8d3e0b5a234793ce38e77f174cd8a68de2, this way we're going to install the NPM version shipping with the latest node 8, 8.10 at this time which already bundles v5.6.0 of npm

PR #10110, opened 2018-03-07
Made kafka optional

TheHollidayInn:

PR #10129, opened 2018-03-12
Make Kafka dep optional

paglias:
@TheHollidayInn @Alys this is based off #10110 but it also makes kafka an optional dep and only load it where it's supported.

@Alys can you try if it works for you?


Alys:
Yes, this is good, thank you! My local install works when this is merged in.

paglias:
Will merge this as soon as the build passes

PR #10142, opened 2018-03-16
Fixes asynchronous cron bug that allows cron to run twice for a user within seconds.

### LINKED ISSUE #8991: cron happening twice at almost the same time (e.g., double quest progress / boss damage)
In the Report a Bug guild, Birkwelch (310aea01-13f5-4fb6-929e-be39d63e665b) reported that cron ran twice:
"Just a few minutes ago, I seem to have cronned twice in the same instant, according to my party log? I did identical damage, but the first time received (correctly) no damage, having completed my dailies, but the second time the system acted as if I had completed nothing, killing me and mortally wounding most of my party."

I've confirmed this from the quest logs (below) and Amplitude, which shows two cron events for the same timestamps as below.

```
        {
            "id": "cfd796f8-e9e4-43a8-a5d6-34585d288b9d",
            "text": "`Birkwelch attacks The Hippo-Crite for 197.5 damage.` `The Hippo-Crite attacks party for 46.6 damage.`",
            "timestamp": 1503684094495, //      UTC: 2017-08-25 18:01:34
            "flagCount": 0
        },
        {
            "id": "50640ade-5e5d-403c-bf7f-5f07a6889e87",
            "text": "`Birkwelch attacks The Hippo-Crite for 197.5 damage.` `The Hippo-Crite attacks party for 0.0 damage.`",
            "timestamp": 1503684089368, //      UTC: 2017-08-25 18:01:29
            "flagCount": 0
        },
```

Hus274:
There is an asynchronous path that can cause two crons for the same user to run within seconds. The following is a description of how this can happen in the current code path.

Let's say the user's client triggers two calls to the cron api. The requests each go to a different server, server1 and server2. The following list is an ordered operation of events

1. Server1 [gets the user object from the database](https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/cron.js#L47) stores it locally and it looks like the following:
```
{
   _cronSignature: NOT_RUNNING
   lastCron: YESTERDAY
}
```

2. Server2 [gets the user object from the database](https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/cron.js#L47) stores it locally and it looks like the following:
```
{
   _cronSignature: NOT_RUNNING
   lastCron: YESTERDAY
}
```

3. Server1 moves to the [checkForActiveCron method](https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/cron.js#L10)

4. Server1 updates the user object with a mongo statement that validates the _cronSignature isn't running or is ready for a retry.
MongoDatabase before the update:
```
{
   _cronSignature: NOT_RUNNING
   lastCron: YESTERDAY
}
```
After the update:
```
{
   _cronSignature: NOW
   lastCron: NOW
}
```

5. Server1 still has the original User object in memory:
```
{
   _cronSignature: NOT_RUNNING
   lastCron: YESTERDAY
}
```
which allows it to pass the lastCron check [here](https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/cron.js#L58)

6. Server1 proceeds to [execute the rest of the cron code updating the user and the group](https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/cron.js#L64-L103)

7. Server1 [releases it's running cron lock on the user object](https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/cron.js#L38-L44)
Mongo Actual before the update:
```
{
   _cronSignature: NOW
   lastCron: NOW
}
```
Mongo Actual After the update:
```
{
   _cronSignature: NOT_RUNNING
   lastCron: NOW
}
```

8. Server2 moves to the [checkForActiveCron method](https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/cron.js#L10)

9. Server2 updates the user object with a mongo statement that validates the _cronSignature isn't running or is ready for a retry.
MongoDatabase ACTUAL before the update. Remember, Sever1 has already released the lock of _cronSignature.
```
{
   _cronSignature: NOT_RUNNING
   lastCron: NOW - a small period of time
}
```
After the update:
```
{
   _cronSignature: NOW
   lastCron: NOW
}
```

10. Server2 also still has the original User object in memory:
```
{
   _cronSignature: NOT_RUNNING
   lastCron: YESTERDAY
}
```
which allows it to pass the lastCron check [here](https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/cron.js#L58)

11. Server2 proceeds to [execute the rest of the cron code updating the user and the group](https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/cron.js#L64-L103) in addition to the processing that was already handled by Server1

fixes #8991 

### Changes
The primary issue that I've described above is that the lastUpdatedTime check is dependent on the time of the lastCron which was gathered before the cron lock is acquired. This change acquires the lock for the cron (no one can change the lastCron time), then looks at the lastCron time and determines if it needs a cron. 

Currently what's happening is the server asks when the last cron happened then before it executes, a different server finishes a cron and this server still doesn't see that the lastCron has been updated.


paglias:
Thanks for the detailed explanation @Hus274 this totally makes sense.

I'm wondering if instead of reloading the user from database we can simply set lastCron to now to the in memory copy we have.

Also upgrading the user variable isn't going to update res.locales.user so they'll differ, it should be

```
user = res.locals.user = new version
```

Hus274:
Your second comment makes sense, I can go ahead and make that change. 

The first one I don’t think is possible though. The reason we’re reloading the user is to see if a different server has modified the lastCron field on the object between when the request hit the server and when the lock was acquired for the user. I could make the load hit smaller by only pulling in the lastCron instead of the whole user document but it’s just a smaller load. Setting the local lastCron doesn’t have a check determining whether another server has run Cron between when the request came in and the lock was acquired.

Hus274:
Now that I think about it a bit more, we could add another conditional check when we try to set Cron to running to fail if the lastCron is within the last hour or something like that which would prevent this issue.

Alys:
There are times when cron should run twice in an hour, for example if you
use Habitica for the first time one day at 23:30, then mark your Dailies
complete before midnight, then cron again (correctly) just after midnight.
We don't want to break that use case so I'm thinking the extra database
transactions are worth it.

Hus274:
@paglias I've made the change to ensure that user and res.locals.user stay in sync. Let me know if you have any additional thoughts on the database load.

This should be ready for review again.

TheHollidayInn:
I've found this logic as well, but haven't been able to replicate with automated tests. I was able to replicate many other recurrent routes.

So, the theory is sound, but have you been able to write a test to recreate this?

Hus274:
@TheHollidayInn I was able to write a test that replicates this bug but it did require modifying the source code with a thread.sleep type behavior to force the paths to be executed in the correct order which is why I didn't push it. I couldn't think of a better way to simulate a Garbage Collection or some other stop the world behavior on box. 

I've pushed the test I wrote (I apologize that it's gross, I was just trying to get it working at the time). You can see the commit here: https://github.com/Hus274/habitica/commit/07145c1f044870410b1e72389983e6660cbd1541

TheHollidayInn:
Ah perfect! Let me read through that then. Thanks for posting it.

Hus274:
Sure thing, let me know if you have any questions!

beffymaroo:
@TheHollidayInn Is this still in need of review? Just checking in since it's marked Important and it's been a minute since the last update.

TheHollidayInn (approved):

PR #10146, opened 2018-03-17
update package-lock.json

Alys:
This is the package-lock.json file that was automatically regenerated after I deleted it and node_modules and then did an `npm install`

Feel free to close this if you don't want this version of the file committed. I'm guessing this new version will start appearing in PRs from contributors but there might be a reason why it hasn't been committed to `develop`.


paglias:
I'm going to close this, sometimes different systems produce different package locks based on which deps are necessary

PR #10147, opened 2018-03-17
increase Hourglasses and gemCapExtra promptly when multi-month subscription renews - fixes #4819

### LINKED ISSUE #4819: Subscription renewal hourglasses

Alys:
Fixes #4819

The bug in #4819 is that when a multi-month recurring subscription renews, the user does not immediately get the new Hourglasses and increased gem cap that they've paid for. Instead, as each month after that passes, they build up credit towards one new Hourglass in the same way that a one-month recurring subscription does. This is unfair since they've already paid for the full multi-month period.

This PR fixes that, mostly. We can't increment the Hourglasses and extra gem cap at the exact date that the subscription renews because we don't track payments to the extent needed for that. 
Instead, this PR waits until the start of the month immediately after the subscription has renewed and then gives the extra Hourglasses and gem cap.

It's not a perfect solution because there's already a lot of subscriptions that have completed their first period and are part way through a second/third/etc period. For example, consider a 6-month subscription that's four months into its second year. When that user crons at the start of the next month after this PR goes live, they'll be given two new Hourglasses and increased gem cap, even though they actually got one of those new Hourglasses already at the end of the previous three-month period - i.e., we're giving them slightly more than they should have. However many of the subscribers in this situation will have suffered through months of not having the correct benefits in advance, so this balances the scales a bit. In the case of that example, their increased subscription benefits will continue to be offset by four months indefinitely, which isn't ideal but is still to their benefit and if anyone reports the problem we can manually adjust their data easily enough (just as we do for reports that are happening now). I'm considering modifying the code to take into account the current `count` and to adjust the new benefits and offset to match it, but I don't want to make the code too complex and risk introducing new bugs.


website/server/libs/cron.js
```diff
@@ -70,16 +71,49 @@ function grantEndOfTheMonthPerks (user, now) {
...
       plan.consecutive.count++;
 
-      if (plan.consecutive.offset > 1) {
-        plan.consecutive.offset--;
-      } else if (plan.consecutive.count % 3 === 0) { // every 3 months
-        if (plan.consecutive.offset === 1) plan.consecutive.offset--;
-        plan.consecutive.trinkets++;
-        plan.consecutive.gemCapExtra += 5;
-        if (plan.consecutive.gemCapExtra > 25) plan.consecutive.gemCapExtra = 25; // cap it at 50 (hard 25 limit + extra 25)
+      plan.consecutive.offset--;
```
paglias, line 79: why was the check for this to be > 1 removed?

Alys: With the new code, the offset is always decremented, and when it's reached -1, that's a sign that we need to look into special handling for multi-month subscriptions that renewed in the previous month. The comments below this line go into some detail about that but I'm happy to make it clearer if needed!

paglias:
Is this ready for review @Alys ?

Alys:
No, I'm still adding tests. I'll comment and change the labels when I'm
done.

Alys:
The code and tests are finished and this is ready for review.

test/api/v3/unit/libs/cron.test.js
```diff
@@ -143,15 +144,16 @@ describe('cron', () => {
...
       user.purchased.plan.consecutive.count = 5;
-      user.purchased.plan.consecutive.offset = 1;
       cron({user, tasksByType, daysMissed, analytics});
       expect(user.purchased.plan.consecutive.gemCapExtra).to.equal(5);
-      expect(user.purchased.plan.consecutive.offset).to.equal(0);
+      expect(user.purchased.plan.consecutive.count).to.equal(6);
     });
 
     it('increments plan.consecutive.gemCapExtra multiple times if user has been absent with continuous subscription', () => {
+      // TODO if approved, delete this because the new tests below include this situation
```
paglias: what about these comments @Alys ?

paglias (approved):
This is looking good, just left one comment

test/api/v3/unit/libs/cron.test.js
Alys: It's up to you. If you're happy with that the new tests cover what these old tests cover, and if you don't want to keep the old tests because they're related to other tests around them, then I can remove them now.

paglias: ah ok, let's delete them then!

Alys: I've deleted those four tests.

PR #10150, opened 2018-03-18
Added paging to Challenge lists 

TheHollidayInn:


website/server/controllers/api-v3/challenges.js
```diff
@@ -341,28 +341,61 @@ api.getUserChallenges = {
...
+    }
+
+    if (owned && owned === 'owned') {
+      query.$and.push({leader: user._id});
+    }
+
+    if (req.query.search) {
+      const searchOr = {$or: []};
+      const searchWords = req.query.search.split(' ').join('|');
+      const searchQuery = { $regex: new RegExp(`${searchWords}`, 'i') };
```
paglias: Should we escape the search string here? To avoid the problem with had when searching for stuff like c++ in the guilds page? There's a lodash utility that does it

TheHollidayInn: Sure!

paglias: this should work https://github.com/HabitRPG/habitica/pull/10092/files#diff-b950a7c557b35ea58bc2685532c46170R340

Issue #10011, opened 2018-02-16
Purchase API Refactoring - Overview


negue:
Goal:
- extract / abstract every purchasable operation into classes
- use a factory to get the needed class
- each operation has its own checks but also the inner checks for each currency-type
- checks to prevent bulk purchases on types that can't be purchased in bulk

Todo: 
- [x] Armoire [Gold] (#10153)
- [x] Market Gear [Gold] (#10010)
- [x] Health Potion [Gold] (#10152)
- [x] Quest [Gold] (#10244)
- [x] Buy Gem [Gold] (#10271)
- [x] Special Spell [Gold] (#10305)
- [x] Mount [Hourglass] (#10482)
- [ ] Pets [Hourglass]
- [x] Quest [Gem] (#10476)
- [ ] Mystery Set [Hourglass]
- [ ] Spell [Gem]
- [ ] Eggs [Gem]
- [ ] Hatching Potions [Gem]
- [ ] Food [Gem]
- [ ] Gear [Gem]
- [ ] Bundles [Gem]
- [ ] create / use factory
- [ ] fix / add new **TESTS!**

seekheart:
@negue maybe time to move to typescript? It supports interfaces which may help with this

negue:
Typescript would be nice yes, but we'd have a lot work before having a working server just with the typescript compiler instead of babel (atleast thats what I think)


Currently the classes are in ES6 so it would be already working with typescript

seekheart:
It's not a whole lot more work though. Mostly we could put all the `client` code in to TS and just have typescript transpile it or gulp it?

paglias:
I personally like Typescript but for now I'd say let's not introduce new tools :)

TheHollidayInn:
Agreed. Typescript is something we can consider after we remove most of the code "smells"

seekheart:
Are we following an OOP paradigm currently?

paglias:
Is this still in progress?

negue:
Not currently, no

beffymaroo:
@negue Should I remove the in progress label from this issue, possibly add Help Wanted?

PR #10152, opened 2018-03-19
Purchase API Refactoring: Health Potion

negue:
Part of #10011

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -151,7 +151,7 @@
     },
     "@sinonjs/formatio": {
       "version": "2.0.0",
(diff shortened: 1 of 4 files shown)


paglias:
schermata 2018-03-23 alle 17 05 19

There are a lot of broken tests, I think it's due to using an es6 export while before we had an old type of export (which is made using exports.defaults). I don't remember exactly but there were some problems using es6 exports here so better stick to the old `module.exports`

negue:
the named exports works fine, those even help with importing the classes (in webstorm) instead of having an `module.export` so I'd rather keep em

no it was my fault, I somehow forgot to change the tests too

PR #10156, opened 2018-03-20
Add tooltip to character buff icon

ESWAT:
### Changes
Add tooltip to character buff icon. Makes it clearer what this icon is if the user hovers over it.

----

### DIFF
--- a/website/client/components/memberDetails.vue
+++ b/website/client/components/memberDetails.vue
@@ -17,7 +17,7 @@
       .d-flex.flex-column.profile-name-character
         h3.character-name
           | {{member.profile.name}}
-          .is-buffed(v-if="isBuffed")
+          .is-buffed(v-if="isBuffed", v-b-tooltip.hover.bottom="$t('buffed')")
             .svg-icon(v-html="icons.buff")
         span.small-text.character-level {{ characterLevel }}
     .progress-container(v-b-tooltip.hover.bottom="$t('health')")
--- a/website/client/components/tasks/task.vue
+++ b/website/client/components/tasks/task.vue
@@ -81,7 +81,7 @@
                 span.m-0(v-if="task.up && task.down") &nbsp;|&nbsp;
                 span.m-0(v-if="task.down") -{{task.counterDown}}
             .d-flex.align-items-center(v-if="task.challenge && task.challenge.id")
-              .svg-icon.challenge(v-html="icons.challenge", v-if='!task.challenge.broken', v-b-tooltip.hover.bottom="$t('challenge')")
+              .svg-icon.challenge(v-html="icons.challenge", v-if='!task.challenge.broken', v-b-tooltip.hover.bottom="`${task.challenge.shortName}`")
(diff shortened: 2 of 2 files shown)


paglias:
Hey @ESWAT, thanks for the PR but I'm going to ping our designer to see if this is wanted and in case it's maybe we want to apply it to more than a single icon? @Tressley

paglias:
Can you also post a screenshot of how this looks? Thanks

ESWAT:
@paglias I revised this branch to use a `tooltip` instead of `title` attribute, the intention is to display a clearer explanation of what this icon is.

screen shot 2018-03-24 at 20 54 47

Tressley:
@ESWAT -- Awesome work on this! Is it possible to add this to Task icons? Icons that need this are:

- Options
- Streak
- Due Date
- Challenge
- Broken Challenge

cc @paglias

ESWAT:
@Tressley I added tooltips for Streak, Challenge and Broken Challenge, however:

- I’m not sure what you mean by _Options_
- I don’t believe Due Date needs a tooltip… seems the text itself makes it clear what it is
- We might need new locale strings for Challenge and Broken Challenge (I’m using _Challenge_ and _Broken Challenge Link_ currently)

Tressley:
@ESWAT -- Great! Let me see if I can help. "Options" is the three dot icon that appears when you hover a Task. While you're correct in "Due Date" being fairly intuitive, we're aiming to establish a UX pattern with these tooltips. If all of the icons but the Due Date has a hover state on a Task, a user may perceive this as an error.

ESWAT:
@Tressley OK added those. I used _Show More_ for the text for the options icon since there isn’t an existing one for the word _Options_.

paglias:
@ESWAT in that case it's fine to add a new string for "Options" in common/locales/en unless @Tressley thinks that Show more is fine.

Also the show more is appearing at the top, not sure if this is okay @Tressley 

schermata 2018-03-31 alle 13 15 02

ESWAT:
@paglias @Tressley I decided to put it up top because it would block the display of the options menu a bit if the player didn’t move their pointer. Admittedly it does looks a bit weird like this.

negue:
@Tressley we could also show the challenge name and not just "challenge" , what do you think?

If `challenge` should be kept then its ready to be merged 👍

ESWAT:
FWIW I tried looking at moving the _Show More_ tooltip to the bottom and dismissing it when you click on the menu, but it doesn’t seem emitting `bv::hide::tooltip` does anything when the tooltip to hide is invoked by hover?

ESWAT:
@negue @Tressley I like the idea of putting the challenge name in the tooltip, so will explore that today or tomorrow.

Tressley:
@ESWAT -- Love the idea of using the Challenge name. Good idea! I think the tooltip appearing above is fine. We have some tooltip positioning hiccups currently. 👍

SabreCat:
Thanks for this, @ESWAT! Noting toward your next contributor tier!

PR #10161, opened 2018-03-21
Wrap attribute cell value text

ESWAT:
### Changes
When they become floating point or have more than 3 charcters they wrap to the next line. This CSS change prevents that.

Before:
screen shot 2018-03-20 at 22 23 06

After:
screen shot 2018-03-20 at 22 23 25

----


SabreCat:
Thanks for this, @ESWAT! Welcome to tier 2 of the Blacksmiths!

PR #10167, opened 2018-03-22
Stringified mem report

TheHollidayInn:


website/server/server.js
```diff
@@ -7,7 +7,7 @@ import http from 'http';
 import memwatch from 'memwatch-next';
 
 memwatch.on('leak', (info) => {
-  const message = `Memory leak detected:\n${info}`;
+  const message = `Memory leak detected:\n${JSON.stringify(info)}`;
```
paglias: you can also pass any JSON object directly to the logger: `logger.error(message, info);
 `

PR #10186, opened 2018-03-26
round priority on update too

### LINKED ISSUE #9275: Round task difficulty value

negue:
fixes #9275  (now on `PUT` route)


paglias:
looks good but can we fix this at the model level instead?

negue:
any idea why `gives out super awesome Masterclasser achievement when quests done out of order` is failing due to my changes ?

paglias:
I'm trying to figure out why that thing keep failing but it's not due to
your changes

Il Sab 31 Mar 2018, 20:16 negue <notifications@github.com> ha scritto:

website/server/models/task.js
```diff
@@ -131,6 +131,10 @@ TaskSchema.plugin(baseModel, {
       delete taskObj.value;
     }
 
+    if (taskObj.priority && !Number.isNaN(Number.parseFloat(taskObj.priority))) {
+      taskObj.priority = Number(taskObj.priority.toFixed(1));
```
paglias: shouldn't this line be `Number(taskObj.priority).toFixed(1);` since toFixed works on numbers, so it makes sense that you first make sure it's a number with Number and then call the toFixed method on the number

paglias (changes requested):

negue:
@paglias I refactored the code, now its only one parsing/casting instead of two

PR #10171, opened 2018-03-23
update antidote display and functionality - fixes #9758 and #10160

caleb-fung:
update antidote display and functionality - 

Fixes #9758 
Fixes #10160

[//]: # (Note: See http://habitica.wikia.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API for more info)

[//]: # 10160
https://github.com/HabitRPG/habitica/issues/10160 
https://github.com/HabitRPG/habitica/issues/9758

### Changes
[//]: # I created an additional `purchaseType` for `spell` attribute and corresponding case in `website/common/script/ops/buy/buy.js`. I also added `spell` to the `hideAmountSelectionForPurchaseTypes` constant in `website/client/components/shops/buyModal.vue` to display the purchase modal without numeric input (similar to greeting cards).

[//]: # a95371c2-08e6-4b73-a11d-7a87bfa5f81c

----
UUID: a95371c2-08e6-4b73-a11d-7a87bfa5f81c

PR #10199, opened 2018-03-29
antidote display and functionality

### LINKED ISSUE #9758: antidotes for avatar transformation buffs should not allow multiple purchases (salt, etc)
When a [Transformation Item](http://habitica.wikia.com/wiki/Transformation_Items) such as Snowball is used on you, the antidote (e.g., Salt) appears in your Rewards column to buy with gold. That works correctly, however the buy modal gives you the option to buy multiples, which isn't helpful and will be confusing (e.g., makes the users think they can store salt for use later). 

The buy modal should not have the text "How many would you like to buy?" or the field for selecting a number.

I'm labelling this as high priority since people will be using it a lot at the moment.

### LINKED ISSUE #10160: Transformation Item Antidotes do not work
I'm seeing reports of this in both staging and production, and so far it's been confirmed for at least Shiny Seeds and Spooky Sparkles.

When one goes to purchase an antidote, one is able to buy multiples which may not be desired behavior since those items were instant use. Upon purchasing any amount of them, no gold is deducted and my avatar is not un-transformed. This persists upon refresh or reload.

negue:
based on #10171, thanks to @CalebFung

fixes #9758 
fixes #10160

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -232,6 +232,7 @@
       "version": "2.0.2",
       "resolved": "https://registry.npmjs.org/acorn-dynamic-import/-/acorn-dynamic-import-2.0.2.tgz",
       "integrity": "sha1-x1K9IQvvZ5UBtsbLf8hPj0cVjMQ=",
+      "dev": true,
       "requires": {
         "acorn": "4.0.13"
       },
@@ -239,7 +240,8 @@
         "acorn": {
           "version": "4.0.13",
           "resolved": "https://registry.npmjs.org/acorn/-/acorn-4.0.13.tgz",
-          "integrity": "sha1-EFSVrlNh1pe9GVyCUZLhrX8lN4c="
+          "integrity": "sha1-EFSVrlNh1pe9GVyCUZLhrX8lN4c=",
+          "dev": true
         }
       }
     },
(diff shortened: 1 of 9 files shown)


SabreCat:
Looks pretty good! Just one issue, fairly minor: you can attempt to pin the antidote item in Rewards, which doesn't work (correct, shouldn't be able to pin), but it throws a strange error, `The item type "" is not valid.`

Should either (preferably) not show the pinning UI element at all, or (if that's not feasible) give a more informative error message like "This item is temporary and cannot be pinned."

caleb-fung:
@SabreCat I will take a look at it as soon as I can. I wasn't sure if @negue was wrapping this up. 
Are there any other features you wanted added/removed?

SabreCat:
@CalebFung Not to worry, @negue should be on it from here!

caleb-fung:
Okay, thanks for letting me help out! This is my first contribution to the Habitica repo. 
I hope to submit more thorough fixes in the future.

SabreCat:
@CalebFung Normally it's not a problem to have some back and forth during review, and to take a few days to implement each round of fixes. This was an unusual case!

paglias:
The problem mentioned by @SabreCat has been fixed so I'm merging

PR #10177, opened 2018-03-25
Move notification snackbars when resting bar is shown

ESWAT:
Fixes #10169

### Changes
Moves notification snackbars when resting bar is shown

screen shot 2018-03-24 at 21 46 06

----
UUID: ffee6b77-0bf8-422a-a65c-c20de17f2b32

PR #10203, opened 2018-03-31
Revert "Move notification snackbars when resting bar is shown (#10177)"

Alys:
This reverts #10177 because the console was showing errors like this:

[Vue warn]: Error in render: "TypeError: this.user is null"
found in
---> <App> at website/client/app.vue
       <Root>

This was happening in our staging server and also in my local install and so was preventing new code being tested effectively.

I'm going to merge this to staging now so that staff and mods can keep using staging for their regular use of Habitica and so that all contributors can keep basing new code off develop without seeing unrelated errors, but we can investigate what went wrong with #10177 and do a proper fix for it.

### DIFF
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -10,7 +10,7 @@ div
         p {{currentTip}}
   #app(:class='{"casting-spell": castingSpell}')
     amazon-payments-modal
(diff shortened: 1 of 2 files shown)


Alys:
The failed test is failing on other PRs too so it's not related to this change.

PR #10205, opened 2018-03-31
prevent a muted user from creating a public guild (API change)

Alys:
This PR modifies the API to prevent a public guild being created by a user who has had their chat privileges revoked. E.g., a young child is not allowed to create a public guild because they might put personal details in the description.

This does not prevent creating a party or private guild, both of which are still desirable for muted users so they can do solo quests or host private challenges for themselves. (NB https://github.com/HabitRPG/habitica/pull/10194 will prevent the user from inviting any other players to a private guild or party.)

We should also make UI changes in the website and mobile apps to make it clearer in advance that a muted user cannot create a public guild or invite users to a private guild, but that's less important than this API change to protect the community.

This PR also removes the `roundBigNumber` filter from the user's gem display in the header. This is because my test user happened to have more than 1000 gems and I needed to see the exact gem number to ensure gems weren't removed inappropriately. A similar requirement is likely to exist for other developers in future so I believe making this change permanent is advantageous.
The chance of a real user in production having more than 1000 gems is so small that I don't believe we need to consider it. Even if a user did have that many gems, they'd want to see the exact number to ensure that purchases were happening correctly.

### DIFF
--- a/test/api/v3/integration/groups/POST-groups.test.js
+++ b/test/api/v3/integration/groups/POST-groups.test.js
@@ -136,6 +136,21 @@ describe('POST /group', () => {
           },
         });
       });
+
+      it('returns an error when a user with no chat privileges attempts to create a public guild', async () => {
+        await user.update({ 'flags.chatRevoked': true });
+
+        await expect(
+          user.post('/groups', {
+            name: 'Test Public Guild',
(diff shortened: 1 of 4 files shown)


Alys:
Test fix : #10218

PR #10206, opened 2018-03-31
Stop using deprecated `moment.zone()`

ESWAT:
Replace all method calls with `moment.utcOffset()`. This stops the warnings thrown by `moment` in the console on app start since the former method is deprecated: http://momentjs.com/guides/#/warnings/zone/

**Note:** No tier contribution needed


negue:
Could check and fix the `npm run test:api-v3:unit` tests? seems that change did break something

TheHollidayInn:
@paglias I think we need a migration for this because we save each user's offset.

ESWAT:
@negue While testing locally I didn’t hit any unit test errors that were specific to this branch, just timeouts on the hooks that were already failing on `develop`.

Alys:
This needs to be done extremely cautiously to avoid cron failures. It needs more than just changing calls to the zone function because the zone data is stored in each player's account and controls cron's behaviour. We'd also need to check the mobile apps to see if they're saving zone data to the users' accounts and if so, make sure it's compatible with what the website does (that might turn out to be completely irrelevant but it should be checked). We'd want this change tested on a test website using a test database - we'd need to set up some test accounts first using the current `develop` code, put them through at least one cron, then deploy the zone change, and test thoroughly. I wrote some notes about this change a while back at [converting from moment().zone to moment().utcOffset](https://github.com/orgs/HabitRPG/teams/gryphons/discussions/2) (only visible to admins; that was partly as a test of github's team discussion feature, which probably isn't worth us keeping on using since it's too hidden even from us).

Alys:
See also notes here: https://github.com/HabitRPG/habitica/pull/7053
although I'm not sure how relevant that conversation is to our current code bases.

Alys:
I'm actually going to log a new issue for us to plan how to safely remove moment.zone() and I'll copy our thoughts about it so far into the issue so it's all in one place. I'll close this PR for now but we can reopen it later if needed. Thanks @ESWAT

PR #10208, opened 2018-03-31
Use consistent elements in footer links (fix add-on/forum link colors)

ESWAT:
### Changes
Use consistent elements in footer links

_Add-on_ and _Forum_ links appear differently due to how they are rendered. This commit is to address that, but is dependant on stripping out element tags from remaining translations, as demonstrated in `locales/en/front.json` This also removes the duplication in translations since both values reference the same URLs.

Before:
screen shot 2018-03-31 at 16 14 50

After (dependent on translations):
screen shot 2018-03-31 at 16 24 16

----

### DIFF
--- a/website/client/components/appFooter.vue
+++ b/website/client/components/appFooter.vue
@@ -135,23 +135,6 @@
         margin-bottom: 7px;
       }
     }
-
-    &.expanded {
-      padding-left: 6em;
-      padding-right: 6em;
-      padding-top: 3em;
-      background: #e1e0e3;
-      color: #878190;
-      min-height: 408px;
-
-      a {
-        color: #878190;
-      }
-
-      .logo {
-        color: #c3c0c7;
-      }
-    }
   }
 
   h3 {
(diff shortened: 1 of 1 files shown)


Alys:
The Wikia URLs need to be in the locales files so that translators can insert URLs for the non-English wikis. If you can't fix the display bug when the URLs are in the same string as the text then you can put the URLs in their own strings.

negue:
There are two issues here (on the current code, not on your side @ESWAT):

1. `a(v-html..` results in `<a><a href="...` which can result in expected issues ^^
2. since the style is `scoped` all elements get a random attribute to identify the scope, and elements that are added with`v-html`  don't receive this random attribute and so the style isn't applied

Could you extract the links styles so that those arent scoped, this way the styles will be applied all the time

ESWAT:
@Alys @negue OK, scrapped my initial commit and made new ones to address items 1 and 2. :)

screen shot 2018-04-20 at 21 38 42

paglias:

PR #10212, opened 2018-04-01
Fix locked class-specific gear after death fixes #10025

### LINKED ISSUE #10025: locked class-specific gear for your own class should show error message
When death has caused you to lose two or more equipment pieces that belong to your class and you try to buy back a more expensive one first, the website seems to let you buy it but the purchase fails with no notification.

What should happen is that the equipment should be locked with an explanation that you can't buy it using the `tierLockedItem` string ("This item is only available once you've purchased the previous items in sequence. Keep working your way up!"). I.e., the same behaviour should occur as for a new user who has not previously bought the equipment.

_Here's an example of a report from Dec (96749e48-ee70-4c3c-9829-e652a6836385) in the Report a Bug guild:_
_"I've died several times so I have quite a few items to buy again. I click Golden Sword, confirm purchase and it says "you bought golden sword!" but no gold is subtracted and gold sword is still in rewards and is not in the equipment tab. I resynced, closed habitica and opened, closed browser and opened - each time same behaviour. I realised, maybe I need to buy lower-tier items first. Ok, bought the axe fine. But why then is a) the golden sword displayed in rewards b) message "you bought golden sword" displayed to me instead of some error message."_

TNychka:

Fixes #10025
When death has caused you to lose two or more equipment pieces that belong to your class and you try to buy back a more expensive one first, the website seems to let you buy it but the purchase fails with no notification.
### Changes
Fixed incorrect boolean value. If the item.canOwn is false the item should be locked. New behavior is high tier gear is locked if death causes you to lose the higher and lower tier of that gear. No tests added because there aren't any item tests for the market (can add the test if desired).
Sorry for the delay on this fix

----

### DIFF
--- a/test/common/libs/shops.js
+++ b/test/common/libs/shops.js
@@ -59,6 +59,78 @@ describe('shops', () => {
       expect(specialCategory.items.find((item) => item.key === 'weapon_special_critical'));
       expect(specialCategory.items.find((item) => item.key === 'weapon_armoire_basicCrossbow'));// eslint-disable-line camelcase
     });
+
+    it('does not show gear when it is all owned', () => {
+      let userWithItems = generateUser({
+        stats: {
+          class: 'wizard',
+        },
+        items: {
+          gear: {
+            owned: {
+              weapon_wizard_0: true, // eslint-disable-line camelcase
+              weapon_wizard_1: true, // eslint-disable-line camelcase
+              weapon_wizard_2: true, // eslint-disable-line camelcase
+              weapon_wizard_3: true, // eslint-disable-line camelcase
+              weapon_wizard_4: true, // eslint-disable-line camelcase
+              weapon_wizard_5: true, // eslint-disable-line camelcase
+              weapon_wizard_6: true, // eslint-disable-line camelcase
+              armor_wizard_1: true, // eslint-disable-line camelcase
+              armor_wizard_2: true, // eslint-disable-line camelcase
+              armor_wizard_3: true, // eslint-disable-line camelcase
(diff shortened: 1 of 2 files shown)


TNychka:
Failing completely unrelated tests:
```
  1) GET challenges/user
       filters and paging
         does not page challenges if page parameter is absent:
      AssertionError: expected 14 to deeply equal 12
      + expected - actual
      -14
      +12
      
      at Context.it (test/api/v3/integration/challenges/GET-challenges_user.test.js:270:36)
      at <anonymous>
      at process._tickDomainCallback (internal/process/next_tick.js:228:7)
  2) POST /group
       Guilds
         public guild
           returns an error when a user with no chat privileges attempts to create a public guild:
     AssertionError: expected promise to be rejected but it was fulfilled with { Object (leaderOnly, quest, ...) }
```
None of the related market tests are failing so it should be good to go

TNychka:
I'm going to do some more testing on this, it might not be as easy as I thought it was.

negue:
@TNychka thanks for PR 👍 , the fix looks correct for me, I wonder why I didn't found this issue back then 🤔 , like you already wrote there aren't many tests available, could you try to add one for this particular issue?

You can take `shows relevant non class gear in special category` as an example

TNychka:
Sure, I'll can add a few tests :)

TNychka:
@negue Tests added, should be good to go!

negue:
Sorry for the long response time, yes thanks for the tests! 👍

paglias:
Thanks for this! Notated towards your next tier

PR #10216, opened 2018-04-01
pinned items fixes #10012

### LINKED ISSUE #10012: some pinned items should not become unpinned when you buy them (quests, eggs, potions, food)
When you pin any of the following items to your Rewards column and then buy it from that column, it unpins itself as soon as the purchase has been made:
- Quest scroll
- Pet Egg
- Hatching Potion
- Food

The correct behaviour is that it should remain pinned until the user chooses to unpin it.

Other items such as the greetings cards, transformation buffs (Snowball, etc), subscriber Gems, etc should also not become unpinned but they are already working correct and so no changes are needed for them.
The fix for this issue should NOT affect Equipment - gear pieces SHOULD become unpinned as soon as they are bought.

_Thanks to Remedy (5a008b2d-aedf-4f9f-bcf7-c02e9a9faecc)  for reporting this in the Report a Bug guild._

TNychka:

Fixes #10012
When you pin any of the following items to your Rewards column and then buy it from that column, it unpins itself as soon as the purchase has been made:
The correct behavior is that it should remain pinned until the user chooses to unpin it.

### Changes
Added a check on the type of item in buy to ensure that only gear items are removed.
**Assumes that ALL quests, eggs, potions, bundles, food, quests should remain pinned after purchase.**

----

### DIFF
--- a/website/common/script/ops/buy/purchase.js
+++ b/website/common/script/ops/buy/purchase.js
@@ -144,8 +144,11 @@ module.exports = function purchase (user, req = {}, analytics) {
     throw new NotAuthorized(i18n.t('notEnoughGems', req.language));
   }
 
-  let itemInfo = getItemInfo(user, type, item);
-  removeItemByPath(user, itemInfo.path);
+  let singlePurchaseTypes = ['gear'];
+  if (singlePurchaseTypes.indexOf(type) !== -1) {
+    let itemInfo = getItemInfo(user, type, item);
+    removeItemByPath(user, itemInfo.path);
+  }
 
   for (let i = 0; i < quantity; i += 1) {
     purchaseItem(user, item, price, type, key);
--- a/website/common/script/ops/buy/purchase.js
+++ b/website/common/script/ops/buy/purchase.js
@@ -106,6 +106,8 @@ function purchaseItem (user, item, price, type, key) {
   }
 }
 
+const acceptedTypes = ['eggs', 'hatchingPotions', 'food', 'quests', 'gear', 'bundles'];
+const singlePurchaseTypes = ['gear'];
 module.exports = function purchase (user, req = {}, analytics) {
   let type = get(req.params, 'type');
   let key = get(req.params, 'key');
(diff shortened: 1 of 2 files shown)


website/common/script/ops/buy/purchase.js
```diff
@@ -144,8 +144,11 @@ module.exports = function purchase (user, req = {}, analytics) {
     throw new NotAuthorized(i18n.t('notEnoughGems', req.language));
   }
 
-  let itemInfo = getItemInfo(user, type, item);
-  removeItemByPath(user, itemInfo.path);
+  let singlePurchaseTypes = ['gear'];
+  if (singlePurchaseTypes.indexOf(type) !== -1) {
```
negue: this line kinda confused me a bit 😁 , this could be converted to `.includes(type)` which is more understandable right away

website/common/script/ops/buy/purchase.js
```diff
@@ -144,8 +144,11 @@ module.exports = function purchase (user, req = {}, analytics) {
     throw new NotAuthorized(i18n.t('notEnoughGems', req.language));
   }
 
-  let itemInfo = getItemInfo(user, type, item);
-  removeItemByPath(user, itemInfo.path);
+  let singlePurchaseTypes = ['gear'];
```
negue: could you move this types-array and `acceptedTypes` outside of the method? it doesn't need to be initialized on every method call since its static anyway

negue (changes requested):
the fix looks correct, just a few requests and could also check the lint issue ?

website/common/script/ops/buy/purchase.js
TNychka: There was another line of code (132-133 purchase.js) which uses a similar style, would you like me to switch that one too? It's unrelated to this issue but it's in the same function.
```
 let acceptedTypes = ['eggs', 'hatchingPotions', 'food', 'quests', 'gear', 'bundles'];
  if (acceptedTypes.indexOf(type) === -1) {
```

TNychka:
Alrighty, I updated both my own and the previous check to use .includes now. Also I moved the variables outside the function, I'm not sure how else to prevent redeclaration. Is this the correct way to do that? Basing [this off of this](https://stackoverflow.com/questions/35271157/are-variables-initialized-inside-function-declaration-always-re-initialized-when?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa)

TNychka:
Should be good now. The coverage decreased due to moving those variables outside the function, can't be covered with tests as far as I'm aware.

negue:
kinda odd that this change decreases coverage

paglias:
thanks for this @TNychka , notated towards your next tier

PR #10220, opened 2018-04-02
Show toast when special pet is fed, don't throw error

### LINKED ISSUE #10217: "Can't feed this pet" error message should be visible to user not shown in the console
When you try to feed a Special pet (which can't be fed; e.g., Veteran Wolf), the message "Can't feed this pet" is reported in the browser's javascript console as shown below. The correct behaviour is that the user should see a "growl" notification showing that message (that's one of the notifications that appears near the top right of the website and then disappears after a couple of seconds).

```
Unhandled promise rejection 
httpCode: 401
​message: "Can't feed this pet."
​name: "t"
​stack: ""
```

This is the code:
https://github.com/HabitRPG/habitica/blob/e50bc189aa5afe211daf6f66cd88594140baaf80/website/common/script/ops/feed.js#L50-L52

I expect that the other errors listed there might also have the same problem so the fix for this issue should include them too.

ESWAT:
Fixes #10217

### Changes
Instead of issuing an exception when feeding pets of type `special` we just return back the _Can’t feed this pet_ message to be shown as a notification.

screen shot 2018-04-02 at 17 37 18

----

### DIFF
--- a/website/common/script/ops/feed.js
+++ b/website/common/script/ops/feed.js
@@ -47,12 +47,15 @@ module.exports = function feed (user, req = {}) {
     throw new NotFound(i18n.t('messageFoodNotFound', req.language));
   }
 
-  if (pet.type === 'special') {
-    throw new NotAuthorized(i18n.t('messageCannotFeedPet', req.language));
-  }
-
   if (user.items.mounts[pet.key]) {
     throw new NotAuthorized(i18n.t('messageAlreadyMount', req.language));
+  }
+
+  if (pet.type === 'special') {
+    return [
+      userPets[pet.key],
+      i18n.t('messageCannotFeedPet', req.language),
+    ];
   }
 
   let message;


paglias:
The right way to do this would be to wrap the call to feed in a try/catch block and if an exception is thrown show the message, a bit like here https://github.com/HabitRPG/habitica/blob/develop/website/client/components/tasks/column.vue#L628

Alys:
Also, can you please check the other errors listed in that file to see if
they have the same problem? The fix for the issue should include them too.

Can you also please comment on the issue to say you are working on it? That
will ensure that GitHub emails you if the issue has any comments added to
it. It's actually safest to post to an issue before you start working and
then wait for an admin's response - that ensures that no one else
duplicates the effort to fix it and that the issue isn't on hold for any
reason. Have a look at [this wiki page] (
http://habitica.wikia.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API)
for more details and other useful tips. :)

ESWAT:
Won’t have time to look into this PR further; will leave a note in the original issue about vacancy.

PR #10221, opened 2018-04-03
Wrap creator icon + text in @click event

ESWAT:
### Changes
Previously only clicking an icon would activate tabs in the creator. Personally I found it annoying that clicking on the associated text wouldn’t activate the tab as well, so this changes that.

screen shot 2018-04-02 at 23 06 20

----


SabreCat:
These proactive improvements are appreciated! Getting close to tier 3, methinks...

PR #10226, opened 2018-04-05
Prevent accidental submission of Tavern/Guild posts after pasting

### LINKED ISSUE #10117: Ctrl-V in a chat message can cause it to send
We've had two reports indicating that hitting Ctrl-V to paste text into a chat message can cause the message to be sent. 

The first was from donotgogentle (65be576c-c412-41c7-a31f-94994c8c4fc6) about six days ago:

_"Sorry for the double post. Whenever I use Ctrl+V my post gets sent automatically (happens in all guilds) without my clicking 'Send'"_

They never responded to my request for more information but I'll chase them about it now that we've had a second report. Here's what I asked: "Is that on both Firefox and Chrome? Are you using Windows or Mac OS or Linux? Does it still happen if you deliberately copy less than a full line of text? E.g., type a few words in a text editor, select only some of the words from the **middle** of the line, copy that, then paste it."

The second report is from kAbi, child of Light (7adf6ada-3c05-4054-b5df-fa7d49d3b9eb) today:

_"in guilds like the short and long term accountability guilds and in the tally board of the duelling grounds I have to write reports and for each bullet point I will use an emoji. I use the pc, so I will copy in my emoji with ctrl+V and then type my message, then hit enter 2 times, then hit ctrl+v to copy in my emoji (e.g white_check_mark for when the task got done, or x for when the task didn't got done) repeat._

_"it often happens, and I mean really often, that the message is getting send before I finish writting my whole report._

_"I'm aware there is the shortcut to hit ctrl+enter to send the message. now I'm wondering, whether I'm just too fast in typing, and so when I hit enter two times to create the space needed to separate the paragraph and then hit ctrl+v maybe I'm too fast in hitting ctrl and my computer interprets it as me hitting enter+ctrl, hence sends the message._

_"though, this all happens so fast, that usually I'm in the midst of the sentence I wanted to write when suddenly my message is gone from the writing space because it got posted. usually then a part of what I intended to write will appear in the writting window, the starting won't be there, probably because during that time it was sending the message._

_"I don't know how to hinder this from happening. I tried typing slower and it seemed to work, but I'm not a patient person and a fast typer. and this is happening sooo often I really wonder whether there might be a bug happening. Am I just too clumsy?"_

 kAbi, child of Light  says they are using Firefox version 58.0.2, Windows 10 Version 1709

siege918:
Fixes #10117

### Changes
Upon pasting into the post boxes on the Tavern or Guild pages, a boolean flag is set to temporarily disable the ctrl-enter shortcut for posting. Currently, this disables the shortcut for .5 seconds, but this value can be easily adjusted (on lines website/client/components/groups/group.vue:571 and website/client/components/groups/tavern.vue:764).

I had issues running tests locally (both on develop and my own branch), so I don't know how they behave until I see the results from submitting this PR.

----
UUID: 8b5e8110-0d52-43d5-8c20-998182d0b13c


website/client/components/groups/group.vue
```diff
@@ -552,6 +552,30 @@ export default {
       this.$store.state.memberModalOptions.fetchMoreMembers = this.loadMembers;
       this.$root.$emit('bv::show::modal', 'members-modal');
     },
+    disableMessageSendShortcut () {
+      // Some users were experiencing accidental sending of messages after pasting
+      // So, after pasting, disable the shortcut for a second.
+      this.chatSubmitDisable = true;
```
negue: custom variables should go into `data()`

website/client/components/groups/group.vue
```diff
@@ -552,6 +552,30 @@ export default {
...
+      if (this.chatSubmitTimeout) {
+        // If someone pastes during the disabled period, prevent early re-enable
+        clearTimeout(this.chatSubmitTimeout);
+        this.chatSubmitTimeout = null;
+      }
+
+      let This = this;
+
+      this.chatSubmitTimeout = window.setTimeout(() => {
+        This.chatSubmitTimeout = null;
```
negue: there shoudn't be a need to use temp variables for `this` during compilation everything inside a `lamdba method` can access the outer scope

negue (changes requested):
the fix looks solid, could you try to extract the textarea from both `vue`s and have it inside its own component? that way we'd only have one component to fix in the future

siege918:
I'll take a look at these this afternoon.

siege918:
Committed a couple of changes to address your comments, but going to take a little longer to see if I can abstract the textbox. This is admittedly my first time working in vue.js but I'll see what I can do.

siege918:
I've tried to extract the component, but it's clear that I don't know vue.js or Habitica well enough to make sure all the components and dependencies for guild chat are carried into the component correctly.

website/client/components/groups/group.vue
```diff
@@ -397,6 +397,10 @@ export default {
         description: true,
         challenges: true,
       },
+	  chat: {
+		SubmitDisable: false,
```
negue: could you use the `camelCase` naming for the inner properties here?

negue:
@siege918 don't worry, if you can't extract it I can do it in a different PR and ping you in that one (if you want to see the refactoring^^)

if you fix the lint issues then I guess its ready 👌

website/client/components/groups/group.vue
siege918: My mistake! Missed this when I refactored. I'll get that fixed.

siege918:
Definitely do ping me on the refactor, I'm interested to see!

siege918:
@negue Should be in good shape now!

paglias (changes requested):
@siege918 thanks for this, the code looks good but it looks like it's duplicate between two files, can you consolidate the shared code in a single component / mixin? Thanks

siege918:
Hi @paglias! As I mentioned further up, I don't feel comfortable doing so due to this being my first time using vue.js and looking at Habitica, and I couldn't wrap my head around the dependencies that I'd have to change to make it work.

@negue offered to make this change in a separate PR instead.

paglias:
Oh, sorry, I missed that part!

negue:
Sorry it took so long, LGTM 👍 and I'll ping you once this is merged for the extracted component

SabreCat:
Thanks for this, @siege918, and welcome to the ranks of the Blacksmiths at tier 1! This will undergo additional testing in a staging environment for a few days, and if all is well go live in production around the end of next week.

Note that achieving further tiers requires increasing amounts of effort from one to the next, so Tier 2 may not come immediately after your next PR. But keep helping out and we'll express our gratitude accordingly!

PR #10236, opened 2018-04-09
adding name attribute to radio inputs

### LINKED ISSUE #10234: Radio button for daily frequency is broken
### General Info
  * Browser: Google Chrome
  * OS: Windows 10

### Description
[//]: # When trying to change the repeat options for a daily, the radio button for changing it to "day of the week" is highlighted without unhighlighting "day of the month". It can be unselected by opening the tab button, and does not persist when you save the task.

image

fentie:
Fixes #10234

### Changes
Adds a shared `name` attribute to the two "repeats on" radio buttons so the browser treats them as a set vs. two unrelated radios. This disallows selecting more than one at a time.

----

### DIFF
--- a/website/client/components/tasks/taskModal.vue
+++ b/website/client/components/tasks/taskModal.vue
@@ -129,10 +129,10 @@
             label.d-block(v-once) {{ $t('repeatOn') }}
             .form-radio
               .custom-control.custom-radio.custom-control-inline
-                input.custom-control-input(type='radio', v-model="repeatsOn", value="dayOfMonth", id="repeat-dayOfMonth")
+                input.custom-control-input(type='radio', v-model="repeatsOn", value="dayOfMonth", id="repeat-dayOfMonth" name="repeatsOn")
                 label.custom-control-label(for="repeat-dayOfMonth") {{ $t('dayOfMonth') }}
               .custom-control.custom-radio.custom-control-inline
-                input.custom-control-input(type='radio', v-model="repeatsOn", value="dayOfWeek", id="repeat-dayOfWeek")
+                input.custom-control-input(type='radio', v-model="repeatsOn", value="dayOfWeek", id="repeat-dayOfWeek" name="repeatsOn")
                 label.custom-control-label(for="repeat-dayOfWeek") {{ $t('dayOfWeek') }}
 
         .tags-select.option(v-if="isUserTask")


paglias:
this seems to work well, thanks @fentie ! I'm merging it now given the importance of the bug

paglias:
@fentie  thanks for this! I'm merging the PR and awarding your contributor tier.

SabreCat:
@paglias @fentie I'm not seeing this work locally. The radio button appears to work, but it throws a console error `Computed property "repeatsOn" was assigned to but it has no setter.` and doesn't save when the user completes the workflow.

fentie:
@SabreCat interesting, I'll take a look tonight. Sounds like I need to somehow exclude that property from the model

paglias:
Hmm I hadn't noticed the console error, I gave a look and think this could be the right solution:

```
matteo@macbook~/dev/habitrpg/habitica$ git diff
diff --git a/website/client/components/tasks/taskModal.vue b/website/client/components/tasks/taskModal.vue
index b95bdbb34..576b6e7a5 100644
--- a/website/client/components/tasks/taskModal.vue
+++ b/website/client/components/tasks/taskModal.vue
@@ -702,6 +702,7 @@ export default {
         con: 'constitution',
         per: 'perception',
       },
+      repeatsOn: 'dayOfMonth', // default
     };
   },
   watch: {
@@ -732,6 +733,9 @@ export default {
     'task.frequency' () {
       this.calculateMonthlyRepeatDays();
     },
+    repeatsOn () {
+      this.calculateMonthlyRepeatDays();
+    },
   },
   computed: {
     ...mapGetters({
@@ -787,17 +791,6 @@ export default {
         return task.everyX === 1 ? this.$t('year') : this.$t('years');
       }
     },
-    repeatsOn: {
-      get () {
-        let repeatsOn = 'dayOfMonth';
-
-        if (this.task.type === 'daily' && this.task.weeksOfMonth && this.task.weeksOfMonth.length > 0) {
-          repeatsOn = 'dayOfWeek';
-        }
-
-        return repeatsOn;
-      },
-    },
     selectedTags () {
       return this.getTagsFor(this.task);
     },
(END)
```

Basically when the radio input changes the `calculateMonthlyRepeatDays` function is called, that should set the correct values for task.daysOfMonth and `task.daysOfMonth` and `task.weeksOfMonth` while before we did the inverse and changed the radio input based on the values of `task.daysOfMonth` and `task.weeksOfMonth`.

I tested it locally and it seems to work fine in that the task is saved, and the data seems correct but there's a lot of logic around so I'm not sure. 

@fentie sorry about this but I'm going to post a PR right now since it's an important bug and we usually deploy a new version of the website every Thursday

paglias:
Actually went with a simpler approach, see https://github.com/HabitRPG/habitica/pull/10247

fentie:
@SabreCat interesting, I'll take a look tonight. Sounds like I need to somehow exclude that property from the model

fentie:
@paglias no problem at all. That does look like a more a more robust solution. Good catch @SabreCat

PR #10248, opened 2018-04-12
Apple pay tests

### LINKED ISSUE #10246: iOS: recreating ongoing subscriptions as new ones causes failure in cron's monthly subscription benefits
This occurs for only Apple subscriptions but @phillipthelen says it's not occuring in the app (https://github.com/HabitRPG/habitica-ios/issues/684#issuecomment-380716620).

Mershmella (6c8e2dc6-d78a-4774-b33c-bda79d88192d) reported in the Report a Bug guild that their `gemsBought` value had not reset during cron at the beginning of this month (i.e., they couldn't buy gems with gold this month). I've confirmed that from Amplitude data and I've reset it manually. This also happened to them last month and at that time I suspected it was caused by them having too much history in some tasks (which can cause cron to time out part way through). I cleared all their history last month. This month they still have small history arrays, so I don't think that can be the cause of their gems not resetting this time.

I think the problem was caused by their subscription being recreated from scratch at the start of each month and that seems to be happening only for Apple subscriptions.

Mershmella has an Apple subscription:
```
{
    "_id": "6c8e2dc6-d78a-4774-b33c-bda79d88192d",
    "profile": {
        "name": "Mershmella"
    },
    "purchased": {
        "plan": {
            "consecutive": {
                "count": 8,
                "offset": 0,
                "gemCapExtra": 10,
                "trinkets": 0
            },
            "quantity": 1,
            "extraMonths": 0,
            "gemsBought": 35,
            "mysteryItems": [],
            "additionalData": "MIIhww...GYUA==",
            "customerId": "32...73",
            "dateCreated": "2017-07-01T17:38:21.482Z",
            "dateTerminated": null,
            "dateUpdated": "2018-04-01T16:14:01.847Z",
            "nextPaymentProcessing": "2018-04-10T18:02:51.592Z",
            "owner": "6c8e2dc6-d78a-4774-b33c-bda79d88192d",
            "paymentMethod": "Apple",
            "planId": "basic_earned"
        }
    },
}
```

Whenever a new subscription is created (not when an ongoing subscription renews), a message is posted to the moderators' Slack to help us troubleshoot subscription data, and we have multiple such records for Mershmella this year:
```
Mershmella 6c8e2dc6-d78a-4774-b33c-bda79d88192d undefined bought a 1-month recurring subscription using Apple on Thu Feb 15 2018 04:16:43 GMT+0000 (UTC)
Mershmella 6c8e2dc6-d78a-4774-b33c-bda79d88192d undefined bought a 1-month recurring subscription using Apple on Thu Feb 15 2018 04:16:43 GMT+0000 (UTC)
Mershmella 6c8e2dc6-d78a-4774-b33c-bda79d88192d undefined bought a 1-month recurring subscription using Apple on Thu Feb 15 2018 04:16:43 GMT+0000 (UTC)
Mershmella 6c8e2dc6-d78a-4774-b33c-bda79d88192d undefined bought a 1-month recurring subscription using Apple on Thu Mar 01 2018 19:16:37 GMT+0000 (UTC)
Mershmella 6c8e2dc6-d78a-4774-b33c-bda79d88192d undefined bought a 1-month recurring subscription using Apple on Thu Mar 01 2018 19:16:37 GMT+0000 (UTC)
Mershmella 6c8e2dc6-d78a-4774-b33c-bda79d88192d undefined bought a 1-month recurring subscription using Apple on Thu Mar 01 2018 19:16:38 GMT+0000 (UTC)
Mershmella 6c8e2dc6-d78a-4774-b33c-bda79d88192d undefined bought a 1-month recurring subscription using Apple on Sun Apr 01 2018 16:14:01 GMT+0000 (UTC)
```
There's supposed to be just one of those records each time a subscription is created, so seeing seven makes it look as if the subscription was created from scratch seven times this year. I asked Mershmella if they've been manually cancelled and recreating their subscription each month, but they said "I have automatically renewed since I started my subscription (through the Apple store)".

So it looks like something in Habitica's code is causing an ongoing subscription to be recreated as if it's a new subscription. Looking through our Slack records, I can see similarly duplicated records for about 40 users in March, all for Apple subscriptions (none for any other payment method). This is a problem because when an ongoing subscription is recreated as if it were a new subscription at the beginning of the month, it causes a faiure in the process that adjusts the user's monthly subscription benefits. In detail:

From the above data, you can see that the new subscription was created for Mershmella at Apr 01 2018 16:14:01, and this resulted in their subscription's `dateUpdated` value being set to "2018-04-01T16:14:01.847Z" - that value is always set when a new subscription is created, by this code: https://github.com/HabitRPG/habitica/blob/fa945c768949bf00adba2281483677cb66dc2c8c/website/server/libs/payments/payments.js#L331

Mershmella's cron on 1 April ran a few minutes after their subscription was recreated: 2018-04-01T16:21 ("April 1, 2018 4:21:16 pm UTC" in Amplitude). Part of cron's processing is to check the `dateUpdated` value for a subscription - if it is in a previous month, then cron runs the code that resets the user's monthly gem cap and increases their consecutive month benefits:
https://github.com/HabitRPG/habitica/blob/fa945c768949bf00adba2281483677cb66dc2c8c/website/server/libs/cron.js#L65-L68

Because `dateUpdated` had already been set to a date in April before cron ran, cron couldn't tell that it needed to reset their gem cap.

I.e., having Apple subscriptions recreated from scratch each month is damaging if it happens before the user's first cron of the month. Users affected by this will never have their monthly gem caps reset and won't earn consecutive month benefits. So ideally we'd either work out what is causing Apple subscriptions to be recreated from scratch, or we'd need to adjust how cron assesses subscriptions so that it can cope with the dateUpdated value being reset before the first cron of the month.

Until this is fixed, admins will need to find all affected users each month and manually:
- reset their `purchased.plan.gemsBought` value (set to 0)
- increase their `purchased.plan.consecutive.count` by 1
- decrease their `purchased.plan.consecutive.offset` by 1 if it's not 0
- increase `purchased.plan.consecutive.gemCapExtra` by 5 if `count` is now at a multiple of 3 and `offset` is 0 and the value is not already 25 (never let it be higher!)
- increase `purchased.plan.consecutive.trinkets` (Hourglasses) by 1 if `count` is now at a multiple of 3 and `offset` is 0

@phillipthelen says "This might be caused by something in the habitica-email-server code, since that goes through the user data. Although I’m not sure right now that it actually updates anything on the subscription, since it should mostly just check with apple if the subscription is still active and cancel it on our side if not." -- https://github.com/HabitRPG/habitica-ios/issues/684#issuecomment-380716620

By the way, Mershmella is being affected by the "RECEIPT_ALREADY_USED" problem. I don't know if that's related. Here's one example from Loggly:
```
message:NotAuthorized: RECEIPT_ALREADY_USED
    at Object.subscribe (/usr/src/habitrpg/website/transpiled-babel/libs/applePayments.js:143:29)
    at <anonymous>
method:POST
originalUrl:/api/v3/iap/ios/subscribe
```
and screenshots of others from around 1 April:
6c8e2dc6-d78a-4774-b33c-bda79d88192d

TheHollidayInn:
Fixes #10246

This also adds other tests besides the fix above because I was already working on this yesterday.

### DIFF
--- a/gulp/gulp-tests.js
+++ b/gulp/gulp-tests.js
@@ -167,7 +167,7 @@ gulp.task('test:content:safe', gulp.series('test:prepare:build', (cb) => {
 
 gulp.task('test:api-v3:unit', (done) => {
   let runner = exec(
-    testBin('node_modules/.bin/istanbul cover --dir coverage/api-v3-unit --report lcovonly node_modules/mocha/bin/_mocha -- test/api/v3/unit --recursive --require ./test/helpers/start-server'),
+    testBin('node_modules/.bin/istanbul cover --dir coverage/api-v3-unit node_modules/mocha/bin/_mocha -- test/api/v3/unit --recursive --require ./test/helpers/start-server'),
     (err) => {
       if (err) {
         process.exit(1);
--- a/gulp/gulp-tests.js
+++ b/gulp/gulp-tests.js
@@ -167,7 +167,7 @@ gulp.task('test:content:safe', gulp.series('test:prepare:build', (cb) => {
 
 gulp.task('test:api-v3:unit', (done) => {
   let runner = exec(
-    testBin('node_modules/.bin/istanbul cover --dir coverage/api-v3-unit node_modules/mocha/bin/_mocha -- test/api/v3/unit --recursive --require ./test/helpers/start-server'),
+    testBin('node_modules/.bin/istanbul cover --dir coverage/api-v3-unit --report lcovonly node_modules/mocha/bin/_mocha -- test/api/v3/unit --recursive --require ./test/helpers/start-server'),
     (err) => {
       if (err) {
         process.exit(1);
--- a/test/api/v3/unit/libs/payments/apple.test.js
+++ b/test/api/v3/unit/libs/payments/apple.test.js
@@ -57,6 +57,18 @@ describe('Apple Payments', ()  => {
         });
     });
 
+    it('should throw an error if getPurchaseData is invalid', async () => {
+      iapGetPurchaseDataStub.restore();
+      iapGetPurchaseDataStub = sinon.stub(iapModule, 'getPurchaseData').returns([]);
+
+      await expect(applePayments.verifyGemPurchase(user, receipt, headers))
+        .to.eventually.be.rejected.and.to.eql({
+          httpCode: 401,
+          name: 'NotAuthorized',
+          message: applePayments.constants.RESPONSE_NO_ITEM_PURCHASED,
+        });
+    });
+
     it('errors if the user cannot purchase gems', async () => {
       sinon.stub(user, 'canGetGems').returnsPromise().resolves(false);
       await expect(applePayments.verifyGemPurchase(user, receipt, headers))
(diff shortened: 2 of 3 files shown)


paglias (approved):
LGTM

PR #10257, opened 2018-04-15
increase user count on home page from 2.5 to 3 million

Alys:
This edits the /static/home page to change the number of users from 2.5 to 3 million.

It also modifies the locales string to use a variable instead of hard-coding the number. Because of this, before this is deployed, we'll need to push the locales files to Transifex and pull them back down again, otherwise non-English users will see error messages from their locales string not containing the variable. FYI @SabreCat 

This PR also removes some old, unused locales strings and an associated variable from when we had a million users.

image


paglias (approved):

hayamiz:
@Alys @paglias 

I've just noticed this PR from Habitica Forge tweet.

For Japanese, translating `<%= userCountInMillions %> million` naturally is very difficult, and I imagine the same problem could happen for other languages. So I suggest hard-coding the number is better.

Natural expression of "XXX million" in Japanese varies depending on `XXX`, and cannot expressed with a single placeholder.
For example:
- when `XXX` is non-integer like `Y.Z` (`Y < 10`): "`Y`百`Z`十万" or "`YZ`0万"
- when `XXX` is a single digit integer `Y`: "`Y`百万" or "`Y`00万"
- when `XXX` is a two digit integer `YZ`: "`Y`千`Z`百万" or "`YZ`00万"
- when `XXX` is a three digit integer `YZW`: "`Y`億`Z`千`W`百万" or "`Y`億`ZW`00万"

I understand the motivation to remove the hard-coded number, which needs translation update for each update of the number. However, this PR could result in strange expression in some (at least Japanese) languages. Bad thing is that even active translators, who frequently checks untranslated strings on Transifex, cannot notice the situation.
"Guidance for Linguists" says "It is better to have no translation than a bad translation because poor quality translations result in complaints and can damage Habitica's reputation." And I think this PR potentially cause a bad translation silently.
For user count string, hard-coding the number is not elegant way but it is better than a parameterized string.

Alys:
@hayamiz Thank you for raising that! I didn't know.

For the moment, we can ignore numbers greater than 9, so the only the first two situations would need to be considered.

For 2.5 million, is `2.5 百万` never written in Japanese? If a Japanese speaker saw that, would they at least understand what it meant? Or is a decimal point not used at all in Japan - is there some other symbol that's used in maths?
How bad is that from a grammar perspective? Are there any ways that translators work around similar problems on the rest of the site (there's other places where numbers can be either single digits or x.y)?

hayamiz:
@Alys 

In contrast to suffixes for each 3-digit (thousand or kilo, million, billion) in English, big number is usually suffixed for each 4-digit in Japanese, and `万` is suffix for 10,000, `億` is for 100,000,000, etc.

> `2.5百万` never written in Japanese?

This is unnatural and we rarely see it.

> If a Japanese speaker saw that, would they at least understand what it meant? Or is a decimal point not used at all in Japan - is there some other symbol that's used in maths?

`百` is hundred, and `百万` is a suffix meaning "hundred `万`" == million. So I think a Japanese speaker can understand at least what is meant by `2.5百万`.
Use of decimal point itself is very usual in Japanese, but "decimal point + `百万`" is unusual.

If we consider user count up to 9 million, natural expression should be like these:
- 2.5 million : 250万
- 3 million: 300万
- 3.5 million: 350万
- 4 million: 400万
- ...

> How bad is that from a grammar perspective? Are there any ways that translators work around similar problems on the rest of the site (there's other places where numbers can be either single digits or x.y)?

For suffixed numbers like "XXX million", I feel the difficulty translating the string with placeholder.
For smaller numbers that have no suffix, there might be no such difficulties in translation, but I'm not sure because I've just started to translate Habitica this month.
"0.6 damage", "100 damage", "1.5 gold", "1000 gold" or something like that is easily translated into Japanese.
Numeric expression without suffix like "1,000,000 gold" also make sense.

Alys:
For this specific translation problem, I think the benefit of having the same number consistent across all translations outweighs the precision of the translation in one language. It's possible we won't update the number again until there's 4 million users so you could translate it with that assumption.

hayamiz:
@Alys 
I understand. As long as `userCountInMillions` is single or two digit integer, there will be no problem for Japanese ( "`userCountInMillions`00万人" gives natural expressions). If possible, please add this limitation as a comment for future updates.
Thank you for taking time.

PR #10188, opened 2018-03-27
moving developer-only strings to api messages

negue:
closes #8574

Issue #8574, opened 2017-03-16
Migrate API errors to apiMessages from i18n (no translations for developer-only messages)


paglias:
https://github.com/HabitRPG/habitica/commit/939712ad1f8e8c8e228907b30844fcd3bb5dae51 introduced the `apiMessages` library  (https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/apiMessages.js) because we've decided that API messages aimed at developers will only be in English from now on.

We should migrate old translated messages to the new lib.

madpink:
are you interested in this being multiple files from the outset, or lumping all the apiMessages into one and possibly splitting it out later?

Also, just to make sure I know what is involved before I volunteer, I believe that the tasks involved in this would be:

1. Copying all of the api messages into the new lib
2. Removing those same api messages from their current locations & translations
3. updating the appropriate pages to read from the apiMessages library

paglias:
@madpink if we start adding all message maybe it's better to split into multiple files.

And the steps are okay, the important thing is not to remove anything used in the client side and to only move API messages meant for developers, so I'd start with things about `req.params.xxxxx must be ...` and similar

SabreCat:
Re step 2, you would only need to remove them from the English translation files. Changes there will propagate to other languages via a scripted process we run.

Alys:
I've increased the priority on this because as noted in https://github.com/HabitRPG/habitica/issues/10008#issuecomment-367126096 having excess strings is an accessibility issue that slows down the website.

negue:
@paglias  @Alys 

Which error messages can be considered as developer-only messages? I'd select all errors messages which are checked if req.params exists but any others too? like if a group/task doesn't exist?

paglias:
I'd say everything with req.something is a good start

PR #10258, opened 2018-04-15
moving developer-only strings to api/common messages

negue:
Splitting the messages lib between api and common

based on #10188 
ref #8574


paglias:
This look good, can you add an explanation of why the strings are split with comments on apiMessages and commonMessages? I think people will find it a bit confusing that some

paglias:
Can you fix the conflicts here?

negue:
@paglias fixed

paglias:
@negue I think this needs one last change to account for your PR that refactored purchasing gems :) after that we can merge

PR #10260, opened 2018-04-16
Adding hand cursor to FAQ headings, and ability to address answers via URL hash

### LINKED ISSUE #10055: FAQ: make sub-sections linkable, adjust cursor
Two unrelated changes for the FAQ:

1. On the old website, the sections in the FAQ could be linked to individually, which was very useful when pointing new players at the specific information they needed. It would be good if that could be done again.

2. As reported by Shadow53 (aa21c230-34a1-44e6-92ea-07a5c46cf516): "On the static FAQ, the CSS should be modified to include the following:
`.faq-question accordion:hover { cursor: pointer; }`
"The FAQ questions are clickable and expand the answer when clicked, but the cursor used is the "select text" cursor, which makes it appear as though you do not click the text. The above CSS changes the pointer to the same one used for clicking buttons and links so users know to click the titles."

fentie:
Fixes #10055

### Changes
Change H2 accordian text to links to make them addressable via URL fragments. Also gives us the hand cursor for free.
Overriding link styling (color & such)
Adding Vue ref to headings
Adding onload handler to check for existing fragment, open that answer, and scroll the answer into view.

----

### DIFF
--- a/website/client/components/static/faq.vue
+++ b/website/client/components/static/faq.vue
@@ -3,32 +3,21 @@
     .row
       .col-12.col-md-6.offset-md-3
         h1#faq-heading {{ $t('frequentlyAskedQuestions') }}
-        b-card.faq-question(no-body, v-for='(heading, index) in headings', :key="index")
+        .faq-question(v-for='(heading, index) in headings', :key="index")
           h2(role="tab", v-b-toggle="heading", @click="handleClick($event)", variant="info") {{ $t(`faqQuestion${index}`) }}
           b-collapse(:id="heading", :visible="isVisible(heading)", accordion="faq", role="tabpanel")
-            b-card-body(v-markdown="$t('webFaqAnswer' + index, replacements)")
+            div.card-body(v-markdown="$t('webFaqAnswer' + index, replacements)")
         hr
         p(v-markdown="$t('webFaqStillNeedHelp')")
 </template>
 
 <style lang='scss' scoped>
-  .card {
-    background: transparent;
-    border: 0;
-    border-radius: 0;
-  }
-
   .card-body {
       margin-bottom: 1em;
   }
 
   .faq-question h2 {
     cursor: pointer;
-  }
-
-  .faq-question .card {
-    background: none;
-    border: none;
   }
 
   .faq-question .card-body {
--- a/website/client/components/static/faq.vue
+++ b/website/client/components/static/faq.vue
@@ -6,7 +6,7 @@
         .faq-question(v-for='(heading, index) in headings')
           h2.accordion(:ref='uniqueRef(index)')
             a(href="#", :id='heading', @click.stop.prevent='setActivePage(heading)') {{ $t(`faqQuestion${index}`) }}
-          div(v-if='pageState[heading]', v-markdown="$t('webFaqAnswer' + index, replacements)")
+          div(:class='classObject(heading)', v-markdown="$t('webFaqAnswer' + index, replacements)")
         hr
         p(v-markdown="$t('webFaqStillNeedHelp')")
 </template>
@@ -80,6 +80,9 @@
     methods: {
       setActivePage (page) {
         this.pageState[page] = !this.pageState[page];
+      },
+      classObject (heading) {
+        return { 'd-none': !this.pageState[heading] };
       },
       uniqueRef (index) {
         return this.headings[index];
--- a/website/client/components/static/faq.vue
+++ b/website/client/components/static/faq.vue
@@ -17,6 +17,9 @@
   .faq-question {
     margin-bottom: 1em;
   }
+  .faq-question h2 {
+      cursor: pointer;
+  }
 
   .faq-question .card {
     background: none;
--- a/website/client/components/static/faq.vue
+++ b/website/client/components/static/faq.vue
@@ -5,7 +5,7 @@
         h1 {{ $t('frequentlyAskedQuestions') }}
         .faq-question(v-for='(heading, index) in headings')
           h2.accordion(:ref='uniqueRef(index)')
-            a(href="#", :id='heading', @click.stop.prevent='setActivePage(heading)') {{ $t(`faqQuestion${index}`) }}
+            a(v-bind:href="`#${heading}`", id='heading', @click.stop.prevent='setActivePage(heading)') {{ $t(`faqQuestion${index}`) }}
           div(:class='classObject(heading)', v-markdown="$t('webFaqAnswer' + index, replacements)")
         hr
         p(v-markdown="$t('webFaqStillNeedHelp')")
--- a/website/client/components/static/faq.vue
+++ b/website/client/components/static/faq.vue
@@ -1,12 +1,14 @@
 <template lang="pug">
-  .container-fluid
+  .container-fluid(role="tablist")
     .row
       .col-12.col-md-6.offset-md-3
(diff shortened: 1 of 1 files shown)


website/client/components/static/faq.vue
```diff
@@ -4,7 +4,8 @@
       .col-12.col-md-6.offset-md-3
         h1 {{ $t('frequentlyAskedQuestions') }}
         .faq-question(v-for='(heading, index) in headings')
-          h2.accordion(@click='setActivePage(heading)') {{ $t(`faqQuestion${index}`) }}
+          h2.accordion(:ref='uniqueRef(index)')
+            a(v-on:click.stop.prevent="" href="#", :id='heading', @click='setActivePage(heading)') {{ $t(`faqQuestion${index}`) }}
```
negue: you can combine `v-on:click.stop.prevent=""` and `@click='setActivePage(heading)'` together

negue: the `href='#'` results in never having a link with hash which can be copied

negue: you could just use `cursor: pointer;` in your style, instead of adding `a` just for the cursor

negue (changes requested):
I don't get any URLs with hash. only '/static/faq#'

@fentie did you tried this just with anchor/links? -  instead of having to watch the hash inside vue and `window.scroll()` to that element

I know how its currently created, the answers only created on click which is kinda hard to use anchors

website/client/components/static/faq.vue
fentie: This is true but it does still allow external pages to target specific answers... I didn't want to force the page to only allow a single answer to be open at once, since that would be losing functionality it already has

fentie:
Wow sorry about the delay on this. I didn't get any activity notices about your responses 😖

website/client/components/static/faq.vue
fentie: cool, didn't know that. will update. thanks

fentie: The anchor tags were for the benefit of the external pages, not just for the pointer

fentie:
@negue the `window.scroll()` bit is there to counteract a visual oddity I found. If you arrive at the page via a link to, say, `static/faq#party-with-friends`, that question will open but it will be partially hidden by the header. The scroll call adjusts for the header size so the question and answer are both immediately visible

fentie:
I also updated the way the answers are created, so rather than waiting for a Vue change to re-render the template, it uses a CSS class change instead with all the answers rendered, just hidden

negue:
@fentie 

With the current code: 

clicking on a question doesn't show me the url with the hash, so there is no way to copy the url for someone else

also:

> I also updated the way the answers are created, so rather than waiting for a Vue change to re-render the template, it uses a CSS class change instead with all the answers rendered, just hidden

could this code may still be in your local git repo and not pushed to the github one?

fentie:
You're correct about the links. I didn't want to try to force the URL hash to match the open question because more than one question can be open at a time. This still allows us to create/derive links directly to specific sections of the FAQ though, which I thought was the main goal.

As far as that other change goes... yeah, I either didn't push it, or reverted it. I'll find it or redo it, and push again. It's basically changing the `v-if` to a conditional `:class=d-none`, still based on pageState.

fentie:
The new commit with the rendering change got pushed. should be ready for discussion again

fentie:
Is this still being looked at?

negue:
Sorry for the long response time!

I've rechecked it, works as expected! Thanks 👍 

Could you add `'#'+heading` to `href`, that way users can use the `open in new tab` functionality of any browser to have the target link ready to sent.

fentie:
@negue I think I've got that added. `'#'+heading` (and a number of other things I tried) didn't work but I found something that did. If there's a more elegant way to handle it I'm all ears, otherwise I think this is ready.

paglias:
So is the desired behavior to have the url with the hash that can be opened in a new window or copied to be shared but when opening a faq the hash isn't updated? 

because I just tested this locally and IMO it's confusing. Most people won't recognize that the link can be copied if they don't see it change in the address bar and when you open a link with the hash pointing to a specific question you and then open a different question the old hash remains

@Alys what do you think?

fentie:
@paglias I know, it's confusing. I'm not entirely happy with it, but I didn't want to remove the ability to have multiple FAQ sections open at once. If you're OK with that, I can redo this so that the hash works the way everyone expects.

Otherwise I'll see if I can change the hash on section change?

Alys:
I think it would be nicest if the address bar always contained the URL+hash for the answer most recently opened. 

I'm not sure it's necessary to keep the ability to have multiple answers open at once. I don't know how often that would be useful. I'd be fine with it being removed, but @fentie you'd want to wait for a staff member to comment before going ahead with that.

paglias:
Opinions @beffymaroo @Tressley @TheHollidayInn @SabreCat ?

Tressley:
I would prefer the address bar reflecting the URL + hash. I don't think multiple questions need to be opened at once.

fentie:
@negue are you still "requesting changes"? iirc that was from your initial review about a month ago.

negue:
I have weird jumping if I click on a question.

All of this tested on a mac:

on Vivaldi: the page scrolls under the answer text (out of visible bounds) - at opening the question
on FIrefox: the page scrolls under the answer text too but at closing the question
on Chrome: same as firefox

just to be sure I retested it again after I reinstalled the packages and it still happens to me

paglias:
Is this ready for review?

fentie:
@paglias  yes

website/client/components/static/faq.vue
```diff
@@ -1,18 +1,47 @@
...
-  .container-fluid
+  .container-fluid(role="tablist")
     .row
       .col-12.col-md-6.offset-md-3
-        h1 {{ $t('frequentlyAskedQuestions') }}
-        .faq-question(v-for='(heading, index) in headings')
-          h2.accordion(@click='setActivePage(heading)') {{ $t(`faqQuestion${index}`) }}
-          div(v-if='pageState[heading]', v-markdown="$t('webFaqAnswer' + index, replacements)")
+        h1#faq-heading {{ $t('frequentlyAskedQuestions') }}
+        b-card.faq-question(no-body, v-for='(heading, index) in headings', :key="index")
```
negue: is there any reason to use `b-card`/`b-card-body` instead of a `div` I didn't see any differences with `div`

website/client/components/static/faq.vue
```diff
@@ -69,8 +95,13 @@
...
     },
     methods: {
-      setActivePage (page) {
-        this.pageState[page] = !this.pageState[page];
+      isVisible (heading) {
+        return this.visible && this.visible === heading;
+      },
+      handleClick (e) {
+        if (!e) return;
+        const heading = e.originalTarget.nextElementSibling.id;
```
negue: `originalTarget` isn't available on every browser, here it would be better to use `target`

negue (changes requested):
Thanks you, now it works perfectly :)

Just the two mentioned requests and its ready to be merged.

Thanks for your work 👍

fentie:
@negue here ya go. thanks for all the reviewing

negue (approved):

SabreCat:
Contributor credit added!

PR #10266, opened 2018-04-19
Fix API challenges export CSV bug (Fixes #8350)

### LINKED ISSUE #8350: Unable to Export to CSV Challenge Data
## General Info

Opera 42.0.2393.85
Windows 10 32-bit

## Description

When export CSV challenge data receiving error message on screen:
{"success":false,"error":"InternalServerError","message":"An unexpected error occurred."}

Example Challenge:
[Clean the Dungeon (dailies slayer) – December – Winter Wonderland (Snowball all the monsters of the day) ID: 3f0be455-990c-49b0-90cc-2ef3b9214949](https://habitica.com/#/options/groups/challenges/3f0be455-990c-49b0-90cc-2ef3b9214949)

No Console error messages,

----

**edit by Alys 2017-08-10:**

## Suggested Fix

**The explanation for this bug and a suggested fix is in [this comment below](https://github.com/HabitRPG/habitica/issues/8350#issuecomment-321517442).**

----

## What to do if this happens to you

Post to the Report a Bug guild and copy the URL of the challenge into your post. An admin will probably be able to make the CSV file available.

ghost:
Fixes #8350

### Changes
In the challenge-export-to-csv API, two if-statements were added to make sure that an array includes a certain item before getting a property of that certain item (which would throw an error if that item didn't exist in the array). The first if-statement makes sure that ```resArray[index]``` exists before trying to access ```resArray[index][0]```. The second if-statement makes sure that ```resArray[index]``` exists before trying to call ```resArray[index]push({...})```. A simple test was also added for this case (which fails if the two mentioned if-statements are deleted/commented and passes if the if-statements are included in the codebase).

----
UUID: 98a3ed00-32d2-4a96-bc33-daa639b1c047

### DIFF
--- a/test/api/v3/integration/challenges/GET-challenges_challengeId_export_csv.test.js
+++ b/test/api/v3/integration/challenges/GET-challenges_challengeId_export_csv.test.js
@@ -71,4 +71,9 @@ describe('GET /challenges/:challengeId/export/csv', () => {
     expect(splitRes[4]).to.equal(`${sortedMembers[3]._id},${sortedMembers[3].profile.name},habit:Task 1,0,,0,todo:Task 2,0,,0`);
     expect(splitRes[5]).to.equal('');
   });
+
+  it('should successfully return when it contains erroneous residue user data', async () => {
+    await members[0].update({challenges: []});
+    await members[1].get(`/challenges/${challenge._id}/export/csv`);
+  });
 });
--- a/test/api/v3/integration/challenges/GET-challenges_challengeId_export_csv.test.js
+++ b/test/api/v3/integration/challenges/GET-challenges_challengeId_export_csv.test.js
@@ -74,6 +74,13 @@ describe('GET /challenges/:challengeId/export/csv', () => {
 
   it('should successfully return when it contains erroneous residue user data', async () => {
     await members[0].update({challenges: []});
-    await members[1].get(`/challenges/${challenge._id}/export/csv`);
+    const res = await members[1].get(`/challenges/${challenge._id}/export/csv`);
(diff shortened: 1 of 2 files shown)


test/api/v3/integration/challenges/GET-challenges_challengeId_export_csv.test.js
```diff
@@ -71,4 +71,9 @@ describe('GET /challenges/:challengeId/export/csv', () => {
     expect(splitRes[4]).to.equal(`${sortedMembers[3]._id},${sortedMembers[3].profile.name},habit:Task 1,0,,0,todo:Task 2,0,,0`);
     expect(splitRes[5]).to.equal('');
   });
+
+  it('should successfully return when it contains erroneous residue user data', async () => {
+    await members[0].update({challenges: []});
+    await members[1].get(`/challenges/${challenge._id}/export/csv`);
+  });
```
Alys: Could you also please test that the exported file contains the correct data? It should be the same data as for the previous test except that there'll be no data for `members[0]`

Thanks for this PR! It will be a relief when we don't have to manually repair challenges any more. :)

Alys (changes requested):

ghost:
This PR is ready for review again.

### Changes
---
As requested, a test was added to ensure that the exported CSV file contains the correct data. Additionally, the fix itself was refactored so that instead of there being two similar if-statements, there is now only one if-statement that ensures that a given task (in the task for-each loop) belongs to a user who is still involved in the challenge.

paglias:
@asherdale  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #10270, opened 2018-04-21
pinning mongodb container to recommended, supported version

fentie:
### Changes
Pinning the mongodb container to version 3.4, as specified in the install instructions, so Docker users don't have to install mongo locally.

----

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       - mongo
 
   mongo:
-    image: mongo
+    image: mongo:3.4
     ports:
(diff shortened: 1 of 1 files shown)


fentie:
@Alys saw you were discussing the install instructions in the blacksmiths guild. This should obviate the need for a local mongo install when using Docker.

paglias:
Thanks for this! Notated towards your next tier

Alys:
@fentie Thanks very much for this! I've removed the mongo install instructions from the new version of the Docker wiki page that I'm writing.

PR #10271, opened 2018-04-21
Purchase API Refactoring: Gems [Gold]

negue:
This is part of ongoing refactoring process of #10011 

Also contains following changes:
- removed `keyRequired` and use `missingKeyParam` (both strings have the "same" value, so it was redundant) 
- changed `plusOneGem` to also accept a quantity variable
- now operations can also configure their `analyticsLabel()`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5631,7 +5631,7 @@
       "dependencies": {
         "onetime": {
           "version": "1.1.0",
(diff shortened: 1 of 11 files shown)


paglias:
Is this ready for review?

negue:
now it is

PR #10283, opened 2018-04-26
Added bulk allocation

TheHollidayInn:


website/client/components/userMenu/profileStats.vue
```diff
@@ -120,7 +120,7 @@
       .row.title-row
         .col-12.col-md-6
           h3(v-if='userLevel100Plus', v-once, v-html="$t('noMoreAllocate')")
-          h3(v-if='user.stats.points || userLevel100Plus')
```
paglias, line 123: looks good, why was this check removed?

TheHollidayInn: The title was hidden when the user spent points, but that looked odd. Must have been pulled over from the old site.

PR #10285, opened 2018-04-26
Remove Incorrect Experience Notification When Leveling Up (fixes #9339)

### LINKED ISSUE #9339: Incorrect Experience Notification on Level-Up
We're getting reports that when people level up, they're getting notifications saying that they lost experience. While technically correct (since their experience reset), it's misleading. They shouldn't get the experience lost notification when they level up.

ghost:
Fixes #9339

### Changes
This PR fixes the following bug: previously, upon leveling up, a notification containing the text "You Lost Some Experience" was shown. This PR ensures that this incorrect notification is not shown. The function `getXPMessage ` returns undefined if a user loses more than 50 experience (which they technically do upon leveling up because they go from some high amount of XP before leveling up to a very low amount of XP after leveling up), so now with the added if-statement the user only sees a notification if the XP difference is bigger than -50.

----
UUID: 98a3ed00-32d2-4a96-bc33-daa639b1c047

### DIFF
--- a/website/client/mixins/notifications.js
+++ b/website/client/mixins/notifications.js
@@ -22,8 +22,10 @@ export default {
       this.notify(this.$t(type, { val }), 'success');
     },
     exp (val) {
-      let message = getXPMessage(val);
-      this.notify(message, 'xp', 'glyphicon glyphicon-star', this.sign(val));
+      const message = getXPMessage(val);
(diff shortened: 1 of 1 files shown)


paglias:
thanks! notated towards your next tier

PR #10305, opened 2018-04-30
Purchase API Refactoring: Spells [Gold]

negue:
This is part of ongoing refactoring process of #10011

Also:
- removed `purchaseWithSpell.js`
~~changed `purchaseType` of `special` to `spells`~~


test/common/ops/buy/buy.js
```diff
@@ -111,7 +111,7 @@ describe('shared.ops.buy', () => {
       params: {
         key: 'thankyou',
       },
-      type: 'special',
```
paglias: is this a breaking api change? is there a way to avoid it?

paglias (changes requested):

test/common/ops/buy/buy.js
negue: if we want to keep the `special` as type then I could revert this part, I just thought it would make it a bit less confusing for the developers and api users

paglias: yeah, it's something we can consider for a new api version but for now we shouldn't change it

Issue #9872, opened 2018-01-24
Menu's zindex is below snackbars


TheHollidayInn:
Menus are hidden behind the snackbar notifiactions

![image](https://user-images.githubusercontent.com/2686586/35277741-c11137b0-0015-11e8-979d-29ef0448a2f9.png)

The zindex should be changed to be above 1041

librarianmage:
Should this be marked as "In Progress"?

PR #10338, opened 2018-05-06
Beard and mustache facial hairs now can be bought as a full set for 5 gems

### LINKED ISSUE #10169: Notifications are covered by resting in the inn status banner
As reported by citrusella:

> The inn banner forces the navbar down but doesn't push down the notifications for gold and such, which places the navbar over top of the notifications (presumably because of its z-index? I think?).

I've confirmed this behaviour on production and staging. Here's a screenshot. 
image

### LINKED ISSUE #10335: beard and mustache facial hairs can't be bought as a full set (twice the correct gem cost is charged)
As reported in the Report a Bug guild by GeraldThePixel (3d129848-79e2-4d89-a3d7-5941a44c9a84):

"Hey! I was looking at the mustaches in character customization, and noticed that each of the two mustaches costs two gems, or you can buy both for five. I know this is less of an bug, but more of a content issue. But I figured here would be a good place for it."

Here's a screenshot showing the problem:

image

The bug is that it should cost 5 Gems to buy ALL of the facial hair together - beards AND mustaches. 

This is not just a display bug. If you try to buy the full set, you're charged 5 Gems (I confirmed that) but you end up with only half the set:

image

----

The correct behaviour is:
- a single 5 Gem purchase should give you all five facial hairs
- the 5 Gem button should appear only once on the facial hair screen to make it clearer that it gives you all five hairs (having it appear twice is misleading and does not allow the user to make an informed choice about whether to buy individual items)

pengfluf:
Fixes #10335: beard and mustache facial hairs can't be bought as a full set (twice the correct gem cost is charged)
Fixes #10169

### Changes
- There's only one button for buying the full set for 5 gems
- Notification 'Not enough Gems' is now visible

<details>
<summary>Before</summary>

before

</details>

<details>
<summary>After</summary>

after

</details>

---

Habitica UUID:
7aa1622e-18f9-4af6-86b3-8c5a017039fd

### DIFF
--- a/website/client/components/creatorIntro.vue
+++ b/website/client/components/creatorIntro.vue
@@ -1337,28 +1337,33 @@ export default {
       * @returns {boolean} - Determines whether the "Purchase All" button is needed (true) or not (false).
     */
     isPurchaseAllNeeded (types, keySets) {
-      const allItemsLengths = [];
       const purchasedItemsLengths = [];
-
-      // Key set must be specify correctly
-      keySets.forEach((keySet) => {
-        allItemsLengths.push(Object.keys(this[keySet]).length);
-      });
-      const allItems = allItemsLengths.reduce((acc, val) => acc + val);
-
-      // Type can be undefined, so we must check it
+      // Types can be undefined, so we must check them.
       types.forEach((type) => {
         if (this.user.purchased.hair[type]) {
           purchasedItemsLengths
             .push(Object.keys(this.user.purchased.hair[type]).length);
         }
       });
-      if (purchasedItemsLengths.length > 0) {
-        const purchasedItems = purchasedItemsLengths.reduce((acc, val) => acc + val);
-        const unpurchasedItems = allItems - purchasedItems;
-        return unpurchasedItems > 2;
+      // We don't need to count the key sets (below)
+      // if there are no purchased items at all.
+      if (purchasedItemsLengths.length === 0) {
+        return true;
       }
-      return true;
+
+      const allItemsLengths = [];
+      // Key sets must be specify correctly.
+      keySets.forEach((keySet) => {
+        allItemsLengths.push(Object.keys(this[keySet]).length);
+      });
+
+      // Simply sum all the length values and
+      // write them into variables for the convenience.
+      const allItems = allItemsLengths.reduce((acc, val) => acc + val);
+      const purchasedItems = purchasedItemsLengths.reduce((acc, val) => acc + val);
+
+      const unpurchasedItems = allItems - purchasedItems;
+      return unpurchasedItems > 2;
     },
     prev () {
       this.modalPage -= 1;
--- a/website/client/components/creatorIntro.vue
+++ b/website/client/components/creatorIntro.vue
@@ -171,11 +171,6 @@ b-modal#avatar-modal(title="", :size='editing ? "lg" : "md"', :hide-header='true
             .gem-lock(v-if='option.locked')
               .svg-icon.gem(v-html='icons.gem')
               span 2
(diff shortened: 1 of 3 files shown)


Alys:
Thanks for working on this so quickly!

I like that the "not enough gems" message appears over the top of the avatar modal - that's definitely an improvement! - but unfortunately it brings back the problem of the menu being hidden by the snackbar notifications (https://github.com/HabitRPG/habitica/issues/9872). To test for that, click a Habit multiple times then open a menu. (It also prevents the right-most menus being opened when the user is resting in the inn because the banner pushes the menu header down, but that's the fault of this bug: https://github.com/HabitRPG/habitica/issues/10169). If you can adjust the z-indexes so that the menus are always on top, that would seem like a good fix to me but there might be other factors I'm not considering (@TheHollidayInn would know better).

I've tested buying the facial hair and the button definitely gives you all five for 5 Gems which is perfect. However the button doesn't appear if you've already separately bought the two mustaches because of this line (line 182 in the PR's current code):
`.col-12.text-center(v-if='!userOwnsSet("hair", baseHair6Keys, "mustache")')`

I wonder if it's best to fix this by swapping the order of the beards and mustaches on the page (it makes sense for the mustaches to be above the beards anyway) and then that line could check for the existence of all the beards (not mustaches) - if the user owns all beards then you don't want to show the "purchase all" button because even if you don't own any mustaches yet, it's still cheaper to buy them individually. I think that would actually be better than checking if they own the full set of all 5 items.

By the way, when you want to start working on an issue, it's safest if you comment on the issue to say you're interested (that prevents two people working on the same issue at the same time) and then ideally wait for an admin to say you can go ahead (that saves your time in cases when an issue marked as help wanted should actually be on hold or needs extra information added to it, which does happen sometimes). However I've marked the issue as in progress now for your fix - thanks! :)

pengfluf:
> By the way, when you want to start working on an issue, it's safest if you comment on the issue to say you're interested and then ideally wait for an admin to say you can go ahead.

Okay.

### New changes:
- User menu, navbar and modal appearance fixed
- Mustaches and beards swapped
- The "Purchase All" button is visible only when there are more than 2 unpurchased facial hairs for 4 gems

facial-hair

Alys:
Fixes #10169
(Nice!)

Alys:
I've tested this locally and I like how it works!
- The purchase all button disappears correctly when there's only two items left to buy, regardless of what they are (a nice touch, and I like that you made that function generic instead of locking it down to facial hair).
- The purchase all button does correctly purchase all appropriate items (and the individual purchases still work as they do on the live site). 
- The error notification snackbars are above the modal but below the menus, which is good! The expected side-effect of that is of course that the XP/GP/etc snackbars also appear above modals but I think that's tolerable, and it's certainly better than having the error messages obscured (we see comments about that fairly often in the Report a Bug guild). If we get many complaints about the XP/etc notifications being over modals, we can create a new issue to find a way to treat them differently than the errors.

I haven't done a code review because I'm not strong on vue.

pengfluf:
@Alys, if you have any questions about the code, feel free to ask.

### New changes:
- The isPurchaseAllNeeded function refactored. Now we don't even start to count all the items, if there are no purchased items at all.

pengfluf:
### New changes:
Function isPurchaseAllNeeded is upgraded, now you can:
- specify any category, not only 'hair'
- optionally select which types (subcategories) you need
  - if not selected, all the purchased items will be recursively found in the specified category

paglias:
So this looks good but I just tested locally and the snackbars appear over the menus, can you fix it and then I'll merge the PR! I'm using chrome if it can help

paglias:
Thanks for this! Welcome to tier 2

PR #10340, opened 2018-05-07
Maximum length for summaries for Guilds and Challenges (#9244)

### LINKED ISSUE #9244: API must enforce maximum length for summaries for Guilds and Challenges

marvinrabe:

Fixes #9244

### Changes

Added express validations for summary field with tests.

----
UUID: b3eeebf4-b03e-48dd-9e3c-abb27410d1d4

### DIFF
--- a/test/api/v3/integration/groups/PUT-groups.test.js
+++ b/test/api/v3/integration/groups/PUT-groups.test.js
@@ -3,6 +3,7 @@ import {
   generateUser,
   translate as t,
 } from '../../../../helpers/api-v3-integration.helper';
+import {MAX_SUMMARY_SIZE_FOR_GUILDS} from '../../../../../website/common/script/constants.js';
 
 describe('PUT /group', () => {
   let leader, nonLeader, groupToUpdate, adminUser;
@@ -43,6 +44,26 @@ describe('PUT /group', () => {
     expect(updatedGroup.leader._id).to.eql(leader._id);
     expect(updatedGroup.leader.profile.name).to.eql(leader.profile.name);
     expect(updatedGroup.name).to.equal(groupUpdatedName);
+  });
+
+  it('updates a group summary', async () => {
+    let updatedGroup = await leader.put(`/groups/${groupToUpdate._id}`, {
+      summary: 'New summary',
+    });
+
+    expect(updatedGroup.summary).to.equal('New summary');
+  });
+
+  it('returns error when summary is too long', async () => {
+    let invalidSummary = '#'.repeat(MAX_SUMMARY_SIZE_FOR_GUILDS + 1);
+
+    await expect(leader.put(`/groups/${groupToUpdate._id}`, {
+      summary: invalidSummary,
+    })).to.eventually.be.rejected.and.eql({
+      code: 400,
+      error: 'BadRequest',
+      message: t('invalidReqParams'),
+    });
   });
 
   it('updates a group categories', async () => {
--- a/test/api/v3/integration/challenges/POST-challenges.test.js
+++ b/test/api/v3/integration/challenges/POST-challenges.test.js
@@ -4,12 +4,27 @@ import {
   translate as t,
 } from '../../../../helpers/api-v3-integration.helper';
 import { v4 as generateUUID } from 'uuid';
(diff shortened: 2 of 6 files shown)


marvinrabe:
Travis failed because of:

>    1) GET /challenges/:challengeId/members
>        supports using req.query.lastId to get more members:
>      Error: Timeout of 8000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/travis/build/HabitRPG/habitica/test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js)

Seems like a random error. Not my fault. 😔

test/api/v3/integration/groups/PUT-groups.test.js
```diff
@@ -3,6 +3,7 @@ import {
   generateUser,
   translate as t,
 } from '../../../../helpers/api-v3-integration.helper';
+import {MAX_SUMMARY_SIZE_FOR_GUILDS} from '../../../../../website/common/script/constants.js';
```
paglias, line 6: shouldn't this check be at the model level instead of doing it for every route?

marvinrabe: See: https://github.com/HabitRPG/habitica/issues/9244#issuecomment-351550431

> So, I think the best start here is to validate length during the api request.

marvinrabe: I dont know what the reason for this is. Though if requested I could move it to the model level instead. Seems more reasonable anyway…

paglias: Hmm I see, I think the problem might be that we don't want to start throwing errors if someone had a longer summery before the validation is introduced

Alys:
@marvinrabe Just to confirm, this won't give an error if the summary does not exist, right?

marvinrabe:
@alys No, it wont. There are tests for the "no summary case" and they are green. For example this one: https://github.com/HabitRPG/habitica/blob/bbbd1f9f73ab69905f9e5957c7055a6206a14d30/test/api/v3/integration/challenges/POST-challenges.test.js#L318

paglias (commented):
The code is okay but honestly I'm worried about the validation not being in the model. @Alys what do we want to do with summaries that are too long?

If we could cut them with a migration then we could move the validation to the model

Alys:
We can certainly fix any too-long summaries before this goes live but I don't think it's worth writing a migration script. There'll be so few that we can easily handle them manually and we'd want to adjust them manually anyway (e.g., reduce unneeded words rather than cut at after 250 characters). I've searched challenges and guilds just now and found only one that had a long summary, which I've shortened. I'll repeat the search before go-live and again just after (if someone could remind me when this is merged and again when deployed, that would be helpful but I'll make a note to check every day or so).
Sample code:
`db.groups.find({"summary": { "$regex": /^.{251,}$/} },{_id:1, summary:1},{multi:true});`

However if the validation is in the model, doesn't that mean it will be checked every time anything happens with the guild (e.g., a user viewing it), rather than just when it's needed (when the summary is changed)? If that's what you want, I'm fine with it but it seems like overkill. I could well be misunderstanding though!

paglias:
The validation at the model level would only happen when the field is modified so it's not a problem.

@marvinrabe can you change the code to have the validation happen at the model level?

marvinrabe:
@paglias Sure!

beffymaroo:
@marvinrabe Are you still working on this issue? Let us know when you have a moment. Thank you!

Issue #10294, opened 2018-04-28
To-Do audio sound effect plays only once


Alys:
If you mark a To-Do complete, you hear the expected sound effect but then if you mark others complete, you don't hear the sound.

If you then complete a different type of task and then a To-Do again, you hear the To-Do noise, but again it's just for one To-Do - further ones are silent again.

Note that when testing this on a local install, you can avoid interference from the drop sound effects by setting your `items.lastDrop.count` to a high value.

Amethystix:
I'm actually not even hearing the first sound effect; to-dos are just completely silent for me.  I tried refreshing the page, checking off something else before checking off a to-do, and even just creating and checking off more, but it's still silent while the other sounds play normally.  I'm using Chrome.

paglias:
Weird, this works locally but not on habitica.com (or staging)

paglias:
https://habitica.com/static/audio/danielTheBard/Todo.ogg (for example) doesn't exist in prod, that's why it's not playing

PR #10346, opened 2018-05-09
Fixed Lunar Battle quest scrolls showing wrong unlock conditions (Issue #10294 )

### LINKED ISSUE #10295: Lunar Battle quest scrolls show wrong unlock conditions
The three Lunar Battle quest scrolls are unlocked after a certain number of check-ins. However for the second and third scrolls, the Quest Shop says that you can unlock them by completing the preceding quests, which is wrong, as well as saying they unlock after check-ins, which is correct. 

In the two screenshots below, only the first sentence is correct. The second sentence should not appear.

image

image

For comparison, the text for the first Lunar Battle quest scroll is correct:
image

DoguD:

Fixes #10295

### Changes
Even though the Lunar Battle quest scrolls unlocked according to check-in count, in Shop/Quests, locked Lunar Battle Part 2 & 3 shows a message which implies the scroll will be unlocked when the user finishes the previous quest. This message is removed.
In order to achieve this, a new method is added to website/client/components/shops/quests/index.vue file which checks whether the key of quest scroll item is in a specified list of keys. In the future, other quest keys can be added to the list as well if their unlock condition is not dependent on the completion of previous quests.
Before: 

After:
screen shot 2018-05-03 at 16 51 25

----
UUID: 1b2e5d14-e918-44aa-b669-7c309781fdba


paglias:
@DoguD  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #10349, opened 2018-05-10
groups: Fix Markdown for claimed task messages

aszlig:
So far if a task contained Markdown, a system message like this would have been posted to the group chat:

```markdown
foo has claimed "Some [link](http://example.org/)"
```

Also, if the Markdown contained backticked code fragments, the whole text would be displayed in red except the code part.

The reason for this is because the system message is already in Markdown and a backticked task text would result in the following Markdown:

```markdown
`foo has claimed "Foo `bar`"`
```

Here there are two code blocks, one with `foo has claimed "Foo ` and another which only has `"`.

This is fixed by running the task text through the Markdown parser and stripping all HTML tags from it.

### DIFF
--- a/website/common/locales/en/groups.json
+++ b/website/common/locales/en/groups.json
@@ -321,7 +321,7 @@
   "approvalsTitle": "Tasks Awaiting Approval",
   "upgradeTitle": "Upgrade",
   "blankApprovalsDescription": "When your group completes tasks that need your approval, they'll appear here! Adjust approval requirement settings under task editing.",
-  "userIsClamingTask": "`<%= username %> has claimed \"<%= task %>\"`",
+  "userIsClamingTask": "`<%= username %> has claimed:` <%= task %>",
   "approvalRequested": "Approval Requested",
   "refreshApprovals": "Refresh Approvals",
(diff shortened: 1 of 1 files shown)


Alys:
For tasks such as your `[link](http://example.org/)` example, I think it could be important to still show the URL in the message in group chat, in some way (e.g., as an actual link that you can hover over to see the URL if you don't show the URL as plain text). If there are two or more tasks with similar text but different links, the URLs provide useful information.

Would it be better to use a chat message something like below? I.e., you see the "foo has claimed this task:" part in a code block but then you see the text of the task looking and working exactly the way it looks in the task dashboard.

`foo has claimed:` Some [link](http://example.org/)

aszlig:
@Alys: Okay, this is actually a better idea. I've replaced the commit with one that just updates the `en` localization string for `userIsClamingTask`.

paglias:
Thanks for this @aszlig , if you can provide us with an Habitica user id I'll be happy to award you a contributor tier for this PR

aszlig:
@paglias: Thanks, but I don't have a user.

PR #10350, opened 2018-05-10
redirects: Fix parsing BASE_URL with port number

aszlig:
The parsing in the redirects module was simply determining the base host via trimming off everything up to `//`, so a `BASE_URL` like `http://localhost:3000` will result in the host name `localhost:3000`,
which isn't a valid host name.

So the problem here is that `BASE_URL_HOST` is used for determining whether the client should be redirected and it's comparing the host name of the request object with `BASE_URL_HOST`.

For example if we have the aforementioned `BASE_URL`, we get to the following comparison:

```js
req.hostname !== BASE_URL_HOST
```

Which expands to:

```js
"localhost" !== "localhost:3000"
```

So in order to get rid of the port number, we now use `url.parse()` to get the right host name.

### DIFF
--- a/website/server/middlewares/redirects.js
+++ b/website/server/middlewares/redirects.js
@@ -1,11 +1,11 @@
 import nconf from 'nconf';
+import url from 'url';
 
 const IS_PROD = nconf.get('IS_PROD');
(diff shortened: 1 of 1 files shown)


paglias:

PR #10352, opened 2018-05-11
only store necessary data for social login

phillipthelen:
Previously we stored some data about facebook and google users that is not entirely necessary. This pull request reduces that to the minimum of the login ID and the displayName (which is saved directly as the profile name, since that was the only use for that).

### DIFF
--- a/website/server/controllers/api-v3/auth.js
+++ b/website/server/controllers/api-v3/auth.js
@@ -329,6 +329,7 @@ api.loginSocial = {
         auth: {
           [network]: {
             id: profile.id,
+            email: profile.email,
           },
         },
         profile: {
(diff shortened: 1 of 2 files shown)


paglias:
The build failed for an unrelated timeout, I've restarted it. The code looks good!

TheHollidayInn:
I think we still want to store email from the user

paglias:
Oh yeah, we definitely need to keep the email

paglias:
The code looks good @phillipthelen but we still have some functions using data that has been removed, for example [here](https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/email.js#L33) it should get the name from `user.profile.name` instead that from `auth.facebook`.

Also the key for email addresses can be plural, see https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/email.js#L53

paglias:
Closed in favor of #10395

Issue #10267, opened 2018-04-20
Quest information modal doesn't respond to changing window size


shanaqui:
From a user in the Bug Guild:

> Summary: Parts of quest preview is obstructed from view in smaller window width
> 
> Steps to Reproduce:
> 
> Go to Inventory > Items (or Shops > Quests)
> Locate Quests section (if you have quest items)
> Open a quest and reduce window size (approximately half a laptop screen)
> Notice that the side information for the quest is hidden and cannot be viewed
> Expected Behavior: I should be able to scroll sideways or the quest information modal should be responsive to the changing window width.
> 
> Observed Behavior:
> 
> In shop: https://ibb.co/mKONmn
> In inventory: https://ibb.co/dqNhmn

And:

> Summary: Horizontal and vertical scrollbars on quest preview/information modal window in shop
> 
> Steps to Reproduce:
> 
> Navigate to Shops > Quests
> Select a quest (I don't have all unlocked but I have confirmed that this is reproduced in all of what I can select)
> Expected Behavior: I should see no scrollbar in the information window (similar to the quest information window from the Inventory > Items page's Quests section)
> 
> Observed Behavior: I see a vertical and horizontal scrollbar.

*UPDATE*

The responsive part has been fixed in https://github.com/HabitRPG/habitica/pull/10354

yb50110:
I have a quick idea for this that will prevent overlapping for smaller by fiddling with the CSS. Should I put up a pull request or just provide what I have adjusted as a comment?

yb50110:
I added this to the `modal.scss` file:
```
#start-quest-modal, #buy-quest-modal {
  @media only screen and (max-width: 1200px) {

    .modal-dialog {
      max-width: 33%;

      .left-panel {
        left: initial;
        width: 100%;
        right: 100%;

        .col-4 {
          width: 100px;
        }
      }

      .side-panel, .right-sidebar {
        left: calc(100% - 10px);
        max-width: 100%;
        right: initial;

        .questRewards {
          width: 90%;

          .reward-item {
            width: 100%;
          }
        }
      }
    }
  }
}
```

I couldn't figure out why the CSS/SCSS changes weren't being reflected when I made changed in the `startQuestModal.vue`. And as my two cents, I think in smaller windows, stacking these side panels below the main content would be better (more legible?) than having them side by side.

SabreCat:
@yb50110 Please do submit a PR and we'll take a look!

yb50110:
[...]

PR #10354, opened 2018-05-11
Modal width responsiveness (partial fix for #10267)

yb50110:
Partial fix for #10267
https://github.com/HabitRPG/habitica/issues/10267

### Changes
Added media queries so that the modal window for quests responds to the smaller window sizes. The following is how the modal window looks in with the addition of my PR:
screen shot 2018-05-11 at 18 56 24
screen shot 2018-05-11 at 18 58 49

UUID: 5c9c0406-e564-4f72-bb07-f10c6002ba86


website/client/assets/scss/modal.scss
```diff
@@ -35,5 +35,36 @@
     overflow-y: hidden;
     text-overflow: ellipsis;
   }
-
 }
+
+#start-quest-modal, #buy-quest-modal {
```
paglias, line 40: can this code go directly in the .vue component file used for the modals?

yb50110: When I put those code in startQuestModal.vue file, it doesn't seem to be able to find the styles I created. However, it worked fine in the buyQuestModal.vue. Because they have similar styles, I put their code together and in one spot (in case something needs to be changed later, it can happen at one place). I am not sure why it doesn't work in startQuestModal.vue. I agree that it would be ideal for their specific styles to go directly into their vue files.

paglias:
So I've tested the change and it helps with smaller windows but make the modal unusable on mobile:

schermata 2018-05-17 alle 12 16 45

While this is the current behavior

schermata 2018-05-17 alle 12 16 31

Not perfect but at least you can start the quest.

@Tressley do you have any idea on how to make this design usable on smaller browser windows?

Tressley:
@paglias -- The Rewards panel will need to have its content folded into the main modal.

yb50110:
@Tressley What exactly do you mean by "folded"?

Tressley:
Take the content from the Quest Rewards side panel and place it in the main modal, under the difficulty section.

yb50110:
@paglias I made some changes so it's a bit more responsive! Please take a look at it. 

I also moved the buyQuestModal css code into its vue file. I tried to do the same for startQuestModal but I wasn't able to :/ 

(As a side note, I have to run and test my local install through a VM. I was able to get a copy of habitica working but I failed the second time when I tried to do it the proper way involving git's fork functionality. So I wrote and tested my code in the working version and pasted into the properly forked version... it's really not ideal but that's how I've been working on this so far)

paglias:
Thanks @yb50110 , if we have to split the same css maybe it's better to keep it in a css file so at least we don't have to deal with duplicate code, can you put everything back in the external css file? Sorry for the confusion

yb50110:
@paglias Sorry; I wasn't able to work on it yet. I will get to it by next week!

yb50110:
@paglias  So I tried to put it all in one file but then I am facing issues where I will have to use `!important` since there are already CSS in the buy-quest-modal.vue file for specific elements I need to change. Buy putting the CSS in its file, I avoid using `!important`. Also, in the modal.scss file, although the two CSS seems identical, there have been some changes (ie. using different orders between start-quest-modal and buy-quest-modal). These difference exist because I had to make more changes to make the modals looks better in smaller screens. 

Do you think I should still combine the CSS into one file and use `!important` ?

paglias:
Yeah I think it's better to have everything in one file even if it means using some `!important` rules

yb50110:
@paglias I made some changes! :)

paglias:
@yb50110  thanks for this! I'm merging the PR and awarding your contributor tier.

Issue #10023, opened 2018-02-19
iOS Push Notifications: switch to Provider Authentication Tokens and node-apn module


paglias:
As explained here https://github.com/node-apn/node-apn/blob/master/doc/provider.markdown#provider-certificates-vs-authentication-tokens they provide some beneifts over our current method and we should switch to them.

cc @phillipthelen

fentie:
I think I can handle switching from `push-notify` to `node-apn`, although I'm not sure I've got all the credentials to make the switch from certs to tokens.

Do we have any tests for ios push notifications currently? I see a disabled unit test for APN, and some API tests that are more about the `push-devices` contract

paglias:
@fentie thanks! if you want to make a PR to switch from certificates to tokens that would be great, we'll add the right credentials later. As per the tests if they're disabled it means they weren't working so feel free to change them in your pr

fentie:
So I made a PR for this... my initial reaction would be to include the team ID and key ID in config.json but the wiki makes it pretty clear not to edit that file, so... here we are. Otherwise hardcode those 2 values and upload the key and if the planets align you should (continue to) have iOS push notifications. I don't have an iDevice to test with so sadly I can't verify this myself.

Alys:
> in config.json but the wiki makes it pretty clear not to edit that file,

Ah! That's another thing that can be improved on the wiki. :) It's fine to modify `config.json.example` for the purposes of submitting a legit change in a PR. The wiki is just trying to say that you shouldn't try to modify the example file for your own local purposes (i.e., don't modify it by accident when you're trying to change `config.json`).

fentie:
@Alys I assumed that was the case, but better safe than sorry. In that case I'll make new keys for those values unless you all would prefer them stored elsewhere

PR #10355, opened 2018-05-12
Convert node apn

fentie:
Partial fix of https://github.com/HabitRPG/habitica/issues/10023 (still needs new credentials)

### Changes
Removed `push-notify`, replaced with (updated version of) `node-apn`
Converted to using Provider token for APN config (though no token present in this PR)

----


website/server/libs/pushNotifications.js
```diff
@@ -12,37 +11,27 @@ const FCM_API_KEY = nconf.get('PUSH_CONFIGS:FCM_SERVER_API_KEY');
...
-    }).promise(),
-  ])
-    .then(([certObj, keyObj]) => {
-      let cert = certObj.Body.toString();
-      let key = keyObj.Body.toString();
+  S3.getObject({
+    Bucket: S3_BUCKET,
+    Key: 'apple_apn/APNsAuthKey.p8',
+  }).promise().then((data) => {
+    const key = data.Body.toString();
```
paglias, line 24: how long is this key going to be? We used to store it on S3 because the certificate exceeded the limit for env variables but if it's a shorter string you can add it directly to config.json.example

website/server/libs/pushNotifications.js
```diff
@@ -12,37 +11,27 @@ const FCM_API_KEY = nconf.get('PUSH_CONFIGS:FCM_SERVER_API_KEY');
...
         key,
-        cert,
-      });
-
-      apn.on('error', err => logger.error('APN error', err));
-      apn.on('transmissionError', (errorCode, notification, device) => {
-        logger.error('APN transmissionError', errorCode, notification, device);
-      });
-    });
+        keyId: 'key-id',
```
paglias: where are these keys and team id coming from? they should be in the config?

fentie: I'll move them to config.json

website/server/libs/pushNotifications.js
fentie: I'm not actually seeing anywhere that puts a hard limit on size for these, but this issue at least implies that it should work: https://github.com/node-apn/node-apn/issues/594

paglias: The limit was in our web server

paglias:
@phillipthelen can you get a token from apple? So I'll be able to make the switch from certificates to tokens

SabreCat:
Poking @phillipthelen

paglias:
Going to take over and finish this, thanks @fentie and I'm noting this towards your next contributor tier

PR #10357, opened 2018-05-13
disabling checking off a subtask if not assigned to a user

### LINKED ISSUE #10254: checklist checkboxes should not be clickable in Group Plan's "Task Board" page
If you try to complete a checklist item in the Task Board page in a Group Plan, a tickmark appears in the checkbox and you see the message "Task not found" (see screenshot).

The correct behaviour is that the checklist boxes should not be clickable at all, in the same way that the main checkbox for the entire task is not clickable.

The "Task not found" message should NOT be removed from the backend code - we still need an error to be thrown in case the user tries to complete a checklist item using an API command or a third-party tool.

screenshot_2018-04-15_122237

fentie:
Fixes #10254

### Changes
Disabling task checkboxes if not assigned to a user

----

### DIFF
--- a/website/client/components/tasks/task.vue
+++ b/website/client/components/tasks/task.vue
@@ -64,7 +64,7 @@
               type="checkbox",
               :checked="item.completed",
               @change="toggleChecklistItem(item)",
-              :disabled="castingSpell",
+              :disabled="castingSpell || !isUser",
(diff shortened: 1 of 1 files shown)


paglias:
Thanks, welcome to tier 2!

PR #10358, opened 2018-05-14
Change reward popunder

fentie:
### Changes
As discussed in the blacksmith guild, this makes the popunder for "add multiple X" reflect the column type. Previously it was always "Tasks" which looked confusing in the Rewards column.

----

cc: @Alys


paglias:
thanks! Notated towards your next tier

PR #10362, opened 2018-05-14
extract chat

negue:
the `chat` was inside `tavern` and `group`, to reduce the duplicate code and possible fixes its better to extract the duplicate code to a component, which can be re-used.

@siege918 Here's how it could look like :) 
ref #10226


website/client/components/groups/staffList.js
```diff
@@ -0,0 +1,97 @@
+export default [
```
paglias: can this file be moved to client/libs? in components we should only have .vue files

website/server/models/group.js
```diff
@@ -346,7 +346,13 @@ schema.statics.toJSONCleanChat = async function groupToJSONCleanChat (group, use
 
   // Convert to timestamps because Android expects it
   toJSON.chat.forEach(chat => {
-    chat.timestamp = chat.timestamp ? chat.timestamp.getTime() : new Date().getTime();
+    if (chat.timestamp) {
+      if (chat.timestamp.getTime) {
+        chat.timestamp = chat.timestamp.getTime();
+      }
+    } else {
+      chat.timestamp = new Date().getTime();
```
paglias: is the timestamp undefined sometimes or it simply doesn't have the getTime value? In that case shouldn't we convert the existing timestamp to a date? `chat.timestamp = new Date(chat.timestamp).getTime();
 `

website/client/components/groups/chatTextarea.vue
```diff
@@ -0,0 +1,223 @@
...
+
+      community-guidelines
+
+      slot(
+        name="additionRow",
+      )
+
+      .row
+        .hr.col-12
+        chat-message(:chat.sync='group.chat', :group-id='group._id', :group-name='group.name')
```
paglias: So this isn't just the text area but also show the list of messages? in that case I'd rename the component name

website/client/components/groups/chatTextarea.vue
```diff
@@ -0,0 +1,223 @@
...
+    props: ['label', 'group', 'placeholder'],
+    components: {
+      autocomplete,
+      communityGuidelines,
+      chatMessage,
+    },
+    data () {
+      return {
+        newMessage: '',
+        icons: Object.freeze({}),
```
paglias: this can be removed

paglias (changes requested):
a few comments but overall this is a great change!

website/server/models/group.js
negue: `timestamp` was a number on some of my chat messages instead of a `Date` thats why it hadn't any methods on it

website/client/components/groups/chatTextarea.vue
negue: yeah I didn't think we'd need add chat-message as a slot on both uses, hmm any name suggestions ? `chat` since its both the new message and the chatlist itself

website/server/models/group.js
paglias: Ok, but you're creating a new date with a value of now, `chat.timestamp = new Date(chat.timestamo).getTime();` will convert it to the right date

website/client/components/groups/chatTextarea.vue
paglias: chat sounds good to me

website/server/models/group.js
negue: I just took the old code and added a check for `getTime` I guessed it was correct before.

The current code checks if there even is a value in timestamp if not just set a new one

paglias: I see, if the number you're talking about is a numeric timestamp then I think there's a bug in the current code because we should convert the timestamp to a date non create a new date with a value of now

negue: if `chat.timestamp` is `undefined` then we can't just convert the `undefined` to a date^^

The old chats have a timestamp as number, the new ones use date , so either way there won't a message which is null or undefined, so I'll refactor my changes to remove the unneeded `new Date().getTime()`  part

paglias:
Seems to be working well, thanks!

PR #10364, opened 2018-05-15
Avoid using media element with empty src attribute

aszlig:
Whenever the client starts up, the following is emitted in the Firefox console:

```
Invalid URI. Load of media resource  failed.
All candidate resources failed to load. Media load paused.
```

This happens because the `<source/>` tags are preinitialized with a `src` attribute of `""`.

So what we're doing instead is initialize the `<audio/>` element without any children and add the children as soon as the first audio file needs to be played. This also has the advantage that we can determine at runtime whether the browser supports Ogg/Vorbis or whether we should fall back to MPEG layer 3 so only one source element is needed.

### DIFF
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -44,8 +44,6 @@ div
             router-view
         app-footer
         audio#sound(autoplay, ref="sound")
-          source#oggSource(type="audio/ogg", :src="sound.oggSource")
-          source#mp3Source(type="audio/mp3", :src="sound.mp3Source")
 </template>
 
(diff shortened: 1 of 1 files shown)


aszlig:
Hm, the Travis errors seem to be unrelated to this change, here are the test results on top of 4.42.3: https://headcounter.org/hydra/eval/399917?filter=tests&compare=399868

negue:
Thanks this is better! 👍

paglias:

PR #10367, opened 2018-05-15
tests: Increase timeouts instead of disabling them

aszlig:
Some tests were disabled in ba799c67f9f31e94bd3cc97975d1b5768b45817c and 10567d81e2efa5256242d396116c1973dfb170fc, because they tend to frequently time out after 8 seconds.

Instead of disabling the tests (which IMHO is bad, because tests are there for a reason), we're now increasing the timeout to 30 seconds just for these tests. It would probably be a good idea to somehow add an ESLint warning, but I haven't found a rule so far that is very suitable for adding a warning that contains a portion of a comment.

So for now I've marked it with an `XXX` and added the `no-warning-comments` ESLint rule to the test suite.

I also needed to change the arrow notation for the test cases to use the function keyword, because otherwise we don't have `this.timeout()` available.

Cc: @paglias

### DIFF
--- a/test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js
+++ b/test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js
@@ -123,8 +123,8 @@ describe('GET /challenges/:challengeId/members', () => {
     });
   });
 
-  // @TODO times out too many times (when it takes more than 8s)
-  xit('supports using req.query.lastId to get more members', async () => {
+  it('supports using req.query.lastId to get more members', async function () {
+    this.timeout(30000); // @TODO: times out after 8 seconds
     let group = await generateGroup(user, {type: 'party', name: generateUUID()});
     let challenge = await generateChallenge(user, group);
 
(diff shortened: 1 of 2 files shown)


test/.eslintrc
```diff
@@ -8,5 +8,8 @@
   },
   "globals": {
     "_": true,
+  },
+  "rules": {
```
paglias: What's this new rule for?

aszlig: As mentioned in the commit message, this is so that ESLint displays a warning whenever a `XXX`, `TODO` or `FIXME` comment occurs within the test suite.

paglias: Ah okay, in that case we've been using `// @TODO` for this purpose but would prefer not to have this show up in the warning, can you convert the comment to @TODO and remove the rule? 

When refactoring we'll go through the @TODO comments anyway

I'll merge after this change, thanks!

aszlig:
@paglias: Commit amended.

paglias:
Thanks @aszlig , it'll get merged before the end of the week

paglias:

PR #10368, opened 2018-05-16
Remove Ethereal Surge notification

### LINKED ISSUE #10313: "Your Skill backfires" message for Mage's Ethereal Surge skill is misleading
As described in https://github.com/HabitRPG/habitica/pull/10121, Ethereal Surge does not give Mana to Mages, and when you cast the skill, if your Party contains other Mages, you see this message in a snackbar notification:
`"Your Skill backfires when mixed with another's magic. Only non-Mages gain MP."`
(That's the `spellWizardNoEthOnMage` string in spells.json.)

We tried to write that message carefully enough that it wouldn't be interpretted as an error condition or user mistake but unfortunately it is being misinterpretted. I've seen several questions about it, such as this one:
_"I'm confused: almost every time I cast Ethereal Surge, I get a pop-up that says: 'Your skill backfires when mixed with another's magic. Only non-mages gain MP.' What's confusing is 1) that this happens even when it has been more than 24 hours since anyone has cast Ethereal Surge for my party; and 2) I thought buffs 'stacked,' so aren't the non-mages in my party getting a mana-buff from my Ethereal Surge, plus their mana-buff from my party's other mage's Ethereal Surge, plus their mana-buff from my party's other mage's Ethereal Surge? (And if indeed they are, how is this 'back-firing'?) Thanks."_

It can also be a bit of a downer: _"This is so absurd but it's actually kind of demoralizing to see that my Ethereal Surge skill backfires every time I use it.  It feels like failure and makes the skill seem less worthwhile"_

So I think we should remove the message. It seemed great for gamification but not so good in practice alas.

Do we want to replace the message with something more mundane or just not show a notification at all? I'm leaning towards not showing a notification. The description of the skill says that it does not affect Mages so I don't know if an extra notification every time is necessary. No matter how we word it, some users might still think an error has occurred and it would likely always have a demotivating feeling.

Yutsuten:

Fixes #10313

### Changes
Removes the pop up notification that appears when a mage uses Ethereal Surge in a party with other mages.

----


paglias:
thanks! Notated towards your next tier

PR #10370, opened 2018-05-17
added try catch, added snackbar notification for errors returned

### LINKED ISSUE #10217: "Can't feed this pet" error message should be visible to user not shown in the console

jerell-mendoza:

[//]: #10217 
Fixes #10217

### Changes
[//]: # Added a try catch block around the feed pets command, on catch we emit a an event to create a snackbar/growl message

[//]: #4b9dc32d-ad7d-4d2e-aaad-0543472f87f6

----
UUID: 4b9dc32d-ad7d-4d2e-aaad-0543472f87f6

### DIFF
--- a/website/client/components/inventory/stable/index.vue
+++ b/website/client/components/inventory/stable/index.vue
@@ -935,7 +935,20 @@
         }
       },
       async feedAction (petKey, foodKey) {
-        const result = await this.$store.dispatch('common:feed', {pet: petKey, food: foodKey});
+        let result;
+        try {
+          result = await this.$store.dispatch('common:feed', {pet: petKey, food: foodKey});
+        } catch (e) {
+          const errorMessage = e.message || e;
+
+          this.$store.dispatch('snackbars:add', {
+            title: 'Habitica',
+            text: errorMessage,
+            type: 'error',
+            timeout: true,
+          });
+        }
+
         if (result.message) this.text(result.message);
         if (this.user.preferences.suppressModals.raisePet) return;
         if (this.user.items.pets[petKey] === -1) this.$root.$emit('habitica::mount-raised', petKey);
--- a/website/client/components/inventory/stable/index.vue
+++ b/website/client/components/inventory/stable/index.vue
@@ -935,9 +935,12 @@
         }
       },
       async feedAction (petKey, foodKey) {
-        let result;
         try {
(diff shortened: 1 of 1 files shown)


website/client/components/inventory/stable/index.vue
```diff
@@ -935,7 +935,20 @@
...
+          const errorMessage = e.message || e;
+
+          this.$store.dispatch('snackbars:add', {
+            title: 'Habitica',
+            text: errorMessage,
+            type: 'error',
+            timeout: true,
+          });
+        }
+
```
paglias: I think these 3 lines should be inside the try block as well, or you should return from the catch block because otherwise when an error occurs `result` will be undefined and accessing `result.message` will throw another error

paglias (changes requested):
just one thing to change!

website/client/components/inventory/stable/index.vue
jerell-mendoza: @paglias , gotcha! good catch, I didn't think about that case. Updated

jerell-mendoza:
@paglias , it's been updated! Thanks for the review!

paglias:
thanks @jerellmendoodoo ! If you have an Habitica UUID I'll be happy to award a contributor tier for the PR

jerell-mendoza:
@paglias woohoo thanks for merging it in :) and by UUID, if that the same as my user id then its: 4b9dc32d-ad7d-4d2e-aaad-0543472f87f6. Also should I delete this branch now?

paglias:
@jerellmendoodoo notated towards your second tier! for the branch you can delete it or keep it, it's on your fork so your choice :) But if you want to delete it, that's totally fine as it's been merged now

PR #10376, opened 2018-05-18
client: Fix display of class bonus for other users

aszlig:
Whenever one is hovering an item from another user, the bonuses of these items are shown for the own user. So for example if you're a mage and view a Royal Magus Robe of another mage, the class bonus is 6.

However if you're a warrior, the class bonus displays as 0 because the attributes grid is always using the stats for the own user even if viewing equipment of a different user.

I've fixed this by moving the user object to the properties in `attributesGrid` and passing the current user from every other Vue file that's using `attributesGrid`.

Not sure whether this is the right approach, as I'm no expert in Vue.js but some testing with the client now shows the correct values.

### DIFF
--- a/website/client/components/inventory/equipment/attributesGrid.vue
+++ b/website/client/components/inventory/equipment/attributesGrid.vue
@@ -147,11 +147,13 @@
       item: {
         type: Object,
       },
+      user: {
+        type: Object,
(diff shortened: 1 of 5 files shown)


paglias:
thanks!

PR #10377, opened 2018-05-19
removing the 24px top margin on .modal-dialog .title

fentie:
### Changes
Boss HP and Difficulty on the quest details view are not aligned with their values. Removing the top margin so they line back up again, as discussed in the Aspiring Blacksmiths guild

----

cc: @Alys

### DIFF
--- a/website/client/components/shops/quests/questInfo.vue
+++ b/website/client/components/shops/quests/questInfo.vue
@@ -24,7 +24,7 @@
 
 dt {
   font-family: 'Roboto Condensed';
-  font-size: 20px;
+  font-size: 1.3em;
   line-height: 1.2;
   color: $gray-50;
 }
@@ -47,6 +47,17 @@ dt {
 
   .svg-icon {
     margin-top: 1px;
+  }
+}
+</style>
+
+<style lang="scss">
+.questPopover {
+  dt {
+    font-family: 'Roboto', sans-serif;
+    color: inherit;
+    font-size: 1em;
+    white-space: nowrap;
   }
 }
 </style>
--- a/website/client/components/shops/quests/questInfo.vue
+++ b/website/client/components/shops/quests/questInfo.vue
@@ -25,6 +25,7 @@
   text-align: left;
   font-weight: bold;
   white-space: nowrap;
+  margin-top: 0;
 }
 
 .col-8 {
--- a/website/client/components/shops/quests/questInfo.vue
+++ b/website/client/components/shops/quests/questInfo.vue
@@ -20,11 +20,13 @@
 </template>
 
 <style lang="scss" scoped>
+@import '~client/assets/scss/colors.scss';
+
 dt {
   font-family: 'Roboto Condensed';
   font-size: 20px;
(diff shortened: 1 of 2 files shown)


paglias:
Thanks! Can you post a screenshot of the changes?

fentie:
Sure:

screen shot 2018-05-25 at 8 29 00 am

negue:
your rule is overridden by `.modal-dialog .title ` , if you add `!important; ` then it will work on every modal

fentie:
@negue I'm not sure I follow you. Is this a problem on other modals too? This is the only one I noticed.

negue:
well the `margin-top` rule was overridden on the shop detail modal, also on the inventory items detail modal, so with `!important; ` it really worked everywhere 😄

fentie:
@negue so I looked at this again and I realized part of the problem/confusion was we were using `.title` in a place where it didn't belong... it has no semantic meaning here, it's just a convenient class with a set of styles that (mostly?) match what the implementer was after.

I redid that section with a `<dl>` instead and gave it only the styles it needed.

negue:
Could you add a `smallVersion` style, `questInfo.vue` is also used in the popovers of a quest, now it shows the big labels in that popover

website/client/components/shops/quests/questInfo.vue
```diff
@@ -1,30 +1,30 @@
...
-
-.title {
-  text-align: left;
-  font-weight: bold;
-  white-space: nowrap;
+dt {
+  font-family: 'Roboto Condensed';
+  font-size: 20px;
+  line-height: 1.2;
+  color: #4E4A57;
```
negue: its better to use the colors from `colors.scss` , for this it would be `$gray-50`

fentie:
@negue should be back in business now. busy few days!

website/client/components/shops/quests/questInfo.vue
```diff
@@ -49,6 +51,17 @@
 }
 </style>
 
+<style lang="scss">
+.questPopover {
+  dt {
+    font-family: 'Roboto', sans-serif;
```
paglias: this font-family rule can be removed as it's specified at an higher level

paglias (changes requested):

website/client/components/shops/quests/questInfo.vue
```diff
@@ -21,10 +22,11 @@
 <style lang="scss" scoped>
 @import '~client/assets/scss/colors.scss';
 
-.title {
-  text-align: left;
-  font-weight: bold;
-  white-space: nowrap;
+dt {
+  font-family: 'Roboto Condensed';
```
paglias: same here

paglias (changes requested):

fentie:
@paglias updated, rebased, and pushed

paglias:
this is looking good @fentie, thanks! We're going to merge it at the end of this week or at the beginning of the next one

paglias:
thanks @fentie ! notated towards your next tier

PR #10383, opened 2018-05-21
Fix 9248: challenge creator should not automatically join their own challenge

### LINKED ISSUE #9248: challenge creator should not automatically join their own challenge
Currently, when you create a challenge, you are automatically made a participant.

This should be changed. A challenge creator should not automatically join their own challenge. They should become a participant only when/if they choose to click the "Join" button.

----

_Here are the reasons (just FYI; this is not important for implementing this change):_
- _If you don't yet have the achievement for joining a challenge and you create a challenge in the new website, the achievement pops up in the middle of the challenge creation flow (after filling out the challenge details but before you add tasks), which is disruptive and confusing. You could argue that a person shouldn't be creating challenges until they joined one, but some users might._
- _Users who've created a challenge sometimes ask if it's okay for them to leave the challenge. They worry that it will end or damage the challenge. If they weren't automatically a participant, it would be clear they could leave._
- _There's probably quite a few cases where challenge leaders don't want to be a participant so forcing them to join and then leave is just extra work for them (e.g., they have to remove their copy of the challenge's tag). For the cases where the leader does want to be a participant, they can easily do that by just clicking "join"._
- _Leaders will no longer have excess tasks in their own list if go through a few iterations of adding / deleting challenge tasks before the challenge is ready for use._

paglias:
Fixes #9248

### DIFF
--- a/test/api/v3/integration/challenges/DELETE-challenges_challengeId.test.js
+++ b/test/api/v3/integration/challenges/DELETE-challenges_challengeId.test.js
@@ -41,6 +41,7 @@ describe('DELETE /challenges/:challengeId', () => {
       group = populatedGroup.group;
 
       challenge = await generateChallenge(groupLeader, group);
+      await groupLeader.post(`/challenges/${challenge._id}/join`);
 
       await groupLeader.post(`/tasks/challenge/${challenge._id}`, [
         {type: 'habit', text: taskText},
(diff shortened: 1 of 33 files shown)


paglias:
Looks like this is breaking a ton of tests that relied on the creator joining the challenge automatically

PR #10385, opened 2018-05-22
Added initial user cache

TheHollidayInn:


paglias:
If the user is updated with a .update command the cache won't be updated?

TheHollidayInn:
Will come back to this in v4

PR #10387, opened 2018-05-24
Api quest restrictions - no purchase/start without fulfilling eligibility requirements

### LINKED ISSUE #8530: API allows quests to be purchased and started without fulfilling eligibility requirements
In the Report a Bug guild, Birkwelch writes: 
"I previously mentioned a minor bug in the Android app about being able to start a quest without having completed the requirements; apparently you can also purchase any quest without fulfilling eligibility requirements, i.e. you don't have to be level 60 to purchase Recidivate, or have completed the first two parts to obtain part three."

Even though that can be done on the Android app (and possibly also iOS), the root cause is that the API doesn't enforce the eligibility requirements.  NOTE THOUGH: I have not confirmed that myself. If someone would like to do a test and report the results here, that would be great!

Assuming that's correct, the API should be changed to:
- make it impossible to buy a quest if you aren't eligible to do so
- make it impossible to start a quest if you aren't eligible to do so
- return an informative error message for both those cases
- have the error message documented in the API in an `@apiError` comment
- have tests for the new behaviour

fentie:
Fixes #8530

### Changes
Extracted a "check prerequisites" method for quest purchases that looks for the requested quest's `previous` attribute and compares it with the requesting user's `achievements.quests` list. I also moved the masterclasser check in here too because it seemed like a logical grouping

----

### DIFF
--- a/website/common/script/ops/buy/buyQuest.js
+++ b/website/common/script/ops/buy/buyQuest.js
@@ -57,7 +57,7 @@ export class BuyQuestWithGoldOperation extends AbstractGoldItemOperation {
     }
 
     if (item && item.previous && !user.achievements.quests[item.previous]) {
-      throw new NotAuthorized(this.i18n('mustComplete', item.previous));
+      throw new NotAuthorized(this.i18n('mustComplete', {quest: item.previous}));
     }
   }
 
--- a/website/common/script/ops/buy/buyQuest.js
+++ b/website/common/script/ops/buy/buyQuest.js
@@ -18,7 +18,7 @@ export class BuyQuestWithGoldOperation extends AbstractGoldItemOperation {
     return true;
   }
 
-  userAbleToStartMasterClasser (user) {
+  static userAbleToStartMasterClasser (user) {
     return user.achievements.quests.dilatoryDistress3 &&
       user.achievements.quests.mayhemMistiflying3 &&
       user.achievements.quests.stoikalmCalamity3 &&
@@ -37,10 +37,6 @@ export class BuyQuestWithGoldOperation extends AbstractGoldItemOperation {
     let key = this.key = get(req, 'params.key');
     if (!key) throw new BadRequest(errorMessage('missingKeyParam'));
 
-    if (key === 'lostMasterclasser1' && !this.userAbleToStartMasterClasser(user)) {
-      throw new NotAuthorized(this.i18n('questUnlockLostMasterclasser'));
-    }
-
     let item = content.quests[key];
 
     if (!item) throw new NotFound(errorMessage('questNotFound', {key}));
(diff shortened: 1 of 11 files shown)


website/server/controllers/api-v3/quests.js
```diff
@@ -76,6 +76,7 @@ api.inviteToQuest = {
     if (!user.items.quests[questKey]) throw new NotAuthorized(res.t('questNotOwned'));
     if (user.stats.lvl < quest.lvl) throw new NotAuthorized(res.t('questLevelTooHigh', { level: quest.lvl }));
     if (group.quest.key) throw new NotAuthorized(res.t('questAlreadyUnderway'));
+    if (quest.previous && !user.achievements.quests[quest.previous]) throw new NotAuthorized(res.t('mustComplete', {quest: quest.previous}));
```
paglias: Is this necessary? I mean the check is already done when the quest is bought so this should not be necessary

website/common/script/ops/buy/buyQuest.js
```diff
@@ -49,9 +45,22 @@ export class BuyQuestWithGoldOperation extends AbstractGoldItemOperation {
       throw new NotAuthorized(this.i18n('questNotGoldPurchasable', {key}));
     }
 
+    this.checkPrerequisites(user, key);
+
     this.canUserPurchase(user, item);
   }
 
+  checkPrerequisites (user, questKey) {
+    const item = content.quests[questKey];
```
paglias, line 54: can you pass the item directly instead of passing the key and then refetching the item?

paglias (commented):
Left two small comments

website/server/controllers/api-v3/quests.js
fentie: I thought that too but without it we'd be left with a (potentially very small) hole where people who have bought quests that they shouldn't be able to would be able to start them before they were supposed to. I'm fine with removing it, but the issue said to change it as well, so I put it in

website/common/script/ops/buy/buyQuest.js
fentie: I can but I'd have to either also pass the `questKey` param (for the check on line 55) or pull that code back out into a separate method. If I only passed the item, I'd no longer have access to the untranslated quest name.

Which do you prefer... pass item & key, leave it as is, or pull the masterclasser check back out?

website/server/controllers/api-v3/quests.js
paglias: Opinions @Alys ? i feel like it would be better to leave the loophole and impose the restrictions only for newly bought quests

website/common/script/ops/buy/buyQuest.js
paglias: Ah okay, that makes sense, let's keep it this way

website/server/controllers/api-v3/quests.js
Alys: @paglias If you'd prefer it that way, that's fine with me. @fentie is right that's it's a small loophole - not very important.

fentie: Ok, removed and pushed.

paglias:
thanks! notated towards your next tier

PR #10418, opened 2018-06-02
tests/cron: Fix tests that involve mocked time

aszlig:
Ah, the joys of global state... :unamused:
    
Whenever some test failed which has mocked the time using `useFakeTimers()`, other test that are run after that test would fail (or even time out) as well, which is a bit confusing to debug.
    
Some of the tests even had a cleanup routine in `afterEach()` but most of them didn't, so I rearranged them in a way so that we have a clock variable for **all** of the subtests, which initially is null and then a cleanup handler (also for **all** of the subtest) calls `clock.restore()` if the value isn't `null`.
    
In order to avoid calling `clock.restore()` twice, I have removed all the `clock.restore()` calls at the end of the tests that set the clock to a specific value.

With the chain reaction gone, the test case that really failed became more clear (`for a 3-month gift subscription (non-recurring)`), so I fixed the test as well.

Cc: @Alys

### DIFF
--- a/test/api/v3/unit/libs/cron.test.js
+++ b/test/api/v3/unit/libs/cron.test.js
@@ -475,7 +475,7 @@ describe('cron', () => {
       // user3g has a 3-month gift subscription starting today
       user3g.purchased.plan.customerId = 'Gift';
       user3g.purchased.plan.dateUpdated = moment().toDate();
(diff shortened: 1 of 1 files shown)


paglias:
Thanks for this @aszlig , you've been extremely helpful with these tests PRs :)

PR #10420, opened 2018-06-02
Flag / Report Private Messages (PMs)

### LINKED ISSUE #9204: private messages (PMs) in inbox can't be flagged / reported
When a player receives an inappropriate private message, they need to be able to flag it to alert moderators in the same way they can flag a message in a guild.

_edit 2018-04-18: Code for some (perhaps most) of the features below exists from work done in https://github.com/HabitRPG/habitica/pull/9761 . Any contributor working on this should investigate the existing code before creating new code. The "Report" link on the PMs doesn't exist so your first action might be to add / enable that (check to see if it's in the vue code but commented out) - that might then help you test for the rest of the features._

Features required:
- [ ] Each PM should have a "Report" link and the same icon that you can see next to "Report" in guild messages. 
   - [ ] That **includes** PMs sent by the player themself - they must be able to flag their own PMs (e.g., when needing to report both sides of a conversation to moderators).
- [ ] Clicking on the Report link will give the same confirmation screen that you see when you flag a guild message.
- [ ] After any PM is flagged, it should contain a permanent notice something like "You have reported this message to moderators." 
   - [ ] That permanent notice should appear as soon as the flag has been **successfully** saved to the database and the report to the mods has been **successfully** sent. 
   - [ ] It should not appear if those actions have failed.
   - [ ] If those actions fail, the user should see an error message (e.g., a red snackbar notification, that disappears after a few seconds) so that they know they need to try to flag the PM again.
   - [ ] It should not be necessary to reload or sync the website to see the permanent notice.
- [ ] A flagged PM should remain visible (if the player doesn't want to see it, they can delete it).
- [ ] When a PM is flagged, the same actions should occur to report it to mods as when a guild message is flagged: 
   - [ ] an email should be sent to the same recipients as for guild messages
   - [ ] a notification should be sent to the moderator's slack channel. For reporting posts in guilds/parties/Tavern, that notification contains the phrase "flagged a message", but for PMs, the notification should contain the phrase "flagged a PM". The "flagged a message" text for guilds/parties/Tavern should be changed to "flagged a group message" to make it clearer to mods that it's not a PM.
   - [ ] the notifications in the slack channel need to make it clear who posted the message - see below for details
- It is not necessary to allow a player to unflag a PM. If there's demand for that, we can add it later.

----

_The material below has been taken from feedback about https://github.com/HabitRPG/habitica/pull/9761. It's probably still relevant for this issue. Please ask if any of this isn't clear!_

The notifications in the slack channel need to make it clear who posted the message. The screenshot below shows the **same** message flagged twice - once by the person who wrote the message and once by the person who received it. When the person who posted it flagged it ("Black Goblin, Go!"), it looks like the message had been written by the recipient ("ALittleYellowSpider"). Below the screenshot is the matching data from the two player's inbox objects.

img

**From the inbox of the recipient:**
```
"cc710462-4a08-4a45-9ed5-985dba9e08b4": {
	"id": "cc710462-4a08-4a45-9ed5-985dba9e08b4",
	"text": "Alys testing PM flagging:  Black Goblin, Go! is writing this message to ALIttleYellowSpider.",
	"timestamp": 1515496466366,
	"likes": {},
	"flags": {
		"3e595299-3d8a-4a10-bfe0-88f555e4aa0c": true
	},
	"flagCount": 1,
	"uuid": "98089587-1bbe-4de2-92d3-0bbe76513a3f", // *********** this is the person who sent the message ***********
	"contributor": {
		"admin": false
	},
	"backer": {},
	"user": "Black Goblin, Go!", // *********** this is the person who sent the message ***********
	"sort": 183
}
```

**From the inbox of the person who sent the message:**
```
"bbf3e017-685d-4820-9e95-d6837681ed15": {
	"sent": true,   // *********** this field does not exist in the recipient's copy of the message so this is how you can tell whether it's being flagged by the recipient or the sender ***********
	"id": "bbf3e017-685d-4820-9e95-d6837681ed15",
	"text": "Alys testing PM flagging:  Black Goblin, Go! is writing this message to ALIttleYellowSpider.",
	"timestamp": 1515496466371,
	"likes": {},
	"flags": {
		"98089587-1bbe-4de2-92d3-0bbe76513a3f": true
	},
	"flagCount": 1,
	"uuid": "3e595299-3d8a-4a10-bfe0-88f555e4aa0c", // *********** this is the recipient ***********
	"contributor": {
		"level": 7,
		"text": "Test",
		"contributions": "This is Alys's second account.",
		"admin": false
	},
	"backer": {},
	"user": "ALittleYellowSpider", // *********** this is the recipient ***********
	"sort": 34
}
```

negue:
- [x] Each PM should have a "Report" link and the same icon that you can see next to "Report" in guild messages. 
   - [x] That **includes** PMs sent by the player themself - they must be able to flag their own PMs (e.g., when needing to report both sides of a conversation to moderators).
- [x] Clicking on the Report link will give the same confirmation screen that you see when you flag a guild message.  
- [x] After any PM is flagged, it should contain a permanent notice something like "You have reported this message to moderators." 
   - [x] That permanent notice should appear as soon as the flag has been **successfully** saved to the database and the report to the mods has been **successfully** sent. 
   - [x] It should not appear if those actions have failed.
   - [x] If those actions fail, the user should see an error message (e.g., a red snackbar notification, that disappears after a few seconds) so that they know they need to try to flag the PM again.
   - [x] It should not be necessary to reload or sync the website to see the permanent notice.
- [x] A flagged PM should remain visible (if the player doesn't want to see it, they can delete it).
- [x] When a PM is flagged, the same actions should occur to report it to mods as when a guild message is flagged: 
   - [x] an email should be sent to the same recipients as for guild messages
   - [x] a notification should be sent to the moderator's slack channel. For reporting posts in guilds/parties/Tavern, that notification contains the phrase "flagged a message", but for PMs, the notification should contain the phrase "flagged a PM". The "flagged a message" text for guilds/parties/Tavern should be changed to "flagged a group message" to make it clearer to mods that it's not a PM.
   - [x] the notifications in the slack channel need to make it clear who posted the message - see below for details in #9204 

Also TODO:
- [x] tests
- [x] update flag-o-rama for private messages

fixes #9204


Alys:
> Clicking on the Report link will give the same confirmation screen that you see when you flag a guild message. @Alys I assume you mean the notification message?

It's not a notification in Habitica's usual sense of the word.
When you click the report link on a guild message, you see a modal as in the screenshot below. You should see something similar when you click report on a PM.

image

negue:
@Alys ah ok, I thought by "confirmation" you meant something after this modal dialog^^, then is already added :)

Also if a private message is reported, should both users see the permanent notice? or just the one that reported it ?

Alys:
Only the one who reported it.

negue:
message_reported

@Tressley any thoughts on the position / color of `You have reported this message to moderators.`

Tressley:
The red copy is a bit far from the actual message. Can we bring that closer to ~16-24px and use our Habitica red #F74E52.

paglias:
is this ready for review or needs extra work?

negue:
Its ready for review

paglias:
I think for this it might be better to wait for #10428 since it'll require some changes

paglias:
@negue finally merged the inbox Pr I think this one can be merged now if the conflicts are fixed

website/client/components/chat/chatCard.vue
```diff
@@ -13,29 +13,32 @@ div
...
-          | {{$t('copyAsTodo')}}
-        span.action(v-if='!inbox && user.flags.communityGuidelinesAccepted && msg.uuid !== "system"', @click='report(msg)')
-          .svg-icon(v-html="icons.report")
-          | {{$t('report')}}
+          .svg-icon(v-html="icons.copy", v-once)
+          span(v-once)  {{$t('copyAsTodo')}}
+        span.action(v-if='(inbox || (user.flags.communityGuidelinesAccepted && msg.uuid !== "system")) && !isMessageReported', @click='report(msg)')
+          .svg-icon(v-html="icons.report", v-once)
+          span(v-once) {{$t('report')}}
           // @TODO make flagging/reporting work in the inbox. NOTE: it must work even if the communityGuidelines are not accepted and it MUST work for messages that you have SENT as well as received. -- Alys
```
Alys: This comment can be removed.

website/client/components/chat/chatCard.vue
```diff
@@ -227,6 +237,9 @@ export default {
       const message = this.msg;
       return achievementsLib.getContribText(message.contributor, message.backer) || '';
     },
+    isMessageReported () {
+      return this.msg.reported || this.reported;
```
Alys: I'm probably misunderstanding the code, but does this return `true` for messages that have been reported in previous sessions? I.e., for messages that have a flag from the user already recorded in the data?  I think it should return true in that case (to prevent them reporting it again and to put the `reportedMessage` text on the message)  but I can't see how that would happen.

website/client/store/actions/chat.js
```diff
@@ -69,10 +69,18 @@ export async function like (store, payload) {
 }
 
 export async function flag (store, payload) {
-  const url = `/api/v4/groups/${payload.groupId}/chat/${payload.chatId}/flag`;
+  let url = '';
+
+  if (payload.groupId === 'privateMessage') {
+    url = `/api/v4/members/flag-private-message/${payload.chatId}`;
```
Alys, line 54: I'm wondering if this should in in api/v3, if that's possible, to allow existing third-party tools to use it.

If it has to be v4, then this comment in the API's code should be changed:
`/api/v3/members/flag-private-message/:messageId Flag a private message`
and it's probably worth adding words to the comments there that it's only for v4 to make it more obvious.

website/server/controllers/api-v3/members.js
```diff
@@ -634,6 +635,43 @@ api.sendPrivateMessage = {
   },
 };
 
+
+/**
+ * @api {post} /api/v3/members/flag-private-message/:messageId Flag a private message
+ * @apiDescription A message will be hidden immediately if a moderator flags the message. An email is sent to the moderators about every flagged message.
```
Alys: This part of that comment should be removed:
"A message will be hidden immediately if a moderator flags the message."

website/server/controllers/api-v3/members.js
```diff
@@ -634,6 +635,43 @@ api.sendPrivateMessage = {
...
+ * @apiName FlagPrivateMessage
+ * @apiGroup Member
+ *
+ * @apiParam (Path) {UUID} messageId The private message id
+ *
+ * @apiSuccess {Object} data The flagged chat message
```
Alys: Change "chat" to "private" maybe?

website/server/controllers/api-v3/members.js
```diff
@@ -634,6 +635,43 @@ api.sendPrivateMessage = {
...
+ * @apiSuccess {UUID} data.id The id of the message
+ * @apiSuccess {String} data.text The text of the message
+ * @apiSuccess {Number} data.timestamp The timestamp of the message in milliseconds
+ * @apiSuccess {Object} data.likes The likes of the message
```
Alys: It might be worth changing this to "The likes of the message (always an empty object)" to avoid questions about whether private messages should be able to have likes. }

That's a minor point though. :) I only thought about it because I thought first "hey wait, likes aren't possible!" and then I checked the code to see that the object will be there but always empty.

website/server/controllers/api-v3/members.js
```diff
@@ -634,6 +635,43 @@ api.sendPrivateMessage = {
...
+ * @apiSuccess {Object} data.flags The flags of the message
+ * @apiSuccess {Number} data.flagCount The number of flags the message has
+ * @apiSuccess {UUID} data.uuid The user id of the author of the message
+ * @apiSuccess {String} data.user The username of the author of the message
```
Alys: The above two lines should read something like this:
```
 * @apiSuccess {UUID} data.uuid The user id of the author of the message, or of the recipient if `sent` is true
 * @apiSuccess {String} data.user The username of the author of the message, or of the recipient if `sent` is true
```

website/server/libs/chatReporting/inboxChatReporter.js
```diff
@@ -0,0 +1,114 @@
...
+    return {message, userComment};
+  }
+
+  async notify (message, userComment) {
+    const group = {
+      type: 'private messages',
+    };
+
+    await super.notify(group, message);
+
```
Alys: I haven't read through this code in detail so this might be irrelevant but are there any problems with the references below to the "group" name, URL, and _id since the inbox won't have them?

If default values need to be supplied in the email, then I think "N/A" would work for all three.

website/server/libs/chatReporting/inboxChatReporter.js
```diff
@@ -0,0 +1,114 @@
...
+    if (!message.flags) message.flags = {};
+    // TODO fix error type
+    if (message.flags[this.user._id] && !this.user.contributor.admin) {
+      throw new NotFound(this.res.t('messageGroupChatFlagAlreadyReported'));
+    }
+
+    return this.updateMessageAndSave(message, (m) => {
+      m.flags[this.user._id] = true;
+
+      // Log total number of flags (publicly viewable)
```
Alys: This comment shouldn't say " (publicly viewable)" since the messages are private.

For the code below, I don't think there's any need for a different flagCount for admins compared to other players but of course I'm fine with this being left as it is for consistency with group messages if that's desirable. :)  See also my comment below about how mods won't be flagging a private message (unless it appears in their own inbox of course).

website/server/libs/slack.js
```diff
@@ -77,14 +89,75 @@ function sendFlagNotification ({
...
+  userComment,
+}) {
+  if (SKIP_FLAG_METHODS) {
+    return;
+  }
+  let titleLink;
+  let authorName;
+  let title = `Flag in ${flagger.profile.name}'s Inbox`;
+  let text = `${flagger.profile.name} (${flagger.id}; language: ${flagger.preferences.language}) flagged a PM`;
+  let footer = `<${SLACK_FLAGGING_FOOTER_LINK}?userId=${flagger.id}&chatMessageId=${message.id}|Flag this message>`;
```
Alys: This footer link isn't needed for private messages because mods won't be flagging them  (the only reason for a mod to flag a group message is to hide it from all non-mods and that doesn't apply to private messages - if the user wants to not see the message, they can delete it after flagging it).

footer can be set to an empty string.

website/server/libs/slack.js
```diff
@@ -9,6 +9,10 @@ const SLACK_FLAGGING_URL = nconf.get('SLACK:FLAGGING_URL');
 const SLACK_FLAGGING_FOOTER_LINK = nconf.get('SLACK:FLAGGING_FOOTER_LINK');
 const SLACK_SUBSCRIPTIONS_URL = nconf.get('SLACK:SUBSCRIPTIONS_URL');
 const BASE_URL = nconf.get('BASE_URL');
+const IS_PRODUCTION = nconf.get('IS_PROD');
+
+const SKIP_FLAG_METHODS = IS_PRODUCTION && !SLACK_FLAGGING_URL;
+const SKIP_SUB_METHOD = IS_PRODUCTION && !SLACK_SUBSCRIPTIONS_URL;
```
Alys, line 15: Just checking: This won't prevent the message being sent to slack when a post is flagged in staging, beta, gamma, and delta, will it? We do want flagging to work normally there, both for testing flagging and for cases where mods are using staging or the other sites for their normal use of Habitica (which is much of the time for staging). It might be worth adding a comment to clarify that.

website/server/libs/slack.js
```diff
@@ -77,14 +89,75 @@ function sendFlagNotification ({
...
+  let title = `Flag in ${flagger.profile.name}'s Inbox`;
+  let text = `${flagger.profile.name} (${flagger.id}; language: ${flagger.preferences.language}) flagged a PM`;
+  let footer = `<${SLACK_FLAGGING_FOOTER_LINK}?userId=${flagger.id}&chatMessageId=${message.id}|Flag this message>`;
+
+  if (userComment) {
+    text += ` and commented: ${userComment}`;
+  }
+
+  if (!message.user && message.uuid === 'system') {
+    authorName = 'System Message';
```
Alys: This isn't needed for inbox messages.

website/server/libs/slack.js
```diff
@@ -77,14 +89,75 @@ function sendFlagNotification ({
...
+  } else {
+    authorName = `${message.user} - ${authorEmail} - ${message.uuid}`;
+  }
+
+  let messageText = message.text;
+
+  if (flagger.id === message.uuid) {
+    messageText += `: ${flagger.profile.name} is writing to itself.`;
```
Alys: This isn't necessarily true. `message.uuid` will be the User ID of the recipient if `sent` is true, or of the sender otherwise, and the flagger could be flagging either message.

It is possible for a player to send themselves a private message and then to flag it, but the chances of that happening are small so I don't think it's worth having code to add a special message for that. It should be clear enough from the other data in the slack message anyway.

Just as a general note though in case it's useful elsewhere, "itself" should be "themself" (or "themselves" if you prefer) because "it" isn't used to refer to people even when gender is unknown. :)

website/server/libs/slack.js
```diff
@@ -77,14 +89,75 @@ function sendFlagNotification ({
...
+
+    if (message.sent) {
+      sender = flagger.profile.name;
+      recipient = message.user;
+    } else {
+      sender = message.user;
+      recipient = flagger.profile.name;
+    }
+
+    messageText += `: ${sender} is writing this message to ${recipient}.`;
```
Alys: Could `sender` and `recipient` please include the email address and User ID in brackets after the name? I.e., the code should produce a message like this:
"ALittleYellowSpider (a@a.com; 3e595299-3d8a-4a10-bfe0-88f555e4aa0c) wrote this message to Alys (b@b.com; d904bd62-da08-416b-a816-ba797c9ee265)"

That will be helpful to mods in the case where two players have similar or identical names.

I think it might be good if the "sender is writing this message to recipient" message was put into the `authorName` variable, rather than into messageText. The `flagSlack.send` code below will then put that message near the top of the slack post.

In case it's helpful, this is what moderators see in slack now when a group message is flagged:

image

So if that were a private message, I'm proposing that the line that says this:
"ALittleYellowSpider - alice.harris@oldgods.net - 3e595299-3d8a-4a10-bfe0-88f555e4aa0c"
be replaced with this:
"ALittleYellowSpider (a@a.com; 3e595299-3d8a-4a10-bfe0-88f555e4aa0c) wrote this message to Alys (b@b.com; d904bd62-da08-416b-a816-ba797c9ee265)"

website/server/libs/slack.js
```diff
@@ -77,14 +89,75 @@ function sendFlagNotification ({
...
+  let footer = `<${SLACK_FLAGGING_FOOTER_LINK}?userId=${flagger.id}&chatMessageId=${message.id}|Flag this message>`;
+
+  if (userComment) {
+    text += ` and commented: ${userComment}`;
+  }
+
+  if (!message.user && message.uuid === 'system') {
+    authorName = 'System Message';
+  } else {
+    authorName = `${message.user} - ${authorEmail} - ${message.uuid}`;
```
Alys: This isn't necessarily the author of the post. It will be the recipient when the flagger is flagging a message they've sent (that's something that will happen legitimately when the flagger wants to report both sides of the conversation to mods). See also my two comments below which are related. One of them has a suggestion for `authorName`.

website/server/libs/slack.js
```diff
@@ -77,14 +89,75 @@ function sendFlagNotification ({
...
+  }
+
+  flagSlack.send({
+    text,
+    attachments: [{
+      fallback: 'Flag Message',
+      color: 'danger',
+      author_name: authorName,
+      title,
+      title_link: titleLink,
```
Alys, line 162: `title_link` can have the value "N/A" for private messages

website/client/store/actions/chat.js
paglias: The client only uses api v4, as far as the server is concerned we can put new methods in v3 but I'd prefer to avoid it

Alys: No worries!

Does that mean the method shouldn't be in website/server/controllers/api-v3/members.js ?

website/client/components/chat/chatCard.vue
negue: Oh I guess I wanted to use the prop `msg.reported` to check it, in a prior design, fixed!

website/client/store/actions/chat.js
negue: moved it completely to `v4`

website/server/libs/chatReporting/inboxChatReporter.js
negue: oh I oversaw this one, thanks 👍 , gonna add "N/A"

website/server/libs/slack.js
negue: can the url be empty? so that it doesn't show it as an url ?

website/server/libs/chatReporting/inboxChatReporter.js
negue: will be removed and just set `flagCount = 1` or should the other user able to flag it too?

website/server/libs/slack.js
negue: do staging / beta / gamma / delta have the "IS_PROD" value? if not I may have to rethink this skips

website/server/libs/slack.js
negue: so I didn't had to update https://github.com/HabitRPG/flag-o-rama/pull/5 ?

website/server/libs/slack.js
Alys: Assuming the Slack side of the integration can handle that, sure! It's worth a try and if it breaks we can change it during testing.

website/server/libs/chatReporting/inboxChatReporter.js
Alys: `flagCount = 1` is fine

The PM in the recipient's inbox is a separate message from the one in the sender's inbox, so there'll only ever be one person flagging each message.
Even if PMs move to a new data structure the way that group messages did, we'd want there to be a separate copy for each of the two users so that each has control over how long the message remains in their inbox. It's just like email in that respect.

website/server/libs/slack.js
Alys: In my opinion, there's no need to update it. However since you've written that code, I will check with the other mods to see if there is any use case that I haven't considered. I feel though that letting the user have the final control over the visibility of their messages is the best option. I'll report back. :)

website/server/libs/slack.js
Alys: I'm not sure. I think they would not. However if they DO have the `SLACK_FLAGGING_URL` and `SLACK_SUBSCRIPTIONS_URL` (which I expect they do since flagging works there now, at least for staging) then the `SKIP_....` constants will be false, which is what we want. 

Although I'm now wondering if the `SKIP_...` constants were meant to be false when IS_PROD is false. We wouldn't need the slack messages to be sent from flags made on a contributor's local install for instance. 
Should it be more like this:
`SKIP_FLAG_METHODS = !SLACK_FLAGGING_URL || !IS_PRODUCTION || !IS_STAGING || ....[beta, gamma, delta]`;
We'd want to tidy that up I guess, and I might be on the wrong track anyway. :)

website/client/components/chat/reportFlagModal.vue
```diff
@@ -106,14 +106,16 @@ export default {
       this.$root.$emit('bv::hide::modal', 'report-flag');
     },
     async reportAbuse () {
-      this.notify('Thank you for reporting this violation. The moderators have been notified.');
+      this.text('Thank you for reporting this violation. The moderators have been notified.');
```
Alys: For group messages, that text is good. For private messages, can it be this:
"Thank you for reporting this message. The moderators have been notified. You can now delete the message if you wish."

Also can that be a translatable string?

website/server/libs/slack.js
Alys: It's confirmed that moderators won't need to flag a private message after a player has reported it. 

I've asked elsewhere in this PR for "You can now delete the message if you wish." to be added to the string the users see after reporting, so that covers the case where the user wants the message to disappear.

paglias:
Looks like @Alys had some comments which have not been addressed yet, I'll tag this `needs-work` again. Ping me when it's ready

website/server/libs/slack.js
negue: @paglias @SabreCat do we have any other config-/env-vars to identify if we are running under staging / beta / etc ? couldn't find any checks in our current code

website/server/libs/slack.js
negue: Ok! 😁 well thats why have reviews 👍  thanks for pointing this out, I thought that might by helpful^^ - gonna remove it.

website/client/store/actions/chat.js
Alys: This might not be necessary but should this file be under a `v4` directory too?
test/api/v3/integration/members/POST-flag_private_message.test.js

website/server/libs/slack.js
paglias: We don't have a separate env var but staging runs in production

website/server/libs/inbox/index.js
```diff
@@ -45,3 +45,13 @@ export async function clearPMs (user) {
...
 }
+
+export async function updateMessage (message) {
+  const messagesInDb = await Inbox
+    .find({id: message.id})
+    .exec();
+
+  const messageInDb = messagesInDb[0];
+
+  await messageInDb.update(message);
```
paglias: shouldn't this follow mongo's update syntax? I don't see any `$set` or anything

website/client/store/actions/chat.js
paglias: yep, the test should be in the v4 folder as well

website/server/controllers/api-v3/chat.js
```diff
@@ -39,6 +39,11 @@ const FLAG_REPORT_EMAILS = nconf.get('FLAG_REPORT_EMAIL').split(',').map((email)
  * @apiError (404) {badRequest} chatIdRequired A chat ID is required
  */
 
+/**
+ * @apiDefine MessageIdRequired
+ * @apiError (404) {badRequest} messageIdRequired A message ID is required
```
paglias: 404 badrequest? A BadRequest is 400

website/server/libs/chatReporting/inboxChatReporter.js
```diff
@@ -0,0 +1,106 @@
...
+    this.req.checkParams('messageId', apiError('messageIdRequired')).notEmpty();
+
+    let validationErrors = this.req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    if (this.user.contributor.admin && this.req.query.userId) {
+      this.inboxUser = await User.findOne({_id: this.req.query.userId});
+    }
+
+    const messages = await inboxLib.getUserInbox(this.inboxUser);
```
paglias: what if line 34 isn't executefd and inboxUser isn't fetched?

website/server/libs/chatReporting/inboxChatReporter.js
```diff
@@ -0,0 +1,106 @@
...
+
+    const message = _find(messages, (m) => m.id === this.req.params.messageId);
```
paglias: instead of fetching the entire inbox can't you fetch just the message directly from the collection?

paglias (changes requested):
left some comments

website/server/libs/chatReporting/inboxChatReporter.js
negue: `inboxUser` is set at line [24](https://github.com/HabitRPG/habitica/pull/10420/files/2c250bfcd9820f95e74f45c817f17d869c1cbbe9#diff-609388719783e1ac10548c4df5e90399R24) 

but I guess this condition (line 33) isn't needed anymore, since there won't be a frag-o-rama link to delete it by an admin/mod, I'll remove this ^^

website/server/libs/inbox/index.js
negue: changed it to mongo's update syntax, but I didn't use $set, will change it needed to use it

website/server/libs/chatReporting/inboxChatReporter.js
negue: updated that too

website/server/libs/inbox/index.js
```diff
@@ -45,3 +49,9 @@ export async function clearPMs (user) {
     Inbox.remove({ownerId: user._id}).exec(),
   ]);
 }
+
+export async function updateMessage (message) {
+  await Inbox.update({
```
paglias: if the message is a mongoose model why not change the message object directly and then call `.save` on it? Instead of updating the entire object on the db. Note that in this case you'll have to mark the `flags` as modified with `message.markModified('flags')` when changing them

paglias (changes requested):
@negue sorry for the delay:

1. I've added a comment on a thing that should be changed
2. The v4 integration tests are failing
3. There are some conflicts

If you can fix those I'll merge it, thanks

Alys:
Once those changes are made, I'd like to have a chance to test it on my local install before it's merged if that's okay. If you can ping me when it's ready, I'll try to get it tested within a couple of days of that.

website/server/libs/inbox/index.js
negue: oh thats even better^^, not that experienced with mongoose^^ thankfully we have reviews haha, changed to `.save`

paglias:
@Alys it should be ready now

website/common/locales/en/groups.json
```diff
@@ -156,11 +156,13 @@
   "cannotSendGemsToYourself": "Cannot send gems to yourself. Try a subscription instead.",
   "badAmountOfGemsToSend": "Amount must be within 1 and your current number of gems.",
   "report": "Report",
+  "canDeleteNow": "You can now delete the message if you wish.",
```
Alys: Should this be in the messages.json file, just under the `reportedMessage` string? That's at line 75.

website/server/controllers/api-v4/members.js
```diff
@@ -0,0 +1,42 @@
+import { authWithHeaders } from '../../middlewares/auth';
+import { chatReporterFactory } from '../../libs/chatReporting/chatReporterFactory';
+
+let api = {};
+
+/**
+ * @api {post} /api/v4/members/flag-private-message/:messageId Flag a private message
+ * @apiDescription An email is sent to the moderators about every flagged message.
```
Alys: Should this comment read "An email and Slack message are sent to..." ?

Alys:
- [ ] There's a double horizontal line in PMs and group chat posts, where on production there's only a single line. From localhost:
image
It appears to be the extra line that's needed when the "you have reported..." message is inserted (see screenshot below).

- [ ] When a mod/staff flags a PM, they see "Message flagged once, not hidden" at the top, which is needed only on flagged group posts (i.e., visible in groups, not in PMs).
image

- [ ] When a mod/staff flags a group post, they see "You have reported this message to moderators." at the bottom, which should be visible in PMs, not groups.
image

Alys:
Should the flag-private-message route be in API v3 so that third-party tools can implement it? But I might be misunderstanding how v3 / v4 works and its use in v3 might already be possible, so ignore this if it's not relevant.

Alys:
Ugh, sorry, definitely ignore that. I'd already asked about v3 last month and paglias said it should be v4 because that's what the client uses. https://github.com/HabitRPG/habitica/pull/10420#discussion_r221571763

We can put something in the wiki to say that private message flagging can't be done on third-party tools that use API v3.
(FYI @cTheDragons since you wrote an inbox tool!).

website/client/components/chat/chatCard.vue
```diff
@@ -18,24 +18,29 @@ div
...
       .action.d-flex.align-items-center(v-if='!inbox', @click='copyAsTodo(msg)')
         .svg-icon(v-html="icons.copy")
         div {{$t('copyAsTodo')}}
-      .action.d-flex.align-items-center(v-if='!inbox && user.flags.communityGuidelinesAccepted && msg.uuid !== "system"', @click='report(msg)')
-        .svg-icon(v-html="icons.report")
-        div {{$t('report')}}
+      .action.d-flex.align-items-center(v-if='(inbox || (user.flags.communityGuidelinesAccepted && msg.uuid !== "system")) && !isMessageReported', @click='report(msg)')
+        .svg-icon(v-html="icons.report", v-once)
+        div(v-once) {{$t('report')}}
         // @TODO make flagging/reporting work in the inbox. NOTE: it must work even if the communityGuidelines are not accepted and it MUST work for messages that you have SENT as well as received. -- Alys
```
Alys: This TODO comment can be removed.

negue:
@Alys all of the mentioned changes fixed now

Alys:
@negue Thank you! :)

@SabreCat Is it possible to merge this into any one of the beta/gamma/delta sites so I can test the emails and slack notifications?

website/common/locales/en/groups.json
```diff
@@ -161,6 +161,7 @@
   "abuseFlagModalBody": "Are you sure you want to report this post? You should <strong>only</strong> report a post that violates the <%= firstLinkStart %>Community Guidelines<%= linkEnd %> and/or <%= secondLinkStart %>Terms of Service<%= linkEnd %>. Inappropriately reporting a post is a violation of the Community Guidelines and may give you an infraction.",
   "abuseFlagModalButton": "Report Violation",
   "abuseReported": "Thank you for reporting this violation. The moderators have been notified.",
+  "pmReported": "Thank you for reporting this message. The moderators have been notified. You can now delete the message if you wish.",
```
Alys: I _think_ from what I've observed, this `pmReported` message is shown to the user as soon as they have submitted the report form - is that right? If so, I'd like the text changed to just this: `"Thank you for reporting this message."`  We don't want them to think they can delete the message until we're sure that moderators have actually been notified. 

The "you can now delete..." text that appears in their inbox after a sync is good - it should stay there since it doesn't appear unless it's certain that the report has been sent to email and slack.

website/server/libs/slack.js
```diff
@@ -77,14 +89,71 @@ function sendFlagNotification ({
...
+
+  if (message.sent) {
+    sender = flaggerFormat;
+    recipient = messageUserFormat;
+  } else {
+    sender = messageUserFormat;
+    recipient = flaggerFormat;
+  }
+
+  authorName = `${sender} is writing this message to ${recipient}.`;
```
Alys: Can you please change "is writing" to "wrote"?

website/client/components/chat/chatCard.vue
```diff
@@ -17,25 +17,28 @@ div
       span.mr-1(v-if="msg.username") •
       span(v-b-tooltip="", :title="msg.timestamp | date") {{ msg.timestamp | timeAgo }}
     .text(v-html='atHighlight(parseMarkdown(msg.text))')
+    .reported(v-if="isMessageReported && inbox")
+      span(v-once) {{ $t('reportedMessage')}}
+      br
+      span(v-once) {{ $t('canDeleteNow') }}
```
Alys, line 17: These two texts (`reportedMessage` and `canDeleteNow`) are still appearing in group chats when a moderator has flagged a message.

image

website/server/controllers/api-v4/members.js
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @apiSuccess {Object} data The flagged private message
+ * @apiSuccess {UUID} data.id The id of the message
+ * @apiSuccess {String} data.text The text of the message
+ * @apiSuccess {Number} data.timestamp The timestamp of the message in milliseconds
+ * @apiSuccess {Object} data.likes The likes of the message (always an empty object)
+ * @apiSuccess {Object} data.flags The flags of the message
+ * @apiSuccess {Number} data.flagCount The number of flags the message has
+ * @apiSuccess {UUID} data.uuid The user id of the author of the message, or of the recipient if `sent` is true
+ * @apiSuccess {String} data.user The username of the author of the message, or of the recipient if `sent` is true
```
Alys: "The username of" should be changed to "The Display Name of" because we currently aren't using the Username in this code.

website/server/libs/chatReporting/inboxChatReporter.js
```diff
@@ -0,0 +1,105 @@
...
+    await super.notify(group, message);
+
+    const groupUrl = getGroupUrl(group);
+    sendTxn(FLAG_REPORT_EMAILS, 'flag-report-to-mods-with-comments', this.emailVariables.concat([
+      {name: 'GROUP_NAME', content: group.name},
+      {name: 'GROUP_TYPE', content: group.type},
+      {name: 'GROUP_ID', content: group._id},
+      {name: 'GROUP_URL', content: groupUrl || 'N/A'},
+      {name: 'REPORTER_COMMENT', content: userComment || ''},
+    ]));
```
Alys: The emails that mods receive aren't always correct. For example, in the email body below, the flagged PM was in "A Little Yellow Spider"'s inbox, "A Little Yellow Spider" both wrote and reported the PM, and "Author" should therefore be "A Little Yellow Spider" but it shows the person that the PM was sent to.

Ideally the email would say:
- who wrote the PM
- who it was sent to
- who reported it

We can create a new email template in Mandril for this if needed.

Also, the "View group" link at the bottom of the email is non-functional, but that's a really minor point. If removing it is not trivial, just leave it there.

The emails are correct when you're flagging a post that someone else sent to you.


```
Offending Content
Tue Nov 27 2018 08:19:31 GMT+0000 (UTC)
alystest: pm from ALittleYellowSpider to Alys. ALittleYellowSpider to flag.

Reporter's Comments
This is ALittleYellowSpider flagging the PM ALittleYellowSpider sent to Alys.

Persons Involved
Author: lady_alys@oldgods.net - Alys - d904bd62-da08-416b-a816-ba797c9ee265
View Profile

Reporter: alice.harris@oldgods.net - A Little Yellow Spider - 3e595299-3d8a-4a10-bfe0-88f555e4aa0c
View Profile

Group Details
Name: N/A
Type: private messages
ID: N/A
View group
```

Alys (changes requested):
This is working pretty well! I've added some comments but I haven't found any critical bugs.

It would be good if the Username was included in the Slack messages and the emails, as well as the Display Name.
We've asked for the same thing to be done for group flagging, so it's possible that there's already some work underway for that. If you wanted to include Username changes for flagging PMs in the work that will be done for flagging group messages rather than adding it in this PR, that would be okay with me. I'm not sure how other mods feel about it though (I can easily look up Usernames with a quick database search but they can't).

website/client/components/chat/chatCard.vue
negue: I can't reproduce this

- my test user is admin
- reported a message
- refresh 
- only the "message hidden" is visible the other ones not

website/common/locales/en/groups.json
negue: `pmReported` is used for the notification text, do you mean this or the text under a private message?

website/server/libs/chatReporting/inboxChatReporter.js
negue: - added `REPORTER_DISPLAYNAME` and `AUTHOR_DISPLAYNAME` to the email-variables (which can be used for the normal chat report email template)

I use the same template for the inbox reporting. This template has only a `message user` and the `reporting user`, so it isn't able to choose the correct text for the email. 

I think it would be better to create a new email template just for private messages
- @Alys can you create it? 
- or @paglias @SabreCat could you give me access to the templates? or are those hosted in this repo too?

website/client/components/chat/chatCard.vue
Alys: Perhaps the text only appears if you've flagged a group message after recently having the private message inbox open?

website/server/libs/chatReporting/inboxChatReporter.js
Alys: Yes, I can create a new template. I'll do that on the weekend. They're stored in Mandrill / MailChimp. I'll report back here with the details you'll need to know after I've done it.

website/common/locales/en/groups.json
Alys: Yes, I mean the notification text. It seems to be shown to the user as soon as they've submitted the form, so at that point we can't be sure that the moderators have received the slack post and email, and so we don't want the user to prematurely delete the message they tried to report.

The text under the private message doesn't need to have the deletion tip removed since it doesn't have the same potential problem (it's only shown to the user after the code has confirmed that the email and slack message were sent correctly, right?)

website/server/libs/chatReporting/inboxChatReporter.js
Alys: I've looked through the email template and I'd actually like us to keep using the same one for flagging both PMs and group messages for ease of maintenance of the templates and because I think it will be easy enough to get the sender/recipient details correct with an `if` statement. Even if a new template was used, we'd still need that `if` to correctly distinguish between sender/recipient. I have adjusted the template so it can be used with PMs without affecting its use for group messages.

Here's the changes that are needed in the code:
- _(This should already be correct for both PMs and group messages.)_ The "REPORTER" details (`REPORTER_UUID`, etc) should always be the person doing the flagging.
- [x] For private messages, the "AUTHOR" (`AUTHOR_UUID`, etc) should always be the person who sent the message. That can be determined by looking at `msg.sent` and then using either `msg.uuid` (if `msg.sent` is `false`) or `msg.ownerId`/`this.inboxUser` (if `msg.sent` is `true`)
- _(This should already be correct.)_ For group messages, the "AUTHOR" should always be the person who sent the message.
- [x] For private messages, `AUTHOR_USERNAME` should be changed to contain `user.auth.local.username`. For group messages, it's desirable that the same change be made but not essential (we can change it in a later PR).
- [x] For private messages, `AUTHOR_DISPLAY_NAME` should be added and should contain `user.profile.name`. For group messages, it's desirable that the same addition be made but not essential.
- [x] For private messages, new parameters will be needed: `RECIPIENT_EMAIL`, `RECIPIENT_DISPLAY_NAME` (`user.profile.name`), `RECIPIENT_USERNAME` (`user.auth.local.username`), `RECIPIENT_UUID`. They should always indicate the person who received the message.
- _(This is already correct.)_ For group messages, the "RECIPIENT" fields can be left undefined - they do not need to be in the code at all. The email template can handle that.
- Ignore this comment from above, it's not worth fixing: _Also, the "View group" link at the bottom of the email is non-functional, but that's a really minor point. If removing it is not trivial, just leave it there._

paglias:
Moved back to `needs work` as there are some unresolved comments, but move it back to `ready for review` when the changes are made!

website/common/locales/en/groups.json
negue: yes the text under the private message is only show (if the message has flags OR when the API-Result comes back with an `ok` )

website/client/components/chat/chatCard.vue
negue: will try to reproduce this tmrw

negue:
Refactored the email-variables (and the other requested changes) it should be good to go now

(although I don't understand the weird git history is visible, I merged as I always do, this shouldn't list other commits .. weird^^)

website/client/components/chat/chatCard.vue
Alys: This is still happening although only for moderators/staff.

website/server/controllers/api-v4/members.js
```diff
@@ -0,0 +1,42 @@
...
+ * @apiSuccess {Number} data.timestamp The timestamp of the message in milliseconds
+ * @apiSuccess {Object} data.likes The likes of the message (always an empty object)
+ * @apiSuccess {Object} data.flags The flags of the message
+ * @apiSuccess {Number} data.flagCount The number of flags the message has
+ * @apiSuccess {UUID} data.uuid The user id of the author of the message, or of the recipient if `sent` is true
+ * @apiSuccess {String} data.user The Display Name of the author of the message, or of the recipient if `sent` is true
+ *
+ * @apiUse MessageNotFound
+ * @apiUse MessageIdRequired
+ * @apiError (400) {BadRequest} messageGroupChatFlagAlreadyReported The message has already been flagged
```
Alys: The text for this is "You have already reported this message" (i.e., that's the text that already appears in messages.json so the only change needed is to this apiError comment here).

website/server/controllers/api-v4/members.js
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @apiSuccess {Object} data The flagged private message
+ * @apiSuccess {UUID} data.id The id of the message
+ * @apiSuccess {String} data.text The text of the message
+ * @apiSuccess {Number} data.timestamp The timestamp of the message in milliseconds
+ * @apiSuccess {Object} data.likes The likes of the message (always an empty object)
+ * @apiSuccess {Object} data.flags The flags of the message
+ * @apiSuccess {Number} data.flagCount The number of flags the message has
+ * @apiSuccess {UUID} data.uuid The user id of the author of the message, or of the recipient if `sent` is true
+ * @apiSuccess {String} data.user The Display Name of the author of the message, or of the recipient if `sent` is true
```
Alys, line 22: `data.username` can be added here now, with text like this:
"The Username of the author of the message, or of the recipient if `sent` is true"

Also in the `data.uuid` comment just above here, can you capitalise "User ID" like that, for consistency with how we usually refer to it?

website/server/libs/chatReporting/inboxChatReporter.js
Alys: I'm not certain all of these parameters are being passed through to the email template. For example, I can't see `RECIPIENT_DISPLAY_NAME` in the code, and it's blank in the emails that are sent. The problem might be that "_DISPLAYNAME" is used in the code instead of "_DISPLAY_NAME". It's probably worth fixing just that and then getting it pushed to delta and I'll test again.

NB we use USERNAME with no  underscore and DISPLAY_NAME with an underscore because that most closely mimics how they're written in user-visible text ("Username" and "Display Name").

Alys (changes requested):
The behaviour that the user sees in their inbox when flagging is all good. The messages sent to moderator slack are great!

website/client/components/chat/chatCard.vue
negue: I set my account to `admin: true` and `level: 8` and I still don't get the `reportedMessage`/`canDeleteNow` messages:

Reported messages in party / groups / tavern / guilds and private messages and I still not getting em.

Do you have an example group / database you could send me to test it?

negue:
changed to `_DISPLAY_NAME` now those should be ok in the email templates too

website/client/components/chat/chatCard.vue
Alys: https://habitrpg-delta.herokuapp.com/groups/tavern

image

```
{
    "_id": "9c0b5eb8-6c86-48a8-b2c0-fcb300dbdbf2",
    "flagCount": 5,
    "flags": {
        "d904bd62-da08-416b-a816-ba797c9ee265": true
    },
    "id": "9c0b5eb8-6c86-48a8-b2c0-fcb300dbdbf2",
    "text": "Hi",
    "timestamp": "2019-01-06T10:28:50.214Z",
    "likes": {},
    "client": "ios",
    "uuid": "xxxxxx,
    "contributor": {},
    "backer": {},
    "user": "xxxxx",
    "username": "xxxxx",
    "groupId": "00000000-0000-4000-A000-000000000000",
    "__v": 0
}
```

I've redacted the user's names and ID since it's an innocent message.  The post was made about 25 minutes ago and I flagged it about 20 minutes ago.

Are you doing your testing on delta or in your local install? Maybe there's code locally that hasn't been pushed to GitHub yet? Or perhaps there's code in this PR that's not on delta yet, although I thought Sabe deployed it fairly recently (at the very least in the past month unless I'm badly misremembering).

If you like, put this specific problem on hold until after the next deployment to delta. We can at least test the `_DISPLAY_NAME` change now. I'll report back here after that.

@SabreCat when you have time, can you deploy this PR to delta again?

paglias:
The new changes are on delta

website/client/components/chat/chatCard.vue
negue: I'm testing locally, I don't have mod/staff level on delta ^^ 

But still its very weird issue, the messages are only shown if this is inside the inbox, so for that `:inbox='true'` must be set. so it should be only visible in the private message log

I'll change the props to use the other syntax with type and default value = false

Alys: It's still happening on delta after the latest delploy.

negue: I refactored the properties to add a default value, what I guess is that the binding of `undefined to :inbox="inbox"` may caused the inner checks of `inbox` being true which shouldn't be, if the flagged text still exists after that I'm out of ideas, since I can't reproduce it locally

Alys:
@SabreCat Can this please be deployed to delta again when convenient? It's not urgent.

paglias:
@negue looks like some tests are failing

negue:
@paglias should be good to go :)

Alys:
@SabreCat Can this be deployed to delta please, when you have time?

paglias:
sorry @negue can you give a look at the conflicts?

negue:
@paglias fixed

Alys:
@SabreCat can this please be merged to delta?

paglias:
@Alys deployed

Alys:
Thanks @paglias 

@negue It's stopped putting that "you have reported this message to moderators" text on group posts - thanks! I'll test all the other functionality as soon as I can.

Alys:
I've done some thorough test of guild, party, Tavern, and private message flagging, including checking the messages mods see in Habitica, in Slack, and in emails. It all looks good!

The only problem is that when flagging PMs, you can't see the blue notification that appears because its z-index is lower than that of both the modals that are open at the time (you can _just_ make it out if you look carefully but can't read the text), but I don't think that's important. The red message that appears on the PM after it's been flagged is plenty good for telling the user their flagging worked. I'm only mentioning this to be thorough. :)

I think this is fine to go live. Thanks @negue!

Alys (approved):

paglias:
Merged!

PR #10427, opened 2018-06-04
Inbox opt in / opt out: perf improvement and new wording

### LINKED ISSUE #10422: Inbox / Private Messages opt-in opt-out switch: server errors, sync delay, toggle not updating
When you toggle the switch for opting in / out of Private Messages, the sync bar takes a long time to disappear. When it finally does, the message below appears in the console:

```
Unhandled promise rejection 
Error: Request failed with status code 502
Stack trace:
t.exports@https://habitica.com/static/js/vendor.d27c35f55e47c473e365.js:6:130061
t.exports@https://habitica.com/static/js/vendor.d27c35f55e47c473e365.js:12:1220
t.exports/</h[p]@https://habitica.com/static/js/vendor.d27c35f55e47c473e365.js:6:129075
es6.promise.js:110
```

In addition, for some of my tests, I received a 502 error message. Nothing appeared in Loggly for my account. I was not using Habitica in another location at the same time and had reloaded the website before and during my tests so version errors and similar can't have been affecting this.

After the sync finally finishes, the change you made sometimes IS recorded in the database (i.e., your value of `inbox.optOut` sometimes is changed appropriately), even when the 502 error occurs. 

Also, I think the label on the toggle doesn't always change to the correct value after the sync finally occurs (i.e., if it said "Disable Private Messages" before you toggle it, it still says that afterwards) - but that might be an intermittent problem or might not actually be occurring - because of the confusion caused by the delay, I'm not sure and I haven't bothered testing more thoroughly for that aspect of this bug because we need to fix the sync delay / server error anyway, and we can test for the toggle changing appropriate when the fix is being tested.

The wording used might also be confusing, as described in comments from AnneLine (16a107de-d519-43ce-8c30-8da5cbcbb237) in the Report a Bug guild (below), although once the switch is working correctly, that confusion might be wholely or partly removed. I suggest that for now we fix the bug with the toggle and then adjust the wording in a different fix later if that still seems to be needed.

----

AnneLine: _"Not sure if this a bug or just something new.  Is there a new 'feature' for turning off/on private messages?  I just received a report via a guild chat that a message to me was refused, and when I went to my pm box, it looked as if there was a new toggle in upper right corner that was grey - and now I have switched it to purple. Is this something that's been there for a while and I just discovered it, or is it new? Either way, any chance it could be put into a Bailey announcement because I'm seeing lots of references to people not getting messages, and maybe there's a connection?"_

citrusella, the serving wench: _"Technically it was a feature before the redesign and was only reintroduced within the past week. (Before that, I believe whatever the switch was on was honored but there was no way to change it because the switch wasn't accessible.) But yes, the switching being back is new."_

AnneLine: _"thanks for clarifying that. It really has created a big problem for me (and apparently for others) because it's very hard to know which way is turned on and which way is turned off, and the wording is very, very confusing. It took three tries for me to get it right, with several people having to send me PMs to confirm what was going on. Would there be a way to have it say 'open for messages' and 'closed for messages' reflecting the status or something? Many people have trouble distinguishing those colors, and many people (me, me!) reverse left and right."_

paglias:
Fixes #10422

- Improved performances when updating the user preferences (locally went from 300ms to 30 for an user with a lot of messages)

WIP

- change the wording for the toggle (waiting for the final text from @Alys  and @shanaqui )

### DIFF
--- a/website/server/controllers/api-v3/user.js
+++ b/website/server/controllers/api-v3/user.js
@@ -294,7 +294,9 @@ let checkPreferencePurchase = (user, path, item) => {
  */
 api.updateUser = {
   method: 'PUT',
-  middlewares: [authWithHeaders()],
+  middlewares: [authWithHeaders({
+    userFieldsToExclude: ['inbox.messages'],
+  })],
   url: '/user',
   async handler (req, res) {
     let user = res.locals.user;


paglias:
Nevermind, the pef change is impossible because it would break the API

PR #10428, opened 2018-06-04
Move inbox to its own model

### LINKED ISSUE #10422: Inbox / Private Messages opt-in opt-out switch: server errors, sync delay, toggle not updating

### LINKED ISSUE #10621: Outgoing private messages appear doubled on client

### General Info
  * UUID: 7f14ed62-5408-4e1b-be83-ada62d504931
  * Browser: Firefox
  * OS: Windows 10

### Description

When you first send off a private message to a user, the message is duplicated in the messages UI, making it appear to have been sent twice. After a refresh, the copy goes away. Reproduced in both staging and production environments.

image

#### Console Errors

None.

paglias:
WIP

Fixes #10422
Fixes #10621

Started by creating a single schema to use for both inbox and chat messages

TODO
- [x] inbox in its own model
- [x] cache user styles for inbox messages as well
- [x] update client to use new routes (get inbox, delete message, clear messages...)
- [x] client: store messages in object and delete/add messages from there, remove all code mentioning inbox.messages
- [x] handle loading for habitica::new-inbox-message
- [x] use cached user styles in client
- [x] migration
- [x] indexes
- [x] return sent message when using members/send-private-message (missing tests)
- [x] api.blockUser
- [x] `also. for the process of moving over to the new API call: would it be an option to have the API call available while still having the messages stored in the user object and then once the apps have been updated we give users another week or so and then remove the messages from the user?`

### Routes
- [x] POST /members/send-private-message
- [x] POST /members/transfer-gems
- [x] GET /inbox/messages
- [x] DELETE /inbox/messages/:messageid
- [x] Clear PMs
- [x] update GET /export/inbox.html
- [x] update GET other exports
- [x] get inbox for v3
- [x] implement GET /user
- [x] tests for GET /user in v4
- [x] api.enterCouponCode
- [x] api.castSpell
- [x] api.updateUser
- [x] api.getUserAnonymized (not migrated, it also returns tasks)
- [x] api.userRebirth
- [x] api.userReroll
- [x] api.userReset
- [x] calls to API v4 routes with params (cast spell, enter coupon code) are being sent to v3...
- [x] api.registerLocal
- [x] get user -> userLib ?

After the migration, merge and deploy #10487


website/server/controllers/api-v3/inbox.js
```diff
@@ -0,0 +1,30 @@
...
+/* NOTE this route has also an API v4 version */
+/**
+ * @api {get} /api/v3/inbox/messages Get inbox messages for a user
+ * @apiName GetInboxMessages
+ * @apiGroup Inbox
+ * @apiDescription Get inbox messages for a user
+ *
+ * @apiSuccess {Array} data An array of <a href='https://github.com/HabitRPG/habitica/blob/develop/website/server/models/group.js#L51' target='_blank'>chat messages</a>
+ */
+api.getInboxMessages = {
```
TheHollidayInn: why do we have a v3 and v4 version of this route?

paglias: It's a leftover from when they were behaving differently but the v4 one can be removed as the code is now the same

PR #10430, opened 2018-06-05
Open correct group modal from header

### LINKED ISSUE #10238: "Invite friends" button in header doesn't work as expected
When you have a party with only yourself and the header shows the Invite friends button and you click on it the list of party members is shown instead of the invite modal

probaton:

Fixes #10238

### Changes
Clicking one of the various party buttons in the header now does a check on the status of your party before serving the user a modal. 
- If the user had no party at all, the 'create a party' modal is shown. 
- If the user is in a party on his/her own, the 'invite friends' modal is shown.
- If the user has party members, the party overview modal is shown. 

On a totally unrelated note I also ran into an outdated README.md in the server folder which only contained information relevant to the site overhaul. As the overhaul is finished, the file did not seem to serve any purpose other than to confuse new users, so I removed it. 

----

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -1,3 +1,0 @@
-`/website/client` contains the source files for the new client side that is being developed as part of the Habitica.com redesign. Static assets for the new client can be found in `website/static`.
-
-The old client side files can be found in `/website/client-old`, they are still used on Habitica.com while the redesign is in progress.
--- a/website/client/components/groups/group.vue
+++ b/website/client/components/groups/group.vue
@@ -1,7 +1,6 @@
 <template lang="pug">
 .row(v-if="group._id")
   group-form-modal(v-if='isParty')
-  invite-modal(:group='this.group')
   start-quest-modal(:group='this.group')
   quest-details-modal(:group='this.group')
   group-gems-modal
@@ -260,7 +259,6 @@ import membersModal from './membersModal';
 import startQuestModal from './startQuestModal';
 import questDetailsModal from './questDetailsModal';
 import groupFormModal from './groupFormModal';
-import inviteModal from './inviteModal';
 import groupChallenges from '../challenges/groupChallenges';
 import groupGemsModal from 'client/components/groups/groupGemsModal';
 import questSidebarSection from 'client/components/groups/questSidebarSection';
@@ -288,7 +286,6 @@ export default {
     membersModal,
     startQuestModal,
     groupFormModal,
-    inviteModal,
     groupChallenges,
     questDetailsModal,
     groupGemsModal,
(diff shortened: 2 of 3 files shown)


paglias:
@probaton I've tested the fix and there's a problem, it works only after you reload the page, until you reload the start a party text remains and clicking on it doesn't do anything

probaton:
Aaaarg. I don't remember actually changing anything in that behavior, but that definitely sounds broken to me. I'll take a look as soon as I can find a spare half hour.

probaton:
Wow, that took waaaay longer than it should have. I spent a million years trying to figure out why the update mechanism was broken before finally noticing the `v-once` tag and guessing at its purpose. I should probably invest in a basic Vue tutorial at some point...

paglias:
@probaton thanks! This works well but there's another problem now, the text changes correctly but if you create a party until the page is reloaded clicking the invite friends button doesn't work

schermata 2018-06-12 alle 11 26 21

^ this is what appears in the console, probably has something to do with the party members not being loaded correctly?

probaton:
@paglias Sorry about that! Was far too focused on getting a fix out to bother with anything as trivial as actually testing my own code. 

Spent another couple hours trying to figure out how Vue handles dependencies between computed values and this is what rolled out. I even got off my lazy butt to test the new implementation, so fingers crossed you don't find anything this time around. 

I even found another bug: the invite modal was loading twice when clicking the header Invite button because both `index.vue` and `group.vue` were loading `invite-modal`. As a result I've removed `invite-modal` from `group.vue` entirely. That means that the Invite button in the Party page will only work as long as the header is loaded. I couldn't think of a scenario in which that wasn't the case, but please let me know if I'm missing a use case, or if there's a more elegant solution.

paglias:
@probaton I haven't tested it yet but my main concern is about this line https://github.com/HabitRPG/habitica/pull/10430/files#diff-8cbf1baa36a2fdb9de136c1614f62226R3 it specifies the group to use in the invite modal as the party, so maybe it won't work for other groups?

probaton:
Wow, you are really good at making me feel like an idiot. I'd think you were doing on purpose if you weren't also right all the time. So I suppose I'm actually really good at making myself feel like an idiot :)!

Once again; good catch! The header now has it's own fancy event listener. I'm not entirely certain that's the right place to leave it, but the alternatives all seemed slightly worse? If you guys prefer I move it to e.g. app.vue just let me know.

paglias:
@probaton I'm just more used to this codebase :) it's more a matter of time spent on it and it's far more difficult to understand a codebase when it's already been built then if you wrote at least parts of it!

Tested and it's working well! We'll merge it in the next few days

paglias:
thanks @probaton ! notated towards your 3rd tier

PR #10442, opened 2018-06-10
Habits: store one history entry per day

### LINKED ISSUE #10386: cron halts part way through when history arrays are large
This isn't a new bug and it's been mentioned in other issues (e.g., https://github.com/HabitRPG/habitica/issues/7727, https://github.com/HabitRPG/habitica/issues/7909, https://github.com/HabitRPG/habitica/issues/9068) but it's still happening so I think it's worth having a new issue opened for it.

When a player has Habits or Dailies with large `history` arrays or if their `user.history` array is large, cron will sometimes halt part way through, presumably because it crashes when adding to or preening the arrays. Sometimes Loggly shows VersionErrors but not always.

Admins can fix this for individual players when it's reported by emptying the arrays. Cron then runs normally on the next day. My personal feeling for when the arrays start to cause problems is around 300 entries.

We could fix this automatically by having cron purge the history arrays when they're too large, but that's not ideal, especially for subscribers who have paid to have history kept longer.

Another option would be to wrap try-catch blocks or similar error handling around the cron code that processes individual tasks so that if one task causes a failure, the rest of cron can continue (including processing the other tasks). That would also make cron more robust for any other current or future problems that might cause failures. I don't know how possible this would be but I think it might be worth investigating, rather than purging the history prematurely.

I believe the staff have plans to keep improving Habitica's architecture to prevent VersionErrors and similar problems, but I don't know the timeframes on that so a workaround now might be helpful.

I'm marking this suggestion-discussion for us to determine the best approach.

paglias:
This PR refactors the task history for habits to make sure only one entry is stored per habit per day.

It also adds two new fields for habits history entries: `scoredUp` and `scoredDown` to store the number of times the habit was scored up and down during a day.

It also removes the support for scoreNotes

Fixes #10386


Alys:
How much notice about this can we give to Comrades after it's finished and before it goes live? If it's going to be less than a couple of weeks, it's probably worth making a blog post about it now to provide as much advance notice as possible.

paglias:
I'd say definitely less than two weeks so if someone has time to write a blog post it would be great

Alys:
_This comment is for the benefit of third-party developers. I'm putting it here so I can link to it from various places in Habitica._

This pull request changes how history data is stored for Habits. It is a **breaking change**. I.e., any third-party tool that uses or displays Habit history data might stop working correctly because of this PR (depending on what the tool does with the data).

The current situation (i.e., before this PR goes live) is that every time you click + or - on a Habit, an entry is created in the task's `history` array, containing the exact date and time of your click and the [value](https://habitica.wikia.com/wiki/Task_Value) of the Habit after the click. Thus you can see the exact time of all clicks on your Habits, at least for recent clicks ([older data is compressed](https://habitica.wikia.com/wiki/Task_History_Compression)). Here is an example of a Habit's `history` data:
```
{
    "_id": "15d64372-ca78-404b-adb1-dacdbab8ef28",
    "text": "Sample Habit",
    "type": "habit",
    "value": -1.026657310999762,
    "history": [
        {
            "date": 1466294947739,
            "value": -1.026657310999762
        },
        {
            "date": 1466294948433,
            "value": -2.081442996376489
        },
        {
            "date": 1466294949443,
            "value": -3.165953482194662
        },
        {
            "date": 1491765475188,
            "value": -1.026657310999762
        },
        {
            "date": 1498324356562,
            "value": -1.026657310999762
        }
    ],
    ...
    other fields
    ...
}
```

This PR changes that. Instead of one history entry for each click, there will be only one history entry for each **day**. The entry will contain a timestamp indicating the day, the value of the Habit at the end of that day, and two new fields, `scoredUp` and `scoredDown`, which contain the total number of clicks on the plus and minus buttons for that day. An example is below.

For the current day, the timestamp will change each time the user clicks the Habit. It will not change when cron runs.

(_to be confirmed:_ [Task History Compression](https://habitica.wikia.com/wiki/Task_History_Compression) will still happen for days far in the past.)

```
{
    "_id": "15d64372-ca78-404b-adb1-dacdbab8ef28",
    "text": "Sample Habit",
    "type": "habit",
    "value": 1.0000021865081459,
    "history": [
        {
            "date": 1528626801010,
            "value": 1.0,
            "scoredUp": 1,
            "scoredDown": 0
        },
        {
            "date": 1528713151881,
            "value": 1.0000021865081459,
            "scoredUp": 5,
            "scoredDown": 4
        }
    ],
    ...
    other fields
    ...
}
```
In that sample data, you can see that on the first day, the user clicked plus on the Habit once, and did not click minus at all. On the second day, the user clicked plus five times and clicked minus four times.


When this change goes live, staff will run a migration script that will change the `history` arrays for all existing Habits, so that all old history data will be converted to the new format. This migration will take into account the user's current time zone and their Custom Day Start time so that history entries will be merged into the correct day from the point of view of the user.

### How this might affect your third-party tool

If your third-party tool does not display or use any of the Habit's `history` data, then it should not be affected by this PR and you shouldn't need to make any changes.

If your tool displays the number of clicks that the user made on a Habit each day, then your tool will stop working correctly and must be modified. For example, instead of counting the number of entries for each day in the `history` array, your code will need to find the history entry for each day and read the `scoredUp` and `scoredDown` fields.

If your tool displays the time of day that each Habit was clicked (e.g., as in the "Habit History" section in the [Official User Data Display Tool](https://oldgods.net/habitrpg/habitrpg_user_data_display.html)), then that feature will break and cannot be fixed - after this change is released, it will not be possible to tell when a Habit was clicked, not even for Habit clicks that occurred before this change.

### Where to ask questions about this change

If you have any questions about this change and its effect on your third-party tool, please ask them in the [Aspiring Comrades](https://habitica.com/groups/guild/2ff9822b-27f2-4774-98da-db349b57a38e) guild, **not** in this pull request. Your questions will be answered in that guild and then this post will be updated with any necessary clarifications.

Alys:
For anyone who wants to experiment with the new history format before it's released, it's available now on Habitica's beta website: https://habitrpg-beta.herokuapp.com/
That site uses a different database than the habitica.com site. The database was copied from habitica.com to the beta site some time ago, so you might find that you have an account there already but with old data in it. If you don't have an old account or if you don't want to use an old account, you can create one or more new test accounts. (Please be aware that the beta site and database might be deleted at any time with no notice, so don't use the beta site for storing real data.)
On the beta site, old Habit history entries have **not** been migrated to the new format so to experiment with the new format, you'll need to add clicks to existing Habits or create new Habits.

The API can be used with the beta site. You use it in the same way that you would for the production site but change habitica.com to habitrpg-beta.herokuapp.com in the API's URLs. For example, to use the API on the beta site to create a To-Do: 
`curl -X POST -d "text=TEST&type=todo" -H "x-api-user: $MY_USERID" -H "x-api-key: $MY_API_TOKEN" https://habitrpg-beta.herokuapp.com/api/v3/tasks/user/`

Alys:
The migration mentioned above will be started as soon as the change is deployed however it will take up to 10 hours to run. During that time, each Habit could have history entries in both the old and new format. This might cause problems for any tool that has been updated to use only the new format, however as soon as the migration has finished, those problems will stop. After that time, there will be no Habits with the old history format.

PR #10443, opened 2018-06-11
Add timestamp to moderator Slack messages (fixes #10441)

### LINKED ISSUE #10441: timestamp needed in moderator Slack messages
_I'm marking this "good first issue" but it might be a little harder than those often are. If you start this and find you need help with it, post here._

When a chat message is flagged, moderators receive a notification in a Slack channel as shown in the screenshot below. If you use `git grep flagSlack` you'll find the code responsible for it.

We'd like the date and time of the message to be included. That data is stored in the message's `timestamp` field like this:
```
{
	"_id" : "34df5f7f-205b-4797-ae7d-443777bae1ec",
	"flagCount" : 5,
	"flags" : {
		"d904bd62-da08-416b-a816-ba797c9ee265" : true
	},
	"id" : "34df5f7f-205b-4797-ae7d-443777bae1ec",
	"text" : "Test post for flagging.",
	"timestamp" : ISODate("2018-06-09T23:56:26.367Z"),
	"likes" : { },
	"uuid" : "3e595299-3d8a-4a10-bfe0-88f555e4aa0c",
	"user" : "ALittleYellowSpider",
	"groupId" : "85dd1be6-336b-4e53-9949-2c09150b4035",
	"userStyles" : { ... },
	"__v" : 0
}
```

The format in the message moderators see in Slack should be `yyyy-mm-dd HH:MM UTC`. For example, the timestamp for the message above would be reported as `2018-06-09 23:56 UTC`. We don't want the second or milliseconds to show - that would just be cluttery. :) moment().format() is the best way to format a date. You can see examples of it in our code with `git grep 'moment(' | grep 'format('`

The timestamp in the message is already in UTC and there's no need to try to convert it to any other time zone, so we just need the "UTC" label added as a reminder to mods that it's UTC not their own time zone.

The timestamp should appear in the Slack message on a line by itself directly above the guild's name, like this:
```
ALittleYellowSpider - alice.harris@oldgods.net - 3e595299-3d8a-4a10-bfe0-88f555e4aa0c
2018-06-09 23:56 UTC
Flag in Chaos Guild - (private guild)
Test post for flagging.
Flag this message
```

It should be in the same grey text as the user's name, email address, and User ID.

_When a message is flagged, other things happen too (e.g., an email is sent to various email addresses), but those other actions do NOT need to be changed._

thefifthisa:

Fixes #10441 

### Changes

Added the date and time (in UTC) to the Slack notification moderators receive when a chat message gets reported. The sendSlackNotification function in website/server/libs/slack.js was changed.

----

### DIFF
--- a/website/server/libs/slack.js
+++ b/website/server/libs/slack.js
@@ -11,6 +11,7 @@ const BASE_URL = nconf.get('BASE_URL');
 
 let flagSlack;
 let subscriptionSlack;
+let moment = require('moment');
 
 try {
   flagSlack = new IncomingWebhook(SLACK_FLAGGING_URL);
@@ -29,7 +30,6 @@ function sendFlagNotification ({
   if (!SLACK_FLAGGING_URL) {
     return;
   }
-  let moment = require('moment');
   let titleLink;
   let authorName;
   let timestamp;
@@ -56,7 +56,7 @@ function sendFlagNotification ({
 
   let date = moment(message.timestamp).format('YYYY-MM-DD');
   let time = moment(message.timestamp).utc().format('HH:mm');
-  timestamp = date + " " + time + " UTC";
+  timestamp = `${date} ${time} UTC`;
 
   flagSlack.send({
     text,
--- a/website/server/libs/slack.js
+++ b/website/server/libs/slack.js
@@ -3,6 +3,7 @@ import { IncomingWebhook } from '@slack/client';
 import logger from './logger';
 import { TAVERN_ID } from '../models/group';
 import nconf from 'nconf';
+import moment from 'moment';
 
 const SLACK_FLAGGING_URL = nconf.get('SLACK:FLAGGING_URL');
 const SLACK_FLAGGING_FOOTER_LINK = nconf.get('SLACK:FLAGGING_FOOTER_LINK');
@@ -11,7 +12,6 @@ const BASE_URL = nconf.get('BASE_URL');
 
 let flagSlack;
 let subscriptionSlack;
-let moment = require('moment');
 
 try {
   flagSlack = new IncomingWebhook(SLACK_FLAGGING_URL);
@@ -32,7 +32,6 @@ function sendFlagNotification ({
   }
   let titleLink;
   let authorName;
-  let timestamp;
   let title = `Flag in ${group.name}`;
   let text = `${flagger.profile.name} (${flagger.id}; language: ${flagger.preferences.language}) flagged a message`;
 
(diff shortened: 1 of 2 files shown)


website/server/libs/slack.js
```diff
@@ -11,6 +11,7 @@ const BASE_URL = nconf.get('BASE_URL');
 
 let flagSlack;
 let subscriptionSlack;
+let moment = require('moment');
```
paglias: moment should be imported at the top with `import moment from 'moment'`

website/server/libs/slack.js
```diff
@@ -31,6 +32,7 @@ function sendFlagNotification ({
   }
   let titleLink;
   let authorName;
+  let timestamp;
```
paglias: if you're not changing the value of timestamp it can be `const timestamp = moment...` right when you need it at line 59

website/server/libs/slack.js
```diff
@@ -52,12 +54,17 @@ function sendFlagNotification ({
     authorName = `${message.user} - ${authorEmail} - ${message.uuid}`;
   }
 
+  let date = moment(message.timestamp).format('YYYY-MM-DD');
+  let time = moment(message.timestamp).utc().format('HH:mm');
```
paglias: why one is utc and the other isn't?

website/server/libs/slack.js
```diff
@@ -52,12 +54,17 @@ function sendFlagNotification ({
     authorName = `${message.user} - ${authorEmail} - ${message.uuid}`;
   }
 
+  let date = moment(message.timestamp).format('YYYY-MM-DD');
```
paglias: you should only call moment once, also for variable that do not change it's better to use `const` instead of `let`. I think you should be able to condense everything in one line here

```
const timestamp = `${moment(message.timestamp).utc().format('YYYY-MM-DD HH:mm')} UTC` 
```

paglias (changes requested):
left some comments!

thefifthisa:
@paglias Ahhh, thanks! Using `const` and condensing everything makes so much sense in hindsight. Fixed them in the latest commit.

paglias (changes requested):
The code looks good, thanks @thefifthisa 

I'm wondering if passing the timestamp inside the `attachments` is going to produce the desired effect. According to https://api.slack.com/docs/messages/builder the `timestamp` property isn't supported. So maybe you should add the timestamp to the `text` variable?

thefifthisa:
@paglias Oh okay! I added it to `author_name` and it looks good on Slack API preview (see below), but one of the tests is still failing?

test

paglias:
That looks great! The tests are failing because they don't expect the timestamp, for example here it's expecting a string without the timestamp https://github.com/HabitRPG/habitica/blob/b1b1e512f57fe4efe825b7fcb04b6cdb6ba59860/test/api/v3/unit/libs/slack.js#L54, you should update that test to also include `\n${timestamp}`

thefifthisa:
Got it, now passing all the tests!

paglias:
@thefifthisa  thanks for this! I'm merging the PR and awarding your contributor tier.

Issue #10124, opened 2018-03-11
group plan "quantity" changes need to be more reliable


Alys:
When the leader of a group plan invites new members or removes existing members, the group's `quantity` field is updated and I think the change is also sent to Stripe / Amazon at the same time, rather than at the next scheduled billing period.

This can sometimes fail. E.g., a few months back, shanaqui's group plan had three members in it, then one member was removed. However the guild's `quantity` field stayed at 3 and payments continued to be made for three members ($15 per month).

The code that adds or removes members to/from a group plan needs to have checking in place to ensure that all necessary actions succeed. If any don't succeed, the code should either try again (and again and again until successful) or notify admins (and check that the notification was successful).

------

With PR #12335 the Habitica codebase has been updated to support MongoDB 4.2 and transactions making it possible to fix this issue. 

Transactions allow operations on multiple documents to be executed ensuring that either all of them are executed correctly or none, making it possible to fix this issue.

An example on using transactions can be found at https://mongoosejs.com/docs/transactions.html while more info at https://docs.mongodb.com/manual/core/transactions/. If you want to work on this issue and have any question please leave a comment!

shanaqui:
I thought you fixed the quantity problem with my group plan anyway, but this month it was still $15 as if there were three members (nope)...

Alys:
Yes, I had adjusted the quantity field back when we first found that problem in your group plan but I think this part is important for correct billing: "the change is also sent to Stripe / Amazon at the same time, rather than at the next scheduled billing period". I think if that's not done automatically at the time that the membership is changed, then Stripe never finds out that it should be charging more/less. Basically the quantity field is irrelevant for that, I think. We hadn't realised that when we were investigating your plan's problem earlier this year.

TheHollidayInn:
I did fix this a while back: https://github.com/HabitRPG/habitica/blob/develop/website/server/models/group.js#L1409
https://github.com/HabitRPG/habitica/blob/f59210375491c8a5de5cb30da53f016ba6e3e7b3/website/server/controllers/api-v3/groups.js#L574
https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/stripePayments.js#L102
https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/amazonPayments.js#L275

shanaqui:
@Alys Aha, I understand now!

[...]

PR #10449, opened 2018-06-14
Better group plan member counts

SabreCat:

Further fixage for #10124 .

### Changes

1. Adds a migration to reconcile `memberCount` with data from the user collection for all existing Group Plans.
2. Adds a step at time of group plan creation to reconcile the group's `memberCount` before setting up subscriptions.
3. Adds a server config variable for the connection string used by migrations. This should be more secure than editing connection strings directly in migration files.
4. Cleans up an old migration that had a garbage connection string.

### To Do

* [x] Run some local tests / dry runs against beta server

### DIFF
--- a/migrations/groups/reconcile-group-plan-members.js
+++ b/migrations/groups/reconcile-group-plan-members.js
@@ -1,0 +1,67 @@
+import monk from 'monk';
+import nconf from 'nconf';
+import stripePayments from '../../website/server/libs/payments/stripe';
+
+/*
+ * Ensure that group plan billing is accurate by doing the following:
+ * 1. Correct the memberCount in all paid groups whose counts are wrong
+ * 2. Update the above's subscription counts in Stripe via their API
+ */
+
+const CONNECTION_STRING = nconf.get('MIGRATION_CONNECT_STRING');
+
+let dbGroups = monk(CONNECTION_STRING).get('groups', { castIds: false });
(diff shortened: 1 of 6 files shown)


TheHollidayInn:
Woot!

migrations/groups/reconcile-group-plan-members.js
```diff
@@ -0,0 +1,67 @@
...
+          ]
+      }
+    );
+    if (group.memberCount !== canonicalMemberCount) {
+      pause();
+      return dbGroups.update(
+        {_id: group._id},
+        {$set: {memberCount: canonicalMemberCount}}
+      ).then(() => {
+        await stripePayments.chargeForAdditionalGroupMember(group);
```
TheHollidayInn: We need to check to ensure this is a group plan with Stripe. Amazon is also an option, but doesn't need a charge like this.

Alys:
Should this also fix `quantity` and use that as another check for whether Stripe needs to be updated?

SabreCat:
@Alys It does fix `quantity` via this line https://github.com/HabitRPG/habitica/blob/04d7ff13def23a4265075a903f586eb18684a099/website/server/libs/payments/stripe.js#L160, but I can definitely check to see if there are any discrepancies in `quantity` that don't correspond to discrepancies in `memberCount`!

PR #10450, opened 2018-06-15
Prevent text overflow on failed login notification (#10447)

### LINKED ISSUE #10447: Text overflow in failed login notification

### Description

When a user enters incorrect login credentials, the resulting error bubble doesn't fit the text especially well:

"Uh-oh" message spills outside of a red bubble

jrobind:
Fixes #10447 

### Changes
The error notification width has been increased slightly along with the addition of some padding and a reduction in border-radius. I've also made sure the text doesn't overflow in other languages!

Before:
screenshot 2018-06-15 13 10 26

After:
screenshot 2018-06-15 13 13 28

----
UUID: 7187b35d-3ae8-4b5c-8578-be7f582ff437

### DIFF
--- a/website/client/components/snackbars/notification.vue
+++ b/website/client/components/snackbars/notification.vue
@@ -50,6 +50,10 @@ transition(name="fade")
 
   .error {
     background-color: #f74e52;
+    border-radius: 60px;
+    width: 320px !important;
+    padding: 10px 5px;
+    margin-left: 0;
     color: #fff;
   }
 


website/client/components/snackbars/notification.vue
```diff
@@ -50,6 +50,10 @@ transition(name="fade")
 
   .error {
     background-color: #f74e52;
+    border-radius: 60px;
+    width: 320px !important;
```
paglias, line 54: Really like these changes @jrobind , is there a way to set 320px as the max-width? Also pinging @Tressley for the design

Tressley: Oof. This is an odd instance. It looks like we set the border-radius to 60px to create the pill shape for the snacks, but that doesn't hold up with larger text. Is there any way to set the border-radius to 4px when it's past two lines?

jrobind: Hey @paglias, applying max-width to the error class will not work because the width is overridden by the 300px width set by the notification class. I avoided touching the width on the notification class just in case it created problems in general for notifications. If I apply max-width of 320px on the notification class this would work however! What do you think? I'm still a bit new to all this, so apologies if I've not explained myself properly!

paglias: The 20px difference between 320 and 300 make a big difference? 

@Tressley I don't think that's possible with css, the problem seems to be that the padding is applied to the rectangular shape of the box, without considering the border radius, so when there's text near the borders it falls in the area used by the border radius.

Not sure if it could help to have a wrapper div with the border radius and then an internal div for the text

SabreCat:
Thanks for this, @jrobind! Added as credit toward contributor tier 2. It'll be merged soon!

PR #10452, opened 2018-06-16
remove unused UNSUBSCRIBED code (a `challenge.broken` and `group.broken` value)

Alys:
DO NOT MERGE YET: might need discussion; will need a data search and maybe a small migration.

This removes some incomplete code for an "UNSUBSCRIBED" value for `challenge.broken` in the task schema. 

It looks like that code might have been previously used in cases when a user had left a challenge but I don't know when - it must have been a very long time ago, if indeed it was ever fully implemented, because I don't recall seeing it used in all the time I've been troubleshooting challenge task problems.

There's no need for it anymore in normal use (when a user leaves a challenge, all tasks are immediately deleted or converted to non-challenge tasks), and if it's manually added to a challenge task by an admin, the user is still unable to delete the task since there's no enabled support  for it ("CHALLENGE_DELETED" can be used for that, which is strictly inaccurate but since it's for cases where the user is in a bug-ridden situation anyway, I don't think that matters).

Removing it from the schema could be considered a breaking change but since it's not in use in official tools, I don't see that that matters either. Third-party tools should not be using it. I checked the Android and iOS source code to ensure it's not used there.

NOTE that before (if) this is merged, we should check all tasks to ensure it's not used in any of them. I can do that.

### DIFF
--- a/website/client/components/tasks/brokenTaskModal.vue
+++ b/website/client/components/tasks/brokenTaskModal.vue
@@ -16,13 +16,6 @@
         div
           button.btn.btn-primary(@click='unlink("keep-all")') {{ $t('keepThem') }}
           button.btn.btn-danger(@click='unlink("remove-all")') {{ $t('removeThem') }}
-      // @TODO: I ported this over, but do we use it anymore?
-      //div(v-if='brokenChallengeTask.challenge.broken === "UNSUBSCRIBED"')
-        p {{ $t('unsubChallenge') }}
-        p
-          a(@click="unlink('keep-all')") {{ $t('keepThem') }}
-          | &nbsp;|&nbsp;
-          a(@click="unlink('remove-all')") {{ $t('removeThem') }}
 </template>
 
 <style scoped>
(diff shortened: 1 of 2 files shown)


paglias:
I'm not againsta removing this but before this PR is merged we should run a migration to change the tasks using it in the db or it'll break the saving for tasks with this value.

Alys:
Yes, no worries, that's the check I mentioned, I just didn't want to run it if you didn't want to merge this. :) I'll do it on the weekend if not before.

SabreCat:
1. Would it be feasible to commit the migration we'd want to run, here, for archival purposes if nothing else?
2. I recall the query on the data ended up being a resource drain when attempted, so I imagine that came out inconclusive. Can we run that test against a local backup, create a temporary index, or the like so we can try that again?

Alys:
Yeah, I'm planning to run it on a local copy of the database, I've just been waiting to visit a friend who has better internet than I have for the download. I'll close the pr until then. I don't think there's any point committing the script - it's just a batched search, not a search plus update because I'm not expecting to find more than a handful of tasks at most, probably none since the only way this value can be used is from an admin adding it - easy to fix manually if needed (although I've added it to a couple of my own tasks to be sure the search is actually working).

PR #10453, opened 2018-06-16
API v4

paglias:
As discussed internally.

Done:
- All `/api/v3` requests can be made to `/api/v4` and will work exactly the same
- Added ability to disable v3 routes in v4 and to override them with new methods
- Added test commands and folders for v4
- Renamed `test/api/v3/unit` to `test/api/unit`, same thing for the npm test command which is now `npm run test:api:unit`

Questions:
- Stop loading entire user when only needed for auth?


paglias:
Now includes #10455 as well

PR #10456, opened 2018-06-17
Client: use api v4

paglias:
To be merged after #10453 

This PR switches the client side to use `/api(v4` instead of `/api/v3` for all server requests


paglias:
Closed in favor of #10457

PR #10459, opened 2018-06-19
Remove "add multiple" tip once a task has been added (fixes #10440)

### LINKED ISSUE #10440: "add multiple" tip interferes with editing a new task
As reported by Ensorceled (4428e727-2a0e-4072-b578-fc28029b33f3):

"Is there anyway to make the [tip] for adding tasks go away once I've added a task? Or just make it go away altogether? It's very annoying that, when trying to edit an added task, I must first click to close the [tip] and then click to edit the task. ...  Here's what I mean:"

scr

I.e., when you add a task, if you want to edit it immediately (a common desire), then you have to click twice on the task to make it open in the edit window because all that the first click does is make the tip disappear. You can work around that by rapidly double-clicking the new task but that's not intuitive.

thefifthisa:

Fixes #10440

### Changes

The tip is hidden after the Enter key has been pressed and no longer prevents the user from editing a new task immediately.

**Edit**: Forgot that the Enter key is also pressed when adding multiple tasks, so the tip is hidden only after the tasks have been created.

addmultipletip

----


website/client/components/tasks/column.vue
```diff
@@ -19,7 +19,7 @@
     textarea.quick-add(
       :rows="quickAddRows",
       v-if="isUser", :placeholder="quickAddPlaceholder",
-      v-model="quickAddText", @keypress.enter="quickAdd",
+      v-model="quickAddText", @keypress.enter="quickAdd($event); quickAddFocused = false",
```
paglias: the code looks good but I think the `quickAddFocused = false` should be moved inside the quickAdd function, we try to keep code in templates to the minimum

paglias (changes requested):

website/client/components/tasks/column.vue
thefifthisa: Done!

paglias:
thanks @thefifthisa! This has been notated towards your second tier

PR #10466, opened 2018-06-22
Fix documentation about rejecting group invitation

phillipthelen:
While Trying to implement rejecting invitations in the mobile app I noticed that the documentation was not correct and the name of the route was wrong.


paglias:
thanks!

PR #10484, opened 2018-06-29
Added login path for Spritely

TheHollidayInn:


website/client/components/settings/api.vue
```diff
@@ -85,13 +85,28 @@ export default {
...
+  },
   computed: {
     ...mapState({user: 'user.data', credentials: 'credentials'}),
     apiToken () {
       return this.credentials.API_TOKEN;
     },
   },
   methods: {
+    receiveMessage (eventFrom) {
+      if (event.origin !== 'https://www.spritely.app') return;
```
paglias, line 102: should the origin also check for the url without the `www`?

TheHollidayInn: I think not, because www is forced for Spritely

PR #10493, opened 2018-07-02
Fix Modal Stack - reopening modals

### LINKED ISSUE #9986: modals reappear after being closed

negue:
fixes #9986 - use fallback target.id - the normal target is sometimes empty, which results in having the last opened modal not removed from the stack

this also removes the second scrollbar when a modal is open (only the modal itself will now scroll)

### DIFF
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -475,7 +475,6 @@ export default {
       });
 
       this.$root.$on('bv::modal::hidden', (bvEvent) => {
-
         let modalId = bvEvent.target && bvEvent.target.id;
 
         // sometimes the target isn't passed to the hidden event, fallback is the vueTarget
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -483,40 +483,24 @@ export default {
         }
 
         if (!modalId) {
-          console.error('There is no id to on this hidden event');
-          debugger;
           return;
-        } else {
-          console.warn(`Hidden: ModalId = ${modalId}`);
         }
 
         const modalStack = this.$store.state.modalStack;
 
         const modalOnTop = modalStack[modalStack.length - 1];
 
-        console.info('Stack', modalStack, 'modalOnTop', modalOnTop);
-
         // Check for invalid modal. Event systems can send multiples
-        if (!this.validStack(modalStack)) {
-          console.warn(`Hidden: ModalId = ${modalId} was invalid so isn't removed`);
-          return;
-        }
+        if (!this.validStack(modalStack)) return;
 
         // If we are moving forward
-        if (modalOnTop && modalOnTop.prev === modalId) {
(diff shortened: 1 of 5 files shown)


TheHollidayInn:
Can you post your steps to replicate? This would be a super helpful fix :D

negue:
@Alys ' Steps helped to recreate that bug:

1. Create a challenge
2. Don't add task
3. Leave the challenge, choosing the option to remove all tasks
4. The challenge creation modal reappeared automatically

SabreCat:
I'm still getting the challenge creation modal popping back up in the following flow:

1. Start with a fresh account, or at least one who has not created a challenge before.
2. Create a challenge. This will trip the achievement popup for creating a challenge.
3. Click away to dismiss the achievement popup. The challenge creation modal springs back up, despite your already being on the page for creating challenge tasks.

That's on Firefox, haven't checked with other browsers.

negue:
Its a different kind of this bug (but yeah same result), I could track it down, the `challenge-modal` isn't  removed due to this condition `modalOnTop && modalOnTop.prev === modalId`

So the achievement modal is opened while the challenge modal isn't fully closed, which results in this behavior.

Two possible solutions for this:
a) remove the condition, but this is one probably needed in different modal-queues
b) move the saving of stuff (like challenge) after the modal is closed, but that would also remove the ability to show server-side validations

any other ideas?

TheHollidayInn:
I'd like us to move the challenge achievement to a notification or a snack bar that has to be clicked to open the modal rather than just opening the modal. That should fix this behavior

negue:
@TheHollidayInn do you want to do that in a different PR or should I move it to a notification (which shows the modal then?)

TheHollidayInn:
If it's fine with you, let's just add the new notification+popup flow to this PR since it is pretty small and still relevant.

SabreCat:
I found another issue along the way https://github.com/HabitRPG/habitica/issues/10544 , but confirmed that's present in the live site. This looks good, sending it into Staging!

Issue #10379, opened 2018-05-19
Flicker when user mouses over the very leftmost edge of party member avatar


SabreCat:
[//]: # (Before logging this issue, please post to the Report a Bug guild from the Habitica website's Help menu. Most bugs can be handled quickly there. If a GitHub issue is needed, you will be advised of that by a moderator or staff member -- a player with a dark blue or purple name. It is recommended that you don't create a new issue unless advised to.)

[//]: # (Bugs in the mobile apps can also be reported there.)

[//]: # (If you have a feature request, use "Help > Request a Feature", not GitHub or the Report a Bug guild.)

[//]: # (For more guidelines see https://github.com/HabitRPG/habitica/issues/2760)

[//]: # (Fill out relevant information - UUID is found from the Habitia website at User Icon > Settings > API)
### General Info
  * UUID: Various
  * Browser: Chrome, others?
  * OS: Windows 10, others?

### Description
[//]: # (Describe bug in detail here. Include screenshots if helpful.)

There is a small area (1-2 pixels wide perhaps?) in the party header where if the user mouses over it, it flickers rapidly and unpleasantly between displaying and hiding the first party member's mouseover details.

Video: https://drive.google.com/file/d/1q7ONERmRuZXTo5VVq6GQ9sv_DHUAQSJj/view

haydenbetts:
Hi @SabreCat !

I was able to replicate this 
Browser: Chrome (it doesn't affect Firefox)
OS: MacOS 10.12

I also found a one line CSS fix that doesn't result in any noticeable change to the UI. If you assign this to me, I am ready to submit a PR! 

Video of behavior in Chrome with fix: https://drive.google.com/open?id=1FSM-PFIjGzr1yEvs6OkjQ2vcQaUKd5i6

I believe that the issue had something to do with the way Chrome processes animation effects. I think a border animation was causing the cursor to repeatedly hover an unhover an element.

paglias:
Go ahead @haydenbetts !

beffymaroo:
@haydenbetts Are you still working on this issue? Please let us know when you have a moment. Thank you!

haydenbetts:
Hi  @beffymaroo. No I am not currently working on this. I got stuck trying to write tests for my fix. The WIP PR that I submitted does contain a solution to the issue itself though!

SabreCat:
This doesn't seem like the sort of fix we'd need a test for anyway! I'll take another look at that PR.

PR #10407, opened 2018-05-28
WIP Fix flicker when user mouses over the very leftmost edge of party member avatar

haydenbetts:
Fixes #10379 

### Changes
Add one pixel of margin-right to the .member-stats class.

Video Before: https://drive.google.com/file/d/1q7ONERmRuZXTo5VVq6GQ9sv_DHUAQSJj/view
Video After: https://drive.google.com/open?id=1FSM-PFIjGzr1yEvs6OkjQ2vcQaUKd5i6

PR #10504, opened 2018-07-06
member-details: Only add 1px margin when condensed

### LINKED ISSUE #10502: Profile modal layout changed, avatar and stats now on different lines (party members, etc)
Reported by citrusella today, confirmed by me in Google Chrome:

image

### LINKED ISSUE #10510: party members list is misformatted
When you click on the "View Party" button or on the Party page's member list badge, the list of party members is no longer displayed properly. It now looks like the screenshot below. From memory, it used to be that all of the details for one member were in a single row (avatar, name, stats bars).

_Thanks to Ireanna for reporting this._

scr

aszlig:
As reported in #10502, adding a 1 pixel right margin to *all* nodes with the `member-stats` class will affect the display when you click on the avatar as well, which will then break the layout because of that additional margin.

Instead of adding the margin to `.member-stats` in general, let's just add it to the condensed version so that it really just addresses the flickering on Chrome/Chromium as reported in #10379.

I've tested whether the flickering still happens via a small `xdotool`-loop (just to make sure I don't get too shaky with the mouse):

```sh-session
$ for i in $(seq 500 508); do xdotool mousemove "$i" 250; sleep 5; done
```

The flickering doesn't happen anymore and the layout in the party members overview and the stats overview when you click on the avatar is no longer distorted.

Fixes #10502 
Fixes #10510

### DIFF
--- a/website/client/components/memberDetails.vue
+++ b/website/client/components/memberDetails.vue
@@ -48,7 +48,6 @@
   .member-stats {
     padding-left: 12px;
     padding-right: 24px;
-    margin-right: 1px;
     opacity: 1;
     transition: width 0.15s ease-out;
   }
@@ -148,6 +147,7 @@
       right: 100%;
       height: calc(100% + 18px);
       margin-top: -9px;
+      margin-right: 1px;
       padding-top: 9px;
       padding-bottom: 24px;
       padding-right: 16px;


negue:
I couldn't get the flickering on the current website either.

But the dialog layout is now fixed :)

aszlig:
@negue: The flickering is fixed since #10407, so the current website shouldn't have that flickering. It should only occur with 4.51.0 and below. This is only a fix of the layout without reintroducing the flickering.

SabreCat:
Thanks @aszlig, this will go into staging for next week's contributor PRs release!

PR #10513, opened 2018-07-09
Fix Modal positioning wrong in specific cases

### LINKED ISSUE #10475: Modal positioning wrong in specific cases
From Report a Bug:

> Charlie (41fa2ec5-0f75-4b52-9038-4209eb9fe5ac) 
> @shanaqui I worked out how to reliably reproduce the z-index issue. It looks like it is only caused by a certain set of steps. If I click help (https://habitica.com/static/faq) then the habitica logo on the menu bar then the modal always sits lower than the menu until I reload all assets (CSS, etc) with a Ctrl+F5/Ctrl+R Screenshot: https://imgur.com/a/0xa1Y88

I've confirmed this myself.

image

vinmegamitensei:

Fixes #10475

### Changes
Created a _.modal_ class and set it's properties to:
_top: 50px;_
Tested on Mozilla Firefox and Google Chrome.

**More details, please check the #10475 discussion.**

Some personal thoughts:
I couldn't find a way to give this property to all modals without duplicating this CSS class, as the original one comes from Bootstrap. Checked if it gets duplicated on production environment, and it does too. This is the simplest solution I've found, but I'll be checking here to see if anyone has a better idea. Thank you.

----
UUID: c7190c89-8947-45a5-81d5-b6946e55ff15

### DIFF
--- a/website/client/assets/scss/modal.scss
+++ b/website/client/assets/scss/modal.scss
@@ -12,8 +12,16 @@
 }
 
 .modal {
-  top: 75px;
+  top: 50px;
 }
+
+.restingInn {
+  .modal {
+    top: 90px;
+  }
+}
+
+
 
 .modal-dialog {
   width: auto;
--- a/website/client/assets/scss/modal.scss
+++ b/website/client/assets/scss/modal.scss
@@ -11,6 +11,10 @@
   }
 }
 
+.modal {
+  top: 50px;
+}
+
 .modal-dialog {
   width: auto;
 
--- a/website/client/assets/scss/modal.scss
+++ b/website/client/assets/scss/modal.scss
@@ -12,7 +12,7 @@
 }
 
 .modal {
-  top: 50px;
+  top: 75px;
 }
 
 .modal-dialog {


negue:
Could you try to increase the margin between the bottom of the navbar and and top most position of the modal to be around 48px, also keep in mind that if a user paused his damage another toolbar shows above the navbar adding additional space

vinmegamitensei:
Hey @negue , I guess setting the navbar margin won't help us in this case - as it will push the website content down, not the modal screen.
image

In any case, I'll put some work on that _"also keep in mind that if a user paused his damage another toolbar shows above the navbar adding additional space"_ part. Thank you for your feedback, hope to be back with some solution soon.

vinmegamitensei:
@negue I've changed it from 50px to 75px. This is an ok distance. But I have some things to add:
I would say we need to study a better way to apply these modals. Their maintenance is quite tiring, even for quick minor changes.

This next part may sound a little weird, as english is not my main language. Anyway, here it goes:
I wanted to make something like:

_**v-if restingBanner is active do:**
`.modal top: 75px`
**else**
`.modal top: 50px`_

But I don't know how to do it, yet. If anyone has a way to implement this idea, I'm willing to help right away. Thank you for you attention!

negue:
@viniciusperuzza You could move the class `restingInn`(https://github.com/HabitRPG/habitica/blob/develop/website/client/app.vue#L26) to an higher element and then you can use this class inside your css rules

vinmegamitensei:
@negue Thanks to your ideas and guidance, I think I achieved a satisfatory answer to this issue, with just a few changes.
Here it goes:

First, when restingInn is deactivated, we have a default modal position of 50px.
image

Then, if restingInn is active, we have a 90px distance from top:
image

This is the results, with and without the restingInn banner on top.
image
image

In my opinion, this is a better solution to the problem - in comparison with the one I proposed earlier. What do you think? I'll commit this solution to this Pull Request. Thank you in advance, this was really interesting to learn!

negue:
You need to move the `restingInn` class to an higher element, since only some modals are added inside `menu.vue` all others won't receive the new `top` 

if moving the class creates more problems you could just use a different class name which is set when `showRestingBanner` is binded


ping @Tressley

negue:
@viniciusperuzza did you try my suggestion from the last comment?

so that all modals would have the margin

TheHollidayInn:
Sorry for the late response.

Have you tried using a more specific selector? I believe every modal is a sub element of our #app ID. You should be able to override bootstrap styles by using a selector like #app .modal.

Or, you could add the ids of all the modal to the CSS selector. We should have IDs for all elements.

And finally, you can try using !important, then add a note about why you needed to override

vinmegamitensei:
Guys, I really can't help with these changes as I previously said on other topics, I'm really really busy with work (totally unexpected, sorry for this). So you guys can feel free to do any changes here or even close this PR. Again, I'm really sorry for this inconvenience.

TheHollidayInn:
Not a problem at all. I didn't see your busy message. Thanks for getting it started.

Tressley:
Just now catching the ping on this issue. Sorry for the late reply. Is there a specific design question I missed?

TheHollidayInn:
Nope, you are good now. 

Closing this PR, but feel free to reopen.

PR #10515, opened 2018-07-11
Group Tasks Shared Completion

### LINKED ISSUE #10365: Group tasks not behaving as expected

### General Info
  * UUID:  1e7fda50-e6b7-4047-9bcd-70110c713b01
  * Browser: Chrome v66
  * OS: Windows 7

### Description

Tasks which are assigned to group members do not share progress between members. Assigning a task to multiple people seems to copy the task to the assignee rather than share it. From what I can tell, this means that even if two people are assigned the same task, if one person completes the task the other person still has to complete it too. Similarly, if a task is edited from the group page the changes are not updated on the users task page. They must unclaim and reclaim a task to get updated changes.

Steps to reproduce:

1. Go to the "Task Board" for a Group Plan you own.
2. Create a task.
3. Assign it to multiple Group Plan members.
4. In each member's account, sync to make the task appear in the user's Tasks screen.
5. In one of the member's account, mark the task as complete. On the second member's account, sync, observe that the task remains for the second member and its appearance is not changed.

SabreCat:

Fixes #10365 

### Changes

Adds a new field to group plan To-Dos, "Shared Completion". This has three options:

* None (Individual Completion) - current, default behavior. Assigned members' task completions have no effect on the master group task.
* Single - When any assigned user completes the task, the master task is completed.
* All - When all assigned users have completed the task, the master task is completed.

To-Dos:

* [x] Finish tests (not currently working)
* [x] Improve labeling/help text? Consult with design/UX folks
* [x] Currently, tasks just drop off the group board. Ideally, they'd show up under "Completed" list.


website/server/controllers/api-v3/tasks.js
```diff
@@ -653,6 +656,49 @@ api.scoreTask = {
...
+    // Handle shared completion
+    if (task.type === 'todo' && direction === 'up' && task.group && task.group.sharedCompletion === 'singleCompletion') {
+      let masterTask = await Tasks.Task.findOne({
+        _id: task.group.taskId,
+      }).exec();
+
+      masterTask.completed = true;
+      promises.push(masterTask.save());
+
+      const tasksToRemove = await Tasks.Task.find({
```
paglias: can this be a remove mongodb operation directly instead of a find + remove for each individual task?

website/server/controllers/api-v3/tasks.js
```diff
@@ -653,6 +656,49 @@ api.scoreTask = {
...
+          {userId: {$exists: true}},
+          {userId: {$ne: user._id}},
+        ],
+      }).exec();
+
+      tasksToRemove.forEach(async (taskToRemove) => {
+        promises.push(taskToRemove.remove());
+      });
+    } else if (task.type === 'todo' && direction === 'up' && task.group && task.group.sharedCompletion === 'allAssignedCompletion') {
+      let allAssignedComplete = await _.reduce(task.group.assignedUsers, async (result, id) => {
```
paglias: _.reduce doesn't return a promise so I don't think this piece of code is working correctly

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -362,9 +363,44 @@ api.approveTask = {
...
-    await Promise.all(managerPromises);
+    // Handle shared completion
+    let masterTask = await Tasks.Task.findOne({
+      _id: taskId,
+    }).exec();
+
+    if (task.type === 'todo' && masterTask && masterTask.group && masterTask.group.sharedCompletion === 'singleCompletion') {
+      masterTask.completed = true;
+      approvalPromises.push(masterTask.save());
+      const tasksToRemove = await Tasks.Task.find({
```
paglias: same here remove instead of find + remove

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -362,9 +363,44 @@ api.approveTask = {
...
 
-    managerPromises.push(task.save());
-    managerPromises.push(assignedUser.save());
-    await Promise.all(managerPromises);
+    // Handle shared completion
+    let masterTask = await Tasks.Task.findOne({
+      _id: taskId,
+    }).exec();
+
+    if (task.type === 'todo' && masterTask && masterTask.group && masterTask.group.sharedCompletion === 'singleCompletion') {
```
paglias: if this code is the same as `website/server/controllers/api-v3/tasks.js` maybe it should be moved to a library file

paglias (dismissed):

website/server/libs/groupTasks.js
```diff
@@ -0,0 +1,46 @@
...
+      return masterTask;
+    } else {
+      return undefined;
+    }
+  } else {
+    return undefined;
+  }
+}
+
+export {handleSharedCompletion};
```
paglias: why not export the function directly at line 3? so `export async function...`

website/server/libs/groupTasks.js
```diff
@@ -0,0 +1,46 @@
...
+
+async function handleSharedCompletion (groupMemberTask) {
+  let masterTask = await Tasks.Task.findOne({
+    _id: groupMemberTask.group.taskId,
+  }).exec();
+
+  if (!masterTask || !masterTask.group || masterTask.type !== 'todo') return undefined;
+
+  if (masterTask.group.sharedCompletion === 'singleCompletion') {
+    await Tasks.Task.deleteMany({
```
paglias: let's make sure all these queries are backed by indexes

website/client/components/tasks/taskModal.vue
```diff
@@ -691,6 +700,7 @@ export default {
         calendar: calendarIcon,
       }),
       requiresApproval: false, // We can't set task.group fields so we use this field to toggle
+      sharedCompletion: 'individualCompletion',
```
TheHollidayInn: From the tests, `individualCompletion` is not an option.

website/client/components/tasks/taskModal.vue
```diff
@@ -896,6 +907,8 @@ export default {
         this.task.group.assignedUsers = this.assignedMembers;
```
TheHollidayInn, line 908: Let's make a todo here to flatten out group tasks and add correct permissions so we don't have to do this hack. We will want to do this soon.

website/server/controllers/api-v3/tasks.js
```diff
@@ -5,6 +5,7 @@ import {
 } from '../../libs/webhook';
 import { removeFromArray } from '../../libs/collectionManipulators';
 import * as Tasks from '../../models/task';
+import * as groupTasks from '../../libs/groupTasks';
```
TheHollidayInn: Change these to only import what you need. So this should be `import { handleSharedCompletion } from  '../../libs/groupTasks';`

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -362,9 +365,12 @@ api.approveTask = {
       direction,
     });
 
-    managerPromises.push(task.save());
-    managerPromises.push(assignedUser.save());
-    await Promise.all(managerPromises);
+    let masterTask = await groupTasks.handleSharedCompletion(task);
+    if (masterTask) approvalPromises.push(masterTask.save());
```
TheHollidayInn: If you are awaiting above, then you do not need to push the promise because it will be completed already. I don't think we want to run `handleSharedCompletion` in parallel with `task.save`, so we should just remove pushing it, and only await it.

website/server/libs/groupTasks.js
```diff
@@ -0,0 +1,46 @@
...
+      }).exec();
+    }
+    if (completions >= masterTask.group.assignedUsers.length) {
+      masterTask.completed = true;
+      return masterTask;
+    } else {
+      return undefined;
+    }
+  } else {
+    return undefined;
```
TheHollidayInn: We won't need to return undefined

website/server/libs/groupTasks.js
```diff
@@ -0,0 +1,46 @@
+import * as Tasks from '../models/task';
+
+async function handleSharedCompletion (groupMemberTask) {
+  let masterTask = await Tasks.Task.findOne({
+    _id: groupMemberTask.group.taskId,
+  }).exec();
+
+  if (!masterTask || !masterTask.group || masterTask.type !== 'todo') return undefined;
+
+  if (masterTask.group.sharedCompletion === 'singleCompletion') {
```
TheHollidayInn: Let's break the code in both of these if branches into functions. 

`if (masterTask.group.sharedCompletion === 'singleCompletion') {` everything under this should be a function

`} else if (masterTask.group.sharedCompletion === 'allAssignedCompletion') {` everything under this should be in a function.

website/server/libs/groupTasks.js
TheHollidayInn: My personal preference is to have the functions at the bottom so you can quickly read the "api" of the module. Similar to reading an interface in OO

website/server/libs/taskManager.js
```diff
@@ -96,6 +96,7 @@ export async function createTasks (req, res, options = {}) {
       if (taskData.requiresApproval) {
         newTask.group.approval.required = true;
       }
+      newTask.group.sharedCompletion = taskData.sharedCompletion || 'individualCompletion';
```
TheHollidayInn: Ensure individualCompletion is an option. Also, let's move all these string constants to our groupTasks Library.

website/server/models/task.js
```diff
@@ -111,6 +111,7 @@ export let TaskSchema = new Schema({
       requested: {type: Boolean, default: false},
       requestedDate: {type: Date},
     },
+    sharedCompletion: {type: String, enum: ['individualCompletion', 'singleCompletion', 'allAssignedCompletion'], default: 'individualCompletion'},
```
TheHollidayInn: Let's have these string constants come from our groupTasks lib.

website/server/libs/groupTasks.js
SabreCat: It'll be more relevant to readability as we expand the library!

website/client/components/tasks/taskModal.vue
SabreCat: Can you elaborate? Do you mean I should add a test that the behavior stays the same (i.e., that the master task is never completed) while it's set to `individualCompletion`?

website/server/libs/groupTasks.js
SabreCat: https://github.com/HabitRPG/habitica/blob/48bcc1e968264e9e41d7074b14dc33587624a95d/migrations/docs/mongo-indexes.md#tasks should cover it, yes? `group.taskId` and `group.userId`

SabreCat: I suppose `group.approval.approved` isn't indexed, nor `completed`. We can add those if you think it'd be valuable! But I'm only querying that in conjunction with `group.taskId` so I thought it'd be a pretty narrow pool to begin with.

website/client/components/tasks/taskModal.vue
TheHollidayInn: Ohh. Right. How about we rename it to recurringCompletion?

website/server/libs/groupTasks.js
```diff
@@ -1,46 +1,60 @@
...
-    if (completions >= masterTask.group.assignedUsers.length) {
-      masterTask.completed = true;
-      return masterTask;
-    } else {
-      return undefined;
-    }
-  } else {
-    return undefined;
+  if (masterTask.group.sharedCompletion === SHARED_COMPLETION.single) {
+    _deleteUnfinishedTasks(groupMemberTask);
```
paglias: these functions are async, shouldn't they be awaited?

PR #10524, opened 2018-07-15
prevent multiple notifications

### LINKED ISSUE #10356: duplicate/multiple notifications for new guild/party messages
As reported by Gumnos (43e2413a-f440-43c8-b96c-85b93024488c): 

"Until recently, I'd only get one notification per guild/party (3 guilds + 1 party meant usually no more than 4 notifications unless there was the occasional extra stat-point to allocate, a quest notification, or a DM).  Sometimes I'd hit a glitch and it would give me one duplicate, but refreshing my browser always fixed/consolidated duplicates down to a single instance.

Today I encountered multiple notifications for the same guild but it wouldn't compact them when I'd refresh. That said, clicking on one of the notifications cleared all of the duplicates."

screenshot

negue:
fixes #10356

the added test `Prevent multiple notifications` worked even without adding the `notifactionAlreadyExists`-check in `addUserNotification`

@paglias @SabreCat any idea on how to reproduce the issue ?
Should the post messages be fired at the same time instead of waiting of prior one?

### DIFF
--- a/test/api/v3/integration/notifications/prevent-multiple-notification.js
+++ b/test/api/v3/integration/notifications/prevent-multiple-notification.js
@@ -2,7 +2,7 @@ import {
   createAndPopulateGroup,
 } from '../../../../helpers/api-integration/v3';
 
-describe.only('Prevent multiple notifications', () => {
+describe('Prevent multiple notifications', () => {
   let partyLeader, partyMembers, party;
 
   before(async () => {
@@ -22,12 +22,13 @@ describe.only('Prevent multiple notifications', () => {
   it('does not add the same notification twice', async () => {
     const multipleChatMessages = [];
 
-    multipleChatMessages.push(
-      partyMembers[0].post(`/groups/${party._id}/chat`, { message: 'Message 1'}),
-      partyMembers[1].post(`/groups/${party._id}/chat`, { message: 'Message 2'}),
-      partyMembers[2].post(`/groups/${party._id}/chat`, { message: 'Message 3'}),
-      partyMembers[3].post(`/groups/${party._id}/chat`, { message: 'Message 4'})
-    );
+    for (let i = 0; i < 4; i++) {
+      for (let memberIndex = 0; memberIndex < partyMembers.length; memberIndex++) {
+        multipleChatMessages.push(
+          partyMembers[memberIndex].post(`/groups/${party._id}/chat`, { message: `Message ${i}_${memberIndex}`}),
+        );
+      }
+    }
 
     await Promise.all(multipleChatMessages);
 
--- a/test/api/v3/integration/notifications/prevent-multiple-notification.js
+++ b/test/api/v3/integration/notifications/prevent-multiple-notification.js
@@ -1,0 +1,33 @@
+import {
+  createAndPopulateGroup,
+} from '../../../../helpers/api-integration/v3';
+
+describe('Prevent multiple notifications', () => {
+  let partyLeader, partyMembers, party;
+
+  before(async () => {
+    let { group, groupLeader, members } = await createAndPopulateGroup({
+      groupDetails: {
+        type: 'party',
+        privacy: 'private',
+      },
+      members: 4,
+    });
+
+    party = group;
+    partyLeader = groupLeader;
+    partyMembers = members;
+  });
+
+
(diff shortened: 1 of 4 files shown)


test/api/v3/integration/notifications/prevent-multiple-notification.js
```diff
@@ -0,0 +1,33 @@
...
+    });
+
+    party = group;
+    partyLeader = groupLeader;
+    partyMembers = members;
+  });
+
+
+  it('does not add the same notification twice', async () => {
+    await partyMembers[0].post(`/groups/${party._id}/chat`, { message: 'Message 1'});
```
paglias: yeah, multiple messages should be posted concurrently

website/server/models/user/methods.js
```diff
@@ -150,11 +150,17 @@ schema.methods.sendMessage = async function sendMessage (userToReceiveMessage, o
  * @param  seen  If the notification should be marked as seen
  */
 schema.methods.addNotification = function addUserNotification (type, data = {}, seen = false) {
-  this.notifications.push({
-    type,
-    data,
-    seen,
-  });
+  // only check if chat notification exists
+  const notifactionAlreadyExists = type === 'NEW_CHAT_MESSAGE' ?
```
paglias: I don't think this will make any difference because chat notifications are added directly to the db using a different method https://github.com/HabitRPG/habitica/blob/develop/website/server/models/group.js#L603

paglias (changes requested):

website/server/models/user/methods.js
negue: oh wow o.O this place I didn't even saw at all ..., well gonna try to fix it there

test/api/v3/integration/notifications/prevent-multiple-notification.js
negue:
```
 const multipleChatMessages = [];

    multipleChatMessages.push(
      partyMembers[0].post(`/groups/${party._id}/chat`, { message: 'Message 1'}),
      partyMembers[1].post(`/groups/${party._id}/chat`, { message: 'Message 2'}),
      partyMembers[2].post(`/groups/${party._id}/chat`, { message: 'Message 3'}),
      partyMembers[3].post(`/groups/${party._id}/chat`, { message: 'Message 4'})
    );

    await Promise.all(multipleChatMessages);

```

won't change the result, its still showing only one notification 

Any other idea to create a test for this?

website/server/models/user/methods.js
negue: is there a way to have a check before `$push` is called?

test/api/v3/integration/notifications/prevent-multiple-notification.js
paglias: hmm no idea, sorry, I think it might only be reproducible in production where there's more latency

website/server/models/user/methods.js
paglias: what type of check? You can make another call to the db but won't change anything as it's without transactions anyway

negue: for this: `User.update(query, lastSeenUpdateAddNew, {multi: true}).exec();` 

I'd like to add a check if this notification already added, or would it be ok to go through each user and do the check+ add of notification instead of the

paglias: Not that I know and tbh it's not something so serious to require an extra query imho, also because if you add a second query you can still have the same problems as now

website/server/models/group.js
```diff
@@ -609,7 +609,31 @@ schema.methods.sendChat = function sendChat (message, user, metaData) {
   };
 
   User.update(query, lastSeenUpdateRemoveOld, {multi: true}).exec().then(() => {
-    User.update(query, lastSeenUpdateAddNew, {multi: true}).exec();
+    User.find(query).exec(async (err, users) => {
```
paglias: I don't think that this will solve the problem completely because not having transactions the problem will always remain unless we can fit the check in the update query... anyway fetching the user and saving it is too slow to do it every time a notification is added to an user

paglias:
I think the only option we have is to add the check here https://github.com/HabitRPG/habitica/blob/develop/website/server/models/userNotification.js#L74

PR #10525, opened 2018-07-15
FIX: Challenge Creation without Group not found error

### LINKED ISSUE #10228: creating a challenge when not in a party shows you the "Group not found or you don't have access" error
If you are not in a party and you start to create a new challenge, as soon as the challenge creation form opens, you see the error message "Group not found or you don't have access" in a red snackbar notification. Server logs show that the error is being thrown from the `GET /api/v3/groups/party` command (which needs to be run to find the name of your party for the case when you are in one). 

That error message should not be visible to the user. Either the code should suppress the user-visible notification or it should check if the user has a value for `party._id` before trying to do the GET.

~I'll leave this as suggestion-discussion for a few days for staff to weigh in on which fix they think is best.~

This isn't just a cosmetic error. It stops the player from creating "Public Challenges" (i.e., Tavern challenges to use the old terminology) because the "Public Challenges" entry doesn't appear in the "Add to" drop-down after the party error is thrown.

negue:
prevent loading the party, if the user isn't part of one

fixes #10228

### DIFF
--- a/website/client/components/challenges/challengeModal.vue
+++ b/website/client/components/challenges/challengeModal.vue
@@ -318,7 +318,7 @@ export default {
     async shown () {
       this.groups = await this.$store.dispatch('guilds:getMyGuilds');
 
-      if (this.user.party) {
+      if (this.user.party && this.user.party._id) {
         await this.$store.dispatch('party:getParty');
         const party = this.$store.state.party.data;
         if (party._id) {
--- a/website/client/components/challenges/challengeModal.vue
+++ b/website/client/components/challenges/challengeModal.vue
@@ -317,14 +317,17 @@ export default {
   methods: {
     async shown () {
       this.groups = await this.$store.dispatch('guilds:getMyGuilds');
-      await this.$store.dispatch('party:getParty');
-      const party = this.$store.state.party.data;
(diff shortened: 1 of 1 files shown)


website/client/components/challenges/challengeModal.vue
```diff
@@ -317,14 +317,17 @@ export default {
...
-      await this.$store.dispatch('party:getParty');
-      const party = this.$store.state.party.data;
-      if (party._id) {
-        this.groups.push({
-          name: party.name,
-          _id: party._id,
-          privacy: 'private',
-        });
+
+      if (this.user.party) {
```
paglias: I think this should also check `this.user.party._id` as `user.party` always exist

paglias: I think this should also check this.user.party._id as user.party always exist

paglias (changes requested):

website/client/components/challenges/challengeModal.vue
negue: it didn't exist for my test user, or it did exist but it never passed the if statement, I'll add the _id check

paglias (approved):

PR #10532, opened 2018-07-19
Re-enable start quest button

### LINKED ISSUE #9701: incorrect active/disabled behaviour for "invite party to quest" button
As reported by Conduit (a2eed704-33d9-4c2f-955c-f24af6653c72) in the Report a Bug guild:

"On the inventory page, when selecting a quest, the 'Invite Party to Quest' button is active on the quest overlay, even if the player does not meet the quest's eligibility criteria.
When clicking the Invite button a small, red message appears in the top-right corner to inform the player they do not meet the criteria and the Invite button greys-out.
However, selecting any other quest, or escaping the quest overlay (via keyboard or off-overlay click), leaves the invite button greyed-out until either selecting a different page and returning or reloading; the red eligibility message does not reappear when clicking the greyed-out button.
Presumably the button should recolour whenever a different quest is selected or, better yet, simply be tied to the player eligibility."

probaton:

Fixes #9701

### Changes
- Expanded the flow that determines which quest is pre-selected in the start quest modal. Instead of blindly selecting the first quest in the list, it now selects the first quest the user _has a scroll for_.
- Added a try/catch around the invite-group-to-quest API call so that the invite button is re-enabled if the call errors. 

----

### DIFF
--- a/website/client/components/groups/group.vue
+++ b/website/client/components/groups/group.vue
@@ -546,7 +546,7 @@ export default {
         eventLabel: 'Start a Quest',
       });
 
-      let hasQuests = find(this.user.items.quests, (quest) => {
+      let hasQuests = this.user.items.quests.some((quest) => {
         return quest > 0;
       });
 
@@ -554,7 +554,6 @@ export default {
         this.$root.$emit('bv::show::modal', 'start-quest-modal');
         return;
       }
-      // $rootScope.$state.go('options.inventory.quests');
     },
     showGroupGems () {
       this.$root.$emit('bv::show::modal', 'group-gems-modal');
--- a/website/client/components/groups/group.vue
+++ b/website/client/components/groups/group.vue
@@ -538,23 +538,6 @@ export default {
       this.$store.state.upgradingGroup = this.group;
       this.$router.push('/group-plans');
     },
-    clickStartQuest () {
-      Analytics.track({
-        hitType: 'event',
-        eventCategory: 'button',
-        eventAction: 'click',
-        eventLabel: 'Start a Quest',
-      });
-
-      let hasQuests = this.user.items.quests.some((quest) => {
-        return quest > 0;
-      });
-
-      if (hasQuests) {
-        this.$root.$emit('bv::show::modal', 'start-quest-modal');
-        return;
-      }
-    },
     showGroupGems () {
       this.$root.$emit('bv::show::modal', 'group-gems-modal');
     },
--- a/website/client/components/groups/startQuestModal.vue
+++ b/website/client/components/groups/startQuestModal.vue
@@ -182,13 +182,14 @@ export default {
       let groupId = this.group._id || this.user.party._id;
 
       const key = this.selectedQuest;
-      const response = await this.$store.dispatch('guilds:inviteToQuest', {groupId, key});
-      const quest = response.data.data;
+      try {
+        const response = await this.$store.dispatch('guilds:inviteToQuest', {groupId, key});
+        const quest = response.data.data;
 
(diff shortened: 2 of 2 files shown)


website/client/components/groups/startQuestModal.vue
```diff
@@ -177,13 +182,14 @@ export default {
...
-      if (this.$store.state.party.data) this.$store.state.party.data.quest = quest;
-
-      this.loading = false;
-
+      try {
+        const response = await this.$store.dispatch('guilds:inviteToQuest', {groupId, key});
+        const quest = response.data.data;
+
+        if (this.$store.state.party.data) this.$store.state.party.data.quest = quest;
+      } finally {
```
paglias, line 190: is the finally necessary? `this.loading = false;
 ` wouldn't be executed anyway if you move it outside the try/finally block like the $root.$emit below?

probaton: `guilds:inviteToQuest` with an invalid quest key throws an exception. In the old version, that would abort the function, which means `this.loading = false` (as well as the `emit`) was never executed. As the `disabled` property of the start-quest button is dependent on `this.loading`, that meant the button would never re-enable, even though the function that was "loading" had long since returned. 

By adding the `finally`, the app will always snap out of its loading state once the call returns, regardless of the outcome. The error will still throw, so the user will get the feedback they need to fix the problem, but that won't prevent the user from trying the button again. 

I left the `emit` outside the `finally` because, unlike the `this.loading` line, you actually want to prevent the pop-up from closing if the start-quest flow errors. In pretty much every scenario I could think of, users will want to try again with a another quest if the invitation fails, so closing the dialog would be confusing and annoying. 

An error in the call will now result in both an open dialog and an enabled button, so the user can immediately select and start a different quest. 

edit: Rephrased for clarity. I hope.

paglias: Ok, but if you remove the code from the final block and move it outside, if an error happens in the try block it won't throw so it'll keep going, right?

```
try {
...
} catch (e) {} // empty catch block

... other code ...

```

anyway, not a big problem, I was worried about finally not being supported but looks like it's not a problem

probaton: Aaaah, that makes more sense. I thought you meant removing it altogether, but you're talking about replacing `finally` with `catch`. You're right that either will work fine. I suppose I picked this route because the split between `this.loading` (which you always want to trigger) and `emit` (which you only want to trigger on a successful call) was more obvious with this implementation.

website/client/components/groups/group.vue
```diff
@@ -546,15 +546,14 @@ export default {
         eventLabel: 'Start a Quest',
       });
 
-      let hasQuests = find(this.user.items.quests, (quest) => {
+      let hasQuests = this.user.items.quests.some((quest) => {
```
paglias: `user.items.quests` is an object, not an array, so `some` isn't available as a methods

paglias (changes requested):

website/client/components/groups/group.vue
probaton: I was arguing my way through the best solution when I found out the entire method isn't actually used. Apparently whoever extracted the quest sidebar from the group page forgot to remove it. 

Have to say the result for users that have no quests is sort of ugly, though. The quest modal would definitely benefit from a default image, maybe a big question mark? Anyway, feels out of scope, so I just removed the function and everything seems dandy.

paglias:
thanks @probaton , will be merged later this week!

paglias:
welcome to tier 3 @probaton !

probaton:
Woohoo! Now to rub it in all my friends' faces.

PR #10540, opened 2018-07-20
Reset gift message

### LINKED ISSUE #9187: Gifting message is repeated
When you send a gift with a message, if you go to send another gift to a different user, the same message is already in the modal and the same gem amount is already selected. You should start with a blank message box and gem amount.

TheHollidayInn:
Fixes #9187

### DIFF
--- a/website/client/components/payments/sendGemsModal.vue
+++ b/website/client/components/payments/sendGemsModal.vue
@@ -1,5 +1,5 @@
 <template lang="pug">
-b-modal#send-gems(:title="title", :hide-footer="true", size='lg')
+b-modal#send-gems(:title="title", :hide-footer="true", size='lg', @hide='onHide()')
   .modal-body(v-if='userReceivingGems')
     .panel.panel-default(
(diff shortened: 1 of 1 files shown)


SabreCat:
Aww, sadness. I found this bug useful :(

TheHollidayInn:
Lol are you mass gifting?

SabreCat:
For the monthly Take This challenges, I gift 2 Gems to each of the Challenge runners-up. Saves a bit of time and copy-pasting!

Alys:
@SabreCat : cTheDragons's [Challenge Data Tool](https://oldgods.net/habitica/cTheDragons/challenge.html) lets you bulk-send gems with a message. Load the challenge data, select all the players you want to give gems to, click "Gift Gems"; it will let you enter the number of Gems and then a message. It will be slow to load all participants for a big challenge, but you can leave it ticking away in the background while it does that.

SabreCat:
This appears to reset the message but not the Gem amount spinner.

PR #10542, opened 2018-07-20
Removed duplicate code and added member modal event

TheHollidayInn:


website/client/components/groups/membersModal.vue
```diff
@@ -363,6 +376,7 @@ export default {
         });
         this.invites = invites;
       }
+      console.log(this.$store.state.memberModalOptions.viewingMembers)
```
paglias: extra console.log

paglias (changes requested):
one extra console.log

website/client/components/groups/membersModal.vue
paglias: Apart from this 👍

PR #10546, opened 2018-07-21
Added user field limit to some routes

TheHollidayInn:


website/server/controllers/api-v3/tasks.js
```diff
@@ -287,7 +287,7 @@ api.getUserTasks = {
   method: 'GET',
   url: '/tasks/user',
   middlewares: [authWithHeaders({
-    userFieldsToExclude: ['inbox'],
+    userFieldsToInclude: ['_id'],
```
Alys: Just checking... Getting tasks involves ordering them based on user.tasksOrder and also calls `setNextDue` which uses user.preferences for date-related fields. This new limitation won't affect that will it?

TheHollidayInn: You are right and the tests caught this. I'll push those and wait for another travis build. I'll most likely revert this route before pushing anything live.

PR #10547, opened 2018-07-22
Esc only closes active tags box, not entire task

thefifthisa:
[//]: # (Note: See http://habitica.wikia.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API for more info)

[//]: # (Put Issue # here, if applicable. This will automatically close the issue if your PR is merged in)
Fixes #10520 

### Changes
[//]: # (Describe the changes that were made in detail here. Include pictures if necessary)

Pressing the escape key closes only the tags box if it's open, and the entire task if it's not.

[//]: # (Put User ID in here - found on the Habitica website at User Icon > Settings > API)

----
UUID: 9ca2bb65-ce6c-4e4c-a367-5ef9b5c7421e

PR #10550, opened 2018-07-22
Clicking out only closes active tags box, not entire task

thefifthisa:

### Changes

Similar to #10547, clicking out closes only the tags box if it's open, and the entire task if it's not.

~However, this fix only works when you click somewhere in the background, not within the task itself.~

----

### DIFF
--- a/website/client/components/tasks/taskModal.vue
+++ b/website/client/components/tasks/taskModal.vue
@@ -1,7 +1,7 @@
 <template lang="pug">
-  form(v-if="task", @submit.stop.prevent="submit()")
+  form(v-if="task", @submit.stop.prevent="submit()", @click="handleClick($event)")
     b-modal#task-modal(v-bind:no-close-on-backdrop="showTagsSelect", size="sm", @hidden="onClose()", @show="handleOpen()", @shown="focusInput()")
-      .task-modal-header(slot="modal-header", :class="cssClass('bg')")
+      .task-modal-header(slot="modal-header", :class="cssClass('bg')", @click="handleClick($event)")
         .clearfix
           h1.float-left {{ title }}
           .float-right.d-flex.align-items-center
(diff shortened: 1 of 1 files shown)


paglias:
@thefifthisa I think this one is pretty similar so it might be helpful https://github.com/HabitRPG/habitica/blob/develop/website/client/components/ui/customMenuDropdown.vue#L77

If you are going to reuse the same code, let's move this to a mixin/directive if possible

thefifthisa:
@paglias Thanks for that! I adapted it and now the tags box also closes if you click anywhere else in the task—except the space beside the Tags title, which is part of the `tagsBox` area that's considered valid to click on in the `handleClick` method.

Do you think it's okay? I'm still trying out separate `ref` attributes to exclude that space, but this seems to be the simplest way to listen for outside clicks so far.

paglias:
@thefifthisa it's possible to add a new ref on the popup only, maybe here https://github.com/HabitRPG/habitica/pull/10550/files#diff-e5a5e507174ee52bb57f5acd2e857a15L150? If it's not too difficult I'd like to exclude the Tags title

website/client/components/tasks/taskModal.vue
```diff
@@ -988,6 +988,11 @@ export default {
     focusInput () {
       this.$refs.inputToFocus.focus();
     },
+    handleClick (e) {
+      if (this.$refs.popup && !this.$refs.popup.$el.contains(e.target)) {
```
negue, line 992: like in #10547 wouldn't be a `showTagsSelect` enough to check if its opened ?

paglias: in this case we want to check if the click was inside or outside the popup not only if it was open so this looks good

paglias:
Thanks for all these PRs @thefifthisa , welcome to tier 3!

PR #10552, opened 2018-07-24
Fix drag scrolling tasks when character has abilities

texastoland:
The default scroll sensitivity of the task columns is 30 px from the bottom of the screen. The collapsed spells drawer renders 32 px from the bottom of the screen intercepting most drag events. Increases the scroll sensitivity to double the height of the blocking element.

Also ignores the Yarn lockfile. `yarn --ignore-engines` builds successfully and tests pass with Node 10.6.0 and MongoDB 4.0.0.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ dist-client
 test/client/unit/coverage
 test/client/e2e/reports
 test/client-old/spec/mocks/translations.js
+yarn.lock
 
 # Elastic Beanstalk Files
 .elasticbeanstalk/*
--- a/website/client/components/tasks/column.vue
+++ b/website/client/components/tasks/column.vue
@@ -37,7 +37,7 @@
     draggable.sortable-tasks(
       ref="tasksList",
       @update='taskSorted',
-      :options='{disabled: activeFilter.label === "scheduled"}',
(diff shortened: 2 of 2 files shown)


texastoland:
@TheHollidayInn this is one I mentioned on Trello super straightforward.

negue:
This fixes the dragging, good catch 🎉 

As for the yarn lockfile, I'd say its ok, since we're using npm self so we shouldn't have to create both lockfiles all the time.

The lockfiles usually used for production environments to ensure the correct packages are used during build

texastoland:
I don't mind removing it.

paglias:
thanks @texastoland ! If you have an Habitica UUID I'll be happy to award you a contributor tier

texastoland:
@paglias thanks for the reminder. I made the PR inside GitKraken so no template. UUID: 5e965e84-1343-4a7c-8f6f-148ec27dee50.

paglias:
notated towards your next tier :)

texastoland:
@paglias will I need multiple PRs for the next one?

paglias:
@texastoland yep! each contributor tier is more difficult to obtain than the previous one http://habitica.wikia.com/wiki/Contributor_Rewards

Issue #10298, opened 2018-04-29
Death modal appears multiple times when dying after Record Yesterdays Activites


Alys:
Embrace (9030eb72-300a-4d82-997d-8094504a3c99) reports: 

**"When you die, it's super buggy. The modal div appears and then instantly goes away, and then appears and then instantly goes away, and then appears again. On the third time it finally stays put like it's supposed to."**

**"It's happening when I die from not completing dailies on the previous day. The modal pops up asking which ones you completed, and if I click 'Start my new day' and yesterday's uncompleted dailies then kill me, the 'you're dead' modal (paraphrasing) flashes on and off the screen and the sound plays multiple times."**

I was able to replicate it although the modal only appeared twice in my case.

haydenbetts:
@Alys 

I tested with:
Browser: Chrome 66.0.3359.181 (Official Build) (64-bit)
OS: 10.12.6
Local Habitica installation

I was not able to replicate. I couldn't replicate this behavior with either the death modal that appears after you do not complete dailies or the regular death modal. 

Do you have replication tips?

TheHollidayInn:
These things have been hard to replicate for me as well, but we do want to move to a new UX. -
- Basically, when you die, there should be a snackbar (top right notification like xp and gold) that does not disappear until clicked
- Then user can then click that notification to open the modal

That will prevent the death modal from going on our stack and causing the above issue. (we also want to do this with many other modals)

So, if you would like to move the death modal to that flow, we would be happy to accept a PR :D

haydenbetts:
Ah got it. Ya you can assign this to me.

TheHollidayInn:
Thanks :D

beffymaroo:
@haydenbetts Are you still working on this issue? Please let us know when you have a moment. Thanks!

haydenbetts:
@beffymaroo No I am not currently working on this issue.

beffymaroo:
@haydenbetts Thank you for the update.

TheHollidayInn:
@negue can you apply the same technique from: https://github.com/HabitRPG/habitica/pull/10493 to to the Death modal?

negue:
On it

negue:
@TheHollidayInn If I move it to a notification (without a timeout) then its maybe not that visible to a user instead of a modal, plus all of the website functionality itself still "works" even though the user isn't revived yet.

Should I put it on "readonly" mode or something else until a user is revived? 
[...]

PR #10564, opened 2018-07-28
move modals to notifications (to open the modals)

negue:
Move Modals to Notifications (which onClick shows the Modal):

- [x] Challenge Joined Achievement - **Looks good!**
- [x] Ultimate Gear Achievement - **Visual issue, see comment**
- [x] Rebirth Achievement - **Functional issue, see comment**
- [x] Guild Joined Achievement - **Looks good!**
- [x] Invited Friend Achievement - **Looks good!**
- [x] New Contributor Level - **Looks good!**
- [x] Debounced Checks (before: multiple modals were opened in short time) - **Not able to reproduce any oddity here, seems OK**
- ~~Death Modal (#10298)~~
  - ~~Auto Revive~~

Any other Modals that should be moved to a notification ?


website/client/components/notifications.vue
```diff
@@ -313,6 +346,34 @@ export default {
...
+      if (forceToModal) {
+        this.$root.$emit('bv::show::modal', config.modalId);
+      } else {
+        this.text(config.label(this.$t), () => {
+          this.$root.$emit('bv::show::modal', config.modalId);
+        }, false);
+      }
+    },
+    debounceCheckUserAchievements: debounce(function debounceCheck () {
+      this.checkUserAchievements();
```
paglias, line 386: why is this debounced?

negue: the checkUserAchievements was called multiple times, so multiple dialogs opened + sound was played so to prevent that I debounced the call

SabreCat:
image

Ultimate Gear achievement growl is awkward on two fronts:

1. Half the text falls outside of the bubble.
2. It says "You have attained the following complete sets." with no follow-up saying what sets have been attained.

I'd probably cut the "You have attained" sentence from the growl (leave it in the modal), and expand the bubble if possible.

SabreCat:
Using an Orb of Rebirth hard refreshes the page, causing the achievement notification to be briefly visible, but it vanishes before the user has a chance to read and click.

negue:
@SabreCat should the refresh be executed like X seconds later ? I don't think it'll work better when its only refreshed if the user clicks on the notification

SabreCat:
@negue I'm not sure what the best solution would be... if it's on a delay, the user clicks the notification and gets a modal, will the refresh then kick in while they're trying to read the modal?

negue:
We could just:
- trigger it on the server 
- show the notification
- if the user hasn't opened the notification in 60seconds then we just refresh
- if it was opened then we'll refresh it on leaving the modal or on the refresh button in the modal

SabreCat:
@negue OK, sounds like a plan! Is that in scope to do here?

paglias:
took a while but it's merged!

Alys:
I've marked this as notify helpers so that we can tell the Socialites if this isn't mentioned in a Bailey.

SabreCat:
@Alys OK! Is there something Bailey-worthy here? I wouldn't expect such a minor change in workflow to merit an announcement...

Alys:
I'm not saying there should be a Bailey, I'm just saying if there isn't, I want to tell the Socialites about the change so that they know about it in case they see questions. Also the wiki will need changes; for example, the page about Rebirth currently says there's a popup.

SabreCat:
Ok, makes sense!

PR #10565, opened 2018-07-30
Added context message to streak achievement notification

### LINKED ISSUE #10014: Streak achievement notfications are displaying like streak bonuses
EDIT by Alys 2018-07-29:

See [this comment](https://github.com/HabitRPG/habitica/issues/10014#issuecomment-408665995) for the current status of this bug.

---

### General Info
  * UUID: 2d6ef231-50b4-4a22-90e7-45eb97147a2c
  * Browser: Firefox ESR 52.6.0 (most up-to-date ESR version)
  * OS: Windows 7 SP1

### Description
image of notifications in vertical order ... You gained some gold: +1.06, Streak Bonus!: +0.43, Streak Bonus!: 864, and +1.0 damage to boss

Alys asked me to write this here: Streak achievements (21 day multiples of one streak) are displaying in the notifications as if they're streak bonuses: the "message" says "Streak Bonus!" and it displays the number of total streaks you have next to a gold icon (though without any sort of plus or minus indicator like an actual gold increase notification would have). That's misleading at best, since it's not a gold increase or decrease notification, it's just telling you about a number (and it's also not a streak bonus, it's a streak achievement, so that's incorrect too).

TheHollidayInn:
Fixes #10014

### DIFF
--- a/website/client/components/notifications.vue
+++ b/website/client/components/notifications.vue
@@ -465,7 +465,7 @@ export default {
             this.$root.$emit('bv::show::modal', 'won-challenge');
             break;
           case 'STREAK_ACHIEVEMENT':
-            this.text(`${this.$t('currentStreak')}${this.user.achievements.streak}`);
+            this.text(`${this.$t('streaks')}: ${this.user.achievements.streak}`);
             this.playSound('Achievement_Unlocked');
             if (!this.user.preferences.suppressModals.streak) {
               this.$root.$emit('bv::show::modal', 'streak');
--- a/website/client/components/notifications.vue
+++ b/website/client/components/notifications.vue
@@ -465,7 +465,7 @@ export default {
             this.$root.$emit('bv::show::modal', 'won-challenge');
             break;
           case 'STREAK_ACHIEVEMENT':
-            this.text(this.user.achievements.streak);
+            this.text(`${this.$t('currentStreak')}${this.user.achievements.streak}`);
             this.playSound('Achievement_Unlocked');
             if (!this.user.preferences.suppressModals.streak) {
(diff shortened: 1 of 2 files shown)


citrusella:
I feel weird commenting on this ~~(because PRs scare me)~~, but I got curious and then started digging around to find [what the notification looked like before the redesign](http://habitica.wikia.com/wiki/File:Streak_Achievement_Notification.png): Should this maybe say something like [this string](https://github.com/HabitRPG/habitica/blob/592cfef6c6cc1c62f3e5ab620fa675a0586f8c42/website/common/locales/en/tasks.json#L107) (the "streaks" string, which is "Streak Achievements") instead of "current streak"? (Maybe not that *exact* string alone since it lacks ": " on the end, but...)

(In fact, just before I hit send on this comment, I thought to look at the fork I took of Habitica forever ago ~~and then literally never remembered to touch again~~, and that string I linked *is* the one used in [the pre-redesign notification](https://github.com/citrusella/habitica/blob/develop/website/client-old/js/services/notificationServices.js#L97). :open_mouth: )

Figured I'd ask since "current streak" could be confusing (perhaps it could make someone think that's the number of times they've done that particular daily? my brain got tripped up on that meaning and I came into the PR *knowing* what the context was 0_o), and also because the pre-redesign notification notates it as "streak achievements". \*shrug*

~~Please ignore me if this was somehow a bad question or you had a reason to use that phrasing that I just didn't understand. XP~~

TheHollidayInn:
Thanks for this! I did not know we had an image of the achievement before. I will definitely update the copy to use that text instead.

PR #10566, opened 2018-07-30
Added mute end date

TheHollidayInn:
Allows for admin to set an end date for muting.

### DIFF
--- a/website/client/components/hall/heroes.vue
+++ b/website/client/components/hall/heroes.vue
@@ -60,12 +60,14 @@
                     label
                       input(type='checkbox', v-if='hero.flags', v-model='hero.flags.chatRevoked')
                       strong Chat Privileges Revoked
-                  div(v-if='hero.flags.chatRevokedEndDate')
+                  div(v-if='hero.flags.chatRevoked && hero.flags.chatRevokedEndDate')
                     strong User is currently muted until
                     br
-                    div {{ hero.flags.chatRevokedEndDate }}
+                    div {{ userRevokedEndDate(hero) }}
+                  div(v-else-if='hero.flags.chatRevoked')
+                    strong User is currently muted indefinitely
                   div
-                    strong How many days do you want to mute this user?
+                    strong For how many days from today do you want to mute this user? Leave as 0 for indefinite.
                     br
                     input(type='number', v-model='numberOfDaysToMute')
                 .form-group
@@ -154,6 +156,10 @@ export default {
     ...mapState({user: 'user.data'}),
   },
   methods: {
+    userRevokedEndDate (hero) {
+      if (moment().isAfter(moment(hero.flags.chatRevokedEndDate))) return 'User is no longer muted';
+      return moment(hero.flags.chatRevokedEndDate).format(this.user.preferences.dateFormat.toUpperCase());
+    },
     getAllItemPaths () {
       // let questsFormat = this.getFormattedItemReference('items.quests', keys(this.quests), 'Numeric Quantity');
       // let mountsFormat = this.getFormattedItemReference('items.mounts', keys(this.mountInfo), 'Boolean');
@@ -184,7 +190,7 @@ export default {
     async loadHero (uuid, heroIndex) {
       this.currentHeroIndex = heroIndex;
       let hero = await this.$store.dispatch('hall:getHero', { uuid });
-      this.hero = Object.assign({}, this.hero, hero);
+      this.hero = Object.assign({}, hero);
       if (!this.hero.flags) {
         this.hero.flags = {
           chatRevoked: false,
@@ -206,6 +212,7 @@ export default {
       this.heroID = -1;
       this.heroes[this.currentHeroIndex] = heroUpdated;
       this.currentHeroIndex = -1;
+      this.numberOfDaysToMute = 0;
     },
     populateContributorInput (id, index) {
       this.heroID = id;
--- a/website/client/components/hall/heroes.vue
+++ b/website/client/components/hall/heroes.vue
@@ -55,10 +55,19 @@
               div(v-if="expandAuth")
                 pre {{hero.auth}}
                 .form-group
+                  h5 User Mute Settings
                   .checkbox
                     label
                       input(type='checkbox', v-if='hero.flags', v-model='hero.flags.chatRevoked')
-                      | Chat Privileges Revoked
+                      strong Chat Privileges Revoked
+                  div(v-if='hero.flags.chatRevokedEndDate')
+                    strong User is currently muted until
+                    br
+                    div {{ hero.flags.chatRevokedEndDate }}
+                  div
+                    strong How many days do you want to mute this user?
+                    br
+                    input(type='number', v-model='numberOfDaysToMute')
                 .form-group
                   .checkbox
                     label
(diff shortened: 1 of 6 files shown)


Alys:
This might need extra code to handle the case where a user who has previously been muted for a specific number of days tries to post a slur. The automatic slur  blocker will set the chat revoked flag to true but that won't have any effect if there's already an end date in the past so the slur blocker will need to delete the end date. It should even delete an end date that's in the future because slur muting is often permanent.

TheHollidayInn:
Good catch! Updated.

Alys:
Can the slur muting just delete the end date instead of setting it to far in the future? Slur muting isn't always for life - it's often permanent but if the user apologizes appropriately, we set chatRevoked to false. If slur muting deletes the end date, that has the same effect as a lifetime ban, but still allows us to reverse the ban if we want to.

TheHollidayInn:
Setting chat revoked to false does unmute the user (no matter what the date field says): https://github.com/HabitRPG/habitica/pull/10566/files#diff-de07389f36251fa6205b76174b233e8eR4

Alys:
Ah yeah, that's true.

website/client/components/hall/heroes.vue
```diff
@@ -55,10 +55,19 @@
...
                     label
                       input(type='checkbox', v-if='hero.flags', v-model='hero.flags.chatRevoked')
-                      | Chat Privileges Revoked
+                      strong Chat Privileges Revoked
+                  div(v-if='hero.flags.chatRevokedEndDate')
+                    strong User is currently muted until
+                    br
+                    div {{ hero.flags.chatRevokedEndDate }}
+                  div
+                    strong How many days do you want to mute this user?
```
Alys: Can this text be changed to this:

"For how many days from today do you want to mute this user? Leave as 0 for indefinite."

website/client/components/hall/heroes.vue
```diff
@@ -55,10 +55,19 @@
...
+                  h5 User Mute Settings
                   .checkbox
                     label
                       input(type='checkbox', v-if='hero.flags', v-model='hero.flags.chatRevoked')
-                      | Chat Privileges Revoked
+                      strong Chat Privileges Revoked
+                  div(v-if='hero.flags.chatRevokedEndDate')
+                    strong User is currently muted until
+                    br
+                    div {{ hero.flags.chatRevokedEndDate }}
```
Alys: Can this date be formatted with the moderator's locale? I.e., more human-readable, in the moderator's own time zone, and with the time zone shown.

Alys: When there's no end date specified, can it say this:
"User is currently muted indefinitely."

Alys:
I've tested locally and left a couple of comments in the code.

Also, there needs to be a way to convert a mute with an end date to one with no end date, for example, a button labelled "make mute indefinite" which deletes the end date. Or if you prefer, we could make it so that if we set the number of days back to 0, the end date is deleted (that might seem a little counterintuitive but it's consistent with the default case of leaving the days set to 0 when we don't want to set an end date on a new mute).

If you mute one user (or view one user who has been muted before) and then view a second user who has not been muted, the second user looks like they have been muted - the chat revoked checkbox is ticked and the end date from the first user is still present. I.e., the two chatRevoked* fields are not cleared when a second user is viewed.

When you unmute a user by unticking the checkbox, then save the change and view the user again (after a reload to remove caching), the old date is still visible and the text says the user is currently muted. The text and date should be hidden when the user is unmuted or the end date could be automatically deleting when the checkbox is unticked.

When you view a user who was given an end date that is now in the past, there's no clear indication that the user is no longer muted - at a glance it looks like they still are muted and you have to read the date to find out that they aren't. For example, in the screenshot below, that date is in the past. In stressful situations such as when a previous troll has returned and is causing trouble, this could lead to the troll not being muted again if we misread the date. Can there be an indication that the date is no longer relevant and the user is not muted?

image



This is just a note for the record / for any other mods reading this PR:
Sometimes we mute for less than a day, which isn't supported here but I don't think it's worth adding more complexity to the code. If we decide we want to keep muting for less than a day in specific cases, we can do it with our current method (set an indefinite mute then remove it later).

TheHollidayInn:
You can mute for less than a day. For example using .5 will mute for half a day.

For this:
```
Also, there needs to be a way to convert a mute with an end date to one with no end date, for example, a button labelled "make mute indefinite" which deletes the end date. Or if you prefer, we could make it so that if we set the number of days back to 0, the end date is deleted (that might seem a little counterintuitive but it's consistent with the default case of leaving the days set to 0 when we don't want to set an end date on a new mute).
```
Let's simply add a large date for now. For example, if you want to indefinitely mute this user, just put in 1 million days. Just to keep this simple.

Alys:
I hadn't tested this again since your last changes.

I'll make it an urgent task for me to test it in staging as soon as I can, but we'll need to delay any deployments of staging to production until I've done that, and then I'd like it to sit in staging for a full week for mods to use it before it's released. It's essential for moderation that we can be certain the mute function is working perfectly, including any displays it shows to users. 

@SabreCat FYI about delaying the next deployment.

TheHollidayInn:
The last changes I made were simply cosmetic on the client. 

Letting it sit for a while sounds good.

Alys:
Cosmetic changes are visible to the user so I believe it's still useful to test them before merging.

From testing in staging:

- [ ] the "User is currently muted until" timestamp needs to display the time of day and time zone as well as the date. It should be converted to the user's time zone (with the time zone being displayed so that we remember it is in our own zone).
- [ ] The text "User is currently muted indefinitely" appears as soon as you tick the checkbox (i.e., before you've saved your change), which is confusing. You could misinterpret it as showing that the user was already muted but the Hall took a couple of seconds to fetch the data, or it could make you think that you don't need to save. In a high-pressure situation when we're muting a few trolls urgently, potential confusion is undesirable. The text should appear only when loading a user who is already muted (it would be ideal if it also appeared after saving a new mute if the server has confirmed the save was successful, but that's not essential).
- [ ] When a user had been muted with an end date that now has passed:
   - [ ] When you load that user in the Hall, both "User is currently muted until" and "User is no longer muted" appear. Only "User is no longer muted" should appear.
   - [ ] Also when you load that user in the Hall, the checkbox is still ticked. It should be unticked (even if `flags.chatRevoked` is still true).
   - [ ] In the player's profile, the mute/unmute icon incorrectly indicates that the user is still muted (it shows the icon you see when the user is still muted and on hover it says "reinstate chat privileges").
   - [ ] Clicking the mute icon in the player's profile should re-mute the user. This means that either:
      - [ ] clicking the mute icon should automatically delete the end date (simplest) or
      - [ ] the mute UI in the profile should be expanded to allow selection of a new end date or manual deletion of the old end date

I've discussed this with the other mods and while we all agree that this is a nice feature to have, we don't feel it's needed urgently. We'd like to suggest that it be taken out of staging while it's being developed. That means that the current fixes and then more testing don't have to be done urgently, and that the next staging deployment doesn't need to be delayed until this feature is finished, and that we can continue to use staging for moderation (ideal as always because it allows us to help with the normal testing of staging in general).

The other mods also indicated that it's desirable to be able to manually changing a mute that had an end date to be indefinite (i.e., by deleting the end date). It avoids the annoyance of having to enter a large number of days and the potential problems that would occur if we accidentally didn't make the number large enough. It also assists in high-pressure situations by making it immediately clear when a mute is indefinite, without us needing to parse the date (and potentially misreading it). Changing a temporary mute to an indefinite one is a common action.

PR #10567, opened 2018-07-31
limit chat message flagging ability for new players - fixes #10069

### LINKED ISSUE #10069: limit chat message flagging ability for new players
Currently, when a player clicks the report button on a chat message to flag it for moderators' attention, the message's `flagCount` is increased by 1:
https://github.com/HabitRPG/habitica/blob/d5e4be85e991740e6ec622c41f7c38a307b802ea/website/server/libs/chatReporting/groupChatReporter.js#L81

We don't want that to happen if the player's account is less than three days old. I.e., the flagCount should NOT be increased when a new player flags a message, All the other actions that occur when flagging should still happen.

There should be no change to the behaviour when a player with an account older than three days flags a message. 

This change should be made in the API so that it applies regardless of which platform the player is using.

The PR for this will need to include tests for players older and younger than three days to ensure the behaviour is correct for both.

_This issue was suggested by @phillipthelen to prevent a troll making a new account to flag innocent posts to hide them. It's unlikely to have a significant detrimental effect on moderation because moderators will still be notified about the flagged messages and can then hide them themselves._

Alys:
Fixes https://github.com/HabitRPG/habitica/issues/10069

This PR affects the flagging ability for players who have created their account less than three days ago. They are still able to flag messages, the moderators are still notified with an email and Slack notification, but the flag count of the message is not increased (to prevent a troll making a new account to flag innocent posts to hide them).

The Slack notification has a small text change to ensure that moderators realise that the post has not been partially hidden.

This PR also removes a couple of unrelated pieces of commented-out code that we don't need (see commit comments for details if desired).

### DIFF
--- a/test/api/unit/libs/slack.js
+++ b/test/api/unit/libs/slack.js
@@ -58,7 +58,7 @@ describe('slack', () => {
           title: 'Flag in Some group - (private guild)',
           title_link: undefined,
           text: 'some text',
-          footer: sandbox.match(/<.*?groupId=group-id&chatId=chat-id\|Flag this message>/),
+          footer: sandbox.match(/<.*?groupId=group-id&chatId=chat-id\|Flag this message.>/),
           mrkdwn_in: [
             'text',
           ],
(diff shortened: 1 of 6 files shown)


Alys:
Still fixing tests; not ready for review yet.

Alys:
ready for review

paglias:
LGTM @Alys

Issue #9231, opened 2017-10-18
Due until Completed, X times every Y intervals


librarianmage:
Copied from #6941 

If we are trying to implement use cases like

* Cleaning bathrooms once every month
* Replacing your home air filter once every three months
* Getting a free credit check once every four months

I feel like a better representation would be creating "blocks" of time in which the daily is active until the period passes, after which the damage is dealt. It might not look like a significant change, but it would keep each period the same length and would allow use cases like

* Water plants thrice per week
* Exercise 10 times per month

There could also be a completion display, like how To-Dos with checklists are displayed.

---

Images:

Basic Daily (Repeat 1x a week on Sundays)
------------------------------------------------------
![img_201709022120240](https://user-images.githubusercontent.com/12386291/30000092-4cf59354-9025-11e7-82ea-db107d134131.jpg)

Basic Daily w/ Due Until Completed (Repeat 1x a week on Sundays)
-------------------------------------------------------------------------------------
![img_201709022120241](https://user-images.githubusercontent.com/12386291/30000094-4cf7f9be-9025-11e7-88d7-cadfb6277484.jpg)

X Intervals since Completion (7 days since completion) #6941 
---------------------------------------------------------------------
![img_201709022120242](https://user-images.githubusercontent.com/12386291/30000095-4cf8b07a-9025-11e7-830f-607ea2bcc34d.jpg)

X Intervals since Completion w/ Due Until Completed (7 days since completion)
--------------------------------------------------------------------------------------------------
![img_201709022120243](https://user-images.githubusercontent.com/12386291/30000093-4cf7dbb4-9025-11e7-8ded-f2a7e981d696.jpg)

X Times Y Intervals (1 time every week)
-------------------------------------------------
![img_201709022120244](https://user-images.githubusercontent.com/12386291/30000096-4cf94bde-9025-11e7-9f84-4acf84e3d394.jpg)

librarianmage:
@SabreCat Since I'm planning on working on them in tandem, is it okay that I combined them into one?

SabreCat:
Sure, thanks!

librarianmage:
@Sabrecat I was thinking about how this best be implemented on the client,
and I was wondering: should the designs be created first?

lemoness:
@MathWhiz Yes, that's probably for the best if you don't mind waiting. We've got quite a lot of designs to implement at the moment, but this one is on our list!

librarianmage:
Darn. Well, I can wait!

[...]

Issue #6941, opened 2016-03-22
Repeat Every X Days Since Last Completed


anwilli5:
In the 'Repeat' section under 'Advanced Options' for a Daily, it'd be great to have an entry called something like 'Every X Days Since Last Completed'.  This should add a Daily that becomes active on the specified 'Start Date' and, upon completion, gets scheduled for X days from the completion date.  It should remain active and count as a missed Daily every day until it is completed.  If it is checked while not active, the Daily should be scheduled for X days from the date of being checked.

Example use cases:
- Cleaning bathrooms once every month
- Replacing your home air filter once every three months
- Getting a free credit check once every four months

There has been a lot of discussion about implementing something like this on the Trello card linked to below, and there isn't a clear consensus about how this and similar features (Ex: repeat every X days, but have the Daily remain active until completed) should be implemented.  The approach described above seems reasonable, though, and is in line with what sites like Todoist do (they have the 'every' keyword tied to repeat every X days and 'every!' which would be the equivalent to this feature.)  Anyway, I wanted to make a ticket for this functionality and post a bounty to encourage someone to implement it unless there are serious reservations about adding this.

https://trello.com/c/q7uM2Ymq/92-dailies-v-3-dailies-minimum-does-not-apply-button-weekly-rolling-over-streak-history

[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=32123883)](https://www.bountysource.com/issues/32123883-repeat-every-x-days-since-last-completed?utm_source=32123883&utm_medium=shield&utm_campaign=ISSUE_BADGE)

Alys:
@anwilli5 We actually use github almost exclusively for bugs, rather than feature requests. You can make suggestions through Help > Request a Feature from the Habitica website's menu, which will take you to a Trello card with instructions for where to put your request. In this particular case, you can write it on https://trello.com/c/XuVokK2V/108-repeat-every-days-weekly-monthly but for future requests, the instructions will guide you to the correct card.

anwilli5:
@Alys I was trying to mirror was was done with #4173.  Given that the BountySource page seems to tie funding to specific github tickets, is there no way currently to put money behind new feature development (since tickets are only for bugs?)

Alys:
I understand now, and thank you for being willing to support our developers! I'm sorry I missed your comment about the bounty when I first read this.

[...]

PR #10577, opened 2018-08-06
combine cron's Resting in the Inn code with non-sleeping code - fixes #5232 etc

### LINKED ISSUE #5232: Allow Habits and To-Dos to "age" while Resting in the Inn

Alys:
This PR combines the cron code for resting in the inn with the normal cron code, as described at https://github.com/HabitRPG/habitica/pull/9627#discussion_r156749416 and https://github.com/HabitRPG/habitica/issues/6941#issuecomment-387927531
The intention is to:
- make it clearer which cron actions are not performed when sleeping (very few of them!)
- remove some bugs (e.g., cronCount was not being incremented when sleeping)
- prevent future bugs (we had a habit of adding new cron code to the non-sleeping section and forgetting to add it to `performSleepTasks`)
- remove the blocker for https://github.com/HabitRPG/habitica/issues/6941 / https://github.com/HabitRPG/habitica/issues/9231 ("Repeat Every X Days Since Last Completed" / "Due until Completed, X times every Y intervals") - that's an example of the previous item (the current code in the gamma site doesn't run when the user is Resting) 
- fix https://github.com/HabitRPG/habitica/issues/5232 ("Allow Habits and To-Dos to "age" while Resting in the Inn") - kind of an incidental fix since aging while sleeping happens automatically now that the code has been combined.
- allow the Perfect Day buff to be obtained when Resting in the Inn if you have completed all your due Dailies.

----

_As notes to myself for after/if this goes live:_
- [ ] _edit the [cron section of the Rest in the Inn wiki page](https://habitica.wikia.com/wiki/Rest_in_the_Inn?action=edit&section=4)_
- [ ] _tell socialites about some of the changes (e.g., perfect day is now given when resting; aging of tasks)_
- [ ] _see about getting the code merged into the blocked PR mentioned above_

### DIFF
--- a/website/server/libs/cron.js
+++ b/website/server/libs/cron.js
@@ -489,19 +489,6 @@ export function cron (options = {}) {
     mp: user.stats.mp - beforeCronStats.mp - (oldCronNotif ? oldCronNotif.data.mp : 0),
   });
 
-  // TODO: Clean PMs - keep 200 for subscribers and 50 for free users. Should also be done while resting in the inn
-  // let numberOfPMs = Object.keys(user.inbox.messages).length;
-  // if (numberOfPMs > maxPMs) {
-  //   _(user.inbox.messages)
-  //     .sortBy('timestamp')
-  //     .takeRight(numberOfPMs - maxPMs)
-  //     .forEach(pm => {
-  //       delete user.inbox.messages[pm.id];
-  //     })
-  //
-  //   user.markModified('inbox.messages');
-  // }
-
   // Analytics
   user.flags.cronCount++;
   trackCronAnalytics(analytics, user, _progress, options);
--- a/website/server/libs/cron.js
+++ b/website/server/libs/cron.js
@@ -172,27 +172,6 @@ function resetHabitCounters (user, tasksByType, now, daysMissed) {
   });
 }
 
-function performSleepTasks (user, tasksByType, now, daysMissed) {
-  user.stats.buffs = _.cloneDeep(CLEAR_BUFFS);
-
-  tasksByType.dailys.forEach((daily) => {
-    let completed = daily.completed;
-    let thatDay = moment(now).subtract({days: 1});
-
-    if (shouldDo(thatDay.toDate(), daily, user.preferences) || completed) {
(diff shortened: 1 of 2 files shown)


Alys:
This is ready for review. It might be easiest to review each commit rather than the whole PR at once. I tried to keep the commits arranged logically with only a relatively small set of closely-related changes in each one.

website/server/libs/cron.js
```diff
@@ -461,20 +435,23 @@ export function cron (options = {}) {
...
-  _progress = progress.toObject(); // clone the old progress object
-  progress.down = -1300;
-  _.merge(progress, {down: 0, up: 0, collectedItems: 0});
-
-  // Send notification for changes in HP and MP
+  if (!user.preferences.sleep) {
+    let progress = user.party.quest.progress;
+    _progress = progress.toObject(); // clone the old progress object
+    progress.down = -1300;
+    _.merge(progress, {down: 0, up: 0, collectedItems: 0});
```
paglias: this is not specifically related to this PR but we're setting progress.down to -1300 and right after to 0?

paglias (commented):
left one comment but it looks good!

website/server/libs/cron.js
Alys: That is weird! Well spotted! It was introduced in 98c019a0b658df1fb5f0c6db5b09b379d6cb58e2 and I'm guessing it was added for testing. I've added a commit to remove it.

paglias:
Looking good!

PR #10582, opened 2018-08-08
Filter dailies by due/not due in group plan and challenge page

### LINKED ISSUE #10444: Tasks cannot be sorted by all/due/not due in group plan overview
Reported in Report a Bug, confirmed by me. Trying to filter by due or not due doesn't change anything in the group plan task board.

image

image

thefifthisa:

Fixes #10444

### Changes

Dailies in the group plan and challenge page can now be filtered by whether or not they're due.

~However, it doesn't seem to work for newly added tasks because they don't have the `isDue` field until you refresh the page.~

----

### DIFF
--- a/website/client/components/tasks/column.vue
+++ b/website/client/components/tasks/column.vue
@@ -362,7 +362,7 @@ export default {
           type: this.type,
           filterType: this.activeFilter.label,
         }) :
-        this.filterByCompleted(this.taskListOverride, this.activeFilter.label);
+        this.filterByLabel(this.taskListOverride, this.activeFilter.label);
 
       let taggedList = this.filterByTagList(filteredTaskList, this.selectedTags);
       let searchedList = this.filterBySearchText(taggedList, this.searchText);
@@ -598,10 +598,12 @@ export default {
         }
       });
     },
-    filterByCompleted (taskList, filter) {
+    filterByLabel (taskList, filter) {
       if (!taskList) return [];
       return taskList.filter(task => {
         if (filter === 'complete2') return task.completed;
+        if (filter === 'due') return task.isDue;
+        if (filter === 'notDue') return !task.isDue;
         return !task.completed;
       });
     },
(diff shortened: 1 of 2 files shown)


negue:
I tested the filter and its fixed, thanks! :)

On the `isDue` issue, you could use the result in https://github.com/HabitRPG/habitica/blob/45ca090105b7e05e214f8b000983d7a17468fc40/website/client/components/tasks/taskModal.vue#L932 and set it to `this.task` that way you get the server created task (which contains `isDue`)

thefifthisa:
Hi @negue! Do you mean `this.task =  await this.$store.dispatch('tasks:createGroupTasks', {`? I tried that and it seems to work, but I get a `Vue warn` that says "avoid mutating a prop directly." 😕

I also noticed that the filter functions are different for users and groups. Is there a way to use the same one for both?

paglias:
@thefifthisa instead of changing the entire prop (this.task = ...) you might use Object.assign to only update changed/new properties, so 

```
const response =           await this.$store.dispatch('tasks:createGroupTasks', {
            groupId: this.groupId,
            tasks: [this.task],
          });

Object.assign(this.task, response.data.data);
```

paglias:
@thefifthisa instead of changing the entire prop (this.task = ...) you might use Object.assign to only update changed/new properties, so 

```
const response = await this.$store.dispatch('tasks:createGroupTasks', {
 groupId: this.groupId,
 tasks: [this.task],
});

Object.assign(this.task, response);
```

thefifthisa:
Thanks @paglias! It's working now.

Alys:
@thefifthisa I've tested this locally and the Dailies due/not due filters for Group Plan tasks are working correctly, even for new tasks at the moment they are created - nice work! 

I happened to notice though that the Challenge page has the same problem that Group Plans had where the filters don't work for newly added tasks until you refresh the page, and I'm guessing it could be fixed with the same fix. The code seems to be just above the code you changed. This isn't strictly related to the issue though, so if you'd prefer to not work on it someone else could fix that later! :)

thefifthisa:
@Alys Sure thing, just added it in the latest commit!

Alys:
For the record, paglias added this PR to @thefifthisa's progress towards the next tier.

@thefifthisa Thanks for this, especially for adding the challenge fix! :)

PR #10589, opened 2018-08-10
Correct timing on updating Group Plan member quantities

SabreCat:

### Changes

Moves some logic when updating Group Plan member counts to later in the relevant functions, hopefully ensuring that the data we're relying on to reconcile the numbers will be up to date when we draw upon it.

### DIFF
--- a/website/server/controllers/api-v3/groups.js
+++ b/website/server/controllers/api-v3/groups.js
@@ -587,14 +587,6 @@ api.joinGroup = {
     }
     if (!isUserInvited) throw new NotAuthorized(res.t('messageGroupRequiresInvite'));
 
-    // @TODO: Review the need for this and if still needed, don't base this on memberCount
-    if (!group.hasNotCancelled() && group.memberCount === 0) group.leader = user._id; // If new user is only member -> set as leader
-
-    if (group.hasNotCancelled())  {
-      await payments.addSubToGroupUser(user, group);
-      await group.updateGroupPlan();
-    }
-
     group.memberCount += 1;
 
     let promises = [group.save(), user.save()];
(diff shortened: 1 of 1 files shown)


website/server/controllers/api-v3/groups.js
```diff
@@ -944,6 +940,12 @@ api.removeGroupMember = {
       member.save(),
       group.save(),
     ]);
+
+    if (group.hasNotCancelled())  {
```
Alys: Is it worth making this `if (isInGroup & group.hasNotCancelled())  {` so that we don't run the check if the user was merely invited? That's slightly less load on the database (not really an issue I guess) and it also protects the user if `cancelGroupSubscriptionForUser()` develops any bugs in future that cause it to cancel subscriptions from group plans inappropriately (perhaps that's just paranoia).

TheHollidayInn: HasNotCancelled does call the DB. But for this route we are assuming they are in a group. So isInGroup wouldn't prevent that. Or it would always be false because we are cancelling after removing them now.

Alys: `isInGroup` is just a variable ("guild" / "party", and effectively a binary flag when used like this) and it isn't changed after they've been removed so it shouldn't be false if it started as true at the beginning of the function.
The function is also run if `isInvited` is true (i.e., to remove an invitation to a group, rather than removing a group member) and that's the case where it's not necessary to attempt to remove a free subscription.

paglias: I don't think that HasNotCancelled calls the db, it only uses local data

Alys: Yeah, `hasNotCancelled` won't do a db search, but `cancelGroupSubscriptionForUser` will. I guess my main point is the code inside this `if` statement doesn't need to be executed for the case where an invitation is being removed, so running it only when `isInGroup` is true seems more efficient and limits the exposure of any potential future bugs in `cancelGroupSubscriptionForUser`

TheHollidayInn: Oh I didn't realize we use this same route for invitations. Then yes, you are correct. I'll add that check in

PR #10590, opened 2018-08-11
Show accurate XP gain in notification on level up

### LINKED ISSUE #10297: show accurate Experience (XP) gain in notification on level-up
As described in https://github.com/HabitRPG/habitica/issues/9339, the Experience notification was deliberately removed when you level up because it was reporting that you lost XP. Strictly speaking, it was correct that you'd lost XP because on level-up you go from a high XP to 0 XP, however it's demotivating to see a message about losing XP.

The Experience notification should appear and should correctly display how much XP you gained when you level up.

However this might not be a trivial fix because our notifications are based on the total change in stats.

Arsnael:

Fixes #10297

### Changes
Fixing the issue by recalculating the experience needed to level up from the previous level, then substracting the previous state of XP to it then adding the new state of XP to it.  

I think there is an edge case though, like if the user levels up more than once, the result might likely be wrong (as the notifications are working on the previous and new state of XP in this case, I don't see how we can see if we get many level ups in there). But that case should happen rarely? (like a low level user in a party that just slayed a big monster perhaps?). Let me know what you think of it...

----

### DIFF
--- a/website/client/components/notifications.vue
+++ b/website/client/components/notifications.vue
@@ -88,6 +88,7 @@ import axios from 'axios';
 import moment from 'moment';
 import throttle from 'lodash/throttle';
 
+import { toNextLevel } from '../../common/script/statHelpers';
 import { shouldDo } from '../../common/script/cron';
 import { mapState } from 'client/libs/store';
 import notifications from 'client/mixins/notifications';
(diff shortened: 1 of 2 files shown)


paglias:
@Arsnael  thanks for this! I'm merging the PR and awarding your contributor tier.

Arsnael:
@paglias  Happy to help, and will try to contribute more :)

PR #10593, opened 2018-08-12
Chat approval 2

### LINKED ISSUE #8063: Optional moderator approval for chat posts from all new users

TheHollidayInn:
Fixes #8063


website/server/libs/apiMessages.js
```diff
@@ -0,0 +1,25 @@
+// A map of messages used by the API that don't need to be translated and
```
paglias, line 1: I think this lib has been implemented already somewhere else

website/server/libs/config.js
```diff
@@ -0,0 +1,18 @@
+import nconf from 'nconf';
+
+const CHAT_APPROVAL_REQUIRED = nconf.get('CHAT_APPROVAL_REQUIRED');
+const CHAT_APPROVAL_DAYS = nconf.get('CHAT_APPROVAL_DAYS');
+
+let api = {};
```
paglias, line 6: Instead of a new  lib I think we should keep using nconf.set like we already do for some config cars here https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/setupNconf.js, since all these values are static

website/server/libs/apiMessages.js
TheHollidayInn: Yeah, this was cherry picked by accident

webpack/config/prod.env.js
```diff
@@ -34,7 +34,7 @@ let env = {
   },
 };
 
-'NODE_ENV BASE_URL GA_ID STRIPE_PUB_KEY FACEBOOK_KEY GOOGLE_CLIENT_ID AMPLITUDE_KEY PUSHER:KEY PUSHER:ENABLED LOGGLY_CLIENT_TOKEN'
+'NODE_ENV BASE_URL GA_ID STRIPE_PUB_KEY FACEBOOK_KEY GOOGLE_CLIENT_ID AMPLITUDE_KEY PUSHER:KEY PUSHER:ENABLED LOGGLY_CLIENT_TOKEN CHAT_APPROVAL_REQUIRED'
```
Alys, line 37: Should this have CHAT_APPROVAL_DAYS in it too?

website/common/locales/en/groups.json
```diff
@@ -446,6 +446,11 @@
   "worldBossBullet2": "The World Boss won’t damage you for missed tasks, but its Rage meter will go up. If the bar fills up, the Boss will attack one of Habitica’s shopkeepers!",
   "worldBossBullet3": "You can continue with normal Quest Bosses, damage will apply to both",
   "worldBossBullet4": "Check the Tavern regularly to see World Boss progress and Rage attacks",
+
+  "approvalNotRequired": "Approval not required",
+  "messagedFlaggedNotHidden": "Message flagged once, not hidden",
+  "messageHidden": "Message hidden",
+  "approvalNotRequired": "Approval not required",
```
Alys, line 453: There's some duplication in these four strings and in the other three new ones at the bottom of this file.

Also all of these new strings won't ever need to be translated so they could be put into the file where non-translated strings are stored (I can't recall where that file is).

webpack/config/prod.env.js
TheHollidayInn: Nah, these are for the client and that variable is only used on the server.

website/common/locales/en/groups.json
TheHollidayInn: Ah thanks. I messed this up while cherry-picking. I'll move the strings. the lib is in common now.

PR #10596, opened 2018-08-12
Added amoire food to user immediately

### LINKED ISSUE #9934: Food from Enchanted Armoire should be visible immediately
When you obtain Food from the Enchanted Armoire, and then go to your Inventory or Stable to see it, it is not immediately visible. You have to sync before it appears.

The desired behaviour is that it should appear immediately.

TheHollidayInn:
Fixes #9934

### DIFF
--- a/website/client/store/actions/shops.js
+++ b/website/client/store/actions/shops.js
@@ -64,6 +64,12 @@ async function buyArmoire (store, params) {
     if (item.type === 'gear') {
       store.state.user.data.items.gear.owned[item.dropKey] = true;
     }
+
+    if (item.type === 'food') {
+      if (!store.state.user.data.items.gear.owned[item.dropKey]) store.state.user.data.items.gear.owned[item.dropKey] = 0;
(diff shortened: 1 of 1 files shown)


website/client/store/actions/shops.js
```diff
@@ -64,6 +64,12 @@ async function buyArmoire (store, params) {
     if (item.type === 'gear') {
       store.state.user.data.items.gear.owned[item.dropKey] = true;
     }
+
+    if (item.type === 'food') {
+      if (!store.state.user.data.items.gear.owned[item.dropKey]) store.state.user.data.items.gear.owned[item.dropKey] = 0;
```
paglias: I think the food should go under user.items.food[key] instead of under items.gear.owned

paglias (changes requested):

website/client/store/actions/shops.js
TheHollidayInn: uhh.. yea. Not sure what I did here.

PR #10598, opened 2018-08-14
Fixed concurrency issues with push devices

TheHollidayInn:


website/server/controllers/api-v3/pushNotifications.js
```diff
@@ -87,8 +87,8 @@ api.removePushDevice = {
       throw new NotFound(res.t('pushDeviceNotFound'));
     }
 
-    pushDevices.splice(indexOfPushDevice, 1);
-    await user.save();
+    const pullQuery = { $pull: { pushDevices: { $elemMatch: { regId } } } };
+    await user.update(pullQuery).exec();
 
     res.respond(200, user.pushDevices, res.t('pushDeviceRemoved'));
```
paglias: LGTM but both here and in the other route the response will contain the pushDevices before the change, I'd keep `pushDevices.splice(indexOfPushDevice, 1);	 ` and ` pushDevices.push(item);` without the save so the response will be up to date. 

And actually pushdevices are subdocs with a model (models/pushDevice) so when adding a new one without goign through mongoose we should build the model `const pushDevice = (new PushDevice(...)).toJSON()` and pass that to the update command

paglias (dismissed):

website/server/controllers/api-v3/pushNotifications.js
TheHollidayInn: Oh interesting. I didn't think about the subdoc nature. And I'll fix the response as well.

PR #10599, opened 2018-08-14
Fixed mp rounding

### LINKED ISSUE #5233: MP rounds down, MP alerts round up...
It makes sense for MP to round down... if it rounded up, it'd look like you could afford skills you can't.

However, MP alerts round up, which makes for some surprising total amounts of MP now. This is especially apparent because of the 1/4 amounts Habits give you.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/15592221-mp-rounds-down-mp-alerts-round-up?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F68393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheHollidayInn:
Fixes #5233

### DIFF
--- a/website/client/mixins/notifications.js
+++ b/website/client/mixins/notifications.js
@@ -2,6 +2,12 @@ import habiticaMarkdown from 'habitica-markdown';
 import { mapState } from 'client/libs/store';
 import { getDropClass, getXPMessage, getSign, round } from 'client/libs/notifications';
 
+// See https://stackoverflow.com/questions/4187146/truncate-number-to-two-decimal-places-without-rounding
+function toFixedWithoutRounding (num, fixed) {
+  const re = new RegExp(`^-?\\d+(?:\.\\d{0,${(fixed || -1)}})?`);
+  return num.toString().match(re)[0];
+}
+
 export default {
   computed: {
     ...mapState({notifications: 'notificationStore'}),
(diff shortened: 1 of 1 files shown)


website/client/mixins/notifications.js
```diff
@@ -46,7 +46,7 @@ export default {
       this.notify(parsedMarkdown, 'info');
     },
     mp (val) {
-      this.notify(`${this.sign(val)} ${this.round(val)}`, 'mp', 'glyphicon glyphicon-fire', this.sign(val));
+      this.notify(`${this.sign(val)} ${this.round(val, 2)}`, 'mp', 'glyphicon glyphicon-fire', this.sign(val));
```
Alys: This needs to always round down, never round up so that it matches the type of rounding used in the mana stats bar. For example, 1.19 should be changed to 1.1 (not 1.2). I think the `this.round()` function is the one below, in which case it will round up for digits above 5, so that's not what we want.
https://github.com/HabitRPG/habitica/blob/f810fff6fc7ac4c0bb263e552d5892ab7b0ddcab/website/client/libs/notifications.js#L45-L47

Also displaying one decimal place instead of two would be best. Basically, the round() function here needs to be replaced with one that just truncates after one decimal place.

TheHollidayInn: This actually truncates for digits greater than 2 which ends up rounding down. 1.234 becomes 1.23. were you getting different results locally?

Alys: I didn't test locally because I can't think of how to easily control the amount of mana given by a task but I was extrapolating behaviour from the toFixed docs. 1.234 will become 1.23 (which is good) but 1.236 will become 1.24. We want rounding down in every case (so for our use, 1.26 would need to become 1.3).

TheHollidayInn: Ah I didn't see you reference the docs. I see now that toFixed is the issue: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed#Using_toFixed

SabreCat:
Possible cause of https://github.com/HabitRPG/habitica/issues/10623 ?

TheHollidayInn:
Ah yep. I'll fix that

PR #10601, opened 2018-08-14
Refactor/market vue

negue:
- [x] InventoryDrawer
- [x] Featured Items Header
- [x] Page Layout
- [x] Sidebar
- [x] Equipment
- [x] Inventory
- [x] sellModal (use the bootstrap-vue way to open the modal)
- [x] clean up

Fixed bugs:
- [x] count badge doesn't get updated on buying an item - fixed by @TheHollidayInn 👍 
- [x] repurchase old armoire gear #10659

### DIFF
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -29,7 +29,6 @@ div
           buyModal(
             :item="selectedItemToBuy || {}",
             :withPin="true",
-            @change="resetItemToBuy($event)",
             @buyPressed="customPurchase($event)",
             :genericPurchase="genericPurchase(selectedItemToBuy)",
 
@@ -565,13 +564,6 @@ export default {
           eventAction: 'click',
           eventLabel: 'Gems > Wallet',
         });
-      }
-    },
-    resetItemToBuy ($event) {
-      // @TODO: Do we need this? I think selecting a new item
-      // overwrites. @negue might know
-      if (!$event && this.selectedItemToBuy.purchaseType !== 'card') {
-        this.selectedItemToBuy = null;
       }
     },
     itemSelected (item) {
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -104,6 +104,7 @@ div
 <style lang='scss'>
   @import '~client/assets/scss/colors.scss';
 
+  /* @TODO: The modal-open class is not being removed. Let's try this for now */
   .modal {
     overflow-y: scroll !important;
   }
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -104,9 +104,8 @@ div
 <style lang='scss'>
   @import '~client/assets/scss/colors.scss';
 
-  /* @TODO: The modal-open class is not being removed. Let's try this for now */
   .modal {
-    overflow-y: scroll !important;
+    overflow-y: auto !important;
   }
 
   .modal-backdrop.show {
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -105,7 +105,7 @@ div
   @import '~client/assets/scss/colors.scss';
 
   .modal {
-    overflow-y: auto !important;
+    overflow-y: scroll !important;
   }
 
   .modal-backdrop.show {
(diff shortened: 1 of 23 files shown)


website/client/app.vue
```diff
@@ -104,9 +104,8 @@ div
 <style lang='scss'>
   @import '~client/assets/scss/colors.scss';
 
-  /* @TODO: The modal-open class is not being removed. Let's try this for now */
   .modal {
-    overflow-y: scroll !important;
+    overflow-y: auto !important;
```
TheHollidayInn: Can you revert this. We still need the overflow hack

website/client/components/shops/market/index.vue
```diff
@@ -23,223 +23,84 @@
...
-        :item="selectedItemToSell",
-        :itemType="selectedDrawerItemType",
-        :itemCount="selectedItemToSell != null ? userItems[drawerTabs[selectedDrawerTab].contentType][selectedItemToSell.key] : 0",
-        :text="selectedItemToSell != null ? getItemName(selectedDrawerItemType, selectedItemToSell) : ''",
-        @change="resetItemToSell($event)"
-      )
+            h4.popover-content-title(slot="popoverContent") {{ ctx.itemName }}
+
+      sellModal
         template(slot="item", slot-scope="ctx")
```
TheHollidayInn: I thin we can remove this slot since we don't customize the sell modal "item" slot anywhere. We should be able to move all child elements to the modal's component.

website/client/components/shops/market/index.vue
```diff
@@ -23,223 +23,84 @@
...
-      drawer(
-        :title="$t('quickInventory')"
-        :errorMessage="inventoryDrawerErrorMessage(selectedDrawerItemType)"
-      )
-        div(slot="drawer-header")
-          drawer-header-tabs(
-            :tabs="drawerTabs",
-            @changedPosition="tabSelected($event)"
+      inventoryDrawer(:showEggs="true", :showPotions="true")
+        template(slot="item", slot-scope="ctx")
```
TheHollidayInn: Same here for the slot. I don't think we customize the slot anywhere, so we should remove the slot and move the child elements to the component.

website/client/app.vue
negue: oh, it just looked ugly on windows :D and I thought if I change it to auto then it will be visible if needed

website/client/components/shops/market/index.vue
negue: this one has to stay, the market just has items but the stable needs dragable ones (yes its currently broken 😢  )

website/client/app.vue
TheHollidayInn: Yea, it's a hack :P Can you add the note back too? You have PR for fixing modals, so after that is merged in we can test removing it.

website/client/components/shops/market/index.vue
TheHollidayInn: Ah okay, I didn't see the second use of it :P That's perfectly fine then.

TheHollidayInn:
@negue not sure if you saw this, but I posted a PR to fix the count: https://github.com/HabitRPG/habitica/pull/10613

website/client/components/shared/inventoryDrawer.vue
```diff
@@ -0,0 +1,169 @@
...
+            show: () => this.showPotions,
+          },
+          {
+            key: 'special',
+            contentType: 'food',
+            label: this.$t('special'),
+            show: () => true,
+          },
+        ],
+        selectedDrawerTab: this.defaultSelectedTab,
```
paglias, line 96: does it work to reference a prop here ?

paglias:
schermata 2018-09-01 alle 19 11 40

I'm getting this error locally

website/client/components/shared/inventoryDrawer.vue
negue: yep :) at first I thought I'd have to use something like `watch` for that but it worked just like that

negue:
@paglias could you tell where you got this error? on which item I mean , maybe my local db doesn't have it listed

negue:
@paglias the method your error was raised, seems to be an obsolete one, where the reference opened the modal itself instead of modal-events

Alys:
I've marked this as notify helpers to remind me to tell Il Mago when the fix has gone live because she's recently had all her gear removed by me for repurchasing.  https://github.com/HabitRPG/habitica/issues/10659

PR #10603, opened 2018-08-15
Fixing healing light not being castable when user full hp

### LINKED ISSUE #10548: Healing Light should not be castable if the player already has 50 health
Currently, it's possible to cast the Healer's "Healing Light" skill (which increases the player's own health) even when their health is already 50. 

The desired behaviour is that when the player tries to cast the skill, if they already have 50 HP, the skill cast should fail. 

The test for the current HP should be at the API level and NOT also at the UI level. I.e., even if the website is showing that your health is 50, you should still be able to try to cast the skill. The server would then read your actual health value and reject the cast (not take your mana) if your health is at maximum. This would avoid any problems with sync errors or local display errors causing the website to think you had full health when really you did need to cast the skill (admittedly, it wouldn't be often that a player would want to cast the skill if it looked like they had full health, but a player who is aware of sync errors and is using the DDT or an app might want to).

For reference, https://github.com/HabitRPG/habitica/issues/6384 was for the same problem for Health Potions. The PR that fixed it was https://github.com/HabitRPG/habitica/pull/7934/files and some of its code can be reused (for example, the `messageHealthAlreadyMax` string should be used for this issue).

_We could also do the same fix for the Blessing skill which adds health to all party members, but a server-side check for the health of all players might make the skill too slow? Perhaps we should ignore Blessing for now._

_Thanks to citrusella for pointing this out._

Arsnael:

Fixes #10548

### Changes
Prevent the user to be able to cast Healing light as a Healer if he has full health already (50 HP). The check is being made in the common script both used by client and server sides. 

Added some tests cases as well.

Wondering too if it wouldn't be better to send the cast request to the server first before casting it on client side as well (current logic seems to be that it's first being casted on client's side).

EDIT : Apologies for the non-respect of the ESLint syntax... Fixed!

----

### DIFF
--- a/test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
+++ b/test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
@@ -58,6 +58,21 @@ describe('POST /user/class/cast/:spellId', () => {
       });
   });
 
+  it('returns an error if use Healing Light spell with full health', async () => {
+    await user.update({
+      'stats.class': 'healer',
+      'stats.lvl': 11,
+      'stats.hp': 50,
+      'stats.mp': 200
+    });
+    await expect(user.post('/user/class/cast/heal'))
+      .to.eventually.be.rejected.and.eql({
+        code: 401,
+        error: 'NotAuthorized',
+        message: t('messageHealthAlreadyMax'),
+      });
+  });
+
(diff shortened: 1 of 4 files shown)


paglias:
thanks @Arsnael , this is working well! I've notated it towards your second tier

Arsnael:
You're welcome !

PR #10604, opened 2018-08-15
Added smartbanner code to suggest iphone/android apps for mobile users - fixes #9901

### LINKED ISSUE #9901: Add smart app banner to web site
See more - https://github.com/ain/smartbanner.js

flhat:
Fixes #9901

### Changes
- Adds appropriate meta data to show smartbanner for iphone/android
- Includes iTunes / Google Play Store logo images in /static/icons/
- Adds /static/smartbanner/smartbanner.min.css and /static/smartbanner/smartbanner.min.js needed for the component

----


Alys:
@BlkStormy Thanks! Can you include a screenshot please?

flhat:
@Alys Sure - sorry, I forgot to do that with the message:

**iPhone**
iphone

**Android**
android

flhat:
Hmm, I saw that after the pull request was made, it gets run through 2 different tests.  Both failed, but what I'm trying to figure out is if there's a way I can run the tests on my own locally, before doing the request?

TheHollidayInn:
Hey! Looks like your problem is with lint. You can run `npm run lint` but the issues is the new JS file you included. Is there a hosted version of the library on a CDN?

flhat:
@TheHollidayInn Hmm, I don't see one.  There is a NPM package here:  https://www.npmjs.com/package/smartbanner.js  I'm unfamiliar with how to get it to load once installed though...  Thoughts?

Alys:
@BlkStormy In case it's of use later, there's information about the tests in the "Creating and Running Automated Tests" section of [Using Your Local Install to Modify Habitica's Website and API](https://habitica.wikia.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API#Creating_and_Running_Automated_Tests)

TheHollidayInn:
@BlkStormy ah I see. They don't really document how to import for newer front ends.

Let's try
- Go to https://github.com/HabitRPG/habitica/blob/develop/website/client/main.js
- `npm i smartbanner.js`
- Then add `import smartbanner.js from 'smartbanner.js'; // eslint-disable-line
- Then maybe try importing the css like so `@import '~smartbanner.js/dist/smartbanner.min.css';` here https://github.com/HabitRPG/habitica/blob/develop/website/client/app.vue#L105

I'm not exactly sure though. Maybe @paglias has some idea.

And sorry for the odd workaround, but we'd like to keep the dependency separate.

flhat:
@TheHollidayInn No problem, thanks for the suggestion.  I'll play around with it and see what I can do.

paglias:
the steps suggested by @TheHollidayInn should work!

flhat:
Hmm, well, I got it worked out, but not quite sure how to handle the coverage issue...  :/

Alys:
Coveralls does that sometimes. You don't need to worry about it. :)

flhat:
@Alys Oh, okay.  So this pull request is ready for review then?

Alys:
@BlkStormy Good point, I've added the "ready for review" label.

paglias:
So this looks good but I think the icons used should be the Habitica apps icons instead of the google play / app store ones? In which case you can find them in website/static/presskit/Logo

website/client/main.js
```diff
@@ -9,6 +9,7 @@ import getStore from './store';
 import StoreModule from './libs/store';
 import './filters/registerGlobals';
 import i18n from './libs/i18n';
+import smartbanner from 'smartbanner.js/dist/smartbanner'; // eslint-disable-line
```
paglias: I think the `// eslint-disable-line` can be removed

TheHollidayInn: I think that we will get an unused var warning, but definitely double check me.

paglias: You're right... I think then we should just do `import smartbanner.js/dist/smartbanner`

flhat:
@paglias @TheHollidayInn thanks for reviewing!

1) Good call on the icon - updated and remove the 2 new images I had added

2) I changed the import to `import 'smartbanner.js/dist/smartbanner';`

website/client/app.vue
```diff
@@ -103,7 +103,8 @@ div
 
 <style lang='scss'>
   @import '~client/assets/scss/colors.scss';
-
+  @import '~smartbanner.js/dist/smartbanner.min.css';
```
paglias: This should be loaded like other deps here https://github.com/HabitRPG/habitica/blob/develop/website/client/app.vue#L634 so in this case
`<style src="smartbanner.js/dist/smartbanner.min.css"></style>`

website/client/main.js
```diff
@@ -9,6 +9,7 @@ import getStore from './store';
 import StoreModule from './libs/store';
 import './filters/registerGlobals';
 import i18n from './libs/i18n';
+import 'smartbanner.js/dist/smartbanner';
```
paglias: instead of loading the version in /dist which is already compiled using browserify we should load the original version that will then be compiled directly in Habitica, so the file to import should be `smartbanner.js/src/index.js`

paglias (changes requested):
Left two comments, then it should be good to go!

website/client/app.vue
flhat: Got it - I have this ready - thanks!

website/client/main.js
flhat: Hmm, the /src/ directory isn't included in the node module only /dist/ which contains smartbanner.js and smartbanner.min.js

paglias: I see... In that case the non .min version will work

TheHollidayInn: Shouldn't we be able to just import smartbanner.js? It is a module

paglias: Looks like it's mostly published to npm but doesn't follow many conventions and if I remember correctly it doesn't have a main entry in package.json so i don't think it'll work, but we can try

flhat: Updated - just committed the change

paglias:
So I just tested this change and it works well, but I noticed we also have this warning 
schermata 2018-08-20 alle 15 09 38

which I think should be removed if we add this?

@Tressley @TheHollidayInn

paglias:
This is the code to remove

```
diff --git a/website/client/components/header/index.vue b/website/client/components/header/index.vue
index fb7e2965c..c17a24c2a 100644
--- a/website/client/components/header/index.vue
+++ b/website/client/components/header/index.vue
@@ -35,8 +35,6 @@ div
         span.small-text(v-html="$t('inviteFriendsParty')")
         br
         button.btn.btn-primary(@click='createOrInviteParty()') {{ user.party._id ? $t('inviteFriends') : $t('startAParty') }}
-  a.useMobileApp(v-if="isAndroidMobile()", v-once, href="https://play.google.com/store/apps/details?id=com.habitrpg.android.habitica") {{ $t('useMobileApps') }}
-  a.useMobileApp(v-if="isIOSMobile()", v-once, href="https://itunes.apple.com/us/app/habitica-gamified-task-manager/id994882113?mt=8") {{ $t('useMobileApps') }}
 </template>

 <style lang="scss" scoped>
@@ -155,12 +153,6 @@ export default {
     ...mapActions({
       getPartyMembers: 'party:getMembers',
     }),
-    isAndroidMobile () {
-      return navigator.userAgent.match(/Android/i);
-    },
-    isIOSMobile () {
-      return navigator.userAgent.match(/iPhone|iPad|iPod/i);
-    },
     expandMember (memberId) {
       if (this.expandedMember === memberId) {
         this.expandedMember = null;
(END)
```

Tressley:
@paglias / @TheHollidayInn  -- I would like to keep the messaging for now. The smart banner doesn't contain any messaging around the lack of true mobile optimization.

I would, however, like to update the messaging with our visual styles. I believe the current implementation is a carry over from Keith's initial pass to get some mobile features available.

paglias:
Sounds good, in that case we can merge this PR

paglias (approved):

SabreCat:
Thanks for this, @BlkStormy, and welcome to the ranks of the Blacksmiths at tier 1! This PR will undergo some additional testing in a staging environment, and if all is well go live late next week.

Note that further contributor tiers require increasing amounts of effort from one to the next, so tier 2 will require multiple PRs or a larger PR to attain. But keep helping out and we'll express our gratitude accordingly!

PR #10606, opened 2018-08-15
allow challenge leader/owner to view/join/modify challenge in private group they've left - fixes #9753

### LINKED ISSUE #9753: Can't view/join//edit/close/delete a challenge you own if you've left its private guild/party
When a user creates a challenge in a private guild or party and then leaves that group, they are still able to see the challenge under My Challenges or Discover Challenges (which is correct) but they are not able to open the challenge into its own page and so they can't use any of the features that should be available on that page (e.g., to close the challenge). When they try to view the challenge, they see the `challengeNotFound` message ("Challenge not found or you don't have access.")

This is because of code like this which checks to see if the user is able to see the group that the challenge is in:
https://github.com/HabitRPG/habitica/blob/3b4b459e6861fd5e1c1abc5471fff95007559ff1/website/server/controllers/api-v3/challenges.js#L502-L503
That code is correct for the case where the user is not the challenge owner, but it should be bypassed for the owner in both that location and probably everywhere else that we have code checking for group permissions.

Alys:
fixes #9753

When a user creates a challenge in a private guild or party and then leaves that group, they still need to have full control over the challenge. This PR allows that.

### DIFF
--- a/website/server/models/challenge.js
+++ b/website/server/models/challenge.js
@@ -84,13 +84,14 @@ schema.methods.canModify = function canModifyChallenge (user) {
 // Returns true if user can join the challenge
 schema.methods.canJoin = function canJoinChallenge (user, group) {
   if (group.type === 'guild' && group.privacy === 'public') return true;
+  if (this.isLeader(user)) return true; // for when leader has left private group that contains the challenge
   return user.getGroups().indexOf(this.group) !== -1;
 };
 
 // Returns true if user can view the challenge
 // Different from canJoin because you can see challenges of groups you've been removed from if you're participating in them
 schema.methods.canView = function canViewChallenge (user, group) {
-  if (this.isMember(user)) return true;
+  if (this.isLeader(user) || this.isMember(user)) return true;
   return this.canJoin(user, group);
 };
 
--- a/website/server/models/challenge.js
+++ b/website/server/models/challenge.js
@@ -91,7 +91,7 @@ schema.methods.canJoin = function canJoinChallenge (user, group) {
 // Returns true if user can view the challenge
 // Different from canJoin because you can see challenges of groups you've been removed from if you're participating in them
 schema.methods.canView = function canViewChallenge (user, group) {
-  if (this.isLeader(user) || this.isMember(user)) return true;
+  if (this.isMember(user)) return true;
   return this.canJoin(user, group);
 };
 
--- a/website/server/models/challenge.js
+++ b/website/server/models/challenge.js
@@ -66,6 +66,11 @@ schema.statics.sanitizeUpdate = function sanitizeUpdate (updateObj) {
   return this.sanitize(updateObj, noUpdate);
 };
 
+// Returns true if user is the leader/owner of the challenge
+schema.methods.isLeader = function isChallengeLeader (user) {
+  return this.leader === user._id;
+};
+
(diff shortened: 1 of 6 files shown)


Alys:
[sorry, hit enter by accident]

fixes #9753

When a user creates a challenge in a private guild or party and then leaves that group, they still need to have full control over the challenge. This PR allows that.

TheHollidayInn:
Looks good! This only allows leaders to rejoin. Is that okay?

Also, can you add tests for this?

Alys:
I really should add a test for this. I knew there was something I was missing. Not ready for review yet.

Alys:
Oh, you already asked about tests. :) That hadn't loaded before. Yes, certainly sorry. I had a nagging feeling I'd forgotten something when I pushed it!

It's fine that only leaders can rejoin. It would be too hard to work out if someone else trying to join had been a previous participant, and if they've left the challenge and also left the private group, they probably shouldn't be allowed back into the challenge anyway.

TheHollidayInn:
@Alys Np :D  And what I meant was that this PR forces users to rejoin to manager their challenge. I think that is a correct way to implement the feature. I'm just making sure that is what you intended.

Alys:
This is ready for review.

website/server/models/challenge.js
```diff
@@ -66,27 +66,33 @@ schema.statics.sanitizeUpdate = function sanitizeUpdate (updateObj) {
...
   return user.getGroups().indexOf(this.group) !== -1;
 };
 
 // Returns true if user can view the challenge
-// Different from hasAccess because you can see challenges of groups you've been removed from if you're partecipating in them
+// Different from canJoin because you can see challenges of groups you've been removed from if you're participating in them
 schema.methods.canView = function canViewChallenge (user, group) {
-  if (this.isMember(user)) return true;
-  return this.hasAccess(user, group);
+  if (this.isLeader(user) || this.isMember(user)) return true;
```
paglias: the this.isLeader is also checked in canJoin, is it necessary to have it here as well?

Alys: Yes, canView and canJoin are used at different times so they both need the check. canJoin is relevant only when the challenge owner is trying to rejoin their own challenge, not when they're trying to administer it without joining. And this has reminded me that I meant to respond to THI's post about that in the PR's comments.  I'll post a reply in a few minutes.

Alys:
> what I meant was that this PR forces users to rejoin to manager their challenge. I think that is a correct way to implement the feature. I'm just making sure that is what you intended.

Ah no, sorry, I should have mentioned something about that in the PR's top post. There was already some support for a challenge creator administering their challenge after they'd left the private group without having joined the challence, and it still should be possible for them to administer without joining. 

We want it to be optional for a challenge owner to join their own challenge because it's not uncommon for an owner to create a challenge that's solely for the benefit of other people (e.g.. Sabe creating the Take This challenges, and similar things can occur in private groups). In particular, for the case of a challenge that's hosted in a group that the challenge owner is no longer a member of, the most common action is for the owner to close or delete the challenge, and so making them join it first would be pointless from their point of view.

website/server/models/challenge.js
Alys: After a DM from paglias, I looked more closely and canView calls canJoin so the isLeader check here is indeed unnecessary. I've removed it. :)

PR #10614, opened 2018-08-21
always reload completed todos

### LINKED ISSUE #9806: Completed To-Dos disappear if you sync after viewing them
Completed To-Dos behave in a way that appears to be inconsistent and buggy although it's a sort-of-expected side-effect of the way that we don't load completed To-Dos at all until the user clicks on the "Complete" tab in the To-Dos column.

When the user does that, the completed To-Dos are read from the database and they appear after that data is synced to the website. However if the user clicks the website's sync button while they are viewing the "Complete" tab, the completed To-Dos vanish (I assume this is because the full-account sync causes the To-Dos data to be replaced with all active To-Dos and no completed ones). If the user moves to the Active or Scheduled tab and then back to Complete, the completed To-Dos do NOT reappear. The user has to reload the website to be able to view completed To-Dos again.

The reload should not be necessary.

Every time the user goes to the "Complete" tab, their completed To-Dos should always be reloaded from the database.

As well as that, ideally if the user clicks the sync button while the "Complete" tab is active, then the completed To-Dos should be synced along with the uncompleted ones. However it's possible that that might be too difficult to do - I haven't looked into it whether the sync action can get information about whether the "Complete" tab is visible.

----

_It's possible that there are also other bugs as indicated in the messages below from Kella (b50b5e4d-0134-4e3b-9627-565840558f59) in the Report a Bug guild. However we should fix the sync behaviour first and then investigate further to see if more fixes are needed - if so, we can log a new issue once this one is closed._

_Kella: "On the tasks page, in the To-Do's column, when I select the "Complete" tab, the list of tasks goes blank. The to-do's I've completed recently show up in the Data Display tool just fine but after refreshing the page, the Completed list of to-do's is still empty."_

_Alys: "Is it possible it's just taking a while for the completed To-Dos to appear? When I view them, they do appear but only after a couple of seconds. Can you try staying in the completed list for ten seconds to see if that makes a difference?"_

_Kella: "I tried waiting 10 seconds for my completed to-do's to show up and nothing. Since then, I checked another time and only the to-do I completed that day showed up. Then I tried syncing the page, which caused that completed to-do to disappear again. Then I refreshed the webpage, and the to-do's all appeared. This morning, went back to look and all the completed to-do's were gone again. Refreshed the website again and they came back."_

negue:
fixes #9806 

- always reload completed todos if you select the tab
- reload completed todos after you triggered a re-sync

### DIFF
--- a/website/client/components/header/menu.vue
+++ b/website/client/components/header/menu.vue
@@ -396,12 +396,13 @@ export default {
     toggleUserDropdown () {
       this.isUserDropdownOpen = !this.isUserDropdownOpen;
     },
-    sync () {
+    async sync () {
       this.$root.$emit('habitica::resync-requested');
-      return Promise.all([
+      await Promise.all([
         this.$store.dispatch('user:fetch', {forceLoad: true}),
         this.$store.dispatch('tasks:fetchUserTasks', {forceLoad: true}),
       ]);
+      this.$root.$emit('habitica::resync-completed');
     },
     async getUserGroupPlans () {
       this.$store.state.groupPlans = await this.$store.dispatch('guilds:getGroupPlans');
--- a/website/client/components/tasks/column.vue
+++ b/website/client/components/tasks/column.vue
@@ -449,9 +449,9 @@ export default {
     });
 
     if (this.type !== 'todo') return;
-    this.$root.$on('habitica::resync-requested', () => {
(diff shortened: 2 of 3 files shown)


SabreCat:
There's a momentary flicker as the To-Dos disappear and are reloaded, but that's extremely minor--the issue is fixed! Gonna send this in.

PR #10615, opened 2018-08-22
Only notify if pushedItems isn't empty - fixes #10496

### LINKED ISSUE #10496: Newly subscribed accounts receive erroneous notification that they have Mystery Items
If an account is newly subscribed, in at least some cases a notification appears that they have Mystery Items available, even though they do not. 

This one is verified by me. In this specific case the account (which did not have a subscription) was invited to a group plan (thus receiving a 'free' subscription).

flhat:
Fixes #10496

### Changes
- Checked to make sure pushedItems wasn't empty before user.addNotification() is called

----

### DIFF
--- a/website/server/libs/payments/subscriptions.js
+++ b/website/server/libs/payments/subscriptions.js
@@ -37,7 +37,9 @@ function revealMysteryItems (user) {
     }
   });
 
-  user.addNotification('NEW_MYSTERY_ITEMS', { items: pushedItems });
+  if (pushedItems.length !== 0) {
+    user.addNotification('NEW_MYSTERY_ITEMS', { items: pushedItems });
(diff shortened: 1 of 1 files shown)


TheHollidayInn:
Thanks @BlkStormy! Can you add a test around this?

@paglias could this have caused empty notifications?

paglias:
@TheHollidayInn no, I don't think so this would just cause a mystery items notification with no items inside, but the remaining data is still there

flhat:
I haven't written a test yet.  Let me look into how to do that and I'll give it a shot.

paglias:
Any update on this one?

paglias:
Closing for lack of updates

PR #10620, opened 2018-08-24
Social user username

TheHollidayInn:
This generates usernames for users who sign in with social.

I'm not sure about using the unicode characters, but we can't use the suggested uppercase letters because we need unique to be based on lowercase.


paglias:
I think we're using the existence of auth.local.username to distinguish between social and non social users so this may create some issues

TheHollidayInn:
Oh, where we are doing that?

paglias:
Looks like mostly in the auth controller https://github.com/HabitRPG/habitica/blob/develop/website/server/controllers/api-v3/auth.js#L70 and https://github.com/HabitRPG/habitica/blob/2946f0df15ba719c5e1ccc305fb86839e3a91c2b/website/server/controllers/api-v3/auth.js#L490

website/server/libs/auth/generateId.js
```diff
@@ -0,0 +1,9 @@
+import shortid from 'shortid';
+
+shortid.characters('0123456789abcdefghijklmnopqrstuvwxyzⓐⓑⓒⓓⓔⓕⓖⓗⓘⓙⓚⓛⓜⓝⓞⓟⓠⓡⓢⓣⓤⓥⓦⓧⓨⓩ-_');
```
paglias: are the chars after the normal alphabet expected?

test/api/v3/integration/user/auth/POST-user_auth_social.test.js
```diff
@@ -39,15 +39,17 @@ describe('POST /user/auth/social', () => {
...
         authResponse: {access_token: randomAccessToken}, // eslint-disable-line camelcase
         network,
       });
 
       expect(response.apiToken).to.exist;
       expect(response.id).to.exist;
       expect(response.newUser).to.be.true;
+      expect(response.username).to.exist;
       await expect(getProperty('users', response.id, 'profile.name')).to.eventually.equal('a facebook user');
+      await expect(getProperty('users', response.id, 'auth.local.lowerCaseUsername')).to.exist;
```
Alys, line 53: `'a facebook user'` in line 51 would need to be replaced with the generated name.  [EDIT: This and my other comments were made before the purpose of this PR had been explained. The top post in the PR didn't make it clear and we'd previously discussed not using the default profile name that social media sites supplied, so I'd been assuming that's what the PR was for.]

website/server/libs/auth/generateId.js
```diff
@@ -0,0 +1,9 @@
...
+
+function generateUsername () {
```
Alys: Can this be renamed to `generateProfileName`? That will be more accurate and could prevent confusion in future code changes.

website/server/libs/auth/social.js
```diff
@@ -0,0 +1,124 @@
...
+        id: profile.id,
+        emails: profile.emails,
+      },
+      local: {
+        username: generatedUsername,
+        lowerCaseUsername: generatedUsername,
+      },
+    },
+    profile: {
+      name: profile.displayName || profile.name || profile.username,
```
Alys, line 59: I might be misunderstanding, but aren't these `profile....` values supplied by the social media account? If so, they're the ones we want to replace with the generated profile name.

Alys:
Unicode characters are best avoided because they're painful in cases where you need to manually type someone's name. We permit them in names that users choose themselves but I think it would be easier for everyone to not have them in the default name.

We might want to take out the digits since while they're permitted for profile names, they're not very commonly chosen by the players.

Once the code no longer sets auth.local.username, it will be fine to use uppercase characters. `profile.name` is the only thing that needs to be set.

I'm wondering if constructing the profile name from random characters is the best approach - some of the names will look pretty weird. Are there libraries for generating random names from two or more words concatenated? E.g., there's an image sharing website that does random image URLs like "HungryBadger.png" so maybe there's freely available libraries to do something like that. Random adjective + noun might create some fun, pronouncable profile names, although mods would want to vet the adjectives used by the library. :)

TheHollidayInn:
The idea here is not to generate profile names, it is to generate usernames. And these are intended to be changed but used to help with notifications if a user has not selected a username.

website/server/libs/auth/social.js
TheHollidayInn: This is copied from existing code. Nothing I wrote.

website/server/libs/auth/generateId.js
TheHollidayInn: Generating username is the intention here.

test/api/v3/integration/user/auth/POST-user_auth_social.test.js
TheHollidayInn: That's the existing test. Nothing I'm adding here.

Alys:
Okay, so it's not a fix for the privacy problem of displaying the user's real name by default? It's for a yet-to-be-built feature for notifying users if their messages have been replied to in Tavern/Guild/Party chat?

TheHollidayInn:
I actually thought we had the privacy issue on the username. So, no this doesn't fix that. And yes, this is for future use: invite by username and mentioning (with notifs).

I can fix the privacy issue in a separate PR as well.

Alys:
Okay, cool, then upper-case characters are undesirable as you say since we need the names to be unique (easily to ensure that if it's all lower-case and numbers). I still think we should avoid unicode just in case there's ever a need for a mentions-name to be typed manually (e.g., in third-party tools). 

I still think starting the user with a pronounceable mentions-name will help. Again, it makes it easier to type if it's not random characters, and also to read after inviting someone if you later want to confirm that you have invited the correct person.

Since we'll be generating profile names as well, would it make sense to use the same generated name for both profile name and mentions name?

TheHollidayInn:
Yea, I need something to replace the uppercase characters. I also am not a fan of the unicode, but the libraries algorithm requires 64 characters.

These usernames are not meant to be used. So, I we don't intent on support easy use of generated usernames. The users can copy and paste if needed, but the idea is that when a user has trouble, they should edit the username. We also have plans to think about ways to promote username editing.

Generating pronounceable names is hard to do and stay unique. So I'm open if you have an algorithm/library for this, but I'll definitely do it separately for profile names as to not hold this back.

paglias:
I haven't checked it in depth but there's this library https://github.com/oculus42/short-uuid that doesn't require the alphabet to be 64 characters

TheHollidayInn:
Nice! I'll read over it

TheHollidayInn:
It was discussed on call to represent habitica. Up to you if you want to keep it or not.

phillipthelen:
Considering we changed the plan to require users to confirm/change/set their username, do we think this is still necessary? Since we won't make usernames public without a users approval, I don't think we will have to set it to anything as long as we allow users to set a username without an existing one.

PR #10624, opened 2018-08-24
Added Contributor Titles to Names on Hover (fixes #10611)

### LINKED ISSUE #10611: Contributor titles don't show in full on hover over player's name
When you hover over a contributor's name in chat, you should see, for example, "Friend Socialite". Currently you just see "Socialite". The tier types are "Friend", "Elite", "Champion", Legendary", "Guardian", "Heroic". You can see code related to them here:
https://github.com/HabitRPG/habitica/blob/1eb1fe76a8917cf5cd67d317fb91f577e7470d51/website/common/script/libs/achievements.js#L9-L32

And in fact that function might be useful for fixing this bug although I don't guarantee that since I haven't looked into it in detail. :)

For cases where the contributor has more than one tier type, you should see, for example, "Elite Socialte, Comrade" (i.e., the "Friend"/"Elite"/etc portion doesn't need to be repeated for each tier type.

legitmaxwu:

Fixes #10611

### Changes

When you hover on a person's name in chat, their Contributor Title will now be included in the tooltip. Ex: In the picture below, only "TestTitle" would be displayed in the past. Now, "Legendary" is also appended to the front.
image

This was done by editing `ChatCard.vue`, adding a computed property in the file to find the prefix. I also added a function to `achievements.js` to allow the computed property to access the `contribText()` function in `achievements.js`.

----
UUID: 215e51f6-beb1-4617-b359-fcb09681e6fd


website/common/script/libs/achievements.js
```diff
@@ -313,6 +313,12 @@ achievs.getAchievementsForProfile = function getAchievementsForProfile (user, la
       achievements: _getSpecialAchievements(user, language),
     },
   };
+
+  return result;
+};
+
+achievs.updateContribText = function updateContribText (user, language) {
```
paglias: I think getContribText would be a better name for this function here

paglias: Actually, this function seems to only call another function so maybe instead of creating a new one you could just make the original `contribText` function avaiable externally as well?

So achievs.getContribText = contribText; ?

website/client/components/chat/chatCard.vue
```diff
@@ -7,7 +7,7 @@ div
       h3.leader(
         :class='userLevelStyle(msg)',
         @click="showMemberModal(msg.uuid)",
-        v-b-tooltip.hover.top="('contributor' in msg) ? msg.contributor.text : ''",
+        v-b-tooltip.hover.top = "('contributor' in msg) ? tierTitle : ''",
```
paglias: looks like `contributor in msg` is always true so this check can be removed, `v-b-tooltip.hover.top="tierTitle"` will work

website/client/components/chat/chatCard.vue
```diff
@@ -221,6 +223,11 @@ export default {
       }
       return this.icons[`tier${message.contributor.level}`];
     },
+    tierTitle () {
+      const message = this.msg;
+      const isContributor = Boolean(message.contributor && message.contributor.level);
```
paglias: I don't think this check is necessary as the ´contribText´ function makes the same exact change so in case the user is not a contributor it'll just return undefined and at that point you can return an empty string: `return achievementsLib .getContribText(...) || ''`

paglias (changes requested):
Left some comments but nothing major!

legitmaxwu:
@paglias Made the fixes! Some of these adjustments were so obvious in hindsight and I should've addressed them before submitting in the first place.

Thank you so much for the comments and patience.

legitmaxwu:
Note: the latest update is the correct version (no need to check the previous commits). I was being stupid and kept committing broken versions of code.

paglias (commented):
@legitmaxwu thanks for the update, the PR is looking good, the only problem is the .project file that got committed, can you remove it?

legitmaxwu:
Oops my bad. Just deleted it!

SabreCat:
Thanks for this, @legitmaxwu, and welcome to the ranks of the Blacksmiths at tier 1! This will go into a staging environment for a few days of additional testing, then if all is well go live in production late next week.

Note that further tiers require increasing amounts of effort from one to the next, so tier 2 will require multiple PRs or a larger PR to attain. But keep helping out and we'll express our gratitude accordingly!

PR #10625, opened 2018-08-25
Validated todo date

### LINKED ISSUE #8405: API: dates may be invalid

TheHollidayInn:
Fixes #8405

### DIFF
--- a/website/server/libs/taskManager.js
+++ b/website/server/libs/taskManager.js
@@ -89,6 +89,8 @@ export async function createTasks (req, res, options = {}) {
     // Validate that task.type is valid
     if (!taskData || Tasks.tasksTypes.indexOf(taskData.type) === -1) throw new BadRequest(res.t('invalidTaskType'));
 
+    if (taskData.date && !moment(taskData.date).isValid()) throw new BadRequest(res.t('invalidTaskDate'));
+
     let taskType = taskData.type;
     let newTask = new Tasks[taskType](Tasks.Task.sanitize(taskData));
 
--- a/test/api/v3/integration/tasks/POST-tasks_user.test.js
+++ b/test/api/v3/integration/tasks/POST-tasks_user.test.js
@@ -416,6 +416,19 @@ describe('POST /tasks/user', () => {
   });
 
   context('todos', () => {
+    it('errors if date is invalid', async () => {
+      await expect(user.post('/tasks/user', {
+        text: 'test todo',
+        type: 'todo',
+        notes: 1976,
+        date: 'randome-date',
+      })).to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: t('invalidTaskDate'),
+      });
+    });
+
     it('creates a todo', async () => {
       let task = await user.post('/tasks/user', {
         text: 'test todo',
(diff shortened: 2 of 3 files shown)


website/server/libs/taskManager.js
```diff
@@ -89,6 +89,8 @@ export async function createTasks (req, res, options = {}) {
     // Validate that task.type is valid
     if (!taskData || Tasks.tasksTypes.indexOf(taskData.type) === -1) throw new BadRequest(res.t('invalidTaskType'));
 
+    if (taskData.date && !moment(taskData.date).isValid()) throw new BadRequest(res.t('invalidTaskDate'));
```
paglias, line 92: Should the check go in the model? So it's applied on updates as well

TheHollidayInn: There is a note on the model about needing a migration. I was going to add that as well.

TheHollidayInn: This breaks earlier though, so could be nice.

paglias: Ah yes this needs a migration for existing data...

paglias: But once the migration is done I think the check should be moved to the model, so we don't have duplicates nor have to remember to update any route that changes the date

TheHollidayInn: Hmm. I'm still in favor of validating before we save. Save happens pretty late in a request. There is probably a better way to implement it using a Policy pattern.

paglias: Checking in the model, apart from being the standard we use for every other field which would reduce complexity, means that there's only one place where we have to worry about checking for validity and if all db ops go through the model validation (and they should) means we don't have to worry about invalid data popping up in the database, a problem we had in multiple occasions.

It also means that if we add new route in 3 years that change the date we have to hope someone remembers about this check and add it there to.

If we're worried about the validation coming too late (and I can see the problem in this case since we do some db ops before saving) we can alway call `task.validateSync()` as soon as it's been created

paglias: Actually we already call validateSync in createTasks, just a bunch of lines after the check you added https://github.com/HabitRPG/habitica/pull/10625/files#diff-cb32c22fd9c0bad615b785ed3f988796R113

TheHollidayInn: If you look right above this line, we validate the type before saving. I don't disagree that we want to rember the always validate. There are multiple patterns to do so. Moving the code out of the route and have a task repository to validate tasks input (with unit tests) is another option.

Anyway it was just a thought.

paglias: I'm fine with moving the validation outside the model, I was just a bit concerned about having all the validations there except this one. Yes, the type is validated above but mostly because the call `new Tasks[taskType](Tasks.Task.sanitize(taskData));` must be passed a valid type or it'll throw

TheHollidayInn: We should probably move to something like this: https://github.com/i0natan/nodebestpractices/blob/master/sections/errorhandling/failfast.md

But to keep the change small, I agree with you and think moving the validation to the model layer to running the migration is better. My fix is currently a hack because we don't standardize like the above.

PR #10628, opened 2018-08-25
Check user version before adding notifications

### LINKED ISSUE #9242: Guild Notifications and Party Notifications sometimes do not automatically clear.
Notifications for Guild and Party messages don’t reliably clear when you click them and go to the guild/party they are from. I’m pretty sure that Party never clears, and Guild notifications don’t clear if you go from one Guild to another, but we should definitely carefully investigate.

TheHollidayInn:
Fixes #9242

### DIFF
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -103,7 +103,7 @@ div
 
 <style lang='scss'>
   @import '~client/assets/scss/colors.scss';
-  
+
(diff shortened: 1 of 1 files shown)


TheHollidayInn:
@paglias does this look good to you?

paglias (approved):
Looks good!

paglias:
This has been reverted because it made notifications never go away from the panel unless the user used the sync button

PR #10629, opened 2018-08-25
Prevented users with lvl less than 10 from seeing mana

### LINKED ISSUE #8893: mana notifications are shown to users under level 10
Under some circumstances, a player below level 10 can see a notification showing that mana has been removed. Mana should be invisible to the player until they reach level 10 and choose a class (not visible if they opt out of the class system).

I don't yet know what triggers the notification but I've just rebirthed so I'll watch for it.

This was reported by _Meakuel (4382653f-4dbe-4f0f-84dd-6127c8a6b466) in the Report a Bug guild: "I saw the message "-12 Mana". I am still not level 10 and this time it was the website version." They'd previously reported gaining mana on the Android app (https://github.com/HabitRPG/habitica-android/issues/793).

EDIT:

It seems that the mana notification appears when mana is removed, rather than when it's given. That would be a good place to start investigating what fix is needed.

TheHollidayInn:
Fixes #8893

### DIFF
--- a/website/client/components/notifications.vue
+++ b/website/client/components/notifications.vue
@@ -200,7 +200,7 @@ export default {
       return !this.user.flags.classSelected && this.user.stats.lvl >= 10;
     },
     userHasClass () {
-      return this.user.stats.lvl >= 10 && this.user.flags.classSelected && !this.user.preferences.disableClasses;
+      return this.$store.getters['members:hasClass'](this.user);
     },
     invitedToQuest () {
       return this.user.party.quest.RSVPNeeded && !this.user.party.quest.completed;
@@ -253,11 +253,9 @@ export default {
     },
     userMp (after, before) {
       if (after === before) return;
-      if (!this.$store.getters['members:hasClass'](this.user)) return;
+      if (!this.userHasClass) return;
 
       const mana = after - before;
-
-      if (!this.userHasClass) return;
       this.mp(mana);
     },
     userLvl (after, before) {
--- a/website/client/components/notifications.vue
+++ b/website/client/components/notifications.vue
@@ -509,7 +509,7 @@ export default {
           case 'CRON':
             if (notification.data) {
               if (notification.data.hp) this.hp(notification.data.hp, 'hp');
-              if (notification.data.mp && !this.userHasClass) this.mp(notification.data.mp);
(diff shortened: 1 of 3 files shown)


Alys:
Ah sorry, the title of that issue is a bit misleading. This from the description is the important part:

"Mana should be invisible to the player until they reach level 10 and choose a class (not visible if they opt out of the class system)."

I.e., Mana should not be visible unless `flags.classSelected` is true and `preferences.disableClasses` is false.

TheHollidayInn:
Ah that makes sense. I'll add that in, thanks!

website/client/components/notifications.vue
```diff
@@ -199,6 +199,9 @@ export default {
     userClassSelect () {
       return !this.user.flags.classSelected && this.user.stats.lvl >= 10;
     },
+    userHasClass () {
+      return this.user.stats.lvl >= 10 && this.user.flags.classSelected && !this.user.preferences.disableClasses;
```
Alys: Love it! That will be useful in lots of other places. It's probably best to leave out the level check though because a level 10 player with a class who dies will be level 9 but still have their class.

Alys: Sorry, accidentally deleted a bunch of words while editing and didn't notice. :)

Love it! That will be useful in lots of other places. It's probably best to leave out the level check though because a level 10 player with a class who dies will be level 9 but still have their class.

TheHollidayInn: :) why leave out the level though?

paglias: like in the other PR, we should probably use the hasClass getter in client/store/getters/members

TheHollidayInn: Ah didn't realize we had a getter.

website/client/components/notifications.vue
```diff
@@ -506,7 +509,7 @@ export default {
           case 'CRON':
             if (notification.data) {
               if (notification.data.hp) this.hp(notification.data.hp, 'hp');
-              if (notification.data.mp) this.mp(notification.data.mp);
+              if (notification.data.mp && !this.userHasClass) this.mp(notification.data.mp);
```
paglias: shouldn't this be `&& this.userHasClass?`

paglias:
Removed the .only from the test and merged

PR #10633, opened 2018-08-25
add top margin on all modals to prevent overlap with nav/messaging - fixes #10475

### LINKED ISSUE #10475: Modal positioning wrong in specific cases

flhat:
Fixes #10475

### Changes
- Added 100px top margin on all modals to prevent overlap issues with the nav and message banner

screen shot 2018-08-24 at 9 34 12 pm

----

### DIFF
--- a/website/client/assets/scss/modal.scss
+++ b/website/client/assets/scss/modal.scss
@@ -13,6 +13,7 @@
 
 .modal-dialog {
   width: auto;
+  margin-top: 100px;
 
   .title {
     min-height: 24px;
--- a/website/client/assets/scss/modal.scss
+++ b/website/client/assets/scss/modal.scss
@@ -13,7 +13,7 @@
 
 .modal-dialog {
   width: auto;
-  margin-top: 100px;
+  margin-top: 60px;
 
   .title {
     min-height: 24px;
@@ -36,6 +36,10 @@
     overflow-y: hidden;
     text-overflow: ellipsis;
   }
+}
+
+.restingInn .modal-dialog {
+  margin-top: 100px;
 }
 
 #start-quest-modal, #buy-quest-modal {
--- a/website/client/assets/scss/modal.scss
+++ b/website/client/assets/scss/modal.scss
@@ -11,9 +11,12 @@
   }
 }
 
+.modal {
+  z-index: 1300;
+}
+
 .modal-dialog {
   width: auto;
-  margin-top: 60px;
 
   .title {
     min-height: 24px;
@@ -36,10 +39,6 @@
     overflow-y: hidden;
     text-overflow: ellipsis;
   }
-}
-
-.restingInn .modal-dialog {
-  margin-top: 100px;
 }
 
 #start-quest-modal, #buy-quest-modal {


paglias:
@BlkStormy thanks, can you post a screenshot of how this looks without the "checked into the inn" blue banner? Would it be possible to apply the extra margin only when the banner is present?

negue:
This way all modals would stay at the same position. 

@Tressley what do you think?

Avatar Modal checked into the inn:
avatar modal checked into the inn


Avatar Modal without:
image

Tressley:
Is there a reason we can't have the modal positioned in the same spot relative to the navigation? I am not a fan of how the modal overlays the navigation when the user is checked into the inn.

negue:
@BlkStormy could you try the idea I mentioned in
- https://github.com/HabitRPG/habitica/pull/10513#issuecomment-404268215
- https://github.com/HabitRPG/habitica/pull/10513#issuecomment-404568554

flhat:
Sorry for the radio silence on this, I was gone for a few days.  I've made the update and I think this works well:

screen shot 2018-08-30 at 1 44 06 pm

screen shot 2018-08-30 at 1 44 18 pm

Tressley:
I think this is the right fix.

paglias:
Hi @BlkStormy thanks for the help here, I just tested this locally and while it works in some situations there are still some problem, in particular:

1. As mentioned in the comments linked by @negue (https://github.com/HabitRPG/habitica/pull/10633#issuecomment-417361160) the .restingInn class is only applied to the menu and header, so this change only affects modal that are defined inside the header/menu of the website. For example if you go to the market while resting in the inn and try to buy an item you'll see that the modal isn't positioned correctly
2. This is less important and not a big deal if you can't find the source but if you follow the instructions to reproduce the bug at the top of https://github.com/HabitRPG/habitica/issues/10475 you'll see that even with your fix the modals seem to be positioned behind the app menu, so when you scroll them they disappear behind the menu

flhat:
@paglias Got it - I ended up backing out the margin changes and updating the z-index on modals.  The scenarios that I've seen so far all check out on this end.

paglias:
schermata 2018-09-04 alle 12 23 19
schermata 2018-09-04 alle 12 23 29

^ This is how it looks now.

So the issue with the modal being under the navigation is fixed, although I think @Tressley wanted to have all the modals start a bit lower?

Tressley:
Yeah, the modals need to be lower. They should be roughly ~48px from the bottom of the navigation.

paglias:
Any update on this one?

paglias:
Closing for lack of updates

PR #10636, opened 2018-08-25
prevent Quest progress message in Party chat when user is Resting in the Inn

Alys:
Currently, if you're on a Quest and are Resting in the Inn, when cron runs it posts a message to Party chat: "Player attacks Boss for 0.0 damage. Boss attacks party for 0.0 damage." The attack numbers are always zero and user.party.quest.progress is not changed, so the bug is just that the message is posted. 

This PR fixes that.

The bug is a side-effect of https://github.com/HabitRPG/habitica/pull/10577 - my fault, I'm sorry. :(

### DIFF
--- a/test/api/unit/models/group.test.js
+++ b/test/api/unit/models/group.test.js
@@ -251,17 +251,20 @@ describe('Group Model', () => {
           let [
             updatedLeader,
             updatedParticipatingMember,
+            updatedSleepingParticipatingMember,
             updatedNonParticipatingMember,
             updatedUndecidedMember,
           ] = await Promise.all([
             User.findById(questLeader._id),
             User.findById(participatingMember._id),
+            User.findById(sleepingParticipatingMember._id),
             User.findById(nonParticipatingMember._id),
             User.findById(undecidedMember._id),
           ]);
 
           expect(updatedLeader.stats.hp).to.eql(42.5);
           expect(updatedParticipatingMember.stats.hp).to.eql(42.5);
+          expect(updatedSleepingParticipatingMember.stats.hp).to.eql(42.5);
           expect(updatedNonParticipatingMember.stats.hp).to.eql(50);
           expect(updatedUndecidedMember.stats.hp).to.eql(50);
         });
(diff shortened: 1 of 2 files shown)


website/server/models/group.js
```diff
@@ -1101,7 +1103,7 @@ schema.statics.processQuestProgress = async function processQuestProgress (user,
 
   let questType = quest.boss ? 'Boss' : 'Collection';
 
-  await group[`_process${questType}Quest`]({
+  await group[`_process${questType}Quest`]({ // processBossQuest, processCollectionQuest
```
Alys: I added this comment because without it there's no way to find where those two functions are called unless you remember that the function names are created on the fly.

paglias: sounds good but I think it should have an underscore in front of the methods so if you're searching for the full method name like `_ processBossQuest` thill will get matched! Then feel free to merge

Alys: Ha, of course, you'd think I would have thought of that. :) Thanks! I've changed it and I'll merge when the tests pass.

Alys:
A test was failing because the test party now includes an extra member (the sleeping quest participant) so I fixed that test and also adjusted a bunch of others to include the new member, even when it wasn't strictly necessary to make the tests pass. I think it's good to include a sleeping user in those tests.

I haven't merged because this latest commit isn't a trivial change. I think it's okay though. If the tests are still failing tomorrow, I'll fix them again then (bed time here now).

Alys:
Tests are passing. Let me know if you'd like me to change any of them!

test/api/unit/models/group.test.js
```diff
@@ -532,8 +539,10 @@ describe('Group Model', () => {
           let [
             updatedLeader,
             updatedParticipatingMember,
+            updatedSleepingParticipatingMember,
```
paglias, line 542: `updatedSleepingParticipatingMember ` matches the third element array but `User.findById(sleepingParticipatingMember._id), ` is the second, not third element

paglias (commented):
Left one comment, not sure if it's an error or not

test/api/unit/models/group.test.js
Alys: Ah yeah, you're right. It wouldn't have caused the test to fail since all three members should have the same data but it was definitely wrong and inconsistent. I've pushed a fix. Thanks!

PR #10638, opened 2018-08-26
changed @mentioning-auto-complete from text- to caret-based. (Fixes #9694)

### LINKED ISSUE #9694: issues with the drop-down that appears when @mentioning an other user
### General Info
  * UUID: c073342f-4a65-4a13-9ffd-9e7fa5410d6b
  * Browser: Google Chrome Versione 62.0.3202.94 (Build ufficiale) (a 64 bit)
  * OS: Windows 7

### Description
I wrote this in Report a Bug guild and Alys directed me here.
_The list of tags doesn't disappear if you are adding a second mention and then delete the @. And it doesnt appear if you are adding a mention before an other one. But if you have a list of tags that don't want to go away and type before an other mention, then it goes away. If needs more explanation I will try, maybe with screenshots..._

So, issues with a second mention after first one (example `@Alys @Lemoness`)
 - first scenario: starting to type the second mention then changing idea and deleting the mention and the `@`, the list of names stay there hanging. It happens also if you click on a name in the list and then decide to delete it, the list reappear while deleting the letters and then stay there.
 - second scenario: you want to add a second mention before the first one (one of those posts in whichc you write to multiple people and then want to address the first paragraph also to someone else), the list of names doesn't appear.
A thing I noticed is that if you have the list of names hanging there and you type before an other mention, then it disappear.

### Note

### LINKED ISSUE #10646: List of possible usernames to mention doesn't close
From Il Mago in the Bug guild, using Safari on iOS. Reproduced in Chrome on Windows 10.

> Bug: When you type a @, there is the list of names that appear, but if you delete the @ it stays there and disappear only when typing a space
> if the `@` is the first character you type in then deleting it makes the list of names disappear. If you instead type it immediately after an other character, (for example `hi@`) then the list disappear only typing in a space or deleting till the previous space

legitmaxwu:

Fixes #9694
Fixes #10646

### Changes
The bugs described in #9694 were mainly due to the `@mentioning` autocomplete only checking the end of an inputted message for possible tags. Autocomplete suggestions were also only added when the message text itself was edited.

I changed the code so that the autocomplete suggestions update based on caret position instead, similar to how Facebook handles `@mentioning`. The autocomplete suggestions should now be reactive to everything (clicking around on the inputted text, using arrow keys to move the caret, etc.), not just the text itself being updated. This change addresses the bugs in #9694.

----
UUID: 215e51f6-beb1-4617-b359-fcb09681e6fd

### DIFF
--- a/.project
+++ b/.project
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>habitica</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+	</buildSpec>
+	<natures>
+	</natures>
+</projectDescription>
--- a/website/client/components/chat/autoComplete.vue
+++ b/website/client/components/chat/autoComplete.vue
@@ -99,7 +99,6 @@ export default {
         this.searchActive = true;
         startIndex++;
       } else return currIndex;
-      console.log(startIndex);
       return startIndex;
     },
   },
--- a/website/client/components/chat/autoComplete.vue
+++ b/website/client/components/chat/autoComplete.vue
@@ -48,18 +48,17 @@ export default {
   },
   watch: {
     carPos (newCarPos) {
-      let index = newCarPos - 1;
+      let startIndex = newCarPos - 1;
       let newText = this.text;
-      if (index >= newText.length) index = newText.length - 1;
-      // if ((!newText[index] || newText[index] === ' ') && (!newText[index + 1] || (newText[index + 1] && newText[index + 1] === ' '))) {
-      //   this.searchActive = false;
-      //   return;
-      // }
+      if (startIndex >= newText.length) startIndex = newText.length - 1;
+      if (!newText[startIndex] || newText[startIndex] === ' ') {
+        this.searchActive = false;
+        return;
+      }
       this.searchActive = false;
 
-      let startIndex = index;
       while (newText[startIndex] !== '@' && startIndex >= 0) {
-        // if (!newText[startIndex] || (newText[startIndex] === ' ' && newText[startIndex + 1] === ' ')) return;
+        if (newText[startIndex] === ' ') return;
         startIndex--;
       }
       if (newText[startIndex] === '@') {
@@ -92,8 +91,7 @@ export default {
       }
     },
     select (result) {
-      let index = this.carPos;
-      let startIndex = index;
+      let startIndex = this.carPos;
       while (this.text[startIndex] !== '@' && startIndex >= 0) {
         startIndex--;
       }
--- a/website/client/components/chat/autoComplete.vue
+++ b/website/client/components/chat/autoComplete.vue
@@ -51,13 +51,15 @@ export default {
       let index = newCarPos - 1;
       let newText = this.text;
       if (index >= newText.length) index = newText.length - 1;
-      if (!newText[index] || newText[index] === ' ') {
-        this.searchActive = false;
-        return;
-      }
+      // if ((!newText[index] || newText[index] === ' ') && (!newText[index + 1] || (newText[index + 1] && newText[index + 1] === ' '))) {
+      //   this.searchActive = false;
+      //   return;
+      // }
+      this.searchActive = false;
+
       let startIndex = index;
       while (newText[startIndex] !== '@' && startIndex >= 0) {
-        if (this.text[startIndex] === ' ') break;
+        // if (!newText[startIndex] || (newText[startIndex] === ' ' && newText[startIndex + 1] === ' ')) return;
         startIndex--;
       }
       if (newText[startIndex] === '@') {
--- a/website/client/components/chat/autoComplete.vue
+++ b/website/client/components/chat/autoComplete.vue
@@ -48,23 +48,15 @@ export default {
   },
   watch: {
     carPos (newCarPos) {
-      let startIndex = newCarPos - 1;
+      let index = newCarPos - 1;
       let newText = this.text;
-      if (startIndex >= newText.length) startIndex = newText.length - 1;
-      if (!newText[startIndex] || newText[startIndex] === ' ') {
+      if (index >= newText.length) index = newText.length - 1;
+      if (!newText[index] || newText[index] === ' ') {
         this.searchActive = false;
         return;
       }
       this.searchActive = false;
-
-      while (newText[startIndex] !== '@' && startIndex >= 0) {
-        if (newText[startIndex] === ' ') return;
-        startIndex--;
-      }
-      if (newText[startIndex] === '@') {
-        this.searchActive = true;
-        startIndex++;
-      } else return;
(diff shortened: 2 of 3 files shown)


legitmaxwu:
added a sliver of code to slightly optimize performance. when checking for an '@' symbol, the while loop will exit if it encounters a space (names are not supposed to have spaces in them).

I noticed that some names in the tavern actually do have a space in them though. How does that happen? Were they specially assigned?

legitmaxwu:
Sorry for adding so many commits. Is there any way to run the travis-ci tests locally?

Alys:
Profile names are allowed to have spaces in them and many do.

Alys:
The full test suite can be run locally. I'm on my phone so I can't find a link easily but look at the section about tests in the wiki page Using Your Local Install To.... (search those words to find the page).

legitmaxwu:
Oh snap, I will edit my code then to account for spaces. The original code did not handle spaces for tagging so I was confused there.

Alys:
Ah yeah, I should have said, sorry, it's a known problem that @ tagging doesn't work the same for players with spaces in their profile names, and since it's a pre-existing problem that's not directly related to the issue you're solving, I don't think you should feel obliged to fix it too.

In particular, if a fix to handle spaces in names turns out to make the message box less responsive / slower, I think we'd prefer to not have the fix. Also, at some future time, probably not too far away, there'll be a different kind of change that will remove the problem of @ tagging someone with spaces in their profile name, so it might be best to ignore it now. :)

Alys:
But thank you for considering it and being willing to fix it!

legitmaxwu:
@Alys maybe we could save the space problem for another day. When you change your profile name to something like "Name____123" (with more than 1 space), it shows as 1 space on the Chat Card, but when someone wants to mention you, they actually need to type extra spaces.

I think it would be good to add code to reduce repeated spaces down to 1 space during profile-name-changing itself. So if I wanted to change my name to "Name____234," the system wouldn't let me and automatically make it "Name 234." (This is done on the Chat Card anyways so it shouldn't hurt, right?)

If this is done, then we can come back to this issue and fix it to handle spaces. I wrote code to handle spaces already, but if I change my name to "test2___xD" and "test2 xD" and send a message under each of these names, then the autoselection treats the names as different and we have duplicates as shown below. (They still show up in the dropdown as "test2 xD," but one of them is actually "test2___xD" with extra spaces.

image

Basically, it looks kind of weird so I think we should just not handle spaces for now.

(Pretend underscores are spaces, Github won't let me put multiple spaces in a row either)

Alys:
> I think we should just not handle spaces for now.

I agree! I think that's the best approach.

The other change that I hinted at earlier will probably make the profile name space problem completely redundant so collapsing 2+ spaces to 1 won't be necessary then. Sorry about the air of mystery that I'm adding to this but that other change is still in progress. :)

legitmaxwu:
Coolio, I guess that the current pull request is ready for review then!

@Alys haha now I'm curious.

legitmaxwu:
accidentally used two variables for something that needed only 1, so i fixed it.

website/client/components/chat/autoComplete.vue
```diff
@@ -80,7 +91,12 @@ export default {
       }
     },
     select (result) {
-      let newText = this.text.slice(0, this.currentSearchPosition + 1) + result;
+      let startIndex = this.carPos;
+      while (this.text[startIndex] !== '@' && startIndex >= 0) {
```
negue: could you extract the search for the correct startIndex to a method?

website/client/components/chat/autoComplete.vue
```diff
@@ -80,7 +91,12 @@ export default {
...
+        startIndex--;
+      }
+      startIndex++;
+      let newText = this.text.slice(0, this.currentSearchPosition) + result + this.text.slice(this.currentSearchPosition + index - startIndex);
```
negue: did you had another variable `index` during development? cause it doesn't exist anymore ^^

```
Uncaught ReferenceError: index is not defined
    at VueComponent.select (webpack-internal:///2130:103)
    at click (webpack-internal:///2348:18)
    at invoker (webpack-internal:///27:2022)
    at HTMLDivElement.fn._withTask.fn._withTask (webpack-internal:///27:1821)
```

negue (changes requested):
could you fix those two? then I'll take a look again

legitmaxwu:
@negue fixed!

.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
negue, line 1: could you remove this file ?

negue:
I have two issues:

1. press `@` 
2. select a name
3. the textarea has los the focus, so you cannot begin to write your message right after


the other issue, if you have a user named `test`:

1. the message already contains a mention`@otherUser hi`
2. you press `@` before the other mention, write `te`
2. select the user `test`
3. you'll see `@testte`

legitmaxwu:
@negue fixed both errors!

However, the caret will always move to the end of the text when the textarea refocuses. I could not figure out how to move the caret to where it was before the user selects the autocompletion.

One note is that putting a setSelectionRange() function call in the selectedAutoComplete function does not work because it is overridden by some other function that I couldn't identify. If you try this approach, the caret will only move to the right position if selecting the autocomplete does not update any text. (i.e. typing in @testUser and then clicking "textUser" as a selection).

Any ideas on how to fix this? Of course, most tags will happen at the end of the text message anyways so this is not a huge issue.

legitmaxwu:
Hello, I was wondering if there are any updates/feedback on this pull request?

.project
paglias: yeah this file should be removed

website/client/components/groups/chat.vue
```diff
@@ -135,8 +140,9 @@
         }, 500);
       },
 
-      selectedAutocomplete (newText) {
+      async selectedAutocomplete (newText) {
```
paglias, line 143: why `async` here? there are no `await`ed calls

paglias (changes requested):

paglias:
Hey @legitmaxwu sorry for the delay, I've left two comments but I'm having some issues replicating the bugs in the first place, the list of usernames always disappear for me. Also can you give a look at the conflict?

paglias:
I was able to reproduce the issue and after fixing the conflicts your PR works well! 

Regarding `setSelectionRange` being overridden isn't that due to `updateCarretPosition`?

I think the issues comes from mixing Vue's code (with its own DOM modification) and normal DOM functions like `appendChild`, ... it's usually fixed by using for everything Vue's directives

paglias:
Hi @legitmaxwu , I was wondering if you had any update on the status of this PR. thanks!

paglias:
I'm going to close this but if you want to work on it just let me know! Also since it's some good work I'll link to it in the issues in case it's useful for some future fixes

PR #10639, opened 2018-08-26
Issue 10414 - Remove Member Option from View Party link.

### LINKED ISSUE #10414: Remove Member option doesn't appear when clicking View Party from main tasks page

### Description

When viewing a party's member list from the Party page proper, the party leader has the option to remove party members. This option is not, however, available from the View Party button on the main Tasks page, despite other options like Send Message and Promote to Leader showing up OK there.

titchimoto:

Fixes #10414 

### Changes

From some initial investigating, it appeared that this issue was occurring due to some permissions issues, because in `index.vue` when it's retrieving the `members-modal` by clicking the 'View Party' link, it's retrieving the `party` from the `this.user` object, which is returning a sort of condensed version of the party/group object? But it doesn't contain some of the details needed for some of the permission checks. 

I've tried to get around this by whenever this link is clicked, it does an `async/await` to retrieve the specific full party details and passes those details to the `members-modal` meaning that those permissions checks for 'Remove Member' etc. can occur and display (i.e only for Group Leaders). This should also fix this issue occurring on other pages like 'Challenges' where the header/index is.

Not being totally familiar with the entire project, I'm not sure if this might cause other issues, but hopefully nothing obvious I'm missing! The `async/await` should only occur when the link is clicked and so shouldn't make too much of an impact at all on performance.

----
UUID: 0168bfd7-3cc2-4017-92bf-7aeb168e3b8e


TheHollidayInn:
Thanks @titchimoto!

I think we've changed this part recently because of loading the party too many times @paglias, does that sound familiar?

If so, can you add a check to ensure we don't load the party twice?

negue:
The change does fix the issue 👍 , but it also loads the party each time you open it from the header. 

Like @TheHollidayInn already mentioned, could you just add a check to see if the party has been already loaded

website/client/components/header/index.vue
```diff
@@ -175,11 +175,17 @@ export default {
         this.$root.$emit('bv::show::modal', 'create-party-modal');
       }
     },
-    showPartyMembers () {
+    async showPartyMembers () {
+      // If party, retrieve the full party details to allow permissions checks.
+      if (this.user.party._id) {
+        await this.$store.dispatch('party:getParty', true);
+        this.party = this.$store.state.party.data;
```
paglias: is this.party defined only here?

negue: you're right, @titchimoto could you add `party` to the data object?

titchimoto:
@negue @TheHollidayInn Whoops, my bad, I think I accidentally assumed it was already on the data object, added now! :) 

Good point, too, about it loading the party details each time... if I've understood the way it retrieves correctly, I've changed the check to see if there is already a party loaded... if so, totally skip over the `dispatch` and just go straight to opening the modal with the current `party` details. 

Let me know if this works for you if there's anything else obvious I'm missing! 

Cheers :)

website/client/components/header/index.vue
```diff
@@ -175,11 +176,18 @@ export default {
         this.$root.$emit('bv::show::modal', 'create-party-modal');
       }
     },
-    showPartyMembers () {
+    async showPartyMembers () {
```
paglias, line 157: sorry for the late response @titchimoto as you correctly guessed the problem is because `this.user.party ` that was used here isn't the entire party but just some data for it. 

Just two small notes about the fix:

- `await this.$store.dispatch('party:getParty', true);` by default loads the party again only if it's not been loaded already, while with `true` passed as the second parameter it loads the entire party from the server even if the data is already here.
- So calling `await this.$store.dispatch('party:getParty');` returns the party from memory if it's already been loaded, otherwise it loads it from the server
- There's no need to assign the party to this.party since it's only used in this function

I'd refactor the code removing the `party: {}` inside `data ()` and then the function like this

```
      ...
      // Retrieve the party data
      const party = await this.$store.dispatch('party:getParty');

       // Set the party details for the members-modal component
      this.$store.state.memberModalOptions.groupId = party.data._id;
      ...
      this.$store.state.memberModalOptions.group = party.data;
```

This should work properly without loading the party more than it's needed. I'm not sure if the $dispatch returns the party data directly or inside the `data` property, so please double check that

paglias (changes requested):
I've left one comment, let me know if it's not clear or tou have any question!

titchimoto:
@paglias Awesome! Thanks for all the feedback and suggestions, really helpful & makes much more sense now. Just refactored based on your suggestions and is working fine & much cleaner. Hopefully that should do it! But let me know if there's anything else I can do to improve it. :) Thanks!

negue:
@titchimoto could also remove the `Remove Member Option` if its the logged in user? cause nobody can remove itself^^

SabreCat:
@negue I think that might be separately fixed on my https://github.com/HabitRPG/habitica/pull/10704 - in which case, maybe this is ready to go?

negue:
@SabreCat yes, your PR fixes that issue 👍 

@titchimoto can you resolve the conflicts?

paglias:
@titchimoto I'm extremely sorry for the delay here but we've finally managed to merge your PR, thanks!

I've awarded your contributor tier.

PR #10641, opened 2018-08-27
(WIP) Exp notifs

### LINKED ISSUE #10635: Death causes an incorrect notification that you gained Experience
As reported by Il Mago (c073342f-4a65-4a13-9ffd-9e7fa5410d6b) and confirmed by me:

"My character just died and I received a notification that said that I gained experience (when instead I lost them), about 1500xp which was the amount I needed to reach next level before death"

The user's level and experience are adjusted correctly, it's just the notification that's misleading.

This will be very confusing for new users who already often ask what the effects of death are so I'm marking this as high priority.

TheHollidayInn:
Fixes #10635

- this depends on a PR that is going to be merged soon that contains the test base
- I found a few cases we were not correctly displaying for
- From the current context, we can't determine if the user levels up with exact xp or if they died. Both of these cases result in the scenario where after === 0 and before is either 0 or greater.
- The above is b/c we simply set the exp to 0 rather than always accumulated xp and displaying a 0 value

Soooo, see the test for examples, but two ways to fix this:
- Change xp to never go to 0 (on the user object) and display the current xp minus all gained xp 
- Or, change stat notifications to view the whole state delta

We will probably go with the latter solution since it is easier and still somewhat correct

### DIFF
--- a/website/client/components/notifications.vue
+++ b/website/client/components/notifications.vue
@@ -224,14 +224,7 @@ export default {
       if (after < 0) this.playSound('Minus_Habit');
     },
     userExp (after, before) {
-      if (after === before) return;
-      if (this.user.stats.lvl === 0) return;
-
-      let exp = after - before;
-      if (exp < -50) { // recalculate exp if user level up
-        exp = toNextLevel(this.user.stats.lvl - 1) - before + after;
-      }
-      this.exp(exp);
+      this.displayUserExpNotifications(after, before);
     },
     userGp (after, before) {
       if (after === before) return;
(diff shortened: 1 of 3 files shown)


website/client/components/notifications.vue
```diff
@@ -199,6 +199,9 @@ export default {
     userClassSelect () {
       return !this.user.flags.classSelected && this.user.stats.lvl >= 10;
     },
+    userHasClass () {
+      return this.user.stats.lvl >= 10 && this.user.flags.classSelected && !this.user.preferences.disableClasses;
```
paglias, line 203: for this check you should be able to use the hasClass getter in client/store/getters/members

Arsnael:
Hi ! I don't want to bother too much, but I feel somewhat concerned, as my previous fix created this issue^^' So I know this PR is not finished, but I don't think this can solve the issue... I don't see how you can really notice the user got multiple level ups or died... If you check the state of the variable this.user.stats.lvl (like with console.log or anything you want), you will see that it's **always** showing the lvl of the user **after update**. Like if you get multiple levels like from 10 to 13, the value will be 13 (so no way to know from which level you started). If you were lvl 20 and you die, then you will get the value 19. So your logic will always think that user got just 1 lvl up, even on multiple lvl ups or death cases (I tested this manually to confirm).

I did think before as well that maybe should notify the exp from somewhere else, like the updateStats file where the exp is actually being updated with the loop... but that script is used by both the server and the client, so I guessed it's a no go...  

I hope I don't cross the line here though (sorry if I do), I just want to help :) 

EDIT : or could try something like https://stackoverflow.com/questions/42737034/vue-js-watch-multiple-properties-with-single-handler ? No idea if this works, but if you could put the watch of lvl and exp changes in the same handler, that could resolve the issue I guess... What do you think?

TheHollidayInn:
No worries! I did mention above the same thing. There isn't an accurate way to check death versus exact level up. I posted two fixes as well.

paglias:
Related #10297 #10590

PR #10644, opened 2018-08-28
New inbox client

paglias:
Part of #10428

- Moves GET /inbox/messages to v3 from v4
- Add more tests for POST-send_private_message.test.js
- Update tests for DELETE-user_messages in v3
- New v4 route to delete messages
- Fix a bug when overriding routes for v4
- Inbox: use new routes to load messages and delete

WIP this was meant to work with other changes but it's now been split, so there's a bug to fix in the client side view before it can be merged


paglias:
Fixed, had to make sure sendPrivateMessage returned messages with `sent: true`

PR #10648, opened 2018-08-29
Fix new party member cannot join pending quest

### LINKED ISSUE #9293: New Party members can't join pending Quests
After a new user joins a Party, if there is a pending quest that is still in the invitation state, the new Party member cannot join the pending quest. If they choose Accept, they get an error message: "Your party is already on a quest. Try again when the current quest has ended." If the player joins the party while the quest is still in the invitation stage, they SHOULD get an invitation to the quest. They should not if they join the party after the quest has started.

screen shot 2017-10-26 at 6 06 11 pm

Arsnael:

Fixes #9293

### Changes
Well the issue was that for a new member that just joined a party, if a quest is pending , then an error was raised at the allocation of partyData variable. This issue was client side, on the server side, no problem, the user gets added to the quest invitation process when joining a party (https://github.com/HabitRPG/habitica/blob/develop/website/server/controllers/api-v3/groups.js#L552)

It seems store.state.partyMembers.data.length was null => the party ID is set up in store when user joins the party, but not this one (https://github.com/HabitRPG/habitica/blob/develop/website/client/store/actions/guilds.js#L76)

So I check first instead if store.state.party exists, as after joining a guild the players is being redirected to the party page, and I believe this is being fetched like this there : https://github.com/HabitRPG/habitica/blob/develop/website/client/components/groups/group.vue#L457

Sometimes with the new member I get the party buttons (accept / reject) showing up normally, sometimes I need to hit the sync button first to see them (no idea what's the issue there... but seems for a lot o stuff it's like that). But in any case, when I click on 'Accept' it goes through now !

Let me know if there is any concerns here :)

----

### DIFF
--- a/website/client/store/actions/quests.js
+++ b/website/client/store/actions/quests.js
@@ -6,15 +6,16 @@ import * as Analytics from 'client/libs/analytics';
 
 export async function sendAction (store, payload) {
   // @TODO: Maybe move this to server
-  let partyData = {
-    partyID: store.state.user.data.party._id,
-    partySize: store.state.partyMembers.data.length,
-  };
-
+  let partyData = {};
   if (store.state.party && store.state.party.data) {
     partyData = {
       partyID: store.state.party.data._id,
       partySize: store.state.party.data.memberCount,
+    };
+  } else {
+    partyData = {
+      partyID: store.state.user.data.party._id,
+      partySize: store.state.partyMembers.data.length,
     };
   }
 


negue:
I can't reproduce the issue mentioned in #9293 at all.

1. having two members in my party
2. starting a quest, quest starter already accepted, waiting on the second user 
3. inviting a third user
4. invited user joins the party, but no errors

Arsnael:
@negue : you mean can't reproduce issue with the PR or without? Without before you could see that when you click on accept button for the pending quest nothing happened (and looking in the console from the dev toolbar, you would get an exception). However, if you refresh the all page (F5) then I guess the client gets synced correctly with the server so no error anymore

negue:
I tested it without your changes, the invited user could join without any issues, but what I don't see is the question if the user would like to join the quest, this question only appears on a page refresh (F5), but then it just works.

I had the same result using your changes.

Currently on Windows 10, Vivaldi.

@TheHollidayInn could you take a look ?

Arsnael:
Ah yeah on a page refresh (F5) it's fine. I think because you probably sync back with the server then. But try again by just clicking the sync button (the one between money and notifications scrollbar icons?),not refresh... then you should see the quest pending invitation and that's the case the bug happens

negue:
@Arsnael your hint with the sync button helped :), and yes your change fixed it! Thanks!

Arsnael:
Great to hear @negue ! Happy to help ;)

paglias:
thanks @Arsnael ! notated towards your third tier :)

Arsnael:
Thanks @paglias !

PR #10651, opened 2018-08-30
Creating default encryption test to improve coverage

lucasheim:
### Changes
Added a new unit test for default password encryption when using salt. This seemed to be the only branch missing to get 100% coverage on this file.

----
UUID: 535a72cc-f949-4ba6-8570-639c5f763da8

### DIFF
--- a/test/api/unit/libs/password.test.js
+++ b/test/api/unit/libs/password.test.js
@@ -107,6 +107,25 @@ describe('Password Utilities', () => {
       }
     });
 
+    it('defaults to SHA1 encryption if salt is provided', async () => {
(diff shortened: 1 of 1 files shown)


SabreCat:
Thanks for this, @lucasheim, and welcome to the ranks of the Blacksmiths at tier 1! This will go into a staging environment for a few days, then if all is well go live late next week. (Which is sort of academic for a tests PR, but just letting you know of the process for future reference!)

Note that further contributor tiers require increasing amounts of effort from one to the next, so tier 2 will require multiple PRs or a larger PR to attain. But keep helping out and we'll express our gratitude accordingly!

PR #10655, opened 2018-08-31
Saved sort selection into local storage for later use - fixes #10432

### LINKED ISSUE #10432: Sort options in Stable should be persistent

### Description

As reported in the bug guild:

> 𝕹𝖆𝖊𝕿𝖍𝖊𝕹𝖊𝖗𝖉 (8f89c805-334e-48cf-b095-0eba848f5974) 
>This hasn't ever impacted badly so I've never reported it, but it's getting on my nerves. every time I go to the stable, I select "sort by colour" and every time I come back it's reverted to sorting by pet type. 

Sort options should stick within a session at least, and maybe even across sessions.

flhat:
Fixes #10432

### Changes
- save the sort selection into local storage
- read back from local storage when displaying the stable and sorting

----

### DIFF
--- a/website/client/components/inventory/stable/index.vue
+++ b/website/client/components/inventory/stable/index.vue
@@ -399,6 +399,8 @@
   import openedItemRowsMixin from 'client/mixins/openedItemRows';
   import petMixin from 'client/mixins/petMixin';
 
+  import { CONSTANTS, setLocalSetting, getLocalSetting } from 'client/libs/userlocalManager';
+
   // TODO Normalize special pets and mounts
   // import Store from 'client/store';
   // import deepFreeze from 'client/libs/deepFreeze';
@@ -430,8 +432,7 @@
       mousePosition: MouseMoveDirective,
     },
     data () {
-      let savedSelectedSortBy = 'standard';
-      if (localStorage.getItem('selectedSortBy')) savedSelectedSortBy = JSON.parse(localStorage.getItem('selectedSortBy'));
+      const stableSortState = getLocalSetting(CONSTANTS.keyConstants.STABLE_SORT_STATE) || 'standard';
 
       return {
         viewOptions: {},
@@ -439,7 +440,7 @@
         searchText: null,
         searchTextThrottled: '',
         // sort has the translation-keys as values
-        selectedSortBy: savedSelectedSortBy,
+        selectedSortBy: stableSortState,
         sortByItems: [
           'standard',
           'AZ',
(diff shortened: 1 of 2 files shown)


website/client/components/inventory/stable/index.vue
```diff
@@ -461,6 +464,12 @@
         let search = this.searchText.toLowerCase();
         this.searchTextThrottled = search;
       }, 250),
+      selectedSortBy: {
+        handler () {
+          localStorage.setItem('selectedSortBy', JSON.stringify(this.selectedSortBy));
+        },
+        deep: true,
```
paglias: deep here isn't necessary as we're not dealing with an object but just a string

website/client/components/inventory/stable/index.vue
```diff
@@ -461,6 +464,12 @@
         let search = this.searchText.toLowerCase();
         this.searchTextThrottled = search;
       }, 250),
+      selectedSortBy: {
+        handler () {
+          localStorage.setItem('selectedSortBy', JSON.stringify(this.selectedSortBy));
```
paglias: JSON.stringify isn't necessary since we already have a string, it's only used when you want to transform an object into a json string

website/client/components/inventory/stable/index.vue
```diff
@@ -430,13 +430,16 @@
       mousePosition: MouseMoveDirective,
     },
     data () {
+      let savedSelectedSortBy = 'standard';
+      if (localStorage.getItem('selectedSortBy')) savedSelectedSortBy = JSON.parse(localStorage.getItem('selectedSortBy'));
```
paglias: instead of calling getItem twice you could do something like this

```
const savedSelectedSortBy = localStorage.getItem('selectedSortBy') || 'standard';
```

Also I feel like the key where this is saved (selectedSortBy) should be a bit more descriptive, something like `habitica-stable-sort-preference`

paglias (changes requested):
Left two comments about the code, but then I noticed that we already have a module to make it easier to manage this settings, https://github.com/HabitRPG/habitica/blob/develop/website/client/libs/userlocalManager.js, it would be great if you could add this in there like in this file https://github.com/HabitRPG/habitica/blob/develop/website/client/components/tasks/spells.vue#L181

flhat:
@paglias Thanks for the feedback!  I just committed an update

paglias (approved):

paglias:
thanks @BlkStormy ! and welcome to tier 2!

flhat:
@paglias - Awesome, thanks!  :D

PR #10657, opened 2018-09-01
prevent buying market gear if class doesn't match - fixes #10183

### LINKED ISSUE #10183: should not be able to buy equipment from a class that you are not currently playing
When you lose an old Grand Gala equipment piece due to death and then change class and then try to buy the piece back, you get no notification that you're not able to buy it. If you try to buy it, you see a notification saying "You bought [gear name]" (although gold is not subtracted and the item is not actually bought).

The correct behaviour is:
- When you click on the buy button, you should see a message saying that you can't buy that item because you are not the right class. This message must come from the API so that it works regardless of how you try to buy the item (e.g., using a third-party tool or a direct API command).
- The buy button should not be clickable (you might want to implement this after fully implementing and testing the above point).

When you try to buy a non-Gala piece of gear from the wrong class, you CAN buy it. This is completely wrong - you should not be able to. The API and website's UI should both prevent it as described above.

(I suspect that the reason you can't buy the Gala gear is because the pieces I tested on were from a non-Spring gala and so the seasonal limitation was taking effect. If you tried to buy a piece from the current Gala that you'd lost recently due to death and then changed class, you'd probably be able to buy it.)

_Thanks to lulumiller (49e7da73-62a4-43eb-a292-a48beb43c5e5) for reporting this._

flhat:
Fixes #10183

### Changes
- Check to make sure the item class matches the user class in the buyMarketGear functionality

Tested directly with the api with my warrior:
> $ curl 'http://localhost:8080/api/v4/user/buy/armor_rogue_1' -X POST -H 'Origin: http://localhost:8080' -H 'Accept-Encoding: gzip, deflate, br' -H 'Accept-Language: en-US,en;q=0.9,la;q=0.8' -H 'Cookie: logglytrackingsession=*****; connect:*****; connect:*****; language=en_US; amazon-pay-connectedAuth=lwa_general; session-set=true' -H 'Connection: keep-alive' -H 'x-api-user: *****' -H 'x-client: habitica-web' -H 'Content-Length: 0' -H 'Pragma: no-cache' -H 'x-user-timezoneOffset: 420' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36' -H 'Accept: application/json, text/plain, */*' -H 'Cache-Control: no-cache' -H 'Referer: http://localhost:8080/shops/market' -H 'DNT: 1' -H 'x-api-key: *****' --compressed
>> {"success":false,"error":"NotAuthorized","message":"You can't buy this item."}

Without this update, the call goes through.

----

### DIFF
--- a/website/common/script/ops/buy/buyMarketGear.js
+++ b/website/common/script/ops/buy/buyMarketGear.js
@@ -24,16 +24,16 @@ export class BuyMarketGearOperation extends AbstractGoldItemOperation {
     return false;
   }
 
+  canUserPurchase (user, item) {
+    return super.canUserPurchase(user, item) && (item.klass !== 'special' && item.klass !== user.stats.class);
+  }
+
   extractAndValidateParams (user, req) {
     let key = this.key = get(req, 'params.key');
     if (!key) throw new BadRequest(errorMessage('missingKeyParam'));
 
     let item = content.gear.flat[key];
     if (!item) throw new NotFound(errorMessage('itemNotFound', {key}));
-
-    if (item.klass && item.klass !== 'special' && item.klass !== user.stats.class) {
-      throw new NotAuthorized(this.i18n('cannotBuyItem'));
-    }
 
     this.canUserPurchase(user, item);
 
--- a/website/common/script/ops/buy/buyMarketGear.js
+++ b/website/common/script/ops/buy/buyMarketGear.js
@@ -31,6 +31,10 @@ export class BuyMarketGearOperation extends AbstractGoldItemOperation {
     let item = content.gear.flat[key];
     if (!item) throw new NotFound(errorMessage('itemNotFound', {key}));
 
+    if (item.klass && user.stats.class !== item.klass) {
+      throw new NotAuthorized(this.i18n('cannotBuyItem'));
+    }
+
     this.canUserPurchase(user, item);
 
     if (user.items.gear.owned[item.key]) {
--- a/website/common/script/ops/buy/buyMarketGear.js
+++ b/website/common/script/ops/buy/buyMarketGear.js
@@ -31,7 +31,7 @@ export class BuyMarketGearOperation extends AbstractGoldItemOperation {
     let item = content.gear.flat[key];
     if (!item) throw new NotFound(errorMessage('itemNotFound', {key}));
 
-    if (item.klass && user.stats.class !== item.klass) {
+    if (item.klass && item.klass !== 'special' && item.klass !== user.stats.class) {
       throw new NotAuthorized(this.i18n('cannotBuyItem'));
     }
 


website/common/script/ops/buy/buyMarketGear.js
```diff
@@ -31,6 +31,10 @@ export class BuyMarketGearOperation extends AbstractGoldItemOperation {
     let item = content.gear.flat[key];
     if (!item) throw new NotFound(errorMessage('itemNotFound', {key}));
 
+    if (item.klass && item.klass !== 'special' && item.klass !== user.stats.class) {
+      throw new NotAuthorized(this.i18n('cannotBuyItem'));
+    }
+
     this.canUserPurchase(user, item);
```
paglias: shouldn't the check go into this function?

website/common/script/ops/buy/buyMarketGear.js
```diff
@@ -31,6 +31,10 @@ export class BuyMarketGearOperation extends AbstractGoldItemOperation {
     let item = content.gear.flat[key];
     if (!item) throw new NotFound(errorMessage('itemNotFound', {key}));
 
+    if (item.klass && item.klass !== 'special' && item.klass !== user.stats.class) {
```
paglias: no need to check for item.klass at the beginning of the if

paglias (changes requested):

website/common/script/ops/buy/buyMarketGear.js
flhat: which check function?  canUserPurchase() ?

paglias: Yes, canUserPurchase

negue: Oh I guess I forgot this condition :( , but @BlkStormy thanks for fixing it :)

`canUserPurchase` is from `AbstractGoldItemOperation` so a check for `item.klass` isn't something that needs to be checked for every gold-purchasable-item.

You could either override `canUserPurchase` (with the call to `super.canUserPurchase`) in this class and add the check there - OR - extend the checks for `item.canOwn` cuase this already checked inside `canUserPurchase` https://github.com/HabitRPG/habitica/blob/develop/website/common/script/ops/buy/abstractBuyOperation.js#L143

flhat: @negue I just added another commit - can you take a look?  Is this what you meant?

website/common/script/ops/buy/buyMarketGear.js
```diff
@@ -24,6 +24,10 @@ export class BuyMarketGearOperation extends AbstractGoldItemOperation {
     return false;
   }
 
+  canUserPurchase (user, item) {
+    return super.canUserPurchase(user, item) && (item.klass !== 'special' && item.klass !== user.stats.class);
```
negue, line 28: `canUserPurchase` doesn't return a bool, I guess the name is misleading, sorry about that!

`super.canUserPurchase` throws errors in the conditions if the user can't purchase it, so your prior code was right, you only needed to move it here under the the `super.canUserPurchase`-call

paglias:
Any update on this one?

paglias:
@negue can you take over this one?

negue:
will do 👍

PR #10658, opened 2018-09-01
Allow admins to edit challenge tasks - fixes #8149

### LINKED ISSUE #8149: moderators can't edit challenge tasks

flhat:
Fixes #8149

### Changes
- Rather than checking if the user is the owner of the challenge, check canModify(user) instead

----

### DIFF
--- a/test/api/v3/integration/tasks/challenges/POST-tasks_challenge_id.test.js
+++ b/test/api/v3/integration/tasks/challenges/POST-tasks_challenge_id.test.js
@@ -54,6 +54,21 @@ describe('POST /tasks/challenge/:challengeId', () => {
     expect(tasksOrder.habits).to.include(task.id);
   });
 
+  it('allows non-leader admin to add tasks to a challenge when not a member', async () => {
+    const admin = await generateUser({'contributor.admin': true});
+    let task = await admin.post(`/tasks/challenge/${challenge._id}`, {
+      text: 'test habit from admin',
+      type: 'habit',
+      up: false,
+      down: true,
+      notes: 1976,
+    });
+
+    let {tasksOrder} =  await user.get(`/challenges/${challenge._id}`);
+
+    expect(tasksOrder.habits).to.include(task.id);
+  });
+
   it('returns error when user tries to create task with a alias', async () => {
     await expect(user.post(`/tasks/challenge/${challenge._id}`, {
       text: 'test habit',
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -247,7 +247,7 @@ api.createChallengeTasks = {
 
     // If the challenge does not exist, or if it exists but user is not the leader -> throw error
     if (!challenge) throw new NotFound(res.t('challengeNotFound'));
-    if (challenge.leader !== user._id) throw new NotAuthorized(res.t('onlyChalLeaderEditTasks'));
+    if (!challenge.canModify(user)) throw new NotAuthorized(res.t('onlyChalLeaderEditTasks'));
 
     let tasks = await createTasks(req, res, {user, challenge});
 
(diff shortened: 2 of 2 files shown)


paglias:
@BlkStormy the code is looking good, can you add a test for this functionality? One test per route will be enough. Let me know if you need help about writing tests

flhat:
@paglias Thanks - yes, can you help me figure out how to write a test for this?  I'm new to writing tests for this codebase.  I read the instructions on the wiki, but perhaps I could see an example to start with?

paglias:
@BlkStormy sure, taking `api.createChallengeTasks` as an example, but it's more or less the same with the other routes, the integration tests for that route can be found here https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/tasks/challenges/POST-tasks_challenge_id.test.js

as you can see there's a test https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/tasks/challenges/POST-tasks_challenge_id.test.js#L42 to check that a leader can add tasks even if not a member, you want to add a new test that checks the same thing but for an user with admin properties. So you should create a new user with the properties `contributor.admin` set to true and use that user to check for the same thing:

Something like this

```
  it('allows admin to add tasks to a challenge when not a member', async () => {
    const admin = await generateUser({'contributor.admin': true});
    let task = await admin.post(`/tasks/challenge/${challenge._id}`, {
      text: 'test habit',
      type: 'habit',
      up: false,
      down: true,
      notes: 1976,
    });

    let {tasksOrder} =  await admin.get(`/challenges/${challenge._id}`);

    expect(tasksOrder.habits).to.include(task.id);
  });
```

paglias:
Any update on this one?

Alys:
I've written a test for this and I'm pushing it to the moderators-edit-challenge-tasks branch on BlkStormy/habitica to update this PR, but the push is running very slowly. I'm going to leave it to do its thing and go to bed. I'll check on it tomorrow. If it turns out I've messed up the PR, I'll fix it tomorrow or on the weekend. 
@BlkStormy Git's version control means that your work can't be lost so don't worry if things seem to go a bit wonky for now. :)

Alys:
Right, so that's every commit ever added to this PR. :roll_eyes: Sorry. I shouldn't have made a new branch based on `develop` and merged in this PR. Let me see if I can fix this. If I can't, I'll make a new PR with @BlkStormy's edits and my test, and I'll link it back to this one so @BlkStormy's credit is clear.

Alys:
That's all fixed and pretty now. :) Tests are passing, including the new one. It's ready for review again.

I have not tested this manually on a local install because I'm guessing it was tested manually during the previous review, but if it wasn't, let me know and I'll test it.

paglias:
merged! thanks @Alys and @BlkStormy ! @BlkStormy this has been added to your list of contributions

PR #10661, opened 2018-09-02
Fix initial position item info when selecting one item after another (fixes #10077)

### LINKED ISSUE #10077: inventory selected item should not be offset when selecting one item after another
_The bug report below was posted by LeonZaLion (d4a926a7-9b69-4f50-aa0c-48c5d7afdb56) in the Report a Bug guild. I've confirmed it on Firefox on Linux._

In your inventory under items, when selecting on item and having it in your hand, and selecting another item of the same type (e.g. grabbing a base potion and selecting a desert potion without deselecting the base potion) will cause the image of the newly selected item's picture and description to pop up where the original item's image and description should appear until your cursor moves. Try selecting a potion, and then selecting another one but don't move your cursor after you've clicked it. Notice that the image & description of the new potion appears where the old potion's image and description should be, then move your mouse and it goes back to following your cursor.

I'm using Chrome...

I've got a link of a screen recording of the glitch: https://drive.google.com/open?id=1UJrcPyeM09dsnvo66gAnrhtX-SJBqXec

lucubro:
Fixes #10077.

### Changes
While dragging the potion/egg, the variable `lastMouseMoveEvent`, which was keeping track of the mouse position, was not updated so if a second item was selected while dragging another the variable was still pointing at the position of the first click.

This changes the code to update `lastMouseMoveEvent` even while dragging another object. I've also added a couple of comments to help people that read that code for the first time. I hope that's welcomed.

----
UUID: bd2b5eb2-04d4-4604-a972-e19ae09de30d

### DIFF
--- a/website/client/components/inventory/items/index.vue
+++ b/website/client/components/inventory/items/index.vue
@@ -405,6 +405,8 @@ export default {
         this.currentDraggingEgg = egg;
         this.eggClickMode = true;
 
+        // Wait for the div.eggInfo.mouse node to be added to the DOM before
+        // changing its position.
         this.$nextTick(() => {
           this.mouseMoved(lastMouseMoveEvent);
         });
@@ -427,6 +429,8 @@ export default {
         this.currentDraggingPotion = potion;
         this.potionClickMode = true;
 
+        // Wait for the div.hatchingPotionInfo.mouse node to be added to the
+        // DOM before changing its position.
         this.$nextTick(() => {
           this.mouseMoved(lastMouseMoveEvent);
         });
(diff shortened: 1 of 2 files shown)


lucubro:
Whoops! Sorry for arriving too late. I've noticed that there was the same exact bug when dragging food in the stable, so I fixed that too. If you prefer to fix that in a separate PR, I can either revert the last commit or close this one and open another clean PR.

paglias:
that's fine @lucubro, thanks! It'll get merged in the next few days

paglias:
thanks @lucubro ! notated towards your second tier!

PR #10663, opened 2018-09-05
Guild textarea at list positioning

### LINKED ISSUE #10612: Username @ list in chat is positioned incorrectly from multiline messages

### Description

If you include carriage returns in a chat message, then attempt to @ mention a chat user, the dropdown with name selections appears further down than it ought. The more lines in your message, the more pronounced the issue:

image

rkojima:

Fixes #10612

### Changes
Username @ list chat is now positioned so that if at the end of the text box, the @ list will line up right below the text box. 
Didn't make the change where the text box height increases as you add more text.

**Before:**
screen shot 2018-09-05 at 12 46 24 pm

**After:**
screen shot 2018-09-05 at 12 45 32 pm

Also added minimum height to text box.

----
UUID: 70c6bc6f-3edc-40e1-b9bc-0b7289484f42

### DIFF
--- a/website/client/components/chat/autoComplete.vue
+++ b/website/client/components/chat/autoComplete.vue
@@ -14,7 +14,7 @@ div.autocomplete-selection(v-if='searchResults.length > 0', :style='autocomplete
 import groupBy from 'lodash/groupBy';
 
 export default {
-  props: ['selections', 'text', 'coords', 'chat'],
+  props: ['selections', 'text', 'coords', 'chat', 'textbox'],
   data () {
     return {
       currentSearch: '',
@@ -25,12 +25,12 @@ export default {
   },
   computed: {
     autocompleteStyle () {
-      function heightToUse (topCoords) {
-        let textAreaHeight = document.getElementsByClassName('user-entry')[0].clientHeight;
-        return topCoords < textAreaHeight ? topCoords + 30 : textAreaHeight + 10;
+      function heightToUse (textBox, topCoords) {
+        let textBoxHeight = textBox['user-entry'].clientHeight;
+        return topCoords < textBoxHeight ? topCoords + 30 : textBoxHeight + 10;
       }
       return {
-        top: `${heightToUse(this.coords.TOP)}px`,
+        top: `${heightToUse(this.textbox, this.coords.TOP)}px`,
         left: `${this.coords.LEFT + 30}px`,
         marginLeft: '-28px',
         marginTop: '28px',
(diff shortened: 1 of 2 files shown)


website/client/components/chat/autoComplete.vue
```diff
@@ -25,9 +25,15 @@ export default {
   },
   computed: {
     autocompleteStyle () {
+      function heightToUse (topCoords) {
+        let textAreaHeight = document.getElementsByClassName('user-entry')[0].clientHeight;
```
negue: I think `document.getElementsByClassName('user-entry')` could be passed as a prop (using `$refs` instead of going through the DOM)

website/client/components/groups/chat.vue
```diff
@@ -187,7 +187,7 @@
     position: relative;
 
     textarea {
-      height: 150px;
+      min-height: 150px;
```
negue, line 193: I checked `min-height` and `height` it doesn't seem to change anything? I can still resize the textarea

negue (changes requested):
Looks good 👍 could you check my comments and fix those too?

website/client/components/groups/chat.vue
rkojima: I did that so that it can't resize to a height < 150px. Does adding this line of code not do that?

rkojima:
@negue I will definitely need some checks since I don't know if I added in the props correctly.

website/client/components/groups/chat.vue
```diff
@@ -62,6 +64,7 @@
           TOP: 0,
           LEFT: 0,
         },
+        textbox: this.$refs,
```
negue, line 67: I think it would be better to use `this.$refs['user-entry']` that way `autocomplete` wouldn't need to use a specific named textarea

website/client/components/groups/chat.vue
negue: on my first try it doesn't seemed to use the `min-height`, retesting it again, worked, maybe some browser hiccup

negue (changes requested):
All changes work, now only one little change left 👍

website/client/components/groups/chat.vue
rkojima: Hmm, when I set `textbox` to `this.$refs['user-entry']`, `this.textbox` on line 33 of autoComplete.vue is set to undefined, but if it's just `this.$refs`, then `this.textbox` is set to an Object with a `user-entry` property, which has a value of the element and its properties. So I think setting it to `this.$refs` is better.

paglias:
@rkojima sorry for the delay here, we've finally managed to merge your PR, thanks! I've notated this PR towards your second tier

PR #10671, opened 2018-09-10
WEB [Nov 6] Added username invite

TheHollidayInn:
To do:
- [x] Frontend changes for inviting to Parties / Guilds (modals)
- [x] If a user types in @phillipthelen to invite someone instead of phillipthelen The server should strip that @, so that it still works.


phillipthelen:
TODO: If a user types in `@phillipthelen` to invite someone instead of `phillipthelen` The server should strip that `@`, so that it still works.

saraolson:
@phillipthelen is this PR tied to the ability to start a message with username on the mobile apps?

SabreCat:
Merged into #10784

PR #10672, opened 2018-09-11
Css font home page update

### LINKED ISSUE #9650: Update CSS fonts on Habitica homepage
Certain sections of the Habitica homepage are using the incorrect font. Should be using [Varela Round](https://fonts.google.com/specimen/Varela+Round) from Google Fonts.

**Existing**
image

**Designed**
image

If you are interested in working on this issue, please comment an email address to be invited to our Zeplin project.

Arsnael:

Fixes #9650

### Changes
Update of the font used on the home page from Roboto to Valera Round according to the zeplin design.

----

### DIFF
--- a/website/client/components/static/header.vue
+++ b/website/client/components/static/header.vue
@@ -35,11 +35,18 @@
 <style lang='scss' scoped>
   @import '~client/assets/scss/colors.scss';
 
+  @import url('https://fonts.googleapis.com/css?family=Varela+Round');
+
   @media only screen and (max-width : 750px) {
     .login-button {
       margin: 0 auto !important;
       margin-top: 1.8em !important;
     }
+  }
+
+  nav, .btn {
+      font-family: 'Varela Round', sans-serif;
+      font-weight: normal;
   }
 
   .btn-primary.pull-right {
@@ -66,9 +73,8 @@
 
   .nav-item {
     .nav-link {
-      font-size: 16px;
+      font-size: 16px !important;
       color: $white;
-      font-weight: bold;
       line-height: 1.5;
       padding: 16px 24px;
       transition: none;
--- a/website/client/components/static/header.vue
+++ b/website/client/components/static/header.vue
@@ -35,8 +35,6 @@
 <style lang='scss' scoped>
   @import '~client/assets/scss/colors.scss';
 
-  @import url('https://fonts.googleapis.com/css?family=Varela+Round');
-
   @media only screen and (max-width : 750px) {
     .login-button {
       margin: 0 auto !important;
@@ -44,9 +42,9 @@
     }
   }
 
-  nav, .btn {
-      font-family: 'Varela Round', sans-serif;
-      font-weight: normal;
+  .home-header, .home-header .btn {
+    font-family: 'Varela Round', sans-serif;
+    font-weight: normal;
   }
 
   .btn-primary.pull-right {
--- a/website/client/components/static/home.vue
+++ b/website/client/components/static/home.vue
@@ -125,6 +125,8 @@
 <style lang="scss" scoped>
 @import '~client/assets/scss/colors.scss';
 
+@import url('https://fonts.googleapis.com/css?family=Varela+Round');
+
   #front {
     .form-text a {
       color: #fff !important;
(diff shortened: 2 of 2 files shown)


negue:
LGTM 👍

paglias:
Sorry about this but I got a few concerns here:

- Is it necessary to @import the font twice?
- We should find a way to set the different font-family only once and have it applied for all elements otherwise we'll end up forgetting it when a new element is added
- I want to be sure the new font is only loaded on the static pages and not even if you only use the app

Arsnael:
Don't be sorry @paglias , no problem, it's valid concerns actually :)

To answer your questions : 

- regarding the import twice, I didn't like it either, but the new font on the mockup is being used in the home page, but also the header of that home page, which is in a different file. But I can try to recheck and think again if cannot do better there.
- Applying it on all elements is I believe not really possible at the moment, more in the way that in the zeplin mockup, not all the text is using Valera Round. A lot of text is still using Roboto as well if you look carefully. Is it it a mistake in the mockup or did I misunderstand your question?
- Logically the @import for Valera Round is only on the scoped css for the header and the home static pages, they shouldn't be loaded anywhere else normally (if I understood correctly how scoped css worked in Vue.js)

Arsnael:
@paglias : so I tried to dig a bit more and review a bit all of that, based on your concerns. This is what I could resume : 
- if you look at the zeplin design, for example there : 
habit_home_2
You can see that still Robot and Valera are being used (Blue boxes => Valera, red boxes Roboto). I can try to maybe change the code to unify it more between those 2 fonts, so that we reduce the risks of forgetting something in the future, but you will always have that risk I think, as long as you have different fonts cohabiting with each other.
- It's scoped css, so normally shouldn't propagate outside of those static pages. But seems I can just declare the import on the home static page, the static header seems to catch up on it. It's not loading in the app for sure, I think the scoped css is making sure of it, and I checked as well with the dev toolbar if the Valera font was loaded in the app, and it's not. So should be safe on that part.
- However I went a bit fast on the header part... Seems the static header is shared between the home page and other static pages that you have when you get logged in (contact us, How it works, etc). 
habit_how
That was probably not required by the task. But if I remove the @import there like I said in the previous point, might help doing the trick.

To resume : let me work a bit more on this :)

paglias:
@Tressley is the screenshot above with both fonts ok?

Tressley:
That's true that not all text is the same. Mainly the headlines are using Varela Round and the body is Roboto. And yeah, I think that static case is okay for now. We still need to address those page designs.

Arsnael:
Hm ok... Well let me know if you need me to change something or not at the end then !

Arsnael:
@Tressley , @paglias : so need to change something or not at the end? :)

website/client/components/static/header.vue
```diff
@@ -35,13 +35,20 @@
...
+
   @media only screen and (max-width : 750px) {
     .login-button {
       margin: 0 auto !important;
       margin-top: 1.8em !important;
     }
   }
 
+  nav, .btn {
+      font-family: 'Varela Round', sans-serif;
```
paglias: the indentation here should be 2 spaces

website/client/components/static/header.vue
```diff
@@ -35,13 +35,20 @@
 <style lang='scss' scoped>
   @import '~client/assets/scss/colors.scss';
 
+  @import url('https://fonts.googleapis.com/css?family=Varela+Round');
```
paglias: In a previous comment you said

> It's scoped css, so normally shouldn't propagate outside of those static pages. But seems I can just declare the import on the home static page, the static header seems to catch up on it. It's not loading in the app for sure, I think the scoped css is making sure of it, and I checked as well with the dev toolbar if the Valera font was loaded in the app, and it's not. So should be safe on that part.

so maybe this second import can be removed?

paglias (commented):
sorry for the delay, left two comments

website/client/components/static/header.vue
Arsnael: Oups my bad. Need to fix it indeed

website/client/components/static/header.vue
Arsnael: Ok let me check that!

Arsnael:
@paglias : changes made, please review !

Arsnael:
Hi ! Any updates on this? :)

paglias:
Sorry for the delay @Arsnael and thanks again for the PR! It's been notated towards your next tier

Arsnael:
No problem @paglias . Was just reminding to be sure was not forgotten, you all seem quite busy with the new username feature recently :) Thanks !

PR #10676, opened 2018-09-12
Show accurate experience notifications

### LINKED ISSUE #10297: show accurate Experience (XP) gain in notification on level-up

Arsnael:

Fixes #10297

### Changes
I managed to put a watch on a concatenated value of experience and level. From there, we get the change of exp and the change of levels, which allows us to do pretty much anything regarding showing correct notifications. This time I took the following cases into consideration and believe show the correct exp notification for : 
- user gains exp but no level up
- user gains exp and levels up once (in both cases where user had already more xp than necessary or not)
- user gains exp and multi-levels up (in both cases where user had already more xp than necessary or not)
- user dies (showing exp loss).

I added the proper unit test cases (greatly inspired by @TheHollidayInn from #10641) for all of those above.  

I hope that this time it will be alright :)

----

### DIFF
--- a/website/client/components/notifications.vue
+++ b/website/client/components/notifications.vue
@@ -204,7 +204,7 @@ export default {
       return this.user.party.quest.RSVPNeeded && !this.user.party.quest.completed;
     },
     userExpAndLvl () {
-      return `${this.user.stats.exp}|${this.user.stats.lvl}`;
+      return [this.user.stats.exp, this.user.stats.lvl];
     },
   },
   watch: {
@@ -273,10 +273,7 @@ export default {
       this.$root.$emit('bv::show::modal', 'quest-invitation');
     },
     userExpAndLvl (after, before) {
-      const [beforeExp, beforeLvl] = before.split('|');
-      const [afterExp, afterLvl] = after.split('|');
-
-      this.displayUserExpAndLvlNotifications(afterExp, beforeExp, afterLvl, beforeLvl);
+      this.displayUserExpAndLvlNotifications(after[0], before[0], after[1], before[1]);
     },
   },
   mounted () {
--- a/website/client/components/notifications.vue
+++ b/website/client/components/notifications.vue
@@ -88,6 +88,7 @@ import axios from 'axios';
 import moment from 'moment';
 import throttle from 'lodash/throttle';
 
+import { toNextLevel } from '../../common/script/statHelpers';
 import { shouldDo } from '../../common/script/cron';
 import { mapState } from 'client/libs/store';
 import notifications from 'client/mixins/notifications';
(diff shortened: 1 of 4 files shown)


website/client/components/notifications.vue
```diff
@@ -204,6 +203,9 @@ export default {
     invitedToQuest () {
       return this.user.party.quest.RSVPNeeded && !this.user.party.quest.completed;
     },
+    userExpAndLvl () {
+      return `${this.user.stats.exp}|${this.user.stats.lvl}`;
```
negue: could you check if its possible to use an array of those values instead of an concated string?

negue (commented):
No more negative exp notifications 🎉

Could you check my comment and see if that works

Arsnael:
Well @negue except when you die, then you get loss of exp notification :)

I will check tomorrow regarding your comment. I think I read the way with an array was possible too when I was searching the solution.

Arsnael:
@negue Made the requested change, please review :)

Arsnael:
Hi ! Any updates on this? :)

majestic-owl448:
Would this still work for when you loose experience without death or loosing a level? (For example when simply unticking a task)

Arsnael:
@Ieahleen : I didn't test but that's a good question. I would say normally yes, but that case would deserve to be covered by the test file too. I will check it when I have time. Anything else?

negue:
@Ieahleen good point, just testet it with having 2exp - unchecked a task , got to -5 exp, but it didn't de-lvl

@paglias @Alys I couldn't find any code which would de-lvl a character once it gets under zero exp, so I guess this is a bug? Or should we just prevent the exp from going under zero?

paglias:
The user is deleveled in https://github.com/HabitRPG/habitica/blob/develop/website/common/script/fns/updateStats.js

paglias:
Which is called by scoreTask and a few other functions

Arsnael:
@negue , @paglias , @Ieahleen : alright I will take a look at it and add test cases for it then, thanks !

EDIT : by a quick look to updateStats file it doesn't look like to me that the user is deleveled in the code... but i will look it more into details later !

Alys:
It's intentional that when your Experience drops below zero, the player is not de-leveled. There are some comments about that at https://github.com/HabitRPG/habitrpg/issues/5736#issuecomment-127677025 and below but in brief, we believe it's simpler and less bug-prone to allow negative XP after unticking a task than to reverse a level-up.

Arsnael:
I see so it 's wanted, thanks for the link @Alys ! I will still add a test case for unchecking a task and loosing xp, but the rest should be alright

Arsnael:
So to resume : as stated by @Alys , the negative xp is a wanted choice. So it works then, we get the negative notification when loosing xp, and I added those cases as unit tests too.

Let me know if there is other matters :)

paglias:
thanks @Arsnael ! and welcom to the 3rd tier!

Arsnael:
\o/ Thank you !

SabreCat:
Man, this is so smart. Going live shortly!

PR #10691, opened 2018-09-18
Animal ears after death

### LINKED ISSUE #10668: Animal Ears should not be purchasable with gems after losing them
When you lose Animal Ears due to Death (or the future Empty the Armory) feature, they appear in the Market to be bought back with gold (which is the correct behaviour and shouldn't be changed).

However they also appear in the User Icon > Avatar section to be bought back for Gems. This is not correct - once a player has bought them with Gems, they should never need to spend Gems on them again.

_This old issue is related so I'm linking it here in case anyone's interested in the discussion there: [Pet Ears cannot be obtained for free after Death or Orb of Rebirth](https://github.com/HabitRPG/habitica/issues/5236)_

_Thanks to Il Mago for reporting this._

kirsty-tortoise:

Fixes #10668

### Changes
* Can now purchase ears/tails for gold (and not gems) from the User / Edit Avatar screen if you have bought them before but lost them to death.
 * Unlocking all ears/tails now removes them from pinned items. 

----

### DIFF
--- a/website/client/components/creatorIntro.vue
+++ b/website/client/components/creatorIntro.vue
@@ -591,20 +591,21 @@ b-modal#avatar-modal(title="", :size='editing ? "lg" : "md"', :hide-header='true
     }
   }
 
-  .text-center .gem-lock {
-    display: inline-block;
-    margin-right: 1em;
-    margin-bottom: 1.6em;
-    vertical-align: bottom;
+  .text-center {
+    .gem-lock, .gold-lock {
+      display: inline-block;
+      margin-right: 1em;
+      margin-bottom: 1.6em;
+      vertical-align: bottom;
+    }
   }
 
-  .gem-lock {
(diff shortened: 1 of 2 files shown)


kirsty-tortoise:
Purchasing with gold doesn't currently have a confirm dialog which needs changing but then also they maybe all should be changed to the confirm modals anyway.

Alys:
Currently it's intentional that gold-purchasable items don't have confirmation dialogs (or modals) because buying with gold is a safe activity and confirmations can be annoying (especially if you're rebuying much gear after the upcoming Empty the Armoiry feature!). If you do buy something accidentally and wanted the gold for something else, you can use Fix Character Values to give it back to yourself.  Gems purchases have confirmations because that involves real-world money.

paglias:
@kirsty-tortoise  thanks for this! I'm merging the PR and awarding your contributor tier.

Issue #10695, opened 2018-09-19
Fix display issues in mobile browsers


phillipthelen:
List of fixes, so that we can link users to a working settings page

- [x] Make banner for paused dailies display correctly
- [x] Display settings pages in one column instead of two.

majestic-owl448:
Other things that could be considered:

often the modals are a nightmare to close (once I was blocked there trying to tap anywhere on the margins to close one until I figured out that as they have the little cross in the corner on the website I could use the same corner to close them - it was a profile modal and there it is not visible)

Scrolling through tasks, you need to use the margins of the screen or you will end moving the tasks around

phillipthelen:
The modals are a good point, since these can pop up on every page. However regarding the tasks the focus right now is on the settings page, since we will send users to that page from an email. We want to work on improving the mobile website soon though!

PR #10700, opened 2018-09-20
Fix website issues on mobile devices

phillipthelen:
This is not completely done yet, but addresses the issues in #10695 

The banner for paused dailies now wraps to multiple lines if necessary and has a shorter version on smartphones, so that it doesn't take up 3+ rows there. The content is correctly positioned below the banner.
The settings page now also has only one column on phones, since the layout is unusable otherwise.

### DIFF
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -17,8 +17,9 @@ div
       template(v-if="isUserLoaded")
         div.resting-banner(v-if="showRestingBanner")
           span.content
-            span.label {{ $t('innCheckOutBanner') }}
-            span.separator |
+            span.label(v-if="!isNarrowScreen") {{ $t('innCheckOutBanner') }}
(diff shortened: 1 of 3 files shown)


website/client/app.vue
```diff
@@ -17,8 +17,9 @@ div
       template(v-if="isUserLoaded")
         div.resting-banner(v-if="showRestingBanner")
           span.content
-            span.label {{ $t('innCheckOutBanner') }}
-            span.separator |
+            span.label(v-if="!isNarrowScreen") {{ $t('innCheckOutBanner') }}
+            span.label(v-if="isNarrowScreen") {{ $t('innCheckOutBannerShort') }}
```
paglias: in this case you can use `v-else` instead of another `v-if` with the opposite condition, https://vuejs.org/v2/api/#v-else

phillipthelen:
Got the banner offset working correctly now, so this should be ready once it is reviewed.

Issue #10696, opened 2018-09-19
Backend changes for Username changes


phillipthelen:
- [x] Add `verifiedUsername` flag to user
- [x] Add verification to usernames by blocking slurs using the [bannesSlurs](https://github.com/HabitRPG/habitica/blob/90117625d72e71ca94af4915886b10b4952d0747/website/server/libs/bannedSlurs.js) list
- [x] Change username API to not require a password
- [x] URL to confirm username without changing it. This should open the settings page. This will be used for the emails we send out
- [x] Limit display names to 30 characters

PR #10703, opened 2018-09-21
Username backend changes

phillipthelen:
This implements the things mentioned in #10696 

Usernames are validated by:

- Length (1 to 20 characters
- only containing valid characters (letters, numbers, _, -)
- Not containing slurs
- being available


Alys:
I think the `bannedSlurRegexs` in its current form will block legitimate words that happen to have a sequence of letters that would be a slur if they were in a word by themselves (the [Scunthorpe problem](https://en.wikipedia.org/wiki/Scunthorpe_problem) - content warning: that page contains some examples of words that are not permitted in Habitica's public spaces).
Slurs (and also swear words from website/server/libs/bannedWords.js) should be prevented only if there's a word boundary around them. For example, these are bad:
- `TESTPLACEHOLDERSWEARWORDHERE`
- `_TESTPLACEHOLDERSWEARWORDHERE_`
- `TESTPLACEHOLDERSWEARWORDHERE_OTHERWORD`
- `TESTPLACEHOLDERSWEARWORDHERE-OTHERWORD`

but these are acceptable:
- `NOTATESTPLACEHOLDERSWEARWORDHERE`
- `TESTPLACEHOLDERSWEARWORDHERETHISISOKAY`

phillipthelen:
I used this regex, since I figured that the words in the slur file wouldn't really have an acceptable use within words. But there might be some, so not sure if it's better to be a bit more restrictive or not.

Alys:
It's definitely better to allow some deliberate slurs to slip through (mods will handle them as we do now) than to block legitimate words that happen to contain a slur in a syllable. We want swear words to be prevented too and that increases the number of normal words that would be accidentally blocked. Non-English names will be allowed of course, again increasing the chance of accidental blocks.

phillipthelen:
After discussing this again, we have decided to keep the check the way it is now and not add a check for the swear words, since that would be too restrictive considering some of the words on the list. Out of all the users that logged in in the past 3 months, there were zero legitimate uses of any of the words on the slurs list. Additionally to that most of the offending usernames (most of which are very far from acceptable) are written without any separators between the slur and the rest of the username. So the only way to have catch then with the check is by keeping it the way it is now. Considering there are currently no active users (and potentially also no inactive) users that would be accidentally affected by this, I don't think the the regex should be changed.

website/server/libs/user/index.js
```diff
@@ -239,4 +242,38 @@ export async function rebirth (req, res, { isV3 = false }) {
   await Promise.all(toSave);
 
   res.respond(200, ...rebirthRes);
-}
\ No newline at end of file
+}
+
+const bannedSlurRegexs = bannedSlurs.map((word) => new RegExp(`.*${word}.*`, 'i'));
+
+function usernameContainsSlur (username) {
```
paglias: Sorry about this but this index.js file is already big enough, can you move the functions to a different file? like libs/user/username?

PR #10704, opened 2018-09-21
Group Management Menu Fixes

SabreCat:
### Changes

* Rearranges and renames the options in the member action dropdown ("..." button in party/group) to the following as hashed out with @Tressley 
  * **Send Message**
  * **Transfer Ownership**
  * **Assign/Unassign Manager**
  * **Remove Member**
* Makes the above show and hide more intelligently:
  * Don't show Assign Manager if target user is already a manager or party leader
  * Don't show Unassign Manager if target user is not a manager or active user is not the leader
  * Don't show Remove Member if target user is the group leader
* Removes a few unused EN strings related to this functionality


negue:
I have a few issues:
- if the leader assigns a manager the menu doesn't show "remove manager" only after the refreshing the page, but isn't that big of deal
- if a manager wants to remove another manager I get the error "that only the leader can remove managers"

Tressley:
@negue -- Group Leaders are the only Group member with administration rights to add/remove Group Managers.

SabreCat:
@Tressley: Yeah, I think the implied suggestion is that "Unassign Manager" should not show up if the user is only a manager, not a leader!

Tressley:
@SabreCat -- Ah, of course! My apologies @negue!

SabreCat:
OK, got around to fixing that trivial thing finally, hah

website/client/components/groups/membersModal.vue
```diff
@@ -33,29 +33,29 @@ div
...
+            b-dropdown-item(@click='removeManager(member._id)', v-if='shouldShowRemoveManager(member._id)')
               span.dropdown-icon-item
                 .svg-icon.inline(v-html="icons.removeIcon")
                 span.text {{$t('removeManager2')}}
             b-dropdown-item(@click='viewProgress(member)', v-if='challengeId')
               span.dropdown-icon-item
                 span.text {{ $t('viewProgress') }}
+            b-dropdown-item(@click='removeMember(member, index)', v-if='shouldShowLeaderFunctions(member._id)')
+              span.dropdown-icon-item
+                .svg-icon.inline(v-html="icons.removeIcon", v-if='isLeader')
```
paglias: I think the `v-if='isLeader'` here can be removed

paglias (commented):
left one small comments but otherwise it's good to go

PR #10713, opened 2018-09-24
Decrease mana when removing stat points from int

### LINKED ISSUE #10712: Mana Points (MP) not decreased when removing Intelligence (INT) Stat Points
This is a client-side display problem, not a problem with the underlying API call for allocating Stat Points.

When you allocate Stat Points to Intelligence, you're given 1 MP for each point and your total mana shown in the website is increased immediately, before you have saved your Stat Points allocation. If you change your mind and remove Stat Points from INT before saving, your Mana shown on the website is not decreased. Therefore after you do save, it looks like you have more Mana than you really do. When you sync, your displayed Mana becomes correct. 
Your actual Mana as recorded in your account in Habitica's database is only updated when you save your changes to your Stat Points so it's always correct.

I.e., there needs to be a client-side change so that when you remove a Stat Point allocated to Intelligence, your Mana shown on the website is immediately decreased, OR a sync is immediately forced to ensure that the Mana you're seeing is correct.

Steps to reproduce bug:
1. Have at least one Stat Point available to be allocated (e.g., level up).
1. Click the website's sync button to ensure you're seeing the latest data then record your Mana.
1. Allocate the Stat Point to INT (do NOT click "Save"). Observe that you see a notification for gaining 1 MP.
1. Remove the Stat Point from INT. Observer that you do NOT see a notification for losing 1 MP (ideally you should although that's not critical).
1. Click "Save".
1. Observer your Mana shown in your stat bars in the website's header. It will be 1 greater than what it was before (it should be the same as it was before). This is the part that most needs to be fixed.
1. Sync. Observe that your Mana is now correct (the same as it was before).

kirsty-tortoise:

Fixes #10712

### Changes
If you deallocate a point from int, your mana is decremented (it is incremented when you allocate a point) on the client side, (and a notification appears, this happened automatically).

----

### DIFF
--- a/website/common/script/ops/stats/allocate.js
+++ b/website/common/script/ops/stats/allocate.js
@@ -24,9 +24,6 @@ module.exports = function allocate (user, req = {}) {
   if (user.stats.points > 0) {
     user.stats[stat]++;
     user.stats.points--;
-    if (stat === 'int') {
-      user.stats.mp++;
-    }
   } else {
     throw new NotAuthorized(i18n.t('notEnoughAttrPoints', req.language));
   }
--- a/website/common/script/ops/stats/allocate.js
+++ b/website/common/script/ops/stats/allocate.js
@@ -12,6 +12,7 @@ import hasClass from '../../libs/hasClass';
 
 module.exports = function allocate (user, req = {}) {
   let stat = get(req, 'query.stat', 'str');
+  let ignoreManaPoints = get(req, 'query.ignoreManaPoints', false);
 
   if (ATTRIBUTES.indexOf(stat) === -1) {
     throw new BadRequest(errorMessage('invalidAttribute', {attr: stat}));
@@ -24,6 +25,9 @@ module.exports = function allocate (user, req = {}) {
   if (user.stats.points > 0) {
     user.stats[stat]++;
     user.stats.points--;
+    if (!ignoreManaPoints && stat === 'int') {
+      user.stats.mp++;
+    }
   } else {
     throw new NotAuthorized(i18n.t('notEnoughAttrPoints', req.language));
   }
--- a/website/common/script/ops/stats/allocate.js
+++ b/website/common/script/ops/stats/allocate.js
@@ -12,7 +12,6 @@ import hasClass from '../../libs/hasClass';
 
 module.exports = function allocate (user, req = {}) {
   let stat = get(req, 'query.stat', 'str');
-  let ignoreManaPoints = get(req, 'query.ignoreManaPoints', false);
 
   if (ATTRIBUTES.indexOf(stat) === -1) {
     throw new BadRequest(errorMessage('invalidAttribute', {attr: stat}));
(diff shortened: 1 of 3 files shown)


kirsty-tortoise:
Made changes as mentioned in #10712 so that the mana points and notifications are only added after saving.

Also made changes so you cannot deallocate points you have already saved in the user interface, as this isn't possible so was not being sent to the server and the user and the server were ending up out of sync.

kirsty-tortoise:
Apparently I broke the thing in shared/common, so I'll check where it is used on the server side and probably add an option to add mana or not to.

website/common/script/ops/stats/allocate.js
```diff
@@ -24,9 +24,6 @@ module.exports = function allocate (user, req = {}) {
   if (user.stats.points > 0) {
     user.stats[stat]++;
     user.stats.points--;
-    if (stat === 'int') {
```
paglias: this is going to affect the server side code as well, so I don't think that it can be removed

kirsty-tortoise: Yep, got it. Plan is to add a "ignoreManaPoints" parameter to the request?

paglias: Actually we've been discussing this internally and it's possible that this feature (dealloation) isn't meant to exist unless in some specific cases so I'd wait a bit to see what @Tressley and @SabreCat think

kirsty-tortoise: This isn't deallocation, it is allocation?

kirsty-tortoise:
I've pushed some changes so that the behaviour on the client side is what I wanted and the server side is unchanged. I'm leaving it for now.

website/common/script/ops/stats/allocate.js
paglias: You're right, so the problem now is that the server is unchanged but anyone can pass the `ignoreManaPoints` parameter to the API and have this change affect the server side as well which we don't want to allow

kirsty-tortoise: Would giving it as a third parameter `options` to `allocate` which is always set by the server and the client be better? I considered this first, but no other functions in common had `options`, so I didn't think it was necessary. Alternatively we can validate the request before passing it to allocate.

I don't feel like it is a huge issue, since I can't see any reason why someone would want to avoid adding 1 mana point to themselves by manually sending a request, but I see it is possibly not ideal.

paglias:
@SabreCat did we reach a decision on this one?

kirsty-tortoise:
I've asked on the issue about whether it would be better to leave all allocation on the client side until save is pressed, so that the client and server don't get out of sync if the user edits stat points without saving and then tries to do other stuff.

kirsty-tortoise:
Okay, I think I've finished rewriting this for now, I'm waiting for some feedback on the issue and depending on that I might revert the last commit, but otherwise it is ready for review I think.

negue:
ping @paglias @SabreCat 

as for the changes, they work as described, all changes just at client-side until you save it
Code LGTM 👍

paglias:
@kirsty-tortoise sorry for all the delay here but I finally managed to merge your 3 PRs, thanks!

I've also awarded your second tier and your 3rd one is in progress :)

PR #10726, opened 2018-09-28
Username Frontend Changes

SabreCat:
[//]: # (Note: See http://habitica.wikia.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API for more info)

[//]: # (Put Issue # here, if applicable. This will automatically close the issue if your PR is merged in)
Fixes #10694

### Changes
[//]: # (Describe the changes that were made in detail here. Include pictures if necessary)
Makes modifications to the User Settings page in preparation for username changes. Requires #10703 to work without errors.

* Removes the password field associated with changing the Username (fka Login Name), as this can now be done by social auth users who lack a password.
* Removes the summary section showing user's current username and email, instead defaulting those into the appropriate input fields.
* Adds fields and functionality to change the user's Display Name (fka Profile Name).
* Tweaks various strings to use new terminology.

[//]: # (Put User ID in here - found on the Habitica website at User Icon > Settings > API)

PR #10729, opened 2018-10-01
Username announcement

phillipthelen:
This PR contains changes from both #10703 and #10726. Since they depend on eachother it was easier to do the finishing touches with both PRs combined.

This PR contains all changes required for the announcement phase.


website/client/components/auth/authForm.vue
```diff
@@ -106,7 +112,50 @@ export default {
...
+      this.validateUsername(this.username);
+    },
+  },
   methods: {
+    // eslint-disable-next-line func-names
+    validateUsername: debounce(function (username) {
+      if (username.length <= 3) {
+        return;
+      }
+      this.$store.dispatch('auth:verifyUsername', {
```
paglias, line 149: Is the `this` here pointing to the right context?

website/client/components/header/notificationsDropdown.vue
```diff
@@ -179,6 +182,15 @@ export default {
         });
       }
 
+      if (this.user.flags.verifiedUsername !== true) {
```
paglias, line 185: so the notification is being pushed in the client side, not on the server? So it's never saved? Also it won't be created again and again every time this function is called?

website/client/components/static/home.vue
```diff
@@ -600,6 +615,14 @@
         if (this.email.length <= 3) return false;
         return !this.validateEmail(this.email);
       },
+      usernameValid () {
```
paglias, line 618: is there a way to share all these validations in a mixin or something? Because they're duplicated in 4 places, same for the validateUsername fn. Mixins are computed properties/methods, ... that can be shared across components https://vuejs.org/v2/guide/mixins.html

website/client/components/settings/site.vue
```diff
@@ -233,19 +249,33 @@ export default {
         password: '',
         confirmPassword: '',
       },
+      usernameIssues: [],
```
paglias, line 256: I think this array should be reset if the user leaves the page without saving

paglias (changes requested):

website/client/components/header/notificationsDropdown.vue
phillipthelen: The notification is shown as long is the flag isn't true. So as long as the user hasn't verified it, it's being shown (It should still always only be there once. Hasn't shown twice for me during my testing). Once the username is verified, the notification isn't shown anymore.

website/client/components/settings/site.vue
phillipthelen: Won't the data object be recreated in that case? I was under that assumption. But I can also reset it in `destroyed` (I guess that would be the right place?)

Tressley:
When an input has a value, it should default to the :active font-color #4e4a57.

image

website/client/components/settings/site.vue
paglias: you're right it's being recreated automatically (just tested on the settings page)

PR #10732, opened 2018-10-02
Fixing typo on portuguese due date

wladimirgramacho:
### Changes
It fixes a typo regarding Portuguese due date field on tasks page. Quick and easy PR.

----
UUID: fa777e46-d2bf-4b48-a253-7804c377690e

### DIFF
--- a/website/common/locales/pt/tasks.json
+++ b/website/common/locales/pt/tasks.json
@@ -79,7 +79,7 @@
     "complete2": "Feito",
     "dated": "Com data",
     "today": "Hoje",
-    "dueIn": "Prazo<%= dueIn %>",
+    "dueIn": "Prazo <%= dueIn %>",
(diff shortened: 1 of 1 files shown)


majestic-owl448:
I am not an admin, but this seems something you would want to fix on transifex, where Habitica is translated, you can join the translation team there if you so wish: https://www.transifex.com/habitrpg/habitrpg/dashboard/

wladimirgramacho:
I would agree with you if it was a translation, but I'm just adding a space needed for the phrase. The phrase needs a space to make sense.

example

Alys:
@wladimirgramacho Thank you for wanting to improve the translations! @Ieahleen is correct that this needs to be done in Transifex. The non-en files can't be changed through a pull request. Please see the _README_FIRST.md file in the pt directory for more details.

PR #10734, opened 2018-10-03
Issue 10728: sort equipment by stat descending on Market page

### LINKED ISSUE #10728: Discrepancy in sorting between Market and Equipment
Per Il Mago again:

> In the Market if I sort by an attribute I have lower attribute first (https://snag.gy/zMOPoW.jpg), in the Equipment section it is instead higher attribute first

trvrfrd:

Fixes #10728 

### Changes
I changed the `sortGearItems()` computed property of the `equipmentSection.vue` component of the Market screen to reverse the order of the items when sorting by a stat. This will display the highest-stat items first, the same as on the Equipment section of the Inventory. (The alternative would be to sort in ascending order on both pages, which seems to make less sense)

While I was making changes, I noticed that sorting by PER was not actually implemented. I couldn't find an open issue related to this, but I went ahead and included the fix.

I also took the liberty of moving the filtering part of `sortGearItems()` into its own method, since `sortGearItems()` was getting fairly long and complicated, and the filtering logic seems to stand on its own. I'm not super familiar with Vue, though, so maybe it makes more sense as a computed property than a method?

I'm open to feedback on all of these changes!

Also, I didn't see any obvious place to add tests for this, so I didn't, which is probably bad :eyes: 

----
UUID: 7620c72a-bb5a-4798-b3c0-846dacb2b44d

### DIFF
--- a/website/client/components/shops/market/equipmentSection.vue
+++ b/website/client/components/shops/market/equipmentSection.vue
@@ -63,6 +63,7 @@ layout-section(:title="$t('equipment')")
 
   import _filter from 'lodash/filter';
   import _sortBy from 'lodash/sortBy';
+  import _reverse from 'lodash/reverse';
   import pinUtils from '../../../mixins/pinUtils';
 
   const sortGearTypes = ['sortByType', 'sortByPrice', 'sortByCon', 'sortByPer', 'sortByStr', 'sortByInt'].map(g => ({id: g}));
@@ -137,9 +138,17 @@ layout-section(:title="$t('equipment')")
           return !this.userItems.gear.owned[gear.key];
         });
 
-        // first all unlocked
-        // then the selected sort
-        result = _sortBy(result, [(item) => item.locked, sortGearTypeMap[this.selectedSortGearBy.id]]);
+        let selectedSortKey = sortGearTypeMap[this.selectedSortGearBy.id];
+
+        result = _sortBy(result, selectedSortKey);
+
+        if (selectedSortKey !== 'type' && selectedSortKey !== 'value') {
+          // sorting by a stat, need to reverse to get high-to-low sort
+          result = _reverse(result);
+        }
+
+        // put unlocked items first
+        result = _sortBy(result, (item) => item.locked);
 
         return result;
       },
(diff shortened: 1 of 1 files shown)


website/client/components/shops/market/equipmentSection.vue
```diff
@@ -116,9 +118,38 @@ layout-section(:title="$t('equipment')")
         return this.marketGearCategories.filter(c => c.id === this.selectedGroupGearByClass)[0];
       },
       sortedGearItems () {
+        let result = this.filterGearItems();
+        let selectedSortKey = sortGearTypeMap[this.selectedSortGearBy.id];
+
+        result = _sortBy(result, selectedSortKey);
```
paglias: hi @trvrfrd , sorry for the delay here. The code looks good, I'm just wondering if it's possible to condense these 3 sorting functions (2 sortBy and the reverse) into just one since with many items it would be a very expensive operation.

For example you could save the call to reverse() by using orderBy instead of sortBy and change to `desc` as the third parameter in case `selectedSortKey !== 'type' && selectedSortKey !== 'value' === true`

Something like this should work:

```
let order = selectedSortKey !== 'type' && selectedSortKey !== 'value' ? 'desc' : 'asc';
result = _orderBy(result, [selectedSortKey, locked], ['', 'asc'];
```

paglias (commented):
left one comment, overall it looks good!

website/client/components/shops/market/equipmentSection.vue
trvrfrd: Now I'm sorry for the delay on my end! I've been dealing with something of a family emergency all week, I'll take a look at your feedback and incorporate it ASAP. Thanks @paglias!

paglias: no worries @trvrfrd ! hope everything's all right

trvrfrd:
hi @paglias, implemented the change you suggested. let me know if all is good!

paglias:
it's working well, thanks @trvrfrd ! I've awarded your second tier :)

trvrfrd:
awesome! thanks for your help @paglias :grin:

PR #10738, opened 2018-10-05
add verified Usernames to inbox and chat messages.

phillipthelen:
This adds a users username to inbox and chat messages as long as they have a verified username. This will eventually be needed for displaying the username on messages and also for @messaging. It should be added before we release those features, so that on release there are already messages with usernames which improves how those features work at launch.

### DIFF
--- a/test/api/v3/integration/chat/POST-chat.test.js
+++ b/test/api/v3/integration/chat/POST-chat.test.js
@@ -448,6 +448,22 @@ describe('POST /chat', () => {
     expect(messageBackerInfo.tokensApplied).to.equal(backerInfo.tokensApplied);
   });
 
+  it('adds username if user has verified', async () => {
+    const newMessage = await user.post(`/groups/${groupWithChat._id}/chat`, { message: testMessage});
+
+    expect(newMessage.message.username).to.equal(user.auth.local.username);
+  });
+
+  it('does not add username if user has not verified', async () => {
+    const unverifiedUser = await generateUser({
+      'flags.verifiedUsername': false,
+    });
+    await unverifiedUser.sync();
+    const newMessage = await unverifiedUser.post(`/groups/${groupWithChat._id}/chat`, { message: testMessage});
+
+    expect(newMessage.message.username).to.not.exist;
+  });
+
   it('sends group chat received webhooks', async () => {
     let userUuid = generateUUID();
     let memberUuid = generateUUID();
--- a/test/api/v3/integration/members/POST-send_private_message.test.js
+++ b/test/api/v3/integration/members/POST-send_private_message.test.js
@@ -151,13 +151,16 @@ describe('POST /members/send-private-message', () => {
     let sendersMessageInSendersInbox = _.find(updatedSender.inbox.messages, (message) => {
       return message.uuid === receiver._id && message.text === messageToSend;
     });
-    expect(response.message.username).to.equal(userToSendMessage.auth.local.username);
+    expect(response.message.username).to.equal(receiver.auth.local.username);
     expect(sendersMessageInReceiversInbox.username).to.equal(userToSendMessage.auth.local.username);
-    expect(sendersMessageInSendersInbox.username).to.equal(userToSendMessage.auth.local.username);
+    expect(sendersMessageInSendersInbox.username).to.equal(receiver.auth.local.username);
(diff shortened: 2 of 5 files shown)


website/server/models/user/methods.js
```diff
@@ -148,18 +151,21 @@ schema.methods.sendMessage = async function sendMessage (userToReceiveMessage, o
...
+
   if (!sendingToYourself) {
     newSenderMessage = new Inbox({
       sent: true,
       ownerId: sender._id,
     });
     Object.assign(newSenderMessage, messageDefaults(senderMsg, userToReceiveMessage));
     setUserStyles(newSenderMessage, sender);
+    if (sender.flags.verifiedUsername === true) {
+      newSenderMessage.username = sender.auth.local.username;
```
Alys: If I understand this code correctly....

When you send a PM to another player, this block of code puts a copy of your PM into your own inbox. That copy will contain the recipient's profile.name and User ID (not your profile.name).
Doesn't that mean that the username should be the recipient's username, not the sender's?

If that's true, I wonder if lines 163 to 165 (and lines 120 to 122) could be put into the messageDefaults() function. That would reduce a bit of code duplication and ensure that the same user's profile.name and username were always displayed together.

phillipthelen: I misunderstood how the user data is stored (Not sure why, since I've worked with that object in the past and wondered why it was modeled that way, so I should've known). I changed it and also moved the assignment to the `messageDefaults` function as suggested.

test/api/v3/integration/chat/POST-chat.test.js
```diff
@@ -448,6 +448,22 @@ describe('POST /chat', () => {
...
+    const newMessage = await user.post(`/groups/${groupWithChat._id}/chat`, { message: testMessage});
+
+    expect(newMessage.message.username).to.equal(user.auth.local.username);
+  });
+
+  it('does not add username if user has not verified', async () => {
+    const unverifiedUser = await generateUser({
+      'flags.verifiedUsername': false,
+    });
+    await unverifiedUser.sync();
```
paglias, line 461: sync here shouldn't be necessary

test/api/v3/integration/members/POST-send_private_message.test.js
```diff
@@ -133,6 +133,59 @@ describe('POST /members/send-private-message', () => {
...
+
+  it('does not add username if user has not verified', async () => {
+    let receiver = await generateUser({
+      'flags.verifiedUsername': false,
+    });
+    receiver.sync();
+    const unverifiedUser = await generateUser({
+      'flags.verifiedUsername': false,
+    });
+    await unverifiedUser.sync();
```
paglias, line 167: sync here is not necessary I think

test/api/v3/integration/members/POST-send_private_message.test.js
```diff
@@ -133,6 +133,59 @@ describe('POST /members/send-private-message', () => {
...
+    expect(response.message.username).to.equal(receiver.auth.local.username);
+    expect(sendersMessageInReceiversInbox.username).to.equal(userToSendMessage.auth.local.username);
+    expect(sendersMessageInSendersInbox.username).to.equal(receiver.auth.local.username);
+  });
+
+  it('does not add username if user has not verified', async () => {
+    let receiver = await generateUser({
+      'flags.verifiedUsername': false,
+    });
+    receiver.sync();
```
paglias, line 163: sync here is not necessary I think

website/server/controllers/api-v3/chat.js
```diff
@@ -177,6 +177,11 @@ api.postChat = {
     }
 
     const newChatMessage = group.sendChat(req.body.message, user);
+
+    if (user.flags && user.flags.verifiedUsername === true) {
```
paglias, line 181: is this necessary? Because the same block of code in `messageDeafult` should be used when sending chat as well so i don't think this is necesssary

paglias (changes requested):
a few minor things

SabreCat:
I think I tried to do the same thing here in my PR: https://github.com/HabitRPG/habitica/pull/10764/files#diff-388968ce2a61bca656313c066ddbda7a Should I remove that so we don't conflict?

PR #10739, opened 2018-10-05
Implement Bailey CMS API

phillipthelen:
This brings a new admin-only view that allows admins to create, edit and delete bailey posts.

Changes/News features:

- Create and change baileys without needing github
- /api/v4/news returns posts as JSON
- no need to run a migration for `flags.newStuff`, since the server now stores `flags.lastNewStuffRead`, which contains the unique ID of the last post the user read. When creating or updating baileys, a cached value is updated that has the ID of the most recent post. If the users last read ID differs from the correct ID, the `flags.newStuff` property is set to true in the user object.

The PR still needs more extensive tests (particularly the handling of the newStuff needs tests), but I'm opening it now for feedback on the general approach.

P.S.: I'm not sure why it's listing 95 commits as being part of this PR.

UPDATE: I removed the admin interface, since it will be available elsewhere.

TODO 
- [x] do not import default from uuid
- [x] id -> _id
- [x] author vs credits?
- [x] published vs publishedData?
- [x] tellMeLaterNews identical between api v3 and v4?
- [x] tellMeLaterNews: title all upper case?
- [x] modal: date filter should respect user preferences
- [x] author should be private? any other private field?
- [x] sort order for returned posts
- [x] how many posts to show in modal? the last one + any draft if it exists
- [x] modal design: update border radius to match other modals
- [x] update / create news post: required fields / fields that cannot be updated
- [x] check that static news page works: needed at all? Need to split modal and posts
- [x] use res.respond instead of res.send?
- [x] api docs
- [x] breaking change: impossible to mark news as read in v3?
- [x] parse markdown in api v3 response
- [x] tests
- [x] Database indexes
- [x] website/common/locales/en/news.json : anything needed for weblate?
- [x] page with all posts?


paglias:
Should this be put back to `in review`?

news_cache/en.json
```diff
@@ -0,0 +1 @@
...ver. You can find them in User > Avatar > Extra. Enjoy!"},{"title":"August Subscriber Set Revealed!","authors":"Beffymaroo","date":"2018-08-23T00:00:00.000Z","text":"Promo mystery 201808Subscriber Items for August have been revealed: the Lava Dragon Item Set! You only have until August 31 to receive the item set when you subscribe. If you're already an active subscriber, reload the site and then head to Inventory > Items to claim your gear!\nSubscribers also receive the ability to buy Gems for Gold -- the longer you subscribe, the more Gems you can buy per month! There are other perks as well, such as longer access to uncompressed data and a cute Jackalope pet. Best of all, subscriptions let us keep Habitica running. Thank you very much for your support -- it means a lot to us.\n\n# Blog Post: Creating a Unique Experience\nScene casting spellsThis month's featured Wiki article is about using Habitica's features to create a unique experience! We hope that it will help you as you customize Habitica to make the app even more motivating and fun. Be sure to check it out, and let us know what you think by reaching out on Twitter, Tumblr, and Facebook.\nby shanaqui and the Wiki Wizards"}]
```
paglias: Is this file used somewhere?

website/server/models/newsPost.js
```diff
@@ -0,0 +1,73 @@
...
+  if (!isAdmin) {
+    posts = this.find({
+      published: true,
+      publishDate: { $lte: Date() },
+    })
+      .select('title publishDate credits text');
+  } else {
+    posts = this.find();
+  }
+  return posts.sort({ publishDate: -1 });
```
paglias: I think `exec` is missing, without it you don't get a full promise
```suggestion
  return posts.sort({ publishDate: -1 }).exec();
```

website/server/models/newsPost.js
```diff
@@ -0,0 +1,73 @@
...
+schema.statics.getNews = async function getNews (isAdmin) {
+  let posts = [];
+  if (!isAdmin) {
+    posts = this.find({
+      published: true,
+      publishDate: { $lte: Date() },
+    })
+      .select('title publishDate credits text');
+  } else {
+    posts = this.find();
```
paglias: I think we should add some pagination here. We have a lot of Baileys posts from the past

website/server/models/newsPost.js
```diff
@@ -0,0 +1,73 @@
...
+  noSet: ['_id'],
+  timestamps: true,
+});
+
+schema.statics.getNews = async function getNews (isAdmin) {
+  let posts = [];
+  if (!isAdmin) {
+    posts = this.find({
+      published: true,
+      publishDate: { $lte: Date() },
```
paglias: is this necessary? I mean we're filtering for things published and not published in the future?

If we keep it, it should be `new Date()`

website/server/models/newsPost.js
```diff
@@ -0,0 +1,73 @@
...
+  return posts.sort({ publishDate: -1 });
+};
+
+let cachedLastNewsPostID = null;
+let cachedLastNewsPostDate = null;
+let cachedLastNewsPostTitle = null;
+
+schema.statics.lastNewsPostID = async function lastNewsPostID () {
+  if (cachedLastNewsPostID === null) {
+    const lastPost = (await this.getNews(false))[0];
```
paglias: can't we use a `findOne` query here? Otherwise we fetch all the posts but only use one

website/server/models/newsPost.js
```diff
@@ -0,0 +1,73 @@
...
+    if (lastPost !== undefined && lastPost !== null) {
```
paglias:
```suggestion
    if (lastPost) {
```

website/server/models/newsPost.js
```diff
@@ -0,0 +1,73 @@
...
+      cachedLastNewsPostID = lastPost.id;
+      cachedLastNewsPostDate = lastPost.publishDate;
+    }
+  }
+  return cachedLastNewsPostID;
+};
+
+schema.statics.lastNewsPostTitle = async function lastNewsPostTitle () {
+  if (cachedLastNewsPostTitle === null) {
+    const lastPost = (await this.getNews(false))[0];
```
paglias: same thing here about the query, also we're doing a new query just to fetch the title? Couldn't we combine the various functions in one?

website/server/libs/user/index.js
```diff
@@ -27,6 +27,7 @@ export async function get (req, res, { isV3 = false }) {
     userToJSON.needsCron = false;
     if (daysMissed > 0) userToJSON.needsCron = true;
     User.addComputedStatsToJSONObj(userToJSON.stats, userToJSON);
+    userToJSON.flags.newStuff = await user.checkNewStuff();
```
paglias: so `newStuff` has been removed from the user model but we still want to return it to the user? Only in this route?

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,207 @@
...
+
+const api = {};
+
+/**
+ * @apiDefine postIdRequired
+ * @apiError (400) {BadRequest} postIdRequired A postId is required
+ */
+
+/**
+ * @api {get} /api/v3/news Get latest Bailey announcement
```
paglias: says `v3`

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,207 @@
...
+    let isAdmin = false;
+    if (user && user.contributor) {
+      isAdmin = user.contributor.admin;
+    }
+    const results = await NewsPost.getNews(isAdmin);
+    res.respond(200, results);
+  },
+};
+
+api.createNews = {
```
paglias, line 70: missing apidocs

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,207 @@
...
+      published: req.body.published,
+      credits: req.body.credits,
+      text: req.body.text,
+    };
+
+    const newsPost = new NewsPost(postData);
+    await newsPost.save();
+
+    if (newsPost.published) {
+      NewsPost.updateLastNewsPost(newsPost);
```
paglias: this is an async function

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,207 @@
...
+    req.checkParams('postId', apiError('postIdRequired')).notEmpty();
+    const { user } = res.locals;
+    let isAdmin = false;
+    if (user && user.contributor) {
+      isAdmin = user.contributor.admin;
+    }
+
+    const newsPost = await NewsPost.findById(req.params.postId);
+    if (!isAdmin && !newsPost.isPublished) {
+      res.respond(401, {});
```
paglias: should this be a 404?

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,207 @@
...
+  noLanguage: true,
+  async handler (req, res) {
+    req.checkParams('postId', apiError('postIdRequired')).notEmpty();
+    const { user } = res.locals;
+    let isAdmin = false;
+    if (user && user.contributor) {
+      isAdmin = user.contributor.admin;
+    }
+
+    const newsPost = await NewsPost.findById(req.params.postId);
```
paglias: what if the post isn't found?

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,207 @@
...
+    res.respond(200, results);
+  },
+};
+
+api.createNews = {
+  method: 'POST',
+  url: '/news',
+  middlewares: [authWithHeaders(), ensureAdmin],
+  async handler (req, res) {
+    const validationErrors = req.validationErrors();
```
paglias: we're not validating anything so these lines can be removed

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,207 @@
...
+
+    const postData = {
+      title: req.body.title,
+      publishDate: req.body.publishDate,
+      published: req.body.published,
+      credits: req.body.credits,
+      text: req.body.text,
+    };
+
+    const newsPost = new NewsPost(postData);
```
paglias: should call NewsPost.sanitize on `postData`

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,207 @@
...
+
+    user.flags.lastNewStuffRead = await NewsPost.lastNewsPostID();
+
+    await user.save();
+    res.respond(200, {});
+  },
+};
+
+/**
+ * @api {post} /api/v3/news/tell-me-later Get latest Bailey announcement in a second moment
```
paglias: v3

test/api/unit/models/user.test.js
```diff
@@ -59,7 +59,7 @@ describe('User Model', () => {
     expect(userToJSON.stats.toNextLevel).to.not.exist;
     expect(userToJSON.id).to.not.exist;
 
-    User.transformJSONUser(userToJSON);
+    await User.transformJSONUser(userToJSON);
```
paglias: has this become an async function?

paglias (changes requested):
left some comments

website/server/models/newsPost.js
phillipthelen: This allows us to create posts in advance, that automatically go live on the day. Probably not 100% necessary, but imho a nice feature.

website/server/libs/user/index.js
phillipthelen: We still want it on the user for backwards compatability. Since clients are checking for that field to determine if the user has unread news posts.

paglias: makes sense! In this case I'd keep a commented out version of the old field in the schema to remember where it's coming from. Like `// newStuff: { $type: Boolean, default: false },` with a small explanation. 

Also this way it'll only be shown from this route which could be a problem, at the same time it's an async function and we don't want to really call it every time the user is returned

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,250 @@
...
+    req.checkParams('postId', apiError('postIdRequired')).notEmpty();
+    const { user } = res.locals;
+    let isAdmin = false;
+    if (user && user.contributor) {
+      isAdmin = user.contributor.admin;
+    }
+
+    const newsPost = await NewsPost.findById(req.params.postId);
+    if (!newsPost || (!isAdmin && !newsPost.isPublished)) {
+      res.respond(404, {});
```
paglias: this should be a `throw new NotFound(...)`

paglias (commented):
Hmm the tests haven't run for the last few commits, not sure why

website/server/libs/user/index.js
phillipthelen: Added that.
Technically this is an async function, but it should really only run async code for the first user after the server is started. After that the date of the last post is cached and retrieved without running any async code. Not sure if there is a better way to put this call, so that it is always returned when appropriate.

website/client/src/components/achievements/newStuff.vue
```diff
@@ -36,33 +51,110 @@
...
 <script>
-import axios from 'axios';
+import moment from 'moment';
+import habiticaMarkdown from 'habitica-markdown';
 import { mapState } from '@/libs/store';
 
 export default {
+  filters: {
+    date (value) {
+      // @TODO: Vue doesn't support this so we cant user preference
```
paglias: what's that vue doesn't support?

website/client/src/components/achievements/newStuff.vue
```diff
@@ -36,33 +51,110 @@
...
     ...mapState({ user: 'user.data' }),
   },
   async mounted () {
-    this.$root.$on('bv::show::modal', async modalId => {
-      if (modalId !== 'new-stuff') return;
-      const response = await axios.get('/api/v4/news');
-      this.html = response.data.html;
-    });
+    this.posts = (await this.$store.dispatch('user:getNews')).data.data;
+    if (this.posts.length > 2) {
```
paglias: we want to show just the two most recent posts?

website/server/controllers/api-v3/news.js
```diff
@@ -68,11 +67,14 @@ api.tellMeLaterNews = {
...
     const { user } = res.locals;
 
-    user.flags.newStuff = false;
+    user.flags.lastNewStuffRead = await NewsPost.lastNewsPostID();
+    const title = await NewsPost.lastNewsPostTitle();
 
-    const existingNotificationIndex = user.notifications.findIndex(n => n && n.type === 'NEW_STUFF');
-    if (existingNotificationIndex !== -1) user.notifications.splice(existingNotificationIndex, 1);
-    user.addNotification('NEW_STUFF', { title: LAST_ANNOUNCEMENT_TITLE }, true); // seen by default
+    if (user.notifications) {
```
paglias: is this if necessary? In that case `user.addNotification` should be there as well?

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,254 @@
...
+
+const api = {};
+
+/**
+ * @apiDefine postIdRequired
+ * @apiError (400) {BadRequest} postIdRequired A postId is required
+ */
+
+/**
+ * @api {get} /api/v4/news Get latest Bailey announcement
```
paglias: we should add something about the parameter to paginate the request here

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,254 @@
...
+ */
+api.deleteNews = {
+  method: 'DELETE',
+  url: '/news/:postId',
+  middlewares: [authWithHeaders(), ensureAdmin],
+  async handler (req, res) {
+    const validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    const newsPost = await NewsPost.findById(req.params.postId);
```
paglias: .exec() ?

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,254 @@
...
+  noLanguage: true,
+  async handler (req, res) {
+    req.checkParams('postId', apiError('postIdRequired')).notEmpty();
+    const { user } = res.locals;
+    let isAdmin = false;
+    if (user && user.contributor) {
+      isAdmin = user.contributor.admin;
+    }
+
+    const newsPost = await NewsPost.findById(req.params.postId);
```
paglias: .exec() ?

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,254 @@
...
+api.updateNews = {
+  method: 'PUT',
+  url: '/news/:postId',
+  middlewares: [authWithHeaders(), ensureAdmin],
+  async handler (req, res) {
+    req.checkParams('postId', apiError('postIdRequired')).notEmpty();
+    const validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    const newsPost = await NewsPost.findById(req.params.postId);
```
paglias: .exec()?

website/server/libs/user/index.js
paglias: Hmm yeah the problem is that even if it's doesn't make any asyn request it being a promise causes it to be handled as one, and there's the problem that this way only this route will have the updated value, not all the others where the user is returned. I think we'll have to find a different solution here

website/server/models/newsPost.js
```diff
@@ -0,0 +1,90 @@
...
+    .limit(POSTS_PER_PAGE)
+    .skip(POSTS_PER_PAGE * Number(page))
+    .exec();
+};
+
+let cachedLastNewsPostID = null;
+let cachedLastNewsPostDate = null;
+let cachedLastNewsPostTitle = null;
+
+schema.statics.lastNewsPostID = async function lastNewsPostID () {
```
paglias: I think this one and `lastNewsPostTitle` should be merged into a single function since in both cases both the `id` and `title` are fetched

paglias:
For the caching issue I'd say we can either use redis, mongodb change streams (needs to upgrade to 4.0) or poll for updates with a timer

website/client/src/components/achievements/newStuff.vue
phillipthelen: In this screen, probably yes. Since it's supposed to only show the recent ones. (Right now it also only shows the last one)

website/server/models/newsPost.js
phillipthelen: I changed the logic for this a bit

phillipthelen:
I implemented an approach where the cached post is re-checked after 5 minutes (although this could also be set higher). This way after 5 minutes, all users will have the notification available that there is a new post. We can also look into a redis solution or update mongodb, but I felt like those are probably bigger projects

website/server/models/newsPost.js
```diff
@@ -0,0 +1,96 @@
...
+schema.plugin(baseModel, {
+  noSet: ['_id'],
+  timestamps: true,
+});
+
+schema.statics.getLastPost = async function getLastPost () {
+  const post = await this.findOne({
+    published: true,
+    publishDate: { $lte: new Date() },
+  }).sort({ publishDate: -1 }).exec();
```
paglias: Reminder for myself to consider adding an index

website/server/models/newsPost.js
paglias: I'd say this is a good solution, just two things:

- I'd stil merge `lastNewsPostID()` and `lastNewsPostTitle()` into a single function that gets and cache the last post
- When comparing dates I'd use moment's diff, it's a bit cleaner and `new Date()` are objects not numbers so it works because it gets casted to a number but it's a bit confusing, something like `moment().diff(timeStampCachedLastNews, 'minutes') < 5`

website/server/libs/user/index.js
paglias: So I think this should be moved to `toJSONTransform` here https://github.com/HabitRPG/habitica/blob/bd8e67a2ea3d6109ab2e43057ecdfe8f1be2e363/website/server/models/user/hooks.js#L25 so that it's present in every request and not only when getting the user. Also to avoid the issue with the `await`, if we refresh the last cached post with a callback called at an interval of X minutes (using `setInterval`) and not with a call to `await NewsPost.lastNewsPostID();` we can then skip the async call

test/api/v4/news/DELETE-news.test.js
```diff
@@ -0,0 +1,41 @@
...
+    await expect(nonAdminUser.post('/news')).to.eventually.be.rejected.and.eql({
+      code: 401,
+      error: 'NotAuthorized',
+      message: 'You don\'t have admin access.',
+    });
+  });
+
+  it('deletes news posts', async () => {
+    const existingPost = await user.post('/news', newsPost);
+    await user.del(`/news/${existingPost.id}`, {
```
paglias: LGTM but why is this passed a payload?

test/api/v4/news/GET-news.test.js
```diff
@@ -0,0 +1,16 @@
...
+describe('GET /news', () => {
+  let api;
+
+  beforeEach(async () => {
+    api = requester();
+  });
+
+  it('returns the latest news in json format, does not require authentication', async () => {
+    const res = await api.get('/news');
+    expect(res).to.be.a.string;
```
paglias: is res a string? also I'd test that it contains some fields or try to add 2 posts and see that it's returning the last one

test/api/v4/news/POST-news.test.js
```diff
@@ -0,0 +1,38 @@
+import {
+  generateUser,
+} from '../../../helpers/api-integration/v4';
+
+describe('POST /news/read', () => {
```
paglias: should it be POST /news?

test/api/v4/news/POST-news.test.js
```diff
@@ -0,0 +1,38 @@
...
+      code: 401,
+      error: 'NotAuthorized',
+      message: 'You don\'t have admin access.',
+    });
+  });
+
+  it('creates news posts', async () => {
+    const response = await user.post('/news', newsPost);
+
+    expect(response.title).to.equal(newsPost.title);
```
paglias, line 35: I would also check the result in the GET /news route

test/api/v4/news/POST-news_read.test.js
```diff
@@ -0,0 +1,20 @@
+import {
+  generateUser,
+} from '../../../helpers/api-integration/v4';
+import { model as NewsPost } from '../../../../website/server/models/newsPost';
+
+describe('POST /news', () => {
```
paglias: POST /news/read?

test/api/v4/news/PUT-news.test.js
```diff
@@ -0,0 +1,41 @@
+import {
+  generateUser,
+} from '../../../helpers/api-integration/v4';
+
+describe('PUT /news/:newsID', () => {
```
paglias: file name should be PUT-news_newsId

website/client/src/components/achievements/newStuff.vue
```diff
@@ -4,21 +4,36 @@
...
+        <small
+          v-if="!post.published"
+          class="draft"
+        >DRAFT</small><h2 class="title">
+          {{ post.publishDate | date }} - {{ post.title }}
+        </h2>
+        <hr>
+        <div
+          class="markdown"
+          v-html="renderMarkdown(post.text)"
```
paglias: for rendering markdown there's a `v-markdown` directive that seems to be used elsewhere

website/server/models/newsPost.js
```diff
@@ -0,0 +1,82 @@
...
+
+const NEWS_CACHE_TIME = 5 * 60 * 1000;
+
+let cachedLastNewsPost = null;
+
+schema.statics.lastNewsPost = function lastNewsPost () {
+  return cachedLastNewsPost;
+};
+
+schema.statics.updateLastNewsPost = async function updateLastNewsPost (newPost) {
```
paglias: the `async` should be removed

website/server/models/newsPost.js
```diff
@@ -0,0 +1,82 @@
...
+export const model = mongoose.model('NewsPost', schema);
+
+setInterval(async () => {
+  const lastPost = await model.getLastPost();
+  if (lastPost) {
+    model.updateLastNewsPost(lastPost);
+  }
+}, NEWS_CACHE_TIME);
+
+model.getLastPost().then(newsPost => {
```
paglias: can we add a `.catch` here? Promises outside of the request / response flow must be catched explicitly ,

```
.catch(err => logger.error(err))
``` will be enough

File: website/server/models/newsPost.js [hunk 37]
```diff
@@ -0,0 +1,82 @@
...
+  if (!cachedLastNewsPost || cachedLastNewsPost.id !== newPost.id) {
+    if (!cachedLastNewsPost || cachedLastNewsPost.publishDate < newPost.publishDate) {
+      cachedLastNewsPost = newPost;
+    }
+  }
+};
+
+export const model = mongoose.model('NewsPost', schema);
+
+setInterval(async () => {
```
paglias: setInterval doesn't really support `async` function so I would use a normal `.then` like below + a `.catch` in case of errors

website/server/libs/user/index.js
paglias: I think this has to be removed now? since it's added in `toJSONTransform`

website/server/controllers/api-v3/news.js
```diff
@@ -31,7 +31,6 @@ api.getNews = {
           <div class="mr-3 ${baileyClass}"></div>
           <div class="media-body">
             <h1 class="align-self-center">${res.t('newStuff')}</h1>
-            <h2>2/13/2020 - ${LAST_ANNOUNCEMENT_TITLE}</h2>
```
paglias: is this HTML still used?

website/server/controllers/api-v3/news.js
paglias: I think it can be removed

paglias (commented):
Sorry i got to this late, functionality wise it looks good 👍 I mostly left comments about some typos + a few questions about the tests which I'm also unsure if they're passing or not, I think this PR was created before we enabled Github Actions so it might only start running tests after develop is merged

paglias:
Given the recent discussion in Slack there are some things regarding permissions to be addressed before this can be merged

website/server/controllers/api-v3/news.js
phillipthelen: Not really. I left this here for compatibility reasons of the API. But it should probably be changed to return the html rendered version of the latest post? Or should I just remove the API call?

paglias: in theory we should have it return the html rendered version of the last post for compatibility but since it's used only internally afaik we might just skip this

paglias: Actually it might not be too difficult to return the HTML version of the post? Not sure if it's worth it though, depends on whether any 3rd party tool uses it

paglias:
I'm going to mark this one as `needs work` pending the updates for the permissions and the API v3 compatibility

paglias:
@phillipthelen looks good to me I think the only 2 things missing are:

- an unit test for the new permission middleware like we have for `ensureAdmin` https://github.com/HabitRPG/habitica/blob/develop/test/api/unit/middlewares/ensureAccessRight.test.js
- an update to the api v3 route so that it returns the most recent news https://github.com/HabitRPG/habitica/pull/10739/files#diff-1831d72ccd46d51a77b01ef87da3a614L31 - I'd say just some basic html with title and text should be enough

website/server/models/user/methods.js
```diff
@@ -295,6 +296,13 @@ schema.statics.transformJSONUser = function transformJSONUser (jsonUser, addComp
   if (!jsonUser.flags.verifiedUsername) jsonUser.auth.local.username = null;
 
   if (addComputedStats) this.addComputedStatsToJSONObj(jsonUser.stats, jsonUser);
+
+  jsonUser.flags.newStuff = this.checkNewStuff(jsonUser.flags.lastNewStuffRead);
+};
+
+schema.statics.checkNewStuff = function checkNewStuff (lastReadID) {
```
paglias: what's the difference between this static `checkNewStuff` and the method one defined below? Can we add a comment explaining the difference?

paglias: Actually it looks like this one isn't used anywhere

website/server/models/newsPost.js
```diff
@@ -0,0 +1,84 @@
+import mongoose from 'mongoose';
+import baseModel from '../libs/baseModel';
+import logger from '../libs/logger';
+
+const { Schema } = mongoose;
+const POSTS_PER_PAGE = 10;
+
+export const schema = new Schema({
+  title: { $type: String },
```
paglias: this isn't super important given access to the routes will be limited but there's no check nor here nor at the routes level for required fields or things that cannot be updated. Not sure if it's something we want to check for

paglias (commented):
@phillipthelen I think there's an unused method that can be removed + another comment about which I'm not sure. Apart from this I'd say it can go to delta (it can also go to delta before these comments are addressed)

website/server/models/user/methods.js
```diff
@@ -490,6 +498,15 @@ schema.methods.isAdmin = function isAdmin () {
   return this.contributor && this.contributor.admin;
 };
 
+schema.methods.isNewsPoster = function isNewsPoster () {
+  return this.contributor && this.contributor.newsPoster;
+};
+
+schema.methods.checkNewStuff = function checkNewStuff () {
+  const { id } = NewsPost.lastNewsPost();
```
paglias: I think the tests are failing because there's no id but `_id`

paglias (changes requested):
comment about the failing tests

paglias (commented):
@phillipthelen I think the tests are still failing because:

- you used `_id` on `statics.checkNewStuff` (which is not used) not on `methods.checkNewStuff`
- there are still a few tests relying on `.id`
- `lastNewsPost()` returns null if there's no post in the database so there should be a check like this

```
schema.methods.checkNewStuff = function checkNewStuff () {
  const lastNewsPost = NewsPost.lastNewsPost();
  return Boolean(lastNewsPost && this.flags.lastNewStuffRead !== lastNewsPost._id);
};

File: test/api/v4/news/GET-news.test.js [hunk 42]
```diff
@@ -0,0 +1,30 @@
...
+    title: 'New Post',
+    publishDate: new Date(),
+    published: true,
+    credits: 'credits',
+    text: 'news body',
+  };
+  beforeEach(async () => {
+    api = requester();
+    const user = await generateUser({
+      'contributor.admin': true,
```
paglias: should this be `contributor.newsPoster`?

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,257 @@
...
+    optional: true,
+  })],
+  noLanguage: true,
+  async handler (req, res) {
+    const { user } = res.locals;
+    const { page } = req.query;
+
+    let isAdmin = false;
+    if (user && user.contributor) {
+      isAdmin = user.contributor.admin;
```
paglias: should this be `contributor.newsPoster`?

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,257 @@
...
+  middlewares: [authWithHeaders({
+    optional: true,
+  })],
+  noLanguage: true,
+  async handler (req, res) {
+    req.checkParams('postId', apiError('postIdRequired')).notEmpty();
+    const { user } = res.locals;
+    let isAdmin = false;
+    if (user && user.contributor) {
+      isAdmin = user.contributor.admin;
```
paglias: should this be `contributor.newsPoster`? Also this check is duplicated, I think we could create a function like the `user.isAdmin()` one in the user model

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,257 @@
...
+ * @apiSuccessExample {json} Post:
+ *     HTTP/1.1 200 OK
+ *     {
+ *       "title": "News Title",
+ *       ...
+ *     }
+ *
+ * @apiUse postIdRequired
+ *
+ * @apiPermission Admin
```
paglias: should this be @apiPermission NewsPoster?

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,257 @@
...
+ * @apiSuccess {Object} data The create news post (See <a href="https://github.com/HabitRPG/habitica/blob/develop/website/server/models/newsPost.js" target="_blank">/website/server/models/newsPost.js</a>)
+ *
+ * @apiSuccessExample {json} Post:
+ *     HTTP/1.1 200 OK
+ *     {
+ *       "title": "News Title",
+ *       ...
+ *     }
+ *
+ * @apiPermission Admin
```
paglias: should this be @apiPermission NewsPoster?

website/server/controllers/api-v4/news.js
paglias: Looks like there's one alreadt 🎉 , sorry i had missed it, then we could use that one

test/api/v4/news/PUT-news_newsId.test.js
```diff
@@ -0,0 +1,41 @@
...
+    text: 'news body',
+  };
+  beforeEach(async () => {
+    user = await generateUser({
+      'contributor.newsPoster': true,
+    });
+  });
+
+  it('disallows access to non-admins', async () => {
+    const nonAdminUser = await generateUser({ 'contributor.': false });
```
paglias: 'contributor.newsPoster'

paglias (changes requested):

website/server/models/user/methods.js
phillipthelen: This one is used in the `transformJSONUser` method. Since that one is also static, I needed to add this static method because the method one isn't available. (If I misunderstood that somehow let me know. but it did give me an error before)

website/client/src/components/achievements/newStuff.vue
```diff
@@ -4,21 +4,36 @@
...
-      ></div>
+        v-for="(post, index) in posts"
+        :key="index"
+        :class="index == (posts.length - 1) ? 'bailey bailey-last' : 'bailey'"
+      >
+        <small
+          v-if="!post.published"
+          class="draft"
+        >DRAFT</small><h2 class="title">
+          {{ post.publishDate | date }} - {{ post.title }}
```
Alys: I don't know if you're interested in feedback about this sort of thing (so feel free to ignore this if it's not appropriate!) but is it possible for `post.publishDate` to use the user's preferred date format? It's the sort of thing that they're likely to ask for. Here's an example of how the format is used in case it speeds things up:
https://github.com/HabitRPG/habitica/blob/be622e8b10ce7f9e59056c77ad01762320d85064/website/client/src/components/userMenu/profile.vue#L775-L778

phillipthelen: That's a good point! I'll make that change. Thanks.

website/server/controllers/api-v3/news.js
```diff
@@ -71,11 +83,14 @@ api.tellMeLaterNews = {
   async handler (req, res) {
     const { user } = res.locals;
 
-    user.flags.newStuff = false;
+    const { id, title } = NewsPost.lastNewsPost();
```
paglias: should be _id instead of id?

website/server/controllers/api-v3/news.js
```diff
@@ -71,11 +83,14 @@ api.tellMeLaterNews = {
...
+    user.flags.lastNewStuffRead = id;
```
paglias: should be _id instead of id?

website/server/controllers/api-v3/news.js
```diff
@@ -71,11 +83,14 @@ api.tellMeLaterNews = {
...
 
-    const existingNotificationIndex = user.notifications.findIndex(n => n && n.type === 'NEW_STUFF');
-    if (existingNotificationIndex !== -1) user.notifications.splice(existingNotificationIndex, 1);
-    user.addNotification('NEW_STUFF', { title: LAST_ANNOUNCEMENT_TITLE }, true); // seen by default
+    if (user.notifications) {
```
paglias: I don't think this check is necessary

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,252 @@
...
+ *
+ */
+api.MarkNewsRead = {
+  method: 'POST',
+  middlewares: [authWithHeaders()],
+  url: '/news/read',
+  async handler (req, res) {
+    const { user } = res.locals;
+
+    const { id } = NewsPost.lastNewsPost();
```
paglias: should be _id instead of id?

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,252 @@
...
+ *
+ */
+api.tellMeLaterNews = {
+  method: 'POST',
+  middlewares: [authWithHeaders()],
+  url: '/news/tell-me-later',
+  async handler (req, res) {
+    const { user } = res.locals;
+
+    const { id, title } = NewsPost.lastNewsPost();
```
paglias: should be _id instead of id?

website/server/controllers/api-v4/news.js
```diff
@@ -0,0 +1,252 @@
...
+    user.flags.lastNewStuffRead = id;
+
+    if (user.notifications) {
```
paglias: I don't think this check is necessary

website/server/models/newsPost.js
```diff
@@ -0,0 +1,84 @@
...
+const NEWS_CACHE_TIME = 5 * 60 * 1000;
+
+let cachedLastNewsPost = null;
+
+schema.statics.lastNewsPost = function lastNewsPost () {
+  return cachedLastNewsPost;
+};
+
+schema.statics.updateLastNewsPost = function updateLastNewsPost (newPost) {
+  if (!cachedLastNewsPost || cachedLastNewsPost.id !== newPost.id) {
```
paglias: should be _id instead of id?

paglias (commented):
some occurences of id instead of _id

website/server/models/user/methods.js
paglias: You're right, but (this is super confusing and we should add at least a comment) `transformJSONUser` is only used [here](https://github.com/HabitRPG/habitica/blob/b947c714f0efb6de1b0558fff885913168890c82/website/server/controllers/api-v3/members.js#L398) because the User model isn't available when using `.lean()` and in that method `flags.newStuff ` should not be returned because it's not an user public field.

To get this field into normal user responses we should actually use the `schema.methods.checkNewStuff` in [here](https://github.com/HabitRPG/habitica/blob/b947c714f0efb6de1b0558fff885913168890c82/website/server/models/user/hooks.js#L25). Something like this

```
    ...
    if (originalDoc.isSelected('flags.lastNewStuffRead')) { // make sure the necessary field to calculate newStuff are selected
        plainObj.flags.newStuff = originalDoc.checkNewStuff();
    }

    return plainObj;
  },
```

website/client/src/components/achievements/newStuff.vue
```diff
@@ -36,33 +51,110 @@
...
   computed: {
     ...mapState({ user: 'user.data' }),
   },
   async mounted () {
-    this.$root.$on('bv::show::modal', async modalId => {
-      if (modalId !== 'new-stuff') return;
-      const response = await axios.get('/api/v4/news');
-      this.html = response.data.html;
-    });
+    this.posts = (await this.$store.dispatch('user:getNews')).data.data;
```
paglias: I've noticed that this API call on delta is made as soon as the page loads even if the bailey modal isn't shown, probably because it gets `mounted` immediately. it should probably be made only when the modal is shown by listening to `this.$root.$emit('bv::show::modal', 'new-stuff');`

paglias (changes requested):

website/common/locales/en/news.json
```diff
@@ -0,0 +1,8 @@
+{
```
paglias: this has to be uploaded manually to weblate?

paglias (approved):

SabreCat (approved):

PR #10665, opened 2018-09-05
 Don't show "no guilds" texts while loading

cvuorinen:
[//]: # (Note: See http://habitica.wikia.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API for more info)

[//]: # (Put Issue # here, if applicable. This will automatically close the issue if your PR is merged in)
Fixes #10662

### Changes
[//]: # (Describe the changes that were made in detail here. Include pictures if necessary)
Previously My Guilds page displayed the "You aren't member of any Guilds" message (and My Challenges page displayed "You don't have any challenges" message) while it was still loading them from the API and also when filtering so that there are no matches.

This PR adds loading indicator (just text "Loading..." as used in Discover Guilds and Discover Challenges pages) that is displayed while the API request is pending and only shows "no guilds" or "no challenges" message after loading is finished. Also, when filtering is used and there are no matches (but user has guilds/challenges), then nothing is shown (as in Discover Guilds and Discover Challenges pages). 

[//]: # (Put User ID in here - found on the Habitica website at User Icon > Settings > API)

----
UUID: fa756acf-9127-4a3c-8dac-8ff627d30073

PR #10744, opened 2018-10-09
No matching Guilds/Challenges message

cvuorinen:
See previous discussion here: https://github.com/HabitRPG/habitica/pull/10665#discussion_r219885307

### Changes
Display "We couldn't find any matching Guilds." message on 
* My Guilds page when user has Guilds but filters don't match anything
* Discover Guilds page when filters don't match anything

screenshot from 2018-10-09 09-46-05

Display "We couldn't find any matching Challenges." message on 
* My Challenges page when user has Challenges but filters don't match anything
* Discover Challenges page when filters don't match anything

screenshot from 2018-10-09 09-45-23

Also, hide the "Load More" button in Discover Challenges page when 
* there is nothing to load (filters don't match anything)
* while already loading.

----

### DIFF
--- a/website/client/components/challenges/findChallenges.vue
+++ b/website/client/components/challenges/findChallenges.vue
@@ -22,8 +22,9 @@
     .row
       .col-12.col-md-6(v-for='challenge in filteredChallenges')
         challenge-item(:challenge='challenge')
+
     .row
-      .col-12.text-center
+      .col-12.text-center(v-if='!loading && filteredChallenges.length > 0')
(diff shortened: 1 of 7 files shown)


negue:
If you search by text and delete it again, then the filtered out guilds won't reset, also changing the category won't help.

But it works for challenges 👍

Could you take a look at the guild search ?

cvuorinen:
@negue yeah, I'll take a look at it

cvuorinen:
Fixed the Guild search.

paglias:
thanks @cvuorinen ! notated towards your next tier

PR #10759, opened 2018-10-15
fix for #10496: Newly subscribed accounts receive erroneous notification that they have Mystery Items

### LINKED ISSUE #10496: Newly subscribed accounts receive erroneous notification that they have Mystery Items

derekdkim:
Fixes #10496

Issue #10496: Newly subscribed accounts receive erroneous notification that they have Mystery Items

Hi, I've fixed this issue but I'm stuck on implementing the automated unit test. Sorry about wasting your precious time. Although I studied Javascript, I have no experience with any of its frameworks or an app of this scale. Automated unit testing is also new for me. This is what I've done so far.

## Reproducing the issue: 
1. Created testHabitica account
2. Granted that account's guild (testGuild) a group subscription using the mongodb command 
3. Created testHabitica2 account
4. invited testHabitica2 by id to the testGuild
5. The erroneous notification appears.
Note: This doesn't appear on the test user that was granted 1 month of sub using the mongodb command.

## Fix:

```
if (pushedItems.length > 0) {
user.addNotification('NEW_MYSTERY_ITEMS', { items: pushedItems });
}
```
in revealMysteryItems() in subscriptions.js.

## Concerns (Need help):

I tested this manually using a third new account, ran through ```npm run test```. I'm currently in the process of writing an automated test for this but I'm completely lost.

The closest thing I could find to revealMysteryItems() in the unit tests is context('Mystery Items'). I'm not sure if this is what I should be adding my unit test to.

I'm also unfamiliar with the BDD and unit testing syntax. I cobbled together the following based on what I would like to test, but I have no idea how to properly write it.

```
it('does not send notification if pushedItems is empty', async () => {
    if (user.purchased.plan.mysteryItems.length == 0) {
        expect(user.addNotification('NEW_MYSTERY_ITEMS', { items: pushedItems })).to.not.have.been.called;
    }
});
```
Thanks for your patience. I have a lot to learn.

### DIFF
--- a/test/api/unit/libs/payments/payments.test.js
+++ b/test/api/unit/libs/payments/payments.test.js
@@ -434,7 +434,7 @@ describe('payments/index', () => {
         fakeClock.restore();
       });
 
-      it('does not awards mystery items when not within the timeframe for a mystery item', async () => {
+      it('does not add a notification for mystery items if none was awarded', async () => {
         const noMysteryItemTimeframe = 1462183920000; // May 2nd 2016
         let fakeClock = sinon.useFakeTimers(noMysteryItemTimeframe);
         data = { paymentMethod: 'PaymentMethod', user, sub: { key: 'basic_3mo' } };
@@ -442,6 +442,7 @@ describe('payments/index', () => {
         await api.createSubscription(data);
 
         expect(user.purchased.plan.mysteryItems).to.have.a.lengthOf(0);
+        expect(user.notifications.find(n => n.type === 'NEW_MYSTERY_ITEMS')).to.be.undefined;
 
         fakeClock.restore();
       });
--- a/test/api/unit/libs/payments/payments.test.js
+++ b/test/api/unit/libs/payments/payments.test.js
@@ -434,6 +434,18 @@ describe('payments/index', () => {
         fakeClock.restore();
       });
 
+      it('does not awards mystery items when not within the timeframe for a mystery item', async () => {
+        const noMysteryItemTimeframe = 1462183920000; // May 2nd 2016
+        let fakeClock = sinon.useFakeTimers(noMysteryItemTimeframe);
+        data = { paymentMethod: 'PaymentMethod', user, sub: { key: 'basic_3mo' } };
+
+        await api.createSubscription(data);
+
+        expect(user.purchased.plan.mysteryItems).to.have.a.lengthOf(0);
+
+        fakeClock.restore();
+      });
+
       it('does not add a notification for mystery items if none was awarded', async () => {
         const noMysteryItemTimeframe = 1462183920000; // May 2nd 2016
         let fakeClock = sinon.useFakeTimers(noMysteryItemTimeframe);
--- a/website/server/libs/payments/subscriptions.js
+++ b/website/server/libs/payments/subscriptions.js
@@ -36,8 +36,9 @@ function revealMysteryItems (user) {
       pushedItems.push(item.key);
     }
   });
-
-  user.addNotification('NEW_MYSTERY_ITEMS', { items: pushedItems });
+  if (pushedItems.length > 0) {
+    user.addNotification('NEW_MYSTERY_ITEMS', { items: pushedItems });
+  }
 }
 
 // @TODO: Abstract to payment helper


paglias:
Hi @derekdkim,

the fix looks good! As far as the test you'll want to test that when calling the `revealMysteryItems` if no mystery item is added to the user no notification is shown either.

Ideally you'd test the `revealMysteryItems` function directly but unfortunately that fn is not exported anywhere so you cannot test it directly. But it's called from this function that instead is exported revealMysteryItems (see the end of the file).

And it's unit tests here https://github.com/HabitRPG/habitica/blob/develop/test/api/unit/libs/payments/payments.test.js#L71 with something similar to what you want to do here

You'll want to add a test case that looks like this (copied from the test linked above):

```
 it('does not add a notification for mystery items if none was awarded', async () => {
        const noMysteryItemTimeframe = 1462183920000; // May 2nd 2016
        let fakeClock = sinon.useFakeTimers(noMysteryItemTimeframe);
        data = { paymentMethod: 'PaymentMethod', user, sub: { key: 'basic_3mo' } };

        await api.createSubscription(data);

        expect(user.purchased.plan.mysteryItems).to.have.a.lengthOf(0);
        expect(user.notifications.find(n => n.type === 'NEW_MYSTERY_ITEMS')).to.be.undefined; // NOTE: I added this line

        fakeClock.restore();
});
```

derekdkim:
Thanks for the help, @paglias . 

I ran ```npm run test:api:unit``` and I confirmed that these tests passed.

 Mystery Items
        ✓ awards mystery items when within the timeframe for a mystery item (72ms)
        ✓ does not add a notification for mystery items if none was awarded (47ms)
        ✓ does not award mystery item when user already owns the item
        ✓ does not award mystery item when user already has the item in the mystery box (47ms)
I'll add the test to the PR.

There's something I want to try to understand for future reference. What is the purpose of the timer in the context of checking for the presence of a mystery item?

Edit: Ruh roh, it seems like it fails the CI test now.

1) GET /hall/patrons
       returns only first 50 patrons per request, more if req.query.page is passed:
     Error: Timeout of 8000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/travis/build/HabitRPG/habitica/test/api/v3/integration/hall/GET-hall_patrons.test.js)

I don't think this is related to the code changes I made since all I did was make an if statement and added a line in an existing test in payments test. Do you know how I can fix this?

test/api/unit/libs/payments/payments.test.js
```diff
@@ -434,14 +434,15 @@ describe('payments/index', () => {
         fakeClock.restore();
       });
 
-      it('does not awards mystery items when not within the timeframe for a mystery item', async () => {
```
paglias: @derekdkim the code looks good but this way one of test has been removed, can you create a separate one for the notification while keeping the old one in place? they're very similar but it's better to have two of them

paglias (changes requested):

paglias:
@derekdkim don't worry about the failure, it's unrelated, I've left one comment but otherwise it looks good.

> There's something I want to try to understand for future reference. What is the purpose of the timer in the context of checking for the presence of a mystery item?

It's there to make sure the server thinks we're not in a period of time when the mystery items should be awarded. Because otherwise tests running while mystery items are awarded would fail.

derekdkim:
@paglias Done. Sorry, I should have known better than to remove a previous test.

> It's there to make sure the server thinks we're not in a period of time when the mystery items should be awarded. Because otherwise tests running while mystery items are awarded would fail.

Many thanks for the explanation. It all makes sense now.

Also, is there a way for me to "resolve" a review? or is it something that an admin/maintainer manually decides?

paglias:
@derekdkim no worries about the test! Regarding the review I think that a maintainer has to manually resolve it, I'll do it in the next few days (we usually review PRs at the end of each week)

paglias:
@derekdkim  thanks for this! I'm merging the PR, if you have an Habitica UUID I'll be happy to award you a contribution tier.

derekdkim:
@paglias Thanks! I learned a great deal from this and I'll definitely keep helping out as my skills improve.

My Habitica UUID is 49b3a33c-6f44-453f-96b6-0c163a3154ac

paglias:
tier awarded!

PR #10760, opened 2018-10-15
check every armoire gear 'canOwn' method

### LINKED ISSUE #10659: some lost Armoire equipment is incorrectly locked and can't be found in Shop
MarieFrog (26cace60-f1f3-4678-bd66-23d9571dfc8c):
"I've lost a level and lost a piece of equipment which is now pinned to my rewards column, so far so good. I can't buy it though because it says "locked until you've purchased the previous items in the sequence". It's the Jeweled Archer Armor"

Confirmed. When I set `items.gear.owned.armor_armoire_jeweledArcherArmor` to `false` on my account and add it to my `pinnedItems` array, it does appear in my pinned items on the website but is locked with the `tierLockedItem` message. It should be unlocked because it's an Armoire piece - when you lose it due to Death, you should be able to buy it back immediately.

Also the item does not appear in the Shop for me to buy back. It should appear under the "None" class, otherwise if you don't have it pinned, you can't rebuy it.

The same problems do not occur for `items.gear.owned.weapon_armoire_poisonedGoblet` - when I set it to have a false value, it is not locked and it does appear in the Shop.

The incorrect behaviour for the former piece might be because it has "armor" in its key - that might be triggering code that should be used only for the basic class-specific gear (which should be locked if you don't have cheaper items in the set).

negue:
adding a test to check all armoire gear if the canOwn methods work as expected

fixes #10659

### DIFF
--- a/test/common/ops/armoireCanOwn.js
+++ b/test/common/ops/armoireCanOwn.js
@@ -1,0 +1,21 @@
+import armoireSet from '../../../website/common/script/content/gear/sets/armoire';
+
+describe('armoireSet items', () => {
+  it('checks if canOwn has the same id', () => {
+    for (const type of Object.keys(armoireSet)) {
+      for (const itemKey of Object.keys(armoireSet[type])) {
+        const ownedKey = `${type}_armoire_${itemKey}`;
+
+        expect(armoireSet[type][itemKey].canOwn({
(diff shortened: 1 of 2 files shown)


paglias:
perfect, thanks for the PR and the test!

PR #10763, opened 2018-10-16
Add config option to restrict access to test servers

phillipthelen:
This change allows a server to be marked as `IS_TEST_SERVER`. This disables registration and only allows users with `flags.isTrustedTestUser` to log in and perform actions. This way those servers don't have to be secured by HTTP auth and we can additionally use the new flag for client side testing in p.e. the mobile apps.


website/common/locales/en/generic.json
```diff
@@ -296,5 +296,6 @@
   "selected": "Selected",
   "howManyToBuy": "How many would you like to buy?",
   "habiticaHasUpdated": "There is a new Habitica update. Refresh to get the latest version!",
-  "contactForm": "Contact the Moderation Team"
+  "contactForm": "Contact the Moderation Team",
+  "testServerNoRegistration": "Test servers don't allow registration"
```
SabreCat: I don't expect this message to be something the general population will see (if they do, there's already been at least a minor goof somewhere). I don't think this needs to be a translated string, then, though maybe @paglias can weigh in.

SabreCat (commented):
Might we want to actually remove the HTTP Basic Auth layer? This seems like a superior approach in a few ways, and it'd cut a bit of code cruft.

Also, might this merit at least a basic unit test?

website/server/controllers/api-v3/auth.js
```diff
@@ -67,11 +67,15 @@ api.registerLocal = {
  * @apiSuccess {String} data.apiToken The user's api token that must be used to authenticate requests.
  * @apiSuccess {Boolean} data.newUser Returns true if the user was just created (always false for local login).
  */
+const IS_TEST_SERVER = nconf.get('IS_TEST_SERVER');
 api.loginLocal = {
   method: 'POST',
   url: '/user/auth/local/login',
   middlewares: [],
   async handler (req, res) {
+    if (IS_TEST_SERVER === true) {
```
paglias, line 76: why on the login route as well?

website/server/middlewares/testServerMode.js
```diff
@@ -0,0 +1,25 @@
+import nconf from 'nconf';
+import {model as User} from '../models/user';
+
+const IS_TEST_SERVER = nconf.get('IS_TEST_SERVER');
+
+module.exports = function maintenanceMode (req, res, next) {
```
paglias: the name of the function is still maintenanceMode

website/common/locales/en/generic.json
paglias: agree with @SabreCat here, if it's used in multiple places you can put it in libs/apiMessages but I don't think it's necessary to have it in transifex

website/server/libs/auth/index.js
```diff
@@ -71,7 +72,12 @@ function hasBackupAuth (user, networkToRemove) {
   return hasAlternateNetwork;
 }
 
+const IS_TEST_SERVER = nconf.get('IS_TEST_SERVER');
+
 async function registerLocal (req, res, { isV3 = false }) {
+  if (IS_TEST_SERVER === true) {
+    throw new NotAuthorized(res.t('testServerNoRegistration'));
```
paglias: I'm a bit worried about disabling registration on test servers, it's a feature often used especially on servers that don't use the production database where otheriwse it would be impossible to regsiter a new user and also on staging if we want to test something related to user registration

website/server/middlewares/testServerMode.js
```diff
@@ -0,0 +1,25 @@
...
+  if (IS_TEST_SERVER !== true) return next();
+  const userId = req.header('x-api-user');
```
paglias: My worry here is that this will always return false for routes that don't require authentication, making it impossible to use them on test servers

paglias (commented):
I like it as an alternative to HTTP auth but I feel like there are some issues to be resolved

paglias:
Yep an unit test here would be quite important, I like it as an alternative to the HTTP auth (which should then be removed from the middlewares) but right now it's two problems

1. It makes it impossible to register new users on test servers which is important both to be able to test changes to the register and login features (also would make it impossible to login since that route is disabled) and would make it impossible to use test servers using a different database (see beta)
2. It makes it impossible to access any route /page that doesn't require authentication like the static pages, the route to fetch a public profile, ...

website/server/middlewares/testServerMode.js
phillipthelen: I believe they worked for me, but I can check again.

website/server/libs/auth/index.js
phillipthelen: This is a flag we don't have to use on all servers. Which is also why I wouldn't remove the HTTP auth option. Servers that use the production database can use this system and those that do not can use HTTP Auth.

phillipthelen:
As mentioned in my other comment, I would keep the HTTP auth option. That way we can still use that for servers where we want to test registration and/or don't use the production database. The idea for this was to remove the need for HTTP auth on staging specifically which would allow us to add and remove users from being able to test on that more easily.

I'll fix the login route, since it wasn't intended to be completely disabled.

The static pages worked for me when testing, but I can check again to make extra sure.

config.json.example
```diff
@@ -4,6 +4,7 @@
     "IP":"0.0.0.0",
     "WEB_CONCURRENCY":1,
     "BASE_URL":"http://localhost:3000",
+    "IS_TEST_SERVER": false,
```
paglias: this should be a string, because locally you can use booleans but on the prod server env variables are strings

website/server/middlewares/testServerMode.js
paglias: So I checked and they work because if no user is found line 17 here is never reached so it goes to the next directly after. So if you pass a wrong user id or no user id at all the check is skipped.

Now that I think about it there's a security issue here because we're just checking the user id which is publicly avaible so you just have to copy the user id of an admin and you'll be able to access the test server freely

phillipthelen: But shouldn't that then fail the actual authentication? So that while this test might be bypassed, the user could still not perform any actions since the server will then deny access because the api key won't match. But I can also add a check for the key here.

phillipthelen:
Refactored the middleware into the auth middleware. But I'm running into issues creating a unit test. Is there a way to set the nconf variable so that the unit tests runs as a test server?

paglias:
You should be able to do it by using a mix of `sandbox` and `requireAgain` like here https://github.com/HabitRPG/habitica/blob/develop/test/api/unit/middlewares/analytics.test.js#L29-L37

Issue #9368, opened 2017-11-02
Odd Pet Feeding Behavior


lemoness:
We've received some reports that sometimes when a user quick-hatches a pet where they already have the mount (not sure if it happens other times), the greyed-out don't-have-this-pet-but-do-have-the-mount icon doesn't update to a regular colored have-this-pet icon until a refresh. This means that, until I reload, they can still click on the greyed-out icon and get the dialogue for hatching the pet again, except that the buttons don't do anything any more. This seems to be a similar problem to the one that people sometimes report when they tame a pet into a mount where the feeding bar disappears, but the pet icon doesn't grey out until they reload the page, preventing them from hatching a new one.

We've also had a follow-up report: "The thing I discovered today when hatching pets (in cases where I already have the mount) was that the bug only seemed to affect pets hatched after the first. So if i hatch just one pet for which I have the mount, the new pet's full-color icon will show up. But the second, third, and presumably etc. pets I hatch in that situation don't get their full-color icons until I reload the page. I can't test other pet growth stages right now, but hopefully this is at least another data point for those working to resolve the issue."

Beth has added: "When I raise a pet to a mount, the pet icon does grey out. But, when I hatch the pet to replace the mount, 1) there does not seem to be any obvious indicator that I can or cannot hatch the pet 2) when I hatch the new pet the icon stays greyed out until I do a full refresh (but not always- weirdly this issue seems to happen with the classic pet set but maybe not with quest ones?)

I've also had a couple where it seems like it auto-hatched the new pet after I raised the mount. I raised a Zombie seahorse, but due to sync errors clicked it maybe an extra time? Then all of a sudden I had the pet without seeing the hatching popup at all. But then on refresh I had to hatch it again."

We should definitely investigate these.

TheHollidayInn:
This issue doesn't seem to be happening for me. I'd suggest we close unless someone else can recreate or we get more reports.

Alys:
A related problem was just posted by NoCyg (154802ac-9567-4916-ac66-79ea5ba81ec7) in the Report a Bug guild:

"In Chrome, I discovered that when you raise a pet to a mount, the pet icon does not automatically switch to grayed out (indicating that you hatched the pet but do not currently have one). This led to a weird instance where I'd raised the pet to a mount, then clicked on its icon to try to re-hatch it, but instead ended up setting the pet (which I didn't actually have) as my avatar pet. This did resolve on a refresh (although I did have to re-select the avatar pet I actually wanted)."

[...]

PR #10768, opened 2018-10-18
refactor - pet image states

### LINKED ISSUE #10273: Pet image / hatching / rehatching confusion in the Stable; inconsistency with mobile apps
We've had some bug reports about the way that pets are displayed in the Stable on the website and how it's confusing for telling which pets/mounts you own, harder to tell how complete your pet/mount collection is, and inconsistent with the mobile apps:

"a large portion of the pets in my stable have reverted to the black paw with question mark. I had the full set of basic ones and now about 2/3 are reset to not hatched. i have tried logging in and out a few times no effect. The odd thing is that they are all still there on my iphone app!"

"On log on yesterday, I noticed that about half of my hatched pets and fed mounts had disappeared - I assumed it was a glitch of some sort and left it at that. It is still like that this morning. My OCD about completing the rows gets jittery every time I look at the page :("

"Having a similar issue as others here. My three pet snakes have reverted to empty paw prints, though they show in the mounts list. Incidentally, this happened right after I completed the quest that makes those eggs (though I hatched the ones missing a long while ago)."

"I have most of the pets (I am lacking some squirrels) and have fed almost all of them into mounts. Now, if I get new eggs, the black paw print shows up instead of the faded pet, so it looks like I don't have the mount. If I don't have eggs, the faded pet still shows."

"a large portion of my pets have dissappeared. My phone app still shows me as having all the basic pets hatched and my acheivements still supports this, but on my PC version more than half are reset back to "not there" I have tried logging in and our a few times."

"Did they change something in the Stables? Previously, pets that turn into mounts are grayed out but now, they turn back into a paw (which only shows if I haven't hatched that pet yet)."

"My stable on the mobile app appears to be full (I have 90/90) but on the desktop version I now have available spaces for all of the pets that have already grown into mounts. Deleting browser cache doesn't appear to help, nor does logging out and coming back in again. Seems to happen in both Safari and Chrome."

It's come from https://github.com/HabitRPG/habitica/pull/10243, which has resulted in these possible states on the Pets section of the Stable on the website (point 4 is the new one that's causing the bug reports):

1. You've never hatched the pet and you can't hatch it now: grey pawprint.
2. You've never hatched the pet but you can hatch it now: black pawprint.
3. You've hatched the pet and turned it into a mount and cannot re-hatch it now: greyed-out pet image.
4. You've hatched the pet and turned it into a mount and can re-hatch it now: black pawprint.

Previously for point 4, you'd see the greyed-out pet image for that condition, which was consistent with the mobile apps, made it clear which mounts you had, and promoted a better sense of completion towards collecting the full set of pets and mounts. Now it looks like you've never owned that pet.

https://github.com/HabitRPG/habitica/pull/10243 was intended to fix https://github.com/HabitRPG/habitica/issues/9368 but that issue is still happening, so I'm wondering if it would be acceptable for https://github.com/HabitRPG/habitica/pull/10243 to be reverted. Or was there an additional reason for that change to be made?

If the new behaviour is going to be kept, the same change might need to be considered for the mobile apps, if consistent behaviour across platforms is desired.

negue:
Refactor `petItem` and the pet image states

closes #10273

### DIFF
--- a/website/client/components/inventory/stable/petItem.vue
+++ b/website/client/components/inventory/stable/petItem.vue
@@ -5,7 +5,10 @@ div
       :class="{'item-empty': emptyItem, 'highlight': highlightBorder}",
     )
       slot(name="itemBadge", :item="item")
-      span.item-content(:class="itemContentClass")
+      span.item-content.hatchAgain(v-if="mountOwned && isHatchable")
+        span.egg(:class="eggClass")
+        span.potion(:class="potionClass")
+      span.item-content(v-else, :class="getPetItemClass(item)")
       span.pet-progress-background(v-if="item.isAllowedToFeed() && progress > 0")
         div.pet-progress-bar(v-bind:style="{width: 100 * progress/50 + '%' }")
     span.item-label(v-if="label") {{ label }}
@@ -15,7 +18,17 @@ div
     :triggers="showPopover ? 'hover' : ''",
     :placement="popoverPosition",
   )
-    slot(name="popoverContent", :item="item")
+    div.hatchablePopover(v-if="item.isHatchable()")
+      h4.popover-content-title {{ item.name }}
+      div.popover-content-text(v-html="$t('haveHatchablePet', { potion: item.potionName, egg: item.eggName })")
+      div.potionEggGroup
+        div.potionEggBackground
+          div(:class="potionClass")
+        div.potionEggBackground
+          div(:class="eggClass")
+    div(v-else)
+      h4.popover-content-title {{ item.name }}
+
 </template>
 
 <style lang="scss">
@@ -32,6 +45,29 @@ div
     height: 4px;
     background-color: #24cc8f;
   }
+
+  .hatchAgain {
+    display: inline-flex;
+    align-items: center;
+
+    width: 94px;
+    height: 94px;
+
+    .egg {
+      position: absolute;
+      left: 4px;
+      top: 14px;
+      z-index: 1;
+      transform: scale(1.2);
+    }
+
+    .potion {
+      position: absolute;
+      right: 4px;
+      top: 14px;
+      transform: scale(1.2);
+    }
+  }
 </style>
 
 <script>
@@ -41,9 +77,6 @@ div
     props: {
       item: {
         type: Object,
-      },
-      itemContentClass: {
-        type: String,
       },
       label: {
         type: String,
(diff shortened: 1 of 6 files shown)


website/client/components/inventory/stable/petItem.vue
```diff
@@ -78,6 +111,43 @@ div
       click () {
         this.$emit('click', {});
       },
+      getPetItemClass (pet) {
```
paglias: why is this a method? since the passed argument (`pet`) can be substitued with a `this.item` it could be a computed property

paglias: why is this a method? since the passed argument (`pet`) can be substitued with a `this.item` it could be a computed property

website/client/components/inventory/stable/petItem.vue
```diff
@@ -78,6 +111,43 @@ div
...
+      },
+    },
+    computed: {
+      potionClass () {
+        return `Pet_HatchingPotion_${this.item.potionKey}`;
+      },
+      eggClass () {
+        return `Pet_Egg_${this.item.eggKey}`;
+      },
+      isOwned () {
```
paglias: computed properties like this seem used only in the getPetItemClass, why don't call the functions directly?

paglias (changes requested):

website/client/components/inventory/stable/petItem.vue
negue: oh yeah that shouldn't be there ...

website/client/components/inventory/stable/petItem.vue
negue: only `mountOwned && isHatchable` in the vue, but I could change it to the method content too 

https://github.com/HabitRPG/habitica/pull/10768/files#diff-47365d041780039f2290fda1096a54feR8

website/client/components/inventory/stable/petItem.vue
```diff
@@ -78,6 +111,40 @@ div
       click () {
         this.$emit('click', {});
       },
+      getPetItemClass () {
```
paglias, line 114: can this be a compputed property instead?

paglias:
The code works but it looks like if you hatch a new pet in the stable it doesn't show up until you sync or refresh

citrusella:
I think that's due to #9368 which was supposed to be fixed by the pull request that caused the issue that this is fixing? Is this intended to fix #9368 too? (I made this post just to highlight that info in case, but now I'm genuinely curious? XP)

paglias:
> The code works but it looks like if you hatch a new pet in the stable it doesn't show up until you sync or refresh

I think the issue is that the computed property that manages the class isn't updated correctly when `isOwned()` changes

negue:
@paglias I refactored the `isOwned` and other checks, but that didn't work with computed properties so I had to use methods instead

paglias:
@negue sounds good, I fixed a lint issue and merged

Issue #10721, opened 2018-09-25
Contrast very low on reward control area


citrusella:
(Was asked by Alys to post here when I asked the best place to bring this up in Aspiring Socialites; spent a few days seeing if I could find other *glaring* issues (there are other places on the website I'm pretty sure also fail [WCAG (Web Content Accessibility Guidelines) AA contrast minimums](https://www.w3.org/TR/WCAG/#contrast-minimum), but the reward column was the most problematic and I didn't manage to get the chance to see if there were other especially bad areas in contrast or other accessibility)

The better part of a week ago, there was discussion in Aspiring Socialites about the similarity in color between tier 7 and the moderator color (though there are other non-color indicators, which helps on some level). That inspired me to run the site through a colorblindness simulation, at which point I remembered I'd thought a few times that the reward price area (the "checkoff" section of the reward) had very low contrast between the yellow price and the slightly paler yellow background, so I included it in the Imgur album I made of screenshots I took from colorblindness simulations: https://imgur.com/a/J7OtRGQ  (That was mostly comparison of tier colors since I was still doing it mostly out of curiosity when I started; the reward price is at the end and wasn't put through as many tests. The extensions I used are mentioned in the album description.)

After that, I got interested in some overall accessibility so I tried running the site through a few accessibility check extensions (specifically, [this](https://addons.mozilla.org/en-US/firefox/addon/wave-accessibility-tool) and [this](https://addons.mozilla.org/en-US/firefox/addon/tota11y-accessibility-toolkit)) and my browser wasn't having it so I didn't get other types of accessibility information (or much info about other low contrast areas).

However, I was concerned enough that I did put the values from the reward column's text and background color into a website and found out [its contrast ratio is approximately 1.59](https://contrastchecker.com/?c=ffa623&b=f3e6d5). (That's using the hex color I got from the eyedropper in Firefox's site inspector since the actual value used for the color is translucent so [this other site I used that allows for that kind of value](https://contrast-ratio.com/#%23FFA623-on-rgba%28255%2C%20217%2C%20160%2C%200.32%29) couldn't give anything firmer than between 1 (which would mean the background and foreground are the same color) and 4.72 (which would be a AA pass at the very least for any size). ...That feels like a functionally useless assessment.)

[...]

PR #10769, opened 2018-10-18
Custom reward action background and font-color updates

Tressley:
Partially addresses the feedback in #10721

**Before:**
image

**After:**
image

### DIFF
--- a/website/client/assets/scss/task.scss
+++ b/website/client/assets/scss/task.scss
@@ -243,7 +243,7 @@
     &-control {
       &-bg { 
         background: rgba(255, 217, 160, 0.32) !important;
-        .small-text { color: #EE9109 !important; }
+        .small-text { color: $orange-10 !important; }
 
         &:hover { background: rgba(255, 217, 160, 0.48) !important; }
       }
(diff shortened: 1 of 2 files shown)


website/client/assets/scss/task.scss
```diff
@@ -243,7 +243,7 @@
     &-control {
       &-bg { 
         background: rgba(255, 217, 160, 0.32) !important;
-        .small-text { color: $yellow-10 !important; }
+        .small-text { color: #EE9109 !important; }
```
SabreCat: This is in colors.scss as `$orange-10` (https://github.com/HabitRPG/habitica/blob/v4.65.5/website/client/assets/scss/colors.scss#L63), so you can use that instead of the raw hex value.

SabreCat (changes requested):
Yay first PR! One simple suggestion inline.

website/client/assets/scss/task.scss
Tressley: It looks like we have two instances of `$orange-10` is that a concern?

PR #10783, opened 2018-10-23
If user's cron will happen later today, set the Daily's Start Date to yesterday.

### LINKED ISSUE #5525: Newly created Daily's Start Date should be yesterday if the time of creation is between midnight and CDS
_edit by admin:_

_Currently, when a Daily is created, the start date is set to the current real-world date (i.e., the date that your PC's clock reports)._ 
_However if you have a Custom Day Start (CDS) set and if you create the Daily after midnight but before the CDS, then the start date should be set to yesterday (because HabitRPG-time is still "yesterday" until the CDS time is reached). This allows the new Daily to be active as soon as it is created._

**Suggested fix**:

When a Daily is created, get the current hour from the date object. If the hour is lower than the Custom Day Start, subtract 24 hours from the date and then set it as the startDate. Here's how it's already done in the iOS app:
https://github.com/HabitRPG/habitrpg-ios/blob/master/HabitRPG/TableViewController/Tasks/HRPGFormViewController.m#L119 

(Thanks @vIiRuS)

---

**Edit by hairlessbear**: for context, the original title of this issue was "Newly created dailies are automatically grey, but are still due every day". Original post is below.

---

[Video of the behavior](https://vid.me/ty0q)

In the video, a tag is selected when the daily is created, but the behavior occurred regardless of tag selection. It also occurred after reloading the page, as well as logging out and back in again.

In the DDT, the new "grey" dailies don't actually appear to be grey. They are categorized as normal dailies, so I suspect this is just a display problem.

If it helps, UUID is 8ac7242d-8f18-4d6f-b4f4-8f6302f7fa0b. I noticed this behavior occurring at around 00:30, and my cron is at 1.

natfarleydev:

Fixes #5525 

### Changes
When creating a task of any kind, if cron happens later in the same day (i.e. if a task is created between midnight and a custom day start), the start date for the task is now yesterday.

----
UUID: ecc53331-4aa1-4b02-b3b3-e794d0ff6827

### DIFF
--- a/test/common/libs/taskDefaults.test.js
+++ b/test/common/libs/taskDefaults.test.js
@@ -1,8 +1,9 @@
 import taskDefaults from '../../../website/common/script/libs/taskDefaults';
+import { generateUser } from '../../helpers/common.helper';
 
 describe('taskDefaults', () => {
   it('applies defaults to undefined type or habit', () => {
-    let task = taskDefaults();
+    let task = taskDefaults({}, generateUser());
     expect(task.type).to.eql('habit');
     expect(task._id).to.exist;
     expect(task.text).to.eql(task._id);
@@ -18,7 +19,7 @@ describe('taskDefaults', () => {
   });
 
   it('applies defaults to a daily', () => {
-    let task = taskDefaults({ type: 'daily' });
+    let task = taskDefaults({ type: 'daily' }, generateUser());
     expect(task.type).to.eql('daily');
     expect(task._id).to.exist;
     expect(task.text).to.eql(task._id);
@@ -42,7 +43,7 @@ describe('taskDefaults', () => {
   });
 
   it('applies defaults a reward', () => {
-    let task = taskDefaults({ type: 'reward' });
+    let task = taskDefaults({ type: 'reward' }, generateUser());
     expect(task.type).to.eql('reward');
     expect(task._id).to.exist;
     expect(task.text).to.eql(task._id);
@@ -52,7 +53,7 @@ describe('taskDefaults', () => {
   });
 
   it('applies defaults a todo', () => {
-    let task = taskDefaults({ type: 'todo' });
+    let task = taskDefaults({ type: 'todo' }, generateUser());
     expect(task.type).to.eql('todo');
     expect(task._id).to.exist;
     expect(task.text).to.eql(task._id);
--- a/test/common/libs/taskDefaults.test.js
+++ b/test/common/libs/taskDefaults.test.js
@@ -73,7 +73,7 @@ describe('taskDefaults', () => {
 
     expect(task.startDate).to.eql(
       moment()
-        .utcOffset(user.preferences.timezoneOffset, 'hour')
+        .zone(user.preferences.timezoneOffset, 'hour')
         .startOf('day')
         .subtract(1, 'day')
         .toDate()
--- a/test/common/libs/taskDefaults.test.js
+++ b/test/common/libs/taskDefaults.test.js
@@ -2,7 +2,6 @@ import moment from 'moment';
 
 import taskDefaults from '../../../website/common/script/libs/taskDefaults';
 import { generateUser } from '../../helpers/common.helper';
-import { constants } from 'os';
 
 describe('taskDefaults', () => {
   it('applies defaults to undefined type or habit', () => {
--- a/test/common/libs/taskDefaults.test.js
+++ b/test/common/libs/taskDefaults.test.js
@@ -1,5 +1,8 @@
+import moment from 'moment';
+
 import taskDefaults from '../../../website/common/script/libs/taskDefaults';
 import { generateUser } from '../../helpers/common.helper';
+import { constants } from 'os';
 
 describe('taskDefaults', () => {
   it('applies defaults to undefined type or habit', () => {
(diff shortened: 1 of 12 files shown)


natfarleydev:
I'm unsure of how to fix the tests as I use `moment()` to test whether the user's CDS is in the future.

Any advice on this would be appreciated.

EDIT: commits after this message halfway through attempting to correct the tests.

paglias:
@nasfarley88 looking at the tests for the `Added default dayStart to taskDefaults. ` commits looks like the only problem is in the client side tests, but they don't seem related to your changes

natfarleydev:
@paglias This is what I thought, but if I remove my changes to `taskDefaults`, then the tests don't seem to break. However, I'm quite new to the Habitica codebase so happy to be proven wrong.

paglias:
@nasfarley88 ok, let me know if you have any update otherwise I'll give it a look in the next few days

natfarleydev:
OK, it looks like I have problems with my personal setup as the `develop` branch gives me the same errors with `npm run test:common`. Since these are not related to the changes for #10783, if TravisCI passes I'm happy for this to be merged.

I've rebased on the latest `develop` and the client tests now work.

website/client/components/chat/copyAsTodoModal.vue
```diff
@@ -45,7 +45,7 @@ export default {
         type: 'todo',
         notes,
       };
-      this.task = taskDefaults(newTask);
+      this.task = taskDefaults(newTask, message.user);
```
paglias: I think message.user here is a string not an object, you can access it from the store `this.$store.state.user.data`

website/common/script/libs/taskDefaults.js
```diff
@@ -9,7 +9,7 @@ import moment from 'moment';
 
 const tasksTypes = ['habit', 'daily', 'todo', 'reward'];
 
-module.exports = function taskDefaults (task = {}) {
+module.exports = function taskDefaults (task = {}, user = { preferences: { dayStart: 0 } }) {
```
paglias: The default user parameter works well but I think there should be a check instead that throws an error is the user object isn't passed since in that case the startDate would be incorrect

paglias (changes requested):
left a couple of comments!

website/common/script/libs/taskDefaults.js
natfarleydev: What's the best way to raise an error in this context? Something like:

```javascript
throw new Error('No user passed to taskDefaults.')
```
(I come from Python, so still figuring out the JavaScript equivalents.)

Or perhaps it would be better to remove default arguments all together (since `taskDefaults` is never called without a task in the current codebase).

paglias: Actually yeah I think removing the default parameter should be enough as it'll throw an error automatically if no user is passed

natfarleydev:
I forgot to fix the test for `taskDefaults`, I'll fix that when I get in front of my usual dev rig.

paglias:
I think the tests are fine, just add back the default parameter for `task` and they should work

natfarleydev:
I thought it would be neater to fix the tests since we'll always need the user otherwise they'll be an error when testing to see when the users CDS is. 

This PR is now ready for merging.

website/common/script/libs/taskDefaults.js
```diff
@@ -76,7 +76,10 @@ module.exports = function taskDefaults (task = {}) {
         s: true,
         su: true,
       },
-      startDate: moment().startOf('day').toDate(),
+      // If cron will happen today, start the daily yesterday
+      startDate: moment().startOf('day').add(user.preferences.dayStart, 'hour').isAfter(moment()) ?
```
paglias: I think there are two changes to make here:

- it should also use the user's timezone to determine when the start of the day is (user.preferences.timezoneOffset)
- `moment().startOf('day')` should be cached

test/common/libs/taskDefaults.test.js
```diff
@@ -1,8 +1,9 @@
 import taskDefaults from '../../../website/common/script/libs/taskDefaults';
+import { generateUser } from '../../helpers/common.helper';
 
 describe('taskDefaults', () => {
```
paglias, line 6: can you add a new test for the new functionalities?

paglias (changes requested):

test/common/libs/taskDefaults.test.js
natfarleydev: Happy to add this test. The current tests don't work, I think due to my timezone (UK), so I may have to fix these first.

natfarleydev:
Tests added+fixed and now uses utcOffset to calculate the right start date.

website/common/script/libs/taskDefaults.js
```diff
@@ -65,6 +65,14 @@ module.exports = function taskDefaults (task, user) {
   }
 
   if (task.type === 'daily') {
+    let now = moment().utcOffset(user.preferences.timezoneOffset);
```
paglias: can you use `.zone` instead? it's deprecated but it's the correct method to use with the data we store

test/common/libs/taskDefaults.test.js
```diff
@@ -1,5 +1,8 @@
+import moment from 'moment';
+
 import taskDefaults from '../../../website/common/script/libs/taskDefaults';
 import { generateUser } from '../../helpers/common.helper';
+import { constants } from 'os';
```
paglias: can you remove this? it's not used anywhere

natfarleydev: Oops, accidental automatic import from VSCode! Now fixed.

natfarleydev:
The last two commits should fix the remaining issues.

paglias (approved):

paglias:
this is working well, thanks @nasfarley88 ! I've notated this PR towards your 4th tier

cc @Alys in case you want to notify someone about this change

PR #10784, opened 2018-10-26
(WIP) Web Usernames MVP

### LINKED ISSUE #10697: WEB [Nov 2] Implement modal that forces users to change username
Once we release the username feature, users have to set/confirm their username

Fixes needed from review:

- [x] Allow modal to grow correctly to accommodate validation errors
- [x] Check on why z-indexing issue sometimes recurs
- [x] Implement modal stack quick fix suggested by Matteo
- [x] Move email validations to mixin using `validator` library https://github.com/HabitRPG/habitica/pull/10757#commitcomment-31038859
- [x] Remove unnecessary CSS prefixes https://github.com/HabitRPG/habitica/pull/10757#commitcomment-31038852
- [x] Allow length-1 usernames https://github.com/HabitRPG/habitica/pull/10757#pullrequestreview-166752463
- [x] Try another modal stack fix, sigh
- [x] Only import relevant methods from `validator`

SabreCat:
**NOT FOR PRODUCTION AT THIS TIME** - This is meant to go live when we are ready to require usernames for all users.

Fixes #10697 
Fixes https://github.com/HabitRPG/habitica-private/issues/12
Fixes https://github.com/HabitRPG/habitica-private/issues/17
Fixes https://github.com/HabitRPG/habitica-private/issues/18

## Username Verification Modals

* Adds a modal that appears automatically if the user has not yet verified their username. The modal gives the user the opportunity to change their display name and/or username and cannot be dismissed except by providing valid input for both fields.
* Adjusts z-indexing of modals so they don't fall behind the "Resting in the Inn" banner. Again. Hopefully.
* Changes the first page of the Justin onboarding flow to give the user a chance to review and update their display name and username.

## Name and Email Verification

* Adds an API route `verify-display-name`, based on `verify-username`, so we can check display names for slurs and length inline.
* Cleans out a few unused name-related strings.
* Fixes an error during email validation in front page login/register flows.
* Replaces in-component regex on emails with the `isEmail` method from `validator`.

## Chat and Private Message Updates

* Cleans up a bunch of styling in Guild/Tavern/Party chat to better fit designs.
* Adds username to the chat message model.
* Shows @username at the top of chat messages, when present.
* Shows both display name (fka profile name) and @username in autocomplete searches, and allows searching on either name.
* Limits autocomplete searches to top 5 matches.
* Adds mouseover highlighting to autocomplete options.
* Adds highlight to @mentions in chat messages.
* Shows @username in inbox conversation list, and implements the full intended styling for profile names there.
* Shows @username in party header, including mouseover for members.
* Shows @username in member profile modal.

## Group Invitation Updates

* Allows users to invite others to groups via username, based on @TheHollidayInn's work.
* Adds an API route `GET /members/username/:username` to allow lookup of public user data via username. It returns not-found if the target username is unverified.


SabreCat:
@paglias Addressed comments via https://github.com/HabitRPG/habitica/pull/10784/commits/5299c8d406ddd7efed355f1a79088f7453a961c0 - now that we're using `validator`, what do you think should be in a mixin? Bundle up the various email/username/etc. validations I guess?

paglias:
@SabreCat I had some trouble moving the functions to a mixin because there were some slight changes between them but if we manage to it would be great

Just one thing about validator, since it's quite big can we load just the necessary functions? So `import validator/lib/isEmail`

website/server/controllers/api-v3/members.js
```diff
@@ -119,6 +119,33 @@ api.getMember = {
...
+  async handler (req, res) {
+    req.checkParams('username', res.t('invalidReqParams')).notEmpty();
+
+    let validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    let username = req.params.username.toLowerCase();
+
+    let member = await User
+      .findOne({'auth.local.lowerCaseUsername': username})
```
paglias: are lowercase usernames unique though?

SabreCat: oh, blurgh, you're right. I suppose this could be set up to return multiple...

PR #10793, opened 2018-10-29
Fixes typo

scheras:
### Changes
zapoměl -> zapomněl


majestic-owl448:
Thank you for wishing to make Habitica better by fixing this small things!
The translation work is done on transifex so you would need to make that correction there -> https://www.transifex.com/habitrpg/habitrpg/

paglias:
@scheras thanks for the fix but @Ieahleen is right, you can find more info at https://github.com/HabitRPG/habitica/blob/develop/website/common/locales/README.md

PR #10808, opened 2018-11-03
 fix: special are not "AllowedToFeed"

### LINKED ISSUE #10801: Special Pets should not have feed bars
_Currently this bug exists only in staging and delta, not production. If possible, it would be nice to have it fixed before it goes into production because it's the sort of thing users will notice and we'd have to handle reports. If it's not possible to fix by then, then when the next deployment from staging to production is made, the Socialites should be told that it's a known error and it doesn't need to be reported when anyone notices it. FYI @SabreCat since you'll likely be doing that deployment. :)_

Special Pets cannot be fed but some of them are displaying feed bars:

image

The reason that some do not display feed bars is that matching Special Mounts exist, but that should be irrelevant for Special Pets - none of them should look feedable.

Here's how it should look (and how it currently does look in production):

image

negue:
fixes #10801 caused by #10768

before #10768 the feeding bar was filtered out for special pets, after the refactor this was removed

### DIFF
--- a/website/client/libs/createAnimal.js
+++ b/website/client/libs/createAnimal.js
@@ -27,7 +27,7 @@ export function isAllowedToFeed (animal, userItems) {
     !isOwned('mount', animal, userItems);
 }
 
-export function createAnimal (egg, potion, type, content, userItems) {
+export function createAnimal (egg, potion, type, _content, userItems) {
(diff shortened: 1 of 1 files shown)


paglias:
@negue thanks, can you fix the lint issue?

Issue #10753, opened 2018-10-11
Very large Guild member counts overflow the badge


SabreCat:
[//]: # (Before logging this issue, please post to the Report a Bug guild from the Habitica website's Help menu. Most bugs can be handled quickly there. If a GitHub issue is needed, you will be advised of that by a moderator or staff member -- a player with a dark blue or purple name. It is recommended that you don't create a new issue unless advised to.)

[//]: # (Bugs in the mobile apps can also be reported there.)

[//]: # (If you have a feature request, use "Help > Request a Feature", not GitHub or the Report a Bug guild.)

[//]: # (For more guidelines see https://github.com/HabitRPG/habitica/issues/2760)

[//]: # (Fill out relevant information - UUID is found from the Habitia website at User Icon > Settings > API)

### Description
[//]: # (Describe bug in detail here. Include screenshots if helpful.)

When a Guild reaches something like 100,000 members, it stops displaying nicely in its member count badge:

https://i.snag.gy/8ncewM.jpg

and in the Guild info itself:

https://i.snag.gy/cm7TCb.jpg

We should probably round these off to remove the decimals, or expand the badge area to fit the count.

Dimini:
Hello, I'd like to work on this issue :)

SabreCat:
@Dimini OK, go for it, with our thanks! Let us know here or in the Aspiring Blacksmiths Guild if you need a hand with anything.

Dimini:
Issue reproduced.

**Count display in the member count badge overflows** -  reproduced after I changed the Guild Member count to 100 010 and more.
**Count display in the Guild info overflows**- reproduced after I changed desktop resolution to value below 1024x768 or after I reduced the size of the window + I changed the Guild Member count to 1010 and more.

The necessary ingredient is to have more than 1k members and the **second to last digit** is not 0.

Thus, 1100 is displayed as 1.1k which is ok, 1101 as 1.1k as well which is ok, but 1010 as 1.01k which is not ok.

I'm going to reduce the number of decimal digits to 1 first and test the results.

Dimini:
**Reducing the amount of decimal numbers to 1** resolved the first issue - **Count display in the member count badge overflows**.

With the same solution **the second issue - **Count display in the Guild info overflows** - is resolved for resolution 1024x768**.

For 800x600 however, the issue still persists. Removing the decimals completely would result in slight overflow for 800x600 resolution.

[...]

PR #10812, opened 2018-11-05
"Very large Guild member counts overflow the badge" partial fix for #10753

Dimini:
### Description

> When a Guild reaches something like 100,000 members, it stops displaying nicely in its member count badge

Issue reproduced.

Count display in the member count badge overflows - reproduced after I changed the Guild Member count to 100 010 and more.
Count display in the Guild info overflows- reproduced after I changed desktop resolution to value below 1024x768 or after I reduced the size of the window + I changed the Guild Member count to 1010 and more.

The necessary ingredient is to have more than 1k members and the second to last digit is not 0.

Thus, 1100 is displayed as 1.1k which is ok, 1101 as 1.1k as well which is ok, but 1010 as 1.01k which is not ok since it overflows.

### Changes
partial fix for #10753 

Reducing the amount of decimal numbers to 1 resolved the first issue - Count display in the member count badge overflows.

With the same solution the second issue - Count display in the Guild info overflows - is resolved for resolution 1024x768.

For 800x600 however, the issue still persists. Removing the decimals completely would result in slight overflow for 800x600 resolution.

Moreover, I used floor function instead of the round function to avoid rounding up.
Eg. If the guild has 4567 members, it shouldn't show 4.6k. It will show 4.5k (which makes more sense).

### Another solution
Regarding the second solution proposed by @SabreCat , the badge would have to be increased in size by at least 15px for the text appear inside the text area. The badge however shouldn't change size, a new badge image with wider text area could be created for Guilds with more than 100k members.

The main root cause of the second issue - Count display in the Guild info overflows - is that while the graphical components are browser responsive, the text is not. This way the text area is shrinking with the decrease of resolution and the text overflows.

### DIFF
--- a/website/client/mixins/groupsUtilities.js
+++ b/website/client/mixins/groupsUtilities.js
@@ -4,7 +4,7 @@ export default {
   filters: {
     // https://stackoverflow.com/questions/2685911/is-there-a-way-to-round-numbers-into-a-reader-friendly-format-e-g-1-1k
     abbrNum: (number) => {
-      let decPlaces = 2;
+      let decPlaces = 1;
       decPlaces = Math.pow(10, decPlaces);
 
       let abbrev = ['k', 'm', 'b', 't'];
@@ -12,7 +12,7 @@ export default {
         let size = Math.pow(10, (i + 1) * 3);
 
         if (size <= number) {
-          number = Math.round(number * decPlaces / size) / decPlaces;
+          number = Math.floor(number * decPlaces / size) / decPlaces;
 
           if (number === 1000 && i < abbrev.length - 1) {
             number = 1;


paglias:
@Dimini thanks for the PR! I think fixing it for resolutions down to 1024x768 is fine and covers most use cases.

I'm merging the PR and I'll be happy to award you a contributor tier if you have an Habitica UUID.

Dimini:
Thanks @paglias, I'm happy I could help :)

My UUID: 1d3232a3-6424-47e7-87fe-2ae78021529e

Looking forward to future cooperation :)

paglias:
tier awarded :)

PR #10818, opened 2018-11-06
prevent buying market gear if class doesn't match

### LINKED ISSUE #10183: should not be able to buy equipment from a class that you are not currently playing

negue:
continuation of #10657 + added test

fixes #10183 
closes #10657

### DIFF
--- a/test/api/v3/integration/user/buy/POST-user_buy_gear.test.js
+++ b/test/api/v3/integration/user/buy/POST-user_buy_gear.test.js
@@ -53,4 +53,15 @@ describe('POST /user/buy-gear/:key', () => {
         message: 'You need to purchase a lower level gear before this one.',
       });
   });
+
+  it('returns an error if tries to buy gear from a different class', async () => {
+    let key = 'armor_rogue_1';
+
+    return expect(user.post(`/user/buy-gear/${key}`))
+      .to.eventually.be.rejected.and.eql({
+        code: 401,
+        error: 'NotAuthorized',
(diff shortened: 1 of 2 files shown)


paglias:
@negue thanks, in the description you mention a test but it looks like it might have been lost because I can't find it in the changes

negue:
I had it ready, but I didn't pushed it ... 😅

PR #10826, opened 2018-11-09
Empty the armory feature

### LINKED ISSUE #7991: Empty the Armory : create feature and achievement
Now that the [Orb of Rebirth no longer removes equipment](https://github.com/HabitRPG/habitrpg/issues/7917), we need a similar feature that does remove gear (but does nothing else).

Features:
- [ ] The name for it is tentatively *"Empty the Armory"* but other suggestions are welcome.
- [ ] It will cost 2 gems.
- [ ] For players who are at or above level 100, it will be free.
- [ ] It will not remove gear that costs zero gold (because it's trivial to "buy" back so there's no point removing it).
- [ ] It must not reintroduce this bug: [Orb of Rebirth causes animal ears in 'Customize Avatar' interface to need to be repurchased](https://github.com/HabitRPG/habitrpg/issues/7662)
- [ ] It must have its own "Empty the Armory" achievement, which increases by one each time the emptying is done (works a similar way as `achievements.rebirths`).
- [ ] **For discussion**: We need to decide when the Empty the Armory feature becomes unlocked. At level 50? After receiving one Ultimate Gear achievement? **Suggestions welcome.**
- [ ] **For discussion**: We need to decide if the achievement increases by one only if you use the Empty the Armory feature at the same level or higher than previously (as is done for the Rebirth achievement). **Suggestions welcome.**
- [ ] [Artwork is being created for the achievement badge](https://trello.com/c/xfXldYs1/11-badges-open). Until it's ready, I suggest creating a placeholder image using a copy of the Orb's badge.
Disagreements welcome.
- [ ] I suggest that the achievement should not cause an increase in the player's drop chance (an increase is caused by Rebirth), because drop chances already have quite a few boosting events for long-term users. Disagreements welcome.
- [ ] The Seasonal Shop displays the `seasonalShopRebirth` message ("If you've used the Orb of Rebirth, you can repurchase this equipment in the Rewards Column...."). That message should be changed to mention the Empty the Armory feature as well as the Orb (rebirth should remain in that message for a while, for the benefit of players who rebirthed before it was nerfed).
- [ ] `settings.json` contains `resetText2` which ends with "You might prefer to use an Orb of Rebirth instead, which is a much safer option and which will preserve your tasks.". That should be changed to say "... and which will preserve your tasks and your equipment."  We should have made that change in #7917 but we overlooked it.

Much of the code for this new feature will be similar to code for the original Rebirth feature, so here's some links to **some** of the relevant files from the production website release from before Rebirth was nerfed:
- https://github.com/HabitRPG/habitrpg/blob/31a1a14baecb464e66c94d259cbac9cdcbbc7e89/website/views/shared/modals/rebirth.jade
- https://github.com/HabitRPG/habitrpg/blob/31a1a14baecb464e66c94d259cbac9cdcbbc7e89/common/locales/en/rebirth.json
- https://github.com/HabitRPG/habitrpg/blob/31a1a14baecb464e66c94d259cbac9cdcbbc7e89/common/script/ops/rebirth.js
- https://github.com/HabitRPG/habitrpg/blob/31a1a14baecb464e66c94d259cbac9cdcbbc7e89/test/api/v3/integration/user/POST-user_rebirth.test.js
- https://github.com/HabitRPG/habitrpg/blob/31a1a14baecb464e66c94d259cbac9cdcbbc7e89/test/common/ops/rebirth.js

**Actions for staff/admins after a contributor has created this feature:**
- [ ] Write a Bailey message to introduce the feature and to say that all players who obtained Orb of Rebirth achievement(s) before 11 September (when the Orb was nerfed) have been given the same number of Empty the Armory achievements (in addition to keeping their Rebirth achievements).
- [ ] Give the Empty the Armory achievement to those players. Alys has a list.
- [ ] Ask the wiki editors to adjust the Start Over Options page, as well as adding a new page for this new feature.

_I'm marking this issue as both help welcome and for suggestion/discussion, because a coder could start on most of it while we're deciding on the final details of some of the points listed here._

Arsnael:

Implements #7991 (Closes #7991)

### Changes
Implemented the 'Empty the armory' feature. I keep the WIP for now as I still need to test it a bit more and write more tests, but as of now it looks quite ok. So you can start checking and also maybe coming up with better text messages for this feature than mine. Here is a list : 

* similar to 'Orb of rebirth', can buy at lvl50 for 2 gems and free at lvl 100,
* resets all equipment,
* has its own achievement,
* when equipment is reset, can buy it back at the market for gold (class none for most of items, normal behavior I guess, I didn't change anything special here),
* no bug with animal ears, thanks to https://github.com/HabitRPG/habitica/issues/10668 (one less issue to tackle^^),
* rebased with recent code (2 days ago).

What's left : 
* writing more tests,
* update messages? My guess is I didn't put much thoughts in it yet, and this is something you probably want to change/discuss as well,
* should we add a loading screen tip for this feature too?
* I noticed some v4 implementation in the code lately server side. Should I take it into consideration? Didn't look into it to be honest.

If you notice anything else, please report ! And sorry for the delay, been busy lately :)

Some quick screenshots : 
emptyarmory1
emptyarmory2
emptyarmory3
emptyarmory4
emptyarmory5

----


migrations/empty_armory.js
```diff
@@ -0,0 +1,90 @@
+let migrationName = 'empty_armory.js';
+let authorName = 'Arsnael'; // in case script author needs to know when their ...
+let authorUuid = 'bbffb2f4-9bf8-46c4-b749-523e2ca93ef6'; // ... own data is done
+
+/*
+ * set the newStuff flag in all user accounts so they see a Bailey message
```
Alys: This comment should be changed.

website/common/locales/en/emptyArmory.json
```diff
@@ -0,0 +1,22 @@
+{
+    "emptyArmoryNew": "Empty the armory: New Challenge Available!",
```
Alys: "Empty the Armory" should be written with capital E and A everywhere that it occurs, since it's the title of the feature.

website/common/locales/en/emptyArmory.json
```diff
@@ -0,0 +1,22 @@
...
+    "emptyArmoryUnlock": "You've unlocked 'Empty the Armory'! This special Market item allows you to reset your equipment. Use it to breathe new challenge into Habitica if you feel you've achieved it all, or if you just like running around naked!",
```
Alys: We can't use "running around naked" because it's something of an adult topic. :)

website/common/locales/en/emptyArmory.json
```diff
@@ -0,0 +1,22 @@
...
+    "emptyArmoryText": "Began <%= emptyArmorys %> New Challenges",
+    "emptyArmoryKey": "Used an armory key to reset equipment after attaining Level <%= level %>.",
+    "emptyArmoryKey100": "Used an armory key to start over after attaining Level 100 or higher.",
+    "emptyArmoryKeyNoLevel": "Used an armory key to start over.",
+    "emptyArmoryPop": "Instantly reset your equipment!",
+    "emptyArmoryName": "Armory key",
+    "reset": "Reset Equipment, max level <%= reLevel %>",
+    "confirmReset": "Are you sure?",
+    "emptyArmoryComplete": "You have reset your equipment!"
+}
```
Alys, line 14: I think not all of the strings above are used in your new code. You can delete any that aren't used. 

I'm guessing the unused ones are because they've been copied from the rebirth.json file. There's probably a bunch of strings in that file that aren't used anywhere. That's because changes have been over time without a full clean-up of the unused strings. We don't need to exactly duplicate what's been done for rebirth in cases like that.

It's probably unnecessary to have a dedicated locale file for this (although I realise rebirth has one). There's extra work needed related to translations when a new locale file is created, so avoiding that is best when it's not essential. My personal feeling is that the new strings could go into the achievements.json file but if any of the staff say otherwise, follow their advice not mine.

Once any unnecessary new strings have been removed, we can work on changing the text if needed.

website/common/locales/en/rebirth.json
```diff
@@ -21,7 +21,7 @@
...
-    "rebirthPop": "Instantly restart your character as a Level 1 Warrior while retaining achievements, collectibles, and equipment. Your tasks and their history will remain but they will be reset to yellow. Your streaks will be removed except from challenge tasks. Your Gold, Experience, Mana, and the effects of all Skills will be removed. All of this will take effect immediately. For more information, see the wiki's <a href='http://habitica.wikia.com/wiki/Orb_of_Rebirth' target='_blank'>Orb of Rebirth</a> page.",
+    "rebirthPop": "Instantly restart your character as a Level 1 Warrior while retaining achievements, collectibles, and equipment. Your tasks and their history will remain but they will be reset to yellow. Your streaks will be removed except from challenge tasks. Your Gold, Experience, Mana, and the effects of all Skills will be removed. All of this will take effect immediately. For more information, see the wiki's <a href='http://habitica.wikia.com/wiki/Orb_of_Rebirth' target='_blank'>Orb of Rebirth</a> page. <b>If you want to reset your equipment as well, you better use the 'Armory key' item first!</b>",
```
Alys: The new text at the end of this line can be:
`If you want to reset your equipment as well, use \"Empty the Armory\" first!`

Alys:
Thanks for this! It's quite a large piece of work so it's nice to see it being worked on! No worries about the delay. :)

The feature shouldn't remove all equipment; anything that costs zero Gold should remain (see the issue for a bit more detail). The resetGear function is for use by the "Reset" option under Settings, which is a different case because it's designed for new players who've been experimenting with Habitica and now want to make a fresh start - in that case it makes sense for all equipment to be removed.

I'm about to add a few comments to the code. This isn't a full code review so it's likely that the staff will add more later. Also, I haven't answered your questions but the staff will do that.

There's a few places in the new code where the Gem cost is referred to (2 or 0.5 depending on whether it's number of Gems or fraction of `balance`). That might be better handled by a constant with a value of 2, and then when you need to refer to `balance`, you can use `[name of constant] / 4`

Arsnael:
Thanks for your comments and your quick review @Alys , I will look into it as soon as I can :)

website/common/locales/en/emptyArmory.json
Arsnael: I know I know. Just sounded funny at the moment, thought would take it out later... but forgot, my bad :)

website/common/locales/en/emptyArmory.json
Arsnael: Actually you are quite right... That was the result of a cheap copy/paste at the beginning, and I forgot to look more back into it at the end. It seems nearly half of those strings are not being used actually, so yeah might be better to put it in achievements.json file. Thanks :)

EDIT : or maybe not... actually it's more than just an achievement, you have the item in the shop and the feature, etc... Will see what other devs think on that matter :)

Arsnael:
Regarding the not removing zero gold items... Actually when I looked into the code from back when rebirth used to remove items, it seems was using resetGear function to me. Which is why I reused it. But will look more into it. After reading again the discussion on the task page, it looks like death is not removing zero gold items. So should be something similar right? Will study that piece of code :)

website/common/locales/en/emptyArmory.json
Alys: It did sound funny. It made me smile. :)

Alys:
> Regarding the not removing zero gold items... Actually when I looked into the code from back when rebirth used to remove items, it seems was using resetGear function to me. Which is why I reused it. 

I'm sure Rebirth did originally remove all equipment. I had thought that we'd changed it to remove only non-zero-gold equipment before we changed it to not remove any gear but I might well be misremembering!

But regardless, as @Ieahleen pointed out in the issue, there's a case to be made for removing zero-gold items so we'll think about that a bit more.

> it looks like death is not removing zero gold items. So should be something similar right? Will study that piece of code :)

Yes, that's correct. If we do end up not removing zero-gold items, then the death code could be a good example.

Arsnael:
@Alys : after testing more deeply, I realized that actually using the `resetGear` function doesn't remove the 0 gold items. I did some tests locally and I added a unit test for that in the code too (along with other tests). So my guess is it's ok. I think it's ready for review now so will just wait for a dev to check it and telling me if we need to change anything or not. Enjoy !

Alys:
Ah, okay! That sounds fine to me. I still think [@Ieahleen's point](https://github.com/HabitRPG/habitica/issues/7991#issuecomment-437887272) will be relevant for at least some players but I suggest that for simplicity and speed, we go with the default `resetGear` behaviour. We could consider changes later based on feedback from the players.

Alys:
FYI staff: before this can be released, this new bug in the equipment purchase code needs to be fixed: https://github.com/HabitRPG/habitica/issues/10849

(We still need to look at the new strings too and I'm hoping to do a first pass over that today.)

website/common/locales/en/emptyArmory.json
Alys: I'm wondering if the word "challenge" will be confusing to some players, although I can't think of other phrasing that could be used. Rebirth uses "adventure" but I think you're right to not copy that here because it would be nice to have different phrasing between the two (and buying gear is less adventurous than starting over from scratch!)

Apart from that I have no further comments about these strings. Thanks for the changes you've made!

I haven't reviewed the code or done any testing.

citrusella: New... closet? Trunk? Armory? Load of clothes? Fashion line?

I don't know how helpful this is but I wanted to try. XP

(The other alternative I have in my head is to go whole hog and phrase the description like the user *lost* something, like pairing "Oh dear! You've broken the lock to your [equipment place name]! Nothing to do now but take this fresh key and start a new one! This is [#] for you, and you've reached [level]." or "Someone's broken into your equipment stash and taken it all!" or something. XD ...Don't listen to me, I'm just having fun here. XP)

Arsnael:
The issue you are mentioning @Alys made me realize that when you have been changing classes a few times, you cannot rebuy items you had from a previous class.

For example I was a rogue, I bought all the base gear from the rogue in the shop, then I do a rebirth as a healer. Then I do a reset equipment. I cannot buy back the rogue gear I had previously. Is that normal behavior or not? (not sure) seems those are being pinned as well, but cannot buy them... What do you think on this?

citrusella:
I'm not Alys, but: Not being able to buy back gear when it's for a class you currently aren't is the expected behavior AFAIK, even back in old-style rebirth. The pinning sounds "not right", but I don't know that for sure. (Back when rebirth did the armory-emptying, it was before pinning was a thing. Equipment only showed up for purchase in the rewards column, and you couldn't see equipment for other classes than the one you were at that moment.)

Alys:
Pinning when you lose something due to death is definitely deliberate, and once something is pinned, I believe there's no code to unpin it automatically, so it's expected that it will still be pinned after you've changed class. Unpinning manually is easy so I don't think that's a problem, especially since some players might want it to remain pinned (e.g., as something to aim for when they change class again).

Arsnael:
@citrusella , @Alys : I had a small doubt so I  just wanted to be sure. Thanks both of you for the clarification !

paglias:
I haven't followed the discussion in the PR much, is this ready for review or there's something still missing?

Also there seem to be some conflicts!

Arsnael:
@paglias yes it is ready for review :) 

Ah conflicts with the sprites again... always new sprites being added regularly, hard to keep up with it :D will try to check when I have time and readjust the sprites correctly thanks for reporting it !

paglias:
Ah sorry @Arsnael I didn't notice the conflict was with the sprites, in that case don't worry about fixing it :)

SabreCat:
I'm just going to make a regular practice of removing compiled spritesheets from PRs when I encounter them. Hope that's OK @Arsnael! Doing some review now.

SabreCat:
This is working quite nicely! The only functional thing I've encountered so far is that it doesn't reset your pinned items; if you've bought a couple tiers of class equipment, then when you Empty the Armory it (1) leaves items pinned that you can no longer directly buy, and (2) the starter-tier items for your class are not pinned to Rewards like you might expect.

website/client/components/achievements/emptyArmory.vue
```diff
@@ -0,0 +1,48 @@
+<template lang="pug">
+  b-modal#empty-armory(:title="$t('modalAchievement')", size='md', :hide-footer="true", @hidden="reloadPage()")
+    .modal-body
+      .col-12
+        achievement-avatar.avatar
+      .col-6.offset-3.text-center
+        div(v-if='user.achievements.emptyArmoryLevel < 100')
+          | {{ $t('emptyArmoryAchievement', {number: user.achievements.emptyArmorys, level: user.achievements.emptyArmoryLevel}) }}
+        div(v-if='user.achievements.emptyArmoryLevel >= 100')
+          | {{ $t('emptyArmoryAchievement100', {number: user.achievements.emptyArmorys}) }}
```
SabreCat, line 10: I get that this is modeled after the Orb of Rebirth, but given how little association there is between equipment and level, it seems really strange to track max level the same way. It might be more fun to track just how many pieces of equipment the user has dumped back into the void...

Alys:
> if you've bought a couple tiers of class equipment, then when you Empty the Armory it (1) leaves items pinned that you can no longer directly buy, and (2) the starter-tier items for your class are not pinned to Rewards like you might expect.

Ooo, I wonder if that's a bug in Reset as well.
If not, Reset would have code to steal for that.
If it is, we'd want both fixed the same way I imagine.

website/client/components/achievements/emptyArmoryEnabled.vue
```diff
@@ -0,0 +1,33 @@
+<template lang="pug">
+  b-modal#empty-armory-enabled(:title="$t('emptyArmoryNew')", size='md', :hide-footer="true")
+    .modal-body
+      .col-12
+        .armory_key
+        p
+          span {{ $t('emptyArmoryUnlock') }}
```
SabreCat, line 7: This same modal seems to appear when the user reaches level 50 and when they reach level 100. Should the message perhaps be a little different in the latter case, so it's clear the feature is now free to use?

SabreCat:
Would it be possible to force a sync after the operation completes? At the moment, your avatar keeps its currently equipped gear after you've given it up, until the next time you click sync or the like.

Arsnael:
Hi @SabreCat ! Thank you for the review and your comments !

> if you've bought a couple tiers of class equipment, then when you Empty the Armory it (1) leaves items pinned that you can no longer directly buy, and (2) the starter-tier items for your class are not pinned to Rewards like you might expect.

I noticed the (1) and we had a discussion about that a bit above with @Alys and @citrusella . I didn't pay attention enough to (2) so I will check. I will check the rest of your comments when I have time as well :)

SabreCat:
Not being able to rebuy items from other classes is intentional. I believe we've made it so you can rebuy _seasonal_ equipment from other classes, but the basic gear progression should stay class-bound. That's a bit of an aside in any case; I'm talking about higher tier items from one's own class, in this scenario. E.g. I'm a Warrior, I've bought several items and the Ruby Helm is currently showing in my Rewards column. I empty the Armory, setting me back to 0. I need to buy the first few Helms before I can buy the Ruby one, but it's still pinned.

citrusella:
> I believe we've made it so you can rebuy seasonal equipment from other classes

Random question: *When* would this have happened? (Just asking because it's news to me; last I knew that wasn't the case.)

SabreCat:
@citrusella I could be mistaken, sorry! We've changed things around so many times with the gear logic I probably got confused. @_@

Arsnael:
@SabreCat Ah indeed I didn't test that case with the higher tier items. I will see if I can pin them down.

I think I saw this as well with the seasonal equipment but my memory could be wrong too... Will try to recheck that too !

website/client/components/achievements/emptyArmoryEnabled.vue
Arsnael: No the modal doesn't reappear in level 100 (the rebirth doesn't too, so I just applied the same here). Should that change?

Arsnael:
> Would it be possible to force a sync after the operation completes? At the moment, your avatar keeps its currently equipped gear after you've given it up, until the next time you click sync or the like.

Well when I was developing the feature, it seems some stuff changed with notifications and I had to rebase and adapt. It seems to be that way for rebirth and a few other events too. You need to click on the notification on the right which opens the modal and when you close it, it reloads automatically. Or maybe I didn't get everything when rebasing...

Alys:
@Arsnael I'm sorry we've been silent about this for two months!  There's been testing done and some comments were made in a private channel. I've just today collated them all and I've added them to the issue to keep them in the same place as the original requirements. See https://github.com/HabitRPG/habitica/issues/7991#issuecomment-463121169

Arsnael:
@Alys : no problem, myself I don't have a lot of time anymore these days anyway. I will try to review and think about your remarks and requirements when I can :)

SabreCat:
If we end up moving ahead with this, it'll be a project for the Habitica team to tackle! I've moved the associated issue ticket to our internal project board.

PR #10832, opened 2018-11-10
WIP: Improve User model performances

paglias:
We found out that the schema definition for user documents has become so big that's creating perf problems.

This PR tries to improve the situation by not defining every single items the user can have as a field in mongoose and instead use several mixed containers.

From some local testing this seems to improve the average loading time for user documents from ~55ms to ~9ms (vs ~2ms when using mongoose `.lean()` option)

It's a WIP and shouldn't be merged yet

Ready for testing


paglias:
The following screenshots are a list of all places affected by this change

SabreCat:
Ooh, good to see you've gotten a chance to circle back to this. How close is it getting?

paglias:
Fixed issue with the validation in the hall controller

paglias:
Added tests

paglias:
Should be ready for testing!

PR #10833, opened 2018-11-10
fixes issue #10758 ("Market incorrect positioning of items: single item on previous line")

### LINKED ISSUE #10758: Incorrect item wrapping on Market pages
Il Mago (c073342f-4a65-4a13-9ffd-9e7fa5410d6b):
"In the Market sometimes there is a strange positioning of the items, for example in this case when there should be a single item on a new line it is instead positioned as last item of the previous line, and to the right of everything else:"
scr

shanaqui:
"Before I add that to Github, do you have any idea how to get that to happen? Resizing windows, maybe?"

Il Mago:
"I have no idea, it is the first time I noticed it, but it is also the first time in which I could have had a row with only one item inside. I will try to buy some things to see if in an other category i can have that one-row-item to display correctly or if it happens again (it may be connected to the one-row-item that doesn't want to display)"
 
Il Mago:
"I have more details on the issue with the positioning of items in the Market
 I hope this is explanatory enough: [screenshot below]
If not I will try to maybe find a screen recorder or something"
scr

Phelilix:

Fixes #10758

### Changes
I added a css rule in items.scss.
The rule changes margin in every div that is a child of items that is a child of row-items.
The previous margin remains but is overwritten by this more specific rule.

I apparently also change white space on line 27. But I currently don't know how to easily change that.

Before:
screenshot 2018-11-10 11 32 50

After:
screenshot 2018-11-10 11 28 23

----
UUID: c494f93f-7b9a-4f4a-beb3-b6af7a20712d

### DIFF
--- a/website/client/assets/scss/item.scss
+++ b/website/client/assets/scss/item.scss
@@ -9,6 +9,10 @@
   margin-right: 0px;
 }
 
+.item-rows > .items > div {
+  margin: 6px 12px 6px 12px;
+}
+
 .item-wrapper {
   position: relative;
   display: inline-block;
@@ -24,7 +28,7 @@
   @media only screen and (min-width: 1440px){
     margin-right: 1.71em;
   }
-  
+
   // Desktop L (1280)
   @media only screen and (min-width: 1280px) and (max-width: 1439px) {
     margin-right: 0.43em;


paglias:
@Phelilix thanks for the PR! 

I'm wondering why the old rules are not deleted since this new one is more specific, in particular this one https://github.com/HabitRPG/habitica/pull/10833/files#diff-d5ff7a0f110e5e65926c99fe07124be0R3

Phelilix:
> 
> 
> @Phelilix thanks for the PR!
> 
> I'm wondering why the old rules are not deleted since this new one is more specific, in particular this one https://github.com/HabitRPG/habitica/pull/10833/files#diff-d5ff7a0f110e5e65926c99fe07124be0R3

Because ```items``` is used on multiple places. At least in the skills bar and the quest-shop.
If I were to remove/modify their margin, the layout would have shifted.

I had to modify the margin of ```items``` when they have more than one row. Or else this problem couldn't have been fixed. 
```item-rows``` was only mentioned in itemRows.vue, and it really looks like it is meant to be used when ```items``` are going to be more than one row.

paglias:
@Phelilix thanks, I've tested  this and it looks like there's an issue in some cases, see the screenshot
schermata 2018-11-18 alle 19 43 09

See the line of times at the top is slightly moved to the right after this change, tested with a new account. It doesn't happen currently

paglias:
Hi @Phelilix ! Do you have any update on the status of this PR?

paglias:
Hi @Phelilix , I'm going to close the PR but if you want to work on this again just leave a comment

PR #10844, opened 2018-11-16
Implement URL handling for profile modal

phillipthelen:
This PR adds some URL handling to the profile modal.

Details:

- When you open a profile from chats, the header or other social spaces, the url is updated to be `/profile/USERID`.
- Closing/tapping outside of the modal correctly sets the URL back to what it was before.
- You can now also use the middle mouse button to open a new tab with the users profile (or whatever interaction method makes the browser do that)
- If the url is opened in a new tab, the profile is not rendered in a modal, but instead displayed on a blank page.
- The url is also updated when you open the profile/stats/achievement page from the user menu in the top right, meaning those pages are linkable now too. ( to `/user/profile`, `/user/stats` or `/user/achievements`)
- Since the UserID is now included in the URL when 

I also added some minor design tweaks to the profile page, to bring it closer to the original designs, since I had to make changes to the html anyway.


paglias:
Working well! The only thing that might be worth fixing is that clicking the back button in the browser doesn't do anything if you have a profile open

phillipthelen:
I also noticed that after posting, but haven't been able to quite figure out how to fix that. vue router seems to ignore this back action, because the browser history is out of sync with the router history (which does then fix itself if you go back again, so that behaviour does seem pretty ok). So I'll probably have to add a listener to a history event and figure out if the modal needs to be closed or not.

paglias:
@phillipthelen  is this ready or do you want to make other changes?

phillipthelen:
I still want to try to improve the handling of the back button. I'll work on that in the next days.

phillipthelen:
I couldn't find a way to make it actually close the modal when using the back button, but I was able to make it not break. If you now use the back button, the url is updated but the modal not closed. If you then close the modal normally, the url isn't updated again. It's not ideal, but since most users probably won't use the back button in this case anyway, it should be ok.

website/client/components/chat/chatCard.vue
```diff
@@ -234,18 +189,6 @@ export default {
       }
       return likeCount;
     },
-    tierIcon () {
```
paglias, line 238: (not related to this PR in particular, but since in another one this function is used as well, can it be moved to a mixin? I think the existing mixins/styleHelper would be perfect)

website/client/components/userLink.vue
```diff
@@ -1,21 +1,114 @@
...
+  :class='levelStyle()',
+  v-b-tooltip.hover.top="tierTitle",
+  v-if='displayName')
+    | {{displayName}}
+    .svg-icon(v-html="tierIcon()")
 </template>
 
+<style scoped lang="scss">
+  @import '~client/assets/scss/colors.scss';
+  @import '~client/assets/scss/tiers.scss';
```
paglias: like in the other PR this should be included only once in scc/index

paglias (changes requested):

paglias:
@phillipthelen just tested this one again and I feel like the handling for the back button is okay. Not sure if it's wanted or not but the spacing between the avatar and the progress bars seems a bit weird and it breaks on smaller screens

schermata 2018-12-04 alle 10 14 37
schermata 2018-12-04 alle 10 14 42

Other than this and the conflicts it LGTM

phillipthelen:
PR Should be ready now.

PR #10845, opened 2018-11-16
Attach client to chat messages

phillipthelen:
Small change that attaches the content of the x-client header to a chat message. The content of that field is also displayed to all users with a contributor level higher than 3. Reasoning being, that this will hopefully simplify user support a little bit, by allowing socialites and mods to see what client the user is using. I set it to only show to users above a certain tier, since this information isn't relevant for most users and only adds more noise/clutter to chat.

screenshot 2018-11-16 at 19 00 05

### DIFF
--- a/website/server/controllers/api-v3/chat.js
+++ b/website/server/controllers/api-v3/chat.js
@@ -175,7 +175,10 @@ api.postChat = {
       throw new NotAuthorized(res.t('messageGroupChatSpam'));
     }
 
-    const client = req.headers['x-client'].replace('habitica-', '');
+    let client = req.headers['x-client'] || '3rd Party';
+    if (client) {
+      client = client.replace('habitica-', '');
+    }
     const newChatMessage = group.sendChat(req.body.message, user, null, client);
     let toSave = [newChatMessage.save()];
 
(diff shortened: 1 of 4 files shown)


saraolson:
suggested change to "@Username - a minute ago (web)" /(Android)/(iOS) for space reasons when it eventually gets to mobile :D i think the slightly more technical label feels appropriate since this will only be visible for high tier people for bug purposes.

website/server/controllers/api-v3/chat.js
```diff
@@ -175,7 +175,8 @@ api.postChat = {
       throw new NotAuthorized(res.t('messageGroupChatSpam'));
     }
 
-    const newChatMessage = group.sendChat(req.body.message, user);
+    const client = req.headers['x-client'].replace('habitica-', '');
```
paglias: this will throw an error if the `x-client` header isn't set

paglias (changes requested):
just one thing, I think otherwise it can be merged

website/server/controllers/api-v3/chat.js
paglias: for example if the message is sent from a 3rd party client

phillipthelen: Good point. Originally didn't have the replace part and then forgot to handle that case. I now added something for that and also set a client for when no header was passed.

paglias (approved):

Alys:
Here's a screenshot showing one example of this feature, to help low-level Socialites and Wiki WIzards understand it:
image

Issue #10834, opened 2018-11-11
pet feeding behaviour and display problems


Alys:
In the Report a Bug guild, Praxis (0ce6d43b-efa2-45a3-acaa-59daad7d658d) reported the bugs below. **All words in** normal font **below this line are theirs, not mine; my comments are added** _[in italics in brackets like this]_.
Below their report is a related report from Aerilita (3b724a41-48d3-4048-ae27-70b258cf5a72).

----

Hi, guys! It seems the Stable was recently changed, and glitches were introduced, sometime between November 1st and today.

When I hatched and raised a full set of Ghost pets and mounts on November 1st, the Stable interface was behaving normally, as it has been for some time: It worked smoothly, quickly and functioned perfectly (I use the web interface).

Today, when I hatched a couple of pairs of matching Gator pets and mounts, the following glitches occurred (all of these were new... and I've raised a LOT of pets):

The initial hatching (upon clicking the black pawprint for the first time) worked normally: Pop-up window with purple button to hatch, pet icon appeared in its proper spot in the lineup.

1) But then the feeding progress bar did not advance when the pet was fed. I had to manually re-synch to get it moving, even though the purple re-synching bar DID move across the top of the page each time (I know enough to wait for this) and even though I got the blue "really likes" confirmation message each time. After I manually re-synched (after about 3-4 feedings where feeding the progress bar did NOT move -- this happened on both pets I tested), then the feeding progress bar continued to advance properly in time with confirmation messages until the pet became a mount.

_[I've replicated everything up to here.]_

2) At this point the normal pop-up annoucing that I had successfully raised my pet into a mount appeared, and I clicked on the purple "onward" button. Then I noticed that the system had automatically equipped the PET that I had just raised into a MOUNT onto my Avatar, substituting it for the pet that I previously had equipped (this happened for sure on the second pet that I tested, not sure about the first one) -- even though now all I really had in my stable was the MOUNT of that color.

_[I can't replicate this (the Pet I just fed to a Mount was not equipped automatically) but this might be an inconsistent bug that depends on other factors. As I'll mention below, there seems to be some other inconsistencies with the new bugs. I suggest we fix the ones that can be replicated and then test again for the ones that can't. The initial fixes might solve the other problems anyway.]_

3) The pawprint position of the pet that I had just raised into a mount was still black (as normally expected) but it had become un-clickable -- completely unresponsive.

[...]

PR #10853, opened 2018-11-18
stable-state refresh 

negue:
should fix most of #10834 

the issue that I could reproduce was that, even if the state changed the computed methods didn't get any refreshes (even though `userItems` was used inside the computed prop)


paglias:
thanks, I'm merging this as it fixes the issue although I'm unsure why the computed properties are not working anymore

PR #8962, opened 2017-08-17
Implements Dailies repeat every X days since last completion  (Fixes #6941) [reverted for extra testing]

asifmallik:
fixes https://github.com/HabitRPG/habitica/issues/6941

### Changes

- Implemented repeat after completion for both the new and old UI
- Added tests for the feature in shouldDo.test.js

UUID: bc6a5bb0-b458-4526-b822-a1e63c581ce4

PR #10854, opened 2018-11-19
Implements repeat every X days since last completion - fixes #6941 continues #8962 by asifmallik

### LINKED ISSUE #6941: Repeat Every X Days Since Last Completed

Alys:
fixes https://github.com/HabitRPG/habitica/issues/6941

This is a continuation of https://github.com/HabitRPG/habitica/pull/8962 by asifmallik. It contains asifmallik's code with minor changes made to it to bring it up to date with `develop`

### Changes
- Implemented repeat after completion
- Added tests for the feature in shouldDo.test.js

asifmallik UUID: bc6a5bb0-b458-4526-b822-a1e63c581ce4


Alys:
This has been only briefly tested on my local install. It needs thorough testing. Now that Username testing has been finished on delta, can we put this on it for Socialites to test on?

paglias:
@SabreCat can this go on delta? Either now or after the flag pm testing is done

Alys:
Is still like it to be on delta but if like the testers to be given a
warning 24 hours before it goes there. It has a small chance of breaking
cron so they need to make a choice about risking that. Also the Dailies
with the new feature won't behave correctly on production so they'll need
to know that. I'm happy to write all that myself if you want me to - just
let me know roughly when it would be deployed to delta.

From the point of view of testers its fine to be testing this and PM
flagging together, but I understand that it could make merging harder. I'm
the only person testing PM flagging so it could go on beta or gamma if that
makes it easier.

SabreCat:
@Alys Maybe this and #10870 together, as they're both cron-related things that could use some tester wind-up? 24 hours from now, or 48, just tell me when and I'll send it up.

paglias:
@SabreCat I think the newest changes have not been pushed to delta

Alys:
Back in December, I talked to Sabe about when to deploy in a private chat so as not to spam people with notifications from this PR, and there was a deployment on 12 December (after that latest commit) but it's certainly confusing now that the conversation wasn't recorded here! I'll keep all requests for deployment in PRs from now on and we should add a note in the PRs when each deployment has been done. Thanks paglias for folowing up!

paglias:
Okay @Alys ! Let me know when you think it's received enough testing then

website/server/libs/cron.js
```diff
@@ -314,6 +314,13 @@ export function cron (options = {}) {
     let scheduleMisses = daysMissed;
 
     if (completed) {
+      if (task.history && task.history.length) {
+        const lastHistoryEntry = task.history[task.history.length - 1];
+        task.lastCompleted = moment(lastHistoryEntry.date).toDate();
```
paglias, line 360: why do we need a new field instead of using the value from `task.history[task.history.length - 1] `?

Alys: So a field like `task.lastCompleted` that's updated at the time that the Daily is marked completed by the user?
I did consider doing that but I realised that if the user completes a Daily by mistake then later un-completes it, we'd need to set `lastCompleted` back to its earlier value, which means we would have needed to store the earlier value in yet another field when we updated `lastCompleted` or we'd need to read the earlier value from the history array, and it seemed simpler to just use the history array in cron.

Alys: sorry, ignore that, I think I misread your questions. a real answer will be here soon. :)

Alys: The `task.lastCompleted` code was added by the original contributor and approved in https://github.com/HabitRPG/habitica/pull/8962/files but I can rework it to use the history array directly. It shouldn't be hard. I'll try to get that done in the next couple of days while it's fresh in my head!

Mostly as a reminder to myself:
Once that's approved and added back to delta, I'll remove the `lastCompleted` fields from the Dailies belonging to anyone who's used the delta test site.

paglias: Ah sorry I missed that it wasn't added here! i'd say if it's not yet used and it's possible to use directly the history array it'd be better to do that

Alys: It turns out to be not possible to use the history array directly, alas. Cron adds a history entry to each Daily that wasn't completed on the previous day, so the most recent entry isn't always the date that the Daily was last completed.

Unrelated to that, I'm adding more tests. I'll post to the PR when it's ready to be reviewed again.

paglias:
@Alys let me know if I misunderstood but this has been removed from delta

Alys:
@paglias Yeah, that's fine. I know it's a pain to have to keep merging PRs into delta. Sorry I'm taking so long with this.

From all my testing, the code is working fine and the new tests I wrote are passing, but they're causing the tests after them to fail, I think because the clock isn't being restored. I can't work out why it's not being restored though because the afterEach should be handling that.
https://github.com/HabitRPG/habitica/blob/3c0cd7067a3608f4f071ea139f3103d3dd9e0be2/test/api/unit/libs/cron.test.js#L37-L41

paglias:
@Alys I'm not sure why but if you split the texts making sure only one `clock` assignment exist per test they works

test/api/unit/libs/cron.test.js
```diff
@@ -1896,3 +1834,116 @@ describe('recoverCron', () => {
     }
   });
 });
+
+describe('cron for Dailies using every X days after completion', () => {
```
paglias: If it's a new top level describe box it should go in a separate file, this one is already huge also

Alys:
> @Alys I'm not sure why but if you split the texts making sure only one clock assignment exist per test they works

Sorry, I just saw that. Ignore the commit I just pushed; it's not meant to be related to your advice.

Thank you for the advice!

test/api/unit/libs/cron.test.js
Alys: No worries. You can ignore that commit though; it's a temporary one that I was using just to see if putting the new tests in their own box would help prevent timeouts (running the full test suite on my local install isn't usually reliable). I'm not expecting to keep the tests separated like that and I'm intending to move them back to where they were. 

I created it before I saw your comment about not using two or more clocks in one test and that idea is what I'm working on now.

Alys:
I've replaced `clock()` in the tests by using cron's `now` parameter. I'd forgotten it existed! It's much better for this sort of thing. :)

Note the TODO comment in the new tests in test/api/unit/libs/cron.test.js - it's for you to consider if you want me to change how the new tests are done. Splitting up the `expect()` calls into multiple tests will make it a bit easier to troubleshoot if the tests ever fail but it will increase the size of the tests.

This is ready to be reviewed now.

paglias:
This looks good to me codewise, I'll put it on delta

Alys:
I haven't found any bugs with this and no one else has reported any since the latest code went to delta. My feeling is it's ready to be deployed.

However @SabreCat is there anything that you want to consider for shared completion of Dailies in Group Plans?
As a summary of this PR:
For any Daily that has its `frequency` set to `daily`, there's a new boolean option `repeatAfterCompletion`. It's labelled on the Daily's edit form as "Repeat Every [x] Days Since Last Completed", where "x" is the number selected in the `everyX` ("Repeat Every") field.
There's also a new `lastCompleted` field that records the timestamp that the Daily was most recently marked complete.
When the `repeatAfterCompletion` option is set to `true`, if a Daily isn't completed on a day that it was due, it will remain due on all future days until it is completed.
When the Daily is completed, it will be rescheduled to the `everyX` number of days after the `lastCompleted` date.

If a Group Plan Daily uses `repeatAfterCompletion`, I'd expect it all to work correctly if the Daily has to be completed by every assigned person. 
However for a Group Plan Daily that's supposed to be completed by just one of the assigned people, when `repeatAfterCompletion` is `true` I think there'd need to be special code to work out which member's `lastCompleted` date should be used for rescheduling.
The simplest option might be to disable the `repeatAfterCompletion` feature for Group Plan Dailies (at least for now; once we've seen how they work in production for normal Dailies, then maybe they could be adjusted for Group Plans).

What do you think? Is there any extra info you'd like from me or anything you want me to look into?

MeanderingCode:
I just saw this and I don't have time to look into it right now, so I'm
responding from intuition.

A single completion shared daily should probably repeat since the last
completion by any member.

I will look at this code in the next couple of days and respond more
completely.

Also note that I'm mostly through with some advances in the logic regarding
"syncing" completion for single and all type shared dailies which takes
into account timezones and CDS. It will also record history of who
completed it in the master task.



Also, this is a great feature!

Alys:
@MeanderingCode Thanks very much! Sorry for not tagging you on my comment - I should have since you're working on shared Dailies. This sounds good to me: _"A single completion shared daily should probably repeat since the last completion by any member."_

paglias:
Is the testing done for this?

Alys:
Mostly. I need to make an improvement to how RYA handles every X days since completion Dailies, and I want to use the `yesterDailyScored` code from https://github.com/HabitRPG/habitica/pull/11179 (EDIT: https://github.com/HabitRPG/habitica/pull/11390 now) if it's approved, so prioritise that PR over mine. :)

paglias:
Sounds good! Although that PR is quite big and I think @SabreCat is handling it with the rest of the group plans stuff

paglias:
Since that PR is taking a long time, do we want to merge this in the meantime?

paglias:
@Alys I can't fix the conflicts on one of your branches so you'll have to do it or the PR will have to be reopened against a branch on the main repo

paglias:
Ok, I found out how to push commits to your branch so the conflicts should be resolved now but an extra check won't hurt

Alys:
@paglias Thanks very much for doing that! I've started testing locally but I might not be able to finish it for a couple of days. I'll post here when it's all tested.

[for my reference, my new local branch is new-dailies-repeat-X-days-since-completion-201811]

paglias:
The merge I did shouldn't have changed anything so it should still work as it did before

Alys:
Ah right, sorry! I thought when you said "extra check" you meant testing but you actually meant just checking that the merge conflicts were gone? I can't find any. It all looks good, thank you!

SabreCat:
This hasn't been worked on in quite some time. Closing for now--if we end up prioritizing this project again, we'll revisit!

PR #10859, opened 2018-11-20
additional checks on gear purchase

### LINKED ISSUE #10849: Enchanted Armoire equipment must be purchasable after losing from Death, etc
When your avatar dies or you use the upcoming Empty the Armory feature, if you lose a piece of gear from the Enchanted Armoire, currently you are not able to buy it back. It appears in your Rewards area and the Equipment shop but when you open its modal to try to buy it, the "Buy" button is greyed out.

The correct behaviour is that the equipment is purchasable directly with gold at any time that the player wants to buy it back. (It should also be available randomly in the Armoire but I've seen nothing to indicate that that's not working.)

This would need to be fixed before the Empty the Armory feature goes live.

negue:
fixes #10849

- [x] fix the purchase of previously owned special-items
- [x] added test

### DIFF
--- a/test/common/ops/buy/buyMarketGear.js
+++ b/test/common/ops/buy/buyMarketGear.js
@@ -246,5 +246,14 @@ describe('shared.ops.buyMarketGear', () => {
 
       expect(user.items.gear.owned).to.have.property('head_special_2', true);
     });
+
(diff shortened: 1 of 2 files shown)


paglias:
thanks for the test :)

PR #10870, opened 2018-11-27
Prevent progress being cleared when quest ends

### LINKED ISSUE #9698: prevent boss damage being cleared when a quest ends
Currently, when a quest ends, `user.party.quest.progress.up` and `user.party.quest.progress.collectedItems` are set to zero for all party members.

This should no longer be done. Any quest progress that a player builds up should remain in their account until they go through cron.

This is partly to deal with https://github.com/HabitRPG/habitica/issues/3693 and partly to prevent players feeling unmotivated or disappointed when their quest progress disappears.

Arsnael:

Fixes #9698

### Changes
This changes fixes the issue that user's progress on a quest is being cleared when those events occur : 
* Quest is finished and user was part of the quest (standard case)
* Quest is started and user is not part of the quest 
* User rejects a quest
* Quest gets cancelled/aborted by owner
* User leaves current quest

Now it keeps the progress (damage and items collected) and clear it only during the cron event.

----

### DIFF
--- a/test/api/unit/models/group.test.js
+++ b/test/api/unit/models/group.test.js
@@ -32,8 +32,19 @@ describe('Group Model', () => {
       privacy: 'private',
     });
 
+    let _progress = {
+      up: 10,
+      down: 0,
+      collectedItems: 5,
+    };
+
     questLeader = new User({
-      party: { _id: party._id },
+      party: {
+        _id: party._id,
+        quest: {
+          progress: _progress,
+        },
+      },
       profile: { name: 'Quest Leader' },
       items: {
         quests: {
@@ -45,20 +56,40 @@ describe('Group Model', () => {
     party.leader = questLeader._id;
 
     participatingMember = new User({
-      party: { _id: party._id },
+      party: {
+        _id: party._id,
+        quest: {
+          progress: _progress,
+        },
+      },
       profile: { name: 'Participating Member' },
     });
     sleepingParticipatingMember = new User({
-      party: { _id: party._id },
+      party: {
+        _id: party._id,
+        quest: {
+          progress: _progress,
+        },
+      },
       profile: { name: 'Sleeping Participating Member' },
       preferences: { sleep: true },
     });
     nonParticipatingMember = new User({
-      party: { _id: party._id },
+      party: {
+        _id: party._id,
+        quest: {
+          progress: _progress,
+        },
+      },
       profile: { name: 'Non-Participating Member' },
     });
     undecidedMember = new User({
-      party: { _id: party._id },
+      party: {
+        _id: party._id,
+        quest: {
+          progress: _progress,
+        },
+      },
       profile: { name: 'Undecided Member' },
     });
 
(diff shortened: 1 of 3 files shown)


test/api/unit/models/group.test.js
```diff
@@ -19,7 +19,7 @@ import * as email from '../../../../website/server/libs/email';
 import { TAVERN_ID } from '../../../../website/common/script/';
 import shared from '../../../../website/common';
 
-describe('Group Model', () => {
+describe.only('Group Model', () => {
```
Alys: .only should be removed. :)  (This isn't a full code review, just one thing that I noticed.)

Arsnael: Ah probably I forgot to take it out after finishing testing... Rookie mistake sorry :) Taking that out pronto !

Alys: Common mistake, not just for rookies. :)

Arsnael: Fixed !

Arsnael:
Something wrong with Travis btw?

Alys:
Yeah, something looks wrong with Travis. I restarted that build but it all failed again. Let's just wait a few hours and then try again.

Arsnael:
https://github.com/dominictarr/event-stream/issues/116

Looks like related, the so called package is the one that `npm ci` cannot find (end of the issue people saying it returns 404 now as well). Might be interesting for the devs to take a look to that post I think :)

Edit : seems flatmap-stream@0.1.1 includes malicious code, but seems to affect more like crypto currency apps (like stealing bitcoin). I guess you don't use crypto in Habitica right?

paglias:
Yeah I saw that yesterday, we are not affected but I'll look into updating the deps to a safe version

paglias:
@Arsnael ok it should be fixed now in the latest `develop` branch

Arsnael:
Great work @paglias . Should I rebase this branch with develop then?

paglias:
@Arsnael yes, either rebase or merge the latest develop

Arsnael:
Done !

website/server/controllers/api-v3/quests.js
```diff
@@ -245,7 +245,7 @@ api.rejectQuest = {
     if (group.quest.members[user._id]) throw new BadRequest(res.t('questAlreadyAccepted'));
     if (group.quest.members[user._id] === false) throw new BadRequest(res.t('questAlreadyRejected'));
 
-    user.party.quest = Group.cleanQuestProgress();
+    user.party.quest = Group.cleanQuestUser(user.party.quest.progress.toObject());
```
paglias: is `user.party.quest.progress` always defined? otherwise we can add a check to make sure we're not calling toObject on null or undefined

test/api/unit/models/group.test.js
```diff
@@ -32,8 +32,19 @@ describe('Group Model', () => {
       privacy: 'private',
     });
 
+    let _progress = {
+      up: 10,
+      down: 0,
```
paglias: if nothing is preventing it I'd like to change the `down` value here to something different than `0` so that we can test it doesn't get reset

paglias (commented):
Left a couple of comments

paglias:
@SabreCat do you think this can be merged directly or should go to delta?

test/api/unit/models/group.test.js
Arsnael: Changed ! But regarding the start quest case, progress.down is being reset by the code (didn't modify) for participating members, so the fact I still check towards 0 in the tests on those cases is expected.

website/server/controllers/api-v3/quests.js
Arsnael: Fixed. Let me know if it's alright or if there was a better way of doing it

Arsnael:
Fixed regarding @paglias 's comments. Please review ! :)

paglias (approved):

paglias:
Sorry for the delay here @Arsnael but we had to do some testing !

paglias:
Notated towards your 4th tier

Arsnael:
No worries ! Thanks @paglias :)

PR #10872, opened 2018-11-27
Improve @mention handling

phillipthelen:
This PR adds a new API call `/members/find/STRING`, that searches for users whose username contains the searched string. It also implements this autocompleting for the group chats.

If a user wants to, they can also opt out of appearing in these searches through a new settings in the preferences.

NOTE: This has to be merged/deployed together or after #10844


website/client/components/chat/chatCard.vue
```diff
@@ -248,6 +248,16 @@ export default {
       return achievementsLib.getContribText(message.contributor, message.backer) || '';
     },
   },
+  mounted () {
+    const links = this.$refs.markdownContainer.getElementsByTagName('a');
+    for (let i = 0; i < links.length; i++) {
+      const link = links[i];
+      links[i].onclick = () => {
+        event.preventDefault();
+        this.$router.push({ path: link.getAttribute('href')});
```
paglias: just curious, is this necessary? i mean keeping a normal `a` tag with no `onclick` handler doesn't work?

website/client/components/groups/chat.vue
```diff
@@ -155,6 +168,30 @@
...
+        this.updateChatInput();
+      },
+      updateChatInput () {
+        let innerText = this.$refs['user-entry'].innerText;
+        if (innerText[innerText.length - 1] === '\n') {
+          innerText = innerText.slice(0, -1);
+        }
+        this.newMessage = innerText;
+      },
+      getTierIcon (member) {
```
paglias: `getTierIcon` and `tierIcon` seem the same function

website/client/components/groups/chat.vue
```diff
@@ -196,20 +235,32 @@
...
-      border: solid 1px $gray-400;
-      font-style: italic;
+      box-shadow: 0 0 3pt 2pt white;
+      border-radius: 2px;
       line-height: 1.43;
-      color: $gray-300;
       padding: .5em;
+      -moz-appearance: textfield-multiline;
+      -webkit-appearance: textarea;
+      background-color: -moz-field;
```
paglias: is this rule firefox specific?

website/client/components/groups/chat.vue
```diff
@@ -233,6 +284,84 @@
...
+    }
+  }
+
+  .v-tribute {
+    width: 100%;
+  }
+</style>
+
+<style lang="scss">
+  @import '~client/assets/scss/tiers.scss';
```
paglias: it looks like this file is being imported multiple times acrsso the website, resulting in duplicate css code, can you remove all the `@import '~client/assets/scss/tiers.scss';` (there are a couple more) and import it only once in `assets/scss/index`

website/client/components/groups/chat.vue
```diff
@@ -233,6 +284,84 @@
...
+    margin: 0;
+    margin-top: 2px;
+    padding: 0;
+    list-style: none;
+    background: #fff;
+    border-radius: 4px;
+    border: 1px solid rgba(#000, 0.13);
+    background-clip: padding-box;
+    overflow: hidden;
+    -moz-transition: none;
```
paglias: it's not necessary to add browser-specific rules, we have a a plugin that automatically adds them when needed, for example in this case `transition: none;` alone will do the trick

website/client/components/groups/chat.vue
```diff
@@ -233,6 +284,84 @@
...
+    -webkit-transition: none;
+    transition: none;
   }
 
+  .tribute-container li {
+    color: $gray-200;
+    padding: 12px 24px;
+    cursor: pointer;
+    font-size: 14px;
+    -moz-transition: none;
```
paglias: same here (about browser specific rules)

website/server/controllers/api-v4/members.js
```diff
@@ -0,0 +1,34 @@
...
+    let username = req.params.username.toLowerCase();
+    if (username[0] === '@') username = username.slice(1, username.length);
+
+    if (username.length < 1) {
+      res.respond(200, []);
+      return;
+    }
+
+    let members = await User
+      .find({'auth.local.lowerCaseUsername': {$regex: `.*${username}.*`}, 'flags.verifiedUsername': true, 'preferences.searchableUsername': {$ne: false}})
```
paglias: let's make sure we have the necessary indexes here or this'll be super slow

website/client/components/groups/chat.vue
```diff
@@ -68,6 +76,38 @@
...
+          tier6,
+          tier7,
+          tier8,
+          tier9,
+          tierNPC,
+        }),
+        autocompleteOptions: {
+          async values (text, cb) {
+            if (text.length > 0) {
+              let suggestions = await axios.get(`/api/v4/members/find/${text}`);
```
paglias: this should be throttled (or debounced) or it'll flood the server with requests

website/client/components/groups/chat.vue
```diff
@@ -1,33 +1,27 @@
...
+        .user-entry(:placeholder='placeholder',
+                  @input="updateChatInput"
                   ref='user-entry',
                   :class='{"user-entry": newMessage}',
-                  @keydown='updateCarretPosition',
                   @keyup.ctrl.enter='sendMessageShortcut()',
                   @paste='disableMessageSendShortcut()',
-                  maxlength='3000'
+                  maxlength='3000',
+                  :contenteditable='true'
```
paglias: this will make the textarea render any type of HTML, I understand why it was added but I feel like it we should add some sort of sanitization or people will be able to copy and paste entiere word docs or html pages and have them rendered

website/server/controllers/api-v4/members.js
paglias: tbh I'm a bit worried about the number of requests this route will get and the performances, considering the $regex type of search. Can we at least make it so the prefix part is fixed so that the search is faster as explained here https://docs.mongodb.com/manual/reference/operator/query/regex/#index-use?

paglias (changes requested):
Left some comments, but didn't test it live. Looks like there's a small conflict to fix

package.json
```diff
@@ -90,6 +90,7 @@
     "svg-url-loader": "^2.3.2",
     "svgo": "^1.0.5",
     "svgo-loader": "^2.1.0",
+    "tributejs": "^3.4.0",
```
paglias: is this used anywhere? it looks like it's installed but not used

paglias (changes requested):

website/server/libs/highlightMentions.js
```diff
@@ -0,0 +1,22 @@
+import {model as User} from '../models/user';
+
+const mentionRegex = new RegExp('\\B@[-\\w]+', 'g');
+
+export async function highlightMentions (text) {
+  const mentions = text.match(mentionRegex);
```
paglias, line 6: can we move this call https://github.com/HabitRPG/habitica/commit/d934d9d759e75cc786163646f44c00e52e4cfc98#diff-ade116c537945ae89e2f5f059d4d577fR199 to a single place so we don't count mentions too as regexps can be expensive?

paglias (changes requested):

website/server/libs/highlightMentions.js
phillipthelen: I wasn't sure how/where to handle that best. Do you have a suggestion for where to make this call?

website/client/components/chat/chatCard.vue
phillipthelen: No, since the markdown parser creates the `a` tags so that they open in a new tab.

website/server/controllers/api-v4/members.js
phillipthelen: We have an index on that field.

I can also change it to work that way, however then you can only search by typing the beginning of the name correctly. But that's probably fine at least for now.

package.json
phillipthelen: It's used for the autocomplete here https://github.com/HabitRPG/habitica/pull/10872/files/8641a98107210cd1b25f06008c7ea38d41d872bb#diff-f9db085b45d481950984d26ae407e383R7

website/client/components/chat/chatCard.vue
paglias: makes sense, thanks!

website/server/controllers/api-v4/members.js
paglias: I'd say it's fine for now, if people complain or there are no performance issues we can extend it but I'd prefer to start with the faster version if possible

package.json
paglias: Oh I see it's a peerDependency of vue-tribute

website/server/libs/highlightMentions.js
paglias: hmmm... maybe `highlightMentions()` could return both the text and the mentions? so array of [text, mentions] and use the mentions returned in api.postChat?

website/client/components/chat/chatCard.vue
paglias: Actually I just tested this and it still opens the new tab, you are missing the event in the function definition: `links[i].onclick = () => {` ->  `links[i].onclick = (event) => {`

paglias: (and it should not be called `event` but `ev` or something else because `event` is defined in the global scope so it will throw an eslint error)

website/client/components/groups/chat.vue
phillipthelen: Not entirely sure how to handle this (or big of an issue this actually is, since it would only affect how it's displayed in this box and not the actual chat message). I could import a library to handle this sanitizing? Writing that myself might be quite some work.

paglias: I think the problem is that it would affect the displayed chat as well because markdown renders HTML code as well so it would be rendered. My biggest fear is that it could contain unsanitized JavaScript as well which could result in serious security issues but I haven't tested it

phillipthelen: This wouldn't be a security issue. Since the message/input has to be sanitized on the server side to catch any security issues. The issue this could cause is that if you copy styled text, it might show up as styled text in the input box. But if unsanitized js can be posted, then that is an issue that has to be fixed on the postChat API call.

phillipthelen:
Apart from the contenteditable thing, this PR should now also be ready. With that, I'm still not sure how to handle it. I guess the options are either using a library to sanitize the input here, or just leave it, since it is a local display issue that is probably not very common anyway?

website/server/controllers/api-v4/members.js
paglias: Hey @phillipthelen looks like this wasn't actually changed, the regexp

> A regular expression is a “prefix expression” if it starts with a caret (^) or a left anchor (\A), followed by a string of simple symbols. For example, the regex /^abc.*/ will be optimized by matching only against the values from the index that start with abc.

paglias (changes requested):
There is a conflict and i left a comment on the regexp. Also, as discussed internally would you add the cache headers to `api.getUsernameAutocompletes`? It should be enough to add the following inside the route:

```
  res.set('Cache-Control', 'public, max-age=300000'); // 5 minutes
```

website/server/controllers/api-v4/members.js
paglias: Marked as unresolved

paglias:
Not sure if this can go directly on staging or if you want to test on delta

website/server/controllers/api-v4/members.js
phillipthelen: Sorry, I thought we'd leave the regex as is for now, since you last comment was it's fine for now. I changed it to work the way you requested

phillipthelen:
As discussed with @saraolson, the API now has an optional parameter to give a "context" for autocompletion, which means it can search according to a specific guild or party. For now it limits results to members of the party or guild and does not show other results.

website/server/controllers/api-v4/members.js
```diff
@@ -0,0 +1,49 @@
...
+
+let api = {};
+
+
+api.getUsernameAutocompletes = {
+  method: 'GET',
+  url: '/members/find/:username',
+  middlewares: [authWithHeaders()],
+  async handler (req, res) {
+    res.set('Cache-Control', 'public, max-age=300000'); // 5 minutes
```
paglias: I fee like now that we have different results based on which user makes the request this is going to create some problems

phillipthelen: Oh, that's a good point. I could try if it works to only cache some of the calls? So that only the public searches are cached and all searches in private spaces are not cached. Do you know if that will work with googles caching?

paglias: So we can cache everything that's the same for all users, everything else should not be cached. So I guess if the results for public searches is the same for everyone it'd be okay to cache them

phillipthelen:
This got a bunch more complicated.

What now happens: if a context+id is passed, the recent chat messages for that group are searched for matching users and those are used for autocompleting. If there are less than 5 results, it also does another search for users that have not chatted recently. It also takes into account blocked and muted users and does not autocomplete those.

website/server/controllers/api-v4/members.js
```diff
@@ -23,25 +25,64 @@ api.getUsernameAutocompletes = {
...
+      }));
+
+      let recentChatQuery = Object.assign({}, commonQuery);
+      recentChatQuery._id = {$in: recentChatters};
+      query._id = {$nin: recentChatters};
+      members = await User
+        .find(recentChatQuery)
+        .select(['profile.name', 'contributor', 'auth.local.username'])
+        .sort({'auth.timestamps.loggedin': -1})
+        .limit(5 - members.length)
```
paglias: member before this query isn't always an empty array? So members.length will always be 0, or not?

paglias (commented):
Left one comment but tbh this is going to be quite slow if at every request it has to fecth 200 messages chat messages everytime and in some cases make an extra User query

phillipthelen:
I updated the title for the issue, since I removed the serverside autocomplete code. So now the PR has the @mentioning highlighting in messages and will send push notifications once I have that code done.

website/server/controllers/api-v3/auth.js
```diff
@@ -308,6 +308,8 @@ api.resetPassword = {
       ]);
 
       await user.save();
+    } else {
+      let user = await User.findOne({ 'auth.google.email': email }).exec();
```
paglias: is this used anywhere?

phillipthelen:
Implemented sending push notifications on mentioning. I made the following rules for that:

1. There are three settings. "Mention in party", "Mention in guild user is a member of", "Mention in guild user is NOT a member of"
2. If a user has party activity notifications active and is mentioned, the party activity notification is not sent. (unless user has disabled "Mention in party"
3. For private guilds, notifications are only sent if user is member of the party

I think that should cover the usecases?

website/client/components/chat/chatCard.vue
```diff
@@ -217,6 +216,21 @@ export default {
       return 'Message hidden (shadow-muted)';
     },
   },
+  mounted () {
+    const links = this.$refs.markdownContainer.getElementsByTagName('a');
+    for (let i = 0; i < links.length; i++) {
+      const link = links[i];
+      if (links[i].getAttribute('href').startsWith('/profile/')) {
```
paglias, line 223: what's the default action that's prevented here? I mean what's wrong with keeping the normal link? that it opens them in a new window?

website/server/models/user/schema.js
```diff
@@ -469,6 +469,7 @@ let schema = new Schema({
     reverseChatOrder: {$type: Boolean, default: false},
     background: String,
     displayInviteToPartyWhenPartyIs1: {$type: Boolean, default: true},
+    searchableUsername: {$type: Boolean, default: false},
```
paglias: this doesn't seem to be used anywhere

paglias (commented):
left a couple of comments and there are a few failing tests

website/client/components/chat/chatCard.vue
phillipthelen: Yes, exactly. Normally it would open in a new tab, but this way it opens in the same window as the modal. (same as when you click on their username at the top of a chat message.). This solution also still allows users to CMD+Click to open in a new tab in the background.

paglias (approved):

PR #10874, opened 2018-11-27
Move group chat analytics event to server and track @mention count

phillipthelen:
This way mobile apps are also tracked and we can get an idea for how many @mentions messages usually have.


SabreCat:
Cherry-picked onto `release` for today's deploy.

PR #10891, opened 2018-12-04
The Istanbul command string in the gulp-tests.js file for each tests are breaking the test run on Windows machines. 

sukh0128:
### Changes

The gulp/gulp-tests.js file is edited from having 
"node_modules/.bin/istanbul" string for the istanbul file location to just "istanbul". The previous version breaks the test suite run on Windows machines and removing the directory for the file actually runs the test fine. Tested on 3 different local install of Habitica on windows machines.

Fixed Error:

- 'node_modules' is not recognized as an internal or external command, operable program or batch file.

Result:

- Now the full test suite and both the individual tests run on windows machine.


gulp/gulp-tests.js
```diff
@@ -167,7 +167,7 @@ gulp.task('test:content:safe', gulp.series('test:prepare:build', (cb) => {
 
 gulp.task('test:api:unit', (done) => {
   let runner = exec(
-    testBin('node_modules/.bin/istanbul cover --dir coverage/api-unit node_modules/mocha/bin/_mocha -- test/api/unit --recursive --require ./test/helpers/start-server'),
+    testBin('istanbul cover --dir coverage/api-unit node_modules/mocha/bin/_mocha -- test/api/unit --recursive --require ./test/helpers/start-server'),
```
paglias, line 170: does this work even if the `istanbul` package isn't installed globally?

paglias:
can you revert the changes to package-lock.json? thanks!

paglias:
thanks @sukh0128 ! If you have an Habitica UUID I'll be happy to award you a contributor tier

PR #10898, opened 2018-12-06
Usernames Misc: Bulk Email and New Party Modal

SabreCat:

### Changes

* Updates the bulk email script to use Mongoose models and sets it up as a reusable template.
* Edits the "New Party" dialog to use Usernames instead of UUIDs, including some style updates from Zeplin.
* Archives several old-style migrations that still used `monk`.


website/client/components/groups/createPartyModal.vue
```diff
@@ -61,140 +75,120 @@ b-modal#create-party-modal(title="Empty", size='lg', hide-footer=true)
...
+    background-color: $white;
+    border: 0px;
+    border-radius: 0px;
+    color: $gray-300;
+    padding: 0rem 0.1rem 0rem 0.75rem;
   }
 
   .join-party {
-    background-image: url('~client/assets/images/party@3x.png');
+    background-image: url('~client/assets/images/party.png');
```
paglias, line 115: is this going to look bad on hd displays?

paglias (commented):
just one comment on an image

website/client/components/groups/createPartyModal.vue
SabreCat: At the moment it looks bad at 1920x1080, so I'm not sure which way is preferable..

paglias: Have you tried the @3x version?

SabreCat: Yep, that's what it was before. It scales down and gets blurry.

paglias: Hmm it looks like that with higher def images you have to specify the background-size https://stackoverflow.com/questions/16154494/retina-displays-high-res-background-images to the original size, but if it looks good, i'd say it's fine

PR #10909, opened 2018-12-13
extract site urls from translations

### LINKED ISSUE #10213: Change links in translation strings to variables
**See [this comment](https://github.com/HabitRPG/habitica/issues/10213#issuecomment-403245134) for current status of this issue**

Report from a user: 

"Windows/Desktop browser + language Spanish, When opening an egg with the potion, appears the window:
¡Has eclosionado una nueva mascota! Cachorro de panda Base ¡Visita el Establo para alimentar y equipar a tu nueva mascota!
The "stable" link points to "inventario/establo", not "inventory/stable", producing a 404 error page."

Confirmed in Spanish language, doesn't occur in French (as one example).

negue:
- [x] extracted all links that point to local pages

fixes #10213

should external urls be extracted too? those can be easier translated if they stay as they are.

also I couldn't find the string "landingend2" anywhere, was it removed?


Alys:
Links to the wiki should stay in the translatable strings - you're right that it's easier that way.

paglias:
thanks!

PR #10911, opened 2018-12-15
drag tags to reorder

### LINKED ISSUE #7541: Tags cannot be sorted / reordered

negue:
re-add the ability to drag tags

image
@Tressley 

closes #7541


Tressley:
I think this looks good. Would it be possible to have the knurling (six dots) be #C3C0C7 and #878190 on hover? Also, the re-order doesn't stay if the user clicks cancel, correct?

negue:
color changed, and now it only changes the tags on save :)

Tressley:
I'll definitely want to play with a bit on Staging to see how it feels, otherwise LGTM 👍

paglias:
Tried this locally and it works well! @Tressley it'll be on staging in about 30 minutes

PR #10912, opened 2018-12-15
fix emails in chat

### LINKED ISSUE #10875: Email address display breaks
From RaB:

> I don't know that it's a valid email address anyhow, but I've seen email examples (or "blanked out" address names here in the report a bug guild) where the username part ends in a dot breaking the email link display like so:
> 
> @yahoo.com" target="_blank" rel="noopener">.....@yahoo.com [see screenshot]
> 
> Even if the address isn't real, I don't think that should happen. (I know at some point it was happening with all email address links. Similar issue?)

image

negue:
- added additional regex checks to ignore the <tag attribute="content"> if an email contains the same name a username/displayname has

fixes #10875


paglias:
@negue thanks! it looks like there's an issue with the regexp:
schermata 2018-12-23 alle 19 37 24


I tried with a chat message with the following text `hello y.@gmail.com other words`

paglias:
still getting the same error @negue

negue:
I "love" RegEx ... 😅 

I never had this issue, it worked in Chrome^^, now I also tested in Firefox and it was finally the same error you had.

After searching what the issue could be, it is that the RegEx `lookbehind` rules (I used for checking possible HTML-Tags to prevent the issue in #10875) are not available in all browsers, or they are just available in Chrome.

Thanks to: https://stackoverflow.com/a/33824409/1298154 I could rewrite the expression without `lookbehind`.

Edit: the tests worked before I pushed it, gonna test it again
Edit2: the previously needed check for html tags was for the double regex, now that this was already removed, the checks not needed anymore

TL;DR fixed

paglias:
thanks @negue ! sorry I didn't specify it was firefox :)

PR #10913, opened 2018-12-15
Streak-Achievement notification/modal timeout 

### LINKED ISSUE #10813: streak achievement notification disappears; modal nearly inaccessible
When you get a streak achievement (i.e., the streak on a Daily reaches a multiple of 21), a blue streak achievement notification appears near the top right of the website and you can click on it to see the streak achievement modal. However the notification disappears after a few seconds so you have to click very quickly.

Also, it's not at all obvious that you can click on the notification to see the modal. The only reason I knew you could was because I'd seen the PR about it. Because it's not obvious, we've had a bug report saying that the streak achievement modal no longer appears (from JoeHandyman, 7438155c-e523-4afa-a3ac-d57ba9f94bdd in the Report a Bug guild).

My suggestions:

- If the player has `preferences.suppressModals.streak` set to `false`, then make the streak achievement notification remain on screen until it's clicked (in the same way that error notifications stay on screen).
- If the player has `preferences.suppressModals.streak` set to `true`, make the streak achievement notification disappear after a couple of seconds just as it does now. This is because the player has chosen to not see the streak achievement modal so forcing them to click on the notification will just annoy them. 

Regardless of what setting the user has chosen for `preferences.suppressModals.streak`, if the player does click on the notification, the streak achievement modal should appear.

I.e., I'm suggestion that we change the effect that the `preferences.suppressModals.streak` setting has. It should **not** control whether the modal appears or not. Instead, it should control whether the notification disappears automatically or not. I think this is the best compromise between the old behaviour of letting the user choose to see the modal or not, and the new behaviour where you get a notification that has to be clicked to see the modal.

EDIT: I'm leaving this crossed-out because we'd want the link to be designed by the staff. This could be added later if desired. ~Also, I suggest that the notification should have something on it to indicate that you can click it for more information (for example, text saying "more information" at the bottom right of the notification, formatted to look like a link, but there might well be a better way of indicating that it's clickable).~

I'll leave this as suggestion-discussion for a few days for the staff to comment and then if there's no objections, I'll mark it as help wanted.

negue:
removes the timeout if the `suppressModals.streak` option is not enabled

fixes #10813

### DIFF
--- a/website/client/components/notifications.vue
+++ b/website/client/components/notifications.vue
@@ -569,10 +569,8 @@ export default {
             break;
           case 'STREAK_ACHIEVEMENT':
             this.text(`${this.$t('streaks')}: ${this.user.achievements.streak}`, () => {
-              if (!this.user.preferences.suppressModals.streak) {
-                this.$root.$emit('bv::show::modal', 'streak');
-              }
-            });
+              this.$root.$emit('bv::show::modal', 'streak');
+            }, this.user.preferences.suppressModals.streak);
             this.playSound('Achievement_Unlocked');
             break;
           case 'ULTIMATE_GEAR_ACHIEVEMENT':


paglias:
thanks @negue !

PR #10936, opened 2019-01-06
Staff sweetness duplicated two handed description

### LINKED ISSUE #10929: Staff of Sweetness duplicating part of description
Lucas Heim @lucasheim (535a72cc-f949-4ba6-8570-639c5f763da8): "Hello!
The special item "Staff of Sweetness" is duplicating the "Two-handed item." description. Currently, it displays like this:
"This is no ordinary lollipop! The glowing orb of magic sugar atop this staff has the power to make good habits stick to you. Increases Intelligence by 15 and Perception by 7. Limited Edition 2018 Autumn Gear. Two-handed item. Two-handed item."

It seems to be a pretty easy fix, in the locale file the text was put explicitly on the item description, but I think this is added automatically (https://github.com/HabitRPG/habitica/blob/4f3536e88756001d8d1fb3da867a5a0cafbfd337/website/common/locales/en/gear.json)
If it's just this change on all locales, I can try to do it when I get home and issue a PR :)

Happy new year!"

lucasheim:
Fixes #10929 

### Changes
Removed all "Two-handed item" descriptions that were hardcoded in the locale files. This description is already added dynamically in [weapon.js](https://github.com/HabitRPG/habitica/blob/develop/website/common/script/content/gear/weapon.js)

----
UUID: 535a72cc-f949-4ba6-8570-639c5f763da8


negue:
~you only need to remove the "two handed item" - string from the `en` file , all other files are managed in transifex :)~

see Alys comment :)

Alys:
In this specific case, it might be worth merging all these files since they have been edited now. We're no longer using Transifex and it could be a while before we have a replacement service set up.

There's a post from beffymaroo in the [Commonwealth of i18n (Aspiring Linguists)](https://habitica.com/groups/guild/7732f64c-33ee-4cce-873c-fc28f147a6f7) guild with details about leaving Transifex, if anyone is interested.

paglias:
thanks @lucasheim ! notated towards your 2nd tier

PR #10952, opened 2019-01-18
Upgrade to NPM Amplitude package

SabreCat:

### Changes

The existing client SDK for Amplitude analytics is being sunset soon. This PR adds the `amplitude-js` NPM package so we can use that instead of the old script, currently fetched inline.


paglias:
@SabreCat looks good, I'd just check that calling `amplitude.logEvent` works fine because here https://amplitude.zendesk.com/hc/en-us/articles/115001361248-JavaScript-SDK-Installation#tracking-events and here https://amplitude.zendesk.com/hc/en-us/articles/115002889587#class-amplitude it seems that first you have a to create an instance with `amplitude.getInstance()` and then use the returned object to log events, ... but I'm not sure

paglias:
just a reminder that this needs to go live before Jan 31st

SabreCat:
Tested on Delta, confirmed events and user properties are coming through OK there. Will see how automated tests go, next--might need to update the syntax there as well.

SabreCat:
This is... a very strange way for the tests to fail. 🤔 Investigating tho...

PR #10957, opened 2019-01-22
Add else block to prevent selling option for non-sellable items (saddle)

### LINKED ISSUE #9596: Saddles Should Not Be Sellable

ChesterSng:
[//]: #9596  (Put Issue # here, if applicable. This will automatically close the issue if your PR is merged in)
Fixes #9596 

### Changes
It looked like the else block was missing in the code. Added an else block so items with sellWarningNote() will not have sell options displayed in the modal. 

image

----


ChesterSng:
Hi @Alys is this PR okay? :)

paglias:
Hi @ChesterSng, thanks for the PR! Looks like two extra commits got included here, could you remoe them? If you don't know how to it might be a good idea to create this PR and open a new one, thanks! And let us know if you have any question

ChesterSng:
Hi @paglias, removed them and re-opened at #10965!

Thanks :)

PR #10971, opened 2019-01-31
refresh gear overview

### LINKED ISSUE #10920: when buying new equipment, item does not disappear from Market or Seasonal Shop

negue:
fixes #10920

at first I thought it was the the seasonal gear didn't refresh the "pinnedItems", then the refresh in market did update the seasonal pinned gear, but not all unpinned equipment in the seasonal shop

then it was clear, it was "just" the vue watcher that didn't get any updates (e.g. no pinned gear was bought, the pinned array didn't change => no object change => no refresh)

now refreshing `user._v++` (and subscribing in the seasonal market object) fixed the refresh 🎉

### DIFF
--- a/website/common/script/libs/shops.js
+++ b/website/common/script/libs/shops.js
@@ -417,6 +417,8 @@ shops.getSeasonalGearBySet = function getSeasonalGearBySet (user, set, officialP
 shops.getSeasonalShop = function getSeasonalShop (user, language) {
   let officialPinnedItems = getOfficialPinnedItems(user);
 
+  // subscribe to user._V, to let vue refresh the seasonal shop overview
+  // eslint-disable-next-line
   const myUserVersion = user._v;
 
   let resObject = {
@@ -430,8 +432,7 @@ shops.getSeasonalShop = function getSeasonalShop (user, language) {
       text: i18n.t(seasonalShopConfig.featuredSet),
       items: shops.getSeasonalGearBySet(user, seasonalShopConfig.featuredSet, officialPinnedItems, language, true),
     },
-    // return / subscribe to user._V, to let vue refresh the seasonal shop overview
-    _v: myUserVersion,
+
   };
 
   return resObject;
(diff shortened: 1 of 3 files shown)


website/common/script/libs/shops.js
```diff
@@ -428,6 +430,8 @@ shops.getSeasonalShop = function getSeasonalShop (user, language) {
       text: i18n.t(seasonalShopConfig.featuredSet),
       items: shops.getSeasonalGearBySet(user, seasonalShopConfig.featuredSet, officialPinnedItems, language, true),
     },
+    // return / subscribe to user._V, to let vue refresh the seasonal shop overview
+    _v: myUserVersion,
```
paglias: I think this is returned by default under `userV`as a top level response property (next to `success` and `data`)

negue: I just used this to store the `myUserVersion` so that I don't get an `unused-var` lint error, I could remove this and just have the following:


```
  // eslint-disable-next-line
  const myUserVersion = user._v;
```

this seems to work with it too

paglias: Ok I think you can remove the changes in this file and it'll work anyway because user._v is returned anyway in the response top level object

paglias (changes requested):

website/common/script/libs/shops.js
negue: yeah I don't need the return value at all, I just need to "subscribe" in the vue-way (see line [420](https://github.com/HabitRPG/habitica/pull/10971/files/b3d3d5d5cffef06696569858cebf9c47025f19f6#diff-bbbfee4959388a7be92457cbcc12434fR420) ) so that I can get the updates inside the web app, without it it'll stay "unchanged" for vue

paglias: could this be moved to a clien side file only with a comment on why it's needed? maybe where it calls the common function? Otherwise I'd still add a more exhaustive comment on why it's needed and I'd remove it from the return value as it's confusing

negue: subscribed in the vue-file itself, somehow I thought it would be better to do inside the lib^^, should be good now

negue:
the test that failed was `sends appropriate emails when subscribed member of group must manually cancel recurring Android subscription` which shouldn't be affected by my changes, @paglias could you restart travis

paglias:
thanks!

PR #10973, opened 2019-02-02
Fix #10907 - Removes Grey Disabled style from Challenge Progress

### LINKED ISSUE #10907: challenge participant progress page should not use grey for tasks
_EDIT: There's a partial PR at https://github.com/HabitRPG/habitica/pull/10973 which a contributor could build on if they didn't want to work on this issue from scratch._

When you are looking at a challenge's page (e.g., a page such as `https://habitica.com/challenges/6c565677-65fc-400f-b39e-52bd01e4cd7a`) and you use the "View Progress Of" > "Select a Participant" menu to view the progress of a participant, completed Dailies and To-Dos should not be grey - they should show their red/orange/yellow/green/blue colour so that you can judge the participant's history of performing that task.

For example, in the screenshot below, the Daily should not be grey.
image

Whoever works on this issue should test that their change does not affect any tasks (including challenge tasks) in a player's own tasks list, or in the "Group Tasks List" screen of a guild or party with a Group Plan. For information about adding a Group Plan to a guild/party, see "Creating Test Data" in [Using Your Local Install to Modify Habitica's Website and API](https://habitica.wikia.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API)

Mira-M:
- Includes a fix that displays task styles based on tab location
- Adds unit test coverage to ensure styles exist (or do not exist) depending on location
- Fixes #10907 

### Changes

Approach:
Since the behaviors were different depending on the location, I passed the route's path name to the task class creation functionality so that it was aware of where the request originated.  I was then able to ensure that todo and daily tasks originating from the challenge page did not get the disabled style applied. 

I added unit tests testing the functionality from the tasks page, the guild/group task board and the challenge page.

#### Before

image

#### After

image

image

#### Regression: User Tasks View (Completed Todo/Daily Tasks are still grey)
image

----
UUID: 

ca561633-c0bb-4544-adf4-dda0aaa23795

### DIFF
--- a/website/client/components/tasks/task.vue
+++ b/website/client/components/tasks/task.vue
@@ -598,13 +598,13 @@ export default {
       return false;
     },
     controlClass () {
-      return this.getTaskClasses(this.task, 'control', this.dueDate);
+      return this.getTaskClasses(this.task, 'control', this.dueDate, this.$route.name);
     },
     contentClass () {
       const type = this.task.type;
 
       const classes = [];
-      classes.push(this.getTaskClasses(this.task, 'control', this.dueDate).content);
+      classes.push(this.getTaskClasses(this.task, 'control', this.dueDate, this.$route.name).content);
 
       if (type === 'reward' || type === 'habit') {
         classes.push('no-right-border');
--- a/website/client/store/getters/tasks.js
+++ b/website/client/store/getters/tasks.js
@@ -52,7 +52,7 @@ export function getTaskClasses (store) {
   // Edit Modal: edit-modal-bg, edit-modal-text, edit-modal-icon
   // Create Modal: create-modal-bg, create-modal-text, create-modal-icon
   // Control: 'control'
-  return (task, purpose, dueDate) => {
+  return (task, purpose, dueDate, loc) => {
     if (!dueDate) dueDate = new Date();
     const type = task.type;
     const color = getTaskColor(task);
@@ -82,12 +82,14 @@ export function getTaskClasses (store) {
       case 'control':
         if (type === 'todo' || type === 'daily') {
           if (task.completed || !shouldDo(dueDate, task, userPreferences) && type === 'daily') {
-            return {
-              bg: 'task-disabled-daily-todo-control-bg',
-              checkbox: 'task-disabled-daily-todo-control-checkbox',
-              inner: 'task-disabled-daily-todo-control-inner',
-              content: 'task-disabled-daily-todo-control-content',
-            };
+            if (!(loc === 'challenge')) {
+              return {
+                bg: 'task-disabled-daily-todo-control-bg',
+                checkbox: 'task-disabled-daily-todo-control-checkbox',
+                inner: 'task-disabled-daily-todo-control-inner',
+                content: 'task-disabled-daily-todo-control-content',
+              };
+            }
           }
 
           return {
--- a/test/client/unit/specs/store/getters/tasks/getTaskClasses.js
+++ b/test/client/unit/specs/store/getters/tasks/getTaskClasses.js
@@ -1,7 +1,7 @@
 import generateStore from 'client/store';
 
 describe('getTaskClasses getter', () => {
-  let store, getTaskClasses;
+  let store, getTaskClasses, getDisabledClassObject;
 
   beforeEach(() => {
     store = generateStore();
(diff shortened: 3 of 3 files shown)


website/client/store/getters/tasks.js
```diff
@@ -82,12 +82,14 @@ export function getTaskClasses (store) {
       case 'control':
         if (type === 'todo' || type === 'daily') {
           if (task.completed || !shouldDo(dueDate, task, userPreferences) && type === 'daily') {
-            return {
-              bg: 'task-disabled-daily-todo-control-bg',
-              checkbox: 'task-disabled-daily-todo-control-checkbox',
-              inner: 'task-disabled-daily-todo-control-inner',
-              content: 'task-disabled-daily-todo-control-content',
-            };
+            if (!(loc === 'challenge')) {
```
negue, line 85: this could be `loc !== 'challenge'` ^^

website/client/components/tasks/task.vue
```diff
@@ -598,13 +598,13 @@ export default {
       return false;
     },
     controlClass () {
-      return this.getTaskClasses(this.task, 'control', this.dueDate);
+      return this.getTaskClasses(this.task, 'control', this.dueDate, this.$route.name);
```
negue, line 601: I think this `this.$route.name` should come from a prop of `task.vue` itself so that its not based on the route

negue (changes requested):
I think using the route to determine the list to change the colors isn't really the right way, like what if we get another view which is on the same route as tasks just on a different modal or something then this would break it right away, I'd suggest to use a fix enum (or string) on each task list

Mira-M:
> I think using the route to determine the list to change the colors isn't really the right way, like what if we get another view which is on the same route as tasks just on a different modal or something then this would break it right away, I'd suggest to use a fix enum (or string) on each task list

Thanks for the feedback.  Just a bit of clarification, are you saying that we shouldn't use the origin in general to determine the styles, or that the origin shouldn't be populated from `this.$route`?

> I'd suggest to use a fix enum (or string) on each task list

Are you saying that the origin (or some flag of some sort) should be added to the task list itself and then use that to decide which styles to be applied?  I ask because this was my original approach and there were other side effects of doing this where for some reason Todos were being stubborn and the flag wasn't toggling, but this was not the case with dailies.

> like what if we get another view which is on the same route as tasks just on a different modal or something then this would break it right away

I considered this, and thought that since I was ignoring any changes not related to challenges it would be ok, I also added unit tests that should catch if the styles were not as expected.  Next time I'll ask if the approach would be sufficient 🦄 

I apologize in advance for the "duh" questions =^.^=

Nonetheless, I can return to that approach and report any specifics I'm seeing in that behavior instead.  I'll be able to get to it this weekend. Thanks for the info! 😻

negue:
> > I think using the route to determine the list to change the colors isn't really the right way, like what if we get another view which is on the same route as tasks just on a different modal or something then this would break it right away, I'd suggest to use a fix enum (or string) on each task list
> 
> Thanks for the feedback. Just a bit of clarification, are you saying that we shouldn't use the origin in general to determine the styles, or that the origin shouldn't be populated from `this.$route`?

Sorry for the cryptic responds^^, I meant to say that using the route could lead to potential weird bugs in the long run, using an enum/string would be a bit 'stronger' against that (see the answer below)

> 
> > I'd suggest to use a fix enum (or string) on each task list
> 
> Are you saying that the origin (or some flag of some sort) should be added to the task list itself and then use that to decide which styles to be applied? I ask because this was my original approach and there were other side effects of doing this where for some reason Todos were being stubborn and the flag wasn't toggling, but this was not the case with dailies.

Yes, a fixed source added by code (instead of the variable route) should be the way to go. I mean you can use the current code (well with a bit more around it) and then it should work as it does right now.

Maybe this side effect happened of something else? Hard to guess anything without the code back-then

> 
> > like what if we get another view which is on the same route as tasks just on a different modal or something then this would break it right away
> 
> I considered this, and thought that since I was ignoring any changes not related to challenges it would be ok, I also added unit tests that should catch if the styles were not as expected. Next time I'll ask if the approach would be sufficient 🦄

The test is already correct, you just need to give this 'challenge' flag to the outer task-list itself and then this is good to go ^^

Alys:
As @Mira-M says at https://github.com/HabitRPG/habitica/issues/10907#issuecomment-502362635 this PR is available for someone else to pick up.

paglias:
I'm going to close this one.

PR #10977, opened 2019-02-06
Track specific items dropped by Armoire

SabreCat:

### Changes

**Before**, users' purchases of the Enchanted Armoire showed up in logs merely as a gold purchase of the Armoire, with no indication of what the user obtained from the random roll. This proved unhelpful during troubleshooting, when a user might e.g. complain that their item didn't appear on the client, but we had no way of verifying what the server gave them.

**Now**, if a user's Armoire click produces a food or gear item, this will additionally log a "dropped item" event, including an "Armoire" method and the key of the specific item dropped. If such an event does not accompany an Armoire purchase event, by process of elimination we know that the user got an Experience payout instead.

### DIFF
--- a/website/common/script/ops/buy/buyArmoire.js
+++ b/website/common/script/ops/buy/buyArmoire.js
@@ -68,6 +68,19 @@ export class BuyArmoireOperation extends AbstractGoldItemOperation {
     ];
   }
 
+  _trackDropAnalytics (userId, key) {
+    this.analytics.track(
+      'dropped item',
+      {
+        uuid: userId,
+        itemKey: key,
+        acquireMethod: 'Armoire',
+        category: 'behavior',
+        headers: this.req.headers,
+      },
+    );
+  }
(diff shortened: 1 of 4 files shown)


website/common/script/ops/buy/buyArmoire.js
```diff
@@ -89,6 +102,8 @@ export class BuyArmoireOperation extends AbstractGoldItemOperation {
 
     removeItemByPath(user, `gear.flat.${drop.key}`);
 
+    this._trackDropAnalytics(user._id, drop.key);
```
paglias: looks like some tests are failing, maybe this should be inside an `if (this.analytics) {`? (Haven't tested it)

SabreCat: Oh, whoops, I fixed this in one place but not the other!

paglias:
LGTM

PR #10978, opened 2019-02-08
catch promise rejection when user submits a message to group. fixes #10970

### LINKED ISSUE #10970: Profanity blocker keeps blocking even after blocked word removed
_@thebat137 (dea8a9e8-c084-438a-9a14-7fe96a73fd0a): "I attempted to post a message in a public guild (New Year's Resolutions) that inadvertently contained a banned word, and got the usual warning telling me to edit my message.  So I edited my message and tried to hit "send", but the warning remained in place even though the banned word was gone.  I couldn't figure out how to get the edit detected correctly, so eventually I had to copy my message to a scratch document, refresh the page, and then copy the message back into the chat pane.  It posted correctly after this procedure, without further editing, so I know that the banned word was definitely already gone."_

I've tested and confirmed this.

ghost:
# #10970 
Fixes #10970 

### Changes
Add a catch block to the message dispatch function & access the response on success only to prevent an uncaught rejection blocking the rest of the sendMessage function. 

----
UUID: 4fb08296-55c0-4d21-9a13-a33a9a40cf98


paglias:
hi @bonnieetc ! thanks for the PR

When using `async/await` instead of `.catch()` like normal PRs it would be better to use a `try {} catch {}` block

Also, does this change still show the error message when a profanity is posted?

ghost:
>hi @bonnieetc ! thanks for the PR ...

@paglias thanks! It does still show the error message to the user, yeah. I wasn’t sure how to implement a try/catch block, as the exception seems to be handled elsewhere (the error message still displaying). 

Would an empty catch block suffice, or is there something else that should be done when the rejection is caught? 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

paglias:
@bonnieetc sorry for the super late response!

One wait to use a try/catch would be like this

```
try {
    let response = await ...
} catch {
    // empty catch block to let the function continue after an error and cleanup the input
}
```

ghost:
No worries at all, @paglias, thanks for getting back to me!

Thank you! I implemented your code with a slight tweak - I defined `response` outside the try block so it doesn't fall out of scope.

paglias:
this is looking good! Will be merged on friday

SabreCat:
Thanks for this, @bonnieetc! It'll go into a staging environment for a few days, then if all is well go live around this time next week.

Welcome to the ranks of the Blacksmiths at tier 1! You'll see your rewards on Habitica when next you sync there. Note that successive tiers require increasing amounts of effort from one to the next, so tier 2 will require multiple PRs or a larger PR to attain. But keep helping out and we'll show our gratitude accordingly!

PR #10979, opened 2019-02-08
Enable Username to be searched in Hall of Heroes - fixes #10972 

### LINKED ISSUE #10972: Username should be searchable in Hall of Heroes
When moderators view the [Hall of Heroes](https://habitica.com/hall/contributors) (`website/client/components/hall/heroes.vue`) they have a search form to search for any user (not just a contributor) by User ID. See below for a screenshot.

That search feature should also accept Usernames (not Display Names).
 
The placeholder text in the search form should be changed from "User ID" to "Username or User ID".
A new text string will need to be created for that but it should NOT be put in the `habitica/website/common/locales` files because it won't need to be translated (admin-only features are just displayed in English). 
_There's a file somewhere else for storing such strings but I can never remember where. :scream: I'll update this issue later when I find it again. If I haven't done that by the time a contributor wants to work on this, ask here. There's website/common/errors/commonErrorMessages.js but that seems to be just for errors; I think there's another file like it somewhere else._

A contributor can see that search feature on their local install by changing their test user to an admin. One way to do that is with the "Make Admin" button in the Debug menu in the footer of the test site.

image

ChesterSng:
[//]: #10972 (Put Issue # here, if applicable. This will automatically close the issue if your PR is merged in)
Fixes #10972

**In Progress**

### Changes
Added a check or username first before checking for UUID.

----

### DIFF
--- a/website/server/controllers/api-v3/hall.js
+++ b/website/server/controllers/api-v3/hall.js
@@ -164,17 +164,29 @@ api.getHero = {
   async handler (req, res) {
     let heroId = req.params.heroId;
 
-    req.checkParams('heroId', res.t('heroIdRequired')).notEmpty().isUUID();
+    req.checkParams('heroId', res.t('heroIdRequired')).notEmpty();
 
     let validationErrors = req.validationErrors();
     if (validationErrors) throw validationErrors;
 
     let hero = await User
-      .findById(heroId)
+      .findOne({'auth.local.username': heroId})
       .select(heroAdminFields)
       .exec();
 
-    if (!hero) throw new NotFound(res.t('userWithIDNotFound', {userId: heroId}));
+    if(!hero) {
+      req.checkParams('heroId', res.t('heroIdRequired')).notEmpty().isUUID();
+      validationErrors = req.validationErrors();
+      if (validationErrors) throw validationErrors;
+
+      hero = await User
+        .findById(heroId)
+        .select(heroAdminFields)
+        .exec();
+
+      if (!hero) throw new NotFound(res.t('userWithIDNotFound', {userId: heroId}));
+    }
+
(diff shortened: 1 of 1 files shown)


Alys:
In reply to [your questions](https://github.com/HabitRPG/habitica/issues/10972#issuecomment-461748488) from the issue:

One useful way to troubleshoot when you can't work out what's gone wrong is to go back to the `develop` code and test with that. Assuming that works correctly (it should), add back your changed code in the smallest increments possible and test after each step. You'd have to choose which small bits of new code could be added without knowingly introducing a bug. 

You might also want to check that you haven't accidentally made changes to the vue files.

I'll post a couple of comments on the code too.

If this isn't enough to help you work out what's wrong, tell us and then we might be able to take a closer look at your new code.

website/server/controllers/api-v3/hall.js
```diff
@@ -164,17 +164,29 @@ api.getHero = {
...
 
     let hero = await User
-      .findById(heroId)
+      .findOne({'auth.local.username': heroId})
       .select(heroAdminFields)
       .exec();
 
-    if (!hero) throw new NotFound(res.t('userWithIDNotFound', {userId: heroId}));
+    if(!hero) {
+      req.checkParams('heroId', res.t('heroIdRequired')).notEmpty().isUUID();
```
Alys, line 178: The `notEmpty` check here isn't necessary since you've done one above.

However you might be able to use the `isUUID` function to simplify the code. If the specified search term is a UUID, you can perform one type of search, otherwise you perform the other type.  You'd then need one `if(!hero)` section after both searches and you'd need only one set of `validationErrors` lines instead of two.

ChesterSng:
Hi @Alys thanks for your comment! I have opened another PR #10980 after branching off `develop` again.

PR #10980, opened 2019-02-09
Enable Username to be searched in Hall of Heroes - fixes #10972

### LINKED ISSUE #10972: Username should be searchable in Hall of Heroes

ChesterSng:
[//]: #10972 (Put Issue # here, if applicable. This will automatically close the issue if your PR is merged in)
Fixes #10972
- [x] Need to change error displayed and placeholder text for search box
- [x] Need to change test to expect difference response
- [x] Add test case for username input

### Changes
Added an if-else block that searches for username when `.isUUID()` fails.

----

### DIFF
--- a/website/server/controllers/api-v3/hall.js
+++ b/website/server/controllers/api-v3/hall.js
@@ -6,6 +6,7 @@ import {
 } from '../../libs/errors';
 import _ from 'lodash';
 import apiError from '../../libs/apiError';
+import validator from 'validator';
 
 let api = {};
 
@@ -162,30 +163,25 @@ api.getHero = {
   url: '/hall/heroes/:heroId',
   middlewares: [authWithHeaders(), ensureAdmin],
   async handler (req, res) {
-    let heroId = req.params.heroId;
     let validationErrors;
-    let hero;
-
     req.checkParams('heroId', res.t('heroIdRequired')).notEmpty();
 
     validationErrors = req.validationErrors();
     if (validationErrors) throw validationErrors;
 
-    req.checkParams('heroId', res.t('heroIdRequired')).isUUID();
+    const heroId = req.params.heroId;
 
-    validationErrors = req.validationErrors();
-
-    if (validationErrors) {
-      hero = await User
-        .findOne({'auth.local.username': heroId})
-        .select(heroAdminFields)
-        .exec();
+    let query;
+    if (validator.isUUID(heroId)) {
+      query = {_id: heroId};
     } else {
-      hero = await User
-        .findById(heroId)
-        .select(heroAdminFields)
-        .exec();
+      query = {'auth.local.username': heroId};
     }
+
+    const hero = await User
+      .findOne(query)
+      .select(heroAdminFields)
+      .exec();
 
     if (!hero) throw new NotFound(res.t('userWithIDNotFound', {userId: heroId}));
     let heroRes = hero.toJSON({minimize: true});
--- a/website/server/controllers/api-v3/hall.js
+++ b/website/server/controllers/api-v3/hall.js
@@ -142,7 +142,7 @@ api.getHeroes = {
 const heroAdminFields = 'contributor balance profile.name purchased items auth flags.chatRevoked';
 
 /**
- * @api {get} /api/v3/hall/heroes/:heroId Get any user ("hero") given the UUID
+ * @api {get} /api/v3/hall/heroes/:heroId Get any user ("hero") given the UUID or Username
  * @apiParam (Path) {UUID} heroId user ID
  * @apiName GetHero
  * @apiGroup Hall
--- a/website/server/controllers/api-v3/hall.js
+++ b/website/server/controllers/api-v3/hall.js
@@ -173,7 +173,7 @@ api.getHero = {
 
     let query;
     if (validator.isUUID(heroId)) {
-      query = {'_id': heroId};
+      query = {_id: heroId};
     } else {
       query = {'auth.local.username': heroId};
     }
--- a/website/server/controllers/api-v3/hall.js
+++ b/website/server/controllers/api-v3/hall.js
@@ -163,16 +163,27 @@ api.getHero = {
   middlewares: [authWithHeaders(), ensureAdmin],
   async handler (req, res) {
     let heroId = req.params.heroId;
+    let validationErrors;
+    let hero;
 
-    req.checkParams('heroId', res.t('heroIdRequired')).notEmpty().isUUID();
+    req.checkParams('heroId', res.t('heroIdRequired')).notEmpty();
 
-    let validationErrors = req.validationErrors();
+    validationErrors = req.validationErrors();
     if (validationErrors) throw validationErrors;
 
-    let hero = await User
-      .findById(heroId)
-      .select(heroAdminFields)
-      .exec();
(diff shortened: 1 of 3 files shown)


ChesterSng:
Hi @Alys after branching off `develop` again I can successfully search for the User ID and see the intended effect after making the code changes. However, the username still returns 404. I have used the same search query in `mongo` and it successfully returns the user object. 

Edit: This is now resolved.

ChesterSng:
@Alys can u point me to the file to change the placeholder text as mentioned in the issue #10972 ? Thanks!

Alys:
@paglias Which file is best for adding a string that doesn't need to be translated because it's for an admin-only feature? Here's what I wrote in the issue about that:

> A new text string will need to be created for that but it should NOT be put in the `habitica/website/common/locales` files because it won't need to be translated (admin-only features are just displayed in English).
> There's a file somewhere else for storing such strings but I can never remember where. scream I'll update this issue later when I find it again. If I haven't done that by the time a contributor wants to work on this, ask here. There's website/common/errors/commonErrorMessages.js but that seems to be just for errors; I think there's another file like it somewhere else.

ChesterSng:
We just need to know where the 'UUID' is stored right?
https://github.com/HabitRPG/habitica/blob/ef412c7185a82f2e2b5a3fdc776a60380a35bebd/website/client/components/hall/heroes.vue#L11

Alys:
@ChesterSng Not exactly. :)  For normal text (anything visible to regular players), finding where a similar string is stored is a good way to choose a location for a new string, but in this case the "UUID" string is used in pages visible to all players (and so needs to be translated) but the new string is only for mods so it shouldn't be translated (that would give unnecessary work to the translators).

However in case it's useful in future, you can find where the "UUID" string is stored (or any other translated string) by going to the `website/common/locales/en` directory (which contains all the English language strings) and doing `git grep "UUID"`. You'll see the results below and the line that has `"UUID"` to the left of the colon is where the UUID string is defined.
```
website/common/locales/en $ git grep "UUID"
challenge.json:84:  "challengeIdRequired": "\"challengeId\" must be a valid UUID.",
challenge.json:85:  "winnerIdRequired": "\"winnerId\" must be a valid UUID.",
communityGuidelines.json:88:  "commGuidePara060B": "If your account is banned (a severe consequence), you will not be able to log into Habitica and will receive an error message upon attempting to log in. <strong>If you wish to apologize or make a plea for reinstatement, please email the staff at <a href='mailto:admin@habitica.com' target='_blank'>admin@habitica.com</a> with your UUID</strong> (which will be given in the error message). It is <strong>your</strong> responsibility to reach out if you desire reconsideration or reinstatement.",
contrib.json:49:  "UUID": "User ID",
contrib.json:53:  "invalidUUID": "UUID must be valid",
front.json:213:  "useUUID": "Use UUID / API Token (For Facebook Users)",
front.json:292:  "memberIdRequired": "\"member\" must be a valid UUID.",
front.json:293:  "heroIdRequired": "\"heroId\" must be a valid UUID.",
settings.json:165:    "invalidWebhookId": "the \"id\" parameter should be a valid UUID.",
spells.json:75:  "targetIdUUID": "\"targetId\" must be a valid User ID.",
tasks.json:156:  "itemIdRequired": "\"itemId\" must be a valid UUID.",
tasks.json:158:  "tagIdRequired": "\"tagId\" must be a valid UUID corresponding to a tag belonging to the user.",
```

ChesterSng:
Ahh okay that's very useful, thanks @Alys!

paglias:
@Alys @ChesterSng regarding

> @paglias Which file is best for adding a string that doesn't need to be translated because it's for an admin-only feature? Here's what I wrote in the issue about that:

In this case it should go directly in the .vue file as a standard string because for the client we don't have a place to store strings which aren't translated

website/server/controllers/api-v3/hall.js
```diff
@@ -163,16 +163,29 @@ api.getHero = {
...
 
-    let validationErrors = req.validationErrors();
+    validationErrors = req.validationErrors();
     if (validationErrors) throw validationErrors;
 
-    let hero = await User
-      .findById(heroId)
-      .select(heroAdminFields)
-      .exec();
+    req.checkParams('heroId', res.t('heroIdRequired')).isUUID();
```
negue: I think the double validationError checking seems to be a bit off.

you could keep the first one with `notEmpty` and then test the param with `validator.isUUID()` without using `validationErrors()` again

paglias: agreed

ChesterSng: Sorry I'm not really familar with the methods here...

`    if (req.checkParams('heroId', res.t('heroIdRequired')).isUUID() === true) {
      hero = await User
        .findById(heroId)
        .select(heroAdminFields)
        .exec();
    } else {
      hero = await User
        .findOne({'auth.local.username': heroId})
        .select(heroAdminFields)
        .exec();
    }`
By doing this, the code path always goes to the else block. And thus searching by user id will not work.

paglias: I think what @negue is suggesting is that `req.validationErrors();` is only called once, so the second check can be done with the `validator` library, something like this:

```
import validator from 'validator';
...
req.checkParams('heroId', res.t('heroIdRequired')).notEmpty();
const validationErrors = req.validationErrors();
....
const heroParam = req.params.heroId;

let query;
if (validator.isUUID(heroParam)) {
    query = {'_id': heroParam};
} else {
    query = 'auth.local.username': heroId};
}

await User.findOne(query).select(...).exec()
...
```

it should work but I haven't tested this code so there are likely errors

ChesterSng: It seems like `.isUUID()` does not return `true` or `false`

paglias: what does it return? if you search for `validator.isUUID` in the code you'll see it's used in many places and it returns true or false

ChesterSng: Ahh okay I tried it. I was confused because `params.isUUID()` did not return `true` or `false`, but I think the result is given in `.validationErrors()` instead. 

`validator.isUUID()` returns true or false.

Thanks for the help @paglias!

SabreCat:
Thanks for this, @ChesterSng! This will go into a staging environment for live testing, then go into production next week. Credited toward your next contributor tier!

PR #10982, opened 2019-02-10
fix mention highlight 

### LINKED ISSUE #10875: Email address display breaks

negue:
Second try: fixes #10875 

- now only `@mentions` are highlighted
- everything inside an "email" isn't highlighted, content or href


paglias:
this seems to be working great now!

PR #10984, opened 2019-02-10
Update conditional render to only show Apply Sort... text in party modals. fixes #10968 

### LINKED ISSUE #10968: "Apply Sort Options to Party Header" should not appear in Guilds
As reported by @ravenlune (9272ab2c-4b66-4b74-aebd-799725885d1c):

"... when I go to a Guild page and check the member list. After I choose a different sorting method from the drop-down menu, the "Apply Sort Options to Party Header" button will show up. Said button will then stay until I refresh my browser. If I click it, my party header will be sorted according to the method I picked. But I'm assuming this isn't actually the desired behaviour, since I'm on a Guild list, not my party list."

The correct behaviour is for that "Apply..." link to appear only when you're sorting while viewing the member list for your Party.

CBillings92:

fixes #10968 

### Changes
I've gone ahead and modified the conditional render `v-if` code in `membersModal.vue` to only render the "Apply Sort Options to Party Header" link button if the current group is of type 'party'. This group data was already accessible, so no changes needed to be changed to get new data to the page. 

I've manually tested locally by creating two user accounts, a party, and a guild, making sure that the Apply Sort Options to Party Header link only displays after changing the filter options within a party.

I ran the automated tests (using `npm run test`) on both develop and my new branch and I've attached images of the results. Looks like there are some failing tests on `develop` that my branch also encounters.

Let me know if there's anything else I need to do or am missing on this issue!

----
new-branch-failed-tests
failed_tests_develop

UUID: 003b1968-38ba-4bcc-9cf6-25926c4d5f11

### DIFF
--- a/website/client/components/groups/membersModal.vue
+++ b/website/client/components/groups/membersModal.vue
@@ -20,7 +20,7 @@ div
           select.form-control(@change='changeSortDirection($event)')
             option(v-for='sortDirection in sortDirections', :value='sortDirection.value') {{sortDirection.text}}
 
-    .row.apply-options.d-flex.justify-content-center(v-if='sortDirty')
+    .row.apply-options.d-flex.justify-content-center(v-if='sortDirty && group.type === "party"')
       a(@click='applySortOptions()') {{ $t('applySortToHeader') }}
     .row(v-if='invites.length > 0')
       .col-6.offset-3.nav


SabreCat:
Thanks for this, @CBillings92! It'll go into a staging environment now for live testing, then if all is well move to production next week.

Welcome to the ranks of the Blacksmiths at tier 1! Note that further tiers require increasing amounts of effort from one to the next, so tier 2 will take multiple PRs or a larger PR to attain. But keep helping out and we'll express our gratitude accordingly!

PR #10988, opened 2019-02-12
Set registered platform on analytics only once

SabreCat:

### Changes

**Before**, the `registeredPlatform` analytics property was considered mutable, just like any other user property. This could lead to unintuitive results, where a user who registered under one platform would later have that "registered platform" updated to something else, or worse, have the property removed from their record.

**Now**, when we set this property, we use the [$setOnce key of the identify method](https://amplitude.zendesk.com/hc/en-us/articles/205406617-Identify-API-Modify-User-Properties#keys-for-the-identification-argument) to set the value and leave it read-only thereafter.

**Before**, the `AMPLITUDE_KEY` constant in the server analytics library had a typo, `AMPLIUDE_KEY`.
**Now**, all instances read `AMPLITUDE_KEY`, as intended.

### DIFF
--- a/website/server/libs/analyticsService.js
+++ b/website/server/libs/analyticsService.js
@@ -10,7 +10,7 @@ import {
 } from 'lodash';
 import { content as Content } from '../../common';
 
-const AMPLIUDE_TOKEN = nconf.get('AMPLITUDE_KEY');
+const AMPLITUDE_TOKEN = nconf.get('AMPLITUDE_KEY');
 const GA_TOKEN = nconf.get('GA_ID');
 const GA_POSSIBLE_LABELS = ['gaLabel', 'itemKey'];
 const GA_POSSIBLE_VALUES = ['gaValue', 'gemCost', 'goldCost'];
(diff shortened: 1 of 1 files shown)


website/server/libs/analyticsService.js
```diff
@@ -271,11 +267,28 @@ let _sendPurchaseDataToGoogle = (data) => {
   });
 };
 
+let _setOnce = (data) => {
+  return new Promise((resolve, reject) => {
```
paglias: LGTM, just one thing here, it looks like `amplitude.identify` already returns a promise, why is it wrapped in another promise?

SabreCat: oh, I was just following the patterns we had elsewhere--let me try it without the additional wrapper

website/server/libs/analyticsService.js
```diff
@@ -271,11 +267,24 @@ let _sendPurchaseDataToGoogle = (data) => {
   });
 };
 
+let _setOnce = (data) => {
+  return amplitude.identify({
+    user_properties: {
+      $setOnce: data,
+    },
+  });
```
SabreCat, line 275: Better @paglias ?

paglias: 👍

website/server/libs/analyticsService.js
paglias: Yeah that wrapper is needed when the method is using callbacks but in this case it's already returning a promise

PR #11013, opened 2019-02-20
fix header design

negue:
Requested Changes
- Health, Experience, and Mana bars should have a border-radius of 1px
- Fix placement of Health, Experience, and Mana bar tooltips (2px under the bar)

In order to change the tooltip position, its only possible to change by the tooltip-class (no ability to add a custom class)

I extracted the progress and tooltip to a custom component, which then could be styled within the container


paglias:
schermata 2019-03-03 alle 17 20 31

looks like there's a minor problem here with the tooltips appearing under the header

negue:
thanks for catching that one, it was a scss-scope issue

paglias:
thanks!

Tressley:
These LGTM. 👍

PR #11030, opened 2019-02-26
Ensure official challenges are listed first (fixes #11018)

### LINKED ISSUE #11018: Official Challenges not sorting to top
### Description
Challenges marked "Habitica Official", such as the monthly Take This challenges, should show up at the top of the "Discover Challenges" view. Currently they aren't--you have to either filter on the category or do some very painful manual scrolling to find them.

#### Console Errors
None relevant in evidence.

alecbrick:

Fixes #11018

### Changes
When fetching challenges from the database, we were originally querying a page of 10 challenges, _then_ sorting the official challenges to the top of those 10. This pull request changes the database query to an aggregate, which will determine whether a challenge has the "habitica_official" slug, then sort based on that result.

----


website/server/controllers/api-v3/challenges.js
```diff
@@ -386,39 +386,52 @@ api.getUserChallenges = {
       query.categories = { $elemMatch: { slug: {$in: categorySlugs} } };
     }
 
-    let mongoQuery = Challenge.find(query)
-      .sort('-createdAt');
+    // Ensure that official challenges are always first
+    let mongoQuery = Challenge.aggregate()
```
negue: could you extract the Challenge-query to a function? that way if we need to fix this query we won't have to change two queries

alecbrick: Done.

website/server/controllers/api-v3/challenges.js
```diff
@@ -386,39 +387,31 @@ api.getUserChallenges = {
...
     // Instead of populate we make a find call manually because of https://github.com/Automattic/mongoose/issues/3833
-    await Promise.all(resChals.map((chal, index) => {
+    await Promise.all(challenges.map((chal, index) => {
       return Promise.all([
         User.findById(chal.leader).select(nameFields).exec(),
         Group.findById(chal.group).select(basicGroupFields).exec(),
       ]).then(populatedData => {
-        resChals[index].leader = populatedData[0] ? populatedData[0].toJSON({minimize: true}) : null;
-        resChals[index].group = populatedData[1] ? populatedData[1].toJSON({minimize: true}) : null;
+        challenges[index].leader = populatedData[0] ? populatedData[0].toJSON({minimize: true}) : null;
```
paglias: is this working correctly? The reason why we first serialized the challenges in plain json objects and then applied the leader and group data is to avoid mongoose removing the changes since they aren't strings as in the schema

website/server/libs/challenges/index.js
```diff
@@ -103,3 +103,29 @@ export function cleanUpTask (task) {
 
   return cleansedTask;
 }
+
+// Create an aggregation query for querying challenges.
+// Ensures that official challenges are listed first.
+export function createChallengeQuery (query) {
```
paglias, line 115: thanks @alecbrick , we're currently having some perf issues with this query so it might take a bit before this is merged. I was wondering if it's possible to get the same result without using the aggregation framework as its purpose is to aggregate multiple docs into a single one and using it like this is a bit outside its scope. But apart from this I like that we're switching from client side sorting to something on the db's side

alecbrick: As far as I can tell, there’s no way to sort by whether a challenge has the “is_official” slug when we’re running normal queries on the database. If we brought back the “official” property, that would work. It might be redundant information, but it makes queries easier.

website/server/controllers/api-v3/challenges.js
alecbrick: I think so. aggregate() returns plain JSON objects, so we don’t have to call toJSON() on the results.

paglias: Hmm does this means defaults aren't applied as well by Mongoose

alecbrick: That’s a good question. I can check it out, if you want.

paglias: Yes that would be useful, thanks. In fact if a plain object is returned we'd have to go back to the server side sorting because we rely on the data from the db passing through mongoose to be sure private data is not shared, defaults are applied correctly, etc...

alecbrick: I'm looking at this right now, but I'm not exactly sure what I should be looking for. Is there something in particular I can do to check whether defaults are being applied?

I'm also not super sure what the issue is regarding the plain object. The Mongoose [docs](https://mongoosejs.com/docs/api.html#aggregate_Aggregate) say that a plain object is returned instead of a document "since any shape of document can be returned". I think the data still passes through Mongoose normally.

paglias:
That means that mongoose returns directly what's given to it from mongodb
and unfortunately that's not possible because this way all the schema
options aren't applied. We'll have to use the server side sorting as we did
previously

Il Lun 4 Mar 2019, 13:47 Alec Brickner <notifications@github.com> ha
scritto:

alecbrick:
Confirmed that aggregate() isn't applying defaults. 

The issue is that we don't really have a "server side sorting" to go back to - that code was sorting the 10 results returned from the database, which usually didn't include official challenges.

paglias:
Hmm I see, so the problem is that with pagination it becomes difficult to load official challenges before normal ones...

I guess one solution could be:

1. We make a query to load the official challenges only
2. if there are less than 10 or if we're not using pagination then we load the normal challenges

But it becomes a bit complex when there are more than 10 official challenges

Otherwise we could take the results from `.aggregate()`, pass them to `new Challenge()`to make sure all defaults are applied correctly and then serialize that (@alecbrick in that cause you should save isOfficial somewhere and then re-add it to the result of `new Challenge()`)

alecbrick:
That could work. Are the performance issues with aggregate() significant?

SabreCat:
If we need to keep the list down to <10 official challenges as an administrative rule to help this work, that's doable. We have a bunch of them scattered around right now, but most of them are "archival" versions of past Challenges that don't really need to be sorted to the top, so we can take the category off of them.

I'm not entirely opposed to restoring use of the `official` db field either, though it'd be best if it were controlled through the category list rather than as a separate checkbox.

paglias:
I feel like remembering to only have 10 official challenges or less is something very error prone so I'd prefer to avoid it.

> I'm not entirely opposed to restoring use of the official db field either, though it'd be best if it were controlled through the category list rather than as a separate checkbox.

is there a specific reason for this? We could keep both the category and the boolean flag maybe?

> Are the performance issues with aggregate() significant?

I haven't investigated that yet but I don't think it'll be an issue and actually if we can offload some calculations from the server to the db even better

SabreCat:
Official flag controlled through category list: for ease of use / UX, mainly. It's not a huge deal because it's an administrative function, but I like having it part of the workflow when setting up Challenge categories, without an extra step or click. Having both the boolean and the category value in the record is fine!

alecbrick:
If aggregate() doesn't cause performance issues, it might just be easier to pass the objects from the query into Challenge objects to ensure defaults are applied instead of worrying about handling the boolean flag.

paglias:
Hi @alecbrick ! Do you have any update on the status of this PR? thanks!

alecbrick:
Not yet! Sorry, school got a little busy around this time, and I never quite got back to this PR. If we're alright with just passing the results from aggregate() into Challenge objects, then I'll just go for that solution.

paglias:
@alecbrick no worries! and yes I think passing aggregate results into Challenge to apply the defaults is a good solution

alecbrick:
I'm back! I'll try to have something up by tomorrow.

alecbrick:
@paglias Alright, I think this is (finally) ready for review now!

website/server/libs/challenges/index.js
paglias: @SabreCat @Alys since it's been a while, do you remember what type of perf issues we had here?

paglias (commented):
Left a comment but this is working well and will be merged soon

website/server/libs/challenges/index.js
Alys: Sorry, no.

paglias:
thanks @alecbrick ! We're going to test this for a few days but if we don't find any perf issue it'll go live by the end of the week!

paglias:
And noted towards your 3rd tier

paglias:
Had to add a small fix for some old challenges that don't have any category https://github.com/HabitRPG/habitica/commit/276e882092974da16dbe5416f59fb9300aa4d0a4

alecbrick:
Whoops. Nice catch!

PR #11031, opened 2019-02-26
Multiple checks for pinnedItems

### LINKED ISSUE #10134: Entire rewards column missing

### General Info
  * UUID: 
  * Browser: Chrome Version 64.0.3282.186 (Official Build) (64-bit)
  * OS: Windows 10 Home (Version 1709) (OS Build 16299.309)

### Description

Relaunched the website after restarting my computer, and the entire rewards column failed to appear.

image

#### Console Errors

image

The problem seems to be this error:

    vue.runtime.esm.js:1713 TypeError: Cannot read property 'key' of undefined
    at e.exports (app.78dc0f64afd5fc1a7fd8.js:1)
    at app.78dc0f64afd5fc1a7fd8.js:1
    at Array.map (<anonymous>)
    at e.exports (app.78dc0f64afd5fc1a7fd8.js:1)
    at a.inAppRewards (8.ea5e7c1a746c2218feac.js:1)
    at Ci.get (vue.runtime.esm.js:3111)
    at Ci.evaluate (vue.runtime.esm.js:3218)
    at a.inAppRewards (vue.runtime.esm.js:3470)
    at a.initialColumnDescription (8.ea5e7c1a746c2218feac.js:1)
    at Ci.get (vue.runtime.esm.js:3111)

negue:
This adds some additional checks for the pinnedItems - issue occurred in #10134

- check the pinnedItems for null/undefined entries and remove them on toggle
- filters out undefined items when getting the reward-items

- also fixed the drawer error which was also in #10134 

@paglias @Alys any other ideas where that undefined could be from?

Fixes #10134

### DIFF
--- a/website/common/script/libs/inAppRewards.js
+++ b/website/common/script/libs/inAppRewards.js
@@ -4,6 +4,7 @@ import getOfficialPinnedItems from './getOfficialPinnedItems';
 import compactArray from 'lodash/compact';
 
 import getItemByPathAndType from './getItemByPathAndType';
+import {checkPinnedAreasForNullEntries} from '../ops/pinnedGearUtils';
 
 /**
  * Orders the pinned items so we always get our inAppRewards in the order
@@ -32,6 +33,8 @@ function sortInAppRewards (user, items) {
 }
 
 module.exports = function getPinnedItems (user) {
+  checkPinnedAreasForNullEntries(user);
+
   let officialPinnedItems = getOfficialPinnedItems(user);
 
   const officialPinnedItemsNotUnpinned = officialPinnedItems.filter(officialPin => {
@@ -41,11 +44,13 @@ module.exports = function getPinnedItems (user) {
 
   const pinnedItems = officialPinnedItemsNotUnpinned.concat(user.pinnedItems);
 
-  let items = pinnedItems.map(({type, path}) => {
-    let item = getItemByPathAndType(type, path);
+  let items = pinnedItems
+    .filter(item => Boolean(item)) // filter out nullable entries, which shouldn't be even possible
+    .map(({type, path}) => {
+      let item = getItemByPathAndType(type, path);
 
-    return getItemInfo(user, type, item, officialPinnedItems);
-  });
+      return getItemInfo(user, type, item, officialPinnedItems);
+    });
 
   shops.checkMarketGearLocked(user, items);
 
--- a/website/common/script/libs/inAppRewards.js
+++ b/website/common/script/libs/inAppRewards.js
@@ -45,7 +45,6 @@ module.exports = function getPinnedItems (user) {
   const pinnedItems = officialPinnedItemsNotUnpinned.concat(user.pinnedItems);
 
   let items = pinnedItems
-    .filter(item => Boolean(item)) // filter out nullable entries, which shouldn't be even possible
     .map(({type, path}) => {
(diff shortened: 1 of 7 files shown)


Alys:
For the cases of this bug that I've investigated, when I've checked the
user's pinned items in the database, I haven't found any with undefined or
inappropriate values. That could be because the problem fixes itself before
I investigate, but that seems unlikely and as you said in the issue,
there's nothing that should be inserting undefined/null entries.

it sort of looks as if the undefined entries happen only client-side,
possibly if data is not fully loaded from the database (due to a timeout or
something maybe). I.e., it might be that the user's `pinnedItems` array in
the database is correct, but the website doesn't get all the data from the
array. if that is what's happening, will this PR result in the user's
correct `pinnedItems` array being overwritten by an empty or partial array?

negue:
Yes, the added =>  `checkPinnedAreasForNullEntries` (in  https://github.com/HabitRPG/habitica/pull/11031/files#diff-b9ee32d8fb1fc0b4590ca444ef865dc8R36) cleans the array of undefined entries so that the toggled items will never have empty entries (which was error in #10134 )

The same `checkPinnedAreasForNullEntries` is also called every toggle.

Alys:
That feels like it might be dangerous since we don't know what causes the Rewards column bug and haven't found any direct evidence of incorrect `pinnedItems` entries in the database. If it turns out to be a client-side problem where good data is not being loaded correctly, we wouldn't want the good data to be removed from the database. Also, if there is bad data in the database, this would prevent us finding it so we'd have less chance of tracking down what's causing the bug. Is it possible to get the website to just ignore bad data when it's displaying the Rewards column rather than trying to remove it, at least until we have a better idea of what's causing the bug?

paglias:
Hmmm... this could be something similar to the issue we had with null/undefined notifications ending up in the db despite it being forbidden by the schema. In that case the checks @negue added would be the best solution as anything client side will probably need extra fixes on mobile apps while an API fix is global.

That being said I agree with @Alys and I'd like to first see a database entry with such null/undefined issues before merging this otherwise we risk the issue being something else

negue:
I removed the cleaning-up of pinnedItems on item-toggle (server-side) now only the client-/api-side call of getting the inAppRewards removes empty entries

website/common/script/libs/inAppRewards.js
```diff
@@ -41,11 +44,13 @@ module.exports = function getPinnedItems (user) {
 
   const pinnedItems = officialPinnedItemsNotUnpinned.concat(user.pinnedItems);
 
-  let items = pinnedItems.map(({type, path}) => {
-    let item = getItemByPathAndType(type, path);
+  let items = pinnedItems
+    .filter(item => Boolean(item)) // filter out nullable entries, which shouldn't be even possible
```
paglias: aren't this already filtered by the call to `checkPinnedAreasForNullEntries` above?

paglias:
Left one comment, apart from that it LGTM. When that's clarified we'll put this on delta for some extra testing

website/common/script/libs/inAppRewards.js
negue: yeah I had this one added  before I added the `checkPinnedAreasForNullEntries` , will be removed

paglias:
Merged to `delta`, if no issues arise next week it can be merged

PR #11032, opened 2019-02-27
Ensure profile page gets synced (fixes #11026)

### LINKED ISSUE #11026: Stats modal caches old data / reload needed before allocating point after level-up
When your Stats Points change, the User Icon > Stats modal does not show your current Stats Points until you reload the website with your browser's refresh button. A sync is not enough to make it show your current Stat Points (but it should be enough).

There's at least a couple of ways you can see the effects of this bug

---

### First way:
1. Make a note of what your Stat Points are.
2. Make a change to your Stats Points on the mobile app (or if you're using your local install, do a direct database command to change your `stats` object).
3. Click the website's sync button.
4. Go to User Icon > Stats - observe that it's still showing the Stat Point allocations from step 1. The correct behaviour is that it should show the change to the Stats Points that you made in step 2.
5. Reload the website.
6. Go to User Icon > Stats - observe that it's now showing the correct Stat Point allocations.

---

### Second way:

As reported by SheWhoCantKnit @shewhoknits (e1d43f13-4a7b-4188-8844-127b4f63c67c):

"I level up, get a notification that I have a stat point to allocate, but when I open the stats page via the notification popup, there's no point to allocate. If I check off another task, I get the message again with the same result.
I've found out by now that I have to reload the page (F5) and then open the stats page via the user menu to allocate the point."

The correct behaviour in this situation is that when you click on the notification menu's entry that says "You have [n] unallocated Stat Points", the User Icon > Stats modal should immediately show your current allocations and the new Stat Point. No sync should be needed in that case since the website should already be aware of your current Stats.

alecbrick:

Fixes #11026

### Changes
In profile.vue, the user in the store was not being directly used - rather, some modifications were being made to it upon load, then the modified user was used. As a result, store updates did not affect the user object, and so any syncs that updated the store did not affect the values of the user. This fix should update the modified user object whenever the user object in the store is updated.

----

### DIFF
--- a/website/client/components/userMenu/profile.vue
+++ b/website/client/components/userMenu/profile.vue
@@ -495,6 +495,9 @@ export default {
     async userId () {
       this.loadUser();
     },
+    userLoggedIn () {
+      this.loadUser();
+    },
   },
   methods: {
     async loadUser () {


paglias:
@alecbrick  thanks for this! I'm merging the PR and awarding your contributor tier.

Issue #10997, opened 2019-02-17
incorrect Pets (and maybe Mounts) displayed / starred in Stable


Alys:
There's client-side bugs in the Stable related to searching for and equipping pets.

It might affect mounts too. I haven't tested since I assume that if mounts are broken too, the pets fix would fix them as well. However any coder working on this might want to test mounts either before their fix or at least afterwards.

Here's a report by Praxis  @Praxissima (0ce6d43b-efa2-45a3-acaa-59daad7d658d) and below that is the results of my testing:

> Shortly after I purchased the Rose Quartz hatching potions and created some matching pets and mounts, I noticed a new bug with the stable selection interface, when using the search function (which is much needed once the Stable reaches a significant population). Since it was behaving perfectly before, maybe there is a connection between the two.
> 
> To illustrate the problem, here's what happened when I tested again just now:
> 
> Typed "cupid" in the search bar, to bring up all possible Cupid type pets in the main Stable graphical menu.
> 
> Selected Cupid wolf cub: got Purple wolf cub in Avatar window.
> 
> Typed/selected Shimmer wolf cub: same result.
> 
> Typed/selected Fairy wolf cub: same result.
> 
> Then I went back to Cupid wolf cub, same result.
> 
> Then I tried to select Cupid panda cub: got FAIRY panda cub.
> 
> From that point, ALL the pet selections went to Fairy type instead of the kind that I typed and I selected (i.e. correct species, wrong color/type). Then I noticed something in the main Stable menu window (in addition to the wrong pet displaying in the Avatar window):
> 
> Comparing the avatar display with the graphical menu in the main stable window, The Fairy lion cub (which I may have purposely chosen) was displaying in the Avatar window, but BOTH the Purple lion cub AND the Fairy lion cub showed (with a "star" in the upper left corner) in the main Stable menu as the currently selected pets.
> 
> Earlier, I found that it won't always even display the right species. I started having problems when the Rose Quartz flying piglet seemed to set every subsequent selection to flying pig, regardless of the species of pet selected, only in the selected color/type (wrong pet, right color).
> 
> In other words, it seems somewhat random, but with SOME consistent patterns, and it all started shortly after the introduction of the Rose Quartz pets (at least that gives you a time frame to check on recent adjustments to the Stable code, in case it's not directly related to the Rose Quartz pets).
> 
[...]

PR #11041, opened 2019-03-02
Fix mismatch between pet and star badge

alecbrick:

### Changes
In #10997, some people reported that incorrect pets are equipped if you click on them while searching. It looks like this was due to the incorrect variable being referenced on the star badges - they reference "item", which is a variable from outside the slot's scope, but they should be referencing "context.item", which is the item that the star badge should have access to. To be quite honest, I'm not exactly sure why referencing "item" led to the observed behavior - Vue has a lot of moving parts, and I'm not 100% sure how they all fit together. However, this fix seems to have done the trick.

I have not been able to reproduce all of the reports seen in #10997, and so I can't say for certain whether this will fix the issue. However, it should fix some of the problems.

----

### DIFF
--- a/website/client/components/inventory/stable/index.vue
+++ b/website/client/components/inventory/stable/index.vue
@@ -92,7 +92,11 @@
             @click="petClicked(item)"
           )
             template(slot="itemBadge", slot-scope="context")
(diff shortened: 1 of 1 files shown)


paglias:
Confirmed that the fix for some of the problems seem to work, thanks @alecbrick  and welcome to the 2nd tier :)

PR #11046, opened 2019-03-04
fix: shop ui

negue:
fixes:
- use the same filter logic on all shop pages (as other pages have)
- fix margins/spacings


Tressley:
@negue -- Now the Quest containers are a bit too big due to the margin-bottom.
image

Can we lower the opacity of the price for items that are locked to the user? That's shown here: https://zpl.io/VYM1NdV

Tressley:
Margins look good. 👍

Lowering the price opacity was missed, but we can handle that in another PR. I have design updates for shop items.

PR #11048, opened 2019-03-06
Make username searches in Hall of Fame lowercase

### LINKED ISSUE #11047: Username searches in the Hall of Heroes should be case-insensitive
As described in issue https://github.com/HabitRPG/habitica/issues/10972 and its pull request https://github.com/HabitRPG/habitica/pull/10980 , the Hall of Heroes allows moderators to search for users by the Username. Currently that search is case-sensitive. It should be case-insensitive. I.e., searching for "examplename" would let you find a user called "ExampleName". 

Any contributor who wants to work on this should read the top post at https://github.com/HabitRPG/habitica/issues/10972 to learn about how the moderator search feature works and how you can access it on your local install. https://github.com/HabitRPG/habitica/pull/10980/files will show you how the current case-sensitive search is implemented.

_(NB We're not considering this to be a bug in https://github.com/HabitRPG/habitica/pull/10980. We neglected to say that the search should be case-insensitive and so that PR did implement the search feature as intended at the time!)_

JSn1nj4:
Fixes #11047: Username searches in the Hall of Heroes should be case-insensitive

Changes:
- Create separate lowercase copy of `heroId` for lowercase username searches
- Query `auth.local.lowerCaseUsername` instead of `auth.local.username`

----
UUID: b1781515-6c3f-4314-a126-705b1e8390e8


JSn1nj4:
Going to look into using Mongoose to do the case-insensitive search directly instead of needing a separate lowercase copy of the variable. This would mean not needing to modify the data before searching.

website/server/controllers/api-v3/hall.js
```diff
@@ -175,7 +175,7 @@ api.getHero = {
     if (validator.isUUID(heroId)) {
       query = {_id: heroId};
```
JSn1nj4, line 176: @Alys do you think I should apply the case-insensitive check to the UUID search as well? If all UUIDs are stored lowercase, it may be beneficial in case an uppercase letter makes it to this point.

Alys:
Thank you for asking but no that wouldn't be necessary. All of the code
base is set up to use lowercase User IDs and if that were changed
accidentally our tests would pick it up and we'd revert.

JSn1nj4:
Alright, I'll leave it as-is then. Thanks!

JSn1nj4:
Looking at this again, I don't think switching to using the lowercase username was necessary after all, since the regex itself is case-insensitive. 🤷‍♂️

paglias:
@JSn1nj4 as mentioned in the other PR I'm going to close this one but it's been added to your list of contributions!

JSn1nj4:
Thanks @paglias!

PR #11065, opened 2019-03-14
Performance:  My Challenges

negue:
Using the same paging feature as for the "find challenges" view:
- server side filtering 
- loads only 10 items per call


website/client/components/challenges/myChallenges.vue
```diff
@@ -117,44 +122,86 @@ export default {
...
-      let user = this.$store.state.user.data;
+      const filters = this.filters;
+      const user = this.user;
+
+      return this.challenges.filter(challenge => {
+        let isMember = true;
 
-      // @TODO: Move this to the server
-      return this.challenges.filter((challenge) => {
-        return this.filterChallenge(challenge, filters, search, user);
```
paglias, line 133: why isn't this function used anymore? also it looks like it was only used here so if it's really not needed it can be removed from the mixin?

paglias:
Left one comment but I think we should also have a `Loading...` status while loading new challenges (like guilds do)

website/client/components/challenges/myChallenges.vue
negue: most of it is in the server now, the function will be removed

negue:
moved the `loading` text above the `load more`

paglias:
Putting this on delta

negue:
@paglias could you put this on delta again once the CI is done ?

paglias:
Done! It'll go live in about 30 minutes, when it does can you mention the update in the testing-web channel? thanks!

website/client/components/challenges/findChallenges.vue
```diff
@@ -160,14 +167,14 @@ export default {
         this.challenges = this.challenges.concat(challenges);
       }
 
-      this.loading = false;
+      // this.loading = false;
```
paglias: is this right?

negue: nope, removed

paglias:
thanks @negue !

website/client/components/userLink.vue
```diff
@@ -11,10 +11,6 @@
 <style scoped lang="scss">
   @import '~client/assets/scss/colors.scss';
 
-  a {
-    color: $gray-50;
```
SabreCat, line 15: @negue What was this for? It's thrown off the styling of non-contributors in chat--gonna work on a hotfix to put it back, but curious what this was intended to do

negue: I removed that to be able to show the contribution level (colors) but I guess that color was needed for all other links, I'll push that back in on the infinite scrolling PR

SabreCat: OK! We sent out a hotfix in 3eb164a and d7f6c3b, so it's working now but ideally we shouldn't be using `!important` for the CSS there, heh

PR #11068, opened 2019-03-16
allow moderators/staff to always see Report button for a flagged message

Alys:
Currently, when a moderator flags a message, the "Report" button disappears, which means they can't unflag. This behaviour started with the recent flagging change but it's my fault for not testing that functionality.

This PR causes the "Report" button to always be visible for a moderator.


paglias:
LGTM, it's a bit late in the week to merge new PRs into staging so let's wait for Friday

PR #11077, opened 2019-03-21
performance: private messages - API

negue:
This is the first part to add paging for private messages.

### DIFF
--- a/test/api/v3/integration/inbox/GET-inbox_messages.test.js
+++ b/test/api/v3/integration/inbox/GET-inbox_messages.test.js
@@ -45,4 +45,23 @@ describe('GET /inbox/messages', () => {
     expect(messages[2].text).to.equal('second');
     expect(messages[3].text).to.equal('first');
   });
+
+  it('returns no messages when using page-query ', async () => {
+    const promises = [];
+
+    for (let i = 0; i < 10; i++) {
+      promises.push(user.post('/members/send-private-message', {
+        toUserId: user.id,
+        message: 'fourth',
+      }));
+    }
+
+    await Promise.all(promises);
+
+    await user.sync();
+
+    const messages = await user.get('/inbox/messages?page=1');
+
+    expect(messages.length).to.equal(4);
+  });
 });
--- a/test/api/v3/integration/inbox/GET-inbox_messages.test.js
+++ b/test/api/v3/integration/inbox/GET-inbox_messages.test.js
@@ -27,8 +27,6 @@ describe('GET /inbox/messages', () => {
       toUserId: user.id,
       message: 'fourth',
     });
-
-    await user.sync();
   });
 
   it('returns the user inbox messages as an array of ordered messages (from most to least recent)', async () => {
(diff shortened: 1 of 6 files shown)


test/api/v3/integration/inbox/GET-inbox_messages.test.js
```diff
@@ -45,4 +45,23 @@ describe('GET /inbox/messages', () => {
...
+    for (let i = 0; i < 10; i++) {
+      promises.push(user.post('/members/send-private-message', {
+        toUserId: user.id,
+        message: 'fourth',
+      }));
+    }
+
+    await Promise.all(promises);
+
+    await user.sync();
```
paglias: is this call necessary?

test/api/v3/integration/inbox/GET-inbox_messages.test.js
```diff
@@ -45,4 +45,23 @@ describe('GET /inbox/messages', () => {
     expect(messages[2].text).to.equal('second');
     expect(messages[3].text).to.equal('first');
   });
+
+  it('returns no messages when using page-query ', async () => {
```
paglias: can you clarify the text here? I mean why it says returns no messages when in fact 4 are returned?

website/server/libs/inbox/index.js
```diff
@@ -1,12 +1,25 @@
 import { inboxModel as Inbox } from '../../models/message';
 
-export async function getUserInbox (user, asArray = true) {
-  const messages = (await Inbox
+const PM_PER_PAGE = 10;
+
+export async function getUserInbox (user, options = {asArray: true, page: 0}) {
+  if (typeof options.asArray === 'undefined') {
```
paglias, line 6: why check if the option is undefined if we already provide a default in the parameter definition?

paglias (commented):
left some comments

test/api/v3/integration/inbox/GET-inbox_messages.test.js
negue: I copied it from the `before` , tested both without the ´user.sync` guess it's not needed anymore

test/api/v3/integration/inbox/GET-inbox_messages.test.js
negue: I had the best first just without results, then I thought It would make more sense to have some entries so I changed the test, gonna rename the test now

website/server/libs/inbox/index.js
negue: if you call it like 
```
const userInbox = await inboxLib.getUserInbox(user, {
      page,
    });
```

then `asArray` won't be set even it is available as the default parameter thats why the second  check is needed

paglias:
LGTM, thanks @negue ! If you want to start working on using this new option on web it would be great!

paglias:
I just realized there's a small problem with this change, since messages are grouped by conversation it would be useless to only load 10 without considering the context like if you opened a specific conversation...

Might make more sense to offer the pagination ability with messages grouped by conversation?

negue:
Yeah just a bit after this was merged I realized this myself.

I'll create a new api to list all conversations first and then load the messages per conversation (while using pagination)

Issue #11015, opened 2019-02-21
Scrollbar disabled after certain modals (popups) appear


shanaqui:
> _@ravenlune (9272ab2c-4b66-4b74-aebd-799725885d1c): "I am very tired right now, so forgive me if I'm awfully mistaken, but I've noticed a very weird thing just now... When I equip any item on the website, my scrollbar disappears, and I can't scroll the page at all. This happens when I click on the item, after I click equip or click the X on the pop-up my scrollbar goes blank (before I tried clear browser data it was even before the pop-up fully appeared). If it's worth anything, it goes blank for maybe a second or less after I click the item, then appears again with the pop-up, then disappears again when I close it. If I click on the tabs at the top (Stable or Items) or any of the links in the top bar, it doesn't come back. I have to refresh to get it back. This all happened starting today, never before. I don't understand why. I tried clear browser data, and I turned off all my extensions that could likely affect it, and nothing worked. I did install a new extension today but, again, disabling it didn't help. I am using the latest version on Chrome on Windows 10. Thanks for your work. I hope this made sense..."_

Confirmed in Chrome on Windows 10.

stuttste:
Hi Habitica team, new potential contributor here. I've confirmed this issue happens to me in Chrome and Internet Explorer on Windows 10 as well. I think this issue should be a good initial issue to dig into. Can I get confirmation from an admin to begin? I should be able to start looking into it this weekend.

SabreCat:
We'll probably look at it ourselves too, but feel free to dig into it, @stuttste !

robert-osborne:
Confirm this happens reliably on this page.  But I've seen the disappearing scrollbars on other pages as well (just not reliably duplicatable).  Especially switching to Shops -> Quests or Inventory -> Stable so this might be related.

brightbold:
Can also confirm this happens (and appears to be consistently reproducible) in Chrome, Firefox, and Safari on Mac.

Alys:
There's a report from @Ensorceled (4428e727-2a0e-4072-b578-fc28029b33f3) indicating that this bug might occur under other circumstances too:

"the scroll bars will sometimes disappear when I do an action (this is new behaviour):

1. Inventory -> Equipment equip an item and then the scroll bar disappears and I have to reload the page (refresh every time)
2. Switch to Shops -> Quests or Inventory -> Stable ... sometimes no scrollbar, have to refresh the screen (occasionally)
[...]

PR #11079, opened 2019-03-22
Less hacky (but still hacky) fix for equip modal shenanigans

SabreCat:

Continuing partial fixage for #11015 

### Changes

While researching to provide more information to the Bootstrap-Vue folks on https://github.com/bootstrap-vue/bootstrap-vue/issues/2809, I discovered that if we toggle the `visible` property of the modal component rather than emit `bv::hide::modal`, the modal closes cleanly without needing to hack the document body ourselves.

I haven't been able to reproduce the other situations where modal scrollbars get stuck, unfortunately. :\

### DIFF
--- a/website/client/components/inventory/equipment/equipGearModal.vue
+++ b/website/client/components/inventory/equipment/equipGearModal.vue
@@ -4,7 +4,6 @@
     v-if="item != null",
     :hide-header="true",
     @change="onChange($event)"
-    @hide="fixDocBody()"
   )
     div.close
       span.svg-icon.inline.icon-10(aria-hidden="true", v-html="icons.close", @click="hideDialog()")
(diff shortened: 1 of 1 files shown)


SabreCat:
@paglias @randi2kewl

website/client/components/inventory/equipment/equipGearModal.vue
```diff
@@ -187,17 +186,9 @@
...
-      fixDocBody () {
-        document.body.classList.remove('modal-open');
-        if (document.body.getAttribute('data-modal-open-count') <= 1) {
-          document.body.removeAttribute('data-modal-open-count');
-        } else {
-          document.body.setAttribute('data-modal-open-count', document.body.getAttribute('data-modal-open-count') - 1);
-        }
-        document.body.removeAttribute('data-padding-right');
-        document.body.style.removeProperty('padding-right');
+        this.visible = false;
```
paglias, line 189: Hmm, why are we setting `:visible` to true in this modal? I don't remember it with any other modal out there.

It seems to come from this very old PR from @negue

SabreCat: I'm not entirely sure--I `grep`ped for other instances of `:visible`, and there aren't many examples. website/client/components/inventory/stable/welcomeModal.vue I think is all.

paglias: Hmm maybe it's worth trying to remove it and see what happens?

SabreCat: I suspect we'd need to restructure how the modal is called. Just taking the property off means nothing happens when the user clicks an equipment item 😅

paglias: Hmm yeah I was just wondering if maybe it was the cause...

SabreCat: I had the same thought--for whatever reason, this modal is the most reliably affected by the issue, so there's _something_ different going on here. Unfortunately, other modals do run into it, just not in a way I've figured out how to reproduce yet.

negue: @paglias `:visible="true"` was back then because it used the `v-if` to show and remove it, pretty soon after that most of the dialogs have been refactored to use the show-modal `$emit` .

Guess this should be changed here too

SabreCat:
Bootstrap-vue's latest update appears to fix this more cleanly, so I'm reverting the old hack and closing this!

PR #11106, opened 2019-04-05
Add message of cancelled quest in party chat

### LINKED ISSUE #11093: cancelled quests should have message in party chat
When the quest owner or party leader cancels a quest that's still in the invitation phase, there's no indication for other party members that that's what happened and it can lead to confusion (there was a recent bug report from someone who understandably thought that something had gone wrong with the quest).

There should be a message posted to party chat to say, for example, "Alys cancelled the party quest Vice, Part 1: Free Yourself of the Dragon's Influence."

A similar message already exists for the case where a running quest is aborted so similar code could be used for this issue:
```
$ git grep "the party quest"
test/api/v3/integration/quests/POST-groups_groupid_quests_abort.test.js:130:    expect(Group.prototype.sendChat).to.be.calledWithMatch(/aborted the party quest Wail of the Whale.`/);
website/server/controllers/api-v3/quests.js:425:    const newChatMessage = group.sendChat(`\`${user.profile.name} aborted the party quest ${questName}.\``);
/code/habitica 2048 $ 
```

Also, as part of the fix for this issue, please change the comment shown below to remove ` (see questCancel for BEFORE)`  because the function it refers to is now called `cancelQuest` and it's likely that we'll forget to update the comment if we rename it again :)
https://github.com/HabitRPG/habitica/blob/a2261e35919a0396d2e74ea70f209b68ebc004c6/website/server/controllers/api-v3/quests.js#L408

HydeHunter2:
Fixes #11093

### Changes
- A message of cancelling a quest that's still in the invitation phase will be displayed in the party chat

github

- Change the comment due to function's renaming (`questCancel`->`cancelQuest`)

----

### DIFF
--- a/website/server/controllers/api-v3/quests.js
+++ b/website/server/controllers/api-v3/quests.js
@@ -372,12 +372,12 @@ api.cancelQuest = {
 
     let questName = questScrolls[group.quest.key].text('en');
     const newChatMessage = group.sendChat(`\`${user.profile.name} cancelled the party quest ${questName}.\``);
-    await newChatMessage.save();
 
     group.quest = Group.cleanGroupQuest();
     group.markModified('quest');
 
     let [savedGroup] = await Promise.all([
+      newChatMessage.save(),
       group.save(),
       User.update(
         {'party._id': groupId},
--- a/website/server/controllers/api-v3/quests.js
+++ b/website/server/controllers/api-v3/quests.js
@@ -363,13 +363,13 @@ api.cancelQuest = {
     if (validationErrors) throw validationErrors;
 
     let group = await Group.getGroup({user, groupId, fields: basicGroupFields.concat(' quest')});
-    
+
     if (!group) throw new NotFound(res.t('groupNotFound'));
     if (group.type !== 'party') throw new NotAuthorized(res.t('guildQuestsNotSupported'));
     if (!group.quest.key) throw new NotFound(res.t('questInvitationDoesNotExist'));
     if (user._id !== group.leader && group.quest.leader !== user._id) throw new NotAuthorized(res.t('onlyLeaderCancelQuest'));
     if (group.quest.active) throw new NotAuthorized(res.t('cantCancelActiveQuest'));
-    
+
     let questName = questScrolls[group.quest.key].text('en');
     const newChatMessage = group.sendChat(`\`${user.profile.name} cancelled the party quest ${questName}.\``);
     await newChatMessage.save();
--- a/website/server/controllers/api-v3/quests.js
+++ b/website/server/controllers/api-v3/quests.js
@@ -377,8 +377,8 @@ api.cancelQuest = {
     group.markModified('quest');
 
     let [savedGroup] = await Promise.all([
-      newChatMessage.save(),
       group.save(),
+      newChatMessage.save(),
       User.update(
         {'party._id': groupId},
         Group.cleanQuestParty(),
--- a/website/server/controllers/api-v3/quests.js
+++ b/website/server/controllers/api-v3/quests.js
@@ -363,11 +363,16 @@ api.cancelQuest = {
     if (validationErrors) throw validationErrors;
 
     let group = await Group.getGroup({user, groupId, fields: basicGroupFields.concat(' quest')});
+    
     if (!group) throw new NotFound(res.t('groupNotFound'));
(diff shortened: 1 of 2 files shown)


HydeHunter2:
While writing tests, I noticed that `partyMembers[1]` doesn't accept the invitation to quest:

https://github.com/HabitRPG/habitica/blob/f09b65e1084bacad76d954dac62424c58f86cbbf/test/api/v3/integration/quests/POST-groups_groupid_quests_cancel.test.js#L99-L110

Although in tests for aborting quest, `partyMembers[1]` accepts the invitation:

https://github.com/HabitRPG/habitica/blob/f09b65e1084bacad76d954dac62424c58f86cbbf/test/api/v3/integration/quests/POST-groups_groupid_quests_abort.test.js#L88-L91

Could I fix this in my commit, which adds tests?

Alys:
@HydeHunter2 In the 'cancels a quest' test, it's actually correct that `partyMembers[1]` hasn't accepted the invitation. :) The cancel command can be done only while the quest is still in the invitation stage, so in the test it's important to have a party member who hasn't accepted or rejected yet (after all members have accepted/rejected, the invitation phase ends automatically).

But thank you for asking about this! I guess it isn't clear at first and it does look like an inconsistency. I wonder if it's worth adding a brief comment under line 101 to explain that `partyMembers[1]` doesn't accept so that the quest stays in the invitation stage.

HydeHunter2:
Ok. I understood.
Then I'll add a brief comment.

Thanks for the clarification!

website/server/controllers/api-v3/quests.js
```diff
@@ -363,12 +363,17 @@ api.cancelQuest = {
...
+
     if (!group) throw new NotFound(res.t('groupNotFound'));
     if (group.type !== 'party') throw new NotAuthorized(res.t('guildQuestsNotSupported'));
     if (!group.quest.key) throw new NotFound(res.t('questInvitationDoesNotExist'));
     if (user._id !== group.leader && group.quest.leader !== user._id) throw new NotAuthorized(res.t('onlyLeaderCancelQuest'));
     if (group.quest.active) throw new NotAuthorized(res.t('cantCancelActiveQuest'));
 
+    let questName = questScrolls[group.quest.key].text('en');
+    const newChatMessage = group.sendChat(`\`${user.profile.name} cancelled the party quest ${questName}.\``);
+    await newChatMessage.save();
```
paglias: thanks @HydeHunter2 ! Just one quick change is needed, could you move this save to the Promise.all down at line 380? So we don't block the other save operations while waiting for this one? Thanks

paglias (changes requested):

website/server/controllers/api-v3/quests.js
HydeHunter2: Of course

HydeHunter2:
Like this?
https://github.com/HabitRPG/habitica/blob/710aabbf4760233dd99ada44a6c2e16c0fe7857d/website/server/controllers/api-v3/quests.js#L373-L381

website/server/controllers/api-v3/quests.js
```diff
@@ -372,12 +372,12 @@ api.cancelQuest = {
 
     let questName = questScrolls[group.quest.key].text('en');
     const newChatMessage = group.sendChat(`\`${user.profile.name} cancelled the party quest ${questName}.\``);
-    await newChatMessage.save();
 
     group.quest = Group.cleanGroupQuest();
     group.markModified('quest');
 
     let [savedGroup] = await Promise.all([
+      newChatMessage.save(),
```
paglias: thanks! last thing, putting newChatMessage.save() at the beginning of the array will cause `savedGroup` to get its value instead of `group.save()`. It should be at least in second place inside the array

paglias: thanks! last thing, putting newChatMessage.save() at the beginning of the array will cause savedGroup to get its value instead of group.save(). It should be at least in second place inside the array

paglias (changes requested):

website/server/controllers/api-v3/quests.js
HydeHunter2: Sorry that I couldn't do it from the first commit(

Thanks for the review!

paglias:
@HydeHunter2  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11110, opened 2019-04-08
Inbox: Add API to list conversations

negue:


website/server/libs/inbox/index.js
```diff
@@ -1,14 +1,23 @@
...
-export async function getUserInbox (user, options = {asArray: true, page: 0}) {
+export async function getUserInbox (user, options = {asArray: true, page: 0, conversation: null}) {
   if (typeof options.asArray === 'undefined') {
     options.asArray = true;
   }
 
+  const findObj = {ownerId: user._id};
+
+  if (options.conversation) {
+    findObj.uuid = options.conversation;
```
paglias, line 18: added an index for this

website/server/controllers/api-v4/inbox.js
```diff
@@ -72,4 +72,17 @@ api.clearMessages = {
   },
 };
 
+api.conversations = {
```
paglias, line 83: can you add some apidoc comments for this?

test/api/v4/inbox/GET-inbox-conversations.test.js
```diff
@@ -0,0 +1,43 @@
...
+      toUserId: user.id,
+      message: 'fifth',
+    });
+  });
+
+  it('returns the conversations', async () => {
+    const result = await user.get('/inbox/conversations');
+
+    expect(result.length).to.be.equal(3);
+    expect(result[0].username).to.be.equal(user.profile.name);
```
paglias: this should check for the username instead of the profile name?

website/server/libs/inbox/index.js
```diff
@@ -29,12 +38,46 @@ export async function getUserInbox (user, options = {asArray: true, page: 0}) {
...
+      {
+        $sort: {timestamp: -1},
+      },
+    ]);
+
+  const conversationList = (await query.exec()).map(c => c._id);
+
+  const users = await User.find({_id: {$in: conversationList}})
+    .select('_id profile.name auth.local.username')
+    .exec();
```
paglias, line 64: I think to make it a bit faster it might be worth to use `.lean()` here

paglias (changes requested):
left a couple of comments but overall looks good, sorry for the delay!

negue:
@paglias any idea why the test passes sometimes and other times it returns the wrong order ?

paglias:
which test @negue? do you have a link to the travis build?

negue:
This test => https://github.com/HabitRPG/habitica/pull/11110/files#diff-6d0b7cf49acf3d515cfeecb237024e64R37

I ran it multiple times locally, it may works and may not ... which is weird with the same code :D 
Travis => https://travis-ci.org/HabitRPG/habitica/jobs/517471325#L843

paglias:
Hmm looks like the sorting isn't stable? maybe try to sort by _id here https://github.com/HabitRPG/habitica/pull/11110/files#diff-ebb5855000db09391acfb6b15aed8fc5R62

negue:
@paglias 

I tried another approach, load the conversations and sort those after, but even this doesn't seem to work.

Filter by `_id` wouldn't work either since the ids are random and not counting from 0 ^^

paglias:
Hmm ok, I'll give it a look tomorrow!

paglias:
@negue should be working now

negue:
@paglias awesome :> guess I used the wrong the orderBy syntax ...

 is there anything left for this api change?

paglias:
Nope, merging now

PR #11112, opened 2019-04-11
Challenges: infinite scroll

negue:
infinity scrolling with a 1s debounce  - or else - the loading would trigger on every scrolling pixel ^^


paglias:
Looking good but there's a conflict? I think with the hotfix for the links styling

negue:
@SabreCat how about this approach? all without a tier will get the class `.no-tier` with the "normal" color and no `!important`

SabreCat:
That should do it! Thank you!

PR #11125, opened 2019-04-22
Reload after rebirth

### LINKED ISSUE #11114: Orb of Rebirth should reload website
When you use the Orb of Rebirth, the website should reload itself as soon as the Rebirth has occurred so that all changes caused by Rebirth are seen immediately.

Currently, there's no visual change after Rebirth has happened so it looks like Rebirth hasn't worked. Worst of all, if the user was at or above level 100 and so had Rebirth available for free, it looks like it's still free so they try to use it again not realising they'll be charged gems.

I'm marking this as high priority since it's causing players to buy Rebirth multiple times, which wastes their gems.

HydeHunter2:
Fixes #11114

### Changes
- Add restart after rebirth ( i.e. after purchasing `Orb of Rebirth` )
- Remove restart after closing achievement: `Began a New Adventure`
  - This reload is not needed, as page now reloads immediately after purchasing `Orb of Rebirth`

----


paglias:
Wow @HydeHunter2 ! Thanks for all these PRs and for the clean code in all of them!

paglias:
Welcome to Tier number 2 and well in progress for the third!

PR #11126, opened 2019-04-22
Fix 500 server error when trying to log in with wrong username

phillipthelen:
Adds a check to see if the user even exists before checking if a hashed password exists. Otherwise there is a 500 server error, because user is null and then you can't access the auth property on that.


paglias:
LGTM

PR #11147, opened 2019-04-30
fix loading owned messages

### LINKED ISSUE #11145: My Challenges is blank when filtering for Owned
As reported by Ceran, Hamster Juggler (@Ceran, aabe567b-aa55-42c1-b612-b7699a7ba9b7):

"Addition to the problems on the "my challenge" page I posted a while ago: I now can't see all of my challenges. Some just won't show up there and I have to go to the guild they're posted in and access them from the guild challenge list there. When I try to filter by challenges owned, I get shown just very few challenges, most don't show up at all (even those not affected by the above named issue). I think the new "show more" button used to be at the bottom, but it's just not there now using the filter, so no way to actually show the other challenges."

I've confirmed this on my own account. I own several challenges but when I got to https://habitica.com/challenges/myChallenges and click the "Owned" filter checkbox, I see the "You don't have any Challenges" message.

negue:
fixes #11145 

this was a filter issue having a list of the group-challenges and using $and with `leader` resulted in none, this should be correct now

### DIFF
--- a/website/server/controllers/api-v3/challenges.js
+++ b/website/server/controllers/api-v3/challenges.js
@@ -364,12 +364,14 @@ api.getUserChallenges = {
       $and: [{$or: orOptions}],
     };
 
-    if (owned && owned === 'not_owned') {
-      query.$and.push({leader: {$ne: user._id}});
-    }
+    if (owned) {
+      if (owned === 'not_owned') {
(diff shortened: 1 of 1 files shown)


paglias:
thanks @negue !

PR #11148, opened 2019-05-02
Fix challenge update

HydeHunter2:
Fixes an issue that we discussed via email (@phillipthelen is aware)

----

### DIFF
--- a/test/api/v3/integration/challenges/PUT-challenges_challengeId.test.js
+++ b/test/api/v3/integration/challenges/PUT-challenges_challengeId.test.js
@@ -56,11 +56,11 @@ describe('PUT /challenges/:challengeId', () => {
       tasksOrder: 'new order',
       official: true,
       shortName: 'new short name',
+      leader: member._id,
 
       // applied
       name: 'New Challenge Name',
       description: 'New challenge description.',
-      leader: member._id,
     });
 
     expect(res.prize).to.equal(0);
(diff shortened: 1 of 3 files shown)


paglias (changes requested):
@HydeHunter2 thanks! The changes look good but since we already have a list of fields that cannot be updated at https://github.com/HabitRPG/habitica/blob/develop/website/server/models/challenge.js#L65 could you add `leader` there?

HydeHunter2:
Oh yes, this is a better solution

paglias (approved):

paglias:
Thanks a lot @HydeHunter2 for the investigation and the fix, notated in your profile, welcome to the 3rd tier :)

Alys:
Recently I had to track down why this change was made so I figure I may as well share what I found. :) It was done to fix a vulnerability - that's why the pull request does not give details (quite correctly). Since it's been merged though, it's no longer a security risk to reveal the details, so here are excerpts from the email in which HydeHunter2 kindly told us about the vulnerability they found:

> I found an exploit that allows you to add any challenge to anyone.  I.e. this exploit can for example:
> - Publish a public challenge on behalf of the admin.
> - Spam tab "my challenge" to someone.
> 
> The vulnerability is that when updating the challenge, Leader has the ability to change any challenge parameters (including the leader's UUID) via a PUT request to the server at /api/v4/challenge/{Challenge_ID}
> 
> Thus, an attacker can create a challenge of any content, and then "update" the leader to any other user. 
>....
> Thus, an attacker can:
> - Create phishing challenges on behalf of the administration.
> - Spam the "My challenge" tab to someone.
> - And so on..



It was originally intentional that a challenge owner be able to select a new owner, and there are still cases where that would be useful, but as HydeHunter2 realised, the ability to change the leader without the new leader's permission made this feature too open to abuse.

PR #11150, opened 2019-05-04
Remove markdown elements from PM

### LINKED ISSUE #11027: markdown should be parsed in Private Message (PM) previews
As reported by @ravenlune (9272ab2c-4b66-4b74-aebd-799725885d1c):

"Markdown doesn't render in PM previews, when it does in the actual message being previewed."

image

I.e., markdown parsing needs to be done in the preview page on the left-hand side of the PM modal.

HydeHunter2:
Fixes #11027

### Changes
- Remove markdown elements from Private Message

Before | After
--------|--------
Git | Git

----

### DIFF
--- a/website/client/components/userMenu/inbox.vue
+++ b/website/client/components/userMenu/inbox.vue
@@ -35,7 +35,7 @@
             .time
               span.mr-1(v-if='conversation.username') @{{ conversation.username }} •
               span {{ conversation.date | timeAgo }}
-            div {{conversation.lastMessageText ? conversation.lastMessageText.substring(0, 30) : ''}}
+            div {{ conversation.lastMessageText ? substring(removeTags(parseMarkdown(conversation.lastMessageText)), 30) : '' }}
       .col-8.messages.d-flex.flex-column.justify-content-between
         .empty-messages.text-center(v-if='!selectedConversation.key')
           .svg-icon.envelope(v-html="icons.messageIcon")
@@ -218,6 +218,7 @@ import filter from 'lodash/filter';
 import sortBy from 'lodash/sortBy';
 import groupBy from 'lodash/groupBy';
 import { mapState } from 'client/libs/store';
+import habiticaMarkdown from 'habitica-markdown';
 import styleHelper from 'client/mixins/styleHelper';
 import toggleSwitch from 'client/components/ui/toggleSwitch';
 import axios from 'axios';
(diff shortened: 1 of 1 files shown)


HydeHunter2:
~~Could you put a label `needs work` for this PR? (or just don't make a review yet)
I looked at the project in Zeplin and realized that it's necessary to limit the length of the text not by number of characters (since they are of different lengths), but somehow limit the number of lines.
I don’t know how to implement it yet, but I'll try (maybe CSS will help)~~

HydeHunter2:
I did it!
~~*But without ellipsis at the end. No idea how to implement it.~~

HydeHunter2:
###### Don't pay attention to past messages.
Now the length of the text is limited to 2 lines and there is ellipsis at the end.
Github

website/client/components/userMenu/inbox.vue
```diff
@@ -209,6 +209,16 @@
     color: $gray-200;
     margin-bottom: 0.5rem;
   }
+
+  .messagePreview {
+    display: block;
+    display: -webkit-box;
+    -webkit-line-clamp: 2;
```
paglias, line 216: can you remove the -webkit prefixes from rule names? it'll be added automatically at compile time

Also it doesn't look like firefox and IE11 support it... is it really needed? @Tressley 

I'm wondering if we could get the same result by using `overflow: hidden;` and setting the height of the box given that we know the height of each line

paglias:
Left one comment but it's working well as usual @HydeHunter2 :)

I'll put this up for testing as well

website/client/components/userMenu/inbox.vue
HydeHunter2: I've read that `-webkit` is not supported by some browsers, but I found a similar solution and thought it's normal.

https://github.com/HabitRPG/habitica/blob/4289becccc8663985736094e9fa2360d44733f0c/website/client/components/challenges/challengeItem.vue#L45-L60

But if you say this isn't a good solution, I'll fix up it.
(But I still don't know how to implement an ellipsis using `overflow: hidden;` :( If it(ellipsis)'s definitely necessary, I'll try again)

paglias: Ah, i didn't noticed it was used elsewhere , in that case it's fine as long as we make sure the layout isn't too broken on Firefox for example

Tressley: Sorry, just getting a chance to answer a few pings; I do think it's necessary for this to work in Firefox, but not IE.

Tressley:
This looks good to me 👍. There was a part of me that thought we should still show emojis in the preview, but I think the current implementation is a good introduction @HydeHunter2 . We can run with this and add emoji support in the future if we hear feedback.

paglias:
> This looks good to me 👍. There was a part of me that thought we should still show emojis

Ah, you're right! I forgot about it but yeah it would be a big change from the current implementation so maybe we can revisit this later when we go through the other emojis issues that we have?

PR #11157, opened 2019-05-07
Performance: Inbox Paging / loading

negue:
- [x] Load Conversations
- [x] Load only messages by conversation
- [ ] ~~Infinite Scroll~~
- [x] "Load More"-Button Designed
- [x] keep scroll position on load more

@paglias could you take a look at the mugen-scroll, I use the same code as for the challenges, yet it doesn't trigger the loading


paglias:
@negue are you still having issues with mugen scroll? If yes I can give it a look tomorrow morning, otherwise let's go with the Load More button so we can get this out

negue:
I removed mugen-scroll and use a simple scroll event, but the fact that messages are added to the top, it scroll position stays from the top, so everything jumps a bit, if I can't fix this kind of jumping, I probably will use the button way

paglias:
Oh I see, it probably isn't meant to work with new content being added to the top. I'd say to go with the button then because we don't want to maintain two different types of infinite scroll imo

paglias:
Also looks like there are a couple of server side failing tests

paglias:
@negue regarding mugen-scroll there's https://github.com/egoist/vue-mugen-scroll/issues/4 does it help maybe?

negue:
the `format` test, fails because it expects => `contributor: {
            toObject () {
              return 'contributor object';
            },
          },` to be just `contributor` 

should I rather use `toObject` at https://github.com/HabitRPG/habitica/pull/11157/files#diff-388968ce2a61bca656313c066ddbda7aR100 or leave the current code as is?

paglias:
Hmm weird about the test.. is the contributor object saved correctly? Because userCopy should have everything already as plain objects. But if it's saved correctly I'd change the test

negue:
could this be pushed to gamma @paglias, once the tests are done

paglias:
It's on delta @negue @Tressley 

One thing I noticed is that the load more button when clicked shows you the oldest message instead of the first one that wasn't visible before (probably because it doesn't keep the scroll position)

negue:
yeah I tried to keep the scroll position before the loading and reset it once the data is arrived but not really "working" , I could remove it but then it would stay the same scrollPosition where the button was but adds all new messages (under the load more, if 10 messages were loaded, otherwise hides the button)

paglias:
I think there's a bug with messages sent to yourself , posted about it in the opensource slack

Tressley:
@negue -- Mock-up for button https://zpl.io/VDDXgwv

The jump is a bit jarring, so I'd like to work on getting the scroll position to stay as I think it's fairly necessary for the proper experience.

Tressley:
@negue -- A couple of bugs I encountered:

When I send a message, the "Report" and "Delete" options aren't available.
image
Additionally, if I send new messages and click "Load Earlier Messages", my new messages appear as blank.
image

Tressley:
@negue -- Still seeing a bit of a jump from scroll position when loading earlier messages. This looks to occur when a loaded message is tall (has a lot of content.) Here's a video that catches the jump in action: https://streamable.com/ybifn

paglias:
So this seems good to go! Will go live today probably

PR #11163, opened 2019-05-12
FIX:custom day start with todos

### LINKED ISSUE #11139: Custom Day Start time is not used for displaying To-Do due dates
As reported by Arkadiy @arkadiyjs (4401fa80-29a5-4f9e-8a51-94081fb3bffd) who has their Custom Day Start time set to 4am:

'It seems that the custom day start is not functional [for To-Do due dates]. [To-Dos] that are "Due Today" change to "Due a day ago" after [midnight] even though the day is supposed to start at 4 am.'

sayansur25:

Fixes put_#_and_issue_numer_here
Custom Day Start time is not used for displaying To-Do due dates #11139

Fixes #11139

### Changes
timeTillDue() is changed.

----
UUID:

### DIFF
--- a/website/client/components/tasks/task.vue
+++ b/website/client/components/tasks/task.vue
@@ -634,8 +634,8 @@ export default {
     },
     timeTillDue () {
       // this.task && is necessary to make sure the computed property updates correctly
-      const endOfToday = moment().endOf('day');
-      const endOfDueDate = moment(this.task && this.task.date).endOf('day');
+      const endOfToday = moment();
+      const endOfDueDate = moment(this.task && this.task.date).startOf('day').add(this.user.preferences.dayStart, 'hours');
 
       return moment.duration(endOfDueDate.diff(endOfToday));
     },


paglias:
LGTM @sayansur25 thanks! I've asked @Alys to check if everything's fine and after that we can merge

Alys:
This successfully considers the Custom Day Start time (CDS) when displaying a due date, however it has side-effects, at least some of which I think are because the due date is now calculated from the start of the day rather than the end of the day.

Tasks that are/were due in the near future/past are shown as being due in a number of minutes or hours. The production site delineates them by day only, which is still the desired situation. Specifying minutes and hours implies that you can choose a due time as well as a due date, which will lead to confusion and unnecessary bug reports from players who can't find a time setting.

Any tasks that are due today are displayed as "Due x minutes/hours ago" (i.e., at the exact CDS time), making it look as if the player has failed to complete them on time, when really they have the rest of the day to complete them before they could be considered overdue. This will cause unnecessary stress for some players (we already get occasional complaints about the red highlighting for "Due Today" making the tasks look "bad"; "due x hours ago" will have a similar effect).

Tasks that were due yesterday are sometimes displayed as "Due 2 days ago", depending on how far through the day you are.

Below are screenshots from production and from my local install using this PR to illustrate these problems. In all screenshots, the CDS is 0 for simplicity, the current date is 21 May, and my PC's time was set to either shortly after midnight or around midday or after 11pm. The due days displayed should be identical between production and local install. 

----

Production at 00:01am:
image

Local install with this PR at 00:01am:
image

----

Production at around midday:
image

Local install with this PR at around midday:
image

----

Production at about 11:55pm:
image

Local install with this PR at about 11:55pm:
image

sayansur25:
But the due time is given accurately, so the players can see them and schedule there work accordingly. Otherwise, if the day ends in 5 mins this will show 'due in a day'. Users who don't know about their CDS, So they can schedule the time accordingly. So what you want to do?

Alys:
A To-Do that's due today should show "Due Today" regardless of how far through the day we are (production is doing that correctly). "Due in a day" would be a bug. To-Dos don't have due times so an exact time shouldn't be shown.

For the case you describe where the day ends in five minutes, "Due Today" should be all that the player needs to see to know that they have to schedule the task right now. :)

paglias:
Hi @sayansur25 ! Do you have any update on the status of this PR? thanks!

paglias:
Hi @sayansur25 ! I'm going to close this one but if you want to work again on it just leave a message

PR #11178, opened 2019-05-18
adjust error messages for muted and banned users, and system flagging error

Alys:
No issue was logged for this.

This PR improves some of the error notification messages that players see if they have had their chat permissions revoked.

It also improves the message you see if you try to flag/report a `system` message.

These changes have been discussed with and approved by beffymaroo and the other mods.

I've tested them all on a local install and they all work. A couple of examples are below.

image

```
$ curl -d "" -X POST -H "x-api-user: ba381bcd-831c-4d5a-9d91-e4d936139f77" -H "x-api-key: ..." http://localhost:8080/api/v3/groups/cb32f49e-2d95-473a-bc96-7dd96da4ebbe/chat/d5decc58-f7dc-4a80-85c3-620991dff5d8/flag
{"success":false,"error":"BadRequest","message":"You cannot report a system message. If you need to report a violation of the Community Guidelines related to this message, please email a screenshot and explanation to our Community Manager at admin@habitica.com."}
```


Alys:
The failing test seems unrelated.
```
  1) payments - stripe - #checkout
       verifies credentials:
      AssertionError: expected { Object (code, error, ...) } to deeply equal { Object (code, error, ...) }
      + expected - actual
       {
         "code": 401
         "error": "Error"
      -  "message": "Invalid API Key provided: aaaabbbb********************1111"
      +  "message": "Invalid API Key provided: ****************************1111"
       }
      at getBasePromise.then.then.newArgs (node_modules/chai-as-promised/lib/chai-as-promised.js:302:22)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

Alys:
The latest commit improves the wording in the error notification that a player sees when they've been blocked.

It also also changes markdown links in that message to plain text because raw markdown was being shown in some locations (for example when you saw the error messages on the command line while using an API command).

BEFORE, showing raw markdown:
image

AFTER:
image

website/common/locales/en/groups.json
```diff
@@ -266,9 +266,9 @@
   "userRequestsApproval": "<%= userName %> requests approval",
   "userCountRequestsApproval": "<%= userCount %> members request approval",
   "youAreRequestingApproval": "You are requesting approval",
-  "chatPrivilegesRevoked": "You cannot do that because your chat privileges have been revoked.",
-  "cannotCreatePublicGuildWhenMuted": "You cannot create a public guild because your chat privileges have been revoked.",
-  "cannotInviteWhenMuted": "You cannot invite anyone to a guild or party because your chat privileges have been revoked.",
+  "chatPrivilegesRevoked": "You cannot do that because your chat privileges have been removed. For details or to ask if your privileges can be returned, please email our Community Manager at admin@habitica.com or ask your parent or guardian to email them. Please include your @Username in the email. If a moderator has already told you that your chat ban is temporary, you do not need to send an email.",
```
paglias: this new part here seems common to several strings, could it be moved into its own string?

paglias (commented):
Left a comment about the possibility of condensing some repeated content in only one string in order to avoid inconsistencies between translations and load less data :)

If it's too difficult feel free to merge anyway

website/common/locales/en/groups.json
Alys: I've replaced the near-duplicate messages with the one generic one.

That was actually the original suggestion I made to the other mods and they were fine with it so this is okay from a moderator perspective.

Ready for review again.

PR #11179, opened 2019-05-18
[WIP] Shared completion for Group Dailies

### LINKED ISSUE #10690: State of Dailies does not change in Group Plan taskboard
When an assigned user (or any user) in a Group Plan completes a Daily that they are able to claim, the state of that Daily does not change on the Group Task board. E.g. the color stays the same, the streak count does not change, and the Daily remains in the "Due" tab rather than the "All" tab. Refreshing on the part of either party involved does not resolve this.
group_task_board_view
assigned_user_view

MeanderingCode:
### NOTE

this depends on https://github.com/HabitRPG/habitica/pull/11389 now

----

This PR adds shared completion for Group Dailies as well as allows "uncompletion" for Group Dailies and Todos.

I am highlighting several places where I would like discussion on the feature or information on how Habitica works.

Fixes #10690 

### Changes

* "Single" Shared Completion Dailies are marked completed for all assigned users [whose "day" is the same as the completing user] when one assigned user completes
* Cron checks for completion of "Single" Shared Completion Dailies
* Master Daily (Single Completion) on Group Task Board is marked complete when one assigned user completes
* Dailies and Todos can be "uncompleted", updating/syncing that Task to assigned users and the Master Task
  * User cannot "uncomplete" a Single Shared Daily another assigned user completed
* Adds optional `/yesterdaily` segment to the end of the task scoring route which modifies history date to user's "yesterday"
* Completed Dailies due/not due filters work in Group Task Board list
* Fixes counting bug causing Shared Completion All Approval Required Tasks to complete when only one assigned user still needs approval and an approved user completes the task
* Add Mongo version key (`__v`) to `omitAttrs` in `syncableAttrs` for Tasks
* Add tests for Todos and Dailies shared completion syncing

### WIP / Todo

* More tests
* History for Shared Completion All type tasks
* Uncomplete Group (Master) Dailies on someone's cron

### Postponed for later

* Track Group Board Dailies completion based on user's "day", vs "first" user cron?
* Sync tasks to assigned users on master task save
* Master task streaks for dailies and habits
* *Potentially:* Reward Group with Gems based on Streaks?

----
UUID: 7d328dff-0f2f-4ecb-b4d4-dd323b1b676d


website/server/libs/groupTasks.js
```diff
@@ -1,24 +1,73 @@
...
+    } else {
+      // The task was uncompleted by the group member and should be recreated for assignedUsers
+      let group = await Groups.findById(masterTask.group.id);
+      let userList = [];
+      masterTask.group.assignedUsers.forEach(userId => {
+        let query = {_id: userId};
+        userList.push(query);
+      });
+      // @REVIEW There has to be a better way to do this
+      // async arrow function callback to the find and/or a forEach resulted in Mongoose ParallelSaveErrors
```
MeanderingCode: There are a mix of Promises and async functions in the codebase. Using Promises from the Group schema functions and the callback structure of the Mongoose find() functions was problematic.

Any JS advice here?

website/server/libs/groupTasks.js
```diff
@@ -1,24 +1,73 @@
...
+      let assignedUsers = await Users.find({
+          $or: userList
+        }).exec();
+      for (let i=0; i < assignedUsers.length; i++) {
+        let promises = [];
+        promises.push(group.syncTask(masterTask, assignedUsers[i]));
+        promises.push(group.save());
+        await Promise.all(promises);
+      }
```
MeanderingCode: There are a mix of Promises and async functions in the codebase. Using Promises from the Group schema functions and the callback structure of the Mongoose find() functions was problematic.

Any JS advice here?

website/server/libs/groupTasks.js
```diff
@@ -1,24 +1,73 @@
...
+    // Complete or uncomplete the task on other users' lists
+    await Tasks.Task.find({
+      'group.taskId': groupMemberTask.group.taskId,
+      $and: [
+        {userId: {$exists: true}},
+        {userId: {$ne: groupMemberTask.userId}}
+      ]},
+      function (err, tasks) {
+        // @REVIEW How does Habitica handle errors?
+        if (err) return;
```
MeanderingCode: WIP. How does Habitica handle errors?

website/server/libs/groupTasks.js
```diff
@@ -1,24 +1,73 @@
...
+
+        tasks.forEach (task => {
+          // Ajdust the task's completion to match the groupMemberTask
+          // @REVIEW Completed or notDue tasks have no effect at cron
+          // This maintain's the user's streak without scoring the task if someone else completed the task
+          // If no assignedUser completes the due daily, all users lose their streaks at their cron
+          // An alternative is to set the other assignedUsers' tasks to a later startDate
+          // Should we break their streaks to encourage competition for the daily?
+          task.completed = groupMemberTask.completed;
```
MeanderingCode: This is the main point of feature discussion.
I based my implimentation (complete other users' dailies) on cron: https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/cron.js#L310

The implications of completing the task are that:
* The user who completes the Daily will score the task
* Other users will see the Daily is completed, will not score the task, will not lose their streak or take damage
* Other users will have their MP replenishment math include the done daily
* Other users "uncompleting" the daily has a bug in scoring (see https://github.com/HabitRPG/habitica/pull/11179#issuecomment-494096089 and following)

More discussion on how the feature should work may be found in #10690

MeanderingCode:
I am not sure I am done with this feature, but wanted to put a draft up to start getting feedback.

website/server/libs/groupTasks.js
paglias: This should be converted to a promise (see the comment above about using `exec()`) and in that case errors will be handled automatically (unless they need some special handling but it doesn't seem the case)

website/server/libs/groupTasks.js
paglias: Hi @MeanderingCode ! Can you link me to where callback functions are used?

The `mongoose.find` function by default uses callbacks but if you call `.exec()` on it at the end a promise will be returned

paglias (commented):
@MeanderingCode thanks for the PR! I left a couple of comments about promises and callbacks but regarding the PR feature I'm going to assign it to @SabreCat since he's working on Group Plans

website/server/libs/groupTasks.js
MeanderingCode: Ah, I realize I've seen `.exec()`s in most places but don't now recall seeing callbacks passed to mongoose.

website/server/libs/groupTasks.js
MeanderingCode: Makes sense.

website/server/libs/groupTasks.js
MeanderingCode: Given that, should I be `await`ing the Promises' resolution, as above and in the `findById()` for the Group on line 29, or using `Promise.then()`?

For the Group `findById()`, `exec()` seems clean. I don't like the `for` loop on the `assignedUsers`. It seemed to need to execute serially with `await` on the Promises to avoid mongoose `ParallelSaveErrors`, ~but I see no other uses of `await` in the server side code.~

**EDIT:** Not sure why I didn't see them before. Found several `await`s.

MeanderingCode: Pushed rework. Commenting on it.

website/server/libs/groupTasks.js
MeanderingCode: Pushed rework.

website/server/libs/groupTasks.js
```diff
@@ -1,24 +1,68 @@
...
+        userList.push(query);
+      });
+      await Users.find({
+        $or: userList
+      }).then(async assignedUsers => {
+        for (const assignedUser of assignedUsers) {
+          let promises = [];
+          promises.push(group.syncTask(masterTask, assignedUser));
+          promises.push(group.save());
+          await Promise.all(promises);
```
MeanderingCode: @paglias If I'm understanding Promises and async correctly, this series of `await Users.find().then(async {await Promise.all()})` will:
* Resolve the Promise returning an array of Users
* Call the `async` arrow function on the array of Users
* Loop over that array, but `await` the `Promise.all()` resolution before the next loop tick
* Finally return to the `await` of the `Users.find().then()`

I have not written much JS, and not used these structures.

MeanderingCode:
In more reworking and testing, I have discovered a bug. If UserA completes a task and UserB realizes the task is not done and "uncompletes" it, UserA's streak remains unchanged, but UserB's streak decriments.

I need to look into how to avoid that. My first thought is to look into the task history property of the various copies of the task. This can only apply to Dailies, as single completion shared Todos are deleted from the other users' tasks.

MeanderingCode:
Having read more of the scoring, it looks like there are two approaches; one is messy and the other requires some careful work in https://github.com/HabitRPG/habitica/blob/develop/website/common/script/ops/scoreTask.js#L187

In the messy approach, `groupTasks.js` could:
1. If a task is completed, copy the user's last history entry, add their userId, and append it to the master task's history
2. If a task is "uncompleted", read the userId from the last history entry in the master task, remove the history entry, and if that userId is *not* the user who "uncompleted" the task: score down the user who completed it and score up the user who "uncompleted" it.

This is messy b/c it shows down and up scores for the "correcting" user and may not stay up to date if code changes in other areas.

It would be far better to adjust `scoreTask()` so that it had logic to handle group shared tasks. This also opens the door for using scoring, streaks, etc on group shared tasks.

Alternatively, an earlier approach I had thought of could be taken: Instead of completing the dailies on other assignedUsers' lists, simply set a startDate of tomorrow, making them  grey (not due). It doesn't tell the user the task has been done (mobile apps can't see the group task board), but it does tell them they don't have to do it.

In looking through code for this, it appears challenges have a lot of the features I'm looking for (though not all), and I am very curious to know more of the discussion or design thinking going into groups. I have heard of some redesign talk, but I have not been able to find anything useful in trello or github about it.

@SabreCat , perhaps you have insight or can point me at where such discussion may be happening?

MeanderingCode:
There appear to be many things at play here.
* Modifying the `task.startDate` could have adverse effects on shared scheduling of `everyX days` dailies
* Modifying the `task.startDate` requires integration with `syncTask`
* Marking other users' `task.completed` has implications on their MP (actually okay, i think), but has side effects I'm looking into WRT "uncompleting" tasks
* Perhaps the best option is to integrate checking for group completion in the `shouldDo()` function. Still investigating: https://github.com/HabitRPG/habitica/blob/develop/website/common/script/cron.js#L98

I am sorry to be having a one-sided design discussion here. I continue posting as I'm learning in case someone wants to weigh in or has knowledge that may point to the right direction.

paglias:
Hello again @MeanderingCode ! I just wanted to let you know that we've not forgot about this PR and we're going to discuss it in a couple of days and let you know :)

SabreCat:
Hi @MeanderingCode--thanks for diving into this! A few comments and questions for you...

1. Gem awards are out of scope for addressing this ticket, and not something we'd want anyone starting on at this time. It's a clever idea, though, and something I'd recommend including in an email to admin@habitica.com as Group Plans feedback!
1. Similarly, if the "uncomplete" flows (or streak syncing, etc.) are proving to be a hassle, I'd recommend implementing only shared completion of Dailies in this PR. We can get that feature functional and polished before moving on to another workflow!
1. Do I understand correctly that so far, this only implements the "single" completion condition for Dailies, not "all assigned"?
1. Re how to mark individual users' Dailies as done after the master task is done via single-completion, I would definitely prefer setting the `completed` status rather than adjusting the Start Date. The Start Date of a task can be used in meaningful ways by Group Plan managers/leaders to note when a daily responsibility begins, information that would be lost if overwritten when the task is done on a specific day.

MeanderingCode:
Hi @SabreCat. Thanks for taking a look! My family is very excited to have this feature, so I'd love to move this forward.

>     1. Gem awards are out of scope for addressing this ticket, and not something we'd want anyone starting on at this time. It's a clever idea, though, and something I'd recommend including in an email to [admin@habitica.com](mailto:admin@habitica.com) as Group Plans feedback!

I agree it's an enhancement that's beyond this PR. I'll email about it!

>     2. Similarly, if the "uncomplete" flows (or streak syncing, etc.) are proving to be a hassle, I'd recommend implementing only shared completion of Dailies in this PR. We can get that feature functional and polished before moving on to another workflow!

I will make the cut where it makes sense. I do not want to disallow "uncompleting", because it could prove problematic if parties are relying on the status of tasks which need to get done. More below.

>     3. Do I understand correctly that so far, this only implements the "single" completion condition for Dailies, not "all assigned"?

No, this works for both "single" and "all" shared completion types, for both completing and "uncompleting".

>     4. Re how to mark individual users' Dailies as done after the master task is done via single-completion, I would definitely prefer setting the `completed` status rather than adjusting the Start Date. The Start Date of a task can be used in meaningful ways by Group Plan managers/leaders to note when a daily responsibility begins, information that would be lost if overwritten when the task is done on a specific day.

That echoes my own intuition and thinking, without specific knowledge of how it is used.

With regards to the bug where UserB "uncompletes" a task UserA erroneously completed and the scoring/streak discrepancies, I think this can fit into an enhancement workflow around group task scoring. If UserB "uncompletes" a task UserA completed, they should be able to reverse the damage by completing the task. They won't get credit for doing it, but neither will they get damage. Leaving "uncomplete" in allows for the user who made the accidental click to rectify the mistake.

With enhancements [later] for a group scoring feature, the necessary enhancements will make it easier to do things like changing the `shouldDo()` and `setNextDue()` logic.

If all that is agreeable, I will finalize this PR and mark it ready for review!

MeanderingCode:
Addressed all but two linting concerns:
* Indentation is as it should be, as far as I think. I could move the `.then(task => {` to a new line at the same indentation as the line it is currently on, which may address the linter's concerns, though I find mixed usage in the code. Thoughts?
* What about the `await` in the `for` loop? It is done to address the `ParallelSaveError`s from mongoose. Please comment on above code comment thread on this topic.

MeanderingCode:
There is, indeed, a bug with approval. I will retitle this PR with "WIP". Please refrain from merging.

website/server/libs/groupTasks.js
```diff
@@ -1,24 +1,68 @@
...
+      });
+    }
+  } else {
+    // Complete or uncomplete the task on other users' lists
+    await Tasks.Task.find({
+      'group.taskId': groupMemberTask.group.taskId,
+      $and: [
+        {userId: {$exists: true}},
+        {userId: {$ne: groupMemberTask.userId}}
+      ]}).then(tasks => {
```
paglias: why is then used here? Instead of something like `const tasks = await Tasks.Task.find(..) ` and then iterate on them?

paglias: why is then used here? Instead of something like `const tasks = await Tasks.Task.find(..) ` and then iterate on them?

website/server/libs/groupTasks.js
```diff
@@ -1,24 +1,68 @@
...
+        tasks.forEach(task => {
+          // Adjust the task's completion to match the groupMemberTask
+          // Completed or notDue dailies have little effect at cron (MP replenishment for completed dailies)
+          // This maintain's the user's streak without scoring the task if someone else completed the task
+          // If no assignedUser completes the due daily, all users lose their streaks at their cron
+          // @TODO Should we break their streaks or increase their value to encourage competition for the daily?
+          task.completed = groupMemberTask.completed;
+          task.save();
```
paglias: async ops and array.forEach don't really work well together. You'll want to use a map that return a promise. So for example

```
const tasks = await Tasks.Task.find(...)
await Promise.all(tasks.map(t => {
    ... do stuff with each task...
    return task.save() // return a promise that can be awaited
}))

File: website/server/libs/groupTasks.js [hunk 8]
```diff
@@ -1,24 +1,68 @@
...
+      let group = await Groups.findById(masterTask.group.id).exec();
+      let userList = [];
+      masterTask.group.assignedUsers.forEach(userId => {
+        let query = {_id: userId};
+        userList.push(query);
+      });
+      await Users.find({
+        $or: userList,
+      }).then(async assignedUsers => {
+        for (const assignedUser of assignedUsers) {
```
paglias: the problem with this for is that it will run sequentially instead of in parallel.

My suggestion would be to only use await and not `then` to start with as it's very error prone

paglias (changes requested):

website/server/libs/groupTasks.js
MeanderingCode: Resolve duplicate comment

website/server/libs/groupTasks.js
MeanderingCode: Due to the `syncTask()` function returning a `Promise.all()` that includes a `masterTask.save()`, they must be done in sequence and waited on or mongoose will throw a `ParallelSaveError` because several asynchronous calls will be trying to update the same document. This sort of emulates the client making separate API calls to sync each assigned user. The only alternative I know is to reimplement the `syncTask()` function. There may be a way to massage mongoose to pipeline the saves, since sequencing the already in memory document saving doesn't prevent the kind of overwrites it is worried about, anyway, given that they are passed to the function as opposed to loaded within it.

I do not think the `group.save()` needs to happen after each user sync, though. That can likely come out of the loop and happen once.

Also, see the other comment thread asking about `then` vs `await` and iteration.

website/server/libs/groupTasks.js
MeanderingCode: It seemed an appropriate pattern, though I do not have a lot of experience here.

Why not use `then`, and what's better about splitting it out like that? In another comment you mention that it is error prone. How?

website/server/libs/groupTasks.js
MeanderingCode: Elegant use of `await Promise.all(tasks.map())`! Thanks!
It also makes a clearer case for not using the `.then()`, though I'm still curious in the thread above.

website/server/libs/groupTasks.js
MeanderingCode: Okay, I had an aha moment, but it is incomplete...
I do not totally understand the difference, but applied the pattern and it worked. I will take the question to Stackoveflow, since it is beyond the scope of this PR.

Thanks!

website/server/libs/groupTasks.js
paglias: Sorry, I didn't explain myself well.

Using `.then` with promises isn't error prone at all but it's the same as doing `await promise` and when you start using async/await or then/catch it's better to only use one method only to avoid confusion. We prefer async/await as it's usually easier to follow and read the code

website/server/libs/groupTasks.js
paglias: > I do not think the group.save() needs to happen after each user sync, though. That can likely come out of the loop and happen once.

That would be great

> they must be done in sequence and waited on or mongoose will throw a ParallelSaveError because several asynchronous calls will be trying to update the same document

Database operations are usually quite slow, is there a way to only update each task once?

MeanderingCode: The `group.taskSync()` function calls `.save()` on the `masterTask` passed to it. It does this because it is used (the only place, afaict) to assign a user to a task, and so will push the user into the `task.assignedUsers` array.

https://github.com/HabitRPG/habitica/blob/develop/website/server/models/group.js#L1393

I see two option to optimize this for the purpose of being used both by the API and by this code (as well as hooking to saving the group task, which I'd like to do later):

1. Move `.push()`ing the `taskToSync` (aka `masterTask`, in this context) into the conditional that checks to see if the `user._id` is in the assigned users (line 1309). This strikes me as suboptimal, since it slightly complicates the logic of the function, and there could be some strange edge case if there is a race condition with un-assigning users and having assigned users un/completing tasks.
2. Move the assign/unassign into it's own function in `group.js` which saves the master task and then calls `syncTask()` and have the API handler call that. This seems far cleaner.

For optimization, I will take the second approach and push something tomorrow.

website/server/libs/groupTasks.js
MeanderingCode: Ah, this makes complete sense.

website/server/libs/groupTasks.js
MeanderingCode: In the end, I used that pattern for re-adding "uncompleted" todos, but optimized this block far more by using mongoose's `updateMany()` instead of `find`ing Users, modifying them, and `save`ing them.

website/server/libs/groupTasks.js
MeanderingCode: I refactored the group model's `syncTask` to have a separate `linkTask` that will call it (and updated the API handler). This allowed me to effectively parallelize this section.

Thanks for the insights!!

MeanderingCode:
@SabreCat I have done a significant amount of work to improve logic around Single Completion Shared Dailies to take into account users' timezones and custom day starts. This now treats completion as happening in the user's "day", and will set the completed status of users who are in the "same day" and have already cron'd for the day. It also adds a function to cron to check the master task history and do calculations about what "day" the history entry happened on, based on the completing user's TZ and CDS, and determines if it was completed in the cron'ing user's "day".

I _think_ this logic advances this work to the point where it will work as expected for users croning at different times and/or in disparate timezones, and I think it is ready for a re-review.

Some points:

* Date logic gets complex when interacting between users. Please check closely all changes regarding day comparisons (cron, groupTasks, scoreTask)
  * Particularly with `scoreTask`, that runs on the client as well as the server. Please think through any discrepencies
* Date and Moment seem to interact well as I have used them, but I am unfamiliar with Moment. Please check uses carefully.
* To check for shared task completion on cron, I had to introduce an async function into a previously synchronous cron function. Commenting directly on the line.
* Group Task List Dailies completion is not reset on any cron. I must still decide when this should happen (first user cron? enhance third party tools as stopgap to see history for parties with "split" days?)
* I still need to write more tests

This feature grew to be a lot larger than I had anticipated due to nuances of differences in users' timezones, day starts, and cron and usage/checkin. I imagine there will need to be some extensive testing, but all of my local testing now shows "expected behaviors".

website/server/libs/cron.js
```diff
@@ -381,7 +383,19 @@ export function cron (options = {}) {
...
+    // If this is a shared task, check if another user completed it in the "same" day the user is "starting"
+    if (task.group && task.group.sharedCompletion === SHARED_COMPLETION.single) {
+      // @REVIEW This introduces an async call into this cron function.
+      // The function is called from ../middlewares/cron.js asyncCron(), which is async and suggests this is okay
+      // Does this cause issues?
+      // Pushing these closures to an array of Promises outside the forEach
+      // This allows both sequential processing and async checking for shared completion
+      groupSharedSingleDailies.push(async function determineGroupCompletion (memberTask, memberUser, memberTime) {
+        memberTask.completed = await groupTaskCompleted(memberTask, memberUser, memberTime);
+      }(task, user, now));
```
MeanderingCode, line 395: The function this is inside (`cron (options = {})`) was synchronous. It was called from the asynchronous `cronAsync()` in the `website/server/middlewares/cron.js` file.

1. I do not know the ramifications of `await`ing this cron call from the middleware. On the surface, it should only increase the time to process, and only if the user has Single Completion Shared Dailies.
2. I tried to make this pattern of pushing an executed asynchronous closure to an array `await`ed outside the `forEach` loop more elegant with an arrow function, but it never executed. If there are suggestions about making this more elegant in any way, please let me know. I wanted to keep the sequential `forEach` for two reasons:
    * It keeps the pattern used to iterate the other task types
    * It keeps the logic sequential. Since there are variables outside the loop scope that are modified and checked, this will avoid tricky issues now and in the future.

website/common/script/ops/scoreTask.js
```diff
@@ -277,22 +278,45 @@ module.exports = function scoreTask (options = {}, req = {}) {
...
+            // This Daily could have been completed "today" by someone else.
+            // Since this function runs on the client, as well as the server, let's make assumptions and keep things simple
+            // Otherwise we have to load the Master Task and check who did what when
+            let o = sanitizeOptions(user.preferences);
+            let startOfDayWithCDSTime = startOfDay(o);
+            if (moment(historyEntry.date).isBefore(startOfDayWithCDSTime)) {
+              // This Daily was completed by us before "today"
+              // It was probably completed "today" by someone else
+              task.history.push(historyEntry);
+              throw new NotAuthorized(i18n.t('messageOtherAssigneeCompleted', req.language));
```
MeanderingCode, line 304: This catch for "uncompleting" a daily another assigned user completed makes a significant assumption. Please review.

MeanderingCode:
It looks like I broke several things, according to the tests. I will look into them, though anyone is welcome to do some reviewing, anyway.

website/common/script/ops/scoreTask.js
paglias: Just to make sure it's clear, given the `throw` here the function will stop running and nothing will be saved

MeanderingCode:
Yes, the intention is that you cannot uncomplete a daily another assigned
user completed.

paglias:
Ok then what's the point of pushing something in the task history at the
line before if it'll never be saved due to the error?

Il Lun 10 Giu 2019, 13:25 Sean Leonard <notifications@github.com> ha
scritto:

MeanderingCode:
Good point. I'll remove, as well as clean up comments slightly if the
approach seems valid.

MeanderingCode:
@paglias 

> Ok then what's the point of pushing something in the task history at the line before if it'll never be saved due to the error?

I had done that, as well as moved the check before the `task.completed = false`, because the task remained "unchecked" after the `throw`, and I didn't want to risk a modified client-side history to be inadvertently saved to the server in some later send. Is that a risk? What do you think?

---
_Aside:_ I wish GH email reply comments properly nested under the thread you started.

MeanderingCode:
The test failures were mostly due to needing to change the cron tests to async/await to match my change to the cron function. Another test (one of mine) needed to call a function for more depth regarding task completion, and one potentially good catch led to a defensive logic change for potential edge cases.

I think this is ready for review, @paglias and @SabreCat.

website/common/script/ops/scoreTask.js
```diff
@@ -277,22 +278,38 @@ module.exports = function scoreTask (options = {}, req = {}) {
         task.completed = true;
 
         // Save history entry for daily
+        // If this is a yesterDaily being scored, set the date to the last minute of the user's "yesterday"
+        let dateScored = task.yesterDailyScored ? startOfDay(user.preferences).subtract(1, 'minute') : moment();
```
Alys, line 282: @MeanderingCode If I understand this correctly, `task.yesterDailyScored` is true if and only if the Daily is being marked complete in the "Check off any Dailies you did yesterday" screen - is that right? 

Also, the code later under the comment `// Set whether we're scoring a yesterdaily` -- that code is executed if the user is completing the Daily in the "Check off any Dailies you did yesterday" screen, right?

MeanderingCode: @Alys Yes. I purposefully did not as the property to the model, so it will never be persisted. It is set in the browser for all (loaded) tasks collected by the yesterdailies modal, then the new optional route segment is added to the score API endpoint. On the server, the property is set on the loaded task if the optional segment is included, and is thus available to all functions where the task is passed as a parameter (and not reloaded from the DB).

Alys: @MeanderingCode Cool, that all sounds good!

I think the "yesterdaily" terminology makes that a little confusing though (not your fault). A Daily's `yesterDaily` attribute is true to indicate that the Daily should appear in the "Check off any Dailies you did yesterday" screen if it was due but not complete, so strictly speaking "a yesterDaily" would be any Daily that's **capable** of appearing in that screen, not necessarily one that is currently being marked complete in that screen. Because of that, when I first looked through your new `yesterDailyScored` code, I thought it was being applied to all such Dailies (which would have been a bug) but I see now that it's applied only to those being completed in the screen (which is good). Maybe the comments could be adjusted slightly to prevent other devs having the same initial misconception I had?

For example, this:
	// If this is a yesterDaily being scored, set the date to the last minute of the user's "yesterday"
could be adjusted to something like this:
	// If this is a yesterDaily being scored in the RYA screen, set the date to the last minute of the user's "yesterday"

("RYA" or "Record Yesterday's Activity" is the official name of the "Check off any Dailies you did yesterday" feature, although unfortunately the name doesn't appear anywhere on the website any more.)

Also this:
	// Set whether we're scoring a yesterdaily
could perhaps become something like this:
	// Set whether we're scoring a yesterdaily in the RYA screen.

And:
// Insert history at the right point, since this could be a yesterdaily scoring after someone else's today scoring
could be:
// Insert history at the right point, since this could be a yesterdaily scored in someone's RYA screen after someone else's today scoring

This is all assuming I'm understanding it correctly now. :) How do you feel about that? If you think my suggested changes to the comments aren't correct, you should definitely say so!

SabreCat:
FYI we're putting this through some QA with the help of volunteer testers from the community. @MeanderingCode you've got it tagged as WIP still--is there more you plan to implement yet?

MeanderingCode:
Hey @sabrecat. Glad to hear it!

As the description states, I need to add some more tests, add history to "All" type tasks (simple), and decide when and how to cron the group dailies. If I don't cron them, the group task list dailies stay completed (though individuals' linked tasks cron with them).

website/common/script/ops/scoreTask.js
MeanderingCode: Updated comments for clarity. Thanks for the suggestion!

MeanderingCode:
Hi @SabreCat. Any updates on how testing is going? Any bugs?

I will try to write the tests soon, and am happy to push the additional things to another PR if it makes this easier.

SabreCat:
I checked in with the crew again today, and while no bugs have been reported, it doesn't sound like intensive testing's been done--the trials of being a small team with no dedicated QA staff, tch. We're going to set aside one of our regular meeting times to get together and try to break things in the next few days, though!

SabreCat:
Hi @MeanderingCode -- I committed a fix to one issue we ran across, but other than that the functionality has been working well! It might make sense to wrap up the tests here and push any further nice-to-have functionality to another PR after we merge.

paglias:
Any update on this?

paglias:
Hi @MeanderingCode, we were wondering if you plan to add the missing tests or if you need help with them! If you can't no worries and we'll add them ourselves :) thanks and let us know

MeanderingCode:
@paglias @SabreCat I had planned on adding them, but I have had no time for it in a long while.

When I get home, I can copy my notes on what tests I was planning to make it easier for someone to know what to cover. If no one picks it up, I may find time in the next couple weeks.

paglias:
@MeanderingCode no worries, and yeah if you can post here the list of what you planned to cover it would be very useful!

SabreCat:
@MeanderingCode @paglias Took a stab at adding the missing group daily uncomplete at cron in 14d68e0 . Pretty first-drafty, thoughts?

@MeanderingCode Getting anywhere with the checklist of tests to add?

~Also after merging latest `develop` a relevant test is failing? 🤔 Odd...~ Oh, guess that was flukey.

paglias:
left some comments!

paglias:
I'm closing this in favor of #11390 (needed because of a change in the order of PRs that will be merged)

PR #11226, opened 2019-06-15
Performance/inbox paging ui

negue:
fixes based on: #11157

- adds `.fromUUID`, prevent resetting `.uuid`
- fix correct `canLoadMore`/`page` state between conversations


website/server/controllers/api-v4/inbox.js
```diff
@@ -95,16 +95,19 @@ api.conversations = {
...
     newChat.toUser = newChat.user;
     newChat.toUserName = newChat.username;
     newChat.toUserContributor = newChat.contributor;
     newChat.toUserBacker = newChat.backer;
-    newChat.uuid = user._id;
+    newChat.fromUUID = user._id;
     newChat.user = user.profile.name;
     newChat.username = user.auth.local.username;
     newChat.contributor = user.contributor;
     newChat.backer = user.backer;
```
Alys, line 108: It looks like this code is re-mapping other attributes too, for example, `user`. Is that going to cause similar problems? Even if it doesn't, will it be confusing to future coders if `user`, `username`, etc mean one thing in the schema and something different in the code?
It might also create extra work for Comrades when (if?) API v4 becomes the official API for third-party tools (they'll have to rewrite code that might not have needed rewriting if the meaning of the attributes hadn't changed).

If mapping is helpful for the inbox vue code to be clearer (highly likely since the schema uses some confusing naming), would it be possible to map to completely new attribute names? For example:
```
newChat.fromProfileName = user.profile.name;
newChat.fromUsername = user.auth.local.username;
newChat.fromContributor = user.contributor;
```

In fact, if there was remapping for both the `from` and `to` data using attribute names that are consistent with how we think about users in other sections of the site (e.g., `newChat.toProfileName` and `newChat.fromProfileName` instead of `newChat.toUser` and `newChat.user`) would that make the inbox vue code clearer? It would however increase the size of the data returned by the API. :(

negue: @paglias @SabreCat any thoughts on this? should I also add the other properties  like `fromProfileName` and so on?

paglias: @Alys @negue sorry for the delay, so the issue here is to make sure the mobile apps don't break, for now I'd say to only add the fields necessary but before we make API v4 official we should definitely revisit this

negue: @phillipthelen any other properties needed to be reverted / converted to `from*` ?

phillipthelen: Sorry for the late response, I somehow missed the email.

From what I can tell, `user`, `username`, `contributor`, `backer` were all changed in the same way? If yes, those will also have to remain with the old behaviour, since the apps use those fields.

PR #11228, opened 2019-06-16
Free group plan subscription from a party/guild no longer gets cancelled after leaving a guild - fixes #11055

### LINKED ISSUE #11055: free Group Plan subscription from a Party should not be cancelled when leaving a Guild
If you have a free subscription because you are a member of a **Party** that has a Group Plan, leaving any Guild will cause your free subscription to be terminated. That should not happen. Your free subscription should remain for as long as you are a member of one or more Guilds/Party that still has an active Group Plan.

Whenever someone with a free subscription leaves a Guild or Party, there is code that is executed to check whether you're still in any other groups that have an active Group Plan (and if you are, your free subscription is not removed), but it's clearly not working for the case where the Group Plan is in a Party. It _might_ also not be working when the Group Plan is in a guild - that should be tested as part of the fix for this although I've seen no evidence that it's not working in that case. Below are pointers to parts of that code; the first one is in the `api.leaveGroup` route (`/api/v3/groups/:groupId/leave`):

https://github.com/HabitRPG/habitica/blob/ee20b1eea8d761f16ba6ba3aba715f3f89a2467f/website/server/controllers/api-v3/groups.js#L780-L783

https://github.com/HabitRPG/habitica/blob/ee20b1eea8d761f16ba6ba3aba715f3f89a2467f/website/server/models/user/methods.js#L392-L398

https://github.com/HabitRPG/habitica/blob/ee20b1eea8d761f16ba6ba3aba715f3f89a2467f/website/server/models/user/methods.js#L361-L372

https://github.com/HabitRPG/habitica/blob/ee20b1eea8d761f16ba6ba3aba715f3f89a2467f/website/server/models/user/methods.js#L42-L48

I've replicated this bug three times. I'm marking it as important priority for a fast fix.

If any contributors want to work on this, you should definitely post here first to say you're interested and wait for a response before you start. That's a good idea for any issue but it's especially important for this one because the staff might decide to fix it themselves - asking first will save your time if the staff haven't yet updated this issue.
In [Using Your Local Install to Modify Habitica%27s Website and API](https://habitica.fandom.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API), you can find information about how to add a free subscription from a Group Plan to any user, and how to add a Group Plan to a Guild or Party. 

Note that when testing this bug or the code that fixes it, it's important that your test user isn't in any Guilds that still have a Group Plan - if it is, that will likely ensure that your test user keeps its free subscription (that's what's supposed to happen).

_As a note for myself: After the fix for this goes live, the emails that are sent when your free Group Plan subscription is removed for any reason should be tested (either I'll test them myself or make a note somewhere for others to test them). The reason is that at the time you are affected by this bug, you receive an email saying that the leader of the Guild you just left has cancelled the Group Plan in that guild, which is clearly not correct in this situation. There might be other bugs that send you the wrong email when you lose your free subscription for legitimate reasons._

doublej89:

Fixes #11055

### Changes

If a user attempts to leave a guild, the code now checks whether that user is a member of at least one other group (party/guild) with an active group plan. If the check succeeds, the user's free subscription will not get cancelled. 

Previously, the buggy code would check if _all_ of the user's groups had a group plan. So if just _one_ of those groups did not have a group plan, the check would fail and the user's free subscription would get cancelled.

That doesn't happen anymore.

The fix was made by simply replacing the ```.every()``` with the ```.some()```  array method in the group membership checking function.

Tests have been added to ensure that when leaving a group a user doesn't have to worry about losing their free subscription as long as they are a member of at least one group plan.

An additional bug fix is in the code that handles cancelling group subscription for users. The code used to incorrectly check user membership in a group with a group plan using the query condition ```'purchased.plan.dateTerminated': null```. This would return not only groups that have a ```'purchased.plan.dateTerminated'``` field with ```null``` value, but also groups that don't have a ```'purchased.plan.dateTerminated'``` field.

To make sure the query only returns groups that have a  ```'purchased.plan.dateTerminated'``` field with ```null``` value, the query condition should be ```'purchased.plan.dateTerminated': { $type: 'null' }```. That was the fix.   

----
UUID: 
395c7c76-da35-4145-85dd-783bf753e3ea

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10753,14 +10753,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
-          "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
-          "optional": true
+          "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c="
         },
         "brace-expansion": {
           "version": "1.1.11",
           "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
           "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
@@ -10775,20 +10773,17 @@
         "code-point-at": {
           "version": "1.1.0",
           "resolved": "https://registry.npmjs.org/code-point-at/-/code-point-at-1.1.0.tgz",
-          "integrity": "sha1-DQcLTQQ6W+ozovGkDi7bPZpMz3c=",
-          "optional": true
+          "integrity": "sha1-DQcLTQQ6W+ozovGkDi7bPZpMz3c="
         },
         "concat-map": {
           "version": "0.0.1",
           "resolved": "https://registry.npmjs.org/concat-map/-/concat-map-0.0.1.tgz",
-          "integrity": "sha1-2Klr13/Wjfd5OnMDajug1UBdR3s=",
-          "optional": true
+          "integrity": "sha1-2Klr13/Wjfd5OnMDajug1UBdR3s="
         },
         "console-control-strings": {
           "version": "1.1.0",
           "resolved": "https://registry.npmjs.org/console-control-strings/-/console-control-strings-1.1.0.tgz",
-          "integrity": "sha1-PXz0Rk22RG6mRL9LOVB/mFEAjo4=",
-          "optional": true
+          "integrity": "sha1-PXz0Rk22RG6mRL9LOVB/mFEAjo4="
         },
         "core-util-is": {
           "version": "1.0.2",
@@ -10905,8 +10900,7 @@
         "inherits": {
           "version": "2.0.3",
           "resolved": "https://registry.npmjs.org/inherits/-/inherits-2.0.3.tgz",
-          "integrity": "sha1-Yzwsg+PaQqUC9SRmAiSA9CCCYd4=",
-          "optional": true
+          "integrity": "sha1-Yzwsg+PaQqUC9SRmAiSA9CCCYd4="
         },
         "ini": {
           "version": "1.3.5",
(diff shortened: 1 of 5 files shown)


paglias:
Hi @doublej89 and thanks for the PR! Sorry for the delay here but during this week I'll make sure to review it

paglias:
@doublej89 could you add some tests to make sure the functionality is working as intended? let me know if you need any help

doublej89:
@paglias Sorry for the delay as well! I've added some tests. Hope it helps. Let me know if more needs to be done.

Alys:
@doublej89  I've done some tests on my local install to compare the behaviour between `develop` and this PR and this is looking great! From everything I can see, you've fixed the bug described in https://github.com/HabitRPG/habitica/issues/11055. I appreciate that you've also made the code improvements I suggested in that issue - thank you!
Below is a quick-and-dirty description of the tests that I did and the results. If anyone wants me to provide more details about the tests, just ask.

I have not done a code review; the staff will do that.


Party has G.Plan
User is also a member of two guilds with no G.Plan
Leave one guild - `develop` & `PR 11228`: keep subscription (correct)
Leave your last guild - `develop`: lose subscription (bug); `PR 11228`: keep subscription (correct - **bug fixed**)

Party has G.Plan
Guild has G.Plan
Leave party - `develop` & `PR 11228`: keep subscription (both correct)

Party has G.Plan
Guild has G.Plan
Leave guild - `develop`: lose subscription (bug); `PR 11228`: keep subscription (correct - **bug fixed**)


Guild has G.Plan
User is also a member of two guilds with no G.Plan
Leave a non-G.Plan guild - `develop` & `PR 11228`: keep subscription (both correct)
Leave your last guild without G.Plan - `develop` & `PR 11228`: keep subscription (both correct)
Leave party - `develop` & `PR 11228`: keep subscription (both correct)

Guild has G.Plan
User is also a member of two guilds with no G.Plan
Leave a non-G.Plan guild - `develop` & `PR 11228`: keep subscription (both correct)
Leave party - `develop` & `PR 11228`: keep subscription (both correct)
Leave your last guild without G.Plan - `develop`: lose subscription (bug); `PR 11228`: keep subscription (correct - **bug fixed**)

Guild has G.Plan
Second guild has G.Plan
Leave one guild - `develop`: lose subscription (bug); `PR 11228`: keep subscription (correct - **bug fixed**)

paglias:
thanks @doublej89 the tests look good and thanks @Alys ! Just one last thing before we can merge this, could you undo the changes to package-lock.json? thanks!

doublej89:
@paglias done. package-lock.json should now be the same as in the main repo

doublej89:
@paglias hmm, coverage seems to have gone down a bit. Let me know if more tests need to be written

paglias:
Hi @doublej89 ! Don't worry about the coverage, sometimes it's buggy. I'm currently on vacation so I'll merge this when I'm back on the 9th or 10th.

SabreCat:
Thanks for this, @doublej89, and welcome to the ranks of the Blacksmiths at tier 1! This will go into a staging environment for a few days, and you should see it live on the production site around this time next week.

Note that subsequent contributor tiers take increasing amounts of effort from one to the next, so Blacksmith 2 will require multiple PRs or a larger PR to attain. But keep helping out and we'll express our gratitude accordingly!

PR #11230, opened 2019-06-16
Fix task dropdown menu overflow

### LINKED ISSUE #11227: website footer covers task three-dots menu (z-index problem?)
As ieahleen reports in the Report a Bug guild:

"It is not possible to access the three-dots-menu of the tasks at the bottom of the screen, the menu appear under the footer:"
image

It might be because the z-index of the footer is too large.

You can replicate this with a task that has no note. (You don't see the bug with a task that has a note because the note will push the bottom of the task down far enough that the three-dots menu doesn't reach the site footer.)

----

_edit by Alys 2020-02-02:_

This was fixed in https://github.com/HabitRPG/habitica/pull/11230 but has started happening again as shown in the recent comment below. That PR might contain useful information for fixing the bug again.

The bug is slight different now than originally: 
When you move your mouse away from the drop-down menu it becomes fully visible. It's only hidden by the footer when your mouse is over the menu, so the problem might lie in styles that are only used for the on-hover state.

freyacodes:
Fixes #11227

### Changes
image

While I am essentially just moving two lines of code, there is a lot of trial-and-error behind them. To make the dropdown menu properly overlap, I had to solve two obstacles:

1. The `container-fluid` class needs to allow overflow, otherwise it will clip the dropdown. This was accomplished by simply removing `overflow-x: hidden;`, reverting it to the default `visible`. What has this to do with the x-axis? No idea. CSS is magic. This breaks the layout, so I moved `overflow-x: hidden;` to `#app` instead.
2. For z-index to render over another div, the DOM elements needs to be in the same stacking context[1], as far as I understand. This is accomplished by explicitly setting the z-index of the footer-row. Here I can just move it from the header.

[1] https://philipwalton.com/articles/what-no-one-told-you-about-z-index/

----
UUID: de91221c-6b7e-4614-8854-c5c6a6d3ffe7

### DIFF
--- a/website/client/app.vue
+++ b/website/client/app.vue
@@ -57,6 +57,7 @@ div
     display: flex;
     flex-direction: column;
     min-height: 100vh;
+    overflow-x: hidden;
   }
 
   #loading-screen-inapp {
@@ -89,7 +90,6 @@ div
   }
 
   .container-fluid {
-    overflow-x: hidden;
     flex: 1 0 auto;
   }
 
--- a/website/client/components/appFooter.vue
+++ b/website/client/components/appFooter.vue
@@ -113,11 +113,11 @@
   .footer-row {
     margin: 0;
     flex: 0 1 auto;
+    z-index: 17;
   }
 
   footer {
     color: #c3c0c7;
-    z-index: 17;
     padding-bottom: 3em;
 
     a {


paglias:
Hi @Frederikam   and thanks for the PR! Sorry for the delay, we'll review it this week

paglias:
@Frederikam  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11235, opened 2019-06-18
Enable keyboard shortcuts for autocomplete @username selection in chat

### LINKED ISSUE #9291: @Username selection with keyboard arrow keys in drop-down mentions list in chat
When tagging someone's name in chat, you can no longer select the person's name using arrow keys. Instead, you need to use the mouse. You should be able to use arrow keys to select someone's name.

jackcogdill:

Fixes #9291 

### Changes

- Enable keyboard shortcuts for chat autocomplete
  - Tab/down to select next
  - Shift-tab/up to select previous
  - Enter to make selection
  - Escape to cancel
- Chat autocomplete mouse hover styling: use cursor pointer to feel more clickable.

----
UUID: 5ffee717-5618-4429-b161-981d79842f8a

### DIFF
--- a/website/client/components/chat/autoComplete.vue
+++ b/website/client/components/chat/autoComplete.vue
@@ -27,6 +27,7 @@
 
   .hover-background {
     background-color: rgba(213, 200, 255, 0.32);
+    cursor: pointer;
   }
 
   .hover-foreground {
--- a/website/client/components/chat/autoComplete.vue
+++ b/website/client/components/chat/autoComplete.vue
@@ -3,8 +3,8 @@
   .autocomplete-results.d-flex.align-items-center(
     v-for='result in searchResults',
     @click='select(result)',
-    @mouseenter='result.hover = true',
-    @mouseleave='result.hover = false',
+    @mouseenter='setHover(result)',
+    @mouseleave='resetSelection()',
     :class='{"hover-background": result.hover}',
   )
     span
@@ -87,6 +87,7 @@ export default {
         tier9,
         tierNPC,
       }),
+      selected: null,
     };
   },
   computed: {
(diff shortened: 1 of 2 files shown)


paglias:
Hi @jackcogdill and thanks for the PR! Sorry for the delay, we'll review it this week

website/client/components/chat/autoComplete.vue
```diff
@@ -3,8 +3,8 @@
   .autocomplete-results.d-flex.align-items-center(
     v-for='result in searchResults',
     @click='select(result)',
-    @mouseenter='result.hover = true',
-    @mouseleave='result.hover = false',
+    @mouseenter='resetSelection(); result.hover = true',
```
paglias: could `result.hover = true` be moved inside `resetSelection `?

jackcogdill: No. `resetSelection()` sets `hover` to `false` for all the results, so it wouldn't make sense.

jackcogdill: I suppose it's possible to pass result into the function as an optional parameter, and then set it inside the function. Would you prefer that?

paglias: yeah it would be better, we try to keep the js into templates to a minimum, thanks!

jackcogdill: @paglias I updated the code. Please take another look

paglias:
@jackcogdill thanks for the PR, this seems to be working very well and the code is fine as well. We're going to test this internally for about a week and if not issues arise it'll go live soon, I'll let you know if there's any problem

paglias:
@jackcogdill  I'm merging the PR and awarding your contributor tier.

PR #11237, opened 2019-06-18
Removed unnecessary "Equipment" sub-heading in Market.

### LINKED ISSUE #11232: unnecessary Equipment sub-heading in Market
As reported by @ieahleen (c073342f-4a65-4a13-9ffd-9e7fa5410d6b):

"In the Market there is an Equipment sub-header in the Items section, below Food and Saddles and above Cards, I don't think it should be there:"

scr

Scens:

Fixes #11232 

### Changes
Zrzut ekranu 2019-06-18 o 17 02 44

----
UUID: dd59e7c1-b76a-466e-b301-783b852da699

### DIFF
--- a/website/client/components/shops/market/filter.vue
+++ b/website/client/components/shops/market/filter.vue
@@ -3,11 +3,11 @@
   h2(v-once) {{ $t('filter') }}
   .form-group
     checkbox(
-      v-for="category in categories",
-      :key="category.identifier",
-      :id="`category-${category.identifier}`",
-      :checked.sync="viewOptions[category.identifier].selected",
-      :text="category.text"
+      v-for="viewOptionKey in Object.keys(viewOptions)",
+      :key="viewOptionKey",
+      :id="`category-${viewOptionKey}`",
+      :checked.sync="viewOptions[viewOptionKey].selected",
+      :text="viewOptions[viewOptionKey].text"
     )
   div.form-group.clearfix
     h3.float-left(v-once) {{ $t('hideLocked') }}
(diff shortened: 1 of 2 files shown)


Alys:
@Scens Thanks for looking into this! 

Can you confirm that removing equipment from the categories array doesn't have any other effects? For example, does the Equipment section still appear at the top of the Market page and does the equipment filter checkbox in the left-hand column still work?

Those are just a couple of examples to check. Have a look for every place in the code that the category array is used to see if there's anything else that might be affected. :)

Alys:
Those are just a couple of examples to check. Have a look for every place in the code that the category array is used to see if there's anything else that might be affected. :)

Scens:
> @Scens Thanks for looking into this!
> 
> Can you confirm that removing equipment from the categories array doesn't have any other effects? For example, does the Equipment section still appear at the top of the Market page and does the equipment filter checkbox in the left-hand column still work?
> 
> Those are just a couple of examples to check. Have a look for every place in the code that the category array is used to see if there's anything else that might be affected. :)

I've checked it at found out that the filter doesn't work - now it works :).
The issue was that the same array was used to create filters and the actual rows/components to be filtered.

The fact that the EquipmentSection component is not a generated category has complicated the matter. 
Screenshot 2019-06-19 at 22 49 02

Someone has tried to solve it by adding to category array "helper category" with "Equipment" text, so it displays in filter and added some ifs to the EquipmentSection. But then it also displays as a new CategoryRow, hence the topic of this issue.
Now viewOptions (previously used only to create filter checks) keeps additional info about text to be displayed in filter and the categories array is no longer passed to the filter component.

But I don't know why do coverage checks fail. Could you explain it to me?

paglias:
thanks @Scens ! the PR seems to be working well

paglias:
@Scens  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11239, opened 2019-06-23
add shadow-muting chat hiding feature (automatic flagging of public posts from shadow-muted players) - fixes 10851

### LINKED ISSUE #10851: shadow-mute feature is needed
Currently Habitica has a mute feature. We want to add an additional "shadow mute" feature. (This comes from a discussion involving beffymaroo and the moderators.)

---

Shadow muting will work in a similar way to muting so here's a brief description of how the existing mute feature works (this feature does not need any changes so this description is just FYI):
- The current mute feature uses the `flags.chatRevoked` attribute in the User model ([user/schema.js](https://github.com/HabitRPG/habitica/blob/9a3cdb5deb6af9054c6566634005bf174f63d880/website/server/models/user/schema.js#L225))
- If a player has that flag set to `true`, then:
   - They are unable to post to the Tavern or public guilds. When they try to they are shown an error message that tells them their chat permissions have been revoked.
   - They can't send Private Messages to any player. When they try to they are shown an error message.
   - They CAN post to their party or any private guild they are a member of.
- A moderator can turn the feature on and off from the [Hall of Contributors](https://habitica.com/hall/contributors) ([heroes.vue](https://github.com/HabitRPG/habitica/blob/9a3cdb5deb6af9054c6566634005bf174f63d880/website/client/components/hall/heroes.vue#L57-L61)) or from icons at the top of a player's profile ([loadAdminTools](https://github.com/HabitRPG/habitica/blob/9a3cdb5deb6af9054c6566634005bf174f63d880/website/client/components/userMenu/profile.vue#L16-L31)).
- To find all relevant code for the existing mute feature, grep the code base for `chatRevoked`. You won't need to change that code but copying some of it will help you create the new shadow mute feature.

---

The new "shadow mute" feature will work in a similar way to the existing mute feature but the player will not see any obvious sign that it's been applied to them.
- [x] It will require a new user attribute `flags.chatShadowMuted` which by default will have no value (as for `flags.chatRevoked`).
- [x] If a player has that flag set to `true`, they will be shadow-muted. When the flag is set to false or does not exist, all shadow muting effects will cease to apply to the player. The effects are described below.
- [ ] The Hall of Contributors and the player profile screen will need to have controls for turning the flag on and off. Copy what's done for `chatRevoked` and use different icons / text. Those controls will be visible to only mods and staff so their appearance won't need any significant design decisions; the contributor that works on this feature can do something simple initially and if we think it's not quite clear enough we'll give feedback about tweaking it.

When a player has `flags.chatShadowMuted` set to `true`, then:
- [x] They CAN send Private Messages to any player. (We might change this later but initially I think it's the best and simplest approach, especially since we want shadow muting to be unnoticable.)
- [x] They CAN post to their party or any private guild they are a member of. Those posts will be seen by party/guild members as normal - no special treatment will be applied to those posts.
- [x] They CAN post to the Tavern and public guilds BUT those posts are automatically hidden at the moment they are created. The posts will still be visible to the player themself and to mods and staff, but to no one else. This can be achieved by setting the post's `flagCount` to greater than 1 (there is already code to hide those messages).
- [ ] When a shadow-muted player posts to the Tavern or a public guild, moderators need to be notified in two ways:
   - [x] an email should be sent to moderators (in the same way that an email is sent when a normal post is flagged)
   - [x] a Slack message should be sent to the moderator Slack "flagged-posts" channel (in the same way that a Slack message is sent when a normal post is flagged)
   - The code for notifying moderators when a normal post is flagged is at [groupChatReporter.js](https://github.com/HabitRPG/habitica/blob/9a3cdb5deb6af9054c6566634005bf174f63d880/website/server/libs/chatReporting/groupChatReporter.js#L52-L72)
   - [x] Something similar can be done for these shadow-muted posts but the email and Slack message must make it clear that the posts have been automatically hidden because the player is shadow-muted.

Regarding setting the post's `flagCount`:
This is the code that hides a post from everyone but mods and the player who posted it when the flagCount is 2:
https://github.com/HabitRPG/habitica/blob/9a3cdb5deb6af9054c6566634005bf174f63d880/website/server/models/group.js#L337
- [x] As you see, currently the value "2" is hard-coded there. It should be converted to a constant and then the constant can be used both there and in the new shadow-muting code

----

_For anyone curious about how this shadow muting feature will be used, moderators are expecting to use it in only very rare, extreme cases. For example, if we're trying to control a persistent troll who is making new accounts to avoid a normal mute, we would shadow-mute them to decrease their impact on the community until staff have had time to set up an IP address ban._

_The primary reason we want this feature is because its code will become the basis of the feature described in https://github.com/HabitRPG/habitica/issues/8063_

_The current `flags.chatRevoked` mute feature will continue to be used in almost all cases where muting is appropriate._

Alys:
fixes https://github.com/HabitRPG/habitica/issues/10851

This PR allows moderators to mark a player as being shadow-muted, which will automatically hide public posts from that player. 

In detail:

- Mods can use the Hall to turn shadow-muting on and off:
image

- Mods can use icons in a player's profile to turn shadow-muting on and off. This PR also converts the profile admin tools from icons to links, and allows the crown icon to be used as a toggle to show or hide the links. These changes are only visible to moderators and staff. Regular users will not see anything different in the profile so normal visual design considerations aren't needed for this.

   - admin tools toggled off:
image
   - admin tools toggled on for a player who has not been muted or banned:
image
   - admin tools toggled on for a player who has been banned and had both types of muting applied (this also shows the on-hover effect for one of the links):
image

When a player has been shadow-muted, the only function that's affected is posting to the Tavern and public guilds. All other functions in Habitica work normally. Any public or private posts made before the player was shadow-muted are not affected.

When a shadow-muted player posts to the Tavern or a public guild:
- Their post is automatically flagged to be hidden immediately.
- Only mods/staff and the player themself can see the post.
- The player sees no indication that their post is hidden.
- Mods see a message on the post (screenshot below).
- Mods can un-hide the post after they've reviewed it (in the same way they can un-hide manually-flagged posts).
- Mods receive a notification in Slack that a shadow-muted post has been made.
- Mods receive an email for the shadow-muted post. _(I still need to create the Mandrill template for this - I'll do that after/if this PR is approved.)_
image

The first three commits modify existing tests so you might want to review them separately:
1. 4127ac5b36bcece03335075879faca2b51601cb5 - moves two existing tests into the 'mute user' describe block, where there's already one test that they are related to
2. 9e0b663706356d9c71c13c06ca88204e41043b2d - changes those two existing tests to have names consistent with the other one in the mute block and to use const instead of let
3. a6f0ded2bc3923b8fff0c1c9c2c0536abca0f4f8 - improves methods for restoring chat permissions to test users (handles an old TODO comment)

The fourth commit (acdd3d88ad0f42c55ebeb9bf0a92ac53d379a9ca) adds new tests for shadow-muting. It also defines three new constants, all related to flagging / hiding messages:
```
export const CHAT_FLAG_LIMIT_FOR_HIDING = 2; // hide posts that have this many flags
export const CHAT_FLAG_FROM_MOD = 5; // a flag from a moderator counts as this many flags
export const CHAT_FLAG_FROM_SHADOW_MUTE = 10; // a shadow-muted user's post starts with this many flags
// @TODO use those constants to replace hard-coded numbers
```
I've used those constants in the code that I added and modified in this PR. but there's still other pieces of code that don't use them yet (code that's not directly related to this change). I'll adjust the remaining code to use the constants everywhere after/if this PR is accepted.

The remaining commits implement the shadow-muting feature itself.

When/if you're happy with the code, can it please be merged to delta so that I can test the email and slack notification code, and let the other mods test it?

Things that I need to do after/if this PR is accepted:
- [x] Create a Mandrill template `shadow-muted-post-report-to-mods`, similar to the `slur-report-to-mods template`.
- [x] Test the Slack message from delta or staging. It must clearly indicate that the post has been hidden automatically.
- [ ] Ask the other mods that they find the Slack notification to be clear enough.
- [x] Test the notification email from delta or staging. It must indicate that the post has been hidden automatically.
- [ ] Create an issue to remind myself to replace hard-coded numbers with the new constants (CHAT_FLAG_LIMIT_FOR_HIDING, CHAT_FLAG_FROM_MOD, CHAT_FLAG_FROM_SHADOW_MUTE)

---

_For anyone curious about how this shadow muting feature will be used, moderators are expecting to use it in only very rare, extreme cases. For example, if we're trying to control a persistent troll who is making new accounts to avoid a normal mute, we would shadow-mute them to decrease their impact on the community until staff have had time to set up an IP address ban._

_The primary reason we want this feature is because its code will become the basis of the feature described in https://github.com/HabitRPG/habitica/issues/8063_

_The current `flags.chatRevoked` mute feature will continue to be used in almost all cases where muting is appropriate._


website/client/components/chat/chatCard.vue
```diff
@@ -1,8 +1,9 @@
 <template lang="pug">
 div
   .mentioned-icon(v-if='isUserMentioned')
-  .message-hidden(v-if='!inbox && msg.flagCount === 1 && user.contributor.admin') Message flagged once, not hidden
-  .message-hidden(v-if='!inbox && msg.flagCount > 1 && user.contributor.admin') Message hidden
+  .message-hidden(v-if='!inbox && user.contributor.admin && msg.flagCount === 1') Message flagged once, not hidden
```
paglias: these three `message-hidden ` might become a bit more readable if we could use only one and then display the string based on a computer prop that returns the string maybe?

website/server/controllers/api-v3/chat.js
```diff
@@ -186,7 +187,43 @@ api.postChat = {
...
+        {name: 'AUTHOR_EMAIL', content: authorEmail},
+        {name: 'AUTHOR_MODAL_URL', content: `/profile/${user._id}`},
+
+        {name: 'GROUP_NAME', content: group.name},
+        {name: 'GROUP_TYPE', content: group.type},
+        {name: 'GROUP_ID', content: group._id},
+        {name: 'GROUP_URL', content: groupUrl},
+      ];
+
+      sendTxn(FLAG_REPORT_EMAILS, 'shadow-muted-post-report-to-mods', report);
```
paglias, line 215: I'll have to check but we might have to add the template before the PR goes live or it might throw an error on the email server

paglias:
@Alys you want this go on delta for some this?

paglias:
Actually nevermind, I see that you asked for this to be put on delta. The code looks good but I think we should first add the template to mandrill

Alys:
Sure, no worries! I just didn't want to create the mandrill template before the PR had been reviewed in case any of the parameters needed to be changed. :) I'll post here when I've created it and made the code improvement you suggested.

website/client/components/chat/chatCard.vue
Alys: I think I've done that how you wanted it but if it's not exactly what you had in mind, tell me and I'll change it. https://github.com/HabitRPG/habitica/pull/11239/commits/180ae50fb74444a70a8e0ff639890ee280c5e182
You were right - it's nicer now!

The new code has a check for msg.flagCount being non-zero in two places (in the `.message-hidden(v-if` line and in `flagCountDescription()`) which is redundant but I think it's clearer this way. However I'm happy to remove the check from either location if you want.

I still need to create the mandrill template. I'll post again when that's done. Feel free to ignore this until then if you want. :)

paglias (commented):
lgtm!

Alys:
I've created the Mandrill template `shadow-muted-post-report-to-mods`.

This is ready to go to delta now, but there's no need to rush - just when you have time and it's convenient.

paglias:
pushed to delta!

Alys:
The slack notification and email is working well from delta (paglias, thanks for deploying!)

The other mods are testing now.

I've realised that when a post is automatically hidden because of shadow-muting, the new message notification should not be sent to all guild members because there's nothing new they can see. I've added a commit for that. In the test file, I've added a new `context` block around existing tests so you'll probably want to review it with whitespace differences ignored:
https://github.com/HabitRPG/habitica/pull/11239/commits/64a92669087632104d786e2ecb967195e4b4e7e7

Alys:
This has worked well for me in delta and none of the other mods have made requests for changes so this is ready to be merged if you're happy with the code.

It doesn't need to stay on delta any longer, so if you need to recreate delta again before this is merged, you don't have to add this PR to it again.

PR #11250, opened 2019-06-30
Only includes backticks if they haven't been already included. Fixes …

### LINKED ISSUE #11202: quest system messages sometimes have double backticks (rage attack, etc)
As reported by @Chandramani_Adil (58de0b72-3af8-4f2a-b9a1-f7d196b29757), some quest system messages are incorrectly formatted because they have nested backticks:

index

It happens when the messages already contain backticks in the translatable strings. The code that displays the messages always inserts its own backticks around the translatable strings. It was reported for a boss_rage message but other types of system messages are also affected. I've included details below.

Perhaps the easiest fix for this would be for the code that displays the messages to first check if there's at least one backtick already in the message and then not display it. ~`I'll leave this issue as suggestion-discussion for a few days for the staff to comment if they wish, and then if there's no objections I'll mark it as help wanted for that fix.`~

This is the code that inserts the backticks:
https://github.com/HabitRPG/habitica/blob/e4b2ef65993f03c1acfc88d8923363ebf917b028/website/server/libs/chat/group-chat.js#L36-L106

For example, for the exact instance shown in the screenshot above:
https://github.com/HabitRPG/habitica/blob/e4b2ef65993f03c1acfc88d8923363ebf917b028/website/server/libs/chat/group-chat.js#L50
(For those new to JavaScript, the outer backticks on that line are just for quoting the string in a way that allows variable interpolation; the inner backticks which are escaped by backslashes are the ones that are printed as part of the string.)

Some or all of the relevant strings are in `website/common/locales/en/questsContent.json` and this is the one shown in the screenshot:
https://github.com/HabitRPG/habitica/blob/e4b2ef65993f03c1acfc88d8923363ebf917b028/website/common/locales/en/questsContent.json#L399

This issue can't be fixed by removing the backticks from the translatable strings because for some examples, only part of the string is formatted with backticks:
https://github.com/HabitRPG/habitica/blob/e4b2ef65993f03c1acfc88d8923363ebf917b028/website/common/locales/en/questsContent.json#L436

In any case where the translatable string includes one or more backticks, the code that displays that string should not insert its own backticks.

In case it helps, this is the relevant data for the quest "system" message stored in the `chats` collection in the database:
```
{
    "_id": "1cb8587f-740e-46e4-ae8f-ef07cca1c05b",
    "text": "`Magical Axolotl uses AXOLOTL REGENERATION!`\n\n`A curtain of colorful bubbles obscures the monster for a moment, and when it clears, some of its wounds have vanished!`",
    "info": {
        "type": "boss_rage",
        "quest": "axolotl"
    },
    "timestamp": "2019-06-01T04:11:34.396Z",
    "uuid": "system",
},
```

Scens:
…the problem.

Fixes #11202 
Screenshot 2019-06-29 at 19 04 38

### Changes
Affects only messages created by translateMessage function in /website/server/libs/chat/group-chat.js. There doesn't seem to be any other case of displaying double backticks.

Code only includes backticks if they haven't been already included (because they might have been hardcoded in questContent.json).

----
UUID: dd59e7c1-b76a-466e-b301-783b852da699

### DIFF
--- a/website/server/libs/chat/group-chat.js
+++ b/website/server/libs/chat/group-chat.js
@@ -35,23 +35,23 @@ export function translateMessage (lang, info) {
 
   switch (info.type) {
     case 'quest_start':
-      msg = `\`${shared.i18n.t('chatQuestStarted', {questName: questScrolls[info.quest].text(lang)}, lang)}\``;
+      msg = `${shared.i18n.t('chatQuestStarted', {questName: questScrolls[info.quest].text(lang)}, lang)}`;
       break;
 
     case 'boss_damage':
-      msg = `\`${shared.i18n.t('chatBossDamage', {username: info.user, bossName: questScrolls[info.quest].boss.name(lang), userDamage: info.userDamage, bossDamage: info.bossDamage}, lang)}\``;
+      msg = `${shared.i18n.t('chatBossDamage', {username: info.user, bossName: questScrolls[info.quest].boss.name(lang), userDamage: info.userDamage, bossDamage: info.bossDamage}, lang)}`;
       break;
 
     case 'boss_dont_attack':
-      msg = `\`${shared.i18n.t('chatBossDontAttack', {username: info.user, bossName: questScrolls[info.quest].boss.name(lang), userDamage: info.userDamage}, lang)}\``;
+      msg = `${shared.i18n.t('chatBossDontAttack', {username: info.user, bossName: questScrolls[info.quest].boss.name(lang), userDamage: info.userDamage}, lang)}`;
       break;
 
     case 'boss_rage':
-      msg = `\`${questScrolls[info.quest].boss.rage.effect(lang)}\``;
+      msg = `${questScrolls[info.quest].boss.rage.effect(lang)}`;
       break;
 
     case 'boss_defeated':
-      msg = `\`${shared.i18n.t('chatBossDefeated', {bossName: questScrolls[info.quest].boss.name(lang)}, lang)}\``;
+      msg = `${shared.i18n.t('chatBossDefeated', {bossName: questScrolls[info.quest].boss.name(lang)}, lang)}`;
       break;
 
     case 'user_found_items':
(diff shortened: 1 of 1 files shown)


paglias:
Hi @Scens  and thanks for the PR! Sorry for the delay here but we'll review the pr soon

website/server/libs/chat/group-chat.js
```diff
@@ -35,73 +35,76 @@ export function translateMessage (lang, info) {
...
-      msg = `\`${quests[info.quest].boss.desperation.text(lang)}\``;
+      msg = `${quests[info.quest].boss.desperation.text(lang)}`;
       break;
 
     case 'claim_task':
       msg = `${shared.i18n.t('userIsClamingTask', {username: info.user, task: info.task}, lang)}`;
       break;
   }
 
+  if (!msg.includes('\`')) {
```
paglias: is the backslash here necessary?

website/server/libs/chat/group-chat.js
```diff
@@ -35,73 +35,76 @@ export function translateMessage (lang, info) {
 
   switch (info.type) {
     case 'quest_start':
-      msg = `\`${shared.i18n.t('chatQuestStarted', {questName: questScrolls[info.quest].text(lang)}, lang)}\``;
+      msg = `${shared.i18n.t('chatQuestStarted', {questName: questScrolls[info.quest].text(lang)}, lang)}`;
```
paglias: this looks good but here like in other messages below given that the string is only the result of the i18n.t call it's not necessary to use the backslashes, so

> msg = `${shared.i18n.t('chatQuestStarted', {questName: questScrolls[info.quest].text(lang)}, lang)}`;

can be

> msg = shared.i18n.t('chatQuestStarted', {questName: questScrolls[info.quest].text(lang)}, lang);

paglias (changes requested):
left two small comments!

Scens:
> left two small comments!

Thanks! You are right, I have taken care of these. I've tested this again and everything works fine.

SabreCat:
Thanks, @Scens! This will go live on the production site around this time next week. Credited toward your second Blacksmith tier!

PR #11251, opened 2019-06-30
Ask for confirmation when buying quests

### LINKED ISSUE #11244: purchasing quest scrolls / bundles should have confirmation popup
Quest scroll purchases don't have a confirmation step.

When you buy other gem-purchasable items such as hatching potions, you see a confirmation dialog that tells you how many gems the total purchase will be and asks you to confirm:
image

The same thing should happen when you buy quest scrolls or quest scroll bundles.

domtriola:
Fixes #11244 

When a user clicks "Buy Now" for an item with currency 'gems' or 'hourglasses', a confirmation dialog prompts the user to confirm their purchase. This should also happen for quests.

Some quests cost gold, but the default behavior is to not ask for confirmation for gold purchases. As suggested by @Alys I have made it so that quests that cost gold-purchasable quests trigger a confirmation modal as well.

### Changes
* Create `confirmPurchase` mixin method for the `buy` modals
* Add `purchaseForGold` key to `website/common/locales/en/character.json`
* Use the `confirmPurchase` method in the `buyQuestModal` to prompt user for confirmation on "Buy Now" click
* Refactor `buyModal` to also use `confirmPurchase` method

### Testing Steps

* Confirm quest buy modals trigger confirmation dialogs for:
  * Quest bundles
  * Quests that cost gold
  * Quests that cost gems
quest-confirmations

* Confirm other items trigger confirmation dialogs (with the exception of gold-purchasable items):
  * Items that cost gold (doesn't trigger dialog)
  * Items that cost gems
  * Items that cost hourglasses
item-confirmations

----
UUID: f19ed7b1-366c-4999-b77f-9bd48cc88eaf

### DIFF
--- a/website/client/components/shops/buyModal.vue
+++ b/website/client/components/shops/buyModal.vue
@@ -402,13 +402,8 @@
           return;
         }
 
-        if (this.item.currency === 'gems' &&
-          !confirm(this.$t('purchaseFor', { cost: this.item.value * this.selectedAmountToBuy }))) {
-          return;
-        }
-
-        if (this.item.currency === 'hourglasses' &&
-          !confirm(this.$t('purchaseForHourglasses', { cost: this.item.value }))) {
+        const shouldConfirmPurchase = this.item.currency === 'gems' || this.item.currency === 'hourglasses';
+        if (shouldConfirmPurchase && !this.confirmPurchase(this.item.currency, this.item.value * this.selectedAmountToBuy)) {
           return;
         }
 
--- a/website/client/components/shops/quests/buyQuestModal.vue
+++ b/website/client/components/shops/quests/buyQuestModal.vue
@@ -265,6 +265,11 @@
         this.$emit('change', $event);
       },
       buyItem () {
+        if (this.item.currency === 'gems') {
+          if (!confirm(this.$t('purchaseFor', { cost: this.item.value * this.selectedAmountToBuy }))) {
+            return;
+          }
+        }
         this.makeGenericPurchase(this.item, 'buyQuestModal', this.selectedAmountToBuy);
         this.purchased(this.item.text);
         this.hideDialog();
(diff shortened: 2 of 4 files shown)


website/client/components/shops/buyModal.vue
```diff
@@ -402,13 +402,8 @@
...
 
-        if (this.item.currency === 'gems' &&
-          !confirm(this.$t('purchaseFor', { cost: this.item.value * this.selectedAmountToBuy }))) {
-          return;
-        }
-
-        if (this.item.currency === 'hourglasses' &&
-          !confirm(this.$t('purchaseForHourglasses', { cost: this.item.value }))) {
+        const shouldConfirmPurchase = this.item.currency === 'gems' || this.item.currency === 'hourglasses';
+        if (shouldConfirmPurchase && !this.confirmPurchase(this.item.currency, this.item.value * this.selectedAmountToBuy)) {
```
domtriola, line 406: If it is desirable to ask for confirmation for gold-purchasable items as well(?) I could update this line to be:
```suggestion
        if (!this.confirmPurchase(this.item.currency, this.item.value * this.selectedAmountToBuy)) {
```

Alys: For most gold-purchasable items, I'd say that there should not be a confirmation step because it's a bit annoying and because losing small amounts of gold accidentally isn't a big issue. Avoiding the annoyance of the extra step is more important.

Gold-purchasable quest scrolls are an exception because they cost a LOT of gold (so accidental purchases can be very upsetting to users who don't know about [FCV](http://habitica.fandom.com/wiki/Fix_Character_Values)), and they are bought rarely so the annoyance of the confirmation step isn't a big factor.

domtriola: Makes sense to me. I'll leave the proposed change as-is then.

paglias:
Hi @domtriola  and thanks for the PR! Sorry for the delay here but we'll review the pr soon

paglias:
@domtriola  thanks for this PR, it worked perfectly

I'm merging the PR and awarding your contributor tier.

PR #11252, opened 2019-07-01
Add audio theme demo button

domtriola:
Trello feature request: https://trello.com/c/tljx8gKC/18-music-sound-effects

> Gustavo Silveira Jul 28, 2018 at 8:58 AM
> Hey, is there already a way to preview an audio theme? Today I was actually struggling with this, as to check an audio theme I had to change it, them go back to my tasks and tick something, change the audio, tick a task, etc... If theres no way of previewing a theme now, I would like to suggest this feature

I'm not sure if a different UX/UI would be desired for this feature (or if the feature is desired to release at all), but here is a potential solution. I found that exploring the different audio options in the current app is too tedious to bother with, and when I looked through the sound effects Trello card I noticed that I wasn't the only one that wished there was a sound demo feature.

### Changes
* Adds a "Demo" button to the audio theme selection setting. 
  * When the button is clicked a sound sample from that theme will be played. 
  * Each additional click plays the next sound in that theme's list, cycling back to the first sound when the end is reached. 
  * Changing themes resets the sound index to 0.

**Old UI**
Screen Shot 2019-06-30 at 4 29 25 PM

**New UI**
Screen Shot 2019-06-30 at 3 58 49 PM

**Demo with log statements**:
sound-demo

----
UUID: f19ed7b1-366c-4999-b77f-9bd48cc88eaf

### DIFF
--- a/website/client/components/settings/site.vue
+++ b/website/client/components/settings/site.vue
@@ -25,10 +25,12 @@
       hr
 
       .form-horizontal
-        h5 {{ $t('audioTheme') }}
-        select.form-control(v-model='user.preferences.sound',
-          @change='set("sound")')
-          option(v-for='sound in availableAudioThemes', :value='sound') {{ $t(`audioTheme_${sound}`) }}
+        .form-group
+          h5 {{ $t('audioTheme') }}
+          select.form-control(v-model='user.preferences.sound',
+            @change='changeAudioTheme')
+            option(v-for='sound in availableAudioThemes', :value='sound') {{ $t(`audioTheme_${sound}`) }}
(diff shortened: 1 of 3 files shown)


paglias:
Hi @domtriola  and thanks for the PR! Sorry for the delay here but we'll review the pr soon

paglias:
Sorry for the delay here but we're discussing internally about this PR

paglias:
@domtriola the code looks good and works as intended! we're testing this on an internal version to make sure the UX feels right

SabreCat:
Thanks, @domtriola! You should see this go live on the production site around this time next week. Credited toward your second Blacksmith tier!

PR #11258, opened 2019-07-08
Performance/inbox paging 

negue:
All changes based on PR #11157 / #11226

### DIFF
--- a/website/client/components/chat/chatCard.vue
+++ b/website/client/components/chat/chatCard.vue
@@ -273,5 +273,8 @@ export default {
       return habiticaMarkdown.render(String(text));
     },
   },
+  mounted () {
+    this.$emit('item-mounted', this.msg.id);
+  },
 };
 </script>
--- a/website/client/components/chat/chatCard.vue
+++ b/website/client/components/chat/chatCard.vue
@@ -274,7 +274,7 @@ export default {
     },
   },
   mounted () {
-    this.$emit('item-mounted', this.msg.id);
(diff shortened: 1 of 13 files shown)


paglias:
Looks like there are some lint issues

website/client/components/chat/chatCard.vue
```diff
@@ -273,5 +273,8 @@ export default {
       return habiticaMarkdown.render(String(text));
     },
   },
+  mounted () {
+    this.$emit('item-mounted', this.msg.id);
```
paglias: $emit is broadcasted to all components upstream? In that case a more detailed name might help to avoid confusion like `message-component-mounted`

website/server/controllers/api-v4/inbox.js
```diff
@@ -93,4 +93,52 @@ api.conversations = {
   },
 };
 
+function mapMessage (newChat, user) {
```
paglias: can this be moved to the message model?

paglias (commented):
couple of comments but this looks good (I have not tested it)

website/client/components/chat/chatCard.vue
negue: not to all components, just to the ones that use this event inside the template

website/server/controllers/api-v4/inbox.js
negue: I tried the `schema.methods`-way but none the properties were changed, so I just moved the method to the same file but outside of the `schema.methods`

Issue #11264, opened 2019-07-13
display `null` challenge prize as 0 gems


Alys:
As described at https://github.com/HabitRPG/habitica/issues/9745 it's possible for the gem prize for a challenge to be null which results in the prize box on the challenge having a blank space where the number of gems should be, as shown in the screenshot below.

As I mentioned in [one of my comments there](https://github.com/HabitRPG/habitica/issues/9745#issuecomment-508986367) a `null` prize doesn't seem to cause any problems apart from that display glitch, so this issue is to have **the website's front-end code modified so that when the prize is `null`, the number `0` is shown in the prize box**.

![screenshot_2017-12-26_19-56-52](https://user-images.githubusercontent.com/1495809/34363766-eb37fa24-ea76-11e7-8302-40a88e5bfea3.png)

_(https://github.com/HabitRPG/habitica/issues/9745 is being left open for a fix to the back-end code and a migration to correct all the affected challenges.)_

kmacart:
I'll take the issue =)

kmacart:
I've fixed it I think! Took me hours =|

sanjayn2096:
Is this issue resolved? can I work on it?

dl1683:
Is this open?

Alys:
@kmacart has been looking into this so I'm marking it as in progress for them.

@kmacart Would you like to submit a pull request for the fix? So far it's just in a commit in your clone of the repo and it's easier for us to review and comment on it if there's a PR. :)

akmaekki:
Hi, is this issue still open? I would like to work on it.

paglias:
@akmaekki for now someone is still working on it but if we don't get an answer in a couple of weeks it'll be made available again

akmaekki:
Ok, thank you for your answer. Just one more question. Is there another
possibility to contribute? Like reviewing pull requests, or something code
related?

paglias:
@akmaekki reviewing pull requests is done by the staff usually but there are a lot of issues to work on if you want to ! Just pick any that doesn't have an `in progress` or `on hold` ticket and comment that you want to work on it

evelynstender:
Hi! Is this still open? I'd like to work on this!

paglias:
The server side change is done but missing the migration, the client side one is still missing, see https://github.com/HabitRPG/habitica/pull/11267#issuecomment-511350354

PR #11275, opened 2019-07-19
Front-end change for the null gem option.

kmacart:
Changes it to be a 0 when null in the challenge page only. Still working on the other pages.

[//]: #11264 Front-end fix for null gem prize
### Changes
[//]: # Changed lines 33 and 34 to be challenge.prize(v-if=null) {{challenge.prize = 0}}
          .details(v-once) {{$t('prize')}}

[//]: # 0fb3db86-7009-46d6-9eea-485a8f7e6734

----
UUID:

### DIFF
--- a/website/client/components/challenges/challengeDetail.vue
+++ b/website/client/components/challenges/challengeDetail.vue
@@ -30,7 +30,7 @@
           .details(v-once) {{$t('participantsTitle')}}
         .box
           .svg-icon.gem-icon(v-html="icons.gemIcon")
-          | {{challenge.prize}}
+          challenge.prize(v-if=null) {{challenge.prize = 0}}
           .details(v-once) {{$t('prize')}}
     .row.challenge-actions
       .col-12.col-md-6


website/client/components/challenges/challengeDetail.vue
```diff
@@ -30,7 +30,7 @@
           .details(v-once) {{$t('participantsTitle')}}
         .box
           .svg-icon.gem-icon(v-html="icons.gemIcon")
-          | {{challenge.prize}}
+          challenge.prize(v-if=null) {{challenge.prize = 0}}
```
paglias, line 33: Is this necessary with the other PR + migration going to make it impossible to have challenges with a prize of 0? I'd only add a new check when creating the challenge

kmacart: What do you mean? The reason I did this was because that way the currently null challenge prizes that haven't been changed would show up as 0 gem prize - do you mind elaborating a bit? Thanks heaps! =)

Kira

kmacart: What do you mean? The reason I did this was because that way the currently null challenge prizes that haven't been changed would show up as 0 gem prize - do you mind elaborating a bit? Thanks heaps! =)

Kira

paglias: Sure! The reason why I don't think this is necessary is that when the other PR will be merged we'll also run a migration to convert all existing null prize challenges to having a prize of 0. 

Also, there's a problem with this code: in vue templates there should only be view logic (for example show a 0 if challenge.prize is null) while in this case it would be modifying the value of challenge.prize. Also the syntax of the v-if is wrong because it checks if the value passed to it is converted to true or false and null is always false.

It should be used like this at the existing like 33

```
          | {{challenge.prize || 0}}
```

which means should challenge.prize or 0 if the value is absent

let me know if anything is not clear :)

paglias:
Hi @kmacart ! do you have any update on the status of this PR? thanks!

paglias:
Hi @kmacart , I'm going to close this one but if you want to work again on this just leave a comment

PR #11293, opened 2019-07-31
More sophisticated migration tracking

SabreCat:

### Changes

1. Adds a `migrations` field to the user, deprecating `migration`.
2. Changes the routine `take-this.js` migration to use the new field, as an example:
* Trims existing data to a maximum of 5 entries in date order
* Uses the existence of a `migrations.name` value of the migration's name to check if the migration has already run on that user, instead of the single value of the `migration` field
* Saves information about the migration to the `migrations` field, in the form

```
migrations: [
  {
    name: '20190801_take_this',
    processed: new Date(),
    item: 'shield_special_takeThis', // arbitrary field name & data for troubleshooting
  },
  {
    name: '20190909_fix_things',
    processed: new Date(),
    fixed: 'party._id guilds',
  },
]
```


migrations/users/take-this.js
```diff
@@ -10,33 +11,50 @@ let count = 0;
 async function updateUser (user) {
   count++;
 
-  const set = {};
-  let push;
+  let set;
+
+  let migrationsArray = sortBy(user.migrations, ['processed']);
```
paglias: is the sortBy necessary with an array now? it's already sorted with the last one added at the end

website/server/models/user/schema.js
```diff
@@ -73,7 +73,8 @@ let schema = new Schema({
   // We want to know *every* time an object updates. Mongoose uses __v to designate when an object contains arrays which
   // have been updated (http://goo.gl/gQLz41), but we want *every* update
   _v: { $type: Number, default: 0 },
-  migration: String,
+  migration: String, // deprecated
+  migrations: Array, // {name: '20190801_take_this', processed: new Date(), item: 'shield_special_takeThis'}
```
paglias: can we specify this is an array of mixed objects? or if it always follow the `{name: '20190801_take_this', processed: new Date(), item: 'shield_special_takeThis'}` structure maybe define it?

paglias (commented):
left a couple of comments but otherwise looks good

paglias:
is this something we still want?

SabreCat:
Ah yeah, might not be a bad idea! I can still clean this up if desired.

SabreCat:
OK, how's that look?

paglias (commented):
LGTM, I wonder if we could abstract a bit the pieces of code that are common to all migrations, for example in the migration runner script, to avoid having to remember to copy and paste code from an old migration.

paglias:
@SabreCat I'm leaving this to you if you want to merge it

SabreCat:
Going to close this for now--we haven't really had a need for this during the last several migrations we've run, so maybe it would complicate the schema for no noticeable benefit. We can always reopen this if it turns out we need it.

PR #11294, opened 2019-07-31
Post-outage groups cleanup migration

SabreCat:

### Changes

Adds a migration to patch up a few lingering data discrepancies from the Guilds/Parties outage of a couple weeks back:

* Recalculates `memberCount` for all groups
* Realigns party members' individual `party.quest` data with their statuses in `group.quest.members`
* Outputs (without making any changes) an alert if a group has an assigned leader that's not actually a member of that group

### DIFF
--- a/migrations/archive/2019/20190731_outage_repair_groups.js
+++ b/migrations/archive/2019/20190731_outage_repair_groups.js
@@ -1,0 +1,74 @@
+/* eslint-disable no-console */
+const MIGRATION_NAME = '20190731_outage_repair_groups';
+import { model as Group } from '../../website/server/models/group';
+import { model as User } from '../../website/server/models/user';
+import forEach from 'lodash/forEach';
+
+const progressCount = 1000;
+let count = 0;
+
+async function updateGroup (group) {
+  count++;
+
+  const set = {};
+
+  if (group.type === 'guild') {
+    set.memberCount = await User.countDocuments({guilds: group._id});
+  } else {
+    set.memberCount = await User.countDocuments({party._id: group._id});
+  }
+
+  if (group.quest && group.quest.key && group.quest.members) {
+    forEach(group.quest.members, async (status, memberId) => {
+      if (status === null) await User.update({_id: memberId, {$set: {
+        'party.quest.RSVPNeeded': true,
+        'party.quest.key': group.quest.key,
+      }}});
+      if (status === true) await User.update({_id: memberId, {$set: {
+        'party.quest.key': group.quest.key,
+      }}})
+    });
+  }
+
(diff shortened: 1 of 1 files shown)


migrations/archive/2019/20190731_outage_repair_groups.js
```diff
@@ -0,0 +1,74 @@
...
+
+  if (group.quest && group.quest.key && group.quest.members) {
+    forEach(group.quest.members, async (status, memberId) => {
+      if (status === null) await User.update({_id: memberId, {$set: {
+        'party.quest.RSVPNeeded': true,
+        'party.quest.key': group.quest.key,
+      }}});
+      if (status === true) await User.update({_id: memberId, {$set: {
+        'party.quest.key': group.quest.key,
+      }}})
```
Alys: It might be worth setting `party.quest.RSVPNeeded` to false for both `(status === true)` and `(status === false)`. We see a trickle of people needing RSVPNeeded to be turned off (not necessarily from the recent group troubles, just as a general thing; usually it's because they're no longer in a party but not always).

SabreCat: Updated!

migrations/archive/2019/20190731_outage_repair_groups.js
```diff
@@ -0,0 +1,82 @@
...
+  const set = {};
+
+  if (group.type === 'guild') {
+    set.memberCount = await User.countDocuments({guilds: group._id});
+  } else {
+    set.memberCount = await User.countDocuments({party._id: group._id});
+  }
+
+  if (group.quest && group.quest.key && group.quest.members) {
+    forEach(group.quest.members, async (status, memberId) => {
```
paglias, line 22: lodash's for each doesn't work as expected with promises or async. It'll go ahead without waiting. Since this is an object you can use a normal `for...of` where you can use await but it'll process everything in sequence not parallel. Otherwise you can convert `group.quest.members ` to an array and use .map + Promise.all

paglias (changes requested):
There's a small problem with the way quests members are processed but otherwise looks good

migrations/archive/2019/20190731_outage_repair_groups.js
SabreCat: argh, I keep making that mistake. >.< Someday I'll learn!

Alys:
Did this script end up working? If so, would it be okay if we did keep a copy of it in the repo so that we can more easily make a cut-down version just for fixing member counts? That would be handy to have since incorrect counts are an ongoing problem.

PR #11296, opened 2019-08-05
[WIP] Add API Call to bulk score tasks

phillipthelen:
Not quite ready for merging!

This PR adds a new API call `/api/v4/tasks/bulk-score`. This allows the user to pass data in the following format:

```
[
  {
    id: "123",
    direction: "up"
  }, {
    id: "456",
    direction: "down"
  }
]
```

This scores all tasks that are passed in the correct order and also allows tasks to be passed multiple times, to tscore them more than once.
The code unifies most of the scoring code between both the normal call and the new bulk call. The return of the old API call is (as far as I can tell) unchanged.

I'm not super happy with the return data for the old call, because it currently can't map everything that might have happened during the call.

```
{
  "data": {
    "deltas": [
      {
        "id": "f437911f-56e0-45b5-a42d-3d41c3a6ba57",
        "delta": 0.5886769031421589
      },
      {
        "id": "275a2057-809a-44f1-a04b-fd27a45b838b",
        "delta": -1.573889391038481
      }
    ],
    "_tmp": {
      "quest": {
        "progressDelta": 0.295810143828934
      }
    },
    "buffs": {
      "str": 0,
      "int": 0,
      "per": 0,
      "con": 0,
      "stealth": 0,
      "streaks": false,
      "snowball": false,
      "spookySparkles": false,
      "shinySeed": false,
      "seafoam": false
    },
    "training": {
      "int": 0,
      "per": 0,
      "str": 0,
      "con": 0
    },
    "hp": 18.9,
    "mp": 13.625,
    "exp": 60,
    "gp": 22.089986247298715,
    "lvl": 3,
    "class": "warrior",
    "points": 3,
    "str": 0,
    "con": 0,
    "int": 0,
    "per": 0
  },
}
```

The big issue is the `_tmp` field, which currently contains mostly data from the last task scored. Some data, like the quest progress delta could just be added together, but for others like drops this isn't as easy. What do others feel would be the best approach here? Move the _tmp_ field to the individual task data (currently called `deltas`)?


website/common/script/ops/scoreTask.js
```diff
@@ -197,7 +197,7 @@ module.exports = function scoreTask (options = {}, req = {}) {
 
   // This is for setting one-time temporary flags, such as streakBonus or itemDropped. Useful for notifying
   // the API consumer, then cleared afterwards
-  user._tmp = {};
+  if (!user._tmp) user._tmp = {};
```
paglias: this way it'll never be cleared?

website/server/libs/taskManager.js
```diff
@@ -281,3 +291,172 @@ export function moveTask (order, taskId, to) {
...
+      logger.error(e);
+      reject();
+    }
+  }
+  resolve();
+}
+
+export async function scoreTasks (user, taskScorings, req, res) {
+  let taskIds = taskScorings.map(taskScoring => taskScoring.id).filter(id => id !== undefined);
+  if (taskIds.length === 0) throw new BadRequest();
```
paglias: this should have an error message specified

website/server/libs/taskManager.js
```diff
@@ -281,3 +291,172 @@ export function moveTask (order, taskId, to) {
...
+  let tasks = {};
+  (await Tasks.Task.findMultipleByIdOrAlias(taskIds, user._id)).forEach(task => {
+    tasks[task._id] = task;
+    if (task.alias !== undefined && task.alias.length > 0) {
```
paglias: I'd just check for `if (task.alias)` it covers all the falsey values including undefined and empty strings

website/server/libs/taskManager.js
```diff
@@ -281,3 +291,172 @@ export function moveTask (order, taskId, to) {
...
+  });
+  if (tasks.length === 0) throw new NotFound(res.t('taskNotFound'));
+  let scorePromises = [];
+  taskScorings.forEach(taskScoring => {
+    if (tasks[taskScoring.id] !== undefined) {
+      scorePromises.push(scoreTask(user, tasks[taskScoring.id], taskScoring.direction, req, res));
+    }
+  });
+  let returnDatas = await Promise.all(scorePromises);
+  if (returnDatas.length === 0) throw new BadRequest();
```
paglias: this is not necessary

website/server/libs/taskManager.js
```diff
@@ -281,3 +291,172 @@ export function moveTask (order, taskId, to) {
...
+  let savePromises = [returnDatas[returnDatas.length - 1].user.save()];
+  returnDatas.forEach(returnData => {
+    if (returnData.taskOrderPromise) savePromises.push(returnData.taskOrderPromise);
```
paglias: can we combine all the taskOrderPromise in one single update? We risk conflicts by moving multiple things in parallell in separate requests

website/server/controllers/api-v3/tasks.js
```diff
@@ -550,139 +547,16 @@ api.scoreTask = {
...
-
-    if (task.group && task.group.taskId) {
-      await handleSharedCompletion(task);
-    }
-
-    // Save results and handle request
-    if (taskOrderPromise) promises.push(taskOrderPromise);
-    let results = await Promise.all(promises);
+    const data = await scoreTasks(user, [{id: taskId, direction}], req, res);
+    user = data.user;
```
paglias: I don't think this is necessary, if the context of the user object isn't lost (and it shouldn't) at this point the previous `user` value will have the updated data

paglias (commented):
left some comments. I'm a bit worried about all the save requests we do to the database, this risks being very slow

website/common/script/ops/scoreTask.js
phillipthelen: Is this field even saved to the database? From my testing it looked like it wasn't saved.

website/server/libs/taskManager.js
phillipthelen: Do you have a suggestion for how to handle that? since it's only pushing and pulling todos, I could have the `returnData` contain if it should be pushed/pulled and then iterate over that to create one user update operation? Does that sound good?

website/common/script/ops/scoreTask.js
paglias: it's not saved but what happens if when bulk scoring 2 tasks both return a drop? only the first one will be added to user._tmp, the other will be ignored

website/server/libs/taskManager.js
paglias: Hmm if we want to keep things simple I would probably try to simplify the entire thing by simply creating one API endpoint that accepts multiple requests' data and for each one simulate the scoring of a single task, save the result and then all the results are returned to the client in an array

phillipthelen: That would make the number of save requests even higher than this version. Since then it would be 2-3 x numberOfTasks (depending on if it's a todo or not and the task has to be moved). Right now my PR saves 1-2 x numberOfTasks + 1. With my proposed change it should be numberOfTasks + 2. (one for the user and one for the task moving)

website/common/script/ops/scoreTask.js
phillipthelen: Yes. That is the part I meant I'm not yet happy with. I'm still thinking about how to solve this while still maintaining backwards compatability. I'm thinking about adding a separate field that is a list of _tmp fields, that the API could return when bulk scored?

paglias: Yeah so that's why I suggested making a simpler version even if it results in more save requests, right now the API for scoring is geared towards one single task and I feel like it would take some time to correctly plan a new version that works well with multiple ones but I haven't looked in depth

phillipthelen:
Made some changes to have this fully work the way I originally intended. for multiple tasks there is now an individual list of the deltas and _tmp data. Single tasks behave the same.

If I'm not mistaking, it should now do 2 save operations for habits and dailies and 3 for todos, when scoring a single task. For multiple tasks it's `numberOfTasksScored + 1` or `numberOfTasksScored + 2` if at least one todo was scored.

website/server/libs/taskManager.js
```diff
@@ -281,3 +291,176 @@ export function moveTask (order, taskId, to) {
...
+  resolve();
+}
+
+export async function scoreTasks (user, taskScorings, req, res) {
+  let taskIds = taskScorings.map(taskScoring => taskScoring.id).filter(id => id !== undefined);
+  if (taskIds.length === 0) throw new BadRequest(res.t('taskNotFound'));
+  let tasks = {};
+  (await Tasks.Task.findMultipleByIdOrAlias(taskIds, user._id)).forEach(task => {
+    tasks[task._id] = task;
+    if (task.alias && task.alias.length > 0) {
```
paglias: I still think this will result in tasks with an alias being added twice

website/server/libs/taskManager.js
```diff
@@ -281,3 +291,176 @@ export function moveTask (order, taskId, to) {
...
+      tasks[task.alias] = task;
+    }
+  });
+  if (Object.keys(tasks).length === 0) throw new NotFound(res.t('taskNotFound'));
+  let scorePromises = [];
+  taskScorings.forEach(taskScoring => {
+    if (tasks[taskScoring.id] !== undefined) {
```
paglias: I'd check for `!tasks[taskScoring.id] ` to also catch `null` values

website/server/controllers/api-v3/tasks.js
```diff
@@ -550,139 +547,15 @@ api.scoreTask = {
...
-    if (taskOrderPromise) promises.push(taskOrderPromise);
-    let results = await Promise.all(promises);
+    const data = await scoreTasks(user, [{id: taskId, direction}], req, res);
 
-    let savedUser = results[0];
-
-    let userStats = savedUser.stats.toJSON();
-    let resJsonData = _.assign({delta, _tmp: user._tmp}, userStats);
+    let userStats = user.stats.toJSON();
+    let resJsonData = _.assign({delta: data.taskResponses[0].delta, _tmp: data.taskResponses[0]._tmp}, userStats);
```
paglias, line 556: hmm so only one drop result is shown?

phillipthelen: This is for the single scoring API call, so yes. Since only one task is scored, there can only be one drop anyway

website/server/libs/taskManager.js
phillipthelen: Yes. They will be added twice to this object. But they will only be processed and saved once. This was needed, since the list it iterates on for scoring the tasks might not have the task ID and only has the alias. Having both the ID and alias available is the easiest way around this. Since the object (should) use references to the same task anyway the overhead should be negligible.

phillipthelen:
Could we push this to delta for testing maybe?

paglias:
@phillipthelen looks like some tests are failing

paglias:
and a new conflict :(

paglias:
Closed in favor of #11389

PR #11303, opened 2019-08-07
feat(styles): improve weekday boxes

### LINKED ISSUE #10041: hard to tell which Daily repeat days are ticked
If a Daily has only some days selected, as in the screenshot below, it's hard to work out which day the tick marks apply to for any days except Sunday and Saturday. We had a bug report about this from someone whose Dailies were becoming due on the wrong days.

It would be clearer if the spacing on either side of the checkboxes wasn't equal (i.e., each checkbox was closer to its own label) or if something else more clearly tied the checkboxes to their labels.

screenshot_2018-02-23_07-38-30

@Tressley

SebastiaanNijland:

Fixes #10041 

### Changes
- Adds styles specifically for the weekday selection

----
UUID: 2f3f22af-3abf-4236-bf9e-67ff88051df9


SebastiaanNijland:
Note that this may need some more testing/finetuning, but I'd like some feedback on this first :) Is this a good solution for you?

paglias:
Hi @SebastiaanNijland sorry for the delay and thanks for the PR!

I just tested it and it looks like there are some visual problems at least on chrome, I'm attaching a couple of screenshots

Schermata 2019-08-14 alle 12 16 46
Schermata 2019-08-14 alle 12 16 50

SebastiaanNijland:
I'll have a look!

paglias:
thanks @SebastiaanNijland ! The checkboxes are gone and the behavior seems to be fine but I want to pass this through our designer

Tressley:
Seeing some inconsistencies between the mock-up and the implementation. Mainly the border-colors and border-radius.

@SebastiaanNijland -- Are you a part of our Zeplin project? We use Zeplin to hand-off designs to developers to assist with implementation. If you'd like to be invited, please leave an email to invite.

Once you're a member, here's the Zeplin link for reference: https://zpl.io/a3gknQy

SebastiaanNijland:
> Seeing some inconsistencies between the mock-up and the implementation. Mainly the border-colors and border-radius.
> 
> @SebastiaanNijland -- Are you a part of our Zeplin project? We use Zeplin to hand-off designs to developers to assist with implementation. If you'd like to be invited, please leave an email to invite.
> 
> Once you're a member, here's the Zeplin link for reference: https://zpl.io/a3gknQy

I am not, you can send an invite to sebastiaannijland@gmail.com - thanks!

paglias:
thanks @SebastiaanNijland ! this is not part of the zeplin design but could you also add a `cursor: pointer;` css rule for the `:hover`, `:focus` and `:active` states?

Tressley:
@SebastiaanNijland -- Still seeing some small items. Here's an updated Zeplin link that better exemplifies the `:hover`, `:focus`, and `:active` states. Also, would you mind removing the transitions for now?

**Zeplin link:** https://zpl.io/aRG3XvN

SebastiaanNijland:
> @SebastiaanNijland -- Still seeing some small items. Here's an updated Zeplin link that better exemplifies the `:hover`, `:focus`, and `:active` states. Also, would you mind removing the transitions for now?
> 
> **Zeplin link:** https://zpl.io/aRG3XvN

The transitions are there by default, I will disable them for these checkboxes specifically, though that is a bit of a code smell.

I'll take a look at the changes for the hover / focus / active states!

paglias:
@SebastiaanNijland sorry again but it looks like the borders between two days get duplicated

Schermata 2019-09-20 alle 16 57 47

We might have to only apply the border right to every element and then the left side one for the first one

SebastiaanNijland:
I made it so that every border is always visually 1px wide, so between two days there is a .5px border from each box. This is because else the border colors will be incorrect for deselected days that are next to selected days, like in your example.

paglias:
Oh I see then I think it might be a good idea to change the inner borders to be 0.25 so together they match the 0.5px of the other borders? Because this way the difference is noticeable. But @Tressley will know more

SebastiaanNijland:
Well, currently all borders actually are the same size - but I guess it might not look that way indeed. I am fine with your solution as wel, I implemented it that way at first (but then some borders will appear to have the incorrect color, if you uncheck certain days).

paglias:
> Well, currently all borders actually are the same size

I get it but the problem is that between two different days there are two 0.5px borders that add up to a 1px border while on the external they're all 0.5.

If we change the right side only border to be 0.25 then they should add up to 0.5px like the other borders while keeping the correct color

SebastiaanNijland:
> > Well, currently all borders actually are the same size
> 
> I get it but the problem is that between two different days there are two 0.5px borders that add up to a 1px border while on the external they're all 0.5.

That is indeed true for the top and bottom borders - but I can easily change those to be 1px as well. I think that would be the better solution in this case! Will take a look soon(ish).

paglias:
Ok, so it looks like the border still looks weird... sorry about that.

One solution seems to use a negative margin like Bootstrap does

> They applied a margin of -1px to every button with the exception of the first-child.

An example is here https://www.w3schools.com/bootstrap/tryit.asp?filename=trybs_button_group&stacked=h and it seems to solve the issue with the border being larger between buttons

SebastiaanNijland:
So the issue is that the .5px border probably gets painted incorrectly - adding a negative margin sadly doesn't really fix it. Also, when using a .25px border, it doesn't render at all.

> We might have to only apply the border right to every element and then the left side one for the first one

I think this is the best solution in this case. I'll update the pr.

SebastiaanNijland:
So the issue is that the .5px border probably gets painted incorrectly - adding a negative margin sadly doesn't really fix it. Also, when using a .25px border, it doesn't render at all.

> We might have to only apply the border right to every element and then the left side one for the first one

I think this is the best solution in this case. I'll update the pr.

Tressley:
@SebastiaanNijland -- I think you're making this more difficult than it needs to be. There isn't a reason to use non-whole numbers in the `border-width` property. We're not trying to combine two borders, but rather overlay the previous button's border. As you can see in my screenshot below, I was able to do this by simply adding `margin-left: -1px;` to `.#task-modal .weekday-check` to produce the expected result.

image

SebastiaanNijland:
I was trying to implement a solution that would show the correct border-colors based on whether the checkboxes are selected or not, but indeed that made it impossible to have the correct border-widths.

I am not a big fan of negative margins in this case, it feels like a code smell to me. It would also result in the same layout as using a 1px border in the way I implemented it now.

Tressley:
I disagree. Negative margins aren't a code smell unless they're covering up an error that was made elsewhere. The next element in order overlaying the previous element's border isn't an error as much as how the HTML is parsed and rendered.

SebastiaanNijland:
I fully agree, it is a very useful feature, and indeed it's only a code smell if you use it to cover up an error in other code - but that is exactly what it would do in this case. To me, it feels like it would cover up the border of the previous element - in which case it would be better to simply not set that border at all.

SebastiaanNijland:
Hi, I'm not sure what more work is still needed for this one - feel free to either merge or close this PR, all my spare programming time is currently used up by a different project :)

paglias:
This is being addressed in #12142

PR #11306, opened 2019-08-08
Add translated gear name to open-mystery-item call

phillipthelen:
To make it easier for the clients to display the mystery item you received (and also to prevent issues if the data wasn't yet refreshed), this adds the translated name for the gear you receive.

### DIFF
--- a/test/common/ops/openMysteryItem.js
+++ b/test/common/ops/openMysteryItem.js
@@ -36,7 +36,9 @@ describe('shared.ops.openMysteryItem', () => {
 
     expect(user.items.gear.owned[mysteryItemKey]).to.be.true;
     expect(message).to.equal(i18n.t('mysteryItemOpened'));
-    expect(data).to.eql(content.gear.flat[mysteryItemKey]);
+    let item = content.gear.flat[mysteryItemKey];
+    item.text = content.gear.flat[mysteryItemKey].text();
+    expect(data).to.eql(item);
     expect(user.notifications.length).to.equal(0);
   });
 });
--- a/test/common/ops/openMysteryItem.js
+++ b/test/common/ops/openMysteryItem.js
@@ -36,7 +36,7 @@ describe('shared.ops.openMysteryItem', () => {
 
     expect(user.items.gear.owned[mysteryItemKey]).to.be.true;
     expect(message).to.equal(i18n.t('mysteryItemOpened'));
-    let item = content.gear.flat[mysteryItemKey];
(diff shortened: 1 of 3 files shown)


test/common/ops/openMysteryItem.js
```diff
@@ -36,7 +36,9 @@ describe('shared.ops.openMysteryItem', () => {
 
     expect(user.items.gear.owned[mysteryItemKey]).to.be.true;
     expect(message).to.equal(i18n.t('mysteryItemOpened'));
-    expect(data).to.eql(content.gear.flat[mysteryItemKey]);
+    let item = content.gear.flat[mysteryItemKey];
```
paglias: hmm this changes the value of the item text for all the tests after this, we shouldn'd really mutate this object

website/common/script/ops/openMysteryItem.js
```diff
@@ -24,6 +24,7 @@ module.exports = function openMysteryItem (user, req = {}, analytics) {
   if (mysteryItems.length === 0) markNotificationAsRead(user);
 
   item = cloneDeep(content.gear.flat[item]);
+  item.text = content.gear.flat[item.key].text();
```
paglias: shouldn't this be passed the language?

test/common/ops/openMysteryItem.js
paglias: also I'm not sure this is necessary, I mean before it returned the english text? In that case it should work anyway as the tests should use english unless otherwise specified

paglias (changes requested):
I'm not sure about this, I only looked at the code but let me know if you want me to give this a test

test/common/ops/openMysteryItem.js
phillipthelen: without the line 40 here, the item does not contain the name in english. instead `text` is a function reference. Which is also why the API doesn't return it, since the function was never actually called.

Not sure how I should handle this test otherwise? Should I create a copy of the item and use that?

website/common/script/ops/openMysteryItem.js
phillipthelen: I assumed it would get the language automatically, but you're probably right.

test/common/ops/openMysteryItem.js
paglias: yeah I'd probably clone it with lodash's cloneDeep function

paglias:
LGTM!

PR #11307, opened 2019-08-09
Always send push notifications in recipients language

### LINKED ISSUE #8718: Push notifications are localized in the language of the sender
When push notifications are sent from p.e. a quest invitation, these are localized using the language of the sender and not the languages of the recipients.

phillipthelen:
This fixes issue #8718. Some notifications were already handling this correctly, so I didn't update all of them.

Fixes #8718

### DIFF
--- a/website/server/libs/pushNotifications.js
+++ b/website/server/libs/pushNotifications.js
@@ -35,8 +35,6 @@ function sendNotification (user, details = {}) {
   let payload = details.payload ? details.payload : {};
   payload.identifier = details.identifier;
 
-  console.log(details);
-
   _.each(pushDevices, pushDevice => {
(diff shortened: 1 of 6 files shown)


website/server/libs/pushNotifications.js
```diff
@@ -35,6 +35,8 @@ function sendNotification (user, details = {}) {
   let payload = details.payload ? details.payload : {};
   payload.identifier = details.identifier;
 
+  console.log(details);
```
paglias: can you remove this?

paglias (changes requested):
One console.log to remove but otherwise looks good!

paglias (approved):

PR #11318, opened 2019-08-16
gear: Fix stats for the "Brilliant Bloom" weapon

aszlig:
This weapon has been introduced in the Summer Splash of 2019.

With an INT bonus of 15 and also being a one-handed weapon, I think it is way too powerful given that it has a low price and also should be relatively easy to get.

I've set the stats to match the mage gear for other events and appropriately made it a two-handed weapon.

Also included small migration script which unequips the shield if a player happens to have the Brilliant Bloom and a shield.

### DIFF
--- a/migrations/archive/2019/20190816_brilliant_bloom_fix.js
+++ b/migrations/archive/2019/20190816_brilliant_bloom_fix.js
@@ -1,0 +1,52 @@
+/* eslint-disable no-console */
+const MIGRATION_NAME = '20190816_brilliant_bloom_fix';
+import { v4 as uuid } from 'uuid';
+
+import { model as User } from '../../../website/server/models/user';
+
+const progressCount = 1000;
+let count = 0;
+
+async function updateUser (user) {
+  count++;
+  if (count % progressCount === 0) console.warn(`${count} ${user._id}`);
+  return await User.update(
+    { _id: user._id },
+    { $unset: {'items.gear.equipped.shield': 1} }
+  ).exec();
+}
+
+module.exports = async function processUsers () {
+  let query = {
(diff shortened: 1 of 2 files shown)


SabreCat:
The Summer Splash 2019 Mage gear was unusual in that it included a shield, the Drops of Pure Water: https://github.com/HabitRPG/habitica/blob/develop/website/common/script/content/gear/sets/special/index.js#L3925-L3933 Implementing this PR would mean players couldn't equip the two thematically connected items together as intended.

Between the two items, the price and stats are equivalent to that of a two-handed Mage item. While this does allow a bit more minmaxing of INT than usual, in that the player can use a shield-hand item with +INT alongside this high-INT mainhand, it's not really a problem.

aszlig:
@SabreCat: Okay, so it *was* intended, I wasn't quite sure because to my knowledge makes it the best weapon available if you're maximizing on INT. Something like the Nomad's Scimitar would be roughly equal (+1 bonus) but you don't get the class bonus on that one. So if you say use the Wakizashi and the Brilliant Bloom as a mage, you'll get a bonus of 39.5.

SabreCat:
I suppose if we'd been concerned about the bit of power creep here, we could have had both the shield and main-hand weapon have split INT and PER values. But I'm OK with rewarding folks for going all-in on optimizing their gear sets!

(Stats could use a high-level rebalancing pass again anyway, I fear...)

wolverdude:
@SabreCat It looks like your Fall Festival refactor introduced a bug in this weapon as the stats now include 7 PER. It looks like you just need to override the PER in the original definition. I'll go ahead and submit a PR for that.

Screen Shot 2021-01-03 at 12 15 24 PM

PR #11319, opened 2019-08-16
Add push notification for party activity

phillipthelen:
This adds a new type of push notification for party activity. Additionally I added some metadata to the inbox push notifications to make them more appealing on android, since I noticed that most of the android push notifications were not working correctly.

This PR deploy has to be coordinated with the [Android 2.1](https://github.com/HabitRPG/habitica-android/milestone/12) update and the [iOS 2.4.1](https://github.com/HabitRPG/habitica-ios/milestone/16). Since otherwise the apps will not handle the notifications correctly and not open the party page when tapped.


website/server/models/user/schema.js
```diff
@@ -500,6 +500,7 @@ let schema = new Schema({
       questStarted: {$type: Boolean, default: true},
       invitedQuest: {$type: Boolean, default: true},
       majorUpdates: {$type: Boolean, default: true},
+      partyActivity: {$type: Boolean, default: true},
```
paglias, line 506: Hmm the settings page on the website doesn't support push notifications settings that don't have an email setting so it'll need a small change

website/server/controllers/api-v3/chat.js
```diff
@@ -229,6 +229,8 @@ api.postChat = {
     if (group.type === 'party') {
       user.party.lastMessageSeen = newChatMessage.id;
       toSave.push(user.save());
+
+      sendChatPushNotifications(user, group, newChatMessage, res.t);
```
paglias: Can this be moved to inside the `group.sendChat ` function code? Since the other notifications code is there already

website/server/models/user/schema.js
phillipthelen: Do we want/need the setting there? Without a mobile device, the setting is useless anyway. I don't think making the list of notifications longer on the website would help.

paglias: right! If we don't care about having it on web (and indeed it doesn't really make sense) that's fine

paglias:
I haven't tested this but the code looks good if you want to merge it (there's a conflict)

phillipthelen:
I'll wait with resolving the conflict until the android and iOS updates are released, since we can't release this before that.

paglias:
Before it's released there seem to be some failing tests

phillipthelen:
Improved the way the notifications are displayed on iOS (this change affects all push notifications, since now they all have a title and a body which makes them more in line with the android push notifications). The mobile updates are also both rolled out, so this PR is ready for merging from my end

paglias:
Looks like there's a new conflict here

PR #11324, opened 2019-08-22
This change prevents checklists from being toggled when casting skills or spells.

### LINKED ISSUE #11028: Using a skill on a task can check off checklist items
_EDIT: See the first comment for a confirmation of the desired fix._

> thebat137   @thebat137 (dea8a9e8-c084-438a-9a14-7fe96a73fd0a) [web] 
> I am seeing what I believe is a regression in the way using skills on tasks works.
> 
> What I expect (and what I think used to happen): I click on the skill, my cursor picks up the skill, and then I click anywhere on the box for a task and the skill is applied to that task, without checking any checkboxes or otherwise affecting the task status.
> 
> What actually happens: If, while my cursor is carrying a skill, I click on a task and happen to do it in an area belonging to one of that task's checklist items, not only is the skill applied to the task, the checklist item is also checked off (or unchecked if it was already checked).
> 
> I'm not sure if other skills have this problem, but the one I've been noticing it with is the warrior skill Brutal Smash.

Tested and confirmed in Windows 10 / Chrome.

I agree with the poster that this behaviour is unexpected, and I think at the very least should be consistent (i.e. clicking with the skill near the main checkbox should check off the whole task, if clicking with the skill can check off the subtasks).

rjbernaldo:
Fixes #11028

### Changes
The top level event handler `castEnd` was being called before the `toggleChecklistItem` event and thus all checks for `if (this.castingSpell) return;` do not get triggered.

All in all a very challenging issue as I have not worked with vue before (only react/redux) but very satisfying nonetheless.

Thanks for the opportunity @Alys .

----
UUID: 3c8f6348-db5f-4162-82af-590260c05f89

### DIFF
--- a/website/client/components/tasks/task.vue
+++ b/website/client/components/tasks/task.vue
@@ -1,6 +1,6 @@
 <template lang="pug">
 .task-wrapper
-  .task
+  .task(@click='castEnd($event, task)')
     approval-header(:task='task', v-if='this.task.group.id', :group='group')
     .d-flex(:class="{'task-not-scoreable': isUser !== true}")
       // Habits left side control
@@ -690,8 +690,6 @@ export default {
 
       if (!this.$store.state.spellOptions.castingSpell) {
         this.$emit('editTask', task);
-      } else {
-        this.$root.$emit('castEnd', task, 'task', e);
       }
     },
     moveToTop () {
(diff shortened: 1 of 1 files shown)


paglias:
sorry for the delay @rjbernaldo ! I checked out the PR and it works well! 
I'm merging the PR and awarding your contributor tier.

rjbernaldo:
Happy to help @paglias. Til the next one 👍

PR #11325, opened 2019-08-23
Hourglass Quest

SabreCat:

### Changes

* Introduces the ability to buy Quests marked with a category of `timeTravelers` from the Time Travelers Shop using Mystic Hourglasses.
* Adds the first such quest, "Mysterious Mechanical Marvels", for Robot pet eggs


website/server/controllers/api-v3/user.js
```diff
@@ -1000,6 +1000,10 @@ api.userPurchaseHourglass = {
   url: '/user/purchase-hourglass/:type/:key',
   async handler (req, res) {
     let user = res.locals.user;
+    req.params.hourglass = true;
```
paglias: Hmm this feels hacky... `req.body.quantity` is lacking validation on it being a number and apidocs and I'd not change the original `req.params` object

Is it necessary for quantity to be in req.params? Cannot you access `req.body.quantity` in `common.ops.buy`?

website/client/store/actions/shops.js
```diff
@@ -112,12 +112,13 @@ export function purchaseMysterySet (store, params) {
...
 export function purchaseHourglassItem (store, params) {
+  const quantity = params.quantity || 1;
   const user = store.state.user.data;
-  let opResult = hourglassPurchaseOp(user, {params});
+  let opResult = hourglassPurchaseOp(user, {params, quantity});
 
   return {
     result: opResult,
-    httpCall: axios.post(`/api/v4/user/purchase-hourglass/${params.type}/${params.key}`),
+    httpCall: axios.post(`/api/v4/user/purchase-hourglass/${params.type}/${params.key}`, {quantity}),
```
negue, line 121: if quantity would be also a `/{params.quantity}`-param, then it wouldn't have to be in body

SabreCat: That would require creating a whole new route, wouldn't it?

website/server/controllers/api-v3/user.js
SabreCat: Hah, this was modeled after some other routes where we do weird things with `req.params.quantity`, but you're absolutely right this was one of the things I was nervous about. 78de2a5 better?

website/client/store/actions/shops.js
negue: I think when you add a param it would still accept the route (without the newly added one)

PR #11330, opened 2019-08-25
Conversation-API: Add last message 

negue:
Refactored the conversation query (thx to @phillipthelen).

The Website UI already used the same properties, so those are now `auto-filled` (instead of filled on selecting a conversation)


phillipthelen (approved):

paglias (commented):
LGTM but do we need to update / add any test or api docs?

Issue #11257, opened 2019-07-08
Update achievements to work with notification object


phillipthelen:
The way achievements are displayed on web is often done by comparing the fields directly. Instead the server should update the `notifictations` property on the user to add information about the gained notification. Same way it is already done for the [Ultimate Gear achievements](https://github.com/HabitRPG/habitica/blob/0b8ce63c760838cd56e94ceea0a1df554ff72c0a/website/common/script/fns/ultimateGear.js#L19) or [Back to basics](https://github.com/HabitRPG/habitica/blob/34faf2fd50eb06987f50352726d6c5c3af7b849c/website/common/script/ops/hatch.js#L50). This is essential for the achievements to work properly across the mobile apps and the website.

List of achievements that need to be updated:

- [Beast Master/Mount Master/Triad Bingo](https://github.com/HabitRPG/habitica/blob/043696c225bda140fccbb412d22a435da68b6656/website/server/models/user/hooks.js#L190)
- [Party Up](https://github.com/HabitRPG/habitica/blob/7baa7427a044060bb0fa0444f98ef34e4de9dbeb/website/server/controllers/api-v3/groups.js#L609)
- [Party On](https://github.com/HabitRPG/habitica/blob/7baa7427a044060bb0fa0444f98ef34e4de9dbeb/website/server/controllers/api-v3/groups.js#L615)
- [Mind over matter](https://github.com/HabitRPG/habitica/blob/34faf2fd50eb06987f50352726d6c5c3af7b849c/website/client/components/header/notifications/mindOverMatter.vue)
- [Lost Masterclasser](https://github.com/HabitRPG/habitica/blob/34faf2fd50eb06987f50352726d6c5c3af7b849c/website/client/components/header/notifications/lostMasterclasser.vue)
- [just add water](https://github.com/HabitRPG/habitica/blob/34faf2fd50eb06987f50352726d6c5c3af7b849c/website/client/components/header/notifications/justAddWater.vue)

I marked this as Help Wanted in case a contributor wants to work on one or multiple of these before I have time.

Xaz16:
Hello @phillipthelen!
Is it still Help wanted issue? As I can see we already have PR for that

paglias:
@Xaz16 I'll remove the label as this is being handled by @phillipthelen

Xaz16:
@paglias thank you :)

beffymaroo:
@SabreCat Should this one get a Help Wanted tag again?

[...]

PR #11332, opened 2019-08-26
Add more achievements to notifications list

phillipthelen:
This is for #11257 

Currently the website shows a persistent toast for the achievements. Not sure if we want to change that. We also need better copy for the `partyUp` and `partyOn` achievement notifications.


website/server/models/user/hooks.js
```diff
@@ -254,6 +258,10 @@ schema.pre('save', true, function preSaveUser (next, done) {
         this.notifications.push(lastExistingNotification);
       }
     }
+
+    notificationsToAdd.forEach(notification => {
```
paglias: since we've been using it across the codebase could you change this to use `user.addNotification()` instead of pushing to the array ?

paglias: Also no need to create a temporary array and then push them together at the end

website/client/components/notifications.vue
```diff
@@ -165,6 +165,31 @@ const NOTIFICATIONS = {
...
+    achievement: true,
+    label: ($t) => `${$t('achievement')}: ${$t('achievementPartyUp')}`,
+    modalId: 'generic-achievement',
+  },
+  ACHIEVEMENT_PARTY_ON: {
+    achievement: true,
+    label: ($t) => `${$t('achievement')}: ${$t('achievementPartyOn')}`,
+    modalId: 'generic-achievement',
+  },
+  ACHIEVEMENT_BEAST_MASTER: {
```
paglias, line 188: for the beast master, mount master and triad bing achievements we didn't have anything to show them on the website?

paglias (changes requested):
left some comment about small changes and a question

I would also add some tests about the notifications being added correctly

website/client/components/notifications.vue
phillipthelen: No, we didn't show anything when people got that achievements. Just checked again on prod to make sure.

phillipthelen:
Found another issue where notifications weren't added correctly and added two new tests

paglias:
Looks like some tests broke

PR #11344, opened 2019-09-03
Make saddles not sellable

### LINKED ISSUE #11280: Saddles should not be sellable using the API
On the website, if you try to sell a Saddle, you're prevented from doing so and you see the message "Hey! This is a pretty useful item! Are you familiar with how to use a Saddle with your Pets?" (that's the locales string `foodSaddleSellWarningNote`). This was implemented because people were selling the Saddle by accident. It was based on the assumption that no one will ever want to sell a Saddle, which I think is still a correct assumption since Saddles are a very rare prize.

However the code to prevent selling it is only in the website's front end, which means you can still sell it using the API, and therefore using the mobile apps too (one person reported today that they sold a Saddle by accident on the app).

If the API prevents selling Saddles, then it won't be possible to do so on the apps. A user who tried to might see an ugly error message but I think that will be an improvement over accidentally losing a Saddle. The apps could be adjusted later as a lower priority to give better feedback when trying to sell.

~I'll leave this issue as suggestion-discussion for a few days in case the staff disagree but then if there's no objections I'll mark it as help wanted.~

This is the website code that prevents selling:
https://github.com/HabitRPG/habitica/blob/20fc233ad7116385594deff19ae619059e21c2d8/website/client/components/shops/market/sellModal.vue#L24-L26

It uses the `sellWarningNote` parameter in the Saddle's `content` data, which should already be available through the API:
https://github.com/HabitRPG/habitica/blob/20fc233ad7116385594deff19ae619059e21c2d8/website/common/script/content/index.js#L484-L492

The API route for selling is [/api/v3/user/sell/:type/:key](https://habitica.com/apidoc/#api-User-UserSell), which is defined here:
https://github.com/HabitRPG/habitica/blob/20fc233ad7116385594deff19ae619059e21c2d8/website/server/controllers/api-v3/user.js#L1238
although this function is probably the best place to make this change:
https://github.com/HabitRPG/habitica/blob/20fc233ad7116385594deff19ae619059e21c2d8/website/common/script/ops/sell.js#L15-L54

Note that a Saddle is defined as `food`, which is included in the `ACCEPTEDTYPES` constant (referred to in the above code and defined in the code below), so that's why it's currently sellable. We don't want to stop treating it as food because that would involve too many changes, some of them breaking changes.
https://github.com/HabitRPG/habitica/blob/20fc233ad7116385594deff19ae619059e21c2d8/website/common/script/ops/sell.js#L13

gpalsingh:

Fixes #11280

### Changes

Added check for items with `sellWarningNote` like `Saddle`.
I'm not sure if it's okay to just reuse existing error string or create a new one for this case because that'd require additional translations to be added.

----
UUID: 9c3d2300-1bc4-4898-8a78-b7ffe8ac8e37

### DIFF
--- a/test/common/ops/sell.js
+++ b/test/common/ops/sell.js
@@ -85,6 +85,19 @@ describe('shared.ops.sell', () => {
     }
   });
 
+  it('returns error when trying to sell Saddle', (done) => {
+    const foodType = 'food';
+    const saddleKey = 'Saddle';
+    user.items[foodType][saddleKey] = 1;
+    try {
+      sell(user, {params: {type: foodType, key: saddleKey}});
+    } catch (err) {
+      expect(err).to.be.an.instanceof(NotAuthorized);
+      expect(err.message).to.equal(i18n.t('foodSaddleSellWarningNote'));
+      done();
+    }
+  });
+
   it('reduces item count from user', () => {
     sell(user, {params: { type, key } });
 
--- a/website/client/components/shops/market/sellModal.vue
+++ b/website/client/components/shops/market/sellModal.vue
@@ -21,7 +21,7 @@
 
         h4.title {{ itemContextToSell.itemName }}
 
-        div(v-if="item.sellWarningNote")
+        div(v-if="item.text() === $t('foodSaddleText')")
           div.text {{ item.sellWarningNote() }}
           br
 
--- a/website/common/script/ops/sell.js
+++ b/website/common/script/ops/sell.js
@@ -44,7 +44,7 @@ module.exports = function sell (user, req = {}) {
   }
 
   if (content[type][key].sellWarningNote) {
-    throw new NotAuthorized(i18n.t('positiveAmountRequired', req.language));
+    throw new NotAuthorized(content[type][key].sellWarningNote());
   }
 
   user.items[type][key] -= amount;
--- a/website/common/script/ops/sell.js
+++ b/website/common/script/ops/sell.js
@@ -43,6 +43,10 @@ module.exports = function sell (user, req = {}) {
     throw new NotFound(i18n.t('userItemsNotEnough', {type}, req.language));
   }
 
+  if (content[type][key].sellWarningNote) {
+    throw new NotAuthorized(i18n.t('positiveAmountRequired', req.language));
+  }
+
   user.items[type][key] -= amount;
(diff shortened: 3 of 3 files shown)


Alys:
The `positiveAmountRequired` text isn't really suitable for use for this purpose. If you go to the root directory of your local install and type this: `git grep positiveAmountRequired` you'll see all the places where that string appears in the source code, and it will show you that the English text that it represents is "Positive amount is required". So that won't really help any API users who try to sell a Saddle - they won't understand why they can't sell it.

Fortunately there's already a string that you can use. Have a look at the code snippets displayed in the issue. The first snippet (from the sellModal.vue file) is the code that displays a message to a user when they try to sell a Saddle using the website. `item.sellWarningNote()` contains that text. See if you can use that same attribute to display the message when a user uses the API to try to sell a Saddle. Post here if you have questions! :)

gpalsingh:
Hi, I made the changes. `sellWarningNote` is always in english though. What are the ways to fix it without having to make changes in the vue file?
Thanks!

Alys:
`content[type][key].sellWarningNote()` is a strong of text so the translated version can be inserted using the same function as for other text. See how `userItemsNotEnough` is translated just above your new code.

paglias:
Hi @gpalsingh and thanks for the PR! The error message looks good but given that only the Saddle has the `sellWarningNote` I feel like it can be a bit confusing to rely on the existence of that field to prevent selling the Saddle.

The best solution would be to have a `canSell` function on each item like we have the `canBuy` one but since this is just for the saddle now I think it's fine to check explicitly for it.

I'd simply change `if (content[type][key].sellWarningNote) {` to `if (type === 'food' && key === 'Saddle') { ...```

I would also add a test to `test/common/ops/sell.js`, let me know if you need any help :)

Alys:
@paglias Ah sorry, it was my idea to use `sellWarningNote` in the `if` statement (https://github.com/HabitRPG/habitica/issues/11280#issuecomment-527371363). Would you want the existing Vue code changed as well?

paglias:
Ah I missed that, in this case yeah changing the Vue code as well would be better

gpalsingh:
HI, I made the changes and added the test

paglias:
@gpalsingh  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11345, opened 2019-09-04
Fix: reset streak for habits after rebirth

### LINKED ISSUE #11334: Orb of Rebirth should reset Habit streaks
Reported via RaB by Eliot the New Recruit [tmto]   @EliotThe (0d77107e-9128-4656-8c02-6e06331fa396):

> The streaks on a couple of my Habits, which reset monthly, did not reset to 0 when I rebirthed a couple days ago. I don't really mind, although I assume they were supposed to. The Habits are not part of a challenge---although, they were part of a challenge which I left and kept the tasks from.

I have not tested this personally.

Xaz16:

Fixes https://github.com/HabitRPG/habitica/issues/11334

### Changes
- Modify current test of reseting streaks
- Add new if branch for task with "habit" type

----


Alys:
@Xaz16 If you test this in your local install, I think you'll find that it doesn't cause the Habit streaks to be reset. I know I said in the issue that the streaks should be reset in the same was as for Dailies, but I'm guessing that was confusing - sorry about that! The code won't be exactly the same for Habits as for Dailies.

Create some Habits in your local install, score them up and down, then examine the data for those Habits in the database, either by doing a `db.tasks.find` mongo search (there'll be some advice about running mongo searches in the wiki) or by using your local install's API to fetch your test user's tasks. By examining the data and comparing it to what you see in your local instal's website, you'll see how streaks on Habits are recorded and you can use that information to work out the code to use to reset the streaks.
You could also look at the data schema for Habits (and the other task types) in the `website/server/models/task.js` file.

Xaz16:
@Alys 
Sure, sorry it is my fault. I don't know why i'm not checked models. I'm tested it with "streak" key reseting of habit and I was really sure that I see how it was reset to zero. 

Now I'm checked model, make fix for habit and checked it locally

PR #11347, opened 2019-09-04
Teams Updates 201908

SabreCat:

### Changes

* Padlocked Group Plan tasks no longer have a hover state
* Various style changes for assigned/unassigned/etc. task states to better match designs
* Managers and leaders receive notifications when a task is claimed
* Adds `auth.timestamps.updated` that holds a timestamp when the user last made a significant interaction with the app
* Uses the updated timestamp to tell if an assigned group task should age or cause damage at cron--if it was assigned since last interaction, don't penalize the user


website/client/components/tasks/approvalFooter.vue
```diff
@@ -76,17 +83,18 @@ export default {
         return this.$t('taskIsUnassigned');
       }
     },
+    userIsManager () {
```
paglias, line 86: there's already an `isManager` function in `website/client/components/groups/group.vue`, maybe we can move it to a mixin and use it instead of duplicating the code?

website/client/components/tasks/approvalFooter.vue
```diff
@@ -76,17 +83,18 @@ export default {
...
+      if (this.group && (this.group.leader.id === this.user._id || this.group.managers[this.user._id])) return true;
+    },
     approvalRequested () {
-      if (this.task.approvals && this.task.approvals.length === 1) return true;
+      if (this.task.approvals && this.task.approvals.length === 1 || this.task.group && this.task.group.approval && this.task.group.approval.requested) return true;
```
paglias, line 90: I know this isn't the part that got changed in the PR but where is `task.approvals ` coming from? I don't see it in the task model

website/server/models/group.js
```diff
@@ -1431,6 +1431,7 @@ schema.methods.syncTask = async function groupSyncTask (taskToSync, user) {
     matchingTask.group.id = taskToSync.group.id;
     matchingTask.userId = user._id;
     matchingTask.group.taskId = taskToSync._id;
+    matchingTask.group.assignedDate = moment().toDate();
```
paglias: no need to use moment here I think :), `new Date()` should work as well

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -200,13 +200,25 @@ api.assignTask = {
...
+    const userName = user.profile.name;
+
+    if (user._id === assignedUserId) {
+      const managerIds = Object.keys(group.managers);
+      managerIds.push(group.leader);
+      const managers = await User.find({_id: managerIds}, 'notifications').exec();
+      managers.forEach((manager) => {
+        if (manager._id === user._id) return;
+        manager.addNotification('GROUP_TASK_CLAIMED', {
+          message: res.t('taskClaimed', {userName, taskText}),
```
paglias: this will add the notification in the language of the user not of the manager

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -200,13 +200,25 @@ api.assignTask = {
...
-      const taskText = task.text;
-      const managerName = user.profile.name;
-
+    const taskText = task.text;
+    const userName = user.profile.name;
+
+    if (user._id === assignedUserId) {
+      const managerIds = Object.keys(group.managers);
+      managerIds.push(group.leader);
+      const managers = await User.find({_id: managerIds}, 'notifications').exec();
```
paglias: instead of loading all the managers and then saving them which is quite slow, can we just push the notification with a direct update? using there's `pushNotification` under the User model's static methods

website/server/models/user/schema.js
```diff
@@ -68,6 +68,7 @@ let schema = new Schema({
     timestamps: {
       created: {$type: Date, default: Date.now},
       loggedin: {$type: Date, default: Date.now},
+      updated: {$type: Date, default: Date.now},
```
paglias, line 71: for this one, instead of setting it manually can we use the `pre('save'` and `pre('update'`user hooks that we already use to increase the user version field?

paglias (commented):
Left some comments, I haven't tested the client side changes

website/client/components/tasks/approvalFooter.vue
SabreCat: hmm, turns out that function isn't used in /components/groups/group.vue. I'll remove it and the several other unused functions in there.

website/client/components/tasks/approvalFooter.vue
SabreCat: Oh yeah, that gave me quite the headache until I figured out how it works. It's set up in this block, calling the `GET /approvals` route: https://github.com/HabitRPG/habitica/blob/sabrecat/teams-201908/website/client/components/group-plans/taskInformation.vue#L200-L213

website/server/controllers/api-v3/tasks/groups.js
SabreCat: Hmm, can I pass the manager's `preferences.language` as an argument? I _think_ so from looking at `i18n.t` but not 100% sure...

website/server/models/user/schema.js
SabreCat: The problem with that is it'll cause the updated timestamp to change on any number of updates to the user that don't indicate they've interacted with the site or app. For instance, if they receive a notification, `pre` hooks will fire, but that happens without any interaction whatsoever on their part.

website/server/controllers/api-v3/tasks/groups.js
SabreCat: ohh, I thought that method was about _push_ notifications specifically, like for mobile devices 🤦‍♂ I'll try that!

website/server/controllers/api-v3/tasks/groups.js
paglias: Yeah let's use `manager. preferences.language `

website/server/controllers/api-v3/tasks/groups.js
paglias: Yeah that's a bit confusing

website/server/models/user/schema.js
paglias: Hmm right.. I still don't like having to set it manually, it's very error prone

website/client/components/tasks/approvalFooter.vue
paglias: Ah ok, so it's client side only. I guess it's fine now but we might want to move this to the server when we make groups available on mobile

PR #11350, opened 2019-09-06
Dockerfile: Modified git clone to only clone with depth 1

Zocker1999NET:

### Changes
Before this change docker downloads the whole repository with all commits for building the project. However the git history is not used inside the docker image, only the current version named `release`. So I added the option `--depth 1` to the git clone command so only the required version without the history will be downloaded inside the docker container. This saves disk space and speeds up the process of building the docker image.

See [this page of the git manual](https://www.git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt) to read about the option `--depth`.

----
UUID: f43b87aa-e434-4564-8932-76b76784edf0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN npm install -g gulp-cli mocha
 # Clone Habitica repo and install dependencies
 RUN mkdir -p /usr/src/habitrpg
 WORKDIR /usr/src/habitrpg
-RUN git clone --branch release https://github.com/HabitRPG/habitica.git /usr/src/habitrpg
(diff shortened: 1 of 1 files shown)


SabreCat:
Ahh, smart! That'll make the Docker images a lot more slender. Thank you!

I've awarded you your first Blacksmith tier on Habitica, congrats! Keep in mind that contributor tiers require increasing amounts of effort from one to the next, so tier 2 will take multiple PRs or a larger PR to attain, but keep helping out and we'll express our gratitude accordingly!

PR #11352, opened 2019-09-06
[Stlye] fix yesterdaily modal

negue:
Fixed the Modal based on zeplin.


website/client/components/yesterdailyModal.vue
```diff
@@ -7,27 +7,42 @@
...
+        :dueDate="dueDate",
+      )
+    .start-day.text-center
+      button.btn.btn-primary(@click='close()') {{ $t('yesterDailiesCallToAction') }}
 </template>
 
+<style lang="scss">
+  #yesterdaily {
+    .modal-dialog {
+      width: 362px;
```
paglias: @Tressley are we sure we want to use such precise pixels dimensions? In this case I think it will work well on small devices since the width is small but we risk making this very error prone due to unrelated future css changes

Tressley: @paglias -- I'd prefer we start using `rem` as our base units. If we set `font-size: 100%` on the root element, that gives us a 16px baseline to work with. In this case it'd become `width: 22.625rem;` If necessary, we could adjust the width slightly to use whole numbers instead.

paglias: @Tressley I agree about rems (we can already use them, bootstrap's base css already sets the correct size for the root element) but I feel like using absolute sizes in a responsive layout  is a bit problematic per se, for example in this case while nothing disastrous happens if you resize the screen up to a certain size you get this:

Schermata 2019-09-09 alle 16 06 54

A couple of pixels more and the modal gets wider:

Schermata 2019-09-09 alle 16 07 01

paglias:
@negue I think there's some with the Start My New Day button position (see the screenshot in the other discussion)

negue:
weird - I wonder what broke the margin to the `Start my day` button :/

paglias:
Is this ready to go?

Tressley:
I'm not sure when this happened, but there looks to be a bug with the purple overlay appearing and then disappearing. We'll need to fix that.

Also, there's a rogue 4px margin-bottom being added to `.task-wrapper`. Let's remove that and keep the 2px set on `.task`.

negue:
Changed the modal-backdrop (the bootstrap-vue changed the classes, there is no . show on .modal-backdrop anymore).

paglias:
Can be merged now?

Tressley:
This LGTM. 👍

PR #11353, opened 2019-09-07
Fix: Antidotes to Avatar Transformation Items should be added to Rewards by API

### LINKED ISSUE #10920: when buying new equipment, item does not disappear from Market or Seasonal Shop

### LINKED ISSUE #11289: Antidotes to Avatar Transformation Items should be added to Rewards by API
When you've been affected by one of the avatar [Transformation Items](https://habitica.fandom.com/wiki/Transformation_Items) (e.g., Spooky Sparkles), your Rewards column contains an antidote (e.g., Opaque Potion). Currently the antidote is inserted into your Rewards column by client-side code, so it's visible only on the website.

This should be changed so that the API inserts the antidote into a player's Rewards whenever they are affected by a Transformation Item. 
_(This will allow the antidote to appear in the mobile apps with little or no app-specific code being needed. Modifying the apps is not in scope for this issue)._

To find all the current website code for the antidote, use `git grep opaquePotion` (the other antidotes could also be searched for but their code will be in the same place as the Opaque Potion's code).

_(After the fix for this has gone live, we should check to see if the potion appears automatically on the mobile apps, and if not log issues to have that investigated.)_

Xaz16:

Fixes https://github.com/HabitRPG/habitica/issues/11289

### Changes

- Move logic of choosing proper debuf potion from vue component to website commons
- introduce new function to get debuffSpellItems
- introduce set debuff spell item function to understand what should we do with user which have some of buffs about transformation, delete or set antidote to pinned items in rewards
- introduce new type for debuffPotion to recognise and properly use it on client

At all I can say that it is minimal solution to satisfy the issue report, in fact I tried to move anitdotes to user inventory, but I face some extra work to make it work as "self" spell. So for now -- we retrieve it in pinned items of user data which we receive from data and I'm not significantly edit website code to work with such item. 

Here we have issue -- if I unpin the item from rewards sections I can't return it back due to we don't have it in inventory at all, if we check transformation items it is in inventory and I think antidotes should be too. To forbid unpin for such debuffPotion type items we can broke path in content of item, but it is not a clear solution for it...

I open for discuss to understand in which way we should store it and how it should work, feel free to participate :) 

----


Xaz16:
@Alys I have made preparation for getting correct antidote (moved it to common) function to use it then on server. Can you please review and answer if I'm going in right way? I'm really new in this project, so will be happy about your feedback :)

Alys:
@Xaz16 That's something that Habitica's staff should comment on so I'll leave it for them to answer you but thank you for checking with us!

Xaz16:
@Alys I finished work on this one, could you please put "ready for review" label? Thank you

website/common/script/index.js
```diff
@@ -77,6 +77,9 @@ api.updateStore = updateStore;
 import inAppRewards from './libs/inAppRewards';
 api.inAppRewards = inAppRewards;
 
+import setDebuffPotionItem from './libs/setDebuffPotionItem';
```
paglias: is there a specific reason why the `getDebuffPotionItem ` file isn't added as well?

website/common/script/libs/getItemInfo.js
```diff
@@ -177,6 +177,25 @@ module.exports = function getItemInfo (user, type, item, officialPinnedItems, la
         pinType: 'seasonalSpell',
       };
       break;
+    case 'debuffPotion':
+      itemInfo = {
```
paglias, line 181: Sorry but I'm not super familiar with this code, where is this taking the values for `item.mana `, item.cast, item.immediateUse... ? Because here https://github.com/HabitRPG/habitica/pull/11353/files#diff-d0d5d38815f41cc5237729df242e5ea1R17 they don't seem to exist

website/common/script/libs/setDebuffPotionItem.js
```diff
@@ -0,0 +1,30 @@
...
+      debuffPotionItems.forEach(debuffPotion => {
+        if (!isPresent) {
+          isPresent = debuffPotion.path === pinnedItem;
+        }
+      });
+      return isPresent;
+    });
+
+    if (!isUserHaveDebuffInPinnedItems) {
+      user.pinnedItems.push(...debuffPotionItems);
```
paglias: is this going to add duplicates for the debuff potions already in the user pinned items? For example if you should have 4, but have removed 1 from the pinned items this code will add the other 3 again?

paglias:
@negue might have some ideas about the pinned items issue

paglias:
@phillipthelen is this going to cause any issue with the mobile apps? (A new type of items added)

paglias (commented):
left some comments! Note that I haven't live-tested it

website/common/script/index.js
Xaz16: I lead by principle -- make all as private as possible... so in server code I see only reasons to set item for user and incapsulate all knowledge about it inside set function. Of course I don't familiar with the project so it might be wrong...

website/common/script/libs/getItemInfo.js
Xaz16: The main idea here that getItemInfo using get function of lodash with content and path with type params, in content we have spells.js file, where we have all this info in special key https://github.com/HabitRPG/habitica/blob/develop/website/common/script/content/spells.js#L312

I tested it locally through vue dev tools and I can show that all of this data we have on client side later (previous image was wrong)

website/common/script/libs/setDebuffPotionItem.js
Xaz16: Good note! I think I need to test it by my self and cover it with unit tests. Thank you :)

website/common/script/index.js
paglias: Yeah so that file used to be the gate to all the common code but on the client we no import only the necessary files but it's still being used in some places on the server so I'd say let's include and we'll remove it when we finally do some refactoring :)

Xaz16: Ok, sure :)

negue:
1. On the pinning of debuffPotions we could 
a) not let those potions be unpinned
or
b) add a pin on the items page for these potions


2. Also the pinned potions seem to have an image at the Rewards-Column.

3. If you have one debuff and cast another one, then two debuff-potions are pinned

Xaz16:
Hi folks!
@negue 
1. Agree with b) but it will be only market one, due to debuff potions is immediateUse items... Or is it not a good idea to extend functionality of market, because on web app we don't have such functionality for immediateUse items. I think the a) would be better but still it means that android/ ios applications should implement "not unpinned items" functionality too. @paglias what do you think about it? 
2. Sure, 
image
3. Agree, will cover my function  setdebufPotionItems with appropriate tests. But in fact due to web app logic we can't have more than 1 debuff  on user https://github.com/HabitRPG/habitica/blob/develop/website/common/script/content/spells.js#L356 (as example seafoam in cast set to false all other debufs)

website/common/script/libs/getItemInfo.js
Xaz16: Thats how we convert path and type of item to content especially for opaquePotion 
image
@paglias

Xaz16:
@paglias what do you think about @negue summary and my questions about it above? 
Could you please set labels for this PR as work or something about it?

phillipthelen:
@paglias Adding new items should not cause an issue with the apps. But it kinda depends on how exactly these are "purchased". Which API call would this use?

website/common/script/libs/setDebuffPotionItem.js
paglias: yeah some unit tests would be great!

paglias:
@phillipthelen they would come through the content api, as far as purchasing they are added to your rewards list

paglias:
Hmm @Xaz16, yesterday I had missed the fact that these are only implemented as pinned items, would it be possible to have them as normal rewards without having them appear in the market?

Xaz16:
@paglias can you elaborate please? Should it work as an heal potion (as it works now)? It mainly looks like as you wish, but it is strange thing -- if I will unpin heal potion, does it mean that I lose ability to buy them? 

Of course I will do as you said, but I need a comprehensive requirement.

paglias:
@Xaz16 sorry I wasn't very clear but to be honest I think this is a question that @Alys is more suited to answer than me.

It should work like an heal potion but that one is always available in your rewards list, it's not a pinned item. I think that this should work like that so it should also appear in the market (only if you can buy it)

Xaz16:
@paglias ok, I tested that we can't unpin health potion and will do the same thing with debuff potions. Thank you :)

@Alys should debuff potions be available in market if user have any transform buff as @paglias said?

website/common/script/libs/setDebuffPotionItem.js
Xaz16: Added

Alys:
I agree that having the the debuff item work in Rewards like the Health potion is good - i.e., you shouldn't be allowed to unpin it. (We might get a couple of complaints about that but we can easily explain that it's much better to avoid accidental unpinning.)

I don't have a strong opinion about whether the debuff item should be in the Market or not. On the one hand there's no real need for it there but on the other hand it wouldn't hurt if it was there. From memory, it wasn't in the Market on the original website. 
For code neatness, I suggest going with whatever the Market's default state is - if the debuff item appears automatically in an **appropriate** section in the Market, leave it there; if it doesn't appear, then don't create code to add it.

Xaz16:
@Alys thank you. As I understand my current code already satisfy all that you write. I just want to notice a message that we have when we try to unpin unpinable item -- 
` "cannotUnpinArmoirPotion": "The Health Potion and Enchanted Armoire cannot be unpinned."`
For some reason it is strictly bound to the especial items... What should we do in this case?
 
Could you set appropriate label please?

Alys:
This should do for all items that can't be unpinned:
"This item cannot be unpinned."

Thanks for asking!

Xaz16:
> This should do for all items that can't be unpinned:
> "This item cannot be unpinned."
> 
> Thanks for asking!

Ok, but it is a task for linguists. I'm I right or should I change translation only for en and rename key of translation to appropriate one like "cannotUnpinItem"?

Alys:
You can change just the `en` version - see the _README_FIRST.md file in the `en` directory for details.

It's probably a good idea to rename the key for clarity as you suggest. It's used in only two places so it won't be hard to update them both.

Xaz16:
@Alys, ok got it, thank you :)

Xaz16:
Hi @paglias, any plans for this PR? Should I extend or fix something?

paglias:
@Xaz16 sorry for the delay, there are no problems with the PR but I've not been able to review it in depth last week, should get to it this one

Xaz16:
@paglias got it, thank you for attention :)

website/common/script/content/spells.js
```diff
@@ -382,6 +388,7 @@ spells.special = {
     cast (user) {
       user.stats.buffs.seafoam = false;
       user.stats.gp -= 5;
+      user.pinnedItems = user.pinnedItems.filter(item => !item.path.includes('sand'));
```
paglias: can we use the exact path here? So if in future we have a new item with `sand` in its path we don't risk removing it by accident. Same for the other places

paglias (commented):
Left one comment! I'm going to put this on a testing website for a few days

website/common/script/content/spells.js
Xaz16: Sure, 
I'm not sure that understand you clearly... How can we resolve it here without using 'sand' string to filter items?

Xaz16: Now I understand, we should use full path of item here. Will be done @paglias

Xaz16: Done

paglias:
@Xaz16 merged in develop, after some final testing this week it should go live around Friday. Thanks for the PR! Notated towards your 3rd tier

paglias:
Looks like the mobile apps require an update to work with this PR so I'm going to revert it and merge it again once they're ready

Xaz16:
@paglias ok, got it. Thank you for info :)

paglias:
Reopened at https://github.com/HabitRPG/habitica/pull/11407

PR #11349, opened 2019-09-06
Fix: Edit/deleting tasks in a challenge appears possible for people who are not admin/the creator

Xaz16:
[//]: # (Note: See http://habitica.fandom.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API for more info)

[//]: # (Put Issue # here, if applicable. This will automatically close the issue if your PR is merged in)
Fixes https://github.com/HabitRPG/habitica/issues/11137

### Changes
[//]: # (Describe the changes that were made in detail here. Include pictures if necessary)

Introduce showOptions propr for column and task and pull it from challenge container to task itself to restrict editing for non leader of challenge users. By default this prop is true, so it will not hide options in place where we not pass such prop. 

[//]: # (Put User ID in here - found on the Habitica website at User Icon > Settings > API)

----
UUID: d94eca66-c6de-423c-a864-9cb103b9eb77

PR #11357, opened 2019-09-11
Refactoring & Feature: edit/delete controls of tasks

Xaz16:

Improve https://github.com/HabitRPG/habitica/pull/11349#issuecomment-529470815

### Changes

- Remove showOption from tasks props
- Pass all needed data to task for understand in mapGetter function which controls we should show
- Improve current solution with edit and delete logic

----

### DIFF
--- a/test/client/unit/specs/store/getters/tasks/canDelete.js
+++ b/test/client/unit/specs/store/getters/tasks/canDelete.js
@@ -6,7 +6,7 @@ describe('canDelete getter', () => {
 
 
     const task = {userId: 1, challenge: {id: 2}};
-    expect(store.getters['tasks:canDelete'](task)).to.equal(false);
+    expect(store.getters['tasks:canDelete'](task, 'challenge', true, null, null)).to.equal(false);
   });
 
   it('can delete broken challenge task', () => {
@@ -14,6 +14,6 @@ describe('canDelete getter', () => {
 
 
     const task = {userId: 1, challenge: {id: 2, broken: true}};
-    expect(store.getters['tasks:canDelete'](task)).to.equal(true);
+    expect(store.getters['tasks:canDelete'](task, 'challenge', true, null, null)).to.equal(true);
   });
 });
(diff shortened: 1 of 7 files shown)


Xaz16:
@paglias could you check this one as well, please? :)

paglias:
thanks @Xaz16 ! Working very well and the code is clean and well organize ! Notated towards your next tier

Xaz16:
@paglias thank you for review :) 
Doing my best

SabreCat:
FYI we had to revert this on production as it was making Challenge tasks completely uneditable from users' main task lists, which isn't correct--there are fields, such as the Notes, that users should be able to edit in their own copies of Challenge tasks.

Xaz16:
> FYI we had to revert this on production as it was making Challenge tasks completely uneditable from users' main task lists, which isn't correct--there are fields, such as the Notes, that users should be able to edit in their own copies of Challenge tasks.

@SabreCat 
Sorry for that... as I understand notes it is part of edit modal and when I checked it, it is really strange when you trying to save task and you receive "you can't edit this task" message... So should I fix it in such way?

Alys:
@Xaz16 In case it helps, the wiki's [Challenges](https://habitica.fandom.com/wiki/Challenges) page describes what elements of a challenge task can be edited by a participant - see the "Challenge Participant's Permissions" section. It also links to a similar section describing the permissions of challenge owners.

Ask here if you have questions about that.

Xaz16:
@Alys the main question still is: " when you trying to save task and you receive "you can't edit this task" message is it correct behavior?" Even for notes, which you can edit

Alys:
Not correct on the website. The website would prevent you changing any of those elements that you aren't allowed to edit - you shouldn't be able to even use the form fields for those elements. When you change any fields you are allowed to edit, the task should save as normal with no error message.

For editing the task through the API:
If you edit only elements that you're allowed to edit, the edits should save with no error.
If you edit any elements that you don't have permission to change, then your API request should return an error message.

Let me know if that's not clear!

Xaz16:
@Alys yes, it is clear. I will modify this functionality to show for user and create another pull request

PR #11113, opened 2019-04-12
WIP: Private messages overhaul

SabreCat:
[//]: # (Note: See http://habitica.fandom.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API for more info)

[//]: # (Put Issue # here, if applicable. This will automatically close the issue if your PR is merged in)

### Changes
[//]: # (Describe the changes that were made in detail here. Include pictures if necessary)

Move private messages from modal to page and fix up a bunch of style issues

[//]: # (Put User ID in here - found on the Habitica website at User Icon > Settings > API)

PR #11360, opened 2019-09-15
PMs rebuild

SabreCat:

Rebuild of #11113 

### Changes

Reconciles the style changes from #11113 with various updates to the private messages feature since then, such as the improved loading mechanisms.

Could be merged from here, or used as the basis for a project to move private messages to a page instead of a modal.

Known issues:

* There's some odd sizing going on with the modal when you switch from the default empty view to a specific conversation. If we want to deploy these changes with the modal, we'll probably want to figure that out and fix it, but it could be a waste of time if we're going to scrap the modal anyway.
* Avatars' margins vs. the conversation cards look better at some window sizes than others. I've fought with this a lot because of the weird way we position the avatar within its box. I can try futzing with it still more, or maybe someone else has a good idea for improving it?

For consideration by @negue. Do you want to pick it up from here, or shall I keep working on it?


negue:
Sure, I can take over :)

website/client/components/chat/chatMessages.vue
```diff
@@ -207,12 +189,6 @@ export default {
       this.lastOffset = container.scrollTop - (container.scrollHeight - container.clientHeight);
       // disable scroll
       container.style.overflowY = 'hidden';
-
-      const canLoadMore = this.inbox && !this.isLoading && this.canLoadMore;
```
negue: @SabreCat was this removed by intention ? or just too much refactored ^^

negue: I see the inbox-part `chatMessages` was extracted to `messageList`, currently I can't load more messages, anyway now I know where to look correctly ^^

negue:
Left todo / bugs:
- [x] fix open conversation from outside the private-messages route (like from a profile modal
- [x] fix `load more`
- [x] add avatar on an empty conversation
- [x] context menu on each conversation item
- [x] scrollbar / padding adjustments 
- [ ] no avatar when opening a new conversation

paglias:
what's the status of this one?

negue:
- added `avatar on an empty conversation` & pushing to delta 

the only issue left is that the face-avatar doesn't have all sprites / positions correctly, unless there aren't any other serious issues left, this could be fixed in a later push.

paglias:
Not my call on this, let's ask @Tressley

Tressley:
@negue — We can remove the face-avatar for now and then put it back in once we have it sorted.

negue:
Removed and pushed to delta

Tressley:
I thought this was good to go, but I noticed that new messages don't load in the window when you're on the private messages page. Syncing displays a new message notification, but trying to navigate to messages from messages doesn't work.

paglias:
@negue just fyi there's a comment from Tressley you might want to look at. I didn't notice it initially because it's an edit, so I'm just pinging you in case you missed it :)

negue:
Will check that issue! sounds a like weird "can load more" thing

negue:
Ok now the notification, header menu and refresh should work too

negue:
@paglias again this weird lint issue where even nothing is marked as error, just warnings

I hoped I already fixed it last-time #11494  but now its just back..


any ideas?

paglias:
@negue I think GH actions are cutting the logs after a certain amount of lines but there were a couple of errors, fixed in https://github.com/HabitRPG/habitica/pull/11360/commits/d1768bf982957c7ad7719583a9ea0d3f991cee64

negue:
@paglias awesome! 🎉

I tried to put the vue-cli-service to --quiet, but it seems that the cli doesn't accept it at all https://github.com/vuejs/vue-cli/issues/4560  

I also tried to use only eslint, but with that the normal errors didn't appear :/

Tressley:
Still encountering the issue with new messages not appearing in the conversation when trying to navigate to messages from the messages link in the user menu. If you sync, you receive a message notification in the user menu, but clicking that doesn't load the new message.
image

You can see in the Messages column (left) that there's a new message, but it's not propagating in the conversation (right). The only way to get the message to appear in the conversation is a hard refresh.

negue:
On a refresh, the list of conversations was reloaded, the messages of the first conversation were also loaded BUT only if there weren't any messages (for this conversation) , now it'll also force the reload on this refresh. So it should be fixed now :)

Tressley:
Last two things and I think we should be good to go:

1. Center "Messages" horizontally in the left-column header.

image

2. Prevent message preview from being cut off when hovering.

image
image

negue:
fixed those and pushed to delta

Tressley:
I think we're finally there. This LGTM. 👍

website/client/src/router/index.js
```diff
@@ -75,6 +75,8 @@ const GroupPlanIndex = () => import(/* webpackChunkName: "group-plans" */ '@/com
 const GroupPlanTaskInformation = () => import(/* webpackChunkName: "group-plans" */ '@/components/group-plans/taskInformation');
 const GroupPlanBilling = () => import(/* webpackChunkName: "group-plans" */ '@/components/group-plans/billing');
 
+const MessagesIndex = () => import(/* webpackChunkName: "group-plans" */ '@/pages/private-messages');
```
paglias: why is this part of the group plans chunk? maybe it should have its own

paglias (commented):
added a small comment but feel free to merge this when that's fixed

website/client/src/router/index.js
negue: oops, fixed

PR #11367, opened 2019-09-19
Add missing translations from transifex

phillipthelen:
This PR adds the partially translated languages from transifex. It also changes how languages are loaded. Instead of loading all languages from the directory, it only loads languages in the `approvedLanguages` list. This allows us to load the translations into weblate without having them active.


paglias:
@phillipthelen looks like the unit tests are failing because while only some languages are loaded all end up in the list of available ones

I'd also add a new test to make sure not approved languages are not loaded / added to the list of languages

phillipthelen:
@paglias I fixed the failed test. This also changed the test that checks that all translations are loaded to ensure that no extra ones are loaded. (since the `approvedLanguages` list is compared to `translations` and that would fail if there were extra ones.

paglias:
@phillipthelen thanks! It looks like the linting is failing on `i18n.test.js `

phillipthelen:
I always seem to miss _something_

paglias (approved):

PR #11368, opened 2019-09-19
load user info per conversation.id

negue:
Load user-infos after conversations - in order to get the correct userstyles

fixes https://github.com/HabitRPG/habitica-android/issues/1223

PR #11374, opened 2019-09-22
load user info per conversation.id - 2nd try

negue:
In #11368 I didn't counted in deleted users.

Now I load the user-infos of the conversations itself (2nd query)

~Todo / check~
- [x] get the "latest" user data (for avatar)  (always latest due to **$last**)

### DIFF
--- a/website/server/libs/inbox/index.js
+++ b/website/server/libs/inbox/index.js
@@ -86,12 +86,8 @@ async function usersMapByConversations (owner, users) {
       {
         $group: {
           _id: '$uuid',
-          user: {$last: '$user' },
-          username: {$last: '$username' },
           userStyles: {$last: '$userStyles'},
           contributor: {$last: '$contributor'},
-
-          count: {$sum: 1},
         },
       },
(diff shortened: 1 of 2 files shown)


website/server/libs/inbox/index.js
```diff
@@ -74,7 +74,41 @@ export async function getUserInbox (user, options = {asArray: true, page: 0, con
...
+      },
+      {
+        $group: {
+          _id: '$uuid',
+          user: {$last: '$user' },
+          username: {$last: '$username' },
+          userStyles: {$last: '$userStyles'},
+          contributor: {$last: '$contributor'},
+
+          count: {$sum: 1},
```
paglias: is this used anywhere?

paglias (commented):
I'm putting this on delta

website/server/libs/inbox/index.js
negue: no, not in the 2nd query - I removed the duplicated properties of the 2nd query

PR #11384, opened 2019-09-28
Fix Purchase Background Modal on "Edit Avatar"

negue:
This PR removes an older, now obsolete, z-index override.  (Navbars should not stay above modals, and notifications are positioned under the navbar)

Also removes my previous override (that was added because of the previous mentioned one^^)


paglias:
working well!

SabreCat:
Sending out as a hotfix because we're deploying new Backgrounds!

PR #11389, opened 2019-09-30
Add API Call to bulk score tasks

### LINKED ISSUE #11265: Level up modal (popup) doesn't always appear
@shewhoknits (e1d43f13-4a7b-4188-8844-127b4f63c67c) reports:

**Sometimes I don't get the "Level up" popup although the notification bubble on the right side shows it. By now I'm pretty sure it happens on levels 15, 30, 40 and 60 when we get an unlockable quest.**
**If I'm right, it's just a minor bug, but I thought I'd tell here anyway.**

I haven't verified this myself but I'll do some tests in the next couple of days to see if I can confirm it. I'll also look through the code to check that it's not actually a deliberate feature, for example to prevent getting too many modals at once.

### LINKED ISSUE #12248: bulk scoring should be used in RYA ("Check off any Dailies you did yesterday" screen)
Currently, when you complete Dailies in the "Check off any Dailies you did yesterday" screen (aka RYA / Record Yesterday's Dailies), each Daily is scored individually. This leads to issues such as the ones described at https://github.com/HabitRPG/habitica/issues/11058 and in its comments and the other issues it links to.

The desired approach is that when bulk scoring is available (https://github.com/HabitRPG/habitica/pull/11389), it should be used for the RYA screen. I.e., in that screen you'd tick all the Dailies that should be completed and then when you click the "Start My New Day" button, the bulk score API route scores them all, and after that cron is run.

I'm leaving this on hold until the bulk scoring PR is completed. I wanted to log this though so that we don't forget about it and so that I can explain that this fix will be done to a user currently being affected by one of the RYA issues mentioned above.

paglias:
This is the continuation of #11296 

Fixes #12248 
Fixes #11265

<details>
  <summary>Done</summary>
  
- [x] handleGroupTask is not awaited
- [x] handleGroupTask / handleChallengeTask are run at the end, with no error handling
- [x] missing unit tests
- [x] api.scoreTasks req.body is not validated
- [x] simplify code?
- [x] when to send webhook? alongside group and challenge tasks handling? also analytics
- [x] prevent task from being scored twice (also case where same task is passed once with id and once with alias)
- [x] integration tests: lacking checks on responses
- [x] what if one fails and the others don't? for example older version here https://github.com/HabitRPG/habitica/pull/12142/commits/2a504392de70932ada54da1bdce509046ef1e12e#diff-7e3afbc9a950c0819d20e0d1c440ee4aR818 - report error but do not fail everything
- [x] how to report errors in bulk scoring? extract from error handler
- [x] group task scoring with approval saves the task separately https://github.com/HabitRPG/habitica/pull/12142/commits/2a504392de70932ada54da1bdce509046ef1e12e#diff-7e3afbc9a950c0819d20e0d1c440ee4aR812
- [x] if a task scoring fails, the user is still saved ....
- [x] update api docs for api.scoreTasks
- [x] https://github.com/HabitRPG/habitica/issues/12248
- [x] loading spinner https://projects.lukehaas.me/css-loaders/ `I like first option in the first row and third option in the first row`
- [x] handle changes in approval request https://github.com/HabitRPG/habitica/pull/12142/commits/2a504392de70932ada54da1bdce509046ef1e12e#diff-b855044e08b1bf5cecdf1f99aad1e734R1086
- [x] user _tmp is cloned but scoreTasks run in parallel, switch to sequential? 
- [x] is reset every time common.ops.scoreTask is called
- [x] https://github.com/HabitRPG/habitica/blob/398b3c1f0d6ade9d27699b0e73dc24fc83ccc07c/website/server/models/user/hooks.js#L25-L36 returned as part of user.toJSON()
- [x] _tmp.leveledUp is used for webhooks https://github.com/HabitRPG/habitica/blob/398b3c1f0d6ade9d27699b0e73dc24fc83ccc07c/website/server/models/user/hooks.js#L360
- [x] sent in task scored webhook https://github.com/HabitRPG/habitica/blob/398b3c1f0d6ade9d27699b0e73dc24fc83ccc07c/website/server/libs/webhook.js#L137
- [x] approval request does not return error BREAKING CHANGE
- [x] change https://github.com/HabitRPG/habitica/pull/12142/files#diff-7e3afbc9a950c0819d20e0d1c440ee4aR787 to not throw an error anymore - tests + client handling + update api docs

client integration
- [x] sounds
- [x] group tasks https://github.com/HabitRPG/habitica/blob/develop/website/client/src/mixins/scoreTask.js#L24-L32
- [x] analytics call https://github.com/HabitRPG/habitica/blob/develop/website/client/src/mixins/scoreTask.js#L56
- [x] post request notifications https://github.com/HabitRPG/habitica/blob/develop/website/client/src/mixins/scoreTask.js#L58-L133
- [x] 1 overall notification for GP, HP from the task scoring itself
- [x] streakBonues comes from taskScore in common -> can be moved to post request notification https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/notifications.vue#L461-L470
- [x] prevent modal from reopening due to automatic close when running cron and there's a modal (for example for a login incentive)
- [x] group plans: refactor action into async resource, remove extra api call
- [x] https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/notifications.vue#L640-L646 - used here https://github.com/HabitRPG/habitica/blob/develop/website/common/script/fns/updateStats.js#L95-L98
- [x] _tmp.leveledUp is never reset on client , https://github.com/HabitRPG/habitica/pull/11389/commits/7382cb44740d3bf5e1252097ee9f80b31fc74945 possible problem? - never used
- [x] level ups, deaths, ... modals are not notifications and appear as soon as the user is synced, before the modal is closed
- [x] handle errors
</details>

TODO
- [ ] transactions?


website/server/models/task.js
```diff
@@ -192,6 +189,34 @@ TaskSchema.statics.findByIdOrAlias = async function findByIdOrAlias (
   return task;
 };
 
+TaskSchema.statics.findMultipleByIdOrAlias = async function findByIdOrAlias (
```
paglias, line 210: we should add some unit tests for this

website/server/libs/taskManager.js
```diff
@@ -284,3 +295,213 @@ export function moveTask (order, taskId, to) {
...
+
+/**
+ * Scores a task.
+ * @param user the user that is making the operation
+ * @param task The task to score
+ * @param direction "up" or "down" depending on how the task should be scored
+ *
+ * @return Response Data
+*/
+async function scoreTask (user, task, direction, req, res) {
```
paglias, line 356: now that these methods have been extracted in a library it might be a good idea to cover them with unit tests

paglias (commented):
@phillipthelen left a couple of comments about tests

paglias:
@phillipthelen what's the status of this one?

phillipthelen:
I haven't had time to work on tests for this, but will get to that next.

website/client/src/assets/scss/button.scss
```diff
@@ -5,7 +5,7 @@
   font-weight: bold;
   line-height: 1.71;
   border: 1px solid transparent;
-  padding: 0.25rem 1rem;
+  padding: 0.219rem 1rem;
```
negue, line 8: those paddings (in most of the zeplins I saw are `0.25rem 0.75rem`  (which I currently have in my PR) but which buttons need `0.219rem` ? maybe have them a special class (or apply it on the component itself instead for all buttons)

paglias:
I think I changed the padding because the buttons where slightly larger
than the designs

Il sab 22 ago 2020, 01:06 negue <notifications@github.com> ha scritto:

negue:
I had those issues as well, in the end it was a wrong line-height setting which caused that

negue:
could you tell me which buttons needed the smaller paddings? then I could take a look at them as well

paglias:
@negue I think it was all the buttons with the default styles, they were 1px too high, for example the RYA modal one is 33px vs 32px in production 
Schermata 2020-08-23 alle 18 52 30

But if the issue was in the line height we can revert the padding changes and change the line height instead

PR #11405, opened 2019-10-06
fix(chat): change class name because of active route

### LINKED ISSUE #11404: Username styling incorrect on profile links
Reported in RaB, confirmed by me:

> Aerilita   @Aerilita (3b724a41-48d3-4048-ae27-70b258cf5a72) [web] 
> Minor bug.
> 
> So I opened a profile in a new link to send a message later, and when I tried to, the recepient's name was wrong colour. Happens on both Chrome and Firefox, when accessed from a direct link or via a new tab, not happening when the profile is opened by a click from the Habitica's page (although the URL changes too).

image

Markkop:
Fixes #11404

### Changes
Renamed like `active` class name to `activeLike`, so it is not get confused with current active route class.

The bug in the issue was occurring because [Vue Router applies a class name](https://router.vuejs.org/api/#active-class) to links that are active routes. This class name is not applied when the route is activated by a link - which are 90% of the cases -, but it is when accessed from a new tab or url.
Also, by default, this active route class name is `router-link-active`, but in the Habitica's project it is replaced via the `linkActiveClass` router constructor option in `router/index.js`.

linkActiveClass

The `active` class name was being applied to the like button likebutton when the user has liked someone's post **and also when displaying a user's chat message while visiting its profile directly (new tab/direct url)**.

activeLike

The fix is changing this `active` like class name to `activeLike`, so it doens't get "triggered" by active route class name.

----
UUID: 40387571-91ee-489e-960f-278bf8fd503a

### DIFF
--- a/website/client/components/chat/chatCard.vue
+++ b/website/client/components/chat/chatCard.vue
@@ -26,10 +26,10 @@ div
         .svg-icon(v-html="icons.delete", v-once)
         div(v-once) {{$t('delete')}}
       .ml-auto.d-flex(v-b-tooltip="{title: likeTooltip(msg.likes[user._id])}", v-if='!inbox')
-        .action.d-flex.align-items-center.mr-0(@click='like()', v-if='likeCount > 0', :class='{active: msg.likes[user._id]}')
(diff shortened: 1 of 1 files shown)


paglias:
thanks for the investigation and the fix @Markkop !

paglias:
I've notated the PR towards your 4th contributor tier!

PR #11409, opened 2019-10-08
[FIX] Challenge Creation/Edit Screen Should Not Close Except By X  Button

### LINKED ISSUE #11394: challenge creation/edit screen should not close except by buttons
As reported by @caesg (e05cd0c8-038c-4606-82cf-744eea3568f6):
"If I click outside the foreground challenge box, while creating or editing the challenge, then the challenge box disappears and work up to that point is lost."

The desired behaviour is that the challenge creation / editing modal (screen) should close only if you click on the "x" button on the top right corner or the "Add Challenge Tasks" / "Update Challenge" button at the bottom of the form.

Clicking outside the modal should not close it.

_Currently, the modal can also be closed if you hit the Escape key on your keyboard. My personal opinion is that that should be prevented as well. It's convenient to use Esc if you've opened the screen by accident, but that won't happen often. It's definitely not convenient for all your work to be lost if you accidentally hit Esc while creating a challenge and I think that's the more important case to be considered. However I'm not sure what the intention is for whether Esc can be used for closing the modal, so if the staff post here to say that Esc should still close, listen to them not me._

--------------------

FYI, for challenge creators who want to avoid this bug until it's fixed, @caesg has suggested this workaround:

"For challenge creation: Type something brief/inconsequential in each * box item and save immediately.
For editing challenge: Create text in a separate word processor. Copy text and paste into challenge."

leodalmacio:
Fixes #11394

Added :no-close-on-backdrop and :no-close-on-esc on challegeModal.vue

User ID: bad1be4e-a56c-495b-9225-8dc52e916484


axrwl (approved):

paglias:
Fixed some conflicts generated by changes in the client side code

paglias:
thanks @leodalmacio ! this is working well and will be merged as soon as possible

paglias:
@leodalmacio  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11411, opened 2019-10-08
Gradually decrease PMs read count when reading instead of setting it to zero immediately

### LINKED ISSUE #11172: Visiting top level private message list should not clear all PM notifications

### Description

AnneLine in the Report a Bug Guild outlined this usability struggle:

>Not sure if this is a bug or intended behavior. Sometimes I get 3 or 4 PM messages at a time. When I come on Habitica, I see a flag saying that. Nice! But once I have gone into the box to check, there is no longer a flag for the messages, even if I only replied to one. Several times, I've not noticed I had not read nor responded to messages that way. Is it behaving as intended, or could it be made to count down the way the guild notifications do?
>
> May be just something that's an issue for me, but not for others, but here's where the problem happens. I look at one PM from the App (which triggers the turn off of the number), both that one AND any others that might be in there... but I'm not on a computer, and/or I'm only focused on doing that one thing that made me go there.... But maybe it's something that has to be done from the computer rather than the app. Or I need to fetch some info before replying. Now I might remember THAT task... but there's no longer a 'flag' for me that there might be another message or two hiding in there... and it might be a day or two longer till I get an additional message... and HOPEFULLY see other PMs and realize I din't read them and didn't respond.
>
>Could it possible to set up be something like a toggle to allow to continue to show additional messages or something? This is probably veering very close to a request rather than a bug - but it did delay a response to Alys about something for a few days.....

This workflow difficulty is worth addressing. If we kept the notification bubble live until each individual unread conversation were visited, that'd help ensure that users don't miss any notifications while processing their private messages.

ResamVi:
Fixes #11172

### Changes
I've added a field to the inbox schema to track whether a message has been read already.
If it has been read you can decrement the counter.
This required changing the markPmsRead API because the amount of read messages to decrement needs to be known.
Also needed was knowledge of the opposing user that is part of the conversation so we don't
set the "isRead" field of all messages received to true

I have a gif to showcase the change. Focus on the counter that decreases in the blurred background when clicking on a conversation.

vue change

#### Current pet peeves
It was difficult to name a variable containing the uuid of the "opposing user that is part of the conversation" but I named it "toUserUuid" if that is not confusing.

As I've noticed a message sent has 2 copies. So the one sending the message may also
mark his own message as "isRead" which has no effect whatsoever but seems weird so
I should mention it.

----
UUID: 6e6d1e09-ea96-486c-b964-b0cdfab59bd8


website/server/controllers/api-v3/user.js
```diff
@@ -1508,9 +1511,13 @@ api.markPmsRead = {
   url: '/user/mark-pms-read',
   async handler (req, res) {
```
paglias, line 1512: unfortunately this is a breaking change that would make existing calls to this route fail and we can't accept it under API v3

paglias: It would have to be moved to API v4

paglias: But then it's probably going to create issues with the mobile apps using v4, right @phillipthelen ?

paglias: The best option might be to add, in api v4, a new route to selectively mark PMs as read while keeping this one and its original behavior. Also I'm not sure having a `count` parameter is very useful as at the end we mark an entire conversation as read anyway. Instead of passing it we could count the number of messages to read directly from the database

paglias: We would also need tests for the new route

SabreCat:
Closing due to age of the PR and the pending need for design on the associated issue ticket.

PR #11455, opened 2019-10-21
Hook up label to input for dontShowAgain checkbox

cstyles:

### Changes
Now if the user clicks on the label for the checkbox, it will (un)check the associated checkbox.

----
UUID: `67b7e848-03f4-48af-b87b-6a715a729730`

### DIFF
--- a/website/client/src/components/achievements/streak.vue
+++ b/website/client/src/components/achievements/streak.vue
@@ -33,8 +33,9 @@
             v-model="user.preferences.suppressModals.streak"
             type="checkbox"
             @change="suppressModals"
+            id="user-preferences-suppressModals-streak"
           >
-          <label>{{ $t('dontShowAgain') }}</label>
+          <label for="user-preferences-suppressModals-streak">{{ $t('dontShowAgain') }}</label>
         </div>
       </div>
     </div>


website/client/src/components/achievements/streak.vue
```diff
@@ -29,12 +29,14 @@
           {{ $t('dontStop') }}
         </button>
         <div class="checkbox">
-          <input
-            v-model="user.preferences.suppressModals.streak"
-            type="checkbox"
-            @change="suppressModals"
-          >
-          <label>{{ $t('dontShowAgain') }}</label>
+          <label>
```
paglias: thanks @cstyles , I think that for accessibility reasons it would be better to keep the label and input as siblings while adding an unique `id` attribute to the input and a matching `for` attribute to the label meaning that the label is linked to the input when using a screen readeer, etc...

Putting the input inside the labels works as well but the w3 site suggests using explicit labels (https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-implicitly) and Bootstrap's styles work better with the label used as a sibling of the input (https://getbootstrap.com/docs/4.3/components/forms/)

paglias (changes requested):

website/client/src/components/achievements/streak.vue
cstyles: Yeah, I would have preferred that too but as I'm not familiar with the codebase I didn't know what to pick to avoid conflicts. Would something like `dontShowAgain` suffice?

paglias: I would probably use something that matches the user preference so in this case `user-preference-suppressModals-streak`

cstyles: OK, I just pushed the change. Let me know what you think!

paglias:
Sorry for the delay @cstyles ! This has been merged and will go live next week, I've notated the PR towards your 2nd contributor tier

PR #11460, opened 2019-10-23
 fixes #11457 - remove z-index for three-dots menu on tasks

### LINKED ISSUE #11457: z-index problems with three-dots menu on tasks
There's two problems that will probably have the same or similar fixes.

Taes / @AnArtificer (6660db98-624a-4459-ac4e-570c006c26ed) reports;
**"When opening the menu dropdown of a reward at the bottom of the rewards column, items in that section are rendered in front of the menu when it has mouse focus. As soon as I mouse away, the menu renders correctly. A refresh does not fix this issue."**

I've confirmed it (in staging as well as production).

image

---

@KGHN (98dadc6c-b4a6-4c41-9856-82db6632d98c) reports:

**"For a couple of weeks now, some habits start with the full set of edits (edit, move to top, move to bottom, delete) but then get stuck with just Edit. With a long list, I really miss the Move options. To move something from the bottom to the top, I have been opening the edit, deleting, and then re-adding it at the top, which works but isn't congenial. ... The item I'm having trouble with is a dummy marker item that I move down the list as I complete sections of my list. It starts out freshly created at the top of the list with a full set of three-dot editing options. After it has been moved several times, the three-dot options suddenly are just Edit only. Returning it to the top position does not fix the problem. ... I just now tried creating several new dummies, with text of ===, ===1, ===2, and aaa. They all had crippled edit options after being moved down individually 2 or 3 times - note that's without scrolling the list - this is movement within the first 8 items or so of a long list. Hmmm, it looks as if the other 3 editing options are hidden behind the next item down the list, like what happens intermittently with the footer hiding edit options when I ask for the options on the last item or two at the bottom of the list. ... I just noticed I can just identify a few pixels of the rest of the option list where the grey line between items normally is, below the one displaying its edit options. I think that confirms that the menu is being covered, not truncated."**

@citrusella and I have both confirmed that one (me in staging), using different browsers. Here's a screenshot of the partly-hidden three-dots menu after I moved one task down one spot:
image

tanmoyopenroot:
#### PR checklist

- [x] Addresses an existing issue: fixes #11457 
- [x] bugfix
- [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
Remove `z-index` from `hover` from class `task`

#### Before Fix:
Screenshot 2019-10-23 at 2 18 36 PM

#### After Fix:
Screenshot 2019-10-23 at 2 13 39 PM

### DIFF
--- a/website/client/src/components/tasks/task.vue
+++ b/website/client/src/components/tasks/task.vue
@@ -371,6 +371,7 @@
 
     &:hover {
       box-shadow: 0 1px 8px 0 rgba($black, 0.12), 0 4px 4px 0 rgba($black, 0.16);
+      z-index: 11;
     }
   }
 
--- a/website/client/src/components/tasks/task.vue
+++ b/website/client/src/components/tasks/task.vue
@@ -371,7 +371,6 @@
 
     &:hover {
       box-shadow: 0 1px 8px 0 rgba($black, 0.12), 0 4px 4px 0 rgba($black, 0.16);
-      z-index: 10;
     }
   }
 


paglias:
Hi @tanmoyopenroot and thanks for the PR. I noticed that the z-index was added as part of a previous PR last month to fix some groups issues https://github.com/HabitRPG/habitica/commit/01d272d2c4af3e1a5937c6963f4caab76913a414#diff-dd4708ce55cca58aefa2a330050b372cR136 so it's probably needed.

I think this can be also solved by upping the z-index on task:hover to 11, could you use that solution instead? This way we don't risk breaking any group related feature

tanmoyopenroot:
> Hi @tanmoyopenroot and thanks for the PR. I noticed that the z-index was added as part of a previous PR last month to fix some groups issues [01d272d#diff-dd4708ce55cca58aefa2a330050b372cR136](https://github.com/HabitRPG/habitica/commit/01d272d2c4af3e1a5937c6963f4caab76913a414#diff-dd4708ce55cca58aefa2a330050b372cR136) so it's probably needed.
> 
> I think this can be also solved by upping the z-index on task:hover to 11, could you use that solution instead? This way we don't risk breaking any group related feature

Sorry, I should have thought about it before removing the code and i have added back the `z-index` with a value of `11`

paglias:
no worries and thanks for the update!

paglias:
thanks @tanmoyopenroot ! This has been merged and will go live next week, if you have an Habitica account just tell me your username and I'll be happy to award a contributor tier for the PR

PR #11474, opened 2019-10-23
Minimum password length + Static Pages fixes

paglias:
Enforce a minimum password length of 8 characters.


website/client/src/components/static/home.vue
```diff
@@ -844,6 +860,14 @@ export default {
       if (this.username.length < 1) return false;
       return !this.usernameValid;
     },
+    passwordValid () {
+      if (this.password.length <= 0) return false;
+      return this.password.length >= MINIMUM_PASSWORD_LENGTH;
+    },
+    passwordInvalid () {
+      if (this.password.length <= 0) return false;
```
Alys, line 857: Is this <=0 line correct? Doesn't it mean that `passwordInvalid` will be `false` (i.e., password will be considered valid) if the length is 0 or less?

Are this line and also line 864 needed at all? I think the checks against MINIMUM_PASSWORD_LENGTH will handle all the cases.

website/client/src/components/static/home.vue
```diff
@@ -844,6 +860,14 @@ export default {
...
+      return this.password.length < MINIMUM_PASSWORD_LENGTH;
+    },
     passwordConfirmValid () {
       if (this.passwordConfirm.length <= 3) return false;
```
Alys, line 861: Should this be changed to 8 or the whole line deleted? I don't think a length check is needed here because the only important check is that the password confirm value is the same as the password value (and we're already checking length for the password).

Same for passwordConfirmInvalid below.

paglias: The <= 3 was a previous check that I think was supposed not to show the error until the user has typed at least 3 characters, which I feel like makes sense from an UX perspective but I'm not 100% sure

website/client/src/components/static/home.vue
paglias: > Is this <=0 line correct? Doesn't it mean that passwordInvalid will be false (i.e., password will be considered valid) if the length is 0 or less?

The idea here is (taken from other fields) not to show the error class until the user has typed something int to field

PR #11478, opened 2019-10-27
Use username in challenge member dropdown

### LINKED ISSUE #11456: challenge participant lists / search should use Username not Display Name
Ceran, Star Keeper / @Ceran (aabe567b-aa55-42c1-b612-b7699a7ba9b7) points out that that challenge winner selection might be easier if it used Usernames instead of Display Names. I've copied their post in full below. I'll leave this as suggestion-discussion for a few days and then if there's no objections, I'll mark it as help wanted to get changes made to <strike>the CSV export files and to</strike> the winner selection UI elements in the website.

--------------------------------------------

"When awarding challenges via the website, the drop-down menu of participants to chose the winner from shows their display names. I think it might be worth considering changing this to show the user names instead, because the display names showing causes issues with awarding:

1. If I don't get around to award the winner immediately after downloading the challenge datas and the person who won changes their display name in the meantime, I can't find it in the list anymore and have to go hunt down the person's new name, which eats up needless time.
1. If someone has a display name that consists out of non-latin letters, some special looking latin script or emojis, most of the time it can not be displayed in the list and these parts are cut out, turned into a garble or other characters or replaced by empty boxes. That makes it hard to recognize the person in the list. Often I can only find them because "since they have an unusual display name, the array of empty boxes must be them".
1. The downloadable data often also has this problem with displaying non-standard latin characters. If a challenge has a lot of participants, and the person who won does not appear right at the top of the drop-down menu for awarding, I have to type in the person's name in the box to make it show. If it consists of mentioned letters however, I am unable to type the name, and since the spreadsheet has the same problem, I can not copypaste their name from their either, nor from their user profile, since only the messed up version shows in the drop-down menu anyway. In this case I have to go over the challenge data tool to be able to award a winner.

In summary, I think all of these issues could be easily avoided if the participants list to pick the winner from was changed to show the user name instead of the display name."

adamfitzgibbon:

Fixes #11456 

### Changes
update challenge member dropdown and api to display and search by username rather than display name
Screen Shot 2019-10-27 at 1 00 32 AM

----
UUID: 2d02f021-be7b-4be4-b20e-3488a00336c8

### DIFF
--- a/test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js
+++ b/test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js
@@ -206,16 +206,16 @@ describe('GET /challenges/:challengeId/members', () => {
     for (let i = 0; i < 3; i += 1) {
       usersToGenerate.push(generateUser({
         challenges: [challenge._id],
-        'profile.name': `${i}profilename`,
+        'auth.local.username': `${i}username`,
       }));
     }
     const generatedUsers = await Promise.all(usersToGenerate);
-    const profileNames = generatedUsers.map(generatedUser => generatedUser.profile.name);
+    const usernames = generatedUsers.map(generatedUser => generatedUser.auth.local.username);
 
-    const firstProfileName = profileNames[0];
-    const nameToSearch = firstProfileName.substring(0, 4);
+    const firstUsername = usernames[0];
+    const nameToSearch = firstUsername.substring(0, 4);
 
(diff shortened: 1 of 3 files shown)


paglias:
thanks @adamfitzgibbon, this is looking good and will be merged at the end of the week or after we deploy the current `develop` branch

paglias:
@adamfitzgibbon thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11479, opened 2019-10-27
Allow update webhook label with an empty string

### LINKED ISSUE #11477: webhook label should be able to be set back to empty string
_As reported by @rhitakorrr (cab16cfa-e951-4dc3-a468-1abadc1dd109)_:

"Updating a webhook with an empty string as a label does not clear the label. Example:"
```python
>>> response = fetch("https://habitica.com/api/v3/user/webhook", requests.post, { "url": "https://example.com", "label": "Example" })
>>> response.json()["data"]["label"]
'Example'
>>> webhook_id = response.json()["data"]["id"]
>>> modified_webhook = fetch(f"https://habitica.com/api/v3/user/webhook/{webhook_id}", requests.put, { "label": "" })
>>> modified_webhook.json()["data"]["label"]
'Example'
>>> modified_webhook_2 = fetch(f"https://habitica.com/api/v3/user/webhook/{webhook_id}", requests.put, { "label": " " })
>>> modified_webhook_2.json()["data"]["label"]
' '
```

"I found the issue in the source a while back (though I don't remember exactly where now), and if I recall correctly, a conditional was treating empty string as if no value was provided. This was due to the false-y nature of empty string in JavaScript. Changing the conditional to check for `== null` would probably fix this."

_Note that the default value for the label in a new webhook is an empty string (see the dump below for a new webhook created through the habitica.com website), so it should be permissable to set the label back to an empty string._
```
    "webhooks": [
        {
            "type": "taskActivity",
            "label": "",
            "enabled": true,
            "id": "f02900eb-f401-4e24-bc65-2d5b677601f3",
            "options": {
                "created": false,
                "updated": false,
                "deleted": false,
                "checklistScored": false,
                "scored": true
            },
            "url": "http://alsdkjf.cdl/",
            "createdAt": "2019-10-27T04:22:42.497Z",
            "updatedAt": "2019-10-27T04:22:42.497Z"
        }
    ]
```

evelynstender:
Fixes #11477 

### Changes
Checking if `label` is `null` or `undefined` instead of checking if it's also an empty string + tests

image

image

----
UUID: 4c0fa3ff-440c-418a-85ee-1ad7ea998a20

### DIFF
--- a/website/server/controllers/api-v3/webhook.js
+++ b/website/server/controllers/api-v3/webhook.js
@@ -179,7 +179,7 @@ api.updateWebhook = {
       webhook.url = url;
     }
 
-    if (label) {
+    if (label !== null && label !== undefined) {
       webhook.label = label;
     }
 
--- a/website/server/controllers/api-v3/webhook.js
+++ b/website/server/controllers/api-v3/webhook.js
@@ -179,6 +179,7 @@ api.updateWebhook = {
       webhook.url = url;
     }
 
+    // using this check to allow the setting of empty labels
     if (label !== null && label !== undefined) {
       webhook.label = label;
     }
--- a/test/api/v3/integration/webhook/PUT-user_update_webhook.test.js
+++ b/test/api/v3/integration/webhook/PUT-user_update_webhook.test.js
@@ -63,6 +63,26 @@ describe('PUT /user/webhook/:id', () => {
     expect(webhook.options).to.eql(options);
   });
 
+  it('updates a webhook with empty label', async () => {
+    const url = 'http://a-new-url.com';
+    const type = 'groupChatReceived';
+    const label = '';
+    const options = { groupId: generateUUID() };
+
+    await user.put(`/user/webhook/${webhookToUpdate.id}`, {
+      url, type, options, label,
+    });
+
+    await user.sync();
+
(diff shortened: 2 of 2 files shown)


website/server/controllers/api-v3/webhook.js
```diff
@@ -179,7 +179,7 @@ api.updateWebhook = {
       webhook.url = url;
     }
 
-    if (label) {
+    if (label !== null && label !== undefined) {
```
paglias, line 183: could you add a comment here to explain that we're using this check to allow the setting of empty labels? Thanks!

paglias (changes requested):
@evelynstender thanks for the PR! I left a comment but the code is looking good and I'll merge this as soon as possible when the comment is added

website/server/controllers/api-v3/webhook.js
evelynstender: Hi! Sure, will do! :D

evelynstender: @paglias added the comment!

paglias:
Sorry for the delay @evelynstender ! This has been merged and will go live next week, I've also notated it towards your 2nd tier

PR #11481, opened 2019-10-27
Fix market dropdownmenu default text

### LINKED ISSUE #11423: dropdown menus in Market have no default text
As reported by citrusella:

"The dropdowns for sorting and grouping on the [Shops > Market] page only show the arrow. (As compared to, say, pet or equipment dropdowns showing the text and the arrow.)"

image

For an example of how it should work, see the Inventory > Items page where the "Sort By" menu has the default sort order displayed as soon as you load the page.

saraheisa:
Add text attribute to b-dropdown in filterDropdown component

Fixes #11423
### Changes

----
UUID:

### DIFF
--- a/website/client/src/components/ui/filterDropdown.vue
+++ b/website/client/src/components/ui/filterDropdown.vue
@@ -1,7 +1,9 @@
 <template>
   <span>
     <span class="dropdown-label">{{ label }}</span>
-    <b-dropdown right="right">
+    <b-dropdown
+    :text="(selectedItem.id.includes('sortBy')) ? selectedItem.id.substring(6) : selectedItem.id"
+    right="right">
       <span
         slot="text"
         :class="{'dropdown-icon-item': withIcon}"
--- a/website/client/src/components/ui/filterDropdown.vue
+++ b/website/client/src/components/ui/filterDropdown.vue
@@ -1,11 +1,9 @@
 <template>
   <span>
     <span class="dropdown-label">{{ label }}</span>
-    <b-dropdown
-    :text="(selectedItem.id.includes('sortBy')) ? selectedItem.id.substring(6) : selectedItem.id"
-    right="right">
+    <b-dropdown right="right">
       <span
-        slot="text"
+        slot="button-content"
         :class="{'dropdown-icon-item': withIcon}"
       >
         <slot


saraheisa:
solving issue #11423

website/client/src/components/ui/filterDropdown.vue
```diff
@@ -1,7 +1,9 @@
 <template>
   <span>
     <span class="dropdown-label">{{ label }}</span>
-    <b-dropdown right="right">
+    <b-dropdown
+    :text="(selectedItem.id.includes('sortBy')) ? selectedItem.id.substring(6) : selectedItem.id"
```
paglias: this doesn't seem to be using the correct string version but simply the key meaning it won't be translated correctly, seeing this instead of the correct string that would be the capitalized "Warrior"
Schermata 2019-10-31 alle 12 17 58

I mean that the default selection should match the item listed in the dropdown (with the same text and icon)

paglias (changes requested):

website/client/src/components/ui/filterDropdown.vue
saraheisa: I'm new to vue 
So would you please help me with this
How can I grab the value of the dropdown menu

I found out that "text" attribute returns the same value in the dropdown but it only works for class

paglias: I think this is something that might have broken with a recent version of bootstrap vue because the slot is there so it might be a breaking change we missed. @negue any idea? The slot exists here https://github.com/HabitRPG/habitica/pull/11481/files#diff-b7d94aa4b827758be8c8eddcd201c4e4R7 but is not rendered

website/client/src/components/ui/filterDropdown.vue
```diff
@@ -1,7 +1,9 @@
...
+    right="right">
       <span
         slot="text"
```
paglias: @prosarahgamal I would try to change `slot="text"` with `slot="button-content"` here and remove your other changes

saraheisa: @paglias Thanks a lot for your valuable comments
It works fine now

website/client/src/components/ui/filterDropdown.vue
negue: @prosarahgamal already found the correct slot name :tada:

@paglias yeah another vue-bootstrap upgrade issue I'll go through other components and see if we still have some removed properties / bindings

paglias:
thanks @prosarahgamal, if you have an Habitica UUID or username I'll be happy to award you a contribution tier

saraheisa:
> thanks @prosarahgamal, if you have an Habitica UUID or username I'll be happy to award you a contribution tier

Yeah sure!
my username @dev_sarah (ba216907-332d-461a-8e41-c1c5d295e4de)

paglias:
@prosarahgamal awarding your contributor tier :)

PR #11492, opened 2019-10-28
Show specific number of pets/mounts in button "Show More"

### LINKED ISSUE #9786: Show specific number of pets/mounts in "Show More"
Right now, it's not always obvious to users that the Show More dropdowns contain the rest of their pets and mounts. For that reason, we'd like to show the total number of pets and mounts in a given section, as follows:

new stable

We may also want to apply this to other "Show More" buttons around the site.

akmaekki:
Fixes #9786

### Changes
This fix displays the number of already owned pets (mounts) in the button "Show more". Therefore, this button would contain the text "Show more (2 Standard Pets)" in case this user already owns 2 pets.

image

To test this fix pets (mounts) should be added to the users collection. This can be done by means of
`> db.users.update({'auth.local.username': 'testusername'}, { $set: {'items.pets.Wolf-Base': 1} })`

----
UUID: 7c99d188-7c2e-45cc-9da3-130a26b63cf5

### DIFF
--- a/website/client/src/components/inventory/stable/index.vue
+++ b/website/client/src/components/inventory/stable/index.vue
@@ -187,7 +187,7 @@
           class="btn btn-flat btn-show-more"
           @click="setShowMore(petGroup.key)"
         >
-          {{ $_openedItemRows_isToggled(petGroup.key) ? $t('showLess') : $t('showMore') }}
+          {{ $_openedItemRows_isToggled(petGroup.key) ? $t('showLess') : $t('showMore') + getTextOwnedAnimals(petGroup, $t('pets')) }}
         </div>
       </div>
       <h2>
@@ -246,7 +246,7 @@
           class="btn btn-flat btn-show-more"
           @click="setShowMore(mountGroup.key)"
         >
-          {{ $_openedItemRows_isToggled(mountGroup.key) ? $t('showLess') : $t('showMore') }}
+          {{ $_openedItemRows_isToggled(mountGroup.key) ? $t('showLess') : $t('showMore') + getTextOwnedAnimals(mountGroup, $t('mounts')) }}
         </div>
       </div>
       <inventoryDrawer>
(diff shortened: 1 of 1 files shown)


website/client/src/components/inventory/stable/index.vue
```diff
@@ -246,7 +246,8 @@
           class="btn btn-flat btn-show-more"
           @click="setShowMore(mountGroup.key)"
         >
-          {{ $_openedItemRows_isToggled(mountGroup.key) ? $t('showLess') : $t('showMore') }}
+          <!--{{ $_openedItemRows_isToggled(mountGroup.key) ? $t('showLess') : $t('showMore') }}-->
```
paglias: can this comment be removed?

website/client/src/components/inventory/stable/index.vue
```diff
@@ -778,6 +779,16 @@ export default {
 
       return animals;
     },
+    getTextOwnedAnimals (petGroup, pets) {
+      return ` (${this.getCountOwnedAnimals(petGroup, 'pet')} ${petGroup.label} ${pets})`;
+    },
+    getCountOwnedAnimals (animalGroup, type) {
```
paglias, line 784: Instead of introducing a new computed property can we reuse this new one inside `countOwnedAnimals` so the code isn't duplicated? Thanks

paglias (changes requested):
thanks for the PR! I left a couple of comments

website/client/src/components/inventory/stable/index.vue
akmaekki: Yes, sorry. Removed the comment.

website/client/src/components/inventory/stable/index.vue
akmaekki: Ok, incorporated changes!

Btw, is the name `getCountOwnedAnimals` ok, or should it better be `getCountOwnedPets`?

paglias:
@SabreCat given the perf problems we've had with the Stable, what do you think about this?

@akmaekki sorry for the delay!

paglias (approved):

paglias:
I previously added some comments thinking this was using computed properties but turns out it's using methods and it might require some extra perf tweaks

paglias:
Sorry about this @akmaekki but we're experiencing some perf problems with the stable right now and this would add even more load. The way to solve this would be to refactor the page to use computed properties instead of methods to get the number of pets.

If you want to try it, but it might not be the easiest thing ever, let me know and I'll be happy to help with any question or problem

SabreCat:
Given the age of this PR and the lack of response to the conversation on the associated issue ticket, closing this as abandoned! Feel free to reopen if you come back to it, @akmaekki.

PR #11494, opened 2019-10-29
Storybook Example

negue:
Example again from scratch, fully working with the countBadge-component, including '@/assets style import' and added the normal scss too.

I think extracting it to a separate folder won't work (or atleast will have multiple complications), like the webpack.config, right now it just uses the current vue-config without adding any settings to point to it.

We'd have to see how much longer the install now takes with storybook in it.


website/client/package.json
```diff
@@ -52,5 +54,13 @@
     "vuedraggable": "^2.23.1",
     "vuejs-datepicker": "git://github.com/habitrpg/vuejs-datepicker.git#5d237615463a84a23dd6f3f77c6ab577d68593ec",
     "webpack": "^4.41.2"
+  },
+  "devDependencies": {
```
paglias: can you move these deps to normal ones instead of devdeps? Otherwise it's not possible to build storybook when deploying because they're not installed in prod

paglias (changes requested):

website/client/package.json
negue: moved to normal dependencies

negue:
Can't figure out what else is broken on lint, all I see there are warnings

paglias:
can you try to merge develop in here to see if the error goes away?

negue:
Nope, mergin from master not helping either, all lint messages are warnings, are now warnings handled as errors?

paglias:
Warnings are not handled as errors but from the logs it looks like for some reasons the lint process stall with these changes

negue:
I ran the lints again, added dist and node_modules as ignores (without it, lint is running for 20mins), hopefully it'll work now

PR #11499, opened 2019-11-01
Permit changing the name of a party without adding a description

### LINKED ISSUE #10828: Cannot change name of party without adding a description
Reported in Report a Bug:

> Birdskull (051409ee-97be-42ae-9916-dd27fb22f113) 
> It's impossible to change the name of a party without also adding a description.

Confirmed by me.

It seems to be desirable that people should be able to change their party name without adding a description; parties aren't public and descriptions aren't that important.

_edit by Alys 2020-01-19:_

This is the desired fix for this issue:

- When you edit a Party, the Description field must be optional - you should not be forced to enter a Description before you can click the "Update Party" button.
- When you edit a Guild, there should be no change to the current behaviour: a Description is mandatory (i.e., it should not be possible to delete the Description then click the "Update Guild" button).
- When you are editing a Party, you should not see an asterisk in the Description label.
- You should still see that asterisk when you are editing a Guild.
- When you are editing either a Party or a Guild, you should not see an asterisk in the "Privacy Settings" label. It's not mandatory to tick the checkbox there so that field shouldn't be marked as mandatory. (This isn't strictly related to this issue but it's a trivial change - just delete the asterisk for all cases.)

akmaekki:
Fixes #10828

### Changes
Until now a description had to be added when changing the name of a party. Now the party name can be changed without also changing its description.

As the picture below shows, the Button is enabled even though the description was not changed.

image

UUID: 7c99d188-7c2e-45cc-9da3-130a26b63cf5

### DIFF
--- a/website/client/src/components/groups/groupFormModal.vue
+++ b/website/client/src/components/groups/groupFormModal.vue
@@ -254,7 +254,7 @@ label.custom-control-label(v-once) {{ $t('allowGuildInvitationsFromNonMembers') 
         <button
           v-if="workingGroup.id"
           class="btn btn-primary btn-md"
-          :disabled="!workingGroup.name || !workingGroup.description"
+          :disabled="!workingGroup.name"
         >
           {{ isParty ? $t('updateParty') : $t('updateGuild') }}
         </button>


akmaekki:
PR ready for review

website/client/src/components/groups/groupFormModal.vue
```diff
@@ -540,7 +540,6 @@ export default {
       if (!this.workingGroup.name) errors.push(this.$t('nameRequired'));
       if (!this.workingGroup.summary) errors.push(this.$t('summaryRequired'));
       if (this.workingGroup.summary.length > MAX_SUMMARY_SIZE_FOR_GUILDS) errors.push(this.$t('summaryTooLong'));
-      if (!this.workingGroup.description) errors.push(this.$t('descriptionRequired'));
```
paglias: thanks @akmaekki can you revert this line? I'm saying this because while the error is not used anymore when updating parties it's still used when creating one, see line https://github.com/HabitRPG/habitica/pull/11499/files#diff-5b256091a52b128af86f73ea9ec9e4e0R250

paglias (changes requested):

website/client/src/components/groups/groupFormModal.vue
akmaekki: @paglias 
Easy to revert, but for me it seems illogical that a description has to be specified when a party is created but can be empty when party name changes.

paglias: agreed but that modal is also used by guilds where description is mandatory

akmaekki: Ah, I see. Thanks for the explanation.

paglias:
Sorry @akmaekki but looks like the change I requested is not working because now you get an error if you don't specify the description in a party. 

I feel like this need some extra refactoring:

- the button should be enabled based on the presence of anything inside the `errors` array, not based on a separate property
- when checking for valid date before submitting (https://github.com/HabitRPG/habitica/pull/11499/files#diff-5b256091a52b128af86f73ea9ec9e4e0L540) there should be an exception when we're in a party for the missing description so that the condition is still enforced when working with guilds

Alys:
@akmaekki How are you going with this? Is it something you'd still like to work on? If it is, great, we'd still appreciate this fix!

If you've found that real life has got in the way and you no longer have time for this, just let us know - we know what that's like :) - and then we'll open up the issue for another contributor to take up. 

If you would like to keep going with it, can you please make another small change so that the asterisk (*) doesn't appear in the Description's label when you're editing a party? It should still appear when you're editing a Guild.

Also, this isn't directly related but it's a trivial fix to the same screen: Could you remove the asterisk from the "Privacy Settings" label for both Parties and Guilds? It's not mandatory to tick the checkbox there so that field shouldn't be marked as mandatory.

I'm about to add these two new points to the issue, so if you'd rather not keep working on this issue that's fine - the next contributor will see my notes there.

akmaekki:
Hi,

Ok, I cannot continue working on this at the moment. Maybe if I have free hours, I'll look into it, but the ticket should be revoked from me.

Alys:
@akmaekki No worries, thanks very much for letting us know!

PR #11502, opened 2019-11-01
Prevent duplicate challenge tasks

### LINKED ISSUE #11295: duplicate challenge tasks in participant's task list
[EDIT: See the first comment below for an important update.]

@SuperSaraA (dbf5390d-6a57-42ad-ae14-af8fd8e570ca) has reported a recurring problem with duplicate challenge tasks appearing on the website. In the most recent case at least the duplicates later disappeared by themselves (unless @sabrecat cleaned them up and didn't post about it?). All the details we have so far are below.

For the moment, I'm marking this as needs investigation. If it happens again and I can capture details of the tasks before they disappear, I'll update this issue.

@SuperSaraA: "a continuation of a problem that I have had more than once. I am somehow in one challenge more than one time. I have no idea how this keeps happening to me. Lucky for me there are not any daily tasks to this challenge. Is there any way that you could please get rid of all the extra copies of the to-do tasks from the Celebrate August challenge (I think I counted that there are about 7 or 8 extra copies of all the tasks)? If this is something you can't do, I guess I can just leave the challenge and re-sign up."

@shanaqui: "have we ever logged this as a bug? I'll tag in @Sabrecat to take a look at cleaning this up for you"

@SuperSaraA [roughly a day after the initial report]: "The extra tasks are now gone. I wonder if it happens when I am trying to join a challenge with a not so great internet connection."

@Alys: "Were those extra tasks visible on the Android app or website or both?"

@SuperSaraA: "I could see all the extra to dos on the website but I'm not sure about on the android app. I remember thinking my to do list was getting way too long because the little scroll line on the right side was so tiny (and I was starting to forget the stuff at the top by the time I would get to the bottom of the list) . So I want onto the website version to look over my tasks in more detail and that is when I for sure saw all the extra copies of the tasks. Sorry, I can't remember more! If it happens again, I will be sure to take notes of if I can see it on which platform."

cvuorinen:

Fixes #11295

### Changes
Created a separate method `addToUser` in challenge model that handles adding the challenge to users challenges atomically (with a condition that checks that it is not there already) to prevent multiple concurrent requests from passing through. See issue for more details (#11295).

Also renamed `syncToUser` method to `syncTasksToUser` since now that adding the challenge to user is not there anymore, it's main purpose is to sync the tasks.

----

### DIFF
--- a/migrations/challenges/sync-all-challenges.js
+++ b/migrations/challenges/sync-all-challenges.js
@@ -10,7 +10,7 @@ async function syncChallengeToMembers (challenges) {
 
     const promises = [];
     users.forEach(user => {
-      promises.push(challenge.syncToUser(user));
+      promises.push(challenge.syncTasksToUser(user));
       promises.push(challenge.save());
       promises.push(user.save());
     });
--- a/test/api/unit/models/challenge.test.js
+++ b/test/api/unit/models/challenge.test.js
@@ -120,7 +120,7 @@ describe('Challenge Model', () => {
         expect(updatedNewMember.challenges.length).to.eql(1);
       });
 
-      it('syncs a challenge to a user', async () => {
+      it('syncs challenge tasks to a user', async () => {
         await challenge.addTasks([task]);
 
         const newMember = new User({
@@ -128,7 +128,7 @@ describe('Challenge Model', () => {
         });
         await newMember.save();
 
-        await challenge.syncToUser(newMember);
+        await challenge.syncTasksToUser(newMember);
 
         const updatedNewMember = await User.findById(newMember._id);
         const updatedNewMemberTasks = await Tasks.Task.find({ _id: { $in: updatedNewMember.tasksOrder[`${taskType}s`] } });
@@ -146,7 +146,7 @@ describe('Challenge Model', () => {
         expect(syncedTask.attribute).to.eql('str');
       });
 
-      it('syncs a challenge to a user with the existing task', async () => {
+      it('syncs challenge tasks to a user with the existing task', async () => {
         await challenge.addTasks([task]);
 
         let updatedLeader = await User.findOne({ _id: leader._id });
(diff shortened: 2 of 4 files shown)


cvuorinen:
It's a bit... "challenging" to develop since I can't run the tests in my local install.
Always just get this error:
```
      at listOnTimeout (internal/timers.js:531:17)
      at processTimers (internal/timers.js:475:7)
```

I am using Docker and I have it otherwise working (actual website works in browser and e.g. `npm run lint` works), but running `npm run test:api:unit` or `npm run test:api-v3:integration` gives that error.

Anyone know how to fix that?

paglias:
@cvuorinen have you tried running only the test file you're interested with using `.only` on the `describe` call? Also make sure `TEST_DB_URI` is set as an env variable. Is that the only output you get?

cvuorinen:
@paglias yes, I've tried with `.only` but doesn't help. I'll check out the `TEST_DB_URI` next.

cvuorinen:
But seems the tests passed anyway :laughing:

SabreCat:
Except for `lint`, it turns out! Looks like some simple fixes needed per https://github.com/HabitRPG/habitica/pull/11502/checks?check_run_id=285137134

cvuorinen:
Oh, and by the way. I have a local setup of the Android app that's connected to my local Habitica install, so I tested it with that too and it seems to fix the issue. Although it does display a notification of something like "you are already in the challenge" for the requests that get blocked by this change. But I guess that's ok, and it could be fixed in the app so it does not allow multiple concurrent requests.

paglias:
thanks @cvuorinen and sorry for the delay. Given the sensitive change I'm going to put this through some extra testing but if nothing comes up we'll merge this next week

Alys:
@cvuorinen :
> ... Android app... does display a notification of something like "you are already in the challenge" for the requests that get blocked by this change. But I guess that's ok, and it could be fixed in the app so it does not allow multiple concurrent requests.

I agree that's okay. After the fix for the issue is merged, I'll create a low-priority issue in the Android repo to suggest a change to prevent multiple requests.

I've marked this as "notify helpers" as a reminder to me that I need to do something later. :)

paglias:
Haven't forgot about this one but given it's Thanksgiving's week we're not merging anything important this week :)

paglias:
thanks @cvuorinen ! Notated towards your 6th tier

PR #11505, opened 2019-11-03
Fixes #11297 (Orb of Rebirth remains non-free an extra day)

### LINKED ISSUE #11297: Orb of Rebirth remains non-free an extra day (creates misleading message on website)
When the Orb of Rebirth stopped being free consistently from level 100, it was announced in the Aspiring Socialites guild with this message (emphasis is mine):
> "Hey Socialites! We deployed a tweak with this Gala that power users might run into. Previously, there was a sort of exploit where folks could use Fix Character Values + the Orb of Rebirth to evade having to pay Gems to change classes. Now, the freebie level 100+ Rebirth is only **usable once every 45 days**. Users who are affected by the limitation will see a small orange banner when they open the Orb from the Market saying how many days are left in this countdown."

However with the current code, it actually becomes free again after 46 days, not 45. I realise this is pretty minor but it would be easy to fix, and it's causing a misleading display message on the website (described below).

The current API code is this: https://github.com/HabitRPG/habitica/blob/2bfc41b1c3174db869209d6e4d74950f6920d222/website/common/script/libs/isFreeRebirth.js#L4-L14

It could be changed to match the official announcement by changing `> 45` to `>= 45`

---

The website's behaviour matches the API (i.e., it shows the Orb as free or not free according to the decision made by the API), but because the API makes you wait one extra day, the website's message is confusing on the final day (the day when it should already be free again) - on that day the website says "0 days until FREE Orb of Rebirth", as reported by @Evil_Overmom (0396e6e1-0edc-46f5-9449-24891ba16698):
 
> "Hi -- I'm having a problem with the orb of rebirth. It tells me I have 0 days til I get a free orb, but no option to get the free orb.
Selection_001

If the API's code is fixed as described above, then the Orb will become available a day earlier and you'll never see that misleading message.

FYI this is the website's code but it does not need to be changed if the API change above is made:
https://github.com/HabitRPG/habitica/blob/2bfc41b1c3174db869209d6e4d74950f6920d222/website/client/components/shops/buyModal.vue#L382-L384
https://github.com/HabitRPG/habitica/blob/2bfc41b1c3174db869209d6e4d74950f6920d222/website/client/components/shops/buyModal.vue#L82-L85

mistergij:

Fixes #11297 

### Changes
Updates Orb of Rebirth API to match website code.

----
UUID: 
b762e018-642b-4b78-a735-354a52c0a905

### DIFF
--- a/website/common/script/libs/isFreeRebirth.js
+++ b/website/common/script/libs/isFreeRebirth.js
@@ -10,5 +10,5 @@ export default function isFreeRebirth (user) {
     daysFromLastFreeRebirth = 999;
   }
 
-  return user.stats.lvl >= MAX_LEVEL && daysFromLastFreeRebirth > 45;
+  return user.stats.lvl >= MAX_LEVEL && daysFromLastFreeRebirth >= 45;
 }


paglias:
@mistergij  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11507, opened 2019-11-04
Improvements and fixes for push notifications

phillipthelen:
This contains an improvement for push notifications with user generated content by adding a `unformattedText` field, that contains the chat message without markdown formatting and using that for the push notification.

It also adds the `type` field to the chat mention push notification payload, so that the apps can correctly open the right screens in the app when a user opens the notification.


paglias:
@phillipthelen sorry for not looking at this earlier, it's ok but I think we should really add some tests showing that the unformatted text is always added correctly.

If it's important for mobile apps for this to go out soon and you don't have time to add the tests (I'm in the middle of some achievements related changes and don't have much time) i'll add a reminder to add them later and this can be merged (ping @SabreCat if it needs to go out sooner than next week)

paglias:
looks like there are a couple of lint errors, then for me it's good to go

Issue #6169, opened 2015-11-03
limited time purchasable items (e.g., Magic Hatching Potions, Grand Gala equipment) can't be bought at the end of the end of the event


Alys:
_EDIT 2019-11-03: See https://github.com/HabitRPG/habitica/issues/6169#issuecomment-549111718 for the latest status. -- Alys_

_EDIT 2019-11-09: The comments just below that one for a method to replicate the bug and a suggested unconfirmed fix. -- Alys_

**EDIT 2020-01-09: The staff have approved the fix described in this comment: https://github.com/HabitRPG/habitica/issues/6169#issuecomment-552068039 -- Alys**

In the final day or so of any Grand Gala, some players find that when they buy the special class gear and other limited edition items, the items disappear after a sync.

I believe this is because of timezone conflicts between the user's browser and the Habitica server. If that's true, then we should be able to fix it by using the user's value of `preferences.timezoneOffset` (e.g., -600 for a user in the GMT+10 timezone). I.e., the code that decides whether to display the items in the rewards column should consider the user's timezone as well as the date range of the availability period so that the items are not visible to the user after the server knows that the date range is over.

The code should assume a zone of GMT (0) if the user doesn't have a value for timezoneOffset.

---
### What to do if this bug has happened to you:

**If you've purchased seasonal items and they have disappeared, please do NOT comment on this issue. This issue is only for discussing the code required for a permanent fix. To help us focus on a fix for this, we might delete any comments here that aren't directly related to fixing it.**

**Please describe your missing items on https://github.com/HabitRPG/habitrpg/issues/6644 and we'll give them back to you.**

TheHollidayInn:
When you say the items disappear, do you mean they disappear from the user's inventory, the rewards column or the time shop? Or a mix?

Also, are you able to replicate this in any way?

SabreCat:
@TheHollidayInn: They disappear from the Rewards column--or, even worse, they sometimes _don't_ immediately disappear from the Rewards column, but when you buy them, they vanish from inventory and refund the Gold upon next sync.

Replicating it should be possible, if a little tricky. This would do it:
1. Change the `event` date range of a seasonal set in the `content` file so that the item is active through the current date
2. Run your Habitica server using a time zone that's ahead of the one you'll connect to with your client
3. Log in with the above mismatched zones, and keep the app open across the divide where the server reaches a new day outside of the event date range 
4. Attempt to buy the seasonal items while the client is still on a day within the stated range of availability

[...]

PR #11519, opened 2019-11-06
Award mystery items at cron

SabreCat:

### Changes

**Before**, mystery items were only ever awarded via migration or during the subscription process.
**Now**, if a user crons during the availability period of a mystery set, the mystery items will be awarded at that point.

This will remove the need for a manual migration and give us more flexibility in scheduling mystery item rollouts.


paglias (approved):

Alys:
So if I understand correctly, if a user with an active subscription doesn't cron during the month that a mystery set is available, they won't ever be given that set? If so, I'll make sure the wiki docs are updated for that when this goes live. (We'd also need to remember this when doing assessments of whether a user is missing Hourglasses, since they might have spent Hourglasses on items that were available during their subscription.)

If I'm wrong, just a simple "no" will suffice. :) No need to spend time explaining more.

SabreCat:
@Alys Hmm, that's a good point. We could maybe backfill missed months based on the subscription's `updatedDate` vs. the cron date?

Alys:
I'm not sure if the `updatedDate` is always reliable for that.

What about this scenario (it might be unlikely but not impossible):
1. User signs up for a 12-month subscription in January.
2. User stops using Habitica in March - `updatedDate` will be some time in
March.
3. A second user gives the first user a gift subscription in June (perhaps
a friend trying to entice them back) - `updatedDate` will then be some time
in June I THINK (I might be wrong here [EDIT: nope, see comment below]).
4. User logs in again in August. The backfill uses the June `updatedDate`
so the user ends up with Mystery Items for June, July, August, but not for
April or May.

Maybe. :)

Alys:
There *might* also be odd interactions between `updatedDate` and non-recurring subscriptions that end while the user is inactive (or recurring ones that are force-terminated by us due to payment problems) and/or with extraMonth amounts that need to be applied and/or with gift subscriptions given after a terminated subscription has ended.

Alys:
I've done some testing and this is not a problem - I was wrong:

> 3. A second user gives the first user a gift subscription in June (perhaps
> a friend trying to entice them back) - `updatedDate` will then be some time
> in June I THINK (I might be wrong here).

I.e., if you give a gift subscription to an inactive player who already has a subscription, then `dateUpdated` is not changed by the gift subscription, so that scenario isn't a barrier to backfilling missed months based on the subscription's updatedDate vs. the cron date.

I still have a vague sense of unease about it though, in case there's odd situations where it might fail, especially since it's possible to perform some actions in Habitica without cronning if you use the API or third party tools, or if cron just won't run for you. For example, if you've had a long absence and then start using Habitica on a device that has a different time zone than your previous device (quite possible due to DST), cron will be delayed for a day (the standard protection to prevent early cron after a time zone change) and during that day you might use Hourglasses to buy Mystery Sets that would otherwise have been given to you for free as soon as cron did run.

SabreCat:
@Alys That requires rather a lot of things to go wrong in very specific ways, so if that's the one edge case where a user might overpay or need to reach out to admin@habitica for help that sounds pretty good. I'll put together some code for backfilling and see if we can get the more common situations covered!

Alys:
Something that _might_ need to be considered (definitely not a show-stopper and I might be wrong anyway)... 

For people in time zones ahead of UTC (like me at UTC+10), there might be an annoying timing issue that might stop the Mystery Items being given for a month if cron runs right at the start of the month. For example, if I cron at the start of 1 December in my time, it's still 30 November in UTC / server time. The grantEndOfTheMonthPerks() function would run [I _think_] and my `updatedDate` value would be changed to the new month, BUT if the code for comparing my current time with the Mystery Item `start` date doesn't consider time zones, then I might be too early to get the items, and since `updatedDate` is changed, I wouldn't get them on the future days that month either. I think. :)
https://github.com/HabitRPG/habitica/blob/99599953f55766fb447707594fe1c66a78d9e89d/website/server/libs/payments/subscriptions.js#L30

It's sort of the opposite of this bug: https://github.com/HabitRPG/habitica/issues/6169#issuecomment-552058556 and might need the .utc() method to be used (if I'm even right about whether that method would help with that other bug).

Worth noting that I've been neck-deep in time-related code for a couple of hours now and it's been doing my head in so I might be mistaken. :) I just have a sort of gut-level maybe-it's-a-problem feeling and I wanted to note it down before it went out of my head. Take it with a grain of salt. :)

In contrast to all that, it's possible that grantEndOfTheMonthPerks() doesn't actually run for me until the 2nd of the month in my time zone, in which case this definitely isn't a problem for Mystery Items. It's been too long since I paid attention to exactly when my consecutive month count ticks over.

tl;dr Time zones are evil.

SabreCat:
@Alys Interesting. Again a pretty strange case, but definitely less strange than the prior scenario. That said, I thought `grantEndOfTheMonthPerks` used server time? I'll see what I can puzzle out.

Alys:
Cantras has mentioned that this string would need to be adjusted for the new timing ("... This is usually released about one week before the end of the month...")

https://github.com/HabitRPG/habitica/blob/7d78a7b3201edd410834c5bd41bb48b88483371d/website/common/locales/en/generic.json#L70

website/common/script/content/mystery-sets.js
```diff
@@ -304,6 +25,17 @@ const mysterySets = {
     end: '2014-04-01',
   },
 };
+const FIRST_MYSTERY_SET = moment('2014-02-01');
+
+for (let mysteryMonth = FIRST_MYSTERY_SET;
```
paglias: I got slightly confused by the indentation here :P I wasn't exactly sure what was going on, could we change it to something like:

```
for (
    let mysteryMonth = FIRST_MYSTERY_SET;
    ...
) {
...
}
```

?

paglias (commented):
LGTM, just left one small comment

website/common/script/content/mystery-sets.js
SabreCat: Heh, lint's indentation rules are not flawless for ensuring readability 😸 Will fix!

PR #11539, opened 2019-11-13
Avatar Edit Modal: addtional none options

negue:
Added a none option to:
- glasses
- animal ears / tails
- headbands


paglias:
Not sure if it's related but I'm getting this error in the console `Error processing the common message "missingTypeKeyEquip"`

paglias:
Yeah it happen when you select the new none items. 

Another issue is that you can only select the new none options only after having selected another item previously (and it's not marked as selected when you don't have any accessory of the category selected)

negue:
Should be fixed now :)

paglias:
LGTM!

PR #11542, opened 2019-11-14
update members list and count when joining or leaving challenge

### LINKED ISSUE #11537: challenge participant count and list should update when joining / leaving
When you join a challenge, the displayed number of participants does not increase immediately, and if you click on the Participants button/icon as soon as you've joined, you don't see yourself in the participant list.
Similarly, when you leave a challenge, the number and participant list are not updated immediately.
After joining / leaving, you need to reload the page to see the number and list change.

The ideal behaviour would be that as soon as you click the join or leave button, the number and list of participants should change immediately.

Note that this is just a client-side issue. When you join or leave a challenge, in the database the challenge's `memberCount` and your participant status are correctly updated, so server/API changes are not needed for this.

_Thanks to @DarkPassenger28 for reporting this._

adamfitzgibbon:

Fixes #11537 

### Changes

Update the count and members list when leaving/joining a challenge on the front-end so that it is immediately reflected in the UI without having to reload data.

----
UUID: 2d02f021-be7b-4be4-b20e-3488a00336c8

### DIFF
--- a/website/client/src/components/challenges/challengeDetail.vue
+++ b/website/client/src/components/challenges/challengeDetail.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="row">
     <challenge-modal @updatedChallenge="updatedChallenge" />
-    <leave-challenge-modal :challenge-id="challenge._id" />
+    <leave-challenge-modal :challenge-id="challenge._id" @update-challenge="updateChallenge" />
     <close-challenge-modal
       :members="members"
       :challenge-id="challenge._id"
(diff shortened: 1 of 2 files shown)


paglias:
thanks for the PR @adamfitzgibbon but I think it would be better if instead of changing client side data directly (duplicating the logic) we use the response from the join or leave API call and use that data on the client side. This also avoids the issue of pushing the entire user object (`this.user`) inside the list of members (which include just some user fields)

adamfitzgibbon:
Yeah this change was motivated by the ticket saying that only client side logic needs to be updated. That’s a simple enough change though. I’ll look into those requests and see if they return the updated logic

adamfitzgibbon:
@paglias I updated the code to use the response on join. I noticed the leave api call didn't return a response so rather than updating that api call I opted to reload the challenge via a separate api request.

paglias:
thanks @adamfitzgibbon ! this is working well, I've merged it and noted the PR towards your 3rd tier

PR #11543, opened 2019-11-15
Feature 11522 group plan tags

### LINKED ISSUE #11522: Group Plan tags are not deletable after leaving group
When you're in a guild/party that has a Group Plan, a tag for the group is added to your personal tags (this is correct behaviour).

When you are no longer in that group, the tag remains and you are not able to edit or delete it. I'm not sure if this happens when you leave the group or when the leader removes you or both.

The correct behaviour is that the tag should become one of your personal tags. I.e., the `group` attribute should be removed from the tag (see the sample tag below which shows that attribute).

The tag should not be deleted automatically because the player might have used it to tag their own tasks, which is permissible.

This was reported by @Ceran (aabe567b-aa55-42c1-b612-b7699a7ba9b7):
_"I still have the tag from the group I used to be in. I don't see a way for me to remove this tag. Could this tag be removed, please? "The Groupish Group". Only group tag I have."_
I've confirmed they are no longer in the associated group and have deleted the tag for them. This is what the tag used to be:

```
    "tags": [
        {
            "id": "d2397cb7-e92d-42dc-aba8-6686cab105e3",
            "name": "The Groupish Group",
            "group": "d2397cb7-e92d-42dc-aba8-6686cab105e3"
        },
 ```

For any coder wanting to work on this, [Using Your Local Install to Modify Habitica%27s Website and API](https://habitica.fandom.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API) shows you how to add a Group Plan to a guild or party (in the "Creating Test Data" section). So, in your local install, you can create a guild or party as normal with one test user account as leader, run that database command to convert it to a group plan, invite a second test user to it, then use that second user for testing how tags behave on leaving / being removed. At least in theory. :) Post here if that doesn't work (say what goes wrong) and if you need help working out how to replicate this bug.

nikonhub:

Fixes #11522

### Changes
Client : 
- Allow edit names of all tags
- Allow remove all tags except the current party one

Server :
- Do not force keep all group tags on user tags update. Only keep the current one.

----


paglias:
Hi @Nikosmonaut , thanks for the PR! Is this PR ready? In that case can you change its status from "draft"? thanks

PR #11546, opened 2019-11-17
Update rebirth purchase modal text to reference group plans (fixes #8834)

### LINKED ISSUE #8834: Habit counters should be set to zero by Orb of Rebirth (not for challenges or group plans)
When you use the Orb of Rebirth, it should set the counters for all your own Habits to zero. This will be consistent with how the streaks of Dailies are reset to zero.

Habit streaks should NOT be reset to zero for any Habits that are part of active Challenges or are from Group Plans (also consistent with Rebirth's affects on Dailies).

_EDIT 2019-11-16: The two points above might already be done. The coder who woks on this might want to test Rebirth on their local install before starting to investigate the code._

The `rebirthPop` text string should be adjusted to match:

https://github.com/HabitRPG/habitica/blob/020726f6f83030cc087dac27679eb17b0ce544f6/website/common/locales/en/rebirth.json#L24

Currently it says:
"... Your streaks will be removed except from challenge tasks. ..."
That should be changed to something like this:
"... Your streaks will be removed except from tasks belonging to active Challenges and Group Plans. ..."
Note that "Challenges" and "Group Plans" should be capitalised as shown.

The `rebirthAdvList3` string also mentions streaks but it's not currently used so it doesn't need to be changed.

_Thanks to cTheDragons for noticing this._

citrusella:

Fixes #8834 

### Changes
The behavior the issue asks for (habit counters resetting to zero except for challenge and group plan tasks) appears to already happen when rebirth happens, so this just makes the requested change to the related string (rebirthPop) using the wording used in the issue's first post (that is "challenge tasks" was changed to "tasks belonging to active Challenges and Group Plans".)

----

### DIFF
--- a/website/common/locales/en/rebirth.json
+++ b/website/common/locales/en/rebirth.json
@@ -21,7 +21,7 @@
     "rebirthOrb": "Used an Orb of Rebirth to start over after attaining Level <%= level %>.",
     "rebirthOrb100": "Used an Orb of Rebirth to start over after attaining Level 100 or higher.",
     "rebirthOrbNoLevel": "Used an Orb of Rebirth to start over.",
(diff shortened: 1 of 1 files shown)


paglias:
thanks @citrusella , noted in your profile :)

PR #11548, opened 2019-11-17
remain on party invites tab after removing invite - fixes #11523

### LINKED ISSUE #11523: party/group leader should remain on invitations tab after removing an invitation
As reported by @MaybeSteveRogers:

"When viewing the guild member list on web, there's a tab that takes you to 'invitations'. Removing the invitation from any one user results in the modal "tabbing" back to the 'members' list. I'm not sure if it's intentional this way, but it *is* annoying when you're trying to clear all the invitations to have to spend time going back to the 'invitations' tab every single time."

The desired behaviour is that  you remain on the invitations tab after removing an invitation.

<strike>My guess would be that it's okay for this to be changed so that you remain on the invitations tab after removing an invitation, however I'll leave it as suggestion-discussion for a few days. If there's no disagreements from the staff, I'll then mark it as help wanted.</strike>

alicemliu:
Fixes #11523 

### Changes
When invite is removed from party:
if there are other invites pending, stays on invites tab.
if no other invites are pending, switch back to members tab (since invites tab disappears).

note: did not write tests for change, but willing to add if necessary!

f8773841-dd31-466e-bce9-e412190f414c

----
UUID:

### DIFF
--- a/website/client/src/components/groups/membersModal.vue
+++ b/website/client/src/components/groups/membersModal.vue
@@ -656,7 +656,6 @@ export default {
         groupId: this.groupId,
       });
       if (this.invites.length === 0) this.viewMembers();
-      else this.viewInvites();
     },
     async promoteToLeader (member) {
       const groupData = { ...this.group };
--- a/website/client/src/components/groups/membersModal.vue
+++ b/website/client/src/components/groups/membersModal.vue
@@ -655,7 +655,8 @@ export default {
         memberId: member._id,
         groupId: this.groupId,
       });
-      this.viewMembers();
+      if (this.invites.length === 0) this.viewMembers();
+      else this.viewInvites();
     },
     async promoteToLeader (member) {
       const groupData = { ...this.group };


website/client/src/components/groups/membersModal.vue
```diff
@@ -655,7 +655,8 @@ export default {
         memberId: member._id,
         groupId: this.groupId,
       });
-      this.viewMembers();
+      if (this.invites.length === 0) this.viewMembers();
+      else this.viewInvites();
```
paglias: is the else needed? I mean if there is at least one invite and viewMembers is not called it shouldn't switch at all?

paglias (commented):
left one comment @alicemliu !

website/client/src/components/groups/membersModal.vue
alicemliu: agreed, it is redundant! I've tested to make sure and updated my pull request to reflect that change. Thank you!

alicemliu (commented):
updated pull request to remove line 659, so as to avoid redundant call.

paglias:
@alicemliu  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11549, opened 2019-11-17
Fix can not remove Group Plan tag

### LINKED ISSUE #11522: Group Plan tags are not deletable after leaving group

nikonhub:
- Allow to change name and remove a group tag
- Add a rule to check if the group tag is an active one to hide the
trash icon

Fixes #11522

### Changes
image

Client : 
- Allow to edit names of all tags
- Allow to remove all tags except if the user belongs to it. (guild or party group plan subscription)

Server :
- api/v4/user update tags. Change rules of tags that will be kept. Do not by default keep all group tags but check if it belongs to an active user guild or party instead.
- When a group leader assigns a new task to a member, do not reset the tag name of this member.

----

### DIFF
--- a/test/api/unit/models/group.test.js
+++ b/test/api/unit/models/group.test.js
@@ -1164,6 +1164,23 @@ describe('Group Model', () => {
         });
       });
 
+      it('unlink group tag', async () => {
+        participatingMember.tags.push({
+          name: party.name,
+          id: party._id,
+          group: party._id,
+        });
+
+        await participatingMember.save();
+        await party.leave(participatingMember);
+
+        participatingMember = await User.findOne({ _id: participatingMember._id });
+        const groupTag = participatingMember.tags.find(tag => tag.id === party._id);
+
+        expect(groupTag).to.not.be.undefined;
+        expect(groupTag.group).to.be.undefined;
+      });
+
       it('deletes a private party when the last member leaves', async () => {
         await party.leave(participatingMember);
         await party.leave(sleepingParticipatingMember);
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -1345,6 +1345,8 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all', keepC
     .map(task => this.unlinkTask(task, user, keep, false));
   await Promise.all(assignedTasksToRemoveUserFrom);
 
+  this.unlinkTags(user);
+
   // the user could be modified by calls to `unlinkTask` for challenge and group tasks
   // it has not been saved before to avoid multiple saves in parallel
   const promises = user.isModified() ? [user.save()] : [];
@@ -1405,6 +1407,15 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all', keepC
   promises.push(group.update(update).exec());
 
   return Promise.all(promises);
+};
+
+schema.methods.unlinkTags = function unlinkTags (user) {
+  const group = this;
+  user.tags.forEach(tag => {
+    if (tag.group && tag.group === group._id) {
+      tag.group = undefined;
+    }
+  });
 };
 
 /**


website/server/libs/user/index.js
```diff
@@ -126,39 +126,34 @@ export async function update (req, res, { isV3 = false }) {
...
         }
-      });
 
-      user.tags = oldTags;
-
-      val.forEach(t => {
-        const oldI = removedTagsIds.findIndex(id => id === t.id);
-        if (oldI > -1) {
-          removedTagsIds.splice(oldI, 1);
+        if (user.party._id && user.party._id === tag.group) {
```
paglias: this seems to be only checking the user's party but you can also be part of a group that is not your party (see `user.groups`)

paglias (changes requested):
left one comment but this is also going to need some integration / unit tests to make sure the API is working correctly, thanks @Nikosmonaut !

website/server/libs/user/index.js
nikonhub: Do you mean user.guilds ? If yes I check it just after at line 140.

nikonhub:
> this is also going to need some integration / unit tests to make sure the API is 

I added 2 integrations tests. No sure about about unit ones though.

website/server/libs/user/index.js
paglias: sorry I missed that

paglias (changes requested):
@Nikosmonaut sorry for the delay. I gave a better look at the PR and I think it's a bit overly complicated. 

Instead of changing the way the tags are updated I think it would be better to remove the `group` attribute from your tag when you leave or are removed from a group, like the original issue said

```
The correct behaviour is that the tag should become one of your personal tags. I.e., the group attribute should be removed from the tag (see the sample tag below which shows that attribute).
```

This way the changes to the client side won't be necessary. The code executed when you leave a group is here https://github.com/HabitRPG/habitica/blob/develop/website/server/models/group.js#L1315

nikonhub:
Hello.

No problem for the delay. (But yes it kind of gets old in my head the reasons for this solution).
Like I said [here](https://github.com/HabitRPG/habitica/issues/11522#issuecomment-552503278) we may face another unwanted behavior if the user is reinvited twice to the same group. Because he's "personal" old group tag would be reused.

```
    "tags": [
        {
            "id": "d2397cb7-e92d-42dc-aba8-6686cab105e3",
            "name": "The Groupish Group",
            "group": "d2397cb7-e92d-42dc-aba8-6686cab105e3"
        },
```
I think the perfect solution would be to not have the same id. Tag id !== group. So we could just delete the group key and we're done. But it would need to change the whole database :/

If you're ok with this side effect of reusing the same tag, I revert my changes to just remove the group key.

Edit : 
It is this code I'm talking about.
Maybe another solution would be to findIndex by { group: group._id }. From now have id !== group and keep old tags as is. But I it seems confusing.

```
  const userTags = user.tags;
  const i = _.findIndex(userTags, { id: group._id });

  if (i === -1) {
    userTags.push({
      id: group._id,
      name: group.name,
      group: group._id,
    });
  }
```

paglias:
> I think the perfect solution would be to not have the same id. Tag id !== group. So we could just delete the group key and we're done. But it would need to change the whole database :/

> If you're ok with this side effect of reusing the same tag, I revert my changes to just remove the group key.

Good point. I'd say I'm okay with reusing the same id but let's see if @Alys is okay with that. If you don't hear anything in the next few days feel free to reuse the id

Alys:
@Nikosmonaut I'm okay with reusing the same id too. Thanks for thinking through the options and implications!

paglias:
Thanks @Nikosmonaut and sorry for the delay! The code fix looks good but it would need some tests, if you want to add them great but if you have any problem let me know and I can add them :)

nikonhub:
Hello @paglias.

Just added a test.
Sorry for the delay. Christmas holidays not obvious to find time.

paglias:
Hi @Nikosmonaut  , sorry for the delay but I'm still catching up from the end of year holidays, I hope to get to this PR in the next couple of days

paglias:
It took a while but I finally managed to merge this, thanks @Nikosmonaut ! I've notated this PR towards your 3rd tier

PR #11563, opened 2019-11-19
when buying new equipment, item does not disappear from Market or Seasonal Shop

### LINKED ISSUE #10920: when buying new equipment, item does not disappear from Market or Seasonal Shop

Xaz16:

Fixes https://github.com/HabitRPG/habitica/issues/10920

### Changes

I'm not sure in current solution, due to I mainly working with angular, but what I know about vue change detection (CD) is that we should reassign whole object for trigger CD of vue. 
As approve of this solution the doc of vue reactivity: https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats

Last paragraph:

> Sometimes you may want to assign a number of properties to an existing object, for example using Object.assign() or _.extend(). However, new properties added to the object will not trigger changes. In such cases, create a fresh object with properties from both the original object and the mixin object.

In fact after my changes I can't reproduce the issue, so I think I hit the issue...

----


Xaz16:
Hello!
@paglias any comments about this PR? :)

paglias:
sorry for the delay, I think the fix is okay in that it was probably broken by a server side change we made a few months ago to improve performances that removed the default value for some paths in the user model and Vue 2 (Will be fixed in Vue 3) doesn't track changes to new properties of objects. 

I'm going to leave a comment on the code and will review this in a few days because I want to see if we can get a fix that's only client side since backend code isn't really affected

website/common/script/ops/pinnedGearUtils.js
```diff
@@ -95,7 +95,10 @@ export function removePinnedGearAddPossibleNewOnes (user, itemPath, newItemKey)
   // remove the old pinned gear items and add the new gear back
   removePinnedGearByClass(user);
 
-  user.items.gear.owned[newItemKey] = true;
+  const oldOwnedUserItems = { ...user.items.gear.owned };
```
paglias: Is this one necessary?

I think, but I'm not sure that it would work with one statement instead of two, the important is that the object is completely reassigned:

```

  const newOwnedUserItems = { ...user.items.gear.owned, [newItemKey]: true };
  user.items.gear.owned = newOwnedUserItems; 
```

paglias (commented):
A couple of things that I think are related to this PR and I just noticed:

- A similar change should be applied every time a new gear is awarded, not just here?
- There's a similar bug in the stable with new pets

website/common/script/ops/pinnedGearUtils.js
Xaz16: yeah, it will work in that way to, just for better and comprehensive code I create another not necessary (maybe just for now) variable. I will remove it.

Xaz16:
> A couple of things that I think are related to this PR and I just noticed:
> 
> * A similar change should be applied every time a new gear is awarded, not just here?
> * There's a similar bug in the stable with new pets

Yes, we should update all properties clearly for Vue reactivity, otherwise we can see such trouble that is not updated app state

Xaz16:
@paglias 
I started to move some other code to my solution but noticed such function _markModified_, which we have in function which I change in this PR. But for some reason user.markModified is undefined, thats why no changes applied on client (maybe). I see it everywhere but I can't find any source of it. Could you please explain what is it and why we don't have it all the time?

It means that app have broken change detection in some places but it hard to understand it due to if statement only check function is defined in user object and don't give any warnings if not...

paglias:
@Xaz16 sorry for the delay but I was away this weekend. The `markModified` function is only available on the server to tell Mongoose that some schema less objects in a database document have been changed and need to be saved, that's the reason why it's undefined on the client.

So when used in `common` code we check if it's available first:

```
if (doc.markModified) doc.markModified(path);
```

Xaz16:
> @Xaz16 sorry for the delay but I was away this weekend. The `markModified` function is only available on the server to tell Mongoose that some schema less objects in a database document have been changed and need to be saved, that's the reason why it's undefined on the client.
> 
> So when used in `common` code we check if it's available first:
> 
> ```
> if (doc.markModified) doc.markModified(path);
> ```

I suggest to make in issue to refactor all such code... Due to it really error prone if using on client.
I will try to make it

paglias:
@Xaz16 We used to have the `user.markModified` function point to an empty function on the client so that it could be used without issues but we decided to change it as it was used without knowledge of what it was doing or not doing. What's your idea for refactoring such code?

Xaz16:
@paglias Hm, in fact it is not a refactoring from my side. More like make it a bit harder but ensure triggering from Vue reactivity... But it is a lot of work to replace all assignments to replacing with new object and I'm not really sure that it should be applied to every such place. I suggest to be aware of it and fix problems when needed... But in fact it is a not good choice due to all code base will be in such state where we are really not sure if we triggered update from Vue

paglias:
Ah I thought you were talking about the `markModified` function. Regarding the Vue updates I agree. What happened is that we changed the server side schema to avoid some performances problems but that resulted in these problems.

It's happening for user's keys that are not defined when the user is returned from the API, in particular it should affect setting new keys for user's items https://github.com/HabitRPG/habitica/blob/develop/website/server/models/user/schema.js#L278 so `items.gear.owned`, `items.pets`, `items.eggs`, `items.mounts`, `items.hatchingPotions`, `items.food` and `items.quests` if a new key is added on the client

Xaz16:
> Ah I thought you were talking about the `markModified` function. Regarding the Vue updates I agree. What happened is that we changed the server side schema to avoid some performances problems but that resulted in these problems.
> 
> It's happening for user's keys that are not defined when the user is returned from the API, in particular it should affect setting new keys for user's items https://github.com/HabitRPG/habitica/blob/develop/website/server/models/user/schema.js#L278 so `items.gear.owned`, `items.pets`, `items.eggs`, `items.mounts`, `items.hatchingPotions`, `items.food` and `items.quests` if a new key is added on the client

Ok, so when we determinate the problem it makes really obvious which places we should fix. I will create apart pull request based on this comment soon.

paglias:
thanks @Xaz16 !

paglias:
I'm going to close this in favor of #11637

PR #11564, opened 2019-11-20
Force reward value to 0 when empty field

### LINKED ISSUE #11560: saving a Reward with no value / cost should not be possible
As reported by Togohogo1 / @yiybtheumes (9bb5ff83-dac5-450e-89a9-213ef54c97d6) and confirmed by @citrusella:

"if you set your custom rewards amount to no value and click on it, the site freezes, and reloading it does not unfreeze it. After opening it again in a new tab, your gold value for the reward is set to the value that you last assigned it."

The desired behaviour is that you should not be able to save a Reward if it does not have a value ("cost"). Note that a value of zero IS allowed! The situation that needs to be avoided is having no value in the cost field.

nikonhub:

Fixes #11560 

### Changes
When a user edits or adds a new reward. Force the value to be equal to 0 if the field is empty.

----

### DIFF
--- a/website/client/src/components/tasks/taskModal.vue
+++ b/website/client/src/components/tasks/taskModal.vue
@@ -1437,6 +1437,10 @@ export default {
         this.task.group.sharedCompletion = this.sharedCompletion;
       }
 
+      if (this.task.type === 'reward' && this.task.value === '') {
+        this.task.value = 0;
+      }
+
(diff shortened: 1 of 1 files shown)


paglias:
@Nikosmonaut  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11591, opened 2019-11-27
Fix edit challenge twice bug

### LINKED ISSUE #11561: editing Challenge twice gives "string not found" categories and prevents saving
If you edit a challenge and save it, and then immediately edit it again, the second edit has the following problems:
- the assigned categories stop being displayed at the bottom of the edit screen and instead you see "string * not found" (screenshot below)
- the "Update Challenge" button cannot be pressed meaning that your edits cannot be saved

I don't know if those two problems are related.

If you reload your browser in between the two edits, the problems don't occur.

The desired behaviour is that a second edit should work correctly without a reload being needed. The "string * not found" text should not appear.

This was reported by @BlueSky-9 (910410cb-71e0-47dd-a9b4-faa3fbe9c99c) who has seen it happen twice. I replicated it, although in one case I had to edit three times in a row before the problems happened.

image

nikonhub:
Description: Use of a same object property group for differents types between a
challenge and workingChallenge
workingChallenge use group as an id but this.challenge have a group
object

Fix: Wait for server response for challenge edit. Set this.challenge
from response and not directly from form data workingChallenge

Fixes #11561

### Changes
After clicking on "Update Challenge" button wait for server response to set this response as this.challenge. WorkingChallenge is just an image of form data. Should not be used as is.

PS : Like this comment says it is error prone
image

Maybe group should be groupID instead.

----

### DIFF
--- a/website/client/src/components/challenges/challengeModal.vue
+++ b/website/client/src/components/challenges/challengeModal.vue
@@ -595,7 +595,7 @@ export default {
       this.$root.$emit('bv::hide::modal', 'challenge-modal');
       this.$router.push(`/challenges/${challenge._id}`);
     },
-    updateChallenge () {
+    async updateChallenge () {
       const categoryKeys = this.workingChallenge.categories;
       const serverCategories = [];
       categoryKeys.forEach(key => {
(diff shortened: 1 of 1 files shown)


paglias:
thanks @Nikosmonaut ! I've noted this PR towards your 2nd tier

PR #11594, opened 2019-12-01
APIDoc update for Tasks: fixes #10943

### LINKED ISSUE #10943: apidocs errors / omissions
There's some problems with the API documentation at https://habitica.com/apidoc/ as reported by Dan Dulaney @duplaja (801998d9-f208-4357-8d2a-c792f23059a1):

- [ ] "I noticed an API documentation bug, found here (possibly the same item in other sublinks): https://habitica.com/apidoc/#api-Task-CreateUserTasks  Under frequency, for allowed values, it lists only "daily" and "weekly", but it looks like "monthly" and "yearly" are also options on site."
- [ ] "The documentation also says that everyX only applies to frequency=daily, but it looks like you can now set this for weekly? (and I assume but haven't checked, monthly and yearly)"
- [ ] "It looks like monthly type has another parameter it receives that's undocumented, telling whether the task should be checked on day of the month, or day of the week."

I haven't verified these so if the person who works on this has questions or needs clarification, comment here.

SunSparc:
As posted in issue 10943, there were some fields relating to adding/updating tasks that had not yet been updated.

Fixes #10943

### Changes
Adding `monthly` and `yearly` to the `frequency` parameter.
Adjusting `frequency` description to include `daysOfMonth` and `weeksOfMonth`, and to clarify that  `everyX` can be used by every value.
Adding entries for the missing fields: `daysOfMonth` and `weeksOfMonth`.

----
UUID: 639ec312-3ddb-4c11-a903-4914b72efeb9

### DIFF
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -76,7 +76,7 @@ const requiredGroupFields = '_id leader tasksOrder name';
  *                                         include: a UUID, startDate and time.
  *                                         For example {"id":"ed427623-9a69-4aac-9852-13deb9c190c3",
  *                                         "startDate":"1/16/17","time":"1/16/17" }
- * @apiParam (Body) {String="weekly","daily"} [frequency=weekly] Value "weekly" enables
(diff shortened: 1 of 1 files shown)


paglias:
Thanks @SunSparc ! I've notated this towards your 2nd tier

PR #11595, opened 2019-12-01
Modified the sort for name to be case-insensitive, fixes #11547

### LINKED ISSUE #11547: sorting in member/participant lists should use case folding for Parties, Guilds, Challenges
As reported by @GeminiForce (bec264b9-5776-47fa-8cb6-672d33ea8fc2): 

"On the desktop, when I sort my party alphabetically by [Display Name], it will sort first alphabetically by names with the first letter capitalized and then by names with the first letter not capitalized. So that the names are listed like A, B, C, D, a, b, c, d."

The desired behaviour is that sorting should use [locale-sensitive case folding](https://www.w3.org/International/wiki/Case_folding).

The same should apply to sorting in the member lists for Guilds and in the participant lists for Challenges.

NB some Display Names start with emoji or other special characters. I don't have any particular expectations for how sorting would handle them so I think whatever position they happen to fall into will be fine - I don't think we need to have specific code to handle them.

moniwang:

[//]: #11547 
Fixes #11547 

### Changes
Before, when sorting members by name, the result would be according to case (A, B, C, a, b, c). This is changed to a technique where the case is not sensitive (a, A, b, B, c, C) as well as maintaining the ability to compare in different languages (using localCompare).

43f0b0bd-c910-4453-bd41-d8f38a7adc59

----
UUID:

### DIFF
--- a/website/client/src/components/groups/membersModal.vue
+++ b/website/client/src/components/groups/membersModal.vue
@@ -480,6 +480,8 @@ export default {
       if (!isEmpty(this.sortOption)) {
         // Use the memberlist filtered by searchTerm
         if (this.sortOption.value === 'profile.name') {
+          // If members are to be sorted by name, use localeCompare for case-
+          // insensitive sort
           sortedMembers.sort(
             (a, b) => a.profile.name.localeCompare(b.profile.name),
           );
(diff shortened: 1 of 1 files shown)


moniwang:
Force push resulted from accidentally pushing changes from another pull request on this branch.

paglias:
@moniwang  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11614, opened 2019-12-04
Fixed disappearing email form bug by forcing a refresh after submission, Fixes #11565 

### LINKED ISSUE #11565: Email edit form in Settings misbehaves after submission

### Description

On the Settings page, if you change your account's email address, the form fields vanish after you click Submit--and are replaced with the "Add email to account" fields. A sync or refresh corrects it, but it's causing some confusion for users. Folks think either that the email reg was _removed,_ or they try to fill out the newly appeared fields, resulting in errors due to the chosen email address already being in existence.

The page should update immediately to show the new email, or invalidate the user session and require a new login, not switch modes as if the email were no longer present.

MeisterRed:

Fixes #11565 

### Changes
This pull request resolves the issue involving the user changing the email address for their account. When the user filled out the form and clicked submit to change their email address, it originally updated the email for the account but also removed the "Change Email Address" and "Change Password" forms from the settings (acting as if the account didn't have an email). This caused confusion for users since they didn't know if the email for their account was actually updated until they refreshed the page.

The solution I created was to force the page to refresh after the user fills out the form and clicks submit. This will avoid causing confusion for the user

----
UUID: 7d427213-9991-49fb-8513-77e1456da77e

### DIFF
--- a/website/client/src/components/settings/site.vue
+++ b/website/client/src/components/settings/site.vue
@@ -870,7 +870,7 @@ export default {
       this.soundIndex = (this.soundIndex + 1) % sounds.length;
     },
     reloadPage () {
-    window.location.reload();
+      window.location.reload();
     },
   },
 };
--- a/website/client/src/components/settings/site.vue
+++ b/website/client/src/components/settings/site.vue
@@ -812,8 +812,8 @@ export default {
         this.localAuth.username = this.user.auth.local.username;
         this.user.flags.verifiedUsername = true;
       } else if (attribute === 'email') {
-        this.user.auth.local.email = updates[attribute];
-        window.location.reload();
+        this.user.auth.local.email = updates.newEmail;
+        window.alert(this.$t('emailSuccess'));
       }
     },
     async changeDisplayName (newName) {
--- a/website/client/src/components/settings/site.vue
+++ b/website/client/src/components/settings/site.vue
@@ -443,7 +443,7 @@
             <button
               class="btn btn-primary"
               type="submit"
-              @click="changeUser('email', emailUpdates); reloadPage();"
+              @click="changeUser('email', emailUpdates)"
             >
               {{ $t('submit') }}
             </button>
@@ -813,6 +813,7 @@ export default {
         this.user.flags.verifiedUsername = true;
       } else if (attribute === 'email') {
         this.user.auth.local.email = updates[attribute];
+        reloadPage();
       }
     },
     async changeDisplayName (newName) {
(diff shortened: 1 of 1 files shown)


paglias:
Thanks @MeisterRed ! There is a problem in the code because the `changeUser` function is an async one so `reloadPage` will be called immediately but the request to the server might not yet have completed which might cause problems.

Also, would you be open to try a fix where the page is not fully reloaded but instead the buttons don't disappear? Thanks

MeisterRed:
Alright, I'll see if I can make a few modifications so the refresh happens after the request is sent to the server.

I originally tried to find a fix that doesn't involve refreshing the whole page but I couldn't think of a proper way of preventing the form from disappearing. if you wish, I can try my best and try a fix within those requirements but I can't promise anything.

paglias:
> I originally tried to find a fix that doesn't involve refreshing the whole page but I couldn't think of a proper way of preventing the form from disappearing. if you wish, I can try my best and try a fix within those requirements but I can't promise anything.

No worries, if you want to try that would be great and I'm happy to answer any question or doubt you might have but if you don't think it's possible the full refresh will work for now

MeisterRed:
Thank you very much and sorry for for the late response. I'll fix up the code soon

MeisterRed:
Alright, I adjusted the code a bit by adding window.location.reload inside of the changeUser function instead of having it run side by side with the changeUser function. I'm hoping that this will allow the email to update before the page refreshes entirely. 

If there's anything that needs to be fixed or adjusted, please let me know

paglias:
@MeisterRed so i looked at the code and while this fix works it would be great to avoid having to reload the whole page.

The reason this bug is happening is that this line https://github.com/HabitRPG/habitica/blob/0935c6e28edca4caed8054cb04b919f81e8816ab/website/client/src/components/settings/site.vue#L815 replaces the email value (used to decide if the `Change Email` part is displayed) to the result of the response from the server.

But the server doesn't return an attribute named `email` but `newEmail` so the email is set to undefined and the form disappears.

It can be fixed by changing that line to 

```
      } else if (attribute === 'email') {
        this.user.auth.local.email = updates.newEmail;
```

Let me know if you have any question

MeisterRed:
I've made the changes you have suggested and that solved the bug without having to refresh the page. I also added a popup window that tells the user that the email has been successfully changed so they're not left hanging after clicking submit.
image

paglias:
@MeisterRed  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11619, opened 2019-12-08
WIP enabling add challenge button after selecting new group

### LINKED ISSUE #11545: Add Challenge Tasks button should be enabled after selecting new group
As reported by @citrusella and confirmed by me:

"I inadvertently tried to create a challenge in a guild that has challenges limited to the guild leader. That brought an error and disabled the creation (add tasks) button, so I tried to switch it to a few other guilds and the button didn't reenable. It appears only a full reload reenables the button, even if you've set it after to a guild you should be able to put the challenge in.

The desired behaviour is that as soon as you select a different group from the "Add to" drop-down menu, the "Add Challenge Tasks" button should immediately be usable.

alicemliu:
partial fix #11545 

### Changes

The pull request is a WIP for a client side solution to the issue. The conditional should also be checking if the group is not a "only leaders can add challenges" type (but I've encountered some issues getting that field).

The original issue was the button to add a challenge task would not reenable after attempting to add a task to a group which the user is not authorized for (e.g. is not the leader for a group that only allows the leader to create challenges).

I wanted to get some feedback on if this issue might be better resolved if the list of groups that a user could select would already be filtered to those which they have authorization to (hence the client side solution). I noticed the check for this currently occurs on the server side in the API (and throws an error when not authorized), so the existing design makes it difficult to adjust UI changes for a particular error.

Any guidance on this method or if there's a better approach would be appreciated!

[//]: #f8773841-dd31-466e-bce9-e412190f414c

----
UUID:

### DIFF
--- a/website/client/src/components/challenges/challengeModal.vue
+++ b/website/client/src/components/challenges/challengeModal.vue
@@ -71,7 +71,7 @@
           <option v-for="group in authorizedCreateChallenge"
             :key="group._id"
             :value="group._id"
-            >
+           >
             {{ group.name }}
           </option>
         </select>
@@ -534,10 +534,8 @@ export default {
       if (!this.workingChallenge.summary) errors.push(this.$t('summaryRequired'));
       if (this.workingChallenge.summary.length > MAX_SUMMARY_SIZE_FOR_CHALLENGES) errors.push(this.$t('summaryTooLong'));
       if (!this.workingChallenge.description) errors.push(this.$t('descriptionRequired'));
-
       if (!this.workingChallenge.group) errors.push(this.$t('locationRequired'));
       if (!this.workingChallenge.categories || this.workingChallenge.categories.length === 0) errors.push(this.$t('categoiresRequired'));
-
       if (errors.length > 0) {
         window.alert(errors.join('\n'));
         this.loading = false;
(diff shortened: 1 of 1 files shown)


Alys:
@alicemliu Is this a fix for a specific issue? If so, can you please edit this PR's description and write at the top of it `fixes #12345` (where 12345 is the issue number)? That will give us a link to the issue and will also let GitHub automatically close the issue if this PR is merged.

alicemliu:
updated PR description!

paglias:
@alicemliu sorry for the delay, I may be wrong but this is going to prevent users in guilds where everyone is allowed to create challenges from creating them? Since the UI now only allows you to choose a challenge where you are the leader

paglias:
I think the problem that the issue wanted to point out is that if you try to create a challenge where you can't the button is, rightly so, disabled but it shouldn't stay disabled when you switch to another group

paglias:
I think the problem that the issue wanted to point out is that if you try to create a challenge where you can't the button is, rightly so, disabled but it shouldn't stay disabled when you switch to another group

paglias:
I think the issue comes from this line https://github.com/HabitRPG/habitica/pull/11619/files#diff-a73635db46b2d5f7643520a5e8ebcbf1L523 setting `loading` to `true` that disables the creation button and it never being set again to `false` because it's stopped before by the permission error

SabreCat:
Looks like the source issue was addressed in later code while we were awaiting an update from the contributor. Closing!

PR #11618, opened 2019-12-08
fix(challengeItem.vue): guild name hidden caused by the fixed height

ShupingHe:
Fixes #11500

Changes:
remove the fixed height of the `.meta-info` container

Issue #11500, opened 2019-11-01
Guild name hidden by tags on challenge page


shanaqui:
Maybe Steve Rogers @MaybeSteveRogers (767e5d92-0e13-4e30-acb1-d8bba62824fc): "(Long) guild name is hidden under tags on the challenge page - take a look:"

![image](https://user-images.githubusercontent.com/1495809/68064302-86680900-fd65-11e9-81a2-5960e2ac4d58.png)

Confirmed in Google Chrome.

ShupingHe:
I'd like to work on this issue

Alys:
@Stevens1995 great, thanks! Please go ahead.

Issue #11259, opened 2019-07-09
Challenges UI layout/formatting on My / Discover pages can be wrong; Username needed


Alys:
The screenshots below show some examples of Challenges on the My Challenges and Discover Challenges pages where the formatting isn't quite right (e.g., categories overwriting the Guild name, inconsistent wrapping of the Created By line).

The layout issues might be related to some or all of these factors:
- creator name longer than average
- creator name longer than average AND the challenge is hosted in a guild (as opposed to being a "public challenge" with no guild name shown)
- size of gem prize 
- length of title
- other things?

Also the Created By line should probably contain the Username as well as or instead of the Display Name. Display Names can change often and it can be difficult to use them to recognise a player.
If it was only the Username, any issues with player name length would then probably disappear.

In addition to that, it might be worth considering changing the label "Created By" to something like "Owned By" or "Challenge Owner" or "Challenge Leader" for accuracy in all cases. It's not too uncommon for a challenge to be moved by an admin from the person who created it to another person who will then run the challenge, or for a challenge created by one person to be directly copied by another (I've seen this done with the creator's approval). In those cases "Created By" isn't accurate and it gives the credit for the imagination and hard work of thinking up the tasks to the wrong person. Using a different label would avoid that problem, while still making it clear who's currently responsible for the challenge.

I'm marking this as "on hold - needs design" in case @Tressley wants to make any design-related comments or point to an existing layout design that mightn't have been implemented as desired for all cases.

![1](https://user-images.githubusercontent.com/1495809/60875117-9589ef00-a27c-11e9-8679-e5b5a2cd7202.png)

![image](https://user-images.githubusercontent.com/1495809/60875188-b0f4fa00-a27c-11e9-893a-b4bf895b49d9.png)

![image](https://user-images.githubusercontent.com/1495809/60875242-c5d18d80-a27c-11e9-9993-2d9b33f46f9a.png)

Thanks to @ieahleen and @citrusella for reporting the layout issues.

Tressley:
Going to loop @negue in as he was the last one to touch the Challenges styles.

Alys:
See also https://github.com/HabitRPG/habitica/issues/11500 for the specific case of the Guild name being hidden by tags. It's been marked with the 'sprint' label for a fix to be done by a group of coders doing a November sprint, so while that issue is open, no one should try to fix the Guild name / label problem as part of this issue. 
If that other issue ends up being closed with a fix, I'll update this issue to remove mention of that specific problem.

[...]

PR #11620, opened 2019-12-08
fix challenge layout 

negue:
This fixes issues on smaller devices

### DIFF
--- a/website/client/src/components/challenges/challengeItem.vue
+++ b/website/client/src/components/challenges/challengeItem.vue
@@ -295,8 +295,9 @@
       justify-content: space-evenly;
       background-color: $gray-700;
       text-align: center;
-      padding: 16px;
+      padding: 8px;
       border-radius: .25em;
+      flex-wrap: wrap;
 
       > div {
         .value {
@@ -327,6 +328,8 @@
       }
 
       > div.muted {
+        margin: 8px;
+
         .value {
           opacity: 0.5;
           font-size: 20px;
--- a/website/client/src/components/challenges/findChallenges.vue
+++ b/website/client/src/components/challenges/findChallenges.vue
@@ -68,6 +68,17 @@
 <style lang='scss' scoped>
   @import '~@/assets/scss/colors.scss';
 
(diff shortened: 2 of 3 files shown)


Alys:
@negue We have a couple of issues related to challenge layout problems and one of them has someone working on it now, so I'm just wondering if this is related to either of those issues and/or if the contributor needs to know about this fix. I don't suppose you have before and after screenshots for this PR? Or are there any descriptions about what it changes that you could add, maybe something that's already been written about the problem somewhere else that could be pasted in quickly?

These are the issues I'm thinking of:
- https://github.com/HabitRPG/habitica/issues/11500 (in progress by a contributor)
- https://github.com/HabitRPG/habitica/issues/11259 (on hold waiting for design decisions)

Alys:
Actually, sorry, I see that you've already approved a PR by the contributor for the first issue I listed, so they definitely don't need to know about this. :) 
The other issue might be affected by this though - after this is merged will we need to check #11259 to see if the problems described in it have changed?

negue:
The issues I fixed in this PR in a narrow view:
- (where two challenge cards were side by side) the task-counts per type didn't wrap at all => here I added a wrap to give them some "space" to breath ^^
- all challenge cards were only listed as 80% width => changed to 100%
- added a margin under the add-challenge-button 

I couldn't find any other (challenge) layout issues, if someone does please add them here :)

#11500 (fixed by #11618) shouldn't be affected by these changes

Alys:
@negue Great, thanks! I've marked this PR as "notify helpers" so that when it's merged, I'll be reminded about it and then I'll update https://github.com/HabitRPG/habitica/issues/11259 if needed.

Tressley:
Still seeing display issues. This one seems to be when the username is too long for the line and wraps into the categories.

Screen Shot 2019-12-12 at 9 16 21 AM

negue:
@Tressley this issue will be fixed by #11618

paglias:
Is this ready to go?

Tressley:
@paglias -- LGTM. 👍

PR #11635, opened 2019-12-13
Updates notifications screen to a more consistent UI

gms8994:
#### Updates notifications screen to a more consistent UI

The user interface on the notifications page was inconsistent and not very pretty; I've cleaned it up so that it feels better.

I've also slightly adjusted the order of the items in the list (taken after the "after" screenshot) as `Important announcements` feels like it's more important than `Received Private Message`, though I'm not married to that change as much.

Before:
image

After:
image

----
UUID: 65d86915-6f66-400c-8d58-80ab61bb8ca2

### DIFF
--- a/website/client/src/components/settings/notifications.vue
+++ b/website/client/src/components/settings/notifications.vue
@@ -28,39 +28,38 @@
       <small>{{ $t('unsubscribeAllEmailsText') }}</small>
     </div>
     <div class="col-8">
-      <table class="table"></table>
-      <tr>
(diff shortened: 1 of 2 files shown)


paglias:
Thanks @gms8994 , cc'ing our designer here @Tressley

Tressley:
Digging this. LGTM. 👍

paglias:
@gms8994 thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11636, opened 2019-12-15
Fix incorrect error when sending private messages to a player you have blocked (fixes #11612)

### LINKED ISSUE #11612: incorrect error when sending private messages (PMs) to a player you have blocked
When you have blocked a player from sending private messages to you, if they try to message you they see this error: 
"You can't send a message to this player because they have chosen to block messages."
That's the `notAuthorizedToSendMessageToThisUser` string in `website/common/locales/en/groups.json`
All of that is working correctly and doesn't need to be changed.

You are also not able to send PMs to the player you've blocked. That's also the desired behaviour, however in that situation you currently also see the `notAuthorizedToSendMessageToThisUser` error, which is not correct. The error that you should see should say something like:
"You can't send a message to this player because you have blocked them from sending messages to you."

_Thanks to EliotThe for finding and reporting this._

tsukimi2:
Fixes #11612 

Introduce a new error type blockedToSendToThisUser for the cases where you have blocked a player from sending private messages to you, if you try to message them or if you try to transfer gems to them, you will see the error "You can't send to this player because you have block this player."

### DIFF
--- a/test/api/v3/integration/members/POST-send_private_message.test.js
+++ b/test/api/v3/integration/members/POST-send_private_message.test.js
@@ -66,7 +66,7 @@ describe('POST /members/send-private-message', () => {
     })).to.eventually.be.rejected.and.eql({
       code: 401,
       error: 'NotAuthorized',
-      message: t('notAuthorizedToSendMessageToThisUser'),
+      message: t('blockedToSendToThisUser'),
     });
   });
 
--- a/test/api/v3/integration/members/POST-transfer_gems.test.js
+++ b/test/api/v3/integration/members/POST-transfer_gems.test.js
@@ -89,7 +89,7 @@ describe('POST /members/transfer-gems', () => {
     })).to.eventually.be.rejected.and.eql({
       code: 401,
       error: 'NotAuthorized',
-      message: t('notAuthorizedToSendMessageToThisUser'),
+      message: t('blockedToSendToThisUser'),
     });
   });
 
--- a/website/common/locales/en/groups.json
+++ b/website/common/locales/en/groups.json
@@ -151,6 +151,7 @@
   "toUserIDRequired": "A User ID is required",
   "gemAmountRequired": "A number of gems is required",
   "notAuthorizedToSendMessageToThisUser": "You can't send a message to this player because they have chosen to block messages.",
+  "blockedToSendToThisUser": "You can't send to this player because you have blocked this player.",
   "privateMessageGiftGemsMessage": "Hello <%= receiverName %>, <%= senderName %> has sent you <%= gemAmount %> gems!",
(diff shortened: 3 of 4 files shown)


tsukimi2:
Sorry, just realize that I forgot to add the changes to the other languages as well (only did it for the EN locale.  Will do another updated pull request later today.

Alys:
Changing only the EN locale files is correct! :)  You shouldn't edit the
other files. See the _README* files in the other directories for details.

tsukimi2:
@Alys Oh, ic.  Thanks for the info! ^_^

paglias:
Hi @tsukimi2 ,

looking at the original issue it said

>  All of that is working correctly and doesn't need to be changed.

But it looks like this PR is replacing the existing behavior instead of introducing a new one alongside it?

tsukimi2:
@paglias  Sorry for the late reply.  For the first paragraph of the issue, "When you have blocked a player from sending private messages to you, if they try to message you", this PR should not be changing this part.  The code will still use the notAuthorizedToSendMessageToThisUser string, as the original issue mentioned that for this part, "All of that is working correctly and doesn't need to be changed".

It‘S the second paragraph, ie "When you have blocked a player, if you try to message that blocked player", that needs to be changed and that‘s what this PR is for.  Instead of using the notAuthorizedToSendMessageToThisUser string which results in the inaccurate error message in this scenario currently, the new error string blockedToSendToThisUser is introduced, giving the error message "You can't send to this player because you have blocked this player" instead.

Also, while debugging the code, I find that a similar issie exists for transferring gems.  Ie.  when you have blocked a player and when you try to transfer gems to that player, right now it gives the notAuthorizedToSendMessageToThisUser string as error message ("You can't send a message to this player because they have chosen to block messages.").  This PR modifies it so that it also uses the newly introduced blockedToSendToThisUser string, thus giving the error message "You can't send to this player because you have blocked this player.", which is the same error message as the case of "
When you have blocked a player, if you try to message that blocked player".

paglias:
Sorry I missed that @tsukimi2 ! I'll review this again in early January

paglias:
Hi @tsukimi2  , sorry for the delay but I'm still catching up from the end of year holidays, I hope to get to this PR in the next couple of days

paglias:
thanks @tsukimi2 and sorry for the long time it took! If you have an Habitica account just send me your username or UUID and I'll notate this among your contributions :)

tsukimi2:
@paglias My Habitica username is osiris65.  Thanks!

cefa15ee-c1bc-46e2-95ec-bbf579cd2d85

paglias:
@tsukimi2  thanks! I've awarded your contributor tier.

PR #11646, opened 2019-12-18
Fix purchasing non-recurring subscriptions

phillipthelen:
This fixes a bug where non-recurring subscriptions were not correctly purchased on iOS.

### DIFF
--- a/website/server/libs/payments/apple.js
+++ b/website/server/libs/payments/apple.js
@@ -198,8 +198,8 @@ api.noRenewSubscribe = async function noRenewSubscribe (options) {
 
   for (const purchaseData of purchaseDataList) {
     if (purchaseData.productId === sku) {
-      const transactionId = purchaseData.transactionId;
-      const existingReceipt = await IapPurchaseReceipt.findOne({
+      const { transactionId } = purchaseData;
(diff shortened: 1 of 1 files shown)


website/server/libs/payments/apple.js
```diff
@@ -194,48 +194,48 @@ api.noRenewSubscribe = async function noRenewSubscribe (options) {
...
+        data.gift = gift;
+        data.paymentMethod = this.constants.PAYMENT_METHOD_GIFT;
+      }
+
+      await payments.createSubscription(data);
+      correctReceipt = true;
       break;
     }
   }
+  if (!correctReceipt) throw new NotAuthorized(api.constants.RESPONSE_INVALID_ITEM);
```
SabreCat, line 234: I see that you changed the error code here from `INVALID_RECEIPT` to `INVALID_ITEM` -- is that more appropriate for the fail case if a correct receipt is not found in `purchaseDataList`?

phillipthelen: Yes, since this happens when the receipt itself is valid, but the contents are not correct (p.e. there is no purchase that matches the SKU provided)

SabreCat (approved):

PR #11647, opened 2019-12-21
Fix issues in NPCs should have NPC style on Display Name in chat, Hall. (fixes #11617)

### LINKED ISSUE #11617: NPCs should have NPC style on Display Name in chat, Hall, etc
When [players who are also NPCs](https://habitica.fandom.com/wiki/Kickstarter#Non-Playing_Characters_.28NPCs.29) post to chat, their Display Names don't have the correct formatting for an NPC. You can see what the format should be by looking at Bailey's name in the "Staff and Moderators" list on the right-hand side of the [Tavern](https://habitica.com/groups/tavern). 

For an NPC who is also a contributor or moderator, the style should be the NPC style not the contributor/moderator style. 
An NPC who is not a contributor should of course have the NPC style not the default black name.

The fix to apply the NPC style should be applied in all the places where the contributor/moderator style is applied.

There's already some support in the code for this but it's not working (I haven't looked into why):
https://github.com/HabitRPG/habitica/blob/develop/website/client/src/mixins/styleHelper.js

~Note that currently I'm only assuming that we want a fix for this issue. Name styles used to work as I've described and I'm guessing that removing the NPC stye was accidental, since the NPC players paid a lot in the original kickstarter to be an NPC.~
~I'll leave this issue as suggestion-discussion for a few days so that staff can comment on it if they wish to. If it's intentional that the NPC style should no longer be used, then we should change the formatting on Bailey's name on the Tavern sidebar, and probably also remove any outdated code related to applying NPC styling (e.g., in the file I linked to above).~

~If there's no objections after a few days, I'll change this to help wanted.~
I'll also make a similar issue in the Android repo since the same problem exists there. EDIT: https://github.com/HabitRPG/habitica-android/issues/1254
Thanks to cTheDragons for reporting this for both platforms. We might need a similar issue for iOS too.

Any blacksmith who works on this after (if) it's marked as help wanted can make a test account on their local install be an NPC by adding this to it:

```
    "backer": {
        "npc": "NPC Title",
        "tier": 800,
        "tokensApplied": true,
    }
```
When you post a chat message with that account, you'll see the name is the default black -  it should have the NPC styling.

You can make the NPC also be a moderator by adding this to the account:
```
    "contributor": {
        "text": "Moderator Title",
        "admin": true,
        "level": 8,
    },
```
The name should still have the NPC styling after that change.

tsukimi2:
Fixes #11617 
See the problem description there.

After the bug fix, NPCs now have NPC style display (satou2 as NPC + moderator, satou3 as NPC, and satou4 as moderator in the screen cap below).
Screenshot 2019-12-21 at 11 15 20 PM


website/server/libs/inbox/index.js
```diff
@@ -141,6 +142,7 @@ export async function listConversations (owner) {
   const conversations = conversationsList.map(res => ({
     uuid: res._id,
     ...res,
+    backer: Array.isArray(res.npc) && res.npc.length > 0 ? { npc: res.npc[0] } : { npc: null },
```
paglias: @negue could you give a look at this part since you worked the most on it? thanks!

website/client/src/components/userLink.vue
```diff
@@ -92,9 +92,9 @@ export default {
     },
     isNPC () {
       if (this.backer) {
-        return this.backer.level;
+        return this.backer.tier;
```
paglias, line 95: good catch!

website/server/libs/inbox/index.js
paglias: couldn't `usersMap[res._id].backer` be used here like it's used for `contributor`? By adding backer here https://github.com/HabitRPG/habitica/pull/11647/files#diff-ebb5855000db09391acfb6b15aed8fc5R96 ?

paglias (changes requested):
left one comment

website/server/libs/inbox/index.js
negue: yeah adding `backer: { $last: '$backer' },` to line 97 and then also here as mentioned `usersMap[res._id].backer` should be enough , then the changes in line 128 and 145 are obsolete

negue (changes requested):
Only one thing (two changes)  but otherwise LGTM

tsukimi2:
@paglias  Hi, I‘m on trip now and do not have access to my computer until after jan 3.  Is it ok to wait until afterward for me to modify my pull request based on the above suggested changes? Or is the issue urgent enough to let someone else available to take over the changes?

paglias:
@tsukimi2 no worries, it's totally fine to wait until the holidays are over :)

tsukimi2:
@paglias Requested changes have been submitted.  Pls let me know if further reworks are needed.  Thx.  ^_^

paglias:
Hi @tsukimi2  , sorry for the delay but I'm still catching up from the end of year holidays, I hope to get to this PR in the next couple of days

paglias:
@tsukimi2 looks like some conflicts appeared, probably due to the Inbox PR we merged recently, could you try to solve them? Thanks!

tsukimi2:
@paglias No prob.  It's the biggest holiday of the year afterall! ^_^  Anyway, did the merging as requested, see if there are any problem with them.  Thanks.

website/client/src/components/settings/subscription.vue
```diff
@@ -1,5 +1,23 @@
 <template>
   <div class="standard-page pt-0 px-0">
+    <div class="banner-g1g1 mx-n3 d-flex justify-content-center">
```
paglias: @tsukimi2 did you add this code? It might be the result of some merging gone wrong because I don't think it's supposed to be here

website/client/src/components/settings/notifications.vue
```diff
@@ -55,7 +58,7 @@
               type="checkbox"
               @change="set('pushNotifications', notification)"
             >
-          <td v-else>
+          </td><td v-else>
```
paglias, line 61: good catch!

website/server/libs/inbox/index.js
```diff
@@ -80,6 +80,76 @@ export async function getUserInbox (user, options = {
   return messagesObj;
 }
 
+async function usersMapByConversations (owner, users) {
```
paglias: I think this might be the result of some merge issues because the code for this is currently in website/server/libs/inbox/conversation.methods.js

paglias (changes requested):
Left a couple of comments regarding some merge issues but otherwise it looks good @tsukimi2 !

tsukimi2:
@paglias Thanks for pointing out.  Changes made as requested.

website/client/src/components/settings/subscription.vue
tsukimi2: @paglias Yes, I think it's merging gone wrong, I didn't add this code.  It's been deleted in the new push.

website/server/libs/inbox/index.js
tsukimi2: @paglias Yes, I must have made a mistake during merging.  Not only the function usersMapByConversations, but also the listConversations function (which is supposed to be in conversation.methods.js file also) are incorrectly merged into this file in my previous PR.  The two functions have now been deleted from the .../inbox/index.js file in the new push.

paglias:
thanks @tsukimi2 ! Notated towards your 2nd tier

PR #11690, opened 2020-01-04
Disabled auto notification dismissal for received card on click fix for #11688

### LINKED ISSUE #11688: inconsistent notification dismissal for Cards (greetings, birthday, etc)
If two or more players each send you the same kind of card (e.g., if you have three Birthday cards), you have a notification for each of them (see the screenshot below). That's correct behaviour so far.

There's two ways that you can get to the cards and remove the notifications.

### First way:
1. You click on one notification. **That one notification immediately disappears**; the other notification(s) remain.
2. You're taken to Inventory > Items where you can see the cards.
3. You click on one card and read it.
4. You click the OK button on the card to close it. **A second notification disappears**.

I.e., if you were sent three cards, you've now read only one card but you have only one notification left. That's incorrect; you should have two notifications left.

### Second way:
1. You go directly to the Inventory > Items screen, without going through the notification menu. All notifications are still there (that's correct behaviour).
2. You click on one card and read it.
3. You click the OK button on the card to close it. **A notification disappears**. 

I.e., if you were sent three cards, you've now read one card and you have two notifications left. That's correct behaviour for this way.

Both ways are legitimate paths for the user to take so the dismissal of the notifications should work equally well for each method. The behaviour for the second way is currently correct. The behaviour for the first way is wrong because there are two notification dismissal events when there should be just one dismissal event.

I suggest that the best way to fix this is to prevent the card notifications dismissing themselves when you click on them. I'll leave this as suggestion-discussion for a few days then if there's no objections I'll mark it as help wanted for this fix.

Note that regardless of what changes are made for this issue, the card notifications SHOULD still be dismissed if you click the "x" button on the right-hand side or if you click the "Dismiss All" link.

image

Kevin-DDR:

Fixes #11688

### Changes
[//]: # Disabled the auto dismissal of notification when clicking on a card received notification

----
UUID: 641375c3-346d-4797-adaf-7c0366f9cab4

### DIFF
--- a/website/client/src/components/header/notifications/cardReceived.vue
+++ b/website/client/src/components/header/notifications/cardReceived.vue
@@ -3,7 +3,7 @@
     :can-remove="canRemove"
     :has-icon="true"
     :notification="notification"
-    :read-after-click="true"
+    :read-after-click="false"
     @click="action"
   >
     <div


paglias:
@Kevin-DDR  thanks for this! I'm merging the PR and awarding your contributor tier.

Kevin-DDR:
Thank you !

PR #11702, opened 2020-01-07
Navbar component

Trigur:
### Changes
Minor refactoring for navigation components

----
UUID: f6a74dee-f832-4860-8f9c-63c2fd26ab43


paglias:
Hi @Trigur  , sorry for the delay but I'm still catching up from the end of year holidays, I hope to get to this PR in the next couple of days

website/client/src/components/challenges/index.vue
```diff
@@ -1,33 +1,37 @@
...
 <script>
-import SecondaryMenu from '@/components/secondaryMenu';
+import Navbar from '@/components/ui/simpleNavbar';
 
 export default {
   components: {
-    SecondaryMenu,
+    Navbar,
+  },
+  computed: {
```
paglias, line 21: I think this can be a normal data property instead of a computed one because the data is static

website/client/src/components/groups/index.vue
```diff
@@ -32,13 +13,32 @@
...
 
 export default {
   components: {
-    groupFormModal,
-    SecondaryMenu,
+    GroupFormModal,
+    Navbar,
+  },
+  computed: {
+    tabs () {
```
paglias, line 25: doesn't have to a computed prop

website/client/src/components/groups/membersModal.vue
```diff
@@ -10,7 +10,6 @@
       id="members-modal"
       :title="$t('createGuild')"
       size="md"
-      :hide-footer="true"
```
paglias, line 13: why was this removed?

website/client/src/components/groups/membersModal.vue
```diff
@@ -234,7 +220,7 @@
           </div>
         </div>
       </div>
-      <div class="modal-footer">
+      <div slot="modal-footer">
```
paglias, line 223: I think this might break the layout of the modal

website/client/src/components/hall/index.vue
```diff
@@ -1,34 +1,37 @@
...
-import SecondaryMenu from '@/components/secondaryMenu';
+import Navbar from '@/components/ui/simpleNavbar';
 
 export default {
   components: {
-    SecondaryMenu,
+    Navbar,
+  },
+  computed: {
+    tabs () {
```
paglias, line 22: data prop

website/client/src/components/inventory/equipment/index.vue
```diff
@@ -442,9 +430,9 @@ export default {
     ) === CONSTANTS.equipmentDrawerTabValues.COSTUME_TAB;
   },
   methods: {
-    selectDrawerTab (tabName) {
+    selectDrawerTab (index) {
```
paglias, line 433: why this change?

website/client/src/components/inventory/index.vue
```diff
@@ -1,35 +1,41 @@
...
-import SecondaryMenu from '@/components/secondaryMenu';
+import Navbar from '@/components/ui/simpleNavbar';
 
 export default {
   components: {
-    SecondaryMenu,
+    Navbar,
+  },
+  computed: {
+    tabs () {
```
paglias, line 22: data prop

website/client/src/components/settings/index.vue
```diff
@@ -1,62 +1,53 @@
...
-import SecondaryMenu from '@/components/secondaryMenu';
+import Navbar from '@/components/ui/simpleNavbar';
 
 export default {
   components: {
-    SecondaryMenu,
+    Navbar,
+  },
+  computed: {
+    tabs () {
```
paglias, line 22: data prop

paglias (changes requested):
Left some comments @Trigur but we definitely appreciate abstracting some components in the codebase :)

Trigur:
oh, my english very bad...

website/client/src/components/groups/membersModal.vue
Trigur: because i change class="modal-footer" to slot. And it work's fine
with class:
Снимок экрана 2020-01-19 в 15 48 19
with slot:
Снимок экрана 2020-01-19 в 15 54 10

website/client/src/components/hall/index.vue
Trigur: i think computed property better for few reasons:
- you may need to add conditions of show for some tabs;
- you can incapsulate this property in mixin later;
- i don't want to allow somebody to add tab in method - it's bad practice (i think);

website/client/src/components/inventory/equipment/index.vue
Trigur: i didn't want to change how drawerHeaderTabs works

website/client/src/components/hall/index.vue
paglias: That's true @Trigur but right now there are no conditions, if we ever add them we can move the prop to a computed later.

> i don't want to allow somebody to add tab in method - it's bad practice (i think);

that's right! methods should be avoided when only used to calculate some data

website/client/src/components/groups/membersModal.vue
paglias: thanks, that makes sense!

paglias:
No worries @Trigur , I'd still like the computed props that don't use anything dynamic to be transformed into normal data props

Sorry for the delay!

Trigur:
y y, sure. Waste of time. Gl with project.

PR #11704, opened 2020-01-08
fix Guild and Challenge categories bugs

### LINKED ISSUE #11162: Guild and Challenge categories bugs: duplication, inability to change
As reported by @-R (3807fed9-ada8-41eb-9174-baa26246b57e):

**"on the website I noticed my private guild's assigned categories are repeated ? 4 categories instead of 2."**

**"The refresh [reloading the webpage] didn't change anything. Here's the URL: https://habitica.com/groups/guild/5189073f-56cd-4e2b-9cb6-817567371887 "**

I confirmed in the database that the guild had two categories assigned but each appeared twice in the `categories` array.

I asked "What happens if you edit the guild to try to change the categories? If simply editing and saving doesn't help, try unselecting both of the current categories and selecting another one (any other one will do) and save that, then edit again and put the categories back to what they should be. Does that help?"

@-R said:
**"Nothing changed when editing and saving. I clicked each category tag to manually uncheck it but the change wasn't implemented. I also tried wiping every category and saving a new one (hobbies), but it added hobbies to the category list on top of the duplicates, now I can't remove that either lol!
There's only meant to be 2 assigned categories: Time-Management and Getting Organised. It seems that new category selections can be added but saved category selections cannot be removed. When saving is attempted the category tags momentarily display `*string not found`. Perhaps the database isn't updating correctly because existing parameters aren't being located?"**

Here's the categories array after they tried those edits, still showing a duplicate:
```
    "categories": [
        {
            "_id": "5ccf6a4e8b3fbc0012f0e8ee",
            "slug": "getting_organized",
            "name": "getting_organized"
        },
        {
            "_id": "5ccf6a4e8b3fbc0012f0e8ed",
            "slug": "time_management",
            "name": "time_management"
        },
        {
            "_id": "5cd2f1c8ba5dc10011ff9c23",
            "slug": "hobbies_occupations",
            "name": "hobbies_occupations"
        },
        {
            "_id": "5cd2f1c8ba5dc10011ff9c22",
            "slug": "time_management",
            "name": "time_management"
        }
    ],
```

I've fixed the immeidate problem by deleting the duplicate and the hobbies category that was added just for testing. 

There's a second report that's probably related and I'll update this issue when I have more details about it:

Richard Russell   @Rick30024 (27294fb1-52ed-4b30-a565-3d250ca7febe):
**"Hi when i try to update the tags of my guild it will not save update and displays strings not found when its exits from edit window"**

Trigur:
Fixes #11162
----
UUID: f6a74dee-f832-4860-8f9c-63c2fd26ab43

### DIFF
--- a/website/server/models/challenge.js
+++ b/website/server/models/challenge.js
@@ -409,4 +409,12 @@ schema.methods.closeChal = async function closeChal (broken = {}) {
   Promise.all(backgroundTasks);
 };
 
+schema.methods.sanitizeUpdate = function challangeSanitizeUpdate (updateObj) {
+  // _.assign(this, schema.static.sanitizeUpdate(updateObj)); ?
+  const sanitizedUpdateObj = schema.statics.sanitizeUpdate(updateObj);
+
+  _.merge(this, sanitizedUpdateObj);
+  _.assign(this, { categories: sanitizedUpdateObj.categories });
+};
+
 export const model = mongoose.model('Challenge', schema); // eslint-disable-line import/prefer-default-export
--- a/website/server/controllers/api-v3/challenges.js
+++ b/website/server/controllers/api-v3/challenges.js
@@ -710,7 +710,7 @@ api.updateChallenge = {
     if (!group || !challenge.canView(user, group)) throw new NotFound(res.t('challengeNotFound'));
     if (!challenge.canModify(user)) throw new NotAuthorized(res.t('onlyLeaderUpdateChal'));
 
-    _.merge(challenge, Challenge.sanitizeUpdate(req.body));
+    challenge.sanitizeUpdate(req.body);
 
     const savedChal = await challenge.save();
     const response = savedChal.toJSON();
--- a/website/server/controllers/api-v3/groups.js
+++ b/website/server/controllers/api-v3/groups.js
@@ -475,7 +475,7 @@ api.updateGroup = {
 
     if (req.body.leader !== user._id && group.hasNotCancelled()) throw new NotAuthorized(res.t('cannotChangeLeaderWithActiveGroupPlan'));
 
-    _.assign(group, _.merge(group.toObject(), Group.sanitizeUpdate(req.body)));
+    group.sanitizeUpdate(req.body);
 
     const savedGroup = await group.save();
     const response = await Group.toJSONCleanChat(savedGroup, user);
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -1686,6 +1686,17 @@ schema.methods.updateGroupPlan = async function updateGroupPlan (removingMember)
   }
 };
 
+schema.methods.sanitizeUpdate = function groupSanitizeUpdate (updateObj) {
+  // _.assign(this, schema.static.sanitizeUpdate(updateObj)); ?
+  const sanitizedUpdateObj = schema.statics.sanitizeUpdate(updateObj);
+
+  _.assign(
+    this,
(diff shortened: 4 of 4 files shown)


paglias:
Hi @Trigur  , sorry for the delay but I'm still catching up from the end of year holidays, I hope to get to this PR in the next couple of days

website/server/models/challenge.js
```diff
@@ -409,4 +409,12 @@ schema.methods.closeChal = async function closeChal (broken = {}) {
   Promise.all(backgroundTasks);
 };
 
+schema.methods.sanitizeUpdate = function challangeSanitizeUpdate (updateObj) {
+  // _.assign(this, schema.static.sanitizeUpdate(updateObj)); ?
```
paglias, line 413: can this comment be removed?

website/server/models/challenge.js
```diff
@@ -409,4 +409,12 @@ schema.methods.closeChal = async function closeChal (broken = {}) {
...
+  const sanitizedUpdateObj = schema.statics.sanitizeUpdate(updateObj);
+
+  _.merge(this, sanitizedUpdateObj);
+  _.assign(this, { categories: sanitizedUpdateObj.categories });
```
paglias, line 417: @Trigur could you explain why this new assignment is necessary?

paglias (changes requested):
Left some comments @Trigur , also instead of adding a new method, `sanitizeUpdate` could we reuse the existing static methods `Challenge.sanitizeUpdate` and `Group.sanitizeUpdate`, thanks!

website/server/models/challenge.js
Trigur: Снимок экрана 2020-01-19 в 15 24 57
Снимок экрана 2020-01-19 в 15 25 46
Снимок экрана 2020-01-19 в 15 26 18
Снимок экрана 2020-01-19 в 15 26 42

Trigur: merge works wrong

website/server/models/challenge.js
Trigur: I think `_.assign` better than `_.merge` in that case. But i don't know app good enough, and `_.assign` may broke something.

Trigur:
> Left some comments @Trigur , also instead of adding a new method, `sanitizeUpdate` could we reuse the existing static methods `Challenge.sanitizeUpdate` and `Group.sanitizeUpdate`, thanks!

i think `group.sanitizeUpdate(req.body);` looks much better in controller than `_.assign(group, _.merge(group.toObject(), Group.sanitizeUpdate(req.body)));`. And you can use it in many places.

website/server/models/challenge.js
Trigur: So it's up to you

website/server/models/challenge.js
paglias: Merge only checks first level properties, if you use assign instead of merge in the first case it should work

paglias (commented):
@Trigur it's true that `group.sanitizeUpdate` looks better but we're mixing static and instance methods:

```

schema.methods.sanitizeUpdate = function groupSanitizeUpdate (updateObj) {
  const sanitizedUpdateObj = schema.statics.sanitizeUpdate(updateObj);  // Static method inside an instance one
```

I would prefer it to be refactored using only static methods or only instance ones

Trigur:
y y, shure. Gl with project

PR #11707, opened 2020-01-10
Quest invitation webhook

### LINKED ISSUE #11682: Webhook for quest invitation

### Description

Currently there is no webhook for quest invitation. Quest invitation also does not post any message to party chat (that could be detected with the `groupChatReceived` webhook). So currently the only way to programmatically handle quest invitations is to poll user data for the `RSVPNeeded` flag etc. 

There already seems to be a webhook type `questActivity` that is triggered on quest start and finish. So could we add one more trigger to that webhook for quest invitation? 

I think many people are using auto join scripts that are polling Habitica every hour (for example this one https://habitica.fandom.com/wiki/Google_Apps_Script#Auto_Accept_Quests), some parties even require all members to set it up. With this webhook we could achieve similar thing without any polling and thus reduce server load. Is there any specific reason this does not exist yet? Also wondering why quest invitation does not post any message to party chat?

I am willing to work on this myself if you think it would be useful.

cvuorinen:

Fixes #11682 

### Changes
Sends webhook on quest invitation for members that have it activated.

Added questActivity webhook to API docs with an example + also added an example for userActivity.

----

### DIFF
--- a/test/api/unit/libs/webhooks.test.js
+++ b/test/api/unit/libs/webhooks.test.js
@@ -43,6 +43,7 @@ describe('webhooks', () => {
       options: {
         questStarted: true,
         questFinised: true,
+        questInvited: true,
       },
     }, {
       id: 'userActivity',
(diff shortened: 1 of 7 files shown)


paglias:
thanks @cvuorinen this is looking good with tests and everything! Notated towards your 6th tier!

PR #11708, opened 2020-01-12
stop dropdown menu getting stuck on hover - fixes #11615

### LINKED ISSUE #11615: Dropdown menu can get "stuck" until you mouse over it again
_Edit by Alys 2020-01-04:_

See [this comment](https://github.com/HabitRPG/habitica/issues/11615#issuecomment-570763238) for a way to reproduce this issue.

In brief, reloading a page causes the mouse on-hover effect for the menu of THAT page to be reversed (hovering over it hides the menu, moving the mouse away reveals the menu).

_end of edit_

----

> @JohnJSal (624e48a3-cbea-4bbe-ae44-ca9183f49626): "I don't know if this is a bug, exactly, but I've noticed for a while that when I hover the mouse over the "Guilds" tab at the top of the website, the dropdown menu often remains stuck on the screen even when I've moved the cursor away or even navigate to a different part of the site. It doesn't go away until I move the cursor back over the Guilds tab and sort of "unstick" it. It doesn't happen every single time, but often."

I've seen this as well in Google Chrome. I can't seem to trigger it on demand, though, so it might be difficult to reproduce.

alexthomson1:

Fixes  #11615

### Changes
We noticed that when we loaded a page the top nav item of that page had the class 'down' added to it. This caused the function for the mouse enter and mouse leave events to close a navigation dropdown menu that needed to be open and open a dropdown menu that needed to be closed. We replaced this with two functions; one to open dropdown menus and one to close them.  We also removed the code that added the down class to an item on loading the page. This fixed the issue causing the navigation to stick. 

After fixing this issue we noticed that the dropdown menus had transparent backgrounds on their initial hover. We updated the background color of the dropdown menu to be the correct color.

----
UUID: 907b95cd-02ea-4d27-9bd7-5fdf1284c537


paglias (commented):
Hi @alexthomson1 and thanks for the PR, I'm wondering if it's possible to reuse some of the existing code, in particular the `dropdown (element)` function instead of adding two new functions, open and close dropdown. In particular because the `dropdown (element)` function is still used by `dropdownMobile`

alexthomson1:
> Hi @alexthomson1 and thanks for the PR, I'm wondering if it's possible to reuse some of the existing code, in particular the `dropdown (element)` function instead of adding two new functions, open and close dropdown. In particular because the `dropdown (element)` function is still used by `dropdownMobile`

Thanks for the feedback.  I have refactored the code to remove the dropdown(element) function and instead call openDropdown(element) and closeDropdown(element) from the dropdownMobile function.  I think its important to open and close the dropdown in separate functions because the cause of the bug in the desktop nav was that mouseenter and mouseleave events could cause either a dropdown open or close.  I do agree that we had duplicate code, however, and hopefully this refactor is more clean.

paglias:
thanks @alexthomson1 , the code is definitely clearer and I couldn't reproduce the bug so I'm merging the PR :)

I've notated the PR toward your 2nd tier!

paglias:
@alexthomson1 had to revert the CSS changes because they broke the style of the dropdown menus https://github.com/HabitRPG/habitica/commit/4e6b411763275cf7d1ca512750ac1e9af0e4ff6b, see https://github.com/HabitRPG/habitica/issues/11800#issue-558543003

Issue #11083, opened 2019-03-24
closing one modal sometimes causes previous modal to reopen


Alys:
We've known this bug exists for a while but I can't find an issue for it.

Sometimes when you close a modal, a previous one that you looked at reopens itself.  Once this starts, it generally keeps happening until you reload the website.

Here's a report from @minervathene (d3de6635-37f7-4369-99c3-399d036d0898) today:

"Every time I hatch a pet (doing the new celestial ones) when I click the 'onward' confirmation it reopens the last dialog (a user lookup) that I had open. Happens repeatedly, but not in a fresh session."

jackcogdill:
Hey, wondering if I can work on this? Just encountered this bug myself.

paglias:
sure @jackcogdill ! thanks for the help and let me know if you have any question

jackcogdill:
Has anyone discovered clear steps to reproduce this? I'm having a hard time debugging. Although I suspect it has to do with `modalStack`

Alys:
Just to confirm, this is still happening. There's a new report below. I've never found a way to reproduce it delibeately.

@JohnJSal (624e48a3-cbea-4bbe-ae44-ca9183f49626): "Recently I'm noticing on the website whenever I click away from a popup screen (like the Messages window, or someone's profile), instead of just going straight back to the page I'm on, another popup screen will appear first, usually my own profile window, or that of a random other user."

shanaqui:
@jackcogdill Are you still interested in working on this issue, or should I mark it as help wanted again? Thanks!

jackcogdill:
@shanaqui You can mark it as help wanted.

shanaqui:
Alright, I've done that! Thanks for letting me know.

PR #11739, opened 2020-01-17
Correct some modal repops

SabreCat:

Partial fix for #11083 

### Changes

Allows `app.vue` to receive a new event `habitica::dismiss-modal`. When emitted, this event deletes the passed modal ID from the modal stack, ensuring that the modal stack handler won't attempt to re-show it after a later modal is closed. ~~Note that the Bootstrap-Vue hide modal event still needs to be emitted, or a router push initiated to change the view, to actually close the modal.~~

I found two places where I could reproduce the original issue and where emitting the new event resolved it:

- Creating/updating/awarding/deleting Challenges
- Sending a gift to another user

I wasn't able to reproduce the specific scenario described in the issue ticket, with pet hatching / user profiles, so I'm not counting this as a complete fix.

Also corrects one place in the Challenge flow where we unnecessarily used a hardcoded English string instead of a localized string.

### DIFF
--- a/website/client/src/app.vue
+++ b/website/client/src/app.vue
@@ -612,10 +612,18 @@ export default {
 
       this.$root.$on('habitica::dismiss-modal', oldModal => {
         if (!oldModal) return;
-        const removeIndex = this.$store.state.modalStack
+        let removeIndex = this.$store.state.modalStack
           .map(modal => modal.modalId)
           .indexOf(oldModal);
-        if (removeIndex >= 0) this.$store.state.modalStack.splice(removeIndex, 1);
+        if (removeIndex >= 0) {
+          this.$store.state.modalStack.splice(removeIndex, 1);
+        }
+        removeIndex = this.$store.state.modalStack
+          .map(modal => modal.prev)
+          .indexOf(oldModal);
+        if (removeIndex >= 0) {
+          delete this.$store.state.modalStack[removeIndex].prev;
+        }
       });
     },
     validStack (modalStack) {
--- a/website/client/src/app.vue
+++ b/website/client/src/app.vue
@@ -610,8 +610,11 @@ export default {
         if (modalBefore) this.$root.$emit('bv::show::modal', modalBefore, { fromRoot: true });
       });
 
+      // Dismiss modal aggressively. Pass a modal ID to remove a modal instance from the stack
+      // (both the stack entry itself and its "prev" reference) so we don't reopen it
       this.$root.$on('habitica::dismiss-modal', oldModal => {
         if (!oldModal) return;
+        this.$root.$emit('bv::hide::modal', oldModal);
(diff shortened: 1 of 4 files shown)


website/client/src/components/payments/sendGemsModal.vue
```diff
@@ -294,6 +294,7 @@ export default {
       this.sendingInProgress = false;
     },
     close () {
+      this.$root.$emit('habitica::dismiss-modal', 'send-gems');
```
paglias, line 297: could we have the `this.$root.$emit('bv::hide::modal', modalName);` call inside the `habitica::dismiss-modal` event handler?

website/client/src/app.vue
```diff
@@ -609,6 +609,22 @@ export default {
 
         if (modalBefore) this.$root.$emit('bv::show::modal', modalBefore, { fromRoot: true });
       });
+
+      this.$root.$on('habitica::dismiss-modal', oldModal => {
```
paglias, line 615: can we add a comment explaining what this does and how it works?

website/client/src/app.vue
```diff
@@ -609,6 +609,22 @@ export default {
...
+        }
+        removeIndex = this.$store.state.modalStack
+          .map(modal => modal.prev)
+          .indexOf(oldModal);
+        if (removeIndex >= 0) {
+          delete this.$store.state.modalStack[removeIndex].prev;
+        }
+      });
     },
     validStack (modalStack) {
```
paglias, line 632: I was wondering if we could extend this function to add the new functionality since it seems to have some overlapping

website/client/src/components/payments/sendGemsModal.vue
SabreCat: Hmm, I think when I tried that it caused some issues, but I'll try again and document what happened--maybe you'll know how to fix it!

website/client/src/app.vue
SabreCat: Hmm, can you elaborate? `validStack` seems to be used to clean things up when a single modal event adds multiple copies of a modal to the stack, which isn't the problem at hand--the problem is that our `modalStack` function does exactly what it's supposed to do, but we don't actually want to return to the modals earlier in the sequence.

website/client/src/components/payments/sendGemsModal.vue
SabreCat: Ahh, I think that must have been before I implemented the bit that removes the `prev` data. So it does work, and that makes it much easier to use--we can just call `habitica::dismiss-modal` anywhere we want instead of `bv::hide::modal`!

website/client/src/app.vue
paglias: got it!

paglias (approved):
LGTM

PR #11766, opened 2020-01-21
Fixing layout issues for the private messages page

negue:
- [x] First batch of layout issues (max/min heights)
- [x] Added autosizing textarea
- [x] clickable face-avatar to open the profile
- [x] @username on a second line
- [x] refresh on sync
- [x] reset un-read counter
- [x] new conversation (item) layout
- [x] added descriptions why the textarea is disabled
- [x] un-block users from the conversation list
- [x] autosize textarea after sending the message
- [x] show only the current avatar on PM page


Tressley:
This was reported by Alys in the testing Slack: 
>  If you're on the Inbox page (https://habitrpg-delta.herokuapp.com/private-messages) and new PM message arrives from a player who hasn't PMed you before (i.e., when you get an email saying there's a new message from that person), clicking the sync button in the header doesn't make the new conversation appear. You have to reload the page before it appears.

We need the empty state for a new user that doesn't have any messages: https://zpl.io/VY1reLE

Added an updated design that accounts for the username and time on the second line in the conversations column, reduced height of the textarea. https://zpl.io/2yGqrDo

Alys:
Here's another couple of non-layout bug reports, related to the one Tressley posted above:

"If you're on the Inbox page (https://habitica.com/private-messages) and new PM message arrives (in either a new conversation or an existing conversation), after you've reloaded the page and viewed the message, the red Messages counter in the header doesn't disappear."
_[that report was originallly from last year for the test site but it's still happening in production so I've edited it to change the URL]_

_From an RaB report this month:_
"When I already have the inbox open, and a new message comes in, syncing will not make that message appear. I have to click user-> messages as if I were on a different page, and only then the new message will show up."
I've replicated that (reloading the page also makes the message appear).

negue:
All mentioned issues should be fixed now. :)

Tressley:
Some additional feedback:
- Need to tighten up the different padding going on in the conversations column.
image

compared to the design

image

- The textarea has weird styling going on. It looks like the placeholder color is being applied to the value. It also jumps in height when I start to type. It should only grow in height once the text wraps.
image

- The conversations column width jumps when using the search.

negue:
Fixed the style issues, note: all controls had `color: $gray-200` as default, which then "looked like" as it's being in the placeholder color, now the "default" is the focus color `$gray-50`, ping me if this isn't the expected "normal" color^^

On the textarea: there is a weird issue, the browser only removes the scrollHeight / clientHeight in steps per key event - 56-54-52 which then looks weird 
> if you had content for 2 lines and then removes the last line and then the control scrollHeight (which is used to set the minHeight) is decreasing per key
as a little "fix" atleast when textarea was emptied it jumps back to the minimal height (not the scrollHeight at this point)

negue:
@phillipthelen I've added `optOut` and `chatRevoked` to the `/api/v4/inbox/conversations` api results, might be handy for the mobile apps :)

Alys:
> @phillipthelen  I've added optOut and chatRevoked to the /api/v4/inbox/conversations api results, might be handy for the mobile apps :)

@negue Does this mean that regular players will be able to tell when another player has had their chat permissions revoked? If so, that might have implications related to moderation.

Does it also mean that when they can't send a message to another player, they'll be able to tell whether it's because the player has opted out of all messages or has blocked them specifically? We've previously not made any distinction for privacy reasons (if you've blocked a specific player you might not want that player to know that you blocked only them).

negue:
@Alys  I see what you mean, currently only the following states are visible:

- the logged in user has opt-out of pm's 
- the logged in user has revoked chat rights
- the conversation partner has opt-out of pm's

I can remove the `chatRevoked` from the conversations results again, guess I took one step too far :grin:

Alys:
Ah cool, thanks! :) I have no objection at all to having a discussion about what user-facing features would make it desirable for `chatRevoked` to be known. There might be some other way we could enable those features without releasing that information. E.g., if the intention was to warn you that if you send a message the player won't be able to reply, then maybe we could change the `chatRevoked` status so that it prevents you receiving messages as well as sending them.

For this part:

> they'll be able to tell whether it's because the player has opted out of all messages or has blocked them specifically

An alternative option would be to add an attribute called something like `canSendMessages` to the `conversation` object. It would be true if nothing was stopping you sending a PM to that person, and false if the person had opted out OR had blocked you. That would allow the clients to prevent you typing a message that you wouldn't be able to send, without telling you whether the person had blocked you specifically.

By the way, moderators and staff should always be able to send a PM regardless of whether the recipient has tried to block the mod/staff, or has opted out of all PMs, or has chatRevoked set to true. The easiest way to enable that on all clients might be for `canSendMessages` to always be `true` if the logged-in user has `contributor.admin` set to `true`

website/client/src/pages/private-messages.vue
```diff
@@ -691,6 +737,31 @@ export default {
...
+      }
+
+      if (this.user.inbox.optOut) {
+        return {
+          title: this.$t('PMDisabledCaptionTitle'),
+          description: this.$t('PMDisabledCaptionText'),
+        };
+      }
+
+      console.info(this.selectedConversation);
```
paglias: Looks like this is breaking the build on delta, can you remove it?

Tressley:
@negue -- Looking at the conversation between you and Alys, I think we may just want to have the following states:
- **Logged in user has disabled private messages**
> **Private messages are disabled**
You can still send messages, but no one can send them to you.
- **Receiving user has disabled private messages**
> **You can not reply to this conversation**
This user is no longer receiving private messages.
- **Logged in user is blocked by receiving user**
> **You can not reply to this conversation**
This user is no longer receiving private messages.
- **Logged in user has blocked receiving user**
> **You can not reply to this conversation**
Unblock this user to continue sending and receiving messages.

**Note:** I noticed when using the "Block" option from the conversations menu, the menu option doesn't switch to "Unblock."

- **Incorrect styling for messaging** (This should be using the same color as the placeholder update `$gray-50`.)
image

website/client/src/pages/private-messages.vue
negue: oof sorry, removed! I wonder why this check isn't added on the current CI

negue:
@Alys now there is only `canReceive` which is false if the target user has `optOut` or `chatRevoked`

Todo: 3rd State
- [ ] set `canReceive` to false if the target user has blocked you

Alys:
@negue `canReceive` sounds good! Can you please remove the `chatRevoked` part from it though? I've checked with Beffymaroo and the other mods and we do want players who've had their chat permissions revoked to be able to receive PMs. There's use cases where it makes sense, for example when muting is only temporary (which is pretty common), or if a parent wants to send a message or a Gem gift to a child.

@Tressley Love that list of states  - very thorough and clear!

negue:
there was also an issue I just recognized, if you click on the face-avatar / user-link it opens the profile modal BUT not for the conversation target ... fixed.

@Alys removed `chatRevoked` and added the isAdmin() check, so these should be all checks for now ^^

Tressley:
- I can't seem to send messages between two users. I only see the "This user is no longer receiving private messages" message.
- Only the logged in user can see their avatar in the user modal from the conversation. The receiving user's avatar doesn't show in their modal.
image
- No empty state for a user without any messages. (https://zpl.io/aBG30Gm)
image
- Spacing is still incorrect for the conversations in the conversations column. This takes up way too much space to not have a message.
image

Alys:
The missing avatar and stats in the user modal are unrelated to this PR. It's from this bug: https://github.com/HabitRPG/habitica/issues/11830

Alys:
I'm seeing "An unexpected error occurred" in the Messages page on delta and none of my PMs load. This is in Loggly:
```
json:
httpCode:500
isHandledError:false
level:error
message:TypeError: Cannot read property 'includes' of undefined
    at /app/website/transpiled-babel/libs/inbox/conversation.methods.js:152:62
    at Array.map (<anonymous>)
    at listConversations (/app/website/transpiled-babel/libs/inbox/conversation.methods.js:140:43)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async handler (/app/website/transpiled-babel/controllers/api-v4/inbox.js:123:20)
method:GET
originalUrl:/api/v4/inbox/conversations
timestamp:2020-02-11T22:14:57.909Z
```

User ID d904bd62-da08-416b-a816-ba797c9ee265

I'm guessing it's specific to particular data otherwise everyone testing PMs on delta would see it so if I can provide any information about PMs I have, I'll be happy to - send me a message privately somewhere. :)

negue:
> I can't seem to send messages between two users. I only see the "This user is no longer receiving private messages" message.

I found an issue when you start a new conversation then I also received `This user is no longer receiving private messages` - this is fixed now - if it wasn't like that I'd need more data to reproduce ^^

> No empty state for a user without any messages.

Was a broken if-statement :smile:  should be fixed now

> Spacing is still incorrect for the conversations in the conversations column. This takes up way too much space to not have a message.

its set to max-height now, so it'll autosize now

> message:TypeError: Cannot read property 'includes' of undefined
    at /app/website/transpiled-babel/libs/inbox/conversation.methods.js:152:62

this is an issue when a user hasn't blocked anyone .. I didn't think that this field could be `undefined` => fixed

Pushed to delta

Tressley:
I think this is good to go! 👍

Tressley:
Starting to encounter some new bugs.
- Search string isn't correct in the blue banner
- Messages missing/out of order after using "Jump to recent messages." I think this is due to the search string remaining. Perhaps we should clear the search string after jumping to recent.
- Can only "load newer messages" once.
 
On Chrome (Version 80.0.3987.149, Win 10)
- Noticing messages jumping when trying to highlight text. This is due to a browser scrollbar appearing and disappearing.
- Textarea shrinks in size when typing a message

PR #11769, opened 2020-01-27
Challenge exception-handling / clientside prize validation

### LINKED ISSUE #11765: Attempting to create a Challenge with not enough Gems locks Challenge submission until refresh

### Description

If you attempt to create a Challenge with a Gem prize larger than your current available Gems, you get an error that you don't have enough Gems and need to reduce the prize or buy some Gems first. All well and good! But if you then reduce the prize or get Gems, as directed, the Create Challenge button stays disabled. You can't create the Challenge unless you refresh the page first.

#### Console Errors

Unhandled error in Vue.js code. logging.js:30
Error: Error: "Request failed with status code 401"
    createError createError.js:16
    settle settle.js:17
    handleLoad xhr.js:59

aristizabal95:

Fixes #11765 

### Changes
a client-side validation has been added, which checks that the challenge prize doesn't surpass the `maxPrize` the user can pay for. Additionally, the `loading` property is set to false once the create challenge request is finished, so that the submit button doesn't stay disabled.

----
UUID: 88643b14-48fa-4ae1-b45d-5ff7b793a257


aristizabal95:
Added exception handling for any error the server may generate when creating the challenge. Not sure if this should be in a separate commit?

Also, I tried to make this try/catch block more specific by checking the error name, but it seems that the generated error has no name. Would be a good idea to include names to the errors generated by the server, so that they can be handled independently.

aristizabal95:
Adding the try/catch block may also solve #11545

website/client/src/components/challenges/challengeModal.vue
```diff
@@ -556,14 +557,21 @@ export default {
...
+        if (this.cloning) {
+          challenge = await this.$store.dispatch('challenges:cloneChallenge', {
+            challenge: challengeDetails,
+            cloningChallengeId: this.cloningChallengeId,
+          });
+          this.cloningChallengeId = '';
+        } else {
+          challenge = await this.$store.dispatch('challenges:createChallenge', { challenge: challengeDetails });
+        }
+      } catch (e) {
```
paglias, line 578: thanks @aristizabal95 ! I see that you're returning inside the `catch` block which makes sense but it'll prevent these 2 lines from being executed while they probably should https://github.com/HabitRPG/habitica/pull/11769/files#diff-a73635db46b2d5f7643520a5e8ebcbf1R601-R603


It might be a good idea to refactor this to use a try/catch/finally flow with all the cleanup that is shared by the success and failure state in the `finally` block

paglias (changes requested):
Left a comment but the direction is correct :)

website/client/src/components/challenges/challengeModal.vue
aristizabal95: Hi @plagias, sure! Could do that change, although I would like to discuss it from an UX perspective. 

At the moment, if a server-side error is thrown during the challenge creation an alert is displayed describing such error and the form stays up for the user to correct the changes. Adding the `finally` block clearing the working challenge contents and closing the form modal would remove all unsaved progress, forcing the user to rewrite everything from scratch if they made a mistake.

Now, with the changes I made an user could just resubmit the form without any change, which may also not be good UX. I'd say a good solution would be to keep the submit button disabled until a change is made on any of the inputs (preferably on the input that's causing errors, but I would need to know which one was from the server). I believe this was the preferred behaviour for issue #11545 .

Tell me your opinion and I'll get working on your preferred solution!

paglias: @aristizabal95 we defenitely want people to be able to correct the error and resubmit, I missed that part :)

> I'd say a good solution would be to keep the submit button disabled until a change is made on any of the inputs (preferably on the input that's causing errors, but I would need to know which one was from the server)

This looks like a good idea, I'd say for now we're not going to change the server response so let's stick to keeping it disabled until at least an input is modified

thanks!

aristizabal95:
@paglias Just made the change. I feel there's a lot that could be done to the client-side validation to make it more user-friendly. If it's okay, I could add an issue and get working on that, as well as adding the same validation to challenge update (I see there's a `TODO` saying that's missing). For now I kept it focused on the issue and implemented the fix we discussed. Hope everything looks okay! 😄

paglias:
Sorry for the delay here but I won't have time to review PRs until the 15th, regarding the new issue go ahead and create it! I'll give it a look and if everything's fine you can make a pr

aristizabal95:
Don't worry, review this whenever you can. I've also been busy so no rush!

website/client/src/components/challenges/challengeModal.vue
```diff
@@ -618,6 +632,12 @@ export default {
     toggleCategorySelect () {
       this.showCategorySelect = !this.showCategorySelect;
     },
+    enableSubmit () {
```
paglias: just one thing, could you make this function throttled so that it's only called every 250ms like [here](https://github.com/HabitRPG/habitica/blob/f8bab04a0e66a819f6d7ca5b210d27d991446bf3/website/client/src/components/challenges/sidebar.vue#L199)? 

The throttle function is a lodash module that can imported this way https://github.com/HabitRPG/habitica/blob/f8bab04a0e66a819f6d7ca5b210d27d991446bf3/website/client/src/components/challenges/sidebar.vue#L92

paglias (changes requested):
This is working well! I only left a comment but after the change is done we can merge it

website/client/src/components/challenges/challengeModal.vue
aristizabal95: Had no idea about throttling. Sure thing!

paglias:
@aristizabal95  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11791, opened 2020-01-29
clarify all "Invalid uuid" messages by making each unique

Alys:
This changes all 'Invalid uuid' error messages so that each one is unique.

This will make it much faster to work out what's causing one of those errors.


paglias (approved):
LGTM and definitely useful!

PR #11792, opened 2020-01-29
Improve G1G1 process

SabreCat:

### Changes

* Adds a new notification for when we run the "Gift One, Get One" promotion, providing a colorful item in the notifs dropdown to draw attention to the event
* Adds a modal called up by the above notification, that allows the user to search for another user to gift to
* Fixes a couple of style issues on the payment success modal that was causing the checkmarked circle not to appear correctly

Note: we will likely want to use this modal elsewhere (e.g. as a link from the subscriptions page)--in that case, we'll take off the promo-specific div. But for now this is safe to have in develop/prod even with the promo verbiage, as it isn't called up by anything live.

EDIT 2/20: The core changes were included in the recent Subscriptions page overhaul, as it wanted the gifting modal as well, so this PR now handles the optimizations requested by @paglias in the comments below, and removes a couple of now unnecessary modal hide calls after updates to the `dismiss-modal` functionality.


paglias:
I think the 2 PNG files con be optimized, as right now we only compress files that end up in sprites

website/client/src/components/header/notifications/g1g1.vue
```diff
@@ -0,0 +1,95 @@
+<template>
+  <div
```
paglias, line 2: since most of the strings here are static a few `v-once` will help with performances

website/client/src/components/payments/selectUserModal.vue
```diff
@@ -0,0 +1,225 @@
+<template>
+  <b-modal
+    id="select-user-modal"
+    :hide-header="true"
+    :hide-footer="true"
+    @hide="onHide()"
+  >
+    <div class="g1g1 mb-4 text-center">
+      <div class="heading">Gift a Subscription, Get a Subscription</div>
```
paglias: do we want to use translatable strings here?

website/client/src/components/payments/selectUserModal.vue
```diff
@@ -0,0 +1,225 @@
...
+      <div
+        class="modal-close"
+        @click="close()"
+      >
+        <div class="svg-icon" v-html="icons.close"></div>
+      </div>
+    </div>
+    <div class="d-flex flex-column align-items-center">
+      <div class="mb-2">
+        <strong> {{ $t('sendGiftToWhom')}} </strong>
```
paglias: v-once

website/client/src/components/header/menu.vue
```diff
@@ -4,6 +4,7 @@
     <profileModal />
     <report-flag-modal />
     <send-gems-modal />
+    <select-user-modal />
```
paglias: given that this modal seems specific for G1G1 maybe we shouldn't load it as long as it's not active?

website/client/src/components/header/notifications/g1g1.vue
SabreCat: OK! Does that cause trouble when users change their locale preference, though?

website/client/src/components/payments/selectUserModal.vue
SabreCat: Yeah, this was sort of a placeholder because we're not actually running the promo right now. I can go ahead and get the strings in for translation and do things proper-like, though!

website/client/src/components/header/menu.vue
SabreCat: Good call!

SabreCat: ...although see https://github.com/HabitRPG/habitica/pull/11823/files for possible non-G1G1 usage

website/client/src/components/header/notifications/g1g1.vue
Alys: I don't know if this helps, but when you change your language at User Icon > Settings, the website reloads itself.

paglias: yes, since the site is reloaded when the locale changes there should be no issues

website/client/src/components/header/menu.vue
paglias: in that case nevermind!

PR #11793, opened 2020-01-30
Sign in with Apple

phillipthelen:
This adds a new API call that allows users to log in through apple.


website/client/src/components/static/appleRedirect.vue
```diff
@@ -0,0 +1,29 @@
...
+<style lang='scss'>
+@import '~@/assets/scss/static.scss';
+</style>
+
+<script>
+
+export default {
+  data () {
+    return {
+      html: '',
```
paglias: I'm probably missing something but this property doesn't seem to be ever changed?

website/server/controllers/api-v3/auth.js
```diff
@@ -146,6 +146,35 @@ api.loginSocial = {
...
+// Called as a callback by Apple. Internal route
+api.loginApple = {
+  method: 'GET',
+  middlewares: [authWithHeaders({
+    optional: true,
+  })],
+  url: '/user/auth/apple',
+  async handler (req, res) {
+    req.body.network = 'apple';
+    return loginSocial(req, res);
```
paglias, line 180: So, here, `loginSocial` with `req.body.network = 'apple';` calls `_appleProfile` which seems to need either `req.body.code` or `req.body.id_token` but this one being a GET request it'll never have a body?


```

  if (req.body.code) {
    const response = await auth.accessToken(req.body.code);
    idToken = jwt.decode(response.id_token);
  } else if (req.body.id_token) {
    idToken = await jwt.verify(req.body.id_token, applePublicKey, { algorithms: ['RS256'] });
  } 
```

paglias (commented):
Left a couple of questions

paglias:
Ah, a small thing, it looks like in the auth form at `/static/plans` the Apple button is lacking an icon

Schermata 2020-02-01 alle 10 56 04

paglias:
And we need to add some tests for the new Apple auth, currently there are for FB/Google at

- `test/api/v3/integration/user/DELETE-user.test.js`
- `test/api/v3/integration/user/auth/POST-user_auth_social.test.js`
- `test/api/v3/integration/user/auth/DELETE-user_auth_social_network.test.js`

website/client/src/components/static/appleRedirect.vue
phillipthelen: You're right. I copied the page from a different static page and forgot to remove this

website/server/controllers/api-v3/auth.js
phillipthelen: You're right. But it does work 🤔 . Does express maybe put the query arguments to also be available through `req.body`?

phillipthelen:
That page probably has the white apple icon. I'll fix that.
I'll also add some tests to match the existing ones.

website/server/controllers/api-v3/auth.js
phillipthelen: https://www.npmjs.com/package/body-parser I think the query params are available in `req.body` because of this package that we use.

paglias: hmm weird, that should only fill `req.body` with data from the body. Anyway, just to be clearer, could we change the code to use `req.query` to specify where it's coming from?

phillipthelen: we _do_ need query for handling iOS login unfortunately. But I'll change it, so that it checks both of those values.

website/server/libs/auth/social.js
```diff
@@ -37,11 +37,13 @@ const auth = new AppleAuth(JSON.stringify({
 
 async function _appleProfile (req) {
   let idToken = {};
-  if (req.body.code) {
-    const response = await auth.accessToken(req.body.code);
+  const code = req.body.code ? req.body.code : req.query.code;
+  const passedToken = req.body.id_token ? req.body.id_oken : req.query.id_token;
```
paglias: I think there's a typo here, `req.body.id_oken`

paglias: I think there's a typo here, `req.body.id_oken`

website/client/src/store/actions/auth.js
```diff
@@ -82,6 +82,24 @@ export async function socialAuth (store, params) {
...
 
+export async function appleAuth (store, params) {
+  const url = '/api/v4/user/auth/apple';
+  const result = await axios.get(url, {
+    params: { code: params.code },
+  });
+
+  const user = result.data.data;
+
+  const userLocalData = JSON.stringify({
```
paglias, line 96: not specifically related to this PR but the code to stringify and save in localstorage the auth keys is duplicated in a few places in this file

website/server/libs/auth/social.js
```diff
@@ -21,16 +24,49 @@ function _passportProfile (network, accessToken) {
...
+  let idToken = {};
+  const code = req.body.code ? req.body.code : req.query.code;
+  const passedToken = req.body.id_token ? req.body.id_token : req.query.id_token;
+  if (code) {
+    const response = await auth.accessToken(code);
+    idToken = jwt.decode(response.id_token);
+  } else if (passedToken) {
+    idToken = await jwt.verify(passedToken, applePublicKey, { algorithms: ['RS256'] });
+  }
+  return {
```
paglias: as discussed outside github there should be a check for when `idToken` doesn't have all the necessary data

paglias (changes requested):

website/server/libs/auth/social.js
```diff
@@ -21,16 +24,49 @@ function _passportProfile (network, accessToken) {
...
 }
 
+const applePrivateKey = nconf.get('APPLE_AUTH_PRIVATE_KEY');
+const applePublicKey = nconf.get('APPLE_AUTH_PUBLIC_KEY');
+
+const auth = new AppleAuth(JSON.stringify({
+  client_id: nconf.get('APPLE_AUTH_CLIENT_ID'), // eslint-disable-line camelcase
+  team_id: nconf.get('APPLE_TEAM_ID'), // eslint-disable-line camelcase
+  key_id: nconf.get('APPLE_AUTH_KEY_ID'), // eslint-disable-line camelcase
+  redirect_uri: 'https://habitica.com/api/v4/user/auth/social', // eslint-disable-line camelcase
```
paglias: this won't work on delta since it tries to hit the main server

paglias (changes requested):

paglias:
Current issues:

- [ ] The profile name is not filled

Schermata 2020-02-27 alle 14 31 34

this is how it looks with google sign in

Schermata 2020-02-27 alle 14 32 22

- [x] The email is not stored correctly (should be stored in an array under the `value` property)

Schermata 2020-02-27 alle 14 34 47

this is how it looks with google sign in

Schermata 2020-02-27 alle 14 35 34

- [x] Handle missing `idToken` https://github.com/HabitRPG/habitica/pull/11793/files/6d402f083353b119eba13d58ac7f2c5b1e441da6#r383370095

- [ ] Tests mirroring what we have for FB / Google

paglias (approved):

PR #11795, opened 2020-01-31
Prevent server from cancelling still active subs

phillipthelen:
Just a small check for the call that cancels android subs. If the autoRenews field is true, the sub is still active on google play and should not be cancelled. Probably not strictly needed, but a good safety measure.

### DIFF
--- a/test/api/unit/libs/payments/google.test.js
+++ b/test/api/unit/libs/payments/google.test.js
@@ -278,29 +278,6 @@ describe('Google Payments', () => {
         });
     });
 
-    it('should not throw an error if subscription is still valid', async () => {
-      iap.getPurchaseData.restore();
-      expect(iapSetupStub).to.be.calledOnce;
-      expect(iapValidateStub).to.be.calledOnce;
-      expect(iapValidateStub).to.be.calledWith(iap.GOOGLE, {
-        data: receipt,
-        signature,
-      });
-      expect(iapIsValidatedStub).to.be.calledOnce;
-      expect(iapIsValidatedStub).to.be.calledWith({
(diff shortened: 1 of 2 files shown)


paglias:
Is this actually a problem? I mean did it happen that we canceled a subscription that wasn't expired? Because we only cancel Android subscriptions automatically from here https://github.com/HabitRPG/habitrpg-email-server/blob/master/libs/googlePayments.js#L68

phillipthelen:
The next android update will also start using this api call, so that when the user cancels their subscription and then views the page in the app, we can reflect the cancelled state faster. I don’t expect that to cause any issues, but think the additional check gives us a bit more safety.

--  
Phillip Thelen
Sent with Airmail

phillipthelen:
I also noticed that right now the api prevents the cancellation from happening before the sub is actually over. But since the `cancelSubscription` method handles that properly by adding the time to `extraMonths` we want the API call to work before the end date, so that we can display to the user that their sub is cancelled but still active untill the end date.

Alys:
I don't know if this level of detail is important here so feel free to just ignore this if it's not helpful. :)

The `cancelSubscription` method shouldn't be adding to `extraMonths`. It should be setting a termination date based on what's left of the subscription. 
If there's any time stored in `extraMonths` from a previously purchased subscription, it should be used to extend the termination date.

phillipthelen:
You’re right about that. I somehow managed to look at what it does and then describe it incorrectly within a few minutes 😬 I didn’t make any changes to that function anyway. So the way it works now is also the way it’ll work after this or

--  
Phillip Thelen
Sent with Airmail

PR #11798, opened 2020-02-01
Show "no owner" when challenge leader cannot be found and allow tasks to be seen fixes #11768

### LINKED ISSUE #11768: challenge bugs when owner/leader deletes their account
When a challenge owner deletes their Habitica account, any challenges they own have these problems:

1. In the "My Challenges" and "Discover Challenges" pages, the "Created By" field is blank. The first screenshot below shows an example of this problem in the "Test Challenge - don't join" challenge.
It should say "no owner" as decided [here](https://github.com/HabitRPG/habitica/issues/5979#issuecomment-151280530). A translatable string already exists for this and should be used instead of creating a new one:
`challenge.json:  "noChallengeOwner": "no owner",`

2. In the challenge's own page, the tasks are invisible. This is because when the challenge page is loading, there are errors when fetching data about the owner (you can see the errors in the javascript console), and (I assume) this prevents the page from displaying all of the challenge's other data.
The desired behaviour is that the challenge page should recognise that there's no owner and handle that gracefully instead of throwiing errors. The "Created By" field should say "no owner" and the tasks should appear as normal.
You can see this error displayed in the second and third screenshots below. The second screenshot is a test challenge that has me as the owner. The third screenshot is the same challenge after I have changed the owner to a User ID that is not in use (i.e., that mimics the case of the challenge owner deleting their account).
The test challenge is available here: https://habitica.com/challenges/f21ad727-3d0d-47c5-a5a3-e6311aa1e3e6

image

image

image

SpeedyLom:
Fixes #11768

### Changes
- Check if `challenge.leader` is `null` before trying to display the user's name after "Created By:"
- Check the `challenge.leader` isn't `null` before trying to get its `id` for comparison when determining if the current user `canDelete` or `canEdit`

----

### DIFF
--- a/website/client/src/components/challenges/challengeDetail.vue
+++ b/website/client/src/components/challenges/challengeDetail.vue
@@ -18,7 +18,10 @@
           <div>
             <span class="mr-1 ml-0 d-block">
               <strong v-once>{{ $t('createdBy') }}:</strong>
-              <user-link
+              <span v-if="challenge.leader === null">
+                {{ $t('noChallengeOwner') }}
+              </span>
+              <user-link v-else
                 class="mx-1"
                 :user="challenge.leader"
               />
--- a/website/client/src/components/challenges/challengeItem.vue
+++ b/website/client/src/components/challenges/challengeItem.vue
@@ -43,7 +43,14 @@
         >
           <div class="owner-item">
             <strong>{{ $t('createdBy') }}:</strong>
+            <span
+              v-if="challenge.leader === null"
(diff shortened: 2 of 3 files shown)


paglias:
Sorry for the delay here but I won't have time to review PRs until the 15th

SpeedyLom:
Thanks for the update much appreciated 🙂

paglias:
@tthorpe88  thanks for this! I'm merging the PR and awarding your contributor tier.

Issue #11794, opened 2020-01-31
some Premium and Wacky Hatching Potions have errors in _addlNotes data


Alys:
This bug can be seen when you use the API to fetch Habitica's `content` data with this command this:
```
curl -X GET 'https://habitica.com/api/v3/content?language=en'
```

For some items, such as the Royal Purple Hatching Potion, you'll see the error shown below in the `_addlNotes` (additional notes) data:

```
"premiumHatchingPotions": {
	"RoyalPurple": {
		"value":2,
		"text":"Royal Purple",
		"limited":true,
		"key":"RoyalPurple",
		"notes":"Pour this on an egg,
		 and it will hatch as a Royal Purple pet.",
		"_addlNotes":"Available for purchase until String 'dateEndundefined' not found..",
		"premium":true
	},
```

That error should not be there.

Below is some of the code related to this issue.

https://github.com/HabitRPG/habitica/blob/257a52cba1712f4fbc5b0265fd03dbd4fe1a1afe/website/common/locales/en/limited.json#L153

https://github.com/HabitRPG/habitica/blob/257a52cba1712f4fbc5b0265fd03dbd4fe1a1afe/website/common/script/content/hatching-potions.js#L268-L303

https://github.com/HabitRPG/habitica/blob/257a52cba1712f4fbc5b0265fd03dbd4fe1a1afe/website/common/script/libs/getItemInfo.js#L85-L97

This seems to be the offending code:
```
    _addlNotes: t('eventAvailability', {
      date: t(`dateEnd${pot._season}`),
    }),
```
```
	notes: `${item.notes(language)} ${item._addlNotes(language)}`,
```

`pot._season` is not always defined. When it doesn't have a value, `_addlNotes` shouldn't be populated.
When `_addlNotes` doesn't have a value, the `notes` attribute shouldn't be populated.

MynahMarie:
Hi! I'm a new contributor but I'd like to have a try at solving this issue. Let me know if that's ok 😃

Alys:
@MynahMarie Sure, thank you! Please go ahead and post here if you have any questions.

PR #11806, opened 2020-02-02
Fix errors in _addlNotes data

MynahMarie:
Fixes Issue #11794

### Changes
To fix this issue, I followed what was mention in the posted issue here:

> pot._season is not always defined. When it doesn't have a value, _addlNotes shouldn't be populated.
> When _addlNotes doesn't have a value, the notes attribute shouldn't be populated.
> 

In `hatching_potions.js`, I'm adding a check in the `_addlNotes` field to first check if `pot._season` is defined. If not, I'm setting the value of this field to `null`. Then, in `getItemInfo.js`, I first check to see if the `addlNotes` field is defined before populating the `notes` field.

I noticed there was a similar problem with the wacky potions when the value `_PENDING_` is given to the `_season` field so I added a similar check for this as well.

It might be a bit of a simplistic solution but just checking to see if this is the right direction to solve this problem? Or do we need to have a different value then `null` being passed in case `pot._season` is not defined?

---- 
UUID: 27395fd0-198c-49a1-8ebb-64c963b6a192


MynahMarie:
Some tests are failing but these failures don't seem to be because of my changes... Please let me know if you think this is a problem with my local install and will fix. Thank you!

Alys:
@MynahMarie The failing lint test might be because the package-lock.json file was committed in this PR. When it changes locally, it doesn't need to be committed or pushed. It contains a change to the lint package versions so that's probably causing lint to have different behaviour than normal. If you remove it from the PR, I'd guess that the tests are likely to succeed. Let us know if you need advice about that! :)

MynahMarie:
@Alys I removed `package-lock.json` from the PR but now all the tests are failing... I don't think the linting tests fail because of package-lock, on my local install I get the same errors when running the tests, both in my branch and in the develop branch.

The problem seems to be a specific eslint rule that doesn't allow `require` statements in some of the files... If you think it's a problem with the tests, or the eslint configuration, I can open another issue to work on this since I think the fix would be pretty easy... But maybe there's something I'm not understanding correctly about the guidelines for making a PR? If that's the case, I'm very sorry... It should become easier once I'm more familiar with the code base.

EDIT: I added back the `package-lock.json` file in this branch to get back to the results I previously had with the tests. I thought that was better then having them all fail... I'd love some advice on this issue... 

Here are some screenshots of the errors I get locally:

My branch:
Screenshot 2020-02-04 10 24 56

Develop branch:
Screenshot 2020-02-04 10 29 48

paglias:
@Mahendran-Nadesan hi, the errors are definitely unrelated to this PR so I wouldn't worry about them, I'll give it a look in the next few days. That being said @Alys is right and changes to `package-lock.json` should not be committed

MynahMarie:
@paglias  @Alys Ok got it, thank you! Now I see my mistake in my first commit where I commited the changes for the package-lock.json. Any suggestions on how to fix this? Since it's a very short PR, maybe I should close this one and start fresh on a new branch?

Alys:
@MynahMarie I think this process should remove the package-lock.json file:

1. Check out the branch that you made this PR from if it's not already your active branch (`git checkout fix_addlNotes` probably)
2. Revert the file to the original version: `git checkout upstream/develop -- package-lock.json` 
3. Commit the change: `git commit -m "Removed package-lock.json from pull request"
4. Push the change to the PR: `git push origin fix_addlNotes` (you might need to do a force push, which is often described as something to avoid but it's usually okay if you're the only person working on a PR) 

Note that using `upstream` and `origin` in those commands assumes that you followed the process described in "Fork and Clone Habitica" in [Setting up Habitica Locally](https://habitica.fandom.com/wiki/Setting_up_Habitica_Locally). If you didn't, your remotes might have different names and you'd need to adjust those commands accordingly.

Anyway, give that a try and if it doesn't work, it's fine to open a new PR.

MynahMarie:
@Alys Thank you so much! That worked like a charm and I learned something new :) And now all the checks are passing :)

paglias:
Sorry for the delay here but I won't have time to review PRs until the 15th

MynahMarie:
@paglias No problem! I'll check back in after the 15th :)

website/common/script/content/hatching-potions.js
```diff
@@ -290,9 +290,9 @@ each(wacky, (pot, key) => {
     notes: t('hatchingPotionNotes', {
       potText: pot.text,
     }),
-    _addlNotes: t('eventAvailability', {
+    _addlNotes: pot._seasont && pot._season !== '_PENDING_' ? ('eventAvailability', {
```
paglias: should this be `pot._season` without an ending `t`?

website/common/script/libs/getItemInfo.js
```diff
@@ -72,7 +72,7 @@ export default function getItemInfo (user, type, item, officialPinnedItems, lang
       itemInfo = {
         key: item.key,
         text: i18n.t('potion', { potionType: item.text(language) }),
-        notes: item.notes(language),
+        notes: item._addlNotes ? item.notes(language) : null,
```
paglias: is it necessary to check for the existence of `_addlNotes` here since it doesn't seem to be used ? @Alys

paglias (changes requested):
left a couple of comments :) let me know if you have any question

paglias:
Ah it looks like there's a conflict to be solved

website/common/script/libs/getItemInfo.js
Alys: `_addlNotes` is definitely used for some items so I think the check is needed.

Alys: Although I haven't looked through the new code in a lot of detail and I might be misunderstanding the question, sorry. :(

Here's an example of where it's used in case that helps:
https://github.com/HabitRPG/habitica/blob/0e590c362318e6e5b8fc525df60ecaa77f64c02d/website/common/script/content/hatching-potions.js#L134-L137

website/common/script/content/hatching-potions.js
MynahMarie: Yes you are right, I added the `t` :)

MynahMarie:
I updated my local install and integrated the changes from the develop branch to fix the merge conflict. This takes out one of the changes I made in this PR but the new solution seems to work much better (but it looks like it makes `pot._season` variable obsolete?) However, now reviewing the output I noticed another similar problem in the output from a different part of the code...

Screenshot 2020-02-18 11 30 57

This error comes from:
https://github.com/HabitRPG/habitica/blob/c6f6722af74bd3af644f3818c30b84444ae6ea4f/website/common/script/content/hatching-potions.js#L161-L168

Is it possible that it comes from the newly integrated changes? Seems like the problem is only from this 1 value.

website/common/script/libs/getItemInfo.js
MynahMarie: Seems like this change is now obsolete since I don't see anymore the error I saw when working on it. So I removed it... Let me know what you think.

website/common/script/libs/getItemInfo.js
```diff
@@ -86,7 +86,7 @@ export default function getItemInfo (user, type, item, officialPinnedItems, lang
       itemInfo = {
         key: item.key,
         text: i18n.t('potion', { potionType: item.text(language) }),
-        notes: `${item.notes(language)} ${item._addlNotes(language)}`,
+        notes: item._addlNotes ? `${item.notes(language)} ${item._addlNotes(language)}` : null,
```
MynahMarie: I removed this change as well for the same reason as above...

MynahMarie:
@Alys @paglias I think there's only one possibly relevant change in this PR since applying the latest changes. Please review and let me know what you think and also if I should try to fix that last error I see or if that's something best addressed with the person who did the latest changes on that file...

paglias:
The error you see if because the string with key `june2018` doesn't exist, @Alys how do we want to handle that?

Alys:
We'll need to add `june2018` to the locales/en/limited.json file. It must have been there in the past because it's in some of the other languages:

website/common/locales/de/limited.json:164:    "june2018": "Juni 2018",
website/common/locales/es/limited.json:161:    "june2018": "Junio 2018",
website/common/locales/fr/limited.json:164:    "june2018": "Juin 2018",
etc

paglias:
@MynahMarie if you add the missing `june2018` string I'll give this a final look and then merge :)

paglias:
@MynahMarie  thanks for this! I'm merging the PR and awarding your contributor tier.

MynahMarie:
@paglias Sorry for the late reply! I was travelling and didn't have the chance to check back on this until now.

Thank you for merging and for the contributor tier! Definitely will continue to help out as I can :)

PR #11826, opened 2020-02-05
[Fix #11824] Show member Display Name if Username is blank

### LINKED ISSUE #11824: Users who do not have usernames show up only as @ in challenges
In old challenges, there are sometimes participants who have never confirmed a username. These users appear in the "view progress" tool as "@", and also in the "End Challenge" username selection.

image
image

Example of a challenge with this issue: https://habitica.com/challenges/8aa90845-80ad-4c1b-ac93-e28d9c55c647

Reported by citrusella in Report a Bug.

tuttiq:

Fixes https://github.com/HabitRPG/habitica/issues/11824

### Changes

Added a method `memberName` to `memberSearchDropdown` that checks if `member.auth.username` is blank, it will return `member.profile.name` instead.

I preferred to use a method rather than an `OR` or ternary operator to include the logic that if it's a Username it will append the `@` symbol before, if it's a Display Name it won't.

### Screenshots (reproduced locally)

Scenario: User `@tuttiq` has username. User `kayoreena` (display name) doesn't have username.

#### Before

_Challenge page_

_End challenge modal_

#### After

_Challenge page_

_End challenge modal_

----
UUID: 7e8107e4-c309-43cc-93c9-50311506e53f

### DIFF
--- a/website/client/src/components/members/memberSearchDropdown.vue
+++ b/website/client/src/components/members/memberSearchDropdown.vue
@@ -62,7 +62,7 @@ export default {
       this.$emit('member-selected', member);
     },
     memberName (member) {
-      if (member.auth.local.username) {
+      if (member.auth.local && member.auth.local.username) {
         return `@${member.auth.local.username}`;
       }
       return member.profile.name;
--- a/website/client/src/components/members/memberSearchDropdown.vue
+++ b/website/client/src/components/members/memberSearchDropdown.vue
@@ -19,7 +19,7 @@
       :key="member._id"
       @click="selectMember(member)"
     >
-      @{{ member.auth.local.username }}
+      {{ memberName(member) }}
     </b-dropdown-item>
   </b-dropdown>
 </template>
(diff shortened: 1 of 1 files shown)


paglias:
Sorry for the delay but I won't have time to review PRs until the 15th

website/client/src/components/members/memberSearchDropdown.vue
```diff
@@ -61,6 +61,12 @@ export default {
     selectMember (member) {
       this.$emit('member-selected', member);
     },
+    memberName (member) {
+      if (member.auth.local.username) {
```
paglias: I think the check here should be `member.auth.local && member.auth.local.username` because old users that signed up using facebook or google and without a set username might have an empty `auth.local` object

paglias (changes requested):
left one small comment @tuttiq but it's looking good :)

website/client/src/components/members/memberSearchDropdown.vue
tuttiq: Thanks for the heads up, updated according to your suggestion!

paglias:
@tuttiq  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11856, opened 2020-02-12
Update user count and write text consistently. fixes #11790

### LINKED ISSUE #11790: inconsistent user count on home page
Mentioned this in the Bug Guild and was told by Alys to create an issue.

The user count on the Habitica home page is inconsistent - at the top it says 3 million, but at the bottom it says 2 million. 

Top of page:

habiticahometop

Bottom of page: 

habiticahomebottom

Pretty sure this is because the relevant line hard-codes in "2,000,000" instead of using the `userCountInMillions` variable. 

https://github.com/HabitRPG/habitica/blob/02d6b77a0364e16013862abb3cde0d51ecf412d7/website/common/locales/en/front.json#L324

neepsnikeep:
The user count is currently 4,671,530 according to Alys. So the million count goes up to 4 and the texts mentioning the number of users is supposed to display this count. There have been two texts and the count is only in one of them with the other hardcoded. Seems like a good thing to not actually calculate the number of users for performance reasons. This fixes #11790

----
UUID: @neepsnikeep

### DIFF
--- a/website/client/src/components/static/home.vue
+++ b/website/client/src/components/static/home.vue
@@ -289,7 +289,7 @@
       <div class="container featured">
         <div class="row text-center">
           <h3 class="col-12">
-            {{ $t('joinMany') }}
+            {{ $t('joinMany', {userCountInMillions}) }}
           </h3>
         </div>
         <div class="row">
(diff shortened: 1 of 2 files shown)


paglias:
@neepsnikeep thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11857, opened 2020-02-12
fix for three dots menu getting covered up by footer issue: #11227

### LINKED ISSUE #11227: website footer covers task three-dots menu (z-index problem?)

actuday6418:

Fixes #11227

### Changes
[//]: The z-index of the footer was lowered from 17 to 10

[//]: #771c60c8-ae6d-4767-b577-36bb98bf72e4

----
UUID:

### DIFF
--- a/website/client/src/components/appFooter.vue
+++ b/website/client/src/components/appFooter.vue
@@ -326,7 +326,7 @@
   .footer-row {
     margin: 0;
     flex: 0 1 auto;
-    z-index: 17;
+    z-index: 10;
   }
 
   footer {


paglias:
Thanks @actuday6418 ! This seems to be working well and will go live at the end of the week if no issues arise in the meantime :) If you have an Habitica account just comment with your username or user id and I'll be happy to award a contribution tier!

PR #11832, opened 2020-02-08
improvements to apidocs comments

Alys:
This makes various changes to some apidocs comments for clarity, accuracy, consistency, grammar, or formatting.

No code was modified.

I'm happy to explain my reasons for any of this if desired.

If anyone knows of other apidocs comments that could be improved, comment here and I'll see about including them too, if you comment before this is reviewed.

PR #11859, opened 2020-02-12
Fixing some typos, etc. in the API doc

citrusella:

I reported an issue with some of the API doc having parts that looked weird in Report a Bug back in November and was asked by Alys if I could do this (and that if I didn't she could make an issue) but it's been this long (goodness!) and an issue was never created, so this doesn't close any specific one.

### Changes
My initial message was my noticing that a few 401 errors like the one at https://habitica.com/apidoc/#api-Task-CreateChallengeTasks so I initially had set out to plan on fixing those instances, but then I noted a lot of headers on example blocks didn't have text and so were displaying a bit from the start of the example instead and decided to try adding titles to those example blocks (I would have tried to make these consistent, but the existing ones are a bit inconsistent (some say stuff like "Success-Example" and others say things like "example return" and still others describe the exact situation the example is showing like "not enough gems" or something), and I didn't know if I should go through and try to make them consistent without asking if or how I should do it, so I just ended up trying to follow the doc's lead where possible and then settled into a rhythm...).

While combing the doc, I changed a few other things that looked wrong, like adding a descriptive title to one in the user part that didn't have one, putting all the text in a few titles that were split in half onto a newline so it didn't appear with the latter part as the name and the prior part as the end of the URL (i.e. like https://habitica.com/apidoc/#api-User-UserAllocate), a typo or two (i.e. "bed" to "bad"), etc.

(There's one thing I'm unsure of: At https://habitica.com/apidoc/#api-Member-GetMemberAchievements the "label" part of the success table looked wrong, so I found that removing the asterisks from the success stuff fixed it (but also made a lot of other fields appear in the doc, which makes me unsure whether my fix would be desired (i.e. were those hidden on accident or on purpose)?)

I... *think* I've described all the changes I made but please ask if it seems like I haven't or if there's anything else I should add/do.

----

### DIFF
--- a/website/server/controllers/api-v3/tasks/groups.js
+++ b/website/server/controllers/api-v3/tasks/groups.js
@@ -112,8 +112,8 @@ api.getGroupTasks = {
 };
 
 /**
- * @api {post} /api/v3/group/:groupId/tasks/:taskId/move/to/:position Move a group task to
- * a specified position
+ * @api {post} /api/v3/group/:groupId/tasks/:taskId/move/to/:position
+ * Move a group task to a specified position
  * @apiDescription Moves a group task to a specified position
  * @apiVersion 3.0.0
  * @apiName GroupMoveTask
--- a/website/server/controllers/api-v3/auth.js
+++ b/website/server/controllers/api-v3/auth.js
@@ -219,7 +219,7 @@ api.updateUsername = {
 };
 
 /**
- * @api {put} /api/v3/user/auth/update-password
+ * @api {put} /api/v3/user/auth/update-password Update password
  * @apiDescription Update the password of a local user
  * @apiName UpdatePassword
  * @apiGroup User
--- a/website/server/controllers/api-v3/coupon.js
+++ b/website/server/controllers/api-v3/coupon.js
@@ -25,7 +25,7 @@ const api = {};
  *
  * @apiSuccess {String} Coupons in CSV format
  *
- * @apiSuccessExample {String}
+ * @apiSuccessExample {String} Example success:
  * code,event,date,user
  * GJG4-WEA4-QX3P,wondercon,1476929528704,user-uuid
  * TT32-EYQA-JPBT,wondercon,1476929528705,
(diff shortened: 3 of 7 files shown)


Alys:
@citrusella Thank you for this!

I forgot we'd talked about it though. :(  I made a PR a little while back that might conflict with this (https://github.com/HabitRPG/habitica/pull/11832) but I've just posted to it to ask that yours be reviewed first, and then I'll fix any conflicts that might be in mine, although from a quick glance at your changes, I hadn't found most of the problems you fixed! :)

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -112,8 +112,8 @@ api.getGroupTasks = {
 };
 
 /**
- * @api {post} /api/v3/group/:groupId/tasks/:taskId/move/to/:position Move a group task to
- * a specified position
+ * @api {post} /api/v3/group/:groupId/tasks/:taskId/move/to/:position
+ * Move a group task to a specified position
```
Alys, line 116: Interesting! Does breaking that `@api` line at the start of the text instead of in the middle fix the problem of only the second part of the text appearing in the left-hand column of the apidocs page?

If so, that's better than my fix of putting it all on one line and disabling the lint max-length check. :) https://github.com/HabitRPG/habitica/pull/11832/files#diff-754aecadb3f23b61d07ebc934534a697

citrusella: Yes, it does! I hadn't realized there was a max-length check until it tripped an error (I had originally put those all on the first line with the URL), and while thinking of how to resolve it without just getting rid of that change, I wondered about that and tested it out to make sure it worked. I didn't occur to me to disable the lint check (and even if it did I probably would have avoided it XD).

citrusella:
Ah, I didn't realize you'd made an API doc PR, too! And I made sure to look to see what API-related PRs were opened semi-recently to make sure I wasn't going to make changes that were already done or anything. ._.

citrusella:
Added a commit fixing a missing quotation mark in https://habitica.com/apidoc/#api-User-UserPurchase (hadn't caught it on my first pass-through but figured I'd add it since I noticed it when trying to help someone in Aspiring Comrades who was trying to use it).

paglias:
thanks @citrusella ! I've merged the PR and noted it in your profile

PR #11862, opened 2020-02-13
Resolves #11861 — Makes "quest already started" error descriptive

### LINKED ISSUE #11861: clarify messages shown when a quest has already started
The message below is used when four events occur:
"Your party is already on a quest. Try again when the current quest has ended."
https://github.com/HabitRPG/habitica/blob/e6ba1ccca72221a80e6e869bb2b87e0ad8faaafe/website/common/locales/en/quests.json#L104

Those events are when you try to:
1. issue an invitation for a new quest while your party is on a quest
2. force-start the current quest that has already started
3. accept the quest invitation for the quest that's already started
4. reject the quest invitation for the quest that's already started

The message is very appropriate for event 1 and that shouldn't be changed.

It's less appropriate for event 2. Better would be something like: **"That quest has already started."** (No need for "try again" because your intention was to start the current quest.)

The message is confusing for events 3 and 4. Better would be something like: **"That quest has already started."** Maybe it could also contain an encouraging and friendly message about waiting for your party's next quest.

The strings for the new messages should be placed next to `questAlreadyUnderway` to keep similar strings together.

You can find where each of those four events are coded by using `git grep questAlreadyUnderway` - that will show you the code that needs to be changed to use the new strings, and the tests that will need to be adjusted. All those changes should be simple so I'm marking this as "good first issue" so that a new contributor can try it.

ianrandmckenzie:
Fixes #11861 making "Quest already started" error messages more descriptive of the context in which the error is being displayed. I attempted to use Google translate for the non-English ones that are more complete, but as I went through the motions of doing that, I found there was still some English in a lot of the translated files. I decided to just make them all English because I assumed it's protocol for Habitica to have native speakers of the language do the translations. I did, however, try my best at my pirate dialect.

### Changes
Adds applicable translations to all locales, changes which translations are used for error popup messages, and updated tests to reflect these changes.

----
UUID: 9bbf6561-a894-4c31-856e-eafe79f1ce8b


paglias:
thanks @ianrandmckenzie ! The code looks good but as explained in the README [here ](https://github.com/HabitRPG/habitica/blob/develop/website/common/locales/README.md) only the english locales should be changed as the others are handled automatically, could you remove the changes from non english languages? 

I'll merge it then, thanks!

ianrandmckenzie:
@paglias I'll do this tomorrow. Because I'm creating new translation variables should I just have empty strings or remove them entirely?

Alys:
@ianrandmckenzie The strings should be removed entirely from the non-"en" locales files.

If you're not sure how to remove changes from a PR, this comment might help:
https://github.com/HabitRPG/habitica/pull/11806#issuecomment-582156012

So for example, this is one of the commands you'd do:
`git checkout upstream/develop --  website/common/locales/hu/quests.json`

You might find it faster to do this to remove ALL the locales files:
`git checkout upstream/develop --  website/common/locales/`
(I _think_ that will work) and then add back just the english one:
`git add website/common/locales/en/quests.json`
and then `git commit`
I think that will work but if it all seems to go horribly wrong, don't worry, your original changes will still be saved and accessible. :)

ianrandmckenzie:
Hey @Alys I fixed up the non-en locales. You can go ahead and tag this as ready for review.

paglias:
@ianrandmckenzie  thanks for this! I'm merging the PR and awarding your contributor tier.

paglias:
@ianrandmckenzie  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11866, opened 2020-02-15
Closing droppable menu after clicking an item fixes #11833

### LINKED ISSUE #11833: navigation menu leaves ghost image visible after clicking on a menu item
This bug started about a day ago, so it's from a recent change.

JohnJSal (624e48a3-cbea-4bbe-ae44-ca9183f49626) reports:

"when I go to a menu item that has a drop-down list (e.g. Shops, Guilds, Challenges) and click on one of the list items (e.g. Market, My Guilds, Discover Challenges), it leaves a 'ghost' image of the drop-down list. This doesn't seem to happen if you are already on that particular main page (e.g. Challenges and you click one of the Challenge items), but it does happen when you go to a different page (e.g. you're on Challenges and then click 'Market' in the Shops drop-down). This last aspect makes it seem similar to the other problem that had been happening with the drop-down menu getting stuck.

"[Below is] a screenshot of what happens. Notice the Challenge drop-down list after I clicked 'Discover Challenges' from the Guilds page.

"... it does not persist. It goes away when the cursor is moved away. I also want to add that it happens even when you click the main menu category (Inventory, Shops, Guilds, and Challenges) and not just a drop-down menu item.

"... this issue doesn't seem to happen when clicking on the 'Help' menu or its drop-down items. The drop-down menu does remain open until I move the cursor away though. Is that intended, or should it immediately disappear?"

The Help menu is configured a bit differently than the others, which explains the difference in behaviour.

The bug was confirmed by geolojess, shanaqui, and me.

The desired fix is that the menu should close as soon as you've clicked on it.
~I guess it could be fixed by either keeping the menu looking normal after you've clicked on it until you've moved away or making the menu close as soon as you've clicked on it. I've marked this as "on hold - needs design" for a design-related decision to be made about that. @Tressley~

image

artur-borys:

Fixes #11833

### Changes
This pull request changes the code in the menu component in the `closeMenu()` method.
Previously the droppable menu didn't close on desktop after clicking on item because code that was supposed to close the menu was ran only in the mobile state of the page. Right now, menu closes on both mobile and desktop states of the page. Previously the code also changed the class list of the element and it's inline style without utilizing an already existing method `closeDropdown()`, so this change also uses that method.

Before:
before-change
After:
after-change

You've got to believe in it working, but if you don't - I can record a video later.

----
UUID: 6965e69b-a0e3-4af6-8765-bb14e0635e65


paglias:
thanks @artur-borys ! This is working well, I've notated the PR towards your second tier

PR #11870, opened 2020-02-16
Reward with negative cost can no longer be created, fixes #11855

### LINKED ISSUE #11855: Negative rewards can reduce health or kill you immediately
Reported via email for iOS and web, confirmed by me on web.

If you create a reward with a negative cost (e.g. -10 gold), it still removes gold in the same way as a normal reward, but it seems that it doesn't check whether you have enough gold to purchase it. If you do purchase it and your gold goes into the negative, it then starts taking away health. If the gold reward is larger than -50g, it just kills your avatar immediately.

SabreCat speculated that the behaviour stems from an old (long unused) mechanic whereby if you couldn't afford rewards, you could trade health for them instead of gold. Normally the check that you have enough gold to purchase the reward prevents this from happening.

The desired fix is to disallow negative rewards, and also to ensure that if someone *does* have such a reward already, the usual check that they have enough gold is performed.

hamboomger:
Fixes #11855

### Changes
Rewards can no longer be created(or updated) with negative cost.
All the existing rewards with negative cost will no longer damage user character in case there's not enough gold to buy it.
After user clicks one time on reward with negative cost, cost value sign flips to positive (with same number). This behaviour is rather a workaround in order to pass validation, than some design decision.  

----
UUID: 

### DIFF
--- a/migrations/tasks/rewards-flip-negative-costs.js
+++ b/migrations/tasks/rewards-flip-negative-costs.js
@@ -1,0 +1,54 @@
+// @migrationName = 'RewardsMigrationFlipNegativeCostsValues';
+// @authorName = 'hamboomger';
+// @authorUuid = '80b61b73-2278-4947-b713-a10112cfe7f5';
+
+/*
+ * For each reward with negative cost, make it positive
+ * by assigning it an absolute value of itself
+ */
+
+import { Task } from '../../website/server/models/task';
+
+async function flipNegativeCostsValues () {
+  const query = {
+    type: 'reward',
+    value: { $lt: 0 },
+  };
+
+  const fields = {
+    _id: 1,
+    value: 1,
+  };
+
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
+    // eslint-disable-next-line no-await-in-loop
+    const rewards = await Task
+      .find(query)
+      .limit(250)
+      .sort({ _id: 1 })
+      .select(fields)
+      .lean()
+      .exec();
+
+    if (rewards.length === 0) {
+      break;
+    }
+
+    const promises = rewards.map(reward => {
+      const positiveValue = Math.abs(reward.value);
+      return Task.update({ _id: reward._id }, { $set: { value: positiveValue } }).exec();
+    });
+
+    // eslint-disable-next-line no-await-in-loop
+    await Promise.all(promises);
+
+    query._id = {
+      $gt: rewards[rewards.length - 1]._id,
+    };
+  }
+
+  console.log('All rewards with negative values were updated, migration finished');
+}
+
+export default flipNegativeCostsValues;
--- a/website/common/script/ops/scoreTask.js
+++ b/website/common/script/ops/scoreTask.js
@@ -341,12 +341,7 @@ export default function scoreTask (options = {}, req = {}, analytics) {
     // Don't adjust values for rewards
     delta += _changeTaskValue(user, task, direction, times, cron);
     // purchase item
-    stats.gp -= Math.abs(task.value);
-    // hp - gp difference
-    if (stats.gp < 0) {
-      stats.hp += stats.gp;
-      stats.gp = 0;
-    }
+    stats.gp -= task.value;
   }
 
   req.yesterDailyScored = task.yesterDailyScored;
--- a/website/common/script/ops/scoreTask.js
+++ b/website/common/script/ops/scoreTask.js
@@ -189,6 +189,37 @@ function _updateCounter (task, direction, times) {
   }
 }
 
+function _lastHistoryEntryWasToday (lastHistoryEntry, user) {
+  if (!lastHistoryEntry || !lastHistoryEntry.date) {
+    return false;
+  }
+
+  const { timezoneOffset } = user.preferences;
+  const { dayStart } = user.preferences;
+
+  // Adjust the last entry date according to the user's timezone and CDS
+  const dateWithTimeZone = moment(lastHistoryEntry.date).zone(timezoneOffset);
+  if (dateWithTimeZone.hour() < dayStart) dateWithTimeZone.subtract(1, 'day');
+
+  return moment().zone(timezoneOffset).isSame(dateWithTimeZone, 'day');
+}
+
+function _updateLastHistoryEntry (lastHistoryEntry, task, direction, times) {
+  lastHistoryEntry.value = task.value;
+  lastHistoryEntry.date = Number(new Date());
+
+  // @TODO remove this extra check after migration
+  // has run to set scoredUp and scoredDown in every task
+  lastHistoryEntry.scoredUp = lastHistoryEntry.scoredUp || 0;
+  lastHistoryEntry.scoredDown = lastHistoryEntry.scoredDown || 0;
+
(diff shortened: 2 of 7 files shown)


website/server/models/task.js
```diff
@@ -96,7 +96,17 @@ export const TaskSchema = new Schema({
     validate: [v => validator.isUUID(v), 'Invalid uuid for task tags.'],
   }],
   // redness or cost for rewards Required because it must be settable (for rewards)
-  value: { $type: Number, default: 0, required: true },
+  value: {
```
paglias, line 99: Unfortunately adding the validator here will cause existing rewards with a negative to be un-saveable due to the new check that would also fails. This will happen when you try to update the rewards or even when scoring it. Could you instead move the check to only happen when the reward is being update or created? Maybe directly in the controller function

website/common/script/ops/scoreTask.js
```diff
@@ -189,6 +189,37 @@ function _updateCounter (task, direction, times) {
   }
 }
 
+function _lastHistoryEntryWasToday (lastHistoryEntry, user) {
```
paglias, line 192: are the changes in this file expected?

paglias (changes requested):
Hi @hamboomger I've left a couple of comments and it also looks like there's a conflict with your changes, could you give it a look? Thanks and let me know if you have any question

website/common/script/ops/scoreTask.js
hamboomger: No, this commit is not related to the issue, it's just something I stumbled upon. There is no change in logic, just (I hope) improvement in readability. But I can discard this commit if you want or make a separate branch for refactoring

paglias: ah okay, let's leave it here since it's definitely and improvement!

website/common/script/ops/scoreTask.js
```diff
@@ -333,13 +341,12 @@ export default function scoreTask (options = {}, req = {}) {
   } else if (task.type === 'reward') {
     // Don't adjust values for rewards
     delta += _changeTaskValue(user, task, direction, times, cron);
+    // Since historically some users can have rewards with negative cost (which will cause
+    // validation failure), in order to 'buy' a reward and update it in database we need to ensure
+    // the reward cost is a positive number
+    task.value = Math.abs(task.value);
```
paglias: what do you think of this @Alys ? Basically if you have a reward with a negative value it'll be treated as if the value is positive, so a value of -5 will be treated as +5

website/server/models/task.js
paglias: I see that in the process of scoring a task you automatically convert the value to a positive one so in that case the validation error should not occur but only when updating the reward and keeping a negative value

hamboomger: Yes, that's why I've made a change that flips negative reward value into a positive with same number before save. Maybe it would be even better to create a migration script which changes all user rewards with negative values into rewards with positive value?

hamboomger: Oh I see, I'll check that behavior and update commit when I got home

paglias: At this point I think a migration might be the best way to solve this, what do you think @Alys?

hamboomger:
Yep, I actually wanted to propose this :)

website/server/models/task.js
hamboomger: Yep, I actually wanted to propose this :)

Alys: I agree that a migration would be good!

paglias: let's go with the migration then! @hamboomger you can see an example of how to write a migration here https://github.com/HabitRPG/habitica/blob/develop/migrations/users/bulk-email.js:

- a new file with `date_name.js` would go into `/migrations`
- you can search for all users or a part of them tweaking the query here https://github.com/HabitRPG/habitica/blob/develop/migrations/users/bulk-email.js#L29 (the only part that has to stay is the `migration: { $ne: MIGRATION_NAME }`
- and finally update the users here https://github.com/HabitRPG/habitica/blob/develop/migrations/users/bulk-email.js#L14

Let me know if you have any question or need help with the migration

migrations/tasks/rewards-flip-negative-costs.js
```diff
@@ -0,0 +1,57 @@
...
+import { Task } from '../../website/server/models/task';
+
+async function flipNegativeCostsValues () {
+  const query = {
+    type: 'reward',
+    value: { $lt: 0 },
+  };
+
+  const totalCount = await _getTotalCount(query);
+  console.log(`Total documents to be changed: ${await Task.find().exec()}`);
```
paglias: can you remove this line? Otherwise it'll try to load all tasks in the database :)

migrations/tasks/rewards-flip-negative-costs.js
```diff
@@ -0,0 +1,57 @@
...
+  if (totalCount === 0 || totalCount > 1) {
+    return;
+  }
+
+  let skip = 0;
+  const limit = 200;
+
+  while (skip < totalCount) {
```
paglias: skip and limit work but tend to have some serious performance problems when delaing with large number of documents, what we usually do is sort by `_id` and then pick only the results above the last `_id`, you can see an example here https://github.com/HabitRPG/habitica/blob/develop/migrations/archive/2020/20200131_habit_birthday.js#L71-L86

paglias (commented):
left some comments about the migration!

migrations/tasks/rewards-flip-negative-costs.js
hamboomger: Oopsie daisy, I couldn't run the migration tool properly so I tested how many documents there are at all (it showed 0), it seems I accidentally left it this way

website/server/models/task.js
hamboomger: @paglias I actually ran into trouble which is described in the related issue, could you take a look at it pls?

paglias: sorry again for the delay, but I answered there, if you can't find a solution don't worry and we'll handle the migration

migrations/tasks/rewards-flip-negative-costs.js
hamboomger: @paglias I pushed new changes, can you take a look pls?

paglias: @hamboomger thanks for the update and yes I'll give a look at all updated PRs tomorrow :) Unfortunately we're a small team, I'm the only one reviewing PRs and I don't work full time on this project so I try to review PRs every Monday. Thanks again for the PRs, it's very appreciated :)

paglias (approved):
This is looking good @hamboomger , thanks! I've added a todo to my list about the migration and will run it (and merge the PR) as soon as possible although it'll probably have to wait until next week as these days we're going to be busy with an important database maintenance

paglias:
@hamboomger  thanks for this! I'm merging the PR and awarding your contributor tier.

---- 

Sorry it took so long but we finally made it :) It'll go live by the end of the week with our next weekly deployment

hamboomger:
@paglias yeey, wait for another 2 PR's this week :)

PR #11883, opened 2020-02-20
Adventure Guide Prep

SabreCat:

Implements the changes from "Adventure Guide Pre-reqs" on the project board

### Changes

* User can now receive random drops prior to attaining level 3. Notifications and modals related to the level 3 unlock are removed.
* When the user scores their second task (determined as the first positive task score after receiving the task score achievement), they receive a random hatching potion and a random pet egg. A modal notification shows them what they got and leads them to the Inventory page.
* An analytics event is now logged when the user finishes their onboarding checklist.
* Analytics events are logged for item drops, pet hatching, and pet feeding during the user's first week.
* Tasks created by the onboarding process (prior to user's "welcomed" flag being set) no longer record analytics events as if the user had manually created them.


website/client/src/components/achievements/firstDrops.vue
```diff
@@ -0,0 +1,113 @@
...
+    :hide-footer="true"
+  >
+    <div class="text-center">
+      <div
+        class="modal-close"
+        @click="close()"
+      >
+        <div class="svg-icon" v-html="icons.close"></div>
+      </div>
+      <h2 class="mt-3 mb-4">{{ $t('foundNewItems') }}</h2>
```
paglias: this file could use a few v-once for static strings :)

website/client/src/components/achievements/firstDrops.vue
```diff
@@ -0,0 +1,113 @@
...
+  computed: {
+    eggClass () {
+      return `Pet_Egg_${this.$store.state.firstDropsOptions.egg}`;
+    },
+    potionClass () {
+      return `Pet_HatchingPotion_${this.$store.state.firstDropsOptions.potion}`;
+    },
+  },
+  methods: {
+    close () {
```
paglias, line 118: can we clear  `$store.state.firstDropsOptions` once this modal is closed?

website/server/models/userNotification.js
```diff
@@ -34,6 +34,7 @@ const NOTIFICATION_TYPES = [
   'NEW_STUFF',
   'NEW_CHAT_MESSAGE',
   'LEVELED_UP',
+  'FIRST_DROPS',
```
paglias, line 37: can we add a `// unused` comment to the `DROPS_ENABLED` notification here?

paglias (commented):
@SabreCat Left a couple of comments, nothing serious. I haven't checked the design

paglias:
Ah, we might want to add a test for this behavior:

> When the user scores their second task (determined as the first positive task score after receiving the task score achievement), they receive a random hatching potion and a random pet egg. A modal notification shows them what they got and leads them to the Inventory page.

website/client/src/components/tasks/task.vue
```diff
@@ -1068,6 +1069,13 @@ export default {
         }
       }
 
+      if (firstDrops) {
+        Vue.set(user.items.eggs, firstDrops.egg, 0);
```
paglias: Just one quick thing, here should we run Vue.set only if the user doesn’t have any of that egg/potion? Otherwise we reset it to zero even if they already bought one. Should be pretty rare at this stage but still possible

SabreCat: We should only be running the first drops code if the user has never had any potions or eggs. But I can definitely put a safety check in, as I did elsewhere!

website/common/locales/en/pets.json
```diff
@@ -48,11 +48,10 @@
...
   "haveHatchablePet": "You have a <%= potion %> hatching potion and <%= egg %> egg to hatch this pet! <b>Click</b> the paw print to hatch.",
   "quickInventory": "Quick Inventory",
-  "foodText": "food",
-  "food": "Food and Saddles",
-  "noFoodAvailable": "You don't have any Food.",
+  "food": "Pet Food and Saddles",
+  "noFoodAvailable": "You don't have any Pet Food.",
   "noSaddlesAvailable": "You don't have any Saddles.",
-  "noFood": "You don't have any food or saddles.",
+  "noFood": "You don't have any Pet Food or Saddles.",
```
Alys: I think noFood is no longer used. I can't find it in the code anywhere.
(I went looking through the code to answer a Socialite question and then got curious about this part because I didn't recall any place where you'd see your owned Food and Saddles together. :) )

SabreCat: ah right! The Inventory only says "You don't own any of these." Good catch, I'll delete it.

PR #11884, opened 2020-02-20
Enchanted Armoire immediately updates XP stat in header (fixes #11827) and does level-up if needed (partial fix #2433)

### LINKED ISSUE #2433: Failed to Level Up after gaining Experience (XP) from Completing Quest

### LINKED ISSUE #11827: Experience gained from Enchanted Armoire should appear in header stats immediately
When you gain experience from the Enchanted Armoire, your experience bar in the header does not change immediately. You have to click the sync button to make the new experience appear.

The desired behaviour is that as soon as you gain experience from the Armoire, your exerperience bar should increase and thee number to the right of it should change. Clicking the sync button should not be necessary.

mayajosyula:

Fixes #11827, also partial fix for #2433 as the Armoire will now level you up.

### Changes
Currently, if someone gets Experience from the Enchanted Armoire, they have to click the sync button to see their Experience bar update. When they sync, they also see the default XP notification saying "You gained some Experience" which makes it look like they received Experience twice. 

This PR changes three things: 

-  It forces a sync for the user if the Armoire gives Experience. This makes the bar in the header update immediately without the user needing to click the sync button.
-  It calls `updateStats` after the user's Experience is modified. Since there is now a sync, this will level up the user if they have enough Experience to do so. 
-  The sync causes the Armoire XP notification ("You wrestle with the Armoire and gain Experience. Take that!") and the default XP notification ("You gained some Experience") to display at the same time, which again makes it look like the user received Experience twice. It's not possible to get rid of the default notification, so as a workaround, this PR removes the XP amount from the Armoire notification. (The XP amount is still shown by the default notification.)

After the changes, this is what the notifications look like if the Armoire gives Experience: 
testexp

The XP notification problem is discussed in more detail over on the issue thread #11827. (In particular, Alys in the second half of https://github.com/HabitRPG/habitica/issues/11827#issuecomment-587337769 explains the problem/solution probably much better than I can.)

----
UUID: 9261598f-9f1a-44ef-bcfb-dc8bc4a70d86

### DIFF
--- a/website/client/src/store/actions/shops.js
+++ b/website/client/src/store/actions/shops.js
@@ -93,6 +93,10 @@ async function buyArmoire (store, params) {
       timeout: true,
       ...notificationOptions,
     });
+
+    if (isExperience) {
+      await store.dispatch('user:fetch', { forceLoad: true });
+    }
   }
 }
 
--- a/website/client/src/store/actions/shops.js
+++ b/website/client/src/store/actions/shops.js
@@ -78,9 +78,8 @@ async function buyArmoire (store, params) {
     // @TODO: We might need to abstract notifications to library rather than mixin
     const notificationOptions = isExperience
       ? {
-        text: `+ ${item.value}`,
-        type: 'xp',
-        flavorMessage: message,
+        text: message,
+        type: 'success',
       }
       : {
         text: message,
--- a/website/common/script/ops/buy/buyArmoire.js
+++ b/website/common/script/ops/buy/buyArmoire.js
@@ -10,6 +10,7 @@ import {
 import randomVal, * as randomValFns from '../../libs/randomVal';
 import { removeItemByPath } from '../pinnedGearUtils';
 import { AbstractGoldItemOperation } from './abstractBuyOperation';
+import updateStats from '../../fns/updateStats';
 
 // TODO this is only used on the server
 // move out of common?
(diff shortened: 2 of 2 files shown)


Alys:
As a note to myself: when this is merged I should tell the Socialites and Wiki Wizards that the Armoire will cause level-ups when needed, and that the Armoire's notification screenshots in the wiki should be updated. I've labelled this with "notify helpers" to remind me.

paglias:
thanks @agentx999 ! this is working well, noted towards your 4th tier

PR #11886, opened 2020-02-20
Change where members info is retrieved in Group component: Fixes #11737

### LINKED ISSUE #11737: member list should update when viewing different guilds
If you visit a guild and click on its "Member List" badge, you see a list of the guild's members. If you then immediately visit a different guild and click on its badge, you see the SAME list - i.e., the list for the first guild. You have to reload the website to see the second guild's member list.

The desired behaviour is that a reload should not be necessary.

tsukimi2:
Fixes #11737

In Group component, changes the place where members data is retrieved from the load function originally to within the watch group function, so that when group data has changed, the code will update the corresponding members data as well.  This fixes the issue in #11737 where members data is not updated when the guild is changed while the Group component has already been loaded, as detailed in #11737.  So instead of only relying on the load function to retrieve the members data when the Group component is first created, by a watcher on the groups data, it makes the Group component to automatically update the members data when the group data changes (e.g. when the guild is changed.)


website/client/src/components/groups/group.vue
```diff
@@ -509,11 +515,7 @@ export default {
         // Load invites
       }
       await this.fetchGuild();
-      // Fetch group members on load
-      this.members = await this.loadMembers({
-        groupId: this.group._id,
-        includeAllPublicFields: true,
-      });
+
       this.$root.$on('updatedGroup', group => {
```
paglias, line 513: (unrelated but this probably leads to a memory leak as the event listener is never cleared)

website/client/src/components/groups/group.vue
```diff
@@ -486,6 +486,12 @@ export default {
         this.$router.push('/');
       }
     },
+    async group () {
```
paglias: instead of watching the entire group object could we watch just the id? So it's only executed when the group actually changes `async 'group._id'`

paglias (commented):
left one comment, I wonder if instead of using the watcher function we could instead load the members in the `fetchGuild` method that is called every time the guild is changed. Not that the current PR doesn't work but this way we have one less watcher (fetchGuild is already executed based on the route change, see https://github.com/HabitRPG/habitica/pull/11886/files#diff-ef32c4433a696d3ba3b69b9fb8d7b676L480) and an easier to follow logic

paglias:
thanks @tsukimi2 ! Post your Habitica username if you want this noted in your profile!

Also, I noticed that in your PRs @tsukimi2 is the account opening the PRs and writing the comments but commits come from @kazekunGb ?

tsukimi2:
> thanks @tsukimi2 ! Post your Habitica username if you want this noted in your profile!
> 
> Also, I noticed that in your PRs @tsukimi2 is the account opening the PRs and writing the comments but commits come from @kazekunGb ?

@paglias Thanks! My Habitica username is  osiris65.  _[Edit by Allys:  cefa15ee-c1bc-46e2-95ec-bbf579cd2d85 ]_

About my github username not matching between PR opening and commit, thanks for notifying.  I just realized that my git config email has been set incorrectly to another github account that I have not been using for many years already.  I've corrected the config now and shouldn't see the @kazekunGb anymore in the future.

paglias:
welcome to the 2nd tier :)

PR #11889, opened 2020-02-22
Allow group links on challenge details to be opened in a new tab

### LINKED ISSUE #11461: challenge screen does not use a normal guild URL (won't open in new tab)
This issue occurs on My Challenges page, the Discover Challenges page, and the the pages for individual Challenges.

When the Challenge belongs to a Guild or Party, there's a link to the guild/party. The link is implemented with JavaScript instead of using a normal URL. This means that while the link does work if you do a normal click on it, it doesn't work if use a right-click to open it in a new tab. If you try to do that, the new tab contains just a second copy of the Challenge.

Being able to open the Challenge's guild in a new tab while still looking at the Challenge is useful (guild challenges often refer to the guild, or involve chat messages in the guild, or you might want to look for related challenges in the guild, etc).

----

I've been noticing this for a long time but it hasn't bothered me quite enough to take the time to report it, however now there's reports of it in the Report a Bug guild, including suggestions that the link might not even work correctly all the time when you do a normal click on it.

@tseren (8d1bee26-4a44-496d-89b5-1e8778d82c7d):
_"On the webpage of a challenge, there's a line below the creator's name which identifies the guild hosting the challenge. It reads along the lines of Guild: Name-of-Guild-Here, and the name of the guild is a link. Logically, this ought to be a link to the guild's page. However, the link points to the challenge's page, i.e. the page you're already viewing. In detail, this is the URL of the challenge followed by a hash, which I'm assuming is meant as the usual pointer. This has been the way of things for a while. I'm not sure it's a bug per se, but it doesn't seem like it's the correct intention, I think?"_

@Ceran (aabe567b-aa55-42c1-b612-b7699a7ba9b7): 
_"I have noticed that, too. It seems to happen especially with challenges that I have cloned, though I couldn't say if it happens only with those. If I refresh the page, it often correctly leads to the guild, so I kind of assumed it is an issue with my browser. But if it happens to you too, maybe not."_

@Ceran (aabe567b-aa55-42c1-b612-b7699a7ba9b7):
_"I have not payed attention to wether the link shows a hashtag or not. I noticed it happening when I try to open the guild in a new tab by clicking on the link with my mouse wheel, often it will just open the challenge page again. I can't really say wether or not it also happens by just clicking normally to open in the same tab. I'll pay more attention now that I know it's not just a problem for me, and report back when it happens again."_

----

I'll leave this as suggestion-discussion for a few days in case anyone has any comments and then if there's no objections I'll mark it help wanted for the javascript link to be changed to a normal "a href" link with the guild or party URL in it.

SpeedyLom:
Fixes #11461

### Changes

- Replace `b-link` with `router-link` and set the `:to` based on the type of `group` that is related to the challenge

----

### DIFF
--- a/website/client/src/components/groupLink.vue
+++ b/website/client/src/components/groupLink.vue
@@ -1,5 +1,6 @@
 <template>
   <router-link
+    v-if="group"
     :to="toPath"
   >
     {{ group.name }}
--- a/website/client/src/components/groupLink.vue
+++ b/website/client/src/components/groupLink.vue
@@ -1,10 +1,9 @@
 <template>
-  <b-link
-    v-if="group"
-    @click.prevent="goToGroup"
+  <router-link
+    :to="toPath"
   >
     {{ group.name }}
-  </b-link>
+  </router-link>
 </template>
 
 <script>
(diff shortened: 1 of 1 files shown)


website/client/src/components/groupLink.vue
```diff
@@ -1,26 +1,36 @@
 <template>
-  <b-link
-    v-if="group"
```
paglias, line 3: the existing `v-if` here should be kept?

paglias (commented):
left a comment but otherwise looks good @tthorpe88 , thanks!

website/client/src/components/groupLink.vue
SpeedyLom: Good spot. I'll get that added back in.

paglias:
thanks @tthorpe88 ! noted towards your 2nd tier

PR #11905, opened 2020-02-26
API-v4 route added: 'api/v4/faq' fixes #11801

### LINKED ISSUE #11801: move FAQ questions and answers from API v4 content route to dedicated route
_Note for contributors: This issue relates to API version 4, which is used internally by the official website and mobile apps. All third-party tools must still use version 3. You're welcome to work on this issue after it's been marked as help wanted but be careful that your fix changes only [API v4](https://github.com/HabitRPG/habitica/tree/develop/website/server/controllers/api-v4). It must not change [API v3](https://github.com/HabitRPG/habitica/tree/develop/website/server/controllers/api-v3) because that would be a breaking change that would negatively affect third-party tools. Links to the [apidocs](https://habitica.com/apidoc/) are provided here for reference but those docs are for API v3 so they might not be accurate for v4._

---

When you use the [content](https://habitica.com/apidoc/#api-Content-ContentGet) route with the API v4 command below, the content includes FAQ data as well as other data:

`curl -X GET 'https://habitica.com/api/v4/content?language=en'`

Sample FAQ data:

```
"faq":{
	"questions": [
		{
			"question":"I'm confused. Where do I get an overview?",
			"ios":"First, you'll set up tasks...",
			"android":"First, you'll set up tasks...",
			"web":"First, you'll set up tasks..."
		},
		{
			"question":"How do I set up my tasks?",
			"ios":"Good Habits (the ones with a +) are...",
			"android":"Good Habits (the ones with a +) are...",
			"web":"* Good Habits (the ones with a :heavy_plus_sign:) are..."
		}
	],
	"stillNeedHelp": {
		"ios":"If you have a question that...",
		"web":"If you have a question that..."
	}
}
```

We want this FAQ content to be removed from the `content` route and added to a new `faq` route.

This must be done in SEPARATE stages:
1. **Create the new `faq` route.** This is the only action needed to resolve this issue.
2. LATER: Update the official mobile apps and website to use the `faq` route and wait long enough for most app users to upgrade the app.
2. LATER: Delete the faq data from the `content` route. We should be certain that the non-updated mobile apps won't crash when that's done. It's okay if the FAQ screen on non-updated apps doesn't work - the user can update to fix that.

So this issue is only for item 1 above. After item 1 is done, we'll log new issues for the other items.

The new `faq` route should take an optional query parameter that can have one of these values: `ios`, `android`, `web`:
- If the parameter is supplied, then only the questions and the matching answers should be returned, not all three answers for each question. See in the data above how there's different answers for different platforms).
- If the parameter is supplied and one of the items doesn't have a response for that platform, the `web` response must be returned. We'll assume `web` is always populated. See the `stillNeedHelp` item for an example of that.
- If the parameter is supplied and it's not one of the allowed values, an error should be thrown.
- If the parameter is not supplied, the full `faq` object should be returned.

----

_This is being done to slim down the `content` route to improve loading times, especially for the mobile apps. The new `faq` route can be loaded only if the user accesses the FAQ screen._

hamboomger:

Fixes #11801

### Changes
The new **faq** route created for API v4, which is available at **/api/v4/faq** url.
Parameters:

- **language**(optional)
- **platform**(optional), supported values: [`web`, `android`, `ios`]. Parameter behaviour is described in the [issue](https://github.com/HabitRPG/habitica/issues/11801).

----
UUID: 

### DIFF
--- a/website/server/controllers/api-v4/faq.js
+++ b/website/server/controllers/api-v4/faq.js
@@ -1,0 +1,77 @@
+import _ from 'lodash';
+import { query } from 'express-validator/check';
+import { langCodes } from '../../libs/i18n';
+import apiError from '../../libs/apiError';
+import common from '../../../common';
+import { localizeContentData } from '../../libs/content';
+
+const { content } = common;
+const { faq } = content;
+
+const api = {};
+
+function _deleteProperties (obj, keysToDelete, platform) {
+  // if there is no description for specified platform, use 'web' description by default
+  if (obj[platform] === undefined) {
+    delete obj.ios;
+    delete obj.android;
+    return;
+  }
+
+  keysToDelete.forEach(key => delete obj[key]);
+}
+
+function _deleteOtherPlatformsAnswers (faqObject, platform) {
+  const faqCopy = _.cloneDeep(faqObject);
+  const keysToDelete = _.without(['web', 'ios', 'android'], platform);
+
+  _deleteProperties(faqCopy.stillNeedHelp, keysToDelete, platform);
+  faqCopy.questions.forEach(question => {
+    _deleteProperties(question, keysToDelete, platform);
+  });
+
+  return faqCopy;
+}
+
+/**
+ * @api {get} /api/v4/faq Get faq in json format
+ * @apiDescription Does not require authentication.
+ * @apiName FaqGet
+ * @apiGroup Content
+ *
+ * @apiParam (Query) {String="bg","cs","da","de",
+ *                   "en","en@pirate","en_GB",
+ *                    "es","es_419","fr","he","hu",
+ *                    "id","it","ja","nl","pl","pt","pt_BR",
+ *                    "ro","ru","sk","sr","sv",
+ *                    "uk","zh","zh_TW"} [language=en] Language code used for the items'
+ *                                                     strings. If the authenticated user makes
+ *                                                     the request, the content will return with
+ *                                                     the user's configured language.
+ *
+ *
+ * @apiSuccess {Object} data FAQ in a json format
+ */
+api.faq = {
+  method: 'GET',
+  url: '/faq',
+  middlewares: [
+    query('platform')
+      .optional()
+      .isIn(['web', 'android', 'ios']).withMessage(apiError('invalidPlatform')),
+  ],
+  async handler (req, res) {
+    const validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    const proposedLang = req.query.language && req.query.language.toString();
+    const language = langCodes.includes(proposedLang) ? proposedLang : 'en';
+
+    const { platform } = req.query;
+
+    const dataToLocalize = platform ? _deleteOtherPlatformsAnswers(faq, platform) : faq;
+    res.respond(200, localizeContentData(dataToLocalize, language));
+  },
+};
+
+export default api;
--- a/gulp/gulp-content.js
+++ b/gulp/gulp-content.js
@@ -6,7 +6,7 @@ gulp.task('content:cache', done => {
   // Requiring at runtime because these files access `common`
   // code which in production works only if transpiled so after
   // gulp build:babel:common has run
-  const { CONTENT_CACHE_PATH, getLocalizedContent } = require('../website/server/libs/content'); // eslint-disable-line global-require
+  const { CONTENT_CACHE_PATH, getLocalizedContentResponse } = require('../website/server/libs/content'); // eslint-disable-line global-require
   const { langCodes } = require('../website/server/libs/i18n'); // eslint-disable-line global-require
 
   try {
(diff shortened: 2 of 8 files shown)


website/server/controllers/api-v4/faq.js
```diff
@@ -0,0 +1,83 @@
...
+ *                                                     the user's configured language.
+ *
+ *
+ * @apiSuccess {Object} data FAQ in a json format
+ */
+api.faq = {
+  method: 'GET',
+  url: '/faq',
+  middlewares: [
+    query('platform')
```
paglias, line 59: is this the syntax for the newest version of `express-validator`? We're still using an older version

website/server/controllers/api-v4/faq.js
```diff
@@ -0,0 +1,83 @@
...
+      .optional()
+      .isIn(['web', 'android', 'ios']).withMessage(apiError('invalidPlatform')),
+  ],
+  async handler (req, res) {
+    const validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    const proposedLang = req.query.language && req.query.language.toString();
```
paglias, line 67: is toString necessary?

website/server/controllers/api-v4/faq.js
```diff
@@ -0,0 +1,83 @@
...
+    const language = langCodes[proposedLang] !== null ? proposedLang : 'en';
```
paglias: I think `langCodes[proposedLang] !== null` will fail because missing languages will result in `undefined` which is different from `null`. `langCodes[proposedLang] ? ...` should work

paglias (changes requested):
Left a couple of comments. We discussed this internally and we've decided that for now we'll implement this without the caching until we revisit how it's implemented and maybe switch to something more appropriate like redis, could you remove that part?

thanks again for your patience

website/server/controllers/api-v4/faq.js
hamboomger: This syntax already was available for [`5.2` version](https://express-validator.github.io/docs/5.2.0/validation-chain-api.html), which is listed as a dependency in a `package.json`. I also tested it inside of a container and it works like a charm :)

website/server/controllers/api-v4/faq.js
hamboomger: Not sure, copied this from v3 `content` route

hamboomger: @paglias looks like it's unnecessary, works without toString()

website/server/controllers/api-v4/faq.js
hamboomger: @paglias When I'm looking at it right now it's actually even worse, because langCodes is an array, not an object, so I think this way of checking whether value exists is kind of misleading (at least for me as a Java developer). I think I'll better change it to `langCodes.includes(proposedLang)`

hamboomger: I mean it worked for sure, you can see it by the completion of a test which specifically checks this behaviour (`'falls back to English if the desired language is not found'`), but it's better to change it anyways

hamboomger: @paglias I think PR is ready, could you take a look pls here too?

website/common/script/libs/replaceIdsWithTranslations.js
```diff
@@ -0,0 +1,21 @@
+import _ from 'lodash';
+
+/*
+ * Returns a copy of an object where identifiers of translations are
+ * replaced with actual translations in specified language
+ */
+export default function replaceIdsWithTranslations (
```
paglias: Two things about this function:

- in /common we try to keep only code used both on client and server and since this one is only used on the server, could you move it to website/server/libs/i18n.js?
- I think the name of the function is not very clear, although I don't have any different idea right now

website/server/controllers/api-v4/faq.js
paglias: thanks :)

website/server/controllers/api-v4/faq.js
```diff
@@ -0,0 +1,83 @@
...
+
+    const proposedLang = req.query.language && req.query.language.toString();
+    const language = langCodes.includes(proposedLang) ? proposedLang : 'en';
+
+    const { platform } = req.query;
+
+    const dataToTranslate = platform ? _deleteOtherPlatformsAnswers(faq, platform) : faq;
+    const translatedFAQ = replaceIdsWithTranslations(dataToTranslate, language);
+
+    res.set({
```
paglias: given that we're not caching anymore I think this can be simplified by using `res.respond(200, translatedFAQ)`. it'll return json data automatically :)

paglias (changes requested):
@hamboomger left a couple of comments but nothing major and the rest looks good! I'll also try to review any change before the next monday

website/common/script/libs/replaceIdsWithTranslations.js
hamboomger: @paglias how about localizeData in libs/content.js? Or maybe localizeContentData to be more precise

website/server/controllers/api-v4/faq.js
hamboomger: Okay, makes sense 👍

paglias:
sorry again for the delay but this is looking good, it'll go live by the end of the week, I've also noted it towards your 2nd tier!

PR #11906, opened 2020-02-27
Remove max-height for app-header

### LINKED ISSUE #11071: header misformatted in Firefox when browser default font size is reduced
@LillaSmith (b5a69bc4-e325-417f-befe-32e3ce6f0f80) reported: **"MacOS Mojave's latest firefox version renders the header strangely, see on this image [below]. i checked on chrome (same machine), looks goog there"**

shanaqui asked: "does that persist even after a hard refresh? Just wondering if it was a single odd loading or if it looks like that permanently."

LillaSmith replied: **"yes, it still exist after cache clearing. i have checked the inspector tool, it renders the following css attribute for the "app-header": max-height: 10.25rem; if i turn this off in the inspector tool the issue is gone."**

image

I'm asking LillaSmith if there's any extensions that might be causing it and will report back with the answer. I wanted to record all this first though so it can't go missing if we need it.

### LINKED ISSUE #11822: header partially hidden by body of page
_@Redpinkandwhite (7e608437-5403-42e2-81bf-c25cce8f6464) reported:_

"I am running into a small bug where my bar is overlapping my party images, partially cutting off my mana bar. I have tried refreshing the page and closing the tab.

I just tried closing the browser window as well and reloading Firefox. The images are still set under or over the bar. Image below:"

scr

"... I tried the [https://habitica.com/static/clear-browser-data] link and clicked the button. Logged on again. Same problem. 

All of my extensions I have used in the past I disabled before I asked. (Originally, this is what I thought caused the problem). I even uninstalled the add-on I was using in Firefox to manage the extensions. Still have a problem."

_I can reproduce the bug when I zoom out so I asked them to reset their browser's zoom but that didn't help:_

"... I found the View > Reset, but the 'Zoom text only' was not clicked. I checked it and clicked reset. Nothing happened. I unchecked the 'zoom text only' and hit reset again. Nothing.

I just noticed my browser also doing this. Hope it helps gain some insight into what is happening.

One with pictures over the background:
scr

One with pictures under the background:
scr

It tends to always be in the under the background mode, but it will sometimes go into the 'over the background' mode and then revert back to the 'under the background' mode."

_I noticed that their browser tab was pinned, which we know can cause various problems so I asked them to unpin it and clear the zoom again, but that didn't help:_

"The unpin did not help. Same problem.
Browser: Firefox 72.0.2 (64-bit)
OS: Windows 10 version 1903"

DXCanas:

Tossed the REM changes I'd made before for a light-touch fix: removing the max-height. Want to flag that the commit that made the change (53555a0f161a883a9a9bb3c801cf63e36153b8ed) seemed to have applied the CSS with good reason - I just can't figure out what that was.

Fixes #11822 
Fixes #11071

### Changes

Allows sprites to expand the `app-header` container to whatever size they need, rather than being restricted by the root node's font size.

----
UUID: 69f621bd-7d92-4516-aadb-c1764c930c5d

### DIFF
--- a/website/client/src/components/header/index.vue
+++ b/website/client/src/components/header/index.vue
@@ -83,7 +83,6 @@
     color: $header-color;
     flex-wrap: nowrap;
     position: relative;
-    max-height: 10.25rem;
   }
 
   .hide-header {


DXCanas:
@Tressley Tried my hand, but ran into some issues. Any guidance to offer?

SabreCat:
I'm struggling to even reproduce the original issue, so having a hard time digging into this. @DXCanas, can you elaborate on what you don't think is working in this PR? What sprites aren't positioned correctly, and how are they incorrect?

@Alys, it looks like you were able to reproduce the issue described in the original ticket--would you be able to give this PR a look?

DXCanas:
@SabreCat I think it's most noticeable at extremes, but if you look at the screencap, the avatar is not at center, as is usually the case. They're shifted down and to the right. 

To replicate, you should be able to manually assign the `font-size` property on the root `<html>` element to something lower than `16px`.

Alys:
@SabreCat I can replicate the issue by turning on View > Zoom > Zoom Text Only and then zooming out.

When I run this on my local install I see this which I guess is from the sprite issue that @DXCanas is asking about:

image

Tressley:
After investigating this a bit more, I think we'd need to convert each sprite slice to `rem` in order to properly prevent the overlap/misalignment from happening. I don't think that's something we want to do just yet without thorough testing. The next solution would be to keep the sprites using pixels as a unit and set a `min-height` on containers with `rem` that have sprites.

DXCanas:
@Tressley Would it make more sense to to use `px` for `min-height`, since `rem` is subject to change?

@Alys I actually hadn't seen that - I was just referring to how the character was no longer in the center of the avatar window. That's seems much more serious!

Tressley:
@DXCanas -- Yep! `px` would be the unit for containers that have sprite objects.

DXCanas:
I'll look at this sometime later this week, folks! Sorry about the delay.

DXCanas:
@Tressley It's a simple change, but did the trick for me locally

paglias:
Thanks @DXCanas ! This seems to be fixing the issue, I'm merging the PR and awarding your contributor tier.

DXCanas:
Sure thing, makes total sense. Just happy to get the change in, as it’s been affecting me personally 🙂

PR #11912, opened 2020-03-01
Fix for party members missing from header after viewing private messages / inbox / PMs

### LINKED ISSUE #11869: most party members missing from header after viewing private messages / inbox / PMs
EDIT 2020-02-22:

As described in the next comment, the bug occurs when you reload the website while viewing the private messages page. To reproduce:

1. Go to the [private messages page](https://habitica.com/private-messages)
2. Reload using your browser's refresh button.
3. Use the top navigation menu to visit any other page (don't use the Help menu as some parts of it work differently).
4. Observe that all but one of your party mates do not appear. (Reloading on that page makes them reappear).

----

As reported by @geolojess (da27bf23-48f7-469d-9a32-58521f9c84a9) a few days ago:

**"(Web) looks like whenever I check my PMs, then leave that page and go to another one, all but one of my party members shown in my header disappear. Refreshing the page brings the rest back. Here's a screenshot."**

image

I've confirmed the party has five members, not two.

A little later she said: 
**"... I tried [again] and the bug didn't occur this time! Now I'm unsure if that's to be expected or if the PMs is not the cause. If it happens again I'll try your instructions again [for getting console messages] and report back."**

A couple of days later she reported it happened again and she sent us the errors below from the browser's console _[EDIT: I've deleted them since they turned out to not be related]._ There was also nothing in Loggly that could be relevant (e.g., no errors related to loading party member data).

~I'm marking this as needs investigation and minor priority since it's only been reported by one person. We can increase the priority if we see more reports.~

tsukimi2:
Fixes #11869

Actually, after some further testing, this issue appears when you go to page that displays the app header from any page that does not display the app header.  Currently, both the private-messages page and the group-plan page do not display the app header component, so the issue appeared upon going to any page that displays the app header (e.g. Tasks page) from the group-plan page as well.

To fix the issue, added code to set the this.currentWidth in AppHeader component to the clientWidth of the partyMembersDiv (the div containing the list of party members in the app header component) whenever the computed property hideMember is called.  It works in manual testing.  But Not entirely sure this is a good solution in vuejs though, since it seems like I'm making a side effect (setting the this.currentWidth which is a data of the app header component) within a computed property.  Please let me know if there's a better solution.

The reason for the above code change is because the culprit of the issue is that when private-messages page is displayed, the app header has css property display:none to prevent the app header from displaying, but it is still being rendered.  Since its css display value is "none" at that time, the clientWidth of the partyMembersDiv within the app header is 0, and since the calculation of how many members to be included in the display of the party members list is based on that clientWidth, this causes most of the party members to be missing from the header.  But of course this is not a problem at this point since the app header containing the party member list is not visible on the private-messages page anyway.

However, upon going to a page that displays the AppHeader from the private-messages page, the AppHeader component, since it has already been rendered already, will not be rendered again and so its this.currentWidth (on which the calculation of the number of members to be displayed in the party member list is based on) remains 0; that's why you'll see that most of the party member list is missing.  A browser resize or page reload would put the correct party member list display back on the page because upon resize, the calculation of the number of members to be displayed in the party member list is triggered again.


paglias:
thanks for the investigation on the source of the issue @tsukimi2 !

I agree that having side effects in a computed property should be avoided. What about watching for it to change and calling `setPartyMembersWidth` when it changes?

You can watch properties like it's done here https://github.com/HabitRPG/habitica/blob/f8bab04a0e66a819f6d7ca5b210d27d991446bf3/website/client/src/components/ui/checkbox.vue#L32, the full api is described here instead https://vuejs.org/v2/api/#watch

tsukimi2:
@paglias Do you mean adding a watch on hideHeader like this?
```
  watch: {
    hideHeader () {
      this.$nextTick(() => {
        if (this.$refs.partyMembersDiv) {
          this.setPartyMembersWidth({ width: this.$refs.partyMembersDiv.clientWidth });
        }
      });
    },
  },
```

paglias:
@tsukimi2 yeah, I think the `$nextTick` might not be necessary when you use a watcher but I haven't tested it

tsukimi2:
> @tsukimi2 yeah, I think the `$nextTick` might not be necessary when you use a watcher but I haven't tested it

I was thinking of keeping the $nextTick in the watch function because when the app header is being displayed again after coming back from a page without header, there's no guarantee that the header will have been displayed already before the watch function watching the hideHeader calls the this.setPartyMembersWidth() function with the parameter this.$refs.partyMembersDiv.clientWidth.  In fact, when I manually tested it, without the nextTick this.$refs.partyMembersDiv.clientWidth equals 0 the watcher function is calling the this.setPartyMembersWidth() function.

tsukimi2:
@paglias I'm using vi editor to change code this time.  The result of git status before running lint was as follow:
```
osiris@osiris-LIFEBOOK-SH572:~/habitica$ git status
On branch bugfix-party-members-missing
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   website/client/src/components/header/index.vue

no changes added to commit (use "git add" and/or "git commit -a")
```
Result of running git status immediately after running lint:
```
osiris@osiris-LIFEBOOK-SH572:~/habitica$ git status
On branch bugfix-party-members-missing
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   website/client/src/components/achievements/firstDrops.vue
	modified:   website/client/src/components/header/index.vue
	modified:   website/client/src/components/messages/conversationItem.vue
	modified:   website/client/src/pages/private-messages.vue

no changes added to commit (use "git add" and/or "git commit -a")
```

I've discarded the modified files firstDrops.vue, conversationItem.vue, and private-messages.vue (i.e. only keeping the modified header index.vue file that I modified myself) before pushing the PR this time.

paglias:
You were right about `npm run lint` modifying the files, I'll try to remember to use it more often locally so that PRs won't have extraneous changes

paglias:
thanks for this one as well @tsukimi2 , it's working well :) noted towards the 3rd tier

PR #11915, opened 2020-03-01
Ensure first-time purchases are synced

### LINKED ISSUE #11907: Purchased items don't sync when buying them for the first time
### General Info
  * UUID: 3c2b1d99-c6f0-4e6e-a67a-bd8e023cfc7e
  * Browser: Chrome 
  * OS: Windows

### Description
When buying food, potions, or eggs for the first time (i.e. having never had this item in your inventory before), the store doesn't sync correctly. In other words, the item in the store won't show an itemCount badge until you perform an action that causes the page to load (e.g. switching tabs in the store). The same goes for your current inventory at the bottom of the screen - it won't be updated unless you switch tabs in the inventory.

My current guess is that this is caused by the fact that when you've never had an item before, it doesn't show up in the JSON representation of your inventory; buying it adds it to the JSON object with a count of 1. For example:
```
{
  fish: 1
}
```
becomes
```
{
  fish: 1,
  chocolate: 1
}
```
 However, when you HAVE had that item, the item key remains in the JSON object. When you have none of that item, it's represented with a count of 0, and so buying the item increases the count to 1. For example:
```
{
  fish: 1,
  chocolate: 0
}
```
becomes
```
{
  fish: 1,
  chocolate: 1
}
```
I haven't confirmed that this is the source of the errors, but it feels very related.

#### Console Errors
None.

alecbrick:

Fixes #11907. 

### Changes
The issue we were having was that when a user bought an item for the first time, they wouldn't see the item counts update. This was due to the fact that Vue isn't reactive when we add a key to an object.

One workaround to adding a new key is replacing the object containing the key. Instead of incrementing the count of the item we buy in `purchase.js`, we replace the `users.items[type]` object entirely.

----

### DIFF
--- a/website/common/script/ops/buy/purchase.js
+++ b/website/common/script/ops/buy/purchase.js
@@ -67,7 +67,7 @@ function purchaseItem (user, item, price, type, key) {
     user.items[type] = {
       ...user.items[type],
       [key]: user.items[type][key] + 1,
-    };
+    }; 
     if (user.markModified) user.markModified(`items.${type}`);
   }
 }
--- a/website/common/script/ops/buy/purchase.js
+++ b/website/common/script/ops/buy/purchase.js
@@ -67,7 +67,7 @@ function purchaseItem (user, item, price, type, key) {
     user.items[type] = {
       ...user.items[type],
       [key]: user.items[type][key] + 1,
-    }; 
+    };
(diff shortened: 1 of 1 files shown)


paglias:
@alecbrick thanks, it's working well! Noted towards your 3rd tier

PR #11918, opened 2020-03-01
Special case for member profile name.

### LINKED ISSUE #11914: sorting by name of party members in header should use case folding
The party members that appear in the website's header can by sorted by name (click the "View Party" button and select "Sort by Name" from the drop-down then click "Apply Sort Options to Party Header"). When you do that, you see that sorting is done case-sensitively (A, B, C, a, b, c). 

The desired behaviour is that sorting by name should use [locale-sensitive case folding](https://www.w3.org/International/wiki/Case_folding).

Previously, this was also a problem in the modal that appears when you click "View Party" (this issue: https://github.com/HabitRPG/habitica/issues/11547). It was fixed for the modal in this PR:  https://github.com/HabitRPG/habitica/pull/11595, and you can probably do something similar for sorting in the header.

Note though that it's important that the `direction` option is still used when sorting by name (see https://github.com/HabitRPG/habitica/issues/11913 for more details about that).

image

nikonhub:

Fixes #11914 

### Changes
Sorting members by name should be case insensitive. Add special case when comparing profile.name by passing a callback function to lowercase the name.

----

### DIFF
--- a/website/client/src/components/header/index.vue
+++ b/website/client/src/components/header/index.vue
@@ -170,9 +170,30 @@ export default {
       return Math.floor(this.currentWidth / 140) + 1;
     },
     sortedPartyMembers () {
-      const sortBy = this.user.party.order === 'profile.name' ? member => member.profile.name.toLowerCase() : this.user.party.order;
+      let sortedMembers = this.partyMembers;
+      const { order, orderAscending } = this.user.party;
 
-      return orderBy(this.partyMembers, [sortBy], [this.user.party.orderAscending]);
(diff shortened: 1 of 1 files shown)


Alys:
@Nikosmonaut The case folding needs to work for names in all languages, so converting to llower case isn't always suitable. See the link about locale-aware case folding in the issue for more details and recommendations. I believe the older PR that's linked from the issue uses an appropriate method for folding (although it doesn't consider direction, which is also important).

PR #11919, opened 2020-03-01
Fix asc/desc sort order when sorted by name

### LINKED ISSUE #11913: Party Sort Order Asc / Desc (Ascending / Descending) fails for "Sort by Name" in modal
When you click the "View Party" button in the header, a modal opens and you see a drop-down with "Sort by" options and another drop-down for  "Asc" (Ascending) and "Desc" (Descending).

The Asc / Desc options don't make any difference when "Sort by Name" has been selected.

The expected behaviour is that if you select "Sort by Name" and then switch between Asc and Desc, the party members in the modal should reverse order.

Asc / Desc does work as required for the other "Sort by" options.

image

nikonhub:

Fixes #11913 
### Changes
Reuse lodash orderby function for all cases to take care of sort order. Special case for profile.name when passing a function instead of a object path.

----

### DIFF
--- a/website/client/src/components/groups/membersModal.vue
+++ b/website/client/src/components/groups/membersModal.vue
@@ -478,20 +478,12 @@ export default {
       let sortedMembers = this.members;
 
       if (!isEmpty(this.sortOption)) {
-        // Use the memberlist filtered by searchTerm
-        if (this.sortOption.value === 'profile.name') {
-          // If members are to be sorted by name, use localeCompare for case-
-          // insensitive sort
-          sortedMembers.sort(
-            (a, b) => a.profile.name.localeCompare(b.profile.name),
-          );
-        } else {
-          sortedMembers = orderBy(
-            sortedMembers,
-            [this.sortOption.value],
-            [this.sortOption.direction],
-          );
-        }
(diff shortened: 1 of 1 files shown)


website/client/src/components/groups/membersModal.vue
```diff
@@ -478,20 +478,12 @@ export default {
       let sortedMembers = this.members;
 
       if (!isEmpty(this.sortOption)) {
-        // Use the memberlist filtered by searchTerm
-        if (this.sortOption.value === 'profile.name') {
-          // If members are to be sorted by name, use localeCompare for case-
-          // insensitive sort
-          sortedMembers.sort(
-            (a, b) => a.profile.name.localeCompare(b.profile.name),
```
Alys, line 486: localeCompare or a similar locale-aware case folding technique is still desirable for this so that names in all languages are handled correctly.

nikonhub: Sorry ! My bad. Case folding rhymed with case insensitive in my head.

paglias:
thanks @Nikosmonaut for both this and the other PR, both are working well. And welcome to the 3rd tier 🎉

PR #11954, opened 2020-03-08
Fix collapse checklist tooltip placement

### LINKED ISSUE #11888: "Collapse checklist" hovertext persists on mouseover and blocks checkboxes for checklist items
> "This was mentioned in the Tavern and I'm seeing it too: in the tasklist on the web, if I hover over the Collapse Checklist button on a task, and then move my mouse down so it's on the label popup, the popup stays until I move my mouse off the popup. The popup covers the top couple checklist items so they can't be checked off (until you move the mouse all the way out of range to get rid of the popup).
> I know some popups are designed to stay onscreen if you mouse over them, which makes sense when the popup is designed to time out. The label popup doesn't time out, so I don't think it needs to be 'paused' with a mouseover."

Reproduced in Firefox on Windows.

cvuorinen:

Fixes #11888

### Changes
As suggested in #11888, first tried placing the tooltip to `top`, but then it goes over the issue title which didn't look very nice:
Screenshot from 2020-03-08 23-08-35

Placing it to `right` looks nicer as there's more room there with plain white background:
Screenshot from 2020-03-08 23-08-58

----

### DIFF
--- a/website/client/src/components/tasks/task.vue
+++ b/website/client/src/components/tasks/task.vue
@@ -167,7 +167,7 @@
             <div class="d-inline-flex">
               <div
                 v-if="isUser"
-                v-b-tooltip.hover.bottom="$t(`${task.collapseChecklist
+                v-b-tooltip.hover.right="$t(`${task.collapseChecklist
                   ? 'expand': 'collapse'}Checklist`)"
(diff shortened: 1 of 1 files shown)


paglias:
thanks @cvuorinen , noted in your profile

PR #11967, opened 2020-03-13
[WIP] Move task icons up to the same line as the task name, saving vertical space

roastchicken:
Especially on smaller screens, space is precious. Having the icons on a separate line takes up vertical space, when there's plenty of room for them to sit to the right of the rest of the task content.

I've also created an issue for this in habitica-android, I'm not sure how separated the codebases are: https://github.com/HabitRPG/habitica-android/issues/1270.

### Changes
Set `position: absolute` on `.icons` in `task.vue`.

Before:
Skjermbilde 2020-03-13 kl  10 54 17

After:
Skjermbilde 2020-03-13 kl  10 54 10

A small possible issue is a slight overlap of text, this is the most overlap I was able to acheive:

Skjermbilde 2020-03-13 kl  10 59 53

This could be remedied by increasing `padding-right` on `.task-notes`, or `margin-bottom` on `.task-notes.markdown p`.

----
UUID: 7b9c79ad-ad01-4ef0-a250-01d0e4a73bf3

### DIFF
--- a/website/client/src/components/tasks/task.vue
+++ b/website/client/src/components/tasks/task.vue
@@ -605,19 +605,15 @@
     font-style: normal;
 
     &-right {
-      flex-grow: 1;
+      margin: 0;
+      padding: 7px 0;
+      padding-right: 8px;
+
+      flex-direction: column;
+      justify-content: space-between;
+      align-items: end;
+      flex-shrink: 0;
     }
-  }
-
-  .icons-right {
-    margin: 0;
-    padding: 7px 0;
-    padding-right: 8px;
-
-    flex-direction: column;
-    justify-content: space-between;
-    align-items: end;
-    flex-shrink: 0;
   }
 
   .icons-right .svg-icon {
--- a/website/client/src/components/tasks/task.vue
+++ b/website/client/src/components/tasks/task.vue
@@ -508,7 +508,6 @@
   }
 
   .task-content {
-    position: relative;
     padding-top: 0px;
     padding-bottom: 7px;
     flex-grow: 1;
@@ -592,10 +591,7 @@
   }
 
   .icons {
-    position: absolute;
-    right: 0;
-    bottom: 7px;
-
+    margin-top: 4px;
     color: $gray-300;
     font-style: normal;
 
--- a/website/client/src/components/tasks/task.vue
+++ b/website/client/src/components/tasks/task.vue
@@ -84,74 +84,6 @@
                 class="task-title"
                 :class="{ 'has-notes': task.notes }"
               ></h3>
-              <menu-dropdown
-                v-if="!isRunningYesterdailies && showOptions"
-                ref="taskDropdown"
-                v-b-tooltip.hover.top="$t('options')"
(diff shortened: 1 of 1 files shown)


paglias:
thanks @roastchicken ! pinging our designer @Tressley

Tressley:
@roastchicken -- The task icons are on their own line to account for the various widths of the task title, task description, checklist items, and task options. Additionally, the layout should be consistent for each task type. Your solution would break down on a To-Do that has a due date.

However, I do feel that there are spacing updates that could be made to shrink the current task height. Here's an example of what a Daily should look like at its smallest height (with icons.)

Red

roastchicken:
Ah, I hadn't realized that todo-dates are left-aligned.

As I mentioned in https://github.com/HabitRPG/habitica-android/issues/1270, having the icons under the task name makes it too high. There is no reason to have that extra height. This is especially noticable on mobile, where screen real-estate is quite limited.

To alleviate the issue of overlap with various widths, I suggest another solution which I have just commited: moving the icons to the right while keeping the due date on the bottom.

I don't fully see a reason for the layout being consistent between other task types and To-Dos. To-Dos are the only task type (AFAIK) that have an icon left-aligned. @Tressley could you explain your thoughts on that subject?

# Changes

My technical changes are outlined in the commits.

## Desktop

### Before
desktop_original

### After
desktop_move-task-icons

## Mobile

### Before
mobile_original

### After
mobile_move-task-icons

IMHO this design looks better.

Some problems remain, namely borders and the options button. I'll continue work on these; I'm not sure why the border is applied to the individual children instead of the parent.

roastchicken:
I've now fixed the border and options button. I've committed the options button, but because the border issue is logically disconnected from this change I've decided to make a separate pull request with it. Once that PR is accepted I'll rebase (and probably squash) this branch off of [develop](/HabitRPG/habitica/tree/develop).

Here's a screenshot of how it looks all together:
screenshot

I welcome any feedback you have @Tressley :)

Tressley:
@roastchicken -- Thanks so much for doing this, but I wish you'd have followed my initial feedback before making these sweeping changes.

> There is no reason to have that extra height.

As I stated before, the reasoning behind our design decision is to ensure the design doesn't break down with the various task types. While I agree this does save space, the most important parts of a task are the title and notes, so they should not compete visually with the icons. Additionally, this allows us to be more efficient in our code.

Again, I have revised task designs that include margin and padding updates that address your vertical height concerns. If you're interested in implementing those designs, I can share the Zeplin artboards with you (just need your email.) However, placing icons and the title on a single line is not the direction we'd like to pursue at this time.

roastchicken:
@Tressley
Respectfully, I completely disagree. The design I have implemented does not break any task types, as evidenced by my above screenshots. I don't believe moving the icons slightly up makes them compete with the title or the notes. Further, I don't understand what kind of efficiency you're referring to.

But very well, it seems I won't convice anyone to merge this. Feel free to send the Zeplin artboards to me at the email listed on my GitHub profile.

PR #11968, opened 2020-03-13
Remove strange disjointed :hover effect on .task

roastchicken:
This reverts part of #8926, specifically ["tasks hover state"](https://github.com/HabitRPG/habitica/pull/8926/commits/5f4fb6264141655a6ffa31324e6ea0dd5ebebece).

This commit is from almost three years ago, so I'm not sure why the border was added in such a strange way.  Regardless, this method is no longer needed as simply setting a border on .task works perfectly fine.

This has the advantage of allowing children to be added to .task without requiring more .no-right-border code. Something I need for #11967.

I believe this commit removes all CSS that references .groupTask, which leads me to believe that 'groupTask': task.group.id is unused. But I am not 100% sure, so I left it in.

## Changes
Replaces the border set on :hover on `.task-content`, `.left-control`, and `.right-control` with a border set on `.task`.

### Before
screenshot before

### After
screenshot after

----
UUID: 7b9c79ad-ad01-4ef0-a250-01d0e4a73bf3

### DIFF
--- a/website/client/src/components/tasks/task.vue
+++ b/website/client/src/components/tasks/task.vue
@@ -371,22 +371,6 @@
     position: relative;
 
     &:hover {
-      box-shadow: 0 1px 8px 0 rgba($black, 0.12), 0 4px 4px 0 rgba($black, 0.16);
-      z-index: 11;
-    }
-  }
-
-  .task:not(.groupTask) {
-    &:hover {
-      .left-control, .right-control, .task-content {
-        border-color: $purple-400;
-      }
-    }
-  }
-
-  .task.groupTask {
-
-    &:hover {
       border: $purple-400 solid 1px;
       border-radius: 3px;
       margin: -1px; // to counter the border width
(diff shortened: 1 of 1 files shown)


paglias:
Hi @roastchicken, thanks for the PR. I'm seeing a weird effect on dailys now

https://www.dropbox.com/s/7cs3fkf3bpdm9tc/Registrazione%20schermo%202020-03-16%20alle%2019.09.38.mov?dl=0

They seem to be missing 1px or so on their control on the left 
Schermata 2020-03-16 alle 19 07 49

paglias:
@roastchicken I think the disjointed hover effect might only exist in #11967 because I can't reproduce it on the current develop

roastchicken:
@paglias I'm not sure how you tried to reproduce, but it's possible it only existed with my changes. Regardless, now that #11967 has been closed I've lost the motivation to work on this. It has no effect anyways, unless another child element is added to `.task`.

PR #11991, opened 2020-03-18
Display Name and Profile blurb editing checks for slurs and banned words

### LINKED ISSUE #11865: Display Name and Profile blurb editing should include checks for swear words and slurs
As you probably know, Habitica's chat feature has code that prevents you posting a message that contains swear words or slurs - when you try to post such a message, you see an error and the message is not posted. Also, If you try to post a slur, your chat privileges are automatically revoked.

We want the same checks to be done when someone edits their profile, which they can do by clicking the Edit button on the [User Icon > Profile](https://habitica.com/user/profile) page.

I.e., when the user clicks Save on the Profile edit screen, the text in the Display Name and About fields should be checked to ensure that they have no swear words or slurs. If they do, the change should not be saved. If a slur is used, the user's public chat permissions should be revoked and moderators should be notified in the same way that they are for chat messages that contain slurs.

You can see at least some of the current code that checks for swears and slurs in chat posts by doing `git grep Banned` (case-sensitive).

As you're looking through the code base, you might come across the files `website/server/libs/bannedWords.js` and `website/server/libs/bannedSlurs.js` - be aware that that's where the bad words are stored, so don't read through them if you don't want to see the words. The top several lines of each file are safe and give you information about how the words are used. There's a large amount of space under that information so you should be able to read the top part of the files without seeing the bad words at the bottom.

For testing on your local install, you can use "TESTPLACEHOLDERSWEARWORDHERE" and "TESTPLACEHOLDERSWEARWORDHERE1" as swear words, and "TESTPLACEHOLDERSLURWORDHERE" and "TESTPLACEHOLDERSLURWORDHERE1" as slurs. They'll give the same behaviour as real swears or slurs would. To restore chat privileges to your test account, set `flags.chatRevoked` to `false`
 
Ideally, the checking of the Display Name and Profile blurb should be done with code in the API, not client-side code (this will later make it easier for the same checks to be added to the mobile apps or to third-party tools). HOWEVER it's important that the checks are done only when the user has changed their Display Name or Profile blurb. I.e., when any other change is saved to the user's account, the name and blurb should not be checked for bad words. If that proves to be difficult or impossible, post here and we'll consider options.

I'm leaving this as suggestion-discussion for about a week in case the staff want to make any comments about it but after that if there's no objections I'll mark it as help wanted.

---

_NB: In case anyone's curious, when a player uses a slur and has their chat permissions automatically revoked, the team of moderators are notified and we always review those repors promptly. In cases where the slur was used accidentally or innocently, we restore the chat permissions as quickly as we can and PM the user to explain what happened (e.g., if the user is posting a literary quote that contains a slur but that is otherwise unobjectionable)._

inftp:

Fixes #11865
### Changes
+ Add `blurbIssueSlur` to `website/common/locales/en/settings.json`
+ Add `stringContainsBannedWord()`
+ If user-submitted display name or blurb contain a banned word or slur, reject the change
+ If user-submitted display name or blurb contain a slur, set `chatRevoked` flag to true
* Rename `nameContainsSlur()` to `stringContainsSlur()`

----
UUID: ce9f5bff-a9f4-4bcc-acb0-4c0f8f7ad0f3


website/server/libs/user/index.js
```diff
@@ -112,7 +112,22 @@ export async function update (req, res, { isV3 = false }) {
     const newName = req.body['profile.name'];
     if (newName === null) throw new BadRequest(res.t('invalidReqParams'));
     if (newName.length > 30) throw new BadRequest(res.t('displaynameIssueLength'));
-    if (nameContainsSlur(newName)) throw new BadRequest(res.t('displaynameIssueSlur'));
+    if (stringContainsSlur(newName)) {
+      user.flags.chatRevoked = true;
+      await user.save();
```
paglias, line 121: @Alys I read the original issue but I'm a bit confused about this part:

> If they do, the change should not be saved. If a slur is used, the user's public chat permissions should be revoked and moderators should be notified in the same way that they are for chat messages that contain slurs.

If the change is not saved the user won't have a slur in their username, so is it still necessary to revoke their chat permissions?

paglias: Also I'm slightly concerned about the fact that we change an user property but return an error, so the user  won't see the fact that they're chat privileges have been revoked until they refresh the website.

paglias (commented):
Thanks @inftp ! 

I've left a comment but it's unrelated to the code, I've noticed that this part might be missing

> ...and moderators should be notified in the same way that they are for chat messages that contain slurs.

Also it would be great if you could add some tests of the functionality at test/api/v3/integration/user/PUT-user.test.js

website/server/libs/user/index.js
Alys: > > If a slur is used, the user's public chat permissions should be revoked

That's necessary because the user might go on to use a version of the slur that has one letter replaced to avoid the blocker and could then post to any public chat. Basically, if a player tries to use a slur in their name or profile, it's a danger sign for us and we wouldn't want them to have chat permissions until they'd emailed us and we could get them to agree in an email to follow the Community Guidelines.

> Also I'm slightly concerned about the fact that we change an user property but return an error, so the user won't see the fact that they're chat privileges have been revoked until they refresh the website.

That's a great point! We could change the error messages so that they contain more detail, for example using words from the error message they see if they try to post a slur to chat. If that's okay with you, I'll put some words together and run them past beffymaroo and the other mods.

paglias: Sounds good to me

inftp:
Thanks @paglias I will amend as suggested. 
> Also it would be great if you could add some tests of the functionality at test/api/v3/integration/user/PUT-user.test.js

Just to clarify: I should specifically be using the v3 test path and not `test/api/v4/user/PUT-user.test.js`?

inftp:
I've made some commits to my fork but I'm unsure how best to test the changes involving emails and slack messages? Is there a simple way to test this on a local development build?

A few other questions and things of note:
1. Should I also be writing tests to ensure that the user flag is set? Are there existing examples of such tests?
2. Should I be writing tests for the email/slack messages and are there examples?
3. I notice that even after receiving the error message about inappropriate language, the frontend still behaves as if the changes were made (e.g. profile name/blurb shows the slur in place) until the page is refreshed. Should #11865 also include fixes to prevent this?

Thanks all

website/server/libs/slack.js
```diff
@@ -239,7 +239,7 @@ function sendShadowMutedPostNotification ({
 function sendSlurNotification ({
   authorEmail,
   author,
-  group,
+  group = undefined,
```
paglias, line 242: undefined isn't the default?

website/server/libs/user/index.js
```diff
@@ -1,13 +1,17 @@
...
   BadRequest,
   NotAuthorized,
 } from '../errors';
 import { model as User, schema as UserSchema } from '../../models/user';
-import { nameContainsSlur } from './validation';
+import { stringContainsSlur, stringContainsBannedWord } from './validation';
+import { sendTxn, getUserInfo } from '../email';
+import * as slack from '../slack';
 
+const FLAG_REPORT_EMAILS = nconf.get('FLAG_REPORT_EMAIL').split(',').map(email => ({ email, canSend: true }));
```
paglias, line 14: given that this code is now used in multiple places I think the list of emails should be exported from server/libs/email instead of re-creating it every time

website/server/libs/slack.js
```diff
@@ -248,14 +248,21 @@ function sendSlurNotification ({
...
 
   let titleLink;
-  let title = `Slur in ${group.name}`;
+  let title;
 
-  if (group.id === TAVERN_ID) {
-    titleLink = `${BASE_URL}/groups/tavern`;
-  } else if (group.privacy === 'public') {
-    titleLink = `${BASE_URL}/groups/guild/${group.id}`;
+  if (group !== undefined) {
```
paglias, line 253: if (group) will work fine

website/server/libs/user/index.js
```diff
@@ -112,7 +116,68 @@ export async function update (req, res, { isV3 = false }) {
...
+        author: user,
+        message: `Name change: "${newName}"`,
+      });
+
+      throw new BadRequest(res.t('displaynameIssueSlur'));
+    }
+    if (stringContainsBannedWord(newName)) throw new BadRequest(res.t('displaynameIssueSlur'));
+  }
+
+  if (req.body['profile.blurb'] !== undefined) {
```
paglias, line 150: here I'd check for `req.body['profile.blurb'] != null` which will cover both null and undefined, usually I'd check directly for `if (req.body['profile.blurb'])` but that would not allow the user to remove the blurb once set

website/server/libs/user/index.js
```diff
@@ -112,7 +116,68 @@ export async function update (req, res, { isV3 = false }) {
     const newName = req.body['profile.name'];
     if (newName === null) throw new BadRequest(res.t('invalidReqParams'));
     if (newName.length > 30) throw new BadRequest(res.t('displaynameIssueLength'));
-    if (nameContainsSlur(newName)) throw new BadRequest(res.t('displaynameIssueSlur'));
+    if (stringContainsSlur(newName)) {
```
paglias, line 119: I think this might return an error if newName is undefined?

paglias:
sorry for the delay @inftp 

> I've made some commits to my fork but I'm unsure how best to test the changes involving emails and slack messages? Is there a simple way to test this on a local development build?
> Should I be writing tests for the email/slack messages and are there examples?

It's not possible to check that it works but you can write tests that make sure the right methods are called, some examples can be found here https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/chat/POST-chat.test.js#L153-L174 where it checks the right methods to send emails and messages to slack are called

> Should I also be writing tests to ensure that the user flag is set? Are there existing examples of such tests?

Ideally yes, for example here https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/chat/POST-chat.test.js#L406 it's tested by trying to send a new message after posting a slur and checking that it's rejected. Otherwise you can update the user `await user.sync()` and then check the flag `expect(user.flags.chatRevoked).to.eql(true)`

> I notice that even after receiving the error message about inappropriate language, the frontend still behaves as if the changes were made (e.g. profile name/blurb shows the slur in place) until the page is refreshed. Should #11865 also include fixes to prevent this?

Good point, looks like we're not handling the failures here https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/userMenu/profile.vue#L905

we would have to wrap the call in a try/catch and revert the changes when an error is throw

```
try {
 await this.$store.dispatch('user:set', values);
} catch (e) {

}
```

but if it's too much for this PR we can create a new issue and fix it later.

Let me know if you have any doubt

Alys:
Regarding the error messages:

---

Firstly, just for reference, we have these ones already:

"displaynameIssueSlur": "Display Names may not contain inappropriate language.",

"usernameIssueSlur": "Usernames may not contain inappropriate language.",

"usernameIssueForbidden": "Usernames may not contain restricted words.",

"bannedSlurUsed": "Your post contained inappropriate language, and your chat privileges have been revoked.",

"bannedWordUsed": "Oops! Looks like this post contains a swearword, religious oath, or reference to an addictive substance or adult topic (<%= swearWordsUsed %>). Habitica has users from all backgrounds, so we keep our chat very clean. Feel free to edit your message so you can post it!",

---

Below are my draft suggestions for the new errors. The mods and beffymaroo haven't objected but there might be small wording changes later.

Slur used in Profile blurb:
"Your About text contains inappropriate language and your chat privileges have been revoked. Please email admin@habitica.com if you would like to discuss having your privileges restored."

Swear used in Profile blurb:
"Oops! Looks like your new About text contains a swearword, religious oath, or reference to an addictive substance or adult topic (<bad word(s) listed here>). Habitica has users from all backgrounds, so we keep profiles very clean. Feel free to edit your text so you can save it!"

Slur used in Username or Display Name:
"Your name contains inappropriate language and your chat privileges have been revoked. Please email admin@habitica.com if you would like to discuss having your privileges restored."

Swear used in Username or Display Name:
"Oops! Looks like your new name contains a swearword, religious oath, or reference to an addictive substance or adult topic (<bad word(s) listed here>). Habitica has users from all backgrounds, so we keep our names very clean. Feel free to choose a different name!"

website/server/libs/slack.js
inftp: It is, but explicitly declaring the parameter this way creates JSDoc type-hinting to IDEs (see attached screenshot) that specify the param is optional (in this case, because a slur report will no longer always relate to a group chat). 

If this goes against Habitica's coding style guidelines I will happily remove it. 

sendSlurNotification

website/server/libs/slack.js
inftp: Strict checking prevents other falsy values from being flagged and I was unsure whether all falsy values (e.g. 0) were invalid for group values. I will amend.

website/server/libs/slack.js
paglias: sorry for the delay but yeah I would remove this, i understand it might help with some IDEs but I'd prefer not to introduce the change just here

website/server/libs/slack.js
paglias: yeah, all falsy values are invalid for groups as it's an object

Alys:
I'm closing this but I'm about to say in the associated issue that the comments made here should be reviewed by the next contributor, and the code can be considered for re-use. Thanks @inftp !

PR #11993, opened 2020-03-19
Fixing the equipment popover message when equipment is from a different class - Fixes #11988

### LINKED ISSUE #11988: wrong locked equipment message for gear from another class
The Equipment section on Shops > Market gives the wrong message for basic equipment from another class. It says:

"This item is only available once you've purchased the previous items in sequence. Keep working your way up!" (that's the `tierLockedItem` string).

You should see that string only when you're in the same class as the equipment. When you're in a different class you should see this message:

"This item is only available to a specific class. At level 10 or above, you can change your class under the User icon > Settings > Character Build!" (`classLockedItem`)

The screenshot below show the bug. I lost the equipment shown when I was a Warrior. Now I am a Rogue and you can see it's displaying the wrong message.

image

This bug was reported by lex / @wlw_witch (87b9857a-48e8-48f8-8aae-fa52dd81f895) and confirmed by me.

laurel-thomson:

Fixes #11988

### Changes
I adjusted the popover message for Equipment in the Shops -> Market section.  Previously, the `tierLockedItem` message was displaying when viewing equipment from a class other than your own.  This was the incorrect message however, as it implied that the Equipment was unavailable because the user hadn't bought the previous items in the tier, when really the Equipment was unavailable because its class didn't match the user's class.  I updated the popover message to display the `classLockedItem` message when the Equipment's 'klass' and 'specialClass' properties don't match the user's class.

----
UUID: 04ec9b30-1464-41f5-af67-fb2482f108a0


website/client/src/components/inventory/equipment/attributesPopover.vue
```diff
@@ -5,7 +5,7 @@
         {{ `${$t('lockedItem')}` }}
       </h4>
       <div
-        v-if="item.specialClass"
+        v-if="item.specialClass !== user.stats.class && item.klass !== user.stats.class"
```
paglias: Hi @laurel-thomson ! Should the check here be different in case the item doesn't have the `specialClass`
property? Also i think all items with a special class have `item.klass` set to `special` which will never match the user's own class. Here it's how it's checked on the server side https://github.com/HabitRPG/habitica/blob/bd8e67a2ea3d6109ab2e43057ecdfe8f1be2e363/website/common/script/ops/buy/buyMarketGear.js#L27-L28 (btw, if it's possible it would be great if the check could be abstracted an reused)

cc'ing @Alys since she knows more about this

paglias (commented):
Left one comment about the `v-if` check

website/client/src/components/inventory/equipment/attributesPopover.vue
Alys: > Should the check here be different in case the item doesn't have the specialClass property?

Yes, there'd need to be a check for specialClass existing. @laurel-thomson in case it helps, here's an example of an item that just has klass, taken from the output of the API's `content` route:

"weapon_warrior_0":{
   "text":"Training Sword","notes":"Practice...","value":1,"type":"weapon","key":"weapon_warrior_0","set":"warrior-0",
   **"klass":"warrior",**
   "index":"0","str":0,"int":0,"per":0,"con":0
}

> Also i think all items with a special class have item.klass set to special which will never match the user's own class.

Yes. Here's an example of a special class item:

"weapon_special_spring2020Warrior": {
   "event":{"start":"2019-03-17","end":"2020-05-02"},
   **"specialClass":"warrior",**
   "set":"spring2020BeetleWarriorSet","text":"Sharpened Wing","notes":"Fight...","value":90,"str":15,"type":"weapon","key":"weapon_special_spring2020Warrior",
   **"klass":"special",**
   "index":"spring2020Warrior","int":0,"per":0,"con":0
},

The code that paglias referenced looks good for checking klass and specialClass, although it would be clearer if `checkKlass` was renamed to `wrongKlass` and `checkSpecialClass` was renamed to `wrongSpecialClass`:
https://github.com/HabitRPG/habitica/blob/bd8e67a2ea3d6109ab2e43057ecdfe8f1be2e363/website/common/script/ops/buy/buyMarketGear.js#L24-L34


However from looking at that code and then doing a test, I've realised that the API doesn't prevent you buying a "tierLocked" item. I.e., if you don't own either of these:
- armor_warrior_2
- armor_warrior_4

then you should be allowed to buy armor_warrior_2 but not armor_warrior_4. The API DOES allow you to buy armor_warrior_4 before armor_warrior_2 using the command below - that's a bug. 
`curl -X POST -d "" -H "x-api-user: d904bd62-da08-416b-a816-ba797c9ee265" -H "x-api-key: ......." https://habitica.com/api/v3/user/buy-gear/armor_warrior_4`


I'm sorry @laurel-thomson it's a bit of a complex situation and we didn't realise that the API had a different bug than the website. :-/ If you'd like to keep working on this that would be awesome and we'll be happy to answer any questions you have!

laurel-thomson: Thanks for all the information!  I can keep working on this.

laurel-thomson:
I've updated the "wrong class" check to use the same logic as the backend.  I've also abstracted this check to be a computed property on the ShopItem component.  Let me know what you guys think.

paglias (commented):
Sorry for the delay! The code looks good, I just have one doubt, the old check `item.specialClass` was defined directly inside the attributes popover component while the new one is passed as a proposition from the shopItem component so it won't be passed when used in `equipment/index.vue`. @Alys is the check needed only for shop items? I think it could be a good idea to move the check to a computed property inside the attributesPopover component directly so it doesn't have to be passed around

Alys:
The message in the popover will be needed in the Rewards column when the item has been pinned there, so it sounds like a computed property inside the attributesPopover component would be a good idea.

laurel-thomson:
Sounds good! I can do that

laurel-thomson:
Okay, I've updated the isWrongClass function to be a computed property on the attributesPopover component.  I had a couple issues with my text editor making whitespace changes to other files, but I believe those are all reverted now.

paglias:
@laurel-thomson  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #11994, opened 2020-03-21
edit foundNewItemsExplanation string to say Pet Food, not food

Alys:
This changes the `foundNewItemsExplanation` to say "Pet Food" rather than just "food".

It also capitalises terms in that string that are usually capitalised these days.

Note that I haven't run this past any staff member so it might not be desirable but doing a PR was about as fast as mentioning the potential issue somewhere. :)


paglias:
Looks like every other reference of Food has been changed so this seems correct 👍 thanks

PR #11995, opened 2020-03-22
[WIP]:Refactor/auth component

### LINKED ISSUE #11482: Client Auth: refactor components
We have three different components that more or less do the same thing: signup and signin.

They're located in
- website/client/src/components/auth/registerLoginReset.vue
- website/client/src/components/auth/authForm.vue
- website/client/src/components/static/home.vue

They share most design and almost all the business logic and should be combined into a single one where possible.

It would also make sense to abstract the auth verification logic (used when choosing or changign your username)

ShupingHe:
fix #11482 

### Changes
add 3 components for login、register and reset password,still work in progress, need to remove duplicate css style and abstract the verification logic

----
UUID:

### DIFF
--- a/test/api/unit/libs/highlightMentions.test.js
+++ b/test/api/unit/libs/highlightMentions.test.js
@@ -105,4 +105,18 @@ describe('highlightMentions', () => {
       expect(result[0]).to.equal('[@user](/profile/111) `@user`');
     });
   });
+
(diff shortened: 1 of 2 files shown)


paglias:
@Stevens1995 I'm going to mark this as `needs works`, just leave a comment when it's ready :)

PR #12010, opened 2020-03-24
Changed private message" button to use <a href> to open Private Messa…

### LINKED ISSUE #11910: private messages (PMs) screen is obscured by challenge/guild participant/member lists
@mikoboy (0731c2a5-7ca8-446f-a86c-8121c1be3f1c) repored:

**"Opened my [challenges]. Opened the list of participants. And clicked to send a message I was in a chat room with this [person], but there were a list that covered up the visibility of the chat."**

It happens for me too with challenges and guilds. Below are the steps to reproduce, with screenshots. 

Step 1: view list of challenge participants or guild members:
bug3

Step 2: view profile of one participant:
bug2

Step 3: click private message icon, get taken to private message screen, observe that profile screen overlays it:
bug1

tsukimi2:
Fixes #11910

Changed private message" button to use <a href to open Private Message page in same tab with normal mouse click, and open in a new tab with right mouse click mouse context menu.  Pass the selected user ID in profile component as part of url query when clicking the private message icon and opening the PMs page.


website/client/src/components/userMenu/profile.vue
```diff
@@ -10,10 +10,12 @@
           class="btn btn-secondary message-icon"
           @click="sendMessage()"
         >
-          <div
-            class="svg-icon message-icon"
-            v-html="icons.message"
-          ></div>
+          <a :href="'/private-messages?uuid=' + user._id">
```
paglias: can we use a `router-link` component? It'll be rendered as an `a` tag but meant to work with Vue router links https://router.vuejs.org/api/#router-link-props

website/client/src/pages/private-messages.vue
```diff
@@ -630,7 +630,16 @@ export default {
 
     await this.reload();
 
-    const data = this.$store.state.privateMessageOptions;
+    let data = this.$store.state.privateMessageOptions;
```
paglias: I'd say this is the perfect place to use Navigation Guards https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards, they'll be called when the route is loaded

paglias (changes requested):

paglias:
@tsukimi2 sorry but I'm still seeing the issue with the current code, the members modal remains open in the PM pages

tsukimi2:
> the members modal remains open in the PM pages

@paglias By "the members modal remains open in the PM pages", do you mean members modal remains open when the PM page is opened in an existing tab (i.e. when mouse left-click on the message icon), or do you mean members modal remains open on the original tab when the PM page is being opened in a new tab (i.e. the case when mouse right-click and then choosing "open in a new tab")?

paglias:
@tsukimi2 in an existing tab

tsukimi2:
@paglias Sorry, tested the previous code on mac and seemed to work properly there but did not realize that for some reason it's not working on linux when testing the workflow more than once.

Anyway, the updated code was manually tested on chrome and safari on mac, and chrome and firefox on linux.  But I'm not entirely sure whether this is the correct way to solve the problem.  Under further testing, I found that when I use the code (in line 654 on private-messages.vue)

this.$root.$emit('bv:hide::modal', 'members-modal');

alone to close the members modal, it works only on mac but not on linux.  However, when I use the code

this.$root.$emit('habitica::dismiss-modal', 'members-modal');

alone it works on linux and not on mac.  So for now I put both lines I put both lines into the modified code and seems to work on both mac and linux (have not tried out on windows since I have not set up the habitica development environment on windows), though I do not understand why this works.  Feeling a bit uneasy about this addition since it feels a bit hack-ish and I do not have the dev environment to test it out on windows yet.

I also tried to observe "this.$store.state.modalStack" (line 696) when running the code, which is the part of the codebase where it actually hides modals.
https://github.com/HabitRPG/habitica/blob/8efed3724150628ace963e4ca2a0554814a73ecf/website/client/src/app.vue#L583-L612

the code seems to be able to clear off the this.$store.state.modalStack with length = 0 after clicking the private messages icon on the profile.vue, so not entirely sure myself why the original submitted PR was not able to work properly on some platforms in the first place.

paglias:
> @paglias Sorry, tested the previous code on mac and seemed to work properly there but did not realize that for some reason it's not working on linux when testing the workflow more than once.

Actually I had tested on Mac as well but using Edge (which should be pretty similar to Chrome), @SabreCat have you ever seen something similar with modals?

paglias (commented):
sorry for the delay, I tested it and it works but I think `this.$root.$emit('habitica::dismiss-modal', 'members-modal'); ` is enough since it also calls `this.$root.$emit('bv:hide::modal', 'members-modal'); `

tsukimi2:
> sorry for the delay, I tested it and it works but I think `this.$root.$emit('habitica::dismiss-modal', 'members-modal'); ` is enough since it also calls `this.$root.$emit('bv:hide::modal', 'members-modal'); `

@paglias Yes, I was aware that habitica::dismiss-modal calls "this.$root.$emit('bv:hide::modal".  However, when I tested it on mac, the members modal would still be displayed when I used only the line "this.$root.$emit('habitica::dismiss-modal', 'members-modal')".  When I tried to step over the code while the code was running, habitica::dismiss-modal did indeed seem to have called the line "this.$root.$emit('bv:hide::modal" within it, but for reason I do not yet understand it was not able to hide the members modal from there.  This happened on my mac running on the following:
Dev env: docker
Chrome: (version 80.0)
safari (v. 13.1)
edge (v.80.0)

In fact, on my mac, even if I just included the line "this.$root.$emit('bv:hide::modal" without including "this.$root.$emit('habitica::dismiss-modal", it would still work when tested against chrome, safari, and edge.  But on my ubuntu it's the other way around.

paglias:
Hmm but I tested on mac as well and got the opposite result... @negue do you have any idea ?

website/client/src/components/userMenu/profile.vue
```diff
@@ -5,16 +5,18 @@
...
-          <div
-            class="svg-icon message-icon"
-            v-html="icons.message"
-          ></div>
-        </button>
+        <router-link :to="{ path: '/private-messages', query: { uuid: user._id } }">
+          <button
+            v-b-tooltip.hover.left="$t('sendMessage')"
+            class="btn btn-secondary message-icon"
+            @click="sendMessage()"
```
negue: I think having a simple click executes the button click and the router-link which is then the reason the modals are opening up again.

negue: `sendMessage` should be probably removed

website/client/src/pages/private-messages.vue
```diff
@@ -630,8 +648,13 @@ export default {
 
     await this.reload();
 
-    const data = this.$store.state.privateMessageOptions;
+    // close members modal if the Private Messages page is opened in an existing tab
+    if (this.hasPrivateMessageOptionsOnPageLoad) {
+      this.$root.$emit('habitica::dismiss-modal', 'members-modal'); // hide members modal on linux
```
negue: this would be needed once you also update the `membersModal.vue` :P 

but for now `profile` would be the better solution

negue (changes requested):
OK, after I wrote some information and tested it out github didn't save my message ... so rewriting some of it again.

I tested with Mac Vivaldi / Chrome and Windows Firefox, the current way worked (even *without* the additional `bc:hide::modal` call).

When you click the button/link (see comments) you will be redirected to the PM page, BUT pressing browser back, the modal repears again. I think this is because the modals are in our history-stack. Maybe we need something like https://router.vuejs.org/api/#replace to move to the PM page, but then the users won't be able to come back (with the browser back issue they can't anyway)

website/client/src/components/userMenu/profile.vue
paglias: agreed with @negue , using the route itself seems to be enough

paglias: So I think the button and the `sendMessage ` method can be removed

website/client/src/pages/private-messages.vue
paglias: @negue are you suggesting to remove the line above or to simply change it to `this.$root.$emit('habitica::dismiss-modal', 'profile');`?

I'm also wondering if the line below is actually needed

paglias (changes requested):

paglias:
thanks @tsukimi2 ! it's working well now, noted towards your 3rd tier

PR #12016, opened 2020-03-26
stop 'pts' from overflowing to next line

### LINKED ISSUE #11738: Allocating 100 stat points to one stat causes text to overflow from the box
image

> Not so much a bug as it is a formatting issue, but I just noticed when I reached level 100 (woo!) and put my 100th stat point into intelligence, that the "pts" text after the number 100 gets pushed out of the box and moved underneath it. I assume this is because the number has become three digits and the entire "100 pts" text doesn't fit in the box.

Reported by JohnJSal in Report a Bug.

inftp:

Fixes #11738 

### Changes
+ Add `stat-points` class to stat point container divs
+ Add `white-space: nowrap` to class `stat-points`

This appears to fix the overflow issue for the "pts" text however I have noticed numerous other text overflow issues at some media widths (both before and after the changes in this PR). Unsure whether these are to be addressed under #11738 or other ticket(s)?

----
UUID: ce9f5bff-a9f4-4bcc-acb0-4c0f8f7ad0f3


website/client/src/components/userMenu/profileStats.vue
```diff
@@ -266,7 +266,7 @@
           class="col-12 col-md-3"
         >
           <div class="box white row col-12">
-            <div class="col-9">
+            <div class="col-9 stat-points">
```
paglias: could we use the `text-nowrap` class that comes with bootstrap? this way we won't have to create a new one every time

paglias (changes requested):
thanks @inftp ! the code works well, I just left one comment, regarding the other places where this issue happens, could you file a new issue with all of them? thanks!

paglias:
@inftp thanks for this! I'm merging the PR and awarding your contributor tier.

Issue #7599, opened 2016-06-03
challenge tasks in user's task lists are in wrong sort order after joining


Alys:
When you join a challenge, the tasks that are added to your own task list are not in the same order as they are in the challenge as shown in the Challenges page. They should be.

For challenges where the order is important, challenge owners can number each task and write instructions in the challenge for the participants to follow the numbers. That's probably a good idea even after this bug is fixed, since participants will still be free to reorder their own tasks.

hmacphail:
Hello, trying to replicate this to see if I can fix it, but I'm a bit confused. 

Is there a certain way to order your challenge tasks or is this just referring to "the order in which the tasks are displayed"? Also, I found the tasks to all add in the correct order, but maybe I'm missing a step. In my test, the challenge creator had all the tasks in reverse order (is this the bug?), while for normal challenge participants, the tasks were all in correct order.

hmacphail:
Just noticed as well, I cannot rearrange the order of the challenge tasks on the challenge creation/edit pages. When I click to drag or click the push to top arrow, nothing happens. Tested on both chrome and firefox.

Alys:
The order in which tasks are displayed on the challenge's page is the order in which they should be displayed in your own task lists after you've joined the challenge. (Although you are permitted to reorder them in your own lists after you've joined.)

There is a different bug that prevents a challenge creator from sorting their tasks, but that's not related to this. https://github.com/HabitRPG/habitrpg/issues/1703

A fix for this issue can assume that the creator has put the tasks in the correct order, and should mimic that order when giving the tasks to the participants.

Alys:
A fix for this issue can assume that the creator has put the tasks in the correct order, and should mimic that order when giving the tasks to the participants.

ShulamitLH:
When I recently added a new challenge, the order was completely rearranged when I hit save. It bears no resemblance to the order I added them.

ghost:
[memorable]

ShulamitLH:
I just tried creating a new challenge with a variety of tasks and difficulties. After saving, I had a different member of our party join the challenge. All of the tasks had randomly organized themselves and had all reset the difficulty to easy, despite that I had set them about 1/3 easy, 1/3 medium and 1/3 hard. I was unable to make any changes on the challenge using edit.

tricksyfox:
[...]

PR #12021, opened 2020-03-29
Make challenge and group tasks draggable

### LINKED ISSUE #1703: challenge tasks don’t retain sort order when creating or editing a challenge
One can reorder items on the Challenges tab but the sort reverts. 

EDIT by Alys 2020-05-13: Before working on this, please read the comments below and also the comments in this pull request: https://github.com/HabitRPG/habitica/pull/12021 . If you think the code in that PR is suitable for re-use in your solution, you're welcome to use it - please mention in your PR that you've done so.

EDIT by Alys 2017: This issue will be easier to fix now because  there is code for sorting tasks in Group Plans (#8358) and it can be reused for challenges. Copying from @TheHollidayInn's [comment below](https://github.com/HabitRPG/habitica/issues/1703#issuecomment-274308541):
"the code allows to order challenge tasks (and can be shared between all three). Although, there needs to be a taskOrder object on the challenge - not sure if that exists."

EDIT by Alys 2020-02-28: What's supposed to happen is that the order of the tasks you see on the challenge's screen should be the same as the order of the task IDs in the challenge's `tasksOrder` object. Below is an example of that object in a challenge's document.

An additional problem is that the challenge's page has lost the ability for the challenge owner to reorder the tasks. The owner should be able to drag and drop in the same way that a user can drag and drop their own tasks. Admins (moderators and staff) should also be able to drag and drop the challenge's tasks (there's already code for challenges that gives admins the same rights at the challenge owner). Other players should not be able to drag a challenge's tasks in the challenge's own page (of course they can drag their own copies of the tasks after they join the challenge).

```
{
    "_id": "b7fbd6a2-6d5e-4f28-a81f-12f42f9cf447",
    "tasksOrder": {
        "habits": [
            "990933b0-4853-4f38-92b0-3864ce576f9e",
            "3d131996-768b-46e9-937d-993452a78df2",
            "5ca7fbf1-b7b5-498e-9c6d-85d328a50099"
        ],
        "dailys": [
            "30ef773b-292e-46e7-beb1-ecf2df66cbb3",
            "62782db7-8bbd-48e4-b0b0-2ab86a5d1031"
        ],
        "todos": [
            "85a5e240-8c64-446d-baab-db6debbabdd0",
            "71a67058-e022-497a-8319-d52584383b66"
        ],
        "rewards": [
            "2c0f8ed9-a61f-4ba4-9ad4-5f36a25db02a"
        ]
    },
    "official": false,
    "memberCount": 0,
    "prize": 0,
    "name": "My Challenge",
    .... etc ....
}
```

alecbrick:

Fixes #1703  

### Changes
Currently, challenge and group tasks can't be dragged and reordered. This change allows them to be reordered by anyone who can create or modify tasks.

However, one of the issues with allowing challenge and group tasks to be reordered is that challenges and groups don't respect the `tasksOrder` object. As such, this PR also ensures that the tasks are always ordered according to `tasksOrder`, which is done by ordering the tasks in the GET /tasks/user endpoint.

One thing to note is that in the front end, challenges and groups currently add tasks at the _end_ of their lists, while user tasks are added at the _beginning_ of their lists. However, this all occurs in one place in the backend, where the tasks are added to the beginning of their `tasksOrder` lists. In order to unify these, I've changed challenges and groups so that new tasks are added at the beginning of their lists, rather than at the end. Let me know if that's not something we want!

Additionally, I've refactored some of the code in the Tasks API that checks whether a user can modify a task.

----


Alys:
> to unify these, I've changed challenges and groups so that new tasks are added at the beginning of their lists, rather than at the end. Let me know if that's not something we want!

In my (non-staff) opinion, I think that's a sensible decision. It will take a bit of time for existing players to get used to the change but consistency in behaviour and in code seems best to me. I've labelled this as notify helpers to remind me to post to various guilds about the change when it goes live, and also to update https://github.com/HabitRPG/habitica/issues/7599 to say that the challenge's tasksOrder arrays can be used.

paglias:
paging @Tressley about the tasks order

test/api/v3/integration/tasks/POST-tasks_unlink-one_taskId.test.js
```diff
@@ -92,16 +92,16 @@ describe('POST /tasks/unlink-one/:taskId', () => {
 
   it('unlinks a task from a challenge and saves it on keep=keep', async () => {
     await user.post(`/tasks/challenge/${challenge._id}`, tasksToTest.daily);
-    let [, daily] = await user.get('/tasks/user');
```
paglias, line 95: all these changes from `[, task]` to `[task]` are because new tasks are now pushed to the top of the list, right?

website/server/libs/tasks/index.js
```diff
@@ -1,63 +1,25 @@
-import moment from 'moment';
 import _ from 'lodash';
-import * as Tasks from '../models/task';
+import {
```
paglias, line 2: thanks for reorganizing this file, makes much more sense this way and I've wanted to do it for a long time, but I feel like the unit tests sould be changed as well as right now they're still in test/api/unit/libs/tasksManager and we try to have tests files match the code file name

paglias: Also, I'm slightly confused on way some functions are in this file while others are in tasks/utils

And I'd prefer if they were not exported from two different places, for example all the ones in tasks/utils are exported both from there and from here

website/server/libs/tasks/index.js
```diff
@@ -232,62 +194,88 @@ export async function getTasks (req, res, options = {}) {
...
 
-  if (currentIndex !== -1) order.splice(currentIndex, 1);
-  if (to === -1) {
-    order.push(taskId);
-  } else {
-    order.splice(to, 0, taskId);
+
+async function getChallengeFromTask (task) {
+  if (task.challenge.id && !task.userId) {
+    return Challenge.findOne({ _id: task.challenge.id });
```
paglias: .exec() at the end here, it's confusing but findOne doesn't return a real promise untill you call .exec()

test/api/v3/integration/tasks/challenges/GET_tasks_challenge.id.test.js
```diff
@@ -84,4 +84,35 @@ describe('GET /tasks/challenge/:challengeId', () => {
       });
     });
   });
+
+  it('maintains challenge task order', async () => {
```
paglias, line 88: this test should be in POST-tasks_taskId_move_to? same for the group one

website/server/controllers/api-v3/tasks.js
```diff
@@ -937,20 +919,30 @@ api.moveTask = {
...
+
+    if (!task) {
+      throw new NotFound(res.t('taskNotFound'));
+    }
+
+    const group = await getGroupFromTaskAndUser(task, user);
+    const challenge = await getChallengeFromTask(task);
+    verifyTaskModification(task, user, group, challenge, res);
 
     if (!task) throw new NotFound(res.t('taskNotFound'));
```
paglias: is this a duplicate of line 924-926?

website/server/controllers/api-v3/tasks.js
```diff
@@ -961,12 +953,12 @@ api.moveTask = {
...
       $position: position,
     };
-    await user.update(updateQuery).exec();
+    await owner.update(updateQuery).exec();
 
     // Update the user version field manually,
     // it cannot be updated in the pre update hook
     // See https://github.com/HabitRPG/habitica/pull/9321#issuecomment-354187666 for more info
-    user._v += 1;
+    owner._v += 1;
```
paglias: the _v field exists only on users

paglias (commented):
first of all thanks for the PR and the general re-org that you made, I left some comments but most are about the organization, not the code itself

test/api/v3/integration/tasks/challenges/GET_tasks_challenge.id.test.js
paglias: sorry, I misread the PR description... anyway should we add a test to check that challenge and group tasks can be moved in POST-tasks_taskId_move_to? since right now it only tests users' tasks

paglias:
regarding 

> However, one of the issues with allowing challenge and group tasks to be reordered is that challenges and groups don't respect the tasksOrder object. As such, this PR also ensures that the tasks are always ordered according to tasksOrder, which is done by ordering the tasks in the GET /tasks/user endpoint.

I'm surely missing something but when getting tasks for a challenge or a group it uses getTasks in the old tasksManager file which has the code to respect the tasks order

test/api/v3/integration/tasks/POST-tasks_unlink-one_taskId.test.js
alecbrick: It's because previously, I think the tasks were sorted by the order in which they were added - since the user was generated with a Todo, that Todo originally came first. Now, however, this sorts the challenges according to task order. Since it wasn't clear how to return the different task types when we ask for all the types at the same time, I chose the order `habits`, `dailies`, `todos`, and `rewards` (the same way they're displayed on the tasks page). As such, the daily now comes before the Todo, and I've updated the call to reflect that.

website/server/libs/tasks/index.js
alecbrick: The functions in this file are largely ones that access the database, while the functions in `tasks/utils` are those that don't access the database. I did so in order to remove circular dependencies, since `Challenge` and `Group` rely on functions from `tasks/utils`.

I exported the functions from both files in order to allow anything that needs a function from the `tasks` library to just import `/lib/tasks` instead of worrying about the underlying structure. However, I got this idea from Stack Overflow, and it didn't really turn out the way I liked - the `Challenge` and `Group` files still import from `/tasks/utils`. I can remove the unnecessary exports.

website/server/libs/tasks/index.js
alecbrick: Nice catch!

test/api/v3/integration/tasks/challenges/GET_tasks_challenge.id.test.js
alecbrick: Sure, that sounds like a good idea.

website/server/controllers/api-v3/tasks.js
alecbrick: It is, thanks!

website/server/controllers/api-v3/tasks.js
alecbrick: Whoops, thanks!

alecbrick:
> I'm surely missing something but when getting tasks for a challenge or a group it uses getTasks in the old tasksManager file which has the code to respect the tasks order

Nope. Currently, `getTasks` only gets the tasks in order when you pass in the `?type=` query parameter.  See [line 235 in taskManager](https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/taskManager.js#L235).

Tressley:
Placing the Challenge and Group tasks at the beginning of the user's list is a good idea. Let's move forward. 👍

website/server/controllers/api-v3/tasks/groups.js
```diff
@@ -11,7 +11,7 @@ import {
   createTasks,
   getTasks,
   moveTask,
```
paglias, line 13: moveTask is in `libs/tasks/utils`

test/api/unit/libs/taskManager.js
```diff
@@ -3,7 +3,7 @@ import {
   getTasks,
   syncableAttrs,
   moveTask,
```
paglias, line 5: moveTask is in the utils file

test/api/v3/integration/tasks/challenges/GET_tasks_challenge.id.test.js
paglias: I think this one slipped through :)

paglias (changes requested):
left a few small comments

paglias:
Sorry for the delay @alecbrick , the PR works good, I've left a few small comments but nothing major.

Regarding this 

> Nope. Currently, getTasks only gets the tasks in order when you pass in the ?type= query parameter. See line 235 in taskManager.

I actually realized that the user tasks are not ordered either, it's not just an issue for group and challenge tasks. They get ordered on the client here https://github.com/HabitRPG/habitica/blob/develop/website/client/src/store/actions/tasks.js#L46

I'm checking with the rest of the staff if there's a reason for doing it on the client but if there's probably it should be done there for challenges and groups tasks as well, otherwise we can remove the order on client code for user tasks as well

paglias:
Actually, it looks like they don't get ordered because a single array is returned instead of them divided by type in 4 different arrays. It's definitely something we should change and will add it to the list of API improvements but for now I think ordering on the client only is better since the apps and website already have the code to handle them

Alys:
I'm closing this but I'm linking to it from the top post of the associated issue so that another contributor can refer to the changes and comments here. Thanks @alecbrick for the work so far!

PR #12034, opened 2020-03-31
cancelSubscription function is no longer called twice when user leaves group fixes #11664

### LINKED ISSUE #11664: extraMonths lost when free subscription from Group Plan is terminated
If you have a free subscription from a Group Plan and have `extraMonths` stored up from a previous subscription, you lose those extra months when you leave the Group Plan.

Steps to reproduce:
1. Add a Group Plan to a guild (you can learn how to do that on a local install at [Using Your Local Install to Modify Habitica's Website and API](http://habitica.fandom.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API)).
2. Invite a another user to the guild.
3. Make that user accept the invitation.
4. Modify that user's subscription data to set `purchased.plan.extraMonths` to any non-zero value (e.g., 12).
5. Make that user leave the guild.
6. Observe that the user's `purchased.plan.dateTerminated` value is a couple of days in the future and the user's `purchased.plan.extraMonths` is now zero.

The expected behaviour is that the user's `purchased.plan.dateTerminated` value should be as far in the future as the number of extraMonths that they had (e.g., one year in the future if you set extraMonths to 12).

----

To illustrate the bug, here's subscription data from a test user account while it was in a Group Plan with `// XXXXXXX` marking the fields of interest:
```
{
    "_id": "a56d9849-a7a4-4a1d-9653-248a5f821fd7",
    "purchased": {
        "plan": {
            "consecutive": { "count": 7, "offset": 0, "gemCapExtra": 10, "trinkets": 2 },
            "quantity": 1,
            "extraMonths": 12.052668652006172, // XXXXXXXXXXXXXXXXXXXXXX
            "gemsBought": 0,
            "mysteryItems": [ ],
            "planId": "group_plan_auto",
            "customerId": "group-plan",
            "dateUpdated": "2019-12-27T11:04:06.397Z",
            "paymentMethod": "Group Plan",
            "dateTerminated": null, // XXXXXXXXXXXXXXXXXXXXXXXXXXX
            "dateCreated": "2019-05-11T23:39:32.913Z",
            "owner": "a56d9849-a7a4-4a1d-9653-248a5f821fd7",
            "lastBillingDate": null,
            "nextPaymentProcessing": null,
            "nextBillingDate": null,
            "additionalData": null
        }
    },
}
```

Here's the same test user account after leaving the guild that had the Group Plan:
```
{
    "_id": "a56d9849-a7a4-4a1d-9653-248a5f821fd7",
    "purchased": {
        "plan": {
            "consecutive": { "count": 7, "offset": 0, "gemCapExtra": 10, "trinkets": 2 },
            "quantity": 1,
            "extraMonths": 0, // XXXXXXXXXXXXXXXXXXXXXX
            "gemsBought": 0,
            "mysteryItems": [ ],
            "planId": "group_plan_auto",
            "customerId": "group-plan",
            "dateUpdated": "2019-12-27T11:04:06.397Z",
            "paymentMethod": "Group Plan",
            "dateTerminated": "2019-12-29T00:00:00.000Z", // XXXXXXXXXXXXXXXXXXXXXX
            "dateCreated": "2019-05-11T23:39:32.913Z",
            "owner": "a56d9849-a7a4-4a1d-9653-248a5f821fd7",
            "lastBillingDate": null,
            "nextPaymentProcessing": null,
            "nextBillingDate": null,
            "additionalData": null
        }
    },
```
---

The bug is happening because when you leave the guild, the `cancelGroupSubscriptionForUser()` function is run twice (you can confirm that on a local install with a console.log line in that function). Here's the two places that run it:

https://github.com/HabitRPG/habitica/blob/33b13a86da42c255bcdff9193b04c972104c7bb7/website/server/models/group.js#L1362
https://github.com/HabitRPG/habitica/blob/33b13a86da42c255bcdff9193b04c972104c7bb7/website/server/controllers/api-v3/groups.js#L865

That function calls `cancelSubscription()` as you can see at the bottom of this code:
https://github.com/HabitRPG/habitica/blob/33b13a86da42c255bcdff9193b04c972104c7bb7/website/server/libs/payments/groupPayments.js#L214-L243

So the `cancelSubscription()` function is run twice too. That function is responsible for calculating `dateTerminated` and then setting `extraMonths` to zero, as shown here:
https://github.com/HabitRPG/habitica/blob/33b13a86da42c255bcdff9193b04c972104c7bb7/website/server/libs/payments/subscriptions.js#L332-L344

The first time that `cancelSubscription()` runs, it correctly reads the value of `extraMonths`, correctly sets `dateTerminated`, then correctly sets `extraMonths` to zero - all what it's supposed to do.
The second time it runs, because `extraMonths` is now set zero, `dateTerminated` is calculated without using the proper value of `extraMonths`, so the date is wrong.
You can confirm all that too with console log lines in `cancelSubscription()` that print out the values of `extraMonths` and `dateTerminated`.

---

I'm going to leave this as suggestion-discussion for a couple of days in case Habitica's staff want to comment on this or state that they'll fix it, then if there's no objections, I'll mark it as help wanted.

hamboomger:

Fixes #11664

### Changes
* cancelSubscription() is no longer called twice when user leaves group with active group plan.
* unit tests created for calculating the terminationDate.

----
UUID: 

### DIFF
--- a/test/api/v3/integration/groups/POST-groups_groupId_leave.test.js
+++ b/test/api/v3/integration/groups/POST-groups_groupId_leave.test.js
@@ -2,6 +2,7 @@ import { v4 as generateUUID } from 'uuid';
 import {
   each,
 } from 'lodash';
+import moment from 'moment';
 import {
   generateChallenge,
   checkExistence,
@@ -12,6 +13,7 @@ import {
 } from '../../../../helpers/api-integration/v3';
 import { model as User } from '../../../../../website/server/models/user';
 import payments from '../../../../../website/server/libs/payments/payments';
+import { calculateSubscriptionTerminationDate } from '../../../../../website/server/libs/payments/util';
 
 describe('POST /groups/:groupId/leave', () => {
   const typesOfGroups = {
@@ -335,6 +337,50 @@ describe('POST /groups/:groupId/leave', () => {
         await member.post(`/groups/${groupWithNoPlan._id}/leave`);
         await member.sync();
         expect(member.purchased.plan.dateTerminated).to.not.exist;
+      });
+    });
+  });
+
+  each(typesOfGroups, (groupDetails, groupType) => {
+    context(`Leaving a group with extraMonths left plan when the group is a ${groupType}`, () => {
+      const extraMonths = 12;
+      let groupWithPlan;
+      let member;
+
+      beforeEach(async () => {
+        const { group, members } = await createAndPopulateGroup({
+          groupDetails,
+          members: 1,
+          upgradeToGroupPlan: true,
+        });
+        [member] = members;
+        groupWithPlan = group;
+        await member.update({
+          'purchased.plan.extraMonths': extraMonths,
+        });
+      });
+      it('calculates dateTerminated and sets extraMonths to zero after user leaves the group', async () => {
+        const userBeforeLeave = await User.findById(member._id).exec();
+
+        await member.post(`/groups/${groupWithPlan._id}/leave`);
+        const userAfterLeave = await User.findById(member._id).exec();
+
+        const dateTerminatedBefore = userBeforeLeave.purchased.plan.dateTerminated;
+        const extraMonthsBefore = userBeforeLeave.purchased.plan.extraMonths;
+        const dateTerminatedAfter = userAfterLeave.purchased.plan.dateTerminated;
+        const extraMonthsAfter = userAfterLeave.purchased.plan.extraMonths;
+
+        const expectedTerminationDate = calculateSubscriptionTerminationDate(null, {
+          customerId: payments.constants.GROUP_PLAN_CUSTOMER_ID,
+          extraMonths,
+        }, payments.constants);
+
+        expect(extraMonthsBefore).to.gte(12);
+        expect(extraMonthsAfter).to.equal(0);
+        expect(dateTerminatedBefore).to.be.null;
+        expect(dateTerminatedAfter).to.exist;
+
+        expect(moment(dateTerminatedAfter).diff(expectedTerminationDate, 'days')).to.equal(0);
       });
     });
   });
--- a/website/server/controllers/api-v3/groups.js
+++ b/website/server/controllers/api-v3/groups.js
@@ -861,11 +861,6 @@ api.leaveGroup = {
       if (guildIndex >= 0) user.guilds.splice(guildIndex, 1);
     }
 
-    const isMemberOfGroupPlan = await user.isMemberOfGroupPlan();
-    if (!isMemberOfGroupPlan) {
-      await payments.cancelGroupSubscriptionForUser(user, group);
-    }
-
     if (group.hasNotCancelled()) await group.updateGroupPlan(true);
     res.respond(200, {});
   },
--- a/website/server/controllers/api-v3/groups.js
+++ b/website/server/controllers/api-v3/groups.js
@@ -1315,7 +1315,7 @@ api.getGroupPlans = {
       .select('leaderOnly leader purchased name managers')
       .exec();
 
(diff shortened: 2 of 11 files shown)


hamboomger:
Hmm, I tried to run failed test locally and it passes, and what's even stranger is that this test is not even related to any changes I've made

paglias:
@hamboomger I'll handle any conflict

test/api/v3/integration/groups/POST-groups_groupId_leave.test.js
```diff
@@ -338,4 +340,48 @@ describe('POST /groups/:groupId/leave', () => {
...
+
+        await member.post(`/groups/${groupWithPlan._id}/leave`);
+        const userAfterLeave = await User.findById(member._id).exec();
+
+        const dateTerminatedBefore = userBeforeLeave.purchased.plan.dateTerminated;
+        const extraMonthsBefore = userBeforeLeave.purchased.plan.extraMonths;
+        const dateTerminatedAfter = userAfterLeave.purchased.plan.dateTerminated;
+        const extraMonthsAfter = userAfterLeave.purchased.plan.extraMonths;
+
+        const expectedTerminationDate = calculateSubscriptionTerminationDate(null, {
```
paglias, line 373: about this check, we know that the server will call `calculateSubscriptionTerminationDate` so if we check the server result against the `calculateSubscriptionTerminationDate` call we risk not catching  issues with `calculateSubscriptionTerminationDate`. Could we change `expectedTerminationDate` to a static value that can be checked against the server result?

If it becomes difficult to know the exact date we can relax the check at line 383 to return true even if there's a slight difference

paglias (commented):
@hamboomger sorry again for the delay, I left just one comment about a test, but if you don't want to fix it I understand , just let me know and I'll fix it

test/api/v3/integration/groups/POST-groups_groupId_leave.test.js
hamboomger: @paglias I think any issues with the `calculateSubscriptionTerminationDate` function should be discovered by the unit tests I wrote for this exact function (and the function is pure, no side effects)

paglias: Ok but if the functions returned a termination date in 2050 we wouldn't notice because `dateTerminatedAfter` and `expectedTerminationDate` are both calculated the same way. We should either have unit tests on `calculateSubscriptionTerminationDate` or the integration test should be `we know that this subscription termination date should be set in ~12 months from now`, after calling the `leave` route it's set to the correct value? Does it make sense?

hamboomger: > We should either have unit tests on `calculateSubscriptionTerminationDate` or the integration test should be...

Hmm, but we **do** have unit tests on `calculateSubscriptionTerminationDate`, it's in the same commit, first file

paglias: sorry, you're right I don't know how I missed

website/server/controllers/api-v3/groups.js
```diff
@@ -861,11 +861,6 @@ api.leaveGroup = {
       if (guildIndex >= 0) user.guilds.splice(guildIndex, 1);
     }
 
-    const isMemberOfGroupPlan = await user.isMemberOfGroupPlan();
```
paglias, line 864: I'm merging and also deleting the lines 859-862 which aren't used anymore now

paglias:
Sorry for the mess with the review, welcome to the 2nd tier and thanks for the PR!

hamboomger:
@paglias no problem, thanks for the second tier!

PR #12035, opened 2020-04-01
Admin Panel to replace admin features in Hall of Heroes

Alys:
Partial fix for https://github.com/HabitRPG/habitica-private/issues/119

This is still in progress but I'm doing a PR now to check that there's nothing significantly wrong with the approach. Upcoming changes are listed in the issue.

[As discussed](https://habitica.slack.com/archives/C02V10HPE/p1581933204139600), this brings the admin panel back into the site as a separate module under /website
Other advantages are that it'll be easier for Blacksmiths to use for adjusting their test accounts, and a bit faster for mods to get to (useful in emergencies).

It's access by a Help menu item that's visible to only mods and staff:
image

Initial view:
image

After a user has been loaded:
image

Samples from the sections:

image

image

Player is not in a party:
image

Player is in a party and quest:
image

image

image

image

image

image

image

image


Alys:
The recent commits have added English names for all of the items.

For equipment pieces, descriptions of the gear sets have been added too, where possible - this includes some of the sets that are listed in the `content` and some sets that I made up because they'll be useful when doing user support (e.g., if a player is asking about missing gear from the Check-In Incentives). Links to wiki page are included when they might be useful.

The screenshot below shows examples of item names (gear and eggs) and gear sets.

image

paglias (commented):
I haven't reviewed it deeply but it looks good, the only thing I understood this was going to be developed separately from the main Vue app so for example in `/website/admin-panel` to keep things separate. But if that would require some more work and as this seems an improvement of the hall of heroes functionality and not something completely new I guess it's okay. Do we want to remove the existing code in the hall of heroes?

Alys:
> I understood this was going to be developed separately from the main Vue app so for example in /website/admin-panel to keep things separate

Ah I'm sorry! I thought that was only needed if it used different libraries and I managed to get it working without them, but I misunderstood. I'm happy to adjust it so that it's separate.

> Do we want to remove the existing code in the hall of heroes?

Yes, definitely. Once I'm sure it's working correctly and does match the needs of the other mods and Blacksmiths (at least as well as the Hall does), I'll be deleting the Hall code.

How would you feel about this going onto delta now so that the mods can try it? If you'd prefer to wait until I've made it separate, that's fine.

Alys:
For this:
> I understood this was going to be developed separately from the main Vue app so for example in /website/admin-panel to keep things separate

to make sure I have the right idea, is this what I should do?
- create the directory `website/admin-panel/`
- put a package.json file in it (like `website/client/package.json` but without any of the libraries and scripts that aren't needed for the admin panel)
- edit the top-level package.json file:
   - copy the "run serv" and "run build" commands to make similar ones for the admin panel
   - add the admin panel to the lint and postinstall commands
- move `website/client/src/components/admin-panel/index.vue` into a subdirectory under `website/admin-panel/`
- copy any essential components from `website/client` to `website/admin-panel` (or can I keep referring to them from the `website/client` directory?)

paglias:
I think it would be a bit more complicated, it should also handle authentication so if it's just an extended hall of heroes it might not be worth it, but maybe we can discuss it at the meeting? It really depends on what the long term plans for this are

website/client/src/router/index.js
```diff
@@ -47,6 +47,9 @@ const HallPage = () => import(/* webpackChunkName: "hall" */'@/components/hall/i
 const PatronsPage = () => import(/* webpackChunkName: "hall" */'@/components/hall/patrons');
 const HeroesPage = () => import(/* webpackChunkName: "hall" */'@/components/hall/heroes');
 
+// Admin Panel
+const AdminPanelPage = () => import(/* webpackChunkName: "admin-panel" */'@/components/admin-panel');
```
paglias, line 54: I think that other than the menu item being hidden, if someone tries to open the route and they're not allowed they should be redirected somewhere else, see here https://github.com/HabitRPG/habitica/blob/develop/website/client/src/router/index.js#L356-L368 for an example

website/client/src/components/admin-panel/index.vue
```diff
@@ -0,0 +1,754 @@
...
+  <div
+    v-if="user.contributor.admin"
+    class="row standard-page"
+  >
+    <div class="well">
+      <h1>Admin Panel</h1>
+      <div class="row">
+        <div class="form-inline col-12">
+          <div class="form-group">
+            <input
```
paglias: instead of adding two event listeners, one for the input and one for the button we should use a form where the button has `type="submit"` and the user is loaded on submit event

website/client/src/components/admin-panel/index.vue
```diff
@@ -0,0 +1,754 @@
+<template>
+  <div
+    v-if="user.contributor.admin"
```
paglias, line 3: is this v-if necessary? if we add a redirect when not an admin I think it can be removed

website/client/src/components/admin-panel/index.vue
```diff
@@ -0,0 +1,754 @@
...
+          + 'An admin needs to delete the invitation or assign the Quest key.';
+      }
+      if (questKey) {
+        return `${questKey} : Invitation has been accepted. `
+          + 'Quest is either running or still in invitation stage.';
+      }
+      return 'No (party might have a quest but user is not in it).';
+    },
+  },
+  // async mounted () {
```
paglias: this should be removed

website/client/src/components/admin-panel/index.vue
```diff
@@ -0,0 +1,754 @@
...
+import moment from 'moment';
+
+import markdownDirective from '@/directives/markdown';
+import styleHelper from '@/mixins/styleHelper';
+import { mapState } from '@/libs/store';
+import content from '@/../../common/script/content';
+import notifications from '@/mixins/notifications';
+
+export default {
+  components: {
```
paglias, line 43: this should be removed

website/client/src/components/admin-panel/index.vue
```diff
@@ -0,0 +1,754 @@
...
+    },
+  },
+  // async mounted () {
+  // },
+  methods: {
+    authMethodExists (authMethod) {
+      if (this.hero.auth[authMethod] && this.hero.auth[authMethod].length !== 0) return true;
+      return false;
+    },
+    formatDate (inputDate) {
```
paglias: this should probably be a filter

website/client/src/components/admin-panel/index.vue
```diff
@@ -0,0 +1,754 @@
...
+      const date = moment(inputDate).utcOffset(0).format('YYYY-MM-DD HH:mm');
+      return `${date} UTC`;
+    },
+    formatTimeZone (inputTimeZoneOffset) {
```
paglias: filter

website/client/src/components/admin-panel/index.vue
```diff
@@ -0,0 +1,754 @@
...
+      const timeZone = (inputTimeZoneOffset / 60) * -1;
+      const sign = (timeZone < 0) ? '-' : '+';
+      return `${sign}${timeZone} UTC`;
+    },
+    formatEquipment (gearWorn) {
```
paglias: filter

website/client/src/components/admin-panel/index.vue
```diff
@@ -0,0 +1,754 @@
...
+        // no point displaying set details for gear where it's obvious
+        return '';
+      }
+      const klassNames = {
+        healer: 'Healer',
+        rogue: 'Rogue',
+        warrior: 'Warrior',
+        wizard: 'Mage',
+      };
+      const loginIncentivesGear = ['armor_special_bardRobes', 'armor_special_dandySuit', 'armor_special_lunarWarriorArmor', 'armor_special_nomadsCuirass', 'armor_special_pageArmor', 'armor_special_samuraiArmor', 'armor_special_sneakthiefRobes', 'armor_special_snowSovereignRobes', 'back_special_snowdriftVeil', 'head_special_bardHat', 'head_special_clandestineCowl', 'head_special_dandyHat', 'head_special_kabuto', 'head_special_lunarWarriorHelm', 'head_special_pageHelm', 'head_special_snowSovereignCrown', 'head_special_spikedHelm', 'shield_special_diamondStave', 'shield_special_lootBag', 'shield_special_wakizashi', 'shield_special_wintryMirror', 'weapon_special_bardInstrument', 'weapon_special_fencingFoil', 'weapon_special_lunarScythe', 'weapon_special_nomadsScimitar', 'weapon_special_pageBanner', 'weapon_special_skeletonKey', 'weapon_special_tachi'];
```
paglias: for this list and all the others, it's possible to get the value from `common/content` instead of writing them manually here? it's very error prone if we have to update this list every time something changes

website/client/src/components/admin-panel/index.vue
```diff
@@ -0,0 +1,754 @@
...
+          <div class="accordion-group">
+            <h3
+              class="expand-toggle"
+              :class="{'open': expandPriv}"
+              @click="expandPriv = !expandPriv"
+            >
+              Privileges, Gem Balance
+            </h3>
+            <div v-if="expandPriv">
+              <div class="form-group">
```
paglias: this should probably be a form

website/client/src/components/admin-panel/index.vue
```diff
@@ -0,0 +1,754 @@
...
+                <small class="muted">
+                  Enter the <strong>item path</strong>. E.g.,
+                  <code>items.pets.BearCub-Zombie</code> or
+                  <code>items.gear.owned.head_special_0</code> or
+                  <code>items.gear.equipped.head</code>.
+                  You can find all the items in the "Item" section above.
+                  The "Change" buttons will insert the path and the current value automatically.
+                </small>
+                <br>
+                <input
```
paglias: this should be a form

website/client/src/components/admin-panel/index.vue
```diff
@@ -0,0 +1,754 @@
...
+          <div class="accordion-group">
+            <h3
+              class="expand-toggle"
+              :class="{'open': expandContrib}"
+              @click="expandContrib = !expandContrib"
+            >
+              Contributor Details
+            </h3>
+            <div v-if="expandContrib">
+              <div class="form-group form-inline">
```
paglias: should be a form

paglias:
Do we want to add support for the new secret field? https://github.com/HabitRPG/habitica/pull/12121

The descriptions in the screenshot below should be under the input?
Schermata 2020-05-03 alle 19 31 29

Also I'd like to remove the code to update the contributor status from the hall now that we have this to avoid duplication.

And, last thing, the component is pretty big at 700+ lines, it could be a good idea to split it into smaller ones

--------------

Left some comments but nothing that prevents it from being tested on delta, but I'll need the conflicts to be resolved

paglias:
I've also tested it and it works well!

Alys:
@paglias   Thanks very much for all the comments! I'll fix the conflicts then work through the comments.

> Also I'd like to remove the code to update the contributor status from the hall now that we have this to avoid duplication.

Please don't do that yet! I'd like to be sure the new code works as needed in production before the admin features in the Hall are removed. If for any reason moderators lose access to the ability to adjust contributor information and other user details, it could cripple our ability to moderate effectively.

paglias:
> Please don't do that yet! I'd like to be sure the new code works as needed in production before the admin features in the Hall are removed. If for any reason moderators lose access to the ability to adjust contributor information and other user details, it could cripple our ability to moderate effectively.

No worries, I meant as part of this PR after it's been extensively tested. Or in a separate PR later, just so that we don't forget.

> @paglias Thanks very much for all the comments! I'll fix the conflicts then work through the comments.

My comments are mostly about style / organization / some html changes for usability (the form element) but nothing that prevents it from being tested so I think it can go to delta as soon as the conflicts are fixed and the rest can be done separately

paglias:
Given that the Bailey API PR is almost ready (#10739) I don't know if we want to add the UI for that here

Alys:
Is it better to keep the Bailey CMS separate from these admin features? They have very different purposes and different users (the CMS is just staff; the user-modification admin features will be used by mods, staff, plus contributors locally). They might need different libraries too.

If it's a problem that this PR uses the name "admin panel" I can easily replace that with something else.

paglias:
I don't think the name is an issue, I think we can decide when the Bailey PR has been tested on delta

website/client/src/router/index.js
Alys: I've tried that in this commit:
https://github.com/HabitRPG/habitica/pull/12035/commits/a4f9c754ad3886466925de4d8ed9a5ac55d2f999
but it needs help as shown in TODO comment near the bottom of the commit.

By the way, I've done it with a `requiresAdmin` true/false option but it could also be done with an integer option called `requiresMinimumContributorTier` which could be set to 8 for the user editing feature (mods and staff) and 9 for the CMS (staff only), with 0 as the default. Let me know if you prefer that! (The TODO comment would still need to be fixed - that's not dependent on which option is used.)

website/client/src/components/admin-panel/index.vue
Alys: I don't think that would be safe enough. If a future change puts a bug into the router redirection code, anyone would be able to see and partially use the admin panel. The API would prevent them from viewing private data for any other user and would prevent them from making changes they're not authorised for BUT that would not not be obvious to them. It would appear to them that they had obtained access to a feature that totally breaks user privacy. The negative publicity could be very damaging if it spread widely.

Even though the chance of such a bug happening may be small, it's not impossible and the consequences could be bad. I don't think it's worth the risk for the sake of removing a v-if. :)

Alys:
@paglias I've added some commits but ignore all of them for now except the one discussed in this comment:
https://github.com/HabitRPG/habitica/pull/12035#discussion_r456129030

All the others will be easier to review when I've finished breaking up the file into components. I've also been working through all your other comments of course, but for now I pushed just enough to get that one commit that I wanted to talk about into the PR. :)

website/client/src/router/index.js
paglias: Hmm I see, the issue is that the user is not loaded yet at this point, looking at the Vue router docs these navigation guards can return a promise so we can try transforming the function to an async function but to avoid it impacting all routes it should wait for the user to be loaded only when admin privileges are required:

- I would remove `const isAdmin`
- transform the if to something like this

```
if (routeRequiresAdmin) {
    await this.$store.dispatch('user:fetch')
    const isUserAdmin = isUserLoaded && store.state.user.data.contributor
    && store.state.user.data.contributor.admin;

    ....
}
```

Note that I haven't tried the fix

website/client/src/components/admin-panel/index.vue
paglias: I'm not sure I agree :) Yeah we could have a bug that removes the redirection check but I don't think it's a good reason to keep the extra check in place, as you said the API access would still be restricted and it would require a bunch of things to go wrong:

1. A bug in the redirection code
2. No one noticing it
3. Someone without admin privileges that manually types the url 

It's not like the extra check is a big deal to maintain but I don't think it's necessary or gives any advantage, we don't have the same check for all the app pages where login is necessary and as long as API data is protected (it is and even if it wasn't there would be other ways to access it) I think the chances of this being useful are so slim it's not worth it

paglias: I don't think it's useful and usually I don't like having extra checks when not necessary but if you want to keep it that's fine

Alys:
This is still WIP and not ready for a full review but I have a couple of questions that I'm about to post.

There may be some lint errors but ignore those. I'll clean them up later. (I'm running a slightly less strict lint config at the moment to make development easier.)

website/client/src/components/admin-panel/index.vue
Alys: I was definitely happy to do that! :) I thought I needed to use similar ways of submitting that were already in use but I'd much rather use normal HTML forms. :) 

Done in 26fdcbbee51937d39a149a20bf7ccd1cf674ca18 and 4a46661fbf6f22d29583fb700634b35d05d9431f but let me know if you want other improvements there.

website/client/src/components/admin-panel/index.vue
Alys: Done in 26fdcbbee51937d39a149a20bf7ccd1cf674ca18

website/client/src/components/admin-panel/index.vue
Alys: Done in 26fdcbbee51937d39a149a20bf7ccd1cf674ca18

website/client/src/components/admin-panel/index.vue
Alys: done in d57008704e7ebf9ef8c27ed461a6423397302794

website/client/src/components/admin-panel/index.vue
Alys: also done in d57008704e7ebf9ef8c27ed461a6423397302794

website/client/src/components/admin-panel/index.vue
Alys: I haven't done this yet because I can't work out how to call the getItemDescription() method from within the filter. Based on what I've read, I'm not sure if it's possible. How would you like me to handle that?

website/client/src/components/admin-panel/index.vue
Alys: Done.

website/client/src/components/admin-panel/index.vue
Alys: It's not really possible to do that, at least not without a lot of extra code (e.g., parsing through all the quests code to pull out data from the rewards), and even then the wiki links would have to be inserted manually.  I don't think the extra code is worth it. 

These are mostly items that won't change at all often or ever, and if new items like these are added to the content, it won't be disastrous if they aren't described like this.

website/client/src/components/admin-panel/index.vue
Alys: Done

Alys:
> the component is pretty big at 700+ lines, it could be a good idea to split it into smaller ones

I'm afraid I'm stuck on this. This simple case works (although if you want me to change any of that I'm happy to):
https://github.com/HabitRPG/habitica/pull/12035/commits/80cedabf1a159309fd16dce8ded726d1419a8182

This more complex case I'm having trouble with:
https://github.com/HabitRPG/habitica/pull/12035/commits/b0cb3f9fd301872bafd8e49b92de23bc308ecde4
It all works fine when you fetch data for one user. However if you try to fetch for a second user, the second user's data is all correct, but the component doesn't reset the automatic display of error messages and the automatic expanding of the section when there's an error.
E.g., If the first user's data had an error, when you fetch a new user you still see the error messages even though you can see that the second user's data does not have the error.

I've tried a few things to re-run the checks that display the error messages(e.g., moving the code to a function and running the function when a prop changes), and I've tried to completely re-instantiate the component when a new user is fetched, but I haven't been successful. What's the best way to do something like that?

paglias:
> This more complex case I'm having trouble with:
b0cb3f9
It all works fine when you fetch data for one user. However if you try to fetch for a second user, the second user's data is all correct, but the component doesn't reset the automatic display of error messages and the automatic expanding of the section when there's an error.

Do you have an example of this

> I've tried a few things to re-run the checks that display the error messages(e.g., moving the code to a function and running the function when a prop changes), and I've tried to completely re-instantiate the component when a new user is fetched, but I haven't been successful. What's the best way to do something like that?

it might be that it doesn't recognize changes to objects, you could try passing the `heroId` as a prop and reset the errors when it changes with:

```
....
watch: {
    heroId () { ... reset errors here ...}
}
....
```

website/client/src/components/admin-panel/index.vue
paglias: it might be that filters don't support accessing `this`, let's keep it as a method then!

Alys:
Oh!!! I think I know what I doing was wrong. I did try watching a prop and that didn't help BUT I think I was relying on this magically executing itself each time a new user was fetched: 

    return {
      cronError: false,
      timezoneDiffError: false,
      timezoneMissingError: false,
      errorsOrWarningsExist: false,
      expand: false,
    };

I'll try again. Don't spend any more of your time thinking about it unless I scream for more help. :) Thanks very much!

Alys:
Sorry, hit enter too soon.

So yeah I'll try again. Don't spend any more of your time thinking about it unless I scream for more help. :) Thanks very much!

website/client/src/components/admin-panel/index.vue
paglias: okay ! This seemed a bit of anightmare to keep up to date but if you think it won't have to be updated often and automating it would be too complex we can keep it like this

Alys:
Yeah, the prop watch method works; when I tried it before I forgot to actually set the errors to false first. `facepalm`

Is this okay? https://github.com/HabitRPG/habitica/pull/12035/commits/3d9757a1f543d2ef3299b71d79c1790163a7a6e1

It has to watch a a property that isn't directly connected to a user because the user you fetch the second time may actually be the first user, just after you've fixed their data with a database update command. So instead of watching heroId, I'm watching an integer counter called `resetCounter`. Let me know if you want changes to its name or anything.

If you're happy with this I'll finish splitting the main component into smaller ones, then work through the remaining fixes, and then add some new features.

paglias:
@Alys that is looking good, sorry for the delay

Alys:
This is not urgent at all but you may want to look at c076d1be53c1d676a6ab86e3f5f225a6091a2272 specifically because it fixes a bug that I think must have been in the Hall code for a while. If it's not self-explanatory, let me know.

Alys:
> the component is pretty big at 700+ lines, it could be a good idea to split it into smaller ones

That's finished, finally! So sorry it took so long, there was a learning curve. For the Items section, I also made significant improvements to how the items are edited - there's a small in-place form for each item and the old "Update Items" section is now deleted (screenshot below).

I still have more to do of course, including your other comments here that I haven't finished yet, but if you think this is good enough to go to delta again, would that be possible (assuming tests pass)? I'd like to get feedback from the other mods about the new Item editing mechanism.

image

Alys:
For the rest of this comment from paglias: https://github.com/HabitRPG/habitica/pull/12035#issuecomment-623149722

> Do we want to add support for the new secret field? https://github.com/HabitRPG/habitica/pull/12121

That's done now in https://github.com/HabitRPG/habitica/pull/12035/commits/358d8773801f2de55df261cfb62a29eac866c72a (screenshot below).



> The descriptions in the screenshot below should be under the input?

I wasn't able to reproduce that layout problem (Firefox on Ubuntu). If you're still seeing it after the next push to delta, can you give me browser details?



> Also I'd like to remove the code to update the contributor status from the hall now that we have this to avoid duplication.

I've added that at https://github.com/HabitRPG/habitica-private/issues/119#issuecomment-598802430 so we can't forget it.


image

paglias:
@Alys it's headed up to delta

paglias:
Schermata 2020-09-05 alle 13 03 27

I'm still seeing the description next to the input, I'm using Edge but it's the same on Safari / Firefox / Chrome, it only happens when the page is more than 1189px wide

website/client/src/router/index.js
Alys: That worked! Thanks! It's done here: https://github.com/HabitRPG/habitica/pull/12035/commits/566d9a9a95358073a971f30372398d199b62da64

Alys:
> I'm still seeing the description next to the input, I'm using Edge but it's the same on Safari / Firefox / Chrome, it only happens when the page is more than 1189px wide

Ah thanks! That's fixed in https://github.com/HabitRPG/habitica/pull/12035/commits/9d8eb7de0037ca9d5bea2cca750960738f1a8089 but tell me if you'd prefer different html code for it.

---

https://github.com/HabitRPG/habitica/pull/12035/commits/49e410dc808a38f2c0a4b7374b8885792a369356 allows a user to be loaded by specifying their Username or User ID in the URL. That was the biggest blocker for removing the mod tools from the Hall of Heroes. That commit includes:
- moving most of the code from the top-level admin-panel index.vue file to a new index.vue file in the user-support subdirectory
- adding a child in the router for specifying the user in the URL
- making the User ID links in the Hall point to the admin panel
- adding an admin panel link in the admin section of the user profile modal

---

I'm about to fix the conflict that I've just noticed and after that it will be ready for review again, when you have time.

Alys:
Conflict fixed, tests passing.

paglias:
Updated PR is on its way to delta

Alys:
Thanks for deploying to delta!

The latest set of commits adds a feature to show the user's API Token (only some of its characters) and to change it for a new random one.

It's ready for review but it's totally fine if you don't have time for a while.


Before change:

image

After change:

image

Alys:
The most recent two commits:
- remove the obscured version of the API Token (it will return later with tighter permissions)
- add the Custom Day Start time (a quick fix for a request from shanaqui)

image

Alys:
Tests are passing. This is ready for review.

website/server/controllers/api-v3/hall.js
```diff
@@ -310,16 +321,63 @@ api.updateHero = {
...
+
+    const query = { _id: groupId };
+
+    const party = await Group
+      .findOne(query)
+      .select(heroPartyAdminFields)
+      .exec();
+
+    if (!party) throw new NotFound(apiError('groupWithIDNotFound', { groupId }));
+    const partyRes = party.toJSON({ minimize: true });
```
paglias, line 378: is there a reason for using `minimize: true` here? We set it to `false` by `default` so that all objects are returned to the client even if empty, removing the need to check if an object is empty or not before using it on the client

website/server/controllers/api-v3/hall.js
```diff
@@ -310,16 +321,63 @@ api.updateHero = {
...
+  middlewares: [authWithHeaders(), ensureAdmin],
+  async handler (req, res) {
+    req.checkParams('groupId', apiError('groupIdRequired')).notEmpty().isUUID();
+
+    const validationErrors = req.validationErrors();
+    if (validationErrors) throw validationErrors;
+
+    const { groupId } = req.params;
+
+    const query = { _id: groupId };
```
paglias, line 370: should this also have a `type: 'party'`? Otherwise it can be used to fetch guilds as well

website/client/src/router/index.js
```diff
@@ -50,6 +50,10 @@ const HallPage = () => import(/* webpackChunkName: "hall" */'@/components/hall/i
...
+const AdminPanelUserPage = () => import(/* webpackChunkName: "admin-panel-user" */'@/components/admin-panel/user-support');
```
paglias, line 55: `webpackChunkName` here should be `admin-panel` as well because it'll make sure the two files are loaded at the same time

website/client/src/router/index.js
```diff
@@ -338,6 +342,22 @@ const router = new VueRouter({
...
+
+    {
+      name: 'adminPanel',
+      path: '/admin-panel',
+      component: AdminPanelPage,
+      meta: { requiresAdmin: true },
+      children: [
+        {
+          name: 'adminPanelUser',
+          path: '/admin-panel/:userIdentifier', // User ID or Username
```
paglias, line 354: shouldn't path here be simply `:userIdentifier`?

test/api/v3/integration/hall/GET-hall_heroes_party_groupId.test.js
```diff
@@ -0,0 +1,64 @@
...
+      message: t('noAdminAccess'),
+    });
+  });
+
+  it('validates req.params.groupId', async () => {
+    await expect(user.get('/hall/heroes/party/invalidUUID')).to.eventually.be.rejected.and.eql({
+      code: 400,
+      error: 'BadRequest',
+      message: t('invalidReqParams'),
+      // message: apiError('groupIdRequired'), // XXX why doesn't this work?
```
paglias, line 33: this is expected because validation errors are all bundled together under an `invalidReqParams` the detailed errors are under `res.errors`

test/api/v3/integration/hall/GET-hall_heroes_party_groupId.test.js
```diff
@@ -0,0 +1,64 @@
...
+      '_id', 'id', 'balance', 'challengeCount', 'leader', 'leaderOnly', 'memberCount',
+      'purchased', 'quest', 'summary',
+    ]);
+    expect(partyRes.summary).to.eq(' ');
+    // NB: 'summary' is NOT a field that the API route retrieves!
+    // It must not be retrieved for privacy reasons.
+    // However the group model automatically adds a summary for reasons given here:
+    // https://github.com/HabitRPG/habitica/blob/8da36bf27c62ba0397a6af260c20d35a17f3d911/website/server/models/group.js#L161-L170
+
+    // XXX need something like this? expect(heroRes.profile).to.have.all.keys(['name']);
```
paglias, line 62: should this be removed?

website/client/src/components/admin-panel/index.vue
```diff
@@ -0,0 +1,76 @@
...
+        <form
+          class="form-inline"
+          @submit.prevent="loadHero(userIdentifier)"
+        >
+          <input
+            v-model="userIdentifier"
+            class="form-control"
+            type="text"
+            :placeholder="'User ID or Username; blank for your account'"
+            :style="{ 'min-width': '45ch' }"
```
paglias, line 19: we should use a class here instead of a style binding since there's nothing dynamic

website/client/src/components/admin-panel/user-support/index.vue
```diff
@@ -0,0 +1,175 @@
...
+          :hero="hero"
+          :reset-counter="resetCounter"
+          @clear-data="clearData"
+        />
+      </div>
+    </div>
+  </div>
+</template>
+
+<style lang="scss">
```
paglias, line 57: this should be `scoped` so it's not applied across the website

website/client/src/components/admin-panel/user-support/index.vue
```diff
@@ -0,0 +1,175 @@
...
+    clearData () {
+      this.hero = {};
+    },
+
+    async loadHero (userIdentifier) {
+      const id = userIdentifier.replace(/@/, ''); // allow "@name" to be entered
+      this.$emit('changeUserIdentifier', id); // change user identifier in Admin Panel's form
+
+      const hero = await this.$store.dispatch('hall:getHero', { uuid: id });
+      this.hero = { ...hero };
```
paglias, line 144: `hero` is a prop so we should always change it from the upstream component, in this case I'd fetch the data in the admin panel component when `changeUserIdentifier` is emitted

website/client/src/components/admin-panel/user-support/index.vue
```diff
@@ -0,0 +1,175 @@
...
+
+      if (!this.hero.flags) {
+        this.hero.flags = {
```
paglias, line 147: unless we use `minimize` on the server `this.hero.flags` will always exist

website/client/src/components/admin-panel/user-support/index.vue
```diff
@@ -0,0 +1,175 @@
...
+      }
+
+      this.hasParty = false;
+      this.partyNotExistError = false;
+      if (this.hero.party && this.hero.party._id) {
+        let party;
+        try {
+          party = await this.$store.dispatch('hall:getHeroParty', { groupId: this.hero.party._id });
+          this.hasParty = true;
+          this.party = { ...party };
```
paglias, line 160: I don't think `...party` is necessary when replacing an entire property

website/client/src/components/admin-panel/user-support/itemsOwned.vue
```diff
@@ -0,0 +1,287 @@
...
+                  <div
+                    v-if="item.modified"
+                    class="form-inline"
+                  >
+                    <input
+                      v-if="item.valueIsInteger"
+                      v-model="item.value"
+                      class="form-control"
+                      type="number"
+                      :style="{ 'width': '10ch' }"
```
paglias, line 86: class?

website/client/src/components/admin-panel/user-support/itemsOwned.vue
```diff
@@ -0,0 +1,287 @@
...
+              </li>
+            </ul>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+</template>
+
+<style lang="scss">
```
paglias, line 105: scoped

website/client/src/components/admin-panel/user-support/privilegesAndGems.vue
```diff
@@ -0,0 +1,138 @@
...
+        </div>
+        <div class="form-inline">
+          <label>
+            Balance
+            <input
+              v-model="hero.balance"
+              class="form-control"
+              type="number"
+              step="0.25"
+              :style="{ 'width': '15ch' }"
```
paglias, line 53: should be in a class or in a normal style attribute (style="width: 15ch;") not in a dynamic binding,

website/client/src/components/admin-panel/user-support/index.vue
paglias: otherwise it'll only update here because props flow downstream, see https://vuejs.org/v2/guide/components-props.html#One-Way-Data-Flow

paglias:
Schermata 2020-10-01 alle 18 20 17

Layout wise it looks like there are a lot of `row` classes being used without `col-xxx` inside which is causing some issues

paglias:
Schermata 2020-10-01 alle 20 44 47
Might be due to the un-scoped CSS but the header behaves strangely after an user is loaded

Alys:
> Layout wise it looks like there are a lot of `row` classes being used without `col-xxx` inside which is causing some issues

Are there any guidelines about which col-xxx should be used in different situations? I can't find anything in the README files or code comments.

paglias:
the `col-` classes are from Bootstrap and are documented here https://getbootstrap.com/docs/4.5/layout/grid/

If you want to keep the existing layout with a single column a `col-12` per row should be enough. The important is that each row has one or more cols inside

website/client/src/components/admin-panel/user-support/index.vue
Alys: I took that from the original Hall code and it looks like minimize has been used deliberately:

https://github.com/HabitRPG/habitica/blob/efe7ea52bc0e3c27563eb2ebfe9ae5f082be20d4/website/server/controllers/api-v3/hall.js#L190-L195

paglias: ah okay, I can't remember why we're using minimize: true but let's keep it in case it had some useful purpose

website/client/src/components/admin-panel/user-support/index.vue
Alys: I've made the other changes locally but for this one, I wouldn't mind leaving it as it is if that's okay with you? Once other features are added to the admin panel, such as the CMS, it will be neater if all the code for user support is in one sub-directory. 
(I think it still makes sense for the search form to be on the top level page so that you can access it as soon as you reach the admin panel.)

paglias: Sorry, I got confused, I thought `hero` was passed as a prop to this component while it's here that it's fetched the first time and then passed to the other components, so that's fine

PR #12037, opened 2020-04-02
Automatically Logout Banned Users

paglias:
Fixes https://github.com/HabitRPG/habitica-private/issues/126

Also fixes a typo that resulted in an extra call to the database for every authenticated request

### DIFF
--- a/test/api/unit/libs/language.test.js
+++ b/test/api/unit/libs/language.test.js
@@ -1,0 +1,125 @@
+import {
+  getLanguageFromBrowser,
+  getLanguageFromUser,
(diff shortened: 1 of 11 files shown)


Alys:
For the benefit of contributors reading this (e.g., wiki editors):

Automatic logout means that when a moderator or staff member ban an account, the next time the owner of the account takes an action on the website that causes a sync, the website will see that the user has been banned and will log them out immediately.
The player won't see a notification that they've been banned.

If the player tries to log back in to the same account, the login will fail and they will then see a notification saying that they are banned, as shown in the screenshot below.

image

PR #12063, opened 2020-04-06
Removed loading screen tips, fixes #11834

### LINKED ISSUE #11834: loading tips can't be read when loading is too quick
This isn't exactly a bug but...

As reported by @Acorn_Sprout (2a5618da-d8c4-4530-a943-3fa871ef4859):

_"This may already be known and it's not precisely a bug, but when using a browser the introductory tip lasts no more than about a second. FWIW the platform I'm currently using is Chrome version 79.0.3945.130 (Official Build) (64-bit) on Windows 10. With Internet Explorer on the same machine, version 11.1387.15063.0, the tip lasts perhaps two seconds - better, but still hard to read quite everything._

_"I believe Chrome on Android 5.1.1 on a Motorola G3 has similarly brief results, but it's a little hard to quantify because my phone is extremely laggy at present."_

The loading tips were introduced to give people something to read while the website slowly loaded, and it's good that the website now loads for some people too quickly for that to be necessary, but it does make the loading tips look like a bit of a tease. :)

~I don't know if there's anything we could or would want to do to "help" with this (we certainly wouldn't want to make loading deliberately slower) but I'm putting this here so that we have it on record. I'm marking it suggestion-discussion. If it's decided to close it as a won't-fix, we can at least direct people to it for an explanation of why the tips don't show up for long.~

EDIT: See the comments below for the desired fix (in brief, remove the loading tips).

**For anyone who wants to see all the loading tips, they're on the wiki at [Loading Page Tips](https://habitica.fandom.com/wiki/Loading_Page_Tips)**

anoopravella:
Fixes #11834

Removed the div that displayed tips on the loading screen from the app.vue file. Also removed the code that selected the random tip to be displayed, also in app.vue.

When I ran the test suite locally, 11 tests failed, but the same 11 tests also failed when I ran them on the develop (untouched) branch locally.

Could not screenshot the change (as per the development guide) because the loading screen finished too quickly.


paglias:
thanks @anoopravella ! could you also remove the `website/common/locales/en/loadingScreenTips.json` file as it contains the loading tips that we don't need anymore so that it's not loaded?

@Alys do we want to add a note about the file being removed so that we'll remember it's in the git history if we want to add these back?
@SabreCat quick question about Weblate, when removing a `.json` file is it okay to remove it only for English? or should it be removed for every language? Not sure how it's handled

Alys:
> do we want to add a note about the file being removed so that we'll remember it's in the git history if we want to add these back?

Good idea! The wiki's [Loading Page Tips](https://habitica.fandom.com/wiki/Loading_Page_Tips) page already had some information about the code so I've updated that and added a temporary note, linking to the PR (it's all in the Information for Developers section at the bottom - click a button there to reveal it). Once this PR is merged and live, I'll make further changes to the page, and also link to the final commit in `develop`. I'm labelling this as "notify helpers" as a reminder to me. I'll also want to tell the Socialites so they can answer questions about why the tips were removed.

paglias:
thanks @anoopravella ! This will go live by the end of the week, if you have an Habitica UUID i'll be happy to award you a contribution tier

PR #12069, opened 2020-04-09
Don't link user in markdown code blocks - Fixes #[11504]

### LINKED ISSUE #11504: @username inside markdown code backticks should not be converted to link
In group chat messages,  if you type an @Username inside backticks for markdown "code" highlighting, you expect to see exactly what you type with the markdown code formatting (red text on a white background), however the @Username is converted into a link and you see the markdown link code.

For example, if you type exactly this: `` `@username` ``

then you expect to see this: `@username` (in red text on a white background)

but instead you see this: `[@username](/profile/d904bd62-da08-416b-a816-ba797c9ee265)` (and the @Username part is blue instead of red).

See the screenshot below for an example.

I believe this counts as a bug because the markdown code highlighting is intended for displaying exactly the text you type.

<strike>I'll leave this issue as suggestion-discussion for a few days and then if there's no objections I'll mark it as help wanted, with the fix being to not have the @Username converted to a link and to have the @Username appear as red text, not blue text.</strike>
EDIT: 
The desired fix is that when @Username appears inside backticks, it should not be converted to a link and should appear as red text, not blue text.
This should be done for private messages as well as for group posts (see the comment below from Nikosmonaut about PMs).

image

benkelaar:
Fixes #11504 

### Changes
Adjusts the highlightMentions functions to not create links inside code blocks. Uses the markdown-it markdown parser from habitica-markdown in an attempt to guarantee functional parity between the code block rendering and the code block determination for link generation.

However, there's still quite some work required to reason back from the code block tokens to parts of the text strings, so perhaps the test set should be expanded further.

### highlightUsers
I've stayed away from the display rendering in `highlightUsers` for now, since I think that should be done in habitica-markdown instead. See the issue for a more in-depth discussion.

----

### DIFF
--- a/test/api/unit/libs/highlightMentions.js
+++ b/test/api/unit/libs/highlightMentions.js
@@ -13,9 +13,12 @@ describe('highlightMentions', () => {
         return this;
       },
       exec () {
-        return Promise.resolve([{
-          auth: { local: { username: 'user' } }, _id: '111',
-        }, { auth: { local: { username: 'user2' } }, _id: '222' }, { auth: { local: { username: 'user3' } }, _id: '333' }, { auth: { local: { username: 'user-dash' } }, _id: '444' }, { auth: { local: { username: 'user_underscore' } }, _id: '555' },
+        return Promise.resolve([
+          { auth: { local: { username: 'user' } }, _id: '111' },
+          { auth: { local: { username: 'user2' } }, _id: '222' },
+          { auth: { local: { username: 'user3' } }, _id: '333' },
+          { auth: { local: { username: 'user-dash' } }, _id: '444' },
+          { auth: { local: { username: 'user_underscore' } }, _id: '555' },
         ]);
       },
     };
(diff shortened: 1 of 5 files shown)


paglias:
I was slightly worried about the performance implications of the extra markdown parsing but it doesn't seem to be an issue. And the moment we start parsing the markdown on the server it might make sense to store the resulting HTML in the database instead of re-parsing it on the client every time. 

Thanks for the comments and the tests :) I'm going to leave this one in review as we usually deploy PRs on Thursday and if I merge it today it won't have time to be properly tested

benkelaar:
Sounds good, that would also work well with the strategy I proposed for the quick-fix in #12071

As you can tell by the comments in the issue, as I dove into properly figuring out where code blocks are in markdown, I figured out you can not know the distribution of code blocks without knowing the complete distribution of blocks in a markdown document. 

Which means you have to parse all the markdown to make an exception for code blocks. Good to hear it did not significantly impact performance.

paglias:
@benkelaar  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #12071, opened 2020-04-10
Fixes issue where usernames that are sandwiched with underscores are not properly formatted [fixes #12033]

### LINKED ISSUE #12033: Usernames with underscores on each side aren't displayed correctly
If a player has a Username that starts and ends with an underscore (e.g., `@_spider_`), when their Username is mentioned in a chat post, there are three formatting problems:

- the name is in italics (that's kind of expected because underscores indicate italics in markdown but markdown parsing within a Username is not desired) 
- the underscores are not visible (a side-effect of the previous point but this is undesirable because you're not seeing the player's proper Username)
- the text colour of Username is wrong (screenshot below)

The screenshot below was taken from a post that contained this raw text:
`@_spider_ and @blackgoblingo`
Both of those Usernames were in use at the time I made the post (the `_spider_` user has a different name now).
The formatting for the `blackgoblingo` user is correct.
The formatting for the `_spider_` user is wrong.

Both usernames do correctly point to the right profile screens, so that's something. :)

The expected behaviour is that you see `@_spider_` and that it's formatted in the same way as the `blackgoblingo` username.

image

_Thanks to @EliotThe (0d77107e-9128-4656-8c02-6e06331fa396) for reporting this!_

kareenf:

Fixes #12033

### Changes
Markdown uses underscores on both sides of a word to indicate that this word must be italicized. Therefore usernames such as `@_spider_` became @_spider_ and the color was not the usual purple when mentioning this user in a chat. This is not the proper formatting of usernames in Habitica. The usernames become italicized, they lose the underscores and the color is not correct. To fix this, I added in the em tag to the mention regex such that it would catch any time Markdown accidentally turned a username into italicized. Then in the body of the text.replace function, I added in logic to find em tags and replace them with the proper underscores such that the username would now be displayed correctly. I also added a component test to check for this specific issue.   

----
UUID: 5ba76e0c-ed09-4a9b-87f7-b28a8971377d

### DIFF
--- a/website/client/src/libs/highlightUsers.js
+++ b/website/client/src/libs/highlightUsers.js
@@ -17,7 +17,7 @@ export function highlightUsers (text, userName, displayName) { // eslint-disable
     let fixedStr = mentionStr.replace(/<\/?em>/g, '_');
     fixedStr = fixedStr.replace(/<\/?strong>/g, '__');
 
-    const isUserMention = currentUser.includes(mentionStr) ? 'at-highlight' : '';
+    const isUserMention = currentUser.includes(fixedStr) ? 'at-highlight' : '';
 
     return fullMatched.replace(mentionStr, `<span class="at-text ${isUserMention}">${fixedStr}</span>`);
   });
--- a/website/client/src/libs/highlightUsers.js
+++ b/website/client/src/libs/highlightUsers.js
@@ -1,7 +1,7 @@
 import escapeRegExp from 'lodash/escapeRegExp';
 
 const optionalAnchorTagRegExStr = '(<\\w[^>]*)?'; // everything including the anchor tag is recognized
-const mentionRegExStr = '(@[<em>\\w-</em>]+)';
+const mentionRegExStr = '(@(?:<(?:em|strong)>)?[\\w-]+(?:<\\/(?:em|strong)>)?)';
 const optionalPostMentionRegExStr = '(\\.\\w+)?'; // like dot-TLD
 
 const finalMentionRegEx = new RegExp(`${optionalAnchorTagRegExStr}${mentionRegExStr}${optionalPostMentionRegExStr}`, 'gi');
@@ -14,11 +14,9 @@ export function highlightUsers (text, userName, displayName) { // eslint-disable
       return fullMatched;
     }
 
-    let fixedStr = mentionStr;
-    if (mentionStr.includes('<em>')) {
-      fixedStr = mentionStr.replace('<em>', '_');
-      fixedStr = fixedStr.replace('</em>', '_');
-    }
+    let fixedStr = mentionStr.replace(/<\/?em>/g, '_');
+    fixedStr = fixedStr.replace(/<\/?strong>/g, '__');
+
     const isUserMention = currentUser.includes(mentionStr) ? 'at-highlight' : '';
 
     return fullMatched.replace(mentionStr, `<span class="at-text ${isUserMention}">${fixedStr}</span>`);
--- a/website/client/src/libs/highlightUsers.js
+++ b/website/client/src/libs/highlightUsers.js
@@ -1,7 +1,7 @@
 import escapeRegExp from 'lodash/escapeRegExp';
 
 const optionalAnchorTagRegExStr = '(<\\w[^>]*)?'; // everything including the anchor tag is recognized
-const mentionRegExStr = '(@[\\w-]+)';
+const mentionRegExStr = '(@[<em>\\w-</em>]+)';
 const optionalPostMentionRegExStr = '(\\.\\w+)?'; // like dot-TLD
 
 const finalMentionRegEx = new RegExp(`${optionalAnchorTagRegExStr}${mentionRegExStr}${optionalPostMentionRegExStr}`, 'gi');
@@ -14,9 +14,14 @@ export function highlightUsers (text, userName, displayName) { // eslint-disable
       return fullMatched;
     }
 
+    let fixedStr = mentionStr;
+    if (mentionStr.includes('<em>')) {
+      fixedStr = mentionStr.replace('<em>', '_');
+      fixedStr = fixedStr.replace('</em>', '_');
+    }
     const isUserMention = currentUser.includes(mentionStr) ? 'at-highlight' : '';
 
-    return fullMatched.replace(mentionStr, `<span class="at-text ${isUserMention}">${mentionStr}</span>`);
+    return fullMatched.replace(mentionStr, `<span class="at-text ${isUserMention}">${fixedStr}</span>`);
   });
 
   return text;
--- a/website/client/tests/unit/libs/highlightUsers.spec.js
+++ b/website/client/tests/unit/libs/highlightUsers.spec.js
@@ -17,6 +17,14 @@ describe('highlightUserAndEmail', () => {
     expect(result).to.contain('<span class="at-text at-highlight">@user</span>');
   });
 
+  it('highlights username sandwiched with underscores', () => {
+    const text = 'hello @_user_';
+
+    const result = highlightUsers(text, 'user', 'displayedUser');
+    expect(result).to.contain('<span class="at-text at-highlight">@user</span>');
+    expect(result).to.not.contain('<em>');
+  });
+
   it('not highlights any email', () => {
     const text = habiticaMarkdown.render('hello@example.com');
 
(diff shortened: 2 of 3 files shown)


website/client/src/libs/highlightUsers.js
```diff
@@ -1,7 +1,7 @@
 import escapeRegExp from 'lodash/escapeRegExp';
 
 const optionalAnchorTagRegExStr = '(<\\w[^>]*)?'; // everything including the anchor tag is recognized
-const mentionRegExStr = '(@[\\w-]+)';
+const mentionRegExStr = '(@[<em>\\w-</em>]+)';
```
benkelaar: Between square brackets is a character class, i.e. "Match any of the characters between these brackets". This way it's actually not a legal regular expression. 

I would recommend always trying out your regular expression strings in a place like https://regex101.com/

Adding the characters in the character class (like this `(@[<>\\/\\w-]+)` would also match things like `@<<<<<<<<<` or `@<a` which is not desirable. I think to get the effect you're looking for you can try something like this `(@(?:<(?:em|strong)>)?[\\w-]+(?:<\\/(?:em|strong)>)?)`. It maintains the original semantics of the regex but enhances it with optional _em_ or _strong_ tag matches. All the `?:` characters are to prevent these new bracket groups from messing up the existing grouping semantics.

website/client/src/libs/highlightUsers.js
```diff
@@ -14,9 +14,14 @@ export function highlightUsers (text, userName, displayName) { // eslint-disable
       return fullMatched;
     }
 
+    let fixedStr = mentionStr;
+    if (mentionStr.includes('<em>')) {
+      fixedStr = mentionStr.replace('<em>', '_');
+      fixedStr = fixedStr.replace('</em>', '_');
+    }
```
benkelaar: Well done, this should work fine!

However, you could also use a regular expression here to do the two things in a single call. Also since the replace match will simply not match if `<em>` is not present, you can drop the if check as well.

You should be able to replace the entire block by `let fixedStr = mentionStr.replace(/<\/?em>/g, '_');`. Notice that I'm using a _regular expression literal_ here, which is a javascript native way to easily represent regular expressions (You usually only use the `new RegEx` constructor if you're dynamically creating the regular expression, like is being done on line 7)

website/client/src/libs/highlightUsers.js
benkelaar: Gave the expanded regular expression a whirl on regex101: https://regex101.com/r/4nFDHZ/1

It seems to work well (BTW, in the comment above I wrote it as it should be injected in the string. The syntax on regex101 is what it would look like as RegEx literal. The difference is in the escaping of backslashes (\\) ).

kareenf: I'm not very used to RegExs so thank you for your suggestion! I actually did use that website to check my RegEx but I think your RegEx is much more to the effect that I was trying to get at. I'll add it in and check that everything still works!

website/client/src/libs/highlightUsers.js
kareenf: This is great! I was thinking about using a RegEx to do it however I wasn't sure how to go about it so I tried to do it in this more simplified way :) I'll add it in!

website/client/tests/unit/libs/highlightUsers.spec.js
```diff
@@ -17,6 +17,24 @@ describe('highlightUserAndEmail', () => {
     expect(result).to.contain('<span class="at-text at-highlight">@user</span>');
   });
 
+  it('highlights username sandwiched with underscores', () => {
+    const text = 'hello @_user_';
```
benkelaar: This is the wrong input data. Remember that highlightUsers is being called with the __output__ of habiticaMarkdown.parse -> This means that at this point in call stack, the underscores would already have been replaced by emphasis and strong tags. The correct input at this point would be `'hello @<em>user</em>'`.
```suggestion
    const text = 'hello @<em>user</em>';
```

website/client/tests/unit/libs/highlightUsers.spec.js
```diff
@@ -17,6 +17,24 @@ describe('highlightUserAndEmail', () => {
...
+
+    const result = highlightUsers(text, '_user_', 'displayedUser');
+    expect(result).to.contain('<span class="at-text at-highlight">@_user_</span>');
+    expect(result).to.not.contain('<em>');
+    expect(result).to.not.contain('</em>');
+  });
+
+  it('highlights username sandwiched with double underscores', () => {
+    const text = 'hello @__user__';
```
benkelaar: Similar to the test case above, the html tags need to already be present in the input data to provide a realistic scenario.

```suggestion
    const text = 'hello @<strong>user</strong>';
```

website/client/tests/unit/libs/highlightUsers.spec.js
```diff
@@ -17,6 +17,24 @@ describe('highlightUserAndEmail', () => {
...
+
+    const result = highlightUsers(text, '__user__', 'displayedUser');
+    expect(result).to.contain('<span class="at-text at-highlight">@__user__</span>');
```
benkelaar: I would let one of these test cases cover the other branch (The highlighted user is not the logged in user) as to cover more variance in the test cases.

```suggestion
    const result = highlightUsers(text, 'differentUser', 'displayedDifferentUser');
    expect(result).to.contain('<span class="at-text">@__user__</span>');
```

website/client/src/libs/highlightUsers.js
```diff
@@ -14,9 +14,12 @@ export function highlightUsers (text, userName, displayName) { // eslint-disable
       return fullMatched;
     }
 
+    let fixedStr = mentionStr.replace(/<\/?em>/g, '_');
+    fixedStr = fixedStr.replace(/<\/?strong>/g, '__');
+
     const isUserMention = currentUser.includes(mentionStr) ? 'at-highlight' : '';
```
benkelaar: This check should also compare to fixedStr, otherwise you're comparing the name including html tags against the normal user name.

```suggestion
    const isUserMention = currentUser.includes(fixedStr) ? 'at-highlight' : '';
```

benkelaar (changes requested):
Much better like this!

There's one bug still in there though, I noticed it and then was very puzzled that the tests were actually green. I hadn't noticed the problem with the added test cases either since it's a bit of a mind bender!

Almost lucky that you left the bug in!

website/client/tests/unit/libs/highlightUsers.spec.js
kareenf: Oh gosh yes that makes much more sense, great catch. I've added in your suggestions. Thank you so much for all of your help thus far!!

paglias:
Thanks @kareenf ! I'm going through all the markdown PRs today, I see that https://github.com/HabitRPG/habitica/pull/12089 is meant to cover this issue as well? I don't know if you and @benkelaar have already talked about this but let me know if there's some overlapping work. Also I'd be happy to invite you to our Slack channel for open source contributors where it would be easier to discuss

kareenf:
Hi @paglias ! I'm not sure what @benkelaar 's plans/timeline is since it seems as though he wants to do an overhaul of habitica-markdown but if you all want a quick surefire fix that works right now, my PR should take care of issue right now! I'd be honored if my PR was accepted since this is my first time ever open source contributing and I'd love to get an invite for the Slack channel :D

benkelaar:
@kareenf Oh, my apologies - the plugin is also ready 😶

@paglias Is it an option to have this fix in the release tomorrow and the "overhaul" next week? Then we can also fix the remaining backend issues and release that in a single go.

paglias:
@kareenf for the Slack channel, what's your email address? If you don't want to share it publicly just send me a PM on Habitica (`paglias`) or an email at matteopagliazzi@gmail.com

kareenf:
@paglias for Slack, my email is kareenf@umich.edu!

paglias:
Didn't manage to get them merged for today but yeah I think it's a good idea to merge the ones in this repository (will do tomorrow) and then see if we can do a more complete overhaul with the changes in the `habitica-markdown` repo

paglias:
I've also invited both of you to Slack

kareenf:
Awesome, I'm excited to see it! I also just joined the Slack, thanks a lot! @paglias

paglias:
@kareenf  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #12072, opened 2020-04-10
To-Do cron task decay fixes

### LINKED ISSUE #6488: Freeze completed todos value when and un-freeze if they are un)done
Currently completed todos value is updated even after the todo is done, this should be changed so that:
1. After a todo has been completed its value should not change
2. In case a completed todo is moved back to the Active column (when it's undone) its value should be updated to take into account the days after completion when the value remained unchanged. So if it's been done for 2 days when it's undone it's value should be updated to take into account those 2 days and thus the todo should be redder then when it was first archived

### LINKED ISSUE #8590: To-Dos with checked-off checklist items turn red faster

### Description

If a To-Do has checklist items marked off, its task value changes faster than To-Dos with no checklist or with checklists that have no accomplished items.

To reproduce in debug mode:

1. Create several To-Dos.
2. Add checklists to some of them.
3. On some of the To-Dos with checklists, check off one or more items.
4. Advance the day using the debug buttons and sync. To-Dos with accomplished checklist items turn orange immediately, while those without stay yellow.

While a relatively innocuous issue, this is strange behavior that doesn't strike me as intentional. Why should partly-done To-Dos be treated differently than any other To-Do when it comes to cron aging?

SabreCat:
Fixes #6488
Fixes #8590

### Changes

**Previously**, completed To-Dos continued to update their values (becoming "redder") during each day's "cron" operation.
**Now**, completed To-Dos retain their value from one day to the next.

**Previously**, having completed checklist items in a To-Do accelerated the pace of the To-Do becoming redder.
**Now**, checklisted To-Dos decay at the same rate regardless of how many items are checked off within them.


paglias (commented):
LGTM! Do we want to add a test case to cover the new behavior?

Issue #10924, opened 2018-12-30
'@' in URL is incorrectly treated as a Username signifier


Alys:
If you post a URL that contains an '@' symbol, the website treats the URL as if it has a Username in it and it makes the URL unusable both for reading and clicking. For example, this URL was posted:

`https://medium.com/@schlaf/how-to-conduct-an-annual-life-review-that-will-catapult-you-into-the-new-year-d5aaffebac1f`

This is what the visible portion of the link looked like in the website:

`@schlaf/how-to-conduct-an-annual-life-review-that-will-catapult-you-into-the-new-year-d5aaffebac1f" target="_blank" rel="noopener">https://medium.com/@schlaf/how-to-conduct-an-annual-life-review-that-will-catapult-you-into-the-new-year-d5aaffebac1f`

This is the URL that the link tried to send you to:

`https://medium.com/%3Cspan%20class=`

_EDIT 2020-04-10: See the comments for additional important information._

Alys:
One way to fix this and any other similar problems that might exist would be to limit the parsing of Usernames so that it only occurs if @xxxxx is surrounded on both sides by whitespace or by the start/end of the message.

There might be rare cases where that doesn't correctly handle something that should be treated as a Username (e.g., if a player types two Usernames with no spaces between them  @Alys@ALittleYellowSpider) but I think that's an acceptable risk, especially since some examples of @xxxxx@yyyyy might actually not be intended to be two Usernames.

I'll leave this as suggestion-discussion for a few days to give the staff time to comment and then if there's no objections I'll mark it as help wanted.

---

For the benefit of any contributor who will work on this regardless of what solution is chosen:

This is purely a client-side change - the website's Username parsing code in a vue file is what will need to be fixed.

arjunchawla:
Hi Alys, I am willing to contribute and fix this problem.

arjunchawla:
Please let me know if i should go ahead and work on it.

Alys:
I've thought of one case where you would want the Username to be parsed when it has a character directly next to it on the left-hand side: if the user put the Username inside brackets like "(@alys )"
The best way of handling that and any other cases that we haven't thought of might be to automatically insert a space before any Username when it's been selected from the drop-down list. There's already automatically a space inserted after it.

margsliu:
@Alys
Hello, @pskupinski and I are able to take on this issue and could have a pull request submitted within the coming week. Would it be okay for us to start working on it?

Alys:
[...]

PR #12089, opened 2020-04-13
Issue 12033 - Use version of habitica-markdown that includes mentions

### LINKED ISSUE #12033: Usernames with underscores on each side aren't displayed correctly

benkelaar:
Fixes #12033

Also fixes frontend parts of #11504 and #10924 

~Is in draft since it is dependent on https://github.com/HabitRPG/habitica-markdown/pull/87 to be merged first.~

~Edit: Now also dependent on #12069 -> Please merge that one before reviewing this one, this branch is dependent on that one and as such this diff view is polluted with the contents of that branch.~

### Changes
- Updated the dependency on habitica-markdown to 2.0.0 for both server and client.
- Removed the highlightUser method
- Updated the cards and tests to use markdown directly with passed in username and displayname

----


website/client/package-lock.json
```diff
@@ -11676,7 +11676,7 @@
       }
     },
     "habitica-markdown": {
-      "version": "1.4.0",
```
paglias, line 11679: I think you'll have to manually install `habitica-markdown@2` with `npm i habitica-markdown@2` or the lockfile won't be updated properly

website/client/package-lock.json
```diff
@@ -11676,7 +11676,7 @@
...
+      "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/habitica-markdown/-/habitica-markdown-1.4.0.tgz",
```
paglias: I think you'll have to manually install habitica-markdown@2 with npm i habitica-markdown@2 or the lockfile won't be updated properly

paglias (changes requested):

website/client/package-lock.json
benkelaar: Yeah I think it went wrong because the version wasn't available at the time. That situation has improved now! 🥳

paglias:
Tested it a bit and looks like nothing broke 🎉 I guess all the markdown stuff is done for now, thanks for all the PRs @benkelaar !

benkelaar:
Wow, genuinely feels like an accomplishment! Like I said in the other PR, thanks a lot for your support! The team is doing a great job helping out a new contributor like myself!

Now all we need to do is sit back and wait for the bug reports to come rolling in 😉. Or do one of the other issues... Or spoiler tags... Or ninja emoji... Or emoji suggestion in web... Or link detection in Android... Or markdown previews... Or WYSIWIG... Or... Always plenty to do for a blacksmith in habitica! 😁

PR #12090, opened 2020-04-13
Handle simultaneous quest accept/reject

### LINKED ISSUE #11398: simultaneous quest invitation accept/reject/force-start are not all recorded correctly
Miikethepriest (ee687a55-8b32-4e74-b935-2da33248a033) reported that they accepted a quest invitation but did not become a participant in the quest. Usually when this happens it's because the quest owner force-started the quest first, but in this case I've confirmed they really did accept the invitation in time and really were not recorded as a participant. 

From Amplitude data I found that another party member accepted the quest invitation about a quarter of a second after Miikethepriest. I'm assuming that the second acceptance prevented the first from being permanently stored in the database.

The party chat and Amplitude data showing this is below.

I'm marking this as minor priority since it's unlikely to happen often, and suggestion-discussion for a fix to be discussed.

```
quest owner sending invitations to the quest:
{
"client_event_time":"2019-09-26 21:04:03.645000"
"event_properties":{"owner":true,
"category":"behavior",
"questName":"dilatory_derby",
"response":"accept"},
}

Miikethepriest accepting the invitation:
{
"client_event_time":"2019-09-26 21:04:44.749000"
"event_properties":{"owner":false,
"category":"behavior",
"questName":"dilatory_derby",
"response":"accept"},
"user_id":"ee687a55-8b32-4e74-b935-2da33248a033",
}

another party member accepting the invitation in the same second:
{
"client_event_time":"2019-09-26 21:04:44.976000"
"event_properties":{"owner":false,
"category":"behavior",
"questName":"dilatory_derby",
"response":"accept"},
"user_id":"f836...",
}

quest owner force-starting the quest:
{
"client_event_time":"2019-09-27 10:25:59.690000"
"event_properties":{"owner":true,
"category":"behavior",
"questName":"dilatory_derby",
"response":"force-start"},
}

{
    "text": "`Your quest, The Dilatory Derby, has started.`",
    "info": {
        "type": "quest_start",
        "quest": "dilatory_derby"
    },
    "timestamp": "2019-09-27T10:25:59.283Z",
    "groupId": "3dc87ceb-35f0-4e69-b356-3f91a1078c4e",
    "_meta": {
        "participatingMembers": "[various party members including f836... but not including Miikethepriest]",
    },
}
```

cvuorinen:

Fixes #11398

### Changes
Handle quest accept/reject atomically so that only the single users true/false status is changed, so that two simultaneous accept/reject request will not override changes from each other. 

Also try to reduce the possibility of force-start conflicting with simultaneous accept/reject by saving the `quest.members` array first (so that incoming accept/reject requests will not do anything while processing the rest of `startQuest` method). 

----

### DIFF
--- a/website/server/controllers/api-v3/quests.js
+++ b/website/server/controllers/api-v3/quests.js
@@ -193,11 +193,13 @@ api.acceptQuest = {
     if (group.quest.active) throw new NotAuthorized(res.t('questAlreadyStartedFriendly'));
     if (group.quest.members[user._id]) throw new BadRequest(res.t('questAlreadyAccepted'));
 
+    const acceptedSuccessfully = await group.handleQuestInvitation(user, true);
+    if (!acceptedSuccessfully) {
+      throw new NotAuthorized(res.t('questAlreadyAccepted'));
+    }
+
     user.party.quest.RSVPNeeded = false;
     await user.save();
-
-    group.markModified('quest');
-    group.quest.members[user._id] = true;
 
     if (canStartQuestAutomatically(group)) {
       await group.startQuest(user);
@@ -252,12 +254,14 @@ api.rejectQuest = {
     if (group.quest.members[user._id]) throw new BadRequest(res.t('questAlreadyAccepted'));
     if (group.quest.members[user._id] === false) throw new BadRequest(res.t('questAlreadyRejected'));
 
+    const rejectedSuccessfully = await group.handleQuestInvitation(user, false);
+    if (!rejectedSuccessfully) {
+      throw new NotAuthorized(res.t('questAlreadyRejected'));
+    }
+
     user.party.quest = Group.cleanQuestUser(user.party.quest.progress);
     user.markModified('party.quest');
     await user.save();
-
-    group.quest.members[user._id] = false;
-    group.markModified('quest.members');
 
     if (canStartQuestAutomatically(group)) {
       await group.startQuest(user);
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -662,7 +662,7 @@ schema.methods.handleQuestInvitation = async function handleQuestInvitation (use
   ).exec();
 
   if (result.nModified) {
-    // update also current instance so future operations wil work correctly
+    // update also current instance so future operations will work correctly
     this.quest.members[user._id] = !!accept;
   }
 
@@ -698,6 +698,11 @@ schema.methods.startQuest = async function startQuest (user) {
 
   // Changes quest.members to only include participating members
   this.quest.members = _.pickBy(this.quest.members, _.identity);
+
+  // Persist quest.members early to avoid simultaneous handling of accept/reject
+  // while processing the rest of this script
+  this.update({ $set: { 'quest.members': this.quest.members } }).exec();
+
   const nonUserQuestMembers = _.keys(this.quest.members);
   removeFromArray(nonUserQuestMembers, user._id);
 
--- a/website/server/models/group.js
+++ b/website/server/models/group.js
@@ -649,6 +649,9 @@ schema.methods.sendChat = function sendChat (options = {}) {
 };
 
 schema.methods.handleQuestInvitation = async function handleQuestInvitation (user, accept) {
+  if (!user) throw new InternalServerError('Must provide user to handle quest invitation');
+  if (accept !== true && accept !== false) throw new InternalServerError('Must provide accept param handle quest invitation');
+
   // Handle quest invitation atomically (update only current member when still undecided)
   // to prevent multiple concurrent requests overriding updates
   // see https://github.com/HabitRPG/habitica/issues/11398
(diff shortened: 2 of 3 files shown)


website/server/controllers/api-v3/quests.js
```diff
@@ -193,12 +193,14 @@ api.acceptQuest = {
...
+    const acceptedSuccessfully = await group.handleQuestInvitation(user, true);
+    if (!acceptedSuccessfully) {
+      throw new NotAuthorized(res.t('questAlreadyAccepted'));
+    }
+
     user.party.quest.RSVPNeeded = false;
     await user.save();
 
-    group.markModified('quest');
-    group.quest.members[user._id] = true;
```
paglias, line 200: even though it's not used anymore, could we keep this line? And the similar one for `api.rejectQuest`? Because the quest data is returned from the API call and we should return the updated state

website/server/models/group.js
```diff
@@ -648,6 +648,27 @@ schema.methods.sendChat = function sendChat (options = {}) {
...
+    },
+    { $set: { [`quest.members.${user._id}`]: !!accept } },
+  ).exec();
+
+  if (result.nModified) {
+    // update also current instance so future operations will work correctly
+    this.quest.members[user._id] = !!accept;
+  }
+
+  return !!result.nModified;
```
paglias: can we use `Boolean(result.nModified)` instead of the two `!!`? it's a bit easier to read, for the two occurances above (`!!accept`), we're converting to a boolean but the function already requires a boolean and being used internally I don't think a forced conversion is necessary. 

What I'd do maybe is throw an error if the user or accept param isn't passed

website/server/models/group.js
```diff
@@ -648,6 +648,27 @@ schema.methods.sendChat = function sendChat (options = {}) {
...
 
+schema.methods.handleQuestInvitation = async function handleQuestInvitation (user, accept) {
+  // Handle quest invitation atomically (update only current member when still undecided)
+  // to prevent multiple concurrent requests overriding updates
+  // see https://github.com/HabitRPG/habitica/issues/11398
+  const Group = this.constructor;
+  const result = await Group.update(
+    {
+      _id: this._id,
+      [`quest.members.${user._id}`]: { $exists: true, $eq: null },
```
paglias: I think for this check you can use this query https://docs.mongodb.com/manual/tutorial/query-for-null-fields/#type-check

website/server/models/group.js
```diff
@@ -677,6 +698,11 @@ schema.methods.startQuest = async function startQuest (user) {
 
   // Changes quest.members to only include participating members
   this.quest.members = _.pickBy(this.quest.members, _.identity);
+
+  // Persist quest.members early to avoid simultaneous handling of accept/reject
+  // while processing the rest of this script
+  this.update({ $set: { 'quest.members': this.quest.members } }).exec();
```
paglias: I'm a bit confused about this update here, I mean which bug would it prevent

paglias (changes requested):
left a few comments but nothing major :) thanks

website/server/models/group.js
cvuorinen: Yeah, seems that should do the same thing. I can change it.

website/server/controllers/api-v3/quests.js
cvuorinen: It's already handled inside the `group.handleQuestInvitation` method.

website/server/models/group.js
cvuorinen: The issue states that another cause of bugs is simultaneous force-start call with an accept/reject call. Previously the `quest.members` list is updated here to only have members who accepted the invite, then it goes on to do a lot of work updating each of the users etc. so it could take a while, until finally `group.save()` is called in the controller which is when the members list gets persisted to db. In the mean time if someone else makes accept/reject call it would update the db, but then it would get overridden by the `group.save()` of the force-start call.

And there is also the potential of executing all this startQuest logic twice simultaneously (although it's pretty unlikely), if the force-start is called at the same time as the last member to accept/reject (which then goes on to execute startQuest also). This change should make that even less likely by updating the db early before all the user updates. Because then if another accept/reject call comes in while doing the user updates, the member would already be removed from the `quest.members` list and so the query will not match and they get an error.

website/server/controllers/api-v3/quests.js
paglias: I had missed that :)

website/server/models/group.js
paglias: thanks for the explanation, quite an edge case but let's keep it, the only thing is that we need to handle any possible error from the op, if we put an `await` in front and it fails it would stop the rest of the operations but I would still do it as the user can then retry. The other option would be to attach a `.catch` and just log the error while allowing the rest of the operations to continue

paglias:
thanks @cvuorinen ! noted in your profile

PR #12095, opened 2020-04-16
Allow changing name of party without adding description

### LINKED ISSUE #10828: Cannot change name of party without adding a description

anoopravella:

Fixes #10828

### Changes
Fixes the following issues with editing parties and guilds:
- Users can now click the "Update Party" button in the party edit menu without having to add a description
- Removed asterisk from "Description" label in party edit menu
- Removed asterisk from "Privacy settings" label in both party and guild edit menus

New edit party and edit guild menus:
Screen Shot 2020-04-15 at 11 20 45 PM
Screen Shot 2020-04-15 at 11 21 48 PM

----
UUID: c2079065-97bd-44a5-b45b-4fccf9455ce2

### DIFF
--- a/website/client/src/components/groups/groupFormModal.vue
+++ b/website/client/src/components/groups/groupFormModal.vue
@@ -115,8 +115,8 @@ label.custom-control-label(v-once) {{ $t('allowGuildInvitationsFromNonMembers') 
       </div>
       <div class="form-group">
         <label>
-          <strong v-if="isParty" v-once>{{ $t('groupDescription') }}</strong>
-          <strong v-if="!isParty" v-once>{{ $t('groupDescription') }} *</strong>
+          <strong v-if="isParty">{{ $t('groupDescription') }}</strong>
+          <strong v-else>{{ $t('groupDescription') }} *</strong>
         </label>
         <a
           v-markdown="$t('markdownFormattingHelp')"
@@ -255,8 +255,7 @@ label.custom-control-label(v-once) {{ $t('allowGuildInvitationsFromNonMembers') 
         <button
           v-if="workingGroup.id"
           class="btn btn-primary btn-md"
-          :disabled="(isParty && !workingGroup.name) ||
-            (!isParty && (!workingGroup.name || !workingGroup.description))"
+          :disabled="!workingGroup.name || (!isParty && !workingGroup.description)"
         >
           {{ isParty ? $t('updateParty') : $t('updateGuild') }}
         </button>
--- a/website/client/src/components/groups/groupFormModal.vue
+++ b/website/client/src/components/groups/groupFormModal.vue
@@ -18,7 +18,7 @@
       </div>
       <div class="form-group">
         <label>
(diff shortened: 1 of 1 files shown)


website/client/src/components/groups/groupFormModal.vue
```diff
@@ -115,7 +115,8 @@ label.custom-control-label(v-once) {{ $t('allowGuildInvitationsFromNonMembers')
       </div>
       <div class="form-group">
         <label>
-          <strong v-once>{{ $t('groupDescription') }} *</strong>
+          <strong v-if="isParty" v-once>{{ $t('groupDescription') }}</strong>
+          <strong v-if="!isParty" v-once>{{ $t('groupDescription') }} *</strong>
```
paglias: could we use a `v-else` for the second condition? We should also remove the `v-once`(s) since they prevent rendering based on a condition

website/client/src/components/groups/groupFormModal.vue
```diff
@@ -254,7 +255,8 @@ label.custom-control-label(v-once) {{ $t('allowGuildInvitationsFromNonMembers')
         <button
           v-if="workingGroup.id"
           class="btn btn-primary btn-md"
-          :disabled="!workingGroup.name || !workingGroup.description"
+          :disabled="(isParty && !workingGroup.name) ||
```
paglias: I think this can be simplified to `!workingGroup.name || (!isParty && !workingGroup.description)`

paglias (changes requested):

paglias:
left a couple of comments!

website/client/src/components/groups/groupFormModal.vue
anoopravella: Done!

website/client/src/components/groups/groupFormModal.vue
anoopravella: Done, thanks!

paglias:
@anoopravella  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #12115, opened 2020-04-22
Fix group Send Invites button

### LINKED ISSUE #12114: Group Invites: Send Invites button not updated correctly
Schermata 2020-04-20 alle 23 59 56

The modal to invite users to a guild or party has a few issues:

- The send invites button starts enabled while it should be disabled until there's at least one invite
- When typing an email or username the validity isn't checked until the focus change from the input to another element. If you input a valid email the Send Invites button will stay disabled until you click on another HTML element: it should instead be updated immediately after the user stops typing using a `debounced` (see [example](https://github.com/HabitRPG/habitica/blob/f8bab04a0e66a819f6d7ca5b210d27d991446bf3/website/client/src/mixins/challengeMemberSearch.js#L6)) function to make sure the server doesn't receive a request for every character typed

cvuorinen:

Fixes #12114

### Changes
Better handling of group Send Invites button: 

* Run validation check from `@input` event rather than `@change` so that it is executed after user stops typing instead of on blur.
* Fix `cannotSubmit` (bound to button :disabled) so that it correctly checks that at least one invite is filled and all filled invites are valid.
* Change debounce time for the validation check from 250ms to 500ms since it's more reasonable for user typing.
* Clear errors from the current invite once user starts typing (since the error condition is not valid anymore and needs to be checked again when they stop typing) 

----

### DIFF
--- a/website/client/src/components/groups/inviteModal.vue
+++ b/website/client/src/components/groups/inviteModal.vue
@@ -128,8 +128,6 @@
 <script>
 import clone from 'lodash/clone';
 import debounce from 'lodash/debounce';
-import filter from 'lodash/filter';
-import forEach from 'lodash/forEach';
 import isEmail from 'validator/lib/isEmail';
 import isUUID from 'validator/lib/isUUID';
 import { mapState } from '@/libs/store';
@@ -152,9 +150,8 @@ export default {
   computed: {
     ...mapState({ user: 'user.data' }),
     cannotSubmit () {
-      const filledInvites = filter(this.invites, invite => invite.text.length);
-      const validInvites = filter(filledInvites, invite => invite.valid);
-      return !filledInvites.length || validInvites.length !== filledInvites.length;
+      const filledInvites = this.invites.filter(invite => invite.text.length);
+      return !filledInvites.length || filledInvites.some(invite => !invite.valid);
     },
     inviter () {
       return this.user.profile.name;
@@ -162,12 +159,11 @@ export default {
   },
   methods: {
     checkInviteList: debounce(function checkList () {
-      this.invites = filter(
-        this.invites,
+      this.invites = this.invites.filter(
         (invite, index) => invite.text.length > 0 || index === this.invites.length - 1,
       );
       while (this.invites.length < 2) this.invites.push(clone(INVITE_DEFAULTS));
-      forEach(this.invites, (value, index) => {
+      this.invites.forEach((value, index) => {
         if (value.text.length < 1 || isEmail(value.text)) {
           return this.fillErrors(index);
         }
@@ -217,7 +213,7 @@ export default {
         uuids: [],
         usernames: [],
       };
-      forEach(this.invites, invite => {
+      this.invites.forEach(invite => {
         if (invite.text.length < 1) return;
         if (isEmail(invite.text)) {
           invitationDetails.emails.push({ email: invite.text });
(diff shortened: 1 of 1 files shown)


website/client/src/components/groups/inviteModal.vue
```diff
@@ -29,7 +29,7 @@
             :class="{
               'input-valid': invite.valid, 'is-invalid input-invalid': invite.valid === false}"
             @keyup="expandInviteList"
-            @change="checkInviteList"
+            @input="checkInviteList(); clearErrors(invite);"
```
benkelaar: I think it would be better if `clearErrors` was called from within `checkInviteList`. That matches better with how `fillErrors` is also called from there and keeps the javascript and html nicely separate. As an alternative you could define a third function (perhaps `processInput`) that calls these two functions but I don't think that's really necessary.

website/client/src/components/groups/inviteModal.vue
```diff
@@ -152,12 +152,9 @@ export default {
...
     cannotSubmit () {
-      const filteredInvites = filter(
-        this.invites,
-        invite => invite.text.length > 0 && !invite.valid,
-      );
-      if (filteredInvites.length > 0) return true;
-      return false;
+      const filledInvites = filter(this.invites, invite => invite.text.length);
+      const validInvites = filter(filledInvites, invite => invite.valid);
+      return !filledInvites.length || validInvites.length !== filledInvites.length;
```
benkelaar:
```suggestion
      const filledInvites = this.invites.filter(invite => invite.text.length);
      return !filledInvites.length || filledInvites.some(invite => !invite.valid);
```

I think for arrays it's better to prefer `Array.prototype.filter` instead of lodash's `filter` method, since it's closer to ES native and (IMO reads nicer). I also think using the `Array.prototype.some` method reads easier than the length check.

benkelaar: Alternatively you could do `!filledInvites.length || !filledInvites.every(invite => invite.valid)`, but I think having the negation inside the arrow function reads easier.

benkelaar (changes requested):
Couple of suggestions for changes.

website/client/src/components/groups/inviteModal.vue
paglias: Agreed, where possible we should use ES native methods (these are all available on all modern browsers but even if they were not babel would take care of them and inject a polyfill)

website/client/src/components/groups/inviteModal.vue
paglias: 👍

cvuorinen: Yeah, would have done it that way myself, but the problem is that we only want to clear the error for this particular invite and not all of them. Inside `checkInviteList` method there is no handle to a particular invitation, it operates on the whole list. Of course could change the method signature so that we pass in the invitation, but I think that defeats the purpose since it's not actually required for the main functionality of that method.

website/client/src/components/groups/inviteModal.vue
cvuorinen: Yeah, I prefer using native array methods myself too, just didn't know what was the reason lodash methods were used there before so just kept the same style. But I can change them to native.

website/client/src/components/groups/inviteModal.vue
cvuorinen: Ok, I made a new method `inviteUpdated(invite)` that acts as the event handler for `@input` and it will then call `clearErrors` and `checkInviteList`. That way the html reads nicer and also SRP :+1:

paglias:
thanks!

PR #12116, opened 2020-04-22
Don't bill (subtract gems) multiple times for multiple unlock item set calls

### LINKED ISSUE #11050: Gem purchases of avatar customisation items should not be possible when you own them
Black Rabbit / @niela (a49e76ca-5f53-44dc-9c20-36abfbec375a) says:

"I just lost 5 gems 'cause the habitica sync didn't show me that I already brought my animals' ears pack."

Confirmed in Amplitude (links will work for only admins):

[event 1](https://analytics.amplitude.com/habitica/project/132025/search/a49e76ca-5f53-44dc-9c20-36abfbec375a?eventFilter=purchase&eventFilter=acquire%20item&sessionHandle=WlV9FDr_WlV9FSa_D%252Bjsav_gO5_070511e6-a0e0-49d5-936b-e34f81197752&eventId=0b49717c-4083-11e9-a76b-066ed5d8f214)

```
March 7, 2019 2:45:16 am UTC
acquire item

acquireMethod: Gems
gemCost: 5
itemKey: items.gear.owned.headAccessory_special_bearEars,items.gear.owned.headAccessory_special_cactusEars,items.gear.owned.headAccessory_special_foxEars,items.gear.owned.headAccessory_special_lionEars,items.gear.owned.headAccessory_special_pandaEars,items.gear.owned.headAccessory_special_pigEars,items.gear.owned.headAccessory_special_tigerEars,items.gear.owned.headAccessory_special_wolfEars
```

[event 2](https://analytics.amplitude.com/habitica/project/132025/search/a49e76ca-5f53-44dc-9c20-36abfbec375a?eventFilter=purchase&eventFilter=acquire%20item&sessionHandle=WlV9FDr_WlV9FSa_D%252Bjsav_gO5_070511e6-a0e0-49d5-936b-e34f81197752&eventId=15b32914-4083-11e9-a76b-066ed5d8f214)
```
March 7, 2019 
2:45:33 am UTC
acquireMethod: Gems
gemCost: 5
itemKey: items.gear.owned.headAccessory_special_bearEars,items.gear.owned.headAccessory_special_cactusEars,items.gear.owned.headAccessory_special_foxEars,items.gear.owned.headAccessory_special_lionEars,items.gear.owned.headAccessory_special_pandaEars,items.gear.owned.headAccessory_special_pigEars,items.gear.owned.headAccessory_special_tigerEars,items.gear.owned.headAccessory_special_wolfEars
```

`balanceGemAmount` on events before and after those shows that 10 Gems were indeed removed.

I've refunded the extra 5 Gems.

I'm leaving this as suggestion-discussion for the staff to comment on an appropriate way to fix it. The avatar customisation purchases for items like ears, backgrounds, etc, need changes for other reasons too (as discussed elsewhere in the past) so the fix for this issue might be best rolled into that.

benkelaar:
Fixes #11050 

### Changes
- Added `Mixed` items field declaration to user schema
- Added some tests for unlock.js
- Cleaned up some code in unlock.js
- Added a debug run target to allow attaching a debugger to the server

### Possible additional work I can pick up
- Create a migration to fix it for existing purchases of existing users
- Investigate why the tests didn't catch this
- Fix the TODOs (test coverage and partial unlocks)
- Refactor significantly (All the if checks are a clear smell that the abstraction here is totally wrong)

Let me know what you'd like to see included in this merge request

----

### DIFF
--- a/test/api/v3/integration/user/POST-user_unlock.js
+++ b/test/api/v3/integration/user/POST-user_unlock.js
@@ -5,7 +5,7 @@ import {
 
 describe('POST /user/unlock', () => {
   let user;
-  const unlockPath = 'shirt.convict,shirt.cross,shirt.fire,shirt.horizon,shirt.ocean,shirt.purple,shirt.rainbow,shirt.redblue,shirt.thunder,shirt.tropical,shirt.zombie';
+  const unlockPath = 'items.gear.owned.headAccessory_special_bearEars,items.gear.owned.headAccessory_special_cactusEars,items.gear.owned.headAccessory_special_foxEars,items.gear.owned.headAccessory_special_lionEars,items.gear.owned.headAccessory_special_pandaEars,items.gear.owned.headAccessory_special_pigEars,items.gear.owned.headAccessory_special_tigerEars,items.gear.owned.headAccessory_special_wolfEars';
   const unlockCost = 1.25;
   const usersStartingGems = 5;
 
@@ -32,6 +32,27 @@ describe('POST /user/unlock', () => {
     await user.sync();
 
     expect(response.message).to.equal(t('unlocked'));
+    expect(user.balance).to.equal(usersStartingGems - unlockCost);
+  });
+
+  it('does not reduce a user\'s balance twice', async () => {
+    await user.update({
+      balance: usersStartingGems,
+    });
+    let response = await user.post(`/user/unlock?path=${unlockPath}`);
+    await user.sync();
+
+    expect(response.message).to.equal(t('unlocked'));
+    expect(user.balance).to.equal(usersStartingGems - unlockCost);
+
+    expect(user.post(`/user/unlock?path=${unlockPath}`))
+      .to.eventually.be.rejected.and.to.eql({
+        code: 401,
+        error: 'NotAuthorized',
+        message: t('alreadyUnlocked'),
+      });
+    await user.sync();
+
     expect(user.balance).to.equal(usersStartingGems - unlockCost);
   });
 });
--- a/test/api/v3/integration/user/POST-user_unlock.js
+++ b/test/api/v3/integration/user/POST-user_unlock.js
@@ -39,7 +39,7 @@ describe('POST /user/unlock', () => {
     await user.update({
       balance: usersStartingGems,
     });
-    let response = await user.post(`/user/unlock?path=${unlockPath}`);
+    const response = await user.post(`/user/unlock?path=${unlockPath}`);
     await user.sync();
 
     expect(response.message).to.equal(t('unlocked'));
--- a/test/api/v3/integration/user/POST-user_unlock.js
+++ b/test/api/v3/integration/user/POST-user_unlock.js
@@ -5,7 +5,8 @@ import {
 
 describe('POST /user/unlock', () => {
   let user;
-  const unlockPath = 'items.gear.owned.headAccessory_special_bearEars,items.gear.owned.headAccessory_special_cactusEars,items.gear.owned.headAccessory_special_foxEars,items.gear.owned.headAccessory_special_lionEars,items.gear.owned.headAccessory_special_pandaEars,items.gear.owned.headAccessory_special_pigEars,items.gear.owned.headAccessory_special_tigerEars,items.gear.owned.headAccessory_special_wolfEars';
+  const unlockPath = 'shirt.convict,shirt.cross,shirt.fire,shirt.horizon,shirt.ocean,shirt.purple,shirt.rainbow,shirt.redblue,shirt.thunder,shirt.tropical,shirt.zombie';
+  const unlockGearSetPath = 'items.gear.owned.headAccessory_special_bearEars,items.gear.owned.headAccessory_special_cactusEars,items.gear.owned.headAccessory_special_foxEars,items.gear.owned.headAccessory_special_lionEars,items.gear.owned.headAccessory_special_pandaEars,items.gear.owned.headAccessory_special_pigEars,items.gear.owned.headAccessory_special_tigerEars,items.gear.owned.headAccessory_special_wolfEars';
   const unlockCost = 1.25;
   const usersStartingGems = 5;
 
@@ -39,13 +40,13 @@ describe('POST /user/unlock', () => {
     await user.update({
       balance: usersStartingGems,
     });
-    const response = await user.post(`/user/unlock?path=${unlockPath}`);
+    const response = await user.post(`/user/unlock?path=${unlockGearSetPath}`);
     await user.sync();
 
     expect(response.message).to.equal(t('unlocked'));
     expect(user.balance).to.equal(usersStartingGems - unlockCost);
 
-    expect(user.post(`/user/unlock?path=${unlockPath}`))
+    expect(user.post(`/user/unlock?path=${unlockGearSetPath}`))
       .to.eventually.be.rejected.and.to.eql({
         code: 401,
         error: 'NotAuthorized',
--- a/test/common/ops/unlock.js
+++ b/test/common/ops/unlock.js
@@ -8,7 +8,7 @@ import {
   BadRequest,
 } from '../../../website/common/script/libs/errors';
 
-describe.only('shared.ops.unlock', () => {
+describe('shared.ops.unlock', () => {
   let user;
   const unlockPath = 'shirt.convict,shirt.cross,shirt.fire,shirt.horizon,shirt.ocean,shirt.purple,shirt.rainbow,shirt.redblue,shirt.thunder,shirt.tropical,shirt.zombie';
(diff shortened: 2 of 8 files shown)


website/server/models/user/schema.js
```diff
@@ -187,6 +187,10 @@ export default new Schema({
       $type: Schema.Types.Mixed,
       default: () => ({}),
     },
+    items: {
```
paglias: I think this can be removed now

website/common/script/ops/unlock.js
```diff
@@ -13,6 +10,46 @@ import { removeItemByPath } from './pinnedGearUtils';
...
+  if (user.markModified) user.markModified(path);
+}
+
+function purchaseItem (path, user) {
+  if (path.includes('gear.')) {
+    setAsObject(user, path, true);
+    const itemName = splitPathItem(path)[1];
+    removeItemByPath(user, `gear.flat.${itemName}`);
+    if (path.includes('gear.owned')) markModified(user, 'items.gear.owned');
+  }
```
paglias, line 58: should there be an `else` block here since gears should not be recorded in `user.purchased`?

test/api/v3/integration/user/POST-user_unlock.js
```diff
@@ -5,7 +5,7 @@ import {
 
 describe('POST /user/unlock', () => {
   let user;
-  const unlockPath = 'shirt.convict,shirt.cross,shirt.fire,shirt.horizon,shirt.ocean,shirt.purple,shirt.rainbow,shirt.redblue,shirt.thunder,shirt.tropical,shirt.zombie';
+  const unlockPath = 'items.gear.owned.headAccessory_special_bearEars,items.gear.owned.headAccessory_special_cactusEars,items.gear.owned.headAccessory_special_foxEars,items.gear.owned.headAccessory_special_lionEars,items.gear.owned.headAccessory_special_pandaEars,items.gear.owned.headAccessory_special_pigEars,items.gear.owned.headAccessory_special_tigerEars,items.gear.owned.headAccessory_special_wolfEars';
```
paglias: can we also test for the old unlock path so that we test both gear and other items

test/common/ops/unlock.js
```diff
@@ -45,23 +45,23 @@ describe('shared.ops.unlock', () => {
 
   it('returns an error when user already owns a full set', done => {
     try {
-      unlock(user, { query: { path: unlockPath } });
-      unlock(user, { query: { path: unlockPath } });
+      unlock(user, { query: { path: unlockGearSetPath } });
```
paglias, line 64: here as well can we test both `unlockGearSetPath `and `unlockPath `?

website/common/script/ops/unlock.js
```diff
@@ -36,27 +73,17 @@ export default function unlock (user, req = {}, analytics) {
...
-    if (alreadyOwnedItems === setPaths.length) {
+    const alreadyOwnedItems = setPaths.filter(p => get(user, `purchased.${p}`)).length;
+    const totalItems = setPaths.length;
+    if (alreadyOwnedItems === totalItems) {
       throw new NotAuthorized(i18n.t('alreadyUnlocked', req.language));
-    // TODO write math formula to check if buying
-    // the full set is cheaper than the items individually
-    // (item cost * number of remaining items) < setCost`
-    } /* else if (alreadyOwnedItems > 0) {
+    } else if ((totalItems - alreadyOwnedItems) < 3) {
```
paglias: For this one I'd prefer to handle it in the next PR so that we can have a precise check on the number of items in the set and their cost since we want to be flexible about how many items there are in a set. In the PR I'll make right after this one I'll be sure to check all the sets against the content api for their price and the items in them

paglias (changes requested):
Left a few comments, thanks for cleaning up the code a bit

website/server/models/user/schema.js
benkelaar: Indeed, didn't actually get round to implementing the improvements you suggested over chat. Thanks for noting them down here as well. Will look at them tonight.

website/common/script/ops/unlock.js
benkelaar: Yeah, good point, fixed.

website/server/models/user/schema.js
benkelaar: Done

website/common/script/ops/unlock.js
benkelaar: But the price for all sets is actually set in the code to be 15 gems for background sets and 5 gems for any other set. That combined with the fact that backgrounds cost 7 gems per single and other items cost 2 gems per single means that currently, for all uses of the unlock functionality you need to buy less than 3 new items at once in order for single purchases to be cheaper.

However, I'll re-comment it out and make a follow-up pull request. No need to burden this PR even more 😄.

website/server/models/user/schema.js
paglias: I think there might have been an issue with the push of the last few commits

benkelaar: Changes should now be in 👍

test/api/v3/integration/user/POST-user_unlock.js
benkelaar: Done.

test/common/ops/unlock.js
benkelaar: Done.

website/common/script/ops/unlock.js
```diff
@@ -22,86 +73,45 @@ export default function unlock (user, req = {}, analytics) {
...
-    cost = 3.75;
-  } else if (isBackground) {
-    cost = 1.75;
-  } else if (isFullSet) {
-    cost = 1.25;
-  } else {
-    cost = 0.5;
-  }
+  const isFullSet = path.includes(',');
+  const isBackground = path.includes('background.');
```
benkelaar: Should this be a startsWith instead?

benkelaar: Did it, think it's more descriptive.

website/common/script/ops/unlock.js
```diff
@@ -1,18 +1,69 @@
...
+  }
+  return isFullSet ? 1.25 : 0.5;
+}
+
+function isGear (path) {
+  return path.includes('gear.');
+}
+
+function alreadyOwns (user, path) {
+  return get(user, isGear(path) ? path : `purchased.${path}`);
```
paglias, line 24: I think this is correct for non gear items but for gear, you can actually lose an unlocked gear item in which case `user.items.gear.owned[key]` will be `false`  but you still own it (can buy it back with gold and not gems) so I would check for the key missing (user has never owned the gear):

```
return isGear(path) 
    ? get(user, path) !== undefined;
    : get(user, `purchased.${path}`) === true; // so that it's converted to a boolean
```

paglias: I think this is correct for non gear items but for gear, you can actually lose an unlocked gear item in which case `user.items.gear.owned[key]` will be `false`  but you still own it (can buy it back with gold and not gems) so I would check for the key missing (user has never owned the gear):

```
return isGear(path) 
    ? get(user, path) !== undefined;
    : get(user, `purchased.${path}`) === true; // so that it's converted to a boolean
```

paglias (changes requested):
left one comment but apart from that it's good to go

website/common/script/ops/unlock.js
benkelaar: Ah yeah, good point. However the conversion to boolean is not necessary right? The field is boolean in the database, so what you'd essentially be doing is `true === true`.

benkelaar: BTW how then do you buy back the lost gear item? Is that somewhere in the buy endpoints?

benkelaar: And won't the singles flow break down? Because I think it'll try to set the lost item as preferred. Will look in to working around that.

Edit: Guess I'll have to introduce a difference between `alreadyOwns` and `alreadyUnlocked`

paglias: > Ah yeah, good point. However the conversion to boolean is not necessary right? The field is boolean in the database, so what you'd essentially be doing is true === true.

yes but being a Mixed field when it does not exist it won't be false but simply undefined, so if you don't have the item `get(user, `purchased.${path}`)` will be `undefined` not `false`

> BTW how then do you buy back the lost gear item? Is that somewhere in the buy endpoints?

> And won't the singles flow break down? Because I think it'll try to set the lost item as preferred. Will look in to working around that.
> Edit: Guess I'll have to introduce a difference between alreadyOwns and alreadyUnlocked


let me check

benkelaar: Ah, of course, the preferences field doesn't have any item field in the schema, that's why it doesn't matter. Will leave it as is then and only fix the bug.

paglias:
Thanks @benkelaar ! Welcome to tier 2

benkelaar:
Awesome! Thanks!

PR #12121, opened 2020-04-24
API: Adding secret.text to the user-schema

### LINKED ISSUE #12088: create user.secret.text field for staff/moderator use - server/API changes only
_I've marked this as suggestion-discussion for the staff to comment on the section in bold (and on the rest of it if they wish)._

This change is purely for server-side code (the API and user model). No website/vue changes are needed at this stage.

We need a new text field in the user document that staff and moderators can use to record private information about a player (e.g., the fact that a child has been given parental permission to use Habitica). The field must be invisible to all users who are not staff or mods.

The work required for this issue is:
1. adjust the user model to add the field
2. adjust the API's [getHero](https://github.com/HabitRPG/habitica/blob/57a9fb524132ac2f6479d4e11dd05578c1422a03/website/server/controllers/api-v3/hall.js#L151) route so that the new field is returned with the other user data
3. adjust the API's [updateHero](https://github.com/HabitRPG/habitica/blob/57a9fb524132ac2f6479d4e11dd05578c1422a03/website/server/controllers/api-v3/hall.js#L205) route so that the new field can be modified
4. adjust any other API routes as needed so that the new field cannot be seen or edited by users who aren't staff or mods
5. add tests for the new behaviour

For points 2 and 3:
- It's possible that that behaviour might already happen by default so test it before adding any code.
- Those routes are already locked down to allow only staff and mods to use them (`ensureAdmin` in [this code](https://github.com/HabitRPG/habitica/blob/57a9fb524132ac2f6479d4e11dd05578c1422a03/website/server/controllers/api-v3/hall.js#L169)), so no extra code is needed for that.

For point 4, you might be able to achieve that by adjusting something in one place in the code base, rather than editing every single route. I am being vague here because I haven't looked into it but post to this issue if you can't find a way to do point 4 relatively easily. It's also possible that the desired behaviour will happen by default since the `secret` object is new.

I suggest writing the tests first because they should make it clear where new code is needed for points 2, 3, and 4.

The new field will be added to the user model, which you can see here:
https://github.com/HabitRPG/habitica/blob/develop/website/server/models/user/schema.js

**I propose adding the new field as shown below, at the top level of the user document:**
```
  secret: {
    text: String,
  },
```
**I.e., the new field would be `user.secret.text`. This keeps it out of any other objects that are visible to the players, and allows us to expand it later with other data if needed. However Habitica's staff need to approve this before implementation.**

Characteristics for the field:
- [x] name is `secret.text` as shown above
- [x] type is String
- [x] there should be a comment in schema.js above the top-level `secret` field saying something like "visible to only staff and moderators"
- [x] the data stored in `secret.text` must be visible to only staff and moderators, not to the user themself, or to any other users
- [ ] the data can be added, edited, deleted by only staff and mods, not by the user themself, or by any other users
- [x] you can identify staff and mods by the `contributor.admin` field - any user that has `true` in that field is a staff/mod and must be allowed to see and edit the `secret` field
- [x] a player who is a staff/mod **can** see and edit their own field - i.e., you don't need to write any code to prevent staff/mods from using the field in their own user document

Tests needed:
- Viewing the data:
   - [x] any user with `contributor.admin` set to true can see the field when using the [/api/v3/hall/heroes/:heroId GET route](https://habitica.com/apidoc/#api-Hall-GetHero)
   - [x] a non-admin user cannot see the field for another user's account when using [/api/v3/members/:memberId](https://habitica.com/apidoc/#api-Member-GetMember)
   - [x] a non-admin user cannot see the field for their own account when using [/api/v3/members/:memberId](https://habitica.com/apidoc/#api-Member-GetMember)
   - [x] a non-admin user cannot see the field for their own account when using the [/api/v3/user GET route](https://habitica.com/apidoc/#api-User-UserGet)
   - [x] a non-admin user cannot see the field for their own account when using [/api/v3/user/anonymized](https://habitica.com/apidoc/#api-User-UserGetAnonymized)
- Updating the data:
   - [ ] any user with `contributor.admin` set to true can add/change text in the field when using the [/api/v3/hall/heroes/:heroId PUT route](https://habitica.com/apidoc/#api-Hall-UpdateHero)
   - [ ] a non-admin user cannot add/change text to the field for their own account when using the [/api/v3/user PUT route](https://habitica.com/apidoc/#api-User-UserUpdate)
- Possibly other tests depending on implementation. Don't rely on me having thought of everything that needs testing. :)

NB: as mentioned above, no website/vue changes are needed for this issue - i.e., you don't have to create any user interface (UI) elements for displaying or changing the field. They can be added later. Currently the admin interface that staff and mods use to display and edit users is being redeveloped so there's no point adding the new field to the existing UI.

negue:
closes #12088

Currently can be only set and read by using `/api/v3/hall/heroes/:heroId`, if other routes need be allowed to set / read it, please add a comment.

> Copied from Alys :)

Characteristics for the field:
- [x] name is `secret.text` as shown above
- [x] type is String
- [x] there should be a comment in schema.js above the top-level `secret` field saying something like "visible to only staff and moderators"
- [x] the data stored in `secret.text` must be visible to only staff and moderators, not to the user themself, or to any other users
- [x] the data can be added, edited, deleted by only staff and mods, not by the user themself, or by any other users
- [x] you can identify staff and mods by the `contributor.admin` field - any user that has `true` in that field is a staff/mod and must be allowed to see and edit the `secret` field
- [x] a player who is a staff/mod **can** see and edit their own field - i.e., you don't need to write any code to prevent staff/mods from using the field in their own user document  - using `/hall/heroes/`

Tests needed:
- Viewing the data:
   - [x] any user with `contributor.admin` set to true can see the field when using the [/api/v3/hall/heroes/:heroId GET route](https://habitica.com/apidoc/#api-Hall-GetHero)
   > note: `/api/v3/hall/heroes/:heroId` can be only called as admin
   - [x] a non-admin user cannot see the field for another user's account when using [/api/v3/members/:memberId](https://habitica.com/apidoc/#api-Member-GetMember)
   - [x] a non-admin user cannot see the field for their own account when using [/api/v3/members/:memberId](https://habitica.com/apidoc/#api-Member-GetMember)
   - [x] a non-admin user cannot see the field for their own account when using the [/api/v3/user GET route](https://habitica.com/apidoc/#api-User-UserGet)
   - [x] a non-admin user cannot see the field for their own account when using [/api/v3/user/anonymized](https://habitica.com/apidoc/#api-User-UserGetAnonymized)
- Updating the data:
   - [x] any user with `contributor.admin` set to true can add/change text in the field when using the [/api/v3/hall/heroes/:heroId PUT route](https://habitica.com/apidoc/#api-Hall-UpdateHero)
   - [x] a non-admin user cannot add/change text to the field for their own account when using the [/api/v3/user PUT route](https://habitica.com/apidoc/#api-User-UserUpdate)

Additional tests:
- [x] does not reset the secret, when a user calls `/user/reset`
- [x] can not query `secret` while using `?userFields`-query

I hope I catched all routes

### DIFF
--- a/test/api/v3/integration/hall/PUT-hall_heores_heroId.test.js
+++ b/test/api/v3/integration/hall/PUT-hall_heores_heroId.test.js
@@ -4,7 +4,7 @@ import {
   translate as t,
 } from '../../../../helpers/api-integration/v3';
 
-describe('PUT /heroes/:heroId', () => {
+describe.only('PUT /heroes/:heroId', () => {
   let user;
 
   before(async () => {
@@ -172,6 +172,45 @@ describe('PUT /heroes/:heroId', () => {
     expect(hero.contributor.level).to.equal(5); // doesn't modify previous values
     expect(hero.contributor.text).to.equal('Astronaut');
   });
+
+
+  it.only('updates contributor secret', async () => {
+    const secretText = 'my super hero';
+
+    const hero = await generateUser({
+      contributor: { level: 5 },
+      secret: {
+        text: 'supr hro typo',
+      },
+    });
+    const heroRes = await user.put(`/hall/heroes/${hero._id}`, {
+      contributor: { text: 'Astronaut' },
+      secret: {
+        text: secretText,
+      },
+    });
+
+    // test response
+    expect(heroRes).to.have.all.keys([ // works as: object has all and only these keys
+      '_id', 'balance', 'profile', 'purchased',
+      'contributor', 'auth', 'items', 'flags',
+      'secret',
+    ]);
+    expect(heroRes.auth.local).not.to.have.keys(['salt', 'hashed_password']);
+    expect(heroRes.profile).to.have.all.keys(['name']);
+
+    // test response values
+    expect(heroRes.contributor.level).to.equal(5); // doesn't modify previous values
+    expect(heroRes.contributor.text).to.equal('Astronaut');
+    expect(heroRes.secret.text).to.equal(secretText);
+
+    // test hero values
+    await hero.sync();
+    expect(hero.contributor.level).to.equal(5); // doesn't modify previous values
+    expect(hero.contributor.text).to.equal('Astronaut');
+    expect(hero.secret.text).to.equal(secretText);
+  });
+
 
   it('updates items', async () => {
     const hero = await generateUser();
--- a/test/api/v3/integration/hall/PUT-hall_heores_heroId.test.js
+++ b/test/api/v3/integration/hall/PUT-hall_heores_heroId.test.js
@@ -4,7 +4,7 @@ import {
   translate as t,
 } from '../../../../helpers/api-integration/v3';
 
-describe.only('PUT /heroes/:heroId', () => {
+describe('PUT /heroes/:heroId', () => {
   let user;
 
   before(async () => {
@@ -173,8 +173,7 @@ describe.only('PUT /heroes/:heroId', () => {
     expect(hero.contributor.text).to.equal('Astronaut');
   });
 
-
-  it.only('updates contributor secret', async () => {
+  it('updates contributor secret', async () => {
     const secretText = 'my super hero';
 
     const hero = await generateUser({
@@ -210,7 +209,6 @@ describe.only('PUT /heroes/:heroId', () => {
     expect(hero.contributor.text).to.equal('Astronaut');
     expect(hero.secret.text).to.equal(secretText);
   });
-
 
   it('updates items', async () => {
     const hero = await generateUser();
--- a/test/api/v3/integration/hall/PUT-hall_heores_heroId.test.js
+++ b/test/api/v3/integration/hall/PUT-hall_heores_heroId.test.js
@@ -7,6 +7,12 @@ import {
 describe('PUT /heroes/:heroId', () => {
   let user;
 
+  const heroFields = [
+    '_id', 'balance', 'profile', 'purchased',
+    'contributor', 'auth', 'items', 'flags',
+    'secret',
+  ];
+
   before(async () => {
     user = await generateUser({
       contributor: { admin: true },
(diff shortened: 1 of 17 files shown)


website/server/libs/user/index.js
```diff
@@ -19,6 +20,12 @@ export async function get (req, res, { isV3 = false }) {
     userToJSON = user.toJSON();
   }
 
+  if (!user.isAdmin()) {
```
paglias: hmm, won't the user returned from routes different than GET /user include the secret? Maybe we should add it to `private` [here](https://github.com/HabitRPG/habitica/blob/develop/website/server/models/user/hooks.js#L24) and only return it for the hero routes?

paglias:
Left one comment, all the rest LGTM!

website/server/libs/user/index.js
negue: I just tried using the `private` array but with that I wouldn't be able to read the User-field `secret` for `/hall/heroes/:heroId` or did I miss a flag for that?

paglias: Sorry, yeah if you add it to the private field it won't show up by default but at least it'll apply to all methods where the user is returned. I'd add it to the array and maybe add a new method to return it that you can use where necessary? So the hall routes probably

negue: I'd need a hint on how / what method I could create to read the `secret` , is the `private` just filtered when it is serialized to json?

If so I'll try to add the method and use it to get the secret then

paglias:
Yes, the private field are removed only from the serialized json, so when
manually calling .toJSON or sending the document as an API response but are
otherwise accessible

Il Lun 27 Apr 2020, 21:52 negue <notifications@github.com> ha scritto:

test/api/v3/integration/user/POST-user_reset.test.js
```diff
@@ -117,4 +117,24 @@ describe('POST /user/reset', () => {
     expect(userChallengeTask).to.exist;
     expect(syncedGroupTask).to.exist;
   });
+
+  it('does not delete secret', async () => {
```
Alys, line 121: This is a good test to have - thanks for thinking of it!

test/api/v3/integration/hall/PUT-hall_heores_heroId.test.js
```diff
@@ -7,6 +7,12 @@ import {
 describe('PUT /heroes/:heroId', () => {
   let user;
 
+  const heroFields = [
+    '_id', 'balance', 'profile', 'purchased',
+    'contributor', 'auth', 'items', 'flags',
+    'secret',
```
Alys, line 13: I like how you've pulled this array out of all the individual tests. Much neater.

Alys (approved):
@negue This is great, thanks!

I love the sample texts that you've used in your tests. :laughing:

Issue #10786, opened 2018-10-27
"Invited a Friend" achievement is not awarded; party invitation is not sent (Social Signups only)


Alys:
_This bug has been happening for ages (maybe since the new website) but I've just realised we don't have an issue for it, unless I can't find it._ `shrug`

When you invite someone to Habitica and they register an account from the link in the email they received, you are supposed to get the `invitedFriend` achievement. In some or all cases (I'm not sure which), the achievement is not given.

Also they are supposed to receive an invitation to your party but they don't.

SabreCat:
I think this is happening because the link that people get on invite just goes to the front page. It doesn't hook into anything.

Alys:
There's some additional details found by cTheDragons (on the delta test site but I expect it's the same for the production site):

"When inviting a new player to Habitica via their email. When the account signs up with that email account, the player does not automatically join the party. If you invite them again with their email account it does appear in the accept/reject which you can then accept."

Alys:
We had a report of this today so I'm copying it here as evidence that this bug still happens. We actually see a lot of reports about it but this one more clearly shows both impacts of the bug. It's from @schmyddy (9a3f4c14-3527-48c9-8d49-231cc353b7ed):

"Hey there. I face 2 problems right now: [I use the german version, so forgive me if i translate something wrong]
1.  I have invited someone to my party to get the Inviting-a-friend achievement through entering her e-mail and she clicked the link and registered but doesn't have any notification of an invitation. 
2. I tried to invite her now through her user name, but when i click "send invitation" it says, that this user doesn't exist... But she send me a picture of her profile and she already used both the app and the browser version of Habitica, so she should definitely exist.
Thank you for your effort in advance and thank you for taking care of the community. :)"

SabreCat:
Upgrading to Important as it's a constant drain for user support.

nikonhub:
I tried to invite an email and I think the problem should come from mandillapp configuration.

Because this link :
[...]

PR #12141, opened 2020-04-27
Refactor/auth component

### LINKED ISSUE #11482: Client Auth: refactor components

### LINKED ISSUE #12419: error not shown even after invalid email
To Create problem:
visit this https://habitica.com/static/home
fill the form and in mail section write @g and move to password column you will se there is no error will apear and then fill all the option .
Screenshot (429)

As this should so error but not showing .
now change email to @g.g you will find a error appears .
so I want to fix that and make changes so that for @g also show error like this one 
Screenshot (428)

 
LET ME KNOW IF THIS WILL HELP YOUR PROJECT I WANT TO CONTRIBUTE

ShupingHe:
this PR fixes #11482 
fixes #12419

abstract the auth logic into login.vue、register.vue and also abstract the social auth component

UUID:


ShupingHe:
@paglias there are still some works need to do, but you can start the review, so you can point out the mistakes i have made

website/client/src/components/auth/socialAuth/google.vue
```diff
@@ -0,0 +1,135 @@
...
+  },
+  mounted () {
+    hello.init({
+      facebook: process.env.FACEBOOK_KEY, // eslint-disable-line
+      // windows: WINDOWS_CLIENT_ID,
+      google: process.env.GOOGLE_CLIENT_ID, // eslint-disable-line
+    });
+  },
+  methods: {
+    async socialAuth (network) {
```
paglias: I think this part of the social auth process is shared between facebook and google entirely, partially for apple, could it be abstracted for example using a mixin?

website/client/src/components/auth/register.vue
```diff
@@ -0,0 +1,342 @@
+<template>
+  <div
```
paglias: Can this and other forms use a `form` element, we should have forms being `form` elements to maintain maximum accessibility and make it possible to submit by pressing Enter on any input inside the form itself

paglias:
Thanks @Stevens1995, this looks like a lot of work! I left a couple of comments but I haven't yet reviewed it entirely.

From local testing it looks like there might be some issues with the placeholders on the homepage register form and the username validation not working:

Schermata 2020-04-28 alle 13 20 05

^ the home page doesn't have labels so they currently appear as placeholders on habitica.com

Schermata 2020-04-28 alle 13 21 37

website/client/src/components/auth/socialAuth/google.vue
ShupingHe: yep, one thing of the remain works is to abstract the validate logic and some auth logic, i will finish later this week.

ShupingHe:
@paglias is it possible to unify these placeholder text among different pages? and also the button text. or there would be some conditional directives, which make the codebase complicate.

website/client/src/components/auth/register.vue
ShupingHe: got it~ thx for point this out

ShupingHe:
image

image

for example, i made the register button in goup-plans page and other pages all say `sign up`

paglias:
@Tressley could you give a look at the things pointed out in the comments? In particular Join vs Sign Up and the lack of labels in the home page sign up form

Tressley:
For button copy, let's go with "Sign up"

As for the labels, we should introduce them on the homepage and registration form. The placeholders as labels is pretty poor UX. Using them as examples when associated with labels is preferred.

ShupingHe:
got it, then i will stay the current design, fix them later this week.

ShupingHe:
@paglias i abstracted the social auth logic into actions and also change the `div` element into `form`.
i didn't abstract the verification logic, cause for now only the register form use these logic. you can continue the review

ShupingHe:
and also i'm curious about why not just use the offcial `vuex` for state management instead create a customized one. it's not a good development experience to debug without `devtools`

website/client/src/components/auth/socialAuth/apple.vue
```diff
@@ -0,0 +1,109 @@
...
+  data () {
+    const data = {};
+    data.icons = Object.freeze({
+      appleIcon,
+    });
+    return data;
+  },
+  computed: {
+    registering () {
+      if (this.$route.path.startsWith('/register')) {
```
benkelaar: I'd recommend not inferring this kind of information from context (in this case the path) but to pass it in as required properties instead. That way you keep the responsibility of this knowledge in the place where it needs to be.

(so instead of doing `<apple-auth />` to create this component, you do `<apple-auth registering=true>` and similar things)

website/client/src/components/auth/socialAuth/apple.vue
```diff
@@ -0,0 +1,109 @@
+<template>
```
benkelaar: Great start with adding some much need abstraction layers!

However, looking at the level of duplication between this one and the facebook and google one, perhaps they can be abstracted further.

The only things that are different in the three components are `networkName`, `icon` and the `socialAuth` method. Ideally you could create the three components by only supplying those three differences.

Edit: Now see there's a little extra markup in the apple-icon as well. Perhaps we can add that in as an optional parameter or something like that.

website/client/src/components/auth/socialAuth/apple.vue
```diff
@@ -0,0 +1,109 @@
...
+    }"
+    @click="socialAuth('apple')"
+  >
+    <div
+      class="svg-icon social-icon apple-icon"
+      v-html="icons.appleIcon"
+    ></div>
+    <span>{{ (registering || isHomePage || isGroupPlansPage)
+      ? $t('signUpWithSocial', {social: 'Apple'})
+      : $t('loginWithSocial', {social: 'Apple'}) }}</span>
```
benkelaar:
```suggestion
    <span>{{ $t(showSignUp ? 'signUpWithSocial' : 'loginWithSocial', {social: 'Apple'}) </span>
```

website/client/src/components/auth/socialAuth/apple.vue
benkelaar: Continuing on this path, what the code here needs to know is not "where it is shown", but it needs to now what it needs to do. "Should it show the signup instead of the login?" "Should it show the fancy graphics on the button?".

So then ideally the inclusion would be something like this: `<apple-auth show-sign-up looks-fancy/>`. (The properties don't need a value since boolean properties default to true when they're present)

There's probably a better name for `looks-fancy`, but you get the gist. This way you keep the API of the component about what this component can do and you fully decouple it from where it is used.

benkelaar (commented):
Hi @Stevens1995 ! I was looking at this same code for https://github.com/HabitRPG/habitica/issues/10786 . You're doing some awesome work here! 

I left a couple of review comments, do with them what you will. If you think my suggestions would be too much work let me know, I can look at them in a follow-up issue.

> and also i'm curious about why not just use the offcial vuex for state management instead create a customizedone. it's not a good development experience to debug without devtools

I'm also curious, just went on a pretty big bender trying to figure out why the vue component unit tests don't isolate properly and I think it might be related to this. @paglias Do you know why this was done?

website/client/src/components/auth/socialAuth/apple.vue
ShupingHe: @benblank  thx for this suggestion! indeed template should not contain much logic, so i add a computed property `showSignUp`.

website/client/src/components/auth/socialAuth/apple.vue
ShupingHe: i abstracted multi social auth component into one, and indeed they repeat lots of codes

website/client/src/components/auth/socialAuth/apple.vue
ShupingHe: well..i delete this computed property... instead use a prop named `login`

paglias:
If I remember correctly it was because it wasn’t possible to have 
Vuex mutations work with the code in habitica/website/common/script

But if someone wants to look into getting Vuex to work (as long as it doesn’t require huge rewrites) that’s fine for me. The API of our custom implementation should be the same except for the lack of mutations

paglias:
I've created an issue about switching to Vuex https://github.com/HabitRPG/habitica/issues/12155

paglias:
@Stevens1995 this is now done and ready for review?

website/client/src/components/auth/register.vue
```diff
@@ -0,0 +1,343 @@
...
+      }
+      return false;
+    },
+    isPlansPage () {
+      if (this.$route.path.startsWith('/static/plans')) {
+        return true;
+      }
+      return false;
+    },
+    usernameValid () {
```
benkelaar, line 117: Here too I would recommend not checking the path of the current page. It is knowledge this abstraction is not supposed to have (And makes future changes in the other pages likely to cause bugs in this component without the person changing that page realising it). A quick win would be to replace these computed values with injected properties just like you did for social-auth.

And here also I'd recommend going for descriptive properties instead of consumption-driven properties. A large part of the home page variation seems to be a "showLabels" variation. However remember that you can also let parents apply styling to divs in child components using [deep selectors](https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors). You could for example do something like:

```css
// pseudo-css since I don't know what the actual id of surrounding element is
#register-form >>> label {
  display: none;
}
```
This could also be a solution for the "in what page am I being displayed"-specific css.

Depending on which CSS pre-processor Habitica uses you may need to use one of the other deep selectors (`/deep/` or `::v-deep`) but the link should have the info you need.

website/client/src/components/auth/socialAuth.vue
```diff
@@ -0,0 +1,125 @@
...
+  computed: {
+    isHomePage () {
+      if (this.$route.path.startsWith('/static/home')) {
+        return true;
+      }
+      return false;
+    },
+    translateOptions () {
+      return {
+        social: this.network.charAt(0).toUpperCase() + this.network.slice(1),
```
benkelaar: Nice! Even better would be to use [lodash#startCase](https://lodash.com/docs/4.17.15#startCase).

website/client/src/components/auth/register.vue
paglias: `::v-deep` should be used

website/client/src/components/auth/socialAuth.vue
```diff
@@ -0,0 +1,125 @@
...
+    });
+    return data;
+  },
+  computed: {
+    isHomePage () {
+      if (this.$route.path.startsWith('/static/home')) {
+        return true;
+      }
+      return false;
+    },
```
benkelaar: This social-auth element is looking awesome! 

Only thing I think is still a risk is this path check. Either pass it in as property or fix via deep css selectors. (See [comment on `register.vue`](https://github.com/HabitRPG/habitica/pull/12141/files#r419163040) for more elaborate reasoning).

website/client/src/components/auth/socialAuth.vue
ShupingHe: resolved.

website/client/src/components/auth/login.vue
```diff
@@ -0,0 +1,244 @@
+<template>
+  <form
+    v-if="forgotPassword"
+    id="forgot-form"
+    @submit.prevent="forgotPasswordLink"
```
paglias, line 5: are both `@submit` and `@keyup.enter` necessary? I think `@keyup.enter` might be redundant

website/client/src/components/auth/login.vue
```diff
@@ -0,0 +1,244 @@
...
+        id="usernameInput"
+        v-model="username"
+        class="form-control"
+        type="text"
+        :placeholder="$t('emailPlaceholder')"
+      >
+    </div>
+    <div class="text-center">
+      <div
+        v-once
```
paglias, line 32: should this be a `button` with `type=submit`?

website/client/src/components/auth/login.vue
```diff
@@ -0,0 +1,244 @@
...
+    }
+    :-ms-input-placeholder { /* IE 10+ */
+      color: $purple-400;
+    }
+    :-moz-placeholder { /* Firefox 18- */
+      color: $purple-400;
+    }
+    ::placeholder { //  Standard browsers
+      color: $purple-400;
+  }
```
paglias: this should be indented once more

website/client/src/components/auth/login.vue
```diff
@@ -0,0 +1,244 @@
...
+
+    .header {
+      h2 {
+        color: $white;
+      }
+
+      color: $white;
+    }
+
+    .gryphon {
```
paglias: I can't find some of these classes and ids used in this component (`.gryphon`, `.habitica-logo`, `#confirmPasswordInput`)

website/client/src/components/auth/register.vue
```diff
@@ -0,0 +1,229 @@
...
+        <router-link :to="{name: 'login'}" exact="exact">
+          <a v-once class="toggle-link" v-html="$t('alreadyHaveAccountLogin')"></a>
+        </router-link>
+      </div>
+    </div>
+  </form>
+</template>
+
+<style lang="scss" scoped>
+  @import '~@/assets/scss/colors.scss';
```
paglias: I think this block can be removed as it simply imports color variables but they're not used then

website/client/src/components/auth/registerLoginReset.vue
```diff
@@ -481,53 +153,68 @@
       height: 31px;
       margin: 2em auto;
     }
+    ::v-deep .register-form {
```
paglias, line 156: is some of the css here duplicated from the  register form component?

website/client/src/components/auth/reset.vue
```diff
@@ -0,0 +1,193 @@
...
+    ::-moz-placeholder { /* Firefox 19+ */
+      color: $purple-400;
+    }
+    :-ms-input-placeholder { /* IE 10+ */
+      color: $purple-400;
+    }
+    :-moz-placeholder { /* Firefox 18- */
+      color: $purple-400;
+    }
+    ::placeholder { //  Standard browsers
```
paglias: This is not specific to this PR but in all places where `::placeholder` is defined all the others aren't really necessary, we use a module called `autoprefixer` that transforms `::placeholder {}` in the various browser specific versions so they can be removed

website/client/src/components/auth/reset.vue
```diff
@@ -0,0 +1,193 @@
...
+      password: '',
+      passwordConfirm: '',
+      resetPasswordSetNewOneData: {
+        hasError: null,
+        code: null,
+      },
+    };
+  },
+  computed: {
+    passwordValid () {
```
paglias, line 117: would it make sense to move these helper functions to a mixin so they can be re-used across various components?

website/client/src/components/auth/socialAuth.vue
paglias: 👍

website/client/src/store/actions/auth.js
```diff
@@ -108,3 +109,26 @@ export function logout (store, options = {}) {
...
+  } catch (e) {} // eslint-disable-line
+
+  try {
+    const auth = await socialLogin(network, redirectUrl);
+
+    await store.dispatch('auth:socialAuth', {
+      auth,
+    });
+
+    setUpAxios();
```
paglias, line 129: Is there a reason why the original code (`window.location.href = '/';`) was changed to

```
setUpAxios();
await store.dispatch('user:fetch', { forceLoad: true });
```

?

website/client/src/store/actions/auth.js
```diff
@@ -108,3 +109,26 @@ export function logout (store, options = {}) {
...
+    await store.dispatch('user:fetch', { forceLoad: true });
+  } catch (err) {
+    console.error(err); // eslint-disable-line no-console
+    // logout the user
+    await socialLogout(network);
+    store.dispatch('auth:facebookAuth', params);
```
paglias, line 135: is this expected?

paglias (commented):
I've left some comments @Stevens1995 , let me know if something isn't clear or if I missed something. Thanks again for the great PR

paglias:
Also I haven't tested it extensively but when the code is ready we'll try it for a few days before it goes live

benkelaar:
About the conflict in home.vue, you can just override it with the version from your branch (It was a small touchup I did in a different branch).

Also when you merge the current state of develop in, the home.spec.js unit test will start failing. Easiest fix for now would be to just replace the `shallowMount` in that test with a `mount` and then I think it should start working again.

website/client/src/components/auth/login.vue
```diff
@@ -0,0 +1,244 @@
...
+      } else {
+        redirectTo = '/';
+      }
+
+      // @TODO do not reload entire page
+      // problem is that app.vue created hook should be called again
+      // after user is logged in / just signed up
+      // ALSO it's the only way to make sure language data
+      // is reloaded and correct for the logged in user
+      window.location.href = redirectTo;
```
benkelaar, line 192:
```suggestion
      // @TODO do not reload entire page
      // problem is that app.vue created hook should be called again
      // after user is logged in / just signed up
      // ALSO it's the only way to make sure language data
      // is reloaded and correct for the logged in user
      window.location.href = this.$route.query.redirectTo || '/';
```

This was the touchup change I did which is the current conflict with develop.

website/client/src/components/auth/reset.vue
ShupingHe: for now only register.vue use these verification logic, so i didn't separate them from register.vue

ShupingHe:
well, there might still some details to be fixed, redundant code or something,i will check them more carefully, and also should i rebase all these commits into just one?

website/client/src/store/actions/auth.js
ShupingHe: well, my bad, authForm.vue and home.vue got different behaviors after dispatch socialAuth actions

paglias:
@Stevens1995 no worries about the commits, we squash all commits into a single one when a PR is merged so it's not necessary to rebase, but if you prefer to do it anyway for code organization that's fine

website/client/src/components/auth/register.vue
```diff
@@ -0,0 +1,229 @@
...
+      </div>
+      <p v-once class="form-text small" v-html="$t('termsAndAgreement')">
+      </p>
+    </div>
+    <div class="text-center">
+      <button
+        type="submit"
+        class="btn btn-block btn-info sign-up"
+        :disabled="signupFormInvalid"
+        @click="register"
```
paglias: this should be removed since there's the @submit handler on the form or it'll result in two requests to the server

paglias:
Outdated

paglias:
Hi @Stevens1995 , is this ready for review now or there's something else that still needs to be fixed?

paglias:
I've done a new round of testing:

### signup
- the focus style (purple border) is missing for fields with a valid value
- usernames validation is only triggered if you type at least 4 characters but usernames can also have 1 character so they should get validation as soon as something is typed
- inputs with invalid data have purple text instead of white

Schermata 2020-06-01 alle 15 52 48

- error messages text style and margins are still a bit different

### group plans signup
- invalid emails don't get any error message shown


^ I'm going to cancel my previous report so we don't have duplicates

paglias:
@Stevens1995 made a new round of testing and we're almost there!

The last two issues I could find are:

### signup
- if you type something in the username field then delete everything the empty field is sent for validation resulting in an `Invalid Input` error. I think this check that was removed https://github.com/HabitRPG/habitica/pull/12141/commits/2b4c4ba82b86f551e30229cf8e694ae3397d631f#diff-2876acc73666f2a3715d582e7f02ccc0L163 can be kept with:

```
      if (username.length < 1) {	
        return;	
      }
```

### group plan signup
- invalid emails don't get any error message shown under the input

ShupingHe:
@paglias sorry for my late reply. one problem left is the different behavior after social logining between group-plans route and other route. I will figure it out this weekend and finish this pr.

paglias:
no worries @Stevens1995, take your time!

ShupingHe:
@paglias  i found out that the current online register component don't show any error message either, do you want me to add this feature instead just do the refactoring work?

paglias:
@Stevens1995 you mean about the invalid email error? if it's not too much work yes I'd add it here

and

> one problem left is the different behavior after social logining between group-plans route and other route. I will figure it out this weekend and finish this pr.

is this issue still present?

ShupingHe:
@paglias ok i'm going to add the invalid email error tips here.

ShupingHe:
```text
one problem left is the different behavior after social logining between group-plans route and other route. I will figure it out this weekend and finish this pr.
```
still need to figure it out

paglias:
> still need to figure it out

Ok, let me know if you need any help with that, I'm not exactly sure what the different behavior is

ShupingHe:
@paglias the current email validator doesn't return any error messages, instead just a boolean type result, so if we want show any messages about the email format, we would need to rewrite the email validator., which may be some extra works, we'd better not do here.

paglias:
@Stevens1995 that's fine! Let's just show a generic error when the email validator returns false, if you need a string to use you can use the string `notAnEmail` which already exists and in english is simply `Invalid email address.`

paglias:
@Stevens1995 is it now ready for review?

ShupingHe:
@paglias on the group-plans page, after register,the page should not redirect to the root route, which is different with other pages, have you noticed this issue? this is the last issue to fix.

paglias:
@Stevens1995 sorry for the delay, can we pass a parameter when registering from the group plans route to redirect to the different url? let me know if you need help with that

paglias:
It should be possible to use the `redirectTo` query parameter since there's already a condition where it uses the value of that query parameter to redirect to a different url instead of root

paglias:
Hi @Stevens1995 ! Sorry to be bothering you but I just wanted to check on the status here :) This PR has been a lot of work and I'd love for it to get merged, but if you cannot finish it that's totally understandable, just let me know and I'll work on the final touches

ShupingHe:
@paglias ...you can finish the rest if you want, I am not sure if I have time to finish the remaining work...

SabreCat:
With @paglias moving on from the project by and large, and the PR author unable to continue, I will regretfully close this one!

benkelaar:
Alright, I thought it was a promising effort though, so I might pick it up at some point.

PR #12143, opened 2020-04-27
Reset should unassign Stat Points (previously known as Attribute Points) #12123

### LINKED ISSUE #12123: Reset should unassign Stat Points (previously known as Attribute Points)
When you use [Reset Account](https://habitica.fandom.com/wiki/Settings#Reset_Account) (from User Icon > Settings > "Danger Zone" section), it correctly sets your Level to 1 but it does not reset your Stat Points. It should reset them as described below.

Ideally we'd fix this with the generic function described at https://github.com/HabitRPG/habitica/issues/6482 but that requires a large body of work and no one has taken it up yet. Doing a fix for Reset alone will be much simpler and will bring immediate benefits (e.g., for new players who were experimenting with Habitica and want a fresh start).

I'm marking this as "good first issue" although it's larger than we'd normally use that label for. The information below should help a new contributor though and you can comment here if you have any questions.

Stat Points are stored in a user's account like this:
```
{
    "stats": {
        "lvl": 14,
        "points": 2,
        "str": 3,
        "con": 3,
        "int": 3,
        "per": 3
    }
}
```

That's a user at level 14. They have 3 points allocated to each of the Stats (STRength, CONstitution, INTelligence, PERception), and they have 2 points unassigned.

After using Reset currently, the player's account will look like this, which is wrong:
```
{
    "stats": {
        "lvl": 1,
        "points": 2,
        "str": 3,
        "con": 3,
        "int": 3,
        "per": 3
    }
}
```

It **should** look like this:
```
{
    "stats": {
        "lvl": 1,
        "points": 1,
        "str": 0,
        "con": 0,
        "int": 0,
        "per": 0
    }
}
```

I.e., `stats.points` should the same as `stats.lvl` (1, the user's level) and the individual Stats should be set to 0.

This is where Reset sets the level to 1 so it would be a good place to reset the Stat Points:
https://github.com/HabitRPG/habitica/blob/d03e5e93b09f5a16598ed63b911c074d28c62368/website/common/script/ops/reset.js#L4-L9

We would want tests for this too. Below is the test for how Reset changes the level to 1. Tests for setting `stats.points` to 1 and the individual stats to 0 should be in the same file and can use similar code.

https://github.com/HabitRPG/habitica/blob/d03e5e93b09f5a16598ed63b911c074d28c62368/test/common/ops/reset.js#L47-L53

PitiTheGrey:
Fixes https://github.com/HabitRPG/habitica/issues/12123


test/common/ops/reset.js
```diff
@@ -52,6 +52,20 @@ describe('shared.ops.reset', () => {
     expect(user.stats.lvl).to.equal(1);
   });
 
+  it('resets user\'s stat points (str, con, int, per)', () => {
+    user.stats.str = 2;
+    user.stats.con = 2;
+    user.stats.int = 2;
+    user.stats.per = 2;
```
Alys, line 59: Could you please add stats.points to this and to the expect() statements below? :)

In case you're not aware, you can make another commit on the same branch in your local install and push it to your github fork, and it will automatically appear in this PR.

PitiTheGrey: Thank you for your support and encouragement, Alys! You are super!

paglias:
Hi @PitiTheGrey , I'm not exactly sure what went wrong but your fork of Habitica seems to be a bit out of date:

Schermata 2020-04-28 alle 12 51 47

And it seems to have resulted in changes to the reset op to be applied on an old version of the code, you can see in the diff for example:

Schermata 2020-04-28 alle 12 52 40

But the current code is already using `export default` https://github.com/HabitRPG/habitica/blob/develop/website/common/script/ops/reset.js#L4

Could you try to update your fork to the current version of the `develop` branch in Habitica and update the PR? Thanks

PitiTheGrey:
Please close this pull request.  
There is a new one with the correct database. ;)

Alys:
Closing in favour of https://github.com/HabitRPG/habitica/pull/12148

PR #12146, opened 2020-04-28
fix: return full inbox on user-v3-api 

### LINKED ISSUE #12144: /api/v3/user GET route is returning only 10 inbox messages - should return all
_Note to contributors: I'm marking this as help wanted but before you start working on it it's important that you post here first and wait for a reply because it's possible that Habitica's staff will be fixing it themselves, which means if you've already starting on it your time would be wasted!_

Until recently, the [https://habitica.com/api/v3/inbox/messages](https://habitica.com/apidoc/#api-Inbox-GetInboxMessages) GET route was returning all inbox messages.

Currently it's returning only the most recent 10.

I think this problem started in just the past few days. I'm pretty sure that I used that route via a third-party tool on the weekend without noticing any problems, although if I was focusing on just recent messages it's possible the bug was happening then too.

negue:
changed the default option param to `undefined` so that only routes with `page` would get the paginated results.

fixes #12144


Alys (approved):
I've tested this on a local install and curl with the `http://localhost:8080/api/v3/user` GET route is correctly returning all of the inbox messages.

test/api/v3/integration/user/GET-user.test.js
```diff
@@ -38,4 +38,27 @@ describe('GET /user', () => {
...
+
+    const allMessagesPromise = [];
+    for (let i = 0; i < amountOfMessages; i += 1) {
+      allMessagesPromise.push(
+        otherUser.post('/members/send-private-message', {
+          toUserId: user.id,
+          message: `Message Num: ${i}`,
+        }),
+      );
+    }
```
benkelaar: Can be done without local mutation with `lodash/range`. Also think `Promise` should be the plural part, since it's a list of message promises, not a promise for all messages.

```suggestion
    const allMessagePromises = range(amountOfMessages)
      .map(i => otherUser.post('/members/send-private-message', {
        toUserId: user.id,
        message: `Message Num: ${i}`,
      }));
```

Or if you want to get fancy about it:
```javascript
    const allMessagePromises = range(amountOfMessages)
      .map(i => `Message Num: ${i}`)
      .map(message => ({ toUserId: user.id, message }))
      .map(body => otherUser.post('/members/send-private-message', body));
```

Once you've got that statement down to a single statement you might even consider inlining it in the `await Promise.all(/*...*/);`, but I'm not sure if I like that.

test/api/v3/integration/user/GET-user.test.js
```diff
@@ -38,4 +38,27 @@ describe('GET /user', () => {
...
+
+    await Promise.all(allMessagesPromise);
+
+    const returnedUser = await user.get('/user');
+
+    expect(returnedUser._id).to.equal(user._id);
+    expect(returnedUser.inbox).to.exist;
+    expect(Object.keys(returnedUser.inbox.messages)).to.have.a.lengthOf(12);
```
benkelaar: Why not compare with `amountOfMessages`?

negue: yeah, wonder why I didn't use the variable, thanks for catching

PR #12148, opened 2020-04-28
Reset should unassign Stat Points (previously known as Attribute Points) #12123

### LINKED ISSUE #12123: Reset should unassign Stat Points (previously known as Attribute Points)

PitiTheGrey:
Fixes #12123 
uuid:  9226cae6-d852-45da-a51b-ec2c9cb759e6

### DIFF
--- a/website/client/src/components/settings/resetModal.vue
+++ b/website/client/src/components/settings/resetModal.vue
@@ -37,8 +37,9 @@ export default {
       this.$root.$emit('bv::hide::modal', 'reset');
     },
     async reset () {
+      this.$router.push('/');
+      setTimeout(() => window.location.reload(true));
       await axios.post('/api/v4/user/reset');
-      window.location.reload(true);
     },
   },
 };
--- a/website/client/src/components/settings/resetModal.vue
+++ b/website/client/src/components/settings/resetModal.vue
@@ -37,9 +37,9 @@ export default {
       this.$root.$emit('bv::hide::modal', 'reset');
     },
     async reset () {
+      await axios.post('/api/v4/user/reset');
       this.$router.push('/');
       setTimeout(() => window.location.reload(true));
-      await axios.post('/api/v4/user/reset');
     },
   },
 };
--- a/website/client/src/components/settings/resetModal.vue
+++ b/website/client/src/components/settings/resetModal.vue
@@ -38,6 +38,7 @@ export default {
     },
     async reset () {
       const response = await axios.post('/api/v4/user/reset');
+      window.location.reload(true);
       // @TODO: Not sure if this is correct
       this.$store.user = response.data.data.user;
       this.$router.push('/');
--- a/website/client/src/components/settings/resetModal.vue
+++ b/website/client/src/components/settings/resetModal.vue
@@ -39,10 +39,6 @@ export default {
     async reset () {
       const response = await axios.post('/api/v4/user/reset');
       window.location.reload(true);
-      // @TODO: Not sure if this is correct
-      this.$store.user = response.data.data.user;
-      this.$router.push('/');
-      this.$root.$emit('bv::hide::modal', 'reset');
     },
   },
 };
--- a/website/client/src/components/settings/resetModal.vue
+++ b/website/client/src/components/settings/resetModal.vue
@@ -37,7 +37,7 @@ export default {
       this.$root.$emit('bv::hide::modal', 'reset');
     },
     async reset () {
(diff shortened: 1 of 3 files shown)


website/client/src/components/settings/resetModal.vue
```diff
@@ -38,6 +38,7 @@ export default {
     },
     async reset () {
       const response = await axios.post('/api/v4/user/reset');
+      window.location.reload(true);
```
paglias: Hmm, I was a bit confused why this was necessary, I thought the reset already had a built in reload... It looks like it didn't although it was necessary, thanks!

Could you remove the extra lines in this method? since they're not used anymore I think it can be simplified to:

```
async reset () {
      await axios.post('/api/v4/user/reset');
      window.location.reload(true);
}
```

paglias (changes requested):
left one comment @PitiTheGrey , but apart from that it looks good!

website/client/src/components/settings/resetModal.vue
PitiTheGrey: Not sure what happend here...  
I cannot read much out of the lint error report. Shall I better put the lines of code back?

website/client/src/components/settings/resetModal.vue
```diff
@@ -37,11 +37,8 @@ export default {
       this.$root.$emit('bv::hide::modal', 'reset');
     },
     async reset () {
-      const response = await axios.post('/api/v4/user/reset');
-      // @TODO: Not sure if this is correct
-      this.$store.user = response.data.data.user;
-      this.$router.push('/');
-      this.$root.$emit('bv::hide::modal', 'reset');
+      await axios.post('/api/v4/user/reset');
```
paglias, line 40: @PitiTheGrey the new changes are working great but I just realized that this way you will not get redirected to the tasks page like the original code,

```
      this.$router.push('/');
      setTimeout(() => window.location.reload(true));
```

^ should fix it, this way you first go to the tasks page and then it immediately reloads the page (the setTimeout is needed to make sure the reload doesn't happen before you reach the tasks page)

paglias (changes requested):

paglias:
Merged in https://github.com/HabitRPG/habitica/commit/ca4efa21cf5527604073c1b9c7ac0b35b4eeaf30

paglias:
thanks @PitiTheGrey ! Noted in your profile and awarded the Blacksmith title

Issue #10209, opened 2018-03-31
converting from moment().zone to moment().utcOffset


Alys:
`moment().zone is deprecated, use moment().utcOffset instead.` http://momentjs.com/guides/#/warnings/zone/

From the utcOffset docs:

> NOTE: Unlike moment.fn.zone this function returns the real offset from UTC, not the reverse offset (as returned by Date.prototype.getTimezoneOffset).

It's not a matter of merely changing the code to use `moment().utcOffset` instead of `moment().zone` because we store the `zone` value in the user document in two places. Here's an example taken just after a user changed time zones from UTC+0 to UTC+10:
```
    "preferences": {
        "timezoneOffset": -600,
        "timezoneOffsetAtLastCron": 0
    },
```
Despite the name of those preference values, they are actually `moment().zone` values.

Those preference values are set by both the mobile apps and the website so we can't change those values from `zone` to `utcOffset` because any player who used the website and a non-upgraded app would get inconsistent data stored in their account. Also the values might be used by third-party tools and so removing or changing them would be a breaking change.

**I guess we can change our code to use `moment().utcOffset` but keep storing `moment().zone` values in the preferences and have our own function to convert between the formats immediately after the preference data is read from the user's document and just before it's written back.** It's a bit messy but it would be backwards compatible. We could change the preferences values in API v4.

Any better ideas?

As @crookedneighbor said in https://github.com/HabitRPG/habitica/pull/7053 this isn't urgent until the next version of moment comes out and we want to upgrade.

Code that uses `moment().zone`:  _[EDIT 2020-03-31: this list of code will be outdated now]_
```
$ git grep "\.zone"
test/api/v3/integration/tasks/GET-tasks_user.test.js:161:    let startDate = moment().zone(timezone).subtract('4', 'days').startOf('day').toISOString();
test/api/v3/integration/tasks/GET-tasks_user.test.js:188:    let startDate = moment().zone(timezone).subtract('4', 'days').startOf('day').toISOString();
test/api/v3/integration/tasks/GET-tasks_user.test.js:214:    let startDate = moment().zone(timezone).subtract('4', 'days').startOf('day').toISOString();
test/api/v3/unit/libs/preening.test.js:12:      now: Number(moment('2013-10-20').zone(0).startOf('day').toDate()),
test/api/v3/unit/models/user.test.js:467:      const timezoneOffset = moment().zone('-06:00').zone();
[...]

PR #12154, opened 2020-05-03
Vue component unit test isolation

benkelaar:
Fixes unit test isolation for Vue.js component unit tests

While working on a solution for #10786 I started to write a unit test for the `home.vue` component. That went fairly smoothly but after I had added it I noticed that stray events from the home.vue localVue I created ended up failing random other tests. Clearly test isolation was failing. I still haven't figured out exactly why. My standing theory is that the await on the `setValue(value)` somehow doesn't completely wait on the events that are being fired by dispatch because it's not marked as an asynchronous function itself. I think this lack of a-synchronicity somehow breaks the Promise chain. In any case adding the async keyword to the dispatch function seems to fix the test isolation issues.

I'm also tracking down the different stacktraces that are being thrown by Vue.js component unit tests, because they unnecessarily confuse investigations into what is happening. Shall I add those to this merge request or shall I create a new merge request for that?

### Changes
Most significant change is the addition of the `async` keyword to the `dispatch` function in `/src/libs/store/index.js`

Beyond that:
- Added unit test for `home.vue` component
- Small cleanup in `home.vue`
- Replaced use of `zone` by `utcOffset` in `/src/store/index.js`
- Couple other touchups.

----

### DIFF
--- a/website/client/src/store/index.js
+++ b/website/client/src/store/index.js
@@ -18,7 +18,7 @@ const IS_TEST = process.env.NODE_ENV === 'test'; // eslint-disable-line no-proce
 let isUserLoggedIn = false;
 
 // eg, 240 - this will be converted on server as -(offset/60)
-const browserTimezoneOffset = -moment().utcOffset();
+const browserTimezoneOffset = moment().zone();
 
 axios.defaults.headers.common['x-client'] = 'habitica-web';
 
--- a/website/client/src/store/index.js
+++ b/website/client/src/store/index.js
@@ -18,7 +18,7 @@ const IS_TEST = process.env.NODE_ENV === 'test'; // eslint-disable-line no-proce
 let isUserLoggedIn = false;
 
 // eg, 240 - this will be converted on server as -(offset/60)
-const browserTimezoneOffset = moment().zone();
+const browserTimezoneOffset = -moment().utcOffset();
 
 axios.defaults.headers.common['x-client'] = 'habitica-web';
 
--- a/website/client/tests/unit/mixins/groupsUtilities.spec.js
+++ b/website/client/tests/unit/mixins/groupsUtilities.spec.js
@@ -2,10 +2,13 @@ import Vue from 'vue';
 import groupsUtilities from '@/mixins/groupsUtilities';
 import { TAVERN_ID } from '@/../../common/script/constants';
 import generateStore from '@/store';
+import Store from '@/libs/store';
+
+Vue.use(Store);
 
 describe('Groups Utilities Mixin', () => {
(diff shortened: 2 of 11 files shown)


benkelaar:
Did a quick check of the usages of dispatch. It looks like all usages either don't care about the return value of dispatch or already handle it via either `await` or promise stacking.

website/client/src/store/index.js
```diff
@@ -18,7 +18,7 @@ const IS_TEST = process.env.NODE_ENV === 'test'; // eslint-disable-line no-proce
 let isUserLoggedIn = false;
 
 // eg, 240 - this will be converted on server as -(offset/60)
-const browserTimezoneOffset = moment().zone();
+const browserTimezoneOffset = -moment().utcOffset();
```
paglias: This probably works well but I think we'll switch all together in one go with the remaining usages of `.zone()` as described in https://github.com/HabitRPG/habitica/issues/10209

website/client/tests/unit/mixins/groupsUtilities.spec.js
```diff
@@ -2,10 +2,13 @@ import Vue from 'vue';
 import groupsUtilities from '@/mixins/groupsUtilities';
 import { TAVERN_ID } from '@/../../common/script/constants';
 import generateStore from '@/store';
+import Store from '@/libs/store';
+
+Vue.use(Store);
```
paglias: should this use a local Vue instance like in the other test?

```
const localVue = createLocalVue();
localVue.use(Store);
```

paglias (changes requested):
left two comments but otherwise looks good 👍 It would also match Vuex's own behavior of always returning a promise from a call to dispatch

website/client/tests/unit/mixins/groupsUtilities.spec.js
benkelaar: Yeah I think so, but there's quite a varied amount of approaches in the different Vue.js component tests. Didn't want to open the pit of unifying them all. This however was required for me to get the tests to run at all. However, perhaps this is no longer necessary, will check.

website/client/src/store/index.js
benkelaar: Cool, will revert for this one.

website/client/tests/unit/mixins/groupsUtilities.spec.js
benkelaar: Alright, got it to work with localVue. Indeed better.

paglias:
thanks @benkelaar ! noted in your profile

PR #12172, opened 2020-05-05
Fix Issue #12103: Wacky potions purchase warning

### LINKED ISSUE #12103: purchase warning for Hatching Potions should stop you buying more than 9 when Mounts don't exist
For premium / wacky Hatching Potions (the ones that can be obtained only with Gems), when you try to buy more than you need, you're given a warning that you don't need that many. (the `purchasePetItemConfirm` string "This purchase would exceed the number of items you need to hatch all possible <%= itemText %> pets. Are you sure?")

This does not work correctly for those potions that you can't get Mounts from (e.g., the new Confection / `Dessert` potion - not to be confused with `Desert`). It should give you the warning when you try to buy more than 9 potions.

https://github.com/HabitRPG/habitica/blob/8da36bf27c62ba0397a6af260c20d35a17f3d911/website/client/src/components/shops/buyModal.vue#L566-L569

<strike>I'm marking this suggestion-discussion in case the staff want to comment on how this should be fixed. If there's no comments or objections from them after about a week, I'll mark it as help wanted with a warning that before a Blacksmith starts coding for it, they should describe their proposed solution here so that it can be assessed for suitability (e.g., for ease of maintenence as other such Hatching Potions are introduced).</strike>

fmaximus:

Fixes #12103

### Changes
For wacky potions lower petsRemaining by 9,
to exclude the mounts that don't exist.

### Tests
- [x] manual
- [x] npm test 

Verified that it shows the warning when the total would exceed 9, for Confection Potion.
Verified that it only shows the warning when the total would exceed 18 for Premium potion like Silver.

----

### DIFF
--- a/website/client/src/components/shops/buyModal.vue
+++ b/website/client/src/components/shops/buyModal.vue
@@ -377,6 +377,7 @@
 
 <script>
 import keys from 'lodash/keys';
+import size from 'lodash/size';
 import reduce from 'lodash/reduce';
 import moment from 'moment';
 
@@ -407,8 +408,12 @@ import Avatar from '@/components/avatar';
 
 import seasonalShopConfig from '@/../../common/script/libs/shops-seasonal.config';
 import { drops as dropEggs } from '@/../../common/script/content/eggs';
+import { drops as dropPotions } from '@/../../common/script/content/hatching-potions';
 
 const dropEggKeys = keys(dropEggs);
+
+const amountOfDropEggs = size(dropEggs);
+const amountOfDropPotions = size(dropPotions);
 
 const hideAmountSelectionForPurchaseTypes = [
   'gear', 'backgrounds', 'mystery_set', 'card',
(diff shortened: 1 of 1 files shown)


website/client/src/components/shops/buyModal.vue
```diff
@@ -542,6 +542,9 @@ export default {
         }, 0);
         if (this.item.pinType === 'premiumHatchingPotion') {
           petsRemaining -= this.user.items.hatchingPotions[this.item.key] + 2 || 2;
+          if (this.item.path.indexOf('wackyHatchingPotions.') !== -1) {
+            petsRemaining -= 9;
```
paglias: Instead of removing 9 from the count regardless of how many the user is trying to buy, shouldn't the count start from 9 in the case of wacky potions at line 534?

fmaximus: That would be another option, but in that case we should also set it to 18 for normal premium hatching potions, 
instead of subtracting 2 extra.

paglias: I was slightly confused by how this code was originally meant to work but I tried a couple of different combinations and it works well (20 - 2 - 9 leaves exactly 9 available to buy)

I think we should refactor it to something clearer but for the scope of this PR this looks good! I'll merge it at the end of the week

fmaximus:
@paglias I added the refactoring in this commit on another branch.
https://github.com/fmaximus/habitica/commit/1fd447dfcfefcfdd1c6991c4cf3eb12e9af5d020

If you prefer that solution, I will add it here as well.

paglias:
the current fix in this PR works well but it's pretty confusing so yeah I'd use the refactored one, thanks!

paglias:
thanks @fmaximus ! noted towards your 3rd tier

PR #12173, opened 2020-05-05
#11828: Show a boss damage notification for skills

### LINKED ISSUE #11828: Brutal Smash and Burst of Flames skills should show a boss damage notification
When you are in a boss quest and you complete a task, you see a "damage to boss" notification, like the one at the bottom of the screenshot below.

You should see the same notification when you do boss damage from using the Brutal Smash and Burst of Flames skills. Those skills used to give a notification but at some point they've stopped doing that.

image
Ignore the grey box at the right of the screenshot, it's just an artifact of how I cropped the screenshot.

fmaximus:
Brutal Smash and Burst of Flames skills should show a boss damage notification

Fixes #11828

### Changes

Check for damage done was done for spells targeting party instead of spells targeting tasks.

----

### DIFF
--- a/website/client/src/components/tasks/spells.vue
+++ b/website/client/src/components/tasks/spells.vue
@@ -300,7 +300,7 @@ export default {
       const { user } = this;
       if (!user.party.quest) return 0;
 
-      const userQuest = this.quests[user.party.quest.key];
+      const userQuest = this.quests.quests[user.party.quest.key];
 
       if (!userQuest) {
         return 0;
--- a/website/client/src/mixins/spells.js
+++ b/website/client/src/mixins/spells.js
@@ -77,8 +77,7 @@ export default {
     },
     async castEnd (target, type) {
       if (!this.$store.state.spellOptions.castingSpell) return null;
-      let beforeQuestProgress;
-      if (this.spell.target === 'task') beforeQuestProgress = this.questProgress();
+      const beforeQuestProgress = this.questProgress();
 
       if (!this.applyingAction) return 'No applying action';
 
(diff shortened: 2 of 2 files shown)


website/client/src/mixins/spells.js
```diff
@@ -78,7 +78,7 @@ export default {
     async castEnd (target, type) {
       if (!this.$store.state.spellOptions.castingSpell) return null;
       let beforeQuestProgress;
-      if (this.spell.target === 'party') beforeQuestProgress = this.questProgress();
+      if (this.spell.target === 'task') beforeQuestProgress = this.questProgress();
```
paglias: this way the notification won't work when the target is `party`? Maybe we should calculate the quest progress in both case?

Also since not all tasks spells result in damage to the boss, we should maybe try do something more generic like:

1. get boss damage before spell is cast
2. get boss damage after spell is cast
3. if the difference is > 0 show the notification, otherwise not?

This would'n require checking for the spell to see if it casts or not damage to the boss

website/client/src/components/tasks/spells.vue
```diff
@@ -300,7 +300,7 @@ export default {
       const { user } = this;
       if (!user.party.quest) return 0;
 
-      const userQuest = this.quests[user.party.quest.key];
+      const userQuest = this.quests.quests[user.party.quest.key];
```
paglias, line 303: hmm I thought this was an error but looks like we've been using the wrong object for a long time... good catch!

paglias (commented):
thanks for the PR @fmaximus ! I've left a couple of comments

website/client/src/mixins/spells.js
fmaximus: The checking of `difference > 0` already happens.
I checked all spells, currently all spells that invoke boss damage are `task` spells.

paglias: I understand but in case we change a spell which does not target a task it would be nice to get it to show the boss damage without any code change here

paglias:
thanks @fmaximus ! welcome to the 2nd tier

PR #12176, opened 2020-05-06
FIX: scroll/loading guilds

negue:
Using the same debounce time as challenges of 1s that way it won't trigger to load on every single pixel-scroll

### DIFF
--- a/website/client/src/components/groups/discovery.vue
+++ b/website/client/src/components/groups/discovery.vue
@@ -55,7 +55,6 @@ b-dropdown(:text="$t('sort')", right=true)
             :handler="triggerFetchGuilds"
             :should-handle="!loading && !hasLoadedAllGuilds"
             :handle-on-mount="true"
-            :threshold="20"
           >
             <span v-once>{{ $t('loading') }}</span>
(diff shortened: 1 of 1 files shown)


website/client/src/components/groups/discovery.vue
```diff
@@ -52,9 +52,10 @@ b-dropdown(:text="$t('sort')", right=true)
           />
           <mugen-scroll
             v-show="loading"
-            :handler="fetchGuilds"
+            :handler="triggerFetchGuilds"
             :should-handle="!loading && !hasLoadedAllGuilds"
             :handle-on-mount="true"
+            :threshold="20"
```
paglias: I think threshold should be between 0 and 1 https://github.com/egoist/vue-mugen-scroll#threshold

paglias (changes requested):
I think there's just one thing to be changed then it can be merged

website/client/src/components/groups/discovery.vue
negue: oops, I mistook it with scrolling threshold ... removed it again.

Good catch :tada:

PR #12177, opened 2020-05-06
fix(chat) - Mention dot doesn't show if mention is preceded by weird …

benkelaar:
As discovered by @citrusella  in the Coding for habitica guild, the mention highlight dot does not display if the mention is preceded by a mention with a different character (Not space or @).

This merge request fixes that.

### Changes
- Removed buggy optimisation for @mention determination.

----


paglias:
I saw the issue on the Blacksmiths guild, I wonder if there's also a problem with the server side push notifications system since @citrusella mentioned they haven't received push notifications on their phone during the last week. I gave a look at the push notifications logs and couldn't find anything

paglias:
thanks! noted in your profile

PR #12195, opened 2020-05-11
Bolding text inside a link in task title/notes. (#12178)

### LINKED ISSUE #12178: Bolding text inside a link on a task causes new tab to open with link + task to open for editing
Reported by JohnJSal:

> Either a bug or just weird behavior: I noticed when I make a clickable link in a task note and put it in bold at the top, it will open that link in a new tab, but when I return to the original tab, it also opens up that task for editing, as if I had just clicked on it normally. This was the task:
> 
> `[**New Year's Resolution**](https://habitica.com/groups/guild/6e6a8bd3-9f5f-4351-9188-9f11fcd80a99) Post in the Guild chat about how things are going, or respond to someone else's comment. Share any successes or difficulties and support your fellow Guild members! * *Note that completing this Daily is not a requirement to be eligible to win the Challenge*`
> 
> I added that first line.

I've tested this out in Firefox, and it turns out it only happens with that specific format; the bolding needs to be inside the link. This would be fine:

> `**[New Year's Resolution](https://habitica.com/groups/guild/6e6a8bd3-9f5f-4351-9188-9f11fcd80a99)**`

fmaximus:

Fixes #12178

### Changes

Ascend up the ancestors of the click target, up until the node defining the click handler.
If any of them is an `<a>` element, don't open the edit task popup.

When clicking on a link, with a bold and/or italic link description: 
Before:
* link target opened in new tab
* task edit popup opened

Now:
* only link target opened in new tab

----

### DIFF
--- a/website/client/src/components/tasks/task.vue
+++ b/website/client/src/components/tasks/task.vue
@@ -976,10 +976,16 @@ export default {
     edit (e, task) {
       if (this.isRunningYesterdailies || !this.showEdit) return;
 
-      // Prevent clicking on a link from opening the edit modal
       const target = e.target || e.srcElement;
 
-      for (let element = target; element.className === undefined || element.className.indexOf('task-clickable-area') === -1; element = element.parentNode) {
+      /*
+       * Prevent clicking on a link from opening the edit modal
+       *
+       * Ascend up the ancestors of the click target, up until the node defining the click handler.
+       * If any of them is an <a> element, don't open the edit task popup.
+       * Needed in case of a link, with a bold and/or italic link description
+       */
+      for (let element = target; !element.classList.contains('task-clickable-area'); element = element.parentNode) {
         if (element.tagName === 'A') return; // clicked on a link
       }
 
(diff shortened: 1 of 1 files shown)


website/client/src/components/tasks/task.vue
```diff
@@ -979,7 +979,9 @@ export default {
       // Prevent clicking on a link from opening the edit modal
       const target = e.target || e.srcElement;
 
-      if (target.tagName === 'A') return; // clicked on a link
+      for (let element = target; element.className === undefined || element.className.indexOf('task-clickable-area') === -1; element = element.parentNode) {
```
paglias: Thanks @fmaximus , I think instead of using `element.className` and checking for `undefined` using `classList` and `.contains` should be easier to read https://developer.mozilla.org/en-US/docs/Web/API/Element/classList https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList 

I'd also add a comment explaining what the for loop does and why it's necessary

paglias:
thanks @fmaximus , noted towards your 2nd tier

PR #12196, opened 2020-05-12
Display descriptive error when mongod isn't running

### LINKED ISSUE #12101: Display descriptive error when tests can't be run due to absent database
### General Info
  * UUID: d71d2c57-a73d-4591-b64d-e688584a9092
  * Browser: Command line (zshell for me)
  * OS: Definitely macOS, but also in docker containers

### Description
When developing for habitica locally, running `npm run test:api:unit` without a mongo database running on localhost (default port) will timeout and give an undescriptive timeout error.

It's rather unintuitive that running the unit tests require a database to be up, so it can be hard to figure out that that is the way to fix this problem.

It does seem to specifically be the test:api:unit target that seems to have this problem, running `npm run test` works fine (Except that there's a test in test:common that seems to have a DST or time-zone problem or something (It keeps failing with day-off-by-one errors around midnight)).

In any case, throwing a descriptive error i.o. "Timeout exceeded" would really help new contributors get up to speed with how to run the tests.

#### ~Console~ Terminal Errors
```
[00:12:17] Starting 'test:api:unit'...


  1) "before all" hook
  2) "after all" hook

  0 passing (16s)
  2 failing

  1) "before all" hook:
      at listOnTimeout (internal/timers.js:549:17)
      at processTimers (internal/timers.js:492:7)

  2) "after all" hook:
      at listOnTimeout (internal/timers.js:549:17)
      at processTimers (internal/timers.js:492:7)

```

fmaximus:

Fixes: #12101

### Changes
Add test:prepare:mongo task to check is mongod is running.

----

### DIFF
--- a/gulp/gulp-tests.js
+++ b/gulp/gulp-tests.js
@@ -160,7 +160,7 @@ gulp.task('test:content:safe', gulp.series('test:prepare:build', cb => {
   pipe(runner);
 }));
 
-gulp.task('test:api:unit', done => {
+gulp.task('test:api:unit:run', done => {
   const runner = exec(
     testBin('istanbul cover --dir coverage/api-unit node_modules/mocha/bin/_mocha -- test/api/unit --recursive --require ./test/helpers/start-server'),
     err => {
@@ -174,7 +174,7 @@ gulp.task('test:api:unit', done => {
   pipe(runner);
 });
 
-gulp.task('test:api:unit:watch', () => gulp.watch(['website/server/libs/*', 'test/api/unit/**/*', 'website/server/controllers/**/*'], gulp.series('test:api:unit', done => done())));
+gulp.task('test:api:unit:watch', () => gulp.watch(['website/server/libs/*', 'test/api/unit/**/*', 'website/server/controllers/**/*'], gulp.series('test:api:unit:run', done => done())));
 
 gulp.task('test:api-v3:integration', done => {
   const runner = exec(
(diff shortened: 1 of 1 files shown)


benkelaar:
@paglias Tested it locally, works well!

@fmaximus Awesome work!

gulp/gulp-tests.js
```diff
@@ -235,14 +235,22 @@ gulp.task('test', gulp.series(
   'test:sanity',
   'test:content',
   'test:common',
-  'test:api:unit',
+  'test:prepare:mongo',
+  'test:api:unit:run',
```
benkelaar, line 239: This one can still reference the new series right? (And the same for test:api-v3)

paglias:
The bug only affects unit tests and not integration ones? Also the mongo prepare task deletes the entire database when it's run. It's okay since we're talking about test databases but it's a bit dangerous in case someone runs the tests against a working database by error without realizing the entire database gets dropped.

Can we add a new task that checks the connection without dropping the database?

fmaximus:
I had the issue also for the integration tests.

It drops the database defined in `TEST_DB_URI`, 
which is defined in `config.json.example` as `mongodb://localhost:27017/habitrpg_test`.

Even if your server would still be running, it would not delete the development database
`mongodb://localhost:27017/habitrpg`.

`test:prepare:mongo` was already used in `test:prepare:server`, 
which is meant to be used together with `test:api-v3:integration:separate-server`.

So in my opinion there isn't much difference with running the server in process.

paglias:
@negue @SabreCat what do you think about the database being dropped before the tests run?

negue:
Oh I thought I already answered here ^^

Since my local integration tests not working at all (timeouts to local docker mongo), and I have to use mongodb atlas , I always got the error that it couldn't drop the database at all, but thats just my problem. 

As long the tests (here on the CI) still work with this change, I don't see any problem with this :)

SabreCat:
Given it's referring to the `TEST_DB_URI`, which you'd have to be _deliberately_ awful with your setup to make unsafe to drop, I'm OK with tests dropping the db before start.

paglias:
Sounds good, merging!

thanks @fmaximus , noted towards your 4th tier

PR #12205, opened 2020-05-16
Store original quantities, to avoid reordering in Inventory Items screen

### LINKED ISSUE #12040: When hatching the egg at the end of a row, eggs rearrange
Report from Geolojess: 
> Also, found another display bug! When I'm on the items page and hatching eggs, if I hatch the egg at the very end of the row, then that egg and the eggs in the second row will all switch places! But ONLY if you hatch the one on the end of the first row. I tested to see if hatching any of the eggs in the second row produced the same results but they did not. Only discovered this because I was hatching hippos and when I went to hatch a second I ended up hatching a cuttlefish, because the hippo egg moved! XD (and I'll keep the cuttlefish, no big deal!)

To reproduce...

1. Egg at the end of the row is hatchable:
image

2. Hatch it.

3. It has swapped with an egg from the row below.
image

_EDIT by Alys 2020-05-13:_

As described in the comments below, reordering happens because the numbers of the items change (items are sorted by how many you own). It's disconcerting though when you're still using them.

**The desired fix is that the items should not change order while looking at the page.**
**Reordering should only happen if you reload the page, or sync, or go to another page then back again.**

fmaximus:

Fixes: #12040

### Changes

Create a quantity snapshot when loading the screen.
Sort based on the snapshot after updates.

Items_before

Hatching a Cotton Candy Pink Parrot

Items_after

----

### DIFF
--- a/website/client/src/components/inventory/items/index.vue
+++ b/website/client/src/components/inventory/items/index.vue
@@ -466,18 +466,11 @@ export default {
         });
 
         if (this.sortBy === 'quantity') {
-          const quantitySnapshot = this.quantitySnapshot[groupKey];
-          if (quantitySnapshot) {
-            itemsArray.sort((a, b) => quantitySnapshot[b.key] - quantitySnapshot[a.key]);
-          } else {
-            itemsArray.sort((a, b) => b.quantity - a.quantity);
-          }
-
-          if (quantitySnapshot === null) {
-            this.quantitySnapshot[groupKey] = Object.fromEntries(
-              itemsArray.map(item => [item.key, item.quantity]),
-            );
-          }
+          const quantitySnapshot = this.quantitySnapshot[groupKey] || Object.fromEntries(
+            itemsArray.map(item => [item.key, item.quantity]),
+          );
+          itemsArray.sort((a, b) => quantitySnapshot[b.key] - quantitySnapshot[a.key]);
+          this.quantitySnapshot[groupKey] = quantitySnapshot;
         } else {
           itemsArray.sort((a, b) => a.text.localeCompare(b.text));
         }
--- a/website/client/src/components/inventory/items/index.vue
+++ b/website/client/src/components/inventory/items/index.vue
@@ -466,6 +466,7 @@ export default {
         });
 
         if (this.sortBy === 'quantity') {
+          // Store original quantities, to avoid reordering when using items.
           const quantitySnapshot = this.quantitySnapshot[groupKey] || Object.fromEntries(
             itemsArray.map(item => [item.key, item.quantity]),
           );
--- a/website/client/src/components/inventory/items/index.vue
+++ b/website/client/src/components/inventory/items/index.vue
@@ -419,6 +419,11 @@ export default {
         cardType: '',
         messageOptions: 0,
       },
+      quantitySnapshot: {
+        eggs: null,
+        hatchingPotions: null,
+        food: null,
+      },
     };
   },
   computed: {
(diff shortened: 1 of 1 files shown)


website/client/src/components/inventory/items/index.vue
```diff
@@ -458,12 +465,22 @@ export default {
...
+            itemsArray.sort((a, b) => quantitySnapshot[b.key] - quantitySnapshot[a.key]);
+          } else {
+            itemsArray.sort((a, b) => b.quantity - a.quantity);
+          }
+
+          if (quantitySnapshot === null) {
+            this.quantitySnapshot[groupKey] = Object.fromEntries(
+              itemsArray.map(item => [item.key, item.quantity]),
+            );
+          }
```
benkelaar: You can avoid the initial branching if you invert the order
```suggestion
          const quantitySnapshot = this.quantitySnapshot[groupKey] || Object.fromEntries(
            itemsArray.map(item => [item.key, item.quantity]),
          );
          itemsArray.sort((a, b) => quantitySnapshot[b.key] - quantitySnapshot[a.key]);
          this.quantitySnapshot[groupKey] = quantitySnapshot;
```

website/client/src/components/inventory/items/index.vue
```diff
@@ -458,12 +465,15 @@ export default {
           }
         });
 
-        itemsArray.sort((a, b) => {
-          if (this.sortBy === 'quantity') {
-            return b.quantity - a.quantity;
-          } // AZ
-          return a.text.localeCompare(b.text);
-        });
+        if (this.sortBy === 'quantity') {
```
paglias, line 468: can we add a comment explaining the reasoning behind the need for the quantity snapshot?

website/client/src/components/inventory/items/index.vue
```diff
@@ -458,12 +465,15 @@ export default {
...
+          const quantitySnapshot = this.quantitySnapshot[groupKey] || Object.fromEntries(
```
paglias, line 470: Given an user can have a big number of items and the rendering is already a bit slow, instead of calculating the snapshot here by iterating on the existing `itemsArray` could we do it in the `each` block at line 445 where we already iterate on the items?

paglias (commented):
it's working as expected but left two small comments

website/client/src/components/inventory/items/index.vue
fmaximus: I tried moving it inside, but it doesn't seem to improve the rendering.
I've added `console.time` and `console.timeEnd` around the each + sort:
| time (ms)                | size | develop | PR   | proposal |
|-----------------|------|---------|------|----------|
| eggs            | 8    | 10.8    | 5.5  | 12.8     |
| hatchingPotions | 13   | 12.6    | 12.8 | 14.0     |
| food            | 5    | 4.3     | 3.0  | 2.7      |
| special         | 4    | 3.0     | 1.8  | 5.9      |
| quests          | 2    | 1.8     | 1.0  | 2.6      |

Adding it inside the each, requires the use of an if block,
to check if it's already cached. 
Doing the check for every element seems slower,
than running it twice, and doing the check once.
Once a snapshot is available, the time will drop even further.

Also notice that by pulling up the `sortBy` check out of the comparator,
performance has already been improved.

paglias: thanks for checking this, let's leave this as it is then

website/client/src/components/inventory/items/index.vue
paglias: If you can add a comment here I'll then merge it , thanks :)

paglias:
thanks! noted towards your 3rd tier

PR #12218, opened 2020-05-19
fix(chat) - Issue 12217 - Allow url-escapable characters in links.

### LINKED ISSUE #12217: URLs containing non-English letters cannot be posted in markdown links
If a URL contains non-English characters such as this Russian wiki page: https://habitica.fandom.com/ru/wiki/Снаряжение
then it can't be posted as a markdown URL in a chat message in Habitica. 

I.e., posting this to a guild will fail:
`[test](https://habitica.fandom.com/ru/wiki/Снаряжение)`

This shorter version with only one non-English letter fails too:
`[test](https://habitica.fandom.com/ru/wiki/н)`

This version with all English letters succeeds so the bug is defniitely triggered by non-English letters:
`[test](https://habitica.fandom.com/ru/wiki/C)`

The user sees "An unexpected error occurs" and this appears in the server logs:
```
message:TypeError: Cannot read property 'index' of null
    at /usr/src/habitrpg/website/transpiled-babel/libs/highlightMentions.js:132:15
    at Array.forEach (<anonymous>)
    at findTextBlocks (/usr/src/habitrpg/website/transpiled-babel/libs/highlightMentions.js:128:22)
    at highlightMentions (/usr/src/habitrpg/website/transpiled-babel/libs/highlightMentions.js:165:22)
```

The expected behaviour is that any valid URL should be able to be posted in a markdown link.

I'm marking this as high priority since it's preventing Habitica's wiki URLs being posted as markdown for a variety of non-English wikis (and I assume also some links from other web sites).

benkelaar:
Fixes #12217 

### Changes
Use broader selector for link matching source-mapper.

----

### DIFF
--- a/test/api/unit/libs/highlightMentions.test.js
+++ b/test/api/unit/libs/highlightMentions.test.js
@@ -100,6 +100,13 @@ describe('highlightMentions', () => {
       const result = await highlightMentions(text);
       expect(result[0]).to.equal('http://www.medium.com/@user/blog [@user](/profile/111)');
     });
+
+    // https://github.com/HabitRPG/habitica/issues/12217
+    it('doesn\'t highlight user in link with url-escapable characters', async () => {
+      const text = '[test](https://habitica.fandom.com/ru/@wiki/Снаряжение)';
+      const result = await highlightMentions(text);
+      expect(result[0]).to.equal(text);
+    });
   });
 
   describe('exceptions in code blocks', () => {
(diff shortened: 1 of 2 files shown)


benkelaar:
I guess the "neater" way to fix this would be to un-url-escape the url and then regExpEscape it, but then you need to be able to guarantee feature-parity between markdown-it's url escaping and our url unescaping. I think this way is less error-prone for the future.

paglias:
thanks for the quick fix @benkelaar, noted towards your next tier

PR #11390, opened 2019-09-30
[WIP] Shared completion for Group Dailies

paglias:
TODO 

- [ ] tests

-------------

This is the continuation of https://github.com/HabitRPG/habitica/pull/11179 and now depends on #11389

PR #12219, opened 2020-05-20
Group Dailies Shared Completion redux

SabreCat:

Continuation of #11390

### Changes

Repaired build of @MeanderingCode's shared completion functionality for team Dailies. It should be the same as #11390, except:

- Missing commits (mainly to comply with new eslint checks) are added
- Code related to @phillipthelen's bulk task scoring is removed
- Code related to "manager notes" functionality is removed
- I refactored a couple of places to not import the entire `_` Lodash library
- Nonfunctional code style differences e.g. linebreaks
- New bug fixes:
  - Various places where we were looking for `group` instead of `group.id`, causing functionality to be invoked outside of the teams context
  - Bad path for `requestedDate` causing a `TypeError`

All functional work credits to @MeanderingCode!


SabreCat:
Hmm, I might be running into some bugs with the cron/uncheck flow. Going to run the workflow in a "realistic" scenario over tonight/tomorrow to see if I can reproduce it.

SabreCat:
OK, there are at least these issues, upon a run through the "a day passes" flow, in decreasing order of severity:

1. Completed master Dailies do not reset after users cron. They remain checked off.
2. For some assigned users, a Daily completed the previous day still shows up as incomplete in the Record Yesterday's Activity modal
3. (minor, probably easily fixed) If the master Daily was completed, and is then assigned to a user, it still shows up as needing completion that day

Depending on how frequently **2** happens, that'd be enough to make this functionality a tough sell; and **1** is definitely a can't-go-live level of trouble. @MeanderingCode FYI in case you want to take a look, otherwise I'll see if I can dig into fixing it!

test/api/unit/libs/cron.test.js
```diff
@@ -41,42 +41,42 @@ describe('cron', () => {
     analytics.track.restore();
```
paglias, line 41: should we have tests on the new logic introduced into the cron libs?

website/server/controllers/api-v3/tasks.js
```diff
@@ -751,6 +751,11 @@ api.scoreTask = {
       }
     }
 
+    // Set whether we're scoring a yesterdaily from the RYA dialog
+    if (req.params.yesterdaily && req.params.yesterdaily === 'yesterdaily') {
+      task.yesterDailyScored = true;
```
paglias, line 756: this field doesn't exist on the task schema, also will it require changes on the mobile apps?

paglias: Also I feel like given that it's optional a query parameter seems better than an url parameter. We should also document it in the api route docs

website/server/libs/cron.js
```diff
@@ -434,7 +440,26 @@ export function cron (options = {}) {
...
 
-    task.completed = false;
+    // Check if another user completed shared task in the "same" day the user is "starting"
+    if (task.group && task.group.id && task.group.sharedCompletion === SHARED_COMPLETION.single) {
+      // @REVIEW This introduces an async call into this cron function.
+      // The function is called from ../middlewares/cron.js asyncCron(), which is already async
+      // Does this cause issues?
+      // Pushing these closures to an array of Promises outside the forEach
+      // This allows both sequential processing and async checking for shared completion
+      groupSharedDailies.push(
```
paglias, line 450: I think using a plain promise here would be cleaner:

groupTaskCompleted(task, user, now).then(result => memberTask.completed = result);

website/server/libs/cron.js
```diff
@@ -434,7 +440,26 @@ export function cron (options = {}) {
...
+        (async function determineGroupCompletion (memberTask, memberUser, memberTime) {
+          memberTask.completed = await groupTaskCompleted(memberTask, memberUser, memberTime);
+        }(task, user, now)),
+      );
+    } else {
+      task.completed = false;
+      if (task.group && task.group.id) {
+        groupSharedDailies.push(async () => {
```
paglias, line 458: This can be simplified in `groupSharedDailies.push(groupTaskNewDay(task, user))` since later it's awaited in Promise.all anyway

website/server/libs/cron.js
```diff
@@ -434,7 +440,26 @@ export function cron (options = {}) {
       });
     }
 
-    task.completed = false;
+    // Check if another user completed shared task in the "same" day the user is "starting"
+    if (task.group && task.group.id && task.group.sharedCompletion === SHARED_COMPLETION.single) {
+      // @REVIEW This introduces an async call into this cron function.
```
paglias, line 445: I don't think the async call is an issue per se but cron is already very slow if you have many dailies

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
...
 
 const SHARED_COMPLETION = {
   default: 'recurringCompletion',
   single: 'singleCompletion',
   every: 'allAssignedCompletion',
 };
 
-async function _completeMasterTask (masterTask) {
-  masterTask.completed = true;
+// @TODO Group task scoring
```
paglias, line 18: is this @TODO still necessary?

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
+import defaults from 'lodash/defaults';
+import moment from 'moment';
 import * as Tasks from '../models/task'; // eslint-disable-line import/no-cycle
+import { model as Groups } from '../models/group'; // eslint-disable-line import/no-cycle
+import { model as Users } from '../models/user/index'; // eslint-disable-line import/no-cycle
+import { InternalServerError } from './errors';
+import common from '../../common';
+
+const { sanitizeOptions } = common;
```
paglias, line 9:
```suggestion
const { sanitizeOptions, startOfDay } = common;
```

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
...
+  if (!masterTask || !masterTask.group) return false;
+
+  if (masterTask.history && masterTask.history.length > 0) {
+    // I hope we don't really need to loop the history and check every date
+    // (User with a late day start in Australia vs and early one in Hawaii?)
+    const taskLastHistory = masterTask.history[masterTask.history.length - 1];
+    // Check if the history entry has a user who completed the task
+    // TODO When group scoring is implemented, add history entries for group missed dailies
+    if (taskLastHistory.userId) {
+      const lastCompletingUser = await Users.findById(taskLastHistory.userId, 'preferences');
```
paglias, line 196: .exec()

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
...
+  }).exec();
+
+  if (!masterTask || !masterTask.group) return false;
+
+  if (masterTask.history && masterTask.history.length > 0) {
+    // I hope we don't really need to loop the history and check every date
+    // (User with a late day start in Australia vs and early one in Hawaii?)
+    const taskLastHistory = masterTask.history[masterTask.history.length - 1];
+    // Check if the history entry has a user who completed the task
+    // TODO When group scoring is implemented, add history entries for group missed dailies
```
paglias, line 194: is this TODO necessary? Also I can't find where we're pushing the userId into the history of the task

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
...
+    if (taskLastHistory.userId) {
+      const lastCompletingUser = await Users.findById(taskLastHistory.userId, 'preferences');
+      if (lastCompletingUser) {
+        // Check what the completing user's "day" was when the task was completed
+        const taskLastCompletedDay = startOfDay(
+          defaults(
+            { now: moment(taskLastHistory.date) },
+            lastCompletingUser.preferences.toObject(),
```
paglias, line 202: is toObject necessary? I'm seeing it's called here but for example not in scoreTask

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
...
+  return false;
 }
 
 async function handleSharedCompletion (groupMemberTask) {
   const masterTask = await Tasks.Task.findOne({
     _id: groupMemberTask.group.taskId,
   }).exec();
 
-  if (!masterTask || !masterTask.group || masterTask.type !== 'todo') return;
+  if (!masterTask || !masterTask.group || masterTask.type === 'habit') return;
```
paglias, line 220: should we also exclude rewards since `handleSharedCompletion ` seems to be called for all tasks inclusing group rewards?

website/client/src/components/tasks/column.vue
```diff
@@ -666,9 +666,10 @@ export default {
     filterByLabel (taskList, filter) {
       if (!taskList) return [];
       return taskList.filter(task => {
+        if (filter === 'all') return true;
```
paglias, line 669: is the last filter (return !task.completed;) used somewhere now that we have the `all` one? Is it okay to change the behavior from `!task.completed` to showing all tasks?

website/common/script/ops/scoreTask.js
```diff
@@ -297,24 +298,48 @@ export default function scoreTask (options = {}, req = {}, analytics) {
         task.completed = true;
 
         // Save history entry for daily
+        // Set date to the last minute of the user's "yesterday" for yesterDailies from RYA dialog
+        const dateScored = task.yesterDailyScored ? startOfDay(user.preferences).subtract(1, 'minute') : moment();
         task.history = task.history || [];
         const historyEntry = {
-          date: Number(new Date()),
+          date: Number(new Date(dateScored)),
```
paglias, line 305: can this be considered a breaking change?

test/api/unit/models/group_tasks.test.js
```diff
@@ -233,10 +235,60 @@ describe('Group Task Methods', () => {
           expect(syncedTask.checklist.length).to.equal(0);
           expect(syncedMemberTask.checklist.length).to.equal(0);
         });
+
+        it('removes and reassigns completed and unchecked todos to other assignees', async () => {
+          if (taskType !== 'todo') return;
+
+          const updatedLeader = await User.findOne({ _id: leader._id });
```
paglias, line 242: .exec()

test/api/unit/models/group_tasks.test.js
```diff
@@ -233,10 +235,60 @@ describe('Group Task Methods', () => {
...
+          const updatedLeadersTasks = await Tasks.Task.find({ _id: { $in: updatedLeader.tasksOrder[`${taskType}s`] } });
```
paglias, line 243: .exec()

test/api/unit/models/group_tasks.test.js
paglias: or can we use leader.sync()?

test/api/unit/models/group_tasks.test.js
```diff
@@ -233,10 +235,60 @@ describe('Group Task Methods', () => {
...
+          const syncedTask = find(updatedLeadersTasks, findLinkedTask);
+
+          syncedTask.completed = true;
+          await handleSharedCompletion(syncedTask);
+
+          const updatedMember = await User.findOne({ _id: newMember._id });
```
paglias, line 249: .exec() or newMember.sync()

test/api/unit/models/group_tasks.test.js
```diff
@@ -233,10 +235,60 @@ describe('Group Task Methods', () => {
...
+          const updatedMemberTasks = await Tasks.Task.find({ _id: { $in: updatedMember.tasksOrder[`${taskType}s`] } });
```
paglias, line 250: .exec()

test/api/unit/models/group_tasks.test.js
```diff
@@ -233,10 +235,60 @@ describe('Group Task Methods', () => {
...
+          const syncedMemberTask = find(updatedMemberTasks, findLinkedTask);
+
+          syncedTask.completed = false;
+          await handleSharedCompletion(syncedTask);
+
+          const reUpdatedMember = await User.findOne({ _id: newMember._id });
```
paglias, line 256: .exec() or newMember.sync()

test/api/unit/models/group_tasks.test.js
```diff
@@ -233,10 +235,60 @@ describe('Group Task Methods', () => {
...
+          const reUpdatedMemberTasks = await Tasks.Task.find({ _id: { $in: reUpdatedMember.tasksOrder[`${taskType}s`] } });
```
paglias, line 257: .exec()

paglias: actually is it necessary to load all tasks since we always seem to only need one? can we search by id?

test/api/unit/models/group_tasks.test.js
```diff
@@ -233,10 +235,60 @@ describe('Group Task Methods', () => {
...
+          const reSyncedMemberTask = find(reUpdatedMemberTasks, findLinkedTask);
+
+          expect(syncedMemberTask).to.not.exist;
+          expect(reSyncedMemberTask).to.exist;
+        });
+
+        it('syncs daily completion to other assignees', async () => {
```
paglias, line 264: a few issues of missing .exec() or possibility to use .sync() here as well

test/api/unit/models/group_tasks.test.js
```diff
@@ -233,10 +235,60 @@ describe('Group Task Methods', () => {
...
+          if (taskType !== 'daily') return;
+
+          const updatedLeader = await User.findOne({ _id: leader._id });
+          const updatedLeadersTasks = await Tasks.Task.find({ _id: { $in: updatedLeader.tasksOrder[`${taskType}s`] } });
```
paglias, line 268: need all tasks?

website/server/models/group.js
```diff
@@ -1597,7 +1593,16 @@ schema.methods.syncTask = async function groupSyncTask (taskToSync, user) {
   // add tag if missing
   if (matchingTask.tags.indexOf(group._id) === -1) matchingTask.tags.push(group._id);
 
-  toSave.push(matchingTask.save(), taskToSync.save(), user.save());
+  toSave.push(matchingTask.save(), user.save());
+  return Promise.all(toSave);
+};
+
+schema.methods.linkTask = async function groupLinkTask (linkingTask, user) {
```
paglias, line 1600: we should start to split up this file, it's really huge, might be a good occasion to start by moving these tasks related methods

website/server/models/group.js
```diff
@@ -1597,7 +1593,16 @@ schema.methods.syncTask = async function groupSyncTask (taskToSync, user) {
   // add tag if missing
   if (matchingTask.tags.indexOf(group._id) === -1) matchingTask.tags.push(group._id);
 
-  toSave.push(matchingTask.save(), taskToSync.save(), user.save());
+  toSave.push(matchingTask.save(), user.save());
```
paglias, line 1596: we're changing `taskToSync` but not saving it anymore?

website/server/models/group.js
paglias: What's the difference between link and sync task? Can we add some comments explaining how they work and how they're supposed to be used?

website/server/models/group.js
```diff
@@ -1597,7 +1593,16 @@ schema.methods.syncTask = async function groupSyncTask (taskToSync, user) {
...
+  return Promise.all(toSave);
+};
+
+schema.methods.linkTask = async function groupLinkTask (linkingTask, user) {
+  const toSave = [];
+  if (linkingTask.group.assignedUsers.indexOf(user._id) === -1) {
+    linkingTask.group.assignedUsers.push(user._id);
+  }
+  toSave.push(linkingTask.save(), this.syncTask(linkingTask, user));
```
paglias, line 1605: syncTask changes `linkingTask` but doesn't save it, it's saved while `syncTask ` runs so it might not catch the changes made there?

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
...
+    completed: true,
+    type: 'daily',
+  }).exec();
+
+  if (!masterTask) return;
+
+  if (masterTask.history && masterTask.history.length > 0) {
+    const taskLastHistory = masterTask.history[masterTask.history.length - 1];
+    if (taskLastHistory.userId) {
+      const lastCompletingUser = await Users.findById(taskLastHistory.userId, 'preferences');
```
paglias, line 247: .exec()

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
...
+    _id: groupMemberTask.group.taskId,
+    completed: true,
+    type: 'daily',
+  }).exec();
+
+  if (!masterTask) return;
+
+  if (masterTask.history && masterTask.history.length > 0) {
+    const taskLastHistory = masterTask.history[masterTask.history.length - 1];
+    if (taskLastHistory.userId) {
```
paglias, line 246: the logic here between `groupTaskNewDay ` and `groupTaskCompleted ` is very similar, maybe we can combine the two or share part of the code?

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
...
+      const lastCompletingUser = await Users.findById(taskLastHistory.userId, 'preferences');
+      if (lastCompletingUser) {
+        const taskLastCompletedDay = startOfDay(defaults(
+          { now: moment(taskLastHistory.date) },
+          lastCompletingUser.preferences.toObject(),
+        ));
+        const userDay = startOfDay(defaults({ now }, user.preferences.toObject()));
+        if (!userDay.isSame(taskLastCompletedDay)) {
+          masterTask.completed = false;
+          masterTask.save();
```
paglias, line 256: await?

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
...
+      }).exec();
+    } else {
+      // The task was uncompleted by the group member and should be recreated for assignedUsers
+      const group = await Groups.findById(masterTask.group.id).exec();
+      const userList = [];
+      masterTask.group.assignedUsers.forEach(userId => {
+        const query = { _id: userId };
+        userList.push(query);
+      });
+      const assignedUsers = await Users.find({
```
paglias, line 46: .exec()

paglias: also is it necessary to load the entire user objects?

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
...
+        $or: userList,
```
paglias, line 47: while $or? Not `{_id: {$in: usersIdsList}}`?

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
...
+          _id: { $in: masterTask.group.assignedUsers },
+        },
+        {
+          _id: { $ne: groupMemberTask.userId },
+        },
+      ],
+    };
+    const groupMemberQuery = groupMemberTask.userId;
+    const [tasks, assignedUsers, groupMember] = await Promise.all([
+      Tasks.Task.find(tasksQuery),
```
paglias, line 78: .exec()

website/server/libs/groupTasks.js
```diff
@@ -1,61 +1,268 @@
+import defaults from 'lodash/defaults';
+import moment from 'moment';
 import * as Tasks from '../models/task'; // eslint-disable-line import/no-cycle
+import { model as Groups } from '../models/group'; // eslint-disable-line import/no-cycle
+import { model as Users } from '../models/user/index'; // eslint-disable-line import/no-cycle
```
paglias, line 5: why Users and not User?

paglias (commented):
I'm a bit concerned about:

- all those async calls with multiple database queries in cron, I think the code is so complex we are losing track of what's being loaded and saved and how many times
- the logic is complex
- could be considered a breaking change since it changes the tasks history behavior
- the apps will require updates

MeanderingCode:
Thanks for taking the time to rework this onto the tree again, @SabreCat!

> * Completed master Dailies do not reset after users cron. They remain checked off.

I did not have that experience. I will test again on this reworked branch.

> * For some assigned users, a Daily completed the previous day still shows up as incomplete in the Record Yesterday's Activity modal

Can you be more specific about repro to help me look into this?

> * (minor, probably easily fixed) If the master Daily was completed, and is then assigned to a user, it still shows up as needing completion that day

Interesting. Must be related to the way the User tasks are "cloned" from the Group task. Should be simple. Will have a look.

I will take a look at these when I can. Thanks for the work and the testing!

MeanderingCode:
@paglias  

> I'm a bit concerned about:
>
>    * all those async calls with multiple database queries in cron, I think the code is so complex we are losing track of what's being loaded and saved and how many times

I definitely have concerns about this. At the time, I didn't see a simple way to avoid it. Maybe another look would present a way. Any advice?

>    * the logic is complex

The core logic is fairly clean, but delicate enough that it deserves several people thinking it over. What specific concerns do you have in which specific areas?

The cron logic, taking into account the possibility of different "day" overlaps seemed the most convoluted part to me.

>    * could be considered a breaking change since it changes the tasks history behavior

As far as the core functionality goes, it does not seem to be breaking. The history behavior for task completion is unchanged on the user side. Some Dailies may be marked as complete while no history entry exists for that user, but that is incidental and does not effect any internal functionality I could see.

Adding history to the Group task is non-breaking, as nothing assumes or uses history on Group tasks since it is unimplemented.

>    * the apps will require updates

What changes? I didn't think it did. I haven't looked at the mobile code, but at least on the web there is little need for difference. The mobile apps already do not support management of Group Tasks, so it is only the User views, which don't need to behave differently.

I will take a look through your line comments tomorrow, though I am not sure when I will have time to start retesting on this branch and addressing anything.

paglias:
Regarding the mobile apps it looks like tasks scored in the RYA modal should be scored passing a new parameter to the URL which is what I meant by needing an update

SabreCat:
Closing this for now. We may want to take another stab at this functionality as we continue to enhance Teams in the back half of 2020, but this specific implementation didn't work out.

kendonB:
@SabreCat I have recently adopted groups and find it's a great way to manage tasks for our family. I am sad to see that this feature got so close and was never completed. Is there a continuation of this work somewhere else?

CuriousMagpie:
Hi @kendonB! We are in the process of implementing some new Teams functionality right now--drop us a line at admin@habitica.com with your username and your group's name and we'll get your group access to the beta server.

enjibby:
Hello there! Shared completion sounds like a feature that I'd definitely love to make use of with my family group plan. Is there still progress being made on this?

PR #12225, opened 2020-05-21
fix(chat) - graceful for failure and fix link regex

### LINKED ISSUE #12223: the same URL cannot be posted twice in a markdown link
If you try to post the same URL twice in the same message (example below), it fails.
```
[here](http://habitica.wikia.com/wiki/The_Keep:Pirate_Cove/FAQ) 
[hier](http://habitica.wikia.com/wiki/The_Keep:Pirate_Cove/FAQ)
```

The error also occurs if there is text between the two links. Posting the same URL (with different text) is essential for multi-lingual posts.

The player sees "An unexpected error occurred" and the server logs show:
```
TypeError: Cannot read property 'index' of null
    at /usr/src/habitrpg/website/transpiled-babel/libs/highlightMentions.js:132:15
    at Array.forEach (<anonymous>)
    at findTextBlocks (/usr/src/habitrpg/website/transpiled-babel/libs/highlightMentions.js:128:22)
    at highlightMentions (/usr/src/habitrpg/website/transpiled-babel/libs/highlightMentions.js:165:22)
    at handler (/usr/src/habitrpg/website/transpiled-babel/controllers/api-v3/chat.js:246:88)
```

The "Cannot read property 'index' of null" message also occurred in this issue: https://github.com/HabitRPG/habitica/issues/12217 which is now fixed but I've checked on our staging site and this new bug still occurs there.

The desired fix for this issue is to allow the same URL to be posted twice (or three or more times) but if possible it would also be good to have something that checks for whether the "Cannot read property 'index' of null" error would be thrown and if so, don't execute the code that would throw it. If that means that a more minor error occurs (e.g., formatting problems with the link), that might be acceptable, at least as a temporary workaround.

I'm marking this as high priority because the ability to post the same link with different text is very helpful for guilds where two or more languages are spoken. If the best solution for this would take some time but a smaller partial fix could be done quickly, it would be acceptable in my opinion for us to have the quick fix in place rapidly, with a second PR done later.

benkelaar:
Fixes #12223 

### Changes
 - Fix overeager dot matcher in link matcher that causes failure in double link
 - Make source-map regexes resilient to unexpected failure

----

### DIFF
--- a/website/server/libs/highlightMentions.js
+++ b/website/server/libs/highlightMentions.js
@@ -86,7 +86,7 @@ function toSourceMapRegex (token) {
   } else if (type === 'link_open') {
     const texts = token.textContents.map(escapeRegExp);
     regexStr = markup === 'linkify' || markup === 'autolink' ? texts[0]
-      : `\\[.*${texts.join('.*')}.*\\]\\([^)]+\\)`;
+      : `\\[[^\\]]*${texts.join('[^\\]]*')}[^\\]]*\\]\\([^)]+\\)`;
   } else {
     throw new Error(`No source mapping regex defined for ignore blocks of type ${type}`);
   }
@@ -110,6 +110,11 @@ function findTextBlocks (text) {
   ignoreBlockRegexes.forEach(regex => {
     const targetText = text.substr(index);
     const match = targetText.match(regex);
+
+    if (!match) {
+      // Should not happen, but insert to handle bugs gracefully
+      return;
+    }
 
     if (match.index) {
       blocks.push({ text: targetText.substr(0, match.index), ignore: false });
--- a/test/api/unit/libs/highlightMentions.test.js
+++ b/test/api/unit/libs/highlightMentions.test.js
@@ -150,12 +150,12 @@ describe('highlightMentions', () => {
       expect(result[0]).to.equal('[@user](/profile/111) `@user`');
     });
 
-    it('does not crash when a link is repeated in the message.', async () => {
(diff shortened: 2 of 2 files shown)


website/server/libs/highlightMentions.js
```diff
@@ -111,6 +111,11 @@ function findTextBlocks (text) {
     const targetText = text.substr(index);
     const match = targetText.match(regex);
 
+    if (!match) {
+      // Should not happen, but insert to handle bugs gracefully
+      return;
```
benkelaar, line 116: Ideally I would like to log an error here. If `console#log` is not the way to do that (Based on that there's a linter rule against it) what would be the best way to do that?

paglias: You can use the logger interface like here https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/pushNotifications.js#L149. It will log to the console in development and to an external service in production. Ideally the first argument should be an `Error` object while the second an object with extra optional data

benkelaar: Awesome. Exactly what I needed. I don't expect to be able to hook up my laptop to the internet any time soon so perhaps merge this and I'll do the proper logging in a followup PR?

paglias: Agreed, this way it'll go live with a deployment later today

Alys:
Thanks so much for the fast fix and fast review @benkelaar @paglias !

paglias:
noted in your profile

PR #12227, opened 2020-05-22
Blockquote styling (fixes #9476)

### LINKED ISSUE #9476: Website Markdown does not currently support quoting with > and >> (blockquotes)
Website Markdown does not currently support quoting with > and >>. It should be updated so that it supports this again.

citrusella:

Fixes #9476 

### Changes

~~Resisting urge to just scream "It's block quotes! They're pretty! Aaah!"~~

Applies styling to blockquotes based on the design on Zeplin ([image and link of Zeplin here](https://github.com/HabitRPG/habitica/issues/9476#issuecomment-463671342)).

I've (maybe redundantly) put the same CSS in typography.scss *and* markdown.scss, mostly because I thought it'd be easier in the future if for some reason different styling might want to be done for places with the markdown CSS class (like tasks, and challenge/guild descriptions/summaries) as opposed to the overall style. (Chat messages and private messages don't use the markdown class.)

Before (taken on production):

Blockquoted lorem ipsum but it just looks like plain text

After:

Example of no blockquote, a blockquote alone, and some nested blockquotes with the same lorem ipsum text

I've showed off the nesting and such because for a second when I had just started playing around with the styling, I thought the spacing looked odd (but I think it looks fine now?). There's nothing too unexpected or anything about the way it looks in guild/challenge/etc. since I forgot to screenshot those...

----


website/client/src/assets/scss/markdown.scss
```diff
@@ -41,4 +41,9 @@
   img {
     max-width: 100%;
   }
+  
+  blockquote {
+    padding: 0 16px;
```
benkelaar, line 46:
```suggestion
    padding-left: 16px;
```

Or was the additional `padding-right` also by design?

website/client/src/assets/scss/typography.scss
```diff
@@ -38,6 +38,11 @@ a, a:not([href]):not([tabindex]) {
   }
 }
 
+blockquote {
+  padding: 0 16px;
```
benkelaar, line 42: Same as above.

benkelaar (commented):
Looks good! Small question on the padding.

website/client/src/assets/scss/markdown.scss
citrusella: I tried playing with using different padding for left and right for a little bit just before I settled on this, but I felt like it looked "wrong" on my local install in regards to where the words wrapped, when compared to what I was looking at on Zeplin. (I tried out 16 padding on the left, alone, as well as 16 left and a smaller one on the right, based on what I thought the numbers on Zeplin were suggesting.)

Basically, I settled on what I used because it appeared to be most faithful to the example in terms of appearance. ._.

(Sidenote: It's also incredibly similar to the pre-2017-redesign padding (which was the same padding code but 15px), though that doesn't really play into why I went with it...)

benkelaar: Cool. Aren't you going to run in to the issue that you're basically running out of space for words if you nest enough blockquotes? Or is there a limit to the amount of nesting for blockquotes?

Alys: We can probably rely on common sense to avoid that most of the time. I'd suggest ignoring the problem for the sake of simpler code (unless it becomes obvious that it happens often enough for us to care).

benkelaar: Sounds good.

citrusella: I know the conversation has more or less reached a resolution (impasse? undetstanding? IDK?) but: Staring at my screenshot right now, I think removing right padding would only allow one extra level of quote if extreme nesting over the current code anyway, based on where the words wrap. OTOH, I might also be tired with little time (gotta get some house cleaning done) to fire up my installation and *confirm* that and thereby am working entirely off the way my tired mind is parsing the visual alone.

website/client/src/assets/scss/typography.scss
```diff
@@ -38,6 +38,11 @@ a, a:not([href]):not([tabindex]) {
   }
 }
 
+blockquote {
```
paglias: thanks for adding it here as well but given that right now we don't have any blockquote element and this risks getting out of sync with the markdown css, could you remove it?

paglias (changes requested):

website/client/src/assets/scss/typography.scss
citrusella: ...I'm confused. The blockquote code added here is the only thing that makes the blockquote element have styling in chat posts, so removing it will make blockquotes work on tasks, guild/challenge sidebars, etc (since they use the `.markdown` class), but be unstyled/look like plain text in chat (since it doesn't use the `.markdown` class). I didn't want to start trying to add CSS classes to chat posts without knowing if it was desired (particularly since there are a couple things that are in both files but different between them), so my original changes were to the typography.scss since the Zeplin example was of styling on a chat post. Is there a different file that should be changed instead to get that styling in chat posts that I just didn't think of?

(At any rate, I may not be able to make any changes *today* specifically since it's my birthday and I'm busy enough with birthday stuff somehow that I've been interrupted five times just trying to write this reply! 😲 But I wanted to reply right now because of my confusion, instead of putting it off and potentially forgetting.)

paglias: Hmm, sorry I got confused, I didn't realize the chat posts don't have the `markdown` class applied, I thought it was added to typography.css only as a possible future use...

@Tressley is it the expected behavior that chat posts don't get the markdown specific css styling? or maybe it's a bug we didn't notice

citrusella: For what it's worth, if I remove it from markdown.scss it does still work everywhere--that was the one I added for possible future use/differentiation.

website/client/src/assets/scss/markdown.scss
Tressley: Sorry I'm a little late to the conversation, the original issue mentioned that we wouldn't be allowing nesting for the first pass. Thanks for Citrus' design eye, I thought nesting looked good and don't mind supporting it. This looks good. 👍

website/client/src/assets/scss/typography.scss
Tressley: We pretty much want Markdown supported anywhere a user has a text input. There are some exceptions, e.g. We want to limit what Markdown is supported in the task title and checklist items to prevent the task layout from breaking.

Additionally, chats/tasks have a slightly different type scale for Markdown to prevent the task layout from breaking. Can I see an example of a task with nested quotes?

website/client/src/assets/scss/markdown.scss
citrusella: To be fair regarding my "design eye", I haven't included any nesting-*specific* code (the markdown already supported nesting the element), I just remembered halfway through testing it out that nesting is a thing that can usually be done so I just tested it to be sure it a) worked the way I thought it did (i.e. that the markdown on Habitica allowed for it), and b) wasn't a big ugly mess, heh.

In all honesty, every time I code anything I'm really just pleasantly surprised that it doesn't explode?

website/client/src/assets/scss/typography.scss
paglias: Sorry what I meant is that the chat messages (not sure if it's a recent change or has always been the case) don't get the `markdown` CSS class applied and so they don't get the markdown specific css styles we have everywhere else. The css applied to the markdown everywhere except for the chat message is here https://github.com/HabitRPG/habitica/blob/develop/website/client/src/assets/scss/markdown.scss

citrusella: I went ahead and fired up my local install to get images of tasks (and guild/challenge descriptions... and profiles, which I forgot could hold text for a hot second) anyway. (All three use the `markdown` class, for reference.)

citrusella:
(@Tressley since they asked for some of these (but not all of them, I just thought I'd be thorough this time with my screenshots))

Additional screenshots for reference, of things that aren't the chat window that render blockquotes (these things all have the markdown class, and the chat window does not, though since at this point the CSS is identical for blockquotes that functionally means less than it could):

Tasks manually created by me on the task page, nested and non-nested (this made me realize I actually had forgotten to test task appearance and just thought I had done so...):

a task with double blockquotes followed by a task with single blockquotes

Additionally, a task with line breaks/paragraphs in the title (allowing a more likely (in my opinion) nesting situation to occur), achieved by copying as to-do from a message posted in my Tavern:

task title with nested blockquotes copied from the Tavern

Guild description (and the summary, didn't think I'd gotten that in there) with challenge box visible (guild summaries as well as the challenge sidebar function identically to the guild description in terms of appearance):

guild description with nested and non-nested blockquotes, with a challenge with both of those visible at the bottom

My computer's color settings must have been different in some slight way when I initially tested on descriptions because I *really* don't remember a striking lack of contrast where the sidebar is concerned. (I would have brought it up if I *did* see it.)

Profiles:

nested and non-nested blockquotes in profile

Tressley:
I think this looks good! The blockquote in the task title pains me, but I think we'll have to rely on user's not wanting their tasks to breakdown. 😄

paglias:
Sorry for the delay here @citrusella , after speaking with @Tressley the desired behavior is that all places where markdown is used should get the `.markdown` class meaning that the blockquote style can be there only. 

There are a couple of places where the `markdown` directive isn't used and for this reason the class isn't applied and must be applied manually:

- https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/chat/chatCard.vue#L37
- https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/messages/messageCard.vue#L24

citrusella:
...I was *just about* to ask if there were any updates or things I needed to do. o.o Will take a look and see what I can get done after I get some other tasks I've been in the middle of done.

citrusella:
Added the classes and removed the code from typography.scss, so the blockquotes are still preserved everywhere... but adding the class changed the spacing at the very end of chat posts (but not between paragraphs). Is that okay or do I have another edit to make?

Current spacing without my change (note the space between the end of the post and the horizontal line):

production spacing (before markdown class)

Spacing after my change:

local spacing (after markdown class application)

Message spacing currently:

local spacing messages (after markdown class application)

And after my change:

production spacing messages (before markdown class)

paglias:
@citrusella looks like removing these 2 negative margins [here](https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/chat/chatCard.vue#L169) and [here](https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/messages/messageCard.vue#L123) will fix the margin issue

citrusella:
I noticed that was causing it on the chat posts but for some reason hadn't seen the same CSS on the message card, so thanks for the link; I've gotten that done!

paglias:
thanks @citrusella !

PR #12228, opened 2020-05-24
Fix empty link bug and proper error logging in highlightMentions.js

benkelaar:
…source-mapping failures

Follow-up on #12225 fix for #12223 

### Changes
- Log error when source mapping fails for a message (But still don't fail posting the message)
- Properly support links without URLs
- Add list of known unsupported markdown cornercases

----

### DIFF
--- a/website/server/libs/highlightMentions.js
+++ b/website/server/libs/highlightMentions.js
@@ -71,8 +71,17 @@ function withOptionalIndentation (content) {
   return content.split('\n').map(line => `\\s*${line}`).join('\n');
 }
 
-// This is essentially a workaround around the fact that markdown-it doesn't
-// provide sourcemap functionality and is the most brittle part of this code.
+/* This is essentially a workaround around the fact that markdown-it doesn't
+ * provide sourcemap functionality and is the most brittle part of this code.
+ *
+ * Known errors (Not supported markdown link variants):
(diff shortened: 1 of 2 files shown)


website/server/libs/highlightMentions.js
```diff
@@ -112,7 +122,10 @@ function findTextBlocks (text) {
     const match = targetText.match(regex);
 
     if (!match) {
-      // Should not happen, but insert to handle bugs gracefully
+      logger.error(
+        new Error('Failed to match source-mapping regex to find ignore block'),
+        { text, targetText, regex },
```
paglias: this looks good but can we convert the regex to a string (String(regex) should work), we've had some issues in the past with the logging not working well with uncommon data types and anyway it'll be converted to JSON. This way we'll be sure we're getting a string logged

paglias (changes requested):

website/server/libs/highlightMentions.js
benkelaar: Ah yeah, it looked fine locally with the Console transport configured but I don't know how it behaves with the production transports of course. Will fix 👍

benkelaar: Fixed in fbcb25d8288d1e83d09c53a99a0c7fee20c0f70a

paglias:
thanks @benkelaar and welcome to the 4th tier!

Issue #11051, opened 2019-03-07
Profile pages show previous player temporarily - loading is slow


Alys:
When you click on a player's avatar or name to view their profile, as soon as the profile modal appears, you see the details of the previous profile that you viewed. It's visible for long enough to be confusing (it looks like a bug or that you've clicked on the wrong person or that the person has dramatically changed their name and avatar since the time they posted a message). After a second or so (usually) the wrong profile is replaced by the correct one but the delay can be longer.

I think it would be less confusing if the profile modal initially appeared blank rather than showing cached data from the previous person, or if it didn't appear at all until the new player's data was ready to be displayed in it.

I'm leaving this as suggestion-discussion for the staff to comment on.

---

In addition to that, FolleMente / @OuttaMyMind (270f8744-3c6f-48d3-9bcf-e9dc79104168) has reported that sometimes the correct profile never appears:

"When I try to click on another persons profile, I have to reset the page in order to see their profile. I see the first person I originally clicked on pull up..."
"... As for the loading, it depends. Sometimes it'll load after a few seconds, but other times, it won't load at all. I usually keep it up for a few minutes."

I am guessing (although not with any evidence) that sometimes the new player's details weren't loading at all because of the recent server problems, and so I suggest that the slowness issue described above be fixed first, and then if we still have reports of the correct player's details never appearing, we can investigate further.

godless-toast:
> When you click on a player's avatar or name to view their profile, as soon as the profile modal appears, you see the details of the previous profile that you viewed.

Yes, I've experienced this as well.

But another thing I experience is viewing a profile, and then I go to another page, say tasks. I open and start editing a task, then when I save it, sometimes that profile shows up again.

Then I might look at someone else's profile. Close it. Old one shows up again. It will keep showing up here and there, but refreshing fixes it. This is on desktop in Firefox.

My problem doesn't sound like this issue or #11066. Is there another issue for this?

Alys:
@godless-toast I'm sorry we didn't reply to you before now. The issue you described with a profile reopening is https://github.com/HabitRPG/habitica/issues/11083

SabreCat:
Do we have a good way of reproducing this? When I open profiles one after another, there is a bit of delay, but during that delay there's only a tiny strip of the modal visible, not the whole profile. Is it browser-specific?

I'll still flag this as worthy of help (in case we can just do a cache purge irrespective of visible issues), but it'd help contributors to know how to get this to happen!

[...]

PR #12229, opened 2020-05-24
Make message links to deleted profiles display 404: (fixes 11020)

### LINKED ISSUE #11020: profile URL for non-existent / deleted account should show an error message
If you click on a profile URL for an account that does not exist (e.g., when someone has deleted their account), you see just whitespace where the body of the page should be - i.e., there's no message to indicate why loading the profile failed. You can see an example of this at https://habitica.com/profile/12345678-b5b9-4bb4-8b82-123456789abc which uses a User ID that isn't associated with any account.

There should be a message to say that the account does not exist. 

I suggest using the existing string `messageDeletedUser` ("Sorry, this user has deleted their account.") which is what you see when you click on a name in chat after the associated account has been deleted. If we decide that a different message would be appropriate, we might want to consider modifying `messageDeletedUser` instead of creating a new string to keep the message consistent in both places where it can appear.

@Tressley I'm marking this as needs design

_Thanks to citrusella for reporting this._

jacobguinther:

Fixes #11020

### Changes

Changes in chatMessages:
  - Move Red 'Deleted Account' Notification Trigger
      - From chatMessages -> profile
  - Other changes in chatMessages affect avatar link

Changes in profile:
  - If user doesnt exist:
    - Render 404
    - Display Red 'Deleted Account' Notification

Now the 'Deleted Account' notification fires on deleted user profile page as well as modals.

Clicking Avatar or Username
Screen Shot 2020-05-24 at 11 18 18 AM

Displays 404 modal with notification
Screen Shot 2020-05-24 at 11 20 14 AM

Deleted Users Profile page now looks like this:
Screen Shot 2020-05-24 at 11 20 35 AM
( displays same notification as well (not shown in picture))

----
UUID: 234319eb-3c97-4cf7-9248-9feff04a31b6


website/client/src/components/userMenu/profile.vue
```diff
@@ -836,8 +841,23 @@ export default {
...
+        try {
+          const response = await this.$store.dispatch('members:fetchMember', {
+            memberId: profileUserId,
+          });
+          // debugger;
+          user = response.data.data;
+        } catch (e) {
+          // user = null;
+          user.active = false;
+          console.log('fkdsafj;dsak;', user.active);
```
paglias: this should be removed, as well as the other comments like // debugger, //user = null

website/client/src/components/userMenu/profile.vue
```diff
@@ -826,7 +831,7 @@ export default {
   },
   methods: {
     async loadUser () {
-      let user = this.userLoggedIn;
+      let user = await this.userLoggedIn;
```
paglias: why this await? this.userLoggedIn isn't a promise as far as I know

website/client/src/components/userMenu/profile.vue
```diff
@@ -733,6 +737,7 @@ export default {
   components: {
     MemberDetails,
     profileStats,
+    error,
```
paglias: can this be renamed to something clearer? like 404Error?

website/client/src/components/userMenu/profile.vue
```diff
@@ -836,8 +841,23 @@ export default {
...
-        user = response.data.data;
+        try {
+          const response = await this.$store.dispatch('members:fetchMember', {
+            memberId: profileUserId,
+          });
+          // debugger;
+          user = response.data.data;
+        } catch (e) {
+          // user = null;
+          user.active = false;
```
paglias: In case the user cannot be fetched, why not keep `this.user` undefined and in the component above do `v-if="!user"` instead of adding a new property to the user object?

website/client/src/components/userMenu/profile.vue
```diff
@@ -836,8 +841,23 @@ export default {
...
       if (profileUserId && profileUserId !== this.userLoggedIn._id) {
-        const response = await this.$store.dispatch('members:fetchMember', { memberId: profileUserId });
-        user = response.data.data;
+        try {
+          const response = await this.$store.dispatch('members:fetchMember', {
+            memberId: profileUserId,
+          });
+          // debugger;
+          user = response.data.data;
+        } catch (e) {
```
paglias: This will catch all errors even if they're not 404s

paglias (changes requested):
left some comments!

website/client/src/components/userMenu/profile.vue
jacobguinther: I agree with you. I don't think I should be adding a new property to the user. However, when I try and do it your suggested way.. I keep getting this bug where the 404 will flash on the modal for every active user before their actual profile displays. Adding the new property is the only way I could get it to work without flashing 404 for active users.

I am still looking into this.

Alys: That seems related to this issue: https://github.com/HabitRPG/habitica/issues/11051
(I.e., it's known that profiles can be slow to load and if caching of a previous profile hasn't happened, you see a blank profile screen before the data loads). That issue is tolerable - we rarely see complaints about it, but if the default state was the 404 error, I'm sure that would cause confusion and more bug reports.

So, I'm a big fan of not flashing up the 404 message initially. :)
If it tunrs out to not be possible to find a way to avoid that without adding a new properly, perhaps we could have the new property as a temporary workaround and then adjust the code later once https://github.com/HabitRPG/habitica/issues/11051 is fixed?

jacobguinther: Sounds good to me. Im having a hard time getting it to work correctly without adding the property. @paglias Any thoughts?

paglias: @jacobguinther can you post the version without the property? I'll give a look and see if I can get it to work

jacobguinther: eh. Found another bug Im working on. 
logged_in_user_404_bug
Currently. If I refresh the page the  logged in users profile modal shows up fine. However if i click any deleted user.. Now the logged in users profile shows 404 also.

I will post version without property once I fix this.

jacobguinther: Ok. I fixed the above bug and got it to work without adding the extra property on user object.

website/client/src/components/userMenu/profile.vue
```diff
@@ -766,7 +771,7 @@ export default {
       achievements: {},
       achievementsCategories: {}, // number, open
       content: Content,
-      user: undefined,
+      user: {},
```
paglias: why not keep undefined? This way `v-if="!user"` will never evaluate to true because `{}` is considered a truthy value

jacobguinther: This was to keep 404 from flashing on modal. I set user to undefined only if 'fetchMember' returns a 404.

If it wasn't for flashing 404 bug I wouldn't have done this.

jacobguinther: Would you like me to try and find another way to fix the flash 404 bug without changing the initial state of user?

paglias: sorry for the delay, yes if you could try another way it would be great but if it's not possible let me know

website/client/src/components/chat/chatMessages.vue
```diff
@@ -331,6 +326,9 @@ export default {
       if (profile && !profile.rejected) {
```
paglias: here (lines 326-331) basically the same thing happens in both cases, if profile is rejected or not because `profile._id` and `memberId` should be the same value

website/client/src/components/chat/chatMessages.vue
```diff
@@ -315,12 +315,7 @@ export default {
       if (!profile._id) {
```
paglias: This is probably unrelated to this PR, but since we're here, we're fetching the profile data (cached or not) and then redirecting to the modal page, so the fetched was not used at all, I guess this whole function could be simplified into 

```
async showMemberModal (memberId) {	
    this.$router.push({ name: 'userProfile', params: { userId: memberId } });
}
```

paglias (changes requested):

website/client/src/components/userMenu/profile.vue
jacobguinther: Im having a hard time finding a way around this. Any suggestions?

website/client/src/components/userMenu/profile.vue
```diff
@@ -1,8 +1,11 @@
 <template>
   <div
-    v-if="user"
+    v-if="!user"
     class="profile"
   >
+    <error404 />
+  </div>
+  <div v-else>
```
paglias: this is missing the `profile` class

paglias (changes requested):
Sorry for the delay @jacobguinther but with the following changes all issues seems to be solved:

```diff
diff --git a/website/client/src/components/userMenu/profile.vue b/website/client/src/components/userMenu/profile.vue
index 9a7dc4eb6b..621b7f960b 100644
--- a/website/client/src/components/userMenu/profile.vue
+++ b/website/client/src/components/userMenu/profile.vue
@@ -1,11 +1,10 @@
 <template>
   <div
-    v-if="!user"
-    class="profile"
+    v-if="!user && userLoaded"
   >
     <error404 />
   </div>
-  <div v-else>
+  <div class="profile" v-else-if="userLoaded">
     <div class="header">
       <span
         class="close-icon svg-icon inline icon-10"
@@ -772,7 +771,8 @@ export default {
       achievements: {},
       achievementsCategories: {}, // number, open
       content: Content,
-      user: {},
+      user: null,
+      userLoaded: false,
     };
   },
   computed: {
@@ -832,11 +832,12 @@ export default {
   },
   methods: {
     async loadUser () {
-      let user = {};
+      let user = null;
 
       // Reset editing when user is changed. Move to watch or is this good?
       this.editing = false;
       this.hero = {};
+      this.userLoaded = false;
       this.adminToolsLoaded = false;
 
       const profileUserId = this.userId;
@@ -844,7 +845,7 @@ export default {
       if (profileUserId && profileUserId !== this.userLoggedIn._id) {
         const response = await this.$store.dispatch('members:fetchMember', { memberId: profileUserId });
         if (response.response && response.response.status === 404) {
-          user = undefined;
+          user = null;
           this.$store.dispatch('snackbars:add', {
             title: 'Habitica',
             text: this.$t('messageDeletedUser'),
@@ -880,6 +881,8 @@ export default {
       } else {
         this.user = this.userLoggedIn;
       }
+
+      this.userLoaded = true;
     },
     selectPage (page) {
       this.selectedPage = page || 'profile';
```

paglias:
thanks @jacobguinther ! Noted towards your 2nd tier, sorry it took so long to get this merged

PR #12247, opened 2020-05-25
Pinned Off-season items should be unavailable to purchase - fixes #9933

### LINKED ISSUE #9933: pinned off-season Magic Hatching Potions, Greeting Cards, Quest Bundles should not appear to be purchasable
If you pin a Magic Hatching Potion to your Rewards list, when that potion is no longer available for purchase, the pinned item acts as if you can still purchase it. You can click on it, select the number to be bought, and then confirm the purchase. You see a message saying that you've bought the Potion(s).

However the Potions are not actually added to your Inventory and the Gems are not subtracted. This part is correct (outside the approved time for the Potions, you are not meant to be able to buy them).

The desired behaviour is something like one of these:
- you can't click on the pinned item and if you try to, you see a message saying that the item is no longer available for purchase (e.g., an on-hover message)
- you can click on it but the modal that appears tells you it's no longer available and doesn't contain the purchase options.
@Tressley for a design

That behaviour should be automatic as soon as the approved time for the Potion has ended so the code should read that date from Habitica's config for the Potion.

Regardless of how this fix is implemented, the user should always be able to unpin the item when they choose to (or leave it pinned permanently if they prefer that).

qrdean:
Fixes #9933 

### Changes
- Exposed the `canBuy()` and `yearRound()` functions values that appear on some item structures to the front end. 
- Make checks based on these exposed values to determine if we display an item unavailable message and disable the Buy button on the `buyModal` component
- Add the item unavailable message to common/locales/en/npc.json
- Added `canBuy: () => true` to items of type gear. Gear type items have a default `canBuy: () => false` . For that items state to match the front end explicit `canBuy: () => true` was added
- For the `itemAvailable()` function if the item is not of type card and does not have canBuy value of type boolean the function defaults to true. 

Habitica ID: @qdean
----
UUID:


website/client/src/components/shops/buyModal.vue
```diff
@@ -164,6 +165,18 @@
       ></span>
       <span class="limitedString">{{ limitedString }}</span>
     </div>
+    <div
+      v-if="!itemAvailable"
+      class="notAvailable"
+    >
+      <span
+        class="svg-icon inline icon-16 clock-icon"
```
paglias, line 173: can we add a `v-once` directive here to make sure the icon is only rendered once? helps with performances

website/client/src/components/shops/buyModal.vue
```diff
@@ -164,6 +165,18 @@
...
+        v-html="icons.clock"
+      ></span>
+      <span
+        class="notAvailableString"
```
paglias, line 177: this string never changes so `v-once` will make sure Vue does not re-render it

website/common/script/content/gear/index.js
```diff
@@ -53,6 +53,7 @@ each(GEAR_TYPES, type => {
         int: 0,
         per: 0,
         con: 0,
+        // @TODO: add checks to switch to true
```
paglias, line 56: is this comment needed?

website/client/src/components/shops/buyModal.vue
```diff
@@ -511,6 +545,18 @@ export default {
...
       return (!this.user.purchased.plan.customerId && !this.user.purchased.plan.consecutive.trinkets && this.getPriceClass() === 'hourglasses');
     },
+    itemAvailable () {
+      // for card types the item has the yearRound flag
+      if (this.item.purchaseType === 'card') {
+        return this.item.yearRound;
+      }
+      // Check if it has the canBuy property of type boolean. If not then
+      // we just return true.
+      if (this.item && (typeof this.item.canBuy === 'boolean')) {
```
paglias, line 555: why is this checking for `this.item` since it's already used in the check above so if it was undefined line 550 would have already thrown an error

paglias (commented):
Hi @qrdean , thanks for the PR!

Let me know if I'm missing something, but wouldn't it make more sense to not specify the `canBuy` function when it's going to return true and in `getItemInfo` do something like this:

```
canBuy: typeof item.canBuy === function ? item.canBuy() : true,
```

Also is there a reason why `potion`, `armoire` and `card` don't get the canBuy property? And why `card` has to use `item.yearRound` instead of a custom `canBuy` function?
@SabreCat could you give a look at this PR?

SabreCat:
This issue was addressed in later code work while we were waiting for updates from the contributor. Closing this PR!

Issue #11911, opened 2020-02-29
Completed To-Dos tab doesn't explain why not all To-Dos appear


Alys:
The completed To-Dos tab no longer explains that only the 30 most recent completed To-Dos are shown. This leads to confusion as shown by this report from Ben Bruce / @umdotdotben (580dbf6e-0975-43f3-8c5d-d0949c9cbd61) who was trying to review his completed Group Plan To-Dos: 

> "I'm using the new group function thats in beta and i noticed that when my friend or i check off a todo it no longer shows up in the completed list of todos. this is unfortunate because i'm paying even more for the group function than i was the premium account and being able to see whats been completed is very important. thanks!"
> 
> "upon continuing my search for completed group to-do's i discovered that if i delete a completed to-do the next missing todo appears at the bottom, it seems like theres a limit to how many are being displayed at one time? is there anything i can do about this"

For reference, the current text on the completed To-Dos tab is this (also shown in the screenshot below):
> Completed To-Dos are deleted after 30 days for non-subscribers and 90 days for subscribers.
> 
> These are your To-Dos
> To-Dos need to be completed once. Add checklists to your To-Dos to increase their value.

In addition, the following text used to be displayed:
> Your most recent 30 completed To-Dos are shown here. You can see older completed To-Dos from Data > Data Display Tool or Data > Export Data > User Data.

It's still available in the locales files:
https://github.com/HabitRPG/habitica/blob/a7fd8f7f527a7e3f8881fda9afffd1aaa63da6a4/website/common/locales/en/tasks.json#L6
The final part of it would need to be changed since the menu structure is different now.

I'm marking this issue as suggestion-discussion for now.

![image](https://user-images.githubusercontent.com/1495809/75608229-6bf24080-5b49-11ea-8463-1720a0a9269d.png)

Alys:
This issue still exists and is still causing confusion. @jo_march (f3f17ccf-e6ea-43fd-b1e7-cf39962ff972) reported in the Report a Bug guild:

**"I'm a subscriber so should be able to view 90 days of completed to-dos,  yet it looks as if I can only view the past month of to-dos. Can you check on this please? Thanks."**

saraolson:
We decided to remove the 30 to do limit because of the confusion it caused and leave the limit to only be time-based in order to better honor the subscriber perk. All to do's completed in the last 30/90 days now show within the completed to do UI.

PR #12250, opened 2020-05-28
remove references to codetriage.com

Alys:
This is a suggested change to remove the codetriage.com information and badge that had been added in https://github.com/HabitRPG/habitica/pull/10028 and https://github.com/HabitRPG/habitica/pull/11205 . I have not discussed this with any staff; I figured I might as well explain in a PR that shows the change. I understand you will want to close this if it's not desirable. :)

It's a nice service but it doesn't yet give the ability to control which issues are advertised by it. As shown at https://www.codetriage.com/habitrpg/habitica#help-out , it's currently presenting issues that are "in progress" (e.g., https://github.com/HabitRPG/habitica/issues/12248), "on hold: needs design" (e.g., https://github.com/HabitRPG/habitica/issues/11911), probably other kinds of issues too that also aren't "help wanted". It's also showing pull requests.

This means that it's going to be more misleading than useful for anyone subscribed to it, since we have a large proportion of issues that are not "help wanted". The ones that are "help wanted" are advertised in the Aspiring Blacksmiths guild, which all code contributors should be a member of, so codetriage currently isn't offering significant benefit over that.

It looks like codetriage has plans to allow label filtering in future (https://github.com/codetriage/codetriage/issues/498, https://github.com/HabitRPG/habitica/pull/11205#issuecomment-499693510) so we could add it back after that.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -13,8 +13,3 @@ Habitica uses [this Google form](https://docs.google.com/forms/d/e/1FAIpQLScPhrw
 # Contributing Code
 
 See [Contributing to Habitica](http://habitica.fandom.com/wiki/Contributing_to_Habitica#Coders_.28Web_.26_Mobile.29)
-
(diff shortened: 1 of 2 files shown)


paglias:
LGTM 👍

Issue #6867, opened 2016-03-12
Should not be able to recast Chilling Frost and Stealth skill / spell that gives no additional benefit


cdouglass:
Casting Chilling Frost (which prevents streaks from resetting on the next cron) costs 40 mana each time, but there is no benefit from doing so twice in one day. There is also no indicator of whether it has already been cast today, so it's easy to forget and cast it extra times, spending mana for no purpose. Proposed solutions: either it should only be possible to cast this spell once per day or there should be some indicator of whether it's already been cast.

My user id: be3421a9-9f57-4ed1-8804-7b1ba8702dd0

---

_edit by Alys, 2018-09-16:_

This is happening again for both Chilling Frost and Stealth. It was originally fixed (and the fix was working) in https://github.com/HabitRPG/habitica/pull/7110 / https://github.com/HabitRPG/habitica/commit/b97d7cc9f989cd8089d3a0883933dbe80faade6a so possibly some of that code has been accidentally removed, or something new has caused it to stop working. Regardless, looking at the previous fix is probably a good way to start investigating this.

Alys:
You can see that it's been cast if you look on the User --> Stats page, where it will say "Streaks Frozen" if you've cast it. The Data > Data Display Tool also tells you in the "Skills and Buffs" section. However both of those are FAR from obvious! 

You're right that it should not be possible to cast it more than once per day. After you've cast it:
- [ ] the skill should be greyed out (or some other visual effect to show it's no longer available)
- [ ] both hovering over it and clicking on it should give you a message saying that you've already cast it, your streaks are already frozen, and there's no need to cast it again.

That could be done on the website's user interface as a purely client-side change. It could also be done separately on the mobile apps, although probably not for some time because the apps have greater priorities. Ideally it should also be done by modifying the API so that it's impossible to cast the skill more than once per day, however that API change cannot be done until version 3 of the API is released some time in the future (there's a change freeze on non-essential API changes until then).

I'm going to leave this issue as is for a couple of days, and then if there's no complaints, I'm going to mark it as "help welcome now" for ONLY the change to the website's user interface. I believe it should be possible to make that client-side change with no API changes. If that goes ahead, we'd also mention this to the app developers as a possible future feature request.

[...]

PR #12255, opened 2020-05-31
Issue 12220 stealth tooltip show protected dailies

### LINKED ISSUE #12220: Stealth tooltip should state how many Dailies it will protect
In the tooltip for Stealth, it should tell you how many Dailies will not cause damage if you cast the skill.

(This was a feature on the old site and was logged in https://github.com/HabitRPG/habitica/issues/9136 for a fix. Moving it to here to split up that issue.)

alexandreabeh:
Fixes #12220

### Changes
A feature on the old site showed the number of dailies protected by casting Stealth. This is now showing again in the Stealth tooltip. The skillNotes section was not being called. It adds additional skill-specific info: the dailies avoided by stealth, whether stealth no longer needs to be cast (dailies already avoided) and whether frost is no longer useful to cast.

Speculation: the spellDisabled method had some commented out code regarding Stealth, which may have broken due to a change in how dailies are referenced. I have fixed this line, so it seemed alright to keep the entirety of the skillNotes function as it had been. However, this includes more than just showing dailies in the tooltip.

Behavior changes:
- tooltip shows dailies avoided for Stealth
- tooltip shows maxed out message for Stealth when all dailies are covered
- tooltip shows frost already cast message when frost has already been cast
- logic for whether stealth should be cast again had incorrectly previously calculated based on all dailies; now we look at all due and incomplete dailies

Screenshots:

All_dailies_avoided
Dailies_avoided_shown_in_tooltip

UUID: 4a07c222-b810-4559-a574-b95785cafac4

### DIFF
--- a/website/client/src/components/tasks/spells.vue
+++ b/website/client/src/components/tasks/spells.vue
@@ -267,14 +267,10 @@ export default {
       );
     },
     spellDisabled (skill) {
-      let isSpellDisabled = false;
+      if (skill === 'frost' && this.user.stats.buffs.streaks) return true;
+      if (skill === 'stealth' && this.user.stats.buffs.stealth >= this.user.tasksOrder.dailys.length) return true;
 
-      if ((skill === 'frost' && this.user.stats.buffs.streaks)
-        || (skill === 'stealth' && this.user.stats.buffs.stealth >= this.user.tasksOrder.dailys.length)) {
-        isSpellDisabled = true;
-      }
-
-      return isSpellDisabled;
+      return false;
     },
 
     skillNotes (skill) {
--- a/website/client/src/components/tasks/spells.vue
+++ b/website/client/src/components/tasks/spells.vue
@@ -270,8 +270,10 @@ export default {
       );
     },
     spellDisabled (skill) {
+      const incompleteDailiesDue = this.getUnfilteredTaskList('daily').filter(daily => !daily.completed && daily.isDue).length;
+
       if (skill === 'frost' && this.user.stats.buffs.streaks) return true;
-      if (skill === 'stealth' && this.user.stats.buffs.stealth >= this.getUnfilteredTaskList('daily').length) return true;
+      if (skill === 'stealth' && this.user.stats.buffs.stealth >= incompleteDailiesDue) return true;
 
       return false;
     },
--- a/website/client/src/components/tasks/spells.vue
+++ b/website/client/src/components/tasks/spells.vue
@@ -206,7 +206,7 @@
 <script>
 import spells from '@/../../common/script/content/spells';
 
-import { mapState } from '@/libs/store';
+import { mapState, mapGetters } from '@/libs/store';
 import notifications from '@/mixins/notifications';
 import spellsMixin from '@/mixins/spells';
 import Drawer from '@/components/ui/drawer';
@@ -237,6 +237,9 @@ export default {
   },
   computed: {
     ...mapState({ user: 'user.data' }),
+    ...mapGetters({
+      getUnfilteredTaskList: 'tasks:getUnfilteredTaskList',
+    }),
     openStatus () {
       return this.$store.state.spellOptions.spellDrawOpen ? 1 : 0;
     },
@@ -268,7 +271,7 @@ export default {
     },
     spellDisabled (skill) {
       if (skill === 'frost' && this.user.stats.buffs.streaks) return true;
-      if (skill === 'stealth' && this.user.stats.buffs.stealth >= this.user.tasksOrder.dailys.length) return true;
+      if (skill === 'stealth' && this.user.stats.buffs.stealth >= this.getUnfilteredTaskList('daily').length) return true;
 
       return false;
     },
--- a/website/client/src/components/tasks/spells.vue
+++ b/website/client/src/components/tasks/spells.vue
@@ -40,7 +40,7 @@
                 v-for="(skill, key) in spells[user.stats.class]"
                 v-if="user.stats.lvl >= skill.lvl"
                 :key="key"
-                v-b-popover.hover.auto="skill.notes()"
+                v-b-popover.hover.auto="skillNotes(skill)"
                 class="col-12 col-md-3"
                 @click="castStart(skill)"
               >
@@ -267,18 +267,16 @@ export default {
       );
     },
     spellDisabled (skill) {
-      if (skill === 'frost' && this.user.stats.buffs.streaks) {
-        return true;
-      }
-      // @TODO: Implement
-      // } else if (skill === 'stealth' && this.user.stats.buffs.stealth
-      // >= this.user.dailys.length) {
-      //   return true;
-      // }
+      let isSpellDisabled = false;
 
-      return false;
+      if ((skill === 'frost' && this.user.stats.buffs.streaks)
+        || (skill === 'stealth' && this.user.stats.buffs.stealth >= this.user.tasksOrder.dailys.length)) {
+        isSpellDisabled = true;
+      }
+
+      return isSpellDisabled;
     },
-    // @TODO is this used?
+
     skillNotes (skill) {
       let notes = skill.notes();
 
--- a/website/client/src/components/tasks/spells.vue
+++ b/website/client/src/components/tasks/spells.vue
@@ -274,7 +274,7 @@ export default {
       if (skill === 'frost' && this.user.stats.buffs.streaks) return true;
       if (skill === 'stealth' && this.user.stats.buffs.stealth >= incompleteDailiesDue) return true;
 
-      return isSpellDisabled;
+      return false;
     },
     skillNotes (skill) {
       let notes = skill.notes();
--- a/website/client/src/components/tasks/spells.vue
+++ b/website/client/src/components/tasks/spells.vue
@@ -204,7 +204,7 @@
 </style>
 
 <script>
-import spells from '@/../../common/script/content/spells';
+import spells, { stealthBuffsToAdd } from '@/../../common/script/content/spells';
 
 import { mapState, mapGetters } from '@/libs/store';
 import notifications from '@/mixins/notifications';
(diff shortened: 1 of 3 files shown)


Alys:
This is not the initial review you asked for and I think you're right to wait for Habitica's staff to comment on your approach but in my opinion those three behaviour changes are fantastic! It looks like you're fixing https://github.com/HabitRPG/habitica/issues/6867 too so I've marked that issue as in progress.

website/client/src/components/tasks/spells.vue
```diff
@@ -267,18 +267,16 @@ export default {
       );
     },
     spellDisabled (skill) {
-      if (skill === 'frost' && this.user.stats.buffs.streaks) {
-        return true;
+      let isSpellDisabled = false;
+
+      if ((skill === 'frost' && this.user.stats.buffs.streaks)
```
paglias: Especially if we add more conditions here in future I think this would look clearer if we do something like 

```
    spellDisabled (skill) {
        if (skill === 'frost' && this.user.stats.buffs.streaks) return true;
        if (skill === 'stealth' && this.user.stats.buffs.stealth >= this.user.tasksOrder.dailys.length) return true;

       return false

**Review by paglias — CHANGES_REQUESTED** [review]
The changes looks good, I left just one stylistic comment, regarding the other points that were mentioned in the comments:

- @Alys The tooltip should say `Number of dailies avoided: X` like it does now or `Number of dailies that will be avoided: X`? The current version seems to imply that they've already been avoided
- regarding #6867 this would fix the issue at the client side level, not at the API level (I would do both so on the website it is told immediately to users but also on the API so that it covers usage from the mobile apps or 3rd party scripts)
- tests: that would be great @alexandreabeh, thanks!

File: website/client/src/components/tasks/spells.vue (hunk 1 above)
Comment (alexandreabeh) [reply website/client/src/components/tasks/spells.vue]: Thanks @paglias ! Good point on the extensibility of those conditions, I'll update that. I'll likely have time to work on tests next weekend.

It sounds like #6867 will only be partially solved by this PR then.

Comment (Alys) [reply website/client/src/components/tasks/spells.vue]: > The tooltip should say `Number of dailies avoided: X` like it does now or `Number of dailies that will be avoided: X`? The current version seems to imply that they've already been avoided

Ah yes, `Number of dailies that will be avoided: X` is better!

**Comment by Alys** [comment]
> regarding #6867 this would fix the issue at the client side level, not at the API level (I would do both so on the website it is told immediately to users but also on the API so that it covers usage from the mobile apps or 3rd party scripts)

@paglias Good point! Thank you! I've updated that issue. 

When this PR is merged, I'll update the issue again to say what still needs to be done, and I'll mark it as help wanted again.
I'm labelling this PR as "notify helpers" to remind me to do that (I monitor that label every few days for closed PRs), and I'm commenting here rather than in the thread above so I can see this comment more easily later. :)

**Comment by alexandreabeh** [comment]
@paglias I'm trying to run the client unit tests locally and keep getting the following when I run `npm run client:unit` at the root of the repo. I've looked in the documentation and only see instructions about running unit tests for the api and not the client. Am I missing a configuration/installation?

`Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: Can't find stylesheet to import.
    ╷
106 │ @import '~@/assets/scss/colors.scss';
    │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ╵`

**Comment by paglias** [comment]
Hmm @alexandreabeh I tried running `npm run client:unit` and the tests run successfully on the `develop` branch

**Comment by paglias** [comment]
But if adding tests is not possible we can add them in a separate PR

**Comment by alexandreabeh** [comment]
@paglias personally I would like to learn how to run the tests locally and be able to add more, since there seems to be a need for it especially on the client side. I will have limited time in the next couple weeks to do so though. I'll let it be your call on whether to address that in a separate PR, since this is my first contribution and I don't yet know what is standard procedure.

File: website/client/src/components/tasks/spells.vue [hunk 2]
```diff
@@ -267,18 +267,12 @@ export default {
...
-      if (skill === 'frost' && this.user.stats.buffs.streaks) {
-        return true;
-      }
-      // @TODO: Implement
-      // } else if (skill === 'stealth' && this.user.stats.buffs.stealth
-      // >= this.user.dailys.length) {
-      //   return true;
-      // }
+      if (skill === 'frost' && this.user.stats.buffs.streaks) return true;
+      if (skill === 'stealth' && this.user.stats.buffs.stealth >= this.user.tasksOrder.dailys.length) return true;
```
paglias: Instead of using the `user.tasksOrder` array to get the number of dailies (which in some cases might not include the correct number of tasks) I would add a getter for the tasks, so inside the computed properties I would add

```
 ...mapGetters({
      getUnfilteredTaskList: 'tasks:getUnfilteredTaskList',
    }),
```

mapGetters will have to be imported with `import { mapState, mapGetters } from '@/libs/store';`

and then can be used here with `this.getUnfilteredTaskList('daily').length;`

paglias (changes requested):
I noticed a small issue in the code, for the tests let's see when this issue is fixed, I'm able to run the tests correctly though so it might be some issue with your git repository not being up to date maybe?

alexandreabeh:
Thanks @paglias! I should be able to spend some time on this over the weekend

alexandreabeh:
@paglias - ok, the good news is that I was able to work on this a bit today :) I made a slight modification from what you suggested because I noticed that otherwise, if I had completed all dailies, the conditional in spellDisabled would still evaluate to false. If I understand correctly, stealth should only be applying to cover dailies that are due and not complete. So I've added a filter for that - let me know if that makes sense.

The bad news - still encountering the same issues trying to run the tests locally. I can run ones for the api, but not the client.

paglias:
> The bad news - still encountering the same issues trying to run the tests locally. I can run ones for the api, but not the client.

This is quite weird... I would try:

1. Merge the most recent `develop` branch from Github into your branch
2. delete the `node_modules` folders both at the top level and under `website/client`, then from the main directory run `npm i` again

If it still doesn't work, I would then try to do `cd website/client` and then `npm run test:unit`

alexandreabeh:
Agreed, it is quite weird. I tried all of the following, with still the same errors
- merge most recent develop into my branch
- delete all node modules
- reinstall node modules
- npm run client:unit at the root
- cd into website/client (since that's what client:unit was supposed to do...and the console output for the above step did show that happening correctly) then `npm run test:unit`
- try the same thing on the develop branch
- found this post https://stackoverflow.com/questions/58546521/node-sass-and-sass-loader-module-build-failed-typeerror-this-getresolve-is-n and tried deleting node modules and downgrading the sass-loader to `7.3.1`
- also found https://github.com/vuejs/vue-cli/issues/4526 but it looks like those issues are fixed as of vue-cli v4 and above, which is what the project is set to use

More info:
I'm on Windows Version	10.0.18363 Build 18363
^ wish I could run Docker, but would need to upgrade my Windows version
npm version 6.9.0
node version 12.6.0

paglias:
Hmm I don't have experience with Windows, could you try posting about your problem in the Aspiring Blacksmiths guild https://habitica.com/groups/guild/68d4a01e-db97-4786-8ee3-05d612c5af6f ?

Regarding this PR I think I'll probably go ahead and merge it in the next few days if everything looks good

alexandreabeh:
Ok, thanks! I've posted in the guild.

website/client/src/components/tasks/spells.vue
```diff
@@ -267,18 +270,14 @@ export default {
...
-      // }
+      const incompleteDailiesDue = this.getUnfilteredTaskList('daily').filter(daily => !daily.completed && daily.isDue).length;
+
+      if (skill === 'frost' && this.user.stats.buffs.streaks) return true;
+      if (skill === 'stealth' && this.user.stats.buffs.stealth >= incompleteDailiesDue) return true;
 
       return false;
     },
-    // @TODO is this used?
+
```
paglias: So I just tested this PR locally and while the code works fine I think that `spellRogueStealthDaliesAvoided` will always show that 0 dailies will be avoided because `this.user.stats.buffs.stealth ` ( see lines 289-292 of this file ) is increased after the skill is casted not before. Am I correct @SabreCat @Alys ?

alexandreabeh: Hmm my followup questions would be - in the code for casting Stealth, the dailies are calculated using `user.tasksOrder.dailys.length` and we discussed above that in some cases this may not include the right number of tasks. What kinds of cases are those?
https://github.com/HabitRPG/habitica/blob/b5182ed7d454f0a01497da8472b54d642fbe43ee/website/common/script/content/spells.js#L212

If stealth increases only after the skill is cast, it seems like we'd want to make the following changes
- have a common method in spell.js that calculates the stealth buff to be added (lines 211 and 212 of spells.js)
- refactor spells.js to use the method --> user.stats.buffs.stealth +=  stealthBuffToBeAdded()
- for the stealth note previewing num dailies to be avoided, use the return value of stealthBuffToBeAdded() instead of the current stealth buff

Looking at git history though, using `this.user.stats.buffs.stealth` for that tooltip has been in place for the last 3 years. So is this a long standing bug or is there some context we're missing? 😅

Alys: @paglias I think you're correct.

@alexandreabeh I think it's long standing bug, alas. :-/ 
From what I can tell, the answer to that comment `// @TODO is this used?` has been No for possibly the last 3 years. :laughing:   I've checked the `frost` skill and it's special note isn't used either.

From my point of view, the approach you've described looks good but wait for paglias to comment and listen to him not me if he has other suggestions. :)

paglias: that approach looks good @alexandreabeh !

alexandreabeh: @paglias I've pushed changes with the approach we just talked about. I was able to see the '0' dailies behavior you were talking about, and now the tooltip gives the same preview number as whatever is calculated during the actual casting.

I decided to leave the dailies length code alone in the spell.js file after looking at the [Rogue wiki](https://habitica.fandom.com/wiki/Rogue#Stealth) and realizing the calculation there is supposed to include both complete and incomplete dailies. So that's different from the way we need to look at incomplete + due dailies for whether the spell should be disabled.

paglias:
@alexandreabeh thanks for this! I'm merging the PR and awarding your contributor tier.

alexandreabeh:
@paglias and @alys thanks for your help as well! Especially as a newbie contributor I appreciate your guidance throughout this process, and I'm definitely interested in contributing more in the future 👍

PR #12288, opened 2020-06-09
More Armoire content refactoring

SabreCat:

### Changes

The recent content deploy included a lot of slimming-down of the Armoire and Backgrounds content files. I spotted a little more I could do to make the Armoire builds easier, but the logic is a little complex due to the way things are structured, so I figured I'd throw this out for review and see if folks think it's an improvement or not.

**Previously**, the `notes` property of each Armoire item was manually constructed.
**Now**, we supply a default according to the following scheme:

1. If the gear item provides no stat bonuses (I don't think we have any like this now, but we have in the past), then construct it like `t('gearslotArmoireGearNameNotes')`
2. If the gear item has a single stat bonus, or more than one stat bonus where not all the stat bonuses are equal, then construct it like `t('gearslotArmoireGearNameNotes', {stat1: value, stat2: value})`
3. If the gear item has multiple stat bonuses all of which are equal, then construct it like `t('gearslotArmoireGearNameNotes', {attrs: value})`.

Some old items don't follow the `attrs` scheme, instead having a notes field that spells out each stat boost individually even though they're all the same; those have been left manually specified.

### DIFF
--- a/website/common/script/content/gear/sets/armoire.js
+++ b/website/common/script/content/gear/sets/armoire.js
@@ -1092,12 +1092,14 @@ forEach({
     const gearStats = {};
     const gearStatValues = [];
     let notes;
-    forEach(ATTRIBUTES, stat => {
-      if (gearItem[stat]) {
-        gearStats[stat] = gearItem[stat];
-        gearStatValues.push(gearItem[stat]);
-      }
-    });
+    if (!gearItem.notes) {
+      forEach(ATTRIBUTES, stat => {
+        if (gearItem[stat]) {
+          gearStats[stat] = gearItem[stat];
+          gearStatValues.push(gearItem[stat]);
+        }
+      });
+    }
     if (gearStatValues.length > 0) {
       if (gearStatValues.length === 1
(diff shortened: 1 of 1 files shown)


website/common/script/content/gear/sets/armoire.js
```diff
@@ -1304,8 +1089,29 @@ forEach({
   weapon,
 }, (set, setKey) => {
   forEach(set, (gearItem, gearKey) => {
+    const gearStats = {};
+    const gearStatValues = [];
+    let notes;
+    forEach(ATTRIBUTES, stat => {
```
paglias: should this be skipped if `gearItem.notes` is manually defined?

paglias (commented):
left one comment but otherwise looks good! Always a pleasure to see code lines going down :)

website/common/script/content/gear/sets/armoire.js
SabreCat: ah, that'd be a good shortcut, will do!

paglias (approved):

PR #12308, opened 2020-06-15
WIP. Changed the way how memberCount is incremented or decremented. Fixes #12275

### LINKED ISSUE #12275: Member Counts Inaccurate in Parties
### Description

Clean-up of #7454. You can visit that ticket to see the full history of discussion on this issue.

`memberCount` in group records is a stored computed value that we currently attempt to keep up to date by incrementing and decrementing as members join and leave. It frequently gets out of sync, leading to people with 6-member parties showing a member count of 7, etc.; this is a nuisance at best, and in some cases involving the 30-person party limit, can actually interfere with people joining or inviting members.

This is the desired fix:

> change the group remove, leave and join routes to recompute `memberCount` each time it needs to be updated (i.e., count all existing members, rather than add or subtract one from the current value of `memberCount`)

...when **Parties** are involved. At this time, we're leaving Guilds alone for performance reasons. See #12286 for the status of that arm of the issue.

JalanshMunshi:
### Changes
The `memberCount` attribute was earlier modified by incrementing/decrementing it by 1 when a user is added or removed respectively. This PR contains changes that will instead get the total count of the members currently present in the group and add or subtract 1 from it.

Fixes #12275.

This PR is not complete yet. We still need to have a look if we need to change any existing test cases or add any new ones.


Alys:
This isn't a full review (Habitica's staff will do that later), but keep in mind that this change needs to be for only Parties, not Guilds, for performance reasons (the issue has some more details about that). The functions that need to be modified are used for both Parties and Guilds, so the original code in those functions will need to be left in place for the case where they're used for a Guild. The group's `type` attribute will be either `party` or `guild`.

It also needs to apply to the case where the user leaves the Party voluntarily. `git grep memberCount` will show you all the places where the memberCount is updated.

JalanshMunshi:
Thanks for the feedback, @Alys !

I have made the suggested changes.

However, I am not sure about the changes for the case when a user leaves voluntarily. The current implementation executes the `leave` method for this scenario. 

https://github.com/JalanshMunshi/habitica/blob/44028dea6538bcd44450faa9fc80e9231a3f5302/website/server/models/group.js#L1356

Do we need to decrement the memberCount explicitly here?

website/server/controllers/api-v3/groups.js
```diff
@@ -603,7 +603,15 @@ api.joinGroup = {
       group.leader = user._id; // If new user is only member -> set as leader
     }
 
-    group.memberCount += 1;
+    if(group.type === 'party')
```
paglias: looks like the `lint` command is complaining about the code style, in particular I think the issue is with the lack of space after the if and the bracket not on the same line, something like this should work

```
if (group.type === 'party') {
```

paglias: can we add a comment explaining why the difference between guilds and parties exist when updating the `memberrCount`? Something like "For parties we count the number of members from the database to get the correct value, see https://github.com/HabitRPG/habitica/issues/12275 on why this is necessary and only done for parties."

paglias (changes requested):
I've left a couple of comments but the code looks good. There's an instance here /website/server/models/group.js#L1438 where the memberCount is updated differenty and also here https://github.com/HabitRPG/habitica/blob/9014943a86e07141853c1c861bdfcf8124c3a13d/website/server/models/group.js#L1724-L1729 could we update the code to use `await this.getMemberCount();` instead of the custom `User.find` query?

Thanks!

website/server/controllers/api-v3/groups.js
JalanshMunshi: Added the comment and changed the code style.

JalanshMunshi: Since you wanted to replace the custom queries, I used `await group.memberCount()` for both, guilds and parties. Please let me know if we need to keep the original code for guilds or not. 

Which instance related to `memberCount` are you referring to here (/website/server/models/group.js#L1438)? 

Also, I noticed that the lines https://github.com/JalanshMunshi/habitica/blob/6cccb17415ac6037d73460dca83d17f2f96d42f0/website/server/models/group.js#L1422-L1429 and https://github.com/JalanshMunshi/habitica/blob/6cccb17415ac6037d73460dca83d17f2f96d42f0/website/server/models/group.js#L1434-L1441 are same. 

I do not have much experience with Habitica's code, but as per the comments, they are performing the same set of tasks.

website/server/controllers/api-v3/groups.js
```diff
@@ -948,7 +955,12 @@ api.removeGroupMember = {
     }
 
     if (isInGroup) {
-      group.memberCount -= 1;
+      if (group.type === 'party') {
```
paglias, line 960: can we add the same comment of above here?

website/server/controllers/api-v3/groups.js
paglias: Good catch, I think the lines going from 1420 to 1429 can be removed

website/server/models/group.js
```diff
@@ -1410,6 +1410,7 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all', keepC
     // double check the member count is correct
     // so we don't accidentally delete a group that still has users in it
     let members;
+    this.memberCount = group.getMemberCount();
```
paglias: Good catch here as well but at this should be `await group.getMemberCount()` and also the lines belowe (1414-1420) aren't needed anymore now, could all be substituted by:

```
const memberCount = await group.getMemberCount();
if ((memberCount - 1) === 0) { // -1 to make sure the user leaving is not counted
    promises.push(group.remove());
    return Promise.all(promises);
}
```

paglias (changes requested):

website/server/controllers/api-v3/groups.js
JalanshMunshi: Added the comment.

website/server/models/group.js
JalanshMunshi: Thanks for the comments!

I removed the lines that you indicated. Those lines also included code that contained the list of members that need to be deleted.

Four test cases are failing due to the newly pushed changes. I tried to figure out the reason and I believe that the members are not being removed from a group. The line `_.remove(members, { _id: user._id });` was instrumental in removing the members but we are currently not doing so, as far as my understanding is concerned. This is what we are doing in the newly pushed changes.

https://github.com/HabitRPG/habitica/blob/abe60489fd77875292e50729679dca921c2ba9bc/website/server/models/group.js#L1414-L1417

Hence, the test cases are failing. Can you please help me out with this one?

website/server/models/group.js
```diff
@@ -1409,29 +1409,12 @@ schema.methods.leave = async function leaveGroup (user, keep = 'keep-all', keepC
   if (group.memberCount <= 1 && group.privacy === 'private') {
     // double check the member count is correct
     // so we don't accidentally delete a group that still has users in it
-    let members;
-    if (group.type === 'guild') {
-      members = await User.find({ guilds: group._id }).select('_id').exec();
-    } else {
-      members = await User.find({ 'party._id': group._id }).select('_id').exec();
-    }
+    this.memberCount = await group.getMemberCount();
```
paglias: Sorry for the delay @JalanshMunshi , and yes I think that this piece of code will have to be reverted. But the code is correct, the problem we're running into with tests is likely that here https://github.com/HabitRPG/habitica/pull/12308/files#diff-b6f239870392ed0ec2a2b73cbebaa699R1396-R1402 the operation to remove the guild or party from the user is already underway but we're not awaiting for it to finish meaning that when we call 

```
await group.getMemberCount();
```

it could already have finished or not resulting in an incorrect result. For example: It's finished so memberCount doesn't include the user leaving, if there were only 2 members and 1 left doing memberCount - 1 will result in the group being deleted even if there's still a member

So yeah I guess this will have to be reversed, the rest of the PR looks good!

paglias (changes requested):

website/server/models/group.js
JalanshMunshi: Thanks for getting back to me, @paglias !

I have reverted the changes. I guess these will be the final changes for this issue. Please have a final look at the PR and let me know if something is missing.

paglias: thanks @JalanshMunshi ! The PR looks good, I'll merge it at the end of the week

paglias (approved):

paglias:
Thanks @JalanshMunshi ! The PR will go live next week, if you post your Habitica UUID or username I'll award you a contributor tier

JalanshMunshi:
Pleasure's all mine, @paglias ! It has been a wonderful experience contributing to Habitica and getting myself more experienced with the world of Open Source. Looking forward to more! My Habitica username is `Jalansh`.

paglias:
Tier awarded!

PR #12312, opened 2020-06-20
Added code to update user tag list along with the existing code that already involves updating the user documents of challenge members

### LINKED ISSUE #12224: New tasks added to existing challenge should always show up with challenge tag
When you join a challenge, a tag for that challenge is created in your account (at least it is when [this bug](https://github.com/HabitRPG/habitica/issues/10689) doesn't happen). You have the ability to delete the tag if you want to while still in the challenge (that's intentional and shouldn't be changed; e.g., it's useful if you've retagged the challenge tasks with one of your own tags).

If the challenge owner adds new tasks to the challenge after you've joined, those tasks will be added to your task list but if the challenge tag does not current exist, the tasks won't be tagged probably. They will have the challenge/tag id in their `tags` list but since the tag itself doesn't exist, that isn't any use - you can't filter to find the new tasks.

The desired behaviour is that when new tasks from a challenge are added to your list, the tag is created IF it doesn't already exist. If it does already exist, a new copy of the tag should not be created.

However I'm not actually sure how easy this would be and it might add extra load if the only way to achieve it was to check for the presence of the tag for every challenge task, including the tasks that are added when you first join the challenge.

~I've labelled this suggestion-discussion for us to decide if it is a fix we want.~

_Thanks to @Altariel (be317a32-c027-4032-bbd5-25ccb21eb912) for reporting this._

tsukimi2:
When a new task is being added to a challenge, added code to update user tag list along with the existing code that already involves syncing / updating the user documents of challenge members.

Fixes #12224

See related issue (#12224) for discussion of proposed solution.

----
UUID:


website/server/models/challenge.js
```diff
@@ -235,7 +235,18 @@ async function _addTaskFn (challenge, tasks, memberId) {
   });
 
   // Update the user
-  toSave.unshift(User.update({ _id: memberId }, updateTasksOrderQ).exec());
+  const addToChallengeTagSet = {
```
paglias, line 240: can we add a comment explaining what this new code does? Also the `// Update the user` comment should be probably removed or moved right before `toSave.unshift..`

paglias (changes requested):
@tsukimi2 thanks! This looks good I only left one comment but we should also add a test to cover this behavior probably here (https://github.com/HabitRPG/habitica/blob/develop/test/api/unit/models/challenge.test.js) or in an integration test

test/api/unit/models/challenge.test.js
```diff
@@ -146,6 +173,60 @@ describe('Challenge Model', () => {
...
+      it('should add challenge tag back to user upon syncing challenge tasks to a user with challenge tag removed', async () => {
+        await challenge.addTasks([task]);
+
+        const newMember = new User({
+          guilds: [guild._id],
+        });
+        await newMember.save();
+        await challenge.syncTasksToUser(newMember);
+
+        let updatedNewMember = await User.findById(newMember._id);
```
paglias: this should have an `.exec()` at the end

test/api/unit/models/challenge.test.js
```diff
@@ -146,6 +173,60 @@ describe('Challenge Model', () => {
...
+        updatedNewMember.tags = [];
+        await updatedNewMember.save();
+
+        const taskValue2 = tasks2ToTest[taskType];
+        const task2 = new Tasks[`${taskType}`](Tasks.Task.sanitize(taskValue2));
+        task2.challenge.id = challenge._id;
+        await challenge.addTasks([task2]);
+        await challenge.syncTasksToUser(updatedNewMember);
+
+        updatedNewMember = await User.findById(updatedNewMemberId);
```
paglias: this should have an `.exec()` at the end

test/api/unit/models/challenge.test.js
```diff
@@ -146,6 +173,60 @@ describe('Challenge Model', () => {
...
+      it('should not add a duplicate challenge tag to user upon syncing challenge tasks to a user with existing challenge tag', async () => {
+        await challenge.addTasks([task]);
+
+        const newMember = new User({
+          guilds: [guild._id],
+        });
+        await newMember.save();
+        await challenge.syncTasksToUser(newMember);
+
+        let updatedNewMember = await User.findById(newMember._id);
```
paglias: this should have an `.exec()` at the end

test/api/unit/models/challenge.test.js
```diff
@@ -146,6 +173,60 @@ describe('Challenge Model', () => {
...
+        const updatedNewMemberId = updatedNewMember._id;
+
+        const taskValue2 = tasks2ToTest[taskType];
+        const task2 = new Tasks[`${taskType}`](Tasks.Task.sanitize(taskValue2));
+        task2.challenge.id = challenge._id;
+        await challenge.addTasks([task2]);
+        await challenge.syncTasksToUser(updatedNewMember);
+
+        updatedNewMember = await User.findById(updatedNewMemberId);
```
paglias, line 222: this should have an `.exec()` at the end

website/server/models/challenge.js
```diff
@@ -234,8 +234,21 @@ async function _addTaskFn (challenge, tasks, memberId) {
...
-  // Update the user
-  toSave.unshift(User.update({ _id: memberId }, updateTasksOrderQ).exec());
+  // Update the tag list of the user document of a participating member of the challenge
+  // such that a tag representing the challenge into which the task to be added will
+  // be added to the user tag list if and only if the tag does not exist already.
+  const addToChallengeTagSet = {
+    $addToSet: {
+      tags: {
+        id: challenge._id,
+        name: challenge.name,
```
paglias: I just tested this and while the code works correctly it creates a new tag when you already have the original one because `tag.name` should be `challenge.shortName` not `challenge.name`, see https://github.com/HabitRPG/habitica/blob/b947c714f0efb6de1b0558fff885913168890c82/website/server/models/challenge.js#L142

paglias (changes requested):
left a few comments about a minor issue (findById should be followed by `.exec()`) and one about the tag name being wrong, but it works fine!

paglias:
thanks @tsukimi2 ! Welcome to the 3rd tier

PR #12142, opened 2020-04-27
Teams UI Redesign and A11y Updates

negue:
Fixes #8730
Fixes #8764 
Fixes #9876
Fixes #10041
Fixes #11422 

Extracted to a different PR: #12274
- New Task Colors
- Button Styles
- Positive / Negative Habit Buttons 
- Title / Notes
- Delete Task Icon + Text

Extracted to a different PR: #12287
- Checklist 
- Delete Icons

Changes in this PR:
- [x] Modal in general
  - [x] Checklist
  - [x] Tags
- [x] Habit Modal
  - [x] Difficulty
  - [x] Reset Streak
  - [x] Advanced Settings
    - [x] Positive Streak
    - [x] Negative Streak
  - Team Options
    - [x] Assign to
    - ~Can be claimed~
    - [x] Requires approval
- [x] Daily Modal
  - [x] Difficulty
  - [x] Start date
  - [x] Repeats
  - [x] Repeat every
  - [x] Repeat on
  - [x] Advanced Settings
    - [x] Adjust Streak
  - Team Options
    - [x] Completion condition
    - [x] Assign to
    - ~Can be claimed~
    - [x] Requires approval
- [x] Todo Modal
  - [x] Difficulty
  - [x] Due date
  - Team Options
    - [x] Assign to
    - ~Can be claimed~
    - [x] Requires approval

PR #12314, opened 2020-06-20
Groups: always show checklist-toggle

negue:
Fixing the visibility of checklist-toggle, mentioned in #12142

I think `isUser` is outdated there, if the checklist items itself are visible

Now everyone can change the collapseChecklist field.

> But that also means if User A collapse it, and User B expands it, it'll be expanded again when User A refreshes it

### DIFF
--- a/website/client/src/components/tasks/task.vue
+++ b/website/client/src/components/tasks/task.vue
@@ -169,7 +169,6 @@
           >
             <div class="d-inline-flex">
               <div
-                v-if="isUser"
                 v-b-tooltip.hover.right="$t(`${task.collapseChecklist
                   ? 'expand': 'collapse'}Checklist`)"
                 class="collapse-checklist d-flex align-items-center expand-toggle"
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -26,11 +26,25 @@ import common from '../../../common';
 import logger from '../../libs/logger';
 import apiError from '../../libs/apiError';
 
-function canNotEditTasks (group, user, assignedUserId) {
+function canNotEditTasks (group, user, assignedUserId, taskPayload = null) {
   const isNotGroupLeader = group.leader !== user._id;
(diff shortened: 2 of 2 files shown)


website/client/src/components/tasks/task.vue
```diff
@@ -169,7 +169,6 @@
           >
             <div class="d-inline-flex">
               <div
-                v-if="isUser"
```
paglias, line 172: `collapseChecklist()` makes a PUT request to `/api/v4/tasks/${task._id}` so I think it'll result in a 401 error for users not allowed to edit the group tasks

Tressley: Yeah, I think I ran into this during testing. Would receive a red snack notification.

Tressley: image

negue: ok so what are our options:
- just let it save on the session not in the database? => no api call on the groups page
- we'd need a different place to store the combination of user, group and task to save the collapse state
- remove the toggle checklist (as it was) on the group page BUT without the space it had

did I miss anything?

paglias: Maybe we can give all users in the group the permission to toggle the checklist? Having the option individually for each user would be quite complicate as not all users have a copy of the task

Tressley: Yeah, ALL users are still able to collapse checklists on tasks, just not complete them.

paglias: @negue after talking with @Tressley the desired behavior is that the `collapsedChecklist` field of group tasks should be editable by everyone not just by the managers or the leader of the group but by all members

negue: ok, I didn't know there is also this kind of field :) , will enable it for all team users

Tressley:
This LGTM. 👍 

...merge it!

paglias:
Can we add a test for the new API permission?

PR #12332, opened 2020-06-24
tags names do not collapse when view width of the browser is changed, fixes #12253

### LINKED ISSUE #12253: tag screen too narrow for wide tags, especially as window size shrinks
_I'm marking this as suggestion-discussion for Habitica's staff to consider whether the Tags screen could be widened as described below, or whether another fix would be desirable._

When you click the "Tags" button to open the pop-up screen that displays your tags, the screen is rather narrow and tags that aren't very short will wrap. See the first screenshot below with my browser at full screen width.

The problem becomes more severe as the window width shrinks. See the second screenshot.

Since the new design of the Tags screen means that you can't use your tasks while you're looking at your tags, I'm wondering if it would be better if the tags screen took up more of the window space - you wouldn't lose any functionality from that and the Tags screen itself would be more usable.

_Thanks to Jubineo / @crThiago (12601b79-cf8e-44b7-94b9-2bafc97adc16) for reporting this._

**This screenshot was taken when my browser was at the full width of my screen:** 
image

**This screenshot was from my browser width was at a bit less than two-thirds of my screen (more than half the screen):**
image

wwyl1234:

Fixes #12253

### Changes
-Changed the flex box direction to column so that the tag headers are on the top of the tag names.
-Changed other CSS properties like min-height and max-height to the flex box so that tag names would not collapse.
-Adjusted the the flex basis on tag header so it takes less space.

Here are snapshots before the changes:
Screen Shot 2020-06-23 at 9 12 23 PM
Screen Shot 2020-06-23 at 9 12 40 PM

Here are some snapshots after the changes:
Screen Shot 2020-06-23 at 9 20 10 PM
Screen Shot 2020-06-23 at 9 19 57 PM
Screen Shot 2020-06-23 at 9 19 48 PM

This has been tested on Mac OS 10.13.6 on Google Chrome 83.0.4103.116 and Firefox 77.0.1 .

----
UUID: 89436423-878e-4ca6-bbd2-33eb0e65f61e


paglias:
@wwyl1234  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #12333, opened 2020-06-24
implemented profanity filter and racial slur filter checking to profi…

NickHolden:
…le blurb and username

Fixes put_#_and_issue_number_here

### Changes

----
UUID:

### DIFF
--- a/website/common/locales/en/settings.json
+++ b/website/common/locales/en/settings.json
@@ -210,5 +210,6 @@
     "mentioning": "Mentioning",
     "suggestMyUsername": "Suggest my username",
     "everywhere": "Everywhere",
-    "onlyPrivateSpaces": "Only in private spaces"
+    "onlyPrivateSpaces": "Only in private spaces",
+    "blurbIssueSlur": "About description may not contain inappropriate language nerd"
 }
--- a/website/server/libs/user/index.js
+++ b/website/server/libs/user/index.js
@@ -6,7 +6,7 @@ import {
   NotAuthorized,
 } from '../errors';
 import { model as User, schema as UserSchema } from '../../models/user';
-import { nameContainsSlur } from './validation';
+import { stringContainsBannedWord, stringContainsSlur } from './validation';
 
 export async function get (req, res, { isV3 = false }) {
   const { user } = res.locals;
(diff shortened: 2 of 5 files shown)


website/common/locales/en/settings.json
```diff
@@ -210,5 +210,6 @@
     "mentioning": "Mentioning",
     "suggestMyUsername": "Suggest my username",
     "everywhere": "Everywhere",
-    "onlyPrivateSpaces": "Only in private spaces"
+    "onlyPrivateSpaces": "Only in private spaces",
+    "blurbIssueSlur": "About description may not contain inappropriate language nerd"
```
Alys, line 214: This isn't at all a full review (one will be done later) but the final word in this string looks like it might be a typo. :)

I don't recall if the issue contained specific text to use. If it does, use that, and if it doesn't, can you change the string to the one below? Note the full stop at the end as well as the text changes.
"The About text may not contain inappropriate language."

Can you please also edit the top post on the PR to insert the issue number (that will create an automatic link to the issue) and to add your User ID? See the comments for where they should go. Your User ID is visible at [Settings > API](https://habitica.com/user/settings/api) on the website.

paglias:
Hi @NickHolden , can you link the issue this PR is fixing? thanks

paglias:
Hi @NickHolden , looks like the PR is causing some tests to fail and it's also missing tests for the new functionalities

website/server/libs/user/index.js
```diff
@@ -111,7 +111,25 @@ export async function update (req, res, { isV3 = false }) {
     const newName = req.body['profile.name'];
     if (newName === null) throw new BadRequest(res.t('invalidReqParams'));
     if (newName.length > 30) throw new BadRequest(res.t('displaynameIssueLength'));
-    if (nameContainsSlur(newName)) throw new BadRequest(res.t('displaynameIssueSlur'));
+
+    if (stringContainsSlur(newName)) {
```
paglias, line 115: can this and the next if be combined into one since they have the same code?

```
if (stringContainsSlur(newName) || stringContainsBannedWord(newName))
```

website/server/libs/user/validation.js
```diff
@@ -29,7 +47,8 @@ function usernameContainsInvalidCharacters (username) {
 export function verifyDisplayName (displayName, res) {
   const issues = [];
   if (displayName.length < 1 || displayName.length > 30) issues.push(res.t('displaynameIssueLength'));
-  if (nameContainsSlur(displayName)) issues.push(res.t('displaynameIssueSlur'));
+  if (stringContainsSlur(displayName)) issues.push(res.t('displaynameIssueSlur'));
```
paglias, line 50: can these two ifs be combined into one? If both are true you would push the same error text twice

paglias: or maybe we can have different strings for slurs and banned words

website/server/libs/user/validation.js
```diff
@@ -1,20 +1,38 @@
...
 import forbiddenUsernames from '../forbiddenUsernames';
 
 const bannedSlurRegexs = bannedSlurs.map(word => new RegExp(`.*${word}.*`, 'i'));
+const bannedWordsRegexs = bannedWords.map(word => new RegExp(`.*${word}.*`, 'i'));
 
-export function nameContainsSlur (username) {
-  for (let i = 0; i < bannedSlurRegexs.length; i += 1) {
-    const regEx = bannedSlurRegexs[i];
+export function stringContainsSlur (username) {
+  let x = 0;
```
paglias, line 10: is there a reason this was changed from a for loop to a while?

website/server/libs/user/validation.js
```diff
@@ -1,20 +1,38 @@
...
 import { getMatchesByWordArray } from '../stringUtils';
 import forbiddenUsernames from '../forbiddenUsernames';
 
 const bannedSlurRegexs = bannedSlurs.map(word => new RegExp(`.*${word}.*`, 'i'));
+const bannedWordsRegexs = bannedWords.map(word => new RegExp(`.*${word}.*`, 'i'));
 
-export function nameContainsSlur (username) {
-  for (let i = 0; i < bannedSlurRegexs.length; i += 1) {
-    const regEx = bannedSlurRegexs[i];
+export function stringContainsSlur (username) {
```
paglias, line 9: the parameter here (and also in stringContainsBannedWord()) should be changed from `username` to something more generic

PR #12337, opened 2020-06-25
added tests: if collapseChecklist is allowed and all others are still blocked

negue:
tests for #12314


paglias (changes requested):
looks like there's a lint problem!

PR #12338, opened 2020-06-26
Added the task type to the "Delete" alert window

### LINKED ISSUE #6146: Deleting a reward asks you if you want to delete the *task*

Shuyinsama:

Fixes #6146

### Changes

- Updated the text when deleting a task to reflect the type of the task that is deleted.
- Edited the "Delete this ..." button in the edit task modal to represent the task type

----
UUID: 4464eb1b-ea57-4490-a3d1-d0f3ed4a7d7a


website/common/locales/en/tasks.json
```diff
@@ -122,8 +122,8 @@
   "fortifyText": "Fortify will return all your tasks, except challenge tasks, to a neutral (yellow) state, as if you'd just added them, and top your Health off to full. This is great if all your red tasks are making the game too hard, or all your blue tasks are making the game too easy. If starting fresh sounds much more motivating, spend the Gems and catch a reprieve!",
   "confirmFortify": "Are you sure?",
   "fortifyComplete": "Fortify complete!",
-  "deleteTask": "Delete this Task",
-  "sureDelete": "Are you sure you want to delete this task?",
+  "deleteTask": "Delete this <%= type %>",
+  "sureDelete": "Are you sure you want to delete this <%= type %>?",
```
Alys: @Shuyinsama This isn't a full review (Habitica's staff will do that later), but this is one thing I've noticed that I think will need to be changed. BTW, there's no way you could have known about this, so it's definitely not a criticism. :)

Can you please choose new names for the two new strings? E.g., `deleteTaskType` and `sureDeleteType` 
(It doesn't have to be those exact names, they're just examples.)

I.e., your pull request should still delete the old `deleteTask` and `sureDelete` strings but it should add two new strings with new names.

The rest of this comment is just an explanation for why this may be necessary. Feel free to ask for clarification if I don't explain it well!

The reason is related to our current translation process. When a string is changed in the English locales files, the English change is copied to Weblate (where translations are done) but the old strings in other languages are **not** automatically changed. The other languages have to be manually updated by volunteers, and it can be months before all languages are updated.

For simple changes in wording, that's not usually a problem, but when a variable is added to a string, it's important that all languages use the new string immediately. For any language that still has the old string, players using that language will see error messages in Habitica because the string doesn't have the variable.

So, the best way to ensure that all languages use the new string is to give the string a new name. The new string still may not be translated into all languages for several months, but any language that doesn't have the new string will use the English version. A new name ensures that players don't see errors from missing variables.

paglias: 👍

paglias (changes requested):
@Shuyinsama thanks for the PR! It all looks good except for the strings names issue @Alys mentioned, when those are changed i'll merge the PR! thanks!

Shuyinsama:
@paglias & @Alys thank you for the feedback! I have changed the keys and the proper files to reflect that.

paglias:
Merged with https://github.com/HabitRPG/habitica/commit/a352a351bf68d4df55b04781332aa475af6f40ee

paglias:
Thanks @Shuyinsama ! The PR has been merged and noted in your profile towards your 2nd tier

PR #12342, opened 2020-06-27
Fix bug where updated webhook options failed to save (fixes #12336)

### LINKED ISSUE #12336: Webhook update route requiring "type" to be explicitly sent when changing options

### General Info
  * UUID: cab16cfa-e951-4dc3-a468-1abadc1dd109

### Description

From the Report a Bug Guild:

>There is an bug with the "Edit a webhook" (PUT https://habitica.com/api/v3/user/webhook/:id) endpoint which prevents you from modifying the webhook's options field. The update request itself actually returns the properly updated webhook, but if you fetch the webhook again you'll see that the changes don't persist. I did discover a workaround for this: if you change the type of the webhook (e.g. from taskActivity to userActivity) and then change the type back and provide the new options you want, the options will actually be saved.

Looking at the `formatOptions` function called from the webhook model, the `options` field is discarded if an appropriate `type` isn't found:

https://github.com/HabitRPG/habitica/blob/release/website/server/models/webhook.js#L90-L131

I would have expected that to get pulled in when the target webhook is found here in the controller...

https://github.com/HabitRPG/habitica/blob/release/website/server/controllers/api-v3/webhook.js#L194

...but possibly that's not happening as expected?

#### Console Errors

Searching server logs for this user's webhook-related requests, there were some timeout, missing auth header, and connection refused errors, but nothing that clearly points to this issue.

robwhitaker:
Fixes #12336

### Changes

This bug was caused by Mongoose not being able to tell when `Mixed` schema
objects (such as the `webhook.options` field) have been modified
(https://mongoosejs.com/docs/schematypes.html#mixed). So, although the webhook
was being updated properly, Mongoose was not actually committing it to the database.
Telling Mongoose that the webhook's `options` field has changed via `markModified`
fixes the issue.

Additionally, the relevant API test case was only checking whether or
not the webhook returned from the PUT endpoint matched the expected
update. Since the endpoint was returning the updated webhook without
querying the database again, this test case would pass. It has been
updated to check both the returned webhook as well as the version of the
webhook that is saved to the database against the expected. In other
words:

`assert returned === saved === expected`

----
UUID: cab16cfa-e951-4dc3-a468-1abadc1dd109

### DIFF
--- a/website/server/controllers/api-v3/webhook.js
+++ b/website/server/controllers/api-v3/webhook.js
@@ -223,9 +223,9 @@ api.updateWebhook = {
 
     webhook.formatOptions(res);
 
-    // Tell Mongoose that the webhooks array has been modified
+    // Tell Mongoose that the webhook's options have been modified
     // so it actually commits the options changes to the database
-    user.markModified('webhooks');
+    webhook.markModified('options');
 
     await user.save();
     res.respond(200, webhook);
--- a/website/server/controllers/api-v3/webhook.js
+++ b/website/server/controllers/api-v3/webhook.js
@@ -223,6 +223,10 @@ api.updateWebhook = {
 
     webhook.formatOptions(res);
 
+    // Tell Mongoose that the webhooks array has been modified
+    // so it actually commits the options changes to the database
+    user.markModified('webhooks');
+
     await user.save();
     res.respond(200, webhook);
   },
--- a/test/api/v3/integration/webhook/PUT-user_update_webhook.test.js
+++ b/test/api/v3/integration/webhook/PUT-user_update_webhook.test.js
@@ -127,19 +127,26 @@ describe('PUT /user/webhook/:id', () => {
   it('can update taskActivity options', async () => {
     const type = 'taskActivity';
     const options = {
+      checklistScored: true,
       updated: false,
-      deleted: true,
+      scored: false,
     };
-
-    const webhook = await user.put(`/user/webhook/${webhookToUpdate.id}`, { type, options });
-
-    expect(webhook.options).to.eql({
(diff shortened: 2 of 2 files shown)


website/server/controllers/api-v3/webhook.js
```diff
@@ -223,6 +223,10 @@ api.updateWebhook = {
 
     webhook.formatOptions(res);
 
+    // Tell Mongoose that the webhooks array has been modified
+    // so it actually commits the options changes to the database
+    user.markModified('webhooks');
```
paglias: The fix looks good but this way it's going to send back to the database the entire array even if only one item was modified, it works also if you do `webhook.markModified('options');` this way it's telling mongoose that the Mixed type `options` object is modified. The issue happens both with arrays but also mixed types objects as mongoose doesn't know what's been modified, see https://mongoosejs.com/docs/schematypes.html#mixed

paglias (changes requested):
@robwhitaker thanks for the PR! it looks good apart from the comment I left

website/server/controllers/api-v3/webhook.js
robwhitaker: Ah, okay. I've updated it to use `webhook.markModified('options')` instead. Also, I've updated the PR description to reflect the correction.

paglias: thanks @robwhitaker ! The pr looks good and will be merged at the end of the week

paglias (approved):

paglias:
thanks @robwhitaker ! I've noted this in your profile towards your 3rd tier and added the Blacksmith title

PR #12356, opened 2020-06-30
WIP. Accepting a redundant party invite will not remove the user from the party and let the user still be a part of it. Fixes #12291.

### LINKED ISSUE #12291: If user has redundant party invite, can get kicked out of party attempting to clear it

### General Info
  * UUID: 9bc9d551-61d1-4820-ae3d-9bca2ce2e470

### Description

It may be possible in some circumstances for a user to accept a party invite and end up _leaving_ the party instead of joining it. It's not clear exactly how @beffymaroo tripped it, but one of her test accounts was a member of a party; still appeared to have an outstanding invite to that party; accepted it, and ended up with no party at all.

Looking at this code in the group join route 

https://github.com/HabitRPG/habitica/blob/develop/website/server/controllers/api-v3/groups.js#L549-L559

there's no guard rail for the situation where the invite might be spurious. We go ahead and have the user leave their current party, even if it and the one they're ostensibly trying to join are the same. Could it be that this part of the process went through, but the later database updates meant to add the user _to_ the party failed, leaving them with none at all?

JalanshMunshi:

Fixes #12291 .

### Changes
Added an extra check for checking if the user is already a member of the party that the user is being invited to. The user will leave the party only if the invitation belongs to a party different than the current party. If the invitation is for the same party, then the notification for the party invite will be removed and the user will be redirected to the party.

The PR is not complete yet. We need to add a test case for a redundant party invite.

**Update [7/12/2020]**

Added the relevant test case.


paglias:
The code looks good @JalanshMunshi but I agree on the need to add a test covering this condition, let me know if you need help with that

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
```diff
@@ -203,6 +203,13 @@ describe('POST /group/:groupId/join', () => {
         await expect(invitedUser.get('/user')).to.eventually.have.nested.property('party._id', party._id);
       });
 
+      it('accepting a redundant party invite will let the user stay in the party', async () => {
+        await invitedUser.post(`/groups/${party._id}/join`);
+        await invitedUser.post(`/groups/${party._id}/join`);
```
JalanshMunshi, line 211: @paglias 
I tried to execute the repro steps of this issue i.e., sending the invite, manually adding the user to the party, and then executing the POST call for joining the party. This should be sufficient for testing the new changes.

However, I am not able to pre-add the member to the party. If I try to add using a POST request, then I get an error for either of the 2 reasons - 

1. Invitation cannot be sent as member is already in the party. 
2. A previous invite already exists.

I am not sure how to go ahead with this test. Can you help me out? 
Is there a way by which I can add the specific UUID to the party members list?

paglias: Yeah, so the API in normal circumstances will block you from inviting someone already part of your group. I think the only way to test it would be to directly editing the user document using `await invitedUser.update(... changes go here ...)` you can see in the existing tests how it's used for example here https://github.com/HabitRPG/habitica/blob/f8bab04a0e66a819f6d7ca5b210d27d991446bf3/test/api/v3/integration/user/POST-user_purchase.test.js#L38

JalanshMunshi: @paglias Thanks for the help! :D 

Please have a look at the PR and let me know if we need any other changes.

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
```diff
@@ -203,6 +203,15 @@ describe('POST /group/:groupId/join', () => {
         await expect(invitedUser.get('/user')).to.eventually.have.nested.property('party._id', party._id);
       });
 
+      it('accepting a redundant party invite will let the user stay in the party', async () => {
```
paglias: it's looking good @JalanshMunshi ! Just one last thing, can we add a reference to the fixed issue in the test title so something like `- issue #12291 ` at the end?

paglias (changes requested):

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
JalanshMunshi: @paglias Yeah sure. Like this? 

`Issue #12291: accepting a redundant party invite will let the user stay in the party`

Or add it in a comment just above the title?

paglias: `Issue #12291: accepting a redundant party invite will let the user stay in the party ` is perfect 👍

JalanshMunshi: Done. Resolving this conversation.

test/api/v3/integration/groups/POST-groups_groupId_join.test.js
```diff
@@ -203,6 +203,15 @@ describe('POST /group/:groupId/join', () => {
         await expect(invitedUser.get('/user')).to.eventually.have.nested.property('party._id', party._id);
       });
 
+      it('Issue #12291: accepting a redundant party invite will let the user stay in the party', async () => {
```
paglias, line 206: so I tested this test in `develop` so without your fix but the test passes correctly meaning that either the bug doesn't exist or the test case isn't replicating it correctly

cc @SabreCat

JalanshMunshi: Ouch. I should have checked that!

The bug is replicable on the localhost, so I guess it does exist. The problem seems to be with the test case. I will look into it. Meanwhile, please let me know if you are able to identify the problem.

JalanshMunshi: @paglias I made a minor change in the update statement after looking at some of the code which used the update statement. And since we are making an update to the party attribute, I also added an additional check to verify if the update statement is working or not. 

Could you please check if this fails in the `develop` branch or not? 

```
        await invitedUser.update({
          'party._id': party._id,
        });
```

I would have checked it myself but I am not able to run the integration tests on my system. Getting these errors (Timeout because of `"before all"` and `"after all"`  - 
`[16:17:07] The following tasks did not complete: test:api-v3:integration`
`[16:17:07] Did you forget to signal async completion?`

While we are at it, do you know what's causing this? Or should I ask in the Aspiring Blacksmiths guild?

paglias: it's working now @JalanshMunshi ! I hadn't notice the error with updating the user. Regarding the error you see when running the integration tests it might be that mongodb isn't running

paglias:
thanks @JalanshMunshi ! It'll go live by the end of the week. If you have an Habitica username post it and I'll award a contribution tier!

JalanshMunshi:
> thanks @JalanshMunshi ! It'll go live by the end of the week. If you have an Habitica username post it and I'll award a contribution tier!

Pleasure's all mine, @paglias ! My Habitica username is `Jalansh`.

paglias:
noted towards your 2nd tier!

PR #12358, opened 2020-07-01
update group members after accepting the quest (without page reload)

### LINKED ISSUE #11173: Accepting quest does not update participant numbers until refresh
> Bug: Accepting a quest does not update the participating player count (e.g., 3/10) until refresh.

From Report a Bug. Not confirmed.

Not sure if we expect this to update live, but filing it as the best way to get the question on the radar.

rogesson:
Fixes #11173

### Changes

- Current scenario

As mentioned in the request, when accepting a mission invite, the participant count was not increased by +1, remaining static even when the user has just accepted.

To update the number of participants who accepted, it was necessary to reload the browser page (F5).

- Investigation

I noticed that the number of group members was updated only when the page was loaded, which made the count remain the same until a new update.

- Adjustment

After the player clicked the "Accept" button, I updated the `this.group.quest.members` based on the `quest` returned in the`quests: sendAction` request.

Images:

**Before Click**
alt before_accept

**After Click (Without page reload)**
alt after_accept

- Specs

I tried to find some way to test this functionality unitarily, but I didn't find it, if anyone knows, please tell me :)
I did the tests through the local application (running on my computer) where I managed to run via the docker.

----
**UUID:** ef556559-6c3c-4ac3-8d70-6714007d2e51
First Pull Request =P

### DIFF
--- a/website/client/src/components/groups/questSidebarSection.vue
+++ b/website/client/src/components/groups/questSidebarSection.vue
@@ -425,6 +425,7 @@ export default {
     async questAccept (partyId) {
       const quest = await this.$store.dispatch('quests:sendAction', { groupId: partyId, action: 'quests/accept' });
       this.user.party.quest = quest;
+      this.group.quest.members = quest.members;
     },
     async questReject (partyId) {
       const quest = await this.$store.dispatch('quests:sendAction', { groupId: partyId, action: 'quests/reject' });
--- a/website/client/src/components/groups/questSidebarSection.vue
+++ b/website/client/src/components/groups/questSidebarSection.vue
@@ -424,8 +424,9 @@ export default {
     },
     async questAccept (partyId) {
       const quest = await this.$store.dispatch('quests:sendAction', { groupId: partyId, action: 'quests/accept' });
+      quest.progress.up = this.user.party.quest.progress.up;
       this.user.party.quest = quest;
-      this.group.quest.members = quest.members;
+      this.group.quest = quest;
     },
     async questReject (partyId) {
       const quest = await this.$store.dispatch('quests:sendAction', { groupId: partyId, action: 'quests/reject' });
--- a/website/client/src/components/groups/questSidebarSection.vue
+++ b/website/client/src/components/groups/questSidebarSection.vue
@@ -162,7 +162,7 @@
                   progress on the group doc? Each user could have different progress.-->
                 <span
                   class="float-right"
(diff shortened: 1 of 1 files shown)


website/client/src/components/groups/questSidebarSection.vue
```diff
@@ -425,6 +425,7 @@ export default {
     async questAccept (partyId) {
       const quest = await this.$store.dispatch('quests:sendAction', { groupId: partyId, action: 'quests/accept' });
       this.user.party.quest = quest;
+      this.group.quest.members = quest.members;
```
paglias: Hi @rogesson this code is working well but I noticed there's another small bug, if you're the last member to accepts the quest (so that after you accept it it should start) it doesn't show up as started until you refresh the page. I think it's happening because `this.group.quest` is not updated , I've not tested it but I feel like adding  `this.group.quest = quest;` should fix it

paglias (changes requested):

website/client/src/components/groups/questSidebarSection.vue
rogesson: Hello @paglias 

Yeaah! I will try it, thanks!

paglias:
looks like force pushing the branch doesn't result in a notification, I've set this as ready for review and will give it a final test in the next few days

website/client/src/components/groups/questSidebarSection.vue
```diff
@@ -424,8 +424,9 @@ export default {
     },
     async questAccept (partyId) {
       const quest = await this.$store.dispatch('quests:sendAction', { groupId: partyId, action: 'quests/accept' });
+      quest.progress.up = this.user.party.quest.progress.up;
```
paglias: @rogesson is this line necessary? what does it help with?

paglias: Oh I see, the user's quest object is reset with the party object which lacks the damage info, but @rogesson I think this only works for boss quests not for collection ones for example here https://github.com/HabitRPG/habitica/blob/84869ff6078a37c5cb5d849212292dab3bf1b898/website/client/src/components/groups/questSidebarSection.vue#L116, maybe it'd be better to sync the entire user and not rely on the party quest info to replace this.user.party.quest

cc @Alys

rogesson: Hello @paglias

Yeah! The `progress.up` was not being displayed correctly.

After your comment, I removed the assignment and created a new validation based on the example you had sent me.

Could you check again? Thank you!

paglias:
@rogesson  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #12379, opened 2020-07-07
Add bulk-feed via query parameter

### LINKED ISSUE #12368: Alter API to allow Bulk Feeding
Due to the API Limits to soon come into affect, bulk hatching and feeding pets will become very slow through the API. Tools like the [Bulk Pets Feed Tool](https://oldgods.net/habitica/cTheDragons/feed.html) will take a long time for users to complete triad bingo quickly as well as feed all their quests pets.

Please could we alter the  [/user/feed/](https://habitica.com/apidoc/#api-User-UserFeed) similar to [/user/sell](https://habitica.com/apidoc/#api-User-UserSell) to allow a query parameter of amount. If it is not specified it will assume amount of one.

PitiTheGrey:
example route: https://habitica.com/api/v3/user/feed/Wolf-Base/Meat?amount=8

Bulk feed with an amount of: 
Minimum of:
- amount of food needed to feed pet until it becomes mount
- amount of food in inventory
- amount of the query parameter

If no query parameter is given, the amount is assumed 1.

Fixes put_#_and_issue_number_here

### Changes

----
UUID: 9226cae6-d852-45da-a51b-ec2c9cb759e6
Alter API to allow Bulk Feeding 

Fixes #12368

### DIFF
--- a/website/common/script/ops/feed.js
+++ b/website/common/script/ops/feed.js
@@ -38,8 +38,13 @@ function evolve (user, pet, req) {
 export default function feed (user, req = {}, analytics) {
   let pet = get(req, 'params.pet');
   const foodK = get(req, 'params.food');
+  const queryAmount = get(req.query, 'amount', 1);
 
   if (!pet || !foodK) throw new BadRequest(errorMessage('missingPetFoodFeed'));
+
+  if (queryAmount < 0) {
+    throw new BadRequest(i18n.t('positiveAmountRequired', req.language));
+  }
 
   pet = content.petInfo[pet];
 
@@ -72,6 +77,7 @@ export default function feed (user, req = {}, analytics) {
 
   if (food.key === 'Saddle') {
     message = evolve(user, pet, req);
+    user.items.food[food.key] -= 1;
   } else {
     const messageParams = {
       egg: pet.text(req.language),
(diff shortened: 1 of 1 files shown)


paglias:
After thinking more about this, given that the current API response only includes the value of the pet after feeding is done, that it can't be a breaking change since it needs to be available for API v3 and that with the behavior implemented here where you don't know exactly how many pieces of food have been used, it would require API clients to make a new call to `GET /user` to get the remaining number of food items. 

So we should add a new key to the API response called something like `foodRemaining` where it shows the number of remaining food items after feeding.

As per the code itself @PitiTheGrey it looks good! It's just missing a couple of things:

- updated docs for the api route to show the new behavior here https://github.com/HabitRPG/habitica/blob/develop/website/server/controllers/api-v3/user.js#L858
- tests for the new functionalities here https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/user/POST-user_feed_pet_food.test.js and here https://github.com/HabitRPG/habitica/blob/develop/test/common/ops/feed.js

let me know if you need help with the tests!

PitiTheGrey:
Thank you Paglias for pointing out the importance of returning `foodRemaining`. I am on it.

PitiTheGrey:
I run into trouble with the desired api response-key `foodRemaining`:

The old (current) behaviour of the feed.js is to return the 
- number of the pet feed progress (from 5 to 50 or 0 or -1) and
- the message ('Wolf-Base likes food'). 

Code looks like this: 
`return [
    user.items.pets[pet.key],  // number
    message, // string
  ];` // Array?

___

If I insert a line into my new test-code: 
`return [
  [user.items.pets[pet.key], //number
  { foodremaining: user.items.food[food.key] }, // object
  message, // string
];` Array?

the response changes in a way (I think) that is not compatible with other client code anymore; 
so moving the response into a new Array (I think) is inappropiate. (I'd think the response changes from
result.data = 10 to result.data[0] = 10 (understand?)

If we work around like this: 
`return [
  [ {petFeedStatus: user.items.pets[pet.key], //number
  { foodremaining: user.items.food[food.key] }, // object
  message, // string
];` Array?
... we would hava a totally new key:
result.data[0] = petFeedStatus = 10
result.data[1] = foodRemaining = 14 
But that would totally screw up the old behaviour. 

Other workarounds like: 
`return [
   [user.items.pets[pet.key], 
   { foodRemaining: user.items.food[food.key}, 
   message, 
];`

did not work; they displayed `message: 32` e.g.

paglias:
I think changing the response might not be as easy as thought without introducing a breaking change. In this case I'd go for throwing an error if `?amount` is too much for the pet or if the user doesn't have enough pieces

PitiTheGrey:
Thank you for responding for my (deleted) question, Paglias.  
I will do as you suggest and throw errors.  
Thanks a lot!

paglias:
Closing in favor of #12384

PR #12380, opened 2020-07-09
PR to fix: Disallow line breaks in display names

### LINKED ISSUE #11946: Disallow line breaks in display names
A user managed to put a line break in their display name:

image

This seems like a bad idea, so the desired fix is to disallow any linebreaks in users' display names.

Amberlisi:

[//]: #11946
Fixes #11946

### Changes
This PR would make changes to 3 files:
1. settings.json - Needed for a new error message when a user tries to set their display name using a backslash followed by the letter N
2. validation.js - To add a validation check for the occurrence of a backslash followed by the letter N - Requires checking for '\\\n' and '\\n' because some systems automatically escape characters
3. index.js - To get the validator in the UI to check for the presence of a backslash followed by the letter N, and to prevent the user from updating their name if a backslash followed by the letter N is found

----
UUID: 74ad4bd2-9de5-43f7-8efd-cfd684f47374

### DIFF
--- a/website/server/libs/user/validation.js
+++ b/website/server/libs/user/validation.js
@@ -16,7 +16,7 @@ export function nameContainsSlur (username) {
 }
 
 export function nameContainsNewline (username) {
-  if (username.includes('\n') || username.includes('\\n')) {
+  if (username.includes('\n') {
     return true;
   }
   return false;
--- a/website/server/libs/user/validation.js
+++ b/website/server/libs/user/validation.js
@@ -18,9 +18,8 @@ export function nameContainsSlur (username) {
 export function nameContainsNewline (username) {
   if (username.includes('\n') || username.includes('\\n')) {
     return true;
-  } else {
-    return false;
   }
+  return false;
 }
 
 function usernameIsForbidden (username) {
--- a/website/server/libs/user/validation.js
+++ b/website/server/libs/user/validation.js
@@ -15,6 +15,14 @@ export function nameContainsSlur (username) {
   return false;
 }
 
+export function nameContainsNewline (username) {
+  if (username.includes('\n') || username.includes('\\n')) {
+    return true;
+  } else {
+    return false;
+  }
+}
+
 function usernameIsForbidden (username) {
   const forbidddenWordsMatched = getMatchesByWordArray(username, forbiddenUsernames);
   return forbidddenWordsMatched.length > 0;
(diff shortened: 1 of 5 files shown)


website/server/libs/user/validation.js
```diff
@@ -15,6 +15,13 @@ export function nameContainsSlur (username) {
   return false;
 }
 
+export function nameContainsNewline (username) {
+  if (username.includes('\n') || username.includes('\\n')) {
```
paglias: @Amberlisi can you remind me why we're checking both \n and \\n?

paglias (commented):
I've left a comment but it's looking good, we would need to add some tests covering this new check though. Let me know if you need help with them

website/server/libs/user/validation.js
Amberlisi: When changing your username in the web UI, it automatically escapes the backslash N. So it doesn't see it as a newline. Which is why you have to search for \\\n

The API works differently. It doesn't automatically escape it, so you have to check for a newline which is \\n.

Not sure if that explanation makes sense

paglias: Hmm I see but the moment is escaped (so the api gets `\\n`) is this an issue anymore? since it's escaped it shouldn't result in a visible new line. Not a big deal but I'm not sure the second check is necessary

Amberlisi: This depends on what we want the end functionality to be.

Do we want the users to be able to set their name to something that includes a \ + n using the UI, like so?
image

Or should the ability to do that in the UI be restricted just like it would be in the API? I would think so, for consistency. If something isn't allowed in the API, I don't think we would want to allow it anywhere else. That's just my opinion and we can go whichever way is desired

paglias: > Or should the ability to do that in the UI be restricted just like it would be in the API? 

The UI under the hood uses the API , what we want to prevent here it simply people having actual line breaks (that then get displayed ) in their display name or username, if they have the literal `\n` we don't care as long as it doesn't create a new line

Amberlisi: If that's the case then the second check can be left out. I'll make that change in a moment. If you could help me figure out how to perform tests for this that would be great

paglias: For the tests you should update https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/user/PUT-user.test.js#L55 to add a new case covering a profile.name that includes `\n` and add a check that it fails like it's done for the other invalid profile names. 

Something similar will have to be done for the route verifying the validity of a display name here https://github.com/HabitRPG/habitica/blob/develop/test/api/v4/user/auth/POST-user_verify_display_name.test.js with the difference that in this case each different invalid "case" has a separate test (identified by an `it(...)` block) so you'll have to add a new one for `\n`

Amberlisi: Okay, I added tests. Let me know if I did that right and if not what I need to change. I also removed the second check from the validation.

paglias (approved):
Everything's looking good! I'll merge it at the end of the week

paglias:
thanks @Amberlisi ! merged and noted towards your 2nd tier

PR #12395, opened 2020-07-13
Profile Page: redesign gear label 

### LINKED ISSUE #12381: Profile page out of alignment when translated versions of Habitica are used
From Report a Bug:

> In languages that have longer names for equipment types the profile stats part is no more in neat rows and columns: https://snipboard.io/knP3oq.jpg

image

negue:
restyled / re-aligned gear labels

also to improve development disable lintOnSave on development

fixes #12381

### DIFF
--- a/website/client/src/components/userMenu/profileStats.vue
+++ b/website/client/src/components/userMenu/profileStats.vue
@@ -17,7 +17,7 @@
             <div
               v-if="label !== 'skip'"
               :id="key"
-              class="box"
+              class="gear box"
               :class="{white: equippedItems[key] && equippedItems[key].indexOf('base_0') === -1}"
             >
               <div :class="`shop_${equippedItems[key]}`"></div>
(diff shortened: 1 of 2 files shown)


Tressley:
This LGTM, however, I noticed that the text clips and doesn't wrap:

image

Tressley:
This LGTM!

website/client/src/components/userMenu/profileStats.vue
```diff
@@ -684,4 +693,37 @@ export default {
...
+    .gear.box {
+      vertical-align: top;
+      margin: 0 auto;
+    }
+
+    .gear-label {
+      margin: 0 auto;
+      margin-top: 0.5rem;
+      min-height: 1rem;
+      font-family: Roboto;
```
paglias, line 707: is this rule alongside `font-stretch: normal;` `font-style: normal;` and `letter-spacing: normal;` necessary?

negue: gonna double check if it can used without or if there is a higher class/rule that overrides the "inherit" value

negue: removed them :)

PR #12399, opened 2020-07-18
Updating daysOfMonths array when the startDate of a monthly is updated in the API - Fixes #12041

### LINKED ISSUE #12041: "Monthly" Daily scheduling added to API
This is stemming from an issue in the mobile apps where the scheduling of a monthly Daily won't update correctly if the start date is changed. It was decided that it'd be better to fix it through the API than locally. 

You can see the detailed report from the original issue here: https://github.com/HabitRPG/habitica-android/issues/1272

laurel-thomson:

Fixes #12041

### Changes
I adjusted the 'Update Task' API endpoint to update the daysOfMonth array whenever the start date is updated.  Whenever the start date for a task with a frequency of 'monthly' is updated, the recurrence should be updated so that the daily occurs on the same day of the month.  For example, if a monthly daily is initially created with a start date of 7/1/2020, the daysOfMonth array will equal [1], indicating that the monthly will recur on the 1st of every month.  If this start date is updated to 7/2/2020, the daysOfMonth array should be updated to equal [2].

I also wrote two integration tests in PUT-tasks_id.test.js that test that (1) the daysOfMonth array is updated whenever start date is updated and (2) the nextDue array is updated to the hold the next six due dates with the correct day of the month.

----
UUID: 04ec9b30-1464-41f5-af67-fb2482f108a0

### DIFF
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -678,6 +678,10 @@ api.updateTask = {
       task.group.sharedCompletion = sanitizedObj.sharedCompletion;
     }
 
+    if (task.frequency === 'monthly' && task) {
+      task.daysOfMonth = [ moment(task.startDate).date() ];
+    }
+
     setNextDue(task, user);
     const savedTask = await task.save();
 
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -678,7 +678,12 @@ api.updateTask = {
       task.group.sharedCompletion = sanitizedObj.sharedCompletion;
     }
 
-    if (task.frequency === 'monthly' && task.startDate) {
+    // If the task was set to repeat monthly on a day of the month, and the start date was updated,
+    // the task will then need to be updated to repeat on the same day of the month as the new
+    // start date.  For example, if the start date is updated to 7/2/2020, the daily should
+    // repeat on the 2nd day of the month.  It's possible that a task can repeat monthly on a 
+    // week of the month, in which case we won't update the repetition at all.
+    if (task.frequency === 'monthly' && task.daysOfMonth.length && task.startDate) {
       task.daysOfMonth = [moment(task.startDate).date()];
     }
 
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -678,8 +678,8 @@ api.updateTask = {
       task.group.sharedCompletion = sanitizedObj.sharedCompletion;
     }
 
-    if (task.frequency === 'monthly' && task) {
-      task.daysOfMonth = [ moment(task.startDate).date() ];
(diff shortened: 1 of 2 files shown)


website/server/controllers/api-v3/tasks.js
```diff
@@ -678,6 +678,10 @@ api.updateTask = {
       task.group.sharedCompletion = sanitizedObj.sharedCompletion;
     }
 
+    if (task.frequency === 'monthly' && task.startDate) {
```
paglias: thanks @laurel-thomson ! The PR works well but I have one doubt, pinging @Alys for an extra check.

When a daily is set to repeat monthly `task.frequency === 'monthly'` it can repeat on the same day every month (for example the 18th of every month) or on `Day of the week` (as can be chosen on the website).

The difference between the two repetition schedules is defined by whether the `task.daysOfMonth` or `task.weeksOfMonth` arrays have values in them.

So from my understanding this bug only applies to `daysOfMonth` and it should not be set when `task.weeksOfMonth` has one or more values in it.

@Alys will be able to confirm but if I'm correct this code should only be applied when `task.weeksOfMonth` is empty or `task.daysOfMonth` is not before the update

paglias (commented):
left one comment but everything else looks good!

website/server/controllers/api-v3/tasks.js
Alys: Just to let you know, I'm going to try to look into this today. Depending on work pressures though, it may have to slip a day or two. Sorry @laurel-thomson I'll try to get to it as soon as I can!

Alys: Yes, paglias is right. The `task.daysOfMonth` array should be given a new value only if the player has specified that they want to use the "Day of the Month" "Repeat On" option.

Currently, the website front-end is responsible for setting either `task.daysOfMonth` or `task.weeksOfMonth` depending on which "Repeat On" option the user has chosen ("Day of the Month" or "Day of the Week") - see the website code linked from my  original issue https://github.com/HabitRPG/habitica-android/issues/1272.
The API doesn't yet have the ability to make that decision, and I think that moving all of that front-end code into the API would be a breaking change - i.e., not something we want to do.

Fortunately I don't think moving it all into the API will be necessary to fix the original issue. I think all that's needed is to change the value in the `task.daysOfMonth` array only if the array already has a value. So you can add a check for whether it has a value to your `if` statement on line 681.

paglias: > The API doesn't yet have the ability to make that decision, and I think that moving all of that front-end code into the API would be a breaking change - i.e., not something we want to do.

agreed, a simple if to check the array will be enough in this case

laurel-thomson: Sounds good!  I've updated the PR

paglias:
thanks @laurel-thomson ! Noted in your profile towards your 2nd tier

PR #12404, opened 2020-07-20
Casting Chilling Frost and Stealth skill again will not be processed and return an error instead. Fixes #12361.

### LINKED ISSUE #12361: API should prevent Chilling Frost and Stealth skills being cast when there is no point
The Mage skill Chilling Frost will prevent the streaks from all your incomplete Dailies being set to zero when cron runs. This skill needs to be cast only once on each day. Casting it more than once on the same day has no additional benefit and just wastes Mana.
**The API should prevent Chilling Frost from being cast if it has already been cast that day.**
I.e., if the user has `stats.buffs.streaks` set to `true` and the user tries to cast `frost` again, the API should prevent the skill from having any effect and should return the string `spellWizardFrostAlreadyCast` as an error message.

The Rogue skill Stealth prevents incomplete Dailies from causing damage when cron runs. Only some Dailies are affected each time it's cast, so it is sometimes correct to cast it more than once each day, but after all your Dailies have been protected, the skill has no no additional benefit and just wastes Mana.
**The API should prevent Stealth from being cast again if all Dailies are already protected.**
I.e., if the user's value of `stats.buffs.stealth` is high enough and the user tries to cast `stealth` again, the API should prevent the skill from having any effect and should return the string `spellRogueStealthMaxedOut` as an error message.

See https://github.com/HabitRPG/habitica/pull/12255 for the website front-end change that was done recently to adjust the website's Stealth tooltip so that `spellRogueStealthMaxedOut` is shown in the tooltip when Stealth no longer needs to be cast. In particular, the code within the `skillNotes` function that you can see in the diffs for that pull request may be a useful reference when you're working on this issue.

--- 

_(NB: it would be ideal for the skill note that is returned by the API to tell you whether the skill still needed to be cast, but that's more difficult because the API's content route, which contains the skill note that the API returns, does not need authentication, and so it can't easily be adjusted to display customised information for each user. It's not essential for the note to be adjusted, so just preventing the skills being cast when there's no need for them should be enough to close this issue.)_

### LINKED ISSUE #12362: buttons for website's Chilling Frost and Stealth skills should be unclickable/disabled when there is no point casting them
_I'm marking this as "on hold: needs design" for Habitica's staff (especially @Tressley) to comment about whether design work is is needed for this and/or to give specifications for how the skill buttons should look or behave._

As described in the API issue https://github.com/HabitRPG/habitica/issues/12361, the Mage skill Chilling Frost needs to be cast only once each day, and the Rogue skill Stealth doesn't need to be cast again after it has already protected all incomplete Dailies. In the pull request https://github.com/HabitRPG/habitica/pull/12255 the tooltips for those skills were modified to indicate when the skills no longer need to be cast. In addition to that, **the skills' buttons should not be clickable and/or should look greyed-out or otherwise disabled when there is no longer any need to cast the skills.**

JalanshMunshi:
Added logic for a repeating Chilling Frost skill. Added test case for redundant chilling frost skill cast. Added comments for the logic of repeating Stealth skill because of an error.

Fixes #12361 
Fixes #12362

### Changes

- [x] The API will throw a `BadRequest` when we try to cast a repeating Chilling Frost skill. 
- [x] Test case for the case of a repeated cast of the Chilling Frost skill.
- [x] The API will throw a `BadRequest` when we try to cast a repeating Stealth skill.
- [x] Test case for the case of repeated cast of the Stealth skill.
- [x] Change the message for Stealth and Chilling Frost recast.


website/server/libs/spells.js
```diff
@@ -163,14 +164,30 @@ async function castSpell (req, res, { isV3 = false }) {
...
-    let userToJson = user;
-    if (isV3) userToJson = await userToJson.toJSONWithInbox();
-
-    res.respond(200, {
-      user: userToJson,
-    });
+    // Check if chilling frost or stealh skill has been previously casted or not.
+    // See #12361 for more details.
+    const spellName = spell.key;
+    // const incompleteDailiesDue = getUnfilteredTaskList('daily').filter(daily => (daily.completed == false && daily.isDue === true)).length;
```
JalanshMunshi: @paglias Facing an issue here. When I try to use the `incompleteDailiesDue` function, it imports it from `website/client/src/store/getters/tasks.js` as it has been exported from there. 

However, after doing so, I keep getting this error.
```
/home/jalansh/Desktop/habitica/website/client/src/store/getters/tasks.js:1
import sortBy from 'lodash/sortBy';
^^^^^^

SyntaxError: Cannot use import statement outside a module
```
As of now, I have commented out the logic for the Stealth skill. I referred this method's usage in the PR #12255 . Do you know what might cause this?

As another approach, we can implement the same method or get the incomplete and due dailies for the user in spells.js to avoid this error.

test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
```diff
@@ -161,6 +161,22 @@ describe('POST /user/class/cast/:spellId', () => {
       });
   });
 
+  it.only('Issue #12361: returns an error if chilling frost has already been cast', async () => {
```
JalanshMunshi: This exclusive test and other console statements will be removed in the upcoming commits.

website/server/libs/spells.js
paglias: incompleteDailiesDue is current part of the client side code, to use it on the server it'll have to be duplicated in a new function here or be moved to /common/script so that it can be used in both places

paglias: but in this case I think the code would be different so I'd make a new function

JalanshMunshi: Thanks for the feedback, @paglias ! That helped. I added code for getting incomplete and due dailies and it works.

I also need the logic in `website/common/script/content/spells.js` to avoid the deduction in `MP`. I wrote the same implementation there as well. But I am facing a console error in my browser saying (and the website does not load) - 
```
Uncaught Error: Cannot find module './script/index'
    webpackEmptyContext common sync:2
    <anonymous> index.js:10
    js app.js:1017
    __webpack_require__ app.js:854
    fn app.js:151
    <anonymous> task.js:41
    js app.js:2493
    __webpack_require__ app.js:854
    fn app.js:151
    <anonymous> spells.js:19
    js app.js:1629
    __webpack_require__ app.js:854
    fn app.js:151
    <anonymous> index.js:24
    js app.js:1557
    __webpack_require__ app.js:854
    fn app.js:151
    <anonymous> index.js:9
    js app.js:16252
    __webpack_require__ app.js:854
    fn app.js:151
    <anonymous> analytics.js:23
    js app.js:15580
    __webpack_require__ app.js:854
    fn app.js:151
    <anonymous> cjs.js:17
    ./node_modules/cache-loader/dist/cjs.js?!./node_modules/babel-loader/lib/index.js!./node_modules/cache-loader/dist/cjs.js?!./node_modules/vue-loader/lib/index.js?!./src/components/header/menu.vue?vue&type=script&lang=js& app.js:4539
    __webpack_require__ app.js:854
    fn app.js:151
    <anonymous> menu.vue:1
    ./src/components/header/menu.vue?vue&type=script&lang=js& app.js:12460
    __webpack_require__ app.js:854
    fn app.js:151
    <anonymous> menu.vue:1
    vue app.js:12448
    __webpack_require__ app.js:854
    fn app.js:151
    <anonymous> cjs.js:37
    ./node_modules/cache-loader/dist/cjs.js?!./node_modules/babel-loader/lib/index.js!./node_modules/cache-loader/dist/cjs.js?!./node_modules/vue-loader/lib/index.js?!./src/app.vue?vue&type=script&lang=js& app.js:4012
    __webpack_require__ app.js:854
    fn app.js:151
    <anonymous> app.vue:1
    ./src/app.vue?vue&type=script&lang=js& app.js:8156
    __webpack_require__ app.js:854
    fn app.js:151
    <anonymous> app.vue:1
    vue app.js:8144
    __webpack_require__ app.js:854
    fn app.js:151
    <anonymous> main.js:12
    js app.js:15784
    __webpack_require__ app.js:854
    fn app.js:151
    1 app.js:16265
    __webpack_require__ app.js:854
    checkDeferredModules app.js:46
    <anonymous> app.js:994
    <anonymous> app.js:997
app.js line 1006 > eval:2:10
```
I tried fixing it but was not able to. Can we not use a method of the server side code in `/common/script`? Is there any other method by which I can write this?

SabreCat: `common` files can import from one another, and server and client code can use common code, but common can't use server- or client-only code. Can you walk me through what you're trying to solve and why the server code is needed? I'm not spotting it on a look through the PR as it stands.

JalanshMunshi: @SabreCat Okay, I understand now. I need to get the number of dailies which are incomplete and due. For that, I added the following code - 
```
const incompleteDailiesDue = Tasks.Task.find({
        userId: user._id,
        type: 'daily',
        $and: [
          { type: 'daily', completed: false, isDue: true },
        ],
      }).exec();
```
This works perfectly on the server side. I also need the list in `common` to avoid deducting `MP`. But I won't be able to import the `Tasks` model from server according to what you said. 

Hence, it is best if we implement the function in `common` so that we can use it on the server side as well. but how can I use the `Tasks` model here? Is there any other way for getting the list of incomplete and due dailies? Does this clarify the problem?

SabreCat: Aha, thanks for the clarification, I understand the trouble now!

Probably the most appropriate place to carry out a calculation like that would be in `castSpell` in `website/server/libs/spells.js`... but that'd require making a special case for `stealth` there, or having e.g. an `abortFunction` property in the spell content that provides code for the server to run, neither of which feels great. @paglias what do you think?

paglias: I think the only way to do this is to have a special case for `stealth` as @SabreCat suggested in `website/server/libs/spells.js ` that will return an error in case the skill should not be casted, the code on the client can remain the same. If it turns out that in future a similar exception is needed for more skills we can look into make it more generic but for now I don't think it's necessary

JalanshMunshi: Thanks for the reply, @SabreCat and @paglias !

I am already throwing a `BadRequest` for the repeated `Stealth` skill. 
```
else if (spellName === 'stealth' && user.stats.buffs.stealth >= incompleteDailiesDue.length) {
      throw new BadRequest(res.t('spellRogueStealthMaxedOut'));
    } 
```
By keeping the client-side code same, are you saying that we are letting the `MP` deduct? Do I need to do include additional code in the `castSpell` method in `website/server/libs/spells.js` for avoiding the change in `MP`? Can you please clarify?

website/server/libs/spells.js
```diff
@@ -163,14 +163,32 @@ async function castSpell (req, res, { isV3 = false }) {
...
-    let userToJson = user;
-    if (isV3) userToJson = await userToJson.toJSONWithInbox();
-
-    res.respond(200, {
-      user: userToJson,
-    });
+    // Check if chilling frost or stealh skill has been previously casted or not.
+    // See #12361 for more details.
+    const spellName = spell.key;
+    const incompleteDailiesDue = await Tasks.Task.find({
```
paglias: instead of loading the tasks you can simply count the number of results by doing `.countDocuments().exec()` instead of `.find()`

website/server/libs/spells.js
```diff
@@ -163,14 +163,32 @@ async function castSpell (req, res, { isV3 = false }) {
...
+      userId: user._id,
+      type: 'daily',
```
paglias: can the query be simplified into: 

```
{
    userId: user._id,
    type: 'daily',
    completed: false,
    isDue: true,
}
```

since by default mongo treats all properties of the query as ANDs

website/server/libs/spells.js
```diff
@@ -163,14 +163,32 @@ async function castSpell (req, res, { isV3 = false }) {
...
+      $and: [
+        {type: 'daily', completed: false, isDue: true },
+      ],
+    }).exec();
+    if (spellName === 'frost' && user.stats.buffs.streaks) {
+      throw new BadRequest(res.t('spellWizardFrostAlreadyCast'));
+    }
+    else if (spellName === 'stealth' && user.stats.buffs.stealth >= incompleteDailiesDue.length) {
```
paglias: can we run the query to calculate the number of incomplete dailies only when stealth is being used?

so that it's not run un-necessarily when using other spells

website/server/libs/spells.js
```diff
@@ -163,14 +163,32 @@ async function castSpell (req, res, { isV3 = false }) {
...
+    // See #12361 for more details.
+    const spellName = spell.key;
+    const incompleteDailiesDue = await Tasks.Task.find({
+      userId: user._id,
+      type: 'daily',
+      $and: [
+        {type: 'daily', completed: false, isDue: true },
+      ],
+    }).exec();
+    if (spellName === 'frost' && user.stats.buffs.streaks) {
```
paglias: for `frost` since we only need  `stats.buffs.streaks` to see if it can be used or not and that is available in common code as well could we move this check there? so that we don't have to add a new one on the client

website/server/libs/spells.js
paglias: > By keeping the client-side code same, are you saying that we are letting the MP deduct?

No I meant that the client side already has a check to prevent using `stealth`  when `user.stats.buffs.stealth >= incompleteDailiesDue.length` so it doesn't require any change

paglias (changes requested):
left some comments but it's looking like it's the right direction

website/server/libs/spells.js
JalanshMunshi: Modified. This was a better idea. Thanks!

website/server/libs/spells.js
JalanshMunshi: Modified the query.

website/server/libs/spells.js
JalanshMunshi: I have added the check in the `common` code. Please have a look at it if I need to change something.

website/server/libs/spells.js
JalanshMunshi: Yes, this would be better. The query will now run only when `Stealth` is cast.

website/server/libs/spells.js
JalanshMunshi: Ah, got it. I noticed that `MP` was deducted on a repeating cast, but that was happening only in the UI. The `MP` was not deducted when I reloaded the page. Is this expected? Should this be changed?

SabreCat: Hmm, that does sound incorrect... if the code @paglias mentions were working correctly, then you'd see the alert about not needing to use Stealth again, not a deduction of MP that is later reverted.

JalanshMunshi: @SabreCat The alert will be visible when my changes are merged with the master branch. The API will send a `BadRequest` which will lead to an error. The tooltip (see #12255 ) is also visible when we try to recast Stealth or Chilling Frost. 

What I am saying is, the UI shows that `MP` is deducted and that is why I initially included the logic to avoid `MP` deduction in `website/common/script/content/spells.js`. However, `MP` is not actually deducted in the database according to what I observed on my local install. Hence, it works fine as @paglias mentioned, but I guess ideally the UI should not show the deduction. What do you think? Please check this from your end as well.

SabreCat: OK, I see what's going on with the client--we have logic to show the "you don't need to cast this again" tooltip, but it doesn't actually prevent the cast. For that we'd need to e.g. change line 45 of website/client/src/components/tasks/spells.vue to `@click="!spellDisabled(key) ? castStart(skill) : null"`

So if you'd like to fix that, you can, but it's fine to keep this PR's scope server-side! In which case, the spurious -MP result that's reverted after sync is probably acceptable.

I am seeing that the error message you get for attempting to overuse Stealth talks about streaks being frozen, which is the Chilling Frost description, though.

paglias: If we could fix the MP deduction in the client side in this PR it would be great, I thought it was already fixed in the previous PR but it doesn't look like so

website/server/libs/spells.js
```diff
@@ -163,14 +164,35 @@ async function castSpell (req, res, { isV3 = false }) {
       task: results[1],
     });
   } else if (targetType === 'self') {
-    await castSelfSpell(req, user, spell, quantity);
+    // Check if chilling frost skill has been previously casted or not.
+    // See #12361 for more details.
+    const spellName = spell.key;
+    if (!chillingFrostAlreadyCast(spellName, user)) {
```
paglias: sorry, @JalanshMunshi is there a reason why the check is made here and not in common code so that it applies to the client side as well?

JalanshMunshi: @paglias Would you be so kind as to please clarify where should I add it in common code? I am not able to understand what you're trying to say because the function has been implemented in `common` code and I am calling that function in server code.

Also, it seems that there is something wrong with this as the sanity check says that this import will not work in production. Can you please help me out?

website/server/libs/spells.js
JalanshMunshi: Yes, sure. I will make the changes that @SabreCat suggested in my next commit. While we are it, should I also add the wording change that @veeeeeee suggested in #12361 for both `Stealth` and `Chilling Frost` skills? And if we are keeping the same message for both the skills, then should we keep a common JSON key for both the skills instead of two different keys?

Please note that I have added the test case for both the cases. Please have a look at it and let me know if something needs to be changed.

SabreCat: Ah yes, good catch on V's suggested wording. One string token used in both cases should be fine!

website/server/libs/spells.js
paglias: The sanity check is because importing common modules cannot be done as a normal import, but in this case what I meant was to move the check here https://github.com/HabitRPG/habitica/blob/develop/website/common/script/content/spells.js#L107 so that the code is executed both on the client and the server.

You can see an example of something similar being done for the healing spell here https://github.com/HabitRPG/habitica/blob/develop/website/common/script/content/spells.js#L230

Does it make sense?

JalanshMunshi: @paglias Yes, that helps! Works as expected.

Btw, I did what @SabreCat suggested.
> For that we'd need to e.g. change line 45 of website/client/src/components/tasks/spells.vue to @'click="!spellDisabled(key) ? castStart(skill) : null"

The button becomes disabled, as the code clearly suggests. But the API is never called (I put a `console.log` for `Stealth` skill and checked). As per the issue description, the API should display an error. This change failed the test cases that I wrote for the new changes. 

We can avoid the deduction by keeping the check here - 
https://github.com/HabitRPG/habitica/blob/a501b2a5a631bc06f6c45c2bd023f15a51913094/website/common/script/content/spells.js#L713-L716

But again, we will be able to do this for `Chilling Frost`, but not for `Stealth` as it requires an import of `Tasks` in common which is not possible as per the previous discussion in this PR.

What do you suggest we do in this case?

paglias: > But the API is never called (I put a console.log for Stealth skill and checked). As per the issue description, the API should display an error. This change failed the test cases that I wrote for the new changes.

This is expected as the API call is made when the button is clicked, the issue is about the API retuning an error in this case but if it's not called it cannot return anything, but it's fine the important thing is that the moment the API is called for example with a script or from the mobile apps it return an error

test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
```diff
@@ -161,6 +161,38 @@ describe('POST /user/class/cast/:spellId', () => {
...
+      'stats.mp': 400,
+      'stats.buffs.streaks': true,
+    });
+    await expect(user.post('/user/class/cast/frost'))
+      .to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: t('spellWizardFrostAlreadyCast'),
+      });
+    expect(user.stats.mp).to.equal(400);
```
paglias: the user object here hasn't been updated after the API call so it will always show the previous value, to make sure it's the last version you should call `await user.sync()` before checking the MPs value

test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
```diff
@@ -161,6 +161,38 @@ describe('POST /user/class/cast/:spellId', () => {
...
+      'stats.mp': 400,
+      'stats.buffs.stealth': 1,
+    });
+    await expect(user.post('/user/class/cast/stealth'))
+      .to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: t('spellRogueStealthMaxedOut'),
+      });
+    expect(user.stats.mp).to.equal(400);
```
paglias, line 178: here as well, a call to user.sync wil be necessary

website/server/libs/spells.js
paglias: I'm sorry but I'm a bit confused here:

### For stealth
- the check requires access to the tasks meaning that there are two different checks, one on the client (already added in a previous pr) and one on the server side added in this PR

### For chilling frost
- since the check doesn't require access to the tasks it can be done in a single place inside the common code that gets run both on the client and on the server and it's called on the client before the MPs are deducted if you do as I suggested above and move the check in here https://github.com/HabitRPG/habitica/blob/5bd4c6af4116186e51419a2f1721dcebb9fac606/website/common/script/content/spells.js#L107 instead of calling it in this file that is server side only

Let me know if something isn't clear

SabreCat: Yup, that solution for the client will mean there's no API call made--which is a good thing. We could force a snackbar notification, but with the tooltip saying "there's no need to cast this" I think it's OK to simply have a null response to the click event.

For manual testing of your API changes using the site, you'll need to have the old `@click` code in place!

JalanshMunshi: Ah, I understand now. So sorry for the confusion. Yes, we will indeed need the changes in the API in case it is called by a script or mobile app. 

I have made the requested change in the client code (in `spells.vue`). The appropriate changes to the test case have also been done. They are passing when I use the old code of `spells.vue`. Please check from your end as well. 

Lastly, I started working on the wording change and we will need to change all the JSON files for different languages. I have added the changes for three files in my last commit. Could you please check it and let me know if it is okay? I will modify the other files accordingly. 

Lastly, a doubt (Please bear with me if it a relatively foolish one, but I don't know about this :sweat_smile:). Since the test case would fail if we add the new code in `spells.vue`, should we remove these test cases in the final commit? 

Let me know if anything needs to be changed elsewhere! :smiley:

SabreCat: Only `website/common/locales/en` needs to be changed for language JSON. The rest is handled via Weblate, translate.habitica.com!

What test case is failing when you change the `spells.vue` line? I don't see `client-unit` failing at all.

JalanshMunshi: @SabreCat Oh, I didn't know that. That's awesome! I will revert the unnecessary changes then.

I am talking about the API-V3 Integrations tests. I added 2 new test cases in V3 Integration tests. Am I missing something here?

test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
JalanshMunshi: I missed it. My bad. I have added it to both the places now. Please check.

test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
```diff
@@ -161,6 +161,40 @@ describe('POST /user/class/cast/:spellId', () => {
...
+      'stats.lvl': 15,
+      'stats.mp': 400,
+      'stats.buffs.stealth': 1,
+    });
+    await user.sync();
+    await expect(user.post('/user/class/cast/stealth'))
+      .to.eventually.be.rejected.and.eql({
+        code: 400,
+        error: 'BadRequest',
+        message: t('spellAlreadyCast'),
```
paglias, line 176: @JalanshMunshi looks like the tests are failing because the `spellAlreadyCast` string does not exists in the original English strings (website/common/locales/en/spells.json)

JalanshMunshi: Hmm, it is passing in my local machine. I will make the wording change in the mentioned file and then check again.

JalanshMunshi: All checks are passing now. Please have a look at the PR and let me know if I need to change anything.

@paglias Also, I have added the string `spellAlreadyCast` to `locales/en/spells.json` and reverted the changes I made in the JSON files in my previous commit. We are now using a common string `spellAlreadyCast` instead of two separate ones for `Stealth` and `Chilling Frost` (`spellRogueStealthMaxedOut` and `spellWizardFrostAlreadyCast` respectively). Please let me know if we need to remove the old string messages from other JSON files as well. I have removed them from `locales/en/spells.json`.

website/server/libs/spells.js
SabreCat: Nope, that makes sense. I was puzzled because a change in a `.vue` file should only affect client tests, if anything, but it sounded like you were saying that the `spells.vue` change caused a server test failure!

test/api/v3/integration/user/POST-user_class_cast_spellId.test.js
```diff
@@ -161,6 +161,40 @@ describe('POST /user/class/cast/:spellId', () => {
...
+    await user.update({
+      'stats.class': 'wizard',
+      'stats.lvl': 15,
+      'stats.mp': 400,
+      'stats.buffs.streaks': true,
+    });
+    await user.sync();
+    await expect(user.post('/user/class/cast/frost'))
+      .to.eventually.be.rejected.and.eql({
+        code: 401,
```
SabreCat: Why is this one a status 401 while the other is 400? 400 seems more appropriate.

website/common/script/content/spells.js
```diff
@@ -105,6 +105,9 @@ spells.wizard = {
     target: 'self',
     notes: t('spellWizardFrostNotes'),
     cast (user) {
+      // Check if chilling frost skill has been previously casted or not.
+      // See #12361 for more details.
+      if (user.stats.buffs.streaks === true) throw new NotAuthorized(t('spellAlreadyCast')(user.language));
```
SabreCat: should be BadRequest I think?

SabreCat (changes requested):

website/common/script/content/spells.js
JalanshMunshi: Yes, that's correct. I had earlier thrown a `BadRequest`. But this line is in `website/common/script/content/spells.js` and throwing a `BadRequest` will require and import from `server/libs/errors` ( `import { BadRequest } from '../../../server/libs/errors';`)

Won't the sanity check give an error since we are trying to import from the server in the common code? What do you think?

paglias: You can import a BadRequest error directly from common code like it's done for NotAuthorized at the top of this file

```
import { NotAuthorized } from '../libs/errors';
```

would become 

```
import { NotAuthorized, BadRequest } from '../libs/errors';
```

the `server/libs/errors ` file basically re-exports the errors that were defined in `common/script/libs/errors`

JalanshMunshi: Ah, I see. I have made the changes. Thanks for the help, @paglias ! :)

JalanshMunshi: Please let me know if there are any other changes. I haven't yet updated my develop branch and merged it with this branch.

website/common/script/content/spells.js
```diff
@@ -105,6 +105,9 @@ spells.wizard = {
     target: 'self',
     notes: t('spellWizardFrostNotes'),
     cast (user) {
+      // Check if chilling frost skill has been previously casted or not.
+      // See #12361 for more details.
+      if (user.stats.buffs.streaks === true) throw new BadRequest(t('spellAlreadyCast')(user.language));
```
paglias:
```suggestion
    cast (user, target, req) {
      // Check if chilling frost skill has been previously casted or not.
      // See #12361 for more details.
      if (user.stats.buffs.streaks === true) throw new BadRequest(t('spellAlreadyCast')(req.language));
```

@JalanshMunshi I think this is the last change needed, everything else looks good :) Basically cast accepts 3 parameters: user, target (not applicable in this case and `req`. Using req.language  is better as it allows the user to override the languages stored in their preferences.

Could you also change it for the `heal` spell at line 230? Thanks! After that I'll make a final manual testing and it should be good to go

paglias (changes requested):
one last change :)

website/common/script/content/spells.js
JalanshMunshi: @paglias I am not sure I understand why is the test failing at line 30 in `test/common/ops/spells.js`. The error's `NotAuthorized` and that is what we are expecting. Can you please help?

paglias:
It looks like the last commit is failing because the test here https://github.com/HabitRPG/habitica/blob/develop/test/common/ops/spells.js#L28 is being passed only user and nothing for `req`, changing it to `spell.cast(user, null, { language: 'en' });` should fix it. 

And lastly I'd also move the chilling frost test to that file as it's more appropriate since the check happens in the common code and not in the server one

test/common/ops/spells.js
```diff
@@ -35,4 +35,21 @@ describe('shared.ops.spells', () => {
       done();
     }
   });
+
+  it('Issue #12361: returns an error if chilling frost has already been cast', async () => {
+    await user.update({
```
paglias: sorry, the test structure here should follow the previously existing test in this file (the one for healer) because in integration tests we simulate actual HTTP calls to the API while here we call the cast function directly, so something like this should work

```
    user.stats.class = 'wizard';
    user.stats.lvl = 15;
    user.stats.mp = 400;
    uster.stats.buffs.streaks = true;

    const spell = spells.wizard.frost;

    try {
      spell.cast(user, null, { language: 'en' });
    } catch (err) {
      expect(err).to.be.an.instanceof(BadRequest);
      expect(err.message).to.equal(i18n.t('spellAlreadyCast'));
      expect(user.stats.mp).to.eql(400);

      done();
    }
```

You'll also need to import BadRequest alongside notauthorized at the top of this file

paglias: also this isn't an async function so the async at the top should be removed

JalanshMunshi: Thanks! Yeah, I was able to figure that out from the message. The tests should pass now.

JalanshMunshi: @paglias A doubt. What is `done` here? I noticed that we are using the same name for the arrow function in the common test.

JalanshMunshi: I am always prone to lint errors. :laughing: :see_no_evil:

paglias: lint errors are bit difficult to catch unless you have a plugin for your code editor that highlights them , depending on the code editor you're using there might be one available

regarding done: when Mocha (our test runner) sees a function with the `done` parameter it is there to tell it that until the done parameter is called (it's a function) the test hasn't finished running.

It's used to signal test completion for tests with async callback usually as otherwise it wouldn't be possible to know when the test has finished running. See https://mochajs.org/#asynchronous-code

In this case it's used to make sure that the `catch` block is called as `done()` is called only inside the `catch` so if no error occurs in `spell.cast()` the test would never end and Mocha would signal a problem

paglias:
@JalanshMunshi the changes are looking good, I'm going to page @Tressley for a final design check but it should be good to go

website/client/src/components/tasks/spells.vue
```diff
@@ -42,7 +42,7 @@
                 :key="key"
                 v-b-popover.hover.auto="skillNotes(skill)"
                 class="col-12 col-md-3"
-                @click="castStart(skill)"
+                @click="!spellDisabled(key) ? castStart(skill) : null"
```
paglias, line 45: @Tressley this line disables the spell from being cast when not necessary but since the disabled state for spells is missing it'll show the popover saying it's not useful anymore but it'll keep the pointer cursor and hover state. Should it be reversed waiting for https://github.com/HabitRPG/habitica/issues/12362 or it can be shipped with this anyway in the meantime?

Tressley: This should be reversed until we're able to remove that conflicting states. However, the disabled state is very simple and should just be setting `opacity: 0.5;`

paglias: @JalanshMunshi I'd say the easiest thing is to apply the style @Tressley suggested when the spell is disabled but otherwise reverting the changes to @click is fine as well

paglias: ah, also the cursor should be set to `cursor: default`

JalanshMunshi: @paglias Yeah, we can apply the style in this PR if you want. Will it be possible for you to guide me through it? I tried to fix it on my own, but I was unable owing to lesser experience with Vue. 

Do we have to make the change in the css of `.popover` class when the `skillDisabled` is true?

SabreCat: @JalanshMunshi The basic way to apply a style dynamically in Vue is to add a line like this to the element:

`:class="{class-name: expression}"`

where `expression` is the condition when the class should be applied, and `class-name` is a CSS class name--you can define it in the `scoped` block further down the .vue file.

So here, you'd add something like

`:class="{'disabled': spellDisabled(key)}"`

to the `div` on line 48, then under `.spell` in the CSS (say, around line 118) add

```
&.disabled {
  opacity: 0.5;
  cursor: default;
}
```

Does that make sense?

JalanshMunshi: @SabreCat Yes, it does! Thanks for explaining it so well. The line 
`:class="{'disabled': spellDisabled(key)}"`

was already added in the code at line 50 in `spells.vue`.

Hence, the only thing remaining was to add the specific class named `.disabled`. Please have a look at the change and let me know if we need to change something.

SabreCat: The `:class` line was on the wrong element (in fact, it was on an `a` element with no content... who knows how that got in there in the first place 🤔), so it didn't work. Have you been testing your changes locally after updating the code?

I went ahead and added a commit with the correct fixes, as well as a couple of other bits of cleanup I figured I'd do while I was in there, ha.

JalanshMunshi: I had the same doubt and since it was already there, I thought of asking first. My browser console showed that the style was applied, but I was not sure whether we had to keep it inside the `a` element as your previous comment said that we had to put it in the `div` element. Went through your commit. I understand the changes now.

Thanks, @SabreCat ! :smiley:

JalanshMunshi: Anything else that needs to be changed/added in this PR?

SabreCat: Things are looking good to me. @paglias ?

website/client/src/components/tasks/spells.vue
```diff
@@ -127,7 +130,7 @@
 
       span {
         display: inline-block;
-        width: 50%;
```
paglias: @JalanshMunshi looks like this change broke the layout:

Schermata 2020-08-09 alle 12 28 43

^ this is how it looks in develop

Schermata 2020-08-09 alle 12 31 31

^ this without the `width: 50%;`

paglias (changes requested):
looks like something broke in the layout but otherwise it's looking good

website/client/src/components/tasks/spells.vue
paglias: left one new comment but the disabled state looks good

website/client/src/components/tasks/spells.vue
paglias: actually sorry looks like this was committed by @SabreCat

paglias:
thanks @JalanshMunshi ! Welcome to the 2nd tier!

JalanshMunshi:
Pleasure's all mine, @paglias !

PR #12438, opened 2020-08-03
Remove Tavern from API v3 list of guilds when 'guild' or 'publicGuilds' type parameter get added. (Fixes #12407)

### LINKED ISSUE #12407: API must have exception to not include Tavern in list of guilds
If you use the [api/v3/groups](https://habitica.com/apidoc/#api-Group-GetGroups) route to get a list of guilds for a user, then the returned list includes the Tavern. The Tavern should not be included.

I.e., if you do this command:
`curl -s -X GET -H "x-api-user: $HRPG_USER_USERID" -H "x-api-key: $HRPG_USER_APITOKEN" https://habitica.com/api/v3/groups"?type=publicGuilds&paginate=false"`
then the returned list includes all appropriate guilds but also this entry:
`{"_id":"00000000-0000-4000-A000-000000000000","privacy":"public","memberCount":7577,"balance":0,"leader":"7bde7864-ebc5-4ee2-a4b7-1070d464cdb0","name":"Tavern","type":"guild","categories":[]}`

That entry represents the Tavern and it's included in the data from that route because from a technical perspective the Tavern is stored as a guild (however from a user's perspective it's not a guild).

**The desired fix is that the api/v3/groups GET route should have an exception coded into it to prevent the group with the ID "00000000-0000-4000-A000-000000000000" from being included.**

Note that that group ID is available as a constant:
https://github.com/HabitRPG/habitica/blob/fef9c74f9b9ab9b49ae4a8f9380faa7007fb5535/website/common/script/constants.js#L7

_This issue was created from https://github.com/HabitRPG/habitica/issues/7431_

AlexanderColen:

Fixes #12407

### Changes
Removed the Tavern from showing up when fetching the list of guilds from API v3 and adding `type=guild` or `type=publicGuilds` as a query parameter.

Also added two new tests regarding this behaviour and fixed the failing tests as a result of this change.

----
UUID: 37028869-4c20-4ceb-8345-58b81b2619e8


test/api/v3/integration/groups/GET-groups.test.js
```diff
@@ -236,15 +236,27 @@ describe('GET /groups', () => {
...
       const page2 = await expect(user.get('/groups?type=publicGuilds&paginate=true&page=2'))
-        .to.eventually.have.a.lengthOf(1 + 4); // 1 created now, 4 by other tests
-      expect(page2[4].name).to.equal('guild with less members');
+        // 1 created now, 4 by other tests, -1 for no more tavern.
+        .to.eventually.have.a.lengthOf(1 + 4 - 1);
+      expect(page2[3].name).to.equal('guild with less members');
     }).timeout(10000);
   });
 
+  it('makes sure that the tavern doesn\'t show up when guilds is passed as a query', async () => {
```
paglias, line 245: for this and the next test could we also add a check that `guilds.length` is `0` so that we're sure there's only `guilds[0]` to check?

website/server/controllers/api-v3/groups.js
```diff
@@ -344,7 +345,14 @@ api.getGroups = {
     }
 
     if (req.query.member) {
-      filters._id = { $in: user.guilds };
+      // Includes 'guilds' nor 'publicGuilds' type.
```
paglias: can we move the check to `Group.getGroups ` here https://github.com/HabitRPG/habitica/blob/develop/website/server/models/group.js#L279 so that the tavern is never returned when that method is used instead of just in this route?

website/server/controllers/api-v3/groups.js
```diff
@@ -344,7 +345,14 @@ api.getGroups = {
...
+      if (req.query.type === 'publicGuilds' || req.query.type === 'guilds') {
```
paglias: this could be simplified by doing something like this outside of the if block

```
if (!filters._id) {
    filters._id = {};
}

filters._id['$ne'] = TAVERN_ID;
```

so that multiple cases depending on whether we're fetching guilds or publicGuilds aren't necessary

paglias (changes requested):
left some comments but the direction is the right one!

website/server/controllers/api-v3/groups.js
AlexanderColen: _Move_ the check or keep it here and _add_ the check to `Group.getGroups` as well?

website/server/controllers/api-v3/groups.js
AlexanderColen: Ah I was looking for documentation on how to do that one properly with Mongoose, since I hadn't touched it before. Always nice to learn something new! 👍 

What exactly did you mean with:
> so that multiple cases depending on whether we're fetching guilds or publicGuilds aren't necessary

The documentation for type parameter says:
> Possible values are party, guilds, privateGuilds, publicGuilds, tavern.

Did you mean with this that the tavern should never be fetched? Because I was under the impression that it could be fetched with the `tavern` type parameter. Unless I am misunderstanding the code, it will never be able to fetch the Tavern if `filters._id['$ne'] = TAVERN_ID;` would always be used to filter it out, even with the `tavern` type parameter.

test/api/v3/integration/groups/GET-groups.test.js
AlexanderColen: Because of previous tests, `guilds.length` at that moment is 4 and 2 for type `guilds` and `publicGuilds` respectively.

This seems to be because in the `before` some guilds already got created for other tests, like this one:
https://github.com/HabitRPG/habitica/blob/13a5b276e94e3e571e9a5d1fc641d5d0c82ac138/test/api/v3/integration/groups/GET-groups.test.js#L28-L40

So the two options are then to loop over every single guild and check if they don't contain `TAVERN_ID`, or to reset the database again using `await resetHabiticaDB();` to make sure that there are actually zero extra created guilds.

https://github.com/HabitRPG/habitica/blob/13a5b276e94e3e571e9a5d1fc641d5d0c82ac138/test/helpers/mongo.js#L36-L88

The first has the downside that it could make the tests take unnecessary long in the future, if more new guilds end up being created. The second might cause the tests afterwards to fail, so the tests would have to be moved to the end instead.

Which option would be the wisest to go for in this regard? Personally I think `await resetHabiticaDB();` is the cleaner option, since it makes sure that the Tavern documents becomes part of the guilds before the test.

website/server/controllers/api-v3/groups.js
paglias: I'd move the check

test/api/v3/integration/groups/GET-groups.test.js
paglias: I'd go with the first option and loop over the resulting array to make sure the tavern isn't there, there are few guilds and even if some are added it would take thousands of results to slow down the tests.

```
expect(guilds.find(g => g.id === TAVERN_ID)).to.be.undefined
```

^ something like this should work!

website/server/controllers/api-v3/groups.js
paglias: ah good point... I think the problem might go away when the check is moved inside Group.getGroups since you'll have to filter for TAVERN_ID in the two switch cases about `guilds` and `publicGuilds` so the tavern one should not be touched at all

AlexanderColen:
Ended up resolving the issues in the way that we discussed them. 😃 

Thanks for the clear explanation of what exactly the issues were so that it was an easy fix on my part. 👍

paglias:
@AlexanderColen I'm going to be able to look at this in about 1 week

paglias:
thanks @AlexanderColen ! I've added the blacksmith title to your profile and noted this PR (it'll go live by the end of the week) towards your 3rd tier

PR #12439, opened 2020-08-04
Scheduling summary line for team Dailies and To Do's

SabreCat:

### Changes

Implements [this Zeplin design](https://app.zeplin.io/project/5d372e492873d2b2d2e77534/screen/5d8a3d89c4ffa4166b5307e4) and related views.

**Previously**, on an individual user's copy of a Team Daily or To Do, various setting fields appeared but were disabled.
**Now**, these fields are instead hidden in this scenario, with a summary sentence in plain English instead describing the task's difficulty and schedule or due date (if any).


website/client/src/components/tasks/taskModal.vue
```diff
@@ -1244,6 +1269,23 @@ export default {
     selectedTags () {
       return this.getTagsFor(this.task);
     },
+    summarySentence () {
```
paglias, line 1247: this is not going to be translatable, right?

SabreCat: For this version, correct. There's simply no way we could account for all possible sentence structures, with this format! If we decide we want to support other locales than `en`, we'll probably need to come up with a simpler `label: value` layout. But less than 10% of Group Plans (as determined by looking at the language preference of the group owner) use anything but `en`, so that's what we're running with for MVP.

website/client/src/components/tasks/taskModal.vue
```diff
@@ -1334,6 +1376,104 @@ export default {
...
+    formattedDays (frequency, repeat, daysOfMonth, weeksOfMonth, startDate) {
+      let activeDays;
+      const dayStringArray = [];
+      switch (frequency) {
+        case 'weekly':
+          activeDays = keys(pickBy(repeat, value => value === true));
+          if (activeDays.length === 0) return ' on <strong>no days</strong>';
+          if (activeDays.length === 7) return ' on <strong>every day of the week</strong>';
+          dayStringArray.push(' on <strong>');
+          forEach(activeDays, (value, index) => {
```
paglias: can we use the native `activeDays.forEach` instead of importing an extra lodash module?

paglias (commented):
left just one small comment

website/client/src/components/tasks/taskModal.vue
SabreCat: done!

PR #12444, opened 2020-08-09
fix(banned words): fix partial matching of words containing diacritic…

### LINKED ISSUE #12309: swearword blocker incorrectly blocks some words containing non-English characters
The swearword blocker can prevent an innocent word being posted if the word contains:
- a non-English letter that the blocker interprets as a word boundary and
- a syllable that the blocker interprets as a swear word.

For example, the name "Assétou" cannot be posted in the Tavern or public guilds. The swearword blocker thinks that the three letters before the "é" character are a stand-alone word, and that word happens to be in the list of words that aren't allowed.

**To replicate the bug:**
1. Go to the Tavern on the website.
2. Try to make a post that contains "Assétou" (and optionally any other words).
3. Observe that your message is not posted and you see an error message (_"Oops! Looks like this post contains a swearword, religious oath, or reference to an addictive substance or adult topic (Assé). Habitica has users from all backgrounds, so we keep our chat very clean. Feel free to edit your message so you can post it!"_)
4. You can repeat this test as often as you need to. There's no penalty for triggering this particular type of blocked message error.

**The expected behaviour** is that "Assétou" should be able to be posted with no error message.

However it should not be possible to post a swear word followed by a space or punctuation. For example, `TESTPLACEHOLDERSWEARWORDHERE` is a swear word (see below for why). You should **not** be able to post any of the examples below nor others like them that use any other punctuation marks:
- " TESTPLACEHOLDERSWEARWORDHERE "
- "TESTPLACEHOLDERSWEARWORDHERE."
- "TESTPLACEHOLDERSWEARWORDHERE, "
- "(TESTPLACEHOLDERSWEARWORDHERE)"
That's currently working correctly (you currently cannot post those examples) and it needs to keep working like that.

You **should** be able to post this:
- "TESTPLACEHOLDERSWEARWORDHEREé"  (that's technically the same situation as the "Assétou" example)

I'm making this a high priority because it's not good that our swear word blocker is preventing legitimate non-English names being posted.

Below are details about how the swear word blocker is implemented.

---

The code that prevents swear words being posted is here:
https://github.com/HabitRPG/habitica/blob/679c8f725b8487b25d9d4d0470e8dd5716e64459/website/server/controllers/api-v3/chat.js#L174-L181
and here:
https://github.com/HabitRPG/habitica/blob/679c8f725b8487b25d9d4d0470e8dd5716e64459/website/server/controllers/api-v3/chat.js#L91-L93
and here:
https://github.com/HabitRPG/habitica/blob/679c8f725b8487b25d9d4d0470e8dd5716e64459/website/server/libs/stringUtils.js#L5-L19

None of those files contain the list of banned swear words, so it's safe to read all those files.

The list of banned swear words is in the `website/server/libs/bannedWords.js` file (CONTENT WARNING: it contains unpleasant words!). The top part of that file is safe to read - it has comments explaining the format of the file, but all you really need to know is that the swear words are stored in an array called `bannedWords`, with one swear word in each array element. The array is exported with `module.exports = bannedWords;` Here's the important parts of that `website/server/libs/bannedWords.js` file:
```
const bannedWords = [
  'TESTPLACEHOLDERSWEARWORDHERE',
  'TESTPLACEHOLDERSWEARWORDHERE1',
  // swear words are here
];

export default bannedWords;
```
'TESTPLACEHOLDERSWEARWORDHERE' and 'TESTPLACEHOLDERSWEARWORDHERE1' are actual words in the `website/server/libs/bannedWords.js` file. They behave just like real swear words so if you try to post them to the Tavern you'll see the swear word warning. They're in that file so that you can test the swear word blocker without needing to use a real swear word. Using either of those test words will not cause your chat permissions to be removed.

---

**NOTE:**

**You should not need to modify the `website/server/libs/bannedWords.js` file that contains the swear words. Please don't commit any changes to that file or let any swear words appear in your pull request or comments on this issue.**

Habitica tries to keep all of its public spaces as safe places for people of all ages and backgrounds, and that includes the issues and pull requests in our GitHub repositories. We try to avoid swearing here just like we do in Habitica itself.

If you need to refer to swear words in examples, or test code, etc, you can use `TESTPLACEHOLDERSWEARWORDHERE`. For example, if you do `git grep TESTPLACEHOLDERSWEARWORDHERE` you'll see that some tests use that word.

bigsee:
…s against banned words list (#12309)

Fixes #12309 

### Changes

Added Unicode normalization of input strings in the `getMatchesByWordArray()` function. This prevents diacritics from tripping up the regEx and resulting in false partial matches like the one reported by @Alys. This works since:

`TESTPLACEHOLDERSWEARWORDHEREé`

...is now tested against the banned words list as:

`TESTPLACEHOLDERSWEARWORDHEREe`

...rather than as:

`TESTPLACEHOLDERSWEARWORDHERE,,`

...which of course would, and should, fail.

----
UUID: 7e694dc6-6c80-4d4a-bc7d-264b6d5022a0

### DIFF
--- a/website/server/libs/stringUtils.js
+++ b/website/server/libs/stringUtils.js
@@ -4,8 +4,8 @@ export function removePunctuationFromString (str) {
 
 // NOTE: the wordsToMatch aren't escaped in order to support regular expressions,
 // so this method should not be used if wordsToMatch contains unsanitized user input
-export function getMatchesByWordArray (str, wordsToMatch) {
 
+export function getMatchesByWordArray (str, wordsToMatch) {
   // remove accented characters from the string, which would trip up the regEx
   // later on, by using the built-in Unicode normalisation methods
   // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize
--- a/website/server/libs/stringUtils.js
+++ b/website/server/libs/stringUtils.js
@@ -10,6 +10,7 @@ export function getMatchesByWordArray (str, wordsToMatch) {
   // later on, by using the built-in Unicode normalisation methods
   // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize
   // https://www.unicode.org/reports/tr15/#Canon_Compat_Equivalence
+  // https://unicode-table.com/en/#combining-diacritical-marks
 
   const normalizedStr = str.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
 
--- a/website/server/libs/stringUtils.js
+++ b/website/server/libs/stringUtils.js
@@ -5,11 +5,19 @@ export function removePunctuationFromString (str) {
 // NOTE: the wordsToMatch aren't escaped in order to support regular expressions,
 // so this method should not be used if wordsToMatch contains unsanitized user input
 export function getMatchesByWordArray (str, wordsToMatch) {
+
+  // remove accented characters from the string, which would trip up the regEx
+  // later on, by using the built-in Unicode normalisation methods
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize
+  // https://www.unicode.org/reports/tr15/#Canon_Compat_Equivalence
+
+  const normalizedStr = str.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
+
   const matchedWords = [];
   const wordRegexs = wordsToMatch.map(word => new RegExp(`\\b([^a-z]+)?${word}([^a-z]+)?\\b`, 'i'));
   for (let i = 0; i < wordRegexs.length; i += 1) {
     const regEx = wordRegexs[i];
-    const match = str.match(regEx);
+    const match = normalizedStr.match(regEx);
     if (match !== null && match[0] !== null) {
       const trimmedMatch = removePunctuationFromString(match[0]).trim();
       matchedWords.push(trimmedMatch);
--- a/test/api/unit/libs/stringUtils.test.js
+++ b/test/api/unit/libs/stringUtils.test.js
@@ -8,5 +8,10 @@ describe('stringUtils', () => {
       const matches = getMatchesByWordArray(message, bannedWords);
       expect(matches.length).to.equal(bannedWords.length);
     });
+    it('doesn\'t flag names with accented characters', () => {
+      const name = 'TESTPLACEHOLDERSWEARWORDHEREé';
+      const matches = getMatchesByWordArray(name, bannedWords);
+      expect(matches.length).to.equal(0);
+    });
   });
 });


website/server/libs/stringUtils.js
```diff
@@ -4,12 +4,20 @@ export function removePunctuationFromString (str) {
...
 // NOTE: the wordsToMatch aren't escaped in order to support regular expressions,
 // so this method should not be used if wordsToMatch contains unsanitized user input
+
 export function getMatchesByWordArray (str, wordsToMatch) {
+  // remove accented characters from the string, which would trip up the regEx
+  // later on, by using the built-in Unicode normalisation methods
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize
+  // https://www.unicode.org/reports/tr15/#Canon_Compat_Equivalence
+
+  const normalizedStr = str.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
```
paglias, line 15: could you point me to where the list of accented unicode characters can be found?

website/server/libs/stringUtils.js
```diff
@@ -4,12 +4,20 @@ export function removePunctuationFromString (str) {
 
 // NOTE: the wordsToMatch aren't escaped in order to support regular expressions,
 // so this method should not be used if wordsToMatch contains unsanitized user input
+
 export function getMatchesByWordArray (str, wordsToMatch) {
+  // remove accented characters from the string, which would trip up the regEx
```
paglias, line 9: @Alys what do you think about this fix where accented characters are simply ignored?

paglias (changes requested):
The code is looking good, could you add a unit or integration test to verify that words with accented characters can be posted now?

website/server/libs/stringUtils.js
Alys: I like this approach. I think it will work well for our needs. 

@bigsee Thanks for including those two URLS - they were useful.

website/server/libs/stringUtils.js
bigsee: Sure. Please see the Unicode table [here](https://unicode-table.com/en/#combining-diacritical-marks). You notice that the section I have linked to contains all the diacritical marks. 

My understanding is that our accented characters like é were being parsed as the diacritical mark then the standard letter (e.g. `[accent] e`), which broke the regex. I think the normalize function converts `[accent] e => e` under the hood.

The highlight section are the marks we are normalizing with this change: 

Screen Shot 2020-08-15 at 14 27 11

website/server/libs/stringUtils.js
bigsee: @paglias just FYI, I don't think they are technically ignored. before this change they were ignored, which is what causes the false flags. they are now converted (for validation only) into their non-accented form. If you're happy I'll get on those extra tests.

website/server/libs/stringUtils.js
bigsee: I think I should probably also add the link to the relevant unicode table section in the code comments, so I'll do that when I'm adding the new test.

paglias: thanks! adding a link to the table is definitely a good idea

website/server/libs/stringUtils.js
paglias: I said ignored because `.replace(/[\u0300-\u036f]/g, '');` doesn't replace the chars from u0300 to u036f with an empty char?

Although in this test it didn't seem to change anything:

Schermata 2020-08-17 alle 12 15 38

bigsee: I am only learning this stuff myself so please excuse me if I have something wrong or explain something with inaccurate terms. :) 

I think you are missing the 'NFD' argument in the normalize() function:

Screen Shot 2020-08-17 at 12 45 04

You are correct that the code replaces the chars from u0300 to u036f with an empty char. However, this Unicode range is a range of _diacritical marks_ rather than accented characters. So I would not expect any accented characters to get replaced. I think the accented character are somehow the _combination_ of the non-accented character with a diacritical mark...

...and I think that without this "normalization"/conversion first, `SOMESWEARWORDé` is evaluated as the split version => `SOMESWEARWORD [diacritical mark] e` somehow. So `str.normalize('NFD').replace(/[\u0300-\u036f]/g, '')` aims to remove those diacritical marks and effectively treat the input as one complete string again.

Not sure fully what is happening under the hood but that's my layperson's guess from what I read.

paglias: Oh you're right sorry! I thought NFD was the default parameter when using `normalize` so I skipped it. And unicode is something I definitely need to read more about, your explanation makes total sense :)

bigsee:
@shanaqui this is the one where I need to add a test. working on that now...

paglias:
thanks @bigsee , welcome to the 2nd tier :)

PR #12445, opened 2020-08-09
Add checks for profanity to profile updates

### LINKED ISSUE #11865: Display Name and Profile blurb editing should include checks for swear words and slurs

bigsee:

Fixes #11865 

### Changes

Based on some of the good work already done by @NickHolden in #12333 ...

1. Added checking display name for banned words
2. Added checking 'About' section for banned words and slurs
3. Added revocation of chat privileges in the event that a slur is used
4. Added the fix for #12309 (which is also relevant here).
5. Added new text for informing users that chat privileges have been revoked. Matches text used in group chat but updated to the correct context.
6. Refactored functions used to detect profanity and flag users (taking into account feedback from @paglias on #12333
7. Improved regEx to prevent false partial matches

----
UUID: 7e694dc6-6c80-4d4a-bc7d-264b6d5022a0


website/server/libs/user/index.js
```diff
@@ -102,6 +102,17 @@ function checkPreferencePurchase (user, path, item) {
...
 }
 
+async function checkNewInputForProfanity (user, res, newValue) {
+  if (stringContainsProfanity(newValue, 'slur') || stringContainsProfanity(newValue)) {
+    if (stringContainsProfanity(newValue, 'slur')) {
+      user.flags.chatRevoked = true;
+      await user.save();
+      throw new BadRequest(res.t('bannedSlurUsedInProfile'));
+    }
+    throw new BadRequest(res.t('displaynameIssueSlur'));
```
paglias: the string here should be changed maybe since it's not just about the display name anymore?

website/server/libs/user/index.js
```diff
@@ -102,6 +102,17 @@ function checkPreferencePurchase (user, path, item) {
   return _.get(user.purchased, itemPath);
 }
 
+async function checkNewInputForProfanity (user, res, newValue) {
+  if (stringContainsProfanity(newValue, 'slur') || stringContainsProfanity(newValue)) {
+    if (stringContainsProfanity(newValue, 'slur')) {
```
paglias: can we avoid calling `stringContainsProfanity()` twice and instead saving the result of the check the first time?

paglias (dismissed):
left some comments, looks like the existing tests are failing and we should add some new ones to check the new behavior

bigsee:
@paglias thank you. will take a look

bigsee:
@paglias apologies for the number of commits and, in advance, for the length of this message!

I really struggled with testing - I'm not sure what I'm doing wrong but I don't seem to get the same results locally as the test runner does in Github. For example:

Habitica failing tests

...are the results I am getting for `npm run test:api-v3:integration`. At least half of the tests fail, which have nothing to do with what I was working on. This made it hard to spot what was failing as a result of my changes. I ended up having to push each time to check individual ones failing in Github. Once I spotted them I could run exclusive tests locally at least.

I did look for a steer on testing in the 'Guidance for Blacksmiths', but all I could find was:

Screen Shot 2020-08-15 at 12 53 30

Should contributors run and wait for the whole of `npm run test` every time? Or just guess which ones apply to them, as I did? Not sure if there is a hard and fast rule, but perhaps we could update the docs to give a little guidance for testing n00bs like me... ☺️ 

Hopefully everything is OK now anyway. 🤞🏽

@Alys some tests were failing, I think, because we were actually testing cases we perhaps shouldn't have been. I have removed this, for example:

Incorrect tests for slur

...as I presumed from the issue discussion that users should be able to set `lateforclasses` or `madeinscunthorpe` as a username, which would have failed the test. 

Is that correct? Or should I reinstate that test?

bigsee:
@paglias scratch most of that - I just found more info on the [Using Your Local Install...](https://habitica.fandom.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API) page. Would it be OK to link from the testing section I highlighted above to this page?

I'm thinking "For more detailed information on testing your contribution before submission, see [Using Your Local Install to Modify Habitica's Website and API](https://habitica.fandom.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API).

Alys:
>  I just found more info on the Using Your Local Install... page. Would it be OK to link from the testing section I highlighted above to this page?
> I'm thinking "For more detailed information on testing your contribution before submission, see Using Your Local Install to Modify Habitica's Website and API.

@bigsee The "Tests for the Website and API Code" section in [Guidance for Blacksmiths](https://habitica.fandom.com/wiki/Guidance_for_Blacksmiths#Tests_for_the_Website_and_API_Code) links to [Using Your Local Install ...](https://habitica.fandom.com/wiki/Using_Your_Local_Install_to_Modify_Habitica%27s_Website_and_API) but if you think it would be helpful to add a brief note from the "Technology Stack" section too you're welcome to edit the page.

It may also be helpful to increase the emphasis for the existing links to "Using Your Local Install..." since they're clearly not obvious enough. If you do edit the Guidance page, you could (if you wanted to) review each of those links and add bolding or something to ensure they're highly noticeable.

website/server/libs/user/validation.js
```diff
@@ -1,13 +1,20 @@
...
+import bannedWords from '../bannedWords';
+import {
+  getMatchesByWordArray,
+  normalizeUnicodeString,
+  removePunctuationFromString,
+} from '../stringUtils';
 import forbiddenUsernames from '../forbiddenUsernames';
 
-const bannedSlurRegexs = bannedSlurs.map(word => new RegExp(`.*${word}.*`, 'i'));
+export function stringContainsProfanity (str, profanityType = 'bannedWord') {
```
paglias, line 13: I'm a bit worried about constructing the regexes here every time the function is invoked instead of doing it outside of the function like it was done before

paglias (commented):
I've left some comments, I think @Alys will know better about the tests you removed and whether they're actually needed or not

website/server/libs/user/index.js
bigsee: I didn't change this since it felt like a pretty major change. 

It's now about the display name and the 'about' section, both of which pertain the same screen (user profile modal). I suppose 'userProfileIssueSlur' would have been slightly more accurate but, per the docs, I felt like it was close enough in context not to warrant changing the json keys in all the translation files.

Would you like me to change anyway?

bigsee: Also, do I need to do anything to alert anyone to new/edited translations required?

paglias: Regarding the new / edited translations no other work is needed! They're handled automatically. I'll let @Alys decide on the string change

Alys: @bigsee Regarding renaming the `displaynameIssueSlur` string, thank you for considering that it's a big change but in this case it would be best to change it since the current wording may cause confusion when the name is fine but the About text is bad. You won't need to change any of non-English locales files, just delete the old string from the locales/en/settings.json file and change the string name in the tests and code that uses it (should be about four files in total I think). Our translators will then handle the non-English language files.

`bannedWordUsedInProfile` may be the best new name since (if I'm reading the code correctly) it's not actually used when there's a slur. The text should be something like "Your Display Name or About text contained inappropriate language."

website/common/locales/en/settings.json
```diff
@@ -211,5 +211,7 @@
     "mentioning": "Mentioning",
     "suggestMyUsername": "Suggest my username",
     "everywhere": "Everywhere",
-    "onlyPrivateSpaces": "Only in private spaces"
+    "onlyPrivateSpaces": "Only in private spaces",
+    "blurbIssueSlur": "The About text may not contain inappropriate language.",
+    "bannedSlurUsedInProfile": "Your display name or About text contained inappropriate language, and your chat privileges have been revoked."
```
Alys: Can you please change "display name" to "Display Name"? The convention is that it's always capitalised to emphasise that it has a specific meaning.
Also change "inappropriate language" to "a slur" so that there's no confusion about other kinds of bad words leading to an automatic mute - it's specifically slurs that do that.

Is `blurbIssueSlur` used anywhere? If not, it can be deleted.

Alys (dismissed):
@bigsee Sorry about my delay with this!

I've left a couple of comments. 

It's good that you removed a couple of the test cases - we should never have been testing for those specific texts.

bigsee:
@Alys no problem.

Just got back from holidays so letting you know that I'm on this. Will likely be spending some more time this weekend.

website/server/libs/user/validation.js
bigsee: @paglias Since we need two possible arrays of regexes - one for slurs and one for profanity - wouldn't moving it outside mean that we always construct one of the regex sets for no reason? Since the function only ever wants one. I have moved outside for now anyway and it might be more readable this way.

paglias:
Hi @bigsee , I won't be able to review the pr, cc'ing @SabreCat

bigsee:
@paglias thank you!

SabreCat (approved):

SabreCat:
Thanks much for this, @bigsee, and apologies for the delay in review! I've credited this toward your next contributor tier, and it'll go into our staging environment ahead of deployment to production next week.

bigsee:
@SabreCat no problem at all. It took too long on my side. Thank you for the review!

PR #12460, opened 2020-08-11
PR to fix: Correct handling for adding second @mention in Tavern and Guild chats

### LINKED ISSUE #12377: Correct handling for adding second @mention in Tavern and Guild chats
Scenario: If you write a message with a single @mention and later decide to @mention a second user, the dropdown list of names does not appear for the second @mention. 

Ideally, selecting the desired choice from dropdown menu should not lose focus of the text area. (Currently, selecting your choice requires you to re-select text area in order to continue typing).

This issue is a cleanup of #9694 with additional UI improvement suggestion. 

Amberlisi:
Creates a variable for the 'user-entry' text area, then once the user selects a mentionable user from the dropdown, focus is directed at the text area using that variable

Fixes #12377

### Changes
This PR addresses the second problem described in the issue, where the text area loses focus upon the user selecting an @mention option with a mouse-click. The first problem described in the issue was not applicable. It couldn't be repeated and does not appear to be a problem anymore.

Inside autoComplete.vue, in the select() function I've created a variable that grabs the textarea element that we want to focus on. Once the selection is made by the user, focus will be brought back to the textarea using that variable.

I don't like that the elements class name is being used in order to focus on the element... I feel like there must be a better way to focus on the correct element without the class 'user-entry' being hardcoded. Are there suggestions to a better method to do this?

----
UUID: 74ad4bd2-9de5-43f7-8efd-cfd684f47374


citrusella:
> The first problem described in the issue was not applicable. It couldn't be repeated and does not appear to be a problem anymore.

I'm no reviewer or even *much* of a coder so I always feel afraid to speak up on PRs, but out of curiosity, what did you try to replicate that first problem? The first bug as described in the issue sounds ambiguous but it's probably specifically referring to going backward into the post and @-ing a second user next to the first user or anywhere else earlier in the text, i.e. not at the end of the text where mentioning a second person *does* work. 

(That is to say 

`@thismentionDOESgetadropdown @thismentionDOESNOTgetadropdown This text typed before second mention`  
as opposed to  
`@thismentionDOESgetadropdown This text typed before second mention @thismentionalsoDOESgetadropdown`)

I tried to test out your code out of curiosity if it did fix it and I still had the issue regarding autocomplete not working on an additional mention inserted anywhere but the end of the text.

Amberlisi:
> > The first problem described in the issue was not applicable. It couldn't be repeated and does not appear to be a problem anymore.
> 
> I'm no reviewer or even _much_ of a coder so I always feel afraid to speak up on PRs, but out of curiosity, what did you try to replicate that first problem? The first bug as described in the issue sounds ambiguous but it's probably specifically referring to going backward into the post and @-ing a second user next to the first user or anywhere else earlier in the text, i.e. not at the end of the text where mentioning a second person _does_ work.
> 
> (That is to say
> 
> `@thismentionDOESgetadropdown @thismentionDOESNOTgetadropdown This text typed before second mention`
> as opposed to
> `@thismentionDOESgetadropdown This text typed before second mention @thismentionalsoDOESgetadropdown`)
> 
> I tried to test out your code out of curiosity if it did fix it and I still had the issue regarding autocomplete not working on an additional mention inserted anywhere but the end of the text.

Oh, I see now! I was only trying to add a second @ mention at the end of my text, I didn't think to go in the middle of text. Okay, I'll go back and take a look at that as well. Thanks for that!

I also went back and read @Alys first comment on #12377, which I now see I misunderstood completely :1st_place_medal: 

I don't want to close this PR, but it also isn't ready for review. Can the label on the PR be changed or something?

paglias:
@Amberlisi I'll change the label!

Regarding

> I don't like that the elements class name is being used in order to focus on the element... I feel like there must be a better way to focus on the correct element without the class 'user-entry' being hardcoded. Are there suggestions to a better method to do this?

You're right that it's not the best option and usually when working with Vue we try to limit as much as possible querying the DOM or changing it manually. Luckily there are `refs` which Vue provides and this should be the perfect use case https://vuejs.org/v2/guide/components-edge-cases.html#Accessing-Child-Component-Instances-amp-Child-Elements

Amberlisi:
> @Amberlisi I'll change the label!
> 
> Regarding
> 
> > I don't like that the elements class name is being used in order to focus on the element... I feel like there must be a better way to focus on the correct element without the class 'user-entry' being hardcoded. Are there suggestions to a better method to do this?
> 
> You're right that it's not the best option and usually when working with Vue we try to limit as much as possible querying the DOM or changing it manually. Luckily there are `refs` which Vue provides and this should be the perfect use case https://vuejs.org/v2/guide/components-edge-cases.html#Accessing-Child-Component-Instances-amp-Child-Elements

Yeah, I was trying to use refs to call it but couldn't figure out why it wasn't working. Then I realized I was trying to call it from the wrong file. So I've made that change. Looking into the rest of the issues still

SabreCat:
Looks like this was addressed in #13180. Closing this one for now; thanks for taking a stab at it, @Amberlisi!

PR #12488, opened 2020-08-23
fix some lint warnings

Alys:
This PR fixes some of the warnings that lint throws when you do `npm run lint` on a local install.

You may not want all of these fixes so tell me if I should remove any of them. Or feel free to cherry pick or modify this further yourself if you prefer. I've split this into commits based on each type of fix in case that helps. 

https://github.com/HabitRPG/habitica/commit/723fa38271173c8c005e0e8a2520292b9dd7e738 may be controversial - see the commit comment for why I did this.

There are still some warnings that I didn't fix but they involved larger changes than I felt comfortable making without spending too much time on it.

I did this because it's too hard to look through the full lint output to find errors and warnings that are related to my own code changes.

### DIFF
--- a/website/client/.eslintrc.js
+++ b/website/client/.eslintrc.js
@@ -14,8 +14,6 @@ module.exports = {
     // TODO find a way to let eslint understand webpack aliases
     'import/no-unresolved': 'off',
     'vue/no-v-html': 'off',
-    'vue/require-prop-types': 'off',
-    'vue/require-default-prop': 'off',
     'vue/html-self-closing': ['error', {
       html: {
         void: 'never',
--- a/website/client/.eslintrc.js
+++ b/website/client/.eslintrc.js
@@ -14,6 +14,8 @@ module.exports = {
     // TODO find a way to let eslint understand webpack aliases
     'import/no-unresolved': 'off',
     'vue/no-v-html': 'off',
+    'vue/require-prop-types': 'off',
+    'vue/require-default-prop': 'off',
     'vue/html-self-closing': ['error', {
       html: {
         void: 'never',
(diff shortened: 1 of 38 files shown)


website/client/.eslintrc.js
```diff
@@ -14,6 +14,8 @@ module.exports = {
     // TODO find a way to let eslint understand webpack aliases
     'import/no-unresolved': 'off',
     'vue/no-v-html': 'off',
+    'vue/require-prop-types': 'off',
```
paglias: I would like to keep these ones, we don't respect them often but they're a good practice

paglias:
@Alys I understand this can be annoying, my suggestions are:

1. If your code editor supports it use a plugin for eslint so that it shows errors while editing the code (there are for vscode, sublime text, ..)
2. Eslint has an option to hide warnings but vue-cli does not support it (https://github.com/vuejs/vue-cli/issues/4560) but you can still be used directly with `npx eslint --quiet "{tests,src}/**/*.{js,vue}" .` from `website/client`

I'm fine with silencing individual errors like the ones about alerts (thanks!) but I'd prefer not disabling rules all together

website/client/.eslintrc.js
Alys: No worries! If they're good practice we should keep them. I've removed the two "off" lines, and I'll make sure to follow them in the admin panel code.

Would you like me to create an issue for a contributor to fix the code that does't yet follow `vue/require-prop-types` and `vue/require-default-prop`? Perhaps in batches so it's not overwhelming to fix and review all of them at once?

paglias: Hmm I'd say an issue would be overwhelming, let's keep it like it's and I'll try to follow the convention more closely / remind it to contributors. For your PR I'd say the more important one is the `vue/require-prop-types` and setting `required` to props that are always necessary (it will make  `vue/require-default-prop` not required, it's mandatory only if the prop is not required)

PR #12496, opened 2020-08-25
Frontend support and adjustments of current banned words blocking code to support new implementation (Fixes #12405)

### LINKED ISSUE #12405: banned words (swear words, etc) should be allowed in guilds with a bannedWordsAllowed Group attribute
EDIT 2020-08-17 : Stages 1 and 2 have been completed. **Stage 3 is now ready to be done.**

In https://github.com/HabitRPG/habitica/pull/9253, we prevented swear words being posted to most public guilds. Some guilds are exempt (e.g., because religious guilds need to be able to mention terms that otherwise would be considered banned religious oaths).

The guilds that don't have swear word blocking are listed in [website/server/libs/guildsAllowingBannedWords.js](https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/guildsAllowingBannedWords.js) and the code that applies blocking for everything except those guilds is shown below.
We want to stop keeping the list of guilds in a file. Instead, we want moderators and staff to be able to mark a guild as exempt from swear word blocking with a checkbox in the guild's edit modal.

https://github.com/HabitRPG/habitica/blob/fef9c74f9b9ab9b49ae4a8f9380faa7007fb5535/website/server/controllers/api-v3/chat.js#L173-L180

---

This will need to be done in three stages:

### THIS HAS BEEN DONE - Stage 1. Backend support for marking a guild as exempt

* Adjust the [Group schema](https://github.com/HabitRPG/habitica/blob/develop/website/server/models/group.js) to add an attribute called something like `bannedWordsAllowed`
* The attribute must be optional.
* It should be a boolean (if true, banned words can be posted; if false or not specified, then the normal swear word blocking happens).
* I think that should be everything needed for this step. No actual code that uses the new field needs to be added yet.
* The pull request for this should be merged and deployed to production before Stage 2 is done.

###  THIS HAS BEEN DONE - Stage 2. Mark guilds as being exempt.

Admins will do this step (probably Alys).

A simple mongodb update script will be run to set `bannedWordsAllowed` to true for all the guilds currently listed in `guildsAllowingBannedWords.js`. The contributor who works on this issue won't need to write the script.

### Stage 3. Frontend support and adjusting the swear word blocking code

* Adjust the guild creation/edit form to add a checkbox that only moderators and staff can see.
   * Mods and staff can be identified by `"contributor.admin": true` - see the `isAdmin()` function in [website/client/src/components/groups/group.vue](https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/groups/group.vue)
   * NB: mods and staff already have the ability to edit any public guild.
* Adjust the API so that when a guild is [created](https://habitica.com/apidoc/#api-Group-CreateGroup) or [updated](https://habitica.com/apidoc/#api-Group-UpdateGroup):
   * `bannedWordsAllowed` is set to true if the checkbox in the creation/edit form is ticked. That should happen ONLY if an admin is creating/editing the guild.
   * If an admin is creating/editing the guild and the `bannedWordsAllowed` parameter is not specified, then it should be set to false.
   * If a non-admin user is creating/editing the guild, then `bannedWordsAllowed` should not be changed.
* Adjust the [code that prevents banned words from being posted](https://github.com/HabitRPG/habitica/blob/fef9c74f9b9ab9b49ae4a8f9380faa7007fb5535/website/server/controllers/api-v3/chat.js#L173-L180) so that it uses the guild's value of `bannedWordsAllowed` instead of the `guildsAllowingBannedWords.js` file.
* If the guild doesn't have any value for `bannedWordsAllowed`, then behave as if the value is false.
* Delete the file containing the list of hard-coded guilds ([website/server/libs/guildsAllowingBannedWords.js](https://github.com/HabitRPG/habitica/blob/develop/website/server/libs/guildsAllowingBannedWords.js)).

---

This is the new version of this old issue: https://github.com/HabitRPG/habitica/issues/8813 but everything you need should be in this new issue.

Scottd125:

Fixes #12405 

### Changes
The third and final stage for closing #12405 :)

- Provides front-end support for admins to allow/disallow guilds from using banned words when creating/updating guilds
- API changes for when a guild is created or updated to support the newly added `bannedWordsAllowed` field
- Adjustments to the code that prevents banned words from being posted to use the new `bannedWordsAllowed` field instead of the hard-coded guilds
- Removal of the `guildsAllowingBannedWords.js` file
- Adjustment to the banned words test which relied on the hard-coded guilds to the new implementation

----
UUID: 
9cbaf3a3-1f48-4557-9f0a-4d403df1f3eb


website/server/controllers/api-v3/groups.js
```diff
@@ -114,6 +114,16 @@ api.createGroup = {
   url: '/groups',
   middlewares: [authWithHeaders()],
   async handler (req, res) {
+    // Once the request body gets sanitized by the group variable
```
paglias: @Scottd125 could we keep the ability of setting `bannedWordsAllowed` only in updateGroup and not when creating it?

website/client/src/components/groups/groupFormModal.vue
```diff
@@ -94,6 +94,40 @@ label.custom-control-label(v-once) {{ $t('allowGuildInvitationsFromNonMembers')
...
+            v-model="workingGroup.bannedWordsAllowed"
+            class="custom-control-input"
+            type="checkbox"
+          >
+          <label
+            v-once
+            class="custom-control-label"
+            for="bannedWordsAllowed"
+          >{{ $t('bannedWordsAllowed') }}</label>
+          <div
```
paglias, line 117: v-once

website/client/src/components/groups/groupFormModal.vue
```diff
@@ -94,6 +94,40 @@ label.custom-control-label(v-once) {{ $t('allowGuildInvitationsFromNonMembers')
...
+            id="groupBannedWordsAllowedDescription"
+            class="icon"
+            :title="$t('bannedWordsAllowedDetail')"
+          >
+            <div
+              class="svg-icon"
+              v-html="icons.information"
+            ></div>
+          </div>
+          <b-tooltip
```
paglias, line 129: v-once

test/helpers/api-integration/v3/object-generators.js
```diff
@@ -121,15 +123,16 @@ async function _upgradeToGroupPlan (groupLeader, group) {
 // invitees: an array of user objects that correspond to the invitees of the group
 // leader: the leader user object
 // group: the group object
-export async function createAndPopulateGroup (settings = {}) {
+export async function createAndPopulateGroup (settings = {}, update = {}) {
   const numberOfMembers = settings.members || 0;
   const numberOfInvites = settings.invites || 0;
   const upgradeToGroupPlan = settings.upgradeToGroupPlan || false;
   const { groupDetails } = settings;
+  const { groupDetailsUpdate } = update;
```
paglias: is this needed? why can't `groupDetails ` be used?

paglias (dismissed):
Left some comments @Scottd125 but the direction is looking good, I'd also like to have a couple new tests showing that the `bannedWordsAllowed` setting can only be set by admins, thanks!

website/server/controllers/api-v3/groups.js
Scottd125: The original description of the issue mentioned adding it to both creating and updating the guild. However, I've updated it to just only be within updateGroup and removed it from createGroup

website/client/src/components/groups/groupFormModal.vue
Scottd125: Added

website/client/src/components/groups/groupFormModal.vue
Scottd125: Added

test/helpers/api-integration/v3/object-generators.js
Scottd125: Admittedly, the approach I was taking when updating the test was sloppy due to being quite unfamiliar with the code base. After I got more familiar with it, I noticed the approach I was taking was pretty silly. This has all been amended

website/client/src/components/groups/groupFormModal.vue
```diff
@@ -94,6 +94,46 @@ label.custom-control-label(v-once) {{ $t('allowGuildInvitationsFromNonMembers')
 // "allowGuildInvitationsFromNonMembers": "Allow Guild invitations from non-members",
         -->
       </div>
+      <div
+        v-if="isAdmin && workingGroup.id"
```
Scottd125, line 98: In relation to the request to only allow the ability to change bannedWordsAllowed within updateGroup. I've included this conditional statement to ensure the banned words section/checkbox is only displayed on the update guild form modal

paglias:
Thanks @Scottd125 ! The PR is working well!

@Alys any objection to this being merged?

Alys:
Can it go to delta first?

paglias:
@Alys it's on delta

Alys:
I've tested this on the delta website and it's looking good! I'd be happy for it to go to production.

Thanks @Scottd125 and thanks @paglias for putting it on delta.

Alys:
I've changed the labels to "... merge pending"

paglias:
thanks @Scottd125 ! noted towards your 2nd tier

PR #12497, opened 2020-08-28
Change spellRogueStealthMaxedOut and spellAlreadyCast to a clearer wo…

### LINKED ISSUE #12487: Stealth on-hover text confusing when all Dailies are completed / not due
Stealth's on-hover text is misleading if you have not cast the Skill today and you don't have any remaining Dailies due. This is the current text:

https://github.com/HabitRPG/habitica/blob/d0bc0dbe498adbe036e01c261d8eed12a68c4dc2/website/common/locales/en/spells.json#L79

Possibly it should be changed to make it more generic so that it applies in either case. E.g., "There is no need for you to cast this Skill because all Dailies are not due or completed or already Stealthed." Or leave out the part from "because" onwards to make it brief. Or something else entirely. :)

_I'm marking this as suggestion-discussion for a decision to be made about what it should say._

---

Below is a two-part bug report from @Gumnos (43e2413a-f440-43c8-b96c-85b93024488c) displaying the confusion the current message can cause (slightly edited for brevity):

_"I'm not sure if this is a bug or not, but my rogue's 'Stealth' skill has been grayed out for several days (maybe more, I don't really use it, so I just happened to notice it a couple days ago and then notice again today that it was still grayed out). The message says that I've already used it, but I haven't used it in...ever? I'm level 68 so it's not a matter of having received/unlocked it yet. And I have more than the required 45 mana. Any ideas what's going on? Thanks!"_

_"Ah, I think I figured out what's going on. I'd completed all my dailies for the day so there's nothing for it to 'hide'. Unchecking one of my dailies to show it as uncompleted re-activates my option to use Stealth. So it makes sense to disable it if there's nothing left to be stealthy from, but the message is misleading, telling me that I've already used it, not that I have nothing left to be stealthy from."_

mike-the-monk:
…rding

[//]: # (Issue #12487)
Fixes #12487

### Changes
[//]: # changed spellRogueStealthMaxedOut and spellAlreadyCast to "Using this skill won't have any additional effect." In testing, it works: if all Dailies are completed, Stealth is grayed out and the hover-text reads "Using this skill won't have any additional effect." If instead the player already cast Stealth, the hover-text will mention that the skill was already cast. If Chilling Frost is used, then it's grayed out and the hover-text reads "Using this skill won't have any additional effect."

----
UUID: 0034eb14-b4d8-494e-8386-d3f33cff7922

### DIFF
--- a/website/client/package-lock.json
+++ b/website/client/package-lock.json
@@ -6698,47 +6698,6 @@
             "has-flag": "^3.0.0"
           }
         },
-        "vue-loader-v16": {
-          "version": "npm:vue-loader@16.0.0-beta.5",
-          "resolved": "https://registry.npmjs.org/vue-loader/-/vue-loader-16.0.0-beta.5.tgz",
-          "integrity": "sha512-ciWfzNefqWlmzKznCWY9hl+fPP4KlQ0A9MtHbJ/8DpyY+dAM8gDrjufIdxwTgC4szE4EZC3A6ip/BbrqM84GqA==",
-          "optional": true,
-          "requires": {
-            "@types/mini-css-extract-plugin": "^0.9.1",
-            "chalk": "^3.0.0",
-            "hash-sum": "^2.0.0",
-            "loader-utils": "^1.2.3",
-            "merge-source-map": "^1.1.0",
-            "source-map": "^0.6.1"
-          },
-          "dependencies": {
-            "chalk": {
-              "version": "3.0.0",
-              "resolved": "https://registry.npmjs.org/chalk/-/chalk-3.0.0.tgz",
-              "integrity": "sha512-4D3B6Wf41KOYRFdszmDqMCGq5VV/uMAB273JILmO+3jAlh8X4qDtdtgCR3fxtbLEMzSx22QdhnDcJvu2u1fVwg==",
-              "optional": true,
-              "requires": {
-                "ansi-styles": "^4.1.0",
-                "supports-color": "^7.1.0"
-              }
-            },
-            "has-flag": {
-              "version": "4.0.0",
-              "resolved": "https://registry.npmjs.org/has-flag/-/has-flag-4.0.0.tgz",
-              "integrity": "sha512-EykJT/Q1KjTWctppgIAgfSO0tKVuZUjhgMr17kqTumMl6Afv3EISleU7qZUzoXDFTAHTDC4NOoG/ZxU3EvlMPQ==",
-              "optional": true
-            },
-            "supports-color": {
-              "version": "7.2.0",
-              "resolved": "https://registry.npmjs.org/supports-color/-/supports-color-7.2.0.tgz",
-              "integrity": "sha512-qpCAvRl9stuOHveKsn7HncJRvv501qIacKzQlO/+Lwxc9+0q2wLyv4Dfvt80/DPn2pqOBsJdDiogXGR9+OvwRw==",
-              "optional": true,
-              "requires": {
-                "has-flag": "^4.0.0"
-              }
-            }
-          }
-        },
         "wrap-ansi": {
           "version": "6.2.0",
           "resolved": "https://registry.npmjs.org/wrap-ansi/-/wrap-ansi-6.2.0.tgz",
--- a/website/client/package-lock.json
+++ b/website/client/package-lock.json
@@ -6555,6 +6555,47 @@
             "has-flag": "^3.0.0"
           }
         },
+        "vue-loader-v16": {
+          "version": "npm:vue-loader@16.0.0-beta.5",
+          "resolved": "https://registry.npmjs.org/vue-loader/-/vue-loader-16.0.0-beta.5.tgz",
+          "integrity": "sha512-ciWfzNefqWlmzKznCWY9hl+fPP4KlQ0A9MtHbJ/8DpyY+dAM8gDrjufIdxwTgC4szE4EZC3A6ip/BbrqM84GqA==",
+          "optional": true,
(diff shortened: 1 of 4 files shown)


website/client/package-lock.json
```diff
@@ -20923,78 +20964,6 @@
...
-          "version": "7.1.0",
-          "resolved": "https://registry.npmjs.org/supports-color/-/supports-color-7.1.0.tgz",
-          "integrity": "sha512-oRSIpR8pxT1Wr2FquTNnGet79b3BWljqOuoW/h4oBhxJ/HUbX5nX6JSruTkvXDCFMwDPvsaTTbvMLKZWSy0R5g==",
-          "optional": true,
-          "requires": {
-            "has-flag": "^4.0.0"
-          }
-        }
-      }
-    },
```
mike-the-monk: I don't recall actively changing any of this (i.e. website/client/package-lock.json), not sure why these changes are here or what they are, or how to fix them if I need to. I'm very new at this.

mike-the-monk:
Well, it looks like I should've read Guidance for Blacksmiths better, I should've attached screenshots. Now that I've opened a pull request, I don't know how to set up the local test account to get those screenshots.

website/common/locales/en/spells.json
```diff
@@ -36,7 +36,7 @@
   "spellRogueStealthText": "Stealth",
   "spellRogueStealthNotes": "With each cast, a few of your undone Dailies won't cause damage tonight. Their streaks and colors won't change. (Based on: PER)",
   "spellRogueStealthDaliesAvoided": "<%= originalText %> Number of dailies that will be avoided: <%= number %>.",
-  "spellRogueStealthMaxedOut": "You have already avoided all your dailies; there's no need to cast this again.",
+  "spellRogueStealthMaxedOut": "Using this skill won't have any additional effect.",
```
paglias: the two string now have the same text so maybe we can remove one of them?

paglias:
In this case screenshots should not be needed, paging @Alys to check the text

For the `package-lock.json` changes you can, if you're familiar with git, restore the version from the develop branch

```
git fetch
git checkout origin/develop -- package-lock.json
```

But if you're not familiar with it no worries, we can fix it when merging the pr

Alys (approved):
The new wording looks correct to me!

mike-the-monk:
@paglias – I've never done that before and I'm not confident I do it correctly. I'll leave it to someone else to do and the pull request is merged

paglias (changes requested):
I've left one final comment, one of the two strings are redundant and we should remove it (spellRogueStealthMaxedOut) from the locales files and also change it with spellAlreadyCast where it's being used

website/common/locales/en/spells.json
mike-the-monk: Yeah, makes sense to me

paglias: @Mike-Antonacci the way to do this is to remove the unused string from the `website/common/locales/en/spells.json` file and then to change all the instances where it's used across the site with the other one

mike-the-monk: @paglias I deleted the unused string, and it looks like all instances are for other languages, which seems odd to me, I would've expected that this string would be used at least somewhere. `git grep "spellRogueStealthMaxedOut"` only showed other languages, for example `website/common/locales/ar/spells.json:`) but I know it's not showing all the languages. Question: does `git grep` only show some of the places the string is used? If yes, how do I get around that? Is it possible for me to exclude the languages (or the entire folder `website/common/locales`) from `git grep`?

I also successfully did `git checkout origin/develop -- package-lock.json` as you said.

I thought I committed the changes, and I thought I pushed them, but it said `no changes added to commit (use "git add" and/or "git commit -a")` and I don't know what that means or how to resolve it. I think this is actually two questions:
1) how do I successfully commit the `package-lock.json` changes?
2) how do I successfully commit deleting `spellRogueStealthMaxedOut`?

Thank you for your patience, this is my very first pull request.

paglias: @Mike-Antonacci what does `git status` says?

paglias: The idea is that you remove the line for `spellRogueStealthMaxedOut` from the english file, git status will tell you that a file was changed, then you can commit the change with `git commit -am'an informative message'` and push it

mike-the-monk: `git status` says `no changes added to commit (use "git add" and/or "git commit -a")`

mike-the-monk: I made a commit but it looks like it didn't have all the changes I wanted. I will work on this later today.

paglias: It's possible that if git status says there are no changes that it got added in a different commit, otherwise it would see the difference between what's stored in git and the saved version of the file on your computer

Unfortunately it's a bit difficult to debug these problems from github comments so if you prefer I'm happy to apply the changes here

mike-the-monk: Fully agreed that it's difficult to debug via comments. If you could apply the changes yourself, that would be great. Thank you for that, and thank you for all you do for Habitica

paglias: the string has been removed, I'm not sure why it's still showing the diffs in the package-lock files because they're exactly the same as `develop`

paglias: not sure how but it got fixed 🎉 The PR should be ready now, I'll merge it tomorrow

paglias:
Thanks @Mike-Antonacci ! I've added a Blacksmith title to your profile and noted this PR

PR #12499, opened 2020-08-30
WIP optional moderator approval for new accounts to post chat messages

Alys:
partial fix for https://github.com/HabitRPG/habitica/issues/8063

WIP
Method used is still under consideration and may be deemed unsuitable.
Needs tests.

### DIFF
--- a/website/server/models/user/schema.js
+++ b/website/server/models/user/schema.js
@@ -1,5 +1,6 @@
 import mongoose from 'mongoose';
 import validator from 'validator';
+import nconf from 'nconf';
 import shared from '../../../common';
 import { // eslint-disable-line import/no-cycle
   getDefaultOwnedGear,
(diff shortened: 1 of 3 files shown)


website/server/models/user/schema.js
```diff
@@ -257,8 +262,8 @@ export default new Schema({
       $type: Schema.Types.Mixed,
       default: () => ({}),
     },
-    chatRevoked: Boolean,
-    chatShadowMuted: Boolean,
+    chatRevoked: { $type: Boolean, default: NEW_USER_CHAT_MODE === 'muted' },
```
paglias, line 265: I think it's a smart approach but given the fields didn't have a default value when `NEW_USER_CHAT_MODE` will be set to muted or shadowmuted it won't affect only new users but all the old ones as well

paglias (changes requested):

website/server/models/user/schema.js
Alys: Oh heck. That's a really good point, thanks for picking that up. My existing local test accounts already had false values from previous testing for other changes. :-/

Do you think it's worth doing a migration to set false values for all existing users, or should I abandon this approach? I'm okay with whichever option you think is best. I'm thinking we probably don't want the hassle of a migration.

I guess I could change the actual code that creates a new account so that the flag is set there. That may be a better approach.

paglias: If we need this quickly setting the value when the user is created will be easier and we already do this in a lot of cases either in the signup route or in here https://github.com/HabitRPG/habitica/blob/aa1ea74daa5ff7644d861d8d41e7a4f84325be89/website/server/models/user/hooks.js#L127

Alys:
I'm closing this to get it out of the way. I'll reopen it if I have time to work on it again.

Issue #10008, opened 2018-02-16
Add alt text on navigation and header items


shanaqui:
Much of the text on Habitica's website either has no alt text or is not allowing screen readers to read the alt text. This is a serious accessibility issue as it means that anything that is represented solely by a graphic is essentially invisible to blind users of the site. 

In some cases there is no alt text, but there is a tooltip. However, the screenreader will not read out the text unless you manage to get your cursor over the tooltip -- for which you have to know it's there.

The following screenshots illustrate just some of the areas I found on the site that are lacking alt text. Red boxes indicate areas "invisible" to the screenreader. Tested using NVDA, a free screenreader available online [here](https://www.nvaccess.org/). 

We should probably discuss what we want the alt text to read and what is most helpful/appropriate in various cases, if this is not simply a display issue. There is some advice on how to form useful alt text [here](https://webaim.org/techniques/alttext/).

![screenshot one](https://user-images.githubusercontent.com/8144640/36328504-a7c67c56-1362-11e8-9a41-db32da1a3d5b.png)
![screenshot two](https://user-images.githubusercontent.com/8144640/36328503-a7af10ac-1362-11e8-8bc0-75f61eed5875.png)
![screenshot three](https://user-images.githubusercontent.com/8144640/36328502-a79763e4-1362-11e8-8860-c44e13970496.png)
![screenshot four](https://user-images.githubusercontent.com/8144640/36328500-a77d83de-1362-11e8-9c2b-470d70b34f77.png)
![screenshot five](https://user-images.githubusercontent.com/8144640/36328499-a763a02c-1362-11e8-89d3-a43ec865e7df.png)
![screenshot six](https://user-images.githubusercontent.com/8144640/36328498-a74c49b8-1362-11e8-9785-50bac4185321.png)
![screenshot seven](https://user-images.githubusercontent.com/8144640/36328497-a7319fb4-1362-11e8-851e-c4ed63accac7.png)

Feywood:
At a quick glance, I'm not seeing any alt text tags which is what I believe most screen readers would grab.
The bronze_small is in title tags, but I'm not seeing those either in eg the gem or gold icons in the top toolbar.

shanaqui:
[...]

PR #12500, opened 2020-08-30
Add alt to img tag

projectormato:
Related #10008

### Changes
- Add alt
  - `communityGuidelines.vue`
  - `features.vue`
  - `home.vue`
  - `404.vue`
  - `groupPlans.vue`

- Add `""` for the alt text to Any image that doesn't give obvious information (i.e., any image that's just for decoration).

Not attached to images that are difficult to attach.

----
UUID: df9dbe38-5bb5-4d23-9ba5-b5e11ed03c86

### DIFF
--- a/website/client/src/components/static/communityGuidelines.vue
+++ b/website/client/src/components/static/communityGuidelines.vue
@@ -7,7 +7,10 @@
       {{ $t('commGuideHeadingWelcome') }}
     </h2>
     <div class="media align-items-center">
-      <img src="~@/assets/images/community-guidelines/intro.png">
+      <img
+        src="~@/assets/images/community-guidelines/intro.png"
+        :alt="$t('commGuideHeadingWelcome')"
+      >
       <div class="media-body">
         <p v-html="$t('commGuidePara001')"></p>
         <p v-html="$t('commGuidePara002')"></p>
@@ -23,7 +26,10 @@
         <p v-html="$t('commGuidePara015')"></p>
         <p v-html="$t('commGuidePara016')"></p>
       </div>
-      <img src="~@/assets/images/community-guidelines/publicSpaces.png">
+      <img
+        src="~@/assets/images/community-guidelines/publicSpaces.png"
+        :alt="$t('commGuideHeadingInteractions')"
+      >
     </div>
     <ul>
       <li v-html="$t('commGuideList02A')"></li>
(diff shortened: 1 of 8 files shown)


projectormato:
@paglias 
I couldn't come up with a good name for `backCorner.png` of `communityGuidelines.vue`. please advice?
Please tell me if you find a strange name.

Alys:
@projectormato  Assuming that's the image shown below, then the best option may be to have an empty string as the alt text (i.e., two quote marks: `""`) so that when someone is using a screen reader, nothing is read out for that image. That's the standard approach for alt text for images that don't give any useful information, and I would guess that it's best for efficiency when having the page read to you.

The other alternative would be to describe the scene in the image (e.g., "A Habitican sits alone at a table with a drink, watching shadowy hooded figures have a conversation at a table in the back corner of the room.")  That would be great for adding flavour to the Guidelines for visually impaired users. My feeling though is that since the Guidelines are already quite long and since the important thing is to focus on the rules described in them, the empty string approach is likely to be better for usability.

back corner

projectormato:
@Alys 
Thank you for advice!
I think this method can be used for images that are difficult to name!

Alys:
@projectormato yes, probably! Any image that doesn't give obvious information (i.e., any image that's just for decoration) could have `""` for the alt text.

projectormato:
I tried working on it.
Please let me know if there is a mistake in the implementation!

website/client/src/components/static/communityGuidelines.vue
```diff
@@ -7,7 +7,10 @@
       {{ $t('commGuideHeadingWelcome') }}
     </h2>
     <div class="media align-items-center">
-      <img src="~@/assets/images/community-guidelines/intro.png">
+      <img
+        src="~@/assets/images/community-guidelines/intro.png"
+        :alt="$t('commGuideHeadingWelcome')"
```
paglias, line 12: Hmm I'm not 100% sure this file needs alt tag, the info added to the alt tags is already part of the text and that can be read with screen readers, I'm not sure duplicating it is a good idea . Opinions @Alys @Tressley ?

paglias:
@projectormato sorry about this but I got some doubt on how useful it's to add alt tags to some of the static pages, for now could we focus on the navigation elements described in https://github.com/HabitRPG/habitica/issues/10008, so mostly icons and images that are used to explain something across the website

paglias:
Actually as mentioned by @veeeeeee in the issue, for now:

> This is a good first issue for contributors **particularly for the header elements. For anything below the header, please hold off as staff would like to make sure the taxonomy is handled logically and consistently across the board.**

projectormato:
@paglias 
I understand.
Is there a img that should add alt text in now PR?

paglias:
@projectormato sorry for the delay! After talking with @Tressley for now we would like to focus only on the items that are part of the header of the website for example the gems, gold, user, ... icons

projectormato:
@paglias I understand!
Does that mean adding alt to SVG in the site header?
I want to try it!

PR #12522, opened 2020-08-31
Correctly filter tasks in challenges and remove filters where unhelpful

### LINKED ISSUE #12495: filters on Challenge page for Habits, To-Dos, Rewards don't work (sort of works for Dailies)
From @Alys in #10771 

As described in detail below, most of the normal task filters (due, completed, etc), aren't of much use when looking at a Challenge's main page ([e.g., this page](https://habitica.com/challenges/14cf33bb-1e97-4a31-8f3d-49c0f9781d3a)). The fixes described below should be done.

This does NOT apply to the screens you see when you are looking at an individual member's Challenge progress - those screens should keep all the filter options because they help the Challenge owner assess the player's progress.

Below are the problems that exist for filters. The bolded options are my preferred suggestions for fixes. There's been no objections to them since I bolded them, so they're probably the best fixes to do.

- [ ] For Habits, the "Weak" and "Strong" filters don't work (they show all Habits just like the "All" filter does). <strike>They should either be modified so that they show appropriate Habits based on the progress from all Challenge participants or</strike> **the "All", "Weak", and "Strong" filters should be removed**.

- [ ] For To-Dos, the "Scheduled" filter doesn't work (it shows all To-Dos). I think this one is actually useful so **I suggest keeping the "All" and "Scheduled" filters and making the "Scheduled" one show only To-Dos that have a due date.**

- [ ] For To-Dos, <strike>I'm not sure if the "Completed" filter works. If it did work, you'd want it to show To-Dos that ALL Challenge participants had completed. If that's too difficult,</strike> **the "Completed" To-Dos filter should be removed from the page**.

- [ ] For Rewards, the "All", "Custom", and "Wishlist" filters all show the same thing (all custom rewards). Wishlist rewards don't exist for Challenges so **the filters should be removed for Rewards**.

- [ ] For Dailies, the filters are working correctly except that they filter Dailies into due and not due based on the time zone of the person who created the Challenge, not the person looking at it, so it can be misleading. <strike>I'm not sure if we should try to do something about this, or leave it as it is so that at least the Challenge owner can use it, or</strike> **remove the filters for Dailies**.

---

Anyone testing a proposed fix should check that:
- [ ] the fix works on a Challenge's main page (e.g., a page with a URL like this: https://habitica.com/challenges/14cf33bb-1e97-4a31-8f3d-49c0f9781d3a)
- [ ] the fix does NOT change the screen that shows an individual Challenge participant's progress (that screen should still have all the filters)
- [ ] the fix does NOT change the user tasks page (https://habitica.com/)

kirsty-tortoise:
Fixes #12495

### Changes
 * On challenge pages, the filters for Habits, Dailies and Rewards have been removed.
 * On challenge pages, the filters for Todos have been edited so the only filters are "All" and "Scheduled" and these filters work correctly.
 * Filters on an individual Challenge page are as before and are now all working correctly (previously some, e.g. habits strong/weak did not appear to work).
 * The Due filters on all pages now use the "isDue" property of the task, rather than calculating it internally. I can change this back if necessary, but it is definitely the simpler way to do things. However it might not take users preferences fully into account and might break at cron if tasks aren't reloaded - I need to check this but would like to ask first.

----

### DIFF
--- a/website/client/src/store/actions/tasks.js
+++ b/website/client/src/store/actions/tasks.js
@@ -114,7 +114,7 @@ export async function create (store, createdTask, isUserTask) {
     const tasksArr = store.state.tasks.data[`${taskRes.type}s`];
     const taskDataIndex = tasksArr.findIndex(t => t._id === taskRes._id);
     if (taskDataIndex !== -1) {
-      Object.assign(tasksArr[taskDataIndex], taskRes);
+        tasksArr[taskDataIndex] = { ...tasksArr[taskDataIndex], ...taskRes };
     }
   });
 }
--- a/website/client/src/store/actions/tasks.js
+++ b/website/client/src/store/actions/tasks.js
@@ -1,8 +1,8 @@
 import axios from 'axios';
+import Vue from 'vue';
 import compact from 'lodash/compact';
 import omit from 'lodash/omit';
 import { loadAsyncResource } from '@/libs/asyncResource';
-import { shouldDo } from '@/../../common/script/cron';
 
 export function fetchUserTasks (store, options = {}) {
   return loadAsyncResource({
@@ -85,14 +85,8 @@ function sanitizeChecklist (task) {
   }
 }
 
-function updateIsDue (store, task, isUserTask) {
-  if (isUserTask) {
-    task.isDue = shouldDo(new Date(), task, store.state.user.data.preferences);
-  }
-}
-
 // Supply an array to create multiple tasks
-export async function create (store, createdTask, isUserTask) {
+export async function create (store, createdTask) {
   // Treat all create actions as if we are adding multiple tasks
   const payload = Array.isArray(createdTask) ? createdTask : [createdTask];
 
@@ -101,7 +95,6 @@ export async function create (store, createdTask, isUserTask) {
     const list = store.state.tasks.data[type];
 
     sanitizeChecklist(t);
-    updateIsDue(store, t, isUserTask);
 
     list.unshift(t);
     store.state.user.data.tasksOrder[type].unshift(t._id);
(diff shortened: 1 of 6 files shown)


kirsty-tortoise:
It seems to be working okay with custom day starts etc

paglias:
@kirsty-tortoise thanks for the PR! There's just one minor issue that I found:

If you add a daily, something weird can happen:

- If the `All` filter is active everything works as expected
- If the `Due` one is active instead and you create a task it does not show up until you change filter and then go back to it or add a second task
- With the `Not due` filter it happens the opposite: if you add a task (due by default) it appears there even though it should not until you add a new one or change filter

After this issue is fixed I can merge it :)

kirsty-tortoise:
@paglias Thanks for spotting this! I've hopefully fixed it now, I think what happened was that the isDue property was not automatically added on the client side at task creation. It is then added on the server side and pushed back to the client, which adds this to its task info on the client side, but then Vue didn't automatically update since it does not notice new properties.

I've changed the code so that it is added on the client side at task creation and when edited, which should hopefully fix the problem.

website/client/src/store/actions/tasks.js
```diff
@@ -94,6 +101,7 @@ export async function create (store, createdTask) {
     const list = store.state.tasks.data[type];
 
     sanitizeChecklist(t);
+    updateIsDue(store, t, isUserTask);
```
paglias: @kirsty-tortoise I was thinking, to avoid moving too much logic to the client what we could do, to avoid new properties not being tracked by vue, is to (like it's done elsewhere) change the task object completely instead of just assigning new properties when the server response comes back, so here at line 114 for example:

```
    const tasksArr = store.state.tasks.data[`${taskRes.type}s`];
    const taskDataIndex =tasksArr.findIndex(t => t._id === taskRes._id);
    if (taskDataIndex !== -1) {
        tasksArr[taskDataIndex] = { ...taskData, ...taskRes }; // works like Object.assign but replaces the object
    }
```

how is that looking?

kirsty-tortoise: Seems very sensible, I'll change that.

kirsty-tortoise: So I tried this and it seems like it is working even less well than Object.assign - now it isn't noticing any change at all, not just not noticing new properties being added. 😕 Unless I'm misunderstanding what taskData is supposed to be (I don't seem to have a variable called that).

paglias: could you push the commit?

paglias:
```
    const tasksArr = store.state.tasks.data[`${taskRes.type}s`];
    const taskDataIndex =tasksArr.findIndex(t => t._id === taskRes._id);
    if (taskDataIndex !== -1) {
        tasksArr[taskDataIndex] = { ...tasksArr[taskDataIndex], ...taskRes }; // works like Object.assign but replaces the object
    }
```

there was an error in taskData

kirsty-tortoise: I've pushed it. It still works fine until I remove line 129 which updates isDue on the client side, and then I seem to be getting the same problem as before with Vue as far as I can tell, so it doesn't seem to be helping with having it properly tracked.

paglias: Sorry, could you try one final thing, to use the code in the last comment before yours here? for the code in lines 114-118 ? If it doesn't work no worries, I'll give it a look or revert to your previous version

kirsty-tortoise: Oops sorry - I changed it back for testing purposes and I thought I had posted your code in lines 114-118. I'll change it back in a second. I think it will still work, but if I remove line 129 then it stops working again.

kirsty-tortoise: I hope that's what you meant?

paglias: @kirsty-tortoise sorry it took a while but turns out the problem exists only when creating a new task because Vue doesn't recognize new properties (it'll get fixed in Vue 3), I pushed a commit directly https://github.com/HabitRPG/habitica/pull/12522/commits/cbb8be8ea46cacf3366e23dedd28b1502249eeae

paglias:
Noted towards your 4th tier!

Issue #9957, opened 2018-02-05
Remove unused I18N strings


SabreCat:
### Description
[//]: # (Describe bug in detail here. Include screenshots if helpful.)

As reported by @paglias in our internal bug tracking doc:

> lots of unused / duplicate i18n strings. Should write an automated scripti for this.

Especially after the site redesign, there are many I18N strings that are not actually used. We should hunt these down and remove them to clean up our repo and reduce the burden on our Transifex volunteers.

Related: #9210 

### Progress

- [ ] achievements.json
- [x] backgrounds.json
- [x] challenge.json
- [x] character.json
- [x] communityGuidelines.json
- [x] content.json
- [x] contrib.json
- [x] death.json
- [x] defaultTasks.json
- [x] faq.json #12593
- [x] front.json
- [x] gear.json
- [x] generic.json
- [x] groups.json
- [x] inventory.json
- [x] limited.json #12567
- [x] loginIncentives.json #12566
- [x] maintenance.json #12565
- [x] merch.json #12564
- [x] messages.json #12563
- [x] noscript.json #12562
- [x] npc.json #12561
- [x] overview.json
- [x] pets.json #12559
- [x] quests.json #12558
- [x] questsContent.json
- [x] rebirth.json #12553
- [x] settings.json #12552
- [x] spells.json - no unused strings
- [x] subscriber.json #12548
- [x] tasks.json #12526
- [x] testing.json - no unused strings
- [ ] Staff: write script to remove strings removed previously from English locales

Alys:
Copying viirus's comment from https://github.com/HabitRPG/habitica/issues/9210#issuecomment-362802915 :

"The apps don‘t use any of the strings from this repo directly. They only use the content and response strings from the api"

Therefore when searching for where strings are used, we need to search only the Habit.RPG/habitica repo

Alys:
In https://github.com/HabitRPG/habitica/issues/9210#issuecomment-354697812, @Inshabel said they'd like to work on this so I've marked it as in progress for them.

Alys:
A comment from @SabreCat in https://github.com/HabitRPG/habitica/issues/9210 :

[...]

PR #12526, opened 2020-09-02
Removed unused I18N strings from the `tasks.json` file (related to issue #9957)

RaitheOfDureya:

This PR fixes part of issue #9957

### Changes
I applied RegExp Search (e.g., ```\"(.+)\" , \bclearCompleted\b ```, ...) to find the strings that were used outside the `locales` folder and removed those without matches.

For each string (or blocks of strings) removed I made a commit, so that the proposed changes can be reversed if necessary without major impacts.

P.S.: Although the strings `taskAlias`, `taskAliasPopover`, `taskAliasPlaceholder` and `taskAliasPopoverWarninge` didn't return any results I decided not to remove them because I wasn't sure if they were being used in "3rd party integrations":
```javascript
"taskAliasPopoverWarning": "WARNING: Changing this value will break any 3rd party integrations that rely on the task alias.".
```

----
UUID:6544e17f-237b-4915-ae7e-47744c0dc1bd


paglias:
Thanks @RaitheOfDureya , the PR is very well organized, I'll give it a look in the next few days.

Regarding the alias-related strings I'm pretty sure they're not used right now, the only doubt I have is if we remove them, will Weblate keep them in history so if we add them back again they old translated version will be used again @SabreCat ?

Also @SabreCat, given that Weblate doesn't seem to remove strings from the other locales when they're removed from the English one, I think we should remove them in a PR, will that cause any conflict?

paglias:
Also @RaitheOfDureya unfortunately, especially on the client some strings will not be used directly with their full key, for example some might be used by adding two strings together so they won't show up when searching for the full key, it's something to keep in mind and that we should check for

RaitheOfDureya:
Thanks for the feedback, @paglias 
Since it is possible that I have removed some strings that are used concatenated with others, I will wait for the review before continuing with the removal from the other files. This way I can adapt the procedure to cover the other cases that I was not aware of.

website/common/locales/en/tasks.json
```diff
@@ -16,9 +12,6 @@
   "theseAreYourTasks": "These are your <%= taskType %>",
   "habit": "Habit",
   "habits": "Habits",
-  "newHabit": "New Habit",
-  "newHabitBulk": "New Habits (one per line)",
-  "habitsDesc": "Habits don't have a rigid schedule. You can check them off multiple times per day.",
```
paglias, line 21: this one, alongside `todosDesc`, `dailysDesc` and `rewardsDesc` is still used here https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/tasks/column.vue#L77

I think an "easy" way to check for these types of usages is:

- on the client translated strings are always generated with a call to `$t(stringName)`, so if you search for `$t(` followed by a character that is not `'` it means that the string key is being dynamically created (either by passing a variable with the string key or by constructing a string using ` \` `
- for the server / common code it's the same but using `.t(` instead of `$t(`

website/common/locales/en/tasks.json
```diff
@@ -27,15 +20,11 @@
   "save": "Save",
   "addChecklist": "Add Checklist",
   "checklist": "Checklist",
-  "checklistText": "Break a task into smaller pieces! Checklists increase the Experience and Gold gained from a To Do, and reduce the damage caused by a Daily.",
   "newChecklistItem": "New checklist item",
-  "expandChecklist": "Expand Checklist",
```
paglias, line 32: this one is used here https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/tasks/task.vue#L184

website/common/locales/en/tasks.json
paglias: Let me know if you think this can be a viable option to solve the issue, also just noticed there are a bunch of cases (the only ones in the entire codebase) of `+` being used to construct i18n keys in this file https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/inventory/items/cards-modal.vue

paglias (commented):
Left some comments, regarding the alias-related keys you can remove them

website/common/locales/en/tasks.json
RaitheOfDureya: This is a good example, thanks. I believe I have understood the issue. I will correct the strings you mentioned and watch out for such cases in the next searches.

paglias:
@RaitheOfDureya this regex expression should be enough to check for keys passed dynamically:

- server and common code `\.t\((?!')` (searches for `.t(` not followed by `'`)
- client `\$t\((?!')`

My suggestion would be to first make a list of all keys that fall into these cases and then go ahead with the method you described in the issue making sure not to remove any key from the list made

RaitheOfDureya:
Hi @paglias,

I repeated the searches and was able to find all the cases mentioned before. 

- `\.t\((?!')` - returned 43 results in 16 files

- `\$t\((?!')` - returned 122 results in 59 files

I reviewed each one manually and made notes for the future as you suggested. 

I restored the `sureDeleteCompletedTodos` string, because I believe it is being used in the [taskModal.vue](https://github.com/HabitRPG/habitica/blob/b2833ac4a209c3ecd960170d8295b2b3a694d3e2/website/client/src/components/tasks/taskModal.vue).

To the best of my knowledge the other removed strings are not being used anywhere.

*P.S.: Since I spent more than 2 hours reviewing the deleted strings, I may be overlooking something. Am I?*

paglias:
@RaitheOfDureya thanks for all your work here, just a couple of things:

- `sureDeleteCompletedTodos ` -> I'm not sure it's being used, I found `clearCompletedConfirm` which has a very similar wording (or even the same) but I don't see `sureDeleteCompletedTodos` used anywhere
- Due to a bug in our translations services if we remove a string from the English locales it doesn't get removed from the other languages, so we should do that manually. If it's too much work or annoying that's totally fine, we can do it for english only and then I'll write a script to remove the strings from the other locales later

After these two points are addressed I'm going to merge the PR :)

RaitheOfDureya:
The following unused I18N strings have been removed from the `tasks.json` file in all languages:

- sureDeleteCompletedTodos
- lotOfToDos
- deleteToDosExplanation
- addsingle
- newHabit
- newHabitBulk
- checklistText
- extraNotes
- direction/Actions
- taskAlias
- taskAliasPopover
- taskAliasPlaceholder
- taskAliasPopoverWarning
- attributeAllocation
- attributeAllocationHelp
- newDaily
- newDailyBulk
- repeatHelpTitle
- dailyRepeatHelpContent
- weeklyRepeatHelpContent
- repeatDays
- repeatWeek
- newTodo
- newTodoBulk
- dated
- ingamerewards
- newReward
- newRewardBulk
- clearTags
- hideTags
- showTags
- startDateHelpTitle
- startDateHelp
- streakerAchievement
- fortifyText
- confirmFortify
- emptyTask
- dailiesRestingInInn
- habitHelp1
- habitHelp2
- habitHelp3
- newbieGuild
- dailyHelp1
- dailyHelp2
- dailyHelp3
- dailyHelp4
- dailyHelp5
- toDoHelp1
- toDoHelp2
- toDoHelp3
- toDoHelp4
- rewardHelp1
- rewardHelp2
- rewardHelp3
- rewardHelp4
- clickForHelp
- strengthExample
- intelligenceExample
- perceptionExample
- constitutionExample
- counterPeriod
- counterPeriodDay
- counterPeriodWeek
- counterPeriodMonth
- habitCounter
- habitCounterUp
- habitCounterDown
- repeatZero
- repeatType
- repeatTypeHelpTitle
- repeatTypeHelp
- onDays
- groupTasksByChallenge
- taskNotes
- monthlyRepeatHelpContent
- yearlyRepeatHelpContent
- summaryStart
- yesterDailiesTitle
- yesterDailiesOptionTitle
- yesterDailiesDescription
- repeatDayError
- searchTasks

paglias:
Thanks @RaitheOfDureya ! We're going to test this for a few days to be 100% sure we didn't miss anything but I already tested it myself and it's working well so feel free to go ahead with the other files if you want

RaitheOfDureya:
That's good to hear, @paglias :smile: 
I will start with the next file. 
Have a nice Sunday!

RaitheOfDureya:
@paglias, I have identified the following unused strings in the `subscriber.json` file. I only removed them from the English folder to make it easier to revert if necessary.

- subDescription
- buyGemsGoldText
- retainHistory
- retainHistoryText
- doubleDrops
- doubleDropsText
- supportDevs
- supportDevsText
- exclusiveJackalopePetText
- giftSubscriptionText1
- giftSubscriptionText2
- giftSubscriptionText3
- monthUSD
- organization
- indivPlan1
- indivPlan2
- groupText1
- groupText2
- planFamily
- planGroup
- dedicatedHost
- dedicatedHostText
- individualSub
- subscribed
- manageSub
- cancelSubInfoGoogle
- cancelSubInfoApple
- canceledSubscription
- adminSub
- morePlans
- organizationSub
- organizationSubText
- hostingType
- hostingTypeText
- dedicated
- customDomain
- customDomainText
- maxPlayers
- maxPlayersText
- unlimited
- priSupport
- priSupportText
- timeSupport
- timeSupportText
- gameFeatures
- gold2Gem
- gold2GemText
- infiniteGem
- infiniteGemText
- notYetPlan
- subCanceled
- buyGemsGoldTitle
- becomeSubscriber
- freeGemsTitle
- buyGemsAllow1
- buyGemsAllow2
- purchaseGemsSeparately
- subFreeGemsHow
- seeSubscriptionDetails
- timeTravelersTitleNoSub
- timeTravelersTitle
- timeTravelersPopoverNoSub
- mysticHourglassPopover
- hourglassBuyEquipSetConfirm
- hourglassBuyItemConfirm
- earnGemsMonthly
- receiveMysticHourglass
- receiveMysticHourglasses
- everyMonth
- everyXMonths
- everyYear
- subscribeSupportsDevs
- subscriptionBenefitLeadin
- subscriptionBenefit2
- haveCouponCode
- subscriptionAlreadySubscribed1
- petsNotAllowedHourglass
- mountsNotAllowedHourglass

I also checked if the strings are being generated dynamically and only found one occurrence that I was not sure about:
- `timeTravelersAlreadyOwned` 

in website/common/script/ops/buy/hourglassPurchase.js:
```javascript
.t(`${type}AlreadyOwned` ...
// type === 'pets, type === 'mounts', type === 'backgrounds', type === 'quests', ...
```

Should I keep or remove this string?

Once you review and approve the strings to be removed I can exclude them from other languages.

Cheers!

paglias:
> .t(`${type}AlreadyOwned` ...

I think it can be removed as the check is one `user.items[type]` and `user.items.hourglass` doesn't exist in the user schema

Another thing, could you make a separate pr for each file? You'll have to create a new branch (from develop do `git checkout -b remove-unused-fileName`)

paglias:
Thanks @RaitheOfDureya ! I've noted this towards your 5th tier and awarded the Blacksmith title!

RaitheOfDureya:
Thank you, @paglias 
:hammer:

PR #12592, opened 2020-09-20
fix(export) - Issue 12482 - Fix messages in xml export.

### LINKED ISSUE #12482: Data Export Erroring Out
Users report being unable to export data via XML or JSON in #10100. 

Per suggestion from @paglias, the JSON export can be accomplished via this link: https://habitica.com/user/settings/data-export instead of accessing it directly

Previous discussion located at #10100 but as the cause is likely to be different, we are starting with a new issue ticket.

benkelaar:
Fixes #12482

### Changes
- For XML export, convert ID-keyed newMessages object to list of objects with ID as field.

----


test/api/unit/top-level/dataexport.test.js
```diff
@@ -0,0 +1,79 @@
+import dataexport from '../../../../website/server/controllers/top-level/dataexport';
```
paglias: thanks for the fix @benkelaar ! The only thing is, this test is about a controller so we should either:

1. Transform it into an integration test
2. Move the logic for the XML export outside of the controller into a library file (website/server/libs) and unit test that

benkelaar: I think 2 is a good idea that I might make some time for.

What's the reason behind 1?

paglias: The idea is that for the results of controllers, which is the result of making an HTTP call to them we use integration tests that actually do an HTTP call behind the scenes and check its results. Unit tests instead are used to test the behavior of functions calls. Does it make sense?

So in general:

- controllers -> get tested with integration tests
- anything else -> unit tests

So if in this case the behavior is difficult to test with an integration test we should move the code to the `libs` folder and make an unit test against a function that simply generates the XML, not that calls the XML generator and then checks that the response sent to the HTTP request is correct

benkelaar: Cool, I understand.  Is this strategy documented somewhere?  Might be a good idea to add it to the CONTRIBUTING.md or the wiki.

gulp/gulp-tests.js
```diff
@@ -41,6 +40,15 @@ const testBin = (string, additionalEnvVariables = '') => {
   return `NODE_ENV=test ${additionalEnvVariables} ${string}`;
 };
 
+function runInChildProcess (command, options = {}, envVariables = '') {
+  return done => pipe(exec(testBin(command, envVariables), options, err => done(err)));
```
benkelaar, line 44: Doing `err=> done(err)` instead of the `System.exit()` option gets rid of the "did you forget to call done somewhere" notification when you have failing tests.

gulp/gulp-tests.js
```diff
@@ -41,6 +40,15 @@ const testBin = (string, additionalEnvVariables = '') => {
   return `NODE_ENV=test ${additionalEnvVariables} ${string}`;
 };
 
+function runInChildProcess (command, options = {}, envVariables = '') {
+  return done => pipe(exec(testBin(command, envVariables), options, done));
```
benkelaar, line 44: Passing `done` directly as callback instead of the `System.exit(1)` option gets rid of the `did you forget to call done somewhere` notification when you have failing tests.

test/api/unit/top-level/dataexport.test.js
benkelaar: @paglias So I've been looking at where to put the integration test.  At first I wanted to add a new test run target `test:api-top-level:integration` (That's why I got to slightly refactoring `gulp-tests.js`), but now I notice that the payment endpoints are simply tested under `api-v3`.

Given the division of run targets I think it would be cleaner to add a `test:api-top-level:integration` target and have that run also the payment tests, but that would also mean setting up an extra github action to run those, and I don't think I have the rights to do that.

So shall I just add the integration test to `api-v3` or shall I create a new `api-top-level` test run target?

paglias: Good point, let's add it to the api-v3 for now, apart from the naming which is indeed confusing it's not harmful

benkelaar: Added an integration test and extracted the functionality in a library.

benkelaar:
@paglias Ready for review again.  Ideally I'd add test-cases for the other corner cases in the library, but I'm late for work 😁 .

test/api/v3/integration/GET-dataexport.test.js
```diff
@@ -0,0 +1,424 @@
+import moment from 'moment';
+
+import { generateUser, requester } from '../../../helpers/api-integration/v3';
+import { Task } from '../../../../website/server/models/task';
+
+describe('GET /export/userdata.xml', () => {
```
paglias, line 6: The PR is looking good @benkelaar , I'm a bit concerned about this test because it'll require updates every time something changes in the default user schema (a new gear, a new audio theme available, a new preference...)

paglias: ah, also we have an existing integration test for this route, https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/dataexport/GET-export_userdata.xml.test.js, I'm going to merge the PR removing this test

paglias:
thanks @benkelaar ! Noted in your profile

test/api/v3/integration/GET-dataexport.test.js
benkelaar: Yeah, I was going back and forth about it as well.  Apologies that I missed that test.  Thanks for removing it.  Might take some time next week to document the test conventions somewhere.

Alys: > Might take some time next week to document the test conventions somewhere.

That would be awesome if you want to do that! 

It could go in the tests section of [Using Your Local Install to Modify Habitica's Website and API](https://habitica.fandom.com/wiki/Using_Your_Local_Install_to_Modify_Habitica's_Website_and_API) although we may end up finding that a separate wiki page just for running and writing tests would be helpful.

benkelaar: Yeah, was thinking it might be worthwhile to have a separate dedicated page for that as well.  However, our baby thought my laptop might make a good water receptacle yesterday, so had a bit of a setback work-wise 😆 .   I have a (temporary) replacement laptop now, but will have to re-do some of the things, so might take me a while.

PR #12605, opened 2020-09-22
Less mess for lingering pinned Seasonal Shop gear

### LINKED ISSUE #11151: locked pinned Grand Gala seasonal equipment for your class has wrong date and wrongly mentions class
EDIT: When this becomes help wanted again, note https://github.com/HabitRPG/habitica/issues/11151#issuecomment-626287070 and any comments below it. They are likely to be relevant to the solution.

When seasonal equipment for the class you are currently playing from the Spring 2019 Grand Gala is pinned to your Rewards column and you have never owned it, the on-hover text says **"Available until 31 October 2018"** (at least for the example item shown below; it's possible that other items show different text).

I have no idea where it's getting that date from; I can't find a reference to that date in the code in any context that should apply to those gear pieces.

Regardless, when the gear is no longer available, the message should say that instead of having any date. For example **"This item is no longer available."**

The on-hover text also says that it's available to a specific class, which is irrelevant when you are the right class - that message shouldn't appear.

Here's an example of the problem:
```
{
    "_id": "d904bd62-da08-416b-a816-ba797c9ee265",
    "profile": {
        "name": "Alys"
    },
    "flags": {
        "classSelected": true
    },
    "stats": {
        "class": "wizard"
    }
    "pinnedItems": [
        {
            "path": "gear.flat.weapon_special_spring2019Mage",
            "type": "marketGear"
        }
    ]
}
``` 
image

_Note that the desired behaviour described above applies to only gear that you have never owned._
_If you have previously bought the item and lost it due to death, the equipment will appear in your `items.gear.owned` object with a `false` value, and then there should be no message indicating lack of availability due to the date or class. That situation is currently working and should not be changed._

SabreCat:
Fixes #11151

### Changes

**Previously**, in the event that a user manually pinned a Seasonal Shop item ~~buyable with Gems~~, and that item remained pinned after the end of the Gala, misleading text would appear when hovering over that item. It referred erroneously to class restrictions and provided a useless end date.

**Now**, instead, the hover text will simply say that the item is not available for purchase, and will not display an end date.

To test this, you'll need to change the Gala end date in `website/common/script/content/constants/index.js` and `website/common/script/libs/shops-seasonal.config.js` to something earlier than the current date.

### DIFF
--- a/website/client/src/components/inventory/equipment/attributesPopover.vue
+++ b/website/client/src/components/inventory/equipment/attributesPopover.vue
@@ -5,29 +5,30 @@
         {{ `${$t('lockedItem')}` }}
       </h4>
       <div
-        v-if="isWrongClass"
-        class="popover-content-text"
+        v-if="item.currency === 'gems'"
+      >
+        {{ `${$t('messageNotAvailable')}` }}
+      </div>
+      <div
+        v-else-if="isWrongClass"
       >
         {{ `${$t('classLockedItem')}` }}
       </div>
-      <div
-        v-else
-        class="popover-content-text"
-      >
+      <div v-else>
         {{ `${$t('tierLockedItem')}` }}
       </div>
-      <p></p>
     </div>
     <div v-else>
       <h4 class="popover-content-title">
         {{ itemText }}
       </h4>
-      <div class="popover-content-text">
+      <div>
         {{ itemNotes }}
       </div>
       <attributesGrid
         :user="user"
         :item="item"
+        class="mt-3 mb-2"
       />
     </div>
   </div>
--- a/website/client/src/components/shops/shopItem.vue
+++ b/website/client/src/components/shops/shopItem.vue
@@ -106,7 +106,6 @@
         </div>
         <div
           v-if="item.event"
-          :class="item.purchaseType === 'gear' ? 'mt-4' : 'mt-2'"
         >
           {{ limitedString }}
         </div>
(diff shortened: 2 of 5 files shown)


paglias:
Schermata 2020-09-23 alle 17 58 32

I changed the config to have the event expire early but the hover text doesn't say that it's not available for purchase

Schermata 2020-09-23 alle 17 57 29

And I'm not sure if it's wanted but the buy modal keeps saying that it's available until August 31st, not sure if we want to change that as well

Schermata 2020-09-23 alle 17 57 33

website/client/src/components/inventory/equipment/attributesPopover.vue
```diff
@@ -5,29 +5,30 @@
         {{ `${$t('lockedItem')}` }}
       </h4>
       <div
-        v-if="isWrongClass"
-        class="popover-content-text"
+        v-if="item.currency === 'gems'"
```
paglias, line 8: so I think the comment I left is happening because I tested with a gold purchasable item while here we're checking only for gems, if it's expected just ignore my comment - I noticed that in the linked issue the example is made with a gold purchasable item

SabreCat: ahh, interesting... ok, I missed that. Sounds like I have some more work to do, thanks!

SabreCat:
We've done a number of updates to this for recent events. If this specific issue needs a PR still, we'll likely fix from scratch. Closing!

PR #12612, opened 2020-09-26
Fix vue test stacktraces

benkelaar:
I started this a while back (Which also lead to resolving the zone() deprecation issue) but now that I wanted to look at a unit test for task.vue I decided to wrap this up.  This merge request resolves all the warnings and stack traces that are being thrown from the client tests, so that debugging those tests will be a bit easier.

### Changes
- Introduces extra test setup in client-side unit tests that resolves warnings and stack traces thrown while running them
- Introduces a unit test for `home.vue`
- Replaces a `indexOf === -1` call with an `includes` check in `avatar.vue`

----

### DIFF
--- a/website/client/src/components/avatar.vue
+++ b/website/client/src/components/avatar.vue
@@ -227,7 +227,7 @@ export default {
       return this.member.preferences.costume ? 'costume' : 'equipped';
     },
     specialMountClass () {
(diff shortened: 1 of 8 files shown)


benkelaar:
@paglias Do you have time to take a look at this one?  It's basically only test setup that's slightly expanded.

If it would be better to merge this without the added unit test for home I can take that one out.

paglias:
It's looking good but I can't merge this today

paglias:
Thanks! we were having some issues with the develop branch but it's fixed now so I merged it

PR #12627, opened 2020-09-28
Fixes Issue #12421 - Change page title based on the site section being viewed

### LINKED ISSUE #12421: Change page title based on the site section being viewed
Currently the page title is `Habitica - Gamify You Life` no matter the section of the site you're visiting.

We already support dynamically changing the page title using the `title` property in the app store. See
- https://github.com/HabitRPG/habitica/blob/8003632041f0f62634be1f66b34bbd75ac3add1b/website/client/src/app.vue#L470-L472
- https://github.com/HabitRPG/habitica/blob/8003632041f0f62634be1f66b34bbd75ac3add1b/website/client/src/store/index.js#L54

For example,

- we could show the guild name when viewing a guild, same for challenges parties, ...
- we could show the section, for example Inventory, ...

It would help users using multiple Habitica tabs at the same time to distinguish between them and also to remember the Habitica page currently open while viewing a different site.

I'm labelling this as suggestion-discussion until the exact titles have been decided

jbusa22:

Fixes #12421 

### Changes
Added dynamic title updating by modifying the title attribute in the vue store on component mount.
Current dynamic title sections are:
Tasks
Inventory
Shops
Party
Guilds
Group
Challenges

----


website/client/src/components/challenges/challengeDetail.vue
```diff
@@ -433,6 +433,7 @@ export default {
         this.$router.push('/challenges/findChallenges');
         return;
       }
+      this.$store.state.title = `${this.challenge.name} | Challenges | Habitica`;
```
benkelaar: Looking good @jbusa22 !  However in order to support the right titles in all languages, these titles should probably go through the `$t` translation function.

If you want to find out which keys to use for the menu text, you can find them all in [the menu source file](https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/header/menu.vue), or you can search for the exact match in the [english translation files](https://github.com/HabitRPG/habitica/tree/develop/website/common/locales/en).

As an example, this line would become: 
```suggestion
      this.$store.state.title = `${this.challenge.name} | ${$t('challenges')} | ${$('habitica')}`;
```

Note that here the keys are basically the same term as the english word, and most seem to follow this pattern, but it's best to check either in the menu or in the files.

Edit: You could also consider extracting this format into a helper function, along the line of

```javascript
function createTitle (pageTitle, groupTitleKey) {
  return `${pageTitle} | ${$t(groupTitleKey)} | ${$t('habitica')}`;
} 
```

You'd have to make sure the translation function is also available there though, @paglias , perhaps you have a suggestion how to do that?  Alternatively I guess it's fine to just use the translation function in all the different usages.

jbusa22: Yeah, I'm was thinking of implementing a helper function but it seemed like unnecessary abstraction given the string is only on 1 line per component. In order to separate the translation out, we would need to import the i18n file from common/script/i18n into the util file where the createTitle would be. I think it might just be cleaner to keep it in the string but not sure. Let me know what you think.

benkelaar: Ah yeah, you're right, the injection of the function is trivial, wasn't really thinking straight (Was confused by the fact that the function lives on `this` in the vue components).

The introduction of the abstraction is not about reducing lines, it's much more about codifying the fact that all these titles follow the same logic.  That way you decouple the structure of the title from the content of the title.  As such the implicit knowledge is made explicit and changing the structure of the title in the future becomes easier. 

Technically I think the fact that this solution has duplication (especially in the grouping level) with the menu definition means that this is not really the right place to define it.  It suggests to me that there should be a data structure that holds the grouping information of the links together with the components they link to and the knowledge of how to construct titles based on that grouping and page's content.  The menu can then be rewritten to just generate all the menu items based off of this data structure and inject the titles into the page components also from that data structure (Or from the menu, kinda depends on how you structure it).  That would be a fairly hefty refactoring though, so perhaps as a second step (And definitely only if someone from the core team thinks it's a good idea).

paglias: I think the option of creating a store action that takes parameters similar to those suggested by @benkelaar and changes the state is the best way to go, as I said previously in the issue. I completely agree with the reasoning outline here in the last comment by @benkelaar

paglias: Regarding the need to inject the translation function you can use `this.$t` anywhere in components so nothing needs to be injected and outside components (for example in a store action) you can just `import i18n from '@/../../common/script/i18n';` and do `i18n.t()`

jbusa22: Ok here is what I have so far for the store action. 
export function setTitle (store, params) {
  if (params.second) {
    store.state.title = `${i18n.t(params.first)} | ${i18n.t(params.second)} | ${i18n.t('habitica')}`;
  } else {
    store.state.title = `${i18n.t(params.first)} | ${i18n.t('habitica')}`;
  }
}
Questions I have are:
1. What should the params be named?
2. When i18n doesn't find the string it returns 'String X Not Found'. Obviously, we don't want this showing up in the title so I'm not sure how to get around this. One option is to check all the params first and match the return string for something like 'not found'.  Another option is to put this on the caller of the action. ie setTitle({first: {name: 'Helllo's party', translate: false}})

jbusa22: One more question is what file should we put this function in? Right now it lives in the common.js in the store folder.

paglias: The file should be under store/actions/common.js, I think to avoid issues with the translations the action should be directly passed the translated version (and I don't think Habitica needs to be translated with `i18n.t('habitica')`, let's just use Habitica in english). If a string is missing (String X Not Found) it should be added to `common/locales/en`

benkelaar: The reason I thought you might want to translate Habitica I thought might be for languages that use different alphabets, or do you want to stylize the brandname also in roman characters in those languages?

Come to think about it, what about right-to-left languages? Shouldn't those display the title right to left as well? (Perhaps this might be scope creep, probably better to leave to a follow-up story (if at all necessary))

paglias: I think we want to keep Habitica in latin characters regardless of the user language and regarding RTL languages there's definitely work to be done but we want to tackle that all together and right now it's not in the plans since it would be a big project

paglias:
@jbusa22 I tested the PR and overall it's working well, there are a few issues though:

1. If you edit the name of a guild / party / challenge the name in the title doesn't change until the page is refreshed. A solution might be do add a watcher (https://vuejs.org/v2/guide/computed.html#Watchers) on the group / challenge name and update the title when that changes
2. Looks like the `groupPlans` string already exists

>      Error: groupPlans in ./website/common/locales/en/subscriber.json already exists in ./website/common/locales/en/groups.json.

3. When switching to a page that doesn't have a title like the Settings the name is not reset to Habitica
4. For a group plan page we should use `NAME | Group Plans | Habitica`

And a couple of things for @Tressley 

1. In the issue I noticed you used `Teams` in place of Groups Plans but the wording across the site has not yet been changed to Teams
2. Do we want to add a few more titles here? I noticed that we're not covering: PMs, User's profile / stats / achievement pages in the user modal, settings

Tressley:
@paglias

> In the issue I noticed you used Teams in place of Groups Plans but the wording across the site has not yet been changed to Teams

Yeah, we're not settled on a date for that change, so let's move forward with "Group" until we're ready to move forward on the potential name change.

> Do we want to add a few more titles here? I noticed that we're not covering: PMs, User's profile / stats / achievement pages in the user modal, settings

Yeah, I think this should try to hit all sections on the site. I outlined most of them, but was sure I was missing others. For the sections I missed, they should follow the same structure: `Sub-section | Section | Habitica`

jbusa22:
In response to @paglias,
1. Sounds good I will add watchers and test name changing to the group, guild / party, and challenge sections
2. Easy fix
3. There doesn't seem to be a nice way to check this. If we have all pages outlined here then this won't be a problem so maybe this is okay?
4. I'm a little confused about this. Do you mean on your group's page it should be `NAME | Group Plans | Habitica` and on the actual group plans page just `Group Plans | Habitica`?

paglias:
@jbusa22 

2. I think the merge commit reverted the previous one where the string was removed
3. yes once we have all pages outline it won't be a problem, I'd also add a comment in the router file to explain that each page will have to set its title
4. I mean in `website/client/src/components/groups/group.vue` it's just setting the section but not subSection, I think it should be `group.name | Group Plans | Habitica`. Does it make sense?

jbusa22:
@paglias Yes, that makes sense.

website/client/src/components/groups/group.vue
```diff
@@ -498,7 +498,8 @@ export default {
     if (!this.searchId) this.searchId = this.groupId;
     await this.fetchGuild();
     this.$store.dispatch('common:setTitle', {
-      section: this.group.name,
+      section: this.t$('groupPlans'),
```
paglias: this should be $t not t$

website/client/src/components/groups/group.vue
```diff
@@ -498,7 +498,8 @@ export default {
...
+      subSection:this.group.name,
```
paglias: I think lint is going to complain about the lack of space before this.group.name

jbusa22:
I think I hit all of the sections / subsections. I also added a watcher like you said to the challenges, but just used an already defined function for groups. Also features isn't used in the json, so I added it, not sure if I need to do anything else for translations to work.

website/client/src/components/groups/group.vue
```diff
@@ -485,6 +485,7 @@ export default {
     // call again the method if the route changes (when this route is already active)
     $route: 'fetchGuild',
     partyStore () {
+      debugger;
```
paglias: this should be removed :)

website/client/src/components/challenges/challengeDetail.vue
```diff
@@ -384,6 +384,17 @@ export default {
...
   },
+  watch: {
+    challenge: {
+      handler (newVal) {
+        this.$store.dispatch('common:setTitle', {
+          section: this.$t('challenge'),
+          subSection: newVal.name,
+        });
+      },
+      deep: true,
```
paglias: instead of using `deep` you can limit the watcher to `challenge.name`:

```
watch: {
    'challenge.name': {
      handler (newVal) {
        this.$store.dispatch('common:setTitle', {
          section: this.$t('challenge'),
          subSection: newVal.name,
        });
      },
  }
}
```

website/client/src/router/index.js
```diff
@@ -6,6 +6,8 @@ import handleRedirect from './handleRedirect';
 
 import ParentPage from '@/components/parentPage';
 
+// NOTE: when adding a page make sure to implement setTitle
+
 // Static Pages
 const StaticWrapper = () => import(/* webpackChunkName: "entry" */'@/components/static/staticWrapper');
 const HomePage = () => import(/* webpackChunkName: "entry" */'@/components/static/home');
```
paglias: I think the home is not being handled by the PR, we should still have a call to `setTitle` to make sure if you go there from a different page it'll reset to simply `Habitica`

paglias (changes requested):
thanks @jbusa22 , I left 3 small comments then it should be good to go :)

website/client/src/router/index.js
jbusa22: Do you mean on the /static/home page? For me, it changes correctly. I'm not sure exactly how it is working but when I go to the home page it updates to the title tag in index.html. Should I set it anyways to be more direct?

website/client/public/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <title>Habitica - Gamify Your Life</title>
+    <title>Habitica</title>
```
paglias: sorry, Gamify Your Life can stay in this file

website/client/src/router/index.js
paglias: you're right that right now it's not possible to go to that page without reloading but just in case we add the ability in future it'd be nice to have it set explicitly, in which case `Habitica - Gamify Your Life` is the title to use

website/common/locales/ach/groups.json
```diff
@@ -14,6 +14,7 @@
     "askQuestionGuild": "Ask a Question (Habitica Help guild)",
     "contributing": "Contributing",
     "faq": "FAQ",
+    "features": "Features",
```
paglias: @jbusa22 adding the string is the correct thing to do but it should be added only to the `locales/en/groups.json` file

paglias (changes requested):
Did another round of testing, left some comments, other than that the only two things missing are titles for:

- Login
- Register

website/common/locales/en/groups.json
```diff
@@ -136,7 +137,6 @@
   "gemGiftsAreOptional": "Please note that Habitica will never require you to gift gems to other players. Begging people for gems is a <strong>violation of the Community Guidelines</strong>, and all such instances should be reported to <%= hrefTechAssistanceEmail %>.",
   "battleWithFriends": "Battle Monsters With Friends",
   "startAParty": "Start a Party",
-  "possessiveParty": "<%= name %>'s Party",
```
paglias, line 139: I think this was removed by mistake

website/common/locales/en/front.json
```diff
@@ -52,7 +52,6 @@
   "featurePetHeading": "Pets and Mounts",
   "featureSocialByline": "Join common-interest groups with like-minded people. Create Challenges to compete against other users.",
   "featureSocialHeading": "Social play",
-  "featuredIn": "Featured in",
```
paglias, line 18: I think this was removed by mistake

website/client/src/components/auth/registerLoginReset.vue
```diff
@@ -777,6 +777,7 @@ export default {
       // windows: WINDOWS_CLIENT_ID,
       google: process.env.GOOGLE_CLIENT_ID, // eslint-disable-line
     });
+    this.setTitle();
```
paglias: I think this might not be necessary beacuse the watched on $route has `immediate: true` so it'll fire right away

website/client/src/components/static/features.vue
```diff
@@ -137,6 +137,12 @@ export default {
   directives: {
     markdown: markdownDirective,
   },
+  mounted () {
+    this.$store.dispatch('common:setTitle', {
+      section: this.$t('help'),
```
paglias: sorry, I don't think that for this static pages (except for FAQ). `help` is the right section, I'd just keep the page name, so in this case it would be Features | Habitica

website/client/src/components/userMenu/profile.vue
```diff
@@ -833,6 +833,10 @@ export default {
   mounted () {
     this.loadUser();
     this.selectPage(this.startingPage);
+    this.$store.dispatch('common:setTitle', {
```
paglias: this doesn't seem to update when the page is changed, I think we can simply set it into `selectPage` and it'll work if you change tab

website/client/src/components/static/home.vue
```diff
@@ -909,6 +909,9 @@ export default {
       // windows: WINDOWS_CLIENT_ID,
       google: process.env.GOOGLE_CLIENT_ID, // eslint-disable-line
     });
+    this.$store.dispatch('common:setTitle', {
+      fullTitle: this.$t('socialTitle'),
```
paglias: I can't find this string, also because Habitica - Gamify Your Life should't be changed I think you can set it directly in english so `fullTitle: 'Habitica - Gamify Your Life', `

paglias (changes requested):

website/common/locales/en/groups.json
jbusa22: I just saw those in my editor, it shows duplicate keys in the files, there still should be one there

website/common/locales/en/front.json
jbusa22: Same with this

website/client/src/components/auth/registerLoginReset.vue
jbusa22: Yes I agree

website/common/locales/en/front.json
```diff
@@ -15,7 +15,17 @@
   "emailNewPass": "Email a Password Reset Link",
   "forgotPasswordSteps": "Enter the email address you used to register your Habitica account.",
   "sendLink": "Send Link",
-  "featuredIn": "Featured in",
+  "evagantzQuote": "My very first dentist appointment where the hygienist was actually excited about my flossing habits. Thanks [Habitica]!",
```
paglias: @jbusa22 it's all looking good, but I don't understand why these strings were added here, they're not used anywhere

paglias: I've fixed this

paglias:
@jbusa22  thanks for this PR! I'm merging the PR and awarding your contributor tier.

PR #12637, opened 2020-10-02
Fix overflow in guild member text and keyboard inaccessible member/bank lists fixes #11067

### LINKED ISSUE #11067: Text displaced off Member List button in large guilds
> raventarte   @ravenlune (9272ab2c-4b66-4b74-aebd-799725885d1c) [web] 
> It would appear that there was a CSS change to the website that now causes the text "Member List" to appear off the surrounding badge if a guild has a very large member count, and the number count itself to stop aligning with the shield icon. An example is the Habitica Help guild, versus this guild, which appears properly. Here is a screenshot comparison. I believe it is the fixed height/width values of the badge that causes this issue, and it seems to only occur if the member count is 10.1k or higher (as this is too large for the badge to handle).

Confirmed in Windows / Google Chrome.

J-Sanderson:

Fixes #11067

### Changes
Fixes overflowing text in guild member lists with large member counts.
Fix the guild member list and gem bank divs being inaccessible to keyboard navigation.

Screenshot from 2020-10-01 19-29-41

cbeeb477-d5db-4efd-8f0b-68201870482b

----
UUID:

### DIFF
--- a/website/client/src/components/groups/group.vue
+++ b/website/client/src/components/groups/group.vue
@@ -10,7 +10,7 @@
     <group-gems-modal />
     <div class="col-12 col-sm-8 standard-page">
       <div class="row">
-        <div class="col-12 col-md-6 title-details">
+        <div class="col-12 col-md-4 title-details">
           <h1>{{ group.name }}</h1>
           <div>
             <span class="mr-1 ml-0">
@@ -22,40 +22,36 @@
             </span>
           </div>
         </div>
-        <div class="col-12 col-md-6">
+        <div class="col-12 col-md-8">
           <div class="row icon-row">
             <div
-              :class="{ 'offset-8': isParty }"
+              class="item-with-icon"
+              tabindex="0"
+              role="button"
+              @keyup.enter="showMemberModal()"
+              @click="showMemberModal()"
             >
               <div
-                class="item-with-icon"
-                tabindex="0"
-                role="button"
-                @keyup.enter="showMemberModal()"
-                @click="showMemberModal()"
+                v-if="group.memberCount > 1000"
+                class="svg-icon shield"
+                v-html="icons.goldGuildBadgeIcon"
+              ></div>
+              <div
+                v-if="group.memberCount > 100 && group.memberCount < 999"
+                class="svg-icon shield"
+                v-html="icons.silverGuildBadgeIcon"
+              ></div>
+              <div
+                v-if="group.memberCount < 100"
+                class="svg-icon shield"
+                v-html="icons.bronzeGuildBadgeIcon"
+              ></div>
+              <span class="number">{{ group.memberCount | abbrNum }}</span>
+              <div
+                v-once
+                class="member-list label"
               >
-                <div
-                  v-if="group.memberCount > 1000"
-                  class="svg-icon shield"
-                  v-html="icons.goldGuildBadgeIcon"
-                ></div>
-                <div
-                  v-if="group.memberCount > 100 && group.memberCount < 999"
-                  class="svg-icon shield"
-                  v-html="icons.silverGuildBadgeIcon"
-                ></div>
-                <div
-                  v-if="group.memberCount < 100"
-                  class="svg-icon shield"
-                  v-html="icons.bronzeGuildBadgeIcon"
-                ></div>
-                <span class="number">{{ group.memberCount | abbrNum }}</span>
-                <div
-                  v-once
-                  class="member-list label"
-                >
-                  {{ $t('memberList') }}
-                </div>
+                {{ $t('memberList') }}
               </div>
             </div>
             <div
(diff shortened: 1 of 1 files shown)


website/client/src/components/groups/group.vue
```diff
@@ -58,10 +60,13 @@
...
             <div
               v-if="!isParty"
-              class="col-4"
+              class="offset-1"
             >
               <div
                 class="item-with-icon"
+                tabindex="0"
+                role="button"
+                @keyup.enter="showMemberModal()"
```
paglias: I think here it should call `showGroupGems `

paglias (changes requested):
left one comment :)

website/client/src/components/groups/group.vue
J-Sanderson: copy paste fail, thanks for catching!

website/client/src/components/groups/group.vue
```diff
@@ -25,11 +25,13 @@
         <div class="col-12 col-md-6">
           <div class="row icon-row">
             <div
-              class="col-4 offset-4"
               :class="{ 'offset-8': isParty }"
```
paglias: I think the offset here might not be necessary anymore

website/client/src/components/groups/group.vue
```diff
@@ -225,12 +230,15 @@
     box-shadow: 0 2px 2px 0 rgba(26, 24, 29, 0.16), 0 1px 4px 0 rgba(26, 24, 29, 0.12);
     padding: 1em;
     text-align: center;
-    min-width: 80px;
-    max-width: 120px;
+    min-width: 100px;
```
paglias: I'm not sure exactly why give the min-width was set to 80px but it looks like the item here is currently 120px while with the PR it shrank a bit. Could we change it so that by defaults it's 120px wide?

website/client/src/components/groups/group.vue
J-Sanderson: Yes, looks like that entire container div can go without any problems.

website/client/src/components/groups/group.vue
J-Sanderson: Fixed!

paglias:
Looking good @J-Sanderson !

@Tressley I don't think we had a zeplin for this part but just in case this is something we want to change, the space between the Members Count and Gem Bank is slightly larger than before (it wasn't a specific distance before, just what happened with the previous code) 

Schermata 2020-10-06 alle 23 09 59

Do we want to reduce it?

This is with a margin between the two of 1.5rem

Schermata 2020-10-06 alle 23 13 22

website/client/src/components/groups/group.vue
```diff
@@ -10,7 +10,7 @@
     <group-gems-modal />
     <div class="col-12 col-sm-8 standard-page">
       <div class="row">
-        <div class="col-12 col-md-6 title-details">
+        <div class="col-12 col-md-4 title-details">
```
paglias: I think changing this to md-4 is causing some issues

Schermata 2020-10-07 alle 15 16 34

The title should have more space to expand

paglias: It should be the other way around md-8 for the title and md-4 for the badges

website/client/src/components/groups/group.vue
```diff
@@ -225,12 +226,15 @@
...
+    min-width: 120px;
     height: 76px;
 
+    &:last-of-type {
+      margin-right: 1rem;
+    }
+
     .svg-icon.shield, .svg-icon.gem {
-      width: 28px;
+      min-width: 28px;
```
paglias: I think this should be set back to `width` not `min-width`, it's causing some issues with firefox
image

paglias (changes requested):
@J-Sanderson left two comments about some issues

website/client/src/components/groups/group.vue
J-Sanderson: Ok, want to make sure while I'm here that the spacing looks ok?

Screenshot from 2020-10-07 18-44-34

They also stack on certain smaller sized screens since they can't shrink below 120 pixels, is this acceptable?

Screenshot from 2020-10-07 18-48-50

For reference the same size screen looked like this before:

Screenshot from 2020-10-07 18-47-24

website/client/src/components/groups/group.vue
J-Sanderson: Yeah, that shouldn't have been there, thanks for the catch. Must have done that by accident.

website/client/src/components/groups/group.vue
paglias: I haven't checked but maybe having both col-md-6 instead of 8 and 4 could result in a better support for smaller screens while still allowing longer titles

J-Sanderson: Looks about the same with that. For reference the stacking occurs from widths of 768px to 892px if both are at col-md-6.

paglias: Let's keep the two col-md-6 then

J-Sanderson: Ok - pushing fixes.

paglias:
@J-Sanderson  thanks for this! I'm merging the PR and awarding your contributor tier.

PR #12609, opened 2020-09-24
fix(task) - 11139 use start-from hours in due date calculation

benkelaar:
Fixes #11139

Based on @sayansur25 's work in #11163 , credit should go to them.

Sinned against my own standards and neglected creating a `task.vue` unit test.  Apologies.

### Changes
- Use dayStart information to calculate due date.

----
UUID: d71d2c57-a73d-4591-b64d-e688584a9092

PR #12650, opened 2020-10-05
fix(task) - Add case for when due date is before dayStart

benkelaar:
Continued work on #12609

Changes
- Add still missing test for anti-case
- Make that work as well

### DIFF
--- a/website/client/src/components/tasks/task.vue
+++ b/website/client/src/components/tasks/task.vue
@@ -1071,8 +1071,10 @@ export default {
       this.scoreChecklistItem({ taskId: this.task._id, itemId: item.id });
     },
     calculateTimeTillDue () {
-      const endOfToday = moment().subtract(this.user.preferences.dayStart, 'hours').endOf('day');
-      const endOfDueDate = moment(this.task.date).endOf('day');
+      const endOfDay = time => time.subtract(this.user.preferences.dayStart, 'hours').endOf('day');
+
+      const endOfToday = endOfDay(moment());
+      const endOfDueDate = endOfDay(moment(this.task.date));
 
       return moment.duration(endOfDueDate.diff(endOfToday));
     },
--- a/website/client/tests/unit/components/tasks/task.spec.js
+++ b/website/client/tests/unit/components/tasks/task.spec.js
@@ -79,10 +79,17 @@ describe('Task', () => {
     });
 
     it('formats due date to tomorrow if today but before dayStart', () => {
-      const now = setClockTo('2019-06-12T04:23:37+02:00');
+      const now = setClockTo('2019-06-12T04:23:37+07:00');
       wrapper = makeWrapper({ date: now.add(8, 'hours') }, { preferences: { dayStart: 7 } });
 
       expect(wrapper.vm.formatDueDate()).to.equal('dueIn{"dueIn":"in a day"}');
+    });
+
+    it('formats due date to today if due tomorrow before dayStart', () => {
+      const now = setClockTo('2017-07-02T20:21:27Z');
+      wrapper = makeWrapper({ date: now.add(5, 'hours') }, { preferences: { dayStart: 5 } });
+
+      expect(wrapper.vm.formatDueDate()).to.equal('dueIn{"dueIn":"today"}');
     });
   });
 });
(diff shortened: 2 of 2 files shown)


benkelaar:
@paglias This was the missing test case.

website/client/src/components/tasks/task.vue
```diff
@@ -1071,8 +1071,10 @@ export default {
       this.scoreChecklistItem({ taskId: this.task._id, itemId: item.id });
     },
     calculateTimeTillDue () {
-      const endOfToday = moment().subtract(this.user.preferences.dayStart, 'hours').endOf('day');
-      const endOfDueDate = moment(this.task.date).endOf('day');
+      const endOfDay = time => time.subtract(this.user.preferences.dayStart, 'hours').endOf('day');
+
+      const endOfToday = endOfDay(moment());
+      const endOfDueDate = endOfDay(moment(this.task.date));
```
paglias, line 1077: I think that subtracting the same number of hours from both values and then comparing them results in the same result as not using the dayStart preference at all and might be causing the test to fail

benkelaar: Weird, they succeeded locally -> Perhaps I didn't make it as server-time-independent as I had hoped.

Do note that it's not just "subtracting the same number from both sides of the equation" as what happens is that the dayStart gets subtracted -> And then the time is set to the end of the current day, meaning that the critical issue is whether the subtraction carries that specific time back to before midnight, which should be different for both times.

I'll look at this tonight.

benkelaar: So now they succeeded but I really think I didn't fundamentally change anything.  I only made the test setup more explicit.  I think that's better anyway but I'm worried the test might still be flaky.  I suspect that it might be that `moment().toDate()` perhaps uses server time zone instead of the moment timezone, but that would actually be kinda weird.

Yeah, really don't know why it failed, hoping it's not intermittent.

benkelaar:
@paglias So it seems to be behaving now, but I'm still not sure (Could very well be a time thing) which is why I added the comment.  Beyond that I think this is ready for merging now.

website/client/src/components/tasks/task.vue
paglias: >  I suspect that it might be that moment().toDate() perhaps uses server time zone instead of the moment timezone, but that would actually be kinda weird.

Moment uses the timezone of the server (if it's run on a server - for example in Github actions) or the browser timezone when run in a browser. The only way to force a specific timezone is to use utcOffset https://momentjs.com/docs/#/manipulating/utc-offset/

paglias:
@benkelaar sorry but I don't think this change is working, now it's behaving the same way it did before the previous PR was merged. It might not be reflected on the tests because it uses the timezone of the server but if you try it in the browser you'll see the bug. 

I just replicated by setting by day start at 7PM on the 12th of October (now it's Oct 12th @ 4PM): a todo due on oct 12th (not yet started) is shown as due today with this PR but correctly due tomorrow using `develop`

website/client/src/components/tasks/task.vue
paglias: I don't think that the daystart should be subtracted from both dates, and for the tests we should make sure they're using a reproducible time by correctly setting the timezone

benkelaar: @paglias Ah, I was thinking it would take the timezone from the ISO date, it might be that could be part of the cause.  Can adjust the code to use utcOffset.

There's probably something wrong with my mental model of how this whole thing works.  Apologies for wasting your time like this.  Having a bit of hard time juggling work, family, exercise and this, will make time to look in to this properly somewhere this week.

paglias: No worries @benkelaar , we really appreciate your PRs , whenever you have time is going to be okay :)

SabreCat:
@benkelaar Are you still planning on tackling this? Let us know one way or the other!

benkelaar:
TBH I'm not sure if there's still a problem to fix here.  I think abandoning this PR is acceptable in any case.

SabreCat:
@benkelaar All right!

PR #12652, opened 2020-10-06
Admin Panel to replace admin features in Hall of Heroes - 2

Alys:
Partial fix for HabitRPG/habitica-private#119

Continuation of https://github.com/HabitRPG/habitica/pull/12035

This contains all the commits from the previous PR, plus three new commits, mostly for fixes requested by paglias. 
The comments that this PR fixes start at https://github.com/HabitRPG/habitica/pull/12035#pullrequestreview-500422316

Ready for review (assuming tests pass).


Alys:
The comments that this PR fixes start at
https://github.com/HabitRPG/habitica/pull/12035#pullrequestreview-500422316

paglias:
Pushed to delta, tomorrow I'll make a final review!

website/client/src/components/admin-panel/user-support/index.vue
```diff
@@ -0,0 +1,171 @@
...
+    clearData () {
+      this.hero = {};
+    },
+
+    async loadHero (userIdentifier) {
+      const id = userIdentifier.replace(/@/, ''); // allow "@name" to be entered
+      this.$emit('changeUserIdentifier', id); // change user identifier in Admin Panel's form
+
+      const hero = await this.$store.dispatch('hall:getHero', { uuid: id });
+      this.hero = { ...hero };
```
paglias: ...hero is not necessary when replacing the entire object

website/client/src/router/index.js
```diff
@@ -338,6 +342,22 @@ const router = new VueRouter({
         { name: 'contributors', path: 'contributors', component: HeroesPage },
       ],
     },
+
+    {
+      name: 'adminPanel',
+      path: '/admin-panel',
+      component: AdminPanelPage,
+      meta: { requiresAdmin: true },
+      children: [
```
paglias, line 359: if you Save and Clear data after loading an user in the admin panel the url doesn't change and clicking the Load User button again it refreshes the page

Alys: So sorry, I only just noticed these review comments. I must have missed the email. :(

Yes, that behaviour is intentional.
"Save and Clear" for the contributor section is necessary because increasing the Tier will change other parts of the user (Gems, items, pets) and so you want to ensure that none of the old data remains.

However the User ID is kept in the search box and the URL doesn't change so that you can easily re-fetch the data for the same user, in case you want to check their new settings or make other changes to their account.

The reload is necessary because the vue router can't handle "moving" from one URL to the same URL. The same thing happens if you load a user and then click the "Load User" button a second time (e.g., if you want to back out of any changes you made). That's this code: https://github.com/HabitRPG/habitica/pull/12652/files#diff-7a3e28ebec99ac2c419cdc471d62b162R60-R75

I'm sure there's a nicer way to do all this but I messed around with it for hours and this was the best option that covered all cases. If it was for the general user population it would be worth spending more time on it, but I'm not sure it's worth it for only the staff and mods.

website/client/src/components/admin-panel/user-support/index.vue
Alys: that's fixed

website/client/src/router/index.js
paglias: > However the User ID is kept in the search box and the URL doesn't change so that you can easily re-fetch the data for the same user, in case you want to check their new settings or make other changes to their account.

I'd like this to be changed, I'm fine with keeping the user id in the search box but if we go back to the main page the url should be changed accordingly

> "Save and Clear" for the contributor section is necessary because increasing the Tier will change other parts of the user (Gems, items, pets) and so you want to ensure that none of the old data remains.

I'm fine if we want to keep the Save and Clear but I'm pretty sure the api route used to update the data returns everything for that user so you could take the response and Object.assign(this.hero, response) to update it (but if it's too much hassle, Save and clear is totally fine for me)

website/server/controllers/api-v3/hall.js
```diff
@@ -310,16 +322,63 @@ api.updateHero = {
...
+ * @apiUse NoAccount
+ * @apiUse NoUser
+ * @apiUse NotAdmin
+ * @apiUse groupIdRequired
+ * @apiUse GroupNotFound
+ */
+api.getHeroParty = { // @TODO XXX add tests
+  method: 'GET',
+  url: '/hall/heroes/party/:groupId',
+  middlewares: [authWithHeaders(), ensureAdmin],
```
paglias: so for the routes here, we're adding the new privileges in the user schema but not using them? I mean all the routes here are checking contributor.admin, not one of the new privileges so they're still accessible from the api

paglias:
Looks like some tests are failing

paglias:
Now that https://github.com/HabitRPG/habitica/pull/12627 has been merged we should call the `common:setTitle` action when the admin panel is loaded to make sure it's a proper title

paglias:
@Alys what's the status of this?

Alys:
Not much progress recently. There's been too much bug support needed.
I'm still working on it when I can.

SabreCat:
@Alys Is this still going forward? What remains to be done?

Alys:
The code so far is working well for us.

As part of the API changes, I added a brief bit of preparation for improving/refactoring how mod/admin permissions are handled in future, thinking that we'd be able to do that slowly in bits and pieces. Matteo wanted it all done in this PR though, and it's a **lot** of work to do all at once so I kind of froze up and couldn't face it - there just aren't enough free hours for me to be able to handle it all as a single project. Sorry. :(  I should probably just remove the initial code that I'd added for that and just leave in the bare minimum of permission changes. I also need to make a small change to the API's admin route to prevent someone without the new permission giving that permission to any account.

SabreCat:
That sounds like a good plan! We can tackle a more comprehensive permissions update separately when we want/need to tackle it.

SabreCat:
@Alys Ready for another round of testing/review? Or are you still making modifications?

Alys:
@SabreCat The PR's current state is suitable for deploying to delta but I still have more code to write. If you wanted to deploy it now, that would reduce support hours and let shanaqui take back more of the support, but I know each additional PR in delta is added work, so if you'd rather wait that's fine. :)

SabreCat:
I don't think there's anything else on Delta at the moment, so up it goes!

SabreCat:
@Alys can you take a look at these conflicts?

Alys:
@SabreCat I think at least some of that conflict will disappear by itself once the Hall's current Display Name duplication issue is fixed in Staging. If there's any conflict left after that, I'll be happy to resolve it, but if I make changes before then, it'll probably result in more conflicts later.

Alys:
https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/hall/heroes.vue#L263-L275

https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/hall/heroes.vue#L281-L293

Those are right around the point where this PR adds a link to the admin panel, so the duplication will be interfering with git's assessment of the conflict.

SabreCat:
Ah, I thought the duplication was the result of the conflict, not the other way around--
I _think_ I've correctly removed the duplication in 094cf899d8fb51d6c3cee, but the `v-if` behavior was not acting the way I thought it would in the duplicated code to begin with, so I may still be confused?

SabreCat:
further fixage at 169634f205b2440096b9e514c875fcf0ac3b7246

SabreCat:
there was still one more commit after that, but I think now this is the actual conflict needing to be resolved..?

Alys:
For my reference, this was the final commit mentioned above: 9e740e70928a5fa313a0dfc2e9cadbc0af849d7f

Alys:
I've fixed the conflict by reverting the heroes.vue file to the version from `develop` (details about why are in the commit message but in brief, the changes I'd made to this file were merely an optional convenience).

SabreCat:
Superseded by #13843!

PR #12655, opened 2020-10-08
fix RTL punctuation & alignment

### LINKED ISSUE #11364: punctuation in wrong place in RTL text (Arabic, maybe others)
@sora-01 (30647f40-eeb7-4d3a-a52a-40491b3eab84) reports:

"Hi! I'm from the Pirate Cove. When I posted Arabic messages in a chat, I noticed periods at the end of a paragraph were displayed on the right like English. It's not the correct position, because periods are usually on the left in Arabic. The same problem happened to exclamation marks. This display problem happens on website and I'm using Google Chrome.
I don't really understand Arabic. I found this problem by chance, so I'm letting you know just to be sure. Thanks!"

(sora-01 was pasting Arabic text written by an Arabic speaker for the Pirate Cove guild (a guild that helps other guilds find new leaders and become more active).)

@citrusella added:

"The guild @sora-01 posted to appears to be [Arab Folks](https://habitica.com/groups/guild/6dd4603a-d4d0-4d20-be5b-955bfbb857ad). It seems all LTR text posted in a message done in RTL text could end up in a wrong place syntactually, and periods and exclamations are being treated as LTR text, I guess? (Question marks appear to be a different glyph in Arabic and thus unaffected.) I tried playing around with copying some text from [Arabic Wikipedia's Main Page's talk page](https://ar.wikipedia.org/wiki/%D9%86%D9%82%D8%A7%D8%B4:%D8%A7%D9%84%D8%B5%D9%81%D8%AD%D8%A9_%D8%A7%D9%84%D8%B1%D8%A6%D9%8A%D8%B3%D9%8A%D8%A9) to my party and Habitica treated the strings fairly differently where LTR characters of any kind were involved."

@sora-01 said:

"I posted those messages in my private guild and [Arab Folks](https://habitica.com/groups/guild/6dd4603a-d4d0-4d20-be5b-955bfbb857ad) as @citrusella says. To avoid confusion, I deleted periods at the end of a paragraph in the Arabic Hail for posting Arab Folks.  I also modified [Hails In Other Languages](https://habitica.fandom.com/wiki/The_Keep:Pirate_Cove/Hails_In_Other_Languages).  I tried to recreate the message in my private guild just now: https://snipboard.io/hB8jtd.jpg https://snipboard.io/3RN6DW.jpg _[images copied below]_

"I found a conversation like this in [Arabic Language Learners](https://habitica.com/groups/guild/245a5948-951a-496b-b595-a2c8d5558ad9) (one year ago): 'I did set up the Windows Arabic keyboard on my computer, but at least on Habitica it messes up the punctuation. I'm hoping that it's just Habitica that does that.' 
'Yeah that's a common problem I run into all the time. tbh it doesn't get better and it's always frustrating'"

hB8jtd
3RN6DW

mario130:

Fixes #11364

### Changes
Arabic text is now fixed in chat **(punctuation and alignment)** without affecting other languages.
image

----
UUID: 8eb6f2c2-ead3-4ec6-aa16-729f2f19185a


website/client/src/components/chat/chatCard.vue
```diff
@@ -278,6 +282,15 @@ export default {
...
     },
   },
+  beforeMount () {
+    const regexCheckLang = /^[\u0600-\u06FF]/;
+    const result = regexCheckLang.test(this.msg.text);
+    if (result) {
+      this.languageAlign = 'right';
+      this.languageDirection = 'rtl';
+    }
+    return console.log(this.msg.text);
```
paglias: this console.log() should be removed

website/client/src/components/chat/chatCard.vue
```diff
@@ -164,7 +166,7 @@
     .text {
       font-size: 14px;
       color: #4e4a57;
-      text-align: left !important;
+      text-align: left;
```
paglias: can we keep the !Important here?

paglias: Or it's not necessary at all given that one of `text-left` or `text-right` is going to be applied anyway?

website/client/src/components/chat/chatCard.vue
```diff
@@ -278,6 +282,15 @@ export default {
       return date.toString();
     },
   },
+  beforeMount () {
+    const regexCheckLang = /^[\u0600-\u06FF]/;
```
paglias: I'm wondering what's going to happen if someone posts a message with mostly latin characters but then adds just a word in Arabic for example.

Also I'm not familiar at all with RTL languages support but looking at links like https://stackoverflow.com/questions/12006095/javascript-how-to-check-if-character-is-rtl/14824756 I'm wondering if this regex is enough or if it'll only cover Arabic but not any other RTL language

website/client/src/components/chat/chatCard.vue
```diff
@@ -34,6 +34,8 @@
       <div
         ref="markdownContainer"
         class="text markdown"
+        :class="'text-' + languageAlign"
+        :style="'direction:' + languageDirection"
```
paglias: Can this be changed to use an object syntax? see https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax-1

paglias (commented):
thanks @mario130 ! Left some comment

website/client/src/components/chat/chatCard.vue
mario130: embarrassing lol

website/client/src/components/chat/chatCard.vue
mario130: Honestly I don't know the range of other rtl languages so I can't test them like Arabic

mario130: Also if a message has mixed languages it will be aligned to the right if it detects that the first character of the message is Arabic

website/client/src/components/chat/chatCard.vue
mario130: done

mario130:
@paglias  made some changes

paglias:
@mario130 I've talked about this with the team and:

- When we make this change it should cover all RTL languages, not just Arabic
- If there are mixed RTL and non-RTL languages we should change the alignment only if the first word is RTL

Given that every RTL language would require a specific check I think the easiest fix would be to use the html `dir` attribute set to `auto` (see https://www.w3schools.com/tags/att_global_dir.asp) It won't change the alignment but correctly set the right direction automatically

mario130:
@paglias I've made all the changes you mentioned including the auto direction and made the regular expression check for all RTL languages then I tested it for some languages like Hebrew and Persian.
Everything is working well, waiting for your review.

website/client/src/components/chat/chatCard.vue
```diff
@@ -278,6 +280,13 @@ export default {
       return date.toString();
     },
   },
+  beforeMount () {
```
paglias: thanks @mario130 , we discussed this internally and for now we'd like to limit the changes to the `dir` attribute, without changing the alignment, it's not a perfect solution but it's a start

paglias (changes requested):
left one comment, also could you revert the changes to package-lock.json? If you update your develop branch then `git checkout develop -- package-lock.json` will do the trick

mario130:
@paglias Did the change
But having trouble modifying the package-lock
I tried git pull then your command but it still has many changes how can I update from here?

paglias:
@mario130 I've pushed a commit that also fixes text alignment by setting it based on the `dir` attribute and also added the dir attribute to textareas

paglias:
@mario130  thanks for this PR! I'm merging the PR and awarding your contributor tier.

mario130:
Thanks, I will keep an eye on the issues for the coming weeks, happy to help :smile:

PR #12656, opened 2020-10-09
Fix keyboard inaccessible accordion in guild/party page fixes #12653

### LINKED ISSUE #12653: Accordion on guild/party page is inaccessible to keyboard navigation
Opening as advised in discussion for issue #11067

The accordion menu at the right hand side of the guild and party page uses clickable divs instead of buttons and cannot be accessed via keyboard navigation.

J-Sanderson:
# Fixes #12653

### Changes
The accordion menu on the right hand side of the guild/party page uses clickable divs to open and close items and is inaccessible to keyboard navigation. Replaced the divs with buttons and converted the nested divs holding the svg icons to spans, and styled to appear the same as before.

----

### DIFF
--- a/website/client/src/components/sidebarSection.vue
+++ b/website/client/src/components/sidebarSection.vue
@@ -22,7 +22,10 @@
           :target="tooltipId"
         />
       </div>
-      <SidebarButton :visible="visible" @click="toggle" />
+      <SidebarButton
+        :visible="visible"
+        @click="toggle"
+      />
     </div>
     <div
       v-show="visible"
--- a/website/client/src/components/sidebarSection.vue
+++ b/website/client/src/components/sidebarSection.vue
@@ -22,21 +22,21 @@
           :target="tooltipId"
         />
       </div>
-      <div
+      <button
         class="section-toggle ml-auto"
         @click="toggle"
       >
-        <div
+        <span
           v-if="visible"
           class="svg-icon"
           v-html="icons.upIcon"
-        ></div>
-        <div
+        ></span>
+        <span
           v-else
           class="svg-icon"
           v-html="icons.downIcon"
-        ></div>
-      </div>
+        ></span>
+      </button>
     </div>
     <div
       v-show="visible"
@@ -65,6 +65,8 @@
   }
 
   .section-toggle {
+    border: 0;
+    background: transparent;
     cursor: pointer;
   }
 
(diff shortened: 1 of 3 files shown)


website/client/src/components/sidebarSection.vue
```diff
@@ -22,21 +22,21 @@
           :target="tooltipId"
         />
       </div>
-      <div
+      <button
```
paglias: Looks like focused buttons get an outline 
Schermata 2020-10-10 alle 10 59 12
Schermata 2020-10-10 alle 10 59 51

I'm wondering, given that browsers might apply their own styles to standard buttons if it would be better to use a div with `aria-role="button"`

@negue do you feel like this should be extracted into its own component?

negue: div with `aria-role` should be fine for that.

And yes since this button already has 2 different icons and a custom style it really waits to be its own component 😁

J-Sanderson: I can set that up. Any naming conventions etc I should be aware of?

J-Sanderson: Update: I've set it to a div but it still has the focus ring. I can get rid of it, but will still need to give it some sort of focus style for accessibility. What would work for you?

paglias: @J-Sanderson the black ring as shown in my screenshot is fine, what i wanted to remove was the white outer ring

J-Sanderson: How does this look?

Screenshot from 2020-10-12 15-40-16

paglias: cc @Tressley

Tressley: This should be using focus states similar to the most recent Accessibility updates to the tasks page.

paglias: purple borders? the styles that you see in the screenshot are from the browser itself and they're used everywhere we haven't customized them like the tasks page
Schermata 2020-10-13 alle 09 59 11

Tressley: @paglias -- Ah, I could've been more clear in my initial reply. Yes, purple. Similar to the newest accessibility updates that were completed on the tasks controls.

J-Sanderson: Thanks for the clarification!

paglias:
@J-Sanderson sorry for the delay, the PR is looking good, noted towards your 2nd tier

PR #12659, opened 2020-10-11
fix "push to bottom" for user tasks in API

Alys:
The API's "push to bottom" feature for user tasks has been broken for a few years from the looks of it. This PR fixes it (confirmed by manual testing on a local install).

This PR also:
- Fixes the test for "push to bottom" which has been equally broken.
- Simplifies the position (0 / -1) information in the apidoc comment.
- Makes that information consistent in two similar routes.
- Replaces non-ascii characters in the apidoc comment, which were causing `é` to appear in https://habitica.com/apidoc/#api-Task-MoveTask

----

_Thanks to @khing (ccba770f-b79b-4fb6-8694-53b31bc27d88) for reporting this bug:_
_"How to move task to bottom?_
_In the [api-doc](https://habitica.com/apidoc/#api-Task-MoveTask) it says: "é -1 = bottom of the list". Don't know what does "é -1" mean. I tried -1, api success, but the task is not moved."_

### DIFF
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -794,9 +794,9 @@ api.scoreTask = {
  * @apiGroup Task
  *
  * @apiParam (Path) {String} taskId The task _id or alias
- * @apiParam (Path) {Number} position Where to move the task. 0 = top of the list.
- é                           -1 = bottom of the list.
- é                           (-1 means push to bottom). First position is 0
+ * @apiParam (Path) {Number} position Where to move the task.
+ *                                    0 = top of the list ("push to top").
+ *                                   -1 = bottom of the list ("push to bottom").
  *
  * @apiSuccess {Array} data The new tasks order for the specific type that the taskID belongs to.
  *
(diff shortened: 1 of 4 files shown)


Alys:
This can be tested with this API command:

`curl -X POST -H "x-api-user: b917f6f9-4080-46c6-b25f-4febb9775941" -H "x-api-key: xxxxxxxxxxxxxxx"  http://localhost:8080/api/v3/tasks/14eca9fc-09df-4b7c-a5ef-68a2a15a6c7e/move/to/-1`

where `14eca9fc-09df-4b7c-a5ef-68a2a15a6c7e` is the ID of any task that isn't already at the bottom of the list.

website/server/controllers/api-v3/tasks.js
```diff
@@ -837,9 +837,8 @@ api.moveTask = {
     pullQuery.$pull[`tasksOrder.${task.type}s`] = task.id;
     await user.update(pullQuery).exec();
 
-    // Handle push to bottom
     let position = to;
-    if (to === -1) position = [`tasksOrder.${task.type}s`].length - 1;
```
paglias, line 842: nice catch!

paglias (approved):

PR #12676, opened 2020-10-13
Loggly-only user support events

SabreCat:

### Changes

1. Per internal discussion how how to economize on Amplitude event volume while still providing data valuable to user support, this PR adds a parameter `loggerOnly` to the `analytics.track` function that, when passed `true`, calls `logger.info` as usual with `LOG_AMPLITUDE_EVENTS` on but skips sending the data to the Amplitude service itself.
2. Adds missing `headers` data to a few Analytics service function calls, so we can track e.g. the user's platform for those events.

### DIFF
--- a/website/server/libs/analyticsService.js
+++ b/website/server/libs/analyticsService.js
@@ -183,7 +183,7 @@ function _sendDataToAmplitude (eventType, data, loggerOnly) {
     logger.info('Amplitude Event', amplitudeData);
   }
 
-  if (loggerOnly) return new Promise();
+  if (loggerOnly) return Promise.resolve(null);
 
   return amplitude
     .track(amplitudeData)
--- a/website/server/libs/analyticsService.js
+++ b/website/server/libs/analyticsService.js
@@ -174,7 +174,7 @@ function _formatDataForAmplitude (data) {
   return ampData;
 }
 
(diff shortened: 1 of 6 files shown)


website/server/libs/analyticsService.js
```diff
@@ -183,6 +183,8 @@ function _sendDataToAmplitude (eventType, data) {
     logger.info('Amplitude Event', amplitudeData);
   }
 
+  if (loggerOnly) return new Promise();
```
paglias: @SabreCat I think instead of returning a promise, we could:

1. Transform the function in an `async` one and then just `return`
2. Otherwise we can keep it a normal function but instead of `new Promise()` do `Promise.resolve(returnValue)`

paglias (commented):
left one comment

website/server/libs/analyticsService.js
paglias: @SabreCat given the extra events being sent to amplitude with the a/b tests maybe we should finish this?

SabreCat: I went with `Promise.resolve`, is there anything more meaningful than `null` we could return?

paglias: I think null is fine if we don't have anything more meaningful to return

paglias (approved):

PR #12680, opened 2020-10-19
Prevent the removal of a quest owner from a group

### LINKED ISSUE #12398: party leader should not be able to remove the owner of the party's quest
_I'm marking this as suggestion-discussion for a few days for the staff to comment on whether my suggested fix is desired and then if there's no objections I'll mark it as help wanted._

Currently, it is possible for a Party leader to remove the Party member who is the owner of the Party's current Quest. This leaves the Party's `quest` object in an invalid state (see the sample data at the bottom of this post). As a side-effect of this, the former Quest owner is not able to use any third-party tools that read or display party data (because the data is invalid) and they probably see errors in Habitica's official clients too.

Another side effect is that the former Quest owner is no longer able to control the quest (e.g., they can't abort it) and that means they have lost the Quest scroll and cannot gain the Quest rewards. Since the Quest scroll may have cost Gems, this is a significant problem.

I believe the best fix is that the API's [/api/v3/groups/:groupId/removeMember/:memberId](https://habitica.com/apidoc/#api-Group-RemoveGroupMember) route must throw an error if the Party leader tries to remove the Quest owner. This should happen regardless of whether the Quest is in the invitation stage or is active.

The error message could say something like "You cannot remove this member because they are the owner of your Party's current Quest. If you abort the Quest first, you can then remove this member." 

I'll leave this as suggestion-discussion for a few days for the staff to comment on whether this is the desired fix and then if there's no objections I'll mark it as help wanted.

NB It would be possible for the `removeMember` route to be changed so that it automatically aborted the Quest before removing the former Quest owner, but I believe that would be less desirable. The Party leader may actually not want to remove the member once they realise the member is the Quest owner. 

---

As an example of the invalid data, below is a `quest` object after the Quest owner and sole Quest participant was removed from the Party. The Quest is active but there is no `quest.leader` or `quest.key` field.

```
    "quest": {
        "progress": {
            "collect": {},
            "hp": 100
        },
        "active": true,
        "members": {
            "0d0fd796-3271-43f0-b042-10bfd7ed3d97": true
        },
        "extra": {}
    },
```

thewilloftheshadow:

Fixes #12398

### Changes
This change will throw an error for the `/groups/:groupId/removeMember/:memberId` API route if the user that is being removed is the owner of the quest
This change adds a new string for the new error with the key `cannotRemoveQuestOwner`

----

### DIFF
--- a/test/api/unit/models/user.test.js
+++ b/test/api/unit/models/user.test.js
@@ -592,6 +592,36 @@ describe('User Model', () => {
   });
 
   context('pre-save hook', () => {
+    it('marks the last news post as read for new users', async () => {
+      const lastNewsPost = { _id: '1' };
+      sandbox.stub(NewsPost, 'lastNewsPost').returns(lastNewsPost);
+
+      let user = new User();
+      expect(user.isNew).to.equal(true);
+      user = await user.save();
+
+      expect(user.checkNewStuff()).to.equal(false);
+      expect(user.toJSON().flags.newStuff).to.equal(false);
+      expect(user.flags.lastNewStuffRead).to.equal(lastNewsPost._id);
+    });
+
+    it('does not mark the last news post as read for existing users', async () => {
+      const lastNewsPost = { _id: '1' };
+      const lastNewsPostStub = sandbox.stub(NewsPost, 'lastNewsPost');
+      lastNewsPostStub.returns(lastNewsPost);
+
+      let user = new User();
+      user = await user.save();
+
+      expect(user.isNew).to.equal(false);
+      user.profile.name = 'new name';
+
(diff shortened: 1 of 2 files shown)


paglias:
Hi @thewilloftheshadow , I'm not exactly sure what happened but looks like something broke with the PR, it got merged automatically and it's not showing any commit

paglias:
@thewilloftheshadow I think it's due to this app that you have installed on your fork https://github.com/apps/pull

It hard resets the develop branch to match the one here and your PR was made from there and not from a different branch so it got reset

thewilloftheshadow:
Well, that was the easiest PR I've ever had then! I've uninstalled that app now since I don't even use it, and remade the PR at #12695 @paglias

PR #12695, opened 2020-10-19
Prevent the removal of a quest owner from a group, take 2

### LINKED ISSUE #12398: party leader should not be able to remove the owner of the party's quest

thewilloftheshadow:
Remake of #12680 but with the Pull app uninstalled

Fixes #12398

### Changes
This change will throw an error for the `/groups/:groupId/removeMember/:memberId` API route if the user that is being removed is the owner of the quest
This change adds a new string for the new error with the key `cannotRemoveQuestOwner`

----


paglias (changes requested):
@thewilloftheshadow thanks for the PR! The code is looking good, could you add an integration test that covers this new behavior? See https://github.com/HabitRPG/habitica/blob/develop/test/api/v3/integration/groups/POST-groups_id_removeMember.test.js and let me know if you need any help

thewilloftheshadow:
It's been a while since I've been written any tests but I'll give it a go!

thewilloftheshadow:
Ok, I wrote a test, and it ran both on my machine and [on the CI tests](https://github.com/HabitRPG/habitica/pull/12695/checks?check_run_id=1286198229#step:8:621)

website/server/controllers/api-v3/groups.js
```diff
@@ -911,6 +911,8 @@ function _sendMessageToRemoved (group, removedUser, message, isInGroup) {
  * @apiError (400) {NotAuthorized} onlyLeaderCanRemoveMember Only the group
                                                              leader can remove members.
  * @apiError (400) {NotAuthorized} memberCannotRemoveYourself Group leader cannot remove themselves
+ * @apiError (400) {NotAuthorized} cannotRemoveQuestOwner Group leader cannot remove
```
paglias: the code here is 401 not 400, I think there was a mistake in the previous docs. NotAuthorized is 401, BadRequest is 400

test/api/v3/integration/groups/POST-groups_id_removeMember.test.js
```diff
@@ -251,6 +251,29 @@ describe('POST /groups/:groupId/removeMember/:memberId', () => {
...
+      await party.sync();
+
+      expect(party.quest.members[partyLeader._id]).to.be.true;
+      expect(party.quest.members[partyMember._id]).to.be.true;
+
+      await party.sync();
+
+      expect(leader.post(`/groups/${party._id}/removeMember/${partyMember._id}`))
+        .to.eventually.be.rejected.and.eql({
+          code: 403,
```
paglias: code should be 401,

```
      code: 401,
      error: 'NotAuthorized',
      message: t('cannotRemoveQuestOwner'),
```

paglias (changes requested):
two small issues then it's ready to go!

thewilloftheshadow:
Fixed, and I changed the other NotAuthorized codes to 401 in the apidoc headers as well

paglias:
@thewilloftheshadow  thanks for this PR! I'm merging the PR and awarding your contributor tier.

thewilloftheshadow:
Awesome, thanks! I definitely plan on sticking around, that was fun!

PR #12696, opened 2020-10-21
Make start date and day of month aware of timezones (fixes #12555)

### LINKED ISSUE #12555: "Monthly - Day of  the Month" Dailies should consider time zone during edits
I believe this bug only happens if your time zone is ahead of UTC (UTC+1, UTC+2, etc).

When you edit a Daily to change it to repeat on the same date each month, the player's time zone is not taken into consideration. Since the date due is taken from the Start Date and the Start Date is stored as a date in UTC, that means that the scheduled date is one day behind what it should be.

Note that this bug does not occur when you create a new Daily and set the monthly repeat options during the creation phase (more details about that below).

As a brief summary, I think the code to create a Daily has the correct method for setting `daysOfMonth` but the code to edit a Daily does not. _(I'm marking this as medium difficulty because it may be easily fixable by finding and comparing those two pieces of code. However I have not looked at the code myself recently so don't rely on this if it doesn't seem to help. :smiley:)_

Steps to reproduce the bug by editing an existing Daily:

1. Set your computer to be in a time zone that is ahead of UTC (mine is UTC+10 and that causes the bug to occur).
1. Create a Daily with the default settings (e.g., by using the "Add a Daily" text box at the top of the Dailies column.
1. Edit the Daily:
   * Ensure the Start Date is today (it should be by default)
   * Set the "Repeats" option to "Monthly"
   * Set "Repeat On" to "Day of the Month"
1. Save the Daily.
1. Observe that the Daily is NOT scheduled to be due, even though it should be.
1. View the data from the database, for example, see the data below which has had irrelevant fields removed:
 
```
{
    "_id": "3dedb8f6-03f2-407f-8cd3-08a0d2b46eea",
    "repeat": { "m": true, "t": true, "w": true, "th": true, "f": true, "s": true, "su": true },
    "frequency": "monthly",
    "everyX": 1,
    "nextDue": [
        "2020-10-08T14:00:00.000Z",
        "2020-11-08T14:00:00.000Z",
        ...
    ],
    "type": "daily",
    "text": "Wed 10am two-step",
    "startDate": "2020-09-08T14:00:00.000Z",  # This UTC date equates to 2020-09-09 in my time zone (UTC+10)
    "daysOfMonth": [
        8   # this is taken from the startDate and should be 9 for my time zone
    ],
    "weeksOfMonth": [],
    "createdAt": "2020-09-09T00:07:21.821Z",
    "updatedAt": "2020-09-09T00:07:59.176Z",
    "userId": "d904bd62-da08-416b-a816-ba797c9ee265",
    "isDue": false,
}
```
 

Steps to observe **correct** behaviour by creating a Daily with the monthly repeat option in a single step:

1. Set your computer to be in a time zone that is ahead of UTC (e.g., UTC+10).
1. Create a Daily by clicking the green diamond "+" button at the top right of the tasks area (just below the party member banner). Do NOT save the Daily immediately!
1. While still in the creation form, edit the Daily's settings:
   * Ensure the Start Date is today (it should be by default)
   * Set the "Repeats" option to "Monthly"
   * Set "Repeat On" to "Day of the Month"
1. Save the Daily.
1. Observe that the Daily IS correctly scheduled to be due today.
1. View the data from the database, for example, see the data below which has had irrelevant fields removed:
 
```
{
    "_id": "2fc939c6-f9a9-4678-9d8c-de464054d5a8",
    "repeat": { "m": true, "t": true, "w": true, "th": true, "f": true, "s": true, "su": true },
    "frequency": "monthly",
    "everyX": 1,
    "nextDue": [
        "2020-10-08T14:00:00.000Z",
        "2020-11-08T14:00:00.000Z",
        ...
    ],
    "type": "daily",
    "text": "wed 10am one-step",
    "startDate": "2020-09-08T14:00:00.000Z",  # This UTC date equates to 2020-09-09 in my time zone (UTC+10)
    "daysOfMonth": [
        9   # this is correct for my time zone
    ],
    "weeksOfMonth": [],
    "createdAt": "2020-09-09T00:08:28.166Z",
    "updatedAt": "2020-09-09T00:08:28.166Z",
    "userId": "d904bd62-da08-416b-a816-ba797c9ee265",
    "isDue": true,
}
```

bakerty:

Fixes #12555

### Changes
Bug was caused by updateTask api method setting `daysOfMonth = startDate.date()` and because `startDate` is represented in UTC, `daysOfMonth` would be wrong if user was in a UTC+xxx timezone.

Fixed this by factoring in `user.preferences.timezoneOffset`, which sets `daysOfMonth` to reflect the date in the user's timezone instead of UTC.

Also slightly tweaked the test for this api call so that the updated task start date was different from the original.

----
UUID: 76ad7ada-553e-46bc-a6dc-fcead49f9228

### DIFF
--- a/test/api/v3/integration/tasks/POST-tasks_user.test.js
+++ b/test/api/v3/integration/tasks/POST-tasks_user.test.js
@@ -8,9 +8,14 @@ import {
 
 describe('POST /tasks/user', () => {
   let user;
+  let tzoffset;
+
+  before(async () => {
+    tzoffset = new Date().getTimezoneOffset();
+  });
 
   beforeEach(async () => {
-    user = await generateUser();
+    user = await generateUser({ 'preferences.timezoneOffset': tzoffset });
   });
 
   context('validates params', async () => {
@@ -532,7 +537,7 @@ describe('POST /tasks/user', () => {
       expect(task.everyX).to.eql(5);
       expect(task.daysOfMonth).to.eql([15]);
       expect(task.weeksOfMonth).to.eql([3]);
-      expect(new Date(task.startDate)).to.eql(now);
+      expect(new Date(task.startDate)).to.eql(new Date(now.setHours(0, 0, 0, 0)));
       expect(task.isDue).to.be.true;
       expect(task.nextDue.length).to.eql(6);
     });
--- a/test/api/v3/integration/tasks/PUT-tasks_id.test.js
+++ b/test/api/v3/integration/tasks/PUT-tasks_id.test.js
@@ -10,9 +10,14 @@ import {
 
 describe('PUT /tasks/:id', () => {
   let user;
+  let tzoffset;
+
+  before(async () => {
+    tzoffset = (new Date()).getTimezoneOffset();
+  });
 
   beforeEach(async () => {
-    user = await generateUser();
+    user = await generateUser({ 'preferences.timezoneOffset': tzoffset });
   });
 
   context('validates params', () => {
@@ -503,7 +508,8 @@ describe('PUT /tasks/:id', () => {
     let monthly;
 
     beforeEach(async () => {
-      const date1 = moment.utc('2020-07-01').toDate();
+      // using date literals is discouraged here, daylight savings will break everything
+      const date1 = moment().toDate();
       monthly = await user.post('/tasks/user', {
         text: 'test monthly',
         type: 'daily',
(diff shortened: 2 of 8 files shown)


paglias:
Hi @bakerty , thanks for the PR.

I noticed this is also changing the value for `startDate`, I guess this is being done because it's not set correctly at task creation, maybe it should be fixed there instead of when updating the task?

bakerty:
Thanks for the response @paglias ,

`startDate` gets set properly on task creation. The reason for the change is that when the task gets updated with a new `startDate`, only the date changes and not the time, which could potentially cause a problem if the user changes timezones. This also only affects 'day of month' dailys

If this is actually the intended behaviour then I apologize and I'd be happy to remove that section of code and re-submit.

paglias:
Hmm, I'm going to ping @Alys to have an extra pair of eyes on this

Alys:
I believe changing the Start Date is good!

We used to tell people to change the Start Date to fix Dailies that had become rescheduled onto a different day by a time zone change, but I think that stopped working (fairly recently?). From what I can see, the change in this PR is good.

I wouldn't mind testing it on our delta test site for a few days if that's okay @paglias ?

paglias:
My point was that if there's a problem with the start date not being set correctly, it should be fixed at task creation, not later. Because right now it's not being set this way on creation. @Alys I'm putting this on delta

paglias:
@Alys have you had a chance to look at this?

Alys:
No. The past couple of weeks have had more RaB posts than usual and I've been helping with the excess emails. Testing this is still on my task list.

Alys:
@bakerty Sorry about the delay with my testing! I've tested this on our delta test site and it's great. It fixes https://github.com/HabitRPG/habitica/issues/12555 and changing the value for `startDate` works as you've described, and allows this workaround to be used again:

> We used to tell people to change the Start Date to fix Dailies that had become rescheduled onto a different day by a time zone change, but I think that stopped working

@paglias I'd be happy for this to go live whenever you think best.

paglias:
Thanks @Alys , before we merge this @bakerty I'd like the fix to be applied on task creation as well since it seems obvious that it's not accounting for time zone there, is this something you can add as well? Thanks

Alys:
> I'd like the fix to be applied on task creation as well since it seems obvious that it's not accounting for time zone there

Hmmm, yes. When you create a Daily through the website, your time zone IS accounted for and the Start Date has the correct time in it. When you create a Daily through the API, your time zone is NOT used when setting the Start Date. I didn't realise until now that the API wasn't doing that. This may account for some of the scheduling bugs we see in the mobile apps.

bakerty:
Alright I'll put the change in at task creation as well. Thank you both for all the feedback!

Should I extend the scope of the `startDate` modification to include all dailys, instead of only monthly ones, as well?

paglias:
I'd say all dailies would be better

bakerty:
Hey @Alys or @paglias, I'm looking for some advice here.

I added some code to `createTasks` in `website/server/libs/taskManager.js` that sets `startDate` to midnight in the user's timezone on task creation, and tweaked the relevant tests to test the new behaviour. All of it seems to be working as expected.

However, I ran into a rather major problem. If the user selects a start date far enough in the future, then it's possible that the user's timezone will change before that date thanks to daylight savings. This would mean that `startDate` might be inaccurate by 1 hour in either direction. The same problem occurs for monthly dailys, where some dates in `[nextDue]` are inaccurate because of daylight savings. This causes some test to not pass if they are run in a location where daylight savings happens.

The main issue is that it's impossible to figure out if a submitted start date is meant to be in daylight savings time or standard time server-side. I'm not really sure what the best way to proceed with this problem is. Any help would be greatly appreciated.

paglias:
Hmm let's see what @Alys thinks but i'd say that if it fixes the issue in most cases we should be fine, it's still an improvement on the current situation

paglias:
@Alys any update on this one?

Alys:
Sorry! I missed this and PR 12860 in the pre-Christmas chaos. I'll look through both this weekend.

Alys:
I think this is good as is!

> If the user selects a start date far enough in the future, then it's possible that the user's timezone will change before that date thanks to daylight savings. This would mean that startDate might be inaccurate by 1 hour in either direction.

That's definitely a problem but it always has been, it isn't made worse by this PR, and it's not something easily fixed. I guess we'd have to add a separate time zone field to each task to record the zone when it was created / modified, and then consider that when scheduling the task, but that's far out of scope for this issue.

@bakerty Thanks very much for this fix and I'm so sorry it took me this long to reply!

paglias:
@bakerty  thanks for this PR! I'm merging the PR and awarding your contributor tier.

PR #12700, opened 2020-10-24
Fix Tasks styles

### LINKED ISSUE #12704: Heading level hierarchy is not accurately reflected by heading appearance (e.g., h5 larger than h3)
_The non-italics words below are taken direct from @citrusella's post in the Report a Bug guild. I've added [my own comments in italics and square brackets]._

---

Have been seeing how used to the new font changes I can get and noticing something that feels weird/confusing/possibly-buggy/unintended.

The h1, h2, h3, h4, h5, h6 hierarchy looks like this:

image

_[Below is the markdown code that produced those headings, with blank lines removed to save space.]_
```
# 1
## 2
### 3
#### 4
##### 5
###### 6
```

Is it deliberate that some in the middle are smaller? h3 feels especially confusing--it's the same size (or very slightly smaller) and weight/color (or again, very slightly different) as regular text and nigh indistinguishable from it, which is throwing me off every time I look at my task list because I have "header tasks" for different sections of my task list and now they just... look like not-due tasks. (I'm not sure "header tasks" is a super common use case, but it is one highlighted on the wiki--I was using h3 because it was small enough to not wrap on the majority of the header tasks I'd made and large enough as to be visually distinguishable from standard text.)

For example, here's h3 vs regular text (there's some spacing difference if the lines wrap but that's about it--the color is a little more distinguishable here than it is on tasks, BTW):

image

_[Below is the markdown code that produced those two lines.]_
```
### This is a third-level header, it's quite small, the size of regular text or so.

This is regular text. It is the size of regular text because it IS regular text.
```

I'm on the lookout for other things that look odd with the changes but I think most of the other things I go "weird, don't like that" with are just things I'm not used to yet.

---

_[I'm labelling this as needs design for a design-related decision. @Tressley ]_

negue:
- re-add previous markdown h3/h4 styles 
- remove double transition properties

Fixes #12704

### DIFF
--- a/website/client/src/assets/scss/markdown.scss
+++ b/website/client/src/assets/scss/markdown.scss
@@ -15,7 +15,11 @@
   }
 
   h3 {
+    font-family: 'Roboto Condensed', sans-serif;
+    font-size: 16px;
     margin-bottom: 4px;
+
+    line-height: 1.5;
     margin-top: 6px;
     color: $gray-10;
   }
@@ -27,6 +31,9 @@
   h4 {
     font-family: 'Roboto', sans-serif;
     font-weight: bold;
+    font-size: 14px;
+    line-height: 1.43;
+    color: $gray-100;
   }
 
   a {
--- a/website/client/src/assets/scss/markdown.scss
+++ b/website/client/src/assets/scss/markdown.scss
@@ -4,14 +4,17 @@
   }
 
   h1 {
+    font-size: 24px;
+    line-height: 1.67;
     margin-bottom: 10px;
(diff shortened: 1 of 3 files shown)


paglias:
can we push it to delta instead of beta? It's much more used, beta has a separate database and it's supposed to be used only when testing something that requires a different database

negue:
Oh, sure! I just wanted to pick an un-used one so that we don't have so many conflicts all the time.

Pushed to delta :)

paglias:
LGTM 👍

PR #12715, opened 2020-10-26
Node 14

paglias:


paglias:
cc @Alys @SabreCat @negue

Alys:
Thanks for pinging me! I've updated the wiki docs to say node 14 is preferred and that node 12 will stop working later so everyone should upgrade when they can. I've posted to Aspiring Blacksmiths about it too.

PR #12716, opened 2020-10-26
Feature/enhance get memebers for a challenge fixes #12481

### LINKED ISSUE #12481: enhance "Get members for a challenge" API route to include tasks
Currently, the API route [Get members for a challenge (/api/v3/challenges/:challengeId/members)](https://habitica.com/apidoc/#api-Member-GetMembersForChallenge) returns:
- the participants in a challenge
- their public profile data if `includeAllPublicFields` is set to `true`

In addition to its current behaviour, we it to have a new optional Boolean parameter called something like `includeTasks`
If that parameter is set to true, then each participant's tasks from the challenge should be included in the returned data. 

This will make it much faster for third-party tools to retrieve all the data needed to allow a challenge owner to judge the challenge.

---

I've labelled this "medium level coding" because I _think_ it should be not too hard since you can look at how the `includeAllPublicFields` option was set up, and can look at the task routes for fetching task data, but it may be at the more difficult end of medium. If you take this up and run into trouble, feel free to ask for help!

hamboomger:

Fixes #12481

### Changes
- New `includeTasks` query parameter added to the GET `/challenges/:challengeId/members` route
- Minor code refactor in the `members.js` module - separated into 2 commits to ease the CR process

----
UUID: 

### DIFF
--- a/test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js
+++ b/test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js
@@ -244,7 +244,7 @@ describe('GET /challenges/:challengeId/members', () => {
         expect(task).to.include.all.keys(['type', 'value', 'priority', 'text', '_id', 'userId']);
         expect(task).to.not.have.any.keys(['tags', 'checklist']);
         expect(task.challenge.id).to.be.equal(challenge._id);
-        expect(task.userId).to.be.equal(member.id);
+        expect(task.userId).to.be.equal(member._id);
       });
     });
   });
--- a/test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js
+++ b/test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js
@@ -117,26 +117,7 @@ describe('GET /challenges/:challengeId/members', () => {
     expect(res[0].profile).to.have.all.keys(['name']);
   });
 
-  it('returns only first 30 members if req.query.includeAllMembers is not true and req.query.limit is undefined', async () => {
-    const group = await generateGroup(user, { type: 'party', name: generateUUID() });
-    const challenge = await generateChallenge(user, group);
-    await user.post(`/challenges/${challenge._id}/join`);
-
-    const usersToGenerate = [];
-    for (let i = 0; i < 31; i += 1) {
-      usersToGenerate.push(generateUser({ challenges: [challenge._id] }));
-    }
-    await Promise.all(usersToGenerate);
-
-    const res = await user.get(`/challenges/${challenge._id}/members?includeAllMembers=not-true`);
-    expect(res.length).to.equal(30);
-    res.forEach(member => {
-      expect(member).to.have.all.keys(['_id', 'auth', 'flags', 'id', 'profile']);
-      expect(member.profile).to.have.all.keys(['name']);
-    });
-  });
-
-  it('returns only first 30 members if req.query.includeAllMembers is not defined and req.query.limit is undefined', async () => {
+  it('returns only first 30 members if req.query.limit is undefined', async () => {
     const group = await generateGroup(user, { type: 'party', name: generateUUID() });
     const challenge = await generateChallenge(user, group);
     await user.post(`/challenges/${challenge._id}/join`);
@@ -216,25 +197,6 @@ describe('GET /challenges/:challengeId/members', () => {
       expect(member.profile).to.have.all.keys(['name']);
     });
   }).timeout(30000);
-
-  it('returns all members if req.query.includeAllMembers is true', async () => {
-    const group = await generateGroup(user, { type: 'party', name: generateUUID() });
-    const challenge = await generateChallenge(user, group);
-    await user.post(`/challenges/${challenge._id}/join`);
-
-    const usersToGenerate = [];
-    for (let i = 0; i < 31; i += 1) {
-      usersToGenerate.push(generateUser({ challenges: [challenge._id] }));
-    }
-    await Promise.all(usersToGenerate);
-
-    const res = await user.get(`/challenges/${challenge._id}/members?includeAllMembers=true`);
-    expect(res.length).to.equal(32);
-    res.forEach(member => {
-      expect(member).to.have.all.keys(['_id', 'auth', 'flags', 'id', 'profile']);
-      expect(member.profile).to.have.all.keys(['name']);
-    });
-  });
 
   it('supports using req.query.lastId to get more members', async function test () {
     this.timeout(30000); // @TODO: times out after 8 seconds
--- a/test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js
+++ b/test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js
@@ -259,6 +259,32 @@ describe('GET /challenges/:challengeId/members', () => {
     expect(resIds).to.eql(expectedIds.sort());
   });
 
+  it('supports using req.query.includeTasks in order to add challenge-related tasks of all members', async () => {
+    const group = await generateGroup(user, { type: 'party', name: generateUUID() });
+    const challenge = await generateChallenge(user, group);
+    await user.post(`/challenges/${challenge._id}/join`);
+
+    const usersToGenerate = [];
+    for (let i = 0; i < 8; i += 1) {
+      usersToGenerate.push(generateUser({ challenges: [challenge._id] }));
+    }
+    await Promise.all(usersToGenerate);
+    await user.post(`/tasks/challenge/${challenge._id}`, [{ type: 'habit', text: 'Some task' }]);
+    await user.post(`/tasks/challenge/${challenge._id}`, [{ type: 'daily', text: 'Some different task' }]);
+
+    const res = await user.get(`/challenges/${challenge._id}/members?includeTasks=true`);
+    expect(res.length).to.equal(9);
+    res.forEach(member => {
+      expect(member).to.have.property('tasks');
+      expect(member.tasks).to.be.an('array');
+      expect(member.tasks).to.have.lengthOf(2);
+      member.tasks.forEach(task => {
+        expect(task).to.include.all.keys(['type', 'text', 'id', 'challenge', 'group']);
+        expect(task.challenge.id).to.be.equal(challenge._id);
+      });
+    });
+  });
+
   it('supports using req.query.search to get search members', async () => {
     const group = await generateGroup(user, { type: 'party', name: generateUUID() });
     const challenge = await generateChallenge(user, group);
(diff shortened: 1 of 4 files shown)


hamboomger:
Hey, I apologise in advance for the possible test failures, I still have problems with running them locally so I decided it would be faster if the CI tool will run them for me :)

website/server/controllers/api-v3/members.js
```diff
@@ -402,6 +402,71 @@ function _getMembersForItem (type) {
   };
 }
 
+async function _handleGetMembersForChallenge (req, res) {
```
hamboomger: In the first commit I just copied all the members-related content from the handleGetMembersForItem (:285) without changing anything inside this function for readability purposes, but considering how small the change actually is it might be a better idea to squash this commit with the next one. What do ya think?

paglias: Don't worry about the commits, we squash them anyway when merging the pr

website/server/controllers/api-v3/members.js
```diff
@@ -402,6 +362,109 @@ function _getMembersForItem (type) {
...
+    .limit(limit)
+    .select(fields)
+    .lean()
+    .exec();
+
+  // manually call toJSON with minimize: true so empty paths aren't returned
+  members.forEach(member => User.transformJSONUser(member, addComputedStats));
+  if (req.query.includeTasks === 'true') {
+    const tasksPremises = members.map(
+      member => _getMemberTasksFromChallenge(challengeId, member.id, user, res),
```
paglias: at this point we already have the challenge and group objects, can we pass them to `_getMemberTasksFromChallenge` to avoid having to re-fetch them for every member?

website/server/controllers/api-v3/members.js
```diff
@@ -402,6 +362,109 @@ function _getMembersForItem (type) {
...
+
+  const members = await User
+    .find(query)
+    .sort({ _id: 1 })
+    .limit(limit)
+    .select(fields)
+    .lean()
+    .exec();
+
+  // manually call toJSON with minimize: true so empty paths aren't returned
```
paglias: I don't think this comment is necessary

website/server/controllers/api-v3/members.js
paglias: I also don't think that fetching the member is necessary in `_getMemberTasksFromChallenge` as we already fetched all of them a few lines above here

website/server/controllers/api-v3/members.js
```diff
@@ -402,6 +362,109 @@ function _getMembersForItem (type) {
...
+
+  const challengeTasks = await Tasks.Task.find({
+    userId: memberId,
+    'challenge.id': challengeId,
+  })
+    .select('-tags') // We don't want to return the tags publicly TODO same for other data?
+    .exec();
+
+  return challengeTasks.map(task => {
+    task.checklist = []; // Clear checklists as they are private
```
paglias: I know this is old code but I was wondering if maybe instead of manually removing the checklist data it wouldn't be easier to simply not return it from the database like it's done for tags

website/server/controllers/api-v3/members.js
```diff
@@ -402,6 +362,109 @@ function _getMembersForItem (type) {
...
+  // optionalMembership is set to true because even if you're
+  // not member of the group you may be able to access the challenge
+  // for example if you've been booted from it, are the leader or a site admin
+  const group = await Group.getGroup({
+    user, groupId: challenge.group, fields: '_id type privacy', optionalMembership: true,
+  });
+  if (!group || !challenge.canView(user, group)) throw new NotFound(res.t('challengeNotFound'));
+  if (!challenge.isMember(member)) throw new NotFound(res.t('challengeMemberNotFound'));
+
+  const challengeTasks = await Tasks.Task.find({
```
paglias: I was thinking, since the number of results could be pretty high here we should probably use `.lean()` to make sure the server doesn't incur in performance issues

website/server/controllers/api-v3/members.js
paglias: I'd like to change this so that instead of calling `_getMemberTasksFromChallenge ` for every challenge member (creating a lot of promises and database queries) if we could transform it so that only one call is necessary, as I said challenge and group data has already been fetched, same for members.

For the tasks it'd be enough to make a single query with `userId:{$in: anArrayOfAllMembersIds}, `

paglias (changes requested):
@hamboomger thanks for the PR! I left some comments about trying to reduce the number of , other than that I think that it might be a good idea to move the logic (_handleGetMembersForChallenge) to a new file under website/server/challenges because the controller is getting too big.

website/server/controllers/api-v3/members.js
hamboomger: The only problem I see is that it uses custom `.toJson()` schema method which replaces `_id` with `id` and (probably?) makes some other changes to the data. I can create some util method like `modelToJson(options)` and use it inside of both [`schema.options.toJson.transform` function](https://github.com/HabitRPG/habitica/blob/9213ee92de43d58c3980d6f89d87c052da627e93/website/server/libs/baseModel.js#L60) and this function, but maybe you have any other ideas?

hamboomger:
> @hamboomger thanks for the PR! I left some comments about trying to reduce the number of , other than that I think that it might be a good idea to move the logic (_handleGetMembersForChallenge) to a new file under website/server/challenges because the controller is getting too big.

Maybe you mean under `website/server/libs/challenges`?

website/server/controllers/api-v3/members.js
paglias: That's something we can revisit in future but for now let's skip this part, in this case the custom toJSON doesn't add anything really necessary

paglias:
> Maybe you mean under website/server/libs/challenges?

yeah sorry

website/server/controllers/api-v3/members.js
hamboomger: Just to clarify, each member of a challenge has the same set of tasks related to the challenge, so there is no need to go over each member and fetch his tasks separately, right?
Update: nevermind, I forgot that the route should also provide task completion for each member

website/server/libs/challenges/handleGetMembersForChallenge.js
```diff
@@ -0,0 +1,92 @@
...
+    .sort({ _id: 1 })
+    .limit(limit)
+    .select(fields)
+    .lean()
+    .exec();
+
+  members.forEach(member => User.transformJSONUser(member, addComputedStats));
+  if (req.query.includeTasks === 'true') {
+    const memberIdToTasksMap = await getMembersTasksForChallenge(members, challenge);
+    for (const member of members) {
```
paglias: instead of having two loops here could you add the tasks to the member inside the first forEach?

```javascript
  const includeTasks = req.query.includeTasks === 'true';
  let memberIdToTasksMap;
  if (includeTasks) {
    memberIdToTasksMap = await getMembersTasksForChallenge(members, challenge);
  }
  members.forEach(member => {
       User.transformJSONUser(member, addComputedStats);
       if (includeTasks) member.tasks = memberIdToTasksMap[member._id]; // id might not be returned when using lean()
  });
```

something like this

website/server/libs/challenges/handleGetMembersForChallenge.js
```diff
@@ -0,0 +1,92 @@
...
+  if (req.query.search) {
+    query['auth.local.username'] = { $regex: req.query.search };
+  }
+
+  if (lastId) query._id = { $gt: lastId };
+
+  let limit = req.query.limit ? Number(req.query.limit) : 30;
+
+  // Allow for all challenges members to be returned
+  if (req.query.includeAllMembers === 'true') {
```
paglias: Actually since we're here let's remove this parameter `req.query.includeAllMembers `. it was supposed to be removed a long time ago and could cause serious perf issues

paglias: It's a breaking change but it's not used anywhere

website/server/controllers/api-v3/members.js
```diff
@@ -402,6 +363,25 @@ function _getMembersForItem (type) {
   };
 }
 
+async function _getMemberTasksFromChallenge (challenge, member, user, res) {
```
paglias: since this is only used by one route I don't think it makes sense to have it an separate method, unless we can merge this and `getMembersTasksForChallenge` in a single one

test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js
```diff
@@ -259,6 +259,33 @@ describe('GET /challenges/:challengeId/members', () => {
...
+    const res = await user.get(`/challenges/${challenge._id}/members?includeTasks=true`);
+    expect(res.length).to.equal(9);
+    res.forEach(member => {
+      expect(member).to.have.property('tasks');
+      expect(member.tasks).to.be.an('array');
+      expect(member.tasks).to.have.lengthOf(2);
+      member.tasks.forEach(task => {
+        expect(task).to.include.all.keys(['type', 'value', 'priority', 'text', '_id', 'userId']);
+        expect(task).to.not.have.any.keys(['tags', 'checklist']);
+        expect(task.challenge.id).to.be.equal(challenge._id);
```
paglias, line 246: can we add a check that `task.userId` matches the member's id?

paglias (changes requested):
@hamboomger thanks about this! I left some comments but they're pretty minor, once those are fixed we can merge this one

website/server/libs/challenges/handleGetMembersForChallenge.js
hamboomger: It's up to you to decide, but maybe it would be better to have the `includeTasks`-related logic to be enclosed in a separate block? The algorithmic complexity is still the same with both solutions, I just thought that it might be a tiny bit more readable this way.

website/server/libs/challenges/handleGetMembersForChallenge.js
hamboomger: Got it, I'll remove this along with test suite and swagger api declaration

test/api/v3/integration/challenges/GET-challenges_challengeId_members.test.js
hamboomger: sure

website/server/libs/challenges/handleGetMembersForChallenge.js
paglias: The big-O complexity is indeed the same but since we're talking about small numbers having 2 separate loops can result in an important performance hit so i'd prefer if we could limit the loops to one, thanks!

hamboomger: Ok, no problem :)

paglias:
Looks like an integration test if failing

paglias:
Thanks @hamboomger , this is looking good and will be merged soon

paglias (approved):

paglias:
Noted in your profile!

PR #12736, opened 2020-10-29
Task and party menus use normal links (fixes #12703)

### LINKED ISSUE #12703: Tasks and Party menu items should be normal links
The "Tasks" and "Party" items in the website's navigation menu are not normal links.

This means you can't right-click on them to open them in a new tab.

It also means you can't hover over them to see the URL they'll take you to - you see just the URL of the page you are currently on (screenshot below).

image

_Thanks to Nakonana and citrusella for reporting this._

J-Sanderson:
Fixes #12703

### Changes
Fixes the problem with the task and party menu items using a different component and not functioning correctly as links, by using the router-link component instead of b-nav-item. Note b-nav-item is still used for the link item where the user is not part of a party as this pops up a modal to create one rather than going to another page.

----

### DIFF
--- a/website/client/src/components/header/menu.vue
+++ b/website/client/src/components/header/menu.vue
@@ -46,15 +46,18 @@
         class="collapse navbar-collapse"
       >
         <b-navbar-nav class="menu-list">
-          <b-nav-item
-            class="topbar-item"
-            :class="{'active': $route.path === '/'}"
-            tag="li"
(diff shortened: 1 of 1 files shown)


paglias:
@J-Sanderson I replied in the issue because I can't replicate the bug in the first place

paglias:
Sorry @J-Sanderson but I'm going to close the PR since it looks like it fixed itself

PR #12737, opened 2020-10-29
Fixes #12730 - Tag filter not always cleared correctly when deleting active tag

### LINKED ISSUE #12730: Tag filter not always cleared correctly when deleting active tag
As reported by @citrusella:

**"Deleting a tag while it's active (filtered on) causes the tag to at least appear to un-select (I'm not sure if it actually does) and show all your tasks again... but selecting another tag afterwards causes the newly-selected tag to appear empty even if it's not, until you use the clear all button."**

I was able to reproduce that a couple of times but then when I went to write up steps to reproduce, I couldn't. @citrusella then investigated further and posted:

**"I can reproduce it, but I think I've determined it happens only if the tag panel doesn't disappear/close (or "clear all" isn't selected) before I select the next tag. That happens a lot easier when something like Tags Always Visible is on (because the panel is wider and its behavior means I take more care to hover over it consistently when doing tag things like that), but it's also pretty easy to have happen without that because I just made the disappear finding in a private window without user styles or scripts active once I thought it might be a possibility (tested twice in a private window: the first time I made sure the tag panel didn't close, the second time I made sure it did)."**

(Note that the [Tags Always Visible](https://habitica.fandom.com/wiki/Tags_Always_Visible) browser user script and style don't cause the bug. When I reproduced it, I had both of them disabled, as did citrusella in her private window test.)

jbusa22:
Fixes #12730 

### Changes
A quick fix to remove deleted tag from selected from temporarilySelectedTags so tag filtering works correctly after a tag is deleted and the menu is still open.

----


paglias:
thanks @jbusa22 ! noted towards your 2nd tier

PR #12758, opened 2020-11-02
update dropdowns to use the updated / styled components

negue:
Updates all dropdowns to use the new `selected` style

### DIFF
--- a/website/client/src/components/ui/filterDropdown.vue
+++ b/website/client/src/components/ui/filterDropdown.vue
@@ -3,8 +3,7 @@
     <span class="dropdown-label">{{ label }}</span>
     <select-list :items="items"
                  :value="selectedItem"
-                 class="array-select"
-                 :class="'inline'"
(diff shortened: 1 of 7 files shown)


negue:
Those are done on the "style" part, code review left :)

website/client/src/components/ui/filterDropdown.vue
```diff
@@ -1,35 +1,32 @@
...
-      <b-dropdown-item
-        v-for="item in items"
-        :key="item.id"
-        :active="selectedItem.id === item.id"
-        @click="selectItem(item)"
-      >
+    <select-list :items="items"
+                 :value="selectedItem"
+                 class="array-select"
+                 :class="'inline'"
```
paglias: inline seems a static class value, can we move it into the normal class list?

paglias (commented):
left one comment about a minor thing to change :)

PR #12779, opened 2020-11-11
Change $type of date for todos

### LINKED ISSUE #8405: API: dates may be invalid

thewilloftheshadow:
Fixes #8405 

### Changes
Changes the $type of the startDate in the task schema to "Date" instead of "String"
Changes the apidocs to reflect this change
Adds a migration script to fix old tasks
Adds a test to check for this new functionality

----

### DIFF
--- a/migrations/20201111_api_date.js
+++ b/migrations/20201111_api_date.js
@@ -1,0 +1,69 @@
+/*
+ * Fix dates in the database that were stored as $type string instead of Date
+ */
+/* eslint-disable no-console */
+
+const MIGRATION_NAME = '20201111_api_date';
+
+import * as Tasks from '../website/server/models/task';
+
+const progressCount = 1000;
+let count = 0;
+
+async function updateUser (user) {
+  count++;
+
+  let set = {};
+  let newDate = new Date(user.date);
+  if(isValidDate(newDate)) {
+    set = {
+      'date': newDate
+    }
+  } else {
+    set = {
+      'date': ''
+    }
+  }
+
+  set.migration = MIGRATION_NAME;
+
+  if (count % progressCount === 0) console.warn(`${count} ${user._id}`);
+  return await Tasks.Task.update({_id: user._id}, {$set: set}).exec();
+}
+
+module.exports = async function processUsers () {
+  let query = {
+    migration: {$ne: MIGRATION_NAME}
+  };
+
+  const fields = {
+    _id: 1,
+    items: 1,
+  };
+
+  while (true) { // eslint-disable-line no-constant-condition
+    const users = await Tasks.Task // eslint-disable-line no-await-in-loop
+      .find(query)
+      .limit(250)
+      .sort({_id: 1})
+      .select(fields)
+      .lean()
+      .exec();
+
+    if (users.length === 0) {
+      console.warn('All appropriate tasks found and modified.');
+      console.warn(`\n${count} tasks processed\n`);
+      break;
+    } else {
+      query._id = {
+        $gt: users[users.length - 1],
+      };
+    }
+
+    await Promise.all(users.map(updateUser)); // eslint-disable-line no-await-in-loop
+  }
+};
+
+function isValidDate(d) {
+  return Object.prototype.toString.call(d) === "[object Date]" && !isNaN(d.getTime());
+}
--- a/migrations/archive/2020/20201111_api_date.js
+++ b/migrations/archive/2020/20201111_api_date.js
@@ -5,28 +5,20 @@
 
 const MIGRATION_NAME = '20201111_api_date';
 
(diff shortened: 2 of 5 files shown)


migrations/20201111_api_date.js
```diff
@@ -0,0 +1,69 @@
+/*
```
paglias: for the lint errors here, just move the file to migrations/archive/2020

migrations/20201111_api_date.js
```diff
@@ -0,0 +1,69 @@
...
+  if (count % progressCount === 0) console.warn(`${count} ${user._id}`);
+  return await Tasks.Task.update({_id: user._id}, {$set: set}).exec();
+}
+
+module.exports = async function processUsers () {
+  let query = {
+    migration: {$ne: MIGRATION_NAME}
+  };
+
+  const fields = {
```
paglias: this is going to select only _id and items (which is a user field, not a task one)

migrations/20201111_api_date.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  set.migration = MIGRATION_NAME;
+
+  if (count % progressCount === 0) console.warn(`${count} ${user._id}`);
+  return await Tasks.Task.update({_id: user._id}, {$set: set}).exec();
+}
+
+module.exports = async function processUsers () {
+  let query = {
```
paglias: this query will find all tasks and unfortunately that would take ages to finish as we have more than 60 milion ones, can we restrict it so that only the affected tasks (those with a date field using mongo's $exist parameter) are targeted?

paglias: also migration can be removed as it's only used for users

migrations/20201111_api_date.js
```diff
@@ -0,0 +1,69 @@
...
+    set = {
+      'date': newDate
+    }
+  } else {
+    set = {
+      'date': ''
+    }
+  }
+
+  set.migration = MIGRATION_NAME;
```
paglias: let's remove this one

migrations/20201111_api_date.js
```diff
@@ -0,0 +1,69 @@
...
+        $gt: users[users.length - 1],
+      };
+    }
+
+    await Promise.all(users.map(updateUser)); // eslint-disable-line no-await-in-loop
+  }
+};
+
+function isValidDate(d) {
+  return Object.prototype.toString.call(d) === "[object Date]" && !isNaN(d.getTime());
```
paglias: I think the first check will always return true as you passed the date through `new Date`, the second one should be enough (!isNaN(d.getTime()))

paglias (changes requested):
left some comments on the migration, the rest looks good :)

migrations/archive/2020/20201111_api_date.js
```diff
@@ -0,0 +1,62 @@
...
+
+  let set = {};
+  const newDate = new Date(user.date);
+  if (isValidDate(newDate)) {
+    set = {
+      date: newDate
+    }
+  } else {
+    set = {
+      date: ''
```
paglias: here I'd unset the field to `undefined` instead of setting it to an empty string

paglias (changes requested):
@thewilloftheshadow I left one comment but otherwise it's looking good, as soon as that's fixed we'll decide on when to run the migration

thewilloftheshadow:
Done!

paglias:
Thanks @thewilloftheshadow , the code is looking good, we're going to decide when it's best to run the migration and will merge the code then but consider this one done :)

migrations/archive/2020/20201111_api_date.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+  };
+
+  if (count % progressCount === 0) console.warn(`${count} ${user._id}`);
+  return await Tasks.Task.update({_id: user._id}, {$set: set}).exec();
+}
+
+module.exports = async function processUsers () {
+  let query = {
+    date: {$exists: true}
```
paglias, line 27: @thewilloftheshadow can we limit the query to `type: 'todo'`? the date field exists only for them

paglias (changes requested):

thewilloftheshadow:
Done! :)

paglias (approved):

paglias:
I've added a fix because without passing `type: 'todo'` in the update op the update didn't work, I've also reduced the number of updates that get performed by making one only if the date is invalid

paglias:
@SabreCat I think this one still need the migration to be run

paglias:
thanks @thewilloftheshadow ! sorry for the delay and welcome to the 2nd tier :)

PR #12783, opened 2020-11-12
partial fix for Issue 12760 Title gets updated after profile modal closes

### LINKED ISSUE #12760: HTML page Title tag does not change back after viewing Profile modal
As reported by @BellaBelinda (88a47bf3-e4bb-4952-9415-afb22148dd9e):

**"If I'm in a guild or the Tavern or even my tasks page and I click on someone's profile, the tab name thing changes to "Profile | User | Habitica", like usual, but when I click back out of the profile, it doesn't change back to the name of wherever I was before. It can be fixed by reloading,"**

It will also happen on any other page if you access a Profile  modal by clicking on a party member's avatar or your own avatar.

aevix:

Fixes #12760 

### Changes
@Alys It is still throwing some errors but its doing what we are wanting to do so I will link this PR. Its a bit of a long explanation sorry... So I finally understand what @paglias meant when he said this is caused by router not being standard. Modals are handled through bootstrap vue and so when it opens the modal the whole app doesn't need to re render since nothing changed in the app. So I tried to create a new router for the profile modal thinking its just another component that needs to be hooked in but that got very messy and not exactly doing what I thought it would. So I went back to see how the profile component was being routed (not the modal) and tried just mimicking what it was doing but with no luck, it wasn't reacting at all until I was about to give up and tried adding this.$router.push() instead of .go(-1) and it finally updated the title! I am not sure 100% but some errors are being thrown and it may have something to do with promise with routing but if someone knows what is going on feel free to let me know. If someone has better idea or want me to change anything let me know also.

----
UUID: 228f739f-b42c-4801-932c-10b4d739b460


website/client/src/components/userMenu/profileModal.vue
```diff
@@ -54,8 +61,40 @@ export default {
     },
     onHidden () {
       if (this.$route.path !== window.location.pathname) {
-        this.$router.go(-1);
+        this.$router.push({ path: this.$route.path });
```
paglias: Hi @aevix , I'm seeing a few problems with this, instead of trying to recreate the previous title , my suggestion would be to store it in a variable before it's set [here](https://github.com/HabitRPG/habitica/blob/316ed391e8c92894bdafe31a41bb8ec6d1e3f418/website/client/src/components/userMenu/profile.vue#L902) and then restore it when `onHidden` is called. I made an example here https://github.com/HabitRPG/habitica/issues/12760#issuecomment-725464599

website/client/src/components/userMenu/profileModal.vue
```diff
@@ -35,15 +35,22 @@ export default {
...
+    },
+  },
   mounted () {
     this.$root.$on('habitica:show-profile', data => {
       this.userId = data.userId;
       this.startingPage = data.startingPage || 'profile';
       this.path = data.path;
       this.$root.$emit('bv::show::modal', 'profile');
     });
+    this.selectPage(this.startingPage);
```
paglias: why is this necessary?

paglias: Actually I'm not seeing where this function is declare or `this.selectedPage` used

website/client/src/components/userMenu/profileModal.vue
aevix: I am not sure if I am not doing something wrong but I have tried this already and it won't change the title. Maybe you could try it and see if it works on your end. Maybe its the browser I am using but I tested it on safari and google chrome. document.title from the app.js is being triggered and working via `common:setTitle` but I think like I mentioned in the issue, vue router has some sort of disconnect between dom being rendered and the actual browser updating. Only way I got it to update was by changing it to push.

website/client/src/components/userMenu/profileModal.vue
aevix: Oh I must have left it by accident I was testing some code and forgot to take it out. Ill push a commit without that function.

website/client/src/components/userMenu/profileModal.vue
paglias: could you post the commit where you tried that and I'll give it a look? thanks

aevix: I made a commit sorry it took so long to respond I didn't get a chance to sit down and work on it until now.

website/client/src/components/userMenu/profileModal.vue
```diff
@@ -35,6 +35,7 @@ export default {
       userId: undefined,
       startingPage: undefined,
       path: undefined,
+      selectedPage: '',
```
paglias: this can be removed I think

website/client/src/components/userMenu/profile.vue
```diff
@@ -838,6 +839,11 @@ export default {
   mounted () {
     this.loadUser();
     this.selectPage(this.startingPage);
+    this.$root.$on('restoreTitle', () => {
```
paglias: to make sure we don't keep this event listener even after the component is not being used it should be removed before the component is destroyed with the following hook(you can place it right after the `mounted` hook

```
  beforeDestroy () {
    this.$root.$off('habitica: restoreTitle');
  },
```

website/client/src/components/userMenu/profile.vue
```diff
@@ -899,6 +905,7 @@ export default {
     selectPage (page) {
       this.selectedPage = page || 'profile';
       window.history.replaceState(null, null, '');
+      this.oldTitle = this.$store.state.title;
```
paglias: I haven't tested this @aevix but my guess is that `selectPage` is called not only the first time the modal is opened but also when switching, for example, from Stats to Achievements, in which case it would store the wrong title. I think this should be called in `mounted` so it's called just once

website/client/src/components/userMenu/profileModal.vue
```diff
@@ -55,6 +56,7 @@ export default {
     onHidden () {
       if (this.$route.path !== window.location.pathname) {
         this.$router.go(-1);
+        this.$root.$emit('restoreTitle');
```
paglias: to follow internal conventions, can you rename the event `habitica: restoreTitle`?

paglias (changes requested):
@aevix left some minor comments but it's looking good! It's a very clean and concise solution

paglias:
Also there seems to be an issue with the branch, it should target `develop` not `delta` (but if it's an issue don't worry I'll fix it before merging)

aevix:
@paglias This commit doesn't solve the issue. I am not sure if you have tried running this but on my browser title doesn't get updated like it should. I just created the commit to show that this way doesn't seem to work I am not sure why it won't update. My previous commit f35030c does update the title but it throws errors on the console. Do you still want me to edit the commit with your comments? Also, I tried creating this PR for the develop but for some reason it was saying my file didn't match the develop version and it was trying to change hundreds of files on this PR but when I switched to delta it only changed what I edited so you might have to fix it before merging it.

paglias:
@aevix I got confused and hadn't tested it earlier, I tested it today and with a few changes it's going to work!

In particular:

- The old title was being saved after the new was already set by `this.startingPage()`
- The event was being fired after the `hidden` event which meant the profile component had already been updated.

This is the diff of the changes I applied (I also changed the event name to `habitica:restoreTitle` without the whitespace):

```

**Comment by paglias** [comment]
@aevix I got confused and hadn't tested it earlier, I tested it today and with a few changes it's going to work!

In particular:

- The old title was being saved after the new was already set by `this.startingPage()`
- The event was being fired after the `hidden` event which meant the profile component had already been updated.

This is the diff of the changes I applied to make it work locally (I also changed the event name to `habitica:restoreTitle` without the whitespace):

```
matteo@macbook-matteo habitica % git diff
diff --git a/website/client/src/components/userMenu/profile.vue b/website/client/src/components/userMenu/profile.vue
index 01dfd1e498..bb33920c1c 100644
--- a/website/client/src/components/userMenu/profile.vue
+++ b/website/client/src/components/userMenu/profile.vue
@@ -782,7 +782,7 @@ export default {
       content: Content,
       user: null,
       userLoaded: false,
-      oldTitle: '',
+      oldTitle: null,
     };
   },
   computed: {
@@ -838,16 +838,18 @@ export default {
   },
   mounted () {
     this.loadUser();
-    this.selectPage(this.startingPage);
     this.oldTitle = this.$store.state.title;
-    this.$root.$on('habitica: restoreTitle', () => {
-      this.$store.dispatch('common:setTitle', {
-        fullTitle: this.oldTitle,
-      });
+    this.selectPage(this.startingPage);
+    this.$root.$on('habitica:restoreTitle', () => {
+      if (this.oldTitle) {
+        this.$store.dispatch('common:setTitle', {
+          fullTitle: this.oldTitle,
+        });
+      }
     });
   },
   beforeDestroy () {
-    this.$root.$off('habitica: restoreTitle');
+    this.$root.$off('habitica:restoreTitle');
   },
   methods: {
     async loadUser () {
diff --git a/website/client/src/components/userMenu/profileModal.vue b/website/client/src/components/userMenu/profileModal.vue
index 251cb3ba75..06b2611cc1 100644
--- a/website/client/src/components/userMenu/profileModal.vue
+++ b/website/client/src/components/userMenu/profileModal.vue
@@ -4,6 +4,7 @@
     size="lg"
     :hide-footer="true"
     :hide-header="true"
+    @hide="beforeHide"
     @hidden="onHidden"
     @shown="onShown()"
   >
@@ -52,10 +53,14 @@ export default {
     onShown () {
       window.history.pushState('', null, this.path);
     },
+    beforeHide () {
+      if (this.$route.path !== window.location.pathname) {
+        this.$root.$emit('habitica:restoreTitle');
+      }
+    },
     onHidden () {
       if (this.$route.path !== window.location.pathname) {
         this.$router.go(-1);
-        this.$root.$emit('habitica: restoreTitle');
       }
     },
   },
(END)
```

paglias:
Unfortunately fixing the PR base was not possible (you probably created your branch from the `delta` one instead of develop originally) so I merged it manually https://github.com/HabitRPG/habitica/commit/df1e4af7fc244e929c900a842048cf39f8dead61 with a direct commit.

@aevix thanks for this PR! I'm awarding your contributor tier.

aevix:
@paglias Thank you so much for showing me what you did. I was completely stumped on this issue and probably bit off a little more than I could chew lol. I will continue to help out any way I can 😄

PR #12791, opened 2020-11-18
Miscellaneous lint fixes: move watch & mounted properties; apply automatic fixes

Alys:
This applies lint's automatic fixes and it moves some watch and mounted properties to different positions as advised in lint's warnings.

### DIFF
--- a/website/client/src/components/challenges/challengeDetail.vue
+++ b/website/client/src/components/challenges/challengeDetail.vue
@@ -384,16 +384,6 @@ export default {
       memberResults: [],
     };
   },
-  watch: {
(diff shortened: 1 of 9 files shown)


paglias:
thanks @Alys ! Btw some suggested changes from eslint get applied automatically when running `npm run lint`

Alys:
> Btw some suggested changes from eslint get applied automatically when running npm run lint

Yes, that's how I found these ones. :) They were applied automatically to my admin panel branch after I ran lint there, and I didn't want them in a commit in that branch since they were unrelated.

PR #12793, opened 2020-11-20
Update Party / Group Sidebar / Quest states

negue:
update / restyle the sidebars / quests

Todo / done:
- [x] Menu 
- [x] Sections (larger divider)
- [x] Quest States
  - [x] Member / No Quest
  - [x] Quest Leader / Quest not started
  - [x] Member/Quest Invite Pending
  - [x] Collection Quest/Quest Owner Participating
  - [x] Not participating
  - [x] Boss Quest
  - [x] Boss Rage
- [x] No Challenge


Tressley:
This is ready for review.

website/client/package-lock.json
```diff
@@ -10094,6 +10094,12 @@
           "resolved": "https://registry.npmjs.org/emoji-regex/-/emoji-regex-8.0.0.tgz",
           "integrity": "sha512-MSjYzcWNOA0ewAHpz0MxpYFvwg6yjy1NG3xteoqz644VCo/RPgnr1/GGt+ic3iJTzQ8Eu3TdM14SawnVUmGE6A=="
         },
+        "emojis-list": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/emojis-list/-/emojis-list-3.0.0.tgz",
+          "integrity": "sha512-/kyM18EfinwXZbno9FyUGeFh87KC8HRQBQGildHZbEuRyWFOmv1U10o9BBp8XVZDVNNuQKyIGIu5ZYAAXJ0V2Q==",
+          "optional": true
+        },
```
SabreCat: These package-lock updates don't look relevant. Is any of this actually meant to be part of the PR? If not, this file can be removed.

website/client/src/components/groups/questSidebarSection.vue
```diff
@@ -143,26 +113,43 @@
...
                     you don't underestimate damage needed to end quest-->
                 </span>
               </div>
               <div
-                v-if="userIsOnQuest"
+                v-if="userIsOnQuest && user.party.quest.progress.up"
                 class="col-6"
               >
                 <!-- @TODO: Why do we not sync quest
                   progress on the group doc? Each user could have different progress.-->
```
SabreCat, line 151: I have noticed this doesn't update on a click of the sync button--user needs to refresh the whole page to get updates.

website/client/src/components/groups/questSidebarSection.vue
```diff
@@ -143,26 +113,43 @@
...
-                  class="float-right"
-                >{{ (user.party.quest.progress.up || 0) | floor(10) }} {{ $t('pendingDamageLabel') }}</span> <!-- eslint-disable-line max-len -->
+                <span class="float-right pending-value">
+                  <div
+                    class="svg-icon sword-icon"
+                    v-html="icons.swordIcon"
+                  ></div>
+                  {{ (user.party.quest.progress.up || 0) | floor(10) }}
+                  {{ $t('pendingDamageLabel') }}
+                </span> <!-- eslint-disable-line max-len -->
```
SabreCat: What is this for? There's no long line...

website/client/src/components/groups/questSidebarSection.vue
```diff
@@ -6,11 +6,11 @@
     >
       <div class="col-12 text-center">
         <div
-          class="svg-icon"
+          class="svg-icon quest-icon color"
           v-html="icons.questIcon"
```
SabreCat, line 10: Icons can be `v-once`, yeah?

website/client/src/components/groups/questSidebarSection.vue
```diff
@@ -248,35 +303,88 @@
...
+    }
+
+    .quest-icon {
+      width: 4.25rem;
+      height: 4.25rem;
+    }
+  }
+
+  .full-width {
+    width: 100%;
```
SabreCat: Doesn't Bootstrap's `w-100` already do this?

website/client/src/components/groups/rightSidebar.vue
```diff
@@ -0,0 +1,231 @@
...
+          <b-dropdown
+            right="right"
+            toggle-class="with-icon"
+            class="ml-2"
+            :no-caret="true"
+          >
+            <template v-slot:button-content>
+              <span
+                class="svg-icon inline menuIcon"
+                v-html="icons.menuIcon"
```
SabreCat, line 29: Icons can be `v-once` in this file too

website/client/src/components/groups/rightSidebar.vue
```diff
@@ -0,0 +1,231 @@
...
+              >
+              </span>
+            </template>
+            <b-dropdown-item
+              v-if="isLeader && !group.purchased.active && group.privacy === 'private'"
+              class="selectListItem"
+              @click="$emit('upgradeGroup')"
+              :style="{'--hover-color': '#1ca372', '--hover-background': '#77f4c73F'}"
```
SabreCat: why `:style` and hardcoded `#color` strings?

website/client/src/components/sidebarButton.vue
```diff
@@ -28,13 +28,18 @@
     background: transparent;
     cursor: pointer;
     &:focus {
+      // Fix keyboard inaccessible #12656
```
SabreCat: Should this be an issue link in the PR instead of a code comment?

SabreCat (changes requested):

website/client/package-lock.json
negue: I don't really know why those we're added - let me delete it and reinstall it

website/client/src/components/groups/questSidebarSection.vue
negue: need to check how we did the other sync button refreshes again - wonder why its not working "out of the box" here too

website/client/src/components/groups/questSidebarSection.vue
negue: the previous code had it :D   I refactored it - check the previous code on this line - it needed this ignore - will be deleted now

website/client/src/components/groups/questSidebarSection.vue
negue: oh here yeah - will do

website/client/src/components/groups/questSidebarSection.vue
negue: can try it - didn't know about the `w-100` 

for some of the bootstraps classes I tend not to use them - some just can't be bend to make the design perfect :D 


but the `w-100` will probably work then - gonna check if there are other places with that as well

website/client/src/components/groups/rightSidebar.vue
negue: need to revisit that - the hover color seems to be useable as one of the $scss vars but the hovher background with the special alpha channel might still end up as not a $scss var ... but lemme see when I get to the part

website/client/src/components/sidebarButton.vue
negue: Either I copied from a different place - or the merge of develop added this - I don't think I wrote that on purpose :D 

I can replace it with a github link - but I also need to find the source of that as well

website/client/package-lock.json
negue: when I delete and let it reinstall it 

it re-adds all of it again - I'll just use the develop version but its still odd that it changes these 


maybe we need to go stricter on the packages for that, e.g.
`   "uuid": "8.3.2",`
instead of
`   "uuid": "^8.3.2",`

website/client/src/components/sidebarButton.vue
negue: oh it was at first a conflict and then I added the comment to know why outline was needed - now I remember hah

website/client/src/components/groups/rightSidebar.vue
negue: the colors were changed - the hover backgrounds doesn't have any specific scss var

website/client/src/components/groups/questSidebarSection.vue
negue: ok soo the `party` property is inside the fetch catch ... so I wonder what can be done to make it refresh ?

website/client/src/components/groups/questSidebarSection.vue
negue: I updated it but github doesn't show it o.O thats odd

website/client/src/components/groups/questSidebarSection.vue
SabreCat: There's a `RESYNC_COMPLETED` event that gets emitted after the fetch. Maybe update a key on the component to prompt Vue to rewrite it?

SabreCat:
Bouncing back to you for now--you're probably better equipped than I to work through these merge conflicts, and we might still be able to fix that sync button

website/client/src/components/groups/questSidebarSection.vue
negue: for example `user.party....` its all from the state - which gets updated before `RESYNC_COMPLETED` 

I just don't know where / how I should reset the property  if its already from the store

negue:
fixed conflicts less see if the CI goes through

website/client/src/components/groups/questSidebarSection.vue
SabreCat: Looks like Tressley pulled this out to a GitHub ticket for us, so we can look into it later!

mornir:
@negue Thanks for this update! 💜 So great to have all that information neatly displayed on the website.

PR #12820, opened 2020-11-27
use or clause for display name and username search

### LINKED ISSUE #12729: Guild member list currently allows searching by Display Name only
As reported by @Nakonana (33bb14bd-814d-40cb-98a4-7b76a752761c):

"... when searching for a user via the guild member list, one can only search by Display Name and not by Username. I find searching by username much more practical because people change their display name frequently while they usually stick with their username once it's established :)"

There's two ways we could fix this:

1. ~Allow searching by Username only.~
2. **Allow searching by either Username or Display Name.**

I suggest option 2 (Username and Display Name) because that will make it much easier when searching for players who have a default hb-12345... Username. (Note though that challenges allow searching by Username only.)

~I'm marking this as suggestion-discussion but if there's no disagreement to that in a week, I'll mark it as help wanted for Option 2 (Username and Display Name).~
(If that's done, we could consider enabling Display Name searching for challenges too.).

jbusa22:

Fixes #12729

### Changes
Just changed the query to an or query for display name and username

----

### DIFF
--- a/website/server/controllers/api-v3/members.js
+++ b/website/server/controllers/api-v3/members.js
@@ -358,7 +358,7 @@ function _getMembersForItem (type) {
       if (req.query.search) {
         // Creates a RegExp expression when querying for profile.name
         const escapedSearch = escapeRegExp(req.query.search);
-        query['profile.name'] = { $regex: new RegExp(escapedSearch, 'i') };
+        query['$or'] = [{'profile.name': { $regex: new RegExp(escapedSearch, 'i') }}, {'auth.local.username': { $regex: req.query.search }}];
       }
     } else if (type === 'group-invites') {
       if (group.type === 'guild') { // eslint-disable-line no-lonely-if
--- a/website/server/controllers/api-v3/members.js
+++ b/website/server/controllers/api-v3/members.js
@@ -359,8 +359,8 @@ function _getMembersForItem (type) {
         // Creates a RegExp expression when querying for profile.name
         const escapedSearch = escapeRegExp(req.query.search);
         query.$or = [
-          { 'profile.name': { $regex: new RegExp(escapedSearch, 'i') } }, 
-          { 'auth.local.username': { $regex: req.query.search } }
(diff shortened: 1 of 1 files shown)


jbusa22:
I've tested this briefly with a couple of differently named users and it appears to work great. However, I think the code in this function specifically could use cleanup but this might not be the place.

website/server/controllers/api-v3/members.js
```diff
@@ -358,7 +358,10 @@ function _getMembersForItem (type) {
       if (req.query.search) {
         // Creates a RegExp expression when querying for profile.name
```
paglias: can we update the comment here to mention usernames?

website/server/controllers/api-v3/members.js
```diff
@@ -358,7 +358,10 @@ function _getMembersForItem (type) {
...
         const escapedSearch = escapeRegExp(req.query.search);
-        query['profile.name'] = { $regex: new RegExp(escapedSearch, 'i') };
+        query.$or = [
+          { 'profile.name': { $regex: new RegExp(escapedSearch, 'i') } },
+          { 'auth.local.username': { $regex: req.query.search } },
```
paglias: is there a reason why this isn't using `new RegExp(escapedSearch, 'i')` like the profile name search?

paglias (changes requested):
left a couple of comments

website/server/controllers/api-v3/members.js
jbusa22: I just copied it from the challenge username search about 10 lines above, but presumably, it should be using RegExp not sure if there was a reason.

paglias: let's make it like the profile one, the escaping might not be necessary here but the `i` flag is useful

paglias:
thanks @jbusa22 , this is looking good and will be merged in the next few days!

paglias (approved):

paglias:
@jbusa22 welcome to the 2nd tier!

PR #12836, opened 2020-12-03
added scroll bar to tag filter

### LINKED ISSUE #12613: can't scroll through tag filter screen with keyboard when not enough tasks are shown
If you have a lot of tags, then when you select one that has no tasks or only a few tasks assigned to it, you cannot use the keyboard arrow keys to scroll all the way down the tag filter panel. You can still use the mouse/trackpad to scroll.

Reported by @shewhoknits (e1d43f13-4a7b-4188-8844-127b4f63c67c), reproduced by me. Firefox on Ubuntu for me.

Steps to reproduce:
1. Give yourself very many tags. I'll paste my set of test tags into a separate comment here. It exactly mimics shewhoknits's tags except with the names changed for privacy. I don't know if that many are needed, but a small set of tags does not allow the bug to be reproduced.
1. Reload the website to ensure it has picked up the latest tag data for your account.
1. Go into the tag filter panel and select one tag. Ensure that it is a tag that does not have any tasks assigned. (The same bug occurs if you have only a small number of tasks assigned to it but not if you have a lot of tasks.)
1. Reopen the tag panel if it closed itself (e.g., if you accidentally moved your mouse too much).
1. Attempt to scroll down through the tag filter panel using the arrow keys on your keyboard. Observe that you can't scroll all the way down.
1. Attempt to scroll down through the tag filter panel using your mouse or trackpad. Observe that you can scroll all the way down.

The desired behaviour is that scrolling with the arrow keys should allow you to get to the bottom of your tags list, just like the mouse does. (This is useful for, for example, people with manual dexterity problems who have trouble using a mouse.)

_I've labelled this as "medium level coding" but that's a bit of a guess. If you're not reasonably confident you could work out how to solve this, you may want to pick a different issue._

hazwong:
Fixes #12613

Added a scroll bar to the tag filter screen such that when tags exceed a certain number, the height of the tag window is fixed and a scroll bar appears. As a result, the tag panel doesn't extend beyond the page when a lot of tags are added. The user can click within the tags window to scroll through the tags list with either a mouse or the up/down arrow keys. They can click outside the tags window if they want to scroll up/down the tasks page.

UUID: cc74fe22-6903-434d-a60f-ee9cd29deb9f

Screenshot 2020-12-04 at 12 43 40 AM

### DIFF
--- a/website/client/src/components/tasks/user.vue
+++ b/website/client/src/components/tasks/user.vue
@@ -300,6 +300,11 @@
       }
     }
 
+    .tags-list {
+      max-height: 400px;
+      overflow-y: scroll;
+    }
+
     .tag-edit-input {
       border-bottom: 1px solid $gray-500 !important;
 


website/client/src/components/tasks/user.vue
```diff
@@ -300,6 +300,11 @@
       }
     }
 
+    .tags-list {
+      max-height: 400px;
+      overflow-y: scroll;
```
paglias, line 305: @hazwong does this work with `overflow-y: auto; `? so that the scrollbar doesn't appear if not necessary?

hazwong: @paglias Currently, the scrollbar only appears when the tags list reaches a height of greater than 400px. The scrollbar doesn't appear if it's not necessary.

paglias:
I'm sorry @hazwong but it looks like the issue is not happening anymore. That being said we have updated designs for the tags component in case you want to work on them

SabreCat:
Closing this PR, looks like it's no longer needed. Thanks for jumping in on it, though, @hazwong !

PR #12854, opened 2020-12-08
Fixed "@Username" mentions in a private message so that it properly links immediately

### LINKED ISSUE #12764: @Username in private message does not behave as a link until after a reload
As reported by @citrusella:

"If you type and send an @ mention in your PMs (for instance, to have a direct reference to someone's profile), then even if it's typed right, it has all the elements of a mention except... the link. The link happens, but you can only access it if you visit a different area of Habitica entirely (not just another PM) and then go back to the PM, at which point it's properly linked."

Confirmed.

Steps to reproduce:

1. Send a new PM to someone (to yourself if you wish). Include an @Username in it.
1. Observe that the Username looks like a link - e.g. the screenshot below.
1. Hover over the Username. Observe that it has incorrect behaviour:
   * your mouse cursor does NOT change to an arrow
   * the profile URL for the Username does not appear in your browser's status bar
   * the Username is not clickable
1. Reload the website OR go to a different page and then back to your private messages.
1. Hover over the Username. Observe that the Username now behaves correctly:
   * your mouse cursor change to an arrow
   * the profile URL for the Username appears in your browser's status bar
   * the Username is clickable and opens the user's profile page in a new tab

image

kevechen:

Fixes #12764 

### Changes
Previously, if you sent a private message to a user with a "@Username" in the text, you wouldn't be able to click on that mention link immediately after the message was sent. It would look like it's clickable, but you would have to refresh the page or go to a different page and come back for that mention link to be available. This change fixes that problem so that immediately after you send a text with a mention, you would be able to click on it, which would lead to that user's profile page (image provided below for context). 

In terms of specific code changes, I added several lines that would update the message properties with the correct text (after the @ text was processed with a link) once the response came back.

mention_ex

----
UUID: 127290bf-75df-4d7d-a94f-8136cb7d7032


website/client/src/pages/private-messages.vue
```diff
@@ -882,12 +882,11 @@ export default {
     },
     sendPrivateMessage () {
       if (!this.newMessage) return;
-
       const messages = this.messagesByConversation[this.selectedConversation.key];
 
       messages.push({
         sent: true,
-        text: this.newMessage,
+        text: this.newMessage, // will update if text has @mention link
```
paglias: can we remove this comment?

website/client/src/pages/private-messages.vue
```diff
@@ -914,7 +913,7 @@ export default {
         this.initiatedConversation = null;
       }
 
-      this.selectedConversation.lastMessageText = this.newMessage;
+      this.selectedConversation.lastMessageText = this.newMessage; // update when result comes back
```
paglias: this one too

website/client/src/pages/private-messages.vue
```diff
@@ -925,7 +924,12 @@ export default {
       }).then(response => {
         const newMessage = response.data.data.message;
         const messageToReset = messages[messages.length - 1];
-        messageToReset.id = newMessage.id; // just set the id, all other infos already set
+
+        // Update newest message with id and text if there's an @ mention
```
paglias: @kevechen I think there's actually a bug at line 933 with `Object.assign(messages[messages.length - 1], messageToReset);` since it's assigning an object to itself (`messageToReset` is `messages[messages.length - 1];`) so I think the easiest way to fix this would be to change the code to something like this:

```
}).then(response => {
  const newMessage = response.data.data.message;
  this.selectedConversation.lastMessageText = newMessage.text;
  Object.assign(messages[messages.length - 1], newMessage);
  this.updateConversationsCounter += 1;
});
```

paglias (changes requested):
Hi @kevechen , I left some comments!

website/client/src/pages/private-messages.vue
kevechen: @paglias I've modified it with your suggestion, but the message does not show up correctly. It might be because the response's message has a different data structure than the message we're trying to set (results below -- this is a conversation from testing1 to testing2).

*With those changes:*
Screen Shot 2020-12-13 at 1 23 05 PM

*After reloading the page:*
Screen Shot 2020-12-13 at 1 23 17 PM

I think the reason we had messageToReset was to get the original structure of a single message first, then set the changes, and copy it over. We may not even need that variable, so maybe this is better:
```
}).then(response => {
        const newMessage = response.data.data.message;
        messages[messages.length - 1].id = newMessage.id;
        messages[messages.length - 1].text = newMessage.text;
        this.selectedConversation.lastMessageText = newMessage.text;
        this.updateConversationsCounter += 1;
});
```

kevechen:
@paglias Ahh my bad, I forgot to fetch before pushing my changes! It seems I added a bunch of other changes and commits that were merged from other PRs. Should I close this PR and open another one?

website/client/src/pages/private-messages.vue
paglias: Hmm I think that could be because Vue by default doesn't recognize new properties. Can you fix the commits or open a new PR? If we change the object completely it should work, so something like `messages[messages.length - 1] = newMessage`

kevechen: @paglias I made a new PR #12883 with the suggested changes, so I'll go ahead and close this one!

PR #12857, opened 2020-12-10
Winter 2021

SabreCat:

### Changes

Winter Wonderland 2020-2021 content, and some substantial refactoring of the related subscription promotion.

All components of the event should turn on and off at the beginning and end of the event without manual intervention or deployment. Where possible, this decision comes from the `world-state` API.

* Seasonal variations of NPC images
* Buyable date ranges for quests, quest bundle, magic hatching potions, gear, and snowballs
* Opening and closing of the Seasonal Shop
* Avatar images in Change Class modal

### Known Issues and Questions

* Some item descriptions are incomplete. They shouldn't throw "string missing" or similar errors, but may appear blank. This will be addressed when the relevant content is written.
* Some date ranges are hard-coded--can these spots access the `world-state`? (Or maybe just the EVENTS constants...)
* To get the dynamic URLs for NPC images to load correctly, I moved them from `/client/src/assets/images` to `/client/public/static`. Was this necessary, or is there another way?
* If a user pins a limited-time item, it will remain pinned but unpurchasable at the end of the event. If a solution isn't found in time, we can remove these via migration.


website/server/libs/payments/paypal.js
```diff
@@ -162,6 +163,8 @@ api.checkoutSuccess = async function checkoutSuccess (options = {}) {
 
     data.paymentMethod = 'PayPal (Gift)';
     data.gift = gift;
+
+    promoSubscription(data);
```
paglias: I feel like there could be some issues with this:

1. There's no error handling
2. The subscription is applied before the payment is done (while usually it's done later, see line 173 here)

I would maybe move the application of the promo sub to `payments/subscriptions#createSubscription` so that it also avoids having to change the code for each payment provider

website/client/src/components/header/banners/giftPromo.vue
```diff
@@ -0,0 +1,107 @@
...
+    :show="showGiftPromoBanner"
+    height="3rem"
+  >
+    <div
+      slot="content"
+      :aria-label="$t('subscription')"
+      class="content d-flex justify-content-around align-items-center ml-auto mr-auto"
+      @click="showSelectUser"
+    >
+      <div
```
paglias, line 13: v-once for icons / static text / ...

website/client/src/components/header/banners/giftPromo.vue
```diff
@@ -0,0 +1,107 @@
...
+  data () {
+    return {
+      icons: Object.freeze({
+        gifts,
+      }),
+    };
+  },
+  methods: {
+    showSelectUser () {
+      /* Analytics.track({
```
paglias: do we want this?

website/client/src/components/inventory/stable/index.vue
```diff
@@ -644,6 +645,10 @@ export default {
     anyFilterSelected () {
       return Object.values(this.viewOptions).some(g => g.selected);
     },
+    mattClass () {
```
paglias: I think maybe some of these helper functions could go in a mixin? not necessary but I see they're used often

paglias: Something like a method that given a npc name returns the correct class?

website/client/src/components/settings/subscription.vue
```diff
@@ -283,22 +283,36 @@
...
-      >
+    <div class="d-flex justify-content-center align-items-start">
+      <div class="d-flex flex-column align-items-center mt-4">
+        <div
+          v-once
+          class="svg-icon svg-gift-box m-auto"
+          v-html="icons.giftBox"
+        >
+        </div>
+        <div class="muted mx-auto mt-3 mb-1">
```
paglias, line 300: v-once missing in a few places in this file

website/common/script/content/hatching-potions.js
```diff
@@ -311,12 +319,14 @@ const premium = {
...
     text: t('hatchingPotionAurora'),
     limited: true,
+    _addlNotes: t('eventAvailabilityReturning', {
+      availableDate: t('dateEndJanuary'),
+      previousDate: t('decemberYYYY', { year: 2019 }),
+    }),
+    event: EVENTS.winter2021,
     canBuy () {
-      return moment().isBetween('2019-12-19', '2020-02-02');
+      return moment().isBetween('2020-12-22T08:00-04:00', '2021-01-31T20:00-04:00');
```
paglias, line 328: not now but maybe we could change these dates to use the event start and end dates directly? instead of having to declare them for each item

website/server/libs/slack.js
```diff
@@ -188,7 +189,9 @@ function sendSubscriptionNotification ({
   let text;
   const timestamp = new Date();
   if (recipient.id) {
-    text = `${buyer.name} ${buyer.id} ${buyer.email} bought a ${months}-month gift subscription for ${recipient.name} ${recipient.id} ${recipient.email} using ${paymentMethod} on ${timestamp}`;
+    const currentEvent = getCurrentEvent();
+    const promoString = currentEvent && currentEvent.promo && currentEvent.promo ? ' and got a promo' : '';
```
paglias: double check for currentEvent.promo

website/common/script/content/hatching-potions.js
SabreCat: In this specific case, the hatching potions come out on 12/22 which doesn't correspond to an event as such. But some of the other ones I may be able to fix, I'll take a look!

Issue #11199, opened 2019-05-31
Notifications not always received for private messages (PMs)


Alys:
Experienced players who receive a lot of PMs are reporting that they don't always get notifications for them. _Some_ of those case may be caused by https://github.com/HabitRPG/habitica/issues/11172 but there seems to be other cause(s) too. I haven't noticed the bug in my testing but I get far fewer PMs than AnneLine, who has observed this a few times now (see https://github.com/HabitRPG/habitica/issues/11172 for related feedback from her).

From @AnneLine (16a107de-d519-43ce-8c30-8da5cbcbb237) three days ago:

**I'm back with another couple of items of feedback regarding notifications. I'm really starting to think there might be a bug going on. But it's really hard to prove a negative. But for about the 3rd time in as many weeks, I've had to send a message like this to a Badger 101 player who had requested a gem prize via PM: "Wow, I missed a notification, I guess, because I didn't see last week's 'badger gem' request! Well done!!!!"**

**Now that's not a huge thing in the bigger scheme of things, but... within the last 48 hours, both @blakejones99 and I have missed fairly important messages to each other because we had no way of knowing the other had sent one. We're having to resort to 'have you checked messages' notes in the STAG guild... Any chance something really is going on that hasn't been seen? There have been a lot of times in the last few weeks that I thought I was just being kinda absent minded, but now I'm wondering....**

From @AnneLine (16a107de-d519-43ce-8c30-8da5cbcbb237) today:

**I'm just putting in another round of 'I'm definitely not getting notifications of messages - just missed some from both @agateAngel and @Ceran.... fortunately, I found the email telling me about both of them....**

From Ceran, Hamster Juggler / @Ceran (aabe567b-aa55-42c1-b612-b7699a7ba9b7) today:

**Joining the ranks of not-PM-notification-getters... Ironically a PM from @AnneLine. I only noticed it was there because I saw the notification on the app (the one that shows a preview of the message), but the website was missing the little red 1 showing that there's an unread message, even after syncing and refreshing. (no, I didn't open the message on the app before that to test it out)** 
I'm asking Ceran if they opened the Messages index screen even though they didn't open the individual message, as that would be enough to clear the notification, however that would be extra evidence that a usability change is needed as outlined in https://github.com/HabitRPG/habitica/issues/11172

----

With PR #12335 the Habitica codebase has been updated to support MongoDB 4.2 and transactions making it possible to fix this issue. 

[...]

PR #12887, opened 2020-12-16
Count down notification regarding issue-11172

### LINKED ISSUE #11172: Visiting top level private message list should not clear all PM notifications

aevix:
Sorry for taking so long to come up with the solution for this issue, I am still pretty new to coding and it took me a hot minute to figure out how the frontend and backend were talking to each other. I got it working and I am sure there are things I missed but this is what I have so far.

I want to give some of the credit to @ResamVi. I missed that he already submitted a PR few months ago regarding this issue and I wish I saw his PR earlier because that could have saved me a few days of research. His PR had a lot of the things solved already that I took and tweaked to work with the new develop version. 

Before:
When a sender sends a message to the receiver, it creates two messages, one for the sender and one for the receiver. It also increments the `store.state.user.data.inbox.newMessage` for the receiver per message. When user goes to the messages under user dropdown menu, it calls the api v3 mark_pms_read which resets the `inbox.newMessages` counter to 0 and refreshes the frontend.

Fixes #11172 

### Changes
[//]: # Starting from the backend, I added a read property to the inbox schema and set it to false on default. When user sends the message, in the methods.js file I added a read property set as true for the sender's message in the sender's inbox so that it doesn't count the message as unread later in the chain. I also added an if statement that basically resets the counter to 0 if for some reason the counter goes to a negative number. I noticed during testing when counter became negative it was triggering all sorts of weird issues for notification and that might have some relations to issue #11199. 
In the private-messages.vue file, under `loadMessages` method, I added the function that @ResamVi had in his code that looks for any messages that were loaded that had the `read`property as false and saves it in `unreadMessagesCount` object. It emits a function call for `markPrivMessagesRead` which updates the `newMessages` counter on the frontend side. Also api is called using `axios.get` and passes `unreadMessagesCount` and `toUserUuid `to backend in api-v4/inbox.js file. This api takes the request and passes it to `markPmsRead` from inbox/index.js which subtracts the `unreadMessagesCount` from `user.inbox.newMessages` and saves the state and changes the message read property to true. 

----
UUID: 228f739f-b42c-4801-932c-10b4d739b460

### DIFF
--- a/test/api/v3/integration/user/POST-user_mark_pms_read.test.js
+++ b/test/api/v3/integration/user/POST-user_mark_pms_read.test.js
@@ -15,9 +15,9 @@ describe('GET /user/mark-pms-read', () => {
     await user.update({
       'inbox.newMessages': 1,
     });
-    const unreadMessageCount = 1
+    const unreadMessageCount = 1;
     await user.get(`/user/mark-pms-read?count=${unreadMessageCount}&to=${user._id}`);
     await user.sync();
-    expect(user.inbox.newMessages).to.equal(1);
+    expect(user.inbox.newMessages).to.equal(0);
   });
 });
--- a/test/api/v3/integration/user/POST-user_mark_pms_read.test.js
+++ b/test/api/v3/integration/user/POST-user_mark_pms_read.test.js
@@ -1,11 +1,11 @@
 import {
   generateUser,
-} from '../../../../helpers/api-integration/v3';
+} from '../../../../helpers/api-integration/v4';
 
-describe('POST /user/mark-pms-read', () => {
+describe('GET /user/mark-pms-read', () => {
   let user;
 
-  beforeEach(async () => {
+  before(async () => {
     user = await generateUser();
   });
 
@@ -15,8 +15,9 @@ describe('POST /user/mark-pms-read', () => {
     await user.update({
       'inbox.newMessages': 1,
     });
-    await user.post('/user/mark-pms-read');
+    const unreadMessageCount = 1
+    await user.get(`/user/mark-pms-read?count=${unreadMessageCount}&to=${user._id}`);
     await user.sync();
-    expect(user.inbox.newMessages).to.equal(0);
+    expect(user.inbox.newMessages).to.equal(1);
   });
 });
--- a/website/server/models/message.js
+++ b/website/server/models/message.js
@@ -45,6 +45,7 @@ const inboxSchema = new mongoose.Schema({
   // we store two copies of each inbox messages:
   // one for the sender and one for the receiver
   ownerId: { $type: String, ref: 'User' },
+  read: { $type: Boolean, default: false },
   ...defaultSchema(),
 }, {
   minimize: false, // Allow for empty flags to be saved
--- a/website/client/src/pages/private-messages.vue
+++ b/website/client/src/pages/private-messages.vue
@@ -830,8 +830,6 @@ export default {
 
       await this.loadConversations();
 
-      await this.$store.dispatch('user:markPrivMessagesRead');
-
       this.loaded = true;
     },
     async loadConversations () {
@@ -964,6 +962,11 @@ export default {
       const requestUrl = `/api/v4/inbox/paged-messages?conversation=${conversationKey}&page=${this.selectedConversation.page}`;
       const res = await axios.get(requestUrl);
       const loadedMessages = res.data.data;
+
+      const unreadMessageCount = loadedMessages.filter(msg => !msg.read).length;
+      const toUserUuid = this.selectedConversation.key;
+      await this.$store.dispatch('user:markPrivMessagesRead', unreadMessageCount);
+      await axios.get(`/api/v4/user/mark-pms-read?count=${unreadMessageCount}&to=${toUserUuid}`);
 
       /* eslint-disable max-len */
       this.messagesByConversation[conversationKey] = this.messagesByConversation[conversationKey] || [];
--- a/website/client/src/store/actions/user.js
+++ b/website/client/src/store/actions/user.js
@@ -175,6 +175,7 @@ export function unblock (store, params) {
 
 export function markPrivMessagesRead (store, ct) {
   markPMSRead(store.state.user.data, ct);
+  return axios.post('/api/v4/user/mark-pms-read');
 }
 
 export function newPrivateMessageTo (store, params) {
--- a/website/client/src/store/actions/user.js
+++ b/website/client/src/store/actions/user.js
@@ -173,9 +173,8 @@ export function unblock (store, params) {
   return axios.post(`/api/v4/user/block/${params.uuid}`);
 }
 
-export function markPrivMessagesRead (store) {
-  markPMSRead(store.state.user.data);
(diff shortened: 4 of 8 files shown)


aevix:
I changed mark-pms-read test so it passes but I am not sure why its failing on the send-private-message test. I don't think the changes I made would have affected any of the api for sending messages.

test/api/v3/integration/user/POST-user_mark_pms_read.test.js
```diff
@@ -1,11 +1,11 @@
 import {
   generateUser,
-} from '../../../../helpers/api-integration/v3';
+} from '../../../../helpers/api-integration/v4';
 
-describe('POST /user/mark-pms-read', () => {
+describe('GET /user/mark-pms-read', () => {
```
paglias, line 5: Looks like this is a breaking change which we can't have. Any existing route should keep working as it's now, basically you can add something new but not break an existing functionality

test/api/v3/integration/user/POST-user_mark_pms_read.test.js
```diff
@@ -1,11 +1,11 @@
...
   let user;
 
-  beforeEach(async () => {
+  before(async () => {
```
paglias, line 8: why was this changed?

paglias:
Hi @aevix , I'm very sorry for the delay. I think this approach might be a bit too complex as it requires changes to the existing clients (web & mobile) and also changes the existing data structure for PMs in the database. I think something simpler would be to add a new route (or a parameter to the existing one) that when called reduces the number of unread pms by 1

aevix:
@paglias I will try and come up with a simpler solution and let you know.

aevix:
@paglias I've been thinking about a way to just use the api already set up to simply subtract the notification by 1 every time the user opens a new message but the problem is that since the notification isn't set up to track individual messages, it just stores the static number of new messages received in the receiver's inbox when sender sends their message, so if user opens a specific new message two or three times it will continue to subtract from the newMessage counter even though there are still messages that were not read. Hopefully I am making sense.

I don't see any work around this issue unless we add something to the inbox schema to keep track of every message whether it is read or not and show notification for only messages that are marked not read. There might be a easier method that I am not seeing but this seems to be the only solution I can come up with. I would like to hear your input to maybe clue me into something I am not seeing.

Thanks!

website/server/models/message.js
```diff
@@ -45,6 +45,7 @@ const inboxSchema = new mongoose.Schema({
   // we store two copies of each inbox messages:
   // one for the sender and one for the receiver
   ownerId: { $type: String, ref: 'User' },
+  read: { $type: Boolean, default: false },
```
paglias, line 48: My main concern with this one is that we'll have to run an huge migration to mark as read all existing messages and we actually won't know if they've been read or not

paglias:
@aevix 

> @paglias I've been thinking about a way to just use the api already set up to simply subtract the notification by 1 every time the user opens a new message but the problem is that since the notification isn't set up to track individual messages, it just stores the static number of new messages received in the receiver's inbox when sender sends their message, so if user opens a specific new message two or three times it will continue to subtract from the newMessage counter even though there are still messages that were not read. Hopefully I am making sense.

That's a good point, hmm.. I left a few comments but I don't think there's an easy solution and changing the schema, apart from being a little complex right now, has the problem that we don't know which PMs have been read at the moment so we won't know if they have to be marked as read or unread in the database

aevix:
> 

> @aevix
> 
> > @paglias I've been thinking about a way to just use the api already set up to simply subtract the notification by 1 every time the user opens a new message but the problem is that since the notification isn't set up to track individual messages, it just stores the static number of new messages received in the receiver's inbox when sender sends their message, so if user opens a specific new message two or three times it will continue to subtract from the newMessage counter even though there are still messages that were not read. Hopefully I am making sense.
> 
> That's a good point, hmm.. I left a few comments but I don't think there's an easy solution and changing the schema, apart from being a little complex right now, has the problem that we don't know which PMs have been read at the moment so we won't know if they have to be marked as read or unread in the database

since we don't have any info on whether message is read or not now, couldn't we just make default as false and then create a button that marks all messages as read? Although with this solution I am not sure if that would cause any issues on the backend server side with potentially thousands of requests at once changing each message as read but that could be one solution. 

Another solution i thought was do two updates where initially mark every message as read on migration and then do a second update to the schema to make default to false so any new messages that are created after the second update will be correctly marked. I am not sure if either of these would be a good solution but it would be interesting to hear your thoughts.

test/api/v3/integration/user/POST-user_mark_pms_read.test.js
aevix: Okay, I will have to go back and create a separate test for the new api and revert all the changes. Ill update it once I am done and get it working.

test/api/v3/integration/user/POST-user_mark_pms_read.test.js
aevix: I don't think that should have been changed, it might have been a typo on my part. Ill go ahead and fix it.

paglias:
> Another solution i thought was do two updates where initially mark every message as read on migration and then do a second update to the schema to make default to false so any new messages that are created after the second update will be correctly marked. I am not sure if either of these would be a good solution but it would be interesting to hear your thoughts.

But you can't know which of the existing messages have been read or if they're still to be read

aevix:
@paglias I don't know if there is a way to figure out if a message is read since current inbox schema doesn't track that information. We could maybe add a function for users to mark a message as unread if they want to go back manually.

SabreCat:
Closing per author's comment here: https://github.com/HabitRPG/habitica/issues/11172#issuecomment-784700783

This may need some discussion between UX and technical devs, ultimately, before we open it back up for contributors. Thank you for the effort diving into this, though, @aevix!

PR #12915, opened 2021-01-01
Hourglass items should be unpinned after purchase

### LINKED ISSUE #12905: Hourglass-purchasable Equipment should be unpinned when purchased
As reported by @Nakonana and confirmed by me:

"When buying a pinned equipment set (Time Travellers) with an hourglass, the set is not automatically unpinned after purchase, even after refreshing the site."

The desired behaviour is that as soon as you've purchased a Mystery Item set with an Hourglass, if it was previously pinned to your Rewards column, it should immediately become unpinned.

The bug occurs regardless of whether you buy the item from Rewards or from the Time Travellers Shop.

_(The bug does not occur with Equipment that is purchasable with Gold or Gems, nor with Backgrounds purchasable with Hourglasses, nor with Quests purchased with Hourglasses - they're meant to stay pinned since you may wish to buy another later.)_

samamill:
Fixes #12905

### Changes
- Pets, mounts, and mystery sets now are unpinned when purchased. 
- Backgrounds are already unpinned when purchased.  No change needed.  
- Quests should remain pinned since a user can purchase more than one.  No change needed.
- Added tests for pets, mounts, and background.  Added a check to the existing mystery set test.

UUID:  bb6ce8ce-17d5-49de-a4fc-b3486714266f

Note that the cron tests are failing for me locally, both in my branch and on master.  But the new tests I added are passing.  I read that tests are automatically run when I submit a PR.  Let me know if you want me to do any additional validation.


Alys:
> Note that the cron tests are failing for me locally, both in my branch and on master

That can happen with cron tests around the change of month due to things like time zone issues. The tests are passing now.

paglias:
@negue could you give a look to this one?

test/common/ops/buy/buy.js
```diff
@@ -197,4 +201,62 @@ describe('shared.ops.buy', () => {
...
+      },
+      quantity: 1,
+    },
+    analytics,
+    {
+      hourglass: true,
+    });
+
+    expect(user.purchased.background[key]).to.be.true;
+    expect(pinnedGearUtils.removeItemByPath.calledOnce).to.equal(true);
```
negue, line 260: I think it would be the better - to let the real function be used and then check on the pinnedItems after that

website/common/script/ops/buy/buyMount.js
```diff
@@ -41,6 +42,7 @@ export class BuyHourglassMountOperation extends AbstractHourglassItemOperation {
     if (user.markModified) user.markModified('items.mounts');
 
     this.subtractCurrency(user);
+    removeItemByPath(user, `timeTravelStable.mounts.${this.key}`);
```
negue: at first I thought we'd need to use `getItemInfo` which would have the path already, but on the other hand we'd create this object just to get the path out of it

@paglias what do you think ?

negue (commented):
I think the code itself LGTM

Just two ideas we might need to discuss first with @paglias:
- use `getItemInfo` so that we always get the correct path (in case of a change or something)
- to use the correct `removeItemByPath` method itself

test/common/ops/buy/buy.js
paglias: agreed!

website/common/script/ops/buy/buyMount.js
paglias: I think that using `getItemInfo` would be more future proof in case the path changes

paglias:
@samamill there are a couple of comments to address but then it should be ready to merge

test/common/ops/buy/buy.js
samamill: Sorry, I don't quite understand.  I am calling "buy".  What do you mean by "the real function"?

samamill:
@paglias Can you take another look?  I addressed the comment in the main code.  I didn't quite understand the comment in the test code.

paglias:
Sorry Sam but for the time being I won't be able to review PRs, @Sabrecat
should be able to help you though

Il lun 25 gen 2021, 01:28 Sam <notifications@github.com> ha scritto:

test/common/ops/buy/buy.js
SabreCat: I believe what's meant is, rather than spying on the `removeItemByPath` function and expecting it to be called once, we should put the item into the test user's `pinnedItems` array, run the `buy` function, and then `expect` the pinned item to be missing at the end.

SabreCat (changes requested):
This doesn't appear to be working, on a test of the functionality.

1. Pin a quest, a background, a pet, a mount, and a Mystery Set in the Time Travelers Shop
2. Purchase the quest from your pinned items--it stays--OK
3. Purchase the background from your pinned items--it _looks_ like it goes away, but...
4. Purchase the pet from your pinned items--it's still there
5. Sync--now the background shows up again, but the pet goes away

So it seems like we need to update the client data when these items are removed from the pinned array, and revisit the logic for removing the background.

SabreCat:
@samamill It's been about a month--will you have time soon to revisit this, or shall we open it back up for someone else to have a look?

SabreCat:
Closing as abandoned and will mark the original issue as available for help again.

PR #12921, opened 2021-01-02
Fix shop item popover staying open after click

### LINKED ISSUE #12884: Shop on-hover popup should always disappear when moving mouse off item
As reported by @johnjsal (624e48a3-cbea-4bbe-ae44-ca9183f49626):

**"when I go to Shops > Quests and click on a quest, the quest description pops up like normal, but when I close this pop-up, the smaller quest text window (the one that appears when you hover over a quest) remains locked on the screen. It won't disappear if you move the cursor around, but it will when you click elsewhere on the screen."**

The same applies to Shops > Market and Shops > Time Travellers and probably Shops > Seasonal Shop. The same fix should be applied to all of them (it may be shared code so it's possible that a fix in one place may fix all of them).

z0glen:

Fixes #12884 

### Changes

These changes will prevent the tooltip from appearing after clicking on a shop item and closing the modal. 

The idea here is that the bootstrap modal component allows for an additional argument to choose where to send focus on close. In this case, when the modal is triggered by a click, focus will go to the root of the app, instead of back to the shop item, which will avoid the tooltip returning. The side effect is that a user can't click a shop item and then keyboard navigate around from there, so that sort of hybrid navigation would be restricted.

I limited these changes to the shop, as in the original reported bug, but it seems like this issue would be repeated anywhere there is a focus popover and click event on the same component. I'm happy to expand this out, but these changes were getting into a number of places already and I didn't want to go too crazy right off the bat.

----
UUID: 49ac1d9d-bdaa-4a89-9ad2-790e396b8952


SabreCat:
Interesting--I definitely find the fix to be an improvement over the previous annoying behavior, but I'm concerned that the disruption to keyboard navigation may be too steep a price to pay. What do you think, @Tressley ?

Tressley:
While I agree that certain parts of the behavior feel better, I do not want to lose the keyboard navigation accessibility in return.

SabreCat:
OK, so it sounds like the solution in this PR won't work out. Maybe instead of forcing the focus elsewhere upon closing the modal, we could force the popover closed instead?

z0glen:
Yeah that makes sense, I feel like there was a reason I didn't start out with that but I can't quite remember now, I can try it out!

SabreCat:
@z0glen Did you have any luck finding a solution that preserves keyboard focus? If you're stumped, we can open this back up for someone else to have a look!

z0glen:
Hey @SabreCat I just pushed up a new version that I think is a bit cleaner - it's just disabling the popover once the modal is closed, and re-enabling when the button loses focus.

SabreCat:
All right, this is working great now! Thanks, @z0glen ! Welcome to the ranks of the Blacksmiths at tier 1. Further tiers require multiple or larger PRs to attain, but keep helping out and we'll express our gratitude accordingly!

PR #12932, opened 2021-01-04
Hotfix: moving pinned items

negue:
This fixes / increases / refactors the checks on moving items

PR #12935, opened 2021-01-06
Hotfix: moving pinned items 

negue:
This fixes / increases / refactors the checks on moving items

2nd try #12932

### DIFF
--- a/website/server/controllers/api-v3/user.js
+++ b/website/server/controllers/api-v3/user.js
@@ -1716,7 +1716,7 @@ api.movePinnedItem = {
 
     const { user } = res.locals;
     const { path } = req.params;
-    const position = Number(req.params.position);
+    let position = Number(req.params.position);
 
     // If something has been added or removed from the inAppRewards, we need
     // to reset pinnedItemsOrder to have the correct length. Since inAppRewards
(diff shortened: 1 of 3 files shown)


paglias:
LGTM!

website/server/controllers/api-v3/user.js
```diff
@@ -1727,17 +1727,25 @@ api.movePinnedItem = {
       user.pinnedItemsOrder = currentPinnedItems.map(item => item.path);
     }
 
+    const itemExistInPinnedArray = user.pinnedItems.findIndex(item => item.path === path);
+
+    if (itemExistInPinnedArray === -1) {
+      throw new BadRequest(res.t('wrongItemPath', { path }, req.language));
```
Alys, line 1736: Will this cause a bug when a player tries to move one of the automatically pinned items that isn't in their `pinnedItems` array?

paglias: Oh I thought this had been merged but given this doubt let's wait for @negue

negue: good point! gonna rework this a bit and try to add test for that as well - cause yes at that point that item isn't in user list (atleast not in the database array) the result however has those items .... so lets see how do it properly ^^

negue: the new push should allow to move official items too

negue: or not... `import getOfficialPinnedItems from '../../../common/script/libs/getOfficialPinnedItems';` can't be used in the user controller ..? any workaround to get this there too ?

Edit: using only the `common` import it worked out :)

SabreCat:
Looks good on my run through!

PR #12962, opened 2021-01-11
9845 - Extract search bar and tag filter popup into separate components

### LINKED ISSUE #9845: Tags filter button colour change is not obvious
In the Trello [ACCESSIBILITY SETTINGS](https://trello.com/c/pxTcbKuD/381-accessibility-settings) card, S. L. Casteel wrote:
"The Tags dropdown button on the Tasks page has very little contrast between the gray when no tags are selected and the purple when tags are selected. I only realized today that the color changes, and even now that I know it, I have trouble seeing which color it is. It'd be helpful to adjust one of the colors to increase the contrast."

I also hadn't noticed that the colour changed until I read this, although that might be partly because I rarely use tags anymore now that it's two clicks and scrolling to select them instead of one click. I think though the colour change would benefit from being more obvious.

benkelaar:
Fixes #9845

### Changes
- Add "Clear tags" link to search bar
- Extract search bar, tag filter popup & tag filter button into separate components
- Refactor Tag filter panel to have a single set of components.
- Remodel Tag filter panel according to new design: https://app.zeplin.io/project/59c16a1433075573feaa48fb/screen/5ffc6fcaf519de768e71b9b1

Pretty happy about the result, ~but didn't get round to building unit tests~.

----


benkelaar:
I think this one is now ready to be reviewed+merged.

benkelaar:
@SabreCat @paglias I think this one has been done for a while, is there anything I can do to help get it merged?  I guess it would first need to go to Delta?

SabreCat:
This is working well, and should be a good refactor! I'm noticing just one small oddment--the width, or possibly position, of the search bar seems to no longer be correct. On production:
image

With this PR:
image

The wider search field leads the overall block of search bar + tag menu to be off-center with respect to the rest of the view. @Tressley do you know off hand which width is correct? If the smaller width is right, we should correct the width; if the larger, then I suspect we need to move the elements for better centering.

benkelaar:
Hi @SabreCat !  Awesome, thanks!

Good feedback too, I think I know how to fix it, but won't have time to look in to it until at least next weekend.

Tressley:
The width itself can be fluid based on the user's window size, but the search input + tag dropdown should be centered as one grouped object.

SabreCat:
@benkelaar It's been about a month, here--will you be circling back to this, or should we open it up for someone else to take across the finish line?

SabreCat:
Closing as abandoned and will mark the original ticket as needing help again.

PR #12967, opened 2021-01-15
Fix error logs in client unit tests from setTitle

benkelaar:
While fixing #9845 I noticed new unnecessary error logs had been introduced in the website unit tests with the introduction of the common:setTitle function.  This merge request fixes those.  There's also a log statement that has been showing up in merge requests, I removed it, but if it's necessary I can also add `// eslint-disable-line no-console` to it instead.

I also noticed there were two duplicate instances of `home.spec.js`.  I think I created that at some point so I removed the one that seemed least up to date.

### Changes
- Add `common:setTitle` stubs for unit tests that need them.
- Remove duplicate home.spec.js
- Remove console log statement

----


benkelaar:
@paglias Small merge request to speed up the client unit tests.  Thought it was too disjoint from #12962 to do it in the same merge request.

SabreCat:
Thanks for doing this cleanup, @benkelaar! Noted toward your next Blacksmith tier.

PR #12968, opened 2021-01-15
12922 - Use absolute user URLs on production

### LINKED ISSUE #12922: @Username automatic profile URLs in chat messages should be absolute not relative
As reported by @citrusella in her post copied below, when you type an @Username in chat, the generated profile link is relative, not absolute.

This means that the link works when you're viewing the post in the habitica.com website but doesn't work in the Android app or in third-party tools. I'm not aware of whether the links work in the iOS app but I suspect they wouldn't.

I'm logging this here rather than in the Android app's repo because the link is created by the API and any fix would need to be done in the API (unless a decision is made to implement a fix separately in each of the mobile apps and for all third-party tool authors to implement their own fix).

_I'm marking this as suggestion-discussion for Habitica's staff to make a decision about how to fix this._

Note that if it is fixed in the API, I believe  it would be important for the `habitica.com` domain name to be hard-coded, rather than being read from the `BASE_URL` value in the server's config. If `BASE_URL` was used, then when we made posts in the staging or online test sites, the profile URLs in those posts would permanently link to the staging/test sites, making them unusuable by most players.
However for testing on local installs, we WOULD want `BASE_URL` to be used, so that profile links worked correctly on the local site. I guess the value of `NODE_ENV` could help with this - if it's set to `development` then `BASE_URL` should be used to create the absolute profile links; if it's set to `production`, then `habitica.com` should be used.
Note though that if Habitica's staff say something different than this, listen to them not me. :)

----

Here is citrusella's post from the Report a Bug guild:

> So... in some ways this is an app bug because it does not happen on the website (due to how browsers treat certain kinds of links) but I thought it over and decided to report it here (or at least *start* here) because the problem users are most likely to encounter has to do with profile links in mentions (which are generated by the... server? API?) and it could be causing problems with every third party tool (unless there's a way to take it into account) and/or the apps?
> 
> When you type an @ mention (i.e. `@citrusella`), the hyperlink that's generated goes to `/profile/2d6ef231-50b4-4a22-90e7-45eb97147a2c` (in my case, since I'm using me as an example). That works completely fine on the website, because it's recognized as a link to a page within the same website.
> 
> The Android app does not understand this link, at all. It attempts to open `/profile/2d6ef231-50b4-4a22-90e7-45eb97147a2c` in a browser and doesn't even give the choice to open it in Habitica. This results in an error page because `/profile/2d6ef231-50b4-4a22-90e7-45eb97147a2c` on its own is not a real URL. (By comparison, current versions of the Android app can open <https://habitica.com/profile/2d6ef231-50b4-4a22-90e7-45eb97147a2c> just fine, giving you the option to open in a browser or Habitica and functioning as intended if opened on the app. I have no idea how iOS treats either of these situations.)
> 
> Now, I suppose it *could* be possible to just fix this on the app somehow (make it understand /blahblahblah links as internal?), but as part of investigating how this could be affecting other tools, I used the Party and Guild Data Tool to check mention URLs on the chat history that tool offers, and it experiences issues because it considers the pages to be internal to oldgods.net (i.e. <https://oldgods.net/profile/2d6ef231-50b4-4a22-90e7-45eb97147a2c>, which is not a URL that exists. I suppose any tool that displays chat posts like that could have the issue.
> 
> Granted, *any* URL written this way (i.e. [/groups/tavern](/groups/tavern]) will cause problems on external tools like this, but given mention links are perhaps the most common instance of this, I figured it was worth some sort of report, i.e. to fix @ mention urls specifically, or find some way that the intended full URLs are given via the API somehow when partial links like that are encountered (spitballing here, I don't actually know what's possible)... ~~or to get told to report to mobile@habitica.com XD~~

benkelaar:
Fixes #12922

### Changes
- Add `habitica.com` root url to injected user links

----

### DIFF
--- a/test/api/unit/libs/highlightMentions.test.js
+++ b/test/api/unit/libs/highlightMentions.test.js
@@ -11,14 +11,14 @@ describe('highlightMentions', () => {
       lean () {
         return this;
       },
-      exec () {
-        return Promise.resolve([
+      async exec () {
+        return [
           { auth: { local: { username: 'user' } }, _id: '111' },
           { auth: { local: { username: 'user2' } }, _id: '222' },
           { auth: { local: { username: 'user3' } }, _id: '333' },
           { auth: { local: { username: 'user-dash' } }, _id: '444' },
           { auth: { local: { username: 'user_underscore' } }, _id: '555' },
-        ]);
+        ];
       },
     };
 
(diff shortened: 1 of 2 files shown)


SabreCat (approved):

SabreCat:
Great! Another one toward Tier 6 (getting close there I should say), thank you!

PR #12969, opened 2021-01-15
12893 - Keep other chat messages right-aligned when avatar doesn't render

### LINKED ISSUE #12893: chat posts by others look like yours when avatar images don't load
As reported by @SuperSaraA (dbf5390d-6a57-42ad-ae14-af8fd8e570ca) and other players in the past:

"... the random display bug that shows other users posts as if they were me is acting up again. Both in the Tavern chat and also the message before mine here in the Report a bug guild."

She took screenshots and I've copied some representative samples of the bug below. The top post on the first screenshot was written by 
SuperSaraA. All others were written by other players.

It looks like the issue is that sometimes the players' avatar images don't load, and that means the posts aren't pushed over to the right the way that they would be if the avatars were displaying. This makes it look like the posts were written by you, since we're all used to seeing our own posts aligned to the left.

Ideally, any post by a player other than yourself should appear aligned to the right even if their avatar isn't visible.

Note that I would guess it's possible that sometimes the player's own avatar may not load. In that case, the player's own posts should still be aligned to the left. Whatever fix is done for this issue should not cause the player's own posts to bbe right-aligned when their own avatar isn't shown.

SuperSaraA reported that the problem disappeared later, so I would assume that reloading would fix it every time. 

**See the comments below from @veeeeeee and @paglias for the required fix, including the second comment by @paglias after @citrusella noted that the div being missing is the issue.**
**As noted in the post from @citrusella the fix should NOT affect system posts (the ones that have** `"uuid": "system"`**). They deliberately don't have avatars on either side and should stretch all the way across the chat space.**

image

image

benkelaar:
…nder

Fixes #12893

### Changes
- Moved check if no user available from `v-if` to `invisible` class.  That way spacing is retained when it's not available.
- Added explicit system message check to not show any avatar at all.
- Reduced some duplication in the HTML code.

----

### DIFF
--- a/website/client/src/components/chat/chatMessages.vue
+++ b/website/client/src/components/chat/chatMessages.vue
@@ -30,47 +30,22 @@
         {{ $t('loading') }}
       </h2>
     </div>
-    <!-- eslint-disable vue/no-use-v-if-with-v-for -->
     <div
-      v-for="msg in messages"
-      v-if="chat && canViewFlag(msg)"
+      v-for="msg in messages.filter(m => chat && canViewFlag(m))"
       :key="msg.id"
     >
-      <!-- eslint-enable vue/no-use-v-if-with-v-for -->
-      <div
-        v-if="user._id !== msg.uuid"
-        class="d-flex"
-      >
+      <div class="d-flex">
         <avatar
-          v-if="msg.userStyles
-            || (cachedProfileData[msg.uuid] && !cachedProfileData[msg.uuid].rejected)"
+          v-if="user._id !== msg.uuid && msg.uuid !== 'system'"
           class="avatar-left"
+          :class="{ invisible: avatarUnavailable(msg) }"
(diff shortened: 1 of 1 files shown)


Alys:
I'm having trouble replicating the bug on my local install so I'm not completely certain this fixes it but I have confirmed locally that this PR still displays messages correctly when the bug does not exist for these cases:
- system message
- message by the user viewing chat
- message by any other user
So at the very least this isn't any worse than the current situation!

Also, the code itself definitely looks better. I love that some duplication has been removed and that it's no longer necessary to use `eslint-disable vue/no-use-v-if-with-v-for`

@benkelaar Thank you! I've noted this in your Tier 6 progress.

benkelaar:
I had the same thing.  I mostly "reproduced" it by adding `&& false` to the relevant conditionals.

However since those conditionals now only manage visibility (which doesn't affect spacing, opposed to display, which does) it should fallback gracefully.

PR #12986, opened 2021-01-19
Use mongoose session to ensure user and group are updated in same transaction

### LINKED ISSUE #12124: guild creation failing, gems being taken
Recently we've seen two reports of people trying to make a guild, having their Gems removed, but the guild not existing.

This one from @Muamua_ (1039bae3-1101-4c9f-8ad1-416b04ecd93d) was 15 hours before now:
_"Then I logged in on website version and create guild but it occur error. After that, my gems disappeared but my guild didn't created.."_

This one from @tymeczekxpompa (1e3a7b23-5fde-44d5-b742-12e848f93f4d) was from 10 days ago:
_"When I was creating the private guild and clicked "create a guild" button an error happened which took my 4 gems but hasn't created any guild."_

In both cases I found a group ID in their `guilds` array that did not correspond to a guild that currently exists (ca200a70-1f0f-48f3-8423-4ed02c8b05bb and 2d69ff05-e763-488a-9706-b3a0f9bed748 respectively). Those IDs are still there but we should remove them eventually.

The Amplitude/Loggly event type for creating a guild is `json.event_type:"join group"` ([code ref.](https://github.com/HabitRPG/habitica/blob/d03e5e93b09f5a16598ed63b911c074d28c62368/website/server/controllers/api-v3/groups.js#L167)), i.e., the same as for joining a guild, however it would have `json.event_properties.owner:"true"` (true rather than false). I haven't found such a record in Loggly for @Muamua_, except for one from a few minutes ago (2020-04-26 04:00:12.240) which was after I gave them Gems and told them to true again. Either their failed attempt happened before that data started to go into Loggly, or (more likely) the guild creation failed before it got to the point where analytics were recorded.

I haven't found Loggly errors associated with this but I have **not** searched thoroughly. We probably should, I'm just a bit short of time today.

I've given 4 Gems to both of those people.

----

With PR #12335 the Habitica codebase has been updated to support MongoDB 4.2 and transactions making it possible to fix this issue. 

Transactions allow operations on multiple documents to be executed ensuring that either all of them are executed correctly or none, making it possible to fix this issue.

An example on using transactions can be found at https://mongoosejs.com/docs/transactions.html while more info at https://docs.mongodb.com/manual/core/transactions/. If you want to work on this issue and have any question please leave a comment!

samamill:
Fixes #12124

### Changes
Use mongoose session to ensure user and group are updated in same transaction

UUID: bb6ce8ce-17d5-49de-a4fc-b3486714266f

This fix seems like it would be best tested with a unit test.  However, I haven't been able to find the server tests in the code.   Can you point me to the right directory?

### DIFF
--- a/website/server/controllers/api-v3/groups.js
+++ b/website/server/controllers/api-v3/groups.js
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import nconf from 'nconf';
+import mongoose from 'mongoose';
 import { authWithHeaders } from '../../middlewares/auth';
 import {
   model as Group,
@@ -137,7 +138,13 @@ api.createGroup = {
       user.party._id = group._id;
     }
 
-    const results = await Promise.all([user.save(), group.save()]);
+    const session = await mongoose.startSession();
+    let results;
+    await session.withTransaction(async () => {
+      results = await Promise.all([user.save(), group.save()]);
+    });
+    session.endSession();
+
     const savedGroup = results[1];
 
     // Instead of populate we make a find call manually because of https://github.com/Automattic/mongoose/issues/3833


website/server/controllers/api-v3/groups.js
```diff
@@ -137,7 +138,13 @@ api.createGroup = {
       user.party._id = group._id;
     }
 
-    const results = await Promise.all([user.save(), group.save()]);
+    const session = await mongoose.startSession();
+    let results;
+    await session.withTransaction(async () => {
+      results = await Promise.all([user.save(), group.save()]);
+    });
+    session.endSession();
```
benkelaar, line 146: Don't you need to explicitly commit the transaction?

```javascript
await session.commitTransaction();
```

samamill: Not according to the documentation.  https://mongoosejs.com/docs/transactions.html

lmk if I am misunderstanding something.



> The previous examples explicitly create a transaction and commits it. In practice, you'll want to use the session.withTransaction() helper instead. The session.withTransaction() helper handles:
> 
> Creating a transaction
> Committing the transaction if it succeeds
> Aborting the transaction if your operation throws
> Retrying in the event of a transient transaction error.

benkelaar: Ah, indeed, you're right, I had read it as an invocation of `startTransaction` instead.

It's a pity that withTransaction doesn't return the underlying value on successful commit, that could make this code a bit nicer.

Anyway, LGTM.

SabreCat:
>This fix seems like it would be best tested with a unit test. However, I haven't been able to find the server tests in the code. Can you point me to the right directory?

You want `test/api`, probably `/v3/integration`! https://github.com/samamill/habitica/tree/samamill_transactions/test/api

SabreCat:
Closing this as a stale PR. If you come back to this, @samamill, we can reopen your request!

PR #12987, opened 2021-01-20
#12781 Add a unified component for small modals

### LINKED ISSUE #12781: Shared style for small modals
All the new small modals (level up, drop cap reached, payments success info, won challenge, ...) share some basic style elements:

- Title color and size
- Paddings
- In some cases the footer
- Same width

We should abstract the styles in a shared component or css class so they're not repeated every time.

Some examples:

Schermata 2020-11-11 alle 16 00 13

Schermata 2020-11-11 alle 16 00 52

Schermata 2020-11-11 alle 15 59 57

benkelaar:
Fixes #12781 

### Changes
- Introduce a small-modal component that takes care of unified elements.
- Refactor levelUp modal to use this component
- Refactor wonChallenge modal to use this component
- Refactor dropCapReached modal to use this component
- Refactor successModal modal to use the css mixin and reduce duplication

I think it works quite well now, only open question is whether there are other modals I should look at and unit tests and perhaps storybook integration?

----


SabreCat:
Looking good so far, thanks for this! One other modal that comes to mind is the confirmation of payment success, https://github.com/HabitRPG/habitica/blob/develop/website/client/src/components/payments/successModal.vue:

image

benkelaar:
@SabreCat Yeah, I had found that one too.  It made me realise that I put too much knowledge in the small-modal component, so I refactored the sparkled and reward parts to their own components.  That made room for properly including this one.  I'll look in to doing that tonight.  I'll put the merge request to WIP.

benkelaar:
@SabreCat Funnily enough the successModal is a bootstrap modal with `size="md"` (Not set so revert to default "md").  However the size is then pushed to 20.5rem.

Going off of this I searched around, there's a couple of ways that different modals get pushed to almost similar sizes, here are the results.

Modals found for: `size="sm"`:
- onboardingComplete.vue (also `330px`)
- leaveChallengeModal.vue
- canceledModal.vue
- canceledModalConfirm.vue
- brokenTaskModal.vue
- taskModal.vue (But this resets the width to `448px`)

Modals found for: `width: 330px`
- genericAchievement.vue
- groupPlan.vue
- equipGearModal.vue
- hatchedPetDialog.vue
- mountRaisedModal.vue
- buyModal.vue
- sellModal.vue
- private-messages.vue

Modals found for: `width: 21rem`
- firstDrops

So the current setup (of having a custom small-modal component) makes it really easy to one-by-one opt modals in to this standard set of styles.  However, the small-modal component as is is quite a dumb component and it's a little risky since it's just a thin wrapper around the bootstrap modal.  This means that whenever a new implementation needs access to properties that are currently not forwarded of the bootstrap modal, the small modal would have to be extended.

What might be a better approach is to have these styles applied globally on the `.modal-sm` class selector and have the unification automatically applied to all small bootstrap modals.  I think that would be a more scalable approach for the future, but would mean going through all these modals to update them (and hopefully remove duplication from them).

What we could do is that I continue like this for the successModal, we merge & test these four and then look at the next step in a follow-up issue, would that be a good idea?

benkelaar:
@SabreCat @paglias So the success modal wasn't as good a fit as the other modals, so what I've done now is also made the scss available as a mixin.  I also refactored the successModal to reduce the duplication in its templating.

I think this merge request is ready to be reviewed now.

SabreCat (approved):

SabreCat:
All right, looking good! We'll let this run in the staging environment for a few days, but if nothing crops up we should get this into production next week. Credited toward your next contributor tier!

SabreCat:
We're seeing some style issues in staging that resolve when I revert this PR, so I'm moving this to "on hold" while we investigate.

benkelaar:
Cool, sounds good, what kind of style issues are we talking about?

PR #13044, opened 2021-02-10
WIP: Open Datepicker depending on the available space

negue:
The logic in the datepicker repo actually works for "normal" pages - so if the datepicker is added in a normal view.

But in our modals - the document itself doesn't have a scrollTop which probably result in a wrong calculation and then end up showing the datepicker always below.

No idea yet how to get this working other than searching for a datepicker based on bootstrap itself 😅


SabreCat:
This PR is real old--are you still planning to tackle this, @negue, or shall we clean it out of the queue?

SabreCat:
With no response in a week, I'm going ahead and closing this. Feel free to pop it back open if you resume work on it!

PR #13108, opened 2021-03-12
Fixes #12966 remove create challenge button for non-leaders if restricted

### LINKED ISSUE #12966: "Create Challenge" button should not appear in groups where you can't create one
When a Guild or Party has a ticked checkbox for "Only the Leader can create Challenges", currently all players see the "Create Challenge" button on it.

The desired behaviour is that only the leader sees that button.

I'm proposing this change because currently if a player clicks the button, they can get all the way through entering the challenge title, summary, and description without any warning that they are wasting their time. It's only when they click the "Add Challenge Tasks" button that they're told they can't create a challenge in that guild. If they've typed a lot of detail, that would be very disheartening.

~~I'm marking this as suggestion-discussion for the staff to decide if the proposed change is appropriate but if there's no objections in a couple of weeks I'll mark this as help wanted.~~

SpeedyLom:
Fixes #12966

### Changes
- Pass the `group` to `groupChallenges` and don't display the "**Create Challenge**" button if it would fail from the modal

can_create_challenge
leader_only_not_a_leader

----


SabreCat:
Looks like this picked up a ton of unrelated changes. I've cherry-picked the relevant commit onto a new branch so this PR can continue in #13221.

PR #13112, opened 2021-03-14
Update Quest Dialogs

negue:
Based on #12793 branch


website/client/config/storybook/mock.data.js
```diff
@@ -72,4 +72,11 @@ export const userStyles = {
     maxHealth: 50,
     maxMP: 158,
   },
+  profile: {
+    name: 'user',
+  },
+  _id: '1337',
```
SabreCat: Cute, but shouldn't this be a UUID? (maybe it doesn't matter outside the API...)

website/client/src/components/groups/invitationListModal.vue
```diff
@@ -0,0 +1,271 @@
...
+    <div
+      v-for="member in members"
+      :key="member._id"
+      class="member-row"
+    >
+      <div class="class-icon">
+        <class-badge
+          v-if="member.stats"
+          :member-class="member.stats.class"
+          style="--badge-size: 40px"
```
SabreCat: Given this is a component, we should be able to pass a prop rather than use an inline style

website/client/src/components/groups/invitationListModal.vue
```diff
@@ -0,0 +1,271 @@
...
+      </div>
+      <div :class="{
+        'status': true,
+        'accepted': member.accepted === true,
+        'declined': member.accepted === false,
+        'pending': member.accepted === null
+      }">
+        <div
+          v-if="member.accepted === true"
+          class="accepted float-right"
```
SabreCat, line 40: A lot of elements could be `v-once` here, static translation strings and icons

website/client/src/components/groups/invitationListModal.vue
```diff
@@ -0,0 +1,271 @@
...
+        </div>
+      </div>
+    </div>
+  </b-modal>
+</template>
+
+<style lang='scss'>
+  #invitation-list {
+    .modal-header {
+      background-color: #edecee;
```
SabreCat: Would it be possible to get these into the `scoped` section so we can use the `colors.scss` variables instead of hardcoded hex values?

website/client/src/components/groups/questDetailModal.vue
```diff
@@ -0,0 +1,450 @@
...
+            :value="sortBy"
+            class="inline"
+            :inline-dropdown="false"
+            @select="sortBy = $event"
+          />
+        </div>
+      </div>
+
+      <div class="quest-items">
+        <!-- eslint-disable vue/no-use-v-if-with-v-for -->
```
SabreCat: You don't have a `v-if` in here anyway, no need to disable this lint

website/client/src/components/groups/questDetailModal.vue
```diff
@@ -0,0 +1,450 @@
...
+      >
+        <button
+          class="btn btn-primary mt-0 invite-btn"
+          :disabled="!Boolean(selectedQuest) || loading"
+          @click="questInit()"
+        >
+          {{ $t('inviteToPartyOrQuest') }}
+        </button>
+      </div>
+      <div v-if="fromSelectionDialog"
```
SabreCat: These indents are weird, would be just a nudge more readable if they were consistent with the rest of the code style in this file!

website/client/src/components/groups/questDetailModal.vue
```diff
@@ -0,0 +1,450 @@
...
+        <div>
+          <button
+            v-if="!onActiveQuest"
+            v-once
+            class="btn btn-success mb-3"
+            @click="questConfirm()"
+          >
+            {{ $t('startQuest') }}
+          </button>
+          <!-- @TODO don't allow the party leader to
```
SabreCat: If these aren't addressed within the PR, we should probably write them up as tickets--we don't have a good track record of revisiting our `TODO`s

website/client/src/components/groups/questDetailModal.vue
```diff
@@ -0,0 +1,450 @@
...
+      const accepted = await this.questActionsCancelOrAbortQuest();
+
+      if (accepted) {
+        this.close();
+      }
+    },
+    goBackToQuestSelection () {
+      this.selectMode = true;
+    },
+    handleOpen (_, selectQuestPayload) {
```
SabreCat, line 438: The underscore as a placeholder parameter is a weird pattern, what's that for?

website/client/src/components/memberDetailsNew.vue
```diff
@@ -0,0 +1,330 @@
...
+      />
+    </div>
+  </div>
+</template>
+
+<style lang="scss" scoped>
+  @import '~@/assets/scss/colors.scss';
+
+  .stats-bar {
+    margin-left: 0 !important;
```
SabreCat: What's overriding this margin such that we need an `!important`?

website/client/src/components/shops/quests/questDialogContent.vue
```diff
@@ -4,9 +4,15 @@
...
-    <h4 class="title">
+    <h3 class="text-center">
       {{ itemText }}
-    </h4>
+    </h3>
+    <div
+      v-if="group && leader"
+      class="leader-label"
+    >
+      <span>Quest-Owner: </span><user-label :user="leader" />
```
SabreCat: Hardcoded English text. Also, double-check with TC/SO, I'm guessing this shouldn't be a hyphenated term

website/client/src/components/shops/quests/questDialogContent.vue
```diff
@@ -22,36 +28,80 @@
...
+    },
+  },
+  async created () {
+    if (this.group && this.group.quest && this.group.quest.active) {
+      try {
+        const fetchMemberResult = await this.$store.dispatch('members:fetchMember', {
+          memberId: this.group.quest.leader,
+        });
+
+        // this should be handled in fetchMember instead
```
SabreCat: a TODO? something else we should ticket I guess?

website/client/src/components/shops/quests/questPopover.vue
```diff
@@ -0,0 +1,74 @@
...
+      const questsNotDependentToPrevious = ['moon2', 'moon3'];
+      if (item.key in questsNotDependentToPrevious) return false;
+      return true;
+    },
+  },
+};
+</script>
+
+<style scoped>
+
```
SabreCat: funny--I guess it's necessary to have a `<style>` block even if there's nothing in it? 😅

website/client/src/components/shops/quests/questRewards.vue
```diff
@@ -0,0 +1,213 @@
...
+      @click="toggle"
+    >
+      <span class="d-flex justify-content-center">
+        <div
+          v-once
+          class="your-rewards d-flex align-items-center"
+        >
+          <span
+            class="sparkles"
+            v-html="icons.sparkles"
```
SabreCat, line 14: `v-once`, as usual

SabreCat (changes requested):
Phew, what a lot of hard work you've put in here! Thanks for adding that `wizard` to `mage` thing, hah

website/client/src/components/groups/invitationListModal.vue
negue: v-if and v-once wouldn't work together - should the label be in its own span with a v-once in it ?

website/client/src/components/groups/invitationListModal.vue
negue: we can still use the colors here - It didnt worked with the scoped version even with ::v-deep - here always get into the conditional bootrap restyling for stuff  - which not always works in a scoped way

website/client/src/components/groups/questDetailModal.vue
negue: since this whole file was "re-added" and not moved .... (wonder why)

+ refactoring this was probably from previous code I assume - will be removed

website/client/src/components/groups/questDetailModal.vue
negue: since the first param isn't used - it can't stay there as unused

_  is usually the pattern for "don't mind me, I'm unused" .... well except for lodash but thats a different story :D

website/client/src/components/memberDetailsNew.vue
negue: I wonder that myself now :D - maybe the copied version of the memberDetail had it ?  yeah can be removed

website/client/src/components/shops/quests/questRewards.vue
negue: some lines higher the element is already [`v-once`](https://github.com/HabitRPG/habitica/pull/13112/files/f028bcbc4b674cfabd3f83f9863766fe5b27e93d#diff-d41252d940211986ef3af224b38110c17263dfe343458c9edb41a48edf6fd2a0R9)

website/client/src/components/shops/quests/questDialogContent.vue
negue: I meant with that - that the `fetchMember` return needs to be unpacked `data.data` first before its used - I can create a refactor issue for that

negue: or change that method so that all usings don't need to go for `.data.data` all the time

website/client/src/components/groups/questDetailModal.vue
negue: extracted as an Issue: https://github.com/HabitRPG/habitica/issues/13219

website/client/src/components/groups/invitationListModal.vue
SabreCat: ah, that's fair--I suspect I've gotten a little reflexive about that after a hundred such comments from MP 😸

SabreCat (approved):

SabreCat:
All right, let's goooooo! Thanks much, @negue!

PR #13165, opened 2021-04-09
Various Patches (L10N boss num, selList overflow, chat middle mention+refocus on click)

### LINKED ISSUE #9470: Localize comma vs period
We should have a string related to the decimal separator, for example when calculating boss health. Some languages use a comma instead of a period, so numbers such as "977.11 / 1000.00" look very odd.

### LINKED ISSUE #12377: Correct handling for adding second @mention in Tavern and Guild chats

### LINKED ISSUE #12891: Dropdowns in members modal overflow in some languages
### Description

In some languages, the label text for sorting the member list in Parties and Guilds doesn't fit the dropdown element:

image

Helcostr:

* Fixes #9470
* Fixes #12891
* Fixes #12377

### Changes
* Uses browser JS `toLocaleString` to convert numbers to appropriately "marked" numbers
* Limit `membersModal` list selector to 12-unit grid (to prevent wrap around on long text for `sortOptions`)
* Add ellipse based overflow for `membersModal` list selector
* Add padding so that the checkmark for `membersModal` list doesn't overlap with text
* Allow adding `@mention` anywhere (previously was fixed at the end of the text)
* Add re-focus after mouseclick on `@mention` choice

----
UUID: 10cbe941-0e10-4763-9f4f-7a3ae5bb2198

### DIFF
--- a/website/client/src/components/chat/autoComplete.vue
+++ b/website/client/src/components/chat/autoComplete.vue
@@ -135,12 +135,10 @@ export default {
     text (newTextParam, prevTextParam) {
       const delCharFocusBool = prevTextParam.length > newTextParam.length;
       const caretPosition = this.textbox.selectionEnd;
-      const charFocus = (
-        delCharFocusBool ? prevTextParam : newTextParam
-      )[caretPosition - (delCharFocusBool ? 0 : 1)];
+      const lastFocussedChar = isDeleting ? prevText[caretPosition] : newText[caretPosition - 1];
       if (
         newTextParam.length === 0 // Delete all
-        || (charFocus === ' ' && !delCharFocusBool) // End Search
+        || /\s/.test(lastFocussed) // End Search
         || (charFocus === '@' && delCharFocusBool) // Cancel Search
       ) {
         this.searchActive = false;
@@ -172,11 +170,8 @@ export default {
       const lowerCaseSearch = this.currentSearch.toLowerCase();
 
       return this.tmpSelections
-        .filter(option => { // eslint-disable-line arrow-body-style
-          return option.displayName.toLowerCase().indexOf(lowerCaseSearch) !== -1
-            || (option.username
-              && option.username.toLowerCase().indexOf(lowerCaseSearch) !== -1);
-        })
+        .filter(option => option.displayName.toLowerCase().includes(lowerCaseSearch)
+            || (option.username && option.username.toLowerCase().includes(lowerCaseSearch)))
         .slice(0, 4);
     },
     resetDefaults () {
--- a/website/client/src/components/chat/autoComplete.vue
+++ b/website/client/src/components/chat/autoComplete.vue
@@ -90,8 +90,6 @@ export default {
   props: ['selections', 'text', 'caretPosition', 'coords', 'chat', 'textbox'],
   data () {
     return {
-      // Old Regex
-      // atRegex: /(?!\b)@([\w-]*)$/,
       atRegex: /@([\w-]*)$/,
       currentSearch: '',
       searchActive: false,
@@ -132,23 +130,21 @@ export default {
     },
   },
   watch: {
-    text (newTextParam, prevTextParam) {
-      const delCharFocusBool = prevTextParam.length > newTextParam.length;
+    text (newText, prevText) {
+      const delCharsBool = prevText.length > newText.length;
       const caretPosition = this.textbox.selectionEnd;
-      const lastFocussedChar = isDeleting ? prevText[caretPosition] : newText[caretPosition - 1];
+      const lastFocusChar = delCharsBool ? prevText[caretPosition] : newText[caretPosition - 1];
       if (
-        newTextParam.length === 0 // Delete all
-        || /\s/.test(lastFocussed) // End Search
-        || (charFocus === '@' && delCharFocusBool) // Cancel Search
+        newText.length === 0 // Delete all
+        || /\s/.test(lastFocusChar) // End Search
+        || (lastFocusChar === '@' && delCharsBool) // Cancel Search
       ) {
         this.searchActive = false;
         this.searchResults = [];
       } else {
-        if (charFocus === '@') {
-          this.searchActive = true;
-        }
+        if (lastFocusChar === '@') this.searchActive = true;
         if (this.searchActive) {
-          this.searchResults = this.solveSearchResults(newTextParam.substring(0, caretPosition));
+          this.searchResults = this.solveSearchResults(newText.substring(0, caretPosition));
         }
       }
     },
@@ -161,9 +157,9 @@ export default {
     this.grabUserNames();
   },
   methods: {
-    solveSearchResults (searchTextWip) {
+    solveSearchResults (textFocus) {
       if (!this.searchActive) return [];
-      const regexRes = this.atRegex.exec(searchTextWip);
+      const regexRes = this.atRegex.exec(textFocus);
       if (!regexRes) return [];
       this.currentSearch = regexRes[1]; // eslint-disable-line vue/no-side-effects-in-computed-properties, max-len, prefer-destructuring
 
--- a/website/client/src/components/chat/autoComplete.vue
+++ b/website/client/src/components/chat/autoComplete.vue
@@ -90,12 +90,14 @@ export default {
   props: ['selections', 'text', 'caretPosition', 'coords', 'chat', 'textbox'],
   data () {
     return {
-      atRegex: /(?!\b)@[\w-]*$/,
+      // Old Regex
+      // atRegex: /(?!\b)@([\w-]*)$/,
+      atRegex: /@([\w-]*)$/,
       currentSearch: '',
       searchActive: false,
-      searchEscaped: false,
       currentSearchPosition: 0,
       tmpSelections: [],
+      searchResults: [],
       icons: Object.freeze({
         tier1,
         tier2,
@@ -114,7 +116,7 @@ export default {
   computed: {
     autocompleteStyle () {
       function heightToUse (textBox, topCoords) {
-        const textBoxHeight = textBox['user-entry'].clientHeight;
+        const textBoxHeight = textBox.clientHeight;
         return topCoords < textBoxHeight ? topCoords + 30 : textBoxHeight + 10;
       }
       return {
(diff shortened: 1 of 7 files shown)


website/client/src/components/chat/autoComplete.vue
```diff
@@ -128,38 +130,29 @@ export default {
...
-        })
-        .slice(0, 4);
-    },
-
   },
   watch: {
-    text (newText) {
-      if (!newText[newText.length - 1] || newText[newText.length - 1] === ' ') {
+    text (newTextParam, prevTextParam) {
+      const delCharFocusBool = prevTextParam.length > newTextParam.length;
```
benkelaar: I would name this something like `wereCharsRemoved` or `isDeleting`

website/client/src/components/chat/autoComplete.vue
```diff
@@ -128,38 +130,29 @@ export default {
...
+      const caretPosition = this.textbox.selectionEnd;
+      const charFocus = (
+        delCharFocusBool ? prevTextParam : newTextParam
+      )[caretPosition - (delCharFocusBool ? 0 : 1)];
```
benkelaar: I think this line would be clearer if you stick to a single ternary (And I'd recommend renaming `charFocus` to something like `lastFocussedChar`)

```suggestion
      const lastFocussedChar = isDeleting ? prevText[caretPosition] : newText[caretPosition - 1];
```

website/client/src/components/chat/autoComplete.vue
```diff
@@ -128,38 +130,29 @@ export default {
...
-              && option.username.toLowerCase().indexOf(lowerCaseSearch) !== -1);
-        })
-        .slice(0, 4);
-    },
-
   },
   watch: {
-    text (newText) {
-      if (!newText[newText.length - 1] || newText[newText.length - 1] === ' ') {
+    text (newTextParam, prevTextParam) {
```
benkelaar: I don't think adding `Param` to parameters is adding useful information to the parameter names.  It's not something I've seen in the rest of the codebase so I'd recommend leaving it out here as well.

website/client/src/components/chat/autoComplete.vue
```diff
@@ -90,12 +90,14 @@ export default {
   props: ['selections', 'text', 'caretPosition', 'coords', 'chat', 'textbox'],
   data () {
     return {
-      atRegex: /(?!\b)@[\w-]*$/,
+      // Old Regex
+      // atRegex: /(?!\b)@([\w-]*)$/,
+      atRegex: /@([\w-]*)$/,
```
benkelaar, line 93: Why remove the negative word boundary lookahead?  I agree that a positive lookbehind for negated word boundary would probably make more sense but the search should not start when typing e.g. an e-mail address right?

Ah is this to enable searches anywhere in the text?  If that's the case you should drop the end-of-text anchor, however that worries me that you might find earlier mentions in the same text block with this regex and the selection will jump back to that mention instead.  I didn't run your branch yet, so not sure, but would be good to check.

Also it's not necessary to leave old code around.  It's all in the git history anyway.

website/client/src/components/chat/autoComplete.vue
```diff
@@ -128,38 +130,29 @@ export default {
...
-      if (!newText[newText.length - 1] || newText[newText.length - 1] === ' ') {
+    text (newTextParam, prevTextParam) {
+      const delCharFocusBool = prevTextParam.length > newTextParam.length;
+      const caretPosition = this.textbox.selectionEnd;
+      const charFocus = (
+        delCharFocusBool ? prevTextParam : newTextParam
+      )[caretPosition - (delCharFocusBool ? 0 : 1)];
+      if (
+        newTextParam.length === 0 // Delete all
+        || (charFocus === ' ' && !delCharFocusBool) // End Search
```
benkelaar: Is it enough to check on space characters for search end?  What about other types of whitespace characters?  Also I don't think the negative deletion check is necessary for spaces, at that point the search should have ended already anyway right?

```suggestion
        || /\s/.test(lastFocussed) // End Search
```

website/client/src/components/groups/questSidebarSection.vue
```diff
@@ -149,7 +149,19 @@
...
+                      .toLocaleString(user.preferences.language, {
+                        minimumFractionDigits: 2,
+                        maximumFractionDigits: 2
+                      })
+                  }} / {{
+                    parseFloat(questData.boss.hp)
+                      .toLocaleString(user.preferences.language, {
+                        minimumFractionDigits: 2,
+                        maximumFractionDigits: 2
+                      })
```
benkelaar: Would be good to externalize this functionality into a method.  Perhaps something like `localizeNumber(number, digits)`

website/client/src/components/groups/questSidebarSection.vue
```diff
@@ -160,9 +172,15 @@
...
-                >{{ (user.party.quest.progress.up || 0) | floor(10) }} {{ $t('pendingDamageLabel') }}</span> <!-- eslint-disable-line max-len -->
+                <span class="float-right">
+                  {{
+                    (Math.floor((user.party.quest.progress.up || 0) * 10) / 10)
+                      .toLocaleString(user.preferences.language, {
+                        minimumFractionDigits: 1,
+                        maximumFractionDigits: 1
+                      })
+                  }} {{ $t('pendingDamageLabel') }}
+                </span> <!-- eslint-disable-line max-len -->
```
benkelaar: Line is no longer longer than max-len.
```suggestion
                </span>
```

website/client/src/components/ui/selectList.vue
```diff
@@ -53,9 +55,19 @@
 <style lang="scss" scoped>
   @import '~@/assets/scss/colors.scss';
 
+  .select-list .select-list-button-wrap {
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
   .select-list ::v-deep .selectListItem {
```
benkelaar: Here you can see the `::v-deep` selector being used, that did not work for you here?

website/client/src/components/chat/autoComplete.vue
```diff
@@ -170,11 +163,26 @@ export default {
     this.grabUserNames();
   },
   methods: {
+    solveSearchResults (searchTextWip) {
```
benkelaar: Is Wip for `Work in progress`?  Why not just `searchText`?

website/client/src/components/chat/autoComplete.vue
```diff
@@ -170,11 +163,26 @@ export default {
...
+      this.currentSearch = regexRes[1]; // eslint-disable-line vue/no-side-effects-in-computed-properties, max-len, prefer-destructuring
+
+      const lowerCaseSearch = this.currentSearch.toLowerCase();
+
+      return this.tmpSelections
+        .filter(option => { // eslint-disable-line arrow-body-style
+          return option.displayName.toLowerCase().indexOf(lowerCaseSearch) !== -1
+            || (option.username
+              && option.username.toLowerCase().indexOf(lowerCaseSearch) !== -1);
+        })
```
benkelaar:
```suggestion
        .filter(option => option.displayName.toLowerCase().includes(lowerCaseSearch)
            || (option.username && option.username.toLowerCase().includes(lowerCaseSearch)))
```

benkelaar (commented):
Awesome merge request!  Thanks a lot!

I added a couple suggestions for things I think would improve the code even more.

website/client/src/components/chat/autoComplete.vue
Helcostr: Agreed.

website/client/src/components/chat/autoComplete.vue
Helcostr: Ah yeah, this was relic code that I had when I stored to focusText (including the focusChar)

Which is why the double ternary still exists.

Good suggestion.

website/client/src/components/chat/autoComplete.vue
Helcostr: Hehe... my secret weapon... XD

First of all, negative lookahead is awkward since the mentions trigger without boundaries, so typing an email is something I gotta let slide. If the user is annoyed while typing an email, they can escape the search (or the search will pop up invalid).

> Try typing `@mention1@mention2` and sending it. It works. The script will auto insert spaces, but if ur a human typing, u can type w/o the spaces. I just did think of an edge case that I need to check though... thx for reminding me. Edit: I think the edge case I was thinking of is fine... can't remember what the issue was exactly anyways...

The end of line is necessary to find the last mention of the textblock, since I didn't start the regex with a "start here" index.

Also old code: whoops, missed cleaning the regex comment when I was keeping it around as a backup (a big flag for me)

The comment issue has been resolved. Looking for feedback on the regex in question.

website/client/src/components/chat/autoComplete.vue
Helcostr: Relic code I was mimicking, will implement this regex.

website/client/src/components/groups/questSidebarSection.vue
Helcostr: Okeydokey, I was wondering if I should do this.

website/client/src/components/groups/questSidebarSection.vue
Helcostr: No idea how eslint works so I kept it in. Will remove.

website/client/src/components/ui/selectList.vue
Helcostr: It didn't.

I can try again... maybe I might need to do a double v-deep 🤔...

Anyways, wrapping feels nicer XD. But I will try the v-deep method again to avoid too many DOM objects.

Edit: wow... v-deep worked 🤔 haha

website/client/src/components/chat/autoComplete.vue
Helcostr: cuz the text is being worked on at the moment (as opposed to the trailing end of the text that is dropped right b4 this parameter is passed into the method: that isn't being worked on).

tbh, I would not name it search. probably I would name it `textWip` since that's the client's focus currently.

Oh maybe I should name it `textFocus`?

website/client/src/components/chat/autoComplete.vue
Helcostr: It's to prevent conflicting naming schemes, since I noticed a few lines up (or down, I don't remember what direction), newText is a totally different meaning: it means text that the programmer generated, not the new text as part of the parameter of watch.

website/client/src/components/chat/autoComplete.vue
Helcostr: I think I maybe hitting the esline limit... gotta check

Helcostr:
@benkelaar am I suppose to re-request another review? There is a button to do so, but it's greyed out.

website/client/src/components/chat/autoComplete.vue
benkelaar: I'm not sure what you mean, the whole point of the negative lookahead is to make the mentions trigger only without boundaries right?  Historically the choice was made to favour not launching searches in e-mails over being able to type two mentions after each other without spaces.

I'm not sure why that behaviour should change now.  Was that discussed in the issue perhaps?  I do agree that the negative lookahead is not necessary, since the word boundary match is a zero-character match anyway. `/\B@([\w-]*)$/` should have the exactly the same behaviour.

Or am I missing something?

(And for the other change, the addition of a capturing group for the username capture is a good idea. I guess you could argue to only match after the first character has been typed though, (So that would be `/\B@([\w-]+)$/`) but I guess that would also be a change from current behaviour)

website/client/src/components/chat/autoComplete.vue
benkelaar: Yeah, might be a personal preference thing, I always try to put as little acronyms in my code as possible since they require more mental overhead to parse from the reader (and assume the reader knows about them in the first place).

Perhaps something along the lines of `textFilter` or `textQuery`?

Come to think of it `solve` is also a bit non-specific.  Perhaps `filterSearchResults` would be a better method name?

website/client/src/components/chat/autoComplete.vue
Helcostr: Oh dang! Where is the history?

Anyways, there was no bug report about this, but I removed it because I believe that it should work this way.

Plus it's not like each search sends a data request to Habitica...  so "phantom searches" in my book are fine. Especially when a user is typing fast. It's not like the logic of tagging is determined client side: it's done server side. And a no-space-tag is totally possible.

The user experience on my end seems fine... 🤔 

And no it's slightly different behavior... the search will include or not include the word boundary.

benkelaar:
> @benkelaar am I suppose to re-request another review? There is a button to do so, but it's greyed out.

Yeah, I can't "officially approve" the PR, since I'm just another contributor and as such my opinion doesn't matter to the merging process.

As far as I know, the person who was primarily responsible for handling these merges left for another job, so there's a bit of a backlog from the core team handling PRs.  I'm just trying to help out to get the PRs into the best shape possible so they don't have to spend too much time on them 😄.

As far as I know @SabreCat is the person now responsible for looking in to reviews, but I'm sure he'll get to it when he'll get to it 😃

website/client/src/components/chat/autoComplete.vue
Helcostr: Filter is weird. Query is more accurate, but still weird.

Search results is completely wrong.

I stuck with `textFocus` eventually.

Helcostr:
> Yeah, I can't "officially approve" the PR, since I'm just another contributor and as such my opinion doesn't matter to the merging process.

@benkelaar I was thinking of just requesting ur help with the review again... but I guess re-requesting is only available for those who actually have approving-the-PR-permissions?

benkelaar:
I guess it depends on how you "finalize" the review.  I didn't set the state to "request changes".  If I had I think you would have been able to re-request a review using that button.  But comments always work 😆

Helcostr:
This PR is ready for review.

Alys:
@Helcostr Thank you for your work on this! Could you please resubmit with one PR per issue? That will make it easier for us to test and review them, and it will also help if we ever need to refer back to how each issue was fixed (e.g., if a future code change makes the issue or a similar one re-appear, which happens sometimes).

Helcostr:
@Alys breakout done.

Issue #10934, opened 2019-01-06
Challenge and Group Plan tags cannot be sorted


Alys:
Recently, https://github.com/HabitRPG/habitica/pull/10911 was done to allow a user's personal tags to be sorted. I'm opening this issue to allow sorting to be done for Challenge tags and Group Plan tags too.

I'm marking this as a medium level task since it's likely that the code that was created for personal task sorting can be reused or expanded to allow sorting the other types of tags.

Note that each type of tag can only be sorted separately. E.g., you can sort all Challenge tags into a particular order amongst themselves but you can't mix Challenge and Group Plan and personal tags together - the front-end structure of the new website no longer allows that. :(

Mira-M:
I can take this one, please!

Alys:
@Mira-M I'm sorry about the delay in us replying! Please do go ahead with this if you're still interested! Comment here if you have questions.

shanaqui:
@Mira-M Are you still interested in working on this? If so, please let us know!

shanaqui:
I'll set this back to help wanted since there was no reply. :)

codexdelta:
hi, I would be very interested in resolving this. 

I can see that I can sort the tags over here but not the challenges. The desired behavior is to make the challenges and groups sortable over here, please correct me if I am wrong.


![image](https://user-images.githubusercontent.com/6303170/104452834-8ef2db00-55c9-11eb-8a8a-2f519279e465.png)

lizapritchard:
Hi @Alys I noticed there hasn't been much activity here in a while, could I work on this issue?

lizapritchard:
I was able to make the challenges draggable like the tags but am having a little trouble handling groups. Right now, groups aren't editable at all from this module (they can't be added, deleted, or dragged) should groups be editable? Or just draggable? 

I'm also having a little trouble adding groups to my local db to test the new functionality. I modified my groups document to include the following and I'm seeing the group under the groups tab but it still does not appear in my tags in the tasks tab
`"purchased" : {
        "plan" : {
            "consecutive" : {
                "count" : 0,
                "offset" : 0,
                "gemCapExtra" : 0,
                "trinkets" : 0
            },
            "quantity" : 3,
            "extraMonths" : 0,
            "gemsBought" : 0,
            "mysteryItems" : [],
            "customerId" : "cus_123456789",
            "dateCreated" : ISODate("2021-04-11T17:49:59.589Z"),
            "dateTerminated" : null,
[...]

PR #13175, opened 2021-04-14
Challenge tags can be sorted, Partial fix for #10934

lizapritchard:
Partial Fix for #10934 

### Changes
In the tags modal on the tasks screen, challenges can now be dragged to reorder similar to the existing behavior of tags created in #10911. Saving these changes allows the order to persist after page reload. 
Screen Shot 2021-04-14 at 1 44 54 PM
Screen Shot 2021-04-14 at 1 44 44 PM

(Reverted commits are for separate PR)

----
UUID: 6179e6e4-1070-43a1-b1df-c1aad12a3ef7

### DIFF
--- a/website/client/src/components/chat/autoComplete.vue
+++ b/website/client/src/components/chat/autoComplete.vue
@@ -90,12 +90,12 @@ export default {
   props: ['selections', 'text', 'caretPosition', 'coords', 'chat', 'textbox'],
   data () {
     return {
-      atRegex: /(?!\b)@[\w-]*$/,
+      atRegex: /(?!\b)@([\w-]*)$/,
       currentSearch: '',
       searchActive: false,
-      searchEscaped: false,
       currentSearchPosition: 0,
       tmpSelections: [],
+      searchResults: [],
       icons: Object.freeze({
         tier1,
         tier2,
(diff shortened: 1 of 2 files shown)


website/client/src/components/challenges/findChallenges.vue
```diff
@@ -147,7 +147,7 @@ export default {
           value: 'none',
         },
         {
-          text: this.$t('participants'),
+          text: this.$t(' ipants'),
```
SabreCat: This looks unintentional. iPants!

website/client/src/components/tasks/user.vue
```diff
@@ -56,9 +56,8 @@
                   class="row"
                   :class="{'no-gutters': !editingTags}"
                 >
-                  <template v-if="editingTags && tagsType.key === 'tags'">
+                  <template v-if="editingTags && tagsType.key !== 'groups'">
```
SabreCat: Could we remove the exclusion logic for groups so that it works there as well?

SabreCat (changes requested):
One question, one file that looks like it was included unintentionally; if we get those cleaned up, this will be ready to go!

website/client/src/components/tasks/user.vue
SabreCat: oh, never mind, group tags aren't editable

Issue #12971, opened 2021-01-17
XML Data export error


shanaqui:
Report from Ceran:

> I tried to export my user date with the options in the settings, using the xml version, but it only gave me a white page with the error message
> 
> {"success":false,"error":"InternalServerError","message":"An unexpected error occurred."}
> 
> Checked, exporting the cvs habit history worked, but since I wanted my to-dos, not much use. Exporting the json file works too, but I don't know how to open that file to check contents.

Replicated by me.

Similar to (or the same as) previously fixed #12482 and #10100 

---

EDIT by Alys 2021-04-19

When this issue is fully fixed, the fix must include an update for the apidoc comments for the XML export route. Currently they say:
> `@api {get} /export/userdata.xml Export user data in XML format`
> `@apiName ExportUserDataXml`
> `@apiDescription This XML export feature is not currently working (https://github.com/HabitRPG/habitica/issues/10100).`

habitrpguser:
I just want to confirm this error. It is exactly the same for me.

Alys:
This happens for me too, and we have a report in RaB today.

Helcostr:
I will investigate to see if I can fix this issue.... I am pretty sure I can, but I will double check.

Alys:
@Helcostr Thank you!

Helcostr:
With a fresh account, XML exporting is fine. And since it's an InternalServerError, I don't have many hints to go off of.

Can anyone narrow this issue down?

I think my contribution here is to wrap this function in an error handler so that the error is more clear.

Edit:

There are tons of warning signs around to not leak stuff... I am not sure how best to handle error returning, and so I decided to kick the key issue down the road and have it always return xml, at the cost of obfuscating the bad key characters. (Assuming the issue is with a "uuid key" issue)

Alys:
@Helcostr There's nothing in Habitica's server logs for my failed attempt to export my data, so narrowing it down will be hard. :(

I like this partial fix you've done! That looks promising.

---

As a note for the future when (if) this issue is fully fixed, we should update the apidoc comments for the XML export route. Currently they say:
> `@api {get} /export/userdata.xml Export user data in XML format`
> `@apiName ExportUserDataXml`
> `@apiDescription This XML export feature is not currently working (https://github.com/HabitRPG/habitica/issues/10100).`

I've added a note to the top post about this.

Alys:
[...]

PR #13181, opened 2021-04-17
Prevent invalid characters from being used in XML - partial fix for #12971

Helcostr:

Partial fix for #12971

### Changes
Patching issue of never returning XML, at the cost of "slightly obfuscated" xml key (I expect issues to be sent)

----
UUID: 10cbe941-0e10-4763-9f4f-7a3ae5bb2198

### DIFF
--- a/website/server/libs/xmlMarshaller.js
+++ b/website/server/libs/xmlMarshaller.js
@@ -25,6 +25,7 @@ export function marshallUserData (userData) {
 
   return js2xml.parse('user', userData, {
     cdataInvalidChars: true,
+    replaceInvalidChars: true,
     declaration: {
       include: false,
     },


Alys:
Since this simply turns on an option for the js2xmlparser library, it should be safe. It seems like it has a good chance of helping with the issue, or at least can't make the situation noticeably worse. Therefore I'm merging this.

Thank you @Helcostr ! This gives you Tier 1 as a Blacksmith. Congratulations!

Note that as your tier increases, it takes progressively more work to achieve the next one, so for the higher tiers, larger PRs or a few PRs are required. You can read more about [contributor tiers and rewards here on the wiki](http://habitica.fandom.com/wiki/Contributor_Rewards).

PR #13183, opened 2021-04-21
Fix "Not Participating" filter checkbox on Discover Challenges screen, Fixes #13025

### LINKED ISSUE #13025: "not participating" search/filter checkbox doesn't work in Discover Challenges page
On the [Discover Challenges](https://habitica.com/challenges/findChallenges) page, if you tick the "Not Participating" checkbox (in the Role section of the filters), you see Challenges that you are participating in.

The expected behaviour is that you see only Challenges that aren't currently joined by you.

The same occurs if you tick both "Not Participating" and "Not Owned" so anyone fixing this bug should check that the fix works for that use case too.

_Thanks to @LindyBlues for reporting this!_

lizapritchard:

Fixes #13025

### Changes

- Filtering added on the front-end of the "discover challenges" view,  to allow the user to view challenges they _are_ and _are not_ currently participating in. Mimics function in "my challenges" view. 
- Removed "role" filter from "my challenges" and "discover challenges" views (deemed extraneous by @saraolson in #13025 followups)
Screen Shot 2021-04-21 at 10 26 30 AM
Screen Shot 2021-04-21 at 10 26 41 AM

----
UUID:  6179e6e4-1070-43a1-b1df-c1aad12a3ef7

### DIFF
--- a/website/client/src/components/challenges/findChallenges.vue
+++ b/website/client/src/components/challenges/findChallenges.vue
@@ -31,7 +31,8 @@
       </div>
       <div class="row">
         <div
-          v-if="!loading && filteredChallenges.length === 0"
+          v-if="!loading &&
+          this.filteredChallenges.length === 0"
           class="no-challenges text-center col-md-6 offset-3"
         >
           <h2 v-once>
@@ -41,7 +42,7 @@
       </div>
       <div class="row">
         <div
-          v-for="challenge in filteredChallenges"
+          v-for="challenge in this.filteredChallenges"
           :key="challenge._id"
           class="col-12 col-md-6"
         >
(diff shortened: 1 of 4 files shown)


website/client/src/components/challenges/findChallenges.vue
```diff
@@ -171,7 +171,23 @@ export default {
   computed: {
     ...mapState({ user: 'user.data' }),
     filteredChallenges () {
-      return this.challenges;
+      const { filters } = this;
```
SabreCat: The code is identical between the two Challenge Vue files, now--could this be abstracted as a mixin?

website/client/src/components/challenges/sidebar.vue
```diff
@@ -34,7 +34,7 @@
           </div>
         </div>
       </filter-group>
-      <filter-group :title="$t('role')">
+      <!-- <filter-group :title="$t('role')">
```
SabreCat: OK to delete these rather than commenting out. Git history will preserve it if we need to restore it!

website/client/src/components/challenges/sidebar.vue
```diff
@@ -231,6 +244,7 @@ export default {
         categories: this.categoryFilters,
         roles: this.roleFilters,
```
SabreCat: We can remove role filtering here too

SabreCat (changes requested):
Thanks for your help here! See minor comments in review, but this shouldn't be difficult to move across the finish line!

website/client/src/components/challenges/findChallenges.vue
SabreCat: We already have the `challengeUtilities` mixin, so it could live there

SabreCat:
One functional thing I'm seeing: selecting _both_ Participating and Not Participating causes the filter to show only the Not Participating challenges. Compare the behavior of the Owned and Not Owned, which if you tick both acts the same as if neither is checked, i.e., not excluding any Challenges from the current list.

lizapritchard:
@SabreCat thanks for the feedback! I think I addressed all of these issues in the new commits, but I was a little unsure about the mixin stuff so there might be some style conventions here that I'm missing

SabreCat:
Looks great! Thanks for the updates! We'll get this into the staging environment and from there to production. Credited toward your second contributor tier!

PR #13222, opened 2021-04-29
babel: enable optional chaining

negue:
This enables `optional chaining` in our website

example, instead of 

`if (someProp && someProp.subProp && someProp.subProp.realProp)` 

you can write

`if (someProp?.subProp?.realProp)`


SabreCat (approved):
Cool. This should make a number of things more readable if we update them as we go!

PR #13249, opened 2021-05-06
Disable Loggly client by providing an empty token.

EliRibble:
This is useful for self-hosters that intend to do logging a different way.


SabreCat (approved):
Makes sense to me! Thanks!

PR #13300, opened 2021-05-27
Rework Notifications:  separate and bundle to an amount of 4 

negue:
Instead of having all notifications visible at once (and hidden one per second)

The notifications now bundled into two pairs which are shown after the previous pair is done.

The logic for that .. was quite "interesting" to do :D - it might've been easier with rxjs


website/client/src/components/snackbars/notification.vue
```diff
@@ -111,6 +108,8 @@
     width: 300px;
     margin-left: 1em;
     margin-bottom: 1em;
+
+    transition: opacity .5s, top .5s;
```
negue, line 129: will be removed

website/client/src/components/snackbars/notifications.vue
```diff
@@ -19,40 +29,294 @@
...
 
 <script>
 import { mapState } from '@/libs/store';
 import notification from './notification';
+import { sleepAsync } from '../../../../common/script/libs/sleepAsync';
+
+const notificationsVisibleAtOnce = 4;
+const removalInterval = 2500;
+const delayDeleteAndNew = 60;
+const delayFilling = 240;
```
SabreCat: For magic numbers and other simple constants defined at top of file, we usually use ALL_CAPS_SNAKE naming convention

website/client/src/components/snackbars/notifications.vue
```diff
@@ -19,40 +29,294 @@
...
   },
+  props: {
+    preventQueue: {
+      type: Boolean,
+      default: false,
+    },
+    debugMode: {
+      type: Boolean,
+      default: false,
+    },
```
SabreCat, line 89: Do we need to keep the debug functionality in all the way to production?

website/client/src/components/snackbars/notifications.vue
```diff
@@ -19,40 +29,294 @@
...
+      }
+
+      this.updateAllowedToFillAgain();
+    },
+    async triggerFillUntilFull () {
+      for (let i = 0; i < notificationsVisibleAtOnce; i += 1) {
+        this.debug(`fill ${i}`);
+        this.fillVisibleNotifications(this.notificationStore);
+
+        await sleepAsync(delayFilling); // eslint-disable-line no-await-in-loop
```
SabreCat: Could this be done by `map`ping a slice of notificationStore instead of ignoring the relevant eslint rule?

SabreCat (changes requested):
Pretty minor stuff on the code review. Will do a short functional test as well!

website/client/config/storybook/sprites.scss
```diff
@@ -0,0 +1,3 @@
+@for $i from 0 through 29 {
```
SabreCat: We're up to 30 now 😩  I should probably implement something like this in app.vue, though.

website/client/src/components/snackbars/notifications.vue
SabreCat: Oh, never mind, I misread this.

website/client/src/components/snackbars/notifications.vue
negue: sadly there aren't any compiler flags that could remove it on production...  (atleast I don't know any, but if there are we could use something like this?)

I think it should stay (even if its only enabled in storybook) the amount of how often it had to be debugged - these code / debug-logs helped alot to speed things up - otherwise me or someone else who needs to debug it for (hopefully no) other issues it would speed up to get all needed logs in place instead of going through again and adding them again and again

(thats why the property to enable it)

website/client/config/storybook/sprites.scss
negue: I can add it in app itself then always the same file used by both app and storybook

website/client/src/components/snackbars/notifications.vue
negue: will change those 👍

website/client/config/storybook/sprites.scss
SabreCat: That sounds good!

negue: should the way it is in the app.vue still be in separated <style>-Tags? or using it like that it would be combined into one ? would this "break" something? like was it separated into css files needed?

I know the image file itself that those are split up that was needed but css files ?

negue: on the other side - now the sprites.css file is only ONE request instead of 31 (added the /sprites.css into it as well) which should speed things up ... I think :D

SabreCat (approved):

SabreCat:
We'll stick this on Staging for a few days just in case, but looking good!!

PR #13315, opened 2021-06-05
Correct member modal behavior in the party section when selecting the invite tab before the invited user accepts the invitation 

### LINKED ISSUE #13107: Party members list can get out of sync in some flows with invitations
Reported by @citrusella in the Report a Bug Guild:

> Strange behavior I encountered on my test account when doing some party shenanigans (in order to easily access a new, second test account's profile to send it some gems to test a thing):
>
> Here are the steps that happened that seem relevant in order, though I'm not sure every piece would be required to replicate it:
>
>    I invited a user (the second test account) to my test account's party
>    I opened the member list modal in the party and switched to the invited users tab
>    (Here I did some probably irrelevant things, like sending the new test account gems and having it try to buy a quest)
>    Invited user joined the party (because I suddenly wanted to try to start the quest using it, which I did and then aborted)
>    Invited user deleted their account
>    Member list modal (which was last open in the invited state) wouldn't display anything below the sort/search header, even though it should have listed @citrustest as the only member at that point
>    This persisted between pages (i.e. if I left the party page and came back) and didn't clear up until I had reloaded the website completely (though I didn't try to use the sync button)
>
>I suspect this has to do with it having been on the invite tab and either happened when the account joined the party or when it got deleted.

After some further investigation, @citrusella found a simpler flow to reproduce the situation:

>  Invite someone
>  Open the member modal before they accept, switch to the invite tab, close the modal and don't open it again
>  User accepts invite and now there are zero invites

AlanGCruz:
Fixes #13107 

### Changes
Update the `getMembers` method to change the selectedPage(or tab selected) data to be equal to `'members'` if the last tab selected by the user was the `invites tab` but there's no pending invites left.

UUID: 79fc8f1a-fa61-4ed2-aeee-999d4c3f3ee1

### DIFF
--- a/website/client/src/components/groups/membersModal.vue
+++ b/website/client/src/components/groups/membersModal.vue
@@ -577,6 +577,7 @@ export default {
           groupId,
           includeAllPublicFields: true,
         });
+        if (this.selectedPage === 'invites' && invites.length === 0) this.viewMembers();
         this.invites = invites;
       }
     },


SabreCat:
Thanks for this, @AlanGCruz, and welcome to the ranks of the Blacksmiths at tier 1! Further tiers require multiple or larger PRs to attain, but keep helping out and we'll express our gratitude accordingly!

PR #13317, opened 2021-06-05
Quest Sidebar: Boss Pending Health inside the normal Bar

negue:
This puts the "pending HP bar" inside the normal one:

grafik

Example ^  - 20 HP pending - of 30 HP current


SabreCat:
Looking good! Thanks @negue !

PR #13327, opened 2021-06-08
Fix: Quest Rewards

### LINKED ISSUE #13326: Quest modals say "New Item" where they should say what the item is
When a Quest modal shows you the rewards for a quest, any reward that is not Gold or Experience is listed as "New Item", meaning that you have no way to tell what the item is if you don't recognise the image (that's especially undesirable for Pet Quests since some of the Eggs have names that are noticeably different than the Quest / Boss name, and so you can't guess what they'll be called, and so, for example, you can't use the rewards list to decide whether you want to spend gems on that particular quest or not).

This bug occurs in the Quest modal that you see in all locations
   * Shops > Quests
   * Inventory > Items
   * Party > Select Quest
   * when a Quest finishes
   * etc
There'll most likely be a single piece of shared code that you can update to fix this in all places at once.

The names of the items can be found in the `text` attribute in the `drop.items` data for each quest. For example in the lower half of this snippet:
https://github.com/HabitRPG/habitica/blob/07b9cc533ba79aa9bc7598d9d142eef442eca134/website/common/script/content/quests.js#L473-L497

image

image

---

_Thanks to  @EricadawnK and @Acorn_Sprout for reporting this._

negue:
- Added Pets / Mounts in the Preview
- Added all other popover labels - fix #13326

### DIFF
--- a/website/client/src/components/shops/quests/buyQuestModal.vue
+++ b/website/client/src/components/shops/quests/buyQuestModal.vue
@@ -158,28 +158,6 @@
       vertical-align: middle;
     }
 
-    .value {
-      width: 28px;
-      height: 32px;
-      font-family: Roboto;
-      font-size: 24px;
-      font-weight: bold;
-      line-height: 1.33;
-      vertical-align: middle;
-
-      &.gems {
-        color: $green-10;
-      }
-
-      &.gold {
-        color: $yellow-10;
-      }
-
-      &.hourglasses {
-        color: $hourglass-color;
-      }
-    }
-
     button.btn.btn-primary {
       margin-top: 24px;
       margin-bottom: 24px;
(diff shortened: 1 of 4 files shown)


SabreCat (changes requested):
I see just one issue: pet and mount rewards push the "New Item" text out of the div:

image

If we can get that lined up correctly, this is good to go!

negue:
the mounts / pets have different sizes (since all sprites have lots of additional space inside the image data) which makes it hard to position right 

I could try to apply a special rule for mounts / pets ... lets hope it works generally on mount/pet and not that it needs a new negative margin for each mount/pet

negue:
Added max sizes for all images and Pet images now pushed to the top by `-20px`.

Pushed to delta to try out

SabreCat:
Excellent! Looks great now!

SabreCat (approved):

PR #13330, opened 2021-06-10
Class icon misalignment fixed. fixes #13316

### LINKED ISSUE #13316: Class icon is in middle of avatar in Profile screens and Edit Avatar screen, and might be wrong in header
When you look at a profile screen or the edit avatar screen, the class icon is around the middle of the avatar.

Marking this "needs design" for a decision on where it's meant to be.

image

image

FarisPalayi:

[//]: # (Put Issue  #13316 , if applicable. This will automatically close the issue if your PR is merged in)
Fixes #13316 

### Changes

Solved the issues related to size and position of the **class icon** in avatar profile and profile edit modal.

Before: 
<div align='left'>

modal before

avatar before

</div>

After :

<div align='left'>

modal after

avatar after

</div>

### Problem No: 1 - CSS variable with no unit value

`:style="{'--badge-size': badgeSize }"` - this inline style v-bind takes a prop named _badgeSize_ which is a type of Number
```js run
badgeSize: {
      type: Number,
      default: 32,
    },
```
So, after compilation, the variable becomes `--badge-size: 32` and what it should be is `--badge-size: 32px`.

So I changed the inline style to this - `:style="{'--badge-size': badgeSize + 'px'}"`.

### Problem No: 2 - Deprecated SASS syntax (I think)

There was a CSS property which used the deprecated '/' syntax for division --> ` bottom: -var(--badge-size) / 2;` 

So I changed that into this --> `bottom: calc((var(--badge-size) / 2) * -1);`

- I also doubt that the problem occurred because of the hyphen `-` before `var()`. But I'm not sure.
Whichever the case, it stayed the same `bottom: -var(--badge-size) / 2;` even after compilation.

----
UUID: 3fbf39e8-67a1-48cc-a5a5-821a2e33d2ad


SabreCat:
Thanks for this, @FarisPalayi, looks good. Welcome to the ranks of the Blacksmiths at tier 1! Further tiers will take multiple or larger PRs to attain, but keep helping out and we'll express our gratitude accordingly!

PR #13382, opened 2021-07-07
WIP: Only unlock Masterclasser 2-4 if 1 is unlocked

### LINKED ISSUE #13013: inconsistent unlock behaviour for Mystery of Masterclassers quest line quests
_This is suggestion-discussion. It's borderline between bug and feature change request, since I believe the current code is behaving as written, but I thought it best to put here. Even if it's closed with a note that the behaviour is intentional, it will be good to have in GitHub so that we can reference it if it's reported again and/or link to it from the wiki's Quests page._

_As reported by @Nakonana (33bb14bd-814d-40cb-98a4-7b76a752761c):_

"Might not be a bug, but it's still a bit funny:

"To purchase the 1st quest of the Mystery of Masterclassers quest line, [Read Between the Lines](https://habitica.fandom.com/wiki/Read_Between_the_Lines) [`lostMasterclasser1`], you have to complete the last parts of the other quest lines. But even if you don't meet the requirements yourself, you can still participate in the Mystery of the Masterclassers quests, if a party mate, who meets the requirements, invites you to those quests. So far, so good and as expected, however after finishing the party mate's Read Between the Lines, you are now able to purchase [Assembling the a'Voidant](https://habitica.fandom.com/wiki/Assembling_the_aVoidant) [`lostMasterclasser2`], but not Read Between the Lines. [screenshot below]

"I would expect that the quests are either all unlocked after I complete them or that they all remain locked until I meet the requirements for the 1st quest. Having only one of them remain locked after completion seems a bit inconsistent."

image

ashaindlin:

Fixes #13013

### Changes

Purchasing any of the quests in the Mystery of the Masterclasser questline now requires the user to have completed the last part of each of the four quest chains that unlock the first one, even if they've already completed the previous quest in the questline in a party. The quest popover displays the unlock requirement text for lostMasterclasser1 in addition to the requirement for unlocking the previous Masterclasser quest when necessary.

It works as-is, but I'm making it a draft because I have some outstanding questions/I think it could be improved.

Here's how it looks right now:

image

A more concise option would be to just change the message from "To unlock this quest, complete…" to "To unlock this quest, unlock and complete…", instead of displaying the two conditions separately, but that would require adding some new translatable strings, and changing a bit more of the logic in `questPopover.vue`.

### Open questions

Working on this revealed some strange behavior with the way locked quests are handled in general, especially ones with multiple unlock requirements. The approach I took is convenient because it doesn't require adding any new strings and it also doesn't require any significant refactoring, but there are a few reasons I'm not satisfied with it:

1. It shows both parts of the message whenever lostMasterclasser2, 3, or 4 is locked, even if one of the two requirements has been met. This is consistent with the behavior of other quest unlock popovers—for example, a user who has completed Attack of the Mundane Part 1 in a party but hasn't yet reached level 15 will see both the message about needing to complete AtoM 1 and about needing to reach level 15—but it feels confusing to me.
2. It's not clear to the user whether both requirements need to be met, or only one of them. I found an old issue (#10295) implying that users read multiple conditions as an "or" rather than an "and", but the PR that closed that issue actually didn't change anything (I'll open a separate issue about that when I have a minute), and no one seems to have complained about it in the meantime, so maybe most people do understand that it's an "and"?
3. It's just a lot of text to show all at once :(

It seems to me like there's a deeper design question here, which is whether the text in the locked quest popover is meant to be a general statement about the requirements for unlocking the quest, or an instruction informing the user what they still need to do to unlock the quest, taking into account what they've already achieved. Because of the way the logic in `questPopover.vue` is structured, requirements like "log in N times" or "reach level N" disappear as soon as the user satisfies them, but requirements of the "complete another quest" variety are displayed whenever the quest is locked, even if it's only locked for some other reason.

Like I mentioned above, this brings Masterclassers in line with the other quest lines in terms of this behavior, so maybe it's fine, but I think it would be nice to eventually make all the different types of unlock requirements behave consistently.

One other tiny thing to note is that the function `userAbleToStartMasterclasser` now exists in two places, which isn't ideal:

https://github.com/ashaindlin/habitica/blob/5d4f1b868618e36a55ae6c1c5ec0b9ff5cd34bdd/website/common/script/libs/getItemInfo.js#L11
https://github.com/ashaindlin/habitica/blob/5d4f1b868618e36a55ae6c1c5ec0b9ff5cd34bdd/website/common/script/ops/buy/buyQuestGold.js#L17

But the logic itself was already duplicated; all I've done is pull that logic out to a function in `getItemInfo.js`, so I don't think it makes the situation worse. There's probably a way to move this function into the user object itself by putting it in `server/models/user/methods.js`, but I wasn't able to figure out how to make it work (I haven't used Mongoose before) and I'm not sure if it's worth it.

----
UUID: 451e223c-f9b6-4a19-af4b-2a45aba3044d

### DIFF
--- a/website/client/src/components/shops/quests/questPopover.vue
+++ b/website/client/src/components/shops/quests/questPopover.vue
@@ -13,7 +13,7 @@
       {{ item.text }}
     </h4>
     <div
-      v-if="item.locked && item.key === 'lostMasterclasser1'"
+      v-if="item.locked && item.key.startsWith('lostMasterclasser')"
       class="popover-content-text"
     >
       {{ $t('questUnlockLostMasterclasser') }}
--- a/website/common/script/libs/getItemInfo.js
+++ b/website/common/script/libs/getItemInfo.js
@@ -8,8 +8,18 @@ import isPinned from './isPinned';
 import isFreeRebirth from './isFreeRebirth';
 import getOfficialPinnedItems from './getOfficialPinnedItems';
 
+function userAbleToStartMasterclasser (user) {
+  return user.achievements.quests.dilatoryDistress3
+    && user.achievements.quests.mayhemMistiflying3
+    && user.achievements.quests.stoikalmCalamity3
+    && user.achievements.quests.taskwoodsTerror3;
+}
+
 function lockQuest (quest, user) {
-  if (quest.key === 'lostMasterclasser1') return !(user.achievements.quests.dilatoryDistress3 && user.achievements.quests.mayhemMistiflying3 && user.achievements.quests.stoikalmCalamity3 && user.achievements.quests.taskwoodsTerror3);
+  if (quest.key === 'lostMasterclasser1') return !userAbleToStartMasterclasser(user);
+  if (quest.key === 'lostMasterclasser2' || quest.key === 'lostMasterclasser3' || quest.key === 'lostMasterclasser4') {
+    return !(userAbleToStartMasterclasser(user) && user.achievements.quests[quest.previous]);
+  }
   if (quest.lvl && user.stats.lvl < quest.lvl) return true;
   if (quest.unlockCondition && (quest.key === 'moon1' || quest.key === 'moon2' || quest.key === 'moon3')) {
     return user.loginIncentives < quest.unlockCondition.incentiveThreshold;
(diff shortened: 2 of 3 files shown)


SabreCat:
Hi @ashaindlin, are you still planning to finalize this? I think it's OK to have the more verbose requirement text for now, if that's holding you up.

As for `userAbleToStartMasterclasser`, it doesn't necessarily have to be a user method--you could abstract it into a function file similar to something like `website/common/script/libs/hasClass.js` and import it where needed.

SabreCat:
Closing for now, but let us know if you're planning to run with this again!

PR #13383, opened 2021-07-08
fix(cors): Allow all origins to support 3rd party tools

### LINKED ISSUE #12383: CORS Origin errors when using third-party tools from two or more domains - API headers may need Vary: Origin
_I'm labelling this suggestion-discussion for the staff to consider if `Origin` should be added to the `Vary` header (Mozilla docs referenced below)._

---

When a player is using third-party tools from two different domains in the same browser at about the same time, they'll see inappropriate CORS errors.

For example, Ceran (aabe567b-aa55-42c1-b612-b7699a7ba9b7) was using [a third-party tools on oldgods.net](https://oldgods.net/habitica/cTheDragons/challenge.html), then started using [a third-party tool at ieahleen.github.io](https://ieahleen.github.io/habitican-randomizer/) and saw this console message there:

"Access to fetch at 'https://habitica.com/api/v3/content?language=en' from origin 'https://ieahleen.github.io' has been blocked by CORS policy: The 'Access-Control-Allow-Origin' header has a value 'https://oldgods.net' that is not equal to the supplied origin. Have the server send the header with a valid value, or, if an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
script.js:348 GET https://habitica.com/api/v3/content?language=en net::ERR_FAILED"

The same thing happened to me (d904bd62-da08-416b-a816-ba797c9ee265) but in reverse. I used the ieahleen.github.io tool successfully, then went to an oldgods.net tool and saw the CORS policy message saying that access from origin oldgods.net had been blocked because the header value was ieahleen.github.io.

The CORS errors tend to stop happening after one or a few reloads, and they don't always happen, but I've been able to reproduce it more than once. It seems to be because the browser is caching the the origin domain from site and using the cached domain on a different site, when both sites are sending requests to the same destination.

It seems to be a known problem for APIs other than Habitica. For example: https://community.esri.com/thread/238787-how-to-prevent-browser-from-caching-cors-domain (_"... Clearly some caching is happening in the browser, where it's sending the wrong (previous) domain out in the request so the external [destination] server is (correctly) rejecting it..."_). Unfortunately that's just a report of the problem without a solution. :)

---

I _think_ the problem may be fixed by adding `Origin` to the API's Vary header. I believe that may be what's being referred to in the text below, which is from https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMissingAllowOrigin :
**"To allow any site to make CORS requests without using the * wildcard (for example, to enable credentials), your server must read the value of the request's Origin header and use that value to set Access-Control-Allow-Origin, and must also set a Vary: Origin header to indicate that some headers are being set dynamically depending on the origin."**

Habitica does set a non-wildcard origin sometimes as seen in https://github.com/HabitRPG/habitica/blob/develop/website/server/middlewares/cors.js and it sounds like the `Vary` header may help when the browser needs to send requests from third-party tools at two different origins.

I'm still pretty new to CORS headers so I'm not sure this will fix the problem, but regardless of that the Mozilla docs do seem to be saying that `Vary: Origin` is important in general. The examples at https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS have `Vary: Accept-Encoding, Origin` in them, and I think the API is currently sending only `Vary: Accept-Encoding` from what I can see in the headers reported when I use third-party tools.

icco:

Fixes #12383

### Changes

In certain cases, Habitica's servers return incorrect cors headers. This is particularly damaging on 3rd party tools that hit the API from the browser, because it will cause requests to randomly fail. I'm not sure what is causing the origin header to be wrong, as I don't have insight into the routing infra of Habitica. But this should fix the issue for users.

----
UUID: a22efff0-2af3-4bce-8075-e810ea228871

### DIFF
--- a/website/server/middlewares/cors.js
+++ b/website/server/middlewares/cors.js
@@ -1,6 +1,6 @@
 export default function corsMiddleware (req, res, next) {
   res.set({
-    'Access-Control-Allow-Origin': req.header('origin') || '*',
+    'Access-Control-Allow-Origin': '*',
     'Access-Control-Allow-Methods': 'OPTIONS,GET,POST,PUT,HEAD,DELETE',
     'Access-Control-Allow-Headers': 'Authorization,Content-Type,Accept,Content-Encoding,X-Requested-With,x-api-user,x-api-key,x-client',
     // Expose rate limit headers to CORS requests


SabreCat:
Simple and effective! Thanks for this, @icco, and welcome to the ranks of the Blacksmiths at tier 1! Further tiers will require multiple or larger PRs to attain, but keep helping out and we'll express our gratitude accordingly!

Issue #13344, opened 2021-06-19
Task edit dropdown menus for Tags and Difficulty don't close themselves, causing overlap


Alys:
_Below is a report from @fufuj (86f37f0c-ebaf-4ee8-bf9f-00a5aa485501). I've marked this as in progress for them to work on a fix._

On Firefox and Internet Explorer (prolly on other browsers aswell), when you:
    + Click to edit a habit
    + Click on a Tags to expand dropdown list of available Tags,
    + Then, without closing Tags dropdown list, click on Difficulty to expand dropdown list of available Difficulties
    + Tags dropdown list doesn't close, therefore it looks bad when those lists overlaps each other

Also, i can try to fix it, so as soon as you post it on GitHub issues, i will apply to fix it.

Looks like this:

![image](https://user-images.githubusercontent.com/1495809/122635163-a3fa0a00-d125-11eb-9852-e4490217a220.png)

dabrowskif:
Hi, as i wrote in Habitica, i will try to take care of it if possible. :)

dabrowskif:
I've finished fixing this bug, but can't "git push origin name-of-branch" this. Wrote on habitica's aspiring blacksmith guild.

PR #13400, opened 2021-07-19
Fix transform atidote unpin after using rebirthing fixes #13306

### LINKED ISSUE #13306: Orb of Rebirth should unpin Antidotes (Sand, etc) for Transformation Item buffs (Seafoam, etc)
_I'm marking this as "good first issue" and also "minor" priority, so experienced Blacksmiths should leave this issue for a new coder. :)_

If you have had an avatar [Transformation Buff](https://habitica.fandom.com/wiki/Transformation_Items) cast on you, and you then use the Orb of Rebirth, the transformation effect is removed at the same time as the other Buffs (that's the correct behaviour). However Rebirth does not un-pin the antidote to the transformation Buff. 

For example, if you had Seafoam cast on you, the Sand antidote would be automatically pinned in your Rewards section, and then after Rebirth it would still be pinned there.

The expected behaviour is that using Rebirth un-pins any antidotes that are no longer needed.

Below is the code for how cron handles removing all the Buffs and then unpinning the antidotes. Rebirth should do something similar.

https://github.com/HabitRPG/habitica/blob/bc730a0d59676aabaa8bbd55fb198995b7c8ba3a/website/server/libs/cron.js#L508-L511

_Thanks to @clumsybrock (05c4dd65-e87c-448f-981c-5d882efcd60d) for reporting this._

dabrowskif:
fixes #13306

### Changes
Fix transformation buff antidote to properly unpin from rewards shop after using orb of rebirth.
Added the same imported function (  setDebuffPotionItems(user) )to rebirth function as it is in cron.js (new day) file - it checks for transformation debuffs and pin/unpin respective antidotes.

----

### DIFF
--- a/website/client/src/components/tasks/modal-controls/selectMulti.vue
+++ b/website/client/src/components/tasks/modal-controls/selectMulti.vue
@@ -49,7 +49,7 @@ multi<template>
           :key="item.id"
           class="ignore-hide multi-item"
           :class="{ 'none': item.id === 'none', selectListItem: true }"
-          @click.prevent.stop="selectItem(item)"
+          @click="selectItem(item)"
         >
           <div
             v-markdown="item.name"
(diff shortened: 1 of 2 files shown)


dabrowskif:
I think i messed up with doing the first bugfix on the develop branch, that's why my previous commit that fix #13344 issue appears here as well.

SabreCat:
Looks good, @dabrowskif! Added to your contributor credit for tier 2!

PR #13414, opened 2021-07-26
Dark mode input fix (fixes #13312)

### LINKED ISSUE #13312: UI elements taking on system dark mode colours
Created on request from Alys.

### General Info
  * UUID: cbeeb477-d5db-4efd-8f0b-68201870482b
  * Browser: Firefox 57.0
  * OS: Linux Mint 18.3

### Description
Certain UI elements take on dark mode colours when dark mode is set in the system settings. It does seem to be based on system settings as changing Firefox to light mode makes no difference.

It is worst on inputs, as the content is near impossible to see:

http://lilaclynx.net/img/darkmodeinput.png
image

~It also effects elements such as notifications, tooltips, etc, but I am not sure if this is intended behaviour anyway and it doesn't break accessibility like the inputs:~

~http://lilaclynx.net/img/darkmodetooltip.png~
~http://lilaclynx.net/img/darkmodenotification.png~
~http://lilaclynx.net/img/darkmodedrawer.png~

There may be more elements affected, these are just the ones I found. I'd be happy to volunteer to fix the problem for you.

#### Console Errors
None

J-Sanderson:
Fixes #13312

### Changes
Added the form-control class to inputs where it was previously missing. This fixes the issue of inputs taking on dark background colours where the system is set to dark mode, rendering them unreadable and not matching the rest of the site.

----


website/client/src/pages/private-messages.vue
```diff
@@ -421,6 +421,7 @@
     padding-right: 1.5rem;
 
     textarea {
+      background: #fff;
```
SabreCat: Can we use the `$white` variable here instead of hardcoded #fff?

package.json
```diff
@@ -30,6 +30,7 @@
     "express": "^4.16.3",
     "express-basic-auth": "^1.1.5",
     "express-validator": "^5.2.0",
+    "gifsicle": "^5.2.0",
```
SabreCat: Please don't include package.json or package-lock.json files in your PR unless adding a new dependency is necessary to the fix.

SabreCat (changes requested):
Two very minor things, but with that bit of polish this will be ready to go!

SabreCat:
Went ahead and cleaned it up. Thanks for this, and welcome to contributor tier 2!

SabreCat (approved):

PR #13415, opened 2021-07-28
Fix negative quest scroll number on bundle purchase

### LINKED ISSUE #13305: Repair negative quest scrolls on purchase of quest bundles
Over the years, various bug conditions have allowed people to own negative numbers of quest scrolls, which meant that when they bought a new scroll, their total would become less negative, but they still wouldn't own the quest unless they bought enough to get them above zero. We had it recorded at https://github.com/HabitRPG/habitica/issues/2592

Two fixes were done, both of which first checked to see if the current total owned was below zero and if so set it to zero before the purchase was made:
- https://github.com/HabitRPG/habitica/pull/10953 (gold-purchasable quests only)
- https://github.com/HabitRPG/habitica/pull/11399 (gem-purchasable quests)

I suspect that neither of those fixes take effect when purchasing quests via a quest bundle, because we had the issue below reported today in the Report a Bug guild for the Feathered Friends bundle (harpy, owl, falcon). When I checked the affected player's data, they had -8 falcon quest scrolls.

**The desired fix is that when purchasing a quest bundle, before adding each type of quest to the player's inventory, the code should check to see if each type has a negative value and if so should set the value to 0 before proceeding with the purchase.** 
The two PRs linked above will show you the code that was created for the previous fixes.

---

@jaioni (3918a73e-53f4-4a25-b559-360b4c30aafc) reported:

> "Hi! I bought 2 of the current pet quest bundles. I got the quests for Parrots and Owls but none for the Falcons. I'm an android user and I tried checking my inventory in both the app and browser (both mobile and full) but they are nowhere to be found.

### LINKED ISSUE #13306: Orb of Rebirth should unpin Antidotes (Sand, etc) for Transformation Item buffs (Seafoam, etc)

dabrowskif:
fixes #13305

### Changes
Fix negative quest scroll number on quest bundle purchase
It's an additional fix to repair negative quest scroll number when buying quest bundle for gems.
The "if statement" i changed in commit was not checking for bundledKey in the second criterium.
I changed key to bundledKey and now it works properly.

----

### DIFF
--- a/website/common/script/ops/buy/purchase.js
+++ b/website/common/script/ops/buy/purchase.js
@@ -54,7 +54,7 @@ function purchaseItem (user, item, price, type, key) {
   } else if (type === 'bundles') {
     const subType = item.type;
     forEach(item.bundleKeys, bundledKey => {
-      if (!user.items[subType][bundledKey] || user.items[subType][key] < 0) {
+      if (!user.items[subType][bundledKey] || user.items[subType][bundledKey] < 0) {
         user.items[subType][bundledKey] = 0;
       }
       user.items[subType][bundledKey] += 1;


SabreCat:
Thanks for this, @dabrowskif, and welcome to the ranks of the Blacksmiths at tier 1! Further tiers require more or larger PRs to attain, but keep helping out and we'll express our gratitude accordingly!

PR #13449, opened 2021-08-13
Fixes aborted Quest should appear in Quest Inventory without manual sync · Issue #13289

### LINKED ISSUE #13289: aborted Quest should appear in Quest Inventory without manual sync
As reported by @citrusella and confirmed by me:

> If you abort a quest, you won't be able to see it in your inventory (including via the "start a quest" form on the party page) until you manually sync or you reload (or presumably if you do something that causes a full sync to happen).
> For a brief moment I actually thought I lost the quest in question after aborting (which would have been a much worse bug if true), but it was just a visual thing.

The desired behaviour is that as soon as you abort a quest, you should be able to see it in your Inventory or restart it from the Party page without needing to sync manually.

ShiqiRao:

Fixes #13289

### Changes
Refresh the member's data while the inventory page is loading.

----
UUID: 9929ec3f-d3c9-46b1-bedb-be6b234d502a

### DIFF
--- a/website/client/src/components/groups/questActions.mixin.js
+++ b/website/client/src/components/groups/questActions.mixin.js
@@ -53,6 +53,7 @@ export default {
         });
         this.group.quest = quest;
         partyState.data.quest = quest;
+        this.$store.dispatch('party:getMembers', { forceLoad: true });
       } else {
         if (!window.confirm(this.$t('sureCancel'))) {
           return false;


SabreCat:
This doesn't appear to fix the entire problem. Steps to reproduce:

1. In your Party, open the Select Quest form. Note the quest scrolls present.
2. Choose a quest (the issue is clearest with one that has only a single copy) and begin it. Force start if needed.
3. Open "View Details" and abort the quest.
4. Reopen Select Quest. The scroll is still missing.
5. Sync, then reopen Select Quest. The scroll has returned.

Since this is the most common scenario people are going to encounter this issue, I think we should make sure that variation is fixed before merging.

SabreCat (changes requested):
Needs to cover one more workflow from the issue ticket

ShiqiRao:
Hi,I dropped the last commit. And I changed it to sync member data after aborting quest.

SabreCat (approved):

SabreCat:
That worked better! Thanks, and welcome to the ranks of the Blacksmiths at tier 1! Further tiers take multiple or larger PRs to attain, but keep helping out and we'll express our gratitude accordingly!

PR #13495, opened 2021-09-07
remove timer for quests in inventory (fixes #13224)

### LINKED ISSUE #13224: Time-limited quests show the timer in the inventory as well
As Fox_town reported in Report a Bug:

"The seasonal quests (e.g. Waffling with the Fool) have an expiration timer on the hover-over, in the inventory. (It's the same hover-over as when the quest is in the shop.) This is misleading and has unduly led some players to think the scroll will vanish from their inventory if they do not use it by tomorrow. Especially as these are gem purchases, I think we should remove that alarm from the purchased version of the scrolls."

Remeday adds:

"In addition to what Foxtown said, the Find the Cub and Trapper Santa quest scrolls are showing in my inventory the same way, though they're not currently available in the Market at all. That seems even more confusing, since there's not even a "Oh, that just means available for purchase" explanation."

Lyqst:
Fixes #13224 

### Changes
Removes the timer in the quest popup for quests in the inventory.

Before:
inventory-questtimer before
After:
inventory-questtimer after

----

### DIFF
--- a/website/client/src/components/inventory/items/index.vue
+++ b/website/client/src/components/inventory/items/index.vue
@@ -194,7 +194,7 @@
                   <h4 class="popover-content-title">
                     {{ context.item.text }}
                   </h4>
-                  <questInfo :quest="context.item" />
+                  <questInfo :quest="context.item" :purchased="true" />
                 </div>
                 <div v-else>
                   <h4 class="popover-content-title">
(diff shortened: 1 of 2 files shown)


SabreCat:
Thanks for this, @Lyqst, and welcome to the ranks of the Blacksmiths at tier 1! Note that while we award a tier for a first PR of any size, further tiers require more or larger PRs to attain. But keep helping out and we'll show our appreciation accordingly!

Lyqst:
Happy to help!!

PR #13506, opened 2021-09-14
fix cloned challenge task order (#13328)

### LINKED ISSUE #13328: cloned challenges should keep tasks in the same order as the original
As noted at https://github.com/HabitRPG/habitica/issues/1703#issuecomment-856953339 (comment copied below) when you clone a challenge, the tasks sometimes do not keep the same order they were in in the original challenge.

 I reproduced this using https://habitica.com/challenges/79fabb8c-f99a-459f-bdb7-ed810f25661b - the first clone had the tasks in the correct order, but the clone of that clone had the Habits mixed up.
 
 As you can see from the report below, other kinds of tasks can have the same issue.
 
 If anyone would like their own copy of the https://habitica.com/challenges/79fabb8c-f99a-459f-bdb7-ed810f25661b challenge so that you don't have to recreate it manually, post your User ID (not Username) here and I'll use my moderator abilities to clone it and make you the owner of the clone.
 
 ---
 
Original report from https://github.com/HabitRPG/habitica/issues/1703#issuecomment-856953339 by  @habitrpguser :

This bug still happens! I don't know if challenge tasks change order if a challenge is cloned once (I couldn't reproduce that), but they definitely do if a challenge is cloned and then that clone is cloned again. Please see examples:

https://habitica.com/challenges/79fabb8c-f99a-459f-bdb7-ed810f25661b

https://habitica.com/challenges/7f96108c-c086-4ae1-9fa4-7b2006bbb965

https://habitica.com/challenges/6cec332e-8ddc-46f7-afe6-f2a9cb212311

Lyqst:
Fixes #13328

### Changes
https://github.com/HabitRPG/habitica/issues/13328#issuecomment-918588649 :
> I think I've fixed the issue. The problem was that when querying for the tasks to be added to the new challenge, it would just search by the challenge id, which I don't think would necessarily return the tasks in the expected order:
> 
> ```js
> const challengeTasks = await Tasks.Task.find({
>   'challenge.id': challengeToClone._id,
>   userId: { $exists: false }, 
> }).exec();
> ```
> 
> I've changed this to instead iterate over the tasksOrder arrays in the challenge, so we get the tasks in the expected order:
> 
> ```js
> const challengeTaskIds = [
>   ...challengeToClone.tasksOrder.habits,
>   ...challengeToClone.tasksOrder.dailys,
>   ...challengeToClone.tasksOrder.todos,
>   ...challengeToClone.tasksOrder.rewards
> ];
> 
> const challengeTasks = await Promise.all(challengeTaskIds.map(async taskId => {
>   const task = Tasks.Task.findById(taskId).exec();
>   return task;
> }));
> 
> challengeTasks.reverse(); // last task should be added first and vice-versa, since new tasks are prepended
> ```
> 
> I've tested this and it looks to be working, but please let me know what you think before I create the PR.

----

### DIFF
--- a/website/server/controllers/api-v3/challenges.js
+++ b/website/server/controllers/api-v3/challenges.js
@@ -860,10 +860,19 @@ api.cloneChallenge = {
 
     const { savedChal } = await createChallenge(user, req, res);
 
-    const challengeTasks = await Tasks.Task.find({
-      'challenge.id': challengeToClone._id,
-      userId: { $exists: false },
-    }).exec();
+    const challengeTaskIds = [
+      ...challengeToClone.tasksOrder.habits,
+      ...challengeToClone.tasksOrder.dailys,
+      ...challengeToClone.tasksOrder.todos,
+      ...challengeToClone.tasksOrder.rewards
+    ];
+
+    const challengeTasks = await Promise.all(challengeTaskIds.map(async taskId => {
+      const task = Tasks.Task.findById(taskId).exec();
+      return task;
+    }));
+    
+    challengeTasks.reverse(); // last task should be added first and vice-versa, since new tasks are prepended
 
     const tasksToClone = challengeTasks.map(task => {
       const clonedTask = cloneDeep(task.toObject());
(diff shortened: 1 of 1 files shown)


SabreCat:
Looks good to me! Thanks for this--noted toward your second contributor tier! (also I gave you the Blacksmith title, which I think was missing before!)

Lyqst:
@SabreCat I think there's a bug going on with the title, currently says "Friend undefined" (that sounds cool and mysterious though!)

SabreCat:
@Lyqst Oh, is that still happening? I wonder if I initially missed adding Blacksmith, but it's there now!

Lyqst:
It was fixed by @Alys! ;)

PR #13507, opened 2021-09-14
remove sample forms from apidoc - fixes #13012

### LINKED ISSUE #13012: apidocs contain non-working Sample Request forms
~~This issue is suggestion-discussion for staff to comment on which option(s) they'd like to see done. If there's no objections in a week, I'll mark it as help wanted for option 1, or if that's not possible option 2, with my idea being that the contributor who works on it would do the researched needed to work out if option 1 can be done.~~

**EDIT 2021-03-14: Option 2 has been done by @SpeedyLom.  I've set this back to suggestion-discussion. In future, we can look at doing one of the other options if the apidocs package is modified to make it easy to turn off the Sample Request forms or to make it easy to add API authentication fields.** 
 
EDIT 2021-02-07: I've marked this as "help wanted" for for option 1 to be done, or if that's not possible then option 2. The contributor who works on this would do the researched needed to work out if option 1 can be done. Please post a comment here if anyone has any questions about this issue!

---

Habitica's [apidocs](https://habitica.com/apidoc/) now have "Send a Sample Request" form under each route (example screenshot below). Almost all of them don't work because they don't contain fields for the User ID and API Token (and possibly for other reasons, such as other missing fields).
This means they're just unnecessary clutter, will be confusing and frustrating to any Comrades or Blacksmiths who try to use them, and will probably result in questions and bug reports that we'll need to handle.

This issue seems to have arisen due to an intentional change in the https://github.com/apidoc/apidoc/ library. With a quick search, I can't find anything in our own PRs that looks related. It's a great change for that library but our own docs aren't set up to take advantage of it.

There's multiple ways we could fix this:
1. Disable the sample forms, e.g., if the apidoc library gives an easy option for that. This is my favourite approach, at least initially to get a quick fix done to get rid of the non-working forms.
1. If option 1 can't be done quickly, then add a large obvious note to the top of the apidoc page to say that the form won't work. This is my second-favourite option at least initially. It may help to prevent some confusion/frustration/questions.
1. Find some way to add User ID and API Token fields to the top of the apidocs page so that they'll apply to all the forms on the page. E.g., the apidoc library may have a feature to easily enable this.
1. Add the User ID and API Token fields to all of the routes (except the very few that don't need them), so that each sample form works. This is my least favourite approach because it's extra clutter in the docs themselves - option 3 would be better if possible as a long-term solution.
1. Anything else?

Note that options 3 and 4 would also ideally require that we carefully check all the api documentation to ensure that all relevant fields are listed, and listed correctly. This is one of the reasons why I'm suggesting that option 1 be done now as a quick fix. Option 3 or 4 could then be done later.

_This issue is suggestion-discussion for staff to comment on which option(s) they'd like to see done. If there's no objections in a week, I'll mark it as help wanted for option 1, or if that's not possible option 2, with my idea being that the contributor who works on it would do the researched needed to work out if option 1 can be done._

I'll link this issue to the [Aspiring Comrades](https://habitica.com/groups/guild/2ff9822b-27f2-4774-98da-db349b57a38e) and [Aspiring Blacksmiths](https://habitica.com/groups/guild/68d4a01e-db97-4786-8ee3-05d612c5af6f) guilds so that the current members know to not bother trying the forms, and I'll add a note to the wiki docs. 

Thanks to citrusella for noticing and reporting this! I've copied her post below.

image

---

@citrusella wrote:

"A bit of a weird one, but: the [API doc](https://habitica.com/apidoc) now has sample request forms underneath most of the doc options... but it looks like most of them don't actually work (namely the like 99% that require authentication... though weirdly one I thought requires it actually DOES work...), in the sense that they will return a NotAuthorized error.

"It seems as if the sample request forms came about via an update to the dependency that produces the API doc rather than something specifically done to get them to show up. I don't know enough about API requests *or* the way those forms work to know if there's any way to send an authentication header through them at present (or to allow them to send an authentication header themselves)... or if those were previously available in terms of what the thing that generates the doc offers but were somehow suppressed or something... or what. But I figured I'd bring it up because right now for most of them it's kind of a useless form taking up space beneath each entry?"

NicolasCARPi:
### Changes
Remove the sample forms from api documentation, and remove warning about it.

### Side notes

Hello, I am the lead dev of apidoc, taking over the project after its initial author, and I've made **a lot** of changes in the upcoming 0.50.0 version. I'm trying out projects using apidoc with the new version to check if all is good. For instance your script is passing a string in "src" but it was expecting an array, this is now fixed :)

Anyway, this little change is just to remove the warning and sample forms. You should know that your doc builds fine with the newest version (that you can try with: `npm i apidoc/apidoc#master`).

This is what your doc looks like with the latest (unreleased) apidoc version:
2021-09-14-161145_1906x1031_scrot

I'm still ironing out the last bugs before making a release, but it should work fine for you (especially since you don't use sample requests). So feel free to use the version from master if you wish to freshen up your doc ;) Also it removes **many** old dependencies.

The second commit is useful for working with 0.50.0 and doesn't hurt previous versions. Suggestion: put apidoc.json into apidoc folder and make it output html into the same apidoc folder, this way you have all in one place ;) ).


SabreCat:
Oh cool! Thank you for stopping by! If you want to implement this suggestion in the PR

>Suggestion: put apidoc.json into apidoc folder and make it output html into the same apidoc folder, this way you have all in one place ;) ).

we welcome your expertise!

NicolasCARPi:
yeah give me a few minutes I'll add that change then!

NicolasCARPi:
Done. Just remember to change the folder path on your server to point to `apidoc/html` instead of `apidoc_build` ;)

PR #13509, opened 2021-09-15
Fix casting items on parties that exceed it limit by showing Load More Button

### LINKED ISSUE #13394: can't use Avatar Transformation Buffs (e.g., Seafoam) on all party members for parties over 30
If your party has more than 30 members (e.g., due to a bug or because it was created before the party member limit), then you can use avatar transformation buffs such as Seafoam on only 30 of the members. There's two aspects to the bug:
- When you go to Inventory > Items and click on the avatar transformation item, you see the "Select a Party Member" modal and it contains only 30 members. There is no "Load More" button at the bottom for you to see the other party member(s).
- When you enter a player's name in the search form of that modal, it will show you that player if they happen to be one of the 30 in the list. It won't show you the player if they are not in the list of 30. This means you can't cast the buff on anyone who doesn't appear in the list.

This bug also happens in the Android app (we have an emailed report from User ID: 1f472d64-139c-4d34-832e-6b92e448c609).

dabrowskif:
fixes #13394

### Changes

This PR contains 4 commits.

1. st commit: **Fix party size and notification when inviting**
- [ ] Which files: `website/server/models/group.js` && `website/common/script/constants.js`
- [ ] Description: During development i noticed that party limit is actually 31, not 30, as it supposed to be, because party leader wasn't included previously. Code is pretty self-explanatory

2. nd commit: **Fix View Party button in header**

- [ ] Which files: `website/client/src/components/header/index.vue`
- [ ] Description: During development i noticed that when clicking View Party button in header, it doesn't show Load More button.
When clicking Member List (with this shield icon, in Party tab) it worked. I copied some code from `website/client/src/components/groups/group.vue` and modified it, so it doesn't conflict with current code.

3. rd commit:  **Fixed View Party button to properly open party** 
- [ ] Which files: `website/client/src/components/header/index.vue`
- [ ] Description: It was just a small fix to 2nd commit, as i didn't notice that after refreshing the page, View Party button doesn't work.

4. th commit: **Fixed SelectMembersModal to properly show Load More button** 
- [ ] Which files: ` website/client/src/components/selectMembersModal.vue`
- [ ] Description: It was the main fix for the issue with an id 13394 (clickable number is on top of this file). First of all i copied Load More button from previous commits, then i copied some code from previous commits as well. The main thing was to get `group` from dispatch, get some values out of it and save them to the store. I also added `const invite` and assigned it to dispatched `getGroupInvites`, but i wasn't sure what was this fragment of code doing and couldn't find it out, so i just left it here. (lines 269-273)

----

### DIFF
--- a/website/common/script/constants.js
+++ b/website/common/script/constants.js
@@ -28,7 +28,7 @@ export const SUPPORTED_SOCIAL_NETWORKS = [
 
 export const GUILDS_PER_PAGE = 30; // number of guilds to return per page when using pagination
 
-export const PARTY_LIMIT_MEMBERS = 30;
+export const PARTY_LIMIT_MEMBERS = 29;
 
 export const MINIMUM_PASSWORD_LENGTH = 8;
 
--- a/test/api/v3/integration/groups/POST-groups_invite.test.js
+++ b/test/api/v3/integration/groups/POST-groups_invite.test.js
@@ -7,7 +7,7 @@ import {
 } from '../../../../helpers/api-integration/v3';
 
 const INVITES_LIMIT = 100;
-const PARTY_LIMIT_MEMBERS = 30;
+const PARTY_LIMIT_MEMBERS = 29;
 const MAX_EMAIL_INVITES_BY_USER = 200;
 
 describe('Post /groups/:groupId/invite', () => {
--- a/test/api/v3/integration/groups/POST-groups_invite.test.js
+++ b/test/api/v3/integration/groups/POST-groups_invite.test.js
@@ -650,7 +650,7 @@ describe('Post /groups/:groupId/invite', () => {
         .to.eventually.be.rejected.and.eql({
           code: 400,
           error: 'BadRequest',
-          message: t('partyExceedsMembersLimit', { maxMembersParty: PARTY_LIMIT_MEMBERS }),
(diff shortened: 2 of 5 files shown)


dabrowskif:
Also the search field in Select Members modal is completely bugged (and it was before), but I just have no time to fix this.

website/common/script/constants.js
```diff
@@ -28,7 +28,7 @@ export const SUPPORTED_SOCIAL_NETWORKS = [
 
 export const GUILDS_PER_PAGE = 30; // number of guilds to return per page when using pagination
 
-export const PARTY_LIMIT_MEMBERS = 30;
+export const PARTY_LIMIT_MEMBERS = 29;
```
SabreCat, line 31: This is causing an integration test to fail--I understand from the PR description what this is for, but you'll need to update the expectation in the test suite!

SabreCat (changes requested):
Looking pretty good! Just need to fix the broken test.

website/common/script/constants.js
SabreCat: Went ahead and updated the tests!

SabreCat:
Thanks for this, @dabrowskif! Added to your contributor credit toward Tier 3 Blacksmith!

PR #13530, opened 2021-10-03
Feature: new "report a bug" modal

negue:
This will replace the current way bug reports are handled.

Internally an email template will be used to make everything easier.

- [x] Added Tests / checks for the `sendTxn`
- [ ] Added Tests for `/api/v4/bug-report`


website/client/src/libs/consts.js
```diff
@@ -0,0 +1,4 @@
+export const PAGES = {
+  PRIVATE_MESSAGES: '/private-messages',
+  BUG_REPORT: '/bug-report',
```
negue: will be removed again

website/server/controllers/api-v4/bug-report.js
```diff
@@ -0,0 +1,81 @@
+import nconf from 'nconf';
+import { authWithHeaders } from '../../middlewares/auth';
+import { convertVariableObjectToArray, sendTxn } from '../../libs/email';
+
+const api = {};
+
+/**
+ * @api {post} /api/v4/bug-report Report an issue
```
negue, line 7: todo test

website/client/src/app.vue
```diff
@@ -177,6 +179,10 @@ import {
   removeLocalSetting,
 } from '@/libs/userlocalManager';
 
+const bugReportModal = () => import(/* webpackChunkName: "bug-report-modal" */'@/components/bugReportModal');
```
negue, line 182: I think this should be used in more modals - so that the intial load isn't to huge due to all the other modals (and surely other stuff as well ^^)

website/server/controllers/api-v4/bug-report.js
negue: I tried an api-v4 integration but I can't get the `sendTxn` to "work" or I have no idea on how to check anything there (since the its not in the same process) 

 but I think this needs to be an unit test ... or atleast I'm trying it

website/client/src/components/bugReportModal.vue
```diff
@@ -0,0 +1,243 @@
...
+
+    if (!email && user.auth?.google?.emails) {
+      email = user.auth.google.emails?.[0]?.value;
+    }
+
+    if (!email && user.auth?.apple?.emails) {
+      email = user.auth.apple.emails?.[0]?.value;
+    }
+
+    this.email = email;
```
SabreCat, line 240: What does this do if we don't find an email at all? I guess the form will go in but it says `undefined`?

website/client/src/components/bugReportSuccessModal.vue
```diff
@@ -0,0 +1,154 @@
...
+
+.text {
+  min-height: 0;
+}
+
+</style>
+
+<script>
+import closeIcon from '@/components/shared/closeIcon';
+import checkCircleIcon from '@/assets/svg/check_circle.svg';
```
SabreCat, line 131: In website/client/src/components/payemnts/successModal.vue we just use the regular check SVG and surround it with a circle using CSS--do you have a feeling about which way is preferable?

SabreCat (approved):
A couple of questions, but nothing that should hold this up really!

website/client/src/components/bugReportModal.vue
negue: when its `undefined` (value) the field will be just empty - since we don't use any `undefined +  ''` to make it `undefined` (string)

website/client/src/components/bugReportSuccessModal.vue
negue: I think when the icon would need a bigger padding between the icon and the circle then the raw icon + CSS is the way to go 

if its all combined in an svg (and is the one we need to use) then I'd go with the svg way

PR #13556, opened 2021-10-11
FIX: notifications position with 2nd banner - WIP

negue:
Now it reacts when a banner is visible / shown.

Only issue currently is: the position only updates when you scrolled at least once, guess somehow only then the currentState is updated?


SabreCat (approved):
Yay, looks great!

PR #13557, opened 2021-10-14
Api task streak default value

### LINKED ISSUE #13372: Set streak to 0 if it has no integer value
If someone edits their streak on web and saved the task with an empty field, that task will disappear on mobile. If they go back and type a 0 into the field, then the task appears normally again. We'd like to change the API so all platforms will always save the field as 0 if its empty to make sure this is fixed across all platforms for good.

original issue was posted to mobile, but we decided to make it an API change. Can reference that issue here: https://github.com/HabitRPG/habitica-android/issues/1569

Lyqst:
Fixes #13372

### Changes
Added default value to task streak (0) when no value is entered. Also, added truncation to ensure integer value.

----

### DIFF
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -688,6 +688,7 @@ api.updateTask = {
           -user.preferences.timezoneOffset,
         ).date()];
       }
+      if (task.streak === undefined) task.streak = 0;
     }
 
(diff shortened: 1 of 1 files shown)


SabreCat (approved):

SabreCat:
Looks good! Thanks for this, and welcome to contributor tier 2!

PR #13559, opened 2021-10-16
Prevent Cron from incorrectly updating loggedIn

### LINKED ISSUE #12914: /api/v3/cron POST route sets auth.timestamps.loggedin even when cron doesn't run
~~This is suggestion-discussion for us to consider the code implications and whether we want to fix this.~~

If you have already cronned since your CDS time today and you run the `/api/v3/cron` POST route (e.g., with the command below), then the value for `auth.timestamps.loggedin` is updated (as is `lastCron` but that's no as much of an issue). The cron route doesn't make any other changes (which is as expected and as desired).

The `loggedin` value normally indicates the most recent time that a true cron ran, so the value is misleading when it's set by that route at a time when cron doesn't actually run.

@cTheDragons reported this and pointed out that it can lead to incorrect stats being reported by third-party tools. E.g., any tasks that were completed after today's real cron but before the new `loggedin` time may be displayed as if they were completed before a real cron.

sample command:
`curl -X POST -H "x-api-user: 12345678-3d8a-4a10-bfe0-123456789abc" -H "x-api-key: 12345678-3d8a-4a10-bfe0-123456789abc"  https://habitica.com/api/v3/cron`

DarknessCatt:
Fixes #12914

### Changes
Changed where `auth.timestamps.loggedin` is set so it is only updated in the end of Cron (and if not aborted). Followed @SabreCat suggestions (thanks for that!) in the original post and it's working as expected.

It still updates the value from if the user hasn't logged in when the endpoint is triggered (as in, the endpoint is called before the user logging in the current day), but I believe that's expected.

----
UUID: 8c0d0692-ce4f-4b08-8737-4599533807c2


SabreCat (approved):

SabreCat:
I fixed one linting issue, but with that corrected it looks good!

Thanks for this, @DarknessCatt, and welcome to the ranks of the Blacksmiths at tier 1! Further tiers require multiple or larger PRs to attain, but keep helping out and we'll express our gratitude accordingly!

PR #13581, opened 2021-10-20
Fix chat @-mentions not opening in a modal

### LINKED ISSUE #13276: Profile links generated from @-mentions open in new tab

### Description

If you mention a user with the @ feature in chat, the resulting link opens the standalone profile page in a new tab, instead of the modal profile view you get when e.g. clicking a user's name at the top of a chat message they posted. We should be continuing to leverage the modal in this case!

Originally reported in the bug guild by @citrusella!

Lyqst:
Fixes #13276

### Changes
Extract the relative link using pathname property to correctly add the event listener to open the modal on mention links (since those are absolute links).

----

### DIFF
--- a/website/client/src/components/chat/chatCard.vue
+++ b/website/client/src/components/chat/chatCard.vue
@@ -285,7 +285,7 @@ export default {
       let link = links[i].pathname;
 
       // Internet Explorer does not provide the leading slash character in the pathname
-      link = link.charAt(0) === '/' ? link : '/' + link;
+      link = link.charAt(0) === '/' ? link : `/${link}`;
 
       if (link.startsWith('/profile/')) {
         links[i].onclick = ev => {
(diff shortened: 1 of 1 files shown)


SabreCat:
Looks great! Thanks for this! Credited toward your third tier.

SabreCat (approved):

citrusella:
So I noticed this behavior now existing on the delta site and I *also* noticed something else that's not exactly an @ mention... linking manually to a profile will also produce modal behavior instead new tab behavior when it didn't previously, I don't believe.

I think that's cool (and it's still possible to open in new tab just fine using a right click), I just wanted to comment on it in case it was something that wasn't known or something undesired.

SabreCat:
I think we're fine with that, it's a nice consistent experience, but thanks for pointing it out @citrusella!

citrusella:
Another thing I noticed that might be related to this but also might be just how things work, I dunno: @ mentions of a username in a private message still open in a new tab and don't open in a modal.

Threw me off enough I wanted to mention it even though this is merged and (I think?) deployed.

SabreCat:
@citrusella Hmm, that's interesting--it's possible that's an oversight that we could also fix. You can reopen #13276 with a comment to that effect!

PR #13582, opened 2021-10-20
Frost spell buff won't affect challenge or group dailies

### LINKED ISSUE #7809: Chilling Frost skill freezes the streaks on challenge Dailies
Chilling Frost freezes streaks on Challenge Dailies.

<strike>This is consistent with being able to edit the streaks on Dailies (necessary for recovering from unexpected crons), but not consistent with the policy that skills can't be used to affect challenge tasks. Perhaps we didn't change this when API v3 was being built due to just an oversight (I didn't think of it until cTheDragons asked about it). I'm pretty much on the fence about whether it should affect challenge tasks or not - perhaps slightly on the side of not affecting them. What do others think?</strike>

The skill should be changed so that it does NOT freeze the streak for Challenge Dailies.

Lyqst:
Fixes #7809

### Changes
Chilling frost applies a buff that is only checked in the scoreTask function in the line
```javascript
if (!user.stats.buffs.streaks) task.streak = 0;
```
Changed this to
```javascript
if (!user.stats.buffs.streaks || task.challenge.id || task.group.id) task.streak = 0;
```
So now the buff won't apply to challenge or groups dailies.

----

### DIFF
--- a/website/common/script/ops/scoreTask.js
+++ b/website/common/script/ops/scoreTask.js
@@ -286,7 +286,8 @@ export default function scoreTask (options = {}, req = {}, analytics) {
     if (cron) {
       delta += _changeTaskValue(user, task, direction, times, cron);
       _subtractPoints(user, task, stats, delta);
-      if (!user.stats.buffs.streaks) task.streak = 0;
+      // Chilling frost should not affect challenge or group dailies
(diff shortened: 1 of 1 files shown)


SabreCat:
Straightforward and effective! Thanks for this--added to your credits for next contributor tier!

SabreCat (approved):

PR #13589, opened 2021-10-26
Add Transaction log for gem and hourglass changes

phillipthelen:
This adds the following things:

- All changes to the amount of gems or hourglasses a user has are tracked in a `Transaction` model.
- In the users settings, there is a newt Tab "Transactions". This tab contains a list of all changes that happened to both their gems and hourglasses (this tab is currently hidden for non-admin users until we are happy with how everything works. The view itself could be opened by non-admins by typing in the URL directly, since I did not want to go through too much trouble of restricting access too it and it is not actually an issue if users are able to see it.)
- When editing a user from the hall of heroes, there is a new expandable section "Transactions", that contains the same table that is displayed in the settings.


test/api/v4/members/GET-purchase_history.test.js
```diff
@@ -0,0 +1,38 @@
...
+    const member = await generateUser();
+    const nonAdmin = await generateUser();
+    await expect(nonAdmin.get(`/members/${member._id}/purchase-history`)).to.eventually.be.rejected.and.eql({
+      code: 401,
+      error: 'NotAuthorized',
+      message: t('noAdminAccess'),
+    });
+  });
+
+  it('returns achievements based on given user', async () => {
```
SabreCat: achievements?

test/common/ops/changeClass.js
```diff
@@ -19,51 +19,48 @@ describe('shared.ops.changeClass', () => {
     user.stats.flagSelected = false;
   });
 
-  it('user is not level 10', done => {
+  it('user is not level 10', async () => {
     user.stats.lvl = 9;
     try {
-      changeClass(user, { query: { class: 'rogue' } });
+      await await changeClass(user, { query: { class: 'rogue' } });
```
SabreCat, line 25: double await?

test/common/ops/releaseMounts.js
```diff
@@ -23,87 +23,85 @@ describe('shared.ops.releaseMounts', () => {
...
   });
 
-  it('returns an error when user does not have all pets', done => {
+  it('returns an error when user does not have all pets', async () => {
     const mountsKeys = Object.keys(user.items.mounts);
     delete user.items.mounts[mountsKeys[0]];
 
     try {
-      releaseMounts(user);
+      await await releaseMounts(user);
```
SabreCat: 'nother double await. Might want to do a quick `grep` on this pattern, if it's not intentional

website/client/src/components/hall/heroes.vue
```diff
@@ -384,11 +421,24 @@ export default {
       this.heroes[this.currentHeroIndex] = heroUpdated;
       this.currentHeroIndex = -1;
     },
+    async clearHero () {
```
SabreCat: doesn't await anything, does this need to be async?

website/client/src/components/ui/purchaseHistoryTable.vue
```diff
@@ -0,0 +1,156 @@
...
+          <td>
+            <span v-html="entryReferenceText(entry)"></span>
+          </td>
+        </tr>
+      </table>
+    </div>
+  </div>
+</template>
+
+<style lang="scss">
```
SabreCat, line 72: Indentation looks weird here, should be 2 spaces per level in our code style

website/common/locales/en/settings.json
```diff
@@ -184,5 +184,25 @@
...
     "everywhere": "Everywhere",
     "onlyPrivateSpaces": "Only in private spaces",
-    "bannedSlurUsedInProfile": "Your Display Name or About text contained a slur, and your chat privileges have been revoked."
+    "bannedSlurUsedInProfile": "Your Display Name or About text contained a slur, and your chat privileges have been revoked.",
+    "transactions": "Transactions",
+    "gemTransactions": "Gem Transactions",
+    "hourglassTransactions": "Hourglass Transactions",
+    "noGemTransactions": "You don't have any gem transactions yet.",
+    "noHourglassTransactions": "You don't have any hourglass transactions yet.",
+    "transaction_debug": "Debug Action",
```
SabreCat, line 193: Do these strings even need to be translated, given they don't face the end user?

SabreCat (changes requested):
Minor nitpicky stuff and questions! Looking good!

website/common/locales/en/settings.json
phillipthelen: The `transaction_debug` one might not be, that's true. The other ones are user facing though in the transaction table when you don't have any transactions yet as an empty state.

phillipthelen: Although the debug action one is technically also user facing, since it is also shown in the UI. (although this would only happen on local dev instances)

SabreCat: kk!

SabreCat (approved):

PR #13591, opened 2021-10-31
Feature/warn missing config

jbryson3:
### Changes

#### TLDR
- Add logger.warn method for messages that may need attention but may not be errors
- Add a warning when starting the local server without a `config.json` present

#### Background and Solution
As a new developer to the project, I read the install instructions carefully the first time (cloning the lib). The second time I did the install process (forking) I forgot the step to copy `config.json.example` to `config.json`. I got an unhelpful error when starting up. 

image

I would have appreciated a bit of help to find out what was wrong, so I put in a helpful warning log. Hopefully it should make getting started developing the project just a tiny bit easier.

image

I had to expose a `warn` method in the logger to elevate it from `info`, but I don't think its an `error` as it looks like there are different ways to boot up the app without the use of `config.json`

----
UUID: e22287a3-189f-4acc-aa0a-36bb30fd4664

### DIFF
--- a/website/server/libs/logger.js
+++ b/website/server/libs/logger.js
@@ -120,7 +120,7 @@ if (IS_PROD) {
   _config.loggingEnabled = false;
 }
 
-// exports a public interface insteaf of accessing directly the logger module
+// exports a public interface instead of accessing directly the logger module
 const loggerInterface = {
   info (...args) {
     if (!_config.loggingEnabled) return;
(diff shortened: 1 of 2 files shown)


SabreCat:
Hmm, it looks like the logger changes are causing a bunch of tests to fail? Can those messages be skipped in a testing environment?

SabreCat:
Closing as abandoned.

PR #13679, opened 2021-12-07
FIX: refactor notifications position based on different promo banners

negue:
extending the previous work of #13556  to use both types of promo banners

### DIFF
--- a/website/client/src/components/snackbars/notifications.vue
+++ b/website/client/src/components/snackbars/notifications.vue
@@ -103,7 +103,7 @@ export default {
       removalIntervalId: null,
       notificationTopY: '0px',
       preventMultipleWatchExecution: false,
-      gemsPromoBannerHeight: null,
+      eventPromoBannerHeight: null,
       sleepingBannerHeight: null,
     };
   },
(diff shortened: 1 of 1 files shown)


website/client/src/components/snackbars/notifications.vue
```diff
@@ -355,10 +356,31 @@ export default {
...
     },
+
+    updateEventBannerHeight () {
+      if (this.isEventActive) {
+        this.eventPromoBannerHeight = getBannerHeight(this.currentEventBannerName());
+      }
+    },
+
+    currentEventBannerName () {
+      // if there are any other types of promo bars
```
negue, line 371: @SabreCat is this here a good place or should this be moved to the events.js ? Then a second export would be to be changed if an event has a promo banner

SabreCat: Hmm, I feel like the best thing would be to always use the `promo` field in `events.js`, with either `gems` or `g1g1` so far, and just read from that..?

SabreCat (approved):

website/client/src/components/snackbars/notifications.vue
negue: I would've used it - but - the value inside promo is a different one that the banners component name (used in the template) 

if the promo and the name in the component would be the same then we could just use it yes

PR #13698, opened 2021-12-13
Fixed intermittent issues with enter key in chat

### LINKED ISSUE #12397: Intermittent issues with creating a new line in chat - Enter key not working
AnneLine reports:

> I've been noticing an intermittent issue few times on web browser version (Windows 10 and chrome right now, but also on Windows 8 and chrome in recent weeks) where when I start to do a new paragraph in chat, the window will NOT allow me to do a 'return' to a new line. Its not consistent, did it a few moments ago abut now it is fine. Solution seems to be a cut and paste of what was typed, a refresh, and then I am good to go. I'm wondering if I'm the only one experiencing this, or if it is a known issue. Anything else I can do to help you identify it? Happened about 5 minutes ago if that is helpful for you, but not an issue in THIS chat window. Weird.

Ceran corroborates:

> I've had that too a few times! I thought it was an issue on my side for some weird reason and not happening to others. Also not being able to make a new line with hitting "enter", or not selecting a user from the dropdown menu when @ing. Refreshing is the only thing that works then.

Nakonana added some info:

> I've encountered the same problem as Ceran and AnneLine using Firefox on Win 10, in case it's relevant. iirc it usually happened when I wrote my reply in a word document and then pasted it into the Habitica text box (but not 100% sure).

Though it's worth noting that it happens to AnneLine while composing in the text box, so it isn't always due to pasting in. It may be related to #11516, but Ceran confirms it can't always be due to that.

maryketa:

Fixes #12397

### Changes
Changed enter key functionality when username autocomplete menu is open in the chat. 
If a username is being hovered over, enter key selects that username. If no username is currently hovered over, enter creates a newline. This fixes the issue where the enter key sometimes does not work in the chatbox, specifically when the autocomplete dropdown menu is active.

----
UUID: f540999b-468c-476d-8773-4c3affd7905b


SabreCat:
Merged in f01c352 !

Thanks for this, @maryketa, and welcome to the ranks of the Blacksmiths at tier 1! Further tiers require multiple PRs or a larger PR to attain, but keep helping out and we'll express our gratitude accordingly!

PR #13700, opened 2021-12-15
[WIP] New Stuff posting time line

### LINKED ISSUE #12920: date on New Stuff by Bailey news posts varies based on time zone / locale
@SnowCrasher and @beinginhisshadow and @citrusella all reported that the "2021 IS HERE!" / January Mystery Items Bailey was dated 31 December 2020.  @citrusella's time zone is UTC-5 and her screenshot is the first one below.

@SnowCrasher's post about this was in the Wizards of the Wiki guild since it was related to how the news is to be displayed on the wiki:
_"I'm confused. Today's news says 12/31/2020 but it was posted today, and the wiki is edited for today as well. Which is it?"_

When I look at the Bailey I see it dated 01/01/2021. My timezone is UTC+10 and my screenshot is the second one below.

_I'm marking this as suggestion-discussion for Habitica's staff to consider if / how this should be fixed._

image

image

Lyqst:
Fixes #12920
### Changes
I added the line as shown below:
image

As @citrusella mentioned here https://github.com/HabitRPG/habitica/issues/12920#issuecomment-960312085:

> ...this doesn't specify daylight or standard time, does that mean it should always be EST (as it is in the code Lyqst is working on to my understanding)? Or does it matter when Eastern is in daylight time (especially as all of Eastern observes it AFAIK, hrmmmm)? (And then, does it matter that many people tend to use EST as "the time offset I'm at right now" (i.e. EST = EDT even if that means UTC -4 which is not what EST actually is) vs using it to mean "standard time" (i.e. EST ≠ EDT because they are different things, and it always means UTC -5)?)

My idea is to use [.tz](https://momentjs.com/timezone/) for this:

```javascript
var jun = moment("2014-06-01T12:00:00Z");
var dec = moment("2014-12-01T12:00:00Z");

jun.tz('America/New_York').format('ha z');     // 8am EDT
dec.tz('America/New_York').format('ha z');     // 7am EST`
```
----


SabreCat:
The suggested use of `tz` looks fine to me, any reason you wouldn't want to go ahead with that @Lyqst ?

SabreCat:
Went ahead and made that change. Thanks for this, @Lyqst, added to your contributor credits toward tier 3!

PR #13701, opened 2021-12-15
Fixed issue with mystery sets, pets, and mounts staying pinned after purchase. Fixes issue #12905

### LINKED ISSUE #12905: Hourglass-purchasable Equipment should be unpinned when purchased

abigley:
Added logic to look up item info and remove from pinned list by path.

Fixes #12905 

### Changes
Added logic to look up item info and remove from pinned list by path using existing getItemInfo and removeItemByPath functions. The item is removed from the pinned list after it is purchased and disappears from the user's rewards panel.

----
UUID: 11334cbf-3129-4764-8970-79fcdcd4a3e6

### DIFF
--- a/website/common/script/ops/buy/buyMysterySet.js
+++ b/website/common/script/ops/buy/buyMysterySet.js
@@ -8,6 +8,8 @@ import {
   NotFound,
 } from '../../libs/errors';
 import errorMessage from '../../libs/errorMessage';
+import { removeItemByPath } from '../pinnedGearUtils';
+import getItemInfo from '../../libs/getItemInfo';
 
 export default function buyMysterySet (user, req = {}, analytics) {
   const key = get(req, 'params.key');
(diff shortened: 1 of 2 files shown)


SabreCat:
Thanks for this, @abigley, and welcome to the ranks of the Blacksmiths at tier 1! Further tiers require multiple PRs or a larger PR to attain, but keep helping out and we'll express our gratitude accordingly!

PR #13734, opened 2022-01-07
Fix: Inconsistent Quest Unlock Behavior

### LINKED ISSUE #13013: inconsistent unlock behaviour for Mystery of Masterclassers quest line quests

### LINKED ISSUE #13206: Locked quests are still buyable even though they appear not to be
image

The Golden Knight Part 3 should be locked for the user testing (NYFairy), and it does appear to be... but the quest was purchaseable and appears in their inventory, including after refreshing.

CuriousMagpie:
Fixes #13013 and #13206

### Changes
- grouped and alphabetized all quests in preparation for future refactoring of `quests.js`
- updated quest data in `quests.js` to have multiple previous steps in sequentially unlocking quests
- updated function `unlockQuest` in `getItemInfo.js` to check for multiple steps
- improved readability of function `unlockQuest` in `getItemInfo.js`
- removed `isGearLocked` from `quests/index.vue`
- added 'if (item.locked) return;' to `selectItem(item)` in `quests/index.vue`

**Note:** The fix to prevent users from purchasing locked quest items has been implemented in the client side only. Further investigation is needed to determine how to fix this in the API and a new issue will be opened.

----


website/common/script/content/quests.js
```diff
@@ -14,1850 +14,1815 @@ const CURRENT_EVENT = find(
 );
 const userCanOwnQuestCategories = USER_CAN_OWN_QUEST_CATEGORIES;
 
+// separatihg & sorting quests by type --
+// generic unlockable, chain, masterclasser, seasonal, pet, potion, time travelers, world
+// with intent of eventually refactoring
```
SabreCat: Maybe we leave this for a separate refactoring PR, then? It'd be good to be able to review the functionally relevant additions without having to wade through all the cut-paste intended for later work.

website/common/script/libs/getItemInfo.js
```diff
@@ -9,12 +9,27 @@ import isFreeRebirth from './isFreeRebirth';
...
-    return user.loginIncentives < quest.unlockCondition.incentiveThreshold;
+
+  // checks unlockCondition.incentiveThreshold
+  if (
+    quest.unlockCondition
+    && quest.unlockCondition.incentiveThreshold
+    && user.loginIncentives < quest.unlockCondition.incentiveThreshold
+  ) return true;
+  if (user.achievements.quests) {
+    return quest.previous
```
SabreCat: After this update, do we even need to reference `quest.previous`? Given that all those should be in `quest.prereqQuests` now?

website/common/script/libs/getItemInfo.js
```diff
@@ -52,7 +67,9 @@ export default function getItemInfo (user, type, item, officialPinnedItems, lang
 
   let itemInfo;
 
-  switch (type) { // eslint-disable-line default-case
+  switch ( // eslint-disable-line default-case
+    type
+  ) {
```
SabreCat: this is odd, did the linter ask you to do this?

website/common/script/ops/buy/buyQuestGem.js
```diff
@@ -8,6 +8,7 @@ import content from '../../content/index';
 
 import errorMessage from '../../libs/errorMessage';
 import { AbstractGemItemOperation } from './abstractBuyOperation';
+// import itemInfo from '../../libs/getItemInfo';
```
SabreCat: We decided against doing this, no need to add a comment for it.

website/common/script/ops/buy/buyQuestGold.js
```diff
@@ -53,7 +46,8 @@ export class BuyQuestWithGoldOperation extends AbstractGoldItemOperation { // es
 
   checkPrerequisites (user, questKey) {
     const item = content.quests[questKey];
-    if (questKey === 'lostMasterclasser1' && !this.userAbleToStartMasterClasser(user)) {
+    if (questKey.startsWith('lostMasterclasser')
+    && !this.userAbleToStartMasterClasser(user)) {
```
SabreCat: This is the cause of one of your test failures. You removed the `userAbleToStartMasterClasser` function but still try to invoke it here.

website/common/script/ops/buy/abstractBuyOperation.js
```diff
@@ -139,6 +139,10 @@ export class AbstractGoldItemOperation extends AbstractBuyOperation {
     if (item && item.canOwn && !item.canOwn(user)) {
       throw new NotAuthorized(this.i18n('cannotBuyItem'));
     }
+
+    if (item.locked) {
+      throw new NotAuthorized(this.i18n('cannotBuyItem'));
+    }
```
SabreCat: I think this is causing the other test failure. Note that above we short-circuit if `item` isn't available--apparently we use this code for a subscriber to buy Gems using Gold, but there isn't a full `item` to represent a Gem.

So if we were doing this, the fix would be to change the condition line to `if (item && item.locked) {`
...but, I think we backed off making changes to `abstractBuyOperation` anyway? I don't think these conditions are ever going to hit, given the buy operation doesn't have access to `locked`. I think the actual correction to be made for this PR is to remove the changes to this file.

SabreCat (changes requested):
Questions and suggestions included, have a look!

website/common/script/libs/getItemInfo.js
CuriousMagpie: Probably not, we could likely change it to `return true`--I'll test that out.

website/common/script/libs/getItemInfo.js
CuriousMagpie: It did! I couldn't get it to not throw an error unless I made the change. I suspect my linter may be using slightly stricter rules.

website/common/script/ops/buy/buyQuestGem.js
CuriousMagpie: That's an artifact of a previous attempt--will delete.

website/common/script/ops/buy/buyQuestGold.js
CuriousMagpie: putting it back in--there should be a way to call `locked` into `checkPrerequisites` but I think that's going to be part of the server-side fix when it's implemented

website/common/script/ops/buy/abstractBuyOperation.js
CuriousMagpie: Took it out!

website/common/script/libs/getItemInfo.js
```diff
@@ -9,12 +9,25 @@ import isFreeRebirth from './isFreeRebirth';
...
+  // checks unlockCondition.incentiveThreshold
+  if (
+    quest.unlockCondition
+    && quest.unlockCondition.incentiveThreshold
+    && user.loginIncentives < quest.unlockCondition.incentiveThreshold
+  ) return true;
+  if (user.achievements.quests
+    && !user.achievements.quests[quest.previous]);
+  // then if we've passed all the checks
   return quest.previous;
```
SabreCat: this last part of the logic has me puzzled--
what's with the `if` statement that just ends in a semicolon?
and why is the final return still `quest.previous`?

I'm guessing it should be

```
if (quest.previous) return !(user.achievements.quests && user.achievements.quests[quest.previous]);
return false;
```

website/common/script/content/quests.js
SabreCat: Because you have the refactoring PR out now, would it be OK to skip the alphabetization changes in quests.js for this PR? It's not necessary to the fix, and it's going to introduce conflicts with the refactor.

eh, on second thought, never mind--it will actually be more trouble to revert it, because we're also adding in the `prereqQuests` property. We'll probably need to manually add those to the refactor PR once this is merged.

SabreCat (changes requested):
couple things that need a look still on the code side. I'll do some functional testing next!

website/common/script/libs/getItemInfo.js
SabreCat: or, wait, I thought you said you were removing the `quest.previous` check entirely... I am confuzzled @.@

website/common/script/ops/buy/abstractBuyOperation.js
SabreCat: looks like it's still there?

SabreCat:
This is odd--only the middle part of the Lunar Battle chain reports a need to finish the prior entry in the series:

image
image

Should they both list sequential prereqs, or neither?

Oh, I see-- the `previous` property has been changed to `previous1` on `moon3` and I don't think anything checks for that.

website/common/script/ops/buy/abstractBuyOperation.js
CuriousMagpie: I'm not seeing it? Here's the current code for `canUserPurchase` on the branch, lines 131 through 150:

```
  canUserPurchase (user, item) {
    this.item = item;
    const itemValue = this.getItemValue(item);

    const userGold = user.stats.gp;

    if (userGold < itemValue * this.quantity) {
      throw new NotAuthorized(this.i18n('messageNotEnoughGold'));
    }

    if (item && item.canOwn && !item.canOwn(user)) {
      throw new NotAuthorized(this.i18n('cannotBuyItem'));
    }
  }

  async subtractCurrency (user, item) {
    const itemValue = this.getItemValue(item);

    user.stats.gp -= itemValue * this.quantity;
  }
```

website/common/script/libs/getItemInfo.js
CuriousMagpie: I think I meant to and spaced out on it--I've made a change and am going to test it:

```
  if (user.achievements.quests
    && !user.achievements.quests[quest.prereqQuests]);
  // then if we've passed all the checks
  return quest.prereqQuests;
```

If that doesn't work, then I'll try `return false;`

CuriousMagpie:
Yeah, that `previous1` should have been `previous`...

website/common/script/libs/getItemInfo.js
CuriousMagpie: `return false;` is the way to go!

SabreCat: right! `return quest.prereqQuests` would send back the whole array, when we're expecting a true/false.

SabreCat:
Updates look good! I merged the quests refactor, though, so we will probably need to make the changes to individual quests in the refactored directory tree now. Let me know if you need a hand with that.

website/common/script/libs/getItemInfo.js
```diff
@@ -9,13 +9,26 @@ import isFreeRebirth from './isFreeRebirth';
...
+  if (quest.lvl && user.stats.lvl < quest.lvl) return true;
+
+  // checks unlockCondition.incentiveThreshold for Lunar Battle
+  if (
+    quest.unlockCondition
+    && quest.unlockCondition.incentiveThreshold
+    && user.loginIncentives < quest.unlockCondition.incentiveThreshold
+  ) return true;
+  if (user.achievements.quests
+    && !user.achievements.quests[quest.prereqQuests]);
```
SabreCat: I still don't understand what's going on here. `quest.prereqQuests` is an array, which is unlikely to behave nicely used like this. And what is this `if` block doing, ending in a semicolon with no action following?

CuriousMagpie: I removed those lines and everything seems to be working as it should when I tested it.

SabreCat (approved):

PR #13793, opened 2022-01-24
Fixes Issue13749 Alternative way

### LINKED ISSUE #13749: Negative habits don't display 0 correctly
### General Info
  * UUID: 7c94e21b-8ff2-44a7-a5f1-cfdef807a5f5
  * Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.110 Safari/537.36
  * OS: Windows

### Description
From user: 

> I have a negative habit defined, with no positive equivalent defined. I have not touched it for about a month, and it is showing "-0" instead of just zero. I understand that this probably has a minus because it is a bad habit but "-0" just looks weird/confusing because zero is not positive or negative, it is just zero.

streak-counter-negative

I checked the user's value for this habit in the database and it is currently 0.

A fix for this would remove the minus symbol from the streak if the value is equal to 0.

KrSkander:
Fixes #13749

Negative habits don't display 0 correctly #13749

Changes :
In order to fix the issue of Negative habits ,(Issue #13749), not displaying as 0 but as -0,In the Task.vue file I added a 0 check in the span v-if and added a span after it that checks if the "task.up" or "task.down" is active to display 0.

Just a value check basically :

* If the value is not 0 and (+/-) value is active show the number with a +/-

* If the value is 0 and (+/-) value is active show the number 0

(this applies to all habits as long as it's a 0 it won't have the +/- next to it)
 The Issue : 

Issue13749 on Live Version

The Solution : 

My Solution on Local Setup

----
UUID: 14267ccc-dff1-4dce-9836-e4aab5c5c8b6

### DIFF
--- a/website/client/src/components/tasks/task.vue
+++ b/website/client/src/components/tasks/task.vue
@@ -268,17 +268,19 @@
                 <span v-if="task.type === 'daily'">{{ task.streak }}</span>
                 <span v-if="task.type === 'habit'">
                   <span
-                    v-if="task.up"
+                    v-if="task.up && task.counterUp != 0"
                     class="m-0"
                   >+{{ task.counterUp }}</span>
+                  <span v-else-if="task.up">0</span>
                   <span
                     v-if="task.up && task.down"
                     class="m-0"
                   >&nbsp;|&nbsp;</span>
                   <span
-                    v-if="task.down"
+                    v-if="task.down && task.counterDown != 0"
                     class="m-0"
                   >-{{ task.counterDown }}</span>
+                  <span v-else-if="task.down">0</span>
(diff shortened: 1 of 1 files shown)


CuriousMagpie:
Hey there, since this was assigned to someone else, I'm going to go ahead and close this PR. Please feel free to look through the "help wanted" tag and see if there's anything there that strikes your fancy (and leave a comment so we can assign it to you).

KrSkander:
@CuriousMagpie Thank you for the update ! I will do keep an eye on it

PR #13816, opened 2022-01-31
Social auth enhancements -- prerequisites for Facebook removal

SabreCat:
This is a recreation of reverted PR #13683, to receive further testing on Delta.


SabreCat:
We've identified a couple of issues that will need to be addressed before this goes live:

* Password reset emails against an account with no local authentication are failing
* If a social auth account gets an automatically assigned email, the user can't change that email or add a password

PR #13818, opened 2022-02-03
Update: Quest Refactoring

CuriousMagpie:
The existing `quests.js` file is extremely difficult to navigate and cumbersome to update. This PR moves all the quest objects into constants based on quest type.

### Changes
- created `QUEST_GENERIC`, `QUEST_MASTERCLASSER`, `QUEST_SERIES`, `QUEST_SEASONAL`, `QUEST_PETS`, `QUEST_POTIONS`, `QUEST_TIMETRAVEL`, and `QUEST_WORLD` to hold quest objects. Quests have been generally alphabetized within each constant, with the exception is `QUEST_MASTERCLASSER`, where quests are listed in the order of the storyline.
- Quest objects is now called from the constants into `quests.js`, where they are combined into a single object using the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax).

----


SabreCat:
OK, after a find-and-fix on an `import` statement and taking the opportunity to fix a production bug with seasonal quests being available out of season, this looks ok!

My one concern is that I don't agree with `content/constants/questXXXX.js`, organization-wise? we have `content/gear` and `content/appearance`, which are technically constants too but they each have their own space within the app. Why not make a directory `website/common/script/content/quests` and have `generic.js`, `masterclasser.js`, `pets.js` and so on, in there?

SabreCat (changes requested):
Code is working, I have one suggestion about the file hierarchy that I added in PR conversation!

CuriousMagpie:
> My one concern is that I don't agree with `content/constants/questXXXX.js`, organization-wise? we have `content/gear` and `content/appearance`, which are technically constants too but they each have their own space within the app. Why not make a directory `website/common/script/content/quests` and have `generic.js`, `masterclasser.js`, `pets.js` and so on, in there?

This is a great suggestion and I went ahead and did that--I think I didn't initially because I wasn't sure if there was a reason why `website/common/script/content/quests` shouldn't exist.

SabreCat (approved):

PR #13821, opened 2022-02-05
Add automatic year detection on appFooter

rarysson:
### Changes
Hi, happy belated new year, I was thinking to just replace 2021 to 2022, but to prevent this happening again I've replaced the year for a variable that get the current year.
I've used the built-in Date to get the current year, but I was not sure if I could use it or I should use moment.

If there's anything wrong just let me know.

----
UUID: 4a7ed516-186a-4b97-ba86-b28b311d88ab


SabreCat:
Needed to clean up some linter errors, but not a bad idea! Thanks for this, @rarysson -- updated your contributor credits!

PR #13843, opened 2022-02-09
Adminpanel and revamped permissions

phillipthelen:
This adds some changes to the existing admin panel PR #12652. Mainly just some checks to restrict permission changing etc.

This also includes changes to how permissions are handled in general to allow for more granular permissions


phillipthelen:
For some reason the Push tests are all failing, but the PR test run works?

PR #13845, opened 2022-02-12
Fixed column help text being overlaid by rewards and tasks. Fixes #13542

### LINKED ISSUE #13542: Task column help text hidden by tasks
[//]: # (Before logging this issue, please contact site administrators from "Report a Bug" in the Habitica website's Help menu. If a GitHub issue is needed, staff will let you know and will most likely log one on your behalf. It is recommended that you don't create a new issue unless advised to.)

[//]: # (Bugs in the mobile apps can be reported via Menu > About > Support.)

[//]: # (If you have a feature request, use "Help > Request a Feature", not GitHub.)

[//]: # (Fill out relevant information - UUID is found from the Habitica website at User Icon > Settings > API)
### General Info
  * UUID: 1643f1c4-3199-47b6-b0fc-99a3e567c092 
  * Browser: Firefox
  * OS:

### Description
With some numbers of tasks, the help text for the Rewards column can be overlaid by the reward icons:

image

Ideally this should flow downward so it remains readable, or be hidden instead of partially covered.

Crats1:
Fixes #13542

### Changes

Fixed issue with column help text being overlaid by the reward icons or tasks by not rendering the help text if there are 3 or more tasks in the column.

----
UUID: 5672b3a8-b851-4196-aac3-27654fe54ec3

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1512,18 +1512,18 @@
       }
     },
     "@types/eslint": {
-      "version": "8.2.1",
-      "resolved": "https://registry.npmjs.org/@types/eslint/-/eslint-8.2.1.tgz",
-      "integrity": "sha512-UP9rzNn/XyGwb5RQ2fok+DzcIRIYwc16qTXse5+Smsy8MOIccCChT15KAwnsgQx4PzJkaMq4myFyZ4CL5TjhIQ==",
+      "version": "8.4.1",
+      "resolved": "https://registry.npmjs.org/@types/eslint/-/eslint-8.4.1.tgz",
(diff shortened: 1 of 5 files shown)


SabreCat:
Unfortunately this fix doesn't successfully address the situation that happens in Rewards--
and it's even a bit of a degradation on the first three columns, because the help text vanishes at three tasks even if there's plenty of vertical space for it, which is not the case without implementing this commit.

I think what's actually needed is to detect when the combination of reward tasks + pinned items runs to a height that collides with the help text, rather than relying on a strict count of tasks. It's the pinned items that give us trouble, not the tasks.

SabreCat (changes requested):
1. Doesn't actually correct the linked issue, alas
2. Unless the package updates are necessary to the fix at hand, don't commit changes to `package.json` and `package-lock.json`.

PR #13860, opened 2022-02-14
Show "Next Hourglass" Month

negue:


test/common/libs/cron.test.js
```diff
@@ -181,4 +181,63 @@ describe('cron utility functions', () => {
...
+      };
+    }
+
+    it('offset 0, next date in 3 months', () => {
+      const user = baseUserData(60, 0, 'group_plan_auto');
+
+      const planContext = getPlanContext(user, now);
+
+      expect(planContext.nextHourglassDate)
+        .to.be.sameMoment('2022-08-10T02:00:00.144Z');
```
SabreCat, line 222: Does this happen on the renewal date? or should it be at the start of the month because this is handled in `grantEndOfTheMonthPerks`?

hmm ok--so the subscription form does say just the month, and mentions that the user will get the Hourglass within 3 days of the start of the month, but you can still mouse over the month text and get a potentially misleading specific date.

SabreCat (changes requested):

test/common/libs/cron.test.js
negue: this is just to get the "context" - the nextHourglassDate is only used in the UI not in the server / cron method - I just added this property since it was mostly in the same-ish context 

the hover was for me for potential debugging - I can remove this hover text as well

SabreCat: ok! yeah, let's just remove any debugging-only code/functionality and it should be good, then!

negue: removed the tooltip

SabreCat (approved):

PR #13862, opened 2022-02-15
Improve flows for social auth users

phillipthelen:
- allows the reset password field to also accept a username
 - fixes resetting password without an email set 
 - allows email to be changed without a password if the user has no password


website/client/src/components/settings/site.vue
```diff
@@ -291,14 +291,22 @@
             </li>
           </ul>
           <hr>
-          <div v-if="!user.auth.local.email">
-            <p>{{ $t('addLocalAuth') }}</p>
+          <div v-if="!user.auth.local.has_password">
```
SabreCat, line 294: Is this supposed to be `hashed_password`? I'm not seeing a `has_password` field or method anywhere.

SabreCat (changes requested):
Just one question that might need a trivial fix. I'll send this to Delta for QA once that's answered!

SabreCat (approved):

website/common/locales/en/front.json
```diff
@@ -125,7 +125,7 @@
   "passwordConfirmationMatch": "Password confirmation doesn't match password.",
   "minPasswordLength": "Password must be 8 characters or more.",
   "passwordResetPage": "Reset Password",
-  "passwordReset": "If we have your email on file, instructions for setting a new password have been sent to your email.",
+  "passwordReset": "If we have your emailor username on file, instructions for setting a new password have been sent to your email.",
```
SabreCat: trivial typo here

website/common/locales/en/settings.json
```diff
@@ -134,6 +134,7 @@
     "saveCustomDayStart": "Save Custom Day Start",
     "registration": "Registration",
     "addLocalAuth": "Add Email and Password Login",
+    "addPasswordAuth": "Add Password ",
```
SabreCat: odd trailing space?

website/server/controllers/api-v3/auth.js
```diff
@@ -341,7 +341,15 @@ api.resetPassword = {
     if (validationErrors) throw validationErrors;
 
     const email = req.body.email.toLowerCase();
-    const user = await User.findOne({ 'auth.local.email': email }).exec();
+    const user = await User.findOne({
+      $or: [
+        { 'auth.local.username': email.replace('@', '') },
```
SabreCat: is there a risk of this picking up someone who coincidentally has a username that's similar to someone else's email address?

SabreCat (changes requested):
couple of typos we can clean up whenever, one logic design question

SabreCat:
Also looks like one of the integration tests is landing the wrong error message

website/server/libs/auth/index.js
```diff
@@ -128,7 +128,11 @@ async function registerLocal (req, res, { isV3 = false }) {
   }, { 'auth.local': 1 }).exec();
 
   if (user) {
-    if (email === user.auth.local.email) throw new NotAuthorized(res.t('emailTaken'));
+    if (existingUser) {
+      if (email === user.auth.local.email && existingUser._id !== user._id) throw new NotAuthorized(res.t('emailTaken'));
+    } else {
```
SabreCat: I think this is what's causing the test failures--
`user` comes from an `$or` search, either we find the email address or we find the lowercase username;
but if we find it on the lowercase username, we end up here, reporting that the email address is already taken even if the email address passed the uniqueness check.

website/server/controllers/api-v3/auth.js
phillipthelen: Good point. I. don't think this would really be an issue, but the more correct solution either way would be to only replace the @ if it's the first character

SabreCat (approved):

PR #13872, opened 2022-02-22
Fix validation for cancelling android subscriptions

phillipthelen:
The server has been using the wrong property/an inexisting property to check if a sub is still active before cancelling it. This change fixes that and also adds a new check to confirm the fix.


test/api/unit/libs/payments/google.test.js
```diff
@@ -256,7 +256,7 @@ describe('Google Payments', () => {
           expirationDate,
         });
       iapGetPurchaseDataStub = sinon.stub(iap, 'getPurchaseData')
-        .returns([{ expirationDate: expirationDate.toDate() }]);
+        .returns([{ expirationDate: expirationDate.toDate(), autoRenewig: false }]);
```
SabreCat, line 259: new typo!

test/api/unit/libs/payments/google.test.js
```diff
@@ -325,5 +325,25 @@ describe('Google Payments', () => {
         headers,
       });
     });
+
+    it('should not cancel a user subscription with autorenew', async () => {
+      iapGetPurchaseDataStub = sinon.stub(iap, 'getPurchaseData')
+      .returns([{ autoRenewig: true }]);
```
SabreCat, line 331: typo also here

SabreCat (changes requested):
Because this is meant as a hotfix, I changed the target branch to `release`. Looks like the main thing you'll need to adjust as a result is to reset these package files to the latest release commit.

Also found one issue in the tests!

SabreCat:
Cherry-picked to release and fixed up!

PR #13884, opened 2022-03-11
Update: Revised text string for zodiacZookeeper and birdsOfAFeather achievements

CuriousMagpie:
Updated `achievementZodiacZookeeperText` and `achievementBirdsOfAFeatherText` to include the phrase "all standard colors" so the achievement requirements are clearer for users.

----


SabreCat:
@CuriousMagpie Looks like this needs to be updated to include Roosters and Peacocks? Thus the conflict..

SabreCat (approved):

PR #13892, opened 2022-03-17
Fix cron not running if previous run failed

phillipthelen:


phillipthelen:
Same thing as with the admin panel PR. not sure why the push tests are failing

SabreCat (approved):

PR #13900, opened 2022-03-21
On-the-fly fix for tasks missing from tasksOrder

SabreCat:
Adds a check in the task move-to-position API route that, if the UUID for the task being moved is missing from the relevant `tasksOrder` array, looks up the tasks for that user/group/challenge and type and adds any missing entries back to the array before performing the requested move.

This should help regenerate sort arrays recently emptied in a bad database operation, and catch other scenarios where potentially a task exists but has no corresponding entry (as sometimes can be spotted using e.g. the Task Adjustor tool). The user experience should be better than the current scenario, where unsorted tasks jump around unpredictably and it's necessary to move every unpaired task individually before the list as a whole starts to behave.

### DIFF
--- a/website/server/controllers/api-v3/tasks.js
+++ b/website/server/controllers/api-v3/tasks.js
@@ -838,6 +838,28 @@ api.moveTask = {
 
     // In memory updates
     const order = owner.tasksOrder[`${task.type}s`];
+    if (order.indexOf(task._id) === -1) { // task is missing from list, list needs repair
+      const taskListQuery = { type: task.type };
+      if (group) {
+        taskListQuery['group.id'] = owner._id;
+      } else if (challenge) {
+        taskListQuery['challenge.id'] = owner._id;
+      } else {
+        taskListQuery.userId = owner._id;
+      }
+      const taskList = await Tasks.Task.find(
+        taskListQuery,
+        { _id: 1 },
+      ).exec();
+      for (const foundTask of taskList) {
+        if (order.indexOf(foundTask._id) === -1) {
+          order.push(foundTask._id);
+        }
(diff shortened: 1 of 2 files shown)


website/server/controllers/api-v3/tasks.js
```diff
@@ -838,6 +838,28 @@ api.moveTask = {
 
     // In memory updates
     const order = owner.tasksOrder[`${task.type}s`];
+    if (order.indexOf(task._id) === -1) { // task is missing from list, list needs repair
+      const taskListQuery = { type: task.type };
+      if (group) {
+        taskListQuery['group.id'] = owner._id;
+      } else if (challenge) {
+        taskListQuery['challenge.id'] = owner._id;
```
Alys, line 848: I believe this search will return both the Challenge's own tasks and copies of the tasks that are in participants' task lists. It would need to also check for userId not existing, to get only the former.

From memory, something similar may happen for the group search too.

website/server/controllers/api-v3/tasks.js
```diff
@@ -838,6 +838,28 @@ api.moveTask = {
 
     // In memory updates
     const order = owner.tasksOrder[`${task.type}s`];
+    if (order.indexOf(task._id) === -1) { // task is missing from list, list needs repair
+      const taskListQuery = { type: task.type };
+      if (group) {
+        taskListQuery['group.id'] = owner._id;
```
Alys, line 845: This group fix might also be beneficial in website/server/controllers/api-v3/tasks/groups.js although I'd guess that's a lower priority than the current issue, so perhaps just a TODO comment here?

PR #13910, opened 2022-03-28
Redesign: "Day Start Adjustment" 

negue:
This merges  the "Custom Day Start" and "Timezone" Sections into one Component


website/common/locales/en/settings.json
```diff
@@ -40,10 +40,12 @@
     "xml": "(XML)",
     "json": "(JSON)",
     "customDayStart": "Custom Day Start",
+    "adjustment": "Adjustment",
+    "dayStartAdjustment": "Day Start Adjustment",
     "sureChangeCustomDayStartTime": "Are you sure you want to change your Custom Day Start time? Your Dailies will next reset the first time you use Habitica after <%= time %>. Make sure you have completed your Dailies before then!",
     "customDayStartHasChanged": "Your custom day start has changed.",
     "nextCron": "Your Dailies will next reset the first time you use Habitica after <%= time %>. Make sure you have completed your Dailies before this time!",
-    "customDayStartInfo1": "Habitica defaults to check and reset your Dailies at midnight in your own time zone each day. You can customize that time here.",
+    "customDayStartInfo1": "Habitica defaults to check and reset your Dailies at 12:00 AM in your own time zone each day. Shift when your day switches over past the default time.",
```
SabreCat: The new copy text here is more confusing than the original--I think we need to consult with the UX folks on what this language should be.

website/common/locales/mn/settings.json
```diff
@@ -159,7 +159,7 @@
     "amazonPayments": "Amazon Payments",
     "amazonPaymentsRecurring": "Ticking the checkbox below is necessary for your subscription to be created. It allows your Amazon account to be used for ongoing payments for <strong>this</strong> subscription. It will not cause your Amazon account to be automatically used for any future purchases.",
     "timezone": "Time Zone",
-    "timezoneUTC": "Habitica uses the time zone set on your PC, which is: <strong><%= utc %></strong>",
+    "timezoneUTC": "Your time zone set by your computer, which is: <strong><%= utc %></strong>",
```
SabreCat: This is not as odd as the other entry, but could also use another set of writerly eyes I think!

SabreCat (commented):
whoops, this wasn't even supposed to be reviewed yet. I'll change this to Draft so it's clearer where it's at!

website/common/locales/en/settings.json
negue: @Tressley

Alys: "12:00 AM" isn't the best term to use because it means different things to different people. https://en.wikipedia.org/wiki/12-hour_clock#Confusion_at_noon_and_midnight shows some examples.

Can we keep using "midnight" for clarity?

Alys: By the way, changing the words does mean that all the translations have to be changed, and also all the wikis (we typically include screenshots and/or text copies of words on the site). That's not worth considering if the words really do need to be changed, but if it's decided that a change isn't needed, that means less work for the contributors.

Tressley: We were experimenting with new language and I agree that the new copy is reading a little odd. Additionally, we were using "12:00 AM" to connect from the time actually selected from the list. Taking feedback into account, here is revised copy that flows better and reverts back to using midnight:
> Habitica checks and resets your Dailies at midnight in your own time zone each day. You can adjust when that happens past the default time here.

website/common/locales/mn/settings.json
Tressley: @negue -- This should be:
> Your time zone is set by your computer, which is:

website/common/locales/en/settings.json
saraolson: some context for the changes: we wanted to edit the wording and UI of the feature to make it more clear that its pushing the time back from the original start time, due to user reports of people setting their day start to later in the day and being confused why their dailies aren't showing on their 'proper' days. so we do think asking for another pass of translation is worth it to hopefully clear up some confusion.

CuriousMagpie: Could it say "12:00 midnight in your own time zone"?

website/common/locales/mn/settings.json
negue: just saw that this is even the wrong file ... woops - reverting that and changing it in the correct one

website/common/locales/en/settings.json
Tressley: I think we'll move forward with just "midnight" for now and adjust should we hear about any static.

website/common/locales/en/settings.json
```diff
@@ -162,8 +164,8 @@
...
-    "timezoneInfo": "If that time zone is wrong, first reload this page using your browser's reload or refresh button to ensure that Habitica has the most recent information. If it is still wrong, adjust the time zone on your PC and then reload this page again.<br><br> <strong>If you use Habitica on other PCs or mobile devices, the time zone must be the same on them all.</strong> If your Dailies have been resetting at the wrong time, repeat this check on all other PCs and on a browser on your mobile devices.",
+    "timezoneUTC": "Your time zone is set by your computer, which is: <strong><%= utc %></strong>",
+    "timezoneInfo": "If that time zone is wrong, first reload this page using your browser's reload or refresh button to ensure that Habitica has the most recent information. If it is still wrong, adjust tcustomDayStartInfo1he time zone on your PC and then reload this page again.<br><br> <strong>If you use Habitica on other PCs or mobile devices, the time zone must be the same on them all.</strong> If your Dailies have been resetting at the wrong time, repeat this check on all other PCs and on a browser on your mobile devices.",
```
SabreCat: looks like there was a weird copy-paste error here? `tcustomDayStartInfo1he` ?

SabreCat (changes requested):
Just the one typo I think, nothing too alarming here! I can fix this myself and merge it if you don't get the chance within ~24 hours.

website/common/locales/en/settings.json
negue: oh wow ... how did that happen :O - yeah somehow I pasted it there ... woops :D

negue: fixed

SabreCat (approved):

PR #13930, opened 2022-04-13
Removes the Year Sub Menu for Unlocked Background (Fixes Issue #13373)

### LINKED ISSUE #13373: Backgrounds modal: When "Hide locked backgrounds" is on, year tabs are still there despite all backgrounds being on one page.
### Description
When selecting "Hide locked backgrounds" on the Backgrounds modal, all unlocked backgrounds are shown on one page - and this is definitely a feature, please leave that as it is! But the year tabs (2021 - 2014) are still there and can be selected as if the backgrounds were grouped by year. This is misleading and makes no sense. (The tab selection does not change anything, all tabs show the same complete collection of all unlocked backgrounds of all years, plus the free ones and the steampunk ones if they are unlocked.)

I'd suggest not showing the tabs when "Hide locked backgrounds" is switched on, or making them unclickable and changing the tab labeling, or otherwise making the design and text consistent with the fact that these are all unlocked backgrounds from all years (and the free ones and the steampunk ones if one has them) on one page. Also, while this is being changed, it would also be nice to give users a hint that this feature exists, by not just labeling the toggle "Hide locked backgrounds", but "Hide locked backgrounds and show all unlocked backgrounds on one page", or just "Show unlocked backgrounds on one page", or similar. (Otherwise one can just discover this very useful "side effect" by accident.)

In my opinion, this is a very low-priority bug with no need to fix it any time soon. But now it is documented here and can be linked to from the wiki and from "Ask a Question"/"Report a Bug" answering posts. And if there is ever a time when Habitica programmers got nothing more urgent to do or a new programmer wants to fix a small thing, this is on the list.

I created this issue with permission from @shanaqui after first discussing it in the Report a Bug guild.

### Screenshot Link (Temporary)
This screenshot is on snipboard.io and may therefore be deleted automatically after a while (if I understand correctly how snipboard.io works). But it can easily be recreated.

[https://snipboard.io/NxL37C.jpg](https://snipboard.io/NxL37C.jpg)

### General Info
  * UUID: 1aff9116-34b0-438c-9877-06038bdce161
  * Browser: Chrome, Firefox
  * OS: Windows

wllyng:

Fixes #13373

### Changes
The changes that were made were adding the conditional v-if="!filterBackgrounds" to the submenu so that the years would only show up if the "Show Unlocked Backgrounds" toggle was not toggled.  We also changed the "Hide Locked Background" toggle string to say 
"Show Unlocked Backgrounds" instead

asdaddsda

adasddddd

----
UUID: b271337f-3964-4e58-92e0-af9cbbeba79a

### DIFF
--- a/website/client/src/components/creatorIntro.vue
+++ b/website/client/src/components/creatorIntro.vue
@@ -159,7 +159,7 @@
           <toggle-switch
             v-model="filterBackgrounds"
             class="backgroundFilterToggle"
-            :label="'Hide locked backgrounds'"
+            :label="'Show Unlocked Backgrounds'"
           />
         </div>
         <div
@@ -247,6 +247,7 @@
           </div>
         </div>
         <sub-menu
+          v-if="!filterBackgrounds"
           class="text-center"
           :items="bgSubMenuItems"
           :active-sub-page="activeSubPage"


SabreCat:
This looks solid, I just need to check with our UX staff to make sure this is the final copy we want. Once we've got that settled, we'll probably want to change that string from hardcoded English to a translatable token using the `t` function.

SabreCat:
Went ahead and made the string token change myself, but the fix itself looks good! @wllyng that user ID doesn't appear to be correct, can you double-check? Or let us know your Habitica @ -username instead!

PR #13942, opened 2022-04-20
Fix XML Data Export Error by Modifying XML Format

johnl2102:
… XML + XML Marshall tests

Tentative fix for #12971

### Changes

Changes the Export XML function in user settings to abide by the format suggested in #12971. Purchased items' IDs are now enclosed in tags of that item's type (e.g. <skin>eb052b</skin>), rather than the original way which produced invalid keys and caused error when exporting XML.

This is done by copying the json object passed into the function marshallUserData() in xmlMarshaller.js, and modifying its format. Then it is converted to XML and returned.

Also included in PR are test cases testing the new functionality.

----
UUID: 096a7abe-b408-4441-a3d9-039bba61505e
(uuid obtained from local dev build)


SabreCat:
Hi @johnl2102, looks like there's still some lint errors here. I pushed a possible fix to your branch, but it doesn't appear to have updated the PR automatically--possibly because it's on your `develop` rather than a work branch. Try adding that commit here and we'll give this another look!

Also, a UUID from your local development environment won't help us award you contributor credit on habitica.com fyi!

SabreCat:
oh, it went through eventually--odd!

Anyway, we'll get this sent on to our staging environment and then to production! If you want contributor credit, let us know a user ID or @-username from habitica.com.

PR #13946, opened 2022-04-23
Fix char values issue

### LINKED ISSUE #13774: Fix Character Values should not allow operands
### Description
User email:

> I wanted to reach out about the fact that if you try to fix a value by adding a certain amount to it directly (ex 11+ 5 health) then the game will assume that the value should be zero, and will kill you, but you don’t actually die, you just enter a glitched state where whenever you reload the game you get the death screen and you have zero health. 

I was able to replicate this in development:

Added "+ 25" to the health field:
Screen Shot 2022-01-20 at 12 27 21 PM

You died screen:
Screen Shot 2022-01-20 at 12 27 44 PM

If you X out of the modal without taking the option to refill your health, then your character appears like this on screen--you lose a piece of equipment but you do not lose your gold when this happens:
Screen Shot 2022-01-20 at 12 28 00 PM

I also tested this on the other fields in FCV and was able to wipe out my experience, gold, mana, and 21 day streaks. I was not able to use operands in the level field.

The fix here is probably some smarter validation that disallows operands in all fields, not just level.

vanathi-g:
Fixes #13774

### Changes
Added a method validateInputs which validates the following conditions:
- All the input values are not empty strings (happens when operands are entered)
- level is an integer >= 1
- streak is an integer >= 0

Since negative values are valid under certain conditions for HP and EXP, positive number validation is not performed for those fields as well as for GP and MP (just to be safe). 

If any of the conditions fail, the fields are reset to the last correct value and the method returns false. On clicking the save button, if incorrect values are entered, the respective fields are reset and the modal remains open.

On discarding changes when incorrect values are entered, the method is called to ensure the incorrect fields don't become blank.

----
UUID: e6563bdb-d493-4207-a3f6-8b4da3e547e1


SabreCat:
Looks good! Inline validations might have been a shinier way to accomplish this, but as this is sort of an under-the-hood feature to begin with, a little bit of scuffed UX is not so terrible. I've added this to your contributor credits--welcome to Blacksmith tier 2!

vanathi-g:
Hi @SabreCat , I checked the Hall of Contributors page! I think a 'title' hasn't been assigned to me so it just says 'Friend undefined'. Could you assign me a title as well, thanks!

SabreCat:
Ah! whoops! There you go, should be better now!

SabreCat:
Huh, looks like there was a bug in the tool we'd been using to award titles 😅 Fixed now!

PR #13947, opened 2022-04-24
Habit streak Fix 

### LINKED ISSUE #13749: Negative habits don't display 0 correctly

KrSkander:
Fixes https://github.com/HabitRPG/habitica/issues/13749

Negative habits don't display 0 correctly https://github.com/HabitRPG/habitica/issues/13749

Changes :
In order to fix the issue of Negative habits ,(Issue https://github.com/HabitRPG/habitica/issues/13749), not displaying as 0 but as -0, In the Task.vue file I added a 0 check, so if one value is null the other won't display +/- sign, if both are null  

Just a value check basically :

If the value is not 0 and (+/-) value is active show the number with a +/-

If the value is 0 and (+/-) value is active show the number 0

if the Up/Down value is 0 remove the (+/-) for the other none null value

(this applies to all habits as long as it's a 0 it won't have the +/- next to it)
The Issue :

Issue13749

My Solution : 

My Solution

@SabreCat @Tressley 
----
UUID:
14267ccc-dff1-4dce-9836-e4aab5c5c8b6


SabreCat:
Looking at your screenshot, this is pretty close to what we need--but a display of `0 | 2` or `2 | 0` isn't correct on a two-direction Habit. The direction sign should still appear on the nonzero value.

KrSkander:
> Looking at your screenshot, this is pretty close to what we need--but a display of `0 | 2` or `2 | 0` isn't correct on a two-direction Habit. The direction sign should still appear on the nonzero value.

Ah that's no problem ! The first fix had the signs I thought based on the suggestions you guys wanted them off when there is a 0.
I'll fix it right away

KrSkander:
@SabreCat I've made the changes, this is how the habits shall look like :

MySolution updated

SabreCat:
Made a tiny layout adjustment, but I think this looks good! I'll merge it into our staging environment for Tressley to peek at, then it'll go live next week!

I've added this to your contributor credits on the site. Welcome to the ranks of the Blacksmiths at tier 1!

PR #13952, opened 2022-04-26
Server setting to disallow chat from new accounts

SabreCat:

Partial implementation of needs from #8063.

### Changes

Adds a new server environment variable, ACCOUNT_MIN_CHAT_AGE. Accounts created less than ACCOUNT_MIN_CHAT_AGE minutes ago will encounter an error when attempting to post a chat message.

- [x] Fix various tests disrupted by the timeout
- [x] Add analytics logging when the error occurs, for better incident tracking


SabreCat:
@Alys since this is a technical PR related to community management, would you be up to reviewing it?

Alys:
I've reviewed this and it works as written.

I think it's worth releasing this now because it will give us some extra control over spammers. We can then modify it further as needed.

On Habitica's server, we should set ACCOUNT_MIN_CHAT_AGE set to 0 of course before/at the same time as this is released.
I think it's worth setting it to 0 in config.json.example too, because most Blacksmiths won't be working on this feature, and it will get in the way of sending test messages, and could cause confusion for any Blacksmiths who aren't aware of the feature.

